PR #1424, opened 2019-05-07
[Security] [Conf] add flag to load path from env-var

jaeyun-jung:
Parse ini file and add sub-plugins path from env-var. (default false)
For unittest, add temorary configuration to read env variables.


jaeyun-jung:
How can I set conf file path in yocto build?

jaeyun-jung:
Please review again. I updated this PR to parse the conf file first.

PR #1433, opened 2019-05-09
[Conf/Tensorflow] remove default flag in conf setting

jaeyun-jung:
Remove default flag about tensorflow memory optimization.
In tensorflow sub-plugin, read this option using custom-bool function.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.c
@@ -81,7 +81,8 @@ tf_loadModelFile (const GstTensorFilterProperties * prop, void **private_data)
     }
   }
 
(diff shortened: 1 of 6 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.c
```diff
@@ -81,7 +81,8 @@ tf_loadModelFile (const GstTensorFilterProperties * prop, void **private_data)
     }
   }
 
-  tf_mem_optmz = nnsconf_get_value_bool (NNSCONF_VAL_TF_MEM_OPTMZ);
+  tf_mem_optmz = nnsconf_get_custom_value_bool ("tensorflow",
```
helloahn: The method `nnsconf_get_custom_value_bool()` reads `nnstreamer.ini.in` file and set the variables.
If we read the `mem_optimization` value only from `.ini` file, however, we can't set the option with the environment variable.

PR #1438, opened 2019-05-13
[Meson] remove unnecessary option

jaeyun-jung:
For opencv test, 'enable-opencv-test' is unnecessary option.
Check dependency of opencv pkg and build libraries when found the dependency.


nnstreamer_example/custom_example_opencv/meson.build
```diff
@@ -1,17 +1,17 @@
-opencv_dep = dependency('opencv')
+opencv_dep = dependency('opencv', required: false)
```
kparichay: Will this work in the case when opencv is installed from source to a local location (without updating pkg-config)?
You can consider using ldconfig to verify is opencv is available (refer to #1336)

kparichay: Yes, the old script would break as well.

PR #1447, opened 2019-05-15
[pytorch] Added pytorch subplugin support @open sesame 06/04 11:35

kparichay:
1. Added pytorch subplugin support to run inference of networks
2. Added support to run on gpu for the inference
3. Added test cases for the subplugin

Added a sample file torch saved file used to test the code

**Self evaluation:**


gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -55,6 +55,7 @@ typedef struct
   gchar *pathDECODERS[CONF_SOURCES];       /**< directory paths for DECODERS */
   gchar *pathCUSTOM_FILTERS[CONF_SOURCES]; /**< directory paths for CUSTOM FILTERS */
   gchar *valueTF_MEM_OPTMZ[CONF_SOURCES];  /**< value of TF_MEM_OPTMZ */
+  gchar *valueTORCH_USE_GPU[CONF_SOURCES]; /**< value of TORCH_USE_GPU */
```
jaeyun-jung: Plz consider to use custom-bool function, not a default conf str.
(See PR #1433 for tensorflow changes)

kparichay: Updated.

meson_options.txt
```diff
@@ -3,6 +3,8 @@ option('install-test', type: 'boolean', value: false)
 option('enable-tensorflow-lite', type: 'boolean', value: true)
 option('enable-tensorflow', type: 'boolean', value: true)
 option('enable-tensorflow-mem-optmz', type: 'boolean', value: true)
+option('enable-pytorch', type: 'boolean', value: true)
+option('enable-pytorch-use-gpu', type: 'boolean', value: false)
```
myungjoo: We need some "explanation" for this option; readers may think GPU usage is deteremined at build-time. But, this is rather a default value that can be overriden at run-time. (adding a line of comment should suffice.)

packaging/nnstreamer.spec
```diff
@@ -178,8 +178,9 @@ enable_tf=true
 %else
 enable_tf=false
 %endif
+enable_torch=false
 
-meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} --libdir=%{_libdir} --bindir=%{nnstexampledir} --includedir=%{_includedir} -Dinstall-example=true -Denable-tensorflow=${enable_tf} %{api} -Denable-env-var=false build
+meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} --libdir=%{_libdir} --bindir=%{nnstexampledir} --includedir=%{_includedir} -Dinstall-example=true -Denable-tensorflow=${enable_tf} -Denable-pytorch=${enable_torch} %{api} -Denable-env-var=false build
```
myungjoo: Use %, not $ in .spec script.

Actually, ${enable_tf} is incorrect as well.

meson_options.txt
kparichay: Done.

packaging/nnstreamer.spec
kparichay: Done
Will update enable_tf in a different PR

tests/nnstreamer_filter_pytorch/runTest.sh
```diff
@@ -0,0 +1,81 @@
...
+	source $search
+	printf "${Blue}Independent Mode${NC}
+"
+fi
+
+# This is compatible with SSAT (https://github.com/myungjoo/SSAT)
+testInit $1
+
+# Test constant passthrough custom filter (1, 2)
+PATH_TO_PLUGIN="../../build"
```
leemgs: Where is network model files of Squeezenet when the pytorch sub-plugin will be enabled?
Which package (e.g., RPM, DEB, etc) that put the model files for runTest.sh?

kparichay: Pytorch does not give model files to download (Squeezenet model in this case).
They are embedded in torchvision library (different from pytorch).
Currently, squeezenet model is added as part of this PR, but as myungjoo said, its too big (4.94MB). I have posted the possible solutions in comment below.

kparichay:
> Please use lighter models for unit-test: BIN +4.94 MB tests/test_models/models/pytorch_squeezenet_1.1_224.pt

This is the smallest model for object classification based on imagenet (this allows the test case between tflite and pytorch to be same). There are two solutions - 
1. Generate the model at run-time as this is using the pretrained model provided by torch. This will require torchvision package (which will have to be packaged as deb and uploaded to launchpad)
2. change the unit-test to a simpler form (like MNIST). This would still require storing the model file but the model file would be quite small (150KB for MNIST)
Which one would you recommend?

packaging/nnstreamer.spec
```diff
@@ -179,7 +179,9 @@ enable_tf=true
 enable_tf=false
 %endif
 
-meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} --libdir=%{_libdir} --bindir=%{nnstexampledir} --includedir=%{_includedir} -Dinstall-example=true -Denable-tensorflow=${enable_tf} %{api} -Denable-env-var=false -Denable-symbolic-link=false build
+%define enable_torch false
```
jaeyun-jung: [Question] enable_torch, is this always false in tizen? then add option -Denable-pytorch=false.

kparichay: it will be always false for now (till pytorch is supported in tizen)
Updated.

tests/nnstreamer_filter_pytorch/runTest.sh
kparichay: Using MNIST model file (150kb), which are added as a PR.
No official place to download trained models for pytorch for MNIST.

kparichay:
V2:
Add pytorch in nnstreamer.mk
install pytorch tests
add nnstreamer-pytorch package

V3:
Disable pytorch for tizen build until its supported

V4:
Disable pytorch tests when disabling pytorch builda

V5:
Corrected variable in spec file
Explanation for enable-pytorch-use-gpu option

V6:
Using mnist based unit-test case instead of imagenet based model
to reduce the model file size in use

kparichay:
V7:
Using input and output properties of tensorfilter than specify in the model file

PR #1454, opened 2019-05-23
[Source/AMC] Source element for Android MediaCodec (AMC) @open sesame 6/3 11:23 

dongju-chae:
This PR includes codes of a source element for Android MediaCodec (AMC).
It's related to issue https://github.com/nnsuite/nnstreamer/issues/1281.
Also, I made another PR for example app (https://github.com/nnsuite/nnstreamer-example/pull/33).

**Changelog:**
1. Fix doxygen errors
2. Add README.md
3. Fix minor bugs
4. Change the termination logic

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/android_source/gstamcsrc.c
+++ b/ext/nnstreamer/android_source/gstamcsrc.c
@@ -1,0 +1,990 @@
+/**
+ * GStreamer Android MediaCodec (AMC) Source
+ * Copyright (C) 2019 Samsung Electronics Co., Ltd. All rights reserved.
+ * Copyright (C) 2019 Dongju Chae <dongju.chae@samsung.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ */
+/**
+ * @file	  gstamcsrc.c
+ * @date	  19 May 2019
+ * @brief	  GStreamer source element for Android MediaCodec (AMC)
+ * @see		  http://github.com/nnsuite/nnstreamer
+ * @author	Dongju Chae <dongju.chae@samsung.com>
+ * @bug		  No known bugs except for NYI items
+ */
+
+/**
+ * SECTION:element-amcsrc
+ *
+ * #amcsrc extends #gstpushsrc source element to reuse the preprocessing capability of 
+ * Android's standard MMFW (i.e., StageFright). It feeds the decoded frames from 
+ * Android MediaCodec (AMC) into a gstreamer pipeline.
+ * 
+ * Note that it's recommended to use this element within Android JNI applications.
+ *
+ * <refsect2>
+ * <title>Example launch line</title>
+ * |[
+ * gst-launch amcsrc location=test.mp4 | autovideosink
+ * ]|
+ * </refsect2>
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "gstamcsrc.h"
+#include "gstamcsrc_looper.h"
+
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <limits.h>
+#include <string.h>
+#include <endian.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <poll.h>
+#include <assert.h>
+/** Glib related */
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <glib/gprintf.h>
+/** Gstreamer related */
+#include <gst/base/base.h>
+#include <gst/video/video.h>
+/** Android/JNI related */
+#include <jni.h>
+#include <android/log.h>
+#include <android/native_window_jni.h>
+#include <media/NdkMediaError.h>
+#include <media/NdkMediaCodec.h>
+#include <media/NdkMediaExtractor.h>
+
+#define TAG "AMCSRC"
+#define LOGI(...) ((void)__android_log_print(ANDROID_LOG_INFO, TAG, __VA_ARGS__))
+#define LOGW(...) ((void)__android_log_print(ANDROID_LOG_WARN, TAG, __VA_ARGS__))
+#define LOGE(...) ((void)__android_log_print(ANDROID_LOG_ERROR, TAG, __VA_ARGS__))
+
+#ifndef S_ISREG
+#define S_ISREG(mode) ((mode)&_S_IFREG)
+#endif
+#ifndef S_ISDIR
+#define S_ISDIR(mode) ((mode)&_S_IFDIR)
+#endif
+#ifndef S_ISSOCK
+#define S_ISSOCK(x) (0)
+#endif
+#ifndef O_BINARY
+#define O_BINARY (0)
+#endif
+
+/**
+ * @brief Template for src pad.
+ */
+static GstStaticPadTemplate src_factory = GST_STATIC_PAD_TEMPLATE ("src",
+    GST_PAD_SRC,
+    GST_PAD_ALWAYS,
+    GST_STATIC_CAPS_ANY);
+
+/**
+ * @brief Private members in GstAMCSrc
+ */
+struct _GstAMCSrcPrivate
+{
+  GMutex mutex;
+  GstDataQueue *outbound_queue; 
+
+  /** media file info */
+  gchar *filename;
+  gint fd;
+  guint64 pos;
+  guint64 size;
+
+  /** media format */
+  gint32 width;
+  gint32 height;
+  gint32 framerate;
+  gint64 duration;
+
+  gboolean seekable;
+  gboolean is_regular;
+  gboolean started;
+  GstClockTime previous_ts;
+
+  gint64 renderstart;
+  gboolean renderonce;
+  gboolean isPlaying;
+  gboolean sawInputEOS;
+  gboolean sawOutputEOS;
+
+  /** Android MediaCodec */
+  AMediaExtractor *ex;
+  AMediaCodec *codec;
+  void *looper;
+};
+
+#define GST_AMC_SRC_GET_PRIVATE(obj)  \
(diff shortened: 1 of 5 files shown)


ext/nnstreamer/android_source/gstamcsrc_looper.h
```diff
@@ -0,0 +1,83 @@
...
+ *
+ */
+/**
+ * @file	  gstamcsrc_looper.h
+ * @date	  19 May 2019
+ * @brief	  GStreamer source element for Android MediaCodec (AMC)
+ * @see		  http://github.com/nnsuite/nnstreamer
+ * @author	Dongju Chae <dongju.chae@samsung.com>
+ * @bug		  No known bugs except for NYI items
+ */
```
myungjoo: Please add more information about looper: What it is & Why it is needed. (It appears that looper is a "thread" for source-android)

dongju-chae: OK. I'll update this.

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,971 @@
...
+  switch (transition) {
+    case GST_STATE_CHANGE_PLAYING_TO_PAUSED:
+      Looper_post (priv->looper, MSG_CODEC_PAUSE, self, false);
+      LOGI ("PLAYING => PAUSED");
+      break;
+    default:
+      break;
+  }
+
+  if (buffer_state_change_success == FALSE)
```
jaeyun-jung: where this flag can be changed to false?

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,971 @@
...
+  GstBuffer *buffer;
+  GstMemory *mem;
+  GstDataQueueItem *item;
+  GstClockTime duration = GST_CLOCK_TIME_NONE;
+  GstClockTime current_ts = GST_CLOCK_TIME_NONE;
+  GstClock *clock;
+
+  GstWrappedBuf *wrapped_buf;
+
+  if ((clock = GST_ELEMENT_CLOCK (self))) {
```
jaeyun-jung: consider to get referenced ```clock = gst_element_get_clock (GST_ELEMENT (self))``` and remove gst_object_ref (clock);

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,971 @@
...
+  GstAMCSrc *self = GST_AMC_SRC (src);
+  GstAMCSrcPrivate *priv = GST_AMC_SRC_GET_PRIVATE (self);
+  GstCaps *caps = gst_caps_new_empty ();
+  GstStructure *format;
+
+  GST_OBJECT_LOCK (self);
+
+  /** width, height, and framerate were obtained from the media format */
+  format = gst_structure_new ("video/x-raw",
+      "format", G_TYPE_STRING, "NV12",  /** TODO Support other formats? */
```
kparichay: You can set the caps in the template, and simply use `gst_pad_get_pad_template_caps (pad)` or `gst_pad_get_current_caps (pad)`

ext/nnstreamer/android_source/gstamcsrc_looper.h
myungjoo: Great!

jni/nnstreamer.mk
```diff
@@ -60,6 +60,10 @@ NNSTREAMER_DECODER_DV_SRCS := \
 NNSTREAMER_DECODER_IL_SRCS := \
     $(NNSTREAMER_EXT_HOME)/tensor_decoder/tensordec-imagelabel.c
 
+# source Android MediaCodec (AMC)
```
leemgs: At this time,  I am not sure that we have to append gstamcsrc*.c in ./jni/nnstreamer.mk. 
Basically. jni/nnstreamer.mk is intended to guide a basic usage for Android JNI interface.

I think that we may add how to write android source code at nnstreamer-example repository. 
* https://github.com/nnsuite/nnstreamer-example/pull/33

ext/nnstreamer/android_source/gstamcsrc.c
dongju-chae: Oh, I'll remove this.

ext/nnstreamer/android_source/gstamcsrc.c
dongju-chae: OK. I'll check this.

jni/nnstreamer.mk
dongju-chae: OK. I'll fix this.

ext/nnstreamer/android_source/gstamcsrc.c
dongju-chae: Hmm. I just wanted to notify available caps (obtained from the media format of an input file) to other elements.
Is it possible to perform the same behavior using your suggestion?

kparichay: Yes, you are right. Cannot set in the template.
You can consider taking intersection of the new caps formed here with gst_pad_get_current_caps (pad), if it is not NULL.

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,976 @@
...
+ * @brief set file location for media codec
+ * @param[in] location Name of the media file
+ */
+static gboolean
+gst_amc_src_set_location (GstAMCSrc * self, const gchar * location,
+    GError ** err)
+{
+  GstAMCSrcPrivate *priv = GST_AMC_SRC_GET_PRIVATE (self);
+
+  if (location == NULL) {
```
jaeyun-jung: let's consider old file name (if se priv->filename is not null (old file name), need to free string.)

jaeyun-jung: and if old file was opened, it needs to close and clear the media info.

dongju-chae: You're right. It will be updated. FYI, a file descriptor is closed in advance when media codec is configured.

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,984 @@
...
+      "format", G_TYPE_STRING, "NV12",  /** TODO Support other formats? */
+      "width", G_TYPE_INT, priv->width,
+      "height", G_TYPE_INT, priv->height,
+      "interlaced", G_TYPE_BOOLEAN, FALSE,
+      "pixel-aspect-ratio", GST_TYPE_FRACTION, 1, 1,
+      "framerate", GST_TYPE_FRACTION, priv->framerate, 1, NULL);
+
+  if (caps) {
+    GstCaps *intersection;
+    intersection = gst_caps_intersect_full (format, caps, GST_CAPS_INTERSECT_FIRST);
```
jaeyun-jung: format is not a GstCaps.

dongju-chae: OMG. It will be fixed.

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,986 @@
...
+  if (current_caps) {
+    GstCaps *intersection =
+      gst_caps_intersect_full (current_caps, caps, GST_CAPS_INTERSECT_FIRST);
+
+    gst_caps_unref (current_caps);
+    gst_caps_unref (caps);
+    caps = intersection;
+  }
+
+  gst_structure_free (format);
```
jaeyun-jung: plz check this line.
gst_caps_append_structure() takes the structure and it does not need to free.

dongju-chae: Right... 'gst_caps_append_structure' does not copy 'format'. I will fix this.

ext/nnstreamer/android_source/gstamcsrc_looper.cc
```diff
@@ -0,0 +1,185 @@
...
+  num_msg = 0;
+
+  pthread_mutex_init(&mutex, NULL);
+  pthread_cond_init(&cond, NULL);
+
+  pthread_attr_init (&attr);
+  pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED); 
+
+  pthread_create (&thread, &attr, entry, this);
+}
```
jaeyun-jung: plz check destroy function (pthread_attr_destroy (&attr)) after creation.

dongju-chae: You're right. It will be updated.

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,987 @@
...
+/**
+ * @brief finalize the instance
+ */
+static void
+gst_amc_src_finalize (GObject * object)
+{
+  GstAMCSrc *self = GST_AMC_SRC (object);
+  GstAMCSrcPrivate *priv = GST_AMC_SRC_GET_PRIVATE (self);
+
+  Looper_post (priv->looper, MSG_CODEC_DONE, self, TRUE);
```
kparichay: Would it be better to call `AMediaCodec_stop()` in `_stop()`
and delete the media codec object with `AMediaCodec_delete()` in `_finalize()`?

kparichay: `_start()` and `_stop()` are called when going upward and downward the states.
As `_start()` creates the looper and starts media codec, I think `looper_exit()` should be done in `_stop()`

ext/nnstreamer/android_source/gstamcsrc.c
```diff
@@ -0,0 +1,987 @@
...
+/**
+ * @brief stop function, called when state changed ready to null
+ */
+static gboolean gst_amc_src_stop (GstBaseSrc * src)
+{
+  GstAMCSrc *self = GST_AMC_SRC (src);
+  GstAMCSrcPrivate *priv = GST_AMC_SRC_GET_PRIVATE (self);
+
+  g_mutex_lock (&priv->mutex);
+
```
kparichay: reset `priv->previous_ts` here? Otherwise duration might be spurious when element restarts, without de-allocation.

dongju-chae: Yeap, you're right. It'll be fixed.

ext/nnstreamer/android_source/gstamcsrc.c
dongju-chae: I see. I think the logic of termination should be changed. thanks!

myungjoo:
Looks good to me.

myungjoo:
@leemgs @kparichay : You need to approve this as you've added quite a few comments before.

PR #1456, opened 2019-05-24
[Conf] Handle the case of using the default 'sysconfdir' value of meson @open sesame 7/25 19:20

wooksong:
Since the default value of 'sysconfdir' in the Meson build system is just 'etc' instead of '/etc', NNSTREAMER_CONF_FILE could be set to wrong path when the default value is used. This patch fixes this issue.

CC: @jaeyun-jung 
See also: https://github.com/nnsuite/nnstreamer/pull/1424

**Changes**
- V4: Add #define to the header file and remove 'const static' in the c file
- V3: Regard the path of root directory for "windows" and "others" as "c:\" and "/" n, respectively
- V2: Use g_path_is_absolute() instead of NNSTREAMER_CONF_FILE[0] == ' /'

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_conf.c
+++ b/gst/nnstreamer/nnstreamer_conf.c
@@ -235,6 +235,7 @@ _fill_in_vstr (gchar *** fullpath_vstr, gchar *** basename_vstr,
 gboolean
 nnsconf_loadconf (gboolean force_reload)
 {
+  const gchar root_path_prefix[] = NNSTREAMER_SYS_ROOT_PATH_PREFIX;
   g_autoptr (GKeyFile) key_file = NULL;
   guint i;
 
@@ -265,7 +266,14 @@ nnsconf_loadconf (gboolean force_reload)
   }
 
   /* Read from the conf file first */
-  conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
+  if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
+    conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
+  } else {
+    /** default value of 'sysconfdir' in meson is 'etc' */
+    conf.conffile = g_build_path (G_DIR_SEPARATOR_S, root_path_prefix,
+        NNSTREAMER_CONF_FILE, NULL);
+  }
+
   if (!g_file_test (conf.conffile, G_FILE_TEST_IS_REGULAR)) {
     /* File not found or not configured */
     g_free (conf.conffile);
--- a/gst/nnstreamer/nnstreamer_conf.h
+++ b/gst/nnstreamer/nnstreamer_conf.h
@@ -36,6 +36,13 @@
 #include <glib.h>
 G_BEGIN_DECLS
 
+/* Hard-coded system-dependent root path prefix */
+#ifdef G_OS_WIN32
+#define NNSTREAMER_SYS_ROOT_PATH_PREFIX "c:\\"
+#else
+#define NNSTREAMER_SYS_ROOT_PATH_PREFIX "/"
+#endif /* G_OS_WIN32 */
+
 /* Env-var names */
 #define NNSTREAMER_ENVVAR_CONF_FILE     "NNSTREAMER_CONF"
 #define NNSTREAMER_ENVVAR_FILTERS       "NNSTREAMER_FILTERS"


gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -264,7 +264,13 @@ nnsconf_loadconf (gboolean force_reload)
   }
 
   /* Read from the conf file first */
-  conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
+  if (NNSTREAMER_CONF_FILE[0] != '/') {
+    /** default value of 'sysconfdir' in meson is 'etc' */
+    conf.conffile = g_strdup_printf("/%s", NNSTREAMER_CONF_FILE);
```
jaeyun-jung: please consider windows

jaeyun-jung: Can we check and change this path in meson script?

wooksong: @jaeyun-jung https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-path-is-absolute

wooksong: @jaeyun-jung IMO, g_path_is_absolute can cover both windows and linux environments.

myungjoo: But, ```"/%s"``` can't.. can it?

myungjoo: And.. how does meson's sysconfdir affect this?

jaeyun-jung: meson default ```sysconfdir``` value now changed, from ```/etc``` to ```etc```.

wooksong: > But, `"/%s"` can't.. can it?
@myungjoo No, as you said, it can't.

> And.. how does meson's sysconfdir affect this? Can you supply an input that creates the issue you mentioned?
@myungjoo 
```bash
$ meson build
The Meson build system
Version: 0.50.999
Source dir: /home/wook/Work/prj_nnstreamer/nnstreamer
Build dir: /home/wook/Work/prj_nnstreamer/nnstreamer/build
Build type: native build
Project name: nnstreamer
Project version: 0.1.3
Native C compiler: ccache cc (gcc 5.4.0 "cc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609")
Native C++ compiler: ccache c++ (gcc 5.4.0 "c++ (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609")
Build machine cpu family: x86_64
Build machine cpu: x86_64
...
Dependency libpng found: YES 1.2.54
Program cp found: YES (/bin/cp)
Dependency GTest found: YES 1.9.0
Configuring nnstreamer-test.ini using configuration
Configuring nnstreamer.ini using configuration
Configuring nnstreamer.pc using configuration
Build targets in project: 32
Found ninja-1.8.2 at /usr/bin/ninja

$ cd build
$ ninja
...
[3/104] ccache cc -Igst/nnstreamer/73d9e45@@nnstreamer@sha -Igst/nnstreamer -I../gst/nnstreamer ... 
'-DNNSTREAMER_CONF_FILE="etc/nnstreamer.ini"' 
...
...
```

wooksong: > meson default `sysconfdir` value changed, from `/etc` to `etc`.
> 
> now nnstreamer sets ini install dir from this.
> nnstreamer_inidir = get_option('sysconfdir')
> 
> how about changing this? (like other options)
> nnstreamer_libdir = join_paths(nnstreamer_prefix, get_option('libdir'))
> nnstreamer_inidir = join_paths(nnstreamer_prefix, get_option('sysconfdir'))
> 
> if user sets --prefix=/usr --sysconfdir=/etc, nnstreamer_inidir=/etc
> if user sets --prefix=/usr --sysconfdir=etc (or use default), nnstreamer_inidir=/usr/etc

@jaeyun-jung 
IMO, if user sets --prefix=/usr --sysconfdir=/etc, nnstreamer_inidir would be /usr//etc rather than /etc.

jaeyun-jung: @wooksong It is /etc. join_paths() ignores old params if having absolute path.
(https://mesonbuild.com/Reference-manual.html#join_paths)

wooksong: @jaeyun-jung Oh, I see. I have submitted another version of this PR. How about the way I used in another PR?

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -34,6 +34,12 @@ static const gchar *subplugin_prefixes[] = {
   [NNSCONF_PATH_END] = NULL
 };
 
+#ifdef G_OS_WIN32
+static const gchar root_path_prefix[] = "c:\\";
```
leemgs: In case of windows,  is "C:\\"  reasonable to all users?

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -265,7 +271,14 @@ nnsconf_loadconf (gboolean force_reload)
   }
 
   /* Read from the conf file first */
-  conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
+  if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
```
leemgs: Looks good.

gst/nnstreamer/nnstreamer_conf.c
myungjoo: Let's not think too much for WINDOWS at this point. We don't use "etc" in Windows either.

myungjoo:
1. You may need to override nnstreamer.ini path with envvar in yocto.
2. If it denies to read envvar because "enable-envvar" is false by default, let's consider to make it "true" by default IF there is NO /etc/nnstreamer.ini. As the rule is getting a little complicated, we will need a related README.md entry in /gst/nnstreamer/ :)

wooksong:
I've just found out what is wrong with this PR. It will be fixed very soon.

gst/nnstreamer/nnstreamer_conf.c
jaeyun-jung: how about moving these block (root_path_prefix) into the loop?
actually root_path_prefix is used only in the loop (line 278), so remove static and move into the loop.

wooksong: @jaeyun-jung I modified my patch as you commented. Thanks.

Issue #1435, opened 2019-05-10
[Tizen/API] Modification request for Tizen ACR process.


again4you:
Tizen API leader requested that below items should be fixed. 

----
- [x] consider changing file name like other module.

- [x] function naming {Module or Sub-module}_[Subject]_{Verb}_[Object]_[Set of parameters]
  * gethandle => get_handle
  * puthandle => put_handle
  * getstate => get_state

- [x] Need to remove unnecessary information in Public header.
  * TBD TODO .. etc

- [x] Consistency issue in the documentation
  * Gstreamer => GStreamer

- [x] Regarding Error
  * If a function has parameter, it should be able to return INVALID_PARAM error
  * If a module is dependent on specific features, it it should be able to return NOT_SUPPORTED error

- [x] enums format
  * nns_tensor_type_e should have NNS_TENSOR_TYPE_Xxxxx

- [x] 3인칭 단순형으로 시작해야 합니다.
  * @brief Control ... => @breif Controls
  * @brief Close ... => @brief Closes

- [x] Not recommend to open struct type. Plz, consider adding a handle for structure.

- [x] param name
  * valvename => valve_name (plz, use _ )

- [x]  for out param (memory management issue)
  * @remarks The @a <array_name> should be released using free().
  * @remarks The @a <object_name> should be released using <function_name>().
  * @remarks The @a <name> should not be released.

- [x] Fix the below comments
  * Are you sure "API" in this case shouldn't be "function"? If this sentence refers to a single function, not a set of functions, it should be "function".

again4you:
Done!

PR #1463, opened 2019-05-28
[Api/Tizen] change function name and refactor code

jaeyun-jung:
1. change function name and indent rule in tizen-capi.
2. unref object after usage (caps and pad data).
3. add macros for log message.


jaeyun-jung:
Related issue #1435

PR #1469, opened 2019-05-30
[Test/Api/Tizen] add testcases for params and element

jaeyun-jung:
Add testcases for function params and element switch


myungjoo:
Anyway... let's try to run SVACE/Coverity daily automatically... I'm getting a bit uneasy again with such issues :)

Issue #1437, opened 2019-05-13
[Filter] Support Intel Movidius Neural Compute Stick


wooksong:
This issue is for tracking the working status of #1138.

- Requirements
  - [x] (mandatory) Add a tensor filter extension for Intel Neural Compute Stick 1 support
  - [x] (mandatory) Support the version 2 of NCSDK (Software Development Kit for the Neural Compute Stick) only
  - [ ] ~~(mandatory) Add several examples based on this tensor filter extension using TensorFlow and Caffe models~~ 
  - [ ] ~~Optional requirements will be added later~~
**Above sub-tasks are postponed and will be handled in the newly opened other issue**

myungjoo:
@wooksong Please close this if you have no pending tasks related.

wooksong:
@myungjoo Got it.

Issue #1138, opened 2019-02-19
[research] Support Intel Movidius USB Stick


myungjoo:
- Do we need to support a tensor_filter_movidius to run models in movidius stick?

If yes, let's start an EPIC item.

ps. At this stage, this is almost a "hobby" project.

PR #1470, opened 2019-05-30
[Filter/MvNCSDK] Write callbacks for framework invocation

wooksong:
This patch fills the body of the callback function for invoke_NN, which
invokes the framework with the given network model.

Relates to #1437
See also: #1441 #1138

**Change logs**
- V3: Fix bug related to g_return_val_if_fail
- V2: Check the number of input tensor in NN_invoke callback

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [*]Skipped
2. Run test: [X]Passed [ ]Failed [* ]Skipped

Signed-off-by: Wook Song wook16.song@samsung.com

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
@@ -281,13 +281,63 @@ err_destroy_device_h:
  * @param[in] input : The array of input tensors
  * @param[out] output : The array of output tensors
  * @return 0 if OK. non-zero if error.
- * @todo : fill this function
  */
 static int
 _mvncsdk2_invoke (const GstTensorFilterProperties * prop, void **private_data,
     const GstTensorMemory * input, GstTensorMemory * output)
 {
+  ncStatus_t ret_code;
+  mvncsdk2_data *pdata = *private_data;
+  guint32 buf_size;
+
+  g_return_val_if_fail (prop->input_configured, -1);
+  if (prop->input_meta.num_tensors != NNS_MVNCSDK2_MAX_NUM_TENOSORS_SUPPORTED) {
+    g_critical ("The number of input tensor should be one: "
+        "The MVNCSDK API supports single tensor input and output only");
+    goto err_destroy;
+  }
+
+  /* Warning: conversion unsigned long to unsigned int */
+  buf_size = (guint32) input->size;
+  ret_code = ncFifoWriteElem(pdata->handle_fifo_input, input->data,
+      &buf_size, 0);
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
```diff
@@ -281,13 +281,56 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
...
 _mvncsdk2_invoke (const GstTensorFilterProperties * prop, void **private_data,
     const GstTensorMemory * input, GstTensorMemory * output)
 {
+  ncStatus_t ret_code;
+  mvncsdk2_data *pdata = *private_data;
+  guint32 buf_size;
+
+  /* Warning: conversion unsigned long to unsigned int */
+  buf_size = (guint32) input->size;
+  ret_code = ncFifoWriteElem(pdata->handle_fifo_input, input->data,
```
jaeyun-jung: params const GstTensorMemory * input and GstTensorMemory * output

Currently input[16] output[16] passed from tensor_filter main.
This is ok here if this handles only one input/output memory block.

myungjoo: If you want to support single-tensor only, you need to generate warning message or error if multi-tensor is given.

wooksong: @myungjoo IMHO, since num_tensors have been set to NNS_MVNCSDK2_MAX_NUM_TENOSORS_SUPPORTED (i.e., 1) at the stages of getInputDimension and getOutputDimension, the input of this function would be always single-tensor. I will add a g_assert here for just in case.

ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
```diff
@@ -281,13 +281,63 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
...
  */
 static int
 _mvncsdk2_invoke (const GstTensorFilterProperties * prop, void **private_data,
     const GstTensorMemory * input, GstTensorMemory * output)
 {
+  ncStatus_t ret_code;
+  mvncsdk2_data *pdata = *private_data;
+  guint32 buf_size;
+
+  g_return_val_if_fail (-1, prop->input_configured);
```
jaeyun-jung: plz check this line ( cond, val )

wooksong: @jaeyun-jung My bad. Fixed.

PR #1476, opened 2019-06-04
[Test] move data files to model directory

jaeyun-jung:
Move files to data directory and remove duplicated.


tests/nnstreamer_decoder_image_labeling/runTest.sh
```diff
@@ -70,6 +67,10 @@ fi
 
 # Decoding 'orange' tests
 # Since data type of output tensor is uint8, int8 requires another 'quantization' (such as /2)
+PATH_TO_MODEL="../test_models/models/mobilenet_v1_1.0_224_quant.tflite"
+PATH_TO_LABEL="../test_models/labels/labels.txt"
+PATH_TO_IMAGE="../test_models/data/orange.png"
```
leemgs: The changed folder of orange.png is better for me.

kparichay:
I think using the filename for the label (ex. extract label `orange` from `orange.png`) is more convenient than specifying the label manually (`orange`).
Otherwise, LGTM.

jaeyun-jung:
@kparichay I considered to set raw-data file name (not a label), and parse file path with delimiter char.
If you think this is necessary to get the answer from filename, I will change arg index to compare the answer.

kparichay:
@jaeyun-jung its not necessary. Both methods work.

kparichay:
LGTM.

PR #1477, opened 2019-06-05
[Dist] typo fix $ --> % @open sesame 6/6 22:00

myungjoo:
% is the macro character in RPM/spec. $ is a typo.

---

**Self evaluation:**


packaging/nnstreamer.spec
```diff
@@ -179,7 +179,7 @@ enable_tf=true
 enable_tf=false
```
kparichay:
```suggestion
%define enable_tf false
```
and same for `enable_tf=true`

myungjoo: Thanks!

PR #1479, opened 2019-06-05
[Filter/Caffe2] initial commit for caffe2

helloahn:
it will be updated after packaging pytorch properly and example & test also

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
@@ -1,0 +1,194 @@
+/**
+ * GStreamer Tensor_Filter, caffe2 Module
+ * Copyright (C) 2019 Hyoung Joo Ahn <hello.ahn@samsung.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ */
+/**
+ * @file	tensor_filter_caffe2.c
+ * @date	27 May 2019
+ * @brief	Caffe2 for tensor_filter gstreamer plugin
+ * @see		http://github.com/nnsuite/nnstreamer
+ * @author	HyoungJoo Ahn <hello.ahn@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ * This is the per-NN-framework plugin (caffe2) for tensor_filter.
+ * Fill in "GstTensorFilterFramework" for tensor_filter.h/c
+ *
+ */
+
+#include <glib.h>
+#include <string.h>
+
+#include "tensor_filter_caffe2_core.h"
+
+void init_filter_caffe2 (void) __attribute__ ((constructor));
+void fini_filter_caffe2 (void) __attribute__ ((destructor));
+
+/**
+ * @brief internal data of caffe2
+ */
+struct _Caffe2_data
+{
+  void *caffe2_private_data;
+};
+typedef struct _Caffe2_data caffe2_data;
+
+
+/**
+ * @brief Free privateData and move on.
+ */
+static void
+caffe2_close (const GstTensorFilterProperties * prop, void **private_data)
+{
+  caffe2_data *cf2;
+  cf2 = *private_data;
+  caffe2_core_delete (cf2->caffe2_private_data);
+  g_free (cf2);
+  *private_data = NULL;
+}
+
+/**
+ * @brief Load caffe2 modelfile
+ * @param prop property of tensor_filter instance
+ * @param private_data : caffe2 plugin's private data
+ * @return 0 if successfully loaded. 1 if skipped (already loaded).
+ *        -1 if the object construction is failed.
+ *        -2 if the object initialization if failed
+ */
+static int
+caffe2_loadModelFile (const GstTensorFilterProperties * prop,
+    void **private_data)
+{
+  caffe2_data *cf2;
+  if (*private_data != NULL) {
+    /** @todo : Check the integrity of filter->data and filter->model_file, nnfw */
+    cf2 = *private_data;
+    if (g_strcmp0 (prop->model_file,
+            caffe2_core_getModelPath (cf2->caffe2_private_data)) != 0 ||
+        g_strcmp0 (prop->model_file_sub,
+            caffe2_core_getInitModelPath (cf2->caffe2_private_data)) != 0) {
+      caffe2_close (prop, private_data);
+    } else {
+      return 1;
+    }
+  }
+  cf2 = g_new0 (caffe2_data, 1); /** initialize cf2 Fill Zero! */
+  *private_data = cf2;
+  cf2->caffe2_private_data = caffe2_core_new (prop->model_file,
+      prop->model_file_sub);
+  if (cf2->caffe2_private_data) {
+    if (caffe2_core_init (cf2->caffe2_private_data)) {
+      g_printerr ("failed to initialize the object: Caffe2");
+      g_free (cf2);
+      return -2;
+    }
+    return 0;
+  } else {
+    g_printerr ("failed to create the object: Caffe2");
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -0,0 +1,465 @@
...
+int
+Caffe2Core::initInputTensor ()
+{
+  int i = 0;
+  inputTensorMap.clear ();
+  for (; i < inputTensorMeta.num_tensors; i++) {
+    Tensor *inputTensor = workSpace.CreateBlob (inputTensorMeta.info[i].name)
+      ->GetMutable<Tensor> ();
+
+  switch (inputTensorMeta.info[i].type){
```
jaeyun-jung: add space in the loop

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.h
```diff
@@ -0,0 +1,95 @@
...
+#ifndef TENSOR_FILTER_CAFFE2_CORE_H
+#define TENSOR_FILTER_CAFFE2_CORE_H
+
+#include "nnstreamer_plugin_api_filter.h"
+
+#ifdef __cplusplus
+#include <iostream>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wredundant-decls"
```
jaeyun-jung: Question: ignored "-Wredundant-decls",  is it necessary to avid build warning in caffe2?

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -0,0 +1,465 @@
...
+#endif
+  return 0;
+}
+
+/**
+ * @brief extract and store the information of input tensors
+ * @return 0 if OK. non-zero if error.
+ */
+int
+Caffe2Core::setInputTensorProp (const GstTensorFilterProperties * prop)
```
jaeyun-jung: IMO setInputTensorProp() and setOutputTensorProp() these functions are unnecessary. 
plz consider to copy the meta data in init function.

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
```diff
@@ -0,0 +1,192 @@
...
+ */
+static int
+caffe2_loadModelFile (const GstTensorFilterProperties * prop,
+    void **private_data)
+{
+  caffe2_data *cf2;
+  if (*private_data != NULL) {
+    /** @todo : Check the integrity of filter->data and filter->model_file, nnfw */
+    cf2 = *private_data;
+    if (strcmp (prop->model_file,
```
kparichay: use `g_strcmp0()`?

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -0,0 +1,465 @@
...
+  }
+
+  workSpace.RunNet (predictNet.name ());
+
+  for (i = 0; i < outputTensorMeta.num_tensors; i++) {
+    const auto& out = workSpace.GetBlob (outputTensorMeta.info[i].name)
+      ->Get<Tensor> ();
+    switch (outputTensorMeta.info[i].type){
+      case _NNS_INT32:
+        output[i].data = out.data<int32_t>();
```
kparichay: If my understanding is correct, as `allocate_in_invoke` is set to `FALSE`, `GstTensorMemory * output` already has memory allocated, overwriting `output[i].data` will result in loss of that memory.
I think `allocate_in_invoke` should be set to `TRUE`

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -0,0 +1,465 @@
...
+do {\
+  ReinitializeTensor (\
+      inputTensor,\
+      {\
+        inputTensorMeta.info[i].dimension[0],\
+        inputTensorMeta.info[i].dimension[1],\
+        inputTensorMeta.info[i].dimension[2],\
+        inputTensorMeta.info[i].dimension[3]\
+      },\
+      at::dtype<type> ().device (CPU)\
```
kparichay: can we infer the device from the model file to be loaded rather than using fixed `CPU`?
Also, what is the default device (if we donot specify any device)?

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
myungjoo: yup. strcmp might be not that portable especially when a variable may be NULL.

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
helloahn: it's added

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.h
helloahn: it's removed

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
helloahn: it's removed

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
helloahn: it's applied

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
helloahn: it's changed.

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
helloahn: currently, the default device for every nnfw of nnstreamer is CPU,  because there is no package for gpu, include tf, tf-lite. When the gpu packages are updated, we will make another option for it or respective fw, like tf-gpu, caffe2-gpu.

kparichay: ok.

ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -0,0 +1,439 @@
...
+  );\
+} while (0);
```
kparichay: I dont understand - how does the code flow come out of this loop?

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.c
```diff
@@ -0,0 +1,192 @@
...
+    if (g_strcmp0 (prop->model_file,
+            caffe2_core_getModelPath (cf2->caffe2_private_data)) != 0 ||
+        g_strcmp0 (prop->model_file_sub,
+            caffe2_core_getInitModelPath (cf2->caffe2_private_data)) != 0) {
+      caffe2_close (prop, private_data);
+    } else {
+      return 1;
+    }
+  }
+  cf2 = g_new0 (caffe2_data, 1); /** initialize cf2 Fill Zero! */
```
wooksong: Is cf2 freed after returning error?

helloahn: `return 1` doesn't mean an error. It means maintaining current state rather than closing&opening new caffe2 with the new models.

wooksong: @helloahn Do you mean that cf2 would be freed after all in the cases of returning -2 and -1?

helloahn: @wooksong 
I did some tests for that case, and you're right. It's not freed if there is any problem. Specifically, `.close()` is not called.
Furthermore, our nnfws(like tensorflow, tf-lite, pytorch..) have the same problem. I'll fix it at once at another pr.

wooksong: @helloahn 
> Furthermore, our nnfws(like tensorflow, tf-lite, pytorch..) have the same problem.
Right. It is need to be fixed later.

IMHO, in this case, simply add g_free just before ```return -1``` and ```return -2``` would be fixed. Anyway, not a critical issue at this point.

PR #1484, opened 2019-06-12
[Tizen/Api] support tensorflow model in singleshot

jaeyun-jung:
1. add tensorflow pipeline in singleshot
2. testcase with tensorflow speech cmd model

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -149,6 +149,54 @@ ml_single_open (ml_single_h * single, const char *model_path,
           ("appsrc name=srcx ! tensor_filter name=filterx framework=tensorflow-lite model=%s ! appsink name=sinkx sync=false",
           model_path);
       break;
+    case ML_NNFW_TENSORFLOW:
+      if (!g_str_has_suffix (model_path, ".pb")) {
+        ml_loge ("The given model file [%s] has invalid extension.", model_path);
+        return ML_ERROR_INVALID_PARAMETER;
+      }
+
+      if (input_info && output_info) {
+        GstTensorsInfo in_info, out_info;
+        gchar *str_dim, *str_type, *str_name;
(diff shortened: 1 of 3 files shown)


tizen-api/src/tizen-api-util.c
```diff
@@ -0,0 +1,442 @@
...
+}
+
+/**
+ * @brief Checks the availability of the given execution environments.
+ */
+int
+ml_util_check_nnfw (ml_nnfw_e nnfw, ml_nnfw_hw_e hw)
+{
+  /** @todo fill this function */
+#ifndef ENABLE_TENSORFLOW_LITE
```
myungjoo: You cannot determine this at build time.
If someone installs tensorflow-lite subplugin (built separatedly with /ext/nnstreamer/...), you get this supported.
If someone deletes the subplugin, you cannot support it.

You may want to use ```get_subplugin``` API (nnstreamer_subplugin.h).

jaeyun-jung: Updated to check tensorflow using nnstreamer_filter_find()

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -149,6 +149,54 @@ ml_single_open (ml_single_h * single, const char *model_path,
...
+        str_type = gst_tensors_info_get_types_string (&out_info);
+        str_name = gst_tensors_info_get_names_string (&out_info);
+        out_option = g_strdup_printf ("output=%s outputtype=%s outputname=%s",
+            str_dim, str_type, str_name);
+        g_free (str_dim);
+        g_free (str_type);
+        g_free (str_name);
+
+        pipeline_desc =
+            g_strdup_printf
```
kparichay: LGTM.

myungjoo: you need to click "approve". GitHub doesn't recognize your "LGTM".

PR #1490, opened 2019-06-13
[Tizen/Api] Remove Tizen dependency from CAPI set

again4you:
In order to support various platform APIs include Tizen, this PR removes Tizen dependency from CAPI set and reorganizes directory structure. Detailed works are as below.

* Support multiple Log environment such as Android, Tizen and Linux distro instead of dlog API.
* Use the standard error code instead of Tizen's Error code. 
* Reorganizes directory structure to support multiple platform APIs

### Related Issue
* https://github.com/nnsuite/nnstreamer/issues/1489

### Self evaluation:


api/capi/include/nnstreamer.h
```diff
@@ -130,11 +130,13 @@ typedef enum _ml_tensor_type_e
...
-  ML_ERROR_NOT_SUPPORTED			= TIZEN_ERROR_NOT_SUPPORTED, /**< The feature is not supported */
-  ML_ERROR_STREAMS_PIPE			= TIZEN_ERROR_STREAMS_PIPE, /**< Cannot create or access GStreamer pipeline. */
-  ML_ERROR_TRY_AGAIN				= TIZEN_ERROR_TRY_AGAIN, /**< The pipeline is not ready, yet (not negotiated, yet) */
+  ML_ERROR_NONE				= 0, /**< Success! */
+  ML_ERROR_INVALID_PARAMETER			= -EINVAL, /**< Invalid parameter */
+  ML_ERROR_STREAMS_PIPE			= -ESTRPIPE, /**< Cannot create or access GStreamer pipeline. */
+  ML_ERROR_TRY_AGAIN				= -EAGAIN, /**< The pipeline is not ready, yet (not negotiated, yet) */
+  ML_ERROR_UNKNOWN = (-1073741824LL),
+  ML_ERROR_TIMED_OUT,
+  ML_ERROR_NOT_SUPPORTED, /**< The feature is not supported */
```
myungjoo: To comply with Tizen, you may need to keep using ```TIZEN_ERROR_...``` with ```#if defined(__TIZEN__)``` for Tizen. For ```#else``` case, the current fix is great.

api/meson.build
```diff
@@ -0,0 +1,4 @@
+# Build Tizen CAPI
```
myungjoo: ```s/Tizen //```

myungjoo:
Looks mostly good except those few items.

api/capi/include/nnstreamer.h
again4you: OK. I will.

api/meson.build
again4you: ACK!

debian/rules
```diff
@@ -35,7 +35,7 @@ override_dh_auto_clean:
 
 override_dh_auto_configure:
 	mkdir -p build
-	meson --buildtype=plain --prefix=/usr --sysconfdir=/etc --libdir=lib/$(DEB_HOST_MULTIARCH) --bindir=lib/nnstreamer/bin --includedir=include -Dinstall-example=true -Denable-tensorflow=$(enable_tf) build
+	meson --buildtype=plain --prefix=/usr --sysconfdir=/etc --libdir=lib/$(DEB_HOST_MULTIARCH) --bindir=lib/nnstreamer/bin --includedir=include -Dinstall-example=true -Denable-tensorflow=$(enable_tf) -Denable-capi=true build
```
jaeyun-jung: plz consider to set default true in meson_option and revert this change.

again4you:
### Submit v2

* Remove unnecessary `tizen` comment in meson build script.
* Fix the build break error when building debian package
* Use TIZEN_ERROR when API is used for Tizen.

again4you:
### Submit v3
* Tizen related Macro would be updated when CAPI is built for Non-Tizen platform.
* Update omitted `return value` in header files.

PR #1492, opened 2019-06-14
[Conf] If CONFFILE path is not available, use the default @open sesame 6/25 17:25

myungjoo:
If conffile path is not available in envvar, use the default.
We had this feature before and... somehow, this feature
has been removed at some point.
I'm reviving this feature as this makes errors in some systems.

I've added g_assert to avoid similar errors without
detected at the point of failure.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_conf.c
+++ b/gst/nnstreamer/nnstreamer_conf.c
@@ -274,6 +274,12 @@ nnsconf_loadconf (gboolean force_reload)
     conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
   }
 
+  if (conf.conffile == NULL)
+    conf.conffile = g_strdup (NNSTREAMER_DEFAULT_CONF_FILE);
+
+  g_assert (key_file != NULL);
+  g_assert (conf.conffile != NULL);
+
   /* Read the conf file. It's ok even if we cannot load it. */
   if (g_key_file_load_from_file (key_file, conf.conffile, G_KEY_FILE_NONE,
           NULL)) {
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -468,6 +474,9 @@ nnsconf_get_custom_value_string (const gchar * group, const gchar * key)
     if (NULL == value && conf.conffile) {
       g_autoptr (GKeyFile) key_file = g_key_file_new ();
 
+      g_assert (key_file != NULL);
+      g_assert (conf.conffile != NULL);
```
kparichay: A `NULL` check for `conf.conffile` is already performed at 474. I think we can remove one.

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -274,6 +274,12 @@ nnsconf_loadconf (gboolean force_reload)
     conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
   }
 
+  if (conf.conffile == NULL)
+    conf.conffile = g_strdup (NNSTREAMER_DEFAULT_CONF_FILE);
+
+  g_assert (key_file != NULL);
+  g_assert (conf.conffile != NULL);
```
kparichay: IMO, I think this check is not needed.
if `g_strdup()` fails, then as per documentation, `If any call to allocate memory fails, the application is terminated. This also means that there is no need to check if the call succeeded.`

myungjoo: This makes sure that ```NNSTREAMER_DEFAULT_CONF_FILE``` is not NULL

gst/nnstreamer/nnstreamer_conf.c
myungjoo: Will update this.

gst/nnstreamer/nnstreamer_conf.c
kparichay: Right.

PR #1496, opened 2019-06-14
[C-API] appsink function in old version @open sesame 6/19 13:33

jaeyun-jung:
gst_app_sink_try_pull_sample is available at ver 1.10.
check minor version and get buffer from appsink with gst_app_sink_pull_sample.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -333,8 +333,14 @@ ml_single_inference (ml_single_h single,
   }
 
   /* Try to get the result */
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -333,8 +333,14 @@ ml_single_inference (ml_single_h single,
   }
 
   /* Try to get the result */
+#if (GST_VERSION_MINOR >= 10)
```
myungjoo: Check MAJOR along with this?
(e.g., say FALSE for all 0.*, TRUE for 2.*)

jaeyun-jung: changed condition 
```
#if (GST_VERSION_MAJOR >= 1 && GST_VERSION_MINOR >= 10)
```

PR #1511, opened 2019-07-09
[C-Api/Android] add build path for c-api

jaeyun-jung:
define path of c-api for android build


myungjoo:
Anyway, did you omit other files that may refer to ```NNSTREAMER_CAPI_*```?

jaeyun-jung:
@myungjoo now I am preparing android-java library, it will use capi in native build.
Currently I tested single-shot open (ml_single_open) with tf-lite.
e.g., Single s = new Single (File model) in java, then nativeOpen(model.path) will use singleshot c functions.

PR #1525, opened 2019-07-12
[Filter/TF] apply a new constructor for TFBuffer - @open sesame 7/22 15:49

helloahn:
The different constructor is applied according to the version of TF
With this pr, GBS build works well without errors related to tensorflow 1.13.1

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_core.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_core.cc
@@ -420,8 +420,14 @@ class TFBuffer : public TensorBuffer {
   size_t len_;
 
 #if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION < 13)
+  explicit TFBuffer (void* data_ptr) : data_(data_ptr) {}
   void* data () const override { return data_; }
+#elif (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 13)
+  explicit TFBuffer (void* data_ptr) : TensorBuffer (data_ptr) {}
+#else
+#error This supports Tensorflow 1.x only.
 #endif
+
   size_t size () const override { return len_; }
   TensorBuffer* root_buffer () override { return this; }
   void FillAllocationDescription (AllocationDescription* proto) const override {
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_core.cc
```diff
@@ -420,8 +420,12 @@ class TFBuffer : public TensorBuffer {
   size_t len_;
 
 #if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION < 13)
+  explicit TFBuffer (void* data_ptr) : data_(data_ptr) {}
   void* data () const override { return data_; }
+#elif (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION == 13)
```
myungjoo: wouldn't it be ok to say ```>=``` instead of ```==``` ?

myungjoo: It'd be better if you handle TF_MAJOR_VERSION != 1
e.g., 
```
#else
#error This supports Tensorflow 1.x only.
```
until we build & test tensorflow 2.x w/ nnstreamer.

helloahn: > wouldn't it be ok to say >= 13 instead of == 13 ?

the change is maintained after 1.13. So I applied `>= 13`.

> It'd be better if you handle TF_MAJOR_VERSION != 1

it's applied

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_core.cc
```diff
@@ -420,8 +420,16 @@ class TFBuffer : public TensorBuffer {
   size_t len_;
 
 #if (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION < 13)
+  explicit TFBuffer (void* data_ptr) : data_(data_ptr) {}
   void* data () const override { return data_; }
+#elif (TF_MAJOR_VERSION == 1 && TF_MINOR_VERSION >= 13)
+  explicit TFBuffer (void* data_ptr) : TensorBuffer (data_ptr) {}
+#else
+  explicit TFBuffer (void* data_ptr) {
+    g_critical ("This supports Tensorflow 1.x only.");
```
myungjoo: USE #error for build error.

PR #1526, opened 2019-07-15
[C-Api] free element handle - @open sesame 7/15 14:42

jaeyun-jung:
Add code to free pipeline element and hash-table.

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -327,6 +327,8 @@ cleanup_node (gpointer data)
 
   g_mutex_unlock (&e->lock);
   g_mutex_clear (&e->lock);
+
+  g_free (e);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -327,6 +327,8 @@ cleanup_node (gpointer data)
 
   g_mutex_unlock (&e->lock);
   g_mutex_clear (&e->lock);
+
+  g_free (e);
```
myungjoo: Isn't this handled by ```g_free``` registered by ```      g_hash_table_new_full (g_str_hash, g_str_equal, g_free, cleanup_node);``` ?

wooksong: @myungjoo Since the third parameter, g_free, is for the key value and the fourth parameter, cleaup_node (which is this function), is for the data, in my humble opinion, it seems that g_free for the data is required here.

jaeyun-jung: As wooksong said, the value ptr needs to be freed in the value-destroy callback.

myungjoo: Ok. thanks for clarification.

PR #1534, opened 2019-07-17
[C-Api] fix error to run pipeline - @open sesame 7/12 19:55

jaeyun-jung:
1. return valid error code when opening single-shot model.
2. When called destroy() in pipeline, remove all callbacks before changing pipeline state.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -151,9 +151,12 @@ ml_single_open (ml_single_h * single, const char *model,
   /* 2. Determine hw */
   /** @todo Now the param hw is ignored. (Supposed CPU only) Support others later. */
   status = ml_check_nnfw_availability (nnfw, hw, &available);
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -153,7 +153,7 @@ ml_single_open (ml_single_h * single, const char *model,
   status = ml_check_nnfw_availability (nnfw, hw, &available);
   if (status != ML_ERROR_NONE || !available) {
     ml_loge ("The given nnfw is not available.");
-    return status;
+    return ML_ERROR_NOT_SUPPORTED;
```
myungjoo: No.

```return ML_ERROR_NOT_SUPPORTED``` only when available == FALSE.
Otherwise, if status != ML_ERROR_NONE, return status.

jaeyun-jung: I updated this line, please re-sync the changes.

PR #1535, opened 2019-07-18
[C-Api] fix memory leak bug

again4you:
Since developers can reduce the number of Tensors by calling
ml_tensors_info_set_count(), it causes the memory leak when invoking
ml_tensors_info_free() and ml_tensors_data_destroy(). This patch fixes
that bug by checking all possible elements (i.e. ML_TENSOR_SIZE_LIMIT).

### Self evaluation:

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -397,7 +397,7 @@ ml_tensors_info_free (ml_tensors_info_s * info)
   if (!info)
     return;
 
-  for (i = 0; i < info->num_tensors; i++) {
+  for (i = 0; i < ML_TENSOR_SIZE_LIMIT; i++) {
     if (info->info[i].name) {
       g_free (info->info[i].name);
       info->info[i].name = NULL;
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -397,7 +397,7 @@ ml_tensors_info_free (ml_tensors_info_s * info)
   if (!info)
     return;
 
-  for (i = 0; i < info->num_tensors; i++) {
+  for (i = 0; i < ML_TENSOR_SIZE_LIMIT; i++) {
```
myungjoo: At Line 458:
```
  _data->num_tensors = tensors_info->num_tensors;
  for (i = 0; i < _data->num_tensors; i++) {
    _data->tensors[i].size = ml_tensor_info_get_size (&tensors_info->info[i]);
    _data->tensors[i].tensor = g_malloc0 (_data->tensors[i].size);
    if (_data->tensors[i].tensor == NULL)
      goto failed;
  }
```
Is this really ok?

again4you: It's OK. Basically ml_tensors_info_set_count() just updates its `count` and do not freed its allocated memory. So `ml_tensors_info_free()` and `ml_tensors_data_destroy()` check all possible space (i.e. ML_TENSOR_SIZE_LIMIT).

For example, below code currently makes a memory leak.
```cpp
// Set # of Tnesor as 5
ml_tensors_info_set_count(info, 5);
...
ml_tensors_info_set_tensor_name(info, 4, "test");
...
// Re-Set # of Tnesor as 2
ml_tensors_info_set_count(info, 2);
```

myungjoo: I was simply worried about apply g_free() to an unallocated pointer.
However, with
```
  _data = g_new0 (ml_tensors_data_s, 1);
```
allocation, it is filled with zero. so, this is safe.

PR #1540, opened 2019-07-19
[Decoder/Sub] build static lib

jaeyun-jung:
change font file in decoder-subplugins to build static library.


myungjoo:
Why do you need it for static lib?

Can we simply add:
```
-uint8_t rasters[][13] = {
+static uint8_t rasters[][13] = {
```
?

jaeyun-jung:
@myungjoo 
It is necessary to add decoder-subplugins in android ilbrary (current this is static)

PR #1542, opened 2019-07-19
[C-API] check gstreamer initialization status in ml_single_open()

again4you:
Before calling gstreamer function, this patch checks its initialization
status by calling gst_init_check() in ml_single_open().

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -57,6 +57,7 @@ ml_single_open (ml_single_h * single, const char *model,
   ml_pipeline *pipe_h;
   GstElement *appsrc, *appsink, *filter;
   GstCaps *caps;
+  GError *err = NULL;
   int status = ML_ERROR_NONE;
   gchar *pipeline_desc = NULL;
   gchar *path_down;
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -73,6 +74,16 @@ ml_single_open (ml_single_h * single, const char *model,
   /* init null */
   *single = NULL;
 
+  if (FALSE == gst_init_check (NULL, NULL, &err)) {
+    if (err) {
+      ml_loge ("GStreamer has the following error: %s", err->message);
+      g_error_free (err);
+    } else {
+      ml_loge ("Cannot initialize GStreamer. Unknown reason.");
+      return ML_ERROR_STREAMS_PIPE;
```
jaeyun-jung: move this to line 85.

if (err) cannot return error-code.

again4you: Right! thanks

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -73,6 +74,16 @@ ml_single_open (ml_single_h * single, const char *model,
   /* init null */
   *single = NULL;
 
+  if (FALSE == gst_init_check (NULL, NULL, &err)) {
```
myungjoo: Try ```gboolean gst_is_initialized ()``` before calling ```gst_init_check```.

If it is already initialized, don't initialize again.


We actually need to add ``` gst_is_initialized()``` in pipeline API implementation as well.

again4you: `gst_init_check()` function internally checks whether initialization is done or not. So in my opinion, we don't need to check it by calling `gst_is_initialized()`. This mechanism is already applied to `Tizen`.

* https://git.tizen.org/cgit/platform/upstream/gstreamer/tree/gst/gst.c#n341
```cpp
gboolean
gst_init_check (int *argc, char **argv[], GError ** err)
{
  ...
  g_mutex_lock (&init_lock);

  if (gst_initialized) {
    GST_DEBUG ("already initialized gst");
    g_mutex_unlock (&init_lock);
    return TRUE;
  }
  ...
```

myungjoo: Oh. Good!

PR #1551, opened 2019-07-25
[Filter] condition to check model

jaeyun-jung:
Prevent error case if filter framework does not need a model file.
Add a flag in nnfw struct and change condition to check model in transform callback.


jaeyun-jung:
This PR is to register user's filter-framework in code level. (e.g., Android Custom)

myungjoo:
As discussed a few minutes ago, please let the other "conventional" tensor_filter subplugins return ERRORS as before. Probably with a new property for tensor_filter subplugin that is default (0) to return ERRORS if model is not specified.

PR #1553, opened 2019-07-26
[C-Api] Check feature before processing API

again4you:
This patch checks the machine_learning.inference feature before
processing API by calling ml_get_feature_enabled() function.

**Self evaluation:**
1. Build test: [* ]Passed [ ]Failed []Skipped
2. Run test: [* ]Passed [ ]Failed []Skipped

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -345,6 +345,8 @@ ml_pipeline_construct (const char *pipeline_description,
 
   ml_pipeline *pipe_h;
 
+  check_feature_state ();
+
   if (pipe == NULL)
     return ML_ERROR_INVALID_PARAMETER;
 
@@ -487,6 +489,8 @@ ml_pipeline_destroy (ml_pipeline_h pipe)
   GstStateChangeReturn scret;
   GstState state;
 
+  check_feature_state ();
+
   if (p == NULL)
     return ML_ERROR_INVALID_PARAMETER;
 
@@ -545,6 +549,8 @@ ml_pipeline_get_state (ml_pipeline_h pipe, ml_pipeline_state_e * state)
   GstState _state;
   GstStateChangeReturn scret;
 
+  check_feature_state ();
+
   if (p == NULL || state == NULL)
     return ML_ERROR_INVALID_PARAMETER;
 
@@ -573,6 +579,8 @@ ml_pipeline_start (ml_pipeline_h pipe)
   ml_pipeline *p = pipe;
   GstStateChangeReturn scret;
 
+  check_feature_state ();
+
   if (p == NULL)
     return ML_ERROR_INVALID_PARAMETER;
 
@@ -594,6 +602,8 @@ ml_pipeline_stop (ml_pipeline_h pipe)
 {
   ml_pipeline *p = pipe;
   GstStateChangeReturn scret;
+
+  check_feature_state ();
 
   if (p == NULL)
     return ML_ERROR_INVALID_PARAMETER;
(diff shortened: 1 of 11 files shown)


again4you:
### Submit v2
* Check feature before processing API.
  * If `unit test` is enabled or Non-Tizen API, checking process is skipped.
* Update documentation for new feature key of Tizen API.

meson_options.txt
```diff
@@ -15,3 +15,4 @@ option('enable-symbolic-link', type: 'boolean', value: true)
 option('enable-capi', type: 'boolean', value: false)
 option('enable-python', type: 'boolean', value: true)
 option('enable-tizen', type: 'boolean', value: false)
+option('enable-unittest', type: 'boolean', value: false)
```
again4you: Use pre-defined `enable-test` instead of `enable-unittest`

api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -1093,6 +1117,9 @@ ml_pipeline_switch_release_handle (ml_pipeline_switch_h h)
 {
   handle_init (switch, swtc, h);
 
+  if (ML_ERROR_NONE != ml_get_feature_enabled())
```
jaeyun-jung: change to macro check_feature_state()

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -777,3 +815,42 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
...
+int ml_get_feature_enabled (void)
+{
+  /** -1: Not checked yet, 0: Not supported, 1: Supported */
+  static int feature_enabled = -1;
+
+/**
+ * In case of Unit Test or Non-Tizen API, then feature checking is not necessary.
+ * So it always returns ML_ERROR_NONE;
+ */
+#if defined(__UNITTEST__)
```
jaeyun-jung: The option '-Denable-unittest=true' defines '__UNITTEST__' and this always makes the return result ML_ERROR_NONE.

api/capi/src/nnstreamer-capi-pipeline.c
again4you: I will!

meson_options.txt
again4you: Fixed

api/capi/src/nnstreamer-capi-util.c
again4you: Not necessary.

again4you:
### Submit v3
* Fix items that @jaeyun-jung  mentioned.

debian/rules
```diff
@@ -35,7 +35,7 @@ override_dh_auto_clean:
 
 override_dh_auto_configure:
 	mkdir -p build
-	meson --buildtype=plain --prefix=/usr --sysconfdir=/etc --libdir=lib/$(DEB_HOST_MULTIARCH) --bindir=lib/nnstreamer/bin --includedir=include -Dinstall-example=true -Denable-tensorflow=$(enable_tf) -Denable-capi=true build
+	meson --buildtype=plain --prefix=/usr --sysconfdir=/etc --libdir=lib/$(DEB_HOST_MULTIARCH) --bindir=lib/nnstreamer/bin --includedir=include -Dinstall-example=true -Denable-tensorflow=$(enable_tf) -Denable-unittest=false -Denable-capi=true -Denable-tizen=false build
```
jaeyun-jung: plz remove option '-Denable-unittest=false'  (it is unnecessary)

IMO, others are ok.

meson.build
```diff
@@ -205,6 +205,9 @@ if get_option('enable-tizen')
   add_project_arguments('-D__TIZEN__=1', language: ['c', 'cpp'])
 endif
 
+if get_option('enable-test')
+  add_project_arguments('-D__UNITTEST__=1', language: ['c', 'cpp'])
```
myungjoo: With the current logic, ```__UNITTEST__``` should be enabled ONLY WHEN the objects are built for gtest, but you are enabling it when the project has unit-test enabled.

You need to modify how gtest binaries are built, not the global build.

PR #1555, opened 2019-07-26
[Android/Api] fix invalid data and add custom-filter

jaeyun-jung:
1. Fix wrong conversion of byte buffer in native.
2. Add methods to allocate buffer with native byte order.
3. Fix invalid condition to close handle.
4. Add custom-filter in java to execute tensors data in the application.


jaeyun-jung:
Now I added new PR,
1. fix exception case when closing pipeline handle.
2. add custom-filter in java.

Please review the changes again.

PR #1558, opened 2019-07-31
[Android/Api] add script to build api library - @open sesame 8/19 16:47

jaeyun-jung:
Add script to build android-api library.
If runs this script, ext-files are downloaded and run gradle build.

TODO add repo for ext-files (or another way to get ext-files)

### DIFF
--- a/api/android/api/build.gradle
+++ b/api/android/api/build.gradle
@@ -20,7 +20,17 @@ android {
                 if (gstRoot == null)
                     throw new GradleException('GSTREAMER_ROOT_ANDROID must be set, or "gstAndroidRoot" must be defined in your gradle.properties in the top level directory of the unpacked universal GStreamer Android binaries')
 
-                arguments "NDK_APPLICATION_MK=jni/Application.mk", "GSTREAMER_JAVA_SRC_DIR=src", "GSTREAMER_ROOT_ANDROID=$gstRoot", "GSTREAMER_ASSETS_DIR=src/assets"
+                def nnsRoot
+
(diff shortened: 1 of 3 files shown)


build-android-lib.sh
```diff
@@ -0,0 +1,68 @@
...
+# Make directory to build NNStreamer library
+mkdir -p build_android_lib
+
+# Copy the files (native and java to build Android library) to build directory
+cp -r ./api/android/* ./build_android_lib
+
+# Get the prebuilt libraries and build-script
+pushd ./build_android_lib
+# TODO change to nnsuite repo later
+git clone https://github.com/jaeyun-jung/nnstreamer-build-ext.git
```
myungjoo: Git clone might be too heavy for what you need. (You don't need the history)

There are options to skip history and extract the files you need only.

jaeyun-jung: changed to ```svn export ...```

build-android-lib.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
myungjoo: LGTM in overall.

However, please move this script into somewhere android-specific.

jaeyun-jung: moved to ```/api/android```

jaeyun-jung:
Now updated to set nnstreamer root directory in gradle.

jaeyun-jung:
Now updated to set nnstreamer root directory. (for the gradle build in aws)
Please review the changes again.

jaeyun-jung:
Updated (08/12)
limit api module (api:assembleRelease)

PR #1563, opened 2019-08-01
[Tizen/Feature] invalid condition to check feature status

jaeyun-jung:
1. Fix invalid condition when mutex-lock of element handle locked.
2. Add new struct to set mutex locked.


jaeyun-jung:
PR updated (added new struct for feature status)

PR #1567, opened 2019-08-02
[Api/Tizen] update APIs for next process

jaeyun-jung:
PR for next ACR

1. get the byte size with data index.
2. add new function to set the timeout to wait for the output.

### DIFF
--- a/api/capi/include/nnstreamer.h
+++ b/api/capi/include/nnstreamer.h
@@ -700,12 +700,18 @@ int ml_tensors_info_set_tensor_dimension (ml_tensors_info_h info, unsigned int i
 int ml_tensors_info_get_tensor_dimension (ml_tensors_info_h info, unsigned int index, ml_tensor_dimension dimension);
 
 /**
- * @brief Gets the byte size of the given tensors type.
+ * @brief Gets the byte size of the given handle of tensors information.
+ * @details If an application needs to get the total byte size of tensors, set the @a index '-1'. Note that the maximum number of tensors is 16 (#ML_TENSOR_SIZE_LIMIT).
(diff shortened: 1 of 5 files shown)


api/capi/include/nnstreamer.h
```diff
@@ -700,12 +700,17 @@ int ml_tensors_info_set_tensor_dimension (ml_tensors_info_h info, unsigned int i
...
+ * @param[in] info The handle of tensors information.
+ * @param[in] index The index of the tensor.
+ * @param[out] data_size The byte size of tensor data.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported.
+ * @retval #ML_ERROR_INVALID_PARAMETER Given parameter is invalid.
  */
-size_t ml_tensors_info_get_size (const ml_tensors_info_h info);
+int ml_tensors_info_get_tensor_size (ml_tensors_info_h info, unsigned int index, size_t *data_size);
```
myungjoo: What about returning total-size if index=MAX_UINT?

jaeyun-jung: will be updated: if the index is -1, return total byte size of tensors.

myungjoo:
PTAL: @again4you @leemgs

again4you:
LGTM!. If we want to apply the Tizen Release, we need to make a `New` ACR.

myungjoo:
@again4you You need to mark "Approve". I'll proceed assuming that you've simply forgot to click that button.

PR #1569, opened 2019-08-02
[Api/Single] invalid output if error occurs

jaeyun-jung:
1. allocate output buffer after invoke is done.
2. allow one buffer in appsink and drop old buffer.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -304,6 +304,10 @@ ml_single_open (ml_single_h * single, const char *model,
   gst_app_sink_set_caps (GST_APP_SINK (appsink), caps);
   gst_caps_unref (caps);
 
+  /* set max buffer in appsink (default unlimited) and drop old buffer */
+  gst_app_sink_set_max_buffers (GST_APP_SINK (appsink), 1);
+  gst_app_sink_set_drop (GST_APP_SINK (appsink), TRUE);
+
   /* 6. Start pipeline */
   status = ml_pipeline_start (pipe);
   if (status != ML_ERROR_NONE) {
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -442,6 +437,16 @@ ml_single_invoke (ml_single_h single,
...
 
+  /* Allocate output buffer */
+  status = ml_tensors_data_create (&single_h->out_info, output);
+  if (status != ML_ERROR_NONE) {
+    ml_loge ("Failed to allocate the memory block.");
+    *output = NULL;
+    return status;
+  }
+
+  result = (ml_tensors_data_s *) (*output);
```
myungjoo: Why was this code block moved?

jaeyun-jung: When this function returns error (ML_ERROR_TIMED_OUT in line 437), the output-handle is allocated. So I moved these block after retrieving the buffer from appsink.

myungjoo: Then you need
```
gst_sample_unref (sample);
```

jaeyun-jung: 1. sample unref
Right, I added gst_sample_unref() if failed to allocate mem block.

2. buffer
gst_app_src_push_buffer() takes ownership of the buffer.

myungjoo: Ok. LTGM now.

jaeyun-jung: add buffer_unref if failed to push buffer to appsrc (if refcount is 0 already, it will do nothing.)

PR #1571, opened 2019-08-05
[Android/Api] method to get data size

jaeyun-jung:
add method to calculate byte size of tensor data.

### DIFF
--- a/api/android/api/src/com/samsung/android/nnstreamer/TensorsInfo.java
+++ b/api/android/api/src/com/samsung/android/nnstreamer/TensorsInfo.java
@@ -165,6 +165,27 @@ public final class TensorsInfo implements AutoCloseable {
     }
 
     /**
+     * Calculates the byte size of tensor data.
+     *
+     * @param index The index of the tensor information in the list
+     *
+     * @return The byte size of tensor
+     *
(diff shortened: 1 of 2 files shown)


api/android/api/src/com/samsung/android/nnstreamer/TensorsInfo.java
```diff
@@ -242,5 +256,39 @@ public void setDimension(@NonNull int[] dimension) {
...
+                case NNStreamer.TENSOR_TYPE_UINT8:
+                    size = 1;
+                    break;
+                case NNStreamer.TENSOR_TYPE_FLOAT64:
+                case NNStreamer.TENSOR_TYPE_INT64:
+                case NNStreamer.TENSOR_TYPE_UINT64:
+                    size = 8;
+                    break;
+                default:
+                    /* unknown type */
```
myungjoo: ```throw```?

jaeyun-jung: In getTensorSize(), added exception case when returned size is 0 or minus value.

myungjoo: OK

PR #1574, opened 2019-08-06
[Spec] Fix the bug of post uninstall script. @open sesame 8/6 16:08

again4you:
Post uninstall script with `-p` option invokes a single program directly
instead of spawning shell script. Because of this reason, below error
occurs when uninstalling package. This patch fixes that bug.

----------------------------------------------
/sbin/ldconfig: relative path `1' used to build cache
warning: %postun(nnstreamer-0.2.1-6.4.i686) scriptlet failed, exit
status 1

**Self evaluation:**
1. Build test: [* ]Passed [ ]Failed []Skipped

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -206,6 +206,15 @@ ninja -C build %{?_smp_mflags}
 %install
 DESTDIR=%{buildroot} ninja -C build %{?_smp_mflags} install
 
+pushd %{buildroot}%{_libdir}
+ln -sf %{gstlibdir}/libnnstreamer.so libnnstreamer.so
+popd
+
+mkdir -p %{buildroot}%{python_sitelib}
+pushd %{buildroot}%{python_sitelib}
+ln -sf %{_prefix}/lib/nnstreamer/filters/nnstreamer_python2.so nnstreamer_python.so
+popd
+
 %if 0%{?testcoverage}
 ##
 # The included directories are:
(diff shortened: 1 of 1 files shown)


packaging/nnstreamer.spec
```diff
@@ -261,10 +261,11 @@ popd
 %post -n nnstreamer-python2
 ln -s %{_prefix}/lib/nnstreamer/filters/nnstreamer_python2.so %{python_sitelib}/nnstreamer_python.so
 
-%postun -p /sbin/ldconfig
+%postun
 pushd %{_libdir}
 rm libnnstreamer.so
 popd
+/sbin/ldconfig
```
myungjoo: Be careful. Uninstalling things with postun is more difficult than you think:
https://ma.ttias.be/on-removing-users-with-postun-in-rpm-spec-files/

For such a symlink, it's be better done at %install
E.g.,
```
%install
...
...
pushd %{buildroot}%{_libdir}
ln -sf %{gstlibdir}/libnnstreamer.so libnnstreamer.so
popd
...
...
```

This applies to other symlinks as well.

again4you: OK. I will.

again4you:
### Submit v2
* Use `%install` section instead of `%post` and `%postun` section.
* Make link files in install phase and include them into packages.

packaging/nnstreamer.spec
```diff
@@ -206,6 +206,15 @@ ninja -C build %{?_smp_mflags}
 %install
 DESTDIR=%{buildroot} ninja -C build %{?_smp_mflags} install
 
+pushd %{buildroot}%{_libdir}
+ln -s %{gstlibdir}/libnnstreamer.so libnnstreamer.so
```
myungjoo: Use -sf instead of -s in build script. Depending on toolchains, the current code may break the build if libnnstreamer.so does not exist at %{gstlibdir} at this point. Note that not-existing at %{gstlibdir} at this point is not abnormal.

packaging/nnstreamer.spec
```diff
@@ -206,6 +206,15 @@ ninja -C build %{?_smp_mflags}
...
+popd
+
+mkdir -p %{buildroot}%{python_sitelib}
+pushd %{buildroot}%{python_sitelib}
+ln -s %{_prefix}/lib/nnstreamer/filters/nnstreamer_python2.so nnstreamer_python.so
```
myungjoo: Same here

again4you: Fixed.

packaging/nnstreamer.spec
again4you: Fixed.

PR #1579, opened 2019-08-07
[Meson] remove missed option

jaeyun-jung:
remove missed option about tensorflow mem-optimize in meson.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -229,7 +229,6 @@ if get_option('enable-test')
 
   nnstreamer_test_conf.set('ENABLE_ENV_VAR', true)
   nnstreamer_test_conf.set('ENABLE_SYMBOLIC_LINK', false)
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -229,7 +229,6 @@ if get_option('enable-test')
 
   nnstreamer_test_conf.set('ENABLE_ENV_VAR', true)
   nnstreamer_test_conf.set('ENABLE_SYMBOLIC_LINK', false)
-  nnstreamer_test_conf.set('TF_MEM_OPTMZ', false)
```
myungjoo: ```tf_loadModelFile()``` of ```tensor_filter_tensorflow.c``` uses this. Let's keep it although its default value is false.

jaeyun-jung: this was already removed (https://github.com/nnsuite/nnstreamer/pull/1577)

myungjoo: Oh.. ok then.

PR #1583, opened 2019-08-08
[Meson] option to disable orc - @open sesame 8/9 13:35

jaeyun-jung:
Add meson-option to disable orc library (default true).
Set -Denable-orc=false if you need to block the orc library.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -108,14 +108,17 @@ protobuf_dep = dependency('protobuf', version: '>= 3.6.1', required: false)
 # Orc
 have_orcc = false
 
-orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: false)
-orcc = find_program('orcc', required: false)
-if orc_dep.found() and orcc.found()
-  have_orcc = true
-  orcc_args = [orcc, '--include', 'glib.h']
-  add_project_arguments('-DHAVE_ORC=1', language: ['c', 'cpp'])
+if get_option('enable-orc')
+  orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: true)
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -108,15 +108,17 @@ protobuf_dep = dependency('protobuf', version: '>= 3.6.1', required: false)
...
+if get_option('enable-orc')
+  orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: false)
+  orcc = find_program('orcc', required: false)
+  if orc_dep.found() and orcc.found()
+    have_orcc = true
+    orcc_args = [orcc, '--include', 'glib.h']
+    add_project_arguments('-DHAVE_ORC=1', language: ['c', 'cpp'])
+  else
+    message('Cannot find orc library')
+    add_project_arguments('-DDISABLE_ORC=1', language: ['c', 'cpp'])
```
myungjoo: Because we not have enable-orc option, this should break the build.

meson.build
```diff
@@ -108,15 +108,17 @@ protobuf_dep = dependency('protobuf', version: '>= 3.6.1', required: false)
...
+  endif
```
myungjoo: Don't you need to add:
```
else
  add_project_arguments('-DDISABLE_ORC=1', language: ['c', 'cpp'])
```
?

meson.build
jaeyun-jung: I updated for this, set required 'true'  and if failed to find orc library, then stop the build.

meson.build
jaeyun-jung: moved to another loop

Issue #1461, opened 2019-05-27
[Meson] Fix build break of meson on launchpad/nnstreamer


wooksong:
This is a sub-issue of https://github.com/nnsuite/nnstreamer/issues/1440.

- If it is simple to create debian package for meson 0.50.1, I will write new my own debian-packaging related files for it.
- Otherwise, 1) Just skip some failed tests or 2) choose other ways to fix the build break easily will be done.

wooksong:
@nnsuite/nnstreamer Build breaks on [our PPA]Check out [our PPA](https://code.launchpad.net/~nnstreamer/+recipe/nnstreamer-daily) have been just fixed. We can close this issue.

Issue #1440, opened 2019-05-13
[CI/CD] Build failed on Launchpad


wooksong:
- Build log
```bash
Test Group nnstreamer_filter_tensorflow Starts.
[FAILED][Critical] 1:gst-launch of case 1
[FAILED][Critical] 1:Golden test comparison
[FAILED][Critical] 2:gst-launch of case 2
[FAILED][Critical] 2:Golden test comparison
==================================================
[FAILED] Test Group nnstreamer_filter_tensorflow has failed cases (4)
```

Our amd64 build of nnstreamer (xenial/bionic) in launchpad has failed since at some point.

again4you:
Architecture `amd64` and `armhf` are failed. 
* Detailed Log: https://launchpadlibrarian.net/423296474/buildlog_ubuntu-xenial-amd64.nnstreamer_0.1.3.0-0~201905111312~ubuntu16.04.1_BUILDING.txt.gz

kparichay:
I am trying to upload pytorch debian to nnstreamer/ppa
Building pytorch for architecture `armhf` and `arm64` failed because building `nnstreamer` failed - detailed logs - https://launchpad.net/~nnstreamer/+archive/ubuntu/ppa/+build/16798360/+files/buildlog_ubuntu-bionic-armhf.nnstreamer_0.1.3.0-0~201905151322~ubuntu18.04.1_BUILDING.txt.gz

The error however is different compared to the error in the log posted by @again4you

kparichay:
The current error in nnstreamer launchpad for xenial is 
'''
The following packages have unmet dependencies:
 sbuild-build-depends-nnstreamer-dummy : Depends: meson (>= 0.40) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
apt-get failed.
Package installation failed
'''

myungjoo:
Maybe... we need to build meson in launchpad/nnstreamer first.
It appears to emit errors in "test" stage; thus, the first workaround will be "disable test" in the meson debian build script.

wooksong:
@nnsuite/nnstreamer Fixed! We can close this issue now.

Issue #1380, opened 2019-04-17
[Upstream Readiness] meson 0.50 compatibility


myungjoo:
With meson 0.50, compiler checks are more strict:

```
ninja: Entering directory `build'
[1/9] Generating copy-bmp2png with a custom command.
[2/9] Compiling C++ object 'tests/59830eb@@unittest_common@exe/common_unittest_common.cpp.o'.
FAILED: tests/59830eb@@unittest_common@exe/common_unittest_common.cpp.o 
c++ -Itests/59830eb@@unittest_common@exe -Itests -I../tests -Igst/nnstreamer -I../gst/nnstreamer -Igst/nnstreamer/tensor_transform -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/local/include/gstreamer-1.0 -I/usr/include/orc-0.4 -I/usr/src/gtest -I/usr/src/gtest/include -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Werror -std=c++11 -g '-DVERSION="0.1.3"' -DHAVE_ORC=1 -DENABLE_TENSORFLOW=1 -DENABLE_TENSORFLOW_LITE=1 -pthread -MD -MQ 'tests/59830eb@@unittest_common@exe/common_unittest_common.cpp.o' -MF 'tests/59830eb@@unittest_common@exe/common_unittest_common.cpp.o.d' -o 'tests/59830eb@@unittest_common@exe/common_unittest_common.cpp.o' -c ../tests/common/unittest_common.cpp
In file included from ../tests/common/unittest_common.cpp:16:0:
/usr/src/gtest/include/gtest/gtest.h: In instantiation of ?쁳esting::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = unsigned int; T2 = int]??
/usr/src/gtest/include/gtest/gtest.h:1421:23:   required from ?쁲tatic testing::AssertionResult testing::internal::EqHelper<lhs_is_null_literal>::Compare(const char*, const char*, const T1&, const T2&) [with T1 = unsigned int; T2 = int; bool lhs_is_null_literal = false]??
../tests/common/unittest_common.cpp:184:3:   required from here
[...]

PR #1587, opened 2019-08-08
[Meson] Apply work-around for gtest-related failure on meson 0.50.1 @open sesame 8/9 14:30

wooksong:
Fixes https://github.com/nnsuite/nnstreamer/issues/1581
Related to #1380, #1440, #1461 

In the case that gtest has been installed via apt (i.e., apt install libgtest-dev), meson provides the -I options indicating the directory that does not exist. Therefore, it is failed to build when '-Wmissing-include-dirs' is set. This patch works around this issue.

#### Build results on Launchpad

https://launchpad.net/~wooksong/+archive/ubuntu/build-test/+sourcepub/10398603/+listing-archive-extra

amd64: fully built
arm64: fully built

armhf: **FAILED**
```bash
The Meson build system
Version: 0.50.1
...omission...

Did not find CMake 'cmake'
Found CMake: NO
Dependency protobuf found: NO (tried pkgconfig and cmake)
Dependency orc-0.4 found: YES 0.4.25
Program orcc found: YES (/usr/bin/orcc)
Dependency tensorflow-lite found: YES 0.1
Dependency pytorch found: YES 1.1.0
Dependency caffe2 found: YES 1.1.0

meson.build:177:4: ERROR: Problem encountered: Cannot find caffe2
```

I guess pytorch could be found using pkg-config by meson but cmake (which is not installed on armhf Ubuntu by default) is required to be found by meson. IMO, this is not problems of meson or launchpad either.

**Self evaluation:**

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Build-Depends: gcc (>=5), ninja-build, meson (>=0.42),
  libgtest-dev, ssat, libpng-dev, libopencv-dev, liborc-0.4-dev,
  python, python-numpy, python3, python3-dev, python3-numpy,
  tensorflow-lite-dev, pytorch,
- tensorflow-dev [amd64 arm64], python2.7-dev, libprotobuf-dev [amd64 arm64]
+ tensorflow-dev [amd64 arm64], python2.7-dev, libprotobuf-dev [amd64 arm64 armhf]
 Standards-Version: 3.9.6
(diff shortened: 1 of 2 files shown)


wooksong:
## V2 
Append another commit, [Debian/armhf] Add build dependency on protobuf

This patch could fix armhf build failure on launchpad.

PR #1588, opened 2019-08-09
[Filter/Py] remove global val to register subplugin

jaeyun-jung:
SAM result, remove global val about filter-framework.


helloahn:
LGTM

Issue #1598, opened 2019-08-14
[Unittest or Build Fail] Launchpad/ARM


myungjoo:
Case 1. 18.04 / armhf (CAPI unit test fail)
- https://launchpadlibrarian.net/437298176/buildlog_ubuntu-bionic-armhf.nnstreamer_0.2.1.0-0~201908141107~ubuntu18.04.1_BUILDING.txt.gz
- (note that 18.08 / arm64 is ok)

Case 2. 19.04 / armhf & arm64 (CAPI unit test fail)
- https://launchpadlibrarian.net/437298204/buildlog_ubuntu-disco-arm64.nnstreamer_0.2.1.0-0~201908141107~ubuntu19.04.1_BUILDING.txt.gz
- https://launchpadlibrarian.net/437298235/buildlog_ubuntu-disco-armhf.nnstreamer_0.2.1.0-0~201908141107~ubuntu19.04.1_BUILDING.txt.gz

- A few fails are from single-invoke timeout
- Another fail is from NOT-ANALYZED-YET.

Case 3. 19.10 / armhf & arm64 (BUILD FAIL / gcc9 becomes more strict!)
- https://launchpadlibrarian.net/437297993/buildlog_ubuntu-eoan-arm64.nnstreamer_0.2.1.0-0~201908141107~ubuntu19.10.1_BUILDING.txt.gz
- https://launchpadlibrarian.net/437297992/buildlog_ubuntu-eoan-armhf.nnstreamer_0.2.1.0-0~201908141107~ubuntu19.10.1_BUILDING.txt.gz
```
[123/145] Compiling C object 'tests/59830eb@@bmp2png@exe/bmp2png.c.o'.
FAILED: tests/59830eb@@bmp2png@exe/bmp2png.c.o 
[...]

PR #1600, opened 2019-08-14
[CAPI/Unittest] 3s timout too short for emul @open sesame 8/14 22:00

myungjoo:
In launchpad emulator, 3s is not enough.

1. We may need to consider having much longer for Ubuntu...
2. In the future, let the user configure the timeout.

This is for a few cases of #1598

CC: @again4you @helloahn @jaeyun-jung


myungjoo:
Ubuntu unittest-capi is disabled. No more needed.

PR #1606, opened 2019-08-22
[Api] code clean

jaeyun-jung:
1. add struct for state change callback
2. check element pointer before free the resources
3. remove unnecessary code to free buffer mem block

### DIFF
--- a/api/capi/include/nnstreamer-capi-private.h
+++ b/api/capi/include/nnstreamer-capi-private.h
@@ -158,6 +158,14 @@ typedef struct _ml_pipeline_element {
 } ml_pipeline_element;
 
 /**
+ * @brief Internal data structure for the pipeline state callback.
+ */
+typedef struct {
+  ml_pipeline_state_cb cb; /**< Callback to notify the change of pipeline state */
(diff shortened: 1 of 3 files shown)


api/capi/include/nnstreamer-capi-private.h
```diff
@@ -157,6 +157,14 @@ typedef struct _ml_pipeline_element {
...
 } ml_pipeline_element;
 
+/**
+ * @brief Internal data structure for the pipeline state callback.
+ */
+typedef struct {
+  ml_pipeline_state_cb cb; /**< Callback to notify the change of pipeline state */
+  void *user_data; /**< The user data passed when calling the state change callback */
+} pipeline_state_cb_s;
+
```
myungjoo: What's the benefit of this refactoring?

jaeyun-jung: Now I am considering another callback for the resource control in Pipeline struct (e.g., tizen mmfw). This is for the readability.

myungjoo: If we may add another callback, then, it's ok.

PR #1615, opened 2019-08-29
[Build] update build script

jaeyun-jung:
Fix duplicated and unnecessary code with the updation of nnstreamer features.
1. remove src-iio in android-jni (currently excluded to register the element, so unnecessary.)
2. remove option enable-nnfw in .spec (default false)
3. duplicated feature definition for tizen

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -222,15 +222,10 @@ if get_option('enable-python')
   endif
 endif
 
-if get_option('enable-tizen')
-  add_project_arguments('-D__TIZEN__=1', language: ['c', 'cpp'])
-endif
-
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -222,16 +222,11 @@ if get_option('enable-python')
   endif
 endif
```
myungjoo: I don't think you may remove ```__TIZEN__```...

jaeyun-jung: This was duplicated with line 24.
```
if get_option('enable-tizen')
  # Pass __TIZEN__ to the compiler
  add_project_arguments('-D__TIZEN__=1', language: ['c', 'cpp'])
  build_platform = 'tizen'
```

myungjoo: Oh.. great!

PR #1618, opened 2019-08-29
[TESTS] add test cases for nnapi property @open sesame 9/3 07:51

jijoongmoon:
# PR Description
Add test cases for reading nnapi property in tensorflow lite example.

**Self evaluation:**
1. Build test:	 [X]Passed [ ]Failed [ ]Skipped
2. Run test:	 [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: jijoong.moon <jijoong.moon@samsung.com>

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
@@ -32,17 +32,6 @@
 #include "tensor_filter_tensorflow_lite_core.h"
 #include "tensor_common.h"
 
-/**
- * @brief nnapi hw type string
- */
-static const char *nnapi_hw_string[] = {
-  [NNAPI_CPU] = "cpu",
-  [NNAPI_GPU] = "gpu",
-  [NNAPI_NPU] = "npu",
-  [NNAPI_UNKNOWN] = "unknown",
-  NULL
-};
-
 void init_filter_tflite (void) __attribute__ ((constructor));
 void fini_filter_tflite (void) __attribute__ ((destructor));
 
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -145,8 +145,8 @@ TFLiteCore::loadModel ()
       }
     }
 #endif
-
-    g_message ("interpreter->UseNNAPI: %d" , use_nnapi);
+    if(use_nnapi)
+      g_message ("interpreter->UseNNAPI( %s : %s )" , use_nnapi?"true":"false", nnapi_hw_string[accel]);
```
jaeyun-jung: use_nnapi?"true":"false" always 'true'

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
```diff
@@ -32,16 +32,6 @@
 #include "tensor_filter_tensorflow_lite_core.h"
 #include "tensor_common.h"
 
-/**
- * @brief nnapi hw type string
- */
-static const char *nnapi_hw_string[] = {
```
jaeyun-jung: how about moving this string list into _lite_core.cc and make function tflite_get_nnapi_hw_type() instead the static function in _lite.c

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
jijoongmoon: right. I will change it.

Issue #1547, opened 2019-07-24
[Tizen/C-API] GST-LESS Tensor_Filter for Single Shot API


myungjoo:
Another tensor_filter.c-like implementation as a backend of single-shot API, which does not require gstreamer, but can accept tensor_filter's subplugins.

This is for (latency) performance enhancement for single-shot API users who do not reuse handles.

This is Tizen 2020-M1 release item.

kparichay:
Tasks for this epic
- [ ] Make filters independent of gst headers
  - [ ] plugin_api headers independent of gst
  - [ ] filters independent of gst
- [ ] Make tensor-filter without gst
- [ ] Make capi-single using the above tensor-filter without gst dependency

myungjoo:
As we have discussed today, this should NOT interfere with Tizen 5.5 M2 releases. Thus, keep the current single-shot API implementation and make "another" library that does not affect the current C-API. We can later merge the two after M2 release.

kparichay:
Yes, thanks for the direction, will follow.

PR #1619, opened 2019-08-29
[filter_single] Added tensor_filter_single prototype

kparichay:
Added prototype for tensor_filter_single
tensor_filter_single is going to form the backend for singleshot api
to minimize the latency. This is made to be independent of GStreamer,
supporting all the existing tensor_filter plugins

Working on #1547 

**Self evaluation:**

### DIFF
--- a/api/capi/include/tensor_filter_single.h
+++ b/api/capi/include/tensor_filter_single.h
@@ -1,0 +1,94 @@
+/**
+ * Copyright (C) 2019 Parichay kapoor <pk.kapoor@samsung.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ */
+/**
+ * @file	tensor_filter_single.h
+ * @date	28 Aug 2019
+ * @brief	Element to use general neural network framework individually without gstreamer pipeline
+ * @see	  http://github.com/nnsuite/nnstreamer
+ * @author	Parichay Kapoor <pk.kapoor@samsung.com>
+ * @author	MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug	  No known bugs except for NYI items
+ */
+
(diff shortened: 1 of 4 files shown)


myungjoo:
1. Why do we need ```gst/nnstreamer/tensor_filter/tensor_filter_single.h```? tensor_filter_single.c should not export its own interface; its interface should be exactly same with the conventional single-shot API.

2. This should not be built as another "tensor filter subplugin", but as an independent shared library or static library.

kparichay:
> 1. Why do we need `gst/nnstreamer/tensor_filter/tensor_filter_single.h`? tensor_filter_single.c should not export its own interface; its interface should be exactly same with the conventional single-shot API.

tensor_filter_single.c implements tensor_filter without gst support. There will be single-shot c-API new implementation which will have same interface as conventional single-shot API

> 2. This should not be built as another "tensor filter subplugin", but as an independent shared library or static library.

Updated. Moved to `api/capi`. An independent static library for `tensor_filter_single`, which is linked with `capi`.

api/capi/include/tensor_filter_single.h
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+#ifndef __G_TENSOR_FILTER_SINGLE_H__
+#define __G_TENSOR_FILTER_SINGLE_H__
+
+#include <stdint.h>
+#include <glib-object.h>
+
+#include "nnstreamer_subplugin.h"
+#include "nnstreamer_plugin_api_filter.h"
```
myungjoo: They are not in your local directory, try to use <nnstreamer/*.h> and drop nnsinc.

myungjoo:
looks good in overall.

api/capi/include/tensor_filter_single.h
kparichay: updated.

kparichay:
V2:
Updated `#include` paths to format `<nnstreamer/....>`
Removed `nnsinc` include_directories from meson.build
Added more vmethods for tensor_filter single class

api/capi/meson.build
```diff
@@ -18,6 +18,21 @@ if meson.project_name() != 'nnstreamer'
 endif
 
 
+inc = include_directories('include')
+nninc = include_directories('../../gst')
+
+# Create dependency for tensor_filter_single for single-shot capi
+tensor_filter_single_lib = static_library ('tensor_filter_single_lib',
```
myungjoo: Any reason to be a static library?

kparichay: Updated to shared library.

kparichay:
V3:
`tensor_filter_single` is a shared_library
Added `gobject-2.0` as a dependency

V2:
Updated #include paths to format <nnstreamer/....>
Removed nnsinc include_directories from meson.build
Added more vmethods for tensor_filter single class

PR #1624, opened 2019-09-02
[filter_single] Common functions between tensor_filter and tensor_filter_single

kparichay:
Simplified common functions between tensor_filter and tensor_filter_single
Added more implementations in tensor_filter_single

Common Functions between `tensor_filter.c` and `tensor_filter_single.c` have been moved to `tensor_filter_common.c`. These functions do not depend on GStreamer. 

Working on #1547 

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/meson.build
+++ b/gst/nnstreamer/tensor_filter/meson.build
@@ -1,5 +1,6 @@
 tensor_filter_sources = [
   'tensor_filter.c',
+  'tensor_filter_common.c',
   'tensor_filter_custom.c'
 ]
 
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -60,6 +60,7 @@
 #include <string.h>
 
 #include "tensor_filter.h"
+#include "tensor_filter_common.h"
 
 /**
  * @brief Macro for debug mode.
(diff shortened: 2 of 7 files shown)


kparichay:
V2:
Added `tensor_filter_common.c` in `jni\nnstreamer.mk`

gst/nnstreamer/tensor_filter/meson.build
```diff
@@ -1,5 +1,6 @@
 tensor_filter_sources = [
   'tensor_filter.c',
+  'tensor_filter_common.c',
```
kparichay: Should this file be moved out from `gst/nnstreamer/tensor_filter/` plugin folder to `gst/nnstreamer/` given its a common source file for `tensor_filter.c` and `tensor_filter_single.c`?

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -791,125 +588,9 @@ gst_tensor_filter_get_property (GObject * object, guint prop_id,
     case PROP_SILENT:
```
jaeyun-jung: Question: why this property remained in tensor-filter?

kparichay: silent requires access to `self` which is different type for `tensor_filter` and `tensor_filter_single`. (other properties do not require `self` but only `prop`)
The only way I could thing of moving this property to common function (`gst_tensor_common_filter_get_property`) would require to make the common function a macro.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -0,0 +1,421 @@
...
+
+
+/**
+ * @brief Parse the string of model
+ * @param[out] prop Struct containing the properties of the object
+ * @param[in] model_files the prediction model paths
+ * @return number of parsed model path
+ */
+guint
+gst_tensors_parse_modelpaths_string (GstTensorFilterProperties * prop,
```
jaeyun-jung: plz consider to change function name (gst_tensor_filter_xxx) later

kparichay: updated to `gst_tensor_filter_parse_modelpaths_string`

PR #1625, opened 2019-09-02
[TESTHUB] Append 'gtest_output' to report unittest output results

dongju-chae:
This PR appends 'gtest_output' into .spec file to generate .xml gtest output files.
These XML files are reported to TESTHUB (for daily x86_64 build).

**Self evaluation:**


dongju-chae:
@myungjoo For ssat, I will use a simple parsing script to generate a XML file using the existing log.
Both gtest and ssat will have the same XML format.

PR #1629, opened 2019-09-03
[single-shot] benchmarking the latency @open sesame 09/09 17:11

kparichay:
Added benchmarking the latency and running times in the test code
This compares old vs new singleshot api - showing reduction in latency and running-time

These values averaged over 10 continuous runs

|  | Open (us)      | Invoke (ms)           | Close (us)  |
| --- |:-------------:|:-------------:|:-----:|
| New (cache warmup) |  658   | 5195 | 206 |
| Old (cache warmup) |  1228 | 5199   | 5245 |
| New (no warmup) | 1653 | 5205  | 201 |
| Old (no warmup) | 7201 | 5225  | 5299  |

These values are just for the first run

| | Open (us)      | Invoke (ms)           | Close (us)  |
| --- |:-------------:|:-------------:|:-----:|
| New  |  12326   | 5231 | 2347 |
| Old |  58772 | 5250   | 52611 |

Working on #1547

Self evaluation:

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cpp
+++ b/tests/tizen_capi/unittest_tizen_capi.cpp
@@ -1433,6 +1433,99 @@ TEST (nnstreamer_capi_singleshot, invoke_02)
   ml_tensors_info_destroy (in_info);
   ml_tensors_info_destroy (out_info);
 }
+
+/**
+ * @brief Test NNStreamer single shot (tensorflow-lite)
+ * @note Measure the loading time and total time for the run
+ */
+TEST (nnstreamer_capi_singleshot, benchmark_time)
+{
+  ml_single_h single;
+  ml_tensors_info_h in_info, out_info;
+  ml_tensors_data_h input, output;
+  ml_tensor_dimension in_dim, out_dim;
+  int status, count;
+  unsigned long open_duration=0, invoke_duration=0, close_duration=0;
+  gint64 start, end;
+
+  const gchar *root_path = g_getenv ("NNSTREAMER_BUILD_ROOT_PATH");
+  gchar *test_model;
+
+  /* supposed to run test in build directory */
+  if (root_path == NULL)
+    root_path = "..";
+
+  test_model = g_build_filename (root_path, "tests", "test_models", "models",
+      "mobilenet_v1_1.0_224_quant.tflite", NULL);
+  ASSERT_TRUE (g_file_test (test_model, G_FILE_TEST_EXISTS));
+
+  ml_tensors_info_create (&in_info);
+  ml_tensors_info_create (&out_info);
+
+  in_dim[0] = 3;
+  in_dim[1] = 224;
+  in_dim[2] = 224;
+  in_dim[3] = 1;
+  ml_tensors_info_set_count (in_info, 1);
+  ml_tensors_info_set_tensor_type (in_info, 0, ML_TENSOR_TYPE_UINT8);
+  ml_tensors_info_set_tensor_dimension (in_info, 0, in_dim);
+
+  out_dim[0] = 1001;
+  out_dim[1] = 1;
+  out_dim[2] = 1;
+  out_dim[3] = 1;
+  ml_tensors_info_set_count (out_info, 1);
+  ml_tensors_info_set_tensor_type (out_info, 0, ML_TENSOR_TYPE_UINT8);
+  ml_tensors_info_set_tensor_dimension (out_info, 0, out_dim);
+
+  input = output = NULL;
+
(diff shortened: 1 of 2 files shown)


kparichay:
Comparison of the maximum memory consumption between the two API implementations
Both the examples below run the same test case. The difference is that the test case executable has been linked with different API shared library. [mprof](https://github.com/pythonprofilers/memory_profiler) has been used to measure the memory consumption with `--interval 0.01 --include-children` as configuration parameters.

Existing C-API implementation
unittest_tizen_capi

Gst-Less C-API implementation
unittest_tizen_capi_single_new

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -7,6 +7,8 @@
  * @bug         No known bugs
  */
 
+#include <chrono>
```
leemgs: I thought that we have generally used C header files although the file extension is `*.CPP`.  :)
Do you wanna use `chrono `of CPP to use clock and time duration?

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -1264,6 +1266,103 @@ TEST (nnstreamer_capi_singleshot, invoke_02)
...
+
+    start = std::chrono::steady_clock::now();
+    status = ml_single_close (single);
+    end = std::chrono::steady_clock::now();
+    close_duration = std::chrono::duration_cast<std::chrono::microseconds>
+      (end - start).count();
+    EXPECT_EQ (status, ML_ERROR_NONE);
+  }
+
+  g_warning ("Time to open single = %f us", (open_duration * 1.0)/count);
```
leemgs: I like this approach.  🥇

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -493,9 +493,10 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
 
 /**
  * @brief Allocates a tensor data frame with the given tensors info. (more info in nnstreamer.h)
+ * @note Memory for data buffer is not allocated.
  */
 int
-ml_tensors_data_create (const ml_tensors_info_h info,
+ml_tensors_data_create_no_alloc (const ml_tensors_info_h info,
```
leemgs: What is the purpose of this line?

api/capi/meson.build
```diff
@@ -27,13 +30,10 @@ capi_devel_main = []
...
 capi_devel_main += join_paths(meson.current_source_dir(), 'include', 'nnstreamer-single.h')
 
-inc = include_directories('include')
-nninc = include_directories('../../gst')
-
 tizen_deps = []
 
 if (get_option('enable-tizen'))
-  message('CAPI is in Tizen mode')
+  message('capi is in tizen mode')
```
leemgs: Does it mean that message do not handle a capital? Why do we have to use lower-case?

api/capi/src/nnstreamer-capi-util.c
kparichay: This adds another tensor creation method, which does not allocate memory for the data inside it.
It just creates the tensor skeleton appropriately, and data tensor is left for the user to be allocated or set to already allocated memory in the current scenario.

api/capi/meson.build
kparichay: This is my mistake, I will revert and clean these changes.

tests/tizen_capi/unittest_tizen_capi.cpp
kparichay: Yes, `chrono` is used measure time duration. 
I used `chrono` because it is the recommended way for C++
However to keep consistency, I will update to `time.h`

myungjoo:
You may add the experimental results at the README.md in capi src directory

kparichay:
> You may add the experimental results at the README.md in capi src directory

I will update this PR to add the results in README.md

tests/tizen_capi/unittest_tizen_capi.cpp
jaeyun-jung: IMO, g_get_real_time() also possible way.

kparichay: updated to use `g_get_real_time()`

kparichay:
V2:
Code Cleaning
Using `g_get_real_time` instead of `chrono`

kparichay:
V3:
Added README.md for `api/capi/src`

V2:
Code Cleaning
Using `g_get_real_time` instead of `chrono`

PR #1630, opened 2019-09-03
[EMERGENCY FIX] [TF-Lite] Add terminating null

myungjoo:
The string list requires a terminating null.


jaeyun-jung:
TODO gst_tensor_transform_operator_string / gst_tensor_transform_stand_string in tensor-transform should be fixed also.

PR #1631, opened 2019-09-03
[Android] update script to remove warning

jaeyun-jung:
Clean up build script and remove warning about duplicates.


api/android/build-android-lib.sh
```diff
@@ -67,22 +67,33 @@ tar xJf ./ext-files/tensorflow-lite-$nnstreamer_tf_lite_ver.tar.xz -C ./api/jni
...
+    result_directory=android_lib
+    today=$(date '+%Y-%m-%d')
 
-if [[ -e $nnstreamer_android_api_lib ]]; then
     echo "Build procedure is done, copy NNStreamer library to $result_directory directory."
-    mkdir -p $result_directory
-    cp $nnstreamer_android_api_lib ./$result_directory/nnstreamer-api-$(date '+%Y-%m-%d').aar
+    mkdir -p ../$result_directory
+    cp $nnstreamer_android_api_lib ../$result_directory/nnstreamer-api-$today.aar
+    cp $nnstreamer_android_api_sample ../$result_directory/sample-release-unsigned-$today.apk
```
myungjoo: uh.. what is this "sample apk"? is it possible to build an Android app in this git? not in nnstreamer-example.git?

jaeyun-jung: I will update this PR to remove warning and code clean.

api/android/api/jni/Android-nnstreamer.mk
```diff
@@ -23,6 +23,9 @@ LOCAL_SRC_FILES := \
     $(NNSTREAMER_DECODER_IL_SRCS) \
     $(NNSTREAMER_DECODER_PE_SRCS)
 
+# Remove duplicates
+LOCAL_SRC_FILES := $(sort $(NNSTREAMER_SRC_FILES))
```
kparichay: should `sort -u` be used for unique list (to remove duplicates)?

jaeyun-jung: sort -u shows warning in ndk build (no option '-u')
https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

kparichay: Yes, I was looking at GNU core-utils definition of make
https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html
You are right.

PR #1632, opened 2019-09-03
[Api/Single] invalid condition to check model

jaeyun-jung:
1. validate model path before checking the file extension.
2. set pre-defined so extension for custom-filter.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -28,6 +28,7 @@
 #include "nnstreamer-single.h"
 #include "nnstreamer-capi-private.h"
 #include "nnstreamer_plugin_api.h"
+#include "nnstreamer_conf.h"
 
 #undef ML_SINGLE_SUPPORT_TIMEOUT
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -116,6 +117,12 @@ ml_single_open (ml_single_h * single, const char *model,
     }
   }
 
+  if (!model || !g_file_test (model, G_FILE_TEST_IS_REGULAR)) {
+    ml_loge ("The given param, model path [%s] is invalid.",
+        GST_STR_NULL (model));
+    return ML_ERROR_INVALID_PARAMETER;
+  }
```
myungjoo: What if ```enable_symlink=TRUE```?

jaeyun-jung: It is ok to run symlink model file. 
'enable_symlink' option is no effect to model file, do we need to handle symlink option for the NN model?

myungjoo: I thought G_FILE_TEST_IS_REGULAR will fail if model is symlink... But according to https://developer.gnome.org/glib/stable/glib-File-Utilities.html, it is ok. So please never mind.

PR #1634, opened 2019-09-04
[Api/Single] clear old buffer

jaeyun-jung:
1. Add mutex-lock for internal process.
2. Clear old buffer of appsink before pushing new buffer to appsrc.

When old buffer received in appsink after timeout, next output of invoke() may be the previous result.
To prevent this, clear old sample in appsink.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -51,8 +51,73 @@ typedef struct
   ml_tensors_info_s in_info;
   ml_tensors_info_s out_info;
 
-  guint timeout; /* milliseconds */
+  GMutex lock; /**< Lock for internal values */
+  gboolean clear_previous_buffer; /**< Previous buffer was timed out, need to clear old buffer */
+  gboolean is_valid; /**< True if the single handle is valid */
+  guint timeout; /**< Timeout in milliseconds */
 } ml_single;
+
+/**
+ * @brief Gets the tensors info from tensor-filter.
+ */
+static void
+ml_single_get_tensors_info_from_filter (GstElement * filter, gboolean is_input,
+    ml_tensors_info_h * info)
+{
+  ml_tensors_info_s *input_info;
+  GstTensorsInfo gst_info;
+  gchar *val;
+  guint rank;
+  gchar *prop_prefix, *prop_name;
+
+  if (is_input)
+    prop_prefix = g_strdup ("input");
+  else
+    prop_prefix = g_strdup ("output");
+
+  /* allocate handle for tensors info */
+  ml_tensors_info_create (info);
+  input_info = (ml_tensors_info_s *) (*info);
+
+  gst_tensors_info_init (&gst_info);
+
+  /* get dimensions */
+  prop_name = g_strdup_printf ("%s", prop_prefix);
+  g_object_get (filter, prop_name, &val, NULL);
+  rank = gst_tensors_info_parse_dimensions_string (&gst_info, val);
+  g_free (val);
+  g_free (prop_name);
+
+  /* set the number of tensors */
+  gst_info.num_tensors = rank;
+
+  /* get types */
+  prop_name = g_strdup_printf ("%s%s", prop_prefix, "type");
+  g_object_get (filter, prop_name, &val, NULL);
+  rank = gst_tensors_info_parse_types_string (&gst_info, val);
+  g_free (val);
+  g_free (prop_name);
+
+  if (gst_info.num_tensors != rank) {
+    ml_logw ("Invalid state, tensor type is mismatched in filter.");
+  }
+
+  /* get names */
+  prop_name = g_strdup_printf ("%s%s", prop_prefix, "name");
+  g_object_get (filter, prop_name, &val, NULL);
+  rank = gst_tensors_info_parse_names_string (&gst_info, val);
+  g_free (val);
+  g_free (prop_name);
+
+  if (gst_info.num_tensors != rank) {
+    ml_logw ("Invalid state, tensor name is mismatched in filter.");
+  }
+
+  ml_tensors_info_copy_from_gst (input_info, &gst_info);
+  gst_tensors_info_free (&gst_info);
+
+  g_free (prop_prefix);
+}
 
 /**
  * @brief Opens an ML model and returns the instance as a handle.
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -418,6 +418,14 @@ ml_single_invoke (ml_single_h single,
     }
   }
 
+#ifdef ML_SINGLE_SUPPORT_TIMEOUT
+  /* Try to clear old buffer in appsink before pushing the buffer */
+  sample =
+      gst_app_sink_try_pull_sample (GST_APP_SINK (single_h->sink), GST_MSECOND * 10);
```
kparichay: How about keeping a check if the previous buffer was TIMED_OUT based on saved variable?
Current implementation adds 10ms overhead even in cases without timeout.

kparichay: Also, give a warning message that 'Next data sample might be overwritten', as this clearing of old buffer does not guarantee clearing of the old buffer.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -352,6 +356,7 @@ ml_single_close (ml_single_h single)
   }
 
   single_h = (ml_single *) single;
+  g_mutex_lock (&single_h->lock);
```
leemgs: Why do we have to append mutex facility? Is this statement is a critical section to care the consistency of single_h?

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -1767,6 +1767,14 @@ TEST (nnstreamer_capi_singleshot, invoke_timeout)
     EXPECT_EQ (status, ML_ERROR_TIMED_OUT);
     EXPECT_TRUE (output == NULL);
 
+    /* sleep 2 seconds (enough time to arrive the buffer in appsink) */
+    g_usleep (2000 * 1000);
```
leemgs: How did you calculate the 2seconds? Is this heuristic value?

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -597,7 +636,10 @@ ml_single_set_timeout (ml_single_h single, unsigned int timeout)
 
   single_h = (ml_single *) single;
 
+  g_mutex_lock (&single_h->lock);
```
leemgs: Please, note that the mutex cost is not inexpensive even though the modern Linux has introduced the FUTEX mechanism.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -51,7 +51,9 @@ typedef struct
   ml_tensors_info_s in_info;
   ml_tensors_info_s out_info;
 
-  guint timeout; /* milliseconds */
+  GMutex lock; /**< Lock for internal values */
+  gboolean timed_out; /**< Previous buffer was timed out */
```
leemgs: It seems that timed_out and timeout are similar too. Could you improve readability for maintenance?

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -1767,6 +1767,14 @@ TEST (nnstreamer_capi_singleshot, invoke_timeout)
...
+
+    /* check the old buffer is dropped */
```
leemgs: Looks good.

api/capi/src/nnstreamer-capi-single.c
jaeyun-jung: Add mutex-lock and handle timed-out case.

api/capi/src/nnstreamer-capi-single.c
jaeyun-jung: right, for the variables in handle to prevent accessing the pipeline in a time.

tests/tizen_capi/unittest_tizen_capi.cpp
jaeyun-jung: right, img classification model (tf-lite) takes less than 300 ms in my local.

api/capi/src/nnstreamer-capi-single.c
jaeyun-jung: will be changed to 'clear_previous_buffer'

tests/tizen_capi/unittest_tizen_capi.cpp
myungjoo: You can remove this ```g_usleep``` by invoking with ```output2``` in the next invoke.

api/capi/src/nnstreamer-capi-single.c
myungjoo: Which variables do you target to protect?
In general, it appears that you are defining "critical section" too widely.

jaeyun-jung: To prevent invoke() in a time (single pipeline input)

jaeyun-jung:
@kparichay Sorry I cannot understand your comment. plz explain the details.
- In Tizen, gst-version is 1.12 and timeout is available.
- Currently ubuntu with gst-1.8 does not support timeout and when called single-invoke(), it waits for the output buffer without the timeout.

tests/tizen_capi/unittest_tizen_capi.cpp
jaeyun-jung: you are right, I will remove this block in next week.

kparichay:
@jaeyun-jung There is some inconsistency when running capi-unittest between tizen and ubuntu. I am investigating it. Meanwhile, this PR can go forward.

api/capi/src/nnstreamer-capi-single.c
myungjoo: Uh.. then, let's add "valid" variable in ```single_h```, set it true at "open" and set it false at "close" before doing unlock(). Otherwise, invoke might enter between Line 381 and 382.

jaeyun-jung: Add new flag 'is_valid' and check the valid state

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -268,8 +271,11 @@ ml_single_open (ml_single_h * single, const char *model,
   single_h->sink = appsink;
   single_h->filter = filter;
   single_h->timeout = SINGLE_DEFAULT_TIMEOUT;
+  single_h->clear_previous_buffer = FALSE;
+  single_h->is_valid = TRUE;
```
myungjoo: This should be set TRUE only when everything is safe; at the last step right before "return OK".

PR #1635, opened 2019-09-04
[Test] Prevent unit-test errors in slow systems

myungjoo:
In slow systems such as QEMU-ARM, a few test cases fail often.

- Add dynamic timeout for pipeline start
- Add mutex for sink callbacks

@leemgs We can now enable unit-test for ARM in CI server.

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cpp
+++ b/tests/tizen_capi/unittest_tizen_capi.cpp
@@ -87,6 +87,18 @@ TEST (nnstreamer_capi_construct_destruct, failure_02_n)
   EXPECT_EQ (status, ML_ERROR_STREAMS_PIPE);
 }
 
+#define wait_for_start(handle, state, status) \
+do {\
+  int counter = 0;\
+  while ((state == ML_PIPELINE_STATE_PAUSED || \
+          state == ML_PIPELINE_STATE_READY) && counter < 20) {\
+    g_usleep (50000);\
+    counter ++;\
+    status = ml_pipeline_get_state (handle, &state);\
+    EXPECT_EQ (status, ML_ERROR_NONE);\
+  }\
+} while (0)\
+
 /**
  * @brief Test NNStreamer pipeline construct & destruct
  */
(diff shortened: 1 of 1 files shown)


tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -87,6 +87,17 @@ TEST (nnstreamer_capi_construct_destruct, failure_02_n)
   EXPECT_EQ (status, ML_ERROR_STREAMS_PIPE);
 }
 
+#define wait_for_start(handle, state, status) \
+do {\
+  int counter = 0;\
+  while (state == ML_PIPELINE_STATE_PAUSED && counter < 20) {\
```
jaeyun-jung: Question: why the pipeline-state is 'PAUSED', not 'PLAYING'? (Now ml_pipeline_construct() makes the pipeline paused.)

myungjoo: The sequence is:
```
1: construct()
2: state == PAUSED
3: play()
4: state == PAUSED
5: some time after play, the pipeline starts asynchronously...
6: state == PLAYING
```

The time required for step 5 varies and it causes QEMU-ARM systems to use more than 200ms sometimes.

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -211,7 +224,8 @@ TEST (nnstreamer_capi_valve, test01)
   EXPECT_NE (state, ML_PIPELINE_STATE_UNKNOWN);
   EXPECT_NE (state, ML_PIPELINE_STATE_NULL);
 
-  g_usleep (150000); /* 150ms. Let a few frames flow. */
+  wait_for_start (handle, state, status);
```
kparichay: `wait_for_start` should be after `g_usleep`? as the state can be `ML_PIPELINE_STATE_READY` at this moment.

kparichay: How about `wait_for_start` loops till `ML_PIPELINE_STATE_READY < state < ML_PIPELINE_STATE_PAUSED`?
This would eliminate the need for extra `g_usleep` at all the places where `wait_for_start` is called.

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -105,9 +116,10 @@ TEST (nnstreamer_capi_playstop, dummy_01)
   EXPECT_NE (state, ML_PIPELINE_STATE_UNKNOWN);
   EXPECT_NE (state, ML_PIPELINE_STATE_NULL);
 
-  g_usleep (150000); /* 50ms is good for general systems, but not enough for emulators to start gst pipeline. Let a few frames flow. */
```
leemgs: @myungjoo , Is This PR to handle QEMU-Based ARM as well as (hig-performance) X86 for runtest? What is the meaning of the "**in slow systems**" that is described in the PR title? NOTE that the system freeze issue of the unit test in Tizen/GBS can not be completely solved with the short sleep time. (In case of Tizen/GBS based on QEMU/ARM), The issue is generated due to slow software IO emulation as well as slow CPU operation of a process.

myungjoo: **slow systems** means any systems that may incur more than 50ms to configure a pipeline.

tests/tizen_capi/unittest_tizen_capi.cpp
myungjoo: Changed applied.

PR #1641, opened 2019-09-05
[Android] unnecessary log while running android-lib

jaeyun-jung:
1. add condition to check file path or tensor name
2. remove unnecessary res file in android

This will remove unnecessary log-print while running java custom-filter and tf-lite model.

### DIFF
--- a/gst/nnstreamer/tensor_common.c
+++ b/gst/nnstreamer/tensor_common.c
@@ -534,7 +534,9 @@ gst_tensors_info_get_names_string (const GstTensorsInfo * info)
     GString *names = g_string_new (NULL);
 
     for (i = 0; i < info->num_tensors; i++) {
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_common.c
```diff
@@ -534,7 +534,8 @@ gst_tensors_info_get_names_string (const GstTensorsInfo * info)
     GString *names = g_string_new (NULL);
 
     for (i = 0; i < info->num_tensors; i++) {
-      g_string_append (names, info->info[i].name);
+      if (info->info[i].name)
+        g_string_append (names, info->info[i].name);
```
dongju-chae: adding `{ }`  for consistency with the below lines?

jaeyun-jung: updated

Issue #1608, opened 2019-08-26
[DIST] Support macOS as a development environment


wooksong:
Macs has become even more popular as development machines. I think that we need to support macOS as one of our development environments as well.

#### Action items (TBD)
- [ ] Make nnstremer buildable with LLVM/clang on macOS
- [ ] Write a manual for How to build nnstreamer with LLVM/clang on macOS
- [ ] Deploy nnstreamer using '[brew](https://brew.sh/index)'

wooksong:
## Run unit tests
#### September 5, 2019

[PASSED] transform_transpose (7 passed among 7 cases)
[PASSED] nnstreamer_repo_dynamicity (10 passed among 10 cases)
[PASSED] nnstreamer_mux (84 passed among 84 cases)
[PASSED] nnstreamer_merge (89 passed among 89 cases)
[PASSED] transform_dimchg (5 passed among 5 cases)
[PASSED] nnstreamer_split (21 passed among 21 cases)
[PASSED] nnstreamer_filter_python (0 passed among 0 cases)
[FAILED] nnstreamer_repo_rnn (0 passed among 2 cases)
[FAILED] nnstreamer_repo_lstm (0 passed among 2 cases)
[PASSED] transform_stand (3 passed among 3 cases)
[PASSED] nnstreamer_converter (28 passed among 28 cases)
[PASSED] nnstreamer_filter_caffe2 (0 passed among 0 cases)
[PASSED] nnstreamer_decoder_image_labeling (0 passed among 0 cases)
[PASSED] nnstreamer_filter_tensorflow (0 passed among 0 cases)
[PASSED] nnstreamer_decoder (17 passed among 17 cases)
[PASSED] nnstreamer_filter_tensorflow_lite (0 passed among 0 cases)
[FAILED] codegen (4 passed among 12 cases)
[PASSED] nnstreamer_demux (37 passed among 37 cases)
[PASSED] nnstreamer_repo (66 passed among 66 cases)
[PASSED] transform_arithmetic (34 passed among 34 cases)
[PASSED] nnstreamer_decoder_boundingbox (6 passed among 6 cases)
[PASSED] nnstreamer_filter_pytorch (0 passed among 0 cases)
[FAILED] nnstreamer_filter_custom (10 passed among 22 cases)
[PASSED] transform_typecast (20 passed among 20 cases)
==================================================
[FAILED] There are failed test groups! (4)
         TC Passed: 441 / Failed: 24

[...]

PR #1643, opened 2019-09-05
[Tests] Use SO_EXT instead of hard-coded 'so'

wooksong:
This patch replaces the hard-coded filename extension, 'so', in the custom filter, repo_lstm, and repo_rnn test scripts with SO_EXT defined at ssat-api.

Partially resolves a sub-item of #1608 

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_filter_custom/runTest.sh
+++ b/tests/nnstreamer_filter_custom/runTest.sh
@@ -21,9 +21,15 @@ PATH_TO_PLUGIN="../../build"
 
 # Test for opencv installed, enable OPENCV test if opencv is found
 TEST_OPENCV="NO"
-/sbin/ldconfig -p | grep opencv >/dev/null 2>&1
-if [[ "$?" == 0 ]]; then
-    TEST_OPENCV="YES"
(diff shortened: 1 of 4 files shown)


tests/nnstreamer_filter_custom/runTest.sh
```diff
@@ -21,9 +21,15 @@ PATH_TO_PLUGIN="../../build"
...
 TEST_OPENCV="NO"
-/sbin/ldconfig -p | grep opencv >/dev/null 2>&1
-if [[ "$?" == 0 ]]; then
-    TEST_OPENCV="YES"
+if [[ "${KernelName}" != "Darwin" ]]; then
+    /sbin/ldconfig -p | grep opencv >/dev/null 2>&1
+    if [[ "$?" == 0 ]]; then
+        TEST_OPENCV="YES"
+    fi
+fi
```
myungjoo: Not uname to verify kernel names?

wooksong:
#### V2
- Replace hard-coded .so with ${SO_EXT} also in the rnn and lstm test scripts

PR #1650, opened 2019-09-09
[Android/Api] change package name

jaeyun-jung:
1. Change package name of Android API. (com.samsung.android.nnstreamer > org.nnsuite.nnstreamer)
2. Change directory structure (src > src/main)


api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
```diff
@@ -133,11 +133,11 @@ public static boolean initialize(Context context) {
         try {
             System.loadLibrary("nnstreamer-native");
             GStreamer.init(context);
-            return nativeInitialize(context);
         } catch (Exception e) {
             e.printStackTrace();
-            return false;
```
dongju-chae: Is it okay to keep the progress when errors occur?

jaeyun-jung: internally checks gst_is_initialized() when called nativeInitialize

dongju-chae: Oh, Okay

PR #1652, opened 2019-09-09
[Dist/RPM] Added library required by devel package

kparichay:
Added dependency of devel-capi on capi

**Self evaluation:**


packaging/nnstreamer.spec
```diff
@@ -171,6 +171,7 @@ You can construct a data stream pipeline with neural networks easily.
 %package -n capi-nnstreamer-single-new-devel
 Summary:	Tizen Native API Devel Kit for NNStreamer
 Group:		Multimedia/Framework
+Requires:	capi-nnstreamer-single-new = %{version}-%{release}
```
myungjoo: Question: can this be installed with capi-nnstreamer simultaneously?

kparichay: Yes, they both can be installed simultaneously (checked by successfully both of them simultaneously on tizen)

PR #1655, opened 2019-09-10
[Android/Api] add timeout in single-shot

jaeyun-jung:
Add method to set the timeout in single-shot API.

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
@@ -39,6 +39,7 @@ public final class SingleShot implements AutoCloseable {
     private native TensorsData nativeInvoke(long handle, TensorsData in);
     private native TensorsInfo nativeGetInputInfo(long handle);
     private native TensorsInfo nativeGetOutputInfo(long handle);
+    private native boolean nativeSetTimeout(long handle, int timeout);
 
     /**
      * Creates a new <code>SingleShot</code> instance with the given model.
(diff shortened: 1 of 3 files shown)


api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
```diff
@@ -143,6 +148,26 @@ public TensorsInfo getOutputInfo() {
...
 
+    /**
+     * Sets the maximum amount of time to wait for an output, in milliseconds.
+     *
+     * @param timeout The time to wait for an output
+     *
+     * @throws IllegalArgumentException if given param is invalid
+     * @throws IllegalStateException if failed to set the timeout
+     */
+    public void setTimeout(int timeout) {
```
myungjoo: Timeout was unsigned-int in C-API.

api/android/sample/src/main/java/org/nnsuite/nnstreamer/sample/MainActivity.java
```diff
@@ -262,6 +262,9 @@ private void runSingle() {
             TensorsInfo outInfo = single.getOutputInfo();
             printTensorsInfo(outInfo);
 
+            /* set timeout (1 second) */
+            single.setTimeout(1000);
```
myungjoo: Default timeout in C-API was 3 sec.

jaeyun-jung: This is sample code to set 1 second timeout.

api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
jaeyun-jung: Because java does not support unsigned type, I added an exception case in this method when timeout is negative value.

PR #1656, opened 2019-09-11
[Api/Single] invalid condition to check file extension

jaeyun-jung:
1. Fix invalid cond to check file extension.
2. Add proper log message and error code to set tensors info.


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -304,13 +304,16 @@ ml_single_open (ml_single_h * single, const char *model,
   } else {
     ml_tensors_info_h in_info;
 
-    status = ML_ERROR_INVALID_PARAMETER;
-    if (!klass->input_configured (single_h->filter))
+    if (!klass->input_configured (single_h->filter)) {
+      status = ML_ERROR_INVALID_PARAMETER;
```
dongju-chae: How about adding an error msg here too?

jaeyun-jung: I updated this PR to add log message.

PR #1663, opened 2019-09-16
[Filter] add common functions

jaeyun-jung:
Add common functions for tensor-filter and filter-single.
1. add struct for common property
2. reduce duplicated code of tensor-filter

### DIFF
--- a/api/capi/include/tensor_filter_single.h
+++ b/api/capi/include/tensor_filter_single.h
@@ -28,8 +28,7 @@
 #include <stdint.h>
 #include <glib-object.h>
 
-#include <nnstreamer/nnstreamer_subplugin.h>
-#include <nnstreamer/nnstreamer_plugin_api_filter.h>
+#include <nnstreamer/tensor_filter/tensor_filter_common.h>
 
 G_BEGIN_DECLS
 #define G_TYPE_TENSOR_FILTER_SINGLE \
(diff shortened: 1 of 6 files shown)


api/capi/include/tensor_filter_single.h
```diff
@@ -54,15 +55,7 @@ struct _GTensorFilterSingle
...
-  void *privateData; /**< NNFW plugin's private data is stored here */
-  GstTensorFilterProperties prop; /**< NNFW plugin's properties */
-  const GstTensorFilterFramework *fw; /**< The implementation core of the NNFW. NULL if not configured */
-
-  /* internal properties for tensor_filter_single */
-  gboolean silent; /**< Verbose mode if FALSE. int instead of gboolean for non-glib custom plugins */
-  gboolean started; /**< filter has been started */
-  GstTensorsConfig in_config; /**< input tensor info */
-  GstTensorsConfig out_config; /**< output tensor info */
+  GstTensorFilterPrivate priv; /**< Internal properties for tensor-filter */
```
myungjoo: Good!

api/capi/include/tensor_filter_single.h
```diff
@@ -30,6 +30,7 @@
 
 #include <nnstreamer/nnstreamer_subplugin.h>
 #include <nnstreamer/nnstreamer_plugin_api_filter.h>
+#include <nnstreamer/tensor_filter/tensor_filter_common.h>
```
myungjoo: Does this really need to include tensor_filter_common.h?
It'd be better if we can include tensor_filter_common.h in *.c only.

api/capi/include/tensor_filter_single.h
myungjoo: However, if the intention is to hide detailed from app writers, exposing tensor_filter_common.h is not good.

jaeyun-jung:
V2 (2019-09-17)
- remove unnecessary declaration in common header

api/capi/include/tensor_filter_single.h
myungjoo: ah.. this is a private header. please never mind.

api/capi/include/tensor_filter_single.h
```diff
@@ -28,8 +28,7 @@
 #include <stdint.h>
 #include <glib-object.h>
 
-#include <nnstreamer/nnstreamer_subplugin.h>
-#include <nnstreamer/nnstreamer_plugin_api_filter.h>
+#include "tensor_filter/tensor_filter_common.h"
```
myungjoo: uh... this doesn't look alright?
```
#include <nnstreamer/tensor_filter/tensor_filter_common.h>
```
?

myungjoo:
LGTM other than include path designation.

api/capi/include/tensor_filter_single.h
jaeyun-jung: Right, I just updated this PR.

PR #1669, opened 2019-09-17
[Single] Added unittests

kparichay:
Added more unittests for single-shot API testing
1. Multiple runs in parallel with varying timeouts
2. Changing timeout/closing the handle while calling the API in a different thread
3. Closing the handle multiple times
4. Check valid result after increasing the timeout, when earlier it was timing out

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -33,6 +33,8 @@
 
 #include "tensor_filter_single.h"
 
+#define ML_SINGLE_MAGIC 0xfeedfeed
+
 /**
  * @brief Default time to wait for an output in appsink (3 seconds).
  */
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -44,18 +46,27 @@
   (ts).tv_nsec = ((msec) % 1000) * 1000000; \
 } while (0)
 
+/** verify the magic number for ml_single obj */
+#define ML_SINGLE_MAGIC_CHECK(arg) do { \
+  if (arg->magic != ML_SINGLE_MAGIC) { \
+    ml_loge ("The given param, single is invalid."); \
+    return ML_ERROR_INVALID_PARAMETER; \
+  } \
+} while (0)
```
myungjoo: It'd be better to define in a common header for both C files.

kparichay: A common header would introduce big changes, irrelevant to this PR. How about a new PR for that?

myungjoo: Ok.

Please note that you may place an internal header at /api/capi/src/
or
Let nnstreamer-capi-single.c include tensor_filter_single.h, which minimizes changes.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -431,6 +443,9 @@ ml_single_close (ml_single_h single)
   }
 
   single_h = (ml_single *) single;
+  ML_SINGLE_MAGIC_CHECK (single_h);
```
jaeyun-jung: Discussion about validating the handle (maybe TODO list)
How about checking magic inside the mutex lock?
IMO, single_h->is_valid is unnecessary with single-magic.

api/capi/src/nnstreamer-capi-single-new.c
kparichay: I will add after #1672 is merged

api/capi/src/nnstreamer-capi-single.c
kparichay: `is_valid` has been removed
I think checking inside lock will assume that the lock is valid. However, if the object is invalid, then the lock can be invalid as well, which can cause the program to crash when trying to acquire the lock. Please correct me if I am wrong.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -401,7 +411,6 @@ ml_single_open (ml_single_h * single, const char *model,
   }
 
   g_mutex_lock (&single_h->lock);
```
jaeyun-jung: plz remove these two lines also (nothing to do here)

kparichay: done.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -559,7 +561,11 @@ ml_single_invoke (ml_single_h single,
   if (!sample) {
     ml_loge ("Failed to get the result from sink element.");
     single_h->clear_previous_buffer = TRUE;
+#ifdef ML_SINGLE_SUPPORT_TIMEOUT
```
jaeyun-jung: inside the feature (appsink without timeout feature does not need to clear old buffer)

```
#ifdef ML_SINGLE_SUPPORT_TIMEOUT
    single_h->clear_previous_buffer = TRUE;
```

kparichay: Right, updated.

PR #1676, opened 2019-09-19
[single] Thread safety bug fix - @open sesame 09/19 18:24

kparichay:
Added bug fix for thread safe usage
Single shot API use single handle lock for making all its API call thread safe
However, the call to close the API destroys the handle itself making the
call to close not thread safe
Added a global lock to make closing of handles thread safe

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -40,16 +40,32 @@
  */
 #define SINGLE_DEFAULT_TIMEOUT 3000
 
+/**
+ * @brief Global lock for single shot API
+ * @detail This lock ensures that ml_single_close is thread safe. All other API
+ *         functions use the mutex from the single handle. However for close,
+ *         single handle mutex cannot be used as single handle is destroyed at
+ *         close
+ * @note This mutex is automatically initialized as it is statically declared
+ */
+G_LOCK_DEFINE_STATIC (magic);
+
 /** Convert time in millisecond to timespec format */
 #define MSEC_TO_TIMESPEC(ts, msec) do { \
   (ts).tv_sec = (msec) / 1000; \
   (ts).tv_nsec = ((msec) % 1000) * 1000000; \
 } while (0)
 
-/** verify the magic number for ml_single obj */
-#define ML_SINGLE_MAGIC_CHECK(arg) do { \
-  if (arg->magic != ML_SINGLE_MAGIC) { \
+/**
+ * @brief Get valid handle after magic verification
+ * @note Magic lock is acquired after this
+ */
+#define ML_SINGLE_GET_VALID_HANDLE(single_h, single) do { \
+  G_LOCK (magic); \
+  single_h = (ml_single *) single; \
+  if (single_h->magic != ML_SINGLE_MAGIC) { \
     ml_loge ("The given param, single is invalid."); \
+    G_UNLOCK (magic); \
     return ML_ERROR_INVALID_PARAMETER; \
   } \
 } while (0)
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -54,6 +54,16 @@
...
 
+/**
+ * @brief Global lock for single shot API
+ * @detail This lock ensures that ml_single_close is thread safe. All other API
+ *         functions use the mutex from the single handle. However for close,
+ *         single handle mutex cannot be used as single handle is destroyed at
+ *         close
+ * @note This mutex is automatically initialized as it is statically declared
+ */
+static GMutex ml_single_mutex;
```
myungjoo: Use ```G_LOCK_DEFINE_STATIC()``` instead

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -523,10 +533,13 @@ ml_single_close (ml_single_h single)
     return ML_ERROR_INVALID_PARAMETER;
   }
 
+  g_mutex_lock (&ml_single_mutex);
   single_h = (ml_single *) single;
   ML_SINGLE_MAGIC_CHECK (single_h);
 
   single_h->magic = 0;
+  g_mutex_unlock (&ml_single_mutex);
+
```
myungjoo: Don't we need to protect ```single_h->magic``` access with single_h->mutex (what about moving LINE 543 -->  LINE 539)? So that other thread in a critical section of single_h->mutex won't get its magic value reset.?

api/capi/src/nnstreamer-capi-single-new.c
kparichay: updated

api/capi/src/nnstreamer-capi-single-new.c
kparichay: This is resolved with global lock

kparichay:
V2:
Now global lock for the API acquires lock while ensuring that the magic is correct.
The per handle lock is acquired inside this global lock ensuring that the single handle is not closed while using the current passed handle (if single handle lock acquired outside the global lock, handle can be closed in between)

I think this ensures that the CAPI is all thread safe

kparichay:
V3:
Added locking around ml_single_get_input/output_info for new CAPI implementation

V2:
Now global lock for the API acquires lock while ensuring that the magic is correct.
The per handle lock is acquired inside this global lock ensuring that the single handle is not closed while using the current passed handle (if single handle lock acquired outside the global lock, handle can be closed in between)

I think this ensures that the CAPI is all thread safe

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -523,11 +534,14 @@ ml_single_close (ml_single_h single)
     return ML_ERROR_INVALID_PARAMETER;
   }
 
+  G_LOCK (magic);
   single_h = (ml_single *) single;
   ML_SINGLE_MAGIC_CHECK (single_h);
```
myungjoo: Wouldn't be "safer" to have the macro defines:
```
  G_LOCK (magic);
  ML_SINGLE_MAGIC_CHECK (single_h);
```
in a single macro? e.g, ```#define ML_SINGLE_GET_VALID_HANDLE```.

kparichay: Updated

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -40,20 +40,41 @@
...
+/**
+ * @brief Get valid handle after magic verification
+ * @note Magic lock is acquired after this
+ */
+#define ML_SINGLE_GET_VALID_HANDLE(single_h, single) do { \
+  G_LOCK (magic); \
+  single_h = (ml_single *) single; \
+  ML_SINGLE_MAGIC_CHECK (single_h); \
+} while (0)
+
```
myungjoo: No one else is using ```ML_SINGLE_MAGIC_CHECK``` and no one else should use it. Let's remove ```ML_SINGLE_MAGIC_CHECK``` and do everything in this macro.

kparichay: updated

PR #1684, opened 2019-09-20
[single-new] Memory leak bug fix in error cases

kparichay:
Added memory leak bug fix in error cases running

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -72,7 +72,15 @@ G_LOCK_DEFINE_STATIC (magic);
  */
 #define ML_SINGLE_HANDLE_UNLOCK(single_h) g_mutex_unlock (&single_h->mutex);
 
-/* ML single api data structure for handle */
+/** States for invoke thread */
+typedef enum {
+  IDLE = 0,           /**< ready to accept next input */
+  RUNNING,            /**< running an input, cannot accept more input */
+  JOIN_REQUESTED,     /**< should join the thread, will exit soon */
+  ERROR               /**< error on thread, will exit soon */
+} thread_state;
+
+/** ML single api data structure for handle */
 typedef struct
 {
   GTensorFilterSingle *filter;  /**< tensor filter element */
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -557,6 +560,7 @@ ml_single_invoke (ml_single_h single,
 
   single_h->input = input;
   single_h->output = output;
+  *single_h->output = NULL;
```
dongju-chae: redundant with line 529?

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -176,6 +176,9 @@ invoke_thread (void *arg)
       for (i = 0; i < single_h->out_info.num_tensors; i++) {
         out_data->tensors[i].tensor = out_tensors[i].data;
       }
+    } else {
+      for (i = 0; i < single_h->out_info.num_tensors; i++)
+        g_free (out_tensors[i].data);
```
myungjoo: Please describe (comment) what the situation is when this ```else``` statement is used.

api/capi/src/nnstreamer-capi-single-new.c
kparichay: removed.

api/capi/src/nnstreamer-capi-single-new.c

kparichay:
V2:
Added state of the invoke thread
Handled case when repeated invoke called - now returns with TRY_AGAIN if current invoke isn't finished. If the invoke times out, returns and another invoke is called, this will return with TRY_AGAIN, if the previous invoke (which TIMED OUT) hasn't finished

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -545,14 +562,15 @@ ml_single_invoke (ml_single_h single,
     }
   }
 
-  if (single_h->data_ready == TRUE) {
+  if (single_h->state == RUNNING) {
```
dongju-chae: It seems that `single_h->state` should be `IDLE` to work. How about merging these state checking into a single line like `if (single_h->state != IDLE)` at the entry?

kparichay: You are right, updated.

api/capi/src/nnstreamer-capi-single-new.c
myungjoo: Good! I was wondering how it could be FALSE :)

kparichay: Earlier, `data_ready` was set to false in case of timeout by the caller. However, there was still a case missing when memory of output wasn't freed. Thanks for helping in pointing that out.

PR #1690, opened 2019-09-23
[Android/Test] script to run unittest

jaeyun-jung:
Update script to run the unittest on target device.

Note that, the testcases are android instrumented test which requires the target device.

Cmd to build and run test: ./build-android-lib.sh --run_unittest=yes

### DIFF
--- a/api/android/build-android-lib.sh
+++ b/api/android/build-android-lib.sh
@@ -16,6 +16,9 @@ nnstreamer_target_abi="'armeabi-v7a', 'arm64-v8a'"
 # Set tensorflow-lite version (available: 1.9 and 1.13)
 nnstreamer_tf_lite_ver=1.13
 
+# Run unittest after build procedure is done
+run_unittest='no'
+
 # Variables to release library (GROUP:ARTIFACT:VERSION)
 release_bintray='no'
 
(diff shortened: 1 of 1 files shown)


api/android/build-android-lib.sh
```diff
@@ -70,26 +76,27 @@ echo "GStreamer binaries: $GSTREAMER_ROOT_ANDROID"
...
 
 # Copy the files (native and java to build Android library) to build directory
-cp -r $NNSTREAMER_ROOT/api/android/* ./build_android_lib
+cp -r ./api/android/* ./build_android_lib
 
 # Get the prebuilt libraries and build-script
 svn --force export https://github.com/nnsuite/nnstreamer-android-resource/trunk/android_api ./build_android_lib
 
-pushd ./build_android_lib
+cd ./build_android_lib
```
myungjoo: Let's use ```pushd``` for maintainability. It's especially good if it's nested as we do here.

api/android/build-android-lib.sh
```diff
@@ -127,17 +135,27 @@ if [[ -e $nnstreamer_android_api_lib ]]; then
...
+
+        zip -r nnstreamer-api-unittest-$today.zip api/build/reports
+        cp nnstreamer-api-unittest-$today.zip ../$result_directory
+    fi
 else
     echo "Failed to build NNStreamer library."
 fi
 
-popd
+cd ..
```
myungjoo: Let's use popd, making it nested.

api/android/build-android-lib.sh
jaeyun-jung: roll back this change

api/android/build-android-lib.sh
jaeyun-jung: roll back this change

PR #1695, opened 2019-09-24
[Filter/nnfw] Tizen-nnfw skeleton @open sesame 9/27 13:16

myungjoo:
Skeleton code for Tizen-nnfw is added.
New nnfw of Tizen has its own interface; thus, we can no longer
rely on tensorflow-lite/nnapi for this.

This uses https://review.tizen.org/gerrit/gitweb?p=platform/core/ml/nnfw.git;a=summary as the backend.

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -1,3 +1,39 @@
+if get_option('enable-nnfw-runtime')
+  filter_sub_nnfw_sources = ['tensor_filter_nnfw.c']
+
+  nnstreamer_filter_nnfw_sources = []
+  foreach s : filter_sub_nnfw_sources
+    nnstreamer_filter_nnfw_sources += join_paths(meson.current_source_dir(), s)
+  endforeach
+
+  nnfw_dep = dependency('nnfw', required: false)
+  add_args = ''
+  nnstreamer_filter_nnfw_deps = [glib_dep, gst_dep, nnstreamer_dep]
+
+  if nnfw_dep.found()
+    nnstreamer_filter_nnfw_deps += nnfw_dep
+  else
+    # Until nnfw supports pkg-config, we need to do this primitively.
+    add_args = '-lnnfw-dev'
+  endif
+
+
+  shared_library('nnstreamer_filter_nnfw',
+    nnstreamer_filter_nnfw_sources,
+    dependencies: nnstreamer_filter_nnfw_deps,
+    c_args : add_args,
+    install: true,
+    install_dir: filter_subplugin_install_dir
+  )
+  static_library('nnstreamer_filter_nnfw',
+    nnstreamer_filter_nnfw_sources,
+    dependencies: nnstreamer_filter_nnfw_deps,
+    c_args : add_args,
+    install: false,
+    install_dir: filter_subplugin_install_dir
+  )
+endif
+
 if get_option('enable-tensorflow')
   filter_sub_tf_sources = [
     'tensor_filter_tensorflow.c',
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -1,0 +1,111 @@
+/**
+ * GStreamer Tensor_Filter, Tizen NNFW Module
+ * Copyright (C) 2019 MyungJoo Ham <myungjoo.ham@samsung.com>
+ *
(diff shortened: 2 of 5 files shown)


kparichay:
@myungjoo How about we disable aarch64 and armv7l build of nnfw for the time being?

myungjoo:
> @myungjoo How about we disable aarch64 and armv7l build of nnfw for the time being? These can be enabled later.

I think we can resolve this build error soon. I've submitted a fix to their repo today.

myungjoo:
> > @myungjoo How about we disable aarch64 and armv7l build of nnfw for the time being? These can be enabled later.
> 
> I think we can resolve this build error soon. I've submitted a fix to their repo today.

Nah.. but, let's disable nnfw build by default. Turn ```enable_nnfw``` on if you want to build it. They need to merge one more commit and SR it.

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -1,3 +1,36 @@
+if get_option('enable-nnfw-runtime')
+  filter_sub_nnfw_sources = ['tensor_filter_nnfw.c']
+
+  nnstreamer_filter_nnfw_sources = []
+  foreach s : filter_sub_nnfw_sources
+    nnstreamer_filter_nnfw_sources += join_paths(meson.current_source_dir(), s)
+  endforeach
+
+  nnfw_dep = dependency('nnfw', required: false)
```
kparichay: required: true?

myungjoo: No. nnfw is not producing a pkg-config file; thus we cannot use mseon's dependency, yet.

kparichay: I thought given `enable_nnfw` will false remain till this package has all the fixes

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -0,0 +1,110 @@
...
+  return 0;
+}
+
+static gchar filter_subplugin_nnfw[] = "nnfw";
+
+static GstTensorFilterFramework NNS_support_nnfw = {
+  .name = filter_subplugin_nnfw,
+  .allow_in_place = FALSE,
+  .allocate_in_invoke = FALSE,
+  .invoke_NN = nnfw_invoke,
```
jaeyun-jung: consider ```.run_without_model = FALSE``` for configuration

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -1,3 +1,36 @@
...
+  add_args = ''
+  if not nnfw_dep.found()
+    # Until nnfw supports pkg-config, we need to do this primitively.
+    add_args = '-lnnfw-dev'
+  endif
+
+  nnstreamer_filter_nnfw_deps = [nnfw_dep, glib_dep, gst_dep, nnstreamer_dep]
```
jaeyun-jung: plz check dependency

```
add_args = ''
nnstreamer_filter_nnfw_deps = [glib_dep, gst_dep, nnstreamer_dep]

if nnfw_dep.found()
  nnstreamer_filter_nnfw_deps += nnfw_dep
else
  add_args = '-lnnfw-dev'
```

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
myungjoo: Will update this.

packaging/nnstreamer.spec
```diff
@@ -114,6 +128,12 @@ Requires:	nnstreamer = %{version}-%{release}
 %description tensorflow-lite
 NNStreamer's tensor_fliter subplugin of TensorFlow Lite.
 
+%package nnfw
```
jaeyun-jung: ```%if 0%{?enable_nnfw} ```  is it necessary? (or enable_nnfw_r )

Question: why enable_nnfw_r (line 86) is required?

ext/nnstreamer/tensor_filter/meson.build
myungjoo: Updated.

packaging/nnstreamer.spec
myungjoo: With recent commit, ```enable_nnfw_r``` is no more required. Anyway, ```enable_nnfw``` is introduced as the nnfw-devel does not look so stable and this is Tizen-only.

PR #1685, opened 2019-09-20
[API/Tizen] Update the comment for Tizen API

again4you:
When pipeline string directly accesses the camera or recorder device of
the Tizen, appropriate privilege should be written on configuration
file. This patch new adds the 'remarks' comment for this privilege.

Signed-off-by: Sangjung Woo <sangjung.woo@samsung.com>

### Self evaluation:
    Build test: [* ]Passed [ ]Failed [ ]Skipped
    Run test: [* ]Passed [ ]Failed [ ]Skipped

PR #1705, opened 2019-09-26
[Api/Tizen] check cam privilege

jaeyun-jung:
Check camera and recoder privilege for each src element.


jaeyun-jung:
Related PR #1685 (update API description about src element)

PR #1707, opened 2019-09-27
[Tests] Make tool scripts written in python be compatible with python3 @open sesame 9/27 15:09

wooksong:
There are several platforms which do not support python2.7 by default anymore. To prepare this issue, this patch revises all python scripts in tests to be also compatible with python3.


leemgs:
I fixed the report modules of CI system.

PR #1710, opened 2019-09-30
[Android/Test] tc fail case

jaeyun-jung:
Sometimes testcase failed with pipeline state NULL when stopping the pipeline.
Add sleep time before closing the pipeline.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
@@ -86,6 +86,7 @@ public class APITestPipeline {
         try (Pipeline pipe = new Pipeline(desc, null)) {
             Thread.sleep(100);
             assertEquals(NNStreamer.PIPELINE_STATE_PAUSED, pipe.getState());
+            Thread.sleep(100);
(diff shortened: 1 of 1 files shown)


api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
```diff
@@ -119,6 +119,7 @@ public void onStateChanged(int state) {
             Thread.sleep(300);
 
             assertEquals(NNStreamer.PIPELINE_STATE_PAUSED, mPipelineState);
+            Thread.sleep(100);
```
myungjoo: Is 100ms enough always? or would it require dynamic-sleeping (e.g.,  https://github.com/nnsuite/nnstreamer/blob/3095de8e4e6596d80509c06139517cc31790ea2f/tests/tizen_capi/unittest_tizen_capi.cpp#L130 )

jaeyun-jung: IMO, 100ms is enough in this line. This is to prevent the pipeline close for a while in the end of the loop.

myungjoo:
PTAL @again4you @helloahn Let's go release with this! :)

PR #1716, opened 2019-10-01
[Tizen/nnfw] Add open/close callbacks for tensor_filter

myungjoo:
This validates integration mechanisms for nnfw-runtime of Tizen.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -25,9 +25,11 @@
  *
  * @todo Check if nnfw supports dynamic input dimension. (if so, we need to supply setInputDim)
  * @todo Decide whether to let nnfw allocate output buffers or we will feed output buffers to nnfw.
+ * @todo NYI: configuring backends
  *
  */
 
+#include <string.h>
 #include <glib.h>
 #include <tensor_common.h>
 #include <nnstreamer_plugin_api_filter.h>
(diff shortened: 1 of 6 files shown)


myungjoo:
I will add nnfw_runtime_r condition to the unittest invocation in the spec file.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -37,45 +39,130 @@
...
+  if (pdata && pdata->session) {
+    NNFW_STATUS status = nnfw_close_session (pdata->session);
+
+    if (status != NNFW_STATUS_NO_ERROR) {
+      g_printerr ("cannot close nnfw-runtime session for %s",
+          pdata->model_path);
+    }
+  } else {
+    g_printerr ("nnfw_close called without proper nnfw_open");
+    return;
```
jaeyun-jung: free model_path and pdata
when called close callback in tensor_filter, private_data will be set null.

myungjoo: Updated.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -37,45 +39,131 @@
...
 {
-  return 0;
+  NNFW_STATUS status;
+  int err = 0;
+  nnfw_pdata *pdata;
+
+  if (*private_data != NULL) {
+    pdata = *private_data;
+    if (strcmp (prop->model_file, pdata->model_path)) {
+      nnfw_close (prop, private_data);  /* "reopen" */
```
kparichay: do we need to close and reopen the session - how about just skipping the `create_session` in this case?

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -37,45 +39,131 @@
...
+        prop->model_file);
+    goto unalloc_exit;
+  }
+
+  pdata->model_path = g_strdup (prop->model_file);
+
+  goto exit;
+
+unalloc_exit:
+  g_free (pdata);
```
kparichay: add `nnfw_close (prop, private_data)`?

tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cpp
```diff
@@ -0,0 +1,57 @@
...
+    .fw_opened = 0,
+    .model_file = "null.nnfw",
+  };
+  void *data;
+
+  const GstTensorFilterFramework *sp = nnstreamer_filter_find ("nnfw");
+  EXPECT_NE (sp, (void *) NULL);
+
+  ret = sp->open (&prop, &data);
+  EXPECT_NE (ret, 0);
```
kparichay: add close here?

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -37,45 +39,131 @@
...
+  NNFW_STATUS status;
+  int err = 0;
+  nnfw_pdata *pdata;
+
+  if (*private_data != NULL) {
+    pdata = *private_data;
+    if (strcmp (prop->model_file, pdata->model_path)) {
+      nnfw_close (prop, private_data);  /* "reopen" */
+    } else {
+      return 1;
```
jaeyun-jung: TODO we have to handle 'already loaded' case.
e.g, return 1 makes assertion in tf & tf-lite now.

myungjoo: Yup. That is to be handled at tensor_filter_common.c and tensor_filter.c 
You may create a backlog issue and resolve it later :)

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
myungjoo: At this stage we don't know how nnfw-runtime will behave if we call load-model repeatedly in a single session. You may optimize it further later once we get it working.

kparichay: Sounds good.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
myungjoo: The final states of variables are identical with nnfw_close() call; however, if we call nnfw_close(), it will be a inappropriate call for nnfw_close(), which will incur error (```"nnfw_close called without proper nnfw_open"```) by nnfw_close(), because nnfw_close assumes that nnfw_open is ended successfully.

kparichay: Then, I think line 88 and 95 should call `nnfw_close_session()` to cleanup already opened session.

Issue #1717, opened 2019-10-01
[UnitTest FAIL] Parallel CAPI issue not fixed.


myungjoo:
We definitely need to re-enable unit-test in aarch64/Tizen builds.

```
$ gbs build -A aarch64 --define "unit_test 1"
...
[  164s] [       OK ] nnstreamer_capi_singleshot.benchmark_time (1464 ms)
[  164s] [ RUN      ] nnstreamer_capi_singleshot.invoke_03
[  164s] [       OK ] nnstreamer_capi_singleshot.invoke_03 (63 ms)
[  164s] [ RUN      ] nnstreamer_capi_singleshot.failure_01_n
[  164s] [       OK ] nnstreamer_capi_singleshot.failure_01_n (87 ms)
[  164s] [ RUN      ] nnstreamer_capi_singleshot.invoke_timeout
[  168s] [       OK ] nnstreamer_capi_singleshot.invoke_timeout (3910 ms)
[  168s] [ RUN      ] nnstreamer_capi_singleshot.parallel_runs
[  171s] ../tests/tizen_capi/unittest_tizen_capi.cpp:1947: Failure
[  171s]       Expected: status
[  171s]       Which is: -1073741823
[  171s] To be equal to: ML_ERROR_NONE
[  171s]       Which is: 0
[  171s] ../tests/tizen_capi/unittest_tizen_capi.cpp:1948: Failure
[  171s] Value of: output != NULL
[  171s]   Actual: false
[  171s] Expected: true
[  171s] ../tests/tizen_capi/unittest_tizen_capi.cpp:1947: Failure
[  171s]       Expected: status
[  171s]       Which is: -1073741823
[  171s] To be equal to: ML_ERROR_NONE
[  171s]       Which is: 0
...
[  284s] [  FAILED  ] nnstreamer_capi_singleshot.parallel_runs (116530 ms)
[  284s] [ RUN      ] nnstreamer_capi_singleshot.close_while_running
[  286s] [       OK ] nnstreamer_capi_singleshot.close_while_running (1349 ms)
[  286s] [----------] 8 tests from nnstreamer_capi_singleshot (126154 ms total)
[  286s] 
[  286s] [----------] Global test environment tear-down
[  286s] [==========] 31 tests from 8 test cases ran. (133220 ms total)
[  286s] [  PASSED  ] 30 tests.
[  286s] [  FAILED  ] 1 test, listed below:
[...]

PR #1719, opened 2019-10-02
[URGENT] [single] Increase default timeout

kparichay:
Increase the default timeout to pass the test case on a slow machine/not fully
optimized tensorflow-lite. Also, as the running time for tf-lite test case in
aarch64 is over 10x compared to armv7l, higher defailt timeout is better.

In unit-test, increased the corresponding default timeout as well.
Further, when running many test cases in parallel, default timeout is scaled
with number of parallel runs.

Solves #1717

**Self evaluation:**

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cpp
+++ b/tests/tizen_capi/unittest_tizen_capi.cpp
@@ -14,6 +14,8 @@
 #include <glib/gstdio.h> /* GStatBuf */
 #include <nnstreamer-capi-private.h>
 
+#define SINGLE_DEF_TIMEOUT_MSEC 10000
+
 /**
  * @brief Struct to check the pipeline state changes.
  */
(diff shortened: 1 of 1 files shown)


tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -2066,7 +2070,7 @@ TEST (nnstreamer_capi_singleshot, parallel_runs)
 
   for (i=0; i<num_cases; i++) {
     ss_data[i].test_model = test_model;
-    ss_data[i].num_runs = 10;
+    ss_data[i].num_runs = 3;
```
kparichay: `num_runs = 10` takes over 2-3 mins to pass for armv7l and aarch64 tizen build.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -75,9 +75,9 @@ G_LOCK_DEFINE_STATIC (magic);
   g_mutex_unlock (&single_h->lock);
 
 /**
- * @brief Default time to wait for an output in appsink (3 seconds).
+ * @brief Default time to wait for an output in appsink (10 seconds).
  */
-#define SINGLE_DEFAULT_TIMEOUT 3000
+#define SINGLE_DEFAULT_TIMEOUT 10000
```
jaeyun-jung: If default value should be changed, we may need another ACR for release.
Is it possible to remain this value, and change only the testcase? (@again4you plz see this change)

kparichay: > Is it possible to remain this value, and change only the testcase?

@jaeyun-jung I think this can be done.

kparichay: Done.

PR #1728, opened 2019-10-07
[Filter/Sub] already loaded case

jaeyun-jung:
Check returned value when the nn model file is already loaded.


myungjoo:
@jaeyun-jung Do you think this is required by Tizen 5.5?

PR #1730, opened 2019-10-07
[single] Minor updates for modularity

kparichay:
Added minor updates in C-API single shot implementation for improved modularity
This makes the code skeleton at par with existing single-shot implementation

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -72,6 +72,15 @@ G_LOCK_DEFINE_STATIC (magic);
  */
 #define ML_SINGLE_HANDLE_UNLOCK(single_h) g_mutex_unlock (&single_h->mutex);
 
+/** define string names for input/output */
+#define INPUT_STR "input"
+#define OUTPUT_STR "output"
+#define TYPE_STR "type"
+#define NAME_STR "name"
+
+/** concat string from #define */
+#define CONCAT_MACRO_STR(STR1,STR2) STR1 STR2
+
 /** States for invoke thread */
 typedef enum {
   IDLE = 0,           /**< ready to accept next input */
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -198,17 +198,86 @@ invoke_thread (void *arg)
...
+  ml_tensors_info_s *input_info;
+  GstTensorsInfo gst_info;
+  gchar *val;
+  guint rank;
+  gchar *prop_prefix, *prop_name;
+
+  if (is_input)
+    prop_prefix = g_strdup ("input");
+  else
+    prop_prefix = g_strdup ("output");
```
myungjoo: These better be a static constant.
E.g., 
```
static const inputstr[] = "input";
static const outputstr[] = "ouput";
...
  prop_prefix = inputstr;
```

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -198,17 +198,86 @@ invoke_thread (void *arg)
...
+  /* allocate handle for tensors info */
+  ml_tensors_info_create (info);
+  input_info = (ml_tensors_info_s *) (*info);
+
+  ml_tensors_info_initialize (input_info);
+  gst_tensors_info_init (&gst_info);
+
+  /* get dimensions */
+  prop_name = g_strdup_printf ("%s", prop_prefix);
+  g_object_get (filter, prop_name, &val, NULL);
```
myungjoo: property_name of g_object_get is const char *; thus, you don't need strdup. You can even directly give "input" here (or static const gchar *)

api/capi/src/nnstreamer-capi-single-new.c
kparichay: Updated with static string.

api/capi/src/nnstreamer-capi-single-new.c

kparichay:
> There is not too much dynamicity in string values. Let's use static-const gchar* rather than malloc-ed buffers. With a few tweaks, you can handle postfixes as well.

Updated with static definitions with #define and concat these strings defined with macro, which can be combined at compile time.

PR #1735, opened 2019-10-07
[API/Tizen] Remove version number from libcapi-nnstreamer.so @open sesame 10/07 18:53

again4you:
This patch removes the version number from libcapi-nnstreamer shared
object.
* libcapi-nnstreamer.so.0.3 -> libcapi-nnstreamer.so

**Self evaluation:**


packaging/nnstreamer.spec
```diff
@@ -381,25 +381,23 @@ cp -r result %{buildroot}%{_datadir}/nnstreamer/unittest/
 %files -n capi-nnstreamer
 %manifest capi-nnstreamer.manifest
 %license LICENSE
-%{_libdir}/libcapi-nnstreamer.so.*
+%{_libdir}/libcapi-nnstreamer.so
```
myungjoo: What about adding symlinks of libcapi-nnstreamer.so.0 for the compatibility against Tizen-FX C# files?

again4you: ACK!

debian/nnstreamer.install
```diff
@@ -1,4 +1,5 @@
 /usr/lib/nnstreamer/decoders/libnnstreamer_decoder_*.so
 /usr/lib/*/gstreamer-1.0/*.so
+/usr/lib/*/libcapi-*.so
```
myungjoo: Remove ```/usr/lib/*/libcapi-*.so.*``` for Ubuntu.

PR #1740, opened 2019-10-08
[CodeClean] Apply regular expressions

abcinje:
Use regular expressions instead of combined if-statements

Signed-off-by: Injae Kang abcinje@gmail.com


myungjoo:
You'll need to explain why we need this change in the commit message. That is what ```TAOS/pr-prebuild-nobody ....``` wants.

PR #1743, opened 2019-10-08
[Filter/TF-lite] cleanup hw-type string

jaeyun-jung:
1. remove duplicated src and fix indent
2. move hw-type string definition


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
```diff
@@ -69,6 +66,19 @@ get_nnapi_hw_type (const gchar * str)
   return (index < 0) ? NNAPI_UNKNOWN : index;
 }
 
+/**
```
myungjoo: Why have you moved this function definition?

jaeyun-jung: just for the code readability

PR #1744, opened 2019-10-08
[Api/Tizen] remove code which occurs error

helloahn:
remove code which occurs error.
Since the previous code lock the resource internally,
when another pipeline want to contact to the resource, an error is generated.

**Self evaluation:**


myungjoo:
Please elaborate what's the error and how it is fixed.

jaeyun-jung:
@myungjoo When called resource release callback, we should not release the handle. So I requested to remove this and set the flag invalid, so that when the application restarts the pipeline, it will acquire the res handle again.

jaeyun-jung:
@helloahn Please update commit message for reviewers.

helloahn:
@myungjoo 
the comment is updated. sorry for the insincerity.

PR #1749, opened 2019-10-10
[Api/Tizen] handle dpm restriction case

jaeyun-jung:
Check the pipeline with device policy restricted case.
Create dpm handle and when trying to start the pipeline, check device policy and handle error case.

### DIFF
--- a/api/capi/src/nnstreamer-capi-tizen.c
+++ b/api/capi/src/nnstreamer-capi-tizen.c
@@ -28,6 +28,7 @@
 #include <glib.h>
 
 #include <system_info.h>
+#include <restriction.h> /* device policy manager */
 #include <privacy_privilege_manager.h>
 #include <mm_resource_manager.h>
 #include <mm_camcorder.h>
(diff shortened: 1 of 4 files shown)


api/capi/src/nnstreamer-capi-tizen.c
```diff
@@ -474,11 +543,12 @@ ml_tizen_mm_res_release (gpointer handle, gboolean destroy)
  * @brief Function to initialize mm resource manager.
  */
 static int
-ml_tizen_mm_res_initialize (ml_pipeline_h pipe)
+ml_tizen_mm_res_initialize (ml_pipeline_h pipe, gboolean has_video_src, gboolean has_audio_src)
 {
   ml_pipeline *p;
   pipeline_resource_s *res;
   tizen_mm_handle_s *mm_handle;
+  gboolean failed = FALSE;
```
dongju-chae: I think this flag is redundant because `rm_error` always handles failed cases.
In `rm_error`, checking whether `mm_handle` is NULL would be enough.

jaeyun-jung: You are right, I just changed to check the error case with ```status != ML_ERROR_NONE```

Issue #1677, opened 2019-09-19
[Tizen] Write Tensor_Filter::nnfw (2019.10.)


myungjoo:
Write tensor_filter::nnfw that connects: https://review.tizen.org/gerrit/gitweb?p=platform/core/ml/nnfw.git;a=summary

It appears that it now has independent interfaces.

This is not for 5.5 M2, but 6.0 M1. However, this requires to be implemented during 2010.10~11.

The interface file is /runtime/include/nnfw.h
(Do not access other interfaces)

myungjoo:
@kparichay Could you please share the status of tensor_filter::nnfw-runtime here?

kparichay:
Issues related to running nnfw with cpu backend have been resolved
However, that patch has not been uploaded to main tizen repo
So, in order to test this patch, clone nnfw repo and build locally
Also, building nnfw repo installs libbackend_acl_cl.so in /usr/lib in emulator
Delete this file to test nnstreamer with cpu backend with nnfw successfully #1819 
This is because libbackend_acl_cl.so exists, but libOpencl is not

NYI items will be implemented this week

myungjoo:
If there are pending nnfw commits and you need them, you can add nnfw at https://build.tizen.org/project/show/devel:AIC:Tizen:5.0:nnsuite and build the unmerged or un-SR-ed commit there for GBS in our CI server and your own PC.

kparichay:
Managing backends will be done with #1849.
Close after #1842.

PR #1750, opened 2019-10-10
[tizen/nnfw] Add get_input/output_dimensions

kparichay:
Added implementation for get_input/output_dimensions functions

Refer #1677

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -26,6 +26,7 @@
  * @todo Check if nnfw supports dynamic input dimension. (if so, we need to supply setInputDim)
  * @todo Decide whether to let nnfw allocate output buffers or we will feed output buffers to nnfw.
  * @todo NYI: configuring backends
+ * @todo NYI: support quant8,bool type
  *
  */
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -141,14 +142,125 @@ nnfw_close (const GstTensorFilterProperties * prop, void **private_data)
...
+  }
+
+  return err;
+}
+
+
+/**
+ * @brief Copy nnfw format info of tensor to gst format info
+ * @param[in] gst_info info give in gst format
+ * @param[out] nnfw_info info container to receive tensor info in gst format
```
dongju-chae: is it really `param[out]`?

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -141,14 +142,125 @@ nnfw_close (const GstTensorFilterProperties * prop, void **private_data)
...
  */
 static int
 nnfw_getInputDim (const GstTensorFilterProperties * prop,
     void **private_data, GstTensorsInfo * info)
 {
-  return 0;
+  nnfw_pdata *pdata;
+  int err = 0;
+
+  g_assert (*private_data);
```
dongju-chae: How about returning an error instead of g_assert because it's callable from outside?
Also, it seems that the `err` variable is useless.

kparichay: Updated `g_assert` to `g_return_val_if_fail`
Using `err` as I dont want to return a value returned from another function directly.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -141,14 +142,126 @@ nnfw_close (const GstTensorFilterProperties * prop, void **private_data)
...
+
+  if (info->num_tensors > NNS_TENSOR_SIZE_LIMIT)
+    return -ERANGE;
+
+  /** Now fill each outputs */
+  for (idx = 0; idx < info->num_tensors; idx ++) {
+    if (is_input)
+      nnfw_input_tensorinfo (pdata->session, idx, &nnfw_info_t);
+    else
+      nnfw_output_tensorinfo (pdata->session, idx, &nnfw_info_t);
```
myungjoo: Let's handle return values of nnfw_{input/output}_tensorinfo.

kparichay: Updated.

PR #1754, opened 2019-10-11
[tizen/nnfw] Add invoke for nnfw

kparichay:
Added invoke implementation for nnfw along with some bug fixes

Refer #1677

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -162,6 +162,79 @@ nnfw_getOutputDim (const GstTensorFilterProperties * prop,
 }
 
 /**
+ * @brief Convert from nnfw type to gst tensor type
+ * @param[in] type type given in gst format
+ * @param[out] nnfw_type container to receive type in nnfw tensor format
+ * @return 0 on sucess, negative errno on error
+ */
+static int
+nnfw_tensor_type_from_gst (const tensor_type type, NNFW_TYPE * nnfw_type)
+{
+  int err = 0;
+
+  switch (type) {
+    case _NNS_FLOAT32:
+      *nnfw_type = NNFW_TYPE_TENSOR_FLOAT32;
+      break;
+    case _NNS_INT32:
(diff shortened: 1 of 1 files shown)


jijoongmoon:
I think this should be working differently depending on the hw to use and we have to hand over this hw information to nnfw (I'm not sure they have..). If it is, then have to conform it is run on that hw (may be GPU). And in order to get the hw, you could use filter property.

kparichay:
> I think this should be working differently depending on the hw to use and we have to hand over this hw information to nnfw (I'm not sure they have..). If it is, then have to conform it is run on that hw (may be GPU). And in order to get the hw, you could use filter property.

Yes, that is the part of configuring `backends`, which is a NYI item. It will be added as a subsequent PR, once the base version starts to run.
However, the interface for invoke is same for all the HW types, just setting backend is required using the interface provided. 
I will the confirm the behavior after the invoke unit-test is complete.

myungjoo:
When you start working on configuring hw backends of nnfw, please look after the consistency with other filter subplugins; i.e., tf-lite nnapi configurations. To keep the consistency, you may modify tf-lite side for hw backend configurations as well.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -161,6 +161,79 @@ nnfw_getOutputDim (const GstTensorFilterProperties * prop,
...
+  g_return_val_if_fail (prop != NULL, -EINVAL);
+  g_return_val_if_fail (mem != NULL, -EINVAL);
+  g_return_val_if_fail (pdata->session != NULL, -EPERM);
+
+  if (is_input) {
+    g_return_val_if_fail (prop->input_configured == TRUE, -EPERM);
+    num_tensors = prop->input_meta.num_tensors;
+  } else {
+    g_return_val_if_fail (prop->output_configured == TRUE, -EPERM);
+    num_tensors = prop->input_meta.num_tensors;
```
dongju-chae: Oh, it should be `prop->output_meta`.

kparichay: that's big mistake, updated.

PR #1767, opened 2019-10-15
[Android] add log to print version string

jaeyun-jung:
add log to print nnstreamer, gstreamer and glib version info.


myungjoo:
Why do you need this?

jaeyun-jung:
This is to get the version info of gstreamer and glib.

PR #1776, opened 2019-10-16
[tests] Locate all models and data together @open sesame 10/16 18:46

kparichay:
Locate all the models and data together for easier management

**Self evaluation:**


tests/nnstreamer_filter_python/runTest.sh
```diff
@@ -47,13 +47,13 @@ popd
 
 # Passthrough test
 export PYTHONPATH=../../build/ext/nnstreamer/tensor_filter/:$PYTHONPATH
-PATH_TO_SCRIPT="python/passthrough.py"
+PATH_TO_SCRIPT="../test_models/models/passthrough.py"
```
dongju-chae: I think it's more like `data` rather than `models`.

kparichay: But this is passed a model file to tensor_filter as it operates on the data passed to tensor_filter.

dongju-chae: Yes... I reused the existing model parameter to support this. Then, it's fine.

dongju-chae: But... it was related to the internal implementation.
I think it's still weird if these example python custom filter scripts are co-located with `*.pb` in `/models/` folder.

kparichay: I think these are models as they provide the specification to take in data, process them and give output, just like .pb file. The only difference is pb file needs a framework like tensorflow to understand them but these .py files are self sufficient.
They much closer to models than data.

dongju-chae: Yes, makes sence

PR #1777, opened 2019-10-16
[tests] Install models/data with tests @open sesame 10/16 19:28

kparichay:
Install models/data with tests correctly when needed
Install tizen-capi-unittest with install-test option
Install python and caffe2 unittests

**Self evaluation:**
    
### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -96,43 +96,58 @@ if gtest_dep.found()
   endif
 endif
 
+# Install data required for unittest
+unittest_tests_install_dir = join_paths(unittest_install_dir,'tests')
+tensor_filter_ext_enabled = get_option('enable-tensorflow-lite') or
+    get_option('enable-python') or  get_option('enable-tensorflow') or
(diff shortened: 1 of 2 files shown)


tests/meson.build
```diff
@@ -96,43 +96,58 @@ if gtest_dep.found()
   endif
 endif
 
+# Install data required for unittest
+unittest_tests_install_dir = join_paths(unittest_install_dir,'tests')
+tensor_filter_ext_enabled = get_option('enable-tensorflow-lite') or
```
jaeyun-jung: How about installing test_models without the conditions of each nnfw option?

Do we need to check these 'enable-xxx' options?
If 'install-test' is true, then just install all the data in test_models folder.

kparichay: As far as I understand, the test_models are only used by these nnfw options (used in the condition). If neither is being enabled, then I think installing test_models with `install-test` will not be useful.

PR #1778, opened 2019-10-16
[Test] update model in tensor-filter

jaeyun-jung:
Add testcase to re-open nn model in tensor-filter using gst-harness.

### DIFF
--- a/tests/nnstreamer_plugins/unittest_plugins.cpp
+++ b/tests/nnstreamer_plugins/unittest_plugins.cpp
@@ -14,6 +14,7 @@
 #include <gst/check/gsttestclock.h>
 #include <gst/check/gstharness.h>
 #include <tensor_common.h>
+#include <nnstreamer_plugin_api_filter.h>
 
 #include "../gst/nnstreamer/tensor_transform/tensor_transform.h"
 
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_plugins/unittest_plugins.cpp
```diff
@@ -2849,6 +2849,75 @@ TEST (test_tensor_transform, orc_performance)
...
+  EXPECT_TRUE (ret != GST_STATE_CHANGE_FAILURE); \
+  while (cur_state != state && counter < 20) { \
+    g_usleep (50000); \
+    counter++; \
+    ret = gst_element_get_state (element, &cur_state, NULL, 5 * GST_MSECOND); \
+    EXPECT_TRUE (ret != GST_STATE_CHANGE_FAILURE); \
+  } \
+  EXPECT_TRUE (cur_state == state); \
+  g_usleep (50000); \
+} while (0)
```
myungjoo: This is worth to be a static-inline function...?

jaeyun-jung: I wanted to surely guarantee the pipeline state changes (playing > pause > update properties and set playing again)

jaeyun-jung:
PR updated
2019-10-16 add case to push dummy buffer to tensor-filter

PR #1780, opened 2019-10-17
[Security] handle returned value @open sesame 10/17 15:00

jaeyun-jung:
Check mem allocation failure case and refactor file-close in IIO.

This PR is related with ssms-1016-0006.

### DIFF
--- a/api/android/api/src/main/jni/nnstreamer-native-customfilter.c
+++ b/api/android/api/src/main/jni/nnstreamer-native-customfilter.c
@@ -47,6 +47,9 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
   guint i;
   int ret = -1;
 
+  in_data = out_data = NULL;
+  in_info = out_info = NULL;
+
   /* get pipe info and init */
   pipe_info = g_hash_table_lookup (g_customfilters, prop->fwname);
   g_return_val_if_fail (pipe_info, -1);
@@ -55,16 +58,28 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
   g_return_val_if_fail (env, -1);
 
   in_data = g_new0 (ml_tensors_data_s, 1);
-  g_assert (in_data);
+  if (in_data == NULL) {
+    nns_loge ("Failed to allocate memory for input tensors data.");
+    goto done;
+  }
 
   out_data = g_new0 (ml_tensors_data_s, 1);
-  g_assert (out_data);
+  if (out_data == NULL) {
+    nns_loge ("Failed to allocate memory for output tensors data.");
+    goto done;
+  }
 
   in_info = g_new0 (ml_tensors_info_s, 1);
-  g_assert (in_info);
+  if (in_info == NULL) {
+    nns_loge ("Failed to allocate memory for input tensors info.");
+    goto done;
+  }
 
   out_info = g_new0 (ml_tensors_info_s, 1);
-  g_assert (out_info);
+  if (out_info == NULL) {
+    nns_loge ("Failed to allocate memory for output tensors info.");
+    goto done;
+  }
 
   /* convert to c-api data type */
   in_data->num_tensors = prop->input_meta.num_tensors;
@@ -161,6 +176,8 @@ nns_customfilter_set_dimension (const GstTensorFilterProperties * prop, void **p
   JNIEnv *env;
   int ret = -1;
 
+  in = out = NULL;
+
   /* get pipe info and init */
   pipe_info = g_hash_table_lookup (g_customfilters, prop->fwname);
   g_return_val_if_fail (pipe_info, -1);
@@ -169,10 +186,16 @@ nns_customfilter_set_dimension (const GstTensorFilterProperties * prop, void **p
   g_return_val_if_fail (env, -1);
 
   in = g_new0 (ml_tensors_info_s, 1);
-  g_assert (in);
+  if (in == NULL) {
+    nns_loge ("Failed to allocate memory for input tensors info.");
+    goto done;
+  }
 
   out = g_new0 (ml_tensors_info_s, 1);
-  g_assert (out);
+  if (out == NULL) {
+    nns_loge ("Failed to allocate memory for output tensors info.");
+    goto done;
+  }
 
   /* convert to c-api data type */
   ml_tensors_info_copy_from_gst (in, in_info);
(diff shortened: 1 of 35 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -71,6 +71,7 @@ construct_element (GstElement * e, ml_pipeline * p, const char *name,
     ml_pipeline_element_e t)
 {
   ml_pipeline_element *ret = g_new0 (ml_pipeline_element, 1);
+  g_assert (ret != NULL);
```
myungjoo: Asserting (killing the app) in a API library may be... a bit too much.

Can we let it return error along with the users of ``construct_element```?

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -262,6 +263,7 @@ loadImageLabels (bounding_boxes * data)
   labels = g_strsplit (contents, "\n", -1);
   data->total_labels = g_strv_length (labels);
   data->labels = g_new0 (char *, data->total_labels);
+  g_assert (data->labels != NULL);
```
myungjoo: Let's make some "handle-able" error returns instead of assert so that we can let applications do something before killing itself.

myungjoo:
Looks good in overall.

However, let's try to minimize asserting and killing user applications unless it's a case of a critical internal bug. Let the application handle if it's what the application is wrong or the execution environment is wrong.

jaeyun-jung:
PR updated
2019-10-17 I tried to avoid assertion case when mem allocation is failed.

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
```diff
@@ -137,8 +152,10 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
     (*env)->DeleteLocalRef (env, obj_out_info);
   (*env)->DeleteLocalRef (env, cls_custom);
 
-  g_free (in_data);
-  g_free (out_data);
+  if (in_data)
+    g_free (in_data);
```
dongju-chae: Is it necessary? g_free will check internally whether `in_data` is NULL or not

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
```diff
@@ -55,16 +58,28 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
...
-  g_assert (out_data);
+  if (out_data == NULL) {
+    nns_loge ("Failed to allocate memory for output tensors data.");
+    goto done;
+  }
 
   in_info = g_new0 (ml_tensors_info_s, 1);
-  g_assert (in_info);
+  if (in_info == NULL) {
+    nns_loge ("Failed to allocate memory for input tensors info.");
```
myungjoo: it appears that you don't free this when it fails.

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
```diff
@@ -55,16 +58,28 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
...
+    goto done;
+  }
 
   out_info = g_new0 (ml_tensors_info_s, 1);
-  g_assert (out_info);
+  if (out_info == NULL) {
+    nns_loge ("Failed to allocate memory for output tensors info.");
```
myungjoo: here as well.

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
```diff
@@ -137,8 +152,10 @@ nns_customfilter_invoke (const GstTensorFilterProperties * prop, void **private_
     (*env)->DeleteLocalRef (env, obj_out_info);
   (*env)->DeleteLocalRef (env, cls_custom);
 
-  g_free (in_data);
-  g_free (out_data);
+  if (in_data)
```
kparichay: No need of NULL check before g_free
https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-free

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -422,13 +422,16 @@ ml_single_open (ml_single_h * single, const char *model,
...
-    return status;
+    return ML_ERROR_NOT_SUPPORTED;
   }
 
   /** Create ml_single object */
   single_h = g_new0 (ml_single, 1);
-  g_assert (single_h);
+  if (single_h == NULL) {
+    ml_loge ("Failed to allocate the single handle.");
+    return ML_ERROR_STREAMS_PIPE;
```
kparichay: ML_ERROR_STREAM_PIPE is described as `/**< Cannot create or access GStreamer pipeline. */`
As the single-new does not use GStreamer, how about some other error?
Is it possible to introduce equivalent of ENOMEM?

api/capi/src/nnstreamer-capi-tizen.c
```diff
@@ -691,7 +691,11 @@ ml_tizen_mm_res_acquire (ml_pipeline_h pipe,
       mm_res = (pipeline_resource_s *) g_hash_table_lookup (mm_handle->res_handles, res_key);
       if (!mm_res) {
         mm_res = g_new0 (pipeline_resource_s, 1);
-        g_assert (mm_res);
+        if (mm_res == NULL) {
+          ml_loge ("Failed to allocate media resource data.");
+          g_free (res_key);
```
kparichay: Please set the status here before `goto rm_error`.

api/capi/src/nnstreamer-capi-single-new.c
jaeyun-jung: I will change this to ML_ERROR_UNKNOWN now.
But later, we need to update this error code.

TIZEN_ERROR_OUT_OF_MEMORY = -ENOMEM
I will mark TODO to replace no-mem error code.
Let's update error code after 5.5 M2.

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
kparichay: aah, this is duplicate.

api/android/api/src/main/jni/nnstreamer-native-customfilter.c
jaeyun-jung: I removed condition to check null ptr.

Issue #1651, opened 2019-09-09
Flexible Input Dimension for Single-Shot API


kparichay:
Support flexible input dimension for Single-Shot API, where user can adjust input size for each inference. This will not be supported for tensorflow-lite.

myungjoo:
Note that flexible input dimension has become POSSIBLE by the new single-shot API implementation; the old implementation (gst-pipeline) cannot handle it. :)

kparichay:
- [] Add setInOutDimension() callback for tensor filter plugins - sets input and output dimensions
- [] Support setInOutDimension()/setInputDimension() for various tensor filter subplugins which can support flexible input dimensions (setInputDimension() should do a sample run on zero tensor to get output tensor shape if cannot infer the shape from model directly)
  - [] Tensorflow
  - [] Pytorch
  - [] Caffe2
  - [] Tensorflow-lite - https://www.tensorflow.org/lite/api_docs/cc/class/tflite/interpreter#resizeinputtensor
  - [] python
- [] Add interface in nnstreamer.h to set flexible input+output dimension
- [] Add support for flexible input/output dimension in new single-c-api and not implemented for old version
- [] Add test cases

myungjoo:
You don't need to support all frameworks from the start.
I'd recommend to do the whole steps with a single framework (TF or TF-Lite) and do the whole steps up to API test cases.

Then, later, expand the feature for other frameworks by extending internal-APIs (setInputOuputDim) and writing additional callbacks for them.

kparichay:
Ok, then I will start with TF-lite, and extend it later.

PR #1785, opened 2019-10-17
[single] Added API and skeleton for flexible input dimension

kparichay:
Adding support for flexible input dimension
Added API and skeleton for setting input dimension via single API

**Self evaluation:**

### DIFF
--- a/api/capi/include/nnstreamer-single.h
+++ b/api/capi/include/nnstreamer-single.h
@@ -139,6 +139,19 @@ int ml_single_get_input_info (ml_single_h single, ml_tensors_info_h *info);
 int ml_single_get_output_info (ml_single_h single, ml_tensors_info_h *info);
 
 /**
+ * @brief Sets the information (tensor dimension, type, name and so on) of required input data for the given model.
+ * @details Note that a model/framework may not support setting such information.
+ * @param[in] single The model handle.
+ * @param[in] info The handle of input tensors information.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED This implies that the given framework does not support dynamic dimensions.
+ *         Use ml_single_set_input_info/ml_single_get_output_info APIs instead for this framework.
+ * @retval #ML_ERROR_INVALID_PARAMETER Fail. The parameter is invalid.
+ */
+int ml_single_set_input_info (ml_single_h single, const ml_tensors_info_h info);
+
+/**
  * @brief Sets the maximum amount of time to wait for an output, in milliseconds.
  * @since_tizen 5.5
  * @param[in] single The model handle.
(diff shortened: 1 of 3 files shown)


api/capi/include/nnstreamer-single.h
```diff
@@ -138,6 +138,18 @@ int ml_single_get_input_info (ml_single_h single, ml_tensors_info_h *info);
...
 int ml_single_get_output_info (ml_single_h single, ml_tensors_info_h *info);
 
+/**
+ * @brief Sets the information (tensor dimension, type, name and so on) of required input data for the given model.
+ * @details Note that a model may not support setting such information.
+ * @param[in] single The model handle.
+ * @param[in] info The handle of input tensors information.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported.
```
myungjoo: Let's add more description on this error, ```ML_ERROR_NOT_SUPPORTED```, for application developers:
```
Not supported. This implies that the given model or framework does not support dynamic dimensions. Use get input/output ..... APIs instead for this model and framework.
```

myungjoo:
In later commits, let's think about macros / inline-or-simple functions that provides syntactic sugar for making it easier to use dynamic input dimensions. For example:
```
ml_single_invoke_dynamic ( handle[in], input[in], input_dim[in], output[out], output_dim[out]);
```
, which can be something :
```
#define ml_single_invoke_dynamic( ... )  ml-api; ml-api2; ...
```
or 
a static inline function.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -615,7 +615,7 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
 }
 
 /**
- * @brief Gets the type (tensor dimension, type, name and so on) of required input data for the given handle.
+ * @brief Gets the information (tensor dimension, type, name and so on) of required input data for the given handle.
```
dongju-chae: Q. the given handle == the given model?

kparichay: Yes, as this is single API, one handle corresponds to one (single) model.

kparichay:
> In later commits, let's think about macros / inline-or-simple functions that provides syntactic sugar for making it easier to use dynamic input dimensions. For example:

Yes, will add these. Added as TODO in #1651.

api/capi/include/nnstreamer-single.h
kparichay: Updated to :
```This implies that the given **~model or~** framework does not support dynamic dimensions. Use get input/output (( fill these in ..... )) APIs instead for **~this model and~** framework.```

There is no way to distinguish if the given input information set by the user is invalid or the model does allow setting dynamic dimensions. So, I think both these cases should be ML_ERROR_INVALID_PARAMETER. However, for framework, we can return ML_ERROR_NOT_SUPPORTED.

kparichay:
V2:
Added more description for retval of ML_ERROR_NOT_SUPPORTED in documentation

PR #1787, opened 2019-10-17
[single] Implementation of flexible input dimension

kparichay:
Added implementation of flexible input dimension at the API level
Handling flexible input dimension at the tensor filter extension
scope to be done.

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -95,6 +95,7 @@ typedef struct
   GTensorFilterSingle *filter;  /**< tensor filter element */
   ml_tensors_info_s in_info;    /**< info about input */
   ml_tensors_info_s out_info;   /**< info about output */
+  ml_nnfw_type_e nnfw;          /**< nnfw type for this filter */
   guint magic;                  /**< code to verify valid handle */
 
   GThread *thread;              /**< thread for invoking */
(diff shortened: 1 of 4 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -732,6 +734,27 @@ ml_single_set_timeout (ml_single_h single, unsigned int timeout)
...
+
+  if (!single)
+    return ML_ERROR_INVALID_PARAMETER;
+
+  ML_SINGLE_GET_VALID_HANDLE_LOCKED (single_h, single, 0);
+
+  switch (single_h->nnfw) {
+    case ML_NNFW_TYPE_TENSORFLOW_LITE:
+      klass = g_type_class_peek (G_TYPE_TENSOR_FILTER_SINGLE);
+      if (klass->set_input_info (single_h->filter, info) == FALSE)
```
jaeyun-jung: TODO fix possible svace issue later
if (klass == NULL || klass->set_input_info (single_h->filter, info) == FALSE)

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -450,6 +451,7 @@ ml_single_open (ml_single_h * single, const char *model,
    * 3. Construct a pipeline
    * Set the pipeline desc with nnfw.
    */
+  single_h->nnfw = nnfw;
```
dongju-chae: Is there any possibility to overwrite `single_h->nnfw` when this open API is called twice?
Then, it's better to check this at the beginning.

kparichay: With every open to `ml_single_open`, a new `single_h` object is made. So, if this open API is called twice, two `single_h` objects will be allocated and returned to the user.
So, I think overwrite should not happen.

dongju-chae: Oh, okay

api/capi/src/nnstreamer-capi-single-new.c
kparichay: @jaeyun-jung https://github.com/nnsuite/nnstreamer/pull/1820

PR #1790, opened 2019-10-18
[TF-Lite] Added setInputDimension

kparichay:
Added implementation for setInputDimension
This can allow support for flexible input dimensions
The support is limited to only changing dimensions
Data type for the input or the number of inputs cannot be changed with current TF-Lite API
The rank for the input is allowed to be changed if the model allows
The input dimension can return error if the model does not allow dimension to be changed

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
@@ -210,6 +210,23 @@ tflite_getOutputDim (const GstTensorFilterProperties * prop,
   return tflite_core_getOutputDim (tf->tflite_private_data, info);
 }
 
+/**
+ * @brief The optional callback for GstTensorFilterFramework
+ * @param prop property of tensor_filter instance
+ * @param private_data : tensorflow lite plugin's private data
+ * @param in_info The dimesions and types of input tensors
+ * @param[out] out_info The dimesions and types of output tensors
+ */
+static int
+tflite_setInputDim (const GstTensorFilterProperties * prop, void **private_data,
+    const GstTensorsInfo * in_info, GstTensorsInfo * out_info)
+{
+  tflite_data *tf;
+  tf = *private_data;
+  g_assert (*private_data);
+  return tflite_core_setInputDim (tf->tflite_private_data, in_info, out_info);
+}
+
 static gchar filter_subplugin_tensorflow_lite[] = "tensorflow-lite";
 
 static GstTensorFilterFramework NNS_support_tensorflow_lite = {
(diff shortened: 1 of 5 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
```diff
@@ -206,6 +206,23 @@ tflite_getOutputDim (const GstTensorFilterProperties * prop,
   return tflite_core_getOutputDim (tf->tflite_private_data, info);
 }
 
+/**
+ * @brief The optional callback for GstTensorFilterFramework
+ * @param prop property of tensor_filter instance
+ * @param private_data : tensorflow lite plugin's private data
+ * @param in_info The dimesions and types of input tensors
```
dongju-chae: How about adding `@param[in]`?

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.h
```diff
@@ -49,10 +49,13 @@ class TFLiteCore
   int init ();
   int loadModel ();
   const char* getModelPath ();
+  int setTensorProp (const std::vector<int> &tensor_idx_list,
```
myungjoo: private?

kparichay: yes, this should be private.
Updated.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -310,6 +304,58 @@ TFLiteCore::getOutputTensorDim (GstTensorsInfo * info)
...
+{
+  TfLiteStatus status;
+  const std::vector<int> &input_idx_list = interpreter->inputs ();
+
+  /** Cannot change the number of inputs */
+  if (info->num_tensors != input_idx_list.size ())
+    return -EINVAL;
+
+  for (int tensor_idx = 0; tensor_idx < info->num_tensors; ++tensor_idx) {
+    std::vector<int> dims;
```
myungjoo: Do ```std::vector<int> dims(NNS_TENSOR_RANK_LIMIT);``` and remove Line 330

kparichay: Removed reserve and directly allocated memory.

gst/nnstreamer/tensor_common.c
```diff
@@ -222,6 +222,24 @@ gst_tensor_info_copy (GstTensorInfo * dest, const GstTensorInfo * src)
...
+ * @note Minimum rank is 1
+ */
+int
+gst_tensor_info_get_rank (const GstTensorInfo * info)
+{
+  int idx;
+
+  /** rank is at least 1 */
+  for (idx = NNS_TENSOR_RANK_LIMIT - 1; idx > 0; idx--) {
+    if (info->dimension[idx] != 1)
```
jaeyun-jung: Later we need to consider how to get the exact rank.

PR #1802, opened 2019-10-21
[single] simplify get_tensors_info in single API @open sesame 10/22 11:55

kparichay:
Added simplification for get_tensors_info in single API implementation
input and output tensors info is stored in single API
This patch updates to use that information than extracting the info again from element

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single-new.c
+++ b/api/capi/src/nnstreamer-capi-single-new.c
@@ -672,6 +672,7 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
 {
   ml_single *single_h;
   int status = ML_ERROR_NONE;
+  ml_tensors_info_s *input_info;
 
   check_feature_state ();
 
@@ -680,8 +681,22 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
 
   ML_SINGLE_GET_VALID_HANDLE_LOCKED (single_h, single, 0);
 
-  ml_single_get_tensors_info_from_filter (single_h->filter, is_input, info);
+  /* allocate handle for tensors info */
+  status = ml_tensors_info_create (info);
+  if (status != ML_ERROR_NONE)
+    goto exit;
 
+  input_info = (ml_tensors_info_s *) (*info);
+
+  if (is_input)
+    status = ml_tensors_info_clone (input_info, &single_h->in_info);
+  else
+    status = ml_tensors_info_clone (input_info, &single_h->out_info);
+
+  if (status != ML_ERROR_NONE)
+    ml_tensors_info_destroy (input_info);
+
+exit:
   ML_SINGLE_HANDLE_UNLOCK (single_h);
   return status;
 }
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -673,7 +674,14 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
 
   ML_SINGLE_GET_VALID_HANDLE_LOCKED (single_h, single, 0);
 
-  ml_single_get_tensors_info_from_filter (single_h->filter, is_input, info);
+  /* allocate handle for tensors info */
+  ml_tensors_info_create (info);
```
dongju-chae: Is `ml_tensors_info_create` always successful?
Otherwise, `*info` may generate a segmentation fault.

kparichay: you are right, updated.

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -680,8 +681,19 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
...
-  ml_single_get_tensors_info_from_filter (single_h->filter, is_input, info);
+  /* allocate handle for tensors info */
+  status = ml_tensors_info_create (info);
+  if (status != ML_ERROR_NONE)
+    goto exit;
+
+  input_info = (ml_tensors_info_s *) (*info);
+
+  if (is_input)
+    ml_tensors_info_clone (input_info, &single_h->in_info);
```
wooksong: ```ml_tensors_info_clone``` has a return value. Don't we need to check it?

api/capi/src/nnstreamer-capi-single-new.c
```diff
@@ -680,8 +681,19 @@ ml_single_get_tensors_info (ml_single_h single, gboolean is_input,
...
+  else
+    ml_tensors_info_clone (input_info, &single_h->out_info);
```
wooksong: Ditto.

api/capi/src/nnstreamer-capi-single-new.c
myungjoo: input_info is not initialized at this point. ```ml_tensors_info_clone ``` may generate seg-fault or return ML_ERROR_INVALID_PARAMETER (if input_info is NULL)

myungjoo: If app developer's given parameter (info) is incorrect (NULL), ```ml_tensors_info_clone``` may return ERROR.

myungjoo: We need to check it although most of errors of ```ml_tensors_info_clone``` are going to be prevented by checking the return of ```ml_tensors_info_create```.

api/capi/src/nnstreamer-capi-single-new.c
kparichay: Updated.

kparichay:
> `` ml_single_get_tensors_info_from_filter`` checked various error cases. I wonder that the return value of ``ml_tensors_info_clone``` can be ignored.

`ml_tensors_info_clone` returns failure when the source and destination tensors info are invalid. In this function (`ml_single_get_tensors_info`), they are both verified to be valid. So, I think `ml_tensors_info_clone` should not fail in this usage. However, to avoid confusion in future, I have added the error check.

Issue #1809, opened 2019-10-21
[pytorch] Try/Catch for the first run


kparichay:
PyTorch does not have an interface to get the shape of an input. This makes it impossible to verify the shape of the input set by the user without doing the actual. However, this creates a possibility of a run failing, and causing a crash without any handling. I suggest that the first run can be wrapped in a try/catch, so as to exit gracefully for some of the errors - as this wont catch errors like segmentation fault.

@helloahn can you verify if its the same case of Caffe2 as well?

kparichay:
After discussion with @helloahn, similar case exist for caffe2 as well.
Added #1846 for caffe2.

PR #1810, opened 2019-10-22
[pytorch] Do first run with try/catch @open sesame 10/22 15:50

kparichay:
First run after configuring the tensor filter with pytorch framework
is now run with a try/catch.

Check #1809 for more details.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_pytorch_core.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_pytorch_core.cc
@@ -44,6 +44,7 @@ TorchCore::TorchCore (const char *_model_path)
 {
   model_path = _model_path;
   configured = false;
+  first_run = true;
 
   gst_tensors_info_init (&inputTensorMeta);
   gst_tensors_info_init (&outputTensorMeta);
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_pytorch_core.cc
```diff
@@ -44,6 +44,7 @@ TorchCore::TorchCore (const char *_model_path)
 {
   model_path = _model_path;
   configured = false;
+  first_run = false;
```
myungjoo: Why is it ```false``` at constructor?

If you want to check if init() is called before invoke, make it "tri-state" (uninit-ed / first / second).

kparichay: Changed to `true` at constructor. Actually, this value does not matter, as init() will be called before invoke(), which will set the value.

> If you want to check if init() is called before invoke, make it "tri-state" (uninit-ed / first / second).

As this ext is called by tensor_filter, I think there is no need to worry about invoke() before init().

PR #1835, opened 2019-10-28
[Api] error code for no-mem

jaeyun-jung:
define error code for no-mem case.


myungjoo:
Update doxygen entries of API headers. Let developers aware that their API calls may return with NOMEM.

jaeyun-jung:
@myungjoo Added error code for related APIs

PR #1838, opened 2019-10-30
[CodeClean] reduce complexity @open sesame 11/08 07:03

jaeyun-jung:
1. declare function to iterate the elements in pipeline description
2. fix api declaration (params in same line)


leemgs:
After this PR:

image

PR #1846, opened 2019-11-07
[caffe2] Do first run with try/catch 

kparichay:
First run after configuring the tensor filter with caffe2 framework is now run with a try/catch

Check #1809  for more details

**Self evaluation:**

cc. @helloahn

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
@@ -47,6 +47,7 @@ Caffe2Core::Caffe2Core (const char * _model_path, const char *_model_path_sub)
   g_assert (_model_path != NULL && _model_path_sub != NULL);
   pred_model_path = g_strdup (_model_path);
   init_model_path = g_strdup (_model_path_sub);
+  first_run = true;
 
   gst_tensors_info_init (&inputTensorMeta);
   gst_tensors_info_init (&outputTensorMeta);
@@ -86,6 +87,8 @@ Caffe2Core::init (const GstTensorFilterProperties * prop)
     g_critical ("Failed to initialize input tensor\n");
     return -2;
   }
+
+  first_run = true;
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_caffe2_core.cc
```diff
@@ -300,7 +303,27 @@ Caffe2Core::run (const GstTensorMemory * input, GstTensorMemory * output)
...
+    } catch(const std::exception& ex)	{
+      g_critical ("Exception while running the model : %s", ex.what());
+      return -4;
+    } catch (...) {
+      g_critical ("Unknown exception while running the model");
+      return -4;
+    }
+  } else {
+    workSpace.RunNet (predictNet.name ());
+  }
```
dongju-chae: Does it have some problems to wrap `workspace.RunNet()` with try/catch in next runs?

kparichay: In my understanding, wrapping with try/catch block adds overhead, and will make the execution slower.

dongju-chae: Well.. this overhead is very little because running a network already takes a few milliseconds.
We can ignore the costs in most cases because most overheads come from throwing an exception and handling it. 

So, IMO, next runs may also have a possibility to generate runtime errors.
How about just wrapping try/catch for next runs as well?

kparichay: The purpose of this try catch is to only catch errors related to mismatch of input/output.
I didn't know that the overhead of try/catch is very small when exception handling is not very frequent.
Yes, wrapping for all runs makes sense as well.
Then I think this should be done for all the tensor filter extensions?

dongju-chae: Oh, if its purpose is just to check the mismatch of input/output, not runtime errors, I think it's okay.

PR #1847, opened 2019-11-07
[tensor_filter/tflite] Update for supporting various accelerators @open sesame 11/15 15:35

kparichay:
Added update to support various accelerators for tensor filters
~This is an initial design
Please provide your comments.
Note: this only provides the design, and implementation and support for all the sub-plugins will be 
added after initial review.~
Updated tensorflow-lite nnapi setting based on these accelerators
Added testcases for tensorflow-lite

Refer #1849 

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -26,21 +26,69 @@
 
 #include "tensor_typedef.h"
 
+/** Macros for accelerator types */
+#define ACCL_NONE_STR "none"
+#define ACCL_AUTO_STR "auto"
+#define ACCL_CPU_STR  "cpu"
+#define ACCL_GPU_STR  "gpu"
+#define ACCL_NPU_STR  "npu"
+#define ACCL_NEON_STR "neon"
+#define ACCL_SRCN_STR "srcn"
+#define ACCL_DEF_STR  "default"
+
+#define REGEX_ACCL_EACH_ELEM(ELEM) "[^!]" ELEM "[,]?"
+
+#define REGEX_ACCL_AUTO REGEX_ACCL_EACH_ELEM(ACCL_AUTO_STR)
+#define REGEX_ACCL_CPU  REGEX_ACCL_EACH_ELEM(ACCL_CPU_STR)
+#define REGEX_ACCL_GPU  REGEX_ACCL_EACH_ELEM(ACCL_GPU_STR)
+#define REGEX_ACCL_NPU  REGEX_ACCL_EACH_ELEM(ACCL_NPU_STR)
+#define REGEX_ACCL_NEON REGEX_ACCL_EACH_ELEM(ACCL_NEON_STR)
+#define REGEX_ACCL_SRCN REGEX_ACCL_EACH_ELEM(ACCL_SRCN_STR)
+#define REGEX_ACCL_DEF  REGEX_ACCL_EACH_ELEM(ACCL_DEF_STR)
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 /**
- * @brief nnapi hw properties.
+ * @brief acceleration hw properties.
+ * @details Enabling acceleration (choosing any accelerator hardware other
+ *          ACCL_NONE) will enable acceleration for the framework dependent on
+ *          the acceleration supported by the framework.
+ *
+ *          For example, enabling acceleration with tflite will enable NNAPI.
+ *          However, with NNFW will enable GPU/NEON etc.
+ *
+ *          Appropriate acceleration should be used with each framework. For
+ *          example, ACCL_NEON is only supported with NNFW tensor filter. Using
+ *          ACCL_NEON with pytorch would result in a warning message, and
+ *          the accelerator would fallback on ACCL_AUTO.
+ *
+ *          ACCL_AUTO automatically chooses the accelerator based on the ones
+ *          supported by the subplugin framework. However, ACCL_DEFAULT would
+ *          use the accelerator set by the subplugin framework, if any.
+ *
+ * @note Acceleration for a framework can be enabled/disabled in the build. If
+ *       the acceleration for a framework was disabled in the build, setting the
+ *       accelerator while use a tensor filter with that framework will have no
+ *       effect.
  */
 typedef enum
 {
-  NNAPI_CPU = 0,
-  NNAPI_GPU = 1,
-  NNAPI_NPU = 2,
+  /**< no explicit acceleration */
+  ACCL_NONE = 0,    /**< no acceleration (defaults to CPU) */
 
-  NNAPI_UNKNOWN
-} nnapi_hw;
+  /** Enables acceleration, 0xn000 any version of that device, 0xnxxx: device # xxx-1 */
+  ACCL_AUTO = 0x1,        /**< choose optimized device automatically */
+  ACCL_CPU  = 0x1000,     /**< specify device as CPU, if possible */
+  ACCL_GPU  = 0x2000,     /**< specify device as GPU, if possible */
+  ACCL_NPU  = 0x4000,     /**< specify device as NPU, if possible */
+  ACCL_NEON = 0x8000,     /**< specify device as NEON, if possible */
+  ACCL_SRCN = 0x10000,    /**< specify device as SRCN, if possible */
+
+  /** If there is no default config, and device needs to be specified, fallback to ACCL_AUTO */
+  ACCL_DEFAULT = 0x20000    /**< use default device configuration by the framework*/
+} accl_hw;
 
 /**
  * @brief GstTensorFilter's properties for NN framework (internal data structure)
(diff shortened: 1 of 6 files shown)


gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -31,16 +31,44 @@ extern "C" {
...
 {
-  NNAPI_CPU = 0,
-  NNAPI_GPU = 1,
-  NNAPI_NPU = 2,
+  /**< no explicit acceleration */
+  ACCL_NONE = 0,    /**< no acceleration (defaults to CPU) */
+
+  /** Enables acceleration */
+  ACCL_AUTO = 1,    /**< choose device automatically */
+  ACCL_CPU  = 2,    /**< specify device as CPU, if possible */
```
wooksong: Since there is the ACCL_NONE mode, ACCL_CPU might be misled that this means accelerations using SIMD. I suggest merging ACCL_CPU and  ACCL_NEON into ACCL_CPU_SIMD.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -31,16 +31,44 @@ extern "C" {
...
+  ACCL_GPU  = 3,    /**< specify device as GPU, if possible */
+  ACCL_NPU  = 4,    /**< specify device as NPU, if possible */
+  ACCL_NEON = 5,    /**< specify device as NEON, if possible */
+  ACCL_SRCN = 6,    /**< specify device as SRCN, if possible */
```
wooksong: Q: SRCN? What is it?

kparichay: SRCN is the backend supported by NNFW.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: As per [nnapi](https://developer.android.com/ndk/guides/neuralnetworks), NNAPI can fallback on CPU for some operations. However, that can be different from the frameworks default implementation for the CPU (https://github.com/tensorflow/tensorflow/issues/28283#issuecomment-488442568). So, ACCL_CPU denotes accelerators implementation for the CPU, and ACCL_NONE denotes direct support for the CPU by the framework itself.

kparichay: Note that currently, NNAPI does not allow choosing a particular device for acceleration as of now. ACCL_* is just to support this kind of mechanism, if needed in future.

wooksong: > As per [NNAPI](https://developer.android.com/ndk/guides/neuralnetworks), NNAPI can fallback on CPU for some operations. However, that can be different from the frameworks default implementation for the CPU ([tensorflow/tensorflow#28283 (comment)](https://github.com/tensorflow/tensorflow/issues/28283#issuecomment-488442568)). So, ACCL_CPU denotes accelerators implementation for the CPU, and ACCL_NONE denotes direct support for the CPU by the framework itself.

I am still confusing:) Are there any ways to accelerate NNs using CPU without SIMD? 
- ACCL_NONE: just CPU
- ACCL_CPU: acceleration using whatever in CPU
- ACCL_NEON: acceleration using ARM SIMD in ARM CPU 
?

Give me more intuition, please.

myungjoo:
This is going to be a major change, setting future directions.
Please try to gather as many reviewers as possible.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
myungjoo: 1. What if a user wants to express "Use either GPU or NPU. Try to avoid using CPU" or "Use NPU-2 for this filter, NPU-1 for that filter."?

2. Can we make this aligned with /api/capi/include/nnstreamer.h::ml_nnfw_hw_e ?

kparichay: > I am still confusing:) Are there any ways to accelerate NNs using CPU without SIMD?
> 
> * ACCL_NONE: just CPU
> * ACCL_CPU: acceleration using whatever in CPU
> * ACCL_NEON: acceleration using ARM SIMD in ARM CPU
>   ?
> 
> Give me more intuition, please.

My bad, I should have made this more clearer - 
- ACCL_NEON: specific for nnfw backend for now. Will represent SIMD if any framework supports it.
- ACCL_NONE: just CPU
- ACCL_CPU: this is still CPU, but now the backend of CPU will use the implementation provided with the acceleration library supported by the framework. For eg - if a framework provides acceleration for CPU using a custom-made BLAS lib, then it would fall in this category. 

Whether they use SIMD internally or not, cant say. However, if some acceleration states that they have explicitly used SIMD (like nnfw `acl_neon` vs `cpu` mode), then ACCL_SIMD could be used.

Another example where ACCL_CPU and ACCL_NONE would be different is with tflite running in a system with no GPU/NPU/DSP support for NNAPI - 
- If NNAPI is used, then tflite would use the NNAPI CPU implementation - ACCL_CPU
- if NNAPI is not used, then tflite would use its own CPU implementation - ACCL_NONE
In this case, if NNAPI gives the option to choose the backend to run (sometime in future), then IMO ACCL_CPU would differ from ACCL_NONE.

Please correct me if I am wrong.

kparichay: > 1. What if a user wants to express "Use either GPU or NPU. Try to avoid using CPU" or "Use NPU-2 for this filter, NPU-1 for that filter."?

I think for this, input can be taken as a regular expression and then interpreted correspondingly.
Like `true:(NPU,GPU,!(CPU))`. 

> "Use NPU-2 for this filter, NPU-1 for that filter."

In this case, as the accelerator property is set for each filter in the pipeline independently, I think this scenario is automatically covered. 

> 2. Can we make this aligned with /api/capi/include/nnstreamer.h::ml_nnfw_hw_e ?

Updated the values for the `enum accl_hw`. Most of the attributes align. However, can `ML_NNFW_HW_ANY` be covered with `ACCL_DEFAULT` or should `ACCL_ANY` be introduced?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -26,21 +26,72 @@
...
+  /**< no explicit acceleration */
+  ACCL_NONE = 0,    /**< no acceleration (defaults to CPU) */
+
+  /** Enables acceleration, 0xn000 any version of that device, 0xnxxx: device # xxx-1 */
+  ACCL_AUTO = 1000,    /**< choose optimized device automatically */
+  ACCL_CPU  = 2000,    /**< specify device as CPU, if possible */
+  ACCL_GPU  = 3000,    /**< specify device as GPU, if possible */
+  ACCL_NPU  = 4000,    /**< specify device as NPU, if possible */
+  ACCL_NEON = 5000,    /**< specify device as NEON, if possible */
+  ACCL_SRCN = 6000,    /**< specify device as SRCN, if possible */
```
myungjoo: 0x is missing?

myungjoo: In case we may extend this with backward compatibility, I'd recommend to use
```
0x1000,
0x2000,
0x4000,
0x8000,
....
```

kparichay: I will update to this style
And update accl_hw_string in `tensor_filter_common.c`. With the enum values growing exponentially, that char * array has exploded.

myungjoo: Ah... yes.. with non-continuous enum, ```*accl_hw_string[ACCL_DEFAULT + 1]``` should be avoided. :)

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -301,9 +313,14 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
...
-  g_object_class_install_property (gobject_class, PROP_NNAPI,
-      g_param_spec_string ("nnapi", "NNAPI",
-          "Enable Neural Newtorks API ?", "", G_PARAM_READWRITE));
+  g_object_class_install_property (gobject_class, PROP_ACCELERATOR,
+      g_param_spec_string ("accelerator", "ACCELERATOR",
+          "Set accelerator for the subplugin with format "
+          "(true/false):(comma separated ACCELERATOR(s)). "
+          "true/false determines if accelerator is to be used. "
+          "list of accelerators determines the backend (ignored with false). "
+          "Example, if GPU and NPU can be used but not CPU - (GPU,NPU,!CPU).",
```
myungjoo: ```(true/false):(comma separated ACCELERATOR(s))```
<->
```(GPU,NPU,!CPU)```
doesn't look consistent.

kparichay: updated the description to `true:(GPU,NPU,!CPU)`

PR #1850, opened 2019-11-07
[Dist] Add Tizen Sensord (sensor-fw) Option

myungjoo:
Enable Tizen Sensord (/platform/core/system/sensord.git)

This is a place-holder without actual code.


ext/nnstreamer/meson.build
```diff
@@ -1,2 +1,2 @@
-subdir('tensor_decoder')
-subdir('tensor_filter')
+subdir('tensor_decoder')
```
kparichay: I can't find and understand the changes in this file.

myungjoo: It has gone thru dos2unix. It shuold've been in another commit.

myungjoo:
Because this will have dependency on Tizen C-API, I'll move it to ext/ and make it independent.

PR #1854, opened 2019-11-08
[nnfw] Fix meson @open sesame 11/15 15:56

kparichay:
Fix variables in meson.build for nnfw and nnfw-runtime

**Self evaluation:**

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -232,11 +232,16 @@ if get_option('enable-python')
   endif
 endif
 
-# nnfw ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
-if get_option('enable-nnfw')
-  add_project_arguments('-DENABLE_NNFW=1', language: ['c', 'cpp'])
-  # For tf-lite/nnapi, enable-nnfw allows to use nnfw as a backend.
-  # This also enables nnfw as a subplugin for Tizen.
+# nnfw-runtime ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
+if get_option('enable-nnfw-runtime')
+  add_project_arguments('-DENABLE_NNFW_RUNTIME=1', language: ['c', 'cpp'])
+  # This enables nnfw as a subplugin for Tizen.
+endif
+
+# nnfw-runtime ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
+if get_option('enable-tflite-nnapi-delegation')
+  add_project_arguments('-DENABLE_TFLITE_NNAPI_DELEGATION=1', language: ['c', 'cpp'])
+  # For tf-lite/nnapi, enable-nnfw allows to use nnfw::tflite::nnapi as a backend.
 endif
 
 # Build nnstreamer (common, plugins)
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -17,6 +17,6 @@ option('enable-python', type: 'boolean', value: true)
 option('enable-tizen', type: 'boolean', value: false)
 option('enable-element-restriction', type: 'boolean', value: false) # true to restrict gst-elements in api
 option('restricted-elements', type: 'string', value: '')
-option('enable-nnfw', type: 'boolean', value: false)
-option('enable-nnfw-runtime', type: 'boolean', value: false)
+option('enable-tflite-nnapi-delegation', type: 'boolean', value: false) # true to enable tensorflow-lite to delegate nnapi interpretation to nnfw backend in tizen
+option('enable-nnfw-runtime', type: 'boolean', value: false) # true to enable nnfw tensor filter element
 option('enable-cppfilter', type: 'boolean', value: true)
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -74,10 +74,10 @@ if get_option('enable-tensorflow-lite')
 
   nnstreamer_filter_tflite_deps = [tflite_dep, thread_dep, libdl_dep, glib_dep, gst_dep, nnstreamer_dep]
 
-  if get_option('enable-nnfw')
-    nnapi_lib = dependency('nnfw', required: true)
-    if nnapi_lib.found()
-      nnstreamer_filter_tflite_deps += nnapi_lib
+  if get_option('enable-tflite-nnapi-delegation')
+    nnfw_nnapi_lib = dependency('nnfw', required: true)
+    if nnfw_nnapi_lib.found()
+      nnstreamer_filter_tflite_deps += nnfw_nnapi_lib
     else
       error('Cannot find nnfw')
     endif
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
@@ -142,7 +142,7 @@ TFLiteCore::loadModel ()
 
     interpreter->UseNNAPI(use_nnapi);
 
-#ifdef ENABLE_NNFW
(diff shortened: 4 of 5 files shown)


myungjoo:
Question:
Does the current Tizen support enable-nnfw as well? (we don't enable it in nnstreamer.spec, yet; thus, we are not testing it.)

kparichay:
> Question:
> Does the current Tizen support enable-nnfw as well? (we don't enable it in nnstreamer.spec, yet; thus, we are not testing it.)

`enable-nnfw` in this meson.build is coming from meson_options.txt, which enables the use of nnapi in tflite. Is `enable-nnfw` of nnstreamer.spec related to this?

leemgs:
To @kparichay 

FYI, the existing issue temporarily results form the storage shortage of the CI server.

meson.build
```diff
@@ -232,11 +232,15 @@ if get_option('enable-python')
...
+if get_option('enable-nnfw-runtime')
+  add_project_arguments('-DENABLE_NNFW_RUNTIME=1', language: ['c', 'cpp'])
+  # This also enables nnfw as a subplugin for Tizen.
+endif
+
 if get_option('enable-nnfw')
   add_project_arguments('-DENABLE_NNFW=1', language: ['c', 'cpp'])
-  # For tf-lite/nnapi, enable-nnfw allows to use nnfw as a backend.
-  # This also enables nnfw as a subplugin for Tizen.
+  # For tf-lite/nnapi, enable-nnfw allows to use nnfw::tflite::nnapi as a backend.
```
myungjoo: Uh... both options are referring to https://review.tizen.org/gerrit/p/platform/core/ml/nnfw
Check its packages (sub-packages) and see if which is for NNAPI and which is for an independent runtime. It is possible that a single package has both roles.

kparichay:
> To @kparichay
> 
> FYI, the existing issue temporarily results form the storage shortage of the CI server.

Thank you for the help.

meson.build
kparichay: Yes, you are right. Both options are referring to https://review.tizen.org/gerrit/p/platform/core/ml/nnfw
However, as we discussed, one option can be set to enable nnfw as backend in tf-lite filter (now `ENABLE_NNFW_TFLITE`), and the other option can be set to enable nnfw as a tensor filter element (`ENABLE_NNFW_RUNTIME`).

kparichay: Please suggest better naming if possible.

meson.build
```diff
@@ -232,11 +232,16 @@ if get_option('enable-python')
...
 endif
 
-# nnfw ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
-if get_option('enable-nnfw')
-  add_project_arguments('-DENABLE_NNFW=1', language: ['c', 'cpp'])
-  # For tf-lite/nnapi, enable-nnfw allows to use nnfw as a backend.
-  # This also enables nnfw as a subplugin for Tizen.
+# nnfw-runtime ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
+if get_option('enable-nnfw-runtime')
+  add_project_arguments('-DENABLE_NNFW_RUNTIME=1', language: ['c', 'cpp'])
```
dongju-chae: This option, `ENABLE_NNFW_RUNTIME`, seems not used yet. Is it reserved for future usage? Then, I think it's better to mention it somewhere.

meson.build
myungjoo: If we do not or cannot specify backend of NNAPI, it should be enable-tensorflow-lite-nnapi (if tf-lite is the only user) or enable-nnapi (if others are going to use it).

If we can and do specify backend of NNAPI, we need something like nnapi-backend=(string) [ nnfw-runtime, armcl, none / disabled, ... ]

meson.build
myungjoo: It is to use Tizen /platform/core/ml/nnfw as yet another tensor-filter subplugin.

kparichay: > This option, `ENABLE_NNFW_RUNTIME`, seems not used yet. Is it reserved for future use? Then, I think it's better to mention it somewhere.

Adding this option creates confusion between using NNFW a tensor-filter subplugin, and NNFW as a backend for tensorflow-lite. Hence, this option is added here, and existing ENABLE_NNFW renamed, to lower burden on the other PR (#1856)

meson.build
kparichay: I cannot find an option to specify backend of NNAPI as of now. 
Currently, tf-lite is the only user in nnstreamer, but caffe2 also supports NNAPI. So, `enable-nnapi` is better, but this option will just enable nnapi and already exists in [nnstreamer.ini.in](https://github.com/nnsuite/nnstreamer/blob/ea2ddd13bcaabc9ed38cf68bb846cb5f69a0bb49/nnstreamer.ini.in#L16).

However, `ENABLE_NNFW_TFLITE` is supposed to enable nnapi with nnfw backend.

myungjoo: If we cannot specify backend, then, let's settle with ENABLE_TFLITE_NNAPI. In such cases, nnfw is meaningless.

Anyway, please note that we need to confirm that nnfw(NNAPI-backend) is working with tflite when nnapi is enabled in Tizen. (but other any backends available in other platforms).

kparichay: > If we cannot specify backend, then, let's settle with ENABLE_TFLITE_NNAPI. In such cases, specifying nnfw is meaningless.

Yes we cannot specify backend in nnapi. However, when ENABLE_NNFW_TFLITE is set (or as you suggested ENABLE_TFLITE_NNAPI), interpreter provided by nnfw is used instead of nnapi's interpreter. So, because backend cannot be specified in nnapi, different interpreter is used. 
Given this, I think ENABLE_TFLITE_NNAPI naming can be confusing as this macro enables uses nnfw interpreter instead of nnapi interpreter (counter to the name which suggests enabling of nnapi in tflite).

> Anyway, please note that we need to confirm that nnfw(NNAPI-backend) is working with tflite when nnapi is enabled in Tizen. (but other any backends available in other platforms).

yes.

myungjoo: Whether it's using nnapi interpreter or nnfw interpreter, with this macro enabled, it uses tensorflow-lite as .tflite parser and nnfw as tf-lite engine. For users of nnstreamer, does the entity that provides nnapi interpreter matter?

In the code affected by this macro, is there any non-tf-lite, but nnfw related code? It only appears to enable "delegation", which I assume it is trying to delegate "nnapi interpretation".

Assuming that another backend is available and we can use this option (in other words, we can let another backend to interprete nnapi instead of both tf-lite and nnfw), having nnfw in the option name may mislead. If activating nnapi is not affected by this macro, but only whether how interpreter is enabled is affected, the name (both option and macro) may need to reflect it: e.g., ENABLE_TFLITE_NNAPI_DELEGATION or ENABLE_TFLITE_NNAPI_INTERPRETER_DELEGATION or something else)

meson_options.txt
```diff
@@ -17,6 +17,6 @@ option('enable-python', type: 'boolean', value: true)
 option('enable-tizen', type: 'boolean', value: false)
 option('enable-element-restriction', type: 'boolean', value: false) # true to restrict gst-elements in api
 option('restricted-elements', type: 'string', value: '')
-option('enable-nnfw', type: 'boolean', value: false)
-option('enable-nnfw-runtime', type: 'boolean', value: false)
+option('enable-nnfw-tflite', type: 'boolean', value: false) # true to enable nnfw backend with tflite filter
```
myungjoo: Based on the above assumption, it'd be # true to enable tensorflow-lite to delegate nnapi interpretation to its backend; i.e., nnfw backend in Tizen for tensorflow-lite.

kparichay: updated.

meson.build
kparichay: Yes, updated to ENABLE_TFLITE_NNAPI_DELEGATION

PR #1860, opened 2019-11-11
[C-API] make other mediatypes available with appsrc

helloahn:
Currently only 'other/tensor' and 'other/tensors' are available for `appsrc` with C-API.
Since there are many `media types`(video, audio, text, octet) that should be used as a `src` data, and they cannot be used currently, this commit makes them available with a partial restriction.

With this commit, I checked the `video/x-raw` works properly, and others will be tested with other sample apps.

Since there will be restriction, however, we need to discuss this method is desirable.

**Self evaluation:**

### DIFF
--- a/api/capi/include/nnstreamer-capi-private.h
+++ b/api/capi/include/nnstreamer-capi-private.h
@@ -169,6 +169,7 @@ typedef struct _ml_pipeline_element {
   gulong handle_id;
 
   GMutex lock; /**< Lock for internal values */
+  gboolean is_media_stream;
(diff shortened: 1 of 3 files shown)


myungjoo:
Your intention is: if it's not other/tensor(s), we are going to accept the whole binary and let the user construct with proper capsfilter or other elements, right?
Then it looks good to me.

helloahn:
@myungjoo

myungjoo:
@jaeyun-jung @again4you Please consider cherry-picking this to 5.5.

PR #1859, opened 2019-11-08
[Tests/Source/IIO] Fix defects (CWE-637) found from Coverity @open sesame 11/11 16:00

wooksong:
This patch fixes defects related to CWE-637 found from Coverity.

See also: https://github.com/nnsuite/nnstreamer/issues/1682

Coverity-scan: CID-263387, CID-263398, CID-263366, CID-263406,
               CID-263410, CID-263372, CID-263390, CID-263361,
	       CID-263414

Signed-off-by: Wook Song <wook16.song@samsung.com>

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [ ]Failed [*]Skipped

PR #1863, opened 2019-11-12
[Tests/Src/IIO] Fix Coverity defects in a macro that generates tests 

wooksong:
This PR addresses the Coverity defects in GENERATE_TESTS_TO_VERIFY_DATA_WO_TRIGGER, which is a macro for test generation. 

- Fix defects about resource leak found from Coverity
  - These defects are caused by the previous PR #1859 
- Fix TOCTOU and NEGATIVE_RETURNS defects as well

CC: @kparichay


wooksong:
> LGTM.
> 
> Anyway... this becomes too lengthy, making it worthy for static/inline function.

Agreed. I will replace this macro with a static function in another PR soon.

Issue #1143, opened 2019-02-21
[EPIC] Tizen Sensor Framework as Input Source


myungjoo:
The output of Tizen sensor framework may become the input for nnstreamer pipeline.

- [x] Determine the rules of ouput format / usage pattern of TIzen sensor framework for nnstreamer
- [x] Determine the tensor types for Tizen sensor framework streams
- [x] Propose and determine the protocol draft with Tizen sensor framework team.
    - [x] Determine if this could be a subplugin of "tensor_source".
    - [x] Determine if this can be in nnstreamer.git/gst, nnstreamer.git/ext, or in another git repo.
- [x] Write the plugin, "tensor_TizenSensor_source"
- [x] Release!

myungjoo:
To Be Determined:
- Include tensor_src_tizensensor in nnstreamer plugin **or** make it an independent plugin & subpackage (and locate at /ext/)

myungjoo:
Done!

PR #1864, opened 2019-11-12
Tensor-src-tizensensor: add get/set properties @open sesame 11/20 16:29

myungjoo:
This adds get/set property methods for tizen sensor extension plugin.
    
This should follow #1857 and addresses #1143
    
### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -126,6 +126,9 @@ enum
  */
 #define DEFAULT_PROP_SEQUENCE 0
 
+#define _LOCK(obj) g_mutex_lock (&(obj)->lock);
+#define _UNLOCK(obj) g_mutex_unlock (&(obj)->lock);
+
 /**
  * @brief Template for src pad.
  */
@@ -324,6 +327,9 @@ gst_tensor_src_tizensensor_init (GstTensorSrcTIZENSENSOR * self)
   /** init properties */
   self->configured = FALSE;
   self->silent = DEFAULT_PROP_SILENT;
+
+  g_mutex_init (&self->lock);
+
   /**
    * @todo TBD. Update This!
    * format of the source since IIO device as a source is live and operates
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -324,6 +327,9 @@ gst_tensor_src_tizensensor_init (GstTensorSrcTIZENSENSOR * self)
   /** init properties */
   self->configured = FALSE;
   self->silent = DEFAULT_PROP_SILENT;
+
+  g_mutex_init (&self->lock);
```
dongju-chae: Needed `g_mutex_clear()` somewhere ?

myungjoo: At ```gst_tensor_src_tizensensor_finalize``` of #1865 ?

dongju-chae: Oh, right.

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -349,7 +384,122 @@ static void
...
+
+      if (new_mode != self->mode) {
+        silent_debug ("Set mode from %d --> %d.", self->mode, new_mode);
+        self->mode = new_mode;
+
+        /* Same sensor is kept. Only mode is changed */
+        ret = _ts_reconfigure (self);
+
+        if (ret) {
+          GST_ELEMENT_ERROR (self, TIZEN_SENSOR, FAILED,
```
kparichay: in this case of failure of `reconfigure`, can value of `mode` be reset to the original value?

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -349,7 +384,122 @@ static void
...
+
+      self->freq_n = gst_value_get_fraction_numerator (value);
+      self->freq_d = gst_value_get_fraction_denominator (value);
+
+      silent_debug ("Set operating frequency %d/%d --> %d/%d",
+          n, d, self->freq_n, self->freq_d);
+
+      if (n != self->freq_n || d != self->freq_d) {
+        /* Same sensor is kept. Only mode is changed */
+        ret = _ts_reconfigure (self);
```
kparichay: same as above, reset freq values if reconfigure fails (This will keep the `get_property` consistent).

Issue #1852, opened 2019-11-08
[API changes] android c#


jaeyun-jung:
Issues from Mr.Ham, data container for Tensor

We implemented two classes for tensor-data and tensor-info.
Need to set metadata in data class.

1. Add metadata in TensorsData class
2. Define Tensor class and contain both metadata and buffer.

We have to consider the tizen c# SingleShot APIs can be updated for next release.

For Android APIs, see attached javadoc.

1. 2019-11-13 defined Tensor class and integrate all methods
[nnstreamer-api-javadoc_2019-11-13.zip](https://github.com/nnsuite/nnstreamer/files/3839418/nnstreamer-api-javadoc_2019-11-13.zip)

2. 2019-11-22 add method to get metadata in TensorsData
[nnstreamer-api-javadoc_2019-11-22.zip](https://github.com/nnsuite/nnstreamer/files/3884960/nnstreamer-api-javadoc_2019-11-22.zip)

3. add clear() method and handle error case to update metadata in TensorsData
[nnstreamer-api-javadoc_2019-11-27.zip](https://github.com/nnsuite/nnstreamer/files/3895387/nnstreamer-api-javadoc_2019-11-27.zip)

4. constructor in TensorsData, CustomFilter and NNStreamer class
[nnstreamer-api-javadoc_2019-11-28.zip](https://github.com/nnsuite/nnstreamer/files/3900941/nnstreamer-api-javadoc_2019-11-28.zip)

5. create TensorsData with TensorsInfo meta (non-null)
[nnstreamer-api-javadoc_2019-12-04.zip](https://github.com/nnsuite/nnstreamer/files/3925132/nnstreamer-api-javadoc_2019-12-04.zip)

6. register multiple sink callback
[nnstreamer-api-javadoc_2019-12-16.zip](https://github.com/nnsuite/nnstreamer/files/3967714/nnstreamer-api-javadoc_2019-12-16.zip)

7. add SNAP, enumeration for data type and nnfw
[nnstreamer-api-javadoc_2019-12-31.zip](https://github.com/nnsuite/nnstreamer/files/4012076/nnstreamer-api-javadoc_2019-12-31.zip)

8. add clone() method in TensorsInfo class to prevent the meta-info change after allocaing the byte-buffer.
[nnstreamer-api-javadoc_2020-01-02.zip](https://github.com/nnsuite/nnstreamer/files/4018347/nnstreamer-api-javadoc_2020-01-02.zip)

[...]

PR #1866, opened 2019-11-12
[Android] tensors-data class with metadata @open sesame 01/06 17:15

jaeyun-jung:
1. Add tensors info in TensorsData class.
2. Handle error case when failed to create pipe info in native functions.
3. Define enumeration for tensor data type and pipeline state.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
@@ -25,19 +25,26 @@ public class APITestCustomFilter {
 
     private Pipeline.NewDataCallback mSinkCb = new Pipeline.NewDataCallback() {
         @Override
-        public void onNewDataReceived(TensorsData data, TensorsInfo info) {
-            if (data == null || data.getTensorsCount() != 1 ||
-                info == null || info.getTensorsCount() != 1) {
+        public void onNewDataReceived(TensorsData data) {
+            if (data == null || data.getTensorsCount() != 1) {
                 mInvalidState = true;
-            } else {
-                ByteBuffer output = data.getTensorData(0);
+                return;
+            }
 
-                for (int i = 0; i < 10; i++) {
-                    float expected = i + 1.5f;
+            TensorsInfo info = data.getTensorsInfo();
 
-                    if (expected != output.getFloat(i * 4)) {
-                        mInvalidState = true;
-                    }
+            if (info == null || info.getTensorsCount() != 1) {
+                mInvalidState = true;
+                return;
+            }
+
+            ByteBuffer output = data.getTensorData(0);
+
+            for (int i = 0; i < 10; i++) {
+                float expected = i + 1.5f;
+
+                if (expected != output.getFloat(i * 4)) {
+                    mInvalidState = true;
                 }
             }
 
@@ -51,13 +58,13 @@ public class APITestCustomFilter {
             mCustomPassthrough = CustomFilter.registerCustomFilter("custom-passthrough",
                 new CustomFilter.CustomFilterCallback() {
                     @Override
-                    public TensorsInfo getOutputInfo(TensorsInfo inInfo) {
-                        return inInfo;
+                    public TensorsInfo getOutputInfo(TensorsInfo in) {
+                        return in;
                     }
 
                     @Override
-                    public TensorsData invoke(TensorsData inData, TensorsInfo inInfo, TensorsInfo outInfo) {
-                        return inData;
+                    public TensorsData invoke(TensorsData in) {
+                        return in;
                     }
                 });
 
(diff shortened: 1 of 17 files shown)


jaeyun-jung:
Related issue : #1852

Define Tensor class and remove TensorsData and TensorsInfo in Java

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
```diff
@@ -279,13 +278,13 @@ public void testRegisterDuplicatedName() {
...
-                    public TensorsInfo getOutputInfo(TensorsInfo inInfo) {
-                        return inInfo;
+                    public Tensor getOutputInfo(Tensor in) {
+                        return in;
                     }
 
                     @Override
-                    public TensorsData invoke(TensorsData inData, TensorsInfo inInfo, TensorsInfo outInfo) {
-                        return inData;
+                    public Tensor invoke(Tensor in) {
```
leemgs: It seems that the changes are good to improve readability.

api/android/api/src/main/java/org/nnsuite/nnstreamer/Tensor.java
```diff
@@ -71,118 +89,204 @@ public void addTensorInfo(int type, @NonNull int[] dimension) {
...
+
+        ByteBuffer buffer = allocateByteBuffer(data.length);
+        buffer.put(data);
+
+        setData(index, buffer);
+    }
+
+    /**
+     * Sets a tensor data.
+     *
```
leemgs: Can not we use @brief for the setData function?

jaeyun-jung: To upload library with javadoc (jcenter), brief tag is not supported.

again4you:
What about adding `Clear()` method into `TensorsData`?

jaeyun-jung:
PR updated
2019-11-27 TensorsData class, add clear() method and handle error case to set new metadata.

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
jaeyun-jung: PR changed (TensorsData class with metadata).

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
```diff
@@ -25,19 +25,26 @@
...
-                for (int i = 0; i < 10; i++) {
-                    float expected = i + 1.5f;
+            TensorsInfo info = data.getTensorsInfo();
 
-                    if (expected != output.getFloat(i * 4)) {
-                        mInvalidState = true;
-                    }
+            if (info == null || info.getTensorsCount() != 1) {
+                mInvalidState = true;
+                return;
```
leemgs: look good.

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
```diff
@@ -25,19 +25,26 @@
...
         @Override
-        public void onNewDataReceived(TensorsData data, TensorsInfo info) {
-            if (data == null || data.getTensorsCount() != 1 ||
-                info == null || info.getTensorsCount() != 1) {
+        public void onNewDataReceived(TensorsData data) {
+            if (data == null || data.getTensorsCount() != 1) {
                 mInvalidState = true;
-            } else {
-                ByteBuffer output = data.getTensorData(0);
+                return;
```
leemgs: :accept:

api/android/api/src/main/java/org/nnsuite/nnstreamer/TensorsData.java
```diff
@@ -33,63 +34,74 @@
      * @param size The byte size of the buffer
      *
      * @return The new byte buffer
+     *
+     * @throws IllegalArgumentException if given size is invalid
      */
     public static ByteBuffer allocateByteBuffer(int size) {
```
kparichay: Is this API really needed as public? I think its better to just allocate allocate on the basis of tensorInfo. Also the returned buffer has no association with the TensorsData object..

jaeyun-jung: For the convenience to pass the buffer to native code. ByteBuffer in java is big-endian and non-direct buffer. It cannot be passed and used in native code.

api/android/api/src/main/java/org/nnsuite/nnstreamer/TensorsInfo.java
```diff
@@ -57,7 +64,7 @@ public int getTensorsCount() {
      * @throws IndexOutOfBoundsException when the maximum number of tensors in the list
      * @throws IllegalArgumentException if given param is null or invalid
      */
-    public void addTensorInfo(int type, @NonNull int[] dimension) {
+    public void addTensorInfo(NNStreamer.TensorType type, @NonNull int[] dimension) {
```
kparichay: With the current implementation, `addTensorInfo` can make make a TensorData inconsistent, as the TensorInfo inside the TensorData can be modified without modifying the memory allocated by TensorData. Should that be synced (in case its already synced), or its responsibility of the caller? This is because addTensorData keeps TensorData and TensorInfo synced.
The cases where TensorData and TensorInfo are not synced (TensorData allocaed without TensorInfo) are ignored.

api/android/api/src/main/java/org/nnsuite/nnstreamer/TensorsData.java
kparichay: got it, thanks for the explanation.

api/android/api/src/main/java/org/nnsuite/nnstreamer/TensorsInfo.java
jaeyun-jung: I supposed that developer should handle proper data conversion between TensorsInfo and TensorsData class. (no-sync after allocating data from TensorsInfo object)

jaeyun-jung: Or, is it necessary to implement cloneable interface?

jaeyun-jung: Implemented clone() method in TensorsInfo class to prevent the meta-info change after allocaing the byte-buffer.

again4you:
Basically LGTM.

BTW, what about using `PipelineState` instead of `Pipeline.State`?
Please refer my work and let me know your opinion. (ACR is not started yet)
* https://github.com/again4you/TizenFX/commit/146ab6f73555262c89bda319dfaa11a9c053676b#diff-4dbc211d01533f0345e5002ebe3054bdR96

jaeyun-jung:
@again4you About the enum of pipeline state,
1. current SingleShot API does not use the pipeline.
2. 'pipeline state' only used in Pipeline APIs.
Because of these, I defined the enum in Pipeline class.

jaeyun-jung:
Discussed with @again4you 
merge this PR, and check the changes after the ACR

PR #1870, opened 2019-11-14
[C-Api] fix invalid sink handle @open sesame 11/15 15:55

jaeyun-jung:
When trying to register sink callbacks to same sink element, it returns null handler.
To fix this, allocate new handle and add it  into the handle list.


myungjoo:
@jaeyun-jung @again4you Do we need to cherry-pick this commit to "tizen 5.5" branch in tizen.org? (even if we do not send SR, we may get issues from partners soon)

jaeyun-jung:
@again4you @helloahn @myungjoo 
need to cherry-pick two PRs: #1860 #1870

Issue #1808, opened 2019-10-21
[other/tensor] Storing the rank of a tensor in tensor_dim


kparichay:
Current setup of using the `tensor_dim dimension` in `other/tensor` does not store the information of rank inside the `dimension` array. For example - two tensors of shape [1,1,10,100] and [1,10,100] are represented in the same manner in the dimension array. Current style sets the extra dimensions to `1`. Although this makes it easier to handle, however the information on actual rank of the tensor is lost.

I propose to set the extra dimensions to `0`. This style allows to implement all the features feasible with the current style, along with storing the information of rank. This enables the following-
- Tensors of same size and similar shape can be distinguished (example - `[1,4]` and `[4]`)
- User set rank can be stored and maintained (current implementation does not allow this)
  - This would allow to directly use the user provided shape as an input for framework (currently, the only way is to try all feasible ranks for frameworks to check which works)

I think the proposed style also brings the use of dimensions more closer to python numpy/tensoflow style, where a dimensions can be squeezed/expanded.

myungjoo:
For clarification:

Current: ```640:480```(C: [1][1][480][640]) == ```640:480:1:1```(C: [1][1][480][640]) . ```640:480:0:0``` == INVALID

Your proposal: ```640:480```(C: [1][1][480][640]) == ```640:480:0:0```(C: [1][1][480][640])  && ```640:480```(C: [1][1][480][640])  != ```640:480:1:1```(C: [1][1][480][640]) .



Additional Considerations:
- Backward compatibility
- Tensor path manipulators (e.g., merge/mux/split/...). 
- GetRank function.

kparichay:
> Additional Considerations:
> 
> * Backward compatibility

Is backward compatibility possible? Current design considers any dimension with value `0` as invalid, but proposed design considers it valid. 
This being said, there can be a helper function to convert a given tensor dimension from proposed design to the current implementation design (the other way around is not possible). However, I am not sure if this counts as backward compatibility.

myungjoo:
Backward compatibility for conventional nnstreamer applications; we will have a few nnstreamer applications in Tizen soon, with the version 1.0.0, which does not have this feature. We cannot enforce them to update apps after the release of next nnstreamer versions.

In other words, the conventional nnstreamer pipelines with tf-lite and custom filters (who didn't need to read any rank numbers) with the conventional notion should work.

We don't need backward compatibility for our own code (nnstreamer elements in nnstreamer.git). We just need to update them all.

[...]

PR #1878, opened 2019-11-18
[CodeClean] cleanup duplicated and update testcases

jaeyun-jung:
1. remove duplicated condition and fix invalid comment
2. add tc to check tensor info

### DIFF
--- a/tools/development/nnstreamerCodeGenCustomFilter.py
+++ b/tools/development/nnstreamerCodeGenCustomFilter.py
@@ -132,8 +132,12 @@ cg_getInputDim (void * _data, const GstTensorFilterProperties * prop,
   in_info->info[0].dimension[1] = 224; /** @todo MODIFY THIS! */
   in_info->info[0].dimension[2] = 224; /** @todo MODIFY THIS! */
 
+  /**
+   * Check max dimension (NNS_TENSOR_RANK_LIMIT) and
+   * fill 1 to uninitialized dimension values.
+   */
   for (i = 3; i < NNS_TENSOR_RANK_LIMIT; i++)
-    in_info->info[0].dimension[i] = 1; /** Fill 1 to uninitialized dimension values */
+    in_info->info[0].dimension[i] = 1;
 
   return 0;
 }}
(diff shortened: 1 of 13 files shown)


jaeyun-jung:
Related Issue: #1808

myungjoo:
I guess there are a lot of documents (.md files) to be updated.

tools/development/nnstreamerCodeGenCustomFilter.py
```diff
@@ -128,12 +128,15 @@ def getinput(text):
   in_info->num_tensors = 1; /** @todo MODIFY THIS! */
   in_info->info[0].name = NULL; /** Optional, default is null. Set new memory for tensor name string. */
   in_info->info[0].type = _NNS_UINT8; /** @todo MODIFY THIS! */
+
+  /**
+   * Check max dimension (NNS_TENSOR_RANK_LIMIT) and
+   * fill 0 to uninitialized dimension values.
```
dongju-chae: Who has the responsibility to do things described in this comment?

jaeyun-jung: If you mean the todo tag, it is developer to set valid tensor metadata.

dongju-chae: Um, what I meant is ... the previous codes initialized remaining dimension slots to 1 (up to `NNS_TENSOR_RANK_LIMIT`).
Now, new codes set dimension values only up to [3]. So, I was confused because I first thought that this comment describes the below codes. 
But, you meant that the duty of filling values for remaining slots is now moved to developers, right? Then, it's fine.

myungjoo:
The code looks good.

But, I have a concern:
```
tensor_ELEMENT_A (out = 3:1) ! (in = 3:1:1:1) tensor_ELEMENT_B
```
How do we keep it working? or do you think we should let it generate error?

jaeyun-jung:
@myungjoo This PR generates error when given tensor dimension is not exactly matched.

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -97,22 +97,45 @@ ml_tensors_info_initialize (ml_tensors_info_s * info)
...
 
+/**
+ * @brief Gets the tensor rank.
+ */
+static guint
+ml_tensor_info_get_rank (const ml_tensor_info_s * info)
+{
+  guint i;
+
+  if (!info)
```
wooksong: Did we change the error checking code following this style to ```info == NULL```, didn't we?

jaeyun-jung: updated.
TODO need to fix coding style later

jaeyun-jung:
@myungjoo I tried to support compatible dimension case, but I could not.

We need to consider adding another plugin for this compatible dim case.
tensor-transform only supports single tensor, the case of multi-tensors it is impossible now.

jaeyun-jung:
PR changed, this is not related with the rank.

PR #1880, opened 2019-11-19
[CAPI] By default a new nnfw is "not available" @open sesame 11/20 16:34

myungjoo:
Until we implement its supporting code, assume a new
nnfw is NOT SUPPORTED.

This makes sure that it does not reply "supported" for
arbitrary new NNFWs.

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -933,6 +933,9 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
   *available = false;
 
   switch (nnfw) {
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -960,6 +963,7 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
       }
       break;
     default:
+      goto done; /* Default = "Not available!" */
       break;
```
dongju-chae: This line will be never reached due to the goto statement.

myungjoo: I'm not sure how SVACE/Coverity will react to this, but yes, let's remove break.

Issue #1817, opened 2019-10-23
Add tensor_filter::armnn


myungjoo:
Let's support armnn.

Note that we ALREADY have armnn in tizen.org

kparichay:
- [ ] Port FlatBufferst to tizen
- [ ] Update armcl and armnn to 19_08 version in tizen
  - [ ] Add pkgconfig for armnn 
- [ ] Add support for armnn tensor filter
  - [ ] Upgrade nnstreamer to c++14

myungjoo:
Close this when you have done. (including test cases & CAPI enum handling)

PR #1885, opened 2019-11-20
[spec/meson] Update nnstreamer package to support c++14 language standard @open sesame 11/21 14:11

kparichay:
Update nnstreamer package to support c++14 language standard
This is primarily needed to support armnn tensor filter subplugin
armnn headers use c++14 compliant code

**Self evaluation:**

More info in #1817

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,8 +2,8 @@ Source: nnstreamer
 Section: libs
 Priority: optional
 Maintainer: MyungJoo Ham <myungjoo.ham@samsung.com>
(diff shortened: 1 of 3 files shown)


kparichay:
Ubuntu failure is most likely because of gcc bug with version 5.2.
I have tested with local build with gcc version 5.4 and it works.
On my local system: `pdebuild` (using gcc 5.2) fails while `debuild` (using gcc 5.4) succeeds.

leemgs:
https://github.com/nnsuite/TAOS-CI/pull/586 solves the incorrect operation of the coverity module.

Issue #1886, opened 2019-11-20
Incorrect Typing: nnstreamer_plugin_api.h/c


myungjoo:
gst_tensor_get_element_count (const tensor_dim dim); 
should be gsize --> guint

gst_tensor_get_element_size (tensor_type type); 
should be guint --> gsize

jaeyun-jung:
gulong gst_tensor_get_element_count()
gsize gst_tensor_get_element_size()

jaeyun-jung:
I will change this with tensor rank changes.

PR #1887, opened 2019-11-20
[Common] change return type

jaeyun-jung:
change improper return type to get tensor element size/count.


jaeyun-jung:
Related issue: #1886

PR #1891, opened 2019-11-21
[tests/coverity] Fix coverity issues related to gtest main

kparichay:
Add try/catch for exceptions thrown by function calls in main bosy when using gtest

**Self evaluation:**

### DIFF
--- a/tests/common/unittest_common.cpp
+++ b/tests/common/unittest_common.cpp
@@ -760,6 +760,18 @@ TEST (conf_custom, env_str_01)
 int
 main (int argc, char **argv)
 {
-  testing::InitGoogleTest (&argc, argv);
-  return RUN_ALL_TESTS ();
+  int ret = -1;
+  try {
+    testing::InitGoogleTest (&argc, argv);
+  } catch (...) {
+    g_warning ("catch 'testing::internal::<unnamed>::ClassUniqueToAlwaysTrue'");
+  }
+
+  try {
+    ret = RUN_ALL_TESTS ();
+  } catch (...) {
+    g_warning ("catch `testing::internal::GoogleTestFailureException`");
+  }
+
+  return ret;
 }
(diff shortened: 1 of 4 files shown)


tests/common/unittest_common.cpp
```diff
@@ -760,6 +760,18 @@ TEST (conf_custom, env_str_01)
 int
 main (int argc, char **argv)
 {
-  testing::InitGoogleTest (&argc, argv);
-  return RUN_ALL_TESTS ();
+  int ret = 0;
+  try {
+    testing::InitGoogleTest (&argc, argv);
+  } catch (...) {
+    g_warning ("catch 'testing::internal::<unnamed>::ClassUniqueToAlwaysTrue'");
```
dongju-chae: Is `...ClassUniqueToAlwaysTrue` the unique exception that can be thrown during `testing::InitGoogleTest()`?

tests/common/unittest_common.cpp
```diff
@@ -760,6 +760,18 @@ TEST (conf_custom, env_str_01)
...
+  }
+
+  try {
+    ret = RUN_ALL_TESTS ();
+  } catch (...) {
+    g_warning ("catch `testing::internal::GoogleTestFailureException`");
```
dongju-chae: IMO, when exception is caught, `ret` needs to denote an error like a negative value.

tests/common/unittest_common.cpp
kparichay: Yes.
Also this exception is never thrown actually - https://github.com/google/googletest/blob/master/googletest/src/gtest.cc#L5711

tests/common/unittest_common.cpp
kparichay: Updated. `ret` is now initialized with -1, this will take care if exception is caught.

leemgs:
https://github.com/nnsuite/TAOS-CI/pull/586 solves the incorrect operation of the coverity module.

PR #1897, opened 2019-11-26
Tensor-src-tizensensor: support custom sensors

myungjoo:
New sensors may be define in runtime.
Because their metadata is unknown at build-time, it should be flexible.


myungjoo:
Looks like custom sensors are obsolete. Let's drop this.

PR #1907, opened 2019-12-02
[PROP] Add new tensor filter prop to indicate a model is updatable - @open sesame 12/03 19:23

dongju-chae:
This PR adds new tensor filter property to indicate a given model
to a tensor filter is updatable or not. The default value is False.

Related issue: https://github.com/nnsuite/nnstreamer/issues/1900

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -65,7 +65,8 @@ enum
   PROP_OUTPUTNAME,
   PROP_CUSTOM,
   PROP_SUBPLUGINS,
-  PROP_ACCELERATOR
+  PROP_ACCELERATOR,
+  PROP_UPDATABLE_MODEL,
 };
 
 /**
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -321,6 +322,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
           "list of accelerators determines the backend (ignored with false). "
           "Example, if GPU, NPU can be used but not CPU - true:(GPU,NPU,!CPU).",
           "", G_PARAM_READWRITE));
+  g_object_class_install_property (gobject_class, PROP_UPDATEMODEL,
+      g_param_spec_string ("updatemodel", "Updatable model",
```
jaeyun-jung: how about changing property name: `is- updatable`

dongju-chae: Looks good. thx.

Issue #1900, opened 2019-11-28
Allow to update models of tensor_filter in run-time.


myungjoo:
1. Allow to update models of tensor_filter in run-time (1st; tf-lite, custom / 2nd: nnfw / 3rd: movidius / 4th: others)
2. Supply corresponding test cases as well.

PR #1908, opened 2019-12-03
[TensorFilter] Add a callback of tensor filter for model update in runtime

dongju-chae:
This PR adds a callback, `reloadModel()`, of tensor filter for model
update in runtime. When the model property is updated, this callback
will be called if `is-updatable` is TRUE.

Also, it adds an event handler to update a model in the same pipeline.
On receiving a model update event, it calls `set_property()` of `PROP_MODEL`.

Related issue: #1900

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -202,6 +202,15 @@ typedef struct _GstTensorFilterFramework
        *
        * @param[in] data the data element.
        */
+
+  int (*reloadModel) (const GstTensorFilterProperties * prop, void **private_data);
+      /**< Optional. tensor_filter.c will call it when a model property is newly configured.
+       * @todo add more detail comments.
+       *
+       * @param[in] prop read-only property values
+       * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, close() frees private_data and set NULL.
+       * @return 0 if ok. < 0 if error.
+       */
 } GstTensorFilterFramework;
 
 /* extern functions for subplugin management, exist in tensor_filter.c */
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -58,6 +58,9 @@
 
 #include "tensor_filter.h"
 
+/** @todo rename & move this to better location */
+#define EVENT_NAME_UPDATE_MODEL "evt_update_model"
+
 /**
  * @brief Macro for debug mode.
  */
@@ -152,6 +155,8 @@ static gboolean gst_tensor_filter_transform_size (GstBaseTransform * trans,
     GstCaps * othercaps, gsize * othersize);
 static gboolean gst_tensor_filter_start (GstBaseTransform * trans);
 static gboolean gst_tensor_filter_stop (GstBaseTransform * trans);
+static gboolean gst_tensor_filter_sink_event (GstBaseTransform * trans,
+    GstEvent * event);
 
 /**
  * @brief Invoke callbacks of nn framework. Guarantees calling open for the first call.
@@ -213,6 +218,9 @@ gst_tensor_filter_class_init (GstTensorFilterClass * klass)
   /* Allocation units */
   trans_class->transform_size =
       GST_DEBUG_FUNCPTR (gst_tensor_filter_transform_size);
+
+  /* setup sink event */
+  trans_class->sink_event = GST_DEBUG_FUNCPTR (gst_tensor_filter_sink_event);
 
   /* start/stop to call open/close */
   trans_class->start = GST_DEBUG_FUNCPTR (gst_tensor_filter_start);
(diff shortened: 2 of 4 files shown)


gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -865,6 +873,54 @@ gst_tensor_filter_transform_size (GstBaseTransform * trans,
...
+      const GstStructure *structure = gst_event_get_structure (event);
+      int ret = -1;
+
+      /** @todo who is supposed to send this event? and how? */
+      if (structure == NULL ||
+          !gst_structure_has_name (structure, EVENT_NAME_UPDATE_MODEL))
+        break;
+
+      if (priv->updatable_model) {
+        const GValue *value = gst_structure_get_value (structure, "model");
```
myungjoo: If the filter subplugin does not support update (updateModel is NULL), you will get ret = -1, which makes an error here.

If a filter subplugin does not support, we should not accept set-property, "updatableModel".

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -214,6 +219,9 @@ gst_tensor_filter_class_init (GstTensorFilterClass * klass)
   trans_class->transform_size =
       GST_DEBUG_FUNCPTR (gst_tensor_filter_transform_size);
 
+  /* setup sink event */
+  trans_class->sink_event = GST_DEBUG_FUNCPTR (gst_tensor_filter_sink_event);
```
myungjoo: What's the rationale of using sink_event for "update model"?

dongju-chae: Well, in order to update model data with some way, I thought that a tensor filter needs to get a related event from somewhere. Actually, I have no idea yet how to send such events to a tensor filter. But, I think an easy way would be to use `gst_element_send_event()` for the filter if we implement gst app.

Anyway, AFAIK, `.send_event()` is called for pads such as sink/src depending on an event's direction. Actually, there was no rationale for why I chose `sink` instead of `src`.

gst/nnstreamer/tensor_filter/tensor_filter.c
dongju-chae: True, fixed.

gst/nnstreamer/tensor_filter/tensor_filter.c
myungjoo: Would it be possible and easy for an app (an owner of pipeline) to invoke this event?

dongju-chae: Well, I'm not sure whether I understood your question correctly, but I'll describe my thought anyway. 

I think pipeline owners (i.e., gst app developers) can easily invoke this event because they can easily access the pointer of each element, and directly call `gst_element_send_event()`. In this case, on-device training may exist as a separated thread or process.

Another scenario is that on-device training co-exists in the same pipeline like the below figure.
image
In this case, on-device training plugin needs to broadcast events to pipeline, and I'm not sure a tensor filter can receive events from a diverged branch like the below figure.
image

In fact, I have no clear working model yet, and we need to discuss more how model-updatable tensor filters and model-updater plugins (i.e., on-device training) can work together in nnstreamer.

myungjoo: We need a small discussion with @jijoonmoon for this.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -202,6 +202,17 @@ typedef struct _GstTensorFilterFramework
        *
        * @param[in] data the data element.
        */
+
+  int (*updateModel) (const GstTensorFilterProperties * prop,
+      void **private_data, const char * model_path);
```
myungjoo: prop already has model info.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -322,7 +323,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
           "list of accelerators determines the backend (ignored with false). "
           "Example, if GPU, NPU can be used but not CPU - true:(GPU,NPU,!CPU).",
           "", G_PARAM_READWRITE));
-  g_object_class_install_property (gobject_class, PROP_UPDATABLE_MODEL,
+  g_object_class_install_property (gobject_class, PROP_UPDATE_MODEL,
```
myungjoo: PROP_MODEL should do this.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
dongju-chae: Yes, I tried to reuse `PROP_MODEL` but it seems that its set_property is designed to handle model reopen. So, I wonder how we can distinguish between reopen vs. update_model

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -202,6 +202,18 @@ typedef struct _GstTensorFilterFramework
...
+      void **private_data, const char * model_path, const char * model_path_sub);
+      /**< Optional. tensor_filter.c will call it when it gets a related event.
+       * @todo add more detail comments.
+       *
+       * @param[in] prop read-only property values
+       * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, close() frees private_data and set NULL.
+       * @param[in] model_path model path to update in runtime
+       * @param[in] model_path_sub sub model path to update in runtime
+       * @return 0 if ok. < 0 if error.
+       */
```
myungjoo: So.. maybe..
```reloadModel (const GstTensorFilterProperties * prop, void **private_data);``` should be enough for this capability.```

File: gst/nnstreamer/tensor_filter/tensor_filter_common.c [hunk 6]
```diff
@@ -129,13 +129,14 @@ nnstreamer_filter_find (const char *name)
...
  * @param[in] model_files the prediction model paths
+ * @param[out] model_file the parsed model_file
+ * @param[out] model_file_sub the parsed sub model_file
  * @return number of parsed model path
  */
 static guint
-gst_tensor_filter_parse_modelpaths_string (GstTensorFilterProperties * prop,
-    const gchar * model_files)
+gst_tensor_filter_parse_modelpaths_string (const gchar * model_files,
+    gchar ** model_file, gchar ** model_file_sub)
```
myungjoo: By using "reload" and "setProperties", you won't need to modify such.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
dongju-chae: Got it. It'll be updated.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
dongju-chae: Got it.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -202,6 +202,15 @@ typedef struct _GstTensorFilterFramework
        *
        * @param[in] data the data element.
        */
+
+  int (*reloadModel) (const GstTensorFilterProperties * prop, void **private_data);
+      /**< Optional. tensor_filter.c will call it when a model property is newly configured.
```
leemgs: Do you need "<" ?

PR #1910, opened 2019-12-04
[C-Api/Single] check invalid tensor info @open sesame 12/05 15:11

jaeyun-jung:
Check input/output tensor info when opening the nn model.
If given info is not matched with model, try to change input info for dynamic mode.

### DIFF
--- a/api/capi/include/nnstreamer-capi-private.h
+++ b/api/capi/include/nnstreamer-capi-private.h
@@ -246,9 +246,13 @@ typedef struct _ml_pipeline_valve {
 
 /**
  * @brief Macro to check the tensors info is valid.
- * @since_tizen 5.5
  */
-#define ml_tensors_info_is_valid(i,v) (ml_tensors_info_validate ((i), &(v)) == ML_ERROR_NONE && (v))
+#define ml_tensors_info_is_valid(i) ({bool v; (ml_tensors_info_validate ((i), &v) == ML_ERROR_NONE && v);})
+
+/**
+ * @brief Macro to compare the tensors info.
+ */
+#define ml_tensors_info_is_equal(i1,i2) ({bool e; (ml_tensors_info_compare ((i1), (i2), &e) == ML_ERROR_NONE && e);})
 
 /**
  * @brief Gets the byte size of the given tensor info.
@@ -264,6 +268,20 @@ size_t ml_tensor_info_get_size (const ml_tensor_info_s *info);
  * @retval #ML_ERROR_INVALID_PARAMETER Given parameter is invalid.
  */
 int ml_tensors_info_initialize (ml_tensors_info_s *info);
+
+/**
+ * @brief Compares the given tensors information.
+ * @details If the function returns an error, @a equal is not changed.
+ * @since_tizen 6.0
+ * @param[in] info1 The handle of tensors information to be compared.
+ * @param[in] info2 The handle of tensors information to be compared.
+ * @param[out] equal @c true if given tensors information is equal, @c false if if it's not equal.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported.
+ * @retval #ML_ERROR_INVALID_PARAMETER Given parameter is invalid.
+ */
+int ml_tensors_info_compare (const ml_tensors_info_h info1, const ml_tensors_info_h info2, bool *equal);
 
 /**
  * @brief Frees the tensors info pointer.
@@ -289,6 +307,7 @@ GstCaps * ml_tensors_info_get_caps (const ml_tensors_info_s *info);
 
 /**
  * @brief Creates a tensor data frame wihout buffer with the given tensors information.
+ * @details If @a info is null, this allocates data handle with empty tensor data.
  * @param[in] info The handle of tensors information for the allocation.
  * @param[out] data The handle of tensors data.
  * @return @c 0 on success. Otherwise a negative error value.
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -322,9 +322,10 @@ ml_single_set_info_in_handle (ml_single_h single, gboolean is_input,
     ml_tensors_info_s * tensors_info)
 {
   ml_single *single_h;
+  ml_tensors_info_h info = NULL;
   ml_tensors_info_s *dest;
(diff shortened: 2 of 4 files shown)


api/capi/include/nnstreamer-capi-private.h
```diff
@@ -250,6 +250,11 @@ typedef struct _ml_pipeline_valve {
  */
 #define ml_tensors_info_is_valid(i,v) (ml_tensors_info_validate ((i), &(v)) == ML_ERROR_NONE && (v))
 
+/**
+ * @brief Macro to compare the tensors info.
+ */
+#define ml_tensors_info_is_equal(i1,i2,e) (ml_tensors_info_compare ((i1), (i2), &(e)) == ML_ERROR_NONE && (e))
```
myungjoo: What about
```
#define ml_tensors_info_is_equal(i1,i2) ({bool e; (ml_tensors_info_compare ((i1), (i2), &e) == ML_ERROR_NONE && e)})
```
?

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -343,18 +345,24 @@ ml_single_set_info_in_handle (ml_single_h single, gboolean is_input,
...
-      ml_single_set_inout_tensors_info (filter_obj, is_input, tensors_info);
-    ml_tensors_info_clone (dest, tensors_info);
-  } else {
-    ml_tensors_info_h info;
-
+  if (configured) {
+    /* get configured info and compare with input info */
     ml_single_get_tensors_info_from_filter (single_h->filter, is_input, &info);
+
+    if (tensors_info && !ml_tensors_info_is_equal (tensors_info, info, equal)) {
```
myungjoo: For correctness you need to update the macro. the return of "is_equal" is not "is_equal'.

api/capi/include/nnstreamer-capi-private.h
jaeyun-jung: updated

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -343,19 +343,25 @@ ml_single_set_info_in_handle (ml_single_h single, gboolean is_input,
...
-  } else {
-    ml_tensors_info_h info;
-
+  if (configured) {
+    /* get configured info and compare with input info */
     ml_single_get_tensors_info_from_filter (single_h->filter, is_input, &info);
+
+    if (tensors_info && !ml_tensors_info_is_equal (tensors_info, info)) {
+      /* given input info is not matched with configured */
+      goto done;
```
myungjoo: Is this ok to return TRUE with this? (it appears to be able to do so..)

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -499,10 +556,13 @@ ml_tensors_data_create_no_alloc (const ml_tensors_info_h info,
...
-  for (i = 0; i < _data->num_tensors; i++) {
-    _data->tensors[i].size = ml_tensor_info_get_size (&tensors_info->info[i]);
-    _data->tensors[i].tensor = NULL;
+  _info = (ml_tensors_info_s *) info;
+  if (_info != NULL) {
+    _data->num_tensors = _info->num_tensors;
+    for (i = 0; i < _data->num_tensors; i++) {
+      _data->tensors[i].size = ml_tensor_info_get_size (&_info->info[i]);
+      _data->tensors[i].tensor = NULL;
+    }
```
myungjoo: Why do you allow to return 0-filled data if info is NULL?

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -2000,51 +2000,83 @@ TEST (nnstreamer_capi_singleshot, failure_01_n)
   ml_tensors_info_create (&out_info);
 
   /* invalid file path */
-  status = ml_single_open (&single, "wrong_file_name", in_info, out_info,
+  status = ml_single_open (&single, "wrong_file_name", NULL, NULL,
```
myungjoo: If you want to test with NULL in/out-info instead of incorrect in/out-info, please create additional test cases.

api/capi/src/nnstreamer-capi-util.c
jaeyun-jung: it is for data conversion in android jni.

tests/tizen_capi/unittest_tizen_capi.cpp
jaeyun-jung: Setting null is to ignore in/out info param error, this line only checks filepath.

jaeyun-jung: Do you mean to separate null-param cases and invalid-param cases?

api/capi/src/nnstreamer-capi-single.c
jaeyun-jung: I added simple testcase with ml_tensors_info_is_equal and ml_tensors_info_is_valid

tests/tizen_capi/unittest_tizen_capi.cpp
myungjoo: We need both null-param cases and invalid-meta cases.
You may separate them or have both of them in a GTEST case.

api/capi/src/nnstreamer-capi-single.c
myungjoo: Ok.. I mean is it ok to return TRUE when ml_tensor_info_is_equal is FALSE.

api/capi/src/nnstreamer-capi-util.c
myungjoo: This function, ...create_noalloc, is called by invoke or data create() and the metadata should be available at that time.

Another point is that this is Public API. Allowing them (app developers) to have a null data-handler appears to be the root of Tizen app bugs. If you need a null data-handler for JNI, I think you need to do it without this API (or some private/hidden API).

jaeyun-jung: ml_tensors_data_create_no_alloc() is private function for internal usage.

myungjoo: Oh.. ok then.

api/capi/src/nnstreamer-capi-single.c
jaeyun-jung: I added flag is_valid which validates the tensor-info at the last step.

jaeyun-jung:
PR updated
2019-12-06 try to update input info (dynamic dimension) when opening with input info handle.

Issue #1920, opened 2019-12-05
Tensor_sink and tensor_source as subplugins?


myungjoo:
Rather than writing every src/sink nnstreamer element, why don't we create tensor_source and tensor_sink as a general element and let developers write subplugins for them as we do with tensor_filter?

Then, how are we going to do with tensor_sink, which is already exposed as public APIs?
Uh... like.. without explicit option, by default, it works as the current tensor_sink...?

myungjoo:
An alternative is to provide typefind for flatbuffer, protobuf, and so on, along with converter/decoder (flatbuf->tensors->flatbuf).

myungjoo:
The alternative is implemented

Issue #1919, opened 2019-12-05
[EPIC] tensor_decoder::flatbuf, tensor_converter::flatbuf


myungjoo:
Timeline: 2020.2Q (implementation) 3Q (polish, extensive testing)

TODO:
- [x] Study flatbuf, give us a seminar
- [x] Maintain flatbuf in Tizen (discuss and choose the version for Tizen-next and upgrade&integrate)
- [x] Define a standard flatbuffers representation of tensor and tensors
- [x] Write tensor_decoder::flatbuf
- [x] Write tensor_converter::flatbuf
- [x] Integrate such subplugins into Tizen
- [x] Integrate such subplugins into Ubuntu
- [x] Test with heterogeneous devices / Provide as example & sample apps
- [x] Write unit test cases

- You may create each subtask as an issue in this github repo.

myungjoo:
Implement flatbuffer-src as tensor_converter subplugin and
implement flatbuffer-sink as tensor_decoder subplugin.
The infrastructure is ready! :)

Issue #849, opened 2018-11-18
[EPIC] tensor_decoder::protobuf / tensor_converter::protobuf


myungjoo:
Timeline: 2020.2Q (implementation) 3Q (polish, extensive testing)

TODO:
- [x] Study protobuf, give us a seminar  @anyj0527 
- [x] Maintain protobuf in Tizen (discuss and choose the version for Tizen-next and upgrade&integrate)
- [x] Define a standard protobuf representation of tensor and tensors
- [x] Write tensor_decoder::protobuf
- [x] Write tensor_converter::protobuf
- [x] Integrate such subplugins into Tizen
- [x] Integrate such subplugins into Ubuntu
- [ ] Test with heterogeneous devices / Provide as example/sample apps
- [x] Write unit test cases

- You may create each subtask as an issue in this github repo.

For future applications with multi-computer devices,
we may need more than ROS; we may need to send protobuf messages via ROS or general UDP.

We need to study and discuss whether we really need to support protobuf or not.
If we need to support protobuf, it means that we need to make tensor-protobuf and protobuf-tensor converters.

PR #1921, opened 2019-12-05
[Sink] Tensor-Sink Framework

myungjoo:
Objectives
- Support tensor-sink subplugins. We can implement sink-flatbuffers, sink-protobuffers, and many similar subplugins easily.

Considrations
- The coventional tensor-sink should behave correctly. Thus, if "framework" property is not given, the conventional tensor-sink should be invoked.

Addressing part of #1920
Assisting part of #1919 and #849

To Be Determined:
- @jaeyun-jung  has suggested to use ```appsink```. We may use "signal" or "app sink callback". In this case, we can forget about "subplugins" and create each tensor_sink_WHATSOEVER with appsink interface.... The problem is how can we provide portable tensor_sink_WHATSOEVER for public APIs as well, describable in gst pipeline string.


myungjoo:
The alternative #1924 is much better!

PR #1925, opened 2019-12-06
[TFLITE/RELOAD] Revise tflite sub-plugin to support model reloading

dongju-chae:
This PR revises tflite sub-plugin to support model reloading.

It implements an interpreter class with its own mutex, which wraps
existing codes. Also, it's designed to support interpreter switching
to hide the overhead of model reloading.

Related issue: #1900

**changelog**
V4: Add extra notes for a `reloadModel` callback.
V3: Add error handling for model mismatch
V2: Remove operator overloading

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
@@ -69,17 +69,18 @@ tflite_loadModelFile (const GstTensorFilterProperties * prop,
     void **private_data)
 {
   tflite_data *tf;
+  int err;
 
   if (*private_data != NULL) {
     /** @todo : Check the integrity of filter->data and filter->model_files, nnfw */
     tf = *private_data;
-    if (g_strcmp0 (prop->model_files[0],
-            tflite_core_getModelPath (tf->tflite_private_data)) != 0) {
-      tflite_close (prop, private_data);
-    } else {
+
+    if (tflite_core_compareModelPath (tf->tflite_private_data, prop->model_files[0]))
       return 1;
-    }
+
+    tflite_close (prop, private_data);
   }
+
   tf = *private_data = g_new0 (tflite_data, 1);
   if (tf == NULL) {
     g_printerr ("Failed to allocate memory for filter subplugin.");
@@ -87,16 +88,26 @@ tflite_loadModelFile (const GstTensorFilterProperties * prop,
   }
 
   tf->tflite_private_data = tflite_core_new (prop->model_files[0], prop->accl_str);
-  if (tf->tflite_private_data) {
-    if (tflite_core_init (tf->tflite_private_data)) {
-      g_printerr ("failed to initialize the object: Tensorflow-lite");
-      return -2;
-    }
-    return 0;
-  } else {
+  if (tf->tflite_private_data == NULL) {
     g_printerr ("failed to create the object: Tensorflow-lite");
-    return -1;
+    err = -1;
+    goto err_free;
   }
+
+  if (tflite_core_init (tf->tflite_private_data) != 0) {
+    g_printerr ("failed to initialize the object: Tensorflow-lite");
+    tflite_core_delete (tf->tflite_private_data);
+    err = -2;
+    goto err_free;
+  }
+
+  return 0;
+
+err_free:
+  g_free (*private_data);
+  *private_data = NULL;
+
+  return err;
 }
 
 /**
@@ -182,6 +193,25 @@ tflite_setInputDim (const GstTensorFilterProperties * prop, void **private_data,
   return tflite_core_setInputDim (tf->tflite_private_data, in_info, out_info);
 }
 
+/**
+ * @brief The optional callback for GstTensorFilterFramework
+ * @param prop property of tensor_filter instance
+ * @param private_data : tensorflow lite plugin's private data
+ * @return 0 if OK. non-zero if error.
+ */
+static int
+tflite_reloadModel (const GstTensorFilterProperties * prop, void **private_data)
+{
+  tflite_data *tf;
+  tf = *private_data;
+  g_assert (*private_data);
+
+  if (prop->num_models != 1)
+    return -1;
+
+  return tflite_core_reloadModel (tf->tflite_private_data, prop->model_files[0]);
+}
+
 static gchar filter_subplugin_tensorflow_lite[] = "tensorflow-lite";
 
 static GstTensorFilterFramework NNS_support_tensorflow_lite = {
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -411,27 +501,58 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo * info)
...
+
+  /* load a model into the sub interp; it's indepedendent with main one's activities */
+  interpreter_sub.loadModel (model_path, use_nnapi);
+
+  /* assign sub interp to main one; it waits if main interp is already doing invoke() */
+  interpreter.lock();
+  interpreter = interpreter_sub;
+  interpreter.unlock();
+
+  interpreter_sub.unlock();
```
myungjoo: You are double-unlocking interpreter_sub and keeping it locked used-to-be-interpreter. (because of ```interpreter = interpreter_sub;```)

dongju-chae: This assignment does not affect the lock because the assignment operator was overloaded.
It just replaced the internal interpreter instance.

myungjoo: Then, don't override an operator = with "model/interpreter assignment", it's not readable.
Recalling OO experts: how do you think about it?: @jaeyun-jung @again4you

dongju-chae: Got it. I'll make an explicit function to perform this action.

dongju-chae:
**changelog**
V3: Add error handling for model mismatch
V2: Remove operator overloading

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -437,68 +384,310 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo * info)
...
+    err = -EINVAL;
+    goto out_unlock;
+  }
+
+  /**
+   * Everything is ready. let's move the model in sub interpreter to main one.
+   * But, it needs to wait if main interpreter is busy (e.g., invoke()).
+   */
+  interpreter.lock();
+  interpreter.moveInternals (interpreter_sub);
```
kparichay: This style of loading multiple interpreter blocks the main interpreter for shorter duration.
However, this causes double memory requirement because both the models are loaded into memory at this moment.
As nnstreamer targets embedded devices with limited resources, I am not sure if this is better (can cause of ENOMEM) or unloading main interpreter first and then loading the new model.

dongju-chae: Yes, I think each impl. has a tradeoff. But, I think the latter one (i.e., unloading old model and loading new model) has additional disadvantages. If a new model has an incompatible hierarchy with the existing pipeline, loadModel() is failed and then we need to reload old model again. That's a quite big overhead.

So, I think it's better to limit this feature by only enabling reload() in devices with enough memory size.

kparichay: Then, lets add a note that this function needs twice the memory of a single model.

dongju-chae: Yeap. Good suggestion.

dongju-chae:
**changelog**
V4: Add extra notes for a `reloadModel` callback.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -437,68 +384,311 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo * info)
...
+}
+
+/**
+ * @brief	get the model path
+ * @return the model path.
+ */
+const char *
+TFLiteCore::getModelPath ()
+{
+  return model_path;
```
kparichay: add lock here, as reload model can update this within its lock?

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -437,68 +384,311 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo * info)
...
- * @param[out]  output : The array of output tensors
- * @return 0 if OK. non-zero if error.
+ * @brief Move the ownership of interpreter internal members
  */
-int
-TFLiteCore::invoke (const GstTensorMemory * input, GstTensorMemory * output)
+void
+TFLiteInterpreter::moveInternals (TFLiteInterpreter& interp)
+{
+  interpreter = std::move(interp.interpreter);
```
kparichay: spaces between brackets - `std::move (....`
Possibly at other places as well

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
dongju-chae: Updated.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
dongju-chae: Updated.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
kparichay: I think this function should return a copy of the model_path stored inside interpreter, as reloadModel can free this model_path making the return value of this function a hazard.

dongju-chae: True. But, I just changed `getModelPath()` to `compareModelPath()` because it's used only to compare model path in `tensor_filter_tensorflow_lite.c`. Making/freeing its duplicate seems not efficient.

kparichay: compareModelPath is good.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.c
```diff
@@ -182,6 +193,29 @@ tflite_setInputDim (const GstTensorFilterProperties * prop, void **private_data,
...
+  tflite_data *tf;
+  tf = *private_data;
+  g_assert (*private_data);
+
+  if (prop->num_models != 1)
+    return -1;
+
+  /* if the model path is the same, skip reload */
+  if (tflite_core_compareModelPath (tf->tflite_private_data, prop->model_files[0]))
+    return 0;
```
myungjoo: What if the weight values are updated with the given model?

dongju-chae: Right. we can provide the same model path with different weights. I'll remove this line.

dongju-chae:
Got it! Thanks for the detail scenarios.

PR #1926, opened 2019-12-06
[TEST/RELOAD] Add unittests for set_property of model/is-updatable and some negative scenarios

dongju-chae:
This PR adds unittest to test set_property of model/is-updatable, and some negative unittests.

It tests reloadModel() of tflite subplugin but it's not asynchronous.
It also adds additional model file, mobilenet v2, used for model reload.

For negative tests, the belows are tested.
1. Update model file name (a.tflite --> b.tflite) but b.tflite not exists               
2. Update model file name (a.tflite --> b.tflite) but b.tflite has different dimensions 
3. Update without file name changes but the file is gone (deleted)                      
4. Update without file name changes but the file has now different dimensions           

Note that testing reloads in runtime will be done in another commit (#1933).

**Self evaluation:**


tests/nnstreamer_plugins/unittest_plugins.cpp
```diff
@@ -2992,6 +2993,352 @@ TEST (test_tensor_filter, reopen_tflite_02_p)
...
+
+  /* close tf-lite model */
+  fw->close (prop, &private_data);
+
+  g_free (prop);
+  g_free (test_model);
+}
+
+/**
+ * @brief Test to reload tf-lite; model has wrong dimension (negative)
```
kparichay: I think this test is for reload model file does not exist.

tests/nnstreamer_plugins/unittest_plugins.cpp
```diff
@@ -2992,6 +2993,352 @@ TEST (test_tensor_filter, reopen_tflite_02_p)
...
+
+  g_free (prop);
+  g_free (test_model);
+  g_free (test_model_renamed);
+}
+
+/**
+ * @brief Test to reload tf-lite; model has wrong dimension (negative)
+ */
+TEST (test_tensor_filter, reload_tflite_same_model_wrong_dims_n)
```
kparichay: Is this testcase needed? I think this is directly overlapped with `reload_tflite_model_wrong_dims_n` as reloadModel implementation does not different models with same or different names.

tests/nnstreamer_plugins/unittest_plugins.cpp
dongju-chae: Oh, wrong comment. will be updated

tests/nnstreamer_plugins/unittest_plugins.cpp
```diff
@@ -2992,6 +2993,352 @@ TEST (test_tensor_filter, reopen_tflite_02_p)
...
+  fw->close (prop, &private_data);
+
+  g_free (prop);
+  g_free (test_model);
+}
+
+/**
+ * @brief Test to reload tf-lite; model has wrong dimension (negative)
+ */
+TEST (test_tensor_filter, reload_tflite_model_wrong_dims_n)
```
kparichay: This comment is not directly related to this PR.
If the new model to be reloaded has different dimension, then the current implementation and test case makes the reloading fail.
However, if the new model to be reloaded allows changing the dimension, and if the dimension for the new model can be changed to the required dimension, then reload can be success.

tests/nnstreamer_plugins/unittest_plugins.cpp
dongju-chae: Yes, the implementation of `reloadModel` does not care about the original model file by overwriting it because the weight values can be changed. But, this testcase simulates various scenarios, and in the near future, its implementations can be changed, satisfying these unittests. So, I think it's meaningful :)

tests/nnstreamer_plugins/unittest_plugins.cpp
dongju-chae: Actually, supporting different dimensions is not a primary target because we're targeting on-device training like changed weights in runtime. If someone wants to reload a new model with different dimensions, we can say like just relaunching the pipeline and re-performing caps navigation.

kparichay: Yes, the scenario I presented is different from on-device training.
In this case, the models default input dimensions are different, but can support the currently set input dimensions. Anyway, this can be considered later when needed.

tests/nnstreamer_plugins/unittest_plugins.cpp
kparichay: ok.

tests/nnstreamer_plugins/unittest_plugins.cpp
dongju-chae: Sure. Let's do it step-by-step.

PR #1930, opened 2019-12-09
Refactor/Conf: avoid using hardcoded categories

myungjoo:
To handle paths/files of each subplugin category,
use loops and enums instead of hardcoded variables.

### DIFF
--- a/gst/nnstreamer/nnstreamer_conf.c
+++ b/gst/nnstreamer/nnstreamer_conf.c
@@ -27,6 +27,33 @@
 
 #include "nnstreamer_conf.h"
 
+/**
+ * Note that users still can place their custom filters anywhere if they
+ * designate them with the full path.
+ */
+
+/* Subplugin Naming Rules */
+#define NNSTREAMER_PREFIX_DECODER	"libnnstreamer_decoder_"
+#define NNSTREAMER_PREFIX_FILTER	"libnnstreamer_filter_"
+#define NNSTREAMER_PREFIX_CUSTOMFILTERS	""
+#define NNSTREAMER_PREFIX_CONVERTER	""
+/* Custom filter does not have prefix */
+
+/* Env-var names */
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -27,6 +27,33 @@
 
 #include "nnstreamer_conf.h"
 
+/**
+ *  Note that users still can place their custom filters anywhere if they
```
dongju-chae: Redundant whitespace :)

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -37,35 +64,31 @@ static const gchar *subplugin_prefixes[] = {
...
-
   /*************************************************
    * Cached Raw Values                             *
    * 0: ENVVAR. 1: CONFFILE, 2: Hardcoded          *
    *************************************************/
-  gchar *pathFILTERS[CONF_SOURCES];        /**< directory paths for FILTERS */
-  gchar *pathDECODERS[CONF_SOURCES];       /**< directory paths for DECODERS */
-  gchar *pathCUSTOM_FILTERS[CONF_SOURCES]; /**< directory paths for CUSTOM FILTERS */
-  gchar *pathCONVERTERS[CONF_SOURCES]; /**< directory paths for CUSTOM FILTERS */
+  gchar *path[CONF_SOURCES]; /**< directory paths */
```
dongju-chae: How about defining `enum CONF_SOURCES`? Then, line 350,369, and 370 can be revised to have straightforward meanings.

myungjoo: Updated. Thanks!

gst/nnstreamer/nnstreamer_conf.c
myungjoo: Fixed

PR #1933, opened 2019-12-10
[TEST/RELOAD] Test runtime model reloads during tflite inferences

dongju-chae:
This PR adds a testcase to test asynchronous model updates during
tflite inferences. It repeatedly switches tflite models between
mobilenet_v1 and mobilenet_v2. It expects the same index for input img.

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_filter_reload/tensor_filter_reload_test.c
+++ b/tests/nnstreamer_filter_reload/tensor_filter_reload_test.c
@@ -1,0 +1,279 @@
+/**
+ * @file    tensor_filter_reload_test.c
+ * @data    19 Dec 2019
+ * @brief   test case to test a filter's model reload
+ * @see     https://github.com/nnsuite/nnstreamer
+ * @author  Dongju Chae <dongju.chae@samsung.com>
+ * @bug     No known bugs except NYI.
+ */
+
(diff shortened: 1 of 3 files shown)


tests/nnstreamer_filter_reload/tensor_filter_reload_test.c
```diff
@@ -0,0 +1,279 @@
...
+  return GST_FLOW_OK;
+}
+
+/**
+ * @brief Reload a tensor filter's model (v1 <-> v2)
+ */
+static gboolean
+reload_model (GstElement *tensor_filter)
+{
+  static gboolean is_first = TRUE;
```
kparichay: Overall LGTM
However, I do not understand the purpose of this variable. As far as I can see, this function is called only once after EVENT_INTERVAL timeout.

dongju-chae: The callback of timeout is repeatedly called if it returns TRUE. 
The period of timeout is interval between events, not an one-shot trigger.

kparichay: aah right, the timeout is called repeatedly.

PR #1942, opened 2019-12-12
[Android] common headers for gstreamer

jaeyun-jung:
define common header path for gstreamer.

### DIFF
--- a/jni/nnstreamer.mk
+++ b/jni/nnstreamer.mk
@@ -1,8 +1,11 @@
 # This mk file defines common features to build NNStreamer library for Android.
 
 ifndef NNSTREAMER_ROOT
-$(warning NNSTREAMER_ROOT is not defined! Using $(LOCAL_PATH)/.. )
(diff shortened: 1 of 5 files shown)


jni/nnstreamer.mk
```diff
@@ -1,8 +1,11 @@
 # This mk file defines common features to build NNStreamer library for Android.
 
 ifndef NNSTREAMER_ROOT
-$(warning NNSTREAMER_ROOT is not defined! Using $(LOCAL_PATH)/.. )
-NNSTREAMER_ROOT := $(LOCAL_PATH)/..
+$(error NNSTREAMER_ROOT is not defined!)
+endif
```
myungjoo: Can you define ```NNSTREAMER_ROOT``` and its friends in nnstreamer.mk and remove them from other .mk files?

jaeyun-jung: NNSTREAMER_ROOT is defined in gradle build script (system.env or local properties of android-studio project)

PR #1943, opened 2019-12-12
[TensorSink/UnitTest] Add a test option for sync property in TensorSink

dongju-chae:
This PR adds a test option, `need_sync`, for sync property in `TensorSink`.

Basically, GstBaseSink has a `sync` property enabled by default, and
`TensorSink` also does. In this case, if the frame is later than max-lateness,
the sink will drop the buffer without calling the render method.

It can be a problem if some unittests (e.g., `signals`) expects the exact
number of received data. In slow system, if a signal callback is too slow,
some buffer can be dropped.

To fix this problem, this patch adds a test option for sync property and
sets `sync` property as `FALSE` by default to disable the buffer dropping
feature for `signals` unittest.

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_sink/unittest_sink.cpp
+++ b/tests/nnstreamer_sink/unittest_sink.cpp
@@ -138,6 +138,7 @@ typedef struct
   TestType test_type; /**< test pipeline */
   tensor_type t_type; /**< tensor type */
   char *tmpfile; /**< tmpfile to write */
+  gboolean need_sync; /**< sync on the clock */
 } TestOption;
 
 /**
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_sink/unittest_sink.cpp
```diff
@@ -1100,6 +1100,8 @@ TEST (tensor_sink_test, signals)
       (GCallback) _eos_cb, NULL);
   EXPECT_TRUE (handle_id > 0);
 
+  g_object_set (g_test_data.sink, "sync", FALSE, NULL);
```
myungjoo: Or.. you can edit line 456:
```
-          "tensor_converter ! tensor_sink name=test_sink", option.num_buffers,
+          "tensor_converter ! tensor_sink sync=false name=test_sink", option.num_buffers,
```

myungjoo: Plus there are so many similar cases if you look at line 456 and its neighbors.

dongju-chae: Yes, the meaning is the same. But, changing flags directly in `TEST_TYPE_VIDEO_RGB` might occur some problems because it's used in other unittests. Actually, other tests may not need this flag because they did not expect the exact number of buffer sinks.

jaeyun-jung: If this can fix the time issue, I think it is better to add this in _setup_pipeline()

dongju-chae: I added a common test option, `need_sync`.

dongju-chae: I added a common test option, `need_sync`.

dongju-chae:
@myungjoo I checked all unittests for x86_64/aarch64/armv7l. But I think I have no permission to merge this PR :(

PR #1949, opened 2019-12-16
[spec] nnfw now packaged only for tizen

kparichay:
nnfw package is now created and corresponding tests performed
only for tizen in the spec file.

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -79,6 +79,9 @@ BuildRequires:	pkgconfig(dlog)
 BuildRequires:	pkgconfig(libmvnc)
 BuildRequires:	gst-plugins-bad-devel
 BuildRequires:	gst-plugins-base-devel
+# For tizen sensor support
+BuildRequires:	pkgconfig(sensor)
+BuildRequires:	capi-system-sensor-devel
 
 %if 0%{?enable_nnfw}
 %ifarch %arm aarch64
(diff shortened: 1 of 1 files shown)


packaging/nnstreamer.spec
```diff
@@ -101,11 +104,6 @@ BuildRequires: ssat
 # For ORC (Oil Runtime Compiler)
 BuildRequires:	pkgconfig(orc-0.4)
 
-%if %{with tizen}
```
dongju-chae: is this not required now?

kparichay: This `if %{with tizen}` block added `BuildRequires` for tizen sensor framework.
This has been moved to line 82 into the existing `if %{with tizen}` block

dongju-chae: Ah.. I missed line 71.

packaging/nnstreamer.spec
```diff
@@ -455,6 +446,13 @@ cp -r result %{buildroot}%{_datadir}/nnstreamer/unittest/
 %files -n nnstreamer-tizen-internal-capi-devel
 %{_includedir}/nnstreamer/nnstreamer-tizen-internal.h
 
+%if 0%{?enable_nnfw_r}
+%files nnfw
+%manifest nnstreamer.manifest
+%defattr(-,root,root,-)
+%{_prefix}/lib/nnstreamer/filters/libnnstreamer_filter_nnfw.so
```
wooksong: Instead of hard-coded 'lib', it is better to use %{_libdir}.

myungjoo: No.
For application/package specific paths, it is /usr/lib/NAME/whatsoever, not %{_libdir}/NAME/whatsoever.

Issue #1849, opened 2019-11-07
[filter/plugins] Manage accelerators for tensor filter plugins properly


kparichay:
Properly manage and set backend accelerators for tensor filter subplugins

PR #1950, opened 2019-12-16
[nnfw] Set accelerator for nnfw

kparichay:
Allow setting of accelerator for nnfw
Parsing accelerator moved to tensor_filter_common. Currently, used in
tflite and nnfw, and use directly for other filters like this.

Related Issue: #1849 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -44,6 +44,26 @@
 #define NNFW_SRCN_BACKEND  "srcn"
 #define NNFW_DEFAULT_BACKEND NNFW_CPU_BACKEND
 
+/** Match all accelerators for nnapi at once */
+#define REGEX_ACCL_NNFW \
+  "(^(true)[:]?([(]?(" \
+  REGEX_ACCL_AUTO "|" \
+  REGEX_ACCL_DEF "|" \
+  REGEX_ACCL_CPU "|" \
+  REGEX_ACCL_GPU "|" \
+  REGEX_ACCL_SRCN "|" \
+  REGEX_ACCL_NEON ")*[)]?))"
+
+/** Match accelerator for nnapi one by one */
+#define REGEX_ACCL_NNFW_ELEM \
+  "(" \
+  "(?<!!)" ACCL_AUTO_STR "|" \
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite_core.cc
```diff
@@ -431,40 +431,17 @@ TFLiteCore::TFLiteCore (const char * _model_path, const char * accelerators)
...
-        g_free (word);
-        break;
-      }
-    }
-    g_match_info_free (match_info);
-    g_regex_unref (nnapi_elem);
-  } else  {
-    goto use_nnapi_ini;
-  }
+  else
```
dongju-chae: It seems `else` is not necessary.

kparichay: updated to set `use_nnapi = TRUE` at the start of the function.

myungjoo:
Could you please add the following diff to PROP_ACCELEARTTOR:
```
diff --git a/gst/nnstreamer/tensor_filter/tensor_filter_common.c b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
index 8a6ac45..7edf4b8 100644
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -268,7 +268,8 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
           "(true/false):(comma separated ACCELERATOR(s)). "
           "true/false determines if accelerator is to be used. "
-          "Example, if GPU, NPU can be used but not CPU - true:(GPU,NPU,!CPU).",
+          "Example, if GPU, NPU can be used but not CPU - true:(GPU,NPU,!CPU). "
+          "Note that only a few subplugins support this property.",
           "", G_PARAM_READWRITE));
   g_object_class_install_property (gobject_class, PROP_IS_UPDATABLE,
       g_param_spec_boolean ("is-updatable", "Updatable model",
```

kparichay:
V2:
Added note that PROP_ACCELERATOR is supported in a few filter extensions only.

kparichay:
> Please add corresponding test cases (including Tizen CAPI) in later PRs.

Yes, will do

PR #1957, opened 2019-12-16
[WIP] Test: filter/edgetpu unit testing

myungjoo:
WIP:
- Create dummy edgetpu library for unit tests
    
TODO:
- Write meson.build & unit-test body
    

tests/meson.build
```diff
@@ -117,6 +117,10 @@ if get_option('enable-nnfw-runtime')
   subdir('tizen_nnfw_runtime')
 endif
 
+if get_option('enable-edgetpu')
```
kparichay: Add `get_option('enable-edgetpu')` to the list of tensor filter which requires `test_models` to be installed along with them (at line 102)

myungjoo:
My dev-machine's network is severed. I'll update when it's recovered.

tests/meson.build
myungjoo: Fixed.

PR #1964, opened 2019-12-17
[Api] add SNAP in nnfw enum type

jaeyun-jung:
Define SNAP in nnfw type, only for Android.


jaeyun-jung:
PR updated
2019-12-24 add snap to set model info in filter

PR #1973, opened 2019-12-19
[single/hw] Check hw support for framework @open sesame 1/716:07

kparichay:
Add implementation to check hw support for framework while checking for its availability.
Added unittests for this functionality with tensorflow-lite

Self evaluation:

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -1030,6 +1030,40 @@ ml_validate_model_file (const char *model, ml_nnfw_type_e * nnfw)
 }
 
 /**
+ * @brief Convert c-api based hw to internal representation
+ */
+static accl_hw
+ml_nnfw_to_accl_hw (const ml_nnfw_hw_e hw)
+{
+  switch (hw) {
+    case ML_NNFW_HW_ANY:
+      return ACCL_DEFAULT;
+    case ML_NNFW_HW_AUTO:
+      return ACCL_AUTO;
+    case ML_NNFW_HW_CPU:
+      return ACCL_CPU;
+    case ML_NNFW_HW_CPU_NEON:
+      return ACCL_CPU_NEON;
+    case ML_NNFW_HW_GPU:
+      return ACCL_GPU;
+    case ML_NNFW_HW_NPU:
+      return ACCL_NPU;
+    case ML_NNFW_HW_NPU_MOVIDIUS:
+      return ACCL_NPU_MOVIDIUS;
+    case ML_NNFW_HW_NPU_EDGE_TPU:
+      return ACCL_NPU_EDGE_TPU;
+    case ML_NNFW_HW_NPU_VIVANTE:
+      return ACCL_NPU_VIVANTE;
+    case ML_NNFW_HW_NPU_SRCN:
+      return ACCL_NPU_SRCN;
+    case ML_NNFW_HW_NPU_SR:
+      return ACCL_NPU_SR;
+    default:
+      return ACCL_AUTO;
+  }
+}
+
+/**
  * @brief Checks the availability of the given execution environments.
  */
 int
@@ -1072,10 +1106,13 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
       break;
   }
 
-  /* finally check filter sub-plugin */
   if (fw_name) {
     if ((fw = nnstreamer_filter_find (fw_name)) != NULL) {
-      *available = true;
+      if (fw->checkAvailability && fw->checkAvailability (ml_nnfw_to_accl_hw (hw)) != 0) {
+        ml_logw ("%s is supported but not with the specified hardware.", fw_name);
+      } else {
+        *available = true;
+      }
     } else {
       ml_logw ("%s is not supported.", fw_name);
     }
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -211,6 +211,13 @@ typedef struct _GstTensorFilterFramework
        * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, close() frees private_data and set NULL.
        * @return 0 if ok. < 0 if error.
        */
+
+  int (*checkAvailability) (accl_hw hw);
+      /**< Optional. Check if the provided hardware accelerator is supported. This check is static or dynamic based on framework support. Positive response of this check does not guarantee successful running of model with this accelerator. The static check can be performed without opening the framework.
+       *
+       * @param[in] hw backend accelerator hardware
+       * @return 0 if supported. -errno if not supported.
+       */
 } GstTensorFilterFramework;
 
 /* extern functions for subplugin management, exist in tensor_filter.c */
--- a/tests/tizen_capi/unittest_tizen_capi.cc
+++ b/tests/tizen_capi/unittest_tizen_capi.cc
@@ -1276,6 +1276,70 @@ TEST (nnstreamer_capi_util, availability_00)
 #endif  /* ENABLE_NNFW_RUNTIME */
 }
 
+#ifdef ENABLE_TENSORFLOW_LITE
+/**
+ * @brief Test NNStreamer Utility for checking availability of Tensorflow-lite backend
+ */
+TEST (nnstreamer_capi_util, availability_01)
+{
+  bool result;
+  int status;
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_ANY, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, true);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_AUTO, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, true);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_CPU, &result);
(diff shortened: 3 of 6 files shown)


gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -211,6 +211,13 @@ typedef struct _GstTensorFilterFramework
        * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, close() frees private_data and set NULL.
        * @return 0 if ok. < 0 if error.
        */
+
+  int (*checkHWSupport) (accl_hw hw);
+      /**< Mandatory. Check if the provided hardware accelerator is supported. This check can be static or dyanmic based on framework support. Positive response of this check does not guarantee successful running of model with this accelerator. This check can be performed without opening the framework.
```
jaeyun-jung: typo dyanmic

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -1043,29 +1078,30 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
 
   switch (nnfw) {
     case ML_NNFW_TYPE_CUSTOM_FILTER:
+      fw = nnstreamer_filter_find ("custom");
       /* Always available */
```
dongju-chae: if it's always available, how about adding `assert`? (because line 1114 implies a potential segmentation fault)

gst/nnstreamer/nnstreamer_plugin_api_filter.h
dongju-chae: I think it's `Optional`, not `Mandatory`.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -28,23 +28,17 @@
...
-#define REGEX_ACCL_CPU  REGEX_ACCL_EACH_ELEM(ACCL_CPU_STR)
-#define REGEX_ACCL_GPU  REGEX_ACCL_EACH_ELEM(ACCL_GPU_STR)
-#define REGEX_ACCL_NPU  REGEX_ACCL_EACH_ELEM(ACCL_NPU_STR)
-#define REGEX_ACCL_NEON REGEX_ACCL_EACH_ELEM(ACCL_NEON_STR)
-#define REGEX_ACCL_SRCN REGEX_ACCL_EACH_ELEM(ACCL_SRCN_STR)
-#define REGEX_ACCL_DEF  REGEX_ACCL_EACH_ELEM(ACCL_DEF_STR)
+#define ACCL_NPU_MOVIDIUS_STR  "npu.movidius"
+#define ACCL_NPU_EDGE_TPU_STR  "npu.edgetpu"
+#define ACCL_NPU_VIVANTE_STR  "npu.vivante"
+#define ACCL_NPU_SRCN_STR  "npu.srcn" /** srcn hardware supported by nnfw */
```
dongju-chae: What does `SRCN` stand for?

tests/tizen_capi/unittest_tizen_capi.cpp
```diff
@@ -1276,6 +1276,49 @@ TEST (nnstreamer_capi_util, availability_00)
...
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, true);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_NPU_SRCN, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, false);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_NPU_EDGE_TPU, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, false);
```
dongju-chae: How about checking other remaining backends as well if this sub-test also checks unimplemented ones?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: SRCN is one of the backends supported by NNFW-runtime.
check https://review.tizen.org/gerrit/gitweb?p=platform/core/ml/nnfw.git;a=summary for more info.

dongju-chae: Thx

tests/tizen_capi/unittest_tizen_capi.cpp
kparichay: Added all of them.

api/capi/src/nnstreamer-capi-util.c
kparichay: done.

gst/nnstreamer/nnstreamer_plugin_api_filter.h

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -211,6 +211,13 @@ typedef struct _GstTensorFilterFramework
        * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, close() frees private_data and set NULL.
        * @return 0 if ok. < 0 if error.
        */
+
+  int (*checkHWSupport) (accl_hw hw);
+      /**< Optional. Check if the provided hardware accelerator is supported. This check can be static or dynamic based on framework support. Positive response of this check does not guarantee successful running of model with this accelerator. This check can be performed without opening the framework.
+       *
+       * @param[in] hw backend accelerator hardware
+       * @return 0 if supported. -errno if not supported.
+       */
```
myungjoo: Question: Is this called even if it's NOT opened?
If it's supposed to be (or if you may modify the flow accordingly) called after it's opened, you can do this with open.

Suggestion: This should not only applied to HW. It should check the general capability and availability of the given condition.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -249,6 +256,12 @@ get_accl_hw_type (const char * str);
 extern const char *
 get_accl_hw_str (const accl_hw key);
 
+/**
+ * @brief check if the given accelerator is supported
+ */
+extern int
+gst_tensor_filter_check_hw_support (const char * const *supported_hw_list, accl_hw hw);
```
myungjoo: Usually, this is supposed to be probed in run-time (check if the hardware is really available)... Then, providing such an internal API is awkward.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1276,6 +1276,61 @@ TEST (nnstreamer_capi_util, availability_00)
...
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, true);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_NPU, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, true);
+
+  status = ml_check_nnfw_availability (ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_NPU_MOVIDIUS, &result);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  EXPECT_EQ (result, false);
```
myungjoo: You are supposed to create more negative cases when you add positive case.
Separate negative test cases (with ```_n``` postfix) and add these "false" asserts there.

kparichay: Got it, updated.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: This API currently determines statically. Later, the implementation of this API will be updated to determine if the hardware is really available (dependent on the platform).

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: As single API provides `int ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw, bool *available);`, this function SUPPORTS calling this even if the framework is NOT opened.

Updated `open()` to ensure checking that the passed hw is supported.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
myungjoo: I don't think we can provide some internal APIs for hardware detection. The responsibility should be on subplugins, which are supposed to be closer to layers of device drivers. Otherwise, we are breaking layers of platform architectures.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: > Suggestion: This should not only applied to HW. It should check the general capability and availability of the given condition.

Are there any other capabilities other than supported `data types`?

myungjoo: Then, Let's rename ```checkHWSupport``` --> ```checkAvailability```. In the future, I feel that we may use GstTensorFilterPropertires instead of accl_hw for this callback; we may use accl_hw instead of const char* for accl_str in GstTensorFilterProperties in the future. Then, we may call ```checkAvailabiliby``` for general availability check for a subplugin without actual model files given.

kparichay: Renamed to `checkAvailability`. 
Using GstTensorFilterProperties in this callback is better - will update later.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Yes, that is right. Updated to directly use `g_strv_contains`.

kparichay:
V2: 
check hardware support performed while parsing for available hardwares
Split availability test cases into positive and negative

V3: 
checkHWSupport->checkAvailability

Issue #1959, opened 2019-12-17
Remove assert from extensions + add error checking testcases


kparichay:
Remove asserts on the return values of the functions used in tensor filter extensions.
This will allow adding unit-tests for failure scenarios.

PR #1975, opened 2019-12-23
[filter/ext] Replace asserts with error handling @open sesame 12/23 19:36

kparichay:
Replace asserts with appropriate error handling in tensor filter extensions
This will allows test cases for failure scenarios and handle erronous inputs
via single-shot API gracefully with error return than shutting down with assert failure

Issue : #1959 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
@@ -90,7 +90,6 @@ std::map <char*, Tensor*> Caffe2Core::inputTensorMap;
  */
 Caffe2Core::Caffe2Core (const char * _model_path, const char *_model_path_sub)
 {
-  g_assert (_model_path != NULL && _model_path_sub != NULL);
   init_model_path = g_strdup (_model_path);
   pred_model_path = g_strdup (_model_path_sub);
   first_run = true;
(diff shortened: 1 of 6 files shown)


kparichay:
V2:
Added updates for custom filter

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
```diff
@@ -90,7 +90,6 @@ std::map <char*, Tensor*> Caffe2Core::inputTensorMap;
  */
 Caffe2Core::Caffe2Core (const char * _model_path, const char *_model_path_sub)
 {
-  g_assert (_model_path != NULL && _model_path_sub != NULL);
   init_model_path = g_strdup (_model_path);
```
myungjoo: What happens if they are NULL?
Do we handle such conditions properly at the start of a pipeline?

kparichay: [Line 465](https://github.com/nnsuite/nnstreamer/blob/37b0603e9826d9328f50f6a3c1d4bdd645c2c06f/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc#L465) in this file ensures that they both are not NULL.

kparichay: Because constructor cannot return with failure, argument checks are performed in caller function.

PR #1976, opened 2019-12-23
[filter/ext] Added basic unittests for all filter extensions

kparichay:
Added basic unittests for all the the filter extensions
This creates failure cases of unittests for the interface defined with GstTensorFilterFramework
for most of the extensions, and performs basic unittests
Updated the unittest list in packaging files as well
    
Also added a script file to run all the test cases inside a folder
    
TODO:  python extensions in this as well

Issue: #1959

Self evaluation:

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -319,32 +319,23 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} --libdir
 ninja -C build %{?_smp_mflags}
 
 export NNSTREAMER_BUILD_ROOT_PATH=$(pwd)
+export GST_PLUGIN_PATH=$(pwd)/build/gst/nnstreamer
+export NNSTREAMER_CONF=$(pwd)/build/nnstreamer-test.ini
+export NNSTREAMER_FILTERS=$(pwd)/build/ext/nnstreamer/tensor_filter
+export NNSTREAMER_DECODERS=$(pwd)/build/ext/nnstreamer/tensor_decoder
 
-## Run unittests
-pushd build
-export GST_PLUGIN_PATH=$(pwd)/gst/nnstreamer
-export NNSTREAMER_CONF=$(pwd)/nnstreamer-test.ini
-export NNSTREAMER_FILTERS=$(pwd)/ext/nnstreamer/tensor_filter
-export NNSTREAMER_DECODERS=$(pwd)/ext/nnstreamer/tensor_decoder
-%if 0%{?unit_test}
-    ./tests/unittest_common --gtest_output="xml:unittest_common.xml"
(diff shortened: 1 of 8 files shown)


packaging/nnstreamer.spec
```diff
@@ -284,28 +284,23 @@ meson --buildtype=plain --prefix=%{_prefix} --sysconfdir=%{_sysconfdir} --libdir
...
-%endif
 %if %{with tizen}
-    %if 0%{?enable_nnfw_r}
-        ./tests/tizen_nnfw_runtime/unittest_nnfw_runtime_raw --gst-plugin-path=. --gtest_output="xml:unittest_nnfw_runtime_raw.xml"
-    %endif
+    bash %{test_script} ./tests/tizen_nnfw_runtime/unittest_nnfw_runtime_raw
     ln -s ext/nnstreamer/tensor_source/*.so .
-    ./tests/tizen_capi/unittest_tizen_sensor --gst-plugin-path=. --gtest_output="xml:unittest_tizen_sensor.xml"
+    bash %{test_script} ./tests/tizen_capi/unittest_tizen_sensor
+%endif
```
dongju-chae: Why these unittests are not inside `unit_test` if-statement? Was it intended?

kparichay: `unit_test` is only enabled in x86_64, and nnfw_runtime is not available for that architecture.
However, tizen sensor framework seems to be available for x86_64 architecture.
If you want to move it under `unit_test condition`, I think a different PR would be good.

dongju-chae: OK

myungjoo: Sensor framework depends on Tizen. We want to test it unconditionally in Tizen because it may get regressions from other Tizen packages. Anyway, we have disabled most unittests in ARM because it slows down the build too much for CI systems.

tests/nnstreamer_filter_extensions_common/unittest_tizen_template.cpp
```diff
@@ -0,0 +1,264 @@
...
+    .model_files = const_cast<const char **>(model_files),
+    .num_models = (int) g_strv_length (model_files),
+  };
+
+  const GstTensorFilterFramework *sp = nnstreamer_filter_find ("EXT_NAME");
+  EXPECT_NE (sp, (void *) NULL);
+
+  /** invoke without open */
+  ret = sp->invoke_NN (&prop, &data, &input, &output);
+  EXPECT_NE (ret, 0);
```
myungjoo: Let's make sure these have ```_n``` postfixes in their test case names.

kparichay:
V2:
Split test cases into successful and negative ones

tests/nnstreamer_filter_extensions_common/unittest_tizen_template.cpp
kparichay: updated - split test cases into successful and negative ones (negative ones suffixed with `_n`)

PR #1978, opened 2019-12-24
[subplugin/ext] Added support for setInputDim/reloadModel in unittests

### LINKED ISSUE #1959: Remove assert from extensions + add error checking testcases

kparichay:
Added check for setInputDim/reloadModel in unittests
Framework functions are reset at close in python subplugin
Framework functions are updated based on available in model file in custom subplugin

Resolves  #1959

Self evaluation:

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_custom.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_custom.c
@@ -176,7 +176,7 @@ custom_getInputDim (const GstTensorFilterProperties * prop, void **private_data,
   g_return_val_if_fail (info != NULL, -EINVAL);
 
   if (ptr->methods->getInputDim == NULL)
-    return -EINVAL;
+    return -ENOENT;
 
   return ptr->methods->getInputDim (ptr->customFW_private_data, prop, info);
 }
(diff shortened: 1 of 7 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_custom.c
```diff
@@ -38,6 +38,7 @@ static void init_filter_custom (void) __attribute__ ((constructor));
 static void fini_filter_custom (void) __attribute__ ((destructor));
 
 static GstTensorFilterFramework NNS_support_custom;
```
dongju-chae: let's remove this cuz its pointer is added.

kparichay: Updated. Removed the added pointer, and kept this instead.

dongju-chae: In fact, I think re-declaration of global variables is not good approach, and its compilation is prohibited in C++. But, anyway, this code works in C.

kparichay: Yes, you are right, C++ gives compilation error. I removed the re-declaration, and used pointer style.

myungjoo:
The commit  3bbe0f679321ec5979848f1aee9005dabb325ca2 appears to have a lot of changes not coherent with the commit message. Is it a rebase mistake?

kparichay:
> The commit [3bbe0f6](https://github.com/nnsuite/nnstreamer/commit/3bbe0f679321ec5979848f1aee9005dabb325ca2) appears to have a lot of changes not coherent with the commit message. Is it a rebase mistake?

Rebased, but still there are a lot of changes - this is due reason that adding testcases of  `setInputDim` needed a lot of change. I will split this into two parts.

Issue #1708, opened 2019-09-27
[Filter] OpenVINO Support


myungjoo:
Accept OpenVINO models for Movidius Sticks.

CC: @yunakim

PR #1989, opened 2020-01-02
[Ext/Filter/Meson] Update meson.build for the OpenVino sub-plugin

wooksong:
Related to #1708.

This patch updates the meson build script for the tensor filter sub-plugins to support OpenVino.

**Self evaluation:**


meson_options.txt
```diff
@@ -2,9 +2,9 @@ option('enable-test', type: 'boolean', value: true)
 option('enable-orc', type: 'boolean', value: true) # default true, use orc when found orc library
 option('install-test', type: 'boolean', value: false)
 option('enable-tensorflow-lite', type: 'boolean', value: true)
-option('enable-tensorflow', type: 'boolean', value: true)
-option('enable-caffe2', type: 'boolean', value: true)
-option('enable-pytorch', type: 'boolean', value: true)
+option('enable-tensorflow', type: 'boolean', value: false)
+option('enable-caffe2', type: 'boolean', value: false)
+option('enable-pytorch', type: 'boolean', value: false)
```
dongju-chae: How about disabling other sub-plugins (e.g., `tensorflow-lite`) as well in `meson_options.txt`?
I think enabling them by default doesn't seem like a good approach due to forced-dependency.
A build script of each system may enable them if required.

wooksong: @dongju-chae You made a good point and we need to discuss it. However, those options are disabled by mistake (for the test purpose). At least, in this PR, I will set them to true:)

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -302,3 +302,34 @@ if get_option('enable-edgetpu')
     install_dir: filter_subplugin_install_dir
   )
 endif
+
+if get_option('enable-openvino')
+  openvino_deps = []
+  openvino_cpp_args = []
+  openvino_deps += dependency('libinference-engine', required: true)
```
myungjoo: Let's update this (after you update openVINO's packaging)

PR #1993, opened 2020-01-03
[Filter/OpenVino] Fill the mandatory callback, invoke

wooksong:
Related to #1708.

This PR adds a method to run inference for given input tensors using InferenceEngine on the acceleration device. It also implements the mandatory callback of GstTensorFilterFramework, invoke, with such method.

Running tests are done with inception v3.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
@@ -56,6 +56,9 @@ public:
   };
 
   static tensor_type convertFromIETypeStr (std::string type);
+  static InferenceEngine::Blob::Ptr convertGstTensorMemoryToBlobPtr (
+      const InferenceEngine::TensorDesc tensorDesc,
+      const GstTensorMemory * gstTensor);
 
   TensorFilterOpenvino (std::string path_model_prefix, accl_hw hw);
   ~TensorFilterOpenvino ();
@@ -65,6 +68,8 @@ public:
 
   int getInputTensorDim (GstTensorsInfo * info);
   int getOutputTensorDim (GstTensorsInfo * info);
+  int invoke (const GstTensorFilterProperties * prop,
+      const GstTensorMemory * input, GstTensorMemory * output);
   std::string getPathModelXml ();
   std::string getPathModelBin ();
 
@@ -78,8 +83,11 @@ private:
   InferenceEngine::CNNNetReader _networkReaderCNN;
   InferenceEngine::CNNNetwork _networkCNN;
   InferenceEngine::InputsDataMap _inputsDataMap;
+  InferenceEngine::TensorDesc _inputTensorDescs[NNS_TENSOR_SIZE_LIMIT];
   InferenceEngine::OutputsDataMap _outputsDataMap;
+  InferenceEngine::TensorDesc _outputTensorDescs[NNS_TENSOR_SIZE_LIMIT];
   InferenceEngine::ExecutableNetwork _executableNet;
+  InferenceEngine::InferRequest _inferRequest;
 
   std::string pathModelXml;
   std::string pathModelBin;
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
```diff
@@ -207,17 +275,21 @@ TensorFilterOpenvino::loadModel ()
     this->_ieCore.AddExtension(
         std::make_shared<InferenceEngine::Extensions::Cpu::CpuExtensions>(),
         "CPU");
-    this->_executableNet = this->_ieCore.LoadNetwork (this->_networkCNN,
-        *strVectorIter);
-    this->isLoaded = true;
 #endif /* __OPENVINO_CPU_EXT__ */
```
dongju-chae: Q: Then, even if `__OPENVINO_CPU_EXT__` is not defined, is it regarded as a successful load?

ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
```diff
@@ -363,19 +437,71 @@ TensorFilterOpenvino::getOutputTensorDim (GstTensorsInfo * info)
...
- * @todo fill this fuction
  */
 static int
 ov_invoke (const GstTensorFilterProperties * prop, void **private_data,
     const GstTensorMemory * input, GstTensorMemory * output)
 {
+  TensorFilterOpenvino *tfOv =
+      static_cast < TensorFilterOpenvino * >(*private_data);
+
+  tfOv->invoke (prop, input, output);
```
dongju-chae: maybe `return tfOv->invoke (...)` ?

ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
wooksong: @dongju-chae A: In that case, by the lines from 559 to 563, this method, loadModel (), never be invoked.

ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
wooksong: Another mistake:(

wooksong:
#### V2
- Rebased onto 4f5852e5e09c0dad9e0855e32e3ba71318f36305

wooksong:
#### V3
- Rebased onto 01a7d2addfc0fec2ea4595a26bba90a946db0f8d

ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
```diff
@@ -363,20 +437,70 @@ TensorFilterOpenvino::getOutputTensorDim (GstTensorsInfo * info)
...
+  InferenceEngine::BlobMap inBlobMap;
+  InferenceEngine::BlobMap outBlobMap;
+  guint num_tensors;
+  guint i;
+
+  num_tensors = (prop->input_meta).num_tensors;
+  for (i = 0; i < num_tensors; ++i) {
+    const GstTensorInfo *info = &((prop->input_meta).info[i]);
+    InferenceEngine::Blob::Ptr blob = convertGstTensorMemoryToBlobPtr (
+        this->_inputTensorDescs[i], &(input[i]));
```
kparichay: Is omitting `.data` from `input[i]` is intentional? Both works, but wanted to know what is a good practice.

wooksong: @kparichay  Yes, it is intentional. In the case of a class that has a standard layout (i.e., C struct), it is legal. It had been regarded as a performance optimization technique (related to the elimination of unnecessary memory access blah blah, I am not sure). However, I think this is a bad habit right now. I will change it. Thanks.

wooksong: @kparichay In the previous comment, there is a misunderstanding about the code you pointed out. 
```
InferenceEngine::Blob::Ptr TensorFilterOpenvino::convertGstTensorMemoryToBlobPtr (
    const InferenceEngine::TensorDesc tensorDesc, const GstTensorMemory * gstTensor)
```
This is a static function taking ``` const GstTensorMemory * gstTensor```.

kparichay: aah, you are right
I thought it was directly passing the GstTensorMemory to make the blob

PR #1994, opened 2020-01-03
[CONF] override NNSTREAMER_CONF_FILE with env-var conf in non-Tizen

dongju-chae:
This PR overrides `NNSTREAMER_CONF_FILE` with env-var conf in
non-Tizen system. So, env-var conf has a higher priority than others.

With this feature, some unittests using a conf file from env-var has
passed even if the default conf file (`/etc/nnstremer.ini`) exists.
Note that Tizen does not allow to set configurations from env-var.

**Problem**
`unittest_common` has failed when `/etc/nnstreamer.ini` exists because it always has a higher priority than others. 

```
[----------] 1 test from conf_custom
[ RUN      ] conf_custom.env_str_01
../tests/common/unittest_common.cpp:786: Failure                                
  fn                                                                            
    Which is: NULL                                                              
  f1                                                                            
    Which is: "/tmp/nns-tizen-SV38C0/filters/libnnstreamer_filter_fantastic.so" 
...
```

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_conf.c
+++ b/gst/nnstreamer/nnstreamer_conf.c
@@ -316,26 +316,41 @@ nnsconf_loadconf (gboolean force_reload)
     /* init with 0 */
     memset (&conf, 0, sizeof (confdata));
   }
-
-  /* Read from the conf file first */
-  if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
-    conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
-  } else {
-    /** default value of 'sysconfdir' in meson is 'etc' */
-    conf.conffile = g_build_path (G_DIR_SEPARATOR_S, root_path_prefix,
-        NNSTREAMER_CONF_FILE, NULL);
-  }
-
+#ifndef __TIZEN__
+  /** if it's not Tizen, configuration from env-var has a higher priority */
+  conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
   if (!g_file_test (conf.conffile, G_FILE_TEST_IS_REGULAR)) {
-    /* File not found or not configured */
     g_free (conf.conffile);
-
-    /* Try to read from Environmental Variables */
-    conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
+    conf.conffile = NULL;
   }
+#endif
+  if (conf.conffile == NULL) {
+    /**
+     * Priority of reading a conf file
+     * 1) read from NNSTREAMER_CONF_FILE
+     * 2) read from NNSTREAMER_DEFAULT_CONF_FILE
+     * 3) read from env-var
+     */
+    if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
+      conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
+    } else {
+      /** default value of 'sysconfdir' in meson is 'etc' */
+      conf.conffile = g_build_path (G_DIR_SEPARATOR_S, root_path_prefix,
+          NNSTREAMER_CONF_FILE, NULL);
+    }
 
-  if (conf.conffile == NULL)
-    conf.conffile = g_strdup (NNSTREAMER_DEFAULT_CONF_FILE);
+    if (!g_file_test (conf.conffile, G_FILE_TEST_IS_REGULAR)) {
+      /* File not found or not configured */
+      g_free (conf.conffile);
+
+      if (g_file_test (NNSTREAMER_DEFAULT_CONF_FILE, G_FILE_TEST_IS_REGULAR)) {
+        conf.conffile = g_strdup (NNSTREAMER_DEFAULT_CONF_FILE);
+      } else {
+        /* Try to read from Environmental Variables */
+        conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
+      }
+    }
+  }
 
   g_assert (key_file != NULL);
   g_assert (conf.conffile != NULL);


jaeyun-jung:
security issue, security team did not allow the env-var in tizen.

dongju-chae:
Then, `env_str_01` is not executed in Tizen?

dongju-chae:
I added `#ifndef __TIZEN__` for the security issue in Tizen.

dongju-chae:
> security issue, security team did not allow the env-var in tizen.

@jaeyun-jung I think some existing unittests in Tizen rely on env var for conf (i.e., /home/abuild/.../nnstreamer-test.ini). It doesn't make sense if Tizen does not allow env var/

jaeyun-jung:
how about changing the priority

1. read from NNSTREAMER_CONF_FILE
2. read from NNSTREAMER_DEFAULT_CONF_FILE
3. read from env-var

```
  /* Read from the conf file first */
  if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
    conf.conffile = g_strdup (NNSTREAMER_CONF_FILE);
  } else {
    /** default value of 'sysconfdir' in meson is 'etc' */
    conf.conffile = g_build_path (G_DIR_SEPARATOR_S, root_path_prefix,
        NNSTREAMER_CONF_FILE, NULL);
  }

  if (!g_file_test (conf.conffile, G_FILE_TEST_IS_REGULAR)) {
    /* File not found or not configured */
    g_free (conf.conffile);

    if (g_file_test (NNSTREAMER_DEFAULT_CONF_FILE, G_FILE_TEST_IS_REGULAR)) {
      /* Try to read from default conf file */
      conf.conffile = g_strdup (NNSTREAMER_DEFAULT_CONF_FILE);
    } else {
      /* Try to read from Environmental Variables */
      conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
    }
  }

  g_assert (key_file != NULL);
  g_assert (conf.conffile != NULL);
```

jaeyun-jung:
Currently before the installation, nnstreamer cannot read ini file.
To run the unittest in the build process, it is necessary to read conf path from env-var.
After the installation, ini file always exists in tizen device.

dongju-chae:
Got it. I understood the situation. I will update this PR soon.

dongju-chae:
PR changed: add `loadconf_env()` instead of changing the priority of conf files

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -317,26 +316,23 @@ nnsconf_loadconf (gboolean force_reload)
...
 
-  /* Read from the conf file first */
-  if (g_path_is_absolute (NNSTREAMER_CONF_FILE)) {
+  /**
+   * Priority of reading a conf file
+   * 1) read from NNSTREAMER_CONF_FILE
+   * 2) read from NNSTREAMER_DEFAULT_CONF_FILE
+   * 3) read from env-var
+   */
+  if (g_path_is_absolute (NNSTREAMER_CONF_FILE) &&
```
jaeyun-jung: plz leave the option to check absolute path, because default value sysconfdirson is etc (not /etc).
if developer just configures `$ meson build`, this always refers NNSTREAMER_DEFAULT_CONF_FILE

dongju-chae: Got it

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -385,6 +392,89 @@ nnsconf_loadconf (gboolean force_reload)
   return TRUE;
 }
 
+/** @brief Public function defined in the header */
+gboolean
+nnsconf_loadconf_env (void)
```
jaeyun-jung: Suggestion: to reduce duplicated code, how about adding a param in old function?
```
gboolean
nnsconf_loadconf (gboolean force_reload, gboolean from_env)
```

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -330,13 +335,15 @@ nnsconf_loadconf (gboolean force_reload)
     /* File not found or not configured */
     g_free (conf.conffile);
 
-    /* Try to read from Environmental Variables */
-    conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
+    if (g_path_is_absolute (NNSTREAMER_DEFAULT_CONF_FILE) &&
```
jaeyun-jung: g_path_is_absolute (NNSTREAMER_DEFAULT_CONF_FILE) is always true

dongju-chae: Got it

gst/nnstreamer/nnstreamer_conf.c
dongju-chae: OK I will fix this

dongju-chae:
PR changed: add an extra argument for `nnsconf_loadconf()`.

myungjoo:
We already using test-only configurations for unit test. I don't think you need this modification for this purpose.

dongju-chae:
> We already using test-only configurations for unit test. I don't think you need this modification for this purpose.

Um, the purpose of this PR is to resolve situations where the test-only configuration does not work in a certain situation due to the priority of conf files.

myungjoo:
Let's allow to override NNSTREAMER_CONF_FILE with _strdup_getenv(NNSTREAMER_ENVVAR_CONF_FILE) if it is NOT Tizen. That's going to be more intuitive for other platforms.

dongju-chae:
PR changed: the comment from Myungjoo is applied.

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -316,26 +316,40 @@ nnsconf_loadconf (gboolean force_reload)
...
-        NNSTREAMER_CONF_FILE, NULL);
-  }
-
+#ifndef __TIZEN__
+  /** if it's not Tizen, configuration from env-var has a higher priority */
+  conf.conffile = _strdup_getenv (NNSTREAMER_ENVVAR_CONF_FILE);
   if (!g_file_test (conf.conffile, G_FILE_TEST_IS_REGULAR)) {
-    /* File not found or not configured */
     g_free (conf.conffile);
+#endif
```
myungjoo:
```
#ifndef __TIZEN__
  ...
  ...
    g_free (conf.conffile);
    conf.conffile = NULL;
  }
#endif
  if (conf.conffile == NULL) {
  ...
  }
```
Then, you may omit the later #ifdef/#endif, which in turn, improves the readability.

tests/common/unittest_common.cc
```diff
@@ -686,6 +686,7 @@ TEST (common_string_util, replace_str_02)
   g_free (result);
 }
 
+#ifndef __TIZEN__
```
myungjoo: You don't need this.
In Tizen, unittest is supposed to be executed in GBS-nnstreamer, which is supposd to have NO nnstreamer.ini at /etc.

gst/nnstreamer/nnstreamer_conf.c
dongju-chae: Applied. thanks

tests/common/unittest_common.cc
dongju-chae: Right. updated.

PR #1999, opened 2020-01-06
[Filter] Add an option to centralize checking the path of model files

wooksong:
Currently, each sub-plugin of the tensor filter should check the validity of the path of the model files. In general, it leads to unnecessary code duplication.  Moreover, sometimes those are not checked
at all. This patch adds an option to centralize checking the path of model files so that each sub-plugin is in charge of the validity of model files only when this option is set to TRUE.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -152,6 +152,43 @@ create_regex (const gchar ** enum_list, const gchar ** regex_utils)
 }
 
 /**
+ * @brief Verify validity of path for given model file if verify_model_path is set
+ * @param[in] priv Struct containing the common tensor-filter properties of the object
+ * @return TRUE if there is no error
+ */
+static inline gboolean
+verify_model_path (const GstTensorFilterPrivate * priv)
+{
+  const GstTensorFilterProperties *prop;
+  gboolean ret = TRUE;
+  int i;
+
+  if (priv == NULL)
+    return FALSE;
+
+  prop = &(priv->prop);
+
+  if (prop == NULL)
+    return FALSE;
+
+  if (g_strcmp0 (prop->fwname, "custom-easy") == 0)
+    return TRUE;
+
+  if ((prop->model_files != NULL) && (priv->fw->verify_model_path == TRUE)) {
+    for (i = 0; i < prop->num_models; i++) {
+      if (!g_file_test (prop->model_files[i], G_FILE_TEST_IS_REGULAR)) {
+        g_critical ("Cannot find the model file [%d]: %s\n",
+            i, prop->model_files[i]);
+        ret = FALSE;
+      }
+    }
+  }
+
+  return ret;
+}
+
+
+/**
  * @brief Validate filter sub-plugin's data.
  */
 static gboolean
@@ -459,22 +496,12 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
     case PROP_MODEL:
     {
       const gchar *model_files = g_value_get_string (value);
-      int idx;
 
       g_assert (model_files);
       gst_tensor_filter_parse_modelpaths_string (prop, model_files);
 
-      /* 'custom-easy' framework has a virtual model file */
-      if (g_strcmp0 (prop->fwname, "custom-easy") != 0) {
-        for (idx = 0; idx < prop->num_models; idx++) {
-          if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) {
-            g_critical ("Cannot find the model file [%d]: %s\n",
-                idx, prop->model_files[idx]);
-          }
-        }
-      }
-
       /* reload model if FW has been already opened */
+      /* TODO: need to define the behaviur according to priv->fw->verify_model_path */
       if (priv->prop.fw_opened && priv->is_updatable) {
         if (priv->fw && priv->fw->reloadModel) {
           if (priv->fw->reloadModel (&priv->prop, &priv->privateData) != 0) {
@@ -810,8 +837,11 @@ gst_tensor_filter_common_open_fw (GstTensorFilterPrivate * priv)
                   priv->prop.num_models > 0 && priv->prop.model_files[0])))
         return;
       /* 0 if successfully loaded. 1 if skipped (already loaded). */
-      if (priv->fw->open (&priv->prop, &priv->privateData) >= 0)
-        priv->prop.fw_opened = TRUE;
+      if (priv->fw->open (&priv->prop, &priv->privateData) >= 0) {
+        if (verify_model_path (priv)) {
+          priv->prop.fw_opened = TRUE;
+        }
+      }
     } else {
       priv->prop.fw_opened = TRUE;
     }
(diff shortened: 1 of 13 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -466,15 +466,20 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
-          if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) {
-            g_critical ("Cannot find the model file [%d]: %s\n",
-                idx, prop->model_files[idx]);
+        if (priv->fw->check_model_path_in_plugin != 0) {
+          /* In this case, the general framework should check the validities of the path of model files */
+          for (idx = 0; idx < prop->num_models; idx++) {
+            if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) {
+              g_critical ("Cannot find the model file [%d]: %s\n",
+                  idx, prop->model_files[idx]);
+              return FALSE;
```
dongju-chae: I'm not sure it's okay to return a value here. IMO, basically, `set_propery` does not has a return value, but in case of `common_set_property`, its return value is used to check `prop_id`'s validity. So, I think this function's doxygen comment (related to return) and line 298 in tensor_filter.c also need to be changed, to support this feature.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -169,6 +169,23 @@ static gboolean gst_tensor_filter_sink_event (GstBaseTransform * trans,
...
 
+/**
+ * @brief Check validity of path for given model file if check_model_path_in_nns is set
+ */
+#define gst_tensor_filter_check_model_path_if_necessary(priv,prop,ret) do { \
+  ret = 0; \
+  if (priv->fw->check_model_path_in_nns == TRUE) { \
+    int idx; \
+    for (idx = 0; idx < prop->num_models; idx++) { \
+      if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) { \
```
dongju-chae: Minor suggestion: how about strictly checking `model_files` like `prop->model_files != NULL`.
Otherwise, it might have potential dangers for a segmentation fault.

wooksong: At the point when the sub-plugin is loaded, prop->num_models is set to 0 by gst_tensor_filter_init. Although adding a null checking is not harmful at all, this seems to be a guard for the segmentation fault that would hardly happen.

dongju-chae: Sure, it's safe for normal users using a standard way to use `PROP_MODEL`. But, IMO, `.num_models` can be specified from a user (like some unittests), and someone may set this value without assigning `.model_files`. Of course, it's just my personal concerns :)

wooksong: @dongju-chae OK:) I will ease your concerns. Thanks!

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -459,22 +459,12 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
 
-      /* 'custom-easy' framework has a virtual model file */
-      if (g_strcmp0 (prop->fwname, "custom-easy") != 0) {
-        for (idx = 0; idx < prop->num_models; idx++) {
-          if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) {
-            g_critical ("Cannot find the model file [%d]: %s\n",
-                idx, prop->model_files[idx]);
-          }
-        }
-      }
```
myungjoo: We were checking file existences if it is not "custom-easy".
It appears that you are disabling it for other subplugins as well.
Please turn them on for subplugins that really use real files as models.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -169,6 +169,27 @@ static gboolean gst_tensor_filter_sink_event (GstBaseTransform * trans,
...
+    } \
+    for (idx = 0; idx < prop->num_models; idx++) { \
+      if (!g_file_test (prop->model_files[idx], G_FILE_TEST_IS_REGULAR)) { \
+        g_critical ("Cannot find the model file [%d]: %s\n", \
+            idx, prop->model_files[idx]); \
+        ret = -1; \
+      } \
+    } \
+  } \
+} while (0);
```
myungjoo: Do you need to make this a macro?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
dongju-chae: IMO, to resolve this issue, it's required to set `check_model_path_in_nns = FALSE` for "custom-easy" framework.
In this manner, we can handle all frameworks using virtual files using `check_model_path_in_nns = FALSE`

myungjoo: Because except for custom-easy, they need to check model_path, they should set it TRUE, but @wooksong set them all FALSE.

Anyway, why don't we change ```check_model_path_in_nns``` to ```virtual_model_names```, which skips path checking if it is TRUE and set TRUE for custom-easy. Then, we can keep checking paths by default.

dongju-chae: Ahh.. understood. I also agree with your opinion.

wooksong: > virtual_model_paths

There are three cases for model checking:
1. virtual model path
2. normal case (which could be simply handled by checking in NNS)
3. I-want-to-check-these-in-my-plugin case (For example, openvino requires two model files, A.xml and A.bin. For the convenience, I make the model properties can take 1) A.xml and A.bin
2) A.xml and B.bin 3) just A)

This is why I am using the name, check_model_path_in_nns, in my PR.

In the first PR, I just set check_model_path_in_nns to false for all the sub-plugins (to keep the existing behavior). And then, check_model_path_in_nns for each sub-plugin is set to TRUE one by one after little investigating the sub-plugin source code.

wooksong:
#### V2
- Add static inline function instead of macro
- Place the checking function at the right after the framework is opened

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: To keep the existing behavior, you need to check files for all subplugins except for custom-easy.

myungjoo: And if that case, ```veify_model_paths``` should be enough. (if_necessary is not necessary)

wooksong: > To keep the existing behavior, you need to check files for all subplugins except for custom-easy.

Currently, all sub plugins have their mechanisms.

wooksong: > And, `verify_model_paths` should be enough. (if_necessary is not necessary)

Oh, that is good idea.

wooksong:
#### V3
- Rename the flag and function names

Issue #1986, opened 2019-12-31
Data layout information needed


kparichay:
An input image in tensor format of size HxWxC can be provided as an input in HxWxC or CxHxW.
These are the most common of the two formats used in ML field (corresponding to NHWC or NCHW format). Generally, NCHW is preferred by GPU backend while NHWC is preferred for CPU backend for performance purposes.

Need to add API to take this information from user, and pass to extensions, if extensions support it (at least nnfw-runtime and armnn (as per the sample example) support it).

jaeyun-jung:
In the first version of sub-plugin SNAP, I simply parsed custom string.
`tensor_filter framework=snap custom=ModelFWType:CAFFE,...,InputDataFormat:NHWC`

kparichay:
If multiple sub-plugins are using this property (nnfw, snap, armnn), should this be added a property in tensor_filter?

myungjoo:
Another constraints or conditions we can consider include:
1. We may keep ourselves totally agnostic to data layout or transparent to data layout (it's something between app developers and model developers?)
2. We handle not only vision models, but also sound/speech models or even models with pure numbers without physical dimensions (tensor outputs of other models). Data layout of w/h/c is very awkward thing; we can consider them simply as 1st, 2nd, and 3rd dimension. Then, this data layout becomes some additional syntactic sugar that we may provide, which simply represents a "conditional tensor_transform".

myungjoo:
If a framework (I hear that some H/W frameworks do) mandates the data layout of input images, it should be handled by pipeline writers (app writers).

Anyway, Single API is the pipeline writer; thus, for Single APIs, we may consider supporting data layouts natively.

kparichay:
> Another constraints or conditions we can consider include:
> 
> 1. We may keep ourselves totally agnostic to data layout or transparent to data layout (it's something between app developers and model developers?)

Yes, nnstreamer does not need to consider these layout for itself. It is to be passed to the framework.

> 2. We handle not only vision models, but also sound/speech models or even models with pure numbers without physical dimensions (tensor outputs of other models). Data layout of w/h/c is very awkward thing; we can consider them simply as 1st, 2nd, and 3rd dimension. Then, this data layout becomes some additional syntactic sugar that we may provide, which simply represents a "conditional tensor_transform".

Yes, this information will be optional from the perspective of nnstreamer.

myungjoo:
[...]

PR #2007, opened 2020-01-08
[tensor] Support data layout for tensor @open sesame 1/16 16:12

kparichay:
Data layout added for other/tensor
Three data layouts are supported
- NHWC - referred as channels last
- NCHW - referred as channels first
- none - unknown or doesnt matter
These data layouts can be different for each other/tensor in other/tensors
Also added their basic support in single API

Refer #1986 

**Self evaluation:**

### DIFF
--- a/api/capi/include/nnstreamer.h
+++ b/api/capi/include/nnstreamer.h
@@ -169,6 +169,17 @@ typedef enum _ml_tensor_type_e
 } ml_tensor_type_e;
 
 /**
+ * @brief Possible data format/layout of Tensor in NNStreamer.
+ * @since_tizen 6.0
+ */
+typedef enum _ml_tensor_layout_e
+{
+  ML_TENSOR_LAYOUT_NONE = 0,      /**< It has unknown layout or it does not care about the data layout */
+  ML_TENSOR_LAYOUT_NHWC,          /**< Channel first layout */
+  ML_TENSOR_LAYOUT_NCHW,          /**< Channel last layout */
+} ml_tensor_layout_e;
+
+/**
  * @brief Enumeration for the error codes of NNStreamer Pipelines.
  * @since_tizen 5.5
  */
--- a/gst/nnstreamer/tensor_common.c
+++ b/gst/nnstreamer/tensor_common.c
@@ -108,6 +108,7 @@ gst_tensor_info_init (GstTensorInfo * info)
 
   info->name = NULL;
   info->type = _NNS_END;
+  info->layout = _NNS_LAYOUT_NONE;
 
   for (i = 0; i < NNS_TENSOR_RANK_LIMIT; i++) {
     info->dimension[i] = 0;
@@ -183,6 +184,10 @@ gst_tensor_info_is_equal (const GstTensorInfo * i1, const GstTensorInfo * i2)
     return FALSE;
   }
 
+  if (i1->layout != i2->layout) {
+    return FALSE;
+  }
+
   for (i = 0; i < NNS_TENSOR_RANK_LIMIT; i++) {
     if (i1->dimension[i] != i2->dimension[i]) {
       return FALSE;
@@ -208,6 +213,7 @@ gst_tensor_info_copy_n (GstTensorInfo * dest, const GstTensorInfo * src,
 
   dest->name = (src->name) ? g_strdup (src->name) : NULL;
   dest->type = src->type;
+  dest->layout = src->layout;
 
   for (i = 0; i < n; i++) {
     dest->dimension[i] = src->dimension[i];
@@ -423,6 +429,44 @@ gst_tensors_info_parse_types_string (GstTensorsInfo * info,
   }
 
   return num_types;
+}
+
+/**
+ * @brief Parse the string of layout
+ * @param info tensors info structure
+ * @param layout_string string of layout
+ * @return number of parsed layouts
+ */
+guint
+gst_tensors_info_parse_layouts_string (GstTensorsInfo * info,
+    const gchar * layout_string)
+{
+  guint num_layouts = 0;
+
+  g_return_val_if_fail (info != NULL, 0);
+
+  if (layout_string) {
+    guint i;
+    gchar **str_layouts;
+
+    str_layouts = g_strsplit_set (layout_string, ",.", -1);
+    num_layouts = g_strv_length (str_layouts);
+
+    if (num_layouts > NNS_TENSOR_SIZE_LIMIT) {
+      GST_WARNING ("Invalid param, layouts (%d) max (%d)\n",
+          num_layouts, NNS_TENSOR_SIZE_LIMIT);
+
+      num_layouts = NNS_TENSOR_SIZE_LIMIT;
+    }
+
+    for (i = 0; i < num_layouts; i++) {
+      info->info[i].layout = gst_tensor_get_layout (str_layouts[i]);
+    }
+
+    g_strfreev (str_layouts);
+  }
+
+  return num_layouts;
 }
 
 /**
@@ -974,6 +1018,43 @@ gsize
 gst_tensor_get_element_size (tensor_type type)
 {
   return tensor_element_size[type];
+}
+
+/**
+ * @brief Get tensor layout from string input.
+ * @return Corresponding tensor_layout.
+ */
+tensor_layout
+gst_tensor_get_layout (const gchar * layoutstr)
+{
+  gsize len;
+  gchar *layout_string;
+  tensor_layout layout = _NNS_LAYOUT_NONE;
+
+  if (layoutstr == NULL)
+    return layout;
+
+  /* remove spaces */
+  layout_string = g_strdup (layoutstr);
+  g_strstrip (layout_string);
+
+  len = strlen (layout_string);
+
+  if (len == 0) {
+    g_free (layout_string);
+    return layout;
+  }
+
+  if (g_ascii_strcasecmp (layoutstr, "NCHW") == 0) {
+    layout = _NNS_LAYOUT_NCHW;
+  } else if (g_ascii_strcasecmp (layoutstr, "NHWC") == 0) {
+    layout = _NNS_LAYOUT_NHWC;
+  } else {
+    GST_WARNING ("Invalid layout, defaulting to none layout.");
+  }
+
+  g_free (layout_string);
+  return layout;
 }
 
 /**
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -80,9 +80,11 @@ enum
   PROP_INPUT,
   PROP_INPUTTYPE,
   PROP_INPUTNAME,
+  PROP_INPUTLAYOUT,
   PROP_OUTPUT,
   PROP_OUTPUTTYPE,
   PROP_OUTPUTNAME,
+  PROP_OUTPUTLAYOUT,
   PROP_CUSTOM,
   PROP_SUBPLUGINS,
   PROP_ACCELERATOR,
@@ -323,6 +325,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
       g_param_spec_string ("inputtype", "Input tensor element type",
           "Type of each element of the input tensor ?", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INPUTLAYOUT,
+      g_param_spec_string ("inputlayout", "Input Data Layout",
+          "Set channel first (NCHW) or channel last layout (NHWC) for input data. "
+          "Layout of the data can be NHWC or NCHW or none for now. ",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
   g_object_class_install_property (gobject_class, PROP_OUTPUTNAME,
       g_param_spec_string ("outputname", "Name of Output Tensor",
           "The Name of Output Tensor", "",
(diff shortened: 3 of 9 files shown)


kparichay:
Note: Input and output layout can be set independently. This is because inputs in other/tensors can have different layouts. Hence, output layout cannot be inferred from input layouts (number of output tensors can be different from input tensors).

gst/nnstreamer/tensor_common.c
```diff
@@ -976,6 +1023,43 @@ gst_tensor_get_element_size (tensor_type type)
...
+  len = strlen (layout_string);
+
+  if (len == 0) {
+    g_free (layout_string);
+    return layout;
+  }
+
+  if (g_ascii_strncasecmp (layoutstr, "NCHW", 4) == 0) {
+    layout = _NCHW;
+  } else if (g_ascii_strncasecmp (layoutstr, "NCHW", 4) == 0) {
```
dongju-chae: `NCHW` --> `NHWC`

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -323,6 +325,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
       g_param_spec_string ("inputtype", "Input tensor element type",
           "Type of each element of the input tensor ?", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INPUTLAYOUT,
+      g_param_spec_string ("inputlayout", "Input Data Layout",
+          "Set channel first (NCHW) or channel last layout (NHWC). "
```
dongju-chae: It will be better if each description has extra info whether it's for either input or output.

gst/nnstreamer/tensor_typedef.h
```diff
@@ -155,6 +166,7 @@ typedef struct
                    and some (tensorflow-lite) do not need this. */
   tensor_type type; /**< Type of each element in the tensor. User must designate this. */
   tensor_dim dimension; /**< Dimension. We support up to 4th ranks.  */
+  tensor_layout layout; /**< Layout. Channel first or channel last layout */
```
dongju-chae: it can be `none_layout` as well.

gst/nnstreamer/tensor_common.c
```diff
@@ -976,6 +1023,43 @@ gst_tensor_get_element_size (tensor_type type)
...
+  g_strstrip (layout_string);
+
+  len = strlen (layout_string);
+
+  if (len == 0) {
+    g_free (layout_string);
+    return layout;
+  }
+
+  if (g_ascii_strncasecmp (layoutstr, "NCHW", 4) == 0) {
```
jaeyun-jung: just use g_ascii_strcasecmp()?

kparichay: updated.

gst/nnstreamer/tensor_typedef.h
kparichay: updated.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: updated.

gst/nnstreamer/tensor_common.c
kparichay: updated.

kparichay:
V2:
Applied suggestion from @dongju-chae and @jaeyun-jung

gst/nnstreamer/tensor_typedef.h
```diff
@@ -133,6 +133,17 @@ typedef union {
...
+/**
+ * @brief Internal tensor layout format for other/tensor
+ */
+typedef enum _nns_tensor_layout
+{
+  _NHWC = 0,    /**< NHWC: channel last layout */
+  _NCHW,        /**< NCHW: channel first layout */
+
+  _none_layout,        /**< Layout unknown or not specified */
+} tensor_layout;
```
dongju-chae: I think it's better for this enum to also have a similar format with 
```
typedef enum _ml_tensor_layout_e
{
  ML_TENSOR_LAYOUT_NHWC = 0,      /**< Channel first layout */
  ML_TENSOR_LAYOUT_NCHW,          /**< Channel last layout */
  ML_TENSOR_LAYOUT_NONE,          /**< Unknown layout */
}
```
For example, like `NNS_TENSOR_LAYOUT_NHWC/NCHW/NONE`?

kparichay: yes, that's better - updated to `_NNS_LAYOUT_*`

api/capi/include/nnstreamer.h
```diff
@@ -168,6 +168,17 @@ typedef enum _ml_tensor_type_e
   ML_TENSOR_TYPE_UNKNOWN         /**< Unknown type */
 } ml_tensor_type_e;
 
+/**
+ * @brief Possible data element types of Tensor in NNStreamer.
+ * @since_tizen 5.5
```
jaeyun-jung: maybe tizen 6.0

gst/nnstreamer/tensor_common.c
```diff
@@ -668,6 +713,7 @@ gst_tensor_config_from_structure (GstTensorConfig * config,
  * @brief Get tensor caps from tensor config
  * @param config tensor config info
  * @return caps for given config
+ * @todo Add layout
```
jaeyun-jung: Discussion: About caps negotiation, is it necessary to indicate data format in caps string?
I think the data format is used only in tensor-filter now, for invoking specific nn-framework like nnfw and snap.

kparichay: In a pipeline with multiple tensor filters (of nnfw or snap), I think having data format/layout in caps might be helpful. It would stop the connection between two tensor filters with same dimensions but different data format (which I think is desirable).
For caps negotiation between data layout/formats, I think the only 1 pair will mis-match - `(NCHW, NHWC)`. All other pairings (even with NONE layout as NONE layout represents `dont care` or `unknown layout`) should match. (This isn't implemented yet). Please provide your input on this.

api/capi/include/nnstreamer.h
kparichay: Updated.

kparichay:
V3:
Added in layout in nnstreamer_examples to avoid getting stuck in some unittests
Tizen 5.5 -> 6.0

gst/nnstreamer/tensor_common.c
jaeyun-jung: I am not sure to do the negotiation with new field 'layout' is possible. (field-type is string and it should be exactly matched)
@helloahn @again4you IMO, this may require huge change in Tizen ACR process. Please see the details.

kparichay: @jaeyun-jung Yes, negotiation with string should match exactly. Removed layout from caps.

api/capi/include/nnstreamer.h
```diff
@@ -168,6 +168,17 @@ typedef enum _ml_tensor_type_e
   ML_TENSOR_TYPE_UNKNOWN         /**< Unknown type */
 } ml_tensor_type_e;
 
+/**
+ * @brief Possible data element types of Tensor in NNStreamer.
```
jaeyun-jung: maybe `data format of tensor`

kparichay: updated.

jaeyun-jung:
Question: will you update/add the related functions in c-api?
e.g., ml_tensors_info_get_layout() ml_tensors_info_set_layout() ml_tensors_info_copy_from_ml()
ml_tensors_info_copy_from_gst() ml_tensors_info_clone()

kparichay:
> Question: will you update/add the related functions in c-api?
> e.g., ml_tensors_info_get_layout() ml_tensors_info_set_layout() ml_tensors_info_copy_from_ml()
> ml_tensors_info_copy_from_gst() ml_tensors_info_clone()

Yes, that I will do in next PR once this is approved.

api/capi/include/nnstreamer.h
```diff
@@ -168,6 +168,17 @@ typedef enum _ml_tensor_type_e
...
 
+/**
+ * @brief Possible data format/layout of Tensor in NNStreamer.
+ * @since_tizen 6.0
+ */
+typedef enum _ml_tensor_layout_e
+{
+  ML_TENSOR_LAYOUT_NHWC = 0,      /**< Channel first layout */
+  ML_TENSOR_LAYOUT_NCHW,          /**< Channel last layout */
+  ML_TENSOR_LAYOUT_NONE,          /**< Unknown layout */
```
myungjoo: Set ```ML_TENSOR_LAYOUT_NONE = 0``` as it is the default value.
```s/Unknown layout/It has unknown layout or it does not care the data layout./```

File: gst/nnstreamer/tensor_filter/tensor_filter_common.c [hunk 10]
```diff
@@ -323,6 +325,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
       g_param_spec_string ("inputtype", "Input tensor element type",
           "Type of each element of the input tensor ?", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INPUTLAYOUT,
+      g_param_spec_string ("inputlayout", "Input Data Layout",
+          "Set channel first (NCHW) or channel last layout (NHWC) for input data. "
+          "Layout of the data can be NHWC or NCHW or none for now. ",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
```
myungjoo: Use ENUM property. Manage default value properly.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -335,6 +342,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
       g_param_spec_string ("outputtype", "Output tensor element type",
           "Type of each element of the output tensor ?", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_OUTPUTLAYOUT,
+      g_param_spec_string ("outputlayout", "Output Data Layout",
+          "Set channel first (NCHW) or channel last layout (NHWC) for output data. "
+          "Layout of the data can be NHWC or NCHW or none for now. ",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
```
myungjoo: Same here.

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -890,6 +890,7 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
       int i;
       gchar **strv = NULL;
 
+      /** TODO: update layout here */
```
myungjoo: With tensor_transform, we may accept (as syntatic suger) ops as: "data-layout-change-to-whs" (this is just a verbose example. the real syntax should be simpler)

api/capi/include/nnstreamer.h
myungjoo: And be aware, you are modifying the public API :)

There will be a lot of follow-up "paper work"

nnstreamer_example/custom_example_passthrough/nnstreamer_customfilter_example_passthrough.c
```diff
@@ -51,6 +51,7 @@ pt_init (const GstTensorFilterProperties * prop)
   data->info.dimension[2] = D3;
   data->info.dimension[3] = D4;
   data->info.type = _NNS_UINT8;
+  data->info.layout = _NNS_LAYOUT_NONE;
```
myungjoo: This should be always "NONE" if not specified. (specifying it is not incorrect. forcing us to specifying is wrong)

gst/nnstreamer/tensor_common.c
```diff
@@ -183,6 +184,10 @@ gst_tensor_info_is_equal (const GstTensorInfo * i1, const GstTensorInfo * i2)
     return FALSE;
   }
 
+  if (i1->layout != i2->layout) {
```
jaeyun-jung: Compare the layouts :
This will make caps-negotiation failure when incoming meta (usually non-layout) is different with the properties in tensor-filter.

kparichay: How about if any of the layout is `none`, then comparison is not performed?

api/capi/include/nnstreamer.h
kparichay: Updated.

nnstreamer_example/custom_example_passthrough/nnstreamer_customfilter_example_passthrough.c
kparichay: Done.

kparichay:
As @jaeyun-jung [pointed out](https://github.com/nnsuite/nnstreamer/pull/2007#discussion_r364535999), keeping layout for caps negotiation will not work. In the current scenario, it seems keeping that keeping the layout for just the tensor_filter is sufficient. So, I will move data layout property to GstTensorFilterProperties.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: In the current implementation using STRING property, `inputlayout` can give a different for each input. However, with ENUM, I am not sure if this is possible.

gst/nnstreamer/tensor_common.c
kparichay: As discussed with @myungjoo , there will be 4 layouts, `ANY`, `NONE`, `NHWC` and `NCHW`. `ANY` can match with all, however others are specific. I will update this PR as per this.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Going to use enum - all the tensors in `other\tensors` will be forced to have the same layout for now.

myungjoo:
Refer to #2050 for more info.

kparichay:
Closing this as this discussion is now in #2051, and this implementation cannot work as non-invasive.

PR #2013, opened 2020-01-09
[subplugin/ext] Extensions not update their functions

kparichay:
NNStreamer extensions for tensor filter update their functions based on the model file
However, this behavior is not thread-safe.
Updated for extensions to not update their functions (for destroyNotify and allocate_in_invoke)

For allocate_in_invoke, tensor_filter needs to know based on the model if the allocate_in_invoke is supported or not
So added another interface in GstTensorFilterFramework - allocateInInvoke

~destroyNotify interface is also updated. private_data for the subplugin is passed embedded in the data field so that
based on the private data, appropriate model specific destroy function for custom filter can be called~

V2:
Added allocate_in_invoke to be set properly in single API implementation of tensor_filter
Also added some minor updates in comment
Added minor bugfix in python extension

PART-2
Python updates in extensions framework functions
This results in non-thread-safe behavior
This PR modifies the corresponding implementation

PART-3
As extensions are supposed to be thread-safe, using class static variables without locks is not stable
Move class static variables to class local object variables for python, tensorflow and caffe2
Also update corresponding usages.

Resolves : #2034

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
@@ -714,7 +714,8 @@ tf_getOutputDim (const GstTensorFilterProperties * prop, void **private_data,
 
 /**
  * @brief The optional callback for GstTensorFilterFramework
- * @param[in] data The data element.
+ * @param data : The data element.
+ * @param private_data : tensorflow plugin's private data
  */
 static void
 tf_destroyNotify (void **private_data, void *data)
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
@@ -74,8 +74,7 @@ public:
   int getInputTensorDim (GstTensorsInfo * info);
   int getOutputTensorDim (GstTensorsInfo * info);
   int run (const GstTensorMemory * input, GstTensorMemory * output);
-
-  static std::map < void *, TF_Tensor * >outputTensorMap;
+  void freeOutputTensor (void *data);
 
 private:
 
@@ -85,6 +84,7 @@ private:
   GstTensorsInfo outputTensorMeta;  /**< The tensor info of output tensors from user input */
 
   std::vector < tf_tensor_info_s > input_tensor_info; /* hold information for TF */
+  std::map < void *, TF_Tensor * > outputTensorMap;
 
   TF_Graph *graph;
   TF_Session *session;
@@ -96,8 +96,6 @@ private:
 
 void init_filter_tf (void) __attribute__ ((constructor));
 void fini_filter_tf (void) __attribute__ ((destructor));
-
-std::map <void*, TF_Tensor*> TFCore::outputTensorMap;
 
 /**
  * @brief	TFCore creator
@@ -590,6 +588,18 @@ failed:
 }
 
 /**
+ * @brief free output tensor corresponding to the given data
+ * @param[in] data The data element
+ */
+void
+TFCore::freeOutputTensor (void *data)
+{
+  TF_Tensor *output_tensor = outputTensorMap.find (data)->second;
+  TF_DeleteTensor (output_tensor);
+  outputTensorMap.erase (data);
+}
+
+/**
  * @brief Free privateData and move on.
  */
 static void
@@ -720,8 +730,11 @@ tf_getOutputDim (const GstTensorFilterProperties * prop, void **private_data,
 static void
 tf_destroyNotify (void **private_data, void *data)
 {
-  TF_DeleteTensor ( (TFCore::outputTensorMap.find (data))->second);
-  TFCore::outputTensorMap.erase (data);
+  TFCore *core = static_cast<TFCore *>(*private_data);
+
+  if (core) {
+    core->freeOutputTensor (data);
+  }
 }
 
 static gchar filter_subplugin_tensorflow[] = "tensorflow";
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -360,11 +360,11 @@ typedef struct _GstTensorFilterFramework
        */
 
        int (*allocateInInvoke) (void **private_data);
-			 /**< Optional. tensor_filter.c will call it when allocate_in_invoke is set to TRUE. This check if the provided model for the framework supports allocation at invoke or not. If this is not defined, then the value of allocate_in_invoke is assumed to be final for all models.
-				*
-				* @param[in] private_data A subplugin may save its internal private data here.
-				* @return 0 if supported. -errno if not supported.
-				*/
+       /**< Optional. tensor_filter.c will call it when allocate_in_invoke is set to TRUE. This check if the provided model for the framework supports allocation at invoke or not. If this is not defined, then the value of allocate_in_invoke is assumed to be final for all models.
+        *
+        * @param[in] private_data A subplugin may save its internal private data here.
+        * @return 0 if supported. -errno if not supported.
+        */
     };
 
     /**
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -357,6 +357,7 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
   GstTensorMemory out_tensors[NNS_TENSOR_SIZE_LIMIT];
   guint i;
   gint ret;
+  gboolean allocate_in_invoke;
 
   self = GST_TENSOR_FILTER_CAST (trans);
   priv = &self->priv;
@@ -376,6 +377,7 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
   /* 0. Check all properties. */
   silent_debug ("Invoking %s with %s model\n", priv->fw->name,
       GST_STR_NULL (prop->model_files[0]));
+  allocate_in_invoke = gst_tensor_filter_allocate_in_invoke (priv);
 
   /* 1. Set input tensors from inbuf. */
   g_assert (gst_buffer_n_memory (inbuf) == prop->input_meta.num_tensors);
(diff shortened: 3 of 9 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_custom.c
```diff
@@ -132,9 +132,18 @@ custom_open (const GstTensorFilterProperties * prop, void **private_data)
...
   if (ptr->methods->allocate_invoke) {
-    NNS_support_custom.allocate_in_invoke = TRUE;
-    NNS_support_custom.destroyNotify = ptr->methods->destroy_notify;
+    filter_framework->allocate_in_invoke = TRUE;
+    filter_framework->destroyNotify = ptr->methods->destroy_notify;
   }
+
+  if (!ptr->methods->getInputDim) {
+    filter_framework->getInputDimension = NULL;
+  } else if (!ptr->methods->getOutputDim) {
```
jaeyun-jung: Question: plz check cond to check filter-callback, 'else if'

jaeyun-jung: also, plz check multi-custom filters (PR #2029 )

kparichay: Thanks for pointing out the mistake.
I will remove the changes in this PR, and modify it to remove existing errors like the one you pointed out.

kparichay: @jaeyun-jung Updated some of the implementations, and added the required interfaces to update this. Please give an initial review.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -393,12 +416,18 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
 
   /* 4. Update result and free map info. */
   for (i = 0; i < prop->output_meta.num_tensors; i++) {
-    if (priv->fw->allocate_in_invoke) {
+    if (allocate_in_invoke) {
+      GDestroyNotify destroy_notify = priv->fw->destroyNotify;
```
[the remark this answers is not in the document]
kparichay: Cannot add another param in destroy-notify callback - `gst_memory_new_wrapped` destroy function (`GDestroyNotify`) only allows one argument for the function - https://developer.gnome.org/gstreamer/stable/GstAllocator.html#gst-memory-new-wrapped

Hence, in the current implementation, privateData is passed embedded in GPtrArray.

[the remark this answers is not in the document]
kparichay: Will update.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
```diff
@@ -75,7 +75,7 @@ class TFCore
   int getOutputTensorDim (GstTensorsInfo * info);
   int run (const GstTensorMemory * input, GstTensorMemory * output);
 
-  static std::map < void *, TF_Tensor * >outputTensorMap;
+  std::map < void *, TF_Tensor * >outputTensorMap;
```
jaeyun-jung: let's try to hide this as private value, if we decide to change the params in destroy callback.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -219,6 +219,13 @@ typedef struct _GstTensorFilterFramework
        * @param[in] hw backend accelerator hardware
        * @return 0 if supported. -errno if not supported.
        */
+
+  int (*allocateInInvoke) (void **private_data);
```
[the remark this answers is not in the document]
kparichay: Either this will have to be made mandatory, or make an assumption that undefined function implies allocate in invoke is not supported for this framework.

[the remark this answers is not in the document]
jaeyun-jung: IMO, optional. If not defined, allocate-invoke is false and tensor-filter main will allocate the output buffer.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
kparichay: Done.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -199,10 +199,10 @@ typedef struct _GstTensorFilterFramework
...
+  void (*destroyNotify) (void * data, void **private_data);
+      /**< Optional. tensor_filter.c will call it when 'allocate_in_invoke' flag of the framework is TRUE and allocateInInvoke also returns enabled. Basically, it is called when the data element is destroyed. If it's set as NULL, g_free() will be used as a default. It will be helpful when the data pointer is included as an object of a nnfw. For instance, if the data pointer is removed when the object is gone, it occurs error. In this case, the objects should be maintained for a while first and destroyed when the data pointer is destroyed. Those kinds of logic could be defined at this method.
        *
-       * @param[in] data the data element.
+       * @param[in] data the data element and the private data. GPtrArray is passed where first element is the data element, and the second element is the subplugin's internal private data.
```
jaeyun-jung: plz update the description

gst/nnstreamer/nnstreamer_plugin_api_filter.h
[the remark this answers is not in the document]
myungjoo: Yup. That's doable. However, we now need to think about backward compatibility. We have private plugins that uses Tizen 5.5 and should be compatible with Tizen 6.0.

Choice1: Create data structure to support old plugins (w/o recompiling?)
Choice2: Some macro for major changes in filter subplugin data structure?
(e.g., #define NNS_FILTER_API (1) ?)

myungjoo:
As long as ```allocate_in_invoke``` is written inside "open", there is no thread-safety issue.
For the backward compatibility, let's try not to make changes in the API data structure except for adding a new method in the end of struct.

If we really need updates in the data structure, we need to resolve #2045 first.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: updated.

kparichay:
`allocate_in_invoke` can be kept for backward compatibility. 

> As long as allocate_in_invoke is written inside "open", there is no thread-safety issue.

However, I think changing `allocate_in_invoke` (even in "open") will interfere with thread safety, as `allocate_in_invoke` variable is shared by all the instances of the sub-plugin. The variable value for custom filter is dependent on model being passed to the tensor_filter, and there can be multiple instances of the filter.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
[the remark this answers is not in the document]
kparichay: I will not remove `allocate_in_invoke` till #2045 is resolved.

myungjoo:
> However, I think changing `allocate_in_invoke` (even in "open") will interfere with thread safety, as `allocate_in_invoke` variable is shared by all the instances of the sub-plugin. The variable value for custom filter is dependent on model being passed to the tensor_filter, and there can be multiple instances of the filter.

Ah.. that's the valid point.
Anyway, we still need to consider the backward compatibility, we have already released nnstreamer 1.0 with Tizen 5.5. We need to keep the ".tpk" apps with Tizen 5.5 keep working.

kparichay:
Currently `allocate_in_invoke` is kept for backward compatibility.

> One fortunate aspect is that we have not released nnstreamer filter subplugin APIs as Tizen APIs, yet. We can setup backward-compatibility mechanism/policy right now and keep moving. (e.g., adding version/sanity check uint64_t at the first word of the struct, consider it as "Version 0" if it does not exist (the current version))

Will add as a new PR?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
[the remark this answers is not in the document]
kparichay: @jaeyun-jung as per [discussion](https://github.com/nnsuite/nnstreamer/pull/2013#issuecomment-576140354), not remove `allocate_in_invoke` flag for now.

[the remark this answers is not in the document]
jaeyun-jung: Got it. Now we need to consider the changes of destroyNotify callback for the backward compatibility.

myungjoo:
> Currently `allocate_in_invoke` is kept for backward compatibility.
> 
> > One fortunate aspect is that we have not released nnstreamer filter subplugin APIs as Tizen APIs, yet. We can setup backward-compatibility mechanism/policy right now and keep moving. (e.g., adding version/sanity check uint64_t at the first word of the struct, consider it as "Version 0" if it does not exist (the current version))
> 
> Will add as a new PR?

You may do this in this PR or make a new PR. Anyway, the change for versioning should be made before this change (adding / modifying the subplugin API struct)

myungjoo:
Because we do not have opened subplugins via Tizen public APIs and have no knowledge on third party subplugins, yet, let's update this API along with our private subplugins "at the same time" and implement versioning feature later. Anyway, before merging this change, let's check if this "version" will survive long enough. (including adding "allocate-in-invoke callback")

gst/nnstreamer/nnstreamer_plugin_api_filter.h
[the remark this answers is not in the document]
myungjoo: Or... possibly...  assuming that we can ignore backward binary compatibility for now, we can remove the flag and define NULL of allocateininvoke callback means it's FALSE.

myungjoo:
Let's finalize the related policies with a short meeting with concerned people next week. @kparichay : please setup a meeting

kparichay:
> Duplicated by #2072 . Rebase after #2072 to reimplement destroyNotify of Caffe2.

I will merge this parts of this commit with #2072, as the interface of `destroyNotify` of version 0 should be updated in #2072. And then rebase this is #2072.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -416,6 +416,27 @@ gst_tensor_filter_parse_modelpaths_string (GstTensorFilterProperties * prop,
...
+ * @brief check if the allocate_in_invoke is valid for the framework
+ * @param[in] priv Struct containing the properties of the object
+ * @return TRUE if valid, FALSE on error
+ */
+gboolean
+gst_tensor_filter_allocate_in_invoke (GstTensorFilterPrivate * priv)
+{
+  int allocate_in_invoke;
+
+  allocate_in_invoke = priv->fw->allocate_in_invoke;
```
dongju-chae: Hmm, I think `allocate_in_invoke` is now use-less because `allocateInInvoke` may decide this (i.e., NULL or not)

myungjoo: Nah.. this is for the backward compatibility. With V1 API, both becomes useless :)

dongju-chae: Oh got it

PR #2016, opened 2020-01-09
[pytorch] Enable setting accelerators @open sesame 1/15 16:20

kparichay:
Enabled setting of accelerators
Enabled getting supported accelerators

See also #1849 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
@@ -43,6 +43,14 @@
 #define INPUT_TENSOR_META_CHAR "InputTensorMeta"
 #define OUTPUT_TENSOR_META_CHAR "OutputTensorMeta"
 
+static const gchar *torch_accl_support[] = {
+  ACCL_AUTO_STR,
+  ACCL_DEFAULT_STR,
+  ACCL_CPU_STR,
+  ACCL_GPU_STR,
+  NULL
+};
+
 /**
  * @brief	ring cache structure
  */
@@ -52,8 +60,7 @@ public:
   TorchCore (const char *_model_path);
   ~TorchCore ();
 
-  int init (const GstTensorFilterProperties * prop,
-      const gboolean torch_use_gpu);
+  int init (const GstTensorFilterProperties * prop);
   int loadModel ();
   const char *getModelPath ();
   int getInputTensorDim (GstTensorsInfo * info);
@@ -64,6 +71,7 @@ private:
 
   char *model_path;
   bool use_gpu;
+  accl_hw accelerator;
 
   GstTensorsInfo inputTensorMeta;  /**< The tensor info of input tensors */
   GstTensorsInfo outputTensorMeta;  /**< The tensor info of output tensors */
@@ -72,6 +80,7 @@ private:
 
   std::shared_ptr < torch::jit::script::Module > model;
 
+  void setAccelerator (const char *accelerators);
   tensor_type getTensorTypeFromTorch (torch::Dtype torchType);
   bool getTensorTypeToTorch (tensor_type tensorType, torch::Dtype * torchType);
   int validateOutputTensor (at::Tensor output);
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -111,16 +120,43 @@ TorchCore::~TorchCore ()
...
+TorchCore::setAccelerator (const char *accelerators)
+{
+  use_gpu = TRUE;
+  accelerator = parse_accl_hw (accelerators, torch_accl_support);
+  if (accelerator == ACCL_NONE)
+    goto use_gpu_ini;
+  if ((accelerator & (ACCL_CPU | ACCL_DEFAULT | ACCL_AUTO)) != 0)
+    use_gpu = FALSE;
+
+  g_message (" 0 gpu = %d", use_gpu);
```
dongju-chae: What's the meaning of a number followed by `gpu`?

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -111,16 +120,43 @@ TorchCore::~TorchCore ()
...
  *        -1 if the model is not loaded.
  */
 int
-TorchCore::init (const GstTensorFilterProperties * prop,
-    const gboolean torch_use_gpu)
+TorchCore::init (const GstTensorFilterProperties * prop)
 {
-  use_gpu = torch_use_gpu;
+  setAccelerator (prop->accl_str);
+  g_message ("gpu = %d, accl = %s", use_gpu, get_accl_hw_str(accelerator));
```
dongju-chae: How about printing this message only when `DBG` is enabled?

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -111,16 +120,43 @@ TorchCore::~TorchCore ()
...
+ * @brief	Set the accelerator for the pytorch
+ */
+void
+TorchCore::setAccelerator (const char *accelerators)
+{
+  use_gpu = TRUE;
+  accelerator = parse_accl_hw (accelerators, torch_accl_support);
+  if (accelerator == ACCL_NONE)
+    goto use_gpu_ini;
+  if ((accelerator & (ACCL_CPU | ACCL_DEFAULT | ACCL_AUTO)) != 0)
```
myungjoo: ACCL_AUTO means no GPU?

kparichay: This is becasuse the pytorch package provided with nnstreamer launchpad does not support GPU. Should I update ACCL_AUTO to imply GPU?

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
kparichay: This was unnecessary logging, removed.

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
kparichay: @dongju-chae need to set DBG variable for ubuntu build.
Plus this needs to be set in other extensions as well.
Will do this separately.

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
myungjoo: It should use GPU if it can use GPU it won't use GPU if it cannot (or is not recommended to) use GPU. You need not to disable it with AUTO.

kparichay: Updated ACCL_AUTO to use GPU.

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -111,16 +120,41 @@ TorchCore::~TorchCore ()
...
+ */
+void
+TorchCore::setAccelerator (const char *accelerators)
+{
+  use_gpu = TRUE;
+  accelerator = parse_accl_hw (accelerators, torch_accl_support);
+  if (accelerator == ACCL_NONE)
+    goto use_gpu_ini;
+  if ((accelerator & (ACCL_CPU | ACCL_DEFAULT)) != 0)
+    use_gpu = FALSE;
```
myungjoo: What does this (Line 134) mean? ```use_gpu``` is in the stack.

kparichay: `use_gpu` is class variable declared on line 73.
accelerator stores the enum value corresponding to the string passed in from the user.
(Line 134) sets `use_gpu` to `FALSE` is string passed in is inferred as `ACCL_DEFAULT` or `ACCL_CPU`.

myungjoo: AH.. ok.

PR #2020, opened 2020-01-10
[CodeClean] remove unnecessary code

jaeyun-jung:
1. remove unnecessary function to get caps from tensor-info.
2. move to proper build option and cleanup unnecessary option.

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -260,29 +260,26 @@ Requires:	libedgetpu1
 Requires:	nnstreamer = %{version}-%{release}
 %description edgetpu
 You may enable this package to use Google Edge TPU with NNStreamer and Tizen ML APIs.
-%define enable_edgetpu -Denable-edgetpu=true
-%else
-%define enable_edgetpu -Denable-edgetpu=false
 %endif
 
 ## Define build options ##
-%define	enable_nnfw_runtime	-Denable-nnfw-runtime=false
+%define enable_tizen -Denable-tizen=false -Denable-tizen-sensor=false
+%define enable_api -Denable-capi=false
+%define enable_mvncsdk2 -Denable-movidius-ncsdk2=false
+%define enable_nnfw_runtime -Denable-nnfw-runtime=false
+%define element_restriction -Denable-element-restriction=false
+
 %if %{with tizen}
 %define enable_tizen -Denable-tizen=true -Denable-tizen-sensor=true
 %define enable_api -Denable-capi=true
 %define enable_mvncsdk2 -Denable-movidius-ncsdk2=true
 %ifarch %arm aarch64
-%define		enable_nnfw_runtime	-Denable-nnfw-runtime=true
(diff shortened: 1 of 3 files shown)


packaging/nnstreamer.spec
```diff
@@ -328,14 +332,15 @@ CFLAGS="${CFLAGS} -fprofile-arcs -ftest-coverage"
...
-	%{enable_python} build
+	--bindir=%{nnstexampledir} --includedir=%{_includedir} -Dinstall-example=true \
+	%{enable_api} %{enable_tizen} %{element_restriction} -Denable-env-var=false -Denable-symbolic-link=false \
+	-Denable-pytorch=false -Denable-caffe2=false %{enable_tf} %{enable_python} \
+	%{enable_nnfw_runtime} %{enable_mvncsdk2} %{enable_armnn} %{enable_edgetpu} \
+	build
 
 ninja -C build %{?_smp_mflags}
 
+%if 0%{?unit_test}
```
dongju-chae: I remember that some unittests should be tested unconditionally in Tizen.
Refer to discussions in https://github.com/nnsuite/nnstreamer/pull/1976.

jaeyun-jung: Question & Discussion: nnfw and tizen-sensor should always run even if unit-test is not enabled?

packaging/nnstreamer.spec
```diff
@@ -345,11 +350,12 @@ export NNSTREAMER_DECODERS=$(pwd)/build/ext/nnstreamer/tensor_decoder
 %define test_script $(pwd)/packaging/run_unittests_binaries.sh
 
 %if %{with tizen}
+    %ifarch %arm aarch64
```
kparichay: This is not needed. The test_script handles if the file exists or not.

packaging/nnstreamer.spec
myungjoo: Requirements & Background:
1. Unit test is disabled for arm/arm64 in CI server because it takes too much time.
2. NNFW is enabled for arm/arm64 only.
3. Unit test for nnfw only won't take too much time.
4. Tizen-sensor has a lot of dependencies on Tizen; thus, I'm worried about regressions from changes in Tizen.org. ==> Do unit-test Tizen-sensor in tizen.org's infrastructure unconditionally. But not for other unit tests in tizen.org's infrastructure (takes too much time)
5. NNFW has some dependencies on Tizen; thus, I'm also worried about regressions from changes in Tizen.org, as well.

In short, for NNFW & Tizen-sensor, regressions from changes in Tizen.org might haunt us. With unit-test (nnfw/tizen-sensor) enabled unconditionally in Tizen, we can be relieved without the high cost of all-unit-test-ARM.

packaging/nnstreamer.spec
jaeyun-jung: updated

packaging/nnstreamer.spec
jaeyun-jung: I understand, rollback to old.

PR #2030, opened 2020-01-14
[Filter/C++] Complete implementation of C++ filter.

myungjoo:
With C++ filter, application authors may supply C++ classes as tensor-filters.
They may provide C++ classes as an independent .so files for other application
authors as well.

@TODO: more tests.
@TODO: usage manual

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
@@ -25,6 +25,10 @@
  *       If you want to write an OpenCV custom filter for nnstreamer, this is a good choice.
  *
  */
+#include <iostream>
+#include <string>
+
+#include <dlfcn.h>
 #include <errno.h>
 #include <string.h>
 #include <glib.h>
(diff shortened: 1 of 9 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
```diff
@@ -104,37 +166,70 @@ static int cpp_invoke (const GstTensorFilterProperties *prop, void **private_dat
...
+  /** The class is deallocated from unordered_map if ref_count hits 0 */
+  cpp->ref_count--;
+
+  if (cpp->ref_count == 0) {
+    size_t num = filters.erase(cpp->name);
+    if (num != 1) {
+      g_critical("cpp_close failed to remove %s from unordered map\n",
+          cpp->name);
+    }
+    dlclose (cpp->handle);
```
dongju-chae: How about checking the validity of `cpp->handle` as well.

myungjoo:
More changes in the class definition coming along with a full set of test cases. Please hold on.

ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
myungjoo: In the new commit, I'll drop the handle property and close handle right after open.

PR #2033, opened 2020-01-14
[Android] custom-filter and amcsrc for native build @open sesame 1/16 16:11

jaeyun-jung:
register custom-filter and amcsrc for native build.


jaeyun-jung:
PR updated
2020-01-20 add filter-subplugin cpp

api/android/api/src/main/jni/nnstreamer-native-api.c
```diff
@@ -572,11 +576,15 @@ nnstreamer_native_initialize (void)
...
-  /* tensor-filter sub-plugin for tensorflow-lite */
-  init_filter_tflite ();
+  /* Android MediaCodec */
+  GST_PLUGIN_STATIC_REGISTER (amcsrc);
 
+  /* tensor-filter sub-plugins */
+  init_filter_cpp ();
+  init_filter_custom ();
+  init_filter_custom_easy ();
+  init_filter_tflite ();
```
myungjoo: We need some mechanism added (preferably with some internal APIs) to prevent calling init/fini twice in a given process.

jaeyun-jung: add flag and global lock to check the plugin registration.

PR #2037, opened 2020-01-15
[Conf/CodeClean] remove duplicated code

jaeyun-jung:
add internal function for sub-plugin scan.

### DIFF
--- a/gst/nnstreamer/nnstreamer_subplugin.c
+++ b/gst/nnstreamer/nnstreamer_subplugin.c
@@ -26,7 +26,7 @@
 #include <dlfcn.h>
 #include <features.h>           /* Check libc version for dlclose-related workaround */
 #include <glib.h>
-#include <gmodule.h>
+
 #include "nnstreamer_subplugin.h"
 #include "nnstreamer_conf.h"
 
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -91,19 +133,11 @@ get_subplugin (subpluginType type, const char *name)
...
-      dlerror ();
-      G_UNLOCK (splock);
-      handle = dlopen (fullpath, RTLD_NOW);
-      /* If this is a correct subplugin, it will register itself */
-      G_LOCK (splock);
-      if (NULL == handle) {
-        g_critical ("Cannot dlopen %s with error %s.", fullpath, dlerror ());
-        continue;
-      }
+      _search_subplugin (type, info.names[i], info.paths[i]);
```
dongju-chae: Now no need to check whether it's successful or not?

jaeyun-jung: currently the result is ignored here for the full-scan and registration. actually result will be checked in line 155.

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -66,24 +63,65 @@ static subpluginSearchLogic searchAlgorithm[] = {
...
+    g_critical ("Cannot dlopen %s(%s) with error %s.", name, path, dlerror ());
+    return NULL;
+  }
+
+  spdata = _get_subplugin_data (type, name);
+  if (spdata == NULL) {
+    g_critical
+        ("nnstreamer_subplugin of %s(%s) is broken. It does not call register_subplugin with its init function.",
+        name, path);
+    dlclose (handle);
```
kparichay: I think the handle should be closed outside the if statement.

myungjoo: No. the handle should be kept alive if its contents (the subplugin) are to be used elsewhere. That's why I'm working on #2049 although I'm experiencing strange behaviors with it.

PR #2039, opened 2020-01-15
New Filter Subplugin Interface Draft: C++ support

myungjoo:
With this, developers can write a C++ class to create
a tensor_filter subplugin without writing the duplicated
C codes.

Changes in V2:
- Redesigned the whole classes
- Applied the recent changes of #2072
Changes in V3:
- Re-redesigned the whole class.
- Build tested without subplugins

TODOS in later PRs:
- Refactor a few subplugins as pure C++ classes and test/update this class.
- Add packaging info
- Add more inline doxygen comments.

With this PR completed in the future, we can remove all the duplicated C code in C++ subplugins.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_cpp.hh
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_cpp.hh
@@ -28,6 +28,13 @@
  * @note This is experimental. The API and Class definition is NOT stable.
  *       If you want to write an OpenCV custom filter for nnstreamer, this is a good choice.
  *
+ * @details
+ *    This is a subplugin for C++ custom filters.
+ *    If you want to write a wrapper for neural network frameworks
+ *    or a hardware adaptor in C++, this is not for you.
+ *    If you want to attach a single C++ class/object as a filter
+ *    (a.k.a. custom filter) to a pipeline, this is what you want.
+ *
  * To Packagers:
  *
  * This is to be exposed with "nnstreamer-c++-dev"
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -1,0 +1,246 @@
+/**
+ * GStreamer Tensor_filter, C++ Subplugin Support. (this is not a subplugin)
+ * Copyright (C) 2020 MyungJoo Ham <myungjoo.ham@samsung.com>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
+ *
+ */
+/**
+ * @file	tensor_filter_support_cc.cc
+ * @date	22 Jan 2020
+ * @brief	Base class for tensor_filter subplugins of C++ classes.
+ * @see		http://github.com/nnsuite/nnstreamer
+ * @author	MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ * @details
+ *    This is not a subplugin, but a helper for C++ subplugins.
+ *    If you want to write a wrapper for neural network frameworks
+ *    or a hardware adaptor in C++, this is what you want.
+ *    If you want to attach a single C++ class/object as a filter
+ *    (a.k.a. custom filter) to a pipeline, use tensor_filter_cpp
+ */
+
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+
+#if __cplusplus < 201103L
+#warn C++11 is required for safe execution
+#else
+#include <type_traits>
+#endif
+
+#define __NO_ANONYMOUS_NESTED_STRUCT
+#include <nnstreamer_plugin_api_filter.h>
+#undef __NO_ANONYMOUS_NESTED_STRUCT
+
+#include "tensor_filter_support_cc.hh"
+
+namespace nnstreamer {
+
+/******************************************************
+ ** Class methods of tensor_filter_subplugin (base)  **
+ ******************************************************/
+
+#define _SANITY_CHECK (0xFACE217714DEADE7ULL)
+
+/**
+ * @brief C tensor-filter wrapper callback function, "open"
+ */
+int tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties * prop,
+    void **private_data)
+{
+  const GstTensorFilterFramework * tfsp = nnstreamer_filter_find (prop->fwname);
+
+  assert (tfsp);
+  assert (tfsp->version == GST_TENSOR_FILTER_FRAMEWORK_V1);
+
+  /* 1. Fetch stored empty object from subplugin api (subplugin_data) */
+  tensor_filter_subplugin *sp = (tensor_filter_subplugin *) tfsp->v1.subplugin_data;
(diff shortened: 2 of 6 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_support_cc.h
```diff
@@ -0,0 +1,83 @@
...
+
+    const char *name; /**< The name of this subplugin. E.g., "edgetpu" */
+    bool allow_in_place;
+    bool allocate_in_invoke;
+    bool run_without_model;
+    bool verify_model_path;
+
+    virtual int invoke_NN (const GstTensorMemory * input, GstTensorMemory * output) = 0;
+    virtual int getInputDimension (GstTensorsInfo * info) = 0;
+    virtual int getOutputDimension (GstTensorsInfo * info) = 0;
```
dongju-chae: How about adding some comments for `getInput/outputDimension` as well (like in `tensor_filter_cpp.h`) ?
Because they can be NULL if `setInputDimension` is defined.

ext/nnstreamer/tensor_filter/tensor_filter_support_cc.h
```diff
@@ -0,0 +1,83 @@
...
+ * To Packagers:
+ *
+ * This is to be exposed with "nnstreamer-c++-dev"
+ */
+#ifndef __NNS_TENSOR_FILTER_CPP_SUBPLUGIN_SUPPORT_H__
+#define __NNS_TENSOR_FILTER_CPP_SUBPLUGIN_SUPPORT_H__
+#ifdef __cplusplus
+
+#include <nnstreamer_plugin_api_filter.h>
+
```
myungjoo: Maybe.. I need to move the class definition to nnstreamer_plugin_api_filter.h

ext/nnstreamer/tensor_filter/tensor_filter_support_cc.h
myungjoo: It'd be a bit different from tensor_filter_cpp.h, but yes I need comments.

Anyway, I'll update them from a pure virtual to virtual so that "not deriving the function" means that the function is not supported.

ext/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
```diff
@@ -0,0 +1,53 @@
...
+  bool isOverridenGetInput (getDimPtr)((*this).*(&tensor_filter_subplugin::getInputDimension)) !=
+      (getDimPtr)(&tensor_filter_subplugin::getInputDimension);
+  bool isOverridenGetOutput (getDimPtr)((*this).*(&tensor_filter_subplugin::getOutputDimension)) !=
+      (getDimPtr)(&tensor_filter_subplugin::getOutputDimension);
+  bool isOverridenSetInput (setDimPtr)((*this).*(&tensor_filter_subplugin::setInputDimension)) !=
+      (setDimPtr)(&tensor_filter_subplugin::setInputDimension);
+#pragma GCC diagnostic pop
+  if (isOverridenGetInput && isOverridenGetOutput)
+    return true;
+  if (isOverridenSetOutput)
```
kparichay: `isOverridenSetInput`?

kparichay:
How will this be packaged? in nnstreamer.so or as a separate subplugin or each subplugin deriving this will package it?

myungjoo:
> How will this be packaged? in nnstreamer.so or as a separate subplugin or each subplugin deriving this will package it?

This .cc is not a subplugin, but a base class implementation for C++ subplugins. Thus, there are two approaches possible:

1. ```%{libdir}/libnnstreamer-cpp.so``` and let C++ subplugins link nnstreamer-cpp. Deploy ```nnstreamer-cpp.rpm```.
    - Pros: clean internal dependency
    - Cons: could be slower
2. Include in ```/usr/lib/gstreamer-1.0/libnnstreamer.so```.
    - Pros: probably faster.
    - Cons: messy internal dependency

I'm inclining to Approach 2 assuming that we can write some reasonable internal dependency statements with meson.

ext/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
myungjoo: With your C-class update, this will be rewritten, but.. yes.. that should be updated.

kparichay:
Approach 1 makes an extra middle-man `nnstreamer-cpp.rpm`. Approach 2 is better.

ext/nnstreamer/tensor_filter/tensor_filter_support_cc.h
kparichay: Given [this](https://github.com/nnsuite/nnstreamer/pull/2039#issuecomment-581772593), moving is better.

ext/nnstreamer/tensor_filter/tensor_filter_cpp.hh
```diff
@@ -28,6 +28,13 @@
  * @note This is experimental. The API and Class definition is NOT stable.
  *       If you want to write an OpenCV custom filter for nnstreamer, this is a good choice.
  *
+ * @details
+ *    This is a subplugin for C++ custom filters.
+ *    If you want to write a wrapper for neural network frameworks
+ *    or a hardware adpator in C++, this is not for you.
```
wooksong: s/adpator/adaptor

gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
```diff
@@ -0,0 +1,246 @@
...
+ * @date	22 Jan 2020
+ * @brief	Base class for tensor_filter subplugins of C++ classes.
+ * @see		http://github.com/nnsuite/nnstreamer
+ * @author	MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ * @details
+ *    This is not a subplugin, but a helper for C++ subplugins.
+ *    If you want to write a wrapper for neural network frameworks
+ *    or a hardware adpator in C++, this is what you want.
```
wooksong: ditto

gst/nnstreamer/tensor_filter/tensor_filter_support_cc.hh
```diff
@@ -0,0 +1,126 @@
...
+ * @date	15 Jan 2020
+ * @brief	Base class for tensor_filter subplugins of C++ classes.
+ * @see		http://github.com/nnsuite/nnstreamer
+ * @author	MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug		No known bugs except for NYI items
+ *
+ * @details
+ *    This is not a subplugin, but a helper for C++ subplugins.
+ *    If you want to write a wrapper for neural network frameworks
+ *    or a hardware adpator in C++, this is what you want.
```
wooksong: ditto

gst/nnstreamer/tensor_filter/tensor_filter_support_cc.hh
```diff
@@ -0,0 +1,126 @@
...
+    virtual tensor_filter_subplugin * getEmptyInstance () = 0;
+        /**< Do return derived_class(), the "empty" object */
+
+    virtual void configure_instance (const GstTensorFilterProperties *prop) = 0;
+        /**< Configure a non-functional "empty" object as a
+             functional "filled" object ready to be invoked */
+
+    virtual ~tensor_filter_subplugin (); /**< called by Close */
+
+    virtual int invoke (const GstTensorMemory * input, GstTensorMemory * output) = 0;
```
wooksong: Q: Is it possible to use references instead of pointers for the const parameters. If it is, we do not need to check whether a given const argument is nullptr or valid in the derived functions.

myungjoo: Oh.. that's good. I've been too much C-oriented (for 30 years... it was the dominant programming language for me)

myungjoo: Now, all the C++ virtual functions of the class use references instead of pointers for their parameters.

PR #2041, opened 2020-01-16
[Meson] Disable external filters for pytorch, caffe2, tf-lite by default 

wooksong:
This PR disables by default the tensor filters to support external neural network frameworks such as pytorch, caffe2, and tensorflow-lite in the meson_options.txt file. According to this change, the .spec and rules files are also modified as well.

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -3,8 +3,11 @@
 %define		gstlibdir	%{_libdir}/%{gstpostfix}
 %define		nnstexampledir	/usr/lib/nnstreamer/bin
 %define		tensorflow_support	0
+%define		tensorflow_lite_support	1
 %define		armnn_support 0
 %define		python_support 1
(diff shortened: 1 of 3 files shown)


packaging/nnstreamer.spec
```diff
@@ -5,6 +5,8 @@
 %define		tensorflow_support	0
 %define		armnn_support 0
 %define		python_support 1
+%define		pytorch_support 0
+%define		caffe2_support 0
```
jaeyun-jung: how about adding 'tensorflow_lite_support'?

wooksong: I thought supporting tf-lite in Tizen is default. But adding tensorflow_lite_support is a good idea.

wooksong: @jaeyun-jung 'tensorflow_lite_support' is added!

PR #2042, opened 2020-01-16
[Meson] Make 'enable-edgetpu' depend on 'enable-tensorflow-lite'

wooksong:
In order to handle the case that 'enable-edgetpu' is set to 'true', meson requires a variable, 'tflite dep', which is a declaration of a dependency on TensorFlow-Lite. Therefore, the 'enable-edgetpu' option
alone cannot properly set the feature. This patch fixes this issue.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -284,24 +284,28 @@ if get_option('enable-cppfilter')
 endif
 
 if get_option('enable-edgetpu')
-  edgetpu_dep = dependency('edgetpu', required: false)
-  if not edgetpu_dep.found()
-    # Ubuntu edgetpu package does not have pkgconfig file
-    edgetpu_dep = cc.find_library('edgetpu')
+  if get_option('enable-tensorflow-lite')
+    edgetpu_dep = dependency('edgetpu', required: false)
+    if not edgetpu_dep.found()
+      # Ubuntu edgetpu package does not have pkgconfig file
+      edgetpu_dep = cc.find_library('edgetpu')
+    endif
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -282,7 +282,7 @@ if get_option('enable-cppfilter')
   )
 endif
 
-if get_option('enable-edgetpu')
+if get_option('enable-tensorflow-lite') and get_option('enable-edgetpu')
```
dongju-chae: How about this style (for better readability, IMO)?
```
if get_option('enable-edgetpu')
  if get_option('enable-tensorflow-lite')
     ...
  else
    warning(...)
endif

Comment (wooksong) [reply ext/nnstreamer/tensor_filter/meson.build]: @dongju-chae Applied. Thanks!

File: ext/nnstreamer/tensor_filter/meson.build [hunk 2]
```diff
@@ -284,24 +284,29 @@ if get_option('enable-cppfilter')
...
-  )
+    shared_library('nnstreamer_filter_edgetpu',
+      nnstreamer_filter_edgetpu_sources,
+      dependencies: [glib_dep, gst_dep, nnstreamer_dep, edgetpu_dep, tflite_dep],
+      install: true,
+      install_dir: filter_subplugin_install_dir
+    )
+  else
+    warning('enable-tensorflow-lite should be set as \'true\' to build the tensor filter for Edge TPU.')
+    warning('-Denable-edgetpu=true is going to be ignored.')
```
myungjoo: Escalate this to error.

wooksong: @myungjoo Escalated to 'error'.

PR #2047, opened 2020-01-20
Filter/C++: Fix potential memory leak.

myungjoo:
dlclose() should not be called if the target-library's contents may
be refered later.
    
dlclose() should be called at the exit of a pipeline to avoid
memory leaks of user applications.
    
Thus, we need to call dlclose for all handles at the ensured exit
point.
    
ps. I'm going to apply similar fixes for subplugin library as well.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
@@ -27,6 +27,7 @@
  */
 #include <iostream>
 #include <string>
+#include <assert.h>
 
 #include <dlfcn.h>
 #include <errno.h>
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
```diff
@@ -200,16 +204,19 @@ int tensor_filter_cpp::open (const GstTensorFilterProperties *prop, void **priva
...
+      dlclose (handle);
       g_printerr_once (__FILE__, __LINE__, "C++ custom filter %s is not found in %s.\n",
           prop->model_files[0], prop->model_files[1]);
       return -EINVAL;
     }
+
+    /** We do not know until when this handle might be required: user may
+      * invoke functions from it at anytime while the pipeline is not
+      * closed */
+    handles.push_back (handle);
```
kparichay: do we need locking around this? (as handle is class static variable shared by all instance of this class).

PR #2048, opened 2020-01-20
[Common/Android] add version info

jaeyun-jung:
1. add new function to get the NNStreamer version string.
2. get nnstreamer version string in Android method.
3. print log about version info when initializing the nnstreamer.

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
@@ -126,7 +126,7 @@ public final class NNStreamer {
     }
 
     /**
-     * Gets the version string of GStreamer and NNStreamer.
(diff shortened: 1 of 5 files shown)


myungjoo:
Why do we want to have version info in .ini file?

jaeyun-jung:
@myungjoo  To get the debug info in target device later

myungjoo:
> @myungjoo To get the debug info in target device later

If you want version info of nnstreamer-library/plugin, add internal API (e.g., at /gst/nnstreamer/nnstreamer.c)

If this is for debugging/devel only, not for actual behavior, let's create another "category":
```
# Values in debug section does not affect the behavior. It's for debugging/development only.
[debug]
version=...
```

jaeyun-jung:
PR updated
2020-01-21 add function to get the version string of nnstreamer

PR #2049, opened 2020-01-21
Fix: subplugin infra: dlclose & free handles at exit

myungjoo:
We need to hold ALL handles if the pipeline is not closed.
However, after the pipeline is closed, all the handles must be
dlclosed; otherwise, the application may hold unnecessary
resources during the execution.

ps. This is supposed to clean up a Coverity issue as well.

Changes in v2:
    Use g_ptr_array instead of garray.
Changes in v3:
    Added workaround for dlclose-racing bug of GLIBC 2.23 (Ubuntu 16.04)

### DIFF
--- a/gst/nnstreamer/nnstreamer_subplugin.c
+++ b/gst/nnstreamer/nnstreamer_subplugin.c
@@ -24,20 +24,27 @@
  */
 
 #include <dlfcn.h>
+#include <features.h>           /* Check libc version for dlclose-related workaround */
 #include <glib.h>
 #include <gmodule.h>
 #include "nnstreamer_subplugin.h"
 #include "nnstreamer_conf.h"
 
+/** @brief Array of dlopen-ed handles */
+static GPtrArray *handles = NULL;
+
+static void init_subplugin (void) __attribute__ ((constructor));
+static void fini_subplugin (void) __attribute__ ((destructor));
+
 typedef struct
 {
   char *name; /**< The name of subplugin */
   const void *data; /**< subplugin specific data forwarded from the subplugin */
-  void *handle; /**< dlopen'ed handle */
 } subpluginData;
 
 static GHashTable *subplugins[NNS_SUBPLUGIN_END] = { 0 };
 
+/** @brief Protects handles and subplugins */
 G_LOCK_DEFINE_STATIC (splock);
 
 /** @brief Private function for g_hash_table data destructor, GDestroyNotify */
@@ -46,8 +53,6 @@ _spdata_destroy (gpointer _data)
 {
   subpluginData *data = _data;
   g_free (data->name);
-  if (data->handle)
-    dlclose (data->handle);
   g_free (data);
 }
 
@@ -103,6 +108,7 @@ get_subplugin (subpluginType type, const char *name)
         g_critical ("Cannot dlopen %s with error %s.", fullpath, dlerror ());
         continue;
       }
+      g_ptr_array_add (handles, (gpointer) handle);
     }
 
     searchAlgorithm[type] = NNS_SEARCH_NO_OP;
@@ -139,6 +145,7 @@ get_subplugin (subpluginType type, const char *name)
       dlclose (handle);
       goto error;
     }
+    g_ptr_array_add (handles, (gpointer) handle);
   }
 
 error:
@@ -192,7 +199,6 @@ register_subplugin (subpluginType type, const char *name, const void *data)
 
   spdata->name = g_strdup (name);
   spdata->data = data;
-  spdata->handle = NULL;
 
   G_LOCK (splock);
   ret = g_hash_table_insert (subplugins[type], g_strdup (name), spdata);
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -29,6 +29,12 @@
 #include "nnstreamer_subplugin.h"
 #include "nnstreamer_conf.h"
 
+/** @brief Array of dlopen-ed handles */
+static GArray *handles = NULL;
+
+static void init_subplugin (void) __attribute__ ((constructor));
```
jaeyun-jung: how about adding priority value?
```
static void init_subplugin (void) __attribute__ ((constructor (101)));
static void fini_subplugin (void) __attribute__ ((destructor (101)));
```

myungjoo: Do you have suggestions on the rules of priority values?

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -216,3 +227,35 @@ unregister_subplugin (subpluginType type, const char *name)
...
+  dlclose (*ptr);
+}
+
+/** @brief Create handles at the start of library */
+static void
+init_subplugin (void)
+{
+  G_LOCK (splock);
+  g_assert (NULL == handles);
+  handles = g_array_new (FALSE, FALSE, sizeof (void *));
```
jaeyun-jung: GPtrArray for array of pointers?

gst/nnstreamer/nnstreamer_subplugin.c
jaeyun-jung: plz ignore this comment

gst/nnstreamer/nnstreamer_subplugin.c
myungjoo: That looks better! Thans!

myungjoo:
I need to test with free functions for g_ptr_array_free further. Do not merge until I verify its behaviors.

myungjoo:
Wow...

https://bugzilla.redhat.com/show_bug.cgi?id=1398370

myungjoo:
Looks like there is a bug in glibc of Ubuntu 16.04: https://bugzilla.redhat.com/show_bug.cgi?id=1264556#c42

I'll need to add glibc version checks in it...

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -216,3 +222,47 @@ unregister_subplugin (subpluginType type, const char *name)
...
+ * Ubuntu 16.04 / GLIBC 2.23 Workaround
+ * If we do dlclose at exit() function, it may incur
+ * https://bugzilla.redhat.com/show_bug.cgi?id=1264556#c42
+ * , which is a GLIBC bug at 2.23.
+ * The corresponding error message is:
+ * Inconsistency detected by ld.so: dl-close.c: 811:
+ * _dl_close: Assertion `map->l_init_called' failed!
+ */
+#if defined(__GLIBC__) && (__GLIBC__ == 2) && (__GLIBC_MINOR__ == 23)
+  return;                       /* Do not call dlclose and return */
```
dongju-chae: Then, in this case, `dlclose()` is not called to release resources?

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -103,6 +108,7 @@ get_subplugin (subpluginType type, const char *name)
         g_critical ("Cannot dlopen %s with error %s.", fullpath, dlerror ());
         continue;
       }
+      g_ptr_array_add (handles, (gpointer) handle);
```
jaeyun-jung: Discussion (later work) : dlopen vs g_module_open
Currently nnstreamer does not have dependency of gmodule.
Is it necessary to use gmodule and remove dl-open/close?

gst/nnstreamer/nnstreamer_subplugin.c
myungjoo: Calling dlclose() breaks the whole application in GLIBC 2.23 if it is called in exit(). (refer to the link in the comment). And we cannot call dlclose() before exit().

gst/nnstreamer/nnstreamer_subplugin.c
myungjoo: If it gives us some advantage, yes. (e.g., using gmodule instead of dlfcn makes the whole nnstreamer compatible with other OS?)

jaeyun-jung: I found gmodule internal code with windows, android, darwin,... So I think it is better to use it.

gst/nnstreamer/nnstreamer_subplugin.c
dongju-chae: Yes, my question was... to avoid this bug, are we going to give up to call `dlcose()` until glibc() is updated? If there is no way to prevent the memory leak in this case, it's fine.

myungjoo: There is a way to prevent it: upgrade GLIBC to 2.24+ or downgrade. :)

So.. actually, we cannot avoid it as long as we cannot predict when the application (nnstreamer user) will stop using a specific subplugin.

gst/nnstreamer/nnstreamer_subplugin.c
myungjoo: If dlfcn.h is not portable, then, yes, that may be added to the future-feature-waiting-list.

PR #2059, opened 2020-01-23
[Fix] Skip duplicated paths from multiple conf sources

### LINKED ISSUE #2057: gst-inspect-1.0 reports duplicated subplugins
`gst-inspect` shows a list of subplugins described in .ini file.
However, it may show duplicated subplugins when there are multiple sources of .ini files.

Initial states
```
$ dchae@dchae-ubuntu:~$ gst-inspect-1.0 tensor_filter | grep sub-plugins -n2
98-                        flags: readable, writable
99-                        String. Default: null
100:  sub-plugins         : Registrable sub-plugins list
101-                        flags: readable
102-                        String. Default: "custom,custom-easy"
```
After copying tflite subplugin
```
dchae@dchae-ubuntu:~$ sudo cp libnnstreamer_filter_tensorflow-lite.so /usr/lib/nnstreamer/filters/
dchae@dchae-ubuntu:~$ gst-inspect-1.0 tensor_filter | grep sub-plugins -n2
98-                        flags: readable, writable
99-                        String. Default: null
100:  sub-plugins         : Registrable sub-plugins list
101-                        flags: readable
102-                        String. Default: "custom,custom-easy,tensorflow-lite,tensorflow-lite"
```

dongju-chae:
This PR skips duplicated paths from multiple conf sources.

It resolves #2057 

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_conf.c
+++ b/gst/nnstreamer/nnstreamer_conf.c
@@ -254,12 +254,19 @@ _fill_in_vstr (gchar *** fullpath_vstr, gchar *** basename_vstr,
 {
   GSList *lstF = NULL, *lstB = NULL;
   vstr_helper vstrF, vstrB;
-  guint i, counter;
+  guint i, j, counter;
 
   counter = 0;
   for (i = 0; i < CONF_SOURCE_END; i++) {
     if (searchpath[i]) {
-      _get_filenames (type, searchpath[i], &lstF, &lstB, &counter);
+      /* skip duplicated paths */
+      for (j = i + 1; j < CONF_SOURCE_END; j++) {
+        if (searchpath[j] && !g_strcmp0 (searchpath[i], searchpath[j])) {
+          break;
+        }
+      }
+      if (j == CONF_SOURCE_END)
+        _get_filenames (type, searchpath[i], &lstF, &lstB, &counter);
     }
   }
 


gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -254,12 +254,19 @@ _fill_in_vstr (gchar *** fullpath_vstr, gchar *** basename_vstr,
...
-  guint i, counter;
+  guint i, j, counter;
 
   counter = 0;
   for (i = 0; i < CONF_SOURCE_END; i++) {
     if (searchpath[i]) {
-      _get_filenames (type, searchpath[i], &lstF, &lstB, &counter);
+      /* skip duplicated paths */
+      for (j = i + 1; j < CONF_SOURCE_END; j++) {
+        if (searchpath[j] && !g_strcmp0 (searchpath[i], searchpath[j])) {
```
kparichay: This logic will avoid duplicated paths but not equivalent paths.
I am not sure if this can happen but unnecessary `. or ..` in the paths can make equivalent paths which aren't equal in string format. How about using `g_file_new_for_path()` to create a GFile and then check equivalence of path with `f_file_equal()`?

dongju-chae: Yes, it could be better, but it may require extra dependency (e.g., `gio`) and I think it's a quite overkill to consider such rare cases.

kparichay: ok.

Issue #2051, opened 2020-01-21
Requirements: data layout


myungjoo:
Related PR: #2007

1. Backward Compatibility: the previous pipelines and filters should work without any changes.
2. Data Layout Agnostic Streams: when not specified either by users or elements, all streams should be kept layout-agnostic (no layout, "none")
3. Auto Layout Negotiation: For example, with a pipeline, ```... ! tensor_converter ! tensor_filter something_mandates_NCHW ! ... ```, the tensor_converter should be able to convert the given inputs to NCHW tensor stream. Anyway, if there is no mandates, it may provide a default data layout that matches the given video/x-raw stream and tensor_filter should be able to accept such unless explicitly specified otherwise.

Before accepting any related changes, we need corresponding additional test cases as well:
1. ```videosrc ! POS ! tensor_converter ! tensor_filter conventional ! POS2 !  ...``` Check if POS has been given the intended data layout automatically.
2. ```videosrc ! POS ! tensor_converter ! tensor_filter requiring NCHW implicitly (automatically) ! POS2 ! ...``` Check if POS has been given the intended data layout (NCHW) automatically and tensor_converter converts the data format automatically.
3. ```videosrc ! POS ! tensor_converter ! tensor_filter requiring NONE implicitly (automatically) ! POS2 ! ...``` Check if POS has been given the intended data layout (NONE) automatically.
4. POS2 above should have NONE layout if no others mandate after that
5. POS2 above may bedome some layout with capsfilter.
6. ```POS2 ! tee name=t t. ! queue ! PIPE1 t. ! queue ! PIPE2```, where PIPE1 uses NCHW and PIPE2 uses NONE or ANY. How are you going to cope with this? (need decision)
7. How will merge/mux/split/demux/aggregator behave with data layout?

Note
- Having a property at tensor_filter so that a filter can transpose internally w/ hardware acceleration is fine as long as it is not invasive (do not make changes in CAPI or common data structure, #2007 is invasive.) E.g., having ```tensor_filter dataop='NHWC2NCHW' ...``` is fine for those who have hardware accelerators for such ops. We may rely on this while we solve this #2051 with the requirements above.

kparichay:
**Base Proposal**
Representing data layout as `gst_type_list` of `g_type_enum` at the template stage. After the capabilities are fixed, this would be fixed as `gst_type_list`. 
PRO: Satisfies above requirements. Existing test cases pass. 
[...]

PR #2065, opened 2020-01-29
[tensor_filter] Support data layout for tensor @open sesame 02/04 16:27

kparichay:
Data layout added for tensor_filter in GstTensorFilterProperties 
Four data layouts are supported
- ANY - unknown does not matter
- NHWC - referred as channels last
- NCHW - referred as channels first
- none - doesnt fit with above options 
These data layouts can be different for each other/tensor in other/tensors 
Layout is taken input as a string, and then parsed for each tensor 
This layout is stored in tensor_filter element itself to be used by its subplugins, and not passed to other elements in the pipeline
Added corresponding unittests as well

This is not the final solution but a temporary solution to provide data layouts to tensor filters. Better solution is being tracked at #2051.

Related Issue: #1986 

**Self evaluation:**

Signed-off-by: Parichay Kapoor 

cc @jaeyun-jung

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -92,6 +92,35 @@ typedef enum
 } accl_hw;
 
 /**
+ * @brief Internal tensor layout format for other/tensor
+ *
+ * The layout is needed by some of the subplugins to appropriately  process the
+ * data based on the axis of the channel in the data. Layout information will be
+ * currently utilized by only some of the sublpugins (SNAP, NNFW), and should be
+ * provided to use some of the subplugins (SNAP).
+ *
+ * Tensor layout is stored locally in the tensor filter element, and not shared
+ * with other elements in the pipeline. Thus, tensor layout is not part of the
+ * capabilities of the element, and does not take part in the caps negotiation.
+ *
+ * NONE layout implies that the layout of the data is neither NHWC nor NCHW. '
+ * However, ANY layout implies that the layout of the provided data is not
+ * relevant.
+ *
+ * @note Providing tensor layout can also decide acceleration to be supported
+ * as not all the accelerators might support all the layouts (NYI).
+ */
+typedef enum _nns_tensor_layout
+{
+  _NNS_LAYOUT_ANY = 0,     /**< does not care about the data layout */
+  _NNS_LAYOUT_NHWC,        /**< NHWC: channel last layout */
+  _NNS_LAYOUT_NCHW,        /**< NCHW: channel first layout */
+  _NNS_LAYOUT_NONE,        /**< NONE: none of the above defined layouts */
+} tensor_layout;
+
+typedef tensor_layout tensors_layout[NNS_TENSOR_SIZE_LIMIT];
+
+/**
  * @brief GstTensorFilter's properties for NN framework (internal data structure)
  *
  * Because custom filters of tensor_filter may need to access internal data
@@ -106,9 +135,11 @@ typedef struct _GstTensorFilterProperties
 
   int input_configured; /**< TRUE if input tensor is configured. Use int instead of gboolean because this is refered by custom plugins. */
   GstTensorsInfo input_meta; /**< configured input tensor info */
+  tensors_layout input_layout; /**< data layout info provided as a property to tensor_filter for the input, defaults to _NNS_LAYOUT_ANY for all the tensors */
 
   int output_configured; /**< TRUE if output tensor is configured. Use int instead of gboolean because this is refered by custom plugins. */
   GstTensorsInfo output_meta; /**< configured output tensor info */
+  tensors_layout output_layout; /**< data layout info provided as a property to tensor_filter for the output, defaults to _NNS_LAYOUT_ANY for all the tensors */
 
   const char *custom_properties; /**< sub-plugin specific custom property values in string */
   const char *accl_str; /**< accelerator configuration passed in as parameter */
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -810,39 +810,49 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
       break;
     }
     case PROP_INPUTLAYOUT:
-      g_assert (!prop->input_configured && value);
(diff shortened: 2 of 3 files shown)


jaeyun-jung:
Discussion (single-shot APIs)
For the Pipeline API, this is ok (tensor-filter property). But with current single-shot functions, there is no way to set the layout in Single C-API.

kparichay:
> For the Pipeline API, this is ok (tensor-filter property). But with current single-shot functions, there is no way to set the layout in Single C-API.

Yes, this interface will have to be added. But the interface should be generic as being discussed in #2051. This PR just provides a temporary solution for internal implementation which is independent of the pipeline caps negotiation.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -80,15 +80,163 @@ enum
...
+  }
+
+  return num_layouts;
+}
+
+/**
+ * @brief Get layout string of tensor layout.
+ * @param layout layout of the tensor
+ * @return string of layout in tensor
+ * @note The returned value should be freed with g_free()
```
jaeyun-jung: plz remove this line (no ned to free string)

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -80,15 +80,163 @@ enum
...
+    case _NNS_LAYOUT_NCHW:
+      return "NCHW";
+    case _NNS_LAYOUT_NHWC:
+      return "NHWC";
+    case _NNS_LAYOUT_NONE:
+      return "NONE";
+    case _NNS_LAYOUT_ANY:
+      return "ANY";
+    default:
+      return NULL;
```
jaeyun-jung: how about returning default 'any'?

kparichay: As function input (tensor_layout) is an enum, IMO passing an invalid value of the enum should result in failure.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -80,15 +80,163 @@ enum
...
+ * @param layout layout of the tensor
+ * @return string of layout in tensor
+ * @note The returned value should be freed with g_free()
+ */
+static const gchar *
+gst_tensor_get_layout_string (tensor_layout layout)
+{
+  switch (layout) {
+    case _NNS_LAYOUT_NCHW:
+      return "NCHW";
```
dongju-chae: I think this needs `g_strdup()`

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: updated.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -106,9 +119,11 @@ typedef struct _GstTensorFilterProperties
 
   int input_configured; /**< TRUE if input tensor is configured. Use int instead of gboolean because this is refered by custom plugins. */
   GstTensorsInfo input_meta; /**< configured input tensor info */
+  tensors_layout input_layout; /**< data layout info for the input */
```
myungjoo: We need to be more specific.
Is this the data layout that the src element (upstream) has given? or the data layout that this filter requires for this input? I guess it is the former, but it might be not that trivial for users. We also need to state that when we need this info and how we will react if this is not given.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -91,6 +91,19 @@ typedef enum
   ACCL_NPU_SR       = 0x4100,     /**< specify device as any SR npu, if possible */
 } accl_hw;
 
+/**
+ * @brief Internal tensor layout format for other/tensor
+ */
```
myungjoo: We need more background info on this. Why do we need this / When do we need this / ...

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: In the current implementation, it's the layout the filter requires for this input and is provided by the user. The data layout that the src element (upstream) has given will be used once tensor_layout becomes a part of the element capabilities.
More info about the layout has been added.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Added.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -360,6 +507,11 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
       g_param_spec_string ("inputtype", "Input tensor element type",
           "Type of each element of the input tensor ?", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INPUTLAYOUT,
+      g_param_spec_string ("inputlayout", "Input Data Layout",
+          "Set channel first (NCHW) or channel last layout (NHWC) or None for input data. "
+          "Layout of the data can be any or NHWC or NCHW or none for now. ",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
```
myungjoo: Use ```g_param_spec_enum``` instead.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -650,6 +809,40 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
         priv->is_updatable = g_value_get_boolean (value);
       break;
     }
+    case PROP_INPUTLAYOUT:
+      g_assert (!prop->input_configured && value);
+      /* Once configured, it cannot be changed in runtime */
```
myungjoo: No. it may be changed (from NULL to valid value). With dynamic pipelines, you don't know when the framework (Gst) will set the value. It may repeatedly call with the same  property value or value->NULL->value.

kparichay: Updated. Now the data layout can be updated as long as the framework is not yet opened.
Changing the data layout after the framework is already opened will be supported later.

PR #2066, opened 2020-01-29
[single/caffe] Enable caffe in single-shot API

kparichay:
Enable support for caffe in single-shot API with armnn as backend framework
Added test cases with caffe and tflite models
Verified by testing on tizen device.

Related issue : #1817 

**Self evaluation:**

### DIFF
--- a/api/capi/include/nnstreamer.h
+++ b/api/capi/include/nnstreamer.h
@@ -128,6 +128,7 @@ typedef enum {
   ML_NNFW_TYPE_OPENVINO = 6,          /**< Intel openVINO. */
   ML_NNFW_TYPE_VIVANTE = 7,           /**< VeriSilicon's Vivante (TBD) */
   ML_NNFW_TYPE_EDGE_TPU = 8,          /**< Google Coral edge TPU (USB) */
+  ML_NNFW_TYPE_ARMNN = 9,             /**< Arm Neural Network framework (support for caffe and tensorflow-lite) */
   ML_NNFW_TYPE_SNAP = 0x2001,         /**< SNAP (Samsung Neural Acceleration Platform), only for Android. */
 } ml_nnfw_type_e;
 
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -518,6 +518,18 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
       status = ML_ERROR_INVALID_PARAMETER;
       goto error;
     }
+  } else if (nnfw == ML_NNFW_TYPE_ARMNN) {
+    /* set input and output tensors information, if available */
+    if (in_tensors_info) {
+      status = ml_single_set_inout_tensors_info (filter_obj, TRUE, in_tensors_info);
+      if (status != ML_ERROR_NONE)
+        goto error;
+    }
+    if (out_tensors_info) {
+      status = ml_single_set_inout_tensors_info (filter_obj, FALSE, out_tensors_info);
+      if (status != ML_ERROR_NONE)
+        goto error;
+    }
   }
 
   switch (nnfw) {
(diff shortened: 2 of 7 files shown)


api/capi/include/nnstreamer.h
```diff
@@ -128,6 +128,7 @@ typedef enum {
   ML_NNFW_TYPE_OPENVINO = 6,          /**< Intel openVINO. */
   ML_NNFW_TYPE_VIVANTE = 7,           /**< VeriSilicon's Vivante (TBD) */
   ML_NNFW_TYPE_EDGE_TPU = 8,          /**< Google Coral edge TPU (USB) */
+  ML_NNFW_TYPE_CAFFE = 9,             /**< Caffe (.caffemodel) */
```
jaeyun-jung: caffe? armnn? which one is correct?
IMO, ML_NNFW_TYPE_ARMNN with armnn-subplugins

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -3022,6 +3022,181 @@ TEST (nnstreamer_capi_singleshot, invoke_05)
...
+        NULL));
+  status = ml_tensors_info_get_tensor_size (in_info, 0, &data_size);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  ASSERT_TRUE (len == data_size/4);
+
+  contents_float = (gfloat *) g_malloc (data_size);
+  for (unsigned int idx=0; idx < len; idx ++) {
+    contents_float[idx] = static_cast<float> (contents_uint8[idx]);
+    contents_float[idx] -= 127.5;
+    contents_float[idx] /= 127.5;
```
dongju-chae: Q: what's the meaning of this calculation? dequantization?

api/capi/include/nnstreamer.h
kparichay: Updated to armnn.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: This calculation takes in a uint8 format data with range [0, 255] and converts to float data type and normalization to range [-1, 1].
Added this in the comment as well.

dongju-chae: Ah, understood.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -3022,6 +3022,183 @@ TEST (nnstreamer_capi_singleshot, invoke_05)
 }
 #endif  /* ENABLE_NNFW_RUNTIME */
 
+#ifdef ENABLE_ARMNN
+/**
+ * @brief Test NNStreamer single shot (caffe/armnn)
+ * @detail Run pipeline with caffe lenet model.
+ */
+TEST (nnstreamer_capi_singleshot, invoke_06)
```
myungjoo: When you add test cases, please add negative cases as well, which test how we behave when the input is incorrect. For each positive case, we are required to have at least one (recommended to have two or more) negative cases.

meson.build
```diff
@@ -244,6 +244,17 @@ if get_option('enable-tflite-nnapi-delegation')
   # For tf-lite/nnapi, enable-nnfw allows to use nnfw::tflite::nnapi as a backend.
 endif
 
+# ArmNN
+if get_option('enable-armnn')
+  armnn_dep = dependency('armnn', required: true)
```
dongju-chae: As it forces `required: true`, the below if-statement is unnecessary.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: Added.

meson.build
kparichay: Updated.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -518,6 +518,18 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
       status = ML_ERROR_INVALID_PARAMETER;
       goto error;
     }
+  } else if (nnfw == ML_NNFW_TYPE_ARMNN) {
```
jaeyun-jung: Question: with armnn, is it possible to get the in/out tensors information from model?

kparichay: Yes, it is possible. However, this is dependent on the framework. 
With tensorflow-lite as backend, all information is available and no information needs to be passed. 
With caffe, input/output layer names, types and num_tensors are needed. Once provided, other information can be taken. Extra information provided from user can also be passed forward.

PR #2070, opened 2020-01-31
[tensor_filter] Unused variable bug resolve

kparichay:
Resolve bug related to unused variable
If DBG is defined, this error is raised as the variable is only used
for check silent property when DBG is not defined

Self evaluation:

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -791,6 +791,12 @@ gst_tensor_filter_fixate_caps (GstBaseTransform * trans,
   silent_debug_caps (caps, "caps");
   silent_debug_caps (othercaps, "othercaps");
 
+  /** Removes no-used-variable warning for priv in when DBG is set */
+  if (priv->fw == NULL) {
+    gst_caps_unref (othercaps);
+    return NULL;
+  }
+
   /**
    * To get the out-caps, GstTensorFilter has to parse tensor info from NN model.
    */


gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -780,12 +781,16 @@ static GstCaps *
 gst_tensor_filter_fixate_caps (GstBaseTransform * trans,
     GstPadDirection direction, GstCaps * caps, GstCaps * othercaps)
 {
+#if defined(DBG_DEFINED)
   GstTensorFilter *self;
   GstTensorFilterPrivate *priv;
+#endif
   GstCaps *result;
 
+#if defined(DBG_DEFINED)
```
dongju-chae: How about adding extra checks using `priv` like other functions?
For example, you may reuse line 791
```
if (priv->fw == NULL)
   return NULL;
```

kparichay: In all the other functions, `priv` is used in other places as well.
Updated as per your suggestion.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -791,6 +791,12 @@ gst_tensor_filter_fixate_caps (GstBaseTransform * trans,
   silent_debug_caps (caps, "caps");
   silent_debug_caps (othercaps, "othercaps");
 
+  /** Removes no-used-variable warning for priv in when DBG is set */
+  if (priv->fw == NULL) {
+    gst_caps_unref (othercaps);
```
myungjoo: Are we safe to do unref othercaps here?
(I now feels that the unref-othercaps at line 805 as well; othercaps is given by the caller and we are unref-ing in the function.)

PR #2071, opened 2020-02-03
[Tests/Filter/OpenVino] Add basic test cases for open/close callbacks

wooksong:
This PR adds basic test cases for the open and close callbacks in the OpenVino tensor filter. To this end, 
- The class definition of TensorFilterOpenVino separated from the existing .cc file into a newly created .hh file.
- The meson build script in the external tensor_filters is updated to provide an internal dependency on the OpenVino tensor filter to the test cases for it.
- The MobileNet V2 model files, pre-compiled via the OpenVino toolkit, are added.

Note that by this PR the line coverage is increased to 40.1%.

See also: #1708 

**Self evaluation:**


tests/nnstreamer_filter_openvino/unittest_openvino.cc
```diff
@@ -0,0 +1,536 @@
...
+
+  g_free (test_model_xml);
+  g_free (test_model_bin);
+  g_free (prop);
+}
+
+/**
+ * @brief Negative test cases for open and close callbacks with wrong model files
+ */
+TEST (tensor_filter_openvino, open_and_close_f0)
```
dongju-chae: How about naming those negative cases `XXX_n` like other unittests?

wooksong: Changed.

wooksong:
#### V2
- Change the names of the negative test cases

wooksong:
#### V3
- Two private member variables, _inputsDataMap and _outputsDataMap, are changed to  protected variables.

wooksong:
#### V4
- Replace inception v3 (>90 MiB) with mobinet v2 ( < 15 MiB)

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -353,4 +353,10 @@ if get_option('enable-openvino')
     install_dir: nnstreamer_libdir
   )
 
+  nnstreamer_filter_openvino_dep = declare_dependency(
+    link_with: nnstreamer_filter_openvino_so_lib,
+    compile_args : openvino_cpp_args,
+    dependencies: [nnstreamer_filter_openvino_deps],
```
myungjoo: You don't need Line 358-360. The code using libnnstreamer_filter_openvino.so does not inherit the build-dependencies of libnnstreamer_filter_openvino.so.

ext/nnstreamer/tensor_filter/tensor_filter_openvino.cc
```diff
@@ -34,74 +34,12 @@
 #include <iostream>
 #include <string>
 #include <vector>
+#include "tensor_filter_openvino.hh"
```
myungjoo: ```s/"tensor_filter_openvino.hh"/<tensor_filter_openvino.hh>/``` ? You have designated the directory of it in meson.

wooksong:
#### V5
- As mentioned in https://github.com/nnsuite/nnstreamer/pull/2071#pullrequestreview-353495056, the dependencies for the unit test executable are revised.
- A note for packagers is added to tensor_filter_openvino.hh

Issue #2045, opened 2020-01-20
Subplugin API Version Management


myungjoo:
Assignees: major subplugin authors

Problem: As in #2013, we may add, remove, update function pointers from subplugin data structures. How can we ensure compatibility and consistency with such updates?

Suggested Requirements:
- If a subplugin binary is not compatible with the given nnstreamer, detect it in runtime and reject it. (buildtime: compiler error).
- Find a mechanism to build subplugins for multiple nnstreamer 'filter subplugin API' versions.
    - Best: a subplugin binary supports multiple API versions
    - Good enough:  from a source, a subplugin binary is created according to the given filter subplugin API version.
    - Good enough: execute old subplugins in a new nnstreamer. (dual /  translation mechanism?)
- FInd a mechanism to gracefully let subplugin authors know there is a subplugin API version mismatch.

myungjoo:
Done with tensor-fitler.

PR #2072, opened 2020-02-03
[subplugins] Refactor subplugin interface

kparichay:
Refactor subplugin interface
Add new interface for the subplugins along with the old interface
Added version number to check if old subplugin interface is used or the new one

Both the interfaces are added as a union,
This will allow for staggered changes rather a big change to be made at once

Related Issue: #2045 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
@@ -559,7 +559,7 @@ caffe2_getOutputDim (const GstTensorFilterProperties * prop,
  * @param[in] data The data element.
  */
 static void
-caffe2_destroyNotify (void *data)
+caffe2_destroyNotify (void **private_data, void *data)
 {
   /* do nothing */
 }
@@ -567,24 +567,25 @@ caffe2_destroyNotify (void *data)
 static gchar filter_subplugin_caffe2[] = "caffe2";
 
 static GstTensorFilterFramework NNS_support_caffe2 = {
-  .name = filter_subplugin_caffe2,
-  .allow_in_place = FALSE,      /** @todo: support this to optimize performance later. */
-  .allocate_in_invoke = TRUE,
-  .run_without_model = FALSE,
-  .verify_model_path = FALSE,
-  .invoke_NN = caffe2_run,
-  .getInputDimension = caffe2_getInputDim,
-  .getOutputDimension = caffe2_getOutputDim,
-  .setInputDimension = NULL,
+  .version = GST_TENSOR_FILTER_FRAMEWORK_V0,
   .open = caffe2_open,
   .close = caffe2_close,
-  .destroyNotify = caffe2_destroyNotify,
 };
 
 /** @brief Initialize this object for tensor_filter subplugin runtime register */
 void
 init_filter_caffe2 (void)
 {
+  NNS_support_caffe2.name = filter_subplugin_caffe2;
+  NNS_support_caffe2.allow_in_place = FALSE;      /** @todo: support this to optimize performance later. */
+  NNS_support_caffe2.allocate_in_invoke = TRUE;
+  NNS_support_caffe2.run_without_model = FALSE;
+  NNS_support_caffe2.verify_model_path = FALSE;
+  NNS_support_caffe2.invoke_NN = caffe2_run;
+  NNS_support_caffe2.getInputDimension = caffe2_getInputDim;
+  NNS_support_caffe2.getOutputDimension = caffe2_getOutputDim;
+  NNS_support_caffe2.destroyNotify = caffe2_destroyNotify;
+
   nnstreamer_filter_probe (&NNS_support_caffe2);
 }
 
--- a/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
@@ -47,19 +47,39 @@ static gchar filter_subplugin_cpp[] = "cpp";
 bool tensor_filter_cpp::dlclose_all_called = false;
 
 static GstTensorFilterFramework NNS_support_cpp = {
-  .name = filter_subplugin_cpp,
-  .allow_in_place = FALSE,      /** @todo: support this to optimize performance later. */
-  .allocate_in_invoke = FALSE,
-  .run_without_model = FALSE,
-  .verify_model_path = FALSE,
-  .invoke_NN = tensor_filter_cpp::invoke,
-  .getInputDimension = tensor_filter_cpp::getInputDim,
-  .getOutputDimension = tensor_filter_cpp::getOutputDim,
-  .setInputDimension = tensor_filter_cpp::setInputDim,
+  .version = GST_TENSOR_FILTER_FRAMEWORK_V0,
   .open = tensor_filter_cpp::open,
   .close = tensor_filter_cpp::close,
 };
 
+G_BEGIN_DECLS
+void init_filter_cpp (void) __attribute__ ((constructor));
+void fini_filter_cpp (void) __attribute__ ((destructor));
+
+/** @brief Initialize this object for tensor_filter subplugin runtime register */
+void
+init_filter_cpp (void)
+{
+  NNS_support_cpp.name = filter_subplugin_cpp;
+  NNS_support_cpp.allow_in_place = FALSE;      /** @todo: support this to optimize performance later. */
+  NNS_support_cpp.allocate_in_invoke = FALSE;
+  NNS_support_cpp.run_without_model = FALSE;
+  NNS_support_cpp.verify_model_path = FALSE;
+  NNS_support_cpp.invoke_NN = tensor_filter_cpp::invoke;
+  NNS_support_cpp.getInputDimension = tensor_filter_cpp::getInputDim;
+  NNS_support_cpp.getOutputDimension = tensor_filter_cpp::getOutputDim;
+  NNS_support_cpp.setInputDimension = tensor_filter_cpp::setInputDim;
+  nnstreamer_filter_probe (&NNS_support_cpp);
+}
+
+/** @brief Destruct the subplugin */
+void
+fini_filter_cpp (void)
+{
+  nnstreamer_filter_exit (NNS_support_cpp.name);
+  tensor_filter_cpp::dlclose_all ();
+}
+G_END_DECLS
 
 #define loadClass(name, ptr) \
   class tensor_filter_cpp *name = (tensor_filter_cpp *) *(ptr); \
@@ -243,10 +263,6 @@ void tensor_filter_cpp::close (const GstTensorFilterProperties *prop, void **pri
   cpp->ref_count--;
 }
 
-G_BEGIN_DECLS
-void init_filter_cpp (void) __attribute__ ((constructor));
-void fini_filter_cpp (void) __attribute__ ((destructor));
-
 /**
  * @brief Call dlclose for all handle
  */
@@ -273,19 +289,3 @@ void tensor_filter_cpp::dlclose_all ()
 #endif
   dlclose_all_called = true;
 }
-
-/** @brief Initialize this object for tensor_filter subplugin runtime register */
-void
-init_filter_cpp (void)
-{
-  nnstreamer_filter_probe (&NNS_support_cpp);
-}
-
-/** @brief Destruct the subplugin */
-void
-fini_filter_cpp (void)
-{
-  nnstreamer_filter_exit (NNS_support_cpp.name);
-  tensor_filter_cpp::dlclose_all ();
-}
-G_END_DECLS
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -40,6 +40,21 @@
 #define ACCL_NPU_SRCN_STR  "npu.srcn" /** srcn hardware supported by nnfw */
 #define ACCL_NPU_SR_STR  "npu.sr"
 
+#define GST_TENSOR_FILTER_FRAMEWORK_BASE (0xDEAFDEAD00000000ULL)
+#define GST_TENSOR_FILTER_FRAMEWORK_V0 (GST_TENSOR_FILTER_FRAMEWORK_BASE)
+#define GST_TENSOR_FILTER_FRAMEWORK_V1 (GST_TENSOR_FILTER_FRAMEWORK_BASE | 0x10000ULL)
+
+/** TODO: update this to 1 after supporting version 1 GstTensorFilterFramework in tensor_filter.c */
+#define GST_TENSOR_FILTER_API_VERSION_DEFINED (0)
+#define GST_TENSOR_FILTER_API_VERSION_MIN (0)	/* The minimum API version supported (could be obsolete) */
+#define GST_TENSOR_FILTER_API_VERSION_MAX (0)	/* The maximum API version supported (recommended) */
+
+/**
+ * @brief Check the value of the version field of GstTensorFilterFramework
+ */
+#define checkGstTensorFitlerFrameworkVersion(value, version) \
+  ((GST_TENSOR_FILTER_FRAMEWORK_BASE | ((version) << 16)) == (value & 0xFFFFFFFFFFFF0000ULL))
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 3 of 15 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
```diff
@@ -241,6 +237,13 @@ void fini_filter_cpp (void) __attribute__ ((destructor));
 void
 init_filter_cpp (void)
 {
+  NNS_support_cpp.allow_in_place = FALSE;      /** @todo: support this to optimize performance later. */
```
kparichay: Should this be changed with `extern C` wrapper as C supports non-trivial designated initializers.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -111,38 +111,106 @@ typedef struct _GstTensorFilterProperties
...
+/**
+ * @brief Tensor_Filter Subplugin definition
+ *
+ * Common callback parameters:
+ * prop Filter properties. Read Only.
+ * private_data Subplugin's private data. Set this (*private_data = XXX) if you want to change filter->private_data.
+ */
+typedef struct _GstTensorFilterFramework
+{
+  uint64_t version; /**< Version of the struct, can be 0x0 (earlier version (_GstTensorFilterFramework_v1)) or 0x2 (newer version (_GstTensorFilterFramework_v2)) */
```
myungjoo: In order to verify the validity in case this struct is given as void * or from dlopen, let's define version as

```
| 32bit (validity check) | 16bit (API version) | 16bit (Subplugin's internal version. Tensor_filter does not case) |
```

Then, let's define:
```
#define GstTensorFilterFrameworkV1 (0xDEAFDEAD00010000)
```
And let the current filters use this macro instead of hardcoded values.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -111,38 +111,106 @@ typedef struct _GstTensorFilterProperties
...
-      void **private_data, GstTensorsInfo * info);
+  /**
+   * @brief Distinct elements between two versions of the subplugin interfaces
+   */
+  union
+  {
+    /**
+     * @brief Tensor_Filter Subplugin definition Version 0
+     */
+    struct /** _GstTensorFilterFramework_v1 */
```
myungjoo: s/v1/v0/

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -184,41 +252,81 @@ typedef struct _GstTensorFilterFramework
...
        * @return 0 if supported. -errno if not supported.
        */
+    };
+
+    /**
+     * @brief Tensor_Filter Subplugin definition Version 2
+     */
+    struct /** _GstTensorFilterFramework_v2 */
+    {
+      int (*getFrameworkInfo) (void **private_data, GstTensorFilterFrameworkInfo *fw_info);
```
myungjoo: I don't think private_data is available for getFrameworkInfo.
prop will be available at that time, though.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -184,41 +252,81 @@ typedef struct _GstTensorFilterFramework
...
+       * @return 0 if OK. non-zero if error.
+       */
+
+      int (*getModelInfo) (const GstTensorFilterProperties * prop,
+          void **private_data, model_info_ops ops, GstTensorsInfo *in_info,
+          GstTensorsInfo *out_info);
+      /**< Mandatory callback. Gets the model related tensor info.
+       * If ops == GET_INPUT_INFO, in_info would contain the input tensor info.
+       * If ops == GET_OUTPUT_INFO, out_info would contain the output tensor info.
+       * If ops == SET_INPUT_INFO, in_info would contain the provided input tensor info, and out_info would contain the updated output tensor info.
```
myungjoo: It appears that you were right about this yesterday. I looks like that we are ok to merge GET-INPUT and GET-OUTPUT as a single op.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -184,41 +252,81 @@ typedef struct _GstTensorFilterFramework
...
+       *
+       * @param[in] prop read-only property values
+       * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer.
+       * @param[in] ops operation to be performed
+       * @param[in/out] in_info structure of input tensor info
+       * @param[out] out_info structure of output tensor info (return value)
+       * @return 0 if OK. non-zero if error. -ENOENT is operation is not supported.
+       */
+
+      int (*sendEvent) (const GstTensorFilterProperties * prop, void **private_data,
```
myungjoo: Wouldn't ```eventHandler``` be more intuitive for developers?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: private_data is optional (it can be NULL).
this is needed for custom filter subplugin as `allocate_in_invoke` property for custom filter is finalized based on the model. 
OR 
remove private_data from here, and move `allocate_in_invoke` property to getModelInfo (note that this property is model dependent only for custom filter subplugin).

myungjoo: Ah.. ok.. then, let's keep both prop and private_data. and write a "caution" statement as a comment that it could be NULL if it's not yet opened.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -111,38 +111,106 @@ typedef struct _GstTensorFilterProperties
...
+
+  int (*open) (const GstTensorFilterProperties * prop, void **private_data);
+  /**< Optional. tensor_filter.c will call this before any of other callbacks and will call once before calling close
+   *
+   * @param[in] prop read-only property values
+   * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. Normally, open() allocates memory for private_data.
+   * @return 0 if ok. < 0 if error.
+   */
+
+  void (*close) (const GstTensorFilterProperties * prop, void **private_data);
```
myungjoo: Another thing you need to verify is: Do we really need double pointers for private_data for all callbacks? Don't we need it for open/close only?

ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
myungjoo: No. it's ok with init_filter_cpp. If you are concerned with the readability, you may move init_filter_cpp to Line 55.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Added.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: done `s/Version 1/Version 0/` and `s/Version 2/Version 1/`

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Merged into a single OP and renamed them,

gst/nnstreamer/nnstreamer_plugin_api_filter.h

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Yes, I updated this for v1.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Added an extra caution statement apart from argument description.

ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc
kparichay: Moved to line 55 for readability.

kparichay:
**V2:**
- Merged operation GET_INPUT_INFO and GET_OUTPUT_INFO into 1. Now atleast one of SET_INPUT_INFO and GET_INOUT_INFO must be supported.
- Added framework versioning helper macros
- Older framework is now version 0 and newer one is version 1
- Renamed sendEvent to eventHandler
- private_data is a double pointer only in open/close in v1
- Invoke in v1 compared to invoke_NN in v0 where private_data is not a double pointer
- Moved init and fini functions in tensor_filter_cpp for readability

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -184,41 +264,93 @@ typedef struct _GstTensorFilterFramework
...
+       * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer.
+       * @param[in] ops operation to be performed
+       * @param[in/out] in_info structure of input tensor info
+       * @param[out] out_info structure of output tensor info (return value)
+       * @return 0 if OK. non-zero if error. -ENOENT is operation is not supported.
        */
 
-  int (*checkAvailability) (accl_hw hw);
-      /**< Optional. Check if the provided hardware accelerator is supported. This check is static or dynamic based on framework support. Positive response of this check does not guarantee successful running of model with this accelerator. The static check can be performed without opening the framework.
+      int (*eventHandler) (const GstTensorFilterProperties * prop, void *private_data,
```
jaeyun-jung: Discussion:
Is it necessary to pass any parameter? (e.g. define new event to change accel : CHANGE_ACCEL / GPU)

kparichay: the parameter to be passed via `void *data` is optional (can be NULL).
The handler event will interpret and use `void *data` as per its need and implementation. 
In the example of change accel, `void *data` will have to interpreted as `accl_hw` or `char *`.
As additional operations should directly be able to add into `enum event_ops`, its `eventHandler` is kept generic.

Updated description of data to allow NULL.

kparichay:
**V4:**
Added to description - eventHandler argument `void *data` is allowed to be NULL

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -111,27 +120,98 @@ typedef struct _GstTensorFilterProperties
   GstTensorsInfo output_meta; /**< configured output tensor info */
 
   const char *custom_properties; /**< sub-plugin specific custom property values in string */
-  const char *accl_str; /**< accelerator configuration passed in as parameter */
+  union {
+    const char **supported_accl; /**< accelerators supported by framework intersected with user provided accelerator preference */
+    const char *accl_str; /**< accelerator configuration passed in as parameter */
```
myungjoo: Why these two values are in a union? ```supported_accl``` is a value configured by a subplugin and ```accl_str``` is a value configured by a user (pipeline), right?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
myungjoo: void * should be enough for the basic need.

However, we may be able to define a standard struct for each event type as a union of multiple struct per event type. The data structure for each event is not determined by sub-plugins; it's determined by tensor_filter and tensor_filter calls eventHandler.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -40,6 +40,15 @@
...
 
+#define GST_TENSOR_FILTER_FRAMEWORK_V0 (0xDEAFDEAD00000000ULL)
+#define GST_TENSOR_FILTER_FRAMEWORK_V1 (0xDEAFDEAD00010000ULL)
+
+/**
+ * @brief Check the value of the version field of GstTensorFilterFramework
+ */
+#define checkGstTensorFitlerFrameworkVersion(value, version) \
+  ((0xDEAFDEAD00000000ULL | ((version) << 16)) == (value & 0xFFFFFFFFFFFF0000ULL))
+
```
myungjoo: Let's add
```
#define GST_TENSOR_FILTER_API_VERSION_DEFINED (1)
#define GST_TENSOR_FILTER_API_VERSION_MIN (0)
#define GST_TENSOR_FILTER_API_VERSION_MAX (1)
```

So that obsolete external (subplugins implemented in other git repos) code may detect it at build time.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Yes, these two values are in union because v0 uses `accl_str` (current implementation), and v1 will use `supported_accl`.
In v1, `tensor_filter.c` will call `getFrameworkInfo` to get the list of all the supported frameworks. `tensor_filter.c` will intersect this with user passed list of accelerators, and the intersection (`supported_accl`) will be passed to subplugin when opening it. Subplugin does not directly need to know the exact value configured by the user in the pipeline (`accl_str`). 

Regarding dynamically updated property values, I think they can be best handled with sendEvent (`updateProperty`).

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Updated. However, currently set `GST_TENSOR_FILTER_API_VERSION_DEFINED` to `0`, and `GST_TENSOR_FILTER_API_VERSION_MAX` to `0`.
Will upgrade these  to 1 once tensor_filter.c supports the new subplugin API.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -111,27 +120,98 @@ typedef struct _GstTensorFilterProperties
...
+  RELOAD_MODEL,   /** Reloads the subplugin with newely provided model */
+} event_ops;
+
+
+/**
+ * @brief Tensor_Filter Subplugin's model related info gathering operations
+ */
+typedef enum
+{
+  GET_INOUT_INFO,   /**< Gets the input and output tensor info */
```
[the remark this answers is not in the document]
kparichay: Agreed, updated.

kparichay:
**V5:**
Added tensor filter API versions for external subplugins to verify compatibility at compile time
Renamed GET_INOUT_INFO to GET_IN_OUT_INFO for readability

gst/nnstreamer/nnstreamer_plugin_api_filter.h
kparichay: Right, it is determined by the caller and not by the callee.
Added standard struct for each event type.

kparichay:
**V6:**
Added more operations to be supported with eventHandler which allow updating the properties
Added GstTensorFilterFrameworkEventData which allows arguments to be passed dependent on the event
GstTensorFilterProperties is passed to all the callbacks

kparichay:
**V7:**
Updated destroyNotify to take private_data as an input for V0, and corresponding changes in tensor_filter and subplugins
Added allocateInInvoke for V0
These changes are taken from #2013

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
```diff
@@ -559,32 +559,35 @@ caffe2_getOutputDim (const GstTensorFilterProperties * prop,
...
-  .getInputDimension = caffe2_getInputDim,
-  .getOutputDimension = caffe2_getOutputDim,
-  .setInputDimension = NULL,
+  .version = GST_TENSOR_FILTER_FRAMEWORK_V0,
   .open = caffe2_open,
   .close = caffe2_close,
-  .destroyNotify = caffe2_destroyNotify,
+  {
+    .name = filter_subplugin_caffe2,
+  }
```
myungjoo: Doesn't this trigger error/warning with SVACE/Coverity?

gst/nnstreamer/nnstreamer_plugin_api_filter.h
myungjoo: If a given single version will use only one part of the union, it's ok.

kparichay: done.

ext/nnstreamer/tensor_filter/tensor_filter_caffe2.cc
kparichay: Moved `name` assignment to inside the `ini_filter_*`.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -40,6 +40,20 @@
...
+/** TODO: update this to 1 after supporting version 1 GstTensorFilterFramework in tensor_filter.c */
+#define GST_TENSOR_FILTER_API_VERSION_DEFINED (0)
+#define GST_TENSOR_FILTER_API_VERSION_MIN (0)	/* The minimum API version supported (could be obsolete) */
+#define GST_TENSOR_FILTER_API_VERSION_MAX (0)	/* The maximum API version supported (recommended) */
+
+/**
+ * @brief Check the value of the version field of GstTensorFilterFramework
+ */
+#define checkGstTensorFitlerFrameworkVersion(value, version) \
+  ((0xDEAFDEAD00000000ULL | ((version) << 16)) == (value & 0xFFFFFFFFFFFF0000ULL))
```
dongju-chae: How about defining `0xDEAFDEAD00000000UL` for readability? (and can be used in both `GST_TENSOR_FILTER_FRAMEWORK_V0/1`)

myungjoo: Appears good.

myungjoo:
This is a huge change that mandates to change version number (#2096). I'd need more reviews/approvals from reviewers.

Those who have written filter-subplugin: @jaeyun-jung @wooksong @helloahn @leemgs

helloahn:
LGTM

PR #2073, opened 2020-02-03
[TestCoverage/Fix] Generate test coverage results in gcc-9 env.

dongju-chae:
This PR generates test coverage results in `gcc-9` env.

But, it requires the latest version of lcov, and some perl modules.
So, enabling `testcoverage` should be done only in CI servers that can hold
cached RPM files (i.e., `lcov`, `perl-PerlIO-gzip`, `perl-JSON`).

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -85,6 +85,7 @@ BuildRequires:	pkgconfig(edgetpu)
 %endif
 
 %if 0%{?testcoverage}
+# to be compatible with gcc-9, lcov should have a higher version than 1.14.1
 BuildRequires: lcov
 # BuildRequires:	taos-ci-unittest-coverage-assessment
 %endif
--- a/tests/unittestcoverage.py
+++ b/tests/unittestcoverage.py
@@ -194,25 +194,24 @@ def check_component(path):
   out = os.popen("gcov -p -r -s " + path + " `find " + buildpath +
                  " -name *.gcno`").read()
   dprint(out)
-  endpoint = len(out) - 1
-  while (out[endpoint] == '\n'):
-    endpoint = endpoint - 1
-  startpoint = endpoint
-  endpoint = endpoint + 1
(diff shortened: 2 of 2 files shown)


packaging/nnstreamer.spec
```diff
@@ -85,7 +85,10 @@ BuildRequires:	pkgconfig(edgetpu)
 %endif
 
 %if 0%{?testcoverage}
+# the latest version of lcov requires extra perl modules
```
myungjoo: Really? ```platform/upstream/lcov``` has no changes for the last 23 months.

I'd rather add Requires: perl-* at lcov.git.

dongju-chae: Yes, because Tizen does not have the latest `lcov` package, we need to use our repositories or cached RPM files for `lcov` and related ones.

FYI, I tested using my repositories:
- lcov: https://github.com/dongju-chae/lcov
- perl-PerlIO-gzip: https://github.sec.samsung.net/dongju-chae/perl-PerlIO-gzip
- perl-JSON: https://github.sec.samsung.net/dongju-chae/perl-JSON

I thought that this flag is enabled in our CI server only, it's okay to put there RPM files to cached one (e.g., `GBS-ROOT-SNAPSHOT/local/repos/tizen/x86_64/RPMS/`)

tests/unittestcoverage.py
```diff
@@ -194,25 +194,21 @@ def check_component(path):
...
+  # Calculate a line coverage per file
+  for each_line in out.splitlines():
+    m = re.match("Lines executed:(\d+.\d+)% of (\d+)$", each_line)
+    if m:
+      rate = float(m.group(1))
+      lines = int(m.group(2))
+
+      total_lines = total_lines + lines
+      total_covered = total_covered + (rate * lines)
+  total_rate = total_covered / total_lines
```
kparichay: you can add a check here that total_lines is not 0.

dongju-chae: Updated

PR #2078, opened 2020-02-06
[Api/Single] add new function about property

jaeyun-jung:
ACR required.
Add new function to set/get the property of tensor-filter in single-shot.

With this function, we can support the tensor layout in single-shot instance.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestSNAP.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestSNAP.java
@@ -78,6 +78,10 @@ public class APITestSNAP {
             /* let's ignore timeout (set 60 sec) */
             single.setTimeout(60000);
 
+            /* set layout */
+            single.setProperty("inputlayout", "NHWC");
+            single.setProperty("outputlayout", "NCHW");
+
             /* single-shot invoke */
             for (int i = 0; i < 10; i++) {
                 /* dummy input */
@@ -105,8 +109,8 @@ public class APITestSNAP {
                 "other/tensor,dimension=(string)3:224:224:1,type=(string)float32,framerate=(fraction)0/1 ! " +
                 "tensor_filter framework=snap " +
                     "model=" + models[0].getAbsolutePath() + "," + models[1].getAbsolutePath() + " " +
-                    "input=3:224:224:1 inputtype=float32 inputname=data " +
-                    "output=1:1:1000:1 outputtype=float32 outputname=prob " +
+                    "input=3:224:224:1 inputtype=float32 inputlayout=NHWC inputname=data " +
+                    "output=1:1:1000:1 outputtype=float32 outputlayout=NCHW outputname=prob " +
                     "custom=" + option + " ! " +
                 "tensor_sink name=sinkx";
 
--- a/api/capi/include/nnstreamer-single.h
+++ b/api/capi/include/nnstreamer-single.h
@@ -186,6 +186,33 @@ int ml_single_set_input_info (ml_single_h single, const ml_tensors_info_h info);
 int ml_single_set_timeout (ml_single_h single, unsigned int timeout);
 
 /**
+ * @brief Sets the property value for the given model.
+ * @details Note that a model/framework may not support to change the property after opening the model.
+ * @since_tizen 6.0
+ * @param[in] single The model handle.
+ * @param[in] name The property name.
+ * @param[in] value The property value.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported.
+ * @retval #ML_ERROR_INVALID_PARAMETER Fail. The parameter is invalid.
+ */
+int ml_single_set_property (ml_single_h single, const char *name, const char *value);
+
+/**
+ * @brief Gets the property value for the given model.
+ * @since_tizen 6.0
+ * @param[in] single The model handle.
+ * @param[in] name The property name.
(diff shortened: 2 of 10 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -954,3 +954,56 @@ ml_single_invoke_dynamic (ml_single_h single,
...
+
+  if (!single || !name || !value)
+    return ML_ERROR_INVALID_PARAMETER;
+
+  ML_SINGLE_GET_VALID_HANDLE_LOCKED (single_h, single, 0);
+
+  filter_obj = G_OBJECT (single_h->filter);
+
+  /** @todo fill here (allow all properties?) */
+  g_object_set (filter_obj, name, value, NULL);
```
kparichay: As not all properties of tensor_filter are string (some are boolean), do we allow setting those properties here by parsing them manually or setting them is not allowed?

jaeyun-jung: IMO, it is better to allow only the properties of tensor info (input,output, inputname, inputtype, ... ) and accelerator.
We cannot change model path and framework after opening the single-shot instance.

kparichay: Right.
However what about properties like `is-updatable`. It expects a boolean value. 
Such properties will also be supported? (will require parsing input `char *` to `boolean`).

jaeyun-jung: Allowed property name:
input, inputtype, inputname, inputlayout
output, outputtype, outputname, outputlayout
accelerator
is-updatable

For the case of boolean type: parse string value (true/false)

api/capi/include/nnstreamer-single.h
```diff
@@ -185,6 +185,33 @@ int ml_single_set_input_info (ml_single_h single, const ml_tensors_info_h info);
...
+ * @param[in] value The property value.
+ * @return @c 0 on success. Otherwise a negative error value.
+ * @retval #ML_ERROR_NONE Successful
+ * @retval #ML_ERROR_NOT_SUPPORTED Not supported.
+ * @retval #ML_ERROR_INVALID_PARAMETER Fail. The parameter is invalid.
+ */
+int ml_single_set_property (ml_single_h single, const char *name, const char *value);
+
+/**
+ * @brief Gets the property value for the given model.
```
kparichay: How about adding a note that the behavior of this function changes before and after an invoke?
Before the first invoke call, the properties can be updated freely as the framework is not yet open.
However, after the first invoke (`ml_single_invoke`), the support of updating the properties will be limited depending on the support provided by the framework.

api/capi/include/nnstreamer-single.h
```diff
@@ -185,6 +185,33 @@ int ml_single_set_input_info (ml_single_h single, const ml_tensors_info_h info);
  */
 int ml_single_set_timeout (ml_single_h single, unsigned int timeout);
 
+/**
+ * @brief Sets the property value for the given model.
```
kparichay: How about adding a note that the behavior of this function changes before and after an invoke?
Before the first invoke call, the properties can be updated freely as the framework is not yet open.
However, after the first invoke (ml_single_invoke), the support of updating the properties will be limited depending on the support provided by the framework.

api/capi/include/nnstreamer-single.h
```diff
@@ -185,6 +185,33 @@ int ml_single_set_input_info (ml_single_h single, const ml_tensors_info_h info);
...
+ * @details Note that a model/framework may not support to change the property after opening the model.
```
kparichay: How about update `opening the model` to `after the first call to invoke`? The user will not know when the model is opened. It might seem that `ml_single_open` might open the model.
Before the first invoke call, the properties can be updated freely as the framework is not yet open.
However, after the first invoke (ml_single_invoke), the support of updating the properties will be limited depending on the support provided by the framework.

jaeyun-jung: ml_single_open() calls the callback start(), then open the model. So I think 'open model' is ok.

kparichay: aah, yes, my bad.

jaeyun-jung:
PR updated
2020-02-07 implement tensors-info update and add android api

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestSNAP.java
```diff
@@ -105,8 +109,8 @@ private void runPipelineCaffe(boolean useGPU) {
                 "other/tensor,dimension=(string)3:224:224:1,type=(string)float32,framerate=(fraction)0/1 ! " +
                 "tensor_filter framework=snap " +
                     "model=" + models[0].getAbsolutePath() + "," + models[1].getAbsolutePath() + " " +
-                    "input=3:224:224:1 inputtype=float32 inputname=data " +
-                    "output=1:1:1000:1 outputtype=float32 outputname=prob " +
+                    "input=3:224:224:1 inputtype=float32 inputlayout=NHWC inputname=data " +
```
dongju-chae: Q: if `inputlayout` is `NHWC`, the layout of input seems like `1:224:224:3`, not `3:224:224:1`.  Or, is there any underlying assumption on this?

jaeyun-jung: in snap subplugin, the dimension is reversed when invoking the model. (3:224:224:1 -> input shape in snap 1:224:224:3)

dongju-chae: Oh, got it.

PR #2087, opened 2020-02-10
[coverity] single-shot API coverity fix

kparichay:
Added coverity fix of checking return value
In ml_single_set_info_in_handle: Value returned from a function is not checked for errors before being used (CWE-252)
    
Fix coverity issue CWE-416
Issue: In ml_single_set_info_in_handle: A pointer to freed memory is dereferenced, used as a function argument, or otherwise used
    
### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -382,9 +382,9 @@ ml_single_set_info_in_handle (ml_single_h single, gboolean is_input,
 
     if (tensors_info && !ml_tensors_info_is_equal (tensors_info, info)) {
       /* given input info is not matched with configured */
+      ml_tensors_info_destroy (info);
       if (is_input) {
         /* try to update tensors info */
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -382,9 +383,9 @@ ml_single_set_info_in_handle (ml_single_h single, gboolean is_input,
 
     if (tensors_info && !ml_tensors_info_is_equal (tensors_info, info)) {
       /* given input info is not matched with configured */
+      ml_tensors_info_destroy (info);
       if (is_input) {
         /* try to update tensors info */
-        ml_tensors_info_destroy (info);
         if (ml_single_update_info (single, tensors_info, &info) != ML_ERROR_NONE)
```
jaeyun-jung: mem-leak : ml_single_update_info() returns newly allocated 'info'. this should be freed

kparichay: done.

PR #2092, opened 2020-02-11
[Coverity] fix coverity CID 1123841, 1123966

anyj0527:
Fix info:
CID 1123841 - check the return value of sensor_listener_start
CID 1123966 - g_free the newly allocated memory

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -590,6 +590,7 @@ _ts_configure_handle (GstTensorSrcTIZENSENSOR * self)
           "The requested sensor sequence %d for sensor %d is not available. The max-sequence is used instead",
           self->sequence, self->type);
       self->sequence = 0;
+      g_free (list);
       return -EINVAL;
     }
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -868,7 +869,10 @@ gst_tensor_src_tizensensor_start (GstBaseSrc * src)
 
   /* 3. Fire it up! */
   self->running = TRUE;
-  sensor_listener_start (self->listener);
+  if (sensor_listener_start (self->listener) != 0) {
+    retval = FALSE;
```
dongju-chae: In this case, I think `self->running = FALSE` is required.

anyj0527: You're right. I'll add it.

dongju-chae: BTW, does `sensor_listener_start` require `self->running = TRUE`?
Otherwise, how about just moving `self->runing = TRUE` to the line 877 for code clean?

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -868,7 +869,11 @@ gst_tensor_src_tizensensor_start (GstBaseSrc * src)
 
   /* 3. Fire it up! */
   self->running = TRUE;
-  sensor_listener_start (self->listener);
+  if (sensor_listener_start (self->listener) != 0) {
+    self->running = FALSE;
+    retval = FALSE;
+    goto exit;
```
kparichay: Although the current implementation of the function can handle already configured sources, you can consider cleaning up the set configuration in case of failure.

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
anyj0527: Yeah just moving the line would be neater.

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
anyj0527: I think `_ts_clean_up_handle (self)` could work for this case... Am I right?

kparichay: yes.

PR #2095, opened 2020-02-11
Support accelerator backends for tensor filter extensions

kparichay:
- Updated the list of HWs supported with openvino. Providing "true:npu" should also work along with "true:npu.movidius". Providing specific movidius accelerator works. However, just providing a generic npu should also default to movidius.
- Added list of supported accelerators as backend for caffe2
- Added list of supported backend accelerators for python extension of tensor filter
- Added supported list of hardware for tensorflow's tensor filter's extension. Added corresponding unit-tests using single API
- Added supported accelerators for custom filters and its unit-tests with single API
- Added unit-tests for nnfw supported accelerators with single API
- Added supported accelerators for ncsdk2 extension and corresponding unit-tests with single API
- Added supported accelerators for armnn and added corresponding unit tests with single API. Tested set accelerators on the device with Caffe model on the device

Resolves: #1849 

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_custom.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_custom.c
@@ -37,6 +37,12 @@
 void init_filter_custom (void) __attribute__ ((constructor));
 void fini_filter_custom (void) __attribute__ ((destructor));
 
+static const gchar *custom_accl_support[] = {
+  ACCL_AUTO_STR,
+  ACCL_DEFAULT_STR,
+  NULL
+};
+
 /**
  * @brief internal_data
  */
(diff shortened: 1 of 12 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_custom.c
```diff
@@ -37,6 +37,12 @@
 void init_filter_custom (void) __attribute__ ((constructor));
 void fini_filter_custom (void) __attribute__ ((destructor));
 
+static const gchar *custom_accl_support[] = {
+  ACCL_AUTO_STR,
+  ACCL_DEFAULT_STR,
+  NULL
+};
```
dongju-chae: Q: Is it now mandatory to define `*_accl_support` and `*_checkAvailability` for all sug-plugins?
I think defining this and `*_checkAvailability` for non-accelerated sub-plugins is somewhat trivial and meaningless. 
What happens if we just set `*_checkAvailability` to NULL?

kparichay: Single-shot C-API provides an interface `ml_check_nnfw_availability` which allows users to check if a particular `nnfw` framework is available with a given `hw` backend. `*_checkAvailability` is required to support this interface.

dongju-chae: Ah, for single-shot APIs. Got it.

kparichay:
Rebased with the master branch.

PR #2102, opened 2020-02-12
[Android] prepare single-only build

jaeyun-jung:
With the request to build single-shot only library, add definition and  update build script.


api/android/api/src/main/jni/nnstreamer-native-api.c
```diff
@@ -24,20 +24,24 @@
...
 extern void init_filter_tflite (void);
-#if defined (ENABLE_SNAP)
-extern void init_filter_snap (void);
-#endif
 extern void init_dv (void);
 extern void init_bb (void);
 extern void init_il (void);
 extern void init_pose (void);
 extern void init_is (void);
+#endif /* NNS_SINGLE_ONLY */
```
dongju-chae: I think `/* NNS_SINGLE_ONLY */` could lead some confusion because it was derived from `#if !defined (NNS_SINGLE_ONLY)`

jaeyun-jung: do you want to change macro to '#ifndef'? most headers use it.

dongju-chae: I meant the comment. When I see `#endif /* NNS_SINGLE_ONLY */`, it looks like a pair of `#if defined (NNS_SINGLE_ONLY)`, not `#if !defined (NNS_SINGLE_ONLY)`.
I am not sure which comment is better, but I think the current comments make some confusion.

jaeyun-jung: ok, I will remove all.

Issue #2105, opened 2020-02-13
[Feature request] Support the `int8` tensor-type in TensorFlow Lite models


ajarthurs:
**Is your feature request related to a problem? Please describe.**
I cannot run TensorFlow Lite models that contain `int8`-type tensors via `tensor_filter` and its TensorFlow Lite framework; however, I have no issue running models containing `uint8`-type tensors. If the model's input tensor-type is `int8`, NNStreamer throws the following negotiation error, erroneously reporting the model's input tensor-type to be `(null)`:
```
tensor_filter tensor_filter.c:557:gst_tensor_filter_configure_tensor:<tensorfilter0> The input tensor is not compatible.
0 : int8 [3:513:513:1] | (null) [3:513:513:1] FAILED
```
It's clear that NNStreamer does not yet support `int8` tensor types in TensorFlow Lite models. For example, [this function](https://github.com/nnsuite/nnstreamer/blob/e67091b777989bc0817ffb14d1e8d2cd51a43d15/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc#L307:L333) does not map `kTfLiteInt8` to `_NNS_INT8`:
```
// ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
tensor_type
TFLiteInterpreter::getTensorType (TfLiteType tfType)
{
  switch (tfType) {
    case kTfLiteFloat32:
      return _NNS_FLOAT32;
    case kTfLiteUInt8:
      return _NNS_UINT8;
    case kTfLiteInt32:
      return _NNS_INT32;
    case kTfLiteBool:
    /** @todo insert `case kTfLiteInt8:` here */
      return _NNS_INT8;
    case kTfLiteInt64:
      return _NNS_INT64;
    case kTfLiteString:
    default:
      /** @todo Support other types */
      break;
  }

  return _NNS_END;
}
```

**Describe the solution you'd like**
Facilitate TensorFlow Lite models containing `int8`-type tensors: I already tried patching the mentioned `getTensorType` function. It resolved the said negotiation error but hangs during playback, so there is more patchwork required.

**Describe alternatives you've considered**
Strictly `uint8`-based quantized models, mostly available from TensorFlow's quantized stock models produced from quantization-aware training.

**Additional context**
[...]

PR #2109, opened 2020-02-17
[Wait for #2119] Support models with signed 8-bit integer I/O

ajarthurs:
TensorFlow Lite models whose I/O is of type `kTfLiteInt8` are rejected,
requiring an I/O wrapper (`Quantize` operators). As signed 8-bit
integers are preferred in TensorFlow's quantization specifications [1],
this patch alleviates the need for an I/O wrapper to run strictly
`kTfLiteInt8`-type models.

Resolves: #2105 

[1] https://www.tensorflow.org/lite/performance/quantization_spec

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -320,6 +320,7 @@ TFLiteInterpreter::getTensorType (TfLiteType tfType)
     case kTfLiteInt32:
       return _NNS_INT32;
     case kTfLiteBool:
+    case kTfLiteInt8:
       return _NNS_INT8;
     case kTfLiteInt64:
       return _NNS_INT64;


myungjoo:
CC: @helloahn , the tf-lite/tf subplugin maintainer.

ajarthurs:
Concerning backward compatibility with pre-1.13 versions of TF Lite, may I suggest adding `enable-tensorflow-lite-legacy` to `meson_options.txt`?

@myungjoo @helloahn

myungjoo:
> Concerning backward compatibility with pre-1.13 versions of TF Lite, may I suggest adding `enable-tensorflow-lite-legacy` to `meson_options.txt`? Or, the alternative could be `enable-tflite-int8` (alongside `enable-tflite-nnapi-delegation`), but that may be too specific.
> 
> @myungjoo @helloahn

enable-tflite-pre-1-13 looks fine. We may need similar features for later versions in the future.

helloahn:
@ajarthursthanks for your suggestion. However, this PR cannot be merged for now because of the compatibility :(

myungjoo:
I'll elaborate a bit here:

There are a few clients using nnstreamer with TF-Lite 1.13, who are following the master branch of this repo; thus, we will need to be prepared for them before merging this. @helloahn will append his commit to this PR and recreate a PR based on it.

After a discussion w/ @helloahn, we've decided to try something like:
```-enable-additional-tflite-types = int8, int16```
Because he will test with a few other options, merging this (along with addtional commits of @helloahn's) may take a few more days.

PR #2115, opened 2020-02-19
Tizen/CAPI: Workaround to support VIVANTE

myungjoo:
We need to rewrite this change to make it more general.

This fixes:
1. if there are multiple files for a model, we need to check the
whole list at once, not indivisually.
2. add VIVANTE detection and verification methods.


myungjoo:
Sorry for mixing up styling fixes. :(

PR #2118, opened 2020-02-20
[CAPI] change the return value of info/data destroy function - @open sesame 03/02 12:16

helloahn:
Currently, when `info/data destroy()` called with the input `NULL`, it returns `ML_ERROR_INVALID_PARAMETER`. IMO, however, it should be done without any kind of error.

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -85,12 +85,12 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
 {
   ml_tensors_info_s *tensors_info;
 
+  if (!info)
+    return ML_ERROR_NONE;
+
   check_feature_state ();
 
   tensors_info = (ml_tensors_info_s *) info;
-
-  if (!tensors_info)
-    return ML_ERROR_INVALID_PARAMETER;
 
   ml_tensors_info_free (tensors_info);
   g_free (tensors_info);
@@ -538,10 +538,10 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
   ml_tensors_data_s *_data;
   guint i;
 
-  check_feature_state ();
-
   if (!data)
-    return ML_ERROR_INVALID_PARAMETER;
+    return ML_ERROR_NONE;
+
+  check_feature_state ();
 
   _data = (ml_tensors_data_s *) data;
 


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -85,13 +85,13 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
...
   ml_tensors_info_s *tensors_info;
 
+  if (!info)
+    return ML_ERROR_NONE;
+
   check_feature_state ();
 
   tensors_info = (ml_tensors_info_s *) info;
 
-  if (!tensors_info)
```
kparichay: can we remove this? I think this handles calling `ml_tensors_info_destroy` on the same info twice.

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -85,13 +85,13 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
...
-    return ML_ERROR_INVALID_PARAMETER;
-
   ml_tensors_info_free (tensors_info);
   g_free (tensors_info);
```
kparichay: how about setting these to null after free? (same for ml_tensors_data_destroy)

helloahn: I added it. please check it is your intention.

kparichay: Yes

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -85,15 +85,16 @@ ml_tensors_info_destroy (ml_tensors_info_h info)
...
+  info = NULL;
```
myungjoo: No effect?

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -553,6 +554,7 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
   }
 
   g_free (_data);
+  data = NULL;
```
myungjoo: Code without effect?

api/capi/src/nnstreamer-capi-util.c
helloahn: initializing the info/data as `NULL` is better than let it go as an unknown state.

myungjoo: Because the argument is ```ml_tensors_info_h info```, ```info=NULL;``` has no effect. It needs to be ```ml_tensors_info_h *info``` and ```*info = NULL;``` to have effects.

helloahn:
> You can't change the value of `info` for the caller if you have the argument as a value, not a reference or pointer of the value.

I agree with your opinion. For the issue I mentioned (it occurs error at g_free() ...) will be fixed with another pr.

helloahn:
it's not matched with tizen's policy. I close it.

PR #2121, opened 2020-02-22
[Api/Single] change return value

jaeyun-jung:
Return not-supported if the property is not available.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -1070,7 +1070,7 @@ ml_single_get_property (ml_single_h single, const char *name, char **value)
     *value = (bool_value) ? g_strdup ("true") : g_strdup ("false");
   } else {
     ml_loge ("The property %s is not available.", name);
-    status = ML_ERROR_INVALID_PARAMETER;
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -1070,7 +1070,7 @@ ml_single_get_property (ml_single_h single, const char *name, char **value)
     *value = (bool_value) ? g_strdup ("true") : g_strdup ("false");
   } else {
     ml_loge ("The property %s is not available.", name);
-    status = ML_ERROR_INVALID_PARAMETER;
+    status = ML_ERROR_NOT_SUPPORTED;
```
kparichay: I think if the argument, property `name`, is a valid property for the `tensor_filter` and is not supported by single-shot API, then `ML_ERROR_NOT_SUPPORTED` as a return value is good.
But if the property asked for is not a valid property for `tensor_filter`, the `ML_ERROR_INVALID_PROPERTY` sounds better.

jaeyun-jung: This is necessary to distinguish the condition - invalid property and invalid value (required C# API)

Issue #2116, opened 2020-02-20
Support TensorFilterFramework V1 based subplugins


kparichay:
This issue tracks supporting GstTensorFilterFramework V1 based subplugins
- [x] Probe and exit #2117 
- [x] Set and get properties #2130
- [x] Basic Callbacks #2131
- [x] Single API support #2132
- [x] EventHandler callbacks for properties update
- [x] Write a subplugin using c++ subplugin interface and test - Waiting for #2194

Issue #2114, opened 2020-02-18
C++ Custom Filter Prop Update


myungjoo:
We need a mechanism to update prop.

We need a policy on updating prop for tensor-filter subplugins in general.
When do we need to update it? (At reload only? or whenever set-property is called?)

kparichay:
For GstTensorFilterFramework V1, `eventHandler` can allow updating some of the properties of the framework. This lets the frameworks decide and support if updating the prop is feasible or not.
This will allow updating the properties whenever set-property is called.
If `eventHandler` fails, then these properties can be cached to be used when reload is called.

myungjoo:
The real problem here is when should we update them.
I didn't do any analyze, but I do feel that calling it with every set-property may be an overkill (and killing the performance)

kparichay:
It is possible to aggregate all the set-property and update all the properties at once with a reload. This will perform faster in some of the cases possibly. However, there are a couple of problems -

- Reload results in closing the existing model, and then loading them again. This will cause unnecessary overhead when the intended update is small (such as just updating the input dimension).
- Single API has function `ml_single_set_property` only. So, setting the property can be delayed till a call to invoke (so that all the property change can be accumulated and done at once), but this will increase the latency for the first call.
- In the pipeline case, the update needs the pipeline to be PAUSE. If the input/output characteristics are changed, then the pipeline needs to be in READY state and do caps negotiations again when going to PAUSE/PLAY. So, all the updates can be done at READY state, and caps negotiation can happen at once the pipeline state goes to PAUSE/PLAY. So, in this case, the aggregation of property change happens at the pipeline level automatically. 

If aggregation of set_property is to be supported at framework level, then how about adding a property to the tensor_filter - `lazy_update`. If this is set, then the update is cached by tensor_filter, and done at once by passing it to the framework right before invoke (this is being lazy), else each call to set_property is passed to the framework (both these cases depends on the support provided by the framework).

myungjoo:
It looks like we need to take care of individual case of property. And I don't feel like to add yet another subplugin property (late_update) for this.

PROP_INPUT/OUTPUT/*TYPE/*NAME:
    - Q: do we need to send event for every change? We do not allow to change it if it's already configured. We may call it when we set "input_configured" or "output_configured". (lazy_update)

PROP_ACCELERATOR:
    - It looks like we need to send the event right away if it's "configured" or "running". ("lazy-update" if it's not configured, yet. "direct-update" if it's configured)

[...]

PR #2130, opened 2020-03-04
[filter] Support get and set properties @open sesame 03/07 05:10

kparichay:
Support get and set properties for GstTensorFilterFramework V1.
Updated GstTensorFilterFramework to include the number of accelerators supported along with its list.
Added TODO where updating of the framework is to be added if the set is called after the framework is opened or input/output is already configured.

Note: All the added TODO with the property update will be handled in a separate PR as per #2114

See also: #2116

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/nnstreamer_plugin_api_filter.h
@@ -52,7 +52,7 @@
 /**
  * @brief Check the value of the version field of GstTensorFilterFramework
  */
-#define checkGstTensorFitlerFrameworkVersion(value, version) \
+#define checkGstTensorFilterFrameworkVersion(value, version) \
   ((GST_TENSOR_FILTER_FRAMEWORK_BASE | ((version) << 16)) == (value & 0xFFFFFFFFFFFF0000ULL))
 
 #ifdef __cplusplus
@@ -158,7 +158,10 @@ typedef struct _GstTensorFilterProperties
 
   const char *custom_properties; /**< sub-plugin specific custom property values in string */
   union {
-    accl_hw *hw_list; /**< accelerators supported by framework intersected with user provided accelerator preference, use in GstTensorFilterFramework V1 only */
+    struct {
+      accl_hw *hw_list; /**< accelerators supported by framework intersected with user provided accelerator preference, use in GstTensorFilterFramework V1 only */
+      int num_hw;       /**< number of hardare accelerators in the hw_list supported by the framework */
+    };
     const char *accl_str; /**< accelerator configuration passed in as parameter, use in GstTensorFilterFramework V0 only */
   };
 
@@ -178,6 +181,7 @@ typedef struct _GstTensorFilterFrameworkInfo
   int run_without_model; /**< TRUE(nonzero) when the neural network framework does not need a model file. Tensor-filter will run invoke_NN without model. */
   int verify_model_path; /**< TRUE(nonzero) when the NNS framework, not the sub-plugin, should verify the path of model files. */
   accl_hw *hw_list; /**< List of supported hardwares by the framework.  Positive response of this check does not guarantee successful running of model with this accelerator. */
+  int num_hw; /**< number of hardware accelerators in the hw_list supported by the framework */
 } GstTensorFilterFrameworkInfo;
 
 /**
@@ -189,6 +193,7 @@ typedef enum
 {
   DESTROY_NOTIFY,   /**< Free the data element allocated in the invoke callback */
   RELOAD_MODEL,     /**< Reloads the subplugin with newely provided model */
+  CUSTOM_PROP,      /**< Update the custom properties for the framework */
   SET_INPUT_PROP,   /**< Update input tensor info and layout */
   SET_OUTPUT_PROP,  /**< Update output tensor info and layout */
   SET_ACCELERATOR,  /**< Update accelerator of the subplugin to be used as backend */
@@ -221,6 +226,11 @@ typedef struct _GstTensorFilterFrameworkEventData
       int num_models;   /**< Updated number of the model files */
     };
 
+    /** for CUSTOM_PROP event */
+    struct {
+      const char *custom_properties; /**< sub-plugin specific custom property values in string */
+    };
+
     /** for SET_INPUT_PROP/SET_OUTPUT_PROP event */
     struct {
       const GstTensorsInfo * info;  /**< The tensor info to be updated to */
@@ -230,6 +240,7 @@ typedef struct _GstTensorFilterFrameworkEventData
     /** for SET_ACCELERATOR event */
     struct {
       accl_hw *hw_list;   /**< accelerators supported by framework intersected with the new user provided accelerator preference */
+      int num_hw;         /**< number of hardare accelerators in the hw_list supported by the framework */
     };
 
   };
@@ -385,7 +396,7 @@ typedef struct _GstTensorFilterFramework
 
       int (*getFrameworkInfo) (const GstTensorFilterProperties * prop,
           void *private_data, GstTensorFilterFrameworkInfo *fw_info);
-      /**< Mandatory callback. Get the framworks statically determined info. Argument 'private_data' can be NULL. If provided 'private_data' is not NULL, then some info, such as 'allocate_in_invoke', can be updated based on the model being used (inferred from the 'private_data' provided). This updated info is useful for custom filter, as some custom filter's ability to support 'allocate_in_invoke' depends on the opened model.
+      /**< Mandatory callback. Get the frameworks statically determined info. Argument 'private_data' can be NULL. If provided 'private_data' is not NULL, then some info, such as 'allocate_in_invoke', can be updated based on the model being used (inferred from the 'private_data' provided). This updated info is useful for custom filter, as some custom filter's ability to support 'allocate_in_invoke' depends on the opened model.
        *
        * @param[in] prop read-only property values
        * @param[in] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer. This parameter can be NULL.
@@ -424,6 +435,7 @@ typedef struct _GstTensorFilterFramework
       /**< Mandatory callback. Runs the event corresponding to the passed operation.
        * If ops == DESTROY_NOTIFY: tensor_filter.c will call it when 'allocate_in_invoke' property of the framework is TRUE. Basically, it is called when the data element is destroyed. If it's set as NULL, g_free() will be used as a default. It will be helpful when the data pointer is included as an object of a nnfw. For instance, if the data pointer is removed when the object is gone, it occurs error. In this case, the objects should be maintained for a while first and destroyed when the data pointer is destroyed. Those kinds of logic could be defined at this method.
        * If ops == RELOAD_MODEL: tensor_filter.c will call it when a model property is newly configured. Also, 'is-updatable' property of the framework should be TRUE. This function reloads a new model passed in as argument via data. Note that it requires extra memory size enough to temporarily hold both old and new models during this function to hide the reload overhead.
+       * If ops == CUSTOM_PROP: tensor_filter will call to update the custom properties of the subplugin.
        * If ops == SET_INPUT_PROP: tensor_filter will call to update the property of the subplugin. This function will take tensor info and layout as the argument. This operation can update input tensor shape, type, name and layout.
        * If ops == SET_OUTPUT_PROP: tensor_filter will call to update the property of the subplugin. This function will take tensor info and layout as the argument. This operation can update output tensor shape, type, name and layout.
        * If ops == SET_ACCELERATOR: tensor_filter will call to update the property of the subplugin. This function will take accelerator list as the argument. This operation will update the backend to be used by the corresponding subplugin.
@@ -434,7 +446,7 @@ typedef struct _GstTensorFilterFramework
        * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer.
        * @param[in] ops operation to be performed
        * @param[in/out] data user sata for the supported handlers (can be NULL)
-       * @return 0 if OK. non-zero if error. -ENOENT if operation is not supported.
+       * @return 0 if OK. non-zero if error. -ENOENT if operation is not supported. -EINVAL if operation is supported but provided arguments are invalid.
        */
     };
   };
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -69,6 +69,12 @@ static const gchar *regex_accl_elem_utils[] = {
 #define g_strfreev_const(x) g_strfreev((void*)(long)(x))
 
 /**
+ * @brief parse user given string to extract list of accelerators based on given regex
+ */
+static GList *parse_accl_hw_all (const gchar * accelerators,
+    const gchar ** supported_accelerators);
+
+/**
  * @brief GstTensorFilter properties.
  */
 enum
@@ -308,7 +314,7 @@ verify_model_path (const GstTensorFilterPrivate * priv)
 {
   const GstTensorFilterProperties *prop;
   gboolean ret = TRUE;
-  int i;
+  int verify_model_path = 0, i;
 
   if (priv == NULL)
     return FALSE;
@@ -321,7 +327,13 @@ verify_model_path (const GstTensorFilterPrivate * priv)
   if (g_strcmp0 (prop->fwname, "custom-easy") == 0)
     return TRUE;
 
-  if ((prop->model_files != NULL) && (priv->fw->verify_model_path == TRUE)) {
+  if (GST_TF_FW_V0 (priv->fw)) {
+    verify_model_path = priv->fw->verify_model_path;
+  } else if (GST_TF_FW_V1 (priv->fw)) {
+    verify_model_path = priv->info.verify_model_path;
+  }
+
+  if ((prop->model_files != NULL) && (verify_model_path == TRUE)) {
     for (i = 0; i < prop->num_models; i++) {
       if (!g_file_test (prop->model_files[i], G_FILE_TEST_IS_REGULAR)) {
         g_critical ("Cannot find the model file [%d]: %s\n",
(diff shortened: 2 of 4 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -668,8 +679,12 @@ gst_tensor_filter_common_free_property (GstTensorFilterPrivate * priv)
   prop = &priv->prop;
 
   g_free_const (prop->fwname);
-  /** TODO: free on the basis of the version of GstTensorFilterFramework */
-  g_free_const (prop->accl_str);
+  if (GST_TF_FW_V0 (priv->fw)) {
+    g_free_const (prop->accl_str);
+  } else if (GST_TF_FW_V1 (priv->fw)) {
+    if (prop->hw_list)
```
dongju-chae: uncessary?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -680,6 +695,55 @@ gst_tensor_filter_common_free_property (GstTensorFilterPrivate * priv)
...
+  }
+  accl_support[info->num_hw] = NULL;
+
+  /** Parse the user given h/w accelerators intersected with supported h/w */
+  match_accl = parse_accl_hw_all (accelerators, accl_support);
+  g_free (accl_support);
+
+  /** Convert the GList to regular array */
+  prop->num_hw = g_list_length (match_accl);
+  if (prop->hw_list)
```
dongju-chae: same here

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Done.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Done.

myungjoo:
To avoid the unexpected unit test errors, you will need to rebase with the recent master; there has been an undapte in GStreamer, which affected the behaviors of unit tests.

kparichay:
> To avoid the unexpected unit test errors, you will need to rebase with the recent master; there has been an undapte in GStreamer, which affected the behaviors of unit tests.

Thanks for the heads up. Updated.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -321,7 +327,13 @@ verify_model_path (const GstTensorFilterPrivate * priv)
   if (g_strcmp0 (prop->fwname, "custom-easy") == 0)
     return TRUE;
 
-  if ((prop->model_files != NULL) && (priv->fw->verify_model_path == TRUE)) {
+  if (GST_TF_FW_V0 (priv->fw)) {
+    verify_model_path = priv->fw->verify_model_path;
+  } else {
+    verify_model_path = priv->info.verify_model_path;
```
myungjoo: This is not very extensible approach.
do:
```
else if {GST_TF_FW_V1 (priv->fw)) {
...
} else {
  FAIL: NOT SUPPORTED API Version.
}
```

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -425,8 +432,7 @@ nnstreamer_filter_probe (GstTensorFilterFramework * tfsp)
 
   g_return_val_if_fail (nnstreamer_filter_validate (tfsp), FALSE);
 
-  if (checkGstTensorFitlerFrameworkVersion (tfsp->version,
-          GST_TENSOR_FILTER_FRAMEWORK_V0)) {
+  if (GST_TF_FW_V0 (tfsp)) {
```
myungjoo: handle !V0 && !V1 case.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -491,13 +497,17 @@ gst_tensor_filter_allocate_in_invoke (GstTensorFilterPrivate * priv)
...
+    allocate_in_invoke = priv->fw->allocate_in_invoke;
+    if (allocate_in_invoke == TRUE && priv->fw->allocateInInvoke) {
+      if (priv->fw->allocateInInvoke (&priv->privateData) == 0) {
+        allocate_in_invoke = TRUE;
+      } else {
+        allocate_in_invoke = FALSE;
+      }
     }
+  } else {
+    allocate_in_invoke = priv->info.allocate_in_invoke;
```
myungjoo: same here. !V0 && !V1 case

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -668,8 +679,11 @@ gst_tensor_filter_common_free_property (GstTensorFilterPrivate * priv)
   prop = &priv->prop;
 
   g_free_const (prop->fwname);
-  /** TODO: free on the basis of the version of GstTensorFilterFramework */
-  g_free_const (prop->accl_str);
+  if (GST_TF_FW_V0 (priv->fw)) {
+    g_free_const (prop->accl_str);
+  } else if (GST_TF_FW_V1 (priv->fw)) {
+    g_free (prop->hw_list);
+  }
```
myungjoo: ditto

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: The successful call to `nnstreamer_filter_validate` ensures that the `tfsp` is either `V0` or `V1`.

kparichay: All the accesses to `priv->fw` happen after the framework has been probed. This means that the filter has been validated. So, this means that the framework version has been checked, and in this case will be limited to V0 and V1.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Will update to 
```
else if {GST_TF_FW_V1 (priv->fw)) {
...
}
```

Is the else case needed. More details on this in https://github.com/nnsuite/nnstreamer/pull/2130#discussion_r389509768

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -764,12 +862,17 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
         }
 
         prop->input_meta.num_tensors = num_dims;
+      } else if (value) {
+        /** Update the properties */
+        if (GST_TF_FW_V0 (priv->fw)) {
+          g_assert (!prop->input_configured);
+        } else {
+          /** TODO: update the properties */
```
myungjoo: You will need to re-negotiate the GST-Caps if you allow this.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -800,12 +909,18 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
         }
 
         prop->input_meta.num_tensors = num_types;
+      } else if (value) {
+        /** Update the properties */
+        if (GST_TF_FW_V0 (priv->fw)) {
+          /* Once configured, it cannot be changed in runtime */
+          g_assert (!prop->input_configured);
+        } else {
+          /** TODO: update the properties */
```
myungjoo: You'll need to re-negotiate here, too. I'm not sure how that can be accomplished.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -837,13 +958,19 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
         }
 
         prop->input_meta.num_tensors = num_names;
+      } else if (value) {
+        /** Update the properties */
+        if (GST_TF_FW_V0 (priv->fw)) {
+          /* Once configured, it cannot be changed in runtime */
+          g_assert (!prop->input_configured);
+        } else {
+          /** TODO: update the properties */
```
myungjoo: Unlike dim/type, this doesn't need re-negotiation. Thus, it appears tractible. However, I don't see any usage cases here.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -856,74 +983,104 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
       break;
     case PROP_CUSTOM:
-      /* In case updated custom properties in runtime! */
-      g_free_const (prop->custom_properties);
-      prop->custom_properties = g_value_dup_string (value);
+      if (!priv->prop.fw_opened) {
+        g_free_const (prop->custom_properties);
+        prop->custom_properties = g_value_dup_string (value);
+      } else {
+        /** TODO: update the properties */
```
myungjoo: 1. uh... why have you added this "if" statement?
2. we may need to send "event" for V1 with this event.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -856,74 +983,104 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
       }
 
-      /** TODO: set this property based on the framework version V0 or V1 */
-      prop->accl_str = g_value_dup_string (value);
+      if (GST_TF_FW_V1 (priv->fw)) {
+        gst_tensor_filter_parse_accelerator (priv, &priv->prop, accelerators);
+        g_free (accelerators);
+      } else {
+        prop->accl_str = accelerators;
+      }
```
myungjoo: we may need event handler here as well.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -856,74 +983,104 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
       break;
     }
     case PROP_IS_UPDATABLE:
     {
-      if (priv->fw->reloadModel != NULL)
-        priv->is_updatable = g_value_get_boolean (value);
+      if (GST_TF_FW_V0 (priv->fw) && priv->fw->reloadModel == NULL) {
+        break;
+      } else if (GST_TF_FW_V1 (priv->fw) &&
+          priv->fw->eventHandler (prop, priv->privateData, RELOAD_MODEL, NULL)
```
myungjoo: You are not supposed to send "RELOAD" even with this!

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -856,74 +983,104 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
+            prop->input_meta.num_tensors != num_layouts) {
+          g_warning ("Invalid input-layout, given param does not fit.");
+        }
+
+        prop->input_meta.num_tensors = num_layouts;
+      } else if (value) {
+        /** Update the properties */
+        if (GST_TF_FW_V0 (priv->fw)) {
+          /* Once configured, it cannot be changed in runtime */
+          g_assert (!prop->input_configured);
```
myungjoo: No. 
This does not incur any GST-CAP renegotiation. You may chage it in runtime w/ V0, I suppose.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: This call to "RELOAD" is to verify if this event is supported. This call will always fail as the passed parameter is NULL, its just that it shouldn't fail with `-ENOENT`.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: This follows the same issue as custom properties.
How will the framework know if the layout has been changed for V0?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: eventHandler is called if the framework is already opened. 
If the framework is not opened, then this information is passed with GstTensorFilterProperties to its open callback.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: 1. For V0, once the framework is opened, there is no mechanism to notify the framework about the update in custom properties. Updating here directly will require each framework to perform check on updates in custom properties at every invoke (python filter uses custom_properties when the filter is opened. Directly changing custom properties will not reflect the corresponding update in the filter). Hence, for now, I have disabled updating custom properties for V0. Please advice.
2. Updated to send the event for V1.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Removed.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Re-negotiation can be done by sending a GST_EVENT_RECONFIGURE event upstream ([link](https://gstreamer.freedesktop.org/documentation/plugin-development/advanced/negotiation.html?gi-language=c#upstream-caps-renegotiation)).

Yes, this is removed for now. Re-negotiation can be supported later.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Yes, this is removed for now. Re-negotiation can be supported later.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -708,18 +770,37 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
       const GstTensorFilterFramework *fw;
 
       if (priv->fw != NULL) {
-        /* close old framework */
-        gst_tensor_filter_common_close_fw (priv);
+        if (priv->prop.fwname != fw_name) {
```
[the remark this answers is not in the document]
kparichay: aah, good catch. 
Updated.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
```diff
@@ -385,7 +390,7 @@ typedef struct _GstTensorFilterFramework
 
       int (*getFrameworkInfo) (const GstTensorFilterProperties * prop,
           void *private_data, GstTensorFilterFrameworkInfo *fw_info);
-      /**< Mandatory callback. Get the framworks statically determined info. Argument 'private_data' can be NULL. If provided 'private_data' is not NULL, then some info, such as 'allocate_in_invoke', can be updated based on the model being used (inferred from the 'private_data' provided). This updated info is useful for custom filter, as some custom filter's ability to support 'allocate_in_invoke' depends on the opened model.
+      /**< Mandatory callback. Get the frameworks statically determined info. Argument 'private_data' can be NULL. If provided 'private_data' is not NULL, then some info, such as 'allocate_in_invoke', can be updated based on the model being used (inferred from the 'private_data' provided). This updated info is useful for custom filter, as some custom filter's ability to support 'allocate_in_invoke' depends on the opened model.
```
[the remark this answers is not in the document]
kparichay: I am not sure if I understand. Should I break this into smaller lines?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -764,12 +865,13 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
         }
 
         prop->input_meta.num_tensors = num_dims;
+      } else if (value) {
+        /** Once configured, it cannot be changed in runtime for now */
+        g_assert (!prop->input_configured);
```
kparichay: @myungjoo is this assert needed or this can be just a warning?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -729,29 +817,56 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
...
+      GstTensorFilterProperties _prop;
 
       g_assert (model_files);
+      _prop.model_files = NULL;
+
+      if (prop->fw_opened) {
+        /** Store a copy of the original prop in case the reload fails */
+        memcpy (&_prop, prop, sizeof (GstTensorFilterProperties));
+        _prop.model_files =
+            (const gchar **) g_strdupv ((gchar **) prop->model_files);
```
myungjoo: if you allocate new memory for ```_prop.model_files```, you should free it somewhere.
You've simply done ```_prop.model_files = NULL;``` at LINE 823 instead.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: We should prohibit set-prop on input dimension if it's already configured (GST-CAP negotiation completed).

It's fine to ignore configuring the value with "VISIBLE" warning; the warning level should be HIGH. (use @jaeyun-jung's recent log APIs)

In fact, we need to limit g_assert for internal error only for higher code quality. External errors (errors incurred by users and applications) should not do assert, but emit proper error messages (visiable and explainable).

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: Then, we need to clearly state that it should return -EINVAL if it supports RELOAD but the given data is NULL (invalid) at the doxygen entry of the API (nnstreamer_plugin_api_filter.h)

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: ah.. yes, that's right. Let's prohibit it for V0.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Memory for `_prop.model_files` is freed at LINE 857 (when reload is successful).
If reload is failed, then `prop.model_files` is freed and `_prop.model_files` is used.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Replaced `g_assert` with meaningful comments.

> HIGHEST for the first few violations and WARN for the rest or for a few seconds for the same errors; for throttling, we will need to upgrade the logging mechanism in the future.

That is a good suggestion and should be done.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Added.

gst/nnstreamer/nnstreamer_plugin_api_filter.h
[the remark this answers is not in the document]
myungjoo: Nah it's ok for now.

PR #2133, opened 2020-03-04
[meson] Add dependency of gtest

kparichay:
Some of the unit-tests depend on gtest
However, their dependency is not checked leading to build failure with incorrect messages
Added dependency check of gtest before building those unittests

**Self evaluation:**

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -150,8 +150,7 @@ if have_movidius_ncsdk2
   subdir('nnstreamer_filter_mvncsdk2')
 endif
 
-
-if get_option('enable-cppfilter')
+if get_option('enable-cppfilter') and gtest_dep.found()
   subdir('cpp_methods')
 endif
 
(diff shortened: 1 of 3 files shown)


tests/meson.build
```diff
@@ -161,7 +161,7 @@ if get_option('enable-movidius-ncsdk2')
 endif
 
 
-if get_option('enable-cppfilter')
+if get_option('enable-cppfilter') and gtest_dep.found()
```
dongju-chae: How about checking `gtest_dep.found()` inside each sub-directory? because such a dependency depends on each filter's test codes.. and for uniformity.

kparichay: Updated.

kparichay:
> Wouldn't it be more readable if we add to tests/meson.build:
> 
> ```
> ...
> if get_option("whatsoever") and gtest_dep.found()
> ...
> endif
> ```
> 
> and in tests/whsoever/meson.build
> 
> ```
> if not gtest_dep.found()
>   FATAL ERROR
> endif
> ```

DJ requested the check to be inside each sub-directory https://github.com/nnsuite/nnstreamer/pull/2133#discussion_r388062816.
I will roll back the commit to the previous version to apply as requested.

kparichay:
@myungjoo Applied your comments.

Issue #2134, opened 2020-03-04
Optimize invoke's critical path


kparichay:
The critical path of `invoke` for all the sub-plugins and the API's needs optimization. This primarily includes moving all the unnecessary computation to other callbacks.

myungjoo:
This is getting critical.. :)

myungjoo:
@jaeyun-jung has done it for Android/JAVA-API.

Revisit this when C-API has similar issues

PR #2135, opened 2020-03-04
[tflite] optimize invoke critical path

kparichay:
Move unnecessary computations out of the critical path to maximize invoke performance for tensorflow-lite.

See also: #2134 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -99,11 +99,13 @@ private:
 
   GstTensorsInfo inputTensorMeta;  /**< The tensor info of input tensors */
   GstTensorsInfo outputTensorMeta;  /**< The tensor info of output tensors */
+  std::vector <TfLiteTensor *> input_tensor_ptr;
+  std::vector <TfLiteTensor *> output_tensor_ptr;
 
   tensor_type getTensorType (TfLiteType tfType);
   int getTensorDim (int tensor_idx, tensor_dim dim);
   int setTensorProp (const std::vector<int> &tensor_idx_list,
-      GstTensorsInfo * tensorMeta);
+      GstTensorsInfo * tensorMeta, std::vector <TfLiteTensor *>& tensor_ptrs);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -380,17 +372,20 @@ TFLiteInterpreter::getTensorDim (int tensor_idx, tensor_dim dim)
  */
 int
 TFLiteInterpreter::setTensorProp (const std::vector<int> &tensor_idx_list,
-    GstTensorsInfo * tensorMeta)
+    GstTensorsInfo * tensorMeta, std::vector <TfLiteTensor *>& tensor_ptrs)
```
dongju-chae: It seems that `tensors_ptrs` is a member variable of `TFLiteInterpreter`. Why this should be passed as an argument?

kparichay: This function is commonly used for both inputs and outputs.
Hence tensor_ptrs is passed which is different for input or output.

dongju-chae: Ah, okay

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -181,30 +183,18 @@ TFLiteInterpreter::invoke (const GstTensorMemory * input,
...
-  TfLiteTensor *tensor_ptr;
   TfLiteStatus status;
   bool size_mismatch = FALSE;
 
-  for (unsigned int i = 0; i < outputTensorMeta.num_tensors; ++i) {
-    tensor_idx = interpreter->outputs ()[i];
-    tensor_ptr = interpreter->tensor (tensor_idx);
-
-    if (tensor_ptr->bytes != output[i].size)
+  for (unsigned int i=0; i < input_tensor_ptr.size(); i++) {
```
wooksong: i = 0?

kparichay: Updated indentation for the changes in the file.

myungjoo:
@kparichay The assumption that the location of input/output buffers are static is NOT TRUE if it is multi-tensor, which incurs #2145. @anyj0527 will create a PR that reverts this and adds a regression test case to prevent this issue.

kparichay:
> @kparichay The assumption that the location of input/output buffers are static is NOT TRUE if it is multi-tensor, which incurs #2145. @anyj0527 will create a PR that reverts this and adds a regression test case to prevent this issue.

That is quite a strange behavior.

PR #2143, opened 2020-03-09
[Api] hw enum for sr-npu

jaeyun-jung:
change enum val for npu-sr, and remove enum npu-srcn.


jaeyun-jung:
This may require new ACR process.

PR #2149, opened 2020-03-11
[Api/Single] add property key custom

jaeyun-jung:
update function to set/get the custom property from filter.


jaeyun-jung:
added testcase for null params

jaeyun-jung:
TC failed, need improvement.
TF_FW_V0 does not allow changing the property after in/out configured.

PR #2151, opened 2020-03-11
[Android] sync to cs-api update

jaeyun-jung:
1. change method name to set/get the property
2. handle empty node name
3. define method to get model file in testcase

### DIFF
--- a/api/android/api/src/main/jni/nnstreamer-native-api.c
+++ b/api/android/api/src/main/jni/nnstreamer-native-api.c
@@ -30,7 +30,6 @@ GST_PLUGIN_STATIC_DECLARE (amcsrc);
 extern void init_filter_cpp (void);
 extern void init_filter_custom (void);
 extern void init_filter_custom_easy (void);
-extern void init_filter_tflite (void);
 extern void init_dv (void);
 extern void init_bb (void);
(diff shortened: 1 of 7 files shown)


jaeyun-jung:
Waiting for ACR process, on API review.

api/android/api/src/main/jni/nnstreamer-native-api.c
```diff
@@ -30,7 +30,6 @@ GST_PLUGIN_STATIC_DECLARE (amcsrc);
 extern void init_filter_cpp (void);
 extern void init_filter_custom (void);
 extern void init_filter_custom_easy (void);
-extern void init_filter_tflite (void);
```
myungjoo: This should be optionally included/excluded by build options along with nn-runtime, snap, and snpe(in the future). The clients want to have multiple versions of .aar depending on the requirements of each application.

Or are you going to make separate .so for each "optional" subplugins in .aar?

jaeyun-jung: I added another PR, build option for subplugins.

myungjoo: If that's the case, this is good for now.

PR #2156, opened 2020-03-12
[Android/Build] clean build script

jaeyun-jung:
1. add flag to disable tf-lite.
2. update build script for each build option.

### DIFF
--- a/api/android/api/src/main/jni/Android.mk
+++ b/api/android/api/src/main/jni/Android.mk
@@ -16,20 +16,38 @@ else
 $(error Target arch ABI not supported: $(TARGET_ARCH_ABI))
 endif
 
-# SNAP (Samsung Neural Acceleration Platform)
-ENABLE_SNAP := false
-
 #------------------------------------------------------
 # API build option
 #------------------------------------------------------
 NNSTREAMER_API_OPTION := all
 
+# tensorflow-lite (nnstreamer tf-lite subplugin)
+ENABLE_TF_LITE := false
+
+# SNAP (Samsung Neural Acceleration Platform)
+ENABLE_SNAP := false
+
+NNS_API_FLAGS := -DVERSION=\"$(NNSTREAMER_VERSION)\"
+NNS_API_STATIC_LIBS :=
+
+ifeq ($(NNSTREAMER_API_OPTION),single)
+NNS_API_FLAGS += -DNNS_SINGLE_ONLY=1
+endif
+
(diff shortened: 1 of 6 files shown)


api/android/build-android-lib.sh
```diff
@@ -13,39 +13,49 @@
...
 # For example, to build library with core plugins for arm64-v8a
 # ./build-android-lib.sh --api_option=lite --target_abi=arm64
 #
 
-# API build option ('lite' to build with GStreamer core plugins)
+# API build option
+# 'all' : default
+# 'lite' : with GStreamer core plugins
+# 'single' : no plugins, single-shot only
+# 'internal' : no plugins, single-shot only, disable SNAP and tf-lite
```
dongju-chae: How about this? `all`, `core`, `single`, `single-internal`

jaeyun-jung: Build for internal release : I am not sure internal release includes only single-shot functions. (TBD)
About the 'lite', in the past we determined the lib name as nnstreamer-lite with minimized plugins.

jaeyun-jung:
2020-03-17 PR updated
- cleanup tf-lite module in mk

jaeyun-jung:
2020-03-18 PR updated
- set default abi arm64-v8a
- update document

api/android/api/src/main/jni/Android.mk
```diff
@@ -82,4 +93,9 @@ GSTREAMER_INCLUDE_CA_CERTIFICATES := no
 
 include $(GSTREAMER_NDK_BUILD_PATH)/gstreamer-1.0.mk
 
+#------------------------------------------------------
+# NDK cpu-features
+#------------------------------------------------------
+ifeq ($(ENABLE_TF_LITE), true)
```
wooksong: Could we eliminate this ifeq (99) and above ifeq (69)? For example, 

```
LOCAL_EXTRA_CFLAGS :=
ifeq ($(ENABLE_TF_LITE), true)
include $(LOCAL_PATH)/Android-tensorflow-lite.mk
LOCAL_EXTRA_CFLAGS += -DENABLE_TF_LITE=1
endif
...omission...
LOCAL_CFLAGS += LOCAL_EXTRA_CFLAGS 
```
How about that?

jaeyun-jung: added new flags as you commented (NNS_API_FLAGS and NNS_API_STATIC_LIBS)

PR #2120, opened 2020-02-21
[TFLite] check supporting datatype

helloahn:
check supporting datatypes with meson_option and return the valid  datatype

Signed-off-by: HyoungJoo Ahn <hello.ahn@samsung.com>

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped

PR #2157, opened 2020-03-13
[Meson/Filter/TFLite] Use has_type to check types tflite supports 

wooksong:
This PR mainly concerns checking which data types the installed TensorFlow Lite supports at build time. In detail, the following commits are included:

- [Meson/TFLite] Check version in ext/nns/filter/meson.build 
  - Move a specific code for the tensor filter extension for tflite from meson.build in the source root to the tensor filter extension directory
- [Meson/Filter/TFLite] Use has_type to check types tflite supports 
  - Instead of checking which data types the installed TensorFlow Lite
supports via hardcoded version information, this patch uses has_type,
which is a function of the compiler object in meson, to check that.

See also: #2139, #2120
cc: @helloahn

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -104,28 +104,19 @@ if get_option('enable-tensorflow-lite')
     endif
   endif
 
-  TFLite_version = tflite_dep.version().split('.')
-  TFLite_major = TFLite_version[0].to_int()
-  TFLite_minor = TFLite_version[1].to_int()
-
   tflite_compile_args = []
-  if TFLite_major == 1
-    if TFLite_minor >= 10
-      tflite_compile_args += '-DTFLITE_INT16=1'
-      tflite_compile_args += '-DTFLITE_COMPLEX64=1'
-      message('data type [INT16] is added for TFLite')
-      message('data type [COMPLEX64] is added for TFLite')
-    endif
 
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -104,6 +104,27 @@ if get_option('enable-tensorflow-lite')
     endif
   endif
 
+  tflite_compile_args = []
+
+  if cxx.has_type('kTfLiteInt8', prefix : '#include <tensorflow/contrib/lite/model.h>')
```
helloahn: Is there any reason that you picked `model.h` as a prefix?
+ for version 1.15 & 1.13, the path of `model.h` is `<tensorflow/lite/model.h>`

wooksong: @helloahn This is because you include that header in the tensor filter extension for tflite:
https://github.com/nnsuite/nnstreamer/blob/c000f9e3d93c213291f531af7e6c98fff1e230b8/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc#L35

helloahn: > As you already knew, we have provided include/tensorflow/contrib/* for backward compatibility.

Oh, I forgot about it. Thanks for reminding me.

PR #2160, opened 2020-03-13
[Filter/Conf] use gmodule to open so file

jaeyun-jung:
use gmodule to open and load sub-plugins.


api/android/api/src/main/jni/Android.mk
```diff
@@ -74,7 +74,7 @@ GSTREAMER_NDK_BUILD_PATH := $(GSTREAMER_ROOT)/share/gst-android/ndk-build/
 include $(LOCAL_PATH)/Android-gst-plugins.mk
 
 GSTREAMER_PLUGINS        := $(GST_REQUIRED_PLUGINS)
-GSTREAMER_EXTRA_DEPS     := $(GST_REQUIRED_DEPS) gio-2.0
+GSTREAMER_EXTRA_DEPS     := $(GST_REQUIRED_DEPS) gio-2.0 gmodule-2.0
```
myungjoo: Does this remove the direct-dependency on dlfcn/ldl ?

jaeyun-jung: I removed dl-dep in nnstreamer.

myungjoo: Oh.. then, it's great!

PR #2167, opened 2020-03-16
[Filter/Sub] preserved name

jaeyun-jung:
do not allow 'auto' as subplugin name.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
@@ -311,4 +311,26 @@ public class APITestCustomFilter {
             /* expected */
         }
     }
+
+    @Test
+    public void testRegisterPreservedName_n() {
+        try {
+            CustomFilter.registerCustomFilter("auto",
+                new CustomFilter.CustomFilterCallback() {
+                    @Override
+                    public TensorsInfo getOutputInfo(TensorsInfo in) {
+                        return in;
+                    }
+
+                    @Override
+                    public TensorsData invoke(TensorsData in) {
+                        return in;
+                    }
+                });
+
+            fail();
+        } catch (Exception e) {
+            /* expected */
+        }
+    }
 }
--- a/gst/nnstreamer/nnstreamer_subplugin.c
+++ b/gst/nnstreamer/nnstreamer_subplugin.c
@@ -184,6 +184,12 @@ register_subplugin (subpluginType type, const char *name, const void *data)
       return FALSE;
   }
 
+  /* check the sub-pugin name */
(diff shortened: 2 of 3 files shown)


jaeyun-jung:
Discussion: consider filter-fw auto detection from @gichan-jang

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
```diff
@@ -311,4 +311,26 @@ public TensorsData invoke(TensorsData in) {
             /* expected */
         }
     }
+
+    @Test
+    public void testRegisterPreservedName_n() {
+        try {
+            CustomFilter.registerCustomFilter("auto",
```
myungjoo: Even if tensor-filter subplugin "auto" is prohibited, a custom plugin "auto" should be allowed. If it's customfilter "auto", its subplugin name will be "custom" or "custom-cpp" or "custom-easy" or so.

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -184,6 +184,12 @@ register_subplugin (subpluginType type, const char *name, const void *data)
       return FALSE;
   }
 
+  /* check the sub-pugin name */
+  if (g_ascii_strcasecmp (name, "auto") == 0) {
+    g_critical ("Failed, the name %s is not allowed.", name);
+    return FALSE;
+  }
+
```
myungjoo: LGTM.

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
jaeyun-jung: custom-filter in java is not a .so file (run without model file). It runs without so, and the fw name would be auto in this case.

myungjoo: So, it is a custom filter with "auto" as its custom filter name?

What if we have two java custom filters in a single pipeline, then?

jaeyun-jung: If developer needs two custom-filters, implement two interfaces with different names and register them.
```
CustomFilter.registerCustomFilter("custom1", new CustomFilter.CustomFilterCallback() { ..
CustomFilter.registerCustomFilter("custom2", new CustomFilter.CustomFilterCallback() { ..

pipeline = "... ! tensor_filter framework=custom1 ! tensor_filter framework=custom2 ! ..."
```

jaeyun-jung:
2020-03-17 PR updated
- add test for preserved subplugin name

PR #2174, opened 2020-03-17
[Tests/Filter/Reload] Check existence of TFLite filters before start

wooksong:
The test cases for reload models require the tensor filter extension for Tensorflow Lite. As other test cases do, this patch adds a procedure that checks existence of the tensorflow-lite filters before starting the test.

See also: #2164 

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_filter_reload/runTest.sh
+++ b/tests/nnstreamer_filter_reload/runTest.sh
@@ -14,11 +14,60 @@ if [[ "$SSATAPILOADED" != "1" ]]; then
 "
 fi
 
+if [ -z ${SO_EXT} ]; then
+    SO_EXT="so"
+fi
+
 # This is compatible with SSAT (https://github.com/myungjoo/SSAT)
 testInit $1
 
 # NNStreamer and plugins path for test
 PATH_TO_PLUGIN="../../build"
+# This test is valid only with the tensor filter extension for tflite
+if [[ -d $PATH_TO_PLUGIN ]]; then
+    ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
+    if [[ -d ${ini_path} ]]; then
+        check=$(ls ${ini_path} | grep tensorflow-lite.${SO_EXT})
+        if [[ ! $check ]]; then
+            echo "Cannot find tensorflow-lite shared lib"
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_filter_reload/runTest.sh
```diff
@@ -19,6 +19,51 @@ testInit $1
 
 # NNStreamer and plugins path for test
 PATH_TO_PLUGIN="../../build"
+# This test is valid only with the tensor filter extension for tflite
+if [[ -d $PATH_TO_PLUGIN ]]; then
+    ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
+    if [[ -d ${ini_path} ]]; then
+        check=$(ls ${ini_path} | grep tensorflow-lite.so)
```
jaeyun-jung: better to set SO_EXT as you fixed today?

wooksong: Ah:( my bad...

tests/nnstreamer_filter_reload/runTest.sh
```diff
@@ -14,11 +14,60 @@ if [[ "$SSATAPILOADED" != "1" ]]; then
...
+if [ -z ${SO_EXT} ]; then
+    SO_EXT="so"
+fi
+
 # This is compatible with SSAT (https://github.com/myungjoo/SSAT)
 testInit $1
 
 # NNStreamer and plugins path for test
 PATH_TO_PLUGIN="../../build"
+# This test is valid only with the tensor filter extension for tflite
```
jijoongmoon: It looks much better for me. Sure. we have not to use relative directory path.

tests/nnstreamer_filter_reload/runTest.sh
```diff
@@ -14,11 +14,60 @@ if [[ "$SSATAPILOADED" != "1" ]]; then
...
+if [[ -d $PATH_TO_PLUGIN ]]; then
+    ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
+    if [[ -d ${ini_path} ]]; then
```
jijoongmoon: It would be better to make the path clear!

PR #2186, opened 2020-03-20
[AHUB/523198] Fix null pointer dereferencing

myungjoo:
This fixes:
WID:16646516 After having been compared to NULL value at tensor_filter_common.c:773, pointer 'priv->fw' is dereferenced at tensor_filter_common.c:791.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -788,7 +788,7 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
       if (fw) {
         /** Get framework info for v1 */
         if (GST_TF_FW_V1 (fw) &&
-            priv->fw->getFrameworkInfo (prop, NULL, &priv->info) < 0) {
+            fw->getFrameworkInfo (prop, NULL, &priv->info) < 0) {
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -788,7 +788,7 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
       if (fw) {
         /** Get framework info for v1 */
         if (GST_TF_FW_V1 (fw) &&
-            priv->fw->getFrameworkInfo (prop, NULL, &priv->info) < 0) {
+            fw->getFrameworkInfo (prop, NULL, &priv->info) < 0) {
```
gichan-jang: This is a question to improve my understanding.
Is there a case where the potive value is returned? 
The description only says '0 if OK. non-zero if error.'

myungjoo: According to the description, positive value == undefined behavior.
This is a callback defined by filter subplugins; thus, we cannot guarantee it.

Note that this is a common practice in C libraries.

Anyway, in the future, we may generate fatal errors if it returns > 0 so that subplugin writers may fix it right away. You will need some survey (on well-known and well-written C programs) to determine if we need to generate such or not.

PR #2190, opened 2020-03-20
[Android/Test] check nnfw availability

jaeyun-jung:
Before running the test, check nnfw (tf-lite or snap) is available and test can be run.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
@@ -368,6 +368,11 @@ public class APITestPipeline {
 
     @Test
     public void testRunModel() {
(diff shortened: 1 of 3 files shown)


myungjoo:
Off-Topic: Be aware that NNFW is probably going to be accessed via SNAP interface in Android.

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
```diff
@@ -368,6 +368,11 @@ public void onNewDataReceived(TensorsData data) {
 
     @Test
     public void testRunModel() {
+        if (!NNStreamer.isAvailable(NNStreamer.NNFWType.TENSORFLOW_LITE)) {
+            /* cannot run the test */
```
jijoongmoon: This makes tests much stable. LGTM.

PR #2191, opened 2020-03-20
[Android/Build] tf-lite data type

jaeyun-jung:
add flags for tf-lite data types.


api/android/api/src/main/jni/Android-tensorflow-lite.mk
```diff
@@ -47,10 +47,13 @@ include $(PREBUILT_STATIC_LIBRARY)
 #------------------------------------------------------
 include $(CLEAR_VARS)
 
+# define types in tensorflow-lite sub-plugin
+TF_LITE_FLAGS := -DTFLITE_INT8=1 -DTFLITE_INT16=1
```
myungjoo: This would be better defined where TF-LITE enable/disable is defined because TF-LITE version will be "known" when someone enable/disable TF-LITE build. That'd be around LINE 40-41 of ```/api/android/api/src/main/jni/Android.mk```

jaeyun-jung: Thanks, I moved this to android.mk.

PR #2193, opened 2020-03-21
[Infra] Move exported headers to .../include/

myungjoo:
Move all exported headers to /gst/nnstreamer/include/
Keep common internal headers at /gst/nnstreamer/

This prepares c++ class headers for subplugins and
"multi-devel" packages for different purposese.


gst/nnstreamer/meson.build
```diff
@@ -1,4 +1,4 @@
-nnstreamer_inc = include_directories('.')
+nnstreamer_inc = include_directories('.', './include/')
```
jaeyun-jung: maybe android build error 

https://github.com/nnsuite/nnstreamer/blob/7b094306f9e5f9f3cae78c0224644c034da5a2ae/jni/nnstreamer.mk#L18

myungjoo: Fixed.

gst/nnstreamer/include/meson.build
```diff
@@ -0,0 +1,25 @@
...
+foreach h : nnst_common_headers
+  nnstreamer_headers += join_paths(meson.current_source_dir(), h)
+endforeach
+
+# Install headers into /{includedir}/nnstreamer
+install_headers(nnstreamer_headers,
+  subdir: 'nnstreamer'
+)
+
+if get_option('enable-filter-cpp-class')
```
jaeyun-jung: Question: tensor_filter_cpp.hh
is it also required to move include dir and install if enable-cppfilter is true?

myungjoo: Yes. it will be installed to include dir.

myungjoo:
Android API header install script updated.

PR #2184, opened 2020-03-19
[C-API] add testcases for a function: `ml_single_invoke_dynamic()`

helloahn:
the success/fail cases are added for the api `ml_single_invoke_dynamic()`

Signed-off-by: HyoungJoo Ahn <hello.ahn@samsung.com>

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped

PR #2201, opened 2020-03-23
[Wait for #2184][CAPI/Unittest] improve the possibility of a memory leak 

helloahn:
free the PATHs before checking `ASSERT`

it resolves https://github.com/nnsuite/nnstreamer/pull/2184#discussion_r395413523

**Self evaluation:**

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cc
+++ b/tests/tizen_capi/unittest_tizen_capi.cc
@@ -15,6 +15,7 @@
 #include <nnstreamer-capi-private.h>
 
 #define SINGLE_DEF_TIMEOUT_MSEC 10000
+#define MAX_PATH_LEN 2048
 
 /**
  * @brief Struct to check the pipeline state changes.
@@ -1039,7 +1040,7 @@ TEST (nnstreamer_capi_switch, dummy_01)
   EXPECT_EQ (status, ML_ERROR_NONE);
 
   if (node_list) {
-    gchar *name = NULL;
+  gchar *name = NULL;
     guint idx = 0;
 
     while ((name = node_list[idx]) != NULL) {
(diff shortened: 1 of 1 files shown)


tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1039,7 +1040,7 @@ TEST (nnstreamer_capi_switch, dummy_01)
   EXPECT_EQ (status, ML_ERROR_NONE);
 
   if (node_list) {
-    gchar *name = NULL;
+  gchar *name = NULL;
```
wooksong: Please check indentation here.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1133,7 +1134,7 @@ TEST (nnstreamer_capi_switch, dummy_02)
...
```
wooksong: ditto

myungjoo:
Do you really need to free things before calling assert?
Recent versions of coverity complain this, but I think Coverity is wrong with it.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1836,14 +1838,16 @@ TEST (nnstreamer_capi_singleshot, invoke_02)
   int status;
 
   const gchar *root_path = g_getenv ("NNSTREAMER_BUILD_ROOT_PATH");
-  gchar *test_model;
+  gchar *tmp, test_model[MAX_PATH_LEN];
```
wooksong: I knew this means that
```c
gchar (*tmp), (test_model[MAX_PATH_LEN]);
```
However, for the readability, please consider
```c
gchar test_model[MAX_PATH_LEN];
gchar *tmp;
```

> It’s also important to comment data, whether they are basic types or derived types. To this end, use just one data declaration per line (no commas for multiple data declarations). This leaves you room for a small comment on each item, explaining its use. 
from https://www.kernel.org/doc/html/v4.10/process/coding-style.html

I think this aims not just room for comment but also the readability of code.

PR #2203, opened 2020-03-23
[Filter/API] Constify subplugin name. @open sesame 3/26 18:35

myungjoo:
Do not let subplugin authors handle alloc/dealloc name
for frequent getFrameworkInfo calls.
    
Allow them to point to a const string instead.
    

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -425,7 +425,7 @@ nnstreamer_filter_probe (GstTensorFilterFramework * tfsp)
 {
   GstTensorFilterFrameworkInfo info;
   GstTensorFilterProperties prop;
-  char *name = NULL;
+  const char *name = NULL;
```
leemgs: Flawless.

PR #2207, opened 2020-03-25
[Android/Build] Add comments on TF-Lite versions and types.

myungjoo:
INT8/INT16 may incur build errors with older TF-Lite.
For users with older TF-Lite, leave some comments.


myungjoo:
Ignoring apptest fail because this touches code-comments only.

PR #2208, opened 2020-03-25
[Decoder/protobuf] Add protobuf tensor_decoder sub-plugin

anyj0527:
Add tensor_decoder sub-plugin "protobuf" to support Protocol Buffers.

pipeline example:
```
gst-launch-1.0 videotestsrc num-buffers=40 ! video/x-raw,format=RGB,width=256,height=256,framerate=30/1 ! \
tensor_converter ! tensor_decoder mode=protobuf ! filesink location=./sample-output.pb
```


ext/nnstreamer/tensor_decoder/meson.build
```diff
@@ -114,3 +114,40 @@ static_library('nnstreamer_decoder_image_segment',
   install: true,
   install_dir: nnstreamer_libdir
 )
+
+# protobuf decoder
+if get_option('enable-protobuf')
+  # Compile protobuf schema: nnstreamer.proto
+  pb_comp = find_program('protoc', required: false)
```
myungjoo, line 121: Make it "required: true" and remove Line 123-125. The behavior will be the same :)

ext/nnstreamer/tensor_decoder/meson.build
```diff
@@ -114,3 +114,40 @@ static_library('nnstreamer_decoder_image_segment',
...
+  if not pb_comp.found()
+    error('protoc is not found')
+  endif
+
+  pb_gen = generator(pb_comp, output: ['@BASENAME@.pb.h', '@BASENAME@.pb.cc'], arguments : ['--proto_path=@CURRENT_SOURCE_DIR@', '--cpp_out=@BUILD_DIR@', '@INPUT@'])
+  pb_gen_src = pb_gen.process('nnstreamer.proto')
+
+  decoder_sub_protobuf_sources = [
+    'tensordec-protobuf.cc',
+  ]
```
myungjoo, line 132: maybe... ```decoder_sub_protobuf_sources += pb_gen_src``` is more neat? (remove pb_gen_src from below)

ext/nnstreamer/tensor_decoder/nnstreamer.proto
```diff
@@ -0,0 +1,34 @@
...
+    NNS_UINT8 = 5;
+    NNS_FLOAT64 = 6;
+    NNS_FLOAT32 = 7;
+    NNS_INT64 = 8;
+    NNS_UINT64 = 9;
+  
+    NNS_END = 10;
+  }
+  Tensor_type type = 2;
+  string dimension = 3;
```
myungjoo, line 22: Why declare dimension as string?

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Library General Public
+ * License as published by the Free Software Foundation;
+ * version 2.1 of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Library General Public License for more details.
```
myungjoo, line 13: For new files, please use SPDX identifier instead. Then, it'd look like: (from tensor-if code)
```
/* SPDX-License-Identifier: LGPL-2.1-only */
/**
 * GStreamer/NNStreamer Tensor-IF
 * Copyright (C) 2020 MyungJoo Ham <myungjoo.ham@samsung.com>
 */
```

ext/nnstreamer/tensor_decoder/meson.build
anyj0527: Sorry for this late. I updated this to be compatible with meson "feature"

ext/nnstreamer/tensor_decoder/meson.build
anyj0527: I updated it. Please see the very next line.

ext/nnstreamer/tensor_decoder/nnstreamer.proto
anyj0527: You're right. I changed it to `repeated uint32 dimension` .

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
anyj0527: Updated!

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  g_assert (outbuf);
+
+  size = tensors.ByteSizeLong ();
+
+  if (gst_buffer_get_size (outbuf) == 0) {
+    out_mem = gst_allocator_alloc (NULL, size, NULL);
+  } else {
+    if (gst_buffer_get_size (outbuf) < size) {
```
dongju-chae, line 120: It seems that `gst_buffer_get_size` is repeatedly called. How about declaring its variable? For example
```
outbuf_size = gst_buffer_get_size (outbuf);
if (outbuf_size == 0) {
  ...
} else {
  if (outbuf_size < size)
    ...
}

```

ext/nnstreamer/tensor_decoder/nnstreamer.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
```
gichan-jang, line 1: Tensor_converter needs this schema file.
How about move the file to `ext/nnstreamer/include`.
And generated file also.

ext/nnstreamer/tensor_decoder/nnstreamer.proto
```diff
@@ -0,0 +1,34 @@
...
+    NNS_END = 10;
+  }
+  Tensor_type type = 2;
+  repeated uint32 dimension = 3;
+  bytes data = 4;
+}
+
+message Tensors {
+  uint32 num_tensor = 1;
+  message frame_rate {
```
myungjoo, line 28: Add a comment that frame_rate = rate_n / rate_d and it may be 0.

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
anyj0527: Thanks for the neat suggestion! I updated the commit.

ext/nnstreamer/tensor_decoder/nnstreamer.proto
anyj0527: I think you're right. The suggestion is applied.

ext/nnstreamer/tensor_decoder/nnstreamer.proto
anyj0527: I added it.

meson.build
```diff
@@ -129,6 +129,9 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
 # Protobuf
 protobuf_dep = dependency('protobuf', version: '>= 3.6.1', required: false)
 
+# Protobuf compiler
+pb_comp = find_program('protoc', required: false)
```
zhoonit, line 133:
```suggestion
pb_comp = find_program('protoc', required:  get_option('protobuf-dec-support'))
```

anyj0527: Thanks for pointing out! I missed it.

ext/nnstreamer/include/nnstreamer.proto
```diff
@@ -0,0 +1,35 @@
...
+    NNS_INT16 = 2;
+    NNS_UINT16 = 3;
+    NNS_INT8 = 4;
+    NNS_UINT8 = 5;
+    NNS_FLOAT64 = 6;
+    NNS_FLOAT32 = 7;
+    NNS_INT64 = 8;
+    NNS_UINT64 = 9;
+  
+    NNS_END = 10;
```
myungjoo, line 19: Do we need to define NNS_END in proto? (in C header, END is for loop-break. but I'm not sure if we need this for proto)

We may have additional types defined later and keep it compatible with older types.

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,168 @@
...
+
+/** @brief tensordec-plugin's GstTensorDecoderDef callback */
+static GstFlowReturn
+pb_decode (void **pdata, const GstTensorsConfig * config,
+    const GstTensorMemory * input, GstBuffer * outbuf)
+{
+  GstMapInfo out_info;
+  GstMemory *out_mem;
+  gsize outbuf_size;
+  size_t size;
```
wooksong, line 80: FYI, gsize and size_t are same. IMHO, it is better to consistently use one of them for the readability (this is just a personal opinion).

ext/nnstreamer/tensor_decoder/meson.build
```diff
@@ -117,3 +117,31 @@ static_library('nnstreamer_decoder_image_segment',
   install: true,
   install_dir: nnstreamer_libdir
 )
+
+# protobuf decoder
+if protobuf_dec_support_is_available
+  pb_gen = generator(pb_comp, output: ['@BASENAME@.pb.h', '@BASENAME@.pb.cc'], arguments : ['--proto_path=@CURRENT_SOURCE_DIR@/../include/', '--cpp_out=@BUILD_DIR@', '@INPUT@'])
+  pb_gen_src = pb_gen.process('../include/nnstreamer.proto')
```
wooksong, line 124: How about using meson.current_source_dir()?

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,168 @@
...
+
+  NNStreamer::Tensors tensors;
+  NNStreamer::Tensors::frame_rate *fr;
+
+  const unsigned int num_tensors = config->info.num_tensors;
+  g_assert (num_tensors > 0);
+  tensors.set_num_tensor (num_tensors);
+
+  fr = tensors.mutable_fr ();
```
wooksong, line 89: If fr could be nullptr, null checking is required.

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,168 @@
...
+  
+  if (outbuf_size == 0) {
+    out_mem = gst_allocator_alloc (NULL, size, NULL);
+  } else {
+    if (outbuf_size < size) {
+      gst_buffer_set_size (outbuf, size);
+    }
+    out_mem = gst_buffer_get_all_memory (outbuf);
+  }
+  g_assert (gst_memory_map (out_mem, &out_info, GST_MAP_WRITE));
```
wooksong, line 127: IMO, gst_memory_map should be out of g_assert.

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
anyj0527: Thanks! I added a null check.

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
anyj0527: You're right. I updated!

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
anyj0527: I accord the type of two variables. Please check!

ext/nnstreamer/tensor_decoder/meson.build
anyj0527: `protoc` complains (and do not run) when absolute path and relative path both are given. I think the current `'./include/nnstreamer.proto'` would be more concise.

ext/nnstreamer/include/nnstreamer.proto
anyj0527: I remove the `NNS_END`. The field number of additional types could be assigned with # starting from 10.

ref: https://developers.google.com/protocol-buffers/docs/proto3?#updating

ext/nnstreamer/tensor_decoder/tensordec-protobuf.cc
```diff
@@ -0,0 +1,177 @@
...
+#include <nnstreamer_plugin_api_decoder.h>
+#include <nnstreamer_plugin_api.h>
+#include <nnstreamer_log.h>
+
+#include "nnstreamer.pb.h"    /* Generated by `protoc` */
+
+void init_pb (void) __attribute__ ((constructor));
+void fini_pb (void) __attribute__ ((destructor));
+
+#define DECODER_PB_BIN_CAPS_STR "application/octet-stream"
```
gichan-jang, line 28: You will need to change when you create a converter subplug-in.

anyj0527: Right! I added a caps for protobuf-converter/decoder in tensor_typedef.h after you.

PR #2212, opened 2020-03-26
[Filter/Python/CXX] Do not use 'register' as a storage-class-specifier @open sesame 03/26 13:46

wooksong:
In C++11, the use of the register keyword as a storage-class-specifier is deprecated [1] and this keyword is excluded from storage-class-specifiers in C++17 [2]. In order to avoid build warnings
from the old python2.7 code, this patch adds a workaround for it.

[1] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4193.html#809
[2] http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4340

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_python.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_python.cc
@@ -41,7 +41,16 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #endif
 
+/**
+ * NOTE: This workaround is only for python2.7. If we are going to use external
+ * open-source projects written in cpp with the 'register' keyword, we need to
+ * project-widely apply this.
+ */
+#if __cplusplus > 199711L
+#define register          /* Deprecated in C++11 */
+#endif /* __cplusplus > 199711L */
 #include <Python.h>
+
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_python.cc
```diff
@@ -41,7 +41,16 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #endif
 
+/**
+ * NOTE: This workaround is only for python2.7. If we are going to use external
+ * open-source projects written in cpp with the 'register' keyword, we need to
+ * project-widely apply this.
+ */
+#if __cplusplus > 199711L
```
helloahn: Q: what `199711L` means? As I search, the value of C++11 is `201103L`.
https://gist.github.com/ax3l/53db9fa8a4f4c21ecc5c4100c0d93c94#meaning-of-__cplusplus

wooksong: @helloahn > 199711L which means >= 201103L

helloahn: I got it :)

PR #2216, opened 2020-03-27
[Tests/CAPI] Abort the current test when ml_single_open is failed @open sesame 03/27 15:12

wooksong:
Using the instance of the model handle after failing to ml_single_open causes undefined behaviour. To prevent such cases, this patch changes the line for checking the result of ml_single_open() from EXPECT_EQ to ASSERT_TRUE that aborts the current test function.

**Self evaluation:**

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cc
+++ b/tests/tizen_capi/unittest_tizen_capi.cc
@@ -1754,7 +1754,7 @@ TEST (nnstreamer_capi_singleshot, invoke_01)
 
   status = ml_single_open (&single, test_model, in_info, out_info,
       ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_ANY);
-  EXPECT_EQ (status, ML_ERROR_NONE);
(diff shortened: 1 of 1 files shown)


wooksong:
> To have minimal changes and make the fail-error messages more readable, replace `EXPECT_EQ` with `ASSERT_EQ` rather than `ASSERT_TRUE`.

Makes sense. Replaced ASSERT_TRUE with ASSERT_EQ.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1754,7 +1754,7 @@ TEST (nnstreamer_capi_singleshot, invoke_01)
 
   status = ml_single_open (&single, test_model, in_info, out_info,
       ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_ANY);
-  EXPECT_EQ (status, ML_ERROR_NONE);
+  ASSERT_EQ (status, ML_ERROR_NONE);
```
jijoongmoon: Yes. The test should stop.

PR #2234, opened 2020-04-02
[Tools/dev] Check if *.py is running on shell

zhoonit:
Add `if __name__ == '__main__'` to check if *.py is running on shell to prevent unwanted crash if users happen to import those files to customize tools.

**Self evaluation:**

### DIFF
--- a/tools/development/count_test_cases.py
+++ b/tools/development/count_test_cases.py
@@ -51,12 +51,12 @@ def readSSAT(filename):
         print("No SSAT results.")
     return (0, 0, 0, 0, 0)
 
-def main():
+if __name__ == '__main__':
     if len(sys.argv) != 3:
         print("Usage:")
         print(" $ "+sys.argv[0]+" <gtest xml path> <ssat summary path>")
         print("")
-        return 1
+        sys.exit(1)
 
     tg = 0
     pg = 0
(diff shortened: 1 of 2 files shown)


tools/development/count_test_cases.py
```diff
@@ -51,39 +51,41 @@ def readSSAT(filename):
...
+        print("GTest (total " + str(tg) + " cases)")
+        print("  Passed: " + str(pg) + " / Failed: " + str(fg) + " / Ignored: " + str(ig) + " | Positive: " + str(posg) + " / Negative: " + str(ng))
+        print("SSAT (total " + str(t) + " cases)")
+        print("  Passed: " + str(p) + " / Failed: " + str(f) + " / Ignored: " + str(i) + " | Positive: " + str(t - n) + " / Negative: " + str(n))
+        print("Grand Total: " + str(pg + t) + " cases (negatives : " + str(ng + n) + ")")
+        print("  Passed: " + str(pg+p) + " / Failed: " + str(fg + f) + " / Ignored: " + str(ig + i))
+        return 0
 
-main()
+    main()
```
dongju-chae, line 91: I think you can just remove `main()`

zhoonit: I'll work on that.

zhoonit:
I deleted main() and instead used sys.exit(#) call to exit :)

PR #2240, opened 2020-04-03
[Android/Test] add tc to check classification result

jaeyun-jung:
Using image classification model (tf-lite), check label with max score.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
@@ -177,8 +177,9 @@ public class APITestPipeline {
 
     @Test
     public void testRegisterNullDataCb_n() {
-        String desc = "videotestsrc ! videoconvert ! video/x-raw,format=RGB ! " +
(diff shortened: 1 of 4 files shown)


api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestPipeline.java
```diff
@@ -428,6 +428,61 @@ public void onNewDataReceived(TensorsData data) {
...
+
+                    ByteBuffer buffer = data.getTensorData(0);
+                    int labelIndex = APITestCommon.getMaxScore(buffer);
+
+                    /* check label index (orange) */
+                    if (labelIndex != 951) {
+                        mInvalidState = true;
+                    }
+
+                    mReceived++;
```
gichan-jang, line 465: Where is mReceived initialized?

jaeyun-jung: `public void setUp()` called before starting each test.

jaeyun-jung:
PR updated (2020-04-03)
- tc revised, with coverage report

PR #2244, opened 2020-04-06
[Tests/CAPI] Add state change check methods

gichan-jang:
Modified to check the state change of each state.
Considering low speed of the arm, I gave more time than it needed.

*This should have been included in the previous PR, but it was not reflected because the modification was late.

**Self evaluation:**
1. Build test: [ *]Passed [ ]Failed []Skipped
2. Run test: [ *]Passed [ ]Failed []Skipped

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cc
+++ b/tests/tizen_capi/unittest_tizen_capi.cc
@@ -489,6 +489,31 @@ file_cmp (const gchar * f1, const gchar * f2)
 }
 
 /**
+ * @brief Wait until the change in pipeline status is done
+ * @return ML_ERROR_NONE success, ML_ERROR_UNKNOWN if failed, ML_ERROR_TIMED_OUT if timeout happens.
+ */
+int
+waitPipelineStateChange (ml_pipeline_h handle, ml_pipeline_state_e state,
+    guint timeout_ms)
+{
+  int status = ML_ERROR_UNKNOWN;
+  guint counter = 0;
+  ml_pipeline_state_e cur_state = ML_PIPELINE_STATE_NULL;
+
+  do {
+    status = ml_pipeline_get_state (handle, &cur_state);
+    EXPECT_EQ (status, ML_ERROR_NONE);
+    if (cur_state == ML_PIPELINE_STATE_UNKNOWN)
+      return ML_ERROR_UNKNOWN;
+    if (cur_state == state)
+      return ML_ERROR_NONE;
+    g_usleep (10000);
+  } while ((timeout_ms / 10) >= counter++);
+
+  return ML_ERROR_TIMED_OUT;
+}
+
+/**
  * @brief Test NNStreamer pipeline sink
  */
 TEST (nnstreamer_capi_sink, dummy_01)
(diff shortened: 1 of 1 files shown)


tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -488,6 +488,31 @@ file_cmp (const gchar * f1, const gchar * f2)
...
+    guint timeout_ms)
+{
+  int status = ML_ERROR_UNKNOWN;
+  guint counter = 0;
+  ml_pipeline_state_e cur_state = ML_PIPELINE_STATE_NULL;
+
+  do {
+    status = ml_pipeline_get_state (handle, &cur_state);
+    EXPECT_EQ (status, ML_ERROR_NONE);
+    if (cur_state == ML_PIPELINE_STATE_UNKNOWN)
```
jaeyun-jung, line 506: status != ML_ERROR_NONE || cur_state == ML_PIPELINE_STATE_UNKNOWN

gichan-jang: Thank you for your comment. 
There seems to be no problem with the proposed revision.
However, if 'ml_pipeline_get_state'  fails, the actual return values vary from ML_ERROR_INVALID_PARAMETER, ML_ERROR_STREAMS_PIPE and etc. 
So, ML_ERROR_UNKNOWN is not an exact return value.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -488,6 +488,31 @@ file_cmp (const gchar * f1, const gchar * f2)
...
+      return ML_ERROR_UNKNOWN;
+    if (cur_state == state)
+      return ML_ERROR_NONE;
+    g_usleep (10000);
+  } while ((timeout_ms / 10) > counter++);
+
+  return ML_ERROR_TIMED_OUT;
```
dongju-chae, line 513: How about check the state before returning timed_out? currently, if timeout_ms is less than 10ms, it always returns timed out.

gichan-jang: Understand!
How about change to
```
  } while ((timeout_ms / 10) >= counter++);
```

dongju-chae: sound good :)

gichan-jang: Updated.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -488,6 +488,31 @@ file_cmp (const gchar * f1, const gchar * f2)
   return cmp;
 }
 
+/**
+ * @brief Wait until the change in pipeline status is done
+ * @return ML_ERROR_NONE success, ML_ERROR_UNKNOWN if failed, ML_ERROR_TIMED_OUT if timeout happens.
+ */
+int
```
myungjoo, line 495: add static. no one is using it externally.

gichan-jang: Defining in /test/unittest_util.c looks good to prevent duplicated code for later.
I will change it.

PR #2245, opened 2020-04-07
[tests] Fix pipeline state check methods

gichan-jang:
The direction of the operator determining the timeout was wrong and it was modified.
The timeout was too long in the test, so I reduced it.

**Self evaluation:**

### DIFF
--- a/tests/cpp_methods/unittest_cpp_methods.cc
+++ b/tests/cpp_methods/unittest_cpp_methods.cc
@@ -92,10 +92,11 @@ TEST (cpp_filter_on_demand, pipeline_01)
   }
 
   if (pipeline) {
-    EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, 500000U), 0);
+    EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
 
     g_usleep (100000);
-    EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, 500000U), 0);
+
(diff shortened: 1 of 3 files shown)


tests/unittest_util.c
```diff
@@ -35,7 +35,7 @@ setPipelineStateSync (GstElement * pipeline, GstState state,
     if (cur_state == state)
       return 0;
     g_usleep (10000);
-  } while ((timeout_ms / 20) < counter++);
+  } while ((timeout_ms / 20) > counter++);
```
myungjoo, line 38: Nice Catch!

tests/cpp_methods/unittest_cpp_methods.cc
```diff
@@ -92,10 +92,11 @@ TEST (cpp_filter_on_demand, pipeline_01)
   }
 
   if (pipeline) {
-    EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, 500000U), 0);
+    EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, 500U), 0);
```
myungjoo, line 95: Let's define a macro for all similar statements and control it centrally.
E.g., ```#define UNITTEST_STATECHANGE_TIMTOUT (500U)```

tests/unittest_util.c
anyj0527: 👍👍👍

tests/cpp_methods/unittest_cpp_methods.cc
gichan-jang: The proposed method seems to be better for managing the tests.
Thank you! Updated.

PR #2246, opened 2020-04-07
[Tizen/Sensor] Do not assert for sensor not available

### LINKED ISSUE #2219: [Tizen 6.0 Sensor Framework] Don't assert for unidentified sensors.
If an unidentified sensor type is given, which may be a new sensor type, at Line 558 of tensor_src_tizensensor.c does g_aseert().

In such a case, we need to stop the pipeline construction and emit readable and visible error messages for application writers.

This will happen with sensor type introduction with sensor-framework API updates and we won't follow up every new sensor in the sensor framework.

This should be done before Tizen 6.0 M1 release!

myungjoo:
If a sensor is not available or unknown,
do not assert, but emit error messages and return errors.

Fixes #2219

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -580,7 +580,7 @@ _ts_configure_handle (GstTensorSrcTIZENSENSOR * self)
   /* Based on Tizen Native App (Sensor) Guide */
   /* 1. Check if the sensor supported */
   ret = sensor_is_supported (self->type, &supported);
-  if (ret != 0) {
+  if (ret != SENSOR_ERROR_NONE) {
     ml_loge ("Tizen sensor framework is not working (sensor_is_supported).\n");
     return -ENODEV;
   }
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -82,6 +82,7 @@
 
 #include <tensor_typedef.h>
 #include <nnstreamer_plugin_api.h>
+#include <nnstreamer_log.h>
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -570,15 +571,19 @@ _ts_configure_handle (GstTensorSrcTIZENSENSOR * self)
...
+    return -ENODEV;
+  }
   self->src_spec = val;
 
   /* Based on Tizen Native App (Sensor) Guide */
   /* 1. Check if the sensor supported */
   ret = sensor_is_supported (self->type, &supported);
-  g_assert (ret == 0);
+  if (ret != 0) {
+    ml_loge ("Tizen sensor framework is not working (sensor_is_supported).\n");
```
gichan-jang, line 584: I don't understand why this part is an error. Can you give me a brief explanation?

gichan-jang, line 584: I don't understand why this part is an error. Can you give me a brief explanation?

myungjoo: if sensor is known to be supported, ```ret == 0``` and ```supported == true```.
if sensor is known to be not supported, ```ret == 0``` and ```supported == false```.
if we do NOT know if it is supported or not, ```ret != 0``` and ```supported is undecided```.

myungjoo: https://docs.tizen.org/iot/api/5.0/tizen-iot-headed/group__CAPI__SYSTEM__SENSOR__MODULE.html#ga1b0149bdb95985188dea080dcb9eca8f

Anyway, I'll update 0 --> SENSOR_ERROR_NONE

PR #2248, opened 2020-04-07
[Mediapipe] add meson option & build script for mediapipe

helloahn:
This is a initial version of meson script about mediapipe. It will be used to development mediapipe as an external filter like other filters.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -382,3 +382,47 @@ if get_option('enable-openvino')
     include_directories: include_directories('.')
   )
 endif
+
+if get_option('enable-mediapipe')
+
+  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
+  MEDIAPIPE_HOME = cmd.stdout().strip()
+  message('['+MEDIAPIPE_HOME+']')
+
+  mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']').returncode()
+  mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode()
+
+  if mediapipe_env_exist == 1 and mediapipe_dir_exist == 0
+    message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
+    mediapipe_incdir = include_directories(
+      MEDIAPIPE_HOME,
+      join_paths(MEDIAPIPE_HOME, 'bazel-bin'),
+      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/eigen_archive'),
+      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
+    )
+
+    cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
+
+    mediapipe_internal_dep = cxx.find_library('mediapipe_internal', dirs: nnstreamer_libdir)
+    mediapipe_external_dep = cxx.find_library('mediapipe_external', dirs: nnstreamer_libdir)
+    
+    nnstreamer_filter_mediapipe_deps = [glib_dep, gst_dep, nnstreamer_dep, mediapipe_external_dep, mediapipe_internal_dep]
+
+    filter_sub_mp_sources = ['tensor_filter_mediapipe.c']
+
+    nnstreamer_filter_mediapipe_sources = []
+    foreach s : filter_sub_mp_sources
+      nnstreamer_filter_mediapipe_sources += join_paths(meson.current_source_dir(), s)
+    endforeach
+    
+    shared_library('nnstreamer_filter_mediapipe',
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -382,3 +382,41 @@ if get_option('enable-openvino')
...
   )
 endif
+
+if get_option('enable-mediapipe')
+  cc = meson.get_compiler('cpp')
+
+  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
+  MEDIAPIPE_HOME = cmd.stdout().strip()
+  message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
+
```
jijoongmoon, line 392: It would be great if error handling is here when there is no $MEDIAPIPE_HOME env variable.

tools/development/mediapipe_external_objs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+for obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/external/ -name '*.o' ); do
+    arr+=($obj)
```
jijoongmoon, line 4: Even though, this script is called by meson.build which has ${MEDIAPIPE_HOEM}, but it would be better check the existence of this variable. and also no need to separate bash scripts (mediapipe_external_objs.sh & mediapipe_internal_objs.sh).  Just use bash function with argument. 
```bash
   mediapipe_objs.sh internal
   mediapipe_objs.sh external
```

ext/nnstreamer/tensor_filter/meson.build
helloahn: I've considered the best way to check the validation of the ENV...
Unfortunately, as I know, there is no function to check the value(string) length or check the file existence in meson.
For this reason, IMO the error printing of meson is the best way for now on, and I remain it as is. any idea to make it better?

tools/development/mediapipe_external_objs.sh
helloahn: I agree. thanks for the advice.

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -382,3 +382,41 @@ if get_option('enable-openvino')
     include_directories: include_directories('.')
   )
 endif
+
+if get_option('enable-mediapipe')
+  cc = meson.get_compiler('cpp')
```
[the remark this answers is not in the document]
helloahn: I got it. Thanks!

tools/development/mediapipe_external_objs.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
helloahn: Thanks for the useful information

ext/nnstreamer/tensor_filter/meson.build
myungjoo: [ -d ] returns 0 if MEDIAPIPE_HOME is undefined. Is this behavior ok?

```
mzx@kohaku:/source/AutoDrv/NNStreamer$ [ -d /source ]
mzx@kohaku:/source/AutoDrv/NNStreamer$ echo $?
0
mzx@kohaku:/source/AutoDrv/NNStreamer$ [ -d ]
mzx@kohaku:/source/AutoDrv/NNStreamer$ echo $?
0
mzx@kohaku:/source/AutoDrv/NNStreamer$ [ -d /ax ]
mzx@kohaku:/source/AutoDrv/NNStreamer$ echo $?
1
mzx@kohaku:/source/AutoDrv/NNStreamer$ 
```

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -382,3 +382,43 @@ if get_option('enable-openvino')
...
+  if run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode() == 0
+    message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
+    mediapipe_incdir = include_directories(
+      MEDIAPIPE_HOME,
+      join_paths(MEDIAPIPE_HOME, 'bazel-bin'),
+      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/eigen_archive'),
+      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
+    )
+
+    cmd = run_command('sh', '-c', '../../../tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
```
dongju-chae, line 400: I think you can utilize `meson.source_root()` to access `tools` directory

helloahn: That's what I need! Thanks!!

ext/nnstreamer/tensor_filter/meson.build
helloahn: it's not ok, obviously. I'll make it better.

helloahn: one more checker with `-z` has been added.

PR #2258, opened 2020-04-08
[Common] Lower cyclomatic-complexity

myungjoo:
The CC value of buffer-sync-collectpad was 23.
Lower it down!

ps. What is cyclomatic complexity: https://en.wikipedia.org/wiki/Cyclomatic_complexity

### DIFF
--- a/gst/nnstreamer/tensor_common.c
+++ b/gst/nnstreamer/tensor_common.c
@@ -1297,6 +1297,82 @@ gst_tensor_time_sync_get_current_time (GstCollectPads * collect,
 }
 
 /**
+ * @brief gst_tensor_time_sync_buffer_from_collectpad's internal function.
+ * SYNC helper "SLOWEST"
+ */
+static gboolean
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_common.c
```diff
@@ -1296,6 +1296,82 @@ gst_tensor_time_sync_get_current_time (GstCollectPads * collect,
...
+  }
+  return TRUE;
+}
+
+/**
+ * @brief gst_tensor_time_sync_buffer_from_collectpad's internal function.
+ * SYNC helper "BASEPAD"
+ */
+static gboolean
+gst_tensor_time_sync_buffer_from_collectpad_SYNC_BASEPAD (GstBuffer ** buf,
```
gichan-jang, line 1342: SLOWEST and BASEPAD functions look almost same.
Is it to reduce CC that does not integrate SLOWEST and BASEPAD?

myungjoo: Updated.

Please review the new code carefully if there is no logic errors.

gichan-jang: There seems to be no problem with the logic!

PR #2260, opened 2020-04-08
[Android/NNFW] add NNFW

jaeyun-jung:
Update build script and API to support NNFW.

TODO : fix/update prebuilt libs of NNFW later (now CPU only)


jaeyun-jung:
2020-04-09 PR updated
- fix nnstreamer version info
- add NNStreamer.java in native build

jaeyun-jung:
2020-04-10 PR updated
- add new constructor in SingleShot (new instance with nnfw type)

PR #2265, opened 2020-04-13
Lower MCD, Increase SAM Score

myungjoo:
SAM analysis says that nnstreamer.c is causing cyclic dependency
between modules.

Although I think this is false positive, let's move nnstreamer.c
to a subdirectory so that SAM won't recognize nnstreamer.c is
"parent module" of other nnstreamer elements.

Because this is the only modules that SAM recognizes as MCD issue,
this may enhance SAM score greatly.

### DIFF
--- a/gst/nnstreamer/registerer/nnstreamer.c
+++ b/gst/nnstreamer/registerer/nnstreamer.c
@@ -47,21 +47,21 @@
 
 #include <gst/gst.h>
 
-#include "tensor_aggregator/tensor_aggregator.h"
-#include "tensor_converter/tensor_converter.h"
-#include "tensor_decoder/tensordec.h"
-#include "tensor_demux/gsttensordemux.h"
-#include "tensor_filter/tensor_filter.h"
-#include "tensor_merge/gsttensormerge.h"
-#include "tensor_mux/gsttensormux.h"
-#include "tensor_repo/tensor_reposink.h"
-#include "tensor_repo/tensor_reposrc.h"
-#include "tensor_sink/tensor_sink.h"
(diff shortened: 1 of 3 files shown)


myungjoo:
Do not merge this until I can confirm that this commit increases SAM score regarding MCD.

gst/nnstreamer/registerer/nnstreamer.c
```diff
@@ -47,21 +47,21 @@
...
-#include "tensor_converter/tensor_converter.h"
-#include "tensor_decoder/tensordec.h"
-#include "tensor_demux/gsttensordemux.h"
-#include "tensor_filter/tensor_filter.h"
-#include "tensor_merge/gsttensormerge.h"
-#include "tensor_mux/gsttensormux.h"
-#include "tensor_repo/tensor_reposink.h"
-#include "tensor_repo/tensor_reposrc.h"
-#include "tensor_sink/tensor_sink.h"
+#include "../tensor_aggregator/tensor_aggregator.h"
```
jaeyun-jung, line 50: this is unnecessary

jaeyun-jung:
This will increase sam score :
4.69 (MCD 4.44) -> 4.88 (MCD 5.0)

gst/nnstreamer/registerer/nnstreamer.c
myungjoo: Updated.

myungjoo:
@jaeyun-jung Please test this commit in SAM whether the MCD score improves or not.

jaeyun-jung:
With sam-tool latest version
SAM 4.87 (MCD 5.0)

PR #2266, opened 2020-04-13
[Tensor-If] Make it buildable

myungjoo:
Add empty functions and build infra for Tensor-If.
    
Tensor-If is a subclass of GstBaseTransform.
    
The src-pad and sink-pad are supposed to have the same
capabilities.
    
It appears that in-place op has no benefit, but
not sure if we need to prohibit it.
    
### DIFF
--- a/gst/nnstreamer/tensor_if/gsttensorif.c
+++ b/gst/nnstreamer/tensor_if/gsttensorif.c
@@ -87,6 +87,9 @@ enum
   PROP_ELSE_OPTION, /**< Option for FALSE Action */
 };
 
+GST_DEBUG_CATEGORY_STATIC (gst_tensor_if_debug);
+#define GST_CAT_DEFAULT gst_tensor_if_debug
+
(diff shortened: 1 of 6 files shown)


gst/nnstreamer/tensor_if/gsttensorif.c
```diff
@@ -130,3 +133,243 @@ static gboolean gst_tensor_if_start (GstBaseTransform * trans);
...
+      "NNStreamer/If",
+      "Controls streams based on the tensor(s) values",
+      "MyungJoo Ham <myungjoo.ham@samsung.com>");
+
+  gst_element_class_add_pad_template (gstelement_class,
+      gst_static_pad_template_get (&src_factory));
+  gst_element_class_add_pad_template (gstelement_class,
+      gst_static_pad_template_get (&sink_factory));
+
+  /* Refer: https://gstreamer.freedesktop.org/documentation/design/element-transform.html */
```
zhoonit, line 173: Refer link seems changed to  https://gstreamer.freedesktop.org/documentation/additional/design/element-transform.html

myungjoo: URL updated.

PR #2267, opened 2020-04-14
[Android/Single] Do not turn off a large code block with #if

myungjoo:
Separate the file and build independently instead.

CC: @jaeyun-jung : This requires your review.


api/android/api/src/main/jni/Android.mk
```diff
@@ -43,6 +41,8 @@ ifeq ($(NNSTREAMER_API_OPTION),single)
 NNS_API_FLAGS += -DNNS_SINGLE_ONLY=1
 endif
 
+include $(NNSTREAMER_ROOT)/jni/nnstreamer.mk
```
jaeyun-jung, line 44: plz move line 37 under this line (NNSTREAMER_VERSION is defined in nnstreamer.mk)
NNS_API_FLAGS += -DVERSION=\"$(NNSTREAMER_VERSION)\"
NNSTREAMER_VERSION

jni/nnstreamer.mk
```diff
@@ -25,6 +25,10 @@ NNSTREAMER_COMMON_SRCS := \
     $(NNSTREAMER_GST_HOME)/nnstreamer_subplugin.c \
     $(NNSTREAMER_GST_HOME)/tensor_common.c
 
+ifneq ($(NNSTREAMER_API_OPTION),single)
```
jaeyun-jung, line 28: how about adding 'tensor_common_pipeline.c' in NNSTREAMER_PLUGINS_SRCS?

api/android/api/src/main/jni/Android.mk
myungjoo: Applied. Thanks!

jni/nnstreamer.mk
myungjoo: Moved to NNSTREAMER_PLUGINS_SRCS along with a few comments to avoid confusion.

PR #2268, opened 2020-04-14
[Converter] Refactored: no more large #if/endif code blocks.

myungjoo:
Do not enclose a large code block with #if/#endif.
Separate such files instead.


gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -46,11 +46,32 @@
 
 #include <string.h>
 #include "tensor_converter.h"
-#include "converter-media-info.h"
+#ifndef NO_VIDEO
```
jaeyun-jung, line 49: IMO, this change will break build, if disable-video-support=true or disable-audio-support=true.

myungjoo: Thanks for pointing it out. Refactored headers to support them.

PR #2269, opened 2020-04-14
[Dec/Subplugins] Remove duplicated code

myungjoo:
Add "utility" internal API to remove
duplicated code.

### DIFF
--- a/ext/nnstreamer/tensor_decoder/meson.build
+++ b/ext/nnstreamer/tensor_decoder/meson.build
@@ -23,7 +23,8 @@ static_library('nnstreamer_decoder_direct_video',
 
 # image labeling
 decoder_sub_image_labeling_sources = [
-  'tensordec-imagelabel.c'
+  'tensordec-imagelabel.c',
+  'tensordecutil.c'
 ]
 
 nnstreamer_decoder_image_labeling_sources = []
@@ -47,6 +48,7 @@ static_library('nnstreamer_decoder_image_labeling',
 # bounding boxes
 decoder_sub_bounding_boxes_sources = [
   'tensordec-boundingbox.c',
+  'tensordecutil.c',
   'tensordec-font.c'
 ]
 
@@ -71,6 +73,7 @@ static_library('nnstreamer_decoder_bounding_boxes',
 # pose estimation
 decoder_sub_pose_estimation_sources = [
   'tensordec-pose.c',
+  'tensordecutil.c',
   'tensordec-font.c'
 ]
 
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -56,6 +56,7 @@
 #include <math.h>               /* expf */
 #include <nnstreamer_plugin_api_decoder.h>
 #include <nnstreamer_plugin_api.h>
+#include "tensordecutil.h"
 
 void init_bb (void) __attribute__ ((constructor));
 void fini_bb (void) __attribute__ ((destructor));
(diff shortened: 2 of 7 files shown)


ext/nnstreamer/tensor_decoder/meson.build
```diff
@@ -23,7 +23,8 @@ static_library('nnstreamer_decoder_direct_video',
 
 # image labeling
 decoder_sub_image_labeling_sources = [
-  'tensordec-imagelabel.c'
+  'tensordec-imagelabel.c',
+  'tensordecutil.c'
```
dongju-chae, line 27: Q: It's minor question but I just wonder what the file naming policy is here. `tensordecutil.c` or `tensordec-util.c` ?

myungjoo: If we name it tensordec-util.c, it would look like a "util" subplugin for tensordec.

dongju-chae: Right. Got it.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -230,68 +187,16 @@ bb_exit (void **pdata)
 {
   bounding_boxes *bdata = *pdata;
 
-  _free_labels (bdata);
+  _free_labels (&bdata->labeldata);
 
-  if (bdata->label_path)
-    g_free (bdata->label_path);
+  if (bdata->labeldata.label_path)
+    g_free (bdata->labeldata.label_path);
```
dongju-chae, line 193: How about moving this line inside `_free_labels`?

ext/nnstreamer/tensor_decoder/tensordecutil.h
```diff
@@ -0,0 +1,43 @@
...
+  guint max_word_length; /**< The max size of labels */
+} imglabel_t;
+
+extern void
+loadImageLabels (imglabel_t *l);
+
+extern void
+initSingleLineSprite (singleLineSprite_t v, rasters_t r, uint32_t pv);
+
+extern void _free_labels(imglabel_t *data);
```
helloahn, line 38: style check: `_free_labels(` -> `_free_labels (`

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
myungjoo: No, we can't. Look at tensordecutil.c::35. label_path may be referred after calling it.

ext/nnstreamer/tensor_decoder/tensordecutil.h
myungjoo: Fixed.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
dongju-chae: Ah. I missed that. Got it.

zhoonit: Arguably, `label_path` shouldn't be part of this `imglabel_t`

1. `label_path` is not highly coupled to the label itself, rather it is a way used to inject data. There might be other way to initiate `imglabel_t` later on.
2. `label_path` always has to be treated separately to `imglabel_t` as seen in tensordecutil.c::35 or here.

ext/nnstreamer/tensor_decoder/meson.build
wooksong, line 27: How about making this src file be a common dep for all decoder sub-plugins?

myungjoo: Image-segment and Directvideo don't use it yet.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
myungjoo: @zhoonit A good point. Let's refactor it.

myungjoo: Updated.

PR #2272, opened 2020-04-14
[Profile] internal log

jaeyun-jung:
add internal profile log function.

### DIFF
--- a/gst/nnstreamer/nnstreamer_profile.c
+++ b/gst/nnstreamer/nnstreamer_profile.c
@@ -1,0 +1,89 @@
+/**
+ * Copyright (c) 2020 Samsung Electronics Co., Ltd. All Rights Reserved.
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * @file	nnstreamer_profile.h
+ * @date	14 April 2020
+ * @brief	Internal util for profile log.
+ * @see		http://github.com/nnstreamer/nnstreamer
+ * @author	Jaeyun Jung <jy1210.jung@samsung.com>
+ * @bug		No known bugs except for NYI items
+ */
+
+#include "nnstreamer_log.h"
+#include "nnstreamer_profile.h"
+
+/**
+ * @brief Hashtable to handle the profile tasks.
+ */
+static GHashTable *g_nns_profile_task = NULL;
+
+/**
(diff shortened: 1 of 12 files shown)


gst/nnstreamer/nnstreamer_profile.c
```diff
@@ -0,0 +1,89 @@
...
+nns_profile_end (const gchar * name)
+{
+  nns_profile_task_s *task;
+  gint64 time_diff, average;
+
+  task = g_hash_table_lookup (g_nns_profile_task, name);
+  g_assert (task);
+
+  time_diff = g_get_monotonic_time () - task->start;
+  task->total += time_diff;
```
wooksong, line 82: Even though using averages, I am not sure that time-diff between two user-level code lines could represent the software-stack overhead. For Android, need to investigate https://developer.android.com/topic/performance/tracing/custom-events. For the other linux-based platforms, need to use perf API. 

Anyway as a fallback for those profiling framework time-diff would be one of our options.

gst/nnstreamer/nnstreamer_profile.c
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @brief Function to set start time.
+ */
+void
+nns_profile_start (const gchar * name)
+{
+  nns_profile_task_s *task;
+
+  if (g_nns_profile_task == NULL) {
+    g_nns_profile_task =
```
dongju-chae, line 54: Q: Is it okay not to consider any locking mechanism to access this global hash table?

gst/nnstreamer/nnstreamer_profile.c
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  task = g_hash_table_lookup (g_nns_profile_task, name);
+  if (task == NULL) {
+    task = g_new0 (nns_profile_task_s, 1);
+    g_assert (task);
+    g_assert (g_hash_table_insert (g_nns_profile_task, g_strdup (name), task));
+  }
+
+  task->start = g_get_monotonic_time ();
```
myungjoo, line 66: This appears too heavy for a profile/instrumentation function.

jaeyun-jung:
plz ignore this.

PR #2194, opened 2020-03-21
C++ Tensor-Filter Class / EdgeTPU C++ class / Enable C++ TF class by default

myungjoo:
This updates edgetpu subplugin as a C++ class to test V1 subplugin API and C++ class as T-F subplugin features.

In the course there are
- (patch0001) The EDGE-TPU C++ (derived) Class
- (patch0002) Enable C++ base class by default
- (patch0003) Spec/Debian update (DIST/packaging)
- (patch0004) C++ base class bugfix.
- (patch0005] Workaround for #2320

Tested with edgetpu, C++ tensor-filter class is ready to be deployed.

Included commits:

commit c4d30225d75fe50d31de0fc703fe81d28e58bc44 (HEAD -> cpp/filter/edgetpu)
Author: MyungJoo Ham <myungjoo.ham@samsung.com>
Date:   Mon Apr 27 18:29:02 2020 +0900

    [Filter/EdgeTPU] Workaround for #2320
    
    Until #2320 is resolved, adding AUTO/DEFAULT to "supported hw"
    list makes it easier to use.
    
    Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

commit 53c730c343a3e1fb039e4fc31eead78420e0611b
Author: MyungJoo Ham <myungjoo.ham@samsung.com>
Date:   Wed Apr 22 14:26:58 2020 +0900

    [Filter/C++Class] Bugfix: close gets a dlb ptr!
    
    Pointer type mismatch fixed!
    
    Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

commit 7c7ca1ad5cc432ba7f52e7feccc58c6723db478b
Author: MyungJoo Ham <myungjoo.ham@samsung.com>
Date:   Fri Apr 17 14:22:26 2020 +0900

    [Spec/Deb] Update C++ devel env.
    
    1. Do not separate C env and C++ env. Merge them.
    Both Ubuntu and Tizen supports C++ by default.
    
    2. Move /nnstreamer/cpp/nnstreamer_cppplugin_api_filter.hh to
    /nnstreamer/nnstreamer_cppplugin_api_filter.hh
    
    Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

commit d8ded8d8dd3003f74115f6ad264a9734fcf71fe7
Author: MyungJoo Ham <myungjoo.ham@samsung.com>
Date:   Tue Mar 31 19:24:52 2020 +0900

[...]

PR #2273, opened 2020-04-16
[Filter/v1] Bugfix of TF-V1 subplugin API / C++ support @open sesame 4/21 10:46

myungjoo:
By testing with C++ class tf-subplugin, "edget-tpu",
bugs were found from the recently introduced TF-V1 subplugin API.

- V1 APIs uses void * private_data instead of void ** private_data.
  There are codes that do not reflect this.

In order to make C++ class subplugins work smoothly,
they need to know "this" pointer when they were invoked
by common C++ wrapper infrastructure, which is a set of
static functions of the base class.

Thus, add a new argument for them to find out "this" pointer.

CC: @kparichay

ps. This breaks C++ subplugin classes, which are, fortunately, not built by default yet. C++ subplugin classes will be working after PR #2194 (it has fixes)

### DIFF
--- a/gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh
+++ b/gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh
@@ -36,8 +36,16 @@
 #define __NNS_TENSOR_FILTER_CPP_SUBPLUGIN_SUPPORT_H__
 #ifdef __cplusplus
 
+#include <stdexcept>
 #include <stdint.h>
+#include <assert.h>
 #include <nnstreamer_plugin_api_filter.h>
+#if __cplusplus < 201103L
+#warn C++11 is required for safe execution
+#else
+#include <type_traits>
+#endif
+
 
 namespace nnstreamer {
 /**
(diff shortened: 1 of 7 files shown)


gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh
```diff
@@ -64,22 +72,68 @@ class tensor_filter_subplugin {
...
     template<typename T>
-    static void unregister_subplugin (T * emptyInstance);
-        /**< Unregister this subplugin class (not object) (usually at so-exit) */
+    static void unregister_subplugin (T * emptyInstance) {
+      GstTensorFilterFrameworkInfo info;
+#if __cplusplus < 201103L
+#warn C++11 is required for safe execution
+#else
+      /** The given class T should be derived from tensor_filter_subplugin */
+      assert((std::is_base_of<tensor_filter_subplugin, T>::value));
```
helloahn, line 128: style check: `assert(` -> `assert (`

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -683,10 +681,10 @@ gst_tensor_filter_transform_caps (GstBaseTransform * trans,
         /* call setInputDimension with given input tensor */
         gst_tensors_info_init (&out_info);
         if (GST_TF_FW_V0 (priv->fw)) {
-          gst_tensor_filter_call (priv, res, setInputDimension, &config.info,
-              &out_info);
+          gst_tensor_filter_v0_call (priv, res, setInputDimension,
```
dongju-chae, line 684: This is not directly related to this PR. But I have some concerns on version checking.
If-statements to check version do not have the same control flow.

For example, some functions check both V0 and V1 using `GST_TF_FW_V0` and `GST_TF_FW_V1`.
But, another functions do this using either `GST_TF_FW_V0` or `GST_TF_FW_V1`, replying on `else` for the other version.

I think there is no problem currently. But, in near future, when newer versions are introduced, this makes some problems.

myungjoo: There was the same discussion before with me and PK. At that time, this was kept because in this context v1/v2 check has been already done and assert-failed if it's not either v1 or v2. Anyway, as #2282 has suggested, we should not rely too much on assert (for NDEBUG) and we need to refactor these as well.

gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh
myungjoo: Updated.

gst/nnstreamer/tensor_filter/tensor_filter.c
dongju-chae: Got it

gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
```diff
@@ -102,30 +117,65 @@ void tensor_filter_subplugin::cpp_close (const GstTensorFilterProperties * prop,
...
  * @brief C V1 tensor-filter wrapper callback function, "invoke"
  */
-int tensor_filter_subplugin::cpp_invoke (const GstTensorFilterProperties *prop,
-    void *private_data, const GstTensorMemory *input,
-    GstTensorMemory *output)
+int tensor_filter_subplugin::cpp_invoke (const GstTensorFilterFramework * tf,
+    const GstTensorFilterProperties *prop, void *private_data,
+    const GstTensorMemory *input, GstTensorMemory *output)
 {
   start_up (obj, private_data);
```
anyj0527, line 124: Looks Good ! 🤗
Minor suggestion: Here and other two lines in this file, how about replace `start_up (obj, private_data)` with
```
tensor_filter_subplugin *obj;
_start_up (obj, private_data);
```
and remove the macro `start_up` ?

myungjoo: Applied a static-inline function and removed both marco definitions.

kparichay:
A little late but thanks for the bugfix.

PR #2277, opened 2020-04-17
[Mediapipe] update meson script

helloahn:
add OpenCV dependency and check the architecture of the system.

Since some of the calculators of mediapipe have dependency at OpenCV, the libraries related to OpenCV are required.
Furthermore, the path of OpenCV libs is created according to the system and architecture of the environment.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -384,45 +384,74 @@ if get_option('enable-openvino')
 endif
 
 if get_option('enable-mediapipe')
-
+  if host_machine.system() != 'linux' and host_machine.cpu_family() != 'x86_64'
+    error('Not Supported System & Architecture. Linux x86_64 is required')
+  endif
+  
   cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
   MEDIAPIPE_HOME = cmd.stdout().strip()
-  message('['+MEDIAPIPE_HOME+']')
 
   mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']').returncode()
   mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode()
 
-  if mediapipe_env_exist == 1 and mediapipe_dir_exist == 0
-    message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
-    mediapipe_incdir = include_directories(
-      MEDIAPIPE_HOME,
-      join_paths(MEDIAPIPE_HOME, 'bazel-bin'),
-      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/eigen_archive'),
-      join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
-    )
-
-    cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
-
-    mediapipe_internal_dep = cxx.find_library('mediapipe_internal', dirs: nnstreamer_libdir)
-    mediapipe_external_dep = cxx.find_library('mediapipe_external', dirs: nnstreamer_libdir)
-    
-    nnstreamer_filter_mediapipe_deps = [glib_dep, gst_dep, nnstreamer_dep, mediapipe_external_dep, mediapipe_internal_dep]
-
-    filter_sub_mp_sources = ['tensor_filter_mediapipe.c']
-
-    nnstreamer_filter_mediapipe_sources = []
-    foreach s : filter_sub_mp_sources
-      nnstreamer_filter_mediapipe_sources += join_paths(meson.current_source_dir(), s)
-    endforeach
-    
-    shared_library('nnstreamer_filter_mediapipe',
-      nnstreamer_filter_mediapipe_sources,
-      dependencies: nnstreamer_filter_mediapipe_deps,
-      include_directories: mediapipe_incdir,
-      install: true,
-      install_dir: filter_subplugin_install_dir
-    )
-  else
+  if mediapipe_env_exist != 1 and mediapipe_dir_exist != 0
     error('Cannot find MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
   endif
+
+  message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
+  mediapipe_incdir = include_directories(
+    MEDIAPIPE_HOME,
+    join_paths(MEDIAPIPE_HOME, 'bazel-bin'),
+    join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/eigen_archive'),
+    join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
+  )
+
+  cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
+  if cmd.returncode() != 0
+    error(cmd.stderr().strip())
+  endif
+
+  mediapipe_internal_dep = cxx.find_library('mediapipe_internal', dirs: nnstreamer_libdir)
+  mediapipe_external_dep = cxx.find_library('mediapipe_external', dirs: nnstreamer_libdir)
+
+  opencv_path = 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -384,45 +384,72 @@ if get_option('enable-openvino')
...
-      nnstreamer_filter_mediapipe_sources,
-      dependencies: nnstreamer_filter_mediapipe_deps,
-      include_directories: mediapipe_incdir,
-      install: true,
-      install_dir: filter_subplugin_install_dir
-    )
+  message('host system: ' + host_machine.system())
+  message('host cpu family: ' + host_machine.cpu_family())
+
+  if host_machine.system() == 'linux' and host_machine.cpu_family() == 'x86_64'
```
myungjoo, line 390: For readability, let's make it
```
if NOT LINUX && NOT x64
  error (and exit)
endif

do everything assuming it's Linux/x64
```

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -384,45 +384,72 @@ if get_option('enable-openvino')
...
+    cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
+    MEDIAPIPE_HOME = cmd.stdout().strip()
+
+    mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']').returncode()
+    mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode()
+
+    if mediapipe_env_exist == 1 and mediapipe_dir_exist == 0
```
myungjoo, line 397: Same here. Check if there is an error and exit if it's an error. Don't nest the code blocks of "no-error" case.

tools/development/gen_mediapipe_libs.sh
```diff
@@ -4,20 +4,24 @@
 if [[ -z "${MEDIAPIPE_HOME}" ]]; then
   echo 'ERROR: Define MEDIAPIPE_HOME First!'
 else
-  # for the external objects
-  for ext_obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/external/ -name '*.o' ); do
-    ext_obj_arr+=($ext_obj)
-  done
-  $(gcc ${ext_obj_arr[@]} -shared -o libmediapipe_external.so -fPIC)
-  $(mv libmediapipe_external.so $1)
+  if [ -d "$1" ]; then
```
myungjoo, line 7: Check if ```$1``` is supplied first.

tools/development/gen_mediapipe_libs.sh
```diff
@@ -4,20 +4,24 @@
...
+    # `for the external objects
+    for ext_obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/external/ -name '*.o' ); do
+      ext_obj_arr+=($ext_obj)
+    done
+    $(gcc ${ext_obj_arr[@]} -shared -o libmediapipe_external.so -fPIC)
```
wooksong, line 12: Is it ok with the failure case of this line?

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -384,45 +384,72 @@ if get_option('enable-openvino')
...
+      opencv_imgproc_dep = cc.find_library('opencv_imgproc', dirs: join_paths(MEDIAPIPE_HOME, 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'))
+      opencv_calib3d_dep = cc.find_library('opencv_calib3d', dirs: join_paths(MEDIAPIPE_HOME, 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'))
+      opencv_highgui_dep = cc.find_library('opencv_highgui', dirs: join_paths(MEDIAPIPE_HOME, 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'))
+      opencv_videoio_dep = cc.find_library('opencv_videoio', dirs: join_paths(MEDIAPIPE_HOME, 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'))
+      opencv_features2d_dep = cc.find_library('opencv_features2d', dirs: join_paths(MEDIAPIPE_HOME, 'bazel-bin/_solib_k8/_U@linux_Uopencv_S_S_Copencv___Uexternal_Slinux_Uopencv_Slib_Sx86_U64-linux-gnu'))
```
dongju-chae, line 417: The joined path `bazel-...` is always same for all dependency. How about defining a path variable (for readability)?

tools/development/gen_mediapipe_libs.sh
```diff
@@ -4,20 +4,24 @@
...
-    if [[ $(nm -Ca $int_obj) != *"T main"* ]]; then
-      int_obj_arr+=($int_obj)
-    fi
-  done
-  $(gcc ${int_obj_arr[@]} -shared -o libmediapipe_internal.so -fPIC)
-  $(mv libmediapipe_internal.so $1)
+    # for the internal objects
+    for int_obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/mediapipe/ -name '*.o' ); do
+      # remove object files which contain main functions
+      if [[ $(nm -Ca $int_obj) != *"T main"* ]]; then
```
wooksong, line 18: In order to provide portability, define variables for gcc and nm (e.g., CC/NM). ?= could be helpful.

tools/development/gen_mediapipe_libs.sh
```diff
@@ -4,20 +4,24 @@
...
-  # for the external objects
-  for ext_obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/external/ -name '*.o' ); do
-    ext_obj_arr+=($ext_obj)
-  done
-  $(gcc ${ext_obj_arr[@]} -shared -o libmediapipe_external.so -fPIC)
-  $(mv libmediapipe_external.so $1)
+  if [ -d "$1" ]; then
+    # `for the external objects
+    for ext_obj in $(find ${MEDIAPIPE_HOME}/bazel-bin/external/ -name '*.o' ); do
+      ext_obj_arr+=($ext_obj)
```
wooksong, line 10: why ($ext_obj)?

```suggestion
      ext_obj_arr+=${ext_obj}
```
or

```suggestion
      ext_obj_arr+=$ext_obj
```

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -384,45 +384,72 @@ if get_option('enable-openvino')
...
-    
-    shared_library('nnstreamer_filter_mediapipe',
-      nnstreamer_filter_mediapipe_sources,
-      dependencies: nnstreamer_filter_mediapipe_deps,
-      include_directories: mediapipe_incdir,
-      install: true,
-      install_dir: filter_subplugin_install_dir
-    )
+  message('host system: ' + host_machine.system())
+  message('host cpu family: ' + host_machine.cpu_family())
```
wooksong, line 388: I am not sure that it is necessary to print out this information.

ext/nnstreamer/tensor_filter/meson.build
helloahn: with this case, similar cases have been updated.

tools/development/gen_mediapipe_libs.sh
helloahn: `-z` option is added

tools/development/gen_mediapipe_libs.sh
helloahn: the logic for the failure case has been updated

ext/nnstreamer/tensor_filter/meson.build
helloahn: the variable has been created

tools/development/gen_mediapipe_libs.sh
helloahn: I added like below.
```bash
CC=$(which gcc)
NM=$(which nm)
```
if there is better way, please let me know :)

tools/development/gen_mediapipe_libs.sh
helloahn: since the object files have to be listed like `aa.o bb.o cc.o ...`, the format has been applied. your recommendation generates the list without space between the file name like `aa.obb.occ.o`

ext/nnstreamer/tensor_filter/meson.build
helloahn: it's removed

tools/development/gen_mediapipe_libs.sh
wooksong: s/=/?=/g

PR #2283, opened 2020-04-18
[Bug] Do not call essential ops in assert(). @open sesame 4/18 12:34

### LINKED ISSUE #2282: Do not make important calls inside g_assert()
Hi there, this is a quick recommendation to improve your code quality. First thing that struck me when I inspected this is code is this type of usage:

```
g_assert (gst_buffer_map (srcbuf, &src_info, GST_MAP_READ));
```

It is common on embedded devices to compile out run-time checks. It's actually a GLib feature. If you construct your code the way you do, the call to gst_buffer_map() will be removed, and your code will crash badly. It is also usually recommended to cleanly handle map failure.

myungjoo:
Some systems may use NDEBUG and skip all assert calls.
Do not call essential ops in there.

Fixes #2282

Reported-by: @ndufresne

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -922,6 +922,7 @@ bb_decode (void **pdata, const GstTensorsConfig * config,
   GstMemory *out_mem;
   GArray *results = NULL;
   const int num_tensors = config->info.num_tensors;
+  gboolean status;
 
   g_assert (outbuf);
   /* Ensure we have outbuf properly allocated */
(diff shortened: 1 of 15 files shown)


ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -933,7 +934,8 @@ bb_decode (void **pdata, const GstTensorsConfig * config,
     }
     out_mem = gst_buffer_get_all_memory (outbuf);
   }
-  g_assert (gst_memory_map (out_mem, &out_info, GST_MAP_WRITE));
+  status = gst_memory_map (out_mem, &out_info, GST_MAP_WRITE);
+  g_assert (status);
```
wooksong, line 938: Instead of g_assert(), how about 
```c
if (status == FALSE)
 return GST_FLOW_ERROR;
```

myungjoo: That'd be yet another topic, requiring yet another commit.

PR #2287, opened 2020-04-20
Meson: Change -Denable-options to features object handler @open sesame 05/06 17:31

### LINKED ISSUE #2281: meson: use feature option instead of boolean
You may want to use [meson feature options](https://mesonbuild.com/Build-options.html#features) instead of booleans as we do in GStreamer. This is the proper way to enable/disable features and also allow to automatically enable features based on detected deps.

zhoonit:
This PR is going to resolve #2281 with following implementation plan.

 1. Feature / feature handler added
 2. Add video support as feature (and wipe out disable-video-support)
 3. Implement enable-tesnorflow, enable-tensorflow-lite ... as feature
 4. Add extra args support
 5. Implement enable-orc, ... as feature (current)
 6. Change docs (Left for later PR)

**This PR has an issue:** 

~`-Denable-*` build flags are going to be deprecated. If there is a script using `-Denable-*` flags, those need to be updated. Alternatively, this PR can add backward compatibility but code will be a lot dirtier.~

**Self evaluation:**
1. Build test: [ * ]Passed [ ]Failed [ ]Skipped
2. Run test: [ * ]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -129,245 +129,203 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
 # Protobuf
 protobuf_dep = dependency('protobuf', version: '>= 3.6.1', required: false)
 
-# Orc
-have_orcc = false
+json_glib_dep = dependency('json-glib-1.0', required: get_option('nnfw-runtime-support')) 
 
-if get_option('enable-orc')
-  orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: true)
-  orcc = find_program('orcc', required: true)
-  if orc_dep.found() and orcc.found()
-    have_orcc = true
-    orcc_args = [orcc, '--include', 'glib.h']
-    add_project_arguments('-DHAVE_ORC=1', language: ['c', 'cpp'])
-  else
-    error('Cannot find orc library')
+#orc
+pg_orcc = find_program('orcc', required: get_option('orcc-support'))
+orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: get_option('orcc-support'))
+
+#movidius
+mvncsdk2_dep = dependency('', required: false)
+if not get_option('mvncsdk2-support').disabled() 
+  mvncsdk2_dep = cc.find_library('mvnc', required: false)
+  if not mvncsdk2_dep.found()
+    if cc.check_header('mvnc2/mvnc.h', required: get_option('mvncsdk2-support'))
+      mvncsdk2_dep = declare_dependency()
+    endif
   endif
-else
-  add_project_arguments('-DDISABLE_ORC=1', language: ['c', 'cpp'])
 endif
 
+## nnfw
+nnfw_dep = dependency('', required: false)
+if not get_option('nnfw-runtime-support').disabled()
+  nnfw_dep = dependency('nnfw', required: false)
+  if not nnfw_dep.found()
+    nnfw_dep = cc.find_library('nnfw-dev', required: get_option('nnfw-runtime-support'))
+  endif
+endif
 
 # features registration to be controlled
 #
-# because dict has no ordering guarantees, list is being used instead'
 # register feature as follows
-# [ <string: feature_name>, {
-#      deps: <list: list of dependencies>,
-#      is_available: <boolean: to be used to check availablity,
-#                    defaults to check 'all deps are found' when omitted>
-#      project_args_when_available: <dict: to be converted to #define TOKEN 4 when enable_criteria is true>,
-#      project_args_when_not_available: <dict: to be converted to #define TOKEN 4 when enable_criteria is false>,
-#      message_when_available: <string: info message to be printed when available>,
-#                                  defaults to be "@feature_name@ is available">
-#      message_when_not_available: <string: warning message to be printed when not available
-#                                  defaults to be "@feature_name@ is not available">
-#      extra_args_when_available: <dict: meson variables to be registered when available, 
-#                                        not recommened to use, this exist more for the backward compatibility>
-#      extra_args_when_not_available: <dict: meson variables to be registered when not available, 
-#                                        not recommened to use, this exist more for the backward compatibility>
-# } ]
-
-features = [
-  [
-     'video-support', {
-        'is_available': get_option('video-support').enabled() or get_option('video-support').auto(),
-        'project_args_when_not_available': { 'NO_VIDEO': 1 },
-      }
-  ]
-]
+#  <string: feature_name> :  {
+#      target: <string>, extra_deps: <list>,  project_args: <dict>,
+#      project_args_disabled: <string>, extra_args: <dict: meson variables to be registered>,
+features = {
+  'video-support': {
+    'project_args_disabled': { 'NO_VIDEO': 1 },
+  },
+  'audio-support': {
+    'project_args_disabled': { 'NO_AUDIO': 1 },
+  },
+  'tf-support': {
+    'target': 'tensorflow',
+    'extra_deps': [ protobuf_dep ],
+    'project_args': { 'ENABLE_TENSORFLOW': 1 },
+  },
+  'tflite-support': {
+    'target': 'tensorflow-lite',
+    'project_args': { 'ENABLE_TENSORFLOW_LITE': 1 }
+  },
+  'pytorch-support': {
+    'target': 'pytorch',
+    'project_args': { 'ENABLE_PYTORCH': 1 }
+  },
+  'caffe2-support': {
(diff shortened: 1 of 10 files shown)


meson_options.txt
```diff
@@ -1,3 +1,8 @@
+# features
+option('video-support', type: 'feature', value: 'enabled')
```
gichan-jang, line 2: Wouldn't it be better to change the default value to disable?

zhoonit: Video support is opposite of no_video. I followed the past options, still it could remain for a discussion.

myungjoo: video support should be kept enabled by default.

meson.build
```diff
@@ -146,11 +146,86 @@ else
   add_project_arguments('-DDISABLE_ORC=1', language: ['c', 'cpp'])
 endif
 
-# NO Video support
-if get_option('disable-video-support')
-  add_project_arguments('-DNO_VIDEO=1', language: ['c', 'cpp'])
-  message('Disable Video Type Support')
-endif
+
```
myungjoo, line 149: Isn't this too complicated and lengthy for the given purpose?

zhoonit: 1. boolean has `true` and `false` but feature has 3 options as below which is making feature handling part more complex although having overlapping flow for each feature. 
   
 - `enabled`: same as required: true
 - `auto`: same as `required: false`, but can be controlled by `auto_features=enabled` to enable all features and disable explicitly only the few they don't want, if any.
 - `disabled`: always return `not_found`

2. With this bit, future features and relative variables can declaratively registered instead of handling flow for every feature. 

3. So, inversion of flow control will making feature pluggable and less error-prone as well as saving LoCs. If we migrate 100% of `enable-*` to registerable feature, it would save 132 - 296 from old meson.build(which expected to be doubled up if we migrate to feature by patching those line due to more complex logic).

Checkout similar strategy used in gst
https://github.com/GStreamer/gst-build/blob/master/meson.build (line 57 -  )

zhoonit: I could still go for the old way if requested :)

meson.build
```diff
@@ -129,170 +129,207 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
+      _deps_all_found = false
+    endif
+  endforeach
 
-  if tf_dep.found() and protobuf_dep.found()
-    add_project_arguments('-DENABLE_TENSORFLOW=1', language: ['c', 'cpp'])
-  else
-    error('Cannot find tensorflow')
+  if not _deps_all_found
+    warning('Target dependency is not found for feature @0@'.format(feature_name))
```
myungjoo, line 224: If dependency it not met, it's an error unless user has specified it is "optional". (Let's not handle "optional" dependencies in this automated script for now; we are not using it here, yet.)

meson.build
```diff
@@ -129,170 +129,207 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
+    'deps': [ orc_dep , pg_orcc ],
+    'project_args_when_available': {'HAVE_ORC': 1},
+    'project_args_when_not_available': { 'DISABLE_ORC': 1 },
+    'extra_args_when_available': {'orcc_args': [pg_orcc, '--include', 'glib.h'] }
+  }
+}
+
+project_args = {}
+# This section controls the flow of feature registration.
+foreach feature_name, data : features
```
myungjoo, line 213: Skip the rest (i.e., do "continue;") if ```get_option(feature_name).disabled()```.

meson.build
```diff
@@ -129,170 +129,207 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
+# } ]
+
+features = {
+  'video-support': {
+    'is_available': not get_option('video-support').disabled(),
+    'project_args_when_not_available': { 'NO_VIDEO': 1 },
+  },
+  'audio-support': {
+    'is_available': not get_option('audio-support').disabled(),
+    'project_args_when_not_available': { 'NO_AUDIO': 1 },
```
myungjoo, line 162: Let's rename the string key: ``` s/'project_args_when_not_available'/'project_args'/ ```  We know they are used if the feature is enabled.

meson.build
myungjoo: Or.. for fallback (or alternative), if the alternative dep check exists, do it here and emit error if that fails too. (similar semantics in similar lines)

meson.build
```diff
@@ -129,170 +129,207 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
+  endforeach
+
+  # handle_msg and halt with error when not_available and not_enabled at the sametime by definition
+  _feature_status = _is_available ? 'available' : 'unavailable'
+
+  _msg = data.get('message' + _variable_postfix, '@0@ is @1@.'.format(feature_name, _feature_status))
+  if _is_available
+    message(_msg)
+  else # not available
+    get_option(feature_name).enabled() ? error(_msg) : warning(_msg)
```
myungjoo, line 255: I think you can simplify this code a lot (short lines, simpler logic for the same effects)

meson.build
```diff
@@ -129,170 +129,207 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
-    endforeach
-    if have_python2
-      python2_deps = declare_dependency(dependencies: python2_dep,
-          compile_args : python2_inc_valid_args)
-    else
-      warning('Found python2.7, but failed to find numpy.')
-      warning('Disable nnstreamer-python2.')
-    endif
-  endif
+pg_pkgconfig = find_program('pkg-config')
```
myungjoo, line 271: Do you need to put python-dep refactoring in the same PR?

meson_options.txt
```diff
@@ -1,30 +1,37 @@
...
+option('caffe2-support', type: 'feature', value: 'auto')
+option('python2-support', type: 'feature', value: 'auto')
+option('python3-support', type: 'feature', value: 'auto')
+option('mvncsdk2-support', type: 'feature', value: 'auto')
+# nnfw-runtime ( details in https://review.tizen.org/gerrit/p/platform/core/ml/nnfw )
+# For tf-lite/nnapi, enable-nnfw allows to use nnfw::tflite::nnapi as a backend.
+option('nnfw-runtime-support', type: 'feature', value: 'auto')
+option('tflite-nnapi-delegation', type: 'feature', value: 'auto')
+option('armnn-support', type: 'feature', value: 'auto')
+option('orcc-support', type: 'feature', value: 'auto')
```
myungjoo, line 16: These are good :)

meson.build
zhoonit: python-deps can be put though, I'll save it for another PR (since this pr is too long)

meson.build
zhoonit: Revised

meson.build
zhoonit: Fixed :)

meson.build
zhoonit: Changed :)

meson.build
zhoonit: I changed not to put any fallback logic for now. 
Fallback has many cases making logic more complex.
eg. `deps_1.found() or deps_2.found()` , `dep1_header_found and deps_2.found() or deps_3.found()`, etc...

So, I revised to handle fallback logic separately for more simplicity.

It will fail at `dependency('...', required: get_option(<feature>))` if it has to be failed.

meson.build
```diff
@@ -129,170 +129,200 @@ thread_dep = dependency('threads') # pthread for tensorflow-lite
...
+#orc
+pg_orcc = find_program('orcc', required: get_option('orcc-support'))
+orc_dep = dependency('orc-0.4', version: '>= 0.4.17', required: get_option('orcc-support'))
 
-# NO Audio support
-if get_option('disable-audio-support')
-  add_project_arguments('-DNO_AUDIO=1', language: ['c', 'cpp'])
-  message('Disable Audio Type Support')
+#movidius
+mvncsdk2_dep = cc.find_library('mvnc', required: get_option('mvncsdk2-support'))
```
myungjoo, line 139: https://mesonbuild.com/Reference-manual.html
> required By default, required is set to true and Meson will abort if no program can be found. If required is set to false, Meson continue even if none of the programs can be found. You can then use the .found() method on the returned object to check whether it was found or not. Since 0.47.0 the value of a feature option can also be passed to the required keyword argument.

if mvncsdk2-support=true, meson build fails if we only have corresponding headers. (LINE 141-143 is never executed)

zhoonit: Thanks. Fixed :)

PR #2289, opened 2020-04-20
[Log] Add fatal-log

### LINKED ISSUE #2288: Internal Logging API Request: "ml_logc" (critical)
Request for internal logging API: log level "critical"

- Message writing: critical API or same with ml_loge
- Additional message: (with another input parameter or macro) show stack
    - This requires boost or libunwind for sound backtraces. Let's forget this for now.

We do not have direct needs for this, yet, but for some critical cases, showing stack with critical-log will shorten follow-up procedures especially when the reporter is not a professional programmer.

myungjoo:
Separate "fatal" level from "error".

When we convert "assert" fail to general error handling,
log-fatal might be appropriate.

Fixes #2288

### DIFF
--- a/gst/nnstreamer/nnstreamer_log.h
+++ b/gst/nnstreamer/nnstreamer_log.h
@@ -19,6 +19,8 @@
  * @see		http://github.com/nnstreamer/nnstreamer
  * @author	Jaeyun Jung <jy1210.jung@samsung.com>
  * @bug		No known bugs except for NYI items
+ *
+ * @todo	Provide print-stack internal API as an extension to ml_logf.
  */
 
 #ifndef __NNSTREAMER_LOG_H__
@@ -41,6 +43,9 @@
 #define ml_logd(...) \
     dlog_print (DLOG_DEBUG, TAG_NAME, __VA_ARGS__)
 
+#define ml_logf(...) \
+    dlog_print (DLOG_FATAL, TAG_NAME, __VA_ARGS__)
+
 #elif defined(__ANDROID__)
 #include <android/log.h>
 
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_log.h
```diff
@@ -56,13 +61,17 @@
...
+#define ml_logc(...) \
+    __android_log_print (ANDROID_LOG_FATAL, TAG_NAME, __VA_ARGS__)
+
 #else /* Linux distro */
 #include <glib.h>
 
 #define ml_logi g_info
 #define ml_logw g_warning
-#define ml_loge g_critical
+#define ml_loge g_error
```
jaeyun-jung, line 72: Q: g_error always fatal. is it ok?

myungjoo: It is neither Tizen or Android, so it isn't that critical, but yes, it is ok as long as we intend to use g_error as a standard message for "error" log. (GLIB is the standard in NNStreamer)

myungjoo: Ah.. or is "g_error" more "critical" than "g_critical"?

myungjoo: OK. error is "fatal" and critical is "critical-warning". I'll rename it ml_logf (fatal).

gst/nnstreamer/nnstreamer_log.h
```diff
@@ -56,15 +61,24 @@
...
+    __android_log_print (ANDROID_LOG_FATAL, TAG_NAME, __VA_ARGS__)
+
 #else /* Linux distro */
 #include <glib.h>
 
 #define ml_logi g_info
 #define ml_logw g_warning
 #define ml_loge g_critical
 #define ml_logd g_debug
+#define ml_logf g_error
```
helloahn, line 74: how about changing `g_critical` and `g_error`?
```c
#define ml_loge g_error
#define ml_logf g_critical
`

myungjoo: That confused me as well. 😄 

g_error is a fatal error and g_critical is a critical warning.

helloahn: wow... I've never expected

PR #2290, opened 2020-04-20
[Tests] Fix sink unittest fail

gichan-jang:
Test fails occasionally due to slow speed of arm.
The test was modified to wait until the pipeline processing the buffers

*Until the Daily build PASSES!

Self evaluation:

### DIFF
--- a/tests/nnstreamer_sink/unittest_sink.cc
+++ b/tests/nnstreamer_sink/unittest_sink.cc
@@ -35,6 +35,8 @@
     goto error; \
   }
 
+const guint TEST_TIME_OUT_MSEC = 2000;
+const guint DEFAULT_TIME_INTERVAL =10000;
 const gulong MSEC_PER_USEC = 1000;
 const gulong DEFAULT_JITTER = 0UL;
 const gulong DEFAULT_FPS = 30UL;
@@ -409,6 +411,25 @@ _test_src_eos_timer_cb (gpointer user_data)
   }
 
   return FALSE;
+}
+
+/**
+ * @brief Wait until the pipeline processing the buffers
+ * @return TRUE on success, FALSE when a time-out occurs
+ */
+static gboolean
+_wait_pipeline_process_buffers (guint expected_num_buffers)
+{
+  guint timer_count = 0;
+  /* Waiting for expected buffers to arrive */
+  while (g_test_data.received < expected_num_buffers) {
+    timer_count++;
+    g_usleep (DEFAULT_TIME_INTERVAL);
+    if (timer_count > (TEST_TIME_OUT_MSEC / 10)) {
+      return FALSE;
+    }
+  }
+  return TRUE;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_sink/unittest_sink.cc
```diff
@@ -411,6 +413,25 @@ _test_src_eos_timer_cb (gpointer user_data)
...
+ * @return TRUE on success, FALSE when a time-out occurs
+ */
+static gboolean
+_wait_pipeline_process_buffers (guint expected_num_buffers)
+{
+  gint timer_count = 0;
+  /* Waiting for expected buffers to arrive */
+  while (g_test_data.received < expected_num_buffers) {
+    timer_count++;
+    g_usleep (10000);
```
wooksong, line 427: Where '10000' come from?

tests/nnstreamer_sink/unittest_sink.cc
```diff
@@ -35,6 +35,8 @@
     goto error; \
   }
 
+#define TEST_TIME_OUT_MSEC 2000
```
wooksong, line 38: Instead of #define, how about use const?

tests/nnstreamer_sink/unittest_sink.cc
```diff
@@ -2211,7 +2254,8 @@ TEST (tensor_stream_test, audio_u16_1000f)
 
   gst_element_set_state (g_test_data.pipeline, GST_STATE_PLAYING);
   g_main_loop_run (g_test_data.loop);
-  g_usleep (jitter);
+
+  EXPECT_TRUE (_wait_pipeline_process_buffers (num_buffers / 2));
```
helloahn, line 2258: is there any reason checking only 250 frames?

tests/nnstreamer_sink/unittest_sink.cc
gichan-jang: There is no special reason. I set it because I thought it was appropriate with one interval.

tests/nnstreamer_sink/unittest_sink.cc
gichan-jang: I will update it!

tests/nnstreamer_sink/unittest_sink.cc
wooksong: Then, how about use a const variable for better readability?

tests/nnstreamer_sink/unittest_sink.cc
gichan-jang: The number of frames per buffer of the tensor converter is twice the number of fames of the source.
```
    case TEST_TYPE_AUDIO_U16_1000F:
      /** audio sample rate 16000 (16 bits, unsigned, little endian), 1000 frames */
      str_pipeline =
          g_strdup_printf
          ("audiotestsrc num-buffers=%d samplesperbuffer=500 ! audioconvert ! audio/x-raw,format=U16LE,rate=16000 ! "
          "tensor_converter frames-per-tensor=1000 ! tensor_sink name=test_sink",
          option.num_buffers);
```

tests/nnstreamer_sink/unittest_sink.cc
gichan-jang: Thank you for suggestion. Updated!

PR #2296, opened 2020-04-21
[Src-IIO] CHANNEL property handling fix

myungjoo:
1. If g_hash_table_insert returns FALSE,
it simply means that the entry is duplicated.
Just skip it, then.

2. If we cannot parse it, don't change status, but
ignore the whole property values.

### DIFF
--- a/gst/nnstreamer/tensor_source/tensor_src_iio.c
+++ b/gst/nnstreamer/tensor_source/tensor_src_iio.c
@@ -1211,7 +1211,7 @@ gst_tensor_src_iio_set_property (GObject * object, guint prop_id,
         gint64 val;
         gchar **strv;
         gchar *endptr = NULL;
-        gboolean status;
+        gboolean status = TRUE;
 
         /**
          * using direct as we only need to store keys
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_source/tensor_src_iio.c
```diff
@@ -1225,17 +1225,24 @@ gst_tensor_src_iio_set_property (GObject * object, guint prop_id,
...
             break;
           }
-          status = g_hash_table_insert (self->custom_channel_table,
-              GINT_TO_POINTER (val), NULL);
-          g_assert (status);
+          if (FALSE == g_hash_table_insert (self->custom_channel_table,
+                  GINT_TO_POINTER (val), NULL)) {
+            /** this means val is duplicated. just skip it, then. */
+            ml_logw
+                ("tensor-src-iio's CHANNELS property balue has duplicated entry, '%s', which is ignored.\n",
```
zhoonit, line 1240: g_hash_table_insert replaces it's previous value in case there is a duplicated entry. So it is not ignored.

https://developer.gnome.org/glib/stable/glib-Hash-Tables.html

zhoonit: I think there is a typo: `balue` -> `value`

myungjoo: - typo fixed.
- It's an integer value. Thus, I rephrased: ignored --> registered only once.

zhoonit: Thanks for the explanation :)

PR #2297, opened 2020-04-21
[Coverage/Fix] Disable optimization and remove redundant gcov run

dongju-chae:
This patch disables code optimization for coverage analysis and remove
redundant gcov run.

In fact, we don't need to run the python script, `unittestcoverage.py`
which recursively executes gcov. Because lcov internally executes gcov,
the results of this script are never used.

FYI...
- When program is compiled with coverage option, `.gcno` files are created.
- When such programs are executed, `.gcda` files are created.
- As lcov's input files are `.gcda`, we don't need to run gcov before lcov.

**Self evaluation:**


packaging/nnstreamer.spec
```diff
@@ -484,7 +487,6 @@ popd
 # 'lcov' generates the date format with UTC time zone by default. Let's replace UTC with KST.
 # If you ccan get a root privilege, run ln -sf /usr/share/zoneinfo/Asia/Seoul /etc/localtime
 TZ='Asia/Seoul'; export TZ
-$(pwd)/tests/unittestcoverage.py module $(pwd)/gst $(pwd)/ext %testtarget
```
jaeyun-jung, line 487: if unittestcoverage.py is unnecessary, how about removing this file?

dongju-chae: Yes but I am not sure it's okay for me to remove this...

myungjoo: You may remove it if it is not used by .spec file.

dongju-chae: Got it. I will update the PR

PR #2302, opened 2020-04-21
[Test] Fix nnstreamer repo test

gichan-jang:
EOS signal is generated before the files are saved and the pipeline is stopped.
Therefore, it seems that enough time is needed to save the files.
Reducing frame rate is not best way to fix the problem, But I can not find any other suitable method in SSAT or element options.
When the test was performed at 10Hz, the repeated test did not fail yet.
There is no reasonable reason to set the frame rate, but it was conservatively set to 3Hz.

Self evaluation:


gichan-jang:
A new idea came up from the shuttle bus going to home..
```
... multifilesink locantion=location=testsequence01_%1d.log next-file=4 max-file-size=700 
```
This way, I think it can mimic un-buffered mode.
I will update it after testing whether it works or not.

tests/nnstreamer_repo/runTest.sh
```diff
@@ -32,16 +32,16 @@ fi
 convertBMP2PNG
 
 # Fail Test : Negotiation Error (dimension)
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} multifilesrc location=testsequence_%1d.png index=0 caps=\"image/png,framerate=(fraction)30/1\" ! pngdec ! tensor_converter ! queue ! tensor_reposink silent=false slot-index=0 tensor_reposrc silent=false slot-index=0 caps=\"other/tensor,dimension=(string)3:100:100:1,type=(string)uint8,framerate=(fraction)30/1\" ! multifilesink location=testsequence01_%1d.log" 0F_n 0 1 $PERFORMANCE
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} multifilesrc location=testsequence_%1d.png index=0 caps=\"image/png,framerate=(fraction)3/1\" ! pngdec ! tensor_converter ! queue ! tensor_reposink silent=false slot-index=0 tensor_reposrc silent=false slot-index=0 caps=\"other/tensor,dimension=(string)3:100:100:1,type=(string)uint8,framerate=(fraction)3/1\" ! multifilesink location=testsequence01_%1d.log" 0F_n 0 1 $PERFORMANCE
```
myungjoo, line 35: What if we simple do NOT specify framerate?

myungjoo: Actually, the best way is to set ```framerate=(fraction)0/1```. However, we have some incompatible elements :( (don't know if it's because of tensor-repo or not)

gichan-jang: If the frame rate set to 0, the pipeline does not work..
And the frame rate cannot be omitted in multifilesrc
```
#GstMultiFileSrc:caps property or a capsfilter to force to caps containing a framerate. Otherwise image decoders send EOS after the first picture.
```
It also failed when the frame rate was ignored only in tensor_reposrc.

PR #2307, opened 2020-04-22
[Filter/Python] Don't assert. Provide exceptions.

myungjoo:
A subplugin shouldn't be arrogant enough to assert.
Return errors or throw exceptions instead.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_python.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_python.cc
@@ -50,6 +50,7 @@
 #define register          /* Deprecated in C++11 */
 #endif /* __cplusplus > 199711L */
 #include <Python.h>
+#include <exception>
 
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
 #include <numpy/arrayobject.h>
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_python.cc
```diff
@@ -630,7 +643,11 @@ PYCore::run (const GstTensorMemory * input, GstTensorMemory * output)
 
   PyObject *result = PyObject_CallMethod(core_obj, (char*) "invoke", (char*) "(O)", param);
   if (result) {
-    g_assert((unsigned int) PyList_Size(result) == outputTensorMeta.num_tensors);
+    if ((unsigned int) PyList_Size(result) != outputTensorMeta.num_tensors) {
+      res = -EINVAL;
+      ml_logf ("The Python allocated size mismatched. Cannot proceed.\n");
+      goto exit_decref;
```
dongju-chae, line 649: Before going to `exit_decref`, you need to call `Py_XDECREF (result)`

myungjoo: Got it.

PR #2315, opened 2020-04-24
[Test/NNFW] separate testcase

jaeyun-jung:
1. separate testcase - invalid model and invalid data.
2. fix mem leak, destroy handle before creating new data.

### DIFF
--- a/tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cc
+++ b/tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cc
@@ -748,7 +748,7 @@ TEST (nnstreamer_nnfw_mlapi, invoke_pipeline_00)
   size_t data_size;
 
   gchar *test_model;
-  guint *sink_called_cnt  = (guint *) g_malloc (sizeof (guint));
+  guint *sink_called_cnt = (guint *) g_malloc0 (sizeof (guint));
 
   test_model = get_model_file ();
   ASSERT_TRUE (test_model != nullptr);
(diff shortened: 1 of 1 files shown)


tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cc
```diff
@@ -748,7 +748,7 @@ TEST (nnstreamer_nnfw_mlapi, invoke_pipeline_00)
   size_t data_size;
 
   gchar *test_model;
-  guint *sink_called_cnt  = (guint *) g_malloc (sizeof (guint));
+  guint *sink_called_cnt = (guint *) g_malloc0 (sizeof (guint));
```
dongju-chae, line 751: Is there any reason to use a pointer of `guint`? What happens if just using `guint sink_called_cnt = 0;`

jaeyun-jung: to avoid reference of local variable in callback.

wooksong, line 751: The type casting might be unnecessary here.

> A pointer to void may be converted to or from a pointer to any object type. A pointer to any object type may be converted to a pointer to void and back again; the result shall compare equal to the original pointer.

jaeyun-jung: But, this makes build failure:
```
error: invalid conversion from 'gpointer' {aka 'void*'} to 'guint*' {aka 'unsigned int*'} [-fpermissive]
```

PR #2318, opened 2020-04-27
[Merge] fix return type

jaeyun-jung:
1. fix invalid return value
2. mem-unmap when failed to get mem ptr in input buffer.

### DIFF
--- a/gst/nnstreamer/tensor_merge/gsttensormerge.c
+++ b/gst/nnstreamer/tensor_merge/gsttensormerge.c
@@ -474,6 +474,7 @@ gst_tensor_merge_generate_mem (GstTensorMerge * tensor_merge,
   uint8_t *inptr, *outptr;
   int num_mem = tensor_merge->tensors_config.info.num_tensors;
   int i, j, k, l;
+  size_t c, s;
   gsize outSize = 0;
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_merge/gsttensormerge.c
```diff
@@ -591,23 +581,22 @@ gst_tensor_merge_generate_mem (GstTensorMerge * tensor_merge,
...
-      ret = FALSE;
+      ret = GST_FLOW_ERROR;
   }
 
+  gst_memory_unmap (outMem, &outInfo);
   gst_buffer_append_memory (tensor_buf, outMem);
   gst_buffer_copy_into (tensor_buf, tensors_buf, GST_BUFFER_COPY_TIMESTAMPS, 0,
       -1);
-  gst_memory_unmap (outMem, &outInfo);
+
```
zhoonit, line 599: Since below is error hadler. How about add `return ret` (or change label name below) here as well?

jaeyun-jung: input-mem should be unmapped, so below lines are necessary here :)

PR #2319, opened 2020-04-27
[CodeClean] fix invalid index

jaeyun-jung:
1. fix invalid index when unmapping the mem block
2. unref buffer when failed to get mem


ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -268,6 +269,7 @@ _tflite_ssd_loadBoxPrior (bounding_boxes * bdata)
     prev_reg = registered;
   }
 
+error:
```
zhoonit, line 272: It seems return with success could come to this line.
I think it'd better change label to `done` or equivalent.

PR #2322, opened 2020-04-27
[Sink] default sync false

jaeyun-jung:
Set default sync option 'false'.
It may be delayed with tensor_filter element, to invoke neural network model.

### DIFF
--- a/gst/nnstreamer/tensor_sink/tensor_sink.c
+++ b/gst/nnstreamer/tensor_sink/tensor_sink.c
@@ -98,6 +98,13 @@ enum
 #define DEFAULT_QOS TRUE
 
 /**
+ * @brief Flag to synchronize on the clock (Default FALSE).
+ * It may be delayed with tensor_filter element, to invoke neural network model.
+ * See GstBaseSink::sync property for more details.
+ */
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_sink/tensor_sink.c
```diff
@@ -282,6 +282,12 @@ gst_tensor_sink_init (GstTensorSink * self)
 
   /** enable qos */
   gst_base_sink_set_qos_enabled (bsink, DEFAULT_QOS);
+
+  /**
+   * Default FALSE (synchronize on the clock)
+   * It may be delayed with tensor_filter element, to invoke neural network model.
+   */
+  gst_base_sink_set_sync (bsink, FALSE);
```
gichan-jang, line 290: How about declaring `DEFAULT_SYNC` to maintain the consistency with other codes?

jaeyun-jung: updated. please see the changes.

gichan-jang: LGTM
When I tested with this PR, the problem was resolved clearly!

PR #2323, opened 2020-04-27
[C-Api/Pipeline] warning message for sync option

jaeyun-jung:
add warning message if sink element does not have sync=false option.

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -453,6 +453,18 @@ iterate_element (ml_pipeline * pipe_h, GstElement * pipeline)
                 /** @todo CRITICAL HANDLE THIS! */
               }
 
+              /* check 'sync' property in sink element */
+              if (element_type == ML_PIPELINE_ELEMENT_SINK ||
+                  element_type == ML_PIPELINE_ELEMENT_APP_SINK) {
+                gboolean sync = FALSE;
+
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -453,6 +453,18 @@ iterate_element (ml_pipeline * pipe_h, GstElement * pipeline)
...
 
+              /* check 'sync' property in sink element */
+              if (element_type == ML_PIPELINE_ELEMENT_SINK ||
+                  element_type == ML_PIPELINE_ELEMENT_APP_SINK) {
+                gboolean sync = FALSE;
+
+                g_object_get (G_OBJECT (elem), "sync", &sync, NULL);
+                if (sync) {
+                  ml_logw
+                      ("Recommended to add 'sync=false' option in the sink element, so that incoming buffer would be passed to the application as fast as possible.");
```
gichan-jang, line 464: I think the more accurate expression is `as soon as the byffer arrives` than the `As fast as possible`

myungjoo: Rephrase:
```
It is recommended to apply 'sync=false' property to a sink element in most AI applications. Otherwise, inference results of large neural networks will be frequently dropped by the synchronization mechanism at the sink element.
```

jaeyun-jung: updated

PR #2325, opened 2020-04-28
[CodeClean] line comment

jaeyun-jung:
remove single line comment.


ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: LGPL-2.1-only
+/* SPDX-License-Identifier: LGPL-2.1-only */
```
myungjoo, line 1: In .cc files, it's ok to have //.

PR #2329, opened 2020-04-28
[Common] Allow to fetch nnstreamer version @open sesame 7/1/ 23:04

### LINKED ISSUE #2294: Allow to get version, both static and dynamic
Provide build-time get-version capability (e.g., "NNSTREAMER_VERSION_MAJOR" macro)
and run-time get-version capability (e.g., get_nnstreamer_version() function)

Consider to add this with Tizen API as well (not sure if it's required for such layers... if required, it'd be 6.0 M2)

myungjoo:
Let external nnstreamer plugins to probe nnstreamer version
at both run-time and build-time.

Fixes #2294

### DIFF
--- a/jni/nnstreamer.mk
+++ b/jni/nnstreamer.mk
@@ -9,10 +9,21 @@ $(error GSTREAMER_ROOT_ANDROID is not defined!)
 endif
 
 NNSTREAMER_VERSION  := 1.5.3
+NNSTREAMER_VERSION_MAJOR := $(word 1,$(subst ., ,${NNSTREAMER_VERSION}))
+NNSTREAMER_VERSION_MINOR := $(word 2,$(subst ., ,${NNSTREAMER_VERSION}))
+NNSTREAMER_VERSION_MICRO := $(word 3,$(subst ., ,${NNSTREAMER_VERSION}))
 
 NNSTREAMER_GST_HOME := $(NNSTREAMER_ROOT)/gst/nnstreamer
 NNSTREAMER_EXT_HOME := $(NNSTREAMER_ROOT)/ext/nnstreamer
 NNSTREAMER_CAPI_HOME := $(NNSTREAMER_ROOT)/api/capi
+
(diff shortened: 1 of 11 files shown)


jni/nnstreamer.mk
```diff
@@ -9,10 +9,16 @@ $(error GSTREAMER_ROOT_ANDROID is not defined!)
...
 
 NNSTREAMER_VERSION  := 1.5.2
+NNSTREAMER_VERSION_MAJOR := $(word 1,$(subst ., ,${NNSTREAMER_VERSION}))
+NNSTREAMER_VERSION_MINOR := $(word 2,$(subst ., ,${NNSTREAMER_VERSION}))
+NNSTREAMER_VERSION_MICRO := $(word 3,$(subst ., ,${NNSTREAMER_VERSION}))
 
 NNSTREAMER_GST_HOME := $(NNSTREAMER_ROOT)/gst/nnstreamer
 NNSTREAMER_EXT_HOME := $(NNSTREAMER_ROOT)/ext/nnstreamer
 NNSTREAMER_CAPI_HOME := $(NNSTREAMER_ROOT)/api/capi
+sed -i "s/@__NNSTREAMER_VERSION_MAJOR__@/${NNSTREAMER_VERSION_MAJOR}/" ${NNSTREAMER_GST_HOME}/include/nnstreamer_plugin_api.h
```
jaeyun-jung, line 19: this block should be in build-android-lib.sh

myungjoo: That bash script won't be aware of the version numbers... :(
We need to execute sed in Android-Makefile somehow. (which this script is failing :( )

myungjoo: Or.. maybe I'll do grep on nnstreamer.mk from build-android-lib.sh :)

myungjoo: > this block should be in build-android-lib.sh

No. In CI, it build nnstreamer.mk without build-android-libs.sh and it's safer to assume that a user may build directly with .mk files. This will support both building with the shell script and building directly with .mk files

PR #2330, opened 2020-04-28
[Test] add tc for element properties

jaeyun-jung:
update simple testcases for the element properties.

### DIFF
--- a/tests/nnstreamer_sink/unittest_sink.cc
+++ b/tests/nnstreamer_sink/unittest_sink.cc
@@ -937,7 +937,7 @@ _setup_pipeline (TestOption & option)
           custom_dir ? custom_dir : "./tests", NNSTREAMER_SO_FILE_EXTENSION, option.tmpfile);
       break;
     case TEST_TYPE_ISSUE739_MERGE_PARALLEL_1:
-      /** 4x4 tensor stream, different FPS, tensor_mux them @ slowest */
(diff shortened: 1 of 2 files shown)


tests/nnstreamer_sink/unittest_sink.cc
```diff
@@ -3229,6 +3229,26 @@ TEST (tensor_stream_test, filter_properties_1)
...
 
+  /* sub-plugins */
+  g_object_get (filter, "sub-plugins", &str, NULL);
+  /* custom / custom-easy are always available */
+  EXPECT_TRUE (strstr (str, "custom") != NULL);
+  EXPECT_TRUE (strstr (str, "custom-easy") != NULL);
+#ifdef ENABLE_TENSORFLOW_LITE
+  EXPECT_TRUE (strstr (str, "tensorflow-lite") != NULL);
+#endif
+  g_free (str);
```
gichan-jang, line 3240: Q. Can NULL be returned in the 3233 line?

jaeyun-jung: Currently no, but I will add this case before checking the sub-plugins.
ASSERT_TRUE (str != NULL);

wooksong: g_free does not require NULL checking.

PR #2337, opened 2020-04-29
[Unittest/IIO] add negative cases - @open sesame 05/06 15:26

helloahn:
testcase for the logic with:
1. invalid frequency
2. invalid base dir path

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_source/unittest_src_iio.cc
+++ b/tests/nnstreamer_source/unittest_src_iio.cc
@@ -1662,6 +1662,53 @@ TEST (test_tensor_src_iio, unusual_cases)
 }
 
 /**
+ * @brief test the logic with invalid frequency value
+ */
+TEST (test_tensor_src_iio, set_frequency_n)
+{
+  iio_dev_dir_struct *dev0;
+  GstElement *src_iio_pipeline;
+  GstElement *src_iio;
+  GstStateChangeReturn status;
+  gchar *parse_launch;
+  gint data_value;
+  guint data_bits;
+  gint ret_freq;
+
+  data_value = DATA;
+  data_bits = 16;
+  /** Make device */
+  dev0 = make_full_device (data_value, data_bits);
+  ASSERT_NE (dev0, nullptr);
+  /** setup */
+  dev0->log_file = g_build_filename (dev0->base_dir, "temp.log", NULL);
+  parse_launch =
+      g_strdup_printf
+      ("%s iio-base-dir=%s device-number=%d name=my-src-iio ! fakesink",
+      ELEMENT_NAME, dev0->iio_base_dir_sim, 0);
+  src_iio_pipeline = gst_parse_launch (parse_launch, NULL);
+  g_free (parse_launch);
+  ASSERT_NE (src_iio_pipeline, nullptr);
+
+  /** get and verify the caps */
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_source/unittest_src_iio.cc
```diff
@@ -1564,6 +1564,85 @@ TEST (test_tensor_src_iio, data_verify_freq_generic_type)
...
+  dev0->log_file = g_build_filename (dev0->base_dir, "temp.log", NULL);
+  parse_launch =
+      g_strdup_printf
+      ("%s iio-base-dir=%s device-number=%d name=my-src-iio ! fakesink",
+      ELEMENT_NAME, dev0->iio_base_dir_sim, 0);
+  src_iio_pipeline = gst_parse_launch (parse_launch, NULL);
+  g_free (parse_launch);
+
+  /** get and verify the caps */
+  src_iio = gst_bin_get_by_name (GST_BIN (src_iio_pipeline), "my-src-iio");
```
wooksong, line 1596: parse_launch might be null when gst_parse_lunch fails. https://gstreamer.freedesktop.org/documentation/gstreamer/gstparse.html?gi-language=c#gst_parse_launch


Note that "you might get a return value that is not NULL even though the error is set. In this case there was a recoverable parsing error and you can try to play the pipeline."

helloahn: `nullptr` checker has been added before calling `g_free`

tests/nnstreamer_source/unittest_src_iio.cc
```diff
@@ -1564,6 +1564,86 @@ TEST (test_tensor_src_iio, data_verify_freq_generic_type)
...
+  /** setup properties */
+  g_object_set (src_iio, "iio-base-dir", "/not/existed/path", NULL);
+  g_object_set (src_iio, "dev-dir", dev0->dev_dir, NULL);
+  g_object_set (src_iio, "device", DEVICE_NAME, NULL);
+
+  status = gst_element_set_state (src_iio, GST_STATE_PAUSED);
+  EXPECT_NE (status, GST_STATE_CHANGE_NO_PREROLL);
+
+  /** teardown */
+  gst_harness_teardown (hrnss);
```
gichan-jang, line 1644: Need `gst_object_unref (src_iio);`?

myungjoo: gst_harness_teardown appears to do it.
We may check if src_iio is available or not after LINE1644 in the code to make sure that it's freed. (don't just add unref there unless you sure that it's not freed at that point)

tests/nnstreamer_source/unittest_src_iio.cc
myungjoo: You need to check ```src_iio_pipeline```, not ```parse_launch``` at LINE 1593.

helloahn: @myungjoo it's fixed

tests/nnstreamer_source/unittest_src_iio.cc
helloahn: I executed Valgrind to check mem leak. Unfortunately, there are some points which should be fixed. I'll make another PR that fixes the problem include this one.

PR #2338, opened 2020-04-29
[Tests/Python] Update test scripts for python extensions

wooksong:
This patch modifies the scripts running the test cases for python filters to use proper PYTHONPATH.

**Self evaluation:**

### DIFF
--- a/packaging/run_unittests_binaries.sh
+++ b/packaging/run_unittests_binaries.sh
@@ -13,24 +13,29 @@ export GST_PLUGIN_PATH=$(pwd)/gst/nnstreamer
 export NNSTREAMER_CONF=$(pwd)/nnstreamer-test.ini
 export NNSTREAMER_FILTERS=$(pwd)/ext/nnstreamer/tensor_filter
 export NNSTREAMER_DECODERS=$(pwd)/ext/nnstreamer/tensor_decoder
-export PYTHONPATH=$(pwd)/ext/nnstreamer/tensor_filter/:$PYTHONPATH
-
+export _PYTHONPATH=${PYTHONPATH}
 
 run_entry() {
   entry=$1
-
-  if [[ $entry == *"python3"* ]]; then
+  if [[ $entry == *"python3"* || $entry == *"python2"* ]]; then
+    PY=$(echo ${entry} | grep -oP "python[0-9]")
     pushd ext/nnstreamer/tensor_filter
-    ln -sf nnstreamer_python3.so nnstreamer_python.so
+    TEST_PYTHONPATH=${PY}_module
+    rm -rf ${TEST_PYTHONPATH}
+    mkdir -p ${TEST_PYTHONPATH}
+    pushd ${TEST_PYTHONPATH}
+    # Covert to an absolute path from the relative path
+    TEST_PYTHONPATH=$(pwd)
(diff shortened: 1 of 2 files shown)


packaging/run_unittests_binaries.sh
```diff
@@ -13,24 +13,29 @@ export GST_PLUGIN_PATH=$(pwd)/gst/nnstreamer
...
+    PY=$(echo ${entry} | grep -oP "python[0-9]")
     pushd ext/nnstreamer/tensor_filter
-    ln -sf nnstreamer_python3.so nnstreamer_python.so
+    TEST_PYTHONPATH=${PY}_module
+    rm -rf ${TEST_PYTHONPATH}
+    mkdir -p ${TEST_PYTHONPATH}
+    pushd ${TEST_PYTHONPATH}
+    # Covert to an absolute path from the relative path
+    TEST_PYTHONPATH=$(pwd)
+    export PYTHONPATH=${TEST_PYTHONPATH}
```
zhoonit, line 29: Should this be `PYTHONPATH=${TEST_PYTHONPATH}:${_PYTHONPATH}` ?

`meson.build` declares requirement for `numpy` when using python. This might override it.

tests/nnstreamer_filter_python/runTest.sh
```diff
@@ -44,11 +44,18 @@ fi
 FRAMEWORK="python2"
 # This symlink is necessary only for testcases; when installed, symlinks will be made
 pushd ../../build/ext/nnstreamer/tensor_filter
-ln -s nnstreamer_python2.so nnstreamer_python.so
+TEST_PYTHONPATH=${FRAMEWORK}_pymodule
+mkdir -p ${TEST_PYTHONPATH}
+pushd ${TEST_PYTHONPATH}
+# Covert to an absolute path from the relative path
+export PYTHONPATH=$(pwd)
```
zhoonit, line 51: ditto.

tests/nnstreamer_filter_python/runTest.sh
```diff
@@ -44,11 +44,18 @@ fi
 FRAMEWORK="python2"
 # This symlink is necessary only for testcases; when installed, symlinks will be made
 pushd ../../build/ext/nnstreamer/tensor_filter
-ln -s nnstreamer_python2.so nnstreamer_python.so
+TEST_PYTHONPATH=${FRAMEWORK}_pymodule
```
zhoonit, line 47: Here is `python2_pymodule` while similar path in `packaging/run_unittests_binaries.sh:23` is `python[23]_module` . IMO this should be the same if there isn't any reason.

packaging/run_unittests_binaries.sh
wooksong: @zhoonit nope. PYTHONPAYH is an extra path. Numpy will be imported from the system path.

tests/nnstreamer_filter_python/runTest.sh
wooksong: It would be good to use same path but not should be. These two scripts are not invoked in the same context. Note that runTest.sh is invoked by SSAT.

tests/nnstreamer_filter_python/runTest.sh
wooksong: ditto.

PR #2339, opened 2020-04-29
[Transform] return value when failed to set option @open sesame 05/06 15:26

jaeyun-jung:
1. return false if failed to set option.
2. add simple testcases for the option string.

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -730,7 +730,7 @@ static gboolean
 gst_tensor_transform_set_option_data (GstTensorTransform * filter)
 {
   gchar *filter_name;
-  gboolean ret = TRUE;
+  gboolean ret = FALSE;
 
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -730,7 +730,7 @@ static gboolean
 gst_tensor_transform_set_option_data (GstTensorTransform * filter)
 {
   gchar *filter_name;
-  gboolean ret = TRUE;
+  gboolean ret = FALSE;
```
gichan-jang, line 733: How about using `filter->loaded` as the return value?

jaeyun-jung: Thank you, I found the error case to revert the option and updated this PR.

About your suggestion, plz consider the case if filter-loaded is true (changing option string)
If option is set and trying to update option string, filter->loaded might be true.

gichan-jang: Understand. Thank you for your kind explanation.

PR #2343, opened 2020-05-04
[Android/AMC] Set the omitted finalize function

again4you:
`gst_amc_src_finalize()` exists in the code but the finalize of GObject
class is not set. Because of this reason, gst_amc_src_finalize() function
is not called at all. This patch fixes this bug.

**Self evaluation:**


myungjoo:
Welcome Back!

myungjoo:
> LGTM.
> 
> Why is this "WIP"?

@again4you Ping?

again4you:
@myungjoo Since @jaeyun-jung already tested it so it's ready to merge.

PR #2348, opened 2020-05-06
[Mediapipe] upload the new subplugin

helloahn:
this subplugin helps running mediapipe graph through nnstreamer

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
@@ -1,0 +1,359 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * GStreamer Tensor_Filter, Mediapipe Module
+ * Copyright (C) 2020 HyoungJoo Ahn <hello.ahn@samsung.com>
+ */
+/**
+ * @file   tensor_filter_mediapipe.cc
+ * @date   06 May 2020
+ * @brief  Mediapipe module for tensor_filter gstreamer plugin
+ * @see    http://github.com/nnstreamer/nnstreamer
+ * @author HyoungJoo Ahn <hello.ahn@samsung.com>
+ * @bug    No known bugs except for NYI items
+ * @todo   If there is other input type (not video), or multiple inputs, current system cannot handle it.
+ *
+ * This is the per-NN-framework plugin (mediapipe) for tensor_filter.
+ */
+#include <iostream>
+#include <string>
+#include <stdexcept>
+
+#include <nnstreamer_cppplugin_api_filter.hh>
+#include <tensor_common.h>
+
+#include <mediapipe/framework/calculator_framework.h>
+#include <mediapipe/framework/formats/image_frame.h>
+#include <mediapipe/framework/formats/image_frame_opencv.h>
+#include <mediapipe/framework/port/commandlineflags.h>
+#include <mediapipe/framework/port/file_helpers.h>
+#include <mediapipe/framework/port/opencv_highgui_inc.h>
+#include <mediapipe/framework/port/opencv_imgproc_inc.h>
+#include <mediapipe/framework/port/opencv_video_inc.h>
+#include <mediapipe/framework/port/parse_text_proto.h>
+#include <mediapipe/framework/port/status.h>
+
+/**
+ * @brief Macro for debug mode.
+ */
+#ifndef DBG
+#define DBG FALSE
+#endif
+
+using nnstreamer::tensor_filter_subplugin;
+
+namespace nnstreamer {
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
```diff
@@ -0,0 +1,358 @@
...
+#include "mediapipe/framework/calculator_framework.h"
+#include "mediapipe/framework/formats/image_frame.h"
+#include "mediapipe/framework/formats/image_frame_opencv.h"
+#include "mediapipe/framework/port/commandlineflags.h"
+#include "mediapipe/framework/port/file_helpers.h"
+#include "mediapipe/framework/port/opencv_highgui_inc.h"
+#include "mediapipe/framework/port/opencv_imgproc_inc.h"
+#include "mediapipe/framework/port/opencv_video_inc.h"
+#include "mediapipe/framework/port/parse_text_proto.h"
+#include "mediapipe/framework/port/status.h"
```
myungjoo, line 33: We'd better provide -I${mediapipe-header-path} to compilers and use <path> instead of "path" for #include ${mediapipe-headers}.

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
```diff
@@ -0,0 +1,358 @@
...
+{
+#if (DBG)
+  gint64 start_time = g_get_real_time ();
+#endif
+  gsize file_size;
+  gchar *content = nullptr;
+  GError *file_error = nullptr;
+  mediapipe::Status status;
+
+  g_assert (config_path != nullptr);
```
myungjoo, line 184: unify assertion usage. use assert here or use g_assert in elsewhere.

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/** 
+ * @brief there is no model info available from the mediapipe.
+ *        For this reason, the acquired properties gotten at configuration will be filled.
+ * @return 0 if OK. non-zero if error.
+ */
+int mediapipe_subplugin::getModelInfo (model_info_ops ops, GstTensorsInfo &in_info, GstTensorsInfo &out_info)
+{
+  if (ops == GET_IN_OUT_INFO) {
```
dongju-chae, line 309: How about this style?
```
if (ops != GET_IN_OUT_INFO)
   return -ENOENT;
gst_...
gst_...
return 0;
```

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
```diff
@@ -0,0 +1,358 @@
...
+  gst_tensors_info_free (&inputInfo);
+  gst_tensors_info_free (&outputInfo);
+
+  for (unsigned int i = 0; i < inputInfo.num_tensors; i++) {
+    status = graph.CloseInputStream (inputInfo.info[i].name);
+    if (!status.ok ()) {
+      std::cerr << "Failed to close input stream" << std::endl;
+    }
+  }
+  status = graph.WaitUntilDone ();
```
dongju-chae, line 118: Q. does this function work without configure/initialization?

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
helloahn: you mean shorter paths? or something else?

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
helloahn: it's applied :)

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
helloahn: it looks better. I applied it :)

ext/nnstreamer/tensor_filter/tensor_filter_mediapipe.cc
helloahn: No, it has to be configured first. Since the configuration is performed right after getting `emptyInstance`, and if it is failed, the `runtime_error` will be called.

Therefore, IMO, we don't need to concern about the NOT configuring case at the destructor.

dongju-chae: Understood.

helloahn:
@dongju-chae I understand what you mean. I'll concern the way to make it better

PR #2360, opened 2020-05-07
[single] cache tensor filter class once at open @open sesame 5/9 16:48

kparichay:
Existing implementation seeks the tensor filter class based on its type everytime a class to class is needed.
Rather, cache the tensor filter class increasing its ref, and at unref at single api close.

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -93,21 +93,22 @@ typedef enum
 /** ML single api data structure for handle */
 typedef struct
 {
-  GTensorFilterSingle *filter;  /**< tensor filter element */
-  ml_tensors_info_s in_info;    /**< info about input */
-  ml_tensors_info_s out_info;   /**< info about output */
-  ml_nnfw_type_e nnfw;          /**< nnfw type for this filter */
-  guint magic;                  /**< code to verify valid handle */
+  GTensorFilterSingleClass *klass;    /**< tensor filter class structure*/
+  GTensorFilterSingle *filter;        /**< tensor filter element */
+  ml_tensors_info_s in_info;          /**< info about input */
+  ml_tensors_info_s out_info;         /**< info about output */
+  ml_nnfw_type_e nnfw;                /**< nnfw type for this filter */
+  guint magic;                        /**< code to verify valid handle */
 
-  GThread *thread;              /**< thread for invoking */
-  GMutex mutex;                 /**< mutex for synchronization */
-  GCond cond;                   /**< condition for synchronization */
-  ml_tensors_data_h input;      /**< input received from user */
-  ml_tensors_data_h *output;    /**< output to be sent back to user */
-  guint timeout;                /**< timeout for invoking */
-  thread_state state;           /**< current state of the thread */
-  gboolean ignore_output;       /**< ignore and free the output */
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -338,9 +299,7 @@ ml_single_set_gst_info (ml_single * single_h, const ml_tensors_info_h info)
     case ML_NNFW_TYPE_CUSTOM_FILTER:
       ml_tensors_info_copy_from_ml (&gst_in_info, info);
 
-      klass = g_type_class_peek (G_TYPE_TENSOR_FILTER_SINGLE);
-      if (klass == NULL
-          || klass->set_input_info (single_h->filter, &gst_in_info,
+      if (single_h->klass->set_input_info (single_h->filter, &gst_in_info,
```
again4you, line 302: In previous invoke() case, logical negation (i.e. !) is used instead of comparing `FALSE`. (160 line) However, comparing `FALSE` is used in this case. It would be better when keeping consistency.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -500,12 +454,20 @@ ml_single_create_handle (ml_nnfw_type_e nnfw)
...
   single_h->state = IDLE;
   single_h->ignore_output = FALSE;
+  single_h->thread = NULL;
 
   ml_tensors_info_initialize (&single_h->in_info);
   ml_tensors_info_initialize (&single_h->out_info);
   g_mutex_init (&single_h->mutex);
   g_cond_init (&single_h->cond);
 
+  single_h->klass = g_type_class_ref (G_TYPE_TENSOR_FILTER_SINGLE);
```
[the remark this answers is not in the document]
kparichay: As per documentation, you are right. However, as per [function definition](https://github.com/GNOME/glib/blob/3dec72b946a527f4b1f35262bddd4afb060409b7/gobject/gtype.c#L2921), it can return NULL when the requested class type cannot be found

api/capi/src/nnstreamer-capi-single.c
kparichay: Updated.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -338,10 +299,8 @@ ml_single_set_gst_info (ml_single * single_h, const ml_tensors_info_h info)
     case ML_NNFW_TYPE_CUSTOM_FILTER:
       ml_tensors_info_copy_from_ml (&gst_in_info, info);
 
-      klass = g_type_class_peek (G_TYPE_TENSOR_FILTER_SINGLE);
-      if (klass == NULL
-          || klass->set_input_info (single_h->filter, &gst_in_info,
-              &gst_out_info) == FALSE) {
+      if (!single_h->klass->set_input_info (single_h->filter, &gst_in_info,
+              &gst_out_info)) {
```
dongju-chae, line 303: `status = single_h->klass->set_input_info (...)` ?

kparichay: return type for `set_input_info` is gboolean, and status is `ML_ERROR` type.

dongju-chae: Ah, right.

Issue #2326, opened 2020-04-28
[Potential Bug] ml_single_set_gst_info handles TFLITE & CUSTOM only


myungjoo:
ml_single_set_gst_info() of nnstreamer-capi-single.c supports
ML_NNFW_TYPE_{TENSORFLOW_LITE, CUSTOM_FILTER} only.

What about other frameworks and hardware accelerators?

kparichay:
@myungjoo Yes, `ml_single_set_gst_info` should treat all the frameworks supported by the single-api same. `g_tensor_filter_set_input_info` in `tensor_filter_single.c` should be able to handle if this the set info is supported or not and return more detailed errors than just `gboolean`.

PR #2361, opened 2020-05-07
[single] single API to process nnfw transparently once opened

kparichay:
single API should process all the frameworks transparently once the single handle has been opened.
The framework check has been performed when the handle was created and opened.
If the operation is not supported, that should be handled in the nnfw rather than in single API.

Resolves: #2326 

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -293,27 +293,21 @@ ml_single_set_gst_info (ml_single * single_h, const ml_tensors_info_h info)
 {
   GstTensorsInfo gst_in_info, gst_out_info;
   int status = ML_ERROR_NONE;
+  int ret = -EINVAL;
 
-  switch (single_h->nnfw) {
-    case ML_NNFW_TYPE_TENSORFLOW_LITE:
-    case ML_NNFW_TYPE_CUSTOM_FILTER:
-      ml_tensors_info_copy_from_ml (&gst_in_info, info);
+  ml_tensors_info_copy_from_ml (&gst_in_info, info);
 
-      if (!single_h->klass->set_input_info (single_h->filter, &gst_in_info,
-              &gst_out_info)) {
-        status = ML_ERROR_INVALID_PARAMETER;
-        goto exit;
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -293,26 +293,23 @@ ml_single_set_gst_info (ml_single * single_h, const ml_tensors_info_h info)
...
       status = ML_ERROR_NOT_SUPPORTED;
-      break;
+    else
+      status = ML_ERROR_INVALID_PARAMETER;
+    goto exit;
   }
 
+  ml_tensors_info_copy_from_gst (&single_h->in_info, &gst_in_info);
+  ml_tensors_info_copy_from_gst (&single_h->out_info, &gst_out_info);
+
```
dongju-chae, line 312: How about this?
```
ret = ...
if (ret == 0) {
  ml_tensors_info_copy_from_gst (&single_h->in_info, &gst_in_info);
  ml_tensors_info_copy_from_gst (&single_h->out_info, &gst_out_info);
} else if (ret == -ENOENT) {
  status = ML_ERROR_NOT_SUPPORTED;
} else {
  status = ML_ERROR_INVALID_PARAMETER;
}
return status;
```


```

File: api/capi/src/tensor_filter_single.c [hunk 2]
```diff
@@ -308,19 +308,18 @@ g_tensor_filter_single_invoke (GTensorFilterSingle * self,
  * @brief Set input tensor information in the framework
  * @param self "this" pointer
  * @param in_info information on the input tensor
- * @return TRUE if there is no error.
+ * @return 0 for success, -errno for failure.
```
dongju-chae, line 311: let's add comments for `out_info`

kparichay: good catch, updated.

api/capi/src/nnstreamer-capi-single.c
kparichay: Updated. Will try to avoid goto unless essentially needed.

PR #2363, opened 2020-05-08
[Test] add simple testcases

jaeyun-jung:
add simple testcases for native c-api and internal functions.

### DIFF
--- a/tests/tizen_capi/unittest_tizen_capi.cc
+++ b/tests/tizen_capi/unittest_tizen_capi.cc
@@ -14,6 +14,7 @@
 #include <glib/gstdio.h> /* GStatBuf */
 #include <nnstreamer-capi-private.h>
 #include <nnstreamer_conf.h> /* NNSTREAMER_SO_FILE_EXTENSION */
+#include <nnstreamer_plugin_api.h>
 #include <unittest_util.h>
 
 static const unsigned int SINGLE_DEF_TIMEOUT_MSEC = 10000U;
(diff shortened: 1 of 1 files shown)


tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -1885,27 +1886,272 @@ TEST (nnstreamer_capi_util, info_set_tname_1)
...
+  ASSERT_EQ (status, ML_ERROR_NONE);
+  status = ml_tensors_info_set_count (info, 1);
+  ASSERT_EQ (status, ML_ERROR_NONE);
+
+  status = ml_tensors_info_get_tensor_name (nullptr, 0, &name);
   ASSERT_EQ (status, ML_ERROR_INVALID_PARAMETER);
+  status = ml_tensors_info_get_tensor_name (info, 0, nullptr);
+  ASSERT_EQ (status, ML_ERROR_INVALID_PARAMETER);
+  status = ml_tensors_info_get_tensor_name (info, 2, &name);
+  ASSERT_EQ (status, ML_ERROR_INVALID_PARAMETER);
```
dongju-chae, line 1903: How about using `EXPECT_EQ()` instead of `ASSERT_EQ()` for line 1898-1903 which this unittest really wants to check.

jaeyun-jung: updated

gichan-jang: There are a lot of continuous ASSERT tests, I think SVACE will make a problem.

PR #2364, opened 2020-05-08
[C-Api/CodeClean] validate model file

jaeyun-jung:
add function for nnfw auto detection and clean up functions to validate model file.

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -898,33 +898,74 @@ ml_initialize_gstreamer (void)
 }
 
 /**
- * @brief Internal helper function to validate arguments
+ * @brief Internal helper function to validate model files.
  */
 static int
-_ml_validate_model_file (char **model, unsigned int num_models,
-    ml_nnfw_type_e * nnfw, gchar ** p1, gchar ** p2)
+_ml_validate_model_file (char **model, unsigned int num_models)
 {
-  if (!model) {
+  guint i;
+
+  if (!model || num_models < 1) {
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -898,33 +898,74 @@ ml_initialize_gstreamer (void)
...
+    *nnfw = ML_NNFW_TYPE_TENSORFLOW_LITE;
+  } else if (g_ascii_strcasecmp (detected, "vivante") == 0) {
+    *nnfw = ML_NNFW_TYPE_VIVANTE;
+  } else if (g_ascii_strcasecmp (detected, "tensorflow") == 0) {
+    *nnfw = ML_NNFW_TYPE_TENSORFLOW;
+  } else if (g_ascii_strcasecmp (detected, "custom") == 0) {
+    *nnfw = ML_NNFW_TYPE_CUSTOM_FILTER;
+  } else {
+    ml_loge ("The given model has unknown or not supported extension.");
+    status = ML_ERROR_INVALID_PARAMETER;
```
again4you, line 955: gst_tensor_filter_framework_auto_detection() can returns other FWs such as caffe2, pytorch as so on.
In that case, does _ml_detect_framework() function return ML_ERROR_INVALID_PARAMETER?

helloahn: how about `ML_ERROR_NOT_SUPPORTED`?

jaeyun-jung: changed to not-supported

PR #2368, opened 2020-05-11
[Demux/Split] unref incoming buffer

jaeyun-jung:
After demux/split the incoming buffer, unref the buffer to decrease ref count.

### DIFF
--- a/gst/nnstreamer/tensor_demux/gsttensordemux.c
+++ b/gst/nnstreamer/tensor_demux/gsttensordemux.c
@@ -462,7 +462,7 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     srcpad = gst_tensor_demux_get_tensor_pad (tensor_demux, &created, i);
 
     outbuf = gst_buffer_new ();
-    mem = gst_buffer_peek_memory (buf, i);
+    mem = gst_buffer_get_memory (buf, i);
     gst_buffer_append_memory (outbuf, mem);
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_demux/gsttensordemux.c
```diff
@@ -462,7 +462,7 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     srcpad = gst_tensor_demux_get_tensor_pad (tensor_demux, &created, i);
 
     outbuf = gst_buffer_new ();
-    mem = gst_buffer_peek_memory (buf, i);
+    mem = gst_buffer_get_memory (buf, i);
```
myungjoo, line 465: This is in a for loop. With up to 16 mem (gst-buffer's memory), you are unref-ing only one of them.

myungjoo, line 465: refer: https://developer.gnome.org/gstreamer/stable/gstreamer-GstBuffer.html#gst-buffer-get-memory

You need gst_memory_unref()

jaeyun-jung: gst_buffer_append_memory() will take the ownership of mem-block, so it is necessary to increase ref count of memory before calling the gst_buffer_unref (buf)

myungjoo: Oh.. great!

PR #2370, opened 2020-05-11
[single] Added documentation for the thread behavior @open sesame 5/12 09:23

### LINKED ISSUE #2185: C-API/Single Invoke/Wait Semantics
We need to document more about the state handling of single/invoke. (about the thread behavior)

kparichay:
Added documentation for the thread behavior in various functions in single API
Also removed ERROR state as it was redundant after #2360

Resolves #2185

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -86,8 +86,7 @@ typedef enum
 {
   IDLE = 0,           /**< ready to accept next input */
   RUNNING,            /**< running an input, cannot accept more input */
-  JOIN_REQUESTED,     /**< should join the thread, will exit soon */
-  ERROR               /**< error on thread, will exit soon */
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -86,8 +86,7 @@ typedef enum
 {
   IDLE = 0,           /**< ready to accept next input */
   RUNNING,            /**< running an input, cannot accept more input */
-  JOIN_REQUESTED,     /**< should join the thread, will exit soon */
-  ERROR               /**< error on thread, will exit soon */
+  JOIN_REQUESTED      /**< should join the thread, will exit soon */
```
myungjoo, line 89: If you remove an element from this public-API-header, you will be required to process ACR. Do you really think this removal is required? (you may keep it as a "reserved" value.)

PR #2372, opened 2020-05-12
[filter] Handle auto/default semantics for tensor_filter

### LINKED ISSUE #2320: [Filter/Main] "AUTO" "DEFAULT" Semantics
The semantics and corresponding behaviors of HW-AUTO and HW-DEFAULT should be defined explicitly.

AUTO/DEFAULT are handled as if they are yet another hardware type; however, this is not appropriate. For example, if a subplugin declares that it supports a hardware type "X", with AUTO or DEFAULT, tensor-filter main should choose "X" transparently from the subplugin and the users.

From https://github.com/nnstreamer/nnstreamer/pull/2194#discussion_r415111018

> I suppose that a subpluigin will simply provide a list of hardware accelerators that it supports and a infra (tensor_filter main) should choose appropriately.

>> Yes, this exists. However, that does not handle default or auto.

> For example if "default or auto" is chosen, a tensor_filter should be able to choose "ACCL_NPU_EDGE_TPU" if the chosen subpluigin is edgetpu. I thought the current tensor-filter does this already. Doesn't it do so?

>> The current tensor-filter sets the accelerator backend out of the list provided by the sub-plugin. Choosing which backend to use in case of default or auto is not yet done in tensor_filter main (tensor_filter has no means to decide the default or auto backend for now especially with many filters supporting multiple backends).

kparichay:
Patch to handle auto/default semantics for tensor_filter

Each tensor_filter can now pass its default/auto accelerator
If not passed, the value defaults to first entry from supported list of accelerators
Workaround used to provide this functionality in c - this helps running exisiting code error free with this change, and does not need each filter to essentially declare auto/default accelerators

Passed user input is parsed and if auto/default is used, corresponding accelerators are set and used.

This resolves #2320 for tensor_filter

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
@@ -45,8 +45,6 @@
 #define OUTPUT_TENSOR_META_CHAR "OutputTensorMeta"
 
 static const gchar *torch_accl_support[] = {
-  ACCL_AUTO_STR,
-  ACCL_DEFAULT_STR,
   ACCL_CPU_STR,
   ACCL_GPU_STR,
   NULL
@@ -131,7 +129,7 @@ TorchCore::setAccelerator (const char *accelerators)
   accelerator = parse_accl_hw (accelerators, torch_accl_support);
   if (accelerator == ACCL_NONE)
     goto use_gpu_ini;
-  if ((accelerator & (ACCL_CPU | ACCL_DEFAULT)) != 0)
+  if ((accelerator & (ACCL_CPU)) != 0)
     use_gpu = FALSE;
 
   return;
@@ -494,7 +492,6 @@ TorchCore::invoke (const GstTensorMemory * input, GstTensorMemory * output)
     ml_loge ("Output is not a tensor.");
     return -3;
   }
-
 
 #if (DBG)
   gint64 stop_time = g_get_real_time ();
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -48,15 +48,20 @@
 #endif
 
 static const gchar *tflite_accl_support[] = {
-  ACCL_AUTO_STR,
-  ACCL_DEFAULT_STR,
-  ACCL_CPU_SIMD_STR,
   ACCL_CPU_NEON_STR,
+  ACCL_CPU_SIMD_STR,
   ACCL_CPU_STR,
   ACCL_GPU_STR,
   ACCL_NPU_STR,
   NULL
 };
+
+#if defined(__x86_64__) || defined(__aarch64__) || defined(__arm__)
+static const gchar *tflite_accl_auto = ACCL_CPU_SIMD_STR;
+#else
+static const gchar *tflite_accl_auto = ACCL_CPU_STR;
+#endif
+static const gchar *tflite_accl_default = ACCL_CPU_STR;
 
 /**
  * @brief Wrapper class for TFLite Interpreter to support model switching
@@ -538,7 +543,8 @@ TFLiteCore::TFLiteCore (const char * _model_path, const char * accelerators)
 void TFLiteCore::setAccelerator (const char * accelerators)
 {
   use_nnapi = TRUE;
-  accelerator = parse_accl_hw (accelerators, tflite_accl_support);
+  accelerator = parse_accl_hw (accelerators, tflite_accl_support,
+      tflite_accl_auto, tflite_accl_default);
   if (accelerators == NULL || accelerator == ACCL_NONE)
     goto use_nnapi_ini;
 
@@ -550,7 +556,7 @@ use_nnapi_ini:
   if (use_nnapi == FALSE) {
     accelerator = ACCL_NONE;
   } else {
-    accelerator = ACCL_AUTO;
+    accelerator = get_accl_hw_type (tflite_accl_auto);
   }
 }
 
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -174,6 +174,8 @@ typedef struct _GstTensorFilterFrameworkInfo
   int verify_model_path; /**< TRUE(nonzero) when the NNS framework, not the sub-plugin, should verify the path of model files. */
   const accl_hw *hw_list; /**< List of supported hardwares by the framework.  Positive response of this check does not guarantee successful running of model with this accelerator. Subplugin is supposed to allocate/deallocate. */
   int num_hw; /**< number of hardware accelerators in the hw_list supported by the framework */
+  accl_hw accl_auto;  /**< accelerator to be used in auto mode (acceleration to be used but accelerator is not specified for the filter) - default -1 implies use first entry from hw_list */
+  accl_hw accl_default;   /**< accelerator to be used by default (valid user input is not provided) - default -1 implies use first entry from hw_list*/
 } GstTensorFilterFrameworkInfo;
 
 /**
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -395,7 +395,7 @@ struct _GstTensorFilterFramework
 
       int (*getFrameworkInfo) (const GstTensorFilterFramework * self,
           const GstTensorFilterProperties * prop, void *private_data,
-	  GstTensorFilterFrameworkInfo *fw_info);
+          GstTensorFilterFrameworkInfo *fw_info);
       /**< Mandatory callback. Get the frameworks statically determined info. Argument 'private_data' can be NULL. If provided 'private_data' is not NULL, then some info, such as 'allocate_in_invoke', can be updated based on the model being used (inferred from the 'private_data' provided). This updated info is useful for custom filter, as some custom filter's ability to support 'allocate_in_invoke' depends on the opened model.
        *
        * @param[in] prop read-only property values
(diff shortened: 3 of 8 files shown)


kparichay:
@myungjoo Please provide an initial review before I proceed forward with this.

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -131,7 +129,7 @@ TorchCore::setAccelerator (const char *accelerators)
   accelerator = parse_accl_hw (accelerators, torch_accl_support);
   if (accelerator == ACCL_NONE)
     goto use_gpu_ini;
-  if ((accelerator & (ACCL_CPU | ACCL_DEFAULT)) != 0)
+  if ((accelerator & (ACCL_CPU)) != 0)
```
myungjoo, line 132: What if ```accelerator == 0x3000``` ?

kparichay: I am not sure if I understand the question.
`parse_accl_hw` will return a accelerator which will ACCL_NONE or one of the accelerators supported by the framework (ACCL_CPU or ACCL_GPU) in this case.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -736,15 +754,20 @@ gst_tensor_filter_parse_accelerator (GstTensorFilterPrivate * priv,
...
   }
 
-  /** Convert the list to string format */
-  accl_support = g_malloc (sizeof (gchar *) * (info->num_hw + 1));
+  /**
+   * Convert the list to string format
+   * Extra 2 entries for basic accelerators : auto and default
+   */
+  accl_support = g_malloc (sizeof (gchar *) * (info->num_hw + 2 + 1));
   if (!accl_support)
```
dongju-chae, line 762: This error handing will be never called. Let's remove it.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1869,20 +1902,64 @@ parse_accl_hw_all (const gchar * accelerators,
...
+  const gchar **accl_support;
+
+  /** Count number of elements for the array */
+  while (supported_accelerators[num_hw] != NULL)
+    num_hw += 1;
+  num_hw += 2;
+
+  /** Allocate the array */
+  accl_support = g_malloc (sizeof (gchar *) * (num_hw + 1));
+  if (!accl_support)
```
dongju-chae, line 1922: same here

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
myungjoo: If user gives ```0x3000``` (CPU | GPU), pytorch may use either CPU or GPU. Thus, disabling GPU and enabling CPU only if 0x1000 (CPU bit) is enabled, is not "incorrect" logically, but, would not meet users' expectation normally. Users will expect (and understand) that it will use GPU if GPU is enabled if it's ```0x3000``` (CPU | GPU)

myungjoo: Ah... parse_accl_hw_util will give a single element. OK.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -48,15 +48,16 @@
...
-  ACCL_AUTO_STR,
-  ACCL_DEFAULT_STR,
   ACCL_CPU_NEON_STR,
   ACCL_CPU_STR,
   ACCL_GPU_STR,
   ACCL_NPU_STR,
   NULL
 };
 
+static const gchar *tflite_accl_auto = ACCL_CPU_NEON_STR;
```
myungjoo, line 58: You may need to determine this based on the given architecture. x64 is not to be ignored :)

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
kparichay: > uld not meet users' expectation normally. Users will expect (and understand) that it will use GPU if GPU is enabled if it's 0x3000 (CPU | GPU

On the side note, user is supposed to give `CPU, GPU` than `CPU | GPU`.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
kparichay: aah yes, that will be good. I will update as per this.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: done.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: done.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
myungjoo: Uh __x86_64__ does not have NEON. Do you intend to designate SSE/AVS w/ NEON?

kparichay: Yes, I was intending NEON was for generic SIMD based processing.
How about rename CPU_NEON -> CPU_SIMD?

myungjoo: For string values (```ACCL_*_STR```), we can add ```"cpu.simd"```
For enum values, let's define ```ACCL_CPU_SIMD = ACCL_CPU_NEON```.
For enum in API, we can define ```ML_NNFW_HW_CPU_SIMD = ML_NNFW_HW_CPU_NEON``` to keep the backward compatibility and minimize API changes, making it easier to proceed with ACR.

kparichay: Added it in #2377 as this PR is already touching too many files.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -172,6 +172,8 @@ typedef struct _GstTensorFilterFrameworkInfo
   int verify_model_path; /**< TRUE(nonzero) when the NNS framework, not the sub-plugin, should verify the path of model files. */
   const accl_hw *hw_list; /**< List of supported hardwares by the framework.  Positive response of this check does not guarantee successful running of model with this accelerator. Subplugin is supposed to allocate/deallocate. */
   int num_hw; /**< number of hardware accelerators in the hw_list supported by the framework */
+  accl_hw accl_auto;  /**< accelerator to be used in auto mode (acceleration to be used but accelerator is not specified for the filter) - default -1 implies use first entry from hw_list */
+  accl_hw accl_default;   /**< accelerator to be used by default (valid user input is not provided) - default -1 implies use first entry from hw_list*/
```
myungjoo, line 176: Question: can each subplugin alter auto/default values in run-time and affect the behaviors?
If then, LGTM.

kparichay: If the plugin is closed, then yes, its effect will be reflected in run-time (opening the plugin would call `getFrameworkInfo()` which will get updated values from the framework itself).
If the plugin is already opened, then only reopen will reflect the update in runtime.

But this is all assuming that the filter returns the updated value with `getFrameworkInfo()`.

kparichay: If the subplugin is already opened, then a reopen can update the filter.
If we want to trigger the change without a reopen, the selected ACCL (based on the user-input) can be made available with `GstTensorFilterProperty` to the plugin, and the plugin needs to do it by itself.

For example - if user input it `true:auto` or `true:auto,gpu,cpu` etc, the `GstTensorFilterProperty` will contain info that accl `auto` is selected (when opening the plugin, this information is not passed. Rather, just the specified accelerator designated is `auto` is passed). Then when the plugin updates its auto/default values, it can check that the `auto` value is being used, and the plugin should by itself update its behaviors.
This change is not being driven by `tensor_filter` as subplugin cannot notify tensor_filter that it has altered any of its properties/configuration (`tensor_filter` always starts the communication between itself and its subplugins).

However, note that this will not support updating the supported accelerator lists because that would require matching it with the user-provided input which is hidden from the subpluins.

myungjoo: If a filter's option is auto and it may use NPU-1 and NPU-2, we may want to use different NPU in run-time (e.g., if NPU-1's utilization by another process is too high, use NPU-2)

We do not have such filters prepared, but in the future, I expect that users will want to use "auto" for such purposes. For now, we just need to "allow" a subplugin to do so. Note that this does not necessarily require yet another callback, we may assume that a subplugin may change it in a invoke callback while a subplugin who does not aware of "auto" or "default" will get one (other than auto/default) automatically). **Anyway.. this looks self-contradicting...**: we cannot achieve both simultaneously in the current structure of nnstreamer. :(

So.. for now... your approach looks fair enough.

LGTM.

kparichay: > If a filter's option is auto and it may use NPU-1 and NPU-2, we may want to use different NPU in run-time (e.g., if NPU-1's utilization by another process is too high, use NPU-2)

In this example, switching between NPU can be done by two entities - 
1. By subplugin itself - this can be done by using the hierarchy in the accl enum. NPU enum would include NPU-1 and NPU-2. NPU-1 enum only marks NPU-1 and same for NPU-2 (if NPU-1 and NPU-2 have the same common enum NPU, then also this applies). If filter auto sets NPU enum, then filter is opened with NPU enum, giving it anonymity to choose which NPU to use and update it in run-time based on utilization or other criteria.
2. By nnstreamer itself - other than scheduling required to be done in nnstreamer, passing just a device enum with every invoke can do the job (this will change the internal APIs again though).  
 
> We do not have such filters prepared, but in the future, I expect that users will want to use "auto" for such purposes. For now, we just need to "allow" a subplugin to do so. 

For this, I will add the property to `GstTensorFilterProperty` as mentioned in https://github.com/nnstreamer/nnstreamer/pull/2372#discussion_r424939871.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1869,20 +1900,62 @@ parse_accl_hw_all (const gchar * accelerators,
...
+  accl_support[idx] = NULL;
+
+  return accl_support;
+}
+
 /**
  * @brief parse user given string to extract accelerator based on given regex
  * @param[in] accelerators user given input
  * @param[in] supported_accelerators list ofi supported accelerators
+ * @param[in] auto_accelerator accelerator to use in auto case (when acceleration is enabled but specific accelerator is not provided or not matching)
```
zhoonit, line 1937: It seems that user might unexpectedly encounter a situation that different accelerator is used against user's choice.
Is there any code bit that warn user when this happens?

kparichay: Good point, added at https://github.com/nnstreamer/nnstreamer/pull/2372/files#diff-97faf0d14d439e63e1544199a92ec25aR1891

zhoonit: Thanks :)

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -496,10 +498,33 @@ extern const char *
...
+ * @note The order of argumemnts for calling this function is:
+ *       - in_accl: user provided input accelerator string
+ *       - sup_accl: list of supported accelerator
+ *       - auto_accl: auto accelerator (optional)
+ *       - def_accl: default accelerator (optional)
+ */
+extern accl_hw parse_accl_hw_fill (parse_accl_args accl_args);
+
+/**
+ * @bried workaround to provide default arguments
```
dongju-chae, line 525: typo detected

kparichay: typo corrected.

PR #2377, opened 2020-05-14
[hw/accl] Add CPU_SIMD as a backend option

kparichay:
Added CPU_SIMD as a backend option
CPU_NEON will be alias for CPU_SIMD in enum but different string representation
Only updated tflite supported accelerators list. Others will be updated after #2372.

Refer https://github.com/nnstreamer/nnstreamer/pull/2372#discussion_r424857942 for more details.

cc. @myungjoo 

### DIFF
--- a/api/capi/include/nnstreamer.h
+++ b/api/capi/include/nnstreamer.h
@@ -139,7 +139,8 @@ typedef enum {
   ML_NNFW_HW_ANY          = 0,      /**< Hardware resource is not specified. */
   ML_NNFW_HW_AUTO         = 1,      /**< Try to schedule and optimize if possible. */
   ML_NNFW_HW_CPU          = 0x1000, /**< 0x1000: any CPU. 0x1nnn: CPU # nnn-1. */
-  ML_NNFW_HW_CPU_NEON     = 0x1100, /**< 0x1100: NEON in CPU. (Since 6.0) */
+  ML_NNFW_HW_CPU_SIMD     = 0x1100, /**< 0x1100: SIMD in CPU. (Since 6.0) */
+  ML_NNFW_HW_CPU_NEON     = 0x1100, /**< 0x1100: NEON (alias for SIMD) in CPU. (Since 6.0) */
   ML_NNFW_HW_GPU          = 0x2000, /**< 0x2000: any GPU. 0x2nnn: GPU # nnn-1. */
   ML_NNFW_HW_NPU          = 0x3000, /**< 0x3000: any NPU. 0x3nnn: NPU # nnn-1. */
   ML_NNFW_HW_NPU_MOVIDIUS = 0x3001, /**< 0x3001: Intel Movidius Stick. (Since 6.0) */
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -21,6 +21,7 @@
 #define ACCL_DEFAULT_STR  "default"
 #define ACCL_AUTO_STR "auto"
 #define ACCL_CPU_STR  "cpu"
+#define ACCL_CPU_SIMD_STR  "cpu.simd"
 #define ACCL_CPU_NEON_STR  "cpu.neon"
 #define ACCL_GPU_STR  "gpu"
 #define ACCL_NPU_STR  "npu"
@@ -86,7 +87,8 @@ typedef enum
 
   /** Enables acceleration, 0xn000 any version of that device, 0xnxxx: device # xxx-1 */
   ACCL_CPU          = 0x1000,     /**< specify device as CPU, if possible */
-  ACCL_CPU_NEON     = 0x1100,     /**< specify device as NEON in cpu, if possible */
+  ACCL_CPU_SIMD     = 0x1100,     /**< specify device as SIMD in cpu, if possible */
+  ACCL_CPU_NEON     = 0x1100,     /**< specify device as NEON (alias for SIMD) in cpu, if possible */
   ACCL_GPU          = 0x2000,     /**< specify device as GPU, if possible */
   ACCL_NPU          = 0x4000,     /**< specify device as any NPU, if possible */
   ACCL_NPU_MOVIDIUS = 0x4001,     /**< specify device as movidius, if possible */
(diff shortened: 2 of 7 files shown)


api/capi/include/nnstreamer.h
```diff
@@ -139,7 +139,8 @@ typedef enum {
   ML_NNFW_HW_ANY          = 0,      /**< Hardware resource is not specified. */
   ML_NNFW_HW_AUTO         = 1,      /**< Try to schedule and optimize if possible. */
   ML_NNFW_HW_CPU          = 0x1000, /**< 0x1000: any CPU. 0x1nnn: CPU # nnn-1. */
-  ML_NNFW_HW_CPU_NEON     = 0x1100, /**< 0x1100: NEON in CPU. (Since 6.0) */
+  ML_NNFW_HW_CPU_SIMD     = 0x1100, /**< 0x1100: SIMD in CPU. (Since 6.0) */
+  ML_NNFW_HW_CPU_NEON     = 0x1100, /**< 0x1100: NEON (alias for NEON) in CPU. (Since 6.0) */
```
myungjoo, line 143: ```/alias for NEON/alias for SIMD/```

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1910,7 +1910,15 @@ accl_hw_get_type (void)
...
       {ACCL_AUTO, ACCL_AUTO_STR, ACCL_AUTO_STR},
       {ACCL_CPU, ACCL_CPU_STR, ACCL_CPU_STR},
+#if defined(__aarch64__) || defined(__arm__)
+      /** Retreive NEON_STR when searching for SIMD/NEON on arm architectures */
       {ACCL_CPU_NEON, ACCL_CPU_NEON_STR, ACCL_CPU_NEON_STR},
+      {ACCL_CPU_SIMD, ACCL_CPU_SIMD_STR, ACCL_CPU_SIMD_STR},
+#else
+      /** Retreive SIMD_STR when searching for NEON/SIMD on x86 architectures */
+      {ACCL_CPU_SIMD, ACCL_CPU_SIMD_STR, ACCL_CPU_SIMD_STR},
+      {ACCL_CPU_NEON, ACCL_CPU_NEON_STR, ACCL_CPU_NEON_STR},
```
myungjoo, line 1920: If it's not ARM, let's not convert 0x1100 to NEON.

kparichay: Did you mean - NEON entry should not be there if it is not ARM?
then the user requesting NEON in non-ARM will not be forwarded to SIMD.

api/capi/include/nnstreamer.h
kparichay: done.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: It's ok to emit errors or ignore if neon is designated for non-ARM machines. They do not have NEON.

SIMD->NEON "translation" is fine, but NEON->SIMD "translation" in non-ARM may be dangerous. Assume that there is a software code that supports NEON but no SSE and will call NEON instructions unconditionally if "neon" is designated (unsafe but not incorrect).

myungjoo: if-ARM section looks good: it states that "neon" "simd" will be translated as NEON and SIMD.
if-x64 section might be undesirable: it will support "neon" in x64. (although it's just X64-SIMD actually.).

tests/nnstreamer_filter_tensorflow_lite/runTest.sh
```diff
@@ -177,7 +177,12 @@ testResult $? 2-16 "NNAPI activation test" 0 1
 
 # Property reading test for nnapi
 run_pipeline true:cpu.neon,cpu
-cat info | grep "nnapi = 1, accl = cpu.neon$"
+arch=$(uname -m)
+if [ "$arch" = "aarch64" ] || [ "$arch" = "armv7l" ]; then
+	cat info | grep "nnapi = 1, accl = cpu.neon$"
+else
+	cat info | grep "nnapi = 1, accl = cpu.simd$"
```
myungjoo, line 184: LGTM. and if accl = cpu.neon in x64, it'd be "Negative Case".

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: That is a very good example. NEON->SIMD is wrong. I will remove this translation, and update x64 section to just have SIMD (neon will not match with SIMD).
This will also impose restrictions on subplugins to not add NEON in supported accelerators list in non-arm architectures (removed ACCL_CPU_NEON_STR in non-arm architectures to check on this).

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -21,7 +21,10 @@
 #define ACCL_DEFAULT_STR  "default"
 #define ACCL_AUTO_STR "auto"
 #define ACCL_CPU_STR  "cpu"
+#define ACCL_CPU_SIMD_STR  "cpu.simd"
+#if defined(__aarch64__) || defined(__arm__)
```
kparichay, line 25: This imposes a restriction on subplugins to not add NEON in supported accelerators list in non-arm architectures.

tests/nnstreamer_filter_tensorflow_lite/runTest.sh
kparichay: this goes to the second option - cpu. Updated.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
myungjoo: This is really a wild assumption, but we may still need to designate NEON in a filter with x64 host; an NPU with ARM core installed in x64 machine. Or in a integrated multi-node pipeline, we may need to say "NEON" from x64 machine where a part of a pipeline might be executed in another node.

We can keep CPU_NEON_STR definition for all arch. We just need to block auto-translation of SIMD -> NEON or NEON -> SIMD in non ARM.

kparichay: The assumption is removed.

PR #2382, opened 2020-05-15
[Flatbuf] Add tensor converter subplugin for flatbuffer

gichan-jang:
Add tensor converter subplugin for flatbuffer.

**Self evaluation:**
1. Build test: [* ]Passed [ ]Failed [ ]Skipped
2. Run test: [ ]Passed [* ]Failed [ ]Skipped

PR #2386, opened 2020-05-18
[Converter] Prepare to register sub-plugin

gichan-jang:
I separated PR from #2382
Prepare to register sub-plugin for flatbuffers and protobuf.

Expected usage in case of flatbuffers :
`flatbuffer byte stream ! application/octet-stream ! tensor_converter mode=flatbuf ! tensor_*`

Self evaluation:


gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -816,25 +849,35 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
       }
       break;
     case _NNS_OCTET:
+      if (self->modename != NULL) {
```
zhoonit, line 852: Q. Is there a case `self->modename == NULL` ?

gichan-jang: Yes, If the user did not assign mode name. 
mode name is used to register sub-plugin.

zhoonit: Thanks for the explanation :)

gst/nnstreamer/include/nnstreamer_plugin_api_converter.h
```diff
@@ -38,38 +38,35 @@ extern "C" {
 /**
  * @brief Converter's subplugin implementation.
  */
-struct _NNStreamerExternalConverter {
+typedef struct _NNStreamerExternalConverter {
```
myungjoo, line 41: We already have ```typedef struct _NNStreamerExternalConverter NNStreamerExternalConverter;``` in tensor_converter.h. Doesn't it matter?

myungjoo:
You can define "other/flatbuf-tensor" and register it for findExternalConverter.
Then, you don't need to add "mode" property.

In pipeline, you may probably reset octet stream as other/flatbuf-tensor with caps-filter. (please test it and tell me if that fails)

gst/nnstreamer/include/nnstreamer_plugin_api_converter.h
gichan-jang: Thank you. Don't need to re-define. I will revert it.

gichan-jang:
> You can define "other/flatbuf-tensor" and register it for findExternalConverter.
> Then, you don't need to add "mode" property.
> E.g.,
> 
> ```
> filesrc whatever.fb ! other/flatbuf-tensor ! tensor_converter ! ...
> ```
> 
> In pipeline, you may probably reset octet stream as other/flatbuf-tensor with caps-filter. (please test it and tell me if that fails)
> 
> The principle when I wrote tensor-converter for non video/audio/text streams (subplugins of converter) was to automate subplugin designation so that tensor-converter can choose subplugins automatically based on the stream type.

I understand principal of tensor converter. I will work with other/flatbuf-tensor

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -79,6 +79,19 @@
...
+#define GST_FLATBUF_TENSOR_CAP_DEFAULT \
+    "other/flatbuf-tensor, " \
+    "framerate = " GST_TENSOR_RATE_RANGE
+    /**
+     * type should be one of types in GST_TENSOR_TYPE_ALL
+     * "type = (string) uint8"
+     * dimension shoule be a formatted string with rank NNS_TENSOR_RANK_LIMIT
+     * "dimension = (string) dim1:dim2:dim3:dim4"
+     */
+
```
gichan-jang, line 94: I'm not sure if this is the right location. I want to hear reviewer's opinion.

gst/nnstreamer/include/nnstreamer_plugin_api_converter.h
```diff
@@ -26,9 +26,11 @@
 
 #include "tensor_typedef.h"
 #include <gst/gst.h>
+#include <tensor_converter/tensor_converter.h>
```
myungjoo, line 29: tensor_convert.h is not to be exposed to other packages.
Which symbol are you going to refer from it?

gst/nnstreamer/include/tensor_typedef.h
myungjoo: if you assume that each gstbuf instance will always have a single frame, that'd be fine. If you need to assume that there might be multiple frames in it, you need to consult with those who have written audio converters.

gst/nnstreamer/nnstreamer_subplugin.c
```diff
@@ -67,7 +67,7 @@ static subpluginSearchLogic searchAlgorithm[] = {
   [NNS_SUBPLUGIN_FILTER] = NNS_SEARCH_FILENAME,
   [NNS_SUBPLUGIN_DECODER] = NNS_SEARCH_FILENAME,
   [NNS_EASY_CUSTOM_FILTER] = NNS_SEARCH_FILENAME,
-  [NNS_SUBPLUGIN_CONVERTER] = NNS_SEARCH_GETALL,
+  [NNS_SUBPLUGIN_CONVERTER] = NNS_SEARCH_FILENAME,
```
myungjoo, line 70: Please state the reason for this.

gst/nnstreamer/include/nnstreamer_plugin_api_converter.h
gichan-jang: Thank you. I reverted it.

gst/nnstreamer/nnstreamer_subplugin.c
gichan-jang: It was changed because `NNS_SEARCH_GETALL` does not seem necessary.
And I'm not aware of what the intention is.

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -1250,6 +1283,48 @@ gst_tensor_converter_parse_octet (GstTensorConverter * self,
...
+  /* get template caps from sub-plugins */
+  total = nnsconf_get_subplugin_info (NNSCONF_PATH_CONVERTERS, &info);
+  if (total > 0) {
+    prefix_str = nnsconf_get_subplugin_name_prefix (NNSCONF_PATH_CONVERTERS);
+    prefix_len = strlen (prefix_str);
+    extension_len = strlen (NNSTREAMER_SO_FILE_EXTENSION);
+
+    for (i = 0; i < total; i++) {
+      len = strlen (info.names[i]) - prefix_len - extension_len;
+      plugin_name = g_strndup (info.names[i] + prefix_len, len);
```
zhoonit, line 1315: Q. is this intended to do `g_size + gchar *` ?

gst/nnstreamer/include/nnstreamer_plugin_api_converter.h
```diff
@@ -38,54 +38,52 @@ extern "C" {
...
-       * @retval Return inbuf if the data is to be kept untouched.
-       * @retval Retrun a new GstBuf if the data is to be modified.
-       */
+  GstBuffer *(*convert) (GstBuffer * buf, gsize * frame_size,
+      guint * frames_in, GstTensorsConfig * config);
+  /**< Convert the given input stream to tensor/tensors stream.
+   * @param[in/out] self A pointer designating "this".
+   * @param[in] buf The input stream buffer
+   * @param[out] frame_size The size of each frame (output buffer)
+   * @param[out] frames_in The number of frames in the given input buffer.
```
dongju-chae, line 52: I think you missed `config`

gichan-jang: Thank you! Updated.

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: For the flatbuffer, the full name of lib file is `libnnstreamer_converter_flatbuf.so`.
This line gets plugin name as `flatbuf`.

zhoonit: Got it. Thanks :)

gichan-jang:
Change tensor config to tensors config.
Assume all media types have single tensor except sub-plugins.

gst/nnstreamer/nnstreamer_subplugin.c
myungjoo: The original intention of GETALL is to load all available converter subplugins so that we can automatically choose subplugins based on the incoming type.

E.g., you need GETALL to attach "flatbuf" plugin automatically for ```/other/flatbuf```.

myungjoo:
PTAL: @jaeyun-jung

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -827,12 +829,45 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
       if (self->externalConverter == NULL ||
           self->externalConverter->convert == NULL)
         return GST_FLOW_NOT_SUPPORTED;
-      inbuf = self->externalConverter->convert (self, buf, &frame_size,
-          &frames_in);
+      inbuf =
+          self->externalConverter->convert (buf, &frame_size, &frames_in,
+          config);
+
+      self->tensors_config = *config;
```
jaeyun-jung, line 836: unnecessary line

gichan-jang: Right. Thank you

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -1622,29 +1685,52 @@ gst_tensor_converter_parse_caps (GstTensorConverter * self,
...
+  tensor_config->rate_d = tensors_config->rate_d;
+}
+
+/**
+ * @brief Get converter's subplugins name
+ */
+static const gchar *
+getExternalConverterName (const char *name)
+{
+  if (g_strcmp0 (name, "other/flatbuf-tensor") == 0) {
```
jaeyun-jung, line 1728: plz mark @todo compare mimetype and return lib name from sub-plugin (not hardcoded value)

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -827,12 +829,45 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
...
+      inbuf =
+          self->externalConverter->convert (buf, &frame_size, &frames_in,
+          config);
+
+      self->tensors_config = *config;
+      if (config->info.num_tensors == 1) {
+        GstStructure *st;
+        GstTensorConfig tensor_config;
+
+        peer_caps = gst_pad_peer_query_caps (self->srcpad, NULL);
```
jaeyun-jung, line 841: need to unref peer-caps

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -463,7 +462,9 @@ gst_tensor_converter_sink_event (GstPad * pad, GstObject * parent,
       silent_debug_caps (in_caps, "in-caps");
 
       if (gst_tensor_converter_parse_caps (self, in_caps)) {
-        out_caps = gst_tensor_caps_from_config (&self->tensor_config);
+        GstTensorConfig tensor_config;
+        gst_tensor_config_from_tensors (&tensor_config, &self->tensors_config);
```
jaeyun-jung, line 466: out_caps maybe other/tensor(s). plz check below.
```
out_caps = gst_tensors_caps_from_config (&self->tensors_config);
if (num-tensors == 1) {
   config single-tensor
  gst_caps_append (out_caps, gst_tensor_caps_from_config (&tensor_config));
}
```

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: Thank you! I will update it

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: But the sub-plugin is not registered yet at this line. You mean `nnstreamer_subplugin.c`?

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: Except for the sub-plugin, the tensor is supported only. 
And sub-plugin updates cap after checking the number of tensor and peer_caps in chain function.

gichan-jang: Modified to support tensor/tensors. Thank you for your valuable comments.

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -457,22 +455,13 @@ gst_tensor_converter_sink_event (GstPad * pad, GstObject * parent,
...
-
-        gst_pad_set_caps (self->srcpad, out_caps);
-
-        gst_event_unref (event);
-        event = gst_event_new_caps (out_caps);
-
-        gst_caps_unref (out_caps);
-        return gst_pad_push_event (self->srcpad, event);
+        return gst_tensor_converter_update_caps (self, self->srcpad,
+            &self->tensors_config);
```
jaeyun-jung, line 464: plz add gst_event_unref (event);

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -824,15 +813,27 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
       break;
     case _NNS_MEDIA_PLUGINS:
     {
+      GstTensorsConfig prev_config;
```
jaeyun-jung, line 816: recommended new_config
if update-caps is ok (line830), then update it. (self->tensors_config = mew_config;)
current patch does not handle this case.

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: Thanks! Added.

gst/nnstreamer/tensor_converter/tensor_converter.c
gichan-jang: I changed it. Please check it one more time.

gichan-jang:
## Changed
 1. Change sub plugin function prototype of the tensor converter
 2. Handle if src pad cap is any

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -1622,29 +1655,93 @@ gst_tensor_converter_parse_caps (GstTensorConverter * self,
...
+static gboolean
+gst_tensor_converter_update_caps (GstTensorConverter * self,
+    GstPad * pad, GstTensorsConfig * config)
+{
+  GstStructure *st;
+  GstCaps *curr_caps, *out_caps = NULL;
+
+  if (config->info.num_tensors == 1) {
+    GstCaps *peer_caps = gst_pad_peer_query_caps (pad, NULL);
+    silent_debug_caps (peer_caps, "********sink event peer caps********");
```
jaeyun-jung, line 1692: plz check additional log message for peer caps

gichan-jang: Ooops, Thanks. This line is removed because it's added for debugging.

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -1247,7 +1256,49 @@ gst_tensor_converter_parse_octet (GstTensorConverter * self,
...
+  gsize prefix_len, extension_len, len;
+
+  /* append default caps */
+  append_video_caps_template (caps);
+  append_audio_caps_template (caps);
+  append_text_caps_template (caps);
+  append_octet_caps_template (caps);
+
+  /* get template caps from sub-plugins */
+  total = nnsconf_get_subplugin_info (NNSCONF_PATH_CONVERTERS, &info);
```
myungjoo, line 1283: These operations are not supposed to be required by plugins. You are doing something duplicated by changing the type to NNS_SEARCH_FILENAME

gichan-jang:
> 1. plugin search mechanism appears to be duplicated
> 2. use load all and search plugins based on the caps.

1. duplicated code is removed.
2. The search algorithm for the converter sub-plugin is changed to `NNS_SEARCH_GETALL`.

gst/nnstreamer/nnstreamer_conf.c
```diff
@@ -37,7 +37,7 @@
 #define NNSTREAMER_PREFIX_DECODER	"libnnstreamer_decoder_"
 #define NNSTREAMER_PREFIX_FILTER	"libnnstreamer_filter_"
 #define NNSTREAMER_PREFIX_CUSTOMFILTERS	""
-#define NNSTREAMER_PREFIX_CONVERTER	""
+#define NNSTREAMER_PREFIX_CONVERTER	"libnnstreamer_converter_"
```
myungjoo, line 40: Why do you need this prefix? You do not use filenames to extract subplugin names, right?

myungjoo: LGTM other than this.

gichan-jang: It is necessary to parse the sub-plugin name from the lib files even with the `NNS_SEARCH_GETALL` option.
`_get_filenames` function requires this prefix.

myungjoo: Uh... does ```_get_filenames``` really need it?

```g_str_has_prefix (string, "")``` returns TRUE
and
```strlen ("")``` returns 0
Thus, it appears that ```_get_filenames()``` will work properly with ```""``` prefixes.

gichan-jang: My explanation was insufficient.
The functions that you mentioned can work well with "".
But when a sub-plugin is added, `findExternalConverter` needs sub-plugin name from config info to register it.
Would it be better to move this change to PR #2382?

gichan-jang: Now I understand. I will modify it according to the direction of you reviewed.

gichan-jang: `findExternalConverter` is changed to get sub-plugins according to the mime type regardless of the name of the library.

PR #2389, opened 2020-05-18
[subplugins] Update subplugins accl support @open sesame 5/28 20:02

kparichay:
Update subplugin's accelerators support as per updated rules for auto/default


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -754,6 +754,9 @@ gst_tensor_filter_parse_accelerator (GstTensorFilterPrivate * priv,
     }
   }
 
+  if (info->num_hw == 0)
+    return;
```
dongju-chae, line 758: let's add some log messages using `ml_logw`

PR #2390, opened 2020-05-18
[single] Add missing gstTensorFilterFramework v1 support

kparichay:
Add missing support for gstTensorFilterFramework in checking framework availability
for a given backend accelerator/hardware

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -1218,16 +1218,11 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
 
   if (fw_name) {
     if ((fw = nnstreamer_filter_find (fw_name)) != NULL) {
-      /** @todo support GstTensorFilterV1 here */
-      /** Only check for specific HW, ANY/AUTO are always supported */
-      if (hw == ML_NNFW_HW_ANY || hw == ML_NNFW_HW_AUTO) {
-        *available = true;
-      } else if (fw->checkAvailability
-          && fw->checkAvailability (ml_nnfw_to_accl_hw (hw)) != 0) {
+      *available = gst_tensor_filter_check_hw_availability
+        (fw, ml_nnfw_to_accl_hw (hw));
+      if (*available == false) {
         ml_logw ("%s is supported but not with the specified hardware.",
             fw_name);
-      } else {
-        *available = true;
       }
     } else {
       ml_logw ("%s is not supported.", fw_name);
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -2038,3 +2038,36 @@ accl_hw_get_type (void)
 
   return g_accl_hw_type_id__volatile;
 }
+
+/**
+ * @brief check if the given hw is supported by the framework
+ */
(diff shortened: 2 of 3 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -1218,12 +1218,11 @@ ml_check_nnfw_availability (ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw,
 
   if (fw_name) {
     if ((fw = nnstreamer_filter_find (fw_name)) != NULL) {
-      if (fw->checkAvailability
-          && fw->checkAvailability (ml_nnfw_to_accl_hw (hw)) != 0) {
+      *available = gst_tensor_filter_check_hw_availability
+        (fw, ml_nnfw_to_accl_hw (hw));
+      if (*available == false) {
```
zhoonit, line 1223: Q. Would there be any difference between `false` and `FALSE` as `tensor_filter_common.c:1946` is using `FALSE`?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1934,3 +1934,36 @@ accl_hw_get_type (void)
...
+    ret = fw->getFrameworkInfo (fw, &prop, NULL, &info);
+    if (ret == 0) {
+      for (idx = 0; idx < info.num_hw; idx++) {
+        if (info.hw_list[idx] == hw) {
+          available = TRUE;
+          break;
+        }
+      }
+    } else if (ret == -ENOENT) {
+      available = TRUE;
```
zhoonit, line 1964: Should this be `available = FALSE`?

api/capi/src/nnstreamer-capi-util.c
kparichay: `FALSE` is a macro by glib which equals to `0`. So, I don't think there would be any difference.
As `available` is a bool (and not `gboolean` type), `false` is used instead of `FALSE`.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: aah, yes. `available` should be `FALSE` for all `ret < 0`.

api/capi/src/nnstreamer-capi-util.c
zhoonit: Thanks for the explanation :)

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1934,3 +1934,34 @@ accl_hw_get_type (void)
...
+  gboolean available = FALSE;
+  GstTensorFilterFrameworkInfo info;
+  GstTensorFilterProperties prop;
+
+  if (GST_TF_FW_V0 (fw) &&
+      (!fw->checkAvailability || fw->checkAvailability (hw) == 0)) {
+    available = TRUE;
+  } else if (GST_TF_FW_V1 (fw)) {
+    gst_tensor_filter_properties_init (&prop);
+    ret = fw->getFrameworkInfo (fw, &prop, NULL, &info);
```
dongju-chae, line 1955: How about checking `!fw->getFrameworkInfo`?

kparichay: done

PR #2391, opened 2020-05-19
[Android] fix lint warning

jaeyun-jung:
fix lint report : set default locale

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
@@ -20,6 +20,8 @@ import android.os.Build;
 
 import org.freedesktop.gstreamer.GStreamer;
 
+import java.util.Locale;
+
 /**
  * Defines the types and limits in NNStreamer.<br>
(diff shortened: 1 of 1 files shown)


api/android/api/src/main/java/org/nnsuite/nnstreamer/NNStreamer.java
```diff
@@ -138,8 +140,8 @@ public static boolean isAvailable(NNFWType fw) {
 
         /* sub-plugin for given framework is available */
         if (available) {
-            String manufacturer = Build.MANUFACTURER.toLowerCase();
-            String hardware = Build.HARDWARE.toLowerCase();
+            String manufacturer = Build.MANUFACTURER.toLowerCase(Locale.getDefault());
```
gichan-jang, line 143: Is this just to fix Lint warning?

jaeyun-jung: yes, toLowerCase() is equivalent toLowerCase(Locale.getDefault())
```
String toLowerCase() {
  return toLowerCase(Locale.getDefault());
}
```

PR #2392, opened 2020-05-19
[Android] enable custom filter in singleshot

jaeyun-jung:
register filter subplugins (custom-filter support) when building single-shot library


api/android/api/src/main/jni/nnstreamer-native-api.c
```diff
@@ -677,6 +673,11 @@ nnstreamer_native_initialize (JNIEnv * env, jobject context)
     init_is ();
 #endif
 
+    /* tensor-filter sub-plugins */
```
gichan-jang, line 676: trivial : need to adjust indent

PR #2398, opened 2020-05-21
[Android] cache method id @open sesame 5/28 20:00

jaeyun-jung:
1. get required method ids to improve performance when initializing the pipe info.
2. add internal method to handle arraylist.

### DIFF
--- a/api/android/api/src/main/jni/nnstreamer-native.h
+++ b/api/android/api/src/main/jni/nnstreamer-native.h
@@ -44,6 +44,18 @@
 #endif
 
 /**
+ * @brief NNStreamer package name.
+ */
+#define NNS_PKG "org/nnsuite/nnstreamer"
+#define NNS_CLS_TDATA NNS_PKG "/TensorsData"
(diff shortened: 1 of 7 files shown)


api/android/api/src/main/jni/nnstreamer-native.h
```diff
@@ -43,6 +43,13 @@
 #define CAST_TO_TYPE(l,type) (type)(jint)(l)
 #endif
 
+/**
+ * @brief NNStreamer package name.
+ */
+#define NNS_PKG "org/nnsuite/nnstreamer"
```
dongju-chae, line 49: BTW, are we still this organization, `nnsuite`?

jaeyun-jung: yes, in android (jcenter) we set the group name 'org.nnsuite'
    userOrg = 'nnsuite'
    repoName = 'nnstreamer'
    groupId = 'org.nnsuite'

PR #2399, opened 2020-05-22
[Unittest] Fix a potential bug of unit testcase

again4you:
This patch fixes the potential bug (i.e. pipeline parsing error) of unit testcase when using Protobuf v3.8.

**Self evaluation:**


myungjoo:
We have decided to downgrade protobuf until such issues are resolved and upgrading tensorflow is postponed.

Both pytorch and tensorflow really need to wake up and use system libraries!!!

PR #2406, opened 2020-05-29
[Filter] release mem block when execution failed @open sesame 5/29 14:35

jaeyun-jung:
release in/out mem blocks when failed to invoke nn framework.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -346,7 +346,7 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
   GstMapInfo out_info[NNS_TENSOR_SIZE_LIMIT];
   GstTensorMemory in_tensors[NNS_TENSOR_SIZE_LIMIT];
   GstTensorMemory out_tensors[NNS_TENSOR_SIZE_LIMIT];
-  guint i;
+  guint i, j;
   gint ret;
   gboolean allocate_in_invoke;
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -420,40 +418,39 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
...
+      if (ret == 0) {
+        GPtrArray *data_array = g_ptr_array_new ();
+        g_ptr_array_add (data_array, (gpointer) self);
+        g_ptr_array_add (data_array, (gpointer) out_tensors[i].data);
+
+        /* filter-subplugin allocated new memory, update this */
+        out_mem[i] = gst_memory_new_wrapped (0, out_tensors[i].data,
+            out_tensors[i].size, 0, out_tensors[i].size, (gpointer) data_array,
+            gst_tensor_filter_destroy_notify);
+      }
```
zhoonit, line 434: Q. Is this fine to not handle the case `allocation_in_invoke && ret != 0`?

jaeyun-jung: ret < 0 means invoke-failure, so it is unnecessary to wrap the output data.
if ret > 0, tensor-filter will drop the incoming buffer.

zhoonit: Thanks for the explanation :)

PR #2410, opened 2020-06-02
[Build/Tests] Fix 'tests/meson.build' to always find 'copy' program @open sesame 6/2 12:41

dongju-chae:
This patch fixes minor build errors caused when `libpng_dep` does not exist.
`tests/nnstreamer_filter_extensions_common/meson.build` is using this
variable regardless of `libpng_dep`'s existence.

**Self evaluation:**

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -14,6 +14,7 @@ library('nnscustom_drop_buffer',
 )
 
 # Build and copy exe for ssat
+copy = find_program('cp', required: true)
 libpng_dep = dependency('libpng', required: false)
 if libpng_dep.found()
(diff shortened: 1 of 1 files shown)


tests/meson.build
```diff
@@ -14,6 +14,7 @@ library('nnscustom_drop_buffer',
 )
 
 # Build and copy exe for ssat
+copy = find_program('cp')
```
wooksong, line 17: If this is mandatory to build the test, 
```suggestion
copy = find_program('cp', required: true)
```

dongju-chae: Looks good.

PR #2415, opened 2020-06-02
[Decoder/Neon] Support neon accl for SNPE deeplab/depth decoder @open sesame 6/5 13:19

dongju-chae:
This patch supports neon accl for SNPE deeplab/depth segmentation decoder

I think we can optimize more as it's a naive impl. 
In fact, setting alpha values (conditionally) makes some inefficiency to use neon instructions.

**Self evaluation:**


dongju-chae:
@anyj0527 Please test with this PR. In my env, it seems to work well.

PR #2419, opened 2020-06-03
[C-Api] function to get pipeline element

jaeyun-jung:
Add new function to get the gst-element from pipeline handle.
For native developers, it is necessary to get pipeline element to use gstreamer functions.

### DIFF
--- a/api/capi/include/nnstreamer-capi-private.h
+++ b/api/capi/include/nnstreamer-capi-private.h
@@ -351,6 +351,14 @@ char* ml_nnfw_to_str_prop (ml_nnfw_hw_e hw);
  */
 const char* ml_get_nnfw_subplugin_name (ml_nnfw_type_e nnfw);
 
+/**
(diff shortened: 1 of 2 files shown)


api/capi/include/nnstreamer-capi-private.h
```diff
@@ -351,6 +351,15 @@ char* ml_nnfw_to_str_prop (ml_nnfw_hw_e hw);
  */
 const char* ml_get_nnfw_subplugin_name (ml_nnfw_type_e nnfw);
 
+/**
+ * @brief Gets the element of pipeline itself (GstElement).
+ * @details With the returned reference, you can use GStreamer functions to handle the element in pipeline.
+ *          Note that caller should release the returned reference using gst_object_unref().
+ * @since_tizen 6.0
```
myungjoo, line 358: This is internal private API. do not add "@since_tizen".

jaeyun-jung: updated

PR #2422, opened 2020-06-03
[Filter/Prop] Add performance properties (lat/fps) for tensor filter

### LINKED ISSUE #2417: Tensor-Filter Property Request (performance info)
## Property "latency"

set: 0: off / 1: on / (we may add other values later. e.g., 2: average 1 sec, 3: average total, ...)
get: -1: INVAL / 0 or higher: average latency of recent 10 inferences in microseconds

## Property "throughput"

set: 0: off / 1: on / (we may add other values later. e.g., 2: average 1 sec, 3: average total, ...)
get: -1: INVAL / 0 or higher: average throughput in "number of outputs per second"

This is required by the demo.

CC: @anyj0527 
in a pipeline, retrieve the object (GstElement) with name (e.g., .. ! tensor_filter name=name1 ... ! ...),
use get/set-property for the GstElement to get the performance values.
then, apply appropriate options and strings for textoverlay to show performance of the two models independently.

dongju-chae:
This patch adds performance properties (latency/throughput) for tensor filter.

Those properties will be used for the overlay information, to show the performance
of tensor filter (e.g., SNPE).

It resolves #2417 

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -331,6 +331,101 @@ gst_tensor_filter_destroy_notify (void *data)
 }
 
 /**
+ * @brief Prepare statistics for performance profiling (e.g, latency, throughput)
+ */
+static void
+prepare_statistics (GstTensorFilterPrivate * priv)
+{
+  priv->stat.latest_invoke_time = g_get_real_time ();
+}
+
+/**
+ * @brief Helper function to accumulate latencies
+ */
+static void
+accumulate_latency (void *data, void *user_data)
+{
+  gint64 *latency = data;
+  gint64 *total_latency = user_data;
+
+  *total_latency += *latency;
+}
+
+#define THRESHOLD_DROP_OLD  (2000)
+#define THRESHOLD_CACHE_OLD (1000)
+
+/**
+ * @brief Record statistics for performance profiling (e.g, latency, throughput)
+ */
+static void
+record_statistics (GstTensorFilterPrivate * priv)
+{
+  gint64 end_time = g_get_real_time ();
+  gint64 *latency = g_new (gint64, 1);
+  GQueue *recent_latencies = priv->stat.recent_latencies;
+
+  *latency = end_time - priv->stat.latest_invoke_time;
+  priv->stat.total_invoke_latency += *latency;
+  priv->stat.total_invoke_num += 1;
+
+  if (g_queue_get_length (recent_latencies) == GST_TF_STAT_MAX_RECENT)
+    g_free (g_queue_pop_head (recent_latencies));
+  g_queue_push_tail (recent_latencies, latency);
+
+  /* the queue should have at least one element */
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -330,6 +330,78 @@ gst_tensor_filter_destroy_notify (void *data)
...
+
+  if (priv->throughput_mode > 0) {
+    gint throughput_int = -1;
+
+    if (priv->stat.total_invoke_latency != 0) {
+      gdouble throughput =
+          (gdouble) (priv->stat.total_invoke_num * G_USEC_PER_SEC);
+
+      throughput /= priv->stat.total_invoke_latency;
+      throughput *= 1000;
```
myungjoo, line 395: For the sake of accuracy, do:
```
throughput = (num * USEC_PER_SEC * 1000.0) / total_latency;
```

myungjoo: Plus, throughput-related variables are monotonously increasing; thus you are averaging the total execution, which worries me with overflows.

myungjoo: BUG: if latency = 0 and througput = 1, you may do divide-by-zero.

dongju-chae: > For the sake of accuracy, do: throughput = (num * USEC_PER_SEC * 1000.0) / total_latency;

Updated, thanks.

> Plus, throughput-related variables are monotonously increasing; thus you are averaging the total execution, which worries me with overflows.

You're right! I changed codes to cache old values if the value exceeds some threshold. But, I'm not sure it's a good way.

> BUG: if latency = 0 and througput = 1, you may do divide-by-zero. and total_invoke_latency is incorrect.

Hmm. I think I don't understand your concerns. Even if `latency_mode = 0` and `throuput_mode = 1`, `total_invoke_latency` is updated. And... the if-statement may prevent divide-by-zero.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -719,6 +751,16 @@ gst_tensor_filter_common_free_property (GstTensorFilterPrivate * priv)
 
   gst_tensors_info_free (&priv->in_config.info);
   gst_tensors_info_free (&priv->out_config.info);
+
+  if (priv->stat.recent_latencies != NULL) {
+    GQueue *queue = priv->stat.recent_latencies;
+    gint64 *latency;
+    for (latency = g_queue_pop_tail (queue); latency != NULL;
+        latency = g_queue_pop_tail (queue)) {
+      g_free (latency);
```
again4you, line 760: To make a code more simple, what about using `while` statement as below?
```cpp
while (latency = g_queue_pop_tail (queue)) {
    g_free (latency);
}
```

dongju-chae: Right! thanks

PR #2426, opened 2020-06-03
[TEST] Test flatbuf and protobuf integration @open sesame 6/4 17:33

gichan-jang:
*Note: PR for test in CI
The unit test fails in my Ubuntu machine, but it passes in the gbs build.
This PR is intended to test whether the test is performed normally in CI.
I will close PR after test.

2. Run test: [ ]Passed [ *]Failed [ ]Skipped


myungjoo:
get_subplugin for GETALL mode has a bug. You will need to fix it to make it work.

gichan-jang:
> get_subplugin for GETALL mode has a bug. You will need to fix it to make it work.

Right. There is conflict in the namespace when flatbuf and protobuf loaded.
I re-defined namespace from `NNStreamer` to `nnstreamer::flatbuf` and `nnstreamer::protobuf`.

gichan-jang:
Flatbuf and protobuf PR are uploaded seperately. Close this PR.

PR #2427, opened 2020-06-03
[Filter/TFLite] remove ineffective codes

helloahn:
# Summary
setting the allocation type is not effective anymore.

# Description
previously(at Tensorflow v1.9), the flag `kTfLiteDynamic` for the allocation option of each tensor of Tensorflow-Lite was required. if it was not, there was no way to set input raw data and to get output raw data without `memcpy`.

However, recently, we met a case about a TFLite model not working with the option `kTfLiteDynamic`, and I tried various ways to solve it.

In conclusion, at TF 1.13.1 and 1.15.2 we provide, the option is not effective anymore. In other words, the raw data pointer that we used to still works properly without any problem like memleak or defects.

For this reason, we can solve the problem(the model not working) by removing the setting phrase without any modification of the current policy(not using the `memcpy`).

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -225,12 +225,6 @@ TFLiteInterpreter::invoke (const GstTensorMemory * input,
     status = kTfLiteError;
   }
 
-  /** if it is not `nullptr`, tensorflow makes `free()` the memory itself. */
-  int tensorSize = tensors_idx.size ();
-  for (int i = 0; i < tensorSize; ++i) {
-    interpreter->tensor (tensors_idx[i])->data.raw = nullptr;
-  }
-
 #if (DBG)
   gint64 stop_time = g_get_real_time ();
   g_message ("Invoke() is finished: %" G_GINT64_FORMAT,
@@ -284,22 +278,6 @@ TFLiteInterpreter::loadModel (bool use_nnapi)
     }
   }
 #endif
-
-  /** set allocation type to dynamic for in/out tensors */
-  int tensor_idx;
-
-  int tensorSize = interpreter->inputs ().size ();
-  for (int i = 0; i < tensorSize; ++i) {
-    tensor_idx = interpreter->inputs ()[i];
-    interpreter->tensor (tensor_idx)->allocation_type = kTfLiteDynamic;
-  }
-
-  tensorSize = interpreter->outputs ().size ();
-  for (int i = 0; i < tensorSize; ++i) {
-    tensor_idx = interpreter->outputs ()[i];
-    interpreter->tensor (tensor_idx)->allocation_type = kTfLiteDynamic;
-  }
-
   if (interpreter->AllocateTensors () != kTfLiteOk) {
     ml_loge ("Failed to allocate tensors\n");
     return -2;


myungjoo:
> setting allocation type is not effective anymore.

please elaborate..

helloahn:
> please elaborate..
> 
> If this is for the tf-lite model that cannot support memcpy-less output buffers, it should be activated optionally (by user configurable properties)

Sorry for the short description. I added the background.

myungjoo:
If the fix works from TF-Lite 1.13, add if or #ifdef that checks the tf-lite version. (I understand the complication of TF-Lite version checks, but there should be some workaround for that, too :smile: )

You don't know when they will require for backporting nnstreamer for older Tizen; we need to keep the backward compatibility.

helloahn:
> If the fix works from TF-Lite 1.13, add if or #ifdef that checks the tf-lite version. (I understand the complication of TF-Lite version checks, but there should be some workaround for that, too 😄 )
> 
> You don't know when they will require for backporting nnstreamer for older Tizen; we need to keep the backward compatibility.

I agree. it will be applied

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -225,11 +225,13 @@ TFLiteInterpreter::invoke (const GstTensorMemory * input,
     status = kTfLiteError;
   }
 
+#if TFLITE_VERSION_MAJOR==1 && TFLITE_VERSION_MINOR==9
```
helloahn, line 228: there is another option like protobuf.
they manage the version like this 3.6.1 == 3006001.
if it looks more standardlike, please leave your opinion.

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -96,6 +96,11 @@ if tflite_support_is_available
 
   tflite_compile_args = []
 
+  tflite_versions = tflite_support_version.split('.')	
+  tflite_compile_args += '-DTFLITE_VERSION_MAJOR='+tflite_versions[0]
+  tflite_compile_args += '-DTFLITE_VERSION_MINOR='+tflite_versions[1]
+  tflite_compile_args += '-DTFLITE_VERSION_PATCH='+tflite_versions[2]
+
```
myungjoo, line 103: You need to do something similar for Android build scripts.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
myungjoo: We didn't check it, but let's say MINOR <= 9 here.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -285,6 +287,7 @@ TFLiteInterpreter::loadModel (bool use_nnapi)
   }
 #endif
 
+#if TFLITE_VERSION_MAJOR==1 && TFLITE_VERSION_MINOR==9
```
myungjoo, line 290: <=9 here as well.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
jaeyun-jung: need to check this with tf-lite 1.9.
IMO we dont need this.

helloahn:
@myungjoo 
with @jaeyun-jung, we checked TFLite 1.9 without the code block, and.....
embarrassingly, there was no problem.
this means the code block was never needed.

So, we need no allocation option and no version check anymore. Just removing can solve it.

myungjoo:
> @myungjoo
> with @jaeyun-jung, we checked TFLite 1.9 without the code block, and.....
> embarrassingly, there was no problem.
> this means the code block was never needed.
> 
> So, we need no allocation option and no version check anymore. Just removing can solve it.

image

PR #2431, opened 2020-06-04
[NNFW] set backend after loading model

jaeyun-jung:
recently updated, nnfw set-backend should be called after loading the model.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -203,14 +203,6 @@ nnfw_open (const GstTensorFilterProperties * prop, void **private_data)
     goto unalloc_exit;
   }
 
-  accelerator = nnfw_get_accelerator (pdata, prop->accl_str);
-  status = nnfw_set_available_backends (pdata->session, accelerator);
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -238,6 +230,14 @@ nnfw_open (const GstTensorFilterProperties * prop, void **private_data)
     goto session_exit;
   }
 
+  accelerator = nnfw_get_accelerator (pdata, prop->accl_str);
+  status = nnfw_set_available_backends (pdata->session, accelerator);
+  if (status != NNFW_STATUS_NO_ERROR) {
+    err = -EINVAL;
+    g_printerr ("Cannot set nnfw-runtime backend to %s\n", accelerator);
+    goto unalloc_exit;
```
again4you, line 238: IMO, it should be `session_exit` instead of `unalloc_exit` since the session is already created by calling `nnfw_create_session()`.

myungjoo: Yes. that has been a bug. It should've been session_exit even before this commit.

jaeyun-jung: updated

Issue #2440, opened 2020-06-09
After passing mux / demux, framerate is extreamly bigger than before


again4you:
## Issue Description

In my opinion, there is a bug in `tensor_mux`.

First, I checked the framerate of below pipeline. I set 'videorate max-rate=1' so its framerate (i.e. fps) should be 1.

```bash
$ GST_DEBUG="GST_TRACER:7" GST_TRACERS="framerate" \
gst-launch-1.0 videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! videorate max-rate=1 ! \
tensor_converter  ! tensor_sink sync=true
...
0:00:05.405294672 14566       0x86ca70 TRACE             GST_TRACER :0:: framerate, pad=(string)videorate0_src, fps=(uint)1;
0:00:05.405351689 14566       0x86ca70 TRACE             GST_TRACER :0:: framerate, pad=(string)videotestsrc0_src, fps=(uint)1;
0:00:05.405368139 14566       0x86ca70 TRACE             GST_TRACER :0:: framerate, pad=(string)capsfilter0_src, fps=(uint)1;
0:00:05.405382281 14566       0x86ca70 TRACE             GST_TRACER :0:: framerate, pad=(string)tensorconverter0_src, fps=(uint)1;
```

However, the framerate is increased after simply adding `tensor_mux ! tensor_demux` between `tensor_converter` and `tensor_sink` as below.

```bash
$ GST_DEBUG="GST_TRACER:7" GST_TRACERS="framerate" \
gst-launch-1.0 videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! videorate max-rate=1 ! \
tensor_converter ! tensor_mux ! tensor_demux ! tensor_sink sync=true
...
0:00:02.582611050 14657      0x1b1a2a0 TRACE             GST_TRACER :0:: framerate, pad=(string)videotestsrc0_src, fps=(uint)2314;
0:00:02.582629105 14657      0x1b1a2a0 TRACE             GST_TRACER :0:: framerate, pad=(string)tensorconverter0_src, fps=(uint)2314;
0:00:02.582633981 14657      0x1b1a2a0 TRACE             GST_TRACER :0:: framerate, pad=(string)tensormux0_src, fps=(uint)2314;
0:00:02.582638150 14657      0x1b1a2a0 TRACE             GST_TRACER :0:: framerate, pad=(string)capsfilter0_src, fps=(uint)2314;
[...]

PR #2441, opened 2020-06-09
[Mux] Set sync-mode to slowest as default @open sesame 6/9 23:16

again4you:
After passing mux / demux, the framerate is extremely bigger than
before. (Issue #2440) This is mainly because sync_mode of tensor_mux
filter is SYNC_NOSYNC. To fix this bug, this patch set sync_mode to
SYNC_SLOWEST as default.

* Related Issue: #2440 

**Self evaluation:**

**Result**
After applying this patch, Default sync-mode is `slowest`.
```bash
$ gst-inspect-1.0 --gst-plugin-path="build" tensor_mux
...
  sync-mode           : Time synchronization mode?
                        flags: readable, writable
                        String. Default: "slowest"
...
```

### DIFF
--- a/gst/nnstreamer/tensor_mux/gsttensormux.c
+++ b/gst/nnstreamer/tensor_mux/gsttensormux.c
@@ -210,7 +210,7 @@ gst_tensor_mux_init (GstTensorMux * tensor_mux)
       tensor_mux);
 
   tensor_mux->silent = TRUE;
-  tensor_mux->sync.mode = SYNC_NOSYNC;
+  tensor_mux->sync.mode = SYNC_SLOWEST;
   tensor_mux->sync.option = NULL;
   tensor_mux->need_buffer = FALSE;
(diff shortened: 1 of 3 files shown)


myungjoo:
It appears that slowest has a bug that incurs infinite loops with a test case (or the test case has a bug that waits indefinitely with slowest mode)

again4you:
The main reason of CI hanging is that `nnstreamer_repo_lstm` test case is not finished when enabling default SYNC_SLOWEST mode.

If I add `sync_mode=nosync` option in nnstreamer_repo_lstm test case, this test case is passed. However, I don't know whether this approach is right or not.

@jijoongmoon, PTAL and give your opinion.

again4you:
### Submit v2
* Use 'nosync' option in LSTM & RNN test since tensor_repo needs that option (verified by Jijoong)

PR #2442, opened 2020-06-09
[Android] build static libs / register native functions

jaeyun-jung:
1. make all sub-plugins static libs and build one native so file.
2. register native functions when loading native lib.

tested with @helloahn
@anyj0527 please check this PR for example app.


api/android/api/src/main/jni/Android.mk
```diff
@@ -126,7 +132,7 @@ include $(LOCAL_PATH)/Android-gst-plugins.mk
 
 GSTREAMER_PLUGINS        := $(GST_REQUIRED_PLUGINS)
 GSTREAMER_EXTRA_DEPS     := $(GST_REQUIRED_DEPS) gio-2.0 gmodule-2.0 json-glib-1.0
-GSTREAMER_EXTRA_LIBS     := $(GST_REQUIRED_LIBS) -liconv
+GSTREAMER_EXTRA_LIBS     := $(GST_REQUIRED_LIBS) -liconv -lcairo
```
myungjoo, line 135: Is this safe? (-lcairo)

jaeyun-jung: With build option, set '-lcairo' (build type all).
Also I updated '-lmediandk' when including amcsrc element.

anyj0527:
@jaeyun-jung Checked. Thanks!

PR #2448, opened 2020-06-10
[Android] do not convert meta in invoke method

jaeyun-jung:
To improve performance, it is necessary to remove metadata conversion in invoke method.
Do not handle dynamic mode in invoke method and update description to change the tensor info.
Also, if output info is changed using set-info method, update cached instance in private data.


myungjoo:
Please update commit title. You don't disable it. You require "set-info" to call it dynamically.

jaeyun-jung:
2020-06-10 commit message updated

api/android/api/src/main/jni/nnstreamer-native-pipeline.c
```diff
@@ -95,15 +96,11 @@ nns_pipeline_sink_priv_set_out_info (element_data_s * item, JNIEnv * env,
     return FALSE;
   }
 
+  ml_tensors_info_free (priv->out_info);
+  ml_tensors_info_clone (priv->out_info, out_info);
+
```
helloahn, line 101: looks better than previous

PR #2452, opened 2020-06-10
[Filter/Custom] copy metadata in registration

jaeyun-jung:
In the filter custom-easy, prevent error case when developer tries to register custom-filter with local variables.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
@@ -38,8 +38,8 @@ void fini_filter_custom_easy (void) __attribute__ ((destructor));
 typedef struct _internal_data
 {
   NNS_custom_invoke func;
-  const GstTensorsInfo *in_info;
-  const GstTensorsInfo *out_info;
+  GstTensorsInfo in_info;
+  GstTensorsInfo out_info;
   void *data; /**< The easy-filter writer's data */
 } internal_data;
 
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
```diff
@@ -66,15 +66,19 @@ NNS_custom_easy_register (const char *modelname,
...
 
   if (!ptr)
     return -ENOMEM;
 
   ptr->func = func;
   ptr->data = data;
-  ptr->in_info = in_info;
-  ptr->out_info = out_info;
+  gst_tensors_info_copy (&ptr->in_info, in_info);
+  gst_tensors_info_copy (&ptr->out_info, out_info);
```
dongju-chae, line 81: I think you need to call `gst_tensors_info_free` as well in error handling (e.g., before `g_free (ptr)` in line 86).

gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
```diff
@@ -66,15 +66,19 @@ NNS_custom_easy_register (const char *modelname,
   if (!func || !in_info || !out_info)
     return -EINVAL;
 
+  if (!gst_tensors_info_validate (in_info) ||
```
gichan-jang, line 69: need to pass the pointer of the info.

jaeyun-jung: in_info is a pointer :)

gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
jaeyun-jung: You are right. I added a function to free internal data.
When unregistering custom filter, it is also necessary to free internal data.

gst/nnstreamer/tensor_filter/tensor_filter_custom_easy.c
gichan-jang: Ah, I confused with line 41. Thanks.

PR #2457, opened 2020-06-11
[Filter/EdgeTPU] Add a 'dummy' device mode for the test purpose @open sesame 6/15 11:46

wooksong:
This patch adds a 'dummy' device mode for the test purpose instead of using the dummy shared library that pretends to be libedgetpu.so.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
@@ -17,8 +17,9 @@
  * @todo A lot of this duplicate tf-lite filter.
  *       We may be able to embed this code into tf-lite filter code.
  */
-#include <iostream>
 #include <fstream>
+#include <iostream>
+#include <sstream>
 #include <string>
 #include <stdexcept>
 
@@ -50,10 +51,34 @@ namespace tensorfilter_edgetpu {
 void _init_filter_edgetpu (void) __attribute__ ((constructor));
 void _fini_filter_edgetpu (void) __attribute__ ((destructor));
 
+
+enum class edgetpu_subplugin_device_type : uint32_t {
+  USB = static_cast<uint32_t>(edgetpu::DeviceType::kApexUsb),
+  PCI = static_cast<uint32_t>(edgetpu::DeviceType::kApexPci),
+  DEFAULT = USB,
+  DUMMY = 99,
+};
+
+static const std::string edgetpu_subplugin_device_type_name (
+    edgetpu_subplugin_device_type t)
+{
+    switch(t)
+    {
+        case edgetpu_subplugin_device_type::PCI:
+          return "pci";
+        case edgetpu_subplugin_device_type::DUMMY:
+          return "dummy";
+        case edgetpu_subplugin_device_type::USB:
+        default:
+          return "usb";
+    }
+}
+
 class edgetpu_subplugin final : public tensor_filter_subplugin {
 private:
   bool empty_model;
   char *model_path; /**< The model *.tflite file */
+  edgetpu_subplugin_device_type device_type; /**< The device type of Edge TPU */
   GstTensorsInfo inputInfo; /**< Input tensors metadata */
   GstTensorsInfo outputInfo; /**< Output tensors metadata */
 
@@ -68,7 +93,8 @@ private:
       /**< Loaded TF Lite model (from model_path) */
   static std::unique_ptr<tflite::Interpreter>
       BuildEdgeTpuInterpreter(const tflite::FlatBufferModel &model,
-          edgetpu::EdgeTpuContext* edgetpu_context);
+          const edgetpu_subplugin_device_type dev_type,
+          edgetpu::EdgeTpuContext* edgetpu_context = nullptr);
 
   /** Internal Utility Functions & Properties ***************************/
   void cleanup ();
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
```diff
@@ -43,10 +44,33 @@ namespace tensorfilter_edgetpu {
...
+{
+    switch(t)
+    {
+        case edgetpu_subplugin_device_type::PCI:
+          return "pci";
+        case edgetpu_subplugin_device_type::DUMMY:
+          return "dummy";
+        case edgetpu_subplugin_device_type::USB:
+        default:
+          return "usb";
```
dongju-chae, line 65: IMO, it's natural to return `dummy` in default case. Is there any implied purpose on this?

wooksong: @dongju-chae This is a sub-plugin for EdgeTPU so that it is natural to work with the device by default for me. The "dummy" mode is only for the testing.

gichan-jang:
> :octocat: **cibot**: @wooksong, A builder checker could not be completed because one of the checkers is not completed. In order to find out a reason, please go to http://nnsuite.mooo.com/nnstreamer/ci/repo-workers/pr-checker/2457-202006151147010.14737606048584-0189285f7ed698c07172006e2477d05f8b397977/.

Since tensorflow-lite-dev pkg is changed, you need to rebase. (#2454 )

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
dongju-chae: Got it

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
```diff
@@ -172,19 +262,37 @@ void edgetpu_subplugin::configure_instance (const GstTensorFilterProperties *pro
...
+      cleanup();
+      throw std::system_error (ENODEV, std::system_category(),
+          "Edge-TPU device is opened, but cannot get its interpreter.");
+    }
+  } else {
+    /* If the device_type is 'dummy', work same as tflite using CPU */
+    tflite::ops::builtin::BuiltinOpResolver resolver;
+
+    edgetpu_context = nullptr;
+    model_interpreter = nullptr;
```
dongju-chae, line 288: I think this line is unnecessary.

wooksong:
> > :octocat: **cibot**: @wooksong, A builder checker could not be completed because one of the checkers is not completed. In order to find out a reason, please go to http://nnsuite.mooo.com/nnstreamer/ci/repo-workers/pr-checker/2457-202006151147010.14737606048584-0189285f7ed698c07172006e2477d05f8b397977/.
> 
> Since tensorflow-lite-dev pkg is changed, you need to rebase. (#2454 )
@gichan-jang Thanks.

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
wooksong: @dongju-chae Right. Removed. Thanks.

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
```diff
@@ -144,10 +173,71 @@ tensor_filter_subplugin & edgetpu_subplugin::getEmptyInstance ()
...
+ * @param custom_prop The given c_str value of the 'custom' property
+ * @return A edgetpu_subplugin_device_type value for the given 'device_type'
+ *        if the value of the 'custom' property. Otherwise,
+ *        edgetpu_subplugin_device_type::USB, which is default, is returned.
+ */
+edgetpu_subplugin_device_type edgetpu_subplugin::parse_custom_prop (
+    const char *custom_prop)
+{
+  if ((!custom_prop) || (strlen (custom_prop) == 0))
+    return edgetpu_subplugin_device_type::USB;
```
kparichay, line 199: minor comment - how about declaring a default device type which can be used here as well as at the constructor of `edgetpu_subplugin`.

wooksong: @kparichay I guess I do not fully understand your comment, but an enum value indicating a DEFAULT device type is added and the constructor and this function are modified to use this enum as well.

However, IMHO, USB is more intuitive than DEFAULT to me.

kparichay: My only concern is that if the default device is updated in future, it would require changes in many locations.
If USB looks more intuitive, then please keep the USB.

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
```diff
@@ -151,10 +181,71 @@ tensor_filter_subplugin & edgetpu_subplugin::getEmptyInstance ()
...
+  cprop_ss = std::stringstream (cprop.substr (0, pos));
+
+  while (std::getline (cprop_ss, token, ':')) {
+    vec.push_back (token);
+    if (vec.size () > max_vec_len)
+      break;
+  }
+
+  if (edgetpu_subplugin::str_tolower (vec[0]) != key)
+    return edgetpu_subplugin_device_type::USB;
```
kparichay, line 227: If DEFAULT is being used, then this and at line 240 would also be DEFAULT.

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
wooksong: @kparichay I have replaced USB, which is used as a default device, with DEFAULT.

ext/nnstreamer/tensor_filter/tensor_filter_edgetpu.cc
wooksong: Right. Thanks.

PR #2462, opened 2020-06-12
[Converter] reduce complexity

jaeyun-jung:
code clean to reduce complexity
1. internal function to push multiple buffers.
2. remove unnecessary args


jaeyun-jung:
2020-06-15 PR updated (external converter with media type)

PR #2472, opened 2020-06-16
[C-Api] error code when failed to allocate tensor data

jaeyun-jung:
Remove unnecessary error code (stream-pipe).
If failed to allocate mem for tensors-data handle, API returns out-of-mem error.

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -621,9 +621,6 @@ ml_tensors_data_create (const ml_tensors_info_h info, ml_tensors_data_h * data)
   if (status != ML_ERROR_NONE) {
     return status;
   }
-  if (!_data) {
-    return ML_ERROR_STREAMS_PIPE;
-  }
 
   for (i = 0; i < _data->num_tensors; i++) {
     _data->tensors[i].tensor = g_malloc0 (_data->tensors[i].size);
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -621,9 +621,6 @@ ml_tensors_data_create (const ml_tensors_info_h info, ml_tensors_data_h * data)
   if (status != ML_ERROR_NONE) {
     return status;
   }
-  if (!_data) {
-    return ML_ERROR_STREAMS_PIPE;
-  }
 
   for (i = 0; i < _data->num_tensors; i++) {
     _data->tensors[i].tensor = g_malloc0 (_data->tensors[i].size);
```
dongju-chae, line 626: I think `g_malloc0` cannot fail. If you wanna handle failure cases, how about using `g_try_malloc0`? Or, you may remove `ML_ERROR_OUT_OF_MEMORY` as well. In fact, size 0 can return NULL, but I think it's not out-of-memory but close to invalid arguments.

jaeyun-jung: Well, internally most APIs and functions in nnstreamer does not use g_try_malloc() now.
As you said, g_malloc internally uses g_error() which aborts the process - always fatal - if failed to allocate mem.
How about adding another PR for changing all allocation (g_try_malloc0 ot g_try_new0)?

dongju-chae: Yes, my comment was not related to this PR :)

PR #2474, opened 2020-06-17
[Android] custom-easy filter @open sesame 7/7 15:16

jaeyun-jung:
Update custom-filter API using custom-easy.
- register custom-filter with in/out info and implement invoke interface only.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
@@ -54,77 +54,68 @@ public class APITestCustomFilter {
 
     private void registerCustomFilters() {
         try {
+            TensorsInfo inputInfo = new TensorsInfo();
+            inputInfo.addTensorInfo(NNStreamer.TensorType.INT32, new int[]{10});
+
+            TensorsInfo outputInfo = inputInfo.clone();
+
             /* register custom-filter (passthrough) */
             mCustomPassthrough = CustomFilter.registerCustomFilter("custom-passthrough",
-                new CustomFilter.CustomFilterCallback() {
-                    @Override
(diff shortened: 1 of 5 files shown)


api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCustomFilter.java
```diff
@@ -286,18 +277,58 @@ public TensorsData invoke(TensorsData in) {
...
+                }
+            });
 
-                    @Override
-                    public TensorsData invoke(TensorsData in) {
-                        return in;
-                    }
+            fail();
+        } catch (Exception e) {
+            /* expected */
```
helloahn, line 296: does it mean the expected exception?

jaeyun-jung: yes, right.

api/android/api/src/main/java/org/nnsuite/nnstreamer/CustomFilter.java
```diff
@@ -65,20 +52,24 @@
...
      *
      * @return {@link CustomFilter} instance
      *
      * @throws IllegalArgumentException if given param is null
      * @throws IllegalStateException if failed to initialize custom-filter
      */
-    public static CustomFilter registerCustomFilter(@NonNull String name, @NonNull CustomFilterCallback callback) {
-        return new CustomFilter(name, callback);
+    public static CustomFilter registerCustomFilter(@NonNull String name, @NonNull TensorsInfo in,
+            @NonNull TensorsInfo out, @NonNull CustomFilterCallback callback) {
```
again4you, line 71: If Android API do not follow the backward compatibility, then it's OK.

myungjoo: It should be ok for now as we do not have corresponding applications, yet.

PR #2479, opened 2020-06-18
[Fix/MemLeak] Prevent potential memory leak in tensor info copy @open sesame 6/19 15:30

dongju-chae:
This patch prevents potential memory leaks in tensor info copy

**Self evaluation:**


dongju-chae:
Let's close this PR because currently there's too many cases copying tensor info without tensor info initialization.
This may result in undesirable side-effects due to uninitialized `name`.

PR #2496, opened 2020-06-26
[Android/Doc] guide to get gstreamer

jaeyun-jung:
update gstreamer version to 1.17.1.
change guide document for gstreamer prebuilt libs.

### DIFF
--- a/api/android/README.md
+++ b/api/android/README.md
@@ -9,7 +9,7 @@ We assume that you already have experienced Android application developments wit
    * Android Studio: Ubuntu version
    * Android SDK: Min version 24 (Nougat)
    * Android NDK: Use default ndk-bundle in Android Studio
-   * GStreamer: gstreamer-1.0-android-universal-1.16.2
+   * GStreamer: gstreamer-1.0-android-universal-1.17.1
 
 ## Build library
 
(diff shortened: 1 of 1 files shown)


myungjoo:
Do you have a reason to change the version of GST?

It's ok to let users change GST versions freely, but changing GST versions of our official releases (it affects products) should be more careful. Note that 1.16.2 is used by Tizen 6.x as well (we can cross-check issues from Tizen to Android products if something happens).

api/android/README.md
```diff
@@ -9,7 +9,7 @@ We assume that you already have experienced Android application developments wit
    * Android Studio: Ubuntu version
    * Android SDK: Min version 24 (Nougat)
    * Android NDK: Use default ndk-bundle in Android Studio
-   * GStreamer: gstreamer-1.0-android-universal-1.16.2
+   * GStreamer: gstreamer-1.0-android-universal-1.17.1
```
myungjoo, line 12: Or.. as this is just a document, let's say:
```
* GStreamer: gstreamer-1.0-android-universal-1.17.1 (you may use other versions)
```

jaeyun-jung:
Let's wait next version (1.18)

PR #2500, opened 2020-06-30
[Tools] GstPipe<-->Pbtxt Conversion / Skeleton

myungjoo:
Skeleton code for GstPipe<-->Pbtxt Conversion.

Ultimately, we can use web-based WYSIWYG tools based on this!

### DIFF
--- a/tools/development/gstPrototxt.py
+++ b/tools/development/gstPrototxt.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+##
+# Prototxt <--> GStreamer pipeline converter
+# Copyright (c) 2020 Samsung Electronics
+##
+# @file   gstPrototxt.py
+# @brief  Prototxt to/from GStreamer pipeline converter
+# @author MyungJoo Ham <myungjoo.ham@samsung.com>
+# @date   30 Jun 2020
+# @bug    No known bugs
+# @todo   WIP
+#
+
+## @todo Get python pbtxt parser
+
+## @todo property link up tensor-repo-*
+
+declaredNames = [] # list of names declared by user.
+generatedNameCounter = 0
+GENERATEDNAMEPREFIX = '__id'
+
+## @brief A class for a filter (object == filter instance)
+class Filter:
+    element = None    # element name (string)
+    name = None       # name property (string)
+    nameGenerated = False   # boolean
+    properties = [] # list of (key, value)
+    src = []        # list of (Filter, string(name of src), string(src-padname. maybe None))
+    sink = []        # list of (Filter, string(name of sink), string(sink-padname. maybe None))
+
+    def __init__(self, elementname, _name):
+        global generatedNameCounter
+        global GENERATEDNAMEPREFIX
+        global declaredNames
+
+        self.element = elementname
+        if _name is not None and len(_name) > 0:
+            name = _name
+            if name in declaredNames:
+                raise RuntimeError("Duplicated name in elements: " + str(name))
+            declaredNames.append(str(name))
+        else:
+            while ((GENERATEDNAMEPREFIX + str(generatedNameCounter)) not in declaredNames):
+                generatedNameCounter = generatedNameCounter + 1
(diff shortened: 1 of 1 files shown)


tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
...
+        if name is None or len(name) == 0:
+            raise RuntimeError("Incorrect usage of __getFullPadname.")
+        if padname is None or len(padname) == 0:
+            return name + "."
+        else:
+            return name + "." + padname
+
+    ## @brief Verify filter and return object if filter is given with name
+    def __verifyfilter(f):
+        _a = a
```
dongju-chae, line 90: Where does `a` come from?

myungjoo: Yes.. it should've been f.

tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
+##
```
kparichay, line 2: for scripts, SPDX can be at max at line 2 (https://lore.kernel.org/patchwork/patch/883424/)

tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
...
+# @todo   WIP
+#
+
+## @todo Get python pbtxt parser
+
+## @todo property link up tensor-repo-*
+
+declaredNames = [] # list of names declared by user.
+generatedNameCounter = 0
+generatedNamePrefix = '__id'
```
kparichay, line 23: this is a global constant than global variable right? How about CAPITALIZING it?

tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
...
+
+        self.element = elementname
+        if _name is not None and len(_name) > 0:
+            name = _name
+            if name in declaredNames:
+                raise RuntimeError("Duplicated name in elements: " + str(name))
+            declaredNames.append(str(name))
+        else:
+            while ((generatedNamePrefix + str(generatedNameCounter)) not in declaredNames):
+                generatedNameCounter = generatedNameCounter + 1
```
kparichay, line 47: this increases the name counter for differently named variables. Is this intended?

tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
...
+            name = generatedNamePrefix + str(generatedNameCounter)
+            nameGenerated = True
```
kparichay, line 49: bring this outside the `if else` and add this name to `declaredNames`?

tools/development/gstPrototxt.py
```diff
@@ -0,0 +1,182 @@
...
+        self.sinks = []
+
+    ## @brief add a filter instnace
+    def addFilter(_filter):
+        self.filters.append(_filter)
+        self.nameList.append(_filter.name)
+
+    ## @brief Get filter with the given name
+    def __getFilterByName(name):
+        for f in self.filters:
```
kparichay, line 74: in python , its recommended to use `filter` than use a loop.
also, as nameList stores all the names of filters, checking directly in nameList would also suffice.

tools/development/gstPrototxt.py
myungjoo: Applied.

tools/development/gstPrototxt.py
myungjoo: Applied.

tools/development/gstPrototxt.py
myungjoo: No, the intention is to manage filters with its id declared (```filtername id=THENAME```) and filters without such (generated) differently. If it is generated (not declared), we can hide the id in GUI as well. If it is declared, we need to show the id in GUI.

tools/development/gstPrototxt.py
myungjoo: Updated.

dongju-chae (approved):
LGTM, after this PR is merged, I will take over this topic.

anyj0527 (approved):
LGTM 👍

PR #2503, opened 2020-07-02
[Filter/TF] code clean  @open sesame 7/3 11:17

jaeyun-jung:
change to static function to release buffer.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
@@ -95,6 +95,7 @@ private:
   tensor_type getTensorTypeFromTF (TF_DataType tfType);
   TF_DataType getTensorTypeToTF (tensor_type tType);
   int validateTensor (const GstTensorsInfo * tensorInfo, int is_input);
+  static void releaseBuffer (void* data, size_t t);
 };
 
 void init_filter_tf (void) __attribute__ ((constructor));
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
```diff
@@ -95,10 +95,13 @@ class TFCore
...
   TF_DataType getTensorTypeToTF (tensor_type tType);
   int validateTensor (const GstTensorsInfo * tensorInfo, int is_input);
+  static void releaseBuffer (void* data, size_t t);
 };
 
-void init_filter_tf (void) __attribute__ ((constructor));
-void fini_filter_tf (void) __attribute__ ((destructor));
+extern "C" {
+  void init_filter_tf (void) __attribute__ ((constructor));
+  void fini_filter_tf (void) __attribute__ ((destructor));
```
myungjoo, line 103: These two functions are defined in this .cc file; thus, the compiler will provide C++ call convention for them. Why do you need extern "C" for them?

jaeyun-jung: actually unnecessary now. I will revert these lines.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow.cc
```diff
@@ -192,8 +193,8 @@ TFCore::getModelPath ()
 /**
  * @brief	the definition of a deallocator method
  */
-static void
-DeallocateBuffer (void* data, size_t t) {
+void
+TFCore::releaseBuffer (void* data, size_t t) {
   std::free (data);
```
wooksong, line 198: This comment is not related to this PR, but how about using g_free here for the consistency of the allocator/deallocator pair.  In the near future, we can replace those g_file_test and g_file_get_contents functions with c++ styles.

PR #2506, opened 2020-07-03
[Tools] GstPipe String Parser

myungjoo:
Rather than rewriting lex/yacc code from scratch for python,
modify GST's lex/yacc (flex/bison) C code and call it
from python.

This is the first commit of parser and simply imports
GST-Parser code without touching it.


tools/development/parser/meson.build
```diff
@@ -0,0 +1,71 @@
+cc = meson.get_compiler('c')
```
anyj0527, line 1: Do not need to declare project?
```
$ meson build
The Meson build system
Version: 0.50.1
Source dir: /home/yongjoo/nnstreamer/tools/development/parser
Build dir: /home/yongjoo/nnstreamer/tools/development/parser/build
Build type: native build

ERROR: First statement must be a call to project
```

tools/development/parser/meson.build
```diff
@@ -0,0 +1,71 @@
...
+
+bison_cdata.set('BISON', bison.path())
+bison_cdata.set('BISON_ARGS', '')
+
+gen_grammar = configure_file(input : 'gen_grammar.py.in',
+  output : 'gen_grammar.py',
+  configuration : bison_cdata)
+
+# Custom targets
+parser = custom_target('parselex',
```
anyj0527, line 60: This could be from wrong version of meson, but may need `python3 = find_program('python3')` ?
```
meson.build:71:0: ERROR: Unknown variable "python3"
```

tools/development/parser/meson.build
myungjoo: In the end, it will be built along with all other "nnstreamer utilities".
Because the code conversion from gst-parse to nnstreamer-parse is not completed, it won't be able to be built anyway.

tools/development/parser/meson.build
myungjoo: That'd be done by a parent meson script.

tools/development/parser/grammar.y
```diff
@@ -0,0 +1,1223 @@
...
+
+/*** define YYPRINTF macro/function if we're debugging */
+
+/* bison 1.35 calls this macro with side effects, we need to make sure the
+   side effects work - crappy bison */
+
+#ifndef GST_DISABLE_GST_DEBUG
+#  define YYDEBUG 1
+
+#  ifdef G_HAVE_ISO_VARARGS
```
helloahn, line 83: Q) is it intended indentation like a code block?

myungjoo: That part is from GStreamer Parser without touching the original code.

helloahn (approved):
LGTM

dongju-chae (approved):
LGTM

PR #2512, opened 2020-07-07
[Tizen/Sensor] remove deprecated api

jaeyun-jung:
API to read sensor data was deprecated since 5.5.
Replace this API (sensor data list).
Also set block size of output buffer when starting the src element.

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -458,6 +458,9 @@ gst_tensor_src_tizensensor_init (GstTensorSrcTIZENSENSOR * self)
    */
   gst_base_src_set_async (GST_BASE_SRC (self), TRUE);
 
+  /** @todo TBD. Let's assume each frame has a fixed size */
+  gst_base_src_set_dynamic_size (GST_BASE_SRC (self), FALSE);
+
   if (NULL == tizensensors) {
     int i;
     tizensensors = g_hash_table_new (g_direct_hash, g_direct_equal);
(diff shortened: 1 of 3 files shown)


jaeyun-jung:
@anyj0527 plz see this changes and do test.

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -1206,24 +1212,29 @@ gst_tensor_src_tizensensor_fill (GstBaseSrc * src, guint64 offset,
   }
 
   if (self->mode == TZN_SENSOR_MODE_POLLING) {
-    sensor_event_s event;
+    sensor_event_s *event;
```
myungjoo, line 1215: ```s/*event/*events/```

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -881,6 +884,13 @@ gst_tensor_src_tizensensor_start (GstBaseSrc * src)
     retval = FALSE;
     goto exit;
   }
+
+  /** @todo TBD. Let's assume each frame has a fixed size */
+  gst_base_src_set_dynamic_size (src, FALSE);
```
myungjoo, line 889: It is OK to call set_synamic_size AFTER listener_start?

jaeyun-jung: blocksize is fixed value. I think this is ok to set the block size before calling gst_base_src_start_complete().

jaeyun-jung: this line is moved to _init() function.

PR #2515, opened 2020-07-08
[Tizen/Restriction] add sensor-src in restriction

jaeyun-jung:
change condition to check restricted element.
tizen sensor-src will be included in restrict list.

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -1271,7 +1271,7 @@ ml_check_plugin_availability (const char *plugin_name, const char *element_name)
   }
 
   /* nnstreamer elements */
-  if (g_str_equal (plugin_name, "nnstreamer") &&
+  if (g_str_has_prefix (plugin_name, "nnstreamer") &&
       g_str_has_prefix (element_name, "tensor_")) {
     return ML_ERROR_NONE;
   }


packaging/nnstreamer.spec
```diff
@@ -386,7 +386,7 @@ You may enable this package to use Google Edge TPU with NNStreamer and Tizen ML
...
+%define restricted_element	'capsfilter input-selector output-selector queue tee valve appsink appsrc audioconvert audiorate audioresample audiomixer videoconvert videocrop videorate videoscale videoflip videomixer compositor fakesrc fakesink filesrc filesink audiotestsrc videotestsrc jpegparse jpegenc jpegdec pngenc pngdec tcpclientsink tcpclientsrc tcpserversink tcpserversrc udpsink udpsrc xvimagesink ximagesink evasimagesink evaspixmapsink glimagesink theoraenc lame vorbisenc wavenc volume oggmux avimux matroskamux v4l2src avsysvideosrc camerasrc tvcamerasrc pulsesrc fimcconvert tizenwlsink tensor_src_tizensensor'
```
myungjoo, line 389: I think you do not need this?

jaeyun-jung: condition for nnstreamer is,
```
  if (g_str_equal (plugin_name, "nnstreamer") &&
      g_str_has_prefix (element_name, "tensor_")) {
  }
```

sensor-src is not in nnstreamer plugin, so this is necessary.

myungjoo: Ah.. then, what about:
```
  if (g_str_has_prefix (plugin_name, "nnstreamer") &&
      g_str_has_prefix (element_name, "tensor_")) {
  }
```

jaeyun-jung: I updated this condition. (plugin prefix 'nnstreamer')

anyj0527: Works on tw3. Checked ✅

PR #2529, opened 2020-07-10
[Converter] bytes to multi tensors @open sesame 7/15 15:00

jaeyun-jung:
parse the list of input info (dimension and type), and convert byte-stream to multi tensors.

### DIFF
--- a/gst/nnstreamer/tensor_converter/tensor_converter.c
+++ b/gst/nnstreamer/tensor_converter/tensor_converter.c
@@ -343,7 +343,7 @@ gst_tensor_converter_init (GstTensorConverter * self)
   self->frame_size = 0;
   self->remove_padding = FALSE;
   self->externalConverter = NULL;
-  gst_tensor_info_init (&self->tensor_info);
+  gst_tensors_info_init (&self->tensors_info);
 
   self->adapter = gst_adapter_new ();
   g_assert (self->adapter != NULL);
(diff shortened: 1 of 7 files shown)


gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -691,6 +710,51 @@ _gst_tensor_converter_chain_timestamp (GstTensorConverter * self,
...
+
+        gst_buffer_append_memory (buffer,
+            gst_memory_new_wrapped (0, data, size, 0, size, data, g_free));
+      }
+
+      gst_memory_unmap (mem, &info);
+      gst_memory_unref (mem);
+
+      /* copy timestamps */
+      gst_buffer_copy_into (buffer, buf, GST_BUFFER_COPY_METADATA, 0, -1);
```
dongju-chae, line 749: Q. why does it have to separate data/metadata copies (i.e., not using `GST_BUFFER_COPY_MEMORY`) ?

jaeyun-jung: The input buffer from byte stream has one memory block. If user configures multiple tensors as output buffer, tensor-converter should divide one memory into multiple blocks (multi tensors).

dongju-chae: Oh, thanks for the explanation.

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -378,19 +378,37 @@ gst_tensor_converter_set_property (GObject * object, guint prop_id,
...
+      value_str = g_value_get_string (value);
+      num = gst_tensors_info_parse_dimensions_string (info, value_str);
+
+      if (num == 0) {
+        GST_WARNING ("%s is invalid dimension string.", value_str);
+      } else if (info->num_tensors > 0 && info->num_tensors != num) {
+        GST_WARNING ("%s, the number of tensor is %u.", value_str, num);
+      }
+
+      info->num_tensors = num;
```
dongju-chae, line 399: Q. Is it okay to perform `info->num_tensors = num` even when `num == 0`? IMO, it's somewhat weird because `info` will be not changed in `gst_tensors_info_parse_dimensions_string()` in this case.

jaeyun-jung: info (self->tensors_info) will be updated if value string is not null.
Actually if developer sets invalid value string (type or dim) in pipeline, tensor-converter may raise an error or work improperly.

dongju-chae: Well.. please refer to the below codes to explain why I thought like that.

https://github.com/nnstreamer/nnstreamer/blob/f7b16f3127235f59479b04e3c10ada7a1ca075d4/gst/nnstreamer/tensor_common.c#L380-L382

Even if value_str is not null, `info` is not changed when its dims (i.e., `num_dims`) are evaluated to 0.
Please let me know if I'm wrong. And, I also totally agree with your thought about an invalid string.

jaeyun-jung: You mean..about the initialization when user changes the property?
(old number 3 > update 2 dimensions, then count is 2 but the array with index 2 has not empty)

dongju-chae: Resolved in offline discussion :)

jaeyun-jung: I will add initialization for dimensions and types.
```
      for (i = num; i < NNS_TENSOR_SIZE_LIMIT; ++i) {
        info->info[i].type = _NNS_END;
      }
```

jaeyun-jung:
2020-07-16 PR updated
- add initialization when setting the dimensions or types.

tests/nnstreamer_sink/unittest_sink.cc
```diff
@@ -2938,6 +2946,85 @@ TEST (tensor_stream_test, octet_2f)
   _free_test_data ();
 }
 
+/**
+ * @brief Test for octet stream.
+ */
+TEST (tensor_stream_test, octet_multi_tensors)
```
myungjoo, line 2952: Need more negative cases for unittest_sink.cc (2 more).

tests/nnstreamer_plugins/unittest_plugins.cc
```diff
@@ -1752,6 +1752,172 @@ TEST (test_tensor_aggregator, aggregate_5)
   gst_harness_teardown (h);
 }
 
+/**
```
myungjoo, line 1755: Need more negative test cases for unittest_plugins.cc (4 or more)

tests/nnstreamer_sink/unittest_sink.cc
jaeyun-jung: add 7 negative tc in unittest_plugins

PR #2535, opened 2020-07-13
[tizensensor] Fix tensor_src_tizensensor to work properly

anyj0527:
- Remove timestamp in `_create()` and `_fill()` and let BaseSrc do timestamp
- Remove `_get_times()`
- Init debug category
- Add query for latency
- set_interval for tizen sensor_listener
- Add debug log for buffer timestamp in `tensor_sink_render_buffer`

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
@@ -59,7 +59,7 @@
  *
  * @todo Add "power management" options (Tizen sensor f/w accepts such)
  *
- * @todo Some sensor tpes are privileged. We need privilege control.
+ * @todo Some sensor types are privileged. We need privilege control.
  * Some sensor types are privileged. An application should have the privilege http://tizen.org/privilege/healthinfo to get handles for the following sensors: SENSOR_HRM, SENSOR_HRM_LED_GREEN, SENSOR_HRM_LED_IR, SENSOR_HRM_LED_RED, SENSOR_HUMAN_PEDOMETER, SENSOR_HUMAN_SLEEP_MONITOR, SENSOR_HUMAN_SLEEP_DETECTOR, and SENSOR_HUMAN_STRESS_MONITOR.
  *
  * @todo Some sensor types appear to have mixed types (float32 & int32).
@@ -136,7 +136,7 @@ enum
 /**
  * @brief Default sensor retrieving frequency
  */
-#define DEFAULT_PROP_FREQ_N 1
+#define DEFAULT_PROP_FREQ_N 10
 #define DEFAULT_PROP_FREQ_D 1
 
 /**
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -961,9 +975,11 @@ _ts_get_gstcaps_from_conf (GstTensorSrcTIZENSENSOR * self)
     const gchar *typestr = gst_tensor_get_type_string (spec->type);
     gchar *dimstr = gst_tensor_get_dimension_string (spec->dimension);
     g_assert (typestr && typestr[0]);
-    tensor = g_strdup_printf ("other/tensor, dimension=%s, type=%s ; "
-        "other/tensors, num_tensors=1, dimensions=%s, type=%s",
-        dimstr, typestr, dimstr, typestr);
+    tensor = g_strdup_printf ("other/tensor,framerate=(fraction)%d/%d,dimension=%s,type=%s ; "
```
jaeyun-jung, line 978: For future work, let's consider gst_tensors_caps_from_config() and gst_tensor_caps_from_config() to get the caps.

anyj0527: I updated the part. Please review.

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -949,26 +963,34 @@ static GstCaps *
 _ts_get_gstcaps_from_conf (GstTensorSrcTIZENSENSOR * self)
 {
   const GstTensorInfo *spec;
-  gchar *tensor;
   GstCaps *retval;
+  GstTensorConfig tensor_config;
```
dongju-chae, line 967: Do you really need to use this extra config?

anyj0527: To make use of `gst_tensor(s)_caps_from_config` the config variable is needed. Else, caps may defined from hard-coded string. So I introduce the `tensor_config`.

And, I moved the declaration of `tensor_config` into else {} statement. Please check!

dongju-chae: Yes, but I think the codes seem to be redundant (e.g., setting rate). Is it possible to reuse `tensors_info` because `num_tensors = 1`?

anyj0527: `GstTensorsConfig` consists of `GstTensorsInfo` and rate info, not of `GstTensorConfig`.  Thus, IMHO, the two config variables (`tensor_config` and `tensors_config`) are necessary. If you have a neat way to address, please let me know..

dongju-chae: Ahh got it

ext/nnstreamer/tensor_source/tensor_src_tizensensor.c
```diff
@@ -949,26 +965,34 @@ static GstCaps *
...
   GstCaps *retval;
 
-  /** @bug the retrieved spec is not valid...? */
-  spec = g_hash_table_lookup (tizensensors, GINT_TO_POINTER (self->type));
+  spec = self->src_spec;
 
   if (FALSE == self->configured || SENSOR_ALL == self->type || NULL == spec) {
-    tensor = g_strdup_printf ("other/tensor; other/tensors, num_tensors=1");
+    retval = gst_caps_from_string (GST_TENSOR_CAP_DEFAULT "; "
+      GST_TENSORS_CAP_DEFAULT);
```
myungjoo, line 974: ```GST_TENSORS_CAP_DEFAULT``` = ```other/tensors, num_tensors = (int) (1, 16], framerate = ...```. It makes pipeline initialization slower than ```other/tensors, num_tensors = 1``` as long as we can sure that the number of tensors is always 1.

anyj0527: Right, thanks! I updated the caps with `GST_TENSORS_CAP_WITH_NUM ("1")`. Please check!

PR #2539, opened 2020-07-15
[SAM/CodeClean] remove global val

jaeyun-jung:
sam report, remove global val in openvino plugin.
update script to install common header file.


jaeyun-jung:
@myungjoo now unittest for openvino includes header to extend openvino class.

PR #2540, opened 2020-07-15
[Demux] Bugfix for set_property function

again4you:
If `tensorpick` property of demux element is set twice by calling
`g_object_set()` function, the previous value should be removed and
new one should be written. However, the value is appended and it can
make a problem. This patch fixes this bug.

**Self evaluation:**


gst/nnstreamer/tensor_demux/gsttensordemux.c
```diff
@@ -546,6 +546,12 @@ gst_tensor_demux_set_property (GObject * object, guint prop_id,
       const gchar *param = g_value_get_string (value);
       gchar **strv = g_strsplit_set (param, ",.;/", -1);
       gint num = g_strv_length (strv);
+
+      /* Before setting the new Tensor Pick data, the existing one should be removed. */
+      if (filter->tensorpick) {
+        g_list_free (filter->tensorpick);
```
jaeyun-jung, line 552: I found that demux has no finalize cb, this would be small mem leak.

PR #2550, opened 2020-07-17
[C-API] Bugfix for abnormal accessing of the callback function

again4you:
If the handle of sink-type element (i.e. tensor_sink and app_sink) is fetched by calling both `ml_pipeline_element_get_handle()` and `ml_pipeline_sink_register()` functions, segmantation fault occurs because of abnormal accessing for callback function. This PR adds this case and fixes this bug

It contains the following patches.
* Bugfix for abnormal accessing of the callback function
* Add test case for abnormal accessing of callback function

**Self evaluation:**

#### Submit v2
* Use `g_list_free_full()` and `GDestroyNotify` instead of iteration of GList.
#### Submit v1
* First draft

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -31,8 +31,9 @@
 #include "tensor_filter_custom_easy.h"
 #include "nnstreamer_plugin_api.h"
 
-#define handle_init(type, name, h) \
-  ml_pipeline_##type *name= (h); \
+
+#define handle_init(name, h) \
+  ml_pipeline_common_elem *name= (h); \
   ml_pipeline *p; \
   ml_pipeline_element *elem; \
   int ret = ML_ERROR_NONE; \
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -332,8 +337,17 @@ cleanup_node (gpointer data)
     e->handle_id = 0;
   }
 
-  if (e->handles)
-    g_list_free_full (e->handles, g_free);
+  l = e->handles;
```
jaeyun-jung, line 340: simply how about adding cb free function?
```
static void
free_element_handle (gpointer data)
{
  ml_pipeline_common_elem *item = (ml_pipeline_common_elem *) data;
  g_free (item->callback_info);
  g_free (item);
}

if (e->handles)
  g_list_free_full (e->handles, free_element_handle);
```

again4you: Nice catch! I will

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -5949,6 +5949,52 @@ TEST (nnstreamer_capi_element, get_property_04_n)
   EXPECT_EQ (status, ML_ERROR_NONE);
 }
 
+TEST (nnstreamer_capi_element, scenario_05_p)
```
myungjoo, line 5952: Add negative cases please!

myungjoo:
We can't proceed unless you add at least two negative cases. The negative-case ratio is under the required threshold.

PR #2551, opened 2020-07-17
[Api/Single] default no timeout when invoking model

jaeyun-jung:
If user does not set the timeout, single-invoke function will wait for the output till the invoke is done.
This requires another ACR.

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
@@ -142,7 +142,7 @@ public final class SingleShot implements AutoCloseable {
      * input data frames of an instance of a model should share the same dimension.
      * To change the input information, you should call {@link #setInputInfo(TensorsInfo)} before calling invoke method.
      *
-     * Note that this has a default timeout of 3 seconds.
+     * Note that this will wait for the result until the invoke process is done.
      * If an application wants to change the time to wait for an output,
      * set the timeout using {@link #setTimeout(int)}.
      *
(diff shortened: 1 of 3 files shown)


api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
```diff
@@ -142,7 +142,7 @@ public SingleShot(@NonNull File[] models, @Nullable TensorsInfo in, @Nullable Te
      * input data frames of an instance of a model should share the same dimension.
      * To change the input information, you should call {@link #setInputInfo(TensorsInfo)} before calling invoke method.
      *
-     * Note that this has a default timeout of 3 seconds.
+     * Note that, this will wait for the result until the invoke process is done.
```
dongju-chae, line 145: Minor comment: unnecessary comma?

jaeyun-jung: yes!

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -839,27 +839,38 @@ ml_single_invoke (ml_single_h single,
...
+      /** Free if any output memory was allocated */
+      if (*single_h->output != NULL) {
+        ml_tensors_data_destroy ((ml_tensors_data_h) *single_h->output);
+        *single_h->output = NULL;
+      }
     }
+  } else {
+    /** @todo calling klass->invoke() may suspend current thread */
+    g_cond_wait (&single_h->cond, &single_h->mutex);
+    status = single_h->status;
```
myungjoo, line 866: Improve this by invoking directly here.

PR #2557, opened 2020-07-21
[C-API] add the EOS push at the destroy moment

helloahn:
for the specific situation, tensor_mux(refresh), the EOS buffer has to be inserted to make the pipeline state as NULL

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -294,6 +294,9 @@ cb_bus_sync_message (GstBus * bus, GstMessage * message, gpointer user_data)
     return;
 
   switch (GST_MESSAGE_TYPE (message)) {
+    case GST_MESSAGE_EOS:
+      pipe_h->isEOS = TRUE;
+      break;
     case GST_MESSAGE_STATE_CHANGED:
       if (GST_MESSAGE_SRC (message) == GST_OBJECT_CAST (pipe_h->element)) {
         GstState old_state, new_state;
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -335,6 +335,16 @@ cleanup_node (gpointer data)
...
 
   g_mutex_lock (&e->lock);
+  if (e->type == ML_PIPELINE_ELEMENT_APP_SRC) {
+    if (gst_app_src_end_of_stream(GST_APP_SRC(e->element)) != GST_FLOW_OK){
+      ml_logw ("Failed to set EOS in %s", e->name);
+    }
+    /**
+     * Run 20 ms sleep mode until the pipeline processes the EOS buffer.
+     */
+    g_usleep(20000);
```
myungjoo, line 345: We have g_usleep(50000; at ```ml_pipeline_destroy```, where ```g_hash_table_remove_all (p->namednodes)``` is already called, invoking ```cleanup_node();```

Do you still need g_usleep(20000; here?

helloahn: Yes, it is still required.  This update performed at the time of `g_hash_table_remove_all()` as you referred, and at this time, the pipeline is still in the state `GST_STATE_PLAYING`. For this reason, when we push EOS, we have to wait enough time until the EOS arrives at the end of the pipeline.

If we don't, we will confront with a deadlock at the state changing step(`scret = gst_element_set_state (p->element, GST_STATE_PAUSED);`).

The 50ms sleep time you mentioned, IMO, is waiting for the state change (PLAYING->PAUSE).

I also checked it WITHOUT 20ms waiting time after push EOS, and it occurs a deadlock.

myungjoo: If you really need to wait for some events, using a constant delay may potentially incur timing errors in slower devices and performance issues in faster devices.

Because this is not just a test case, but a critical CAPI implementation, could you write a check-and-wait loop here?

helloahn: The 'check-and-wait' loop has been applied. Please check it.

PR #2569, opened 2020-07-23
[Single] Add unittest to measure invoke latency by single API

kparichay:
Added unittest to measure the invoke latency by single API.
This latency is measured for tensorflow-lite and nnfw-runtime filters.
The latency is measured by running the invoke with single API and running the invoke over the same data with the same model using filter API directly.
The difference in average execution time is reported as the latency introduced by the single API invoke per call.
The invoke is called 1000 times after a warmup where the output is also verified.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -62,6 +62,8 @@ static const gchar *nnfw_accl_default = ACCL_CPU_STR;
 void init_filter_nnfw (void) __attribute__ ((constructor));
 void fini_filter_nnfw (void) __attribute__ ((destructor));
 
+GstTensorFilterFrameworkStatistics nnfw_internal_stats;
+
 /**
  * @brief private data structure for the nnfw framework
  */
@@ -642,10 +644,12 @@ nnfw_invoke (const GstTensorFilterProperties * prop,
     void **private_data, const GstTensorMemory * input,
     GstTensorMemory * output)
 {
+  int64_t start_time, stop_time;
   nnfw_pdata *pdata;
   int err = 0;
   NNFW_STATUS nnfw_status;
 
+  start_time = g_get_monotonic_time ();
   g_return_val_if_fail (private_data != NULL, -EINVAL);
   pdata = (nnfw_pdata *) * private_data;
 
@@ -656,10 +660,19 @@ nnfw_invoke (const GstTensorFilterProperties * prop,
   err = nnfw_tensor_memory_set (prop, pdata, output, FALSE);
   if (err < 0)
     return err;
+  stop_time = g_get_monotonic_time ();
 
+  nnfw_internal_stats.total_overhead_latency += stop_time - start_time;
+
+  start_time = g_get_monotonic_time ();
   nnfw_status = nnfw_run (pdata->session);
+  stop_time = g_get_monotonic_time ();
+
   if (nnfw_status != NNFW_STATUS_NO_ERROR)
     err = -EINVAL;
+
+  nnfw_internal_stats.total_invoke_latency += stop_time - start_time;
+  nnfw_internal_stats.total_invoke_num += 1;
 
   return err;
 }
@@ -689,6 +702,10 @@ static GstTensorFilterFramework NNS_support_nnfw = {
 void
 init_filter_nnfw (void)
 {
+  nnfw_internal_stats.total_invoke_num = 0;
+  nnfw_internal_stats.total_invoke_latency = 0;
+  nnfw_internal_stats.total_overhead_latency = 0;
+
   NNS_support_nnfw.name = filter_subplugin_nnfw;
   NNS_support_nnfw.allow_in_place = FALSE;
   NNS_support_nnfw.allocate_in_invoke = FALSE;
@@ -699,6 +716,7 @@ init_filter_nnfw (void)
   NNS_support_nnfw.getOutputDimension = nnfw_getOutputDim;
   NNS_support_nnfw.setInputDimension = nnfw_setInputDim;
   NNS_support_nnfw.checkAvailability = nnfw_checkAvailability;
+  NNS_support_nnfw.statistics = &nnfw_internal_stats;
 
   nnstreamer_filter_probe (&NNS_support_nnfw);
 }
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -161,6 +161,16 @@ typedef struct _GstTensorFilterProperties
 } GstTensorFilterProperties;
 
 /**
+ * @brief Structure definition for tensor-filter framework statistics
+ */
+typedef struct _GstTensorFilterFrameworkStatistics
+{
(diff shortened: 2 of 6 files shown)


tests/tizen_capi/unittest_tizen_capi_latency.cc
```diff
@@ -0,0 +1,322 @@
...
+
+      status = sp->getOutputDimension (&prop, &data, &res);
+      EXPECT_EQ (status, 0);
+
+      output.type = res.info[0].type;
+      output.size = gst_tensor_info_get_size (&res.info[0]);
+
+      /** Allocate input and output data */
+      input.data = g_malloc(input.size);
+      output.data = g_malloc(output.size);
```
kparichay, line 216: @myungjoo This is one of the key differences between single and direct invoke. Single API invoke allocates memory for output internally with each call to invoke. So, a direct comparison to invoke between single and direct add extra overhead of this output data allocation. 
Should this allocation be moved to time critical section for a fair comparison?

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -657,6 +661,12 @@ nnfw_invoke (const GstTensorFilterProperties * prop,
   if (nnfw_status != NNFW_STATUS_NO_ERROR)
     err = -EINVAL;
 
+#if (DBG)
+  gint64 stop_time = g_get_real_time ();
+  g_message ("Invoke() is finished: %" G_GINT64_FORMAT,
+      (stop_time - start_time));
+#endif
+
```
myungjoo, line 669: You are meauring incorrect values.

C-API Single Invoke Latency for nnfw = TOTAL-LATENCY(ml_single_invoke) - LATENCY(nnfw_run)

If we do memory_set in invoke, we'd better measure the latency of two memory_set calls as well. (need to see if these calls affect latency; if so, we need to find a way to mitigate them)

tests/tizen_capi/unittest_tizen_capi_latency.cc
myungjoo: If it is malloc'ed by app, it is not our problem (not our latency). If it is malloc'ed by nnstreamer's code (not app or unittest), it is our problem (our latency)

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
kparichay: Updated to measure latency added by single API + latency by the tensor filter extension.

kparichay:
V3:
Sleep for 30sec between consecutive latency tests for different frameworks for cooldown

tests/tizen_capi/unittest_tizen_capi_latency.cc
```diff
@@ -0,0 +1,280 @@
...
+        max_idx = idx;
+        max_value = array[idx];
+      }
+    }
+
+    /**
+     * entry 952 (idx 951) is orange as per
+     * tests/test_models/labels/labels.txt
+     */
+    EXPECT_EQ (max_idx, 951);
```
anyj0527, line 132: How about `951` -> `951U` ? I think compiling on Ubuntu would warn this.

kparichay: Updated.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -160,6 +160,16 @@ typedef struct _GstTensorFilterProperties
   int throughput; /**< The average throughput in the number of outputs per second */
 } GstTensorFilterProperties;
 
+/**
+ * @brief Structure definition for tensor-filter framework statistics
+ */
+typedef struct _GstTensorFilterFrameworkStatistics
+{
+  int64_t total_invoke_num;      /**< The total number of calls to invoke */
```
jaeyun-jung, line 168: how about adding flag to get the latency?
tensor-filter can set this flag true to calculate the latency. (default false to not get the latency)

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -638,10 +640,12 @@ nnfw_invoke (const GstTensorFilterProperties * prop,
     void **private_data, const GstTensorMemory * input,
     GstTensorMemory * output)
 {
+  int64_t start_time, stop_time;
   nnfw_pdata *pdata;
   int err = 0;
   NNFW_STATUS nnfw_status;
 
+  start_time = g_get_real_time ();
```
jaeyun-jung, line 648: Discussion:
https://developer.gnome.org/glib/stable/glib-Date-and-Time-Functions.html#g-get-real-time
`g_get_monotonic_time() is probably more useful for measuring intervals.`
use g_get_monotonic_time()?

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -160,6 +160,16 @@ typedef struct _GstTensorFilterProperties
...
+  int64_t total_invoke_latency;  /**< The total accumulated invoke latency (usec) */
+  int64_t total_overhead_latency;    /**< The total accumulated overhead latency from the extension (usec) */
+} GstTensorFilterFrameworkStatistics;
```
dongju-chae, line 171: I think this feature would be overlapped with `latency` and `throughput` in `GstTensorFilterProperties`.
Could you merge them into a unified one?

kparichay: The latency used by GstTensorFilterProperties measures the latency inside the tensor filter (measuring time around the invoke call). This one allows extensions to measures the latency themselves and tensor filter can extract it if needed.

I can merge the data structure into one. Is that what you are suggesting?

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
kparichay: With the flag to enable the measuring of latency, do you also want to link it with PROP_LATENCY? The description of PROP_LATENCY is towards tensor filter measuring the latency than extension measuring the latency around the backend library invoke call.
If you think its alright to control this latency with PROP_LATENCY, I will update it.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
kparichay: `g_get_monotonic_time` is better. Updated.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
jaeyun-jung: Let's do this later, not this PR.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
myungjoo: Do not merge C-API's infra and NNStreamer-core's infra.

kparichay:
Screenshot for RPI4 benchmark (armv7l) - 
Screenshot from 2020-09-03 12-54-06

Screenshot for x64 benchmark -
image

PR #2575, opened 2020-07-27
[tizensensor] Add tolerance for sensor data check @open sesame 07/28 15:30

anyj0527:
Ease the future timestamp check:
- At the check time, the timestamp of event could be the "next" sample
- Add tolerance amount of sampling rate
- Change Error to Warning

**Self evaluation:**


myungjoo:
Q: Why "WARNING"?, not "ERROR"? even with this tolerance value?

anyj0527:
> Q: Why "WARNING"?, not "ERROR"? even with this tolerance value?

I changed back to "ERROR".
For TCT, I tried simple pipeline repeatedly and the future event case does not occur with the tolerance value. I think existing ERROR would be more suitable.

PR #2577, opened 2020-07-27
[Android/Single] allow 0 timeout @open sesame 7/28 13:39

jaeyun-jung:
Now C-API allows setting 0 timeout val, change condition to set timeout value.


jaeyun-jung:
@again4you please change C# API later if this PR is merged.

myungjoo:
> BTW, should we add the `remarks` section for the default timeout and its behavior?

Yes, we should do it for API doxygen entries.

PR #2583, opened 2020-07-28
[single] Memory leak bug fix

kparichay:
Add memory leak bug fix
When output ml_tensors_data_h fails to allocate, the already computed output must be freed before return

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -152,6 +152,7 @@ __process_output (ml_single * single_h, GstTensorMemory * out_tensors)
   unsigned int i;
   int status = ML_ERROR_NONE;
   ml_tensors_data_s *out_data;
+  gboolean need_free = TRUE;
 
   /** Allocate output buffer */
   if (single_h->ignore_output == FALSE) {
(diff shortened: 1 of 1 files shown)


api/capi/src/nnstreamer-capi-single.c
```diff
@@ -160,6 +160,9 @@ __process_output (ml_single * single_h, GstTensorMemory * out_tensors)
     if (status != ML_ERROR_NONE) {
       ml_loge ("Failed to allocate the memory block.");
       (*single_h->output) = NULL;
+      /** free computed output data */
+      for (i = 0; i < single_h->out_info.num_tensors; i++)
```
jaeyun-jung, line 164: To reduce duplicated code (line 175~180), how about adding goto and checking the flag before return?
```
gboolean need_free = false;
...
process_error:
if (need_free)
  //free mem
```

kparichay: Done.

kparichay:
> With "FALSE" from th eif statement of LINE 158, you are now doing "g_free", which you didn't in previous commit or before this PR.

`g_free` is only performed if `ml_tensors_data_create_no_alloc` fails (in this case, the returned output to user is NULL, so the output should be freed). if it does not fail, `need_free` is set to FALSE at line 169. Before freeing memory, `need_free` variable is checked at LINE 183. This ensures that `g_free` is not called when valid output is to be returned.

PR #2584, opened 2020-07-28
[filter/single] Optimize invoke path for tensor filter single

kparichay:
Tensor filter single performs various check on invoke
However, these checks are done on properties which remain static after open and first invoke
So, these checks are now performed on the first invoke only
Further, some checks were being performed multiple times, which have now been removed.

**Self evaluation:**
    
### DIFF
--- a/api/capi/src/tensor_filter_single.c
+++ b/api/capi/src/tensor_filter_single.c
@@ -102,6 +102,7 @@ g_tensor_filter_single_init (GTensorFilterSingle * self)
   priv = &self->priv;
 
   gst_tensor_filter_common_init_property (priv);
+  self->allocate_in_invoke = FALSE;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


api/capi/src/tensor_filter_single.c
```diff
@@ -214,6 +216,7 @@ g_tensor_filter_single_start (GTensorFilterSingle * self)
     return FALSE;
 
   gst_tensor_filter_load_tensor_info (&self->priv);
+  self->allocate_in_invoke = gst_tensor_filter_allocate_in_invoke (priv);
```
jaeyun-jung, line 219: About first_invoke flag, how about using `priv->configured` flag in g_tensor_filter_single_start() function?
```
if (prop->input_configured && prop->output_configured)
  priv->configured = TRUE;

return priv->configured;
```
then, in g_tensor_filter_single_invoke() we can simply check the configured flag.
```
/** start if not already started */
if (priv->configured == FALSE) {
  if (!g_tensor_filter_single_start (self)) {
    return FALSE;
}
```

kparichay: That looks better. Updated.

PR #2585, opened 2020-07-28
[tensor] Update GstTensorMemory coherent with ml_tensors_data_s

kparichay:
One of the efforts to minimize API latency is to avoid creating data wrappers in
API and internally in nnstreamer and avoid its copying. Rather just interpret
the API data structure as internal definition.
This requires internal and external structures to be coherent.
This commit makes GstTensorMemory coherent with ml_tensor_data_s by removing
`tensor_type type` from GstTensorMemory.

type can be remove from GstTensorMemory as GstTensorMemory is used internally
and its usage is always supported with GstTensorInfo which contains the type information.
Correspondingly update the existing elements and extensions which use GstTensorMemory
to use the GstTensorInfo type information.

Also add bugfix for bug exposed with this patch - 
setInputDimension in python extension updates only the number of tensors that all the info

Related issue: #2587

**Self evaluation:**


kparichay:
Rebased.

PR #2592, opened 2020-07-29
[nnfw] Optimize nnfw invoke

kparichay:
This patch caches the type for the input and output data tensor
rather than calculating it at every invoke call.

**Self evaluation:**
    
### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -74,12 +74,15 @@ typedef struct
   nnfw_session *session;
   gchar *model_file;
   accl_hw accelerator;
+
+  NNFW_TYPE in_type[NNS_TENSOR_SIZE_LIMIT];   /**< cached input tensor types */
+  NNFW_TYPE out_type[NNS_TENSOR_SIZE_LIMIT];  /**< cached output tensor types */
 } nnfw_pdata;
 
 static void nnfw_close (const GstTensorFilterProperties * prop,
     void **private_data);
 static int nnfw_tensors_info_get (const nnfw_pdata * pdata,
-    const gboolean is_input, GstTensorsInfo * info);
+    const gboolean is_input, GstTensorsInfo * info, NNFW_TYPE * type);
 static int nnfw_tensor_type_from_gst (const tensor_type type,
     NNFW_TYPE * nnfw_type);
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -72,12 +72,15 @@ typedef struct
   nnfw_session *session;
   gchar *model_file;
   accl_hw accelerator;
+
+  NNFW_TYPE in_type[NNS_TENSOR_SIZE_LIMIT];   /**< cached input tensor types */
```
jaeyun-jung, line 76: How about keeping nnfw_tensorinfo for in/out information?  This will include rank and dimension also.

kparichay: Yes, nnfw_tensorinfo can be used. However, do we need that info?
Dimension is directly available from nnfw_pdata->in/out_info, and rank is not used internally by nnstreamer.

jaeyun-jung: gst_tensor_info_get_rank() in nnfw_tensor_info_set() (Maybe set-info is not available now)
We can get the exact rank from nnfw_tensorinfo.
or, removing GstTensorsInfo in_info; and GstTensorsInfo out_info;?

kparichay: We cannot extract the exact rank from nnfw_tensorinfo as the info is provided by the user.
The purpose of nnfw_tensor_info_set() is to generate the new nnfw_tensor_info from the provided GstTensorInfo. 
Using the cached nnfw_tensorinfo for this wont be possible.

kparichay:
Rebased.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -591,47 +600,33 @@ nnfw_tensor_memory_set (const GstTensorFilterProperties * prop,
...
   guint idx;
   unsigned int num_tensors = 0;
-  const GstTensorsInfo *info;
 
-  g_return_val_if_fail (prop != NULL, -EINVAL);
-  g_return_val_if_fail (mem != NULL, -EINVAL);
-  g_return_val_if_fail (pdata != NULL, -EINVAL);
-  g_return_val_if_fail (pdata->session != NULL, -EPERM);
+  g_return_val_if_fail (
+      G_UNLIKELY (prop != NULL || mem != NULL || pdata != NULL), -EINVAL);
```
jaeyun-jung, line 608: maybe  `&&` ? not `||`

kparichay: aah, fixed.

kparichay:
rebased.

Issue #2594, opened 2020-07-29
Tensor filter updates properties with empty string


kparichay:
Tensor filter set property updates its properties with empty string.
Single API passes the properties given by the user directly to the tensor_filter_common (after stringify).
In case, for example, if the names aren't set, empty name is given to the property.
gst_tesnor_filter_set_property() does not update the internal properties with empty properties.
However, it does update the number of tensors (to 0 in the above example).

Issue #2593, opened 2020-07-29
Tensor filter memory destroy bug


kparichay:
Memory allocated by the tensor filters inside them (which support allocate_in_invoke)
is not freed appropriately. When using the single API, this memory is wrapped with ml_tensors_data
and freed wtih ml_tensors_data_destroy. However, these memories allocated by the tensor filter itself 
must be destroyed with destroyer of the tensor filter extension.

kparichay:
This issue also exists in the pipeline C-API as well.

myungjoo:
Isn't this resolved?

kparichay:
The memory allocated by custom filters internally is NOT YET tracked by single/pipeline API to be able to call the filter handle later (if open) as of now.

PR #2595, opened 2020-07-29
[C-API] Allocate memory at once for tensor in tensors

kparichay:
This patch changes the semantics of memory allocation for tensor data.
Tensor data is now allocated for all the tensor in tensors.
This reduces the multiple calls to malloc/free to 1, thus reducing the latency.
The final location of the memory in the tensor data is set with the offset.
This updates the memory allocated ml_tensors_data_create() and memory allocated
internally in single API tensor filter.
Correspondongly ml_tensors_data_destroy is also correspondingly update.

Note that this does not update the memory allocation by the pipeline API elements.
The tensor filter in the pipeline API retains its original behavior.

This patch exposes 2 bugs :

1. Memory allocated by the tensor filters inside them (which support allocate_in_invoke)
is not freed appropriately. When using the single API, this memory is wrapped with ml_tensors_data
and freed wtih ml_tensors_data_destroy. However, these memories allocated by the tensor filter itself
must be destroyed with destroyer of the tensor filter extension (issue #2593).
This will be fixed in a different patch.

2. Single API sets the input/output tensors info provided by the user.
However, it sets all the info with empty string which is might not be even provided.
tensor_filter set_property() ignores empty properties for most of the times.
However it updates the number of tensors even with empty properties (issue #2594).
The failing unittest (nnstreamer_capi_singleshot.set_input_info_success_02) has been
fixed for now. A proper fix to the internals will be done in a different patch.

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -550,11 +550,15 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
 
   _data = (ml_tensors_data_s *) data;
 
+  /**
+   * Memory for the all the tensors is allocated at once. Freeing memory at
+   * 0th tensor frees all the allocated memory for the tensors.
+   */
+  if (_data->tensors[0].tensor)
+    g_free (_data->tensors[0].tensor);
+
   for (i = 0; i < ML_TENSOR_SIZE_LIMIT; i++) {
-    if (_data->tensors[i].tensor) {
-      g_free (_data->tensors[i].tensor);
-      _data->tensors[i].tensor = NULL;
-    }
+    _data->tensors[i].tensor = NULL;
   }
 
   g_free (_data);
@@ -609,6 +613,8 @@ ml_tensors_data_create (const ml_tensors_info_h info, ml_tensors_data_h * data)
   gint status = ML_ERROR_STREAMS_PIPE;
   ml_tensors_data_s *_data = NULL;
   gint i;
+  void *tensor;
+  size_t total_size = 0;
 
   check_feature_state ();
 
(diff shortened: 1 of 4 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -550,11 +550,15 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
 
   _data = (ml_tensors_data_s *) data;
 
+  /**
+   * Memory for the all the tensors is allocated at once. Freeing memory at
+   * 0th tensor frees all the allocated memory for the tensors.
+   */
+  if (_data->tensors[0].tensor)
+    g_free (_data->tensors[0].tensor);
```
myungjoo, line 558: No need for null checks for g_free.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -4814,6 +4815,7 @@ TEST (nnstreamer_capi_singleshot, set_input_info_success_02)
   out_dim[2] = 1;
   out_dim[3] = 1;
   ml_tensors_info_set_count (out_info, 1);
+  ml_tensors_info_set_tensor_name (out_info, 0, "output");
```
myungjoo, line 4818: The test should work without tensor names. If you want to test with tensor names, please add more test cases (and negatives...)

kparichay:
> Please check if the "alloc-once" scheme does not have issues with `gst_memory_new_wrapped ( ... )` in a num_tensors loop at LINE 1175 of nnstreamer-capi-pipeline.c
> 
> It appears that you may have g_free() on data->tensors[i], i = 0 .. num_tensors -1.

Yes, this patch does not consider this. However, there is a bigger issue with contiguous data allocation when using that data with a pipeline which has mux/demux (as mux/demux will split/join data and then the semantics of combined data allocation are not valid anymore). I will update this PR to support contiguous data allocation for single API only (single API needs separate data management anyway because of #2593).

kparichay:
Closing as this is not urgent for now, but requires fixes later.

PR #2597, opened 2020-07-29
[filter] Buf fix of setting num_tensors to 0

### LINKED ISSUE #2594: Tensor filter updates properties with empty string

kparichay:
In case, empty string is provided for the property,
tensor filter set_property() sets the number of tensors to 0 in
its private info.

Now with this patch, updating the num_tensors to non-zero is allowed.
However, setting to 0 is not.

Resolves #2594

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -1086,12 +1086,15 @@ _gtfc_setprop_INPUT (GstTensorFilterPrivate * priv,
     num_dims = gst_tensors_info_parse_dimensions_string (&prop->input_meta,
         g_value_get_string (value));
 
-    if (prop->input_meta.num_tensors > 0 &&
-        prop->input_meta.num_tensors != num_dims) {
-      ml_logw ("Invalid input-dim, given param does not match with old value.");
-    }
+    if (num_dims > 0) {
+      if (prop->input_meta.num_tensors > 0 &&
+          prop->input_meta.num_tensors != num_dims) {
+        ml_logw
+            ("Invalid input-dim, given param does not match with old value.");
+      }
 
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1086,12 +1086,15 @@ _gtfc_setprop_INPUT (GstTensorFilterPrivate * priv,
...
-    }
+    if (num_dims > 0) {
+      if (prop->input_meta.num_tensors > 0 &&
+          prop->input_meta.num_tensors != num_dims) {
+        ml_logw
+            ("Invalid input-dim, given param does not match with old value.");
+      }
 
-    prop->input_meta.num_tensors = num_dims;
+      prop->input_meta.num_tensors = num_dims;
```
myungjoo, line 1096: Is it ok to update ```prop->input_meta.num_tensors``` when ```num_dims == 0```? (and same for other similar variables)

kparichay: > Is it ok to update prop->input_meta.num_tensors when num_dims == 0? (and same for other similar variables)

That is exactly what this patch prevents. If `num_dims` interpreted with the given property is 0, then it does not update the `prop->input_meta.num_tensors`.

However, if multiple properties have different dimensions, then the property setting the last dimension wins with the current implementation (supporting changing the num_tensors of an already configured tensor_filter).

PR #2602, opened 2020-07-30
[C-API] int32/uint32 get/set functions support enum type.

again4you:
This patch makes int32/uint32 element-wise getter/setter functions
support Enumeration type. It is much natural and easy to use for
developers. I also add related test cases.

**Self evaluation:**

#### Submit v2
* Remove remark section related with long and enum type.
* Update the negative test case of element-wise control function.

#### Submit v1
* First draft

### DIFF
--- a/api/capi/include/nnstreamer.h
+++ b/api/capi/include/nnstreamer.h
@@ -715,7 +715,6 @@ int ml_pipeline_element_set_property_int32 (ml_pipeline_element_h elem_h, const 
 /**
  * @brief Sets the integer 64bit value of element's property in NNStreamer pipelines.
  * @since_tizen 6.0
- * @remarks This function supports both Integer64 and Long.
  * @param[in] elem_h The target element handle.
  * @param[in] property_name The name of the property.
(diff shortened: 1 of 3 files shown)


again4you:
@jaeyun-jung Could you verify this?

again4you:
>     1. Do you really need to add remarks for enum handling? Can't we just not mention enum here? (it appears to be "undefined" in C/C++)
It just informs users that these APIs support the enum type. If it is common sense, then I will remove it.

>     2. API auditors won't like the usage of capital letters in the remarks.
ACK.


>     3. (not in this PR, but...) `@remarks This function supports both Integer64 and Long`: is this really true? LONG == INT64? not LONG LONG == INT64?

As you already know, it depends on system architecture.
`INT64` (i.e. `gint64`) guaranteed to be 64 bits on __all__ systems. However `LONG` (i.e. `glong`) is different from that.
`glong` is type-casted to `long` and the size of the standard C long type might be 32bit on 32bit machine or 64bit on 64bit machine. Because of this reason, I use INT64 to support both of the machines.

PR #2607, opened 2020-08-02
[example] Use macro to copy tensor info

tony-jinwoo-ahn:
Add a macro to copy tensor info

**Self evaluation:**

### DIFF
--- a/nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
+++ b/nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
@@ -20,6 +20,7 @@
 #include <errno.h>
 #include <math.h>
 #include <tensor_filter_custom.h>
+#include <glib.h>
 
 #define TSIZE   (4)
 
@@ -67,8 +68,27 @@ pt_exit (void *private_data, const GstTensorFilterProperties * prop)
 {
   pt_data *data = private_data;
   assert (data);
+
+  g_free (data->info[0].name);
+  g_free (data->info[1].name);
+  g_free (data->info[2].name);
   free (data);
 }
+
+/**
+ * @brief copy tensor info
+ */
+#define copy_tensor_info(_info, _data) \
+  do { \
+    unsigned int i = 0; \
+    unsigned int num_tensors = 0; \
+    num_tensors = _info->num_tensors; \
+    assert (0 < num_tensors && num_tensors <= NNS_TENSOR_SIZE_LIMIT); \
+    for (i = 0; i < num_tensors; ++i) { \
+      _info->info[i] = _data->info[i]; \
+      _info->info[i].name = g_strdup(_data->info[i].name); \
+    } \
+  } while (0)
 
 /**
  * @brief get the input tensor dimensions of dummy-LSTM (4:4:4 float32, 4:4:4 float32)
@@ -83,10 +103,7 @@ get_inputDim (void *private_data, const GstTensorFilterProperties * prop,
   assert (NNS_TENSOR_RANK_LIMIT >= 3);
 
   info->num_tensors = 3;
-  /** @todo use common function to copy tensor info */
-  info->info[0] = data->info[0];
-  info->info[1] = data->info[1];
-  info->info[2] = data->info[2];
+  copy_tensor_info(info, data);
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
```diff
@@ -70,6 +70,22 @@ pt_exit (void *private_data, const GstTensorFilterProperties * prop)
...
+    unsigned int num_tensors = 0; \
+    assert (_info != NULL); \
+    assert (_data != NULL); \
+    num_tensors = _info->num_tensors; \
+    assert (0 < num_tensors && num_tensors <= NNS_TENSOR_SIZE_LIMIT); \
+    for (i = 0; i < num_tensors; ++i) { \
+      _info->info[i] = _data->info[i]; \
+    } \
+  } while (0)
+
```
myungjoo, line 88: ```memcpy(_info->info, _data->info, sizeof(_info->info[0]) * _info->num_tensors)``` ?

gichan-jang: I think... If info has a name, a fixed-size is not guaranteed.

nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
```diff
@@ -70,6 +70,22 @@ pt_exit (void *private_data, const GstTensorFilterProperties * prop)
...
 
+/**
+ * @brief copy tensor info
+ */
+#define copy_tensor_info(_info, _data) \
+  do { \
+    unsigned int i = 0; \
+    unsigned int num_tensors = 0; \
+    assert (_info != NULL); \
+    assert (_data != NULL); \
```
dongju-chae, line 81: I think such `assert ()` seem to be redundant.

tony-jinwoo-ahn: agree. will update this.

nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
myungjoo: _info->info[i] = _data->info[i] will have the same issue for ```char *name```.  _info->info[i].name will be pointing to the same address.

myungjoo: If "name" matters for the users, you need g_strdup and g_free for cloned tensors (or reset all names to NULL if that doesn't matter)

tony-jinwoo-ahn: @myungjoo Thank you for review. updated patch.

nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
```diff
@@ -67,9 +68,28 @@ pt_exit (void *private_data, const GstTensorFilterProperties * prop)
 {
   pt_data *data = private_data;
   assert (data);
+
+  g_free (data->info[0].name);
+  g_free (data->info[1].name);
+  g_free (data->info[2].name);
```
tony-jinwoo-ahn, line 74: actually, we don't need these 3 lines for this file, because 'name' is NULL in this case, but these code are no harm.
g_free internally checks NULL.

nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
tony-jinwoo-ahn: done.

nnstreamer_example/custom_example_LSTM/dummy_LSTM.c
tony-jinwoo-ahn: someone who use this macro must g_free(info->info[i].name) after this patch is merged; but that is out of scope from this file.

jaeyun-jung: This example did not have dependency to glib, gstreamer and nnstreamer. 
IMO, we dont need to consider the tensor name here.

If it is necessary to handle this in this example, I recommend to use gst_tensors_info_copy() here, include nnstreamer_plugin_api.h with dependency nnstreamer.

PR #2609, opened 2020-08-03
[C-Api] fix typo

jaeyun-jung:
add line comment and missing dot in api description.


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -1187,8 +1187,8 @@ ml_get_nnfw_subplugin_name (ml_nnfw_type_e nnfw)
     [ML_NNFW_TYPE_VIVANTE] = "vivante",
     [ML_NNFW_TYPE_EDGE_TPU] = "edgetpu",
     [ML_NNFW_TYPE_ARMNN] = "armnn",
-    [ML_NNFW_TYPE_SNAP] = "snap",
     [ML_NNFW_TYPE_SNPE] = "snpe",
```
anyj0527, line 1190: How about add `/* Android only */` here for snpe too?

jaeyun-jung: we can build snpe subplugin on ubuntu :)

anyj0527: Oh that's right!

PR #2616, opened 2020-08-05
[Android] build script for nnfw

jaeyun-jung:
Update build script
- Remove unnecessary so in nnfw prebuilt libs
- Add nnfw-ext option


chunseoklee:
LGTM for me

PR #2623, opened 2020-08-06
[Android/AMC] start thread after the construction

jaeyun-jung:
Fix exception case, start amcsrc-looper after calling the constructor.


jaeyun-jung:
@myungjoo fatal sig while running the android instrumentation test (once)

Issue #2645, opened 2020-08-14
Choosing backends for nnfw


semun-lee:
# Issue Description
acl_cl backend doesn't loaded even when I set ML_NNFW_HW_GPU as a hardware type.

```
ml_single_open(&single_h, "...", NULL, NULL, ML_NNFW_TYPE_NNFW, ML_NNFW_HW_GPU);
```
Expected Result
============
nnfw loads acl_cl backend when I set GPU as a hardware.

How to Reproduce
===============
1. call ml_single_open() with ML_NNFW_TYPE_NNFW and ML_NNFW_HW_GPU.

Further Information
===============

Tizen TM4 with OpenCL support. (armv7l binary)

It looks like "acceleartor" property of a filter object is not passed properly to the nnfw.

```
Thread 1 "ml_single_api" hit Breakpoint 3, nnfw_get_accelerator (pdata=0x429200, accelerators=0x0) at /usr/src/debug/nnstreamer-1.5.3/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c:235
235     in /usr/src/debug/nnstreamer-1.5.3/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
(gdb) finish
Run till exit from #0  nnfw_get_accelerator (pdata=0x429200, accelerators=0x0) at /usr/src/debug/nnstreamer-1.5.3/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c:235
nnfw_open (private_data=0x428120, prop=0x428124) at /usr/src/debug/nnstreamer-1.5.3/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c:236
236     in /usr/src/debug/nnstreamer-1.5.3/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
(gdb) p accelerator
$3 = 0xf4afa9a8 "cpu"
```

kparichay:
@semun-lee Thanks for reporting the issue.
I have added patch #2649 to fix this.
Can you please check if this issue is resolved with this patch?

kparichay:
@semun-lee Please check the comments https://github.com/nnstreamer/nnstreamer/pull/2649#issuecomment-674078486 and https://github.com/nnstreamer/nnstreamer/pull/2649#issuecomment-674076325 on the PR #2649.

myungjoo:
I think the workaround has fixed the issue?

semun-lee:
[...]

PR #2649, opened 2020-08-14
[single] Bug fix for setting accelerator @open sesame 8/15 18:50

kparichay:
Bug fix for setting accelerator in the C-API.
When accelerator is set after the framework, tensor_filter does not know the framework
to check its version and store that information appropriately.
So, when framework version does not match, set property returns without setting the property.
This patch changes the order of setting this information for single API.

Related Issue: #2645 #2650

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -666,7 +666,7 @@ ml_single_open_custom (ml_single_h * single, ml_single_preset * info)
   /* set accelerator, framework, model files and custom option */
   fw_name = ml_get_nnfw_subplugin_name (nnfw);
   hw_name = ml_nnfw_to_str_prop (hw);
-  g_object_set (filter_obj, "accelerator", hw_name, "framework", fw_name,
+  g_object_set (filter_obj, "framework", fw_name, "accelerator", hw_name,
       "model", info->models, NULL);
   g_free (hw_name);
 
--- a/tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cc
+++ b/tests/tizen_nnfw_runtime/unittest_tizen_nnfw_runtime_raw.cc
@@ -463,7 +463,7 @@ TEST (nnstreamer_nnfw_mlapi, invoke_single_00)
   ml_tensors_info_set_tensor_dimension (out_info, 0, out_dim);
 
   status = ml_single_open (&single, test_model, in_info, out_info,
-      ML_NNFW_TYPE_NNFW, ML_NNFW_HW_AUTO);
+      ML_NNFW_TYPE_NNFW, ML_NNFW_HW_CPU);
   EXPECT_EQ (status, ML_ERROR_NONE);
 
   /* let's ignore timeout (30 sec) */


kparichay:
@semun-lee This PR fixes setting the accelerator property with `single` C-API. 
However, I think this exposes a bug in nnfw (I am still confirming if it is coming from nnstreamer but from my initial review, it seems to be coming from nnfw) with `acl_neon` backend on aarch64 architecture.
The unittest `nnstreamer_nnfw_mlapi.invoke_single_00` sets the nnfw to `acl_neon` and runs a basic model. That same unittest passes for `armv7l` architecture but fails with `qemu: uncaught target signal 4 (Illegal instruction) - core dumped` on aarch64.

This bug is revealed with this patch as this patch corrects the setting of the backend with `single` C-API.

kparichay:
More details on the failed unittest:
unittest file - tests/tizen_nnfw_runtime/unittest_nnfw_runtime_raw.cpp
unittest name - nnstreamer_nnfw_mlapi.invoke_single_00
Error details - on aarch64 architecture, this unittest fails with the call `nnfw_prepare()` with error `qemu: uncaught target signal 4 (Illegal instruction) - core dumped`. The backend set in this test is `acl_neon` and model file used is `./tests/test_models/models/add.tflite`

Note that this same unittest with the described configuration runs successfully on armv7l.

kparichay:
> However, this is a workaround that does not fix the real problem, but removes the effects of the problem for Single APIs only. Thus, please continue with #2650

Yes, this is a hotfix for single API to enable GPU backends for nnfw. Proper fixing will be done with #2650.

myungjoo:
The fix appears to digged up a potential issue of unit tests:

```
[  201s] + bash /home/abuild/rpmbuild/BUILD/nnstreamer-1.5.3/packaging/run_unittests_binaries.sh ./tests/tizen_nnfw_runtime/unittest_nnfw_runtime_raw
[  201s] ~/rpmbuild/BUILD/nnstreamer-1.5.3/build ~/rpmbuild/BUILD/nnstreamer-1.5.3
[  201s] [==========] Running 11 tests from 2 test cases.
[  201s] [----------] Global test environment set-up.
[  201s] [----------] 2 tests from nnstreamer_nnfw_runtime_raw_functions
[  201s] [ RUN      ] nnstreamer_nnfw_runtime_raw_functions.get_dimension
[  201s] [       OK ] nnstreamer_nnfw_runtime_raw_functions.get_dimension (112 ms)
[  201s] [ RUN      ] nnstreamer_nnfw_runtime_raw_functions.invoke
[  201s] [       OK ] nnstreamer_nnfw_runtime_raw_functions.invoke (7 ms)
[  201s] [----------] 2 tests from nnstreamer_nnfw_runtime_raw_functions (119 ms total)
[  201s] 
[  201s] [----------] 9 tests from nnstreamer_nnfw_mlapi
[  201s] [ RUN      ] nnstreamer_nnfw_mlapi.invoke_single_00
[  201s] qemu: uncaught target signal 4 (Illegal instruction) - core dumped
[  201s] /home/abuild/rpmbuild/BUILD/nnstreamer-1.5.3/packaging/run_unittests_binaries.sh: line 18: 73374 Illegal instruction     (core dumped) ${entry} --gtest_output="xml:${entry##*/}.xml"
[  201s] ~/rpmbuild/BUILD/nnstreamer-1.5.3
[  201s] error: Bad exit status from /var/tmp/rpm-tmp.qrsLbq (%build)
[  201s] 
```

It appears that we cannot use "HW_AUTO" of aarch64 (NEON/aarch64) in GBS/QEMU emulator.
Workaround: do not use "HW_AUTO" in unit-test
Real Fix: with "HW_AUTO" or "NEON", probe if NEON is possible, use it only when it's available.

kparichay:
> Workaround: do not use "HW_AUTO" in unit-test

Not use HW_AUTO only for armv7l as its unittest passes. For aarch64, do not use HW_AUTO.

> Real Fix: with "HW_AUTO" or "NEON", probe if NEON is possible, use it only when it's available.

Will update to use flags such as `HWCAP_NEON` at runtime to check if NEON architecture is supported on the device.

kparichay:
V2:
Updated with unittest to `HW_CPU` than `HW_AUTO`. This will be properly resolved with separate PR.

jaeyun-jung (approved):

PR #2651, opened 2020-08-14
[filter] Independency on order of the properties

### LINKED ISSUE #2650: [filter] Order of setting the properties
With the introduction of two versions of GstTensorFilterFramework, there has come up a bug with setting the properties.
The order of properties has now become important as setting certain properties depends on the version of the framework.
So, the version of framework is needed to set those properties, which creates a required order among some of the properties.

Given the two pipelines below:
```
pipelineA = "tensor_filter accelerator=true:GPU framework=tensorflow_lite"
pipelineB = "tensor_filter framework=tensorflow_lite accelerator=true:GPU"
```
Both the pipelines would have the correct framework. However, `accelerator` for `pipelineA` would be set to NULL because when `accelerator` property is set, `framework` is not yet set. This issue exists for properties `accelerator` and `is_updatable`. 

This is the inherent issue behind issue #2465

kparichay:
This patch adds independence on the order of setting the properties
Properties of tensor filter which depends on the framework can now be set before setting the framework as well
When setting the framework, these properties are set again from the cache

Resolves #2650

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -95,6 +95,10 @@ static const gchar *regex_accl_elem_utils[] = {
  */
 static GList *parse_accl_hw_all (const gchar * accelerators,
     const gchar ** supported_accelerators);
+static gint _gtfc_setprop_IS_UPDATABLE (GstTensorFilterPrivate * priv,
+    GstTensorFilterProperties * prop, const GValue * value);
+static gint _gtfc_setprop_ACCELERATOR (GstTensorFilterPrivate * priv,
+    GstTensorFilterProperties * prop, const GValue * value);
 
 /**
  * @brief GstTensorFilter properties.
@@ -1005,7 +1009,9 @@ static gint
 _gtfc_setprop_FRAMEWORK (GstTensorFilterPrivate * priv,
     GstTensorFilterProperties * prop, const GValue * value)
 {
+  gint status;
   const gchar *fw_name = g_value_get_string (value);
+  GValue val = G_VALUE_INIT;
 
   if (priv->fw != NULL) {
     if (g_strcmp0 (priv->prop.fwname, fw_name) != 0) {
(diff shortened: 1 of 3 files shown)


kparichay:
V2:
Update assert to throw/catch in tensor_filter_support (needed by setting of cached properties)
Added checking of supported accelerators passed by the framework to be available at runtime before using them in tensor_filter

myungjoo (approved):
It's a bit difficult to review. (too many things in a single PR.) LGTM, anyway. (will need more reviewers than normal as this is a big fix)

kparichay:
@myungjoo Noted. Will keep smaller and independent commits per PR in future. 
Modularization related to tensor filter accelerators will be done soon.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1011,6 +1025,29 @@ _gtfc_setprop_FRAMEWORK (GstTensorFilterPrivate * priv,
...
+
+  /** set PROP_IS_UPDATABLE in case it was set before framework */
+  g_value_init (&val, G_TYPE_BOOLEAN);
+  g_value_set_boolean (&val, priv->is_updatable);
+  status = _gtfc_setprop_IS_UPDATABLE (priv, prop, &val);
+  if (status != 0) {
+    g_warning ("Set propery is-updatable failed with error: %d", status);
+    return status;
+  }
+  g_value_unset (&val);
```
dongju-chae, line 1037: How about moving `g_value_unset` to between line 1032 and 1033? Same issue in line 1048.

tests/nnstreamer_filter_tensorflow_lite/runTest.sh
```diff
@@ -96,7 +96,7 @@ function run_pipeline() {
 }
 
 arch=$(uname -m)
-if [ "$arch" = "aarch64" ] || [ "$arch" = "armv7l" ]; then
+if [ "$arch" = "armv7l" ]; then
```
dongju-chae, line 99: Why aarch64 is disabled?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -24,6 +24,14 @@
...
 
 #include <string.h>
+#if defined(__aarch64__) || defined(__arm__)
+#include <sys/auxv.h>
+#if defined(__ANDROID__)
+#include <asm/hwcap.h>
+#else /* __ANDROID __ */
+#include <asm-arm/hwcap.h>
+#endif /* __android __ */
+#endif /* __arch64__ __arm__ */
```
dongju-chae, line 34: IMO, it's better to create an additional header file like this [framework](https://ncnn.docsforge.com/master/api/ncnn/cpu_support_arm_neon/) in another PR,

kparichay: Yes, the accelerator related stuff will move to different files.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kparichay: Sorry for the late reply. this is done.

tests/nnstreamer_filter_tensorflow_lite/runTest.sh
kparichay: This was wrong change. Has been reverted.

kparichay:
@dongju-chae Sorry for the late update. I have updated as per your comments. Please review.

tests/nnstreamer_filter_tensorflow_lite/runTest.sh
```diff
@@ -96,7 +96,7 @@ function run_pipeline() {
 }
 
 arch=$(uname -m)
-if [ "$arch" = "aarch64" ] || [ "$arch" = "armv7l" ]; then
+if ["$arch" = "aarch64"] || [ "$arch" = "armv7l" ]; then
```
dongju-chae, line 99: Mistake here?

kparichay: yes, fixed.

dongju-chae (approved):
LGTM

PR #2671, opened 2020-08-25
[Android] Remove recently added testcases

anyj0527:
- Remove `-Wno-exceptions` flag from Android-snpe.mk
- Currently, the exception from tensor_filter is not handled properly in naive api
and causes test failure in specific build setting, so remove it
- This TCs could be restored after when this issue is fixed.

**Self evaluation:**


gichan-jang:
How about using @ignore annotation?

myungjoo:
@jaeyun-jung I think we have added ```-Wno-exceptions``` intentionally for some reasons related with ndk build. Is this ok now?

jaeyun-jung:
@myungjoo `-fexceptions` to handle exceptions. `-Wno-exceptions` is now unnecessary. (preventing build error with snpe module)

jaeyun-jung (approved):
Later work let's handle exceptions from native code in java.

anyj0527:
@myungjoo @jaeyun-jung The build flag `-Wno-exceptions` is not necessary for SNPE SDK v1.38+ (the very recent version is 1.41)
@gichan-jang The TCs that would be added may not be same as these. I think removing them is more neat.

myungjoo (approved):
Ok then. LGTM.

PR #2673, opened 2020-08-27
[Filter/Cpp] typo in cpp check

jaeyun-jung:
fix typo, cpp std version macro.


again4you (approved):
LGTM!

Issue #2675, opened 2020-08-28
Support for clamping tensor data


dongju-chae:
It seems that some private sub-plugins need to apply `clamp` for input tensor data in pre-processing.
I think `tensor_transform` is the best place to support this feature as an extra option.

PR #2676, opened 2020-08-28
[Clamp] Support 'clamp' as tensor_transform's option

dongju-chae:
This patch supports 'clamp' as tensor_transform's option.
It requires min/max values to apply clamp for input tensor data.

The usage is like below:
```
... ! tensor_transform mode=clamp option=-123.902313:150.837601 ! ...
```

It resolves https://github.com/nnstreamer/nnstreamer/issues/2675

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -95,6 +95,8 @@ GST_DEBUG_CATEGORY_STATIC (gst_tensor_transform_debug);
 #define REGEX_DIMCHG_OPTION "^([0-3]):([0-3])$"
 #define REGEX_TYPECAST_OPTION "(^[u]?int(8|16|32|64)$|^float(32|64)$)"
 #define REGEX_TRANSPOSE_OPTION "^(?:([0-2]):(?!.*\\1)){3}3$"
+#define REGEX_CLAMP_OPTION "^((([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?))):"\
+    "((([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?)))$"
 #define REGEX_ARITH_OPTION "^(typecast:([u]?int(8|16|32|64)|float(32|64)),)?"\
     "(((add|mul|div)(:([-+]?[0-9]*\\.?[0-9]+([eE][-+]?[0-9]+)?))+)(,|))+$"
 #define REGEX_ARITH_OPTION_TYPECAST "(typecast:([u]?int(8|16|32|64)|float(32|64)))"
@@ -200,6 +202,9 @@ gst_tensor_transform_mode_get_type (void)
       {GTT_STAND, "Mode for statistical standardization of tensor, "
             "option=default",
           "stand"},
+      {GTT_CLAMP, "Mode for clamping all elements of tensor into the range, "
+            "option=CLAMP_MIN:CLAMP_MAX",
+          "clamp"},
       {GTT_UNKNOWN, "Unknown or not-implemented-yet mode",
           "unknown"},
       {0, NULL, NULL},
@@ -924,6 +929,45 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
       ret = filter->loaded = TRUE;
       break;
     }
+    case GTT_CLAMP:
+    {
+      gchar **strv = NULL;
+
+      if (!g_regex_match_simple (REGEX_CLAMP_OPTION, filter->option, 0, 0)) {
+        ml_loge
+            ("%s: clamp: \'%s\' is not valid option string: it should be in the form of [CLAMP_MIN:CLAMP_MAX]\n",
+            filter_name, filter->option);
+        break;
+      }
+
+      strv = g_strsplit (filter->option, ":", 2);
+
+      filter->data_clamp.min = g_ascii_strtod (strv[0], NULL);
+      if (errno == ERANGE) {
+        ml_loge ("%s: clamp: CLAMP_MIN value has an invalid range\n",
+            filter_name);
+        g_strfreev (strv);
+        break;
+      }
+      filter->data_clamp.max = g_ascii_strtod (strv[1], NULL);
+      if (errno == ERANGE) {
+        ml_loge ("%s: clamp: CLAMP_MAX value has an invalid range\n",
+            filter_name);
+        g_strfreev (strv);
+        break;
+      }
+
+      g_strfreev (strv);
+
+      if (filter->data_clamp.min > filter->data_clamp.max) {
+        ml_loge ("%s: clamp: CLAMP_MIN is larger than CLAMP_MAX\n",
+            filter_name);
+        break;
+      }
+
+      ret = filter->loaded = TRUE;
+      break;
+    }
     default:
       GST_ERROR_OBJECT (filter, "Cannot identify mode\n");
       ret = FALSE;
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -924,6 +929,26 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
...
+            ("%s: clamp: \'%s\' is not valid option string: it should be in the form of [CLAMP_MIN:CLAMP_MAX]\n",
+            filter_name, filter->option);
+        break;
+      }
+
+      strv = g_strsplit (filter->option, ":", 2);
+
+      filter->data_clamp.min = g_ascii_strtod (strv[0], NULL);
+      filter->data_clamp.max = g_ascii_strtod (strv[1], NULL);
+
```
wooksong, line 947: How about adding code for checking the validity of CLAM_MIN and MAX here?

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -1390,6 +1415,32 @@ gst_tensor_transform_stand (GstTensorTransform * filter,
...
+ */
+static GstFlowReturn
+gst_tensor_transform_clamp (GstTensorTransform * filter,
+    const uint8_t * inptr, uint8_t * outptr)
+{
+  uint32_t *fromDim = filter->in_config.info.dimension;
+  size_t Size = fromDim[3] * fromDim[2] * fromDim[1] * fromDim[0];
+  float *in = (float *) inptr;
+  float *out = (float *) outptr;
+  int i;
```
wooksong, line 1435:
```suggestion
  size_t i;
```

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -1390,6 +1415,32 @@ gst_tensor_transform_stand (GstTensorTransform * filter,
...
+ * @param[out] outptr output tensor
+ * @return Gst flow status
+ */
+static GstFlowReturn
+gst_tensor_transform_clamp (GstTensorTransform * filter,
+    const uint8_t * inptr, uint8_t * outptr)
+{
+  uint32_t *fromDim = filter->in_config.info.dimension;
+  size_t Size = fromDim[3] * fromDim[2] * fromDim[1] * fromDim[0];
+  float *in = (float *) inptr;
```
wooksong, line 1433: Q: Why is float used here instead of NNS_TYPE *in = (NNS_TYPE *) inptr?

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -1390,6 +1415,32 @@ gst_tensor_transform_stand (GstTensorTransform * filter,
...
+  float *out = (float *) outptr;
+  int i;
+
+  for (i = 0; i < Size; i++) {
+    out[i] = CLAMP (in[i], filter->data_clamp.min, filter->data_clamp.max);
```
wooksong, line 1438: FYI, min and max are the double-typed variables.

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: Got it. I will add a routine to check overflow/underflow using `errno`.

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: I just took the similar codes to `gst_tensor_transform_stand`, but yes, they should be changed.

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: will be updated.

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: Checked.

wooksong (approved):
Looks good to me

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -924,6 +929,36 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
...
+
+      strv = g_strsplit (filter->option, ":", 2);
+
+      filter->data_clamp.min = g_ascii_strtod (strv[0], NULL);
+      if (errno == ERANGE) {
+        ml_loge ("%s: clamp: CLAMP_MIN value has invalid range\n", filter_name);
+        g_strfreev (strv);
+        break;
+      }
+      filter->data_clamp.max = g_ascii_strtod (strv[1], NULL);
```
kparichay, line 951: ensure that data_clamp.max > data_clamp.min?

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -1390,6 +1425,58 @@ gst_tensor_transform_stand (GstTensorTransform * filter,
...
+    /* extract the value */
+    data_idx = in_element_size * i;
+    gst_tensor_transform_set_value (filter, &value, in_tensor_type,
+        (gpointer) (inptr + data_idx));
+
+    /* in_tensor_type to double */
+    gst_tensor_transform_typecast_value (filter, &value, _NNS_FLOAT64);
+    gst_tensor_transform_get_value (filter, &value, (gpointer) (&tmp));
+
+    tmp = CLAMP (tmp, filter->data_clamp.min, filter->data_clamp.max);
```
kparichay, line 1465: This method performs the operation in double even for values of int or float which can become quite slow.
Any reason to with double?

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: Yes, updated.

gst/nnstreamer/tensor_transform/tensor_transform.c
dongju-chae: Yes, it makes some inefficient, But as this macro does not consider data types, we might have possible side-effects to handle such various data types. This can be fixed in later PR.

kparichay: Yes, as clamp as a macro doesnt need data type, all that's need to be done is to set the type of min/max/tmp to be same.
Later PR also works.

tests/transform_clamp/generateTest.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+##
+# SPDX-License-Identifier: LGPL-2.1-only
```
kparichay, line 4: SPDX license should be mentioned on the second line of the file

tests/transform_clamp/runTest.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+##
+## SPDX-License-Identifier: LGPL-2.1-only
```
kparichay, line 3: ditto
also single # should be used for SPDX

tests/transform_clamp/generateTest.py
dongju-chae: Well, in nnstreamer, there're many python scripts with the same description

kparichay: Then thats not correct and must be fixed.
Refernce - https://www.kernel.org/doc/Documentation/process/license-rules.rst

kparichay (approved):
Other than the license format, LGTM

tests/transform_clamp/generateTest.py
dongju-chae: I think that's the rule in kernel scripts. Also, I'm not sure that's out-of-scope in this PR. If it's really mandatory, fixing license issues across all python scripts can be handled in another PR.

kparichay: Yes, fixing all the scripts is out of scope for this PR. However, the license for the files added in the PR can be fixed to a proper style.

dongju-chae: I don't want to fix this because I'm not sure this is really necessary. I will keep the same format with other source files in nnstreamer. Also, I think that's not a hard rule. There are many python source files with SPDX specifier, available in many open source repo.

kparichay: ok

PR #2679, opened 2020-09-01
[Filter/NNFW] directory as model parameter

jaeyun-jung:
Support model path as model property for NNFW.

NNFW requires model path to open model.
Current tensor-filter and API does not allow directory string.
Check the model property and if it is directory, set framework to NNFW.

cc @lemmaa @chunseoklee 

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
@@ -180,13 +180,21 @@ public class APITestCommon {
     }
 
     /**
-     * Gets the File object of tensorflow-lite model.
+     * Gets the path string of tensorflow-lite add.tflite model.
+     */
+    public static String getTFLiteAddModelPath() {
+        String root = Environment.getExternalStorageDirectory().getAbsolutePath();
+        return root + "/nnstreamer/test/add";
+    }
+
+    /**
+     * Gets the File object of tensorflow-lite add.tflite model.
      * Note that, to invoke model in the storage, the permission READ_EXTERNAL_STORAGE is required.
      */
     public static File getTFLiteAddModel() {
-        String root = Environment.getExternalStorageDirectory().getAbsolutePath();
-        File model = new File(root + "/nnstreamer/test/add/add.tflite");
-        File meta = new File(root + "/nnstreamer/test/add/metadata/MANIFEST");
+        String path = getTFLiteAddModelPath();
+        File model = new File(path + "/add.tflite");
+        File meta = new File(path + "/metadata/MANIFEST");
 
         if (!model.exists() || !meta.exists()) {
             fail();
--- a/api/capi/src/nnstreamer-capi-util.c
+++ b/api/capi/src/nnstreamer-capi-util.c
@@ -935,13 +935,19 @@ ml_initialize_gstreamer (void)
  * @brief Internal helper function to validate model files.
  */
 static int
-_ml_validate_model_file (char **model, unsigned int num_models)
+_ml_validate_model_file (char **model, unsigned int num_models,
+    gboolean * is_dir)
 {
   guint i;
 
   if (!model || num_models < 1) {
     ml_loge ("The required param, model is not provided (null).");
     return ML_ERROR_INVALID_PARAMETER;
+  }
+
+  if (g_file_test (model[0], G_FILE_TEST_IS_DIR)) {
+    *is_dir = TRUE;
+    return ML_ERROR_NONE;
   }
 
   for (i = 0; i < num_models; i++) {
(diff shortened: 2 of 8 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -192,6 +134,13 @@ nnfw_open (const GstTensorFilterProperties * prop, void **private_data)
     }
   }
 
+  /* validate model file first */
+  if (!g_file_test (prop->model_files[0], G_FILE_TEST_IS_REGULAR)) {
```
lemmaa, line 138: ONE runtime should receive the model path. The path can be a regular file or a directory.

```suggestion
  if (!g_file_test (prop->model_files[0], (G_FILE_TEST_EXISTS | G_FILE_TEST_IS_DIR))) {
```

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -180,7 +123,6 @@ nnfw_open (const GstTensorFilterProperties * prop, void **private_data)
   int err = 0;
   nnfw_pdata *pdata;
   char *model_path = NULL;
-  char *meta_file = NULL;
```
leemgs, line 183: One question. After this PR, how can we get a meta-file name for the "nnfw" tensor filter?

jaeyun-jung: nnstreamer doesn't need to verify the meta. This is up to nnfw library.
If metadata is missing, nnfw will return error now.

Now to run nnfw, metadata is necessary. However, nnfw team considers to remove meta file or other option.
So he requested to remove this validation, and to handle directory path as model property.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
jaeyun-jung: I added new commit to handle directory string in tensor-filter. Please review the changes again.

jaeyun-jung:
PR updated (2020-09-03)
- handle directory string in tensor-filter (if model is directory, set the framework to NNFW)

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -944,6 +945,11 @@ _ml_validate_model_file (char **model, unsigned int num_models)
     return ML_ERROR_INVALID_PARAMETER;
   }
 
+  if (g_file_test (model[0], G_FILE_TEST_IS_DIR)) {
```
kparichay, line 948: does this mean that the `model` for nnfw should be provided as `model folder path` and not `model file path`, then unittests also needs updating.

kparichay: or atleast a new unittest needed with input as model path.

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -1017,16 +1023,30 @@ ml_validate_model_file (char **model, unsigned int num_models,
...
-    return status;
-
   if (!nnfw)
     return ML_ERROR_INVALID_PARAMETER;
 
+  status = _ml_validate_model_file (model, num_models, &is_dir);
+  if (status != ML_ERROR_NONE)
+    return status;
+
+  /* supposed it is ONE if given model is directory */
```
kparichay, line 1038: Add this handling to documentation in `tensor_filter_common.c` for `model` property as well as to the APIs?

kparichay:
This is a question about the abstraction offered by tensor_filter in the nnstreamer. 
tensor_filter has the property `model` (PROP_MODEL), which allows setting `model` for the tensor_filter independent of how the input is needed by the framework associated with the tensor filter. `tensor_filter` receives model_file for all the frameworks and caters the framework's requirement internally (like passing folder) than exposing it outside to the user.
IMO allowing providing model folder path to the property `model` changes the existing abstraction.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -192,6 +134,12 @@ nnfw_open (const GstTensorFilterProperties * prop, void **private_data)
     }
   }
 
+  /* validate model file first */
+  if (!g_file_test (prop->model_files[0], G_FILE_TEST_EXISTS)) {
```
kparichay, line 138: when the input is a model_path, how will this work?

kparichay: how about move this check to line 160 inside the else condition?

jaeyun-jung: https://developer.gnome.org/glib/stable/glib-File-Utilities.html
G_FILE_TEST_EXISTS returns TRUE if the file exists. It may or may not be a regular file.

jaeyun-jung:
@kparichay ONE team requests to enable setting both directory and model path for NNFW.
As you know, NNFW API requires the directory string to open model.

I think it is unnecessary to update API docs, setting directory is only for NNFW only.

api/capi/src/nnstreamer-capi-util.c
jaeyun-jung: Both model file path and directory.
I added testcases to open nnfw single-shot with directory.

api/capi/src/nnstreamer-capi-util.c
jaeyun-jung: IMO this is unnecessary, directory string is only available on NNFW now.

kparichay: ok

kparichay (commented):
LGTM

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
```diff
@@ -180,13 +180,21 @@ public static int getMaxScoreSNPE(ByteBuffer buffer) {
...
      */
     public static File getTFLiteAddModel() {
-        String root = Environment.getExternalStorageDirectory().getAbsolutePath();
-        File model = new File(root + "/nnstreamer/test/add/add.tflite");
-        File meta = new File(root + "/nnstreamer/test/add/metadata/MANIFEST");
+        String path = getTFLiteAddModelPath();
+        File model = new File(path + "/add.tflite");
+        File meta = new File(path + "/metadata/MANIFEST");
 
         if (!model.exists() || !meta.exists()) {
```
kparichay, line 199: remove meta check?

kparichay (approved):

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
jaeyun-jung: To setup the test environment now.
I remained the metadata in testcases, because current NNFW requires metadata.
This also should be removed later.

wooksong (approved):
LGTM

PR #2680, opened 2020-09-01
[Filter/TF] fix deallocator ptr

jaeyun-jung:
fix function ptr when creating new tensor struct in tf sub-plugin.


anyj0527 (approved):
LGTM 👍👍

dongju-chae (approved):
LGTM

PR #2681, opened 2020-09-01
[CodeClean] fix coverity issue

jaeyun-jung:
check return value when setting conf path after test is done.

### DIFF
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -998,18 +998,19 @@ TEST (conf_custom, env_str_01)
   EXPECT_FALSE (nnsconf_get_custom_value_bool ("customW", "n05", FALSE));
   EXPECT_TRUE (nnsconf_get_custom_value_bool ("customW", "n05", TRUE));
 
-  if (confenv) {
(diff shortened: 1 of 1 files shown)


anyj0527 (approved):
LGTM 👍🏼👍🏼

dongju-chae (approved):

tests/common/unittest_common.cc
```diff
@@ -998,18 +998,19 @@ TEST (conf_custom, env_str_01)
...
-  }
   g_free (f1);
   g_free (f2);
   g_free (f3);
   g_free (f4);
   g_free (f5);
   g_free (f6);
+
+  if (confenv) {
+    ASSERT_TRUE (g_setenv ("NNSTREAMER_CONF", confenv, TRUE));
```
wooksong, line 1009: Instead of ASSERT_TRUE, how about EXPECT_TRUE? 
I guess static analysis tools we are using may complains about "confenv is not freed".

jaeyun-jung: updated

wooksong (approved):
LGTM

PR #2682, opened 2020-09-01
[Android] deallocate direct buffer

jaeyun-jung:
Direct buffer allocated from native will not release the buffer.
Create instance in java and set byte array, then call method to add tensor data.


anyj0527 (approved):
LGTM 👍 Verified on S20

jaeyun-jung:
This PR reverts #2656.

wooksong (approved):

PR #2685, opened 2020-09-02
[Bug/TensorSplit] Fix the offset calculation bug

dongju-chae:
This patch fixes the offset calculation bug in tensor split.

Unless tensors have 1-byte data size, each tensor split's src gets invalid data.

**Self evaluation:**


jaeyun-jung (approved):

kparichay (approved):
LGTM

myungjoo (approved):

PR #2686, opened 2020-09-02
[single] Resolve tensor filter memory leak

### LINKED ISSUE #2593: Tensor filter memory destroy bug

kparichay:
Commit 1 : 
When tensor filter with `allocate_in_invoke` allocates the memory,
its memory should be destroyed with the filters destroy notify
However, this was ignored until now for the single API (#2593)

This patch fixes this
Single handle now creates a list of each such data buffer
as well as the single handle is stored in the data buffer.
If the data buffer is deallocated, then destroyNotify is called from the
single handle stored in it.
However, if the single handle is closed before the data buffer is freed,
then all the memory is freed from the list of the data buffers
stored in the single handle.

V2:
Added note to the description of `ml_single_invoke` about data not being available 
single handle is closed when memory is allocated by the framework.

Commit 2 :
Appropriate negative and positive unit tests are included to verify
that the data can be freed by the user before closing the filter or closing the filter frees the data
or partially both with multiple data points.
The negative tests were tried when destroying the data twice (destroying manually
after the single handle is closed) with `EXPECT_DEATH()`. However, the unit test
doesn't always die (modifying freed data is Undefined Behavior). So, it has been commented.
This as a side effect finally adds unit tests for custom framework
using allocate_in_invoke with the single API as well.

Resolves #2593

**Self evaluation:**

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -151,10 +151,10 @@ __setup_in_out_tensors (ml_single * single_h)
  * @note this has not overhead if the allocation of output is not performed by
  * the framework but by tensor filter element.
  */
-static inline void
+static void
 set_destroy_notify (ml_single * single_h, ml_tensors_data_s * data)
 {
-  if (single_h->klass->allocate_in_invoke (single_h->filter) == TRUE) {
+  if (single_h->klass->allocate_in_invoke (single_h->filter)) {
     data->handle = single_h;
     single_h->destroy_data_list = g_list_append (single_h->destroy_data_list,
         (gpointer) data);
@@ -176,16 +176,16 @@ __destroy_notify (gpointer data_h, gpointer single_data)
     single_h->klass->destroy_notify (single_h->filter,
         (GstTensorMemory *) data->tensors);
   }
+  data->handle = NULL;
 }
 
 /**
- * @brief call the framework to destroy the allocated output data before closing
+ * @brief Wrapper for ml_tensors_data_destroy with signature of GDestroyNotify
  */
-static inline void
-__destroy_notify_free_data (gpointer data_h, gpointer single_data)
+static void
+ml_tensors_data_destroy_gwrapper (gpointer data)
 {
-  __destroy_notify (data_h, single_data);
-  g_free(data_h);
+  ml_tensors_data_destroy (data);
 }
 
 /**
@@ -839,9 +839,9 @@ ml_single_close (ml_single_h single)
 
   /** locking ensures correctness with parallel calls on close */
   if (single_h->filter) {
-    g_list_foreach (single_h->destroy_data_list, __destroy_notify_free_data,
-        single_h);
-    g_list_free (single_h->destroy_data_list);
+    g_list_foreach (single_h->destroy_data_list, __destroy_notify, single_h);
+    g_list_free_full (single_h->destroy_data_list,
+        ml_tensors_data_destroy_gwrapper);
 
     if (single_h->klass)
       single_h->klass->stop (single_h->filter);
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -111,6 +111,8 @@ typedef struct
 
   ml_tensors_data_s in_tensors;    /**< input tensor wrapper for processing */
   ml_tensors_data_s out_tensors;   /**< output tensor wrapper for processing */
+
+  GList *destroy_data_list;         /**< data to be freed by filter */
 } ml_single;
 
 /**
@@ -141,6 +143,80 @@ __setup_in_out_tensors (ml_single * single_h)
     out_tensors->tensors[i].size =
         ml_tensor_info_get_size (&single_h->out_info.info[i]);
   }
+}
+
+/**
+ * @brief setup the destroy notify for the allocated output data.
+ * @note this stores the data entry in the single list.
+ * @note this has not overhead if the allocation of output is not performed by
+ * the framework but by tensor filter element.
+ */
+static inline void
+set_destroy_notify (ml_single * single_h, ml_tensors_data_s * data)
+{
+  if (single_h->klass->allocate_in_invoke (single_h->filter) == TRUE) {
+    data->handle = single_h;
+    single_h->destroy_data_list = g_list_append (single_h->destroy_data_list,
+        (gpointer) data);
+  }
+}
+
+/**
+ * @brief To call the framework to destroy the allocated output data
+ */
+static inline void
+__destroy_notify (gpointer data_h, gpointer single_data)
+{
+  ml_single *single_h;
+  ml_tensors_data_s *data;
+
+  data = (ml_tensors_data_s *) data_h;
+  single_h = (ml_single *) single_data;
+  if (G_LIKELY (single_h->filter)) {
+    single_h->klass->destroy_notify (single_h->filter,
+        (GstTensorMemory *) data->tensors);
+  }
+}
+
+/**
+ * @brief call the framework to destroy the allocated output data before closing
+ */
+static inline void
+__destroy_notify_free_data (gpointer data_h, gpointer single_data)
+{
+  __destroy_notify (data_h, single_data);
+  g_free(data_h);
+}
+
+/**
+ * @brief Wrapper function for __destroy_notify
+ */
+int
+ml_single_destroy_notify (ml_single_h single, ml_tensors_data_s * data)
+{
+  ml_single *single_h;
+  int status = ML_ERROR_NONE;
+
+  if (G_UNLIKELY (!single || !data))
+    return ML_ERROR_INVALID_PARAMETER;
+
+  ML_SINGLE_GET_VALID_HANDLE_LOCKED (single_h, single, 0);
+
+  if (G_UNLIKELY (!single_h->filter)) {
+    status = ML_ERROR_INVALID_PARAMETER;
+    goto exit;
+  }
+
+  single_h->destroy_data_list =
+      g_list_remove (single_h->destroy_data_list, data);
+  __destroy_notify (data, single_h);
+
+exit:
+  ML_SINGLE_HANDLE_UNLOCK (single_h);
+
+  if (G_UNLIKELY (status != ML_ERROR_NONE))
+    ml_loge ("Failed to destroy the data.");
+  return status;
 }
 
 /**
@@ -180,6 +256,8 @@ __invoke (ml_single * single_h)
 static inline void
 __process_output (ml_single * single_h)
 {
+  ml_tensors_data_s *out_data;
+
   if (single_h->ignore_output == TRUE) {
     /**
      * Caller of the invoke thread has returned back with timeout
@@ -187,6 +265,9 @@ __process_output (ml_single * single_h)
      */
     ml_tensors_data_destroy (single_h->output);
     single_h->output = NULL;
+  } else {
+    out_data = (ml_tensors_data_s *) single_h->output;
+    set_destroy_notify (single_h, out_data);
   }
 }
 
@@ -501,6 +582,7 @@ ml_single_create_handle (ml_nnfw_type_e nnfw)
   single_h->thread = NULL;
   single_h->input = NULL;
   single_h->output = NULL;
+  single_h->destroy_data_list = NULL;
 
   ml_tensors_info_initialize (&single_h->in_info);
   ml_tensors_info_initialize (&single_h->out_info);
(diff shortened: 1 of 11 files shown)


kparichay:
One approach to handle freeing the data is (1), the `ml_single_close` frees the data allocated by the tensor_filter framework as well as the `ml_tensors_data_s` structure. So, in this case, once the single handle is closed, there is no need to free the data with `ml_tensors_data_destroy` as it the data in it is not available to use anyway. This is intuitive. However, misaligns with the API style where the data created by the user is not needed to be freed by the user. (this is the current implementation)

The alternative is (2) `ml_single_close` just frees the internal data allocated by the tensor_filter framework. The user must free  `ml_tensor_data_s` structure with `ml_tensors_data_destroy`. This aligns more with the API style (user creates and destroys), however, is a bit counter-intuitive for the user to free the data which is not even available after single handle has been closed.

In both the cases, as this data is not available after `ml_single_close`, the internal details about `allocate_in_invoke` are exposed to the user. The only way to avoid this exposure is to perform a `memcpy` before returning the output to the user, which will nullify any advantage of `allocate_in_invoke`.

myungjoo:
If it is fixing a memory leak of API, it is RELEASE-CRITICAL. Get attentions from reviewers; we need to review it quickly and thorughly.

api/capi/src/tensor_filter_single.c
```diff
@@ -276,6 +311,9 @@ g_tensor_filter_single_invoke (GTensorFilterSingle * self,
 
   GST_TF_FW_INVOKE_COMPAT (priv, status, input, output);
 
+  if (self->allocate_in_invoke) {
+  }
```
dongju-chae, line 315: Mistake? Otherwise, please add some comments like `TODO`

api/capi/src/nnstreamer-capi-util.c
```diff
@@ -550,15 +551,19 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
...
-    if (_data->tensors[i].tensor) {
-      g_free (_data->tensors[i].tensor);
-      _data->tensors[i].tensor = NULL;
+  if (_data->handle != NULL) {
+    status = ml_single_destroy_notify (_data->handle, _data);
+  } else {
+    for (i = 0; i < ML_TENSOR_SIZE_LIMIT; i++) {
+      if (_data->tensors[i].tensor) {
+        g_free (_data->tensors[i].tensor);
+        _data->tensors[i].tensor = NULL;
```
wooksong, line 560: This line is unnecessary. g_free does already set _data->tensors[i].tensor to NULL.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -5757,6 +5757,265 @@ TEST (nnstreamer_capi_singleshot, invoke_09_n)
...
+
+  status = ml_tensors_data_destroy (input);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+
+  g_free (test_model);
+  ml_tensors_info_destroy (in_info);
+  ml_tensors_info_destroy (out_info);
+
+  /** This will destroy twice resulting in UB */
+  // EXPECT_DEATH (ml_tensors_data_destroy (output), ".*");
```
dongju-chae, line 5841: Is this your intended comment?

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -143,6 +145,80 @@ __setup_in_out_tensors (ml_single * single_h)
   }
 }
 
+/**
+ * @brief setup the destroy notify for the allocated output data.
+ * @note this stores the data entry in the single list.
+ * @note this has not overhead if the allocation of output is not performed by
+ * the framework but by tensor filter element.
+ */
+static inline void
```
wooksong, line 154: Q: Is there any reason to make this function inline?

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -757,6 +839,10 @@ ml_single_close (ml_single_h single)
 
   /** locking ensures correctness with parallel calls on close */
   if (single_h->filter) {
+    g_list_foreach (single_h->destroy_data_list, __destroy_notify_free_data,
+        single_h);
+    g_list_free (single_h->destroy_data_list);
```
dongju-chae, line 844: Well.. I think this is unnecessary because you already freed all elements in `__destroy_notify_free_data`.

Or, if you want to keep this code style, `g_list_foreach (single_h->destroy_data_list, __destroy_notify, single_h)` is enough for your purpose, and remove unnecessary implementation of `__destroy_notify_free_data`

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -143,6 +145,80 @@ __setup_in_out_tensors (ml_single * single_h)
...
+/**
+ * @brief setup the destroy notify for the allocated output data.
+ * @note this stores the data entry in the single list.
+ * @note this has not overhead if the allocation of output is not performed by
+ * the framework but by tensor filter element.
+ */
+static inline void
+set_destroy_notify (ml_single * single_h, ml_tensors_data_s * data)
+{
+  if (single_h->klass->allocate_in_invoke (single_h->filter) == TRUE) {
```
wooksong, line 157: FYI, https://developer.gnome.org/glib/stable/glib-Basic-Types.html#gboolean

kparichay:
> If it is fixing a memory leak of API, it is RELEASE-CRITICAL. Get attentions from reviewers; we need to review it quickly and thorughly.

I thought this was to be done for the next release. So, I didn't add RELEASE-CRITICAL.
Also, note that this requires an ACR as it updates the single-API dsecription.

api/capi/src/tensor_filter_single.c
kparichay: This is a mistake. Removed.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: Yes, it is intended.
This is to emphasize that destroying twice will result in the death of code. 
However, it isn't guaranteed as this results in modifying already freed memory, which is undefined behavior.

api/capi/src/nnstreamer-capi-util.c
kparichay: I am not sure - https://github.com/bratsche/glib/blob/master/glib/gmem.c#L258

api/capi/src/nnstreamer-capi-single.c
kparichay: Yes, this is in the critical path of the `invoke` function.
Also, this is a very small function that has been moved out of `__process_output` for readability.

wooksong: If this is in the critical path as you mentioned, IMO, adding the ```inline``` keyword to all those functions related to __process_output does not guarantee the performance improvement.  It is possible to make the body in the while loop of invoke_thread () be large.

api/capi/src/nnstreamer-capi-single.c
kparichay: I am not sure if I understand correctly.
`Glist` for each of its elements (`ml_tensors_data_s` here) allocates wrapper around them.
`g_list_foreach` above is supposed to free the `ml_tensors_data_s` structure.
`g_list_free` is supposed to free the wrapper made by the glist.
So, I am not on how to eliminate `g_list_free`?

You are right that `__destroy_notify_free_data` is a bit redundant. I removed this function, and reset `data->handle` in `__destroy_notify`. This allows calling `ml_tensors_data_destroy` with `g_list_free_full` which destroys `ml_tensors_data_s` wrapper along with glist wrapper.

api/capi/src/nnstreamer-capi-single.c
kparichay: Thanks. Updated.

api/capi/src/nnstreamer-capi-single.c
kparichay: As you already know, `__proces_output` and `__invoke` are used in both `invoke_thread` and `ml_single_invoke`.
And I don't think copying the same code in both functions inside is a good choice.

wooksong: > As you already know, `__proces_output` and `__invoke` are used in both `invoke_thread` and `ml_single_invoke`.
> And I don't think copying the same code in both functions inside is a good choice.

I am not saying that this function is unnecessary. I am concerning the ```inline`` keyword only.

kparichay: I meant that given `__process_output` is inline, I didn't want to introduce a non-inline function inside it (which would be against the inline attribute of `__process_output`). Which one do you recommend -
1. don't make `set_destroy_notify`, and let that code directly be in `__process_output`
2. have `set_destroy_notify`
3. have `set_destroy_notify` with inline attribute
4. something else

wooksong: 2

kparichay: done.

api/capi/src/nnstreamer-capi-single.c
dongju-chae: Oh, my bad. Yes, the double linked list in glib contains its wrapping object.

dongju-chae (approved):
LGTM

wooksong (approved):

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -5757,6 +5757,265 @@ TEST (nnstreamer_capi_singleshot, invoke_09_n)
...
+
+  /* generate input data */
+  status = ml_tensors_data_create (in_info, &input);
+  EXPECT_EQ (status, ML_ERROR_NONE);
+  ASSERT_TRUE (input != NULL);
+
+  for (i = 0; i < 10; i++) {
+    int16_t i16 = (int16_t) (i + 1);
+
+    status = ml_tensors_data_get_tensor_data (input, 0, &data_ptr, &data_size);
```
again4you, line 5817: Is your intent? 
If not, what about moving this code out of the loop?

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -5757,6 +5757,265 @@ TEST (nnstreamer_capi_singleshot, invoke_09_n)
...
```
again4you, line 5901: ditto.

tests/tizen_capi/unittest_tizen_capi.cc
```diff
@@ -5757,6 +5757,265 @@ TEST (nnstreamer_capi_singleshot, invoke_09_n)
...
```
again4you, line 5987: ditto.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: not my intent.
moved out.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: done.

tests/tizen_capi/unittest_tizen_capi.cc
kparichay: done.

myungjoo (approved):
LGTM.

You've added a few more positive cases, where we are still required to add more negative cases.
Please add more negative cases soon.

kparichay:
> LGTM.
> 
> You've added a few more positive cases, where we are still required to add more negative cases.
> Please add more negative cases soon.

The negative test cases, in this case, is freeing the memory which has already been freed. 
I tried adding negative test cases with `EXPECT_DEATH`, however, the above scenario results in Undefined Behavior and the program doesn't always die. Can you give some hints for how to create proper negative unit tests for this?

myungjoo:
> > LGTM.
> > You've added a few more positive cases, where we are still required to add more negative cases.
> > Please add more negative cases soon.
> 
> The negative test cases, in this case, is freeing the memory which has already been freed.
> I tried adding negative test cases with `EXPECT_DEATH`, however, the above scenario results in Undefined Behavior and the program doesn't always die. Can you give some hints for how to create proper negative unit tests for this?

Random ideas:

1. test with incorrect file path (e.g., at Line 5784)
2. test with incorrect set_count along with return value testing (e.g., at Line 5791)
3. test with incorrect tensor type or dimension (e.g., at Line 5798, 5799)
4. test with incorrect flags of ml_single_open (e.g., at Line 5803)
5. Trying get_tensor_data w/ null ptr or other "detectable" incorrect arguments (e.g., at Line 5814)
6. ml_single_invoke w/ invalid arguments (incl. null ptrs) (e.g., at Line 5820)
and so on...

You already have so many points that may fail by incorrect API usages.

kparichay:
Aah, you meant to add any negative test cases related to custom filters/single API, and not specifically for automated freeing the output memory with single close with framework which has allocate_in_invoke.

PR #2687, opened 2020-09-03
[Android/NNFW] parse option

jaeyun-jung:
add new constructor and parse option string in NNFW sub-plugin.

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
@@ -54,7 +54,7 @@ public final class SingleShot implements AutoCloseable {
      * @throws IllegalStateException if this failed to construct the pipeline
      */
     public SingleShot(@NonNull File model) {
-        this(model, null, null);
+        this(new File[]{model}, null, null, NNStreamer.NNFWType.TENSORFLOW_LITE, null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -113,6 +113,46 @@ nnfw_get_accelerator (nnfw_pdata * pdata, const char *accelerators)
   }
 }
 
+/**
+ * @brief Parse accelerator and other option.
+ */
+static void
+nnfw_parse_option (const GstTensorFilterProperties * prop, nnfw_pdata * pdata)
```
kparichay, line 120: rename to `nnfw_parse_accl_custom`?

jaeyun-jung: Currently only accelerator, but I am not sure whether they would request another option for NNFW.
nnfw_parse_custom_option() ?

kparichay: nnfw_parse_custom_option() sounds good.

jaeyun-jung: updated

kparichay (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -113,6 +113,47 @@ nnfw_get_accelerator (nnfw_pdata * pdata, const char *accelerators)
...
+
+    for (i = 0; i < len; ++i) {
+      gchar **option = g_strsplit (options[i], ":", -1);
+
+      if (g_strv_length (option) > 1) {
+        g_strstrip (option[0]);
+        g_strstrip (option[1]);
+
+        if (g_ascii_strcasecmp (option[0], "Runtime") == 0) {
+          pdata->accelerator = g_strdup (option[1]);
```
dongju-chae, line 138: Q. Is there any possibility for `pdata->accelerator` to be overwritten in this for loop?

jaeyun-jung: currently no, nnfw_parse_custom_option() is always called with newly allocated nnfw_pdata ptr.

api/android/api/src/main/java/org/nnsuite/nnstreamer/SingleShot.java
```diff
@@ -54,7 +54,7 @@
      * @throws IllegalStateException if this failed to construct the pipeline
      */
     public SingleShot(@NonNull File model) {
-        this(model, null, null);
+        this(new File[]{model}, null, null, NNStreamer.NNFWType.TENSORFLOW_LITE, null);
```
myungjoo, line 57: Why TFLITE is chosen here?

jaeyun-jung: constructor `public SingleShot(@NonNull File model)` with default tf-lite now.

dongju-chae (approved):
LGTM

myungjoo:
For next releases, we will need to parse file extensions and change default engines dynamically with constructors (e.g., nnfw runtime needs to be more prioritized.) and the existence of frameworks should be checked/considered with constructor as well. (we may "cache" status of frameworks for faster constructors.)

PR #2688, opened 2020-09-03
onert(nnfw) uses 'cpu;bcq' as default backend

chunseoklee:
For now, there is no api to use bcq backend from android java
application. Thus, as a workaround, set 'cpu;bcq' as a deafult backend.

Signed-off-by: Chunseok Lee <chunseok.lee@samsung.com>

---
# [Template] PR Description

In general, github system automatically copies your commit message for your convenience.
Please remove unused part of the template after writing your own PR description with this template.
```bash
$ git commit -s filename1 filename2 ... [enter]

Summarize changes in around 50 characters or less

More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical;
various tools like `log`, `shortlog` and `rebase` can get confused 
if you run the two together.

Further paragraphs come after blank lines.

- Bullet points are okay, too
- Typically a hyphen or asterisk is used for the bullet, preceded
  by a single space, with blank lines in between, but conventions vary here.

Resolves: #123
See also: #456, #789

**Self evaluation:**

**How to evaluate:**
1. Describe how to evaluate in order to be reproduced by reviewer(s).

Add signed-off message automatically by running **$git commit -s ...** command.

$ git push origin <your_branch_name>
```

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -37,7 +37,7 @@
 #include <nnfw.h>
 
 /** backends supported by nnfw */
-#define NNFW_CPU_BACKEND  "cpu"
+#define NNFW_CPU_BACKEND  "cpu;bcq"
 #define NNFW_GPU_BACKEND  "acl_cl"
 #define NNFW_NEON_BACKEND "acl_neon"
 #define NNFW_SRCN_BACKEND  "srcn"


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -37,7 +37,7 @@
 #include <nnfw.h>
 
 /** backends supported by nnfw */
-#define NNFW_CPU_BACKEND  "cpu"
+#define NNFW_CPU_BACKEND  "cpu;bcq"
```
kparichay, line 40: This does change the default backend but I recommend to create a NNFW_CPU_BCQ_BACKEND here and use it.
Further, register a ACCL_CPU_BCQ_STR and ACCL_CPU_BCQ at `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h` and then update the auto/default backend at line 55 to 60 for various architectures.

jaeyun-jung:
Discussion

@chunseoklee 
IMO, this is ok if `cpu;bcq` on NNFW determines backend,
1. if nnfw-boost lib exists, run on boost
2. if nnfw-boost is not available, run on cpu

Also, please check the patch https://github.com/nnstreamer/nnstreamer/pull/2687
This PR is to pass option string to NNFW. (e.g. custom str = backend:cpu;bcq)

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
myungjoo: Until we get it updated, let's use either:
```
#define NNFW_CPU_BACKEND  "cpu;bcq"
```
or
```
#define NNFW_DEFAULT_BACKEND NNFW_CPU_BACKEND ";bcq"
```
Let's not touch the API related headers at this moment (release imminent)

myungjoo (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
myungjoo: Besides, as long as bcq enabled just yet another algorithm using CPU instructions, having "cpu;bcq" as CPU_BACKEND  isn't terribly wrong.

leemgs:
[Note]  Dongsoo Lee has made the kernel (BCQ) to run a BIQGEMM-based network model. "BCQ" means Bitwise Compressed Quantization.

PR #2705, opened 2020-09-09
[Android] reduce method call

jaeyun-jung:
While converting tensor data, reduce method call to set each tensor.


anyj0527 (approved):
Verified on S20 with SNPE App (java api and native api)

myungjoo (approved):
LGTM

myungjoo:
You will need to show the performance differences.

PR #2706, opened 2020-09-09
[TF-lite] remove nnfw-delegate feature

jaeyun-jung:
NNFW delegate feature is unnecessary.
NNFW supports method to set backend and no more need to set NNFW-delegation in tf-lite plugin.


myungjoo (approved):
LGTM

jaeyun-jung:
2020-09-09 PR updated
NNFW-delegate is unnecessary in tf-lite subplugin.
Remove this feature.

kparichay:
Does this mean that tflite subplugin does not support nnapi backend anymore or nnapi backend is enabled by default with the tflite version nnstreamer uses? If its the latter case, can we make some documentation/comment about this.

again4you (approved):
Nice catch!

myungjoo:
I'm adding for-next because this does not improve the performance of target devices and does not fix bugs affecting behaviors.
If this is actually required by 1.6 releease, please let me know.

jaeyun-jung:
> Does this mean that tflite subplugin does not support nnapi backend anymore or nnapi backend is enabled by default with the tflite version nnstreamer uses? If its the latter case, can we make some documentation/comment about this.

@kparichay ENABLE_TFLITE_NNAPI_DELEGATE feature is for NNFW runtime, not tf-lite.

Issue #2611, opened 2020-08-03
[Notice] Release Branch Policy (UPDATED 2020-10-08)


myungjoo:
### Merging Branches
- main
    - 1.7.x development. Both features and fixes

For major product releases, I'll maintain release branches

### Other Branches

1. ~~for-next~~ (reactivated when code-freeze is activated)
2. lts/1.0.y : Tizen 5.5 Mx long-term-stable
3. lts/1.6.y : Tizen 6.0 Mx and Android-21-products long-term stable
    - Maintainer: @again4you (Tizen fixes) / @jaeyun-jung (Android fixes)

myungjoo:
Closing: this notice is obsolete — it describes main as 1.7.x development with lts/1.0.y and lts/1.6.y, while the current version is 2.7.0 with LTS branches through lts/2.6.y. A refreshed release-branch policy notice can be posted separately if needed.

PR #2714, opened 2020-09-10
[ext/tensor_decoder] Configure framerate caps for all tensor_decoders

kbumsik:
This patch add framerate caps for in all tensor_decoder modes.

While I was testing the gst-launch examples, I noticed that framerate
caps are dropped by tensor_decoder elements.

There are two problems I can think of:

1. What is the right framerate policies for tensor_decoder? (1)
2. Framerate policies are inconsistent across the tensor_decoder modes.
Some passthrough framerate caps (e.g. tensordec-directvideo.c)
but the others drop framerate caps.

This patch tries to solve at least question 2 by letting them passthrough. I keep @todo tags
regarding question 1 (1) because I do not know how to verify if this policy is correct, though
I think this would be enough.

I made a function in tensordecutil.c to share the same code.

(1): https://github.com/nnstreamer/nnstreamer/blob/a1c9b45d1ac1f3a7b830df2fa17b962d88a128cf/ext/nnstreamer/tensor_decoder/tensordec-imagesegment.c#L303

Signed-off-by: Bumsik Kim <k.bumsik@gmail.com>

**Self evaluation:**

**How to evaluate:**
Run test: looking at how framerate caps in tensordec src pads changes while running nnstreamer-examples with `export GST_DEBUG="tensor_decoder:6`

---
This change is [Reviewable](https://reviewable.io/reviews/nnstreamer/nnstreamer/2714)


ext/nnstreamer/tensor_decoder/tensordecutil.h
```diff
@@ -36,6 +38,8 @@ initSingleLineSprite (singleLineSprite_t v, rasters_t r, uint32_t pv);
 
 extern void _free_labels (imglabel_t *data);
 
+extern void capsConfigFrameRate (GstCaps *caps, const GstTensorsConfig * config);
```
myungjoo, line 41: Please rename the function for readability:

```setFramerateFromConfig ( caps, config )```

File: ext/nnstreamer/tensor_decoder/tensordecutil.c [hunk 2]
```diff
@@ -116,3 +117,25 @@ _free_labels (imglabel_t * data)
...
+  gint fn, fd;
+
+  /** @todo Verify if this rate is ok */
+  fn = config->rate_n;
+  fd = config->rate_d;
+  if (!(fn >= 0 && fd > 0)) {
+    fn = 0;
+    fd = 1;
+  }
+  gst_caps_set_simple (caps, "framerate", GST_TYPE_FRACTION, fn, fd, NULL);
```
myungjoo, line 139: If there is no "valid" framerate in config (i.e., the input caps has no framerate definition), why don't we keep it unconfigured?

myungjoo:
LGTM other than the two points commented.

As we want to merge this to for-next (1.7.0), not 1.6.0, could you please rebase yours so that the three unwanted commits are removed? (they are created by auto-rebase to for-next)

ext/nnstreamer/tensor_decoder/tensordecutil.c
kbumsik: I actually wanted to hear any thoughts about this. I thought it might be safer to keep framerate caps without any particular reason. I will drop framerate caps if you think it's better.

ext/nnstreamer/tensor_decoder/tensordecutil.h
kbumsik: Thanks for a better name :) I will change it.

kbumsik:
Thanks for a quick review. I fixed them and rebased to `for-next` branch.

ext/nnstreamer/tensor_decoder/tensordecutil.c
myungjoo: If there was no framerate caps in input, decoder should not make framerate caps in output.
If there was framerate caps in input, decoder should make the same framerate caps in output.

If there was framerate caps in output (reverse is possible sometiems), decoder should make the same framerate caps in input if it is trying to caps in reverse order.

kbumsik: Thanks for explantion! I fixed it as you said. I probably need to get more experience in gstreamer :)

dongju-chae (approved):
LGTM!

dongju-chae:
> Thanks for a quick review. I fixed them and rebased it to for-next branch. So, is for-next the default development branch for most cases?


Right, please refer to https://github.com/nnstreamer/nnstreamer/issues/2611

kbumsik:
@myungjoo It seems that my PRs have unwanted commits after force-pushed in upstream. Should I rebase them accordingly?

gichan-jang (approved):
LGTM

myungjoo (approved):

PR #2738, opened 2020-09-15
[Build/TensorRT] Add TensorRT dependency

dongju-chae:
This patch adds the build dependency for TensorRT.

CC: @again4you 

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -186,6 +186,36 @@ if not get_option('snpe-support').disabled()
   endif
 endif
 
+# tensorrt
+if not get_option('tensorrt-support').disabled()
+  # check available cuda versions (11.0 and 10.2 are recommended)
+  cuda_vers = [
+    '11.0',
+    '10.2',
+    '10.1',
+    '10.0',
+    '9.2',
+    '9.1',
+    '9.0'
+  ]
+
+  foreach ver : cuda_vers
+    cuda_dep = dependency('cuda-' + ver, required: false)
+    cudart_dep = dependency('cudart-' + ver, required: false)
+    if cuda_dep.found() and cudart_dep.found()
+      if ver != '11.0' and ver != '10.2'
+        message('Warning: the recommended cuda version is at least 10.2')
+      endif
+      break
+    endif
+  endforeach
+
+  nvinfer_lib = cxx.find_library('nvinfer', required: false)
+  nvinfer_dep = declare_dependency(dependencies: nvinfer_lib)
+
+  tensorrt_deps = [ nvinfer_dep, cuda_dep, cudart_dep ]
+endif
+
 # features registration to be controlled
 #
 # register feature as follows
@@ -245,6 +275,9 @@ features = {
   },
   'protobuf-support': {
     'extra_deps': [ pb_comp ]
+  },
+  'tensorrt-support': {
+    'extra_deps': tensorrt_deps
   }
 }
 
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -15,6 +15,7 @@ option('orcc-support', type: 'feature', value: 'auto')
 option('snpe-support', type: 'feature', value: 'auto')
 option('protobuf-support', type: 'feature', value: 'auto')
 option('flatbuf-support', type: 'feature', value: 'auto')
+option('tensorrt-support', type: 'feature', value: 'auto')
 
 # booleans & other options
 option('enable-test', type: 'boolean', value: true)


meson.build
```diff
@@ -186,6 +186,28 @@ if not get_option('snpe-support').disabled()
   endif
 endif
 
+# tensorrt
+tensorrt_dep = dependency('', required: false)
```
wooksong, line 190: Is this effective line?

zhoonit: I think a discussion made at https://github.com/nnstreamer/nntrainer/pull/150#discussion_r439172394 might help this line.

meson.build
```diff
@@ -245,6 +267,9 @@ features = {
   },
   'protobuf-support': {
     'extra_deps': [ pb_comp ]
+  },
+  'tensorrt-support': {
+    'extra_deps': tensorrt_dep
```
myungjoo, line 272: It's ```extra_deps```, not ```extra_dep```; thus, ```[ tensorrt_dep ]``` will look more beautiful :)

meson.build
```diff
@@ -186,6 +186,28 @@ if not get_option('snpe-support').disabled()
...
+  cuda_vers = [
+    '11.0',
+    '10.2',
+  ]
+
+  foreach ver : cuda_vers
+    cuda_dep = dependency('cuda-' + ver, required: false)
+    cudart_dep = dependency('cudart-' + ver, required: false)
+    if cuda_dep.found() and cudart_dep.found()
+      break
```
zhoonit, line 201: is this okay if they are not found for all `cuda_vers` ?

zhoonit: self answered: yes. it is handled at line 269

meson.build
```diff
@@ -186,6 +186,28 @@ if not get_option('snpe-support').disabled()
   endif
 endif
 
+# tensorrt
+tensorrt_dep = dependency('', required: false)
+if not get_option('tensorrt-support').disabled()
+  cuda_vers = [
+    '11.0',
+    '10.2',
+  ]
```
myungjoo, line 195: Do you need to restrict CUDA versions?

meson.build
dongju-chae: Ah, `tensorrt_dep` is actually 'deps'. I will rename it.

meson.build
wooksong: @zhoonit I could not still get it. 
```suggestion
tensorrt_dep = disabler()
```

Do you mean this?

meson.build
```diff
@@ -186,6 +186,28 @@ if not get_option('snpe-support').disabled()
...
+    cudart_dep = dependency('cudart-' + ver, required: false)
+    if cuda_dep.found() and cudart_dep.found()
+      break
+    endif
+  endforeach
+
+  nvinfer_lib = cxx.find_library('nvinfer', required: false)
+  nvinfer_dep = declare_dependency(dependencies: nvinfer_lib)
+
+  tensorrt_dep = [ nvinfer_dep, cuda_dep, cudart_dep ]
```
wooksong, line 208: wih
```
tensorrt_dep = disabler()
```
How about this?
```suggestion
  tensorrt_dep = declare_dependency(
      ependencies: [ nvinfer_dep, cuda_dep, cudart_dep ]
  )
```

dongju-chae: Looks good!

meson.build
dongju-chae: I'm not sure.. but as described in https://github.com/NVIDIA/TensorRT, it seems that cuda-11.0 and cuda-10.2 are recommended.

meson.build
dongju-chae: That's right, updated.

meson.build
zhoonit: I am not sure if a disabler object acts like a dependency object as the reference states that the disabler object has only one method: `found()`. If it is fine to pass a disabler object to a `build_target()`, it is equivalent.

wooksong: @zhoonit If tensorrt_dep.fount() is false, why do we provide this tensorrt_dep to build_target()?

zhoonit: I don't know the exact use case only seeing this line. So I am not sure if `disabler()` can substitute `dependency('', required:false)`

What I know is `found()` is required in this file which makes this line effective.

Although I can think of a usecase for passing not_found to the build_target

```meson
A = dependency('foo', required: false)
B = executable('bar', dependencies: A, fallback: [......])
```
or

```meson
nnfw_dep = dependency('', required: false)
if not get_option('nnfw-runtime-support').disabled()
  nnfw_dep = dependency('nnfw', required: false)
  if not nnfw_dep.found()
    nnfw_dep = cc.find_library('nnfw-dev', required: get_option('nnfw-runtime-support'))
  endif
endif
```
and adding `nnfw_dep` to a `build_target()` will safely throw when needed

meson.build
myungjoo: 1. Such info can be added as a comment.
2. Need some warning messages if CUDA is installed but the version is not supported (or not recommended), then.

meson.build
wooksong: @zhoonit
I hardly think so.
https://mesonbuild.com/Disabler.html

https://git.sailfishos.org/mirror/rygel/blob/134eb29fa1994ca2e97617579a28de0fb69d39bf/meson.build

disabler() is introduced for reducing such fragmented if-else-endif.

dongju-chae: I think `disabler()` looks good. BTW, I have modified this PR to use `tensorrt_deps` not `tensorrt_dep`.

meson.build
dongju-chae: Got it, updated.

meson.build
zhoonit: Yes I think `disabler()` can replace the given line :)

meson.build
```diff
@@ -186,6 +186,36 @@ if not get_option('snpe-support').disabled()
...
+if not get_option('tensorrt-support').disabled()
+  # check available cuda versions (11.0 and 10.2 are recommended)
+  cuda_vers = [
+    '11.0',
+    '10.2',
+    '10.1',
+    '10.0',
+    '9.2',
+    '9.1',
+    '9.0'
```
kparichay, line 199: - as far as I know, if the tensorRT was compiled with a different version of cuda and cudnn, even though a different version of cuda/cudnn is available, it won't work (this is from what I experienced an year ago).
if this happens currently as well, I am not sure if these cuda version versions will help unless there is the specific cuda version check.
- as long as version checks are being added, cudnn version might also be helpful

dongju-chae: That's right. Any installed TensorRT expects the exact version of cuda/cudnn. I will change the PR.

dongju-chae: Hmm... Unfortunately, it's really hard (impossible) to find the corresponding cuda version from libnvinfer. Do you have idea?

kparichay: I cant think of any way either.

kparichay (approved):
LGTM

wooksong (approved):
LGTM

myungjoo (approved):

Issue #1703, opened 2019-09-25
[New Element] Handle if statement nicely.


myungjoo:
Without adding yet another custom filters wrapping others, if would be much better if we can add an "if" statement based on tensor values as a filter.

V1. E.g., (version 1 supports .so only)
```
... ! tensor_if skipif="user_code_with_tensor_if_API_bool.so" ! ...
... ! tensor_if outputselect="user_code_with_tensor_if_API_int.so" name=output  output.0 ! ... output.1 ! ...
```

V2. E.g., (version 2 supports python code)
```
... ! tensor_if skipif="${python code with given API}" ! ...
... ! tensor_if skipif="user_code_with_tensor_if_API_bool.py" ! ...
```
or (version 2 supports some scripts)
```
... ! tensor_if skipif="scripting with what we've defined..." ! ...
```

Idea 20191001: the whole "if-then-else" clauses might need to be enclosed in a "bin"... maybe... maybe not...

If we are using multi tensor with conditional variable, then it could be as below.
![image](https://user-images.githubusercontent.com/23226860/66291974-13ae6e00-e91e-11e9-8457-b9a37c98e8ff.png)

jijoongmoon:
I think we might need the end-if element. if there is separate pipeline, then it is ok without end-if element, but these separate pipeline should be merged into one stream, then we might need end-if element. We could use merge or mux but these are going to wait to the other stream is arrived and it will be stall. Or we could add options in the merge or mux which allow push down stream if one of stream data arrived. 
 
![image](https://user-images.githubusercontent.com/23226860/66355534-9ed14780-e9a2-11e9-8ad4-b334b5ad6c55.png)

myungjoo:
CC: @leemgs  (refer to the code at /gst/nnstreamer/tensor-if )

gichan-jang:
todo list
 1. custom callback function for user-defined condition
 2. determine the srcpads at the negotiation phase
 3. Find a way to connect with the other element. (e.g, input-selector)
  - Or one src pad allow a true / false action? So it can operate as input-selector.
    -> in this case, Allow dynamic configuration? or just allow the same configuration only.
 4. tensor_if control with meta info 
 5. property update during pipeline is on running or paused state. (Only options that do not affect cap, like supported value? )

myungjoo:
This is going in a good direction! Great!

myungjoo:
@gichan-jang I think you've completed this. Please close it.

[...]

PR #2745, opened 2020-09-17
[TensorIF] Add README.md for tensor_if element @open sesame 9/24 15:25

gichan-jang:
Describe for tensor_if definition, its properties and usage cases.
Related issue: #1703 

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_if/README.md
+++ b/gst/nnstreamer/tensor_if/README.md
@@ -1,0 +1,154 @@
+# NNStreamer::tensor\_if
+
+## Supported features
+
+GstTensorIF controls its src-pad based on the values (other/tensor(s)) of its sink-pad.
+
+For example, you may skip frames if there is no object detected with high confidence.
+
+The input/output stream data type is either ```other/tensor``` or ```other/tensors```.
+
+## Properties
+
+- compared_value: Compared Value (CV), operand 1 (from input tensor(s))
+  * A_VALUE:  Decide based on a single scalar value
+  * TENSOR_TOTAL_VALUE: Decide based on a total (sum) value of a specific tensor
+  * ALL_TENSORS_TOTAL_VALUE:  Decide based on a total (sum) value of tensors or a specific tensor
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_if/README.md
```diff
@@ -0,0 +1,130 @@
+# NNStreamer::tensor\_if
+
+## Supported features
+
+GstTensorIF control controls its src-pad based on the values (other/tensor(s)) of its sink-pad.
```
dongju-chae, line 5: typo?

gichan-jang: updated.

myungjoo:
For the documentation. if you have some ideas or remember PL/compiler classes (about parsers), you may suggest more user-friendly properties.

However, for implementation, the first version should implement the most basic (and w/ immediate needs) feature.

gichan-jang:
> For the documentation. if you have some ideas or remember PL/compiler classes (about parsers), you may suggest more user-friendly properties.
> 
> However, for implementation, the first version should implement the most basic (and w/ immediate needs) feature.

Yes. I will start with basic features first. (Passthrough and skip)
I'm considering the structure for feature expansion, should tensor_if element support multiple source pads?
For example, If "then"(true action) picks tensor1, 3 and "else"(false action) picks tensor2, tensor_if should provide multiple source pads. In this case, GstBaseTransform is not support multiple pads. 
So if we need this usecase, I will start writing code without GstBaseTransform.

myungjoo:
> So if we need this usecase, I will start writing code without GstBaseTransform.

Things to consider:
1. can we support them w/ tensor-mux + tensor-if ?
2. how are you going to support dynamic caps? (caps of srcpads will change dynamically according to valus in sinkpad.. would it be manageable? or would it perform highly enough?)

dongju-chae (approved):
LGTM

gichan-jang:
It is inefficient to calculate the sum or average each time for compared value.
Is it better to set the period to calculate?
e,g,. 1 time per second. or 1 timer per 30 frames.

myungjoo (approved):

PR #2763, opened 2020-09-23
[filter/SNPE] Support multiple output layers

anyj0527:
- Add missing doxygen tags
- Existing snpe filter only use the final layer of the given model.
- This patch makes the filter support multiple output layers.

There are two ways of set output layer:
1.  Use custom property
```
"appsrc name=srcx ! " +
"other/tensor,dimension=(string)3:300:300:1,type=(string)float32,framerate=(fraction)0/1 ! " +
"tensor_filter framework=snpe " + "model=" + model.getAbsolutePath() +
" custom=OutputLayer:concat;concat_1 ! " +
"tensor_sink name=sinkx";
```

2. Use tensor info
```
"appsrc name=srcx ! " +
"other/tensor,dimension=(string)3:300:300:1,type=(string)float32,framerate=(fraction)0/1 ! " +
"tensor_filter framework=snpe " + "model=" + model.getAbsolutePath() +
" output=91:1917:1:1,4:1:1917:1 outputtype=float32,float32 outputname=concat,concat_1 ! " +
"tensor_sink name=sinkx";
```

REF: https://developer.qualcomm.com/docs/snpe/group__c__plus__plus__apis.html#a5aa93979416b17df898cb0c6f8425461

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
@@ -76,6 +76,7 @@ class snpe_subplugin final : public tensor_filter_subplugin
   /* options for snpe builder */
   zdl::DlSystem::RuntimeList runtime_list;
   bool use_cpu_fallback;
+  std::vector<std::string> output_layer_names;
 
   std::unique_ptr<zdl::DlContainer::IDlContainer> container;
   std::unique_ptr<zdl::SNPE::SNPE> snpe;
@@ -90,6 +91,7 @@ class snpe_subplugin final : public tensor_filter_subplugin
   void cleanup ();
   bool configure_option (const GstTensorFilterProperties *prop);
   bool parse_custom_prop (const char *custom_prop);
+  bool set_output_layer_names (const GstTensorsInfo *info);
   static void setTensorProp (GstTensorsInfo &tensor_meta, zdl::DlSystem::TensorMap &tensor_map);
   static const char *runtimeToString (zdl::DlSystem::Runtime_t runtime);
 
@@ -115,7 +117,8 @@ const char *snpe_subplugin::name = "snpe";
  */
 snpe_subplugin::snpe_subplugin ()
     : tensor_filter_subplugin (), empty_model (true), model_path (nullptr),
-    snpe (nullptr)
+      runtime_list (zdl::DlSystem::Runtime_t::CPU), use_cpu_fallback (false),
+      container (nullptr), snpe (nullptr)
 {
   inputInfo.num_tensors = 0;
   outputInfo.num_tensors = 0;
@@ -184,7 +187,8 @@ snpe_subplugin::getEmptyInstance ()
 /**
  * @brief Method to get string of SNPE runtime.
  */
-const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
+const char *
+snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
 {
   switch (runtime) {
   case zdl::DlSystem::Runtime_t::CPU:
@@ -198,6 +202,25 @@ const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
   default:
     return "invalid_runtime...";
   }
+}
+
+/**
+ * @brief Internal method to get names of output layers from tensors information.
+ */
+bool
+snpe_subplugin::set_output_layer_names (const GstTensorsInfo *info)
+{
+  for (unsigned int i = 0; i < info->num_tensors; ++i) {
+    if (info->info[i].name == nullptr) {
+      /* failed */
+      nns_loge ("Given layer name with index %u is invalid.", i);
+      output_layer_names.clear ();
+      return false;
+    }
+    nns_logd ("Add output layer name of %s", info->info[i].name);
+    output_layer_names.emplace_back (std::string (info->info[i].name));
+  }
+  return true;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -184,6 +204,22 @@ const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
...
 }
 
+/**
+ * @brief Internal method to get names of output layers from tensors information.
+ */
+bool snpe_subplugin::set_output_layer_names (const GstTensorsInfo *info)
+{
+  for (unsigned int i = 0; i < info->num_tensors; ++i) {
+    if (info->info[i].name == nullptr) {
+      /* failed */
```
jaeyun-jung, line 214: let's consider both 'custom' and 'outputname'.
info[i].name is null, clear `output_layer_names` and return (Using single-shot Java API always returns failure.)

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -184,6 +204,22 @@ const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
...
+      nns_loge ("Given layer name with index %u is invalid.", i);
+    }
+    nns_logd ("Add output layer name of %s", info->info[i].name);
+    output_layer_names.append (info->info[i].name);
```
dongju-chae, line 218: Q: Is it okay to append `nullptr`?

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -184,6 +204,22 @@ const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
...
+ * @brief Internal method to get names of output layers from tensors information.
+ */
+bool snpe_subplugin::set_output_layer_names (const GstTensorsInfo *info)
+{
+  for (unsigned int i = 0; i < info->num_tensors; ++i) {
+    if (info->info[i].name == nullptr) {
+      /* failed */
+      nns_loge ("Given layer name with index %u is invalid.", i);
+    }
+    nns_logd ("Add output layer name of %s", info->info[i].name);
```
dongju-chae, line 217: Let's don't print a name if it's null.

anyj0527: Updated! Please check

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
anyj0527: I replace the type of output_layer_names to `std::vector<std::string>`, because the `zdl::DlSystem::StringList` does not open `clear` method.
Please check!

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -238,6 +276,19 @@ bool snpe_subplugin::parse_custom_prop (const char *custom_prop)
           nns_loge ("Unknown cpu_fallback option");
           invalid_option = true;
         }
+      } else if (g_ascii_strcasecmp (option[0], "OutputLayer") == 0) {
+        gchar **names = g_strsplit (option[1], ";", -1);
+        for (guint i = 0; i < g_strv_length (names); ++i) {
```
wooksong, line 281:
```suggestion
        guint len_str_arr = g_strv_length (names)
        for (guint i = 0; i <len_str_arr; ++i) {
```

It is better to use a local variable here.

anyj0527: I followed your suggestion. Thanks!

dongju-chae (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -202,6 +204,24 @@ const char * snpe_subplugin::runtimeToString (zdl::DlSystem::Runtime_t runtime)
   }
 }
 
+/**
+ * @brief Internal method to get names of output layers from tensors information.
+ */
+bool snpe_subplugin::set_output_layer_names (const GstTensorsInfo *info)
+{
+  for (unsigned int i = 0; i < info->num_tensors; ++i) {
```
myungjoo, line 212: If layers are NOT tensors, why are you using "num_tensors" and use GstTensorsInfo's tensor names for layer names?

anyj0527: The example for convert tensorflow model into dlc model:
```
snpe-tensorflow-to-dlc --input_network ssdlite_mobilenet_v2.pb -d Preprocessor/sub "1,300,300,3" --out_node "concat" --out_node "concat_1" --output_path ssdlite_mobilenet_v2.dlc --allow_unconsumed_nodes
```
In this case, two `out_node`s have LAYER names of "concat" and "concat_1" and each LAYER has "concat:0" and "concat_1:0" as its TENSOR names.
I think the user only deal with LAYER, not with TENSOR, so LAYER could be compatible with TensorInfo.

myungjoo: If you have output layers "layerA" and "layerB" and layerA has two tensors and layerB has three tensors, how are you going to handle?

myungjoo: If this works well with multiple multi-tensor layers, it's good to me.

anyj0527: There could be a model which contains some output nodes that have multiple meaningful output tensors, but not that I know of. Let me test with some self-made models.

anyj0527: @myungjoo 
After some investigation, I found that:

In every TensorFlow model I checked, all output layers have a single output tensor. This could be also checked with the famous keras model build api `Sequential.add`.
It throws `ValueError` in case the layer argument has multiple output tensors (https://www.tensorflow.org/api_docs/python/tf/keras/Sequential#add)

So I think the multi-tensor output layer feature could be supported later with some other PR.

myungjoo (approved):
LGTM.

@anyj0527 Write an ISSUE on what's to be done later.

anyj0527:
UPDATED: I added two testcases

wooksong (approved):
LGTM
@jaeyun-jung Please merge this PR if https://github.com/nnstreamer/nnstreamer/pull/2763/files#r493377086 is resolved.

Issue #2766, opened 2020-09-24
Tensor-Rate (framerate control)


myungjoo:
Control framerate of tensor streams.

1. can we control this only with GST-CAP (capsfilter) and let tensor-filter handle frame rate?
2. do we need yet another gst-filter "tensor-rate" that does something similar with "videorate"?

Required by a client.

Depending on the study/survey results,
1. implement required elements
2. provide test cases and applications
3. the implement mechanism needs to focus on minimizing overheads. (the objective of rate limit is to reduce overheads)

dongju-chae:
I'm not sure but I think that an extra plugin, `tensor-rate`, looks better than internal capability of `tensor-filter` if we think its wide use. We have other plugins using tensors and they also get some benefits from this feature.

myungjoo:
Please do some testing with prototype codes and sample apps including this scenario:
```
... A - q -+-> C
... B - q -+
```
(q: test all combinations with queue and without queue)

If C is at 5 fps, are A and B throttled down to 5 fps?
If C is at 60 fps and A is at 5 fps, what's B if untouched?
If A is at 5 fps and B is at 10 fps configured, what's C's behavior?
(make as many scenarios as you can imagine, find out what would be optimal and what would be alternatives or should be configured by users.)

After getting as many scenarios as possible based on your imaginations, create specifications. You will need some prototype/PoC code for tensor-filter or custom-easy subplugin for testing rate-limiting.

dongju-chae:
Thanks for the guidance . Got it.

jaeyun-jung:
FYI
tensor_sink property `signal-rate` : new data signal per seconds

myungjoo:
Ultimately, this will be related with QoS / Latency Control! (not a trivial problem!)

PR #2769, opened 2020-09-24
[TensorRate] Implement the prototype codes for tensor_rate element @open sesame 10/12 12:19

dongju-chae:
This patch implements the prototype codes for `tensor_rate` element.

`tensor_rate` will include two methods to adjust a frame rate.
1) Either drop or duplicate frames similar to `videorate` plugin
2) Propagate a specified frame-rate to upstream elements.

As a prototype, most of codes are reused from `videorate` sources codes,
and still many things should be revised according to the tensor concept.

But, at least, its functionality was checked if we run this pipeline.

```                      
gst-launch-1.0 videotestsrc                                            
   ! video/x-raw,width=640,height=480,framerate=15/1                   
   ! tensor_converter                                                   
   ! tensor_rate framerate=10/1  
   ! tensor_decoder mode=direct_video                                  
   ! videoconvert                                                      
   ! autovideosink                                                     
```

Related issue: https://github.com/nnstreamer/nnstreamer/issues/2766

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_rate/gsttensorrate.c
+++ b/gst/nnstreamer/tensor_rate/gsttensorrate.c
@@ -1,0 +1,998 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * GStreamer/NNStreamer Tensor-Rate
+ * Copyright (C) 2020 Dongju Chae <dongju.chae@samsung.com>
+ */
+
+/**
+ * @file    gsttensorrate.c
+ * @date    24 Sep 2020
+ * @brief   GStreamer plugin to adjust tensor rate
+ * @see     https://github.com/nnstreamer/nnstreamer
+ * @author  Dongju Chae <dongju.chae@samsung.com>
+ * @bug     No known bugs except for NYI items
+ */
+
+/**
+ * SECTION:element-tensor_rate
+ *
+ * This element controls a frame rate of tensor streams in the pipeline.
+ *
+ * Basically, this element takes an incoming stream of tensor frames, and
+ * produces an adjusted stream that matches the source pad's framerate.
+ * The adjustment is performed by dropping and duplicating tensor frames.
+ * By default the element will simply negotiate the same framerate on its
+ * source and sink pad.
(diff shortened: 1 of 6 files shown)


gst/nnstreamer/tensor_rate/gsttensorrate.c
```diff
@@ -0,0 +1,927 @@
...
+  gst_element_class_add_pad_template (gstelement_class,
+      gst_static_pad_template_get (&src_factory));
+  gst_element_class_add_pad_template (gstelement_class,
+      gst_static_pad_template_get (&sink_factory));
+
+  trans_class->passthrough_on_same_caps = TRUE;
+  trans_class->transform_ip_on_passthrough = FALSE;
+
+  /* Processing units */
+  trans_class->transform_ip = GST_DEBUG_FUNCPTR (gst_tensor_rate_transform_ip);
```
myungjoo, line 202: You use transform_ip only if you touch the contents. Refer to https://gstreamer.freedesktop.org/documentation/base/gstbasetransform.html?gi-language=c#modifications-only-to-the-capsmetadata-of-a-buffer

dongju-chae: IMO, when a framerate is changed, this element will think caps are also different and run on non-passthrough mode.
In this case, `transform` will make a unnecessary buffer copy. I think that's why existing rate-control elements (e.g., [videorate](https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/blob/8ecc3b97307e3a47dc19a024de117cac68ff7d7b/gst/videorate/gstvideorate.c#L1286), [videomaxrate](https://code.cablelabs.com/DLNA-RUI/gst-plugins-bad/-/blob/de138dbc57c5928a25e10d440ff2bb0ac6ae4b4d/gst/videomaxrate/videomaxrate.c#L275)) use `transform-ip`.

gichan-jang (approved):
LGTM

helloahn (approved):
LGTM

PR #2780, opened 2020-09-29
[Ext/Filter] Set verify_model_path as TRUE for TF, PyTorch, and Python sub-plugins

wooksong:
This PR is about setting verify_model_path as TRUE to delegate the given model files' verification to the NNS common framework.
If such verification code is included in the tensor_filter sub-plugin, it is also removed from the sub-plugin.

**Self evaluation:**


dongju-chae (approved):
LGTM

myungjoo (approved):
LGTM

PR #2788, opened 2020-10-06
[Filter/NNFW] dummy invoke after set-info called

jaeyun-jung:
When called set-info function, NNFW updates output shape after the invoke is done.
Add dummy invoke function and call it after input shape is changed.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -88,6 +88,9 @@ static void nnfw_close (const GstTensorFilterProperties * prop,
     void **private_data);
 static int nnfw_tensors_info_get (const nnfw_pdata * pdata,
     const gboolean is_input, GstTensorsInfo * info, NNFW_TYPE * type);
+static int nnfw_invoke (const GstTensorFilterProperties * prop,
+    void **private_data, const GstTensorMemory * input,
+    GstTensorMemory * output);
 
 /**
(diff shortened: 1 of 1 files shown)


jaeyun-jung:
about the output buffer size,
https://github.com/Samsung/ONE/blob/master/runtime/onert/api/include/nnfw.h#L329

jaeyun-jung:
fix dynamic input shape in ONE
https://github.com/Samsung/ONE/pull/4512

myungjoo (approved):

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -514,6 +517,43 @@ nnfw_getOutputDim (const GstTensorFilterProperties * prop,
...
+    for (i = 0; i < out_info->num_tensors; ++i) {
+      output[i].size *= 2;
+      output[i].data = g_realloc (output[i].data, output[i].size);
+    }
+  }
+
+  for (i = 0; i < in_info->num_tensors; ++i)
+    g_free (input[i].data);
+  for (i = 0; i < out_info->num_tensors; ++i)
+    g_free (output[i].data);
```
dongju-chae, line 554: How about setting buffers to nullptr to prevent any possible side-effects?

jaeyun-jung: updated

helloahn (approved):
LGTM

PR #2789, opened 2020-10-07
[TensorRate/Filter] Skip inference when QoS events arrive

dongju-chae:
This patch implements a simple QoS mechanism which skips
inference when QoS events arrive from tensor_rate.      

The below shows test results using tflite object detection example

- Without throttling (original pipeline, eventually saturated 10 FPS due to tensor filter's performance)
```
...
  ! queue leaky=2 max-size-buffers=2 ! videoscale ! video/x-raw,width=640,height=480 ! tensor_converter
  ! tensor_transform mode=arithmetic option=typecast:float32,add:-127.5,div:127.5
  ! tensor_filter framework=tensorflow-lite model=...
...
```
without throttling

- With throttling (`tensor_rate` added, and no `max-size-buffer` setting, we want to limit framerate to 5 FPS)
```
...
  ! queue ! videoscale ! video/x-raw,width=640,height=480 ! tensor_converter
  ! tensor_transform mode=arithmetic option=typecast:float32,add:-127.5,div:127.5
  ! tensor_filter framework=tensorflow-lite model=...
  ! tensor_rate framerate=5/1 throttle=true
...
```

with throttling

In short, even without limiting queue's max buffers, we can adjust a frame rate to 5 FPS, and effectively reduces CPU consumption.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -48,6 +48,19 @@
  *
  * If input is other/tensor C array input[1][224][224][3] and
  * output is other/tensor C array output[1][1][1][1000]
+ *
+ * The current QoS policy: In a nnstreamer pipeline, the QoS is currently satisfied
+ * by adjusting a input or output framerate, initiated by 'tensor_rate' element.
+ * When 'tensor_filter' receives a throttling QoS event from the 'tensor_rate' element,
+ * it compares the average processing latency and throttling delay, and takes the
+ * maximum value as the threshold to drop incoming frames by checking a buffer timestamp.
+ * In this way, 'tensor filter' can avoid unncessary calculation and adjust a framerate,
+ * effectively reducing resource utilizations.
+ * Even in the case of receiving QoS events from multiple downstream pipelines (e.g., tee),
+ * 'tensor_filter' takes the minimum value as the throttling delay for downstream pipeline
+ * with more tight QoS requirement. Lastly, 'tensor_filter' also sends QoS events to
+ * upstream elements (e.g., tensor_converter, tensor_src) to possibly reduce incoming
+ * framerates, which is a better solution than dropping framerates.
  */
 
 #ifdef HAVE_CONFIG_H
@@ -150,6 +163,8 @@ static gboolean gst_tensor_filter_start (GstBaseTransform * trans);
 static gboolean gst_tensor_filter_stop (GstBaseTransform * trans);
 static gboolean gst_tensor_filter_sink_event (GstBaseTransform * trans,
     GstEvent * event);
+static gboolean gst_tensor_filter_src_event (GstBaseTransform * trans,
+    GstEvent * event);
 
 /**
  * @brief initialize the tensor_filter's class
@@ -201,8 +216,9 @@ gst_tensor_filter_class_init (GstTensorFilterClass * klass)
   trans_class->transform_size =
       GST_DEBUG_FUNCPTR (gst_tensor_filter_transform_size);
 
-  /* setup sink event */
+  /* setup events */
   trans_class->sink_event = GST_DEBUG_FUNCPTR (gst_tensor_filter_sink_event);
+  trans_class->src_event = GST_DEBUG_FUNCPTR (gst_tensor_filter_src_event);
 
   /* start/stop to call open/close */
   trans_class->start = GST_DEBUG_FUNCPTR (gst_tensor_filter_start);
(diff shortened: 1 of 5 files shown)


gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -456,6 +461,29 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
       GST_STR_NULL (prop->model_files[0]));
   allocate_in_invoke = gst_tensor_filter_allocate_in_invoke (priv);
 
+  /* skip input data when throttling delay is set */
+  if (priv->throttling_delay != 0) {
```
kparichay, line 465: As per our offline discussion, this would be best applied to tensor_converter and tensor_src*

dongju-chae: Yes, sure. I will make next PRs for that.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -999,6 +1050,48 @@ gst_tensor_filter_sink_event (GstBaseTransform * trans, GstEvent * event)
...
+      if (type == GST_QOS_TYPE_THROTTLE && diff > 0) {
+        priv->throttling_delay = diff;
+        gst_event_unref (event);
+
+        /* enable the average latency profiling */
+        g_object_set (self, "latency", 1, NULL);
+        return TRUE;
+      }
+
+      return gst_pad_push_event (trans->sinkpad, event);
```
jaeyun-jung, line 1085: default evt handle in bast-transform?

dongju-chae: Updated.

gst/nnstreamer/tensor_filter/tensor_filter.c
myungjoo: What if...
```
t.c (doing 100FPS) --> T -+-> T-F model A (may do 10 FPS)
                          +-> T-F model B (may do 200 FPS)
```
Do you still want to handle overflow QoS events at t.c?

dongju-chae: Yes, I have to. But, this PR is targeting to compare its latency and throttling delay from `tensor_rate`, as a prototype.
I will make another PR to consider multiple QoS events from downstream elements.

myungjoo (commented):
Before merging, please describe the QoS policy of tensor_filter as a GST Element Manual entry (at the top section)

dongju-chae:
@myungjoo Sure. Its manual will be updated soon.

myungjoo (approved):
LGTM.

To reviewers: this is a huge update on the core. Please review it carefully.

helloahn (approved):
LGTM

myungjoo:
PTAL: @jaeyun-jung @kparichay

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -457,6 +487,44 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
...
+    if (GST_CLOCK_TIME_IS_VALID (prev_ts)) {
+      GstClockTimeDiff diff = curr_ts - prev_ts;
+      GstClockTimeDiff delay;
+
+      priv->throttling_accum += diff;
+
+      /* check whether the average latency is longer than throttling delay */
+      delay = MAX (priv->prop.latency * 1000, priv->throttling_delay);
+
+      if (priv->throttling_accum < delay) {
```
kparichay, line 508: should this be `throttling_accum > delay`? throttling_accum indicates the time between consecutive frames. If this frame exceeds the allowed, then QoS is send. 
With the current code, if `curr_ts == prev_ts`, `throttling_accum` goes to 0, and a QoS event is sent upstream with positive diff as `delay` (from documentation - A positive diff value indicates how late the buffer with timestamp was.). However, in this case, the buffer is not at all late.

gst/nnstreamer/tensor_filter/tensor_filter.c
kparichay: @dongju-chae I wonder if the images in the PR description have been updated or not? Because the fps of tensor_converter is 36FPS. However, with this change, we expect it to also be 5fps.
Please correct me if I understood wrongly.

gst/nnstreamer/tensor_filter/tensor_filter.c
dongju-chae: > should this be throttling_accum > delay? throttling_accum indicates the time between consecutive frames. If this frame exceeds the allowed, then QoS is send.

No. `throttling_accum` is being used to drop frames. So, until it reaches to `delay`, all incoming frames should be dropped.

kparichay: Understood, I misunderstood buffer timestamp.

kparichay (approved):
LGTM

gst/nnstreamer/tensor_filter/tensor_filter.c
dongju-chae: Yep, it seems like a glitch. It keeps slightly changing over time.

PR #2790, opened 2020-10-07
[Fix/Filter] Use g_value_take_string in get_property

dongju-chae:
This patch uses g_value_take_string, instead of g_value_set_string,
in get_property impl, to avoid redundant invoking g_free.

**Self evaluation:**


jaeyun-jung (approved):

myungjoo (approved):
LGTM

PR #2792, opened 2020-10-07
[Common] Add rank field to GstTensorInfo

again4you:
This patch newly adds the rank field to GstTensorInfo structure. 
When passing the dimension string in tensor filter, the rank value is automatically set. 
For example, the rank value of the dimension string '1:10:1:1' is 4. This rank field is not used in cap-negotiation phase.

This PR contains the below patches.
* Add rank field to GstTensorInfo
* bugfix: Fix the wrong rank of gst_tensor_info_get_rank()

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/include/tensor_typedef.h
+++ b/gst/nnstreamer/include/tensor_typedef.h
@@ -172,6 +172,7 @@ typedef struct
                    and some (tensorflow-lite) do not need this. */
   tensor_type type; /**< Type of each element in the tensor. User must designate this. */
   tensor_dim dimension; /**< Dimension. We support up to 4th ranks.  */
+  unsigned int rank; /** < Rank of the tensor. */
 } GstTensorInfo;
 
 /**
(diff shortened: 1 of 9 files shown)


gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -172,6 +172,7 @@ typedef struct
                    and some (tensorflow-lite) do not need this. */
   tensor_type type; /**< Type of each element in the tensor. User must designate this. */
   tensor_dim dimension; /**< Dimension. We support up to 4th ranks.  */
+  unsigned int rank; /** < Rank of the tensor. */
```
myungjoo, line 175: NACK. We won't add rank in GST-CAPs. It has a lot of adverse effects.

again4you:
> NACK. We won't add rank in GST-CAPs. It has a lot of adverse effects. Rank values (and their overriden rank values) should be effective locally.
> 
> If rank value is assigned locally to tensor-filter/tensor-decoder sub-plugins only, do not add this element to the common data structure.

OK. To comply with the [Rank counting rule](https://github.com/nnstreamer/nnstreamer/wiki/Rank-counting-with-other-tensor-types), I will submit a new PR instead of this one.

PR #2798, opened 2020-10-14
[Meson] Fix the buildbreak of custom_example_tensorrt example

again4you:
Because of the wrong file path in meson script, below error occurs when
running `meson build` command. This patch fixes that bug.

* Error message: nnstreamer_example/meson.build:98:2: ERROR: File
  nnstreamer_customfilter_example_tensorrt_reshape.cc does not exist.

**Self evaluation:**


dongju-chae (approved):
LGTM, Hmm what happens in here?

jaeyun-jung (approved):

PR #2801, opened 2020-10-16
[Filter/TFLite] option to set no of threads

jaeyun-jung:
add option to set the number of threads to tf-lite interpreter.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -29,6 +29,7 @@
 #include <unistd.h>
 #include <limits.h>
 #include <algorithm>
+#include <thread>
 
 #include <nnstreamer_log.h>
 #include <nnstreamer_plugin_api.h>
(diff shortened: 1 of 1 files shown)


again4you (approved):
LGTM!

gichan-jang (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -290,6 +290,8 @@ TFLiteInterpreter::loadModel (bool use_nnapi)
   }
 
   interpreter->UseNNAPI (use_nnapi);
+  if (num_threads > 0)
```
helloahn, line 293: Is there no max limit?

jaeyun-jung: It seems no limit in tf-lite function. I added code to check hardware_concurrency() before setting the num_threads.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -29,6 +29,7 @@
 #include <unistd.h>
 #include <limits.h>
 #include <algorithm>
+#include <thread>
```
myungjoo, line 32: Do you really need this?

myungjoo (approved):
Please check if you really need to include ```thread'''.
Other than that, LGTM.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
jaeyun-jung: it is necessary to use std::thread::hardware_concurrency()

PR #2808, opened 2020-10-20
[Tensor Transform] Fix unsupported orc type cast

gichan-jang:
An orc does not support a 64bit integer typecast.
A problem occurs when the pipeline is launched with the orc option,
  and then the option is changed using set_property to convert type int64 or uint64.
So disables the orc if the int64 or uint64 typecast is required.

Self evaluation:

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -351,7 +351,7 @@ gst_tensor_transform_get_stand_mode (const gchar * str)
       case _NNS_UINT8: orc_func_conv (intype, u8) ((gpointer) o, (gpointer) i, n); break; \
       case _NNS_FLOAT64: orc_func_conv (intype, f64) ((gpointer) o, (gpointer) i, n); break; \
       case _NNS_FLOAT32: orc_func_conv (intype, f32) ((gpointer) o, (gpointer) i, n); break; \
-      default: GST_ERROR_OBJECT (filter, "Unsupported type %d", otype); g_assert (0); break; \
+      default: GST_ERROR_OBJECT (filter, "Unsupported output type %d", otype); g_assert (0); break; \
     } \
   } while (0)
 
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -765,7 +765,11 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
     case GTT_TYPECAST:
     {
       if (g_regex_match_simple (REGEX_TYPECAST_OPTION, filter->option, 0, 0)) {
-        filter->data_typecast.to = gst_tensor_get_type (filter->option);
+        gint type_to = gst_tensor_get_type (filter->option);
```
jaeyun-jung, line 768: plz change to tensor_type
```
tensor_type type_to ...
```

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -765,7 +765,11 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
...
+        if (type_to == _NNS_INT64 || type_to == _NNS_UINT64) {
```
jaeyun-jung, line 769: requires definition HAVE_ORC for filter->orc_supported

#ifdef HAVE_ORC
... condition to set filter->orc_supported
#endif

jaeyun-jung (approved):

gst/nnstreamer/tensor_transform/tensor_transform.c
gichan-jang: Updated.

gst/nnstreamer/tensor_transform/tensor_transform.c
gichan-jang: Thanks! Updated.

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -765,7 +765,13 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
     case GTT_TYPECAST:
     {
       if (g_regex_match_simple (REGEX_TYPECAST_OPTION, filter->option, 0, 0)) {
-        filter->data_typecast.to = gst_tensor_get_type (filter->option);
+        tensor_type type_to = gst_tensor_get_type (filter->option);
+#ifdef HAVE_ORC
+        if (type_to == _NNS_INT64 || type_to == _NNS_UINT64) {
+          filter->orc_supported = FALSE;
+        }
```
myungjoo, line 772: need ```filter->orc_supported = TRUE;``` somewhere in case type_to has been changed from INT64 to INT8 (or others).

myungjoo (changes requested):
Please consider the case the properties are changed in the middle.

gst/nnstreamer/tensor_transform/tensor_transform.c
gichan-jang: Right!
I changed to decide `orc_supported` during the negotiation process.
It covers these cases.

myungjoo (approved):
LGTM

PR #2817, opened 2020-10-22
[NNFW] cache info from nnfw

jaeyun-jung:
Cache nnfw tensors info and handle rank value when setting input info.


dongju-chae (approved):
LGTM

myungjoo (approved):

PR #2818, opened 2020-10-23
[C-Api] prepare new APIs

jaeyun-jung:
prepare next ACR
- function to invoke with preallocated output handle
- function to open nn model with custom options

Change function names and add description for each API.

### DIFF
--- a/api/capi/include/nnstreamer-single.h
+++ b/api/capi/include/nnstreamer-single.h
@@ -76,6 +76,40 @@ typedef void *ml_single_h;
 int ml_single_open (ml_single_h *single, const char *model, const ml_tensors_info_h input_info, const ml_tensors_info_h output_info, ml_nnfw_type_e nnfw, ml_nnfw_hw_e hw);
 
 /**
+ * @brief Opens an ML model and returns the instance as a handle.
+ * @details Even if the model has flexible input data dimensions,
+ *          input data frames of an instance of a model should share the same dimension.
+ * @since_tizen 6.5
+ * @remarks %http://tizen.org/privilege/mediastorage is needed if @a model is relevant to media storage.
+ * @remarks %http://tizen.org/privilege/externalstorage is needed if @a model is relevant to external storage.
+ * @param[out] single This is the model handle opened. Users are required to close
+ *                   the given instance with ml_single_close().
+ * @param[in] model This is the path to the neural network model file.
+ * @param[in] input_info This is required if the given model has flexible input
+ *                      dimension, where the input dimension MUST be given
+ *                      before executing the model.
+ *                      It is required by some custom filters of NNStreamer.
+ *                      You may set NULL if it's not required.
+ * @param[in] output_info This is required if the given model has flexible output dimension.
+ * @param[in] nnfw The neural network framework used to open the given @a model.
+ *                 Set #ML_NNFW_TYPE_ANY to let it auto-detect.
+ * @param[in] hw Tell the corresponding @a nnfw to use a specific hardware.
+ *               Set #ML_NNFW_HW_ANY if it does not matter.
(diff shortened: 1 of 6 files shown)


api/capi/include/nnstreamer-single.h
```diff
@@ -106,6 +141,23 @@ int ml_single_close (ml_single_h single);
  */
 int ml_single_invoke (ml_single_h single, const ml_tensors_data_h input, ml_tensors_data_h *output);
 
+/**
+ * @brief Invokes the model with the given input data and fills the @a output data handle.
+ * @details This function does not allocate new data handle, unlike ml_single_invoke() which returns newly allocated output handle.
+ * @since_tizen 6.5
+ * @param[in] single The model handle to be inferred.
+ * @param[in] input The input data to be inferred.
+ * @param[in] output The output data to be filled.
```
myungjoo, line 150: Need more info for developers:
```
The output data to be filled.
```
==>
```
The output data to be filled by the API. Output should be preallocated before calling the API.
```

api/capi/include/nnstreamer-single.h
```diff
@@ -106,6 +141,23 @@ int ml_single_close (ml_single_h single);
  */
 int ml_single_invoke (ml_single_h single, const ml_tensors_data_h input, ml_tensors_data_h *output);
 
+/**
+ * @brief Invokes the model with the given input data and fills the @a output data handle.
```
myungjoo, line 145: Add more description
```
Invokes the model with the given input data and fills the @a output data handle.
```
==>
```
Invokes the model with the given input data and fills the @a output data handle. This is generally faster than ml_invoke_single.
```

api/capi/include/nnstreamer-single.h
```diff
@@ -106,6 +141,23 @@ int ml_single_close (ml_single_h single);
...
+ * @details This function does not allocate new data handle, unlike ml_single_invoke() which returns newly allocated output handle.
```
myungjoo, line 146:
```
This function does not allocate new data handle, unlike ml_single_invoke() which returns newly allocated output handle.
```
==>
```
The caller should preallocate memory buffers of the given output handle before calling the API.
Note that ml_single_invoke() allocates memory buffers of the output handle in the API, which may incur memcpy.
```

myungjoo (commented):
LGTM. Please rephrase a few doxygen entries.

api/capi/include/nnstreamer-single.h
again4you, line 150: Maybe [in/out]?
`@param[in] output` => `@param[in/out] output`

again4you (approved):
Except comments, LGTM!

api/capi/include/nnstreamer-single.h
kparichay: Addon : `The memory of each output must be contiguous. However, it is not required for multiple outputs to have contiguous shared memory.`?

kparichay:
This is most likely outside the scope of this PR but I want your opinion.
This patch will allow supporting pre-allocated handle to pass to `invoke`.
However, as nnstreamer `tensor_data` c-api does not allow to create a wrapper of `tensor_data` around already allocated memory, I think the usage of this API might be limited. 
How about introducing an additional API like `int ml_tensors_data_create (const ml_tensors_info_h info, ml_tensors_data_h *handle, void *data)` where `void *data` is the pre-allocated memory allocated by the user. This function would just wrap the user allocated memory with `tensor_data` so that it can be used inside nnstreamer. (I know this is currently limited to just 1 tensor but can be extended)

jaeyun-jung:
@kparichay I added new issue about your suggestion. (#2827)

dongju-chae (approved):
LGTM

myungjoo (approved):
LGTM

PR #2838, opened 2020-10-27
[gRPC] Implement tensor_source (client) and tensor_sink (server)      

dongju-chae:
This patch implement a reverse pair of gRPC tensor_source/tensor_sink.
(i.e., tensor_source as client and tensor_sink as server)

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_grpc.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_grpc.c
@@ -379,7 +379,7 @@ gst_tensor_src_grpc_start (GstBaseSrc * src)
 
   grpc_set_callback (self, _grpc_callback);
 
-  ret = grpc_start (self);
+  ret = grpc_start (self, GRPC_DIRECTION_FROM_PROTOBUF);
   if (ret)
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_source/tensor_src_grpc.c
```diff
@@ -493,14 +509,23 @@ gst_tensor_src_grpc_set_property (GObject * object, guint prop_id,
...
+      gchar * host;
+
       if (!g_value_get_string (value)) {
         ml_logw ("host property cannot be NULL");
         break;
       }
-      g_free (self->host);
-      self->host = g_value_dup_string (value);
-      silent_debug ("Set host = %s", self->host);
+      host = g_value_dup_string (value);
```
myungjoo, line 519: Why value cannot be null?

myungjoo (approved):
LGTM

ext/nnstreamer/tensor_source/tensor_src_grpc.c
dongju-chae: Right, I will remove warning message and regard setting to NULL as the unchanged action.

jaeyun-jung (approved):

PR #2848, opened 2020-11-02
[Android] codeclean

jaeyun-jung:
1. move native methods registration to each file.
2. add new function to get data ptr in native.

### DIFF
--- a/api/android/api/src/main/jni/nnstreamer-native.h
+++ b/api/android/api/src/main/jni/nnstreamer-native.h
@@ -180,22 +180,28 @@ extern void
 nns_set_priv_data (pipeline_info_s * pipe_info, gpointer data, nns_priv_destroy destroy_func);
 
 /**
+ * @brief Get element data of given name.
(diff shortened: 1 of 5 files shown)


api/android/api/src/main/jni/nnstreamer-native.h
```diff
@@ -179,23 +179,29 @@ nns_destroy_pipe_info (pipeline_info_s * pipe_info, JNIEnv * env);
 extern void
 nns_set_priv_data (pipeline_info_s * pipe_info, gpointer data, nns_priv_destroy destroy_func);
 
+/**
+ * @brief Get element data of given name.
+ */
+extern element_data_s *
```
wooksong, line 185: Q: Is this ```extern``` keyword required? In general, the ```extern``` keyword is implicitly assumed when a function is declared in the header file.

wooksong (approved):

myungjoo (approved):

PR #2849, opened 2020-11-02
[Android] handle invalid context

jaeyun-jung:
Handle invalid param case -init nnstreamer with invalid context.

### DIFF
--- a/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
+++ b/api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
@@ -338,6 +338,16 @@ public class APITestCommon {
     }
 
     @Test
+    public void testInitWithInvalidCtx_n() {
+        try {
+            NNStreamer.initialize(null);
+            fail();
+        } catch (Exception e) {
+            /* expected */
+        }
+    }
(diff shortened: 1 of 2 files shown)


myungjoo (approved):
LGTM

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
```diff
@@ -337,6 +337,16 @@ public void useAppContext() {
         assertEquals("org.nnsuite.nnstreamer.test", context.getPackageName());
     }
 
+    @Test
+    public void testInitWithInvalidCtx_n() {
+        try {
+            NNStreamer.initialize(null);
+            fail();
```
wooksong, line 344: Q: Is it a reachable code line?

dongju-chae (approved):
LGTM

api/android/api/src/androidTest/java/org/nnsuite/nnstreamer/APITestCommon.java
jaeyun-jung: It is test code to check exception case. If init with null context will not throw exception, this test will be failed.

PR #2863, opened 2020-11-11
[JOIN] Add new element join

gichan-jang:
Add new gstreamer element: join
The join select the out that arrived first among the input streams.

Example launch line. 
If the brightness of the camera is bright, the camera image is output, and if it is dark, the video test source is output.
```
gst-launch-1.0 v4l2src ! videoconvert ! videoscale ! video/x-raw, format=RGB, width=640, height=480, framerate=30/1 ! tensor_converter ! mux.sink_0 \
               videotestsrc ! videoconvert ! videoscale ! video/x-raw, format=RGB, width=640, height=480, framerate=30/1 ! tensor_converter ! mux.sink_1 \
                              tensor_mux name=mux ! tensor_if name=tif compared_value=TENSOR_AVERAGE_VALUE compared-value-option=0 \
                              supplied-value=100 operator=GT then=TENSORPICK then-option=0 else=TENSORPICK else-option=1 \
               tif.src_0 ! queue ! join.sink_0 \
               tif.src_1 ! queue ! join.sink_1 \
               join name=join ! tensor_decoder mode=direct_video ! videoconvert ! ximagesink
```
 - todo
  1. Add unit test cases
  2. Upload sample application 

Self evaluation:


packaging/nnstreamer.spec
```diff
@@ -621,7 +627,7 @@ ninja -C build %{?_smp_mflags}
 
 export NNSTREAMER_SOURCE_ROOT_PATH=$(pwd)
 export NNSTREAMER_BUILD_ROOT_PATH=$(pwd)/build
-export GST_PLUGIN_PATH=${NNSTREAMER_BUILD_ROOT_PATH}/gst/nnstreamer:${NNSTREAMER_BUILD_ROOT_PATH}/ext/nnstreamer
+export GST_PLUGIN_PATH=${NNSTREAMER_BUILD_ROOT_PATH}/gst:${NNSTREAMER_BUILD_ROOT_PATH}/ext
```
dongju-chae, line 630: You have to check `debian/rules` as well.

gst/join/gstjoin.c
```diff
@@ -0,0 +1,1075 @@
...
+#ifndef PACKAGE
+#define PACKAGE "join"
+#endif
+
+GST_PLUGIN_DEFINE (GST_VERSION_MAJOR,
+    GST_VERSION_MINOR,
+    join,
+    "Select the out that arrived first among the input streams",
+    plugin_init, VERSION, "LGPL", "nnstreamer",
+    "GStreamer core elements: input-selector")
```
dongju-chae, line 1075: Well... I recommend you to use `./ext/nnstreamer/registerer` for extra plugins.

dongju-chae: Or, if it's totally independent plugin, I think you don't need to use `nnstreamer` as a package name

gst/join/gstjoin.c
```diff
@@ -0,0 +1,1075 @@
...
+
+  if (!gst_element_register (plugin, "join", GST_RANK_NONE, GST_TYPE_JOIN)) {
+    return FALSE;
+  }
+
+  return TRUE;
+}
+
+#ifndef PACKAGE
+#define PACKAGE "join"
```
dongju-chae, line 1067: BTW, what's the hierarchy of plugin and elements? I think `join` is close to an element, not a plugin. (unless this plugin will contain only a single element)
So, we have to think its plugin name, for example
```
plugin: nnstreamer-extra
- elements: join, ...
```

gichan-jang: Thanks for the suggestion. As we discussed offline, `nnstreamer-misc` seems more suitable as a package name.

gst/join/gstjoin.c
gichan-jang: I think there will be no more elements related to `join` in the future, so registerer seems unnecessary.
And it is independent of nnstreamer, it would be better to create a package called `join`. 
Changed the package name.

dongju-chae: Yeap, If then, looks good.

dongju-chae (approved):
LGTM

packaging/nnstreamer.spec
```diff
@@ -461,6 +462,11 @@ Summary:	NNStreamer developer utilities
 %description util
 NNStreamer developer utilities include nnstreamer configuration checker.
 
+%package misc
+Summary:	NNStreamer extra packages
+%description misc
+Provides additional packages that are not essential for GStreamer/NNStreamer.
```
myungjoo, line 468: JOIN is essential. :)

```
Provides additional gstreamer plugins for nnstreamer pipelines
```

gst/join/gstjoin.c
```diff
@@ -0,0 +1,1076 @@
...
+ * N streams should not operate at the same time.
+ * For example, If one sinkpad is receiving buffer, the others should be stopped.
+ * <refsect2>
+ * <title>Example launch line</title>
+ * gst-launch-1.0 ... (input stream 0) ! join.sink_0 \
+ *                ... (input stream 1) ! join.sink_1 \
+ *                ... \
+                  ... (input stream N) ! join.sink_n \
+                  join name=join ! (arrived input stream) ...
+ * </refsect2>
```
myungjoo, line 28: Please add that all the capabilities (input stream i and output stream) should be the same.

gst/join/gstjoin.c
```diff
@@ -0,0 +1,1076 @@
...
+
+    goto done;
+  }
+flushing:
+  {
+    GST_DEBUG_OBJECT (pad, "We are flushing, discard buffer %p", buf);
+    gst_buffer_unref (buf);
+    res = GST_FLOW_FLUSHING;
+    goto done;
+  }
```
myungjoo, line 721: This goto and label statements are convoluted. Use it only for error handling and it should be linear.
If you want to use such blcok structures, use some static/inline functions

gichan-jang: In this case, should we stop the pipeline or just warn them?

gst/join/gstjoin.c
gichan-jang: In this case, should we stop the pipeline or just warn them?

myungjoo: If JOIN cannot enforce all the caps equalized, it should fail at the negotiation phase.
However, during the negotiation phase, caps are not finalized (it may be flexible) and JOIN should try to equalize them.

packaging/nnstreamer.spec
gichan-jang: Updated!

gst/join/gstjoin.c
gichan-jang: Thank you. Updated.

gst/join/gstjoin.c
gichan-jang: Check capabilities of the sink pads are added.
The caps are checked when they are requested.

debian/nnstreamer.install
```diff
@@ -4,6 +4,6 @@
 /usr/lib/nnstreamer/decoders/libnnstreamer_decoder_image_labeling.so
 /usr/lib/nnstreamer/decoders/libnnstreamer_decoder_direct_video.so
 /usr/lib/nnstreamer/filters/libnnstreamer_filter_cpp.so
-/usr/lib/*/gstreamer-1.0/*.so
+/usr/lib/*/gstreamer-1.0/libnnstreamer.so
```
dongju-chae, line 7: As we discussed in offline, I think it needs a quick patch because the install of grpc plugin is being blocked.
Let's review & merge this PR as soon as possible, or separate the PR.

gichan-jang: I'm not sure how long it will take to merge, so I will separate the PR related to this.

myungjoo (changes requested):
Update ```restricted_element``` of nnstreamer.spec so that pipeline API users of Tizen apps may use JOIN.

gichan-jang:
> Update `restricted_element` of nnstreamer.spec so that pipeline API users of Tizen apps may use JOIN.

`join` is added in `restriced_element`! Thanks.

jni/nnstreamer.mk
```diff
@@ -141,6 +141,10 @@ NNSTREAMER_DECODER_IS_SRCS := \
     $(NNSTREAMER_EXT_HOME)/tensor_decoder/tensordec-imagesegment.c \
     $(NNSTREAMER_EXT_HOME)/tensor_decoder/tensordecutil.c
 
+# decoder image segment
```
jaeyun-jung, line 144: plz fix this comment line

gichan-jang: Ooops, Thanks.

gichan-jang:
Join is added in NNStreamer lib for Android.

myungjoo (approved):

PR #2867, opened 2020-11-12
[Tools/Parser] Update gstparser's types @open sesame 12/3 12:43

myungjoo:
We do not need to load up actual GObjects for this parser.
Thus, the data structures should be simplied.
Define substitute data types for GST Elements as well.


dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM

PR #2868, opened 2020-11-12
[Tools/Parser] Replace GST macros to dummy

myungjoo:
We do not need to create GST objects with this parser.
Do not use GST object invoking macros.

### DIFF
--- a/tools/development/parser/types.c
+++ b/tools/development/parser/types.c
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * GStreamer Pipeline from/to PBTxt Converter Parser
+ * Copyright (C) 2020 MyungJoo Ham <myungjoo.ham@samsung.com>
+ *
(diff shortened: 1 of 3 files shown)


dongju-chae (approved):
LGTM

tools/development/parser/types.c
```diff
@@ -0,0 +1,42 @@
...
+  return quark;
+}
+
+/**
+ * @brief Replacement of gst_parse_element_make
+ */
+_Element *
+nnstparser_element_make (const gchar * element, const gchar * name)
+{
+  _Element ret = g_malloc (sizeof (_Element));
```
jaeyun-jung, line 37: how about using g_malloc0 or g_new0 for null init

jaeyun-jung (approved):

tools/development/parser/types.c
myungjoo: uh oh.. yes, that's better. I'll add later in another PR.

PR #2870, opened 2020-11-12
[Tools/Parser] Parser Rule #1 and #2 updated

myungjoo:
Replacing gst-factory method with simplified implementation.
Updated grammar and its static function for parsing rule
1. element
2. elementary
    
### DIFF
--- a/tools/development/parser/grammar.y
+++ b/tools/development/parser/grammar.y
@@ -293,13 +293,10 @@ error:
   goto out;
 }
 
-static void gst_parse_element_set (gchar *value, GstElement *element, graph_t *graph)
+static void nnstparser_element_set (gchar *value, _Element *element, graph_t *graph)
 {
-  GParamSpec *pspec = NULL;
   gchar *pos = value;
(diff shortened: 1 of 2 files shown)


tools/development/parser/grammar.y
```diff
@@ -324,62 +321,15 @@ static void gst_parse_element_set (gchar *value, GstElement *element, graph_t *g
...
-         g_value_set_object (&v, bin);
-         got_value = TRUE;
-       }
-    }
-    if (!got_value)
-      goto error;
-    g_object_set_property (target, pspec->name, &v);
-  }
+  /* Assign a "name=value" pair to element */
+  prop = g_malloc (sizeof(_Property));
```
dongju-chae, line 325: How about using `g_malloc0` here, too.

myungjoo: Fixed. I'm applying g_new0

jaeyun-jung (approved):

dongju-chae (approved):
LGTM

PR #2877, opened 2020-11-18
[NNFW] print retry count log

jaeyun-jung:
For debug, add log retry count in set-dimension function.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -539,7 +539,7 @@ nnfw_invoke_dummy (const GstTensorFilterProperties * prop, void **private_data,
 {
   GstTensorMemory input[NNS_TENSOR_SIZE_LIMIT] = { 0, };
   GstTensorMemory output[NNS_TENSOR_SIZE_LIMIT] = { 0, };
-  guint i;
+  guint i, retry;
 
   for (i = 0; i < in_info->num_tensors; ++i) {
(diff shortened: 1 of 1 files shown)


myungjoo (approved):
LGTM. Anyway, don't you need a limit for retrying?

jaeyun-jung:
@myungjoo 
When changing input shape in NNFW, we cannot calculate the exact buffer size of output tensor.
Current NNFW can return the output shape after invoke is successfully called.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -552,8 +552,10 @@ nnfw_invoke_dummy (const GstTensorFilterProperties * prop, void **private_data,
     output[i].data = g_malloc (output[i].size);
   }
 
-  while (nnfw_invoke (prop, private_data,  input, output) != 0) {
-    nns_logw ("Invoke failed, reallocate output tensors and retry.");
+  retry = 0;
+  while (nnfw_invoke (prop, private_data, input, output) != 0) {
+    nns_logw ("Invoke failed, reallocate output tensors and retry (%u).", ++retry);
```
myungjoo, line 557: Please mark this logic as ```/** @todo workaround. to be updated when ONE provides get-size API or proper error codes. */```

wooksong (approved):

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
jaeyun-jung: updated

PR #2879, opened 2020-11-18
[Android/Api] method to check element availability

jaeyun-jung:
Add new method to check element available on the pipeline.

### DIFF
--- a/api/android/api/src/main/java/org/nnsuite/nnstreamer/Pipeline.java
+++ b/api/android/api/src/main/java/org/nnsuite/nnstreamer/Pipeline.java
@@ -38,6 +38,7 @@ public final class Pipeline implements AutoCloseable {
     private HashMap<String, ArrayList<NewDataCallback>> mSinkCallbacks = new HashMap<>();
     private StateChangeCallback mStateCallback = null;
 
+    private static native boolean nativeCheckElementAvailability(String element);
     private native long nativeConstruct(String description, boolean addStateCb);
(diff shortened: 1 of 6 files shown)


myungjoo (approved):
LGTM

api/android/api/src/main/java/org/nnsuite/nnstreamer/Pipeline.java
```diff
@@ -157,6 +158,23 @@ public Pipeline(@NonNull String description, @Nullable StateChangeCallback callb
...
+     * Checks the element is registered and available on the pipeline.
+     *
+     * @param element The name of GStreamer element
+     *
+     * @return true if the element is available
+     *
+     * @throws IllegalArgumentException if given param is invalid
+     */
+    public static boolean isElementAvailable(@NonNull String element) {
+        if (element == null || element.isEmpty()) {
```
dongju-chae, line 171: I think this null check is not necessary as `element` is annotated as `@NonNull`

jaeyun-jung: NonNull annotation is just warning message in android-studio. This does not guarantee the passed param is not null.

dongju-chae: Yes.. it's the caller's responsibility.

dongju-chae (approved):
LGTM

PR #2883, opened 2020-11-19
[Android] fix invalid window error

jaeyun-jung:
prevent invalid window handle case.
video-sink does not set new window handle directly. If video-sink accesses released window handle, this will crash an application.


myungjoo (approved):

wooksong (approved):
LGTM!

Issue #2802, opened 2020-10-18
[Edge-AI/New Feature] Remote Query Filter / Remote Server Filter


myungjoo:
Requirements
1. For a single "server" pipeline, there may be multiple clients (and client pipelines)
2. For a request from a client pipeline, its reply should be sent to that exact client only.
3. The sequence of input stream should be guaranteed.

Example

Server Pipeline
```
tensor_server_accept service=A ! ... ... ... ! tensor_server_reply service=A
```

Client Pipelines (multiple instances)
```
... ... ! tensor_client_query address=X.X.X.X:X ! ... ...
```

dongju-chae:
It looks interesting to me.

Q: is there any major difference with the combination of tcp server/client elements? Just wrapping?

myungjoo:
1. This reduces the complexity of pipelines of clients
2. This adds the capability to support multiple clients with a single pipeline of a server (and single entry point of input as well)
3. You need to be ready for network outages. (no hangs!)

dongju-chae:
Thanks for the answer. I want to take this issue for the next item after finishing other tasks.

dongju-chae:
@myungjoo I have one question on it. Do you have any idea which protocol would be appropriate to communicate flatbuffer data via network (e.g., TCP or gRPC) ?

It seems that gRPC also supports flatbuffer data with zero-copy support.
(https://grpc.io/blog/grpc-flatbuffers/#generated-server-implementation)
But, in this case, nnstreamer-edge may have the dependency on gRPC.

myungjoo:
> But, in this case, nnstreamer-edge may have the dependency on gRPC.

1. gRPC for lightweight devives: https://github.com/containerd/ttrpc  https://groups.google.com/g/grpc-io/c/A0PlDMi2an8  ...
2. TCP/UDP

dongju-chae:
> Maybe we need both mechanisms ready. (2 is trivial)

Right. Then, it might be selected using a property.

myungjoo:
If you are going to use "flatbuf-decoder/converter" and "gRPC-sink/src" in "query", you will need to look at GST-BIN class carefully.

myungjoo:
We have additional requirements & usage cases along with a query filter.
You need to discuss with @wooksong for the requirements & usage cases.
Then. please start with the class diagram before starting implementation.

dongju-chae:
Got it. I will follow up the latest requirements and usage cases before implementing gRPC/query.

[...]

PR #2892, opened 2020-11-24
[Edge/Package] Prepare to implement the nnstreamer-edge plugin

dongju-chae:
This patch adds the registerer for nnstreamer-edge plugin.
It currently provides only skeleton codes for initial package.

Related issue: #2802 

The impl. for tensor_server_accept/reply and tensor_client_query
will be updated soon.

**Self evaluation:**


dongju-chae:
As discussed in offline, it's better to implement `tensor_query` as a sub-plugin of tensor filter. Let's close this.

PR #2893, opened 2020-11-25
[Filter/NNFW] invalid tensor info

jaeyun-jung:
1. Fix invalid tensor info ptr when setting new tensor shape.
(Running dummy invoke tries to set old tensor info.)
2. Allocate dummy buffer and init to 0.
3. Handle insufficient output buffer case.


myungjoo (approved):
LGTM.

anyj0527 (approved):
LGTM Thanks!

jaeyun-jung:
I will check next version with this PR.

PR #2906, opened 2020-12-01
[meson] Add deps declaration for tensorrt, grpc

junmin-kim:
command: meson build -Dtensorrt-support=disabled
log: ERROR: Unknown variable "tensorrt_deps".

Without empty dependency declaration, features can not recognize
"tensorrt_deps"
Ditto grpc.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -194,6 +194,10 @@ if not get_option('snpe-support').disabled()
 endif
 
 # tensorrt
+nvinfer_dep = dependency('', required: false)
+nvparsers_dep = dependency('', required: false)
+cuda_dep = dependency('', required: false)
+cudart_dep = dependency('', required: false)
 if not get_option('tensorrt-support').disabled()
   # check available cuda versions (11.0 and 10.2 are recommended)
   cuda_vers = [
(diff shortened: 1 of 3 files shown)


dongju-chae (approved):
LGTM

meson.build
```diff
@@ -298,7 +302,7 @@ features = {
     'extra_deps': [ pb_comp, protobuf_dep ]
   },
   'tensorrt-support': {
-    'extra_deps': tensorrt_deps
+    'extra_deps': [ tensorrt_deps ]
```
dongju-chae, line 305: Well.. I don't think it works. It's already array of deps.
How about defining empty `nvinfer_dep, ...` instead of `tensorrt_deps`.

junmin-kim: I'll declared tensorrt_deps as a empty list.

dongju-chae: Yeap, looks better

meson.build
```diff
@@ -194,6 +194,10 @@ if not get_option('snpe-support').disabled()
 endif
 
 # tensorrt
+nvinfer_dep = dependency('', required: false)
+nvparsers_dep = dependency('', required: false)
+cuda_dep = dependency('', required: false)
+cudart_dep = dependency('', required: false)
```
dongju-chae, line 200: If you've decided to define empty dependency instead of `tensorrt_deps`, please change other build scripts.

For example, in `nnstreamer_example/meson.build` and `ext/nnstreamer/tensor_filter/meson.build`, `tensorrt_deps` --> `tensorrt_support_deps`

junmin-kim:
I didn't change permission, and the meson.build is 644... Why this happened?

dongju-chae:
> I didn't change permission, and the meson.build is 644... Why this happened?

I think it's a bug on `pr-prebuild-executable`. I will fix this.
The result of `pr-prebuild-timestamp` affects the following test which is `pre-prebuild-executable`.

In short, it's due to the error from pr-prebuild-timestamp. Please check the time on your PC.

wooksong (approved):
Looks good to me

myungjoo (approved):

PR #2910, opened 2020-12-02
[Test] path to build dir

jaeyun-jung:
set build root (current dir) if env var is not set.


dongju-chae (approved):

gichan-jang (approved):
LGTM

PR #2913, opened 2020-12-02
[TensorRate/Tests] Add some unittests for tensor_rate @open sesame 12/03 09:50

dongju-chae:
This patch adds some unittests for tensor_rate.
It increases the coverage of `gsttensorrate.c` from 67.7% to 78.3%.

### DIFF
--- a/tests/nnstreamer_rate/unittest_rate.cc
+++ b/tests/nnstreamer_rate/unittest_rate.cc
@@ -1,0 +1,514 @@
+/**
+ * @file    unittest_rate.cc
+ * @date    02 Dec 2020
+ * @brief   Unit test for tensor_rate element
+ * @see     https://github.com/nnstreamer/nnstreamer
+ * @author  Dongju Chae <dongju.chae@samsung.com>
+ * @bug     No known bugs
+ */
+
+#include <gtest/gtest.h>
+#include <glib.h>
+
+#include <nnstreamer_plugin_api_filter.h>
+#include <nnstreamer_plugin_api.h>
+
+#include <unittest_util.h>
+
+#define NNS_TENSOR_RATE_NAME "tensor_rate"
+
+/**
+ * @brief Test tensor_rate existence.
+ */
+TEST (nnstreamer_rate, check_existence)
+{
+  GstElementFactory *factory;
+
+  factory = gst_element_factory_find (NNS_TENSOR_RATE_NAME);
+  EXPECT_TRUE (factory != NULL);
+  gst_object_unref (factory);
+}
+
+/**
+ * @brief Test tensor_rate existence (negative).
+ */
+TEST (nnstreamer_rate, check_existence_n)
+{
+  GstElementFactory *factory;
+  gchar *name;
+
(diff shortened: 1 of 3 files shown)


tests/nnstreamer_rate/unittest_rate.cc
```diff
@@ -0,0 +1,476 @@
...
+
+  _set_default_option (option);
+  option.mode = TENSOR_RATE_MODE_PASSTHROUGH;
+
+  ASSERT_TRUE (_setup_pipeline (option));
+
+  rate = gst_bin_get_by_name (GST_BIN (test_data.pipeline), "rate");
+  ASSERT_TRUE (rate != NULL);
+
+  ret = gst_element_set_state (test_data.pipeline, GST_STATE_PLAYING);
```
gichan-jang, line 325: How about using `setPipelineStateSync ()`?
It will be safer to use this even in slow environments.

dongju-chae: Oh, yes. I will update the codes to use `unittest_utils`.

gichan-jang (approved):
LGTM

tests/nnstreamer_rate/unittest_rate.cc
```diff
@@ -0,0 +1,475 @@
...
+  g_object_get (rate, "in", &in, NULL);
+  g_object_get (rate, "out", &out, NULL);
+  g_object_get (rate, "duplicate", &dup, NULL);
+  g_object_get (rate, "drop", &drop, NULL);
+
+  EXPECT_EQ (in, option.source_num_buffers);
+  EXPECT_EQ (dup, 0);
+
+  /** we don't expect the exact values */
+  EXPECT_GE (out, (guint64) ((option.source_num_buffers / 2) * 0.95));
```
jaeyun-jung, line 380: could you please run these tests on gbs armv7l

dongju-chae: Right.. I think it could have some timing issues.

dongju-chae: I increased the sleep time to 1 sec, and checked all tests works well.

gichan-jang: Giving a long sleep time is not a good solution.
How about using `wait_pipeline_process_buffers`.?

dongju-chae: It seems that `wait_pipeline_process_buffers` is not a generic API. If you're saying the internal function in `unittest_sink`,
I am not sure the same mechanism works as it may control number of output buffers,

gichan-jang: Right, it is impossible to use `wait_pipeline_process_buffers` in this test.
However, I think it seems better to ensure the integrity of the test through check and repeat.
A second doesn't mean much.

dongju-chae: @gichan-jang OK, I understand your concerns. I updated codes to wait the eos event instead of the constant timeout.

jaeyun-jung (approved):

PR #2915, opened 2020-12-03
[Fix/gRPC] Fix gRPC unittests to use an unused port for each test

dongju-chae:
This patch fixes gRPC unittests to use an unused port for each test.
Without this patch, it could make undesirable hangs on the CI server.

### DIFF
--- a/tests/nnstreamer_grpc/get_available_port.py
+++ b/tests/nnstreamer_grpc/get_available_port.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: LGPL-2.1-only
+# Copyright (C) 2020 Samsung Electronics
+#
+# @file get_available_port.py
+# @brief Get available socket port in local machine
+# @author Dongju Chae <dongju.chae@samsung.com>
+
+import socket
+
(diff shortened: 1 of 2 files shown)


gichan-jang (approved):
There seems to be a potential problem of very low probability, but it looks good.

zhoonit (approved):
LGTM!

tests/nnstreamer_grpc/get_available_port.py
```diff
@@ -0,0 +1,19 @@
...
+# @author Dongju Chae <dongju.chae@samsung.com>
+
+import socket
+
+if __name__ == "__main__":
+  s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+  # Bind to any available port
+  s.bind(('', 0))
+  addr = s.getsockname()
```
leemgs, line 16: Looks better. :)

leemgs (approved):
Neat.

dongju-chae:
> There seems to be a potential problem of very low probability, but it looks good.

Right. I think we may use a file lock using `flock` in the shell script.

PR #2917, opened 2020-12-03
[unittest] Fix some unittest for yocto

anyj0527:
- Set src_iio test with channels 3 -> 2
- Skip file permission test if the user is root
- Make division compatible with python3

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -15,6 +15,7 @@
 #include <gst/gst.h>
 #include <nnstreamer_plugin_api_filter.h>
 #include <string.h>
+#include <unistd.h>
 #include <tensor_common.h>
 
 #include "../gst/nnstreamer/tensor_transform/tensor_transform.h"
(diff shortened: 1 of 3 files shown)


gichan-jang:
Q. What's wrong with these in yocto?

anyj0527:
@gichan-jang 
- If user is root, the negative testcase for inaccessible file fails
- It needs more investigation, but I found that in the yocto - qemu environment, the src_iio test keeps failing with 3 channels.
  But it passes with 2 channels.
- And for the python2 v.s. python3 division behavior

tests/nnstreamer_plugins/unittest_plugins.cc
```diff
@@ -4252,6 +4253,10 @@ TEST (test_tensor_filter, framework_auto_ext_tflite_no_permission_n)
   const gchar *fw_name = NULL;
   GstElement *gstpipe;
 
+  /** If the user is root, skip this test */
+  if (getuid () == 0)
```
dongju-chae, line 4257: `geteuid`?

anyj0527: It says both `getuid` and `geteuid` return `0` for the root. `geteuid` would be more proper?

dongju-chae: There's a scenario `euid` can be zero even if `uid` is not zero, and it also has permissions to behave like a root.

anyj0527: Thanks! You're right. I updated with `geteuid`

dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM

PR #2901, opened 2020-11-27
[gRPC] Code refactoring for gRPC's flatbuffer support @open sesame 12/02 12:32

dongju-chae:
This patch performs code refactoring for gRPC's flatbuffer support.
The existing gRPC codes can be re-used for flatbuffer as well.

It also checks the same gRPC testcases for both IDLs (protobuf and flatbuf).

It resolves https://github.com/nnstreamer/nnstreamer/issues/2625.

Signed-off-by: Dongju Chae <dongju.chae@samsung.com>

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed []Skipped
2. Run test: [*]Passed [ ]Failed []Skipped

PR #2925, opened 2020-12-07
[Wait For #2901] [Query] Add the skeleton codes for 'Query' sub-plugin        

dongju-chae:
This patch adds the skeleton codes for `Query` sub-plugin.   
The `tensor_query` will be implemented as sub-plugin of `tensor_filter`.

Related issue: https://github.com/nnstreamer/nnstreamer/issues/2802
                                                             

dongju-chae:
`tensor_query` interface was not decided.

myungjoo:
This is rather tensor-fitler::query-flatbuf. The design requires a common class "query" that supports "query-flatbuf", "query-webhook", "query-protobuf", "query-...".

dongju-chae:
> This is rather tensor-fitler::query-flatbuf. The design requires a common class "query" that supports "query-flatbuf", "query-webhook", "query-protobuf", "query-...".

That's right.

dongju-chae:
According to incoming Edge-AI interface, it becomes too stale codes, thus let's close this.

Issue #2934, opened 2020-12-10
Architecture Discrepancy: Python API as Tensor Filter??


myungjoo:
```/ext/nnstreamer/tensor_filter/tensor_filter_python_api.c``` is NOT a tensor filter.
It may be migrated to https://github.com/nnstreamer/api and initiate the "api" project.

dongju-chae:
That's right and it's not gstreamer element.

dongju-chae:
As discussed in offline, we decided to rename/relocate the file.

PR #2938, opened 2020-12-10
[Python/Cleanup] Remove tensor_filter python2 sub-plugin @open sesame 12/16 12:42

dongju-chae:
This patch removes tensor_filter python2 sub-plugin.

As Python2 is deprecated, we don't need to keep the codes to support
both versions simultaneously.

It resolves https://github.com/nnstreamer/nnstreamer/issues/2934

**Self evaluation:**


packaging/nnstreamer.spec
```diff
@@ -42,9 +42,9 @@
 
 # Enable python if it's Tizen 5.0+
 %if 0%{tizen_version_major} >= 5
-%define		python_support 1
+%define		python3_support 1
 %else
-%define		python_support 0
+%define		python3_support 0
 %endif
```
myungjoo, line 48: Because we are no longer supporting python2, you need to check if it's Tizen 6.5 to support python3. If it is under 6.5, do not enable python3.

dongju-chae: I checked that `Python3` is available and works on `Tizen-6.0` as well.

meson.build
```diff
@@ -371,74 +371,44 @@ if not (pytorch_support_is_available and caffe2_support_is_available)
...
-
-else
-  warning('Python2.7 either not found or disabled')
-endif
-
-# Check python 3.x
-python3_dep = dependency('python3', required: get_option('python3-support'))
-if python3_dep.found()
-  if python3_dep.version().version_compare('>= 3.8')
+if not get_option('python3-support').disabled()
```
myungjoo, line 376: Note that if "python3-support" is enabled and python3 is not found --> BUILD ERROR!
if "python3-support" is auto and python3 is not found --> Just disable python3. (no error)

Your code appears to handle both equally.

myungjoo (commented):
LGTM except for the enable-python3 meson logic.

meson.build
dongju-chae: Applied, thanks!

myungjoo (approved):

wooksong (approved):
This is tough! LGTM!

PR #2942, opened 2020-12-11
[Spec] Disable some components in tizen-6.0 profile @open sesame 12/15 17:00

dongju-chae:
This patch disables some components in tizen-6.0 profile.

The below does not work in tizen-6.0
- tensorflow-lite2
- grpc/flatbuffers
---> This needs some revision on grpc because grpc/protobuf works well. 

**Self evaluation:**
2. Run test: [*] Passed [ ]Failed []Skipped

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -60,6 +60,12 @@
 %define		edgetpu_support 0
 %endif
 
+# tizen 6.0 (or less) backward-compatibility check
+%if ( 0%{?tizen_version_major} == 6 && 0%{?tizen_version_minor} < 5 ) || 0%{?tizen_version_major} < 6
+%define		grpc_support 0
+%define		tensorflow2_lite_support 0
+%endif
+
 # Disable e-TPU if it's not 64bit system
 %ifnarch aarch64 x86_64
 %define		edgetpu_support 0
--- a/tests/nnstreamer_grpc/runTest.sh
+++ b/tests/nnstreamer_grpc/runTest.sh
@@ -18,7 +18,23 @@ fi
 # This is compatible with SSAT (https://github.com/myungjoo/SSAT)
 testInit $1
 
+# NNStreamer and plugins path for test
 PATH_TO_PLUGIN="../../build"
+
+if [[ -d $PATH_TO_PLUGIN ]]; then
+    ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer"
+    if [[ -d ${ini_path} ]]; then
+        check=$(ls ${ini_path} | grep nnstreamer-grpc.so)
+        if [[ ! $check ]]; then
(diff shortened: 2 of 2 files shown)


packaging/nnstreamer.spec
```diff
@@ -60,6 +60,15 @@
 %define		edgetpu_support 0
 %endif
 
+# tizen 6.0 (or less) backward-compatibility check
+%if ( 0%{tizen_version_major} == 6 && 0%{tizen_version_minor} < 5 ) || 0%{tizen_version_major} < 6
```
myungjoo, line 64: ```0%{tizen...``` --> ```0%{?tizen...```

dongju-chae: It seems both work. But It's better to use the formal one, thanks.

myungjoo: It's not just "formal". It behaves differently: https://backreference.org/2011/09/17/some-tips-on-rpm-conditional-macros/

dongju-chae: Oh.. thanks for the info. Actually, I just used the similar conditions in the existing spec file.
https://github.com/nnstreamer/nnstreamer/blob/5ce00070d8423d5f90e88967abb7078f7a7a4aa9/packaging/nnstreamer.spec#L44-L51

Anyway, in this case, `0%{tizen...` and `0%{?tizen...` behaves similarly because we didn't set the default values.
```
%{?mymacro:%{mymacro}} can also be written just as %{?mymacro}
```

myungjoo (approved):
LGTM

packaging/nnstreamer.spec
```diff
@@ -60,6 +60,15 @@
 %define		edgetpu_support 0
 %endif
 
+# tizen 6.0 (or less) backward-compatibility check
+%if ( 0%{?tizen_version_major} == 6 && 0%{?tizen_version_minor} < 5 ) || 0%{?tizen_version_major} < 6
+%define		grpc_support 0
+%define		tensorflow2_lite_support 0
+%ifarch x86_64
+%define		nnfw_support 0
```
jaeyun-jung, line 68: @again4you please check this line (nnfw in 6.0)

again4you: 1) Our Tizen customers request that they need `nnfw` on Tizen 6.0. So I am going to add them to Tizen 6.0 Platform image.
2) After https://github.com/nnstreamer/nnstreamer/pull/2949 is merged, `nnfw` and `nnstreamer-nnfw` supports all Architectures including `x86_64`. So this `if` statement is not necessary.

again4you (changes requested):

packaging/nnstreamer.spec
dongju-chae: I removed this line

jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #2949, opened 2020-12-14
[Spec] Enable nnfw filter on x86 architecture

again4you:
Since nnfw works on on x86 architecture for the Tizen emulator, This
patch enables the nnfw filter when building x86 architecture.

Signed-off-by: Sangjung Woo <sangjung.woo@samsung.com>

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

#### Tizen NNFW Build status
* https://dashboard.tizen.org/index.code?sr=submit/tizen/20201214.093059

PR #2951, opened 2020-12-15
[Spec] Enable nnfw filter on x86 architecture

again4you:
Since nnfw works on on x86 architecture for the Tizen emulator, This
patch enables the nnfw filter when building x86 architecture.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed []Skipped
2. Run test: [X]Passed [ ]Failed []Skipped

#### Origin patch
* https://github.com/nnstreamer/nnstreamer/pull/2949

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -204,15 +204,15 @@ BuildRequires:	gst-plugins-base-devel
 BuildRequires:	pkgconfig(sensor)
 BuildRequires:	capi-system-sensor-devel
 
-%ifarch %arm aarch64
 %if 0%{?nnfw_support}
 # Tizen 5.5 M2+ support nn-runtime (nnfw)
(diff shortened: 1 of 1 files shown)


packaging/nnstreamer.spec
```diff
@@ -57,6 +57,11 @@
 %define		vivante_support 0
 %endif
 
+# Disable NNFW if it's not ARM/x64 and x86
+%ifnarch %arm aarch64 x86_64 i586 i686 %ix86
```
myungjoo, line 61: This is always FALSE in Tizen. We do not add %if for this condition :)
Let's make errors if a new arch (e.g., RISC-V) is introduced.

again4you: Right. I will fix it.

myungjoo (approved):
LGTM

jaeyun-jung (approved):

Issue #2931, opened 2020-12-09
Separate IDL dependencies for gRPC tensor_src/sink


dongju-chae:
Currently, gRPC tensor_src/sink supports both IDLs (protobuf, flatbuf) in runtime (using a property), but requires both dependencies unconditionally. So, we may have to separate such dependencies because someone using protobuf may do not have interests on the other IDL.

PR #2953, opened 2020-12-15
[gRPC/Fix] Decoupling gRPC IDL dependencies @open sesame 12/17 10:05

dongju-chae:
This patch seperates gRPC IDL dependencies (protobuf/flatbuf) using dlopen().
From now on, gRPC plugin does not have mandatory dependencies for its IDLs.

It resolves https://github.com/nnstreamer/nnstreamer/issues/2931

### DIFF
--- a/ext/nnstreamer/extra/nnstreamer_grpc_common.cc
+++ b/ext/nnstreamer/extra/nnstreamer_grpc_common.cc
@@ -13,15 +13,17 @@
  */
 
 #include "nnstreamer_grpc_common.h"
-#include "nnstreamer_grpc_protobuf.h"
-#include "nnstreamer_grpc_flatbuf.h"
+
+#include <gmodule.h>
 
 #include <nnstreamer_log.h>
 #include <nnstreamer_plugin_api.h>
 
-#include <thread>
-
 #include <grpcpp/health_check_service_interface.h>
+
+#define NNS_GRPC_PROTOBUF_NAME   "libnnstreamer_grpc_protobuf.so"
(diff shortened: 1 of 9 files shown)


myungjoo (approved):
LGTM

ext/nnstreamer/extra/nnstreamer_grpc_common.cc
```diff
@@ -13,37 +13,66 @@
...
+    name = NNS_GRPC_PROTOBUF_NAME;
+  else if (idl == GRPC_IDL_FLATBUF)
+    name = NNS_GRPC_FLATBUF_NAME;
+
+  if (name == NULL) {
+    ml_loge ("Unsupported IDL detected: %d\n", idl);
+    return NULL;
   }
+
+  void * handle = dlopen (name, RTLD_LAZY);
```
jaeyun-jung, line 44: how about using gmodule?

dongju-chae: Oh, it looks better. thanks.

ext/nnstreamer/extra/nnstreamer_grpc_common.h
```diff
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <string>
 #include <thread>
+#include <dlfcn.h>
```
jaeyun-jung, line 26: is it necessary?

jaeyun-jung (approved):

ext/nnstreamer/extra/nnstreamer_grpc_common.h
dongju-chae: Thanks, it's not necessary now.

Issue #2956, opened 2020-12-16
[CAPI] Potential danger after ml_pipeline_construct()


dongju-chae:
In `unittest_tizen_capi.cc`, there are some testcases which immediately change properties (even with invalid values) before the pipeline is paused (by `ml_pipeline_construct()`)

Even if the pipeline is not started, each element's main function like `chain` can be called.

This is the reason why aarch64 daily build often fails.

```
[  443s] [ RUN      ] nnstreamer_capi_element.get_property_string_01_p
[  443s] 
[  443s] ** (-0:91449): CRITICAL **: 17:24:12.785: gst_tensor_demux_get_tensor_config: assertion 'index < tensors_config->info.num_tensors' failed
[  443s] 
[  443s] ** (-0:91449): CRITICAL **: 17:24:12.785: gst_pad_set_caps: assertion 'caps != NULL && gst_caps_is_fixed (caps)' failed
[  443s] 
[  443s] (-0:91449): GStreamer-CRITICAL **: 17:24:12.785: gst_buffer_get_memory_range: assertion '(len == 0 && idx == 0 && length == -1) || (length == -1 && idx < len) || (length > 0 && length + idx <= len)' failed
[  443s] 
[  443s] (-0:91449): GStreamer-CRITICAL **: 17:24:12.785: gst_buffer_insert_memory: assertion 'mem != NULL' failed
[  443s] [       OK ] nnstreamer_capi_element.get_property_string_01_p (31 ms)
[  443s] [ RUN      ] nnstreamer_capi_element.get_property_string_02_n
[  443s] [       OK ] nnstreamer_capi_element.get_property_string_02_n (1 ms)
[  443s] [ RUN      ] nnstreamer_capi_element.get_property_string_03_n
[  443s] 
[  443s] ** (-0:91449): CRITICAL **: 17:24:12.815: gst_tensor_demux_get_tensor_config: assertion 'index < tensors_config->info.num_tensors' failed
[  443s] 
[  443s] ** (-0:91449): CRITICAL **: 17:24:12.815: gst_pad_set_caps: assertion 'caps != NULL && gst_caps_is_fixed (caps)' failed
[  443s] qemu: uncaught target signal 11 (Segmentation fault) - core dumped
[...]

PR #2957, opened 2020-12-16
[CAPI] Wait until the pipeline state is changed

dongju-chae:
This patch fixes a potential danger after `ml_pipeline_construct()` is called.

Currently, as `ml_pipeline_construct()` does not wait the pipeline state change,
the following APIs such as `set_property` may affect the initial pipeline processing
with dummy data (even if the pipe line is not started).

For example, the below sequence may incurs some problems because the property
is changed while processing the demux's `chain()`.
```
status = ml_pipeline_construct (pipeline, nullptr, nullptr, &handle);

status = ml_pipeline_element_get_handle (handle, "demux", &demux_h);

status = ml_pipeline_element_set_property_string (demux_h, "tensorpick", "1,2"); 
```

It resolves https://github.com/nnstreamer/nnstreamer/issues/2956

Please check this PR: @jaeyun-jung @again4you 

### DIFF
--- a/api/capi/src/nnstreamer-capi-pipeline.c
+++ b/api/capi/src/nnstreamer-capi-pipeline.c
@@ -630,6 +630,17 @@ construct_pipeline_internal (const char *pipeline_description,
   /* finally set pipeline state to PAUSED */
   if (status == ML_ERROR_NONE) {
     status = ml_pipeline_stop ((ml_pipeline_h) pipe_h);
+
+    if (status == ML_ERROR_NONE) {
+      /**
+       * Let's wait until the pipeline state is changed to paused.
+       * Otherwise, the following APIs like 'set_property' may incur
+       * unintended behaviors. But, don't need to return any error
+       * even if this state change is not finished within the timeout,
+       * just replying on the caller.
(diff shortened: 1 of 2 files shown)


api/capi/src/nnstreamer-capi-pipeline.c
```diff
@@ -630,6 +630,17 @@ construct_pipeline_internal (const char *pipeline_description,
...
+
+    if (status == ML_ERROR_NONE) {
+      /**
+       * Let's wait until the pipeline state is changed to paused.
+       * Otherwise, the following APIs like 'set_property' may incur
+       * unintended behaviors. But, don't need to return any error
+       * even if this state change is not finished within the timeout,
+       * just replying on the caller.
+       */
+      gst_element_get_state (pipeline, NULL, NULL, 10 * GST_MSECOND);
```
gichan-jang, line 642: If you do not check the error and state, what is different from `g_usleep (10 * GST_MSECOND)`?

jaeyun-jung (approved):

myungjoo (approved):

PR #2963, opened 2020-12-17
[Dist/gRPC] Create per-IDL grpc packages

dongju-chae:
This patch creates per-IDL grpc packages to resolve dependency issues.

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -469,9 +469,31 @@ You can include Tizen sensor framework nodes as source elements of GStreamer/NNS
 %package grpc
 Summary:	NNStreamer gRPC support
 Requires:	nnstreamer = %{version}-%{release}
-Requires:	grpc-devel
+%if %{with tizen}
+Recommends:	nnstreamer-grpc-protobuf = %{version}-%{release}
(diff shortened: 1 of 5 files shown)


jaeyun-jung (approved):

packaging/nnstreamer.spec
```diff
@@ -472,6 +472,25 @@ Requires:	nnstreamer = %{version}-%{release}
 Requires:	grpc-devel
 %description  grpc
 NNStreamer's tensor_source/sink plugins for gRPC support.
+
```
myungjoo, line 475: Wait a sec... do we need grpc-devel for grpc? (LIne 473)

Fot Tizen, please add ```Recommends:``` on grpc-protobuf and grpc-flatbuf from grpc.

Reference: https://fedoraproject.org/wiki/Packaging:WeakDependencies

For fine control, you need to understand the following dependency keywords:
Requires, Recommends, Suggests,
Provides, Supplements, Enhances.

Note that TIzen heavily depends on the usages of Requires, Recommends, Suggests, and Provides.

dongju-chae: Oh.. my bad. don't need grpc-devel

myungjoo (approved):
LGTM

PR #2973, opened 2020-12-21
[Filter] release mem block in combi case

jaeyun-jung:
Possible res leak in output combination case, release mem block if output tensor is not in the combination list.


gichan-jang (approved):
How did you find it! Thank you for fixing it.

dongju-chae (approved):
LGTM

PR #2978, opened 2020-12-23
[gRPC/Async] Extend gRPC codes to support async/non-blocking APIs

### LINKED ISSUE #2972: Support for gRPC asynchronous RPC handling
Currently, gRPC support synchronous/blocking RPC handling. But, according to nnstreamer-edge scenarios, a single server may have to handle multiple clients at the same time. So, we have to extend the gRPC plugin to support asynchronous/non-blocking RPC handling.

- https://grpc.io/docs/languages/cpp/async/

dongju-chae:
This patch extends gRPC codes to support async/non-blocking APIs.

Let's discuss about how we can handle the tag of each RPC request
in the gstreamer pipeline.

Resolves #2972


ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
```diff
@@ -88,25 +89,154 @@ Status ServiceImplFlatbuf::_read_tensors (T reader)
...
+  gst_buffer_map (buffer, &map, GST_MAP_READ);
+
+  for (guint i = 0; i < num_tensors; i++) {
+    const GstTensorInfo * info = &config_->info.info[i];
+    gsize tsize = gst_tensor_info_get_size (info);
+
+    tensor_dim = builder.CreateVector (info->dimension, NNS_TENSOR_RANK_LIMIT);
+    tensor_name = builder.CreateString ("Anonymous");
+    tensor_type = (Tensor_type) info->type;
+    tensor_data = builder.CreateVector<unsigned char> (map.data + data_ptr, tsize);
```
jaeyun-jung, line 158: line 160 : condition to check data size after creating vector

ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
```diff
@@ -88,25 +89,154 @@ Status ServiceImplFlatbuf::_read_tensors (T reader)
...
+  std::vector<flatbuffers::Offset<Tensor>> tensor_vector;
+  Tensor_type tensor_type;
+
+  unsigned int num_tensors = config_->info.num_tensors;
+  frame_rate fr = frame_rate (config_->rate_n, config_->rate_d);
+
+  GstMapInfo map;
+  gsize data_ptr = 0;
+
+  gst_buffer_map (buffer, &map, GST_MAP_READ);
```
jaeyun-jung, line 149: plz handle returned value to avoid svace issue.
```
if (!gst_buffer_map (..))
 // handle error
```

ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
dongju-chae: Fixed, thanks.

ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
dongju-chae: Fixed, thanks.

ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
```diff
@@ -29,56 +27,59 @@
...
+/** @brief parse tensors and deliver the buffer via callback */
+void
+ServiceImplFlatbuf::parse_tensors (Message<Tensors> &tensors)
 {
-  return _read_tensors (reader);
+  GstBuffer *buffer;
+
+  _get_buffer_from_tensors (tensors, &buffer);
+
+  if (cb_)
```
jaeyun-jung, line 51: if cb is null, buffer should be released.
how about setting gst-buffer in if condition?
```
if (cb) {
  GstBuffer *buffer = get_buffer_from_tensor()
  cb (...)
}
```

ext/nnstreamer/extra/nnstreamer_grpc_common.h
```diff
@@ -59,6 +59,16 @@ class NNStreamerRPC {
       return handle_;
     }
 
+    /** @brief check it's server instance */
+    gboolean getIsServer () {
```
jaeyun-jung, line 63: minor comment, suggestion getIsServer > isServer

ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
jaeyun-jung: Also plz check the implementation in grpc-protobuf.
IMO others are ok.

dongju-chae: Oh, thanks. I will update.

ext/nnstreamer/extra/nnstreamer_grpc_common.h
dongju-chae: will be updated.

jaeyun-jung (approved):

myungjoo:
PTAL: @wooksong @anyj0527 @gichan-jang

anyj0527 (approved):
LGTM. Thanks!

PR #2980, opened 2020-12-28
[Converter] property to print sub-plugin list

jaeyun-jung:
add new property to get the sub-plugins list of tensor-converter.


dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM

PR #2982, opened 2020-12-28
[If] init property values

jaeyun-jung:
remove unnecessary value and init null to avoid exception case using gst-inspect.

### DIFF
--- a/gst/nnstreamer/tensor_if/gsttensorif.c
+++ b/gst/nnstreamer/tensor_if/gsttensorif.c
@@ -289,8 +289,11 @@ gst_tensor_if_init (GstTensorIf * tensor_if)
 
   tensor_if->num_srcpads = 0;
   tensor_if->srcpads = NULL;
-  tensor_if->custom.func = NULL;
-  tensor_if->custom.data = NULL;
+  tensor_if->cv_option = NULL;
+  tensor_if->then_option = NULL;
+  tensor_if->else_option = NULL;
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_if/gsttensorif.c
```diff
@@ -289,8 +289,11 @@ gst_tensor_if_init (GstTensorIf * tensor_if)
 
   tensor_if->num_srcpads = 0;
   tensor_if->srcpads = NULL;
-  tensor_if->custom.func = NULL;
-  tensor_if->custom.data = NULL;
+  tensor_if->cv_option = NULL;
+  tensor_if->then_option = NULL;
+  tensor_if->else_option = NULL;
+  memset (tensor_if->sv, 0, sizeof (tensor_if_sv_s) * 2);
+  memset (&tensor_if->custom, 0, sizeof (custom_cb_s));
```
dongju-chae, line 296: How about just initializing `tensor_if` itself at the beginning (line 278)?
```
memset (tensor_if, 0, sizeof (GstTensorIf));
```
Otherwise, we may have to repeat this again whenever new variables are added.

jaeyun-jung: cannot do this, passed ptr has parent object.

dongju-chae: Ohh right. my bad.

dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM!

PR #2991, opened 2021-01-07
[If] install header

jaeyun-jung:
add missed header for tensor-if custom callback.


gichan-jang (approved):
LGTM

wooksong (approved):

PR #2994, opened 2021-01-11
[Common] function to compare tensor info

jaeyun-jung:
Prepare API repo migration, add function to compare tensor info in nnstreamer plugin API.


dongju-chae (approved):
LGTM

myungjoo:
We are migrating ML-API to API.git Which of these are to be included (or referred) in ML-API?

jaeyun-jung:
This PR is unnecessary, keep the function in filter. I will update another PR.

PR #3002, opened 2021-01-14
[C-API] fix double free case

jaeyun-jung:
If filter sub-plugin allocates output tensor, single-shot notifies releasing output buffer when closing the single handle.
In this case, we dont need to free the output handle itself. Calling single-close and destroy-outdata functions will make an error.


jaeyun-jung:
This will fix seg-fault to run test with tensorflow (nnstreamer_capi_singleshot, invoke_04)

dongju-chae (approved):
LGTM

myungjoo (approved):

PR #3010, opened 2021-01-19
[Filter/Single] move tensor-filter single class

jaeyun-jung:
move tensor-filter single class to nnstreamer core directory


again4you (approved):
LGMT!

myungjoo (approved):

PR #3013, opened 2021-01-20
[debian] Add nnstreamer-api-common-dev package

zhoonit:
Currently, `capi-ml-common-devel` was added to rpm package but
corresponding package was not available in debian.

This patch adds `nnstreamer-api-common-dev` to the debian which
corresponds with `capi-ml-common-devel` in rpm.

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed [ ]Skipped
2. Run test: [X]Passed [ ]Failed [ ]Skipped

Signed-off-by: Jihoon Lee <jhoon.it.lee@samsung.com>

PR #3014, opened 2021-01-20
[chores] Merge platform/tizen_error into ml_error @open sesame 01/22 15:31

zhoonit:
- ~**#3013** [debian] Add nnstreamer-api-common-dev package~
- [Chores] Merge platform/tizen_error into ml_error
```
`ml-api-common.h` was depending on `tizen_error.h`.

So there was `platform/tizen_error.h` to simulate tizen error codes.

As `ml-api-common.h` is becomming a standalone package
, it is better off `ml-api-common.h` defines error code on it's
own rather depending on a certain platform.

This patch merges `platform/tizen_error` into `ml-api-common.h` to
reduce packaging complexity.

**Self evaluation:**

```

--- v2
merge reverted, instead created another copy of `ml-api-common.h` for packages in other platforms


api/capi/include/ml-api-common.h
```diff
@@ -47,6 +51,35 @@ typedef enum {
...
+ */
+#if !defined(ESTRPIPE)
+#define ESTRPIPE EPIPE
+#endif /* !defined(ESTRPIPE) */
+
+/**
+ * @brief Enumeration for the error codes of NNStreamer for None-tizen platform
+ */
+typedef enum {
+  ML_ERROR_NONE                 = 0, /**< Success! */
```
kparichay, line 70: let's define TIZEN_ERROR_* than ML_ERROR_* again
as adding a new error code, in this case, would require adding it twice in both cases.

zhoonit: This variant is reflection of 
https://github.com/nnstreamer/nntrainer/pull/243#discussion_r447501431

As `Tizen_*` is platform dependent, if we are open to bind other platform dependent error code, IMO, it make sense to keep it this way.

kparichay: cool.

kparichay (approved):
LGTM
Just remember that this will require tizen ACR

zhoonit:
> LGTM
> Note that this will require tizen ACR

@again4you Could you confirm if we require tizen ACR? This patch does not change the enum definition but source code changes, so I wonder.

api/capi/include/ml-api-common.h
```diff
@@ -18,11 +19,14 @@
 #ifndef __ML_API_COMMON_H__
 #define __ML_API_COMMON_H__
 
-#include <tizen_error.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus */
+
+#ifdef __TIZEN__
```
again4you, line 26: If this file is shipped on Tizen package, `#ifdef __TIZEN__` is not allowed.

again4you (changes requested):
If you think `tizen_error.h` is not a good name, what about using `ml_error.h` or other things?

api/capi/include/ml-api-common.h
zhoonit: Thank you for the confirmation, I have made a change, as per request.

as `ml-api-common.h` is a part of `ml-api-common` package. It should be available across platforms. So, I renamed it to be `ml-api-common.h`
 Could you take another look?

zhoonit:
v2--

merge reverted, instead created another copy of `ml-api-common.h` for packages for other platforms

api/capi/include/platform/ml-api-common.h
```diff
@@ -0,0 +1,56 @@
...
+  ML_ERROR_NONE                 = 0, /**< Success! */
+  ML_ERROR_INVALID_PARAMETER    = -EINVAL, /**< Invalid parameter */
+  ML_ERROR_STREAMS_PIPE         = -ESTRPIPE, /**< Cannot create or access the pipeline. */
+  ML_ERROR_TRY_AGAIN            = -EAGAIN, /**< The pipeline is not ready, yet (not negotiated, yet) */
+  ML_ERROR_UNKNOWN              = -0x40000000LL,  /**< Unknown error */
+  ML_ERROR_TIMED_OUT            = -ML_ERROR_UNKNOWN + 1,  /**< Time out */
+  ML_ERROR_NOT_SUPPORTED        = -ML_ERROR_UNKNOWN + 2, /**< The feature is not supported */
+  ML_ERROR_PERMISSION_DENIED    = -EACCES, /**< Permission denied */
+  ML_ERROR_OUT_OF_MEMORY        = -ENOMEM, /**< Out of memory (Since 6.0) */
+} ml_error_e;
```
myungjoo, line 54: The problem is that you are defining ML_ERROR_* twice in the code repo, which usually creates issues undetectable at build-time.

zhoonit:
I consulted @myungjoo and let it leave as it is then having definition to two different places:)

PR #3020, opened 2021-01-22
[Coverity] fix coverity issues

jaeyun-jung:
1. handle error case when failed to map the gst buffer.
2. init variable in cls constructor.


myungjoo (approved):
LGTM

anyj0527 (approved):
LGTM 👍🏼

PR #3024, opened 2021-01-25
[If/Api] api description for if callback

jaeyun-jung:
update description for tensor-if custom callback and set init value for condition.


gichan-jang (approved):
LGTM

myungjoo (approved):

PR #3034, opened 2021-01-28
[Decoder/BoundingBox/TF-SSD] Add output tensors mapping configuration option, plus minor fixes & enhancements

xroumegue:
Hello nnstreamer community,

Please find the few patches I had to develop to get [mobilenet_ssd_v2_coco](https://coral.googlesource.com/edgetpu/+/refs/heads/release-chef/test_data/mobilenet_ssd_v2_coco_quant_postprocess.tflite) neural network model imported from Google corail database running on iMX8MP NXP SoC.

The main enhancement is related to the default output tensor mapping within tensor decoder bounding boxes in TF-SSD mode, which is incompatible with [default object detection NN output tensors mapping delivered by tensor flow lite community](https://www.tensorflow.org/lite/models/object_detection/overview#output_signature) used nowadays.

Even though neural network output mapping could be reorganized offline to fit default tensor decoder bounding boxes mapping, this might be preferable to let the user specify optionally the output tensor mapping.
The option-3 is used to specify the mapping as the following:
option3=Bounding boxes locations tensor index **:** classes tensor index **:** score tensor index **:** number of detected object index

For instance, option3=0:1:2:3 follows [tflite object detection default mapping](https://www.tensorflow.org/lite/models/object_detection/overview#output_signature) convention.

Backward compatibility (similar as option3=3:1:2:0) is preserved if the option is not provided.

Additionally, clamping the bounding boxes location tensor values before processing was required to avoid segfault exception while drawing as this model might output some coordinates as negative value (-0.f), obviously incorrectly interpreted with a simple int cast.

To avoid too many incorrect object detection draw, a score detection threshold is applied as well on TF-SSD, which is configurable thanks to options3, adding ",threshold" to it.

Lastly, an off-by-one error while applying a sanity check on the detected class is fixed, such [label class "0"](https://raw.githubusercontent.com/google-coral/test_data/master/coco_labels.txt)  is assessed as valid.

1. Coding style: [*]Passed [ ]Failed [ ]Skipped
2. Build test: [*]Passed [ ]Failed [ ]Skipped
3. Run test: [*]Passed [ ]Failed [ ]Skipped
4. Tensor decoder bounding box unittest: [*]Passed [ ]Failed [ ]Skipped

Regards,
 Xavier

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -2,6 +2,7 @@
  * GStreamer / NNStreamer tensor_decoder subplugin, "bounding boxes"
  * Copyright (C) 2018 Samsung Electronics Co. Ltd.
  * Copyright (C) 2018 MyungJoo Ham <myungjoo.ham@samsung.com>
+ * Copyright 2021 NXP
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
@@ -98,6 +99,18 @@ typedef enum
 } bounding_box_modes;
 
 /**
+ * @brief TF SSD Output tensor feature mapping.
+ */
+typedef enum
+{
+  TF_SSD_BBOX_IDX_LOCATIONS = 0,
+  TF_SSD_BBOX_IDX_CLASSES = 1,
+  TF_SSD_BBOX_IDX_SCORES = 2,
+  TF_SSD_BBOX_IDX_NUM = 3,
+  TF_SSD_BBOX_IDX_UNKNOWN
+} tf_ssd_bbox_idx_t;
+
+/**
  * @brief List of bounding-box decoding schemes in string
  */
 static const char *bb_modes[] = {
@@ -109,7 +122,7 @@ static const char *bb_modes[] = {
 };
 
 /**
- * @brief Data structure for SSD boundig box info for tf-lite ssd model.
+ * @brief Data structure for SSD bounding box info for tf-lite ssd model.
  */
 typedef struct
 {
@@ -117,6 +130,15 @@ typedef struct
   char *box_prior_path; /**< Box Prior file path */
   gfloat box_priors[BOX_SIZE][TFLITE_SSD_DETECTION_MAX + 1]; /** loaded box prior */
 } properties_TFLite_SSD;
+
+/**
+ * @brief Data structure for SSD bounding box info for tf ssd model.
+ */
+typedef struct
+{
+  /* From option3, output tensor mapping */
+  gint tensor_mapping[TF_SSD_MAX_TENSORS];      /* Output tensor index mapping */
+} properties_TF_SSD;
 
 /**
  * @brief Data structure for boundig box info.
@@ -128,6 +150,7 @@ typedef struct
   union
   {
     properties_TFLite_SSD tflite_ssd; /**< Properties for tflite_ssd configured by option 1 + 3 */
+    properties_TF_SSD tf_ssd; /**< tf_ssd mode properties configuration settings */
   };
 
   /* From option2 */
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -465,10 +533,16 @@ _set_max_detection (bounding_boxes * data, const guint max_detection,
...
  *
  * [Tensorflow SSD Model]
- * The first tensor is num_detection. 1, ANY-TYPE
- * The second tensor is detection_classes. #MaxDetection, ANY-TYPE
- * The third tensor is detection_scores. #MaxDetection, ANY-TYPE
- * The fourth tensor is detection_boxes. BOX_SIZE : #MaxDetection, ANY-TYPE
+ * Tensors mapping is defined through option-3, with following syntax:
+ * LOCATIONS_IDX:CLASSES_IDX:SCORES_IDX:NUM_DETECTION_IDX
+ *
+ * Default configuration is: 3:2:1:0
```
myungjoo, line 539: Doesn't this configuration incur regression errors for deployed applications & unit-tests (the reason why CI's automated test cases of x64 are failing?)
The default configuration should be the configuration of previous versions. (would it be 3:1:2:0 ?)

myungjoo:
Unittest reports:
```

[Starting] nnstreamer_decoder_boundingbox
==================================================
    Test Group nnstreamer_decoder_boundingbox Starts.
[PASSED] 0:gst-launch of case 0
[PASSED] 0-1:TFLITESSD Decode 1
[PASSED] 0-2:TFLITESSD Decode 2
[FAILED][Critical] 0-1:TFSSD Decode 1
[FAILED][Critical] 0-2:TFSSD Decode 2
==================================================
[FAILED] Test Group nnstreamer_decoder_boundingbox has failed cases (2)

```

Your comits are not backward compatible. Probably the default values of the newly introduced options should be corrected.

myungjoo (changes requested):
The unittest results suggest that this is not backward-compatible.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
xroumegue: Thanks for your review.

I don't think this tensor mapping related change introduced the regression ( I checked this behaved the same without it on my board as this was failing..)

I would have rather thought to the changes related to valid class (0) and the applied threshold on the detected objects, which lead obviously to different decoder outputs.

But I am a bit confused as the decoder bounding boxes unittests passed on my device (aarch64, yocto zeus)..

I guess I have to reproduce the issue on an ubuntu build (x86_64) to progress further as the TAOS output logs is .. minimalist.

In the meantime, I might update the PR with less functional changes to narrow down what is wrong.

xroumegue: You were finally right, I made a stupid typo while coding the default values.. my attention was on first and last items of the tensors indexes :(

The threshold was causing issues as well to pass unit test, so the threshold can be set through option-3, and if not specified, all detected objects will be output so backward compatibility is preserved.

myungjoo (approved):
LGTM

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -954,7 +1066,7 @@ bb_decode (void **pdata, const GstTensorsConfig * config,
     return GST_FLOW_ERROR;
   }
 
-  /** reset the buffer with alpha 0 / black */
+                                                                /** reset the buffer with alpha 0 / black */
```
dongju-chae, line 1069: Mistake?

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -118,6 +132,18 @@ typedef struct
...
+ */
+typedef struct
+{
+  /* From option3, output tensor mapping */
+  gint tensor_mapping[TF_SSD_MAX_TENSORS];      /* Output tensor index mapping */
+  gfloat threshold;             /* Detection threshold */
+} properties_TF_SSD;
+
+
+
```
dongju-chae, line 146: Unnecessary new-lines

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
xroumegue: Indeed, not intentional.
This has been introduced on gst-indent commit, not sure why.

Fixed.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
xroumegue: Thanks for your review.

Fixed.

dongju-chae (approved):
LGTM :)

PR #3046, opened 2021-02-03
[Tizen/gRPC] add extra dependencies for gRPC 1.35.0

dongju-chae:
This patch adds extra dependencies for gRPC 1.35.0

CC: @leemgs 

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -220,7 +220,9 @@ BuildRequires:  pkgconfig(amlogic-vsi-npu-sdk)
 %endif
 
 %if 0%{?grpc_support}
-BuildRequires:  grpc-devel
+BuildRequires:	pkgconfig(re2)
+BuildRequires:	pkgconfig(libcares)
+BuildRequires:	grpc-devel
 %endif
 
 %if %{with tizen}


packaging/nnstreamer.spec
```diff
@@ -220,7 +220,9 @@ BuildRequires:  pkgconfig(amlogic-vsi-npu-sdk)
 %endif
 
 %if 0%{?grpc_support}
-BuildRequires:  grpc-devel
+BuildRequires:	pkgconfig(re2)
+BuildRequires:	pkgconfig(libcares)
+BuildRequires:	grpc-devel
```
myungjoo, line 225: If building packages with grpc-devel requires re2 and libcares, grpc package should declare that its devel packages requires the two. Or is it because of our internal needs?

myungjoo: Reference https://build.opensuse.org/package/view_file/devel:tools/grpc/grpc.spec?expand=1

myungjoo (changes requested):
You should not declare secondary dependencies unless it's really inevitable.

In this case, it's not inevitable.

packaging/nnstreamer.spec
dongju-chae: Yes, I know and totally agree with you. But, this was a request from the Tizen gRPC maintainer (by email).
So, then we just let him resolve dependency issues by himself.

dongju-chae:
Let's close this as gRPC packaging was changed :)

PR #3050, opened 2021-02-04
[Build] remove unnecessary option

jaeyun-jung:
Remove unnecessary build option for sub-plugins, default val is auto.


myungjoo (approved):
LGTM

anyj0527 (approved):
LGTM👍

PR #3054, opened 2021-02-08
[gRPC/Test] Skip SSAT tests on qemu/aarch64

dongju-chae:
This patch makes gbs skip ssat tests on qemu/aarch64 env.

In gRPC 1.35.0, some behaviors are not supported on qemu/aarch64.

See the below error log
```
[  119s] qemu: Unsupported syscall: 168
[  119s] E0208 04:45:45.854670498   21089 cpu_linux.cc:41] Error determining current CPU: Function not implemented
[  119s]
[  119s] Unsupported setsockopt level=1 optname=15
[  119s] E0208 04:45:45.874942631   21089 socket_utils_common_posix.cc:222] check for SO_REUSEPORT: {"created":"@1612759545.874068306","description":"Protocol not available","errno":92,"file":"/home/abuild/rpmbuild/BUILD/grpc-1.35.0/src/core/lib/iomgr/socket_utils_common_posix.cc","file_line":198,"os_error":"Protocol not available","syscall":"setsockopt(SO_REUSEPORT)"}
[  119s] qemu: uncaught target signal 4 (Illegal instruction) - core dumped
```

CC: @gichan-jang 

### DIFF
--- a/tests/nnstreamer_grpc/runTest.sh
+++ b/tests/nnstreamer_grpc/runTest.sh
@@ -18,6 +18,13 @@ fi
 # This is compatible with SSAT (https://github.com/myungjoo/SSAT)
 testInit $1
 
+# Skip test on qemu env.
+if [[ "$SKIP_QEMU_ARM64_INCOMPATIBLE_TESTS" == "1" ]]; then
+  echo "Skip ssat tests on qemu/arm64 env."
+  report
+  exit
+fi
+
 # Check gRPC availability
 PATH_TO_PLUGIN=${NNSTREAMER_BUILD_ROOT_PATH}
 if [[ -d $PATH_TO_PLUGIN ]]; then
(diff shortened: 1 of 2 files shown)


myungjoo (changes requested):
1. Call report before exit.
2. Rename ```SKIP_SSAT_TEST_ON_QEMU``` so that readers can understand what is going on.
 (e.g., ```SKIP_QEMU_ARM64_INCOMPATIBLE_TESTS``` or the reason causing the trouble.)
  The current naming suggests that we need to disable all SSAT TCs in QEMU

dongju-chae:
@myungjoo Thanks for the suggestion :) Updated.

tests/nnstreamer_grpc/runTest.sh
```diff
@@ -18,6 +18,13 @@ fi
 # This is compatible with SSAT (https://github.com/myungjoo/SSAT)
 testInit $1
 
+# Skip test on qemu env.
+if [[ "$SKIP_QEMU_ARM64_INCOMPATIBLE_TESTS" == "1" ]]; then
```
gichan-jang, line 22: How about deciding whether to perform the test in the shell?
```
arch=$(uname -m)
if [ "$arch" = "aarch64" ] ; then
...
```

dongju-chae: Well I think this approach is much easier to understand the trouble.
if we get similar problems in future, let's discuss about generalizing this.

gichan-jang (approved):

myungjoo (approved):
LGTM

myungjoo:
tizen-i586 error should be gone if you rebuild (CI tried openvino 6.29 while 6.30 is being released and 6.29 is removed from the package server)
Ubuntu error: let's ignore for now. We need someone else to look at Ubuntu build issue.

Issue #3055, opened 2021-02-08
NNStreamer memory leak is detected by valgrind


ryuseonghan:
Hi, I'm Seonghan Ryu of Samsung Research.

# Issue Description

- I'm using [Pre-built nnstreamer in artifactory](https://art.sec.samsung.net/artifactory/webapp/#/artifacts/browse/tree/General/odin_generic/nnstreamer/libs/2020-12-11_v1.6.0.3) in [ODIN-NLU](https://github.sec.samsung.net/ODIN/ODIN-NLU).
- There is no problem about execution and its result.
- However, memory leak has been reported about nnstreamer on 2/8 by valgrind.
- The leak is detected always.
- The leak is detected in both unittest and actual C++ app.
- The leak was generated by Linux app that uses nnstreamer linux library.
- There is no recent profiling result on Android by us.

Expected Result
============
There should not be any memory leaks.

How to Reproduce
===============
1. clone `dev_r20n` branch of https://github.sec.samsung.net/ODIN/ODIN-NLU
2. Build unittest using build option : `./build.sh -t gtest`
3. Run Valgrind on TC which uses nnstreamer: `valgrind --log-file=./report.memlog --show-possibly-lost=no --leak-check=full ./odin-nlu/bin/odin-test -d --gtest_filter=*TEST_INFERENCEENGINE_INFER_NORMAL_P*`

You can see the leaks present in the report file generated "20210208_cc_report.txt"

[20210208_cc_report.txt](https://github.com/nnstreamer/nnstreamer/files/5943729/20210208_cc_report.txt)

Logs
===============

```shell
==17827== Memcheck, a memory error detector
==17827== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==17827== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==17827== Command: ./odin-nlu/bin/odin-test --gtest_filter=*ODIN_NLU_CAPSULECLASSIFIER*
==17827== Parent PID: 5862
==17827== 
==17827== 
==17827== HEAP SUMMARY:
==17827==     in use at exit: 154,683 bytes in 584 blocks
==17827==   total heap usage: 36,632,693 allocs, 36,632,109 frees, 22,808,269,677 bytes allocated
==17827== 
==17827== 299 bytes in 23 blocks are definitely lost in loss record 344 of 372
[...]

PR #3057, opened 2021-02-09
[Filter] Fix memory leak

gichan-jang:
Fix mem leak of the tensor filter.
Related issue: #3055 

Self evaluation:

### DIFF
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -152,13 +152,9 @@ typedef struct _GstTensorFilterProperties
   unsigned int output_ranks[NNS_TENSOR_SIZE_LIMIT];  /**< the rank list of output tensors, it is calculated based on the dimension string. */
 
   const char *custom_properties; /**< sub-plugin specific custom property values in string */
-  union {
-    struct {
-      accl_hw *hw_list; /**< accelerators supported by framework intersected with user provided accelerator preference, use in GstTensorFilterFramework V1 only */
-      int num_hw;       /**< number of hardare accelerators in the hw_list supported by the framework */
-    };
-    const char *accl_str; /**< accelerator configuration passed in as parameter, use in GstTensorFilterFramework V0 only */
-  };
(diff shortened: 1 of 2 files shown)


jaeyun-jung:
We should keep current nn-framework version in filter-private struct.
Calling gst_tensor_filter_common_free_property() after closing nnfw causes res leak.

We cannot check fw version because fw is null, then prop->accl_str (for V0) prop->hw_list (for V1) is leak.
```
  if (GST_TF_FW_V0 (priv->fw)) {
    g_free_const (prop->accl_str);
  } else if (GST_TF_FW_V1 (priv->fw)) {
    g_free (prop->hw_list);
  }
```

gichan-jang:
> We should keep current nn-framework version in filter-private struct.
> Calling gst_tensor_filter_common_free_property() after closing nnfw causes res leak.
> 
> We cannot check fw version because fw is null, then prop->accl_str (for V0) prop->hw_list (for V1) is leak.
> 
> ```
>   if (GST_TF_FW_V0 (priv->fw)) {
>     g_free_const (prop->accl_str);
>   } else if (GST_TF_FW_V1 (priv->fw)) {
>     g_free (prop->hw_list);
>   }
> ```

Thanks! Updated.

jaeyun-jung (approved):
LGTM. Please add another PR to lts-1.6 branch after merging this to main branch.

dongju-chae:
FYI, Ubuntu build will be working after `pbuilder update` (compatible grpc version: `1.30.5-0~202102090627`).

myungjoo (commented):
I suggest:

1. add a flag in GstTensorFilterProperties for the union, which tells you which to free.
or
2. Un-union the two and free both. (set NULL when unused)

Then you don't need to add an element for the deallocated element, which is very weird and you can remove the if statement of Line 1006.

gichan-jang:
> I suggest:
> 
> 1. add a flag in GstTensorFilterProperties for the union, which tells you which to free.
>    or
> 2. Un-union the two and free both. (set NULL when unused)
> 
> Then you don't need to add an element for the deallocated element, which is very weird and you can remove the if statement of Line 1006.

Accl_str and hW_list are separated. Thanks.

myungjoo (approved):
LGTM

PR #3058, opened 2021-02-09
[grpc] Temporal disable grpc package

gichan-jang:
The grpc version is upgraded, but the grpc package of the nnstreamer ppa is not upgraded properly, so it is temporarily disabled.
After fixing the grpc package, revert this commit.

Discussed with @dongju-chae 

**Self evaluation:**
1. Build test: [ *]Passed [ ]Failed [*]Skipped
2. Run test: [ *]Passed [ ]Failed [*]Skipped


myungjoo (changes requested):
If Ubuntu PPA package is the problem, touch /debian/* files only. (Edit the rule file)

myungjoo:
@dongju-chae Can you update grpc related nnstreamer code for backward compatibility (making it compatible with both grpc 3.12 and grpc 3.14). You may get grpc version from meson and use #ifdef in grpc files.

dongju-chae:
Got it. When I come back to the office, let me check again.

myungjoo:
> Got it. When I come back to the office, let me check again.

If that seems to require too much effort, you can update meson to reject older grpc.
(We should reject incompatible grpc before actually compiling files)

gichan-jang:
> If Ubuntu PPA package is the problem, touch /debian/* files only. (Edit the rule file)

Right. Updated.

myungjoo (approved):
LGTM

jaeyun-jung (approved):

PR #3059, opened 2021-02-09
[C-Api/Tizen] restricted elements for internal function

jaeyun-jung:
Do not check restricted list when constructing a pipeline with internal API.


myungjoo (approved):
LGTM.
We have exempted only privilege checks for internal APIs. We need to exempt plugin checks, too!

gichan-jang (approved):
LGTM!

PR #3061, opened 2021-02-09
[OSX] Build error fix

myungjoo:
OSX does not have sys/auxv.h
Add a macro assuming that SIMD is always available for
OSX-AARCH64 (e.g., M1) devices.

Tested at MacBook-Air M1

### DIFF
--- a/gst/nnstreamer/hw_accel.c
+++ b/gst/nnstreamer/hw_accel.c
@@ -24,7 +24,16 @@
 #endif /* __TIZEN__ */
 #endif /* __arch64__ || __arm__ */
 
+#if !defined(__APPLE__)
 #include <sys/auxv.h>
+#else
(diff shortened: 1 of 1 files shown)


jaeyun-jung (approved):

gst/nnstreamer/hw_accel.c
```diff
@@ -24,7 +24,16 @@
 #endif /* __TIZEN__ */
 #endif /* __arch64__ || __arm__ */
 
+#if !defined(__APPLE__)
```
wooksong, line 27: Oh, we can use ```__APPLE__``` to distinguish macOS from the platforms. I need to change ```-D__MACOS__``` provided by the meson script.

wooksong (approved):
Looks great to me!

PR #3064, opened 2021-02-10
[Bug] Fix memory leak

gichan-jang:
Fix memory leak detected by valgrind.

Self evaluation:

### DIFF
--- a/gst/nnstreamer/tensor_split/gsttensorsplit.c
+++ b/gst/nnstreamer/tensor_split/gsttensorsplit.c
@@ -92,7 +92,7 @@ static void gst_tensor_split_set_property (GObject * object, guint prop_id,
     const GValue * value, GParamSpec * pspec);
 static void gst_tensor_split_get_property (GObject * object, guint prop_id,
     GValue * value, GParamSpec * pspec);
-static void gst_tensor_split_dispose (GObject * object);
+static void gst_tensor_split_finalize (GObject * object);
 
 #define gst_tensor_split_parent_class parent_class
 G_DEFINE_TYPE (GstTensorSplit, gst_tensor_split, GST_TYPE_ELEMENT);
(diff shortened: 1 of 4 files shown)


jaeyun-jung (approved):

gst/nnstreamer/tensor_split/gsttensorsplit.c
```diff
@@ -115,7 +115,7 @@ gst_tensor_split_class_init (GstTensorSplitClass * klass)
 
   parent_class = g_type_class_peek_parent (klass);
 
-  gobject_class->dispose = gst_tensor_split_dispose;
+  gobject_class->finalize = gst_tensor_split_finalize;
```
myungjoo, line 118: Please explain why are you are changing the callback.

myungjoo (commented):
LGTM except for the arise question with gsttensorsplit.c (please explain why "finalize" should be used instead of "dispose". Note that I don't know what "dispose" does, either :D  )

gst/nnstreamer/tensor_split/gsttensorsplit.c
gichan-jang: The GStreamer destruction process is split in two phases: the first phase, executed in the dispose handler is supposed to release all references to other member objects. The second phase, executed by the finalize handler is supposed to complete the object's destruction process. 
Currently, the tensor_split has only dispose function and no finalize function.
So, I changed to finalize the tensor_split instance.


Reference link 1: https://developer.gnome.org/gobject/unstable/gobject-memory.html#gobject-memory-cycles
Reference link 2: https://github.com/ImageMagick/glib/blob/52f7289456c08d74f50d22586a86ea2827d9eec3/gobject/gobject.h#L268

myungjoo (approved):

PR #3069, opened 2021-02-15
[C-Api] fix res leak in timeout mode

jaeyun-jung:
In single-shot timeout case, the flag to release output handle may reset when setting short timeout value.
To prevent this case, add output handle to del-list and change params to set input/output handle in invoke thread.

### DIFF
--- a/api/capi/src/nnstreamer-capi-single.c
+++ b/api/capi/src/nnstreamer-capi-single.c
@@ -106,7 +106,6 @@ typedef struct
   ml_tensors_data_h output;           /**< output to be sent back to user */
   guint timeout;                      /**< timeout for invoking */
   thread_state state;                 /**< current state of the thread */
-  gboolean ignore_output;             /**< ignore and free the output */
   gboolean free_output;               /**< true if output tensors are allocated in single-shot */
   int status;                         /**< status of processing */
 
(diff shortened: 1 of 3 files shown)


api/capi/src/nnstreamer-capi-util.c
```diff
@@ -570,8 +570,8 @@ ml_tensors_data_destroy (ml_tensors_data_h data)
 
   _data = (ml_tensors_data_s *) data;
 
-  if (_data->handle != NULL) {
-    status = ml_single_destroy_notify (_data->handle, _data);
+  if (_data->destroy) {
```
wooksong, line 573: Q: If checking ```_data->destory``` whether NULL or not, is it OK without the checking for ```_data``` ?

jaeyun-jung: It is already checked at line 568.

api/capi/src/nnstreamer-capi-single.c
```diff
@@ -1042,24 +1055,23 @@ _ml_single_invoke_internal (ml_single_h single,
...
       goto exit;
-    __process_output (single_h);
+    __process_output (single_h, single_h->output);
     single_h->state = IDLE;
   }
 
-  if (single_h->ignore_output == FALSE) {
+exit:
+  if (G_UNLIKELY (status != ML_ERROR_NONE)) {
+    ml_loge ("Failed to invoke the model.");
```
dongju-chae, line 1067: Q: at this point (e.g., when `__invoke` was failed), is it okay not to free output data?

jaeyun-jung: output handle is destroyed in __invoke() function, at line 246.

jaeyun-jung:
Now nnstreamer api is migrated to api-repo. I will reopen this PR.

PR #3070, opened 2021-02-16
[Build] disable c-api

jaeyun-jung:
disable c-api build.

TODO: we should remove api directory and related testcases later.


again4you (approved):
LGTM!

After applying this patch and modifying related packages, Tizen platform image is successfully created, and CAPI unit tests are passed.

zhoonit (approved):
LGTM!

PR #3075, opened 2021-02-17
[Build/Dev] internal header in cpp

jaeyun-jung:
add extern for cpp use in internal dev header.


gichan-jang (approved):

again4you (approved):
Nice catch!

Issue #3079, opened 2021-02-18
Unable to build with ArmNN option enabled


ggardet:
# Issue Description
I am unable to build with ArmNN option enabled. Meson does not find ArmNN.
Upstream now provides cmake files: https://github.com/ARM-software/armnn/commit/eac4adb5bf493ba7d1b00ef6ecd6b56a42ea3f3d

Expected Result
============
meson should find ArmNN cmake files

How to Reproduce
===============
Install ArmNN 20.11 and try to build nnstreamer against it.

Further Information
===============
* Exact OS version: openSUSE Tumbleweed on aarch64

dongju-chae:
The Tizen upstream is currently using armnn-20.05. So, we may change the build script and relevant codes to support both versions.

myungjoo:
@ggardet Does this commit (#3083 ) fix the issue?

ggardet:
> @ggardet Does this commit (#3083 ) fix the issue?

No, it did not helped. Files from `/usr/lib64/cmake/armnn/` are : `ArmnnConfig.cmake` `ArmnnTargets.cmake` `ArmnnTargets-relwithdebinfo.cmake`

dongju-chae:
@ggardet Could you share `build/meson-logs/meson-log.txt`?
`dependency('Armnn')` should find and open `ArmnnConfig.cmake`
like
```
...
  -- Found Armnn: /usr/lib/cmake/Armnn/ArmnnConfig.cmake
  -- Armnn Import: /usr/lib/cmake/Armnn/ArmnnTargets.cmake
...
```

ggardet:
@dongju-chae I get:
```
[...]

PR #3083, opened 2021-02-22
[meson] supports alternative target name to find pkg dependency @open sesame 02/22 15:50

dongju-chae:
This patch supports alternative target name to find pkg dependency.

For example, in case of armnn, the dependency name should be
`Armnn` not `armnn` to find its cmake config files.

Related issue: https://github.com/nnstreamer/nnstreamer/issues/3079

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -277,6 +277,7 @@ features = {
   },
   'armnn-support': {
     'target': 'armnn',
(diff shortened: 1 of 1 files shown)


myungjoo (approved):

gichan-jang (approved):
LGTM

PR #3093, opened 2021-02-25
[Dist/RPM] Fix meson path options.

### LINKED ISSUE #2642: Tizen/RPM-Spec's meson options are incorrect although they are harmless
Path designation to meson is incorrect in Tizen RPM-spec although meson's behavior is a bit awkward and there is no harmful effects. We need to fix this assuming that join_path("/abc", "/def/ghi") may results in "/abc/def/ghi").

myungjoo:
With the side-effects of meon join_paths,
absolute paths of bindir/libdir is working,
this is not recommended.

Fix the path inputs; use relative paths along with
prefix.

Fixes #2642

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -13,7 +13,7 @@
 
 %define		gstpostfix	gstreamer-1.0
 %define		gstlibdir	%{_libdir}/%{gstpostfix}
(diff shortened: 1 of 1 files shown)


dongju-chae (approved):
LGTM

anyj0527 (approved):
LGTM✔ Thanks!

PR #3123, opened 2021-03-10
[Transform] tensor type in stand mode

jaeyun-jung:
Currently only float32 can use stand mode in tensor-transform.

Add option to set output type and calc stand mode.

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -200,7 +200,7 @@ gst_tensor_transform_mode_get_type (void)
             "option=D1\':D2\':D3\':D4 (fixed to 3)",
           "transpose"},
       {GTT_STAND, "Mode for statistical standardization of tensor, "
-            "option=default",
+            "option=default[:TYPE]",
           "stand"},
       {GTT_CLAMP, "Mode for clamping all elements of tensor into the range, "
             "option=CLAMP_MIN:CLAMP_MAX",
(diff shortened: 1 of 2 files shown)


jaeyun-jung:
TODO
tensor-if / tensor-transform uses similar struct to handle tensor value.
add new functions for common struct.

gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -200,7 +200,7 @@ gst_tensor_transform_mode_get_type (void)
             "option=D1\':D2\':D3\':D4 (fixed to 3)",
           "transpose"},
       {GTT_STAND, "Mode for statistical standardization of tensor, "
-            "option=default",
+            "option=default[:TYPE]",
```
myungjoo, line 203: This might be a bit difficult for users to understand.
Please update ```gst/nnstreamer/tensor_transform/README.md``` as well.

jaeyun-jung: I will update this on another PR (#3129 to add new option 'DC remove')

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -918,8 +918,17 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
     }
     case GTT_STAND:
     {
-      filter->data_stand.mode =
-          gst_tensor_transform_get_stand_mode (filter->option);
+      gchar **strv = NULL;
+
+      strv = g_strsplit (filter->option, ":", -1);
```
dongju-chae, line 923: Q: Is it possible that `filter->option` is a empty string? Otherwise, need to check this first.

jaeyun-jung: line 740 checks this case.

dongju-chae: I meant a empty string `""` not nullptr.

jaeyun-jung: then gst_tensor_transform_get_stand_mode() will return STAND_END, which means error case.

dongju-chae: I think `strv[0]` can be a problem. 
from the GLib manual
```
As a special case, the result of splitting the empty string "" is an empty vector, not a vector containing a single string. The reason for this special case is that being able to represent an empty vector is typically more useful than consistent handling of empty elements. If you do need to represent empty elements, you'll need to check for the empty string before calling g_strsplit().
```

jaeyun-jung: I will make another PR to handle option string is empty string.

dongju-chae: Got it

dongju-chae (approved):
LGTM

zhoonit (approved):
LGTM :)

myungjoo (approved):

PR #3125, opened 2021-03-10
Enable detection for nntrainer singleshot

zhoonit:
This patch maps *.ini as nntrainer to utilize auto detection for
nntrainer.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -1198,6 +1198,8 @@ gst_tensor_filter_detect_framework (const gchar * const *model_files,
       detected_fw = g_strdup ("python");
     else if (g_str_equal (ext[0], ".graph"))
       detected_fw = g_strdup ("movidius-ncsdk2");
+    else if (g_str_equal (ext[0], ".ini"))
+      detected_fw = g_strdup ("nntrainer");
     else if (g_str_equal (ext[0], ".circle"))
       detected_fw = g_strdup ("nnfw");
     else if (g_str_equal (ext[0], NNSTREAMER_SO_FILE_EXTENSION))


myungjoo:
Does a nntrainer model consist of a single file, ".ini", only?

zhoonit:
> Does a nntrainer model consist of a single file, ".ini", only?

yes. there is weight file(.bin) but it is referenced inside ".ini" file.

jaeyun-jung (approved):

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1198,6 +1198,8 @@ gst_tensor_filter_detect_framework (const gchar * const *model_files,
       detected_fw = g_strdup ("python");
     else if (g_str_equal (ext[0], ".graph"))
       detected_fw = g_strdup ("movidius-ncsdk2");
+    else if (g_str_equal (ext[0], ".ini"))
```
wooksong, line 1201: IMO, since the extension names such as ```ini``` are too broadly used, it would be better to add more conditions to check whether the given file is for entrainer or not.

zhoonit: I agree that it is quite broad. Could you give me a suggestion to detect the framework? 
or could you recommend other supplementary choice?
1. to not support auto detection for now and comment this section
2. to warn user that it is selected as nntrainer but it might not be intended

gichan-jang: If other frameworks use the same .ini extension later, it would be good to use priority like tflite.

gichan-jang (approved):

gst/nnstreamer/tensor_filter/tensor_filter_common.c
myungjoo: *.nntr? or *.nnt?

helloahn: `.nnt` is already in use.
PTAL, https://fileinfo.com/extension/nnt

myungjoo (approved):
LGTM. We may start with .ini and let's update later if required.

zhoonit:
Cool :) Marking this as mergable PR

PR #3129, opened 2021-03-11
[Transform] option for DC removal

jaeyun-jung:
Add new option for DC removal (average value) in stand mode.

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -124,6 +124,7 @@ enum
 
 static const gchar *gst_tensor_transform_stand_string[] = {
   [STAND_DEFAULT] = "default",
+  [STAND_DC_AVERAGE] = "dc-average",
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -701,7 +702,7 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
 
       if (filter->data_stand.mode == STAND_END) {
         ml_loge
-            ("%s: stand: \'%s\' is not valid option string: it should be \'default\', currently the only supported mode.\n",
+            ("%s: stand: \'%s\' is not valid option string: it should be in the form of (default|dc-average)[:TYPE]\n",
```
wooksong, line 705:
```suggestion
            ("%s: stand: \'%s\' is not a valid option string: it should be in the form of (default|dc-average)[:TYPE]\n",
```

wooksong (approved):

PR #3134, opened 2021-03-12
[Decoder] file content size

jaeyun-jung:
When loading data from file, use param to get the contents size.


gichan-jang (approved):

dongju-chae (approved):
LGTM

PR #3135, opened 2021-03-12
[Transform] handle empty option string case

jaeyun-jung:
If option string is empty, tensor-transform cannot parse the mode and will crash application.
Add error case for empty option string.

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -740,6 +740,11 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
   if (filter->mode == GTT_UNKNOWN || filter->option == NULL)
     return TRUE;
 
+  if (strlen (filter->option) == 0) {
+    ml_loge ("Given option string is empty, you should set proper option.");
+    return FALSE;
+  }
+
   filter_name = gst_object_get_name ((GstObject *) filter);
 
   switch (filter->mode) {


gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -740,6 +740,11 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
   if (filter->mode == GTT_UNKNOWN || filter->option == NULL)
     return TRUE;
 
+  if (strlen (filter->option) == 0) {
+    ml_loge ("Given option string is empty, you should set proper option.");
+    return FALSE;
```
myungjoo, line 745: If ```filter->option == NULL``` is fine, can we consider ```filter->option == ""``` equally with NULL?

jaeyun-jung: Default value of filter->option is null, which means it is not parsed yet in this line. (or..developer does not set an option string.

wooksong: @jaeyun-jung Is it necessary? IMO, ```g_regex_match_simple()``` handles the case where ```strlen(filter->option)``` is ```0```?

jaeyun-jung: right, it is better to add new regex for stand mode. (currently no pattern for stand mode)

wooksong: @jaeyun-jung Understood. After merging this PR, I will add ```g_regex_match_simple()``` for the stand mode (and others if any). Thanks.

wooksong (approved):
LGTM!

anyj0527 (approved):
LGTM

gst/nnstreamer/tensor_transform/tensor_transform.c
jaeyun-jung: @wooksong plz consider new option for dc-offset removal (#3129)

PR #3147, opened 2021-03-23
[RPM] Add virtual conf package @open sesame 03/23 14:18

zhoonit:
Add virtual conf package to resolve build dependency issue as suggested
in https://github.com/nnstreamer/api/pull/21#issuecomment-804599118

With this patch, it is now possible to install devel package with
`/etc/nntrainer.ini` or the packager choose to use `test-devel` when
building.

See also nnstreamer/nntrainer#1034

**Evaluation Method**
Cross building with nntrainer package

**Self evaluation:**

Cc: Myungjoo Ham <myungjoo.ham@samsung.com>

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -128,6 +128,7 @@ Source1001:	nnstreamer.manifest
 ## Define requirements ##
 Requires: nnstreamer-core = %{version}-%{release}
 Requires: nnstreamer-configuration = %{version}-%{release}
+Recommends: nnstreamer-default-configuration = %{version}-%{release}
 
 ## Define build requirements ##
 BuildRequires:	gstreamer-devel
@@ -276,10 +277,12 @@ Summary: NNStreamer core package
 NNStreamer is a set of gstreamer plugins to support general neural networks
 and their plugins in a gstreamer stream, this package is core package without configuration
 
-%package configuration
+%package default-configuration
 Summary: NNStreamer global configuration
-%description configuration
-NNStreamer's configuration setup for the end user.
+Provides: nnstreamer-configuration = %{version}-%{release}
+Conflicts: nnstreamer-test-devel
+%description default-configuration
+NNStreamer's global configuration setup for the end user.
 
 # for tensorflow
 %if 0%{?tensorflow_support}
@@ -387,7 +390,7 @@ NNStreamer's tensor_fliter subplugin of caffe2
 
 %package devel
 Summary:	Development package for custom tensor operator developers (tensor_filter/custom)
-Requires:	nnstreamer-core = %{version}-%{release}
+Requires:	nnstreamer = %{version}-%{release}
 Requires:	glib2-devel
 Requires:	gstreamer-devel
 %description devel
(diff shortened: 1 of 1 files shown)


zhoonit:
@again4you I think this change shouldn't affect the build system, but could you review if this patch would provide any build crashes on VD or rootstrap?

packaging/nnstreamer.spec
```diff
@@ -128,6 +128,7 @@ Source1001:	nnstreamer.manifest
 ## Define requirements ##
 Requires: nnstreamer-core = %{version}-%{release}
 Requires: nnstreamer-configuration = %{version}-%{release}
+Recommends: nnstreamer-global-configuration = %{version}-%{release}
```
myungjoo, line 131: ```s/global/default```.

zhoonit: Done :)

packaging/nnstreamer.spec
```diff
@@ -388,6 +391,8 @@ NNStreamer's tensor_fliter subplugin of caffe2
 %package devel
 Summary:	Development package for custom tensor operator developers (tensor_filter/custom)
 Requires:	nnstreamer-core = %{version}-%{release}
+Requires:	nnstreamer-configuration = %{version}-%{release}
+Recommends: nnstreamer-test-devel = %{version}-%{release}
```
myungjoo, line 395: Why do you recommend a test package with devel? test-devel is for gbs-unit-test only.

zhoonit: Right, we don't even need to only `require: nnstreamer-core` so reverted this back to `Requires: nnstreamer`

packaging/nnstreamer.spec
```diff
@@ -411,7 +414,8 @@ Static library package of nnstreamer-devel.
 %package test-devel
 Summary: Development package to provide testable environment of a subplugin (tensor_filter/custom)
 Requires: nnstreamer-devel = %{version}-%{release}
-Conflicts: nnstreamer-configuration
+Provides: nnstreamer-configuration = %{version}-%{release}
```
kparichay, line 417: basic question - where is `nnstreamer-configuration` package defined? 
Also, both `default-configuration` and `test-devel` packages `conflicts` each other but in the end are providing the same package - `nnstreamer-configuration`. what am I missing here?

zhoonit: This is called virtual package.

So, `test-devel` and `default-configuration` provides the capability of `nnstreamer-configuration` which is virtual capability 

When installing nnstreamer, it is `recommended` to have `default-configuration` to satisfy the capability of `nnstreamer-configuration` but even if our system has `test-devel`, nnstreamer will be happy to accept it as a `nnstreamer-configuration`, so this makes the other packager to choose which one to include at will.

So, when we need to have test environment, we require `nnstreamer-test-devel`. The dependency solver will be happy to have `nnstreamer-test-devel` instead of `nnstreamer-default-configuration` to satisfy `nnstreamer-configuration`

referring to below.
https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-advanced-packaging.html#id818305

https://fedoraproject.org/wiki/Packaging:WeakDependencies#Real_life_example

kparichay: Got it. Thanks for the explanation.

kparichay: It might be helpful to add a note that `nnstreamer-configuration` is a virtual capability.

zhoonit: Updated :)

myungjoo (approved):

dongju-chae (approved):
LGTM

PR #3152, opened 2021-03-25
[Spec] Use %{name} macro instead of string

again4you:
This patch uses %{name} macro instead of a string. It causes build
failures on VD QBuild.

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -127,13 +127,13 @@ Release:	0
 Group:		Applications/Multimedia
 Packager:	MyungJoo Ham <myungjoo.ham@samsung.com>
 License:	LGPL-2.1
-Source0:	nnstreamer-%{version}.tar.gz
+Source0:	%{name}-%{version}.tar.gz
 Source1:	generate-tarball.sh
-Source1001:	nnstreamer.manifest
+Source1001:	%{name}.manifest
 
 ## Define requirements ##
-Requires: nnstreamer-core = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


myungjoo:
Wow... this really was the cause?

myungjoo:
@jijoongmoon @kparichay Please take a look at this. NNTrainer may experience the same issue when you move on to VD.

myungjoo (approved):
With the standard implementation of rpmbuild and GBS/OBS, this should behave exactly equivalently with the previous .spec file. However, the implementation of QuickBuild appears not complying the standard. Why in the hell are they using QuickBuild anyway... :(

LGTM.

packaging/nnstreamer.spec
```diff
@@ -772,7 +772,7 @@ cp -r result %{buildroot}%{_datadir}/nnstreamer/unittest/
 %files
 
 %files core
-%manifest nnstreamer.manifest
+%manifest %{name}.manifest
```
kparichay, line 775: I wonder if this should be updated in other locations as well like line 793, 800, etc?

kparichay:
> @jijoongmoon @kparichay Please take a look at this. NNTrainer may experience the same issue when you move on to VD.

Applied to NNTtrainer with https://github.com/nnstreamer/nntrainer/pull/1048.

packaging/nnstreamer.spec
again4you: You are right. I will fix it if this patch is necessary.

PR #3154, opened 2021-03-26
[TRANSFORM] Changed to support tensors

gichan-jang:
Changd tensor transform to support tensors
    
### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -28,8 +28,8 @@
 /**
  * SECTION:element-tensor_transform
  *
- * A filter that transforms tensor dimension or type.
- * The input and output is always in the format of other/tensor
+ * A filter that transforms tensors dimension or type.
+ * The input and output is always in the format of other/tensor or other/tensors.
  *
  * <refsect2>
  * <title>Example launch line</title>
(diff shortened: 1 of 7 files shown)


jaeyun-jung (approved):

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -1606,13 +1662,15 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
...
    */
-  if (in_config.info.type == _NNS_INT64 ||
-      in_config.info.type == _NNS_UINT64 ||
-      out_config.info.type == _NNS_INT64 ||
-      out_config.info.type == _NNS_UINT64) {
-    filter->orc_supported = FALSE;
-  } else {
-    filter->orc_supported = TRUE;
+  for (i = 0; i < in_config.info.num_tensors; i++) {
+    if (in_config.info.info[i].type == _NNS_INT64 ||
```
jaeyun-jung, line 1666: TODO (later work)
we can control orc acceleration for multi-tensors with in/out type.
I will try to handle each buffer with orc after this PR.

myungjoo (approved):
LGTM.

Note somewhere that, in the future, you need to add a property that selects which tensors of the vien tensors are transformed. (e.g., tensor 0 and 2 among 0, 1, 2, and 3.)

PR #3162, opened 2021-03-29
[Transform/Orc] orc acceleration for tensors

jaeyun-jung:
condition to support orc acceleration for each tensor.


gichan-jang (approved):
LGTM

dongju-chae (approved):
LGTM

PR #3171, opened 2021-03-30
[TizenSensor] init sensor type

jaeyun-jung:
set init value of sensor type. 'all' means no specific sensor and if user does not set type in pipeline, it will be failed.


gichan-jang (approved):
We missed it..! LGTM

again4you (approved):
Nice catch!

PR #3173, opened 2021-03-31
[CodeQuality] add NULL checker after allocations

helloahn:
Like others, NULL checker after memory allocations have been added.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_source/tensor_src_grpc.c
+++ b/ext/nnstreamer/tensor_source/tensor_src_grpc.c
@@ -290,6 +290,7 @@ _grpc_callback (void *obj, void *data)
   GST_BUFFER_PTS (buffer) = timestamp;
 
   item = g_new0 (GstDataQueueItem, 1);
+  g_assert (item != NULL);
   item->object = GST_MINI_OBJECT (buffer);
   item->size = gst_buffer_get_size (buffer);
   item->visible = TRUE;
@@ -343,6 +344,7 @@ gst_tensor_src_grpc_init (GstTensorSrcGRPC * self)
   self->out = 0;
 
   self->priv = g_new0 (grpc_private, 1);
+  g_assert (self->priv != NULL);
   grpc_config_init (self);
 
   GST_OBJECT_FLAG_UNSET (self, GST_TENSOR_SRC_GRPC_CONFIGURED);
--- a/gst/nnstreamer/tensor_if/gsttensorif.c
+++ b/gst/nnstreamer/tensor_if/gsttensorif.c
@@ -380,6 +380,7 @@ gst_tensor_if_set_property_supplied_value (const GValue * value,
   }
 
   param = g_value_get_string (value);
+  g_assert (param != NULL);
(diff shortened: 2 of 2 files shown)


anyj0527 (approved):
LGTM ✔

dongju-chae:
Q: Does any static analyzer report that they're defects?

ext/nnstreamer/tensor_source/tensor_src_grpc.c
```diff
@@ -290,6 +290,7 @@ _grpc_callback (void *obj, void *data)
   GST_BUFFER_PTS (buffer) = timestamp;
 
   item = g_new0 (GstDataQueueItem, 1);
+  g_assert (item != NULL);
```
dongju-chae, line 293: Q: Does any static analyzer report that they're defects?

kimjh12 (approved):
LGTM!!

ext/nnstreamer/tensor_source/tensor_src_grpc.c
helloahn: Yes, they did.
Though there are many cases that have no NULL checker after allocations, the recommendation is it.
It sounds quite weird, but they report only these cases as a defect. :sweat_smile:

dongju-chae (approved):
LGTM

myungjoo (changes requested):
Do not add assertions into production library code if such errors may be caused by library users.
Do proper error handling (messages and returns)

In production library code, assertions are allowed ONLY IF such assertion may be caused by internal errors.

ext/nnstreamer/tensor_source/tensor_src_grpc.c
again4you: What about setting it as **False-Positive** instead of adding assertion?

wooksong, line 293: How about handling the NULL cases instead of using ```g_assert```

dongju-chae:
@myungjoo I think they are `internal errors` because `g_new0` never returns `NULL`.
I heard that this patch is for static analyzer's false negative problems.

myungjoo:
> @myungjoo I think they are `internal errors` because `g_new0` never returns `NULL`.
> I heard that this patch is for static analyzer's false positive problems.

Then, mark them "false positive". Refer to https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html (Description above g_new() function description)

Tell SE people that their tools are again reporting false positive and tell them to fix this.

helloahn:
> > @myungjoo I think they are `internal errors` because `g_new0` never returns `NULL`.
> > I heard that this patch is for static analyzer's false positive problems.
> 
> Then, mark them "false positive". Refer to https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html (Description above g_new() function description)
> 
> Tell SE people that their tools are again reporting false positive and tell them to fix this.

Got it.
I'll close this PR because it only contains the same cases.

PR #3181, opened 2021-04-05
Change gboolean comparison

gichan-jang:
Never directly compare the contents of a gboolean variable with the values TRUE or FALSE.
Use `if (condition)` to check a gboolean is "true", instead of `if (condition == TRUE)`. Likewise use `if (!condition)` to check a gboolean is "false".
There is no validation when assigning to a gboolean variable and so it could contain any value represented by a gint. This is why the use of if (condition) is recommended.
All non-zero values in C evaluate to "true".

Refer: https://developer.gnome.org/glib/stable/glib-Basic-Types.html#gboolean

Self evaluation:


kimjh12 (approved):
LGTM!!

myungjoo (approved):
LGTM

jaeyun-jung (approved):

PR #3183, opened 2021-04-06
[Mux/Merge] change property name

jaeyun-jung:
change property name (sync mode and option), remove underbar in property name.

### DIFF
--- a/Documentation/synchronization-policies-at-mux-merge.md
+++ b/Documentation/synchronization-policies-at-mux-merge.md
@@ -30,7 +30,7 @@ For the case of merge/mux element which has three pads with different framerate 
 
 # Slowest
 
-"Slowest" policy (sync_mode=slowest) synchronize tensors based on slowest timestamp among pads.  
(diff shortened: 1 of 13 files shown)


gichan-jang (approved):
LGTM!
These changes synced with element properties of the tensor mux and merge.

myungjoo:
Can you make it backward compatible?

CC: @jaeyun-jung @again4you Wouldn't this make problems in the outside and clients?

jaeyun-jung:
Tizen TCT sync mode and option is not used.
odin only uses single-shot functions, so no problem.
Testcases and sample in nnstreamer-example / api repo need update for this change. (I found 3 files to be modified)

@helloahn plz check esar pipeline includes mux and merge elements.

myungjoo:
There are nnstreamer pipeline users (both GST-API and ML-API) making prototypes. We do not have access to their applications.

jaeyun-jung:
`sync-mode` and `sync_mode` both work well :)

again4you (approved):
LGTM!

PR #3213, opened 2021-04-23
[Converter] support flex tensor

jaeyun-jung:
Add pad template to handle flex tensors.
tensor-converter supports changing capability from flex to static, or static to flex if downstream element has flex-tensor caps.

### DIFF
--- a/gst/nnstreamer/tensor_converter/tensor_converter.c
+++ b/gst/nnstreamer/tensor_converter/tensor_converter.c
@@ -78,6 +78,12 @@
     gst_caps_append (caps, gst_caps_from_string (OCTET_CAPS_STR))
 
 /**
+ * @brief Macro to append template caps for flexible tensor
+ */
+#define append_flex_tensor_caps_template(caps) \
+    gst_caps_append (caps, gst_caps_from_string (GST_TENSORS_FLEX_CAP_DEFAULT))
+
+/**
  * @brief Macro for debug mode.
  */
 #ifndef DBG
@@ -287,8 +293,8 @@ gst_tensor_converter_class_init (GstTensorConverterClass * klass)
 
   /* set src pad template */
   pad_caps =
-      gst_caps_from_string (GST_TENSOR_CAP_DEFAULT "; "
-      GST_TENSORS_CAP_DEFAULT);
+      gst_caps_from_string (GST_TENSOR_CAP_DEFAULT ";"
+      GST_TENSORS_CAP_DEFAULT ";" GST_TENSORS_FLEX_CAP_DEFAULT);
 
   pad_template = gst_pad_template_new ("src", GST_PAD_SRC, GST_PAD_ALWAYS,
       pad_caps);
@@ -304,6 +310,7 @@ gst_tensor_converter_class_init (GstTensorConverterClass * klass)
   append_audio_caps_template (pad_caps);
   append_text_caps_template (pad_caps);
   append_octet_caps_template (pad_caps);
+  append_flex_tensor_caps_template (pad_caps);
 
   /* append sub-plugin template caps */
   str_array = get_all_subplugins (NNS_SUBPLUGIN_CONVERTER);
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -319,6 +319,10 @@ gst_tensor_converter_class_init (GstTensorConverterClass * klass)
     g_strfreev (str_array);
   }
 
+  /* append caps string for flexible tensors */
+  gst_caps_append (pad_caps,
+      gst_caps_from_string (GST_TENSORS_FLEX_CAP_DEFAULT));
```
myungjoo, line 324: Can you do this with the style of LINE 302-206 if this capability is built-in?

myungjoo (commented):
LGTM except for the (mentioned) possible style inconsistency.

Anyway, questions:
Q1. what happens if the pipeline is ```... static-tensor ! tensor-converter ! static-tensor ...```?
Q2. what happens if the pipeline is ```... flex-tensor ! tensor-converter ! flex-tensor ...```?
Q3. what happens if the pipeline is ```... flex-tensor ! tensor-converter ! static-tensor ...``` and the buffer via flex-tensor at nth frame is 100 bytes while stat-tensor has 10 bytes in its size?
Q4. what happens if the pipeline is ```... flex-tensor ! tensor-converter ! static-tensor ...``` and the buffer via flex-tensor at nth frame is 100 bytes while static-tensor has 1000 bytes in its size?

Corresponding behaviors (Q1~Q4) need to be documented later. (we have tensor type document)

gst/nnstreamer/tensor_converter/tensor_converter.c
jaeyun-jung: updated

jaeyun-jung:
> LGTM except for the (mentioned) possible style inconsistency.
> 
> Anyway, questions:
> Q1. what happens if the pipeline is `... static-tensor ! tensor-converter ! static-tensor ...`?
> Q2. what happens if the pipeline is `... flex-tensor ! tensor-converter ! flex-tensor ...`?
> Q3. what happens if the pipeline is `... flex-tensor ! tensor-converter ! static-tensor ...` and the buffer via flex-tensor at nth frame is 100 bytes while stat-tensor has 10 bytes in its size?
> Q4. what happens if the pipeline is `... flex-tensor ! tensor-converter ! static-tensor ...` and the buffer via flex-tensor at nth frame is 100 bytes while static-tensor has 1000 bytes in its size?
> 
> Corresponding behaviors (Q1~Q4) need to be documented later. (we have tensor type document)

Q1. negotiation failure (converter does not have static caps on sink pad)
Q2. Possible, converter just checkes the incoming buffer size.
Q3 and Q4. Failed, incoming buffer size should be exactly matched with converter property.

I will update wiki and readme about flex tensor later.

myungjoo:
> converter just checkes the incoming buffer

This will probably be future work. However,.... shouldn't Q4 be supported if the user provides (new) properties that declare which dimension to be "expanded"? (we may have a working default value for this as well)

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -319,6 +325,9 @@ gst_tensor_converter_class_init (GstTensorConverterClass * klass)
     g_strfreev (str_array);
   }
 
+  /* append caps string for flexible tensors */
+  append_flex_tensor_caps_template (pad_caps);
```
myungjoo, line 329: Move to Line 313?

jaeyun-jung: updated

jaeyun-jung:
> > converter just checkes the incoming buffer
> 
> This will probably be future work. However,.... shouldn't Q4 be supported if the user provides (new) properties that declare which dimension to be "expanded"? (we may have a working default value for this as well)

add todo to change mem block if incoming data is tensor format.

myungjoo (approved):
LGTM

Issue #3196, opened 2021-04-12
MQTT Feature Flash Request: TimeStamp


myungjoo:
Need a general design that is ready for additional metadata, including "network latency", "profile data", ...

Functional Requirement
- MQTT-SINK --> MQTT-SRC should preserve timestamp information.

CC: @helloahn @again4you 

We will be relying on NTP to synchronize clocks across nodes.
1. However, to preserve timestamp information from a previous node and make it valid to the current node, MQTT-SRC/SINK should have an additional mechanism to "encode" and "decode" abstract timestamp <--> relative timestamp.
2. Need to be careful with clock drifts. (may need to design 1. robust enough or reset relative timestamps sometimes (as if "FastForward/Reverse" is done)

wooksong:
FYI, I attached rtspsrc's timestamp-related properties here.

```bash
$ gst-inspect-1.0 rtspsrc
...
  buffer-mode         : Control the buffering algorithm in use
                        flags: readable, writable
                        Enum "GstRTSPSrcBufferMode" Default: 3, "auto"
                           (0): none             - Only use RTP timestamps
                           (1): slave            - Slave receiver to sender clock
                           (2): buffer           - Do low/high watermark buffering
                           (3): auto             - Choose mode depending on stream live
                           (4): synced           - Synchronized sender and receiver clocks
  ntp-sync            : Synchronize received streams to the NTP clock
                        flags: readable, writable
                        Boolean. Default: false
  use-pipeline-clock  : Use the pipeline running-time to set the NTP time in the RTCP SR messages(DEPRECATED: Use ntp-time-source property)
                        flags: readable, writable, deprecated
                        Boolean. Default: false
  ntp-time-source     : NTP time source for RTCP packets
                        flags: readable, writable
                        Enum "GstRTSPSrcNtpTimeSource" Default: 0, "ntp"
                           (0): ntp              - NTP time based on realtime clock
                           (1): unix             - UNIX time based on realtime clock
                           (2): running-time     - Running time based on pipeline clock
                           (3): clock-time       - Pipeline clock time
  max-rtcp-rtp-time-diff: Maximum amount of time in ms that the RTP time in RTCP SRs is allowed to be ahead (-1 disabled)
                        flags: readable, writable
                        Integer. Range: -1 - 2147483647 Default: 1000 
  rfc7273-sync        : Synchronize received streams to the RFC7273 clock (requires clock and offset to be provided)
                        flags: readable, writable
[...]

PR #3225, opened 2021-04-28
[Gst/MQTT] Support timestamp synchronization between nodes and dynamic buffer re-allocation

wooksong:
This PR addresses #3196, which is timestamp synchronization between MQTT pipeline nodes. Other changes included in this PR as follows:
- Introduce a new property, max-buffer-size
  - Using this property, mqttsink could exploit a static-sized buffer instead of dynamically allocating buffers according to the changes of incoming buffer' size.
- Support dynamic buffer re-allocation when the static-sized buffer mode is disabled
  - To handle the case when the size of an incoming buffer is larger than the existing dynamically allocated one, mqttsink frees the existing buffer and re-allocates a new buffer.

**Self evaluation:**

### DIFF
--- a/gst/mqtt/mqttcommon.h
+++ b/gst/mqtt/mqttcommon.h
@@ -30,6 +30,8 @@
  */
 #define GST_MQTT_MAX_NUM_MEMS   16
 
+#define GST_US_TO_NS_MULTIPLIER 1000
+
 /**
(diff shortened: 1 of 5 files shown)


again4you (approved):
LGTM!

helloahn (approved):
It would be nice to have an appropriate error message in the future.

PR #3229, opened 2021-04-29
[Test/NNFW] init info in testcase

jaeyun-jung:
Initialize tensor-info struct before calling filter callback.


myungjoo (approved):
LGTM

gichan-jang (approved):
LGTM

PR #3236, opened 2021-05-01
with GST_PLUGIN_PATH set `gst-inspect-1.0 nnstreamer` works, also renamed Dockerfile to use standard naming convention

shortcipher3:
Change Summary
* renamed Dockerfile to standard naming convention
* added environment variable for nnstreamer plugins to be discoverable by default

**How to evaluate:**
```
cd tools/docker/ubuntu18.04-devel/
docker build -t nnstreamer .
docker run nnstreamer gst-inspect-1.0 nnstreamer
```
Without this patch nnstreamer won't be found in the docker image.


myungjoo (approved):
LGTM. Thanks a lot!

dongju-chae (approved):
LGTM

PR #3239, opened 2021-05-03
[Filter] support flex tensor

jaeyun-jung:
Add pad template to support flex tensor in tensor-filter.

If incoming tensor is flexible, tensor-filter cannot get the meta from caps.
In this case, tensor-filter gets model info from subplugin and compares buffer size in transform().
That means, if the model handles non-static info (e.g., set-dimension with input info), it cannot support flex tensor.


myungjoo (approved):
LGTM.

Please describe behaviors of tensor-filter regarding flex-tensors.
If subplugin implementation is affected by flex-tensors, please describe such, too.

Issue #3242, opened 2021-05-04
[Android] pad_caps of tensor_converter does not have its subplugins


anyj0527:
Currently pad_caps of tensor_converter is set by checking nnsconf file.
Cause there is no such file in Android, should handle the case.

anyj0527:
Resolved by #3243

PR #3243, opened 2021-05-04
[Conf/Subplugin] get registered subplugins

jaeyun-jung:
Util function to get registrable subplugins list.

If conf file is not loaded (or does not exists), nnstreamer cannot get the subplugin list from configuration.
In this case (e.g., Android), conveter cannot set the pad template from subplugins, and will make negotiation failure.

To prevent this error, get subplugins from table and conf file.


myungjoo (approved):
LGTM

anyj0527 (approved):
LGTM 👍 I tested in Android device. This PR resolves #3242

PR #3246, opened 2021-05-06
[Filter] error in query caps

jaeyun-jung:
In transform-caps function, element should not query caps to peer.
The element should add all possible caps here.


gichan-jang (approved):
LGTM

again4you (approved):
LGTM!

PR #3249, opened 2021-05-07
[Filter/tvm] set input without memory copy

kimjh12:
- Fix `set_input` to `set_input_zero_copy`
- Requires input data to be aligned by 128 (https://github.com/apache/tvm/pull/3416#discussion_r301917977)

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
@@ -15,6 +15,7 @@
 #include <glib.h>
 #include <nnstreamer_cppplugin_api_filter.hh>
 #include <nnstreamer_log.h>
+#include <nnstreamer_plugin_api.h>
 #include <tensor_common.h>
 
 #include <dlpack/dlpack.h>
@@ -51,6 +52,7 @@ class tvm_subplugin final : public tensor_filter_subplugin
   tvm::runtime::Module gmod;
   std::vector<DLTensor> input_tensor_list;
   std::vector<DLTensor> output_tensor_list;
+  bool zero_copy_enabled;
 
   static const char *name;
   static const accl_hw hw_list[];
@@ -83,7 +85,8 @@ const accl_hw tvm_subplugin::hw_list[] = { ACCL_CPU, ACCL_GPU };
  */
 tvm_subplugin::tvm_subplugin ()
     : tensor_filter_subplugin (), empty_model (true), model_path (nullptr),
-      device (DLDevice{ kDLCPU, 0 }), mod_factory (nullptr), gmod (nullptr)
+      device (DLDevice{ kDLCPU, 0 }), mod_factory (nullptr), gmod (nullptr),
+      zero_copy_enabled (true)
 {
   gst_tensors_info_init (std::addressof (inputInfo));
   gst_tensors_info_init (std::addressof (outputInfo));
@@ -199,8 +202,8 @@ tvm_subplugin::configure_instance (const GstTensorFilterProperties *prop)
   mod_factory = tvm::runtime::Module::LoadFromFile (model_path, "so");
   gmod = mod_factory.GetFunction ("default") (device);
 
-  inputInfo.num_tensors = (int)gmod.GetFunction ("get_num_inputs") ();
-  outputInfo.num_tensors = (int)gmod.GetFunction ("get_num_outputs") ();
+  inputInfo.num_tensors = (int) gmod.GetFunction ("get_num_inputs") ();
+  outputInfo.num_tensors = (int) gmod.GetFunction ("get_num_outputs") ();
 
   tvm::runtime::NDArray arr;
   const DLTensor *dt;
@@ -252,56 +255,56 @@ bool
 tvm_subplugin::convert_dtype (tensor_type &nns_type, const DLDataType &dtype)
 {
   switch (dtype.code) {
-  case kDLInt:
-    switch (dtype.bits) {
-    case 64:
-      nns_type = _NNS_INT64;
+    case kDLInt:
+      switch (dtype.bits) {
+        case 64:
+          nns_type = _NNS_INT64;
+          break;
+        case 32:
+          nns_type = _NNS_INT32;
+          break;
+        case 16:
+          nns_type = _NNS_INT16;
+          break;
+        case 8:
+          nns_type = _NNS_INT8;
+          break;
+        default:
+          return false;
+      }
       break;
-    case 32:
-      nns_type = _NNS_INT32;
+    case kDLUInt:
+      switch (dtype.bits) {
+        case 64:
+          nns_type = _NNS_UINT64;
+          break;
+        case 32:
+          nns_type = _NNS_UINT32;
+          break;
+        case 16:
+          nns_type = _NNS_UINT16;
+          break;
+        case 8:
+          nns_type = _NNS_UINT8;
+          break;
+        default:
+          return false;
+      }
       break;
-    case 16:
-      nns_type = _NNS_INT16;
-      break;
-    case 8:
-      nns_type = _NNS_INT8;
+    case kDLFloat:
+      switch (dtype.bits) {
+        case 64:
+          nns_type = _NNS_FLOAT64;
+          break;
+        case 32:
+          nns_type = _NNS_FLOAT32;
+          break;
+        default:
+          return false;
+      }
       break;
     default:
       return false;
-    }
-    break;
-  case kDLUInt:
-    switch (dtype.bits) {
-    case 64:
-      nns_type = _NNS_UINT64;
-      break;
-    case 32:
-      nns_type = _NNS_UINT32;
-      break;
-    case 16:
-      nns_type = _NNS_UINT16;
-      break;
-    case 8:
-      nns_type = _NNS_UINT8;
-      break;
-    default:
-      return false;
-    }
-    break;
-  case kDLFloat:
-    switch (dtype.bits) {
-    case 64:
-      nns_type = _NNS_FLOAT64;
-      break;
-    case 32:
-      nns_type = _NNS_FLOAT32;
-      break;
-    default:
-      return false;
-    }
-    break;
-  default:
-    return false;
   }
   return true;
 }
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -322,7 +322,8 @@ features = {
     'extra_deps': [ pahomqttc_dep ]
   },
   'tvm-support': {
-    'extra_deps': [ tvm_dep, dlpack_dep ]
+    'extra_deps': [ tvm_dep, dlpack_dep ],
+    'project_args': {'ALLOC_ALIGNMENT': 1}
```
myungjoo, line 326: Because tvm subplugin may be built independently (any subplugin may be built independently), let's not change core behaviors with a subplugin availability.

Instead, you may add "enable-alloc-alignment" in meson_option, make it default "TRUE".
And generate a warning or an error at build of tvm-support if alloc-alignment is FALSE.

myungjoo: You may need to look at api.git for allocations as well. (Single API)

kimjh12: Sounds right! Updated

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
```diff
@@ -219,7 +219,7 @@ tvm_subplugin::configure_instance (const GstTensorFilterProperties *prop)
 
   for (unsigned int i = 0; i < inputInfo.num_tensors; ++i) {
     arr = getInput (i);
-    dt = arr.operator-> ();
+    dt = const_cast<DLTensor*> (arr.operator-> ());
```
zhoonit, line 222: I wonder the reasoning behind this change.

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
```diff
@@ -320,10 +320,13 @@ tvm_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
   tvm::runtime::NDArray tensor;
 
   /* read functions */
-  tvm::runtime::PackedFunc set_input = gmod.GetFunction ("set_input");
+  tvm::runtime::PackedFunc set_input = gmod.GetFunction ("set_input_zero_copy");
   if (set_input == nullptr) {
-    cleanup ();
-    throw std::runtime_error ("Packed function `set_input` not defined in model");
+    set_input = gmod.GetFunction ("set_input");
```
zhoonit, line 325: Is it okay to _implicitly_ alternate `set_input_zero_copy` to `set_input` in terms of memory management?

gst/nnstreamer/include/nnstreamer_plugin_api.h
```diff
@@ -379,6 +379,16 @@ gst_tensors_get_caps (GstPad * pad, const GstTensorsConfig * config);
 extern gboolean
 gst_pad_peer_has_tensor_caps (GstPad * pad);
 
+
+/**
+ * @brief Allocating memory block with alignment
```
zhoonit, line 384: How about adding note that this function is dependent if certain option is not enabled, it does not align?

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
kimjh12: Apparently there's no need for this..

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
kimjh12: Not sure.. I don't think of any other option other than this or throwing an error. But in normal case `set_input_zero_copy` is supported.

zhoonit: IMO, it's better to throw here if we are not sure, or issue a warning :)

tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
```diff
@@ -393,8 +393,10 @@ TEST (nnstreamerFilterTvm, invoke00)
   };
 
   output.size = input.size = sizeof (float) * 3 * 640 * 480 * 1;
-  input.data = g_malloc (input.size);
-  output.data = g_malloc (output.size);
+
+  input.data = aligned_alloc((size_t)128, input.size);
```
helloahn, line 397: Q) Is the number 128 not related to `GST_TENSOR_ALLOC_ALIGNMENT`?

kimjh12: Yes it is. `GST_TENSOR_ALLOC_ALIGNMENT` is given as bitmask so this also can be written as `GST_TENSOR_ALLOC_ALIGNMENT`+1

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
kimjh12: I agree

ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
```diff
@@ -320,10 +320,13 @@ tvm_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
...
+    if (set_input == nullptr) {
+      cleanup ();
+      throw std::runtime_error ("Packed function `set_input` not defined in model");
```
myungjoo, line 328: What if the pipeline is:
``` SOMESRC ! tensor_filter framework=tvm model=a ! ... ```
and ```SOMESRC``` is a third party element that we do not know if it is aligned or not?

Comment (myungjoo) [reply ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc:329]: Your code needs to be ready for such situation and you need test cases for such cases as well.

Comment (kimjh12) [reply ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc:329]: Is it ok to handle it by try & catch?
If not aligned then zero-copy throws an error so that I can catch it and retry with copying input

File: gst/nnstreamer/tensor_common.c [hunk 7]
```diff
@@ -1219,6 +1219,24 @@ gst_tensor_get_type_string (tensor_type type)
...
+GstMemory *
+gst_allocator_aligned_alloc (GstAllocator * allocator, gsize size)
+{
+#ifdef ALLOC_ALIGNMENT
+  GstAllocationParams alloc_params;
+  gst_allocation_params_init (&alloc_params);
+  alloc_params.align = (gsize) GST_TENSOR_ALLOC_ALIGNMENT;
+  return gst_allocator_alloc (allocator, size, &alloc_params);
+#endif
+  return gst_allocator_alloc (allocator, size, NULL);
```
wooksong, line 1237: With ```-DALLOC_ALIGNMENT```, this line will be never reached.

meson.build
```diff
@@ -186,6 +186,12 @@ if not get_option('tvm-support').disabled()
   if not dlpack_dep.found()
     dlpack_dep = cxx.find_library('dlpack', required: get_option('tvm-support'))
   endif
+  if not get_option('enable-alloc-alignment')
```
wooksong, line 189: Does ```get_option()``` return ```boolean```?  IMO,```.enabled()``` or other functions are missing here.

wooksong (changes requested):
Please see the comments.

ext/nnstreamer/extra/nnstreamer_protobuf.cc
```diff
@@ -89,7 +89,7 @@ gst_tensor_decoder_protobuf (const GstTensorsConfig *config,
   outbuf_size = gst_buffer_get_size (outbuf);
 
   if (outbuf_size == 0) {
-    out_mem = gst_allocator_alloc (NULL, size, NULL);
+    out_mem = gst_allocator_aligned_alloc (NULL, size);
```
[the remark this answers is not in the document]
kimjh12: Right. I'll figure out using `gst_allocator_set_default`. Sounds like quite a challenge tho... Thx for the help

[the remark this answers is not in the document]
dongju-chae: Yes.. you may need to implement new element. But it should be very simple by just wrapping the default allocator.

Of course, you may resolve this concern later if `tvm` is a critical item,

kimjh12:
After a brief discussion, I ended up to provide an API that indicates to set alignment for memory allocation.
This can be specified at initialization time as below (must be set to 127 for use TVM subplugin w/ zero copy)
```
gst_init ();
gst_tensor_alloc_init (127);
```

tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
```diff
@@ -485,6 +488,7 @@ main (int argc, char **argv)
   }
 
   gst_init (&argc, &argv);
+  gst_tensor_alloc_init (127UL);
```
dongju-chae, line 491: How about in tvm subplugin's init function?

wooksong:
> Please see the comments.

@kimjh12 Could you leave some reply comments when you marked the comments resolved?

kimjh12:
> > Please see the comments.
> 
> @kimjh12 Could you leave some reply comments when you marked the comments resolved?

Sorry. I acked all the comments but then I cleared whole previous code so that turned to be outdated..

wooksong:
> > > Please see the comments.
> > 
> > 
> > @kimjh12 Could you leave some reply comments when you marked the comments resolved?
> 
> Sorry. I acked all the comments but then I cleared whole previous code so that turned to be outdated..

You can still leave replies to the outdated comments. Anyway got it.

gst/nnstreamer/tensor_allocator.c
```diff
@@ -0,0 +1,119 @@
...
+ */
+static void
+gst_tensor_allocator_init (GstTensorAllocator * allocator)
+{
+  GstAllocator *sysmem_alloc, *alloc;
+
+  alloc = GST_ALLOCATOR_CAST (allocator);
+  sysmem_alloc = gst_allocator_find (GST_ALLOCATOR_SYSMEM);
+
+  alloc->mem_type = sysmem_alloc->mem_type;
```
wooksong, line 87: What happens if simply do ```alloc = system_alloc```?

kimjh12: I guess `*alloc = *system_alloc` might work. Thx for the comment!

tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
kimjh12: That sounds right. Updated!!

gst/nnstreamer/tensor_allocator.c
kimjh12: Doesn't work by just copying it maybe because of some private values, FYI

dongju-chae (approved):
LGTM

wooksong (approved):
LGTM

gst/nnstreamer/tensor_allocator.c
wooksong: @kimjh12 In https://github.com/GStreamer/gstreamer/blob/master/gst/gstallocator.c, it seems that copying of allocator works. Anyway, at this point, let's merge it. But we need to more investigate it.

kimjh12: Thanks for the detail. I' ll have a closer look

PR #3257, opened 2021-05-10
[Transform] change val name in macro

jaeyun-jung:
Fix svace issue, change local param in macro.


myungjoo (approved):
LGTM

PR #3259, opened 2021-05-10
[TensorMerge] change the default mode and description

helloahn:
* The description of `tensor_merge` is too short to make a pipeline with it.
* Currently, if the property `mode` is empty, `tensor_merge` generates an error because the default of the property is ERR. IMO, since only the mode `linear` is available, it should be a default mode.

Result: 
```
Element Properties:
  name                : The name of the object
                        flags: readable, writable
                        String. Default: "tensormerge0"
  parent              : The parent of the object
                        flags: readable, writable
                        Object of type "GstObject"
  mode                : Tensor Merge mode. Currently, `linear` is available only.
                        flags: readable, writable
                        String. Default: "linear"
  option              : Option for the tensor Merge mode.
			  1) linear mode: it will become the number of rank.
			     (e.g. merge 3:640:480 & 3:640:480 to 3:1280:480, option=1)
			  2) TBU
                        flags: readable, writable
                        String. Default: null
  sync-mode           : Time synchronization mode
			See also: https://github.com/nnstreamer/nnstreamer/blob/main/Documentation/synchronization-policies-at-mux-merge.md
                        flags: readable, writable
                        String. Default: "nosync"
  sync-option         : Option for the time synchronization mode
                        flags: readable, writable
                        String. Default: null
  silent              : Produce verbose output
                        flags: readable, writable
                        Boolean. Default: true
```

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_merge/gsttensormerge.c
+++ b/gst/nnstreamer/tensor_merge/gsttensormerge.c
@@ -149,23 +149,29 @@ gst_tensor_merge_class_init (GstTensorMergeClass * klass)
   gobject_class->set_property = gst_tensor_merge_set_property;
 
   g_object_class_install_property (gobject_class, PROP_SILENT,
-      g_param_spec_boolean ("silent", "Silent", "Produce verbose output ?",
+      g_param_spec_boolean ("silent", "Silent", "Produce verbose output",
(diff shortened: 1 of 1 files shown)


gichan-jang (approved):
LGTM

PR #3265, opened 2021-05-12
[MQTTsrc] fix minors and add comments

helloahn:
* set the DEFAULT_MQTT_KEEP_ALIVE_INTERVAL to 1 minute.
* Since the `cause` of cb_connection_lost is always null,
we cannot let users know the reason of the lost like client-id duplication.
IMO, we need to guide user to check the log messages of the broker.
* remove the callback `cb_mqtt_on_delivery_complete()`. It is useless for the subscriber.

**Self evaluation:**


myungjoo (approved):
LGTM

dongju-chae (approved):
LGTM

PR #3269, opened 2021-05-13
[Fix/Filter] Fix tensors info initialization

kimjh12:
[Fix/Filter] Fix tensors info initialization

- Internal tensors-info is not correctly initialized for some sub filters, causing incompatible format error.


zhoonit (approved):
Nice Catch 👍

anyj0527 (commented):
Thanks! How about call `gst_tensors_info_free`  in their cleanup func ?

kimjh12:
> Thanks! How about call `gst_tensors_info_free` in their cleanup func ?

Alright. thx

anyj0527 (approved):
LGTM ✔

myungjoo (approved):

PR #3273, opened 2021-05-13
[Meta] define tensor meta

jaeyun-jung:
Define new struct for tensor meta.
The flex-tensor buffer will have a fixed-size header per mem block.

### DIFF
--- a/gst/nnstreamer/include/tensor_typedef.h
+++ b/gst/nnstreamer/include/tensor_typedef.h
@@ -36,6 +36,13 @@
 #define NNS_TENSOR_SIZE_LIMIT_STR	"16"
 #define NNS_TENSOR_DIM_NULL ({0, 0, 0, 0})
 
+/**
+ * @brief The maximum rank in meta info (see GstTensorMetaInfo).
+ * This RANK is applied to meta info of other/tensors-flexible only and
+ * does not affect other/tensors(s)'s NNS_TENSOR_RANK_LIMIT.
+ */
+#define NNS_TENSOR_META_RANK_LIMIT	(16)
+
 #define NNS_MIMETYPE_TENSOR "other/tensor"
 #define NNS_MIMETYPE_TENSORS "other/tensors"
 #define NNS_MIMETYPE_TENSORS_FLEXIBLE "other/tensors-flexible"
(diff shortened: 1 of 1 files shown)


myungjoo (commented):
tensor_typedef.h serves as APIs for nnstreamer developers (including external packages).
The internal data structure shouldn't be here.

jaeyun-jung:
IMO, not internal, if developer implements extern plugin and link to tensor-crop, this is necessary. (tensor-crop info pad accepts a buffer, array of crop-info)

jaeyun-jung:
Or make new header for later work, now we are considering GstTensorMeta which includes num-tensors, type, dim and another info like media-type.

jaeyun-jung:
PR updated (2021-05-14)
flex-tensor will have header for each mem block.
TODO
- util functions to parse and update header / get header size / ...
- macro to check the version info
- update element : add header for each output mem block

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -232,4 +232,20 @@ typedef struct
...
+ * Internally NNStreamer handles a buffer with capability other/tensors-flexible using this information.
+ * - version: The version of tensor meta.
+ * - type: The type of each element in the tensor. This should be a value of enumeration tensor_type.
+ * - dimension: The dimension of tensor. This also denotes the rank of tensor. (e.g., [3:224:224:0] means rank 3.)
+ */
+typedef struct
+{
+  uint32_t version;
+  uint32_t type;
+  uint32_t dimension[16];
```
dongju-chae, line 248: Q. `16` means `NNS_TENSOR_SIZE_LIMIT`? Then, you may use this variable.
BTW, why not `NNS_TENSOR_RANK_LIMIT`?

jaeyun-jung: No, max dimension for meta info and flex tensor. In offline discussion, supposed that 16 is enough size to describe a tensor. (e.g. flex-tensor)

myungjoo: Let's define a macro so that its meaning is exposed.

jaeyun-jung:
PR updated (2021-05-17)
- Define max rank in meta.
- Add data format and media type in meta struct.

dongju-chae (approved):
LGTM

jaeyun-jung:
Added PR #3291, to handle meta and header.

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -36,6 +36,11 @@
 #define NNS_TENSOR_SIZE_LIMIT_STR	"16"
 #define NNS_TENSOR_DIM_NULL ({0, 0, 0, 0})
 
+/**
+ * @brief The maximum rank in meta info (see GstTensorMetaInfo).
```
myungjoo, line 40: Add more description (to avoid confusion with RANK for other/tensor(s) GST CAP):
```
* This RANK is applied to meta info of other/tensors-flexible only and does not affect other/tensors(s)'s NNS_TENSOR_RANK_LIMIT. 
```

jaeyun-jung: updated

myungjoo (approved):

PR #3279, opened 2021-05-14
[Gst/MQTT/Src] Keep subscribing to the topic only in the PLAYING state

wooksong:
This patch makes the subscription be removed when the state of the element is changed to PAUSED from PLAYING. In addition, to block the extra invocation of the receiving message handler running on the other thread, checking the subscription flag is added to this handler.

**Changelog**
- Handled the case when the state is changed to PLAYING again from PAUSED (https://github.com/nnstreamer/nnstreamer/pull/3279#pullrequestreview-661585524)

**Self evaluation:**

### DIFF
--- a/gst/mqtt/mqttsrc.c
+++ b/gst/mqtt/mqttsrc.c
@@ -130,6 +130,10 @@ static void cb_mqtt_on_subscribe (void *context,
     MQTTAsync_successData * response);
 static void cb_mqtt_on_subscribe_failure (void *context,
     MQTTAsync_failureData * response);
+static void cb_mqtt_on_unsubscribe (void *context,
+    MQTTAsync_successData * response);
+static void cb_mqtt_on_unsubscribe_failure (void *context,
+    MQTTAsync_failureData * response);
 
 static void cb_memory_wrapped_destroy (void *p);
 
@@ -137,6 +141,8 @@ static GstMQTTMessageHdr *_extract_mqtt_msg_hdr_from (GstMemory * mem,
     GstMemory ** hdr_mem, GstMapInfo * hdr_map_info);
 static void _put_timestamp_on_gst_buf (GstMqttSrc * self,
     GstMQTTMessageHdr * hdr, GstBuffer * buf);
+static gboolean _subscribe (GstMqttSrc * self);
+static gboolean _unsubscribe (GstMqttSrc * self);
 
 /**
  * @brief A utility function to check whether the timestamp marked by _put_timestamp_on_gst_buf () is valid or not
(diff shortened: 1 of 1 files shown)


gst/mqtt/mqttsrc.c
```diff
@@ -445,6 +445,20 @@ gst_mqtt_src_change_state (GstElement * element, GstStateChange transition)
...
+        int sub_ret;
+
+        sub_ret = MQTTAsync_unsubscribe (self->mqtt_client_handle,
+            self->mqtt_topic, &self->mqtt_respn_opts);
+        if (sub_ret != MQTTASYNC_SUCCESS) {
+          GST_ERROR_OBJECT (self,
+              "MQTTAsync_unsubscribe failed with return code %d", sub_ret);
+        }
+
+        self->is_subscribed = FALSE;
```
dongju-chae, line 459: Even if `MQTTAsync_unsubscribe` is failed?

wooksong: > Even if `MQTTAsync_unsubscribe` is failed?

At this point, there is nothing we can do.

dongju-chae: true..

dongju-chae (approved):
LGTM

helloahn (approved):
LGTM

myungjoo (commented):
Do you have re-subscribe mechanism for PAUSE-->PLAYING?

(Tested with start/plyaing --> pause --> resume/playing --> pause --> ... ?)
You'll need corresponding UTC (verifying it's actually "ubsub-ed" for pause, too) for this as well.

wooksong:
> Do you have re-subscribe mechanism for PAUSE-->PLAYING?
> 
> (Tested with start/plyaing --> pause --> resume/playing --> pause --> ... ?)
> You'll need corresponding UTC (verifying it's actually "ubsub-ed" for pause, too) for this as well.

As [this figure](https://gstreamer.freedesktop.org/documentation/application-development/advanced/images/clocks.png) represented, the state of the pipeline only affects the stream time. We are using the clock time so that timestamp conversion and synchronization would be OK, IMO. 
- Tested with sub/unsub and sub again by terminating while running the whole demonstration scenario. 
- Not (Tested with start/plyaing --> pause --> resume/playing --> pause --> ... ?)

gst/mqtt/mqttsrc.c
```diff
@@ -437,7 +439,14 @@ gst_mqtt_src_change_state (GstElement * element, GstStateChange transition)
       diff = GST_CLOCK_DIFF (base_time, cur_time);
       self->base_time_epoch =
           g_get_real_time () * GST_US_TO_NS_MULTIPLIER - diff;
-      GST_INFO_OBJECT (self, "GST_STATE_CHANGE_PAUSED_TO_PLAYING");
+
+      /** This handles the case when the state is changed to PLAYING again */
+      if (self->is_connected && !_do_subscribe (self)) {
+        GST_ERROR_OBJECT (self, "Failed to re-subscribe to %s",
+            self->mqtt_topic);
```
myungjoo, line 446: sub/unsub methods don't look symmetric and touch different properties. Can you improve it?

1. this uses is_connected while the other uses is_subscribed
2. this uses an abstracted func, "_do_subsribe", while the other uses MQTT API directly.
3. this do not update any properties directly, while the other do so.

Recommendation: use the same property to check status and make _do_unsubscribe that is symmetric to _do_subscribe.

Not using mutex in this (while the does does) is also a bit itchy.

myungjoo (approved):
LGTM

PR #3296, opened 2021-05-21
[Flex/Common] util functions to handle pad caps

jaeyun-jung:
add util fucntions to handle pad caps and move these to internal header.


gst/nnstreamer/tensor_common.c
```diff
@@ -1277,6 +1315,8 @@ gst_tensor_get_element_count (const tensor_dim dim)
 gsize
 gst_tensor_get_element_size (tensor_type type)
 {
+  g_return_val_if_fail (type >= 0 && type <= _NNS_END, 0);
```
dongju-chae, line 1318: `type >= _NNS_INT32`? I'm not sure which one is better..

dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM

PR #3304, opened 2021-05-25
[Converter/subplugins] Fix use after free

gichan-jang:
The internal representation of temporary of type std::string is freed by its destructor.
Copy the std::string to avoid use after free.

Self evaluation:

### DIFF
--- a/ext/nnstreamer/extra/nnstreamer_protobuf.cc
+++ b/ext/nnstreamer/extra/nnstreamer_protobuf.cc
@@ -149,7 +149,8 @@ gst_tensor_converter_protobuf (GstBuffer *in_buf, GstTensorsConfig *config, void
 
   for (guint i = 0; i < config->info.num_tensors; i++) {
     const nnstreamer::protobuf::Tensor *tensor = &tensors.tensor (i);
-    const gchar *name = tensor->name ().c_str ();
+    std::string _name = tensor->name ();
+    const gchar *name = _name.c_str ();
 
     config->info.info[i].name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
     config->info.info[i].type = (tensor_type)tensor->type ();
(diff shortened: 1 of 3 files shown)


jaeyun-jung (approved):

ext/nnstreamer/extra/nnstreamer_protobuf.cc
```diff
@@ -149,7 +149,8 @@ gst_tensor_converter_protobuf (GstBuffer *in_buf, GstTensorsConfig *config, void
 
   for (guint i = 0; i < config->info.num_tensors; i++) {
     const nnstreamer::protobuf::Tensor *tensor = &tensors.tensor (i);
-    const gchar *name = tensor->name ().c_str ();
+    std::string _name = tensor->name ();
+    const gchar *name = _name.c_str ();
 
     config->info.info[i].name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
```
wooksong, line 155:
```suggestion
    const gchar *name = g_strdup (tensor->name ().c_str ());

    config->info.info[i].name = name;
```

Is it wrong?

gichan-jang: In this case, even if `tensor->name ()` is "", `name` is not null.
In the test, check whether the name is null pointer (NULL).

wooksong: As we discussed via chat,  a string whose length is 0 is "\0", not NULL. We need to consider it here and in the related unit tests.

gichan-jang: In the process of converting to a stream buffers, NULL was temporarily replaced with "".
So it's converted back to NULL.

zhoonit (approved):
Nice catch!

PR #3305, opened 2021-05-26
Refactor/GRPC: remove duplicated get-property codes

myungjoo:
grpc-sink and grpc-src have duplicated codes of get-property.
Remove them.


jaeyun-jung (approved):

gichan-jang (approved):
LGTM

Issue #3114, opened 2021-03-08
Sharing Model Representation with CAPI


helloahn:
We can access a specific TFLite model through an `interpreter` object,
which can be built with the model path.

With TFLite API, when we want to use multi-threads to make processes faster,
because the `interpreter` is a sharable resource,
we can share the one `interpreter` without creating the new one for each thread.

However, in Single & Pipeline CAPI, there is no way to share the model representations currently.
I had a short discussion with @jaeyun-jung, and we made a few suggestions below.

1. Create a sharing table consists of Key(model path) & Value (`interpreter`) at each filter.
2. If there is any NEW request related to the added key(model path) the saved value(`interpreter`) will be used.
3. If there is no more usage of the shared Value(`interpreter`), it will be disappeared.

Though we still have some side-effects from the above solution, like the policy of dimension-change function, 
to reduce memory consumption, IMHO sharing the duplicated resource is quite important.

Please leave your opinions.

helloahn:
> I wonder if a single interpreter can support different inputs where a thread can set tflite inputs for itself without affecting the other threads?

No, it cannot support it, as I know. To using one interpreter with multi-threads we have to adjust the order one by one.

> Note that this is only required when different threads will try to use the shared interpreter concurrently. If only one interpreter is going to use it at once, then it's fine.

You're right. The main purpose of this issue is to save memory consumption by sharing one interpreter.
If efficiency is the most important point, we can create as much as we need for the threads.
However, as you know, because interpreters require memory,
if memory consumption is important, we need to consider reducing it by sharing.

> Also, you can consider the batch for the inputs, where different threads can set the data in their corresponding location in the batch, and all the threads data can run at once in a single batch.

IMO, It can be another solution for the issue.
We can solve it without additional code fix by giving guide users to change batch size rather than using multi-thread. Good point.
However, is it available for all models (changing batch size)?

kparichay:
> With the CPU-only environment, however, because the parallel calculation is not efficient, multi-thread could be better.

I wonder the difference in performance with the multi-threading done at pipeline level and multi-threading done with `interpreter.SetNumThreads()`.

myungjoo:
For API format, try to keep backward-compatibility. (I believe you can do it.)

helloahn:
> For API format, try to keep backward-compatibility. (I believe you can do it.)

Sure, it is one of the most important points.

kparichay:
[...]

PR #3307, opened 2021-05-26
[Filter/TFLite] change interpreter to pointer

helloahn:
To make the interpreter shareable, need to change to a pointer first.

related to #3114

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -122,7 +122,6 @@ class TFLiteInterpreter
 
   int invoke (const GstTensorMemory *input, GstTensorMemory *output);
   int loadModel (int num_threads, tflite_delegate_e delegate);
-  void moveInternals (TFLiteInterpreter &interp);
 
   int setInputTensorProp ();
   int setOutputTensorProp ();
@@ -204,6 +203,8 @@ class TFLiteInterpreter
 class TFLiteCore
 {
   public:
+  TFLiteCore ();
+  ~TFLiteCore ();
   int init (tflite_option_s *option);
   int loadModel ();
   gboolean compareModelPath (const char *model_path);
@@ -222,8 +223,8 @@ class TFLiteCore
   accl_hw accelerator;
   tflite_delegate_e delegate;
 
-  TFLiteInterpreter interpreter;
-  TFLiteInterpreter interpreter_sub;
+  TFLiteInterpreter *interpreter;
+  TFLiteInterpreter *interpreter_sub;
 
   void setAccelerator (const char *accelerators, tflite_delegate_e d);
 };
@@ -649,19 +650,6 @@ TFLiteInterpreter::setModelPath (const char *_model_path)
 }
 
 /**
- * @brief Move the ownership of interpreter internal members
- */
-void
-TFLiteInterpreter::moveInternals (TFLiteInterpreter &interp)
-{
-  interpreter = std::move (interp.interpreter);
-  model = std::move (interp.model);
-  inputTensorPtr = std::move (interp.inputTensorPtr);
-  outputTensorPtr = std::move (interp.outputTensorPtr);
-  setModelPath (interp.getModelPath ());
-}
-
-/**
  * @brief cache input and output tensor ptr before invoke
  * @return 0 on success. -errno on failure.
  */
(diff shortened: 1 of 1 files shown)


helloahn:
Since I force-pushed the branch, the previous PR was dead :sweat: 
There were few comments like below

@wooksong asked
>`TFLiteInterpreter` looks like a struct without any constructors (default one would be invoked at this line).
>
>Then, are there any reasons why we do not use `TfLiteInterpreterCreate`?
>
>https://github.com/tensorflow/tensorflow/blob/7ce597b5e8467cb662c9ae9a968e1ff10741f734/tensorflow/lite/c/c_api.cc#L131

I replied
> The class `TFLiteInterpreter` is a different one you shared. It's a respective class.
> PTAL https://github.com/nnstreamer/nnstreamer/blob/f1257d7c4d54699de88ce868f446a461084abb9e/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc#L117

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -894,62 +898,59 @@ int
...
-  interpreter.unlock ();
+  interpreter_sub->lock ();
+  interpreter_temp->lock ();
 
-out_unlock:
-  interpreter_sub.unlock ();
+  interpreter = interpreter_sub;
 
+  interpreter_sub->unlock ();
+  interpreter_temp->unlock ();
```
zhoonit, line 951: Maybe we don't have to lock the `interpreter_sub`? 

If we have to lock `interpreter_sub` , I think ,could you consider locking two locks at the same time? Locking two locks in sequence has a potential problem of deadlock.

or how about defining a move assignment operator and move constructor and use lock inside those?

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -894,62 +898,59 @@ int
...
+  delete interpreter_temp;
+out_unlock:
```
zhoonit, line 953: How about removing this label and return directly?

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -222,8 +223,8 @@ class TFLiteCore
   accl_hw accelerator;
   tflite_delegate_e delegate;
 
-  TFLiteInterpreter interpreter;
-  TFLiteInterpreter interpreter_sub;
+  TFLiteInterpreter *interpreter;
```
zhoonit, line 226: This might be a bit out of scope of this PR, If ownership is shared later on, deleting at destruction would be dangerous. How about using `std::shared_ptr` later on?

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
helloahn: it sounds good point. the locking of `interpreter_sub` has been removed!

zhoonit (approved):
LGTM :)

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
helloahn: you're right. it was just the heritage of the previous code. it's removed.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
helloahn: I understand what you worry about.
I will consider the point deeply for better sharing representation.

wooksong: @helloahn 
Q: Is it shared with other instances? or let them own it?

wooksong: Then, in addition to using ```std::sharred_ptr```, how about applying the singleton pattern to it?

helloahn: IMO, it could be a solution.
Since we still need to maintain multiple interpreters respectively,
I'm planning to manage them with a hash table.
The singleton pattern could be used at this point.

jaeyun-jung (approved):

PR #3311, opened 2021-05-28
[Gst/MQTT] Fix bugs and and further stabilize the Gst MQTT elements

wooksong:
This PR is related to bug fixes and further stabilization of the code. In detail,
- Reduced unnecessary mutex lock operations when changing MQTTSink's internal states. In addition, g_cond_wait is replaced with g_cond_wait_until to avoid possible deadlocks.
- Fixed bugs in the re-negotiation procedure in the MQTTSrc element
- Corrected the timeout in the MQTTSink element to wait for the disconnection to the broker using by MQTTSink
 
**Self evaluation:**

### DIFF
--- a/gst/mqtt/mqttsink.c
+++ b/gst/mqtt/mqttsink.c
@@ -551,18 +551,24 @@ gst_mqtt_sink_stop (GstBaseSink * basesink)
   disconn_opts.onFailure = cb_mqtt_on_disconnect_failure;
   disconn_opts.context = self;
 
-  g_mutex_lock (&self->mqtt_sink_mutex);
-  self->mqtt_sink_state = SINK_RENDER_STOPPED;
+  g_atomic_int_set (&self->mqtt_sink_state, SINK_RENDER_STOPPED);
   while (MQTTAsync_isConnected (self->mqtt_client_handle)) {
+    gint64 end_time = g_get_monotonic_time () +
+        DEFAULT_MQTT_DISCONNECT_TIMEOUT / 10;
+    mqtt_sink_state_t cur_state;
+
     MQTTAsync_disconnect (self->mqtt_client_handle, &disconn_opts);
-    g_cond_wait (&self->mqtt_sink_gcond, &self->mqtt_sink_mutex);
-    if ((self->mqtt_sink_state == MQTT_DISCONNECTED) ||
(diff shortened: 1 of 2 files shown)


gst/mqtt/mqttsink.c
```diff
@@ -816,10 +819,8 @@ gst_mqtt_sink_event (GstBaseSink * basesink, GstEvent * event)
 
   switch (type) {
     case GST_EVENT_EOS:
-      g_mutex_lock (&self->mqtt_sink_mutex);
-      self->mqtt_sink_state = SINK_RENDER_EOS;
+      g_atomic_int_set (&self->mqtt_sink_state, SINK_RENDER_EOS);
       g_cond_broadcast (&self->mqtt_sink_gcond);
```
myungjoo, line 823: Doesn't cond-broadcast require to have a mutex locked?

gst/mqtt/mqttsink.c
```diff
@@ -816,10 +819,8 @@ gst_mqtt_sink_event (GstBaseSink * basesink, GstEvent * event)
 
   switch (type) {
     case GST_EVENT_EOS:
-      g_mutex_lock (&self->mqtt_sink_mutex);
-      self->mqtt_sink_state = SINK_RENDER_EOS;
+      g_atomic_int_set (&self->mqtt_sink_state, SINK_RENDER_EOS);
```
myungjoo, line 822: Doesn't applying atomic function for mqtt_sink_state require to have all other mqtt_sink_state to use atomic function? (e.g., Line 555). And how about Line 562-564? (these three lines might be not atomic enough)

wooksong: @myungjoo You are absolutely right. All the state-changing statements should be replaced with ```g_atomic_int_set``` (or ``g_atomic_int_get``` in the case of reading). For 562-565, as you mentioned, only one atomical copying of the state to the local variable is enough, I guess. I will revise that mutex-atomic mess. Thanks.

gst/mqtt/mqttsink.c
wooksong: @myungjoo Right. I will modify it.

myungjoo (approved):
LGTM

jaeyun-jung (approved):

helloahn (approved):
LGTM

PR #3314, opened 2021-05-31
[Crop] crop region

jaeyun-jung:
1. Default mode to crop region of raw data.
2. Add new property lateness.
  If user sets lateness, then tensor-crop will compare the timestamp of incoming buffers and will drop buffer if time-diff is larger than lateness.

TODO later work
handle various crop-mode in tensor-crop.

### DIFF
--- a/gst/nnstreamer/tensor_crop/tensor_crop.c
+++ b/gst/nnstreamer/tensor_crop/tensor_crop.c
@@ -73,6 +73,7 @@ GST_DEBUG_CATEGORY_STATIC (gst_tensor_crop_debug);
 enum
 {
   PROP_0,
+  PROP_LATENESS,
   PROP_SILENT
 };
 
@@ -80,6 +81,11 @@ enum
  * @brief Flag to print minimized log.
  */
 #define DEFAULT_SILENT TRUE
+
+/**
+ * @brief Default value to compare timestamp of raw and info buffer, in milliseconds (-1 means no synchronization).
+ */
+#define DEFAULT_LATENESS (-1)
 
 /**
  * @brief Template for sink pad (raw data).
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_crop/tensor_crop.c
```diff
@@ -51,13 +73,13 @@ GST_DEBUG_CATEGORY_STATIC (gst_tensor_crop_debug);
 enum
 {
   PROP_0,
-  PROP_SILENT
+  PROP_LATENESS
```
myungjoo, line 76: Why silent is deleted?

jaeyun-jung: IMO, unnecessary. Actually this property is not used in tensor-crop.

myungjoo: For debugging, we will need to printout a few information, controlled by slient.
E.g., printing cropping info

jaeyun-jung: updated

myungjoo: Good! :)

myungjoo (approved):
LGTM

gst/nnstreamer/tensor_crop/tensor_crop.c
```diff
@@ -435,19 +483,274 @@ gst_tensor_crop_negotiate (GstTensorCrop * self)
...
+  GstBuffer *buf_raw, *buf_info, *result;
+  tensor_crop_info_s cinfo;
+  gboolean drop_raw, drop_info;
+
+  g_return_val_if_fail (data_raw && data_info, GST_FLOW_ERROR);
+
+  buf_raw = gst_collect_pads_peek (self->collect, data_raw);
+  buf_info = gst_collect_pads_peek (self->collect, data_info);
+  drop_raw = (buf_raw != NULL);
+  drop_info = (buf_info != NULL);
```
gichan-jang, line 677: Q. Is there a case that the buffers are NULL when the chain function is called?

gichan-jang (approved):
LGTM

PR #3317, opened 2021-06-02
Build: create libnnstreamer.so symlink in meson.

### LINKED ISSUE #3316: dependency('capi-nnstreamer') is not enough to link against libnnstreamer.so
# Issue Description
For nnstreamer dependency, using dependency('capi-nnstreamer') in meson results into linker error "error while loading shared libraries: libnnstreamer.so: cannot open shared object file"

Expected Result
============
should work as expected

How to Reproduce
===============
1. create a sample test app in meson and add dependency('capi-nnstreamer')  as dependency and try to invoke any ml_pipeline() results in runtime error.

Further Information
===============
- In nnstreamer.pc the library location mentioned  in /usr/local/lib/x86_64-linux-gnu . but the actual install directory when building nnstreamer from source is /usr/local/lib/x86_64-linux-gnu/gstreamer-1.0/ hence the error.

myungjoo:
Instead of creating symlinks in packaging (RPM/DEB),
create symlinks at meson so that local-build & user-install
works appropriately.

Fixes #3316
CC: @smohantty

### DIFF
--- a/gst/nnstreamer/meson.build
+++ b/gst/nnstreamer/meson.build
@@ -80,6 +80,18 @@ nnstreamer_shared = shared_library('nnstreamer',
   install_dir: plugins_install_dir
 )
 
+## @todo target name using method name() (since 0.54.0)
+# target_name = nnstreamer_shared.name()
+libpath_split = nnstreamer_shared.full_path().split('/')
+target_name = libpath_split[libpath_split.length() - 1]
+
+meson.add_install_script(
+  'sh', '-c',
+  'ln -fs @0@ @1@@2@'.format(
+    join_paths(plugins_install_dir, target_name),
+    '${DESTDIR}',
+    join_paths(nnstreamer_libdir, target_name)))
+
 nnstreamer_static = static_library('nnstreamer',
   nnstreamer_sources,
   dependencies: nnstreamer_base_deps,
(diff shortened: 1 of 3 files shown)


myungjoo:
Although this works with local-build & user-install, it breaks isolated builds such as gbs and pdebuild. Do not merge this until I fix it.

myungjoo:
> Although this works with local-build & user-install, it breaks isolated builds such as gbs and pdebuild. Do not merge this until I fix it. I need to let it be based on DESTDIR or BUILDROOT.

Looks like fixed.

gst/nnstreamer/meson.build
```diff
@@ -80,6 +80,13 @@ nnstreamer_shared = shared_library('nnstreamer',
   install_dir: plugins_install_dir
 )
 
+meson.add_install_script(
```
jaeyun-jung, line 83: how about using target name (other platforms)
```
# TODO target name using method name() (since 0.54.0)
# target_name = nnstreamer_shared.name()
libpath_split = nnstreamer_shared.full_path().split('/')
target_name = libpath_split[libpath_split.length() - 1]

meson.add_install_script(
  'sh', '-c',
  'ln -fs @0@ @1@@2@'.format(
    join_paths(plugins_install_dir, target_name),
    '${DESTDIR}',
    join_paths(nnstreamer_libdir, target_name)))
```

myungjoo: Yup. That's better.

myungjoo: And.. applied.

jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #3329, opened 2021-06-07
[tizen] flatbuffers python-support

gichan-jang:
flatbuffers python-support for tizen.
This will enable tensor converter and decoder unittest of the python.

Self evaluation:

### DIFF
--- a/gst/nnstreamer/tensor_converter/README.md
+++ b/gst/nnstreamer/tensor_converter/README.md
@@ -57,3 +57,44 @@ For each outgoing frame (on the source pad), there always is a **single** instan
 ```
 $ gst-launch videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! tensor_converter ! tensor_sink
 ```
+
+## Custom converter
+If you want to convert any media type to tensors, You can use custom mode of the tensor converter.
+### code mode
+This is an example of a callback type custom mode.
+```
+// Define custom callback function
+GstBuffer * tensor_converter_custom_cb (GstBuffer *in_buf, void *data, GstTensorsConfig *config) {
+  // Write a code to convert any media type to tensors.
+}
+...
+// Register custom callback function
+nnstreamer_converter_custom_register ("tconv", tensor_converter_custom_cb, NULL);
+...
+// Use the custom tensor converter in a pipeline.
+// E.g., Pipeline of " ... (any media stream) ! tensor_converter mode=custom-code:tconv ! (tensors)... "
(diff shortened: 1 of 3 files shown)


jaeyun-jung (approved):

packaging/nnstreamer.spec
```diff
@@ -165,6 +165,7 @@ BuildRequires:  pkgconfig(libpng)
 %if 0%{?flatbuf_support}
 # for flatbuffers
 BuildRequires: flatbuffers-devel
+BuildRequires: flatbuffers-python
```
dongju-chae, line 168: If it's mandatory for `nnstreamer-flatbuf`, you may need `Requires: flatbuffers-python` in line 378.
I don't think it can be automatically resolved because it's not a devel package.

again4you: What if we add `Requires: flatbuffers-python`, then the target platform should have a python environment. `flatbuffers` is used for a light-weight device such as IoT but python is not. Becuase of this reason, `Requires` might make a image break issue.

What about adding a use case to README.md for advanced platform developers?

dongju-chae: Then, how about `Recommends` or `Suggests`.

again4you: Well, it looks good.

gichan-jang: Thanks! `flatbuffers-python` is added as recommends in flatbuf package.
And the package is installed when `unit_test` is enabled.

dongju-chae (approved):
LGTM

gst/nnstreamer/tensor_converter/README.md
```diff
@@ -57,3 +57,44 @@ For each outgoing frame (on the source pad), there always is a **single** instan
...
+// Use the custom tensor converter in a pipeline.
+// E.g., Pipeline of " ... (any media stream) ! tensor_converter mode=custom-code:tconv ! (tensors)... "
+...
+// After everything is done.
+nnstreamer_converter_custom_unregister ("tconv");
+```
+
+### script mode
+* Note: Currently only Python is supported.
+  - If you want to use FlatBuffers Python in Tizen, install package `flatbuffers-python`. It also includes a Flaxbuffers Python.
```
jaeyun-jung, line 83: Flaxbuffers > FlexBuffers?

gichan-jang: Updated.!

PR #3334, opened 2021-06-10
[Transform] set write flag

jaeyun-jung:
In transform() function, set write flag when mapping the output buffer.


gichan-jang (approved):
LGTM!

myungjoo (approved):
LGTM.

How didn't this incur critical errors????

PR #3337, opened 2021-06-11
[Filter] add a hash table for shared key

helloahn:
to manage the shared model representation, a new hash table is required.

TFLite shared model representation reduced memory as below.
```
[share]
MAX RSS: 212172
MAX PSS: 202508

[not share]
MAX RSS: 353576
MAX PSS: 278325
```

+ reload is not supported yet.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -221,7 +221,7 @@ class TFLiteInterpreter
 class TFLiteCore
 {
   public:
-  TFLiteCore ();
+  TFLiteCore (const GstTensorFilterProperties *prop);
   ~TFLiteCore ();
   int init (tflite_option_s *option);
   int loadModel ();
@@ -244,6 +244,8 @@ class TFLiteCore
   TFLiteInterpreter *interpreter;
   TFLiteInterpreter *interpreter_sub;
 
+  gchar *shared_tensor_filter_key;
+  void checkSharedInterpreter (const GstTensorFilterProperties *prop);
   void setAccelerator (const char *accelerators, tflite_delegate_e d);
 };
 
@@ -712,13 +714,21 @@ fail_exit:
 /**
  * @brief	TFLiteCore constructor
  */
-TFLiteCore::TFLiteCore ()
+TFLiteCore::TFLiteCore (const GstTensorFilterProperties *prop)
 {
   num_threads = -1;
   accelerator = ACCL_NONE;
   delegate = TFLITE_DELEGATE_NONE;
   interpreter_sub = nullptr;
-  interpreter = new TFLiteInterpreter ();
+  shared_tensor_filter_key = NULL;
+
+  if (prop->shared_tensor_filter_key) {
+    shared_tensor_filter_key =
+        g_strdup (prop->shared_tensor_filter_key);
+    checkSharedInterpreter (prop);
+  }
+  else
+    interpreter = new TFLiteInterpreter ();
 }
 
 /**
@@ -726,7 +736,59 @@ TFLiteCore::TFLiteCore ()
  */
 TFLiteCore::~TFLiteCore ()
 {
-  delete interpreter;
+  if (shared_tensor_filter_key) {
+    if (nnstreamer_filter_shared_table_remove_this (shared_tensor_filter_key, this)) {
+      if (nnstreamer_filter_referred_list_length (shared_tensor_filter_key) == 0) {
+        /* if there is no more referred instance, remove the key from the shared table */
+        nnstreamer_filter_shared_table_remove (shared_tensor_filter_key);
+        /**
+         * @todo the below g_free () should be moved to the last of ~TFLiteCore () method
+         * To remove it, the has table lookup should be changed. 
+         */
+        g_free (shared_tensor_filter_key);
+        delete interpreter;
+      }
+    }
+  } else {
+    delete interpreter;
+    /** @todo should be removed */
+    g_free (shared_tensor_filter_key);
+  }
+}
+
+/**
+ * @brief	check the shared interpreter
+ * The shared model representation (interpreter) is allocated or shared.
+ * If `shared_tensor_filter_key` is already existed, it will share the TFLiteInterpreter.
+ * in the opposite case, the new TFLiteInterpreter will allocated and registered at the shared table.
+ */
+void
+TFLiteCore::checkSharedInterpreter (const GstTensorFilterProperties * prop)
+{
+  TFLiteInterpreter * ref = (TFLiteInterpreter *) nnstreamer_filter_get_shared_model_representation (shared_tensor_filter_key);
+  if (ref) {
+    if (g_strcmp0 (prop->model_files[0],
+            ref->getModelPath ()) != 0) {
+      ml_loge ("The model paths are different between the tensor filters!");
+      return;
+    }
+    if (!nnstreamer_filter_shared_table_insert (shared_tensor_filter_key, NULL, this)) {
+      ml_loge ("Failed to share the model representation!");
+      return;
+    }
+    interpreter = ref;
+  }
+  else {
+    interpreter = new TFLiteInterpreter ();
+    if (!nnstreamer_filter_shared_table_insert (shared_tensor_filter_key, interpreter, this)) {
+      ml_loge ("Failed to share the model representation!");
+      delete interpreter;
+      return;
+    }
+  }
+
+  ml_logd ("The model representation is shared: key=[%s], num of share: %u",
+      shared_tensor_filter_key, nnstreamer_filter_referred_list_length (shared_tensor_filter_key));
 }
 
 /**
@@ -922,6 +984,12 @@ TFLiteCore::reloadModel (const char *_model_path)
   int err;
   TFLiteInterpreter * interpreter_temp = interpreter;
 
+  if (shared_tensor_filter_key) {
+    /** @todo process should be added if the interpreter is shared */
+    ml_loge ("The reload is not supported with sharing model representation!");
+    return -EROFS;
+  }
+
   if (!g_file_test (_model_path, G_FILE_TEST_IS_REGULAR)) {
     ml_loge ("The path of model file(s), %s, to reload is invalid.", _model_path);
     return -EINVAL;
@@ -1102,7 +1170,7 @@ tflite_loadModelFile (const GstTensorFilterProperties *prop, void **private_data
     tflite_close (prop, private_data);
   }
 
-  core = new TFLiteCore ();
+  core = new TFLiteCore (prop);
   if (core == NULL) {
     g_printerr ("Failed to allocate memory for filter subplugin.");
     return -1;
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -242,7 +250,10 @@ class TFLiteCore
 
   TFLiteInterpreter *interpreter;
   TFLiteInterpreter *interpreter_sub;
-
+  
+  GHashTable *shared_table;
```
myungjoo, line 254: static?

helloahn: The hash table is created and managed at `tensor_filter_common.c`. This pointer is just existed to access the table. The reason why `filter_tf_lite` should access the table is to find the shared interpreter and remove it when the filter is destroyed.

myungjoo:
shared_table is a global variable across the whole libnnstreamer. It shouldn't be re-declared at each instance.
Make it an exported as a global variable of tensor-filter-common.
Make it available as a static variable of tensor-fitler c++ class, which is another "form" of global variable.
You may add internal API functions for this, too, if you feel the need.

CC @kimjh12 : you need to follow-up this.

kimjh12 (approved):
LGTM!!

helloahn:
> shared_table is a global variable across the whole libnnstreamer. It shouldn't be re-declared at each instance.
> Make it an exported as a global variable of tensor-filter-common.
> Make it available as a static variable of tensor-fitler c++ class, which is another "form" of global variable.
> You may add internal API functions for this, too, if you feel the need.
> 
> CC @kimjh12 : you need to follow-up this.

It has been changed to API rather than access the table directly at the external filters.
The static hash table is defined, managed, destroyed at `tensor_filter_common.c`.
The external filters can use the table through the APIs defined at `nnstreamer_plugin_api_filter.h`

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -711,21 +721,87 @@ TFLiteInterpreter::cacheInOutTensorPtr ()
...
+    if (ref) {
+      tflite_shared_s * sTemp = (tflite_shared_s *) ref;
+
+      sTemp->referred_list = g_slist_remove (sTemp->referred_list, this);
+      if (g_slist_length (sTemp->referred_list) == 0) {
+        delete interpreter;
+        g_slist_free (sTemp->referred_list);
+        g_free (sTemp);
+        nnstreamer_filter_shared_table_remove (shared_tensor_filter_key);
+        g_free (shared_tensor_filter_key);
```
myungjoo, line 759: 1. I'd suggest additional API of ```nnstreamer_filter_shared_table_remove_instance ( key, this )```, assuming that any subplugin who wants "shared key" will use the same "s-list of this" or "d-list of this"
2. You need mutex associated with each key (for list removal correctness)

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -548,6 +548,35 @@ extern accl_hw parse_accl_hw_fill (parse_accl_args accl_args);
...
+nnstreamer_filter_shared_table_insert (char *key, void *value);
+
+/* extern functions for shared model representation table */
+/**
+ * @brief remove the shared model representation at the table.
+ * @param[in] key The key to remove the matched value of hash table.
+ * @return TRUE if the new key and value are removed. FALSE if failed to remove it.
+ */
+extern int
+nnstreamer_filter_shared_table_remove (const char *key);
```
myungjoo, line 578: As I've mentioned above, remove (key, value) might be more useful than this. And realization w/ mutex becomes more robust with this semantics.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
myungjoo: 3. In general, try to move most of key/value managing part to the common code so that then another subplugin starts to use this, there will be almost no additional code required.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
myungjoo: If you feel ready, @kimjh12 you may jump in (need to talk w/ @helloahn for intentions and details of this PR and its expected following commits)

helloahn: The code has been updated to use APIs rather than to allow subplugins to access the referred list directly.
There is some update is required (mutex, hash key, reload model).
I'll share the detail with @kimjh12 before I leave and try to follow this.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
myungjoo: @kimjh12 Please follow these requirements up. We need to make it 1. thread-safe and 2. less coupling (this code of tf-lite subplugin should be able to "detach" itself from the table with a single API call.)

dongju-chae (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
myungjoo: With requirement 1., LINE 740 - 742 shouldn't be required (such tasks should be done by common code). The two APIs (referred_list_length and shared_table_remove) shouldn't be required. Any serialization issues should be handled inside common code and subplugins should not take care of such issues.

kimjh12: I just posted a commit following the rules as far as possible.. PTAL!

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -711,21 +713,73 @@ TFLiteInterpreter::cacheInOutTensorPtr ()
...
 
 /**
  * @brief	TFLiteCore destructor
  */
 TFLiteCore::~TFLiteCore ()
 {
-  delete interpreter;
+  if (shared_tensor_filter_key) {
+    if (nnstreamer_filter_shared_table_remove (shared_tensor_filter_key, this)) {
+      if (nnstreamer_filter_referred_list_is_empty (shared_tensor_filter_key)) {
```
myungjoo, line 740: Possible thread errors.
Thread 1, tensor_filter_tensorflow_lite.cc LINE 739 (TRUE)
Thread 1, tensor_filter_tensorflow_lite.cc LINE 740 (TRUE)
Thread 2, tensor_filter_tensorflow_lite.cc LINE 767 (FALSE)
Thread 1. tensor_filter_tensorflow_lite.cc LINE 741
Thread 2.  tensor_filter_tensorflow_lite.cc ==> SEGMENTATION FAULT.

myungjoo: nnstreamer_filter_referred_list_is_empty SHOULD BE NOT NEEDED.
nnstreamer_filter_shared_table_remove should be able to do "delete interpreter" atomically if needed. (defining a callback is fine.)
Otherwise, you'd need a mutex protecting an internal object from external logic (nnstreamer_filter_*), which doesn't look good (increasing coupling between the framework and its drivers).

myungjoo (changes requested):
1. Hazard of segmentation faults with multithreading
2. Unneeded coupling between tensor-common and tensor-filter-tensorflow-lite.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -2786,3 +2801,126 @@ gst_tensor_filter_check_hw_availability (const gchar * name, const accl_hw hw)
...
+/**
+ * @brief Insert the new shared model representation at the table.
+ * @param[in] key The new key to store at hash table.
+ * @param[in] value The new value to store at hash table.
+ * @param[in] instance The instance that is sharing the model representation. It will be registered at the referred list.
+ * @return 0 if the new key and value are inserted. Others are failed to insert it.
+ */
+int
+nnstreamer_filter_shared_table_insert (char *key, void *value, void *instance)
+{
```
myungjoo, line 2815: This function is not thread safe (against remove function). Cover critical section with mutex.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -2786,3 +2801,126 @@ gst_tensor_filter_check_hw_availability (const gchar * name, const accl_hw hw)
...
+  ml_logd ("The referred instance of sharing key: %s has been removed!", key);
+
+  /* remove key from table if list is empty */
+  if (g_list_length (value->referred_list) == 0) {
+    g_list_free (value->referred_list);
+    g_mutex_lock (&mutex);
+    ret = g_hash_table_remove (shared_model_table, key);
+    if (free_callback)
+      free_callback (instance);
+    g_mutex_unlock (&mutex);
```
myungjoo, line 2897: The protected resource, "shared_model_table", is not protected properly. Look at line 2878.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -712,21 +716,80 @@ TFLiteInterpreter::cacheInOutTensorPtr ()
...
+    interpreter = ref;
+  }
+  else {
+    interpreter = new TFLiteInterpreter ();
+    if (!nnstreamer_filter_shared_table_insert (shared_tensor_filter_key, interpreter, this)) {
+      ml_loge ("Failed to share the model representation!");
+      delete interpreter;
+      return;
+    }
+  }
```
myungjoo, line 790: You still have thread-synchronization issue and calling get and then insert is not a good design.
I'd suggest:

```C
///this function
{
   ...
  interpreter = get (key);
  if (!interpreter) { // not available. create interpreter
    interpreter = insert_and_get (key, new-interpreter());
  }
  ...
}

...

///remove function
{
  put (key); // if internal counter or linked-list hits 0, it'll be removed by tensor-filter infra.
}
```

kimjh12: That helps me understand how to synchronize. With that API we might need static mutex to lock around `get ()` and `insert_and_get ()`, right? Just in case remove function runs between the two

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -2786,3 +2801,127 @@ gst_tensor_filter_check_hw_availability (const gchar * name, const accl_hw hw)
...
+    ml_loge ("The interpreter should NOT be NULL!");
+    return NULL;
+  }
+
+  model_rep = (GstTensorFilterSharedModelRepresenatation *)
+      g_malloc0 (sizeof (GstTensorFilterSharedModelRepresenatation));
+  model_rep->shared_interpreter = interpreter;
+  model_rep->referred_list = g_list_append (model_rep->referred_list, instance);
+
+  g_mutex_lock (&mutex);
```
myungjoo, line 2873: What if ```nnstreamer_filter_shared_model_insert_and_get()``` is invoked by two threads?

myungjoo (approved):
Looks Good to me except for the minor potential glitches at LINE 2873~2878.
Please merge when the comment https://github.com/nnstreamer/nnstreamer/pull/3337/files#r694024157 is resolved.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
kimjh12: Oh `g_hash_table_insert` replaces even if the key exists.. Thank you

PR #3339, opened 2021-06-14
[filter/cpp] Catch exceptions in invoke

anyj0527:
- Catch other exceptions and print the error msg in invoke function.
- runtime_error or exception have wide coverage, Let those cpp exceptions return error code EINVAL.
- Developer should throw exceptions with detailed error message.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -91,12 +91,9 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   } catch (const std::system_error &e) {
     _RETURN_ERR_WITH_MSG (e.code ().value () * -1, e.what ());
   } catch (const std::runtime_error &e) {
-    /** @todo return different error codes according to exceptions */
-    _RETURN_ERR_WITH_MSG (-1, e.what ());
+    _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::exception &e) {
-    /** @todo Write exception handlers. */
-    /** @todo return different error codes according to exceptions */
-    _RETURN_ERR_WITH_MSG (-1, e.what ());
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
```diff
@@ -160,12 +160,19 @@ tensor_filter_subplugin::cpp_invoke (const GstTensorFilterFramework *tf,
...
+    _RETURN_ERR_WITH_MSG (e.code ().value () * -1, e.what ());
+  } catch (const std::runtime_error &e) {
+    /** @todo return different error codes according to exceptions */
+    _RETURN_ERR_WITH_MSG (-1, e.what ());
   } catch (const std::exception &e) {
     /** @todo Write exception handlers. */
-
-    return -EINVAL;
     /** @todo return different error codes according to exceptions */
+    _RETURN_ERR_WITH_MSG (-1, e.what ());
```
myungjoo, line 173: Instead of -1, let's use some standard error numbers even if you are going to do ```@todo``` later. (Line 169 and 173)

anyj0527: `std::runtime_error` and `std::exceptions` would cover very general exceptions. Could you suggest proper errorno for those?

myungjoo: Look at ```/usr/include/asm-generic/errno.h``` and ```/usr/include/asm-generic/errno-base.h```. -1 is -EPERM (Operation not permitted). Find the most general ones from the errno list. For example, accessing a subplugin may be seen as accessing a peripheral hardware accelerator, which is an I/O device.

anyj0527: Assuming developers would throw exceptions with detailed error message, IMHO, EINVAL could be an option.

myungjoo (approved):
LGTM

helloahn (approved):
LGTM

PR #3340, opened 2021-06-14
[Transform] support flex tensor

jaeyun-jung:
Add pad template and update chain function to support flexible tensor stream.

TODO: Now max rank supported in tensor-transform is 4.

### DIFF
--- a/gst/nnstreamer/tensor_transform/tensor_transform.c
+++ b/gst/nnstreamer/tensor_transform/tensor_transform.c
@@ -178,6 +178,10 @@ static gboolean gst_tensor_transform_transform_size (GstBaseTransform * trans,
     GstPadDirection direction, GstCaps * caps, gsize size,
     GstCaps * othercaps, gsize * othersize);
 
+static gboolean gst_tensor_transform_convert_dimension (GstTensorTransform *
+    filter, GstPadDirection direction, guint idx, const GstTensorInfo * in_info,
+    GstTensorInfo * out_info);
+
 #define GST_TYPE_TENSOR_TRANSFORM_MODE (gst_tensor_transform_mode_get_type ())
 /**
  * @brief A private function to register GEnumValue array for the 'mode' property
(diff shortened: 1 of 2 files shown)


tests/nnstreamer_plugins/unittest_plugins.cc
```diff
@@ -1909,6 +1909,100 @@ TEST (testTensorTransform, arithmeticChangeOptionString)
   gst_harness_teardown (h);
 }
 
+/**
+ * @brief Test for tensor_transform arithmetic (flex tensor)
```
gichan-jang, line 1913: I think static tensor -> flex tensor transform test is needed to increase the code coverage.

jaeyun-jung: later work : transform-caps in tensor-transform
When called transform-caps(), it should append all possible caps as result, but now only one possible caps supported.

gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_transform/tensor_transform.c
```diff
@@ -933,33 +933,151 @@ gst_tensor_transform_finalize (GObject * object)
...
+ * @brief Dimension conversion calculation
+ * @param[in] filter "this" pointer
+ * @param[in] direction GST_PAD_SINK if input->output conv
+ * @param[in] idx index of the input tensors
+ * @param[in] in_info tensor info structure of source tensor (input if direction is SINK)
+ * @param[out] out_info tensor info structure of destination tensor (output if direction is SINK)
+ * @return TRUE if success
+ */
+static gboolean
+gst_tensor_transform_convert_dimension (GstTensorTransform * filter,
```
myungjoo, line 946: Can you keep the position of this function and only add function declaration to reduce changes? (making reviews easier)

tests/nnstreamer_plugins/unittest_plugins.cc
```diff
@@ -1909,6 +1909,100 @@ TEST (testTensorTransform, arithmeticChangeOptionString)
...
+ */
+TEST (testTensorTransform, arithmeticFlexTensor)
```
myungjoo, line 1915: negative cases?

gst/nnstreamer/tensor_transform/tensor_transform.c
jaeyun-jung: updated

tests/nnstreamer_plugins/unittest_plugins.cc
jaeyun-jung: After changing caps string (about tensor format)
All elements with flex caps should have formal condition to check data size. Now tensor-transform does not compare data size in transform().

myungjoo: Ok. Let's postpone it until then.

myungjoo (approved):
LGTM

PR #3346, opened 2021-06-21
[Common] parse flex config from caps

jaeyun-jung:
When getting config from caps, if caps is flexible, do not get the type and dimension from caps.


helloahn (approved):
LGTM

myungjoo (approved):

PR #3347, opened 2021-06-21
[Crop] flex tensor on raw data pad

jaeyun-jung:
support flex tensor on raw data pad.


helloahn (approved):
LGTM

myungjoo (approved):
LGTM

Issue #3145, opened 2021-03-22
Stream Type: Sparse Tensors


myungjoo:
This is forked from #2936 (compressed tensor)

Requested by "client", transmitting sparse tensors for Edge-AI projects are required.

V1: sparse-tensor (compressed) stream type definition with converter & decoder standard subplugins
V2: sparse-tensor (compressed) stream direct inferences with a few tensor-filter subplugins
V1: TBD 2021
V2: TBD 2021 4Q or 2022.

CC: @jaeyun-jung @anyj0527

kparichay:
I wonder about sparse tensor src/sink if they will also be needed?

myungjoo:
Sparse Tensor is just another stream data type (e.g., other/sparsetensors).

For src, appsrc should be fine. (no need for yet another src element)

Tensorsink... let's see what happens without it. My guess is that we won't need it because users will generally use tensor-converter to get other/tensors. We can easily expand tensor-sink if that's really required. Actually, tensor-sink is not that useful when app-sink is available.

kparichay:
> We can easily expand tensor-sink if that's really required. Actually, tensor-sink is not that useful when app-sink is available.

Agreed.

myungjoo:
I guess this is already

PR #3350, opened 2021-06-21
[SparseTensor] Define SparseTensor and API @open sesame 07/06 09:45

anyj0527:
- Define a tensor_format for SparseTensor
- Define a info structure for SparseTensor which has "length" of non-zero elements
- Add a internal header `tensor_sparse_util.h` for util functions that will be used in related elements
- Add a union in GstTensorMetaInfo for SparseTensor

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [X]Skipped

SparseTensor

SparseTensors could be more efficient on processing of tensors that contain a lot of zero values.
I'll adopt the coordinate list (COO) format for encoding it, same as other NN frameworks.

The COO encoding for sparse tensors is comprised of:  (REF: https://www.tensorflow.org/guide/sparse_tensor)

1. values: A 1D tensor with shape [N] containing all nonzero values.
2. indices: A 2D tensor with shape [N, rank], containing the indices of the nonzero values.
3. dense_shape: A 1D tensor with shape [rank], specifying the shape of the tensor.

For example a Tensor
```
[[1, 0, 0, 0]
 [0, 0, 2, 0]
 [0, 0, 0, 0]]
```
can be encoded as
```
SparseTensor(values=[1, 2], indices=[[0, 0], [1, 2]], dense_shape=[3, 4])
```

In our project, a SparseTensor should be encoded with these information:
```cpp
tensor_type   data_type;
tensor_dim    dense_shape;
unsigned int  nnz; /* the number of non-zero elements */
data_type     values[nnz];
unsigned int  indices[nnz][RANK_LIMIT];
```

Since data_type and dense_shape are could be represented in `GstTensorMetaInfo`,
Add a `GstSparseTensorInfo`, and let it have the `nnz`.
The values and indices are stored in a single memory chunk, and the `length` information would be used for parsing the buffer.

### DIFF
--- a/gst/nnstreamer/include/tensor_typedef.h
+++ b/gst/nnstreamer/include/tensor_typedef.h
@@ -264,6 +264,14 @@ typedef struct
 } GstTensorsConfig;
 
 /**
+ * @brief Internal data structure for sparse tensor info
+ */
+typedef struct
+{
+  uint32_t nnz; /**< the number of "non-zero" elements */
+} GstSparseTensorInfo;
+
+/**
  * @brief Data structure to describe a tensor data.
  * This represents the basic information of a memory block for tensor stream.
  *
@@ -281,6 +289,14 @@ typedef struct
   uint32_t dimension[NNS_TENSOR_META_RANK_LIMIT];
   uint32_t format;
   uint32_t media_type;
+
+  /**
+   * @brief Union of the required information for processing each tensor "format".
+   */
+  union {
+    GstSparseTensorInfo sparse_info;
+  };
+
 } GstTensorMetaInfo;
 
 #endif /*__GST_TENSOR_TYPEDEF_H__*/
--- a/gst/nnstreamer/tensor_sparse/tensor_sparse_util.h
+++ b/gst/nnstreamer/tensor_sparse/tensor_sparse_util.h
@@ -1,0 +1,41 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * GStreamer / NNStreamer Sparse Tensor support
+ * Copyright (C) 2021 Yongjoo Ahn <yongjoo1.ahn@samsung.com>
+ */
+/**
+ * @file	tensor_sparse_util.h
+ * @date	06 Jul 2021
+ * @brief	Util functions for tensor_sparse encoder and decoder.
+ * @see		https://github.com/nnstreamer/nnstreamer
+ * @author	Yongjoo Ahn <yongjoo1.ahn@samsung.com>
+ * @bug		No known bugs except for NYI items
+ */
+
+#ifndef __GST_TENSOR_SPARSE_UTIL_H__
+#define __GST_TENSOR_SPARSE_UTIL_H__
+
+#include <gst/gst.h>
+#include <tensor_common.h>
+
+/**
+ * @brief NYI. Make dense tensor with input sparse tensor.
+ * @param[in,out] meta tensor meta structure to be updated
+ * @param[in] in pointer of input sparse tensor data
+ * @param[out] out pointer of output dense tensor data. Assume that it's already allocated.
+ * @return TRUE if no error
+ */
+extern gboolean
+gst_tensor_sparse_to_dense (GstTensorMetaInfo * meta, gpointer in, gpointer out);
+
+/**
+ * @brief NYI. Make sparse tensor with input dense tensor.
+ * @param[in,out] meta tensor meta structure to be updated
+ * @param[in] in pointer of input dense tensor data
+ * @param[out] out pointer of output sparse tensor data. Assume that it's already allocated.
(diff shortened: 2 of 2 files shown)


gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -239,6 +240,22 @@ typedef struct
   int rate_d; /**< framerate is in fraction, which is numerator/denominator */
 } GstTensorsConfig;
 
+/**
+ * @brief Internal data structure for sparse tensor info
+ */
+typedef struct
+{
+  uint32_t length; /**< the number of "non-zero" elements */
```
dongju-chae, line 248: Q.  It's ambiguous to me. Is `length` commonly used terminology in sparse tensor or matrix?

anyj0527: No, there are no common term I found. `size`, `length`, and `num_elements` were candidates. Maybe other terms be more suitable?

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -257,6 +274,13 @@ typedef struct
   uint32_t dimension[NNS_TENSOR_META_RANK_LIMIT];
   uint32_t format;
   uint32_t media_type;
+
```
helloahn, line 277: It would be better if the information about the union could be provided in the comments.

jaeyun-jung (approved):

gst/nnstreamer/include/tensor_typedef.h
kimjh12: Scipy uses `nnz`, FYI

gst/nnstreamer/include/tensor_typedef.h
anyj0527: I added brief comment on it: `@brief Union of the required information for processing each tensor "format".`
Thanks!

gst/nnstreamer/include/tensor_typedef.h
dongju-chae: Then, let's use `nnz`, `non_zeros`, or other intuitive word.

anyj0527: `nnz` seems cool. I chose it. PTAL

dongju-chae (approved):
LGTM

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -257,6 +274,16 @@ typedef struct
...
+  /**
+   * @brief Union of the required information for processing each tensor "format".
+   * @todo Implement "tensor-query"
+   */
+  union {
+    GstSparseTensorInfo sparse_info;
+    GstTensorQueryInfo query_info;
```
myungjoo, line 284: Query and Sparse-Tensor are orthogonal. Why they are in a single union?

anyj0527: `sparse_info` contains single integer value. Query related functions may be responsible for preserving the information.
Here Query is introduced for showing concept of format-internal data. Later we can decide whether to keep single union or to separate `query_info`.

gst/nnstreamer/include/nnstreamer_plugin_api.h
```diff
@@ -514,13 +514,33 @@ gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem);
...
 
+/**
+ * @brief NYI. Make dense tensor with input sparse tensor.
+ * @param[in,out] meta tensor meta structure to be updated
+ * @param[in] in pointer of input sparse tensor data
+ * @param[out] out pointer of output dense tensor data
+ * @return TRUE if no error
+ */
+extern gboolean
+gst_tensor_sparse_to_dense (GstTensorMetaInfo * meta, gpointer in, gpointer out);
```
jaeyun-jung, line 532: As we talked on offline, move these functions to internal.

anyj0527: Added a internal header tensor_sparse_util.h for util functions that will be used in related elements. PTAL

gst/nnstreamer/include/tensor_typedef.h
myungjoo: Because someone may "query" with "sparse tensors", they should be separated from the beginning.
You need to union mutually exclusive items of the same category; e.g., "tensor format" (default, flex, sparse).
Otherwise, they should exist independently.

myungjoo:
This will resolve #3145

gst/nnstreamer/include/tensor_typedef.h
anyj0527: @myungjoo Right, I excluded query related code in this commit. Those will be properly added by other PRs.

myungjoo (approved):

PR #3366, opened 2021-06-25
[TEST] Fix streambuffers test

gichan-jang:
Test fails on arm arch(gbs build) because buffer conversion takes long time.
The order of buffers is important for this test because the frame rate of the filesource is different.
The purpose of the test is to convert other/tensors to stream buffers, so modifies the test.
 *Consecutive use of the python decoder has some problem, I will fix it next PR.

Self evaluation:

### DIFF
--- a/tests/nnstreamer_converter_python3/runTest.sh
+++ b/tests/nnstreamer_converter_python3/runTest.sh
@@ -116,9 +116,10 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} audiotestsrc num-buffers=1 samplesp
 callCompareTest test.audio16k2c.u16le.origin.log test.audio16k2c.u16le.log 2-6 "Audio16k2c-u16le Golden Test" 0 0
 
 # Test other/tensors
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} tensor_mux name=tensors_mux sync-mode=basepad sync-option=1:50000000 ! tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter mode=custom-script:${PATH_TO_SCRIPT} ! multifilesink location=testsynch19_%1d.log \
-    tensor_mux name=tensor_mux0  sync-mode=slowest ! queue ! tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter ! tensors_mux.sink_0 \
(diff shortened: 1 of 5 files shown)


tests/nnstreamer_converter_python3/runTest.sh
```diff
@@ -116,9 +116,10 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} audiotestsrc num-buffers=1 samplesp
 callCompareTest test.audio16k2c.u16le.origin.log test.audio16k2c.u16le.log 2-6 "Audio16k2c-u16le Golden Test" 0 0
 
 # Test other/tensors
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} tensor_mux name=tensors_mux sync-mode=basepad sync-option=1:50000000 ! tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter mode=custom-script:${PATH_TO_SCRIPT} ! multifilesink location=testsynch19_%1d.log \
-    tensor_mux name=tensor_mux0  sync-mode=slowest ! queue ! tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter ! tensors_mux.sink_0 \
-    tensor_mux name=tensor_mux1  sync-mode=slowest ! queue ! tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter ! tensors_mux.sink_1 \
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} tensor_mux name=tensors_mux sync-mode=basepad sync-option=1:50000000 ! \
+    tensor_decoder mode=flexbuf ! other/flexbuf ! tensor_converter mode=custom-script:${PATH_TO_SCRIPT} ! multifilesink location=testsynch19_%1d.log \
+    tensor_mux name=tensor_mux0  sync-mode=slowest ! tensors_mux.sink_0 \
+    tensor_mux name=tensor_mux1  sync-mode=slowest ! tensors_mux.sink_1 \
```
myungjoo, line 122: Is it supposed to have equally without queue? (tensor_mux w/ basepad)

gichan-jang: I think it will work the same way.
This test adds conversion of the stream buffers to the existing tensor mux test. There is no queue in the original test.

jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #3367, opened 2021-06-25
meson: configure inidir properly when a relative path is given.

### LINKED ISSUE #3161: The path of nnstreamer.ini should be fixed
## Envorinment
- Meson 0.53.2
- Ubuntu 20.04 (aarch64)

## Description
The install path of `nnstreamer.ini` should be fixed to something(like `/etc/nnstreamer.ini`)
If we install it through `meson build` and `ninja install` without any options like `--sysconfdir`,
`nnstreamer.ini` is installed at `/usr/local/etc/nnstreamer.ini`, and it is not searchable and generates the below error.

```bash
(run the example nnstreamer_example_object_detection_tflite)

** (nnstreamer_example_object_detection_tflite:40709): WARNING **: 17:09:07.325: Failed to load the configuration, no config file found.
...
```

myungjoo:
If sysconfdir option is not given or a relative path is given,
add prefix so that its actual install path (which uses prefix
if a relative path is given) matches NNSTREAMER_CONF_FILE
macro in nnstreamer_conf.h

Fixes #3161

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -88,7 +88,8 @@ nnstreamer_prefix = get_option('prefix')
 nnstreamer_libdir = join_paths(nnstreamer_prefix, get_option('libdir'))
 nnstreamer_bindir = join_paths(nnstreamer_prefix, get_option('bindir'))
 nnstreamer_includedir = join_paths(nnstreamer_prefix, get_option('includedir'))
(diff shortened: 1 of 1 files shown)


jaeyun-jung (approved):

dongju-chae (approved):
LGTM

PR #3368, opened 2021-06-25
[Crop] remove tensor-config struct

jaeyun-jung:
Prepare refactoring about tensor config, remove tensor-config struct and replace to tensors-config.


jaeyun-jung:
@kimjh12 This is first step to update / modify tensor caps of all nnstreamer elements.
As we talked on offline, let's remove tensor-config struct first.

kimjh12 (approved):
LGTM!!!

myungjoo (approved):
LGTM

PR #3381, opened 2021-07-05
[Caps/Format] add tensor format in caps

jaeyun-jung:
Add tensor format in caps.
Fix pad template with flex tensor and remove old mimetype.
Now caps string for flex-tensor has framerate field.

### DIFF
--- a/gst/nnstreamer/tensor_common.c
+++ b/gst/nnstreamer/tensor_common.c
@@ -112,8 +112,15 @@ static GstCaps *
 _get_tensors_caps (const GstTensorsConfig * config)
 {
   GstCaps *caps;
+  tensor_format format = _NNS_TENSOR_FORMAT_STATIC;
 
   caps = gst_caps_from_string (GST_TENSORS_CAP_DEFAULT);
+
+  if (gst_tensors_info_is_flexible (&config->info))
+    format = _NNS_TENSOR_FORMAT_FLEXIBLE;
+
+  gst_caps_set_simple (caps, "format", G_TYPE_STRING,
+      gst_tensor_get_format_string (format), NULL);
 
   if (config->info.num_tensors > 0) {
     gchar *dim_str, *type_str;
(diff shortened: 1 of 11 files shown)


gst/nnstreamer/tensor_common.c
```diff
@@ -112,9 +112,16 @@ static GstCaps *
 _get_tensors_caps (const GstTensorsConfig * config)
 {
   GstCaps *caps;
+  tensor_format format;
 
   caps = gst_caps_from_string (GST_TENSORS_CAP_DEFAULT);
 
+  format = config->info.info[0].format;
```
myungjoo, line 119: Question: why "format" is in TensorInfo? not in TensorsInfo? Does anyone know why it should be in TensorInfo?

myungjoo: If it's better be in TensorsInfo, not TensorInfo, it should be dealt with another PR later.

jaeyun-jung: Considered multi tensors with mixed format (0:static / 1:flex / ...). I will check this after #3389 (removing tensor-config).

dongju-chae (approved):
LGTM

gst/nnstreamer/tensor_common.c
myungjoo: If you want to support mixed formats, ```other/tensors,format=${type}``` won't work.

jaeyun-jung: related issue #3405

gichan-jang (approved):
LGTM

anyj0527 (approved):
LGTM 👍👍

gst/nnstreamer/tensor_common.c
jaeyun-jung: Fix this on https://github.com/nnstreamer/nnstreamer/pull/3418

PR #3404, opened 2021-07-14
[Query] Update query_serversrc @open sesame 07/16 13:40

kimjh12:
Update serversrc element to serve remote tensor query
Currently working almost same as tcpserversrc
Create server socket and receive data from client on TCP

can test
```gst-launch-1.0 tensor_query_serversrc ! filesink location=temp.out```  
```echo hi | nc localhost 4953```  

---
**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_common.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_common.c
@@ -26,17 +26,17 @@
  * @param[in] hostname the hostname.
  * @param[in] port a port number.
  * @param[in] cancellable (nullable) GCancellable
+ * @param[out] saddr Socket address
  * @return Newly created socket or NULL on error.
  * @note Caller is responsible for unreferring the returned object with g_object_unref().
  */
 GSocket *
 gst_tensor_query_socket_new (const gchar * hostname, guint16 port,
-    GCancellable * cancellable)
+    GCancellable * cancellable, GSocketAddress ** saddr)
 {
   GSocket *socket;
   GError *err = NULL;
   GInetAddress *addr;
-  GSocketAddress *saddr;
 
   /* look up name if we need to */
   addr = g_inet_address_new_from_string (hostname);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: gst/nnstreamer/tensor_query/tensor_query_common.c [hunk 1]
```diff
@@ -131,11 +138,29 @@ gst_tensor_query_socket_receive (GSocket * socket, GCancellable * cancellable,
...
+    if (need_alloc) {
+      out_mem = gst_allocator_alloc (NULL, read, NULL);
+    } else {
+      if (gst_buffer_get_size (outbuf) < read) {
+        gst_buffer_set_size (outbuf, read);
+      }
+      out_mem = gst_buffer_get_all_memory (outbuf);
+    }
+
+    nns_logi ("allocated buffer size: %lu", gst_buffer_get_size (outbuf));
```
dongju-chae, line 154: I think this log may not provide useful info. when `need_alloc` is true.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -93,6 +93,13 @@ gst_tensor_query_socket_receive (GSocket * socket, GCancellable * cancellable,
   gsize read;
   GError *err = NULL;
   GstMapInfo map;
+  GstMemory *out_mem;
+  gboolean need_alloc;
+
+  if (!socket || !outbuf) {
```
wooksong, line 99: ```socket``` would never be NULL and

```c
*outbuf = gst_buffer_new_and_alloc (0UL);
```
returns NULL according to https://github.com/genesi/gstreamer/blob/9d3066642e54c032200b919281f05427b1014527/gst/gstbuffer.c#L341-L356. Is it right? 

If you want to null-checking for ```outbuf```, do after line 340.

myungjoo:
Off-Topic: in the commit message, you are supposed to describe the reason why you have added another argument and why you have added new conditions. Simply stating "- tcp server src element" is inappropriate.

again4you:
What if this PR is in-progress state and you just submit it to share your code, then please use `Draft submits` or add `[WIP]` tag.

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: Right. Moved it to log new allocated

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: It seems it returns buffer object without memory allocated but NULL. (https://developer.gnome.org/gstreamer/stable/gstreamer-GstBuffer.html#gst-buffer-new-allocate)
While `*outbuf = gst_buffer_new_and_alloc (0UL);` is valid, I removed unnecessary logging. Thanks

dongju-chae (approved):
LGTM

PR #3408, opened 2021-07-15
Decoder: Bounding boxes: Mobilenet SSD mode enhancements

xroumegue:
Hi nnstreamer folks,

This pull request brings some enhancements to bounding boxes decoder element, while running in mobilenet ssd mode.

The first commit adds support for setting the post-processing parameters through option3. If not set, the default values are used, such as backwards compatibility is preserved.

The second commit brings an optimization while decoding, preventing to call costly expf mathematical function for each possible box while assessing a valid detected object. Instead, the comparison is done against values in the sigmoid domain.
This reduces the decoder element latency from ~20 ms to ~2ms on NXP iMX8MP platform.

Regards,
 Xavier

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -35,8 +35,27 @@
  *                     tflite-ssd (deprecated, recommend to use mobilenet-ssd)
  * option2: Location of label file
  *          This is independent from option1
- * option3: Location of box prior file (ssd) or any option1-dependent values
+ * option3: Any option1-dependent values
  *          !!This depends on option1 values!!
+ *          for mobilenet-ssd mode:
+ *            The option3 definition scheme is, in order, the following:
+ *                - box priors location file (mandatory)
+ *                - Detection threshold (optional, default set to 0.5)
+ *                - Y box scale (optional, default set to 10.0)
+ *                - X box scale (optional, default set to 10.0)
+ *                - h box scale (optional, default set to 5.0)
+ *                - w box scale (optional, default set to 5.0)
+ *                - IOU box valid threshold (optional, default set to 0.5)
+ *            The default parameters value could be set in the following ways:
+ *            option3=box-priors.txt:0.5:10.0:10.0:5.0:5.0:0.5
+ *            option3=box-priors.txt
+ *            option3=box-priors.txt::::::
+ *
+ *            It's possible to set only few values, using the default values for
+ *            those not specified through the command line.
+ *            You could specify respectively the detection and IOU thresholds to 0.65
+ *            and 0.6 with the option3 parameter as follow:
+ *            option3=box-priors.txt:0.65:::::0.6
  * option4: Video Output Dimension (WIDTH:HEIGHT)
  *          This is independent from option1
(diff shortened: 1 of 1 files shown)


myungjoo:
Around Line 30~50, please add the descriptions for the additional parts of option3:

```

The option3 definition scheme is, in order, the following:
- box priors location file (mandatory)
- Detection threshold (optional, default set to 0.5)
- Y box scale (optional, default set to 10.0)
- X box scale (optional, default set to 10.0)
- h box scale (optional, default set to 5.0)
- w box scale (optional, default set to 5.0)
- IOU box valid threshold (optional, default set to 0.5)

The default parameters value could be set through the following option3
example:

... option3=box-priors.txt:0.5:10.0:10.0:5.0:5.0:0.5 ...
or
... option3=box-priors.txt ...
or
... option3=box-priors.txt:::::: ...

It's possible to set only few values, using the default values for
those not specified through the command line.
You could specify respectively the detection and IOU thresholds to 0.65
and 0.6 with the option3 parameter as follow:

... option3=box-priors.txt:0.65:::::0.6 ...
```

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -225,6 +226,17 @@ _get_mobilenet_ssd_pp_threshold (bounding_boxes * bdata)
...
 
+/** @brief Mathematic inverse of sigmoid function, aka logit */
+static float logit(float x)
+{
+  if (x < 0)
+    x = 0;
+  if (x > 1)
+    x = 1;
+
+  return log(x / (1.0 - x));
```
myungjoo, line 237: Could you please try to avoid division-by-zero? (unless dbz is intended)

xroumegue: I verified this implementation behaves correctly, with regard to infinity handling in the log().
So I believe it's safe but some exception might be raised in log() depending on the implementation.
Hence, I modified the logit() implementation to avoid infinity handling in log(), and returns directly +- INFINITY for input values out of range.
The simple scalar comparison operation should not be an issue while handling INFINITY in the rest of the code.

myungjoo (approved):
LGTM

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -35,8 +35,27 @@
...
+ *		The default parameters value could be set in the following ways:
+ *		option3=box-priors.txt:0.5:10.0:10.0:5.0:5.0:0.5
+ *		option3=box-priors.txt
+ *		option3=box-priors.txt::::::
+ *
+ *		It's possible to set only few values, using the default values for
+ *		those not specified through the command line.
+ *		You could specify respectively the detection and IOU thresholds to 0.65
+ *		and 0.6 with the option3 parameter as follow:
+ *		option3=box-priors.txt:0.65:::::0.6
```
myungjoo, line 58: You are mostly using tabs here while other similar comments and a few of your comments don't. Please use spaces instead of tabs for the consistency.

xroumegue: Done. Sorry for that. My bad.

anyj0527 (approved):
LGTM👍 Great thanks!

PR #3409, opened 2021-07-15
Common: refactor silent_debug() and make it common

myungjoo:
1. Provide common macro of silent_debug and silent_debug_caps,
 which had been repeatedly defined in each plugin.
2. Refactor the two macros to get "object" as an argument
 so that it wouldn't use implicit argument exceot for "DBG".

The same macros are not touched in /ext directory
because they are not supposed to include "tensor_common.h",
which is an internal header.

@todo There are a few subplugins in /ext (not tensor-filter)
that uses "tensor_common.h". They should be refactored not to
use internal headers so that we can separate them to
independent repo anytime. Although we don't have such plans,
we need to keep them as appropriate examples for users.

### DIFF
--- a/gst/nnstreamer/tensor_common.h
+++ b/gst/nnstreamer/tensor_common.h
@@ -195,5 +195,31 @@ gst_tensor_pad_possible_caps_from_config (GstPad * pad, const GstTensorsConfig *
 extern gboolean
 gst_tensor_pad_caps_is_flexible (GstPad * pad);
 
+/******************************************************
+ ************ Commonly used debugging macros **********
+ ******************************************************
+ */
+/**
+ * @brief Macro for debug message.
+ */
+#define silent_debug(self, ...) do { \
+    if (DBG) { \
+      GST_DEBUG_OBJECT (self, __VA_ARGS__); \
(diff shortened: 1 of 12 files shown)


gst/nnstreamer/tensor_common.h
```diff
@@ -195,5 +195,38 @@ gst_tensor_pad_possible_caps_from_config (GstPad * pad, const GstTensorsConfig *
...
+    } \
+  } while (0)
+
+/**
+ * @brief Macro for capability debug message.
+ */
+#define silent_debug_caps(self, caps, msg) do { \
+  if (DBG) { \
+    if (caps) { \
+      GstStructure *caps_s; \
```
jaeyun-jung, line 217: suggestion : use gst_caps_to_string(caps) it simply returns caps string.

dongju-chae (approved):
LGTM!

gst/nnstreamer/tensor_common.h
myungjoo: Applied.

anyj0527 (approved):
LGTM👍

PR #3413, opened 2021-07-19
[Build] framework priority option

jaeyun-jung:
Add meson-option to set fw priority for file extension (.nb, .bin)


again4you (approved):
Nice work.
```bash
[Normal case]
framework_priority_nb=
framework_priority_bin=openvino

[VD case]
framework_priority_nb=vd_aifw
framework_priority_bin=vd_aifw
```

gichan-jang:
The number of models is also used when auto-detecting the framework.
In this case, there is no overlap yet, so do we need priorities?

jaeyun-jung:
@gichan-jang For only one mode file (.ext), now we didn't consider multi-models case to detect fw from conf file.

gichan-jang (approved):
LGTM

PR #3418, opened 2021-07-20
[Config/Format] tensor format in config

jaeyun-jung:
Set tensor format from caps struct.
tensor stream has single format, so format in tensor-info struct is unnecessary.

### DIFF
--- a/gst/nnstreamer/tensor_common_pipeline.c
+++ b/gst/nnstreamer/tensor_common_pipeline.c
@@ -271,6 +271,9 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
   guint counting, empty_pad;
   GstTensorsConfig in_configs;
   GstClockTime base_time = 0;
+  guint i, n_mem;
+  GstMemory *in_mem[NNS_TENSOR_SIZE_LIMIT];
+  tensor_format in_formats[NNS_TENSOR_SIZE_LIMIT];
 
   g_return_val_if_fail (collect != NULL, FALSE);
   g_return_val_if_fail (sync != NULL, FALSE);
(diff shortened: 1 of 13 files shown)


jaeyun-jung:
To reviewers, please ignore android build error.
PR https://github.com/nnstreamer/api/pull/76 on api repo will fix error.

anyj0527 (approved):
LGTM 👍👍

gst/nnstreamer/tensor_common_pipeline.c
```diff
@@ -374,20 +377,22 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
...
+      if (gst_tensors_config_is_flexible (&in_configs))
+        configs->format = _NNS_TENSOR_FORMAT_FLEXIBLE;
+
       for (i = 0; i < n_mem; ++i) {
-        mem = gst_buffer_get_memory (buf, i);
-        gst_buffer_append_memory (tensors_buf, mem);
+        in_mem[counting] = gst_buffer_get_memory (buf, i);
+
         configs->info.info[counting] = in_configs.info.info[i];
+        in_formats[counting] = in_configs.format;
```
gichan-jang, line 395: Q1. Why do we need `counting`?
Q2. Is it necessary to save the format of the input for each memory?

jaeyun-jung: 1. max number of tensors in gst-buffer is 16. 
2. if tensor-mux has two sink pads with different format - 1st pad is static and 2nd pad is flexible, output buffer should be flexible and tensor-mux should append header in the memories from only 1st pad. The memories from 2nd pad already have header in each mem block because it is flexble tensor.

gichan-jang (approved):
LGTM

PR #3421, opened 2021-07-22
[Query] Define tensor query common data structures.

gichan-jang:
Re-define tensor query common data structures.
 - GSocket or MQTT_handled are managed from tensor_query_common.
 - Each element manages a unique id which is generated using IP and port.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_common.h
+++ b/gst/nnstreamer/tensor_query/tensor_query_common.h
@@ -8,52 +8,101 @@
  * @see    https://github.com/nnstreamer/nnstreamer
  * @author Gichan Jang <gichan2.jang@samsung.com>
  * @bug    No known bugs except for NYI items
- * @see    gsttcpserversink, gsttcpserversrc
  */
 
-#ifndef __GST_TENSOR_QUERY_COMMON_H__
-#define __GST_TENSOR_QUERY_COMMON_H__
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -18,42 +18,72 @@
 #include <gst/gst.h>
```
jaeyun-jung, line 18: TODO : glib, gst, and tensor_common.h
remove these and use std only (e.g. tensor_typedef.h is ok)

gichan-jang: Thanks. Removed!

gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -8,52 +8,94 @@
...
+  int64_t sent_time;
+  uint64_t duration;
+  uint64_t dts;
+  uint64_t pts;
+} TensorQueryDataInfo;
+
+typedef struct
+{
+  uint8_t *data;
+  unsigned long size;
```
jaeyun-jung, line 59: 'size_t' is better for this case :)

jaeyun-jung (approved):

kimjh12 (approved):
Looks great

PR #3426, opened 2021-07-28
Move nnstreamer_example to tests/

myungjoo:
With nnstreamer_example directory at the top,
some users are confused and thought that this directory
is the official nnstreamer example apps.

However, these are merely example codes for unit tests;
thus, they should be moved into tests to prevent
the confusion.


jaeyun-jung (approved):

kimjh12 (approved):
LGTM!!

PR #3427, opened 2021-07-28
[Gst/MQTT] Introduce the Unix epoch synchronization using NTP @open sesame 09/29 10:29

wooksong:
This PR introduces utility functions to get timestamps from the given NTP server and the self-synchronization mechanism of the Unix epoch timestamps using them. 

To resolve the following issue, this PR also includes a patch to apply caching mechanism according to the polling interval given from the NTP server.
> Note that if the 'ntp-sync' property is set to true (which means that if this feature is enabled), the pipeline would be stuck after a while. This is because most of the public NTP servers restrict clients not to poll too frequently. To fix it, polling intervals and caching mechanisms should be applied to the NPT utility functions.

**Self evaluation:**

### DIFF
--- a/gst/mqtt/mqttsink.c
+++ b/gst/mqtt/mqttsink.c
@@ -27,6 +27,7 @@
 #include <nnstreamer_util.h>
 
 #include "mqttsink.h"
+#include "ntputil.h"
 
 static GstStaticPadTemplate sink_pad_template = GST_STATIC_PAD_TEMPLATE ("sink",
     GST_PAD_SINK, GST_PAD_ALWAYS, GST_STATIC_CAPS_ANY);
@@ -52,6 +53,8 @@ enum
   PROP_NUM_BUFFERS,
   PROP_MAX_MSG_BUF_SIZE,
   PROP_MQTT_QOS,
+  PROP_MQTT_NTP_SYNC,
+  PROP_MQTT_NTP_SRVS,
 
   PROP_LAST
 };
@@ -68,6 +71,8 @@ enum
   DEFAULT_MQTT_PUB_WAIT_TIMEOUT = 1,    /* 1 secs */
   DEFAULT_MAX_MSG_BUF_SIZE = 0, /* Buffer size is not fixed */
   DEFAULT_MQTT_QOS = 0,         /* fire and forget */
+  DEFAULT_MQTT_NTP_SYNC = FALSE,
+  MAX_LEN_PROP_NTP_SRVS = 4096,
 };
 
 static guint8 sink_client_id = 0;
@@ -78,6 +83,7 @@ static const gchar DEFAULT_MQTT_CLIENT_ID[] = "$HOST_$PID_^[0-9][0-9]?$|^255$";
 static const gchar DEFAULT_MQTT_CLIENT_ID_FORMAT[] = "%s_%u_sink%u";
 static const gchar DEFAULT_MQTT_PUB_TOPIC[] = "$client-id/topic";
 static const gchar DEFAULT_MQTT_PUB_TOPIC_FORMAT[] = "%s/topic";
+static const gchar DEFAULT_MQTT_NTP_SERVERS[] = "pool.ntp.org:123";
 
 /** Function prototype declarations */
 static void
@@ -131,6 +137,12 @@ static gint gst_mqtt_sink_get_num_buffers (GstMqttSink * self);
 static void gst_mqtt_sink_set_num_buffers (GstMqttSink * self, const gint num);
 static gint gst_mqtt_sink_get_mqtt_qos (GstMqttSink * self);
 static void gst_mqtt_sink_set_mqtt_qos (GstMqttSink * self, const gint qos);
+static gboolean gst_mqtt_sink_get_mqtt_ntp_sync (GstMqttSink * self);
+static void gst_mqtt_sink_set_mqtt_ntp_sync (GstMqttSink * self,
+    const gboolean flag);
+static gchar *gst_mqtt_sink_get_mqtt_ntp_srvs (GstMqttSink * self);
+static void gst_mqtt_sink_set_mqtt_ntp_srvs (GstMqttSink * self,
+    const gchar * pairs);
 
 static void cb_mqtt_on_connect (void *context,
     MQTTAsync_successData * response);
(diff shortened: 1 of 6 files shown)


myungjoo:
CC: @anyj0527 @gichan-jang @kimjh12 You all need to study and understand this.

gst/mqtt/ntputil.c
```diff
@@ -0,0 +1,289 @@
...
+
+/**
+ * @brief Initialize the NTP utility library using pthread_once
+ */
+void
+ntputil_init (void)
+{
+  pthread_spin_init (&lock, PTHREAD_PROCESS_SHARED);
+}
+
```
anyj0527, line 136: No need to add some api for `pthread_spin_destroy` ?

wooksong: @anyj0527 Actually we need to call ```pthread_spin_destory```, but since the ```lock``` is statically allocated and it has the same life cycle as that of this library, this can be omitted (there is no point to invoke ```pthread_spin_destroy```).

gst/mqtt/mqttsink.c
```diff
@@ -1067,6 +1120,92 @@ gst_mqtt_sink_set_mqtt_qos (GstMqttSink * self, const gint qos)
...
+    guint16 port;
+
+    pair = pair_arrs[i];
+    memset (hname, '\0', MAX_LEN_NTP_HNAME);
+
+    sscanf (pair, "%[^:]:%hd", hname, &port);
+    self->mqtt_ntp_hnames[i] = g_strdup (hname);
+    self->mqtt_ntp_ports[i] = port;
+  }
+
```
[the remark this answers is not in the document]
wooksong: @gichan-jang Well, in my intention, ```ntp-srvs``` is an optional property. How about explicitly declaring it in the property description?

gst/mqtt/ntputil.c
```diff
@@ -0,0 +1,289 @@
...
+    }
+    recv_frac = ntohl (packet.xmit_ts.frac);
+
+    /**
+     * @note NTP uses an epoch of January 1, 1900 while the Unix epoch is
+     * the number of seconds that have elapsed since January 1, 1970. For this
+     * reason, we subtract 70 years worth of seconds from the seconds since 1900
+     */
+    ret = (int64_t) (recv_sec - NTPUTIL_TIMESTAMP_DELTA);
+    ret *= NTPUTIL_SEC_TO_USEC_MULTIPLIER;
```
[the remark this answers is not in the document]
wooksong: IMO, the overflow case is impossible since ```((uint32_t) recv_sec - 2208988800)  * 1000000``` is always smaller than ```9223372036854775807```, which is max of ```int64_t```.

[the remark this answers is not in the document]
gichan-jang: Ah, `recv_sec` is uint32.. Thanks for the explanation.

gst/mqtt/mqttsink.c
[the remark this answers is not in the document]
gichan-jang: I asked because I thought specifying the ntp server is using ntp synchronization.
This is already well explained in the property description.
`valid only if ntp-sync is true`
If it's your intention, it's fine.

gichan-jang (approved):
LGTM

gst/mqtt/mqttsink.c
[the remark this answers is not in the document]
wooksong: @gichan-jang Thanks!

gst/mqtt/mqttsink.c
```diff
@@ -1067,6 +1120,92 @@ gst_mqtt_sink_set_mqtt_qos (GstMqttSink * self, const gint qos)
...
+{
+  gchar **pair_arrs = NULL;
+  guint hnum = 0;
+  gchar *pair;
+  guint i;
+
+  if (g_strcmp0 (self->mqtt_ntp_srvs, pairs) == 0)
+    return;
+
+  self->mqtt_ntp_srvs = g_strdup (pairs);
```
dongju-chae, line 1164: Is it okay not to free `self->mqtt_ntp_srvs` before assignment?

gst/mqtt/mqttsink.c
```diff
@@ -1067,6 +1120,92 @@ gst_mqtt_sink_set_mqtt_qos (GstMqttSink * self, const gint qos)
...
+
+  self->mqtt_ntp_num_srvs = hnum;
+  for (i = 0; i < hnum; i++) {
+    gchar hname[MAX_LEN_NTP_HNAME];
+    guint16 port;
+
+    pair = pair_arrs[i];
+    memset (hname, '\0', MAX_LEN_NTP_HNAME);
+
+    sscanf (pair, "%[^:]:%hd", hname, &port);
```
dongju-chae, line 1191: How about checking the return value of `sscanf`?

gst/mqtt/ntputil.c
anyj0527: Right, It could be omitted. Thanks!

gst/mqtt/mqttsink.c
myungjoo: Actually, ```sscanf``` is the common culprit of security checks (this will probably cause SVACE error).

gst/mqtt/mqttsink.c
wooksong: @dongju-chae Right!

gst/mqtt/mqttsink.c
wooksong: @myungjoo @dongju-chae ```scanf``` has been eliminated.

wooksong:
@dongju-chae @myungjoo 

- The missing ```g_free```s are added.
- ```scanf``` is eliminated.

dongju-chae (approved):
LGTM!

gst/mqtt/mqttsink.c
```diff
@@ -1067,6 +1120,103 @@ gst_mqtt_sink_set_mqtt_qos (GstMqttSink * self, const gint qos)
...
+    gulong port_ul;
+
+    pair = pair_arrs[i];
+    hname_port = g_strsplit (pair, ":", 2);
+    hname = hname_port[0];
+    port_ul = strtoul (hname_port[1], &eport, 10);
+    if ((port_ul == 0) || (port_ul > UINT16_MAX)) {
+      self->mqtt_ntp_num_srvs--;
+    } else {
+      self->mqtt_ntp_hnames[j++] = g_strdup (hname);
```
jaeyun-jung, line 1200: 1200 and 1201 both increases index (j++), this may raise exception.

myungjoo: this potentially causes out-of-bounds... and looks like an error.

wooksong: My bad. It is fixed.

jaeyun-jung (approved):

myungjoo:
Merge after #3509

PR #3428, opened 2021-07-29
[Test] fix res leak

jaeyun-jung:
fix res leak in tvm testcase.


kimjh12 (approved):
LGTM! Thanks for the fix

myungjoo (approved):
LGTM

PR #3430, opened 2021-07-30
[Query] Implement common utils for TCP

kimjh12:
Common API functions used for both client and server elements
Currently based on TCP connections

TODO
- Enable timeout

---

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_common.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_common.c
@@ -7,6 +7,7 @@
  * @brief  Utility functions for tensor query
  * @see    https://github.com/nnstreamer/nnstreamer
  * @author Gichan Jang <gichan2.jang@samsung.com>
+ * @author Junhwan Kim <jejudo.kim@samsung.com>
  * @bug    No known bugs except for NYI items
  */
 
@@ -17,17 +18,13 @@
 #include <gio/gio.h>
 #include <gio/gsocket.h>
 #include <stdint.h>
+#include <errno.h>
 #include <nnstreamer_util.h>
+#include <nnstreamer_log.h>
 #include "tensor_query_common.h"
 
-/**
- * @brief hashtable for managing server and client data
- * key: id, value: server or client data
- */
-static GHashTable *table = NULL;
-
-/** @brief Protects table */
-G_LOCK_DEFINE_STATIC (splock);
+#define TENSOR_QUERY_SERVER_DATA_LEN 128
+#define TCP_BACKLOG 1
 
 /**
  * @brief Structures for tensor query server data.
@@ -35,18 +32,18 @@ G_LOCK_DEFINE_STATIC (splock);
 typedef struct
 {
   TensorQueryProtocol protocol;
-  GstTensorsInfo src_info;
-  GstTensorsInfo sink_info;
-
-  unsigned long long current_client_id;
+  gchar *server_host;
+  guint32 server_port;
+  /* network info */
   union
   {
     struct
     {
-      GSocket *src_socket;
-      GSocket *sink_socket;
+      GSocket *server_socket;
       GCancellable *cancellable;
     };
+    // check the size of struct is less
+    guint8 _dummy[TENSOR_QUERY_SERVER_DATA_LEN];
   };
 } TensorQueryServerData;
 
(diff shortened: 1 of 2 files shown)


myungjoo:
https://www.tutorialspoint.com/what-is-the-correct-way-to-use-printf-to-print-a-size-t-in-c-cplusplus

Use %zu instead of %lu for gsize?

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -66,26 +73,447 @@ typedef struct
...
-  UNUSED (id);
-  UNUSED (ip);
-  UNUSED (port);
-  UNUSED (timeout_ms);
-
-  /** NYI: To avoid `defined but not used` */
-  if (NULL == table) {
+  TensorQueryServerData *server_data = g_new (TensorQueryServerData, 1);
+  server_data->protocol = _TENSOR_QUERY_PROTOCOL_END;
+  server_data->src_info = src_info;
```
dongju-chae, line 95: Q: Is it safe just to assign the pointer here?

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -66,26 +73,447 @@ typedef struct
...
+
+  server_data->current_client_id = NULL;
```
dongju-chae, line 97: It should be checked and freed in `gst_query_server_data_free`

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: I think it's safe because `server_data` belongs to `serversrc` conceptually but would this be a bad practice?

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: Fixed!! Thank you

gichan-jang (changes requested):
Let's remove all the dependencies of GStreamer from the tensor_query_common.
And use `uint64_t id` as hash key which is generated by `nnstreamer_query_request_id` not `client_ip:client_port_server_ip:server_port`.

gst/nnstreamer/tensor_query/tensor_query_serversrc.c
```diff
@@ -14,105 +14,319 @@
...
+        mem = gst_buffer_get_all_memory (*outbuf);
+        gst_memory_map (mem, &map, GST_MAP_READWRITE);
+        data.data.size = data_size;
+        data.data.data = map.data;
+        nns_logi ("allocated buffer size: %zu", gst_buffer_get_size (*outbuf));
+        gst_memory_unmap (mem, &map);
+        gst_memory_unref (mem);
+        break;
+
+      case _TENSOR_QUERY_CMD_TRANSFER_DATA:
```
gichan-jang, line 311: Please considering streaming data.
I think received data should be handled here..
And sequence of the command buffers are
`_TENSOR_QUERY_CMD_TRANSFER_START | _TENSOR_QUERY_CMD_TRANSFER_DATA | ... | _TENSOR_QUERY_CMD_TRANSFER_DATA | _TENSOR_QUERY_CMD_TRANSFER_END`
The capability of the server does not change. So you don't nned to send a start cmd buf every time.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -89,3 +140,397 @@ nnstreamer_query_connect (uint64_t id, const char *ip, uint32_t port,
...
+      conn->protocol = sdata->protocol;
+      client_host =
+          g_inet_address_to_string (g_inet_socket_address_get_address (
+              (GInetSocketAddress *) saddr));
+      client_port =
+          g_inet_socket_address_get_port ((GInetSocketAddress *) saddr);
+      conn->socket = client_socket;
+      conn->cancellable = g_cancellable_new ();
+
+      key =
```
gichan-jang, line 413: resource leak?
Um.. Who and How do we destroy the hash table?

kimjh12: It's to be handled by sink. Sink will reuse the connection value, table key can be destroyed once sink has done with sending response

kimjh12:
To make things go quick with others I'm splitting this PR into only common regarding parts. PTAL

gst/nnstreamer/tensor_query/tensor_query_common.c
gichan-jang: The key generated from the source should be managed from the source.
Let's release the key when connection is closed.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -89,3 +140,397 @@ nnstreamer_query_connect (uint64_t id, const char *ip, uint32_t port,
...
+      }
+      g_object_unref (sdata->cancellable);
+      break;
+    default:
+      /* NYI */
+      nns_logw ("Invalid protocol");
+      break;
+  }
+  g_free (sdata->server_host);
+  g_free (sdata);
```
gichan-jang, line 326: Let's free hash table data and destroy hash table here.

gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -76,31 +78,63 @@ typedef struct
...
-nnstreamer_query_connect (uint64_t id, const char *ip, uint32_t port, uint32_t timeout_ms);
+nnstreamer_query_connect (int64_t id, const char *ip, uint32_t port, uint32_t timeout_ms);
 
 /**
  * @brief send command to connected device.
  * @return 0 if OK, negative value if error
  */
 extern int
-nnstreamer_query_send (uint64_t id, TensorQueryCommandData *data, uint32_t timeout_ms);
+nnstreamer_query_send (int64_t id, TensorQueryCommandData *data, uint32_t timeout_ms);
```
gichan-jang, line 98: As we discussed, `tensor_query_client` don't need hash table.
Please replace `int64_t id` as `void * connection`.
And apply it to other APIs commonly used by servers and clients.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -17,36 +18,42 @@
...
+ * key: id, value: TensorQueryConnection
  */
 static GHashTable *table = NULL;
 
 /** @brief Protects table */
 G_LOCK_DEFINE_STATIC (splock);
 
+#define TENSOR_QUERY_SERVER_DATA_LEN 128
+#define TCP_BACKLOG 1
+#define TCP_TIMEOUT 10
```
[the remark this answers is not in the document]
kimjh12: Actually timeout turned out to be passed as function params.. I will remove this. Thanks

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: I see. I think calling `nnstreamer_query_close` (L250) can do.. can it??

gichan-jang:
> To make things go quick with others I'm splitting this PR into only common regarding parts. PTAL

Thanks! It will be easy to implement the tensor query client.

gst/nnstreamer/tensor_query/tensor_query_common.h
kimjh12: Updated!

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: I removed hash table from common which is only dependent to server. It would belong to query_server.c

myungjoo:
Wait wait..

Why ```void *```?

kimjh12:
> Wait wait..
> 
> Why `void *`?

To avoid package dependency of `common.h`?

gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -74,33 +76,58 @@ typedef struct
...
+/**
+ * @brief close connection with corresponding id.
+ * @return 0 if OK, negative value if error
+ */
+extern int
+nnstreamer_query_close (void * connection);
+
+/* server */
+extern void *
+nnstreamer_query_server_accept (void *server_data);
```
myungjoo, line 109: Need more explanation for external interfaces

myungjoo:
> > Wait wait..
> > Why `void *`?
> 
> To avoid package dependency of `common.h`?

If you are using ```void *``` to hide details from users and cut coupling,
please use different names for different pointers: connections vs server_data.
Using ```void *``` for both handles confuses developers.

E.g., ```typedef void * query_server_handle``` and ```typedef void * query_connection_handle```.

Name the functions so that users can identify which  is "get/register/open/new" and which is "put/unregister/close/free".

gst/nnstreamer/tensor_query/tensor_query_common.h
kimjh12: Fixed!

kimjh12:
> > > Wait wait..
> > > Why `void *`?
> > 
> > 
> > To avoid package dependency of `common.h`?
> 
> If you are using `void *` to hide details from users and cut coupling,
> please use different names for different pointers: connections vs server_data.
> Using `void *` for both handles confuses developers.
> 
> E.g., `typedef void * query_server_handle` and `typedef void * query_connection_handle`.
> 
> Name the functions so that users can identify which is "get/register/open/new" and which is "put/unregister/close/free".
> It'd be great if you add a doxygen entry that demonstrates a flow of API usage (e.g., open->run->close)
> 
> Note that these notations will be exported via Tizen Public API later (7.0); thus, you need to name them more carefully.
> Type and Extern-Function Naming Review: @again4you @jaeyun-jung

Updated `void *` types based on your suggestion.
Calling flow and role of functions can be changed as we develop and integrate other protocols so we will keep finding out the best API definition. Thanks for the review

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,422 @@ typedef struct
...
+/**
+ * @brief Connection info structure
+ */
+typedef struct
+{
+  TensorQueryProtocol protocol;
+  gchar *host;
+  guint32 port;
+  /* network info */
+  union
```
wooksong, line 75: Is this required to support other protocols rather than TCP? If it is, IMHO, you could pack those in ```union``` when it is really required.

myungjoo:
> > > > Wait wait..
> > > > Why `void *`?
> > > 
> > > 
> > > To avoid package dependency of `common.h`?
> > 
> > 
> > If you are using `void *` to hide details from users and cut coupling,
> > please use different names for different pointers: connections vs server_data.
> > Using `void *` for both handles confuses developers.
> > E.g., `typedef void * query_server_handle` and `typedef void * query_connection_handle`.
> > Name the functions so that users can identify which is "get/register/open/new" and which is "put/unregister/close/free".
> > It'd be great if you add a doxygen entry that demonstrates a flow of API usage (e.g., open->run->close)
> > Note that these notations will be exported via Tizen Public API later (7.0); thus, you need to name them more carefully.
> > Type and Extern-Function Naming Review: @again4you @jaeyun-jung
> 
> Updated `void *` types based on your suggestion.
> Calling flow and role of functions can be changed as we develop and integrate other protocols so we will keep finding out the best API definition. Thanks for the review

Resolved. LGTM!

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,422 @@ typedef struct
...
+/**
+ * 
+ * @brief return initialized server_data
+ * @param src_info tensors info shared with serversrc element
+ * @return void* casted type of TensorQueryServerData*
+ */
+query_server_handle
+nnstreamer_query_server_data_new (void)
+{
+  TensorQueryServerData *sdata = g_new (TensorQueryServerData, 1);
```
wooksong, line 252: How about using g_try_new(... and adding some null checking after it?

gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -74,33 +79,63 @@ typedef struct
...
+/**
+ * @brief accept client connection
+ * @param server_data TensorQueryServerData
+ * @return void* casted type of TensorQueryConnection*
+ */
+extern query_connection_handle
+nnstreamer_query_server_accept (query_server_handle server_data);
+
+/**
+ * @brief return initialized server_data
```
wooksong, line 120: server_data? Is it a handle of something?

gst/nnstreamer/tensor_query/tensor_query_common.h
```diff
@@ -74,33 +79,63 @@ typedef struct
...
+ * @brief close connection with corresponding id.
+ * @return 0 if OK, negative value if error
+ */
+extern int
+nnstreamer_query_close (query_connection_handle connection);
+
+/* server */
+/**
+ * @brief accept client connection
+ * @param server_data TensorQueryServerData
```
wooksong, line 113: Which one is right between query_server_handle and TensorQueryServerData? and Please write more details for the doxygen tags.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,422 @@ typedef struct
...
+static gboolean
+query_tcp_receive (GSocket * socket, uint8_t * data, size_t size,
+    GCancellable * cancellable)
+{
+  gsize bytes_received = 0;
+  gssize rret;
+  GError *err;
+
+  while (bytes_received < size) {
+    rret = g_socket_receive (socket, (gchar *) data + bytes_received,
```
wooksong, line 443: Is it safe to use mixed signed and unsigned data types here and at line 450?

kimjh12: I think it 's ok as long as we use it for byte buffer.
For L450, it looks fine because size is added only if >= 0

gst/nnstreamer/tensor_query/tensor_query_common.h
kimjh12: Fixed to `query_server_handle`. Thanks

gst/nnstreamer/tensor_query/tensor_query_common.h
kimjh12: It's same as `query_server_handle`. Fixed it to avoid confusion

gst/nnstreamer/tensor_query/tensor_query_common.c
kimjh12: That looks better. Thank you

dongju-chae (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,505 @@ typedef struct
...
+query_tcp_receive (GSocket * socket, uint8_t * data, size_t size,
+    GCancellable * cancellable);
+static gboolean query_tcp_send (GSocket * socket, uint8_t * data, size_t size,
+    GCancellable * cancellable);
+
+/**
+ * @brief Create requested socket.
+ */
+static gboolean
+gst_tensor_query_socket_new (query_connection_handle conn_h,
```
gichan-jang, line 97: Let's pass TensorQueryProtocol and create a socket with the protocol.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,505 @@ typedef struct
...
+        if (g_error_matches (err, G_IO_ERROR, G_IO_ERROR_CANCELLED)) {
+          nns_logd ("Cancelled connecting");
+        } else {
+          nns_loge ("Failed to connect to host");
+        }
+        goto tcp_fail;
+      }
+      g_object_unref (saddr);
+      break;
+    tcp_fail:
```
gichan-jang, line 180: If fail, need to free `conn`.

gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -60,32 +57,505 @@ typedef struct
...
+  }
+
+  *saddr = g_inet_socket_address_new (addr, conn->port);
+  g_object_unref (addr);
+
+  /* create sending client socket */
+  /** @todo Support UDP protocol */
+  conn->socket =
+      g_socket_new (g_socket_address_get_family (*saddr), G_SOCKET_TYPE_STREAM,
+      G_SOCKET_PROTOCOL_TCP, &err);
```
gichan-jang, line 136:
```suggestion
      g_socket_new (g_socket_address_get_family (*saddr), G_SOCKET_TYPE_STREAM,
      conn->protocol, &err);
```

gichan-jang (approved):
LGTM!

myungjoo (approved):
LGTM

PR #3431, opened 2021-07-30
[Build] fix build error

jaeyun-jung:
Fix build error after changing warning level.


myungjoo (approved):
LGTM

PR #3432, opened 2021-08-02
[Subplugin/Py] prevent resizing the array length

jaeyun-jung:
In the sub-plugins for python, prevent resizing the array length when creating new array object.


dongju-chae (approved):
Looks good! It seems better to maintain reference counts.

myungjoo (approved):
LGTM

PR #3455, opened 2021-08-18
[tensor_query] Fix double free in tensor_query_server @open sesame 08/19 10:24

anyj0527:
- Assign NULL value after call `query_server_data_free` to  prevent double free
- Remove some trailing spaces

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_common.h
+++ b/gst/nnstreamer/tensor_query/tensor_query_common.h
@@ -146,7 +146,7 @@ nnstreamer_query_server_data_free (query_server_handle server_data);
 
 /**
  * @brief set server handle params and setup server
- * @return 0 if OK, negative value if error 
+ * @return 0 if OK, negative value if error
  */
 extern int
 nnstreamer_query_server_init (query_server_handle server_data,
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/tensor_query/tensor_query_common.c
```diff
@@ -379,9 +379,9 @@ nnstreamer_query_server_data_new (void)
  * @brief free server handle
  */
 void
-nnstreamer_query_server_data_free (query_server_handle server_data)
+nnstreamer_query_server_data_free (query_server_handle * server_data)
```
dongju-chae, line 382: IMO, it looks not a standard way to use a handle. Such handling should be done in the caller.
For example, after calling `nnstreamer_query_server_data_free`, you may set `src->server_data = NULL`.

anyj0527: Thanks! I think you're suggestion seems right. I updated the code, PTAL.

kimjh12 (approved):
LGTM!

gichan-jang (approved):
LGTM

wooksong (approved):
LGTM

dongju-chae (approved):
LGTM

PR #3469, opened 2021-08-27
[Common/Adapter] aggregation from multi clients

jaeyun-jung:
Use hashtable to aggregate buffers from multiple clients.
- create gst-adapter for each client id
- for normal case, set default adaptor (no meta in gst-buffer)


dongju-chae (approved):
LGTM

jaeyun-jung:
@gichan-jang After merging this PR, please update hash function and key value.
e.g. g_direct_hash() and g_direct_equal() to create hash-table. (key-destroy function should be null)

PR #3479, opened 2021-09-03
[lua/test] Add tc for multiple tensor types in lua filter

anyj0527:
- Add a testcase which covers all tensor types in tensor_filter::lua

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_filter_lua/unittest_filter_lua.cc
+++ b/tests/nnstreamer_filter_lua/unittest_filter_lua.cc
@@ -13,6 +13,7 @@
 #include <gst/gst.h>
 #include <unittest_util.h>
 
+#include <tensor_common.h>
 #include <nnstreamer_plugin_api_filter.h>
 #include <nnstreamer_util.h>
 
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_filter_lua/unittest_filter_lua.cc
```diff
@@ -891,6 +910,53 @@ end
...
+    ret = sp->invoke (NULL, NULL, data, &input, &output);
+    EXPECT_EQ (ret, 0);
+
+    g_free (model);
+  }
+
+  g_free (input.data);
+  g_free (output.data);
+  sp->close (&prop, &data);
+}
```
myungjoo, line 958: Please add negative cases (```${name}_${num}_n```)

tests/nnstreamer_filter_lua/unittest_filter_lua.cc
```diff
@@ -891,6 +892,55 @@ end
...
+  GstTensorMemory input, output;
+  GstTensorFilterProperties prop;
+
+  output.size = input.size = sizeof (int64_t) * 1 * 2 * 2 * 1;
+
+  /* alloc input data without alignment */
+  input.data = g_malloc (input.size);
+  output.data = g_malloc (output.size);
+
+  memset (input.data, 0, input.size);
```
gichan-jang, line 911: How about using g_malloc0?

tests/nnstreamer_filter_lua/unittest_filter_lua.cc
```diff
@@ -891,6 +892,55 @@ end
...
+
+  const GstTensorFilterFramework *sp = nnstreamer_filter_find ("lua");
+  EXPECT_NE (sp, nullptr);
+
+  for (int i = 0; i < _NNS_END; ++i) {
+    tensor_type ttype = (tensor_type) i;
```
gichan-jang, line 917: `i` is duplicated.

tests/nnstreamer_filter_lua/unittest_filter_lua.cc
anyj0527: g_malloc0 seems good. Thanks!

tests/nnstreamer_filter_lua/unittest_filter_lua.cc
anyj0527: I added a negative case for using invalid data type.

tests/nnstreamer_filter_lua/unittest_filter_lua.cc
```diff
@@ -891,6 +892,87 @@ end
...
+
+  const GstTensorFilterFramework *sp = nnstreamer_filter_find ("lua");
+  EXPECT_NE (sp, nullptr);
+
+  /**
+   * Invalid data type `uint128`. Type should be one of
+   * { float32, float64, int64, uint64, int32, uint32, int16, uint16, int8, uint8 }
+   */
+  const char *invalid_data_type = "uint128";
+  const gchar *model = g_strdup_printf ("\
```
jaeyun-jung, line 912: check returned type (char to const char)
g_strdup_printf() returns newly allocated string, need free function.
I recommend to add macro to get model string,
```
#define make_lua_model(dtype) "utTensorsInfo={num=1,dim={{1,2,2,1},},type={" dtype ",}} ...
....

in dataType00_n()
const gchar model[] = make_lua_model ("uint128");

in dataType01()
const gchar model[] = make_lua_model (gst_tensor_get_type_string (ttype));
```

anyj0527: Instead of macro, I updated the code to call `g_free` for allocated strings. Thanks, PTAL!

jaeyun-jung (approved):

myungjoo (approved):

PR #3490, opened 2021-09-13
[Filter/SharedModel] fix svace

jaeyun-jung:
Fix svace issue,
- set next node when releasing the list item
- use macro to define local lock for shared table

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -92,8 +92,11 @@ static gint _gtfc_setprop_IS_UPDATABLE (GstTensorFilterPrivate * priv,
 static gint _gtfc_setprop_ACCELERATOR (GstTensorFilterPrivate * priv,
     GstTensorFilterProperties * prop, const GValue * value);
 
+/**
+ * @brief mutex for shared model table.
+ */
+G_LOCK_DEFINE_STATIC (shared_model_table);
 static GHashTable *shared_model_table = NULL;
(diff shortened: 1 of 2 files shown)


dongju-chae (approved):
LGTM

gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -2942,33 +2958,34 @@ nnstreamer_filter_shared_model_remove (void *instance, const char *key,
...
   model_rep->referred_list = g_list_remove (model_rep->referred_list, instance);
   ml_logd ("The referred instance of sharing key: %s has been removed!", key);
+  ret = TRUE;
 
   /* remove key from table if list is empty */
   if (g_list_length (model_rep->referred_list) == 0) {
     if (free_callback)
       free_callback (model_rep->shared_interpreter);
-    ret = g_hash_table_remove (shared_model_table, key);
+    g_hash_table_remove (shared_model_table, key);
```
zhoonit, line 2985: minor though, how about checking if either comment or the implementation should be changed?
`@return TRUE if the new key and value are removed. FALSE if failed to remove it.`

zhoonit (approved):
LGTM :)

gst/nnstreamer/tensor_filter/tensor_filter_common.c
jaeyun-jung: updated

again4you (approved):
LGTM!

PR #3491, opened 2021-09-13
[Svace] fix res leak case

jaeyun-jung:
Fix svace issue,
- fix invalid condition to check the args.
- possible res leak in query close function.
- condition to free strings for test models.

### DIFF
--- a/gst/nnstreamer/tensor_if/gsttensorif.c
+++ b/gst/nnstreamer/tensor_if/gsttensorif.c
@@ -354,18 +354,14 @@ gst_tensor_if_set_property_supplied_value (const GValue * value,
 {
   gint i;
   gboolean is_float = FALSE;
-  const gchar *param;
-  gchar **strv;
-  gint num;
+  const gchar *param = g_value_get_string (value);
+  gchar **strv = g_strsplit_set (param, delimiters, -1);
+  gint num = g_strv_length (strv);
 
-  if (!value) {
-    ml_loge ("Invalid supplied value. The value is NULL");
+  if (!param) {
+    ml_loge ("Invalid supplied value. The value is NULL.");
     return;
   }
-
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_if/gsttensorif.c
```diff
@@ -354,19 +354,15 @@ gst_tensor_if_set_property_supplied_value (const GValue * value,
 {
   gint i;
   gboolean is_float = FALSE;
-  const gchar *param;
-  gchar **strv;
-  gint num;
+  const gchar *param = g_value_get_string (value);
+  gchar **strv = g_strsplit_set (param, delimiters, -1);
+  gint num = g_strv_length (strv);
```
dongju-chae, line 359: I'm not sure how this patch fixes resource leak.

jaeyun-jung: leak case when calling nnstreamer_query_close() (not this line)

query element allocates TensorQueryConnection handle and tyies to make connection. If connection is failed, it calls nnstreamer_query_close(). In nnstreamer_query_close(), it should be release TensorQueryConnection handle even if closing socket is failed or protocol is not tcp.

https://github.com/nnstreamer/nnstreamer/blob/fcd180087157e3942049565611fa73e66352e4f3/gst/nnstreamer/tensor_query/tensor_query_common.c#L230

dongju-chae: Ah, thanks for the detail explanation. I missed the change of line 361.

dongju-chae (approved):
LGTM

anyj0527 (approved):
LGTM ✔

again4you (approved):
LGTM!

PR #3501, opened 2021-09-20
[armnn] compilation error - unused parameter

jvuillaumier:
warning correction -- breaks build with -Werror :
Some function parameters are unused in body definition.  Those are explicitly marked with UNUSED() macro, same approach as in other elements.

Tests:
ssat suite: passed
armnn GTest suite: passed

correction for multiple occurences of compilation error:

error: unused parameter '[parameter name]' [-Werror=unused-parameter]


myungjoo (approved):
LGTM. Thanks!

jaeyun-jung (approved):

PR #3502, opened 2021-09-21
[Filter/openvino] fix invalid args

jaeyun-jung:
As mentioned in another PR, openvino subplugin calls accel function with invalid args.


myungjoo (approved):
LGTM

dongju-chae (approved):
LGTM

Issue #3508, opened 2021-09-28
[Suggestion] Robustness support of MQTT-Hybrid 


again4you:
To supports the Robustness of MQTT-Hybrid, we need to define the naming rule of the MQTT topic and its scenario.

## Basic Ideas
* MQTT Broker can maintain List of servers and their latest information such as status, algorithm, and so on.
  - Client can subscribe its updated information
  - Client can requests servers' information explicitly.

* Client can subscribe multiple topics of the same domain from various devices.
  - Use 'wildcards' of MQTT filter
    - \+ : single level wildcard that matches any name
    - \# : multi level wildcard used at the end of the topic

## Topic Rule
* Rule: **edge/[type]/[device_name]/[main_operation]/[option]**
  * edge: prefix (fixed)
  * type: 'inference' for server, client / 'sensor' for small sensor device 
  * device_name: Device Name such as 'galaxy_00' or 'bespoke_01'
  * main_operation: Main Operation such as 'object_detection', 'speed
  * option: optional topic for main operation (e.g. mobilev3)

* Examples
  * edge/inference/galaxy_00/object_detection/mobilev3 (for Server / Client)
  * edge/inference/tv_00/object_detection/yolo (for Server / Client)
  * edge/inference/bespoke_00/object_detection (for Server / Client)
  * edge/sensor/aircleaner_01/temperature (for Sensor)

### 1) Servers connect to Brokers
![robust_mqtt_00](https://user-images.githubusercontent.com/433435/135012310-4210fa46-9185-4017-845f-d51db8581f64.JPG)
  - edge/inference/galaxy_00/object_detection/mobilev3
  - edge/inference/TV_00/object_detection/yolo

### 2) Client connect to Broker
![robust_mqtt_01](https://user-images.githubusercontent.com/433435/135012612-f8e7dd73-844e-4679-8643-7161fbcabf4e.JPG)
  - Subscribe 'edge/inference/+/object_detection/#' topic
    - \+ : single level wildcard that matches any name
    - \# : multi level wildcard used at the end of the topic

### 3) Client gets server's Information
![robust_mqtt_02](https://user-images.githubusercontent.com/433435/135012747-3e66c5e4-6853-4ce1-a34a-91fb54b174c3.JPG)
* List of server nodes, and their algorithm, status or performance

### 4) Choose the suitable one
[...]

PR #3510, opened 2021-09-30
[QUERY] Robustness support of MQTT-hybrid

gichan-jang:
Releated issue: #3508
Robustness support of MQTT-hybrid.
If the connected server is stop, re-connect to another availabe server.

Test pipeline.
Server 1
```
gst-launch-1.0 tensor_query_serversrc operation=passthrough/optionalVal ! other/tensors,format=flexible ! tensor_query_serversink
```
Server 2
```
gst-launch-1.0 tensor_query_serversrc operation=passthrough/optional host=127.0.0.1 port=5001 ! other/tensors,format=flexible ! tensor_query_serversink host=127.0.0.1 port=5000
```
Client
```
gst-launch-1.0 v4l2src ! videoconvert ! videoscale ! video/x-raw,width=320,height=240,format=RGB,framerate=30/1 ! tensor_converter ! other/tensors,format=flexible ! tensor_query_client operation=passthrough! tensor_converter input-dim=3:320:240 input-type=uint8 ! tensor_decoder mode=direct_video ! videoconvert ! ximagesink
```

**Self evaluation:**


tests/nnstreamer_query/runTest.sh
```diff
@@ -47,6 +54,46 @@ callCompareTest raw3_0.log result3_0.log 3-3 "Compare 3-3" 1 0
...
+    echo "mosquitto broker is not running"
+    return 0
+}
+
+# Check whether mqtt broker is running or not
+check_mosquitto
+ret=$?
+if [  $ret -ne 1 ]; then
+    echo "Warning! mosquitto is not running so skip MQTT-hybrid test."
+    report
```
anyj0527, line 73: May add `rm *.log` ?

gichan-jang: Right. Updated.

gst/nnstreamer/registerer/nnstreamer.c
```diff
@@ -67,9 +67,11 @@
 #include <tensor_transform/tensor_transform.h>
 #include <tensor_if/gsttensorif.h>
 #include <tensor_rate/gsttensorrate.h>
+#if (ENABLE_NNS_QUERY == 1)
```
wooksong, line 70: IMO, ```tensor_query/tensor_query_*.h``` can hide this line and the following ```#endif```.

gst/nnstreamer/registerer/nnstreamer.c
```diff
@@ -102,9 +104,11 @@ gst_nnstreamer_init (GstPlugin * plugin)
   NNSTREAMER_INIT (plugin, transform, TRANSFORM);
   NNSTREAMER_INIT (plugin, if, IF);
   NNSTREAMER_INIT (plugin, rate, RATE);
+#if (ENABLE_NNS_QUERY == 1)
```
wooksong, line 107: How about defining a macro something like ```NNSTREAMER_QUERY_INIT``` which is valid only if ```ENABLE_NNS_QUERY == 1```?

tests/nnstreamer_query/runTest.sh
```diff
@@ -47,6 +54,47 @@ callCompareTest raw3_0.log result3_0.log 3-3 "Compare 3-3" 1 0
 callCompareTest raw3_1.log result3_1.log 3-4 "Compare 3-4" 1 0
 callCompareTest raw3_2.log result3_2.log 3-5 "Compare 3-5" 1 0
 
+check_mosquitto() {
+    pid=`ps aux | grep mosquitto | grep -v grep | awk '{print $2 }'`
```
wooksong, line 58:
```suggestion
    pid=`ps aux | grep mosquitto | grep -v grep | awk '{print $2}'`
```

```awk (gawk)```  is awsome!

File: tests/nnstreamer_query/runTest.sh [hunk 5]
```diff
@@ -47,6 +54,47 @@ callCompareTest raw3_0.log result3_0.log 3-3 "Compare 3-3" 1 0
...
+
+# Check whether mqtt broker is running or not
+check_mosquitto
+ret=$?
+if [  $ret -ne 1 ]; then
+    echo "Warning! mosquitto is not running so skip MQTT-hybrid test."
+    rm *.log
+    report
+    exit
+fi
```
wooksong, line 76:
```suggestion

    if [ $pid > 0 ]; then
        echo "mosquitto broker is running"
        return
    fi
    echo "Warning! mosquitto is not running so skip MQTT-hybrid test."
    rm *.log
    report
    exit
}
```
This is just a suggestion. BTW, I think it is possible to make these lines (from line 57 to 76) without ```check_mosquitto()```.

tests/nnstreamer_query/runTest.sh
gichan-jang: Updated.

tests/nnstreamer_query/runTest.sh
gichan-jang: Thanks for the suggestion. updated.

gst/nnstreamer/registerer/nnstreamer.c
gichan-jang: Oh.. Interesting.
But the mqtt-hybrid (using `nnsquery) code will be separated from tensor_query soon, so it will be reverted.

gst/nnstreamer/registerer/nnstreamer.c
gichan-jang: Same above.

jaeyun-jung (approved):
LGTM.
We should separate mqtt-hybrid feature and base connection of tensor-query later :)

PR #3529, opened 2021-10-22
[Converter] convert bytes or flex-tensor without properties

jaeyun-jung:
1. convert bytes (octet stream) to flex-tensor when input info (type and dimension) is not denoted.
2. parse tensor-mata and convert to static when input is flexible tensor.

### DIFF
--- a/gst/nnstreamer/tensor_converter/tensor_converter.c
+++ b/gst/nnstreamer/tensor_converter/tensor_converter.c
@@ -829,15 +829,16 @@ _gst_tensor_converter_chain_timestamp (GstTensorConverter * self,
   self->old_timestamp = GST_BUFFER_TIMESTAMP (inbuf);
 }
 
-/** @brief Chain function's private routine to process multi tensor */
+/** @brief Chain function's private routine to process octet stream */
 static GstBuffer *
-_gst_tensor_converter_chain_multi_tensor (GstTensorConverter * self,
-    GstBuffer * buf)
+_gst_tensor_converter_chain_octet (GstTensorConverter * self, GstBuffer * buf)
 {
   GstBuffer *buffer = buf;
+  GstTensorsInfo *_info = &self->tensors_config.info;
+  gboolean multi = (_info->num_tensors > 1);
 
   /* configure multi tensors */
-  if (self->tensors_info.num_tensors > 1) {
+  if (multi || gst_buffer_n_memory (buf) > 1) {
     GstMemory *mem;
     gsize offset, size;
     guint i;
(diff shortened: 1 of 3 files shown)


dongju-chae (approved):
LGTM

jaeyun-jung:
To reviewers,
I updated this PR to handle flexible tensor without input info. Please review this again.
- case1: octet stream to flexible tensor
- case2: flexible tensor to static tensor

myungjoo:
Question: what happens if we have
```
appsrc ! application/octet-stream ! tensor_converter ! tensor_filter some_model ! ...
```
How does this determine if it's flexible or static at tensor_converter?

jaeyun-jung:
@myungjoo 
If tensor_filter returns input caps other/tensors,format=flexible, converter sets output to flexible tensor.
If filter caps is static, error - not negotiated with this PR.

gst/nnstreamer/tensor_converter/tensor_converter.c
```diff
@@ -1121,36 +1121,49 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
       }
       break;
     case _NNS_OCTET:
-      /** get frame size from the properties */
-      g_assert ((buf_size % frame_size) == 0);
-      frames_in = buf_size / frame_size;
+      if (gst_tensors_config_is_flexible (config)) {
+        /* update dimension with buffer size */
+        config->info.info[0].dimension[0] = buf_size;
```
gichan-jang, line 1126: Is there no problem if the octet-stream has multiple memory chunks?

jaeyun-jung: I updated this PR to handle multi memories in in input buffer.

jaeyun-jung:
2021-11-12 PR updated
- handle multi memories in input buffer (octet stream)
- if no properties given, get tensors info from peer and use this to convert octet stream

myungjoo:
> @myungjoo If tensor_filter returns input caps other/tensors,format=flexible, converter sets output to flexible tensor. If filter caps is static, error - not negotiated with this PR. (PR updated 2021-11-12)

What does conventional tensor_filter return? What if the tensor_filter has set_input_dim instead of get_input/outtput/dim? I want it to work when there is NO capsfilter and there is NO format/dim specified.

Or.. is it considered "format=static" because its default value is static?

myungjoo:
> > @myungjoo If tensor_filter returns input caps other/tensors,format=flexible, converter sets output to flexible tensor. If filter caps is static, error - not negotiated with this PR. (PR updated 2021-11-12)
> 
> What does conventional tensor_filter return? What if the tensor_filter has set_input_dim instead of get_input/outtput/dim? I want it to work when there is NO capsfilter and there is NO format/dim specified.
> 
> Or.. is it considered "format=static" because its default value is static?
> 
> I'll be in the office tomorrow, so we can further discuss this tomorrow.

tensor-filter does not support flexible today.
Let's make it support later (TBD2022)

@jaeyun-jung : please register EPIC item for this.

PR #3533, opened 2021-10-28
[Converter/Doc] doc to convert octet stream

jaeyun-jung:
Update document to convert octet stream to static or flexible tensor.


myungjoo (approved):
LGTM

Issue #3442, opened 2021-08-10
[Tizen] Enable SNPE for Tizen


myungjoo:
We now have a client that wants to use Qualcomm SoC for Tizen devices along with Qualcomm's NPU/DSPs.
This client also wants to use nnstreamer and ROS2 in their Tizen devices for realtime inferences.

1. See if there is no license issues with SNPE for public releases. (e.g., mirroring SNPE repo in Tizen is allowed by Qualcomm?)
2. Port SNPE support (we released only for Android) for Tizen
3. Test nnstreamer-ros for ROS2@Tizen .
4. They may request Edge-AI capability as well.

Timeline: TBD. We do not have official requests from this client. It's at "RFP" stage.

myungjoo:
Please visit bugs.tizen.org and create related git repos (and build requests)

PR #3541, opened 2021-11-01
[Spec] Enable SNPE filter on both In-House and public infra @open sesame 11/11 17:20

again4you:
To support NNStreamer filter for SNPE on both In-House build infra and
those of the public, this patch conditionally enables nnstreamer-snpe
rpm package when existing actual snpe library. (i.e. snpe.pc). In case
of the dummy SNPE package, nnstreamer-snpe rpm contains no files.

**Self evaluation:**

**Related Issue**
* https://github.com/nnstreamer/nnstreamer/issues/3442

#### Public Build Infra ####
* `snpe-devel.rpm` contains the `snpe-dummy.pc` so NNStreamer filter for SNPE is not built.
```bash
$ rpm -qlp snpe-devel-0.0.1-1.aarch64.rpm
/usr/lib64/pkgconfig/snpe-dummy.pc

$ rpm -qlp ./nnstreamer-snpe-2.1.0-0.aarch64.rpm
(contains no files)
```

#### In-House Infra ####
* `snpe-devel.rpm` contains the `snpe.pc` file so NNStreamer filter for SNPE is built on In-Hause Infra.
```bash
# Pre-built packages on the build server
$ rpm -qlp snpe-devel-1.53.2-0.aarch64.rpm | grep snpe.pc
/usr/lib64/pkgconfig/snpe.pc

# Build result
$ find . | grep snpe
./nnstreamer-snpe-2.1.0-0.aarch64.rpm
./nnstreamer-snpe-debuginfo-2.1.0-0.aarch64.rpm

$ rpm -qlp ./nnstreamer-snpe-2.1.0-0.aarch64.rpm
/usr/lib/nnstreamer/filters/libnnstreamer_filter_snpe.so
```

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -34,7 +34,7 @@
 %define		mqtt_support 1
 %define		lua_support 1
 %define		tvm_support 1
-%define		snpe_support 0
+%define		snpe_support 1
 
 %define		check_test 1
 %define		release_test 1
@@ -102,6 +102,7 @@
 %define		lua_support 0
 %define		mqtt_support 0
 %define		tvm_support 0
+%define		snpe_support 0
 %endif
 
 # DA requested to remove unnecessary module builds
@@ -114,18 +115,17 @@
 %define		lua_support 0
 %define		mqtt_support 0
 %define		tvm_support 0
-%endif
-
-# TODO: The variable `_with_qc_snpe` is temporary. Make proper one.
-%if 0%{?_with_qc_snpe}
-%ifarch aarch64
-%define		snpe_support 1
-%endif
+%define		snpe_support 0
 %endif
 
 # Release unit test suite as a subpackage only if check_test is enabled.
 %if !0%{?check_test}
 %define		release_test 0
+%endif
+
+# Current Tizen Robot profile only supports aarch64.
+%ifnarch aarch64
+%define		snpe_support 0
 %endif
 
 # If it is tizen, we can export Tizen API packages.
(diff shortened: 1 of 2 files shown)


packaging/nnstreamer.spec
```diff
@@ -116,18 +117,16 @@
 %define		tvm_support 0
```
myungjoo, line 117: Disable snpe for DA, either.

packaging/nnstreamer.spec
```diff
@@ -735,13 +734,6 @@ Provides additional gstreamer plugins for nnstreamer pipelines
 %define enable_tvm -Dtvm-support=disabled
 %endif
 
-# Support snpe
-%if 0%{?snpe_support}
-%define enable_snpe -Dsnpe-support=enabled
-%else
-%define enable_snpe -Dsnpe-support=disabled
-%endif
-
```
myungjoo, line 744: Are you going to let "auto" decide snpe support?
(not a good idea for platform build because we need to ensure features for platform builds.)
Anyway, it's ok until the snpe-support packages are ready.

packaging/nnstreamer.spec
again4you: Right. I will fix it.

packaging/nnstreamer.spec
again4you: Yes. I agreed with your opinion. It's not a good idea.
However, there is no Robot profile option in OBS so package config file (i.e. pc) for snpe-devel rpm is the only way to check whether SNPE is enabled or not.

As you mentioned above, it is a workaround so I will add the comment for this reason.

anyj0527 (approved):
LGTM👍 Thanks!

myungjoo (approved):

PR #3546, opened 2021-11-03
[Common] util function to handle tensors

jaeyun-jung:
With gstreamer plugins, nnstreamer element may receive gst-buffer with invalid memory chunks.
nnstreamer internally checks the number of tensors using the mem blocks in gst-buffer.

To handle multi tensors, add util function to separate memories in gst-buffer.

### DIFF
--- a/gst/nnstreamer/tensor_common_pipeline.c
+++ b/gst/nnstreamer/tensor_common_pipeline.c
@@ -446,6 +446,121 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
 }
 
 /**
+ * @brief Configure gst-buffer with tensors information.
+ * NNStreamer handles single memory chunk as single tensor.
+ * If incoming buffer has invalid memories, separate it and generate new gst-buffer using tensors information.
+ * Note that this function always takes the ownership of input buffer.
+ * @param in input buffer
+ * @param config tensors config structure
+ * @return Newly allocated buffer. Null if failed. Caller should unref the buffer using gst_buffer_unref().
+ */
+GstBuffer *
+gst_tensor_buffer_from_config (GstBuffer * in, GstTensorsConfig * config)
+{
+  GstBuffer *out = NULL;
+  GstMemory *all = NULL;
+  GstMapInfo map;
+  guint i, num;
+  gsize total, offset;
+  gsize mem_size[NNS_TENSOR_SIZE_LIMIT];
+  gboolean configured = FALSE;
+
+  if (!GST_IS_BUFFER (in)) {
+    nns_loge ("Failed to get tensor buffer, invalid input buffer.");
+    return NULL;
+  }
+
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_common_pipeline.c
```diff
@@ -445,6 +445,107 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
...
+  GstBuffer *out = NULL;
+  GstMemory *all = NULL;
+  GstMapInfo map;
+  guint i, num;
+  gsize total, offset;
+  gsize mem_size[NNS_TENSOR_SIZE_LIMIT];
+  gboolean configured = FALSE;
+
+  if (!GST_IS_BUFFER (in) || !gst_tensors_config_validate (config))
+    goto error;
```
dongju-chae, line 469: how about just returning NULL? `gst_buffer_unref` can be called when `in` is a non-buffer instance.

jaeyun-jung: Thanks, updated.

anyj0527 (commented):
LGTM 👍

dongju-chae (approved):
LGTM

gst/nnstreamer/tensor_common_pipeline.c
myungjoo: We have started writing more elaborated error messages for exceptions.
Please describe the error and how the user should correct the error with this error.

gst/nnstreamer/tensor_common_pipeline.c
```diff
@@ -445,6 +445,112 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
...
+
+  /* configure output buffer */
+  out = gst_buffer_new ();
+  all = gst_buffer_get_all_memory (in);
+  offset = 0;
+
+  for (i = 0; i < num; i++) {
+    /* invalid memory size */
+    if (offset + mem_size[i] > total)
+      goto error;
```
myungjoo, line 528: Here, too.

myungjoo (commented):
Please add "useful" error messages (what's wrong & how to fix it) for all error/exceptions.

tests/common/unittest_common.cc
```diff
@@ -1757,6 +1757,185 @@ TEST (versionControl, getVer01)
...
+      EXPECT_EQ (output[j], expected++);
+
+    gst_memory_unmap (mem, &map);
+  }
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (create static tensor buffer)
```
gichan-jang, line 1810: create flexible tensor buffer?

jaeyun-jung: updated

myungjoo (approved):

PR #3547, opened 2021-11-03
[Elements] handle single mem chunk

jaeyun-jung:
If incoming buffer has single memory (case with gst-plugins), nnstreamer elements cannot generate output buffer (invalid memory chunks).

### DIFF
--- a/gst/nnstreamer/tensor_sparse/tensor_sparse_dec.c
+++ b/gst/nnstreamer/tensor_sparse/tensor_sparse_dec.c
@@ -161,6 +161,7 @@ gst_tensor_sparse_dec_init (GstTensorSparseDec * self)
 
   /* init properties */
   self->silent = DEFAULT_SILENT;
+  gst_tensors_config_init (&self->in_config);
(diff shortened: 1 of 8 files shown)


gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_sparse/tensor_sparse_dec.c
```diff
@@ -321,25 +326,21 @@ gst_tensor_sparse_dec_sink_event (GstPad * pad, GstObject * parent,
       gst_event_parse_caps (event, &caps);
       silent_debug_caps (self, caps, "caps");
 
+      /* set in_config */
+      structure = gst_caps_get_structure (caps, 0);
```
myungjoo, line 330: In the future, with later PR, do NOT assume that 0th structure will be the only structure of gst-caps.
In general, there could be multiple structures in gst-cap and you may need to iterate.

myungjoo (approved):
LGTM. Please take a look at the comment anyway.

PR #3550, opened 2021-11-03
[Build] Add g_memdup2() support for glib >= 2.68

jvuillaumier:
Function g_memdup() is deprecated starting from glib 2.68, replaced by g_memdup2() :
_" Use g_memdup2() instead, as it accepts a #gsize argument for byte_size, avoiding the possibility of overflow in a #gsize → #guint conversion."_
https://docs.gtk.org/glib/func.memdup.html

To keep compatibility with the 2 g_memdupX() variants,  option used is to detect glib version from meson script.
Alternative option may be to use in the source code construct like:
`
#ifdef GLIB_CHECK_VERSION(2, 68, 0)

   data = g_memdup2(...);

#else

  data = g_memdup(...);

#endif
`

Self Evaluation:
SSAT tests: passed

Update to fix build with glib version 2.68 and later

Legacy function:
  gpointer g_memdup ( gconstpointer mem, guint byte_size )
is deprecated since version 2.68 and replaced by:
  gpointer g_memdup2 ( gconstpointer mem, gsize byte_size )

Difference is byte_size argument type changed from guint to gsize.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -128,6 +128,9 @@ add_project_arguments('-DNNSTREAMER_CONF_FILE="' + join_paths(nnstreamer_inidir,
 
 # Dependencies
 glib_dep = dependency('glib-2.0')
+if glib_dep.version().version_compare('>= 2.68.0')
+  add_project_arguments('-DGLIB_USE_G_MEMDUP2', language: ['c', 'cpp'])
+endif
 gobject_dep = dependency('gobject-2.0')
 gmodule_dep = dependency('gmodule-2.0')
 gio_dep = dependency('gio-2.0')
--- a/ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
+++ b/ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc
@@ -16,6 +16,7 @@
 
 #include <nnstreamer_log.h>
 #include <nnstreamer_plugin_api.h>
+#include <nnstreamer_util.h>
 
 #include <thread>
 
@@ -119,7 +120,7 @@ ServiceImplFlatbuf::_get_buffer_from_tensors (Message<Tensors> &msg,
     const Tensor * tensor = tensors->tensor ()->Get (i);
     const void * data = tensor->data ()->data ();
     gsize size = VectorLength (tensor->data ());
-    gpointer new_data = g_memdup (data, size);
+    gpointer new_data = _g_memdup (data, size);
 
     memory = gst_memory_new_wrapped ((GstMemoryFlags) 0, new_data, size,
         0, size, new_data, g_free);
--- a/ext/nnstreamer/extra/nnstreamer_grpc_protobuf.cc
+++ b/ext/nnstreamer/extra/nnstreamer_grpc_protobuf.cc
@@ -16,6 +16,7 @@
 
 #include <nnstreamer_log.h>
 #include <nnstreamer_plugin_api.h>
+#include <nnstreamer_util.h>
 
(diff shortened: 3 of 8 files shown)


myungjoo (changes requested):
We can define a macro
```
#if GLIB_USE_G_MEMDUP2
#define _g_memdup(data, size) g_memdup2 (data, size)
#else
#define _g_memdup(data, size) g_memdup (data, size)
#endif
```
in a common header (e.g., nnstreamer_pluigin_api.h).
Let's have minimum #if/#endif in .c/.cc files.

jvuillaumier:
> We can define a macro
> 
> ```
> #if GLIB_USE_G_MEMDUP2
> #define _g_memdup(data, size) g_memdup2 (data, size)
> #else
> #define _g_memdup(data, size) g_memdup (data, size)
> #endif
> ```
> 
> in a common header (e.g., nnstreamer_pluigin_api.h). Let's have minimum #if/#endif in .c/.cc files.

Changeset updated accordingly:
_g_memdup() macro defined in nnstreamer_util.h
_g_memdup() macro used as a replacement of g_memdup() / g_memdup2()

meson.build
```diff
@@ -127,7 +127,12 @@ nnstreamer_conf.set('TRIX_ENGINE_ALIAS', get_option('trix-engine-alias'))
 add_project_arguments('-DNNSTREAMER_CONF_FILE="' + join_paths(nnstreamer_inidir, 'nnstreamer.ini') + '"', language: 'c')
 
 # Dependencies
-glib_dep = dependency('glib-2.0')
+glib_dep = dependency('glib-2.0', version: '>= 2.68.0', required: false)
+if (glib_dep.found())
+  add_project_arguments('-DGLIB_USE_G_MEMDUP2', language: ['c', 'cpp'])
+else
+  glib_dep = dependency('glib-2.0')
```
jaeyun-jung, line 134: how about using macro in header file or version_compare() in meson

Use glib macro in header
```
#if GLIB_CHECK_VERSION (2, 68, 0)
#define _g_memdup g_memdup2
#else
#define _g_memdup g_memdup
#endif
```

or meson.build
```
if glib_dep.version().version_compare('>= 2.68.0')
  add_project_arguments('-DGLIB_USE_G_MEMDUP2', language: ['c', 'cpp'])
endif
```

IMO, 1st one is better.

jvuillaumier: I am fine with dropping meson.build change and go for  GLIB_CHECK_VERSION() check in the header file if all agree with that.

However doing so requires addition of
`
#include <glib.h>
`
at the top of the header file so that the macro is guaranteed to be defined.
May not be a bid deal but my understanding is that at the moment .c/.cc files include glib.h explicitly, not indirectly via other header files.

myungjoo: The first approach (no changes in meson) looks more beautiful, but we cannot ensure the order of ```#include``` statements; the coding-style checks (gst-indent, clang-format, ...) often force to reorder them and having such implicit conditions don't feel right.

Let's proceed with the second approach (adding a few lines in meson.build).

myungjoo (approved):
LGTM

meson.build
jvuillaumier: Patchset updated with suggestion from @jaeyun-jung for option 2 in meson.build:

`
if glib_dep.version().version_compare('>= 2.68.0')
  add_project_arguments('-DGLIB_USE_G_MEMDUP2', language: ['c', 'cpp'])
endif
`

jaeyun-jung (approved):

PR #3551, opened 2021-11-04
[TEST] Disable MQTT Test when broker is not available.

gichan-jang:
If the broker is not available, the mqtt element cannot be changed to the playing state.
However, harnesses are used in the mqtt test, and errors occur if harnesses cannot be changed to the playing state.
So, if there's no broker on the test environment, the test fails.

For example, the build on the lunch pad failed.
https://launchpadlibrarian.net/566158157/buildlog_ubuntu-bionic-amd64.nnstreamer_2.1.0.0-0~202110290743~ubuntu18.04.1_BUILDING.txt.gz

**Self evaluation:**

### DIFF
--- a/tests/check_broker.sh
+++ b/tests/check_broker.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+mosquitto_pub -r -t check/broker/running -m "mosquitto_broker_is_running"
+check=`mosquitto_sub  -t check/broker/running --remove-retained -W 1 | grep mosquitto_broker_is_running`
+
+if [ "$check" = "mosquitto_broker_is_running" ]; then
+    echo 1
+fi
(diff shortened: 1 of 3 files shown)


anyj0527 (approved):
LGTM ✔

tests/check_broker.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+pid=`ps aux | grep mosquitto | grep -v grep | awk '{print $2}'`
```
dongju-chae, line 2: I'm not sure it's enough to check broker's availability. What if its homonym exists..
How about utilizing `mosquitto_pub` and `mosquitto_sub` to ensure the broker's basic functionality?

gichan-jang: Nice suggestion! Updated. Please check it again.

dongju-chae (approved):
LGTM!

PR #3564, opened 2021-11-15
[Query] null ptr in finalize callback

jaeyun-jung:
Clearly check the handles in finalize callback.
(e.g., gst-inspect shows gst-log if given handle is null.)


gichan-jang (approved):
LGTM

myungjoo (approved):

PR #3571, opened 2021-11-19
Test: enhancing log behaviors for unittest

myungjoo:
1. Add a dlog_print hijacker that makes dlog calls independent
    from dlog daemon and drivers. As a result, developers can now
    read the dlog log messages of unittests in gbs environment.
    
    2. Clean up a bit of meson.build dependency declaration of
    unit tests.
    
    3. A unit test without direct dlog call is updated to call
    it directly so that the hijacker is linked (edgetpu).


again4you (approved):
LGTM!

dongju-chae (approved):
LGTM

PR #3577, opened 2021-11-26
tensor_filter TFLite subplugin - XNNPACK delegate @open sesame 11/29 12:40

jvuillaumier:
---

2 items related to XNNPACK delegate usage with TfLite framework (tested on TfLite 2.6):

1) Peculiarity of TfLite XNNPACK delegate usage in TfLite is that both input and output tensors buffer addresses are cached within XNNPACK runtime during first invoke(). NNStreamer tensor_filter input/output buffer memory addresses change from an inference to an other, therefore i/o buffer data has to be copied between Gst memory and TfLite allocated memory for tensors. Failure to do so typically leads to system crash as TfLite interpreter tries to access during subsequent invoke() to the GStreamer buffers populated during 1st invoke that have chances to have been deallocated and unmapped.
Currently only output tensors data copy is provisioned for XNNPACK case.

2) When TfLite library is built with XNNPACK delegate support, XNNPACK delegate is used as a default delegate for the graph, even though not explicitly applied via ModifyGraphWithDelegate().
Issue is that XNNPACK usage requires tensor data copies so for NNStreamer case, XNNPACK should only be used when explicitly selected by the user (custom=Delegate:XNNPACK) so that TfLite subplugin is aware about XNNPACK usage and handle input/output tensors accordingly.
API for OpResolver has been introduced in TfLite 2.4 for this:
tflite::ops::builtin::BuiltinOpResolverWithoutDefaultDelegates resolver;
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/kernels/register.h#L36

**Self evaluation:**
SSAT test suite passed

**How to evaluate:**
1. need TFLite library built with XNNPACK support 
2. enable NNStreamer meson build option -Dtflite2-xnnpack-delegate-support=true
3. use tensor_filter property custom=Delegate:XNNPACK

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -238,6 +238,10 @@ if tflite2_support_is_available
     tflite2_compile_args += '-DTFLITE_EXTERNAL_DELEGATE_SUPPORTED'
   endif
 
+  if get_option('tflite2-xnnpack-delegate-support')
+    tflite2_compile_args += '-DTFLITE_XNNPACK_DELEGATE_SUPPORTED'
+  endif
+
   tflite2_extra_dep = declare_dependency(
     compile_args : tflite2_compile_args
   )
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -333,20 +333,28 @@ TFLiteInterpreter::invoke (const GstTensorMemory *input, GstTensorMemory *output
   start_time = g_get_monotonic_time ();
 
   /**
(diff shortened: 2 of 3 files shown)


jvuillaumier:
Patchset updated to fallback using (TfLiteTensor*)->data.raw (deprecated since TFLite 2.2) instead of more recent (TfLiteTensor*)->data.data as it is breaking some builds.

anyj0527 (approved):
Great thanks for this contribution. I've packaged tflite v2.6.0+ for android and tizen without default delegate feature to address this default delegate issue.

But using `tflite::ops::builtin::BuiltinOpResolverWithoutDefaultDelegates resolver` seems more elegant considering those who manually build nnstreamer with their own tflite support.

LGTM✔

jaeyun-jung (approved):

PR #3586, opened 2021-12-01
[Query] fix invalid socket mode

jaeyun-jung:
Fix socket op mode, blocking IO is unnecessary with multi-threaded server node.


gichan-jang:
The blocking mode option is no longer needed as the server operates in multi-thread mode.
How about just removing it?

again4you (approved):
LGTM!

gichan-jang (approved):
LGTM

PR #3590, opened 2021-12-02
[Query] set default timeout

jaeyun-jung:
Define common timeout for tensor-query.

TODO: we should handle timeout in query-server elements.


myungjoo (approved):
LGTM

gichan-jang (approved):
LGTM

Issue #3591, opened 2021-12-03
Question: rebuilding nnstreamer from source w/ pytorch-1.10.0


byungs2:
# Issue Description
Hi dev.
Now I'm trying to re-build nnstreamer from source with libtorch-1.10.0.
But I met errors while ninja -C build as below.

```
/home/workspace/nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc: In member function 'int TorchCore::invoke(const GstTensorMemory*, GstTensorMemory*)':
/home/workspace/nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc:486:27: error: 'struct c10::IValue' has no member named 'isGenericList'; did you mean 'isGenericDict'?
   } else if (output_value.isGenericList ()) {
                           ^~~~~~~~~~~~~
                           isGenericDict
/home/workspace/nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc:487:66: error: 'struct c10::IValue' has no member named 'toGenericListRef'; did you mean 'toGenericDict'?
     c10::ArrayRef<torch::jit::IValue> output_list = output_value.toGenericListRef ();
                                                                  ^~~~~~~~~~~~~~~~
                                                                  toGenericDict
cc1plus: all warnings being treated as errors
[199/210] Compiling C++ object 'tests/nnstreamer_filter_extensions_common/b8951fb@@unittest_tizen_pytorch@exe/gtest-all.cc.o'.
ninja: build stopped: subcommand failed.
```

and there have a lots of warnning about -Werror=redundant-decls & -Werror=unused-parameter too.
If you need information about those errors, I will add.

I think It occurs from my mistake because I'm not familiar with compile process.

Anyway I hope It is not a question that is not worth answering.

Expected Result
============
complete re-build nnstreamer from source w/ pytorch 1.10.0

How to Reproduce
===============
1. use Dockerfile as below

```
FROM ubuntu:18.04

ARG DEBIAN_FRONTEND=noninteractive
RUN apt update
RUN apt install -y software-properties-common
RUN apt-add-repository ppa:nnstreamer
RUN apt update

RUN apt install -y git vim wget

RUN apt install -y \
    meson \
    ninja-build \
    build-essential \
    pkg-config \
    cmake \
    libglib2.0-dev \
    libgstreamer1.0-0 \
    gstreamer1.0-plugins-base \
    gstreamer1.0-plugins-good \
[...]

PR #3592, opened 2021-12-06
meson: use version_copmare()

myungjoo:
Do not compare version strings with general operators.
Use version_compare() API instead.

Reported by #3591


jaeyun-jung (approved):

anyj0527 (approved):
LGTM👍

PR #3595, opened 2021-12-07
[Query] Change error message when GstMeta is invalid.

gichan-jang:
Some GStreamer element don't copy GstMeta. So tensor query cannot handle buffer.
Change error message when GstMeta is invalid and stop pipeline.
The available elements table is not completed yet and will change frequently, so I will write it on the wiki page first.

*note: query server src does not set format (GST_FORMAT_TIME) currently, compositor also is not availabe.  This will also be changed soon.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_serversink.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_serversink.c
@@ -23,6 +23,7 @@ GST_DEBUG_CATEGORY_STATIC (gst_tensor_query_serversink_debug);
 #define DEFAULT_HOST "localhost"
 #define DEFAULT_PORT_SINK 3000
 #define DEFAULT_PROTOCOL _TENSOR_QUERY_PROTOCOL_TCP
+#define DEFAULT_METALESS_FRAME_LIMIT 1
 
 /**
  * @brief the capabilities of the inputs.
@@ -39,7 +40,8 @@ enum
   PROP_PORT,
   PROP_PROTOCOL,
   PROP_ID,
-  PROP_TIMEOUT
+  PROP_TIMEOUT,
+  PROP_METALESS_FRAME_LIMIT
 };
 
 #define gst_tensor_query_serversink_parent_class parent_class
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_query/tensor_query_serversink.c
```diff
@@ -302,7 +302,10 @@ gst_tensor_query_serversink_render (GstBaseSink * bsink, GstBuffer * buf)
 
   meta_query = gst_buffer_get_meta_query (buf);
   if (!meta_query) {
-    nns_logw ("Cannot get tensor query meta. Drop buffers!");
+    nns_loge ("Cannot get tensor query meta. Drop buffers!\n"
+        "There are elements that are not available on the query server.\n"
+        "Please check available elements on the server."
+        "See: https://github.com/nnstreamer/nnstreamer/wiki/Available-elements-on-query-server");
     return GST_FLOW_OK;
```
myungjoo, line 309: If meta is not available, you cannot reply to clients, right?

Then,... let's generate GST_FLOW_${error} if it fails to get meta repeatedly without having a frame with meta.

E.g., 
```
static int metalessFrameCount = 0;
...
if (meta) {
   process normally;
   metalessFrameCount = 0;
} else {
   write error message;
   metalessFramecount ++;
   
   if (metalessFrameCount > limit) {
      write fatal error message;
      return GST_FLOW_${error};
}
...
return GST_FLOW_OK;
```

gichan-jang: > If meta is not available, you cannot reply to clients, right?

Right. If the client don't receive the reply within timeout, it will stop the playing.
I will change to stop the server if there is no meta in the buffer received from all clients.
Thanks.

myungjoo: Yes. what we need to do is:

notify the server pipeline writer that he/she made a mistake in GSTMETA handling (an element that cannot handle GSTMETA is in the pipeline) as soon as possible and as clear as possible.

myungjoo: And... this ```limit``` might need to be a property of tensor-query-server-sink with some reasonable default value.

gichan-jang: I changed it to compare using the number of connected client instead of `limit`.
Because, If the client doesn't receive the buffer from the server, It doesn't send the buffer to the server again.

myungjoo: No. don't make it #clients.
What if #clients = 100 and only 1 client has actually sent a query?
In most cases, server pipeline writers will want to set limit = 1 unless the pipeline is supposed to create a few non-Meta buffers (to be wasted) along with with-Meta buffers from a single query input. (e.g., for each query-src input frame, three sink output frames are created and only one of the three has a valid Meta.)

gichan-jang: Ok. Default `metaless_frame_limit` value is set to 1 for general case.

gst/nnstreamer/tensor_common_pipeline.c
```diff
@@ -164,7 +164,7 @@ gst_tensor_time_sync_get_current_time (GstCollectPads * collect,
           /* fall-through */
         case SYNC_SLOWEST:
         case SYNC_REFRESH:
-          if (*current_time < GST_BUFFER_PTS (buf))
+          if (*current_time <= GST_BUFFER_PTS (buf))
```
dongju-chae, line 167: Q. Is it related to this PR?

myungjoo (approved):
LGTM

gst/nnstreamer/tensor_common_pipeline.c
gichan-jang: Not directly related the purpose of the this PR. I will make a new PR for this. Thanks.

dongju-chae (approved):
LGTM

PR #3600, opened 2021-12-09
[Query] fix invalid data size

jaeyun-jung:
1. Fix invalid data size - caps string length.
2. Clearly define client and server param and update log message.


myungjoo (approved):

gichan-jang (approved):
LGTM!

Issue #3603, opened 2021-12-10
ml_single_open returns Streams Pipe Error


almirfilho77:
# Issue Description

I get Streams Pipe Error (error code -86) when I try to load the model [yamnet/classification/tflite](https://tfhub.dev/google/lite-model/yamnet/classification/tflite/1) using the following function:

`int status = ml_single_open(&single, TEST_MODEL_PATH, NULL, NULL, ML_NNFW_TYPE_TENSORFLOW_LITE, ML_NNFW_HW_ANY);`

When I use dlog I get the following messages:

I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
I/nnstreamer( 1031): Found tensorflow2-lite for tensorflow-lite.
E/nnstreamer( 1031): Failed to construct interpreter
E/nnstreamer( 1031): Failed to load model

Further Information
===============

When I use the same function to load [the yamnet/tflite model](https://tfhub.dev/google/lite-model/yamnet/tflite/1) or [the mobilenet_v1_1.0_224_quant model](https://github.com/nnstreamer/nnstreamer-example/tree/main/Tizen.native/SingleSample/res) it works fine.

* I'm working on Tizen 6.5
* Code is in C++ 
* nnstreamer 2.0 package
* tensorflow2-lite 2.3 package

EDIT:

I tried to use tensorflow APIs directly as follows:

```
std::unique_ptr<tflite::FlatBufferModel> model = tflite::FlatBufferModel::BuildFromFile(TEST_MODEL);

if (!model)
{
    _ERR("Failed to mmap model");
}

tflite::ops::builtin::BuiltinOpResolver resolver;
std::unique_ptr<tflite::Interpreter> interpreter;
tflite::InterpreterBuilder(*model.get(), resolver)(&interpreter);

if (!interpreter)
{
    _ERR("Failed to create interpreter");
}
```

With dlog I can see the message "Failed to create interpreter", also only for the problematic model.

anyj0527:
[...]

Issue #3475, opened 2021-09-02
[EPIC][VoC] Upgrade error messages (esp. ML-APIs)


myungjoo:
This is requested by clients.

- Need human-readable and more descriptive error messages for ML-API exceptions.
    - Error messages should have enough info to debug the pipelines.
- This includes Web ML API users

PR #3605, opened 2021-12-13
filter/tensorflow-lite: loadModel error message

myungjoo:
For developers, we need to elaborate error messages.
The error messages should have enough information to
understand what's wrong and what to fix.

We will be updating all error messages of nnstreamer.

This helps #3603
This is part of #3475

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -948,12 +948,13 @@ TFLiteCore::init (tflite_option_s *option)
   interpreter->setModelPath (option->model_file);
   interpreter->setExtDelegate (option->ext_delegate_path, option->ext_delegate_kv_table);
   num_threads = option->num_threads;
+  int err;
 
   setAccelerator (option->accelerators, option->delegate);
   g_message ("accl = %s", get_accl_hw_str (accelerator));
 
-  if (loadModel ()) {
-    ml_loge ("Failed to load model\n");
+  if ((err = loadModel ())) {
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -948,12 +948,13 @@ TFLiteCore::init (tflite_option_s *option)
...
   num_threads = option->num_threads;
+  int err;
 
   setAccelerator (option->accelerators, option->delegate);
   g_message ("accl = %s", get_accl_hw_str (accelerator));
 
-  if (loadModel ()) {
-    ml_loge ("Failed to load model\n");
+  if ((err = loadModel ())) {
+    ml_loge ("Failed to load model (TensorFlow-lite interpreter->loadModel() has returned %d. Please check if the model, '%s', is accessible and compatible with the given TensorFlow-lite instance. For example, this TensorFlow-lite's version might not support the given model.\n", err, option->model_file);
```
jaeyun-jung, line 957: how about adding current tf-lite version using predefined values?
(TFLITE_VERSION_MAJOR / MINOR / MICRO)

myungjoo: TFLITE_VERSION_* is determined at build time.
However, users CAN replace it at runtime if we allow to use tf-lite shared library.
Thus, using TFLITE_VERSION_* macro is NOT safe. The version should be determined at run-time.

dongju-chae (approved):
LGTM

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
dongju-chae: I think [TfLiteVersion()](
https://github.com/tensorflow/tensorflow/blob/2afc306a29ce217244784dde30d2171fac46e482/tensorflow/lite/c/c_api.h#L96) can be used but I'm not sure our tflite library supports this.

PR #3606, opened 2021-12-13
[Query] client id in connection handle

jaeyun-jung:
1. Add new function to set client ID.
2. Remove unnecessary value - protocol in cmd data. The connection handle includes this value.


myungjoo (approved):

dongju-chae (approved):
LGTM

PR #3617, opened 2021-12-16
[Filter/TFLite] Fix delegates deleter usage @open sesame 12/17 16:37

jvuillaumier:
Hello,
This change is a proposal to fix a crash visible on TFLite filter using external delegate - same issue is also present for other delegates except NNAPI that uses appropriate unique_ptr declaration.
XNNPACK, NNAPI and external delegate deletion has been verified. Could not verify GPU delegate as it is not available on my setup.
Thank you

On pipeline stop, crash in TFLite filter may be seen with some
delegates with log signature:

Setting pipeline to NULL ...
free(): invalid pointer

Pipeline termination procedure is then aborted.

TFlite delegates shall be released by specific deleter, provided
by their respective interface.
TFLite filter implementation manages delegates instance with
smart pointers for automatic disposal when interpreter goes out of
of scope.

Care should be taken to bind relevant delegate deleter to the
smart pointer, otherwise trivial deleter is used for the managed
pointer which may crash.

---

**Self evaluation:**

SSAT test suite passed.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -198,9 +198,15 @@ class TFLiteInterpreter
   int cacheInOutTensorPtr ();
 
   /** @brief set delegate for the tflite interpreter */
-  void setDelegate (TfLiteDelegate *delegate)
+  void setDelegate (TfLiteDelegate *delegate, void (*deleter) (TfLiteDelegate *))
   {
-    delegate_ = delegate;
+    delegate_ptr = tflite::Interpreter::TfLiteDelegatePtr (delegate, deleter);
+  }
+
+  /** @brief get delegate for the tflite interpreter */
+  TfLiteDelegate *getDelegate ()
+  {
+    return delegate_ptr.get ();
   }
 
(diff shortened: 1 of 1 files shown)


anyj0527 (approved):
Great thanks! I tested the supported delegates including GPU delegate on Android device.
LGTM ✔

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -197,10 +197,16 @@ class TFLiteInterpreter
   /** @brief cache input and output tensor ptr before invoke */
   int cacheInOutTensorPtr ();
 
-  /** @brief set delegate for the tflite interpreter */
-  void setDelegate (TfLiteDelegate *delegate)
+  /** @brief get delegate for the tflite interpreter */
```
anyj0527, line 200: May be a minor typo here.

jvuillaumier: Thank you for the comment - typo fixed in updated patchset

myungjoo (approved):
LGTM

jaeyun-jung (approved):

Issue #3275, opened 2021-05-14
i586/x86_64 unittest failure


dongju-chae:
- Build env: Tizen i586 on gbs
- Test log: http://nnstreamer.mooo.com/nnstreamer/ci/repo-workers/pr-checker/3270-202105131848230.84696292877197-cc1ba12d5efcdd0e0864df73c29f4ae57dc70712/report/build_log_3270_tizen_i586_output.txt

```
[  217s] [----------] 3 tests from cppFilterObj
[  217s] [ RUN      ] cppFilterObj.base01_n
[  217s] C++ custom filter basic_so_03 is not found in ./tests/cpp_methods/libcppfilter_test.so.
[  217s] [       OK ] cppFilterObj.base01_n (26 ms)
[  217s] [ RUN      ] cppFilterObj.base02_n
[  217s] [       OK ] cppFilterObj.base02_n (26 ms)
[  217s] [ RUN      ] cppFilterObj.base03
[  218s] 240, 3837
[  218s] /home/abuild/rpmbuild/BUILD/nnstreamer-1.7.1/tests/cpp_methods/unittest_cpp_methods.cc:284: Failure
[  218s]       Expected: filter_basic::resultCompare (tmp2, tmp3)
[  218s]       Which is: -2
[  218s] To be equal to: 0
[  218s] /home/abuild/rpmbuild/BUILD/nnstreamer-1.7.1/tests/cpp_methods/unittest_cpp_methods.cc:285: Failure
[  218s]       Expected: filter_basic2::resultCompare (tmp5, tmp4)
[  218s]       Which is: -5
[  218s] To be equal to: 0
[  218s] [  FAILED  ] cppFilterObj.base03 (677 ms)
[  218s] [----------] 3 tests from cppFilterObj (729 ms total)
```

myungjoo:
any news? replication? analysis?

dongju-chae:
Well, I couldn't reproduce this error anymore :( let's close and re-open this later if happens again.

dongju-chae:
I still got unittest error in x86_64 tizen build

gichan-jang:
Hmm.. Should we wait until the file is saved to the expected size?
for example, use `_wait_pipeline_save_files` .

myungjoo:
> Hmm.. Should we wait until the file is saved to the expected size? for example, use `_wait_pipeline_save_files` .

Worth trying..

gichan-jang:
Fixed. Close this issue.

PR #3621, opened 2021-12-23
[TEST/util] Change getTempFilename function

gichan-jang:
When creating a temp file name, it is intended to create a unique name, but duplicated temp names may occasionally generated because we remove generated file and get the temp name only.
So sometimes duplicated temp file names are generated when creating temp filenames multiple times.

Change the file to delete after it is used.
Previously, if 1,000 tests were performed, it failed about 15 times on my local PC.
There was no failure when I repeated 2,000 times with this change.

Related issue: #3275 

For example, generate tmp filename 5 times, tmp2 and tmp3 is duplicated.

[  200s] ** (unittest_cppfilter:132252): CRITICAL **: 05:07:52.416: tmp1: /tmp/nnstreamer_unittest_temp_UJOPE1
[  200s] ** (unittest_cppfilter:132252): CRITICAL **: 05:07:52.416: tmp2: /tmp/**nnstreamer_unittest_temp_VIOPE1**
[  200s] ** (unittest_cppfilter:132252): CRITICAL **: 05:07:52.416: tmp3: /tmp/**nnstreamer_unittest_temp_VIOPE1**
[  200s] ** (unittest_cppfilter:132252): CRITICAL **: 05:07:52.416: tmp4: /tmp/nnstreamer_unittest_temp_AJOPE1
[  200s] ** (unittest_cppfilter:132252): CRITICAL **: 05:07:52.416: tmp5: /tmp/nnstreamer_unittest_temp_ILOPE1
 
Self evaluation:

### DIFF
--- a/tests/unittest_util.c
+++ b/tests/unittest_util.c
@@ -42,7 +42,7 @@ setPipelineStateSync (GstElement * pipeline, GstState state,
 
 /**
  * @brief Get temp file name.
- * @return file name (should free string with g_free)
+ * @return file name (should finalize it with g_remove() and g_free() after use)
  */
 gchar *
 getTempFilename (void)
@@ -66,9 +66,6 @@ getTempFilename (void)
   }
 
   g_close (fd, NULL);
-  if (g_remove (tmp_fn) != 0) {
-    _print_log ("failed to remove temp file %s", tmp_fn);
-  }
 
   return tmp_fn;
 }
--- a/tests/cpp_methods/unittest_cpp_methods.cc
+++ b/tests/cpp_methods/unittest_cpp_methods.cc
@@ -106,6 +106,8 @@ TEST (cppFilterOnDemand, pipeline01)
   }
   g_free (str_pipeline);
 
+  g_remove (tmp1);
+  g_remove (tmp2);
   g_free (tmp1);
   g_free (tmp2);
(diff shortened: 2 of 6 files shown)


dongju-chae (approved):
LGTM

anyj0527 (approved):
LGTM, Great👍

tests/unittest_util.c
```diff
@@ -66,9 +66,6 @@ getTempFilename (void)
   }
 
   g_close (fd, NULL);
-  if (g_remove (tmp_fn) != 0) {
-    _print_log ("failed to remove temp file %s", tmp_fn);
-  }
```
again4you, line 69: The users expect that `getTempFilename()` function returns a string without creating the file.
So if you want to remove this line, what about adding some comments as below?
This function is used for test so it looks OK.

```cpp
* @return file name (should finalize it with g_remove() and g_free() after use)
```

again4you (commented):
First of all, nice catch! It looks really hard to find it.
I just adds a simple comment for your work.

tests/unittest_util.c
gichan-jang: Thanks for the suggestion. Updated.

again4you (approved):

PR #3626, opened 2022-01-03
[PKG] Add omitted header to internal dev package

again4you:
nnstreamer_conf.h file is not included in any dev package. To make
some test cases for product-specific ini file, some function in this header
is necessary. This patch adds nnstreamer_conf.h file to
nnstreamer-devel-internal package.

**Self evaluation:**

### DIFF
--- a/debian/nnstreamer-dev-internal.install
+++ b/debian/nnstreamer-dev-internal.install
@@ -1,4 +1,5 @@
 /usr/include/nnstreamer/nnstreamer_internal.h
 /usr/include/nnstreamer/nnstreamer_log.h
 /usr/include/nnstreamer/tensor_filter_single.h
+/usr/include/nnstreamer/nnstreamer_conf.h
 /usr/lib/*/pkgconfig/nnstreamer-internal.pc
--- a/gst/nnstreamer/meson.build
+++ b/gst/nnstreamer/meson.build
@@ -82,6 +82,7 @@ subdir('include')
 # Private header for sub-plugins and native APIs
 nnstreamer_headers += join_paths(meson.current_source_dir(), 'nnstreamer_internal.h')
 nnstreamer_headers += join_paths(meson.current_source_dir(), 'nnstreamer_log.h')
+nnstreamer_headers += join_paths(meson.current_source_dir(), 'nnstreamer_conf.h')
 
 # Build libraries ("both_libraries" are supported from 0.46.)
 nnstreamer_shared = shared_library('nnstreamer',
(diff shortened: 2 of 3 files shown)


jaeyun-jung:
Need discussion, nnstreamer_conf is internal header and marked not to include dev pkg now.

again4you:
> Need discussion, nnstreamer_conf is internal header and marked not to be included in dev pkg now.

OK. Actually, I only need `nnsconf_loadconf ()` function for testing. Except for this issue, I have a question.
If nnstreamer configuration (i.e. /etc/nnstreamer.ini) is changed, can application developers reload it using ML API?
If not, then do we have to make a new API for this?

jaeyun-jung:
> > Need discussion, nnstreamer_conf is internal header and marked not to be included in dev pkg now.
> 
> OK. Actually, I only need `nnsconf_loadconf ()` function for testing. Except for this issue, I have a question. If nnstreamer configuration (i.e. /etc/nnstreamer.ini) is changed, can application developers reload it using ML API? If not, then do we have to make a new API for this?

Now impossible. Does prod-team require this feature? (reload product config file)
I think that, loading conf from default conf and product-conf files is done when starting nnstreamer-based application. So this case - changing ini while running application - is not general app scenario.

again4you:
> > > Need discussion, nnstreamer_conf is internal header and marked not to be included in dev pkg now.
> > 
> > 
> > OK. Actually, I only need `nnsconf_loadconf ()` function for testing. Except for this issue, I have a question. If nnstreamer configuration (i.e. /etc/nnstreamer.ini) is changed, can application developers reload it using ML API? If not, then do we have to make a new API for this?
> 
> Now impossible. Does prod-team require this feature? (reload product config file) I think that, loading conf from default conf and product-conf files is done when starting nnstreamer-based application. So this case - changing ini while running application - is not general app scenario.

VD guys told me that:
1) They want to install their filter such as 'tvdpbsrc'.
2) They want to have their configuration file for updating it whenever they want. For example, restricted_element is updated after installing their package.

In my opinion, we need to talk about this issue face to face.

myungjoo:
Another possible interface for Tizen vendors: capi-machine-learning-tizen-internal-devel (api.git) with https://github.com/nnstreamer/api/blob/main/c/include/nnstreamer-tizen-internal.h

You may add an interface at nnstreamer_util.h , too.


ANYWAY... DO THEY REALLY NEED TO RELOAD DURING PROCESS EXECUTION?
(if they update .ini file and launch the process, they don't need this.)

PR #3634, opened 2022-01-05
[Filter] set default fw in common

jaeyun-jung:
Fix mem leak, init default framework in filter-common function.


gichan-jang (approved):
LGTM

anyj0527 (approved):
LGTM 👍

PR #3650, opened 2022-01-21
[Meson/TRIx] target dependency for trix engine

jaeyun-jung:
Define target library to build filter-subplugin for trix-engine.


myungjoo (approved):

gichan-jang (approved):
LGTM

PR #3659, opened 2022-02-11
[Svace] fix svace issues

jaeyun-jung:
Fix svace issues
- init local variable
- check allocation failure case

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
@@ -110,8 +110,8 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
   struct ncGraphHandle_t *handle_graph = NULL;
   struct ncTensorDescriptor_t tensor_desc_input;
   struct ncTensorDescriptor_t tensor_desc_output;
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/tensor_common_pipeline.c
```diff
@@ -597,6 +597,7 @@ gst_tensor_aggregation_add_data (GHashTable * table, const guint32 key)
   else
     hashkey = key;
   aggr = g_new0 (gst_tensor_aggregation_data_s, 1);
+  g_assert (aggr != NULL);
```
myungjoo, line 600: Don't fix this. Mark this "false positive". I have reported this issue to SE (SVACE maintainer) and it appears that they didn't fix it yet; g_new series are supposed to kill the process if g_new fails.

myungjoo (commented):
LGTM except for checking g_new0 return.

gst/nnstreamer/tensor_common_pipeline.c
myungjoo: Besides, try not to ass g_assert in non-test-case codes. Try to make proper error handling code instead.

jaeyun-jung: updated

myungjoo (approved):

anyj0527 (approved):
LGTM✔

PR #3662, opened 2022-02-15
[MqttSrc/Sink] release mqtt handle

jaeyun-jung:
Destroy mqtt-async handle when the element instance is finalized.


anyj0527 (approved):
LGTM ✔

Issue #3657, opened 2022-02-10
[tensor query] Unittest fails


kparichay:
# Issue Description
nnstreamer_query unit test fails sometimes.

Expected Result
============
All unit tests must pass at all times.

How to Reproduce
===============
I haven't been able to reproduce on my local tizen i586 build with 15 runs and counting. 
The script is still running to find reproduce the error. 
For now, the error logs can still be found at the [CI build](http://nnstreamer.mooo.com/nnstreamer/ci/repo-workers/pr-checker/3653-202202090930080.9313690662384-b517c895e615ee718a40b724de3ca26c21424584/report/build_log_3653_tizen_i586_output.txt) of #3653.

Further Information
===============
The failure can also be currently seen on the CI result of the #3653 itself as of now.

Update: all builds pass for over 200 runs. IMO the script needs to run on the CI over the weekend.

myungjoo:
The log will be removed. Please paste the related part here.

kparichay:
Error log : 
<details><summary> Tizen i586 error log <\summary>
...

[  468s] [Starting] nnstreamer_query
[  496s] ==================================================
[  496s]     Test Group nnstreamer_query Starts.
[  496s] [PASSED] 1-2:gst-launch of case 1-2
[  496s] [PASSED] 1-3:Compare 1-3
[  496s] [PASSED] 1-4:Compare 1-4
[  496s] [PASSED] 1-5:Compare 1-5
[  496s] [PASSED] 2-3:gst-launch of case 2-3
[  496s] [PASSED] 2-4:Compare 2-4
[  496s] [PASSED] 2-5:Compare 2-5
[  496s] [PASSED] 2-6:Compare 2-6
[  496s] [PASSED] 2-7:Compare 2-7
[  496s] [PASSED] 2-8:Compare 2-8
[  496s] [PASSED] 2-9:Compare 2-9
[  496s] [PASSED] 3-2:gst-launch of case 3-2
[  496s] [PASSED] 3-3:Compare 3-3
[  496s] [PASSED] 3-4:Compare 3-4
[  496s] [PASSED] 3-5:Compare 3-5
[  496s] [FAILED][Critical] 5-3:gst-launch of case 5-3
[  496s] [FAILED][Critical] 5-4:Compare 5-4
[  496s] [FAILED][Critical] 5-5:Compare 5-5
[  496s] [FAILED][Critical] 5-6:Compare 5-6
[...]

PR #3663, opened 2022-02-16
[TEST] Increase sleep time of the query test

gichan-jang:
1 second sleep time is not enough to operate query server.
see: https://github.com/nnstreamer/nnstreamer/issues/3657#issuecomment-1038708557

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_query/runTest.sh
+++ b/tests/nnstreamer_query/runTest.sh
@@ -20,7 +20,10 @@ testInit $1
 
 PATH_TO_PLUGIN="../../build"
 PERFORMANCE=0
-TIMEOUT_SEC=5
+# The client has to wait enough time for the server to be ready.
+# Related issue: https://github.com/nnstreamer/nnstreamer/issues/3657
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_query/runTest.sh
```diff
@@ -20,7 +20,8 @@ testInit $1
 
 PATH_TO_PLUGIN="../../build"
 PERFORMANCE=0
-TIMEOUT_SEC=5
+SLEEPTIME_SEC=5
```
kparichay, line 23: I recommend to please add a comment here about setting this as 5 sec or link the issue number of future reference.

kparichay (approved):
LGTM, good find

tests/nnstreamer_query/runTest.sh
gichan-jang: I added the comment. Thanks.

myungjoo (approved):
LGTM

PR #3675, opened 2022-03-01
[SSAT] Enable filter_python3 test to run on target @open sesame 3/7 18:10

ksghaier:
- remove exit if build folder not exists
    - bypass pushd failure if build folder not exists

### DIFF
--- a/tests/nnstreamer_filter_python3/runTest.sh
+++ b/tests/nnstreamer_filter_python3/runTest.sh
@@ -37,23 +37,31 @@ if [[ -d $PATH_TO_PLUGIN ]]; then
     fi
 else
     echo "No build directory"
-    report
-    exit
+    # Check if module is installed.
+    if python3 -c "import nnstreamer_python" &> /dev/null; then
+        echo 'nnstreamer_python module exists'
+    else
+        echo 'nnstreamer_python module does not exist'
+        report
+        exit
+    fi
 fi
 
 FRAMEWORK="python3"
-# This symlink is necessary only for testcases; when installed, symlinks will be made
-pushd ../../build/ext/nnstreamer/tensor_filter
-TEST_PYTHONPATH=${FRAMEWORK}_pymodule
-mkdir -p ${TEST_PYTHONPATH}
-pushd ${TEST_PYTHONPATH}
-# Covert to an absolute path from the relative path
-export PYTHONPATH=$(pwd)
-if [[ ! -f ./nnstreamer_python.so ]]; then
-  ln -s ../../extra/nnstreamer_${FRAMEWORK}.so nnstreamer_python.so
+if [[ -d ../../build ]]; then
+    # This symlink is necessary only for testcases; when installed, symlinks will be made
+    pushd ../../build/ext/nnstreamer/tensor_filter
+    TEST_PYTHONPATH=${FRAMEWORK}_pymodule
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_filter_python3/runTest.sh
```diff
@@ -37,13 +37,11 @@ if [[ -d $PATH_TO_PLUGIN ]]; then
...
 else
     echo "No build directory"
-    report
-    exit
 fi
 
 FRAMEWORK="python3"
 # This symlink is necessary only for testcases; when installed, symlinks will be made
-pushd ../../build/ext/nnstreamer/tensor_filter
+pushd ../../build/ext/nnstreamer/tensor_filter || echo "INFO: ../../build/ext/nnstreamer/tensor_filter folder not found" && pushd ${PWD}
```
gichan-jang, line 44: If a built dir exists, an error occurs because `../../build/ext/nnstreamer/tensor_filter` is pushed twice.
Check the result of the `pushd`.

tests/nnstreamer_filter_python3/runTest.sh
```diff
@@ -37,13 +37,11 @@ if [[ -d $PATH_TO_PLUGIN ]]; then
     fi
 else
     echo "No build directory"
-    report
```
gichan-jang, line 40: It seems that you only want to test the nnstreamer without building.
But I think these lines are necessary.
How about changing like this?
```
# Default installation path. If you installed it somewhere else, change it.
INSTALL_PATH=/usr/lib/nnstreamer/filters/
check=$(ls ${INSTALL_PATH} | grep python3.so)
if [[ ! $check ]]; then
    echo "Cannot find python3 shared lib"
    report
    exit
fi
```

tests/nnstreamer_filter_python3/runTest.sh
ksghaier: yes correct, I updated the code

gichan-jang (approved):
LGTM

ksghaier:
enhanced path search to check if module is installed

tests/nnstreamer_filter_python3/runTest.sh
```diff
@@ -37,23 +37,31 @@ if [[ -d $PATH_TO_PLUGIN ]]; then
...
-    exit
+    # Check if module is installed.
+    INSTALL_PATHS=$(python -c 'import sys; print ("\n".join(sys.path))')
```
gichan-jang, line 41:
```suggestion
    INSTALL_PATHS=$(python3 -c 'import sys; print ("\n".join(sys.path))')
```

Or.. How about import module?
```
    if python3 -c "import nnstreamer_python" &> /dev/null; then
        echo 'nnstreamer_python module exists.'
    else
        echo 'nnstreamer_python module does not exist'
        report
        exit
    fi
```

myungjoo (approved):
LGTM

myungjoo:
CI system malfunction? :(

gichan-jang:
> CI system malfunction? :(

Right.. There is CI system problem with the tensor converter using Python. Test is stopped during destruct the python converter. I'm finding a reason..

ksghaier:
> @ksghaier Please rebase this against the recent main branch. It will reactivate pr-postbuild tests.

rebased

PR #3676, opened 2022-03-02
[Util] util function to compare tensor-config

jaeyun-jung:
remove gst-dependency to compare tensor config.
add return val of invalid condition.


gichan-jang (approved):
LGTM

myungjoo (approved):

PR #3682, opened 2022-03-04
[build/tflite] Check each delegate symbol in build time

### LINKED ISSUE #3674: TensorFlow Lite filter - link issue depending on delegates supported by TFlite lib
Hello all,

# Issue Description
A change has been merged recently into main branch changing the configuration options for NNStreamer TFLite filter:
https://github.com/nnstreamer/nnstreamer/commit/2a86a9c9147a8b59693af9dbee4b96876fb035f9
Before this change, delegates support of GPU, NNAPI and XNNPACK was configurable via meson option.
It is no longer the case and from now on, support of this 3 delegates is expected by NNStreamer filter to be always available.

This is not an issue at build time because headers for every delegate are exported by the lib so can be #include-d by filter.
However it is an issue at  link time, because delegates support in TFLite lib is defined at library configuration time by its cmake:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/CMakeLists.txt#L64
It still stands true on main branch of tensorflow upstream project: it can not be expected that all delegates are built in the lib.
When one delegate is not compiled in TFLite lib, NNStreamer TFLite filter can not link, because it references symbols missing in TFLite lib.  

We have the issue using a TFLite 2.6.0 library without GPU delegate compiled in - error output below.
But similar issue would presumably show up with any missing delegate and with any TFLite version.

Error at link reported during subplugin (.so) build:
tensor_filter_tensorflow_lite.cc.o: in function `TFLiteInterpreter::loadModel(int, tflite_delegate_e)::{lambda(TfLiteDelegate*)#2}::_FUN(TfLiteDelegate*)':
tensor_filter_tensorflow_lite.cc:475: undefined reference to `TfLiteGpuDelegateV2Delete'
tensor_filter_tensor_filter_tensorflow_lite.cc.o: in function `TFLiteInterpreter::loadModel(int, tflite_delegate_e)':
tensor_filter_tensorflow_lite.cc:458: undefined reference to `TfLiteGpuDelegateOptionsV2Default'
tensor_filter_tensorflow_lite.cc:472: undefined reference to `TfLiteGpuDelegateV2Create'
collect2: error: ld returned 1 exit status
 

Expected Result
============
NNStreamer tflite filter should be able to link with multiple flavours of TFLite library, even when all delegates (XNNPACK, GPU, NNAPI) are not built-in.
Note: same NNStreamer change has added a TODO to remove the meson option for external delegate. Doing so would presumably lead to the same issue.

Can the previous meson options be reenabled, as well as conditional build in the NNStreamer tflite filter of the delegate-specific block
option('tflite2-gpu-delegate-support', type: 'boolean', value: 'false')
option('tflite2-nnapi-delegate-support', type: 'boolean', value: 'false')
option('tflite2-xnnpack-delegate-support', type: 'boolean', value: 'false')

How to Reproduce
===============
Build NNStreamer TFLite filter on a file system hosting a TFLite library without GPU delegate.

Thank you for your support.

anyj0527:
- Check each tflite delegate symbol by linking with tflite dep in build time

**Self evaluation:**
1. Build test: [X]Passed [ ]Failed []Skipped
2. Run test: [X]Passed [ ]Failed []Skipped

This resolves #3674

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -223,9 +223,40 @@ if tflite2_support_is_available
     tflite2_compile_args += '-DUSE_TENSORFLOW2_HEADER_PATH=1'
   endif
 
-  # TODO: tflite v2.8.0+ supports build option for external delegate
-  # remove below option after that.
-  if get_option('tflite2-external-delegate-support')
+  # Check each delegate linking with given tflite library
+  ## xnnpack delegate
+  if cxx.links('''
+        #include "tensorflow/lite/delegates/xnnpack/xnnpack_delegate.h"
+        TfLiteXNNPackDelegateOptions xnnpack_options = TfLiteXNNPackDelegateOptionsDefault ();
+        int main() {return 0;}
+      ''', dependencies : [tflite2_support_deps, thread_dep], name : 'xnnpack delegate')
+    tflite2_compile_args += '-DTFLITE_XNNPACK_DELEGATE_SUPPORTED'
+  endif
+
+  ## gpu delegate
+  if cxx.links('''
+        #include "tensorflow/lite/delegates/gpu/delegate.h"
+        TfLiteGpuDelegateOptionsV2 gpu_options = TfLiteGpuDelegateOptionsV2Default ();
+        int main() {return 0;}
+      ''', dependencies : [tflite2_support_deps, libdl_dep], name : 'gpu delegate')
+    tflite2_compile_args += '-DTFLITE_GPU_DELEGATE_SUPPORTED'
+  endif
+
+  ## nnapi delegate
+  if cxx.links('''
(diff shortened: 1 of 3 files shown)


jvuillaumier:
Verified - it works for me.

jaeyun-jung:
One concern is that, compiler.links() is available from meson version 0.60.

eli-schwartz:
> One concern is that, compiler.links() is available from meson version 0.60.

I think there may be some confusion here, since this function is available from Meson version 0.27 (the stone age, basically).

jaeyun-jung:
> > One concern is that, compiler.links() is available from meson version 0.60.
> 
> I think there may be some confusion here, since this function is available from Meson version 0.27 (the stone age, basically).

My bad. YJ plz ignore my comment.

jaeyun-jung (approved):

gichan-jang (approved):
LGTM!

Issue #3684, opened 2022-03-07
[gbs] python converter test stucks


anyj0527:
With latest tizen/base/latest/ and tizen/unified/latest, the test is stucked after python_converter unittest.

With tizen/unified/tizen-unified_20220228.1/ and tizen/base/tizen-base_20220228.1/,
it does not stucked.

May the version of python3 be the reason?
tizen/base/tizen-base_20220228.1/repos/standard/packages/x86_64/**python3-3.9.1-10.28**.x86_64.rpm
vs.
tizen/base/latest/repos/standard/packages/x86_64/**python3-3.9.10-11.3**.x86_64.rpm

REF: #3683

myungjoo:
Is this addressed by #3685 ?

gichan-jang:
> Is this addressed by #3685 ?

No, #3685  is not directly related with this issue.

myungjoo:
#3872 is trying to resolve this...

PR #3696, opened 2022-03-18
[converter/python3] Temporarily remove Py_Finalize

gichan-jang:
Tizen issues: After python version has been upgraded from 3.9.1 to 3.9.10, python converter is stopped at Py_Finalize.
Since Py_Initialize is not called twice from this object, Py_Finalize is temporarily removed.

Related issue: #3684 

**Self evaluation:**


again4you (approved):
After meeting with all of us , we decided that Py_Finalize() is disabled as workaround.
@gichan-jang added enough comments for this issue and we will fix it next time.

jaeyun-jung (approved):

PR #3715, opened 2022-04-08
[CodeClean] fix svace issue

jaeyun-jung:
1. Set interpreter lock while replacing interpreter in tf-lite sub-plugin.
2. Remove unreached code in converter.
3. Clearly check file ext (bmp) in test util.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -1113,20 +1113,29 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo *info)
 int
 TFLiteCore::reloadInterpreter (TFLiteInterpreter * new_interpreter)
 {
-  TFLiteInterpreter * interpreter_temp = interpreter;
-  if (!gst_tensors_info_is_equal (interpreter->getInputTensorsInfo (),
-          new_interpreter->getInputTensorsInfo ())
-      || !gst_tensors_info_is_equal (interpreter->getOutputTensorsInfo (),
-             new_interpreter->getOutputTensorsInfo ())) {
+  TFLiteInterpreter *old_interpreter = interpreter;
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/include/nnstreamer_plugin_api_util.h
```diff
@@ -23,7 +23,6 @@ G_BEGIN_DECLS
 /**
  * @brief If the given string is NULL, print "(NULL)". Copied from `GST_STR_NULL`
  */
-
```
myungjoo, line 26: Remove this file from the commit.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -1113,19 +1113,21 @@ TFLiteCore::setInputTensorDim (const GstTensorsInfo *info)
...
+      new_interpreter->getOutputTensorsInfo ());
+  interpreter->unlock ();
+
+  if (!in_matched || !out_matched) {
     ml_loge ("The model has unmatched tensors info\n");
     return -EINVAL;
   }
 
-  interpreter_temp->lock ();
   interpreter = new_interpreter;
```
myungjoo, line 1130: What if some other thread was using values of interpreter with interpreter->lock()?

Don't you need to unlock after LINE 1130?

gst/nnstreamer/include/nnstreamer_plugin_api_util.h
jaeyun-jung: updated

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
jaeyun-jung: set new/old interpreter lock in this method.

myungjoo (approved):
LGTM

again4you (approved):
LGTM!

PR #3716, opened 2022-04-10
[Refactoring] Apply test fixture for each test case

again4you:
### Summary
To increase maintainability and modifiability, this patch applies the test fixture framework of Google Test.
It removes duplicate code in each test case and makes a `SetUp()` and `TearDown()` method for each test setup and teardown.

### Before
Data structure and variables in `unittest_rate.cc` are defined as `Global` scope and each test case has duplicate code as below.

apply_test_fixture_before_v2

### After
This patch applies the test fixture framework as below. So duplicate code is moved to `SetUp()` and `TearDown()` methods and is used for each test case. It causes increasing maintainability and modifiability.
apply_test_fixture_after_v2


tests/nnstreamer_filter_single/unittest_filter_single.cc
```diff
@@ -13,212 +13,196 @@
...
+    single = (GTensorFilterSingle *) g_object_new (G_TYPE_TENSOR_FILTER_SINGLE, NULL);
+    klass = (GTensorFilterSingleClass *) g_type_class_ref (G_TYPE_TENSOR_FILTER_SINGLE);
+
+    /* supposed to run test in build directory */
+    if (root_path == NULL)
+      root_path = "..";
+
+    model_file = g_build_filename (root_path, "tests", "test_models", "models",
+        "mobilenet_v1_1.0_224_quant.tflite", NULL);
+    if (!g_file_test (model_file, G_FILE_TEST_EXISTS)) {
```
jaeyun-jung, line 62: When model file is not in model directory or reading file content fails (line 62 and 68), all testcase should be stopped. 
How about using ASSERT_TRUE() for this case?

again4you: Good advice! I updated the code as below.
```cpp
ASSERT_TRUE (g_file_test (model_file, G_FILE_TEST_EXISTS));
```

again4you:
### Submit history
#### v2
* Check the file existance in SetUp() method.

gichan-jang (approved):
LGTM!

jaeyun-jung (approved):

PR #3718, opened 2022-04-12
[lts/6.5][fix/snpe] Handle zero dim models

anyj0527:
- Some snpe model has "0" for their output layer which is same as "1"
- Handle this corner case

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
@@ -589,6 +589,10 @@ snpe_subplugin::configureUserBuffer (zdl::DlSystem::UserBufferMap &buffer_map, c
     std::vector<size_t> strides (bufferShape.rank ());
     strides[strides.size () - 1] = sizeof (float);
     for (size_t i = strides.size () - 1; i > 0; --i) {
+      /* Some model has "0" dim, replacing 0 with 1 */
+      if (bufferShape[i] == 0) {
+        bufferShape[i] = 1;
+      }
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -589,6 +589,10 @@ snpe_subplugin::configureUserBuffer (zdl::DlSystem::UserBufferMap &buffer_map, c
     std::vector<size_t> strides (bufferShape.rank ());
     strides[strides.size () - 1] = sizeof (float);
     for (size_t i = strides.size () - 1; i > 0; --i) {
+      /* Some model has "0" dim, replacing 0 with 1 */
+      if (bufferShape[i] == 0) {
```
myungjoo, line 593: Is this the SNPE bug you've mentioned yesterday?

myungjoo: let's add something like:
```
  static int reported = 0;
  if (G_UNLIKELY(!reported)) {
    reported = 1;
    log info-level (....)
  }
```

anyj0527: Right, after replacing 0 with 100, it works. Now I've asked the model making counterpart.

myungjoo (approved):
LGTM. Proceed when you are ready. And please consider that "one-time" log-info printing.

anyj0527:
@myungjoo This PR is WIP, can it be reverted?

Issue #3713, opened 2022-04-07
[Feature] tf-lite custom support 


myungjoo:
Supposed Usage Example
1. tensor_filter framework=tensorflow-lite-custom model=${path}.tflite fw_location=/usr/opt/app/whatsoever/tflite-custom-whatsoever.so
2. ml-inf-single usage

Timeline: no specific due date, but let's try to release w/ Tizen 7.0 M2 (priority is lower than other established PLM/RELEASE tasks)

Plan A: use dlopen. get class instances of required tflite features.

Findings:
1. Cannot use dlopen for getting tflite.so instance. It cannot be used for arbitrary tflite.so (need a "C" exported constructors and destructors of each tflite binary instance!)
2. Plan A is inappropriate.

Plan B: let it be linked to "libtensorflow2-lite-custom.so" and user defines LD path in launch-time. This limits functionality a bit (cannot have multiple different custom tflite custom binaries for a given app.), but it will work. By default, let it try to load it from the "current" path of the app.

myungjoo:
Plan B Issue: when libtensorflow-lite-custom.so is not available at the pipeline construction (tensor-filter loading), we need to generate proper (not killing the app) error!

myungjoo:
#4268 will complete this.

PR #3727, opened 2022-04-15
filter:tf-lite2-custom usage manual

myungjoo:
For #3713, add a usage manual because it may be not too trivial
for non-system-sw developers.

### DIFF
--- a/ext/nnstreamer/tensor_filter/README.md
+++ b/ext/nnstreamer/tensor_filter/README.md
@@ -1,0 +1,48 @@
+---
+title: NNStreamer tensor\_filter default subplugin manual
+...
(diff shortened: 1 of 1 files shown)


myungjoo:
Note: with dladdr, tensor-filter can designate directory of libtensorflow2-lite-custom.so in run-time right before calling dlopen. It designates shared-lib location for dlopen-ed library.

TODO: add a custom property of tensorflow2-lite-custom that designates dladdr location.
TODO2: consider security hazard.

DonghakPark (approved):
LGTM

jaeyun-jung (approved):

PR #3737, opened 2022-04-21
[lts][filter/snpe] Revert "[fix/snpe] Handle zero dim models" and fix the order of output tensors

anyj0527:
This reverts commit https://github.com/nnstreamer/nnstreamer/commit/7826866a7524b6769b9d5476c6d9bf04f0105ea7.
- Still figuring out how to deal with this "zero dim".

Make the order of output tensors be consistent with the user given option
- For multiple output models, let user set output tensors instead of output layers
- And make the order of input / output tensors be consistent.

Updated:
Some snpe models have "0" dim to represent resizable dim.
Add a custom option "MaxResizableDim" to deal with this case.

**Self evaluation:**


myungjoo (approved):
LGTM. / PTAL: @again4you or @jaeyun-jung

again4you (approved):
LGTM!

PR #3744, opened 2022-04-26
[TEST/ntp] Add ntp unittest using GMock.

gichan-jang:
Add ntp uniitest using GMock.
ntp unittest is enabled only on tizen.

**Self evaluation:**

### DIFF
--- a/gst/mqtt/ntputil.c
+++ b/gst/mqtt/ntputil.c
@@ -14,7 +14,7 @@
 
 #include <errno.h>
 #include <netdb.h>
-#include <netinet/ip.h>
+#include <arpa/inet.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
@@ -120,6 +120,15 @@ const char NTPUTIL_DEFAULT_HNAME[] = "pool.ntp.org";
 const uint16_t NTPUTIL_DEFAULT_PORT = 123;
 
 /**
+ * @brief Wrapper function of ntohl.
+ */
+uint32_t
+_convert_to_host_byte_order (uint32_t in)
+{
+  return ntohl (in);
+}
+
+/**
  * @brief Get NTP timestamps from the given or public NTP servers
  * @param[in] hnums A number of hostname and port pairs. If 0 is given,
  *                  the NTP server pool will be used.
@@ -208,8 +217,8 @@ ntputil_get_epoch (uint32_t hnums, char **hnames, uint16_t * ports)
      * such as microsceonds. Note that the bit/byte order of those data should
      * be converted to the host's endianness.
      */
-    recv_sec = ntohl (packet.xmit_ts.sec);
-    recv_frac = ntohl (packet.xmit_ts.frac);
+    recv_sec = _convert_to_host_byte_order (packet.xmit_ts.sec);
+    recv_frac = _convert_to_host_byte_order (packet.xmit_ts.frac);
 
     /**
      * @note NTP uses an epoch of January 1, 1900 while the Unix epoch is
--- a/gst/mqtt/ntputil.h
+++ b/gst/mqtt/ntputil.h
@@ -10,10 +10,14 @@
  * @author  Wook Song <wook16.song@samsung.com>
  * @bug     No known bugs except for NYI items
  */
+#ifndef __NNS_NTP_UTIL_H__
+#define __NNS_NTP_UTIL_H__
 
 #include <stdint.h>
 #include <time.h>
-
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
 /**
  * @brief Get NTP timestamps from the given or public NTP servers
  * @param[in] hnums A number of hostname and port pairs. If 0 is given,
@@ -25,3 +29,15 @@
  */
 int64_t
 ntputil_get_epoch (uint32_t hnums, char **hnames, uint16_t * ports);
+
+/**
+ * @brief Converting network byte order to host byte order.
+ * @note There is a problem in mocking ntohl in GMock, so the wrapper function is temporarily used.
+ * */
+uint32_t
+_convert_to_host_byte_order (uint32_t in);
+
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+#endif /* __NNS_NTP_UTIL_H__ */
--- a/tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
+++ b/tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
@@ -1,0 +1,310 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * @file        unittest_ntp_util_mock.cc
+ * @date        25 Apr 2022
+ * @brief       Unit test for ntp util using GMock.
+ * @see         https://github.com/nnstreamer/nnstreamer
+ * @author      Gichan Jang <gichan2.jang@samsung.com>
+ * @bug         No known bugs
+ */
+
+#include <glib.h>
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <unittest_util.h>
+#include "../gst/mqtt/ntputil.h"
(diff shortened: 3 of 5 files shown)


tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
```diff
@@ -0,0 +1,317 @@
...
+ * @brief Free hostent.
+ */
+static void _hostent_free (struct hostent *host)
+{
+  g_free (host->h_name);
+  g_free (host->h_aliases[0]);
+  g_free (host->h_aliases);
+  g_free (host->h_addr_list[0]);
+  g_free (host->h_addr_list);
+}
```
again4you, line 130: I found out that every test case calls _hostent_init() before test and _hostent_free() after test.
What about applying the Test Fixture of Google Test instead of calling both of them?

For example, `NtpUtilMockTest` class for Test Fixture can be defined as below.

```cpp
clsss NtpUtilMockTest : public::testing::Test {
  ...
  
protected:
  struct hostent host;
  ...
  
public:
  void SetUp() override
  {
    host.h_name = g_strdup ("github.com");
    host.h_aliases = (char **) calloc (1, sizeof (char *));
    host.h_aliases[0] = g_strdup ("www.github.com");
    host.h_addrtype = AF_INET;
    host.h_length = 4;
    host.h_addr_list = (char **) calloc (1, sizeof (char *));
    host.h_addr_list[0] = g_strdup ("52.78.231.108");
  }
  
  void TearDown() override
  {
    g_free (host.h_name);
    g_free (host.h_aliases[0]);
    g_free (host.h_aliases);
    g_free (host.h_addr_list[0]);
    g_free (host.h_addr_list);
  }
};
```

tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
```diff
@@ -0,0 +1,317 @@
...
+      .WillOnce(Return(NTPUTIL_TIMESTAMP_DELTA + 1ULL))
+      .WillOnce(Return(1ULL));
+
+  ret = ntputil_get_epoch (1, (char **)hnames, ports);
+
+  EXPECT_GE (ret, 0);
+
+  _hostent_free (&temp_hostent);
+  delete mockInstance;
+}
```
again4you, line 164: After that, we can run each test case without calling `_hostent_init()` and `_hostent_free() as below.
```cpp
TEST_F (NtpUtilMockTest, Normal_p)
{
  ...
  mockInstance = new NtpUtilMock ();

  EXPECT_CALL (*mockInstance, gethostbyname(_))
      .Times(1).WillOnce(Return((struct hostent *)&temp_hostent));
  ...

  ret = ntputil_get_epoch (1, (char **)hnames, ports);

  EXPECT_GE (ret, 0);
  ...
}
```



In my opinion, it is easy to maintain since the test framework automatically calls both SetUp() and TearDown() methods.

gst/mqtt/ntputil.h
```diff
@@ -10,10 +10,15 @@
  * @author  Wook Song <wook16.song@samsung.com>
  * @bug     No known bugs except for NYI items
  */
+#ifndef __NNS_NTP_UTIL_H__
+#define __NNS_NTP_UTIL_H__
 
 #include <stdint.h>
 #include <time.h>
-
+#include <byteswap.h>
```
again4you, line 18: This header file is not necessary so please remove it.

gst/mqtt/ntputil.c
```diff
@@ -119,6 +119,15 @@ const int64_t NTPUTIL_SEC_TO_USEC_MULTIPLIER = 1000000;
 const char NTPUTIL_DEFAULT_HNAME[] = "pool.ntp.org";
 const uint16_t NTPUTIL_DEFAULT_PORT = 123;
 
+/**
+ * @brief Wrapper function of ntohl.
+ */
+uint32_t
+_convert_to_host_byte_order (uint32_t in)
+{
+  return ntohl (in);
```
again4you, line 128: Question.
I can't find `#include <netinet/in.h>` or `#include <arpa/inet.h>` in this file.
Should we include one of them instead of `#include <netinet/ip.h>`?

gst/mqtt/ntputil.h
```diff
@@ -25,3 +30,15 @@
  */
 int64_t
 ntputil_get_epoch (uint32_t hnums, char **hnames, uint16_t * ports);
+
+/**
+ * @brief Converting network byte order to host byte order.
+ * @note There is a problem in mocking ntohl in GMock, so the wrapper function is temporarily used.
+ * */
+uint32_t
+_convert_to_host_byte_order (uint32_t in);
```
again4you, line 39: Good alternative solution!

At first, I wondered why you define the wrapper function instead of directly making a mock like `connect()`. So I made a simple test code and found out that a redefinition error occurs when making a mock. Your alternative solution looks reasonable since it is directly called `ntohl()`, which is a test target. :+1:

tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
```diff
@@ -0,0 +1,317 @@
...
+  EXPECT_CALL (*mockInstance, connect(_, _, _))
+      .Times(1).WillOnce(Return(0));
+  EXPECT_CALL (*mockInstance, write(_, _, _))
+      .Times(1).WillOnce(Return(0));
+  EXPECT_CALL (*mockInstance, read(_, _, _))
+      .Times(1).WillOnce(Return(0));
+  EXPECT_CALL (*mockInstance, _convert_to_host_byte_order(_))
+      .Times(2)
+      .WillOnce(Return(1ULL))
+      .WillOnce(Return(1ULL));
```
again4you, line 286: Great! Your test code checks its call count :100:

tests/meson.build
```diff
@@ -160,6 +160,18 @@ if gtest_dep.found()
           install_dir: unittest_install_dir)
 
       test('unittest_mqtt', unittest_mqtt, env: testenv)
+
+      # ntp util test
+      gmock_dep = dependency('gmock', required: false)
+      if get_option('enable-tizen') and gmock_dep.found()
```
again4you, line 166: Nice work!
Ubuntu 18.04 does not support `libgmock-dev` package so this condition looks OK!

again4you (commented):
All test scenarios are great!
Each mock function returns an appropriate value for each test case and checks its call count.

In my opinion, there is an unnecessary header in the code so please remove it. Moreover, the Test Fixture of Google test can remove `_hostent_init()` and `_hostent_free()` calling in each test case.

gst/mqtt/ntputil.h
gichan-jang: Right. This header was include during development. Removed. Thanks.

gst/mqtt/ntputil.c
gichan-jang: I changed to include <arpa/inet.h> instead of <netinet/ip.h>. Thanks.

gst/mqtt/ntputil.h
gichan-jang: Right. I also tried to solve the redefinition problem, but I couldn't find a way, so I used the wrapper function as an alternative.

tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
gichan-jang: Right. This method looks better. I applied test fixture. Thanks!

tests/gstreamer_mqtt/unittest_ntp_util_mock.cc
gichan-jang: Thanks! Applied.

tests/meson.build
gichan-jang: Right. I excluded it from Ubuntu because we distribute packages on both 18.04 and 20.04 from Launchpad.

wooksong (approved):
Looks great to me!

PR #3749, opened 2022-05-04
test: disable converter test if python is not available.

myungjoo:
converter test uses python scripts.
disable converter test if python is not available.


anyj0527 (approved):
LGTM ✔

Issue #3757, opened 2022-05-12
Additional instructions for tensor_filter with python3 framework


mholtta:
# Additional instructions for tensor_filter with python3 framework
Would it be possible to get additional information on what is needed to run tensor_filter with framework=python3? And also information on what is needed to setup Python environment such that importing nnstreamer_python in python would work?

My current issue seems to be that GStreamer doesn't find the nnstreamer-python elements and tensor_filter doesn't recognize `framework=python3`. Additionally, nnstreamer_python cannot be imported in Python. Pipelines with tensor_filter with TensorFlow Lite framework work as expected.

How to Reproduce
===============
1. Build dockerimage (image definition below)
2. Rename [passthrough.py](https://github.com/nnstreamer/nnstreamer/blob/2a42dda5b773d4c1a7ba29db7faa611f6eb0c6fe/tests/test_models/models/passthrough.py) to nnnstreamer_filter_passthrough.py, run the dockerimage while mounting nnnstreamer_filter_passthrough.py
3. Run test pipeline (below)
4. See error (below)

**I tested the below pipeline with tensor_filter python3 framework:**
`gst-launch-1.0 videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=280,height=40,framerate=0/1 ! videoconvert ! video/x-raw, format=RGB ! tensor_converter ! tee name=t ! queue ! tensor_filter framework=python3 model=nnnstreamer_filter_passthrough.py input=\"3:280:40:1\" inputtype=\"uint8\" output=\"3:280:40:1\" outputtype=\"uint8\" ! filesink location=\"testcase1.passthrough.log\" sync=true t. ! queue ! filesink location=\"testcase1.direct.log\" sync=true`

The pipeline has been adopted from nnstreamer unittests - [see this link](https://github.com/nnstreamer/nnstreamer/blob/2a42dda5b773d4c1a7ba29db7faa611f6eb0c6fe/tests/nnstreamer_filter_python3/runTest.sh#L68). The nnstreamer_filter_passthrough.py in the pipeline is passthrough.py from nnstreamer unittests - [see link here](https://github.com/nnstreamer/nnstreamer/blob/2a42dda5b773d4c1a7ba29db7faa611f6eb0c6fe/tests/test_models/models/passthrough.py).

**The resulting error from running the pipeline:**
```
[...]

PR #3762, opened 2022-05-17
[Python] path to python helper lib

jaeyun-jung:
Change install path for python-helper library.
This is to fix library link error.

### DIFF
--- a/debian/nnstreamer-python3.install
+++ b/debian/nnstreamer-python3.install
@@ -1,4 +1,4 @@
+/usr/lib/*/nnstreamer_python3.so
 /usr/lib/nnstreamer/filters/libnnstreamer_filter_python3.so
 /usr/lib/nnstreamer/converters/libnnstreamer_converter_python3.so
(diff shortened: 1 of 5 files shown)


gichan-jang (approved):
LGTM.
This patch resolves the library link error mentioned in #3757.

myungjoo (commented):
and... do you think we need to rename it?

e.g., "libnnstreamer-python3.so"

jaeyun-jung:
> and... do you think we need to rename it?
> 
> e.g., "libnnstreamer-python3.so"

I think this is related with python module name (PyInit_nnstreamer_python), so IMO it is unnecessary.

myungjoo:
> > and... do you think we need to rename it?
> > e.g., "libnnstreamer-python3.so"
> 
> I think this is related with python module name (PyInit_nnstreamer_python), so IMO it is unnecessary.

Ok!

myungjoo (approved):
LGTM

PR #3765, opened 2022-05-23
[Meson] Fix warnings related to the run_command call

wooksong:
This patch sets the keyword argument, 'check', of the run_command
function to true in order to fix the following warnings:

WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -526,17 +526,19 @@ if get_option('enable-mediapipe')
     error('Not Supported System & Architecture. Linux x86_64 is required')
   endif
 
-  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
+  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME', check : true)
   MEDIAPIPE_HOME = cmd.stdout().strip()
-
-  mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']').returncode()
-  mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode()
-
-  if mediapipe_env_exist != 1 and mediapipe_dir_exist != 0
+  if MEDIAPIPE_HOME == ''
+    error('MEDIAPIPE_HOME is not set')
+  endif
+  message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
+  # TODO: This is not a portable way to check if a directory exists
+  # After bumping up to 0.53.0, the following line can be replaced with one that uses FS module
+  mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']', check : false).returncode()
+  if mediapipe_dir_exist != 0
     error('Cannot find MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
   endif
 
-  message('MEDIAPIPE_HOME: ' + MEDIAPIPE_HOME)
   mediapipe_incdir = include_directories(
     MEDIAPIPE_HOME,
     join_paths(MEDIAPIPE_HOME, 'bazel-bin'),
@@ -544,10 +546,7 @@ if get_option('enable-mediapipe')
     join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
   )
 
-  cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
-  if cmd.returncode() != 0
-    error(cmd.stderr().strip())
-  endif
+  run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir, check : true)
 
   mediapipe_internal_dep = cxx.find_library('mediapipe_internal', dirs: nnstreamer_libdir)
   mediapipe_external_dep = cxx.find_library('mediapipe_external', dirs: nnstreamer_libdir)
@@ -593,7 +592,7 @@ if get_option('enable-mediapipe')
   )
 endif
 
-run_command('sh', '-c', 'touch filter_snpe_list')
+run_command('sh', '-c', 'touch filter_snpe_list', check : true)
 if snpe_support_is_available
   filter_sub_snpe_sources = ['tensor_filter_snpe.cc']
 
@@ -622,7 +621,7 @@ if snpe_support_is_available
     install: true,
     install_dir: nnstreamer_libdir
   )
-  run_command('sh', '-c', 'echo "%{_prefix}/lib/nnstreamer/filters/libnnstreamer_filter_snpe.so" >> filter_snpe_list')
+  run_command('sh', '-c', 'echo "%{_prefix}/lib/nnstreamer/filters/libnnstreamer_filter_snpe.so" >> filter_snpe_list', check : true)
 endif
 
 if tensorrt_support_is_available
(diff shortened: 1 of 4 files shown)


myungjoo (approved):
LGTM! (however, please address gichan's concern)

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -526,11 +526,11 @@ if get_option('enable-mediapipe')
     error('Not Supported System & Architecture. Linux x86_64 is required')
   endif
 
-  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME')
+  cmd = run_command('sh', '-c', 'echo $MEDIAPIPE_HOME', check : true)
   MEDIAPIPE_HOME = cmd.stdout().strip()
 
-  mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']').returncode()
-  mediapipe_dir_exist = run_command('[', '-d', MEDIAPIPE_HOME, ']').returncode()
+  mediapipe_env_exist = run_command('[', '-z', MEDIAPIPE_HOME, ']', check : true).returncode()
```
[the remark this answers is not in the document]
eli-schwartz: In the context of what you're trying to do here...

You do two run_command invocations, and if the following two conditions are both true:
- the first succeeds
- the second fails

you emit a manual `error()`.

In this case since you're explicitly checking the return value, you should probably use `check: false`

But, there's a bigger issue to consider. The first one can be rewritten to not use run_command at all.

```meson
mediapipe_env_exist = MEDIAPIPE_HOME != ''

# This requires bumping to Meson 0.53.0
fs_mod = import('fs')
mediapipe_dir_exist = fs_mod.is_dir(MEDIAPIPE_HOME)
```

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -544,7 +544,7 @@ if get_option('enable-mediapipe')
     join_paths(MEDIAPIPE_HOME, 'bazel-mediapipe/external/com_google_absl'),
   )
 
-  cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir)
+  cmd = run_command('sh', '-c', meson.source_root() + '/tools/development/gen_mediapipe_libs.sh ' + nnstreamer_libdir, check : true)
   if cmd.returncode() != 0
     error(cmd.stderr().strip())
```
eli-schwartz, line 549: You can now drop this error message, and I don't think you need to save the return value to `cmd`.

The `check: true` change means that Meson now errors out for you.

meson.build
```diff
@@ -157,7 +157,7 @@ flatc = find_program('flatc', required : get_option('flatbuf-support'))
 flatbuf_dep = disabler()
 flatbuf_version_check_dep = disabler()
 if flatc.found()
-  flatc_ver = run_command(flatc, '--version').stdout().split()[2]
+  flatc_ver = run_command(flatc, '--version', check : true).stdout().split()[2]
   flatbuf_dep = dependency('flatbuffers', version: flatc_ver,
```
eli-schwartz, line 161: If you bump to Meson 0.62.0 (yes I know that's really new, you may not be able to), then you can use `flatc.version()`, see https://mesonbuild.com/Reference-manual_returned_external_program.html#external_programversion

meson.build
```diff
@@ -191,11 +191,11 @@ if not get_option('snpe-support').disabled()
...
       # Check whether $SNPE_ROOT (where SNPE SDK is located) is set.
-      cmd = run_command('sh', '-c', 'echo $SNPE_ROOT')
+      cmd = run_command('sh', '-c', 'echo $SNPE_ROOT', check : true)
       SNPE_ROOT = cmd.stdout().strip()
       if SNPE_ROOT != ''
         message('Got $SNPE_ROOT: @0@'.format(SNPE_ROOT))
-        snpe_dir_not_exist = run_command('[', '-d', SNPE_ROOT, ']').returncode()
+        snpe_dir_not_exist = run_command('[', '-d', SNPE_ROOT, ']', check : true).returncode()
         if snpe_dir_not_exist != 0
           error('@0@ does not exists'.format(SNPE_ROOT))
```
eli-schwartz, line 200: This error message can be removed again, if you use `check: true`.

However again, you can use the fs module's `fs_mod.is_dir(SNPE_ROOT)` instead.

tools/development/parser/meson.build
```diff
@@ -4,7 +4,7 @@ flex_cdata = configuration_data()
 flex_min_version='2.5.31'
 flex = find_program('flex', 'win_flex')
 
-flexversion_res = run_command([flex, '--version'])
+flexversion_res = run_command([flex, '--version'], check : true)
```
eli-schwartz, line 7: I think you just want to use `find_program('flex', 'win_flex', version: '>=2.5.31')` here and skip all this manual version checking?

tools/development/parser/meson.build
```diff
@@ -33,7 +33,7 @@ bison_cdata = configuration_data()
 bison_min_version='2.4'
 bison = find_program('bison', 'win_bison')
 
-bversion_res = run_command([bison, '--version'])
+bversion_res = run_command([bison, '--version'], check : true)
```
eli-schwartz, line 36: Same thing here, just use `find_program('bison', 'win_bison', version: '>=2.4')`

eli-schwartz (changes requested):

ext/nnstreamer/tensor_filter/meson.build
myungjoo: I agree.

meson.build
myungjoo: That's good; however, before applying this, we will wait (guess a year or two :( ) until we drop a few old versions of OS supported.

eli-schwartz: Adding a TODO comment for the future is probably the best you can do for now then.

ext/nnstreamer/tensor_filter/meson.build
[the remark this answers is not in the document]
wooksong: @gichan-jang The first line has been replaced and the other one seems to still require it so that ```check: false``` should be provided.

@eli-schwartz Thanks. I've revised my PR based on your comment.

ext/nnstreamer/tensor_filter/meson.build
wooksong: Indeed. The if block to test ```cmd.returncode()``` has been removed.

meson.build
wooksong: I left a TODO comment for it. Thanks again.

meson.build
wooksong: The FS (filesystem) module requires bumping up to 0.53.0 so that I left a TODO comment as you commented.

tools/development/parser/meson.build
wooksong: Thanks! It is applied.

myungjoo (approved):
LGTM

meson.build
```diff
@@ -157,7 +157,9 @@ flatc = find_program('flatc', required : get_option('flatbuf-support'))
 flatbuf_dep = disabler()
 flatbuf_version_check_dep = disabler()
 if flatc.found()
-  flatc_ver = run_command(flatc, '--version').stdout().split()[2]
+  # TODO: After bumping up meson version to 0.52.0, we can use flatc.version()
```
gichan-jang, line 160: typo : 0.62.0

gichan-jang (commented):
LGTM with fixing minor typo.

meson.build
wooksong: Fixed.

gichan-jang (approved):

PR #3770, opened 2022-05-31
[macOS/meson] Add a missing parameter to the sed's in-place edit option

wooksong:
In macOS, the '-i' option of the sed command requires a parameter to
tell what extension to use for the backup file. To this end, this patch
appends '.bak' to the '-i' option to fix the build errors on macOS.

### DIFF
--- a/tests/nnstreamer_filter_extensions_common/meson.build
+++ b/tests/nnstreamer_filter_extensions_common/meson.build
@@ -48,6 +48,10 @@ if have_python3
 endif
 
 sed_command = find_program('sed', required: true)
+sed_command_edit_in_place_option = '-i'
+if build_platform == 'macos'
+  sed_command_edit_in_place_option = '-i .bak'
+endif
 ext_test_template_prefix = 'unittest_tizen_'
 ext_test_template_str = ext_test_template_prefix + 'template.cc.in'
 ext_test_template = files (ext_test_template_str)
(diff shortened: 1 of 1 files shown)


anyj0527 (approved):
LGTM ✔

myungjoo (approved):
LGTM

tests/nnstreamer_filter_extensions_common/meson.build
```diff
@@ -48,6 +48,10 @@ if have_python3
 endif
 
 sed_command = find_program('sed', required: true)
+sed_command_edit_in_place_option = '-i'
+if build_platform == 'macos'
+  sed_command_edit_in_place_option = '-i .bak'
```
eli-schwartz, line 53: macOS sed can be told to edit in place without writing an additional backup file, using

```sh
sed -i '' s/foo/bar/ file.txt
```

tests/nnstreamer_filter_extensions_common/meson.build
```diff
@@ -72,9 +76,9 @@ foreach ext : extensions
     input : ext_test_template,
     output : ext_test_path_each,
     command : [copy, '-f', '@INPUT@', '@OUTPUT@', \
-      '&&', 'sed', '-i', sed_ext_name_option, '@OUTPUT@', \
-      '&&', 'sed', '-i', sed_ext_abbrv_option, '@OUTPUT@', \
-      '&&', 'sed', '-i', sed_ext_mf_option, '@OUTPUT@']
+      '&&', 'sed', sed_command_edit_in_place_option, sed_ext_name_option, '@OUTPUT@', \
+      '&&', 'sed', sed_command_edit_in_place_option, sed_ext_abbrv_option, '@OUTPUT@', \
+      '&&', 'sed', sed_command_edit_in_place_option, sed_ext_mf_option, '@OUTPUT@']
   )
```
eli-schwartz, line 82: This all looks kinda odd. For one thing, you can do multiple sed expressions in one invocation:

```sh
sed -i -e "s/foo/bar/" -e "s/other/thing/" file.txt
```

So here I'd expect to see something more like:
```meson
    command : [copy, '-f', '@INPUT@', '@OUTPUT@', \
      '&&', 'sed', '-i', '-e', sed_ext_name_option, \
      '-e', sed_ext_abbrv_option, \
      '-e', sed_ext_mf_option, '@OUTPUT@']
```

Also I'm surprised this works at all, since generally Meson doesn't document that you may use shell syntax here... huh...

One alternative to cp+sed is to write a python script that takes an input file, an output file, and a list of replacements and uses the `re` module on them.
Another alternative is to use Meson's `capture: true` option for custom_target, and use sed without -i; sed will take the input file, replace it, and emit the result to stdout, Meson will then capture the output and handle writing to the output file on its own.

Whatever the case may be, I'd definitely optimize these 3 seds into one. :)

PR #3773, opened 2022-05-31
[decoder/bounding_box] Fix ssd box decoding without postprocessing

jvuillaumier:

[decoder/bounding_box] Fix ssd box decoding without postprocessing

Boxes whose detection scores do not exceed the specified sigmoid
threshold shall never be reported as valid.

This changeset fixes current implementation whereby valid field
for a box is set only when its detection scores have passed the
threshold tests. Thus, if score test is not passed, valid field for
a box is inherited from the previous box score tests.

Therefore without this change, whenever one box passes the score
criteria, every subsequent boxes will be reported as valid, even though
their detection scores are low.
As a consequence, a very big number of boxes may be passed to the
next stage of NMS processing which causes processing overhead.

**Self evaluation:**
SSAT suite passed

**How to evaluate:**
Run a SSD MobileNet model without post processing and review results[] GArray content prior to executing nms()
There are many duplicate 'valid' boxes in there - GArray content is cleaned with this change. 
```


jaeyun-jung (approved):

myungjoo (approved):
LGTM!

PR #3790, opened 2022-06-09
Query: fix protocol property handling

myungjoo:
Protocols are enum. Use enum-param, not int or string.
With this, users can use both numbers and enum-strings.

PTAL: @jaeyun-jung @gichan-jang


myungjoo:
todo: update property values in test cases. it is no more case insensitive. i.e., tcp -> TCP

jaeyun-jung (approved):

gichan-jang (approved):
LGTM

PR #3792, opened 2022-06-09
[CI TC Fail Fix] test/query: parallel client launching shouldn't use gstTest @open sesame 6/10 10:12

myungjoo:
Don't use & fork with bash function.
Add timeout for launching.


niley7464 (approved):
LGTM 👍

anyj0527 (approved):
LGTM ✔ Great!

PR #3799, opened 2022-06-14
[spec] Fix spec to cut the depedency on custom tflite in 64-bit

anyj0527:
- In 64-bit the dep is like "libtensorflow2-lite-custom.so()(64bit)"
- To cut this properly, replace "so" with "*"

**Self evaluation:**


jaeyun-jung (approved):

again4you (approved):
Nice catch!

PR #3811, opened 2022-06-27
Fixes build breaks of 2.0.y on macOS

wooksong:
This PR fixes build breaks of lts/2.0.y for macOS. The cherry-picked patches from the upstream branch are as follows:

- tensor-query: cross-platform compatibility (non-Linux)
- [macOS/meson] Add a missing parameter to the sed's in-place edit option
- [Tests/IF] Fix a build error related to -Winconsistent-missing-override


jaeyun-jung (approved):

again4you:
Since this PR is backporting from the main branch, I will merge it without CI verification.

PR #3813, opened 2022-06-28
[Filter/TFLite] Fix model tensors allocation not working for some delegates

jvuillaumier:
Hello all,
This is a bugfix for an issue detected with TFLite delegate supporting dynamic shape.
Tensors are not allocated and inference can not run.
AllocateTensors() has to be called consistently before Invoke() even in case of delegate usage.
That is the case in TFLite examples code.
Thanks

[Filter/TFLite] Fix model tensors allocation not working for some delegates

This change is correction for error below reported by the TFlite library
during Invoke():

"Invoke called on model that is not ready."

Tensors allocation may be done by TFLite framework during call to
ModifyGraphWithDelegate() but that is not guaranteed.
Therefore, even if delegates are used, tensors allocation must be done
explicitly with AllocateTensors(), prior to inferencing with Invoke().

**Self evaluation:**
2. Run test: [ *]Passed [ ]Failed []Skipped

SSAT  test suite is passed.


jaeyun-jung (approved):

anyj0527 (approved):
LGTM, Thanks!

Issue #3800, opened 2022-06-15
Increase NNS_TENSOR_RANK_LIMIT, the rank of tensor


anyj0527:
NNS_TENSOR_RANK_LIMIT is currently 4. We may increase it.
Ops like conv3d requires 5-D tensors. And widely used yolov5 pytorch model has 5-D output.

REF:
ONE has rank limit 6.
TFLite seems does not have limit, but its delegates have (xnnpack: 6, gpu: 4, ...)

---
[Added by @myungjoo]

- Development Target: 2022-4Q
- Release Target: 2023-1Q (with Tizen 7.5M1 ACR & GBM Pre-release)
- Priority: Medium
- Functional Requirements
    - RANK_LIMIT = 8 or 16
    - Support all nnstreamer elements
- Non-Functional Requirements
    - Backward compatibility (all previous nnstreamer pipelines should be kept working)
    - Performance & resource-efficiency shouldn't drop
    - Minimize code changes
    - (Recommended) no more hardcoded RANK_LIMIT logic (c.f., tensor\_transform)
- CAUTION
    - tensor\_transform may require a lot of changes
    - be careful with flexible/sparse tensors <-- --> static tensors
    - tensor\_aggregator may require a bit of changes
- You will need to add a lot of test cases to ensure backward compatibility and RANK>4 behaviors.

myungjoo:
If you are going to increase it, let's try 8. :)

We may let each tensor\_* element declare its supported RANK.
Most elements will be trivial for this change; however, ```tensor_transform``` may suffer from a lot of changes and you'll need to delay its changes for RANK>4.

Anyway, as long as we now have models with RANK>4, we need to start this (w/ priority after nnstreamer-edge(+RT) and ml-service-api/Phase2&3).

CC: @jaeyun-jung @again4you @gichan-jang @niley7464

jaeyun-jung:
Let's discuss below items with rank change,
1. dynamic shape (3:?:?:1)
2. scalar (model rank is 0)

myungjoo:
![image](https://user-images.githubusercontent.com/1109011/173759573-42caed30-9232-4df1-8c1b-4b7902c15632.png)

myungjoo:
I think it's done along with test cases, right? @niley7464 
Please close this if you have enough test cases registered. (if you have #negativecases > #positivacase > N > 0 and code-line-coverage > 80%, major-function-coverage == 100% , it's enough.)

[...]

PR #3814, opened 2022-06-29
[Filter] support NNS_TENSOR_RANK_LIMIT 8

niley7464:
This patch changes filter pytorch and tensorflow lites to support NNS_TENSOR_RANK_LIMIT 8

tensor_filter_pytorch's invoke does not have a problem when NNS_TENSOR_RANK_LIMIT is 4.
But when rank limit is 8, model->invoke has an exception while running the model.

Similarly, tensor_filter_tensorflow_lite's tflite_setInputDim has a problem.
TFLiteInterpreter::setInputTensorsInfo now iterates all possible ranks starting from MAX.
If NNS_TENSOR_RANK_LIMIT changed to 8, there is a case which ResizeInputTensor works while AllocateTensors does not work.
So I chagned to iterate from MIN, and Allocate Tensors inside for loop.
related: #3800 

ToDo
- Implement dimchg(tensor_transform) with NNS_TENSOR_RANK_LIMIT 8 and create test data
- Find reason why repo_* have errors when rank limit is 8
- Testcode fix ( add :1:1:1:1 behind to the hardcoded dimension ) -> almost done

Question 
- tensor_transform's transpose are fixed to rank 3. Do I need to modify this limitation?

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
@@ -63,7 +63,7 @@ class TorchCore
   const char *getModelPath ();
   int getInputTensorDim (GstTensorsInfo *info);
   int getOutputTensorDim (GstTensorsInfo *info);
-  int invoke (const GstTensorMemory *input, GstTensorMemory *output);
+  int invoke (const GstTensorFilterProperties *prop, const GstTensorMemory *input, GstTensorMemory *output);
 
   private:
   char *model_path;
@@ -495,7 +495,7 @@ TorchCore::serializeOutput (
  *         -4 if running the model failed.
  */
 int
-TorchCore::invoke (const GstTensorMemory *input, GstTensorMemory *output)
+TorchCore::invoke (const GstTensorFilterProperties *prop, const GstTensorMemory *input, GstTensorMemory *output)
 {
 #if (DBG)
   gint64 start_time = g_get_real_time ();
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -680,7 +680,7 @@ torch_invoke (const GstTensorFilterProperties *prop, void **private_data,
   g_return_val_if_fail (core && input && output, -EINVAL);
   UNUSED (prop);
```
anyj0527, line 681: This line can be removed.

myungjoo (approved):
LGTM.
Please make sure that there is no tensor_filter subplugins not tested or validated. We may have a f ew filters not built or tested in the CI (e.g., NXP's subplugins)

ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
niley7464: Fixed! 
Thank you :)

anyj0527 (approved):
LGTM👍

> tensor_transform's transpose are fixed to rank 3. Do I need to modify this limitation?

@myungjoo may answer this?

myungjoo:
> LGTM👍
> 
> > tensor_transform's transpose are fixed to rank 3. Do I need to modify this limitation?
> 
> @myungjoo may answer this?

Not in this PR. But, yes, someday (e.g., 2023?), it should support rank N.

PR #3818, opened 2022-06-30
[Tests/Meson] Copy bmp2png to the tests directory every build time @open sesame 07/05 10:15

wooksong:
In order to enforce to keep ./tests/bmp2png the same as the one in the build directory, this patch modifies the meson build script to copy bmp2png to the tests directory from the build directory every build time.

**Self evaluation:**

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -18,8 +18,10 @@ if libpng_dep.found()
   custom_target('copy-bmp2png',
     input: b2p,
     output: 'b2p',
-    command: [copy, '@INPUT@', meson.current_source_dir()],
-    build_by_default: true
+    depends: b2p,
+    command: [copy, '@INPUT@', '@CURRENT_SOURCE_DIR@'],
+    build_by_default: true,
+    build_always_stale: true,
   )
 endif
 


eli-schwartz:
Maybe it would be possible to somehow tell the rest of the testsuite to locate the executable in the build directory instead? Perhaps by passing its path as an argument to whatever uses it?

myungjoo:
> Maybe it would be possible to somehow tell the rest of the testsuite to locate the executable in the build directory instead? Perhaps by passing its path as an argument to whatever uses it?

This is due to incorrect design of earlier days when test suite (SSAT) was being detached from nnstreamer. No code or test cases uses bmp2png in nnstreamer. The test suite (ssat) uses it. bmp2png should be migrated to SSAT out of nnstreamer later. Thus, no need to clean this up in nnstreamer. I'll remove bmp2png from nnstreamer when it's ready.

wooksong:
This is a Q related to https://github.com/myungjoo/SSAT/pull/23. Do I need to drop this PR? If so, I will revise the whole test case based on the new SSAT. 


Or, we can just go gradually.

myungjoo:
> This is a Q related to [myungjoo/SSAT#23](https://github.com/myungjoo/SSAT/pull/23). Do I need to drop this PR? If so, I will revise the whole test case based on the new SSAT.
> 
> Or, we can just go gradually.

Let's do it gradually. Loading up SSAT updates take some time in the platforms.

myungjoo (approved):

again4you (approved):
LGTM!

PR #3820, opened 2022-07-01
[Tests/Query] Revise runTest.sh to support macOS

wooksong:
grep in macOS does not have -P, --perl-regexp. In addition, the stat's option to specify a format also differs from that of Linux distros. This patch revises runTest.sh to support macOS by applying the above differences.

**Self evaluation:**

### DIFF
--- a/tests/nnstreamer_query/runTest.sh
+++ b/tests/nnstreamer_query/runTest.sh
@@ -20,11 +20,13 @@ testInit $1
 
 # Skip query test temporarily because of ppa build failure.
 # Related issue: https://github.com/nnstreamer/nnstreamer/issues/3657
-ID=$(grep -oP '(?<=^ID=).+' /etc/os-release | tr -d '"')
-if [ "$ID" = "ubuntu" ]; then
-    echo "Skip query test for ppa build"
-    report
-    exit
+if [ -f /etc/os-release ]; then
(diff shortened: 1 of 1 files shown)


tests/nnstreamer_query/runTest.sh
```diff
@@ -45,7 +47,7 @@ function waitformarker {
   for i in $(seq 1 ${TIMEOUT_SEC})
   do
     if [ -f 'marker.log' ]; then
-      markersize=$(stat -c%s marker.log)
+      markersize=$([[ ${SO_EXT} == "dylib" ]] && stat -f%z marker.log || stat -c%s marker.log)
```
myungjoo, line 50: There's similar script at https://github.com/nnstreamer/nnstreamer/blob/main/tests/nnstreamer_grpc/runTest.sh

anyj0527 (approved):
LGTM ✔

tests/nnstreamer_query/runTest.sh
wooksong: @myungjoo We've already defined ```StatCmd_GetSize``` in ```ssat-api.sh```. Thus, I revised this PR as well as #3825 to use ```StatCmd_GetSize``` .

myungjoo (approved):

PR #3838, opened 2022-07-14
[TEST/query] Do not use fixed port number

gichan-jang:
Do not use fixed port number.
Let's get an available port number so that multiple tests do not use the same port number.

**Self evaluation:**


niley7464 (approved):
Looks good to me

myungjoo (approved):
LGTM

PR #3863, opened 2022-08-02
[Gst/Edge] Add pipeline tests @open sesame 8/8 17:23

anychhoice:
This PR will be improved during the development of edgesrc and edgesink.

Changes:
- Init edgesrc and edgesink unit test.
- Add edgesrc and edgesink pipeline tests.

Signed-off-by: Yechan Choi <yechan9.choi@samsung.com>

**Self evaluation:**

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -145,15 +145,17 @@ if gtest_dep.found()
     test('unittest_query', unittest_query, env: testenv)
 
     # Run unittest_edge
-    unittest_edge = executable('unittest_edge',
-      join_paths('nnstreamer_edge', 'unittest_edge.cc'),
-      dependencies: [nnstreamer_unittest_deps],
-      install: get_option('install-test'),
-      install_dir: unittest_install_dir
-    )
+    if nnstreamer_edge_support_is_available
+      unittest_edge = executable('unittest_edge',
+        join_paths('nnstreamer_edge', 'unittest_edge.cc'),
+        dependencies: [nnstreamer_unittest_deps],
+        install: get_option('install-test'),
+        install_dir: unittest_install_dir
+      )
 
-    test('unittest_edge', unittest_edge, env: testenv)
-
+      test('unittest_edge', unittest_edge, env: testenv)
+    endif
+    
     # Run unittest_mqtt
     if mqtt_support_is_available
       unittest_mqtt_w_helper = executable('unittest_mqtt_w_helper',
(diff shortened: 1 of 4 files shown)


tests/nnstreamer_edge/unittest_edge.cc
```diff
@@ -0,0 +1,44 @@
+/**
```
myungjoo, line 1: Add a SPDX license identifier.

tests/meson.build
```diff
@@ -144,6 +144,16 @@ if gtest_dep.found()
 
     test('unittest_query', unittest_query, env: testenv)
 
+    # Run unittest_edge
+    unittest_edge = executable('unittest_edge',
+      join_paths('nnstreamer_edge', 'unittest_edge.cc'),
+      dependencies: [nnstreamer_unittest_deps],
+      install: get_option('install-test'),
+      install_dir: unittest_install_dir
+    )
```
myungjoo, line 153: You will need a ```if ... endif``` for unitest_edge so that it's activated only when "edge" feature is enabled. (refer to meson_options.txt and its usages in meson.build)

tests/nnstreamer_edge/unittest_edge.cc
anychhoice: Updated.

tests/meson.build
anychhoice: Updated.

tests/nnstreamer_edge/runTest.sh
```diff
@@ -0,0 +1,146 @@
...
+
+## @brief Execute file comparison test if both files exist
+function callCompareTestIfExist() {
+    if [[ ! -f "$1" ]]; then
+        echo "$1 don't exist. Skip $3($4)."
+    fi
+
+    if [[ ! -f "$2" ]]; then
+        echo "$2 don't exist. Skip $3($4)."
+    fi
```
gichan-jang, line 40: It's not skipping the tests.
```suggestion
    if [[ ! -f "$1" ]]; then
        echo "$1 don't exist. Skip $3($4)."
        return
    fi

    if [[ ! -f "$2" ]]; then
        echo "$2 don't exist. Skip $3($4)."
        return
    fi
```

anychhoice: Updated.

gichan-jang (approved):

myungjoo (approved):
LGTM

PR #3870, opened 2022-08-09
tensor-decoder: add ssd_mobilenet v3 support

ksghaier:
ssd_mobilenet v3 models handle 2034 anchor boxes 
while previous models v2 and v1 handles only 1917 boxes
thus updating the MAX value to 2034 adds support for v3 models
this update is needed for models without post-processing layer.


myungjoo (approved):
LGTM

niley7464 (approved):
Looks good to me :)

PR #3874, opened 2022-08-11
[Query] fix res leak case @open sesame 08/11 21:26

jaeyun-jung:
Fix res leak in query elements.
Add destroy-cb function in query-server table to prevent mem leak.
Prevent duplicated creation of edge-handle in event callback.

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_client.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_client.c
@@ -181,6 +181,7 @@ gst_tensor_query_client_init (GstTensorQueryClient * self)
   self->dest_port = TCP_DEFAULT_SRV_SRC_PORT;
   self->topic = NULL;
   self->in_caps_str = NULL;
+  self->edge_h = NULL;
   self->msg_queue = g_async_queue_new ();
 }
 
@@ -195,6 +196,8 @@ gst_tensor_query_client_finalize (GObject * object)
 
   g_free (self->host);
   self->host = NULL;
+  g_free (self->dest_host);
+  self->dest_host = NULL;
   g_free (self->topic);
   self->topic = NULL;
   g_free (self->in_caps_str);
@@ -226,6 +229,7 @@ gst_tensor_query_client_set_property (GObject * object, guint prop_id,
 {
   GstTensorQueryClient *self = GST_TENSOR_QUERY_CLIENT (object);
 
+  /** @todo DO NOT update properties (host, port, ..) while pipeline is running. */
   switch (prop_id) {
     case PROP_HOST:
       if (!g_value_get_string (value)) {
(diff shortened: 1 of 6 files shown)


gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -478,6 +481,36 @@ _nns_edge_event_cb (nns_edge_event_h event_h, void *user_data)
...
+    nns_edge_set_event_callback (self->edge_h, _nns_edge_event_cb, self);
+
+    if (self->topic)
+      nns_edge_set_info (self->edge_h, "TOPIC", self->topic);
+    if (self->host)
+      nns_edge_set_info (self->edge_h, "HOST", self->host);
+    if (self->port > 0) {
+      gchar *port = g_strdup_printf ("%u", self->port);
+      nns_edge_set_info (self->edge_h, "PORT", port);
+      g_free (port);
```
myungjoo, line 507: Is it guaranteed that topic, host, and port will be never changed between init and finalize?

gst/nnstreamer/tensor_query/tensor_query_server.c
```diff
@@ -31,11 +31,33 @@ static GHashTable *_qs_table = NULL;
...
+
+  if (_data->edge_h) {
+    nns_edge_release_handle (_data->edge_h);
+    _data->edge_h = NULL;
+  }
+
+  g_cond_clear (&_data->cond);
+  g_mutex_clear (&_data->lock);
+  g_free (_data->id);
+  g_free (_data);
```
myungjoo, line 53: What if someone else is holding ```_data->lock``` while this function is called?

jaeyun-jung: updated

gst/nnstreamer/tensor_query/tensor_query_client.c
jaeyun-jung: I updated code to compare old caps string when received gst-caps event.

@gichan-jang @myungjoo 
Discussion: Now it does not guarantee the change of host and port after starting the pipeline.
(If it is necessary, we need to consider adding state-change callback and closing edge handle if pipeline state is changed from 'pause' to 'ready'.

myungjoo: You can forbid changing host/port while it is active. (and make it allowed if there is a need :) )

gst/nnstreamer/tensor_query/tensor_query_server.c
myungjoo: It would be safer to do ```g_mutex_unlock``` after ```g_free (_data)```. You can add ```g_mutex *tmp = &_data->lock;``` and control the mutex with ```tmp```.

jaeyun-jung: updated

gst/nnstreamer/tensor_query/tensor_query_client.c
jaeyun-jung: Marked todo to prevent changing the properties while pipeline is running.

gst/nnstreamer/tensor_query/tensor_query_server.c
myungjoo: uh...

```
g_cond_clear ()
void                g_cond_clear                        (GCond *cond);
Frees the resources allocated to a GCond with g_cond_init().

This function should not be used with a GCond that has been statically allocated.

Calling g_cond_clear() for a GCond on which threads are blocking leads to undefined behaviour.

cond :

an initialised GCond
```


looks like a trouble for g_cond_clear() here...

jaeyun-jung: rollbacked, _release_server_data() will be called with table-lock.

myungjoo (approved):

Issue #3873, opened 2022-08-11
[android-resource] File removal request.


myungjoo:
Related: https://lfscanning.org/reports/lfai/nnstreamer-2022-07-02-8eb54f0b-f36e-43ff-826f-80e9f251e17e.html

Please remove the following header files from archives in nnstreamer-android-resource.git

- [x] # 6
- [ ] # 7
```
Finding #6
--
Priority: High
These files are under GPL-3.0+ and MUST BE REMOVED FROM THE REPOSITORY.
4 files (hide files)
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/tensorflow-lite-2.7.0.tar.xz/tensorflow-lite/include/tensorflow/lite/examples/minimal/build/eigen/unsupported/test/mpreal/mpreal.h
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/tensorflow-lite-2.7.0.tar.xz/tensorflow-lite/include/tensorflow/lite/tools/make/downloads/eigen/unsupported/test/mpreal/mpreal.h
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/tensorflow-lite-2.8.1.tar.xz/tensorflow-lite/include/tensorflow/lite/examples/minimal/build/eigen/unsupported/test/mpreal/mpreal.h
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/tensorflow-lite-2.8.1.tar.xz/tensorflow-lite/include/tensorflow/lite/tools/make/downloads/eigen/unsupported/test/mpreal/mpreal.h
```

```
Finding #7
--
Priority: High
These files are under the GPL with GCC Runtime Exception license, which applies when using the GCC compiler. Since you can't guarentee that anyone will be using GCC you should remove these files.
4 files (hide files)
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/pytorch-1.10.1.tar.xz/pytorch/include/c10/util/Array.h
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/pytorch-1.10.1.tar.xz/pytorch/include/c10/util/reverse_iterator.h
nnstreamer-2022-07-02.zip/nnstreamer-android-resource/external/pytorch-1.8.0.tar.xz/pytorch/include/c10/util/Array.h
[...]

PR #3879, opened 2022-08-16
filter:pytorch allow gcc only.

myungjoo:
Do not allow to build with other compilers.
Refer to #3873 and https://lfscanning.org/reports/lfai/nnstreamer-2022-07-02-8eb54f0b-f36e-43ff-826f-80e9f251e17e.html

This is to comply with GPL-3.0 GCC Runtime Exception License of a few headers in PyTorch.


jaeyun-jung (approved):

anyj0527 (approved):
LGTM✔ Thanks

PR #3881, opened 2022-08-16
test/query: revive pid retrieval

myungjoo:
In recent commits, pid assignment after invoking background processes
has disappeared. Revivde pid assignment so that we can kill unnecessary
processes and avoid indefinite waits.


anyj0527 (approved):
LGTM

jaeyun-jung:
need rollback GC's PR (block hybrid testcases)?

gichan-jang:
grpc/runTest.sh also uses `gstTestBackground`.

myungjoo:
grpc test cases updated.

myungjoo:
> need rollback GC's PR (block hybrid testcases)?

Don't need any rollbacks.

jaeyun-jung (approved):

PR #3886, opened 2022-08-18
[test/tvm] Add tvm ssat test for image labeling @open sesame 11/16 11:00

ksghaier:
This commit adds a compiled tvm model and two gst test commands

Sample model (mobilenet_v1_0.75_224_{arch}.so) is compiled from public
tflite model as follow:
- Get the model from:
    http://download.tensorflow.org/models/mobilenet_v1_2018_08_02/mobilenet_v1_0.75_224_quant.tgz
- Follow the steps described in:
    https://tvm.apache.org/docs/how_to/compile_models/from_tflite.html

---
# [Template] PR Description

In general, github system automatically copies your commit message for your convenience.
Please remove unused part of the template after writing your own PR description with this template.
```bash
$ git commit -s filename1 filename2 ... [enter]

Summarize changes in around 50 characters or less

More detailed explanatory text, if necessary. Wrap it to about 72
characters or so. In some contexts, the first line is treated as the
subject of the commit and the rest of the text as the body. The
blank line separating the summary from the body is critical;
various tools like `log`, `shortlog` and `rebase` can get confused 
if you run the two together.

Further paragraphs come after blank lines.

- Bullet points are okay, too
- Typically a hyphen or asterisk is used for the bullet, preceded
  by a single space, with blank lines in between, but conventions vary here.

Resolves: #123
See also: #456, #789

**Self evaluation:**

**How to evaluate:**
1. Describe how to evaluate in order to be reproduced by reviewer(s).

Add signed-off message automatically by running **$git commit -s ...** command.

$ git push origin <your_branch_name>
```

### DIFF
--- a/tests/nnstreamer_filter_tvm/runTest.sh
+++ b/tests/nnstreamer_filter_tvm/runTest.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+##
+## SPDX-License-Identifier: LGPL-2.1-only
+##
+## @file runTest.sh
+## Copyright 2022 NXP
+## @brief SSAT Test Cases for NNStreamer
+##
+if [[ "$SSATAPILOADED" != "1" ]]; then
+    SILENT=0
+    INDEPENDENT=1
+    search="ssat-api.sh"
+    source $search
+    printf "${Blue}Independent Mode${NC}"
+fi
+
+# This is compatible with SSAT (https://github.com/myungjoo/SSAT)
+testInit $1
+
+# Check if architecture is supported
+SUPPORTED_ARCHS="aarch64 armv7l x86_64"
+ARCH=`uname -m`
+echo ${SUPPORTED_ARCHS} | grep -w -q ${ARCH}
+if [ $? -ne 0 ]; then
+  echo "Architecture is not supported - ${ARCH}"
+  report
+  exit
+fi
+
+# NNStreamer and plugins path for test
+PATH_TO_PLUGIN="../../build"
+
+if [[ -d $PATH_TO_PLUGIN ]]; then
+    ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
+    if [[ -d ${ini_path} ]]; then
+        check=$(ls ${ini_path} | grep tvm.so)
+        if [[ ! $check ]]; then
+            echo "Cannot find tvm shared lib"
+            report
+            exit
+        fi
+    else
+        echo "Cannot find ${ini_path}"
+    fi
+else
+    ini_file="/etc/nnstreamer.ini"
+    if [[ -f ${ini_file} ]]; then
+        path=$(grep "^filters" ${ini_file})
+        key=${path%=*}
+        value=${path##*=}
+
+        if [[ $key != "filters" ]]; then
+            echo "String Error"
+            report
+            exit
+        fi
+
+        if [[ -d ${value} ]]; then
+            check=$(ls ${value} | grep tvm.so)
+            if [[ ! $check ]]; then
+                echo "Cannot find tvm shared lib"
+                report
+                exit
+            fi
+        else
+            echo "Cannot file ${value}"
+            report
+            exit
+        fi
+    else
+        echo "Cannot identify nnstreamer.ini"
+        report
+        exit
+    fi
+fi
+
+PATH_TO_MODEL="../test_models/models/mobilenet_v1_0.75_224_quant_${ARCH}.so"
+PATH_TO_LABEL="../test_models/labels/labels.txt"
+PATH_TO_IMAGE="../test_models/data/orange.png"
+PATH_TO_CLASS="class.out.log"
+
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! video/x-raw,format=RGB,framerate=0/1 ! tensor_converter ! tensor_filter framework=tvm model=${PATH_TO_MODEL} custom=device:CPU,num_input_tensors:1 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=${PATH_TO_CLASS}" 1 0 0 $PERFORMANCE
+
+class=`cat ${PATH_TO_CLASS}`
+[ "$class" = "orange" ]
+testResult $? 1 "Golden test comparison" 0 1
+
+# Negative test: wrong input type float32 instead of uint8
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,div:127.5,add:-1.0 ! tensor_filter framework=tvm model=${PATH_TO_MODEL} custom=device:CPU,num_input_tensors:1 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=${PATH_TO_CLASS} " 1_n 0 1 $PERFORMANCE
+
+# Cleanup
+rm *.log
+
+report


tests/nnstreamer_filter_tvm/runTest.sh
```diff
@@ -0,0 +1,84 @@
...
+        exit
+    fi
+fi
+
+PATH_TO_MODEL="../test_models/models/mobilenet_v1_0.75_224_quant_$(arch).so"
+PATH_TO_LABEL="../test_models/labels/labels.txt"
+PATH_TO_IMAGE="../test_models/data/orange.png"
+PATH_TO_CLASS="class.out.log"
+
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! video/x-raw,format=RGB,framerate=0/1 ! tensor_converter ! tensor_filter framework=tvm model=${PATH_TO_MODEL} custom=device=CPU,num_input_tensors:1 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=${PATH_TO_CLASS}" 1 0 0 $PERFORMANCE
```
myungjoo, line 72: custom=device=CPU,...
==>
custom=device:CPU...

tests/nnstreamer_filter_tvm/runTest.sh
```diff
@@ -0,0 +1,84 @@
...
+
+class=`cat ${PATH_TO_CLASS}`
+[ "$class" = "orange" ]
+testResult $? 1 "Golden test comparison" 0 1
+
+# Negative test: wrong input type float32 instead of uint8
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,div:127.5,add:-1.0 ! tensor_filter framework=tvm model=${PATH_TO_MODEL} custom=device=CPU,num_input_tensors:1 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=${PATH_TO_CLASS} " 1_n 0 1 $PERFORMANCE
```
myungjoo, line 79: custom=device=CPU,...
==>
custom=device:CPU...

tests/nnstreamer_filter_tvm/runTest.sh
```diff
@@ -0,0 +1,84 @@
...
+            report
+            exit
+        fi
+    else
+        echo "Cannot identify nnstreamer.ini"
+        report
+        exit
+    fi
+fi
+
```
myungjoo, line 66: Please skip tests if $(arch) is not supported.

myungjoo:
If you can supply lighter models for testing, it'd be great.
However, if you are going to keep them, please ignore ```TAOS/pr-prebuild-filesize — Oooops. File size checker is failed at mobilenet_v1_0.75_224_quant_aarch64```.

myungjoo:
```
[  354s] [Starting] nnstreamer_filter_tvm
[  354s] /home/abuild/rpmbuild/BUILD/nnstreamer-2.1.1/tests/nnstreamer_filter_tvm/runTest.sh: line 67: arch: command not found
[  354s] ==================================================
[  354s]     Test Group nnstreamer_filter_tvm Starts.
[  354s] [FAILED][Critical] 1:gst-launch of case 1
[  354s] [FAILED][Critical] 1:Golden test comparison
[  354s] [PASSED] 1_n:gst-launch of case 1_n
[  354s] ==================================================
[  354s] [FAILED] Test Group nnstreamer_filter_tvm has failed cases (2)
```
Looks like $(arch) is often not supported in a few Linux distro.

tests/nnstreamer_filter_tvm/runTest.sh
ksghaier: updated, thanks

tests/nnstreamer_filter_tvm/runTest.sh
ksghaier: updated, thanks

tests/nnstreamer_filter_tvm/runTest.sh
ksghaier: updated

ksghaier:
> S/pr-prebuild-filesi

unfortunately I have no smaller compiled model than this one right now

myungjoo (approved):
LGTM

myungjoo:
Unfortunately, there are failed unit tests (TVM tests. Both in x64 arch. Maybe the x64 binaries have some problems?)
```
[  264s] [Starting] nnstreamer_filter_tvm
[  264s] ==================================================
[  264s]     Test Group nnstreamer_filter_tvm Starts.
[  264s] [FAILED][Critical] 1:gst-launch of case 1
[  264s] [FAILED][Critical] 1:Golden test comparison
[  264s] [PASSED] 1_n:gst-launch of case 1_n
[  264s] ==================================================
[  264s] [FAILED] Test Group nnstreamer_filter_tvm has failed cases (2)
[  264s] 
````

**Comment by anyj0527** [comment]
> Unfortunately, there are failed unit tests (TVM tests. Both in x64 arch. Maybe the x64 binary has some problems?)

May check this: https://github.com/nnstreamer/nnstreamer/pull/3412#issue-946185613
AFAIK, the existing test models (tvm_add_one_arch.so_) are created with it.

**Comment by anyj0527** [comment]
@ksghaier @myungjoo
The added testcase get runtime error with current Tizen/Debian tvm package (it's v0.7.0).
However, with the recent tvm release (v0.9.0) and slight modification in our source, I verified those TCs are passed in x86_64 ubuntu local env.
I think after the new TIZEN release in this month, tvm packages should be upgraded. Then this PR could be merged.

**Comment by anyj0527** [comment]
> > @ksghaier @myungjoo The added testcase get runtime error with current Tizen/Debian tvm package (it's v0.7.0). However, with the recent tvm release (v0.9.0) and slight modification in our source, I verified those TCs are passed in x86_64 ubuntu local env. I think after the new TIZEN release in this month, tvm packages should be upgraded. Then this PR could be merged.
> 
> @anyj0527 PTAL if it's now ok.
Tizen tvm package need upgrade (0.7.0 -> 0.9+). I'll make a new issue for this.

**Comment by myungjoo** [comment]

## Review — [test/tvm] Add tvm ssat test for image labeling

The test itself is well-constructed: a golden image-labeling comparison plus a negative type-mismatch case, an architecture guard, and the same ini/plugin-discovery preamble the other SSAT suites use. The gap it fills is real — `tests/nnstreamer_filter_tvm/` currently has only a gtest, no pipeline-level test.

The blocking concerns are about the three binary artifacts, not the script.

### 1. (Critical) 16.7 MB of unverifiable prebuilt native libraries — provenance required — cc @myungjoo

This PR adds three prebuilt, executable shared objects:

| File | Size |
| --- | --- |
| `tests/test_models/models/mobilenet_v1_0.75_224_quant_x86_64.so` | 5,800,904 B |
| `tests/test_models/models/mobilenet_v1_0.75_224_quant_aarch64.so` | 5,432,336 B |
| `tests/test_models/models/mobilenet_v1_0.75_224_quant_armv7l.so` | 5,508,020 B |

These are not data files. A TVM `.so` module is native code that gets `dlopen`ed and executed in-process by the TVM runtime. And they *will* execute in the project's build infrastructure: `packaging/nnstreamer.spec:1079-1092` runs `ssat -n -p=1` from `tests/`, which recursively discovers and executes every `runTest.sh` — including this new one — on x86_64, aarch64 and armv7l Tizen builders.

I want to be explicit that I found **no evidence of anything malicious**, and the contributor has given a clear recipe in the PR description (public MobileNet tflite + the standard `from_tflite.html` TVM flow). The problem is structural: the recipe is not reproducible from what is in the repo, so nobody reviewing this PR — now or in five years — can confirm that these 16.7 MB of machine code correspond to that recipe. A compiler-emitted blob is the one artifact type where "looks fine in the diff" is meaningless, because the diff shows nothing at all. That makes this a supply-chain surface that the project cannot audit, and it lands permanently in git history where it cannot be removed without a rewrite.

Requests, in decreasing order of preference:

1. **Do not commit the binaries.** Add a small build/fetch script under `tests/test_models/` that compiles the model from the published tflite file (with a pinned TVM version and a pinned SHA256 of the source tflite), and have `runTest.sh` skip cleanly when the artifact is absent. This is what makes the test auditable *and* keeps the repo small.
2. If the binaries must be committed, publish them as a release asset or in a separate models repository, referenced by URL + SHA256, rather than in this tree.
3. At minimum: record in the PR the exact TVM version, target triple/`-mcpu` flags, host OS, and the SHA256 of both the input tflite and each output `.so`, and add a `README` next to the files carrying that information plus the full command line.

@myungjoo — flagging for your explicit sign-off per the review policy on unverifiable executable content entering the tree. This is a provenance/auditability call, not a finding of malicious code.

### 2. (Major) The test is tied to a TVM ABI with no version guard

Per the discussion in this thread, these modules fail on the TVM 0.7.0 packages shipped by Tizen/Debian and need 0.9+. A compiled TVM module embeds assumptions about the runtime it was built against, so this test will break again on the next TVM major bump — and it will break as a *test failure in the Tizen build*, i.e. it will block unrelated PRs.

Please add a runtime-version guard to `runTest.sh` alongside the existing architecture guard, so an incompatible TVM causes a clean skip (`report; exit`) rather than a red build. The architecture guard at the top is the right pattern to copy.

### 3. (Minor) `rm *.log` is too broad

```sh
rm *.log
```

This deletes every `.log` in the current directory, not just the one the script created (`class.out.log`). SSAT itself writes logs into the test directory. Please narrow it to `rm -f ${PATH_TO_CLASS}` (and note that an unqualified `rm *.log` fails the script if no match exists, since there is no `-f`).

### 4. (Minor) The negative case may not be negative

```sh
# Negative test: wrong input type float32 instead of uint8
gstTest "... tensor_transform mode=arithmetic option=typecast:float32,div:127.5,add:-1.0 ! tensor_filter framework=tvm ..." 1_n 0 1 $PERFORMANCE
```

The `1_n` case relies on the pipeline failing when a float32 tensor meets a uint8 model input. Whether that fails at caps negotiation or is silently accepted depends on the TVM subplugin's `getModelInfo()` behavior — please confirm it actually fails for the reason intended, rather than for an unrelated one (e.g. the missing `videoscale`/`video/x-raw` capsfilter that the positive case has but this one does not, which would make it fail on dimensions rather than on type). A negative test that passes for the wrong reason is worse than none.

Also note this case is directly affected by #4853, which changes tensor_filter invoke failures from `GST_FLOW_ERROR` to a dropped buffer; if `1_n` currently fails at invoke time rather than at negotiation, it will stop being negative once that lands.

### Other checks

- **Does it solve the stated topic?** Yes — golden image-labeling coverage for the TVM subplugin, which did not exist.
- **Regression detection value:** good. It shares `orange.png` and `labels.txt` with the other filter suites, so a change to the shared test data or to `tensor_decoder mode=image_labeling` fails several suites at once. That is the property you want.
- **Will CI actually gate on it?** Yes in Tizen GBS builds (via the recursive `ssat` run above); no in the GitHub Actions workflows, which only invoke `ssat` for a small explicit list of directories. Worth deciding whether it should be added to `ubuntu_clean_meson_build.yml` too, given TVM is available there.
- **Scope / size:** the script is appropriately sized; the 16.7 MB of binaries is the entirety of the size problem.
- **API/ARCHITECTURE:** no API change; no documentation update needed.

myungjoo:
Superseded by #4873.

#4873 adds the SSAT suite this PR was after, but reuses the in-tree `tvm_add_one_*.so_` models instead of adding new prebuilt ones. That sidesteps both blockers this PR ran into: the 16.7MB of prebuilt modules whose provenance could not be verified from the tree, and the tvm runtime version mismatch that stalled this in 2022. A runtime too old to load those in-tree models would already be failing `unittest_filter_tvm`, so there is no version guard to get wrong.

@ksghaier thank you for raising the gap and for the negative-test idea — both carried over. This PR will be closed once #4873 merges; keeping it open until then so the history stays linked.

PR #3890, opened 2022-08-23
[Query/Test] Set listener port for query client

gichan-jang:
tensor client of TEST 5-2 and 5-3 use same default port for listener.
Set random available port for query client.

**Self evaluation:**


jaeyun-jung (approved):
need update of nns-edge lib to handle host addr later

myungjoo (approved):

myungjoo:
The failed TCs are related with the nnstreamer-edge regression, which is to be fixed by awaiting SR.

PR #3896, opened 2022-08-31
[query] Add timeout prop for tensor_query_client

anyj0527:
- Add timeout prop to decide how much time to wait message from server.
- Default value is 0, no wait.
- Add simple unittest.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_client.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_client.c
@@ -44,6 +44,7 @@ enum
   PROP_DEST_PORT,
   PROP_CONNECT_TYPE,
   PROP_TOPIC,
+  PROP_TIMEOUT,
   PROP_SILENT,
 };
 
@@ -51,6 +52,7 @@ enum
 #define TCP_DEFAULT_HOST        "localhost"
 #define TCP_DEFAULT_SRV_SRC_PORT 3000
 #define TCP_DEFAULT_CLIENT_SRC_PORT 3001
+#define DEFAULT_CLIENT_TIMEOUT  0
 #define DEFAULT_SILENT TRUE
 
 GST_DEBUG_CATEGORY_STATIC (gst_tensor_query_client_debug);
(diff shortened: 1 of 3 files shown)


jaeyun-jung (approved):

again4you (approved):
LGTM!

gichan-jang (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -138,6 +140,12 @@ gst_tensor_query_client_class_init (GstTensorQueryClientClass * klass)
           "The main topic of the host.",
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
+  g_object_class_install_property (gobject_class, PROP_TIMEOUT,
+      g_param_spec_uint ("timeout", "timeout value",
+          "A timeout value (in ms) to wait message from server", 0,
```
myungjoo, line 145: What's the behavior if timeout == 0?

anyj0527: It there is no element in queue, do not wait and do not call `gst_pad_push`

myungjoo: Describe the behaviors for tensor_query_client users. What should they expect when they set timeout? Is the behavior kept the same as before if it's set 0? Usually, users will expect that if it's 0, it will wait for the server forever.

songgot (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_client.c
anyj0527: Chill, users won't see the word unless they look into code.
With default 0 timeout, it behaves same as before - It does not wait for the message from tensor_query_serversink to be queued.
@jaeyun-jung and I discussed about this, and we chose default 0 timeout to "no wait" and do not allow infinite wait.

myungjoo: What I'm telling you is that you need to describe it in ```g_param_spec_unit```. This is not an internal string for code readers. GStreamer users see these with gst-inspect, which is the default utility for gstreamer users. We have users who do not use ml-api, but use gstreamer api.

anyj0527: Updated the description as
"A timeout value (in ms) to wait message from query server after sending buffer to server. 0 means no wait."
PTAL

myungjoo (approved):

PR #3920, opened 2022-09-26
[Spec] dependency to nns-edge

jaeyun-jung:
Check dependency to nnstreamer-edge library.


anyj0527 (approved):
LGTM ✔

PR #3942, opened 2022-10-24
[Common] tensor format in struct

jaeyun-jung:
Move enum flag - tensor_format into tensors-info struct.

To pass flexible tensors to filter sub-plugins, we should indicate tensor format in tensors-info struct, not tensors-config. Internally tensors-config struct is used for caps negotiation.

TODO: need discussion, should we separate format in each tensor info, so that each mem chunk can express its format. (then, nnstreamer will combine memories: static+flexible in gst-buffer.)


gichan-jang:
> TODO: need discussion, should we separate format in each tensor info, so that each mem chunk can express its format. (then, nnstreamer will combine memories: static+flexible in gst-buffer.)

If each tensor has a format, I think it has a clear advantage in terms of performance.
It would be good to develop it when we have time to develop it or when there is a user request.

gichan-jang (approved):
LGTM

anyj0527 (approved):
LGTM✔

Issue #1146, opened 2019-02-21
Latency not supported with tensor elements


kparichay:
Tensor elements do not handle the `GST_QUERY_LATENCY`. 
This query is important with live sources, and not handling this query leads to frame drops with live source/sinks.

This needs all the elements to measure their latency/performance (depending on live source or not) and add it to the pipeline latency.

PR #3946, opened 2022-10-26
[filter/core] Add tensor filter latency reporting to the pipeline @open sesame 10/27 23:11

jvuillaumier:
Hello all,

This is a proposal for implementation of filter latency estimate and reporting in the pipeline.
It should help with #1146 for tensor filter element that is the main contributor to the latency.

Tensor filter executes inferences that may take long time depending on the model complexity and the compute capabilities of the platform. When inference duration is long, it happens that some downstream buffers are dropped because associated timestamp is too far in the past.

It results in a shaky pipeline buffer rate and for instance occurences of error messages from the sink when operated in live mode (sync=TRUE): 'There may be a timestamping problem, or this computer is too slow.'

Tensor filter implementation already has the option to computes latency estimates when its element property latency=1 is set.

This change introduces a new property value latency_report=1 that will when enabled use those latency duration estimates to:
- implement LATENCY query handling to include GstTensorFilter latency estimate in the overall pipeline latency computation
- track filter latency estimates updates to notify the bus through a LATENCY message that its latency has changed. It triggers a new round of LATENCY query in the pipeline to take into account the updated filter latency.

**Self evaluation:**
SSAT test suite passed

**How to evaluate:**
unittest_latency created.
Benefit can be seen in real pipeline typically when buildding pipeline with live source, a time-synchronized sync (sync=true) while having a tensor filter whose inference time is fairly long (like hundreds of msec)

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -105,6 +105,20 @@ static GstStaticPadTemplate src_factory = GST_STATIC_PAD_TEMPLATE ("src",
 #define gst_tensor_filter_parent_class parent_class
 G_DEFINE_TYPE (GstTensorFilter, gst_tensor_filter, GST_TYPE_BASE_TRANSFORM);
 
+/**
+ * @brief Headroom (extra duration) added to actual latency estimate reported
+ *        to LATENCY query, to limit number of updates when tracking the
+ *        maximum value - arbitrarily set to 5%.
+ */
+#define LATENCY_REPORT_HEADROOM 0.05
+
+/**
+ * @brief Threshold deciding when tracking latency estimate that current
+ *        value is sufficiently lower than reported value so that a
+ *        notification update is necessary - arbitrarily set to 25%.
+ */
+#define LATENCY_REPORT_THRESHOLD 0.25
+
 /* GObject vmethod implementations */
 static void gst_tensor_filter_set_property (GObject * object, guint prop_id,
     const GValue * value, GParamSpec * pspec);
@@ -121,6 +135,8 @@ static GstCaps *gst_tensor_filter_fixate_caps (GstBaseTransform * trans,
     GstPadDirection direction, GstCaps * caps, GstCaps * othercaps);
 static gboolean gst_tensor_filter_set_caps (GstBaseTransform * trans,
     GstCaps * incaps, GstCaps * outcaps);
+static gboolean gst_tensor_filter_query (GstBaseTransform * trans,
+    GstPadDirection direction, GstQuery * query);
 static gboolean gst_tensor_filter_transform_size (GstBaseTransform * trans,
     GstPadDirection direction, GstCaps * caps, gsize size,
     GstCaps * othercaps, gsize * othersize);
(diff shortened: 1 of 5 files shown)


jvuillaumier:
Change updated to fix CI errors:
- gst-indent fixups
- corrected arch-dependant xprintf() format strings

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1976,6 +1982,12 @@ gst_tensor_filter_common_set_property (GstTensorFilterPrivate * priv,
     case PROP_SHARED_TENSOR_FILTER_KEY:
       status = _gtfc_setprop_SHARED_TENSOR_FILTER_KEY (prop, value);
       break;
+    case PROP_LATENCY_REPORT:
+      priv->latency_reporting = g_value_get_boolean (value);
+      /* latency reporting requires latency_mode=1 to compute latency estimates */
```
myungjoo, line 1987: 1. If it is required, the property description should let users know.

2. I'd suggest to mention that latency_report will be only effective if latency_mode is 1 and ignore latency_report if it is kept 0 in run-time. You may emit a warning message for such inconsistency.

If latency_mode=1 is set forcibly when latency_reporting=1, some frameworks/applications on the of gstreamer that do set-property repeatedly for a given element (e.g., pipeviz) may suffer from undesired behaviors.

Or you may update "need_profiling" accordingly and rely on it around LINE 788

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -721,6 +785,7 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
   GST_TF_FW_INVOKE_COMPAT (priv, ret, invoke_tensors, out_tensors);
   if (need_profiling)
     record_statistics (priv);
+  track_latency (self);
```
myungjoo, line 788: Do we need to track this while we are not profiling anything?
This appears to be an unnecessary workload for most deployed systems.

Isn't it ok to have this inside ```if (need_profiling)```?

gst/nnstreamer/tensor_filter/tensor_filter_common.c
jvuillaumier: Agreed, it is clearer to keep properties and associated variables independent from each others.

Change updated to no longer couple latency_mode variable to the latency_report setting.
Logic has been moved to  need_profiling definition and record_statistics(), solely based on latency_reporting variable that reflects latency_report setting.

gst/nnstreamer/tensor_filter/tensor_filter.c
jvuillaumier: Update done in last changeset.

myungjoo (approved):
LGTM.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -429,6 +454,45 @@ record_statistics (GstTensorFilterPrivate * priv)
...
+ *        (arbitrary value) to a lower latency is also reported to pipeline.
+ *        Notification is done sending LATENCY message to bus. Upon receipt,
+ *        application will initiate a pipeline latency probe via LATENCY query.
+ */
+static void
+track_latency (GstTensorFilter * self)
+{
+  GstTensorFilterPrivate *priv = &self->priv;
+  gdouble estimated, reported, deviation;
+
```
songgot, line 473: How about check 'priv->latency_reporting' in here?
if (!priv->latency_reporting)
  return;
or remove line 479 "(priv->latency_reporting) && " and add it to line 789

479: if (estimated > 0) {

789: if (priv->latency_reporting)
           track_latency (self);

jvuillaumier: Agreed, such slightly different logic could be used with similar functional behavior.
Change has been merged already so can not integrate in this PR - suggestion could make it into future change.
Thanks

jvuillaumier:
Thank you @myungjoo @anyj0527 @songgot for the reviews.

PR #3973, opened 2022-11-25
src/iio: clean up build logic

myungjoo:
related logic rewritten:
- determine if src-iio is built
- determine if src-iio is registered
- determine if src-iio is tested via gtest

all the three decisions are unified for glitches in alpine build.


jaeyun-jung (approved):

songgot (approved):
LGTM

PR #3974, opened 2022-11-25
Add SSAT to RISC-V github-action

myungjoo:
Get SSAT and build SSAT for RISC-V
Then, Run!

@TODO: SSAT TC fails in RISC-V.
   Either fix failed cases or let it ignore in RISC-V build.


songgot (approved):
LGTM

jaeyun-jung (approved):

PR #3980, opened 2022-12-01
github-action: Add libunwind-dev for jammy (22.04)

again4you:
In the case of Ubuntu 22.04, the below error occurs since unmet dependencies. This patch fixes this bug.
- libunwind-14-dev : Breaks: libunwind-dev but 1.3.2-2build2 is to be installed


jaeyun-jung (approved):

myungjoo:
This appears to be a packaging (gcc?) bug of Ubuntu 22.04, but let's add it anyway.

myungjoo (approved):

PR #3996, opened 2022-12-16
[Svace] Fix the array overflow issue

again4you:
The result of 'sizeof (dump_list_type) / sizeof (nnsconf_type_path)' is 4 but dump_list_str array has only 3 items. Because of this reason, it can occur the array overflow issue. This patch fixes this issue.


songgot (approved):
LGTM, Thank you!!

niley7464 (approved):
LGTM :)

jaeyun-jung (approved):

PR #4000, opened 2022-12-22
[tests] change hardcoded dimension compare to use NNS_TENSOR_RANK_LIMIT

niley7464:
This patch changes to use NNS_TENSOR_RANK_LIMIT instead of hardcoded dimension. 
It will be used for increasing NNS_TENSOR_RANK_LIMIT.

Signed-off-by: Yelin Jeong <yelini.jeong@samsung.com>

**Self evaluation:**

### DIFF
--- a/tests/cpp_methods/cppfilter_test.cc
+++ b/tests/cpp_methods/cppfilter_test.cc
@@ -28,7 +28,8 @@ filter_basic::getInputDim (GstTensorsInfo *info)
   info->info[0].dimension[0] = 3;
   info->info[0].dimension[1] = 4;
   info->info[0].dimension[2] = 4;
-  info->info[0].dimension[3] = 1;
+  for (int i = 3; i < NNS_TENSOR_RANK_LIMIT; i++)
+    info->info[0].dimension[i] = 1;
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


anyj0527 (approved):
LGTM✔

tests/cpp_methods/cppfilter_test.cc
```diff
@@ -176,6 +180,8 @@ filter_basic2::getOutputDim (GstTensorsInfo *info)
   info->info[0].dimension[1] = 16;
   info->info[0].dimension[2] = 16;
   info->info[0].dimension[3] = 2;
+  for (int i = 4; i < NNS_TENSOR_RANK_LIMIT; i++)
+    info->info[0].dimension[i] = 1;
   return 0;
```
songgot, line 185: LGTM !! and How about below code?
```
int value[NNS_TENSOR_RANK_LIMIT] = {3, 16, 16, 2, 1, 1, 1, 1};
  for (int i = 0; i < NNS_TENSOR_RANK_LIMIT; i++)
    info->info[0].dimension[i] = value[i];
return 0;
```

niley7464: This PR is for not to use hardcoded dimension.
So I think `{3, 16, 16, 2, 1, 1, 1, 1}` this part may cause a problem when NNS_TENSOR_RANK_LIMIT has a different value.
Currently, NNS_TENSOR_RANK_LIMIT is 4 in main branch :)

songgot: Thank you.
In that case, I think you can use value[8].

songgot (approved):
LGTM!!
I understand it as a temporary code based on the difference in NNS_TENSOR_RANK_LIMIT value.
thank you.

PR #4010, opened 2023-01-03
[fix][tensor_filter] Fix the first output with garbage value with tflite

anyj0527:
- The output buffer address may change after the first invoke.
- To handle this case, Memcpy the output from tflite interpreter at the first invoke.

**Self evaluation:**


songgot (approved):
LGTM

niley7464 (approved):
LGTM 👍

jaeyun-jung (approved):

PR #4013, opened 2023-01-13
[tensor_converter] Remove a seemingly unnecessary caps attribute 'views' to make tensor_converter compatible with vaapih264dec.

tschulz:
When using ` vaapih264dec` (gstreamer v1.20.1) to decode video, I have been unable to run a pipeline with `tensor_converter` successfully. GST_DEBUG=2 helps finding this warning hint:
```
0:00:08.260625980 48364 0x7ff988007520 WARN basetransform gstbasetransform.c:1370:gst_base_transform_setcaps:<videoconvert3> transform could not transform video/x-raw, format=(string)NV12, width=(int)1920, height=(int)1080, framerate=(fraction)30000/1001, interlace-mode=(string)progressive, pixel-aspect-ratio=(fraction)1/1, multiview-mode=(string)mono, multiview-flags=(GstVideoMultiviewFlagsSet)0:ffffffff:/right-view-first/left-flipped/left-flopped/right-flipped/right-flopped/half-aspect/mixed-mono, views=(int)2, chroma-site=(string)mpeg2, colorimetry=(string)bt709 in anything we support
```

It looks like `vaapih264dec` produces a src caps like `video/x-raw, ... multiview-mode=(string)mono, ... views=(int)2`. But the caps restriction of `views=(int)1` prohibits the successful negotiation. 

Now, I don't fully understand the details and side effects, not do I understand if the VAAPI element behavior should be considered a bug or correct. In any case other sink elements can handle this gracefully (almost any other video element I've tried does so).

### DIFF
--- a/debian/nnstreamer-single-dev.install
+++ b/debian/nnstreamer-single-dev.install
@@ -1,6 +1,7 @@
 /usr/include/nnstreamer/nnstreamer_version.h
 /usr/include/nnstreamer/tensor_typedef.h
 /usr/include/nnstreamer/nnstreamer_plugin_api_filter.h
+/usr/include/nnstreamer/nnstreamer_plugin_api_trainer.h
 /usr/include/nnstreamer/nnstreamer_plugin_api_util.h
 /usr/lib/*/pkgconfig/nnstreamer-single.pc
 /usr/lib/*/libnnstreamer-single.a
--- a/gst/nnstreamer/elements/gsttensor_converter.c
+++ b/gst/nnstreamer/elements/gsttensor_converter.c
@@ -1429,7 +1429,7 @@ gst_tensor_converter_parse_video (GstTensorConverter * self,
    */
   GstVideoInfo vinfo;
   GstVideoFormat format;
-  gint width, height;
+  gint width, height, views;
   guint i;
 
   g_return_val_if_fail (config != NULL, FALSE);
@@ -1440,8 +1440,8 @@ gst_tensor_converter_parse_video (GstTensorConverter * self,
   if (!gst_video_info_from_caps (&vinfo, caps)) {
     char *capstr = gst_caps_to_string (caps);
     GST_ERROR_OBJECT (self,
-        "Failed to get video info from caps; gst_video_info_from_caps (&info, \"%s\") has returned FALSE, which means the given caps cannot be parsed as a video.",
-        capstr);
+      "Failed to get video info from caps; gst_video_info_from_caps (&info, \"%s\") has returned FALSE, which means the given caps cannot be parsed as a video.",
+      capstr);
     g_free (capstr);
     return FALSE;
   }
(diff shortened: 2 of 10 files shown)


myungjoo:
Reference: https://gstreamer.freedesktop.org/documentation/additional/design/stereo-multiview-video.html?gi-language=c

The tensor-converter implementation won't understand multi-view video streams.

A multi-view video stream has multiple heterogeneous video streams in it; e.g., stereo vision or multiview sports channels.

To support multi-view with tensor-converter:

Approach A (without modifying tensor_converter, enabled by properly writing gstreamer pipelines)
    - demux a multi-view video stream into multiple single-view video streams. (I don't know which gst-plugin does this, but there should be one..)
    - apply tensor-converter to each single-view video stream.
    - apply tensor-mux to the resulting multiple tensor streams.


Approach B (modify tensor_converter)
    - update tensor_converter so that it converts each view of a multi-view video streams. (multi-view video stream ==> multi-tensor stream)


If you want to simply use the first view of this multi-view video, Approach A will be suffice. You will only need to find the gstreamer plugin that "select" a view with frame id, before feeding it to tensor_converter.


Although it will make the pipeline descriptions shorter and neat, Approach B requires some implementation efforts and we will need to wait for someone who's willing to do it :)

myungjoo:
However, it'd be relatively easy to modify tensor_converter to accept a multi-view stream and convert to a single tensor stream with the first video only. I'll ask for proper warning messages in the case where views>1 (so that users can see that the second stream is lost.)

tschulz:
Thanks for the response @myungjoo ! I will first try to investigate why I get `views=2, ..., multiview-mode=mono` in the first place. If this turns out to be correct and expected, I will follow your direction and draft a pull request to change tensor_converter like you described.

tschulz:
Note: The situation seems to arise because of the VAAPI decoder behavior, introduced by this commit: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/commit/9f02c3fa4ccf66d0258608ce9eda3051706a835e#3e5beac4a4e4cf0ddd1d7d67a1d8af984a0e77da_1563_1584

tschulz:
@myungjoo I have filed and proposed a fix for the root cause I stumbled upon in the VAAPI element `vaapih264dec` (https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/3724)

However, i"ve also added some code in this pull request to `tensor_converter` to warn and ignore views > 1. I haven't really tested with a truly stereoscopic video, producing legit views=2. I guess that would need some more code in the converter code to ignore the data for the other views.

myungjoo (approved):
LGTM. Thanks!

myungjoo:
Please clean up the PR.

1. Be based on ```main``` branch.
2. Have a single commit as long as the change is small and the topic is coherent.

tschulz:
> Please clean up the PR.
> 
> 1. Be based on `main` branch.
> 2. Have a single commit as long as the change is small and the topic is coherent.
> 3. You can clean up your PR by doing force-push to the pushed branch (eyepop-ai:upstream-main)

@myungjoo I've re-based and squashed the changes. Please let me know if this is OK.

tschulz:
Looks like I am still struggling with git rebase ... I will close this and open a new PR on a clean branch

PR #4021, opened 2023-01-30
[common] Add header structure for 16+th tensors and simple APIs

anyj0527:
- Add header structure for 16+th tensors
- Those tensors would be packed into 16th GstMemory in a GstBuffer.
- Define NNS_TENSOR_SIZE_EXTRA_LIMIT as 200
- Add an init API `gst_tensors_extra_init`
- Add an API `gst_tensors_get_nth_memory` which returns GstMemory from given GstBuffer with index larger than 16.

unittest:
- Add a simple TC which push and get 20 tensors.
- Note that the pipeline used in this TC does not care about the caps or negotiation.

TODO:
- Add APIs for `GstTensorInfo` of 16+th tensors. These will be used in each nnstreamer element for configuration and negotiation.
- Implement proper behavior of tensor_mux and tensor_merge.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -1428,3 +1428,131 @@ gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem)
   gst_memory_unmap (new_mem, &new_map);
   return new_mem;
 }
+
+#define NNS_TENSOR_EXTRA_MAGIC 0xf00dc0de
+
+/**
+ * @brief Initialize GstTensorExtraInfo structure with given @a memory.
+ * @param[in/out] extra GstTensorExtraInfo to be initialized.
+ * @param[in] memory The information of given memory is used to initialize @a extra.
+*/
+void
+gst_tensors_extra_init (GstTensorExtraInfo * extra, GstMemory * memory)
+{
+  guint i;
+
+  extra->magic = NNS_TENSOR_EXTRA_MAGIC;
+  extra->version = 0;
+  extra->num_extra_tensors = 0;
+
+  /* set reserved size of NNS_TENSOR_SIZE_LIMIT-th memory */
+  extra->reserved = gst_memory_get_sizes (memory, NULL, NULL);
+  for (i = 0; i < NNS_TENSOR_SIZE_EXTRA_LIMIT; ++i) {
+    gst_tensor_info_init (&extra->infos[i]);
+  }
+}
+
+/**
+ * @brief Get the nth GstMemory from given @a buffer.
+ * @param[in] buffer GstBuffer to be parsed.
+ * @param[in] info GstTensorsInfo to be used in parsing buffer.
+ * @param[in] index Index of GstMemory to be returned.
+ * @return GstMemory if found, otherwise NULL (Caller should free returned memory using gst_memory_unref()).
+*/
+GstMemory *
+gst_tensors_get_nth_memory (GstBuffer * buffer, const GstTensorsInfo * info, const guint index)
+{
+  guint i, offset = 0;
+  GstMemory *extra_tensors_memory, *res_mem;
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1428,3 +1428,91 @@ gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem)
...
+  for (i = 0; i < NNS_TENSOR_SIZE_EXTRA_LIMIT; ++i) {
+    gst_tensor_info_init (&extra->infos[i]);
+  }
+}
+
+/**
+ * @brief TBU
+*/
+GstMemory *
+gst_tensors_get_nth_memory (GstBuffer * buffer, guint index)
```
jaeyun-jung, line 1454: Plz consider the param `GstTensorsInfo * info` for the case if num-tensors is 16,
```
gst_tensor_get_nth_memory (GstBuffer * buffer, const GstTensorsInfo * info, const guint index)
{
...
  if (info->num_tensors <= NNS_TENSOR_SIZE_LIMIT)
    return gst_buffer_get_memory (buffer, index);
...
}
```

gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1428,3 +1428,91 @@ gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem)
...
+  GstMapInfo extra_tensors_map;
+  GstTensorExtraInfo *extra_info;
+
+  buffer_n_memory = gst_buffer_n_memory (buffer);
+
+  /* If buffer contains less than NNS_TENSOR_SIZE_LIMIT, it's trivial. */
+  if (buffer_n_memory < NNS_TENSOR_SIZE_LIMIT
+      || index < NNS_TENSOR_SIZE_LIMIT - 1) {
+    return gst_buffer_get_memory (buffer, index);
+  }
```
myungjoo, line 1467: As @jaeyun-jung said, if ```buffer_n_memory == 16``` and ```index == 15```, without the num_tensors, you don't know whether you may do the "quick access (traditional)" or "extended access".

gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1428,3 +1428,91 @@ gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem)
...
+*/
+GstMemory *
+gst_tensors_get_nth_memory (GstBuffer * buffer, guint index)
+{
+  guint buffer_n_memory, i, offset = 0;
+  GstMemory *extra_tensors_memory, *res_mem;
+  GstMapInfo extra_tensors_map;
+  GstTensorExtraInfo *extra_info;
+
+  buffer_n_memory = gst_buffer_n_memory (buffer);
```
songgot, line 1461: If the input buffer is not GstBuffer, the buffer_n_memory is 0.
and gst_buffer_get_memory return 'null' if buffer is not GstBuffer;

How about add below to line 1461 ?
g_return_val_if_fail (GST_IS_BUFFER (buffer), NULL);

songgot (approved):
thank you for checking GST_IS_BUFFER (buffer).
LGTM

myungjoo (approved):
LGTM. PTAL: @jaeyun-jung

gst/nnstreamer/nnstreamer_plugin_api_impl.c
anyj0527: Thanks for the suggestion, I updated the code.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
anyj0527: I updated the code, PTAL

jaeyun-jung (approved):

PR #4029, opened 2023-02-07
[common] Add extra info for 16+th tensors 

anyj0527:
- Add a pointer of GstTensorInfo in GstTensorsInfo struct. And name it  as "extra"
- Add create and free APIs for extra info
- Fix get/parse utils for type, dim, name of tensor_info
- Fix tests that check NNS_TENSOR_SIZE_LIMIT in get/parse type/dim/name of tensor_info

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/include/tensor_typedef.h
+++ b/gst/nnstreamer/include/tensor_typedef.h
@@ -246,6 +246,7 @@ typedef struct
 {
   unsigned int num_tensors; /**< The number of tensors */
   GstTensorInfo info[NNS_TENSOR_SIZE_LIMIT]; /**< The list of tensor info */
+  GstTensorInfo *extra; /**< The list of tensor info for tensors whose idx is larger than NNS_TENSOR_SIZE_LIMIT */
   tensor_format format; /**< tensor stream type */
 } GstTensorsInfo;
 
--- a/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
@@ -307,6 +307,53 @@ gst_tensor_info_get_rank (const GstTensorInfo * info)
 }
 
 /**
+ * @brief Allocate and initialize the extra info in given tensors info.
+ * @param[in,out] info tensors info to be updated.
+*/
+gboolean
+gst_tensors_info_extra_create (GstTensorsInfo * info)
+{
+  GstTensorInfo *new;
+
+  g_return_val_if_fail (info != NULL, FALSE);
+
+  if (info->extra) {
+    nns_logd ("Extra tensors info is allocated already");
+    /* clear existing extra info */
+    gst_tensors_info_extra_free (info);
+  }
+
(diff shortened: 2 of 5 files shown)


gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
```diff
@@ -578,16 +660,21 @@ gst_tensors_info_parse_names_string (GstTensorsInfo * info,
     str_names = g_strsplit (name_string, ",", -1);
     num_names = g_strv_length (str_names);
```
songgot, line 662: How about using a variable to sum of NNS_TENSOR_SIZE_LIMIT and NNS_TENSOR_SIZE_EXTRA_LIMIT ?

gst/nnstreamer/include/tensor_typedef.h
```diff
@@ -245,6 +245,7 @@ typedef struct
 {
   unsigned int num_tensors; /**< The number of tensors */
   GstTensorInfo info[NNS_TENSOR_SIZE_LIMIT]; /**< The list of tensor info */
+  GstTensorInfo *extra; /**< Info for tensors whose idx is larger than NNS_TENSOR_SIZE_LIMIT */
```
songgot, line 248: The description of info is as follows.
GstTensorInfo info[NNS_TENSOR_SIZE_LIMIT]; /**< The list of tensor info */

How about using 'list' instead of 'idx'?

jaeyun-jung (approved):

gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
```diff
@@ -306,6 +306,53 @@ gst_tensor_info_get_rank (const GstTensorInfo * info)
...
+gst_tensors_info_extra_create (GstTensorsInfo * info)
+{
+  GstTensorInfo *new;
+
+  g_return_val_if_fail (info != NULL, FALSE);
+
+  if (info->extra) {
+    nns_logd ("Extra tensors info is allocated already");
+    /* clear existing extra info */
+    gst_tensors_info_extra_free (info);
```
niley7464, line 323: Could I know why this function clear existing extra info instead of return FALSE?
It's not suggestion, just a question :)

anyj0527: This function does initializing as well as creating. Add new functions for just creating or initializing could be more proper. I'll consider this later, thanks!

gst/nnstreamer/include/tensor_typedef.h
anyj0527: Thanks for this suggestion, I used the word 'list'

gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
niley7464: Thank you for the explanation

niley7464 (approved):
LGTM :)

PR #4033, opened 2023-02-10
[datareposrc] Add epochs and start/stop-sample-index property @open sesame 04/20 09:25

songgot:
Add start/stop-sample-index and epochs property
- Add start-sample-index and stop-sample-index for setting range of files or samples to read (set none, set both, or set one)
- Add epochs property to set number of repetitions, datareposrc read range of files or samples for epochs
- Add gst_base_src_set_live and move start function to create function.
  a. datareposrc can get media type and size after set_caps(), but set_caps is completed after PAUSED_TO_PLAYING.
  b. filling the buffer should be pending until set_caps().
  c. datareposrc is a live source.

**Self evaluation:**

### DIFF
--- a/gst/datarepo/gstdatareposrc.c
+++ b/gst/datarepo/gstdatareposrc.c
@@ -12,12 +12,12 @@
  *
  * ## Example launch line
  * |[
- * gst-launch-1.0 datareposrc location=mnist_trainingSet.dat ! \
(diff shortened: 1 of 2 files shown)


jaeyun-jung (approved):

gichan-jang (approved):
LGTM

PR #4040, opened 2023-03-08
[fix][API] Fix a bug that free wrong memory area

songgot:
Fix a bug that free wrong memory area
- Free last_memory instead of new_memory

**Self evaluation:**


songgot:
It's my mistake.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1661,7 +1661,7 @@ gst_tensors_extra_append_memory_to_buffer (GstBuffer * buffer,
   /* copy last_memory into new_memory */
   if (!gst_memory_map (last_memory, &last_memory_map, GST_MAP_READ)) {
     nns_loge ("Failed to map last memory");
-    gst_memory_unref (new_memory);
+    gst_memory_unref (last_memory);
```
jaeyun-jung, line 1664: no, last_mem from gst_buffer_peek_memory() is not allocated memory. we should release new_mem here (allocated at line 1649)

songgot: thank you. It's my mistake.

PR #4041, opened 2023-03-10
[datareposrc] Add is-shuffle property

songgot:
If the value is true, samples index are shuffled, default value is true.
The sample index is the order in which samples are stored in the file, starting from 0.

**Self evaluation:**

### DIFF
--- a/gst/datarepo/gstdatareposrc.c
+++ b/gst/datarepo/gstdatareposrc.c
@@ -115,11 +115,13 @@ enum
   PROP_LOCATION,
   PROP_START_SAMPLE_INDEX,
   PROP_STOP_SAMPLE_INDEX,
-  PROP_EPOCHS
+  PROP_EPOCHS,
+  PROP_IS_SHUFFLE
 };
 
 #define DEFAULT_INDEX 0
 #define DEFAULT_EPOCHS 1
+#define DEFAULT_IS_SHUFFLE TRUE
 
 static void gst_data_repo_src_finalize (GObject * object);
 static void gst_data_repo_src_set_property (GObject * object, guint prop_id,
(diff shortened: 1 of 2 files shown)


gst/datarepo/gstdatareposrc.c
```diff
@@ -373,6 +392,54 @@ gst_data_repo_src_get_file_offset (GstDataRepoSrc * src, guint64 sample_index)
...
+    value_j = g_array_index (src->shuffled_index_array, gint, j);
+
+    /* shuffled_index_array->data type is gchar * */
+    *(src->shuffled_index_array->data + (sizeof (gint) * i)) = value_j;
+    *(src->shuffled_index_array->data + (sizeof (gint) * j)) = value_i;
+  }
+
+  for (i = 0; i < src->shuffled_index_array->len; i++) {
+    GST_DEBUG_OBJECT (src, "%d -> %d", i,
+        g_array_index (src->shuffled_index_array, gint, i));
```
gichan-jang, line 422: If this loop for debugging, how about run the loop when `silent=FALSE` or `DEBUG=TRUE`?

gst/datarepo/gstdatareposrc.c
```diff
@@ -373,6 +392,54 @@ gst_data_repo_src_get_file_offset (GstDataRepoSrc * src, guint64 sample_index)
...
+  guint i, j;
+  gint value_i, value_j;
+  g_return_if_fail (src != NULL);
+  g_return_if_fail (src->shuffled_index_array != NULL);
+
+  GST_LOG_OBJECT (src, "samples index are shuffled");
+
+  /* Fisher-Yates algorithm */
+  /* The last index is the number of samples - 1. */
+  for (i = src->num_samples - 1; i > 0; i--) {
```
gichan-jang, line 410: Don't need to shuffle i=0?

gichan-jang (approved):
LGTM with minor comments.

gst/datarepo/gstdatareposrc.c
songgot: This is output only when the GST_LOG level is 4.
The user needs to export GST_DEBUG=element_name:4 or --gst-debug=element_name:4 in the pipeline to output.

gst/datarepo/gstdatareposrc.c
songgot: in each read function, it is checked below.
    if (src->is_shuffle)
      gst_data_repo_src_shuffle_samples_index (src);

anyj0527 (approved):
LGTM

again4you (approved):
LGTM!

PR #4048, opened 2023-03-23
[filter] Fix the wrong index in tflite filter @open sesame 03/27 17:33

niley7464:
This patch fixes the wrong start index in tflite filter. 
If rank starts from 0, some dimension info can be ignored.

See also: #3814 

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -708,6 +708,7 @@ TFLiteInterpreter::setInputTensorsInfo (const GstTensorsInfo *info)
 {
   TfLiteStatus status = kTfLiteOk;
   const std::vector<int> &input_idx_list = interpreter->inputs ();
+  int input_rank;
 
   /** Cannot change the number of inputs */
   if (info->num_tensors != input_idx_list.size ())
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -729,7 +730,8 @@ TFLiteInterpreter::setInputTensorsInfo (const GstTensorsInfo *info)
      * iterate over all possible ranks starting from MIN rank to the actual rank
      * of the dimension array. In case of none of these ranks work, return error
      */
-    for (int rank = 0 ; rank < NNS_TENSOR_RANK_LIMIT ; rank++) {
+    input_rank = gst_tensor_info_get_rank (tensor_info);
+    for (int rank = input_rank; rank < NNS_TENSOR_RANK_LIMIT; rank++) {
```
songgot, line 734: I want to know if it is "input_rank -1" or not.

niley7464: Updated to use just `rank` instead of `rank+1`
So, It will be okay to use `input_rank`.
Thank you :)

songgot: Thank you.

songgot (approved):
Looks good to me.

anyj0527 (approved):
LGTM👍 Thanks!

jaeyun-jung (approved):

PR #4049, opened 2023-03-24
[datareposrc] Add tensor-sequence property @open sesame 4/21 16:51

songgot:
- Tensors in a sample are read into gstBuffer according to input sequence.
    - Only read the set tensors among all tensors in a sample.
    - It is for other/tensors and default value is NULL.
```
e.g, if a sample has '1:1:1:1', '1:1:10:1', '1:1:784:1' and each index is '0,1,2' , 
"tensors-sequence=2,1" means that only '1:1:784:1' then '1:1:10:1' are read. 
```
further explanation
```
I modify MNIST dataset that have two features and labels in a sample.
so..
     offset and sample_offset(byte size) are from 0.
      if the size of one sample is 6352 and num_tensors is 4, dimensions are  '1:1:784:1' , '1:1:10:1',
      '1:1:784:1' and '1:1:10:1' with float32.
      the offset of the second sample is as follows.
        -------------------------------------------------
          sample_offset: 6352
          tensors index: [ 0    | 1    | 2    | 3     ]
           tensors_size: [ 3136 | 40   | 3136 | 40    ]
         tensors_offset: [ 0    | 3136 | 3176 | 6312  ]
              fd offset: [ 6352 | 9488 | 9528 | 12664 ]
        -------------------------------------------------
      if user sets "tensor-sequence=2,1", datareposrc read offset 9528 then 9488.
```
**Self evaluation:**

### DIFF
--- a/gst/datarepo/gstdatareposrc.c
+++ b/gst/datarepo/gstdatareposrc.c
@@ -118,7 +118,8 @@ enum
   PROP_START_SAMPLE_INDEX,
   PROP_STOP_SAMPLE_INDEX,
   PROP_EPOCHS,
-  PROP_IS_SHUFFLE
+  PROP_IS_SHUFFLE,
+  PROP_TENSORS_SEQUENCE,
 };
 
 #define DEFAULT_INDEX 0
@@ -180,7 +181,7 @@ gst_data_repo_src_class_init (GstDataRepoSrcClass * klass)
           GST_PARAM_MUTABLE_READY));
 
   g_object_class_install_property (gobject_class, PROP_START_SAMPLE_INDEX,
-      g_param_spec_int ("start-sample-index", "Start index of samples",
+      g_param_spec_uint ("start-sample-index", "Start index of samples",
           "Start index of sample to read, in case of image, "
           "the starting index of the numbered files. start at 0."
           "Set start index of range of samples or files to read",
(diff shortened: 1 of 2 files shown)


gst/datarepo/gstdatareposrc.c
```diff
@@ -332,6 +348,60 @@ gst_data_repo_src_set_file_path (GstDataRepoSrc * src, const int prop,
...
+
+  length = g_strv_length (strv);
+  if (length > NNS_TENSOR_SIZE_LIMIT) {
+    GST_ERROR_OBJECT (src, "The total number of indices exceeded %d.",
+        NNS_TENSOR_SIZE_LIMIT);
+    goto error;
+  }
+
+  while (strv[i] != NULL && strlen (strv[i]) > 0) {
+    src->tensors_seq[i] = atoi (strv[i]);
```
anyj0527, line 375: How about use sth like `(guint) g_ascii_strtoll (strv[i], NULL, 10)` rather than vulnerable `atoi`?

gst/datarepo/gstdatareposrc.c
```diff
@@ -198,12 +199,24 @@ gst_data_repo_src_class_init (GstDataRepoSrcClass * klass)
           GST_PARAM_MUTABLE_READY));
 
   g_object_class_install_property (gobject_class, PROP_EPOCHS,
-      g_param_spec_int ("epochs", "Epochs",
+      g_param_spec_uint ("epochs", "Epochs",
           "Repetition of range of files or samples to read, set number of repetitions",
           0, G_MAXINT, DEFAULT_EPOCHS,
```
anyj0527, line 204: `G_MAXUINT` ?

songgot: Thank you.
property can get uint value but.
Limit the number of samples to a number in the range int.
However, in the case of epochs, it doesn't seem to make much sense.

gst/datarepo/gstdatareposrc.c
songgot: okay, I'll try and test

anyj0527 (approved):
LGTM ✔

niley7464 (approved):
LGTM!

PR #4059, opened 2023-04-12
[TEST] Invoke tensor_filter with flexible input and output.

gichan-jang:
Testcase for flexible tensor filter is disabled and activate it after development is done.
In order to invoke tensor_filter with flexible input/output, you need to specify custom option.
 ex)  ``` ... ! tensor_filter framework=... invoke-dynamic=TRUE model=... ! ...```
 * Enable tests after implementation of the tensor_filter dynamic invoke

Self evaluation:

Build test: [ ]Passed [ ]Failed [*]Skipped
Run test: [ ]Passed [ ]Failed [*]Skipped

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_trainer.c
+++ b/gst/nnstreamer/elements/gsttensor_trainer.c
@@ -18,6 +18,10 @@
  * ]|
  *
  * Total number of data to be received is 1000((num-training-samples + num-validation-samples) * epochs)
+ * 
+ * output tensors : dimensions=1:1:4, types=float64.
+ * values are training loss, training accuracy, validation loss and validation accuracy.
+ * -INFINITY value is stored if the value fetched from the sub-plugin is not greater than 0.
  */
 
 #ifdef HAVE_CONFIG_H
@@ -28,6 +32,7 @@
 #include <nnstreamer_util.h>
 #include "gsttensor_trainer.h"
 #include <unistd.h>
+#include <math.h>
 
 /**
  * @brief Default caps string for both sink and source pad.
@@ -56,6 +61,19 @@ GST_DEBUG_CATEGORY_STATIC (gst_tensor_trainer_debug);
 G_DEFINE_TYPE (GstTensorTrainer, gst_tensor_trainer, GST_TYPE_ELEMENT);
 
 /**
+ * @brief Statistical from the model being trained
+ * An enum value indicates the value stored at the index of the output tensor.
+ */
+enum
+{
+  TRAINING_LOSS,
+  TRAINING_ACCURACY,
+  VALIDATION_LOSS,
+  VALIDATION_ACCURACY
+};
+#define MODEL_STATS_SIZE 4
+
+/**
  * @brief Default framework property value
  */
 #define DEFAULT_PROP_INPUT_LIST 1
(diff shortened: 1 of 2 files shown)


tests/nnstreamer_filter_custom/unittest_filter_custom.cc
```diff
@@ -0,0 +1,141 @@
...
+  g_signal_connect (sink_handle, "new-data", (GCallback)new_data_cb, NULL);
+
+  filter_received = 0;
+  sink_received = 0;
+  EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_TRUE (wait_pipeline_process_buffers (&sink_received, 9, TEST_TIMEOUT_LIMIT));
+
+  /** cleanup registered custom_easy filter */
+  ret = NNS_custom_easy_unregister ("safe_memcpy_10x10");
```
songgot, line 110: "flexbible_filter" is correct, right?

gichan-jang: Updated. Thanks.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -161,6 +161,8 @@ typedef struct _GstTensorFilterProperties
 
   int latency; /**< The average latency over the recent 10 inferences in microseconds */
   int throughput; /**< The average throughput in the number of outputs per second */
+
+  int invoke_dynamic; /**< True for supporting flexible invoke. NNFW plugin have to allocate output memory. */
```
songgot, line 165: 'plugins have' or 'plugin has' ?

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -666,6 +666,12 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
   out_flexible =
       gst_tensor_pad_caps_is_flexible (GST_BASE_TRANSFORM_SRC_PAD (trans));
 
+  if (priv->prop.invoke_dynamic && !out_flexible) {
```
songgot, line 669: In the case of (priv->prop.invoke_dynamic && !in_flexible), is confirmation required?

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -295,6 +297,18 @@ struct _GstTensorFilterFramework
...
+  /**< Mandatory callback. Invoke the given network model.
+   *
+   * @param[in/out] private_data A subplugin may save its internal private data here. The subplugin is responsible for alloc/free of this pointer.
+   * @param[in] info structure of input tensors info
+   * @param[out] info structure of output tensors info. The subplugin should fill this info.
+   * @param[in] input The array of input tensors. Allocated and filled by tensor_filter/main
+   * @param[out] output The array of output tensors. The subplugin should allocate the memory block for output tensor. (data in GstTensorMemory)
+   * @return 0 if OK. non-zero if error.
+   */
+
```
myungjoo, line 311: The API format looks like v0 format (**private_data). Do not touch v0 format. Update v1 format.
And do not add in the middle of the struct. It breaks ABI of old binaries.

myungjoo (changes requested):
Need to reconsider the interface change.

myungjoo (approved):
LGTM

myungjoo (commented):
And make sure that the gstcap becomes ```format=flexible``` if it should be flexible ( framework supports flexible && the model wants flexible)

Even if framework supports flexible, if the models doesn't need flexible, try to keep it "static" for higher efficiency.,

gst/nnstreamer/tensor_filter/tensor_filter.c
gichan-jang: I will check your comment in next PR. Thanks.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
gichan-jang: I removed `const` from `const prop` in invoke of filter v1 to update in/out tensors info.
If the sub-plugin using filter v0 wants to use a dynamic filter later, it needs to be changed to filter v1.
Thanks.

songgot (approved):
LGTM

PR #4060, opened 2023-04-12
[Tests] Move generateTest script to non-optional test directory

abirlinger:
Hello all,

This commit aims to get rid of SSAT test failures for platform not supporting Pytorch.  As nnstreamer_filter_pytorch is an optional test introduced by 'pytorch_support_is_available' flag when building it could introduce SSAT failures for test suites requiring the generateTest script located into nnstreamer_filter_pytorch directory. This script added to nnstreamer_filter_pytorch is also used by nnstreamer_converter, nnstreamer_filter_tensorflow2_lite, and transform_dimchg. When Pytorch is not supported, the nnstreamer_filter_pytorch directory is not available and other tests above could not access to the script.

The proposal is to move 'generateTest.py' script to nnstreamer_converter directory as the script is not only used by nnstreamer_filter_pytorch, but also by nnstreamer_converter which is a unittest installed by default. Another script used in the same way as 'generateTest.py' (i.e. used by the unittest declared within the same directory and some others) was already located in the nnstreamer_converter directory which was making this unittest a good candidate to hold this script.

- 'generateTest.py' script location used for higher rank unittests

**Self evaluation:**
SSAT test suites passed (nnstreamer_converter, nnstreamer_filter_tensorflow2_lite, transform_dimchg) when pytorch not available

**How to evaluate:**
1. Disable Pytorch support
2. Build NNStreamer
3. Run mentioned SSAT


wooksong (approved):
LGTM

jaeyun-jung (approved):

PR #4072, opened 2023-04-27
[MQTT] Change string copy @open sesame 04/28 16:12

gichan-jang:
Change string copy to avoid gcc-12 build error.

**Self evaluation:**

### DIFF
--- a/gst/mqtt/mqttsink.c
+++ b/gst/mqtt/mqttsink.c
@@ -929,10 +929,10 @@ gst_mqtt_sink_set_caps (GstBaseSink * basesink, GstCaps * caps)
   ret = gst_caps_replace (&self->in_caps, caps);
 
   if (ret && gst_caps_is_fixed (self->in_caps)) {
-    char *caps_str = gst_caps_to_string (caps);
+    gchar *caps_str = gst_caps_to_string (caps);
 
-    strncpy (self->mqtt_msg_hdr.gst_caps_str, caps_str,
-        MIN (strlen (caps_str), GST_MQTT_MAX_LEN_GST_CAPS_STR - 1));
+    g_strlcpy (self->mqtt_msg_hdr.gst_caps_str, caps_str,
+        GST_MQTT_MAX_LEN_GST_CAPS_STR);
     g_free (caps_str);
   }
 


jaeyun-jung (approved):

gst/mqtt/mqttsink.c
```diff
@@ -932,7 +932,7 @@ gst_mqtt_sink_set_caps (GstBaseSink * basesink, GstCaps * caps)
     char *caps_str = gst_caps_to_string (caps);
```
wooksong, line 932: Q: What happens if using the returned ```gchar *``` for ```gst_caps_str ```
```suggestion
    self->mqtt_msg_hdr.gst_caps_str = gst_caps_to_string (caps);
```

gichan-jang: In my opinion, it is better to limit the size for safe string copying.

gst/mqtt/mqttsink.c
```diff
@@ -929,10 +929,10 @@ gst_mqtt_sink_set_caps (GstBaseSink * basesink, GstCaps * caps)
   ret = gst_caps_replace (&self->in_caps, caps);
 
   if (ret && gst_caps_is_fixed (self->in_caps)) {
-    char *caps_str = gst_caps_to_string (caps);
+    gchar *caps_str = gst_caps_to_string (caps);
```
wooksong, line 932:
```suggestion
    g_autofree gchar *caps_str = gst_caps_to_string (caps);
    
    self->self->mqtt_msg_hdr.gst_caps_str = g_strdup (caps);
```

https://docs.gtk.org/glib/func.strlcpy.html

jaeyun-jung: self->mqtt_msg_hdr.gst_caps_str is fixed size string array, cannot set allocated memory.

https://github.com/nnstreamer/nnstreamer/blob/eb54486d4d17f1146fea7cfb6d304d65ebd8dbd8/gst/mqtt/mqttcommon.h#L59

wooksong: @gichan-jang @jaeyun-jung Got it!

wooksong (approved):

PR #4077, opened 2023-05-10
[Filter] PyTorch, ignore extra output tensors @open sesame 05/11 10:51

tschulz:
Allow tensor_filter to be configured with smaller number of output tensors than PyTorch returns native. This is useful e.g. when my Yolov7 converted model (TorchScript) has 4 output tensors, but I am really only interested in in the 1st. Without this fix, the code would crash with a memory overwrite in `TorchCore::serializeOutput` if the number of configured output tensors is smaller than the actual TorchScript model provides..

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
@@ -93,7 +93,7 @@ class TorchCore
   int processIValue (const torch::jit::IValue &value, GstTensorMemory *output,
       unsigned int idx);
   int serializeOutput (const torch::jit::IValue &value, GstTensorMemory *output,
-      unsigned int *idx);
+      unsigned int *idxm, unsigned int limit_idx);
 };
 
 extern "C" { /* accessed by android api */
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc
```diff
@@ -447,8 +447,12 @@ TorchCore::processIValue (
  */
 int
 TorchCore::serializeOutput (
-    const torch::jit::IValue &value, GstTensorMemory *output, unsigned int *idx)
+    const torch::jit::IValue &value, GstTensorMemory *output, unsigned int *idx, unsigned int limit_idx)
```
gichan-jang, line 450: To fix clang format error, please update as below:
```suggestion
TorchCore::serializeOutput (const torch::jit::IValue &value,
    GstTensorMemory *output, unsigned int *idx, unsigned int limit_idx)
```
and you can check error [here](http://ci.nnstreamer.ai/nnstreamer/ci/repo-workers/pr-checker/4077-202305101037130.039629936218262-35bc7428d504c033ed0364f27a86c3e824837bd5/report/clang-format.patch)

And please check DCO and add the sign-off: [DCO](https://github.com/apps/dco)

tschulz: Thanks @gichan-jang! I took care of those two issues. There are still two checks failing and I would need some advise how to address those.

gichan-jang: The test failure is irrelevant to your change. 
I started the TAOS-CI again and it passed now.

anyj0527 (approved):
This change is only valid when "used" output tensors are located from the first order.
A good way to solve this issue is modify the model file to output only needed tensors. But I think it's still valid.
LGTM Thanks!

anyj0527 (approved):
This change is only valid when "used" output tensors are located from the first order.
A good way to solve this issue is modify the model file to output only needed tensors. But I think it's still valid.
LGTM Thanks!

gichan-jang (approved):
LGTM

jaeyun-jung (approved):

PR #4090, opened 2023-05-16
[Trainer] parse config in caps event

jaeyun-jung:
Fix invalid tensor-config,
1. parse config when receiving caps event.
2. init/free config in start/finalize callback.


songgot (approved):
LGTM

niley7464 (approved):
LGTM!

PR #4092, opened 2023-05-17
datarepo: dependency management.

myungjoo:
Datarepo has its own unique dependency.
Thus, make datarepo optional at build so that
we can cut the dependency.

It's set "enabled by default" because most
releases will be using it anyway.


songgot (approved):
LGTM

meson_options.txt
```diff
@@ -50,6 +50,7 @@ option('framework-priority-bin', type: 'string', value: '', description: 'A comm
 option('skip-tflite-flatbuf-check', type: 'boolean', value: false, description: 'Do not check the availability of flatbuf for tensorflow-lite build. In some systems, flatbuffers\' dependency cannot be found with meson.')
 option('trix-engine-alias', type: 'string', value: 'srnpu', description: 'The alias name list of trix-engine sub-plugin. This option provides backward compatibility of the previous framework name.')
 option('enable-float16', type: 'boolean', value: false, description: 'Support float16 streams with GCC extensions')
+option('enable-datarepo', type: 'boolean', value: true, description: 'Data repository sink/src for in-pipeline training')
```
wooksong, line 53:
```suggestion
option('enable-datarepo', type: 'boolean', value: true, description: 'Data repository sink/src for intra/inter-pipeline training')
```

How about it?

anyj0527 (approved):
LGTM✔

meson_options.txt
myungjoo: Nah.. in-pipeline would be more accurate. Inter-pipeline training requires more components; we do not have the methods to transmit models or model-diffs.

PR #4097, opened 2023-05-19
[Common] prepare 0-init dimension

jaeyun-jung:
Prepare increasing max rank and 0-init dimension case.


gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
```diff
@@ -1121,7 +1121,7 @@ gst_tensor_parse_dimension (const gchar * dimstr, tensor_dim dim)
     dim[i] = (uint32_t) val;
     rank = i + 1;
   }
-
+  /** @todo remove below line (init at line 1062, fill remained dimension with 0) */
```
myungjoo, line 1124: Line 1062? Please do not use code lines. It's not static. If  you want to leave a note for next task, please write an issue.

myungjoo (commented):
Is it safe to have that many ```todo``` items?

songgot (approved):
I only saw the code part. 
LGTM.

myungjoo (approved):

PR #4109, opened 2023-06-08
[CodeClean] util function for caps

jaeyun-jung:
Code clean, fix indent and use util function for gst-caps.

### DIFF
--- a/gst/datarepo/gstdatareposrc.c
+++ b/gst/datarepo/gstdatareposrc.c
@@ -268,11 +268,8 @@ gst_data_repo_src_finalize (GObject * object)
   if (src->shuffled_index_array)
     g_array_free (src->shuffled_index_array, TRUE);
 
-  /* Check for gst-inspect log */
-  if (src->caps) {
-    gst_caps_unref (src->caps);
-    src->caps = NULL;
-  }
+  if (src->caps)
+    gst_caps_replace (&src->caps, NULL);
 
   G_OBJECT_CLASS (parent_class)->finalize (object);
 }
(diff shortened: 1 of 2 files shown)


gichan-jang (approved):
LGTM

gst/datarepo/gstdatareposrc.c
```diff
@@ -1012,9 +1009,8 @@ gst_data_repo_get_caps_by_tensors_sequence (GstDataRepoSrc * src)
   GST_DEBUG_OBJECT (src,
       "datareposrc caps by tensors_sequence %" GST_PTR_FORMAT, new_caps);
 
-  gst_caps_replace (&src->caps, new_caps);
+  gst_caps_take (&src->caps, new_caps);
```
songgot, line 1012: Is it okay to put values in src->caps multiple times?
If we use gst_caps_take(), shouldn't ```src->caps``` unref before taking ownership of new_caps?

In case of  'gst_caps_replace', the old_caps is unreffed, the new_caps is referred and change each other.
```
gboolean
gst_caps_replace (GstCaps ** old_caps,
                  GstCaps * new_caps)
```
so src->caps don't need unref().
if use gst_caps_take(), src->caps take ownership of new_caps, so new_caps don't need unref(). but shouldn't src->caps unref() before taking ownership of new_caps in multiple settings?

jaeyun-jung: It is same with gst_caps_replace() - decreases ref count of old value.

myungjoo: http://www.manpagez.com/html/gstreamer-1.0/gstreamer-1.0-1.14.3/GstCaps.php#gst-caps-take

```gst_caps_take()``` = ```atomic { unref(old); ref(new); own(new); }```

myungjoo (approved):

gst/datarepo/gstdatareposrc.c
songgot: > It is same with gst_caps_replace() - decreases ref count of old value.
Thanks for your explanation. :) :+1: :+1:

songgot: > http://www.manpagez.com/html/gstreamer-1.0/gstreamer-1.0-1.14.3/GstCaps.php#gst-caps-take
> 
> `gst_caps_take()` = `atomic { unref(old); ref(new); own(new); }`

Thanks for your explanation. :) :+1: :+1:

songgot:
LGTM

PR #4126, opened 2023-06-26
[CodeClean] use util func for extra tensors

jaeyun-jung:
Code clean, remove unnecessary local val and use util function to handle extra tensors.


anyj0527 (approved):
LGTM✔

niley7464 (approved):
LGTM :)

PR #4131, opened 2023-06-27
[trivial] Fix log level warn -> info

anyj0527:
- Decrease log level since the line is in proper scenario when the num_tensors is exactly 16.


jaeyun-jung (approved):

songgot (approved):
LGTM

Issue #3904, opened 2022-09-13
Config file option to provide settings for element properties


shortcipher3:
**Is your feature request related to a problem? Please describe.**

Some of the filters such as `tensor_decoder` have a lot of properties.  It would be nice to have a config file to specify these, this way when specifying lots of config settings it doesn't clutter the code or command line, which would make the pipeline easier to read for new users.

For example, [this single element from the object detection example](https://github.com/nnstreamer/nnstreamer-example/blob/4efa166d2062d7f620202bb2c3ed663e11fa2683/bash_script/example_object_detection_tensorflow_lite/gst-launch-object-detection-tflite.sh#L7) is 166 columns:

```bash
tensor_decoder mode=bounding_boxes option1=mobilenet-ssd option2=tflite_model/coco_labels_list.txt option3=tflite_model/box_priors.txt option4=640:480 option5=300:300
```

Where best practice for bash scripts is to keep the line under 80 columns.

**Describe the solution you'd like**

These options could easily be specified in a config file, so the element reads:

```bash
tensor_decoder config-file=object_detection.config
```

And the contents of the config file are as follows:

```yaml
mode: bounding_boxes
option1: mobilenet-ssd
option2: tflite_model/coco_labels_list.txt
option3: tflite_model/box_priors.txt
option4: 640:480
option5: 300:300
```

This would make the configuration more readable and less error prone as well.

myungjoo:
Yes, this looks great.

@niley7464 : Could you please look at this one after the NPU driver issue?
 When you start this, please first design and describe its usage and interfaces before writing actual codes.

niley7464:
> Yes, this looks great.
> 
> @niley7464 : Could you please look at this one after the NPU driver issue? When you start this, please first design and describe its usage and interfaces before writing actual codes.
> 
> When you design this feature, please consider writing a common module reusable by different plugins. Possibly, we may use this feature with tensor_if, tensor_filter, and more.

I will look at this after the NPU issue :)

gichan-jang:
@Harshj20 Harsh, Please check this issue.

Harshj20:
> @Harshj20 Harsh, Please check this issue.

Yes, I will.

myungjoo:
#4135 has implemented this.

PR #4135, opened 2023-06-28
Config file for properties

Harshj20:
This includes an API named gst_tensor_parse_config_file defined in nnstreamer_plugin_api_impl.c.

For any plugin, install a new property for defining the config file path and use the API as follows:

```
case PROP_CONFIG:
    {
      const gchar *config_path = g_value_get_string (value);
      gst_tensor_parse_config_file (config_path, object);
      break;
    }
```
The config file is defined as:
```
mode=bounding_boxes
option1=mobilenet-ssd
option2=coco_labels_list.txt
option3=box_priors.txt
option4=640:480
option5=300:300
```

And the property can be used as:
`.. ! tensor_decoder config-file=configFile ! ..`

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -1784,3 +1784,98 @@ gst_buffer_n_tensor (GstBuffer * buffer)
 
   return num_mems;
 }
+
+/**
+ * @brief Sets the value of a property based on the specified property value and GParamSpec.
+ *
+ * @param prop_value A pointer to the GValue where the property value will be set.
+ * @param param_spec A pointer to the GParamSpec that describes the property.
+ * @param property_value A string representing the value to be set for the property.
+ *
+ * @note This API is intended to be used by gst_tensor_parse_config_file ()
+ */
+
+static void
+set_property_value (GValue * prop_value, const GParamSpec * param_spec,
+    const gchar * property_value)
+{
+  GType value_type = G_PARAM_SPEC_VALUE_TYPE (param_spec);
+  g_value_init (prop_value, value_type);
+
+  if (value_type == G_TYPE_BOOLEAN) {
+    gboolean value = g_ascii_strcasecmp (property_value, "true") == 0;
+    g_value_set_boolean (prop_value, value);
+  } else if (value_type == G_TYPE_INT) {
+    gint value = atoi (property_value);
+    g_value_set_int (prop_value, value);
+  } else if (value_type == G_TYPE_UINT) {
+    guint value = atoi (property_value);
+    g_value_set_uint (prop_value, value);
+  } else if (value_type == G_TYPE_FLOAT) {
+    gfloat value = atof (property_value);
+    g_value_set_float (prop_value, value);
+  } else if (value_type == G_TYPE_DOUBLE) {
+    gdouble value = atof (property_value);
+    g_value_set_double (prop_value, value);
+  } else {
+    g_value_set_string (prop_value, property_value); /** default is string */
+  }
+}
+
+/**
+ * @brief Parses a configuration file and sets the corresponding properties on a GObject.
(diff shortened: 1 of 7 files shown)


gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1784,3 +1784,63 @@ gst_buffer_n_tensor (GstBuffer * buffer)
...
+      gchar *property_name = g_strstrip (g_strdup (parts[0]));
+      gchar *property_value = g_strstrip (g_strdup (parts[1]));
+
+      GParamSpec *pdata =
+          g_object_class_find_property (g_object_class, property_name);
+      if (pdata != NULL) {
+        GValue prop_value = G_VALUE_INIT;
+        g_value_init (&prop_value, G_TYPE_STRING);
+        g_value_set_string (&prop_value, property_value);
+        g_object_set_property (G_OBJECT (object), pdata->name, &prop_value);
```
gichan-jang, line 1832: I like this changes. Good!
We will extend this method to various elements such as tensor_filter, tensor_if, and tensor_transform.
The property type may vary not only string but also boolean or int and etc.
So, let's set the property according to type of GParamSpec.

Harshj20: I shall include some common types: int, unit, float, double, boolean, string

gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1784,3 +1784,98 @@ gst_buffer_n_tensor (GstBuffer * buffer)
...
+
+      GParamSpec *pdata =
+          g_object_class_find_property (g_object_class, property_name);
+      if (pdata != NULL) {
+        GValue prop_value = G_VALUE_INIT;
+        set_property_value (&prop_value, pdata, property_value);
+        g_object_set_property (G_OBJECT (object), pdata->name, &prop_value);
+        g_value_unset (&prop_value);
+
+        g_param_spec_unref (pdata); /** Unreference GParamSpec */
```
gichan-jang, line 1870: `g_object_class_find_property ()` does not take ownership of the GparamSpec.
I think we don't need to unref the GparamSpec.

gst/nnstreamer/tensor_common.h
```diff
@@ -108,6 +108,33 @@ typedef struct
...
+ * @brief Sets the value of a property based on the specified property value and GParamSpec.
+ *
+ * @param prop_value A pointer to the GValue where the property value will be set.
+ * @param param_spec A pointer to the GParamSpec that describes the property.
+ * @param property_value A string representing the value to be set for the property.
+ *
+ * @note This API is intended to be used by gst_tensor_parse_config_file ()
+ */
+void
+set_property_value (GValue * prop_value, const GParamSpec * param_spec, const gchar * property_value);
```
gichan-jang, line 121: If this function is used only in gst_tensor_parse_config_file (). Let's move this to nnstreamer_plugin_api_impl.c and makes as a static function.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
Harshj20: Yes there is no need to unref GParamSpec. I shall update it.

gst/nnstreamer/tensor_common.h
Harshj20: I will update this. Thank you

gichan-jang:
Related issue: https://github.com/nnstreamer/nnstreamer/issues/3904
Let's add the config file test by referring to this test: https://github.com/nnstreamer/nnstreamer/blob/main/tests/nnstreamer_decoder_boundingbox/runTest.sh.

gichan-jang (approved):
LGTM

jaeyun-jung (approved):

PR #4140, opened 2023-07-03
[Common] reconfigure gst-buffer with extra tensors

jaeyun-jung:
Handle extra tensors when reconfiguring the gst-buffer.


anyj0527 (approved):
LGTM 👍 This could be used in other elements handling extra tensors

songgot (approved):
LGTM

PR #4149, opened 2023-07-06
[Demux] util functions to handle memories

jaeyun-jung:
Update util functions to handle memories in gst-buffer.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_demux.c
+++ b/gst/nnstreamer/elements/gsttensor_demux.c
@@ -468,24 +468,26 @@ done:
 static GstFlowReturn
 gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
 {
-  guint num_tensors, num_srcs, i;
+  guint num_tensors, num_srcs, i, idx;
   GstFlowReturn res = GST_FLOW_OK;
   GstTensorDemux *tensor_demux;
   GList *list = NULL;
+  GstTensorInfo *_info;
+
   UNUSED (pad);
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/elements/gsttensor_demux.c
```diff
@@ -511,15 +513,20 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
...
       }
       g_strfreev (strv);
       list = list->next;
     } else {
-      mem = gst_buffer_get_memory (buf, i);
-      gst_buffer_append_memory (outbuf, mem);
+      _info =
+          gst_tensors_info_get_nth_info (&tensor_demux->tensors_config.info, i);
+      mem = gst_tensor_buffer_get_nth_memory (buf, i);
+      gst_tensor_buffer_append_memory (outbuf, mem, _info);
```
songgot, line 529: The input argument of each function is checked for null inside the function. 
But what about checking the return value of "gst_tensor_buffer_append_memory" ?
There seem to be many cases where the function returns false.

songgot: If the value inside info entered in this function is always vaild, it's okay.

songgot (approved):
LGTM

gst/nnstreamer/elements/gsttensor_demux.c
jaeyun-jung: Updated, plz see the changes again.

anyj0527 (approved):
LGTM ✔

PR #4159, opened 2023-07-19
[tensor_region] fix memcpy to out buffer

Harshj20:
Initially, the increment size of the output buffer pointer while storing the cropping info was 4 times the required value.


jaeyun-jung (approved):

gichan-jang (approved):
Nice catch, LGTM.

PR #4163, opened 2023-07-27
[tensor_trainer] Add load_model_path to GstTensorTrainerProperties

songgot:
- Add model_load_path to load an existing model to use for training a new model
- Add model_load_path property to tensor_trainer

Related PR : (No dependency) https://github.com/nnstreamer/nntrainer/pull/2258 (review please)

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_trainer.c
+++ b/gst/nnstreamer/elements/gsttensor_trainer.c
@@ -95,6 +95,7 @@ enum
   PROP_FRAMEWORK,
   PROP_MODEL_CONFIG,
   PROP_MODEL_SAVE_PATH,
+  PROP_MODEL_LOAD_PATH,
   PROP_NUM_INPUTS,              /* number of input list */
   PROP_NUM_LABELS,              /* number of label list */
   PROP_NUM_TRAINING_SAMPLES,    /* number of training data */
@@ -127,6 +128,8 @@ static void gst_tensor_trainer_set_prop_framework (GstTensorTrainer * trainer,
 static void gst_tensor_trainer_set_prop_model_config_file_path (GstTensorTrainer
     * trainer, const GValue * value);
 static void gst_tensor_trainer_set_model_save_path (GstTensorTrainer * trainer,
+    const GValue * value);
+static void gst_tensor_trainer_set_model_load_path (GstTensorTrainer * trainer,
     const GValue * value);
 static gboolean gst_tensor_trainer_find_framework (GstTensorTrainer * trainer,
     const char *name);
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -423,9 +441,9 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
   g_return_val_if_fail (trainer != NULL, FALSE);
 
   /* Parameters that can be retrieved from caps will be removed */
-  if (!trainer->fw_name || !trainer->model_config || !trainer->model_save_path
-      || trainer->prop.num_epochs <= 0 || trainer->prop.num_inputs <= 0
-      || trainer->prop.num_labels <= 0) {
+  if (!trainer->fw_name || !trainer->prop.model_config
+      || !trainer->prop.model_save_path || trainer->prop.num_epochs <= 0
```
gichan-jang, line 445: Is it needed to check null for trainer->prop.model_config or trainer->model_save_path?

tests/nnstreamer_trainer/unittest_trainer.cc
```diff
@@ -91,7 +91,10 @@ TEST (tensor_trainer, SetParams)
   EXPECT_STREQ (get_str, model_config_path);
 
   g_object_get (tensor_trainer, "model-save-path", &get_str, NULL);
-  EXPECT_STREQ (get_str, "model.bin");
+  EXPECT_STREQ (get_str, "new_model.bin");
+
+  g_object_get (tensor_trainer, "model-load-path", &get_str, NULL);
+  EXPECT_STREQ (get_str, "old_model.bin");
```
gichan-jang, line 97: Need to free `get_str`

again4you: @gichan-jang  Nice catch!

again4you (approved):
Except @gichan-jang 's comments, it looks good to me.

gst/nnstreamer/elements/gsttensor_trainer.c
gichan-jang: Check null ptr is meaningless, let's compare with the "" string.

songgot:
Thank you for your review, @gichan-jang , @wooksong !!

gst/nnstreamer/elements/gsttensor_trainer.c
songgot: null check is also necessary. thank you.

myungjoo:
Question: what's the behavior if the load_path or save_path is NULL? (I recommend describing it at ```g_object_class_install_property```)

Also, in the next PR or here, please check the availability of file when the property is set. And, add a test cases of that, too. (both negative and positive cases)

songgot:
> availability

songgot:
> Question: what's the behavior if the load_path or save_path is NULL? (I recommend describing it at `g_object_class_install_property`)
> 
> Also, in the next PR or here, please check the availability of file when the property is set. And, add a test cases of that, too. (both negative and positive cases)

Thank you, I will add a guide and test case.

tests/nnstreamer_trainer/unittest_trainer.cc
```diff
@@ -73,7 +73,7 @@ TEST (tensor_trainer, SetParams)
       "gst-launch-1.0 datareposrc location=%s json=%s "
       "start-sample-index=3 stop-sample-index=202 tensors-sequence=0,1 epochs=1 ! "
       "tensor_trainer name=tensor_trainer framework=nntrainer model-config=%s "
-      "model-save-path=model.bin num-inputs=1 num-labels=1 "
+      "model-save-path=new_model.bin model_load_path=old_model.bin num-inputs=1 num-labels=1 "
```
niley7464, line 76: Does `set_property` treat `model_load_path` (with _) and `model-load-path` (with -) same?
Since this test passed, it seems to be treated as same, but just for check cause I didn't know it before

niley7464: How about adding what model_load_path property means in comment like other properties?

songgot: Thank you, I'm sorry , I will fix typo.

gichan-jang (approved):
LGTM

PR #4173, opened 2023-08-04
[Svace] fix svace issue

jaeyun-jung:
Code clean, fix svace issues.
1. prevent duplicated free case.
2. handle returned val when mapping incoming buffer.


again4you (approved):
LGTM!

gichan-jang (approved):
LGTM

PR #4177, opened 2023-08-09
fix: decoder/bounding-box mobilenet threshold logic @open sesame 8/9 17:02

### LINKED ISSUE #4137: tensor-decoder / ssd-mobilenet biased towards first classes
I believe I found an issue in the box decoding of ssd mobilenet (without post processing).

The function iterates over class scores and breaks at the first class above the selected detection threshold (sigmoid_threshold).

However, multiple classes could theoretically have a score above this threshold. In particular, subsequent classes could have a higher score than the first one above the threshold (for example: threshold of 0.5, score for bus 0.6, score for truck 0.7, the chosen class would be bus since it appears first in the list).

https://github.com/nnstreamer/nnstreamer/blob/e7439ca2e78fdc877a86963683dcd2a54ee6d906/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c#L1399

Can you confirm that this is indeed a bug? Then I can submit a fix.

Thanks!

myungjoo:
Let's not ignore labels that may have larger score after the first label exceeding the threshold.

Fixes #4137

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -1391,6 +1391,7 @@ update_centroids (void **pdata, GArray * boxes)
 #define _get_object_i_mobilenet_ssd(bb, index, total_labels, boxprior, boxinputptr, detinputptr, result) \
   do { \
     unsigned int c; \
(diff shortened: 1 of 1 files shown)


songgot (approved):
LGTM

gichan-jang (approved):
LGTM

PR #4182, opened 2023-08-10
tensor_query: nnstreamer-edge dependency fix

myungjoo:
1. nnstreamer-edge.h is from another package. Use #include appropriately.
2. dependency on nnstreamer-edge is not expressed proprely in meson.

Potential issue: Android build w/ nnstreamer-edge might need further care in /jni


myungjoo:
More commits incoming. Do not merge.

niley7464 (approved):
LGTM!

PR #4183, opened 2023-08-14
Clean up build scripts for datarepo

myungjoo:
1. Make it 'feature' so that it can be resolved automatically.
2. Because it has an additional dependency, don't make it default true.
3. Force enable datarepo for Ubuntu/Tizen

CC: @songgot


songgot (approved):
LGTM :+1:

jaeyun-jung (approved):

PR #4202, opened 2023-08-25
memleak fix: tensor_split forgot to free things. @open sesame 8/31 09:05

myungjoo:
1. split->tensorseg is not freed when overwritten.
2. tensordim pointers are never freed but new-ed frequently.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_split.c
+++ b/gst/nnstreamer/elements/gsttensor_split.c
@@ -560,6 +560,15 @@ gst_tensor_split_change_state (GstElement * element, GstStateChange transition)
 }
 
 /**
+ * @brief Glib Array Clear Function
+ */
+static void
+_clear_tensorseg (tensor_dim ** element)
+{
+  g_free (*element);
+}
+
+/**
  * @brief Get property (gst element vmethod)
  */
 static void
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/elements/gsttensor_split.c
```diff
@@ -202,10 +202,13 @@ static void
...
 {
   GstTensorSplit *split;
+  guint i;
 
   split = GST_TENSOR_SPLIT (object);
   gst_tensor_split_remove_src_pads (split);
   g_list_free (split->tensorpick);
+  for (i = 0; i < split->tensorseg->len; i++)
+    g_free (g_array_index (split->tensorseg, tensor_dim *, i));
   g_array_free (split->tensorseg, TRUE);
```
songgot, line 212: As free_segment of g_array_free (split->tensorseg, TRUE) is set to TRUE, elements are automatically freed. right ?

gst/nnstreamer/elements/gsttensor_split.c
```diff
@@ -594,25 +597,36 @@ gst_tensor_split_set_property (GObject * object, guint prop_id,
...
+
       split->num_tensors = g_strv_length (strv);
-      split->tensorseg =
-          g_array_sized_new (FALSE, FALSE, sizeof (tensor_dim *),
-          split->num_tensors);
+      if (NULL == tensorseg) {
+        split->tensorseg =
+            g_array_sized_new (FALSE, FALSE, sizeof (tensor_dim *),
+            split->num_tensors);
+
```
jaeyun-jung, line 607: With @songgot 's opinion, I suggest adding clear function here, then items will be release when calling `g_array_free (split->tensorseg, TRUE);`.

```g_array_set_clear_func (split->tensorseg, g_free);```

File: gst/nnstreamer/elements/gsttensor_split.c (hunk 1 above)
Comment (myungjoo) [reply gst/nnstreamer/elements/gsttensor_split.c:212]: with ```g_array_free ( array, TRUE)```, ```tensor_dim``` is freed. but ```tensor_dim[i]``` are not freed.

Comment (songgot) [reply gst/nnstreamer/elements/gsttensor_split.c:212]: Thank you for the detailed description. :+1:

Comment (myungjoo) [reply gst/nnstreamer/elements/gsttensor_split.c:212]: To have tensor_dim freed with g_array_free (array, TRUE), it should've been ```g_array_size_new (FALSE, FALSE, sizeof (tensor_dim), split->numbers)```.

Comment (myungjoo) [reply gst/nnstreamer/elements/gsttensor_split.c:212]: Anyway, @jaeyun-jung 's idea looks better.

File: gst/nnstreamer/elements/gsttensor_split.c [hunk 3]
```diff
@@ -594,25 +597,36 @@ gst_tensor_split_set_property (GObject * object, guint prop_id,
...
+        for (i = 0; i < split->num_tensors; i++) {
+          tensor_dim *d = g_new0 (tensor_dim, 1);
+          g_assert (d != NULL);
```
wooksong, line 610: ```g_new0()``` aborts the program on failure, which means this line looks unnecessary.

gst/nnstreamer/elements/gsttensor_split.c
myungjoo: https://docs.gtk.org/glib/type_func.Array.set_clear_func.html

```g_array_set_clear_func (split->tensorseg, g_free);``` will do ```g_free (  tensor_dim ** )```, which is still the container of the element. I'll add a custom free function here.

gst/nnstreamer/elements/gsttensor_split.c
myungjoo: +1

jaeyun-jung (approved):

wooksong (approved):
LGTM

PR #4205, opened 2023-08-25
memleak fix: datareposrc json parse leaking.

myungjoo:
set-property may be called repeatedly.
Do not overwrite allocated pointers.

CC: @songgot : please take a look if there are similar errors (overwriting allocated pointers).


songgot (approved):
Okay, Thank you. LGTM

jaeyun-jung (approved):

PR #4218, opened 2023-08-29
[Subplugin/Flatbuf] util to get nth info

jaeyun-jung:
Code clean, remove unnecessary copy of tensors-info and use util to get nth info.

### DIFF
--- a/ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
+++ b/ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
@@ -74,6 +74,8 @@ fbc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
   GstMemory *in_mem, *out_mem;
   GstMapInfo in_info;
   guint mem_size;
+  GstTensorInfo *_info;
+
   UNUSED (priv_data);
 
   if (!in_buf || !config) {
@@ -108,12 +110,15 @@ fbc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
     std::string _name = tensor->Get (i)->name ()->str ();
     const gchar *name = _name.c_str ();
 
-    config->info.info[i].name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
-    config->info.info[i].type = (tensor_type) tensor->Get (i)->type ();
+    _info = gst_tensors_info_get_nth_info (&config->info, i);
+
+    g_free (_info->name);
+    _info->name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
+    _info->type = (tensor_type) tensor->Get (i)->type ();
     tensor_data = tensor->Get (i)->data ();
 
     for (guint j = 0; j < NNS_TENSOR_RANK_LIMIT; j++) {
-      config->info.info[i].dimension[j] = tensor->Get (i)->dimension ()->Get (j);
+      _info->dimension[j] = tensor->Get (i)->dimension ()->Get (j);
     }
     mem_size = VectorLength (tensor_data);
 
--- a/ext/nnstreamer/tensor_converter/tensor_converter_flexbuf.cc
+++ b/ext/nnstreamer/tensor_converter/tensor_converter_flexbuf.cc
@@ -91,6 +91,8 @@ flxc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
   GstMemory *in_mem, *out_mem;
   GstMapInfo in_info;
   gsize mem_size;
+  GstTensorInfo *_info;
+
   UNUSED (priv_data);
 
   if (!in_buf || !config) {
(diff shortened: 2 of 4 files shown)


ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
```diff
@@ -108,12 +110,14 @@ fbc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
     std::string _name = tensor->Get (i)->name ()->str ();
     const gchar *name = _name.c_str ();
 
-    config->info.info[i].name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
-    config->info.info[i].type = (tensor_type) tensor->Get (i)->type ();
+    _info = gst_tensors_info_get_nth_info (&config->info, i);
+
+    _info->name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
```
wooksong, line 115: If ```name``` is NULL, ```g_strdup()``` returns NULL..

jaeyun-jung: Now we initialize the tensor name in tensors-info struct as null, and do not set this if a tensor does not use its name. I am not sure but this may cause unexpected error.
If necessary we need considering another PR to change default value of tensor name.

myungjoo, line 115: What if _info->name is already allocated?

ext/nnstreamer/tensor_converter/tensor_converter_flexbuf.cc
```diff
@@ -124,12 +126,14 @@ flxc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
     flexbuffers::String _name = tensor[0].AsString ();
     const gchar *name = _name.c_str ();
 
-    config->info.info[i].name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
-    config->info.info[i].type = (tensor_type) tensor[1].AsInt32 ();
+    _info = gst_tensors_info_get_nth_info (&config->info, i);
+
+    _info->name = (name && strlen (name) > 0) ? g_strdup (name) : NULL;
```
myungjoo, line 131: With g_strdup, you may incur a memory leak here as well.

ext/nnstreamer/tensor_decoder/tensordec-flexbuf.cc
```diff
@@ -125,41 +125,45 @@ flxd_decode (void **pdata, const GstTensorsConfig *config,
...
+      tensor_type type = _info->type;
 
       fbb.Vector (tensor_key, [&] () {
         fbb += tensor_name;
         fbb += type;
-        fbb.Vector (flxd_config.info.info[i].dimension, NNS_TENSOR_RANK_LIMIT);
+        fbb.Vector (_info->dimension, NNS_TENSOR_RANK_LIMIT);
         fbb.Blob (input[i].data, input[i].size);
       });
       g_free (tensor_key);
```
myungjoo, line 169: Please check if it's safe to call ```g_free (tensor_name)``` after doing ```fbb += tensor_name```. (I don't know.. I just feel something weird with this)

ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
jaeyun-jung: Internally tensor-converter calls convert() callback after initializing the tensors-config struct.
To clearly release tensors-config, I will append init code in this callback.
```
  gst_tensors_config_free (config);
  gst_tensors_config_init (config);
```

ext/nnstreamer/tensor_decoder/tensordec-flexbuf.cc
jaeyun-jung: operator+ seems getting a param as const

ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
jaeyun-jung: updated. plz review the changes again.

ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
```diff
@@ -74,13 +74,17 @@ fbc_convert (GstBuffer *in_buf, GstTensorsConfig *config, void *priv_data)
...
+  GstTensorInfo *_info;
+
   UNUSED (priv_data);
 
   if (!in_buf || !config) {
     ml_loge ("NULL parameter is passed to tensor_converter::flatbuf");
     return NULL;
   }
 
+  gst_tensors_config_init (config);
```
myungjoo, line 86: gsttensor_converter.c::2429 may call it before this callback.

ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
myungjoo: Although gsttensor_converter.c::2429 calls ```gst_tensor_config_init``` before calling ```->convert```, in case someone call this function from another context (this is a callback available to other modules), let's simply add ```g_free(_info->name)``` before line 117, which will do nothing if _info->name is NULL as supposed.

ext/nnstreamer/tensor_converter/tensor_converter_flexbuf.cc
myungjoo: Same here. Let's call g_free, instead of gst_tensors_config_init (which can be duplicated in the context of gst-converter)

ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc
jaeyun-jung: updated

anyj0527 (approved):
LGTM

songgot (approved):
LGTM

PR #4224, opened 2023-08-30
memleak fix: tensor_demux forgot to free strv @open sesame 8/30 15:34

myungjoo:
g_strsplit_set()'s return should be freed by the caller.


songgot (approved):
LGTM

gichan-jang (approved):
LGTM

PR #4227, opened 2023-08-30
memleak fix: converter test case fix

myungjoo:
converter gtest code mem leak.

converter element mem leak fix is coming with another PR that fixes multiple elements: #4238


songgot (approved):
LGTM

PR #4230, opened 2023-09-01
[Split] init dim as 0

jaeyun-jung:
The initial value of tensor dimension is 0. Do not fill remained dims as 1.


myungjoo (approved):

niley7464 (approved):
LGTM! Thank you 👍

PR #4238, opened 2023-09-06
[Doc/API] Clarify GstMemory ownership of gst_tensor_meta_info_append_header

myungjoo:
## What this PR is now

This PR has been reduced to **documentation only**.

`gst_tensor_meta_info_append_header()` copies the content of the given `GstMemory` into a newly allocated one and never takes ownership of the input. The doc comment said only that the *returned* memory must be unref'ed, leaving the fate of the *input* memory ambiguous. That ambiguity is what produced a series of leaks in the converter / mux / transform / filter flexible-tensor paths.

This adds an `@note` to both the implementation (`nnstreamer_plugin_api_impl.c`) and the public header (`nnstreamer_plugin_api.h`, which is what API consumers actually read) stating that the input memory is not consumed and remains the caller's responsibility.

No functional change.

## Why the original code changes were dropped

The original 2023 version of this PR fixed the actual leaks by rewriting the flexible-tensor paths to mutate buffers in place. Every one of those leaks has since been fixed on `main` by the other resolution — keep building a fresh output buffer and simply `gst_memory_unref()` the memory passed to `gst_tensor_meta_info_append_header()`:

| original target | current `main` |
|---|---|
| `gsttensor_converter.c` flex tensor | `gst_memory_unref (mem);` after `append_header` |
| `gsttensor_mux.c` flex tensor | same pattern |
| `gsttensor_transform.c` | literally the `old` / temp-variable pattern this PR proposed |
| `tensor_filter.c` | unref'ed in both the `invoke_dynamic` and `allocate_in_invoke` flexible paths |
| `nnstreamer_plugin_api_impl.c` caller | `gst_memory_unref (in_mem[i]);` |

The second motivation in the original description — `collect_pad->buf` becoming dangling after buffer replacement — no longer applies either. `gst_tensor_mux_collected()` now passes a locally allocated `gst_buffer_new()` filled by `gst_tensor_mux_collect_buffer()`, not a `GstCollectData` buffer, so replacing the buffer instance is safe.

Rebasing the original in-place-mutation diff would therefore re-litigate an already-merged fix with a larger, riskier change. Only the piece that is still genuinely missing — the ownership note — is kept here.

### DIFF
--- a/gst/nnstreamer/include/nnstreamer_plugin_api.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api.h
@@ -108,6 +108,10 @@ gst_tensor_meta_info_parse_memory (GstTensorMetaInfo * meta, GstMemory * mem);
  * @param[in] meta tensor meta structure
  * @param[in] mem pointer to GstMemory
  * @return Newly allocated GstMemory (Caller should free returned memory using gst_memory_unref())
+ * @note This does not take the ownership of @a mem; its content is copied into
+ *       the returned memory. If the caller owns a reference of @a mem (e.g., the
+ *       memory acquired by gst_tensor_buffer_get_nth_memory ()), the caller
+ *       should unref @a mem as well as the returned memory.
  */
 extern GstMemory *
 gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem);
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -1633,6 +1633,10 @@ gst_tensor_meta_info_parse_memory (GstTensorMetaInfo * meta, GstMemory * mem)
  * @param[in] meta tensor meta structure
  * @param[in] mem pointer to GstMemory
  * @return Newly allocated GstMemory (Caller should free returned memory using gst_memory_unref())
+ * @note This does not take the ownership of @a mem; its content is copied into
+ *       the returned memory. If the caller owns a reference of @a mem (e.g., the
+ *       memory acquired by gst_tensor_buffer_get_nth_memory ()), the caller
+ *       should unref @a mem as well as the returned memory.
  */
 GstMemory *
 gst_tensor_meta_info_append_header (GstTensorMetaInfo * meta, GstMemory * mem)


myungjoo:
> [Relayed on behalf of the author by a Claude agent]

Rescoped this PR to documentation only, per the review above.

I re-verified every item in that review against current `main` (`6bea3717`): the converter, mux, transform, `tensor_filter` and `nnstreamer_plugin_api_impl.c` call sites all already `gst_memory_unref()` the memory handed to `gst_tensor_meta_info_append_header()`, and `gst_tensor_mux_collected()` now passes a locally allocated buffer rather than a `GstCollectData` one, so the dangling-`collect_pad->buf` concern is moot as well. The original in-place-mutation diff would re-litigate an already-merged fix, so it has been dropped.

What remains is the one piece that was genuinely still missing: an `@note` documenting that `append_header()` does not take ownership of the input `GstMemory`. I confirmed against the implementation that it only maps/unmaps the input and copies into a freshly allocated block, so the caller retains whatever reference it holds. The note is now on both the implementation and the public header — the header is the declaration API consumers actually read, and the original PR only touched the `.c`.

Branch rebased onto current `main`; the PR is no longer conflicting. No functional change, so the previously noted absence of valgrind gating on the converter/mux/transform/filter flexible paths is unaffected by this PR — that remains a separate follow-up if anyone touches the ownership contract itself.

myungjoo:
> [Relayed on behalf of the author by a Claude agent]

CI is green on `b7d98b99` — all 18 checks pass. Removing the `DO NOT MERGE` label.

Two jobs failed on the first pass and both were re-run; neither was related to this change, which adds only comment lines and zero code:

- **Ubuntu pdebuild** — `packages.microsoft.com` returned `403 Forbidden` on the GitHub runner image, so `apt-get update` exited non-zero. Because the workflow runs `apt-get update && apt-get install -y pbuilder ...` under `bash -e`, and bash's errexit does not fire for a non-final command in an `&&` list, the install was silently skipped and the step still reported success — surfacing 20 seconds later as `sudo: pbuilder: command not found`. Passed on re-run. The workflow fragility is worth fixing separately.
- **Valgrind x86_64** — `datareposink.writeFlexibleTensors_n` failed at `unittest_datareposink.cc:578` (`Expected: (file_info) != (nullptr)`), a filesystem race whose assertion already carries a 5 x 50 ms retry loop that is too short under valgrind. The `ubuntu-22.04-arm` variant of the same job passed, and an unrelated PR failed the sibling `datareposrc.fps30ReadFlexibleTensors` timing test in the same window (being fixed in #4872, which covers only the *src* side). Passed on re-run.

An independent re-review of the rescoped diff came back clean: the `@note` matches the implementation on every path, all 13 call sites in the tree behave as the note describes, the doxygen-tag checker rules are satisfied, and the two copies of the note are byte-for-byte identical.

One optional nit was raised and deliberately left out to keep this PR to the note: the `@return` line still does not mention that the function returns NULL on failure (it has four such paths). Happy to fold that in if a reviewer would prefer it here rather than as a follow-up.

PR #4244, opened 2023-09-07
[pytorch] Fix load model with use_gpu option with recent pytorch

anyj0527:
- Drop pytorch support for version less than 1.2.0 
- In the recent pytorch, `model->to (device)` does not work properly.
- Use `torch::jit::load (model, torch::Device)` instead.

Tested with yolov8 torchscript model with pytorch v2.0.1

**Self evaluation:**


jaeyun-jung (approved):

songgot (approved):
LGTM

PR #4248, opened 2023-09-11
[build] Fix recent meson usage

anyj0527:
- meson may change the default value of `run_command`'s `check` arg (false -> true). Set `check : false` in omitted lines.
- meson complains as `starting in "lib" only works by accident and is not portable`. Change `libSNPE` to `SNPE`.

**Self evaluation:**


jaeyun-jung (approved):

songgot (approved):
LGTM

PR #4254, opened 2023-09-14
tizen: enable/disable nnstreamer-edge explicitly.

myungjoo:
For Tizen distro, do not let "auto" determine essential features. This will prevent regression that implicitly disabling nnstreamer-edge.


jaeyun-jung (approved):

anyj0527 (approved):
LGTM✔

PR #4262, opened 2023-09-19
[tensor_filter] Add tensor filter config property @open sesame 09/20 10:05

linuxias:
tensor filter configuration is very complex, so when you set filter,
 property settings become longer and more complicated, making it
 difficult to find them at once even if human errors occur.

 If config is used, readability can be increased and human errors
 can be reduced due to organized properties.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped
2. Run test: [*]Passed [ ]Failed [ ]Skipped

```bash
ssat -t runTest.sh
==================================================
    Test Group nnstreamer_filter_tensorflow2_lite Starts.
[PASSED] 1:gst-launch of case 1
[PASSED] 1:Golden test comparison
[PASSED] 1:gst-launch of case 1
[PASSED] 1:Golden test comparison(with config_file_golden.0)
[PASSED] 2F_n:gst-launch of case 2F_n
[PASSED] 2F_n(with config_file_golden.1):gst-launch of case 2F_n(with config_file_golden.1)
[PASSED] 3F_n:gst-launch of case 3F_n
[PASSED] 3F_n(with config_file_golden.2):gst-launch of case 3F_n(with config_file_golden.2)
[PASSED] 4_n:gst-launch of case 4_n
[PASSED] 5:gst-launch of case 5
[PASSED] 2-1:accelerator set test
[PASSED] 2-2:accelerator set test
[PASSED] 2-3:accelerator set test
[PASSED] 2-4:accelerator set test
[PASSED] 2-5:accelerator set test
[PASSED] 2-6:accelerator set test
[PASSED] 2-7:accelerator set test
[PASSED] 2-8:accelerator set test
[PASSED] 2-9:accelerator set test
[PASSED] 2-10:accelerator set test
[PASSED] 2-11:accelerator set test
[PASSED] 2-12:accelerator set test
[PASSED] 2-13:accelerator set test
[PASSED] 2-14:accelerator set test
[PASSED] 2-15:accelerator set test
[PASSED] 2-16:accelerator set test
[PASSED] 2-17:accelerator set test
[PASSED] 3:gst-launch of case 3
[PASSED] 3:Golden test comparison
[PASSED] 4:gst-launch of case 4
[PASSED] 4:Golden test comparison
[PASSED] 5:gst-launch of case 5
[PASSED] 5:Golden test comparison
[PASSED] 6:gst-launch of case 6
[PASSED] 6:Compare 6
[...]

PR #4264, opened 2023-09-21
test: tensor_filter/tflite2 correct postfix

myungjoo:
Negative test cases require _n postfix for the test infrastucture to recognize them. The new test cases with property config forgot to add such postfix properly.

This fixes test case infra regression with #4262


wooksong (approved):
LGTM

PR #4269, opened 2023-09-24
[Edge] Fix memory leak when finalize sync

linuxias:
Free dest_host, topic variable memory in GstEdgeSink
 when call finalize method to clear GstEdgeSink memory

**Self evaluation:**

### DIFF
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -256,10 +256,15 @@ static void
 gst_edgesink_finalize (GObject * object)
 {
   GstEdgeSink *self = GST_EDGESINK (object);
-  if (self->host) {
-    g_free (self->host);
-    self->host = NULL;
-  }
+
+  g_free (self->host);
+  self->host = NULL;
+
+  g_free (self->dest_host);
+  self->dest_host = NULL;
+
+  g_free (self->topic);
+  self->topic = NULL;
 
   if (self->edge_h) {
     nns_edge_release_handle (self->edge_h);
--- a/gst/edge/edge_src.c
+++ b/gst/edge/edge_src.c
@@ -228,20 +228,14 @@ gst_edgesrc_class_finalize (GObject * object)
   GstEdgeSrc *self = GST_EDGESRC (object);
   nns_edge_data_h data_h;
 
-  if (self->host) {
-    g_free (self->host);
-    self->host = NULL;
-  }
+  g_free (self->host);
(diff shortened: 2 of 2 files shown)


gichan-jang (approved):
LGTM

songgot (approved):
LGTM

yeonykim2 (approved):
LGTM

gst/edge/edge_sink.c
```diff
@@ -265,6 +265,16 @@ gst_edgesink_finalize (GObject * object)
...
+  if (self->dest_host) {
+    g_free (self->dest_host);
+    self->dest_host = NULL;
+  }
+
+  if (self->topic) {
+    g_free (self->topic);
+    self->topic = NULL;
+  }
+
```
wooksong, line 277:
```suggestion
  g_free (self->dest_host);
  self->dest_host = NULL;
  g_free (self->topic);
  self->topic = NULL;
```

https://developer-old.gnome.org/glib/stable/glib-Memory-Allocation.html#g-free

linuxias: Thanks for your comment!
I know, if mem is Null, it simply returns, so there is no need to check mem against before calling this function.
But i want to unify exception condition other codes in this function.
I think, code uniformity can be an important factor in readability.

So i want to maintain current patch. 
Please advise if the modification is necessary.

Thanks :)

wooksong: @linuxias Then, it is better to fix the other lines for consistency.

linuxias: @wooksong I fixed all lines what free memory in sink and source finalize method.
Please review! Thanks!

gst/edge/edge_sink.c
```diff
@@ -256,15 +256,18 @@ static void
...
+  self->host = NULL;
+
+  g_free (self->dest_host);
+  self->dest_host = NULL;
+
+  g_free (self->topic);
+  self->topic = NULL;
+
+  nns_edge_release_handle (self->edge_h);
+  self->edge_h = NULL;
```
wooksong, line 270:
```suggestion
  if (self->edge_h) {
    nns_edge_release_handle (self->edge_h);
    self->edge_h = NULL;
  }
```
```nns_edge_release_handle``` definitely requires a null check. Otherwise, it complains about that using ```g_critical``` or the equivalent.

PR #4277, opened 2023-10-11
filter: handle error of convert_to_meta

myungjoo:
gst_tensor_info_convert_to_meta may return error
if the configured output tensor is invalid.
Handle the error before feeding it to another function, that may access uninitialized value.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -771,8 +771,15 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
 
     hsize = 0;
     if (out_flexible && !priv->prop.invoke_dynamic) {
-      gst_tensor_info_convert_to_meta (&prop->output_meta.info[i],
+      gboolean ret = FALSE;
+      ret = gst_tensor_info_convert_to_meta (&prop->output_meta.info[i],
           &out_meta[i]);
+      if (TRUE != ret) {
+        ml_loge_stacktrace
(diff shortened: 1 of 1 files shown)


songgot (approved):
LGTM

SeoHyungjun (approved):
LGTM!

I have a personal question.
Is it correct to express it as "if (TRUE != ret)" rather than the code below?
It seems like it improves readability at the cost of making the code longer, but I'm not sure which method is preferable.
```
if ( !gst_tensor_info_convert_to_meta (&prop->output_meta.info[i], &out_meta[i])) {
```
or 
```
ret = gst_tensor_info_convert_to_meta (&prop->output_meta.info[i],
          &out_meta[i]);
if (!ret) {
```

jaeyun-jung (approved):

myungjoo:
> LGTM!
> 
> I have a personal question. Is it correct to express it as "if (TRUE != ret)" rather than the code below? It seems like it improves readability at the cost of making the code longer, but I'm not sure which method is preferable.
> 
> ```
> if ( !gst_tensor_info_convert_to_meta (&prop->output_meta.info[i], &out_meta[i])) {
> ```
> 
> or
> 
> ```
> ret = gst_tensor_info_convert_to_meta (&prop->output_meta.info[i],
>           &out_meta[i]);
> if (!ret) {
> ```

```
if ( !gst_tensor_info_convert_to_meta ...
```
or
```
if (FALSE == gst_tensor_info_convert_to_meta ...
```

is more prefereable.

PR #4279, opened 2023-10-13
[decoder] Let user set yolo decoder parameters

anyj0527:
- Let option3 of yolo decoders can handle user provided parameters.
  For yolov5 and yolov8 mode: The option3 requires up to 3 numbers, which tell
    - whether the output values are scaled or not 0: not scaled (default), 1: scaled (e.g., 0.0 ~ 1.0)
    - the threshold of confidence (optional, default set to 0.25)
    - the threshold of IOU (optional, default set to 0.45)
  An example of option3 is "option3=0:0.65:0.6"

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -38,16 +38,13 @@
  *          This is independent from option1
  * option3: Any option1-dependent values
  *          !!This depends on option1 values!!
- *          for yolov5 mode:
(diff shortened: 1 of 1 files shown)


myungjoo (approved):
Now, consider having some test cases for yolo. (comparing against golden)

myungjoo:
The CI error w/ GBS TIzen build is to be fixed by https://github.com/nnstreamer/nntrainer/pull/2334 
You may proceed without passing that one.

niley7464 (approved):
LGTM!

Issue #3903, opened 2022-09-13
Descriptive names and better documentation for `tensor_decoder` properties 


shortcipher3:
**Is your feature request related to a problem? Please describe.**
It would be great if the properties on `tensor_decoder` had descriptive names and better documentation.  It is really confusing with the `optionX` in the documentation.

**Describe the solution you'd like**
It would be great to do one of the following:
* provide meaningful named options and in the description explain which types of models it is applicable to (eg classifier, object detection, etc.) and what it does with example values
* leave the names the same and in the description explain what the options do for each type of model and what it does with example values
* leave the names and the descriptions the same, but link to documentation that explains what the options are

**Additional context**
```
$ gst-inspect-1.0 tensor_decoder
...
Element Properties:
...
  option1             : option for specific decoder modes, 1st one.
                        flags: readable, writable
                        String. Default: null
  option2             : option for specific decoder modes, 2nd one.
                        flags: readable, writable
                        String. Default: null
  option3             : option for specific decoder modes, 3rd one.
                        flags: readable, writable
                        String. Default: null
  option4             : option for specific decoder modes, 4th one.
                        flags: readable, writable
                        String. Default: null
  option5             : option for specific decoder modes, 5th one.
                        flags: readable, writable
                        String. Default: null
  option6             : option for specific decoder modes, 6th one.
                        flags: readable, writable
                        String. Default: null
  option7             : option for specific decoder modes, 7th one.
                        flags: readable, writable
                        String. Default: null
  option8             : option for specific decoder modes, 8th one.
                        flags: readable, writable
                        String. Default: null
  option9             : option for specific decoder modes, 9th one.
                        flags: readable, writable
                        String. Default: null
```

myungjoo:
The problem is that each subplugin (e.g., bounding-boxes, classification, ...) may have different options, which is why we have allocated option{#N} from the beginning.

1. Having different property names is not possible. We do not know which subplugins will be available for a pipeline before actually running the pipeline. Users may install additional subplugins anytime.

    - In order to allow your suggestion 1, we need to make every subplugins built-in. So, I don't prefer this.

2. In the similar rationale, this is not preferable as well.

3. This would be the viable option. We can place documents for each "upstreamed" subplugins at /ext/nnstreamer/tensor_decoder/README.md and have the link described in gst-inspect.

    - We can further assist this with nnstreamer-check tool (and describe it in gst-inspect text as well).
[...]

PR #4283, opened 2023-10-16
decoder/filter: more info for gst-inspect.

myungjoo:
Show available frameworks / mode with gst-inspect. This is the first step to address #3903

@songgot : please add similar feature to tensor_trainer


jaeyun-jung (approved):

anyj0527 (approved):
LGTM👍

PR #4288, opened 2023-10-17
Decoder: Add detailed descriptions for the bounding box sub-plugin @open sesame 10/18 18:04 

wooksong:
In addition to https://github.com/nnstreamer/nnstreamer/pull/4284, this patch adds descriptions for the bounding box sub-plugin options.

See also: #3903 

**Self evaluation:**

```bash
 $ ./nnstreamer-check
NNStreamer version: 2.4.1
           loaded : TRUE
           path   : /home/wook/Work/NNS/nnstreamer/build/gst/nnstreamer/libnnstreamer.so

NNStreamer configuration:
============================================================
Configuration Loaded: TRUE
Configuration file path: /home/wook/Work/NNS/nnstreamer/build/nnstreamer-test.ini
    Candidates: envvar(NNSTREAMER_CONF): /home/wook/Work/NNS/nnstreamer/build/nnstreamer-test.ini
                build-config: /usr/local/etc/nnstreamer.ini
                hard-coded: /etc/nnstreamer.ini

...omission..

NNStreamer registered sub-plugins:
============================================================

[Filter]

[Decoder]
  bounding_boxes
    - option1: Decoder mode of bounding box: [mobilenet-ssd|mobilenet-ssd-postprocess|ov-person-detection|ov-face-detection|tflite-ssd|tf-ssd|yolov5|mp-palm-detection|yolov8]
    - option2: Location of the label file. This is independent from option1.
    - option3: Sub-option values that depend on option1;
        for yolov5 and yolov8 mode:
                The option3 requires up to 3 numbers, which tell
                - whether the output values are scaled or not
                   0: not scaled (default), 1: scaled (e.g., 0.0 ~ 1.0)
                - the threshold of confidence (optional, default set to 0.25)
                - the threshold of IOU (optional, default set to 0.45)
                An example of option3 is option3 = 0: 0.65:0.6 
        for mobilenet-ssd mode:
                The option3 definition scheme is, in order, as follows
                - box priors location file (mandatory)
                - detection threshold (optional, default set to 0.5)box priors location file (mandatory)
                - Y box scale (optional, default set to 10.0)
                - X box scale (optional, default set to 10.0)
                - H box scale (optional, default set to 5.0)
                - W box scale (optional, default set to 5.0)
                The default parameters value could be set in the following ways:
                 option3=box-priors.txt:0.5:10.0:10.0:5.0:5.0:0.5
                 option3=box-priors.txt
                 option3=box-priors.txt::::::
                It's possible to set only few values, using the default values for those not specified through the command line.
                You could specify respectively the detection and IOU thresholds to 0.65 and 0.6 with the option3 parameter as follow:
                 option3=box-priors.txt:0.65:::::0.6
        for mobilenet-ssd-postprocess mode:
                The option3 is required to have 5 integer numbers, which tell the tensor-dec how to interpret the given tensor inputs.
                The first 4 numbers separated by colon, ':', designate which are location:class:score:number of the tensors.
                The last number separated by comma, ',' from the first 4 numbers designate the threshold in percent.
                In other words, "option3=%i:%i:%i:%i,%i"
        for mp-palm-detection mode:
                The option3 is required to have five float numbers, as follows;
                - box score threshold (mandatory)
                - number of layers for anchor generation (optional, default set to 4)
                - minimum scale factor for anchor generation (optional, default set to 1.0)
                - maximum scale factor for anchor generation (optional, default set to 1.0)
                - X offset (optional, default set to 0.5)
                - Y offset (optional, default set to 0.5)
                - strides for each layer for anchor generation (optional, default set to 8:16:16:16)
                The default parameter value could be set in the following ways:
                 option3=0.5
                 option3=0.5:4:0.2:0.8
                 option3=0.5:4:1.0:1.0:0.5:0.5:8:16:16:16
    - option4: Video Output Dimension (WIDTH:HEIGHT). This is independent from option1.
    - option5: Input Dimension (WIDTH:HEIGHT). This is independent from option1.
    - option6: Whether to track result bounding boxes or not
                 0 (default, do not track)
                 1 (track result bounding boxes, with naive centroid based algorithm)
    - option7: Whether to log the result bounding boxes or not
                 0 (default, do not log)
                 1 (log result bounding boxes)  This is independent from option1
    - option8: Box Style (NYI)
```

## Changelog
- Apply @myungjoo 's [comment](https://github.com/nnstreamer/nnstreamer/pull/4288#discussion_r1361795891)

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
@@ -93,7 +93,7 @@
  * option6: Whether to track result bounding boxes or not
  *          0 (default, do not track)
  *          1 (track result bounding boxes, with naive centroid based algorithm)
- * option7: Whether to log the result boungding boxes or not
+ * option7: Whether to log the result bounding boxes or not
  *          0 (default, do not log)
  *          1 (log result bounding boxes)
  * option8: Box Style (NYI)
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
```diff
@@ -2210,6 +2210,72 @@ void
 init_bb (void)
 {
   nnstreamer_decoder_probe (&boundingBox);
+
+  {
+    g_autofree gchar *sub_desc = g_strjoinv ("|", (GStrv) bb_modes);
+    const gchar *desc =
```
jaeyun-jung, line 2216: const? possible mem leak.

myungjoo: Although init_bb is supposed to be called only once, I'd suggest to have a static variable and let it be gfree-ed if called multiple times. Then, you may choose to gfree it at fini_bb() if you've decided to have a global-static.

myungjoo:
Anyway, except for the possibility of leaking a gchar pointer, this looks great.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c
wooksong: > const? possible mem leak.

Is it freed by https://github.com/nnstreamer/nnstreamer/blob/e7710dff2b94ef809f4794367240d5d4f13e5306/gst/nnstreamer/nnstreamer_subplugin.c#L54 at the time when ```unregister_subplugin``` is called?

myungjoo (approved):

jaeyun-jung (approved):

PR #4291, opened 2023-10-19
[util] Fix gst_tensors_info_to_string

anyj0527:
- Let the method returns string ended with "]" always.
- Let get_dimension_string returns NULL when given dim is filled with 0

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
@@ -852,10 +852,12 @@ gst_tensors_info_to_string (const GstTensorsInfo * info)
     dim = gst_tensor_get_dimension_string (_info->dimension);
 
     g_string_append_printf (gstr, "{\"%s\", %s, %s}%s",
-        name ? name : "", type, dim, (i == info->num_tensors - 1) ? "]" : ", ");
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
```diff
@@ -1098,7 +1100,11 @@ gst_tensor_parse_dimension (const gchar * dimstr, tensor_dim dim)
 gchar *
 gst_tensor_get_dimension_string (const tensor_dim dim)
 {
-  return gst_tensor_get_rank_dimension_string (dim, NNS_TENSOR_RANK_LIMIT);
+  gchar *res =
+      gst_tensor_get_rank_dimension_string (dim, NNS_TENSOR_RANK_LIMIT);
+  if (!res || *res == '\0')
```
myungjoo, line 1105: if ```*res == '\0'```, g_free before return NULL.

anyj0527: Thanks! I updated the code, PTAL.

niley7464 (approved):
LGTM!

myungjoo (approved):

Issue #3017, opened 2021-01-22
Tensor-Composite (for Visualization Aid)


myungjoo:
This is very similar with tensor-merge.
This is supposed to do video-composite with tensor streams (that can be decoded into ```video/x-raw``` with ```tednsor-decoder::directvideo```) and video positions encoded in tensor streams (e.g., in a "tensors" stream, first tensor has video content and second tensor has the top-left coordinates and effective size)

Usage example:
With src0: 640:480:3, 4 and src1 : 200:200:1, 4, sink0: 640:480:3
```
... ! tensor-composite name=EXAMPLE mode=video-merge color=0:original-RGB,1:monotone-red background=0 z-order=0,1 ! ...
```

The second tensor (proposal) [4] = { TOP-LEFT POSITION X, TOP-LEFT POSITION Y, WIDTH, HEIGHT }, this is ignored if it is designated as "background". If it is the designated background, it is already overriden with [ 0, 0, tensor-size-x, tensor-size-y ].

The example draws src1's points at the position and size of src1's 2nd tensor value with "red" if the value is non-zero. We may imagine that if it is "grayscale-blue", then, it will paint [ R=0, G=0, B=255 * tensor-value / max-value ].

TODO:
- [ ] Write tensor-merge
- [ ] (for optimization with tensor-decoder & tensor-converter) Allow tensor-decoder to shoot other/tensors, generating the above src0/src1.

jaeyun-jung:
tensor-flexcrop

input (other/tensors)
- 1st : origin tensor [c,w,h]
- 2nd : list of region info [x,y,w,h] / max num 16  ---> fixed size of buffer (num_crop + sizeof (guint) * 4 * 16)

output (other/tensors)
- original (option) + cropped
- how can we support dynamic dimension? (fixed dimension (limit of max w/h) + meta, or dynamic dimension on t-demux/t-filter)

myungjoo:
> tensor-flexcrop
> 
> input (other/tensors)
> 
> * 1st : origin tensor [c,w,h]
> * 2nd : list of region info [x,y,w,h] / max num 16  ---> fixed size of buffer (num_crop + sizeof (guint) * 4 * 16)
> 
> output (other/tensors, or other/tensors-cropped -flex?)
> 
> * original (option) + cropped
> * how can we support dynamic dimension? (fixed dimension (limit of max w/h) + meta, or dynamic dimension on t-demux/t-filter)

[...]

PR #4298, opened 2023-10-26
tensor_compositor suggestion

myungjoo:
To jump start #3017, a usage example is added.
Let's add more idea before writing codes.

This can be used to put cropped tensors into a normalized dimension, too.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_compositor.c
+++ b/gst/nnstreamer/elements/gsttensor_compositor.c
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * Copyright (C) 2023 Samsung Electronics Co., Ltd.
+ *
+ * @file        gsttensor_compositor.c
+ * @date        26 Oct 2023
+ * @brief       GStreamer element to composite multiple tensor streams
+ * @see         https://github.com/nnstreamer/nnstreamer
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs except for NYI items.
+ */
+
+
+/**
+ * SECTION:element-tensor_compositor
+ *
+ * tensor_compositor is a GStreamer element to composite tensor streams.
+ *
+ * tensor_compositor can have multiple sink pads of other/tensors,
+ * and has a single source pad of other/tensors.
+ * It is similar to the compositor element, which composites video streams.
+ * And intuitively, the input tensor streams can be regarded as video streams
+ * to be composited by the element.
+ *
+ * If the two input tensors are to be composited disjointedly without any
+ * padding or filling, it can be done by tensor_merge, too.
+ *
+ * The input streams may have different dimensions or formats; however,
+ * the input streams should have the same number of tensors (num_tensors),
+ * and the same types for every matching tensor (e.g., uint8, float32, ...).
+ *
+ * The output is always other/tensors,format=static.
+ *
+ * <refsect2>
+ * <title>Example launch line</title>
+ * |[
+ * gst-launch-1.0 videotestsrc ! videoconvert ! video/x-raw,format=RGB,width=640,height=480 ! tensor_converter ! cmp.sink_0
+ *     videotestsrc ! videoconvert ! video/x-raw,format=RGB,width=1920,height=1080 ! tensor_converter ! cmp.sink_1
+         tensor_compositor name=cmp num_tensors=1 types=uint8 dimensions=3:1920:1080 sink_0::z_order=1 sink_1::z_order=2 sink0::pos=0:640:480 sink1::pos=0:0:0 merge_logic=add ! tensor_decoder mode=direct_video ! videoconvert ! autovideosink
+ * ]|
+ * </refsects>
+ */


gst/nnstreamer/elements/gsttensor_compositor.c
```diff
@@ -0,0 +1,42 @@
...
+ * and the same types for every matching tensor (e.g., uint8, float32, ...).
+ *
+ * The output is always other/tensors,format=static.
+ *
+ * <refsect2>
+ * <title>Example launch line</title>
+ * |[
+ * gst-launch-1.0 videotestsrc ! videoconvert ! video/x-raw,format=RGB,width=640,height=480 ! tensor_converter ! cmp.sink_0
+ *     videotestsrc ! videoconvert ! video/x-raw,format=RGB,width=1920,height=1080 ! tensor_converter ! cmp.sink_1
+         tensor_compositor name=cmp num_tensors=1 types=uint8 dimensions=3:1920:1080 sink_0::z_order=1 sink_1::z_order=2 sink0::pos=0:640:480 sink1::pos=0:0:0 merge_logic=add ! tensor_decoder mode=direct_video ! videoconvert ! autovideosink
```
wooksong, line 39: We can implement ```tensor_compositor``` as a GstBin element that contains ```tensor_transform```, ```tensor_merge```, and etc.

myungjoo: Can you do "composite": e.g., [ [10, 11 ], [20, 21] ] ,  [ [ 100 ] ] =composite=> [ [10, 100 ], [ 20, 21 ] ] or [ [10, 11], [100, 21] ], depending on the properties, with transform and merge? (you need to make them overlapped or summed)

myungjoo:

## Review — tensor_compositor suggestion

Reviewing this as what it declares itself to be — a design sketch to jump-start #3017, not an implementation. The design itself reads well: the "input tensor streams behave like video streams to a `compositor`" framing is the right mental model, and calling out that the disjoint no-padding case is already `tensor_merge` is a useful boundary. Comments below are about the form it lands in and about details of the proposal that will matter once someone implements it.

### 1. (Major) A `.c` file with no code is the wrong container for a proposal

`gst/nnstreamer/elements/gsttensor_compositor.c` contains only a header block and a `SECTION:` doc comment. It is not added to `gst/nnstreamer/meson.build`, so it is never compiled, never linted, and never covered by any test — but it sits in the elements directory looking exactly like a real element. Anyone grepping `gst/nnstreamer/elements/` for available elements will find it, and `gtk-doc`-style `SECTION:` blocks are harvested by documentation tooling, so this can surface as a documented element that does not exist.

Please move the content to `Documentation/` (or into #3017 itself, which is where the discussion belongs) and keep `gst/nnstreamer/elements/` for things that build. If you want the design to live next to the future implementation, a `.md` in that directory would at least not be mistaken for a translation unit.

### 2. (Major) The example launch line does not parse

```
gst-launch-1.0 videotestsrc ! ... ! cmp.sink_0
    videotestsrc ! ... ! cmp.sink_1
     tensor_compositor name=cmp num_tensors=1 types=uint8 dimensions=3:1920:1080 sink_0::z_order=1 sink_1::z_order=2 sink0::pos=0:640:480 sink1::pos=0:0:0 merge_logic=add ! ...
```

Three problems, and since this text is the entire specification, they will be copied into the implementation:

- **Pad names are inconsistent within one line:** `sink_0::z_order` / `sink_1::z_order` versus `sink0::pos` / `sink1::pos`. The rest of nnstreamer (`tensor_mux`, `tensor_merge`) uses `sink_%u`, so `sink_0::pos` is the correct form. Pick one and use it everywhere.
- **`pos` has three components** (`pos=0:640:480`) while the element is described as compositing 2D-image-like tensors. Is the leading `0` a channel offset, a tensor index, or a z-coordinate? Given `dimensions=3:1920:1080` puts channels first, an offset that mirrors the dimension order is defensible — but it needs to be stated, because "position" in the `compositor` element it is modelled on is 2D (`xpos`/`ypos`).
- **`merge_logic=add` is never explained.** What are the other values, and what happens where two inputs overlap and `z_order` already defines precedence? If `z_order` decides the winner, `merge_logic` is redundant; if `merge_logic` decides, `z_order` is. This is the single most important open question in the design and the sketch does not answer it.

Also, line 3 of the example is indented with a stray extra space relative to line 2, and there is no `\` continuation — as written it is not copy-pasteable.

### 3. (Minor) Malformed doc block

`<refsect2>` is closed with `</refsects>`. It will not match, and any gtk-doc-style processing of the block will misbehave.

### 4. (Minor) Design questions the sketch should answer before implementation starts

Since the point of the PR is to collect ideas, these are the ones I would want settled in the text:

- **Background / uncovered regions.** With `z_order` and explicit positions, output areas covered by no input are possible. Zero-filled? A `background` property? For a quantized uint8 tensor, "zero" is not neutral.
- **`num_tensors` must match across inputs** is stated; **what happens when it does not** is not. Caps negotiation failure, presumably — worth saying so, since that is the behavior tests will assert.
- **Synchronization.** `tensor_mux`/`tensor_merge` expose `sync-mode`/`sync-option` and the docs for `tensor_mux` explicitly discuss when to create output tensors. A compositor with N sink pads has exactly the same problem and the sketch is silent on it. Reusing the existing `sync-mode`/`sync-option` vocabulary would be the consistent choice.
- **Type promotion.** "the same types for every matching tensor" forbids mixing uint8 and float32 inputs. That is a reasonable v1 constraint, but the stated use case ("put cropped tensors into a normalized dimension") often involves exactly that mix, so it is worth confirming the constraint is intentional.
- **Relationship to `tensor_crop`.** The PR body mentions the cropped-tensor use case; if `tensor_crop` output is the expected upstream, the dimension/format contract between the two should be part of the design.

### Other checks

- **Regression risk:** none. Nothing is built and no existing module is touched.
- **Tests:** none, and none are needed at this stage — but note that if this file lands as-is, there is no CI signal of any kind attached to it, so it cannot rot loudly. That is another argument for item 1.
- **API/ARCHITECTURE docs:** this PR *is* the doc; once an implementation lands, `Documentation/component-description.md` will need a `tensor_compositor` entry alongside `tensor_mux`/`tensor_merge`/`tensor_split`.
- **Security:** none. No back-door-shaped code found (there is no code).

Given the age of this PR (open since 2023) and that it carries no code, consider whether #3017 is the better home for the design text and this PR can be closed once the content is moved there.

myungjoo:
Abandoning this. Anyone can resume it.

PR #4300, opened 2023-10-30
[action] Fix meson install and usage

anyj0527:
- Let ubuntu build install proper meson and python.
- And use meson commands explicilty rather than use outdated action from
  github marketplace

**Self evaluation:**


anyj0527:
This will fix recent CI breaks upon gh action meson build.

jaeyun-jung (approved):

again4you (approved):
Good Job!

PR #4301, opened 2023-10-30
[Common] macro for max number of tensors

jaeyun-jung:
Change definition for the max number of tensors (256), and update related code.


myungjoo (approved):

anyj0527 (approved):
LGTM👍

PR #4305, opened 2023-11-01
[Python] validate tensors info

jaeyun-jung:
Validate output tensors info after parsing pyobject.

### DIFF
--- a/ext/nnstreamer/extra/nnstreamer_python3_helper.cc
+++ b/ext/nnstreamer/extra/nnstreamer_python3_helper.cc
@@ -23,6 +23,10 @@
 #include <nnstreamer_util.h>
 #include "nnstreamer_python3_helper.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /** @brief object structure for custom Python type: TensorShape */
 typedef struct {
   PyObject_HEAD PyObject *dims;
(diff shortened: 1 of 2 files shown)


ext/nnstreamer/extra/nnstreamer_python3_helper.cc
```diff
@@ -480,7 +480,8 @@ parseTensorsInfo (PyObject *result, GstTensorsInfo *info)
     Py_SAFEDECREF (shape_dims);
   }
 
-  return 0;
+  /* Validate output tensors info after parsing python object. */
+  return gst_tensors_info_validate (info) ? 0 : -EINVAL;
```
myungjoo, line 484: An error message explaining this EINVAL would be nice (ml_loge with the contents of info so that developers can debug their pipelines or python code).

jaeyun-jung: Updated, please review the changes again.

ext/nnstreamer/extra/nnstreamer_python3_helper.cc
```diff
@@ -480,6 +484,12 @@ parseTensorsInfo (PyObject *result, GstTensorsInfo *info)
     Py_SAFEDECREF (shape_dims);
   }
 
+  /* Validate output tensors info after parsing python object. */
+  if (!gst_tensors_info_validate (info)) {
+    Py_ERRMSG ("Failed to parse the tensors information from python script, it may include invalid tensor type or dimension value.");
```
myungjoo, line 489: Please print out gst_tensors_info_to_string(info) in the error message. If there is an error in info, it will explain why it is invalid in detail.

jaeyun-jung: updated

gichan-jang (approved):
LGTM

myungjoo (approved):

PR #4307, opened 2023-11-01
[Common] check index to get nth tensor

jaeyun-jung:
Code clean, check index to get nth tensor from gst-buffer and remove unnecessary code.

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -1547,7 +1547,7 @@ GstMemory *
 gst_tensor_buffer_get_nth_memory (GstBuffer * buffer, const guint index)
 {
   guint i, num_tensors, offset = 0;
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -1557,8 +1557,9 @@ gst_tensor_buffer_get_nth_memory (GstBuffer * buffer, const guint index)
   }
 
   num_tensors = gst_tensor_buffer_get_count (buffer);
-  if (num_tensors == 0U) {
-    nns_loge ("num_tensors is 0. Please check the buffer.");
+  if (index >= num_tensors) {
```
songgot, line 1560: Since ```index``` starts from 0, it seems that an exception is handled if the return value of ``` gst_tensor_buffer_get_count()``` is 0.

songgot (approved):
LGTM

PR #4308, opened 2023-11-02
[Common] unref memory when appending tensor buffer

jaeyun-jung:
Fix util function to take gst-memory if failed to append it.


myungjoo (approved):

songgot (approved):
LGTM

PR #4309, opened 2023-11-02
[Common] fix mem leak case

jaeyun-jung:
Fix mem leak case if failed to append memory into gst-buffer.

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -346,7 +346,7 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
   GstTensorsConfig in_configs;
   GstClockTime base_time = 0;
   GstTensorInfo *_info;
-  guint i;
+  guint i, j;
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -510,6 +504,9 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
     }
 
     if (!gst_tensor_buffer_append_memory (tensors_buf, mem, _info)) {
+      for (j = i + 1; j < counting; j++)
+        gst_memory_unref (in_mem[j]);
+
```
songgot, line 509: I do not know well but,
Can this be solved with ```gst_buffer_unref(tensor_buf)``` ?

jaeyun-jung: The memories are not appended yet, so gst_buffer_unref() cannot release these.

songgot: Thank you.

songgot (approved):
LGTM

myungjoo (approved):

PR #4310, opened 2023-11-02
[action] Do not run github workflows on every push event

anyj0527:
- Doing every workflows on each push is not necessary as each
  pull request already checks it.
- Do not run those to save the earth!

[action] Cache gbs once every weekday
- Let the new workflow cache gbs every weekday.

**Self evaluation:**


myungjoo (approved):

songgot (approved):
LGTM

PR #4312, opened 2023-11-06
[Tizen/SensorSrc] fix mem leak case

jaeyun-jung:
Code clean, fix mem leak case when failed to add new gst-buffer.


myungjoo (approved):

gichan-jang (approved):
LGTM

PR #4320, opened 2023-11-10
[Edge] Remove re-connect from nnstreamer element

gichan-jang:
- Since the reconnection has been changed to be done inside the nns-edge, nns_edge_connect for reconnection is deleted from the nnstreamer element.
 - Add state change function and set `playing` prop.

**Self evaluation:**
1. Build test: [O]Passed [ ]Failed [ ]Skipped
2. Run test: [O]Passed [ ]Failed [ ]Skipped

### DIFF
--- a/gst/edge/edge_src.c
+++ b/gst/edge/edge_src.c
@@ -65,6 +65,8 @@ static void gst_edgesrc_set_dest_port (GstEdgeSrc * self,
 static nns_edge_connect_type_e gst_edgesrc_get_connect_type (GstEdgeSrc * self);
 static void gst_edgesrc_set_connect_type (GstEdgeSrc * self,
     const nns_edge_connect_type_e connect_type);
+static GstStateChangeReturn gst_edgesrc_change_state (GstElement *
+    element, GstStateChange transition);
 
 /**
  * @brief initialize the class
@@ -117,6 +119,7 @@ gst_edgesrc_class_init (GstEdgeSrcClass * klass)
 
   gstbasesrc_class->start = gst_edgesrc_start;
   gstbasesrc_class->create = gst_edgesrc_create;
+  gstelement_class->change_state = gst_edgesrc_change_state;
 
   GST_DEBUG_CATEGORY_INIT (GST_CAT_DEFAULT,
       GST_EDGE_ELEM_NAME_SRC, 0, "Edge src");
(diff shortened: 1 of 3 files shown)


gst/edge/edge_src.c
```diff
@@ -345,20 +378,21 @@ gst_edgesrc_create (GstBaseSrc * basesrc, guint64 offset, guint size,
...
   GstBuffer *buffer = NULL;
   guint i, num_data;
   int ret;
 
   UNUSED (offset);
   UNUSED (size);
 
-  data_h = g_async_queue_pop (self->msg_queue);
+  while (self->playing && !data_h) {
+    data_h = g_async_queue_timeout_pop (self->msg_queue, G_USEC_PER_SEC);
```
songgot, line 390: How about setting the property instead of 'G_USEC_PER_SEC'?

songgot (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -690,8 +670,8 @@ gst_tensor_query_client_chain (GstPad * pad,
   g_free (val);
 
   if (NNS_EDGE_ERROR_NONE != nns_edge_send (self->edge_h, data_h)) {
-    nns_logw ("Failed to publish to server node, retry connection.");
-    goto retry;
+    nns_logi ("Failed to publish to server node, retry connection.");
+    goto done;
```
wooksong, line 674: Is it OK without changing ```res``` and returning it?

gichan-jang: It's Ok. Even if it fails to send the buffer, drop the current buffer and send the next incoming buffer.
And if we return error here, the pipeline stops.

gst/edge/edge_src.c
gichan-jang: Thanks for the suggestion.

jaeyun-jung, line 390: Q: timed-out param
Should we set the timeout to wait for new data? If edge device does not send new data in connected state, then this may raise an error.

gichan-jang: This is to prevent infinite waiting of data.
Even if timeout occurs while connected, it waits for an additional 1 second and no error occurs.

jaeyun-jung (approved):

PR #4327, opened 2023-11-21
[Query] Add max-request prop to query client @open sesame 11/22 10:40

gichan-jang:
Currently, the query client sends all input buffers to the query server.
If the processing speed of the server is slower than the client, the result time of the server is gradually different from the input time.
To prevent this, the server processes only limited number of buffers, and drop the remaining input buffers.

Signed-off-by: gichan2-jang <gichan2.jang@samsung.com>

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_client.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_client.c
@@ -46,6 +46,7 @@ enum
   PROP_TOPIC,
   PROP_TIMEOUT,
   PROP_SILENT,
+  PROP_MAX_REQUEST,
 };
 
 #define TCP_HIGHEST_PORT        65535
@@ -54,6 +55,7 @@ enum
 #define TCP_DEFAULT_CLIENT_SRC_PORT 3001
 #define DEFAULT_CLIENT_TIMEOUT  0
 #define DEFAULT_SILENT TRUE
+#define DEFAULT_MAX_REQUEST 2
 
 GST_DEBUG_CATEGORY_STATIC (gst_tensor_query_client_debug);
 #define GST_CAT_DEFAULT gst_tensor_query_client_debug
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -192,6 +198,8 @@ gst_tensor_query_client_init (GstTensorQueryClient * self)
   self->timeout = DEFAULT_CLIENT_TIMEOUT;
   self->edge_h = NULL;
   self->msg_queue = g_async_queue_new ();
+  self->drop = FALSE;
```
wooksong, line 201: How about using the default value that you defined?
```suggestion
  self->drop = DEFAULT_DROP;
```

gichan-jang: Thanks, updated!

wooksong (approved):
LGTM

gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -669,16 +683,23 @@ gst_tensor_query_client_chain (GstPad * pad,
   nns_edge_data_set_info (data_h, "client_id", val);
   g_free (val);
 
-  if (NNS_EDGE_ERROR_NONE != nns_edge_send (self->edge_h, data_h)) {
-    nns_logi ("Failed to publish to server node.");
-    goto done;
+  if (self->drop && self->requested_num >= 2) {
```
jaeyun-jung, line 686: Suggestion: new property guint 'max-request' instead 'drop'
default value 2(?) to allow sending request to server (0 for infinite), so that user can set the max number of query. :)

gichan-jang: Thanks, It would be better for the user. Updated.

gst/nnstreamer/tensor_query/tensor_query_client.c
```diff
@@ -669,16 +685,23 @@ gst_tensor_query_client_chain (GstPad * pad,
   nns_edge_data_set_info (data_h, "client_id", val);
   g_free (val);
 
-  if (NNS_EDGE_ERROR_NONE != nns_edge_send (self->edge_h, data_h)) {
-    nns_logi ("Failed to publish to server node.");
-    goto done;
+  if (self->requested_num > self->max_request) {
```
jaeyun-jung, line 688: condition to drop the input: self->max_request should be positive value.
if (self->max_request > 0 && self->requested_num > self->max_request)

gichan-jang: Right. Updated.

songgot (approved):
LGTM

jaeyun-jung (approved):

PR #4328, opened 2023-11-22
[QueryClient] handle tensor buffer

jaeyun-jung:
1. Decrease the ref count of incoming buffer if an error occurs.
2. Check the number of requested buffer before creating edge-data handle, and try to get the output from query-server.
3. Append gst-buffer using util function if current caps is tensor stream.


gichan-jang (approved):
LGTM, Thanks.

PR #4331, opened 2023-11-22
[Transform] available caps without peer

jaeyun-jung:
tensor-transform cannot determine the caps, this element can set out-caps by using incoming caps. 
Remove unnecessary query for peer caps and update types of out-caps if input has tensors.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -2009,11 +2009,9 @@ gst_tensor_transform_transform_caps (GstBaseTransform * trans,
     gboolean is_types_not_fixed = FALSE;
     GstCaps *result_aux = gst_caps_new_empty ();
 
-    structure = gst_caps_get_structure (caps, i);
-
-    gst_tensors_config_init (&in_config);
     gst_tensors_config_init (&out_config);
 
+    structure = gst_caps_get_structure (caps, i);
     gst_tensors_config_from_structure (&in_config, structure);
 
     if (gst_tensors_config_is_flexible (&in_config)) {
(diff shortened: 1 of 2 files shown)


jaeyun-jung:
@niley7464 please check this PR after clear your current issue :)

anyj0527 (approved):
LGTM, cool!

niley7464 (approved):
LGTM!!
Thank you :)

gst/nnstreamer/elements/gsttensor_transform.c
```diff
@@ -2041,13 +2041,10 @@ gst_tensor_transform_transform_caps (GstBaseTransform * trans,
...
-    }
 
-    /* remove `types` field from caps */
-    if (is_types_not_fixed) {
-      GstStructure *s;
-      for (j = 0; j < gst_caps_get_size (result_aux); ++j) {
-        s = gst_caps_get_structure (result_aux, j);
+      /* remove `types` field from caps */
+      if (is_types_not_fixed) {
+        GstStructure *s = gst_caps_get_structure (result_aux, 0);
```
myungjoo, line 2047: How do you sure that result_aux will have only one GstStructure?

jaeyun-jung: The caps 'result_aux' is created in the loop (line 2010), and new capability is appended once (2040~2043).

myungjoo (commented):
If you can add a test case (a gst pipeline desc) that runs with this commit (a pipeline that can be negotiated only after this commit), it'd be fantastic. It will also provide a regression preventor.

myungjoo (approved):

PR #4332, opened 2023-11-22
[Test] tensorflow-lite version in testcase

jaeyun-jung:
Fix version of tensorflow-lite.
Both version 2.x and 1.x can be used in testcase if tensor-filter property framework is tensorflow-lite. Default version is v2.x, and it has higher priority.

Also, added minor patch to fix indent.


anyj0527 (approved):
LGTM

songgot (approved):
LGTM

myungjoo (approved):

Issue #4235, opened 2023-09-04
result string from ml-agent


jaeyun-jung:
TBU, string from ml-agent interfaces (get model or resource)

```
model_get_all:
[
  {"version":"1","active":"F","path":"/opt/usr/shared/my_model_file_1.pb","description":"","app_info":""}, 
  {"version":"2","active":"F","path":"/opt/usr/shared/my_model_file_2.pb","description":"","app_info":""}, 
  {"version":"3","active":"T","path":"/opt/usr/shared/my_model_file_3.pb","description":"","app_info":""}
]

model_get:
{"version":"2","active":"F","path":"/opt/usr/shared/my_model_file_2.pb","description":"","app_info":""}

model_get_activated:
{"version":"3","active":"T","path":"/opt/usr/shared/my_model_file_3.pb","description":"","app_info":""}

resource_get:
[
  {"path":"/opt/usr/shared/my_resource_file1.dat","description":"","app_info":""},
  {"path":"/opt/usr/shared/my_resource_file2.dat","description":"","app_info":""},
  {"path":"/opt/usr/shared/my_resource_file3.dat","description":"","app_info":""}
]
```

myungjoo:
Closing: the ml-agent result string is now consumed by real code, so the format recorded here is no longer a TBU note.

`gst/nnstreamer/ml_agent.c` parses the stringified JSON returned by the ML Agent and resolves an `mlagent://` URI to a model path; the caller is `mlagent_get_model_path_from()` at `gst/nnstreamer/tensor_filter/tensor_filter_common.c:1378`. The parsing landed in #4336, which referenced this issue.

Coverage exists as well: `tests/unittest_mlagent/` (with `mock_mlagent.cc` standing in for the agent), so the accepted shape of `model_get` / `model_get_activated` is pinned by tests rather than by this issue body.

If the intent was to publish the schema as documentation rather than to have it implemented, please reopen with that scope — this close only covers the implementation side.

PR #4336, opened 2023-11-28
[Filter/MLAgent: Implement the JSON parsing logic to get model's path

wooksong:
This patch adds the logic to parse the C-style stringified JSON data gathered from the ML Agent in order to convert the given URI to the model's path.

See also: #4235

**Self evaluation:**


myungjoo (commented):
Please add a simple "how-to" document in .h/.c or .md file.

gst/nnstreamer/ml_agent.h
```diff
@@ -21,9 +21,9 @@
 /**
  * @brief Parse the given URI into the valid file path string
  */
-const gchar *mlagent_parse_uri_string (const GValue * val);
+gchar *mlagent_parse_uri_string (const GValue * val);
```
myungjoo, line 24: Add a note for the returned pointer; whether the caller should free or not.

gst/nnstreamer/ml_agent.c
```diff
@@ -78,29 +82,74 @@ mlagent_parse_uri_string (const GValue * val)
...
        * @todo The specification of the data layout filled in the third
        *       argument (i.e., stringfied_json) by the callee is not fully decided.
        */
-      rcode = ml_agent_model_get (name, version, stringfied_json, &err);
+      rcode = ml_agent_model_get (name, version, &stringfied_json, &err);
+      if (rcode != 0) {
+        nns_loge
+            ("Failed to get the stringied JSON using the given URI(%s): %s",
+            uri, (err ? err->message : "unknown reason"));
+        goto fallback;
```
myungjoo, line 98: Are you going to return the URI string even when there is an error?

gst/nnstreamer/ml_agent.h
wooksong: Updated it. Thanks.

gst/nnstreamer/ml_agent.c
wooksong: Yes. If the given string is not for ml-agent, ```_gtfc_setprop_MODEL``` is in charge of it when it did before.

myungjoo: Then, how the caller will be notified for such errors? (you need to specify it in the doxygen, too)

gst/nnstreamer/ml_agent.c
```diff
@@ -12,34 +12,42 @@
...
 const gchar URI_SCHEME[] = "mlagent";
 const gchar URI_KEYWORD_MODEL[] = "model";
+const gchar JSON_KEY_MODEL_PATH[] = "path";
 
 /**
  * @brief Parse the given URI into the valid file path string
+ * @param[in] val A pointer to a GValue holding a G_TYPE_STRING value
+ * @return A newly allocated c-string containing the model file path in the case that the valid URI is given.
+ * Otherwise, it simply returns c-string that the val contains.
+ * @note The caller should free the return c-string after using it.
```
myungjoo, line 30: Specify how errors will be handled? (NULL return? ERRPTR? some special strings? ...)

gst/nnstreamer/ml_agent.c
wooksong: This function converts a given URI to a model path c-string when ml-agent is enabled and the given URI is valid. When the ml-agent is disabled, it works in the same way as ```g_value_dup_string``` works (please see ```ml_agent.h```). As ```g_value_dup_string``` does, it returns ```NULL``` when the ```GValue *val``` does not hold a string value. If the string has been verified as a valid URI, then this function converts it to a path c-string. Otherwise, it just returns the string that the ```*val``` holds.

gst/nnstreamer/ml_agent.c
```diff
@@ -78,29 +86,74 @@ mlagent_parse_uri_string (const GValue * val)
...
+        }
+
+        path = json_object_get_string_member (jobj, JSON_KEY_MODEL_PATH);
+        if (path == NULL || !g_strcmp0 (path, "")) {
+          nns_loge
+              ("Failed to get the model's path from the given URI: "
+              "Invalid value for the key, %s", JSON_KEY_MODEL_PATH);
+          goto fallback;
+        }
+        return g_strdup (path);
```
jaeyun-jung, line 149: free uri before return

jaeyun-jung (approved):

gst/nnstreamer/ml_agent.c
```diff
@@ -12,34 +12,42 @@
...
 const gchar URI_SCHEME[] = "mlagent";
 const gchar URI_KEYWORD_MODEL[] = "model";
+const gchar JSON_KEY_MODEL_PATH[] = "path";
 
 /**
- * @brief Parse the given URI into the valid file path string
+ * @brief Get a path of the model file from a given GValue
+ * @param[in] val A pointer to a GValue holding a G_TYPE_STRING value
+ * @return A newly allocated c-string representing the model file path, if the given GValue contains a valid URI
+ * Otherwise, it simply returns c-string that the val contains.
```
myungjoo, line 29: to remove ambiguity,
```Otherwise, it simply returns a duplicated (strdup'ed) c-string that the val contains.```

myungjoo (approved):
Please update doxygen comment at LINE 29. Then, please merge.

PR #4337, opened 2023-11-30
[tensor_trainer] Modify description of model-load-path

songgot:
Change to 'pre-trained' instead of 'existing' expression

**Self evaluation:**


gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -196,7 +196,7 @@ gst_tensor_trainer_class_init (GstTensorTrainerClass * klass)
 
   g_object_class_install_property (gobject_class, PROP_MODEL_LOAD_PATH,
       g_param_spec_string ("model-load-path", "Model load path",
-          "(nullable) Path to load an existing model to use for training a new model",
+          "(nullable) Path to load a pre-trained model to use for training a new model",
```
myungjoo, line 199: It can be both a pre-trained or a non-trained initial model. It can also be an already personalized model that is to be further personalized, too. You do not know the purpose of users at build-time.

```
"(nullable) Path to a model file to be loaded for the given training session."
```

songgot: Thank you. I will fix this.

myungjoo (approved):

PR #4338, opened 2023-12-01
[Doc] remove unnecessary lines from ml-api

jaeyun-jung:
Code clean, now we do not define data structure in ML API repo and use common functions of nnstreamer. 
Remove unnecessary lines defined in api repo.


anyj0527 (approved):
LGTM

myungjoo (approved):

PR #4340, opened 2023-12-04
[Common] check tensor dim to append previous caps

jaeyun-jung:
Prevent appending same caps if tensor dimension has min rank 4. 
(e.g., given dim is 1001:1, previous caps with rank 4 is unnecessary.)


niley7464 (approved):
LGTM!

anyj0527 (approved):
LGTM✔

PR #4341, opened 2023-12-04
[Filter] condition to update dimension

jaeyun-jung:
For caps negotiation, compare tensor dimension if caps is tensor stream and already configured.


niley7464 (approved):
LGTM!
Thank you 👍

anyj0527 (approved):
LGTM

PR #4344, opened 2023-12-08
[Filter/TF-lite] fill 1 when changing input dimension

jaeyun-jung:
After implementing 0-init dims, tf-lite subplugin fills zero when changing input dimension.
Fill 1 if given dimension is 0 dim value.


anyj0527 (approved):
LGTM👍

myungjoo (approved):

PR #3872, opened 2022-08-11
decoder/converter::python make them thread safe. @open sesame 8/18 10:00

myungjoo:
Use GIL instead of mutex. Refer to #3832.

Fixes #3834

Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

Issue #3834, opened 2022-07-07
tensor_decoder / tensor_converter :: python subplugin threading issue


myungjoo:
Fix the issue by using GIL instead of mutex.
Refer to #3832

myungjoo:
Closing: implemented. Python subplugins now use GIL-based thread safety via the PyGILGuard RAII class (ext/nnstreamer/extra/nnstreamer_python3_helper.h), applied across tensor_filter_python3, tensordec-python3, and tensor_converter_python3 through PRs #4348, #4350, #4375, and #4788.

PR #4347, opened 2023-12-12
converter::python make them thread safe with GIL.

myungjoo:
Apply GIL control to converter::python3 as in we do with filter::python3.

Fixes converter part of #3834

Overrides #3872

### DIFF
--- a/ext/nnstreamer/tensor_converter/tensor_converter_python3.cc
+++ b/ext/nnstreamer/tensor_converter/tensor_converter_python3.cc
@@ -222,8 +222,9 @@ PYConverterCore::init ()
   shape_cls = PyObject_GetAttrString (api_module, "TensorShape");
   Py_SAFEDECREF (api_module);
 
-  if (shape_cls == NULL)
+  if (shape_cls == NULL) {
     return -EINVAL;
+  }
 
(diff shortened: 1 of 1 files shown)


jaeyun-jung (approved):

ext/nnstreamer/tensor_converter/tensor_converter_python3.cc
```diff
@@ -212,19 +212,25 @@ PYConverterCore::convert (GstBuffer *in_buf, GstTensorsConfig *config)
 int
 PYConverterCore::init ()
 {
+  int ret = 0;
   /** Find nnstreamer_api module */
   PyObject *api_module = PyImport_ImportModule ("nnstreamer_python");
   if (api_module == NULL) {
-    return -EINVAL;
+    ret = -EINVAL;
+    goto done;
```
SeoHyungjun, line 220: Wouldn't "early return", code before using "goto" in general, be desirable?
It seems to be the same function. Is there a reason for the change?

myungjoo: Thanks. That was a residue from older changes.

myungjoo:
This requires fixes in #4350. Closing this for the duplication.

PR #4348, opened 2023-12-12
decoder::python make them thread safe with GIL.

myungjoo:
Apply GIL control to decoder::python3 as in we do with filter::python3.

Fixes decoder part of #3834

Overrides #3872


anyj0527 (approved):
LGTM👍

jaeyun-jung (approved):

PR #4349, opened 2023-12-12
filter::python3: do not break converter/decoder

myungjoo:
A python-global reference counter is required for py-finalize.
This is workaround before applying such a global reference counter.

A real fix will be applied after #4348 and #4347


songgot (approved):
LGTM

anyj0527 (approved):
LGTM✔

PR #4352, opened 2023-12-18
[Filter/ncnn] Add a sub-plugin filter for ncnn

chosanglyul:
This PR is to newly add [ncnn 20230816](https://github.com/Tencent/ncnn/tree/20230816) subplugin support for tensor_filter.

**Building ncnn:**

Without Vulkan
```bash
sudo apt install build-essential git cmake libprotobuf-dev protobuf-compiler libomp-dev libopencv-dev
cmake -GNinja -Bbuild -S. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DNCNN_SHARED_LIB=ON
sudo ninja -Cbuild install
```

With Vulkan (recommended)
```bash
sudo apt install build-essential git cmake libprotobuf-dev protobuf-compiler libomp-dev libvulkan-dev vulkan-tools libopencv-dev
cmake -GNinja -Bbuild -S. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DNCNN_SHARED_LIB=ON -DNCNN_VULKAN=ON
sudo ninja -Cbuild install
```

Usage Examples:
We added examples about image classification and object detection
See also: [PR](https://github.com/nnstreamer/nnstreamer-example/pull/320)

Special considerations on properties:
- input, inputtype, output, outputtype: It is essential to set these four options correctly. For assistance in configuring the shape and type, please refer to the examples provided.
- accelerator: Vulkan acceleration by setting `accelerator=true:gpu`. This option is applicable if your device is equipped with any Vulkan-acceleratable processor.
- custom: Each entries are separated by ','. Each entries have `property_key:value` format. There must be no spaces.
- use_yolo_decoder (optional, default=false): Enable this option by setting `use_yolo_decoder:true` if your model includes a `Yolov3DetectionOutput` layer or yolo-related output layers, especially when dealing with variable output sizes `(num_detection, 6)`. In such cases, you must also configure `output=(5+num_labels, max_detection, 1)` and `outputtype=float32`. To calculate the max_detection for an input image of size `(w, h)`, use the formula: `(w/32)*(h/32) + (w/16)*(h/16) + (w/8)*(h/8)*3`. See also: https://github.com/nnstreamer/nnstreamer/blob/main/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.c#L1194

**Self evaluation:**

**How to evaluate:**
```bash
cd nnstreamer
ninja -Cbuild install
cd tests
ssat
```

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
@@ -28,7 +28,7 @@
  *
  *  custom:
  *    Each entries are separated by ','
- *    Each entries have property_key=value format.
+ *    Each entries have property_key:value format.
  *    There must be no spaces.
  *
  *    Supported custom properties:
@@ -213,7 +213,7 @@ ncnn_subplugin::configure_instance (const GstTensorFilterProperties *prop)
   // init input matrices
   for (guint i = 0; i < inputInfo.num_tensors; i++) {
     // get dimensions of the input matrix from inputInfo
-    const uint32_t *dim = inputInfo.info[i].dimension;
+    const uint32_t *dim = gst_tensors_info_get_nth_info (&inputInfo, i)->dimension;
     std::vector<int> shape;
     while (*dim)
       shape.push_back (*dim++);
@@ -234,7 +234,8 @@ ncnn_subplugin::configure_instance (const GstTensorFilterProperties *prop)
         in = ncnn::Mat (shape[0], shape[1], shape[2], shape[3]);
         break;
       default:
-        throw std::invalid_argument ("Wrong input dimension");
+        throw std::invalid_argument ("ncnn subplugin supports only up to 4 ranks and does not support input tensors of "
+                                     + std::to_string (shape.size ()) + " dimensions.");
     }
     input_mats.push_back (in);
   }
@@ -260,7 +261,12 @@ ncnn_subplugin::configure_instance (const GstTensorFilterProperties *prop)
 void
 ncnn_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
 {
-  assert (!empty_model);
+  if (empty_model)
+    throw std::runtime_error (
+        "Model is empty: the ncnn instance is not configured and "
+        "its \"invoke\" method is called. This may be an internal bug of "
+        "nnstreamer or ncnn-subplugin unless if you have directly accessed "
+        "ncnn-subplugin.");
 
   // make extractor instance for each inference
   ncnn::Extractor ex = net.create_extractor ();
@@ -305,7 +311,8 @@ ncnn_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
     // write detection-box infos to the output tensor
     for (guint i = 0; i < outputInfo.num_tensors; i++) {
       ncnn::Mat &out = output_mats.at (i);
-      const int label_count = outputInfo.info[i].dimension[0];
+      const int label_count
+          = gst_tensors_info_get_nth_info (&outputInfo, i)->dimension[0];
       float *output_data = (float *) output->data;
       for (int j = 0; j < out.h; j++) {
         float *values = out.row (j);
@@ -398,7 +405,7 @@ ncnn_subplugin::parseCustomProperties (const GstTensorFilterProperties *prop)
 
     for (guint i = 0; i < len; i++) {
       // split with = to parse single option
-      uniq_g_strv option (g_strsplit (options.get ()[i], "=", -1), g_strfreev);
+      uniq_g_strv option (g_strsplit (options.get ()[i], ":", -1), g_strfreev);
 
       // we only have key=value form option
       if (g_strv_length (option.get ()) == 2) {
--- a/ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
@@ -1,15 +1,24 @@
 /* SPDX-License-Identifier: LGPL-2.1-only */
 /**
  * @file    tensor_filter_ncnn.cc
- * @date    13 Sep 2022
+ * @date    15 Dec 2022
  * @brief   NNStreamer tensor-filter sub-plugin for Tencent NCNN
  * @author  Sungbin Jo <goranmoomin@daum.net>
+ * @author  SangLyul Cho <chosanglyul@gmail.com>
  * @see     http://github.com/nnstreamer/nnstreamer
- * @bug     Nothing works.
+ * @bug     No known bugs.
  *
  * This is the NCNN plugin for tensor_filter.
+ *
+ * @details Usage examples
+ *  TODO
+ *
+ * @note Special considerations on props:
+ *  TODO
+ *  TODO make description of yolo filter output size
  */
 
+#include <functional>
 #include <glib.h>
 #include <nnstreamer_cppplugin_api_filter.hh>
 #include <nnstreamer_log.h>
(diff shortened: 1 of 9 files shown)


tests/nnstreamer_filter_ncnn/runTest.sh
```diff
@@ -0,0 +1,64 @@
...
+## @brief SSAT Test Cases for NNStreamer
+##
+if [[ "$SSATAPILOADED" != "1" ]]; then
+    SILENT=0
+    INDEPENDENT=1
+    search="ssat-api.sh"
+    source $search
+    printf "${Blue}Independent Mode${NC}"
+fi
+
```
anyj0527, line 17: Add dependency library check here. Like in https://github.com/nnstreamer/nnstreamer/blob/main/tests/nnstreamer_filter_onnxruntime/runTest.sh#L25

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
```diff
@@ -0,0 +1,504 @@
...
+        break;
+      case 3:
+        in = ncnn::Mat (shape[0], shape[1], shape[2]);
+        break;
+      case 4:
+        in = ncnn::Mat (shape[0], shape[1], shape[2], shape[3]);
+        break;
+      default:
+        throw std::invalid_argument ("Wrong input dimension");
+    }
```
myungjoo, line 242: @niley7464 : is it ok to support up to 4 ranks in a subplugin?

@chosanglyul : because nnstreamer latest supports up to 16 ranks, the error message might need to say "ncnn subplugin supports only up to 4 ranks and does not support input tensors of " + std::string (shape.size()) + "dimensions."

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
```diff
@@ -0,0 +1,504 @@
...
+  empty_model = false;
+}
+
+/**
+ * @brief Invoke ncnn model and get the inference result.
+ */
+void
+ncnn_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
+{
+  assert (!empty_model);
```
myungjoo, line 267: Do not use assert in a case where it can be caused by a user (a developer using this "subplugin").
Use assert only for the case where it must be caused by this code.

Apply exception throwing/handling or error returning.

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
niley7464: https://github.com/nnstreamer/nnstreamer/blob/35642d7b1f060ce6304c0129813f88263715150f/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c#L469

nnfw subplugin uses it's own `NNFW_MAX_RANK` (6) so I think it's okay to support up to 4 rank.

I just realized that nnfw subplugin does not consider `GstTensorInfo *info` rank > `NNFW_MAX_RANK`.
Maybe this kind of exception handling has to be added in nnfw subplugin too.

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
goranmoomin: While the code itself is updated, note that this pattern exists verbatim at `tensor_filter_mxnet.cc`, `tensor_filter_tvm.cc`, and so on. AFAIU there are no codepaths that trigger this assertion since we are setting the property to true at the end of `ncnn_subplugin::configure_instance`.

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
```diff
@@ -0,0 +1,500 @@
...
+ *    please refer to the examples provided.
+ *
+ *  accelerator:
+ *    Enable Vulkan acceleration by setting accelerator=true:gpu.
+ *    This option is applicable if your device is equipped
+ *    with any Vulkan-acceleratable processor.
+ *
+ *  custom:
+ *    Each entries are separated by ','
+ *    Each entries have property_key=value format.
```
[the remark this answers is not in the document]
goranmoomin: Updated.

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
```diff
@@ -0,0 +1,500 @@
...
+  if (inputInfo.num_tensors != input_indexes.size ())
+    throw std::invalid_argument (
+        std::string ("Wrong number of input matrices")
+        + ": Found in argument = " + std::to_string (inputInfo.num_tensors)
+        + ", Found in model file = " + std::to_string (input_indexes.size ()));
+
+  // init input matrices
+  for (guint i = 0; i < inputInfo.num_tensors; i++) {
+    // get dimensions of the input matrix from inputInfo
+    const uint32_t *dim = inputInfo.info[i].dimension;
```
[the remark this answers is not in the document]
goranmoomin: I'm not sure what code path would trigger this – I'd expect that we might have to handle the case where the result of `gst_tensors_info_get_nth_info` is `nullptr`? What would be the appropriate action in that case?

For now, the code just assumes that it is not NULL (as before, just using the functions).

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
goranmoomin: Updated to the suggested error message.

anyj0527 (commented):
Please rebase on the latest main branch

goranmoomin:
> Please rebase on the latest main branch

Done.

anyj0527 (approved):
LGTM👍Great thanks for your contribution!

myungjoo:
> LGTM👍Great thanks for your contribution!

Have you tried it w/ ncnn in your computer? (CI is not covering this yet.)

ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc
```diff
@@ -0,0 +1,503 @@
...
+}
+
+/**
+ * @brief Invoke ncnn model and get the inference result.
+ */
+void
+ncnn_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
+{
+  if (empty_model)
+    throw std::runtime_error ("Model is empty.");
```
myungjoo, line 265: Error message should be useful for those who may read it.
```
Model is empty: the ncnn instance is not configured and its "invoke" method is called. This may be an internal bug of nnstreamer or ncnn-subplugin unless if you have directly accessed ncnn-subplugin.
```

goranmoomin: Done.

chosanglyul:
> > LGTM👍Great thanks for your contribution!
> 
> Have you tried it w/ ncnn in your computer? (CI is not covering this yet.)

Screenshot from 2023-12-19 18-11-55

Yes. We checked that the test worked well in the local Ubuntu 22.04 environment.

meson.build
```diff
@@ -341,6 +341,17 @@ if not get_option('mxnet-support').disabled()
   endif
 endif
 
+# ncnn
```
jaeyun-jung, line 344: Suggestion: pkgconfig 'ncnn.pc'
IMO you can use pkg-config, instead checking the ncnn library (or header) in this block.
```
  'ncnn-support': {
    'target': 'ncnn',
    'project_args': { 'ENABLE_NCNN' : 1 }
  },```

anyj0527:
> > > LGTM👍Great thanks for your contribution!
> > 
> > 
> > Have you tried it w/ ncnn in your computer? (CI is not covering this yet.)
> 
> Screenshot from 2023-12-19 18-11-55
> 
> Yes. We checked that the test worked well in the local Ubuntu 22.04 environment.

I also confirmed the test passed in my local with ncnn/20230816 and also with ncnn/master

jaeyun-jung (approved):

meson.build
goranmoomin: AFAIU the code is already finding the ncnn library with pkg-config (via meson's `cxx.find_library`)? The header checking code is a fallback for when pkg-config doesn't exist, which is copied verbatim from the logic of finding other dependencies (i.e. see MXnet or ARMNN).

jaeyun-jung: OK, we will check this after ncnn packaging is done later.

myungjoo (approved):

myungjoo:
Let's package ncnn for at least one of Ubuntu/Tizen and let it tested in github-action or TAOS-CI later when someone's bored.

PR #4369, opened 2024-01-02
[Pkg] header file for devel package

jaeyun-jung:
1. move util for trainer (single-devel > devel)
2. update library (link to nns-single, internal devel package)


SeoHyungjun (approved):
LGTM!

myungjoo (approved):

PR #4371, opened 2024-01-05
[mlagent] Fix fallback case and unittest_mlagent @open sesame 1/9 16:24

anyj0527:
- Let `mlagent_get_model_path_from` returns uri in fallback case
- Fix memleak (g_strfreev).
- Fix memleak in a testcase.
- Add a missing header <utility> in mock_mlagent.h
  - It declares `std::exchange`
- Fix multiline comments in unittest_mlagent.cc to satisfy CI (doxygen)

**Self evaluation:**

### DIFF
--- a/tests/unittest_mlagent/mock_mlagent.h
+++ b/tests/unittest_mlagent/mock_mlagent.h
@@ -15,6 +15,7 @@
 
 #include <unordered_map>
 #include <vector>
+#include <utility>
 
 struct MockModel;
 
--- a/tests/unittest_mlagent/unittest_mlagent.cc
+++ b/tests/unittest_mlagent/unittest_mlagent.cc
@@ -94,7 +94,8 @@ TEST (testMLAgent, GetModelInvalidURIFormats_n)
 
     path = mlagent_get_model_path_from (&val);
 
-    /* In the case that invalid URIs are given, mlagent_get_model_path_from () returns
+    /**
+     * In the case that invalid URIs are given, mlagent_get_model_path_from () returns
      * the given URI as it is so that it is handled by the fallback procedure (i.e., regarding it as a file path).
      */
     EXPECT_STREQ (uri, path);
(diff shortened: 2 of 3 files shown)


jaeyun-jung (approved):

tests/unittest_mlagent/unittest_mlagent.cc
```diff
@@ -116,14 +117,15 @@ TEST (testMLAgent, GetModelInvalidModel_n)
   g_value_init (&val, G_TYPE_STRING);
 
   g_autofree gchar *uri
-      = g_strdup_printf (valid_uri_format_literal, "InvalidModelName", UINT32_MAX);
+      = g_strdup_printf (invalid_uri_format_literal, "InvalidModelName", UINT32_MAX);
```
wooksong, line 120: This is the case for the valid URI format with an invalid model name (which is an invalid URI).

tests/unittest_mlagent/mock_mlagent.h
```diff
@@ -15,6 +15,7 @@
 
 #include <unordered_map>
 #include <vector>
+#include <utility>
```
wooksong, line 18: Is it necessary?

wooksong: https://stackoverflow.com/questions/69449781/utility-not-necessarily-included-when-swap-is-performed-how-can-this-becom

OK. I think some standard functions like ```std::swap()```, ```std::begin()```, ```std::end()```, ```std::exchange()``` (, which looks like they are already meltdown to the grammar itself), could be used without explicit include statements.

But it was wrong. This line is necessary.

tests/unittest_mlagent/unittest_mlagent.cc
anyj0527: I see. I updated the code ml_agent.c to handle fallback case properly, PTAL

wooksong (approved):
LGTM! Thanks.

myungjoo (approved):

PR #4374, opened 2024-01-08
MLAgent: Unref JsonParser variable using g_object_unref

linuxias:
This patch transforms the JsonParser variable generated by the function
 json_parser_new() by using g_object_unref() to release memory.

 The g_autoptr variable is released at the end of the scope by g_autofree(),
 but should not be released if there is still a reference.

 It should be released only when there is no reference through
 the management using the reference count.

 reference url:
  https://www.manpagez.com/html/json-glib/json-glib-1.2.8/JsonParser.php#json-parser-new

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/ml_agent.c
+++ b/gst/nnstreamer/ml_agent.c
@@ -87,7 +87,7 @@ mlagent_get_model_path_from (const GValue * val)
       g_autofree gchar *name = g_strdup (parts[MODEL_PART_IDX_NAME]);
       guint version = strtoul (parts[MODEL_PART_IDX_VERSION], NULL, 10);
       g_autofree gchar *stringfied_json = NULL;
-      g_autoptr (JsonParser) json_parser = NULL;
+      JsonParser *json_parser = NULL;
       gint rcode;
 
       /**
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/ml_agent.c
```diff
@@ -87,7 +87,7 @@ mlagent_get_model_path_from (const GValue * val)
       g_autofree gchar *name = g_strdup (parts[MODEL_PART_IDX_NAME]);
       guint version = strtoul (parts[MODEL_PART_IDX_VERSION], NULL, 10);
       g_autofree gchar *stringfied_json = NULL;
-      g_autoptr (JsonParser) json_parser = NULL;
+      JsonParser *json_parser = NULL;
```
jaeyun-jung, line 90: Actually unnecessary change.
JsonParser defines auto release function by using g_object_unref (see below:)
https://github.com/GNOME/json-glib/blob/8410c3283ce3d79e6e54612b1b5761b05eedd9de/json-glib/json-parser.h#L194

linuxias: Thanks for your comment.
I checked G_DEFINE_AUTOPTR_CLEANUP_FUNC about JsonParser.

I'll close this PR.

PR #4377, opened 2024-01-09
[fix/onnxruntime-filter] fixed tensor with negative value in shape

yeonykim2:
Some onnx models had free dimensions for batch. ex) [None:3:720:720]
\- It fixed that creating tensor with negative value in shape ```Ort::Value``` is treated  as 1.


SeoHyungjun (approved):
LGTM

jaeyun-jung (approved):

niley7464 (approved):
LGTM!

myungjoo (approved):

PR #4379, opened 2024-01-11
Dist/Tizen: Enable ncnn tensor_filter sub-plugin @open sesame 03/06 15:00

wooksong:
As ncnn packages are released on the Tizen public repository,
this patch enables the ncnn sub-plugin for the tensor filter in Tizen.

**Self evaluation:**


wooksong:
Releasing of NCNN is currently WiP.

myungjoo (approved):

songgot (approved):
LGTM

wooksong:
https://review.tizen.org/gerrit/gitweb?p=platform/upstream/ncnn.git;a=shortlog;h=refs/heads/sandbox/wooksong

NCNN is required to be published on Tizen.

packaging/nnstreamer.spec
```diff
@@ -302,6 +305,10 @@ BuildRequires:	npu-engine-devel
 BuildRequires: onnxruntime-devel
 %endif
 
+%if 0%{?ncnn_support}
+BuildRequires: ncnn-devel
```
myungjoo, line 309: Let's force-enable ncnn build option if ncnn_support is 1. This prevents regression.

wooksong: @myungjoo Updated. PTAL.

myungjoo (approved):

myungjoo:
```
[ 556s] [FAILED] nnstreamer_filter_ncnn (6 passed among 9 cases)
```

jaeyun-jung (approved):

myungjoo (changes requested):
Issue 1: All positive unit TCs failed.
Issue 2: Negative TCs do not have the postfix, ```_n```.

myungjoo:
Issue 1 of TC 1/2 is caused by gstreamer version mismatch with your assumptions.
Older GStreamer (e.g., 1.14) does not have ```num-buffers``` property in the ```imagefreeze``` element.

wooksong:
> Issue 1: All positive unit TCs failed.
> Issue 2: Negative TCs do not have the postfix, `_n`.
> Issue 1 of TC 1/2 is caused by gstreamer version mismatch with your assumptions. Older GStreamer (e.g., 1.14) does not have `num-buffers` property in the `imagefreeze` element.

@myungjoo Well, NCNN has been published, and I almost forgot about this PR. 
Even though I was only in charge of the packaging part (the implementation of the filter and TCs was the work of the SNU students), I will take a look as you commented. Thanks!

wooksong:
> Issue 2: Negative TCs do not have the postfix, `_n`.

This trivial issue has been resolved.

wooksong:
> Issue 1 of TC 1/2 is caused by gstreamer version mismatch with your assumptions. Older GStreamer (e.g., 1.14) does not have `num-buffers` property in the `imagefreeze` element.

The two failed TCs involve several sub-issues. I will summarize them and post them on the [Issues](https://github.com/nnstreamer/nnstreamer/issues). Merging them may take some time.

myungjoo:
You may simple check gstreamer version (or property availability with gst-inspect) at the test script to skip this instance if it's old.

myungjoo:
Be aware that this still breaks CI test.

tests/nnstreamer_filter_ncnn/runTest.sh
```diff
@@ -100,7 +100,7 @@ function run_pipeline() {
 # Property reading test for accelerator
 run_pipeline true:gpu
 cat info | grep "accl = gpu$"
-testResult $? 8 "vulkan accelerator test" 0 1
+testResult $? 2 "vulkan accelerator test" 0 1
```
myungjoo, line 103: Let's check the return value of ```run_pipeline``` as well.

myungjoo: If I check ```$?``` after line 101, it fails.

tests/nnstreamer_filter_ncnn/runTest.sh
```diff
@@ -75,22 +75,22 @@ cat ncnn.out.log | grep "orange"
...
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! videoscale ! video/x-raw,width=227,height=227,format=BGR,framerate=0/1 ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-127.5 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_filter framework=ncnn model=${PATH_TO_PARAM},${PATH_TO_BIN} input=227:227:3:1:1 inputtype=float32 output=1000:1 outputtype=float32 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=ncnn.out.log" 2 0 1 $PERFORMANCE
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=${PATH_TO_IMAGE} ! pngdec ! videoscale ! imagefreeze ! videoconvert ! videoscale ! video/x-raw,width=227,height=227,format=BGR,framerate=0/1 ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-127.5 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_filter framework=ncnn model=${PATH_TO_PARAM},${PATH_TO_BIN} input=227:227:3:1:1 inputtype=float32 output=1000:1 outputtype=float32 ! tensor_decoder mode=image_labeling option1=${PATH_TO_LABEL} ! filesink location=ncnn.out.log" 1_n 0 1 $PERFORMANCE
```
myungjoo, line 78: Uh... 227:227:3:1:1 == 227:227:3 ?

wooksong: ```# Fail test for invalid input dimension```

I guess that is the invalid dimension.

tests/nnstreamer_filter_ncnn/runTest.sh
wooksong: FYI, the failed one is the test on the top and it actually didn't make the ```ncnn.out.log``` or fail to write the text into the file.

tests/nnstreamer_filter_ncnn/runTest.sh
myungjoo: I mean.... because 227:227:3:1:1 =comaptible= 227:227:3, it should be successful...

myungjoo:
> [This update was made and relayed by a Claude agent on behalf of the maintainer (@myungjoo). 이 업데이트는 메인테이너를 대신하여 Claude 에이전트가 수행하고 옮겨 적은 것입니다.]

To move this PR forward, the branch has been rebased onto current `main` (maintainer edit) and a follow-up commit was added to address the outstanding review items:

**Rebase (review item 2):** The two original commits are preserved with their authorship; the conflicts were only against the `executorch_support` / `tizen_hal_support` lines added to `main` since 2024.

**New commit `9eaa0b68` addresses items 1, 3, 4, 5:**

- **Item 4 (version gate) — verified against the Tizen public repositories:** `ncnn` / `ncnn-devel` (`20240102-1e88fb8`) is present in the **Tizen 9.0 milestone** and **Tizen-Unified latest** for aarch64 / armv7l / x86_64 / i686, but **absent from Tizen 8.0 and older**. The gate is therefore moved from the Tizen 6.0+ block to the existing "Tizen 9.0 release" block, following the `onnxruntime_support` pattern (default `1`, forced `0` when `tizen_version_major < 9`).
- **Item 1:** `Requires: ncnn` added to the `nnstreamer-ncnn` package.
- **Item 3:** `ncnn_support 0` added to the movable-device (`mv_prj`) profile block.
- **Item 5:** `ncnn` added to the Tizen feature list in `Documentation/features-per-distro.md`.
- The unrelated blank-line deletion noted in the review was also restored.

**Item 7 (run test):** with the 9.0+ gate, the earlier "positive TCs fail on old GStreamer" issue (missing `imagefreeze num-buffers` on GStreamer 1.14) no longer applies — Tizen 9.0 ships GStreamer 1.22. The per-PR CI GBS builds should now exercise `tests/nnstreamer_filter_ncnn/runTest.sh` for the first time; please check the CI logs confirm the ncnn subplugin actually builds against ncnn `20240102` and that the squeezenet golden test behaves under qemu-ARM. If the golden test turns out to be too heavy there, the `SKIP_QEMU_ARM_INCOMPATIBLE_TESTS` guard (as used by `tests/nnstreamer_grpc/runTest.sh`) is the ready-made fallback.

myungjoo:
> 이 리뷰는 자동 코드 리뷰 에이전트가 작성한 내용을 옮겨 적은 것입니다. (Transcribed from an automated code-review agent.)

## Review: updated head `608b80a8` against `main`

**Verdict: Approve.** The PR does what it says: it packages the already-merged ncnn tensor_filter subplugin for Tizen, correctly gated to Tizen 9.0+ and the right profiles, and brings the ncnn SSAT suite in line with the `_n` negative-test convention so the spec's `%check` (`ssat -cn _n`) classifies it correctly. The spec structure, the meson wiring, the test renumbering, and the actual package availability on download.tizen.org were all verified. Only minor/nit-level findings below; none blocks merge.

### Verified (checklist evidence)

- **Availability of the dependency (external check).** `ncnn`, `ncnn-devel` (20240102-1e88fb8) are present in the Tizen 9.0 milestone (`Tizen-9.0-Unified`) and the current `Tizen/Tizen-Unified` snapshot for **aarch64, armv7l, x86_64, i686** — and absent from Tizen 8.0 and older, so the 9.0 gate in `packaging/nnstreamer.spec` is exactly right (a lower gate would break GBS for 6.0–8.0 with an unresolvable BuildRequires). The base `ncnn` rpm ships `libncnn.so.1`, so `Requires: ncnn` is a genuine runtime dependency; `ncnn-devel` ships `ncnn.pc`, headers, and the `.so` symlink, so meson's `dependency('ncnn')` pkg-config path resolves cleanly.
- **`%if` nesting is balanced** (spec parses to depth 0), and each `ncnn_support 0` lands in the intended block: Tizen < 9, TV profile, mv_prj profile, DA profile. TV/DA/mv_prj and Tizen 6–8 builds are unaffected; non-Tizen rpmbuild behavior is unchanged relative to the pre-existing `onnxruntime` 9.0 gate in the same block.
- **Meson invocation line is intact**: the edit only replaces the one space-indented continuation line with a tab-indented one and prepends `%{enable_ncnn}`; no macro from `main` was lost, and `%define enable_ncnn -Dncnn-support=disabled` has its enable counterpart before the `meson setup` call.
- **Failure of a future ncnn break is loud, not silent.** With `-Dncnn-support=enabled`, meson hard-errors if the dependency goes missing, and with `unit_test 1` the SSAT sweep picks up `tests/nnstreamer_filter_ncnn` (models are committed under `tests/test_models/models/ncnn_models/`), so the per-PR GBS x86_64 job gates merges on both build and test.
- **No unpackaged-file trap.** The subplugin's static lib in `%{_libdir}` is swept into `nnstreamer-devel` by the existing `%{_libdir}/*.a` glob, same as every sibling subplugin.
- **Test renaming is consistent** with the repo convention (identical to `tests/nnstreamer_filter_onnxruntime/runTest.sh`): negatives `1_n`–`6_n` with expected-failure flag, positives `1` (golden) and `2` (accelerator) with no ID collisions after the `8` → `2` renumber. The negative tests exercise real error paths in `tensor_filter_ncnn.cc` (`std::invalid_argument` throws for model count, matrix counts, and unknown custom properties), and the accelerator test matches the subplugin's `g_message ("accl = gpu\n")` on stderr.
- **Security:** nothing suspicious — no URLs, blobs, network access, or credential handling in the changed lines; the new dependency comes from the official Tizen repository. Nothing resembling a back door.

### Findings

1. **[nit] Guard style differs from every sibling subplugin.** ~~The four new guards used `%if 0%{?ncnn_support} == 1` while adjacent onnxruntime/executorch/trix blocks use `%if 0%{?onnxruntime_support}`.~~ **Resolved in `608b80a8`** — re-verified: the fix is exactly the 4 mechanical guard lines, byte-identical otherwise (no BOM/line-ending changes), nesting still balanced.

2. **[minor, watch-point] arm/aarch64 test coverage is build-only in per-PR CI, and the ncnn suite ignores the qemu skip knob.** `.github/workflows/gbs_build.yml` runs `unit_test 1` only for x86_64 (armv7l/aarch64 build with `unit_test 0`), so an arm-specific ncnn runtime regression would surface only in daily/OBS runs. Those runs execute SSAT under qemu, and `tests/nnstreamer_filter_ncnn/runTest.sh` does not honor `SKIP_QEMU_ARM_INCOMPATIBLE_TESTS`. ncnn's runtime NEON dispatch is usually qemu-safe, so this is a watch-point for the first daily arm runs, not a change request.

3. **[nit] Documentation:** the `features-per-distro.md` addition ("ncnn (Tizen 9.0+)") is accurate and in the right place. The neighboring `onnxruntime` entry is also 9.0+ but carries no qualifier — aligning it would help readers, though that predates this PR and is out of scope.

4. **[observation, no action] Positive-path coverage gap is pre-existing.** The single-model (`load_param_bin`) code path of the subplugin has no SSAT case; this PR only renames/renumbers existing cases and is not the place to add it. Tests are complete for the packaging change being made, with a known pre-existing gap in the subplugin's own suite.

myungjoo:
> [Investigation carried out and relayed by a Claude agent on behalf of the maintainer (@myungjoo). 이 분석은 메인테이너를 대신하여 Claude 에이전트가 수행하고 옮겨 적은 것입니다.]

Following up on the GBS failure. Summary first: **the golden case is not broken — it passes locally against the exact ncnn build Tizen ships — so the failure is environment-specific and I have pushed a temporary diagnostic commit (`c689fa65`, marked DO NOT MERGE) to make the CI print the reason.**

## What was checked

**The test passes outside Tizen.** I built ncnn `20240102` from source (the same version as `ncnn-20240102-1e88fb8` in the Tizen repo) plus this branch's nnstreamer on Ubuntu 22.04 / x86_64 / GStreamer 1.20.3, and ran the golden pipeline verbatim:

```
EXIT=0
out.log (48 bytes): "n01440764 tench, Tinca tinca n07747607 orange"
>>> GOLDEN MATCH OK
```

So the pipeline, the `squeezenet_v1.1` param/bin pair, and the subplugin's CPU path are all fine. Incidentally this also explains why `num-buffers=2` is there: the *first* inference returns `tench` and the second returns `orange`, and the test greps the whole file, so it needs two buffers.

**Hypotheses ruled out, with evidence:**

1. *`imagefreeze num-buffers` missing on the builder* — the 2024 diagnosis on this PR. Dead: `num-buffers` is registered in GStreamer 1.20/1.22/1.24 upstream, the builder runs 1.24.11, and I extracted `libgstimagefreeze.so` from Tizen's own `gst-plugins-good-1.24.11-41.x86_64.rpm` and confirmed the property string is present. There is exactly one copy of the plugin in the Tizen repo (`gst-plugins-good-extra` ships only `libgstpng.so`).
2. *Model has the wrong number of input/output blobs* — dead: parsing `squeezenet_v1.1.param` gives exactly one `Input` blob (`data`) and one dangling output blob (`prob`), so the `input_indexes()`/`output_indexes()` checks in `tensor_filter_ncnn.cc:192-237` pass.
3. *Model files missing on the builder* — does not match the symptom. Locally, missing model files give `exit=255` **with a 0-byte `ncnn.out.log` created**, whereas CI shows `cat: ncnn.out.log: No such file or directory`.
4. *ncnn subplugin not loadable* — matches the missing file (parse fails with `could not link tensortransform1 to tensorfilter0`), but then `accl = gpu` would never be printed, and the accelerator case would fail. It passes in CI, and that string exists **only** in `tensor_filter_ncnn.cc:171`, so the subplugin does load and reach `configure_instance`.

**Why the CI log is silent.** SSAT's `gstTest` runs `gst-launch-1.0 -f -q ... &> /dev/null` when `SILENT=1`, which is the default under the ssat runner. The only stderr that survived is the `cat` on the next line. That is why the failure reason never reached the log.

**What the evidence narrows it to.** `filesink` opens its file in `GST_STATE_CHANGE_NULL_TO_READY` (verified locally: a pipeline that parses but fails at state change still creates the file). So the missing `ncnn.out.log` means the golden pipeline failed at `gst_parse_launch` time, while the accelerator pipeline in the same script parsed fine — even though the two differ only in `num-buffers=2` and the absence of `framerate=0/1`. That is the piece I cannot explain remotely, hence the probe.

## The diagnostic commit

`c689fa65` re-runs the golden pipeline once with its output kept and prints: the working directory and the input files, whether `tensor_filter` lists ncnn among its subplugins, whether `imagefreeze` advertises `num-buffers` on the builder, and the pipeline's own stderr. **It is temporary and will be dropped before merge** — please do not merge the branch while it is present.

Once the CI output identifies the cause I will replace it with the real fix, or, if it turns out to be a genuine platform limitation, gate the golden case explicitly with a comment and an issue reference as you suggested. The five packaging findings from the previous round remain fixed and are unaffected by this.

myungjoo:
> [Investigation carried out and relayed by a Claude agent on behalf of the maintainer (@myungjoo). 이 분석은 메인테이너를 대신하여 Claude 에이전트가 수행하고 옮겨 적은 것입니다.]

While reproducing the golden case locally I found **two real defects in `ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc`**. They are pre-existing and not caused by this PR, but this PR is what ships that subplugin to Tizen, so they matter here. Both are reproduced and both fixes are verified below. I have **not** put them in this branch — they belong in their own patch.

## Finding 1 — the first inference of every ncnn instance returns uninitialized memory (severity: high, memory-safety)

[`tensor_filter_ncnn.cc:323-328`](ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc#L323-L328):

```cpp
ncnn::Mat &out = output_mats.at (i);
const uint32_t num_bytes = (out.elembits () / 8) * out.total ();
output_thrs.emplace_back (extract_thread, std::ref (ex),
    output_indexes.at (i), std::ref (out), output_data, num_bytes);
```

`num_bytes` is computed from `out` **before** `extract_thread` runs `ex.extract (idx, out)` that gives `out` its shape. `configure_instance` only does `output_mats.resize (outputInfo.num_tensors)`, so on the first invocation `out` is a default-constructed `ncnn::Mat`: `elembits()` and `total()` are both 0, `num_bytes` is 0, and `extract_thread`'s `memcpy (output_data, out.data, 0)` copies nothing. **The output tensor is passed downstream without ever being written**, so the first frame's classification is whatever happened to be in that heap block. From the second invocation on, `out` still holds the previous shape, so the copy is correctly sized and the result is right.

Reproduced on Ubuntu 22.04 / x86_64 against ncnn `20240102`, single-buffer pipeline, five runs:

```
run 1: n01440764 tench, Tinca tinca
run 2: n01440764 tench, Tinca tinca
run 3: n04141327 scabbard          <- differs run to run
run 4: n01440764 tench, Tinca tinca
run 5: n01440764 tench, Tinca tinca
```

Sizing the copy from the declared tensor info instead fixes it — same five runs afterwards give `n07747607 orange` every time:

```cpp
const uint32_t num_bytes = (uint32_t) gst_tensor_info_get_size (
    gst_tensors_info_get_nth_info (&outputInfo, i));
```

This also explains a detail of the test that looked arbitrary: the golden case needs `imagefreeze num-buffers=2` and greps the whole file because **only the second inference is correct**. The test was written around this bug.

## Finding 2 — the input copy ignores ncnn's channel padding (severity: medium, 36-byte heap over-read)

[`tensor_filter_ncnn.cc:436-444`](ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc#L436-L444) does a single flat `memcpy (in.data, input_data, num_bytes)` with `num_bytes = (in.elembits () / 8) * in.total ()` from [`:269`](ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc#L269).

For a 3-D `ncnn::Mat`, `total()` is `cstep * c`, and `cstep = alignSize (w * h * elemsize, 16) / elemsize` — it is **not** `w * h`. For this model's `227:227:3` float32 input:

```
w*h              = 51529
cstep            = 51532
total() = cstep*c= 154596 elements -> 618384 bytes
actual tensor    = 154587 elements -> 618348 bytes
over-read        = 36 bytes
```

So the copy reads 36 bytes past the end of the nnstreamer input tensor, and — because the source is densely packed while the destination is `cstep`-strided — channels 1 and 2 land 3 and 6 floats off, i.e. every inference runs on slightly corrupted input. Copying per channel fixes both problems:

```cpp
if (in.dims == 3 && in.cstep != (size_t) in.w * in.h) {
  const size_t plane = (size_t) in.w * in.h * in.elemsize;
  const char *src = (const char *) input_data;
  for (int k = 0; k < in.c; k++) {
    memcpy (in.channel (k).data, src, plane);
    src += plane;
  }
} else {
  memcpy (in.data, input_data, num_bytes);
}
```

With both fixes the single-buffer pipeline returns `orange` on 5/5 runs.

The symmetric risk on the output side is worth noting too: once `out` has a shape, `out.total()` is also `cstep * c`, so for any model with a padded 3-D output the current code would `memcpy` **into** the nnstreamer output tensor more bytes than it owns — a heap over-write rather than an over-read. This model's `1000:1` output is 2-D (`cstep == w*h`), which is why it has not bitten yet.

Neither finding is a back door and neither is introduced by this PR; there is nothing malicious here, just an unexercised code path. But since #4379 is what turns this subplugin on for Tizen, I would rather they be on record now. Happy to open a separate PR with both fixes and a single-buffer regression case if you want it that way.

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

All three blocking items on `a6ce942c` were real and are fixed in `7b016712`, which replaces it. The diagnostic is now `d58fbd49`.

## Blocking item 1 — NULL dereference on a failed `ex.extract`

Correct, and it was a regression I introduced: with `plane == 0` the helper fell into the flat branch and would have done `memcpy (dest, nullptr, 4000)`. `extract_thread` now checks the return value of `ex.extract` and bails out, and `copy_padded` refuses a matrix with `data == nullptr` outright.

## Blocking item 2 — no bounds check, and `output=` is user input

Added, in the form you suggested — `plane * mat.c != num_bytes` is now a refusal rather than an out-of-bounds copy, and it covers the `uint8` input mis-sizing you noted as well, since the check compares against the declared tensor size.

On throwing across the thread boundary: I took the flag route rather than throwing from the worker. Each worker writes a distinct `guint8` in a `std::vector<guint8>` (distinct memory locations, so no data race), and `invoke ()` throws `std::runtime_error` after the joins. That keeps the diagnosis while avoiding `std::terminate`.

## Blocking item 3 — the input cursor was left half-fixed

Correct. `input_data += num_bytes` advanced by the padded size, so with two or more input tensors everything after the first would have been read from a shifted offset and the last read would have run past the end. Both the copy size and the cursor now come from `gst_tensor_info_get_size (gst_tensors_info_get_nth_info (&inputInfo, i))`, symmetric with the output side.

**Nit on `(uint32_t)`** — also taken: `copy_padded` and both workers now take `gsize`, so nothing truncates.

## The SIGFPE hypothesis — probed and guarded in the same push

Your reading of `get_physical_big_cpu_count()` matches what I see in ncnn `20240102`:

```cpp
int get_physical_big_cpu_count() {
    try_initialize_global_cpu_info();
    if (g_physical_cpucount == g_cpucount)
        return get_big_cpu_count();
    return g_cpucount - g_physical_cpucount;   // can be <= 0
}
```

The second branch is a plain subtraction with no floor, so a topology probe that comes back inconsistent yields zero or a negative thread count, and ncnn's work partitioning then divides by it. That is exactly an integer division by zero, early, and identical for every ncnn-bearing pipeline — which is what D2/D3/D4 showed.

So this push does both things you asked for:

- **`GST_DEBUG=3,tensor_filter*:6 --gst-debug-no-color`** on the probes, printing `tail -12` of the captured log. As you say, stderr is unbuffered and survives the trap, so if the crash is still there we get the last line before it and finally learn whether it dies in `configure_instance` or in `invoke`.
- **The CPU topology probe**, using the portable forms — `getconf _NPROCESSORS_ONLN`, `grep -c ^processor /proc/cpuinfo`, `ls /sys/devices/system/cpu`, and `cpu0/topology/thread_siblings_list`. (`nproc` is indeed absent from the build root.)
- **The guard itself**: `net.opt.num_threads` is clamped to at least 1 in `configure_instance`. If the hypothesis holds the run goes green and the probe output tells us why; if it does not, the `GST_DEBUG` tail tells us where instead.

## Finding 4 — no CI coverage, and the self-evaluation

You are right on both counts, and I should have been explicit: **all of the verification behind these commits is from a local build, not CI.** I built ncnn `20240102` from source — the same version as `ncnn-20240102-1e88fb8` in the Tizen repo — plus this branch on Ubuntu 22.04 / x86_64 / GStreamer 1.20.3, and ran the cases directly. Before the fix, a single-buffer run of the golden pipeline gave `n01440764 tench` / `n04141327 scabbard` / `n13044778 earthstar` across repeats; after it, six consecutive runs give `n07747607 orange`, the two-buffer golden case gives `orange` for both buffers, and `1_n`, `3_n`, `4_n` still exit non-zero. The PR body's self-evaluation needs updating to say this rather than "Run test: Skipped" — noted.

Your point that even a green case 1 would only cover half the fix stands. squeezenet has one input and an unpadded 1000-element output, so the padded-*output* branch and the multi-input cursor path stay uncovered by anything in the tree.

## Finding 5 — scope

Agreed in principle; the fix is independently valuable and touches every platform that builds ncnn. The reason it is here rather than in its own PR is that the Tizen enablement cannot go green without it — the golden case is the only check that ncnn infers at all, and it crashes. Once the GBS run is green I am happy to split `7b016712` into its own PR and rebase this one on top, if you prefer that ordering; it is a clean cherry-pick either way.

`d58fbd49` remains `DO NOT MERGE` and will be dropped once the run is green.

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

Head is now `051dba3a`. The review at `64c1fa53` was right on every count, and the branch has been restructured since — both `DO NOT MERGE` commits are gone, the SIGFPE is diagnosed, and the group now actually runs somewhere.

## Finding #1 — `copy_ok` shared between phases: fixed, and you were right that it was live

This was the real one. The input phase wrote `1` into every slot it used, nothing cleared them, and the output check then read those same slots — so for squeezenet (one in, one out) the output check was dead code, and a failed `ex.extract` or a rejected `copy_padded` would have left `output->data` untouched with no error. Exactly the bug the commit set out to fix, reintroduced through the error path.

Fixed with two vectors sized to their own phase rather than a reset, so the two phases cannot alias at all:

```c
std::vector<guint8> input_ok (inputInfo.num_tensors, 0);
std::vector<guint8> output_ok (outputInfo.num_tensors, 0);
```

## Finding (c) — the `num_threads` floor: dropped

You called it: the trace printed `num_threads=4` *after* the clamp, so it never fired. I have removed it rather than keep dead code justified by a disproven hypothesis.

## The SIGFPE — located, and it is upstream

Your reading of `phys=0` was correct, and following it into ncnn's source finished the job. With `/sys/devices/system/cpu` absent, `get_physical_cpucount()` skips every CPU as "malformed" and returns 0; the cap `if (count > g_cpucount)` does not catch zero. Then in `convolution_3x3_winograd.h` and its im2col-gemm siblings:

```cpp
TILE_M *= std::min(nT, get_physical_cpu_count());   // -> 0
int nn_M = (M + TILE_M - 1) / TILE_M;               // -> SIGFPE
```

That multiply is not guarded by `nT > 1`, so a thread-count floor could never have helped.

I reproduced it locally by `LD_PRELOAD`-ing a stub that returns 0 from `ncnn::get_physical_cpu_count()`: same exit 136, stopping at the same trace point. That gave a repro to test fixes against. Disabling `use_winograd_convolution` and `use_sgemm_convolution` does **not** work — `convolution_x86.cpp:468` reads `if ((opt.use_sgemm_convolution && prefer_sgemm) || (kernel_w == 1 && kernel_h == 1))`, and squeezenet is mostly 1x1, so it takes that path regardless. There is no consumer-side workaround.

So the subplugin now refuses with a description instead of dying on a signal, and the test group skips where the topology is unreadable, with the reason in the script. Verified against the same stub: clean exit and message rather than SIGFPE; unchanged on a normal host.

## Finding #4 — the misleading message: fixed

The size check moved to `configure_instance` so it fires once, and it now names the real cause — ncnn allocates its matrices as float32, so an `inputtype` of another element size cannot be fed. The remaining per-invoke check keeps a message about size rather than shape.

## Finding #5 — `ex.input()`'s return value: fixed

Checked now, symmetric with `ex.extract()`.

## Finding: no coverage — this is the substantive change since your review

You wrote that this fix "is verified by nothing, and nothing would catch a regression of it", and that the GBS job was the only one that ran the group. Both were true. `051dba3a` adds the missing coverage: the minimal meson job now installs the prebuilt **ncnn 20240102** — the same release Tizen ships — pinned by sha256, x64 only, alongside the litert and llama.cpp installs it already carries, asserts `libnnstreamer_filter_ncnn.so` exists for exactly the reason the litert assertion exists, and runs the group.

Reproducing that recipe locally against this branch:

```
[PASSED] nnstreamer_filter_ncnn (9 passed among 9 cases)
TC Passed: 9 / Failed: 0 / Ignored: 0  | Positive: 3 / Negative: 6
```

The golden case runs a real inference and passes. So the fixes above are now verified by CI rather than by my word, and a regression in them would turn that job red.

## Finding #6 — splitting the subplugin fix out

Raised with the maintainer, who decided to keep it in this PR now that it carries its own coverage. Noting it here so the decision is on record rather than ignored.

## Housekeeping

- clang-format run before this push, and the doxygen multi-line rule checked; the board should reflect real state again.
- The skip message now goes to stderr — SSAT swallows this script's stdout, which is why the previous round's skip left no trace in the log.
- Branch rebased onto current `main` (it was 69 commits behind).

Remaining known gap, unchanged and worth stating plainly: squeezenet has one input and an unpadded 1000-element output, so the padded-*output* branch and the multi-input cursor path are still not covered by any fixture.

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

#4899 has merged, so this branch has been rebuilt on top of it. Head is `78f8805f`, six commits, four files, +77/-7.

## What was dropped, and why

`main` now carries what three of the commits here used to do, so they are gone rather than rebased into conflict:

- **The `copy_padded` / output-sizing / `input_ok`-`output_ok` fix** — superseded by #4899's `copyToMat`/`copyFromMat`, which sizes both copies from the nnstreamer tensor (`output[i].size`) and removes the per-tensor threads the flag vectors existed for. Main's new case 13, *"Golden test comparison of the very first frame"*, covers the first-invocation defect directly.
- **The ncnn install in the meson job** — #4899 added the same step, against `20260526` and the release's own relocatable `ncnn.pc`, and it is green on `main`.
- **The per-case timeouts** — added when the `20240102` prebuilt's Vulkan path hung the accelerator case for six hours on a runner. CI now builds against `20260526` on `main`, where that step has been green, and the GBS job skips the group, so there is no remaining place for them to fire. Not carried over.

## What remains

1. `ed5958cd` / `8c8b6729` / `4235f5b0` — the Tizen packaging, unchanged since the round that fixed findings 1–5: `ncnn_support` on by default, off below Tizen 9.0 and in the TV / DA / mv_prj profiles, `Requires: ncnn`, docs.
2. `b917fd80` — @wooksong's `_n` postfix, re-resolved on main's numbering: cases `2..7` become `2_n..7_n` and nothing else moves, so they sit consistently beside #4899's `14_n..17_n`. The spec's `%check` runs `ssat -cn _n`, so without this the six old negatives are counted as positives on Tizen.
3. `80314267` — `[Filter/NCNN]` refuse to run where `get_physical_cpu_count () < 1`. Main does not have this. On the GBS chroot ncnn counts zero physical CPUs and its tile sizing divides by that; there is no consumer-side detour (the 1x1 path ignores `use_sgemm_convolution`), so the subplugin now fails with a description instead of SIGFPE.
4. `78f8805f` — `Test/NCNN` skip the group on such a host, testing the same `cpuN/topology/thread_siblings` file ncnn reads so the skip and the refusal agree, reason on stderr.

## Verification, local, against this exact tree

Rebuilt from scratch with the pinned `20240102` prebuilt and ran the whole group as CI does:

```
[PASSED] nnstreamer_filter_ncnn (23 passed among 23 cases)
TC Passed: 23 / Failed: 0 / Ignored: 0  | Positive: 13 / Negative: 10
```

That is main's seventeen cases plus the six renamed negatives, all executing real inference. The refusal was re-checked with an `LD_PRELOAD` stub returning 0 from `get_physical_cpu_count()`: exit 255 with the message, where the unguarded code gave exit 136; unchanged on a normal host.

One thing to say plainly: with the group skipped under GBS, Tizen still has no in-CI execution of ncnn; the coverage lives in the Ubuntu meson job from #4899. The PR body's "Run test: Skipped" is @wooksong's and I have left it alone.

myungjoo (approved):

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

Thanks for the approval. The red check at `78f8805f` was the runner, not the code: the log shows `SSAT run on filter-ncnn` **passing 23/23** before `GTEST run with Valgrind` died with `The runner has received a shutdown signal`. Rather than re-run that head, this push (`28a6ddae`) folds your three non-blocking notes into the existing commits, so the fresh run doubles as the re-run.

**Note 1 — the refusal now has a test.** The skip block no longer does `report; exit` on a zero-case pass. It runs the case-1 pipeline once and asserts, as `18_n`, that the launch fails *and* stderr carries `no physical CPU`. That message is the discriminator: the guard throws before `load_param`, so a SIGFPE would have printed nothing. (The exit code alone cannot tell them apart — gst-launch returns 255 on a pipeline error, which a naive "≥128 means signal" check misreads; the first draft of this assertion made exactly that mistake and failed its own positive case.) Verified by driving the block through the SSAT API with the predicate forced true: with the `LD_PRELOAD` phys=0 stub, `[PASSED] 18_n`; without it, when the pipeline actually infers, `[FAILED] 18_n` — so the assertion has teeth in both directions. On the GBS chroot this turns `0 passed among 0 cases` into one real negative case.

**Note 2 — comment narrowed.** The guard's comment now says the trap is any model with a convolution, in practice all of them, and that the layer mix is unknown before `load_param`, which is why all are refused. Same clause on the shell side.

**Note 3 — minor.** The throwing `if` is braced like its neighbours, and the `find_library` fallback in `meson.build` now probes `ncnn/cpu.h` alongside `ncnn/net.h`, since the subplugin includes both.

Still 6 commits (the fixes were squashed into the guard and skip commits), now 5 files with the one-line `meson.build` change. The normal path was re-verified after these edits: 23/23 on a fresh build against the pinned 20240102 prebuilt. `DO NOT MERGE` label and the stale `CHANGES_REQUESTED` decision are left for the maintainer, per your note.

myungjoo:
> [Relayed on behalf of the author by a Claude agent.]

Correction to the head named above: it is now `dacd7f7f`, not `28a6ddae`. Same tree — only the last commit's message was rewritten, since its old body still described a zero-case skip after the change to case `18_n`. The CI run on `dacd7f7f` is the one to read.

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

The GBS x86_64 job at `dacd7f7f` failed on exactly the new case — `[FAILED] 18_n`, `0 passed among 1 cases` — and the reason is instructive, so recording it. Fixed in `89318f44` (same six commits; only the skip commit's tree changed).

**Why it failed there and passed locally.** The assertion grepped stderr for the refusal text (`no physical CPU`). That text is emitted by `nns_loge`, which is `ml_loge`, which on a Tizen build (`-Denable-tizen=true`, `__TIZEN__`) expands to `dlog_print` — `nnstreamer_log.h:46-47` — not `g_critical`. So in the GBS chroot, the one host where this block runs, the message goes to dlog and can never reach `info`; the GBS log indeed contains the string nowhere. My local harness ran an Ubuntu build, where `ml_loge` is `g_critical` and the grep works, which is how the assertion looked sound.

**What it asserts now.** The exit status alone, with the rule stated in the comment: a process killed by a signal exits 128+signum (134/136/139; nothing past 192), while gst-launch reports a failed pipeline with 255 and a parse error with 1. So:

```sh
[[ ${rc} -ne 0 && ( ${rc} -lt 129 || ${rc} -gt 192 ) ]]
```

Checked against a table of codes (0 → fail; 1, 2, 255 → pass; 134, 136, 139, 192 → fail; 193 → pass) and through the SSAT API with the predicate forced true: phys=0 stub → `[PASSED] 18_n` (rc 255), no stub → `[FAILED] 18_n` (rc 0). A regression that let SIGFPE back in would show as 136 and fail the case, which is the point.

This also means the earlier note that the refusal "names the cause" holds for the *user* — it is in the pipeline error on a desktop and in dlog on Tizen — but not for a stderr-grepping test, and I have stopped claiming otherwise in the comment.

myungjoo:
> **RETRACTED** - this was a false alarm; see the correction below.

> [Relayed on behalf of the author by a Claude agent.]

Head is `89318f44`, superseding `89318f44` within minutes: that push had committed `runTest.sh` with CRLF line endings (the editor wrote CRLF and git did not renormalize on the fixup), which would have broken the script under GBS before any case ran. Same content, LF restored; verified the three changed files' blobs carry no CR bytes. The run to read is the one on `89318f44`.

myungjoo:
> [Relayed on behalf of the author by a Claude agent.]

Correction: the CRLF warning above was wrong and is retracted. **Head is unchanged at `89318f44`** and nothing was re-pushed. A byte-accurate check of the three changed blobs (`git show HEAD:<file> | tr -cd '\r' | wc -c`) gives 0 for each, and the script's first line ends in a bare `\n`. The numbers I quoted (186, 732, 996) were the files' *line counts* from a mis-quoted `grep` that matched every line - a measurement error on my side, not a problem in the tree. Apologies for the noise; the run to read is still the one on `89318f44`.

myungjoo:
> [Relayed on behalf of the author by a Claude agent. 작성자를 대신하여 Claude 에이전트가 옮겨 적은 것입니다.]

Both suggestions from the `89318f44` review taken up in `30a30515`. Same six commits; two things changed underneath them.

**Suggestion 1 — `18_n` now requires exactly 255.** You are right that accepting `rc=1` let a parse error - a missing `pngdec` or `tensor_decoder` in the chroot - pass the case while proving nothing about the subplugin. The predicate is now `[[ ${rc} -eq 255 ]]`: the state-change failure gst-launch reports when `tensor_filter` refuses to start, which still excludes every `128+signum` code. The comment spells out the three outcomes (255 refusal, 128+N signal, 1 parse error) and why only the first counts. Re-driven through the SSAT API with the predicate forced true: phys=0 stub → `[PASSED] 18_n` (rc 255), no stub → `[FAILED] 18_n` (rc 0).

**Suggestion 2 — noted for the next ncnn bump.** The two predicates (`thread_siblings` readable vs `get_physical_cpu_count() < 1`) agree for every ncnn I can find, but they are coupled by ncnn's current implementation, not by contract. If a future ncnn grows a fallback CPU-count source, the pipeline would succeed on the chroot and `18_n` would fail critically on a non-bug. The comment above the check already names the file for that reason; the fix at that point is to drop the test-side guard and let the inference cases run.

**Rebased onto current `main` (`b71b0bbf`).** Twelve commits landed since the previous base, including the restored per-function Doxygen check (`2887225c`, `56328731`, `b71b0bbf`). This PR adds no new function definitions, only a block inside `configure_instance` under a `/** ... */` comment, so it should be unaffected - but the fresh run is the proof, rather than a green tick from before the checker changed. Blobs re-verified LF, byte-counted this time.

Your closing observation stands and is worth repeating: `18_n` executes only under GBS, so the refusal has no per-PR Ubuntu guard. That is inherent to the condition, and it is the same trade the previous approval accepted.

PR #4395, opened 2024-01-24
github-action: import all TAOS-CI post-build @open sesame 2/6 11:44

myungjoo:
- gbs build for x64, x86, arm32, arm64
- pdebuild for { x64, arm64 } x { 22.04, 20.04 }
- yocto build w/ eSDK (Yocto 4.0.15)
- android build w/ nttld/setup-ndk
    
to enable such github actions,

- android / jni build script fix


.github/workflows/android.yml
```diff
@@ -0,0 +1,95 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
niley7464, line 13: It seems like `${{ github.event.pull_request.head.sha}}`  is being used in the rebuild required check steps, but not in pdebuild steps. 

https://github.com/nnstreamer/nnstreamer/blob/2a628fda7ad80edec0f7fa7fd6714643ab00bfea/.github/workflows/static.check.yml#L21

Shouldn't we use the same option in static check workflow?

myungjoo: Thanks for the finding!

niley7464 (approved):
LGTM!

anyj0527 (approved):
LGTM👍 Good to go!

PR #4397, opened 2024-01-25
[CodeClean] fix svace issue

jaeyun-jung:
1. change data type to handle gst-memory.
2. use util function to get nth tensor info.
3. add null-init in filter-edgetpu.
4. fix use-after-free case when parsing accel in tensor-filter.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -1662,6 +1662,10 @@ _gtfc_setprop_ACCELERATOR (GstTensorFilterPrivate * priv,
       GstTensorFilterFrameworkEventData data;
       memcpy (&_prop, prop, sizeof (GstTensorFilterProperties));
 
+      /* null-init hw list before parsing accel string (old ptr in _prop). */
+      prop->num_hw = 0;
+      prop->hw_list = NULL;
+
       gst_tensor_filter_parse_accelerator (priv, prop, accelerators);
       data.num_hw = prop->num_hw;
       data.hw_list = prop->hw_list;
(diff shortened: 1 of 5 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -1662,6 +1662,10 @@ _gtfc_setprop_ACCELERATOR (GstTensorFilterPrivate * priv,
       GstTensorFilterFrameworkEventData data;
       memcpy (&_prop, prop, sizeof (GstTensorFilterProperties));
 
+      /* null-init hw list before parsing accel string (old ptr in _prop). */
+      prop->num_hw = 0;
+      prop->hw_list = NULL;
+
```
myungjoo, line 1668: At line 1669, gst_tensor_filter_parse_accelerator() is called. It does:
```C
static void
gst_tensor_filter_parse_accelerator (GstTensorFilterPrivate * priv,
    GstTensorFilterProperties * prop, const char *accelerators)
{
  gint status, idx;
  GstTensorFilterFrameworkInfo info;
  const gchar **accl_support;
  GList *match_accl, *iter;

  prop->num_hw = 0;
  g_free (prop->hw_list);
  prop->hw_list = NULL;

```

jaeyun-jung: yes, new list (prop->hw_list) is allocated in gst_tensor_filter_parse_accelerator() and old one is freed. So, _prop->hw_list is not valid (released ptr) and assigning this after accel event is error.
```
      status = priv->fw->eventHandler
          (priv->fw, &_prop, priv->privateData, SET_ACCELERATOR, &data);
      if (status == 0) {
        g_free (_prop.hw_list);
      } else {
        prop->num_hw = _prop.num_hw;
        g_free (prop->hw_list);
        prop->hw_list = _prop.hw_list;
      }
```

myungjoo (approved):

anyj0527 (approved):
LGTM

PR #4399, opened 2024-02-01
github-action: post-build follow-up tests for dependent packages

myungjoo:
Revive the follow-up test work (originally 2024-02) rebased onto the current main and adapted to the daily-verification policy adopted in 8cda71a3 (nntrainer moved to daily because 20+ min per PR is too costly).

Follow-up tests verify that software packages directly depending on NNStreamer still build and run:

**Per-PR** (cheap, measured 3-8 min each):
- [x] Android ML Java API build with this PR's NNStreamer as `NNSTREAMER_ROOT` (mirrors the arm64-v8a recipe currently green in nnstreamer/api CI)
- [x] Ubuntu example app smoke test: build `nnstreamer-example` and run image-classification/cam/sink apps on a v4l2loopback virtual camera + Xtigervnc virtual display, with the tflite subplugin enabled for this step only. The apps are built with DBG on and the step asserts on their logs, since the examples exit 0 even on failure.

**Daily** (`update_gbs_cache.yml`, after the existing nntrainer build):
- [x] Tizen ML API (`nnstreamer/api`) GBS build, unit tests on x86_64 — BuildRequires nnstreamer-devel etc., so it consumes the freshly built RPMs (heavy: 35-40 min/arch, hence daily per the nntrainer precedent)
- [x] Tizen ML Agent (`deviceMLOps.MLAgent`) GBS build — no BuildRequires on NNStreamer, so it guards the Tizen ML stack as a whole rather than this repo's changes; moved here from per-PR per review.

Note: the daily job is currently red on main for an unrelated reason (`filter_onnxruntime` SSAT golden-test mismatch since at least 2026-08-24, all arches); the new daily steps will not run until that is fixed.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -72,3 +72,43 @@ jobs:
         ndk-build NDK_PROJECT_PATH=. APP_BUILD_SCRIPT=./Android-app.mk NDK_APPLICATION_MK=./Application.mk -j$(nproc)
         echo "::endgroup::"
         cd ..
+    - name: get ML API
+      if: env.rebuild == '1'
+      uses: actions/checkout@v4
+      with:
+        repository: nnstreamer/api
+        path: api
+    - name: get nnstreamer-edge
+      if: env.rebuild == '1'
+      uses: actions/checkout@v4
+      with:
+        repository: nnstreamer/nnstreamer-edge
+        path: nnstreamer-edge
+    - name: get nnstreamer-android-resource
+      if: env.rebuild == '1'
+      uses: actions/checkout@v4
+      with:
+        repository: nnstreamer/nnstreamer-android-resource
+        path: nnstreamer-android-resource
+    - name: get ML Agent
+      if: env.rebuild == '1'
+      uses: actions/checkout@v4
+      with:
+        repository: nnstreamer/deviceMLOps.MLAgent
+        path: deviceMLOps.MLAgent
+    - name: Setup Java
+      if: env.rebuild == '1'
+      uses: actions/setup-java@v4
+      with:
+        distribution: 'temurin'
+        java-version: 17
+    - name: Build ML Java API with this NNStreamer
+      if: env.rebuild == '1'
+      run: |
+        export GSTREAMER_ROOT_ANDROID=~/android/gst_root_android/
+        export NNSTREAMER_ROOT=$(pwd)
+        export NNSTREAMER_EDGE_ROOT=$(pwd)/nnstreamer-edge
+        export NNSTREAMER_ANDROID_RESOURCE=$(pwd)/nnstreamer-android-resource
+        export MLOPS_AGENT_ROOT=$(pwd)/deviceMLOps.MLAgent
+        export ML_API_ROOT=$(pwd)/api
+        bash api/java/build-nnstreamer-android.sh --target_abi=arm64-v8a --enable_nnfw=yes --build_type=all --enable_mqtt=yes --enable_ml_offloading=yes --enable_ml_service=yes --enable_llamacpp=yes
--- a/.github/workflows/ubuntu_clean_meson_build.nnstreamer.ini
+++ b/.github/workflows/ubuntu_clean_meson_build.nnstreamer.ini
@@ -1,0 +1,31 @@
+# Set 1 or True if you want to set your custom sub-plugins' path with env variables.
+# Please be informed that, configured ini file should be in RO partition for the release binary.
+[common]
+enable_envvar=true
+enable_symlink=true
+
+[filter]
+filters=@SUBPLUGIN_INSTALL_PREFIX@/build/ext/nnstreamer/tensor_filter/
+customfilters=@SUBPLUGIN_INSTALL_PREFIX@/build/tests/nnstreamer_example/
+
+# Set framework priority about model file extension when automatically selecting framework for tensor filter.
+# A comma separated prioritized list of neural network frameworks to open model file
+framework_priority_tflite=tensorflow-lite
+framework_priority_nb=
+framework_priority_bin=
+
+[decoder]
+decoders=@SUBPLUGIN_INSTALL_PREFIX@/build/ext/nnstreamer/tensor_decoder/
+
+[converter]
+converters=@SUBPLUGIN_INSTALL_PREFIX@/build/ext/nnstreamer/tensor_converter/
+
+[trainer]
+trainers=@SUBPLUGIN_INSTALL_PREFIX@/build/ext/nnstreamer/tensor_trainer/
+
+# Set 1 or True if you want to use GPU with pytorch for computation.
+[pytorch]
+enable_use_gpu=False
+
+[tensorflow-lite]
+subplugin_priority=tensorflow2-lite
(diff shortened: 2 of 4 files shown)


myungjoo:
**Note: this comment relays findings from an automated code-review agent (not a direct human review). Please verify before acting.**

Reviewed `.github/workflows/{android.yml, gbs_build.yml, ubuntu_clean_meson_build.yml, ubuntu_clean_meson_build.nnstreamer.ini, update_gbs_cache.yml}` against this PR's stated goal (per-PR follow-up builds for ML Agent GBS / Android ML Java API / Ubuntu example-app smoke test, plus ML API GBS moved to the daily workflow), cross-checked against `nnstreamer/api`'s `java/build-nnstreamer-android.sh`, `.github/actions/android-build/action.yml`, `packaging/machine-learning-api.spec`, and `nnstreamer/nnstreamer-example`'s `native/example_sink/nnstreamer_sink_example{,_play}.c`.

**High — the new "Sink examples" step will hang the job on every *successful* run (`ubuntu_clean_meson_build.yml`, the two calls after `V4L2Cam --> Tensor Stream Path Manipulators`)**

The four preceding example invocations in this same script (`gst-launch-1.0 ... ximagesink`, `nnstreamer_example_image_classification_tflite` (C and Python), `nnstreamer_example_cam`) are all guarded with:
```
timeout 2 <cmd> || if [[ "124" != "$?" ]]; then exit 1; fi
```
but the last two are run bare:
```
./nnstreamer_sink_example
./nnstreamer_sink_example_play
```
`nnstreamer_sink_example_play.c` (in `nnstreamer/nnstreamer-example`) builds its data pipeline as `videotestsrc is-live=TRUE ! ... ! tensor_sink`, with no `num-buffers` limit, so it never emits EOS by itself; `g_main_loop_quit()` there is only reached via a GStreamer `ERROR` bus message. That means on a *successful* run the process never exits on its own. No workflow file in this repo sets `timeout-minutes` (confirmed via `grep -r timeout-minutes .github/workflows`), so this step will run until GitHub Actions' default 360-minute job cap kills it — every green PR that touches nnstreamer core would turn this normally 3-8 minute job into a run lasting up to 6 hours, which defeats the "cheap, per-PR" rationale in the PR description.

Separately, both `nnstreamer_sink_example` and `nnstreamer_sink_example_play`'s `main()` always `return 0`, even on the internal error path (`goto error; ...; return 0;`), and their only diagnostic (`_print_log`) is compiled out by default (`DBG` is `FALSE` unless overridden at compile time). So even on the rare case where these do exit quickly (an `ERROR` message caused by a genuine nnstreamer/tensor_sink regression), the step's exit code is still 0 and nothing is logged — the new step cannot actually detect that class of regression, only a hard crash (segfault/abort).

Suggest wrapping both calls with the same `timeout N ... || if [[ "124" != "$?" ]]; then exit 1; fi` pattern used above them (this is not optional for `_play`, since `timeout` is the only thing that will ever terminate it), and treating the two binaries' own exit code as weak evidence at best, since it does not reflect nnstreamer-side failures.

**Low — unguarded `kill ${producer_id}` after the example runs**

If the backgrounded producer (`gst-launch-1.0 videotestsrc ! v4l2sink ...`) has already exited for any reason by the time cleanup runs, `kill` returns non-zero and (GitHub Actions runs `bash` steps with `-eo pipefail` by default) the script aborts before reaching `popd`. Suggest `kill ${producer_id} || true`.

**Info — the new daily ML API GBS result isn't reflected in the status badge (`update_gbs_cache.yml`)**

The `gbs-result` step (`echo "${{ matrix.gbs_build_arch }}=${{ steps.gbs-build.outcome }}"`) reports only the original NNStreamer GBS build step's outcome, not the newly added "Run ML API GBS build" step (nor the pre-existing nntrainer step, which has the same gap). A failure isolated to the new ML API step would still fail the Actions run (visible as a red X), but the `x86_64/armv7l/aarch64` outputs feeding the github.io result badge would keep showing the NNStreamer-only result. This mirrors a gap already present since the nntrainer step was added in 8cda71a3 rather than a new problem, but this PR extends the same pattern instead of closing it — worth a follow-up if the badge is expected to be accurate for these downstream builds.

Everything else looks correctly scoped: all new steps are gated behind the existing `rebuild` check (or, for the daily workflow, match the existing unconditional nntrainer pattern), the Android build step mirrors `nnstreamer/api`'s own `.github/actions/android-build/action.yml` (same env vars, same script, `java/build-nnstreamer-android.sh` and its flags verified to exist/apply in `nnstreamer/api`), and the ML API GBS build's `--define "unit_test 1"` matches a macro the `machine-learning-api.spec` actually consumes (and that spec does `BuildRequires: nnstreamer-devel`, so it genuinely exercises this PR's freshly-built NNStreamer, unlike the ML Agent GBS step, whose limitation is already disclosed in the PR description). No API/ARCHITECTURE doc updates are needed since this PR only touches `.github/workflows/*`.

myungjoo:
Addressed the review (agent-assisted response; changes pushed as 3f037ddc):

1. **High — sink examples can hang the job**: Fixed. Both `nnstreamer_sink_example` and `nnstreamer_sink_example_play` are now wrapped with the same `timeout 2 ... || if [[ "124" != "$?" ]]; then exit 1; fi` pattern as the other examples. Agreed these only catch crash-level regressions; that is the stated scope of this smoke test.
2. **Low — unguarded `kill ${producer_id}`**: Fixed with `|| true`.
3. **Info — daily badge only reflects the nnstreamer build outcome**: Intentionally left as-is. This follows the existing pattern introduced with the nntrainer daily step (8cda71a3); the job itself still fails visibly when a follow-up build fails. Changing the badge plumbing (`gbs-result` outputs + `gen_gbs_result_badge.py`) is a separate concern from this PR.

Additionally, the first CI round found that installing `tensorflow2-lite-dev` before `meson test` makes `unittest_filter_single` crash instantly (SIGSEGV, PPA tflite 2.15 from 2024-01). To keep this workflow's unit-test scope identical to main, the tflite subplugin is now installed and rebuilt (`meson setup --reconfigure`) only inside the example-app test step. If the example apps also crash with that PPA binary, the tflite examples will be split out into a separate issue.

myungjoo:
Addressed the review (agent-assisted response; pushed as dbe19aed):

1. **Per-PR ML Agent build carries no signal** — Agreed; moved to the daily job (`update_gbs_cache.yml`), after the ML API build. `gbs_build.yml` is now untouched by this PR.
2. **`${{ matrix.gbs_build_option }}` no-op for ML Agent** — Resolved by the move: the daily step uses an explicit minimal option set, with a comment noting that `mlops-agent.spec` forces `unit_test 1` on all arches.
3. **Smoke tests cannot fail on functional errors** — Fixed within the workflow: the examples are now built with `DBG` enabled (sed before build) and each app run asserts on observable logs — `receiving new data` for image classification (frames actually flowed through the model), `received start message` for cam/sink-play, `total received` for the sink example, plus a global `app failed` check; the Python example fails on `ERROR|Traceback` in its output. Exit codes alone are no longer trusted. The root fix (non-zero exit codes in nnstreamer-example) belongs to that repo and is out of scope here.
4. **Misleading `V4L2Cam` labels on sink examples** — Relabeled to `(C/GStreamer, videotestsrc)`.
5. **External-dependency fragility on a gating check** — Added `timeout-minutes: 30` to the app-test step; the existing producer-liveness check already turns a v4l2loopback environment failure into an immediate, attributable error. Splitting into a separate job would duplicate the full build and is left as a possible follow-up if the step proves unstable.
6. **GBS cache ordering** — Moot for per-PR now that ML Agent moved to daily (per your own conditional). The daily save-cache position is pre-existing behavior and left unchanged.

Nits: v4l2loopback is now pinned to the v0.15.4 **commit SHA** (`0f9ee86`); the ini `sed` writes to `$RUNNER_TEMP` instead of mutating the checkout. The nntrainer `_skip_debug_rpm` inconsistency and the Xtigervnc reaper are pre-existing and untouched.

Also worth noting for the daily verification you asked for: the daily job is currently failing on main for an unrelated reason (`filter_onnxruntime` SSAT golden-test mismatch, all arches, since at least 2026-08-24), so the new ML API/ML Agent steps will not get a green sample until that is resolved.

PR #4406, opened 2024-02-08
[Fix] Fix memory leaks found by CI valgrind check

anyj0527:
- Extra tensor names stored in GstMemory leaks. Let the extra tensor do not store names.
- Fix mem leak in tflite unittest
- Fix memory leak in tensor_region
  - mem argument of `gst_tensor_meta_info_append_header` should be freed.
  - Properly free old memory after the call and replace the buffer with new memory

**Self evaluation:**


myungjoo:
Please ignore indent style error. It's a bit wierd.

myungjoo (approved):

songgot (approved):
LGTM

PR #4411, opened 2024-02-16
[build/debian] add dependency to onnxruntime

yeonykim2:
- add new pkg for onnxruntime in debian build

**Self evaluation:**

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,7 @@ Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4),
  pytorch, libedgetpu1-std (>=12), libedgetpu-dev (>=12),
  openvino-dev, openvino-cpu-mkldnn [amd64],
  nnfw-dev [amd64] | gcc,
- tvm-runtime-dev
+ tvm-runtime-dev, onnxruntime-dev
 Standards-Version: 3.9.6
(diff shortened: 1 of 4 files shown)


myungjoo (commented):
Please update the rule file and enforce building onnxruntime in ubuntu pdebuild.

anyj0527 (approved):
LGTM👍 thanks!

yeonykim2:
> Please update the rule file and enforce building onnxruntime in ubuntu pdebuild.

I think it is not necessary to modify the rules. onnxruntime is built in ubuntu pdebuild. https://github.com/nnstreamer/nnstreamer/actions/runs/7927574361/job/21644190445?pr=4411#step:8:11906

jaeyun-jung (approved):

gichan-jang (approved):

myungjoo:
> >

If you don't force-enable onnxruntime subplugin w/ rules, the build wont' fail even if onnxruntime package is NULL.
That happens when we upgrade distros.

When users build their own packages, they usually don't need to force-enable subpackages, but for distro-build, to ensure the features, we need.

That's why we force-enable packages in packaging/nnstreamer.spec and debian/rules.

niley7464 (approved):
LGTM!

myungjoo (approved):

PR #4415, opened 2024-02-28
github-action/ubuntu: handle coredump from unit test

myungjoo:
If unit test generates core dump,
show the backtrace in the build log and
upload the coredump file.

@DonghakPark FYI: I thought you may want similar features in your scripts. You can download coredump and see the stack trace with these scripts.


.github/workflows/ubuntu_clean_meson_build.yml
```diff
@@ -37,9 +37,31 @@ jobs:
       run: |
         meson setup build/
         meson compile -C build/
+
+        sudo mkdir /cores
+        sudo chmod 777 /cores
```
wooksong, line 42:
```suggestion
        sudo mkdir -m 777 /cores
```

wooksong (approved):

jaeyun-jung (approved):

PR #4417, opened 2024-03-04
[CodeClean] set null after free

jaeyun-jung:
Clearly set null after releasing the name in tensor-info struct.


gichan-jang (approved):
LGTM

wooksong (approved):

PR #4421, opened 2024-03-08
decoder/direct_video: detect element size mismatch

myungjoo:
Return error with explicit error message (GST_ERROR) if
the element size mismatches.
  
For example, float32 tensor stream cannot be converted
to RGB video stream without tensor_transform.

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-directvideo.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-directvideo.c
@@ -22,8 +22,9 @@
  *
  * @see		https://github.com/nnstreamer/nnstreamer
  * @author	Jijoong Moon <jijoong.moon@samsung.com>
- * @bug		No known bugs except for NYI items
- *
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_decoder/tensordec-directvideo.c
```diff
@@ -174,6 +174,13 @@ dv_getOutCaps (void **pdata, const GstTensorsConfig * config)
   gint width, height, channel;
   GstCaps *caps;
 
+  if (config->info.info[0].type != _NNS_UINT8) {
```
anyj0527, line 177: direct_video also support `_NNS_UINT16`

myungjoo: fixed to take care of element size consistency only.

jaeyun-jung (approved):

anyj0527 (approved):
LGTM✔

PR #4427, opened 2024-03-25
[Refactoring] Split bounding_box struct

niley7464:
This patch splites properties so that bounding_box struct does not have properties for all cases.
Also some functions like `bb_setOptions`, `bb_decode` have too many if cases, so I replaced it to use pure virtual functions.

**Self evaluation:**


ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
```diff
@@ -0,0 +1,1817 @@
...
+    return setVideoSize (param);
+  } else if (opNum == 4) {
+    return setInputModelSize (param);
+  } else if (opNum == 5) {
+    is_track = (int) g_ascii_strtoll (param, NULL, 10);
+    return TRUE;
+  } else if (opNum == 6) {
+    do_log = (int) g_ascii_strtoll (param, NULL, 10);
+    return TRUE;
+  }
```
wooksong, line 886: If they have some meaning, how about making those ```opNum```value constants in the enum class (C++) or enum (C)?

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.h
```diff
@@ -0,0 +1,607 @@
...
+  guint max_detection;
+  guint total_labels;
+};
+
+
+class BoundingBox
+{
+  public:
+  BoundingBox ();
+  ~BoundingBox ();
```
myungjoo, line 411: You will need virtual destructors unless you can guarantee that subclasses don't have dynamically allocated properties.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.h
```diff
@@ -0,0 +1,607 @@
...
+#define MOBILENET_SSD_PARAMS_MAX 6
+
+#define DETECTION_THRESHOLD (.5f)
+#define THRESHOLD_IOU (.5f)
+#define Y_SCALE (10.0f)
+#define X_SCALE (10.0f)
+#define H_SCALE (5.0f)
+#define W_SCALE (5.0f)
+
+class MobilenetSSD : public BoxProperties
```
myungjoo, line 482: You do not need to declare subclasses in the header.
As long as we now can have separated subclasses, I recommend making separated files for subclasses and remove all info about each subclass in the header. (e.g., /ext/nnstreamer/tensor_decoder/boxproperties/mobilenetssd.cc)

Then, you can apply factory method in this base class that creates sub classes according to the property name. By properly designing this, you will not need the enums for each "subclass", too. You will need to think about how to accept "new" subclasses from the base class.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
niley7464: Create `BoundingBoxOption` enum class!
Thank you :)

myungjoo:
For macos CI fail, rebase it to main.

For static checkers fail, please take care of doxygen entries.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.h
myungjoo: The principle is to put commonly referenced classes and definitions in header.

If a class, #define, or a function is to be referenced by a single entity (e.g., by mobile-SSD only), you don't need to put it in a header.

niley7464: Moved subclasses to boxproperties subdir.
Thank you :)

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
```diff
@@ -0,0 +1,947 @@
...
+      case YOLOV5_BOUNDING_BOX:
+        bdata = new YoloV5 ();
+        break;
+      case YOLOV8_BOUNDING_BOX:
+        bdata = new YoloV8 ();
+        break;
+      case MP_PALM_DETECTION_BOUNDING_BOX:
+        bdata = new MpPalmDetection ();
+        break;
+      default:
```
myungjoo, line 727: This creates dependency from the base class to its derived classe.

You may let the derived classes register required data and function callbacks for this switch-case statements.

e.g.,
```
class derived_class : public boxproperty
{
  friend class registerer;
  class registerer {
    registerer() { /* register a function callback doing "return new derived_class();" */ }
  };

  static registerer register;
};
```
in the derived .cc file.

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.h
```diff
@@ -0,0 +1,430 @@
...
+  int setOptionInternal (const char *param);
+  int checkCompatible (const GstTensorsConfig *config);
+  GArray *decode (const GstTensorsConfig *config, const GstTensorMemory *input);
+
+  static const guint MAX_TENSORS = 4U;
+
+  private:
+  gint tensor_mapping[MAX_TENSORS]; /* Output tensor index mapping */
+  gfloat threshold; /* Detection threshold */
+};
```
myungjoo, line 345: With the previous comment satisfied (OCP), you should be able to kick all these derived classes out of this header.

niley7464: Moved derived class to subdir!

ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
niley7464: I misunderstood the previous comment because it was an unfamiliar concept.
Now base class do not need to be modifed when new derived class added or changed.

myungjoo (approved):
LGTM.

We still have points that break OCP, yet, but we can now proceed and fix further later.
(you can find such points by trying (mimicking) to implement another boxproperty subclass. if you need to touch tensordec-boundingbox.h/cc (you actually need at this point), it means you have more to refactor.

However, further refactoring can be in another PR later.

jaeyun-jung (approved):

jaeyun-jung:
@niley7464 I suggest using util function gst_tensors_info_get_nth_info() to get nth tensor-info ptr when iterating struct w/num_tensors. (config->info.info[nth])

myungjoo:
Let's move on and do the rest of refactoring in another PR.

PR #4437, opened 2024-04-12
[Query] util function using integer id

jaeyun-jung:
To prevent double free case, remove query handle and use integer id.
Also, add util function to hide edge handle in query elements and fix null ptr case.


anyj0527 (approved):

anyj0527 (approved):
LGTM👍 Great!

again4you (approved):
Great work!

PR #4454, opened 2024-05-02
[actions] Fix gh actions

anyj0527:
- Since Ubuntu 22.04 supports gbs from this year, let's switch to it.
- This patch makes the daily pdebuild workflow save the cache key with date postfix and outdate previous cache. Then the pdebuild.yml workflow can use the fresh cache.

**Self evaluation:**


myungjoo (approved):

again4you (approved):
Good job!

PR #4461, opened 2024-05-16
[edgesrc] Add checking caps before getting tensor size

songgot:
- Prevent Gstreamer-CRITICAL message when caps is null


jaeyun-jung (approved):

niley7464 (approved):
LGTM!

PR #4482, opened 2024-06-06
Feature/upgrade tensorrt to 10

bveldhoen:
---
# PR Description

Updated tensorrt to version 10.0.1, including the required changes to the tensor_filter subplugin for tensorrt.

- Using onnx parser instead of uff parser.
- Upgrade to tensorrt 10.

See also:
https://github.com/nnstreamer/nnstreamer-example/pull/338

**Self evaluation:**

**How to evaluate:**
1. Run the example in the nnstreamer-example.


ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -670,13 +670,15 @@ if tensorrt_support_is_available
 
   shared_library('nnstreamer_filter_tensorrt',
     filter_sub_tensorrt_sources,
+    link_args: '-L/usr/local/cuda-12.1/targets/x86_64-linux/lib/stubs',
```
myungjoo: Nope. You cannot and should not do this. You cannot guarantee that everyone will install tensorrt binraies to this hardcoded directory (think of other distro of Linux users or even.. Arm/RISC-V users).

Library linking path should be determined dynamically at build-time, usually by pkg-config or library-finding mechanism of meson. Refer to other library finding codes in meson.build files.

tests/nnstreamer_example/meson.build
```diff
@@ -95,11 +95,11 @@ library('nnscustom_framecounter',
...
 
-if tensorrt_support_is_available
-  library('nnstreamer_customfilter_tensorrt_reshape',
-    join_paths('custom_example_tensorrt', 'nnstreamer_customfilter_example_tensorrt_reshape.cc'),
-    dependencies: [glib_dep, gst_dep, nnstreamer_dep, tensorrt_support_deps],
-    install: get_option('install-test'),
-    install_dir: customfilter_install_dir
-  )
-endif
+#if tensorrt_support_is_available
```
myungjoo, line 98: Do not comment out in git. Git leaves code history.

If your changes for CUDA 12 or CUDA 12 itself is not compatible with the current unit tests, please do 
1. Update the unit test compatible for older-CUDA and CUDA-12
OR
2. Disable unit testing only for CUDA-12.

meson.build
```diff
@@ -286,20 +287,20 @@ if not get_option('tensorrt-support').disabled()
...
 
   nvinfer_lib = cxx.find_library('nvinfer', required: false)
-  if nvinfer_lib.found() and cxx.check_header('NvInfer.h')
+  if nvinfer_lib.found() #and cxx.check_header('NvInfer.h')
     nvinfer_dep = declare_dependency(dependencies: nvinfer_lib)
   endif
 
   nvparsers_lib = cxx.find_library('nvparsers', required: false)
-  if nvparsers_lib.found() and cxx.check_header('NvUffParser.h')
+  if nvparsers_lib.found() #and cxx.check_header('NvUffParser.h')
```
myungjoo: Please do not comment it out. Remove or leave it.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -157,8 +157,8 @@ typedef struct _GstTensorFilterFrameworkInfo
 {
   const char *name; /**< Name of the neural network framework, searchable by FRAMEWORK property. Subplugin is supposed to allocate/deallocate. */
   int allow_in_place; /**< TRUE(nonzero) if in-place transfer of input-to-output is allowed. Not supported in main, yet. */
-  int allocate_in_invoke; /**< TRUE(nonzero) if invoke_NN is going to allocate output ptr by itself and return the address via output ptr. Do not change this value after cap negotiation is complete (or the stream has been started). */
-  int run_without_model; /**< TRUE(nonzero) when the neural network framework does not need a model file. Tensor-filter will run invoke_NN without model. */
+  int allocate_in_invoke; /**< TRUE(nonzero) if invoke[_NN] is going to allocate output ptr by itself and return the address via output ptr. Do not change this value after cap negotiation is complete (or the stream has been started). */
+  int run_without_model; /**< TRUE(nonzero) when the neural network framework does not need a model file. Tensor-filter will run invoke[_NN] without model. */
```
myungjoo, line 161: Why have you changed this?

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -670,13 +670,15 @@ if tensorrt_support_is_available
...
     filter_sub_tensorrt_sources,
+    #link_args: '-L/usr/local/cuda-12.1/targets/x86_64-linux/lib/stubs',
     dependencies: nnstreamer_filter_tensorrt_deps,
     install: true,
     install_dir: filter_subplugin_install_dir
   )
 
   static_library('nnstreamer_filter_tensorrt',
     filter_sub_tensorrt_sources,
+    #link_args: '-L/usr/local/cuda-12.1/targets/x86_64-linux/lib/stubs',
```
myungjoo, line 681: Please don't add these commented hardcoded-paths. You can add library-paths when you declare corresponding deps: https://mesonbuild.com/Dependencies.html

tools/docker/Dockerfile
```diff
@@ -99,4 +99,4 @@ RUN if [ ! -n "${MIN_RUN_ENV}" ]; then \
     fi && \
     rm -rf /var/lib/apt/lists/* /root/*.deb
 USER ${USERNAME}
-WORKDIR /home/${USERNAME}
+WORKDIR /home/${USERNAME}
```
myungjoo, line 102: uh....

you may need to clean up commits..

meson.build
```diff
@@ -267,12 +267,13 @@ endif
 
 # tensorrt
 nvinfer_dep = dependency('', required: false)
-nvparsers_dep = dependency('', required: false)
+# nvparsers_dep = dependency('', required: false)
```
myungjoo, line 270: please create nvonnxparser_dep empty dependency here for non-nvidia users. (and don't comment out removed variables)

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -403,74 +726,96 @@ tensorrt_subplugin::allocBuffer (void **buffer, gsize size)
...
+  //   _NNS_UINT16,
+  //   _NNS_INT8,
+  //   _NNS_UINT8,
+  //   _NNS_FLOAT64,
+  //   _NNS_FLOAT32,
+  //   _NNS_INT64,
+  //   _NNS_UINT64,
+  //   _NNS_FLOAT16, /**< added with nnstreamer 2.1.1-devel. If you add any operators (e.g., tensor_transform) to float16, it will either be not supported or be too inefficient. */
+  //   _NNS_END,
+  // } tensor_type;
```
myungjoo, line 815: Please remove useless comments.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
bveldhoen: I believe that the function, to which is referred, is called 'invoke_NN' in GST_TENSOR_FILTER_FRAMEWORK_V0, and 'invoke' GST_TENSOR_FILTER_FRAMEWORK_V1. I can revert, as this is not relevant to this PR.

tests/nnstreamer_example/meson.build
bveldhoen: This PR contains breaking changes w.r.t. existing tensorrt implementations; existing pipeline configurations still using the uffparser and the uff model format will not work anymore. Next to that, TensorRT version 10 is only supported with cuda versions 11.0 and up (earlier cuda versions are not supported for TensorRT 10).

Would it make sense to introduce a separate build option for TensorRT 10? I.e. tensorrt10-support? And treat it as a separate implementation, so that existing implementations and tests (for tensorrt <10) are unaffected?

.gitignore
```diff
@@ -4,6 +4,13 @@
 # ignore cmake build directory
 /build
 
+# ignore generated documentation directories
+/html
+/latex
+
+# ignore docker artifact directory
+/tools/docker/deps
+
```
bveldhoen, line 13: This is related to the Dockerfile.tensorrt (see below), which I used for local testing. I can remove this if that docker file is not supposed to be part of this PR?

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -613,7 +613,7 @@ _gst_tensor_filter_transform_validate (GstBaseTransform * trans,
     return GST_FLOW_ERROR;
   }
 
-  silent_debug (self, "Invoking %s with %s model\n", priv->fw->name,
```
bveldhoen, line 616: This caused logging of uninitialized memory.

meson.build
```diff
@@ -286,21 +287,21 @@ if not get_option('tensorrt-support').disabled()
...
+  if nvinfer_lib.found() and cxx.has_header('NvInfer.h')
     nvinfer_dep = declare_dependency(dependencies: nvinfer_lib)
   endif
 
-  nvparsers_lib = cxx.find_library('nvparsers', required: false)
-  if nvparsers_lib.found() and cxx.check_header('NvUffParser.h')
-    nvparsers_dep = declare_dependency(dependencies: nvparsers_lib)
+  nvonnxparser_lib = cxx.find_library('nvonnxparser', required: false)
+  if nvonnxparser_lib.found() and cxx.has_header('NvOnnxParser.h')
+    nvonnxparser_dep = declare_dependency(dependencies: nvonnxparser_lib)
```
bveldhoen, line 304: NvInferVersion.h contains the version of tensorrt:

```
...
#define NV_TENSORRT_MAJOR 10 //!< TensorRT major version.
...
```
Perhaps I should use that to determine the version of tensorrt?

bveldhoen:
I've made the following changes:
* Clean up unnecessary/unrelated changes (I can clean up further by removing all unrelated changes if desired).
* There should be no changes when compiling against tensorrt version < 10, so this should be backwards compatible (not tested).
* Reverted tensor_filter_tensorrt and added separate file for tensorrt v10 (tensor_filter_tensorrt_v10), following similar approach as for tensor_filter_snpe[_v1].
* Subsequent changes in build files to support the additional tensorrt implementation.
* Replaced error return codes with exceptions.

meson.build
bveldhoen: Done

ext/nnstreamer/tensor_filter/meson.build
bveldhoen: Done

tools/docker/Dockerfile
bveldhoen: Done

meson.build
bveldhoen: Done

meson.build
bveldhoen: Done, see most recent commit.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
bveldhoen: Done

tests/nnstreamer_example/meson.build
bveldhoen: Done, see most recent commit.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt_v10.cc
```diff
@@ -0,0 +1,758 @@
...
+  std::size_t getVolume (const nvinfer1::Dims& shape) const;
+  std::size_t getTensorRTDataTypeSize (nvinfer1::DataType tensorrt_data_type) const;
+  tensor_type getNnStreamerDataType(nvinfer1::DataType tensorrt_data_type) const;
+
+  void constructNetwork(NvInferUniquePtr<nvonnxparser::IParser>& parser) const;
+  void buildSaveEngine () const;
+  void loadEngine ();
+};
+
+const char *tensorrt_subplugin::name = "tensorrt";
```
myungjoo, line 164: Please make the name and its binary file name distinguishable against older tensorrt. (e.g., tensorrt10)
And, let the subplugin check and abort if the linked tensorrt version is incomaptible during the init.

We need a way to identify whether this is for tensorrt+10 or not after they are compiled and deployed although we can deprecate the older one afterwards.

myungjoo:
After allowing the tensorrt 10+ subplugin be deployed along with older tensorrt subplugin and allowing its binaries be distinguishable against older tensorrt subplugin (by naming it differently), please do

1. satisfy the github-action checkers (coding style and commit signing)
2. remove unnecessary or unrelated changes
3. squashing commits to remove unnecessary commits.
    - e.g., commit-1: new tensorrt subplugin, commit-2: docker updates, commit-3: test updates

bveldhoen:
losed in favor of:
https://github.com/nnstreamer/nnstreamer/pull/4489

PR #4487, opened 2024-06-13
[filter] Support snpe filter in android

anyj0527:
- Support snpe (SDK v2) in Android.
  Some additional env variables are requires to be set.
  REF: https://docs.qualcomm.com/bundle/publicresource/topics/80-63442-2/dsp_runtime.html

**Self evaluation:**

### DIFF
--- a/jni/nnstreamer.mk
+++ b/jni/nnstreamer.mk
@@ -109,6 +109,12 @@ NNSTREAMER_FILTER_CAFFE2_SRCS := \
 NNSTREAMER_FILTER_SNPE_SRCS := \
     $(NNSTREAMER_EXT_HOME)/tensor_filter/tensor_filter_snpe.cc
 
+# Check whether the version of SNPE SDK is old (v1, rather than v2)
+ifeq ($(shell test -d ${SNPE_ROOT}/include/zdl; echo $$?),0)
+NNSTREAMER_FILTER_SNPE_SRCS := \
+    $(NNSTREAMER_EXT_HOME)/tensor_filter/tensor_filter_snpe_v1.cc
+endif
+
 # filter snap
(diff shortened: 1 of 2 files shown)


wooksong (approved):

jni/nnstreamer.mk
```diff
@@ -109,6 +109,12 @@ NNSTREAMER_FILTER_CAFFE2_SRCS := \
 NNSTREAMER_FILTER_SNPE_SRCS := \
     $(NNSTREAMER_EXT_HOME)/tensor_filter/tensor_filter_snpe.cc
 
+# Check whether the version of SNPE SDK is old (v1, rather than v2)
+ifeq ($(shell test -d ${SNPE_ROOT}/include/zdl; echo $$?),0)
+NNSTREAMER_FILTER_SNPE_SRCS := \
+    $(NNSTREAMER_EXT_HOME)/tensor_filter/tensor_filter_snpe_v1.cc
+endif
```
myungjoo, line 116: Isn't this too dangerous assumption?

I'd recommend:

- build snpe(v2) if v2 is available
- build snpe(v1) if v1 is available AND v2 is not available.

The current logic is

- build snpe(v2) if v2 is available AND v1 is NOT available
- build snpe(v1) if v1 is available AND.

anyj0527: I assume $SNPE_ROOT is unique and a single SNPE SDK does not have both header directories for V1 and V2. I think there is no possibility to both v1 and v2 available.

myungjoo: ok.. if ```!exist(zdl)``` means !v2, let's proceed.

myungjoo (approved):

PR #4490, opened 2024-06-17
[Filter/DeepViewRT] mem leak case

jaeyun-jung:
Fix mem leak case when loading mapped model file, use member variable.


myungjoo (approved):

myungjoo:
CC: @jvuillaumier

PR #4491, opened 2024-06-18
[Query] case when change pipe state

jaeyun-jung:
Set caps in edge handle before calling change-state callback.

### DIFF
--- a/gst/nnstreamer/tensor_query/tensor_query_serversink.c
+++ b/gst/nnstreamer/tensor_query/tensor_query_serversink.c
@@ -181,6 +181,10 @@ gst_tensor_query_serversink_change_state (GstElement * element,
         nns_loge ("Failed to change state from PAUSED to PLAYING.");
         return GST_STATE_CHANGE_FAILURE;
       }
+
+      caps = gst_pad_peer_query_caps (GST_BASE_SINK_PAD (bsink), NULL);
+      gst_tensor_query_serversink_set_caps (bsink, caps);
+      gst_caps_unref (caps);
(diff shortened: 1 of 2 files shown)


niley7464 (approved):
LGTM!
Thank you :)

gst/nnstreamer/tensor_query/tensor_query_serversink.c
```diff
@@ -202,10 +206,6 @@ gst_tensor_query_serversink_change_state (GstElement * element,
     case GST_STATE_CHANGE_PLAYING_TO_PAUSED:
       gst_tensor_query_server_release_edge_handle (sink->sink_id);
       break;
-    case GST_STATE_CHANGE_PAUSED_TO_PLAYING:
-      caps = gst_pad_peer_query_caps (GST_BASE_SINK_PAD (bsink), NULL);
-      gst_tensor_query_serversink_set_caps(bsink, caps);
-      break;
     default:
       break;
```
djeong20, line 210: It seems like now the switch statement has only one case. maybe it can use an if statement?

```
if (transition == GST_STATE_CHANGE_PLAYING_TO_PAUSED) {
   gst_tensor_query_server_release_edge_handle (sink->sink_id);
}
```

jaeyun-jung: possible, however for future work, use switch-case (guide from gst sample, I am not sure but if adding another state change is required later, better remaining current code.)

anyj0527 (approved):

PR #4494, opened 2024-06-19
[tensor_trainer] Add a function that generates dummy data

songgot:
When the pipeline stops and no data is input, the necessary data for the current epoch is generated for the normal termination of nntrainer. The data generation callback of nntrainer is no longer called until all the necessary data for the current epoch are received.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_trainer.c
+++ b/gst/nnstreamer/elements/gsttensor_trainer.c
@@ -307,6 +307,7 @@ gst_tensor_trainer_init (GstTensorTrainer * trainer)
   trainer->is_training_complete = FALSE;
   trainer->is_epoch_complete = FALSE;
   trainer->cur_epoch_data_cnt = 0;
+  trainer->required_sample = 0;
 
   gst_tensors_config_init (&trainer->in_config);
   gst_tensors_config_init (&trainer->out_config);
@@ -343,6 +344,11 @@ gst_tensor_trainer_finalize (GObject * object)
   g_mutex_clear (&trainer->training_completion_lock);
   g_cond_clear (&trainer->epoch_completion_cond);
   g_mutex_clear (&trainer->epoch_completion_lock);
+
+  if (trainer->dummy_data_thread) {
+    g_thread_join (trainer->dummy_data_thread);
+    trainer->dummy_data_thread = NULL;
+  }
 
   if (trainer->fw_created && trainer->fw) {
     trainer->fw->destroy (trainer->fw, &trainer->prop, &trainer->privateData);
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +493,50 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+  do {
+    GST_INFO_OBJECT (trainer, "cur_epoch_data_cnt=%u",
+        trainer->cur_epoch_data_cnt);
+    GST_INFO_OBJECT (trainer, "num_tensors=%d",
+        trainer->prop.input_meta.num_tensors);
+
+    for (i = 0; i < trainer->output_meta.num_tensors; i++) {
+      dummy_data = g_malloc (trainer->push_tensors[i].size);
+      memset (dummy_data, 1, trainer->push_tensors[i].size);
+      trainer->push_tensors[i].data = dummy_data;
```
songgot, line 521: memory leak.

gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +494,47 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+
+  do {
+    GST_INFO_OBJECT (trainer, "cur_epoch_data_cnt=%u",
+        trainer->cur_epoch_data_cnt);
+    GST_INFO_OBJECT (trainer, "num_tensors=%d",
+        trainer->prop.input_meta.num_tensors);
+
+    for (i = 0; i < trainer->output_meta.num_tensors; i++) {
+      dummy_data[i] = g_malloc (trainer->push_tensors[i].size);
+      memset (dummy_data[i], 1, trainer->push_tensors[i].size);
```
gichan-jang, line 518: Q> Is there any reason to set 1?

gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +494,47 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+      trainer->push_tensors[i].data = dummy_data[i];
+    }
+
+    ret =
+        trainer->fw->push_data (trainer->fw, &trainer->prop,
+        trainer->privateData, trainer->push_tensors);
+    if (ret < 0) {
+      GST_ERROR_OBJECT (trainer, "Failed to push dummy data");
```
gichan-jang, line 526: In case of failure of pushing the data, is it okay to increase `trainer->cur_epoch_data_cnt`?

gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +494,47 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+    }
+
+    for (i = 0; i < trainer->output_meta.num_tensors; i++)
+      g_free (dummy_data[i]);
+    trainer->cur_epoch_data_cnt++;
+  } while (trainer->required_sample > trainer->cur_epoch_data_cnt);
```
gichan-jang, line 532: Don't need to check `trainer->is_training_complete`?

gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +494,47 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+
+  return NULL;
```
djeong20, line 534: quick question! if this function always returns NULL, why can't it be a void function? should GThreadFunc always return some value?

jaeyun-jung: The callback of g_thread_new() - thread function (GThreadFunc) requires return ptr (void *).

gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -488,6 +494,47 @@ gst_tensor_trainer_check_invalid_param (GstTensorTrainer * trainer)
...
+
+  gst_tensor_trainer_stop_model_training (trainer);
+
+  do {
+    GST_INFO_OBJECT (trainer, "cur_epoch_data_cnt=%u",
+        trainer->cur_epoch_data_cnt);
+    GST_INFO_OBJECT (trainer, "num_tensors=%d",
+        trainer->prop.input_meta.num_tensors);
+
+    for (i = 0; i < trainer->output_meta.num_tensors; i++) {
```
jaeyun-jung, line 516: suggestion: is it possible to allocate required memory and release it outside of do {} while () block?
IMO it does not necessary allocating memories at every push_data callback (The param of memories is const).

gst/nnstreamer/elements/gsttensor_trainer.c
songgot: No, I thought it would be better than 0. :)

gst/nnstreamer/elements/gsttensor_trainer.c
songgot: Thank you.

gst/nnstreamer/elements/gsttensor_trainer.c
songgot: In fact, data push is always guaranteed to succeed, and there is no need to check the end of learning here. If it doesn't end, there is no way to terminate it anymore.

gst/nnstreamer/elements/gsttensor_trainer.c
songgot: okay, I will modify it.

jaeyun-jung (approved):

myungjoo:
PING: @djeong20 @gichan-jang

djeong20 (approved):
LGTM

PR #4501, opened 2024-06-24
[Converter] max dim value

jaeyun-jung:
Code clean, check max dim value from buffer size.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_converter.c
+++ b/gst/nnstreamer/elements/gsttensor_converter.c
@@ -1158,6 +1158,7 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     case _NNS_OCTET:
       if (gst_tensors_config_is_flexible (config)) {
         /* update dimension with buffer size */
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/elements/gsttensor_converter.c
```diff
@@ -1158,6 +1158,7 @@ gst_tensor_converter_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     case _NNS_OCTET:
       if (gst_tensors_config_is_flexible (config)) {
         /* update dimension with buffer size */
+        g_assert (buf_size <= G_MAXUINT32);
```
myungjoo, line 1161: buf_size is gsize.
```
typedef unsigned int gsize;
```

This looks meaningless.

jaeyun-jung: yes, but svace tool detects size as long :(

myungjoo: Then, please mark it "false positive" and report SE.

PR #4502, opened 2024-06-25
[filter] Support TFLite QNN Delegate

anyj0527:
- Import QNN delegate header.
- For Android, set certain environment variables for DSP/HTP runtimes.
- Add custom option for QNN Delegate.
- Add custom option for QNN backend and performance mode.
- User can use this in pipeline:
`... custom=Delegate:QNN,QNNBackend:DSP,QNNPerformanceMode:highperformance ...`

Disclaimer: This only be tested in Android.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -125,9 +125,29 @@ typedef enum {
   TFLITE_DELEGATE_NNAPI,
   TFLITE_DELEGATE_XNNPACK,
   TFLITE_DELEGATE_EXTERNAL,
+  TFLITE_DELEGATE_QNN,
 
   TFLITE_DELEGATE_MAX
 } tflite_delegate_e;
+
+/**
+ * @brief Internal enum for QNN backend type.
+ */
+typedef enum QNNBackendType {
+  QNN_BACKEND_UNDEFINED = 0,
+  QNN_BACKEND_GPU,
+  QNN_BACKEND_HTP,
+  QNN_BACKEND_DSP
+} QNNBackendType;
+
+/**
+ * @brief Internal enum for QNN Performance mode.
+ */
+typedef enum QNNPerformanceMode {
+  QNN_PERFMODE_Default = 0,
+  QNN_PERFMODE_HighPerf,
+  QNN_PERFMODE_PowerSaver
+} QNNPerformanceMode;
 
 /**
  * @brief Option to open tf-lite model.
@@ -139,6 +159,8 @@ typedef struct {
   gint num_threads; /**< the number of threads */
   const gchar *ext_delegate_path; /**< path to external delegate lib */
   GHashTable *ext_delegate_kv_table; /**< external delegate key values options */
+  QNNBackendType qnn_backend_type; /**< QNN Delegate backend type */
+  QNNPerformanceMode qnn_performance_mode; /**< QNN Delegate performance mode */
 } tflite_option_s;
 
 /**
@@ -228,6 +250,8 @@ class TFLiteInterpreter
   bool is_xnnpack_delegated; /**< To check if XNNPACK delegate is used */
   char *ext_delegate_path; /**< path to external delegate lib */
   GHashTable *ext_delegate_kv_table; /**< external delegate key values options */
+  QNNBackendType qnn_backend_type; /**< QNN Delegate backend type */
+  QNNPerformanceMode qnn_performance_mode; /**< QNN Delegate performance mode */
 
   std::unique_ptr<tflite::Interpreter> interpreter;
   std::unique_ptr<tflite::FlatBufferModel> model;
@@ -242,6 +266,7 @@ class TFLiteInterpreter
   int setTensorProp (const std::vector<int> &tensor_idx_list, GstTensorsInfo *tensorMeta);
 
   tflite::Interpreter::TfLiteDelegatePtr delegate_ptr; /**< single delegate supported */
+  friend class TFLiteCore;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -1656,6 +1779,134 @@ static GstTensorFilterFramework NNS_support_tensorflow_lite
...
+ */
+void
+init_filter_tflite (JNIEnv *env, jobject context)
+{
+  if (!_tflite_set_env (env, context)) {
+    nns_loge ("Failed to set extra env");
+    return;
+  }
+
+  nnstreamer_filter_probe (&NNS_support_tensorflow_lite);
```
jaeyun-jung, line 1899: recommend adding internal static function to prevent duplicated code in init_filter_tflite() 1914~

myungjoo (commented):
How are you going to enable ```TFLITE_QNN_DELEGATE_SUPPORTED```?
What's the dependency relation between ```TFLITE_QNN_DELEGATE_SUPPORTED``` and ```__ANDROID__```?

According to LINE 1900-1907, it appears that QNN is to be enabled only in Android.. but we already have a Tizen device with QNN...

anyj0527:
> How are you going to enable `TFLITE_QNN_DELEGATE_SUPPORTED`? What's the dependency relation between `TFLITE_QNN_DELEGATE_SUPPORTED` and `__ANDROID__`?
> 
> According to LINE 1900-1907, it appears that QNN is to be enabled only in Android.. but we already have a Tizen device with QNN...

Currently there is no tizen device available for QNN or QNN delegate. I'll build/test this feature on Tizen later.
`TFLITE_QNN_DELEGATE_SUPPORTED` is enabled by Android build script. Please check https://github.com/nnstreamer/api/pull/544

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -117,10 +125,30 @@ typedef enum {
...
+ */
+typedef enum QNNBackendType {
+  QNN_BACKEND_UNDEFINED = 0,
+  QNN_BACKEND_GPU,
+  QNN_BACKEND_HTP,
+  QNN_BACKEND_DSP
+} QNNBackendType;
+
+/**
+ * @brief Internal enum for QNN erformance mode.
```
SeoHyungjun, line 144: It looks like a typo.
```suggestion
 * @brief Internal enum for QNN performance mode.
```

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -1656,6 +1779,134 @@ static GstTensorFilterFramework NNS_support_tensorflow_lite
               .allocateInInvoke = nullptr,
           } } };
 
+#if defined(__ANDROID__)
+/**
+ * @brief Set additional environment (ADSP_LIBRARY_PATH) for tflite
+ */
+static gboolean
+_tflite_set_env (JNIEnv *env, jobject context)
```
myungjoo, line 1787: Name this ```_tflite_set_env_android``` for better readability.
Then, separate conditions for "QNN" and "ANDROID". Don't use the "__ANDROID__" as if that is same with "QNN". You should be able to build Android without QNN without touching any .cc or .h codes.

anyj0527: I added a function `_tflite_set_env_qnn_android`, PTAL

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
anyj0527: Thanks! I refactored the code, PTAL

myungjoo (commented):
If this file is built with ANDROID=no && QNN_DELEGATE=yes, does QNN delegation work? or it does not?

If it does not, we need explicit error at built-time with
```
#if !ANDROID && QNN
#error the error message for developers/integrators...
#endif
```

I'm not sure if OS/BSP support is ready, we have non-Android devices that might support QNN. So, until they are ready, let's emit explicit errors.

anyj0527:
> If this file is built with ANDROID=no && QNN_DELEGATE=yes, does QNN delegation work? or it does not?
> 
> If it does not, we need explicit error at built-time with
> 
> ```
> #if !ANDROID && QNN
> #error the error message for developers/integrators...
> #endif
> ```
> 
> I'm not sure if OS/BSP support is ready, we have non-Android devices that might support QNN. So, until they are ready, let's emit explicit errors.

One could manually build and test it on non-Android. The explicit build error seems no need.

myungjoo:
> > If this file is built with ANDROID=no && QNN_DELEGATE=yes, does QNN delegation work? or it does not?
> > If it does not, we need explicit error at built-time with
> > ```
> > #if !ANDROID && QNN
> > #error the error message for developers/integrators...
> > #endif
> > ```
> >     
> >   
> > I'm not sure if OS/BSP support is ready, we have non-Android devices that might support QNN. So, until they are ready, let's emit explicit errors.
> 
> One could manually build and test it on non-Android. The explicit build error seems no need.

Then. the answer for 

> If this file is built with ANDROID=no && QNN_DELEGATE=yes, does QNN delegation work? or it does not?

is "it may work. there is no missing parts for non-Android as far as you know"?

The concerned case is "there is known missing parts for non-Android", guessed by the big Android-only function, "_tflite_set_env_qnn_android" (the current code incurs the question: do we need something like "_tflite_set_env_qnn_linux"?)

anyj0527:
@myungjoo In `filter_snpe.cc` there is same code for only-android-init. If the target is prepared properly, qnn delegate should work like Tizen-snpe.

myungjoo:
> @myungjoo In `filter_snpe.cc` there is same code for only-android-init. If the target is prepared properly, qnn delegate should work like Tizen-snpe.

Ok. Now, my concerns are resolved. Thanks

myungjoo (approved):

jaeyun-jung (approved):

PR #4514, opened 2024-07-08
[Filter] Let ExecuTorch filter use `Module` api

anyj0527:
- If executorch.pc set link option as something like `-Wl,--whole-archive -lexecutorch  -lportable_ops_lib -Wl,--no-whole-archive`, 
  no manual cpp file include is needed. Let's remove those lines.

- To simplify verbose codes of the lower-level ExecuTorch API, Let it use `Module` APIs
  REF: https://pytorch.org/executorch/stable/extension-module.html

- Add SSAT test for executorch filter.
  Used models are converted to executorch model from those of pytorch ssat test

TODO:
Remove memcpy of result tensors using [a recent api](https://github.com/pytorch/executorch/blob/5584b9e3c865edca239ec5df6346f1d1aabb0276/extension/module/module.h#L205)

**Self evaluation:**


myungjoo (approved):

ext/nnstreamer/tensor_filter/tensor_filter_executorch.cc
```diff
@@ -141,7 +120,9 @@ executorch_subplugin::getEmptyInstance ()
 void
 executorch_subplugin::cleanup ()
 {
-  g_free (model_path);
+  if (model_path)
```
jaeyun-jung, line 123: unnecessary cond? (g_free() internally checks null case.)

ext/nnstreamer/tensor_filter/tensor_filter_executorch.cc
```diff
@@ -177,61 +150,69 @@ executorch_subplugin::configure_instance (const GstTensorFilterProperties *prop)
...
+          return _NNS_END;
+      }
+    };
+
+    // parse input tensors info
+    size_t num_inputs = forward_method_meta->num_inputs ();
+    inputInfo.num_tensors = num_inputs;
+    for (size_t i = 0; i < num_inputs; ++i) {
+      const auto input_meta = forward_method_meta->input_tensor_meta (i);
+
```
jaeyun-jung, line 180: let's use gst_tensors_info_get_nth_info() to get nth tensor-info ptr, to prevent index error.

ext/nnstreamer/tensor_filter/tensor_filter_executorch.cc
```diff
@@ -177,61 +150,69 @@ executorch_subplugin::configure_instance (const GstTensorFilterProperties *prop)
...
-        program->get ().load_method (method_name, memory_manager));
-    ET_CHECK_MSG (method->ok (), "Loading of method %s failed with status 0x%" PRIx32,
-        method_name, (uint32_t) method->error ());
-    ET_LOG (Info, "Method loaded.");
+    // parse output tensors info
+    size_t num_outputs = forward_method_meta->num_outputs ();
+    outputInfo.num_tensors = num_outputs;
+    for (size_t i = 0; i < num_outputs; ++i) {
+      const auto output_meta = forward_method_meta->output_tensor_meta (i);
+
```
jaeyun-jung, line 203: ditto use gst_tensors_info_get_nth_info() to prevent index error.

ext/nnstreamer/tensor_filter/tensor_filter_executorch.cc
anyj0527: Oops I missed that, I updated the code PTAL

ext/nnstreamer/tensor_filter/tensor_filter_executorch.cc
anyj0527: Updated it, PTAL

jaeyun-jung (approved):

PR #4517, opened 2024-07-12
tizen: allow tensorrt

myungjoo:
Tizen is going to accept TensorRT and NVidia GPUs in the build system. Let's allow to enable TensorRT.

However, meson is not configured to force tensorrt, yet, becuase we do not know whether Tizen is going to port TensorRT version >=10 or not.


SeoHyungjun (approved):
LGTM

jaeyun-jung (approved):

PR #4526, opened 2024-07-29
[coverity] fix coverity issues

niley7464:
This patch fixes some coverity issues.
- WID 10976414 : The pointer returned by operator new can't be null.
- WID 10976530 : Constructor may not initialize class members.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -331,6 +331,7 @@ TFLiteInterpreter::TFLiteInterpreter ()
   ext_delegate_path = nullptr;
   ext_delegate_kv_table = nullptr;
   qnn_backend_type = QNN_BACKEND_UNDEFINED;
+  qnn_performance_mode = QNN_PERFMODE_Default;
 
   g_mutex_init (&mutex);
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -966,6 +967,9 @@ TFLiteCore::TFLiteCore (const GstTensorFilterProperties *prop)
       interpreter = new TFLiteInterpreter ();
   } else
     interpreter = new TFLiteInterpreter ();
+
+  if (interpreter == NULL)
```
anyj0527, line 971: Another comparing NULL with result of `new` here?

niley7464: This condition is to check if interpreter is not set when `checkSharedInterpreter (prop)` is true (line 966)!
If interpreter is `NULL` after `new TfliteCore()`, init function will create a seg fault so I added this compare logic.

anyj0527: Oh that's right. Thanks!

anyj0527 (approved):
LGTM👍

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -966,6 +967,9 @@ TFLiteCore::TFLiteCore (const GstTensorFilterProperties *prop)
...
+    nns_logd ("Failed to allocate memory for interpreter");
```
gichan-jang, line 972: How about using try-catch statement to check memory allocation?
```
  try {
    ...
  } catch (std::bad_alloc &exception) {
    ...
  }
```

gichan-jang (approved):
LGTM

PR #4536, opened 2024-08-08
[filter] Add custom prop `OutputTensor`, `OutputType`, and `InputType` for snpe 

anyj0527:
[prop `OutputTensor`]
- Let snpe subplugin set output tensor names provided by `OutputTensor`
  custom property.
- Each name should be sepearted by `;` (if OutputTensor is plural)
- Example usage:
  ... model=model.dlc custom=OutputTensor:"output_0;output_1" ...

[prop `InputType` and `OutputType`]
- Let snpe subplugin set types for input/output tensors.
- The value should be one of {"FLOAT32", "TF8"}
- Each value should be seperated by `;` (if plural)
- The order should follow the order of input/output tensors in the model.
- Example usage:
  ... custom=InputType:TF8,OutputType:"FLOAT32,FLOAT32" ...
- If type is not specified, use default type of the given model.
  - quanitzed model can use both float32 and tf8,
    non-quantized model can only use float32.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
@@ -274,7 +274,7 @@ snpe_subplugin::configure_instance (const GstTensorFilterProperties *prop)
     ret = Snpe_UserBufferMap_Add (bufferMapHandle, tensorName, iub);
   };
 
-  auto parse_custom_prop = [&runtime] (const char *custom_prop) {
+  auto parse_custom_prop = [&runtime, &outputstrListHandle] (const char *custom_prop) {
     if (!custom_prop)
       return;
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -300,6 +319,48 @@ snpe_subplugin::configure_instance (const GstTensorFilterProperties *prop)
...
+              nns_loge ("Given tensor name with index %u is invalid.", i);
+            }
+            nns_logi ("Add output tensor name of %s", names[i]);
+            if (Snpe_StringList_Append (outputstrListHandle, names[i]) != SNPE_SUCCESS) {
+              nns_loge ("Failed to append output tensor name. %s", names[i]);
+            }
+          }
+          g_free (_ot_str);
+          g_strfreev (names);
+        } else if (g_ascii_strcasecmp (option[0], "OutputType") == 0) {
```
jaeyun-jung, line 339: Q: with snpe filter for v1, is it possible using tensor-filter property input/inputtype/inputname and output/outputtype/outputname?

anyj0527: Well I don't think so. There is no API changing Input tensors. And the "TF8" type is should be distinguished from "UINT8" type (NYI).

jaeyun-jung (approved):
plz fix format error.

anyj0527:
> plz fix format error.

Thanks, I fixed the format error.

ext/nnstreamer/tensor_filter/tensor_filter_snpe.cc
```diff
@@ -300,6 +320,49 @@ snpe_subplugin::configure_instance (const GstTensorFilterProperties *prop)
...
           }
+        } else if (g_ascii_strcasecmp (option[0], "OutputTensor") == 0) {
+          /* the tensor name may contain ':' */
+          gchar *_ot_str = g_strjoinv (":", &option[1]);
+          gchar **names = g_strsplit (_ot_str, ";", -1);
+          guint num_names = g_strv_length (names);
+          outputstrListHandle = Snpe_StringList_Create ();
+          for (guint i = 0; i < num_names; ++i) {
+            if (g_strcmp0 (names[i], "") == 0) {
+              nns_loge ("Given tensor name with index %u is invalid.", i);
```
wooksong, line 331: Doesn't need any ```break`` or ```return``` here?

anyj0527: Thanks! I make it throws exception on error case. PTAL

wooksong (approved):
LGTM!

again4you (approved):
LGTM!

PR #4541, opened 2024-08-09
[Element] invalid stream id

jaeyun-jung:
Prevent invalid stream-id case if element has multiple sink pads.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_split.c
+++ b/gst/nnstreamer/elements/gsttensor_split.c
@@ -276,6 +276,8 @@ static GstTensorPad *
 gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
     gboolean * created, guint nth)
 {
+  GstElement *element = GST_ELEMENT_CAST (split);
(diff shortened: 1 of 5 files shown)


anyj0527 (approved):
LGTM✔

gst/nnstreamer/elements/gsttensor_split.c
```diff
@@ -306,6 +308,8 @@ gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
 
   name = g_strdup_printf ("src_%u", split->num_srcpads);
   pad = gst_pad_new_from_static_template (&src_templ, name);
+  stream_id = gst_pad_create_stream_id_printf (pad, element,
```
songgot, line 311: Commit is great!!! 
And '''stream_id``` is good as a variable name. : 👍
There are several places where ```sid``` is used as a variable meaning 'stream id'.
At first, I had a hard time understanding what ```sid``` was.
I think it would be good to change ```sid``` to ```stream_id```. 
How do you feel about this? :)

songgot (approved):

PR #4544, opened 2024-08-13
[Element] util to get nth info

jaeyun-jung:
Code clean, use util function to get nth tensor-info ptr to prevent index error.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_merge.c
+++ b/gst/nnstreamer/elements/gsttensor_merge.c
@@ -380,6 +380,8 @@ static gboolean
 gst_tensor_merge_get_merged_config (GstTensorMerge * tensor_merge,
     const GstTensorsConfig * in_config, GstTensorsConfig * out_config)
 {
+  GstTensorsInfo *in_info;
+  GstTensorInfo *_info;
   gboolean ret = FALSE;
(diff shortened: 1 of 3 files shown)


gst/nnstreamer/elements/gsttensor_merge.c
```diff
@@ -380,6 +380,8 @@ static gboolean
 gst_tensor_merge_get_merged_config (GstTensorMerge * tensor_merge,
     const GstTensorsConfig * in_config, GstTensorsConfig * out_config)
 {
+  GstTensorsInfo *in_info;
+  GstTensorInfo *_info;
```
songgot, line 384: You've done a great job with the code. 👍 
Is there any other variable that can differentiate between tensorsInfo and tensorInfo?
It might be a bit difficult to think about.
```tensors_info``` and ``` tensor_info``` ?  The variable name is long. Is this okay? How do you feel about this?

jaeyun-jung: _info for temporal indexed ptr of input or output info

anyj0527 (approved):
LGTM👍

niley7464 (approved):
LGTM!

PR #4546, opened 2024-08-13
[CodeClean] invalid function to handle mem block

jaeyun-jung:
If gst-buffer has multiple mem blocks, memory from get-all-memory function does not update original data.
Update function - replace old memory.

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-flatbuf.cc
+++ b/ext/nnstreamer/tensor_decoder/tensordec-flatbuf.cc
@@ -172,7 +172,7 @@ fbd_decode (void **pdata, const GstTensorsConfig *config,
   if (gst_buffer_get_size (outbuf) == 0)
     gst_buffer_append_memory (outbuf, out_mem);
   else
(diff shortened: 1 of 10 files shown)


ext/nnstreamer/tensor_decoder/tensordec-flatbuf.cc
```diff
@@ -172,7 +172,7 @@ fbd_decode (void **pdata, const GstTensorsConfig *config,
   if (gst_buffer_get_size (outbuf) == 0)
     gst_buffer_append_memory (outbuf, out_mem);
   else
-    gst_memory_unref (out_mem);
+    gst_buffer_replace_all_memory (outbuf, out_mem);
```
songgot, line 175: I was wondering when ```out_mem``` is freed, but it's done inside the ```_replace_memory()```,
thank you.!! :smile:

songgot (approved):
LGTM :smile:

anyj0527 (approved):
LGTM👍

PR #4548, opened 2024-08-14
[decoder] Add yolov10 bounding box decoder

anyj0527:
- YoloV10 is an end-to-end (NMS-free) object detection model.
  Its output dimension is 6:#MAX_DETECTION:1 (output[#MAX_DETECTION][6]).
  The 6 values are [x, y, w, h, conf, class_id].
- Add simple SSAT testcases with raw files.

**Self evaluation:**

REF: https://docs.ultralytics.com/models/yolov10/


songgot (approved):
Thank you for your effort :smiley:

niley7464 (approved):
LGTM!
Last time I refactored the decoder bounding box, there were only two types of yolo, so I bundled them into one file.
If more types are added in the future, I will have to consider separating them 😆

PR #4550, opened 2024-08-22
[SNPE] function to delete instance

jaeyun-jung:
Remove unnecessary local val and use proper function to delete buffer handle.


anyj0527 (approved):
Great, It seems that low versions of SNPE V2 does not provide API `Snpe_UserBufferEncoding_Delete`.
LGTM👍

niley7464 (approved):
LGTM!

PR #4551, opened 2024-08-22
[Meson] check snpe ver

jaeyun-jung:
Check snpe version by using snpe-util function.
Also, update build message.


anyj0527 (approved):
LGTM👍 This make it cover some old versions of SNPE 2.

niley7464 (approved):
I didn't know meson could do something like this.
LGTM 👍

myungjoo (approved):

PR #4565, opened 2024-09-02
[filter] Fix QNN API usages

anyj0527:
- Set `tensor.v1.memType` properly.
- Fix argument value to `compose_graphs`.
- Use nns_log instead of glib log.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_qnn.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_qnn.cc
@@ -308,7 +308,7 @@ QnnManager::init (const char *model_path, const char *backend_path)
   /* Prepare graphs from model.so */
   model_module = g_module_open (model_path, G_MODULE_BIND_LOCAL);
   if (!model_module) {
-    g_warning ("Failed to open module: %s", g_module_error ());
+    nns_loge ("Failed to open module: %s", g_module_error ());
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_qnn.cc
```diff
@@ -321,7 +321,7 @@ QnnManager::init (const char *model_path, const char *backend_path)
 
   uint32_t graph_count = 0U;
   ModelError_t mrt = compose_graphs (backend_handle_, qnn_interface_, context_handle_,
-      nullptr, 1, &graphs_info_, &graph_count, false, nullptr, QNN_LOG_LEVEL_ERROR);
+      nullptr, 0U, &graphs_info_, &graph_count, false, nullptr, QNN_LOG_LEVEL_ERROR);
```
songgot, line 324: Thank you for PR :+1:
I have a question, I want to know the difference between 1 and 0U. :)

songgot (approved):
LGTM :+1:

ext/nnstreamer/tensor_filter/tensor_filter_qnn.cc
anyj0527: It's the number of `GraphConfigInfo_t *` being passed to compose_graphs and it's nullptr here. So it should be 0.

PR #4567, opened 2024-09-03
[doc] Add a new product using nnstreamer

anyj0527:
- MediaTek uses nnstreamer in their products to show some AI demo apps.


songgot (approved):
LGTM :100:

jaeyun-jung (approved):

PR #4588, opened 2024-09-25
[llama2] Copy input prompt to prevent buffer overflow

niley7464:
This patch fixes input memory buffer overflow.
Sometimes, there is another string right after input data.
llama2 api uses prompt string directly, so unintended prompt is added in that case.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
@@ -149,13 +149,17 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
   char *result = NULL, *prompt, *output_data;
   int len, string_idx = 0, i;
 
-  prompt = (char *) input[0].data;
+  prompt = (char *) malloc (input[0].size + 1);
+  memcpy (prompt, input[0].data, input[0].size);
+  prompt[input[0].size] = '\0';
+
   len = strlen (prompt);
 
-  if (len > 1) {
-    prompt[strlen (prompt) - 1] = ' '; // Remove new line from text file
-  } else {
-    prompt = NULL;
+  if (len > 0) {
+    /* Remove new line from text file */
+    char c = prompt[len - 1];
+    if (c == '\r' || c == '\n')
+      prompt[len - 1] = '\0';
   }
 
   result = get_tokens (transformer, tokenizer, sampler, prompt, steps);
@@ -173,6 +177,7 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
   output[0].data = output_data;
 
   free (result);
+  free (prompt);
 }
 
 /**


ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -149,7 +149,8 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
   char *result = NULL, *prompt, *output_data;
   int len, string_idx = 0, i;
 
-  prompt = (char *) input[0].data;
+  prompt = (char *) malloc (input[0].size);
```
jaeyun-jung, line 152: suggestion: eos
```
prompt = (char *) malloc (input[0].size + 1);
memcpy (prompt, input[0].data, input[0].size);
prompt[input[0].size] = '\0';
```

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -149,13 +149,14 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
...
+  memcpy (prompt, input[0].data, input[0].size);
+  prompt[input[0].size] = '\0';
+
   len = strlen (prompt);
 
-  if (len > 1) {
-    prompt[strlen (prompt) - 1] = ' '; // Remove new line from text file
-  } else {
-    prompt = NULL;
+  if (len > 1 && prompt[len - 1] == '\n') {
```
jaeyun-jung, line 158: need to consider all cr, lf, or cr-lf case? TT
```
  if (len > 1) {
    /* Remove new line from text file */
    char c = prompt[len - 1];
    if (c == '\r' || c == '\n')
      prompt[len - 1] = '\0';
  }
```

jaeyun-jung, line 158: cond len > 0
`if (len > 0 && prompt[len - 1] == '\n')`

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
niley7464: Updated it
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -149,13 +149,14 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
...
+  memcpy (prompt, input[0].data, input[0].size);
+  prompt[input[0].size] = '\0';
+
   len = strlen (prompt);
 
-  if (len > 1) {
-    prompt[strlen (prompt) - 1] = ' '; // Remove new line from text file
-  } else {
-    prompt = NULL;
+  if (len > 0 && prompt[len - 1] == '\n') {
```
jaeyun-jung, line 158: or, replacing cr/lf to space?
```
void
TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *output)
{
  char *result, *prompt;
  int len, string_idx;

  prompt = (char *) malloc (input[0].size + 1);
  memcpy (prompt, input[0].data, input[0].size);
  prompt[input[0].size] = '\0';
  len = strlen (prompt);

  /* Remove new line from text file */
  string_idx = 0;
  while (string_idx < len) {
    if (prompt[string_idx] == '\r' || prompt[string_idx] == '\n')
      prompt[string_idx] = ' ';
    string_idx++;
  }

  result = get_tokens (transformer, tokenizer, sampler, prompt, steps);

  /** @todo Make out_info flexible */
  output[0].size = tokenizer->max_token_length * steps;
  output[0].data = malloc (output[0].size);
  memset (output[0].data, '\0', output[0].size);
  memcpy (output[0].data, result, strlen (result));

  free (result);
  free (prompt);
}
```

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
niley7464: I missed it because I'm not familiar with `\r`
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -149,13 +149,17 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
...
+
   len = strlen (prompt);
 
-  if (len > 1) {
-    prompt[strlen (prompt) - 1] = ' '; // Remove new line from text file
-  } else {
-    prompt = NULL;
+  if (len > 0) {
+    /* Remove new line from text file */
+    char c = prompt[len - 1];
```
songgot, line 160: I may have misunderstood the code. If we modify it like this, will it behave incorrectly?
```
  prompt = (char *) input[0].data;
  len = strlen (prompt);

  if (len > 0) {
    char c = prompt[len - 1];
    if (c == '\r' || c == '\n')
      prompt[len - 1] = '\0';
  }
```

niley7464: If input[0].data is not end with `\r` or `\n` and if there are some data right behind input[0].data, llama2_api could not know where the string ends.
I added memcpy because I thought it would be better to fix the sub plugin than fix the API.

songgot: Thank you for the detailed explanation. :pray:

songgot (approved):
LGTM :+1:

PR #4589, opened 2024-09-25
[tensor_trainer] Remove flexible CAPS from src pad

songgot:
- Remove flexible CAPS from src pad because the output buffer is fixed to the values training_loss, training_accuracy, validation_loss and validation_accuracy

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_trainer.c
+++ b/gst/nnstreamer/elements/gsttensor_trainer.c
@@ -35,9 +35,14 @@
 #include <math.h>
 
 /**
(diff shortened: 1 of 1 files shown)


gichan-jang (approved):
LGTM

gichan-jang:
If the only data to be pushed downstream is stats, I think it can be a sink type element without src pad.

niley7464 (approved):
LGTM!

songgot:
> If the only data to be pushed downstream is stats, I think it can be a sink type element without src pad.

Thank you

PR #4595, opened 2024-10-02
[Filter/llama2] function to allocate mem

jaeyun-jung:
Add internal function to allocate memory, to handle possible error case if result is null or allocation failure.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
@@ -22,7 +22,6 @@
 #include <assert.h>
 #include <errno.h>
 #include <nnstreamer_cppplugin_api_filter.hh>
-#include <nnstreamer_log.h>
 #include <nnstreamer_util.h>
 
 #define TEMPERATURE \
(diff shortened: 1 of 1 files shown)


niley7464 (approved):
LGTM!
Thank you 👍

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -166,17 +167,15 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
...
-  output_data = (char *) malloc (output[0].size);
-  len = strlen (result);
-  for (i = 0; i < len; i++) {
-    output_data[string_idx++] = result[i];
-  }
-  for (i = string_idx; i < (int) output[0].size; i++) {
-    output_data[i] = '\0';
+  output[0].data = allocate (output[0].size);
+
+  if (result) {
```
gichan-jang, line 172: Q. What does it mean if result is NULL?
In this case, is it right to allocate memory?
Don't need to throw exception?

jaeyun-jung: allocating empty string, to prevent pipeline freeze.

gichan-jang (approved):
LGTM

PR #4606, opened 2024-10-17
[filter] Add a new filter exeuctorch-llama

anyj0527:
- Meta's executorch provides llama inference with simple APIs.
- As utilize it, add a simple filter name `exeuctorch-llama`.

REF: https://github.com/pytorch/executorch/blob/main/examples/models/llama
Tested with [this executorch commit](https://github.com/pytorch/executorch/commit/c242c4cec93320a017df8096bab84000d4072c1e).

pipeline usage:
```
echo "Hello who are you and what brings you here?" > input.txt && \
gst-launch-1.0  filesrc location=input.txt ! application/octet-stream ! \
tensor_filter framework=executorch-llama model=llama3_2.pte,llama3.2-1b/tokenizer.model invoke-dynamic=true ! \
other/tensors,format=flexible ! tensor_decoder mode=octet_stream !
application/octet-stream ! filesink location=result.txt --no-position \
&& cat llm.log
> Hello who are you and what brings you here?
I am a 35 year old female and I work for the school system in North Carolina. I am currently in the process of starting a new chapter of the International Association of Transpersonal Psychology. I have been a student in IATP for about a year and I have really enjoyed the work. I have been looking for a new way to reach out to my peers and to have a way to network and share information about the work of the Association.
I have read that there is a new group starting up in Raleigh, North Carolina. Would you be willing to share some information about the
```
```
echo "Once upon a time in Seoul," > input.txt && \
gst-launch-1.0  filesrc location=input.txt ! application/octet-stream ! \
tensor_filter framework=executorch-llama model=xnnpack_stories110M.pte,stories110M.tokenizer.bin invoke-dynamic=true ! \
application/octet-stream ! filesink location=result.txt  --no-position \
&& cat result.txt
> Once upon a time in Seoul,
Near the beach, there was a big sea. On the beach there were many kids. One of the kids was a boy. He was very sad.
The other kids asked him why he was so sad. The boy said he was looking for something special. He had been searching for a long time.
The other kids were so kind. They offered to help him. They all looked for something together. After a while, one of the kids found something in the sand. It was a pretty shell!
```

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/meson.build
+++ b/ext/nnstreamer/tensor_filter/meson.build
@@ -859,6 +859,32 @@ if executorch_support_is_available
   )
 endif
 
+if executorch_llama_support_is_available
+  filter_sub_executorch_sources = ['tensor_filter_executorch_llama.cc']
+  nnstreamer_filter_executorch_llama_deps = [executorch_llama_support_deps, glib_dep, nnstreamer_single_dep]
+
+  executorch_llama_compile_args = cxx.get_supported_arguments([
+    '-Wno-sign-compare',
+    '-Wno-deprecated-declarations'
+  ])
+
+  nnstreamer_filter_executorch_llama_deps += declare_dependency(compile_args: executorch_llama_compile_args)
+
+  shared_library('nnstreamer_filter_executorch-llama',
+    filter_sub_executorch_sources,
+    dependencies: nnstreamer_filter_executorch_llama_deps,
+    install: true,
+    install_dir: filter_subplugin_install_dir
+  )
+
+  static_library('nnstreamer_filter_executorch-llama',
+    filter_sub_executorch_sources,
+    dependencies: nnstreamer_filter_executorch_llama_deps,
+    install: true,
+    install_dir: filter_subplugin_install_dir
+  )
+endif
+
 if onnxruntime_support_is_available
   filter_sub_onnxruntime_sources = ['tensor_filter_onnxruntime.cc']
 
--- a/ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
@@ -1,0 +1,248 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+
+/**
+ * @file    tensor_filter_executorch_llama.cc
+ * @date    16 Oct 2024
+ * @brief   NNStreamer tensor-filter sub-plugin for ExecuTorch Llama
+ * @author
+ * @see     https://github.com/nnstreamer/nnstreamer
+ *          https://github.com/pytorch/executorch/blob/main/examples/models/llama
+ * @bug     No known bugs.
+ * @note    Currently experimental.
+ * @details A pipeline example:
+ *          gst-launch-1.0  filesrc location=input.txt ! application/octet-stream !
+ *          tensor_converter ! other/tensors,format=flexible !
+ *          tensor_filter framework=executorch-llama model=model.pte,tokenizer.model invoke-dynamic=true !
+ *          other/tensors,format=flexible ! tensor_decoder mode=octet_stream !
+ *          application/octet-stream ! filesink location=result.txt
+ */
+
+#include <glib.h>
+#include <nnstreamer_cppplugin_api_filter.hh>
+#include <nnstreamer_log.h>
+#include <nnstreamer_plugin_api_util.h>
+#include <nnstreamer_util.h>
+#include <vector>
+
+#include <executorch/extension/module/module.h>
+#include <executorch/extension/tensor/tensor.h>
+
+#include <executorch/examples/models/llama/runner/runner.h>
+
+using namespace torch::executor;
+
+namespace nnstreamer
+{
+namespace tensorfilter_executorch_llama
+{
+
+#ifdef __cplusplus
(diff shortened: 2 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * @brief Method to prepare/configure ExecuTorch instance.
+ */
+void
+executorch_llama_subplugin::configure_instance (const GstTensorFilterProperties *prop)
+{
+  try {
+    /* Load model file (.pte) */
+    if (!g_file_test (prop->model_files[0], G_FILE_TEST_IS_REGULAR)) {
```
niley7464, line 99: I'm not sure, but should we check this even if verify_model_path is true?

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  ET_CHECK_MSG (
+      runner->generate (prompt, 128,
+          [&result] (const std::string &generated) { result.append (generated); })
+          == Error::Ok,
+      "Generating text failed");
+
+  output[0].size = result.size ();
+  output[0].data = g_malloc0 (output[0].size);
+  memcpy (output[0].data, result.c_str (), result.size ());
```
songgot, line 159: In my personal opinion,
It is necessary to check if the value of ```result.size()``` is valid and if ```output[0].data``` is allocated memory.
After that, doing ```memcpy``` seems to prevent the crash. how about think of this? right?

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
```diff
@@ -0,0 +1,250 @@
...
+ * @brief Method to execute the model.
+ */
+void
+executorch_llama_subplugin::invoke_dynamic (GstTensorFilterProperties *prop,
+    const GstTensorMemory *input, GstTensorMemory *output)
+{
+  if (!input)
+    throw std::runtime_error ("Invalid input buffer, it is NULL.");
+  if (!output)
+    throw std::runtime_error ("Invalid output buffer, it is NULL.");
```
songgot, line 146: The error output code is duplicated. I understand it as a branch according to input and output.
I don't have a good idea, but How about modifying it as below and outputting the point?
Or there may be a way to change the log message.
```
if (!input || !output) {  
        throw runtime_error("There is at least one invaild buffer in input(" + string(reinterpret_cast<void*>(input)) + ") and output(" + string(reinterpret_cast<void*>(output)) + ").");  
    } 
```

myungjoo (approved):

ext/nnstreamer/tensor_filter/meson.build
```diff
@@ -859,6 +859,27 @@ if executorch_support_is_available
   )
 endif
 
+if executorch_llama_support_is_available
+  filter_sub_executorch_sources = ['tensor_filter_executorch_llama.cc']
+  nnstreamer_filter_executorch_llama_deps = [executorch_llama_support_deps, glib_dep, nnstreamer_single_dep]
+  nnstreamer_filter_executorch_llama_deps += declare_dependency(compile_args: ['-Wno-sign-compare'])
+  nnstreamer_filter_executorch_llama_deps += declare_dependency(compile_args: ['-Wno-deprecated-declarations'])
```
jaeyun-jung, line 866: suggestion:
```
executorch_llama_compile_args = cxx.get_supported_arguments([
  '-Wno-sign-compare', '-Wno-deprecated-declarations'
])

nnstreamer_filter_executorch_llama_deps += declare_dependency(compile_args: executorch_llama_compile_args)
```

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/meson.build
anyj0527: That's clean. I updated the code as you suggested.

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
anyj0527: Thanks for pointing it. I changed `verify_model_path` as 0.
I found that it does not care `GST_TF_FW_V1` properly. It should be fixed later.

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
anyj0527: `result` is valid std::string and g_malloc aborts if mem allocation failed. I think it is unnecessary to check here.

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
niley7464: Oh, I tested using llama2c filter and indeed `verify_model_path` is not working at v1 😭

ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
anyj0527: Well I think the current code is more clear.

songgot (approved):
LGTM :+1:

Issue #4553, opened 2024-08-24
Following the meson build tutorial may lead to unexpected install paths. 


aychun:
# Issue Description
Following the [meson build tutorial](https://github.com/nnstreamer/nnstreamer/blob/main/Documentation/getting-started-meson-build.md) led to an incorrect install path **/sr**/lib .. ). Looks like meson `join_path()` isn't returning the correct path for my case (meson version: 0.53.2)

Expected Result
============

`$ meson build --prefix=/ --sysconfdir=/etc --libdir=/usr/lib --bindir=/usr/bin --includedir=/usr/include` to install nnstreamer under /usr/... 

How to Reproduce
===============
I added the following lines in `meson.build` then followed the installation tutorial.
```
message('prefix: @0@'.format(nnstreamer_prefix))
message('libddir: @0@'.format(nnstreamer_libdir))
message('bindir: @0@'.format(nnstreamer_bindir))
message('includdir: @0@'.format(nnstreamer_includedir))
message('inidir: @0@'.format(nnstreamer_inidir))
```

`$ meson build --prefix=/ --sysconfdir=/etc --libdir=/usr/lib --bindir=/usr/bin --includedir=/usr/include`

It outputs the following messages

```markdown
The Meson build system
Version: 0.53.2
Source dir: /home/yooeun/workspace/tmp/nnstreamer
Build dir: /home/yooeun/workspace/tmp/nnstreamer/build
Build type: native build
Project name: nnstreamer
Project version: 2.4.1
C compiler for the host machine: cc (gcc 9.4.0 "cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0")
C linker for the host machine: cc ld.bfd 2.34
C++ compiler for the host machine: c++ (gcc 9.4.0 "c++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0")
C++ linker for the host machine: c++ ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64

.
.
.

Message: prefix: /
Message: libddir: /sr/lib
Message: bindir: /sr/bin
Message: includdir: /sr/include
Message: inidir: /etc
```

`$ cat build/nnstreamer.ini`
```
# Set 1 or True if you want to set your custom sub-plugins' path with env variables.
# Please be informed that, configured ini file should be in RO partition for the release binary.
[common]
enable_envvar=True
enable_symlink=True


[filter]
[...]

PR #4610, opened 2024-10-21
meson: 0.50 doesn't support version with find_program

myungjoo:
version argument of find_program is avaiable since 0.52. Skip it if it's not 0.52 or higher.

Addresses the secondary point of #4553


jaeyun-jung (approved):

anyj0527 (approved):
LGTM👍

PR #4611, opened 2024-10-22
[llama2] Use invoke_dynamic in tensor filter llama2

niley7464:
This patch adds invoke_dynamic funciton to tensor filter llama2.
LLM model's output is flexible, so llama2 filter should use invoke_dynamic.

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
@@ -15,13 +15,15 @@
  * @details Pipeline example
  *          gst-launch-1.0 filesrc location=input.txt ! application/octet-stream !
  *          tensor_converter ! other/tensors,format=flexible !
- *          tensor_filter framework=llama2c model=model.bin,tokenizer.bin !
- *          filesink location=output.txt
+ *          tensor_filter framework=llama2c model=model.bin,tokenizer.bin invoke-dynamic=TRUE !
+ *          other/tensors,format=flexible ! tensor_decoder mode=octet_stream !
+ *          application/octet-stream ! filesink location=output.txt
  */
 
 #include <assert.h>
 #include <errno.h>
 #include <nnstreamer_cppplugin_api_filter.hh>
+#include <nnstreamer_plugin_api_util.h>
 #include <nnstreamer_util.h>
 
 #define TEMPERATURE \
@@ -64,6 +66,8 @@ class TensorFilterLlama2c : public tensor_filter_subplugin
   tensor_filter_subplugin &getEmptyInstance ();
   void configure_instance (const GstTensorFilterProperties *prop);
   void invoke (const GstTensorMemory *input, GstTensorMemory *output);
+  void invoke_dynamic (GstTensorFilterProperties *prop,
+      const GstTensorMemory *input, GstTensorMemory *output);
   void getFrameworkInfo (GstTensorFilterFrameworkInfo &info);
   int getModelInfo (model_info_ops ops, GstTensorsInfo &in_info, GstTensorsInfo &out_info);
   int eventHandler (event_ops ops, GstTensorFilterFrameworkEventData &data);
@@ -82,8 +86,7 @@ class TensorFilterLlama2c : public tensor_filter_subplugin
   static Tokenizer *tokenizer;
   static Sampler *sampler;
   static Config *config;
-  char *prompt; // prompt string
-  int steps; // number of steps to run for
+  int steps = 0; // number of steps to run for
 
   static void *allocate (size_t size);
 };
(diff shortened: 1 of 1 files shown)


niley7464:
https://github.com/nnsuite/llama2.c/blob/0804e8a19f747225f3f92aeda02b480974763439/transformer.c#L13
```c
#define PRAGMA_OMP_PARALLEL_FOR_PRIVATE(x) DO_PRAGMA(omp parallel for private(x))
```

This part causes 
```
Freeing pipeline ...
Caught SIGSEGV
Caught SIGSEGV
Caught SIGSEGV
Caught SIGSEGV
Caught SIGSEGV
Caught SIGSEGV
```

Tested with no openmp-support
```
# features
option('openmp-support', type: 'feature', value: 'disabled')
```

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -165,17 +182,24 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
 
   result = get_tokens (transformer, tokenizer, sampler, prompt, steps);
 
-  /** @todo Make out_info flexible */
-  output[0].size = tokenizer->max_token_length * steps;
+  if (!result) {
```
jaeyun-jung, line 185: Q: is it possible if returned token is empty ?
```
if (!result || (len = strlen (result)) == 0)
```

niley7464: `result`'s memory is allocated in llama2c, so it must be freed even if length is 0.
I added a logic to check if string is empty!
Also, I found `prompt`'s memory is not relased by your review.
Thank you :)

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -15,13 +15,14 @@
  * @details Pipeline example
  *          gst-launch-1.0 filesrc location=input.txt ! application/octet-stream !
  *          tensor_converter ! other/tensors,format=flexible !
- *          tensor_filter framework=llama2c model=model.bin,tokenizer.bin !
- *          filesink location=output.txt
+ *          tensor_filter framework=llama2c model=model.bin,tokenizer.bin invoke-dynamic=TRUE !
+ *          application/octet-stream ! filesink location=output.txt
```
myungjoo, line 19: Is if ok to "decode" ```other/tensors``` to ```application/octet-stream``` with caps-filter?

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -147,7 +150,21 @@ TensorFilterLlama2c::configure_instance (const GstTensorFilterProperties *prop)
...
+}
+
+/**
+ * @brief Invoke llama2c using input tensors
+ */
+void
+TensorFilterLlama2c::invoke_dynamic (GstTensorFilterProperties *prop,
+    const GstTensorMemory *input, GstTensorMemory *output)
+{
+  char *result = nullptr, *prompt;
```
myungjoo, line 167: Be careful. ```prompt``` is also a class property.
Using same name for a local var and a class property is not good.

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -83,7 +86,7 @@ class TensorFilterLlama2c : public tensor_filter_subplugin
   static Sampler *sampler;
   static Config *config;
   char *prompt; // prompt string
```
myungjoo, line 88: This ```prompt``` property is not used...

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
```diff
@@ -165,17 +182,32 @@ TensorFilterLlama2c::invoke (const GstTensorMemory *input, GstTensorMemory *outp
 
   result = get_tokens (transformer, tokenizer, sampler, prompt, steps);
 
-  /** @todo Make out_info flexible */
-  output[0].size = tokenizer->max_token_length * steps;
-  output[0].data = allocate (output[0].size);
+  if (!result) {
```
myungjoo, line 185: LINE 168-181 == strndup

myungjoo: LINE 174 implies that the input tensor[0] is a char string ending with '0'. Is this a safe assumption?

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
myungjoo: And.. what happens if in/out tensors are format=static?

niley7464: Oh, I missed tensor_decoder line 😅 Updated it!
This filter is for text generation, so if the tensor_filter's invoke-dynamic option is FALSE (in/out static) it thorws runtime error. 
Same with #4606

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
niley7464: Removed unused `prompt`
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llama2.cc
niley7464: Used `strndup` for input and output data :)
'\0' is arbitrarily added at LINE 171 to use `strlen`.

anyj0527 (approved):
LGTM👍

myungjoo (approved):

PR #4615, opened 2024-10-22
github-action/risc-v: update ubuntu to 22.04

myungjoo:
20.04 appears to be dropped.
Update to 22.04


myungjoo:
Nah.. 20.04 is still there. This risc-v docker error has something different cause.

PR #4616, opened 2024-10-23
[Trainer] delete unnecessary code

jaeyun-jung:
Code clean, remove unnecessary code and use util function for tensor info.


gst/nnstreamer/elements/gsttensor_trainer.c
```diff
@@ -858,8 +855,7 @@ gst_tensor_trainer_create_output_tensors (GstTensorTrainer * trainer,
     GstMemory ** out_mem, GstMapInfo * out_info)
```
songgot, line 855: In my personal opinion, since we are not creating output tensors inside the function, shouldn't we also change the function name?

myungjoo (approved):

gst/nnstreamer/elements/gsttensor_trainer.c
jaeyun-jung: I changed function name and update code to parse input buffer. plz review the changes again.

PR #4619, opened 2024-10-29
[util] Let `gst_tensors_info_to_string` handle tensor_format

anyj0527:
- Let the util function properly set format information.
- Skip copying n-th info if the format is not static.
- Add missing header `nnstreamer_log.h` in llama2 source.

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
@@ -511,6 +511,9 @@ gst_tensors_info_copy (GstTensorsInfo * dest, const GstTensorsInfo * src)
   num = dest->num_tensors = src->num_tensors;
   dest->format = src->format;
 
+  if (src->format != _NNS_TENSOR_FORMAT_STATIC)
+    return;
+
   for (i = 0; i < num; i++) {
     _dest = gst_tensors_info_get_nth_info (dest, i);
     _src = gst_tensors_info_get_nth_info ((GstTensorsInfo *) src, i);
--- a/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
@@ -787,31 +787,37 @@ gst_tensors_info_to_string (const GstTensorsInfo * info)
   unsigned int limit = info->num_tensors;
   GstTensorInfo *_info;
 
-  g_string_append_printf (gstr, "Num_Tensors = %u, Tensors = [",
-      info->num_tensors);
-  if (limit > NNS_TENSOR_SIZE_LIMIT) {
-    limit = NNS_TENSOR_SIZE_LIMIT;
-    g_string_append_printf (gstr,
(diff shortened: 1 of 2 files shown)


gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
```diff
@@ -511,6 +511,9 @@ gst_tensors_info_copy (GstTensorsInfo * dest, const GstTensorsInfo * src)
   num = dest->num_tensors = src->num_tensors;
   dest->format = src->format;
 
+  if (src->format != _NNS_TENSOR_FORMAT_STATIC)
+    return;
+
```
songgot, line 516: If we look at the API description, it mentions using gst_tensors_info_free() after using this API.

However, if the format is not ```_NNS_TENSOR_FORMAT_STATIC```,  ```gst_tensor_info_copy()``` is not performed, so it seems that ```gst_tensors_info_free()``` should not be performed.
It seems that the comment needs to be edited or code added to check the format in ```gst_tensors_info_free()```.
Right?

```
/**
 * @brief Copy tensor info
 * @note Copied info should be freed with gst_tensors_info_free()
 */
void
gst_tensors_info_copy (GstTensorsInfo * dest, const GstTensorsInfo * src)
{
[...]
 + if (src->format != _NNS_TENSOR_FORMAT_STATIC)
 +   return;
    gst_tensor_info_copy (_dest, _src);
}
```

niley7464 (approved):
LGTM!

gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
anyj0527: I think `gst_tensors_info_init` is called in L511, and makes the free function good to go.

songgot: Thank you, I got it. 
gst_tensors_info_init (info) is called twice, but it is not a problem.
```
/**
 * @brief Free allocated data in tensors info structure
 * @param info tensors info structure
 */
void
gst_tensors_info_free (GstTensorsInfo * info)
{
  guint i;

  g_return_if_fail (info != NULL);

  for (i = 0; i < NNS_TENSOR_MEMORY_MAX; i++) {
    gst_tensor_info_free (&info->info[i]);
  }

  if (info->extra) {
    for (i = 0; i < NNS_TENSOR_SIZE_EXTRA_LIMIT; ++i)
      gst_tensor_info_free (&info->extra[i]);

    g_free (info->extra);
    info->extra = NULL;
  }

  /* Init default */
  gst_tensors_info_init (info);
}
```

songgot (approved):
LGTM :100:

gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
songgot: Let's refactor it later. :smile:

jaeyun-jung (approved):

PR #4625, opened 2024-11-04
[MqttSrc] fix leak case

jaeyun-jung:
Code clean, caps_replace() function increases ref count of new caps.
Fix mem leak case when using caps_replace().

### DIFF
--- a/gst/mqtt/mqttsrc.c
+++ b/gst/mqtt/mqttsrc.c
@@ -1097,6 +1097,7 @@ cb_mqtt_on_message_arrived (void *context, char *topic_name, int topic_len,
   GstBaseSrc *basesrc;
   GstMqttSrc *self;
   GstClock *clock;
+  GstCaps *recv_caps;
   gsize offset;
   guint i;
   UNUSED (topic_name);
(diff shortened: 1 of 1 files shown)


gst/mqtt/mqttsrc.c
```diff
@@ -1141,11 +1141,13 @@ cb_mqtt_on_message_arrived (void *context, char *topic_name, int topic_len,
...
-      gst_mqtt_src_renegotiate (basesrc);
-    } else {
-      gst_caps_replace (&recv_caps, NULL);
+    if (recv_caps) {
+      if (!gst_caps_is_equal (self->caps, recv_caps)) {
+        gst_caps_replace (&self->caps, recv_caps);
+        gst_mqtt_src_renegotiate (basesrc);
+      }
+
+      gst_caps_unref (recv_caps);
```
songgot, line 1150: When ```recv_caps``` is created, it is correct to do ```gst_caps_unref``` in L1150.
However, if the ```if (!gst_caps_is_equal (self->caps, recv_caps))``` is satisfied and ```gst_caps_replace``` is done, the ref of ```recv_caps``` increases.
Therefore, ```gst_catp_ref (recv_caps)``` is needed again.

```
The reference count of @olddata is decreased and the reference count of @newdata is increased.
```
```
gboolean
gst_caps_replace (GstCaps ** old_caps, GstCaps * new_caps)
{
  return gst_mini_object_replace ((GstMiniObject **) old_caps,
      (GstMiniObject *) new_caps);
}
```
```
/**
 * gst_mini_object_replace:
 * @olddata: (inout) (transfer full) (nullable): pointer to a pointer to a
 *     mini-object to be replaced
 * @newdata: (allow-none): pointer to new mini-object
 *
 * Atomically modifies a pointer to point to a new mini-object.
 * The reference count of @olddata is decreased and the reference count of
 * @newdata is increased.
 *
 * Either @newdata and the value pointed to by @olddata may be %NULL.
 *
 * Returns: %TRUE if @newdata was different from @olddata
 */
gboolean
gst_mini_object_replace (GstMiniObject ** olddata, GstMiniObject * newdata)
```

jaeyun-jung: Cannot understand, caps_replace() increases ref count of new caps.
So after calling caps_replace(), the ref count of recv_caps is 2.
After gst_caps_unref (recv_caps) at line 1150 (precondition is caps is not equal), its ref count is 1 and self->caps holds this.

songgot (approved):
LGTM :+1:

gichan-jang:
This change is integrated into https://github.com/nnstreamer/nnstreamer/pull/4631. Close this PR>

PR #4629, opened 2024-11-06
[Spec] Skip test in case that ASAN is enabled

again4you:
In case of GCC v14, the build process is stopped when ASAN option is enabled. Build engineer suggested to skip the test in such case to avoid the build failure.

**Self evaluation:**


songgot (approved):
LGTM :+1:

jaeyun-jung (approved):

PR #4630, opened 2024-11-06
[Action] Generate build result badge

gichan-jang:
- Generate gbs build result badge for each arch.

**Self evaluation:**

### DIFF
--- a/.github/workflows/gen_gbs_result_badge.py
+++ b/.github/workflows/gen_gbs_result_badge.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+##
+# SPDX-License-Identifier: LGPL-2.1-only
+# Copyright (C) 2024 Samsung Electronics
+#
+# @file gen_gbs_result_badge.py
+# @brief A tool for generating a github badge image in svg format
+# @author Gichan Jang <gichan2.jang@samsung.com>
+#
+import sys
+import os
+import json
+from pybadges import badge
+
+##
+# @brief Save the badge on the given path
+# @param[in] path A file path to save the svg file
+# @param[in] badge github badge svg file
+def save_badge (path, badge):
+    try:
+        with open(path, "w") as f:
+            f.write(badge)
+    except IOError as e:
+        print("Failed to save badge:", str(e))
+
+##
+# @brief Generate a github badge svg file representing daily build result
+# @param[in] html build result of arch.
+# @param[in] path A directory path to save the svg file
+def gen_build_result_badge(json_str, out_dir):
+    daily_result="success"
+    for arch in json_str:
+        try:
+            result = json_str[arch]
+            daily_color = color= 'green'
+            if result == 'failure':
+              color = 'red'
+              daily_result = result
+              daily_color = 'red'
+
+            s = badge(left_text='test', right_text=result, right_color=color)
+            path = os.path.join(out_dir, arch + '_reuslt.svg')
+            save_badge (path, s)
+        except Exception as e:
+            print(f"An error occurred while processing {arch} :", str(e))
+    try:
+        s = badge(left_text='Daily build', right_text=result, right_color=color)
+        path = os.path.join(out_dir, 'daily_reuslt.svg')
+        save_badge (path, s)
+    except Exception as e:
+        print("An error occurred while processing daily build badge:", str(e))
+
+if __name__ == '__main__':
+    if len(sys.argv) < 3:
+        exit(1)
+
+    try:
+        with open(sys.argv[1], "r") as input_file:
+            data = json.load(input_file)
+            gen_build_result_badge(data, sys.argv[2])
+    except Exception as e:
+        print("An error occurred:", str(e))
--- a/.github/workflows/update_gbs_cache.yml
+++ b/.github/workflows/update_gbs_cache.yml
@@ -10,6 +10,11 @@ on:
 
 jobs:
   cache_gbs_build:
+    outputs:
+      x86_64: ${{ steps.gbs-result.outputs.x86_64 }}
+      i586: ${{ steps.gbs-result.outputs.i586 }}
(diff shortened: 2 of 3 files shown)


.github/workflows/gen_gbs_result_badge.py
```diff
@@ -0,0 +1,52 @@
...
+
+##
+# @brief Save the badge on the given path
+# @param[in] path A file path to save the svg file
+# @param[in] badge github badge svg file
+def save_badge (path, badge):
+    file = open(path, "w")
+    file.write(badge)
+    file.close()
+
```
songgot, line 23: Thank you for always working hard for the part. 👍 
Thanks for the great PR.

I think it would be good to modify it like this. How about you?
First, ```file.colse()``` can be handled automatically if we use ```with```.
If we add ```try exception```, it would be good because we can handle exception errors when opening or writing.
```
def save_badge (path, badge):    
    try:    
        with open(path, "w") as f:    
            f.write(badge)    
    except IOError as e:    
        print("Failed to save badge:", str(e)) 
```

.github/workflows/gen_gbs_result_badge.py
```diff
@@ -0,0 +1,52 @@
...
+    path = os.path.join(out_dir, 'daily_reuslt.svg')
+    save_badge (path, s)
+
+if __name__ == '__main__':
+    if len(sys.argv) < 3:
+        exit(1)
+
+    file = open(sys.argv[1], "r")
+
+    gen_build_result_badge (json.load(file), sys.argv[2])
```
songgot, line 52: I think it would be good to modify this part as follows. 😄 
```
try:  
        with open(sys.argv[1], "r") as input_file:  
            data = json.load(input_file)  
            gen_build_result_badge(data, sys.argv[2])  
    except Exception as e:  
        print("An error occurred:", str(e))  
```

.github/workflows/gen_gbs_result_badge.py
```diff
@@ -0,0 +1,52 @@
...
+        result = json_str[arch]
+        daily_color = color= 'green'
+        if result == 'failure':
+          color = 'red'
+          daily_result = result
+          daily_color = 'red'
+
+        s = badge(left_text='test', right_text=result, right_color=color)
+        path = os.path.join(out_dir, arch + '_reuslt.svg')
+        save_badge (path, s)
```
songgot, line 40: If we convert a JSON string into a dictionary using the ```json.load()``` method and try to access the key called ```[arch]``` in the dictionary, a ```KeyError``` may occur if the key does not exist.
Although it is not strictly necessary, it is recommended to handle ```KeyError``` in case of such exceptions.
How about you?
```
try:
        result = json_str[arch]
        daily_color = color= 'green'
        if result == 'failure':
          color = 'red'
          daily_result = result
          daily_color = 'red'

        s = badge(left_text='test', right_text=result, right_color=color)
        path = os.path.join(out_dir, arch + '_reuslt.svg')
        save_badge (path, s)
exception KeyError as e:
         print(f"KeyError occurred while processing {arch}: {str(e)}")    
         continue
```

.github/workflows/update_gbs_cache.yml
```diff
@@ -98,3 +110,26 @@ jobs:
...
+          python3 .github/workflows/gen_gbs_result_badge.py gbs_build_result.json ~/testresult
+          git clone https://${{ secrets.TAOS_ACCOUNT }}:${{ secrets.TAOS_ACCOUNT_TOKEN }}@github.com/nnstreamer/nnstreamer.github.io.git
+          pushd nnstreamer.github.io
+          cp -r ~/testresult/*.svg testresult
+          git config user.email "nnsuite@samsung.com"
+          git config user.name "nnsuite"
+          git add *
+          git commit -s -m "Update unit test result badge."
+          git push origin main -f
+          popd
```
songgot, line 135: You seem to be good at using Yaml files. Thank you for the great code, I can learn a lot from this PR. 🙏 
Thank you very mush 👍

.github/workflows/gen_gbs_result_badge.py
gichan-jang: Thanks, I applied your suggestion.

.github/workflows/gen_gbs_result_badge.py
gichan-jang: Updated.

.github/workflows/update_gbs_cache.yml
```diff
@@ -98,3 +110,26 @@ jobs:
...
+          echo '${{ toJSON( needs.cache_gbs_build.outputs) }}' > gbs_build_result.json
+          python3 .github/workflows/gen_gbs_result_badge.py gbs_build_result.json ~/testresult
+          git clone https://${{ secrets.TAOS_ACCOUNT }}:${{ secrets.TAOS_ACCOUNT_TOKEN }}@github.com/nnstreamer/nnstreamer.github.io.git
+          pushd nnstreamer.github.io
+          cp -r ~/testresult/*.svg testresult
+          git config user.email "nnsuite@samsung.com"
+          git config user.name "nnsuite"
+          git add *
+          git commit -s -m "Update unit test result badge."
+          git push origin main -f
```
songgot, line 134: By the way, regarding this part,
I think it could be reused similarly in multiple workflows.
it would be possible to reuse the code if you use 'composite action'. Thank you.

https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action

gichan-jang: Badges and logs should be uploaded for other platforms(android, yocto and etc.).
It would be nice to use composite action or reusable workflow to reduce the duplicated codes.
I will work it in the next PR.

songgot: Thank you, have a nice holiday. :+1:

songgot (approved):
Thank you for the great PR. :100:

jaeyun-jung (approved):

PR #4637, opened 2024-11-14
[Action] Fix for failure case

gichan-jang:
Even if the step failed, the `continue-on-error` was used to continue the next step. 
But `continue-on-error` marks success even if the step fails. 
So, let's use `always()` for subsequent steps.

**Self evaluation:**


myungjoo (approved):

niley7464 (approved):
LGTM!

anyj0527 (approved):

songgot (approved):
Thanks for the great PR. :+1:

PR #4639, opened 2024-11-15
dist: disable tflite2-custom

myungjoo:
tflite2-custom (using custom tflite2 binary per pipeline) is not stable and not maintained.

Disable it from the default builds until it gets stable.


anyj0527 (approved):
LGTM✔

songgot (approved):
LGTM :+1:

jaeyun-jung (approved):

PR #4655, opened 2024-11-28
spec: clean up build options and conditions

myungjoo:
We have a lot of optional external libraries, and
sub-packages created by such libraries.

Place dependency declarations of such external libraries
and build option declarations together for easier
management.

Then, simplify meson option management, too.


myungjoo:
spec/lua recursion is killing it. need to remove recursion.

PR #4656, opened 2024-12-05
[llamacpp] Add llamacpp tensor_filter subplugin 

niley7464:
This PR adds llamacpp tensor_filter subplugin.
llama.cpp supports more options like kv cache and lora adapter than the existing llama2.c

See also : https://github.com/ggerganov/llama.cpp/blob/master/examples/simple/simple.cpp
Model from : https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF

### How to use
```sh
echo "Hello my name is" > input.txt && \
gst-launch-1.0 filesrc location=input.txt ! application/octet-stream ! \
tensor_filter framework=llamacpp model=llama-2-7b-chat.Q2_K.gguf invoke-dynamic=TRUE ! \
application/octet-stream ! filesink location=output.txt && \
cat output.txt
```
Result
```
llama_model_loader: loaded meta data with 19 key-value pairs and 291 tensors from llama-2-7b-chat.Q2_K.gguf (version GGUF V2)
llama_model_loader: Dumping metadata keys/values. Note: KV overrides do not apply in this output.
...
llama_new_context_with_model: graph nodes  = 1030
llama_new_context_with_model: graph splits = 1
Pipeline is PREROLLED ...
Setting pipeline to PLAYING ...
New clock: GstSystemClock
Got EOS from element "pipeline0".
Execution ended after 0:00:00.000100061
Setting pipeline to NULL ...
Freeing pipeline ...
<s> Hello my name is Jake and I'm a 25 year old male from the United States. Unterscheidung between the two is important because they have different meanings and imp
```

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -1,0 +1,345 @@
+/**
+ * GStreamer Tensor_Filter, llamacpp Module
+ * Copyright (C) 2024 Yelin Jeong <yelini.jeong@samsung.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * @file    tensor_filter_llamacpp.cc
+ * @date    5 Dec 2024
+ * @brief   llamacpp module for tensor_filter gstreamer plugin
+ * @see     http://github.com/nnstreamer/nnstreamer
+ *          https://github.com/ggerganov/llama.cpp
+ * @author  Yelin Jeong <yelini.jeong@samsung.com>
+ * @bug     No known bugs except for NYI items
+ * @details Pipeline example
+ *          gst-launch-1.0 filesrc location=input.txt ! application/octet-stream !
+ *          tensor_converter ! other/tensors,format=flexible !
+ *          tensor_filter framework=llamacpp model=llama-2-7b-chat.Q2_K.gguf invoke-dynamic=TRUE !
+ *          other/tensors,format=flexible ! tensor_decoder mode=octet_stream !
+ *          application/octet-stream ! filesink location=output.txt
+ */
+
+#include <assert.h>
+#include <errno.h>
+#include <nnstreamer_cppplugin_api_filter.hh>
+#include <nnstreamer_log.h>
+#include <nnstreamer_plugin_api_util.h>
+#include <nnstreamer_util.h>
+#include <string>
+#include <vector>
+
+#include "ggml.h"
+#include "llama.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif /* __cplusplus */
+void init_filter_llamacpp (void) __attribute__ ((constructor));
+void fini_filter_llamacpp (void) __attribute__ ((destructor));
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+
+namespace nnstreamer
+{
+namespace tensorfilter_llamacpp
+{
+/**
+ * @brief Class for TensorFilterLlamaCpp subplugin
+ */
+class TensorFilterLlamaCpp : public tensor_filter_subplugin
+{
+  public:
+  TensorFilterLlamaCpp ();
+  ~TensorFilterLlamaCpp ();
+
+  /* mandatory methods */
+  tensor_filter_subplugin &getEmptyInstance ();
+  void configure_instance (const GstTensorFilterProperties *prop);
+  void invoke (const GstTensorMemory *input, GstTensorMemory *output);
+  void invoke_dynamic (GstTensorFilterProperties *prop,
+      const GstTensorMemory *input, GstTensorMemory *output);
+  void getFrameworkInfo (GstTensorFilterFrameworkInfo &info);
+  int getModelInfo (model_info_ops ops, GstTensorsInfo &in_info, GstTensorsInfo &out_info);
+  int eventHandler (event_ops ops, GstTensorFilterFrameworkEventData &data);
+
+  /* static methods */
+  static void init_filter_llama ();
+  static void fini_filter_llama ();
+
+  private:
+  static TensorFilterLlamaCpp *registered;
+  static const char *name;
+
+  llama_context *ctx;
+  llama_model *model;
+  llama_sampler *smpl;
+  llama_model_params model_params;
+  llama_sampler_chain_params sampler_params;
+  llama_context_params ctx_params;
+};
+
+void init_filter_llama (void) __attribute__ ((constructor));
+void fini_filter_llama (void) __attribute__ ((destructor));
+
+TensorFilterLlamaCpp *TensorFilterLlamaCpp::registered = nullptr;
+const char *TensorFilterLlamaCpp::name = "llamacpp";
+
+/**
+ * @brief Construct a new llamacpp subplugin instance
+ */
+TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
+    : tensor_filter_subplugin (), ctx (nullptr), model (nullptr), smpl (nullptr)
+{
+}
+
+/**
+ * @brief Destructor of TensorFilterllamacpp
+ */
+TensorFilterLlamaCpp::~TensorFilterLlamaCpp ()
+{
+  if (model) {
+    llama_free_model (model);
+    model = nullptr;
+  }
+
+  if (smpl) {
+    llama_sampler_free (smpl);
+    smpl = nullptr;
+  }
+}
+
+/**
+ * @brief Method to get an empty object
+ */
+tensor_filter_subplugin &
+TensorFilterLlamaCpp::getEmptyInstance ()
+{
+  return *(new TensorFilterLlamaCpp ());
+}
+
+/**
+ * @brief Configure llamacpp instance
+ */
+void
+TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
+{
+  /** @todo: Get ngl value from prop */
+  int ngl = 99;
+
+  if (!prop->invoke_dynamic) {
+    throw std::invalid_argument (
+        "llamacpp only supports invoke-dynamic mode. Set `invoke-dynamic=true`");
+  }
+
+  /** load dynamic backends */
+  ggml_backend_load_all ();
+  model_params = llama_model_default_params ();
+  model_params.n_gpu_layers = ngl;
+
+  model = llama_load_model_from_file ((char *) prop->model_files[0], model_params);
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,337 @@
...
+        break;
+      }
+
+      char buf[128];
+      int n = llama_token_to_piece (model, new_token_id, buf, sizeof (buf), 0, true);
+      if (n < 0) {
+        throw std::invalid_argument ("Failed to convert token to piece");
+      }
+      std::string s (buf, n);
+      result += s;
```
niley7464, line 233: If this filter is used in tensor_generator, this part have to push `s` to `gst_pad` (`gst_pad_push()`)

gichan-jang: How is this function reused in the tensor_generator?
Are you going to extract this function as a common file later?

niley7464: I'm still thinking about it because I can't find an easy way right now...
Any idea will be helpful 🤗

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,337 @@
...
+  n_prompt = -llama_tokenize (model, prompt.c_str (), prompt.size (), NULL, 0, true, true);
+
+  std::vector<llama_token> prompt_tokens (n_prompt);
+  if (llama_tokenize (model, prompt.c_str (), prompt.size (),
+          prompt_tokens.data (), prompt_tokens.size (), true, true)
+      < 0) {
+    throw std::invalid_argument ("Failed to tokenize the prompt");
+  }
+
+  ctx_params = llama_context_default_params ();
```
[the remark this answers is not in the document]
niley7464: Updated to set params(context, sampler) in `configure_instance` and left `n_prompt` related value.
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,339 @@
...
+  static const accl_hw hw_list[];
+  static const int num_hw;
+
+  llama_context *ctx;
+  llama_model *model = nullptr;
+  llama_sampler *smpl = nullptr;
+  llama_model_params model_params;
+  llama_sampler_chain_params sampler_params;
+  llama_context_params ctx_params;
+  llama_batch batch;
```
myungjoo, line 87: batch is used only in ```TensorFilterLlamaCpp::invoke_dynamic``` and its value is never reused.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,339 @@
...
+ * @brief Invoke llamacpp using input tensors
+ */
+void
+TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *output)
+{
+  UNUSED (input);
+  UNUSED (output);
+
+  throw std::runtime_error (
+      "llamacpp only supports invoke-dynamic mode. Set `invoke-dynamic=true`");
```
myungjoo, line 160: Please check and return error at the init phase (CAPS negotiation or configure_instance, whichever earlier possible)

Or...

Find a way to force-enable invoke-dynamic (change the property) at initialization phase and emit warning if it wasn't set true.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,339 @@
...
+ */
+TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
+{
+}
+
+/**
+ * @brief Destructor of TensorFilterllamacpp
+ */
+TensorFilterLlamaCpp::~TensorFilterLlamaCpp ()
+{
```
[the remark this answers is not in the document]
niley7464: Added clenaup to the end invoke functions and destructor.
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: `prop` value is const at the init phase and inaccessible from anywhere else so I added exception instead of change the property.
Thank you!

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: Removed `batch` from class member value and moved it to `invoke_dynamic`.
Thank you!

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,350 @@
...
+#include <vector>
+
+#include "ggml.h"
+#include "llama.h"
+/**
+ * @brief Macro for debug mode.
+ */
+#ifndef DBG
+#define DBG FALSE
+#endif
```
songgot, line 37: Thanks for the great PR. :+1: 
This file does not use DBG, how about removing this?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,350 @@
...
+  static void init_filter_llama ();
+  static void fini_filter_llama ();
+
+  private:
+  static TensorFilterLlamaCpp *registered;
+  static const char *name;
+
+  llama_context *ctx;
+  llama_model *model;
+  llama_sampler *smpl;
```
songgot, line 81: We also use the word ```sampler``` in line 83. How about changing ```smpl``` to ```smpler``` for better code readability?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/**
+ * @brief Configure llamacpp instance
+ */
+void
+TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
+{
+  /** @todo: Get ngl value from prop */
+  int ngl = 99;
```
songgot, line 133: How about renaming 'ngl' to 'num_gpu_layers' for code readability?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,350 @@
...
+
+/**
+ * @brief Invoke llamacpp using input tensors
+ */
+void
+TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
+    const GstTensorMemory *input, GstTensorMemory *output)
+{
+  /** @todo: Get n_predict value from prop */
+  int n_prompt, n_pos, n_predict = 32, n_decode = 0;
```
songgot, line 183: I'm wondering if ``` n_predict' '' is fixed or if it will be brought in as a property.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -0,0 +1,350 @@
...
+    if (llama_decode (ctx, batch)) {
+      throw std::invalid_argument ("Failed to eval");
+    }
+
+    n_pos += batch.n_tokens;
+
+    /** sample the next token */
+    {
+      new_token_id = llama_sampler_sample (smpl, ctx, -1);
+
```
songgot, line 235: Doesn't this function need to throw an exception on failure?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: The value name is from https://github.com/ggerganov/llama.cpp/blob/master/examples/simple/simple.cpp
How about leave it as is for easy comparison with the original code?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: > How about renaming 'ngl' to 'num_gpu_layers' for code readability?

Ditto

> Any plans to bring in the number of layers as a property?

I plan to upload the contents of the prop as a separate PR in the future.
``` /** @todo: Get ngl value from prop */ ```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: It's not fixed so I will updated to use property later.
```  /** @todo: Get n_predict value from prop */```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: Could I know what `failure` means?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
niley7464: Removed unused value!
Thank you :)

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Good idea, Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I'm wondering if a valid value is always passed.

niley7464: I couldn't find the part that validates the id value in several codes in the example 😥

songgot: Okay :smile:

songgot (approved):
Thank you.

myungjoo (approved):

PR #4659, opened 2024-12-11
[orc/transform] Use 64bit int orc funcs in tensor_transform

anyj0527:
- Add orc functions for 64bit integers and let tensor_transform use 64bit orc funcs.
- Add TCs for {add/mul, typecast from/to} orc funcs which handle 64bit integers.

**Self evaluation:**


myungjoo (approved):

wooksong (approved):
LGTM!

Issue #4660, opened 2024-12-11
[Design] tensor-filter HAL for layer separation


myungjoo:
Approach A.
- Interface in C struct with function pointers
- Caution: there can be 2 or more instances of NPU HAL objects! (we already have one coming soon)
- Keep C++ (tensor_filter_subplugin of nnstreamer 2.4.x as HAL 2.0) semantics, but interface in C
    - Each C++ method ==> 3 C functions of HAL
        - E.g., ```tensor_filter_subplugin::configure_instance``` ==>
```
int func_tensor_filter_subplugin_HAL_configure_instance_prepare (struct X(tbd), void **ptr);
int func_tensor_filter_subplugin_HAL_configure_instance_main (struct X(tbd), void **ptr);
int func_tensor_filter_subplugin_HAL_configure_instance_post (struct X(tbd), void **ptr);
```
        - Then, the upper layer (the subplugin in C++) calls:
```
thehardware::configure_instance(P) {
   // some C++ code that needs to interact with nnstreamer

   someptr->configure_instance_prepare (converter(P. TBD), this->ptr);

   // some C++ code that needs to interact with nnstreamer

   someptr->configure_instance_main (converter(P. TBD), this->ptr);

   // some C++ code that needs to interact with nnstreamer

   someptr->configure_instance_post (converter(P. TBD), this->ptr);

   // some C++ code that needs to interact with nnstreamer

}
```

myungjoo:
Closing: the layer separation described here (Approach A) has been implemented and shipped.

The tensor-filter HAL now exists in-tree as `ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc`, with the SNPE and Vivante backends (`tensor_filter_tizen_hal_snpe.cc`, `tensor_filter_tizen_hal_vivante.cc`). The C-interface/C++-semantics split ended up realized through Tizen's own `hal-ml.h` infrastructure rather than the hand-rolled `*_prepare/_main/_post` triples sketched above, which is what the original note anticipated ("assuming that Tizen HAL will provide such infrastructure", #4661).

It is packaged and released: `packaging/nnstreamer.spec` builds it under `tizen_hal_support` and ships `nnstreamer-tizen-hal` with `libnnstreamer_filter_tizen-hal.so`.

The skeleton PR #4661 was closed unmerged; the production implementation superseded it. Please reopen or file a new issue if a non-Tizen HAL binding is still wanted.

PR #4661, opened 2024-12-12
Skeleton for Tizen HAL Support

myungjoo:
This is a skeleton (not buildable) code to show how Approach A of
https://github.com/nnstreamer/nnstreamer/issues/4660 works.

Note that the mechanism to load such a HAL driver is not included, assuming that Tizen HAL will provide such infrastructure.

CC: @again4you @anyj0527


myungjoo:
As yongjoo's Tizen HAL implementation progresses, I'm closing this proposal.

PR #4666, opened 2024-12-20
[Test] lib path to run edge test

jaeyun-jung:
Fix lib path when edge_test is enabled.

This will fix integration test failure on nnstreamer-edge repo.


gichan-jang (approved):
LGTM

again4you (approved):
Nice Fix!

PR #4671, opened 2025-01-03
[Subplugin] util lib for sub-plugins

jaeyun-jung:
Build and install common util library for sub-plugins.


again4you (approved):
I checked the compile time error, caused by `non-virtual-dtor`, is fixed by this patch.

gichan-jang (approved):
LGTM

PR #4682, opened 2025-01-22
action: run ubuntu arm build/test in arm machines

myungjoo:
Building and running at aarch64 machines
is much faster than at x64 machines.

Note: actions/setup-python@v1 is not compatible with arm machines. Update it to the latest.


myungjoo:
Nah.. gbs won't work at aarch64. gbs does not have aarch64 files :(

We will need to build and run tests on ubuntu-arm machines without Tizen.

dkjung (approved):
LGTM

jaeyun-jung (approved):

PR #4699, opened 2025-03-12
[Android] header for nns version

jaeyun-jung:
Generate header file of nnstreamer version info under the local path, to build android library on android-studio.


anyj0527 (approved):

myungjoo (changes requested):
we do not have nnstreamer_version.h
we only have nnstreamer_version.h.in

update android build script along with this update.

jaeyun-jung:
I just found that nnstreamer uses ndk-build, I suggest building android library using the script of api repo.
(clone api repo and build)

myungjoo (commented):
Let's wait for API updates.

songgot (approved):
LGTM, Thank you.

jaeyun-jung:
I updated this PR - generating version header under local-path to construct the ml-api project using android-studio.

myungjoo (approved):

PR #4700, opened 2025-03-13
[Edge] Add custom property

gichan-jang:
Add custom property for edgesrc/edgesink.
, separated options(key:value) are set on the edge handle.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [*]Skipped
2. Run test: [*]Passed [ ]Failed [*]Skipped

### DIFF
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -43,6 +43,7 @@ enum
   PROP_WAIT_CONNECTION,
   PROP_CONNECTION_TIMEOUT,
   PROP_CUSTOM_LIB,
+  PROP_CUSTOM_PROPS,
 
   PROP_LAST
 };
@@ -134,6 +135,10 @@ gst_edgesink_class_init (GstEdgeSinkClass * klass)
       g_param_spec_string ("custom-lib", "Custom connection lib path",
           "User defined custom connection lib path.",
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_CUSTOM_PROPS,
+      g_param_spec_string ("custom-props", "Custom connection props",
+          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options.",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
   gst_element_class_add_pad_template (gstelement_class,
       gst_static_pad_template_get (&sinktemplate));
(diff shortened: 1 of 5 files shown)


myungjoo (approved):
LGTM.

Add a usage example of the custom property on the boilerplate.
E.g.,:
https://github.com/nnstreamer/nnstreamer/blob/3ba5b6c7a980f8a0d5266b38442377b0c63e6eec/gst/join/gstjoin.c#L13-L30

gst/edge/edge_sink.c
```diff
@@ -390,6 +428,11 @@ gst_edgesink_start (GstBaseSink * basesink)
     return FALSE;
   }
 
+  if (0 != nns_edge_connect (self->edge_h, self->dest_host, self->dest_port)) {
+    nns_loge ("Failed to connect to edge device!");
+    return FALSE;
+  }
+
```
myungjoo, line 435: Is this a bugfix, unrelated with this topic?

myungjoo (changes requested):
Regression with edge-connection. (unittest fail due to this change)

myungjoo:
@gichan-jang This PR has been inactive since March 2025 with changes requested, and it now has merge conflicts with main. Are you still pursuing this? If so, please rebase and address the review comments; otherwise I will close it in a few weeks to keep the review queue clean. Feel free to reopen or submit a fresh PR anytime.

myungjoo:
Superseded by #4871, which is now merged — so the `custom-props` feature is in main, with your authorship preserved.

Since this PR had been inactive for a while and had grown merge conflicts, I took it over rather than let it go stale: your commit was rebased onto main and kept as-is, authored by you with your original `Signed-off-by:` intact (`75c0b894` on main).

What changed during the takeover, for the record:

- **Dropped the `nns_edge_connect()` call added to `gst_edgesink_start()`.** This was the edge-connection regression I reported here: edgesink is a PUB node, so connecting at start() broke the NULL→READY transition whenever the peer was not up yet, and made `wait-connection` meaningless.
- **Hardened the parser** and moved it into a shared `gst_edge_parse_custom_props()` in edge_common, so the two elements no longer carry duplicate copies. The old parser read `str_op[1]` without checking the vector length, which was an out-of-bounds read on an empty token (a trailing comma was enough), and it split on every `:`, so a value like `QUEUE_SIZE:10:OLD` was silently truncated.
- **Kept option values out of logs and off the bus.** Since custom-props is the documented place for options of a custom connection library, a value may be a token or credential; a rejected option is now named by its key, or by its position when no key can be parsed out of it.
- **Added tests that actually run the parser.** The tests here only covered the GObject set/get round trip, so the parsing at start() was never executed and none of the above would have been caught by CI.

Thank you for the contribution, @gichan-jang — the feature and your commit are in main. Closing this one as superseded.

PR #4701, opened 2025-03-18
[filter] Add a new filter for general tizen-hal

anyj0527:
- Add a new filter for tizen-hal.
- This filter get the user-given hal-backend name as custom prop.

Usage:
`... ! tensor_filter framework=tizen-hal custom=backend:snpe model=path.to.model ! ...`

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
@@ -1,0 +1,319 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/**
+ * NNStreamer tensor_filter, sub-plugin for Tizen HAL
+ * Copyright (C) 2025 Yongjoo Ahn <yongjoo1.ahn@samsung.com>
+ */
+/**
+ * @file      tensor_filter_tizen_hal.cc
+ * @date      15 Jan 2025
+ * @brief     NNStreamer tensor-filter sub-plugin for Tizen HAL
+ * @see       http://github.com/nnstreamer/nnstreamer
+ * @author    Yongjoo Ahn <yongjoo1.ahn@samsung.com>
+ * @bug       No known bugs except for NYI items
+ *
+ * This is the per-NN-framework plugin (Tizen HAL) for tensor_filter.
+ */
+
+#include <iostream>
+#include <string>
+
+#include <glib.h>
+#include <nnstreamer_cppplugin_api_filter.hh>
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
```diff
@@ -0,0 +1,326 @@
...
+  for (guint op = 0; op < g_strv_length (options); ++op) {
+    gchar **option = g_strsplit (options[op], ":", -1);
+
+    if (g_strv_length (option) >= 2) {
+      g_strstrip (option[0]);
+      g_strstrip (option[1]);
+
+      if (g_ascii_strcasecmp (option[0], "backend") == 0) {
+        backend_name = g_strdup (option[1]);
+        break;
```
jaeyun-jung, line 119: I think break is unnecessary here (printing warning message for useless option?)

ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
```diff
@@ -0,0 +1,326 @@
...
+{
+  if (!hal_handle) {
+    nns_loge ("HAL backend is not configured.");
+    return -1;
+  }
+
+  hal_ml_param_h param = nullptr;
+  int ret = hal_ml_param_create (&param);
+  if (ret != HAL_ML_ERROR_NONE) {
+    nns_loge ("Failed to create hal_ml_param");
```
jaeyun-jung, line 241: need return error

ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
```diff
@@ -0,0 +1,326 @@
...
```
jaeyun-jung, line 273: also, need return error

ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
anyj0527: Thanks! I make it print warning msg

ext/nnstreamer/tensor_filter/tensor_filter_tizen_hal.cc
anyj0527: I've made it return an error, PTAL

jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #4702, opened 2025-03-21
[RepoSrc] mem in out buffer

jaeyun-jung:
When creating buffer, appending tensor may allocate new memory.
Append memory after updating its contents.


gichan-jang (approved):
LGTM

niley7464 (approved):
LGTM

myungjoo (approved):

PR #4710, opened 2025-03-28
[Action] Add nnstreamer testhub action.

gichan-jang:
- Upload coverage results to TestHub
 - Runs an action on only for OASIS org.

**Self evaluation:**

### DIFF
--- a/.github/workflows/parse_ssat.py
+++ b/.github/workflows/parse_ssat.py
@@ -1,0 +1,78 @@
+#!/usr/bin/python
+##
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Copyright (C) 2025 Samsung Electronics
+#
+# @file    parse_ssat.py
+# @brief   Parse SSAT result to xml format
+
+import sys
+
+if len(sys.argv) < 3:
+  sys.exit(1)
+
+LOG_PATH="unittest_result/" + sys.argv[1]
+XML_PATH="unittest_result/" + sys.argv[2]
+TIMESTAMP=sys.argv[3]
+
+log_f = open(LOG_PATH, 'r')
+xml_f = open(XML_PATH, 'w')
+
+xml_dic = {}
+total_tests = 0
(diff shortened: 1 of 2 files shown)


.github/workflows/parse_ssat.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/python
+##
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# Copyright (C) 2025 Samsung Electronics
+#
+# @file    parse_ssat.py
+# @brief   Parse SSAT result to xml formet
```
jaeyun-jung, line 8: typo formet :)

niley7464 (approved):
Checked the execution result of the oasis org and testhub.
LGTM :)

I have one question about "Allow all actions" to nnstreamer-mirror in oasis.
Daily test and per PR check workflows are also triggerd and never end in oasis repo.
I think adding  `if: github.repository_owner == 'nnstreamer'` condition to the rest of the workflow would solve the problem, but I'm not sure if it's a good way 😥

gichan-jang:
> Checked the execution result of the oasis org and testhub. LGTM :)
> 
> I have one question about "Allow all actions" to nnstreamer-mirror in oasis. Daily test and per PR check workflows are also triggerd and never end in oasis repo. I think adding `if: github.repository_owner == 'nnstreamer'` condition to the rest of the workflow would solve the problem, but I'm not sure if it's a good way 😥

I know what you're worried. But I manually disabled all workflows except this workflow. Let me know if there's a problem on mirrored repo.

.github/workflows/parse_ssat.py
gichan-jang: Thanks, updated.

anyj0527 (approved):
LGTM👍

jaeyun-jung (approved):

songgot (approved):
LGTM

PR #4716, opened 2025-04-01
[workflow] Fix daily action failures

anyj0527:
- Fix python version to 3.11, which has `imghdr` module.
- Recent versions of cmake is not compatible with hotdoc.
   Set certain version of cmake to be used.
- Let all workflows use same `setup-python` (3.11)

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [X]Skipped


songgot (approved):
LGTM

jaeyun-jung (approved):

PR #4720, opened 2025-04-03
[Common] unnecessary log

jaeyun-jung:
While appending memory into gst-buffer, nnstreamer internally checks header in the memory to check flexible tensor.
To remove unnecessary log print, fix condition to validate meta info.


songgot (approved):
LGTM

gichan-jang (approved):
LGTM

PR #4725, opened 2025-04-30
[decoder] Do not clear properties_table in BoundingBox destructor

glneo:
The properties_table contains the list of properties for loaded sub-modules. The properties are only added on initial module load which happens once, this is why the table is a static singleton not tied to the lifecycle of any one BoundingBox instance.

The issue is this table is cleared in the BoundingBox destructor, which means after any BoundingBox instance is removed, no new instances will be able to find any sub-module properties.

Fix this by not clearing the properties_table in the destructor.


myungjoo (approved):
Nice Catch! Thanks!

PR #4730, opened 2025-05-08
[Wait #4729][llamacpp] Implement a function that passes the output to tensor_filter by calling the callback

songgot:
[llamacpp] Implement a function that passes the output to tensor_filter by calling the callback
    
llamacpp sub-plugin takes a callback from eventHandler and passes each generated token to tensor_filter by calling that callback.

### How to use
```sh
echo "Hello my name is" > input.txt && \
gst-launch-1.0 filesrc location=input.txt ! application/octet-stream ! \
tensor_converter ! other/tensors,format=flexible ! \
tensor_filter framework=llamacpp model=llama-2-7b-chat.Q2_K.gguf invoke-dynamic=TRUE ! \
other/tensors,format=flexible ! tensor_decoder mode=octet_stream ! \
application/octet-stream ! filesink location=output.txt && \
cat output.txt
```
Result
```
<s> Hello my name is Jake and I am a 23 year old male from the United States. Unterscheidung between the two is important because they have different meanings and connot
```

PR #4729, opened 2025-05-08
[tensor_filter] Implement asynchronous callback for handling multiple…

songgot:
[tensor_filter] Implement asynchronous callback for handling multiple output from sub-plugins

This commit introduces a callback mechanism inside the tensor_filter component to address scenarios where a sub-plugin produces multiple outputs.  When tensor_filter receives an input, it passes it to the specified sub-plugin.  The callback function previously defined by tensor_filter is triggered for each output produced by the sub-plugin.

ToDo: make property for async output callback

PR #4733, opened 2025-05-09
[Wait #4729, #4730][tensor_filter] Add use-asyn-invoke-callback property

songgot:
- Set this property to pass the output to tensor_filter by calling the asynchronous callback in the subplugin.
- Check if this property is set in llamacpp sub-plugin


songgot:
Closed

PR #4735, opened 2025-05-12
[llamacpp] Dispatch output tensor for each token generation

songgot:
When generating each output token, create an output tensor and dispatch it to nnstreamer_filter.

See also : https://github.com/ggerganov/llama.cpp/blob/master/examples/simple/simple.cpp
Model from : https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF

### How to use
```sh
echo "Hello my name is" > input.txt && \
gst-launch-1.0 filesrc location=input.txt ! application/octet-stream ! \
tensor_filter framework=llamacpp model=llama-2-7b-chat.Q2_K.gguf invoke-dynamic=TRUE ! \
application/octet-stream ! filesink location=output.txt && \
cat output.txt
```
Result
```
<s> Hello my name is Jake and I am a 23 year old male from the United States. Unterscheidung between the two is important because they have different meanings and connot
```

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -20,6 +20,8 @@
  */
 
 #include <assert.h>
+#include <atomic>
+#include <condition_variable>
 #include <errno.h>
 #include <functional>
 #include <memory>
@@ -27,7 +29,9 @@
 #include <nnstreamer_log.h>
 #include <nnstreamer_plugin_api_util.h>
 #include <nnstreamer_util.h>
+#include <queue>
 #include <string>
+#include <thread>
 #include <vector>
 
 #include "ggml.h"
@@ -82,7 +86,17 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   int n_gpu_layers = 99;
   int n_predict = 32;
 
+  std::thread output_thread; /* thread for async output */
+  std::queue<std::pair<GstTensorFilterProperties *, std::string>> input_queue;
+  std::mutex queue_mutex;
+  std::condition_variable queue_cv;
+  std::atomic<bool> stop_thread{ false }; /* flag to stop the thread */
   void parseCustomProperties (const GstTensorFilterProperties *prop);
+  void generateTokens (GstTensorFilterProperties *prop, std::string &&prompt,
+      GstTensorMemory *output);
+  GstTensorMemory *createOutputTensor (
+      const char *buf, int n, GstTensorFilterProperties *prop);
+  void outputThreadLoop ();
 };
 
 TensorFilterLlamaCpp *TensorFilterLlamaCpp::registered = nullptr;
@@ -101,14 +115,26 @@ TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
  */
 TensorFilterLlamaCpp::~TensorFilterLlamaCpp ()
 {
-  if (model) {
-    llama_model_free (model);
-    model = nullptr;
+  stop_thread = true;
+  queue_cv.notify_one ();
+  if (output_thread.joinable ()) {
+    output_thread.join ();
+  }
+
+  {
+    std::lock_guard<std::mutex> lock (queue_mutex);
+    while (!input_queue.empty ())
+      input_queue.pop ();
   }
 
   if (sampler) {
     llama_sampler_free (sampler);
     sampler = nullptr;
+  }
+
+  if (model) {
+    llama_model_free (model);
+    model = nullptr;
   }
 }
 
@@ -198,6 +224,9 @@ TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
 
   sampler = llama_sampler_chain_init (sampler_params);
   llama_sampler_chain_add (sampler, llama_sampler_init_greedy ());
+
+  stop_thread = false;
+  output_thread = std::thread (&TensorFilterLlamaCpp::outputThreadLoop, this);
 }
 
 /**
@@ -214,26 +243,60 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
 }
 
 /**
- * @brief Invoke llamacpp using input tensors
+ * @brief Create output tensor
+ *
+ * The memory allocated for the output tensor data using strndup
+ * must be freed by the caller of this function to avoid memory leaks.
+ */
+GstTensorMemory *
+TensorFilterLlamaCpp::createOutputTensor (
+    const char *buf, int n, GstTensorFilterProperties *prop)
+{
+  if (buf == nullptr || n <= 0 || prop == nullptr) {
+    ml_loge ("Invalid arguments passed to createOutputTensor");
+    return nullptr;
+  }
+
+  GstTensorMemory *output = g_new0 (GstTensorMemory, 1);
+  if (!output) {
+    ml_loge ("createOutputTensor: Allocation failed");
+    return nullptr;
+  }
+
+  output[0].data = g_strndup (buf, n);
+  output[0].size = n; /* Stores string length, excluding NULL appended to the end of the string */
+  if (output[0].data == nullptr) {
+    ml_loge ("createOutputTensor: strndup failed");
+    g_free (output);
+    return nullptr;
+  }
+
+  gst_tensors_info_free (&prop->output_meta);
+  prop->output_meta.num_tensors = 1;
+  prop->output_meta.format = _NNS_TENSOR_FORMAT_FLEXIBLE;
+  prop->output_meta.info[0].type = _NNS_UINT8;
+  prop->output_meta.info[0].dimension[0] = output[0].size;
+
+  return output;
+}
+
+/**
+ * @brief Generates output tokens based on the provided input prompt using the
+ * llamacpp and dispatches the generated tokens to tensor_filter asynchronously.
+ * nnstreamer frees GstTensorMemory output after use. Do not free it.
  */
 void
-TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
-    const GstTensorMemory *input, GstTensorMemory *output)
+TensorFilterLlamaCpp::generateTokens (GstTensorFilterProperties *prop,
+    std::string &&prompt, GstTensorMemory *output)
 {
   const llama_vocab *vocab = llama_model_get_vocab (model);
-  llama_context *ctx = nullptr;
   int n_prompt, n_pos, n;
+  llama_context *ctx = nullptr;
   llama_token new_token_id;
   llama_batch batch;
-  std::string prompt, result = "";
-
-  UNUSED (prop);
-
-  prompt.assign ((char *) input[0].data, input[0].size);
-  prompt.erase (prompt.find_last_not_of (" \t\n\r\f\v") + 1);
+  UNUSED (output);
 
   n_prompt = -llama_tokenize (vocab, prompt.c_str (), prompt.size (), NULL, 0, true, true);
-
   std::vector<llama_token> prompt_tokens (n_prompt);
   n = llama_tokenize (vocab, prompt.c_str (), prompt.size (),
       prompt_tokens.data (), prompt_tokens.size (), true, true);
@@ -252,13 +315,19 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
 
   for (auto id : prompt_tokens) {
     char buf[128] = { 0 };
+    GstTensorMemory *output_async = nullptr;
 
     n = llama_token_to_piece (vocab, id, buf, sizeof (buf), 0, true);
     if (n < 0) {
       throw std::invalid_argument ("Failed to convert token to piece");
     }
-    std::string s (buf, n);
-    result += s;
+
+    output_async = createOutputTensor (buf, n, prop);
+    if (!output_async)
+      throw std::runtime_error ("generateTokens: Failed to create output tensor");
+
+    /* nnstreamer frees GstTensorMemory output after use. Do not free it. */
+    nnstreamer_filter_dispatch_output_async (prop, output_async);
   }
 
   batch = llama_batch_get_one (prompt_tokens.data (), prompt_tokens.size ());
@@ -274,6 +343,7 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
     /* sample the next token */
     {
       char buf[128] = { 0 };
+      GstTensorMemory *output_async = nullptr;
 
       new_token_id = llama_sampler_sample (sampler, ctx, -1);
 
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -286,26 +327,40 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
 
-  gst_tensors_info_free (&prop->output_meta);
-  prop->output_meta.num_tensors = 1;
-  prop->output_meta.format = _NNS_TENSOR_FORMAT_FLEXIBLE;
-  prop->output_meta.info[0].type = _NNS_UINT8;
-  prop->output_meta.info[0].dimension[0] = output[0].size;
+  if (!input || !input[0].data || input[0].size == 0) {
+    throw std::invalid_argument ("Invalid input tensor data");
+  }
+  if (output_thread.joinable ()) {
```
gichan-jang, line 354: Q. If GStreamer pauses and then resumes PLAYING while generateTokens is active, will the join() call block the pipeline? This could cause delays as it waits for the prior thread (generateTokens ) to finish.
Shouldn't the thread be properly cleaned up when the GStreamer pipeline's state changed to PAUSED or NULL?

songgot: There seems to be a lot to consider, I will continue to add more features in the next PRs.
Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -252,13 +289,16 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
+    GstTensorMemory *output = nullptr;
 
     n = llama_token_to_piece (vocab, id, buf, sizeof (buf), 0, true);
     if (n < 0) {
       throw std::invalid_argument ("Failed to convert token to piece");
     }
-    std::string s (buf, n);
-    result += s;
+    output = g_new0 (GstTensorMemory, 1);
+    createOutputTensor (buf, n, output, prop);
```
niley7464, line 299: If `nnstreamer_filter_dispatch_output_async` is always called after `createOutputTensor`, how about call this function inside `createOutputTensor`?

songgot: Adding a dispatch function to createOutputTensor reduces clarity as a single function now performs two roles. It also becomes difficult to make the function name reflect its original role, so I thought it would be structurally preferable to separate tensor creation and dispatch processing by function.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
jaeyun-jung: Should we create new thread for each input?
I suggest adding message queue with lock, pushing message (includes input data),  and fetching a message in the thread.

songgot: Thank you. 
Good idea. The problem is that when invoke is called multiple times, thread creation and destruction are repeated. I fixed it.

songgot: I've updated it, thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -286,26 +344,68 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
+    throw std::invalid_argument ("Invalid input tensor data");
+  }
+
+  /* The part that fills in the prompt value was moved out of the thread, because input is freed when the invoke_dynamic function returns.*/
+  prompt.assign ((char *) input[0].data, input[0].size);
+  prompt.erase (prompt.find_last_not_of (" \t\n\r\f\v") + 1);
+
+  {
+    std::lock_guard<std::mutex> lock (queue_mutex);
+    input_queue.push (std::make_tuple (prop, std::move (prompt), output));
```
jaeyun-jung: IMO passing gst-buffer 'output'  is unnecessary. The output result is allocated in thread.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +244,50 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
+}
+
+/**
+ * @brief Generates output tokens based on the provided input prompt using the
+ * llamacpp and dispatches the generated tokens to tensor_filter asynchronously.
+ * nnstreamer frees GstTensorMemory output after use. Do not free it.
+ */
+void
+TensorFilterLlamaCpp::generateTokens (GstTensorFilterProperties *prop,
+    std::string &&prompt, GstTensorMemory *output)
```
jaeyun-jung, line 290: The param output is unnecessary.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +244,50 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
- * @brief Invoke llamacpp using input tensors
+ * @brief Create output tensor
+ *
+ * The memory allocated for the output tensor data using strndup
+ * must be freed by the caller of this function to avoid memory leaks.
  */
 void
-TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
-    const GstTensorMemory *input, GstTensorMemory *output)
+TensorFilterLlamaCpp::createOutputTensor (const char *buf, int n,
```
jaeyun-jung: Suggestion: change params to allocate new memory and return output buffer
```
GstTensorMemory *
TensorFilterLlamaCpp::createOutputTensor (GstTensorFilterProperties *prop, const char *buf, int n)
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I'm thinking about the following structure. If 'invoke async' is false, I need to pass the output to 'invoke dynamic'.
```
  if (prop->invoke_async) {
    {
      std::lock_guard<std::mutex> lock (queue_mutex);
      input_queue.push (std::make_tuple (prop, std::move (prompt), output));
    }
    queue_cv.notify_one (); /* notify waiting thread */
  } else {
    try {
      generateTokens (prop, std::move (prompt), output);
      /* thread exception */
      g_critical ("Exception occurred during token generation: %s", e.what ());
    }
  }
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I'm thinking about the following structure. Both ```OutputTensor``` and ```generateTokens``` require ```output```.
```

void
TensorFilterLlamaCpp::generateTokens (GstTensorFilterProperties *prop,
    std::string &&prompt, GstTensorMemory *output)
{
  const llama_vocab *vocab = llama_model_get_vocab (model);
  int n_prompt, n_pos, n;
  llama_token new_token_id;
  llama_batch batch;
  std::string output_accumulated;

  n_prompt = -llama_tokenize (vocab, prompt.c_str (), prompt.size (), NULL, 0, true, true);
  std::vector<llama_token> prompt_tokens (n_prompt);
  n = llama_tokenize (vocab, prompt.c_str (), prompt.size (),
      prompt_tokens.data (), prompt_tokens.size (), true, true);
  if (n < 0) {
    throw std::invalid_argument ("Failed to tokenize the prompt");
  }

  ctx_params.n_ctx = n_prompt + n_predict - 1;
  /* n_batch is the maximum number of tokens that can be processed in a single call to llama_decode */
  ctx_params.n_batch = n_prompt;

  ctx = llama_init_from_model (model, ctx_params);
  if (ctx == nullptr) {
    throw std::invalid_argument ("Failed to create llama context");
  }

  for (auto id : prompt_tokens) {
    char buf[128] = { 0 };
    GstTensorMemory *output = nullptr;

    n = llama_token_to_piece (vocab, id, buf, sizeof (buf), 0, true);
    if (n < 0) {
      throw std::invalid_argument ("Failed to convert token to piece");
    }

    if (prop->invoke_async) {
      output = g_new0 (GstTensorMemory, 1);
      createOutputTensor (std::string (buf, n), output, prop);
      /* nnstreamer frees GstTensorMemory output after use. Do not free it. */
      nnstreamer_filter_dispatch_output_async (prop, output);
    } else {
      output_accumulated.append (buf, n);
    }
  }

  batch = llama_batch_get_one (prompt_tokens.data (), prompt_tokens.size ());

  for (n_pos = 0; n_pos + batch.n_tokens < n_prompt + n_predict;) {
    /* evaluate the current batch with the transformer model */
    if (llama_decode (ctx, batch)) {
      throw std::invalid_argument ("Failed to eval");
    }

    n_pos += batch.n_tokens;

    /* sample the next token */
    {
      char buf[128] = { 0 };
      GstTensorMemory *output;

      new_token_id = llama_sampler_sample (sampler, ctx, -1);

      /* is it an end of generation? */
      if (llama_vocab_is_eog (vocab, new_token_id)) {
        break;
      }

      n = llama_token_to_piece (vocab, new_token_id, buf, sizeof (buf), 0, true);
      if (n < 0) {
        throw std::invalid_argument ("Failed to convert token to piece");
      }

      if (prop->invoke_async) {
        output = g_new0 (GstTensorMemory, 1);
        createOutputTensor (std::string (buf, n), output, prop);
        /* nnstreamer frees GstTensorMemory output after use. Do not free it. */
        nnstreamer_filter_dispatch_output_async (prop, output);
      } else {
        output_accumulated.append (buf, n);
      }

      batch = llama_batch_get_one (&new_token_id, 1);
    }
  }

  if (!prop->invoke_async) {
    createOutputTensor (output_accumulated, output, prop);
  }
}
```

jaeyun-jung: Then, allocating new memory `output = g_new0 (GstTensorMemory, 1);` is weird, need to use passed param to avoid mem leak case.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
jaeyun-jung: nned allocating new memory for output?
```
  if (prop->invoke_async) {
    {
      GstTensorMemory *async_out = g_new0 (GstTensorMemory, 1);

      std::lock_guard<std::mutex> lock (queue_mutex);
      input_queue.push (std::make_tuple (prop, std::move (prompt), async_out));
    }
    queue_cv.notify_one (); /* notify waiting thread */
  } else {
    try {
      generateTokens (prop, std::move (prompt), output);
      /* thread exception */
      g_critical ("Exception occurred during token generation: %s", e.what ());
    }
  }
```

songgot: GstTensorMemory must be created equal to the number of tokens that will be generated as output.
So, I think it would be better to generate it in generateTokens.
And, if prop->invoke_async == TRUE, I agree that 'output' doesn't require a push.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I will use a different name from the input parameter.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Okay, I will update it. Thank you.

songgot: I've updated it. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I've updated it. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +243,61 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
+    return nullptr;
+  }
+
+  GstTensorMemory *output = g_new0 (GstTensorMemory, 1);
+  if (!output) {
+    g_critical ("createOutputTensor: Allocation failed");
+    return nullptr;
+  }
+
+  std::string s (buf, n);
```
jaeyun-jung: is it necessary?
```
output[0].data = g_strndup (buf, n);
output[0].size = n + 1;
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +243,61 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
+    const char *buf, int n, GstTensorFilterProperties *prop)
+{
+  if (buf == nullptr || n <= 0 || prop == nullptr) {
+    g_critical ("Invalid arguments passed to createOutputTensor");
+    return nullptr;
+  }
+
+  GstTensorMemory *output = g_new0 (GstTensorMemory, 1);
+  if (!output) {
+    g_critical ("createOutputTensor: Allocation failed");
```
jaeyun-jung: Suggestion: This plugin includes nnstreamer_log.h, so use ml_loge() instead.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Yes.
std::string s(buf, n); // Use first n bytes from buf (may not include NULL)
s.length() == n;       // String length excludes NULL
strndup(s.c_str(), n); // Duplicate n bytes, auto appends NULL
output[0].size = n;    // Stores string length, NULL excluded

songgot: If I use 'n', I think a more detailed explanation is needed. I will update this. Thank you.
```
output[0].data = g_strndup (buf, n);
output[0].size = n;
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Thank you. I've updated it.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -284,28 +353,72 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
+    queue_cv.wait (lock, [this] () {
+      return !input_queue.empty () || stop_thread;
+    }); /*wait when condition is false*/
 
-  gst_tensors_info_free (&prop->output_meta);
-  prop->output_meta.num_tensors = 1;
-  prop->output_meta.format = _NNS_TENSOR_FORMAT_FLEXIBLE;
-  prop->output_meta.info[0].type = _NNS_UINT8;
-  prop->output_meta.info[0].dimension[0] = output[0].size;
+    if (stop_thread && input_queue.empty ())
```
jaeyun-jung: condition to stop thread?
(stop_thread || input_queue.empty ()) or need to clear input-queue in destructor.

songgot: Modified to terminate immediately without processing any remaining input in the queue. Thank you.

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +249,59 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
+ * nnstreamer frees GstTensorMemory output after use. Do not free it.
  */
 void
-TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
-    const GstTensorMemory *input, GstTensorMemory *output)
+TensorFilterLlamaCpp::generateTokens (GstTensorFilterProperties *prop,
+    std::string &&prompt, GstTensorMemory *output)
 {
   const llama_vocab *vocab = llama_model_get_vocab (model);
-  llama_context *ctx = nullptr;
```
jaeyun-jung, line 224: plz consider ctx, newly allocating local value and releasing in generateTokens() method.

songgot: Thank you. I've updated it.

anyj0527 (approved):
LGTM👍

niley7464 (approved):
LGTM 👍

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -101,15 +115,27 @@ TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
  */
 TensorFilterLlamaCpp::~TensorFilterLlamaCpp ()
 {
-  if (model) {
-    llama_model_free (model);
-    model = nullptr;
+  stop_thread = true;
+  queue_cv.notify_one ();
```
myungjoo, line 119: Lock before notify.

PR #4736, opened 2025-05-12
[Filter] Add invoke-async property

songgot:
Set this property when sub-plugin receives an input and generates multiple outputs asynchronously.


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -943,6 +943,10 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
           "input and output of the tensor filter. "
           "With this option, the output caps is always in the format of flexible tensors.",
           FALSE, G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INVOKE_ASYNC,
+      g_param_spec_boolean ("invoke-async", "Enable asynchronous invoke",
+          "Use this when the sub-plugin receives an input and generates multiple outputs asynchronously.",
+          FALSE, G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
```
gichan-jang, line 949: You set
```
if (!prop->invoke_async) {
    throw std::runtime_error (
        "llamacpp must invoke output asynchronously. Set `invoke_async=true`");
  }
```
Then, I think it would be more appropriate to set invoke_async to true by default.

songgot: In most cases where tensor_filter is used, invoke-async should be false. It seems better to set it to true only in special cases like llama.cpp.
Thank you.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -135,8 +144,11 @@ typedef struct _GstTensorFilterProperties
   int latency; /**< The average latency over the recent 10 inferences in microseconds */
   int throughput; /**< The average throughput in the number of outputs per second */
   int invoke_dynamic; /**< True for supporting invoke with flexible output. */
-
+  int invoke_async; /**< Use this when the sub-plugin receives an input and generates multiple outputs asynchronously. */
```
myungjoo: 1. Don't insert a new property. Append it.

2. This is "True for supporting invoke asynchronously". GstTensorFilterProperty is a const object that a framework defines. Wheter to invoke async or not should be determined by the user. And using "async" mode should never be madatory for users.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -943,6 +943,10 @@ gst_tensor_filter_install_properties (GObjectClass * gobject_class)
           "input and output of the tensor filter. "
           "With this option, the output caps is always in the format of flexible tensors.",
           FALSE, G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_INVOKE_ASYNC,
+      g_param_spec_boolean ("invoke-async", "Enable asynchronous invoke",
+          "Use this when the sub-plugin receives an input and generates multiple outputs asynchronously.",
```
myungjoo: Need more explanation:
```
Use this when the sub-plugin receives an input and generates multiple outputs asynchronously.
To enable this, the framework needs to support invoke-async and its model needs to generate multiple output frames for a given input frame.
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -286,26 +327,45 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
+  UNUSED (output);
+  std::string prompt;
+
+  if (!input || !input[0].data || input[0].size == 0) {
+    throw std::invalid_argument ("Invalid input tensor data");
+  }
+  if (!prop->invoke_async) {
+    throw std::runtime_error (
+        "llamacpp must invoke output asynchronously. Set `invoke_async=true`");
+  }
```
myungjoo: Do not mandate asynch mode.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +229,48 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
+  if (buf == nullptr || n <= 0 || output == nullptr || prop == nullptr) {
+    throw std::invalid_argument ("Invalid arguments passed to createOutputTensor");
+  }
+
+  std::string s (buf, n);
+  output[0].size = s.length ();
+  output[0].data = strndup (s.c_str (), s.length ());
+
+  if (output[0].data == nullptr) {
+    throw std::bad_alloc ();
```
myungjoo: This function itself is a thread. Do not throw. Handle all exceptions in this thread.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -135,8 +144,11 @@ typedef struct _GstTensorFilterProperties
...
   uint32_t suspend; /**< Timeout (ms) for suspend. (Unload the framework) */
+
+  NNSFilterInvokeAsyncCallback invoke_async_callback;
+  void *async_handle; /**< handle of Framework */
```
myungjoo: Provide more info on these two new properties. You need to prevent misuse of other developers in the future.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -591,6 +603,45 @@ extern void
...
+ *        Sets callback and the handle for asynchronous operations.
+ *
+ * This function is used when the sub-plugin receives an input and generates multiple outputs asynchronously.
+ *
+ * @param[in] callback The callback function to be invoked during async operations.
+ * @param[in] prop GstTensorFilterProperties object.
+ * @param[in] handle The handle associated with async operations.
+ */
+extern void
+nnstreamer_filter_enable_invoke_async (NNSFilterInvokeAsyncCallback callback, GstTensorFilterProperties * prop, void *handle);
```
myungjoo: Who's the user of this new API? (justify why this needs to be in filter API)

songgot: The function is in ```tensor_common.c``` and is used by tensor_filter and tensor_single.
it is necessary to move to ```tensor_common.h```.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
songgot: Okay, thank you.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
songgot: Updated, thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -351,6 +351,11 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
   if (!input || !input[0].data || input[0].size == 0) {
     throw std::invalid_argument ("Invalid input tensor data");
   }
+  if (!prop->invoke_async) {
+    throw std::runtime_error (
```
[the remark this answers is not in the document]
songgot: Okay, I will support both mode.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Okay, I will fix this. Thank you.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
songgot: Thank you. I've updated the detailed information.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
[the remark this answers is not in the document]
songgot: I've updated, thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I've updated, thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -322,12 +317,17 @@ TensorFilterLlamaCpp::generateTokens (GstTensorFilterProperties *prop,
       throw std::invalid_argument ("Failed to convert token to piece");
     }
 
-    output_async = createOutputTensor (buf, n, prop);
-    if (!output_async)
-      throw std::runtime_error ("generateTokens: Failed to create output tensor");
+    if (prop->invoke_async) {
+      output_async = g_new0 (GstTensorMemory, 1);
+      createOutputTensor (std::string (buf, n), output_async, prop);
+      if (!output_async)
```
jaeyun-jung, line 347: createOutputTensor() after checking allocation failure.

songgot: Thank you, I've updated it.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -214,26 +245,78 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
 }
 
 /**
- * @brief Invoke llamacpp using input tensors
+ * @brief Create output tensor
+ *
+ * The memory allocated for the output tensor data using strndup
+ * must be freed by the caller of this function to avoid memory leaks.
+ */
```
songgot, line 252: Need to modify

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I've updated it. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -277,13 +273,37 @@ TensorFilterLlamaCpp::createOutputTensor (
...
+}
+
+/**
+ * @brief Prepare output tensor for invoke_async
+ */
+GstTensorMemory *
+TensorFilterLlamaCpp::prepareOutputAsync (
+    const std::string &tokenStr, GstTensorFilterProperties *prop)
+{
+  GstTensorMemory *output = g_new0 (GstTensorMemory, 1);
```
gichan-jang: Maybe you intended `g_try_new0`?

gichan-jang (approved):
LGTM except minor comment for dead code.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: I've updated it. Thank you.

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
```diff
@@ -146,8 +146,9 @@ typedef struct _GstTensorFilterProperties
   int invoke_dynamic; /**< True for supporting invoke with flexible output. */
   uint32_t suspend; /**< Timeout (ms) for suspend. (Unload the framework) */
 
-  NNSFilterInvokeAsyncCallback invoke_async_callback; /**< If a filter is configured to emit output frames asynchronously (e.g., N output frames for 1 input frame), tensor_filter framework implementation sets and uses this. */
-  void *async_handle; /**< Internal data structure for invoke_async_callback, configured by tensor_filter framework */
+  int invoke_async; /**< The sub-plugin must support asynchronous output to use this option. If set to TRUE, the sub-plugin can generate multiple outputs asynchronously per single input. Otherwise, only synchronous single-output is expected and async callback/handle are ignored. */
```
myungjoo: (repost) This breaks ABI. Move new elements to the end.

songgot:
```
int invoke_async ,
NNSFilterInvokeAsyncCallback invoke_async_callback,
void *async_handle
```
 are added to the end of elements in this PR.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -420,11 +452,20 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
...
-    input_queue.push (std::make_pair (prop, std::move (prompt)));
+  if (prop->invoke_async) {
+    {
+      std::lock_guard<std::mutex> lock (queue_mutex);
+      input_queue.push (std::make_pair (prop, std::move (prompt)));
+    }
+    queue_cv.notify_one (); /* notify waiting thread */
+  } else {
+    try {
+      generateTokens (prop, std::move (prompt), output);
```
myungjoo, line 462: This line makes LINE444 (UNUSED(output)) obsolete.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -248,27 +250,21 @@ TensorFilterLlamaCpp::invoke (const GstTensorMemory *input, GstTensorMemory *out
...
-  output[0].data = g_strndup (buf, n);
-  output[0].size = n; /* Stores string length, excluding NULL appended to the end of the string */
   if (output[0].data == nullptr) {
     ml_loge ("createOutputTensor: strndup failed");
-    g_free (output);
-    return nullptr;
+    return false;
   }
 
   gst_tensors_info_free (&prop->output_meta);
```
myungjoo: Do you really need to free meta (gst_tensors_info) every frame? Do you really need to erase tensor name? Reconfiguring dimension does not require free-reinit.

gst/nnstreamer/tensor_filter/tensor_filter.c
```diff
@@ -1903,7 +1904,9 @@ gst_tensor_filter_stop (GstBaseTransform * trans)
     GST_OBJECT_UNLOCK (self);
   }
 
-  gst_tensor_filter_disable_invoke_async (&priv->prop);
+  if (priv->prop.invoke_async)
```
myungjoo: invoke_async can be updated at run-time by users or applications. You need to check if invoke_async thread is running.

E.g., 
```
1. configure invoke_async=true
2. start pipeline
3. configure invoke_async=false
4. stop pipeline
```

gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
myungjoo: The quoted code still breaks ABI. New element should be at the end, not at the end-1. ```NNSFilterInvokeAsyncCallback invoke_async_callback``` and ```void *async_handle;``` not not new elements of this PR.

songgot: Thank you. I've updated.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Thank you. I've updated.

gst/nnstreamer/tensor_filter/tensor_filter.c
songgot: Thank you. I removed the if condition.

Now
```
Summary of invoke_async Behavior in tensor_filter:
1. Pipeline Start (gst_tensor_filter_start):
- If invoke_async=true:
  The subplugin (e.g., llamacpp) creates a dedicated thread.
  Asynchronous invocation callbacks are enabled for non-blocking model inference.
- If invoke_async=false:
  No thread is created, and all operations run synchronously.

2. Pipeline Stop (gst_tensor_filter_stop):
- Cleanup is unconditional:
   gst_tensor_filter_disable_invoke_async() is called to terminate any active async threads, regardless of the current invoke_async value.
   Ensures safety even if invoke_async was modified mid-pipeline.
- Subplugin cleanup:
  fini_filter_llamacpp() releases all subplugin resources (threads, memory).
3. Note:
- Runtime changes to invoke_async are ignored:
- Modifying invoke_async while the pipeline is running has no effect.
- To apply changes, stop the pipeline first, update the property, then restart.

4. Workflow Example:
[Pipeline Start]  
├─ Case 1: `invoke_async=true`  
│   ├─ Subplugin thread created  
│   └─ Async callbacks enabled  
│  
├─ Case 2: `invoke_async=false`  
│   ├─ No thread (synchronous execution)  
│  
[Pipeline Stop]  
├─ Force-terminate async threads (`gst_tensor_filter_disable_invoke_async()`)  
├─ Release subplugin resources (`fini_filter_llamacpp()`)  
└─ All resources cleaned up  
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Thank you. I've updated.

niley7464 (approved):
LGTM 👍

PR #4746, opened 2025-06-30
meson.build: Support NumPy >= 2.2 header path changes

cythe:
Since NumPy 2.2.0, header files have moved from `numpy/core/include` to `numpy/_core/include` due to the transition to the Meson build system. This patch updates `meson.build` to reflect the new path, ensuring successful builds with newer NumPy versions.

Note: This update does not maintain compatibility with older NumPy (< 2.2.0).

---
# [Template] fix new NumPy versions(>=2.2.0) headers include path

**Self evaluation:**

**How to evaluate:**
1. build with yocto walnascar(5.2)


myungjoo:
Resolved by #4757

PR #4747, opened 2025-07-01
[Common] id in aggregation

jaeyun-jung:
Fix index value type in tensor-aggregation.


jaeyun-jung:
> 32b key is not enough? :)

GstMetaQuery has client-id with int64_t type, so need to update this :)

PR #4751, opened 2025-07-04
[Filter] invalid return in sub-plugins

jaeyun-jung:
Update return value of callbacks in tensor-filter sub-plugins.


again4you (approved):
LGTM!

myungjoo (approved):

PR #4752, opened 2025-07-07
[spec] Enable test-devel package for robot profile

again4you:
This patch enables the nnstreamer-test-devel package when building the robot profile.

**Self evaluation:**

### DIFF
--- a/packaging/nnstreamer.spec
+++ b/packaging/nnstreamer.spec
@@ -132,7 +132,6 @@
 # Disable a few features for movable device
 %if 0%{?mv_prj}
 %define		grpc_support 0
-%define		check_test 0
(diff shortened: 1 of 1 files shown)


packaging/nnstreamer.spec
```diff
@@ -636,6 +635,7 @@ Requires:   nnstreamer-single-devel = %{version}-%{release}
 %description single-devel-static
 Static library package of nnstreamer-single-devel.
 
+%if 0%{?release_test}
```
jaeyun-jung, line 640: need condition at line 424? `Conflicts: nnstreamer-test-devel`

again4you: ACK!

PR #4758, opened 2025-07-23
[Common] caps with empty tensor info

jaeyun-jung:
Check empty info string when getting tensor caps and remove unnecessary code.

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -1278,16 +1278,14 @@ gst_tensor_pad_caps_from_config (GstPad * pad, const GstTensorsConfig * config)
 
 intersectable:
   if (!gst_caps_can_intersect (caps, templ)) {
-    gst_caps_unref (caps);
-    caps = NULL;
+    g_clear_pointer (&caps, gst_caps_unref);
   }
 
 done:
-  gst_caps_unref (templ);
-  if (peer_caps)
-    gst_caps_unref (peer_caps);
-  caps = gst_caps_truncate (caps);
-  return caps;
+  g_clear_pointer (&templ, gst_caps_unref);
+  g_clear_pointer (&peer_caps, gst_caps_unref);
+
+  return gst_caps_truncate (caps);
 }
 
 /**
@@ -1338,11 +1336,10 @@ gst_tensor_pad_possible_caps_from_config (GstPad * pad,
 
   /* if no possible caps for given config, return null. */
   if (gst_caps_is_empty (caps)) {
-    gst_caps_unref (caps);
-    caps = NULL;
+    g_clear_pointer (&caps, gst_caps_unref);
   }
 
-  gst_caps_unref (templ);
+  g_clear_pointer (&templ, gst_caps_unref);
   return caps;
 }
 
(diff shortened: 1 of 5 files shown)


gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -825,6 +825,32 @@ _append_prev_caps (const GstTensorsConfig * config)
...
+/**
+ * @brief Internal function to check tensors-info string includes item (dimension, type, or name).
+ */
+static gboolean
+_is_empty_info_string (const gchar * str)
+{
+  gchar **str_array;
+  guint i, num;
+  gboolean is_empty = TRUE;
+
```
songgot, line 835: How about adding? 

```
if (str == NULL)
   return TRUE;
```

jaeyun-jung: internally returns true(empty string) if str is null (g_strsplit() returns null if str is null)

gst/nnstreamer/nnstreamer_plugin_api_impl.c
```diff
@@ -927,29 +953,44 @@ _get_tensors_caps (const GstTensorsConfig * config)
...
-      gst_caps_set_simple (prev2, "types", G_TYPE_STRING, type_str, NULL);
+      if (has_dim) {
+        g_autofree gchar *dimstr =
+            gst_tensors_info_get_rank_dimensions_string (&config->info,
+            NNS_TENSOR_RANK_LIMIT_PREV, TRUE);
+
+        gst_caps_set_simple (prev2, "dimensions", G_TYPE_STRING, dimstr, NULL);
+      }
+      if (has_type) {
+        gst_caps_set_simple (prev2, "types", G_TYPE_STRING, type_str, NULL);
```
songgot, line 996: Don't prev1 and prev2 need to be freed after use?

gst/nnstreamer/nnstreamer_plugin_api_impl.c
songgot: Then, can't we make it so that it only runs from line 839 onwards when str_array is null?
```
if (!str_array)
  return TRUE;
```

gst/nnstreamer/nnstreamer_plugin_api_impl.c
jaeyun-jung: those are merged into caps.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
jaeyun-jung: well, actually incoming string would not be null if caps contains its field.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
songgot: Thank you. I made a mistake.

gst/nnstreamer/nnstreamer_plugin_api_impl.c
songgot: Okay, thank you.

songgot (approved):
LGTM

again4you (approved):
LGTM!

PR #4762, opened 2025-08-04
[Build] url to get tizen tools

jaeyun-jung:
Fix url to download and build tizen package.


jaeyun-jung:
The path to Tizen tools is not changed, close this PR.

PR #4763, opened 2025-08-08
[Filter/Sub] util to get nth info

jaeyun-jung:
Code clean,
1. use util to get nth tensor info.
2. clear old tensor info and remove unnecessary code.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
@@ -136,6 +136,8 @@ void
 executorch_llama_subplugin::invoke_dynamic (GstTensorFilterProperties *prop,
     const GstTensorMemory *input, GstTensorMemory *output)
 {
+  GstTensorInfo *_info;
+
   if (!input)
     throw std::runtime_error ("Invalid input buffer, it is NULL.");
   if (!output)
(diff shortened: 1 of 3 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_executorch_llama.cc
```diff
@@ -157,11 +159,14 @@ executorch_llama_subplugin::invoke_dynamic (GstTensorFilterProperties *prop,
...
 
-  gst_tensors_info_init (&prop->output_meta);
+  gst_tensors_info_free (&prop->output_meta);
+
   prop->output_meta.num_tensors = 1;
   prop->output_meta.format = _NNS_TENSOR_FORMAT_FLEXIBLE;
-  prop->output_meta.info[0].type = _NNS_UINT8;
-  prop->output_meta.info[0].dimension[0] = result.size ();
+
+  _info = gst_tensors_info_get_nth_info (&prop->output_meta, 0);
```
songgot, line 167: Shouldn't we initialize _info to NULL and check if gst_tensors_info_get_nth_info() returns NULL?

jaeyun-jung: It couldn't be, if it is null (mem allocation failure or other error) this should be assertion case.

hj210 (approved):
LGTM

anyj0527 (approved):
LGTM✔

PR #4764, opened 2025-08-11
[Filter/Llamacpp] fix mem leak

jaeyun-jung:
Revise llamacpp subplugin to reduce duplicated code and fix mem leak when pushing async output to tensor-filter.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -92,12 +92,12 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   std::condition_variable queue_cv;
   std::atomic<bool> stop_thread{ false }; /* flag to stop the thread */
   void parseCustomProperties (const GstTensorFilterProperties *prop);
-  void generateTokens (GstTensorFilterProperties *prop, std::string &&prompt,
-      GstTensorMemory *output);
-  bool createOutputTensor (const std::string &buf, GstTensorMemory *output,
-      GstTensorFilterProperties *prop);
-  GstTensorMemory *prepareOutputAsync (
-      const std::string &tokenStr, GstTensorFilterProperties *prop);
+  void generateTokens (GstTensorFilterProperties *prop, GstTensorMemory *output,
+      const std::string &prompt);
+  bool createOutputTensor (GstTensorFilterProperties *prop,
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -452,22 +434,13 @@ TensorFilterLlamaCpp::invoke_dynamic (GstTensorFilterProperties *prop,
   prompt.erase (prompt.find_last_not_of (" \t\n\r\f\v") + 1);
 
   if (prop->invoke_async) {
-
-    if (prop->invoke_async_callback == nullptr) {
-      throw std::invalid_argument ("invoke_async_callback is null");
-    }
```
songgot: In asynchronous mode (invoke_async=true), the existence of the callback function (invoke_async_callback) is not checked. Is this okay?

jaeyun-jung: IMO it is not sub-plugins's role, tensor-filter should set the internal callback and handle it.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -277,25 +277,36 @@ TensorFilterLlamaCpp::createOutputTensor (const std::string &buf,
...
 {
-  GstTensorMemory *output = g_try_new0 (GstTensorMemory, 1);
+  const llama_vocab *vocab = llama_model_get_vocab (model);
+  char buf[128] = { 0 };
+  int n = llama_token_to_piece (vocab, token, buf, sizeof (buf), 0, true);
 
-  if (!output) {
-    ml_loge ("Memory allocation failed");
-    return nullptr;
+  if (n < 0) {
```
hj210, line 295: Is it not necessary to check n is 0? For example 'if (n <= 0)'

jaeyun-jung: It just checks error code, converting token id to piece.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Okay. Thank you.

songgot (approved):
LGTM

PR #4760, opened 2025-07-29
[Common] tensor data callback

jaeyun-jung:
1. Define callback function for tensor data stream.
2. Revise async-invoke of tensor-filter.

PR #4766, opened 2025-08-11
[Filter] validate properties

jaeyun-jung:
Add util function to validate filter properties before starting the pipeline.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_common.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_common.c
@@ -1885,7 +1885,6 @@ _gtfc_setprop_PROP_INVOKE_DYNAMIC (GstTensorFilterPrivate * priv,
     const GValue * value)
 {
   priv->prop.invoke_dynamic = g_value_get_boolean (value);
-  priv->info.allocate_in_invoke = TRUE;
 
   return 0;
 }
@@ -2214,7 +2213,7 @@ gst_tensor_filter_common_get_property (GstTensorFilterPrivate * priv,
       g_value_set_boolean (value, prop->invoke_dynamic);
       break;
     case PROP_INVOKE_ASYNC:
-      g_value_set_boolean (value, priv->prop.invoke_async);
+      g_value_set_boolean (value, prop->invoke_async);
       break;
     case PROP_SUSPEND:
       g_value_set_uint (value, prop->suspend);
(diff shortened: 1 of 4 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_common.c
```diff
@@ -2366,8 +2365,29 @@ gst_tensor_filter_common_get_out_info (GstTensorFilterPrivate * priv,
...
 
 /**
- * @brief Load tensor info from NN model.
- * (both input and output tensor)
+ * @brief Check the properties are valid.
+ */
+gboolean
+gst_tensor_filter_validate_prop (GstTensorFilterPrivate * priv)
+{
+  if (priv->prop.invoke_async) {
```
songgot, line 2373: How about checking for null before using ```priv```?

jaeyun-jung: It should not be null. We did not check priv ptr in all util functions of tensor-filter.

tests/nnstreamer_filter_single/unittest_filter_single.cc
```diff
@@ -586,22 +586,25 @@ class NNSFilterSingleInvokeAsyncTest : public ::testing::Test
...
-  callback_count++;
-  g_print ("%s", (char *) output->data);
-  g_free (output->data);
+  /* Track the number of async callback invocations. */
+  if (user_data) {
+    guint *callback_count = (guint *) user_data;
+    (*callback_count)++;
+  }
+
+  g_print ("%s", (char *) data->data);
```
hj210: Is it not necessary to check if data->data is not null? 
For example,
if (data->data) { ... }

tests/nnstreamer_filter_single/unittest_filter_single.cc
```diff
@@ -615,14 +618,17 @@ TEST_F (NNSFilterSingleInvokeAsyncTest, invokeAsync_p)
   }
 
   ASSERT_TRUE (this->loaded);
-  callback_count = 0;
 
-  EXPECT_TRUE (klass->set_invoke_async_callback (single, _invoke_async_callback, klass));
+  guint *count = g_new0 (guint, 1);
```
hj210: I have a question. Is the size of the 'guint' not known at compile time?

jaeyun-jung: It is 'unsigned int', we can get its size.

tests/nnstreamer_filter_single/unittest_filter_single.cc
jaeyun-jung: I would not be null. However when data->data is null, this prints (null) and no crash occurs.

tests/nnstreamer_filter_single/unittest_filter_single.cc
hj210: Could you kindly explain why you are dynamically allocating count instead of defining it directly? For example, guint count = 0

tests/nnstreamer_filter_single/unittest_filter_single.cc
jaeyun-jung: I think it is better adding condition to check data and info in async callback.
I will update this on #4760

tests/nnstreamer_filter_single/unittest_filter_single.cc
jaeyun-jung: well, no reason :) 
I just wanted to pass allocated data to async callback.

gst/nnstreamer/tensor_filter/tensor_filter_common.c
songgot: Thank you.

songgot (approved):
LGTM

PR #4772, opened 2025-09-17
[llama.cpp] Apply KV cache trimming and add context length property

songgot:
- When an input exceeds the remaining context window, KV cache trimming is performed.
    - Add context_length property
    - Modify to maintain context for continuous conversations

**Self evaluation:**


myungjoo (approved):

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -175,6 +187,16 @@ TensorFilterLlamaCpp::parseCustomProperties (const GstTensorFilterProperties *pr
         n_gpu_layers = (int) g_ascii_strtoull (option.get ()[1], NULL, 10);
       } else if (g_ascii_strcasecmp (option.get ()[0], "num_predict") == 0) {
         n_predict = (int) g_ascii_strtoull (option.get ()[1], NULL, 10);
+      } else if (g_ascii_strcasecmp (option.get ()[0], "context_length") == 0) {
```
hj210, line 195: Regarding parseCustomProperties,
The function is responsible for parsing strings and setting multiple member variables.
As properties are added, this function will become increasingly lengthy and complex.

Following the Long Method and Single Responsibility Principle, this function could be modified as follows. Please refer to this suggestion.

//Suggested code by Cline.
class TensorFilterLlamaCpp : public tensor_filter_subplugin {
  // ...
  private:
    // ...
    void parseCustomProperties (const GstTensorFilterProperties *prop);    
     using PropertySetter = std::function<void(const gchar*)>;    
    std::map<std::string, PropertySetter> property_setters_;    
    void initializePropertySetters(); 
    // ...
};

void TensorFilterLlamaCpp::initializePropertySetters() {
    property_setters_["num_gpu_layers"] = [this](const gchar* value) {
        this->n_gpu_layers = static_cast<int>(g_ascii_strtoull(value, NULL, 10));
    };
    property_setters_["num_predict"] = [this](const gchar* value) {
        this->n_predict = static_cast<int>(g_ascii_strtoull(value, NULL, 10));
    };
    ......
}

TensorFilterLlamaCpp::TensorFilterLlamaCpp()   
    initializePropertySetters();
}

void
TensorFilterLlamaCpp::parseCustomProperties (const GstTensorFilterProperties *prop)
{
  for (i = 0; i < len; ++i) {
      ......     
      std::string prop_name = g_ascii_strdown(option.get()[0], -1);
      auto it = property_setters_.find(prop_name);

      if (it != property_setters_.end()) {
        it->second(option.get()[1]);
      } else {
        throw std::invalid_argument (
            "Unknown custom property " + std::string (option.get ()[0]));
      }
    }
}

Advantages of above code
-High extensibility when adding new properties - only needs one additional line in initializePropertySetters (Open/Closed Principle).
-The parseCustomProperties function becomes much more concise and clear.
-Each property's parsing logic is separated, making maintenance easier

songgot: That's a really good suggestion :+1: . I'll reflect it in a new PR. Thank you.

songgot: I've updated. Thank you.
https://github.com/nnstreamer/nnstreamer/pull/4780

PR #4774, opened 2025-09-23
[llama.cpp] Add context save/load functionality for context continuity

songgot:
Add file-based context persistence for context continuity.
Users can now save and load context state using 'save_ctx' and 'load_ctx' custom properties.

**TC**
- contextSaveLoad_p
```
data_push ("Hello my name is John. I like programming and artificial intelligence.");
>> data_push ("Hello my name is John. I like programming and artificial intelligence.");
save context.bin
destroy pipeline

crate pipeline
load context.bin
data_push ("What did I just say about myself?");
>> You said: "Hello my name is John. I like programming and artificial intelligence. I
```
- contextFileNotFound_n
```
// Should work normally with fresh context
>> I am a 35 year old man from
```

**Self evaluation:**

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -23,6 +23,8 @@
 #include <atomic>
 #include <condition_variable>
 #include <errno.h>
+#include <filesystem>
+#include <fstream>
 #include <functional>
 #include <memory>
 #include <nnstreamer_cppplugin_api_filter.hh>
@@ -100,6 +102,9 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   float top_p = TOP_P_DISABLED;
   float typical_p = TYPICAL_P_DISABLED;
   float temperature = TEMPERATURE_DISABLED;
+  std::string lora_path;
+  std::string save_ctx_path;
+  std::string load_ctx_path;
 
   std::thread output_thread; /* thread for async output */
   std::queue<std::pair<GstTensorFilterProperties *, std::string>> input_queue;
@@ -115,6 +120,8 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
       std::string &accumulated);
   void outputThreadLoop ();
   void cleanupResources ();
+  bool loadContextFromFile (const std::string &path);
+  bool saveContextToFile (const std::string &path);
 };
 
 TensorFilterLlamaCpp *TensorFilterLlamaCpp::registered = nullptr;
@@ -134,6 +141,12 @@ TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
  */
 TensorFilterLlamaCpp::~TensorFilterLlamaCpp ()
 {
+  if (!save_ctx_path.empty () && ctx) {
+    if (!saveContextToFile (save_ctx_path)) {
+      ml_loge ("Failed to save context to %s", save_ctx_path.c_str ());
+    }
+  }
+
   cleanupResources ();
 }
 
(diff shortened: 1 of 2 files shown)


myungjoo (approved):
LGTM. (for the 4th commit only)

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -109,6 +113,8 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
       std::string &accumulated);
   void outputThreadLoop ();
   void cleanupResources ();
+  bool loadContextFromFile (const std::string &path);
```
hj210, line 123: The implementation is well done and appears to have high test coverage.
The code could be further improved by considering the following additional points:


The TensorFilterLlamaCpp class appears to be taking on too many responsibilities.
It seems to be handling both tensor filtering operations and context management.
(Please refer to : Single Responsibility Principle)

Suggested Refactoring:
We should separate the context management functionality into a dedicated class:
 
  예) 
  '''cpp
  // Proposed new class
class LlamaContextManager {
public:
  bool saveContext (llama_context *ctx, const std::string &path);
  bool loadContext (llama_context *ctx, const std::string &path);
  bool validateContextFile (const std::string &path);
  
private:
  bool writeContextData (llama_context *ctx, const std::string &path);
  bool readContextData (llama_context *ctx, const std::string &path);
};

// Usage in TensorFilterLlamaCpp class
class TensorFilterLlamaCpp : public tensor_filter_subplugin {
private:
  std::unique_ptr<LlamaContextManager> context_manager;
  // ...
};
```

File: ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc [hunk 2]
```diff
@@ -502,6 +530,95 @@ TensorFilterLlamaCpp::cleanupResources ()
...
+ */
+bool
+TensorFilterLlamaCpp::loadContextFromFile (const std::string &path)
+{
+  if (!ctx) {
+    ml_loge ("Context not initialized for loading state");
+    return false;
+  }
+
+  std::vector<uint8_t> ctx_data;
```
hj210, line 540: The ctx_data vector is created without proper size allocation.
This leads to potential memory access violations.
(Invalid Memory Access)

ctx_data.data() points to invalid memory address.
Attempting to read into unallocated memory is dangerous.
(Undefined Behavior)

The file.read() operation will trigger undefined behavior.
This occurs when trying to read data into uninitialized memory space.

Please consider updating the code after reviewring the sample code below.

  ```cpp 
  std::ifstream file (path, std::ios::binary);
  try {
    file.exceptions (std::ifstream::failbit | std::ifstream::badbit);
     ......
    std::vector<uint8_t> ctx_data (ctx_size);
    file.read (reinterpret_cast<char *> (ctx_data.data ()), ctx_size);
    file.close ();

    if (llama_state_set_data (ctx, ctx_data.data (), ctx_size) != ctx_size) {
      ml_loge ("Failed to set context data");
      return false;
    }

    ml_logi ("Successfully loaded context from %s", path.c_str ());
    return true;

    ml_loge ("Failed to read context file: %s (file: %s)", e.what (), path.c_str ());
    if (file.is_open ()) {
      file.close ();
    }
    return false;
  }
  ```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
hj210: Additionally, I think you can consider applying the State Pattern to manage the context.

Example Code:
```cpp
class ContextPersistenceStrategy {
public:
  virtual ~ContextPersistenceStrategy () = default;
  virtual bool save (llama_context *ctx, const std::string &path) = 0;
  virtual bool load (llama_context *ctx, const std::string &path) = 0;
};

class FileContextPersistence : public ContextPersistenceStrategy {
public:
  bool save (llama_context *ctx, const std::string &path) override;
  bool load (llama_context *ctx, const std::string &path) override;
};

class MemoryContextPersistence : public ContextPersistenceStrategy {
public:
  bool save (llama_context *ctx, const std::string &path) override;
  bool load (llama_context *ctx, const std::string &path) override;
};

songgot: That's a good idea. I'll consider it.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Good idea.
But the current version has no issue because ctx_data.data is valid data.

songgot:
> @songgot Please rebase.

Thank you. Done.

Issue #4755, opened 2025-07-07
Bug report (generated) for tensor-filters (tizen-hal added)


myungjoo:
# C/C++ Source Code Bug Analysis Report - ext/nnstreamer/tensor_filter

This report analyzes C/C++ source code in `/ext/nnstreamer/tensor_filter` for correctness bugs and memory issues.

## Summary

After analyzing multiple tensor filter implementation files, several critical categories of bugs were identified:
- Memory management issues (leaks, premature deallocations)
- Resource cleanup problems in error paths
- Python object lifecycle management bugs
- CUDA/GPU resource management issues
- Device handle management problems
- HAL parameter resource leaks
- Logic errors in error handling
- Missing null pointer checks

---

### ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c

**Bug #1: Memory Leak in Custom Option Parsing**
- **Location**: Lines 153-175 in `nnfw_parse_custom_option()`
- **Issue**: When parsing custom options, the allocated `option` array from `g_strsplit()` is freed, but if the accelerator is set multiple times in options, previous `pdata->accelerator` values leak
- **Risk**: Memory leak on repeated property setting
- **Fix**: Free existing `pdata->accelerator` before reassigning

**Bug #2: Potential Double-Free in Error Path**
- **Location**: Lines 255-270 in `nnfw_open()` error_exit
- **Issue**: The `nnfw_close()` function is called on error, but `nnfw_close()` checks and frees `pdata->accelerator` without nullifying pointers properly
- **Risk**: Use-after-free or double-free
- **Fix**: Ensure proper null pointer checks and nullification after freeing

**Bug #3: Missing Null Check in Tensor Info**
- **Location**: Lines 460-470 in `nnfw_tensor_info_set()`
- **Issue**: `gst_tensors_info_get_nth_info()` can return NULL but this is not checked before dereferencing
- **Risk**: Null pointer dereference
- **Fix**: Add null check after getting tensor info

### ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc

**Bug #4: Memory Management Issue in XNNPACK Delegate**
- **Location**: Lines 380-400 in `TFLiteInterpreter::invoke()`
- **Issue**: When XNNPACK delegate is used, memory is copied but if `memcpy()` fails or tensor sizes are mismatched, no cleanup is performed for partially processed tensors
- **Risk**: Memory corruption or inconsistent state
[...]

PR #4781, opened 2025-10-21
bugfix : Memory Leak in Custom Option Parsing(issue #4755)

songgot:
In theory, this is a memory leak, but in practice, this scenario does not occur: nnfw_parse_custom_option() is only called from nnfw_open() nnfw_open() always allocates a new pdata, so pdata->accelerator always starts with NULL There are no repeated calls to the same pdata instance But. If we can support "Runtime:accel_A, Runtime:accel_B", code is memory leak

Bug 1: Memory Leak in Custom Option Parsing
Location: Lines 153-175 in nnfw_parse_custom_option()
Issue: When parsing custom options, the allocated option array from g_strsplit() is freed, but if the accelerator is set multiple times in options, previous pdata->accelerator values leak
Risk: Memory leak on repeated property setting
Fix: Free existing pdata->accelerator before reassigning

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
@@ -161,6 +161,12 @@ nnfw_parse_custom_option (const GstTensorFilterProperties * prop,
          * Update description in init_filter_nnfw() when adding new custom option for nnfw.
          */
         if (g_ascii_strcasecmp (option[0], "Runtime") == 0) {
+          if (pdata->accelerator != NULL) {
+            nns_logw
+                ("Trying to set new runtime '%s', previous runtime '%s' is ignored.",
+                option[1], pdata->accelerator);
+            g_free (pdata->accelerator);
+          }
           pdata->accelerator = g_strdup (option[1]);
         } else {
           nns_logw ("Unknown option (%s).", options[i]);


ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -161,6 +161,11 @@ nnfw_parse_custom_option (const GstTensorFilterProperties * prop,
          * Update description in init_filter_nnfw() when adding new custom option for nnfw.
          */
         if (g_ascii_strcasecmp (option[0], "Runtime") == 0) {
+          if (pdata->accelerator != NULL) {
+            nns_logw ("Duplicate Runtime option detected. Using: %s",
```
jaeyun-jung: Suggestion: logging previous runtime string

```
"Trying to set new runtime '%s', previous runtime '%s' is ignored.", option[1], pdata->accelerator
```

songgot: I've updated. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_nnfw.c
```diff
@@ -161,6 +161,12 @@ nnfw_parse_custom_option (const GstTensorFilterProperties * prop,
          * Update description in init_filter_nnfw() when adding new custom option for nnfw.
          */
         if (g_ascii_strcasecmp (option[0], "Runtime") == 0) {
+          if (pdata->accelerator != NULL) {
```
hj210, line 164: This patch will fix memory leak issue. (Memory Management Code Smell). LGTM!

hj210 (approved):

PR #4782, opened 2025-10-21
bugfix : Memory management issue in XNNPACK delegate (issue #4755)

songgot:
g_assert is completely removed in release mode.
The assertion tensor_ptr->bytes == input[i].size (comparing tensor sizes) is not a developer logic bug, but rather a data inconsistency that can occur at runtime. This is an error that should be handled with an if statement, not g_assert.

Bug 4: Memory Management Issue in XNNPACK Delegate

Location: Lines 380-400 in TFLiteInterpreter::invoke()
Issue: When XNNPACK delegate is used, memory is copied but if memcpy() fails or tensor sizes are mismatched, no cleanup is performed for partially processed tensors
Risk: Memory corruption or inconsistent state
Fix: Add proper validation before memcpy operations

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -369,7 +369,14 @@ TFLiteInterpreter::invoke (const GstTensorMemory *input, GstTensorMemory *output
   if (is_xnnpack_delegated) {
     for (unsigned int i = 0; i < inputTensorMeta.num_tensors; ++i) {
       tensor_ptr = inputTensorPtr[i];
-      g_assert (tensor_ptr->bytes == input[i].size);
+
+      /* Replace g_assert with explicit error check for release builds */
+      if (tensor_ptr->bytes != input[i].size) {
+        ml_loge ("Tensor size mismatch for input tensor %u: expected %zu, got %zu",
+            i, tensor_ptr->bytes, input[i].size);
+        return -EINVAL;
+      }
+
       memcpy (tensor_ptr->data.raw, input[i].data, input[i].size);
     }
   } else {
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -369,8 +369,19 @@ TFLiteInterpreter::invoke (const GstTensorMemory *input, GstTensorMemory *output
...
+      if (tensor_ptr->bytes != input[i].size) {
+        ml_loge ("Tensor size mismatch for input tensor %u: expected %zu, got %zu",
+            i, tensor_ptr->bytes, input[i].size);
+        return -EINVAL;
+      }
+
+      /* Check memcpy result to handle potential memory copy failures */
+      if (memcpy (tensor_ptr->data.raw, input[i].data, input[i].size) == NULL) {
+        ml_loge ("Memory copy failed for input tensor %u", i);
+        return -ENOMEM;
```
myungjoo: memcpy does not have an error return.
It returns dest pointer. (tensor_ptr->data.raw)

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -398,8 +409,19 @@ TFLiteInterpreter::invoke (const GstTensorMemory *input, GstTensorMemory *output
...
+      if (tensor_ptr->bytes != output[i].size) {
+        ml_loge ("Tensor size mismatch for output tensor %u: expected %zu, got %zu",
+            i, tensor_ptr->bytes, output[i].size);
+        return -EINVAL;
+      }
+
+      /* Check memcpy result to handle potential memory copy failures */
+      if (memcpy (output[i].data, tensor_ptr->data.raw, output[i].size) == NULL) {
+        ml_loge ("Memory copy failed for output tensor %u", i);
+        return -ENOMEM;
```
myungjoo: Don't check return value of memcpy. It does not return NULL for error.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
songgot: Okay, thank you, I've updated.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
songgot: Okay, thank you, I've updated.

hj210 (approved):
LGTM!

PR #4783, opened 2025-10-21
bugfix: Potential memory leak in model loading (issue #4755)

songgot:
After creating a delegate, if ModifyGraphWithDelegate fails, the delegate memory is not released.

Bug 5: Potential Memory Leak in Model Loading

Location: Lines 540-580 in TFLiteInterpreter::loadModel()
Issue: If delegate creation succeeds but interpreter->ModifyGraphWithDelegate() fails, the delegate memory is not properly freed
Risk: Memory leak of delegate resources
Fix: Use RAII or ensure cleanup in error paths

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -643,6 +643,7 @@ TFLiteInterpreter::loadModel (int num_threads, tflite_delegate_e delegate_e)
   if (delegate != nullptr) {
     if (interpreter->ModifyGraphWithDelegate (delegate) != kTfLiteOk) {
       ml_loge ("Failed to apply delegate\n");
+      setDelegate (nullptr, [] (TfLiteDelegate *) {});
       return -2;
     }
   }


ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -643,6 +643,7 @@ TFLiteInterpreter::loadModel (int num_threads, tflite_delegate_e delegate_e)
   if (delegate != nullptr) {
     if (interpreter->ModifyGraphWithDelegate (delegate) != kTfLiteOk) {
       ml_loge ("Failed to apply delegate\n");
+      setDelegate (nullptr, [] (TfLiteDelegate *) {});
```
hj210, line 646: This patch will fix the memory leak issue that occurs when failing to create a delegate. LGTM!

anyj0527, line 646: The delegate is unique_ptr created with custom deleter, thus the deleter would be called in return. So I don't think this line is necessary.

songgot: @anyj0527 , @jaeyun-jung , Thank you for the review.
I overlooked that loadModel is called from the init function. With RAII, the smart pointer is freed.

songgot:
@anyj0527 , @jaeyun-jung , Thank you for the review.
I overlooked that loadModel is called from the init function. With RAII, the smart pointer is freed.

PR #4784, opened 2025-10-21
bugfix: Race Condition in Shared Model Management (issue #4755)

songgot:
If thread A succeeds in nnstreamer_filter_shared_model_get, there is no time for thread B to intervene.
However, if thread A fails in nnstreamer_filter_shared_model_get and unlock, thread B can access the interpreter pointr.

Bug 6: Race Condition in Shared Model Management

Location: Lines 980-1020 in TFLiteCore::checkSharedInterpreter()
Issue: The shared model table operations use a global lock, but the interpreter pointer could be accessed after lock release and before usage
Risk: Use-after-free in multi-threaded scenarios
Fix: Extend lock scope or use reference counting

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -1001,37 +1001,48 @@ TFLiteCore::~TFLiteCore ()
 gboolean
 TFLiteCore::checkSharedInterpreter (const GstTensorFilterProperties *prop)
 {
+  TFLiteInterpreter *temp_interpreter = nullptr;
+  gboolean result = FALSE;
+
   G_LOCK (slock);
-  interpreter = (TFLiteInterpreter *) nnstreamer_filter_shared_model_get (
+  temp_interpreter = (TFLiteInterpreter *) nnstreamer_filter_shared_model_get (
       this, shared_tensor_filter_key);
 
-  if (!interpreter) {
+  if (!temp_interpreter) {
     /* create new interpreter */
     TFLiteInterpreter *new_interpreter = new TFLiteInterpreter ();
-    interpreter = (TFLiteInterpreter *) nnstreamer_filter_shared_model_insert_and_get (
+    temp_interpreter = (TFLiteInterpreter *) nnstreamer_filter_shared_model_insert_and_get (
         this, shared_tensor_filter_key, new_interpreter);
-    if (!interpreter) {
-      G_UNLOCK (slock);
+    if (!temp_interpreter) {
       ml_loge ("Failed to insert the model representation!");
       g_free (shared_tensor_filter_key);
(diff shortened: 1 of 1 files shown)


songgot:
> checkSharedInterpreter() is called by the constructor only. Please elaborate how race condition occurs.

While `checkSharedInterpreter` is only called directly from the constructor, the shared model table is accessed from other functions as well:

- `nnstreamer_filter_shared_model_replace()` is called from the `reloadModel()` function
- `nnstreamer_filter_shared_model_remove()` is called from the destructor
- `nnstreamer_filter_shared_model_get()` is called from the `reloadModel()` function

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -1001,37 +1001,54 @@ TFLiteCore::~TFLiteCore ()
...
         this, shared_tensor_filter_key, new_interpreter);
-    if (!interpreter) {
-      G_UNLOCK (slock);
+    if (!temp_interpreter) {
       ml_loge ("Failed to insert the model representation!");
       g_free (shared_tensor_filter_key);
       shared_tensor_filter_key = NULL;
       delete new_interpreter;
-      return FALSE;
+      result = FALSE;
```
jaeyun-jung: line 1022 and 1035 (result = false) unnecessary lines.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
```diff
@@ -1001,37 +1001,54 @@ TFLiteCore::~TFLiteCore ()
...
   /* shared model exists */
-  else if (g_strcmp0 (prop->model_files[0], interpreter->getModelPath ()) != 0) {
-    ml_logw ("The model paths are not equal, models are not shared.");
-    nnstreamer_filter_shared_model_remove (this, shared_tensor_filter_key, free_interpreter);
-    G_UNLOCK (slock);
-    g_free (shared_tensor_filter_key);
-    shared_tensor_filter_key = NULL;
-    return FALSE;
+  else {
+    model_path_copy = g_strdup (temp_interpreter->getModelPath ());
```
jaeyun-jung: Q: why do you make a copy of model path? (it is possible to compare model path using temp_interpreter->getModelPath ())

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
songgot: Thank you. I have fixed it.

ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
songgot: Thank you. I have fixed it.

jaeyun-jung (approved):

PR #4787, opened 2025-10-22
bugfix: Incomplete Error Handling in Run Method (issue #4755)

songgot:
If, for some reason, a Python array object is not properly created in the middle of a loop, or if type/size checks fail (e.g., line 591), an error will occur and the loop will abort.
Therefore, as pointed out in the bug report, if an error occurs within the PYCore::run() function, previously created arrays must be cleaned up.

Bug 10: Incomplete Error Handling in Run Method

Location: Lines 570-620 in PYCore::run()
Issue: If output array creation fails midway through the loop, previous arrays in the map are not cleaned up properly, leading to reference leaks
Risk: Python object reference leaks
Fix: Clean up partial results on failure

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
@@ -50,6 +50,7 @@
 #include <nnstreamer_cppplugin_api_filter.hh>
 #include <nnstreamer_log.h>
 #include <nnstreamer_util.h>
+#include <vector>
 #include "nnstreamer_python3_helper.h"
 
 /**
@@ -541,6 +542,7 @@ PYCore::run (const GstTensorMemory *input, GstTensorMemory *output)
   GstTensorInfo *_info;
   int res = 0;
   PyObject *result;
+  std::vector<void *> newlyAddedItems;
 
 #if (DBG)
   gint64 start_time = g_get_real_time ();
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -593,7 +593,12 @@ PYCore::run (const GstTensorMemory *input, GstTensorMemory *output)
         break;
       }
     }
-
+    if (res != 0) {
+      for (auto &item : outputArrayMap) {
+        Py_SAFEDECREF (item.second);
+      }
+      outputArrayMap.clear ();
```
myungjoo: Is it really ok to remove ALL from outputArrayMap?
or should be remove items inserted in this function only? (what is there is an outputArrayMap entry owned by a caller, which will call eventHandler later?)

songgot: You're right. I'll fix it to only delete the problem part of outputArrayMap.

songgot: Thank you, I've updated.

myungjoo (approved):

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -587,13 +589,23 @@ PYCore::run (const GstTensorMemory *input, GstTensorMemory *output)
...
         ml_loge ("Output tensor type/size is not matched\n");
         res = -2;
         break;
       }
     }
-
+    if (res != 0) {
+      // Clean up only items added in this function call
+      for (void *data : newlyAddedItems) {
+        auto it = outputArrayMap.find (data);
```
jaeyun-jung: suggest calling freeOutputTensors (data) instead, it is already implemented.

songgot: Thank you. I have updated it.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -587,13 +589,19 @@ PYCore::run (const GstTensorMemory *input, GstTensorMemory *output)
...
+        newlyAddedItems.push_back (output[i].data);
       } else {
         ml_loge ("Output tensor type/size is not matched\n");
         res = -2;
         break;
       }
     }
-
+    if (res != 0) {
+      // Clean up only items added in this function call
```
jaeyun-jung: minor, let's keep line-comment rule.
`/* Clean up only items added in this function call */`

jaeyun-jung (approved):

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
songgot: Thank you.

PR #3889, opened 2022-08-19
test:python3 multithreaded python script as a filter (CV2)

myungjoo:
With the fear of regressions related with #3822 as
described in #3885, added a test case for multithreaded python
script of CV2.

TODO: add similar cases for converter and decoder
TODO: add negative cases

Fixes #3885

Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>

PR #4788, opened 2025-10-23
bugfix: Threading issue with Python GIL (issue #4755)

songgot:
- Multiple functions acquire/release GIL but if an exception occurs between lock/unlock, GIL state becomes inconsistent.
- Solved using RAII pattern

Bug 11: Threading Issue with Python GIL

Location: Throughout the Python filter
Issue: Multiple functions acquire/release GIL but if an exception occurs between lock/unlock, GIL state becomes inconsistent
Risk: Deadlock or Python interpreter corruption
Fix: Use RAII pattern for GIL management

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
@@ -192,13 +192,11 @@ PYCore::~PYCore ()
   gst_tensors_info_free (&inputTensorMeta);
   gst_tensors_info_free (&outputTensorMeta);
 
-  Py_LOCK ();
   Py_SAFEDECREF (core_obj);
   Py_SAFEDECREF (shape_cls);
-
   PyErr_Clear ();
-  Py_UNLOCK ();
 
+  g_mutex_clear (&py_mutex);
   dlclose (handle);
 }
 
@@ -674,9 +672,8 @@ TensorFilterPython::TensorFilterPython () : core (nullptr)
 TensorFilterPython::~TensorFilterPython ()
 {
   if (core != nullptr) {
-    PyGILState_STATE gstate = PyGILState_Ensure ();
+    PyGILGuard gil_guard;
     delete core;
-    PyGILState_Release (gstate);
   }
 }
 
@@ -712,22 +709,20 @@ TensorFilterPython::configure_instance (const GstTensorFilterProperties *prop)
       return; /* skipped */
     }
 
-    PyGILState_STATE gstate = PyGILState_Ensure ();
+    PyGILGuard gil_guard;
     delete core;
-    PyGILState_Release (gstate);
   }
 
-  PyGILState_STATE gstate = PyGILState_Ensure ();
+  PyGILGuard gil_guard;
   core = new PYCore (script_path, prop->custom_properties);
   if (core == nullptr) {
     g_printerr ("Failed to allocate memory for filter subplugin: Python\n");
-    goto done;
+    return;
   }
 
   if (core->init (prop) != 0) {
     delete core;
     g_printerr ("failed to initialize the object: Python\n");
-    PyGILState_Release (gstate);
     throw std::runtime_error ("Python is not initialize");
   }
 
@@ -736,11 +731,8 @@ TensorFilterPython::configure_instance (const GstTensorFilterProperties *prop)
   if (cb != cb_type::CB_SETDIM && cb != cb_type::CB_GETDIM) {
     delete core;
     g_printerr ("Wrong callback type\n");
-    goto done;
+    return;
   }
-
-done:
-  PyGILState_Release (gstate);
 }
 
 /**
@@ -749,9 +741,8 @@ done:
 void
 TensorFilterPython::invoke (const GstTensorMemory *input, GstTensorMemory *output)
 {
-  PyGILState_STATE gstate = PyGILState_Ensure ();
+  PyGILGuard gil_guard;
   core->run (input, output);
-  PyGILState_Release (gstate);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -191,12 +201,11 @@ PYCore::~PYCore ()
   gst_tensors_info_free (&inputTensorMeta);
   gst_tensors_info_free (&outputTensorMeta);
 
-  Py_LOCK ();
+  PyGILGuard gil_guard;
```
myungjoo: Possible error.
With dlclose, you don't have link to Python C APIs, but your code will call PyGILRelease after dlclose.
Note that  Py_LOCK/UNLOCK does not call any Python C APIs.

myungjoo:
Enable the testcase, "4-1 (IGNORED) " from #3889 to test this code.

```
[  595s] ==================================================
[  595s]     Test Group nnstreamer_filter_python3 Starts.
[  595s] [PASSED] 1:gst-launch of case 1
[  595s] [PASSED] 1:Compare 1
[  595s] [PASSED] 2:gst-launch of case 2
[  595s] [PASSED] 2:Golden test comparison
[  595s] [PASSED] 3:gst-launch of case 3
[  595s] [PASSED] 3:Golden test comparison
[  595s] [IGNORED] 4-1:gst-launch of case 4-1
[  595s] [PASSED] 4-2:Multithreaded python script as a filter (CV2)
[  595s] ==================================================
```

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
songgot: I've updated. Thank you.

myungjoo:
> Code Flow: cv2 installed: CV2=0 → IGNORE=0 → 4-1 executed, 4-2 compared. cv2 not installed: CV2=1 → IGNORE=1 → 4-1 ignored, 4-2 passed (comparison skipped). Therefore, the fact that 4-1 is ignored means that cv2 is not installed on the system.

Then, enable it by updating workflow and prevent regression.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -731,7 +727,8 @@ TensorFilterPython::configure_instance (const GstTensorFilterProperties *prop)
   }
 
 done:
-  PyGILState_Release (gstate);
+  /* PyGILGuard destructor automatically releases GIL */
+  return;
```
jaeyun-jung: We dont need goto done, just return.

songgot: Nice catch :+1: , Thank you. I've updated.

songgot:
> > Code Flow: cv2 installed: CV2=0 → IGNORE=0 → 4-1 executed, 4-2 compared. cv2 not installed: CV2=1 → IGNORE=1 → 4-1 ignored, 4-2 passed (comparison skipped). Therefore, the fact that 4-1 is ignored means that cv2 is not installed on the system.
> 
> Then, enable it by updating workflow and prevent regression.

Okay, I will try to do this in another PR. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -81,6 +81,27 @@ enum class cb_type {
   CB_END,
 };
 
+/**
+ * @brief RAII wrapper for managing Python GIL.
+ */
+class PyGILGuard
```
jaeyun-jung: plz move this class to common python-util  (It would be a new header file)
and update converter/decoder subplugin for python also.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -192,13 +202,12 @@ PYCore::~PYCore ()
...
 
-  Py_LOCK ();
-  Py_SAFEDECREF (core_obj);
-  Py_SAFEDECREF (shape_cls);
-
-  PyErr_Clear ();
-  Py_UNLOCK ();
-
+  {
+    PyGILGuard gil_guard;
```
jaeyun-jung: you are calling deletion with gil, and set gil in destructor again.
```
PyGILGuard gil_guard;
delete core;
```

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -210,8 +219,9 @@ int
 PYCore::init (const GstTensorFilterProperties *prop)
 {
   int ret = -EINVAL;
+  PyGILGuard gil_guard;
```
jaeyun-jung: I suggest using mutex-lock in py-core, and gil outside of py-core.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
songgot: How about using 'nnstreamer_python3_helper.h'?

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
songgot: Okay, I will update.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
songgot: Thank you, I agree with you.

ext/nnstreamer/tensor_filter/tensor_filter_python3.cc
```diff
@@ -518,6 +516,7 @@ PYCore::setInputTensorDim (const GstTensorsInfo *in_info, GstTensorsInfo *out_in
 void
 PYCore::freeOutputTensors (void *data)
 {
+  Py_LOCK ();
```
jaeyun-jung: deadlock occurs here, plz remove mutex-lock here.

songgot: Oops, thank you.

jaeyun-jung (approved):

again4you (approved):
LGTM!

PR #4789, opened 2025-10-23
bugfix: Resource Leak in Tensor Information Extraction (issue #4755)

songgot:
If session creation succeeds but tensor info extraction fails, ONNX runtime objects are not properly released

The Reported Bug is a **False Positive** because of calling cleanup().
Bug 12: Resource Leak in Error Path
=> However, when an exception occurs in the part where tensor info is retrieved, cleanup is not called.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_onnxruntime.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_onnxruntime.cc
@@ -67,7 +67,7 @@ class onnxruntime_subplugin final : public tensor_filter_subplugin
   static const char *name;
   static onnxruntime_subplugin *registeredRepresentation;
 
-  void cleanup ();
+  void cleanup (bool force = false);
   void clearNodeInfo (onnx_node_info_s &node);
   void convertTensorInfo (onnx_node_info_s &node, GstTensorsInfo &info);
   int convertTensorDim (std::vector<int64_t> &shapes, tensor_dim &dim);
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_onnxruntime.cc
```diff
@@ -298,45 +298,51 @@ onnxruntime_subplugin::configure_instance (const GstTensorFilterProperties *prop
...
 
-  memInfo = Ort::MemoryInfo::CreateCpu (
-      OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault);
+    configured = true;
+    allocator = Ort::AllocatorWithDefaultOptions{ nullptr }; /* delete unique_ptr */
 
-  configured = true;
-  allocator = Ort::AllocatorWithDefaultOptions{ nullptr }; /* delete unique_ptr */
+  } catch (...) {
+    cleanup ();
```
myungjoo: Cleanup() has no effect at error handling because configured may be false.
(```if configured == false```, cleanup() does NOTHING.)

songgot: I added a `force` parameter with a default value of `false` to maintain backward compatibility.
I modified the catch block to call `cleanup(true)` for proper resource cleanup.

myungjoo (approved):

PR #4790, opened 2025-10-23
bugfix: CUDA Memory Leak in Error Path (issue #4755)

songgot:
If cudaMallocManaged() succeeds for output buffer but execute() fails, the allocated CUDA memory is not freed.

Bug 14: CUDA Memory Leak in Error Path

Location: Lines 260-290 in invoke()
Issue: If cudaMallocManaged() succeeds for output buffer but execute() fails, the allocated CUDA memory is not freed
Risk: GPU memory leak
Fix: Free CUDA memory on all error paths

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
@@ -235,11 +235,19 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
   std::vector<void *> bindings = { _inputBuffer, output->data };
   if (!_Context->execute (1, bindings.data ())) {
     ml_loge ("Failed to execute the network");
+    cudaFree (output->data);
+    output->data = nullptr;
     throw std::runtime_error ("Failed to execute the network");
   }
 
   /* wait for GPU to finish the inference */
-  cudaDeviceSynchronize ();
+  cudaError_t status = cudaDeviceSynchronize ();
+  if (status != cudaSuccess) {
+    ml_loge ("Failed to synchronize device: %s", cudaGetErrorString (status));
+    cudaFree (output->data);
+    output->data = nullptr;
+    throw std::runtime_error ("Failed to synchronize device");
+  }
 }
 
 /**


ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -225,21 +225,34 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
   /* Copy input data to Cuda memory space */
   memcpy (_inputBuffer, input->data, input->size);
 
+  /* Initialize output buffer pointer to nullptr for safety */
+  output->data = nullptr;
```
myungjoo: why? in allocBuffer, it calls cudaMallocManaged (&output->data, output->size), which effectively does output->data = malloc(). This appears meaningless.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -225,21 +225,34 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
...
+    cudaDeviceSynchronize ();
+
+  } catch (...) {
+    /* Clean up allocated output buffer on any exception */
+    if (output->data != nullptr) {
+      cudaFree (output->data);
+      output->data = nullptr;
+    }
+    throw std::runtime_error ("TensorRT inference failed");
+  }
```
myungjoo, line 250: execute() already does error check. no need for type-catch with more ambiguous error message.

cudaDeviceSynchronize() returns cudaSuccess or error. Use error return; catch won't catch errors here.

Simply add cudaFree at error cases.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: Oh, that's right. I'll fix the code.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: I will make a simple correction as you said.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: I've updated. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: I've updated. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -235,11 +235,19 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
   std::vector<void *> bindings = { _inputBuffer, output->data };
   if (!_Context->execute (1, bindings.data ())) {
     ml_loge ("Failed to execute the network");
+    cudaFree (output->data);
```
myungjoo, line 238: To ensure the safety of output->data[], check the lifecycle of given output. (tensor_filter_common.c, tensor_filter.c/h, tensor_filter_support_cc.cc)

output->data[] might be allocated by the caller. allocBuffer() might be overwriting a valid pointer.

songgot: TensorRT use `allocate_in_invoke = TRUE`:
1. TensorRT allocates and frees memory directly.
2. tensor_filter.c only assists TensorRT in freeing memory.
3. Safe freeing is achieved by chaining `gst_tensor_filter_destroy_notify_util()` → `eventHandler(DESTROY_NOTIFY)` → `cudaFree()`.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -235,11 +235,19 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
...
+    output->data = nullptr;
     throw std::runtime_error ("Failed to execute the network");
   }
 
   /* wait for GPU to finish the inference */
-  cudaDeviceSynchronize ();
+  cudaError_t status = cudaDeviceSynchronize ();
+  if (status != cudaSuccess) {
+    ml_loge ("Failed to synchronize device: %s", cudaGetErrorString (status));
+    cudaFree (output->data);
```
again4you, line 247: If `cudaDeviceSynchronize()` do not returns `cudaSuccess`, then should `cudaDeviceReset()` be called?
In this case, we do not need to call `cudaFree()` since `cudaDeviceReset()` will release all allocated memory including `cudaMallocManaged()`

- [cudaDeviceReset API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217)

songgot: It can return 'cudaSuccess'.
```
__host__​__device__​[cudaError_t](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gf599e5b8b829ce7db0f5216928f6ecb6) cudaDeviceSynchronize ( void )
Wait for compute device to finish.
Returns
[cudaSuccess](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038e355f04607d824883b4a50662830d591)

Description
Blocks until the device has completed all preceding requested tasks. [cudaDeviceSynchronize()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g10e20b05a95f638a4071a655503df25d) returns an error if one of the preceding tasks has failed. If the [cudaDeviceScheduleBlockingSync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g057e6912c52708b6aa86e79dd83d007c) flag was set for this device, the host thread will block until the device has finished its work.


Note:
Use of cudaDeviceSynchronize in device code was deprecated in CUDA 11.6 and removed for compute_90+ compilation. For compute capability < 9.0, compile-time opt-in by specifying -D CUDA_FORCE_CDP1_IF_SUPPORTED is required to continue using [cudaDeviceSynchronize()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g10e20b05a95f638a4071a655503df25d) in device code for now. Note that this is different from host-side cudaDeviceSynchronize, which is still supported.

Note that this function may also return error codes from previous, asynchronous launches.

Note that this function may also return [cudaErrorInitializationError](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038ce7993a88ecf2c57b8102d55d997a18c), [cudaErrorInsufficientDriver](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038f5e52d1774934b77ba55d2aa2c063067) or [cudaErrorNoDevice](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e0038e942e4cbbd2bef6e92e293253f055613) if this call tries to initialize internal CUDA RT state.

Note that as specified by [cudaStreamAddCallback](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498) no CUDA function may be called from callback. [cudaErrorNotPermitted](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg3f51e3575c2178246db0a94a430e003867b6095ab719b21659a400b553963eb3) may, but is not guaranteed to, be returned as a diagnostic in such case.

See also:

[cudaDeviceReset](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gef69dd5c6d0206c2b8d099abac61f217), [cuCtxSynchronize](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CTX.html#group__CUDA__CTX_1g7a54725f28d34b8c6299f0c6ca579616)
```
Since it is an invoke function, it needs to use cudaFree().
```
cudaFree() vs cudaDeviceReset() Summary
1. Scope of Operation
cudaFree(): Targets a single memory block (local operation)
cudaDeviceReset(****): Affects all resources on the device (global operation)
2. Resources Affected
cudaFree(): Only the specified pointer's memory
cudaDeviceReset(****): All memories, streams, events, and other CUDA objects
3. Thread Safety
cudaFree(): Safe for use in multi-threaded environments (affects only the specified resource)
cudaDeviceReset(****): Not thread-safe (disrupts all CUDA operations in the process)
4. Typical Use Cases
cudaFree(): When you need to release specific memory blocks during execution
cudaDeviceReset(****): When cleaning up all resources at program exit or after critical errors
```

PR #4791, opened 2025-10-23
bugfix: Missing CUDA error handling (issue #4755)

songgot:
CUDA operations like cudaDeviceSynchronize() don't check for errors
- "Silent failures" occur when CUDA function calls go undetected.
- GPU task failures are difficult to debug.
- Memory leak risk.

Bug 15: Missing CUDA Error Handling

Location: Lines 300-320 in allocBuffer()
Issue: cudaMallocManaged() failure is checked, but other CUDA operations like cudaDeviceSynchronize() don't check for errors
Risk: Silent CUDA failures
Fix: Check all CUDA operation return values

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
@@ -149,8 +149,12 @@ tensorrt_subplugin::~tensorrt_subplugin ()
   gst_tensors_info_free (&_inputTensorMeta);
   gst_tensors_info_free (&_outputTensorMeta);
 
-  if (_inputBuffer != nullptr)
-    cudaFree (_inputBuffer);
+  if (_inputBuffer != nullptr) {
+    cudaError_t free_status = cudaFree (_inputBuffer);
+    if (free_status != cudaSuccess) {
+      ml_loge ("Failed to free input CUDA buffer: %s", cudaGetErrorString (free_status));
+    }
+  }
 
   if (_uff_path != nullptr)
     g_free (_uff_path);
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -239,7 +243,11 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
   }
 
   /* wait for GPU to finish the inference */
-  cudaDeviceSynchronize ();
+  cudaError_t sync_status = cudaDeviceSynchronize ();
+  if (sync_status != cudaSuccess) {
+    ml_loge ("CUDA device synchronize failed: %s", cudaGetErrorString (sync_status));
+    throw std::runtime_error ("CUDA device synchronize failed");
+  }
```
myungjoo, line 260: A bit of mix-up with #4790. Anyway, add cudafree on output->data is required in this throw.

songgot: Yes, you're right. I've updated. Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -239,7 +243,13 @@ tensorrt_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *outpu
   }
 
   /* wait for GPU to finish the inference */
-  cudaDeviceSynchronize ();
+  cudaError_t sync_status = cudaDeviceSynchronize ();
+  if (sync_status != cudaSuccess) {
+    ml_loge ("CUDA device synchronize failed: %s", cudaGetErrorString (sync_status));
+    cudaFree (output->data);
```
myungjoo: Without checking the return value of this, you will get the same issue.

songgot: Thank you for the review. I've updated.

myungjoo (approved):

PR #4792, opened 2025-10-24
bugfix: Resource Management in Destructor (issue #4755)

songgot:
Avoid calling `cudaFree()` when the CUDA context has already been destroyed.
If cudaFree() is called but CUDA context is already destroyed, it can cause undefined behavior.

Bug 16: Resource Management in Destructor

Location: Lines 130-150 in ~tensorrt_subplugin()
Issue: If cudaFree() is called but CUDA context is already destroyed, it can cause undefined behavior
Risk: Undefined behavior on cleanup
Fix: Check CUDA context validity before operations

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
@@ -149,8 +149,19 @@ tensorrt_subplugin::~tensorrt_subplugin ()
   gst_tensors_info_free (&_inputTensorMeta);
   gst_tensors_info_free (&_outputTensorMeta);
 
-  if (_inputBuffer != nullptr)
-    cudaFree (_inputBuffer);
+  if (_inputBuffer != nullptr) {
+    /* Check CUDA context validity before freeing memory */
+    int deviceCount = 0;
+    cudaError_t status = cudaGetDeviceCount (&deviceCount);
+
+    if (status == cudaSuccess && deviceCount > 0) {
+      cudaError_t freeStatus = cudaFree (_inputBuffer);
+      if (freeStatus != cudaSuccess) {
+        ml_logw ("Failed to free CUDA buffer: %s", cudaGetErrorString (freeStatus));
+      }
+    }
+    _inputBuffer = nullptr;
+  }
 
   if (_uff_path != nullptr)
     g_free (_uff_path);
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -149,8 +149,19 @@ tensorrt_subplugin::~tensorrt_subplugin ()
   gst_tensors_info_free (&_inputTensorMeta);
   gst_tensors_info_free (&_outputTensorMeta);
 
-  if (_inputBuffer != nullptr)
-    cudaFree (_inputBuffer);
+  if (_inputBuffer != nullptr) {
```
myungjoo, line 152: Why do you need to check device count when destructing the object? (Is this code created by LLM?)

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
```diff
@@ -282,7 +293,21 @@ tensorrt_subplugin::eventHandler (event_ops ops, GstTensorFilterFrameworkEventDa
 {
   if (ops == DESTROY_NOTIFY) {
     if (data.data != nullptr) {
-      cudaFree (data.data);
+      /* Check CUDA context validity before freeing memory */
+      int deviceCount = 0;
+      cudaError_t status = cudaGetDeviceCount (&deviceCount);
```
myungjoo, line 298: Why do you need to check this here?

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: To fix the bug, I tried to check the context before closing CUDA.
If deviceCount > 0, this is one way to reasonably infer that the CUDA runtime is still functioning. However, this may not be the most clear approach.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
songgot: To fix the bug, I tried to check the context before closing CUDA.

ext/nnstreamer/tensor_filter/tensor_filter_tensorrt.cc
myungjoo: cudaGetDeviceCount check the number of CUDA devices (e.g., NVIDIA GPUs)

It has nothing to do with runtime.

myungjoo: Plus, the current tensorrt_subplugin::invoke is running synchronously.

songgot: Thank you for the detailed explanation. I agree with your opinion.

songgot:
Thank you for the detailed explanation. I agree with your opinion.

PR #4793, opened 2025-10-24
bugfix: Device handle leak on failure (issue #4755)

songgot:
If device creation succeeds but graph creation fails, only `ncDeviceDestroy()` is called. `ncDeviceClose()` is required when the device is open, but is not called. This results in a device handle leak.

Bug 17: Device Handle Leak on Failure
Location: Lines 140-170 in _mvncsdk2_open()
Issue: When device creation succeeds but graph creation fails, some error paths don't call ncDeviceDestroy()
Risk: Device handle leak
Fix: Ensure device cleanup on all error paths

Bug 18: FIFO Resource Imbalance
Location: Lines 240-270 in _mvncsdk2_open()
Issue: If output FIFO allocation fails, input FIFO is destroyed but input FIFO handle isn't nullified, potentially causing double-free
Risk: Double-free or use-after-free
Fix: Nullify handles after destruction

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
@@ -91,6 +91,7 @@ _mvncsdk2_close (const GstTensorFilterProperties * prop, void **private_data)
     ncGraphDestroy (&(pdata->handle_graph));
     ncFifoDestroy (&(pdata->handle_fifo_output));
     ncFifoDestroy (&(pdata->handle_fifo_input));
+    ncDeviceClose (pdata->handle_device);
     ncDeviceDestroy (&(pdata->handle_device));
 
     g_free (pdata);
@@ -186,7 +187,7 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
   if (file_model == NULL) {
     g_printerr ("Failed to g_mapped_file_new for the model file, \"%s\"\n",
         prop->model_files[0]);
-    goto err_destroy_graph_h;
+    goto err_close_device_h;
   }
 
   /* Warning: conversion unsigned long to unsigned int */
@@ -200,7 +201,7 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
   g_mapped_file_unref (file_model);
   if (ret_code != NC_OK) {
     g_printerr ("Cannot send the model file to the device\n");
-    goto err_destroy_graph_h;
+    goto err_close_device_h;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
```diff
@@ -273,15 +274,24 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
...
   *private_data = pdata;
 
   return 0;
 
 err_destroy_fifo_h:
   ncFifoDestroy (&handle_fifo_input);
+err_destroy_fifo_input_h:
   ncFifoDestroy (&handle_fifo_output);
+
+err_close_device_h:
```
myungjoo, line 285: Whenever this is reached, is_device_open is always TRUE.

songgot: Oops, my mistake. Thanks for checking. I've updated.

ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
```diff
@@ -230,14 +230,14 @@ _mvncsdk2_open (const GstTensorFilterProperties * prop, void **private_data)
...
-    goto err_destroy_graph_h;
+    goto err_close_device_h;
   }
 
   ret_code = ncFifoCreate (NNS_MVNCSDK2_NAME_OUTPUT_FIFO,
       NC_FIFO_HOST_RO, &handle_fifo_output);
   if (ret_code != NC_OK) {
     g_printerr ("Cannot create FIFO handle for output tensor\n");
-    goto err_destroy_graph_h;
+    goto err_destroy_fifo_input_h;
```
myungjoo: at ```err_destroy_fifo_input_h```, ```handle_fifo_output``` is destroyed. you need to destroy ```handle_fifo_input```.

songgot: Thank you. I've updated it.

myungjoo (commented):
Need ```ncDeviceClose()``` in _mvncsdk2_close (), too.

ext/nnstreamer/tensor_filter/tensor_filter_movidius_ncsdk2.c
```diff
@@ -91,6 +91,7 @@ _mvncsdk2_close (const GstTensorFilterProperties * prop, void **private_data)
     ncGraphDestroy (&(pdata->handle_graph));
     ncFifoDestroy (&(pdata->handle_fifo_output));
     ncFifoDestroy (&(pdata->handle_fifo_input));
+    ncDeviceClose (&(pdata->handle_device));
```
myungjoo: type error

songgot: I'm Sorry, I fixed the error.

myungjoo (approved):

Issue #4753, opened 2025-07-07
Bug report (generated)


myungjoo:
# This is a generated report. This may include hallucination.

# C Source Code Bug Analysis Report

This report analyzes C source code in `/gst` and `/ext` directories for correctness bugs and memory issues.

## Summary

After analyzing multiple C files, several categories of bugs and potential issues were identified:
- Memory management issues (leaks, missing deallocations)
- Missing null pointer checks
- Resource management problems
- Potential race conditions
- Error handling inconsistencies

---

### gst/mqtt/mqttsink.c

**Bug #1: Potential Memory Leak in Error Path**
- **Location**: Lines 569-625 in `gst_mqtt_sink_start()`
- **Issue**: When `MQTTAsync_connect()` fails, the function calls `MQTTAsync_destroy()` but may not clean up allocated `haddr` string in all error paths
- **Risk**: Memory leak
- **Fix**: Ensure `g_free(haddr)` is called before all early returns

**Bug #2: Missing Null Check**
- **Location**: Line ~750+ in `gst_mqtt_sink_render()`
- **Issue**: `gst_buffer_get_all_memory(in_buf)` return value is checked for NULL, but subsequent operations don't verify `in_buf_mem` isn't NULL before using it
- **Risk**: Null pointer dereference
- **Fix**: Add explicit null check after memory allocation

**Bug #3: Static Buffer Size Issue**
- **Location**: Lines 800-850 in `gst_mqtt_sink_render()`
- **Issue**: The static buffer flag `is_static_sized_buf` is declared static but used across multiple instances, causing potential conflicts
- **Risk**: Buffer management issues between multiple sinks
- **Fix**: Move static variable to instance-level

### gst/mqtt/mqttsrc.c

**Bug #4: Resource Cleanup Race**
- **Location**: Finalization functions (observed pattern)
- **Issue**: MQTT client handle destruction may race with callback execution
- **Risk**: Use-after-free or segmentation fault
- **Fix**: Ensure proper synchronization before destroying client handle

### gst/edge/edge_sink.c

**Bug #5: Memory Leak in Error Path**
- **Location**: Lines 450-523 in `gst_edgesink_render()`
- **Issue**: In the `done:` cleanup section, if `nns_edge_data_add()` fails after some memories are mapped, previously mapped memories may not be unmapped
- **Risk**: Memory leak
- **Fix**: Track mapped memories and ensure all are unmapped in error cases

**Bug #6: Inconsistent Error Handling**
[...]

PR #4794, opened 2025-10-28
bugfix: File descriptor leak (issue #4753)

songgot:
Issue: Multiple error paths call goto error_close but the file descriptor close operation uses g_close(src->fd, NULL) without checking if close succeeded Risk: File descriptor leak on close failure
Fix: Check return value of g_close() and log errors

### DIFF
--- a/gst/datarepo/gstdatareposrc.c
+++ b/gst/datarepo/gstdatareposrc.c
@@ -264,6 +264,22 @@ gst_data_repo_src_init (GstDataRepoSrc * src)
 }
 
 /**
+ * @brief Safely close file descriptor with error handling
+ */
+static void
+gst_data_repo_src_safe_close_fd (GstDataRepoSrc * src)
+{
+  if (src->fd) {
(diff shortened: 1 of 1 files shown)


gst/datarepo/gstdatareposrc.c
```diff
@@ -263,6 +263,22 @@ gst_data_repo_src_init (GstDataRepoSrc * src)
...
 }
 
+/**
+ * @brief Safely close file descriptor with error handling
+ */
+static void
+gst_data_repo_src_safe_close_fd (GstDataRepoSrc * src)
+{
+  if (src->fd) {
+    if (g_close (src->fd, NULL) != 0) {
```
myungjoo: BUG!

https://docs.gtk.org/glib/func.close.html

g_close returns TRUE (1) if success.

songgot: Oops, sorry. Thank you. I've fixed it.

jaeyun-jung (approved):

PR #4804, opened 2025-11-21
[Action] Fix cv2 thread safety issue in python3 filter

songgot:
- Fix thread race condition in cv2 python3 filter

Please install python3-opencv for local Ubuntu environment testing. 
```sudo apt-get install -y python3-opencv```

Root Causes:

- Missing python3-opencv dependency in CI environment, causing ImportError
- Thread race condition when multiple threads call cv2.imwrite() simultaneously
- File name conflicts using an identical filename across all threads

Solutions:

- Added python3-opencv to Build-Depends for CI installation
- Implemented thread synchronization using threading.Lock() for cv2 operations
- Used unique filenames with thread ID and timestamp to avoid conflicts

**TC Result** 
```
==================================================
    Test Group nnstreamer_filter_python3 Starts.
[PASSED] 1:Compare 1
[PASSED] 2:gst-launch of case 2
[PASSED] 2:Golden test comparison
[PASSED] 3:gst-launch of case 3
[PASSED] 3:Golden test comparison
[PASSED] 4-1:gst-launch of case 4-1
[PASSED] 4-2:Multithreaded python script as a filter (CV2)
==================================================
[PASSED] Test Group nnstreamer_filter_python3 Passed


==================================================

==================================================

[PASSED] nnstreamer_filter_python3 (8 passed among 8 cases)
==================================================
[PASSED] All Test Groups (1) Passed!
         TC Passed: 8 / Failed: 0 / Ignored: 0 
```

### CODE REVIEW

File: tests/test_models/models/passThrough_CV.py [hunk 1]
```diff
@@ -32,7 +37,20 @@ def getOutputDim(self):
...
+                with cv2_lock:
+                    # Create a simple test image and save it
+                    test_image = np.zeros((240, 320, 3), dtype=np.uint8)
+                    # Use a unique filename to avoid conflicts
+                    filename = f"/tmp/x_{threading.get_ident()}_{int(time.time() * 1000)}.png"
+                    success = cv2.imwrite(filename, test_image)
+                    if not success:
+                        print(f"Warning: Failed to write image to {filename}")
+            except Exception as e:
+                print(f"Warning: cv2 operation failed: {e}")
```
myungjoo: WHen it's failed, it should return error. (so that github-action fails and prevents merging)

songgot: Thank you. I have updated

tests/test_models/models/passThrough_CV.py
```diff
@@ -8,15 +8,20 @@
...
+import threading
+import time
 
 D1 = 3
 D2 = 280
 D3 = 40
 D4 = 1
 
-USE_CV2 = True # [NG] XXX: 'cv2' halts!!! -->  github issue (threads' race condition/deadlock? release GIL? ???)
+USE_CV2 = True # Fixed: Thread-safe cv2 usage with proper error handling
```
jaeyun-jung: If it is always true, I suggest removing this flag, just import cv2 and declare cv2_lock = threading.Lock()

songgot: Thank you. I have updated.

tests/test_models/models/passThrough_CV.py
```diff
@@ -32,7 +34,21 @@ def getOutputDim(self):
...
+        try:
+            # Thread-safe cv2 operation with proper error handling
+            with cv2_lock:
+                # Create a simple test image and save it
+                test_image = np.zeros((240, 320, 3), dtype=np.uint8)
+                # Use a unique filename to avoid conflicts
+                filename = f"/tmp/x_{threading.get_ident()}_{int(time.time() * 1000)}.png"
+                success = cv2.imwrite(filename, test_image)
+                if not success:
+                    print(f"Warning: Failed to write image to {filename}")
```
myungjoo, line 47: this also needs to return error.

songgot: Thank you.

jaeyun-jung (approved):

PR #4806, opened 2025-11-27
[single] Add latency profiling in single filter

anyj0527:
- Introduce latency profiling in single like tensor_filter

**Self evaluation:**

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_single.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_single.c
@@ -389,6 +389,7 @@ g_tensor_filter_single_invoke (GTensorFilterSingle * self,
   GstTensorMemory out_tensors[NNS_TENSOR_SIZE_LIMIT] = { {0} }; /** @todo refactor this local variable */
   guint i;
   gint status;
+  gint64 start_time, end_time;  /* variable for latency profiling */
 
   spriv = G_TENSOR_FILTER_SINGLE_PRIV (self);
   priv = &spriv->filter_priv;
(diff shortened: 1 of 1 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_single.c
```diff
@@ -427,7 +428,15 @@ g_tensor_filter_single_invoke (GTensorFilterSingle * self,
     }
   }
 
+  start_time = g_get_monotonic_time ();
   GST_TF_FW_INVOKE_COMPAT (priv, status, input, _out);
+  end_time = g_get_monotonic_time ();
+
+  /* check invoke latency when latency profiling is enabled */
+  if (priv->latency_mode > 0) {
+    ml_logi ("[%s] Invoke took %.3f ms", TF_MODELNAME (&(priv->prop)),
```
jaeyun-jung, line 437: 1. hide model file path?
2. print inference time if status is 0?
```
invoke_time = g_get_monotonic_time ();
GST_TF_FW_INVOKE_COMPAT (priv, status, input, _out);

if (status == 0) {
  /* check invoke latency when latency profiling is enabled */
  if (priv->latency_mode > 0) {
    g_autofree gchar *model_file = priv->prop.model_files ? g_path_get_basename (priv->prop.model_files[0]) : g_strdup ("No model file");

    invoke_time = g_get_monotonic_time () - invoke_time;
    ml_logi ("[%s][%s] Invoke took %.3f ms", priv->prop.fwname, model_file, invoke_time / 1000.0f);
  }
   ...
```

songgot: Shouldn't we measure the time after the if statement below? line 448?
```
if (_out != output) {...}
```

jaeyun-jung: It is not related with inference time :)

songgot: I had thought that the inference process was complete once the results were copied to the output buffer.
Thank you.

myungjoo: This is for the client's request on the model-framework-invoke latency, which is to ensure that the "long" latency is due to the model's computation complexity itself.

(Such information is highly recommended to be added in the commit message although you discussed it in the meeting.)

PR #4815, opened 2025-12-05
[Filter] static variable for property

jaeyun-jung:
Code clean, remove static variable to set property information.


songgot (approved):
LGTM

PR #4818, opened 2025-12-05
[Filter/Single] log after invoke is done

jaeyun-jung:
Code clean,
1. print latency log when inference is done.
2. hide model path, it would be a private data of an application.


songgot (approved):
LGTM :+1:

PR #4829, opened 2025-12-19
Fix: prevent leaks and exception escapes in tensor_filter_subplugin C wrappers

myungjoo:
## What

Originally: `cpp_open()` leaked the object returned by `getEmptyInstance()` whenever `configure_instance()` threw (identified by cursor). The scope grew through review rounds into making the C++ subplugin C-ABI wrappers exception-safe as a whole:

- `cpp_open()`: the spawned instance is owned by a `std::unique_ptr` and `release()`d into `*private_data` only on success; both `getEmptyInstance()` and `configure_instance()` are guarded, including `catch (...)`.
- `cpp_invoke()`, `cpp_getFrameworkInfo()`, `cpp_getModelInfo()`, `cpp_eventHandler()`, and `GET_TFSP_WITH_CHECKS`: `catch (...)` coverage so no exception can propagate into the C caller (undefined behavior / `std::terminate`).

## API contract (affects out-of-tree subplugin authors)

The object returned by `getEmptyInstance()` must be heap-allocated and singly owned by the framework: it is deleted on close, or immediately when `configure_instance()` throws. Previously a failed configure leaked the object and never ran its destructor; now the destructor always runs, so it must be safe on a partially-configured instance. The rule is documented in `nnstreamer_cppplugin_api_filter.hh` and `Documentation/writing-subplugin-tensor-filter.md`. All 24 in-tree subplugins conform (`return *(new X ());`).

Related: dali cleanup fix #4859; tensorrt10 cleanup rework #4860 (its `cleanup()` must stay idempotent since the framework now also destroys the instance on a failed configure).

## Tests

`testTensorFilterCppSubplugin` (gtest fixture, runs in the gating meson/Valgrind and Tizen GBS unit-test jobs):
- `openFailNoLeak_n`: repeated failing opens with `configure_instance()` throwing at three distinct points (before/after acquiring a resource, after acquire-and-release); asserts, via instance and resource counters, that neither the instance nor its resources leak.
- `openCloseOwnership_p`: successful open transfers ownership to `*private_data`; close destroys the instance and releases its resources.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -34,6 +34,7 @@
 #include <nnstreamer_log.h>
 #include <string.h>
 
+#include <memory>
 #include <system_error>
 
 #define NO_ANONYMOUS_NESTED_STRUCT
@@ -85,15 +86,26 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
 
   /* 2. Spawn another empty object and configure the empty object */
   tensor_filter_subplugin &obj = sp->getEmptyInstance ();
+  /**
+   * NOTE:
+   * getEmptyInstance() is expected to return a heap-allocated object.
+   * If configure_instance() throws, cpp_open() must delete the object to avoid leaks
+   * when open() is repeatedly attempted with invalid properties.
+   */
+  tensor_filter_subplugin *obj_ptr = std::addressof (obj);
   try {
     obj.configure_instance (prop);
   } catch (const std::invalid_argument &e) {
+    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::system_error &e) {
+    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (e.code ().value () * -1, e.what ());
   } catch (const std::runtime_error &e) {
+    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::exception &e) {
+    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   }
 
@@ -103,9 +115,9 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
 
 /* 4. Save the object as *private_data */
 #if __GNUC__ < 5 || __cplusplus < 201103L
-  *private_data = &(obj);
+  *private_data = obj_ptr;
 #else /* It is safer w/ addressof, but old gcc doesn't appear to support it */
-  *private_data = std::addressof (obj);
+  *private_data = obj_ptr;
 #endif
 
   return 0;
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -88,24 +88,21 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   tensor_filter_subplugin &obj = sp->getEmptyInstance ();
   /**
    * NOTE:
-   * getEmptyInstance() is expected to return a heap-allocated object.
-   * If configure_instance() throws, cpp_open() must delete the object to avoid leaks
-   * when open() is repeatedly attempted with invalid properties.
+   * getEmptyInstance() must return a heap-allocated, singly-owned object.
+   * If configure_instance() throws, obj_ptr deletes the object to avoid
+   * leaks when open() is repeatedly attempted with invalid properties.
+   * On success, ownership is transferred to *private_data via release().
    */
-  tensor_filter_subplugin *obj_ptr = std::addressof (obj);
+  std::unique_ptr<tensor_filter_subplugin> obj_ptr (std::addressof (obj));
   try {
     obj.configure_instance (prop);
   } catch (const std::invalid_argument &e) {
-    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::system_error &e) {
-    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (e.code ().value () * -1, e.what ());
   } catch (const std::runtime_error &e) {
-    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::exception &e) {
-    delete obj_ptr;
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   }
 
@@ -113,12 +110,8 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
    * nnstreamer_filter_find) empty object */
   obj.fwdesc.v1.subplugin_data = nullptr;
 
-/* 4. Save the object as *private_data */
-#if __GNUC__ < 5 || __cplusplus < 201103L
-  *private_data = obj_ptr;
-#else /* It is safer w/ addressof, but old gcc doesn't appear to support it */
-  *private_data = obj_ptr;
-#endif
+  /* 4. Save the object as *private_data */
+  *private_data = obj_ptr.release ();
 
   return 0;
 }
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -85,8 +85,16 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   assert (sp->sanity == _SANITY_CHECK); /** tfsp is using me! */
 
   /* 2. Spawn another empty object and configure it; obj_ptr deletes it if configure_instance() throws */
-  tensor_filter_subplugin &obj = sp->getEmptyInstance ();
-  std::unique_ptr<tensor_filter_subplugin> obj_ptr (std::addressof (obj));
+  tensor_filter_subplugin *obj_raw;
+  try {
+    obj_raw = std::addressof (sp->getEmptyInstance ());
+  } catch (const std::exception &e) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getEmptyInstance()");
+  }
+  std::unique_ptr<tensor_filter_subplugin> obj_ptr (obj_raw);
+  tensor_filter_subplugin &obj = *obj_ptr;
   try {
     obj.configure_instance (prop);
   } catch (const std::invalid_argument &e) {
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -233,13 +233,16 @@ tensor_filter_subplugin::cpp_getModelInfo (const GstTensorFilterFramework *tf,
   UNUSED (tf);
   UNUSED (prop);
 
+  int ret;
   try {
-    return obj->getModelInfo (ops, *in_info, *out_info);
+    ret = obj->getModelInfo (ops, *in_info, *out_info);
   } catch (const std::exception &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (...) {
     _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getModelInfo()");
   }
+
+  return ret;
 }
 
 /**
@@ -256,13 +259,16 @@ tensor_filter_subplugin::cpp_eventHandler (const GstTensorFilterFramework *tf,
   UNUSED (tf);
   UNUSED (prop);
 
+  int ret;
   try {
-    return obj->eventHandler (ops, *data);
+    ret = obj->eventHandler (ops, *data);
   } catch (const std::exception &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (...) {
     _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from eventHandler()");
   }
+
+  return ret;
 }
 
 /**
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -88,6 +88,8 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   tensor_filter_subplugin *obj_raw;
   try {
     obj_raw = std::addressof (sp->getEmptyInstance ());
+  } catch (const std::bad_alloc &e) {
+    _RETURN_ERR_WITH_MSG (-ENOMEM, e.what ());
   } catch (const std::exception &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (...) {
@@ -97,6 +99,8 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   tensor_filter_subplugin &obj = *obj_ptr;
   try {
     obj.configure_instance (prop);
+  } catch (const std::bad_alloc &e) {
+    _RETURN_ERR_WITH_MSG (-ENOMEM, e.what ());
   } catch (const std::invalid_argument &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::system_error &e) {
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -62,7 +62,7 @@ namespace nnstreamer
   do {                                                                  \
     try {                                                               \
       obj = get_tfsp_with_checks (private_data);                        \
-    } catch (const std::exception &e) {                                 \
+    } catch (...) {                                                     \
       /** @todo Write exception handlers. */                            \
       return -EINVAL;                                                   \
       /** @todo return different error codes according to exceptions */ \
@@ -84,9 +84,7 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   tensor_filter_subplugin *sp = (tensor_filter_subplugin *) tfsp->v1.subplugin_data;
   assert (sp->sanity == _SANITY_CHECK); /** tfsp is using me! */
 
-  /* 2. Spawn another empty object and configure the empty object.
-   * If configure_instance() throws, obj_ptr deletes the object; see the
-   * ownership contract of getEmptyInstance() in nnstreamer_cppplugin_api_filter.hh */
+  /* 2. Spawn another empty object and configure it; obj_ptr deletes it if configure_instance() throws */
   tensor_filter_subplugin &obj = sp->getEmptyInstance ();
   std::unique_ptr<tensor_filter_subplugin> obj_ptr (std::addressof (obj));
   try {
@@ -178,6 +176,8 @@ tensor_filter_subplugin::cpp_invoke (const GstTensorFilterFramework *tf,
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::exception &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from invoke()");
   }
 
   return 0;
@@ -213,6 +213,8 @@ tensor_filter_subplugin::cpp_getFrameworkInfo (const GstTensorFilterFramework *t
   } catch (const std::exception &e) {
     /** @todo Write exception handlers. */
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getFrameworkInfo()");
   }
   return 0;
 }
@@ -231,7 +233,13 @@ tensor_filter_subplugin::cpp_getModelInfo (const GstTensorFilterFramework *tf,
   UNUSED (tf);
   UNUSED (prop);
 
-  return obj->getModelInfo (ops, *in_info, *out_info);
+  try {
+    return obj->getModelInfo (ops, *in_info, *out_info);
+  } catch (const std::exception &e) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getModelInfo()");
+  }
 }
 
 /**
@@ -248,7 +256,13 @@ tensor_filter_subplugin::cpp_eventHandler (const GstTensorFilterFramework *tf,
   UNUSED (tf);
   UNUSED (prop);
 
-  return obj->eventHandler (ops, *data);
+  try {
+    return obj->eventHandler (ops, *data);
+  } catch (const std::exception &e) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from eventHandler()");
+  }
 }
 
 /**
--- a/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
+++ b/gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
@@ -84,15 +84,10 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
   tensor_filter_subplugin *sp = (tensor_filter_subplugin *) tfsp->v1.subplugin_data;
   assert (sp->sanity == _SANITY_CHECK); /** tfsp is using me! */
 
-  /* 2. Spawn another empty object and configure the empty object */
+  /* 2. Spawn another empty object and configure the empty object.
+   * If configure_instance() throws, obj_ptr deletes the object; see the
+   * ownership contract of getEmptyInstance() in nnstreamer_cppplugin_api_filter.hh */
   tensor_filter_subplugin &obj = sp->getEmptyInstance ();
-  /**
-   * NOTE:
-   * getEmptyInstance() must return a heap-allocated, singly-owned object.
-   * If configure_instance() throws, obj_ptr deletes the object to avoid
-   * leaks when open() is repeatedly attempted with invalid properties.
-   * On success, ownership is transferred to *private_data via release().
-   */
   std::unique_ptr<tensor_filter_subplugin> obj_ptr (std::addressof (obj));
   try {
     obj.configure_instance (prop);
@@ -104,6 +99,8 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
   } catch (const std::exception &e) {
     _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
+  } catch (...) {
+    _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from configure_instance()");
   }
 
   /** 3. Mark that this is not a representative (found by
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,10 @@
 2.6.0 -> 2.7.0
 	- 2.7.0 is a devel version for 2.8.0 release. Experimental features may be added to 2.7.x.
 
+	- C++ subplugin API (tensor_filter_subplugin) behavior change
+		- The framework now deletes the object returned by getEmptyInstance() when configure_instance() throws, instead of leaking it. The object must be heap-allocated and singly owned by the framework, and the derived destructor must be safe on a partially-configured instance.
+		- All in-tree subplugins conform. An out-of-tree subplugin whose getEmptyInstance() returns a static, pooled, or shared instance must be fixed to return a newly allocated object; see Documentation/writing-subplugin-tensor-filter.md.
+
 2.4.4 -> 2.6.0
         - llama.cpp upgrades
                 - Added context save/restore, KV-cache trimming, LoRA adapters, token-by-token samplers, and safer teardown paths so on-device LLMs support long chats and tuning.
--- a/Documentation/writing-subplugin-tensor-filter.md
+++ b/Documentation/writing-subplugin-tensor-filter.md
@@ -35,6 +35,15 @@ As in C subplugin, the derived (concrete) class should be registered at init and
 Subplugin writers are supposed to use the static methods of the base class, ```register_subplugin()``` and ```unregister_subplugin()```; refer to the function ```init_filter_snap()``` and ```fini_filter_snap()``` in the reference example.
 
 
+### Object ownership and exception rules
+
+```getEmptyInstance()``` must return a heap-allocated object that is singly owned by the framework: the framework deletes it (through the virtual destructor) when the filter is closed, or immediately if ```configure_instance()``` throws.
+Do not return a reference to a static, pooled, or otherwise shared instance, and do not retain a reference to the returned object inside the subplugin.
+
+If ```configure_instance()``` throws, clean up any external references to the object (e.g., entries in a global registry) before throwing, because the framework deletes the object right away.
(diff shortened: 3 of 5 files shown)


gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc
```diff
@@ -85,15 +86,26 @@ tensor_filter_subplugin::cpp_open (const GstTensorFilterProperties *prop, void *
...
+   * NOTE:
+   * getEmptyInstance() is expected to return a heap-allocated object.
+   * If configure_instance() throws, cpp_open() must delete the object to avoid leaks
+   * when open() is repeatedly attempted with invalid properties.
+   */
+  tensor_filter_subplugin *obj_ptr = std::addressof (obj);
   try {
     obj.configure_instance (prop);
   } catch (const std::invalid_argument &e) {
+    delete obj_ptr;
```
songgot: How about using code below?
 ```std::unique_ptr<tensor_filter_subplugin> obj_ptr(&sp->getEmptyInstance()); ```
 and remvoe ```delete obj_ptr```.

myungjoo:

## Review — Fix: Prevent memory leaks in tensor_filter_subplugin::cpp_open

The leak is real: before this change, when `configure_instance()` threw, the object returned by `getEmptyInstance()` was simply abandoned. For a pipeline that repeatedly reopens a filter with bad properties, that leaks one full subplugin instance per attempt. The fix is in the right place and the scope is minimal.

I verified the two things that would make `delete obj_ptr` unsafe, and both are fine:

- `tensor_filter_subplugin` declares `virtual ~tensor_filter_subplugin ()` (`gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh:176`), so deleting through the base pointer is well-defined.
- Every in-tree `getEmptyInstance()` returns `*(new Derived ())` — all 24 of them (`ext/nnstreamer/tensor_filter/*.cc`). None returns a static or a singleton, so none would be `delete`d spuriously.

Findings below.

### 1. (Major) This changes an undocumented API contract for out-of-tree subplugins

`nnstreamer_cppplugin_api_filter.hh` documents `getEmptyInstance()` only as "Return a newly created 'empty' object of the derived class." It never states who owns the object if `configure_instance()` throws — and until now the framework's *actual* behavior was "nobody, it leaks". After this PR the framework unconditionally `delete`s it.

This is a public C++ header shipped in the devel package, so out-of-tree subplugins exist. Two concrete ways a conforming-until-now subplugin breaks:

- one that returns a reference to a pooled/cached instance (nothing in the header forbids it) now gets a live object deleted under it;
- one whose `configure_instance()` registers `this` in a global/static registry *before* the throwing step now holds a dangling pointer.

Please make the contract explicit in the header, on `getEmptyInstance()` and on `configure_instance()`: "the object returned by `getEmptyInstance()` must be heap-allocated and singly-owned; if `configure_instance()` throws, the framework deletes it, and the subplugin must not retain any reference to it."

### 2. (Major) Destructors must now survive a partially-configured object

This is the risk that needs a per-subplugin check before merge. Previously a failed `configure_instance()` never ran the derived destructor. Now it always does — on an object whose members may be in whatever state the throwing step left them.

Concretely, `tensorrt10_subplugin::~tensorrt10_subplugin()` calls `cleanup()`, which begins with `if (!_configured) return;` and `_configured` is only set at the *end* of `configure_instance()`. So this PR turns "leak the object" into "leak the object's CUDA allocations instead" for that subplugin — better, but not free, and it is worth confirming the same shape holds for the others (`onnxruntime`, `ncnn`, `qnn`, `python3`, `snpe*`, `executorch*`, `llamacpp`, `tvm`, …). Any destructor that unconditionally frees a member it assumes `configure_instance()` initialized will now double-free or crash on the error path, where before it simply never ran.

A sweep of the ~24 in-tree destructors, plus a note in the header telling subplugin authors that their destructor must be safe on a never-configured instance, would make this change safe to land.

### 3. (Minor) The `#if __GNUC__ < 5` block is now dead code

```cpp
#if __GNUC__ < 5 || __cplusplus < 201103L
  *private_data = obj_ptr;
#else /* It is safer w/ addressof, but old gcc doesn't appear to support it */
  *private_data = obj_ptr;
#endif
```

Both branches are identical, and the comment now describes something the code no longer does (the `std::addressof` call moved up to the `obj_ptr` initialization). Please collapse this to a single `*private_data = obj_ptr;` and drop the stale comment — leaving a no-op conditional is exactly the kind of thing that confuses the next reader.

### 4. (Major) No test

This is a leak fix with no test, so nothing prevents it from silently regressing. `tests/nnstreamer_filter_extensions_common/` already builds a per-subplugin template that opens filters with invalid properties; a case that opens-and-fails in a loop and asserts bounded RSS (or simply runs under the existing valgrind CI job, `GTEST run with Valgrind on a case` in `ubuntu_clean_meson_build.yml`) would both prove the fix and catch a future reintroduction. Without one, the only signal is a valgrind summary that CI is currently configured not to fail on.

Given item 2, a second test is worth having: open a subplugin with properties that make `configure_instance()` throw, and assert the process does not crash — that is the new code path this PR creates, and it currently has no coverage at all.

### Other checks

- **Scope:** correct, one file, no unrelated modules touched.
- **Interaction:** #4854 touches `tensorrt10_subplugin::cleanup()`; see item 2 for how the two interact.
- **Security:** no new attack surface. No back-door-shaped code found.

myungjoo:

## Re-review of 9c314c9

This is a substantially better PR than the previous revision. All four earlier items are addressed, and the `unique_ptr` rewrite is a genuine improvement over the manual `delete` in each catch block — fewer places to forget, and correct by construction.

### Verified

**The tests actually run and actually gate.** Both cases executed in CI (run 32799319402, `[ OK ] testTensorFilterCppSubplugin.openFailNoLeak` / `openCloseOwnership` on both `ubuntu-22.04` and `ubuntu-22.04-arm`), and I traced what each would do against the unpatched code:

- `openFailNoLeak` fails on the **first** iteration without the fix: `getEmptyInstance()` raises `instances_alive` to 2, `configure_instance()` throws, the old catch returns without deleting, so `EXPECT_EQ (instances_alive, 1U)` trips immediately. This is a real regression gate, not a test that merely passes.
- `openCloseOwnership` gates the main hazard the rewrite introduces: writing `obj_ptr.get()` instead of `.release()` would delete the object on return, leaving `instances_alive` at 1 instead of 2 and setting up a double-free in `cpp_close()`. The test catches both halves.

Worth calling out because it is the harder thing to get right — a leak-fix test that does not distinguish patched from unpatched is common, and this one does.

**Other confirmations:** the dead `#if __GNUC__ < 5` block is gone; `<memory>` is included; the base class has a `virtual` destructor (`nnstreamer_cppplugin_api_filter.hh:176`) so `unique_ptr`'s default deleter is well-defined; `_RETURN_ERR_WITH_MSG` expands to a plain `return`, so the guard destructor runs on every catch path; and `cpp_close()`'s `delete obj` + `*private_data = nullptr` matches what `openCloseOwnership` asserts.

A side benefit not mentioned in the commit message: an exception **not** derived from `std::exception` previously leaked the instance outright, since no catch clause matched. With the guard in place, stack unwinding now frees it.

### 1. (Major) The documented contract is stronger than what is tested, and two in-tree subplugins do not meet it

The header now promises:

> the destructor of the derived class must be safe to run on a partially-configured (or never-configured) instance.

The never-configured half was already enforced before this PR — `register_subplugin<T>()` does `new T()` and `unregister_subplugin()` does `delete emptyInstance` at library unload, so every subplugin's destructor already runs on a pristine object. The **partially-configured** half is the genuinely new state this PR creates, and `cpp_mock_subplugin::configure_instance()` throws immediately without acquiring anything, so the test never enters it.

I swept all in-tree C++ subplugins for what actually happens on that path:

| Behavior on `configure_instance()` throw | Subplugins |
| --- | --- |
| Calls its own `cleanup()` before every `throw` (idempotent — each ends by flipping `configured`/`empty_model`) | `qnn`, `executorch`, `tvm`, `llamacpp`, `onnxruntime` |
| Sets `configured = true` *before* acquiring resources, so the destructor cleans fully | `snpe` |
| **Neither** — destructor early-returns and nothing is freed | **`tensorrt10`, `dali`** |

For `tensorrt10`, `cleanup()` opens with `if (!_configured) return;` while `_configured = true` is set only at line 287, after `loadModel()`. A throw inside `loadModel()` therefore leaks the CUDA device buffers and the `cudaStream`. For `dali`, `configured = true` is set at line 297 at the end of `loadPipeline()`, so a throw leaks `_pipeline_handle` and the `_pipeline_path` allocated at line 265.

So after this PR the framework no longer leaks the *object*, but for those two the *resources it acquired* still leak. That is not a defect in this PR's code, but it does mean the PR's title claim is only partly realised, and the header now documents a contract that two in-tree implementations satisfy only in the weak sense of "does not crash".

Two suggestions:

- **Make the test check what the header promises.** Have the mock allocate something in `configure_instance()` before throwing (a `g_malloc`, or just a second counter incremented on acquire and decremented in the destructor) and assert it is released. That converts the documented contract into a checked one, and it is a few lines on top of what is already here.
- **Track the two offenders.** For `tensorrt10` the natural place is #4854, which is already editing that exact `cleanup()` function — and which *widens* the window by moving output-buffer allocation into `loadModel()`. For `dali`, note separately that `cleanup()` never frees `_pipeline_path` even on the success path; that is a pre-existing bug independent of this PR. Simplest fix for both: drop the `if (!configured) return;` guard entirely, since every free inside is already null-safe.

### 2. (Minor) An exception can still escape `cpp_open()` into C

`cpp_open()` is reached through the C function-pointer table in `GstTensorFilterFramework`. If `configure_instance()` throws something outside the `std::exception` hierarchy, it now unwinds cleanly past the guard (good) but then propagates out of a function called from C — undefined behavior, in practice `std::terminate`.

Adding a final clause closes it for the cost of three lines:

```cpp
} catch (...) {
  _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from configure_instance()");
}
```

Given that the header is being amended in this PR to describe what `configure_instance()` may do, this is a natural place to make the framework side total.

### 3. (Minor) Test state is process-global, and an `ASSERT` before cleanup leaks it into the next case

`instances_alive`, `registered` and `throw_on_configure` are static, and both tests call `init()` … `fini()` inline. If `ASSERT_TRUE (fw && fw->open && fw->close)` fires, the test returns immediately, `fini()` never runs, the mock stays registered and the counter stays non-zero — so the *other* test then fails on its first `EXPECT_EQ (instances_alive, 1U)` for an unrelated reason, and the real failure is buried. `throw_on_configure` is likewise never reset in `fini()`.

A small `::testing::Test` fixture with `SetUp()`/`TearDown()` doing `init()`/`fini()` and resetting `throw_on_configure` keeps each failure local. Cheap, and these two tests are exactly the ones a future maintainer will be reading when something goes wrong.

### 4. (Trivial)

- The 100-iteration loop has three `EXPECT_*` per pass, so an actual leak emits roughly 300 failure lines. `ASSERT_EQ` for the counter inside the loop stops at the first and is far easier to read.
- `instances_alive` is a non-atomic `guint`. Correct for single-threaded gtest; a one-line comment saying so pre-empts the question.
- `tests/cpp_methods/` already exists for C++-filter-related tests, though it targets the `cpp` *framework* rather than the subplugin base API, and `unittest_plugins.cc` already hosts the `testTensorFilter*` family. Current placement is defensible — just noting that `unittest_plugins.cc` is now 8550 lines.

### Other checks

- **Scope:** three files, and each one earns its place (implementation, the contract it establishes, the test that gates it). No unrelated modules touched.
- **CI:** all checks green on 9c314c9. Note `Build on MacOS-latest with Meson` passes in 12s, i.e. it is still skipping every step — that is the #4858 issue, not this PR.
- **Regression risk:** the only behavioral change for existing subplugins is that a failed `configure_instance()` now runs their destructor. Per the sweep above, all 24 in-tree implementations are safe under that; out-of-tree ones are now covered by the documented contract.
- **Security:** no new attack surface. No back-door-shaped code found.

myungjoo:

## Review of the full PR (3 commits, HEAD 9e4e6ca8)

The leak fix itself is sound: `getEmptyInstance()`'s return value is now owned by a `std::unique_ptr` from the moment it is created, `configure_instance()`'s throwing paths (including the new `catch (...)`) all unwind through that guard and delete the object, and only a successful configure releases ownership into `*private_data`. I traced this against `cpp_close()` and the virtual destructor in `nnstreamer_cppplugin_api_filter.hh:190` — deleting through the base pointer is well-defined, and all 24 in-tree `getEmptyInstance()` implementations under `ext/nnstreamer/tensor_filter/*.cc` return `*(new Derived())`, so none is broken by the new "framework owns and deletes it" contract. Both prior review rounds are, with one exception below, addressed:

- Round 1's four items (undocumented ownership contract, destructor-safety risk, dead `#if __GNUC__ < 5` block, no test) are all resolved in commit 9c314c90.
- Round 2's items are resolved in 9e4e6ca8: `catch (...)` was added to `cpp_open()` (`tensor_filter_support_cc.cc:102-104`), the mock/tests moved into a `TEST_F` fixture with `SetUp`/`TearDown` so a failed `ASSERT` no longer bleeds registration state into the next case, the mock now acquires a "resource" before throwing (`resources_alive`) so the destructor-safety contract is actually exercised rather than only documented, and the loop uses `ASSERT_EQ` to fail fast. I confirmed `TearDown()` still runs GoogleTest's contract: even when a `TEST_F` body `ASSERT`s early, `TestBody()` returns normally (no C++ exception) and the framework always calls `TearDown()` afterward, so the leak checks in `TearDown` (`instances_alive == 0`, `resources_alive == 0`) fire on every path.

However, **CI is currently red on the exact commit under review**, for a regression introduced by 9e4e6ca8 itself, and the whole PR otherwise looks ready. See Finding 1.

**Verdict: not OK to merge as-is.** The one blocking issue (Finding 1) is a one-line fix. Once that is corrected — re-push and confirm "Static checks" is green — this is good to merge; nothing else found rises above Minor.

### 1. (Critical) The head commit fails the required "Static checks" CI job, and this specific job is a required status check on `main`

`gh pr checks 4829` shows `Static checks` = `fail` on `9e4e6ca8` (https://github.com/nnstreamer/nnstreamer/actions/runs/32810716124/job/97689386771), and `repos/nnstreamer/nnstreamer/branches/main/protection` lists `Static checks` in `required_status_checks.contexts`. The PR cannot be merged through branch protection in its current state.

The failure: `.github/workflows/static.check.scripts/doxygen-tag.sh` rejects any multi-line `/* ... */` comment that doesn't start with `/**`. In `gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc:87-89`:

```cpp
  /* 2. Spawn another empty object and configure the empty object.
   * If configure_instance() throws, obj_ptr deletes the object; see the
   * ownership contract of getEmptyInstance() in nnstreamer_cppplugin_api_filter.hh */
```

This is new in 9e4e6ca8 — at 9c314c90 the same spot was a `/** NOTE: ... */` block (doxygen-style, so it passed), and round 2 of the prior AI review explicitly noted "CI: all checks green on 9c314c9." The rewrite in the latest commit reverted the comment to a plain `/*` opener while keeping it multi-line, which the checker flags every time (log: `[ERROR] File name: gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc, 87 line, Doxygen or multi line comments should begin with /**`).

Fix: change the opening delimiter to `/**`, e.g.:
```cpp
  /**
   * 2. Spawn another empty object and configure the empty object.
   * If configure_instance() throws, obj_ptr deletes the object; see the
   * ownership contract of getEmptyInstance() in nnstreamer_cppplugin_api_filter.hh
   */
```

### 2. (Minor) `catch (...)` is only on `cpp_open()`; the same C-boundary UB is still reachable from the other C wrapper callbacks

Round 2 correctly identified that an exception outside the `std::exception` hierarchy unwinding out of `cpp_open()` (called from C/GStreamer) is undefined behavior, and the fix added `catch (...)` there. The same reasoning applies to `cpp_invoke()` (`tensor_filter_support_cc.cc:167-181`), `cpp_getFrameworkInfo()` (`:211-216`), and `cpp_eventHandler()`/`cpp_getModelInfo()`, none of which have a `catch (...)` guard — a non-`std::exception` thrown from `obj->invoke()` (or `invoke_dynamic`, `getFrameworkInfo`) still propagates into C and can `std::terminate` the pipeline. This is pre-existing behavior, not a regression from this PR, and strictly out of the stated "cpp_open" scope, but since the PR is already establishing the "framework must be total at the C boundary" precedent for `cpp_open`, it's worth a follow-up applying the same guard to the other wrappers.

### 3. (Trivial) Prior review's "track the two offenders" suggestion (tensorrt10, dali) wasn't recorded anywhere in this PR

Round 2 found that `tensorrt10_subplugin` and `dali_subplugin`'s destructors don't release resources acquired before a `configure_instance()` throw (both guard cleanup on a `configured`/`_configured` flag set only at the end of the method), and suggested tracking those two separately (e.g., via #4854 for tensorrt10, and a new note for dali's pre-existing `_pipeline_path` leak) since fixing them is out of this PR's scope. Neither the PR body nor the commits reference a tracking issue for either. Not blocking — the current PR's own contract test (mock resource counter) is scoped correctly to the framework's object-ownership guarantee, not to auditing every subplugin's resource cleanup — but leaving no breadcrumb means this is likely to be forgotten.

### Other checks

- **Scope:** three files (implementation, header contract, tests), each earns its place; no unrelated modules touched.
- **Tests actually gate merges:** `tests/nnstreamer_plugins/unittest_plugins.cc` is registered via `test('unittest_plugins', unittest_plugins, env: testenv)` in `tests/meson.build:114`, built into the `unittest_plugins` executable, and `meson test -C build/ -v` runs in `.github/workflows/ubuntu_clean_meson_build.yml` (the "Build with meson and test with Valgrind in Ubuntu" required check) — a regression in `testTensorFilterCppSubplugin.{openFailNoLeak,openCloseOwnership}` would fail that required job and block merge. Confirmed both cases build and would fail without the fix (traced in round 2; not independently re-run here).
- **Regression risk to the ~24 in-tree C++ subplugins:** none found. The only behavioral change for existing subplugins is that a failed `configure_instance()` now runs their destructor (previously it never ran on that path); this was already the documented expectation once an object is `delete`d via `cpp_close()`, and the round-2 sweep found all in-tree implementations tolerate it (two only in the "does not crash" sense, per Finding 3).
- **Documentation:** the public header `nnstreamer_cppplugin_api_filter.hh` is the only place this contract is documented in-repo (no separate `docs/` reference exists), and its `getEmptyInstance()`/`configure_instance()` comments were updated in this PR to state the new ownership rule — adequate.

myungjoo:

## Follow-up review of 42d680b5

All three findings from my previous review are resolved.

### Finding 1 (was Critical) — Resolved

The line-87 comment in `cpp_open()` is now a single line ending in `*/`:
```cpp
/* 2. Spawn another empty object and configure it; obj_ptr deletes it if configure_instance() throws */
```
This satisfies the doxygen-tag checker's rule (multi-line `/* ... */` blocks must start with `/**`). Confirmed on the actual CI run for this commit: **`Static checks` = pass** (https://github.com/nnstreamer/nnstreamer/actions/runs/32816573380/job/97705962864), which is the required status check that was previously blocking the merge button.

### Finding 2 (was Minor) — Resolved, with one small new observation

`catch (...)` was added to `cpp_invoke`, `cpp_getFrameworkInfo`, `cpp_getModelInfo`, `cpp_eventHandler`, and the shared `GET_TFSP_WITH_CHECKS` macro (widened from `catch (const std::exception &)` to `catch (...)`). All four callbacks now catch every exception type before it can unwind into the C caller.

One thing worth a closer look, since it's a new pattern in this commit: `cpp_getModelInfo` and `cpp_eventHandler` were rewritten from a bare `return obj->method(...);` to

```cpp
try {
  return obj->getModelInfo (ops, *in_info, *out_info);
  _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
} catch (...) {
  _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getModelInfo()");
}
```
with **no statement after the try/catch block** (`tensor_filter_support_cc.cc:239-245` and `:262-268`). Every path inside does return — the try body returns, and both catch handlers return unconditionally via `_RETURN_ERR_WITH_MSG`'s `do { ...; return c; } while (0);` — but the function relies on the compiler's flow analysis to see that through the macro. This project builds with `werror=true` + `warning_level=2` + an explicit `-Werror=return-type` (`meson.build:58`), so a compiler that can't prove it would turn this into a hard build failure rather than a warning. I checked this is not just theoretical risk: it does build clean under at least one toolchain in the matrix — `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)` passed (2m36s, same commit, same `-Werror=return-type`) — and `Static checks`/`MacOS`/`Spell Check` are also green. The rest of the build matrix (RISC-V cross-gcc, Android NDK clang, Tizen GBS cross toolchains, Ubuntu x86_64 Valgrind, LLVM/Clang, Yocto, pdebuild) was still pending at review time; see CI status below.

Given at least one GCC toolchain already compiles it cleanly, I'm not calling this blocking, but for robustness/consistency with the sibling functions `cpp_invoke`/`cpp_getFrameworkInfo` (which keep an explicit trailing `return 0;` after their try/catch instead of returning from inside the try), consider adding an explicit fallback after the try/catch in `cpp_getModelInfo`/`cpp_eventHandler` too — e.g. move the call out of the `try` and keep a plain `return obj->getModelInfo(...)` outside it, mirroring the other two wrappers. (Minor/style, not required for this PR to be correct.)

### Finding 3 (was Trivial) — Resolved

The commit message for 42d680b5 records the tensorrt10/dali follow-up as promised. Verified both references are real and on-topic:
- #4859, "[Filter/dali] Fix resource leaks in cleanup()", open.
- #4854 carries a comment titled "Tracking note from the #4829 review" describing the exact `tensorrt10_subplugin::cleanup()` / `_configured` issue from my prior review.

### CI status at time of this review

- Required check `Static checks`: **pass**.
- `DCO`: pass. `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)`: pass. `Build on MacOS-latest with Meson`: pass (note: this job completes in ~12s because it currently skips build steps — a pre-existing, unrelated issue tracked as #4858, not evidence either way for this PR). `Spell Check with Typos`: pass.
- Still pending at review time (not failed, just not finished): `Analyze (c-cpp)`, `Analyze (python)`, `Android NDK build`, `Build on Ubuntu 22.04 RISC-V 64`, `Build on Ubuntu with LLVM/Clang`, `Build with Yocto`, `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04, x86_64)`, `CPP-Linter`, three `Tizen GBS build` jobs, `Ubuntu pdebuild`. None of these had failed as of this review; please confirm they finish green (particularly the x86_64 Valgrind and LLVM/Clang builds, given the return-type note above) before merging.

## Verdict

**OK to merge once the remaining pending CI checks finish green.** All three prior findings are properly addressed, the core leak fix and its ownership contract remain correct, and the previously-blocking required check (`Static checks`) now passes. I found no new Critical or Major issues — only the Minor stylistic note above (Finding 2), which is not blocking given it already compiles clean under one `-Werror=return-type` GCC toolchain.

myungjoo:

## Follow-up review of 332b163e

This commit addresses the one Minor style note from my previous review: `cpp_getModelInfo` and `cpp_eventHandler` now declare `int ret;` before the `try`, assign it inside the `try`, and `return ret;` after the try/catch block, instead of returning from inside the `try`. This now exactly mirrors `cpp_invoke`/`cpp_getFrameworkInfo`'s pattern (explicit trailing return outside the try/catch), and no longer depends on the compiler proving fall-through-freedom through the `_RETURN_ERR_WITH_MSG` macro. `ret` is unconditionally assigned before every path that reads it (every `catch` returns before reaching `return ret;`), so there is no uninitialized-read risk either. Diff confirmed to touch only these two functions in `tensor_filter_support_cc.cc`, nothing else.

CI on this commit: `Static checks` = pass, `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)` = pass (same `-Werror=return-type`/`werror=true` flags as the rest of the matrix), `CPP-Linter` = pass, `Spell Check` = pass, `DCO` = pass, `Build on MacOS-latest with Meson` = pass (this one always completes in ~10s regardless of content — pre-existing #4858 issue, not a meaningful signal). The remaining matrix jobs (x86_64 Valgrind meson build, LLVM/Clang, RISC-V, Android NDK, Tizen GBS ×3, Yocto, Analyze c-cpp/python, Ubuntu pdebuild) were still pending as of this review and had not failed.

## Verdict

No outstanding findings from any of the three review rounds remain. **OK to merge**, pending the rest of the CI matrix finishing green (expected, given Static checks and one full `-Werror`-enabled build already pass on this exact commit).

myungjoo:

## Follow-up review of the myungjoo-bot CHANGES_REQUESTED review (commits b2527a7a, 86b4382a)

Reviewed against `gh api repos/nnstreamer/nnstreamer/pulls/4829/reviews/5021940256`. All three numbered items and all three nits are resolved; one new, already-fixed regression is noted below.

### Item 1 — `getEmptyInstance()` left unguarded — Resolved

`tensor_filter_support_cc.cc:87-97` now wraps the call exactly as the review suggested:
```cpp
tensor_filter_subplugin *obj_raw;
try {
  obj_raw = std::addressof (sp->getEmptyInstance ());
  _RETURN_ERR_WITH_MSG (-EINVAL, e.what ());
} catch (...) {
  _RETURN_ERR_WITH_MSG (-EINVAL, "Unknown exception from getEmptyInstance()");
}
std::unique_ptr<tensor_filter_subplugin> obj_ptr (obj_raw);
tensor_filter_subplugin &obj = *obj_ptr;
```
`obj_raw` is declared before the `try` and read only after it; every path that reaches the read has completed the assignment (both `catch` clauses return unconditionally via the macro), so there is no uninitialized-read risk despite the lack of an initializer. This is the same "declare-before-try, use-after-catch-all-return" shape validated in the previous round's `cpp_getModelInfo`/`cpp_eventHandler` fix, and it now has direct empirical confirmation too: `Build on Ubuntu with LLVM/Clang`, both `Build with meson and test with Valgrind` jobs (x86_64 and arm) all compiled this clean under the project's `werror=true` + `-Werror=return-type` settings.

### Item 2 — Documentation not updated for the API contract change — Resolved

`Documentation/writing-subplugin-tensor-filter.md` gained an "Object ownership and exception rules" section stating the heap-allocated/singly-owned rule, deletion on close or on a throwing `configure_instance()`, and the requirement that the destructor be safe on a partially- or never-configured instance — matching what the review asked for almost verbatim. The PR title/body (`gh pr view 4829`) were also rewritten to name the full scope (all four hardened wrappers, the API contract change, and the two related PRs), addressing the review's first Nit as well.

### Item 3 — Destructor safety on partially-configured instances / #4860 coordination — Resolved

The mock's `configure_instance()` now throws at three selectable points (`THROW_BEFORE_ACQUIRE`, `THROW_AFTER_ACQUIRE`, `THROW_AFTER_RELEASE`), and `openFailNoLeak_n` iterates all three × 100 iterations, asserting `instances_alive`/`resources_alive` back to baseline after each. I checked the `THROW_AFTER_RELEASE` branch specifically — it decrements `resources_alive`/clears `resource_held` before throwing, so the destructor's own `if (resource_held) resources_alive--;` correctly does *not* double-decrement — the three branches together actually exercise "never acquired," "acquired then object destroyed by framework," and "acquired then self-released before throwing," which is exactly the destructor-safety invariant the new contract depends on. Separately, I fetched and read the cross-link comment posted on #4860 (https://github.com/nnstreamer/nnstreamer/pull/4860#issuecomment-5419231810) — it correctly explains why #4860's reworked `cleanup()` (guard removed, ctor-initialized metas, buffer registered before any throwing call) is idempotent and composes safely with this PR's new "destructor always runs" contract, in either merge order.

### Nits 2-3 — Resolved

Tests renamed `openFailNoLeak` → `openFailNoLeak_n`, `openCloseOwnership` → `openCloseOwnership_p`. The mock's statics (`instances_alive`, `resources_alive`, `throw_point`) are now reset in `SetUp()` in addition to `TearDown()`, so a failure in one test case can no longer bleed into the next.

### New finding, already fixed by 86b4382a

b2527a7a introduced a real `clang-format` nonconformance in `unittest_plugins.cc` (the `throw_point` static-member definition and the `points[]` array initializer, lines ~6036 and ~6074-6076), and `CPP-Linter` failed on that commit as a result. I independently reproduced this locally with `clang-format-16 -style=file --output-replacements-xml` on the exact changed line ranges and got two non-empty replacement blocks, confirming the failure was real (not a CI flake). `CPP-Linter` is **not** in `main`'s required-status-checks list, so this did not block the merge button, but it was still a legitimate lint failure.

86b4382a fixes it. I diffed `b2527a7a..86b4382a` directly: it touches only `tests/nnstreamer_plugins/unittest_plugins.cc`, reflowing exactly those two spots (8 lines changed, 4+/4-) with no identifier, logic, or whitespace-outside-those-two-declarations change. Re-running `clang-format-16 --output-replacements-xml` against the full content of both changed files at the new head now returns zero replacements for both — the tree is fully clang-format-clean.

## CI status on head 86b4382a

Pass: `CPP-Linter`, `DCO`, `Spell Check with Typos`, `CodeQL`, `Analyze (c-cpp)`, `Analyze (python)`, `Build on MacOS-latest with Meson`, `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)`, `Build with meson and test with Valgrind in Ubuntu` (both x86_64 and arm), `Build with Yocto`. Still pending, not failed, at review time: `Static checks`, `Android NDK build`, `Build on Ubuntu 22.04 RISC-V 64`, `Tizen GBS build` (x86_64/aarch64/armv7l), `Ubuntu pdebuild`. I waited roughly 20 minutes across two polling rounds for `Static checks` specifically (the one required check most relevant to this file, since it previously failed once on a comment-formatting regression) and it had not finished; nothing failed in that window. Given the source of the comment (`cpp_open`'s doxygen-adjacent comment on line 87) is unchanged since the commit where `Static checks` last passed (42d680b5/b2527a7a), and I independently re-verified the doxygen-tag script's specific rule (no multi-line `/* ` comment without `/**`) is still satisfied in the current file, I expect it to pass, but please confirm before merging rather than take my word for it.

## Verdict

All items from the myungjoo-bot review, all three nits, and the follow-on `CPP-Linter` regression are resolved. **OK to merge once `Static checks` and the remaining pending jobs finish green** — no unresolved Critical/Major/Minor findings from any review round remain.

myungjoo:

## Final delta review (31af0bb7 → c6d719ba)

Both non-blocking notes from the myungjoo-bot APPROVED review (5026139259) are addressed. Diff is exactly the two described changes, nothing else.

**1. Release note — Resolved.** `CHANGES` gained a "C++ subplugin API (tensor_filter_subplugin) behavior change" entry under `2.6.0 -> 2.7.0`, describing the new delete-on-throw behavior and explicitly calling out what out-of-tree vendors must check (a `getEmptyInstance()` that returns a static/pooled/shared instance must change to return a newly allocated object), with a pointer to `Documentation/writing-subplugin-tensor-filter.md`. Indentation matches the file's convention: I checked byte-for-byte — the pre-existing `2.7.0 is a devel version...` line directly above it uses tab indentation, and the new entry uses the same tab/tab-tab nesting, consistent with its immediate section (the older `2.4.4 -> 2.6.0` section uses spaces instead, but that's a pre-existing inconsistency in the file, not something this entry needed to match).

**2. `std::bad_alloc` → `-ENOMEM` — Resolved, ordering verified correct.** New `catch (const std::bad_alloc &e)` arms were added to both guarded blocks in `cpp_open()` (`tensor_filter_support_cc.cc:91` for `getEmptyInstance()`, `:102` for `configure_instance()`), each placed *before* the corresponding `catch (const std::exception &e)` arm. Since `std::bad_alloc` derives directly from `std::exception`, this ordering is required for the arm to be reachable at all (a base-class catch before a derived-class catch would make the derived one dead code, which most compilers flag) — confirmed both arms are correctly ordered before their `std::exception` sibling in both blocks, and unaffected relative to `std::invalid_argument`/`std::system_error`/`std::runtime_error` since those are unrelated branches of the hierarchy.

**No test impact confirmed.** `tests/nnstreamer_plugins/unittest_plugins.cc` is untouched in this diff. The mock's `configure_instance()` throws `std::invalid_argument` at all three throw points, which still lands in the `catch (const std::invalid_argument &e)` arm exactly as before → still maps to `-EINVAL`, so `openFailNoLeak_n`'s `EXPECT_EQ (fw->open (&prop, &private_data), -EINVAL)` assertions are unaffected by the new `bad_alloc` arm.

## CI status on head c6d719ba

Pass: `Static checks`, `CPP-Linter`, `DCO`, `Spell Check with Typos`, `Analyze (python)`, `Build on MacOS-latest with Meson`, `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)`, `Android NDK build`, `Build with meson and test with Valgrind in Ubuntu` (both x86_64 and arm — these are the two jobs that actually execute `testTensorFilterCppSubplugin`), `Tizen GBS build on Ubuntu (armv7l)`. Still pending, not failed, at review time: `Analyze (c-cpp)`, `Build on Ubuntu 22.04 RISC-V 64`, `Build with Yocto`, `Tizen GBS build` (x86_64, aarch64), `Ubuntu pdebuild`.

## Verdict

**OK to merge.** Both remaining notes from the APPROVED review are correctly addressed, no test regressions, catch-arm ordering is correct, and every gating check that has finished (including both jobs that run the leak-regression tests) is green. The handful of still-pending jobs had not failed as of this review.

PR #4837, opened 2026-02-25
[Filter/llama.cpp] loading lora

jaeyun-jung:
Update changes to load lora adapter.
The function to manual free lora adapter is deprecated, and lora should be loaded before creating context.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -90,7 +90,6 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   llama_model *model;
   llama_context *ctx;
   llama_sampler *sampler;
-  llama_adapter_lora *lora_adapter;
 
   int n_processed = 0;
   int n_gpu_layers = 0;
@@ -132,8 +131,7 @@ const char *TensorFilterLlamaCpp::name = "llamacpp";
  * @brief Construct a new llamacpp subplugin instance
  */
 TensorFilterLlamaCpp::TensorFilterLlamaCpp ()
-    : tensor_filter_subplugin (), model (nullptr), ctx (nullptr),
-      sampler (nullptr), lora_adapter (nullptr)
+    : tensor_filter_subplugin (), model (nullptr), ctx (nullptr), sampler (nullptr)
 {
   initializePropertySetters ();
 }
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -262,6 +260,18 @@ TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
     throw std::runtime_error ("Failed to load model");
   }
 
+  /**
+   * Check for LoRA adapter in model_files.
+   * Note that lora adapter must be loaded before creating context.
+   */
+  if (prop->model_files[1] != nullptr) {
+    llama_adapter_lora *lora = llama_adapter_lora_init (model, prop->model_files[1]);
```
songgot, line 268: It's great that LoRa uploads can now be done with just 'llama_adapter_lora_init', without the need for 'llama_set_adapter_lora'. 👍

songgot (approved):

myungjoo (commented):
## Code Review: Memory Leak and Missing LoRA Application

### BUG 1 (HIGH) — Memory leak: `lora` pointer is never freed

```cpp
llama_adapter_lora *lora = llama_adapter_lora_init (model, prop->model_files[1]);
if (lora == nullptr) {
  cleanupResources ();
  throw std::runtime_error ("Failed to load LoRA adapter");
}
// lora is never stored, never freed, never used again
```

The `lora` pointer is a local variable that is never stored anywhere and never freed. The old code stored it as a class member and freed it in `cleanupResources()`. This PR removes both the storage and the cleanup, resulting in a memory leak on every `configure_instance` call.

Unless `llama_adapter_lora_init()` internally transfers ownership to the model (freed when `llama_model_free()` is called), this leaks the adapter struct. Even if the llama.cpp API does take internal ownership, the return value exists specifically for the caller to manage its lifetime — the old code explicitly called `llama_adapter_lora_free()`.

### BUG 2 (HIGH) — LoRA adapter is never applied to the context

The old code called:
```cpp
llama_set_adapter_lora(ctx, lora_adapter, 1.0f);
```
to activate the adapter on the inference context. This PR removes that call entirely. Without it, even though the adapter is initialized, it may have **no effect on inference output**.

The PR moves LoRA loading to before context creation (which may be correct per the llama.cpp API), but the adapter still needs to be applied to the context after context creation. Otherwise LoRA is effectively a no-op.

### Suggested Fix

Either:
1. Keep `lora` as a class member, call `llama_set_adapter_lora(ctx, lora, 1.0f)` after context creation, and restore `llama_adapter_lora_free()` in `cleanupResources()`.
2. Or, if the newer llama.cpp API auto-applies adapters loaded before context creation, verify this is the case and document it in a comment. The `lora` pointer should still be freed (or documented as model-owned).

jaeyun-jung:
Recent changes on llama.cpp
1. Returned ptr after loading lora is valid ptr before releasing model. (model contains it and freed when calling llama_model_free())
2. llama_set_adapter_lora() is removed function.

myungjoo (approved):

PR #4838, opened 2026-02-25
[Filter/llama.cpp] properties for llama.cpp

jaeyun-jung:
Update properties to run llama.cpp and clean up source code.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -95,7 +95,8 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   int n_processed = 0;
   int n_gpu_layers = 0;
   int n_predict = 32;
-  int n_ctx = 2048;
+  int n_ctx = 0;
+  int n_batch = 512;
   int top_k = 0;
   float top_p = TOP_P_DISABLED;
   float typical_p = TYPICAL_P_DISABLED;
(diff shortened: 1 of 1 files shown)


ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -93,9 +93,10 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
   llama_adapter_lora *lora_adapter;
 
   int n_processed = 0;
-  int n_gpu_layers = 0;
+  int n_gpu_layers = -1;
```
songgot: In llama.cpp, -1 means offloading all layers to the GPU. Shouldn't the default be CPU?

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -93,9 +93,10 @@ class TensorFilterLlamaCpp : public tensor_filter_subplugin
...
   int n_predict = 32;
-  int n_ctx = 2048;
+  int n_ctx = 0;
```
songgot, line 98: If n_ctx is 0 and the user does not set the property, llama_context_default_params() sets the context to 0.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
```diff
@@ -265,7 +272,7 @@ TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
   llama_context_params ctx_params = llama_context_default_params ();
   ctx_params.n_ctx = n_ctx;
   /* n_batch is the maximum number of tokens that can be processed in a single call to llama_decode */
-  ctx_params.n_batch = n_ctx;
+  ctx_params.n_batch = n_batch;
```
songgot, line 281: How about add code below?
```
if (n_ctx <= 0) {
  ml_logw (" ");
  n_ctx = default_ctx_value ;
}
if (n_batch <= 0) {
  throw std::invalid_argument ("batch_size must be positive");
}
```

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
jaeyun-jung: text context from model if n_ctx is 0.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
jaeyun-jung: I add code to validate properties.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
jaeyun-jung: updated

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Okay, Thank you.

ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
songgot: Thank you.

songgot (approved):

myungjoo (commented):
## Code Review: Silent Write Failure in saveContextToFile

### BUG (MEDIUM) — `saveContextToFile`: Silent write failures after removing `file.exceptions()`

The PR removes `file.exceptions(std::ofstream::failbit | std::ofstream::badbit)`. Without this call, `file.write()` will **silently fail** on I/O errors (disk full, permission denied, etc.). The `try/catch` block remains but will **never catch anything** because no exceptions are thrown by default — `std::ofstream` sets failbit/badbit silently.

The function will return `true` claiming success even when the context data was not actually written to disk, which could lead to data corruption when the user later tries to load a truncated/empty context file.

### Suggested Fix

Either:
1. Keep `file.exceptions()` but move it **after** the `is_open()` check:
```cpp
if (!file.is_open()) {
  ml_loge(...);
  return false;
}
file.exceptions(std::ofstream::failbit | std::ofstream::badbit);
file.write(...);
```

2. Or remove exceptions entirely but check for write failure explicitly:
```cpp
file.write(reinterpret_cast<const char *>(ctx_data.data()), ctx_size);
if (file.fail()) {
  ml_loge("Failed to write context data to %s", path.c_str());
  file.close();
  return false;
}
```

jaeyun-jung:
@myungjoo updated, moving file.exceptions() after checking is_open().

myungjoo (approved):

PR #4839, opened 2026-02-25
[Test] revise llama.cpp test

jaeyun-jung:
Code clean, remove unnecessary file and revise testcases using llama.cpp.

### DIFF
--- a/tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc
+++ b/tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc
@@ -13,9 +13,48 @@
 #include <gst/gst.h>
 #include <unittest_util.h>
 
+#define LLAMACPP_TEST_MODEL "TinyStories-656K-Q2_K.gguf"
+#define LLAMACPP_TEST_MODEL_URL \
+  "https://huggingface.co/tensorblock/TinyStories-656K-GGUF"
+#define LLAMACPP_LORA_TEST_MODEL "llama-2-7b-chat.Q4_K_M.gguf"
+#define LLAMACPP_LORA_TEST_MODEL_URL \
+  "https://huggingface.co/TheBloke/Llama-2-7B-Chat-GGUF"
+#define LLAMACPP_LORA_TEST_ADAPTER "ggml-adapter-model.gguf"
 
-/* Static variable to track the number of async callback invocations. */
-static int new_sample_count = 0;
+/**
+ * @brief Structure to hold LoRA-related file paths
+ */
+typedef struct {
+  gchar *adapter_path;
+  gchar *model_path;
+} LoRAPaths;
+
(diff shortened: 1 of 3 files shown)


tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc
```diff
@@ -317,347 +310,240 @@ TEST_F (NNStreamerFilterLlamaCppTest, invalidModel_n)
...
-  new_sample_count = 0;
-  create_pipeline (model, invoke_async, "num_predict:5,context_length:0");
-
-  data_push ("Hello");
-  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
-  g_usleep (3000000);
-
-  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
-  EXPECT_GE (new_sample_count, 0);
+  EXPECT_DEATH (create_pipeline (model, TRUE, "num_predict:5,batch_size:0"), "");
```
songgot: create_pipeline() simply returns EXPECT_NE (pipeline, nullptr) if pipeline creation fails. If a crash does not occur, the test fails.

jaeyun-jung: I updated another PR to validate properties. It throws exception if property value is invalid.

Now if batch size is 0 (without checking valid value), llama.cpp goes to crash.

songgot: Okay, thank you.

songgot (approved):

myungjoo (commented):
## Code Review: Test Refactoring

Overall this is a clean and well-structured refactoring. A few minor concerns:

### Concern (LOW) — Reduced `g_usleep` times may cause flaky tests on slow CI

Many tests changed from `g_usleep(3000000)` (3s) to `g_usleep(100000)` (100ms). With the smaller 656K model this is likely fine on fast machines, but 100ms is tight for inference + GStreamer pipeline state transitions on slow CI runners. Consider using a slightly more generous timeout (e.g., 500ms) to avoid flakiness, or use a condition variable / polling approach instead of fixed sleeps.

### Note (LOW) — `create_pipeline` early-return safety

The `g_return_val_if_fail` calls in `create_pipeline()` will return `FALSE` if intermediate `gst_bin_get_by_name` calls fail, leaving some class members set while others are NULL. This is safe in practice because `TearDown()` always calls `clear_pipeline()` which uses `g_clear_pointer` (NULL-safe). Just noting this relies on the GTest fixture guarantee.

myungjoo (approved):

PR #4851, opened 2026-04-14
[CodeClean] fix svace issue

jaeyun-jung:
Code clean, fix svace issue.
1. check passed ptr is not null.
2. prevent invalid dimension value, division-by-zero.
3. cleanup code to check file extension.


myungjoo (approved):

anyj0527 (approved):
LGTM👍

PR #4852, opened 2026-05-19
[spec] Disable trix-engine build

again4you:
This patch disables the trix-engine build since it is not used in public repository.

**Self evaluation:**


jaeyun-jung (approved):

songgot (approved):
Thank you for the great development.

PR #4853, opened 2026-08-19
[Filter] post error when failed to invoke model

jaeyun-jung:
To handle an error case in tensor-filter, post an error message when failed to invoke a model.

### DIFF
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -954,9 +954,23 @@ _gst_tensor_filter_transform_check_invoke_result (GstBaseTransform * trans,
   }
 
   if (invoke_res < 0) {
-    ml_loge_stacktrace
-        ("Calling invoke function (inference instance) of the tensor-filter subplugin (%s for %s) has failed with error code (%d).\n",
+    GstStructure *s;
+    GstMessage *m;
+    gchar *msg = g_strdup_printf
+        ("Calling invoke function (inference instance) of the tensor-filter subplugin (%s for %s) has failed with error code (%d).",
         prop->fwname, TF_MODELNAME (prop), invoke_res);
+
+    ml_loge_stacktrace ("%s", msg);
+
+    /* Post an error message (custom - application) when failed to invoke. */
+    s = gst_structure_new ("nnstreamer-message",
+                "element", G_TYPE_STRING, "tensor-filter",
+                "type", G_TYPE_STRING, "invoke-failure",
+                "message", G_TYPE_STRING, msg, NULL);
+    m = gst_message_new_custom (GST_MESSAGE_APPLICATION, GST_OBJECT (trans), s);
+
+    gst_element_post_message (GST_ELEMENT_CAST (trans), m);
+    g_free (msg);
     return GST_FLOW_ERROR;
   } else if (invoke_res > 0) {
     /* drop this buffer */


myungjoo:

## Review — [Filter] post error when failed to invoke model

Two things are happening in this PR: (a) posting a bus message on invoke failure, and (b) changing invoke failure from a fatal `GST_FLOW_ERROR` into a dropped buffer. (a) is a clean addition; (b) is a pipeline-wide behavior change and is where the concerns are.

### 1. (Critical) Silent behavior change of a public contract, with no opt-out and no docs

```c
return (invoke_res == 0) ? GST_FLOW_OK : GST_BASE_TRANSFORM_FLOW_DROPPED;
```

Until now, "the subplugin's `invoke()` returned a negative value" propagated as `GST_FLOW_ERROR` and stopped the pipeline. After this change, **every** tensor_filter user — every subplugin, ML-API single/pipeline, nnstreamer-edge, Tizen apps — keeps running on a model that is failing on every frame, and only learns about it if it happens to listen for `GST_MESSAGE_APPLICATION` with `element == "tensor-filter"`.

Applications that today rely on the bus `ERROR` (the overwhelmingly common pattern) silently degrade from "pipeline stops, app knows" to "pipeline runs forever producing nothing". That is a regression in observability for existing users, not just a new feature.

Requests:
- Make the new behavior opt-in (or at least opt-out) via a property, e.g. `stop-on-invoke-error` / reuse of the existing error-handling knobs, so existing deployments keep their semantics.
- The new `nnstreamer-message` structure (`element` / `type` / `message` fields, `type = "invoke-failure"`) is now a **public, application-facing contract**. It is not documented anywhere — `Documentation/` has no description of nnstreamer bus messages, and this is the first `gst_message_new_custom()` in the codebase (`gst/mqtt/mqttsrc.c:792` only mentions it in a comment). Please document the structure name and field schema in the tensor_filter documentation in this PR, otherwise the field names will drift and nobody will be able to depend on them.

### 2. (Major) Deleting 5 negative test cases without replacing the coverage

`tests/nnstreamer_filter_lua/runTest.sh` loses cases `7_n` through `11_n`. Those were the only tests in the tree asserting that a type-mismatched tensor_filter invocation makes the pipeline fail. They are being deleted *because* this PR makes them pass buffers instead of failing — which is understandable — but the coverage is not replaced by anything.

The consequence is exactly the failure mode the project cares about: if a future change makes `invoke()` silently succeed with garbage output, or makes the message stop being posted, **no CI job fails**. Please convert those five into positive assertions of the new behavior rather than deleting them, e.g.:
- the pipeline reaches PLAYING and terminates normally (no `ERROR` on the bus),
- **zero** buffers arrive at the sink (all dropped),
- an `APPLICATION` message with `type=invoke-failure` was posted at least once.

A gtest in `tests/nnstreamer_filter_lua/` or `tests/nnstreamer_plugins/` is the easier place for the third assertion since SSAT cannot inspect the bus.

Related: please confirm a full SSAT run. There are ~39 `_n` gstTest cases across `tests/*/runTest.sh` that expect a pipeline launch to fail; most fail at caps negotiation and are unaffected, but any that fail at invoke time will now start "passing" the wrong way.

### 3. (Major) Per-frame stack trace and per-frame bus message on a persistently failing model

```c
ml_loge_stacktrace ("%s", msg);
...
gst_element_post_message (GST_ELEMENT_CAST (trans), m);
```

Previously a failing `invoke()` produced **one** stack trace and the pipeline stopped. Now, with the pipeline kept alive, a model that fails deterministically produces a full `ml_loge_stacktrace()` dump **plus** a `g_strdup_printf()` allocation **plus** a `GstMessage` **for every single buffer** — 30 per second at 30 fps, forever.

Two concrete problems:
- `ml_loge_stacktrace()` is very expensive (symbol resolution / backtrace). At framerate this alone can dominate the pipeline.
- `GstBus` queues messages; an application without a bus watch (or one that filters `gst_bus_timed_pop_filtered()` on a message mask that does not include `APPLICATION`) accumulates them unboundedly. That is an unbounded memory growth path reachable from ordinary "the model is broken" input.

Please rate-limit or latch: post the message and dump the stack trace on the *first* failure (or on a transition from OK→failing), and demote the repeats to `ml_logd`/a counter.

### 4. (Trivial) Style

- `GST_OBJECT (trans)` → `GST_OBJECT_CAST (trans)` to match the surrounding code.
- The `gst_structure_new()` continuation lines are indented at 16 columns; `gst-indent` in CI will likely reflow them.

### Other checks

- **Scope:** appropriately small; no unrelated modules touched. Memory handling in `_gst_tensor_filter_transform_check_invoke_result()` is unchanged for the `invoke_res != 0` path (inputs and non-`allocate_in_invoke` outputs are still unmapped and unref'd before the return), so the DROPPED path does not leak.
- **Security:** the message payload only contains the framework name, the model name and the error code — no new attack surface. No back-door-shaped code found.

myungjoo:

## Re-review of 610fab7

This resolves the three substantive findings from the previous round. Confirmed against the updated diff:

- **`GST_FLOW_ERROR` is restored.** The `(invoke_res == 0) ? GST_FLOW_OK : GST_BASE_TRANSFORM_FLOW_DROPPED` rewrite is gone; `invoke_res < 0` returns `GST_FLOW_ERROR` and `invoke_res > 0` still returns `GST_BASE_TRANSFORM_FLOW_DROPPED` as before. No behavior contract change for existing applications, so the concern about apps silently losing their bus `ERROR` no longer applies.
- **The five lua negative cases are back.** `tests/nnstreamer_filter_lua/runTest.sh` is untouched by the current diff, so cases `7_n`–`11_n` are retained and the coverage loss is undone. The earlier worry about other `_n` cases across `tests/*/runTest.sh` is moot for the same reason.
- **The message/stack-trace flooding concern is moot.** Because the pipeline now stops on the first failure again, `ml_loge_stacktrace()` and the `GstMessage` fire once, not once per buffer. No unbounded `GstBus` growth.

The PR body was updated to match, so the description no longer claims the pipeline keeps running. Good.

The PR is now a clean, additive change: one bus message on an already-fatal path. Two items remain.

### 1. (Minor, still open) The `nnstreamer-message` structure is an undocumented public contract

```c
s = gst_structure_new ("nnstreamer-message",
            "element", G_TYPE_STRING, "tensor-filter",
            "type", G_TYPE_STRING, "invoke-failure",
            "message", G_TYPE_STRING, msg, NULL);
```

This is the first `gst_message_new_custom()` in the codebase — `gst/mqtt/mqttsrc.c:792` only mentions the call in a comment. The moment it lands, applications will match on the literal strings `"nnstreamer-message"`, `"tensor-filter"` and `"invoke-failure"`, and on the `element`/`type`/`message` field names.

Nothing in `Documentation/` describes nnstreamer bus messages, so there is no place a reader can learn this schema, and nothing stops the next contributor from renaming a field. A short section in the tensor_filter documentation stating the structure name, the three fields, and the `type` values that exist so far would fix that, and it costs a few lines now versus a compatibility problem later.

This matters more, not less, now that the change is purely additive: the message is the entire user-visible surface of the PR.

### 2. (Trivial, still open) Style

- `GST_OBJECT (trans)` → `GST_OBJECT_CAST (trans)`, matching `GST_ELEMENT_CAST (trans)` on the next line and the surrounding code.
- The `gst_structure_new()` continuation lines are indented at 16 columns; `gst-indent` will likely reflow them, so running it locally avoids a CI round-trip.

### Unrelated: the Yocto CI failure on this PR is not yours

`Build with Yocto / meta-neural-network on Ubuntu` failed on run 32469757681 with `fatal: unable to connect to git.yoctoproject.org`. That is the connectivity flakiness #4857 fixes with a GitHub-mirror fallback; #4857 has since merged, so a rebase onto current `main` should clear it.

myungjoo (approved):
Approved.

> _Note: the review text below was produced by another (AI) agent and is transcribed here._

The three blocking findings from the earlier review are resolved in 610fab7:

- `GST_FLOW_ERROR` is restored for `invoke_res < 0`, so there is no change to the pipeline-failure contract that existing applications depend on.
- The five negative cases in `tests/nnstreamer_filter_lua/runTest.sh` (`7_n`–`11_n`) are retained — the file is no longer part of the diff.
- With the pipeline stopping on the first failure again, the stack trace and the bus message fire once rather than once per buffer, so the message-flooding and unbounded-`GstBus` concerns no longer apply.

The PR body was updated to match the code. What remains is a clean additive change: one `GST_MESSAGE_APPLICATION` on a path that was already fatal.

Two non-blocking follow-ups:

1. **Please document the `nnstreamer-message` structure.** It is now a public, application-facing contract — the structure name plus the `element` / `type` / `message` fields and the `type` values that exist — and this is the first `gst_message_new_custom()` in the codebase, so there is no precedent for readers to follow. A short section in the tensor_filter documentation would prevent field-name drift later. This is the entire user-visible surface of the change, so it is worth a follow-up PR rather than leaving it undocumented.
2. `GST_OBJECT (trans)` → `GST_OBJECT_CAST (trans)`, matching `GST_ELEMENT_CAST (trans)` on the next line. Running `gst-indent` locally will also reflow the `gst_structure_new()` continuation lines.

Note the red Yocto check on this PR is unrelated: run 32469757681 failed with `fatal: unable to connect to git.yoctoproject.org`. #4857 has since merged with a GitHub-mirror fallback, so a rebase onto current `main` should clear it.

PR #4854, opened 2026-08-20
[Filter/tensorrt10] Fix segfault: do not pass CUDA managed memory downstream

### LINKED ISSUE #4850: Segfault on reading `tensor_filter framework=tensorrt10` outputs via a `queue`
Hey folks, I'm running into a segfault when reading `tensor_filter framework=tensorrt10` outputs via a queue.

# How to Reproduce

A minimal example using a YOLO model [converted to TensorRT](https://docs.ultralytics.com/integrations/tensorrt/#deployment-options-in-tensorrt):

```bash
#!/usr/bin/env bash
gst-launch-1.0 videotestsrc \
! video/x-raw,width=224,height=224,format=RGB \
! tensor_converter \
! tensor_transform mode=transpose option=1:2:0:3 \
! tensor_transform mode=arithmetic option=typecast:float32,div:255   \
! tensor_filter framework=tensorrt10 model=/path/to/yolo26n-cls.engine \
! queue \
! fakesink dump=true
```
```
[...]
00000c60 (0x34c968c60): 68 2b 8a 3a ff a3 9e 38 91 87 8a 3b 34 6e 98 3a  h+.:...8...;4n.:
Caught SIGSEGV
Spinning.  Please run 'gdb gst-launch-1.0 895346' to continue debugging, Ctrl-C to quit, or Ctrl-\ to dump core.
```

The pipeline runs fine either without the `queue` or without buffer content access (`fakesink dump=false`).

myungjoo:
### Problem

Reported in #4850: `tensor_filter framework=tensorrt10` segfaults when its output is consumed via a `queue`, but only when the buffer content is actually read (`fakesink dump=true`).

### Root cause

The subplugin allocates its output tensors with `cudaMallocManaged()` on every `invoke()` and hands the **CUDA managed pointers** directly to the pipeline (`allocate_in_invoke`). With a `queue`, downstream runs in a separate thread and reads the buffer of frame N while the filter thread is already executing frame N+1's inference (`enqueueV3`). On systems without `concurrentManagedAccess` (Windows, **WSL2**, pre-Pascal GPUs), CPU access to managed memory while the GPU is active is invalid and raises SIGSEGV. Without the queue the read happens synchronously while the GPU is idle, and with `dump=false` nothing touches the memory — which is exactly the reported symptom matrix.

### Fix

- Allocate a persistent **device** buffer for each output at load time and set the output tensor addresses once, symmetric with the inputs.
- In `invoke()`, copy the results into plain host memory (`g_malloc`) and hand only host memory to the pipeline. `eventHandler` no longer consumes `DESTROY_NOTIFY`, so the core releases the buffers with `g_free()`.
- Use `cudaMalloc` instead of `cudaMallocManaged` (the buffers are GPU-only now) and drop the unified-memory capability check, which would needlessly reject devices.
- Free the output buffers when `invoke()` fails mid-way; demote the per-invoke info log to debug.

This also removes a per-frame `cudaMallocManaged`/`cudaFree` pair; the added D2H copy is work the managed path did implicitly via page migration anyway.

### Verification (RTX 4070, WSL2 Ubuntu 22.04, CUDA 13.1, TensorRT 10.15.1, GStreamer 1.20.3)

`cudaGetDeviceProperties`: `managedMemory=1, concurrentManagedAccess=0` on this setup.

1. A standalone CUDA program mimicking both memory patterns (per-invoke managed output handed to a consumer thread vs. persistent device buffer + D2H copy into `malloc`ed host memory): the managed pattern **segfaults**, the host-copy pattern runs clean.
2. The reported pipeline itself, with a tiny ONNX classifier (input `[1,3,224,224]`, output `[1,1000]`, built to `.engine` by this subplugin):
   ```
   videotestsrc num-buffers=200 ! video/x-raw,width=224,height=224,format=RGB \
     ! tensor_converter ! tensor_transform mode=transpose option=1:2:0:3 \
     ! tensor_transform mode=arithmetic option=typecast:float32,div:255 \
     ! tensor_filter framework=tensorrt10 model=tiny_cls.onnx ! queue ! fakesink dump=true
   ```
   - **Before** this patch: `Segmentation fault` (exit 139), crashing mid-hexdump exactly like the report.
   - **After** this patch: exit 0, all 200 frames consumed, clean EOS.

Fixes #4850

Related:
- #4861 adds the SSAT regression case for #4850 (and fixes the library-name gate in runTest.sh)
- #4860 fixes resource leaks when configuration fails midway; rebases over this PR
- Suggested merge order: this PR -> #4860 -> #4861

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -313,10 +313,9 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
     throw std::runtime_error ("Invalid output buffer, it is NULL.");
 
   cudaError_t status;
-  std::size_t i;
 
   /* Copy input data to Cuda memory space */
-  for (i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
+  for (std::size_t i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
     const auto &tensorrt10_tensor_info = _tensorrt10_input_tensor_infos[i];
     g_assert (tensorrt10_tensor_info.buffer_size == input[i].size);
 
@@ -339,7 +338,7 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
    * at all. The core releases these buffers with g_free() since the
    * eventHandler does not handle DESTROY_NOTIFY.
    */
-  for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+  for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
     const auto &tensorrt10_tensor_info = _tensorrt10_output_tensor_infos[i];
     g_assert (tensorrt10_tensor_info.buffer_size == output[i].size);
     output[i].data = g_malloc (output[i].size);
@@ -354,7 +353,7 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
 
     /* Copy the results from the persistent device buffers to host memory */
     /** @todo pageable dst makes this copy synchronous; pin it if it matters */
-    for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+    for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
       const auto &tensorrt10_tensor_info = _tensorrt10_output_tensor_infos[i];
       status = cudaMemcpyAsync (output[i].data, tensorrt10_tensor_info.buffer,
           output[i].size, cudaMemcpyDeviceToHost, _stream);
@@ -373,7 +372,9 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
       throw std::runtime_error ("Failed to synchronize the cuda stream");
     }
   } catch (...) {
-    for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+    /* Drain pending D2H copies targeting output[] before freeing it */
+    cudaStreamSynchronize (_stream);
+    for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
       g_free (output[i].data);
       output[i].data = nullptr;
     }
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -329,7 +329,7 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
     }
   }
 
-  /* Allocate plain host memory for the output tensors.
+  /** Allocate plain host memory for the output tensors.
    * The buffers handed over to the pipeline MUST be ordinary host memory:
    * downstream elements may read them from another thread (e.g., after a
    * queue) while this filter is already running the next inference. CPU
@@ -636,7 +636,7 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
 
     } else if (tensorrt10_tensor_info.mode == nvinfer1::TensorIOMode::kOUTPUT) {
 
-      /* Persistent device buffer for the output; invoke() copies the result
+      /** Persistent device buffer for the output; invoke() copies the result
        * into per-frame host memory before handing it to the pipeline. */
       allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
       if (!_Context->setOutputTensorAddress (
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -314,20 +314,6 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
 
   cudaError_t status;
 
-  /* Copy input data to Cuda memory space */
-  for (std::size_t i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
-    const auto &tensorrt10_tensor_info = _tensorrt10_input_tensor_infos[i];
-    g_assert (tensorrt10_tensor_info.buffer_size == input[i].size);
-
-    status = cudaMemcpyAsync (tensorrt10_tensor_info.buffer, input[i].data,
-        input[i].size, cudaMemcpyHostToDevice, _stream);
-
-    if (status != cudaSuccess) {
-      ml_loge ("Failed to copy to cuda input buffer");
-      throw std::runtime_error ("Failed to copy to cuda input buffer");
-    }
-  }
-
   /** Allocate plain host memory for the output tensors.
    * The buffers handed over to the pipeline MUST be ordinary host memory:
    * downstream elements may read them from another thread (e.g., after a
@@ -345,6 +331,20 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
   }
 
   try {
+    /* Copy input data to Cuda memory space */
+    for (std::size_t i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
+      const auto &tensorrt10_tensor_info = _tensorrt10_input_tensor_infos[i];
+      g_assert (tensorrt10_tensor_info.buffer_size == input[i].size);
+
+      status = cudaMemcpyAsync (tensorrt10_tensor_info.buffer, input[i].data,
+          input[i].size, cudaMemcpyHostToDevice, _stream);
+
+      if (status != cudaSuccess) {
+        ml_loge ("Failed to copy to cuda input buffer");
+        throw std::runtime_error ("Failed to copy to cuda input buffer");
+      }
+    }
+
     /* Execute the network */
     if (!_Context->enqueueV3 (_stream)) {
       ml_loge ("Failed to execute the network");
@@ -372,7 +372,7 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
       throw std::runtime_error ("Failed to synchronize the cuda stream");
     }
   } catch (...) {
-    /* Drain pending D2H copies targeting output[] before freeing it */
+    /* Drain pending copies (H2D reads input[], D2H writes output[]) first */
     cudaStreamSynchronize (_stream);
     for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
       g_free (output[i].data);
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -353,6 +353,7 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
     }
 
     /* Copy the results from the persistent device buffers to host memory */
+    /** @todo pageable dst makes this copy synchronous; pin it if it matters */
     for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
       const auto &tensorrt10_tensor_info = _tensorrt10_output_tensor_infos[i];
       status = cudaMemcpyAsync (output[i].data, tensorrt10_tensor_info.buffer,
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -641,13 +641,14 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
       /** Persistent device buffer for the output; invoke() copies the result
        * into per-frame host memory before handing it to the pipeline. */
       allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
+      /* Register the buffer first so cleanup() can free it on failure */
+      _tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
+
       if (!_Context->setOutputTensorAddress (
               tensorrt10_tensor_info.name, tensorrt10_tensor_info.buffer)) {
         ml_loge ("Unable to set output tensor address");
         throw std::runtime_error ("Unable to set output tensor address");
       }
-
-      _tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
 
     } else {
       ml_loge ("TensorIOMode not supported");
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -158,7 +158,6 @@ class tensorrt10_subplugin final : public tensor_filter_subplugin
   void cleanup ();
   void allocBuffer (void **buffer, gsize size);
   void loadModel (const GstTensorFilterProperties *prop);
-  void checkUnifiedMemory () const;
   void convertTensorsInfo (const std::vector<NvInferTensorInfo> &tensorrt10_tensor_infos,
       GstTensorsInfo &info) const;
   std::size_t getVolume (const nvinfer1::Dims &shape) const;
@@ -207,6 +206,11 @@ tensorrt10_subplugin::cleanup ()
     tensorrt10_tensor_info.buffer = nullptr;
   }
   _tensorrt10_input_tensor_infos.clear ();
+
+  for (auto &tensorrt10_tensor_info : _tensorrt10_output_tensor_infos) {
+    cudaFree (tensorrt10_tensor_info.buffer);
+    tensorrt10_tensor_info.buffer = nullptr;
+  }
 
   if (_model_path != nullptr) {
     g_free (_model_path);
@@ -300,7 +304,7 @@ tensorrt10_subplugin::configure_instance (const GstTensorFilterProperties *prop)
 void
 tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
 {
-  ml_logi ("tensorrt10_subplugin::invoke");
+  ml_logd ("tensorrt10_subplugin::invoke");
   g_assert (_configured);
 
   if (!input)
@@ -309,9 +313,10 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
     throw std::runtime_error ("Invalid output buffer, it is NULL.");
 
   cudaError_t status;
+  std::size_t i;
 
   /* Copy input data to Cuda memory space */
-  for (std::size_t i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
+  for (i = 0; i < _tensorrt10_input_tensor_infos.size (); ++i) {
     const auto &tensorrt10_tensor_info = _tensorrt10_input_tensor_infos[i];
     g_assert (tensorrt10_tensor_info.buffer_size == input[i].size);
 
@@ -324,29 +329,54 @@ tensorrt10_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *out
     }
   }
 
-  for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+  /* Allocate plain host memory for the output tensors.
+   * The buffers handed over to the pipeline MUST be ordinary host memory:
+   * downstream elements may read them from another thread (e.g., after a
+   * queue) while this filter is already running the next inference. CPU
+   * access to CUDA managed memory while the GPU is active is invalid on
+   * systems without concurrentManagedAccess (e.g., Windows/WSL2 and
+   * pre-Pascal GPUs) and causes SIGSEGV; device memory is not CPU-visible
+   * at all. The core releases these buffers with g_free() since the
+   * eventHandler does not handle DESTROY_NOTIFY.
+   */
+  for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
     const auto &tensorrt10_tensor_info = _tensorrt10_output_tensor_infos[i];
     g_assert (tensorrt10_tensor_info.buffer_size == output[i].size);
-    allocBuffer (&output[i].data, output[i].size);
-    if (!_Context->setOutputTensorAddress (
-            tensorrt10_tensor_info.name, output[i].data)) {
-      ml_loge ("Unable to set output tensor address");
-      throw std::runtime_error ("Unable to set output tensor address");
+    output[i].data = g_malloc (output[i].size);
+  }
+
+  try {
+    /* Execute the network */
+    if (!_Context->enqueueV3 (_stream)) {
+      ml_loge ("Failed to execute the network");
+      throw std::runtime_error ("Failed to execute the network");
     }
-  }
 
-  /* Execute the network */
-  if (!_Context->enqueueV3 (_stream)) {
-    ml_loge ("Failed to execute the network");
-    throw std::runtime_error ("Failed to execute the network");
-  }
+    /* Copy the results from the persistent device buffers to host memory */
+    for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+      const auto &tensorrt10_tensor_info = _tensorrt10_output_tensor_infos[i];
+      status = cudaMemcpyAsync (output[i].data, tensorrt10_tensor_info.buffer,
+          output[i].size, cudaMemcpyDeviceToHost, _stream);
 
-  /* Wait for GPU to finish the inference */
-  status = cudaStreamSynchronize (_stream);
+      if (status != cudaSuccess) {
+        ml_loge ("Failed to copy from cuda output buffer");
+        throw std::runtime_error ("Failed to copy from cuda output buffer");
+      }
+    }
 
-  if (status != cudaSuccess) {
-    ml_loge ("Failed to synchronize the cuda stream");
-    throw std::runtime_error ("Failed to synchronize the cuda stream");
+    /* Wait for GPU to finish the inference and the output copies */
+    status = cudaStreamSynchronize (_stream);
+
+    if (status != cudaSuccess) {
+      ml_loge ("Failed to synchronize the cuda stream");
+      throw std::runtime_error ("Failed to synchronize the cuda stream");
+    }
+  } catch (...) {
+    for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
+      g_free (output[i].data);
+      output[i].data = nullptr;
+    }
+    throw;
   }
 }
 
@@ -383,19 +413,15 @@ tensorrt10_subplugin::getModelInfo (
 }
 
 /**
- * @brief Override eventHandler to free Cuda data buffer.
+ * @brief Event handler. Output buffers are plain host memory allocated with
+ * g_malloc() in invoke(); returning -ENOENT for DESTROY_NOTIFY lets the
+ * core release them with g_free().
  */
 int
 tensorrt10_subplugin::eventHandler (event_ops ops, GstTensorFilterFrameworkEventData &data)
 {
-  if (ops == DESTROY_NOTIFY) {
-    if (data.data != nullptr) {
-      cudaFree (data.data);
-    }
-
-    return 0;
-  }
-
+  UNUSED (ops);
+  UNUSED (data);
   return -ENOENT;
 }
 
@@ -515,8 +541,6 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
     throw std::runtime_error ("Unable to set GPU device index");
   }
 
-  checkUnifiedMemory ();
-
   // Parse model from .onnx and create .engine if necessary
   std::filesystem::path model_fs_path (_model_path);
   if (".onnx" == model_fs_path.extension ()) {
@@ -598,7 +622,6 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
         throw std::runtime_error ("Unable to set input shape");
       }
 
-      // Allocate only for input, memory for output is allocated in the invoke method.
       allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
 
       // Register the buffer before any call that may throw so that
@@ -612,6 +635,15 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
       }
 
     } else if (tensorrt10_tensor_info.mode == nvinfer1::TensorIOMode::kOUTPUT) {
+
+      /* Persistent device buffer for the output; invoke() copies the result
+       * into per-frame host memory before handing it to the pipeline. */
+      allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
+      if (!_Context->setOutputTensorAddress (
+              tensorrt10_tensor_info.name, tensorrt10_tensor_info.buffer)) {
+        ml_loge ("Unable to set output tensor address");
+        throw std::runtime_error ("Unable to set output tensor address");
+      }
 
       _tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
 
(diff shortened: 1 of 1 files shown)


myungjoo:

## Review — [Filter/tensorrt10] Fix segfault: do not pass CUDA managed memory downstream

The root cause analysis holds up and the fix is the right shape. Verified against the tree:

- `info.allocate_in_invoke = TRUE` (`tensor_filter_tensorrt10.cc:356`), so the core owns the buffers handed out by `invoke()`.
- `gst_tensor_filter_destroy_notify_util()` (`gst/nnstreamer/tensor_filter/tensor_filter_common.c:800-816`) falls back to `g_free (data)` exactly when the V1 `eventHandler` returns `-ENOENT`. So dropping the `DESTROY_NOTIFY` branch and allocating with `g_malloc()` is consistent — no leak, no mismatched free.
- The new `cudaFree` loop in `cleanup()` matches the new `allocBuffer()` call for outputs in `loadModel()` (allocated before `push_back`, so the pointer stored in `_tensorrt10_output_tensor_infos` is the live one). Symmetric with the pre-existing input loop.
- The `try { … } catch (...) { g_free … ; throw; }` block correctly covers every path after the host allocations complete.

Scope is right: one file, no other modules touched, and the size is proportional to the fix.

### 1. (Minor) Pre-existing CUDA leak window is slightly widened

`cleanup()` starts with `if (!_configured) return;`, and `_configured` is only set at the end of `configure_instance()`. So if `loadModel()` throws *after* some buffers are allocated, the destructor's `cleanup()` does nothing and the device allocations leak. That was already true for inputs; this PR adds output allocations to the same window, so a repeated open-with-bad-model loop now leaks roughly twice as much device memory.

Not introduced here and not a blocker, but since the file is already being touched: either free unconditionally in `cleanup()` (all the `cudaFree`/`g_free` calls are nullptr-safe) or set a separate "partially allocated" flag. Note this interacts with #4829, which makes `cpp_open()` `delete` the object on a `configure_instance()` throw — that will start invoking this destructor path in exactly this scenario.

### 2. (Minor) `cudaMemcpyAsync` into pageable host memory

`output[i].data` comes from `g_malloc()`, i.e. pageable memory. CUDA treats D2H `cudaMemcpyAsync` into pageable memory as effectively synchronous, so the "async + one `cudaStreamSynchronize` at the end" structure does not buy the overlap it looks like it buys. Correctness is fine; if the copy shows up in profiling, a pinned staging buffer (`cudaHostAlloc`) copied into the `g_malloc()` block, or registering the block, would be the follow-up. Worth a `@todo` at most.

### 3. (Major) No test guards this fix

This is the real gap. The reported bug (#4850) is a crash that only shows up with `queue` + an actual read of the buffer, and the fix is precisely "never hand non-host memory to the pipeline". Nothing in the tree will notice if that invariant is broken again:

- `tests/` has no `nnstreamer_filter_tensorrt10` directory at all, so there is no gtest and no SSAT case for this subplugin.
- The verification in the PR description is manual and GPU-bound, so it cannot run in the public CI.

Two things would make future regressions detectable:
1. A gtest under `tests/nnstreamer_filter_tensorrt10/` gated on `tensorrt10_support_is_available`, with at least the `… ! tensor_filter framework=tensorrt10 … ! queue ! fakesink dump=true` pipeline from the report. It will be skipped in the public CI, but it will run wherever a GPU runner exists (including the internal infra) and it documents the invariant.
2. Because the whole point is "the pointer handed downstream must be plain host memory", a cheap and CI-runnable assertion is possible even without a GPU: a unit check that `eventHandler(DESTROY_NOTIFY, …)` returns `-ENOENT` for this subplugin. If someone re-adds a `cudaFree`-based `DESTROY_NOTIFY` handler, that test fails immediately and points straight at this bug.

Without something like this, the next contributor who "optimizes away the extra copy" reintroduces #4850 and CI stays green.

### Other checks

- **Regression to other modules:** none — all changes are inside the tensorrt10 subplugin, and the `DESTROY_NOTIFY` fallback path in the core is pre-existing and used by other subplugins.
- **Dropping `checkUnifiedMemory()`:** correct, since managed memory is no longer used. It did double as a "is CUDA sane at all" probe, but `cudaMalloc()` reports those failures anyway, so nothing meaningful is lost.
- **API/ARCHITECTURE docs:** no public API change; nothing to update.
- **Security:** no new attack surface; the change strictly reduces the amount of memory shared across the GPU/CPU boundary. No back-door-shaped code found.

myungjoo:
Tracking note from the #4829 review (AI-assisted, transcribed by the author's tooling):

Since #4829, a `configure_instance()` that throws now gets its instance deleted by the framework, so the destructor runs on partially-configured objects. `tensorrt10_subplugin::cleanup()` opens with `if (!_configured) return;` while `_configured = true` is only set after `loadModel()` returns — so a throw inside `loadModel()` (after `cudaStreamCreate`/`allocBuffer`) leaks the CUDA stream and device buffers.

Since this PR is already editing `cleanup()` (and moves output-buffer allocation into `loadModel()`, widening that window), it is the natural place to address it: drop the `!_configured` early-return and make each free null-safe, as done for dali in #4859.

myungjoo:
_Note: this review was produced by another (AI) agent and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Independent re-review — HEAD `d05168b` (2 commits: fix + doxygen-style follow-up)

This is a fresh pass against the current HEAD, not a repeat of the 2026-08-21 review. Where it reaches the same conclusion as that review, I've kept it brief and just noted agreement; details below are from re-checking the tree myself.

### Does it fix #4850?

Yes, and the mechanism is correctly closed:

- Output tensors now get a **persistent device buffer** allocated in `loadModel()` (`allocBuffer` + `setOutputTensorAddress`, symmetric with the existing input path), instead of a fresh `cudaMallocManaged()` per `invoke()`.
- `invoke()` now allocates `output[i].data` with plain `g_malloc()`, runs `enqueueV3`, then `cudaMemcpyAsync` D2H from the device buffer into that host buffer, then a single `cudaStreamSynchronize`. Only the host pointer is ever handed to `output[i].data`, i.e. only host memory crosses into the pipeline.
- `eventHandler()` no longer intercepts `DESTROY_NOTIFY`; it unconditionally returns `-ENOENT`. I checked `gst_tensor_filter_destroy_notify_util()` in `gst/nnstreamer/tensor_filter/tensor_filter_common.c` — for a V1 subplugin, `-ENOENT` from `eventHandler` makes the core fall back to `g_free(data)`, which matches the `g_malloc()` allocation. No leak, no mismatched allocator.
- Because `invoke()` now serializes "enqueue → D2H copy → sync" before returning, the reused device output buffer is never being written by the *next* `enqueueV3` while a *previous* frame's D2H copy is still in flight — this is exactly the race the bug report describes (queue + downstream thread reads while GPU already runs frame N+1). The fix removes the shared-memory hazard rather than papering over the symptom.
- `checkUnifiedMemory()` is correctly removed along with `cudaMallocManaged` — it existed solely to gate the managed-memory path, which is gone. I don't see any other caller depending on it. The single side-effect (dropping the old "CUDA ≥ 6.0" version sanity check) is inconsequential; TensorRT itself requires far newer CUDA and `cudaMalloc`/`cudaSetDevice` will fail loudly on anything that old anyway.

### Regression risk to other modules

None found. The diff is confined to `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc`, and the `DESTROY_NOTIFY → -ENOENT → g_free` fallback path exercised here is shared, pre-existing core logic already used by other subplugins — this PR doesn't change that logic, only stops using the other branch of it. Agreed with the prior review: no regression surface.

### Code size / scope

Proportional: one file, +69/-73 lines, no unrelated modules touched. Agreed.

### Test coverage — confirmed gap (Major, same conclusion as 2026-08-21 review)

I looked for anything that would actually catch a re-regression and found nothing that would:

- `tests/nnstreamer_filter_tensorrt10/runTest.sh` **does already exist**, but it predates this PR (added in the original TensorRT10 support commit, untouched here) and does not reproduce the bug's trigger condition at all: no `queue` element, and it sinks to `multifilesink` rather than reading buffer content via `fakesink dump=true`. It would not have caught #4850 and won't catch a re-regression of this fix either.
- None of the 18 CI jobs on this PR (`gh pr checks 4854`) build or run anything TensorRT-related — there's no GPU/CUDA runner in the public CI matrix at all, and no workflow file references `tensorrt`. So even if a gtest/SSAT case existed, public CI has no way to execute it today; only manual/GPU-hosted verification (as done in the PR description) exercises this path.

Net effect: if a future change reintroduces a managed/shared buffer here (e.g., someone "optimizes away" the D2H copy), nothing in the repo or CI will notice, and the exact bug in #4850 comes back silently. Concretely actionable, low-cost options that don't require a GPU runner:
1. A gtest/SSAT case gated on `tensorrt10_support_is_available` using the `... ! tensor_filter framework=tensorrt10 ... ! queue ! fakesink dump=true` shape from the issue — skipped in public CI, but documents the invariant and runs anywhere a GPU is actually available.
2. A cheap, GPU-free assertion that `eventHandler(DESTROY_NOTIFY, ...)` returns `-ENOENT` for this subplugin — this alone would fail immediately if someone re-adds a `cudaFree`-based `DESTROY_NOTIFY` handler, and it needs no CUDA device to run in CI.

I understand test coverage and the related cleanup-leak issue below are being tracked/handled in a follow-up PR/session; flagging here only so this review is complete on its own.

### Minor: `cudaMemcpyAsync` D2H into pageable memory

`output[i].data` is `g_malloc()`'d, i.e. pageable host memory. CUDA implicitly staging-buffers/serializes D2H transfers into pageable memory, so the "async copy + one trailing `cudaStreamSynchronize`" structure doesn't give the overlap it visually suggests — it's not wrong, just not actually async in effect. Same conclusion as the prior review; not a blocker, worth at most a `@todo` if it ever shows up in profiling.

### Minor/Nit: buffer leak on mid-loop allocation failure in `loadModel()` (related to, but distinct from, the already-tracked `_configured`/cleanup issue)

Separate from the previously-flagged `_configured` early-return problem in `cleanup()` (tracked via the #4829 note, already being addressed elsewhere): even with that fixed, there's a narrower pre-existing pattern now duplicated for outputs. In the IO-buffer loop:

```cpp
allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
if (!_Context->setOutputTensorAddress (...)) {
  throw std::runtime_error (...);   // buffer never reaches _tensorrt10_output_tensor_infos
}
_tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
```

If `setOutputTensorAddress` fails after `allocBuffer` succeeds, the just-allocated device buffer is never pushed into the vector `cleanup()` iterates over, so it leaks regardless of the `_configured` fix. This exact pattern already existed for inputs before this PR; this PR extends it symmetrically to outputs, so it isn't a new bug class introduced here — just doubled exposure. Given it's already in the same neighborhood as the tracked cleanup fix, worth folding in together (e.g., push_back before checking `setOutputTensorAddress`, or push_back unconditionally and let `cleanup()` free it either way) rather than as its own change.

### Architecture / API docs

No public API or architecture change; nothing to update here. Agreed with the prior review.

### Static checks

Verified the second commit's doxygen-style fix against `.github/workflows/static.check.scripts/doxygen-tag.sh`'s actual matching logic — the two new multi-line comment blocks now correctly open with `/**`, which is what the script's `$line != *"/**"* ` check requires. All 18 CI checks on this PR (including "Static checks") are green.

### Overall

Root cause and fix are correct and appropriately scoped; no regressions found in other modules. The one **Major** item — no test (SSAT/gtest or CI-runnable invariant check) guards this specific invariant, and public CI cannot exercise TensorRT at all — is the same gap already identified in the prior review and reportedly being tracked separately. The two **Minor** items (pageable-memory async copy, and the mid-loop leak-on-failure duplicated to outputs) are not blockers for this fix itself.

myungjoo:
Author response to the review above (drafted with AI assistance):

1. **Missing regression test (Major)** — agreed; being addressed in a separate PR (in progress) that fixes the `tests/nnstreamer_filter_tensorrt10/runTest.sh` library-name gate (`grep tensorrt.so` does not match `tensorrt10.so`) and adds the `queue ! fakesink dump=true` reproduction case from #4850. Kept out of this PR to keep the crash fix minimal.
2. **Pageable D2H `cudaMemcpyAsync` (Minor)** — resolved in this PR: commit 8932eec4 adds a `@todo` documenting that the copy is effectively synchronous and pointing at a pinned staging buffer as the follow-up if profiling warrants it. No behavioral change intended.
3. **Leak when `setOutputTensorAddress` fails after `allocBuffer` (Minor/Nit)** — agreed, but deferring to the cleanup/`_configured` follow-up PR (in progress): freeing that buffer is only observable once `cleanup()` stops early-returning on unconfigured instances, and fixing it here would conflict with that work touching the same region.

myungjoo:
_Note: this review was produced by another (AI) agent and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Re-review — HEAD `8932eec4`

Checked the new commit and the author's response (https://github.com/nnstreamer/nnstreamer/pull/4854#issuecomment-5406947107) against the three points from my previous pass.

**Commit `8932eec4` diff** (verified directly, `git diff d05168b..8932eec4`):
```cpp
     /* Copy the results from the persistent device buffers to host memory */
+    /** @todo pageable dst makes this copy synchronous; pin it if it matters */
     for (i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
```
Comment-only, no logic change. It's a single-line `/** ... */` block, which is fine under `doxygen-tag.sh`'s rule (the "multi-line comment must open with `/**`" check only fires on lines containing `/*` that *don't* contain `/**`; this line does).

### Item 1 — Missing regression test (Major): deferred, judged acceptable

Deferred to a separate in-progress PR that (a) fixes the SSAT gate in `tests/nnstreamer_filter_tensorrt10/runTest.sh` and (b) adds the `queue ! fakesink dump=true` reproduction from #4850.

I independently checked the specific bug the author cites — the existing gate does `grep tensorrt.so`, and I confirmed with a direct regex test that `tensorrt.so` (with `.` as a single-char wildcard) does **not** match `tensorrt10.so` or `libnnstreamer_filter_tensorrt10.so`. So that SSAT case has apparently never actually exercised the tensorrt10 plugin at all (it always falls into the "Cannot find TensorRT shared lib" early-exit) — this is a real, separate, pre-existing defect, not a rationalization. Fixing it in the same follow-up PR that adds the regression case makes sense (fix the gate, then rely on it to actually run the new case) and is a legitimate reason to keep it out of this crash-fix PR: the two changes are logically distinct (a broken test gate vs. this pointer-lifetime fix) and public CI has no GPU runner regardless, so deferring doesn't change what CI can catch today either way. Acceptable as a tracked follow-up, not a blocker for this PR.

### Item 2 — Pageable D2H `cudaMemcpyAsync` (Minor): resolved in-PR

The `@todo` added in `8932eec4` documents exactly the effect I flagged (copy is effectively synchronous due to pageable destination) and points at the correct follow-up (pinned staging buffer) if it ever matters. This was explicitly a "worth a `@todo` at most" item in my original review — fully addressed as specified.

### Item 3 — Leak on `setOutputTensorAddress` failure after `allocBuffer` (Minor/Nit): deferred, judged acceptable

Deferred to the same follow-up that fixes `cleanup()`'s `_configured` early-return. The author's reasoning — that this leak is only observable once `cleanup()` stops bailing out early on a partially-configured instance, and that fixing it here would edit the same lines the other in-progress PR is about to touch — is sound. Landing both together avoids a rebase collision and a half-fixed intermediate state. Fine to track separately.

### Scope / regression check on the new commit

Single comment line added, zero behavioral change, still confined to `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc`. No new regression surface.

### CI

At the time of this re-review, checks were still re-running for the new commit (`Static checks`, the various `Build on ...` jobs, etc. showed `pending`, with `DCO` and `Build on MacOS-latest` already green). Given the change is comment-only, I'd expect the same all-green result as before, but worth a quick glance before merge to confirm.

## Verdict

All three points from the prior review have been addressed — one in-PR, two deferred to specific, already-scoped follow-up PRs for defensible reasons (keeping this crash fix minimal, and avoiding overlapping edits with the cleanup/`_configured` work). I have no remaining blocking issues against this PR's diff.

**All review points addressed — ok to merge** (once the currently-running CI checks finish green).

myungjoo:
Author response to the automated review above (drafted with AI assistance):

1. **`catch (...)` frees host memory with D2H copies still enqueued** — fixed in this PR: commit 08a8cf0b adds `cudaStreamSynchronize (_stream)` at the top of the catch block (with a comment tying it to the pinned-buffer `@todo`), so the drain stays correct if the copy ever becomes truly asynchronous.
2. **Output device buffer leak when `setOutputTensorAddress()` fails** — deferring to #4860, which owns exactly this failure path ("resource leak when configuration fails midway"). The push_back-first fix only becomes observable once #4860 removes the `!_configured` early-return in `cleanup()`, and #4860 must rebase over this PR's output-allocation change anyway — that rebase is where the output side will get the same push_back-before-set treatment its input side already has.
3. **Cross-links** — added to the PR description: #4861 adds the SSAT regression case for #4850 (and fixes the `grep tensorrt.so` gate that never matched `tensorrt10.so`); #4860 covers partial-configure cleanup. Suggested merge order: this PR → #4860 (rebase + output-side push_back-first) → #4861 (rebase runTest.sh).
4. **Nit: hoisted `std::size_t i`** — fixed in 08a8cf0b; all loops are back to loop-scoped indices, the catch block uses its own.
5. **Nit: input H2D loop outside try** — no action, as noted it is not live; it should be revisited together with the pinning `@todo`.

myungjoo:
_Note: this review was produced by another (AI) agent and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Re-review — HEAD `08a8cf0b`, against the myungjoo-bot review round (2026-08-25 17:33) and the author response

Verified `git diff 8932eec..08a8cf0b` directly and re-read the full `invoke()` function at HEAD. Going through the bot's findings one by one:

### 1. `catch (...)` freeing host memory with copies still enqueued — fixed correctly

The new code:
```cpp
  } catch (...) {
    /* Drain pending D2H copies targeting output[] before freeing it */
    cudaStreamSynchronize (_stream);
    for (std::size_t i = 0; i < _tensorrt10_output_tensor_infos.size (); ++i) {
      g_free (output[i].data);
      output[i].data = nullptr;
    }
    throw;
  }
```
This closes the gap exactly as described: any D2H copies already enqueued on `_stream` for earlier `output[j].data` (j < the index that threw) are now drained before those buffers are freed, so freeing can't race a still-pending GPU write into that memory. Ignoring the `cudaStreamSynchronize` return value here is correct, not sloppy — we're already unwinding via an active exception, there is no meaningful recovery action, and the alternative (throwing a second exception, or swallowing the original one to report the sync failure) would be worse. The comment ties it to the pageable-memory `@todo` on the D2H loop, so if that TODO is ever acted on (pinning the host buffer), this drain is exactly what keeps the exception path safe — good forward-proofing. This is a real fix, not cosmetic: today the risk was latent (masked by pageable-memory D2H being effectively synchronous), and the added sync makes it correct independent of that CUDA implementation detail.

### 2. Output device buffer leak on `setOutputTensorAddress()` failure — deferred to #4860, reasoning holds up

Same conclusion as my previous pass: this is a pre-existing pattern (mirrors the input path) extended symmetrically to outputs. The author's rationale — the leak is only observable once `cleanup()` stops early-returning on `!_configured` (which #4860 addresses), and #4860 has to rebase over this PR's output-allocation code anyway — is sound. Doing the push-before-set fix here would very likely get re-touched or conflict with #4860's own edits to the same lines. Deferring with the coordinated merge order in the PR description (this PR → #4860 → #4861) is a reasonable resolution, not a bypass.

### 3. No automated regression gate — cross-linked, appropriately deferred

The PR description now links #4861 (SSAT regression case for #4850, plus the fix for the `grep tensorrt.so` gate that never matched `tensorrt10.so` — I verified that regex mismatch independently in my first review pass) and #4860, with an explicit suggested merge order. That satisfies the ask: the gap is tracked, discoverable, and sequenced sensibly rather than silently dropped.

### Nits

- Hoisted `std::size_t i` — reverted to loop-scoped declarations everywhere, confirmed via diff. Matches the rest of the file's style again.
- Input H2D loop outside the `try` — no action taken; confirmed this is correct as-is (no output buffers exist yet at that point, so there's nothing to free even if it throws), and deferred together with the pinning `@todo` is the right call rather than fixing it in isolation.

### Scope / regression check on `08a8cf0b`

One functional line added (`cudaStreamSynchronize (_stream);` in the catch block) plus a comment and reverting the index-hoisting refactor. Still confined to `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc`. No new regression surface, and this change makes the failure path strictly safer than before.

---

## CI status — blocking, but unrelated to this PR's diff

**"Static checks" is currently failing** on this commit. I pulled the job log and the actual failure is:

```
Static checks	/Checker/ check-rebuild action self-test	bash: .github/actions/check-rebuild/test_check_if_rebuild_requires.sh: No such file or directory
Static checks	/Checker/ check-rebuild action self-test	##[error]Process completed with exit code 127.
```

I traced this: `main` has since gained commit `b57d9654` ("[CI] Add check-rebuild self-test..."), which adds both a new step to `static.check.yml` *and* the script it calls (`.github/actions/check-rebuild/test_check_if_rebuild_requires.sh`) in the same commit. `static.check.yml`'s checkout step is pinned to `ref: ${{ github.event.pull_request.head.sha }}` (PR head only, not merged with base), so it checks out this PR branch's tree as-is — which predates `b57d9654` and therefore doesn't have the new script — while GitHub still runs the *workflow definition* from current `main`, which now references that script unconditionally. The result: every open PR that hasn't rebased past `b57d9654` will fail this exact step, regardless of what the PR itself changes. I confirmed the script and the workflow step both exist on `origin/main` and are simply absent from this branch's tree.

This is **not a defect introduced by this PR's diff** — the change under review never touches `.github/` — but it is a real blocker for merging as-is, since "Static checks" is a required-looking CI job and is currently red. Rebasing this branch onto current `main` should pick up the missing script and clear it; no code change to `tensor_filter_tensorrt10.cc` is needed for this.

## Verdict

All three substantive points from the bot review are handled appropriately (one fixed in-PR with a correct and verifiable change, two deferred with sound reasoning and now cross-linked), and both nits are resolved or reasonably deferred. **No remaining code-review blocking issues against this PR's diff.**

The PR is not yet mergeable as-is only because of the unrelated CI infra gap above (**Major, but out of scope of this diff**) — rebase onto latest `main` to pick up `b57d9654`, confirm "Static checks" goes green, then this is **ok to merge**.

myungjoo:
Author response to the 4th-round review above (drafted with AI assistance):

**Point 2 (output buffer leak on `setOutputTensorAddress()` failure) — now fixed in this PR.** Commit 7d985643 moves the `push_back` to right after `allocBuffer()`, before `setOutputTensorAddress()` can throw, mirroring the input-side treatment in #4860. You were right that deferring it to #4860 left a gap: its diff predates this PR, so its reordering never touched the output branch this PR introduces. With this commit the PR is self-contained — the change that adds the output-side `allocBuffer` also keeps the buffer reachable by `cleanup()`. As noted, this becomes observable once #4860 drops the `!_configured` early-return; #4860 has been notified that its rebase no longer needs to re-apply anything on the output side.

The remaining notes (no hosted GPU gate — mitigated by the invariant comment and #4861; H2D loop outside the try — to be revisited together with the pinning `@todo`) are acknowledged as-is, no code change.

myungjoo:
_Note: this review was produced by another (AI) agent and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Re-review — HEAD `7d985643`, against the second myungjoo-bot round (2026-08-26 02:01)

Fetched the branch and diffed `c8caebeb..7d985643` directly:

```cpp
       allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
+      /* Register the buffer first so cleanup() can free it on failure */
+      _tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
+
       if (!_Context->setOutputTensorAddress (
               tensorrt10_tensor_info.name, tensorrt10_tensor_info.buffer)) {
         ml_loge ("Unable to set output tensor address");
         throw std::runtime_error ("Unable to set output tensor address");
       }
-
-      _tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);
```

This is the single item the bot withheld approval for (its point 2), and it also called out a real hole in the original plan: #4860's own reorder is written against the pre-#4854 file and only touches the input branch, so a straight rebase would not have picked up the output side (which didn't have `allocBuffer` before this PR existed). Checking the specifics asked about:

**Equivalence of using the local `tensorrt10_tensor_info` after `push_back`.** `NvInferTensorInfo` is a plain-data struct (`const char*`, enums, two `size_t`, one `void*`) with no owning members or custom copy semantics, so `push_back` does a trivial value copy. The local variable and `_tensorrt10_output_tensor_infos.back()` hold identical field values afterward, including the same `buffer` pointer value — so the subsequent `setOutputTensorAddress(tensorrt10_tensor_info.name, tensorrt10_tensor_info.buffer)` call against the local is functionally identical to calling it against the vector's copy. No aliasing bug.

**Reachability by `cleanup()` on failure.** With the buffer now in the vector before `setOutputTensorAddress()` can throw, `cleanup()`'s output loop (`for (auto &t : _tensorrt10_output_tensor_infos) cudaFree(t.buffer)`) will free it — *provided `cleanup()` gets past its `if (!_configured) return;` guard*. On the very first `configure_instance()` call, `_configured` is still false when `loadModel()` throws, so that guard still short-circuits `cleanup()` today — this half of the problem is, as the author's response also states, correctly left to #4860 (which removes that guard). What this commit fixes, independent of that: once #4860 lands (or in a reconfigure scenario where `_configured` is already `true` from a prior successful `loadModel()` — note `loadModel()` unconditionally does `_tensorrt10_output_tensor_infos.clear()` at the top, so a failing second call starts from an empty vector each time), the failing buffer is now reachable instead of orphaned. That matches exactly what the bot's point 2 asked for — it's a fix to "the buffer never entered the vector," not a claim to also fix the separate `_configured`-gate problem, which stays correctly scoped to #4860.

**Double-free.** `cudaFree` is only ever invoked from the one loop in `cleanup()`, which iterates `_tensorrt10_output_tensor_infos` exactly once per entry; nothing else frees this pointer, and the per-iteration local `tensorrt10_tensor_info` has no destructor. No double-free risk from moving the `push_back` earlier.

**Consistency note (non-blocking):** the input branch above still does `allocBuffer → setInputTensorAddress → push_back`, i.e. it keeps the same latent gap the bot separately flagged as pre-existing and out of scope for this PR. That asymmetry is intentional here — the coordinator's framing (fixing only the output-side leak this PR introduces, leaving the input-side pre-existing instance to #4860) matches the bot's own scoping of point 2 to "the output side ... this PR is what introduces." Fine as-is.

The author's response (https://github.com/nnstreamer/nnstreamer/pull/4854#issuecomment-5424745438) and the updated coordination note on #4860 both state the same understanding independently arrived at above (fix is self-contained now; becomes fully effective once #4860 drops the `_configured` guard). I have no correction to that framing.

### Residual non-blocking notes — unchanged, correctly left as-is

- No hosted-GPU CI gate for this invariant: mitigated by the doc comment at `:326-335` and tracked via #4861. Same conclusion as prior rounds.
- H2D input loop outside the `try`: not a live issue (nothing to free yet at that point), deferred together with the pinning `@todo` as before.

### CI

`Static checks` — which was failing in the previous round due to an unrelated, repo-wide CI infra gap (a stale branch missing a script newly added on `main`) — is now green after the branch picked up `main`'s CI fixes (`b57d9654`, `988a128e`) via rebase; commit SHAs for the fix itself changed accordingly (`50c3a59a→11b88622`, etc., same content). Remaining jobs were still finishing at the time of this check but nothing has failed.

## Verdict

All points raised across both bot review rounds are now addressed: the segfault fix is correct, the exception-path stream-drain fix (round 3) is correct, and this round's output-buffer-registration fix correctly closes the vector-membership gap in isolation, with the remaining `_configured`-gate half of the story correctly left to #4860 and now explicitly coordinated (its comment updated to say the output side needs no re-application). No remaining code issues against this PR's diff, and the CI blocker from the previous round is resolved.

**All review points addressed — ok to merge.**

myungjoo:
Author update (drafted with AI assistance): rebased onto current main, which now contains #4860 and #4861 (both merged), and pushed two new commits addressing the remaining review items:

1. **Output device buffer leak (the blocking finding)** — now fully closed. The `push_back()`-before-throw reorder for the output branch was already in `Register output buffer before throwing calls`; after the rebase, #4860's removal of the `if (!_configured) return;` guard makes `cleanup()` actually reach those entries, so the mid-configuration leak path the reviewer traced is closed end-to-end. The rebase also resolved the "#4860's input-only reorder won't cover the output branch" concern: this PR owns the output branch shape directly.
2. **H2D loop outside the `try`** (`Cover H2D copies with the failure drain path`) — host output buffers are now allocated before the `try`, and the H2D copy loop moved inside it, so the existing catch (stream drain + `g_free`) covers every async operation in `invoke()` uniformly. This also future-proofs the pinned-buffer `@todo`.
3. **Symmetry follow-up** (`Clear output infos in cleanup() for idempotency`) — `_tensorrt10_output_tensor_infos` is now cleared after its `cudaFree` loop, matching the input side from #4860.

Note on merge order: #4861's regression case is already on main while this fix is not, so on TensorRT10-capable machines main's SSAT currently reproduces #4850 until this PR merges.

myungjoo:
> _Note: this review was produced by an automated review agent (a Claude sub-agent) and is transcribed here as-is. Please verify each point before acting on it._

## Re-review — HEAD `a44861d7` (post-rebase onto main, which now includes #4860 and #4861; 2 new commits since the last bot review at `8932eec4`/`c8caebeb`)

Read the prior bot review (review id 5026087711, at head `8932eec`) and the author's "Author update" comment, then re-verified the current tree (`ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc`, 830 lines) directly — `cleanup()`, both branches of the IO loop in `loadModel()`, and all of `invoke()` — plus `git diff origin/main...HEAD` to confirm scope, and `git log -- tests/nnstreamer_filter_tensorrt10/` to confirm #4861's test-gate fix is present on this branch post-rebase.

### 1. (Was Major/blocking) Output device buffer leak if `setOutputTensorAddress()` fails — **Resolved**

Commit `e75c31cd` reorders the output branch in `loadModel()` (`:640-652`):

```cpp
allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
/* Register the buffer first so cleanup() can free it on failure */
_tensorrt10_output_tensor_infos.push_back (tensorrt10_tensor_info);

if (!_Context->setOutputTensorAddress (...)) {
  throw std::runtime_error ("Unable to set output tensor address");
}
```

This is now symmetric with the input branch just above it (`:628-638`, from #4860). Combined with `cleanup()` (`:198-225`), which — confirmed by direct read — no longer has an `if (!_configured) return;` guard and unconditionally walks both `_tensorrt10_input_tensor_infos` and `_tensorrt10_output_tensor_infos` calling `cudaFree`, a throw from `setOutputTensorAddress()` now leaves the buffer reachable: `configure_instance()` propagates the throw, the framework deletes the partially-configured instance (per #4829), the destructor runs `cleanup()`, and the just-allocated output buffer is freed via the vector it was pushed into before the throw. Traced the full path; no leak.

### 2. (Non-blocking) H2D loop sat outside the `try` block — **Resolved**

Commit `17189d05` moves the input H2D copy loop inside the `try` (now `:336-347`), after hoisting the output `g_malloc` loop above the `try` (`:328-332`, so every `output[i].data` is a valid pointer before any throw site the catch would need to free). The `catch (...)` block (`:375-383`) now does `cudaStreamSynchronize (_stream)` before freeing `output[]`, and its comment was updated to "Drain pending copies (H2D reads input[], D2H writes output[]) first" — accurately describing both directions now that both are inside `try`. This also happens to resolve the prior review's minor nit about `std::size_t i` being hoisted to function scope: each loop now redeclares `std::size_t i` in its own scope again (`81a70fb1`), matching the rest of the file's style.

Traced the failure modes:
- H2D copy fails on iteration *k*: nothing was queued for iteration *k* (the async submission itself failed), earlier iterations' copies may still be in flight → `cudaStreamSynchronize` drains them before `output[]` is freed. Correct.
- D2H copy or `enqueueV3` fails: same drain-then-free path as before, unchanged behavior, still correct.
- No output tensors (edge case, size 0): loops and drain are no-ops, no issue.

No new correctness issues from the reordering — I checked for uninitialized-use and missing-code risems specifically (per the request, since this file isn't compiled by any CI lane) and found none: `status` is declared once at function scope and only read immediately after each assignment, `i` no longer leaks scope, and the doc comment above the output `g_malloc` loop (`:318-327`) still accurately describes the ownership contract now that it sits before `try` rather than after the (now relocated) H2D loop.

### 3. (Non-blocking) No automated regression gate — **Resolved separately, not in this PR's scope**

This was already tracked as a separate follow-up in the prior rounds. Confirmed via `git log --oneline -- tests/nnstreamer_filter_tensorrt10/` that #4861 (`637dbbef` "Fix SSAT lib gate and add issue #4850 regression case", plus follow-ups `36594a05`, `747ef2fc`, `6bea3717`) is present in this branch's history after the rebase onto main, and `tests/nnstreamer_filter_tensorrt10/runTest.sh` now gates on `tensorrt10.so` (not the old non-matching `tensorrt.so` pattern) and includes the `queue`/`fakesink dump=true` reproduction. Nothing further needed from this PR.

### New commit: `Clear output infos in cleanup() for idempotency` (`a44861d7`)

```cpp
for (auto &tensorrt10_tensor_info : _tensorrt10_output_tensor_infos) {
  cudaFree (tensorrt10_tensor_info.buffer);
  tensorrt10_tensor_info.buffer = nullptr;
}
_tensorrt10_output_tensor_infos.clear ();
```

Correct and exactly symmetric with the pre-existing input-side `_tensorrt10_input_tensor_infos.clear ()` a few lines above (from #4860). Checked idempotency: `cleanup()` is only invoked once in this file (from `~tensorrt10_subplugin()`), so there's no live double-cleanup call site today, but the change is still the right defensive move — a second `cleanup()` call (e.g. if a future change adds an explicit reload/reset path) would now iterate empty vectors and no-op cleanly instead of leaving dangling `NvInferTensorInfo` entries with already-freed `buffer` pointers around. No issue found.

### Rebase integration check

`git diff origin/main...HEAD --stat` shows only `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc` changed (84/84 lines), confirming scope is still confined to this one file. #4860's changes (constructor `gst_tensors_info_init`, no `_configured` guard in `cleanup()`, input push-before-throw reorder, `_stream` nulled after destroy) are all present and correctly interleaved with this PR's changes — no conflicting or reverted hunks found.

### New issues found

None. No new correctness, style, or scope issues in the two new commits or in how they interact with the already-merged #4860/#4861.

### CI

At review time, most build jobs were still `pending` (freshly pushed after rebase); `Static checks`, `CPP-Linter`, `DCO`, and `Spell Check` are already green. The change is confined to a file no public CI lane compiles (TensorRT/CUDA-only), consistent with prior rounds — worth a glance once the pending jobs settle, but nothing here is expected to fail given the diff.

## Verdict

All items from the prior review round — the one blocking (output buffer leak on `setOutputTensorAddress()` failure) and the two non-blocking (H2D-outside-try, missing automated gate) — are resolved: the first two in this PR's new commits, the third by the already-merged #4861. No new issues introduced by the rebase or the two new commits.

**LGTM — ok to merge** (once the currently-pending CI checks finish green).

Issue #4849, opened 2026-04-06
Any plans on migrating TFLite to Litert?


PrathamTI:
Hello team 
TensorFlow Lite versions after 2.20 have been deprecated [0] and shifted completely to LiteRT [1]. 
As LiteRT is now the primary evolution of the project, I wanted to inquire about the future plans of nstreamer. Specifically: 
1. Are there existing plans to migrate NNStreamer’s TFLite filters/plugins to support LiteRT?
2.  Is this migration being considered for the next LTS release?

[0] https://github.com/tensorflow/tensorflow/releases/tag/v2.20.0
[1] https://github.com/google-ai-edge/LiteRT

Regards

myungjoo:
Thanks for raising this, and apologies for the slow response. Yes, we are aware of the TFLite → LiteRT transition, and here is our current position:

**Short term — nothing breaks.** The `tensorflow2-lite` filter uses the classic C++ Interpreter API (`FlatBufferModel`, builtin ops, `ModifyGraphWithDelegate` with XNNPACK/GPU/NNAPI/external/QNN delegates), which LiteRT v1 supports as a drop-in replacement, and the legacy `tensorflow/lite/` include paths are kept in LiteRT for backward compatibility. NNStreamer detects the runtime via pkg-config (`tensorflow2-lite`), so building the runtime from the LiteRT repository instead of the TF repository should require only build-script-level changes, not source changes. We plan to verify this and, if needed, add a `litert-support` build option.

**Mid term — a new subplugin for LiteRT Next.** The new `CompiledModel` API (`litert::Environment` / `Model` / `CompiledModel` / `TensorBuffer`) is a genuinely different API with unified GPU/NPU acceleration (accelerator flags instead of delegates), so it warrants a separate tensor_filter subplugin (tentatively `framework=litert`), coexisting with the current one — the same pattern we used for `tensorrt` vs `tensorrt10`. All future performance/NPU work from Google lands only on that side, so this is where the long-term value is.

Regarding your second question: there is no dated commitment for a specific LTS release yet. We are tracking this in our backlog, and contributions are very welcome — the short-term item in particular is a well-scoped, build-system-only task.

Which platform are you targeting (Ubuntu / Android / Tizen)? That helps us prioritize the packaging side.

myungjoo:
A correction and an update to my earlier comment, after verifying against the current LiteRT releases:

[...]

PR #4855, opened 2026-08-20
[Filter/litert] Add LiteRT 2.x (CompiledModel API) subplugin

myungjoo:
This adds a new tensor_filter subplugin, `framework=litert`, targeting the LiteRT 2.x **CompiledModel C API** — the actively developed successor line of TensorFlow Lite (LiteRT Next, GA since early 2026). It is a separate subplugin from `tensorflow2-lite` (classic Interpreter API); both coexist so users can A/B the two runtimes on the same `.tflite` model, following the `tensorrt`/`tensorrt10` precedent.

Addresses the mid-term plan discussed in #4849. The Yocto CI fixes found while stabilizing this PR are split into #4857.

## Design
- **C API only** (`litert/c/*.h`): the sole runtime dependency is `libLiteRt.so` / `libLiteRt.dylib` (ABI-versioned, `VERS_1.0`). No absl/flatbuffers/cc-SDK build needed.
- **Detection**: pkg-config `litert.pc` first, then `LITERT_ROOT` env var fallback (same pattern as `snpe-support`), plus a compile/link sanity check. `-Dlitert-support=auto|enabled|disabled`. Both detection paths verified.
- **Tensor meta**: element type/name from the model signature; **dimensions from the compiled model's runtime-resolved layouts**, so models declaring dynamic batch (e.g., `[-1,1]`) resolve to concrete shapes — matching classic `AllocateTensors()` semantics.
- **Invoke**: managed tensor buffers created once from compiled-model buffer requirements, reused across invokes with lock/unlock + memcpy (accelerator-agnostic). Zero-copy host wrapping is a follow-up TODO.
- **Properties**: `custom=Accelerators:cpu|gpu|npu` (combinable with `+`), `custom=Signature:<key>`; the standard `accelerator=` property is honored too.

## Build script support
- Ubuntu/Tizen/Yocto: the meson option covers all three (auto-disables silently when the SDK is absent — no packaging changes required at this stage).
- macOS: same meson option; the subplugin builds as `.dylib` and the loader accepts it.
- Android: `NNSTREAMER_FILTER_LITERT_SRCS` added to `jni/nnstreamer.mk` for ndk-build consumers.

## Tests
**gtest** — `tests/nnstreamer_filter_litert/unittest_filter_litert.cc`, 21 cases (positive/negative), coverage exceeding the existing tflite2/onnxruntime suites:
- golden classification results (orange.png → argmax 951) **shared with the tensorflow2-lite unit tests on the same model files**, so any cross-runtime divergence fails at unit test level with a descriptive gtest message
- determinism check (two invokes, identical input → identical output), output-canary overwrite check
- 32-in/32-out model (dynamic batch resolution path), quant + float models, reopen with different model
- negatives: invalid path, garbage flatbuffer, no model, null tensor memories, unknown accelerator, malformed custom property, unknown signature key, wrong pipeline dims/type, unsupported getModelInfo ops

**SSAT** — `tests/nnstreamer_filter_litert/runTest.sh`, 6 cases: golden image-labeling (same model/golden label as the tensorflow2-lite SSAT tests), `Accelerators:cpu`, and negative cases (wrong dims, invalid model, unknown accelerator).

## CI coverage for litert (verified green)
- **Ubuntu x86_64 (meson job)**: installs the LiteRT SDK (headers from `litert_cc_sdk.zip` v2.2.0, `libLiteRt.so` from the official `ai-edge-litert` wheel), then runs the litert gtest suite via `meson test` (21/21 observed), the gtest suite again under Valgrind, and the litert SSAT suite.
- **macOS arm64 (macos job)**: installs the SDK likewise (`libLiteRt.dylib` + Metal GPU accelerator from the macOS wheel; install name pinned to an absolute path since SIP strips `DYLD_LIBRARY_PATH`), verifies the subplugin `.dylib` is built, and runs the litert gtest (21/21 observed) and SSAT suites — the first unit testing ever wired into the macOS job.
- Jobs without the SDK (arm meson, Tizen GBS, Android NDK, RISC-V, pdebuild, Yocto) verify the auto-disable path builds cleanly.

Two pre-existing issues found and fixed on the way, since the macOS enablement depends on them:
- `check-rebuild` computed the rebuild flag with `grep | wc -l`; BSD `wc` pads its output, so every step of the macOS job has been silently skipped (~15s fake-success) since the action was introduced.
- Building with `-Dorcc-support=disabled` (needed on macOS, where the orc test code never compiled) fails under default werror due to an unused parameter in `gsttensor_transform.c`.

Verified on Ubuntu 22.04 x86_64 (WSL2) against LiteRT v2.2.0: full build (werror), 21/21 gtest, 6/6 SSAT, `unittest_common`/`unittest_filter_single` regression check, config sanity with the SDK absent/disabled, and both detection paths (`litert.pc` and `LITERT_ROOT`).

## Follow-ups (out of scope here)
- Zero-copy invoke; `invoke_dynamic` for dynamic shapes
- Accelerator opaque options (GPU precision, NPU compiler/dispatch library paths)
- Android ndk-build module wiring + prepare_litert.sh (AAR extraction); Yocto litert recipe in meta-neural-network
- GPU/NPU accelerator runtime packaging guidance per OS (CI exercises the XNNPACK CPU path; macOS ships the Metal accelerator alongside)

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -32,18 +32,29 @@ jobs:
     - name: install LiteRT SDK
       if: env.rebuild == '1'
       run: |
+        # When bumping LITERT_VER, refresh the pinned hashes below:
+        # the zip hash from the release asset, the wheel hashes from
+        # https://pypi.org/pypi/ai-edge-litert/<ver>/json (digests.sha256).
+        # The wheel hash is per python tag: this job pins python 3.11 above,
+        # so the cp311 macosx_12_0_arm64 wheel is the one downloaded.
         LITERT_VER=2.2.0
+        LITERT_SDK_ZIP_SHA256=0aa619d80aef27303ad9c6e3759a20110f77e7b11ade9b68061b8c6e5904b0c6
+        LITERT_WHEEL_SHA256=60831e57ce2887db83783bcff2cd5e33b24e3b37b23ff29d7b68c511e20f6a5b
         LITERT_ROOT=$HOME/litert-sdk
         mkdir -p $LITERT_ROOT/lib $LITERT_ROOT/include/litert/build_common
         curl -sL -o litert_cc_sdk.zip https://github.com/google-ai-edge/LiteRT/releases/download/v${LITERT_VER}/litert_cc_sdk.zip
+        echo "${LITERT_SDK_ZIP_SHA256}  litert_cc_sdk.zip" | shasum -a 256 -c -
         unzip -q litert_cc_sdk.zip -d /tmp/litert-cc-sdk
         cp -r /tmp/litert-cc-sdk/litert_cc_sdk/litert/c $LITERT_ROOT/include/litert/
         # generate build_config.h (all optional features on; runtime dylib decides)
         sed -e 's/#cmakedefine01 \(.*\)/#define \1 0/' \
           /tmp/litert-cc-sdk/litert_cc_sdk/litert/build_common/build_config.h.in \
           > $LITERT_ROOT/include/litert/build_common/build_config.h
-        # take the prebuilt runtime library from the official pip wheel (macosx arm64)
-        pip download ai-edge-litert==${LITERT_VER} --no-deps -q -d /tmp/litert-wheel
+        # take the prebuilt runtime library from the official pip wheel
+        # (macosx arm64), with the expected wheel content pinned by hash
+        printf 'ai-edge-litert==%s --hash=sha256:%s\n' \
+          "$LITERT_VER" "$LITERT_WHEEL_SHA256" > /tmp/litert-req.txt
+        pip download --require-hashes --no-deps -q -r /tmp/litert-req.txt -d /tmp/litert-wheel
         unzip -q -o /tmp/litert-wheel/*.whl 'ai_edge_litert/*.dylib' -d /tmp/litert-wheel/x
         cp /tmp/litert-wheel/x/ai_edge_litert/libLiteRt.dylib $LITERT_ROOT/lib/
         cp /tmp/litert-wheel/x/ai_edge_litert/libLiteRtMetalAccelerator.dylib $LITERT_ROOT/lib/ || true
@@ -51,6 +62,8 @@ jobs:
         # not depend on DYLD_LIBRARY_PATH (which macOS SIP strips from the
         # environment of protected processes such as the CI step shell)
         install_name_tool -id $LITERT_ROOT/lib/libLiteRt.dylib $LITERT_ROOT/lib/libLiteRt.dylib
+        # modifying the install name invalidates the existing code signature;
+        # re-sign ad-hoc so the (hash-verified, see above) dylib stays loadable
         codesign -f -s - $LITERT_ROOT/lib/libLiteRt.dylib
         echo "LITERT_ROOT=$LITERT_ROOT" >> $GITHUB_ENV
     - name: meson build
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -46,7 +46,8 @@ jobs:
         echo "${LITERT_SDK_ZIP_SHA256}  litert_cc_sdk.zip" | shasum -a 256 -c -
         unzip -q litert_cc_sdk.zip -d /tmp/litert-cc-sdk
         cp -r /tmp/litert-cc-sdk/litert_cc_sdk/litert/c $LITERT_ROOT/include/litert/
-        # generate build_config.h (all optional features on; runtime dylib decides)
+        # generate build_config.h (all optional features off at compile time;
+        # the prebuilt runtime dylib ships its own feature set)
         sed -e 's/#cmakedefine01 \(.*\)/#define \1 0/' \
           /tmp/litert-cc-sdk/litert_cc_sdk/litert/build_common/build_config.h.in \
           > $LITERT_ROOT/include/litert/build_common/build_config.h
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -28,12 +28,64 @@ jobs:
           brew install cask
     - name: install minimal requirements
       if: env.rebuild == '1'
-      run: brew install meson ninja pkg-config cmake libffi glib gstreamer numpy json-glib
-    - uses: BSFishy/meson-build@v1.0.3
+      run: brew install meson ninja pkg-config cmake libffi glib gstreamer numpy json-glib googletest
+    - name: install LiteRT SDK
       if: env.rebuild == '1'
-      with:
-        action: build
+      run: |
+        LITERT_VER=2.2.0
+        LITERT_ROOT=$HOME/litert-sdk
+        mkdir -p $LITERT_ROOT/lib $LITERT_ROOT/include/litert/build_common
+        curl -sL -o litert_cc_sdk.zip https://github.com/google-ai-edge/LiteRT/releases/download/v${LITERT_VER}/litert_cc_sdk.zip
+        unzip -q litert_cc_sdk.zip -d /tmp/litert-cc-sdk
+        cp -r /tmp/litert-cc-sdk/litert_cc_sdk/litert/c $LITERT_ROOT/include/litert/
+        # generate build_config.h (all optional features on; runtime dylib decides)
+        sed -e 's/#cmakedefine01 \(.*\)/#define \1 0/' \
+          /tmp/litert-cc-sdk/litert_cc_sdk/litert/build_common/build_config.h.in \
+          > $LITERT_ROOT/include/litert/build_common/build_config.h
+        # take the prebuilt runtime library from the official pip wheel (macosx arm64)
+        pip download ai-edge-litert==${LITERT_VER} --no-deps -q -d /tmp/litert-wheel
+        unzip -q -o /tmp/litert-wheel/*.whl 'ai_edge_litert/*.dylib' -d /tmp/litert-wheel/x
+        cp /tmp/litert-wheel/x/ai_edge_litert/libLiteRt.dylib $LITERT_ROOT/lib/
+        cp /tmp/litert-wheel/x/ai_edge_litert/libLiteRtMetalAccelerator.dylib $LITERT_ROOT/lib/ || true
+        # pin the install name to an absolute path so that linked binaries do
+        # not depend on DYLD_LIBRARY_PATH (which macOS SIP strips from the
+        # environment of protected processes such as the CI step shell)
+        install_name_tool -id $LITERT_ROOT/lib/libLiteRt.dylib $LITERT_ROOT/lib/libLiteRt.dylib
+        codesign -f -s - $LITERT_ROOT/lib/libLiteRt.dylib
+        echo "LITERT_ROOT=$LITERT_ROOT" >> $GITHUB_ENV
+    - name: meson build
+      if: env.rebuild == '1'
+      # orcc-support is disabled: the orc test code in unittest_plugins.cc
+      # does not compile on macOS (orc_int64 is long while int64_t is
+      # long long); it never has, hidden until now by the check-rebuild bug
+      # that skipped every step of this job.
+      run: |
+        meson setup build -Dorcc-support=disabled
+        meson compile -C build
+    - name: check litert subplugin built
+      if: env.rebuild == '1'
+      run: |
+        ls build/ext/nnstreamer/tensor_filter/ | grep litert
+        test -f build/ext/nnstreamer/tensor_filter/libnnstreamer_filter_litert.dylib || test -f build/ext/nnstreamer/tensor_filter/libnnstreamer_filter_litert.so
+    - name: run litert unit tests
+      if: env.rebuild == '1'
+      run: |
+        export DYLD_LIBRARY_PATH=$LITERT_ROOT/lib
+        meson test -C build unittest_filter_litert -v
+    - name: SSAT run on filter-litert
+      if: env.rebuild == '1'
+      run: |
+        git clone --depth 1 https://github.com/myungjoo/SSAT.git $HOME/ssat
+        export PATH=$PATH:$HOME/ssat
+        export DYLD_LIBRARY_PATH=$LITERT_ROOT/lib
+        export NNSTREAMER_BUILD_ROOT_PATH=`pwd`/build
+        export NNSTREAMER_FILTERS=`pwd`/build/ext/nnstreamer/tensor_filter
+        export NNSTREAMER_DECODERS=`pwd`/build/ext/nnstreamer/tensor_decoder
+        export NNSTREAMER_CONVERTERS=`pwd`/build/ext/nnstreamer/tensor_converter
+        export GST_PLUGIN_PATH=`pwd`/build/gst
+        export NNSTREAMER_CONF=`pwd`/build/nnstreamer-test.ini
+        cd tests/nnstreamer_filter_litert
+        bash runTest.sh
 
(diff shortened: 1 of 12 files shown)


myungjoo:

## Review — [Filter/litert] Add LiteRT 2.x (CompiledModel API) subplugin

Overall this is a well-shaped subplugin addition: the `tensorrt`/`tensorrt10` coexistence precedent is followed, the C-API-only dependency keeps the build surface small, `Documentation/component-description.md` and `ext/nnstreamer/tensor_filter/README.md` are updated in the same PR, and the test suite (21 gtest cases + 6 SSAT cases, sharing golden results with the tensorflow2-lite suite) is genuinely stronger than what most existing subplugins ship with. The cross-runtime golden sharing in particular is the right answer to "will we detect it when another module changes underneath us".

Findings below, ordered by severity.

### 1. (Major) `-Dlitert-support=auto` can hard-fail the whole build

In `meson.build`:

```meson
if not fs.is_dir(LITERT_ROOT)
  error('@0@ does not exist'.format(LITERT_ROOT))
endif

litert_lib = cxx.find_library('LiteRt',
  dirs: join_paths(LITERT_ROOT, 'lib'),
  required: true
)
...
if not fs.is_file(join_paths(litert_include_path, 'litert', 'c', 'litert_common.h'))
  error('LiteRT header files are not found in your LITERT_ROOT path.')
endif
```

All three of these are unconditional hard errors, but the option defaults to `auto`. Anyone who has a stale or partial `LITERT_ROOT` exported in their shell — a very likely state for people experimenting with LiteRT — will find that **every unrelated nnstreamer build on that machine now fails outright**, with no way to recover other than unsetting the env var. `auto` must degrade to "feature off", not abort.

Please gate on the option: `required: get_option('litert-support').enabled()` for `find_library`, and replace the two `error()` calls with `message()` plus resetting `litert_dep` to the not-found sentinel unless the feature is explicitly `enabled`. The `cxx.links()` sanity check further down already implements exactly this pattern correctly — the earlier branches should match it.

### 2. (Major) Supply-chain: CI downloads and executes unpinned third-party binaries

Both `.github/workflows/ubuntu_clean_meson_build.yml` and `.github/workflows/macos.yaml` gain steps that fetch an archive and a wheel and then link and run the resulting native library:

```sh
wget -q https://github.com/google-ai-edge/LiteRT/releases/download/v${LITERT_VER}/litert_cc_sdk.zip
pip download ai-edge-litert==${LITERT_VER} --no-deps -q -d /tmp/litert-wheel
```

Neither is integrity-checked. A GitHub release asset can be replaced in place under the same tag, and `pip download` without `--require-hashes` will accept whatever the index serves. This runs on runners that also have the repository checked out. The upstream is reputable and this is not an objection to LiteRT itself, but for a project that ships into Tizen the norm should be hash pinning: record the SHA256 of both artifacts and verify before unpacking (`sha256sum -c`, and `pip download --require-hashes` with a small requirements file).

Same file, macOS job: `codesign -f -s -` ad-hoc-signs a downloaded dylib. That is required to make it loadable, but it deserves an inline comment saying so, since ad-hoc signing an unverified binary reads badly out of context.

### 3. (Major) `invoke()` memcpys without validating the LiteRT buffer size

```c
LITERT_CHECK (LiteRtLockTensorBuffer (input_buffers[i], &host_mem, kLiteRtTensorBufferLockModeWrite));
std::memcpy (host_mem, input[i].data, input[i].size);
...
std::memcpy (output[i].data, host_mem, output[i].size);
```

The length used is always the *pipeline's* notion of the tensor size, never the size of the buffer LiteRT actually handed back. In the normal pipeline path this is safe — the core rejects a caps/property mismatch against `getModelInfo()` before any buffer flows (`gst/nnstreamer/tensor_filter/tensor_filter_common.c:2423` and `:2449`) — so this is defence-in-depth rather than a live bug. But:

- `LiteRtCreateManagedTensorBufferFromRequirements()` allocates from *buffer requirements*, which may legitimately differ from `rank x element_size` (alignment, padding, accelerator-specific packing). Relying on them being equal is an unstated assumption.
- The subplugin's `invoke()` is also reachable directly, which is exactly what several of the new gtest cases do, and there no core-side validation applies.

Please query the buffer size (`LiteRtGetTensorBufferSize`) once at `createTensorBuffers()` time, store it alongside the buffer, and either assert or throw on mismatch in `invoke()`. A silent `memcpy` past the end of an accelerator-allocated buffer is the worst possible failure mode to debug.

Related, smaller: `convertLayout()` never inspects `layout.has_strides`. If LiteRT ever returns a strided (non-contiguous) layout, the flat `memcpy` silently produces garbage. Throwing on `has_strides` would turn that into a clear error instead.

### 4. (Minor) Scope creep — two unrelated fixes ride along

- `gst/nnstreamer/elements/gsttensor_transform.c`: adding `UNUSED (filter)` to the non-orc branch. This is a core element, not a subplugin.
- `.github/actions/check-rebuild/action.yml`: `grep | wc -l` changed to `grep -c`.

Both are legitimate fixes and both are genuinely *needed* by this PR (the macOS job only builds at all once check-rebuild works, and only compiles once orc is disabled). But the check-rebuild fix changes the behavior of **every workflow in the repository** that uses that action. Its practical effect is that the macOS job starts running for the first time in a long while, which is a latent CI-stability risk for all subsequent PRs, not just this one. Please split those two hunks into a separate small PR that can land and be observed on its own; this PR then rebases on top of it. That also makes the LiteRT change reviewable as a pure addition.

### 5. (Minor) One `LiteRtEnvironment` per filter instance

`configure_instance()` calls `LiteRtCreateEnvironment()` per instance, so a pipeline with N `tensor_filter framework=litert` elements creates N environments. For CPU this is merely wasteful; for GPU/NPU accelerators it may mean N device contexts. Worth a `@todo` at minimum, and worth checking against upstream guidance before this gets used with `Accelerators:gpu` in production.

### 6. (Trivial)

- `setTensorMeta()` rejects `num_tensors > NNS_TENSOR_SIZE_LIMIT` but not `num_tensors == 0`; a zero-signature model yields `num_tensors = 0`, which then fails later in `gst_tensors_info_validate()` with a less specific message. This overlaps with #4856.
- When `num_tensors == 0`, `LiteRtGetCompiledModelOutputTensorLayouts()` is handed a null data pointer.
- `parseAcceleratorValue()` splits on `"+|"`, so `"cpu++gpu"` produces an empty entry and reports `Unknown accelerator ""`, a confusing message for a typo.
- `kLiteRtElementTypeBool` silently maps to `_NNS_UINT8`. Correct in practice, but the mapping deserves a comment since it is the only lossy entry in the table.

### Things I checked and found fine

- `applyHwList()`'s bitmask tests against `ACCL_CPU`/`ACCL_GPU`/`ACCL_NPU` work correctly with the `0xn000`-style `accl_hw` encoding (device index in the low bits), including `ACCL_NPU_SR = 0x4100`.
- `cleanup()` releases tensor buffers, compiled model, model, environment and `GstTensorsInfo` in the right order, and `configure_instance()`'s `catch (...)` covers the `LiteRtOptions` leak correctly.
- `parseCustomProperties()`'s try/catch around the `g_strsplit` result is leak-free on the throw path.
- Test coverage for future cross-module regressions is the strong point here: sharing model files and golden argmax with the tensorflow2-lite suite means a change to the shared test models or to the core caps negotiation fails *both* suites, which is exactly what you want.
- No back-door-shaped code found in the subplugin. The only trust extension is the CI artifact download covered in item 2.

myungjoo:

## Re-review of the 6-commit push

The branch is now split into six commits (subplugin / SSAT + Ubuntu CI / macOS CI / check-rebuild fix / transform fix / docs). Diffing the new head against the previous single-commit version, **the file contents are byte-identical** — this is a commit restructuring plus a rebase onto current `main`, not a content change.

That is a real improvement for reviewability, and it makes the two out-of-scope hunks visible as their own units (`d39d0f6` check-rebuild, `b784b86` transform). Partially addresses the earlier scope comment.

### What it does not address

The earlier point was that `d39d0f6` changes the behavior of **every workflow in the repository** that uses `.github/actions/check-rebuild`, and its practical effect is that the macOS job starts running for the first time in a long while. Splitting it into its own commit does not change that it lands atomically with LiteRT: if the reactivated macOS job turns out to be unstable, reverting it means reverting the whole PR.

Those two commits are self-contained and have no dependency on the LiteRT work — they only need to land *before* it. Landing them as a separate small PR first lets the macOS job be observed on its own for a day, and this PR then rebases on top. That is still the safer sequencing.

### Still open from the previous round

None of the three Major findings were touched (expected, since the content is unchanged) — restating them so they are not lost in the commit-split noise:

1. **`-Dlitert-support=auto` can hard-fail unrelated builds.** `meson.build` still has two unconditional `error()` calls and `cxx.find_library(..., required: true)` inside a branch that only runs because `LITERT_ROOT` happens to be exported. A stale env var breaks every nnstreamer build on that machine. The `cxx.links()` sanity check below it already does this correctly — the earlier branches should match it.
2. **CI downloads unpinned third-party binaries.** `litert_cc_sdk.zip` (GitHub release asset, mutable under its tag) and `pip download ai-edge-litert==2.2.0` (no `--require-hashes`) are fetched, linked, and executed on runners that hold the checkout. SHA256 pinning for both.
3. **`invoke()` memcpys using the pipeline's tensor size, never the LiteRT buffer's.** Safe in the normal pipeline path because the core validates against `getModelInfo()` first, but not when `invoke()` is called directly — which the new gtest cases do. Query `LiteRtGetTensorBufferSize` once in `createTensorBuffers()` and validate.

Plus the smaller ones: `layout.has_strides` unchecked, `num_tensors == 0` unhandled in `setTensorMeta()`, `parseAcceleratorValue()`'s empty-entry message, and the per-instance `LiteRtEnvironment`.

### CI note

`Android NDK build on Ubuntu for arm64-v8a` is red on this PR, but it is not caused by these changes: the job fails in `Prepare Build` at `Download prebuilt Android-GStreamer-1.24.13 libraries` with exit code 8 (server error on the download), before any nnstreamer source is compiled. The same job failed identically on #4854 and #4829 in the same time window, neither of which touches `jni/`. Infrastructure flakiness — worth a re-run rather than investigation.

myungjoo:
Applied the review's M1, M3, and minor findings in e3f67c8 (M2 hash-pinning and the check-rebuild/transform split are deferred to a follow-up decision):

- **M1**: with `-Dlitert-support=auto`, every `LITERT_ROOT` detection failure (missing dir, missing headers, missing lib) now degrades to feature-off with a message instead of aborting; explicit `enabled` still fails loudly. Verified both modes with a stale `LITERT_ROOT=/nonexistent-path`.
- **M3**: `createTensorBuffers()` queries `LiteRtGetTensorBufferSize` per buffer, rejects buffers smaller than the tensor, and `invoke()` validates every memcpy length against the actual LiteRT buffer size — a direct `invoke()` call can no longer read/write past an accelerator-allocated buffer.
- **Minors**: `has_strides` layouts rejected explicitly; zero-input/output signatures rejected with a specific message (and no null pointer handed to `LiteRtGetCompiledModelOutputTensorLayouts`); empty accelerator entries (`cpu++gpu`) tolerated; Bool→UINT8 mapping commented; per-instance `LiteRtEnvironment` recorded as a `@todo`.

Re-verified after the changes: 21/21 gtest, 6/6 SSAT against LiteRT 2.2.0 (Ubuntu 22.04 x86_64).

myungjoo:
Per the review's sequencing request, the two prerequisite commits (check-rebuild fix, transform orc-off fix) are now also submitted as an independent PR: #4858. Once it lands, this branch will rebase onto main — the identical patches will drop out automatically, leaving this PR a pure LiteRT addition. Until then this branch keeps them so its macOS CI coverage stays meaningful.

myungjoo:
M2 (supply-chain) is now addressed in e91eacb: both CI jobs verify `litert_cc_sdk.zip` against a pinned SHA256 right after download, and the `ai-edge-litert` wheel is fetched via `pip download --require-hashes` with the per-platform cp311 wheel hashes taken from the official PyPI metadata. The macOS ad-hoc `codesign` now carries the requested explanation comment. The bump procedure (where each hash comes from, and the python-tag coupling) is documented next to the pins.

Verified locally: the pinned zip hash matches an independent fresh download, and a deliberately wrong wheel hash is rejected by pip with its tamper warning.

With this, all review findings are addressed except the sequencing item, which is in flight: the two prerequisite commits are up as #4858 (17/18 green incl. the first real macOS build; the Android job is red on the ongoing gstreamer.freedesktop.org outage, being retried). Once #4858 lands, this branch rebases onto main and becomes a pure LiteRT addition.

myungjoo:
Thanks for the review — all items addressed or answered below. The branch is now rebased onto current `main` (post-#4858), which drops the two duplicated commits (`d39d0f6`, `b784b86`) as promised earlier; the PR is now a pure LiteRT addition. Three new commits on top:

**1. litert CI coverage can silently disappear** → fixed in d775c76. The Ubuntu job now asserts `libnnstreamer_filter_litert.so` exists right after `meson compile` (ubuntu-22.04 only, where the SDK is installed), mirroring the macOS job's `check litert subplugin built` step. I kept `auto` + assertion rather than `-Dlitert-support=enabled` because the meson setup line is shared with the ubuntu-22.04-arm matrix entry, which installs no SDK. The zero-case `report` in runTest.sh is left as is — that is the established SSAT convention for optional subplugins so local full-suite runs don't fail, and the CI guarantee now comes from the assertion.

**2. `G_MINFLOAT` max-search seed** → fixed in 015f7f3 (`-G_MAXFLOAT`, with a comment). Noted that this file had copied the pattern from the existing filter tests that #4864/#4866 are fixing; this PR no longer reintroduces it.

**3. `>` vs `!=` on the invoke size check** → fixed in 7e9bd81. Good catch — this was the sharpest finding of the round: a smaller-than-model tensor previously passed and left stale data in the reused buffer's tail. `invoke()` now rejects any size mismatch against the *nnstreamer* tensor size (cached at `createTensorBuffers()` time, which still guarantees every LiteRT buffer holds the whole tensor, so equality also rules out overflow in both directions).

**4. Overlap with #4858** → resolved by the rebase; both commits auto-dropped (`git rebase` patch-id detection, no conflicts).

**5. Minor**
- `cleanup()` field reset → fixed in 7e9bd81. Chose reset over dropping the `if (configured) cleanup()` branch: the branch is cheap defence and becomes reachable the moment a RELOAD event handler is added.
- `NNSTREAMER_FILTER_LITERT_SRCS` unconsumed: this matches every other `NNSTREAMER_FILTER_*_SRCS` entry in `jni/nnstreamer.mk` (QNN, SNAP, MXNET, LLAMA2C, ...) — these are exported for downstream ndk-build consumers (nnstreamer-android / ml-api), not consumed in-repo. The litert entry follows that convention, so I'd rather keep it for parity than special-case it.
- Packaging untouched: intentional for this stage, per the PR description — `packaging/nnstreamer.spec` / `debian/` entries will come in the follow-up that introduces `litert.pc` on a shipping target.

Bonus: while verifying item 1, the `build_config.h` generation comment in both workflow files said "features on" while the `sed` defines them to 0 — corrected in d775c76.

These fixes were not re-run against the SDK locally (prepared on a machine without the LiteRT SDK); the Ubuntu/macOS CI on this push is the verification, and the new assertion from item 1 ensures that verification actually exercises the subplugin.

> _Note: this response and the fixes were prepared by an (AI) agent operated by the PR author._

myungjoo:
The non-blocking items deferred from the review rounds (packaging, Android consumer, environment sharing, zero-copy, invoke_dynamic, accelerator options, runTest.sh standalone cosmetics) are now tracked in #4869.

PR #4856, opened 2026-08-21
[Common] check max tensors

jaeyun-jung:
Check max number of tensors and clearly validate it in tensor information.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc
@@ -756,7 +756,15 @@ int
 TFLiteInterpreter::setTensorProp (
     const std::vector<int> &tensor_idx_list, GstTensorsInfo *tensorMeta)
 {
-  tensorMeta->num_tensors = tensor_idx_list.size ();
+  unsigned int num = tensor_idx_list.size ();
+
+  if (num < 1 || num > NNS_TENSOR_SIZE_LIMIT) {
+    ml_loge ("The number of tensors is %u and it should be between 1 and %d.",
+        num, NNS_TENSOR_SIZE_LIMIT);
+    return -1;
+  }
+
+  tensorMeta->num_tensors = num;
 
   for (unsigned int i = 0; i < tensorMeta->num_tensors; ++i) {
     int idx = tensor_idx_list[i];
--- a/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_util_impl.c
@@ -454,13 +454,13 @@ gst_tensors_info_validate (const GstTensorsInfo * info)
     return TRUE;
   }
 
-  if (info->num_tensors < 1) {
+  if (info->num_tensors < 1 || info->num_tensors > NNS_TENSOR_SIZE_LIMIT) {
     nns_logd
-        ("Failed to validate tensors info. the number of tensors: %d. the number of tensors should be greater than 0.",
-        info->num_tensors);
+        ("Failed to validate tensors info. The number of tensors is %u and it should be between 1 and %d.",
+        info->num_tensors, NNS_TENSOR_SIZE_LIMIT);
     _nnstreamer_error_write
-        ("Failed to validate tensors info. the number of tensors: %d. the number of tensors should be greater than 0.",
-        info->num_tensors);
+        ("Failed to validate tensors info. The number of tensors is %u and it should be between 1 and %d.",
+        info->num_tensors, NNS_TENSOR_SIZE_LIMIT);
     return FALSE;
   }
 
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -776,6 +776,21 @@ TEST (commonTensorsInfo, validateInvalidFormat_n)
 }
 
 /**
+ * @brief Test for validating of the tensors info with invalid number of tensors.
+ */
+TEST (commonTensorsInfo, validateInvalidNum_n)
+{
+  GstTensorsInfo info;
+
+  gst_tensors_info_init (&info);
+
+  info.num_tensors = 0;
+  EXPECT_FALSE (gst_tensors_info_validate (&info));
+  info.num_tensors = NNS_TENSOR_SIZE_LIMIT + 1;
+  EXPECT_FALSE (gst_tensors_info_validate (&info));
+}
+
+/**
  * @brief Test for getting nth info with invalid index.
  */
 TEST (commonTensorsInfo, getNthInfoInvalidIndex_n)


myungjoo:

## Review — [Common] check max tensors

The change does what the title says and the upper bound is the right one: `NNS_TENSOR_SIZE_LIMIT` is 256 (`gst/nnstreamer/include/tensor_typedef.h:42`), i.e. the true maximum including the `extra` list, not the 16-entry static array (`NNS_TENSOR_MEMORY_MAX`). So this does **not** regress the extended-tensor path. Scope is minimal (2 files, no unrelated modules touched). Good.

### 1. (Major) No test covers the new boundary

This is the main gap. `tests/common/unittest_common.cc` already has the natural home for it — `commonTensorsInfo.validateInvalidParam_n` / `validateInvalidFormat_n` (around line 760) — and there is currently **no** case that exercises `num_tensors` at all in `gst_tensors_info_validate()`. Without one:

- the newly added `> NNS_TENSOR_SIZE_LIMIT` branch is never executed in CI, so nothing proves it works today;
- more importantly, if someone later raises/lowers `NNS_TENSOR_SIZE_LIMIT` or reworks `gst_tensors_info_get_nth_info()`'s `extra` allocation, **no CI job fails** — the regression ships silently. That is exactly the case this PR is trying to defend against.

Three cheap cases would close it:

```c
info.num_tensors = 0;                        EXPECT_FALSE (gst_tensors_info_validate (&info));
info.num_tensors = NNS_TENSOR_SIZE_LIMIT;    /* boundary, must stay valid when infos are filled */
info.num_tensors = NNS_TENSOR_SIZE_LIMIT + 1; EXPECT_FALSE (gst_tensors_info_validate (&info));
```

Likewise for the tflite side: `TFLiteInterpreter::setTensorProp()` now returns `-1` on a bad tensor count, and the error propagates correctly through `setInputTensorProp()`/`setOutputTensorProp()` up to `TFLiteCore` (`tensor_filter_tensorflow_lite.cc:1114`, `:1707`), but nothing verifies that propagation.

### 2. (Trivial) Format specifier

`num` is `unsigned int` but printed with `%d` in both the new `ml_loge()` and the reworded `nns_logd`/`_nnstreamer_error_write` messages. `%u` is correct; with `%d` a (hypothetical) count above `INT_MAX` prints as a negative number, which is misleading in exactly the log line meant to explain the failure.

### Other checks

- **Regression to other modules:** `gst_tensors_info_validate()` is called broadly (tensor_converter, tensor_filter caps negotiation, the Tizen filter template at `tests/nnstreamer_filter_extensions_common/unittest_tizen_template.cc.in:453`). The added condition only rejects states that were already unrepresentable, so no legitimate caller should start failing. Confirmed that `commonTensorsInfo.printInvalidIndex_n` (which deliberately sets `num_tensors = NNS_TENSOR_SIZE_LIMIT + 1`) exercises `gst_tensors_info_to_string()`, not `validate()`, so it is unaffected.
- **API/ARCHITECTURE docs:** no public API or behavior contract changes; no doc update needed.
- **Security:** no issue; this tightens an input-validation path. No back-door-shaped code found.

myungjoo:

## Re-review of c9b4114

Thanks for the quick turnaround. The `%d` → `%u` fix is done in all three places. On the test, though — I checked whether the new case actually distinguishes a patched tree from an unpatched one, and it does not.

### `validateInvalidNum_n` passes without the patch

```c
info.num_tensors = 0;
EXPECT_FALSE (gst_tensors_info_validate (&info));
info.num_tensors = NNS_TENSOR_SIZE_LIMIT + 1;
EXPECT_FALSE (gst_tensors_info_validate (&info));
```

- The `num_tensors = 0` half was already rejected by the pre-existing `info->num_tensors < 1` branch, so it asserts old behavior.
- The `NNS_TENSOR_SIZE_LIMIT + 1` half also returns `FALSE` on unpatched `main`, but for an unrelated reason: after `gst_tensors_info_init()`, the validation loop reaches `i = 0`, gets `&info->info[0]` whose `type` is `_NNS_END`, and `gst_tensor_info_validate()` returns `FALSE` right there. The loop never gets anywhere near index 256.

Filling the first 256 infos with valid types/dimensions would not help either: the old loop would then reach `i = 256`, `gst_tensors_info_get_nth_info()` returns `NULL` (index is not `< NNS_TENSOR_SIZE_LIMIT`), and `gst_tensor_info_validate(NULL)` returns `FALSE` via its `g_return_val_if_fail`.

So there is no input for which the old and new `gst_tensors_info_validate()` differ in return value. The new upper-bound branch is not observable through the function's return value at all, which means no test of this function can gate it.

### What that implies for the PR

This is not an argument against the change — it is still worth making, but for reasons the commit message does not currently state:

1. **It replaces a confusing failure with a clear one.** Today an out-of-range count surfaces as `Failed to get the information, invalid index 256 (max 256)` from deep inside `gst_tensors_info_get_nth_info()`. The new message names the actual problem.
2. **It removes a real crash risk under `G_DISABLE_CHECKS`.** That is the one place where behavior genuinely changes: with checks compiled out, `g_return_val_if_fail (info != NULL, FALSE)` disappears and the old path dereferences the `NULL` returned by `get_nth_info()`. The new early return prevents that.

Worth putting (2) in the commit message — it is the strongest justification and it is currently invisible.

### The untested half is the tflite side

`TFLiteInterpreter::setTensorProp()` is where this PR makes a genuinely new, observable behavior change: it now returns `-1` where it previously assigned the count and carried on. That error propagates through `setInputTensorProp()`/`setOutputTensorProp()` into `TFLiteCore` (`tensor_filter_tensorflow_lite.cc:1114`, `:1707`, and the recovery path at `:1660`), and none of it is covered.

That is the path a test can actually gate, and it is also the path with a plausible regression: someone refactoring the `tryRecovery` chain could swallow the `-1` and the model would silently load with a bogus tensor count. A case in `tests/nnstreamer_filter_tensorflow2_lite/` that opens a model and asserts the failure propagates would be the useful addition here — more so than another `gst_tensors_info_validate()` case.

### Unrelated: the Yocto CI failure on this PR is not yours

`Build with Yocto / meta-neural-network on Ubuntu` failed on run 32461246477 with `fatal: unable to connect to git.yoctoproject.org`. That is the connectivity flakiness #4857 fixes with a GitHub-mirror fallback; #4857 has since merged, so a rebase onto current `main` should clear it.

myungjoo (approved):
Approved.

> _Note: the review text below was produced by another (AI) agent and is transcribed here._

c9b4114 addresses the format-specifier point (`%d` → `%u` in all three places) and adds `commonTensorsInfo.validateInvalidNum_n`.

On the earlier "the test does not gate" observation — after looking more closely, that is not something this PR can fix, so it should not hold it up. There is no input for which the old and new `gst_tensors_info_validate()` differ in return value: an out-of-range `num_tensors` already ended in `FALSE`, just by way of `gst_tensors_info_get_nth_info()` returning `NULL` and `gst_tensor_info_validate()`'s `g_return_val_if_fail` catching it. The new branch is not observable through the return value, so no test of that function can distinguish patched from unpatched. The added case is still worth having as a regression guard on the existing behavior.

The change itself is safe: it only rejects tensor counts that were already unusable, and it rejects them earlier and with a message that names the actual problem instead of `invalid index 256 (max 256)` from deep inside the accessor. It also removes a genuine crash path under `G_DISABLE_CHECKS`, where `g_return_val_if_fail` compiles out and the old code would dereference the `NULL`. That last point is the strongest justification for the change and is worth a line in the commit message.

One non-blocking follow-up: `TFLiteInterpreter::setTensorProp()` is where this PR makes a genuinely new, observable behavior change (returning `-1` where it previously carried on), and that path is untested. It propagates through `setInputTensorProp()` / `setOutputTensorProp()` into `TFLiteCore` including the `tryRecovery` chain at `tensor_filter_tensorflow_lite.cc:1660`, which is exactly the kind of place a future refactor could swallow the error and load a model with a bogus tensor count. A case under `tests/nnstreamer_filter_tensorflow2_lite/` asserting the failure propagates would be the useful addition — that path *can* be gated, unlike the one above.

Note the red Yocto check on this PR is unrelated: run 32461246477 failed with `fatal: unable to connect to git.yoctoproject.org`. #4857 has since merged with a GitHub-mirror fallback, so a rebase onto current `main` should clear it.

PR #4857, opened 2026-08-21
[CI/Yocto] Fix and harden the Yocto job

myungjoo:
The Yocto CI job has been failing for reasons unrelated to the code under test. Three fixes:

1. **Upgrade poky yocto-5.0.3 → yocto-5.0.19 (both scarthgap).** The GitHub runner image's `tar` recently gained a CVE fix that uses the `openat2` syscall, which the old pseudo does not intercept; `do_package` then dies with `tar: ... Cannot mkdir: Bad address` (observed at `linux-libc-headers` after ~1h of build). The pseudo fix is backported in scarthgap point releases — see the [Yocto 5.0.15 release notes](https://docs.yoctoproject.org/migration-guides/release-notes-5.0.15.html). Cache keys follow the new tag.

2. **Fall back to the GitHub poky mirror on clone failure.** `git.yoctoproject.org` intermittently refuses connections from GitHub-hosted runners, failing the job at the very first step. The official repository stays primary; the mirror is used only when the clone fails (verified working on a PR run during the current outage). A partially created directory from a mid-transfer failure is removed before retrying.

3. **Run the acquire-disk-space step on pull requests too**, exiting early when there is already enough free space (observed runners vary between ~20G and ~87G free). A cold sstate cache — e.g., when the nightly cache refresh itself failed on the outage above — makes PR runs build from scratch, so they need the same headroom the manual dispatch path already provides. The warm case only costs a `df` invocation.

These were found while stabilizing CI for #4855; split into an independent PR since they are unrelated to that feature.

Note: the first run of this PR builds from a cold cache (the old cache key no longer matches) and will take longer than usual; the nightly run on main will repopulate the cache under the new key.

### DIFF
--- a/.github/workflows/yocto.yml
+++ b/.github/workflows/yocto.yml
@@ -1,4 +1,4 @@
-name: Build test / Yocto-5.0.3 (scarthgap-5.0.3)
+name: Build test / Yocto-5.0.19 (scarthgap-5.0.19)
 
 on:
   pull_request:
@@ -64,12 +64,20 @@ jobs:
           /var/cache/yocto/downloads
           /var/cache/yocto/sstate-cache
           /var/cache/yocto/persistent
-        key: yocto-cache-yocto-5.0.3
+        key: yocto-cache-yocto-5.0.19
 
     - name: acquire some disk space
-      if: inputs.acquire-space == 'enable'
+      # Also run on pull requests: when the sstate cache is cold (e.g., the
+      # nightly cache refresh failed), a from-scratch build exhausts the
+      # runner disk. The step exits early when there is already enough room.
+      if: inputs.acquire-space == 'enable' || github.event_name == 'pull_request'
       run: |
           df -h
+          avail_gb=$(df --output=avail -BG / | tail -1 | tr -dc '0-9')
+          if [ "${{ inputs.acquire-space }}" != "enable" ] && [ "$avail_gb" -gt 40 ]; then
+            echo "Enough free disk space (${avail_gb}G); skipping cleanup."
+            exit 0
+          fi
           sudo apt-get update --fix-missing \
             && sudo apt-get remove -y '^dotnet-.*' '^llvm-.*' '^mysql-server-core-.*' '^postgresql-.*' '^temurin-.*' \
             azure-cli google-chrome-stable google-cloud-cli firefox powershell microsoft-edge-stable mono-devel \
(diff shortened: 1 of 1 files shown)


myungjoo:

## Review — CI/Yocto hardening

The topic (fix + harden the Yocto job) is addressed and the diff stays inside `.github/workflows/yocto.yml`, so scope is fine. Three findings.

### 1. (Major) The "acquire some disk space" step will now run on *every* pull request, including the ones that skip the build

```yaml
if: inputs.acquire-space == 'enable' || github.event_name == 'pull_request'
```

Every other expensive step in this job is gated on `steps.rebuild.outputs.rebuild == '1'` (cache restore, the build itself, cache save). This one is not, so a docs-only PR that skips the whole Yocto build will still pay for the `apt-get remove '^dotnet-.*' '^llvm-.*' …` + docker prune pass.

The intended mitigation is the early exit, but the threshold does not actually mitigate anything:

```sh
if [ "${{ inputs.acquire-space }}" != "enable" ] && [ "$avail_gb" -gt 40 ]; then
```

GitHub-hosted `ubuntu-22.04` runners expose roughly 25–30 GB free on `/`, i.e. `avail_gb` is essentially always below 40, so the "skip when there is enough room" branch is effectively dead code and the cleanup always runs. Please either add the `steps.rebuild.outputs.rebuild == '1'` guard, or lower the threshold to a value that a real runner can actually exceed — otherwise the two mechanisms cancel each other out.

### 2. (Minor) `avail_gb` is used unquoted-as-integer without a fallback

```sh
avail_gb=$(df --output=avail -BG / | tail -1 | tr -dc '0-9')
```

If `df` ever produces something unexpected (or the pipeline is run on a runner image where `--output` is unavailable), `avail_gb` becomes the empty string and `[ "" -gt 40 ]` aborts with `integer expression expected`, failing the step and therefore the job. A `: "${avail_gb:=0}"` (or `|| echo 0`) makes the harden-the-job intent actually robust.

### 3. (Minor) Prefer the HTTPS clone as the primary, not the fallback

```sh
git clone git://git.yoctoproject.org/poky -b yocto-5.0.19 \
  || { rm -rf poky; git clone https://github.com/yoctoproject/poky -b yocto-5.0.19; }
```

The fallback is correct and the mirror is the official one, but the ordering means the *unauthenticated, unencrypted* `git://` transport is tried first on every run and the integrity-checked HTTPS path is only the fallback. Since the git protocol is deprecated upstream and the stated failure mode is exactly "`git.yoctoproject.org` intermittently drops connections", swapping the order gives both better reliability and a better security posture at zero cost.

### Tests / regression detection

Nothing to add here — this is CI plumbing and the workflow itself is the test. Worth noting that the *first* run after this lands will have a cold sstate cache (the key changed to `yocto-cache-yocto-5.0.19`), which is precisely the scenario the disk-space change is meant to survive; please watch that first scheduled run.

No API/ARCHITECTURE change, so no documentation update is required. No security issue and no back-door-shaped code found.

jaeyun-jung (approved):

PR #4858, opened 2026-08-24
[CI] Fix check-rebuild on macOS and the orc-disabled werror build

myungjoo:
Self-contained fixes split out of #4855 per its review, so that reactivating the macOS CI job can land and be observed independently of the LiteRT feature (and reverted independently if the job turns out to be unstable):

1. **`.github/actions/check-rebuild`**: the rebuild flag was computed with `grep | wc -l`; BSD `wc` on macOS pads its output with leading spaces, so the padded value never matched the `env.rebuild == '1'` step conditions. **Every step of the macOS job has been silently skipped since the action was introduced** — the job "passed" in ~15 seconds without building anything. `grep -c` prints an unpadded count on all platforms (the check script emits exactly one `REBUILD=YES|NO` line, so the count is always 0 or 1).

2. **`gst/nnstreamer/elements/gsttensor_transform.c`**: in `gst_tensor_transform_typecast()`, the `filter` parameter is used only inside the `HAVE_ORC` block, so building with `-Dorcc-support=disabled` fails under the default werror (`-Werror=unused-parameter`). Mark it `UNUSED` in the non-orc path. This is the configuration that **#4855 will use for the macOS job** (the orc test code in `unittest_plugins.cc` does not compile on macOS — tracked in #4862). Note that the current macOS job builds *with* orc (brew's gstreamer pulls it in), so this fix is exercised by the new non-orc gate below, not by the macOS job on this PR.

3. **Review follow-ups (see the review comments below)**:
   - check-rebuild no longer fails open: an error from `check_if_rebuild_requires.sh` turns the step red instead of silently producing `rebuild=0` and skipping the job, and the script rejects a missing/unreadable file-list argument.
   - The Ubuntu LLVM workflow gains a `-Dorcc-support=disabled` configure-and-compile gate (with clang, like its sibling steps) so the non-orc build path is covered by CI from now on.
   - `test_check_if_rebuild_requires.sh` exercises the checker's YES/NO decision for every mode plus the exact `sh -e` step pattern of `action.yml` (including the `REBUILD=NO`/`|| true` path and the fail-closed error path), and runs in the **Static checks** workflow — which is not gated on the rebuild flag, so these paths are verified by CI on every PR, including this one.

Effect to watch after merge: the macOS job starts doing real builds for the first time, and docs-only PRs must still skip cleanly across all 8 workflows using check-rebuild. The LiteRT-specific macOS steps arrive with #4855, which will rebase on top of this.

### DIFF
--- a/.github/actions/check-rebuild/action.yml
+++ b/.github/actions/check-rebuild/action.yml
@@ -14,16 +14,11 @@ runs:
         tmpfile=$(mktemp)
         git show --pretty="format:" --name-only --diff-filter=AMRC ${{ github.event.pull_request.head.sha}} -${{ github.event.pull_request.commits }} | sort | uniq | awk NF > ${tmpfile}
         echo "changed_file_list=${tmpfile}" >> "$GITHUB_ENV"
-        # Use grep -c instead of "grep | wc -l": the BSD wc on macOS pads its
-        # output with leading spaces, which made `env.rebuild == '1'` always
-        # false there (macOS jobs silently skipped every build step).
-        # Check the script status separately: a script failure must fail this
-        # step instead of silently producing rebuild=0 and skipping the job.
-        out=$(bash .github/actions/check-rebuild/check_if_rebuild_requires.sh ${tmpfile} ${{ inputs.mode }}) || {
-          echo "::error::check_if_rebuild_requires.sh failed"
-          exit 1
-        }
-        rebuild=$(printf '%s\n' "$out" | grep -c "REBUILD=YES" || true)
+        # get_rebuild_flag.sh prints 0 or 1 and exits nonzero on checker
+        # failure, which fails this "sh -e" step instead of silently
+        # skipping the job. It is also exercised by
+        # test_check_if_rebuild_requires.sh in the Static checks workflow.
+        rebuild=$(bash .github/actions/check-rebuild/get_rebuild_flag.sh ${tmpfile} ${{ inputs.mode }})
         echo "Rebuild required: ${rebuild}"
         echo "rebuild=${rebuild}" >> "$GITHUB_ENV"
       shell: sh
--- a/.github/actions/check-rebuild/action.yml
+++ b/.github/actions/check-rebuild/action.yml
@@ -17,7 +17,13 @@ runs:
         # Use grep -c instead of "grep | wc -l": the BSD wc on macOS pads its
         # output with leading spaces, which made `env.rebuild == '1'` always
         # false there (macOS jobs silently skipped every build step).
-        rebuild=`bash .github/actions/check-rebuild/check_if_rebuild_requires.sh ${tmpfile} ${{ inputs.mode }} | grep -c "REBUILD=YES" || true`
+        # Check the script status separately: a script failure must fail this
+        # step instead of silently producing rebuild=0 and skipping the job.
+        out=$(bash .github/actions/check-rebuild/check_if_rebuild_requires.sh ${tmpfile} ${{ inputs.mode }}) || {
+          echo "::error::check_if_rebuild_requires.sh failed"
+          exit 1
+        }
+        rebuild=$(printf '%s\n' "$out" | grep -c "REBUILD=YES" || true)
         echo "Rebuild required: ${rebuild}"
         echo "rebuild=${rebuild}" >> "$GITHUB_ENV"
       shell: sh
--- a/.github/actions/check-rebuild/action.yml
+++ b/.github/actions/check-rebuild/action.yml
@@ -14,7 +14,10 @@ runs:
         tmpfile=$(mktemp)
         git show --pretty="format:" --name-only --diff-filter=AMRC ${{ github.event.pull_request.head.sha}} -${{ github.event.pull_request.commits }} | sort | uniq | awk NF > ${tmpfile}
         echo "changed_file_list=${tmpfile}" >> "$GITHUB_ENV"
-        rebuild=`bash .github/actions/check-rebuild/check_if_rebuild_requires.sh ${tmpfile} ${{ inputs.mode }} | grep "REBUILD=YES" | wc -l`
+        # Use grep -c instead of "grep | wc -l": the BSD wc on macOS pads its
+        # output with leading spaces, which made `env.rebuild == '1'` always
+        # false there (macOS jobs silently skipped every build step).
+        rebuild=`bash .github/actions/check-rebuild/check_if_rebuild_requires.sh ${tmpfile} ${{ inputs.mode }} | grep -c "REBUILD=YES" || true`
         echo "Rebuild required: ${rebuild}"
         echo "rebuild=${rebuild}" >> "$GITHUB_ENV"
       shell: sh
--- a/.github/actions/check-rebuild/check_if_rebuild_requires.sh
+++ b/.github/actions/check-rebuild/check_if_rebuild_requires.sh
@@ -15,8 +15,8 @@
 #                  android: check if jni rebuild is required
 #                  build (default): check if general meson rebuild is required.
 
-if [ -z $1 ]; then
-  echo "::error The argument (file path) is not given."
+if [ ! -f "$1" ]; then
+  echo "::error The argument (file path) is not given or is not a file."
   exit 1
 fi
 
--- a/.github/actions/check-rebuild/get_rebuild_flag.sh
+++ b/.github/actions/check-rebuild/get_rebuild_flag.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file: get_rebuild_flag.sh
+# @brief    Print 1 if the given changed-file list requires a rebuild, else 0.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Argument 1 ($1): the file containing the list of changed files.
+# Argument 2 ($2): build mode to be checked (see check_if_rebuild_requires.sh).
+#
+# Exits nonzero if the checker fails, so that the calling CI step (run under
+# "sh -e") turns red instead of silently skipping the job.
+# "grep -c" is used instead of "grep | wc -l": the BSD wc on macOS pads its
+# output with leading spaces, which made the `env.rebuild == '1'` step
+# conditions always false there (macOS jobs silently skipped every build step).
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+
+out=$(bash "${SCRIPT_DIR}/check_if_rebuild_requires.sh" "$1" "$2") || {
+  echo "::error::check_if_rebuild_requires.sh failed" >&2
+  exit 1
+}
+printf '%s\n' "$out" | grep -c "REBUILD=YES" || true
--- a/.github/actions/check-rebuild/test_check_if_rebuild_requires.sh
+++ b/.github/actions/check-rebuild/test_check_if_rebuild_requires.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file: test_check_if_rebuild_requires.sh
+# @brief    Self-test for check_if_rebuild_requires.sh and the action.yml step pattern.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Runs the same "out=$(...) || exit; grep -c ... || true" pattern as
+# .github/actions/check-rebuild/action.yml under "sh -e", which is how GitHub
+# executes composite "shell: sh" steps. This keeps the REBUILD=NO path and the
+# fail-closed error path exercised by CI on every PR.
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+CHECKER="${SCRIPT_DIR}/check_if_rebuild_requires.sh"
+failed=0
+
+# run_step <file-list-path> <mode>: emulate the action.yml step under sh -e.
+run_step() {
+  sh -ec '
+    out=$(bash "$1" "$2" "$3") || {
+      echo "::error::check_if_rebuild_requires.sh failed" >&2
+      exit 1
+    }
+    printf "%s\n" "$out" | grep -c "REBUILD=YES" || true
+  ' sh "$CHECKER" "$1" "$2"
+}
+
+# expect_rebuild <expected 0|1> <mode> <changed-file>...
+expect_rebuild() {
+  expected=$1
+  mode=$2
+  shift 2
+  list=$(mktemp)
+  printf '%s\n' "$@" > "$list"
+  if actual=$(run_step "$list" "$mode") && [ "$actual" = "$expected" ]; then
+    echo "PASS: mode=${mode} files=[$*] rebuild=${actual}"
+  else
+    echo "FAIL: mode=${mode} files=[$*] expected=${expected} got=${actual:-<step failed>}"
+    failed=1
+  fi
+  rm -f "$list"
+}
+
+# A source change triggers a rebuild in every mode.
+for mode in build gbs debian android; do
+  expect_rebuild 1 "$mode" "gst/nnstreamer/nnstreamer_plugin_api_impl.c"
+done
+
+# Docs-only changes must skip every mode (this is the REBUILD=NO / "|| true"
+# path that lets docs-only PRs pass all 8 consuming workflows).
+for mode in build gbs debian android; do
+  expect_rebuild 0 "$mode" "README.md" "docs/logo.png"
+done
+
+# Mode-specific arms.
+expect_rebuild 1 gbs "packaging/nnstreamer.spec"
+expect_rebuild 0 build "packaging/nnstreamer.spec"
+expect_rebuild 1 debian "debian/rules"
+expect_rebuild 0 android "debian/rules"
+expect_rebuild 1 android "jni/Android.mk"
+expect_rebuild 0 gbs "jni/Android.mk"
+
+# A checker failure must fail the step instead of silently yielding rebuild=0.
+if run_step "/nonexistent/file-list" build > /dev/null 2>&1; then
+  echo "FAIL: a missing file list must fail the step"
+  failed=1
+else
+  echo "PASS: a missing file list fails the step"
+fi
+
+exit ${failed}
--- a/.github/actions/check-rebuild/test_check_if_rebuild_requires.sh
+++ b/.github/actions/check-rebuild/test_check_if_rebuild_requires.sh
@@ -8,26 +8,21 @@
 # @see      https://github.com/nnstreamer/nnstreamer
 # @author   MyungJoo Ham <myungjoo.ham@samsung.com>
 #
-# Runs the same "out=$(...) || exit; grep -c ... || true" pattern as
-# .github/actions/check-rebuild/action.yml under "sh -e", which is how GitHub
-# executes composite "shell: sh" steps. This keeps the REBUILD=NO path and the
-# fail-closed error path exercised by CI on every PR.
+# Runs get_rebuild_flag.sh -- the same script action.yml invokes -- under
+# "sh -e", which is how GitHub executes composite "shell: sh" steps. This
+# keeps the REBUILD=NO path and the fail-closed error path exercised by CI
+# on every PR.
 
 set -u
 
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
-CHECKER="${SCRIPT_DIR}/check_if_rebuild_requires.sh"
+FLAG_SCRIPT="${SCRIPT_DIR}/get_rebuild_flag.sh"
 failed=0
 
 # run_step <file-list-path> <mode>: emulate the action.yml step under sh -e.
 run_step() {
-  sh -ec '
-    out=$(bash "$1" "$2" "$3") || {
-      echo "::error::check_if_rebuild_requires.sh failed" >&2
-      exit 1
-    }
-    printf "%s\n" "$out" | grep -c "REBUILD=YES" || true
-  ' sh "$CHECKER" "$1" "$2"
+  sh -ec 'rebuild=$(bash "$1" "$2" "$3"); printf "%s\n" "$rebuild"' \
+    sh "$FLAG_SCRIPT" "$1" "$2"
 }
 
 # expect_rebuild <expected 0|1> <mode> <changed-file>...
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -144,6 +144,9 @@ jobs:
         # Originally from "pr-prebuild-signed-off-by"
         run: |
           bash .github/workflows/static.check.scripts/signed-off-by.sh ${{ github.event.pull_request.commits }}
+      - name: /Checker/ check-rebuild action self-test
+        run: |
+          bash .github/actions/check-rebuild/test_check_if_rebuild_requires.sh
       - name: /Checker/ shellcheck for shell scripts
         # Originally from "pr-prebuild-shellcheck"
         run: |
--- a/.github/workflows/ubuntu_clean_llvm_build.yml
+++ b/.github/workflows/ubuntu_clean_llvm_build.yml
@@ -41,6 +41,11 @@ jobs:
       if: env.rebuild == '1'
     - run: meson test -C build/ -v
       if: env.rebuild == '1'
+    - name: build with orcc-support disabled
+      if: env.rebuild == '1'
+      run: |
+        meson setup build-noorc/ -Dorcc-support=disabled
+        meson compile -C build-noorc/
     - uses: actions/upload-artifact@v4
       if: failure()
       with:
--- a/.github/workflows/ubuntu_clean_llvm_build.yml
+++ b/.github/workflows/ubuntu_clean_llvm_build.yml
@@ -46,8 +46,13 @@ jobs:
       run: |
         meson setup build-noorc/ -Dorcc-support=disabled
         meson compile -C build-noorc/
+      env:
+        CC: clang
+        CXX: clang++
     - uses: actions/upload-artifact@v4
       if: failure()
       with:
         name: Meson_LLVM_Testlog
-        path: build/meson-logs/testlog.txt
+        path: |
+          build/meson-logs/testlog.txt
+          build-noorc/meson-logs/meson-log.txt
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1309,6 +1309,8 @@ gst_tensor_transform_typecast (GstTensorTransform * filter,
     orc_typecast (inptr, outptr, num, in_info->type, out_info->type);
     return GST_FLOW_OK;
   }
+#else
+  UNUSED (filter);
 #endif
 
   in_element_size = gst_tensor_get_element_size (in_info->type);


myungjoo:

## Review — [CI] macOS prerequisites split out of #4855

This is exactly the split that was asked for: two self-contained commits, 6 added lines, no dependency on the LiteRT work. Scope is right and the sequencing rationale in the description is sound.

### Commit 1 is verified working — the evidence is in the job records

I compared the macOS job before and after:

| Run | Branch | Duration | `homebrew` / `install minimal requirements` / `meson-build` |
| --- | --- | --- | --- |
| 32469757676 (2026-08-21) | `filter-post-error-message` | **10s** | **skipped / skipped / skipped** |
| 32708457950 (this PR) | `ci-macos-prereq` | **1m37s** | **success / success / success** |

So the claim holds and the fix demonstrably works — the job reported green for months while building nothing. Also confirmed the supporting claim: `check_if_rebuild_requires.sh` can only ever emit one `REBUILD=` line (each `REBUILD=YES` is followed by `exit 0`, and the `case` arms are mutually exclusive), so `grep -c` yields exactly `0` or `1`.

### 1. (Major) `|| true` is load-bearing, and the branch it protects has never run

GitHub executes `shell: sh` as `sh -e {0}`. In the old form the pipeline ended in `wc`, which exits 0 regardless. In the new form `grep -c` is the last command and exits **1** when the count is zero — that is, precisely on `REBUILD=NO`. Without `|| true`, every docs-only PR would fail this step across all **8 workflows** that consume this action (`android`, `gbs_build`, `macos`, `pdebuild`, `risc-v`, `ubuntu_clean_llvm_build`, `ubuntu_clean_meson_build`, `yocto`).

The `|| true` is there, so the code is right. The problem is that nothing has proven it: every job on this PR resolves to `REBUILD=YES`, because the changed files (`action.yml`, `gsttensor_transform.c`) hit the catch-all `*` arm of the helper script for every mode. **The `REBUILD=NO` path of the new code has not been exercised by a single run.**

For a change whose blast radius is all 8 workflows, that is worth closing before merge: push a scratch branch whose only change is a `.md` file and confirm the jobs log `Rebuild required: 0` and skip cleanly instead of erroring.

### 2. (Major) Commit 2's stated justification does not apply to this PR, and no CI job compiles the code it fixes

The description says the macOS job uses `-Dorcc-support=disabled`. On this PR's own run it does not — meson reports:

```
Program orcc found: YES (/opt/homebrew/bin/orcc)
Run-time dependency orc-0.4 found: YES 0.4.42
```

`orc` arrives as a transitive dependency of `brew install gstreamer`. So `HAVE_ORC` is **defined** in the build this PR unblocks, and the newly added `#else UNUSED (filter)` branch is never compiled by it. (The build passing here therefore does not validate commit 2 at all.)

Widening the check: `grep -rn "orcc-support" .github/ packaging/ debian/` returns nothing. **No CI configuration anywhere in the repo builds nnstreamer without orc.** The fix is correct and harmless, but it is both unverified and unguarded — if someone reintroduces an unused parameter under `#ifndef HAVE_ORC`, CI stays green and the next person to build with orc disabled eats the error.

Two ways out, and they differ in what they buy:

- **(a)** Move this commit into #4855, where `-Dorcc-support=disabled` is actually set. That keeps 4858 a pure CI-plumbing PR and puts the fix next to the configuration that needs it. Cheapest, but still leaves the configuration covered only by the macOS job.
- **(b)** Keep it here and add one `meson setup ... -Dorcc-support=disabled` configure-and-compile to the Ubuntu job. This is the only option that creates an actual regression gate for the non-orc build.

I'd suggest (b), because the whole premise of this PR is that an untested CI path stayed broken and invisible for a long time.

For what it's worth, the fix itself is complete for this file: I checked the other three in-function `#ifdef HAVE_ORC` sites, and `gst_tensor_transform_typecast()` is the only one with the problem — `gst_tensor_transform_set_property()` (line 1076) uses `filter` in both branches, and `gst_tensor_transform_arithmetic()` (line 1349) uses `filter` again at line 1411, outside the `#endif`. `UNUSED` is available via the `nnstreamer_util.h` include at line 53. The only other `HAVE_ORC` consumers in the tree are `tensor_common.h` and `unittest_plugins.cc`.

### 3. (Minor) The action still fails open

`rebuild=\`... || true\`` means that *any* failure inside the substitution — the helper script erroring on a missing argument, an unexpected `mode`, an empty `tmpfile` — produces `0` and **silently skips the whole job**. That is the exact shape of the bug this PR is fixing, just with a different trigger.

Since the value is now provably `0` or `1`, a three-line assertion converts the next instance from "green for months" into a red job:

```sh
case "$rebuild" in
  0|1) ;;
  *) echo "::error::unexpected rebuild value '$rebuild'"; exit 1 ;;
esac
```

Optional, but this PR is the natural place for it.

### 4. (Trivial) Comment wording

The inline comment explains *what* broke but not why nobody noticed. Adding that the job still reported success in ~10 seconds is the part a future reader actually needs in order to recognise the pattern elsewhere.

### Other checks

- **Regression risk to other modules:** `UNUSED` expands to `do { (void)(expr); } while (0)` — a no-op. The `action.yml` change is behaviour-preserving on GNU platforms (`grep | wc -l` and `grep -c` produce identical unpadded counts there), so the 7 non-macOS workflows should be unaffected; the macOS job changing from "skips everything" to "actually builds" is the intended and only behavioural change.
- **Documentation:** no API or ARCHITECTURE change; nothing to update.
- **Security:** no new attack surface, no new external inputs, no back-door-shaped code.
- **Follow-up noted in the description:** the `orc_int64` vs `int64_t` mismatch in `unittest_plugins.cc` remains open. Worth filing as an issue rather than leaving it in a PR description, since it is the reason #4855 has to disable orc on macOS in the first place.

jaeyun-jung (approved):

myungjoo:

## Follow-up review — [CI] macOS prerequisites split out of #4855

This reviews the latest commit (`Two hardenings requested by the review of this PR`, `63af703`→`+1` commit) against the prior review posted on 2026-08-24, plus a fresh pass over scope/regression/tests/docs. I re-read the four changed files in the worktree (`.github/actions/check-rebuild/action.yml`, `.github/actions/check-rebuild/check_if_rebuild_requires.sh`, `.github/workflows/ubuntu_clean_llvm_build.yml`, `gst/nnstreamer/elements/gsttensor_transform.c`) and cross-checked against `meson.build`/`meson_options.txt`.

### Status of the four prior findings

**#2 (was Major) — orc fix was unguarded by CI → RESOLVED.** `ubuntu_clean_llvm_build.yml` now adds a `build with orcc-support disabled` step (`meson setup build-noorc/ -Dorcc-support=disabled; meson compile -C build-noorc/`) gated on `env.rebuild == '1'`. I confirmed in `meson.build` (line ~575-577) that `orcc-support` is the feature option that sets the `HAVE_ORC` project arg, so `-Dorcc-support=disabled` genuinely undefines `HAVE_ORC` and exercises the new `#else UNUSED (filter)` branch in `gst_tensor_transform_typecast()`. This is option (b) from the prior review and is the right choice — it creates a real, permanent regression gate rather than just moving the fix to #4855.

Two small (non-blocking) observations on this new step:
- (Trivial) The other steps in this same job explicitly set `CC: clang` / `CXX: clang++`; the new step doesn't, so it likely builds with the default system compiler (gcc on `ubuntu-22.04`), not clang, despite the job being named "...with LLVM/Clang". This doesn't undermine correctness — `-Werror=unused-parameter` here comes from the project-wide `warning_level=2` + `werror=true` in `meson.build`, not a compiler-specific flag — but it's a bit inconsistent with the job's stated purpose.
- (Trivial) The `if: failure()` artifact-upload step only captures `build/meson-logs/testlog.txt` (the with-orc build). If `build-noorc/` fails to compile, the job correctly turns red (blocking merge), but no log artifact is attached for that failure — only the inline Action log. Minor DX gap, not a functional issue.

**#3 (was Minor) — check-rebuild failed open on script errors → RESOLVED.** `action.yml` now separates the script's own exit status from the `grep -c` count: `out=$(bash check_if_rebuild_requires.sh ...) || { echo "::error::..."; exit 1; }`, then computes `rebuild` from `$out`. I traced the `sh -e` semantics: because each line is of the form `cmd || { ... }` / `cmd || true`, both are AND-OR lists, which are exempt from `errexit` aborting mid-list — so the `|| { exit 1; }` branch reliably fires on a real script failure, and the `|| true` reliably swallows only the "count is 0" case. `check_if_rebuild_requires.sh` was also hardened: `[ -z $1 ]` (only caught a literally-missing arg, unquoted) → `[ ! -f "$1" ]` (quoted, also rejects a nonexistent/unreadable file). Correct and matches what was asked.

**#4 (was Trivial) — comment should explain why nobody noticed → substantially addressed.** The new inline comment says the padding "made `env.rebuild == '1'` always false there (macOS jobs silently skipped every build step)" — it doesn't restate the "~10s, reported success" specifics from the original suggestion, but it does capture the actionable insight (silent full-job skip). Fine as-is.

**#1 (was Major) — the `REBUILD=NO` path of the new code had never been exercised by an actual CI run → still open, but now explicitly acknowledged and reasonably mitigated.** The PR description now states this directly: a docs-only scratch PR can't exercise the *new* action code before merge (the diff to `action.yml` itself would flip every workflow to `REBUILD=YES` via the catch-all arm), so the author verified all `REBUILD` paths locally under `sh -e` and commits to checking the first post-merge docs-only PR. That's a legitimate structural limitation (chicken-and-egg for CI-plumbing PRs), not negligence, and the reasoning is sound. I'm downgrading this to **Minor**: it's still true that this exact code path (the one that guards 8 workflows from running on every PR) merges without a live-fire test, and the failure mode if it's ever wrong is the same silent-skip pattern this PR was written to fix. Two low-cost options if the author wants to close this fully rather than defer it:
- Actually push a throwaway docs-only branch/PR against this branch before merging and confirm `Rebuild required: 0` in the logs (the description says this "cannot" be done before merge because the action-change itself would flip to YES — that's true for a PR *against this PR's own diff*, but a scratch PR based on `main` pointed at a *separate* already-merged copy of this action, or simply testing via `act`/local `sh -e` harness as already done, are the practical substitutes; the author has effectively done the closest available thing).
- Land the optional `case "$rebuild" in 0|1) ;; *) exit 1;; esac` sanity assertion suggested last time — it doesn't test the `NO` path per se, but it does turn any future silent-open failure mode into a red job, which is the actual risk being managed here. Still not applied; still optional.

Not a blocker given the explanation, but worth a short follow-up comment/issue confirming the first post-merge docs-only PR actually shows `Rebuild required: 0` across all 8 workflows, since that's the real proof this PR is deferring.

### Fresh checks

- **Does it solve the stated problem?** Yes. Commit 1 fixes the padded-`wc` bug (confirmed independently by CI: `Build on MacOS-latest with Meson` on this PR passed in **1m29s**, doing a real build, vs. the ~10-15s skip-everything green it used to be). Commit 2 fixes the unused-parameter build error for `-Dorcc-support=disabled`, now with a CI gate that actually compiles that configuration.
- **Regression risk to other modules:** Low. `UNUSED()` is a no-op wrapper, only compiled into the non-orc path. `grep -c` vs `grep | wc -l` are equivalent on GNU platforms, so the other 7 workflows consuming `check-rebuild` are behavior-preserving for the common case. The two hardening changes only change behavior on the error path (a broken/missing file-list argument), which was already a bug, not a supported case.
- **Scope / size vs. topic:** Appropriately small — 4 files, ~19 net added lines, no unrelated modules touched. Matches the PR's own "self-contained fixes split out of #4855" framing.
- **Tests / durability against future breakage:** The new `build-noorc` CI step is a good, durable regression gate for commit 2 — any future `#ifdef HAVE_ORC` misuse in this function class will now fail CI. There is still no automated test (shell unit test, e.g. bats/shunit2) for `check_if_rebuild_requires.sh`'s own YES/NO logic; a future regression there (e.g., someone reintroducing a piped-`wc` pattern, or breaking the mutual-exclusivity of the `REBUILD=` lines this PR's `grep -c` correctness depends on) would not be caught by CI — it would resurface as the same silent-skip failure mode this PR fixes, which is exactly the class of bug that's hardest to notice. Minor, non-blocking suggestion for a follow-up.
- **CI status at review time:** `Build on MacOS-latest with Meson` — pass (real build, 1m29s). `Static checks`, `CPP-Linter`, `Spell Check`, `DCO` — pass. `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` — still queued/pending as of this review; this is the job carrying the new `orcc-support=disabled` gate for commit 2, so it's worth confirming it goes green before merging (it did pass meson.build's static reasoning check above, but hasn't been observed running yet).
- **Documentation / architecture / API impact:** None applicable — pure CI plumbing plus a one-line `UNUSED()` annotation. Nothing to update.

### Severity summary

| # | Item | Severity | Status |
|---|---|---|---|
| 1 | `REBUILD=NO` path unexercised by a live CI run | Minor (was Major) | Open, explained & mitigated; recommend post-merge confirmation |
| 2 | orc fix previously unguarded by any CI config | — | **Resolved** — new non-orc build gate added and verified against `meson.build` |
| 2a | new step doesn't set `CC=clang`/`CXX=clang++` like its siblings | Trivial | Open, cosmetic |
| 2b | failure-artifact upload doesn't cover `build-noorc/` logs | Trivial | Open, cosmetic |
| 3 | check-rebuild failed open on script errors | — | **Resolved** |
| 4 | comment should explain the "silently green" symptom | Trivial | Resolved in spirit |
| 5 | no automated unit test for `check_if_rebuild_requires.sh` logic | Minor | New observation, non-blocking |

### Verdict

No blocking issues found. Both Major items from the prior review have been addressed — one with a genuine, verifiable CI gate (confirmed against `meson.build`'s `orcc-support` option wiring), the other with a documented, reasonable mitigation for a structural bootstrapping limitation that can only be fully closed post-merge. The Minor item is fully resolved. Remaining points (2a, 2b, 5) are Trivial/Minor polish suggestions, not merge blockers. Recommend merging once `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` (currently pending) is confirmed green, and following up post-merge to confirm the first docs-only PR shows `Rebuild required: 0` cleanly across all 8 workflows.

myungjoo:

## Round-3 review — [CI] macOS prerequisites split out of #4855

This reviews the latest commit (`77f580a0`, "[CI] Add check-rebuild self-test and polish the non-orc gate") against the second-round review posted 2026-08-25 (comment `5406420180`), plus a fresh scope/regression/docs pass. I re-read all five files in the diff, ran `test_check_if_rebuild_requires.sh` locally, cross-checked the actual GitHub Actions logs for this PR's own HEAD commit, checked branch protection, and did targeted mutation testing on the checker script.

### Status of the four open items from round 2

**2a (Trivial) — new non-orc step didn't use clang like its siblings → RESOLVED.** `ubuntu_clean_llvm_build.yml`'s "build with orcc-support disabled" step now carries `env: { CC: clang, CXX: clang++ }`, matching the `meson setup build/` step above it.

**2b (Trivial) — failure-artifact upload didn't cover `build-noorc/` → RESOLVED, with a caveat.** The `upload-artifact` step's `path:` now includes `build-noorc/meson-logs/meson-log.txt`. Caveat: `meson-log.txt` is populated mainly by `meson setup` (the configure step); if the realistic failure mode — `meson compile -C build-noorc/` failing on `-Werror=unused-parameter` — actually occurs, the compiler error itself may not land in that file, only in the raw Action step log (which is still visible, just not as a downloadable artifact). Not a functional gap, just limited DX value; not blocking.

**5 (Minor) — no automated unit test for `check_if_rebuild_requires.sh` → RESOLVED, and verified to be a real (not just for-show) test.**
- It runs where claimed: I pulled the actual log for `Static checks` on this PR's own HEAD (`77f580a0`, run `32817154516`, job `97707652259`) and confirmed the new step `/Checker/ check-rebuild action self-test` executes and all 15 assertions print `PASS` (12 mode/file combinations + the fail-closed "missing file list" case).
- It genuinely fails on a real regression: I mutated a copy of `check_if_rebuild_requires.sh` to drop the `rebuild='1'` assignment from the catch-all `case` arm (a plausible real-world regression) and re-ran the self-test against the mutant — it correctly reported `FAIL: mode=build files=[...] expected=1 got=0` and exited non-zero. This is genuine mutation-tested coverage of the checker's decision logic, not a tautological test.
- It is enforced, not advisory: `Static checks` is in `required_status_checks.contexts` on `main`'s branch protection, so a red self-test genuinely blocks merge — this closes the loop the PR set out to close (CI failing loudly instead of silently skipping).
- Shellcheck gate: will pass. Two independent reasons — (a) `static.check.scripts/shellcheck.sh` never actually fails the job unless shellcheck's raw output exceeds 9000 lines, so it is structurally a soft/informational gate for any realistically-sized script; (b) empirically, the same CI run reports `shellcheck: passed ... There are 0 lines` for both `test_check_if_rebuild_requires.sh` and `check_if_rebuild_requires.sh` — zero findings, not just under-threshold ones.

**1 (Minor, was Major) — `REBUILD=NO` path of `action.yml`'s own code never exercised by a live run → partially closed; see new finding below, which supersedes it.**

### New finding this round

**(Major) The self-test verifies a hand-duplicated copy of `action.yml`'s step, not `action.yml` itself — so a future regression in the actual composite action would still go undetected and merge silently green.**

`test_check_if_rebuild_requires.sh`'s `run_step()` re-implements the `out=$(...) || { exit 1; }; ... grep -c ... || true` pipeline as a literal string inside the test file:

```sh
run_step() {
  sh -ec '
    out=$(bash "$1" "$2" "$3") || {
      echo "::error::check_if_rebuild_requires.sh failed" >&2
      exit 1
    }
    printf "%s\n" "$out" | grep -c "REBUILD=YES" || true
  ' sh "$CHECKER" "$1" "$2"
}
```

I checked whether the test reads, sources, or otherwise executes `action.yml` anywhere: `grep -n "action.yml" test_check_if_rebuild_requires.sh` matches only three comment lines (the file docblock and one inline comment) — never the file's content. The test's own commit message and docblock say it exercises "the exact ... step pattern of `action.yml`," which is only true as of the moment the two were written in sync; nothing keeps them in sync afterward.

Practical consequence: if a future PR edits `action.yml`'s `run:` block — e.g. reverts to `grep | wc -l`, drops the `|| true`, or breaks the fail-closed `|| { exit 1; }` — the self-test still passes (it never touches that file), `Static checks` stays green, and the regression ships to all 8 consuming workflows exactly as invisibly as the bug this PR fixes shipped originally. This is the same failure category the PR exists to eliminate, just moved one file over: `check_if_rebuild_requires.sh`'s logic is now well-guarded (see mutation test above), but `action.yml`'s own script text is not.

This doesn't need to block this PR — the current state is still a large net improvement over "zero tests, fails open" — but I'd flag it as Major rather than close it out, because it's easy to read the round-2/round-3 exchange as "item 1 resolved" when what actually happened is the coverage moved to an adjacent file that can drift. Two low-cost ways to close it, either is fine:
- Have `action.yml`'s `run:` block and the test's `run_step()` both call a single shared script file (e.g. move the `out=$(...) || …; grep -c … || true` pipeline into a tiny `run_check.sh` that `action.yml` invokes and the test also invokes), so there is exactly one copy to drift.
- At minimum, add a one-line comment in both files pointing at each other ("keep this pipeline in sync with test_check_if_rebuild_requires.sh's run_step()" / vice versa) so the drift risk is at least visible to the next editor, even if not mechanically prevented.

### Fresh checks

- **Solves the stated topic?** Yes — all four round-2 follow-ups are addressed (three fully, one partially per above).
- **Regression risk to other modules:** None. This round's diff touches only CI plumbing (`static.check.yml` +3 lines, `ubuntu_clean_llvm_build.yml` a few lines for `CC/CXX` env + one artifact path, plus the new standalone test script). No production code (`gsttensor_transform.c` etc.) changed in this commit; that fix was already reviewed and confirmed in round 2.
- **Scope/size vs. topic:** Appropriate — 85 added lines across 3 files, tightly scoped to exactly the round-2 follow-ups. Nothing unrelated touched.
- **Does the new self-test actually run on `pull_request` for this PR's own workflow file (not the base branch's stale copy)?** Verified empirically, not just reasoned about: the `Static checks` run for this PR's HEAD commit (`77f580a0`) shows `Preparing step 1... Checking out` landing at `77f580a0`, and the new self-test step executing with all 15 `PASS` lines in its log.
- **Test durability against future breakage in other modules:** For `check_if_rebuild_requires.sh`'s own logic, yes (mutation-verified above). For `action.yml` itself, not yet (see Major finding).
- **CI status at review time:** `Static checks` — pass (confirms the new self-test ran and passed). `Build on MacOS-latest with Meson` — pass, 2m2s, doing a real build. `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` — the job carrying the new `clang`-flagged `build-noorc` gate — was still `in_progress` as of this review; both it and `Static checks` are in `main`'s `required_status_checks`, so confirm it goes green before merging.
- **Documentation/architecture/API impact:** None applicable — pure CI plumbing and a self-test, no public API or `ARCHITECTURE.md`-relevant surface touched.

### Severity summary

| # | Item | Severity | Status |
|---|---|---|---|
| 2a | non-orc step didn't use clang | Trivial | **Resolved** |
| 2b | failure-artifact upload missing `build-noorc/` logs | Trivial | **Resolved** (limited practical value — see caveat) |
| 5 | no automated unit test for `check_if_rebuild_requires.sh` | Minor | **Resolved** — verified running, required-check-gated, and mutation-tested |
| 1 | `REBUILD=NO` path unexercised by live CI | Minor | Substantially mitigated for the checker's logic; superseded by the new finding below |
| new | self-test duplicates `action.yml`'s step instead of exercising the real file — future `action.yml` regressions would still merge silently green | Major | Open, non-blocking suggestion |

### Verdict

No hard blockers. All four round-2 items are addressed to a reasonable standard, and two of them (5, and by extension the practical substance of 1) are independently verified here, not just taken on trust — I ran the test locally, checked the live CI logs for this exact commit, confirmed `Static checks` is a required status check, and mutation-tested the checker's decision logic. The one new finding (self-test duplicates rather than exercises `action.yml`) is worth a deliberate decision rather than a silent close-out, but it doesn't regress anything and the PR is still strictly safer than before it existed, so I would not hold up merge for it — a follow-up issue/PR to de-duplicate the pipeline logic (or at minimum cross-reference comments) would be a good use of five minutes. Recommend merging once `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` (pending at review time) is confirmed green.

myungjoo:

## Round-4 review — [CI] macOS prerequisites split out of #4855

This reviews the latest commit `e1338d6e` ("[CI] Share the rebuild-flag logic between action.yml and its test") against the sole open item from the round-3 review (comment `5406532769`): *self-test hand-duplicates `action.yml`'s shell pipeline as a string, so a future edit to the composite action could silently diverge from what the test verifies (Major, non-blocking)*. I re-read the full diff of `e1338d6e`, the current contents of `action.yml`, the new `get_rebuild_flag.sh`, and `test_check_if_rebuild_requires.sh`; ran the self-test locally; and cross-checked this PR's own CI logs at HEAD.

### 1. Is the round-3 Major finding actually resolved? Yes.

The decision-logic that used to be duplicated (`out=$(...) || { exit 1; }; ... | grep -c ... || true`) is now a single file, `.github/actions/check-rebuild/get_rebuild_flag.sh`, and both `action.yml` and `test_check_if_rebuild_requires.sh` invoke it as an external script rather than re-typing its body:

- `action.yml`: `rebuild=$(bash .github/actions/check-rebuild/get_rebuild_flag.sh ${tmpfile} ${{ inputs.mode }})`
- `test_check_if_rebuild_requires.sh`'s `run_step()`: `sh -ec 'rebuild=$(bash "$1" "$2" "$3"); printf "%s\n" "$rebuild"' sh "$FLAG_SCRIPT" "$1" "$2"`

What remains duplicated between the two call sites is now only the generic "assign a command-substitution result under `sh -e`" shape — a single line with no decision logic in it — which is a categorically smaller drift surface than the multi-line pipeline that was duplicated before. If `action.yml`'s script changes what it does (e.g., how the flag is computed, the fail-closed handling), that logic now lives in `get_rebuild_flag.sh` and is exercised by both callers automatically. **This resolves the Major finding as requested.**

### 2. Does `get_rebuild_flag.sh` stay fail-closed under `sh -e`?

Traced this end to end:

- `check_if_rebuild_requires.sh` (unchanged in this commit) exits 0 on both `REBUILD=YES` and `REBUILD=NO`; it only exits nonzero on a real error (e.g. missing/unreadable file-list argument).
- Inside `get_rebuild_flag.sh`, `out=$(bash ".../check_if_rebuild_requires.sh" "$1" "$2") || { echo "::error..." >&2; exit 1; }` — so a genuine checker failure makes `get_rebuild_flag.sh` itself exit 1.
- Its last line, `printf '%s\n' "$out" | grep -c "REBUILD=YES" || true`, always resolves to exit 0 in the normal YES/NO cases (the `|| true` only ever swallows `grep -c`'s "0 matches" exit status, never a real error), so the script's own exit status is a clean signal: **0 = flag computed (value 0 or 1 on stdout), 1 = checker actually failed.**
- In `action.yml`, `rebuild=$(bash .../get_rebuild_flag.sh ${tmpfile} ${{ inputs.mode }})` is a bare top-level statement (not inside `&&`/`||`/`if`), so under POSIX `sh -e` semantics a nonzero exit from the command-substitution aborts the composite step immediately, before `echo "rebuild=..." >> $GITHUB_ENV"` runs — turning the step red instead of silently producing `rebuild=0`. I verified this exact assignment-under-`-e` behavior empirically with `bash --posix`/`sh` locally (`x=$(cmd_that_returns_1)` under `set -e` aborts the script before the next line), which matches the standard POSIX rule that only commands protected by a conditional (`if`, `&&`, `||`, negation) are exempt from `errexit`.

This confirms the PR description's claim ("get_rebuild_flag.sh exits nonzero on checker failure, keeping the fail-closed behavior") holds under the shell semantics GitHub actually uses for `shell: sh` composite steps.

### 3. Quality review of the `e1338d6e` diff itself

- **Shellcheck**: confirmed via this PR's own CI run (`Static checks`, run `32818067382`, commit `e1338d6e`) — `shellcheck: passed. file name: .github/actions/check-rebuild/get_rebuild_flag.sh There are 0 lines.` (zero findings, not just under a threshold).
- **Self-test actually exercises the new file, on this PR's own CI**: same run's `/Checker/ check-rebuild action self-test` step logs all 15 `PASS` lines (12 mode/file combinations + the fail-closed missing-file-list case), confirming the test runs against `get_rebuild_flag.sh` (not a stale/cached copy) at this exact commit.
- **`Static checks` remains a required status check** on `main` (confirmed via `repos/.../branches/main/protection` → `required_status_checks.contexts` includes `"Static checks"`), so a red self-test genuinely blocks merge — the enforcement loop this PR is building is real, not advisory.
- **Locally reproduced**: ran `bash .github/actions/check-rebuild/test_check_if_rebuild_requires.sh` in the checked-out worktree; all 15 assertions pass, exit code 0.
- **(Trivial)** A small stylistic residue: `action.yml` still passes `${tmpfile}` / `${{ inputs.mode }}` unquoted into the command substitution, while the test's `run_step()` quotes `"$1" "$2" "$3"`. This is pre-existing (not introduced by this commit) and practically inert — `tmpfile` comes from `mktemp` and `mode` is always one of a fixed set of literal strings in the calling workflows — but since the two call sites are now meant to mirror each other, quoting them identically would remove the last cosmetic asymmetry. Not blocking.
- **(Trivial)** `get_rebuild_flag.sh` and `check_if_rebuild_requires.sh` both use `#!/usr/bin/env bash` but are always invoked via an explicit `bash script.sh`, so the shebang line is dead weight (harmless, just unused) — consistent with the pre-existing script and not a regression.

### 4. Regression / scope check for this commit

- Only 3 files touched (`action.yml`, new `get_rebuild_flag.sh`, `test_check_if_rebuild_requires.sh`), net +38/-22 lines. No production code touched in this commit (the `gsttensor_transform.c` fix and the `ubuntu_clean_llvm_build.yml` non-orc gate were already reviewed and confirmed in rounds 2–3 and are untouched here).
- Verified the composite action's public contract (`inputs.mode` in, `env.rebuild` out) is unchanged, so none of the 8 consuming workflows (`android`, `gbs_build`, `macos`, `pdebuild`, `risc-v`, `ubuntu_clean_llvm_build`, `ubuntu_clean_meson_build`, `yocto` — confirmed by grepping `uses:.*check-rebuild` across `.github/workflows/`) need any change and none is regressed by this refactor.
- No documentation/architecture/API surface is affected — pure CI-internal refactor.

### 5. CI status at review time (commit `e1338d6e`)

- **Pass**: `Static checks` (carries the self-test + shellcheck gate for this commit's own files), `Spell Check with Typos`, `CPP-Linter`, `Analyze (python)`, `DCO`.
- **Still pending/in-progress** at review time: `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` (carries the `-Dorcc-support=disabled` gate from round 2), `Build on MacOS-latest with Meson`, `Build with meson and test with Valgrind` (both variants), `Tizen GBS build` (all three), `Ubuntu pdebuild`, `Build with Yocto`, `Android NDK build`, `Build on Ubuntu 22.04 RISC-V 64`, `Analyze (c-cpp)`. `Build on Ubuntu with LLVM/Clang` and `Build on MacOS-latest with Meson` are both in `main`'s required-status-checks list, so they should be confirmed green before merging, as recommended in the round-3 review.

### Severity summary

| # | Item | Severity | Status |
|---|---|---|---|
| — | Round-3's sole open item: self-test duplicated `action.yml`'s pipeline as a hand-copied string | Major | **Resolved** — decision logic extracted to `get_rebuild_flag.sh`, shared by both callers |
| — | Fail-closed behavior of `get_rebuild_flag.sh` under `sh -e` | — | Verified correct (traced + reproduced locally) |
| — | Shellcheck / self-test actually running against the new script in this PR's own CI | — | Verified empirically from live CI logs (0 shellcheck findings, 15/15 self-test PASS) |
| 1 | Residual unquoted-vs-quoted asymmetry between `action.yml` and `run_step()`'s invocation of the shared script | Trivial | Open, cosmetic, pre-existing pattern |
| 2 | Unused bash shebang on scripts always invoked via explicit `bash` | Trivial | Open, cosmetic, pre-existing pattern |
| 3 | `Build on Ubuntu with LLVM/Clang` and `Build on MacOS-latest with Meson` (both required checks) still pending at review time | — | Not a code issue; confirm green before merge |

### Verdict

No blocking issues. The one substantive item carried over from round 3 (Major) is resolved in a structurally sound way — a single shared script, verified fail-closed under the shell semantics GitHub actually uses, verified by shellcheck with zero findings, and verified to actually execute (not just exist) against this exact commit via the PR's own required `Static checks` run. The two remaining observations are trivial/cosmetic and not worth blocking on. Recommend merging once the still-pending required checks (`Build on Ubuntu with LLVM/Clang`, `Build on MacOS-latest with Meson`) finish green.

PR #4859, opened 2026-08-25
[Filter/dali] Fix resource leaks in cleanup()

myungjoo:
## What

Fixes resource leaks in `dali_subplugin::cleanup()`:

1. `_pipeline_path` (allocated with `g_strdup()` in `loadPipeline()`) was never freed anywhere, leaking on every close.
2. The `if (!configured) return;` guard skipped all cleanup when `configure_instance()` threw before its last line, where `configured` is set. #4829 (still open; this PR should land after it) makes the framework delete the instance when open fails, so the destructor then runs on such partially-configured instances and must release whatever was acquired — not just `_pipeline_handle` and `_pipeline_path`, but also the two `GstTensorsInfo` metas that `configure_instance()` copies from the filter properties before any of its throw sites.

## How

Drop the guard and free `_pipeline_path`. Every free in `cleanup()` is null-safe on a never-configured instance: members are value-initialized (`{}`), tensor metas are initialized in the constructor, and `_pipeline_handle` is checked before `daliDeletePipeline()`.

`cleanup()` is now fully idempotent, which is what makes dropping the guard safe: `gst_tensors_info_free()` re-initializes the metas, and every raw pointer member — including the `_nns_input_name`/`_nns_output_name` aliases into the tensor metas — is reset to `nullptr`, so no member points at freed memory after `cleanup()` returns.

Identified during the (AI-assisted) review of #4829; the `_pipeline_path` leak is pre-existing and independent of that PR.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_dali.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_dali.cc
@@ -199,17 +199,18 @@ dali_subplugin::~dali_subplugin ()
 void
 dali_subplugin::cleanup ()
 {
-  if (!configured) {
-    return;
-  }
-
   if (_pipeline_handle) {
     daliDeletePipeline (&_pipeline_handle);
     _pipeline_handle = nullptr;
   }
 
+  g_free (_pipeline_path);
+  _pipeline_path = nullptr;
+
   gst_tensors_info_free (&_inputTensorMeta);
   gst_tensors_info_free (&_outputTensorMeta);
+  _nns_input_name = nullptr;
+  _nns_output_name = nullptr;
 
   configured = false;
 }


myungjoo:

## Review — [Filter/dali] Fix resource leaks in cleanup()

Correct fix, minimal scope (+3/-4, one file), and it does what the description says. I verified each safety claim rather than taking them on trust:

- **All members are value-initialized.** `configured{}`, `_pipeline_handle{}`, `_pipeline_path{}` (line 143–164), so a never-configured instance has null everywhere `cleanup()` touches.
- **The tensor metas are initialized in the constructor** (`gst_tensors_info_init()` on both), so `gst_tensors_info_free()` is safe on a pristine object.
- **`g_free(nullptr)` and the `if (_pipeline_handle)` guard** cover the remaining two.

One property worth adding to the commit message, because it is what really makes dropping the guard safe: **`cleanup()` is now fully idempotent.** `gst_tensors_info_free()` ends with `gst_tensors_info_init (info)` (`nnstreamer_plugin_api_util_impl.c`), and `gst_tensor_info_free()` likewise re-inits after `g_free (info->name)`. Combined with the explicit `= nullptr` on the two raw pointers, calling `cleanup()` any number of times is safe. The old `if (!configured) return;` was providing that idempotency as a side effect, so it is worth stating explicitly that nothing was lost by removing it.

### The fix is larger than the description claims

The body lists `_pipeline_handle` and `_pipeline_path` as what leaked on the partially-configured path. It is more than that. `configure_instance()` opens with:

```cpp
gst_tensors_info_copy (std::addressof (_inputTensorMeta), std::addressof (prop->input_meta));
gst_tensors_info_copy (std::addressof (_outputTensorMeta), std::addressof (prop->output_meta));
```

Both happen **before** the `throw std::invalid_argument ("The pipeline file is not given.")` further down. So a throw there left two fully populated `GstTensorsInfo` behind — including every `g_strdup`'d tensor `name`, and the `extra` array if the tensor count exceeded `NNS_TENSOR_MEMORY_MAX` — and the old guard skipped all of it. Since #4829 that path now runs the destructor, so this PR reclaims those too. Worth mentioning, as it makes the case for the change stronger than stated.

### 1. (Minor, worth fixing here) `cleanup()` leaves two dangling aliases behind

```cpp
_nns_input_name  = input_tensor_info->name;   /* line 244 */
_nns_output_name = output_tensor_info->name;  /* line 251 */
```

These are non-owning aliases into `_inputTensorMeta` / `_outputTensorMeta`. `cleanup()` calls `gst_tensors_info_free()` on both metas, which `g_free`s exactly those strings — but unlike `_pipeline_handle` and `_pipeline_path`, the two aliases are **not** reset. After `cleanup()` the object holds two pointers to freed memory.

This is not reachable today: `cleanup()` has a single caller, the destructor (line 195), so the object dies immediately afterwards. But this PR's entire premise is that `cleanup()` should be safe to run at any point in the object's life, and `invoke()` reads `_nns_input_name` at lines 310 and 318 right after its `g_assert (configured)` at line 307. The moment `cleanup()` gains a second caller — a reconfigure path, an error path, anything — that becomes a use-after-free rather than a leak.

Two lines finish the job the PR started:

```cpp
_nns_input_name = nullptr;
_nns_output_name = nullptr;
```

Cheap now, and it keeps the invariant "after `cleanup()`, no member points at anything" true without exception.

### 2. (Minor) Nothing in CI can verify this

`dali-support` is `auto` in `meson_options.txt` and there are no `dali` hits in `.github/workflows/` or `packaging/nnstreamer.spec` — **no CI configuration builds this subplugin**, so neither the leak nor the fix is observable to any job. That is not this PR's fault and is not a reason to hold it, but it does mean the change rests entirely on review.

The constructive angle: this bug is one instance of a general class — "the framework deletes the instance when `configure_instance()` throws, so the destructor must release whatever was already acquired". That class *can* be gated generically, and the natural place is the new `cpp_mock_subplugin` in #4829. Making that mock allocate something in `configure_instance()` before throwing, and asserting it is released, would give a CI-visible guard covering every subplugin's contract compliance, including this one, without needing the DALI SDK. Given that #4829 is where this leak was found, closing the loop there seems worth it.

### 3. (Trivial) `configured` is still live

Just noting for the record, since the guard that read it was removed: `configured` is still read at line 307 (`g_assert (configured);` in `invoke()`), so it has not become a write-only field and does not need removing.

### Other checks

- **Regression risk:** the only behavioral change is that `cleanup()` now does work on instances where it previously returned early. Every operation in it is null-safe and idempotent per the verification above, and the representative empty instance created by `register_subplugin<dali_subplugin>()` — whose destructor runs at `fini_filter_dali()` and previously took the early return — now runs `gst_tensors_info_free()` on constructor-initialized metas, which is a no-op. No other module is touched.
- **Interaction with #4829:** correct dependency direction. This PR only matters because #4829 makes the framework delete the instance on a `configure_instance()` throw; before that, the partially-configured object leaked wholesale and freeing its members would have been moot. Worth noting in the description that this should land after #4829, or at least not before it.
- **Scope:** exactly right — one function, no drive-by changes.
- **API/ARCHITECTURE:** no public API change; nothing to document beyond what #4829 already added to the header.
- **Security:** no new attack surface; the change strictly reduces the amount of memory left allocated. No back-door-shaped code found.

myungjoo:
> _Note: this review was produced by another (AI) agent in a separate session and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Independent re-review — [Filter/dali] Fix resource leaks in cleanup()

This is a follow-up review done independently of the existing AI review comment on this PR (posted earlier). I re-verified the code from scratch against the current `fix/dali-cleanup-leaks` branch (`ext/nnstreamer/tensor_filter/tensor_filter_dali.cc`), and also checked whether the one substantive suggestion from the prior review was applied.

### Does it solve the stated problem? Yes.

- `_pipeline_path` (leaked unconditionally on every close, `g_strdup`'d in `configure_instance`/used by `loadPipeline`) is now freed and nulled in `cleanup()`.
- The `if (!configured) return;` guard is removed, so `cleanup()` now runs fully even when `configure_instance()` throws partway through. I re-checked null-safety of every statement in the resulting `cleanup()` on a never-configured object: `_pipeline_handle{}`/`_pipeline_path{}`/`configured{}` are value-initialized, `_inputTensorMeta`/`_outputTensorMeta` are `gst_tensors_info_init()`-ed in the constructor, `g_free(nullptr)` is a no-op, and the `if (_pipeline_handle)` guard covers `daliDeletePipeline`. Confirmed safe.
- Scope is minimal and correct: one function, +3/-4, no drive-by changes, no other module touched.

### Status of the prior review's finding — NOT applied

The existing AI review comment on this PR (posted 2026‑08‑25T05:53:34Z) flagged, as "(Minor, worth fixing here)":

> `cleanup()` calls `gst_tensors_info_free()` on `_inputTensorMeta`/`_outputTensorMeta`, which frees the exact strings that `_nns_input_name` (line 244) and `_nns_output_name` (line 251) alias — but neither raw pointer is reset afterward, leaving two dangling pointers on the object.

I checked the current branch content directly (not just the diff) and confirmed this is still unaddressed: `cleanup()` frees `_pipeline_handle` and `_pipeline_path` and nulls both, but `_nns_input_name`/`_nns_output_name` are left holding freed addresses. `gst_tensor_info_free()` in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c` (line ~140) does `g_free(info->name); gst_tensor_info_init(info);` — it resets the struct's own `name` field, not any external alias into it.

**Reachability today:** low. `cleanup()` has one caller (the destructor), and nothing in the class reads `_nns_input_name`/`_nns_output_name` afterward — the object is being destroyed. Not exploitable via the current single call site.

**Why it's still worth two lines:** this PR's entire premise is establishing "`cleanup()` is safe to call on a partially-configured object," precisely so a second future caller (a reconfigure path, an error path, whatever comes next) doesn't reintroduce a use-after-free. Leaving these two aliases dangling keeps exactly that class of bug alive for the next person who calls `cleanup()` from anywhere but the destructor. `invoke()` already reads both fields right after `g_assert(configured)` (which is a no-op if assertions are compiled out), so the failure mode if this is ever violated is a use-after-free, not merely a leak. Since this was already raised once and the PR was updated for the `_pipeline_path` leak but not for this, it looks like it was simply missed rather than intentionally deferred — worth a fast follow-up:

```cpp
_nns_input_name = nullptr;
_nns_output_name = nullptr;
```

### New finding — PR description overstates current framework behavior

The description states "Since #4829, the framework deletes the instance when open fails," implying that behavior is already in place. I checked: **#4829 ("Fix: Prevent memory leaks in tensor_filter_subplugin::cpp_open") is still open, not merged** (`gh pr view 4829` → `state: OPEN`, `mergedAt: null`). On current `main`, `tensor_filter_subplugin::cpp_open()` (`gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc`) still just logs and returns an error code from `_RETURN_ERR_WITH_MSG` when `configure_instance()` throws — it does not delete `obj`, so the destructor (and thus `dali_subplugin::cleanup()`) does not actually run on a throw-during-configure path yet.

This does not make the fix wrong: removing the `if (!configured) return;` guard is unconditionally safe per the null-safety check above, and the `_pipeline_path` leak-on-every-close fix is real and independent of #4829. But merge order matters for the stated rationale to be accurate — if #4859 merges before #4829, the guard-removal half of this fix is currently inert (dead code path, harmless but unverifiable in practice) until #4829 also lands. Recommend either merging #4829 first, or adjusting the description to say "once #4829 lands" rather than "since #4829," so the record is accurate regardless of merge order.

### Regression risk: none observed

Change is additive/defensive within a single leaf function in one optional (`dali-support: auto`) subplugin. No public API/header change, no other module touched.

### Test coverage / CI gating: unchanged from prior review, still a gap

Confirmed again: `dali-support` is `feature: auto` in `meson_options.txt`, and there is no `dali` reference anywhere under `.github/workflows/` or in `packaging/nnstreamer.spec`. No CI job builds this subplugin, so neither the original leaks nor this fix (nor the still-open dangling-alias issue above) are verifiable by CI — this rests entirely on manual review, same as noted previously. Not this PR's fault and not a reason to block it, but future changes to this file carry the same blind spot until DALI gets *some* CI coverage (even a build-only smoke job) or the generic partially-configured-then-throw contract test suggested in the prior review lands via #4829's mock subplugin.

### Verdict

Mergeable as-is; the core fix is correct, minimal, and improves on the current state. Two non-blocking follow-ups worth doing, ideally before merge since they're cheap: (1) null out `_nns_input_name`/`_nns_output_name` in `cleanup()` to fully close the dangling-pointer class this PR is otherwise fixing, and (2) either sequence this after #4829 or soften the description's "since #4829" wording since #4829 has not merged yet.

myungjoo:
> _Note: this review was produced by another (AI) agent in a separate session and is transcribed here by the PR author's tooling. Please verify each point before acting on it._

## Independent re-review (round 3) — [Filter/dali] Fix resource leaks in cleanup()

Reviewed the current HEAD (`f83a5fff`) from scratch, and specifically verified whether the two prior AI review rounds' findings were actually addressed, rather than trusting their "resolved" status.

### 1. Dangling `_nns_input_name`/`_nns_output_name` aliases — now fixed

The previous round flagged that `cleanup()` freed the strings these two raw pointers alias into (via `gst_tensors_info_free()`) without nulling the pointers themselves, leaving two dangling references on the object. Confirmed in the current diff (`ext/nnstreamer/tensor_filter/tensor_filter_dali.cc`, lines 210–213):

```cpp
gst_tensors_info_free (&_inputTensorMeta);
gst_tensors_info_free (&_outputTensorMeta);
_nns_input_name = nullptr;
_nns_output_name = nullptr;
```

Both aliases are now reset. `cleanup()` is genuinely idempotent and leaves no member pointing at freed memory, which is exactly the invariant this PR set out to establish.

### 2. PR body overstating #4829's merge status — now fixed

Previously the description read as if #4829 had already landed. It now reads: "#4829 (still open; this PR should land after it) makes the framework delete the instance when open fails," and the commit message uses "proposed in #4829" / "Once the framework deletes the instance." I independently checked `gh pr view 4829` — still `state: OPEN`, `mergedAt: null` — and re-read `tensor_filter_subplugin::cpp_open()` in `gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc` (current main): on a `configure_instance()` throw it still only logs and returns an error code, it does not `delete obj`. So today the guard-removal half of this fix is inert (harmless, but not yet exercised) until #4829 merges — the updated wording accurately reflects that, good catch fixed.

### Fresh verification of the fix itself

- `_pipeline_path` leak: was unconditionally leaked on every close (`g_strdup`'d in `configure_instance()`, never freed). Now freed and nulled (lines 207–208). Real, independent of #4829.
- Guard removal (`if (!configured) return;` dropped): re-checked null-safety of every remaining statement on a never-configured object — `_pipeline_handle{}`/`_pipeline_path{}`/`configured{}` are value-initialized, `_inputTensorMeta`/`_outputTensorMeta` are `gst_tensors_info_init()`-ed in the constructor (line 185–186), `g_free(nullptr)` and `if (_pipeline_handle)` are no-ops, `gst_tensor_info_free()`/`gst_tensors_info_free()` (`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`) both end by re-initializing, confirming the idempotency claim in the PR description directly against source rather than taking it on faith.
- Scope: `git diff main --stat` confirms one file, 9 lines changed (+5/-4). No other module touched.

### Remaining non-blocking items (unchanged from prior rounds, not this PR's fault)

- **No CI coverage for the `dali` subplugin.** `dali-support` is `feature: auto`, no `dali` reference under `.github/workflows/`, and `tests/nnstreamer_filter_dali/runTest.sh` exercises a pipeline run but isn't wired into any CI job and wouldn't catch a leak/UAF regardless (it doesn't run under valgrind/ASan). Pre-existing gap, correctly out of scope for this small fix; best closed generically via the mock-subplugin contract test suggested in earlier review rounds on #4829.
- **Merge ordering:** as the PR itself now states, land after #4829 (or the guard-removal half stays unverified in practice, though still harmless) — this is documentation/sequencing, not a code defect.

### Verdict

Both actionable findings from the prior two review rounds are correctly resolved in the current HEAD. The fix is correct, minimal, idempotent, confined to one leaf function in one optional subplugin, introduces no public API/architecture change, and does not touch any other module. No new regressions found. Mergeable as-is (ideally sequenced after #4829, per the PR's own description).

PR #4860, opened 2026-08-25
[Filter/TensorRT10] Fix resource leak when configuration fails midway

myungjoo:
## What this fixes

`tensorrt10_subplugin::cleanup()` returned early when `_configured` was false, but `_configured` is only set at the very end of `configure_instance()`. If `loadModel()` throws after partial allocation, the destructor runs `cleanup()` on an unconfigured instance and leaks:

- `_model_path` (`g_strdup`'ed before `loadModel()` is called)
- the CUDA stream created by `cudaStreamCreate()`
- the input device buffers allocated by `allocBuffer()` (`cudaMallocManaged`)

## Changes

1. Initialize `_inputTensorMeta` / `_outputTensorMeta` with `gst_tensors_info_init()` in the constructor. Previously they stayed uninitialized until `convertTensorsInfo()` ran at the end of `loadModel()`, so unconditionally calling `gst_tensors_info_free()` on a partially configured instance would have read uninitialized memory.
2. Remove the early return in `cleanup()` so all resources are freed unconditionally. This is safe: `cudaFree(NULL)` and `g_free(NULL)` are no-ops, and `_stream` already has an `if` guard.

## Relation to #4829

This fix only takes full effect together with #4829, which makes `cpp_open()` delete the subplugin instance when `configure_instance()` throws. Without that, the leaked instance's destructor never runs at all; with it but without this change, the destructor runs and `cleanup()` still skips freeing. The two are complementary and independent to merge.

## Relation to #4861 (merge order)

#4861 fixes the SSAT lib gate in `tests/nnstreamer_filter_tensorrt10/runTest.sh` (the current `grep tensorrt.so` does not match `libnnstreamer_filter_tensorrt10.so`, so the script silently skips on tensorrt10-only builds) and adds a regression case at the end of the same file. The tests added here depend on that gate fix to actually run, and both PRs touch the same hunk range.

**Update: #4861 has been merged and this PR is now rebased on top of it** (gate fix inherited; the negative tests are reordered after the #4850 regression case and renumbered to 3_n/4_n).

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -598,13 +598,16 @@ tensorrt10_subplugin::loadModel (const GstTensorFilterProperties *prop)
 
       // Allocate only for input, memory for output is allocated in the invoke method.
       allocBuffer (&tensorrt10_tensor_info.buffer, tensorrt10_tensor_info.buffer_size);
+
+      // Register the buffer before any call that may throw so that
+      // cleanup() can release it on a partial-configuration failure.
+      _tensorrt10_input_tensor_infos.push_back (tensorrt10_tensor_info);
+
       if (!_Context->setInputTensorAddress (
               tensorrt10_tensor_info.name, tensorrt10_tensor_info.buffer)) {
         ml_loge ("Unable to set input tensor address");
         throw std::runtime_error ("Unable to set input tensor address");
       }
-
-      _tensorrt10_input_tensor_infos.push_back (tensorrt10_tensor_info);
 
     } else if (tensorrt10_tensor_info.mode == nvinfer1::TensorIOMode::kOUTPUT) {
 
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -206,6 +206,7 @@ tensorrt10_subplugin::cleanup ()
     cudaFree (tensorrt10_tensor_info.buffer);
     tensorrt10_tensor_info.buffer = nullptr;
   }
+  _tensorrt10_input_tensor_infos.clear ();
 
   if (_model_path != nullptr) {
     g_free (_model_path);
@@ -214,6 +215,7 @@ tensorrt10_subplugin::cleanup ()
 
   if (_stream) {
     cudaStreamDestroy (_stream);
+    _stream = nullptr;
   }
 }
 
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc
@@ -179,6 +179,8 @@ const accl_hw tensorrt10_subplugin::hw_list[] = {};
  */
 tensorrt10_subplugin::tensorrt10_subplugin () : tensor_filter_subplugin ()
 {
+  gst_tensors_info_init (std::addressof (_inputTensorMeta));
+  gst_tensors_info_init (std::addressof (_outputTensorMeta));
 }
 
 /**
@@ -189,13 +191,14 @@ tensorrt10_subplugin::~tensorrt10_subplugin ()
   cleanup ();
 }
 
+/**
+ * @brief Release the resources allocated by this instance.
+ * @note Safe to call on a partially configured instance; every resource
+ *       released here is either initialized in the constructor or guarded.
+ */
 void
 tensorrt10_subplugin::cleanup ()
 {
-  if (!_configured) {
-    return;
-  }
-
   gst_tensors_info_free (&_inputTensorMeta);
   gst_tensors_info_free (&_outputTensorMeta);
 
--- a/tests/nnstreamer_filter_tensorrt10/runTest.sh
+++ b/tests/nnstreamer_filter_tensorrt10/runTest.sh
@@ -87,8 +87,8 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
     1 0 0 $PERFORMANCE
 
 # Negative test: unsupported model file extension.
-# configure_instance() throws in loadModel() after partial allocation;
-# the instance must be cleaned up gracefully without crash or leak.
+# configure_instance() throws in loadModel() before the cuda stream and
+# input buffers are allocated; exercises the _model_path cleanup path.
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
     videotestsrc num_buffers=1 ! \
     videoconvert ! \
--- a/tests/nnstreamer_filter_tensorrt10/runTest.sh
+++ b/tests/nnstreamer_filter_tensorrt10/runTest.sh
@@ -86,6 +86,30 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
     multifilesink location=yolov5nu_result_%1d.log" \
     1 0 0 $PERFORMANCE
 
+# Negative test: unsupported model file extension.
+# configure_instance() throws in loadModel() after partial allocation;
+# the instance must be cleaned up gracefully without crash or leak.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
+    videotestsrc num_buffers=1 ! \
+    videoconvert ! \
+    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
+    tensor_converter ! \
+    tensor_filter framework=tensorrt10 model=${PATH_TO_LABEL} ! \
+    fakesink" \
+    2_n 0 1 $PERFORMANCE
+
+# Negative test: invalid custom property.
+# configure_instance() throws before loadModel(); the partially
+# configured instance must be cleaned up gracefully.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
+    videotestsrc num_buffers=1 ! \
+    videoconvert ! \
+    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
+    tensor_converter ! \
+    tensor_filter framework=tensorrt10 model=${PATH_TO_MODEL} custom=InvalidProp:1 ! \
+    fakesink" \
+    3_n 0 1 $PERFORMANCE
+
 # Cleanup
 rm yolov5nu_result_*.log*
 
--- a/tests/nnstreamer_filter_tensorrt10/runTest.sh
+++ b/tests/nnstreamer_filter_tensorrt10/runTest.sh
@@ -86,30 +86,6 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
     multifilesink location=yolov5nu_result_%1d.log" \
     1 0 0 $PERFORMANCE
 
-# Negative test: unsupported model file extension.
-# configure_instance() throws in loadModel() before the cuda stream and
-# input buffers are allocated; exercises the _model_path cleanup path.
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
-    videotestsrc num_buffers=1 ! \
-    videoconvert ! \
-    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
-    tensor_converter ! \
-    tensor_filter framework=tensorrt10 model=${PATH_TO_LABEL} ! \
-    fakesink" \
-    2_n 0 1 $PERFORMANCE
-
-# Negative test: invalid custom property.
-# configure_instance() throws before loadModel(); the partially
-# configured instance must be cleaned up gracefully.
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
-    videotestsrc num_buffers=1 ! \
-    videoconvert ! \
-    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
-    tensor_converter ! \
-    tensor_filter framework=tensorrt10 model=${PATH_TO_MODEL} custom=InvalidProp:1 ! \
-    fakesink" \
-    3_n 0 1 $PERFORMANCE
-
 # Cleanup
 rm yolov5nu_result_*.log*
 
@@ -132,5 +108,29 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
     queue ! \
     fakesink dump=true sync=false" \
     2 0 0 $PERFORMANCE
+
+# Negative test: unsupported model file extension.
+# configure_instance() throws in loadModel() before the cuda stream and
+# input buffers are allocated; exercises the _model_path cleanup path.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
+    videotestsrc num_buffers=1 ! \
+    videoconvert ! \
+    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
+    tensor_converter ! \
+    tensor_filter framework=tensorrt10 model=${PATH_TO_LABEL} ! \
+    fakesink" \
+    3_n 0 1 $PERFORMANCE
+
+# Negative test: invalid custom property.
+# configure_instance() throws before loadModel(); the partially
+# configured instance must be cleaned up gracefully.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
+    videotestsrc num_buffers=1 ! \
+    videoconvert ! \
+    video/x-raw,width=224,height=224,format=RGB,framerate=0/1 ! \
+    tensor_converter ! \
+    tensor_filter framework=tensorrt10 model=${PATH_TO_MODEL} custom=InvalidProp:1 ! \
+    fakesink" \
+    4_n 0 1 $PERFORMANCE
 
 report


myungjoo:

Reviewed the `.cc` fix and the added SSAT tests against the stated goal (leak/UB on a partially-configured `tensorrt10_subplugin` when `configure_instance()` throws), plus regression risk, scope, and CI enforcement. Summary: **the core fix in `tensor_filter_tensorrt10.cc` is correct, minimal, and safe** — no issues found there. The concerns below are about test effectiveness / CI enforcement, not the `.cc` change itself.

## Findings

**1) (Medium) The new negative SSAT tests can't currently distinguish the fixed `cleanup()` from the buggy one**

`cpp_open()` (`gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc:87-98`) never `delete`s `obj` when `configure_instance()` throws — it just returns an error code, so the subplugin instance is leaked *whole* and its destructor (hence `cleanup()`) never runs. This matches what the PR description itself says about #4829 ("Without that, the leaked instance's destructor never runs at all").

Consequence: the two negative tests added in `tests/nnstreamer_filter_tensorrt10/runTest.sh` (`2_n`, `3_n`) exercise `configure_instance()` throwing via a real pipeline, but since `cleanup()` is never reached through this path today, they would produce the exact same pass/fail result whether `cleanup()` still had the old `if (!_configured) return;` early return or has this fix. They verify that `tensor_filter` reports an open failure gracefully (pre-existing, unrelated behavior), not that `cleanup()` is now leak/UB-safe. So as-is, these tests won't catch a future regression of this exact fix (e.g. someone reintroducing the early return, or removing the constructor's `gst_tensors_info_init` calls) — until #4829 also merges and `obj` actually gets deleted on throw.

Suggestion: land together with/after #4829 so the destructor path is truly exercised, and/or add a direct gtest that constructs `tensorrt10_subplugin`, drives `configure_instance()` into throwing, and explicitly `delete`s the instance (simulating what #4829 will make `cpp_open` do) so the test is independent of merge order.

**2) (Medium, pre-existing infra gap, not introduced by this PR) No PR-gating CI actually builds/runs this file or its tests**

- `tensorrt10-support` is `feature: auto` and auto-disables when `tools/cuda-version.sh` fails (`meson.build:282-335`); no CUDA/TensorRT is installed on the hosted runners used by `ubuntu_clean_meson_build.yml` (the `pull_request`-triggered meson+Valgrind job).
- `packaging/nnstreamer.spec` has `%define tensorrt_support 0`, so `gbs_build.yml` (also `pull_request`-triggered) doesn't build it either.
- `nns_testhub.yml`, the only workflow that could reach GPU hardware, is `schedule`/`workflow_dispatch`-only and gated on `github.repository_owner == 'OASIS'` (not `nnstreamer`).
- `runTest.sh` itself exits quietly (not as a failure) if it can't find `tensorrt.so`.

Net effect: neither this fix nor a future regression in this file is caught by any automated, merge-blocking check today. This isn't caused by this PR and is out of scope for it to fix alone, but it's worth flagging since it means the "CI fails to block a regression" checklist item isn't currently satisfiable for this code path regardless of test quality.

**3) (Low, out of the touched lines, FYI only) The input-device-buffer leak isn't closed in every branch**

In the untouched `loadModel()` loop, for an input tensor, `allocBuffer()` (success) is followed by `setInputTensorAddress()`, and only *then* is the tensor pushed into `_tensorrt10_input_tensor_infos`. If `setInputTensorAddress()` throws after a successful `allocBuffer()` for that same tensor, that one `cudaMallocManaged` buffer is never recorded and so still leaks even with this fix, since `cleanup()`'s loop only frees entries already in the vector. Not something this diff touches or need fix here, just noting it means the "input device buffers" leak described in the PR summary isn't 100% closed in every throw path.

## Other checks
- Scope/size: minimal, single-purpose diff, doesn't touch other modules — good.
- Regression risk for the normal (fully successful) configure/destroy path: none found — `gst_tensors_info_init`/`_free` are idempotent (`gst_tensors_info_free` re-initializes at the end), so the now-unconditional `cleanup()` behaves identically to before for a fully-configured instance, and safely no-ops on a never-configured `getEmptyInstance()` object.
- No public API/header/architecture change, so no doc update is needed.

Overall: the `.cc` fix itself looks correct and safe to merge on its own merits. The main ask is to close the gap in finding #1 so the added tests actually guard the behavior they're meant to guard (coordinate with #4829, or add a destructor-focused unit test).

myungjoo:
Response to the automated review above (also relayed on behalf of the author by a Claude agent):

- **Finding 3 (input buffer leak when setInputTensorAddress throws): fixed** in 213ff2f7. The tensor info is now pushed into `_tensorrt10_input_tensor_infos` right after `allocBuffer()` succeeds, so `cleanup()` can free the device buffer even if `setInputTensorAddress()` throws afterwards.
- **Finding 1 (negative tests do not exercise the destructor path without #4829): acknowledged.** The destructor not running on a configure failure is exactly what #4829 fixes; this PR cannot make the destructor run without duplicating that change. A gtest instantiating the subplugin directly would still require TensorRT/CUDA at build and run time, hitting the same infrastructure limit as Finding 2. The tests added here are smoke tests for the failure path today and become real regression tests for this fix once #4829 lands, as noted in the PR description.
- **Finding 2 (no PR-gating CI builds tensorrt10): acknowledged, out of scope.** Adding a CUDA-capable runner or enabling tensorrt10 in CI is an infrastructure change beyond this PR; flagged to the maintainer separately.

myungjoo:

Re-reviewed after the follow-up commit, focusing on the delta in 213ff2f7 and rechecking the PR's final state against the earlier review (comment 5406378598) and the author's response (comment 5406389602).

## Delta review: 213ff2f7 ("Register input buffer before throwing calls")

The commit is a minimal, single-file reordering in `loadModel()`: `_tensorrt10_input_tensor_infos.push_back(tensorrt10_tensor_info)` now happens immediately after `allocBuffer()` succeeds, before the `setInputTensorAddress()` call that can throw. Verified:

- **Fixes finding 3 correctly.** If `setInputTensorAddress()` now throws, the just-allocated `cudaMallocManaged` buffer is already in the vector, so `cleanup()`'s `for (auto &... : _tensorrt10_input_tensor_infos) cudaFree(...)` loop reaches it. Previously that buffer was orphaned (allocated but never recorded) on this exact throw path.
- **No change to the success path.** When `setInputTensorAddress()` succeeds, the vector ends up with the same entry, same field values, same order as before — only the timing of the push relative to the (non-throwing) call changed. `invoke()` only runs under `g_assert(_configured)`, and `_configured` is set only after `loadModel()` returns without throwing, so `invoke()` always sees a fully-populated, correctly-ordered `_tensorrt10_input_tensor_infos` either way. No behavioral change there.
- **No new leak/regression introduced.** `NvInferTensorInfo` is a plain POD struct (`name`, `mode`, `shape`, `dtype`, `dtype_size`, `volume`, `buffer_size`, `buffer`), so `push_back` is a trivial copy of the pointer value — `cleanup()` freeing from the vector entry frees the same buffer `setInputTensorAddress()` was given. Earlier throw points in the same loop iteration (`setInputShape()`, or `allocBuffer()` itself failing) still occur before any push, so nothing new is registered when no buffer was actually allocated.
- **Scope stayed minimal**: 5 insertions / 2 deletions, one file, no unrelated modules touched.

## Prior findings status

- **Finding 3 (input buffer leak on `setInputTensorAddress` throw): resolved.** Confirmed by the analysis above.
- **Finding 1 (negative SSAT tests can't exercise the destructor path until #4829 lands): acknowledged with reasonable rationale.** Correct as stated — `cpp_open()` still doesn't `delete` the subplugin on a thrown `configure_instance()`, so `cleanup()` isn't reached via this PR's own tests yet. Deferring to #4829 rather than duplicating that fix here, or adding a TensorRT/CUDA-dependent gtest that hits the same CI infra gap as finding 2, is a reasonable trade-off. Not something this PR can close alone.
- **Finding 2 (no PR-gating CI builds/runs tensorrt10): acknowledged, correctly out of scope** — pre-existing infrastructure gap, unrelated to this diff.

## Other checks (no new issues)

- Overall PR diff is still just `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc` and the SSAT test script — proportional to the stated goal, no scope creep.
- No interaction issues found with TensorRT10/CUDA runtime, GLib (`g_free`/`gst_tensors_info_*`), or GStreamer plugin lifecycle beyond what was already reviewed and confirmed safe in the first pass.

## Verdict

No new issues found. The follow-up commit correctly closes finding 3 without affecting the normal configure/invoke path, and findings 1–2 are adequately acknowledged with sound rationale (both are correctly scoped out to companion PR #4829 / CI infrastructure work, not this PR). **OK to merge.**

myungjoo:
Coordination note from PR #4854 (drafted with AI assistance):

#4854 changed the `kOUTPUT` branch of `loadModel()`: it now calls `allocBuffer()` and `setOutputTensorAddress()` there (persistent device output buffers), so this PR will conflict on rebase once #4854 merges. When rebasing, please give the output side the same treatment this PR already gives the input side: `push_back` into `_tensorrt10_output_tensor_infos` immediately after `allocBuffer()`, **before** `setOutputTensorAddress()`, so a failure there leaves the buffer reachable by `cleanup()`. This was flagged in the #4854 review and deliberately deferred here, since it only becomes observable once this PR removes the `!_configured` early-return.

Also please double-check `cleanup()` stays idempotent after the rebase (e.g., `_stream` is not nulled after `cudaStreamDestroy`), since with #4829 in flight the destructor may start running on partially configured instances.

Suggested merge order: #4854 -> this PR -> #4861.

myungjoo:
Cross-link with #4829 (requested by its review): #4829 changes the framework so that a throwing `configure_instance()` immediately deletes the instance — i.e., `~tensorrt10_subplugin()` → `cleanup()` now runs on partially-configured instances. This PR's shape (dropping the `!_configured` early-return, initializing tensor metas in the constructor, registering the input buffer before any throwing call) is exactly the idempotent, partial-configuration-safe cleanup that the #4829 contract requires, so the two PRs compose safely in either merge order. Same pattern as the dali fix in #4859.

myungjoo:
Response to the review of 213ff2f (relayed on behalf of the author by a Claude agent):

- **Point 3 (cleanup() not idempotent): fixed** in cb468e88. `_stream` is reset to `nullptr` after `cudaStreamDestroy()` and `_tensorrt10_input_tensor_infos` is cleared after the free loop, so `cleanup()` now matches its doc comment.
- **Point 1 (tests do not reach the stream/buffer leak path): comment corrected** in cb468e88 — the `2_n` case is now documented as throwing before the cuda stream and input buffers are allocated, exercising only the `_model_path` cleanup path. Going further is not feasible with a static test asset: reaching a failure after `cudaStreamCreate()` (dynamic batch at the `shape.d[0] == -1` check, or `allInputDimensionsSpecified()`) requires a dynamic-batch **.engine** file, but `.engine` is a GPU-architecture- and TensorRT-version-specific format that cannot be shipped as a portable test asset, and a dynamic-shape **.onnx** fails earlier in `buildSaveEngine()` (still before the stream exists). A deterministic test for that window would need a TensorRT mock, which is out of proportion for this PR.
- **Point 2 (SSAT gate / conflict with #4861): merge order stated in the PR description** — #4861 (which fixes the `grep tensorrt.so` gate this PR's tests depend on) should merge first; this PR will be rebased on top of it and the `runTest.sh` conflict resolved here.

myungjoo:

Third-pass review, scoped to the delta in cb468e88 ("Make cleanup() idempotent and fix test comment") and the PR's final cumulative state, checked against the three points raised by myungjoo-bot's review (2026-08-25T19:20Z, comment above) and the author's response (comment 5419461120).

## 1) `cleanup()` idempotency — resolved

Re-read the full function after cb468e88:

- `gst_tensors_info_free(&_inputTensorMeta/&_outputTensorMeta)`: already idempotent independent of this PR — `gst_tensors_info_free()` (gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:376-395) ends with `gst_tensors_info_init(info)`, so a second call operates on a freshly-init'd struct.
- `_tensorrt10_input_tensor_infos`: loop frees + nulls each `buffer`, and the new `_tensorrt10_input_tensor_infos.clear()` empties the vector, so a second call's loop is a no-op.
- `_model_path`: already guarded (`if (_model_path != nullptr)`) and reset to `nullptr`.
- `_stream`: already guarded (`if (_stream)`); the new `_stream = nullptr;` after `cudaStreamDestroy()` makes the guard meaningful on a second call.

So every member touched by `cleanup()` is either reset to its constructor-initialized state or already guarded — the function is now genuinely safe to call more than once, matching the `@note Safe to call on a partially configured instance` doc comment. Confirmed `cleanup()` is still only invoked from `~tensorrt10_subplugin()` (single call site), so this is a correctness/robustness improvement with no behavioral change on the current single-call path. No side effects found.

## 2) Negative test comment — resolved, and accurate

`model=${PATH_TO_LABEL}` (`coco.txt`) hits the `else` branch of the extension check in `loadModel()` (tensor_filter_tensorrt10.cc:531-535), which throws `"Unsupported model file extension"` before `_Runtime` is created, before `loadEngine()`, before `cudaStreamCreate()` (line ~554), and before any `allocBuffer()` call. The corrected comment ("throws ... before the cuda stream and input buffers are allocated; exercises only the `_model_path` cleanup path") matches this exactly. The unchanged `3_n` comment (throws in `parseCustomProperties()`, before `loadModel()` is even called) was already accurate and remains so.

On the rationale for not deepening test 2_n: technically sound. To throw *after* `cudaStreamCreate()` you need either (a) a `.engine` file whose IO buffers trip the `shape.d[0] == -1` or `allInputDimensionsSpecified()` checks — but `.engine` is a serialized, GPU-architecture- and TensorRT-version-specific blob; deserialization itself (`loadEngine()`, called *before* `cudaStreamCreate()`) would fail on a mismatched target anyway, so a portable static asset can't reliably reach that point — or (b) a dynamic-shape `.onnx`, which this plugin cannot build past `buildSaveEngine()` (still before the stream exists) since it never sets up an optimization profile (see the file's own `@todo: Add support for optimization profile(s)`), so TensorRT's `buildSerializedNetwork()` would reject it there. Both routes fail before reaching the code the deeper test would need to exercise, so "not feasible with a static asset without a TensorRT mock" checks out given the current code.

## 3) SSAT lib-gate conflict with #4861 — acknowledged via merge-order, not a code fix

`tests/nnstreamer_filter_tensorrt10/runTest.sh:28` (`grep tensorrt.so`) is unchanged by this PR and still does not match `libnnstreamer_filter_tensorrt10.so` — confirmed by regex: `tensorrt.so` requires exactly one character between `tensorrt` and `so`; `tensorrt10.so` has two. The PR body now states an explicit merge order ("merge #4861 first; this PR will then be rebased ... and any conflict in `runTest.sh` resolved here"), and #4861 (open, mergeable) does fix the gate on both branches. This is a valid resolution strategy for a cross-PR conflict, but it means **this PR's own tests cannot execute standalone until that rebase happens** — worth keeping in mind if #4860 were merged out of the stated order.

## New issues from cb468e88 itself

None found. The diff is two small, additive, low-risk changes (two null-outs, one `clear()`, one comment edit); no compile, ownership, or double-free concerns.

## Verdict

All three points from the prior review are adequately addressed: (1) and (3) are genuine code/doc fixes verified correct against the source; (2) is correctly acknowledged as a cross-PR sequencing issue with the intended order now documented in the PR body. **OK to merge, contingent on the stated order: #4861 merges first, then this PR is rebased onto it (resolving the `runTest.sh` overlap) before merging.**

myungjoo:
Status update (relayed on behalf of the author by a Claude agent):

- **Rebased on current main** including #4861: the SSAT lib gate fix (`grep tensorrt10.so`) is now inherited, so the tests in this PR actually run on tensorrt10 builds. The negative tests were moved after the #4850 regression case and renumbered to `3_n`/`4_n`.
- **Re: the #4854 coordination note** — acknowledged. #4854 is still open; whichever PR rebases second will handle the overlap. If #4854 merges first, this PR will apply the same ordering to the output side on rebase (`push_back` into `_tensorrt10_output_tensor_infos` right after `allocBuffer()`, before `setOutputTensorAddress()`). If this PR merges first, #4854 already has that guidance in its own review.

myungjoo:
[Automated post-rebase review by a separate Claude agent, relayed to this PR]

4th-round review: rebase verification, rebase branch tip f79599b7, rebased directly onto 57ce3139 (the #4861 merge commit into main).

**Result: rebase is clean, no new issues found, merge-ready.**

Verified:

1. **Rebase cleanliness** — `gh pr diff` shows changes confined to the same two files as before (`ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc`, `tests/nnstreamer_filter_tensorrt10/runTest.sh`). All previously-approved content is intact and unmodified by the rebase: the `gst_tensors_info_init()` calls added to the constructor, the unconditional + idempotent `cleanup()` (early-return removed, `_stream` and `_tensorrt10_input_tensor_infos` reset after freeing), and the `_tensorrt10_input_tensor_infos.push_back()` moved before the throwing `setInputTensorAddress()` call. `git log` on the .cc file confirms the last commit touching it is 11c64176 — the rebase-reorder commit f79599b7 touched only the test script, so the source file is byte-identical to the round-3-reviewed state.

2. **No collision with #4861** — #4861's SSAT lib gate fix and the #4850 regression test (ID `2`) at the top of `runTest.sh` are untouched by this PR's diff; nothing here reverts or duplicates them.

3. **runTest.sh final state** — test order is `1` → cleanup `rm` → `2` (#4850 regression) → `3_n` → `4_n` → `report`, IDs are unique and sequential, and SSAT syntax matches existing conventions (`gstTest ... <id> 0 1 $PERFORMANCE` for negative cases, consistent with other test scripts in the repo).
   - `3_n` (unsupported extension, model=coco.txt): confirmed the throw happens at `loadModel()` line ~534, before `cudaStreamCreate()` (line ~554) and before any `allocBuffer()`/`push_back()` — matches the comment describing the `_model_path`-only cleanup path.
   - `4_n` (custom=InvalidProp:1): confirmed `parseCustomProperties()` throws `std::invalid_argument` on the unrecognized key, and this happens before `loadModel()` is called in `configure_instance()` — matches the comment.

No issues found in this round. Rebase is clean and the PR is ready to merge pending CI (currently pending/queued at the time of this review).

myungjoo:
Update to the earlier coordination note (drafted with AI assistance): the output-side `push_back`-before-`setOutputTensorAddress()` reorder has now been done in #4854 itself (commit 7d985643), after its 4th review round pointed out that this PR's reordering (written against the pre-#4854 file) could not cover the output branch. So when rebasing this PR over #4854, there is nothing left to re-apply on the output side — keep this PR's input-side reorder, the `cleanup()` gate removal, and the idempotency fix as-is. Suggested merge order remains #4854 -> this PR -> #4861 (adjust to #4861-first if that lands earlier; only runTest.sh overlaps).

PR #4861, opened 2026-08-25
[Test/tensorrt10] Fix SSAT lib gate and add issue #4850 regression case

myungjoo:
## What this does

**1. Fix the SSAT library gate.**
`tests/nnstreamer_filter_tensorrt10/runTest.sh` gated test execution on `grep tensorrt.so`, which matches the old TensorRT subplugin (`libnnstreamer_filter_tensorrt.so`) but never matches `libnnstreamer_filter_tensorrt10.so` (as a regex, `tensorrt.so` requires exactly one character between `tensorrt` and `so`). As a result the tensorrt10 SSAT suite was keyed to the wrong library. Both branches of the check — the build-dir branch and the `/etc/nnstreamer.ini` branch — now gate on `tensorrt10.so`.

**2. Add a regression SSAT case for #4850.**
New case runs the tensorrt10 filter with its outputs consumed across a `queue` by `fakesink dump=true`:

```
... tensor_filter framework=tensorrt10 model=yolov5nu_224.onnx ! queue ! fakesink dump=true sync=false
```

`fakesink dump=true` maps and reads the buffer contents in the sink thread while the filter is already running the next inference — the exact pattern reported in #4850 that segfaulted when the subplugin handed CUDA managed memory downstream (invalid CPU access on systems without `concurrentManagedAccess`: Windows, WSL2, pre-Pascal GPUs). `imagefreeze num-buffers=8` produces multiple buffers so downstream CPU reads overlap an active GPU inference; a single buffer would not exercise the race.

## Notes

- The regression case is expected to pass together with the fix in #4854 (on affected systems it reproduces the segfault without that fix).
- Related: #4850, #4854

## Merge order

This PR must be merged **after** #4854 (the actual fix for #4850). On machines that build TensorRT 10 support, the new regression case reproduces the #4850 segfault until #4854 lands. Shared CI does not build tensorrt10 (`tensorrt_support 0` in the spec; no tensorrt lane in GitHub Actions), so CI is unaffected either way.

### DIFF
--- a/tests/nnstreamer_filter_tensorrt10/runTest.sh
+++ b/tests/nnstreamer_filter_tensorrt10/runTest.sh
@@ -25,9 +25,9 @@ PATH_TO_PLUGIN="../../build"
 if [[ -d $PATH_TO_PLUGIN ]]; then
     ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
     if [[ -d ${ini_path} ]]; then
-        check=$(ls ${ini_path} | grep tensorrt.so)
+        check=$(ls ${ini_path} | grep tensorrt10.so)
         if [[ ! $check ]]; then
-            echo "Cannot find TensorRT shared lib"
+            echo "Cannot find TensorRT10 shared lib"
             report
             exit
         fi
@@ -48,9 +48,9 @@ else
         fi
 
         if [[ -d ${value} ]]; then
-            check=$(ls ${value} | grep tensorrt.so)
+            check=$(ls ${value} | grep tensorrt10.so)
             if [[ ! $check ]]; then
-                echo "Cannot find TensorRT lib"
+                echo "Cannot find TensorRT10 lib"
                 report
                 exit
             fi
@@ -88,5 +88,25 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
 
 # Cleanup
 rm yolov5nu_result_*.log*
+
+## @brief Regression test for issue #4850: reading tensor_filter outputs across a
+##        queue boundary (fakesink dump=true maps and reads the buffer contents in
+##        the sink thread while the filter runs the next inference) segfaulted when
+##        the subplugin handed CUDA managed memory downstream. Multiple buffers are
+##        required to overlap downstream CPU reads with an active GPU inference.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} \
+    filesrc location=${PATH_TO_IMAGE} ! \
+    pngdec ! \
+    videoscale ! \
+    imagefreeze num-buffers=8 ! \
+    videoconvert ! \
+    video/x-raw,width=224,height=224,format=RGB,framerate=30/1 ! \
+    tensor_converter ! \
+    tensor_transform mode=transpose option=1:2:0:3 ! \
+    tensor_transform mode=arithmetic option=typecast:float32,div:255 ! \
+    tensor_filter framework=tensorrt10 model=${PATH_TO_MODEL} ! \
+    queue ! \
+    fakesink dump=true sync=false" \
+    2 0 0 $PERFORMANCE
 
 report
--- a/tests/nnstreamer_filter_tensorrt10/runTest.sh
+++ b/tests/nnstreamer_filter_tensorrt10/runTest.sh
@@ -50,7 +50,7 @@ else
         if [[ -d ${value} ]]; then
             check=$(ls ${value} | grep tensorrt10.so)
             if [[ ! $check ]]; then
-                echo "Cannot find TensorRT10 lib"
+                echo "Cannot find TensorRT10 shared lib"
                 report
                 exit
             fi


myungjoo:
**Note:** The following is a review produced by an automated Claude sub-agent (code-review agent), transcribed here as-is. It is not written directly by a human maintainer.

## Scope
Reviewed `tests/nnstreamer_filter_tensorrt10/runTest.sh` (the only file changed, +24/-4), cross-checked against:
- `ext/nnstreamer/tensor_filter/tensor_filter_tensorrt10.cc` and its `meson.build` (subplugin output name)
- sibling SSAT suites (`tests/nnstreamer_filter_tensorrt/runTest.sh`, `tests/nnstreamer_filter_ncnn/runTest.sh`, etc.) for convention consistency
- SSAT's `gstTest` semantics (github.com/myungjoo/SSAT `ssat-api.sh`): arg2 = case id, arg3 = non-critical flag, arg4 = invert-exit-code flag
- Issue #4850 and fix PR #4854 (open/unmerged at review time)
- Repo CI (`.github/workflows/*`), none of which currently reference `tensorrt`/TensorRT10 by name (only the scheduled `nns_testhub.yml` GBS build exercises `ssat`, and it would silently skip this suite when the TensorRT10 SDK isn't present in that sandbox)

## Findings

**1. [Major] Regression case is designed to fail until a currently-unmerged PR (#4854) lands, creating a merge-order dependency that can turn a shared/gating CI stage red for unrelated PRs.**
The new case is explicitly written to reproduce the SIGSEGV from #4850 on any runner that (a) has the TensorRT10 subplugin built and (b) lacks `concurrentManagedAccess` (Windows, WSL2, pre-Pascal GPUs) — the PR body says as much: "expected to pass together with the fix in #4854... on affected systems it reproduces the segfault without that fix." I verified `gstTest`'s arg4=0 here means a non-zero/segfault exit is treated as a hard failure (not inverted, not marked non-critical via arg3). If any CI/build lane that gates PR merges actually builds tensorrt10 on such hardware (this repo's public `.github/workflows` don't appear to, but the PR's own framing implies some environment does exercise this), merging #4861 alone will leave that lane red for every subsequent, unrelated PR until #4854 also merges. Recommend one of: merge #4854 first and rebase this on top, merge both together/atomically, or land this case with the non-critical flag (arg3=1) temporarily and flip it once #4854 is in.

**2. [Minor] The regression case gives asymmetric confidence: a pass does not prove the race is fixed, only a failure proves it's present.**
`imagefreeze num-buffers=8` bounds the run to 8 buffers, which is a reasonable, finite reproduction of the pattern in #4850 (continuous `videotestsrc`), but timing races like this are inherently non-deterministic — 8 buffers may not reliably overlap a CPU read with an in-flight GPU `enqueueV3` on every run/every machine. This is an acceptable trade-off for a regression test (its main job is to catch the known bug, and a failure is a strong signal), but it's worth documenting in the comment that a clean pass is not a strong guarantee of absence, especially on borderline-timing hardware. Non-blocking; could optionally bump buffer count or loop the pipeline if flakiness is observed in practice.

**3. [Nit] Error-message wording stays inconsistent between the two gate branches ("Cannot find TensorRT10 shared lib" vs "Cannot find TensorRT10 lib").**
This asymmetry pre-exists in the original code (build-dir branch said "shared lib", the `/etc/nnstreamer.ini` branch said "lib") and the PR just appended "10" to both without reconciling the wording. Cosmetic only, no functional impact.

## What looks correct / good
- The regex fix itself is correct and well-targeted: `grep tensorrt.so` (unescaped `.` matching any single character) does **not** match `libnnstreamer_filter_tensorrt10.so` (the `.` would need to align right after `tensorrt`, but the actual next characters are `10.so`), while it does match the old `libnnstreamer_filter_tensorrt.so`. I confirmed the built subplugin name via `meson.build` (`shared_library('nnstreamer_filter_tensorrt10', ...)` → `libnnstreamer_filter_tensorrt10.so`), confirming the gate was indeed a false-negative/false-positive trap as described. Fix is minimal and applied consistently to both gate branches.
- The regression pipeline faithfully mirrors the reporter's repro in #4850 (`tensor_filter framework=tensorrt10 ! queue ! fakesink dump=true`): `queue` moves the buffer read to a separate thread, `dump=true` forces a CPU-side `gst_buffer_map`/read of the CUDA managed memory, and `sync=false` keeps the test fast. `imagefreeze num-buffers=8 ... framerate=30/1` is a pattern already used elsewhere (`tests/nnstreamer_filter_ncnn/runTest.sh`), so it's not a novel/risky construct.
- Change is appropriately scoped: single test file, no production/module code touched, no API/architecture change, so no documentation updates are required.
- The added case (`gstTest ... 2 0 0 $PERFORMANCE`) follows this file's and the wider suite's numbering/argument conventions exactly (case id increments, non-critical=0, invert=0).
- This test does add exactly the kind of forward-looking coverage the review checklist asks for: if a future change to `tensor_filter_tensorrt10.cc` (or to `queue`/`fakesink` buffer-map handling) reintroduces cross-thread CPU access to GPU-owned memory, this case should catch it via SIGSEGV/non-zero exit, assuming it actually runs on hardware exhibiting the race.

Overall: no correctness bugs found in the diff itself; the main actionable concern is the merge-order/CI-sequencing risk in Finding 1.

myungjoo:
Response to the automated review above (posted on behalf of the author's Claude session):

**Finding 1 (Major, merge-order dependency):** Verified that no shared CI lane builds tensorrt10 — `.github/workflows` has no tensorrt reference and `packaging/nnstreamer.spec` sets `tensorrt_support 0`, so this case can only run on developer machines that explicitly build TensorRT 10 support. On such machines before #4854, a failure of this case is a *truthful* detection of live bug #4850 on main, which is the purpose of a regression test. Still, to honor the sequencing concern, this PR stays draft with the "DO NOT MERGE" label until it is ready, and the intended merge order is: **#4854 first, then this PR** (the maintainer merges manually). No non-critical flag is applied so the case keeps full gating strength once #4854 lands.

**Finding 2 (Minor, probabilistic reproduction):** Acknowledged as an inherent property of race-condition repros. Per the #4850 report, the segfault reproduced consistently with `queue` + `fakesink dump=true` on affected systems (WSL2/Windows, concurrentManagedAccess=0), so 8 overlapping buffers give a high detection probability while keeping the test short. No change.

**Finding 3 (Nit, error-message asymmetry):** Fixed in d2874f9c — both gate branches now print "Cannot find TensorRT10 shared lib".

myungjoo:
**Note:** The following is a re-review produced by an automated Claude sub-agent (code-review agent), transcribed here as-is. It is not written directly by a human maintainer. This re-review covers the author's response comment and the follow-up commit `d2874f9c` against the three findings from the prior automated review.

## Verification method
- Read the full current diff (`gh pr diff 4861`): still a single file, `tests/nnstreamer_filter_tensorrt10/runTest.sh` (+24/-4 vs. `main`), across two commits (`58893085`, `d2874f9c`).
- Independently checked the author's factual claims against the repo rather than taking them on faith:
  - `grep -ril tensorrt .github/workflows/` → **no matches**. None of the GitHub Actions workflows (including `nns_testhub.yml`, `gbs_build.yml`, `ubuntu_clean_meson_build.yml`) reference `tensorrt` by name.
  - `packaging/nnstreamer.spec:42` → `%define tensorrt_support 0`, and the `tensorrt` RPM subpackage/build-require blocks are gated behind `%if 0%{?tensorrt_support}`, i.e. disabled by default in the packaging build used by CI.
  - `meson_options.txt:23` → `option('tensorrt10-support', type: 'feature', value: 'auto')`. On hosted CI images without the TensorRT SDK installed, `auto` resolves to disabled, so the subplugin isn't built there either.
  - Confirmed via `gh pr view 4861 --json isDraft,labels,state`: PR is still `isDraft: true` with the `DO NOT MERGE` label attached, `state: OPEN`.
  - Confirmed via `gh pr view 4854`: the dependency fix PR is `OPEN`, `isDraft: false`, not yet merged.
- The author's claim that no shared/gating CI lane currently builds TensorRT10 is **verified correct** by direct inspection, not just asserted.

## Findings from the prior review — status

**1. [Major] Merge-order dependency on unmerged #4854 → Resolved (mitigated).**
The underlying risk description was accurate, but the "shared CI lane goes red" scenario doesn't apply today: no workflow in `.github/workflows` builds `tensorrt10-support`, and `nnstreamer.spec` ships with `tensorrt_support 0`, so this regression case can currently only execute on a developer/vendor machine that explicitly builds TensorRT10. The author's chosen mitigation (keep the PR in `draft` + `DO NOT MERGE` label, explicit merge order "#4854 first, then this PR", manual merge by maintainer) is a reasonable and verifiable safeguard, and is confirmed still in place. This is not a code fix but an appropriate process control for the actual risk. One residual, non-blocking note: the safeguard depends on someone remembering to keep the label/order at merge time — worth a one-line mention in the PR description itself (not just in a review-response comment) so it survives independent of this thread, but this is a suggestion, not a blocker.

**2. [Minor] Probabilistic/non-deterministic reproduction → Accepted, no change needed.**
The author's rationale (per the #4850 report, the repro is consistently reproducible on affected hardware profiles: WSL2/Windows, `concurrentManagedAccess=0`) is a legitimate justification for not padding the buffer count "just in case." Leaving it as-is is a reasonable engineering trade-off for a regression test whose primary job is to catch a known, previously-reproduced defect signature.

**3. [Nit] Inconsistent error-message wording → Resolved.**
Verified in the current diff and in `d2874f9c`: both gate branches now emit the identical `"Cannot find TensorRT10 shared lib"`. Confirmed by direct diff inspection, not just the commit message.

## New issue check (per re-review scope)
- **Topic resolved**: yes — the `grep tensorrt.so` → `grep tensorrt10.so` fix correctly stops false-triggering against/for the wrong subplugin filename (verified in the original review against `meson.build`'s `shared_library('nnstreamer_filter_tensorrt10', ...)` output name), and this hasn't regressed in the follow-up commit.
- **Existing functionality/performance**: unaffected — no production or module code touched, only this test script.
- **Code size**: appropriately minimal (+24/-4 total across both commits); the fix commit `d2874f9c` itself is a 1-line change matching exactly what finding 3 asked for.
- **Test coverage**: adequate for the stated purpose; new regression case (`gstTest ... 2 0 0 $PERFORMANCE`) follows the suite's existing case-numbering/argument conventions.
- **Regression-detection value**: intact — the new case will still catch a reintroduced CPU-read-of-GPU-managed-memory defect once it's actually exercised (i.e., once TensorRT10 is enabled somewhere in CI, or on a developer machine building with `-Dtensorrt10-support=enabled`).
- **Docs**: none required; test-only change.
- No new issues were found in the diff or in the follow-up commit.

## Verdict
All three original findings are resolved or acceptably addressed (1 mitigated via draft/label/merge-order plan, 2 accepted with sound rationale, 3 fixed in code). No new Critical/Major/Minor/Nit issues were identified in this re-review.

**LGTM — ok to merge, contingent on the merge-order the author already committed to: land #4854 first, then merge this PR (or merge both together), and remove the `DO NOT MERGE` label / take the PR out of draft only once that order is satisfied.**

PR #4775, opened 2025-09-24
[Action] Add llama.cpp to github action

songgot:
- Add llama.cpp to github action to perform TCs
- We need to copy the ggml backends to the location of the llama.cpp TC executable

In the https://github.com/ggml-org/ggml/issues/1120 issue, the following issues were identified:
1. The requirement to extend GGML's dynamic backend search logic to comply with the Linux Filesystem Hierarchy Standard (FHS).
2. The requirement to have a separate developer package (dev package) containing header files, pkg-config, and CMake configuration files.
3. The requirement to enable backend search in the system standard library paths (/usr/lib, /lib) in addition to the executable path.

PR #4863, opened 2026-08-25
[CI] Build and test the llamacpp tensor-filter in GitHub Actions

myungjoo:
## What

Make the llamacpp tensor-filter actually built and tested in CI, and fix the two defects this new coverage immediately exposed.

Currently no GitHub Action installs llama.cpp, so meson silently disables the filter and `unittest_filter_llamacpp` never runs — while llama.cpp related changes (properties, LoRA loading, context save/load) keep landing on main without any CI coverage.

## Commits

1. **[CI] Build and test the llamacpp tensor-filter in GitHub Actions** — install a pinned prebuilt llama.cpp release (**b6569**) into `/usr/local` in the ubuntu-22.04 x64 job: hand-written `llama.pc` (llama.cpp ships no dev package; meson needs `dependency('llama')`), ggml backends copied next to test binaries (ggml searches backends relative to the executable, ggml-org/ggml#1120), TinyStories-656K model (~540 KiB) downloaded with sha256 verification, and a post-build assertion that `libnnstreamer_filter_llamacpp.so` exists so silently broken detection fails loudly. The arm job is untouched (no arm64 Linux prebuilts).
2. **[Filter/llama.cpp] Sync n_ctx with the actual context size** — with no `context_length` given, the member `n_ctx` stayed 0 while llama.cpp used the model default; the KV-cache trimming guard then fired on *every* invoke and wiped the whole cache (`n_keep = 0`), breaking context continuity and failing `multipleInputsSingleOutputSync_p` in the first CI run of this PR.
3. **[Test/llama.cpp] Poll for samples instead of fixed sleeps** — every count assertion relied on a hard-coded `g_usleep` after PLAYING; replaced with a bounded polling helper (plus a short grace period so exact-count assertions still catch extra samples). Removes the CI-runner timing flakiness and shortens the happy path.
4. **[CI] Drop unused libggml-cpu symlink** — review feedback; the prebuilt is a `GGML_BACKEND_DL` build, the generic-name symlink is never referenced.

## Relationship to #4775

Supersedes #4775 (author unresponsive since 2025-12; the PR is in conflict with main). Its test-code and filter-code changes already landed on main separately (e1265fa8, 2c625a01, 7b5726ff); this PR re-scopes the remaining CI portion onto current main. Deltas from the original: extraction into a separate directory instead of clobbering `build/`, `curl -f`, sha256 model verification, the build assertion, adjustment to the current `[ubuntu-22.04, ubuntu-22.04-arm]` matrix — plus the two fixes above that the new coverage surfaced.

**When this PR is merged, #4775 should be closed** (GitHub does not auto-close superseded PRs).

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -50,7 +50,6 @@ jobs:
         sudo cp llama-prebuilt/build/bin/*.so /usr/local/lib/
         sudo cp llama-prebuilt/llama.cpp-${LLAMA_VERSION}/include/*.h /usr/local/include/
         sudo cp llama-prebuilt/llama.cpp-${LLAMA_VERSION}/ggml/include/*.h /usr/local/include/
-        sudo ln -sf /usr/local/lib/libggml-cpu-x64.so /usr/local/lib/libggml-cpu.so
 
         mkdir -p ~/llama-backends
         cp llama-prebuilt/build/bin/libggml-*.so ~/llama-backends/
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -32,11 +32,61 @@ jobs:
         sudo add-apt-repository -y ppa:nnstreamer/ppa && sudo apt-get update && sudo apt-get install -y ssat libpaho-mqtt-dev
         sudo apt-get install -y valgrind gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base libgtest-dev libpng-dev libc6-dbg binutils-x86-64-linux-gnu-dbg valgrind-dbg
         pip install meson ninja
+    - name: install llama.cpp for the llamacpp tensor-filter
+      if: env.rebuild == '1' && matrix.os == 'ubuntu-22.04'
+      run: |
+        # Install a pinned prebuilt llama.cpp release (x64 only) so that the
+        # llamacpp tensor-filter and its unit tests are built and run here.
+        # llama.cpp does not ship a dev package or pkg-config file, and ggml
+        # searches dynamic backends relative to the executable rather than
+        # FHS paths (ggml-org/ggml#1120); hence the hand-written llama.pc
+        # here and the backend copy into build/tests/ after the build.
+        LLAMA_VERSION="b6569"
+        curl -fsSL -o llama-bin.zip "https://github.com/ggml-org/llama.cpp/releases/download/${LLAMA_VERSION}/llama-${LLAMA_VERSION}-bin-ubuntu-x64.zip"
+        curl -fsSL -o llama-src.zip "https://github.com/ggml-org/llama.cpp/archive/refs/tags/${LLAMA_VERSION}.zip"
+        unzip -q llama-bin.zip -d llama-prebuilt
+        unzip -q llama-src.zip -d llama-prebuilt
+
+        sudo cp llama-prebuilt/build/bin/*.so /usr/local/lib/
+        sudo cp llama-prebuilt/llama.cpp-${LLAMA_VERSION}/include/*.h /usr/local/include/
+        sudo cp llama-prebuilt/llama.cpp-${LLAMA_VERSION}/ggml/include/*.h /usr/local/include/
+        sudo ln -sf /usr/local/lib/libggml-cpu-x64.so /usr/local/lib/libggml-cpu.so
+
+        mkdir -p ~/llama-backends
+        cp llama-prebuilt/build/bin/libggml-*.so ~/llama-backends/
+
+        sudo mkdir -p /usr/local/lib/pkgconfig
+        cat << 'EOF' | sudo tee /usr/local/lib/pkgconfig/llama.pc
+        prefix=/usr/local
+        exec_prefix=${prefix}
+        libdir=${prefix}/lib
+        includedir=${prefix}/include
+
+        Name: llama
+        Description: C/C++ inference of LLaMA-family models
+        Version: 0.0.6569
+        Libs: -L${libdir} -lggml -lggml-base -lllama
+        Cflags: -I${includedir}
+        EOF
+        sudo ldconfig
+        rm -rf llama-bin.zip llama-src.zip llama-prebuilt/
+
+        mkdir -p tests/test_models/models
+        curl -fsSL -o tests/test_models/models/TinyStories-656K-Q2_K.gguf \
+          "https://huggingface.co/tensorblock/TinyStories-656K-GGUF/resolve/main/TinyStories-656K-Q2_K.gguf"
+        echo "5a798ebc1cad921e0ce639b2f10c4bcce492654a79d4b4ca52bd0268416f6b87  tests/test_models/models/TinyStories-656K-Q2_K.gguf" | sha256sum -c
     - name: build and unit test
       if: env.rebuild == '1'
       run: |
         meson setup build/
         meson compile -C build/
+
+        if [ '${{ matrix.os }}' == 'ubuntu-22.04' ]; then
+          # Fail loudly if meson stopped detecting llama.cpp; otherwise the
+          # llamacpp tests would silently vanish from this job.
+          test -f build/ext/nnstreamer/tensor_filter/libnnstreamer_filter_llamacpp.so
+          cp ~/llama-backends/libggml-*.so build/tests/
+        fi
 
         sudo mkdir -m 777 /cores
         sudo bash -c 'echo "/cores/coredump" > /proc/sys/kernel/core_pattern'
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -44,6 +44,10 @@ jobs:
         LLAMA_VERSION="b6569"
         curl -fsSL -o llama-bin.zip "https://github.com/ggml-org/llama.cpp/releases/download/${LLAMA_VERSION}/llama-${LLAMA_VERSION}-bin-ubuntu-x64.zip"
         curl -fsSL -o llama-src.zip "https://github.com/ggml-org/llama.cpp/archive/refs/tags/${LLAMA_VERSION}.zip"
+        echo "52cc91b7b0a0d8e40bc79f7049a0b03c9055ac90ddcdf9f1f7307c5f3939634e  llama-bin.zip" | sha256sum -c
+        # Auto-generated source archive: its checksum is stable in practice but
+        # not permanently guaranteed by GitHub; update the hash if it changes.
+        echo "25c8a7c5f5f8411659ccfa7c24caacc5f7d7a36ea758d120e2429a888612d447  llama-src.zip" | sha256sum -c
         unzip -q llama-bin.zip -d llama-prebuilt
         unzip -q llama-src.zip -d llama-prebuilt
 
@@ -63,10 +67,11 @@ jobs:
 
         Name: llama
         Description: C/C++ inference of LLaMA-family models
-        Version: 0.0.6569
+        Version: 0.0.0
         Libs: -L${libdir} -lggml -lggml-base -lllama
         Cflags: -I${includedir}
         EOF
+        sudo sed -i "s/^Version: .*/Version: 0.0.${LLAMA_VERSION#b}/" /usr/local/lib/pkgconfig/llama.pc
         sudo ldconfig
         rm -rf llama-bin.zip llama-src.zip llama-prebuilt/
 
--- a/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc
@@ -298,6 +298,10 @@ TensorFilterLlamaCpp::configure_instance (const GstTensorFilterProperties *prop)
     throw std::runtime_error ("Failed to create llama context");
   }
 
+  /* n_ctx 0 means the model default; sync it for the trimming guard in generateTokens() */
+  if (n_ctx == 0)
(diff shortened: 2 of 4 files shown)


myungjoo:
> **Note:** This review was written and posted by another agent (an automated code-review agent), not typed directly by a human maintainer. Treat it as an automated first pass.

I reviewed `.github/workflows/ubuntu_clean_meson_build.yml` (the only file this PR touches), cross-checked it against `meson.build` / `tests/meson.build`'s `dependency('llama')` detection, `packaging/run_unittests_binaries.sh`, the already-landed `tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc`, and the prior CI-validated approach in #4775. I also let this PR's own draft CI run to completion and inspected the logs (run `32819294678`, job `97713873446`).

## Critical

**The exact CI job this PR turns on is currently red, and the failure is timing-flakiness, not a one-off fluke risk.** In this PR's own live run, `install llama.cpp for the llamacpp tensor-filter` passed fully (download, sha256 match, header/lib copy, `llama.pc` generation, `libnnstreamer_filter_llamacpp.so` build assertion — all verified OK), but `meson test` then failed:

```
../tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc:288: Failure
  *new_sample_count
    Which is: 1
  2
[  FAILED  ] NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p (1017 ms)
...
18/19 nnstreamer:unittest_filter_llamacpp  FAIL  5.68s  exit status 1
```
(`Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)`, https://github.com/nnstreamer/nnstreamer/actions/runs/32819294678/job/97713873446)

`multipleInputsSingleOutputSync_p` pushes 2 prompts in sync mode, then does a fixed `g_usleep(1000000)` (1s) before asserting exactly 2 output samples — a wall-clock race against CPU-only inference on a shared GitHub-hosted runner. This test line is not part of this PR's diff (it already landed on `main`), but this workflow is what makes it run in CI for the very first time (previously `dependency('llama')` was never satisfied, so the whole suite was silently skipped, which is exactly the gap this PR is meant to close).

Since `Build with meson and test with Valgrind in Ubuntu` gates every PR to `main`, merging this change as-is will make that check fail/flake on unrelated PRs going forward — the opposite of the PR's stated goal of "reliable CI coverage." Recommend one of, before this leaves draft:
- Fix the timing assumption in `multipleInputsSingleOutputSync_p` (and any sibling tests using fixed `g_usleep` synchronization) to poll/wait with a generous timeout instead of a fixed sleep, or
- Land that test fix as a prerequisite/companion commit and re-run this workflow to confirm green before un-drafting.

This is a real, reproduced failure from this PR's own CI run, not a hypothetical.

## Minor

The `sudo ln -sf /usr/local/lib/libggml-cpu-x64.so /usr/local/lib/libggml-cpu.so` symlink appears to be dead code. `llama.pc`'s `Libs:` line only references `-lggml -lggml-base -lllama` (not `-lggml-cpu`), and the backend libraries copied into `build/tests/` for ggml's dlopen-based backend search keep their original `libggml-cpu-x64.so` name (copied from `~/llama-backends`, not from `/usr/local/lib`). The observed successful inference runs in the CI log suggest the generic-named symlink in `/usr/local/lib` isn't actually being consulted. Since it's otherwise harmless on an ephemeral CI VM, this is just a nit — either drop it or add a short comment on why it's needed (e.g. if some other component does look it up by the generic name) so future maintainers aren't left guessing.

## Confirmed non-issues (checked, no action needed)

- **`ubuntu-22.04-arm` job unaffected**: the new install step and the build-assertion block are both gated on `matrix.os == 'ubuntu-22.04'`; the live run shows the install step skipped (`-`) on the arm job, and the header/lib copies never touch that job's environment.
- **No collision with `packaging/run_unittests_binaries.sh`**: it selects files by `-name "unittest_*"`, so the `libggml-*.so` files copied into `build/tests/` are never picked up as test binaries.
- **No collision with other subplugins' builds**: only the specific header lists from `ggml/include/*.h` and the llama.cpp top-level `include/*.h` are copied into `/usr/local/include`; no generically-named headers that would plausibly shadow other dependencies.
- **`dependency('llama')` detection matches**: `meson.build`'s `llamacpp-support` feature does `dependency('llama', ...)`, matching the hand-written `llama.pc`'s `Name: llama`; confirmed working end-to-end by the live build (filter `.so` linked, build-assertion passed).
- **sha256 pin is correct**: the live run's `sha256sum -c` reported `tests/test_models/models/TinyStories-656K-Q2_K.gguf: OK`.
- **Fixes a real bug from the superseded #4775**: this PR extracts the release archive into an isolated `llama-prebuilt/` directory instead of the repo root, avoiding the collision #4775 had between its extracted `build/bin/` and meson's own `build/` directory it creates via `meson setup build/`.
- **Scope/size**: single file, 50 lines added, CI-only — appropriately scoped to the stated topic, doesn't touch source, packaging, or other workflows.
- **Docs/architecture**: this is a CI-only change with no API or architectural impact; no documentation update is required, and none was expected.

## Merge readiness

Given the PR is already in draft with a `DO NOT MERGE` label, the live CI failure above is presumably part of what's still being iterated on — but flagging explicitly: **not ok-to-merge as-is** because the workflow this PR adds currently fails its own target job due to the flaky test described above. The workflow YAML itself (structure, guards, dependency wiring, hash pinning, cleanup) looks correct and matches the previously-validated approach from #4775 with real improvements (isolated extraction dir, `curl -f`, sha256 verification, build-assertion regression guard). Once the flaky `multipleInputsSingleOutputSync_p` (and any similarly-timed tests) are stabilized, this looks mergeable.

myungjoo:
Review response (both items addressed):

- **Critical — llamacpp test failure in this PR's own CI run**: root-caused beyond timing. `contextContinuationInConversation_p` (sync, 2 inputs, `context_length:512`, 100 ms sleep) passed while `multipleInputsSingleOutputSync_p` (identical shape, no `context_length`) failed — so this was not merely runner slowness. With no `context_length`, the filter member `n_ctx` stays 0 while llama.cpp uses the model default (512 here), so the trimming guard `n_processed + prompt > n_ctx` fired on every invoke and wiped the entire KV cache (`n_keep = n_ctx/2 = 0`), leaving the second sync invoke with an empty output. Fixed in d7e2674's parent commit (06c4d76) by syncing `n_ctx` from `llama_n_ctx(ctx)` after context creation. The fixed-sleep timing assumption you flagged was real too and is fixed separately (d7e2674) with a bounded polling helper, so the count assertions no longer depend on runner speed.
- **Minor — `libggml-cpu` symlink**: confirmed dead (the prebuilt is a `GGML_BACKEND_DL` build; ggml only scans executable-relative paths for `libggml-cpu-*.so` variants). Removed in e9bdc82.

myungjoo:
> **Note:** This review was written and posted by another agent (an automated code-review agent), not typed directly by a human maintainer. Treat it as an automated second pass (re-review of the 3 commits added since round 1).

Re-reviewed after the 3 new commits (9e19c14 n_ctx sync, d7e2674 polling helper, e9bdc82 symlink removal) that address round-1's Critical (flaky `multipleInputsSingleOutputSync_p`) and Minor (dead `libggml-cpu` symlink) findings. Checked the diff against `ext/nnstreamer/tensor_filter/tensor_filter_llamacpp.cc`, the full `tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc`, `tests/unittest_util.h`, `tests/meson.build`, the upstream `llama.h` at tag `b6569`, and this PR's own live CI runs.

## Verified: both round-1 findings are correctly resolved

**Critical (n_ctx sync, 9e19c14):** Confirmed `llama_n_ctx(const struct llama_context *)` exists in `llama.cpp` `b6569`'s `include/llama.h` and returns `uint32_t`; the `(int)` cast in `tensor_filter_llamacpp.cc:303` is safe for any realistic context size (no overflow risk vs `INT_MAX`). Placement is correct: the sync happens right after `llama_init_from_model()` succeeds and *before* the `load_ctx_path` branch, so a loaded saved-context's KV state doesn't affect what `n_ctx` gets synced to, and `save_ctx`/`load_ctx` paths are unaffected either way. `n_ctx < 0` validation in `parseCustomProperties()` still runs earlier against the *user-supplied* value only, so it doesn't conflict with the post-creation sync of the "0 = model default" case.

I also traced why only `multipleInputsSingleOutputSync_p` failed in round 1 while other default-`context_length` tests (e.g. `applyLoraAdapter_p`) did not: the trimming guard in `generateTokens()` is checked once per invoke, not per generated token. On a *fresh* context (`n_processed == 0`), `n_ctx == 0` makes the guard fire but degenerates to a no-op (`llama_memory_seq_rm(mem, 0, 0, 0)` — nothing to trim yet), so single-invoke tests were silently unaffected. Only tests that invoke `generateTokens()` a second time on the *same* context (like `multipleInputsSingleOutputSync_p`, sync mode, exact-count assertion) actually observed the KV cache getting wiped between invokes. `cacheTrimmingTrigger_p` and `contextSaveLoad_p` were never at risk either way since they already pass `context_length` explicitly (20 and 512). This is all consistent with the fix and I found no gap in it.

**Minor (dead symlink, e9bdc82):** Confirmed clean removal — no remaining reference to `libggml-cpu.so` or `ln -sf` anywhere in `ubuntu_clean_meson_build.yml`.

## Test helper review (d7e2674)

- `wait_for_samples()`'s "poll a non-atomic `guint*` from another thread with `g_usleep` between checks" pattern is not new to this PR — it matches the existing convention in `tests/nnstreamer_plugins/unittest_plugins.cc` (`_harness_wait_for_output_buffer`, same shape: `while (received < expected && count < N) { g_usleep(...); ... }`). No new risk introduced.
- Cross-checked all 18 `wait_for_samples(N)` call sites against their following assertion: every `EXPECT_GT(*count, X)` site polls for `X+1`, every `EXPECT_EQ(*count, X)` site polls for exactly `X`. All consistent (e.g. `singleInputMultipleOutputsAsync_p`: wait 11 / assert `> 10`; `multipleInputsSingleOutputSync_p`: wait 2 / assert `== 2`; `contextSaveLoad_p`'s two stages: wait 1 / assert `== 1` each, with `*new_sample_count = 0` correctly resetting between stages).
- One `g_usleep(100000)` remains unconverted, in `invalidNumPredict_n` (`num_predict:0`). Correct to leave as-is — that test makes no assertion on `*new_sample_count` at all, so there's nothing for a bounded wait to synchronize against.

**Nit (non-blocking):** `wait_for_samples()`'s default timeout is 10s, and `UNITTEST_STATECHANGE_TIMEOUT` (used for both the PLAYING and NULL transitions in every test) is another 10s each (`tests/unittest_util.h:31`). `unittest_filter_llamacpp` has no per-test `timeout:` override in `tests/meson.build` (unlike e.g. `unittest_filter_custom`, which sets `timeout: 120`), so it runs under meson's 30s default. Worst case (PLAYING wait + sample wait + NULL wait all near their ceiling) sits right at that 30s boundary, which would surface as a hard meson TIMEOUT instead of a clean assertion failure. Not a real risk today — the happy-path is now *faster* than before since polling exits as soon as samples arrive — but if this is meant to hold up under slow/loaded runners long-term, consider adding `timeout: 60` (or similar) to the `test('unittest_filter_llamacpp', ...)` registration for headroom.

## Regression / CI evidence

- Directly observed on this PR's own CI: run `32820368054` at commit `d7e2674` (n_ctx fix + polling helper, pre-symlink-removal) — **both** `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` and the `-arm` job completed with `conclusion: success`, i.e. the entire `unittest_filter_llamacpp` suite (Valgrind included) passed with these fixes applied. This directly supersedes round 1's reproduced failure on the same job.
- The final run at `e9bdc82` (`32820414518`, adds only the dead-code symlink removal) was still in progress at review time, but since that commit touches only CI YAML cleanup with no effect on build or test behavior, and the arm leg of that same run already completed successfully, there's no reason to expect a different outcome from the run that already went green at `d7e2674`.
- `async` invoke paths, `cacheTrimmingTrigger_p` (`context_length:20`), and `contextSaveLoad_p` (`context_length:512`, save/load round-trip) all pass in the green run above — no regression from the `n_ctx` sync.

## Scope

96/-19 across 3 files, all changes map directly to the stated fixes with no scope creep. No documentation exists for this filter's `context_length` semantics to begin with, so no doc update is warranted here.

## Conclusion

No new issues found. Both round-1 findings are correctly and minimally fixed, the test-helper rewrite is internally consistent across every call site, and this PR's own CI has already gone green on the combined fix (run `32820368054`). **OK to merge** from a code-review standpoint (the PR remains in draft with a `DO NOT MERGE` label, presumably pending the author's own final CI confirmation and un-drafting — that's a process step, not a code issue).

myungjoo:
Response to the 2026-08-25 automated review (3 non-blocking suggestions):

1. **Archive integrity** — addressed in 19a24b1. Both the prebuilt binary zip and the source zip are now sha256-pinned, consistent with the model file. Note: the source zip is a GitHub auto-generated archive whose checksum is stable in practice but not permanently guaranteed (unlike release assets); a comment in the workflow notes to update the hash if it ever changes — the failure mode is a loud CI error, not a silent drift.
2. **llama.pc Version hardcoding** — addressed in 19a24b1, going one step further than the suggested comment: the Version field is now derived from `LLAMA_VERSION` (`sed` after the quoted heredoc), so a future version bump cannot leave the .pc stale by construction.
3. **Grace period in wait_for_samples()** — keeping as-is, as the review itself deemed it a reasonable compromise. The distinction from the removed fixed sleeps: an insufficient sleep previously caused false *failures*, whereas the 100 ms grace only bounds the *detection power* for unexpected extra samples in exact-count assertions and cannot cause flaky failures. A signal-based wait (appsrc EOS + bus watch) would require restructuring how these tests drive the pipeline and is out of scope here.

PR #4864, opened 2026-08-25
[Test] fix float max-search init in filter unit tests

myungjoo:
## What this fixes

The `check_output` callbacks of the onnxruntime and tensorflow2-lite filter unit tests initialized the float max search with `G_MINFLOAT`, which is the smallest **positive** normalized float (~1.18e-38), not the most negative float. If every model output were negative (e.g., raw logits instead of softmax scores), `max_idx` would incorrectly remain at its initial value and the test would pass/fail for the wrong reason.

## Changes

- Replace the `G_MINFLOAT`-seeded float max searches with a shared `argmax_float ()` helper in `tests/unittest_util.h`, seeded with the first element so all-negative (and -inf/NaN-only) arrays are handled uniformly. It takes and returns `gsize` so the callers' `info_res.size / sizeof (gfloat)` is never truncated.
- Fix the same bug class in the armnn unit test: `get_argmax<T> ()` seeded `max_value` with 0, which breaks argmax for all-negative float arrays. It now seeds with the first element and guards `size == 0`. It stays a separate template because it is generic over `T` and also serves the `uint8` path, where a 0 seed is a valid lower bound.
- Add regression tests covering all-negative inputs and the empty-array guard, which would have caught the original bug: `commonUtil.argmaxFloatAllNegative` in `unittest_common` for the shared helper (always built, so the coverage does not depend on any optional subplugin), and `nnstreamerFilterArmnn.argmaxAllNegative` for the armnn template.

## Scope notes

- Test-only change; no production code is touched (`tests/unittest_util.h` is a test-only shared header).
- The same pattern in `ext/nnstreamer/tensor_decoder/tensordec-pose.c` (production code) is fixed separately in #4866.
- `box_properties/mobilenetssdpp.cc` uses `G_MINFLOAT` as a default threshold sentinel, which is intentional and not changed.
- `tests/tizen_nnfw_runtime`'s `get_argmax ()` keeps its 0 seed: it is `guint8`-only, where 0 is a valid lower bound.
- The armnn unit test now builds and runs in CI as of #4867, so its regression test is exercised there.

### DIFF
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -97,6 +97,18 @@ fill_tensors_config_for_test (GstTensorsConfig *conf1, GstTensorsConfig *conf2)
 }
 
 /**
+ * @brief Test argmax_float with all-negative values (e.g. raw logits)
+ */
+TEST (commonUtil, argmaxFloatAllNegative)
+{
+  const gfloat values[] = { -3.0f, -1.5f, -0.25f, -2.0f };
+
+  EXPECT_EQ (argmax_float (values, 4), 2U);
+  EXPECT_EQ (argmax_float (values, 2), 1U);
+  EXPECT_EQ (argmax_float (values, 0), 0U);
+}
+
+/**
  * @brief Test for int32 type string.
  */
 TEST (commonGetTensorType, failure_n)
--- a/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
+++ b/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
@@ -484,6 +484,18 @@ get_argmax (T *array, size_t size)
 }
 
 /**
+ * @brief Test get_argmax with all-negative values (e.g. raw logits)
+ */
+TEST (nnstreamerFilterArmnn, argmaxAllNegative)
+{
+  float values[] = { -3.0f, -1.5f, -0.25f, -2.0f };
+
+  EXPECT_EQ (get_argmax<float> (values, 4), 2U);
+  EXPECT_EQ (get_argmax<float> (values, 2), 1U);
+  EXPECT_EQ (get_argmax<float> (values, 0), 0U);
+}
+
+/**
  * @brief Test armnn subplugin with successful invoke for tflite advanced model
  */
 TEST (nnstreamerFilterArmnn, invokeAdvanced)
--- a/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
+++ b/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
@@ -467,8 +467,13 @@ size_t
 get_argmax (T *array, size_t size)
 {
   size_t idx, max_idx = 0;
-  T max_value = 0;
-  for (idx = 0; idx < size; idx++) {
+  T max_value;
+
+  if (size == 0)
+    return 0;
+
+  max_value = array[0];
+  for (idx = 1; idx < size; idx++) {
     if (max_value < array[idx]) {
       max_idx = idx;
       max_value = array[idx];
--- a/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
+++ b/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
@@ -86,25 +86,6 @@ _SetFilterProp (GstTensorFilterProperties *prop, const gchar *name, const gchar 
 }
 
 /**
- * @brief internal function to find the index of the maximum float value
- */
-static guint
-_argmax_float (const gfloat *values, guint num)
-{
-  guint idx, max_idx = 0U;
-  gfloat max_value = -G_MAXFLOAT;
-
-  for (idx = 0; idx < num; ++idx) {
-    if (values[idx] > max_value) {
-      max_value = values[idx];
-      max_idx = idx;
-    }
-  }
-
-  return max_idx;
-}
-
-/**
  * @brief Signal to validate the result in tensor_sink
  */
 static void
@@ -135,7 +116,7 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   } else if (is_float == 1) {
     gfloat *output = (gfloat *) info_res.data;
 
-    max_idx = _argmax_float (output, info_res.size / sizeof (gfloat));
+    max_idx = argmax_float (output, info_res.size / sizeof (gfloat));
   }
 
   gst_memory_unmap (mem_res, &info_res);
@@ -145,14 +126,15 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 }
 
 /**
- * @brief Test _argmax_float with all-negative values (e.g. raw logits)
+ * @brief Test argmax_float with all-negative values (e.g. raw logits)
  */
 TEST (nnstreamerFilterOnnxRuntime, argmaxFloatAllNegative)
 {
   const gfloat values[] = { -3.0f, -1.5f, -0.25f, -2.0f };
 
-  EXPECT_EQ (_argmax_float (values, 4), 2U);
-  EXPECT_EQ (_argmax_float (values, 2), 1U);
+  EXPECT_EQ (argmax_float (values, 4), 2U);
+  EXPECT_EQ (argmax_float (values, 2), 1U);
+  EXPECT_EQ (argmax_float (values, 0), 0U);
 }
 
 /**
--- a/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
+++ b/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
@@ -86,6 +86,25 @@ _SetFilterProp (GstTensorFilterProperties *prop, const gchar *name, const gchar 
 }
 
 /**
+ * @brief internal function to find the index of the maximum float value
+ */
+static guint
+_argmax_float (const gfloat *values, guint num)
+{
+  guint idx, max_idx = 0U;
+  gfloat max_value = -G_MAXFLOAT;
+
+  for (idx = 0; idx < num; ++idx) {
+    if (values[idx] > max_value) {
+      max_value = values[idx];
+      max_idx = idx;
+    }
+  }
+
+  return max_idx;
+}
+
+/**
  * @brief Signal to validate the result in tensor_sink
  */
 static void
@@ -115,20 +134,25 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
     }
   } else if (is_float == 1) {
     gfloat *output = (gfloat *) info_res.data;
-    gfloat max_value = G_MINFLOAT;
 
-    for (idx = 0; idx < (info_res.size / sizeof (gfloat)); ++idx) {
-      if (output[idx] > max_value) {
-        max_value = output[idx];
-        max_idx = idx;
-      }
-    }
+    max_idx = _argmax_float (output, info_res.size / sizeof (gfloat));
   }
 
   gst_memory_unmap (mem_res, &info_res);
   gst_memory_unref (mem_res);
 
   EXPECT_EQ (max_idx, 951U);
+}
+
+/**
+ * @brief Test _argmax_float with all-negative values (e.g. raw logits)
+ */
+TEST (nnstreamerFilterOnnxRuntime, argmaxFloatAllNegative)
+{
+  const gfloat values[] = { -3.0f, -1.5f, -0.25f, -2.0f };
+
+  EXPECT_EQ (_argmax_float (values, 4), 2U);
+  EXPECT_EQ (_argmax_float (values, 2), 1U);
 }
 
 /**
--- a/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
+++ b/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
@@ -101,7 +101,7 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   ASSERT_TRUE (mapped);
 
   gint is_float = (gint) * ((guint8 *) user_data);
-  guint idx, max_idx = 0U;
+  gsize idx, max_idx = 0U;
 
   if (is_float == 0) {
     guint8 *output = (guint8 *) info_res.data;
@@ -123,18 +123,6 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   gst_memory_unref (mem_res);
 
   EXPECT_EQ (max_idx, 951U);
-}
-
-/**
- * @brief Test argmax_float with all-negative values (e.g. raw logits)
- */
-TEST (nnstreamerFilterOnnxRuntime, argmaxFloatAllNegative)
-{
-  const gfloat values[] = { -3.0f, -1.5f, -0.25f, -2.0f };
-
-  EXPECT_EQ (argmax_float (values, 4), 2U);
-  EXPECT_EQ (argmax_float (values, 2), 1U);
-  EXPECT_EQ (argmax_float (values, 0), 0U);
 }
 
 /**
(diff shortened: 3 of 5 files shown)


myungjoo:

**Verdict: No blocking issues found. Approve.**

**Bug fix correctness**
- Confirmed the bug: `G_MINFLOAT` (~1.18e-38) is the smallest *positive* normalized float, not the most negative one, so the original `gfloat max_value = G_MINFLOAT;` search would never update `max_idx` away from its initial value if every output were negative (e.g. raw logits). Replacing it with `-G_MAXFLOAT` in both `tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc` and `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc` correctly fixes this.
- The new `argmaxFloatAllNegative` tests use inputs `{-3.0f, -1.5f, -0.25f, -2.0f}` and correctly expect index 2 (full array) and index 1 (first two elements). I traced through the pre-fix code path: with the old `G_MINFLOAT` init, both assertions would have failed (max_idx would stay 0), confirming these tests genuinely catch the original bug and guard against regression.
- The incidental change in the tflite2 file (`max_idx = -1` → `max_idx = 0U`) is harmless — `max_idx` is unconditionally overwritten in both the `is_float==0` and `is_float==1` branches before use, so this is a safe cleanup, not a behavior change.

**Scope / regression risk**
- Change is confined to two test files only; no production/shared code touched. No regression risk to other modules.
- Size is proportional to the fix (small helper extraction + 2 focused unit tests per file).

**CI enforcement**
- Verified both `unittest_filter_onnxruntime` and `unittest_filter_tensorflow2_lite` are registered via `test()` in `tests/meson.build` (gated on `onnxruntime_support_is_available` / `tflite2_support_is_available`).
- Verified `packaging/nnstreamer.spec` runs `bash packaging/run_unittests_binaries.sh ./tests` under `%if 0%{?unit_test}`, which executes every `unittest_*` binary and aborts the GBS build on the first non-zero exit code — so a failing test blocks the RPM build.
- Verified `.github/workflows/gbs_build.yml` runs on every PR to `main` (x86_64, `--define "unit_test 1"`), and `.github/actions/check-rebuild/check_if_rebuild_requires.sh` classifies changes to `tests/*.cc` files as requiring a GBS rebuild (falls into the default `*)` case, not the doc-only exclusion list), so this PR's changes will actually trigger the GBS unit-test run rather than being skipped.
- Default spec flags (`onnxruntime_support`/`tensorflow2_lite_support`) are enabled for the Tizen Unified profile targeted by `.github/workflows/tizen.gbs.conf`, so these two suites do run in that pipeline (consistent with prior known onnxruntime float-test flakiness already tracked for this exact suite).

**Non-blocking observation (out of scope for this PR)**
- The same "wrong sentinel for a max-search" pattern also exists in `tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc`'s templated `get_argmax<T>` helper (`T max_value = 0;`), which is instantiated as `get_argmax<float>(...)` at two call sites. It isn't currently triggered (test values are non-negative), but it's the same latent bug class as the one fixed here. Not requesting any change in this PR — just flagging for a possible future follow-up, in the same spirit as the PR description's own notes about `tensordec-pose.c` and `mobilenetssdpp.cc`.

No API/architecture changes are present, so no documentation updates are required.

myungjoo:

**Re-review of commit a40d5eb9 "[Test] fix generic argmax init in armnn unit test"**

**Verdict: No blocking issues. Mergeable.**

**Correctness of the new fix**
- Confirmed the fix addresses exactly the bug class flagged in the previous review round: the templated `get_argmax<T>()` in `tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc` used to initialize `T max_value = 0;`, which would fail to find the max index if every element of an all-negative `T` array were less than 0 (e.g. `get_argmax<float>` over raw, non-softmaxed logits).
- The new implementation initializes `max_value = array[0]` and starts the scan loop at `idx = 1`, with an explicit `if (size == 0) return 0;` guard before dereferencing `array[0]`. This is correct and generic for any `T` (unlike a fixed sentinel such as `-G_MAXFLOAT`, it needs no per-type minimum value), and preserves the original tie-break semantics (strict `<`, so the earliest index wins on ties).
- Traced all three call sites and confirmed the fix does not change existing expected results:
  - `get_argmax<guint8>((guint8*)output.data, output.size)` (line 588) — expects `951`, unaffected since `guint8` is unsigned and `0` was already a valid minimum; the new code is equivalent here.
  - `get_argmax<float>((float*)output.data, num_labels)` (lines 673, 700) — expects `9` from a softmax/`prob` output, which is non-negative in practice, so behavior is unchanged for this real data; the fix only changes behavior for the previously-latent all-negative edge case.
- `size == 0` guard: previously the size==0 case was implicitly handled by the loop simply not executing (returning the initial `max_idx = 0`); the new code returns the same value via an explicit early return, so this is a no-op behavior-wise, just more explicit/defensive.

**Scope**
- Change is 9 lines, confined to the single test helper in one file. No production code, other modules, or public API touched. Proportionate to the stated topic.

**Two non-blocking observations**
1. **Test coverage for the fix itself**: The sibling commit in this PR (fbe8d0b7) added dedicated `argmaxFloatAllNegative` unit tests for the extracted `_argmax_float()` helper in the onnxruntime/tflite2 suites. This armnn commit does not add an equivalent standalone `TEST` for `get_argmax<T>` with all-negative input (e.g. `get_argmax<float>` over `{-3.0f, -1.5f, -0.25f}` should return `1`). None of the existing armnn tests exercise this path (the real model outputs used are non-negative), so the corrected behavior is currently unverified by any test in the repo, even in environments where armnn is built. Consider adding a small direct test for consistency with the sibling fix, though this is optional.
2. **CI does not build/run this test at all**: `packaging/nnstreamer.spec` hardcodes `%define armnn_support 0` unconditionally (line 26), and only clears it to 0 again for non-ARM targets — it is never set to `1` anywhere in the spec. This means `.github/workflows/gbs_build.yml` (x86_64/armv7l/aarch64) never builds `unittest_filter_armnn` regardless of this change. `.github/workflows/ubuntu_clean_meson_build.yml` also doesn't install an armnn dependency, so meson's `armnn-support` (`type: feature, value: 'auto'`) resolves to disabled there too, per `tests/meson.build`'s `if armnn_support_is_available` gate. Net effect: **no CI job in this repository currently compiles or executes `unittest_filter_armnn.cc`**, for this PR or otherwise. This is a pre-existing, repo-wide gap (not introduced or worsened by this PR), but it does mean that if this fix (or any future change to this file) were subtly wrong, CI would not catch it — only manual/local builds with armnn enabled would. Flagging for visibility per the review checklist; not a reason to block this PR.

No architecture/API changes are present, so no documentation updates are required.

myungjoo:

**Re-review of commit 7e5fba9c "[Test] add all-negative argmax case for armnn unit test"**

**Verdict: No blocking issues. Mergeable.**

**New test correctness**
- `TEST(nnstreamerFilterArmnn, argmaxAllNegative)` uses `values = {-3.0f, -1.5f, -0.25f, -2.0f}`.
- `get_argmax<float>(values, 4)` expects `2U`: traced the post-fix loop (`max_value = array[0]`, scan from idx 1) — -1.5 > -3.0 (idx=1), -0.25 > -1.5 (idx=2), -2.0 not > -0.25 → result 2. Correct.
- `get_argmax<float>(values, 2)` expects `1U`: only {-3.0, -1.5} — -1.5 > -3.0 → result 1. Correct.
- `get_argmax<float>(values, 0)` expects `0U`: hits the `if (size == 0) return 0;` guard added in a40d5eb9. Correct, and this is the guard-path coverage requested in the prior review round.
- Regression check against the pre-fix code (`T max_value = 0; for (idx = 0; ...)`, from before a40d5eb9): all four array values are negative, so `max_value < array[idx]` would never be true, and `max_idx` would stay 0 for both the size=4 and size=2 cases — i.e. both non-guard assertions would have failed pre-fix (expected 2U/1U, actual 0U/0U). Confirms this test genuinely catches the original bug and guards against regression, matching the sibling `argmaxFloatAllNegative` tests added for onnxruntime/tflite2 in fbe8d0b7.

**Prior review items — status**
- (a) Missing dedicated all-negative test for armnn's `get_argmax<T>`: resolved by this commit.
- (b) No CI job builds/runs `unittest_filter_armnn.cc` (armnn_support hardcoded to 0 in packaging/nnstreamer.spec, not installed in ubuntu_clean_meson_build.yml): still true, still out of scope for this PR as previously agreed — tracked as a separate follow-up, not a blocker here.

**Scope / regression / docs**
- Diff is 12 lines, additive only, confined to the existing armnn unit test file; no production code touched, no regression risk to other modules.
- Size is proportionate (one focused test case, consistent style with existing `TEST` blocks and with the sibling onnxruntime/tflite2 tests).
- No API/architecture changes, so no documentation updates are required.

All observations raised across the three review rounds are now addressed within this PR's scope. Approve.

myungjoo:

**Re-review after rebase + commit 6374bafd "[Test] deduplicate float argmax into unittest_util.h"**

**Verdict: No blocking issues. Mergeable.**

**Suggestion ① (unify seeding) + ② (deduplicate `_argmax_float`)**
- `tests/unittest_util.h` now defines `static inline guint argmax_float (const gfloat * values, guint num)`, seeded with `values[0]` (matching armnn's `get_argmax<T>()` pattern) with an explicit `if (num == 0U) return 0U;` guard before dereferencing `values[0]`. This resolves the -inf/NaN-only edge case the old `-G_MAXFLOAT` sentinel in onnxruntime/tflite2 didn't handle, and now all three suites (onnxruntime, tflite2, armnn) share the same "seed with first element" semantics.
- Both `tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc` and `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc` had their local `_argmax_float()` definitions removed entirely; `check_output()` in both now calls the shared `argmax_float()` and still asserts `EXPECT_EQ (max_idx, 951U)` — behavior is unchanged. The `argmaxFloatAllNegative` tests in both files were updated to call the renamed function and gained a third assertion, `EXPECT_EQ (argmax_float (values, 0), 0U)`, exercising the new empty-array guard.
- Traced through: pre-fix (`G_MINFLOAT` in onnxruntime/tflite2, `T max_value = 0` in armnn) all three would have failed on the all-negative regression inputs; post-fix all three pass with the unified seeding.

**Shared-header impact (`tests/unittest_util.h`)**
- Checked all 32 files under `tests/` that include `unittest_util.h`, including the one plain-C consumer, `tests/nnstreamer_filter_reload/tensor_filter_reload_test.c` — no name collisions with `argmax_float` anywhere else in the repo.
- Top-level `meson.build` does not pass `-Wall`/`-Wextra`/`-Wunused*`, only a curated list of specific `-Werror=...` diagnostics, none of which is unused-function/unused-variable. Independent of that, GCC/Clang do not warn on an unused `static inline` function in the first place, so translation units that include the header without calling `argmax_float()` are unaffected either way.
- The function sits inside the header's existing `#ifdef __cplusplus extern "C" { ... }` guard, consistent with the rest of the header; `static inline` with a `const gfloat *` parameter is valid under both the project's `c_std=gnu89` (C consumer) and `cpp_std=c++17` (C++ consumers) settings.

**Rebase / coexistence with main**
- `origin/main` tip is `6bea3717` (post #4861 TensorRT10 rebase), and `git merge-base HEAD origin/main` equals that same commit — the branch is a clean, purely-additive rebase with no stray commits. `git diff origin/main...HEAD` touches exactly the 4 files listed in the PR's file list (`unittest_filter_armnn.cc`, `unittest_filter_onnxruntime.cc`, `unittest_filter_tensorflow2_lite.cc`, `unittest_util.h`).
- Confirmed `origin/main` already contains #4865 (`mobilenet_v2_float.onnx` present, onnxruntime test expects `951U` against the float model) and #4867 (armnn CI enabled via `-Darmnn-support=enabled` in `.github/workflows/ubuntu_clean_meson_build.yml`, rank-aware comments/expectations already in `unittest_filter_armnn.cc`). This PR's `get_argmax<T>()` fix and `argmaxAllNegative` test layer cleanly on top of that pre-existing armnn test content — no overlap/conflict with #4867's changes.

**PR body accuracy**
- Body's Changes section now explicitly covers the armnn `get_argmax<T>()` fix (previously only onnxruntime/tflite2 were mentioned) and references #4866/#4867, matching the actual diff and closing out non-blocking suggestion ③.
- The note "The armnn unit test now builds and runs in CI as of #4867" checks out against `ubuntu_clean_meson_build.yml` on `origin/main`, closing out non-blocking observation ④.

All four non-blocking suggestions from the prior review round are addressed. No new issues found in this round.

myungjoo:

**Re-review of commit 212aeb8f "[Test] widen argmax_float to gsize and test it unconditionally"**

**Verdict: No blocking issues. Mergeable.**

**Nit ① (gsize widening) — resolved**
- `tests/unittest_util.h`: `argmax_float()` now takes `gsize num` and returns `gsize`, with `idx`/`max_idx` also `gsize`. The `if (num == 0U) return 0U;` guard and the seed-with-`values[0]` logic are otherwise unchanged.
- Both call sites (`tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc:119`, `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc:102`) now assign into `gsize idx, max_idx` (was `guint`), so `argmax_float (output, info_res.size / sizeof (gfloat))` — a `gsize` expression — is no longer narrowed. The `uint8` branch's `for (idx = 0; idx < info_res.size; ++idx)` also benefits: `idx` and `info_res.size` (a `gsize` field of `GstMapInfo`) are now the same type, removing even the implicit-promotion comparison that existed before.
- Checked `meson.build`: the project sets `werror=true` + `warning_level=2` project-wide (≈ `-Wall -Wextra` as errors), plus an explicit `-Werror=...` list that does not include `sign-compare`. `EXPECT_EQ (max_idx, 951U)` and `EXPECT_EQ (argmax_float (values, 4), 2U)` compare `gsize` (unsigned) against `unsigned int` literals — both operands are unsigned, so this is an unsigned-unsigned widening comparison, not a signed/unsigned mismatch, and does not trigger `-Wsign-compare` even under `-Wextra`/`warning_level=2`. No compiler could be run in this sandbox to double-check empirically, but the reasoning is unambiguous (no signedness mismatch is present anywhere in the diff).

**Nit ② (deduplicate coverage into an unconditionally-built binary) — resolved**
- The duplicated `TEST(..., argmaxFloatAllNegative)` blocks were removed from both `unittest_filter_onnxruntime.cc` and `unittest_filter_tensorflow2_lite.cc` (which are gated behind `onnxruntime_support_is_available` / `tflite2_support_is_available` in `tests/meson.build`), and a single copy, `TEST(commonUtil, argmaxFloatAllNegative)`, was added to `tests/common/unittest_common.cc`.
- Verified `tests/common/unittest_common.cc` includes `unittest_util.h` (line 19) and that its `executable('unittest_common', ...)` + `test('unittest_common', ...)` registration in `tests/meson.build` (lines 76-83) sits only inside `if gtest_dep.found()` — the same baseline gate every gtest-based suite in this project requires (including the two subplugin suites), not behind any optional-subplugin feature. So the helper's only regression coverage no longer depends on onnxruntime or tflite2 being enabled.
- Confirmed via grep that no stale references to the removed per-subplugin `argmaxFloatAllNegative` tests remain, and there's exactly one `argmax_float` definition.

**Coverage check**
- Both subplugin `check_output()` functions still exercise `argmax_float()` via the real `1U/2U/951U`-branch assertions (`EXPECT_EQ (max_idx, 951U)`), so removing the duplicated unit test did not remove exercising of the helper in those binaries — it only removed redundant, duplicate direct-unit-test coverage of the same helper logic.
- armnn's separate templated `get_argmax<T>()` and its own `TEST(nnstreamerFilterArmnn, argmaxAllNegative)` (lines 467, 489-495 of `unittest_filter_armnn.cc`) are untouched by this commit, as expected — they're a different helper with independent seeding, not affected by this change.

**PR body / scope**
- PR body's "Changes" section now correctly states `argmax_float()` "takes and returns `gsize`" and that the regression test is `commonUtil.argmaxFloatAllNegative` in `unittest_common` ("always built, so the coverage does not depend on any optional subplugin") — matches the diff.
- Diff vs `origin/main` is confined to the 5 files already covered by prior review rounds (`unittest_common.cc`, `unittest_filter_armnn.cc`, `unittest_filter_onnxruntime.cc`, `unittest_filter_tensorflow2_lite.cc`, `unittest_util.h`). No production code touched; no documentation changes required (test-only change, no API/architecture impact).

Both non-blocking nits from the prior review round are fully addressed. No new issues found.

PR #4865, opened 2026-08-25
[Test/onnxruntime] Use float model for result verification

myungjoo:
### Problem

`nnstreamerFilterOnnxRuntime.floatModelResult` (and the ssat golden test of the onnxruntime filter) fails sporadically in the "Tizen GBS build on Ubuntu (x86_64, unit_test 1)" job on PRs that do not touch anything related — e.g. [this run](https://github.com/nnstreamer/nnstreamer/actions/runs/32821326864/job/97719884798) returned `max_idx` 522 instead of 951 and passed on an immediate re-run with no code change.

### Root cause

Not package drift (both the failing and passing runs installed the identical `onnxruntime-1.16.3-1`), not model/file corruption (the model and image are committed in-repo). The cause is the **heterogeneous GitHub runner pool**: onnxruntime dispatches quantized (u8) kernels by CPU feature. VNNI-capable Intel Xeons compute exactly, while AVX2-without-VNNI CPUs (e.g. AMD EPYC 7763) use saturating `vpmaddubsw`-based kernels whose error is large enough to change the argmax of `mobilenet_v2_quant.onnx` for `orange.png` (orange margin over runner-up is only ~0.8 logits). Reproduced empirically: the same onnxruntime 1.16.3 with the same input produces measurably different logits per kernel path (native AVX-VNNI vs QEMU SSE-only), each path bit-deterministic. Upstream: microsoft/onnxruntime#6004, microsoft/onnxruntime#24220.

### Fix

Verify inference *results* with a float-compute model, which is numerically stable across CPU kernel paths, and keep the quantized model for everything that does not assert output values.

- **`tests/test_models/models/mobilenet_v2_float.onnx`** (new, 3.5 MiB): generated from `mobilenet_v2_quant.onnx` by the included `gen_mobilenet_v2_float.py` — all compute in float32, weights kept stored as uint8 via `DequantizeLinear` (so the file stays as small as the quant model instead of ~14 MiB), and per-layer `Clip` restoring the ReLU6 activations that were folded into the quantization ranges. Regeneration is byte-identical.
- `floatModelResult` (gtest) and the ssat "Golden test comparison" now use the float model.
- New `quantModelLaunch00` (gtest) and ssat case 2 keep the quantized model covered end-to-end through the pipeline without asserting the CPU-dependent output value.

### Verification

- Float model: top-1 = 951 ("orange") with **1.85 logit margin**; cross-path variance ~1e-5 across onnxruntime 1.16.3 (CI version) native AVX-VNNI, QEMU SSE-only, 1 and 4 intra-op threads, and onnxruntime 1.29 on Windows.
- Quant model divergence across the same paths confirmed (e.g. class 950 logit 15.254 vs 14.852) — the mechanism behind the CI flake.
- clang-format clean; changed-file sizes within the 5 MiB CI limit.

### Notes for reviewers

- This adds a 3.5 MiB binary to the repo. The generator script is committed alongside for provenance; regeneration is deterministic.
- The alternative (accepting the wrong argmax values per CPU) was rejected as brittle: the saturated output is not a single stable alternative across CPU generations.


myungjoo:
Note: This comment relays a review written by a separate review agent (not the PR author or a human reviewer). All findings below were independently verified against this branch's checked-out source (local regeneration of the model, local `pylint`/`clang-format` runs, local onnxruntime inference), not just read from the diff.

## Summary

The change correctly targets the stated problem: `nnstreamerFilterOnnxRuntime.floatModelResult` and the ssat "Golden test comparison" case switch from asserting a hard-coded argmax on the **quantized** onnxruntime model (whose u8 kernel path — and thus its output — is CPU-feature-dependent) to asserting it on a new **float-compute** model (`mobilenet_v2_float.onnx`), while a new non-asserting smoke test (`quantModelLaunch00` / ssat case 2) keeps the quantized model exercised end-to-end. The change is well-scoped: only `tests/nnstreamer_filter_onnxruntime/*` and `tests/test_models/models/*` are touched, no plugin/library code is modified, and the `_GetModelFilePath` → `_GetModelFilePathByName` refactor is verified to preserve behavior for every other of the 8 existing call sites that still use the quantized model.

I independently verified two of the PR's central technical claims by running the tooling locally on this checkout:

- **Byte-identical regeneration**: running `python tests/test_models/models/gen_mobilenet_v2_float.py` locally (onnx 1.22.0 / numpy 2.5.2) reproduces `mobilenet_v2_float.onnx` byte-for-byte (`git status` clean afterward, sha256 matches).
- **The stated margin**: running onnxruntime 1.29.0 (CPUExecutionProvider) on this machine against `orange.png`: the float model gives argmax=951 ("orange") with a **1.85** logit margin over the runner-up — matching the PR description exactly — while the quantized model on this same run gives argmax=951 with only a **0.80** logit margin, i.e. the fragile margin the PR describes as flipping on AVX2-without-VNNI CPUs. This corroborates the root-cause analysis.

## Findings

### [Major] The added `gen_mobilenet_v2_float.py` will fail the required "Pylint" CI check

`.github/workflows/static.check.scripts/pylint.sh` (invoked from the `Static checkers and verifiers` workflow, which runs on every PR) generates a fresh default `.pylintrc` via `pylint --generate-rcfile` and fails the job (`exit 1`) if **any** convention/warning/error/refactor message is emitted for a changed `.py` file — there is no exclusion for `tests/`.

Running that exact sequence against `tests/test_models/models/gen_mobilenet_v2_float.py` produces 9 convention + 7 warning messages, e.g.:
- `C0114: Missing module docstring`
- `C0103: Constant name "dq_count"/"clip_count" doesn't conform to UPPER_CASE naming style`
- `C0116: Missing function or method docstring` (×5, on `keep_init`, `f`, `to_arr`, `add_float_init`, the nested `addend`)
- `W0603: Using the global statement` (×2)
- `W0621: Redefining name 'axis'/'out'/'ys'/'yzp' from outer scope` (×5)
- `C0301: Line too long (103/100)` (the `emit_clipped(...)` call inside `QLinearAdd` handling)

As currently written, this file will make the static-check job fail, blocking merge until the script is cleaned up (docstrings, renamed locals/constants, or targeted `# pylint: disable=...` with justification) — independent of the PR's own `DO NOT MERGE` draft status. `clang-format` on the `.cc` change, by contrast, is confirmed clean.

### [Minor] The new quantized-model tests cannot detect future numeric regressions in that path

`quantModelLaunch00` (gtest) and ssat case 2 only assert that the pipeline reaches `PLAYING` and that `tensor_sink` receives at least one buffer; neither checks the output values. This is a reasonable and clearly deliberate trade-off given the root cause (the values are legitimately CPU-dependent), but it does mean any future regression that changes the *shape/validity* of quantized inference (without breaking the pipeline outright) — e.g. a change to the onnxruntime EP/session options, or to how nnstreamer's onnxruntime subplugin sets up quantized I/O — will not be caught by these two tests, only by the float-model test (which exercises a different set of ops: `Conv`/`Add`/`MatMul`/`GlobalAveragePool`/`Clip` rather than `QLinearConv`/`QLinearAdd`/`QLinearMatMul`/`QLinearGlobalAveragePool`). Worth a one-line acknowledgment in the PR description/comments that this is a known, accepted coverage gap rather than an oversight (much of it is already implied by "Positive test: ... no result check", so this is a documentation nit more than a code issue).

### [Nit] Generator script gets bundled into on-device test packages for no functional reason

`tests/meson.build` does `install_subdir('test_models', install_dir: unittest_install_dir)`, so placing `gen_mobilenet_v2_float.py` inside `tests/test_models/models/` means it will be installed alongside the actual test assets on target devices (Tizen GBS/Yocto test images) even though it's a one-off host-side generator that requires `onnx`/`numpy` and is never invoked by the test suite itself. Harmless, but it's dead weight in the installed test package; consider a `tools/` or `scripts/` location outside the installed subtree if that's easy, otherwise not blocking.

### Documentation / architecture check: N/A

No API or architecture change is involved (test-only change); no documentation update is required, and none was found to be missing.

## Other checks performed (no issues found)

- **Scope/size**: 239 additions / 5 deletions across 4 files, all under `tests/`; proportionate to the stated goal, no unrelated modules touched.
- **File-size CI gate**: the new binary is 3.5 MiB, under the repo's 5 MiB per-file CI limit (`.github/workflows/static.check.yml`), and is smaller than several pre-existing committed test models in the same directory (e.g. `mobilenet_v2_1.0_224.tflite` at ~13.9 MiB), so it's consistent with existing repo conventions even though it does add binary bulk.
- **Doxygen tag check**: the new `.py` file includes `@brief`, satisfying `doxygen-tag.sh`'s per-file requirement.
- **Regression risk to other modules**: none identified — the refactor of `_GetModelFilePath` into `_GetModelFilePathByName` is backward-compatible for all existing callers, and no plugin/core source is touched.

myungjoo:
Thanks for the review — all findings addressed in caf29e92:

1. **Major (pylint)**: `gen_mobilenet_v2_float.py` was rewritten pylint-clean (class-based converter: no globals, full docstrings, naming rules, line length). Verified locally with the CI's exact sequence (`pylint --generate-rcfile` then `pylint --reports=y`): **10.00/10, zero messages**. The generated node/initializer names changed with the rewrite, so `mobilenet_v2_float.onnx` was regenerated with it (still byte-identical on repeated regeneration; inference re-verified: top-1 = 951, margin 1.8532 — numerically identical logits, only internal tensor names changed).
2. **Minor (quantized numeric coverage gap)**: accepted as a deliberate trade-off, as noted. Any hardware-portable numeric assertion on the quantized path would have to be loose enough to be meaningless, since the divergence originates inside onnxruntime's kernel dispatch, not in nnstreamer. The quantized op path (`QLinearConv` etc.) remains exercised end-to-end by `quantModelLaunch00`/ssat case 2, which still catch loading, layout, and invocation regressions.
3. **Nit (generator installed on devices)**: fixed — `tests/meson.build` now passes `exclude_files: ['models/gen_mobilenet_v2_float.py']` to the `install_subdir` call, so the host-only script is no longer shipped in on-device test packages.

*Note: this comment was written by the author-side agent in response to the relayed review above.*

myungjoo:
Note: This comment relays a second-round review written by a separate review agent (not the PR author or a human reviewer), checking whether the first-round findings above were actually resolved by caf29e92 and whether the rewritten/regenerated artifacts introduce any new issues. As with the first round, everything below was independently re-derived on this branch's checked-out source (fresh `.pylintrc` generation + `pylint --reports=y`, local `clang-format -i` diff, local model regeneration + sha256, local onnxruntime 1.29.0 inference), not just read from the diff.

## Verification of the three first-round findings

1. **[Major] Pylint** — Confirmed fixed. Regenerated a clean default rc (`rm -f ~/.pylintrc && python -m pylint --generate-rcfile > ~/.pylintrc`, matching `pylint.sh`'s exact sequence) and ran `pylint --reports=y tests/test_models/models/gen_mobilenet_v2_float.py`: **10.00/10, 0 convention/warning/error/refactor messages**. `pylint.sh`'s pass condition (`grep` count of `W:`/`C:`/`E:`/`R:` lines == 0) is satisfied. One caveat worth noting for the record, not a blocker: this was verified with pylint 4.0.7 (pip); the CI job installs pylint via `apt-get` on `ubuntu-latest`, which resolves to an older major version (~3.x). Default-ruleset behavior has shifted across pylint 2→4, so a from-scratch CI run is still the authoritative check, but a clean result under a newer/stricter default ruleset is a good sign.
2. **[Nit] Generator script shipped to on-device test packages** — Confirmed fixed. `tests/meson.build`'s `install_subdir('test_models', install_dir: unittest_install_dir, exclude_files: ['models/gen_mobilenet_v2_float.py'])` uses the correct relative path (relative to the `test_models` source subdir being installed, i.e. `test_models/models/gen_mobilenet_v2_float.py`), and `exclude_files` does exact-path matching, not globbing, so it removes only this one host-only script and nothing else installed under `test_models` — no regression to the other assets in that directory or to any other `install_subdir`/`install_data` call in the file (all other install statements in `tests/meson.build` are untouched). No `.spec`/packaging files reference `test_models` explicitly (packaging just archives whatever `ninja install` produced), so this also fixes on-device/Tizen GBS/Yocto packaging, not just the meson-level intent.
3. **[Minor] Quantized-path numeric coverage gap** — Accepted trade-off, now explicitly acknowledged in the code (`@note` on `_GetFloatModelFilePath`, doc comment on `quantModelLaunch00`) and in the author's response above. Reasonable: no CPU-portable numeric assertion is possible here since the divergence is internal to onnxruntime's kernel dispatch, not nnstreamer behavior, and the quantized op path (`QLinearConv`/`QLinearAdd`/`QLinearMatMul`/`QLinearGlobalAveragePool`) is still exercised end-to-end (load, layout, invocation) by `quantModelLaunch00` / ssat case 2, so a future load/pipeline-level regression will still fail CI.

## New-issue check on the rewritten/regenerated artifacts

- **`gen_mobilenet_v2_float.py` (rewritten)**: pylint-clean as above; also re-verified the algorithm itself is unchanged in substance (same QOperator→float rewrite: `DequantizeLinear` for weights, folded int32 bias, `Clip` to reproduce the fused ReLU6, `QLinearAdd`/`QLinearMatMul`/`QLinearGlobalAveragePool` handling) — only refactored into a documented class with named methods instead of nested closures/globals.
- **`mobilenet_v2_float.onnx` (regenerated)**: ran the script locally (onnx 1.22.0 / numpy 2.5.2) — byte-identical (sha256 match, clean `git status` after). Independently ran onnxruntime 1.29.0 (CPUExecutionProvider) inference on both models against `orange.png` (NCHW `[1,3,224,224]`, `typecast:float32,div:127.5,add:-1.0` preprocessing matching the gst pipeline): float model → top-1 class 951 "orange", margin **1.8532**; quantized model → top-1 class 951 "orange" on this machine too, but margin only **0.8029**. This matches the author's reported numbers and confirms the fix's premise: the quantized model's margin is small enough to plausibly flip argmax under a different CPU's int8 kernel path, while the float model's margin is comfortably larger.
- **`tests/meson.build`**: covered above; no effect on other install targets.
- **`unittest_filter_onnxruntime.cc` / `runTest.sh` additions** (`quantModelLaunch00`, ssat case 2, `_GetModelFilePathByName` refactor): `clang-format -i` produces no diff (clean). `UNUSED` macro used in `count_output` resolves via the already-included `nnstreamer_util.h`. Doxygen `@brief`/`@note` tags are present for every new/renamed function, satisfying `doxygen-tag.sh`'s advanced (per-function) check. `_GetModelFilePathByName` is a backward-compatible refactor — verified all 8 other existing call sites of the old `_GetModelFilePath` name still resolve to the quantized model via the new one-line wrapper. New `mobilenet_v2_float.onnx` (3.57 MiB) stays under the 5 MiB CI file-size gate.
- **Scope**: still test-only (`tests/nnstreamer_filter_onnxruntime/*`, `tests/test_models/models/*`, `tests/meson.build`); no plugin/library/core code touched, so no functional regression risk to other modules.
- **CI gate effectiveness**: if a future onnxruntime plugin regression broke quantized-model loading or pipeline invocation outright, `quantModelLaunch00`/ssat case 2 would still fail CI (pipeline must reach PLAYING and emit ≥1 buffer); if it broke float-path inference correctness, `floatModelResult`/ssat case 1 would catch it via the golden-label assertion. Both paths remain covered by tests that can fail the build.

## Conclusion

All three first-round findings are verified resolved, and no new Critical/Major/Minor issues were found in the rewritten generator script, the regenerated model, or the `tests/meson.build` change. **Approve — OK to merge** from this review's perspective (still gated on `DO NOT MERGE` draft status / CI passing, per the PR's own labeling).

PR #4866, opened 2026-08-25
[Decoder/Pose] fix max-score search initialization

myungjoo:
## What this fixes

The keypoint max-score search in `pose_decode ()` (ext/nnstreamer/tensor_decoder/tensordec-pose.c) initialized `max` with `G_MINFLOAT`, which is the smallest **positive** normalized float (~1.18e-38), not the most negative float. If every score in a keypoint's heatmap is negative — raw logits in heatmap-only mode, or subnormal sigmoid outputs (e.g., sigmoid(-100) ≈ 3.7e-44) in heatmap-offset mode — the argmax incorrectly stays at (0, 0) and `p.prob` is reported as `G_MINFLOAT` instead of the true maximum.

## Changes

- Initialize the search with `-G_MAXFLOAT` instead of `G_MINFLOAT`.
- Add an SSAT case in `tests/nnstreamer_decoder_pose/runTest.sh` driving all-negative float heatmap values through the decoder.

## Impact analysis

The bug is currently **latent**: `draw ()` filters out keypoints with `prob < 0.5`, which covers every all-negative case, so the rendered output is identical before and after this fix — i.e., zero regression risk. The fix corrects the argmax and the propagated `prob` so that future changes to the drawing threshold or prob usage do not inherit the wrong result. Because the observable output cannot distinguish buggy from fixed behavior through the current pipeline, a golden-output regression test is not applicable; the added SSAT case keeps the all-negative path exercised.

The same pattern was fixed in the filter unit tests in #4864. The remaining `G_MINFLOAT` in `ext/` (`box_properties/mobilenetssdpp.cc` `THRESHOLD_DEFAULT`) is an intentional threshold sentinel, not a max-search initializer, and is unchanged.

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-pose.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-pose.c
@@ -773,7 +773,7 @@ pose_decode (void **pdata, const GstTensorsConfig * config,
   for (index = 0; index < pose_size; index++) {
     int maxX = 0;
     int maxY = 0;
-    float max = G_MINFLOAT;
+    float max = -G_MAXFLOAT;
     pose p;
     for (j = 0; j < grid_ysize; j++) {
       for (i = 0; i < grid_xsize; i++) {
--- a/ext/nnstreamer/tensor_decoder/tensordec-pose.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-pose.c
@@ -773,7 +773,7 @@ pose_decode (void **pdata, const GstTensorsConfig * config,
   for (index = 0; index < pose_size; index++) {
     int maxX = 0;
     int maxY = 0;
-    float max = -G_MAXFLOAT;
+    float max = -G_MAXFLOAT; /* not G_MINFLOAT, which is the smallest positive float */
     pose p;
     for (j = 0; j < grid_ysize; j++) {
       for (i = 0; i < grid_xsize; i++) {
--- a/tests/nnstreamer_decoder_pose/runTest.sh
+++ b/tests/nnstreamer_decoder_pose/runTest.sh
@@ -32,9 +32,6 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=4 ! videoc
 # TEST WITH MORE BUFFERS
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=20 ! videoconvert ! videoscale ! video/x-raw,width=14,height=14,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:128,div:255 ! tensor_split name=a tensorseg=1:14:14:1,2:14:14:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=14:14 ! fakesink" 2 0 0 $PERFORMANCE
 
-# TEST WITH ALL-NEGATIVE HEATMAP VALUES (the max-score search should still work)
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=4 ! videoconvert ! videoscale ! video/x-raw,width=14,height=14,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-256,div:255 ! tensor_split name=a tensorseg=1:14:14:1,2:14:14:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=14:14 ! fakesink" 4 0 0 $PERFORMANCE
-
 echo "nose 1 2 3 4
 leftEye 0 2 3
 rightEye 0 1 4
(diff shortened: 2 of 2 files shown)


myungjoo:

Reviewed the diff (`ext/nnstreamer/tensor_decoder/tensordec-pose.c` +1/-1, `tests/nnstreamer_decoder_pose/runTest.sh` +3) against the stated bug (max-score search initialized with `G_MINFLOAT`, the smallest **positive** normalized float, instead of `-G_MAXFLOAT`/-infinity, so an all-negative heatmap left argmax pinned at `(0,0)`).

**Fix correctness — OK.** `-G_MAXFLOAT` is the correct sentinel for a float max-search; it is less than any finite float32 value, so normal (non-degenerate) cases behave exactly as before and the degenerate all-negative case is now correctly resolved to the true max/argmax.

**Regression check — OK, claim verified.** `draw()` (line ~701) does `if (XYdata[i]->prob < 0.5) XYdata[i]->valid = FALSE;` before rendering, and `p.prob` is the only consumer of the search result outside the decode loop (no other file/module reads `pose.prob`, `maxX`, or `maxY`). Since `G_MINFLOAT` (~1.18e-38) and any legitimately negative post-fix max are both `< 0.5`, the rendered RGBA buffer is bit-identical before/after the fix for the all-negative case. No regression risk to other modules confirmed.

**Scope — OK.** Change is minimal (1 source line, 3 test lines), touches no unrelated code, no API/architecture change, so no doc updates are required.

**Test pipeline — verified correct.** `tensor_transform mode=arithmetic option=typecast:float32,add:-256,div:255` applies `typecast` first (avoiding uint8 wraparound) and then the remaining ops in the order listed in the option string (confirmed in `gst_tensor_transform_arithmetic`, both ORC and generic paths iterate `filter->operators` in parse order) — i.e. `add(-256)` then `div(255)`. For RGB pixel bytes in `[0,255]` this yields `[-256/255, -1/255]` ≈ `[-1.004, -0.0039]`, guaranteed negative for every element. `option4` (mode) is not set, so `pose_data->mode` defaults to `HEATMAP_ONLY`, meaning the raw (non-sigmoid) values are used directly as `cen` — this matches the "raw logits in heatmap-only mode" trigger described in the PR body. The new case does exercise the intended all-negative path.

**Golden-output claim — verified accurate.** The repo does have byte-exact golden-output comparison infra (`callCompareTest`, used e.g. in `tests/nnstreamer_decoder_boundingbox/runTest.sh`), but per the regression analysis above, this specific bug is unobservable through the rendered output (both buggy and fixed `prob` values fall below the 0.5 draw threshold), so a golden-output comparison genuinely could not have caught it. The PR's claim holds.

**CI enforcement — confirmed.** `tests/nnstreamer_decoder_pose/runTest.sh` is picked up by the Tizen/GBS `unit_test 1` job (`packaging/nnstreamer.spec` %check: `pushd tests; ssat -n -p=1 ...; popd` auto-discovers every `tests/*/runTest.sh`), which is one of this PR's required status checks. So the new case does run in CI.

**One non-blocking observation (informational only):** the new SSAT case only asserts pipeline completion (exit code / buffer count) via `gstTest`, not the internal `prob`/argmax values. Since the pre-fix code also completes this same pipeline without error or crash (`maxX=maxY=0`, `prob=G_MINFLOAT`, still filtered by `draw()`), the added case would pass identically with or without the one-line fix — it exercises the all-negative code path for crash-safety but does not itself act as a regression guard for this specific logic bug. This matches what the PR description already (honestly) claims — it does not claim the SSAT case detects the bug, only that it "keeps the path exercised." No action requested; flagging only for future awareness, since a true regression guard here would need a way to inspect `pose_decode()`'s internal result independent of the 0.5 render threshold, which is a larger change than this fix warrants.

**Verdict: no blocking issues found. This looks mergeable.**

myungjoo:

Re-reviewed commit 70067b84 ("[Decoder/Pose] address review suggestions"), which responds to myungjoo-bot's three non-blocking suggestions.

**1. SSAT comment clarifies no output verification** — Confirmed. `runTest.sh` now reads:
`# TEST WITH ALL-NEGATIVE HEATMAP VALUES (path exercise only: the output is not verified since keypoints with prob < 0.5 are never drawn)`
This matches the earlier regression analysis (both pre-fix `G_MINFLOAT` and post-fix negative `prob` fall below the 0.5 draw threshold, so the case cannot itself detect the bug) and removes the risk of the case being mistaken for a regression guard later. Addressed.

**2. Case 4 moved after case 3** — Confirmed via diff. The block was relocated from between case 2 and the `pose_label.txt` setup to after `rm pose_label.txt` (i.e., after case 3), restoring sequential file order `0_n, 1, 2, 3, 4`. The `gstTest` pipeline string and its trailing args (`4 0 0 $PERFORMANCE`) are byte-identical before and after the move — only the comment and position changed, no pipeline edits. Addressed.

**3. Comment on the initializer explaining why not G_MINFLOAT** — Confirmed. `tensordec-pose.c` line 776 now reads:
`float max = -G_MAXFLOAT; /* not G_MINFLOAT, which is the smallest positive float */`
Addressed.

**SSAT execution impact check** — None. Case 4 still runs after `rm pose_label.txt` and its pipeline has no `option3=pose_label.txt` dependency, so the reorder does not introduce any new file/state dependency. Case label stays `4` (position changed, not renumbered), so there's no collision with the existing `0_n, 1, 2, 3` labels.

**CI** — DCO and CPP-Linter already pass on this commit; the rest of the required checks were still running (pending) at review time. No format/lint issue expected from this change (the added inline comment is left untouched by `clang-format -style=file`, and the project's `indent.sh` static check is warning-only/non-blocking regardless).

No remaining issues found. All three suggestions from the prior approve review are accurately reflected, and the fix itself (previously verified: `-G_MAXFLOAT` sentinel, no regression via the 0.5 draw-threshold analysis) is unchanged. This PR looks mergeable.

PR #4867, opened 2026-08-26
[CI] build and test the armnn subplugin in the Ubuntu meson job

myungjoo:
## Problem

The armnn tensor-filter subplugin (`ext/nnstreamer/tensor_filter/tensor_filter_armnn.cc`) and its unit test (`tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc`) are not built by **any** CI job:

- `packaging/nnstreamer.spec` defines `armnn_support 0` unconditionally (disabled 2022-12-22 in 6571504c when Tizen dropped armnn).
- `debian/control.ubuntu.ppa` (which governs the pdebuild job's build-deps) lists no armnn packages.
- The Ubuntu meson workflow installed no armnn packages, so the `armnn-support` feature (`auto`) always resolved to disabled.

Regressions in the subplugin or its tests were invisible to CI. (Found during review of #4864.)

## Change

Ubuntu 22.04 universe ships Arm NN 20.08 — `libarmnn-dev`, `libarmnntfliteparser-dev`, `libarmnn-cpuref-backend22` — on amd64 and arm64, matching the API era this subplugin targets. This PR:

- installs those packages in `ubuntu_clean_meson_build.yml` (both `ubuntu-22.04` and `ubuntu-22.04-arm` matrix jobs), and
- passes `-Darmnn-support=enabled` to `meson setup`, so if detection ever breaks, the build **fails loudly** (meson.build errors on enabled-but-missing features) instead of silently skipping the subplugin again.

With the feature available, `meson test` builds and runs `unittest_filter_armnn`. The TfLite parser covers all test models (`add.tflite`, mobilenet); the Caffe parser remains optional — those tests handle its absence via the `-EPERM` path. The subplugin's default backend is `CpuRef`, which works on both architectures.

## Scope notes

- CI-only change; no source, packaging, or release-artifact changes.
- Making `nnstreamer-armnn` an actual Ubuntu PPA package (adding armnn to `control.ubuntu.ppa`) is a separate product decision, intentionally not part of this PR.
- Tizen/GBS remains disabled since Tizen no longer packages armnn.

### DIFF
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -31,11 +31,12 @@ jobs:
         sudo apt-get install -y liborc-0.4-dev flex bison libopencv-dev pkg-config python3-dev python3-numpy python3
         sudo add-apt-repository -y ppa:nnstreamer/ppa && sudo apt-get update && sudo apt-get install -y ssat libpaho-mqtt-dev
         sudo apt-get install -y valgrind gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base libgtest-dev libpng-dev libc6-dbg binutils-x86-64-linux-gnu-dbg valgrind-dbg
+        sudo apt-get install -y libarmnn-dev libarmnntfliteparser-dev libarmnn-cpuref-backend22
         pip install meson ninja
     - name: build and unit test
       if: env.rebuild == '1'
       run: |
-        meson setup build/
+        meson setup build/ -Darmnn-support=enabled
         meson compile -C build/
 
         sudo mkdir -m 777 /cores
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -9,6 +9,7 @@ jobs:
     name: Build with meson and test with Valgrind in Ubuntu
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-22.04, ubuntu-22.04-arm ]
 
@@ -32,6 +33,8 @@ jobs:
         sudo add-apt-repository -y ppa:nnstreamer/ppa && sudo apt-get update && sudo apt-get install -y ssat libpaho-mqtt-dev
         sudo apt-get install -y valgrind gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base libgtest-dev libpng-dev libc6-dbg binutils-x86-64-linux-gnu-dbg valgrind-dbg
         sudo apt-get install -y libarmnn-dev libarmnntfliteparser-dev libarmnn-cpuref-backend22
+        # The armnn subplugin auto-selects the NEON-accelerated CpuAcc backend on arm.
+        if [ "$(dpkg --print-architecture)" = "arm64" ]; then sudo apt-get install -y libarmnn-cpuacc-backend22; fi
         pip install meson ninja
     - name: build and unit test
       if: env.rebuild == '1'
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -247,7 +247,7 @@ if gtest_dep.found()
       install_dir: unittest_install_dir
     )
 
-    test('unittest_filter_armnn', unittest_filter_armnn, env: testenv)
+    test('unittest_filter_armnn', unittest_filter_armnn, timeout: 120, env: testenv)
   endif
 
   # Lua unittest
--- a/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
+++ b/tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc
@@ -128,12 +128,13 @@ TEST (nnstreamerFilterArmnn, getDimension)
   ret = sp->open (&prop, &data);
   EXPECT_EQ (ret, 0);
 
+  /** add.tflite tensors are rank 1; unused dimensions are reported as 0 */
   info.num_tensors = 1;
   info.info[0].type = _NNS_FLOAT32;
   info.info[0].dimension[0] = 1;
-  info.info[0].dimension[1] = 1;
-  info.info[0].dimension[2] = 1;
-  info.info[0].dimension[3] = 1;
+  info.info[0].dimension[1] = 0;
+  info.info[0].dimension[2] = 0;
+  info.info[0].dimension[3] = 0;
 
   /** get input/output dimension successfully */
   ret = sp->getInputDimension (&prop, &data, NULL);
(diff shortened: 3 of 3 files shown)


myungjoo:

This review (and the finding below) was produced by an automated Claude review agent; I'm relaying it here as-is.

**Summary:** the scope/design of this CI-only change is sound, but the live CI runs on the current HEAD (`794af4b3`) show a reproducible failure that means the PR's own stated goal — get the armnn subplugin/tests building and passing in at least one CI job — is not yet demonstrated.

### 1. (Blocking) `unittest_filter_armnn` fails to `open()` the model on `ubuntu-22.04-arm`, and the default `fail-fast` matrix behavior cancels the amd64 leg before it can finish

Verified in the run for this HEAD: https://github.com/nnstreamer/nnstreamer/actions/runs/32920501197
- On `ubuntu-22.04-arm` (job `98033009774`), `sp->open()` returns `-22` (`-EINVAL`) for `add.tflite` with the default CpuRef backend, immediately (<1ms). This fails 6 of 13 gtest cases (`openClose01_n`, `getDimension`, `invoke00`, `invoke02_n`, `invoke03_n`, `invoke04_n`), plus cascading garbage-value assertions in the ones that read uninitialized `res` after the failed open.
- The armnn dependency-detection path is identical between arches (`Run-time dependency armnn found: NO (tried pkg-config and cmake)` → raw `cxx.find_library` + header check fallback on both), yet the previous run's amd64 leg (`98026981624`, commit `f9e82b3a`) got past `open()` fine and only failed the two now-fixed dimension assertions. That points to an arm64-specific ArmNN CpuRef-backend runtime-loading/packaging issue (likely `libarmnn-cpuref-backend22` on aarch64), not a bug in the test or nnstreamer code.
- Because `.github/workflows/ubuntu_clean_meson_build.yml`'s `strategy` doesn't set `fail-fast: false`, GitHub Actions' default `fail-fast: true` cancels the sibling matrix job once one leg fails. In both of this PR's runs so far, the still-running leg got cancelled before producing a result (run `32918453151`: arm64 cancelled; run `32920501197`: amd64 cancelled). So neither run actually shows a clean pass on any leg for the final diff.
- Net effect if merged as-is: this workflow will likely stay red on every future PR to `main` because of the arm64 leg, and — more importantly — it will also prevent the pre-existing, unrelated amd64-only "SSAT run with Valgrind on decoder-bounding-box" and "GTEST run with Valgrind on a case" steps from ever completing, since those steps only run when `matrix.os == 'ubuntu-22.04'` and fail-fast cancels that job whenever arm64 fails first. That's a regression to existing, previously-working CI coverage, not just a gap in the new feature.
- Suggestions (pick one): root-cause/fix the arm64 CpuRef backend loading problem; or don't pass `-Darmnn-support=enabled` on the `ubuntu-22.04-arm` leg until it's fixed (leave it `auto` there); or add `strategy.fail-fast: false` so one broken/flaky leg can't take down the passing leg and the unrelated valgrind/SSAT coverage with it.

### 2. (Looks correct on inspection, but not yet confirmed green) Updated dimension expectations

Cross-checked the new expectations in `tests/nnstreamer_filter_armnn/unittest_filter_armnn.cc` — `[1,0,0,0]` for `add.tflite` (rank 1) and `[1001,1,0,0]` for the mobilenet output (rank 2) — against `ArmNNCore::setTensorProp` in `ext/nnstreamer/tensor_filter/tensor_filter_armnn.cc` (~lines 535-547), which reverse-copies `arm_shape` into `dimension[0..num_dim-1]` and zero-fills `dimension[num_dim..NNS_TENSOR_RANK_LIMIT-1]`. The arithmetic checks out and matches commit `0a05b79a`'s "fill 0 dim at nth info" semantics. That said, because of issue #1, the amd64 leg where these two assertions live never actually got to run to completion in the latest CI run, so this fix is verified by code inspection only, not by a green CI run yet.

### 3. (Looks fine) Scope, timeout, hard-error behavior

- The diff is minimal and correctly scoped: only the workflow YAML, `tests/meson.build`, and the one test file are touched; no other subplugin, the Tizen spec, or `debian/control.ubuntu.ppa` are touched, matching the PR's stated CI-only intent. No ARCHITECTURE/API docs need updating.
- `timeout: 120` on `unittest_filter_armnn` matches existing precedent in the same file (`unittest_src_iio`, `unittest_filter_custom` already use `timeout: 120`) and is a reasonable margin for a model-load + CpuRef-inference test.
- The "enabled → hard error if not found" behavior was confirmed by reading meson.build's generic feature-registration loop: `armnn-support` only supplies `extra_deps: [armnn_dep]` (no `target`), so it flows through the generic `_available` check, and `get_option(feature_name).enabled() and not _available` triggers `error(...)` when `armnn_dep` isn't found. So a future detection regression will fail the build loudly instead of silently disabling the subplugin again, as intended.

### Overall
Design direction is right, and keeping this in DRAFT/DO-NOT-MERGE for now looks appropriate: the arm64 leg needs to actually pass (or be excluded/allowed to fail without taking down the amd64 leg) before this can merge, since as of the current HEAD it does not pass and it's actively preventing the amd64 leg — and the pre-existing valgrind/SSAT steps on it — from completing.

myungjoo:

Re-review of `b43d8c06` against the prior blocking review on this PR.

**1. arm64 `-EINVAL` blocker: fixed, and the fix is correct.**
`ArmNNCore::getBackend()` (`ext/nnstreamer/tensor_filter/tensor_filter_armnn.cc:363-378`) maps `ACCL_CPU_NEON` (and its `default:` fallthrough) to `armnn::Compute::CpuAcc`, while `armnn_accl_support[]` (line 54-56) lists `ACCL_CPU_NEON_STR` first as "ACCL for default and auto config". Cross-checked against `parse_accl_hw_util`/`parse_accl_hw_fill` in `gst/nnstreamer/tensor_filter/tensor_filter_common.c` (~2780-2930): on `__aarch64__`, when no `accl_str` is given, `sup_accl[0]` (NEON) becomes the auto accelerator whenever `cpu_neon_accel_available()` reports NEON present — which is true on real arm64 hardware. So the diagnosis is right: auto-detect really does resolve to NEON → `CpuAcc` on arm64, and only `libarmnn-cpuref-backend22` was installed before this commit. Installing `libarmnn-cpuacc-backend22` on that leg directly addresses the root cause (not a workaround).

**2. Confirmed by live CI, not just by inspection.** Pulled the logs for run `32921378728` directly:
- `ubuntu-22.04-arm` (job `98035473282`): `libarmnn-cpuacc-backend22` is installed (log shows it fetched/unpacked only on this leg, absent from the amd64 leg's log — the `dpkg --print-architecture = "arm64"` guard works as intended). `unittest_filter_armnn` → `OK 0.07s`, all 13/13 gtest cases pass, `invokeAdvanced` in 39ms (CpuAcc/NEON), including `EXPECT_EQ(max_idx, 951)`.
- `ubuntu-22.04` (job `98035473381`): `libarmnn-cpuacc-backend22` correctly *not* installed. `unittest_filter_armnn` → `OK 8.88s`, 13/13 pass, `invokeAdvanced` 8865ms (CpuRef), same argmax-951 assertion passes. 8.865s leaves comfortable headroom under the new 120s meson timeout.
- Both legs completed independently (no cancellation) — `fail-fast: false` is doing its job.
- All 18 PR checks are green on `b43d8c06`.

This also resolves the previous review's item #2 ("dimension expectations look correct but not yet confirmed by green CI") — the rank-aware `[1,0,0,0]` / `[1001,1,0,0]` expectations from `794af4b3` are now exercised and passing on both backends.

**3. `fail-fast: false` and the `dpkg --print-architecture` guard: no side effects found.** The matrix here only has two legs (`ubuntu-22.04`, `ubuntu-22.04-arm`) local to this one job; `fail-fast: false` doesn't affect any other workflow's cancellation behavior, only means both legs of this job now run to completion independently on failure — which is exactly what was needed to stop arm64 failures from taking down the amd64-only valgrind/SSAT steps. The apt-get architecture guard is validated empirically by the logs above (installed on arm64 only), and needs no `apt-get update` of its own since it runs in the same shell after the preceding `apt-get update`.

**Overall:** The previously blocking issue is resolved at the root cause, verified with a green CI run on the current HEAD, and no new issues were introduced by the fix commit. All three commits (`f9e82b3a`, `794af4b3`, `b43d8c06`) are minimal, correctly scoped to CI-only changes, and consistent with the PR's stated goal. This PR looks mergeable.

PR #4868, opened 2026-08-26
[CI] Guard the check-rebuild self-test against stale PR branches

myungjoo:
## Problem

For `pull_request` events, GitHub takes the **workflow definition (step list) from the merge ref**, while `static.check.yml` checks out `${{ github.event.pull_request.head.sha }}` — the PR head tree. When main gains a new step that invokes a repository script (as happened with the check-rebuild self-test, b57d9654), every PR branched before that change fails the Static checks job with exit 127 (`No such file or directory`): the new step runs, but the script it invokes is not in the head tree.

Observed in practice on #4863, which had to rebase to get past it. Until rebased or updated, **every stale open PR fails Static checks the same way.**

## Fix

Run the self-test only if the script exists in the checked-out tree, with a skip notice otherwise. Skipping is semantically correct — a head tree that predates the self-test has no self-test target to verify — unlike the changed-file-based lint checkers, which are unaffected because their scripts long predate any open PR. A comment in the workflow documents that any future step running a repository script unrelated to `$changed_file_list` needs the same guard.

## Alternative considered (deferred)

Checking out the merge ref into a separate directory and running **all** repository scripts from there would close the entire hazard class (including step-argument vs. script-version mismatches), but it touches every checker step and changes which tree the scripts come from. If new checker steps become frequent, that restructure is the better long-term shape; for now the one-step guard fixes the only step with this failure mode at minimal risk.

### DIFF
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -146,11 +146,12 @@ jobs:
           bash .github/workflows/static.check.scripts/signed-off-by.sh ${{ github.event.pull_request.commits }}
       - name: /Checker/ check-rebuild action self-test
         # The step list comes from the merge ref while the checkout above is
-        # the PR head, so a branch older than this script would fail with
-        # "No such file or directory". Skip in that case: the head tree then
-        # has no self-test target to verify. Any future step that runs a
-        # repository script unrelated to $changed_file_list needs the same
-        # existence guard.
+        # the PR head, so a tree that lacks this script - whether the branch
+        # predates it or a rename/deletion left it out without updating this
+        # step - would fail with "No such file or directory". Skip only when
+        # this PR did not remove it (see the elif below); any future step
+        # that runs a repository script unrelated to $changed_file_list
+        # needs the same guard.
         run: |
           selftest=.github/actions/check-rebuild/test_check_if_rebuild_requires.sh
           if [ -f "$selftest" ]; then
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -145,8 +145,18 @@ jobs:
         run: |
           bash .github/workflows/static.check.scripts/signed-off-by.sh ${{ github.event.pull_request.commits }}
       - name: /Checker/ check-rebuild action self-test
+        # The step list comes from the merge ref while the checkout above is
+        # the PR head, so a branch older than this script would fail with
+        # "No such file or directory". Skip in that case: the head tree then
+        # has no self-test target to verify. Any future step that runs a
+        # repository script unrelated to $changed_file_list needs the same
+        # existence guard.
         run: |
-          bash .github/actions/check-rebuild/test_check_if_rebuild_requires.sh
+          if [ -f .github/actions/check-rebuild/test_check_if_rebuild_requires.sh ]; then
+            bash .github/actions/check-rebuild/test_check_if_rebuild_requires.sh
+          else
+            echo "Skipped: this PR branch predates the check-rebuild self-test."
+          fi
       - name: /Checker/ shellcheck for shell scripts
         # Originally from "pr-prebuild-shellcheck"
         run: |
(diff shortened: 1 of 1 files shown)


myungjoo:
*(This comment relays a review written by an automated code-review agent, not by a human reviewer.)*

Reviewed the diff, the workflow history, and the live Static checks run for this PR.

**Summary: looks correct and appropriately scoped. No blocking issues.**

Verification performed:
- Confirmed `pull_request` step-list-from-merge-ref vs. checkout-from-head-sha behavior matches the PR's description (this is the documented GitHub Actions quirk also described in `actions/checkout#2035`: "`ref: head.sha` has no effect on the GHA workflow code, which is always as-merged"). The commit message / PR body's technical explanation is accurate.
- `bash -n` on the new step body: syntax OK. `python -c "import yaml; yaml.safe_load(...)"` on the full workflow file: parses fine (24 steps in `simple_script_checkers`, as expected).
- Checked the actual `Static checks` run for this PR (https://github.com/nnstreamer/nnstreamer/actions/runs/32963120827/job/98159664319): since this branch is newer than `b57d9654`, the guard took the "script exists" branch and the real self-test executed and passed — confirms the happy path isn't accidentally short-circuited. Job conclusion: `pass` (34s).
- Confirmed via `git log -- .github/actions/check-rebuild` and `git log -- .github/workflows/static.check.scripts` that the changed-file-based checker scripts predate `b57d9654` (the self-test commit), supporting the PR body's claim that they're currently unaffected by this failure mode. Scoping the fix to just the self-test step (rather than the broader merge-ref restructure described as "deferred") is reasonable for now, and the added comment correctly warns future authors of unconditional-script steps about the same hazard.

One non-blocking gap worth being aware of (severity: **minor**, not a request to change this PR):

The guard is a plain `[ -f script ]` check on the head tree, so it can't distinguish two different situations that both result in "file missing":
1. The branch predates `b57d9654` (the intended, "semantically correct" skip case).
2. A PR branched *after* `b57d9654` deletes/renames `test_check_if_rebuild_requires.sh` in its own commits (e.g. accidental deletion during a refactor).

In case 2 the step would print "Skipped: this PR branch predates the check-rebuild self-test." and pass, even though the message is factually wrong and the deletion is a real regression. I checked and no other workflow in the repo currently exercises `test_check_if_rebuild_requires.sh` (only `action.yml` / `check_if_rebuild_requires.sh` are consumed by the build-gating workflows like `android.yml`, `gbs_build.yml`, `macos.yaml`, etc.), so nothing else in CI would catch that specific deletion — it would go entirely silent. A `git merge-base`-based check (does the file exist at the PR's fork point, not just at HEAD) would close this gap, but given this is a low-probability edge case and a net improvement over the current unconditional failure, I don't think it should block this PR — just flagging it as a possible fast-follow.

**ok-to-merge: yes**, from a static-check/CI-correctness standpoint.

myungjoo:
Addressed the minor review note in 5b63ec0: the guard now distinguishes the two missing-script cases. If this PR's own commit range deletes the self-test script (checked with the same `git show --diff-filter=D` pattern the changed-file-list step uses), the step fails with an explicit error; the silent skip remains only for branches that genuinely predate the script.

myungjoo:
Both points addressed in 15f3627:

1. **Rename escape** — added `--no-renames` so a moved script decomposes into D+A and hits the deletion branch instead of the stale-branch skip, preserving the invariant exactly as you phrased it ('absent from HEAD and not removed by this PR ⇒ stale branch'). The error message now also hints at updating the step when the script moved.
2. **`grep -qxF`** — applied.

Also reworded the skip message to state only what the guard has established (absent and not removed by this PR) rather than asserting 'predates'.

myungjoo:
Addressed the cosmetic note from the approving review too (6d6d21c): reworded the block comment to describe the actual guard condition (rename/deletion without updating this step, not just 'older than the script'), since we're keeping the PR self-contained rather than leaving even trivial follow-ups.

PR #4871, opened 2026-08-26
[Edge] Add custom property (takeover of #4700)

myungjoo:
Takeover of #4700 by @gichan-jang, which has been inactive since March 2025 with changes requested and now conflicts with main. The original commit is preserved with the author's authorship and sign-off; fixes are stacked on top.

**1. [Edge] Add custom property** (author: @gichan-jang, rebased)
Adds the `custom-props` property to edgesink/edgesrc: comma-separated `key:value` options applied to the nnstreamer-edge handle at start(). The unrelated `nns_edge_connect()` call added to `gst_edgesink_start()` is dropped during the rebase: edgesink is a PUB node and the call broke the NULL-to-READY transition when the peer is not up yet (the regression flagged in the original review).

**2. [Edge] Harden custom-props parsing with a common helper**
Replaces the duplicated per-element parsers with `gst_edge_parse_custom_props()` in edge_common, fixing the review findings:
- Heap OOB read on empty tokens (e.g. trailing comma): `g_strsplit("")` returns an empty vector, so `str_op[1]` read out of bounds.
- Values containing `:` were silently truncated (e.g. `QUEUE_SIZE:10:OLD`); now split on the first `:` only.
- Whitespace is stripped; malformed tokens and rejected keys are logged and skipped instead of silently ignored.
- Adds example launch lines to the element boilerplates, as requested in the original review.

**3. [Test/Edge] Exercise custom-props parsing at element start**
The original tests only covered the GObject set/get round trip; the parser never ran. New state-change tests apply custom-props through start() and verify values on the edge handle via `nns_edge_get_info()` (PEER_ADDRESS through the custom test library), plus malformed-input tests (colon-less token, empty key/value, doubled/trailing commas) confirming no crash, start not blocked, and valid tokens still applied. `unittest_edge` now links libnnstreamer-edge (a regular library; the gstedge plugin module remains unlinked).

**4. [Edge] Surface custom-props parse failures on the element bus**
Posts `GST_ELEMENT_WARNING (RESOURCE, SETTINGS)` when some options were not applied, so applications see the failure on the bus rather than in console logs only. The helper's return-value contract is specified in its doxygen, and the property blurb notes when options are applied.

**5. [Test/Edge] Verify custom-props warnings on the pipeline bus**
Asserts that rejected options post that warning — matched on domain/code, the posting element's name, and the message text — and that fully valid options post none, so the warning path itself is covered by CI.

**6. [Edge] Keep custom-props values out of logs and the bus** (+ follow-up commit)
`custom-props` is the documented place for options of a custom connection library, so a value may be a token or credential. A rejected option is now named by its key — or by its position when the separator was mistyped and no key can be parsed out of it — and the bus warning no longer echoes the property string. Also documents that custom-props is applied after the dedicated properties and therefore overrides a duplicated key, covers the branch where a well-formed option is refused by `nns_edge_set_info()`, and switches the edgesink malformed-input test to the custom connection library so it no longer opens a real TCP listener.

Local test: `unittest_edge` 18/18 passed; gst-indent and `clang-format --dry-run -Werror` clean on the changed code.

When this PR is merged, #4700 will be commented and closed.

**Self evaluation:**
1. Build test: [x] Passed [ ] Failed [ ] Skipped
2. Run test: [x] Passed [ ] Failed [ ] Skipped

### DIFF
--- a/gst/edge/edge_common.c
+++ b/gst/edge/edge_common.c
@@ -63,22 +63,23 @@ gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props)
       const gchar *key = g_strstrip (kv[0]);
       const gchar *value = g_strstrip (kv[1]);
 
-      if (*key && *value) {
+      if (!*key) {
+        nns_logw ("Ignored a custom property with an empty key.");
+        all_set = FALSE;
+      } else if (!*value) {
+        nns_logw ("Ignored custom property '%s' with an empty value.", key);
+        all_set = FALSE;
+      } else {
         int ret = nns_edge_set_info (edge_h, key, value);
 
         if (NNS_EDGE_ERROR_NONE != ret) {
-          nns_logw ("Failed to set custom property '%s:%s' (error %d).",
-              key, value, ret);
+          nns_logw ("Failed to set custom property '%s' (error %d).", key, ret);
           all_set = FALSE;
         }
-      } else {
-        nns_logw ("Ignored custom property token '%s' with empty key or value.",
-            tokens[i]);
-        all_set = FALSE;
       }
     } else if (kv[0] && *g_strstrip (kv[0])) {
-      nns_logw ("Ignored malformed custom property token '%s'. "
-          "Expected key:value form.", tokens[i]);
+      nns_logw ("Ignored malformed custom property '%s'. "
+          "Expected key:value form.", kv[0]);
       all_set = FALSE;
     }
     g_strfreev (kv);
--- a/gst/edge/edge_common.c
+++ b/gst/edge/edge_common.c
@@ -64,7 +64,7 @@ gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props)
       const gchar *value = g_strstrip (kv[1]);
 
       if (!*key) {
-        nns_logw ("Ignored a custom property with an empty key.");
+        nns_logw ("Ignored custom property #%u with an empty key.", i + 1);
         all_set = FALSE;
       } else if (!*value) {
         nns_logw ("Ignored custom property '%s' with an empty value.", key);
@@ -78,8 +78,8 @@ gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props)
         }
       }
     } else if (kv[0] && *g_strstrip (kv[0])) {
-      nns_logw ("Ignored malformed custom property '%s'. "
-          "Expected key:value form.", kv[0]);
+      nns_logw ("Ignored malformed custom property #%u. "
+          "Expected key:value form.", i + 1);
       all_set = FALSE;
     }
     g_strfreev (kv);
--- a/gst/edge/edge_common.c
+++ b/gst/edge/edge_common.c
@@ -15,6 +15,7 @@
 #endif
 
 #include "edge_common.h"
+#include "nnstreamer_log.h"
 
 /**
  * @brief register GEnumValue array for edge protocol property handling
@@ -39,4 +40,50 @@ gst_edge_get_connect_type (void)
   }
 
   return protocol;
+}
+
+/**
+ * @brief Parse custom properties and set them on the edge handle.
+ */
+gboolean
+gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props)
+{
+  gchar **tokens;
+  guint i;
+  gboolean all_set = TRUE;
+
+  g_return_val_if_fail (edge_h != NULL, FALSE);
+  g_return_val_if_fail (custom_props != NULL, FALSE);
+
+  tokens = g_strsplit (custom_props, ",", -1);
+  for (i = 0; tokens[i]; i++) {
+    gchar **kv = g_strsplit (tokens[i], ":", 2);
+
+    if (g_strv_length (kv) == 2) {
+      const gchar *key = g_strstrip (kv[0]);
+      const gchar *value = g_strstrip (kv[1]);
+
+      if (*key && *value) {
+        int ret = nns_edge_set_info (edge_h, key, value);
+
+        if (NNS_EDGE_ERROR_NONE != ret) {
+          nns_logw ("Failed to set custom property '%s:%s' (error %d).",
+              key, value, ret);
+          all_set = FALSE;
+        }
+      } else {
+        nns_logw ("Ignored custom property token '%s' with empty key or value.",
+            tokens[i]);
+        all_set = FALSE;
+      }
+    } else if (kv[0] && *g_strstrip (kv[0])) {
+      nns_logw ("Ignored malformed custom property token '%s'. "
+          "Expected key:value form.", tokens[i]);
+      all_set = FALSE;
+    }
+    g_strfreev (kv);
+  }
+  g_strfreev (tokens);
+
+  return all_set;
 }
--- a/gst/edge/edge_common.h
+++ b/gst/edge/edge_common.h
@@ -34,5 +34,17 @@ G_BEGIN_DECLS
  */
 GType gst_edge_get_connect_type (void);
 
+/**
+ * @brief Parse custom properties and set them on the edge handle.
+ * @param[in] edge_h The edge handle to set the parsed options on.
+ * @param[in] custom_props Comma-separated key:value pairs, e.g. "key1:val1,key2:val2".
+ *            A value may contain ':' (e.g. "QUEUE_SIZE:10:OLD"); only the first
+ *            ':' separates the key from the value. Whitespace around keys and
+ *            values is trimmed and empty tokens are ignored.
+ * @return TRUE if all non-empty tokens were parsed and set successfully.
+ *         Malformed tokens or rejected keys are logged and skipped.
+ */
+gboolean gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props);
+
 G_END_DECLS
 #endif /* __GST_EDGE_H__ */
--- a/gst/edge/edge_common.h
+++ b/gst/edge/edge_common.h
@@ -40,9 +40,12 @@ GType gst_edge_get_connect_type (void);
  * @param[in] custom_props Comma-separated key:value pairs, e.g. "key1:val1,key2:val2".
  *            A value may contain ':' (e.g. "QUEUE_SIZE:10:OLD"); only the first
  *            ':' separates the key from the value. Whitespace around keys and
- *            values is trimmed and empty tokens are ignored.
- * @return TRUE if all non-empty tokens were parsed and set successfully.
- *         Malformed tokens or rejected keys are logged and skipped.
+ *            values is trimmed.
+ * @return TRUE if every token was applied. Empty or whitespace-only tokens
+ *         (e.g. from a trailing comma) are silently ignored and do not affect
+ *         the result. Tokens without ':', with an empty key or value, or
+ *         rejected by nns_edge_set_info() are logged, skipped, and make the
+ *         return value FALSE.
  */
 gboolean gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props);
 
--- a/gst/edge/edge_common.h
+++ b/gst/edge/edge_common.h
@@ -46,7 +46,8 @@ GType gst_edge_get_connect_type (void);
  *         the result. Tokens without ':', with an empty key or value, or
  *         rejected by nns_edge_set_info() are logged, skipped, and make the
  *         return value FALSE. Option values may carry credentials of a custom
- *         connection library, so only keys are logged.
+ *         connection library, so a rejected option is named by its key, or by
+ *         its position when no key could be parsed out of it.
  */
 gboolean gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props);
 
--- a/gst/edge/edge_common.h
+++ b/gst/edge/edge_common.h
@@ -45,7 +45,8 @@ GType gst_edge_get_connect_type (void);
  *         (e.g. from a trailing comma) are silently ignored and do not affect
  *         the result. Tokens without ':', with an empty key or value, or
  *         rejected by nns_edge_set_info() are logged, skipped, and make the
- *         return value FALSE.
+ *         return value FALSE. Option values may carry credentials of a custom
+ *         connection library, so only keys are logged.
  */
 gboolean gst_edge_parse_custom_props (nns_edge_h edge_h, const gchar * custom_props);
 
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -153,7 +153,7 @@ gst_edgesink_class_init (GstEdgeSinkClass * klass)
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
   g_object_class_install_property (gobject_class, PROP_CUSTOM_PROPS,
       g_param_spec_string ("custom-props", "Custom connection props",
-          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options. The options are applied when the element starts.",
+          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options. The options are applied when the element starts, after the dedicated properties, so a key that duplicates one of them (e.g. TOPIC) overrides it.",
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
   gst_element_class_add_pad_template (gstelement_class,
@@ -416,7 +416,7 @@ gst_edgesink_start (GstBaseSink * basesink)
     if (!gst_edge_parse_custom_props (self->edge_h, self->custom_props)) {
       GST_ELEMENT_WARNING (self, RESOURCE, SETTINGS,
           ("Some custom-props options were not applied."),
-          ("custom-props: %s", self->custom_props));
+          ("The rejected keys are in the debug log."));
     }
   }
 
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -153,7 +153,7 @@ gst_edgesink_class_init (GstEdgeSinkClass * klass)
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
   g_object_class_install_property (gobject_class, PROP_CUSTOM_PROPS,
       g_param_spec_string ("custom-props", "Custom connection props",
-          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options.",
+          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options. The options are applied when the element starts.",
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
   gst_element_class_add_pad_template (gstelement_class,
@@ -412,8 +412,13 @@ gst_edgesink_start (GstBaseSink * basesink)
   if (self->topic)
     nns_edge_set_info (self->edge_h, "TOPIC", self->topic);
 
-  if (self->custom_props)
-    gst_edge_parse_custom_props (self->edge_h, self->custom_props);
+  if (self->custom_props) {
+    if (!gst_edge_parse_custom_props (self->edge_h, self->custom_props)) {
+      GST_ELEMENT_WARNING (self, RESOURCE, SETTINGS,
+          ("Some custom-props options were not applied."),
+          ("custom-props: %s", self->custom_props));
+    }
+  }
 
   nns_edge_set_event_callback (self->edge_h, _nns_edge_event_cb, self);
 
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -416,7 +416,7 @@ gst_edgesink_start (GstBaseSink * basesink)
     if (!gst_edge_parse_custom_props (self->edge_h, self->custom_props)) {
       GST_ELEMENT_WARNING (self, RESOURCE, SETTINGS,
           ("Some custom-props options were not applied."),
-          ("The rejected keys are in the debug log."));
+          ("The rejected options are in the debug log."));
     }
   }
 
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -10,6 +10,22 @@
  * @bug     No known bugs
  *
  */
+
+/**
+ * SECTION:element-edgesink
+ *
+ * Publish incoming streams to the connected edge devices.
+ * <refsect2>
+ * <title>Example launch line</title>
+ * gst-launch-1.0 videotestsrc ! edgesink port=0
+ * gst-launch-1.0 videotestsrc ! edgesink port=0 custom-props="QUEUE_SIZE:10:OLD"
+ *
+ * The custom-props property passes comma-separated key:value options to the
+ * nnstreamer-edge handle, e.g. queue policy or options of a custom connection
+ * library given by custom-lib. Only the first ':' separates a key from its
+ * value, so a value itself may contain ':'.
+ * </refsect2>
+ */
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
@@ -241,11 +257,9 @@ gst_edgesink_set_property (GObject * object, guint prop_id,
       self->custom_lib = g_value_dup_string (value);
       break;
     case PROP_CUSTOM_PROPS:
-    {
       g_free (self->custom_props);
       self->custom_props = g_value_dup_string (value);
       break;
-    }
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
@@ -353,23 +367,6 @@ _nns_edge_event_cb (nns_edge_event_h event_h, void *user_data)
 }
 
 /**
- * @brief Parse edge custom properties.
- */
-static void
-_edgesink_parse_custom_props (GstEdgeSink *self)
-{
-  gchar **str_ops = g_strsplit_set (self->custom_props, ",", -1);
-  guint i, num = g_strv_length (str_ops);
-
-  for (i = 0; i < num; i++) {
-    gchar **str_op = g_strsplit (str_ops[i], ":", -1);
-    nns_edge_set_info (self->edge_h, str_op[0], str_op[1]);
-    g_strfreev (str_op);
-  }
-  g_strfreev (str_ops);
-}
-
-/**
  * @brief start processing of edgesink
  */
 static gboolean
@@ -416,7 +413,7 @@ gst_edgesink_start (GstBaseSink * basesink)
     nns_edge_set_info (self->edge_h, "TOPIC", self->topic);
 
   if (self->custom_props)
-    _edgesink_parse_custom_props (self);
+    gst_edge_parse_custom_props (self->edge_h, self->custom_props);
 
   nns_edge_set_event_callback (self->edge_h, _nns_edge_event_cb, self);
 
--- a/gst/edge/edge_sink.c
+++ b/gst/edge/edge_sink.c
@@ -43,6 +43,7 @@ enum
   PROP_WAIT_CONNECTION,
   PROP_CONNECTION_TIMEOUT,
   PROP_CUSTOM_LIB,
+  PROP_CUSTOM_PROPS,
 
   PROP_LAST
 };
@@ -134,6 +135,10 @@ gst_edgesink_class_init (GstEdgeSinkClass * klass)
       g_param_spec_string ("custom-lib", "Custom connection lib path",
           "User defined custom connection lib path.",
           "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+  g_object_class_install_property (gobject_class, PROP_CUSTOM_PROPS,
+      g_param_spec_string ("custom-props", "Custom connection props",
+          "User defined custom connection properties. Set the options in key:value form and divide them by , for multiple options.",
+          "", G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
   gst_element_class_add_pad_template (gstelement_class,
       gst_static_pad_template_get (&sinktemplate));
@@ -180,6 +185,7 @@ gst_edgesink_init (GstEdgeSink * self)
   self->wait_connection = FALSE;
   self->connection_timeout = 0;
   self->custom_lib = NULL;
+  self->custom_props = NULL;
   self->is_connected = FALSE;
   self->edge_h = NULL;
   g_mutex_init (&self->lock);
@@ -234,6 +240,12 @@ gst_edgesink_set_property (GObject * object, guint prop_id,
       g_free (self->custom_lib);
       self->custom_lib = g_value_dup_string (value);
       break;
+    case PROP_CUSTOM_PROPS:
+    {
+      g_free (self->custom_props);
+      self->custom_props = g_value_dup_string (value);
+      break;
+    }
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
@@ -277,6 +289,9 @@ gst_edgesink_get_property (GObject * object, guint prop_id, GValue * value,
     case PROP_CUSTOM_LIB:
       g_value_set_string (value, self->custom_lib);
       break;
+    case PROP_CUSTOM_PROPS:
+      g_value_set_string (value, self->custom_props);
+      break;
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
@@ -295,6 +310,7 @@ gst_edgesink_finalize (GObject * object)
   g_clear_pointer (&self->dest_host, g_free);
   g_clear_pointer (&self->topic, g_free);
   g_clear_pointer (&self->custom_lib, g_free);
+  g_clear_pointer (&self->custom_props, g_free);
 
   g_mutex_clear (&self->lock);
   g_cond_clear (&self->cond);
@@ -334,6 +350,23 @@ _nns_edge_event_cb (nns_edge_event_h event_h, void *user_data)
   }
 
   return ret;
+}
+
+/**
+ * @brief Parse edge custom properties.
+ */
+static void
+_edgesink_parse_custom_props (GstEdgeSink *self)
+{
+  gchar **str_ops = g_strsplit_set (self->custom_props, ",", -1);
+  guint i, num = g_strv_length (str_ops);
+
+  for (i = 0; i < num; i++) {
+    gchar **str_op = g_strsplit (str_ops[i], ":", -1);
+    nns_edge_set_info (self->edge_h, str_op[0], str_op[1]);
+    g_strfreev (str_op);
+  }
+  g_strfreev (str_ops);
 }
 
 /**
@@ -381,6 +414,9 @@ gst_edgesink_start (GstBaseSink * basesink)
   }
   if (self->topic)
     nns_edge_set_info (self->edge_h, "TOPIC", self->topic);
+
+  if (self->custom_props)
+    _edgesink_parse_custom_props (self);
 
   nns_edge_set_event_callback (self->edge_h, _nns_edge_event_cb, self);
 
--- a/gst/edge/edge_sink.h
+++ b/gst/edge/edge_sink.h
@@ -55,6 +55,7 @@ struct _GstEdgeSink
   guint64 connection_timeout;
 
   gchar *custom_lib;
+  gchar *custom_props;
   gboolean is_connected;
   GMutex lock;
   GCond cond;
--- a/gst/edge/edge_src.c
+++ b/gst/edge/edge_src.c
@@ -38,6 +38,7 @@ enum
   PROP_CONNECT_TYPE,
   PROP_TOPIC,
   PROP_CUSTOM_LIB,
+  PROP_CUSTOM_PROPS,
 
   PROP_LAST
 };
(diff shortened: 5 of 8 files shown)


myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다. (The following is a relay of a review written by a review agent.)

## Verdict

No blocking issues found. The PR does what it claims — takes over #4700 (`custom-props` for edgesrc/edgesink), hardens the parser, and adds tests that actually execute the parser at `start()`. Scope is tight (only `gst/edge/*`, the edge unit test, and its meson entry). Minor, non-blocking observations below.

## Verified

- **Parser hardening is real.** The original #4700 parser did `g_strsplit (str_ops[i], ":", -1)` and read `str_op[0]`/`str_op[1]` without checking the vector length, so a trailing/doubled comma or a colon-less token was an out-of-bounds/NULL read; it also truncated values containing `:`. `gst_edge_parse_custom_props()` checks `g_strv_length() == 2`, splits on the first `:` only, and strips/validates key and value. Verified against nnstreamer-edge (`nnstreamer-edge-internal.c`, `nns_edge_set_info`): `QUEUE_SIZE` genuinely takes `limit:leaky` (e.g. `10:OLD`), so first-colon-only splitting matches the library contract.
- **Regression from #4700 removed as described.** `gst_edgesink_start()` in this PR contains no `nns_edge_connect()`; the original PR added one (SUB-connects-to-PUB inversion breaking NULL→READY when the peer is down). Confirmed absent.
- **No regression surface elsewhere.** `edge_common.h` is included only by `edge_sink.h`/`edge_src.h` (and `edge_common.c`); it already includes `nnstreamer-edge.h`, so the new prototype using `nns_edge_h` compiles for all consumers. The meson change adds `nnstreamer_edge_support_deps` (a plain pkg-config dep on libnnstreamer-edge) only to `unittest_edge`, inside a subdir gated by `nnstreamer_edge_support_is_available` — no effect on other targets or installs. Keeping the gstedge plugin itself unlinked (headers only) preserves the installed-test loader fix.
- **Cross-repo contract detection.** `sinkCustomProps`/`srcCustomProps` assert via `nns_edge_get_info()` that an unknown key (`PEER_ADDRESS`) round-trips through nnstreamer-edge's metadata fallback and that `QUEUE_SIZE:5:OLD` is accepted. If a future nnstreamer-edge change rejects unknown keys or alters `QUEUE_SIZE` parsing, these tests fail. The tests are registered with meson `test()`, so the per-PR ubuntu/valgrind jobs gate the merge.
- **Boundary cases covered at start():** whitespace around tokens, value containing `:`, colon-less token, doubled comma, whitespace-only token (` : `), empty key (`:v`), empty value (`c:`), trailing comma — for both elements, plus the check that a valid token among malformed ones is still applied.
- **Authorship/sign-off chain preserved.** Commit `177066b2` keeps Gichan Jang as author with his `Signed-off-by`, followed by the maintainer's sign-off, and the commit message documents what the takeover changed. DCO check passes.
- **Docs:** `SECTION:element-*` gtk-doc blocks with launch-line examples (matching the convention used by tensor_* elements) plus the property blurb; no external Documentation/*.md mentions these elements, so nothing else to update.

## Minor (non-blocking)

1. **Return value of `gst_edge_parse_custom_props()` is ignored at both call sites** (`edge_sink.c:416`, `edge_src.c:415`). A fully malformed `custom-props` therefore only emits `nns_logw` lines and the element starts silently. That is a defensible policy (and the header documents it), but from the callers' perspective the `all_set` accounting is currently dead code. Consider either (a) posting a `GST_ELEMENT_WARNING` on the bus when it returns FALSE so pipeline users can see the misconfiguration, or (b) a small direct unit test of the helper asserting the TRUE/FALSE return, so the return-value contract itself is covered.
2. **`all_set` accounting inconsistency (cosmetic):** a whitespace-only token (no colon) is silently skipped and leaves `all_set` TRUE, while ` : ` (colon with empty key/value) sets it FALSE. Invisible today because the return is ignored; worth aligning if item 1 is adopted.
3. **Default-value asymmetry (pre-existing pattern):** the pspec default is `""` but `init()` sets `custom_props = NULL`, so `g_object_get` before any set returns NULL rather than `""`. Same as the existing `custom-lib` property — noting only for the record.
4. **Runtime mutability:** setting `custom-props` on a running element takes effect only at the next `start()`. This matches every other property of these elements, but a `GST_PARAM_MUTABLE_READY` flag or a sentence in the blurb would make it explicit.

Overall: correct fix, well-scoped, tests are the kind that will catch both parser regressions here and behavioral drift in nnstreamer-edge. Recommend merge once CI is green (posted as a comment review; the PR author and reviewer account coincide, so a formal approval cannot be recorded).

myungjoo:

Review feedback addressed in e4702ab5:

1. **Return value ignored** — Fixed. Both call sites now post `GST_ELEMENT_WARNING (RESOURCE, SETTINGS)` on the bus when `gst_edge_parse_custom_props()` reports that some options were not applied, so applications see the failure instead of console logs only.
2. **`all_set` aggregation inconsistency** — The behavior is now specified in the helper doxygen: empty or whitespace-only tokens (e.g. a trailing comma) are silently ignored and do not affect the result, while tokens with an explicit ':' but empty key/value, colon-less tokens, and keys rejected by `nns_edge_set_info()` are logged and make the result FALSE. Rationale: a stray comma is harmless separator noise, whereas a token the user shaped as a pair signals intent and deserves a warning.
3. **pspec default `""` vs init `NULL`** — Kept as is, matching the neighboring `custom-lib` property; changing only `custom-props` would make the two adjacent properties inconsistent. Noted for a potential separate cleanup of both.
4. **Runtime property changes** — The property blurb now states that the options are applied when the element starts. The `GST_PARAM_MUTABLE_READY` flag was not added because none of the neighboring properties (host, port, custom-lib) declare mutability flags either; adding it to one property only would be misleading.

Local re-verification after the fix: build clean, unittest_edge 17/17 passed, gst-indent produces no diff on the changed code (remaining indent deltas in the touched files are pre-existing main code deliberately left untouched).

myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다. (The following is a relay of a review written by a review agent.)

## Re-review of e4702ab5 ("[Edge] Surface custom-props parse failures on the element bus")

### Prior review items — resolution check

1. **Ignored return value** — Resolved. Both call sites (`gst_edgesink_start`, `gst_edgesrc_start`) now post `GST_ELEMENT_WARNING (RESOURCE, SETTINGS, ...)` when the helper returns FALSE, which was option (a) of the original suggestion. The `all_set` accounting is no longer dead code.
2. **`all_set` aggregation inconsistency** — Resolved by specification. The header doxygen now states the exact contract, and I verified it against the implementation in `gst/edge/edge_common.c`: colon-less whitespace-only/empty tokens are silently skipped (the `kv[0] && *g_strstrip (kv[0])` guard), while ` : `-style tokens (explicit ':' with empty key/value), colon-less non-empty tokens, and `nns_edge_set_info()` rejections all log and force FALSE. Doc and code match token-class by token-class. The "separator noise vs. user-shaped pair" rationale is reasonable.
3. **pspec default `""` vs init `NULL`** — Consciously kept for consistency with the adjacent `custom-lib` property; acceptable, and correctly deferred as a joint cleanup.
4. **Runtime mutability** — Blurb now says "The options are applied when the element starts." Declining `GST_PARAM_MUTABLE_READY` to stay consistent with the neighboring properties (none declare mutability flags) is a defensible call.

### New-commit scrutiny (no new issues found)

- **`GST_ELEMENT_WARNING` in `start()` context**: safe. `start()` runs during the NULL→READY (sink) / READY→PAUSED state transitions with the element in a pipeline; posting a warning message there is standard GStreamer practice, and `gst_element_message_full` handles the no-bus case gracefully anyway. A warning does not fail the state change, so the existing "malformed props must not block start" tests (`customPropsMalformed_n`, `srcCustomPropsMalformed_n`) remain valid — `setPipelineStateSync` only checks state-change results.
- **Category choice `RESOURCE, SETTINGS`**: appropriate — `GST_RESOURCE_ERROR_SETTINGS` is the canonical code for "settings could not be applied to the resource" (the edge connection handle here), and matches how this codebase already uses the RESOURCE domain (datareposrc/sink, tensor_reposink). No i18n `_()` wrapper is the codebase norm.
- **Macro usage**: text/debug arguments correctly parenthesized; `self->custom_props` is guaranteed non-NULL inside the guard; the user string is passed as a format argument, not as the format itself, so `%` in property values is harmless.
- **Doxygen/blurb wording**: accurate and consistent between edge_sink.c and edge_src.c.
- Re-scanned the full PR diff (8 files); nothing new beyond what the previous review covered, and no interaction between the new warning and the handle-lifecycle or meson/test changes.

### Remaining (optional, non-blocking)

- The new warning path itself is not asserted by any test (a bus-watch check on the malformed-props tests could pin it down). Nice-to-have only; the behavioral tests already cover the parse outcomes.

## Verdict

All previous review items are resolved or consciously settled with sound rationale, and the new commit introduces no new problems. **No remaining blocking issues — this PR is ready to merge once CI is green.** (Posted as a comment because the PR author and this gh account coincide, so a formal approval cannot be recorded.)

myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다. (The following is a relay of a review written by a review agent.)

## Final re-review of 8335fc5a ("[Test/Edge] Verify custom-props warnings on the pipeline bus")

The one remaining nice-to-have from the previous review — "the new `GST_ELEMENT_WARNING` path itself is not asserted by any test" — is now addressed. I re-checked the new helper in depth and re-swept the whole diff after the force-push.

### `_pop_custom_props_warning()` — correctness

- **Domain/code mapping is exact.** `GST_ELEMENT_WARNING (self, RESOURCE, SETTINGS, ...)` expands to `gst_element_message_full (..., GST_MESSAGE_WARNING, GST_RESOURCE_ERROR, GST_RESOURCE_ERROR_SETTINGS, ...)`, and `gst_message_parse_warning()` hands back a `GError` carrying exactly that quark/code pair, so `g_error_matches (err, GST_RESOURCE_ERROR, GST_RESOURCE_ERROR_SETTINGS)` is the right predicate — no broader or narrower than the code under test.
- **`GError` ownership is correct.** `gst_message_parse_warning()` returns a newly-allocated `GError`; `g_clear_error (&err)` frees it and re-NULLs the variable each iteration, so the reused `err` never triggers the "GError set over the top of a previous GError" case and nothing leaks. Passing `nullptr` for the debug out-param means no debug string is allocated at all. `msg` is unreffed before the `break`, and the bus ref from `gst_element_get_bus()` is released. Clean on every path.
- **No timing race — the assertion is deterministic, not lucky.** `GST_ELEMENT_WARNING` fires inside `gst_edgesink_start()` / `gst_edgesrc_start()`, i.e. during `NULL→READY` (basesink) and `READY→PAUSED` (basesrc), both executed in the caller's thread inside `gst_element_set_state()`. `gst_bus_post()` enqueues synchronously, and the bin's sync handler (`bin_bus_handler` → `gst_bin_handle_message` → default `forward:`) re-posts warnings upward in the same thread, so the message is in the pipeline bus queue before the transition returns. `setPipelineStateSync()` additionally spins on `gst_element_get_state()` until `cur_state == state`, so by the time the pop runs, `start()` has long since completed. Also verified that `GstPipeline` clears bus flushing in the `NULL→READY` pre-chain switch (before children transition), so the warning cannot be dropped by a flushing bus.
- **Nothing else consumes the bus.** `unittest_edge.cc` installs no bus watch or signal watch, and `setPipelineStateSync()` only uses `set_state`/`get_state`. Each test builds its own pipeline via `gst_parse_launch` and unrefs it, so there is no cross-test message bleed either.
- **`EXPECT_FALSE` is not vacuous.** I checked the positive props (` PEER_ADDRESS : tcp://127.0.0.1:1883 , QUEUE_SIZE:5:OLD`) against nnstreamer-edge's `nns_edge_set_info()`: `QUEUE_SIZE` splits on the first `:` and accepts `NEW`/`OLD` case-insensitively, so `5:OLD` returns `NNS_EDGE_ERROR_NONE`; `PEER_ADDRESS` falls through to `nns_edge_metadata_set()`, also `NONE`. So `all_set` stays TRUE and the absence of a warning is a real property of the code, not an artifact. It also means these two tests would now *fail loudly* if nnstreamer-edge ever started rejecting either key — which is the desired cross-repo tripwire.
- **`EXPECT_TRUE` is deterministic too.** Token-by-token on `"foo,,topic:test_topic, : ,c:,:v,"`: `foo` (colon-less, non-empty → FALSE), `""` (silently skipped), `topic:test_topic` (applied), `" : "` (empty key+value → FALSE), `c:` (empty value → FALSE), `:v` (empty key → FALSE), trailing `""` (skipped). `all_set` is FALSE on four independent grounds, so the warning is guaranteed; the test cannot flake.

### Other checks

- **No cross-test / cross-assertion interference.** The drain loop only touches `GST_MESSAGE_WARNING`; ERROR/EOS/STATE_CHANGED messages are untouched, and no subsequent assertion in any of the four tests reads the bus.
- **Full diff re-swept (8 files).** Nothing changed beyond what the two previous reviews covered; `f2c83481` still matches its described content (warning at both call sites, return-value contract in the header doxygen, "applied when the element starts" in the blurb), and the helper in `edge_common.c` is unchanged.
- **Force-push integrity.** `git log --format='%an <%ae>%n%b' main..8335fc5a` confirms `177066b2` is still authored by Gichan Jang \<gichan2.jang@samsung.com\> with his `Signed-off-by:` followed by the maintainer's, and every later commit carries a sign-off. The DCO check is green on GitHub.
- **clang-format fixup verified.** `clang-format --dry-run -Werror` on `tests/nnstreamer_edge/edge/unittest_edge.cc` exits 0 — the formatting squashed into `2d977a87` is complete. CPP-Linter and Static checks are already green on 8335fc5a.

### Optional polish (non-blocking, no change requested)

1. The helper drains *every* queued warning up to the first match, which is more than "pop" suggests. Harmless as used (one call per test, fresh pipeline), but a word in the doc comment would prevent surprise if a future test wants to inspect other warnings after calling it.
2. `GST_MESSAGE_SRC (msg)` is not checked, so in principle any element could satisfy the `EXPECT_TRUE`. None of the elements in these pipelines emit `RESOURCE/SETTINGS`, so it is not a live concern; comparing against the edge element would make it airtight.
3. The debug string (`"custom-props: %s"`) is not asserted — checking that it echoes the offending value would pin the message content as well as its category.

## Verdict

No blocking issues. All items from the first and second reviews are resolved, the new bus-watch coverage is correct and race-free, and the force-push preserved authorship and the sign-off chain. **CI green 전제 하에 merge 가능 — no remaining issues.** (Posted as a comment rather than an approval because the PR author and this gh account coincide.)

myungjoo:

All three polish items from the latest review are addressed in e659d007 (fixed up into the bus-warning test commit):

1. **Helper drains more messages than the name implied** — The doc comment now states it consumes queued warnings up to and including the matching one, and the name reflects that it looks for a specific element's warning.
2. **`GST_MESSAGE_SRC` unchecked** — The helper now takes the element name and requires `GST_OBJECT_NAME (GST_MESSAGE_SRC (msg))` to match, so a warning from any other element cannot satisfy the assertion.
3. **Debug string not asserted** — The match now also requires the debug field to contain `custom-props: `, pinning message content in addition to the domain/code. (Checked as a substring rather than a prefix because `GST_ELEMENT_WARNING` prepends file/line/function location info to the debug string.)

The strengthened assertions are not vacuous: `unittest_edge` still passes 17/17 locally, which means the element name and the debug payload both genuinely match on the positive paths, and `EXPECT_FALSE` still holds where every option is valid. `clang-format --dry-run -Werror` is clean on the test file and the build produces no new warnings.

myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다. (The following is a relay of a review written by a review agent.)

## Review of `e659d007` (4th round, scoped to the three polish items from the previous review)

Diffed `8335fc5a..e659d007`: the only changed file is `tests/nnstreamer_edge/edge/unittest_edge.cc` (helper signature + doc, three-condition match, four call sites). No production code moved, so the scope of this round is exactly the helper.

### Previous polish items — resolution check

1. **Helper drains more than "pop" suggests** — Resolved. The doc block now reads `@brief Look for the warning that the given element posts for rejected custom-props options.` / `@note Consumes the queued warning messages up to and including the matching one.` That matches the implementation: `gst_bus_pop_filtered (bus, GST_MESSAGE_WARNING)` in a loop, `break` on the first match, full drain when there is none.
2. **`GST_MESSAGE_SRC` unchecked** — Resolved. `g_strcmp0 (GST_OBJECT_NAME (GST_MESSAGE_SRC (msg)), elem_name) == 0` pins the poster. The four call sites pass `"sinkx"` / `"srcx"`, which match the `name=` in each test's own `gst_parse_launch` string.
3. **Debug string unasserted** — Resolved. `debug && g_strstr_len (debug, -1, "custom-props: ") != nullptr` pins the payload against the call sites' `("custom-props: %s", self->custom_props)` (`edge_sink.c:419`, `edge_src.c:418`). Substring rather than prefix is the right call, since `gst_element_message_full` prepends `file(line): function (): path:\n` to the user debug.

### Specific risks raised for this round

- **Does the three-way AND make the `EXPECT_FALSE` cases vacuous?** No — every conjunct is independently anchored:
  - *Domain/code and debug substring*: the two `EXPECT_TRUE` tests (`edgeSink.customPropsMalformed_n`, `edgeCustom.srcCustomPropsMalformed_n`) exercise the identical `GST_ELEMENT_WARNING` call sites through the identical helper. If the domain, code, or debug format ever drifts, those two fail loudly and force the helper to be updated — which keeps the `EXPECT_FALSE` tests honest. A silent-pass regression would require both `_n` tests to be deleted first.
  - *Element name*: `sinkCustomProps`, `customPropsMalformed_n`, and `srcCustomProps` each already call `gst_bin_get_by_name (GST_BIN (gstpipe), "sinkx"/"srcx")` under an `ASSERT_NE (…, nullptr)` in the same test body, so a rename of the pipeline's `name=` without updating the literal aborts the test rather than making the assertion vacuous. `srcCustomPropsMalformed_n` is the `EXPECT_TRUE` side, which fails loudly on a name mismatch. There is no test where a stale name literal could pass quietly.
  - The positive props (` PEER_ADDRESS : tcp://127.0.0.1:1883 , QUEUE_SIZE:5:OLD`) still leave `all_set` TRUE, so `EXPECT_FALSE` is asserting a real property (verified in the previous round; unchanged here).
- **Can `GST_MESSAGE_SRC (msg)` be NULL?** Only reachable when the first conjunct already matched, i.e. for a `GST_RESOURCE_ERROR/SETTINGS` warning. `GST_ELEMENT_WARNING` routes through `gst_element_message_full`, which always sets `src` to the element, and no other element in these four pipelines emits that domain/code pair. `GST_OBJECT_NAME` is the idiomatic GStreamer accessor used unguarded throughout this codebase. Not a live risk; adding a NULL guard would be defensive noise.
- **`g_strstr_len (debug, -1, …)`** — correct usage for a NUL-terminated string (`-1` = search to the terminator); `debug` is NULL-guarded first, and both `g_strstr_len` and `g_clear_pointer` already appear in this test tree (`tests/common/unittest_common.cc`), so no new API-availability surface.

### Memory / lifetime on the new path

`gst_message_parse_warning (msg, &err, &debug)` allocates both out-params; `g_clear_error (&err)` and `g_clear_pointer (&debug, g_free)` run *before* the `if (matched) break;`, so both the matching and the draining iterations free everything and re-NULL for reuse. `msg` is unreffed each iteration, and the `gst_element_get_bus()` ref is dropped after the loop. No leak on any path — relevant because these tests run under the valgrind job.

### Other

`clang-format --dry-run -Werror` on `tests/nnstreamer_edge/edge/unittest_edge.cc` exits 0. Authorship and the sign-off chain are unchanged from the previous round (`177066b2` still authored by Gichan Jang with his `Signed-off-by:`); DCO is green. The remaining checks are still running on the freshly pushed head.

## Verdict

**실질적 결함 0건.** All three polish items are properly resolved, and the strengthened assertions introduce no false-negative, NULL-dereference, or leak hazard.

**CI green 전제 하에 merge 가능하며, 추가 리뷰 라운드는 불필요합니다.** (Posted as a comment rather than an approval because the PR author and this gh account coincide.)

myungjoo:

Ready for review. All 18 CI checks are green on e659d007.

One note on the earlier red run: `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` failed once on `datareposrc.fps30ReadFlexibleTensors`, which asserts `EXPECT_LT (0.8, elapsed_time)` and measured 0.787534 s — a 1.5% miss on a wall-clock threshold. It is unrelated to this PR (no datarepo code is touched, and the identical job passed on the arm runner for the same commit), and it passed on re-run. The tight timing margin in that test looks worth loosening separately.

myungjoo:

Thanks — all four items are addressed in 6b8bf37f.

**1. Option values echoed into logs and onto the bus (the pre-merge item).** Agreed, and it applied more widely than the two lines quoted: the malformed-token warnings also echoed the raw token, which leaks a value in the `:v` (empty-key) case. The parser now reports keys only —
- rejected by the library: `Failed to set custom property '%s' (error %d)` with the key alone;
- empty key: no token content logged at all;
- empty value: the key alone;
- colon-less token: the token is a bare key by definition, so it is safe to name.

The element warning's debug payload no longer carries the custom-props string; it points at the debug log instead, where only keys now appear. The helper doxygen records why values are withheld so the constraint survives future edits.

**2. Silent override of the dedicated properties.** Documented in the property blurb of both elements: "applied when the element starts, after the dedicated properties, so a key that duplicates one of them (e.g. TOPIC) overrides it." I chose documenting over rejecting: refusing those keys would break the legitimate case of setting an alias the element has no property for (`IP`, `DEST_IP`, `CAPS`), and the override is what makes custom-props a general escape hatch.

**3. `nns_edge_set_info()` failure branch untested.** New `edgeCustom.sinkCustomPropsRejected_n` uses `custom-props="ID:not_allowed"` — syntactically valid, but nnstreamer-edge explicitly refuses to update `ID` — and asserts the warning is posted. It passes, which confirms the branch fires rather than merely compiling.

**4. Real TCP listener and suite naming.** `edgeSink.customPropsMalformed_n` is now `edgeCustom.sinkCustomPropsMalformed_n` and uses `connect-type=CUSTOM custom-lib=…` like its edgesrc counterpart, so the pair is symmetric and no socket is opened.

Since the warning's identifying text moved from the debug payload to the message, `_pop_custom_props_warning()` now matches on `err->message`, keeping the assertion pinned to message content rather than just the domain/code.

Local verification: `unittest_edge` 18/18 passed, build warning-free, `clang-format --dry-run -Werror` and gst-indent clean.

myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다. (The following is a relay of a review written by a review agent.)

## Review of `6b8bf37f` ("[Edge] Keep custom-props values out of logs and the bus")

Scope of this round: the four items from the previous review, and whether the fix for item 1 introduced anything new. Diffed `e659d007..6b8bf37f` (4 files: `edge_common.c`, `edge_common.h`, `edge_sink.c`, `edge_src.c`, plus the test file). Everything settled in the earlier rounds (parser OOB, first-colon split, authorship/sign-off) was left untouched and is not re-litigated here.

### 1. Value leakage — audited exhaustively, closed on every live path

I enumerated every path in the changed code that can emit the property content, and each token class:

| token class | example | what is now emitted |
|---|---|---|
| library rejects the pair | `ID:secret` | key only — `Failed to set custom property 'ID' (error -22).` |
| empty key | `:s3cr3t` | nothing from the token — `Ignored a custom property with an empty key.` |
| empty value | `TOKEN:` | key only |
| empty / whitespace-only | trailing `,` | nothing (silently skipped) |
| colon-less, non-empty | `foo` | the token itself, as a key |
| bus message | any failure | text `Some custom-props options were not applied.`, debug `The rejected keys are in the debug log.` — no property content in either |

`grep`ing `custom_props` across `gst/edge/` confirms there is no other consumer: `set_property`/`get_property`/`finalize` do not log, and `start()` no longer passes the string into the warning. So of the six classes, five emit nothing that can be a value, and the bus message is fully sanitized. This was worth fixing beyond theory — the pre-fix CI log for `e659d007` literally contains `Ignored custom property token ':v' with empty key or value.` in the pdebuild job output, i.e. the old code really did print token content into a publicly archived log.

**Residual (non-blocking, one class):** the colon-less branch still prints the whole token, on the stated reasoning that "a colon-less token is a bare key by definition". That holds for the *intended* input shape, but the branch exists precisely because the input was **not** shaped as intended, so the code cannot actually tell a bare key from a pair typed with the wrong separator. `custom-props="ACCESS_TOKEN=s3cr3t"` (or a space, or `=` instead of `:`) is a single plausible typo away and would land the secret in the warning log — the very outcome the commit is closing everywhere else. Also worth noting that the header doxygen now asserts "only keys are logged", which is a promise this branch cannot keep in general. Logging the position instead of the content (`Ignored malformed custom property token #%u; expected key:value form.`) would close the class outright and lose nothing diagnostically — the user can count commas. Not merge-blocking, since it needs malformed input, but it is the one remaining hole in an otherwise complete fix.

Out of this PR's control, for the record: `gst-launch-1.0 -v` prints property values via deep-notify, so a credential passed through `custom-props` is visible there regardless. That is generic GStreamer behaviour affecting every string property in the tree and is not something this change can or should address.

### 2. Documenting the override instead of rejecting the keys — agreed

I checked the alternative against `nns_edge_set_info()` upstream before judging. The key space is not a clean superset of the GObject properties: `CAPS`/`CAPABILITY` and `QUEUE_SIZE` have no dedicated property at all, and `IP`/`DEST_IP` are aliases of `HOST`/`DEST_HOST`. A "reject keys that have a dedicated property" rule would therefore need alias-aware logic just to avoid false rejections, and would still amputate the legitimate use of an escape-hatch property. Documenting is the proportionate call, and "applied after the dedicated properties, so a key that duplicates one of them (e.g. TOPIC) overrides it" says exactly what happens. Last-writer-wins also matches how the surrounding `start()` code already behaves. No objection.

### 3. New `sinkCustomPropsRejected_n` — hits the intended branch, contract verified

- **It genuinely reaches the `nns_edge_set_info()` failure branch.** Checked against upstream `nnstreamer-edge/src/libnnstreamer-edge/nnstreamer-edge-internal.c`: `nns_edge_set_info()` has an explicit `else if (0 == strcasecmp (key, "ID") || 0 == strcasecmp (key, "CLIENT_ID")) { /* Not allowed key */ … ret = NNS_EDGE_ERROR_INVALID_PARAMETER; }` arm. Critically, that arm sits **before** the `NNS_EDGE_CONNECT_TYPE_CUSTOM` pass-through at the tail of the function, which only runs `if (ret == NNS_EDGE_ERROR_NONE)` — so using `connect-type=CUSTOM` does not let the custom test library intercept `ID` and turn the rejection into a success. The branch fires deterministically, not incidentally.
- **Contract stability.** `ID` is not an incidental rejection; it guards handle identity, which is why `CLIENT_ID` is refused alongside it, and `nns_edge_get_info()` exposes `ID` read-only. It is unlikely to be relaxed. If it ever were, this test fails loudly (`EXPECT_TRUE` on the warning) rather than passing silently, which is the correct failure direction for a cross-repo tripwire — the same property the earlier `PEER_ADDRESS`/`QUEUE_SIZE` assertions have.
- **CI does gate it.** `unittest_edge` is not built in the Valgrind jobs (`nnstreamer-edge found: NO` there), but the pdebuild and Tizen/GBS `unit_test 1` jobs do build and run it — I confirmed `Linking target tests/nnstreamer_edge/unittest_edge` and its runtime output in the pdebuild log of the previous head. So the new test is verified against the distro's libnnstreamer-edge, not only locally.

### 4. Malformed test moved to CUSTOM — no loss of strength

`edgeCustom.sinkCustomPropsMalformed_n` keeps every assertion it had: the same seven-token input (`foo,,topic:test_topic, : ,c:,:v,`), the same `nns_edge_get_info (…, "TOPIC", …)` / `EXPECT_STREQ ("test_topic", …)` readback proving the valid token among malformed ones is still applied, the same state-change checks, plus the bus-warning assertion. Only the transport changed, and the parser and the `TOPIC` handling are transport-independent (`TOPIC` is resolved in `nns_edge_set_info()`'s own `strcasecmp` chain, and `nns_edge_get_info()` reads `eh->topic` before the custom override is even consulted, so the lowercase-`topic` → uppercase-`TOPIC` round trip is unaffected). The suite rename to `edgeCustom` and dropping the real listener are both improvements. `port=0` is now vestigial for a CUSTOM handle (`self->port > 0` gates it), harmless.

### Helper change to `err->message` — valid, no false positive or negative

`GST_ELEMENT_WARNING(self, RESOURCE, SETTINGS, (text), (debug))` reaches `gst_element_message_full()`, which builds the `GError` as `g_error_new_literal (domain, code, sent_text)` from the **text** argument — the debug argument only ever populates the separate debug string. Since the identifying wording moved from debug to text in this commit, matching on `err->message` is the correct follow-through, not a weakening: it still pins message content on top of domain+code+element name.

- **No false positive.** All four conjuncts must hold; no other element in these pipelines posts `RESOURCE`/`SETTINGS`, and the message literal is not wrapped in `_()`, so no translation can move it out from under the substring check.
- **No false negative.** `"Some custom-props options were not applied."` contains `custom-props`; if that text ever drifts, the three `EXPECT_TRUE` tests fail loudly and force the helper to be updated, which keeps the two `EXPECT_FALSE` tests honest.
- **Lifetime.** Dropping the `debug` out-param means `gst_message_parse_warning (msg, &err, nullptr)` allocates nothing for it, so the removal of `g_clear_pointer (&debug, g_free)` is correct rather than a leak; `g_clear_error (&err)` and `gst_message_unref (msg)` still run on both the matching and the draining iterations. `err->message` is only dereferenced after `g_error_matches()` has established `err != NULL`.

## Verdict

**실질적 결함 0건.** All four items from the previous review are resolved: item 1 on every live path (with one narrow residual on malformed colon-less tokens, noted above as a suggestion, not a request), item 2 by a well-reasoned documentation choice, item 3 by a test that provably reaches the intended branch, item 4 by a symmetric, socket-free rewrite. The fix introduces no leak, no vacuous assertion, and no behavioural change beyond what is logged.

**CI green 전제 하에 merge 가능합니다.** (Posted as a comment rather than an approval because the PR author and this gh account coincide.)

myungjoo:

The residual leak in the colon-less branch is closed in 4d189fe0. The reasoning in the review is right and my earlier justification was wrong: "a colon-less token is a bare key by definition" only holds for *intended* input, and that branch is reached precisely when the input was not what the user intended — `ACCESS_TOKEN=s3cr3t` is indistinguishable from a bare key there.

Malformed options and empty-key options are now named by position rather than content:

```
Ignored malformed custom property #1. Expected key:value form.
Ignored custom property #4 with an empty key.
Ignored custom property 'c' with an empty value.
Ignored custom property #6 with an empty key.
```

That last line is the case that mattered: token 6 is `:v`, which previously printed as `':v'` and exposed the value. Options that yield a parsable key are still named by key, since that is the actionable part and carries no secret. The header doxygen now states the rule for both shapes, so it holds on every branch rather than just most of them.

Local verification on 4d189fe0: `unittest_edge` 18/18 passed, build warning-free, gst-indent clean.

myungjoo (commented):
**Automated review** — 아래 내용은 리뷰 에이전트가 작성한 리뷰를 옮겨 적은 것입니다.

Scope: 마지막 커밋 `4d189fe0` ("[Edge] Name malformed custom-props options by position") 한 건만 좁게 확인했습니다. 이전 라운드에서 이미 검증된 항목(파서 OOB, first-colon 분리, bus 경고, authorship 등)은 재검토하지 않았습니다.

**결론: 실질적 결함 0건. CI green 전제 하에 merge 가능합니다.**

### 1. Value leak 경로 전수 확인 (`gst_edge_parse_custom_props()`)

`edge_common.c:59-86`의 모든 토큰 클래스를 훑었고, 이제 어떤 로그 경로에도 option value가 포함되지 않습니다.

| # | 토큰 형태 | 예 | 로그 | value 노출 |
|---|---|---|---|---|
| 1 | 정상 key:value, `nns_edge_set_info()` 성공 | `topic:t` | 없음 | 없음 |
| 2 | 정상 key:value, `nns_edge_set_info()` 실패 | `ID:not_allowed` | key + errno (L76) | 없음 |
| 3 | 빈 key | `:v`, ` : ` | `#N`만 (L67) | **없음 (이번 커밋)** |
| 4 | 빈 value | `c:` | key만 (L70) | value 자체가 빈 문자열 |
| 5 | colon 없는 non-empty | `ACCESS_TOKEN=s3cr3t` | `#N`만 (L81) | **없음 (이번 커밋)** |
| 6 | 빈/공백 토큰 | `""`, `"  "` | 없음 (silently skipped) | 없음 |

이전 라운드의 잔여 지적(클래스 5에서 토큰 전체를 출력하여 `ACCESS_TOKEN=s3cr3t`의 값이 유출되던 문제)이 정확히 해소되었습니다. 부수적으로 클래스 3(`:v`)도 토큰 내용을 전혀 찍지 않게 되어, "값처럼 보이는 나머지 절반"이 로그로 나갈 여지까지 함께 닫혔습니다. 이제 로그에 등장할 수 있는 사용자 입력 유래 문자열은 **성공적으로 파싱된 non-empty key뿐**입니다.

`guint i` + `%u` 포맷 매칭 정상, 기존 `key`/`value` 지역 변수는 다른 분기에서 계속 쓰이므로 unused 경고 없음.

### 2. 헤더 doxygen 정합성

`edge_common.h:44-50`의 갱신된 문구 — "a rejected option is named by its key, or by its position when no key could be parsed out of it" — 는 구현과 일치합니다. key로 명명되는 경우(클래스 2, 4)와 position으로 명명되는 경우(클래스 3, 5)의 구분이 정확하고, 클래스 3의 "key를 파싱해내지 못한 경우"라는 표현도 (콜론은 있으나 key가 빈 문자열) 실제 동작을 옳게 서술합니다. 클래스 6은 바로 앞 문장("Empty or whitespace-only tokens ... are silently ignored and do not affect the result")이 이미 예외로 명시하고 있어, "Tokens without ':' ... are logged" 문장과 충돌하지 않습니다.

### 3. 인덱스 표기 `i + 1`의 의미 (non-blocking nit)

`i`는 `g_strsplit(custom_props, ",", -1)`의 인덱스이므로 `#N`은 **N번째 콤마 구분 필드**를 뜻하며, 빈 필드도 번호를 소모합니다. 반면 빈 필드는 로그에 전혀 나타나지 않으므로, 빈 토큰이 섞이면 "사용자가 눈으로 세는 옵션 번호"와 어긋납니다.

PR의 기존 테스트 입력 `"foo,,topic:test_topic, : ,c:,:v,"`가 그 예로, 실제 로그는 `#1`(`foo`), `#4`(` : `), `#6`(`:v`)이 찍히지만 비어 있지 않은 옵션만 세면 1, 3, 5번째입니다.

다만 (a) `#N`은 "콤마를 세면 기계적으로 특정 가능한 필드 번호"라는 결정적 정의를 가지고, (b) 빈 필드가 없는 통상적 입력에서는 두 세는 방식이 일치하며, (c) 대안(빈 필드를 건너뛴 순번)은 오히려 원문 문자열과의 대응을 흐리게 만듭니다. 진단 가치를 해치는 수준은 아니라고 판단합니다. 굳이 다듬는다면 메시지에 "(comma-separated field #N)" 정도를 덧붙이는 선택지가 있으나, merge를 막을 사유는 아닙니다.

또 하나의 사소한 문구 드리프트: `edge_sink.c:419` / `edge_src.c:418`의 bus warning debug 문자열이 여전히 "The rejected keys are in the debug log."입니다. 이제 클래스 3/5는 key가 아니라 position으로 기록되므로 엄밀히는 "keys or positions"에 가깝습니다. 순수 cosmetic이며 별도 수정 없이 두어도 무방합니다.

### 4. 기존 테스트/타 코드 영향

- 저장소 전체에서 변경된 두 로그 문자열(`"Ignored ... empty key"`, `"Ignored malformed custom property"`, `"Expected key:value form."`)에 의존하는 코드/테스트는 **없습니다**.
- `tests/nnstreamer_edge/edge/unittest_edge.cc`의 `_pop_custom_props_warning()`은 GStreamer **bus** 메시지만 검사하며, 매칭 조건은 `GST_RESOURCE_ERROR_SETTINGS` + element name + `err->message`의 `"custom-props"` 부분 문자열입니다. 이 커밋은 bus 메시지(`edge_sink.c` / `edge_src.c`)를 건드리지 않았으므로 `edgeSink/edgeSrc` 관련 기존 어서션은 그대로 통과합니다.
- 이번 커밋의 diff는 `nns_logw()` 포맷 문자열 2줄과 헤더 주석뿐이며, 반환값 의미(`all_set`)나 제어 흐름 변화는 없습니다. 따라서 `gst_edge_parse_custom_props()` 호출자 2곳의 동작도 불변입니다.

myungjoo:

Took the cosmetic drift too — the bus debug string said "The rejected keys are in the debug log." while some options are now named by position, so it reads "The rejected options are in the debug log." in f690fda3 (amended into the same commit). No behavior change; no test depends on that string.

On the index nit: agreed that `#N` counts comma-separated fields including empty ones, so it can differ from what a user counting only non-empty options would expect. I left it as is deliberately — `#N` is deterministic and mechanically recoverable by counting commas, whereas a "Nth non-empty option" numbering would need the reader to know which fields the parser considered empty, which is exactly what they are trying to find out.

Final local verification on f690fda3: `unittest_edge` 18/18 passed, build warning-free, clang-format clean, and gst-indent's only remaining delta in the touched files is three lines of pre-existing main code this PR does not modify (`_wait_connection` and two `nns_edge_create_handle` continuations).

myungjoo:

All 18 CI checks are green on f690fda3, and no review items remain open. Ready to merge.

For the record, two red runs along the way were both unrelated flaky tests that passed on re-run with no code change:

- `datareposrc.fps30ReadFlexibleTensors` — asserts `EXPECT_LT (0.8, elapsed_time)` and measured 0.787534 s, a 1.5% miss on a wall-clock threshold.
- `nnstreamerFilterOnnxRuntime.quantModelLaunch00` — sets the pipeline to PLAYING and then straight to NULL with no wait, and expects at least one buffer to have reached the sink; `setPipelineStateSync()` only guarantees the state change, not that a frame traversed pngdec → videoconvert → videoscale → tensor_converter → two transforms → ONNX inference → sink.

Neither touches code this PR modifies, and the immediately preceding commit e659d007 passed both jobs. Both look worth loosening separately so unrelated PRs stop going red on them.

PR #4872, opened 2026-08-26
[Test/datarepo] fix flaky fps30ReadFlexibleTensors

myungjoo:
## Problem

`datareposrc.fps30ReadFlexibleTensors` fails intermittently in the Ubuntu meson CI job. Most recently on [run 32974102457](https://github.com/nnstreamer/nnstreamer/actions/runs/32974102457) for #4864, whose diff only touched unrelated test argmax helpers. The log shows:

```
Elapsed time: 0.758597 second
../tests/nnstreamer_datarepo/unittest_datareposrc.cc:668: Failure
```

## Root cause

The test asserted a hard-coded wall-clock lower bound (`EXPECT_LT (0.8, elapsed_time)`) on the `sync=true` run, which assumes the generated file always holds 30 frames (3 sources x `num-buffers=10`).

It does not. `create_flexible_tensors_test_file ()` feeds **three concurrent** `videotestsrc` into `join`, but `join` forwards events only from the pad that delivered the last buffer:

https://github.com/nnstreamer/nnstreamer/blob/main/gst/join/gstjoin.c — `forward = (pad == active_sinkpad);`

So the first EOS that arrives while its pad is active is forwarded downstream, `datareposink` finalizes the file, and the remaining sources are cut off. How many frames land in the file depends on thread scheduling. This is consistent with `join`'s documented contract — *"N streams should not operate at the same time... If one sinkpad is receiving buffer, the others should be stopped"* — which the helper does not honour, so this is a test-side issue rather than a `join` defect.

`datareposrc` timestamps frame *i* at `(i-1)/fps`, so the last of N frames renders at `(N-2)/fps`:

| frames stored | expected elapsed | vs. 0.8 s bound |
|---|---|---|
| 30 | 0.933 s | passes |
| 25 | 0.767 s | **fails** (matches the observed 0.7586 s) |

## Fix

Count the frames each run actually delivers via the `new-data` signal and derive the expectation from that instead of a magic constant:

- **synced run** must take at least 70% of the stream duration (`frames / fps`). Since the sink waits for every timestamp, elapsed is always >= `(N-2)/fps`, so this holds deterministically for any N >= 7 — and N is never below 10, because the source that emits the terminating EOS has necessarily pushed all 10 of its buffers. Machine load can only make elapsed larger, never smaller.
- **unsynced run** is now compared against the synced run instead of an absolute `0.05 s` bound, so it no longer depends on the runner being fast in absolute terms. For the typical 25-30 frame file this allows ~0.26-0.31 s where the observed value is ~0.02 s.

Both properties the test verified are preserved: `sync=true` is paced by the framerate stored in the json, and `sync=false` is far faster.

## Scope notes

- Test-only change, confined to one test function; no production code touched.
- Assertions moved after the cleanup so a failure cannot leak the pipeline.
- The deeper question of whether `join` should aggregate EOS across all pads (rather than relying on callers to serialize their sources) is left alone deliberately — it would change element behaviour for every `join` user and is a separate decision.

### DIFF
--- a/tests/nnstreamer_datarepo/unittest_datareposrc.cc
+++ b/tests/nnstreamer_datarepo/unittest_datareposrc.cc
@@ -714,7 +714,7 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
    * the expected duration from the frames actually read.
    */
   ASSERT_GT (sync_count, 0);
-  ASSERT_GT (no_sync_count, 0);
+  EXPECT_EQ (sync_count, no_sync_count);
 
   /* sync=true paces the playback with the framerate stored in the json file */
   EXPECT_GT (sync_time, (sync_count / (gdouble) fps) * 0.7);
--- a/tests/nnstreamer_datarepo/unittest_datareposrc.cc
+++ b/tests/nnstreamer_datarepo/unittest_datareposrc.cc
@@ -636,8 +636,10 @@ error:
 TEST (datareposrc, fps30ReadFlexibleTensors)
 {
   gint fps = 30;
+  gint sync_count = 0, no_sync_count = 0;
+  GCallback handler = G_CALLBACK (new_data_cb);
   guint64 start_time, end_time;
-  gdouble elapsed_time;
+  gdouble sync_time, no_sync_time;
   GstElement *tensor_sink;
   GstBus *bus;
   GMainLoop *loop;
@@ -648,6 +650,10 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   create_flexible_tensors_test_file (fps, file_index);
   GstElement *pipeline = gst_parse_launch (str_pipeline, NULL);
   ASSERT_NE (pipeline, nullptr);
+
+  tensor_sink = gst_bin_get_by_name (GST_BIN (pipeline), "tensor_sink0");
+  ASSERT_NE (tensor_sink, nullptr);
+  g_signal_connect (tensor_sink, "new-data", (GCallback) handler, &sync_count);
 
   loop = g_main_loop_new (NULL, FALSE);
   bus = gst_pipeline_get_bus (GST_PIPELINE (pipeline));
@@ -662,11 +668,9 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
 
   setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT);
   end_time = g_get_monotonic_time ();
-  elapsed_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
+  sync_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
 
-  g_print ("Elapsed time: %.6f second\n", elapsed_time);
-  EXPECT_LT (0.8, elapsed_time);
-
+  g_clear_pointer (&tensor_sink, gst_object_unref);
   g_clear_pointer (&pipeline, gst_object_unref);
   g_main_loop_unref (loop);
 
@@ -680,7 +684,9 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   g_clear_pointer (&bus, gst_object_unref);
 
   tensor_sink = gst_bin_get_by_name (GST_BIN (pipeline), "tensor_sink0");
+  ASSERT_NE (tensor_sink, nullptr);
   g_object_set (GST_OBJECT (tensor_sink), "sync", FALSE, NULL);
+  g_signal_connect (tensor_sink, "new-data", (GCallback) handler, &no_sync_count);
 
   start_time = g_get_monotonic_time ();
 
@@ -689,10 +695,7 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
 
   setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT);
   end_time = g_get_monotonic_time ();
-  elapsed_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
-
-  g_print ("Elapsed time: %.6f second\n", elapsed_time);
-  EXPECT_LT (elapsed_time, 0.05);
+  no_sync_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
 
   g_clear_pointer (&tensor_sink, gst_object_unref);
   g_clear_pointer (&pipeline, gst_object_unref);
@@ -700,6 +703,23 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
 
   g_remove ("flexible1.json");
   g_remove ("flexible1.data");
+
+  g_print ("Elapsed time: %.6f second (sync, %d frames), %.6f second (no sync, %d frames)\n",
+      sync_time, sync_count, no_sync_time, no_sync_count);
+
+  /**
+   * The number of frames in the test file is not fixed: the join element feeding
+   * datareposink forwards EOS from the pad that delivered the last buffer, so the
+   * remaining sources are cut off at a point that depends on scheduling. Derive
+   * the expected duration from the frames actually read.
+   */
+  ASSERT_GT (sync_count, 0);
+  ASSERT_GT (no_sync_count, 0);
+
+  /* sync=true paces the playback with the framerate stored in the json file */
+  EXPECT_GT (sync_time, (sync_count / (gdouble) fps) * 0.7);
+  /* sync=false reads the same frames without waiting for the clock */
+  EXPECT_LT (no_sync_time * 3, sync_time);
 }
 
 /**
--- a/tests/nnstreamer_datarepo/unittest_datareposrc.cc
+++ b/tests/nnstreamer_datarepo/unittest_datareposrc.cc
@@ -704,14 +704,14 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   g_remove ("flexible1.json");
   g_remove ("flexible1.data");
 
-  g_print ("Elapsed time: %.6f second (sync, %d frames), %.6f second (no sync, %d frames)\n",
-      sync_time, sync_count, no_sync_time, no_sync_count);
+  g_print ("Elapsed time: %.6f s (sync), %.6f s (no sync), %d frames\n",
+      sync_time, no_sync_time, sync_count);
 
   /**
-   * The number of frames in the test file is not fixed: the join element feeding
-   * datareposink forwards EOS from the pad that delivered the last buffer, so the
-   * remaining sources are cut off at a point that depends on scheduling. Derive
-   * the expected duration from the frames actually read.
+   * The number of frames in the test file is not fixed: the join element
+   * feeding datareposink forwards EOS from the pad that delivered the last
+   * buffer, so the remaining sources are cut off at a point that depends on
+   * scheduling. Derive the expected duration from the frames actually read.
    */
   ASSERT_GT (sync_count, 0);
   EXPECT_EQ (sync_count, no_sync_count);


myungjoo:

I independently verified this fix by reading `gst/join/gstjoin.c`, `gst/datarepo/gstdatareposrc.c`, and `gst/nnstreamer/elements/gsttensor_sink.c` alongside the diff. Summary of findings — no blocking issues, only low-severity optional suggestions.

**Root cause — confirmed accurate.**
- `gst_join_pad_chain()` always forwards every buffer regardless of which pad it arrives on, but sets `sel->active_sinkpad = pad` on each buffer.
- `gst_join_pad_event()` only forwards an event (`forward = (pad == active_sinkpad)`) — so whichever source's EOS happens to arrive while it is still the "most recently chained" pad gets EOS forwarded downstream, closing `datareposink` early. This exactly matches the module's own documented contract at the top of `gstjoin.c`: "Connect recently arrived buffer from N input streams to the output pad. N streams should not operate at the same time." The test's `create_flexible_tensors_test_file()` deliberately runs 3 concurrent `videotestsrc` branches into `join`, which violates that contract and is the real source of the nondeterministic frame count. The fix works around this at the test level rather than touching `join`, which is appropriate given the PR's scope.
- `gst_data_repo_src_set_timestamp()` confirms the `(N-2)/fps` last-frame PTS formula: `running_time`/`n_frame` bookkeeping yields PTS for frame index `k≥1` of `(k-1)/fps`, so the last frame (index `N-1`) gets `(N-2)/fps`. This reproduces the PR's numbers exactly (N=30 → 0.933s, N=25 → 0.767s).

**New assertions — mathematically sound.**
- `sync_time > (sync_count/fps) * 0.7`: since `sync=true` blocks on the pipeline clock, `sync_time` is floor-bounded by `(N-2)/fps` and can only be *pushed higher* by scheduling/CPU load, never lower. At the theoretical minimum `N=10` (guaranteed because whichever `videotestsrc` branch sends the EOS that gets forwarded must have already pushed all 10 of its own buffers), the floor is `8/30 ≈ 0.267s` against a required threshold of `(10/30)*0.7 ≈ 0.233s` — a real, load-independent margin, not a statistical one. This can't be flakified by a slow/oversubscribed runner (only sped up by an incorrect clock, which isn't a realistic failure mode here).
- `no_sync_time * 3 < sync_time`: at `N=10` this requires `no_sync_time < ~0.089s` for a trivial, computation-free "read small file → push buffers" pipeline — comfortable in practice, though it is a tighter absolute margin than the old fixed `0.05s`/`0.8s` pair for very small N. Given this repo's documented history of CI runner heterogeneity causing flakiness (e.g. the ONNX Runtime quantized test issue), this is worth a mention, but I could not identify a plausible mechanism, in this specific lightweight pipeline, to push `no_sync_time` anywhere near that bound. Low severity, non-blocking.

**Frame counting via `new-data` — verified accurate.**
- `gst_tensor_sink_render_buffer()` emits `new-data` unconditionally when `signal-rate` is 0 (the default), and is only invoked from `render`/`render_list`.
- `tensor_sink` never overrides `max-lateness` (stays at `GstBaseSink`'s default of -1/unlimited), so no QoS-driven drop path exists even though `qos=TRUE` is set — `qos` here only affects whether QoS events are sent upstream, not whether buffers are rendered. So neither run can under-count due to lateness-based dropping.
- `tensor_sink` does not override the `preroll` vfunc, so `GstBaseSink`'s standard preroll-then-render sequencing means the first buffer is not double-counted either. `sync_count` and `no_sync_count` should therefore both equal the true frame count `N` written to the shared `flexible1.data` file (both playback pipelines read the exact same file).

**Memory/ref-counting — no leak introduced.**
- Both `tensor_sink` acquisitions via `gst_bin_get_by_name()` are unreffed via `g_clear_pointer(&tensor_sink, gst_object_unref)` before the corresponding pipeline is unreffed, in both runs — balanced.
- The diff actually improves robustness by adding a missing `ASSERT_NE (tensor_sink, nullptr)` after the second `gst_bin_get_by_name()` call, which previously had no null-check.
- Moving the `EXPECT_GT`/`EXPECT_LT` calls to after cleanup is safe: the `ASSERT_GT (sync_count, 0)`/`ASSERT_GT (no_sync_count, 0)` guards run after all `gst_object_unref`/`g_remove` calls, so an early return on assertion failure leaks nothing.

**Property under test — not weakened.** The test still verifies that `sync=true` paces playback by the stored framerate and `sync=false` reads far faster; if anything it is now a more faithful check since it no longer assumes a fixed, sometimes-wrong frame count.

**Scope/size — appropriate.** Only `tests/nnstreamer_datarepo/unittest_datareposrc.cc` is touched (29 insertions / 9 deletions, one test function). No other module is touched, and no attempt is made to "fix" `join` itself, which is reasonable given the stated topic.

**Minor, non-blocking suggestions (optional):**
1. Since both pipeline runs read the identical `flexible1.data`/`flexible1.json`, consider adding `EXPECT_EQ (sync_count, no_sync_count)` as an extra sanity check independent of timing.
2. Given the tight-at-minimum-N margin noted above, consider whether `0.7` and `3x` still leave enough headroom on the project's more heavily loaded/heterogeneous CI runners; this is speculative and I have no evidence it will actually flake.

Overall assessment: the root-cause analysis matches the code, the new assertions are logically and mathematically sound (the sync-time bound is floor-guaranteed rather than statistical), no leaks or scope issues were found. I'd consider this mergeable as-is.

myungjoo:
Addressed the review feedback in `0a146ae4`:

- **Nit (frame counts must match)**: adopted. Both runs read the same `flexible1.data`, and the reviewer confirmed there is no drop path (`tensor_sink` does not override `max-lateness`, so it stays -1), so the counts are asserted equal now instead of only checking the second one is non-zero.

- **Low (margin of `no_sync_time * 3 < sync_time` at the minimum frame count)**: kept as is, because the new bound is strictly *looser* than the one that has been passing all along. The replaced assertion was an absolute `no_sync_time < 0.05 s`. The new bound allows `sync_time / 3`, which is:

  | frames | new budget | old budget |
  |---|---|---|
  | 10 (theoretical minimum) | 0.089 s | 0.05 s |
  | 25 (observed in the failing run) | 0.256 s | 0.05 s |
  | 30 (nominal) | 0.311 s | 0.05 s |

  So for every possible frame count the unsynced run now has more headroom than before, against an observed value of ~0.02 s. This side of the test has never flaked, and the change cannot make it more likely to.

myungjoo:

Re-review of `0a146ae4` on top of `64aa8c4e`. I re-read the final `TEST (datareposrc, fps30ReadFlexibleTensors)` body and `gst/join/gstjoin.c` independently (not just the diff) to check both open items.

**① `EXPECT_EQ (sync_count, no_sync_count)` — correctly applied and safe.**
- The commit does exactly what was asked: `ASSERT_GT (no_sync_count, 0)` → `EXPECT_EQ (sync_count, no_sync_count)`, while `ASSERT_GT (sync_count, 0)` is kept.
- Checked whether a `no_sync_count == 0` outcome (which `EXPECT_EQ` would report but not abort on) can reach undefined/divide-by-zero behavior later: the only arithmetic after this point is `sync_count / (gdouble) fps` (line ~720, `fps` is a nonzero constant 30) and `no_sync_time * 3` (line ~722) — neither divides by `no_sync_count` or `sync_count`, so a zero count cannot cause UB or a crash; it can only surface as a (correct) `EXPECT_EQ` failure. Safe.
- Semantically sound too: both runs read the identical, already-fully-written `flexible1.data`/`flexible1.json` (no concurrent multi-source race at read time — that race only exists in `create_flexible_tensors_test_file()`'s file-generation pipeline, not in the two playback runs), and `tensor_sink` never overrides `max-lateness` (stays at GstBaseSink's default -1), so there's no drop path that could make the two counts differ legitimately.

**② `EXPECT_LT (no_sync_time * 3, sync_time)` margin at minimum N — numerically confirmed, and the floor is stronger than stated.**
- Recomputed the table independently: budget = `((N-2)/fps)/3`. N=10 → 0.0889s, N=25 → 0.2556s, N=30 → 0.3111s — matches the posted numbers, all comfortably above the old absolute `0.05s` bound.
- Went further and checked whether N could ever drop below 10 (the earlier review flagged this as "theoretical minimum," worth confirming from source rather than assumption). Read `gst_join_pad_chain()`/`gst_join_pad_event()` in `gst/join/gstjoin.c`: buffers are forwarded from *every* pad unconditionally, but an EOS event is only forwarded when it arrives on the pad that is currently `active_sinkpad` (i.e., the pad that chained the most recent buffer, set unconditionally in `gst_join_pad_chain`). Since each `videotestsrc` branch is configured with `num-buffers=10` and only emits EOS after its 10th buffer, the branch whose EOS actually gets forwarded must, by construction, have already chained all 10 of its own buffers before that EOS can be the "active" one. So N ≥ 10 is a hard structural floor, not a statistical/empirical one — the N=10 row is in fact the worst case the pipeline can produce, and it already clears the old 0.05s bound with ~78% headroom (0.089s vs 0.05s). This closes the residual concern more tightly than "cannot identify a plausible mechanism" — it's provably unreachable.
- Conclusion: the no-code-change response to ② is both numerically correct and, on inspection of `gstjoin.c`, backed by a proven (not just empirical) lower bound on N. No action needed.

**Full-PR pass (`64aa8c4e` + `0a146ae4`):**
- Root cause: confirmed again directly against `gstjoin.c` — `gst_join_pad_event()`'s `forward = (pad == active_sinkpad)` gating on events (but not on the unconditional buffer-forwarding path in `gst_join_pad_chain()`) is exactly the mechanism described in the commit message. Valid root-cause fix, correctly scoped to the test rather than papering over it or touching `join` itself.
- Memory/refs: both `tensor_sink` acquisitions (`gst_bin_get_by_name`) are paired with `g_clear_pointer (..., gst_object_unref)` before their pipeline is torn down in both runs; no leak introduced by either commit.
- Test strength: not weakened — the two hard-coded absolute wall-clock bounds (which were simply wrong given `join`'s documented single-active-stream contract) are replaced by frame-count-derived relative bounds that still assert the same underlying property (`sync=true` paces to the stored framerate; `sync=false` does not), plus a new equality check that adds coverage the old test didn't have.
- Scope: only `tests/nnstreamer_datarepo/unittest_datareposrc.cc`, one test function, consistent across both commits.

Both previously-raised items are resolved (① by code change, ② by a verified — and in fact provably tighter than assumed — analytical argument). No new issues found. This PR is mergeable as-is.

myungjoo:
Closing this in favour of #4874, which covers everything here and the other half of the same CI flakiness. Written up so the reasoning is on the record (drafted with AI assistance).

**The analysis in this PR is correct and carried over.** #4874 reaches the same root cause for `fps30ReadFlexibleTensors` — `join` forwards events only from the pad that delivered the last buffer (`gst/join/gstjoin.c`, `forward = (pad == active_sinkpad)`), so the first EOS that arrives while its own pad is active cuts off the remaining sources and the stored frame count depends on scheduling. Probing the writer pipeline directly confirms it: 30 frames on every idle run, and 17/21/22/24/25/26/28/30 with the cores saturated. A hard-coded `EXPECT_LT (0.8, elapsed_time)` cannot survive that, which is exactly the 0.758597 s in the report.

**What #4874 adds.** The same CI job also fails on `datareposink.writeFlexibleTensors_n` (`unittest_datareposink.cc:578`, `file_info` NULL), through a different mechanism this PR does not touch: `meson test` runs test binaries in parallel with the build directory as their working directory, and both datarepo binaries create and `g_remove()` the same fixed names (`img_%02d.png`, `img.json`). Each deletes the images the other is feeding to `multifilesrc`, the sink pipeline errors out before caps reach `datareposink`, and since the sink opens its output file on `PAUSED_TO_PLAYING` the file is never created. Locally, `--repeat 12` over both binaries fails 12 of 24 runs, and passes 12/12 with `--num-processes 1`. #4874 gives each binary a private working directory, so both failure modes are closed together.

**Nothing from this PR is lost.** The one check here that #4874 lacked — counting frames on the unsynced run and asserting both runs read the same number — has been added there (`EXPECT_EQ (no_sync_count, buffer_count)`). The remaining differences are margins, and #4874's are the tighter ones: it bounds the synced run at `(n-1)/fps * 0.9` rather than `n/fps * 0.7` (the last frame renders at `(n-1)/fps`, so that is the exact expectation, and the inequality `(n-1)/fps*0.9 >= n/fps*0.7` holds for every `n >= 10`), and it requires at least 10 frames so the unsynced bound cannot collapse toward the fixed pipeline overhead.

The `join` EOS behaviour itself is left alone in both PRs — it is a real data-loss path for any pipeline merging concurrent streams, and changing when it forwards EOS deserves its own PR rather than riding along with a test fix.

PR #4873, opened 2026-08-26
[Test/TVM] Add SSAT test cases for tensor_filter::tvm

myungjoo:
`tests/nnstreamer_filter_tvm/` carried a gtest but no SSAT suite, so the tvm sub-plugin was never exercised through `gst-launch` against the built or installed plugin the way every other filter shipping an in-tree model is. TVM was the only such filter still missing one.

- Add `tests/nnstreamer_filter_tvm/runTest.sh`, reusing the in-tree `tvm_add_one_*.so_` models. No new binaries and no build-system changes.
- Positive case 1: the models are a single `ones_like + add` node over `3:480:640:1` float32, so passing the tee'd input through the filter and subtracting 1.0 must reproduce the input byte for byte. All 921600 elements are compared, where the gtest inspects only the first ten, and the output size is pinned to `3 * 480 * 640 * sizeof(float32)`.
- Positive case 2: `num_input_tensors` omitted, so the sub-plugin has to infer it from the model.
- Negative cases: uint8 input against a float32 model, dimension mismatch, a file that exists but is not a tvm module, unknown `device`, and `num_input_tensors:0`.
- The suite skips cleanly when the sub-plugin is not built or the architecture has no prebuilt model, so it is a no-op on builds without tvm and on architectures such as riscv64. It runs in the Tizen GBS and Debian pdebuild `ssat` phases.

**Relation to #3886:**

This supersedes #3886 by @ksghaier, which proposed the same kind of test. That PR added 16.7MB of prebuilt mobilenet TVM modules whose provenance could not be verified from the tree, and it stalled in 2022 because those modules needed a newer tvm runtime than the packaged one. Reusing the `tvm_add_one_*.so_` models that the existing gtest already loads today avoids both problems: nothing new enters git history, and there is no runtime-version risk, since a runtime that cannot load them would already be failing `unittest_filter_tvm`. Please close #3886 when this merges — GitHub does not auto-close a referenced pull request.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [*]Skipped (no source change)
2. Run test: [ ]Passed [ ]Failed [*]Skipped (needs a tvm-enabled build; verified in CI)

**How to evaluate:**
1. Build with `-Dtvm-support=enabled` into `build/`.
2. `cd tests/nnstreamer_filter_tvm && ssat` — expect 2 positive cases, 1 size assertion and 5 negative cases to pass.
3. On a build without tvm, confirm the group reports no cases instead of failing.

### DIFF
--- a/tests/nnstreamer_filter_tvm/runTest.sh
+++ b/tests/nnstreamer_filter_tvm/runTest.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+##
+## SPDX-License-Identifier: LGPL-2.1-only
+##
+## @file runTest.sh
+## @author MyungJoo Ham <myungjoo.ham@samsung.com>
+## @date Aug 26 2026
+## @brief SSAT Test Cases for NNStreamer tensor_filter::tvm
+##
+
+if [[ "$SSATAPILOADED" != "1" ]]; then
+    SILENT=0
+    INDEPENDENT=1
+    search="ssat-api.sh"
+    source $search
+    printf "${Blue}Independent Mode${NC}"
+fi
+
+# This is compatible with SSAT (https://github.com/myungjoo/SSAT)
+testInit $1
+
+# NNStreamer and plugins path for test
+PATH_TO_PLUGIN="../../build"
+
+if [[ -d ${PATH_TO_PLUGIN} ]]; then
+    filter_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_filter"
+else
+    ini_file="/etc/nnstreamer.ini"
+    if [[ ! -f ${ini_file} ]]; then
+        echo "Cannot identify nnstreamer.ini"
+        report
+        exit
+    fi
+
+    filters_line=$(grep "^filters" ${ini_file})
+    if [[ ${filters_line%%=*} != "filters" ]]; then
+        echo "Cannot identify the sub-plugin path from ${ini_file}"
+        report
+        exit
+    fi
+    filter_path=${filters_line##*=}
+fi
+
+if [[ ! -f ${filter_path}/libnnstreamer_filter_tvm.so ]]; then
+    echo "Cannot find the tvm sub-plugin"
+    report
+    exit
+fi
+
+# The test model is prebuilt per architecture, matching ARCH of unittest_filter_tvm.cc.
+# CI exercises the x86_64 model only: the gbs armv7l and aarch64 jobs build with
+# "unit_test 0" and the pdebuild matrix is amd64, so the other two are not covered there.
+case $(uname -m) in
+    x86_64) MODEL_ARCH="x86_64" ;;
+    aarch64) MODEL_ARCH="aarch64" ;;
+    arm*) MODEL_ARCH="arm" ;;
+    *) MODEL_ARCH="unsupported" ;;
+esac
+
+PATH_TO_MODEL="../test_models/models/tvm_add_one_${MODEL_ARCH}.so_"
+PATH_TO_INVALID_MODEL="../test_models/labels/labels.txt"
+
+if [[ ! -f ${PATH_TO_MODEL} ]]; then
+    echo "No tvm test model for $(uname -m)"
+    report
+    exit
+fi
+
+# tvm_add_one_*.so_ is a single 'ones_like + add' node: 3:480:640:1 float32 in and out.
+# Every pixel value is an integer in [-255, 0] after the offset, so +1.0 and -1.0 cancel
+# bit-exactly and the tvm branch must equal the direct branch byte for byte.
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tee name=t ! queue ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" custom=device:CPU,num_input_tensors:1 ! tensor_transform mode=arithmetic option=add:-1.0 ! filesink location=\"testcase1.tvm.log\" sync=true t. ! queue ! filesink location=\"testcase1.direct.log\" sync=true" 1 0 0 $PERFORMANCE
+callCompareTest testcase1.direct.log testcase1.tvm.log 1 "Compare 1" 0 0
+
+# Pin the output tensor shape and type: 3 * 480 * 640 * sizeof(float32)
+[ -s testcase1.tvm.log ] && [ "$(wc -c < testcase1.tvm.log)" -eq 3686400 ]
+testResult $? 1 "Output tensor size" 0 1
+
+# num_input_tensors is optional; the sub-plugin infers it from the model
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tee name=t ! queue ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" ! tensor_transform mode=arithmetic option=add:-1.0 ! filesink location=\"testcase2.tvm.log\" sync=true t. ! queue ! filesink location=\"testcase2.direct.log\" sync=true" 2 0 0 $PERFORMANCE
+callCompareTest testcase2.direct.log testcase2.tvm.log 2 "Compare 2" 0 0
+
+# Negative test: uint8 input while the model expects float32
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" custom=device:CPU,num_input_tensors:1 ! fakesink" 1_n 0 1 $PERFORMANCE
+
+# Negative test: input dimension does not match the model
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=320,height=240,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" custom=device:CPU,num_input_tensors:1 ! fakesink" 2_n 0 1 $PERFORMANCE
+
+# Negative test: the given file exists but is not a tvm module
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tensor_filter framework=tvm model=\"${PATH_TO_INVALID_MODEL}\" custom=device:CPU,num_input_tensors:1 ! fakesink" 3_n 0 1 $PERFORMANCE
+
+# Negative test: unknown device in the custom property
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" custom=device:NPU,num_input_tensors:1 ! fakesink" 4_n 0 1 $PERFORMANCE
+
+# Negative test: num_input_tensors must be greater than 0
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=480,height=640,framerate=0/1 ! videoconvert ! video/x-raw,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:-255.0 ! tensor_filter framework=tvm model=\"${PATH_TO_MODEL}\" custom=device:CPU,num_input_tensors:0 ! fakesink" 5_n 0 1 $PERFORMANCE
+
+# Cleanup
+rm -f testcase1.tvm.log testcase1.direct.log testcase2.tvm.log testcase2.direct.log
+
+report


myungjoo:

Scope reviewed: the single added file, `tests/nnstreamer_filter_tvm/runTest.sh` (99 lines, SSAT suite for `tensor_filter::tvm`). I cross-checked it against `ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc`, `tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc`, the in-tree `tvm_add_one_{x86_64,aarch64,arm}.so_` fixtures, other `runTest.sh` files in the tree, `tests/meson.build`, `packaging/nnstreamer.spec`, `debian/rules`, `debian/control`, and the per-PR workflows (`.github/workflows/gbs_build.yml`, `pdebuild.yml`) plus `.github/actions/check-rebuild`.

Summary: this is a well-scoped, test-only change that does exactly what it says, and I did not find any blocking issues.

Checklist results:

1. **Solves the stated topic** — Yes. It adds an SSAT suite for the tvm sub-plugin (2 positive cases + a byte-exact tee-compare + a pinned output-size assertion + 5 negative cases), closing the gap called out in the PR description, while deliberately reusing the in-tree `tvm_add_one_*.so_` fixtures instead of repeating #3886's 16.7MB unverifiable-binary / runtime-version-coupling mistakes. The math checks out: the model is `ones_like + add`, so `+1.0` then `-1.0` on integer-valued float32 pixels cancels bit-exactly, matching the direct branch.

2. **No regression to other modules** — Confirmed no risk. The change touches only this one new file; no meson/build/packaging file is modified. The tee-based "filtered branch vs. direct branch" comparison pattern is copy-consistent with `tests/nnstreamer_filter_custom/runTest.sh` (testcase01) and similar patterns in `nnstreamer_filter_pytorch`/`nnstreamer_filter_executorch`, so it isn't introducing a novel, unproven test idiom.

3. **Code size / blast radius proportionate to the topic** — Yes. Single file, test-only, no other module touched.

4. **Test completeness / correctness** — Verified against the actual sub-plugin source:
   - Negative case 3 (file exists but isn't a tvm module, using `test_models/labels/labels.txt`) relies on `tvm::runtime::Module::LoadFromFile` throwing when parsing a non-module file. That exception is not caught inside `tensor_filter_tvm.cc`, but it *is* caught generically in the common C++ subplugin wrapper (`tensor_filter_support_cc.cc`'s `configure_instance` call, catching `std::exception`/`...` and returning `-EINVAL`), so this fails gracefully as a pipeline error rather than crashing/hanging the test runner.
   - Negative case 4 (`device:NPU`) and case 5 (`num_input_tensors:0`) map directly to the `invalid_option` checks in `tvm_subplugin::parse_custom_prop`.
   - Negative cases 1/2 (type and dimension mismatch) rely on the generic tensor_filter caps-negotiation failure path, same convention used by existing pytorch/executorch negative tests.
   - The pinned-size assertion (`3*480*640*4 = 3686400` bytes) and the `callCompareTest ... 0 0` (exact byte match) mode are both correct for this model's known behavior.

5. **Detects future regressions from other modules** — Reasonably good: the byte-exact compare plus the fixed output-size assertion will catch both incorrect-output and unexpected-shape/type regressions in this framework path, not just "did it crash."

6. **CI will actually fail on a real problem** — Verified end-to-end, not just assumed:
   - `packaging/nnstreamer.spec` already runs `ssat -n -p=1 --summary summary.txt -cn _n` from `tests/` (Tizen GBS `unit_test` phase) and `debian/rules` runs the equivalent for the Debian package — SSAT auto-discovers `runTest.sh` per subdirectory, so no additional wiring was needed for this new directory to be picked up.
   - `tvm_support` is `1` in the spec's main profile and `tvm-runtime-dev`/`nnstreamer-tvm` are present in `debian/control`, so the sub-plugin is actually built where these phases run.
   - Both `.github/workflows/gbs_build.yml` and `.github/workflows/pdebuild.yml` trigger `on: pull_request` (not just nightly/manual), and `.github/actions/check-rebuild/check_if_rebuild_requires.sh` maps any file outside `packaging/*`, `debian/*`, `jni/*` (including this new `tests/...` file) to `REBUILD=YES` for both `gbs` and `debian` modes — so this exact PR's own diff triggers both workflows to actually build with tvm enabled and run the new suite, rather than being skipped as a "docs/test-only, no rebuild" change.
   - One caveat worth flagging (low severity, pre-existing, not introduced by this PR): `gbs_build.yml`'s matrix only sets `unit_test 1` for the `x86_64` job; `armv7l` and `aarch64` run with `unit_test 0`, so today only the x86_64 model (`tvm_add_one_x86_64.so_`) is actually exercised in per-PR CI. The existing `unittest_filter_tvm.cc` carries a note ("Currently the test model can be executed only on x86_64"), suggesting the aarch64/arm `.so_` fixtures may not be fully verified to load correctly. This new SSAT script inherits exactly the same exposure as the existing gtest — it isn't a new risk — but if `unit_test` is ever flipped on for the arm/aarch64 GBS jobs, this suite (like the existing gtest) could turn red for reasons unrelated to a real regression in `tensor_filter_tvm`. Suggest carrying the same arch caveat as a comment in `runTest.sh`, or filing a small follow-up to confirm/fix the aarch64/arm fixtures, but this should not block this PR.

7. **Architecture/API changes requiring doc updates** — N/A. No architecture or API change; test-only addition.

Overall: no blocking issues found. The one item above is informational/low severity and pre-existing (inherited from the already-merged gtest and today's CI matrix), not a defect in this PR's diff.

myungjoo:
Thanks — the one item raised (arch coverage in per-PR CI) is now addressed in the script.

Added the caveat as a comment next to the architecture mapping, stating only what is verifiable: per-PR CI exercises the x86_64 model alone, because the `armv7l` and `aarch64` GBS jobs build with `unit_test 0` and therefore never reach the `ssat` phase.

I am deliberately not filing a follow-up to "confirm/fix the aarch64/arm fixtures", for two reasons:

- The exposure is pre-existing and unchanged by this PR. `unittest_filter_tvm.cc` already selects the same three fixtures via its compile-time `ARCH` macro, and `packaging/nnstreamer.spec` already gates the tvm gtest on `%ifarch %arm x86_64 aarch64`. This PR adds no new dependency on the arm/aarch64 fixtures beyond what the merged gtest has had since 2021.
- If an arm or aarch64 fixture genuinely failed to load once `unit_test` were enabled on those jobs, a red test would be the correct signal rather than a false positive. Suppressing or skipping it in advance would hide a real defect, so there is nothing to pre-emptively fix here.

The `@notes Currently the test model can be executed only on x86_64` line in `unittest_filter_tvm.cc` does look stale — it predates commit 076a78ea, which added the aarch64 and arm fixtures — but correcting it would need evidence from an actual arm run that I do not have, so I left that file untouched rather than replace one unverified claim with another.

No other changes; the diff is still a single new file.

myungjoo:

This is a second-round review. I re-read the full diff independently (not just the delta) and separately checked whether the round-1 item and the author's response hold up.

**1. Was the round-1 item resolved, and is the author's reasoning sound?**

Yes on both counts.

- The diff since round 1 is exactly the two comment lines the author described — file grew from 99 to 101 lines, and `git diff` shows only:
  ```
  # The test model is prebuilt per architecture, matching ARCH of unittest_filter_tvm.cc.
  # Only the x86_64 model is exercised by per-PR CI: the armv7l and aarch64 gbs jobs build
  # with "unit_test 0", so ssat does not run there.
  ```
  placed directly above the `case $(uname -m) in` block. No other lines changed.
- I verified the claim itself against the actual CI config rather than taking it on faith: `.github/workflows/gbs_build.yml` sets `unit_test 1` only for the `x86_64` matrix entry and `unit_test 0` for `armv7l`/`aarch64`; `packaging/nnstreamer.spec` gates the entire `ssat -n -p=1 ... -cn _n` invocation (line 1092) behind `%if 0%{?unit_test}`. So the comment's factual claim is correct and independently checkable from files already in the tree — it isn't just an assertion.
- The author's reason for *not* filing a follow-up on the arm/aarch64 fixtures also checks out: `unittest_filter_tvm.cc`'s `ARCH` macro has covered `aarch64`/`arm`/`x86_64` since commit `076a78ea` (2021), and `packaging/nnstreamer.spec` has gated the tvm gtest on `%ifarch %arm x86_64 aarch64` since the same era — so this PR adds no new dependency on the unverified arm/aarch64 fixtures beyond what the merged gtest has carried for years. Declining to "fix" the stale `@notes` comment in `unittest_filter_tvm.cc` without an actual arm run to verify against is a reasonable call; a guess wouldn't be better than the current stale note.
- Net effect: this was a low-severity, non-blocking documentation item in round 1, and it's now closed out correctly with no scope creep (still a single file, no build/meson/packaging changes).

One optional nit, not worth blocking on: the new comment scopes the caveat to "per-PR CI" / the GBS jobs. Worth noting for completeness (not a request to change anything) — the Debian path has the same limitation: `debian/rules` only runs `ssat -n -p=1` under `ifeq ($(DEB_HOST_ARCH), amd64)`, and `.github/workflows/pdebuild.yml`'s matrix is `arch: [ amd64 ]` only. So in practice no CI path in the repo (GBS or Debian) exercises the aarch64/arm fixtures today, not just the per-PR GBS jobs. This doesn't make the comment wrong, just slightly narrower in scope than the full picture — informational only.

**2. Independent full-diff re-check (treating it as a first look):**

- Skip-detection logic is correct: the shared library target in `ext/nnstreamer/tensor_filter/meson.build` is literally named `nnstreamer_filter_tvm`, producing `libnnstreamer_filter_tvm.so`, matching the existence check in the script.
- The arithmetic claim is correct: input pixels are `uint8` in `[0,255]`, `typecast:float32,add:-255.0` maps them to `[-255,0]`; the model computes `ones_like + input` (confirmed by the existing gtest's `EXPECT_EQ(1, output[i])` for zero input); `add:-1.0` afterward cancels that offset exactly since all values involved are integers exactly representable in float32. The byte-exact `callCompareTest ... 0 0` mode is the correct choice here and mirrors the existing convention in `tests/nnstreamer_filter_custom/runTest.sh`.
- The pinned size assertion (`3*480*640*sizeof(float32) = 3686400`) is arithmetically correct and matches the model's `3:480:640:1` shape confirmed in `unittest_filter_tvm.cc`.
- Negative cases map onto real validation code in `ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc`: `device:NPU` and `num_input_tensors:0` both hit `parse_custom_prop`'s `invalid_option` path; the "file exists but isn't a tvm module" case (`test_models/labels/labels.txt`, confirmed present in-tree) throws out of `Module::LoadFromFile`, which is not caught locally in `tensor_filter_tvm.cc` but is caught generically (`catch (const std::exception&)` / `catch (...)`) in the shared C++ wrapper's `configure_instance` call in `gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc`, returning `-EINVAL` rather than crashing.
- `num_input_tensors` omission (case 2) is correctly handled: `configure_instance` only infers `get_num_inputs()` `if (inputInfo.num_tensors == 0)`, matching the "optional, inferred from model" claim.
- CI will actually run this: `.github/actions/check-rebuild/check_if_rebuild_requires.sh` puts any path outside `packaging/*`/`debian/*`/`jni/*` (this new test file included) into the catch-all case that sets `regbs=1` and `redebian=1`, so both the GBS and pdebuild workflows rebuild and run `ssat` for this PR's own diff.
- Structural conventions (the `SSATAPILOADED` sourcing guard, the `PATH_TO_PLUGIN`/`ini_file` fallback lookup, `testInit`/`report`/`exit` skip pattern) match `tests/nnstreamer_filter_pytorch/runTest.sh` and other existing suites closely — this isn't introducing a new, unproven test idiom.
- No build system, packaging, or other module files are touched; this remains a single-file, test-only addition proportionate to the stated goal.

**Conclusion:** No blocking issues, in this round or carried over from round 1. The round-1 item is resolved with an accurate, verifiable comment, and the author's decision not to additionally "fix" the stale docstring in `unittest_filter_tvm.cc` is justified given the lack of verification data. My independent re-check of the full diff found nothing new to flag. Mergeable as-is.

myungjoo:
Took the non-blocking observation as well, since it makes the comment strictly more accurate rather than merely not-wrong.

Verified it first: `debian/rules:70` gates the `ssat` run on `DEB_HOST_ARCH == amd64`, and `pdebuild.yml`'s matrix is `arch: [ amd64 ]`. So the Debian path is amd64-only too, and no CI path in the tree exercises the aarch64 or arm fixtures. The comment now says so:

```sh
# Only the x86_64 model is exercised in CI: the gbs armv7l and aarch64 jobs build with
# "unit_test 0", and debian/rules runs ssat on amd64 only.
```

Amended into the same commit. The diff remains a single new file.

myungjoo:
[Automated review (3rd pass) by a separate Claude agent, relayed to this PR]

Third-round check, scoped per the request to: (a) whether round 2's non-blocking observation was resolved, (b) whether the new comment text is factually accurate against `debian/rules`, `.github/workflows/pdebuild.yml`, `.github/workflows/gbs_build.yml`, and `packaging/nnstreamer.spec`, and (c) whether the amend introduced anything new. I re-fetched the PR head and checked these files directly rather than trusting the prior claim at face value.

**(a) Round 2 observation — addressed in spirit.** The commit now mentions the Debian path alongside GBS, matching the amend described in the author's reply. Confirmed via `git show`/`gh pr diff`: still a single new file, single commit (amended), and the only change since round 2 is the two comment lines above `case $(uname -m) in`.

**(b) The new wording is half right.** Current text:
```sh
# Only the x86_64 model is exercised in CI: the gbs armv7l and aarch64 jobs build with
# "unit_test 0", and debian/rules runs ssat on amd64 only.
```
- GBS half: correct. `gbs_build.yml`'s matrix sets `unit_test 1` only for `x86_64` and `unit_test 0` for `armv7l`/`aarch64`, and `packaging/nnstreamer.spec:1060` gates the `ssat -n -p=1 ...` call behind `%if 0%{?unit_test}`. Verified directly.
- Debian half: **the specific claim is inaccurate**. `debian/rules` reads (line numbers from `git show origin/main:debian/rules`):
  ```
  69:ifeq ($(shell dpkg-vendor --derives-from Ubuntu && echo yes), yes)
  70:ifeq ($(DEB_HOST_ARCH), amd64)
  71:    if [ -f './tests/tizen_nnfw_runtime/unittest_nnfw_runtime_raw' ]; then ...; fi
  72:endif
  73:    cd tests && ssat -n -p=1 && cd ..
  74:else
  75:    echo "Skipping SSAT test because it's not in Debian."
  76:endif
  ```
  The `endif` on line 72 closes the `DEB_HOST_ARCH == amd64` block *before* the `ssat` call on line 73. So `ssat -n -p=1` is **not** arch-gated in `debian/rules` — only the `tizen_nnfw_runtime_raw` binary test above it is. `debian/rules` would run `ssat` on any Debian/Ubuntu-derived architecture it's invoked for; the reason today's Debian CI happens to exercise amd64 alone is that `.github/workflows/pdebuild.yml`'s build matrix is `arch: [ amd64 ]` — a property of the workflow, not of `debian/rules`.

  This exact claim ("`debian/rules` only runs `ssat` under `ifeq ($(DEB_HOST_ARCH), amd64)`") was first asserted, apparently unverified, in the round-2 review comment, then repeated in the author's reply and folded into the commit as-is. It traces to not noticing that the inner `ifeq`'s `endif` precedes the `ssat` line rather than wrapping it.

  Practical impact today: none. The end conclusion — "only x86_64 is exercised by CI, in both the GBS and Debian paths" — is still correct, purely because `pdebuild.yml` only builds amd64. But the comment now asserts a gating mechanism that doesn't exist in `debian/rules`, which could mislead a future reader: if `pdebuild.yml`'s matrix is ever widened to another architecture, `debian/rules` would immediately start running `ssat` there too with no change to `debian/rules` needed — the opposite of what this comment implies.

**(c) No other change since round 2.** Nothing else in the diff moved; still a single new file (101 lines), no build/meson/packaging file touched.

**Severity:** Low / informational, non-blocking. It's a comment-only inaccuracy in a new test script — it doesn't affect what the test does, how CI gates it, or its ability to catch real regressions in `tensor_filter_tvm`. Optional reword for a future touch-up, not required before merge, e.g.: "...and the Debian path only builds amd64 in CI today (`pdebuild.yml`'s matrix), so `ssat` there only ever runs on x86_64 in practice — `debian/rules` itself does not gate `ssat` by architecture."

**Overall verdict: mergeable as-is.** Rounds 1 and 2 already verified the test logic (byte-exact tee-compare, pinned output size, all 5 negative cases mapping to real validation paths in `tensor_filter_tvm.cc`), CI wiring (`check-rebuild` triggers both workflows on this diff, both actually build tvm and reach the `ssat` phase for x86_64), and scope (single file, no build-system changes). This round found no functional defects — only the comment-wording nit above, which does not change any test behavior or CI outcome and does not need to block merge.

myungjoo:
Good catch, and the correction is right — that was my error. I read `debian/rules` from grep output showing `ifeq ($(DEB_HOST_ARCH), amd64)` at line 70 and `ssat -n -p=1` at line 73, and assumed the former enclosed the latter without checking the lines in between.

Reading them: the `endif` at line 72 closes the amd64 conditional before the `ssat` call, so the amd64 gate covers only the `unittest_nnfw_runtime_raw` invocation on line 71. The `ssat` run on line 73 sits inside the outer `ifeq (dpkg-vendor --derives-from Ubuntu)` block, so it is gated on Ubuntu-vs-Debian, not on architecture. Were the pdebuild matrix ever extended to arm64, `debian/rules` would run `ssat` there with no change — the opposite of what my comment implied.

Corrected to attribute the amd64-only coverage to the actual cause:

```sh
# CI exercises the x86_64 model only: the gbs armv7l and aarch64 jobs build with
# "unit_test 0" and the pdebuild matrix is amd64, so the other two are not covered there.
```

Both halves are now verified against the files rather than inferred: `gbs_build.yml` sets `unit_test 1` for x86_64 only while `packaging/nnstreamer.spec` gates the `ssat` phase on `%if 0%{?unit_test}`, and `pdebuild.yml`'s matrix is `arch: [ amd64 ]`.

Amended into the same commit; still a single new file.

myungjoo:
[Automated review (4th pass) by a separate Claude agent, relayed to this PR]

Fourth-round check, scoped per the request to: (1) whether the amended comment wording is now factually accurate against `debian/rules`, `.github/workflows/pdebuild.yml`, `.github/workflows/gbs_build.yml`, and `packaging/nnstreamer.spec`, (2) whether the amend introduced anything new, (3) whether the comment carries a staleness risk. I read each file directly (not via grep line numbers alone, per the round-3 lesson) rather than trusting the prior claim.

**Current text:**
```sh
# CI exercises the x86_64 model only: the gbs armv7l and aarch64 jobs build with
# "unit_test 0" and the pdebuild matrix is amd64, so the other two are not covered there.
```

**(1) Phrase-by-phrase verification:**

- *"the gbs armv7l and aarch64 jobs build with unit_test 0"* — confirmed exactly. `.github/workflows/gbs_build.yml`'s matrix has `gbs_build_arch: x86_64` paired with `--define "unit_test 1"`, and both `armv7l` and `aarch64` paired with `--define "unit_test 0"`, verbatim. `packaging/nnstreamer.spec:1059-1092` gates the entire `ssat -n -p=1 ... -cn _n` invocation behind `%if 0%{?unit_test}`, so `unit_test 0` does correctly mean ssat never runs for those two jobs — this is the mechanism, not just a correlation.
- *"the pdebuild matrix is amd64"* — confirmed exactly. `.github/workflows/pdebuild.yml`'s `strategy.matrix.arch` is `[ amd64 ]`, the only value.
- Critically, the new wording **no longer names `debian/rules` as the gating mechanism** for the Debian path — it attributes amd64-only coverage to the pdebuild workflow's matrix, a true statement about the actual cause. I re-read `debian/rules` lines 65-76 in full sequence to confirm the round-3 finding still stands: the `endif` at line 72 closes the `ifeq ($(DEB_HOST_ARCH), amd64)` opened at line 70 *before* the `ssat -n -p=1` call at line 73, so that call sits only inside the outer `ifeq (dpkg-vendor --derives-from Ubuntu)` block (lines 69/74/76) — gated by Ubuntu-vs-Debian, not architecture. The current comment makes no claim inconsistent with this; it sidesteps the mechanism entirely and states only the observable, correct fact (pdebuild's matrix is amd64-only). Good fix.

**(2) Nothing new introduced.** `gh pr view --json commits` shows a single commit; `changed_files` on the PR is still 1 (`tests/nnstreamer_filter_tvm/runTest.sh`). The amend replaced only the two comment lines above `case $(uname -m) in`; no other line moved.

**(3) Staleness risk — acknowledged but acceptable.** The comment is a snapshot of today's CI config (`unit_test 0/1` per gbs arch, pdebuild's `arch: [ amd64 ]`) rather than a structural invariant, so if either workflow's matrix changes later without touching this comment, it could go stale. This is an inherent property of any comment describing external CI wiring from inside a test script, not a defect introduced here — and the wording is appropriately hedged ("CI exercises... today's config", implicitly) rather than asserting a load-bearing guarantee. Not blocking.

**Overall verdict: no remaining issues. Mergeable as-is.** All three prior rounds already closed out test-logic completeness, CI wiring, and scope (single file, no build-system changes); this round confirms the comment text is now fully accurate, the diff remains minimal, and no new defect was introduced by the fix.

myungjoo:
All 18 checks pass. Recording what the two jobs that actually execute this suite reported, since a guarded SSAT script can also pass by skipping and that distinction matters here.

**Tizen GBS x86_64 (`unit_test 1`)** and **Ubuntu pdebuild amd64** both built `libnnstreamer_filter_tvm.so` and ran the group in full:

```
[Starting] nnstreamer_filter_tvm
[PASSED] 1:Compare 1
[PASSED] 1:Output tensor size
[PASSED] 2:gst-launch of case 2
[PASSED] 2:Compare 2
[PASSED] 1_n:gst-launch of case 1_n
[PASSED] 2_n:gst-launch of case 2_n
[PASSED] 3_n:gst-launch of case 3_n
[PASSED] 4_n:gst-launch of case 4_n
[PASSED] 5_n:gst-launch of case 5_n
[PASSED] nnstreamer_filter_tvm (10 passed among 10 cases)
```

`10 passed among 10 cases` is the point — in the same runs, `nnstreamer_filter_pytorch` and `nnstreamer_filter_executorch` report `0 passed among 0 cases` because those frameworks are absent, which is what a skip looks like. This group is genuinely executing.

Two results are worth calling out:

- **`Compare 1` passing** confirms empirically what the PR description asserts from reading the fixtures: the model is a single `ones_like + add` node, so `+1.0` followed by `-1.0` cancels bit-exactly across all 921600 float32 elements against the tee'd direct branch. That is the coverage the gtest does not provide, since it checks only the first ten outputs.
- **All five negative cases failing the pipeline** addresses the concern raised on #3886 that a negative case can pass for an unintended reason. `1_n` in particular keeps the positive case's `480x640` caps and varies only the tensor type, and `gst_tensor_info_is_equal` compares `type` before `dimension`, so it fails on the type mismatch it is written for.

The group costs about one second of wall clock in both jobs.

myungjoo:
Checked all three. One is fixed here; the other two I am deliberately not doing in this PR, with reasons below.

**제안 3 (nit) — `wc -c` error noise: fixed.**

```sh
[ -s testcase1.tvm.log ] && [ "$(wc -c < testcase1.tvm.log)" -eq 3686400 ]
```

Now short-circuits when case 1 produced no file, so the case still fails but without bash's `integer expression expected` on stderr. Amended into the same commit.

On the other two nits, no change:
- The overlap of `2_n`/`4_n`/`5_n` with `launch01_n`/`launchInvalidDevice_n`/`launchInvalidInputNum_n` is intentional and is the same overlap the pytorch and lua suites have with their own gtests — the SSAT copies verify the `gst-launch` path against the built/installed plugin rather than the filter API.
- `... ! videoconvert ! video/x-raw,format=RGB ! tensor_converter` is copied verbatim from `nnstreamer_filter_lua/runTest.sh:47`. Both orderings exist in the tree (the tvm gtest puts `videoconvert` first), and I would rather match the sibling SSAT suite than introduce a third variant for a no-op element.

**제안 1 (install_subdir) — not doing this here.**

The observation is right: `tests/meson.build` enumerates SSAT directories explicitly and `nnstreamer_filter_tvm` is not among them, so the `/etc/nnstreamer.ini` branch is unreachable today. I also confirmed adding it would be packaging-safe, since `%files unittests` globs the directory (`%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests`) rather than listing files.

But eleven SSAT suites are missing from that list, not one:

```
codegen, gstreamer_join, nnstreamer_decoder_tensor_region, nnstreamer_filter_lua,
nnstreamer_filter_onnxruntime, nnstreamer_filter_tensorrt, nnstreamer_filter_tensorrt10,
nnstreamer_filter_tvm, nnstreamer_grpc, nnstreamer_octet_stream, nnstreamer_sparse
```

Adding `install_subdir` for tvm alone would make it the only recently-added suite that is installed, which is a coincidental fix rather than a real one. The `ini` branch is kept because it is byte-for-byte the sibling idiom and goes live the moment that list is repaired repo-wide. That repair is a separate change across eleven directories and does not belong in a PR that adds one test file.

**제안 2 (model_path leak) — real, and slightly broader than described; separate change.**

Confirmed: `configure_instance()` does `model_path = g_strdup(...)` before `Module::LoadFromFile()`, while `empty_model` is still `true`, so the `if (empty_model) return;` guard at the top of `cleanup()` skips the `g_free`. Also correct that `3_n` is the first test to reach it — the existing negative cases use a nonexistent `temp.so` and are stopped earlier by the `g_file_test` check.

It is a bit wider than the report says, though: the same guard means `model_path` also leaks on the two later throws, when `get_input` or `get_output` is not defined in the module. So the fix is not just reordering the `g_strdup` — it wants the `g_free (model_path)` moved above the `empty_model` early-return in `cleanup()`, which covers every throw path.

That is a change to `ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc`, i.e. sub-plugin behavior, in a PR whose commit is titled `[Test/TVM]` and touches no source. I will send it as its own `[Filter/TVM]` change rather than mixing two subjects into one commit; it is a pre-existing defect and does not gate this PR.

PR #4874, opened 2026-08-26
[Test/datarepo] Fix flaky datarepo unit tests under parallel CI runs

myungjoo:
## The problem

Two datarepo unit tests fail intermittently on shared CI runners and take unrelated PRs down with them:

- `datareposink.writeFlexibleTensors_n` — `Expected: (file_info) != (nullptr)` at `unittest_datareposink.cc:578`
- `datareposrc.fps30ReadFlexibleTensors` — `Expected: (0.8) < (elapsed_time), actual: 0.8 vs 0.758597` at `unittest_datareposrc.cc:668`

Both were seen on 2026-08-26 within minutes of each other on two different PRs, and the `ubuntu-22.04-arm` variant of the same job passed.

## Reproduction

```
meson test -C build unittest_datareposink unittest_datareposrc --repeat 12
```

**12 of 24 runs failed**, with exactly the two assertions above. The same command with `--num-processes 1`: **12/12 passed**. The trigger is concurrency, and there are two independent causes.

### Cause 1 — the two binaries share a working directory

`meson test` runs test binaries in parallel with the build directory as their working directory. Both datarepo binaries have their own `create_image_test_file()` writing the *same* names (`img_%02d.png`, `img.json`), and both `g_remove()` them when done. Each deletes the images the other is feeding to `multifilesrc`.

That explains the NULL `file_info` precisely: with the images gone, `multifilesrc` errors out before caps ever reach `datareposink`, and the sink opens its output file in `PAUSED_TO_PLAYING` (`gst_data_repo_sink_open_file()`, which also requires `data_type` from `set_caps`), so `flexible.data` is never created.

The two binaries share more than the images — `unittest_datareposrc.cc` also reads bare `flexible.json` and `sparse.data`, which only `unittest_datareposink` ever writes — so this PR isolates the whole file namespace rather than renaming the known collisions. The helper lives in `unittest_util` because any test binary writing fixed names has the same exposure.

### Cause 2 — a wall-clock assertion over a variable number of samples

`create_flexible_tensors_test_file()` joins three `videotestsrc` branches (10 buffers each). `join` forwards EOS from whichever sinkpad is currently active (`gst_join_pad_event()`: `forward = (pad == active_sinkpad)`), so when a branch's EOS wins the race the other branches are cut off. Probing the fixture directly:

| condition | total_samples written |
|---|---|
| idle | 30, 30, 30, 30, 30, 30, 30, 30 |
| all cores loaded | 30, 28, 30, 17, 21, 30, 24, 22, 25, 26 |

The test assumed 30 samples at 30 fps (last PTS = 0.967 s) and compared against a hardcoded `0.8`. With 24 samples the stream is only 0.767 s long, so the assertion fails — which is exactly the 0.758597 s the CI log recorded. The `EXPECT_LT (elapsed_time, 0.05)` on the unsynced run had the same shape of problem in the other direction.

The fix counts the buffers actually read and derives the expected play time from them, so the test measures the framerate behaviour it is meant to measure regardless of how much the fixture wrote.

## Verification

| scenario | before | after |
|---|---|---|
| `--repeat 12`, parallel | 12/24 fail | **24/24 pass** |
| `--repeat 20`, all cores saturated | — | **40/40 pass** (buffer counts 18–30 in that run; the old assertion would have failed 11 times) |
| full local suite | — | **no regressions** (checks the shared `unittest_util` change) |

## Not in this PR

The `join` EOS behaviour in cause 2 is a real data-loss path for any pipeline that merges concurrent streams, not just this fixture. Changing when `join` forwards EOS is an element behaviour change that deserves its own PR and review, so this PR only makes the tests independent of it.

### DIFF
--- a/tests/nnstreamer_datarepo/unittest_datareposink.cc
+++ b/tests/nnstreamer_datarepo/unittest_datareposink.cc
@@ -665,6 +665,7 @@ int
 main (int argc, char **argv)
 {
   int result = -1;
+  gchar *work_dir;
 
   try {
     testing::InitGoogleTest (&argc, argv);
@@ -674,11 +675,17 @@ main (int argc, char **argv)
 
   gst_init (&argc, &argv);
 
+  /* These tests write fixed file names, which unittest_datareposrc also uses. */
+  /* Enter after InitGoogleTest, which anchors --gtest_output to the start directory. */
+  work_dir = enterPrivateWorkDir ();
+
   try {
     result = RUN_ALL_TESTS ();
   } catch (...) {
     g_warning ("catch `testing::internal::GoogleTestFailureException`");
   }
+
+  leavePrivateWorkDir (&work_dir);
 
   return result;
 }
--- a/tests/nnstreamer_datarepo/unittest_datareposrc.cc
+++ b/tests/nnstreamer_datarepo/unittest_datareposrc.cc
@@ -637,11 +637,12 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
 {
   gint fps = 30;
   guint64 start_time, end_time;
-  gdouble elapsed_time;
+  gdouble elapsed_time, stream_duration;
   GstElement *tensor_sink;
   GstBus *bus;
   GMainLoop *loop;
   gint file_index = 1;
+  gint buffer_count = 0, no_sync_count = 0;
   const gchar *str_pipeline
       = "datareposrc location=flexible1.data json=flexible1.json ! queue ! tensor_sink name=tensor_sink0 sync=true";
 
@@ -655,6 +656,10 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   gst_bus_add_watch (bus, bus_callback, loop);
   g_clear_pointer (&bus, gst_object_unref);
 
+  tensor_sink = gst_bin_get_by_name (GST_BIN (pipeline), "tensor_sink0");
+  ASSERT_NE (tensor_sink, nullptr);
+  g_signal_connect (tensor_sink, "new-data", G_CALLBACK (new_data_cb), &buffer_count);
+
   start_time = g_get_monotonic_time ();
 
   setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT);
@@ -664,9 +669,20 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   end_time = g_get_monotonic_time ();
   elapsed_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
 
-  g_print ("Elapsed time: %.6f second\n", elapsed_time);
-  EXPECT_LT (0.8, elapsed_time);
+  /**
+   * @brief The writer pipeline does not store a fixed number of samples: join
+   * forwards EOS from the pad that delivered the last buffer, so the remaining
+   * sources are cut off at a point that depends on scheduling. Derive the
+   * expected play time from the samples that were actually read. Fewer than a
+   * third of the 30 samples means the writer, not the timing, is broken.
+   */
+  ASSERT_GE (buffer_count, 10);
+  stream_duration = (buffer_count - 1) / (gdouble) fps;
 
+  g_print ("Elapsed time: %.6f second (%d buffers)\n", elapsed_time, buffer_count);
+  EXPECT_LT (stream_duration * 0.9, elapsed_time);
+
+  g_clear_pointer (&tensor_sink, gst_object_unref);
   g_clear_pointer (&pipeline, gst_object_unref);
   g_main_loop_unref (loop);
 
@@ -680,7 +696,9 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   g_clear_pointer (&bus, gst_object_unref);
 
   tensor_sink = gst_bin_get_by_name (GST_BIN (pipeline), "tensor_sink0");
+  ASSERT_NE (tensor_sink, nullptr);
   g_object_set (GST_OBJECT (tensor_sink), "sync", FALSE, NULL);
+  g_signal_connect (tensor_sink, "new-data", G_CALLBACK (new_data_cb), &no_sync_count);
 
   start_time = g_get_monotonic_time ();
 
@@ -691,8 +709,10 @@ TEST (datareposrc, fps30ReadFlexibleTensors)
   end_time = g_get_monotonic_time ();
   elapsed_time = (end_time - start_time) / (double) G_USEC_PER_SEC;
 
-  g_print ("Elapsed time: %.6f second\n", elapsed_time);
-  EXPECT_LT (elapsed_time, 0.05);
+  g_print ("Elapsed time: %.6f second (%d buffers)\n", elapsed_time, no_sync_count);
+  EXPECT_EQ (no_sync_count, buffer_count);
+  /* Without sync the same samples are read without waiting for the clock. */
+  EXPECT_LT (elapsed_time, stream_duration * 0.5);
 
   g_clear_pointer (&tensor_sink, gst_object_unref);
   g_clear_pointer (&pipeline, gst_object_unref);
@@ -1147,6 +1167,7 @@ int
 main (int argc, char **argv)
 {
   int result = -1;
+  gchar *work_dir;
 
   try {
     testing::InitGoogleTest (&argc, argv);
@@ -1156,11 +1177,17 @@ main (int argc, char **argv)
 
   gst_init (&argc, &argv);
 
+  /* These tests write fixed file names, which unittest_datareposink also uses. */
+  /* Enter after InitGoogleTest, which anchors --gtest_output to the start directory. */
+  work_dir = enterPrivateWorkDir ();
+
   try {
     result = RUN_ALL_TESTS ();
   } catch (...) {
     g_warning ("catch `testing::internal::GoogleTestFailureException`");
   }
+
+  leavePrivateWorkDir (&work_dir);
 
   return result;
 }
--- a/tests/unittest_util.c
+++ b/tests/unittest_util.c
@@ -92,6 +92,99 @@ void removeTempFile (char **file_name)
 }
 
 /**
+ * @brief Remove a directory with everything in it.
+ */
+static void
+removeDirRecursive (const gchar * path)
+{
+  GDir *dir;
+  const gchar *name;
+
+  dir = g_dir_open (path, 0, NULL);
+  if (dir == NULL)
+    return;
+
+  while ((name = g_dir_read_name (dir)) != NULL) {
+    gchar *child = g_build_filename (path, name, NULL);
+
+    if (g_file_test (child, G_FILE_TEST_IS_DIR)
+        && !g_file_test (child, G_FILE_TEST_IS_SYMLINK))
+      removeDirRecursive (child);
+    else
+      g_remove (child);
+
+    g_free (child);
+  }
+
+  g_dir_close (dir);
+  g_rmdir (path);
+}
+
+static gchar *private_work_dir_origin = NULL; /**< Directory to return to. */
+
+/**
+ * @brief Enter a private working directory for the calling test binary.
+ * @return the new working directory, or NULL if the move failed
+ */
+gchar *
+enterPrivateWorkDir (void)
+{
+  gchar *work_dir;
+  gchar *cwd = g_get_current_dir ();
+
+  /* Tests locate their data with this, falling back to a relative path. */
+  if (g_getenv ("NNSTREAMER_SOURCE_ROOT_PATH") == NULL) {
+    gchar *root = g_build_filename (cwd, "..", NULL);
+
+    g_setenv ("NNSTREAMER_SOURCE_ROOT_PATH", root, TRUE);
+    g_free (root);
+  }
+
+  work_dir = g_dir_make_tmp ("nnstreamer_unittest_workdir_XXXXXX", NULL);
+  if (work_dir == NULL) {
+    g_warning ("failed to create a private working directory, "
+               "tests sharing file names may interfere with each other");
+    g_free (cwd);
+    return NULL;
+  }
+
+  if (g_chdir (work_dir) != 0) {
+    g_warning ("failed to enter the private working directory %s, "
+               "tests sharing file names may interfere with each other", work_dir);
+    g_rmdir (work_dir);
+    g_free (work_dir);
+    g_free (cwd);
+    return NULL;
+  }
+
+  g_free (private_work_dir_origin);
+  private_work_dir_origin = cwd;
+
+  return work_dir;
+}
+
+/**
+ * @brief Return to the entry directory and remove the private working directory.
+ */
+void
+leavePrivateWorkDir (gchar ** work_dir)
+{
+  if (work_dir == NULL || *work_dir == NULL)
+    return;
+
+  /* Move out of the directory before removing it; both steps are best effort. */
+  if (private_work_dir_origin != NULL)
+    (void) g_chdir (private_work_dir_origin);
+
+  removeDirRecursive (*work_dir);
+
+  g_free (private_work_dir_origin);
+  private_work_dir_origin = NULL;
+  g_free (*work_dir);
+  *work_dir = NULL;
+}
+
+/**
  * @brief Wait until the pipeline processing the buffers
  * @return TRUE on success, FALSE when a time-out occurs
  */
--- a/tests/unittest_util.h
+++ b/tests/unittest_util.h
@@ -51,6 +51,29 @@ extern gchar * getTempFilename (void);
 extern void removeTempFile (char **file_name);
 
 /**
+ * @brief Enter a private working directory for the calling test binary.
+ * @details meson runs test binaries in parallel with the build directory as
+ *          their working directory, so binaries that create files with fixed
+ *          names overwrite and delete each other's files. Call this before
+ *          running the tests to give the process a directory of its own.
+ *          NNSTREAMER_SOURCE_ROOT_PATH is made absolute first, so that test
+ *          data is still found after the move.
+ * @note Call this after testing::InitGoogleTest(), which resolves a relative
+ *       --gtest_output against the directory the test was started from.
+ *       Entering earlier would write the report into the private directory,
+ *       which leavePrivateWorkDir() then removes.
+ * @return the new working directory, or NULL if the move failed
+ *         (should be released with leavePrivateWorkDir)
+ */
+extern gchar * enterPrivateWorkDir (void);
+
+/**
+ * @brief Return to the directory enterPrivateWorkDir() was called from and
+ *        remove the private working directory with its contents.
+ */
+extern void leavePrivateWorkDir (gchar **work_dir);
+
+/**
  * @brief Wait until the pipeline processing the buffers
  * @return TRUE on success, FALSE when a time-out occurs
  */


myungjoo:
> **Note:** This review was produced by an automated Claude sub-agent (a background code-review agent, not a human maintainer). It is being transcribed here verbatim as a review comment; please weigh it accordingly and verify anything surprising before acting on it.

## Summary

The fix correctly diagnoses and resolves the two documented root causes: (1) two datarepo test binaries writing/deleting the same fixed filenames while meson runs them concurrently with the build dir as CWD, and (2) the `fps30ReadFlexibleTensors` timing assertion hard-coding an assumed sample count that `join`'s EOS-forwarding makes variable. Giving each binary a private `g_dir_make_tmp()` working directory is a reasonable, minimal-footprint fix, and deriving the expected duration from the actually-observed `buffer_count` is the right way to detensionize the timing assertion from `join`'s flakiness. The diff is appropriately scoped (2 test files + the shared helper it needed) and I found no ABI/symbol-collision risk for the other 30 binaries linking `nnstreamer_unittest_util` (none of them reference the new symbols).

However, I found one **Major** issue: the new `enterPrivateWorkDir()`/`leavePrivateWorkDir()` pair changes the test binary's CWD for *every* invocation path, not just `meson test`'s parallel one — and this silently breaks GTest's XML report generation in the Tizen/GBS packaging test flow (`packaging/run_unittests_binaries.sh`).

---

### [Major] GTest `--gtest_output` XML report is misdirected and then deleted under the Tizen GBS `unit_test` flow

`packaging/nnstreamer.spec` (line ~1064-1065) runs `bash %{test_script} ./tests/nnstreamer_datarepo` when `%{?unit_test}` is enabled, and `packaging/run_unittests_binaries.sh` (line 48-80) does:

```sh
pushd build
...
${entry} --gtest_output="xml:${entry##*/}.xml"
```

i.e. it launches `unittest_datareposink`/`unittest_datareposrc` with CWD = `build/`, expecting `unittest_datareposink.xml` etc. to land there.

But in the new `main()`:
```c
work_dir = enterPrivateWorkDir ();      // chdir()s into a private g_dir_make_tmp() dir
result = RUN_ALL_TESTS ();              // GTest's XmlUnitTestResultPrinter writes the XML
                                         // *inside* RUN_ALL_TESTS(), i.e. while CWD is the
                                         // private tmp dir, before this call returns
leavePrivateWorkDir (&work_dir);        // chdir()s to g_get_tmp_dir(), then
                                         // removeDirRecursive(work_dir) — deletes the
                                         // just-written XML along with everything else
```

GTest resolves a relative `--gtest_output=xml:...` path against the process's current working directory *at the moment the report is generated*, which happens inside `RUN_ALL_TESTS()` — before `leavePrivateWorkDir()` runs. Because of the `chdir()` in `enterPrivateWorkDir()`, that moment now occurs inside the private tmp directory instead of `build/`. The result:
1. `build/unittest_datareposink.xml` / `build/unittest_datareposrc.xml` are never created where `run_unittests_binaries.sh` expects them.
2. The copy that *was* written (inside the private tmp dir) is then deleted by `removeDirRecursive()` before the process exits.

Net effect: for these two binaries, the GBS/OBS packaging test run silently produces no XML test report at all. `packaging/nnstreamer.spec` doesn't appear to hard-require the XML for the RPM build to succeed, so this won't turn the GBS build red — it will just silently drop the test-result artifact for exactly the two binaries this PR touches, which is easy to miss until someone goes looking for the report.

Also worth a quick check (not confirmed, lower confidence): if `-Db_coverage=true` (used in the Tizen coverage-enabled spec branch), `.gcda` files are flushed via `atexit`, i.e. *after* `main()` returns — at which point `leavePrivateWorkDir()` has already moved CWD to `g_get_tmp_dir()`. If gcov's data-file paths are ever resolved relative to CWD rather than baked in as absolute at compile time, coverage data for these two binaries could also land in/relative-to the wrong place. I did not verify GCC's exact behavior here, so treat this as a "worth double-checking," not a confirmed bug.

**Why this matters / suggested fix:** The root cause reported in this PR (root cause 1) is specifically about `meson test`'s *parallel* scheduling — GitHub Actions CI uses `meson test` directly, and `run_unittests_binaries.sh` runs binaries strictly sequentially (no backgrounding), so it was never exposed to the original race. A fix scoped to `meson test` — e.g. giving each `test()` entry in `tests/nnstreamer_datarepo/meson.build` its own `workdir:` (a meson `test()` kwarg that sets the CWD for that test run) — would solve the reported flakiness without changing the binaries' behavior for the GBS script or for a developer running the binary directly from `build/` (which also currently loses the convenience of finding `flexible1.data`/`img_*.png` next to the binary after a local run, since they're now generated in an auto-deleted tmp dir). If the current `chdir()`-in-`main()` approach is kept instead, at minimum the private-workdir logic needs to either run only after `--gtest_output` has been written (not currently controllable) or `leavePrivateWorkDir()` needs to copy any generated `*.xml`/coverage artifacts back to the original working directory before deleting the private dir.

---

### [Minor] Silent degradation when `g_dir_make_tmp()`/`g_chdir()` fail in `enterPrivateWorkDir()`

```c
work_dir = g_dir_make_tmp (...);
if (work_dir == NULL) {
  _print_log ("failed to create a private working directory");
  return NULL;
}
if (g_chdir (work_dir) != 0) {
  _print_log ("failed to enter the private working directory %s", work_dir);
  ...
  return NULL;
}
```

`_print_log()` is `#define _print_log(...) do { if (DBG) g_message(...); } while (0)`, and `DBG` defaults to `FALSE`. So on a runner where `TMPDIR`/`g_dir_make_tmp()` fails (disk full, restrictive sandbox, read-only tmp, etc.), the test binary falls back to running in the shared build directory with **zero visible diagnostic** in a default CI log — silently reintroducing the exact race condition (root cause 1) this PR exists to fix, with no signal that it happened. Given the whole point of this PR is to eliminate a flaky-CI source, this specific failure path deserves an unconditional `g_warning()` (or at least a `g_message()` without the `DBG` gate) rather than the same helper used for routine diagnostics.

---

### [Minor] `leavePrivateWorkDir()` has no fallback if `g_chdir(g_get_tmp_dir())` itself fails

```c
if ((tmp_dir = g_get_tmp_dir ()) != NULL)
  (void) g_chdir (tmp_dir);
removeDirRecursive (*work_dir);
```

The `g_chdir()` return value is intentionally discarded (`(void)`), so if it fails, `removeDirRecursive(*work_dir)` proceeds while the process's CWD may still be *inside* the directory being recursively deleted. This is a very unlikely failure (system tmp dir normally exists/is writable), but the outcome is platform-dependent — POSIX generally tolerates removing your own CWD, but this repo's test suite is also built/run on Windows (per this session's environment), where deleting a directory that is a process's current working directory can fail outright (sharing violation), leaving the private tmp dir (and its contents) behind. Low likelihood, but since the whole point of the new code is cleanup hygiene, a comment or an explicit check would be worth it.

---

### [Minor] Second timing margin in `fps30ReadFlexibleTensors` could get unrealistically tight at very low `buffer_count`

```c
ASSERT_GT (buffer_count, 1);
stream_duration = (buffer_count - 1) / (gdouble) fps;
...
EXPECT_LT (elapsed_time, stream_duration * 0.5);
```

`stream_duration` scales linearly down to near 0 as `buffer_count` approaches its allowed floor of 2 (`stream_duration = 1/30 ≈ 0.033s`, so the bound becomes `elapsed_time < 0.0167s`). The second segment's `elapsed_time` includes fixed pipeline construction/state-change/teardown overhead that does **not** shrink proportionally with the file's sample count, so a pathologically low `buffer_count` (well below the PR's own measured 17–30 range) could make this bound fail even though nothing is actually wrong. Given the PR's own measurements (17–30 samples even under full CPU load), this is not expected to trigger in practice, but the `ASSERT_GT(buffer_count, 1)` guard doesn't structurally prevent it, so it's a latent (much less likely) source of exactly the kind of flakiness this PR is trying to eliminate. Consider either a slightly higher floor for `ASSERT_GT` (e.g. `>= 10`) or an additive fixed-overhead term in the second bound (e.g. `stream_duration * 0.5 + margin_const`).

---

### Verified as correct / no issue found

- **`NNSTREAMER_SOURCE_ROOT_PATH` fallback semantics**: `enterPrivateWorkDir()` computes `root = g_build_filename(cwd, "..", NULL)` using the CWD *before* the `chdir()` into the private dir, and only when the env var isn't already set. This correctly reproduces `get_file_path()`'s existing `".."` relative-path fallback (both resolve to the same parent-of-build-dir target), so there's no semantic drift for the "run the binary directly from the build dir" use case. In the normal `meson test` path this branch is dead code anyway, since `testenv` (top-level `meson.build` line 983) always sets `NNSTREAMER_SOURCE_ROOT_PATH` to an absolute path — likewise for the Tizen GBS `.spec`/`run_unittests_binaries.sh` paths, which export it before invoking any binaries.
- **`removeDirRecursive()` symlink handling**: `g_file_test(child, G_FILE_TEST_IS_DIR) && !g_file_test(child, G_FILE_TEST_IS_SYMLINK)` correctly avoids recursing through a symlinked directory (which could otherwise walk outside the private tmp dir or loop); such entries fall to `g_remove()`, which unlinks the symlink itself rather than following it. Dangling symlinks are also handled correctly (`IS_DIR` on a broken symlink is false, so it's `g_remove()`'d as a leaf). `g_dir_read_name()` already skips `.`/`..` per GLib's documented contract, so no manual filtering is needed there.
- **Resource cleanup on `g_dir_make_tmp`/`g_chdir` failure**: the `g_chdir()` failure branch correctly `g_rmdir()`s and `g_free()`s the directory it just created before returning `NULL` — no leak of the allocated string or the just-created (now-empty) directory.
- **`main()` call placement**: `enterPrivateWorkDir()` is called after `gst_init()` in both files (so plugin scanning etc. happens before CWD changes), and `leavePrivateWorkDir()` is called after `RUN_ALL_TESTS()` returns (including after a caught `testing::internal::GoogleTestFailureException`), so cleanup runs on both the normal and the caught-exception path. `work_dir` is unconditionally assigned right after `gst_init()`, so there's no uninitialized-variable read even though it's declared before the `try` block.
- **Scope/size**: the change is limited to the two affected binaries plus the shared helper they need; no unrelated module is touched, and no other of the 32 binaries linking `nnstreamer_unittest_util` reference the new symbols, so there's no risk of behavior change or symbol collision for them (including the Tizen GBS `unit_test` build, aside from the XML issue above which is specific to these two binaries).

---

### [Nit] No direct unit test for the new helper functions themselves

`enterPrivateWorkDir()`/`leavePrivateWorkDir()`/`removeDirRecursive()` have no dedicated test coverage (e.g. in `tests/common/unittest_common.cc`), so a future regression in the recursive-delete or chdir logic wouldn't be caught directly — it would only surface indirectly as flakiness in the two datarepo tests again. That said, this matches the existing convention in this file (`getTempFilename()`/`removeTempFile()` aren't directly tested either), so I'm not blocking on it, just flagging it as a gap given the review's emphasis on regression detection.

---

## Verdict

The core fix is sound and directly addresses the two documented root causes with a well-scoped diff. I would not consider this mergeable as-is, though, until the **Major** finding is addressed (or explicitly accepted as a known trade-off) — the GBS/Tizen packaging path silently loses its GTest XML report for these two binaries. The **Minor** findings are worth a follow-up but are not blocking. Recommend requesting changes rather than an unconditional approve.

myungjoo:
Author response to the automated review above (drafted with AI assistance). Head is now `efbb583a`, rebased onto current main.

**1. [Major] gtest XML report lost in the GBS `unit_test` path — not reproducible; the report is written where the runner expects it.**

Ran exactly what `packaging/run_unittests_binaries.sh:80` does — relative `--gtest_output`, cwd = build dir — against the patched binaries:

```
$ ./tests/nnstreamer_datarepo/unittest_datareposink --gtest_output="xml:unittest_datareposink.xml"
$ ls -la unittest_datarepo*.xml
-rw-r--r-- 1 ... 2971 unittest_datareposink.xml
-rw-r--r-- 1 ... 5162 unittest_datareposrc.xml
$ grep -o '<testsuites[^>]*tests="[0-9]*"[^>]*failures="[0-9]*"' unittest_datareposink.xml
<testsuites tests="12" failures="0"
```

Both reports land in the build directory with complete results, and no temp directory is left behind. The reason is ordering: gtest resolves `--gtest_output` to an absolute path inside `InitGoogleTest()` (`PostFlagParsingInit()` → `ConfigureXmlOutput()` → `UnitTestOptions::GetAbsolutePathToOutputFile()`), which stores the path in the XML printer long before `RUN_ALL_TESTS()` writes it. `enterPrivateWorkDir()` is called after `InitGoogleTest()`, so the path is already anchored to the original directory.

That ordering was load-bearing but undocumented, which is a fair thing to have flagged — both `main()` functions now carry a comment stating it, so a future refactor that moves the call above `InitGoogleTest()` has a warning attached.

On the alternative suggested (meson `workdir:`): it would confine the change to `meson test`, but meson has no portable way to create the directory at configure time, and the isolation would not cover a developer running the binary directly — which is how the collision was first hit locally.

**2. [Minor] Silent degradation when isolation fails — fixed.** `_print_log` is compiled out by default (`DBG FALSE`), so a failure to create or enter the directory would have silently restored the racy behaviour. Both paths now use `g_warning()` and say what the consequence is. `G_DEBUG` in CI is `gc-friendly`, not `fatal-warnings`, and these `main()`s already use `g_warning`, so this cannot turn into an abort.

**3. [Minor] `leavePrivateWorkDir()` ignores the chdir result — documented.** The removal is deliberately best effort: cleanup must not fail a test run, and on POSIX removing the directory one is standing in succeeds anyway. The comment now says so.

**4. [Minor] Tight fast-path margin near the `buffer_count` floor — fixed.** With `buffer_count == 2` the unsynced bound would have been 0.017 s, which fixed pipeline overhead could exceed. The floor is now `ASSERT_GE (buffer_count, 10)`: at 10 samples the bound is 0.15 s against the 0.004–0.015 s actually measured, and anything below a third of the expected 30 samples means the writer fixture is broken, which is worth failing loudly rather than measuring. The stress run that produced 18–30 samples with every core saturated stays well clear of the floor.

**5. [Nit] No unit tests for the new helpers.** Agreed it matches the existing convention for `getTempFilename`/`removeTempFile`. The helpers are exercised by every run of both datarepo binaries, and their failure mode is loud now (point 2).

Re-verified after these changes: 24/24 parallel runs pass, and the XML reports above were produced by the patched binaries.

myungjoo:
> **Note:** This review was produced by an automated Claude sub-agent (a background code-review agent, not a human maintainer), re-reviewing the author's rebuttal/fix round at head `efbb583a`. It is being transcribed here verbatim as a review comment; please weigh it accordingly and verify anything surprising before acting on it.

## Re-review of the prior round (Major 1 / Minor 3 / Nit 1)

### [Major] gtest XML report lost in the GBS `unit_test` path — **Withdrawn, confirmed by direct source inspection**

I independently verified the author's rebuttal against the actual GoogleTest source rather than taking the empirical run on faith. Downloaded and grepped both `googletest/src/gtest.cc` + `gtest-internal-inl.h` at **release-1.11.0** (the version `apt`-installed on `ubuntu-22.04`, matching what GitHub Actions CI uses) and **v1.14.0** (the author's local verification version). Both versions have byte-for-byte identical control flow in the relevant area:

- `UnitTestImpl::AddTestInfo()` (`gtest-internal-inl.h`) captures `original_working_dir_ = FilePath::GetCurrentDir()` the first time any `TEST()`/`TEST_F()` registers itself — this happens via static/global initialization **before `main()` is even entered**, let alone before any `chdir()` inside it.
- `testing::InitGoogleTest()` → `internal::InitGoogleTestImpl()` → `ParseGoogleTestFlagsOnly()` (parses `--gtest_output`) → `GetUnitTestImpl()->PostFlagParsingInit()` → `ConfigureXmlOutput()`, all called synchronously and unconditionally inside `InitGoogleTest()`.
- `ConfigureXmlOutput()` immediately constructs `new XmlUnitTestResultPrinter(UnitTestOptions::GetAbsolutePathToOutputFile().c_str())`, and `GetAbsolutePathToOutputFile()` resolves any relative `--gtest_output=xml:...` path against `UnitTest::GetInstance()->original_working_dir()` (the pre-`main()` CWD captured above) — **not** the CWD at the moment the file is actually written.
- So the absolute output path is baked into the printer object before `InitGoogleTest()` even returns.

I then confirmed in the local diff that both `main()`s call `testing::InitGoogleTest(&argc, argv)` **before** `gst_init()` and before `enterPrivateWorkDir()`, exactly matching the call order the rebuttal describes (`tests/nnstreamer_datarepo/unittest_datareposink.cc:665-694`, `tests/nnstreamer_datarepo/unittest_datareposrc.cc:1163-1190`). Given that ordering, the `chdir()` inside `enterPrivateWorkDir()` happens strictly after the XML output path is already resolved to an absolute path, so it cannot retarget or lose the report. The author's empirical run (12/22-plus tests reporting into `build/*.xml`, no leftover temp dirs) is exactly what this code path predicts.

**Verdict: the Major finding does not hold; withdrawn/resolved.** The added `main()` comments documenting the ordering dependency are a reasonable (if belt-and-suspenders) safeguard against a future refactor breaking this invariant.

### [Minor] Silent degradation on `g_dir_make_tmp`/`g_chdir` failure — **Resolved**
Confirmed in `tests/unittest_util.c:142-155`: both failure branches now call unconditional `g_warning()` (not the `DBG`-gated `_print_log()`), each stating the consequence ("tests sharing file names may interfere with each other"). Checked CI (`ubuntu_clean_meson_build.yml:171`) — `G_DEBUG` is set to `gc-friendly` for the one job that sets it at all, never `fatal-warnings`, and both `main()`s already use `g_warning()` elsewhere (the `catch (...)` blocks), so this is consistent with existing practice and introduces no new abort risk.

### [Minor] `leavePrivateWorkDir()` ignores `g_chdir()` failure — **Accepted (documented trade-off)**
`tests/unittest_util.c:171` now carries an explicit "both steps are best effort" comment. Given cleanup failure here is not supposed to fail the test run, and the residual risk is low-probability/POSIX-only-adjacent, documenting the trade-off is an adequate response; not blocking.

### [Minor] Tight fast-path margin near the `buffer_count` floor — **Resolved, and the new floor is well-justified**
Verified `tests/nnstreamer_datarepo/unittest_datareposrc.cc:636-719`. The floor is now `ASSERT_GE (buffer_count, 10)`, and I traced why 10 is not an arbitrary/probabilistic choice: `create_flexible_tensors_test_file()` (same file, lines 105-136) drives three `videotestsrc num-buffers=10` branches into a `join`; each branch independently produces exactly 10 buffers. Since `join` can only forward as few buffers as the *fastest-finishing single branch* contributes before forwarding EOS, 10 is a **structural lower bound** on `buffer_count`, not just an empirically-observed one — a value below 10 can only happen if a branch itself failed to produce its 10 buffers, i.e. the writer pipeline is genuinely broken. The author's own stress-test range (17-30 buffers under full core saturation) sits comfortably above this structural floor, so raising it to 10 does not introduce new flakiness; if anything it now fails fast and loudly (via `ASSERT_GE`) on a real regression instead of silently producing a near-zero timing bound. The derived second-segment bound (`elapsed_time < stream_duration * 0.5`) at the floor works out to ~0.15s against a measured ~0.004-0.015s of fixed overhead, leaving a healthy margin. Math and reasoning check out.

### [Nit] No dedicated unit test for the new helpers — **Accepted**
Matches existing convention (`getTempFilename`/`removeTempFile` also untested directly), and the failure mode is now loud (per the Minor-1 fix above). Not blocking, as before.

## New issues from re-review

None found. I re-read the full diff (`tests/unittest_util.c`, `tests/unittest_util.h`, both datarepo test files) and re-checked scope: `tests/meson.build` shows the 32 other binaries linking `nnstreamer_unittest_util` don't reference `enterPrivateWorkDir`/`leavePrivateWorkDir`, so there's no behavior change or symbol-collision risk for them. `meson.build:983` confirms `NNSTREAMER_SOURCE_ROOT_PATH` is already set by `testenv` for the standard `meson test` path, so the fallback branch in `enterPrivateWorkDir()` is dead code there exactly as previously verified, and remains correct for the direct-invocation / GBS cases. No `workdir:` kwarg was added to `tests/nnstreamer_datarepo/meson.build` (the alternative suggested last round), but the author's reasoning for preferring the in-process `chdir()` — it also covers a developer invoking the binary directly, which is how the original collision was hit — is sound and doesn't leave the meson-only fix's gap unaddressed for its actual goal.

## Verdict

All prior findings are resolved, withdrawn, or accepted as documented trade-offs, and no new issues surfaced. **LGTM — ok to merge** from this automated review's perspective.

myungjoo:
Update (drafted with AI assistance). Head is now `f94bc6ee`.

**CI failure fixed.** `Static checks` failed on the doxygen-tag checker, and it was caused by this PR: `doxygen-tag.sh:124` rejects any comment that opens with `/*` and does not close on the same line, so the four multi-line comments added here were flagged (`unittest_datareposink.cc:678`, `unittest_datareposrc.cc:672,708,1176`). They are now either single-line comments or proper `/**` blocks. Simulating the checker's exact condition over the changed files reports nothing, and the tests still pass (24/24 parallel runs).

**Superseding #4872.** #4872 fixes `fps30ReadFlexibleTensors` with the same root-cause analysis (join forwards EOS from the pad that delivered the last buffer, so the stored frame count varies) and the same approach (derive the expected play time from the frames actually read). It does not touch the shared working directory, so `datareposink.writeFlexibleTensors_n` — the other half of the flakiness, which fails through a different mechanism — stays broken there.

To make this PR a strict superset, the one check #4872 had that this PR lacked is now included: the unsynced run counts its buffers too and `EXPECT_EQ (no_sync_count, buffer_count)` asserts both runs read the same frames. The remaining differences are margins, and this PR's are the tighter ones: it bounds the synced run at `(n-1)/fps * 0.9` against #4872's `n/fps * 0.7` (the last frame renders at `(n-1)/fps`, so that is the exact expectation), and it requires at least 10 frames so the unsynced bound cannot collapse toward the fixed pipeline overhead.

With that, #4872 can be closed as covered here.

myungjoo:
> **Note:** This review was produced by an automated Claude sub-agent (a background code-review agent, not a human maintainer), re-reviewing this PR after head moved from `efbb583a` (previously given "LGTM — ok to merge" by a prior automated pass) to `f94bc6ee`. It is being transcribed here verbatim as a review comment; please weigh it accordingly and verify anything surprising before acting on it.

## Scope of this round

`git diff efbb583a f94bc6ee` touches exactly two things across the same four files reviewed before (`tests/unittest_util.{c,h}`, both datarepo test `.cc` files): (a) reformatting four `/* ... */` block comments to fix the `Static checks` (doxygen-tag) CI failure, and (b) absorbing #4872's `no_sync_count`/`EXPECT_EQ` verification into `fps30ReadFlexibleTensors`. Nothing else changed. Both are verified below.

## (a) CI fix — verified correct

Reproduced `doxygen-tag.sh`'s exact check (lines 113-127, the `/* ...` rule at line 124 plus the uppercase-tag rule) against the four changed files, using the content as actually stored in git (`git show <rev>:<path>`, LF line endings — the local worktree is CRLF due to `core.autocrlf=true` but no `.gitattributes` forces this on checkout, and there's no CRLF in the committed blobs, so this matches what the Ubuntu CI runner lints):

- At `efbb583a`, my emulation reproduces exactly the four reported violations and nothing else: `unittest_datareposink.cc:678`, `unittest_datareposrc.cc:672`, `unittest_datareposrc.cc:708`, `unittest_datareposrc.cc:1176` — this matches the actual failing CI run's log (`run 32983207698`) byte-for-byte.
- At `f94bc6ee`, the emulation reports zero violations across all four files. I also confirmed directly against GitHub: the `Static checks` job on the current head (`run 32985132366`) is green.
- No new violation was introduced elsewhere: the four comments were either split into two single-line `/* ... */` comments (each opened and closed on the same line) or turned into proper `/** ... */` Doxygen blocks — both forms this checker accepts.
- The ctags-based per-function `@brief` check (`advanced=1`, active via `static.check.yml:85`) is unaffected: no new functions were added in this round (`removeDirRecursive`/`enterPrivateWorkDir`/`leavePrivateWorkDir` already existed as of the previously-approved `efbb583a` and were already confirmed to carry `@brief` tags in the prior review round; the failing CI run at `efbb583a` reported only the four `/*`-rule errors, confirming the ctags-based check was already clean then and is untouched now).

## (b) Comment rewrite — no loss of meaning

Diffed `efbb583a..f94bc6ee` directly:
- Both `main()`s: the invariant "`enterPrivateWorkDir()` must run after `InitGoogleTest()` because that's what anchors `--gtest_output` to the start directory" — the fact independently verified via GoogleTest source in the prior review round — is still stated, just split into two single-line comments immediately above the `enterPrivateWorkDir()` call instead of one multi-line block. Content is preserved verbatim, just reflowed.
- The `ASSERT_GE(buffer_count, 10)` rationale comment became a proper `/** @brief ... */` block; the new wording is slightly more detailed (explicitly names `join`'s EOS-forwarding behavior as the cause) with no information dropped.

## (c) #4872 absorption — verified correct, no new flakiness

- `no_sync_count` is counted via the same well-established `new_data_cb` "new-data" signal-counting pattern already used for `buffer_count` (and reused elsewhere in this file, e.g. `readImageFiles`), connected to the *second* pipeline's `tensor_sink` (line ~701) right after it's fetched via `gst_bin_get_by_name`. `EXPECT_EQ (no_sync_count, buffer_count)` (line ~713) sits right after the second run's teardown/timing capture and before the timing `EXPECT_LT`, matching #4872's check and placement intent.
- Both runs read the *same* `flexible1.data`/`flexible1.json`, written once by `create_flexible_tensors_test_file()` before either read pipeline starts and deleted only at the very end of the test. The read-side pipeline is a plain `datareposrc ! queue ! tensor_sink` — no `join`/race-prone element on the read path (that's only in the writer pipeline, and only runs once, before either read). So `buffer_count == no_sync_count` is a deterministic invariant of reading a static file twice, not a timing-sensitive assertion — I don't expect this to introduce new flakiness.
- Margin-comparison claim: verified the specific claim in the PR description — for the enforced floor `buffer_count >= 10`, `(n-1)/fps*0.9 >= n/fps*0.7` holds for all `n >= 10` (difference `= 0.2n - 0.9`, minimum `1.1 > 0` at `n=10`, growing with `n`), so this PR's synced-run lower bound is provably at least as strict as #4872's for every value `buffer_count` can now take. That part of the "tighter margins" claim checks out exactly as stated.
  - One FYI nuance not called out in the PR description: the no-sync *upper* bound isn't a strict formal superset of #4872's `no_sync_time * 3 < sync_time` in the same way — this PR bounds `no_sync_time` against a fraction of the *derived* `stream_duration` rather than the *measured* `sync_time`, and at larger buffer counts (e.g. the 17-30 range the author measured under load) `stream_duration * 0.5` can numerically exceed `sync_time / 3`, i.e. slightly looser there. In practice this is inconsequential — `no_sync_time` is essentially fixed pipeline overhead (single-digit ms per the author's own measurements) regardless of `n`, so both bounds pass with a wide margin either way — but flagging it since margin coverage was explicitly asked about.

## New issues from this round

None found (Critical/Major/Minor/Nit). The diff since the last "LGTM" is exactly the two targeted fixes described in the PR update, both verified correct above, and CI now confirms the `Static checks` job that motivated this round is green.

## Verdict

**LGTM — ok to merge.** On closing #4872: this PR's `fps30ReadFlexibleTensors` now covers everything #4872 changed in that test (same root-cause fix, same `no_sync_count`/`EXPECT_EQ` check, and a synced-run lower bound that is provably at least as tight for the buffer counts this test can produce) plus the separate `datareposink.writeFlexibleTensors_n`/working-directory-isolation fix #4872 doesn't touch — I agree #4872 can be closed as superseded by this PR.

myungjoo:
Applied the two actionable suggestions from the approving review (drafted with AI assistance). Head is now `d873e00b`.

**1. The ordering constraint is now in the header.** `enterPrivateWorkDir()`'s doxygen block carries an `@note` saying it must be called after `testing::InitGoogleTest()`, and why: gtest resolves a relative `--gtest_output` against the directory the test started in, so entering earlier would write the report into the private directory that `leavePrivateWorkDir()` then removes. The call-site comments stay, but the header is what a future caller reads.

**2. Cleanup returns to the entry directory.** `enterPrivateWorkDir()` records the working directory it was called from and `leavePrivateWorkDir()` chdirs back to it instead of to `g_get_tmp_dir()`. The cwd was already being read for the `NNSTREAMER_SOURCE_ROOT_PATH` fallback, so this just keeps it; both failure paths free it.

**3. The `ASSERT_GE (buffer_count, 10)` floor is left as is, deliberately.** That suggestion says to revisit it once the writer fixture becomes deterministic, and #4875 is exactly that change: it fixes `join` to forward EOS only after every sink pad ends and then asserts the exact sample count. Loosening the floor here would only widen the window before that lands, and tightening it would reintroduce the flakiness this PR removes. Note that #4875 edits the same test function, so whichever merges second needs a rebase — the natural order is this PR first, then #4875 replacing the floor with its exact equality.

Re-verified on a clean checkout of `d873e00b`: 24/24 parallel runs pass, the full local suite passes, `--gtest_output` still lands in the build directory, and no temporary working directory is left behind.

myungjoo:
> **Note:** This review was produced by an automated Claude sub-agent (a background code-review agent, not a human maintainer), re-reviewing this PR after head moved from `f94bc6ee` to `d873e00b` in response to the three non-blocking suggestions left by the prior approving review (`myungjoo-bot`). It is being transcribed here verbatim as a review comment; please weigh it accordingly and verify anything surprising before acting on it.

## Scope of this round

`git diff f94bc6ee d873e00b` touches only `tests/unittest_util.c` and `tests/unittest_util.h`; no other file changed (the two datarepo `.cc` test files are untouched, matching suggestion 3 being deliberately deferred rather than applied). Verified directly against the diff and by reading the current file contents.

## Suggestion 1 - ordering constraint documented in the header - Resolved

`tests/unittest_util.h`'s `enterPrivateWorkDir()` doxygen block now carries:
```
@note Call this after testing::InitGoogleTest(), which resolves a relative
      --gtest_output against the directory the test was started from.
      Entering earlier would write the report into the private directory,
      which leavePrivateWorkDir() then removes.
```
This matches the gtest control flow independently verified in an earlier round of this review (`InitGoogleTestImpl()` -> `ConfigureXmlOutput()` -> `GetAbsolutePathToOutputFile()`, which bakes the absolute output path in before `InitGoogleTest()` returns). The wording is a reasonable practical simplification - the anchor directory is technically captured even earlier via `UnitTestImpl::AddTestInfo()` at static-init time, before `main()` runs - but "call after `InitGoogleTest()`" is the correct and sufficient instruction for a caller, since the absolute path is resolved by the time `InitGoogleTest()` returns either way. Both call sites (`unittest_datareposink.cc:679-680`, `unittest_datareposrc.cc:1181-1182`) still call `enterPrivateWorkDir()` after `testing::InitGoogleTest()`, so behavior matches the new doc.

## Suggestion 2 - `leavePrivateWorkDir()` returns to the entry directory - Resolved, implementation verified

Read the full new implementation (`tests/unittest_util.c:123-185`). Checked the specific correctness properties called out for this round:

- **Lifecycle of `private_work_dir_origin`**: file-scope `static gchar *`, initialized `NULL`. Set only on the success path of `enterPrivateWorkDir()` (after both `g_dir_make_tmp()` and `g_chdir()` succeed), and always freed and reset to `NULL` at the end of `leavePrivateWorkDir()`. No leak on the normal path.
- **Both failure paths of `enterPrivateWorkDir()`** (`g_dir_make_tmp()` fails; `g_chdir()` fails) now `g_free(cwd)` before returning `NULL`, and leave `private_work_dir_origin` untouched - no leak of the newly-fetched `cwd` in either failure branch.
- **Reentrancy**: on a second call to `enterPrivateWorkDir()` without an intervening `leavePrivateWorkDir()`, the success path does `g_free(private_work_dir_origin); private_work_dir_origin = cwd;` - so a stale previous origin is freed, not leaked. Note this is only about avoiding a *leak*: since the second call's `cwd` would be read from inside the first call's private dir (the process never left it), the *origin value* itself would end up wrong for unwinding two nested calls with one `leavePrivateWorkDir()`. This is a real latent fragility if the helper is ever called reentrantly, but it is not exercised by the current callers - each of the two `main()`s calls `enterPrivateWorkDir()`/`leavePrivateWorkDir()` exactly once, so this is a non-blocking observation about future misuse, not a bug in the shipped code.
- **`leavePrivateWorkDir()` chdir-failure path**: unchanged trade-off from the previous round, just now targeting `private_work_dir_origin` instead of `g_get_tmp_dir()`. The `g_chdir()` return value is still discarded and `removeDirRecursive()` still proceeds regardless, with the "best effort" comment retained. Swapping the target from the system tmp dir (almost always present/writable) to the original entry directory doesn't meaningfully change the risk profile in practice for these two callers (the entry directory is the build directory, which necessarily exists for the test binary to have been invoked from it), and this exact trade-off was already reviewed and accepted as documented in the prior round.
- **Thread safety**: the new static is unguarded, but `enterPrivateWorkDir()`/`leavePrivateWorkDir()` are each called exactly once from `main()`, before/after `RUN_ALL_TESTS()`, in both binaries - single-threaded with respect to this state. Not a concern for the current call sites.

## Suggestion 3 - `ASSERT_GE (buffer_count, 10)` floor left as-is, deferred to #4875 - Deferred, judgment reviewed and endorsed

Confirmed `tests/nnstreamer_datarepo/unittest_datareposrc.cc` has no changes in this round; the floor is untouched. Checked the actual state of `nnstreamer/nnstreamer#4875` (`gh pr diff 4875`, currently **DRAFT / DO NOT MERGE**):

- #4875 changes `gst/join/gstjoin.c` to forward EOS from `join` only after every linked sink pad has ended (via a new per-pad `eos` flag, `gst_join_all_sinkpads_eos()`, and hooks in `gst_join_pad_event()`/`gst_join_release_pad()`/`gst_join_change_state()`), which is exactly the structural fix that would make `create_flexible_tensors_test_file()`'s 3-branch `join` write a deterministic 30 samples instead of a load-dependent prefix.
- #4875's diff **does** touch `tests/nnstreamer_datarepo/unittest_datareposrc.cc`, in the same `fps30ReadFlexibleTensors` test function this PR modified, replacing the timing-margin-based `EXPECT_LT` check with an exact `EXPECT_EQ (buffer_count, total_samples)` - confirming the author's statement that the two PRs touch the same function and whichever merges second will need a rebase.
- Given #4875 is a draft that restructures the same test function around a different (exact-count) assertion strategy, not preemptively touching the floor here avoids doing throwaway work now that #4875 would immediately supersede/conflict with. Merging this PR first (isolation fix + load-tolerant floor) and letting #4875 replace the floor with an exact equality once `join` is deterministic is a sound, low-risk sequencing - it doesn't block this PR's actual goal (fixing the parallel-CI flake) on an unrelated, higher-risk element-behavior change that is still in draft.

## New issues from this round

None found (Critical/Major/Minor). Additional checks performed:
- **CI/doxygen-tag compliance**: re-emulated `.github/workflows/static.check.scripts/doxygen-tag.sh`'s two comment-formatting rules (lines 124 and 131) against the actual git-blob content of all four previously-touched files at `d873e00b` (`tests/unittest_util.c`, `tests/unittest_util.h`, both datarepo `.cc` files) - zero violations. Note the local worktree checks out CRLF (`core.autocrlf=true`), but the committed blobs (via `git show`) are plain LF with no `\r`, matching what the Ubuntu CI runner lints, so this reproduces the real CI check rather than risking a false negative from line-ending differences. `Static checks` is green on the current head per `gh pr checks 4874`.
- **Scope**: diff is limited to the two functions discussed above; no other of the ~32 binaries linking `nnstreamer_unittest_util` reference the changed symbols, consistent with prior rounds.
- One **Nit**: the new `@note` block plus the now-unconditional `cwd = g_get_current_dir()` at the top of `enterPrivateWorkDir()` (previously fetched only inside the `NNSTREAMER_SOURCE_ROOT_PATH`-unset branch) is a correct and necessary change to support suggestion 2, not a regression - just flagging it as an intentional behavior widening (one extra syscall per call, negligible) for the record.

## Verdict

All three non-blocking suggestions from the approving review are addressed appropriately: two implemented and verified correct (header doc, return-to-origin chdir), one soundly deferred to the PR that actually owns the root cause it depends on (#4875, confirmed in draft and touching the same test function). No new issues surfaced. **LGTM - ok to merge.**

PR #4875, opened 2026-08-26
[Join] Forward EOS only after every sink pad ends

myungjoo:
## Problem

`datareposrc.fps30ReadFlexibleTensors` fails intermittently in CI. It is the recurring flake of #4184 and #4287, seen again on 2026-08-26 in run [32974102457](https://github.com/nnstreamer/nnstreamer/actions/runs/32974102457):

```
[ RUN      ] datareposrc.fps30ReadFlexibleTensors
Elapsed time: 0.758597 second
../tests/nnstreamer_datarepo/unittest_datareposrc.cc:668: Failure
Expected: (0.8) < (elapsed_time), actual: 0.8 vs 0.758597
```

The cause is not runner latency. `join` forwards every event arriving on its currently active sink pad, and EOS is an event. When several inputs run at once, whichever branch finishes first while it happens to be active tears down everything downstream, and the remaining branches lose their pending buffers.

`create_flexible_tensors_test_file()` builds its fixture from three concurrent `videotestsrc` branches through one `join`, so the file it writes holds an arbitrary prefix of its 30 samples. The reader then paces fewer samples and finishes early. Every elapsed time reported on those issues matches a specific truncated count:

| reported elapsed | samples written |
|---|---|
| 0.320441 s (#4184) | 11 |
| 0.653894 s (#4287) | 21 |
| 0.758597 s (this run) | 24 |
| 0.955010 s (passing) | 30 |

Measured directly by running the writer pipeline 40 times under CPU load and reading `total_samples` back from the JSON:

| | complete fixtures |
|---|---|
| before | 7 / 40 (range 10–30) |
| after | 40 / 40 |

## Change

`gst/join/gstjoin.c` tracks EOS per sink pad and forwards it only once every **linked** sink pad has ended. Unlinked pads are skipped, since no stream can ever end on them and waiting for one would hang the pipeline. The flags are cleared on `FLUSH_STOP` and on `READY_TO_PAUSED` so a seeked or restarted pipeline starts clean.

It also fixes a failover case that was broken before: one branch ending no longer ends the joined stream while another is still producing. The element documentation used to say the inputs never run at the same time, which is exactly the situation this handles; it now says what happens when they overlap.

[...]

PR #4876, opened 2026-08-26
[Join] Forward EOS only after every sinkpad has ended

myungjoo:
## The problem

`gst_join_pad_event()` computes

```c
forward = (pad == active_sinkpad);
```

once, at the top, and applies it to **every** event — EOS included. `gst_join_pad_chain()` makes whichever pad pushed the most recent buffer the active one. So the fate of a branch's EOS depends on a race:

- if the branch is the active pad when its EOS arrives, EOS goes downstream **immediately** and the sink stops accepting data — the buffers the other branches are still pushing are dropped;
- otherwise the EOS is silently unref'd and nothing ever ends the output stream from that branch.

`join` is documented as an N-to-1 element, so neither is right: the output stream ends when *all* the input streams have ended.

## Measured impact

The fixture `create_flexible_tensors_test_file()` in `tests/nnstreamer_datarepo/unittest_datareposrc.cc` joins three `videotestsrc` branches of 10 buffers each into `datareposink`. Probing it directly (10 runs each):

| condition | `total_samples` written |
|---|---|
| idle | 30 ×10 |
| all cores loaded, before this PR | 26, 28, 27, 27, 30, 30, 28, 17, 28, 20 |
| all cores loaded, with this PR | 30 ×10 |

That non-determinism is the second cause analysed in #4874, which deliberately left the element alone and only made the tests independent of the sample count. This PR is the element-side fix; the two do not conflict.

## The change

`gst/join/gstjoin.c`:

- each `GstJoinPad` carries an `eos` flag, the element carries `eos_count`/`eos_sent`, and the `GST_EVENT_EOS` case forwards EOS only when `eos_count == n_pads`. Buffers and all other events keep their existing pass-through behaviour.
- `GST_EVENT_FLUSH_STOP` withdraws the pad from the collected EOS state and clears `eos_sent`, and `READY_TO_PAUSED` resets it for every pad, so a flushing seek or a pipeline restart can end the stream again.
- `release_pad` is implemented (it was missing entirely). A released pad can no longer deliver EOS, so it is dropped from the count, and if the pads left behind have all ended, the release itself forwards EOS. It also decrements `n-pads` and clears `active-pad`, which previously leaked across a pad release.

The element documentation now states the aggregation rule, including its consequence: a requested sinkpad that is never linked keeps the output stream alive.

## Tests

Three new cases in `tests/gstreamer_join/unittest_join.cc`:

| test | what it pins down |
|---|---|
| `join.forwardAllBuffers` | 3 sources × 10 buffers ⇒ 30 buffers at the sink |
| `join.eosAggregation` | EOS on the active pad does not cut off the other stream, and EOS reaches the bus exactly once both pads are done |
| `join.releasePadAfterEos` | releasing the last awaited pad completes the aggregation |

Against the unpatched element all three fail, and `forwardAllBuffers` fails **10/10 even on an idle machine** (10–28 buffers instead of 30). `tensor_sink` is given `qos=false` there so that a late-buffer QoS event cannot make upstream drop frames — that is real GStreamer behaviour and would otherwise be counted as join's loss.

## Verification

| scenario | result |
|---|---|
| `unittest_join`, 40 runs, all 24 cores saturated | 40/40 pass |
| `unittest_join` + `unittest_datarepos{rc,ink}` + `unittest_filter_custom`, `--repeat 6` under load | 24/24 pass |
| full local `meson test` suite | 19/19 pass, no regressions |
| SSAT `tests/gstreamer_join/runTest.sh` pipelines run by hand | 1 and 2 reach EOS and match the goldens; 3_n still fails as expected |

The SSAT cases now produce one extra `testJoin*_4.log` (the branch that used to be cut off finishes); the script only compares indices 0–3 and removes `*.log` afterwards, so it is unaffected.

### DIFF
--- a/gst/join/gstjoin.c
+++ b/gst/join/gstjoin.c
@@ -19,6 +19,9 @@
  * N streams should not operate at the same time.
  * All capabilities (input stream i and output stream) should be the same.
  * For example, If one sinkpad is receiving buffer, the others should be stopped.
+ * As an N-to-1 element, join ends its output stream only after every sinkpad
+ * has received EOS, so a sinkpad that is never linked keeps the output stream
+ * alive.
  * <refsect2>
  * <title>Example launch line</title>
  * gst-launch-1.0 ... (input stream 0) ! join.sink_0 \
@@ -98,6 +101,8 @@ struct _GstJoinPad
 
   GstSegment segment;           /* the current segment on the pad */
   guint32 segment_seqnum;       /* sequence number of the current segment */
+
+  gboolean eos;                 /* TRUE if the pad has received EOS */
 };
 
 /**
@@ -162,7 +167,40 @@ gst_join_pad_reset (GstJoinPad * pad)
 {
   GST_OBJECT_LOCK (pad);
   gst_segment_init (&pad->segment, GST_FORMAT_UNDEFINED);
+  pad->eos = FALSE;
   GST_OBJECT_UNLOCK (pad);
+}
+
+/**
+ * @brief Withdraw the EOS of the given sink pad from the collected EOS state.
+ * @note must be called with the JOIN_LOCK
+ */
+static void
+gst_join_pad_clear_eos (GstJoin * sel, GstJoinPad * pad)
+{
+  if (pad->eos) {
+    pad->eos = FALSE;
+    if (sel->eos_count > 0)
+      sel->eos_count--;
+  }
+}
+
+/**
+ * @brief Reset the EOS state of every sink pad.
+ * @note must be called with the JOIN_LOCK
+ */
+static void
+gst_join_reset (GstJoin * sel)
+{
+  GList *walk;
+
+  GST_OBJECT_LOCK (sel);
+  for (walk = GST_ELEMENT_CAST (sel)->sinkpads; walk; walk = walk->next)
+    gst_join_pad_reset (GST_JOIN_PAD_CAST (walk->data));
+  GST_OBJECT_UNLOCK (sel);
+
+  sel->eos_count = 0;
+  sel->eos_sent = FALSE;
 }
 
 /**
@@ -284,6 +322,33 @@ gst_join_pad_event (GstPad * pad, GstObject * parent, GstEvent * event)
           &selpad->segment);
       break;
     }
+    case GST_EVENT_EOS:
+    {
+      /**
+       * Join is an N-to-1 element, so the output stream ends only after every
+       * input stream has ended. Forwarding EOS as soon as the pad that happens
+       * to be active receives it makes the downstream elements discard the
+       * buffers that the other, still running, input streams are about to push.
+       */
+      if (!selpad->eos) {
+        selpad->eos = TRUE;
+        sel->eos_count++;
+      }
+
+      forward = (!sel->eos_sent && sel->eos_count >= sel->n_pads);
+      if (forward)
+        sel->eos_sent = TRUE;
+
+      GST_DEBUG_OBJECT (pad, "EOS received (%u of %u sink pads), %s",
+          sel->eos_count, sel->n_pads,
+          forward ? "forwarding" : "waiting for the other sink pads");
+      break;
+    }
+    case GST_EVENT_FLUSH_STOP:
+      /* The stream restarts after a flush, so drop the collected EOS state. */
+      gst_join_pad_clear_eos (sel, selpad);
+      sel->eos_sent = FALSE;
+      break;
     default:
       break;
   }
@@ -424,6 +489,9 @@ static void gst_join_get_property (GObject * object,
     guint prop_id, GValue * value, GParamSpec * pspec);
 static GstPad *gst_join_request_new_pad (GstElement * element,
     GstPadTemplate * templ, const gchar * unused, const GstCaps * caps);
+static void gst_join_release_pad (GstElement * element, GstPad * pad);
+static GstStateChangeReturn gst_join_change_state (GstElement * element,
+    GstStateChange transition);
 
 #define gst_join_parent_class parent_class
 G_DEFINE_TYPE_WITH_CODE (GstJoin, gst_join, GST_TYPE_ELEMENT,
@@ -465,6 +533,8 @@ gst_join_class_init (GstJoinClass * klass)
       &gst_join_src_factory);
 
   gstelement_class->request_new_pad = gst_join_request_new_pad;
+  gstelement_class->release_pad = gst_join_release_pad;
+  gstelement_class->change_state = gst_join_change_state;
 }
 
 /**
@@ -482,6 +552,8 @@ gst_join_init (GstJoin * sel)
   sel->active_sinkpad = NULL;
   sel->padcount = 0;
   sel->have_group_id = TRUE;
+  sel->eos_count = 0;
+  sel->eos_sent = FALSE;
 
   g_mutex_init (&sel->lock);
   g_cond_init (&sel->cond);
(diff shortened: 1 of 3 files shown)


myungjoo:

## Verdict

**Approve, with a few Minor/Nit follow-ups.** No Critical or Major issue found. The change does what the title says, the fix is the right one for an N-to-1 element, the blast radius is confined to `gst/join`, and the new tests genuinely fail without the element change.

## What I verified locally (Ubuntu/WSL, meson+ninja)

| check | result |
|---|---|
| `unittest_join` with the PR applied | 6/6 pass |
| `unittest_join` with **only `gst/join` reverted** (tests kept) | 3/6 fail - `forwardAllBuffers` (`received` = 20, expected 30), `eosAggregation`, `releasePadAfterEos` (`n-pads` = 2, expected 1; EOS arrives at the 200 ms negative wait) |
| `unittest_join` x25 with every core saturated | 25/25 pass, no GLib/GStreamer `WARNING`/`CRITICAL` in the logs |
| full `meson test` | 19/19 OK - no regression in `unittest_datareposrc/sink`, `unittest_if`, `unittest_sink`, `unittest_plugins` |
| SSAT `gstreamer_join` case 1 pipeline run by hand, patched vs. unpatched | both emit `testJoin1_0..4.log`, md5 of indices 0-3 identical, so the goldens still match |
| pipeline restart (`NULL -> PLAYING -> NULL -> PLAYING`) with a standalone probe | EOS forwarded in **both** rounds, i.e. the `READY_TO_PAUSED` reset works |

So the regression tests are real regression tests, they are wired into `meson test` (`tests/meson.build:174`) and therefore block CI on failure, and every failure mode is bounded by `TEST_TIMEOUT_LIMIT_MS` (10 s) rather than hanging the runner.

Other things I checked and found clean:

- **Lock ordering.** The new `gst_join_reset()` takes `GST_JOIN_LOCK` -> `GST_OBJECT_LOCK(element)`, which is the same order the pre-existing `gst_join_request_new_pad()` (`gst_element_add_pad`) and `gst_join_get_active_sinkpad()` (`gst_element_iterate_sink_pads`) already use. `gst_join_release_pad()` correctly drops `GST_JOIN_LOCK` before `gst_element_remove_pad()` and before pushing EOS. I could not construct a new deadlock.
- **No EOS-before-buffer reordering.** EOS is forwarded only once every pad is EOS, and per-pad ordering is guaranteed by that pad's own streaming thread, so no branch can still be pushing buffers when EOS goes out.
- **`eos_count` bookkeeping.** Add-pad-after-partial-EOS, release-of-an-EOS-pad, release-of-a-non-EOS-pad and a duplicate EOS on one pad all end up with the right count; `eos_sent` prevents a double EOS between the event path and the release path.
- **The `release_pad` part is not scope creep.** `GstElement` falls back to `gst_element_remove_pad()` when `release_pad` is `NULL`, so before this PR the pad really was removed while `sel->active_sinkpad` kept a ref on it and `n-pads` stayed stale - a genuine latent bug that the EOS state machine forced into the open. Fixing it here is right.
- **Semantics match upstream precedent.** This is what `funnel` does (`gst_funnel_all_sinkpads_eos_unlocked()`), including the "a requested-but-never-linked sinkpad keeps the output stream alive" consequence, which the PR documents.
- **Docs.** No public API or architecture change; `Documentation/component-description.md` only links the directory, and the element's own gtk-doc `SECTION:` block is updated in the PR. That is adequate, except for item 1 below.

---

## Findings

### 1. [Minor] The element doc now contradicts itself

`gst/join/gstjoin.c:19` still says:

> N streams should not operate at the same time. ... For example, If one sinkpad is receiving buffer, the others should be stopped.

while the whole premise of this PR ("the buffers the other branches are still pushing are dropped") and the new `join.forwardAllBuffers` test (3 `videotestsrc` branches running concurrently) depend on N branches being live simultaneously - as does the `datarepo` fixture cited in the description. As it stands a future maintainer cannot tell which sentence is authoritative and could "fix" the new test on the grounds that it violates the documented contract. Please reconcile the two paragraphs in this PR, since it is the paragraph immediately above the one being added.

### 2. [Minor] The `READY_TO_PAUSED` reset has no test - the highest-value gap here

`gst_join_change_state()` (gstjoin.c:800) is load-bearing: drop `sel->eos_sent = FALSE` from `gst_join_reset()` and any pipeline that goes `PLAYING -> NULL -> PLAYING` (or `READY -> PAUSED` twice) never emits EOS again - it hangs. I confirmed by hand that the current code handles two rounds correctly, but nothing in `unittest_join` exercises a restart, so a future refactor that loses the reset would go green in CI and produce a hang in the field. A `join.eosAfterRestart` case (run the `eosAggregation` body twice around a `GST_STATE_NULL` round trip) is roughly ten lines and would close it.

### 3. [Minor] The `FLUSH_STOP` path has no test either

`gstjoin.c:347-350` is the other new branch with zero coverage. A flushing seek after EOS is exactly the scenario it exists for, and `gst_join_pad_clear_eos()`'s `eos_count--` is easy to break silently (e.g. by dropping the `pad->eos` guard, which would underflow the count on a repeated `FLUSH_STOP`). Same argument as item 2.

### 4. [Minor] `wait_pipeline_eos()` cannot distinguish "no EOS" from "the pipeline errored out"

`tests/gstreamer_join/unittest_join.cc:168` pops `EOS | ERROR` and returns `FALSE` for both. That is fine for the `EXPECT_TRUE` sites, but the two negative assertions

```c
EXPECT_FALSE (wait_pipeline_eos (pipeline, 200));   /* :231 and :337 */
```

will happily pass on a pipeline that just posted a fatal `ERROR` - and they also *consume* that message off the bus, so the failure is partly hidden from the later assertions as well. Suggest returning a small enum (or adding an out-param) and asserting `EXPECT_FALSE(got_error)` at both negative sites, so an error is reported as an error rather than as "EOS did not arrive, as expected".

### 5. [Minor] `eos_sent` is latched before the push, so a failed EOS push is unrecoverable

```c
forward = (!sel->eos_sent && sel->eos_count >= sel->n_pads);
if (forward)
  sel->eos_sent = TRUE;          /* gstjoin.c:338-340 */
```

If the subsequent `gst_pad_push_event()` returns `FALSE` (downstream flushing, srcpad not linked, ...), EOS never reached downstream but `eos_sent` says it did, and nothing retries - the stream can then no longer be ended without a flush or a state change. Low probability, but since the point of the PR is "the output stream reliably ends", it is worth either latching only on success or adding a one-line comment saying the loss is deliberate. The same applies to `gstjoin.c:782-783` in `gst_join_release_pad()`.

### 6. [Nit] `eos_count` duplicates state the pad list already carries

The counter must be kept in sync in four places (`GST_EVENT_EOS`, `gst_join_pad_clear_eos`, `gst_join_release_pad`, `gst_join_reset`), and the `if (sel->eos_count > 0)` / `if (sel->n_pads > 0)` underflow guards hint that the invariant is hard to see locally. `funnel` instead derives the predicate by walking `GST_ELEMENT_CAST (sel)->sinkpads` and returning `FALSE` at the first pad without EOS - drift-proof by construction, and `n_pads` is small. I did not find an actual drift case in the current code, so this is a readability/robustness suggestion only.

### 7. [Nit] `pad->eos` is protected by two different locks depending on the call site

`gst_join_pad_reset()` writes `pad->eos` under `GST_OBJECT_LOCK (pad)` (gstjoin.c:170), while `gst_join_pad_event()` and `gst_join_pad_clear_eos()` write it under `GST_JOIN_LOCK` only. In practice every path except `gst_join_pad_init()` holds `GST_JOIN_LOCK`, so it is correct today, but the mixed convention invites a future writer to add an unlocked accessor. Suggest documenting `pad->eos` as JOIN_LOCK-protected in the struct comment and leaving the pad object lock to `segment` only.

### 8. [Nit] `releasePadAfterEos` checks `n-pads` but not `active-pad`

The test asserts `n_pads == 1` after the release, which is good, but the dangling `active_sinkpad` was the more dangerous half of the pre-existing bug (a ref held on a pad that `gst_element_remove_pad()` had already dropped). Releasing the **active** pad and asserting `active-pad == NULL` would pin that half down; as written, the released pad (`sink_1`) is never the active one, so that branch of `gst_join_release_pad()` (gstjoin.c:771-774) is not executed by any test.

### 9. [Nit] Description detail, no code change needed

The PR body says the SSAT cases "now produce one extra `testJoin*_4.log`". I could not reproduce that: the unpatched element also produced `testJoin1_0..4.log` here, with identical md5 sums for indices 0-4. Harmless either way (the goldens for 0-3 match in both cases), but worth not relying on that sentence as evidence.

---

## Summary

| # | severity | item |
|---|---|---|
| 1 | Minor | element doc still says the N streams must not run concurrently, contradicting the PR and `forwardAllBuffers` |
| 2 | Minor | no test for the `READY_TO_PAUSED` EOS reset (a regression there means a pipeline hang) |
| 3 | Minor | no test for the `FLUSH_STOP` clear-EOS path |
| 4 | Minor | `wait_pipeline_eos()` conflates timeout with a bus ERROR, weakening the two `EXPECT_FALSE` assertions |
| 5 | Minor | `eos_sent` latched before the push; a failed EOS push is never retried |
| 6 | Nit | `eos_count` counter vs. iterating the sinkpad list (cf. `funnel`) |
| 7 | Nit | `pad->eos` guarded by `GST_OBJECT_LOCK(pad)` in one place and `GST_JOIN_LOCK` elsewhere |
| 8 | Nit | `releasePadAfterEos` does not cover releasing the *active* pad / `active-pad == NULL` |
| 9 | Nit | the SSAT "extra `_4.log`" claim in the description does not reproduce |

None of these block the merge in my view. Items 1-3 are the ones I would fold into this PR rather than defer: 1 is a two-line doc edit in the block already being touched, and 2-3 close the coverage of the only new code paths that no test executes.

(CI for this head commit was still queued when this review was written; the Ubuntu meson/valgrind job had not reported yet. The arm64 job shows "fail" only because it was *cancelled*, not because of a test failure.)

myungjoo:
Closing as a duplicate of #4875, which addresses the same defect and was opened first.

The two arrived at the same root cause and the same fix independently. #4875's handling is better in one respect — it skips **unlinked** sink pads when deciding whether every stream has ended, so a requested-but-never-linked pad cannot hang the pipeline, whereas this PR waited for it.

The two things that were only in this PR have been carried over to #4875 in bde85434:

- `release_pad`, which join never implemented. Releasing a request pad left it counted in `n-pads` and left `active-pad` referencing a removed pad, and with EOS aggregation in place it became a way to hang a pipeline. Release now also re-evaluates the aggregation, so a release that leaves only ended streams behind forwards EOS.
- `join.forwardAllBuffers`, the datarepo fixture shape reduced to the element: three `videotestsrc` branches of 10 buffers, all 30 required at the sink. It fails 10/10 on an idle machine without the fix.
- Plus `join.eosAfterPadRelease` and `join.eosAfterFlush` for the two paths that had no coverage.

The review left on this PR (#4876 (comment)) was applied to the merged result:

| finding | disposition |
|---|---|
| Minor 1 — element doc contradicts concurrent inputs | fixed in bde85434 |
| Minor 2 — `READY_TO_PAUSED` reset untested | already covered by `join.eosAfterRestart` in #4875 |
| Minor 3 — `FLUSH_STOP` path untested | `join.eosAfterFlush` added |
| Minor 4 — helper conflates timeout with bus ERROR | #4875's `pop_eos_or_error()` returns the message type, so timeout and ERROR are distinct |
| Minor 5 — `eos_sent` latched before the push | not applicable; the merged version has no such flag and uses the source pad's sticky EOS |
| Nit 6 — counters drift | not applicable; the merged version walks the sinkpads instead of counting |
| Nit 8 — releasing the *active* pad uncovered | `join.eosAfterPadRelease` releases the active pad |
| Nit 9 — "extra `testJoin*_4.log`" claim unreproducible | correct, the claim was wrong and is not repeated in #4875 |

PR #4877, opened 2026-08-26
[CI] Fail the pdebuild setup step when apt-get update fails

myungjoo:
## Problem

`.github/workflows/pdebuild.yml` → step **prepare pdebuild** ran setup as one `&&` list:

```bash
sudo apt-get update && sudo apt-get install -y pbuilder debootstrap curl ...
```

The step runs under GitHub's default `shell: /usr/bin/bash -e`. Under errexit, a command that is part of an `&&` list and is **not** the last command in that list does not cause the shell to exit. So when `apt-get update` failed it simply short-circuited the install, errexit never fired, and the step's exit status came from the trailing `sudo ln -s ~/.pbuilderrc /root/` — the step reported **SUCCESS**.

The job then failed ~20 seconds later, several steps downstream, in **make pbuilder base.tgz**:

```
sudo: pbuilder: command not found
```

which points the reader at pbuilder rather than at the setup that never ran.

Observed on #4238, run [32977958537](https://github.com/nnstreamer/nnstreamer/actions/runs/32977958537/job/98207232010) (2026-08-26).

The trigger there was external and transient: `packages.microsoft.com` returned `403 Forbidden` for both the `azure-cli` and `ubuntu/22.04/prod` repos, so `apt-get update` exited non-zero — even though every repo nnstreamer actually needs (azure.archive.ubuntu.com, the nnstreamer PPA) fetched fine. Those Microsoft sources are part of the GitHub-hosted runner image; nothing in this build uses them.

## Change

1. **Honest, immediate failure** — `apt-get update` and `apt-get install` become separate statements, so a setup failure aborts the step where it happens instead of masquerading as a build failure later.

2. **Resilience to unrelated third-party repos** — drop any `sources.list.d` entry pointing at `packages.microsoft.com` before touching apt:

   ```bash
   { sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
   ```

   Matching on the URI rather than the filename covers both the classic `.list` files and deb822 `.sources` files, and survives runner-image renames. The removal happens before `add-apt-repository` (which itself runs an apt update on jammy).

   The `|| true` group is deliberate: `grep` exits 1 on the normal "nothing to remove" path, and this step must not care. Neutralizing it inside the group — rather than relying on the pipeline's last command — keeps the behavior identical whether or not `pipefail` is set, so a later `shell: bash` on this step cannot silently turn the no-op case into a step failure. `-Z`/`-0` delimits filenames unambiguously.

`.github/workflows/update_pbuilder_cache.yml` carried the same step verbatim, so it gets the same fix.

## Scope

The same `apt-get update && apt-get install` shape appears in other workflows, but it only masks a failure when the list is **not** the last command of the step script. `publish_docs.yml:48`, `static.check.yml:24` and `update_gbs_cache.yml:40` are single-line `run:` steps, so their status already propagates correctly. The remaining sites (`codeql.yml`, `gbs_build.yml`, `nns_testhub.yml`, `publish_docs.yml:20`, `ubuntu_clean_meson_build.yml`) are left alone here: several of the packages they install are preinstalled on the runner image, so splitting them would convert currently-passing degraded runs into hard failures — a behavior change that needs per-file judgment rather than a mechanical sweep, and one unrelated to the #4238 failure.

## Verification

- Both files parse as YAML; `actionlint` reports nothing on the changed step (its two findings are pre-existing missing-`description` metadata in `check-rebuild/action.yml` and `s3_upload/action.yml`, untouched here).
- errexit masking reproduced directly: `bash -e -c 'false && true; echo reached'` prints `reached`.
- The cleanup pipeline was exercised against a synthetic `sources.list.d` holding an `azure-cli.list`, a deb822 `microsoft-prod.sources`, and an unrelated `keep.list` — under **both** `bash -e` (the runner default) and `bash -eo pipefail`, for no-match, matches-present, and missing-directory. Exit 0 in all six combinations; both Microsoft entries removed; `keep.list` untouched.
- `check_if_rebuild_requires.sh` routes a `.github/workflows/*` change through the catch-all arm (`redebian=1`), so the pdebuild job on this PR exercises the changed step end to end.

### DIFF
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -66,7 +66,7 @@ jobs:
         # grep exits 1 when there is nothing to remove, which is not a failure here.
         { sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
         echo "::endgroup::"
-        sudo add-apt-repository ppa:nnstreamer/ppa
+        sudo add-apt-repository -y -n ppa:nnstreamer/ppa
         # Keep update and install separate: errexit ignores a failure before "&&".
         echo "::group::apt-get update"
         # "&& break" leans on that same rule: a failed attempt must not abort the step.
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -69,7 +69,15 @@ jobs:
         sudo add-apt-repository ppa:nnstreamer/ppa
         # Keep update and install separate: errexit ignores a failure before "&&".
         echo "::group::apt-get update"
-        sudo apt-get update
+        # "&& break" leans on that same rule: a failed attempt must not abort the step.
+        for i in 1 2 3; do
+          sudo apt-get update && break
+          if [ "$i" = 3 ]; then
+            echo "::error::apt-get update failed after $i attempts"
+            exit 1
+          fi
+          sleep $((i * 10))
+        done
         echo "::endgroup::"
         echo "::group::apt-get install"
         sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
@@ -78,7 +86,7 @@ jobs:
         echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main |deb [trusted=yes] http://ppa.launchpad.net/kubuntu-ppa/backports-extra/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
         cat ~/.pbuilderrc
         sudo mkdir -p /root/
-        sudo ln -s ~/.pbuilderrc /root/
+        sudo ln -sf ~/.pbuilderrc /root/
 
     - name: make pbuilder base.tgz
       if: ${{ env.rebuild == '1' && steps.restore-pbuilder-cache.outputs.cache-hit != 'true' }}
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -60,9 +60,22 @@ jobs:
       if: env.rebuild == '1'
       run: |
         echo "Installing build tools"
+        # The GitHub-hosted runner image ships apt sources for packages.microsoft.com
+        # (azure-cli and ubuntu/<ver>/prod), which nnstreamer does not use. That host
+        # intermittently answers 403 to runners and then "apt-get update" fails for a
+        # reason unrelated to this build; drop those sources before touching apt.
+        echo "::group::drop unused third-party apt sources"
+        sudo grep -rl 'packages\.microsoft\.com' /etc/apt/sources.list.d/ | sudo xargs -r rm -fv
+        echo "::endgroup::"
         sudo add-apt-repository ppa:nnstreamer/ppa
-        echo "::group::apt-get update && apt-get install"
-        sudo apt-get update && sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
+        # Keep update and install as separate statements. In an "A && B" list, a
+        # failing A does not trigger errexit, so a broken apt-get update used to leave
+        # the step green and surface as "sudo: pbuilder: command not found" later on.
+        echo "::group::apt-get update"
+        sudo apt-get update
+        echo "::endgroup::"
+        echo "::group::apt-get install"
+        sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
         echo "::endgroup::"
         echo "DISTRIBUTION=${{ matrix.distroname }}" > ~/.pbuilderrc
         echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main |deb [trusted=yes] http://ppa.launchpad.net/kubuntu-ppa/backports-extra/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -60,17 +60,14 @@ jobs:
       if: env.rebuild == '1'
       run: |
         echo "Installing build tools"
-        # The GitHub-hosted runner image ships apt sources for packages.microsoft.com
-        # (azure-cli and ubuntu/<ver>/prod), which nnstreamer does not use. That host
-        # intermittently answers 403 to runners and then "apt-get update" fails for a
-        # reason unrelated to this build; drop those sources before touching apt.
+        # Drop the runner image's packages.microsoft.com sources: this build does not
+        # use them and that host intermittently answers 403, failing apt-get update.
         echo "::group::drop unused third-party apt sources"
-        sudo grep -rl 'packages\.microsoft\.com' /etc/apt/sources.list.d/ | sudo xargs -r rm -fv
+        # grep exits 1 when there is nothing to remove, which is not a failure here.
+        { sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
         echo "::endgroup::"
         sudo add-apt-repository ppa:nnstreamer/ppa
-        # Keep update and install as separate statements. In an "A && B" list, a
-        # failing A does not trigger errexit, so a broken apt-get update used to leave
-        # the step green and surface as "sudo: pbuilder: command not found" later on.
+        # Keep update and install separate: errexit ignores a failure before "&&".
         echo "::group::apt-get update"
         sudo apt-get update
         echo "::endgroup::"
--- a/.github/workflows/update_pbuilder_cache.yml
+++ b/.github/workflows/update_pbuilder_cache.yml
@@ -37,7 +37,7 @@ jobs:
           # grep exits 1 when there is nothing to remove, which is not a failure here.
           { sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
           echo "::endgroup::"
-          sudo add-apt-repository ppa:nnstreamer/ppa
+          sudo add-apt-repository -y -n ppa:nnstreamer/ppa
           # Keep update and install separate: errexit ignores a failure before "&&".
           echo "::group::apt-get update"
           # "&& break" leans on that same rule: a failed attempt must not abort the step.
--- a/.github/workflows/update_pbuilder_cache.yml
+++ b/.github/workflows/update_pbuilder_cache.yml
@@ -31,17 +31,14 @@ jobs:
       - name: prepare pdebuild
         run: |
           echo "Installing build tools"
-          # The GitHub-hosted runner image ships apt sources for packages.microsoft.com
-          # (azure-cli and ubuntu/<ver>/prod), which nnstreamer does not use. That host
-          # intermittently answers 403 to runners and then "apt-get update" fails for a
-          # reason unrelated to this build; drop those sources before touching apt.
+          # Drop the runner image's packages.microsoft.com sources: this build does not
+          # use them and that host intermittently answers 403, failing apt-get update.
           echo "::group::drop unused third-party apt sources"
-          sudo grep -rl 'packages\.microsoft\.com' /etc/apt/sources.list.d/ | sudo xargs -r rm -fv
+          # grep exits 1 when there is nothing to remove, which is not a failure here.
+          { sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
           echo "::endgroup::"
           sudo add-apt-repository ppa:nnstreamer/ppa
-          # Keep update and install as separate statements. In an "A && B" list, a
-          # failing A does not trigger errexit, so a broken apt-get update used to leave
-          # the step green and surface as "sudo: pbuilder: command not found" later on.
+          # Keep update and install separate: errexit ignores a failure before "&&".
           echo "::group::apt-get update"
           sudo apt-get update
           echo "::endgroup::"
--- a/.github/workflows/update_pbuilder_cache.yml
+++ b/.github/workflows/update_pbuilder_cache.yml
@@ -31,9 +31,22 @@ jobs:
       - name: prepare pdebuild
         run: |
           echo "Installing build tools"
+          # The GitHub-hosted runner image ships apt sources for packages.microsoft.com
+          # (azure-cli and ubuntu/<ver>/prod), which nnstreamer does not use. That host
+          # intermittently answers 403 to runners and then "apt-get update" fails for a
+          # reason unrelated to this build; drop those sources before touching apt.
+          echo "::group::drop unused third-party apt sources"
+          sudo grep -rl 'packages\.microsoft\.com' /etc/apt/sources.list.d/ | sudo xargs -r rm -fv
+          echo "::endgroup::"
           sudo add-apt-repository ppa:nnstreamer/ppa
-          echo "::group::apt-get update && apt-get install"
-          sudo apt-get update && sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
+          # Keep update and install as separate statements. In an "A && B" list, a
+          # failing A does not trigger errexit, so a broken apt-get update used to leave
+          # the step green and surface as "sudo: pbuilder: command not found" later on.
+          echo "::group::apt-get update"
+          sudo apt-get update
+          echo "::endgroup::"
+          echo "::group::apt-get install"
+          sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
           echo "::endgroup::"
           echo "DISTRIBUTION=${{ matrix.distroname }}" > ~/.pbuilderrc
           echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
--- a/.github/workflows/update_pbuilder_cache.yml
+++ b/.github/workflows/update_pbuilder_cache.yml
@@ -40,7 +40,15 @@ jobs:
           sudo add-apt-repository ppa:nnstreamer/ppa
           # Keep update and install separate: errexit ignores a failure before "&&".
           echo "::group::apt-get update"
-          sudo apt-get update
+          # "&& break" leans on that same rule: a failed attempt must not abort the step.
+          for i in 1 2 3; do
+            sudo apt-get update && break
+            if [ "$i" = 3 ]; then
+              echo "::error::apt-get update failed after $i attempts"
+              exit 1
+            fi
+            sleep $((i * 10))
+          done
           echo "::endgroup::"
           echo "::group::apt-get install"
           sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
@@ -49,7 +57,7 @@ jobs:
           echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
           cat ~/.pbuilderrc
           sudo mkdir -p /root/
-          sudo ln -s ~/.pbuilderrc /root/
+          sudo ln -sf ~/.pbuilderrc /root/
 
       - name: get date
         id: get-date


myungjoo:

Scope reviewed: `.github/workflows/pdebuild.yml` and `.github/workflows/update_pbuilder_cache.yml` only (the two files this PR touches). I verified the diff's claims empirically rather than just reading them — reproduced the errexit behavior locally in bash, and pulled the actual raw logs of a recent successful run of this repo's `pdebuild.yml` (`gh run view 32981441328 --log`) to check the real runner's shell invocation and the real Microsoft apt sources present.

**Overall: no blockers or majors. The fix is correct and the diff is well-scoped.** A few non-blocking observations below.

---

**1. Core errexit-split fix — verified correct (no issue)**

Confirmed the exact swallowing behavior described in the PR body: under `set -e` (no pipefail), `false && echo reached` does *not* abort — the shell continues and `reached` prints, exit 0. Splitting into two separate statements (`false` on its own line) does abort as expected. So the `apt-get update` / `apt-get install` split genuinely fixes the "step reports SUCCESS while pbuilder never got installed" bug.

I also confirmed, directly from this repo's own CI logs, that the real GitHub-hosted runner invokes steps with no `shell:` key as literally `shell: /usr/bin/bash -e {0}` — **no `-o pipefail`** (e.g. run 32981441328, "prepare pdebuild" step). This matters for finding 3 below.

**2. `grep -rl ... | xargs -r rm -fv` pipeline safety — verified correct under the actual runner shell (no issue)**

Tested all three cases the PR body claims to have verified, using plain `bash -e` (matching the real runner, see above):
- no match in `sources.list.d`: exit 0, script continues.
- match present: file removed, exit 0.
- directory missing: `grep` writes "No such file or directory" to stderr and exits 2, but since there's no `pipefail`, the pipeline's exit status is `xargs`'s (0, thanks to `-r`), so the script continues.

All three match the PR's claim. `sudo` on the `grep` side is also the right call, not just cosmetic consistency — a root-only-readable source file would otherwise be silently skipped by an unprivileged `grep` and never get matched/removed.

**3. Minor / forward-looking — the pipeline is only safe because pipefail is not set today (Minor)**

This is the one real caveat. I re-ran the same three cases with `set -eo pipefail`: the "no match" case (the common/expected case — see point 5) then aborts the step with exit 1, and the "directory missing" case aborts with exit 2, both before reaching `add-apt-repository`/`apt-get update` at all. GitHub's own docs claim `-eo pipefail` is the default even without a `shell:` key, but that's a long-standing documented-vs-actual discrepancy (actions/runner#353, closed as "working as intended" by GitHub staff in 2020, and still true today per the log evidence in point 1). So the fix is correct today, but it is quietly relying on that platform default staying "no pipefail." If someone later "cleans up" this workflow by adding an explicit `shell: bash` (a very plausible well-intentioned edit, since `shell: bash` is often recommended for portability/clarity), pipefail turns on and this exact step starts failing on the ordinary case where no Microsoft source is present — the opposite of what this PR is trying to achieve (honest failures instead of confusing ones). Worth a one-line comment in the script noting this dependency, or restructuring to not rely on it, e.g. capture the grep output into a variable with `|| true` and only pass it to `rm` if non-empty, instead of piping `grep` straight into `xargs`. Not a blocker — just a latent fragility worth documenting since it's non-obvious and easy to reintroduce by a future "shell hygiene" edit.

**4. Ordering relative to `add-apt-repository` — verified correct (no issue)**

`add-apt-repository` on Jammy runs its own implicit `apt-get update` unless `-n` is passed. The Microsoft-source removal is placed before it in both files, so that implicit update is also protected, not just the explicit one later in the script. Correct ordering, and the PR body's reasoning matches.

**5. Content-based match vs. filename-based — verified against the real runner image (no issue)**

Pulled the actual apt log from a real run of this workflow: the runner currently fetches from both `https://packages.microsoft.com/repos/azure-cli` and `https://packages.microsoft.com/ubuntu/22.04/prod`. Cross-checked against `actions/runner-images` build scripts (`install-azure-cli.sh`, `install-ms-repos.sh`): the azure-cli script explicitly removes `/etc/apt/sources.list.d/azure-cli.list` and `azure-cli.list.save` after installing via Microsoft's own installer, which is a strong signal the actual filename/format for that source has moved around over time (classic `.list` vs. deb822 `.sources`). Matching by content (`packages\.microsoft\.com`) rather than by a specific filename is therefore the right call and more resilient to future runner-image changes than a filename-based `rm` would have been. No risk of over-matching: nnstreamer's own sources (`archive.ubuntu.com`, `ppa.launchpad.net`) don't contain that string.

**6. Scope — other workflows have the identical `&&`-masking pattern, untouched here (Minor, not a request to expand this PR)**

The same `apt-get update && apt-get install` pattern also appears in `codeql.yml`, `gbs_build.yml`, `nns_testhub.yml`, `publish_docs.yml` (twice), `static.check.yml`, `ubuntu_clean_meson_build.yml` (twice), and `update_gbs_cache.yml` — all carry the same "a failing `apt-get update` gets masked by the following `&&`" risk this PR fixes in `pdebuild.yml`/`update_pbuilder_cache.yml`. I don't think this PR should be expanded to cover them — keeping the diff scoped to the two files that actually produced the observed incident (#4238) is the right call and consistent with "don't touch unrelated modules." But since the underlying bug class is identical, a follow-up issue/PR applying the same two-line fix to the rest would be worthwhile so the same "green step / confusing downstream failure" doesn't recur elsewhere.

**7. Test coverage for this exact fix (Nit)**

This is workflow YAML with an inline bash snippet, so there's no natural unit-test hook, and I don't think one is warranted here. The verification the PR body describes (synthetic `sources.list.d`, three cases) is exactly right, but it's manual/described-in-prose rather than something CI re-runs on every future edit to this snippet. If this step gets edited again later, nothing will catch a reintroduction of the `&&`-masking bug or the pipefail fragility in point 3 except noticing it by eye. Given the size of the change, I don't consider this blocking — just noting it since the review checklist asks specifically about regression-detection for future edits.

**8. YAML/lint validity (no issue)**

Both files parse as valid YAML (`yaml.safe_load`) and `actionlint` reports no new findings on either file — the two pre-existing actionlint notices (missing `description:` in the local `check-rebuild` and `s3_upload` composite actions) are unrelated to this diff.

**9. End-to-end CI signal (확인 필요 / needs confirmation)**

At the time of this review, this PR's own `pdebuild.yml` run (32985074747) was still queued and had not executed the changed step yet, so I could not directly observe the new code path running green (or failing) on this exact PR. Worth checking before merge that the "prepare pdebuild" step actually completes as expected.

---

**Summary of findings**
- Blocker: none
- Major: none
- Minor: (3) the pipeline's safety depends on the runner's undocumented "no pipefail by default" behavior — fragile against a future `shell: bash` edit; (6) identical `&&`-masking pattern left unfixed in 7 other workflow files (reasonable to leave for a follow-up, not this PR)
- Nit: (7) no automated regression test for this bash snippet's behavior
- Needs confirmation: (9) this PR's own pdebuild CI run had not completed at review time

No functional or logic errors found in the diff itself; the stated problem (silently-swallowed `apt-get update` failure) is genuinely fixed, and the Microsoft-source workaround is well-targeted and verified against the real runner image's actual apt sources.

myungjoo:

Thanks — the pipefail point is a real one and is now fixed. Per-item response:

### Minor 1 — pipefail dependency: **fixed** in d4c93b52

Reproduced first, to be sure the concern was real and not theoretical:

```
$ bash -e         -c "grep -rl 'packages\.microsoft\.com' empty/ | xargs -r rm -fv; echo REACHED"
REACHED                       # runner default: fine
$ bash -eo pipefail -c "grep -rl 'packages\.microsoft\.com' empty/ | xargs -r rm -fv; echo REACHED"
                              # rc=1, step dies on the *normal* no-match path
```

So the old form worked only because `shell:` was left unspecified. Adding `shell: bash` later — the documented spelling, and a plausible tidy-up — would have broken the step on every run where there is nothing to remove. New form:

```bash
{ sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
```

The group neutralizes grep's status so the pipeline status comes from `xargs` alone, and `-Z`/`-0` delimits filenames unambiguously. Re-verified under **both** `bash -e` and `bash -eo pipefail`, for no-match / matches-present / missing-directory: exit 0 in all six combinations, `keep.list` untouched, both `.list` and deb822 `.sources` removed.

### Minor 2 — the same `&&` pattern in 7 other files: **scope kept as-is**, with a correction

Checking each site, the `&&` only masks when the list is **not** the last command of the step script — otherwise the list's own non-zero status becomes the step's status and the failure is already honest. Three of the cited sites are single-line `run:` steps and are therefore **not** affected:

- `publish_docs.yml:48`, `static.check.yml:24`, `update_gbs_cache.yml:40` — last command, status propagates, no bug.

The genuinely-masking sites are 7, across 5 files: `codeql.yml:76,78`, `gbs_build.yml:42`, `nns_testhub.yml:24`, `publish_docs.yml:20`, `ubuntu_clean_meson_build.yml:31,33`.

Not folding them in here, for a reason beyond PR size: the risk profile differs per file. In pdebuild (and in `gbs_build`/`nns_testhub`) the skipped install is unconditionally fatal downstream — `pbuilder`/`gbs` are not on the runner image — so splitting only moves the failure to its true cause. In `codeql.yml`, `ubuntu_clean_meson_build.yml` and `publish_docs.yml:20`, several of the skipped packages *are* preinstalled on the runner image, so a job that currently limps to green through a transient mirror hiccup would start failing outright. That is a behavior change needing per-file judgment, not a mechanical sweep, and it does not belong in a fix for the #4238 failure.

### Nit — no automated regression test: agreed, and inherent

An inline `run:` script has no unit-test surface in this repo. The practical guard is that `check_if_rebuild_requires.sh` routes any `.github/workflows/*` change through the catch-all arm, setting `redebian=1`, so the pdebuild job itself exercises the changed step on this and every future PR touching it. The shell semantics above were verified by direct execution rather than by assertion.

### 확인 필요 — pdebuild CI not yet observed: tracking

Agreed; the run had not finished at review time. Not marking this ready for review until the pdebuild job on this PR has actually gone green on the changed step.

myungjoo:

**This is a second-pass review.** Scope reviewed: `.github/workflows/pdebuild.yml` and `.github/workflows/update_pbuilder_cache.yml` only (the two files this PR touches). I did not take the prior review (comment 5427560711) or the author's reply (comment 5427590790) at face value — I re-derived every factual claim independently: reproduced the shell semantics myself in a real bash (GNU bash 5.2, GNU grep 3.0, GNU findutils 4.10 — same tool versions as the Ubuntu 22.04 runner), pulled this repo's own raw Actions logs, read `actions/runner-images`' published Ubuntu 22.04 manifest, and read `check_if_rebuild_requires.sh` and every other workflow file cited in the scope discussion directly from the tree.

**Overall: still no blockers or majors. Prior Minor #1 is resolved by d4c93b52. The scope decision (Minor #2) checks out factually. One item remains open, but it is a CI/infra state, not a code defect.**

---

## Disposition of each item from the first review

**Minor 1 (pipefail dependency in the cleanup pipeline) — RESOLVED, verified independently.**

I built three synthetic `sources.list.d` fixtures (no match / a classic `azure-cli.list` + a deb822 `microsoft-prod.sources` + an unrelated `keep.list` / a missing directory) and ran the new line
```
{ sudo grep -rlZ 'packages\.microsoft\.com' /etc/apt/sources.list.d/ || true; } | sudo xargs -0r rm -fv
```
under both `bash -e` and `bash -eo pipefail`. All 6 combinations exit 0, the script continues past the line, only the two Microsoft-matching files are ever removed, and `keep.list` is untouched in every case. I also reproduced the *old*, pre-d4c93b52 form (`grep -rl ... | xargs -r rm -fv`, no group, no `-Z/-0`) side by side: under `bash -e` it prints a following `echo REACHED` (masks fine, matching the original report), but under `bash -eo pipefail` on the ordinary no-match path it aborts before reaching that echo at all (exit 1). So the defect the first review flagged was real, not theoretical, and the new grouping genuinely fixes it. `-Z`/`-0` pairing is the correct GNU grep/xargs idiom for NUL-delimited filenames and both tools on the target image support it.

I also independently confirmed the platform fact this fix depends on: a `run:` step with no `shell:` key on this repo's actual runner is literally invoked as `shell: /usr/bin/bash -e {0}` (checked against the raw log of a completed run in this repo, `gh run view 32981441328 --log`) — no pipefail — while `shell: bash` explicitly would switch to `-eo pipefail`. So the commit message's stated rationale ("adding `shell: bash` later would flip this") is accurate, and the fix's whole point — behaving identically either way — is now true by direct test, not by relying on the undocumented default persisting.

**Minor 2 (same `&&`-masking pattern left in 7 other workflow files) — scope decision independently verified as factually grounded, no issue.**

I re-derived the classification myself instead of trusting the reply:
- Confirmed by exact line number that `publish_docs.yml:48`, `static.check.yml:24`, `update_gbs_cache.yml:40` are each the step's *only* `run:` line (single-line `run:`, no block) — so the `&&` list's own exit code is already the step's exit code. No masking there; author's claim is correct.
- Confirmed by exact line number that `codeql.yml:76,78`, `gbs_build.yml:42`, `nns_testhub.yml:24`, `publish_docs.yml:20`, `ubuntu_clean_meson_build.yml:31,33` are each followed by further commands in the same multi-line `run: |` block (further `apt-get install` lines, `cp`, `pip install`, `meson setup/compile`, etc.) — so these genuinely have the same masking exposure as the bug this PR fixes. Author's claim is correct.
- Checked the "some of these packages are preinstalled on the runner image" justification against GitHub's own published manifest for the image in use (`actions/runner-images`, Ubuntu 22.04 readme). Confirmed preinstalled: `cmake`, `pkg-config`, `flex`, `bison`, `python3`, `python3-pip`, `git`, `binutils`, `shellcheck`, `clang-format`/llvm toolchain — all of which appear in the apt-get install lists of the files left alone (`publish_docs.yml:20` installs `cmake`; `codeql.yml`/`ubuntu_clean_meson_build.yml` install `flex`/`bison`/`pkg-config`/`python3-dev`, etc.). This makes the stated risk (splitting could convert a job that currently limps to green off an already-satisfied dependency into a hard failure on a transient mirror hiccup) a concrete, evidence-backed concern rather than a hand-wave, and supports keeping this PR scoped to the two files that actually caused the #4238 incident.
- Also confirmed `gbs`/`rpm2cpio`/`cpio` (used in `gbs_build.yml`/`nns_testhub.yml`, which the author says are "unconditionally fatal downstream" like pbuilder) are *not* in the preinstalled list — consistent with treating those two as same-risk-profile as pdebuild, which the author's reply does not fold in either (correctly, per its own stated reasoning about differing risk profiles being a per-file judgment call).

Conclusion: the scope-limiting argument is factually sound, not just plausible-sounding. No expansion needed for this PR.

**Nit (no automated regression test for this bash snippet) — still open, and inherently so; the stated mitigation is verified real.**

Read `.github/actions/check-rebuild/check_if_rebuild_requires.sh` directly: any file not matching `*.md|*.png|*.webp|*.css|*.html`, `packaging/*`, `debian/*`, `jni/*` falls into the catch-all case (line 47-52) which sets `redebian=1` unconditionally. `.github/workflows/*.yml` changes hit that catch-all, so the claim that "the pdebuild job itself exercises the changed step on this and every future PR touching it" is correct. This doesn't produce a synthetic-fixture regression test, but it does guarantee the changed step is not silently skipped by the rebuild-gate logic. No code action needed; this is an acceptable, inherent limitation of an inline shell snippet in workflow YAML.

**"확인 필요" #9 (end-to-end CI green on this PR) — still NOT resolved; not a code issue, but worth flagging explicitly before merge.**

I checked this PR's own `pdebuild.yml` runs directly:
- The run for the pre-pipefail-fix commit (`2f826ac5`, run `32985074747`) is still sitting in `queued` status (unstarted) as of this review — consistent with a broader CI backlog I also observed on several unrelated PRs' jobs in this repo at the same time, not specific to this PR.
- The run for the current head commit (`d4c93b5`, run `32985848700`) completed with `conclusion: startup_failure`. I fetched the run page directly: the banner is GitHub's generic "An unexpected error has occurred and we've been automatically notified... Errors are sometimes temporary" with a support request ID — i.e., a GitHub Actions platform-side start failure, not a YAML/workflow content error (the job itself never entered `in_progress`, and 0 annotations were attached to the check run). This is not evidence of a defect in the diff.

So the "changed step actually goes green" confirmation that the author explicitly said would gate marking this ready for review has still not happened, for reasons external to this PR's code. The PR is correctly still `draft` with the `DO NOT MERGE` label at review time, which matches the author's own stated intent ("Not marking this ready for review until the pdebuild job on this PR has actually gone green on the changed step"). Recommend re-running/watching this PR's own `pdebuild.yml` job once the platform/queue issue clears, before flipping to ready-for-merge — this is process hygiene, not a code change request.

---

## Fresh review of the new code in d4c93b52 (not covered by the first pass)

- The `{ ... || true; }` grouping correctly isolates `grep`'s exit status from the pipeline's status without needing `pipefail` either way; `xargs -0r` supplies the actual pipeline exit code (0 unconditionally, thanks to `-r`/`--no-run-if-empty` plus a normal `rm -fv` exit).
- `sudo` is applied on both sides of the pipe (`sudo grep` and `sudo xargs ... rm`), which is correct and not just cosmetic: an unprivileged `grep` could silently fail to read a root-only source file and never report it as a match.
- No risk of over-matching or deleting the wrong file: `-l` only prints filenames of matching files, the fixed unrelated file in my `keep.list` test was untouched in every one of the 6 combinations, and content-matching on the hostname (`packages\.microsoft\.com`) rather than a specific filename correctly covers both classic `.list` and deb822 `.sources` formats.
- Comment density is consistent with this file's pre-existing style (the file already carried explanatory comments before this PR, e.g. `## prevent permission error`), and the two new comments explain non-obvious shell semantics (why `grep`'s exit 1 is not a failure, why update/install must stay split) rather than restating what the code already says. No excess, nothing missing.
- Both commit messages (`2f826ac5`, `d4c93b52`) accurately describe what their respective diffs do and why; I did not find any discrepancy between the prose and the actual code changes or my own test results.
- Checked whether the pre-existing trailing `sudo ln -s ~/.pbuilderrc /root/` (unchanged by this PR) still determines the step's final exit status in a way that matters: it does, but that's expected/correct behavior now — `apt-get update`/`apt-get install` are separate errexit-checked statements ahead of it, so a setup failure aborts before reaching `ln -s` at all (the bug this PR fixes). `ln -s` itself failing (e.g., if the target already existed) would be a genuine, separate, pre-existing-and-untouched-by-this-PR concern, and is not applicable in practice since GitHub-hosted runners are fresh VMs per job. Not an issue introduced or left behind by this diff.
- No regression risk to other modules: the diff touches only these two CI workflow files, does not alter any other job, and does not change any package version or install list — it only reorders/splits/prunes commands that were already there.
- Diff size is proportionate to the bug: ~10 added lines per file, mechanically identical between the two files (as the PR body states, `update_pbuilder_cache.yml` "carried the same step verbatim").

---

## Summary

- Blocker: none
- Major: none
- Minor: none remaining — prior Minor #1 is resolved and independently verified; prior Minor #2's scope justification is independently verified as factually accurate, not merely asserted
- Nit: (unchanged, inherent) no automated regression test for this inline bash snippet; mitigated by `check_if_rebuild_requires.sh` routing every `.github/workflows/*` change through the pdebuild job itself
- Needs confirmation before flipping out of draft: this PR's own `pdebuild.yml` run has not yet gone green on the changed step — currently blocked by what appears to be a transient GitHub Actions platform/queue issue unrelated to this diff's content, not a defect in the code

**Merge verdict: the code change itself is correct, well-scoped, and ready to merge.** No outstanding code-level issues block it. Per the author's own stated (and reasonable) gating condition, hold the actual merge until this PR's own `pdebuild.yml` job is observed going green end-to-end on the changed step — that is a CI-state check, not a code fix.

myungjoo:

### CI confirmation

The pdebuild job on this PR completed green end to end — [run 32985848700, attempt 2](https://github.com/nnstreamer/nnstreamer/actions/runs/32985848700):

```
8. prepare pdebuild      -> success
9. make pbuilder base.tgz -> success     <- the step that produced "sudo: pbuilder: command not found"
10. run pdebuild          -> success
```

Two things in the job log are worth recording, because they confirm assumptions this PR was resting on:

**1. The runner really does use `bash -e` with no pipefail.**

```
shell: /usr/bin/bash -e {0}
```

So the pre-fix pipeline was passing only because pipefail happened to be off, exactly as the first review argued. The `|| true` group in d4c93b52 makes that irrelevant either way.

**2. The real runner ships the Microsoft sources in *two different formats*:**

```
##[group]drop unused third-party apt sources
removed '/etc/apt/sources.list.d/microsoft-prod.list'
removed '/etc/apt/sources.list.d/azure-cli.sources'
```

One classic `.list` and one deb822 `.sources`. This is a good argument for having matched on the URI rather than the filename: the obvious `sudo rm -f /etc/apt/sources.list.d/microsoft*.list` spelling would have removed `microsoft-prod.list` and silently left `azure-cli.sources` behind — i.e. it would have looked like it worked while still failing on exactly the 403 that motivated #4238.

`pbuilder` then installed normally (`Setting up pbuilder (0.231build1) ...`) and both later pbuilder steps succeeded.

### On the earlier red checks

The `startup_failure` / `failure` results on this PR before the reruns were a GitHub Actions platform incident, not this change:

- 16–17 `startup_failure` runs across five unrelated branches at the same time.
- The first pdebuild run was marked `failure` at `16:08:11` while its only job *started* at `16:08:13` — the run was failed two seconds before the job began, and the job was left orphaned in `in_progress` with no retrievable log.
- `Spell Check` reported `failure` with `total_count: 0` jobs — nothing ever ran.
- The Valgrind job was `cancelled`, not failed.

All have been re-run. Spell Check now passes, and was independently verified locally with the repo's own config (`typos --config ./.github/workflows/_typos.toml`, clean on the changed files and on the whole tree), so the added words (`errexit`, `pipefail`) are not a problem for `crate-ci/typos`.

myungjoo:

Thanks for the approval. Two of the three non-blocking suggestions are now addressed in a9f36ebd; the third I'd like to argue against.

### 2 — `apt-get update` retry: **done**

Fair point, and it was the half-finished part of this PR: dropping the Microsoft sources covers the repos we *don't* need, and does nothing for the ones we do. Now:

```bash
for i in 1 2 3; do
  sudo apt-get update && break
  if [ "$i" = 3 ]; then
    echo "::error::apt-get update failed after $i attempts"
    exit 1
  fi
  sleep $((i * 10))
done
```

The `&& break` is deliberate and leans on exactly the errexit rule this series is about — a failed attempt must not abort the step — while exhausting all three still exits non-zero, so the "honest failure" property is preserved rather than traded away for flake tolerance. Exercised locally under `bash -e` against a stub that fails 0 / 2 / 9 times: succeeds immediately, recovers on the third attempt, and exits 1 with the `::error::` annotation respectively; in the first two cases execution continues to the install.

Only the update is retried. An `apt-get install` failure is a real error, not a network blip, and retrying it would re-introduce the class of masking this PR exists to remove.

### 3 — `ln -s` idempotency: **done**

Switched to `ln -sf`. You're right that it's the command whose exit status the step reports, which makes it in-scope for a PR about that step failing honestly. Unreachable on today's fresh-VM runners, but a cached or reused image would resurrect a confusing failure in precisely the step this series just taught to fail correctly — cheap to close off.

### 1 — composite action for the duplicated block: **not in this PR**

Agreed on the diagnosis: verbatim duplication is what let the original bug live in two places. But the extraction isn't as clean as the byte-identical diff suggests — the surrounding steps differ (`update_pbuilder_cache.yml` has no `if: env.rebuild` gate, and its `OTHERMIRROR` omits the kubuntu backports-extra entry), so a shared action would need parameterising beyond the ~10 identical lines. That's a CI restructure with its own review surface, and folding it into a fix for the #4238 incident would make this PR considerably harder to reason about than the bug it closes. Better as its own change.

myungjoo:

**Third-pass review.** Scope: `.github/workflows/pdebuild.yml` and `.github/workflows/update_pbuilder_cache.yml` (the only two files touched), focused on the new commit `a9f36ebd` (retry loop for `apt-get update`, `ln -s` → `ln -sf`) added in response to the APPROVED review's three non-blocking suggestions. I did not take the author's reply at face value; each claim below was independently reproduced.

---

## Disposition of the APPROVED review's 3 non-blocking suggestions

**#2 (retry loop for `apt-get update`) — RESOLVED, verified by reproduction.**

Built a stub `apt-get` controllable to fail N times, and ran the exact loop body from the diff under `bash -e` (the real runner's actual invocation, confirmed by both prior reviews from raw Actions logs: `shell: /usr/bin/bash -e {0}`, no pipefail):

| stub behavior | result |
|---|---|
| succeeds immediately | 1 attempt, exit 0, continues to install |
| fails twice, succeeds 3rd | 3 attempts, exit 0, continues to install |
| always fails | 3 attempts, `::error::apt-get update failed after 3 attempts`, exit 1, step aborts |

All three match the author's claimed 0/2/9-failure verification exactly. Confirmed the specific errexit rule this relies on: for a two-command list `cmd1 && cmd2`, bash's `-e` only aborts on failure of the command *following the final `&&`* — here that's `break`, not `apt-get update` — so a failed attempt genuinely cannot abort the loop early, and this is not coincidental but the documented bash `-e` semantic for AND-lists. The trailing `exit 1` after the 3rd failure is a plain top-level statement (not part of any `&&`/`if` condition), so it does fail the step, and since neither file sets `continue-on-error` anywhere (checked via grep, zero matches), a step failure here fails the job as intended — the "honest failure" goal is preserved, not traded away.

`$i` and `$((i * 10))` are plain POSIX shell constructs; GitHub Actions only rewrites `${{ ... }}` at the workflow-parse stage (which happens before the script ever reaches bash), so there is no interaction with these variables — confirmed no `${{ i }}`-style token appears anywhere near the loop, and both files still parse as valid YAML and produce no new `actionlint` findings (only the two pre-existing, unrelated "description is required" notices on the local composite actions, same as before this commit).

Retrying only `apt-get update` and not `apt-get install` is the right call: `apt-get update` is naturally idempotent/re-runnable (it just re-fetches index files), while retrying a partially-failed `apt-get install` risks masking a real dependency/dpkg-state problem behind a "worked on retry" result — exactly the kind of masking this whole PR series exists to remove.

Cross-checked against this PR's own CI: on the current head (`a9f36ebd`), run 33078141430 / job 98537727572, the "prepare pdebuild" step (containing this exact loop) completed successfully (✓), consistent with the retry loop working correctly on the real runner and not needing to exercise its failure branch this time.

**#3 (`ln -s` → `ln -sf`) — RESOLVED, safe.**

`sudo ln -sf ~/.pbuilderrc /root/` — since `/root/` is passed with a trailing slash and exists as a directory, this creates/overwrites only the specific path `/root/.pbuilderrc` (never touches the directory itself or any sibling file). `-f` unlinks an existing regular file, dangling symlink, or live symlink at that exact path before relinking; it does not follow/replace a directory at that path. No blast radius beyond the one file whose staleness this line exists to fix, and it's a strict superset of the old behavior on today's fresh-VM runners (where `/root/.pbuilderrc` never pre-exists, so `-f` is a no-op there). Matches the author's stated rationale exactly.

**#1 (composite action to de-duplicate) — rejection verified factually correct, no issue.**

Read both files directly to check the author's counter-argument:
- `pdebuild.yml`'s "prepare pdebuild" step carries `if: env.rebuild == '1'` (line 60); `update_pbuilder_cache.yml`'s equivalent step has **no** `if:` gate at all — confirmed, it's unconditional.
- `pdebuild.yml`'s `OTHERMIRROR` line includes a third mirror entry, `...kubuntu-ppa/backports-extra/ubuntu ${{ matrix.distroname }} main`, appended after the nnstreamer PPA entry; `update_pbuilder_cache.yml`'s `OTHERMIRROR` line ends after the nnstreamer PPA entry and does **not** have that third segment — confirmed, byte-for-byte different.

Both differences are real, not hand-waved, so a naive extraction into a shared composite action would need parameters for both the gate and the mirror-list content — it is not the pure copy-paste the identical ~10-line diff might suggest. Deferring that restructure to a separate PR is a reasonable, correctly-justified scope decision, not a rejected-without-basis complaint.

---

## New observations on `a9f36ebd` itself (informational, not blocking)

- **Nit:** the retry loop wraps only the explicit `sudo apt-get update` line. It does not cover the implicit `apt-get update` that `add-apt-repository ppa:nnstreamer/ppa` runs internally immediately beforehand (unless `-n` is passed, which it isn't) — a transient failure there still hard-aborts the step with no retry, same as before this commit. Not a regression introduced by this diff (pre-existing since the earlier commits in this series, and arguably out of scope for a commit titled "retry apt-get update"), just worth knowing if that specific line is ever the next source of flakiness.
- **Nit (carries forward from the 1st/2nd-pass reviews' "no automated regression test" point):** `update_pbuilder_cache.yml` only triggers on `schedule`/`workflow_dispatch`, so this PR's own CI never actually executes that file's copy of the new retry loop / `ln -sf` — verification there rests on the two files being byte-identical (confirmed via diff) plus the reasoning above, not on an observed green run. This is an existing structural property of the file, unchanged by this commit; it will next actually run at the following scheduled cron invocation.

Neither item rises above Nit — both are pre-existing characteristics of the surrounding design, not defects introduced by `a9f36ebd`.

---

## Summary

- Blocker: none
- Major: none
- Minor: none
- Nit: (1) implicit `add-apt-repository` update is not covered by the new retry loop; (2) `update_pbuilder_cache.yml`'s copy of this code is not exercised by pull-request CI, only by the daily/manual schedule
- All three of the APPROVED review's non-blocking suggestions are correctly closed out: two implemented and verified working (retry loop, `ln -sf`), one declined with a rejection rationale that checks out against the actual file contents (differing `if:` gate and `OTHERMIRROR` content).

**Merge verdict: mergeable.** No outstanding code-level issues in `a9f36ebd` or the PR as a whole. This PR's own `pdebuild.yml` run has been observed completing the changed "prepare pdebuild" step successfully on the current head commit.

myungjoo:

Third-pass nits, dispositioned.

### Nit 1 — retry did not cover `add-apt-repository`'s implicit update: **fixed** in 97722481

Correct, and it made the retry half-useless: on jammy `add-apt-repository` runs its own `apt update` after writing the source, on its own line under errexit, so a transient failure there aborted the step regardless of the loop sitting immediately below it. Now:

```bash
sudo add-apt-repository -y -n ppa:nnstreamer/ppa
```

`--no-update` leaves it to only write the source; the retried `apt-get update` that follows fetches the PPA along with everything else. The redundant second update disappears as a bonus, and `-y` matches how `codeql.yml:78` and `ubuntu_clean_meson_build.yml:33` already invoke it.

The behavioral dependency here is that `--no-update` exists on the runner's `software-properties-common`. This PR's own pdebuild job exercises the line, so CI is the check rather than my assertion — flagging it explicitly rather than leaving it implied.

### Nit 2 — `update_pbuilder_cache.yml` is not exercised by PR CI: **acknowledged, not fixable here**

Accurate: it triggers only on `schedule` and `workflow_dispatch`, so no pull_request run touches it, and confidence in its copy rests on the diff being byte-identical to the pdebuild one rather than on an observed run. Adding a `pull_request` trigger to a daily cache-refresh job would be a worse trade (a full pbuilder create + build on every PR) than the risk it retires, so I'm leaving the trigger alone. Worth noting for anyone editing that file later: it has no PR-time safety net.

myungjoo:

### Final CI state — all 12 checks green on 97722481

Including the one behavioral dependency I flagged as needing CI rather than assertion: `--no-update` is supported on the runner and behaves as intended. From the pdebuild job log ([run 33078731290](https://github.com/nnstreamer/nnstreamer/actions/runs/33078731290)):

```
sudo add-apt-repository -y -n ppa:nnstreamer/ppa
Repository: 'deb https://ppa.launchpadcontent.net/nnstreamer/ppa/ubuntu/ jammy main'
More info: https://launchpad.net/~nnstreamer/+archive/ubuntu/ppa
...
##[group]drop unused third-party apt sources
removed '/etc/apt/sources.list.d/microsoft-prod.list'
removed '/etc/apt/sources.list.d/azure-cli.sources'
...
Get:19 https://ppa.launchpadcontent.net/nnstreamer/ppa/ubuntu jammy InRelease [18.0 kB]
Get:20 https://ppa.launchpadcontent.net/nnstreamer/ppa/ubuntu jammy/main amd64 Packages [16.1 kB]
```

`add-apt-repository` writes the source and stops — no apt update of its own — and the PPA is then fetched by the retried `apt-get update` at `Get:19`, i.e. genuinely inside the retry umbrella rather than ahead of it. `Setting up pbuilder (0.231build1) ...` follows, and both later pbuilder steps pass.

### One transient failure along the way, for the record

`Minimal meson build in Ubuntu with Valgrind` failed once on this commit at `unittest_filter_llamacpp.cc:309` (`wait_for_samples (2)` timing out with 1 sample) and passed on re-run of the same commit. That is the known flake tracked as #4883, and it cannot be reached from this diff: this PR changes only `pdebuild.yml` and `update_pbuilder_cache.yml`, neither of which that job reads, so the source under test is byte-identical to `main`. The immediately preceding commit on this same branch (a9f36ebd) passed the same job, and the two commits differ by two lines in files it never opens.

No outstanding review items remain.

Issue #3495, opened 2021-09-16
Migration from JCenter to MavenCentral (or other alternative)


myungjoo:
JCenter is deprecated. We need to migrate to somewhere else.

myungjoo:
Closing: JCenter/bintray shut down in 2021, so the migration target itself no longer exists. The underlying distribution need was solved a different way instead — Android artifacts are published via S3 and surfaced at release.nnstreamer.com through nnstreamer/api's daily-build-android.yml workflow, not through a Maven repository. `org.nnsuite` was never renamed to `org.nnstreamer`, and there is no MavenCentral publishing config in this repo or in nnstreamer/api. If real Maven/MavenCentral artifact publishing is still wanted, it should be re-filed against nnstreamer/api, where the Android bindings actually live. (The stale JCenter/bintray links still in README.md and Documentation/ are being cleaned up separately.)

PR #4878, opened 2026-08-26
[Doc] Replace dead JCenter references with the daily release channel

myungjoo:
## What

Three docs still send Android users to JCenter/bintray, which has served nothing since bintray shut down in 2021:

| Location | Before |
| --- | --- |
| `Documentation/API-reference.md:54` | `[Available at JCenter](https://bintray.com/beta/#/nnsuite/nnstreamer?tab=packages)` |
| `Documentation/component-description.md:190` | `Android WIP: JCenter Repository & Daily Build Release` |
| `README.md:43` | `Android Studio (JCenter, "nnstreamer")` |

All three now point at the daily build release instead.

## Why this target

There is no Maven/MavenCentral publication today — `jcenter`/`bintray`/`maven` appear nowhere in the build or publish configuration of either `nnstreamer` or `nnstreamer/api`, only in these stale doc strings. The real Android distribution is the daily build produced by `nnstreamer/api`'s `.github/workflows/daily-build-android.yml` and surfaced at release.nnstreamer.com.

Verified against the release bucket that `nnstreamer/latest/android/` currently holds `nnstreamer.aar`, `nnstreamer-lite.aar` and the matching native zips, so the artifact name quoted in the docs is the published one. `README.md:42` already links this same path from the publish table, so the three references are now consistent with it.

## Notes

- The `component-description.md` entry is reformatted to match the neighbouring Tizen / Ubuntu / Yocto rows (`- <Platform> [<Channel>](<url>)`).
- The `WIP` marker is dropped from that line. Unlike the macOS entry below it, which has no automated publication, Android arm64 is published automatically and is already marked *Available* with a build badge in the README table, so it belongs with the other live channels.
- `CHANGES:455` and `CHANGES:465` also mention JCenter, but those are historical 0.2/0.3-era release notes describing what was true at the time and are left untouched.

Documentation only; no functional change.

## Relation to #3495

This does **not** implement the JCenter → MavenCentral migration tracked by #3495. That issue asks for a MavenCentral repo, a domain rename and a publishing action — none of which this touches. #3495 stays open; this only stops the docs from advertising a dead channel in the meantime.

## Testing

No test is added: the change is three markdown strings with no build, packaging or code surface. Both changed `Documentation/*.md` files are listed in `Documentation/hotdoc/sitemap.txt`, so they flow into the published doc site; the links are ordinary inline markdown and render like the surrounding entries.

Ran locally against the changed files:

- `static.check.scripts/prohibited-words.sh` — pass
- `git diff --check` — clean, no CRLF introduced
- Confirmed no `jcenter`/`bintray` reference remains in any `*.md`

I looked at wiring a permanent guard (adding `bintray`/`jcenter` to `prohibited-words.txt`) and deliberately did not, for a reason worth recording separately: that checker cannot fail today. See the review discussion below.

### DIFF
--- a/Documentation/API-reference.md
+++ b/Documentation/API-reference.md
@@ -51,7 +51,7 @@ Note that this supports SingleShot/Pipeline subdomains and do not include traini
 - [Android sample app](https://github.com/nnstreamer/nnstreamer-example/tree/master/android/example_app/api-sample) uses JAVA APIs to implement Android-NNStreamer apps.
     - Note that the Android Sample Applications published via Google Play Store, [Source Code](https://github.com/nnstreamer/nnstreamer-example/tree/main/android/example_app), are developed before NNStreamer Java API. They use GStreamer functions.
 - [Implementation](https://github.com/nnstreamer/api/tree/main/java) (stable)
-- [Available at JCenter](https://bintray.com/beta/#/nnsuite/nnstreamer?tab=packages)
+- [Available at Daily Build Release](https://release.nnstreamer.com/#nnstreamer/latest/android/) (`nnstreamer.aar`)
 
 
 ### Python API is planned
--- a/Documentation/component-description.md
+++ b/Documentation/component-description.md
@@ -187,7 +187,7 @@ Note that test elements in /tests/ are not elements for applications. They exist
   - Tizen (since 5.5 M1) [Package Download](https://download.tizen.org/snapshots/TIZEN/Tizen/Tizen-Unified/latest/repos/standard/packages/) [Build & Release Infra](https://build.tizen.org/project/show/Tizen:Unified)
   - Ubuntu [Launchpad PPA](https://launchpad.net/~nnstreamer/+archive/ubuntu/ppa)
   - Yocto/OpenEmbedded [OpenEmbedded Layer, "meta-neural-network"](https://layers.openembedded.org/layerindex/branch/master/layer/meta-neural-network/)
-  - Android WIP: JCenter Repository & Daily Build Release
+  - Android [Daily Build Release](https://release.nnstreamer.com/#nnstreamer/latest/android/)
   - macOS WIP: Daily Build Release
 - [Test cases](https://github.com/nnstreamer/nnstreamer/tree/main/tests/): Mandatory unit test cases required to pass for each PR.
   - Used [SSAT](https://github.com/myungjoo/SSAT).
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ neural network developers to manage neural network pipelines and their filters e
 - Ready: CI system ensures build-ability and unit-testing. Users may easily build and execute. However, we do not have automated release & deployment system for this instance.
 - Available: binary packages are released and deployed automatically and periodically along with CI tests.
 - [Daily Release](https://release.nnstreamer.com/#nnstreamer/latest/)
-- SDK Support: Tizen Studio (5.5 M2+) / Android Studio (JCenter, "nnstreamer")
+- SDK Support: Tizen Studio (5.5 M2+) / Android Studio (`nnstreamer.aar` from [Daily Release](https://release.nnstreamer.com/#nnstreamer/latest/android/))
 - [Enabled features of official releases](Documentation/features-per-distro.md)
 
 


myungjoo:
### Why no permanent guard was added here

The natural place to stop a dead-repository reference from coming back is `prohibited-words.txt`, checked by `/Checker/ prohibited words` in `static.check.yml`. Adding `bintray`/`jcenter` there would be inert, because the checker cannot currently fail:

https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/prohibited-words.sh#L91

```sh
bad_words_log_file=$(mktemp)      # line 74
...
$bad_words_sw $bad_words_rules $target_files > ${bad_words_log_file}   # line 85
cat ${bad_words_log_file}                                             # line 88
result_count=$(cat ${bad_word_log_file} | grep -c '^' )               # line 91 -- bad_word_log_file
```

Line 91 reads `${bad_word_log_file}` (no `s`), which is unset, so `cat` reads stdin instead of the grep output. Under Actions stdin is empty, `result_count` is always 0, and the step always exits 0. Running it locally on this PR's files reproduces it — grep reports a hit and the script still passes:

```
README.md:127:- [TAOS-CI config files for nnstreamer](.TAOS-CI).
There is no prohibited word found in this PR.
exit=0
```

The one-character fix is not safe to bundle here. `prohibited-words.txt` currently contains `file`, and the rules are a plain `grep -f`, so repairing the variable name would immediately fail nearly every PR that touches a `.c`/`.h`/`.md` file — the hit above is exactly that. Reviving the checker means pruning the word list first and deciding what belongs in it, which is a separate change with its own blast radius and does not belong in a three-string docs fix.

Flagging it here rather than silently widening this PR. Happy to file it as its own issue/PR if that is the preference.

myungjoo:
*This review was written by a separate automated review agent (not the PR author/committer), invoked to independently audit this PR. It is being transcribed here as-is.*

## Scope check

The diff touches exactly 3 lines across 3 files (`Documentation/API-reference.md`, `Documentation/component-description.md`, `README.md`), all replacing dead JCenter/bintray references with the existing daily-build release channel. No code, build, or CI files are touched. This matches the stated topic precisely — no scope creep, no unrelated modules touched.

## Verification performed

- Confirmed via `grep -rniE "jcenter|bintray"` across the full working tree (post-patch) that no other doc/build/CI file references JCenter/bintray. The only remaining hits are `CHANGES:455` and `CHANGES:465`, which are historical 0.2/0.3-era release notes correctly left untouched (rewriting historical changelog entries would be inaccurate).
- Confirmed `jcenter`/`bintray`/`maven` do not appear in any build or publish configuration in this repo, so the change is genuinely documentation-only with no functional/packaging impact — no regression risk to other modules.
- Verified the replacement URL's factual claim: queried the S3 bucket backing `https://release.nnstreamer.com/#nnstreamer/latest/android/` directly (`nnstreamer-release.s3-ap-northeast-2.amazonaws.com?prefix=nnstreamer/latest/android/`) and confirmed it currently holds `nnstreamer.aar`, `nnstreamer-lite.aar`, and matching native zips — the artifact name quoted in the docs (`nnstreamer.aar`) matches what's actually published.
- Confirmed both changed `Documentation/*.md` files are listed in `Documentation/hotdoc/sitemap.txt`, so the fix will propagate to the published doc site with no further doc wiring needed.
- Confirmed `README.md:42` (the existing publish table) already links this exact same URL, so the three edits are now consistent with the existing publish table rather than introducing a new/unverified channel.
- `git diff --check` is clean and no CRLF was introduced in the changed lines.
- Confirmed issue #3495 (JCenter → MavenCentral migration) is still open, matching the PR's statement that this change does not resolve #3495 and is only a stop-gap for the dead link.
- Independently reproduced the bug the author flagged in `.github/workflows/static.check.scripts/prohibited-words.sh` (line 91 references `${bad_word_log_file}`, a variable that is never set, instead of `${bad_words_log_file}`, so `result_count` is always 0 and the prohibited-words check can never fail). This confirms the author's stated reason for not adding a permanent `prohibited-words.txt` guard in this PR is accurate, and that the underlying checker bug is pre-existing and out of scope here — correctly not bundled into this docs-only fix.

## Findings

No blocking (High/Medium) issues found.

**Low / informational (non-blocking, does not need to hold up this PR):**

1. There is no automated link-checker in CI (checked all workflows under `.github/workflows/`; no `lychee`/`markdown-link-check`/equivalent). This means a future dead link (including a regression of what this PR just fixed) would not be caught automatically and would silently ship, same as the original JCenter links did. This is a pre-existing repository-wide gap, not something introduced by this PR, and adding a link checker is out of proportion to a 3-line doc fix — flagging only as a possible follow-up issue, not a request to expand this PR.
2. The `prohibited-words.sh` bug described above is real and already fully disclosed by the PR author in the PR description/discussion, with a correct rationale for not fixing it here (the word list would need pruning first, which has much broader blast radius). No action needed on this PR; noting it only to confirm the author's due-diligence claim independently rather than take it at face value.

No security concerns and no backdoor-shaped code found — the change is plain Markdown text pointing at the project's own existing, already-used release domain (`release.nnstreamer.com`), with no code, script, or CI logic changed.

## Verdict

**Approve.** The PR fully resolves the stated topic (removing three dead JCenter/bintray references), is minimal and precisely scoped, does not touch or risk any other module, requires no test cases given its non-functional documentation-only nature, and its factual claims about the replacement URL/artifact were independently verified to be correct.

PR #4880, opened 2026-08-27
[CI] Repair the prohibited-words checker and pin it with a self-test

myungjoo:
## The checker cannot fail today

`/Checker/ prohibited words` in `static.check.yml` runs `prohibited-words.sh`, whose Step 3 is:

```sh
bad_words_log_file=$(mktemp)                                        # line 74
$bad_words_sw $bad_words_rules $target_files > ${bad_words_log_file}  # line 85
cat ${bad_words_log_file}                                             # line 88
result_count=$(cat ${bad_word_log_file} | grep -c '^' )               # line 91
```

Line 91 reads `${bad_word_log_file}` — no `s`. The name is unset, so `cat` reads stdin rather than the grep output. Under Actions stdin is empty, `result_count` is always `0`, and the step always exits 0.

Reproduced locally on a file that does contain a listed word — grep prints the hit and the script still passes:

```
README.md:127:- [TAOS-CI config files for nnstreamer](.TAOS-CI).
There is no prohibited word found in this PR.
exit=0
```

Counting with `wc -l` fixes the name and removes the useless `cat` that let the failure hide behind a stdin read.

## Why the word list has to change at the same time

Repairing the count alone would turn the step red on nearly every PR. `prohibited-words.txt` listed `file`, and the rules are a plain `grep -f`:

- **337 of 392** files in the inspected extensions (`*.c *.h *.cpp *.hpp *.py *.sh *.php *.md`) contain `file`
- The `README.md` hit above is exactly that

`file` is not a prohibited word in any meaningful sense, so it is dropped. Remaining entries: `samsung.net`, `FUCK`.

**One consequence to be aware of.** With the checker actually working, `ext/nnstreamer/tensor_filter/vivante/tensor_filter_vivante.c:47` will fail the step if a future PR touches that file — its header comment cites an internal `github.sec.samsung.net` clone URL for Tizen packaging. That is the checker doing precisely what `samsung.net` is in the list for. I left it alone because rewriting that packaging pointer needs knowledge of where that repo lives now, which this change does not have. Merging this PR does not break anything today: the checker only inspects files changed in a PR.

## Regression test

`test_prohibited_words.sh` follows `.github/actions/check-rebuild/test_check_if_rebuild_requires.sh`, the self-test `static.check.yml` already runs, and is wired in next to it. It pins both properties that were broken:

| Property | Guards against |
| --- | --- |
| Every word in `prohibited-words.txt` must fail the checker | the counting bug fixed here |
| Ordinary prose must pass | an over-broad entry like `file` being re-added |

Both directions were verified by reintroducing each defect:

```
# with the misspelled variable restored
FAIL: prohibited word samsung.net expected exit 1, got 0
FAIL: prohibited word FUCK expected exit 1, got 0

# with "file" put back in the word list
FAIL: plain prose expected exit 0, got 1
FAIL: code-ish prose expected exit 0, got 1
```

Because the test drives the checker through the same entry point CI uses, a future PR that breaks either property fails `Static checkers and verifiers` and cannot merge.

The checker is invoked with stdin closed, matching how GitHub runs it. Without that, restoring the `cat`-reads-stdin bug makes the test hang on a terminal instead of failing — the test would still not pass, but it would waste a CI slot rather than report.

## Local verification

- `test_prohibited_words.sh` — all checks pass; both regression injections fail it as shown above
- Repaired checker run against this PR's own changed files — exit 0
- `executable.sh` on the changed files — pass; new script committed `100644` like its siblings
- Staged blobs confirmed LF (0 CR bytes), `static.check.yml` re-parsed and the new step confirmed registered in `simple_script_checkers`

## Scope

Found while working on #4878, where `prohibited-words.txt` looked like the natural place to guard against dead JCenter/bintray links coming back. It could not serve that purpose, hence this repair. The two PRs are independent and touch disjoint files — this one deliberately does **not** add `bintray`/`jcenter` to the list, because doing so before #4878 lands would fail any PR touching the docs that still mention them.

### DIFF
--- a/.github/workflows/static.check.scripts/prohibited-words.sh
+++ b/.github/workflows/static.check.scripts/prohibited-words.sh
@@ -88,7 +88,7 @@ if [[ -n "${target_files/[ ]*\n/}" ]]; then
   cat ${bad_words_log_file}
 
   # Step 3: Count prohibited words from variable result_content
-  result_count=$(cat ${bad_word_log_file} | grep -c '^' )
+  result_count=$(wc -l < ${bad_words_log_file})
 
   # Step 4: change a value of the check result
   if [[ $result_count -gt 0 ]]; then
--- a/.github/workflows/static.check.scripts/prohibited-words.txt
+++ b/.github/workflows/static.check.scripts/prohibited-words.txt
@@ -1,3 +1,2 @@
 samsung.net
 FUCK
-file
--- a/.github/workflows/static.check.scripts/test_prohibited_words.sh
+++ b/.github/workflows/static.check.scripts/test_prohibited_words.sh
@@ -3,7 +3,7 @@
 ##
 # Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
 #
-# @file: test_prohibited_words.sh
+# @file     test_prohibited_words.sh
 # @brief    Self-test for prohibited-words.sh and its word list.
 # @see      https://github.com/nnstreamer/nnstreamer
 # @author   MyungJoo Ham <myungjoo.ham@samsung.com>
@@ -53,8 +53,9 @@ expect_checker() {
   fi
 }
 
-# Every listed word must actually fail the checker.
-while read -r word; do
+# Every listed word must actually fail the checker. The || [[ -n ]] guard keeps
+# a last line without a trailing newline from being skipped silently.
+while read -r word || [[ -n "$word" ]]; do
   word=${word%$'\r'}
   [[ -z "$word" ]] && continue
   expect_checker 1 "prohibited word ${word}" "A line that mentions ${word} inline."
--- a/.github/workflows/static.check.scripts/test_prohibited_words.sh
+++ b/.github/workflows/static.check.scripts/test_prohibited_words.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file: test_prohibited_words.sh
+# @brief    Self-test for prohibited-words.sh and its word list.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Runs prohibited-words.sh the same way static.check.yml does, from the
+# repository root with a changed-file list. Two properties are pinned: a
+# prohibited word must fail the checker, and ordinary prose must not. The
+# first would have caught the checker silently passing on every input; the
+# second keeps an over-broad entry from turning the word list into a gate
+# that rejects nearly every PR.
+#
+# The checker is invoked with stdin closed, matching how GitHub runs it, so
+# that a future reader-of-stdin bug fails here instead of hanging.
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+REPO_ROOT=$(cd "${SCRIPT_DIR}/../../.." && pwd)
+CHECKER="${SCRIPT_DIR}/prohibited-words.sh"
+WORD_LIST="${SCRIPT_DIR}/prohibited-words.txt"
+workdir=$(mktemp -d)
+failed=0
+
+trap 'rm -rf "$workdir"' EXIT
+
+# expect_checker <expected 0|1> <description> <content>
+# Writes content to a markdown file, runs the checker on it from the repo
+# root, and compares the exit status.
+expect_checker() {
+  local expected=$1 desc=$2 content=$3
+  local target list actual
+
+  target="${workdir}/$(echo "$desc" | tr -c 'a-zA-Z0-9' '_').md"
+  printf '%s\n' "$content" > "$target"
+  list=$(mktemp -p "$workdir")
+  printf '%s\n' "$target" > "$list"
+
+  (cd "$REPO_ROOT" && bash "$CHECKER" "$list") > /dev/null 2>&1 < /dev/null
+  actual=$?
+  [[ $actual -ne 0 ]] && actual=1
+
+  if [[ "$actual" == "$expected" ]]; then
+    echo "PASS: ${desc} (exit ${actual})"
+  else
+    echo "FAIL: ${desc} expected exit ${expected}, got ${actual}"
+    failed=1
+  fi
+}
+
+# Every listed word must actually fail the checker.
+while read -r word; do
+  word=${word%$'\r'}
+  [[ -z "$word" ]] && continue
+  expect_checker 1 "prohibited word ${word}" "A line that mentions ${word} inline."
+done < "$WORD_LIST"
+
+# Ordinary prose must pass. These sentences use the vocabulary that a
+# repository-wide word list is most likely to over-match on.
+expect_checker 0 "plain prose" "This file describes how to build and test the plugin."
+expect_checker 0 "code-ish prose" "Open the file, read the header, then close the file descriptor."
+
+# An empty changed-file list is a no-op, not a failure.
+empty_list=$(mktemp -p "$workdir")
+: > "$empty_list"
+if (cd "$REPO_ROOT" && bash "$CHECKER" "$empty_list") > /dev/null 2>&1 < /dev/null; then
+  echo "PASS: empty file list"
+else
+  echo "FAIL: empty file list expected exit 0"
+  failed=1
+fi
+
+if [[ "$failed" != "0" ]]; then
+  echo "::error test_prohibited_words.sh failed."
+  exit 1
+fi
+echo "test_prohibited_words.sh: all checks passed."
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -162,6 +162,19 @@ jobs:
           else
             echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
           fi
+      - name: /Checker/ prohibited-words self-test
+        # Runs a repository script unrelated to $changed_file_list, so it
+        # needs the same merge-ref guard as the step above.
+        run: |
+          selftest=.github/workflows/static.check.scripts/test_prohibited_words.sh
+          if [ -f "$selftest" ]; then
+            bash "$selftest"
+          elif git show --pretty="format:" --name-only --no-renames --diff-filter=D ${{ github.event.pull_request.head.sha }} -${{ github.event.pull_request.commits }} | grep -qxF "$selftest"; then
+            echo "::error::This PR removes $selftest; remove its workflow step too, or restore the script (also update this step if the script moved)."
+            exit 1
+          else
+            echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
+          fi
       - name: /Checker/ shellcheck for shell scripts
         # Originally from "pr-prebuild-shellcheck"
         run: |


myungjoo:
> **Note:** This review was generated by a separate automated review agent (not the PR author or a human reviewer), on behalf of @myungjoo. It is being posted verbatim as requested.

## Summary: **Approve**, with two low-severity nits (no blockers)

I reproduced/tested each claim in the PR description rather than taking it at face value, by actually re-injecting both fixed bugs and running the new self-test, and by checking the repo's branch-protection config. Details below.

### Confirms the described bug and fix
- Reintroduced the original `bad_word_log_file` (missing `s`) typo in a scratch copy of `prohibited-words.sh` and re-ran the new `test_prohibited_words.sh`: it correctly **fails** ("FAIL: prohibited word samsung.net expected exit 1, got 0" etc.) instead of the checker silently passing, confirming the self-test actually pins the original defect.
- Reintroduced `file` into `prohibited-words.txt`: the self-test correctly **fails** on the "plain prose" / "code-ish prose" cases, confirming the second property (no over-broad word) is pinned too.
- Restored both files afterward; working tree is clean.

### `wc -l < ${bad_words_log_file}` vs. the original `grep -c '^'` intent
- Empty grep output (no matches) → `wc -l` gives `0`, matching the intended "no prohibited words" semantics.
- Tested the "last line lacks trailing newline" edge case directly against the real `grep --color -n -r -H -f ...` invocation used by the script: GNU grep (3.0 here, and the `apt-get install ... grep` package on `ubuntu-latest`) always terminates each matched output line with `\n`, even when the corresponding line in the *source* file has no trailing newline. So `wc -l` and `grep -c '^'` agree in every case I could construct — this is not a live bug, just something worth knowing if the runner's grep implementation ever changed (it won't, on GitHub's Ubuntu runners).
- Filenames-with-spaces: the `for file in `cat $files`` word-splitting in this script would already mis-handle a changed file path containing spaces, but that predates this PR and is unrelated to the `wc -l` change — not introduced or worsened here, so out of scope for this diff.

### Word-list change (`file` removed)
- Recounted the "how many files contain the word `file`" claim independently: 340/395 tracked `*.c *.h *.cpp *.hpp *.py *.sh *.php *.md` files in this checkout contain "file" (case-insensitive), close to the PR's stated 337/392 — confirms `file` was making the checker fail almost every PR and had to go.
- Verified `ext/nnstreamer/tensor_filter/vivante/tensor_filter_vivante.c:47` does contain a literal `https://github.sec.samsung.net/...` URL, so the PR's called-out consequence (a future PR touching that file will trip the `samsung.net` entry) is accurate, and deferring it (since the checker only scans files changed *in* a PR, and this PR doesn't touch that file) is a reasonable, clearly-documented call rather than scope creep.
- Confirmed this PR's own diff does not trip its own newly-repaired checker: `samsung.net`/`FUCK` appear literally only inside `prohibited-words.txt` itself, and `.txt` is not one of the extensions the checker scans (`*.c *.h *.cpp *.hpp *.py *.sh *.php *.md`), so the word list is never checked against itself. No non-`.txt` file in this diff contains either literal string.

### Self-test robustness (hang/flake potential)
- `test_prohibited_words.sh` invokes the checker with `< /dev/null` on every call, so the very stdin-reads-instead-of-grep-output bug this PR fixes would, if reintroduced, fail loudly rather than hang a CI runner — confirmed this by testing (FAIL output appeared immediately, no hang).
- `mktemp -d` / `mktemp -p "$workdir"` are standard GNU coreutils usage and match what's on `ubuntu-latest`.
- `trap 'rm -rf "$workdir"' EXIT` cleans up on both success and the `exit 1` failure path.
- `set -u` doesn't appear to trip on anything — all variables are assigned before use, and the `local expected=$1 desc=$2 content=$3` multi-assignment form is valid bash.

### Workflow wiring
- `static.check.yml` has a single job, `simple_script_checkers` (display name **"Static checks"**). I confirmed via `gh api repos/nnstreamer/nnstreamer/branches/main/protection` that **"Static checks" is a required status check** on `main`. So a future regression that this self-test catches will genuinely block merge, not just show a red X that can be ignored — this satisfies the "CI fail blocks the offending PR" requirement.
- The new step is placed right after the analogous `/Checker/ check-rebuild action self-test` step and before the file-list-driven checkers resume (`shellcheck`, etc.) — a sensible grouping with the workflow's other self-test, and it doesn't depend on `$changed_file_list` so its position relative to that variable's producer step doesn't matter.

### Scope
Diff is minimal and tightly scoped (86+/2- across exactly the 3 files the fix requires: the checker, the word list, and a new sibling self-test + one workflow line). No unrelated modules touched. No architecture/API change, so no documentation update is needed for this PR.

---

### Nits (non-blocking)

1. **Low** — In `test_prohibited_words.sh`, the loop that verifies every listed word fails the checker:
   ```sh
   while read -r word; do
     ...
   done < "$WORD_LIST"
   ```
   Bash's `read` returns non-zero on an unterminated final line, so if `prohibited-words.txt` is ever edited (e.g. by an editor that strips the trailing newline) such that its **last** line loses its trailing `\n`, that last word will be silently skipped by this loop — the self-test will still print "all checks passed" without ever having tested the newest entry. I verified this concretely: adding a new word as the file's last line with no trailing newline caused the self-test to pass without ever exercising that word. Currently harmless (the checked-in `prohibited-words.txt` does end with `\n`), but it's a silent-coverage-gap landmine for the next person who edits the list. Suggest the standard guard: `while read -r word || [[ -n "$word" ]]; do ... done < "$WORD_LIST"`.

2. **Nit / cosmetic** — The new file's header uses `@file: test_prohibited_words.sh` (colon after `@file`), while every sibling script in `.github/workflows/static.check.scripts/` uses `@file     name.sh` (no colon). I checked `Doxyfile.prj`: it does not set `WARN_AS_ERROR`, so even if this produces a doxygen argument-mismatch warning it won't fail the `doxygen-build.sh` checker (warnings don't become a nonzero exit without that flag) — this is purely a style inconsistency, not a functional or CI risk.

Neither nit blocks approval; both are easy one-line tweaks if the author wants to address them.

myungjoo:
Both review points are addressed in 388d0073; neither needed to become a follow-up.

**Low — unterminated last line skipped.** Confirmed, and worth more than its severity suggested: appending a sentinel word without a trailing newline made the suite print `all checks passed` while never testing that word. A test that silently tests nothing is the exact failure mode this PR exists to remove, so it would have been wrong to leave it in. Fixed with `|| [[ -n "$word" ]]`.

Verified after the change:

```
# word list ending without a newline -- the sentinel is now exercised
PASS: prohibited word samsung.net (exit 1)
PASS: prohibited word FUCK (exit 1)
PASS: prohibited word zzsentinelzz (exit 1)
```

An empty word list was also checked, since `[[ -n "$word" ]]` under `set -u` would break if `read` left the variable unset at EOF. It does not — bash assigns an empty value — and the suite runs clean.

Both original regression injections still fail the test after the change: restoring the misspelled variable gives `expected exit 1, got 0` on the word cases, and re-adding `file` gives `expected exit 0, got 1` on the prose cases.

**Nit — `@file:` colon.** Fixed to `@file     test_prohibited_words.sh`, matching the sibling scripts.

Thanks for independently re-deriving the 337/392 figure and for confirming via the branch-protection API that `Static checks` is a required check on `main` — that was the part I could not verify myself, and it is what makes the self-test actually able to block a regression rather than just report one.

myungjoo:
> **Note:** This is a re-review generated by a separate automated review agent (not the PR author or a human reviewer), on behalf of @myungjoo, following up on [the earlier review](https://github.com/nnstreamer/nnstreamer/pull/4880#issuecomment-5433531882) and the fix in 388d0073. It is being posted verbatim as requested.

## Summary: **Approve**

Both previously-reported nits are verified fixed by direct reproduction (not just reading the diff), the fix introduces no new problem under `set -u` in the cases I could construct, and both original regression-detection properties still work after the change. Details below.

### 1. Trailing-newline / unterminated-last-word fix — confirmed

Reproduced the exact scenario from the original nit: temporarily wrote `prohibited-words.txt` with a sentinel word (`zzsentinelzz`) as the **last line with no trailing newline**, then ran `test_prohibited_words.sh`.

- Before 388d0073 (loop without the guard), this silently skips the sentinel and still prints `all checks passed` (exit 0) — I confirmed this by temporarily reverting the guard in a scratch edit and restoring it afterward.
- After 388d0073, the guarded loop (`while read -r word || [[ -n "$word" ]]`) correctly exercises the sentinel: `PASS: prohibited word zzsentinelzz (exit 1)`.

Working tree was restored with `git checkout --` after each experiment; `git status` is clean.

### 2. `@file` colon style — confirmed

`test_prohibited_words.sh` now reads `@file     test_prohibited_words.sh` (no colon), matching every other script in `.github/workflows/static.check.scripts/` (`cppcheck.sh`, `doxygen-build.sh`, `executable.sh`, `flawfinder.sh`, `hardcoded-path.sh`, `misspelling.sh`, `nobody.sh`, `prohibited-words.sh`, `rpm-spec.sh`, `shellcheck.sh`, `signed-off-by.sh`, `sloccount.sh`, `timestamp.sh`). (Aside: `.github/actions/check-rebuild/*.sh` uses `@file:` with a colon, but that's a different directory the new file doesn't share a listing with — matching its immediate siblings, which it does, is the right target.)

### 3. `|| [[ -n "$word" ]]` guard under `set -u` — no new problems found

Tested directly against the running script (not just reasoning about it):

- **Empty word list** (0-byte file): loop body never executes, no unbound-variable error, script proceeds to the prose/empty-list checks and exits 0. Confirmed both against the real script and in an isolated `set -u` snippet — `read -r word || [[ -n "$word" ]]` never dereferences `$word` before `read` has had a chance to assign it (even an empty assignment), so this is safe.
- **Whitespace-only line** (`"   "` as the entire line): bash's `read` with a single variable target strips leading/trailing IFS whitespace, so the line collapses to an empty string and is caught by the existing `[[ -z "$word" ]] && continue` guard, independent of the new `|| [[ -n ]]` addition. No new gap.
- **CRLF-terminated word list** (every line ending `\r\n`, including the last): works correctly end-to-end — the `${word%$'\r'}` strip handles it, all three words are exercised.
- **Pathological case found, out of scope for this PR**: if the word list's *last* line ends in a bare `\r` with **no trailing `\n` at all** (a mixed/corrupted line-ending state, not something the repo's LF-formatted `prohibited-words.txt` is in or that any normal editor produces), GNU grep 3.0's `-f` pattern-file reader takes that trailing `\r` as a literal character in the pattern (it only strips `\r` from patterns that are properly `\n`-terminated). That makes the real checker fail to match the word, and the self-test correctly reports `FAIL: ... expected exit 1, got 0` rather than passing silently — i.e., the self-test's own defense-in-depth catches this pathology instead of hiding it. This is pre-existing `grep -f` behavior in `prohibited-words.sh`'s Step 1 (untouched by this diff) and not something 388d0073 (or this PR) introduces or needs to fix; flagging only as an FYI, not a blocker.

All scratch modifications were reverted with `git checkout --`; the worktree is clean.

### 4. Regression-detection still works — confirmed

Re-ran both original regression injections against the post-388d0073 code:

- Restored the original `bad_word_log_file` (missing `s`) typo in a scratch copy of `prohibited-words.sh` → `test_prohibited_words.sh` correctly fails: `FAIL: prohibited word samsung.net expected exit 1, got 0` / same for `FUCK`.
- Re-added `file` to `prohibited-words.txt` → correctly fails: `FAIL: plain prose expected exit 0, got 1` / `FAIL: code-ish prose expected exit 0, got 1`.

Both match the exact failure signatures shown in the PR description. Neither property regressed as a side effect of the 388d0073 fix.

### Re-check of the broader PR criteria

- **Solves the stated problem**: yes — the checker's silent-always-pass bug is fixed, the over-broad `file` entry is removed, and the fix is now pinned by a self-test wired into `static.check.yml`.
- **No regression to other modules**: the diff touches only `.github/workflows/static.check.scripts/{prohibited-words.sh,prohibited-words.txt,test_prohibited_words.sh}` and `.github/workflows/static.check.yml`. No runtime/library code touched; nothing else to regress.
- **Scope/size proportionate**: 4 files, ~90 lines, all directly required for the fix + its regression test. Not oversized, no unrelated drive-by changes.
- **Future-regression coverage / CI enforcement**: `test_prohibited_words.sh` is wired into the `simple_script_checkers` job in `static.check.yml`, which the earlier review confirmed (via the branch-protection API) is a required status check on `main` — a future regression on either property will genuinely block merge.
- **Docs**: no ARCHITECTURE/API surface changed; no documentation update needed.

No blocking issues found. Approving.

myungjoo:
Rebased onto current `main` (638b53c7). This was not routine housekeeping — the PR had gone `CONFLICTING`, and the conflict turned up a real defect in what I had written.

**Why no CI ran on the previous head.** GitHub does not schedule `pull_request` workflows when the merge commit cannot be computed, so once #4868's follow-ups landed on `main` and touched the same region of `static.check.yml`, the push to 388d0073 produced zero workflow runs — only DCO responded. I spent a while treating that as more of the runner-starvation flakiness from earlier today; the actual cause was the conflict.

**The defect the conflict exposed.** `main` now guards the check-rebuild self-test step, and its comment states the rule directly:

> any future step that runs a repository script unrelated to `$changed_file_list` needs the same guard

My new step is exactly that. Without the guard it would have broken **every currently open PR**: the step list comes from the merge ref while the checkout is the PR head, so any branch predating `test_prohibited_words.sh` would run a step referencing a script its tree does not contain and fail `Static checks` with "No such file or directory". The step now carries the same guard, and the reason is recorded in the commit message rather than only here.

That is a multi-line change made after the approving review, so this needs another look rather than riding on the previous approval — re-review requested.

Nothing else changed. `prohibited-words.sh`, `prohibited-words.txt` and `test_prohibited_words.sh` are byte-identical to the reviewed versions; the diff against `main` is still the same four files. The self-test passes locally after the rebase, and the workflow YAML re-parses with the new step registered in `simple_script_checkers`.

For the record, CI on the pre-rebase commit a1678d16 was fully green — all 12 workflows, including `Static checkers and verifiers`, where both `/Checker/ prohibited words` and `/Checker/ prohibited-words self-test` ran and passed.

myungjoo:
> **Note:** This is a third-round re-review generated by a separate automated review agent (not the PR author or a human reviewer), on behalf of @myungjoo, triggered by the post-approval rebase onto `main` (head now `638b53c7`) and the new merge-ref guard it introduced. It is being posted verbatim as requested.

## Summary: **Approve**

The rebase did not touch the reviewed logic, and the new guard is correct, necessary, and consistent with precedent already in `main`. Details below (all claims reproduced directly, not taken from the diff or the PR description).

### 1. The new merge-ref guard is a byte-for-byte pattern match

Diffed the new `/Checker/ prohibited-words self-test` step against the existing `/Checker/ check-rebuild action self-test` step in `static.check.yml` (638b53c7): identical `if [ -f "$selftest" ] / elif git show --pretty="format:" --name-only --no-renames --diff-filter=D ... | grep -qxF "$selftest" / else` structure, same flags (`--no-renames` to decompose renames into D+A, `--diff-filter=D`, `grep -qxF`), same error/skip message templates. Only the `selftest=` path and the shortened explanatory comment (which correctly points back to "the same merge-ref guard as the step above" instead of duplicating the full rationale) differ. `$selftest` is set to `.github/workflows/static.check.scripts/test_prohibited_words.sh`, which matches the actual repository path exactly.

### 2. The "would have broken every open PR" claim is verified, not just plausible

- Confirmed in `static.check.yml`'s checkout step: `ref: ${{ github.event.pull_request.head.sha }}` — the job checks out the PR **head**, not the merge ref.
- Confirmed this is a real, previously-materialized failure mode in this repo's own history, not a hypothetical: commit `3a7ff6ec` ("[CI] Guard the check-rebuild self-test against stale PR branches") documents that when `main` gained the check-rebuild self-test (`b57d9654`) without a guard, "every PR branched before it started failing the Static checks job with exit 127 ('No such file or directory'), as seen on nnstreamer/nnstreamer#4863" — because the pull_request event's step list is resolved from the merge ref while the checkout is the PR head. That guard was then further hardened over three follow-up commits (`791ef6f3` rename/deletion distinction, `5ba765f0` `--no-renames` fix, `22488067` comment correction) — the exact final form the new step now copies. Given this precedent, the author's claim that an unguarded new step would break all currently-open PRs the moment this PR merges is well-founded.

### 3. This PR's own run takes the correct branch

`test_prohibited_words.sh` is present in the PR head tree at `.github/workflows/static.check.scripts/test_prohibited_words.sh`, so `[ -f "$selftest" ]` is true and the `if` branch (`bash "$selftest"`) runs — the `elif`/`else` never need to trigger for this PR's own diff. Ran the self-test directly against the checked-out head: `PASS` on all 5 cases (`samsung.net`, `FUCK`, plain prose, code-ish prose, empty file list), exit 0.

### 4. Rebase integrity — byte-identical, confirmed by hash, not just `git diff`

The PR's pre-rebase history had a1678d16 ("Repair the prohibited-words checker...", the commit actually CI-tested green) as parent of 388d0073 ("Do not skip an unterminated last word...", the nit-fix commit the second review approved). After rebase these became `bd75be42` and `638b53c7` respectively — same patches, new parent, new hashes, which is expected and not a discrepancy.
- `git diff 388d0073 638b53c7 -- prohibited-words.sh prohibited-words.txt test_prohibited_words.sh` → empty.
- MD5 of all three files at `388d0073` vs `638b53c7`: identical (`d54677e3...`, `e890c4fe...`, `8494609d...` respectively, matching on both sides).

So the byte-identical claim holds against the version that actually received the second-round approval. (Note for the record: the "CI fully green" reference in the author's rebase comment is to `a1678d16`, the commit *before* the nit fixes — accurate as stated, since the push of `388d0073` never triggered a workflow run due to the conflict, but worth flagging so no one mistakes it for CI having validated the final nit-fixed content. Not a blocker; the byte-identical diff above is the stronger and sufficient evidence.)

### 5. Prior findings still hold post-rebase

Since the three content files are byte-identical to the twice-reviewed version, all previously-verified properties carry over unchanged and were spot-re-confirmed:
- Regression detection still works: reintroducing the `bad_word_log_file` typo and re-adding `file` to the word list both correctly fail the self-test (same failure signatures as before).
- The `|| [[ -n "$word" ]]` unterminated-line guard and the `@file` header fix from `388d0073` are both present at head.
- `Static checks` remains the job name / required status check (unchanged — only a step was appended, job name and structure untouched).

### 6. `static.check.yml` YAML validity and placement

Parsed with `yaml.safe_load`: valid. The new step sits inside the same `simple_script_checkers` job, immediately after `/Checker/ check-rebuild action self-test` and before `/Checker/ shellcheck for shell scripts` — grouping the two self-tests together, after the real `/Checker/ prohibited words` step, and before the file-list-driven checkers resume. Position is cosmetic here since the step doesn't depend on `$changed_file_list`, but it's the sensible spot.

### Verification hygiene

All checks were read-only except one self-test execution (no file mutation); `git status` was clean throughout and remains clean. No files were left modified in the worktree.

No blocking issues found. Approving.

PR #4881, opened 2026-08-27
[Elements] use util to get running time

jaeyun-jung:
Code clean, use util function to handle the running time and cleanup codes.

### DIFF
--- a/ext/nnstreamer/android_source/gstamcsrc.c
+++ b/ext/nnstreamer/android_source/gstamcsrc.c
@@ -243,8 +243,9 @@ G_DEFINE_BOXED_TYPE (GstWrappedBuf, gst_wrapped_buf,
 
 /**
  * @brief get system's nanotime
+ * @todo consider using g_get_monotonic_time_ns() (glib >= 2.88)
  */
-static int64_t
+static gint64
 systemnanotime (void)
 {
   struct timespec now;
@@ -642,26 +643,18 @@ feed_frame_buf (GstAMCSrc * self, guint8 * buf, gint idx, gsize real_size,
     gsize buf_size)
 {
   GstAMCSrcPrivate *priv;
-  GstElement *element;
   GstBuffer *buffer;
   GstMemory *mem;
   GstDataQueueItem *item;
   GstClockTime duration = GST_CLOCK_TIME_NONE;
   GstClockTime current_ts = GST_CLOCK_TIME_NONE;
-  GstClock *clock;
   GstWrappedBuf *wrapped_buf;
 
   g_return_if_fail (self != NULL);
   g_return_if_fail (buf != NULL);
 
   priv = GST_AMC_SRC_GET_PRIVATE (self);
-  element = GST_ELEMENT (self);
-
-  if ((clock = gst_element_get_clock (element))) {
-    current_ts =
-        gst_clock_get_time (clock) - gst_element_get_base_time (element);
-    gst_object_unref (clock);
-  }
+  current_ts = gst_element_get_current_running_time (GST_ELEMENT (self));
 
   g_mutex_lock (&priv->mutex);
 
@@ -687,7 +680,7 @@ feed_frame_buf (GstAMCSrc * self, guint8 * buf, gint idx, gsize real_size,
   wrapped_buf = g_new0 (GstWrappedBuf, 1);
   g_assert (wrapped_buf != NULL);
   wrapped_buf->refcount = 1;
-  wrapped_buf->amcsrc = g_object_ref (self);
+  wrapped_buf->amcsrc = gst_object_ref (self);
   wrapped_buf->buf = buf;
   wrapped_buf->idx = idx;
 
@@ -723,7 +716,7 @@ static void
 check_codec_buf (GstAMCSrc * self)
 {
   GstAMCSrcPrivate *priv = GST_AMC_SRC_GET_PRIVATE (self);
-  gint64 delay, presentation_time;
+  gint64 delay, presentation_time, time_ns;
   gssize buf_idx = -1;
   gsize buf_size;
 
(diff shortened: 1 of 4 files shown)


myungjoo (commented):
> _This review was produced by a separate Claude agent and is relayed here by the maintainer's account. Please verify before acting on it._

## Third review at head `a1f353a5` (unchanged since the previous two reviews)

The code has not changed since `7ab9a570`. What has changed is `main`, which now conflicts with this PR, and the review criteria, which are narrower this round.

### Scope ruling: correct incidental fixes are fine

The PR describes itself as "code clean" but also fixes a few small bugs along the way. **For this PR, going beyond pure cleanup is not a reason to hold it, as long as each fix is correct.** Both earlier reviews raised the description mismatch; that note is withdrawn, and no separate description or tests are required for the incidental fixes. I rechecked each one and all are correct:

- `mqttsink.c` debug path: the old code called `gst_clock_get_time()` and `gst_object_unref()` on a possibly-NULL clock. Gone.
- `mqttsrc.c` PAUSED→PLAYING: the old `if (!elem_clock) break;` also skipped the re-subscribe block. That block now always runs.
- `gstamcsrc.c`: taking a single `time_ns` sample keeps `renderstart` and `delay` consistent; `usleep` becomes `g_usleep`; `g_object_ref` becomes `gst_object_ref`. The last was functionally the same before, but now pairs with the unref.

The three items below are different. They are not fixes. They are regressions introduced by this PR, so the ruling does not cover them, and they remain blocking.

### 1. Blocking: `mqttsrc` latency uses an invalid `buf_ts`

[mqttsrc.c:750-760](https://github.com/nnstreamer/nnstreamer/blob/a1f353a5e1497962b086ee2d123b49b89ebee36a/gst/mqtt/mqttsrc.c#L750-L760) no longer checks `buf_ts` before `GST_CLOCK_DIFF (buf_ts, cur)`. The gate above it, [`_is_gst_buffer_timestamp_valid()`](https://github.com/nnstreamer/nnstreamer/blob/a1f353a5e1497962b086ee2d123b49b89ebee36a/gst/mqtt/mqttsrc.c#L161-L167), passes when any one of PTS, DTS or duration is valid. [`_put_timestamp_on_gst_buf()`](https://github.com/nnstreamer/nnstreamer/blob/a1f353a5e1497962b086ee2d123b49b89ebee36a/gst/mqtt/mqttsrc.c#L1451-L1459) sets `pts` only when `hdr->pts` is valid but always copies `duration`. So a PTS-less, duration-carrying buffer reaches this code. For it, `diff` comes out as `cur + 1`, and `GST_QUERY_LATENCY` reports roughly the element's running time as `min_latency`. Please wrap the computation in `if (GST_CLOCK_TIME_IS_VALID (buf_ts))`.

### 2. Blocking: duration-less buffers report no latency

Same block. With the duration unset, `duration == GST_CLOCK_TIME_NONE`, so `duration < (GstClockTime) diff` is always false and `latency` stays `NONE`. The old code did the opposite and kept the measured latency when there was no valid duration. Suggested:

```c
if (diff > 0 && (!GST_BUFFER_DURATION_IS_VALID (*buf) || duration < (GstClockTime) diff))
```

### 3. Blocking (new): the converter hunk conflicts with #4898 and would undo it

`mergeable` is now `CONFLICTING`. The only conflict is in `gsttensor_converter.c`, against `5bad7dc7` "[Converter] Take the running time only while PLAYING" (fixes #4898, merged 2026-09-04).

The PR's version cannot simply be carried over on top of that fix. `gst_element_get_current_running_time()` ([gstelement.c](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gstreamer/gst/gstelement.c)) does not check the element's state:

- `base_time == 0` counts as valid. Before the converter receives its base time, it therefore returns the absolute clock value. That is exactly the #4898 bug, where the timestamp became the system uptime and a sync sink stalled.
- It also drops `main`'s fallback outside PLAYING, which reuses `old_timestamp` and otherwise keeps `segment.start`.

`unittest_converter_timestamp` on `main` pins this behaviour. For example, [`pausedBeforeFirstPlaying_p`](https://github.com/nnstreamer/nnstreamer/blob/192b4ddf9d9768e8f9f44691e0dd902bdeabf16d/tests/nnstreamer_converter/unittest_converter_timestamp.cc#L139-L159) expects 0 while PAUSED with the clock at one hour and base time 0, and the PR's code returns one hour. [`pausedUsesSegmentStart_p`](https://github.com/nnstreamer/nnstreamer/blob/192b4ddf9d9768e8f9f44691e0dd902bdeabf16d/tests/nnstreamer_converter/unittest_converter_timestamp.cc#L190-L210) expects `segment.start` where the PR's code returns one hour.

**Resolution: when rebasing, drop this PR's `gsttensor_converter.c` hunk entirely and keep `main`'s version** ([L819-L841](https://github.com/nnstreamer/nnstreamer/blob/192b4ddf9d9768e8f9f44691e0dd902bdeabf16d/gst/nnstreamer/elements/gsttensor_converter.c#L819-L841)). `main` already reads the clock and base time under one object lock and checks `GST_STATE == PLAYING`, which the utility function cannot do. This also settles the earlier review's item 3 (the lost `segment.start` fallback).

### Recommended, not blocking

- A latency test in `tests/gstreamer_mqtt/` covering one buffer with valid PTS and duration and one without PTS, then asserting `min_latency`. This would keep items 1 and 2 from returning in the next rewrite of this code.
- `gst_element_get_current_running_time()` requires GStreamer 1.18 or newer and is used without a `GST_CHECK_VERSION` guard. Every target built today is new enough; this is only a question of whether that is a deliberate choice.
- [gstamcsrc.c:52](https://github.com/nnstreamer/nnstreamer/blob/a1f353a5e1497962b086ee2d123b49b89ebee36a/ext/nnstreamer/android_source/gstamcsrc.c#L52) still includes `<unistd.h>`, which was only needed for `usleep`.

### CI

The two red checks at this head are unrelated: Android NDK failed on a Maven HTTP 429, and Valgrind ubuntu-22.04 failed on `unittest_filter_llamacpp`, whose intermittent stall `5bad7dc7` attributes to #4898. After a rebase, both should be re-run.

**Verdict: changes requested** on items 1, 2 and 3. The rest of the PR, the incidental fixes included, is fine as it stands.

PR #4882, opened 2026-08-27
[Filter/TVM] Fix model_path leak on configure_instance failure paths

myungjoo:
`tvm_subplugin::cleanup()` starts with `if (empty_model) return;`, but `configure_instance()` assigns `model_path = g_strdup (prop->model_files[0])` *before* the calls that can throw, and `empty_model` is only cleared at the very end of a successful configure. Every throw in between calls `cleanup()`, hits the early return, and leaks the duplicated path.

The affected throw sites, all after the `g_strdup`:
- `tvm::runtime::Module::LoadFromFile()` on a file that exists but is not a tvm module
- the `get_input == nullptr` and `get_output == nullptr` guards
- either `convert_dtype()` failure

- Move `g_free (model_path); model_path = nullptr;` above the `empty_model` guard in `cleanup()`, so the path is released regardless of how far configure got. `model_path` is initialised to `nullptr` in the constructor and reset here, so the now-unconditional `g_free` is safe for the close-before-open and double-close calls the existing tests make.
- Add `openClose01_n`, opening a file that exists but is not a tvm module (`test_models/labels/labels.txt`). The existing negative cases (`openClose00_n`, `launch00_n`) pass a nonexistent `temp.so` and are rejected earlier by the `g_file_test (…, G_FILE_TEST_IS_REGULAR)` check, so no current test reaches the leaking path at all.

Found while reviewing #4873, which adds an SSAT suite whose `3_n` case is the first test in the tree to exercise `LoadFromFile()` throwing. The two PRs are independent and touch disjoint files. This one is not required for #4873 to be correct — the leak is pre-existing and one `g_strdup` per failed configure — but it is worth fixing before the path starts being exercised regularly.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [*]Skipped (needs a tvm-enabled build; verified in CI)
2. Run test: [ ]Passed [ ]Failed [*]Skipped (same)

**How to evaluate:**
1. Build with `-Dtvm-support=enabled`.
2. Run `unittest_filter_tvm`; `openClose01_n` should pass, and should fail on `sp->open()` returning 0 if the exception handling ever regresses.
3. Under valgrind or ASAN, `openClose01_n` shows a definite loss of the model path string without this fix and none with it.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc
@@ -99,16 +99,17 @@ tvm_subplugin::tvm_subplugin ()
 void
 tvm_subplugin::cleanup () noexcept
 {
+  g_free (model_path);
+  model_path = nullptr;
+
   if (empty_model)
     return;
-  g_free (model_path);
 
   input_tensor_list.clear ();
   output_tensor_list.clear ();
   gst_tensors_info_free (std::addressof (inputInfo));
   gst_tensors_info_free (std::addressof (outputInfo));
 
-  model_path = nullptr;
   empty_model = true;
 }
 
--- a/tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
+++ b/tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
@@ -39,7 +39,9 @@ class NNStreamerFilterTVMTest : public ::testing::Test
   const GstTensorFilterFramework *sp;
   const gchar *wrong_model_files[2];
   const gchar *proper_model_files[2];
+  const gchar *invalid_model_files[2];
   gchar *model_file;
+  gchar *invalid_model_file;
   gchar *pipeline;
   GstElement *gstpipe;
   GstElement *sink_handle;
@@ -47,16 +49,25 @@ class NNStreamerFilterTVMTest : public ::testing::Test
   GstTensorMemory output;
 
   /**
+   * @brief Get the absolute path of a file in tests/test_models
+   */
+  gchar *GetTestModelPath (const gchar *subdir, const gchar *filename)
+  {
+    const gchar *src_root = g_getenv ("NNSTREAMER_SOURCE_ROOT_PATH");
+    g_autofree gchar *root_path = src_root ? g_strdup (src_root) : g_get_current_dir ();
+
+    return g_build_filename (root_path, "tests", "test_models", subdir, filename, NULL);
+  }
+
+  /**
    * @brief Set the model file for testing
    */
   gchar *SetModelFile ()
   {
-    const gchar *src_root = g_getenv ("NNSTREAMER_SOURCE_ROOT_PATH");
-    g_autofree gchar *root_path = src_root ? g_strdup (src_root) : g_get_current_dir ();
     g_autofree gchar *model_name
         = g_strdup_printf ("tvm_add_one_%s%s_", ARCH, NNSTREAMER_SO_FILE_EXTENSION);
 
-    return g_build_filename (root_path, "tests", "test_models", "models", model_name, NULL);
+    return GetTestModelPath ("models", model_name);
   }
 
   public:
@@ -64,13 +75,14 @@ class NNStreamerFilterTVMTest : public ::testing::Test
    * @brief Construct a new NNStreamerFilterTVMTest object
    */
   NNStreamerFilterTVMTest ()
-      : sp (nullptr), model_file (nullptr), pipeline (nullptr),
-        gstpipe (nullptr), sink_handle (nullptr)
+      : sp (nullptr), model_file (nullptr), invalid_model_file (nullptr),
+        pipeline (nullptr), gstpipe (nullptr), sink_handle (nullptr)
   {
     input.data = output.data = nullptr;
     input.size = output.size = 0;
     wrong_model_files[0] = wrong_model_files[1] = nullptr;
     proper_model_files[0] = proper_model_files[1] = nullptr;
+    invalid_model_files[0] = invalid_model_files[1] = nullptr;
   }
 
   /**
@@ -97,6 +109,10 @@ class NNStreamerFilterTVMTest : public ::testing::Test
     proper_model_files[0] = model_file;
     proper_model_files[1] = NULL;
 
+    invalid_model_file = GetTestModelPath ("labels", "labels.txt");
+    invalid_model_files[0] = invalid_model_file;
+    invalid_model_files[1] = NULL;
+
     input.size = output.size = 0;
     input.data = nullptr;
     output.data = nullptr;
@@ -110,6 +126,7 @@ class NNStreamerFilterTVMTest : public ::testing::Test
   void TearDown () override
   {
     g_free (model_file);
+    g_free (invalid_model_file);
     g_free (pipeline);
 
     g_clear_object (&gstpipe);
@@ -184,6 +201,28 @@ TEST_F (NNStreamerFilterTVMTest, openClose00)
   sp->close (&prop, &data);
 
   /* double close */
+  sp->close (&prop, &data);
+}
+
+/**
+ * @brief Negative case with a file that exists but is not a tvm module
+ */
+TEST_F (NNStreamerFilterTVMTest, openClose01_n)
+{
+  int ret;
+  void *data = NULL;
+  GstTensorFilterProperties prop;
+
+  ASSERT_TRUE (g_file_test (invalid_model_file, G_FILE_TEST_IS_REGULAR));
+  EXPECT_NE (sp, nullptr);
+
+  /* Test */
+  SetFilterProperty (&prop, invalid_model_files);
+  ret = sp->open (&prop, &data);
+  EXPECT_NE (ret, 0);
+  EXPECT_EQ (data, nullptr);
+
+  /* close after failed open */
   sp->close (&prop, &data);
 }
 


myungjoo:

Reviewed the diff against `ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc` and `tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc`, cross-checked against `gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc` (the C++ subplugin open/close wrapper) and the analogous fix already in `tensor_filter_tensorrt10.cc`.

**Verdict: no blocking issues found. The fix is correct, minimal, and safe to merge.**

Findings:

1. **Correctness of the reorder (main change).** Moving `g_free(model_path); model_path = nullptr;` above the `if (empty_model) return;` guard is correct. `model_path` is only ever set via `g_strdup()` in `configure_instance()`, and it is initialized to `nullptr` in the constructor and reset to `nullptr` immediately after the free. Since `empty_model` stays `true` until the very end of a successful `configure_instance()`, every throw site after `model_path = g_strdup(prop->model_files[0])` (LoadFromFile, get_input/get_output null checks, both convert_dtype failures) previously hit the early return and leaked the duplicated string; now it is freed unconditionally. Confirmed this matches the exact bug described in the PR body.

2. **No double-free / use-after-free across the "double cleanup" pattern.** `tensor_filter_subplugin::cpp_open()` in `tensor_filter_support_cc.cc` wraps the freshly-constructed subplugin object in a `std::unique_ptr`; when `configure_instance()` throws, the explicit `cleanup()` call inside `configure_instance()` runs first, and then the `unique_ptr` going out of scope on the catch path invokes `~tvm_subplugin()`, which calls `cleanup()` a second time. Because the fixed code nulls `model_path` right after freeing it, this second `cleanup()` call does `g_free(nullptr)`, which is a documented no-op in glib. So the reorder is specifically what makes this double-invocation safe with respect to `model_path`; verified this is not a new risk introduced by the change.

3. **No cleanup omission introduced.** The `input_tensor_list.clear()` / `output_tensor_list.clear()` / `gst_tensors_info_free()` calls remain gated behind `if (empty_model) return;`, unchanged from before. This is pre-existing behavior (those resources are effectively no-ops when nothing was ever configured) and the PR does not alter when they run relative to `empty_model` — only `model_path`'s handling moved. No new gap.

4. **This fix follows established precedent in the codebase.** `tensor_filter_tensorrt10.cc::cleanup()` already frees `_model_path` unconditionally (guarded only by a null check, not by an `empty_model`-style flag) as part of an earlier leak fix in that subplugin. This PR brings `tensor_filter_tvm.cc` in line with that existing pattern.

5. **`openClose01_n` exercises the intended path, not a different one.** Traced the call: the test calls `sp->open()` directly (bypassing the GStreamer-property-level `verify_model_path` check in `tensor_filter_common.c`, which only applies to the `gsttensorfilter` element's property setter, not this direct subplugin call). `g_file_test(..., G_FILE_TEST_IS_REGULAR)` at line 219 of `tensor_filter_tvm.cc` passes because `tests/test_models/labels/labels.txt` is a real regular file, unlike the existing `openClose00_n`/`launch00_n` cases which use a nonexistent `temp.so` and are rejected by that same check before `model_path` is ever assigned. Execution then reaches `tvm::runtime::Module::LoadFromFile(model_path, "so")`, which will fail to `dlopen()` a plain-text file as an ELF shared object and throw — this is the exact throw site named in the PR description. The test correctly distinguishes itself from the existing negative tests.

Minor, non-blocking observations:

- The new test asserts `ret != 0` (i.e., that `open()` still correctly rejects the bad file), which is a good regression guard for the exception-handling path itself, but it does not directly assert the absence of the leak — that requires memory instrumentation (valgrind/ASan), as the PR author acknowledges was done manually. I found no sanitizer wiring (`b_sanitize`, `-fsanitize`, valgrind) in any `.github/workflows/*.yml` or the meson build files, and no CI workflow references `tvm` at all in `.github/`. This appears to be a pre-existing, project-wide gap (not something this PR introduces or is expected to fix), but it means a future regression that reintroduces this exact leak pattern would not be caught by GitHub Actions CI via this test alone — only by whatever internal/daily tvm-enabled build+valgrind process is referenced in the PR's self-evaluation section as already being used to verify this fix. Worth being aware of, not a reason to block this PR.
- No architecture or public API changes are involved (`cleanup()` is a private method internal to `tvm_subplugin`), so no documentation updates are needed for this change.
- Scope is appropriately minimal: only the two files described are touched, and the code-side change is a 3-line reorder.

myungjoo:
Thanks. The observation about sanitizer coverage is correct, and it is worse than "may not be detected" — it cannot be detected today, for two independent reasons. I checked both rather than leave it vague.

**The valgrind job does not build tvm.** `ubuntu_clean_meson_build.yml` configures with `meson setup build/ -Darmnn-support=enabled`, and `tvm-support` defaults to `auto`, so the sub-plugin is built only where `tvm-runtime-dev` is present. That job's apt steps (lines 31-38) do not install it, and `tvm` appears nowhere in `.github/workflows/`. So `unittest_filter_tvm` is never built there, and its `--valgrind` step at line 171 never sees this code.

**Even if it did, valgrind findings are non-fatal.** That same step ends with:

```
|| echo "There are Valgrind errors. Please fix it. As we have a lot of Valgrind
errors from different libraries and possible from nnstreamer itself, we are not
halting the build with Valgrind errors until we get them all."
```

So a reintroduced leak would print and pass regardless.

Closing that gap needs `tvm-runtime-dev` added to a job named "Minimal meson build" and a project-wide decision to make valgrind failures fatal — the second of which the workflow comment shows has been deliberately deferred. Neither belongs in a three-line leak fix, so I am leaving it.

What `openClose01_n` does buy, and what I claimed for it in the PR description, is narrower but real: it is the first test in the tree to reach `LoadFromFile()` throwing at all, so it pins the *behavioral* contract — a file that exists but is not a tvm module must make `open()` fail rather than crash or succeed. The leak itself is verifiable by hand under valgrind, per the evaluation steps in the description.

Noted also that `tensor_filter_tensorrt10.cc` already frees its model path unconditionally; good to know this change brings tvm in line with an existing precedent rather than inventing one.

myungjoo:
Correction to my previous reply, and an update to the branch.

**I was wrong that this cannot be detected by per-PR CI.** I only looked at `.github/workflows/*.yml`, where `tvm` indeed never appears, and stopped there. But the GBS job builds from `packaging/nnstreamer.spec`, not from the workflow's apt list:

- `nnstreamer.spec` sets `%define tvm_support 1` for the default profile, and the GBS matrix passes no `profile`, so `tvm-runtime-devel` is a `BuildRequires` and the sub-plugin is built.
- In `%check`, under `%if 0%{?unit_test}` and `%ifarch %arm x86_64 aarch64`, the spec runs `bash %{test_script} ./tests/nnstreamer_filter_tvm`.
- The matrix runs `--define "unit_test 1"` only for `x86_64` (armv7l and aarch64 use `unit_test 0`), so exactly one job runs it — "Tizen GBS build on Ubuntu (x86_64, --define \"unit_test 1\")".
- `packaging/run_unittests_binaries.sh` exports `NNSTREAMER_SOURCE_ROOT_PATH=$(pwd)` before running the binaries, so `openClose01_n`'s path lookup resolves there, and the script propagates the gtest exit code, which fails `%check` and therefore the job.

So the *behavioural* contract — a file that exists but is not a tvm module must make `open()` fail — is enforced per-PR and does block a merge. What is not enforced is the leak itself, since the valgrind job (`ubuntu_clean_meson_build.yml`) does not install `tvm-runtime-dev` and its valgrind step is non-fatal by design. That half of the earlier review stands. For completeness: the pdebuild job does build the sub-plugin (`debian/control` has `tvm-runtime-dev`) but `override_dh_auto_test` does not run `tests/nnstreamer_filter_tvm`, so it contributes nothing here.

**Changes pushed** (branch amended, one commit):
- `openClose01_n` now also asserts that a failed `open()` leaves `private_data` untouched, and calls `close()` afterwards, so the failed-configure teardown path is exercised end to end rather than only through the return code.
- The `NNSTREAMER_SOURCE_ROOT_PATH` lookup that the new case had duplicated is factored out of `SetModelFile()` into `GetTestModelPath()`, and the invalid model path is held in the fixture and freed in `TearDown()` like the other model paths.

The sub-plugin change itself is unchanged.

Two things I looked at and deliberately left alone, so they are on the record rather than lost:
- `configure_instance()` dereferences `prop->model_files[0]` in the `g_file_test` branch without a NULL check — `(std::string) prop->model_files[0]` would be UB if it were NULL. It is unreachable through the element, since `verify_model_path()` in `tensor_filter_common.c` rejects `num_models <= 0` before `open()`, so it is defensive hardening of a different defect class and does not belong in a leak fix.
- Making the leak itself CI-visible needs `tvm-runtime-dev` added to the valgrind job plus a project-wide decision to make valgrind failures fatal, the latter explicitly deferred in that workflow's own comment.

myungjoo:

Reviewed the current head commit (`6d80b8e`) against `ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc` and `tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc`, plus `gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc` for the open/close wrapper, and — since a prior round's CI-detectability claim was corrected mid-thread — I independently re-derived that correction from `packaging/nnstreamer.spec`, `.github/workflows/gbs_build.yml`, `.github/actions/check-rebuild/check_if_rebuild_requires.sh`, and `packaging/run_unittests_binaries.sh` rather than taking it on trust.

**Verdict: no blocking issues. The fix is correct, minimally scoped, and adequately tested. Approve.**

What I checked beyond re-reading the diff:

1. **Fix correctness, independently re-traced.** In `configure_instance()`, `model_path = g_strdup(...)` happens before `LoadFromFile()`, and every throw site after that (`LoadFromFile` itself, the `get_input`/`get_output` null checks, both `convert_dtype` failures) either falls straight into the `unique_ptr` teardown in `cpp_open()` (for `LoadFromFile`) or calls `cleanup()` explicitly (the other three). `empty_model` is only set to `false` at the very end of a fully successful `configure_instance()`, so at every one of those throw sites it is still `true`. Moving `g_free(model_path); model_path = nullptr;` above the `if (empty_model) return;` guard is exactly what's needed to free it on all of them, and nulling it immediately makes the possible second `cleanup()` call (from `~tvm_subplugin()` when the explicit call already ran) a safe `g_free(nullptr)`.

2. **No new leak or double-free from the reorder, checked against the "skip" branch.** For the failure paths where `cleanup()` returns early after freeing `model_path` (still `empty_model == true`), `input_tensor_list`/`output_tensor_list`/`inputInfo`/`outputInfo` are left untouched. I confirmed this is safe, not just pre-existing-and-unexamined: in the per-tensor loops, `_info->name` is always set to `nullptr` and `dimension[]` is a fixed-size array, so `inputInfo`/`outputInfo` never hold separately-heap-allocated data before `empty_model` would flip to `false`. `gst_tensors_info_free()` therefore has nothing to reclaim on these paths regardless of whether it runs, so skipping it here is not a gap this PR introduces or fails to close.

3. **CI-detectability correction, verified from source rather than trusted.**
   - `packaging/nnstreamer.spec`: the default profile sets `%define tvm_support 1` (line 36); it's forced to `0` only under `tv`, `mv_prj`, and `_with_da_profile` profiles (lines 122/147/164), none of which the GBS matrix passes.
   - `.github/workflows/gbs_build.yml`: the matrix runs three `gbs build` invocations with no `profile` define, so the default (`tvm_support=1`) applies to all three; only the `x86_64` entry passes `--define "unit_test 1"` (armv7l/aarch64 get `unit_test 0`).
   - `.github/actions/check-rebuild/check_if_rebuild_requires.sh`: files outside `*.md/png/webp/css/html`, `packaging/*`, `debian/*`, `jni/*` fall into the catch-all branch that sets `regbs=1`. Both files this PR touches (`ext/...tensor_filter_tvm.cc`, `tests/...unittest_filter_tvm.cc`) hit that branch, so the GBS job is not skipped for this PR.
   - `packaging/nnstreamer.spec` (~line 1076-1077): under `%if 0%{?unit_test}` and `%ifarch %arm x86_64 aarch64`, it runs `bash %{test_script} ./tests/nnstreamer_filter_tvm`.
   - `packaging/run_unittests_binaries.sh`: exports `NNSTREAMER_SOURCE_ROOT_PATH=$(pwd)` (so `GetTestModelPath()`'s env lookup resolves), invokes each `unittest_*` binary, and propagates its exit code (`ret=$?`) as the script's own exit code, which fails the RPM build step and thus the "Tizen GBS build on Ubuntu (x86_64, `unit_test 1`)" job on failure.
   - This all matches the corrected claim in the thread: the **behavioral** contract that `openClose01_n` checks (opening a non-tvm regular file must make `open()` fail, not crash) is enforced per-PR by that one GBS job and will block a merge if it regresses. The **leak itself** is not: I found no `tvm-runtime-dev`/`tvm-runtime-devel` install step anywhere under `.github/workflows/`, so the separate valgrind job (`ubuntu_clean_meson_build.yml`) never builds `tensor_filter_tvm` at all, and that job's valgrind step is explicitly non-fatal by its own inline comment regardless. So this correction holds up under independent re-check.

4. **Test itself is sound.** `openClose01_n` correctly reuses the newly-factored `GetTestModelPath()` helper, asserts the target file exists as a precondition (so the test itself doesn't silently pass for the wrong reason if `test_models/labels/labels.txt` ever moves), checks `data == nullptr` on failed open (matching `cpp_open()`'s behavior of never writing `*private_data` before a successful return), and exercises `close()` on that null-data state afterward — which I traced through `cpp_close()`/`get_tfsp_with_checks()` and confirmed is a safe no-op path. Fixture cleanup (`g_free(invalid_model_file)` in `TearDown()`) is present, matching the other model-path fields.

5. **Scope.** Two files, one 3-line functional reorder plus a factored-out helper and one new test case. No unrelated modules touched, no public API or architecture change, so no documentation update is expected or missing.

Non-blocking observations (already surfaced in this thread, restated only for completeness of this pass — not new asks):
- Leak-specific CI coverage remains a gap, but closing it needs a `tvm-runtime-dev` install in the valgrind job plus a project-wide decision to make valgrind failures fatal, which is out of scope for this fix and already flagged as deliberately deferred.
- The pre-existing lack of a NULL check on `prop->model_files[0]` before `(std::string) prop->model_files[0]` in the `g_file_test` branch is a different (currently unreachable, since `verify_model_path()` gates `num_models <= 0` upstream) defect class and correctly left out of this leak fix.

No new issues found. I'd merge this as-is.

myungjoo:
All 18 checks pass on `6d80b8e`.

For the record, the CI-detectability question is now settled empirically rather than by reading the spec. The GBS x86_64 (`unit_test 1`) job log shows the sub-plugin test binary being built and the new case running:

```
[290/418] Compiling C++ object tests/nnstreamer_filter_tvm/unittest_filter_tvm.p/unittest_filter_tvm.cc.o
[412/418] Linking target tests/nnstreamer_filter_tvm/unittest_filter_tvm
[----------] 15 tests from NNStreamerFilterTVMTest
[ RUN      ] NNStreamerFilterTVMTest.openClose01_n
[       OK ] NNStreamerFilterTVMTest.openClose01_n (0 ms)
```

So `openClose01_n` is executed per-PR and a regression in the failure path would fail this job and block a merge. The leak itself still needs valgrind, which no per-PR job applies to this sub-plugin — unchanged and out of scope here.

myungjoo:
Filed the sanitizer-coverage point as #4888 rather than carrying it here.

Investigating it turned up more than "the valgrind job lacks tvm". That job builds 26 unittest binaries and gets through 4 of them: `packaging/run_unittests_binaries.sh:100` breaks out of its loop on the first non-zero return, and `unittest_filter_custom` fails under valgrind every run (`tensorFilterCustom.flexibleInvoke_p`, 13393 ms — timing-sensitive against valgrind's slowdown). Everything after it is silently skipped. On top of that, the valgrind invocation has no `--error-exitcode`, so findings never reach the exit status, and the workflow step swallows the result with `|| echo`.

Also worth correcting something I wrote earlier in this thread: I repeated the workflow's own claim that there are "a lot of Valgrind errors" to fix first. Measured, that is not the obstacle — the 4 binaries that do run total `ERROR SUMMARY` of 0, 5, 8 and 18, with one `definitely lost: 128 bytes in 1 blocks`. The obstacle is four independent layers each neutralising the check while the job stays green.

None of that belongs in a three-line leak fix, and the issue says so. This PR stands on the behavioral contract that `openClose01_n` pins.

PR #4885, opened 2026-08-27
[Test] Install the SSAT suites that were missing from unittest-nnstreamer

myungjoo:
`tests/meson.build` enumerates by hand which SSAT directories go into the `unittest-nnstreamer` package, and that list stopped being maintained around 2021. Twelve suites were absent, so on a target device they do not exist and cannot be run against the installed nnstreamer — even though the gtest binaries from the same directories *are* installed. This fixes the list and then stops it drifting again.

## Changes

**Install the eleven suites that were missing:** `nnstreamer_decoder_tensor_region`, `nnstreamer_filter_lua`, `nnstreamer_sparse`, `nnstreamer_filter_litert`, `nnstreamer_filter_onnxruntime`, `nnstreamer_filter_tensorrt`, `nnstreamer_filter_tensorrt10`, `nnstreamer_filter_tvm`, `nnstreamer_grpc`, `nnstreamer_octet_stream`, `gstreamer_join`. Framework-specific ones are gated on the matching `*_support_is_available` flag, following the `tflite`/`executorch` entries above them.

**Install `get_available_port.py`.** `nnstreamer_grpc/runTest.sh` invokes it as `../get_available_port.py`; without it that suite would *fail* rather than skip once installed.

**Leave `codegen` out**, with a comment saying why.

**Delete the mxnet entry.** It installed `unittest_filter_mxnet`, which is not a directory in this tree — the real one is `tests/nnstreamer_filter_mxnet`, and it has no `runTest.sh`, so there is no SSAT suite to install either way. It went unnoticed because it only fires when `mxnet_support_is_available`, which is never true in CI.

**Extend `tensor_filter_ext_enabled`** — which gates the `test_models` install — with litert, onnxruntime, tensorrt, tensorrt10, tvm and lua. Every suite added here reads from `../test_models`, so without this a build enabling only one of those frameworks would ship a suite without its fixtures. That variable is defined at `tests/meson.build:448` and used only at line 456, so widening it changes nothing else.

**Add a drift check.** The root cause is a hand-maintained list with nothing detecting divergence: no one was told to update it and CI stayed green while twelve entries went missing. `.github/workflows/static.check.scripts/ssat-install-list.sh` compares both directions — a suite with a `runTest.sh` that is not installed, and an entry naming a directory that does not exist — and carries `codegen` as an explicit exemption with its reason next to the name. The mxnet entry above is exactly what the second direction catches. The workflow step uses the same guard as the check-rebuild self-test, since the step list comes from the merge ref while the checkout is the PR head.

## How each suite was checked

Every candidate was inspected for references outside its own directory (ignoring `../test_models`, which is installed, and `../../build`, the build-mode probe):

| suite | external reference | verdict |
| --- | --- | --- |
| `gstreamer_join`, `octet_stream`, `sparse`, `filter_tensorrt`, `filter_tensorrt10`, `filter_tvm`, `filter_litert` | none | install |
| `filter_onnxruntime` | `../` only, inside a deliberately-invalid model path in a negative case | install |
| `decoder_tensor_region` | `../nnstreamer_decoder_boundingbox/{box_priors.txt,coco_labels_list.txt}` — installed unconditionally | install |
| `filter_lua` | `../nnstreamer_filter_python3/checkScaledTensor.py` — installed when `have_python3` | install |
| `grpc` | `../get_available_port.py` — **was not installed**; added here | install |
| `codegen` | `../../tools/development/nnstreamerCodeGenCustomFilter.py`, `../../gst/nnstreamer`, `../../gst/nnstreamer/include` | **exclude** |

`codegen` is the one that genuinely cannot work outside a source tree. `nnstreamer_sparse` is gated on lua rather than a converter flag because its own script says so: *"Check lua libraries are built. This test utilizes it for making 'sparse' dense tensors."* `nnstreamer_decoder_tensor_region` is unconditional because it runs no framework — its `PATH_TO_MODEL` is assigned and never used, and the pipeline feeds pre-dumped `mobilenet_ssd_tensor.{0,1}`.

The audit above covers `../`-relative references pointing outside `test_models`; it does not verify that fixtures *inside* `test_models` resolve. #4890 records one that does not (`tensorrt10` names `yolov5nu_224.onnx`, which is not in the tree) — pre-existing, but this PR is what puts that suite on devices.

## What this does and does not achieve

Nine of the eleven become runnable on a device. `filter_lua` and `sparse` do not: their build-directory probe ends with `echo "No build directory"; report; exit` rather than falling back to `/etc/nnstreamer.ini`, so from an installed tree they report no cases.

That shape is pre-existing — six suites already in the install list behave identically (`converter_python3`, `decoder_python3`, `filter_python3`, `flatbuf`, `flexbuf`, `protobuf`). Teaching all eight installed-mode discovery is a separate change, and doing it for only the two this PR happens to touch would be the same arbitrary partial fix that left this list stale. They are installed here regardless, since that later fix is only useful once the scripts ship.

One consequence worth noting: the `filter_lua` dependency on `checkScaledTensor.py` cannot bite today, because the script exits at the build-directory probe long before reaching that call. If `filter_lua` later gains the fallback, its install should be gated on `have_python3` as well.

## Scope and risk

No test logic changes, and nothing about how CI runs the suites changes — CI always executes them from the build tree, where discovery comes from `GST_PLUGIN_PATH` / `NNSTREAMER_CONF` / `NNSTREAMER_FILTERS` exported by `nnstreamer.spec` and `debian/rules`, not from this list. The only behavioural effect is the contents of the installed test package.

`%files unittests` globs the directory (`%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests`), so no packaging file needs to change. The package grows by about 1.1MB, ~924KB of which is the `mobilenet_ssd_tensor.*` fixtures in `decoder_tensor_region`; every other suite is 8-40KB.

**Self evaluation:**
1. Build test: [*]Passed [ ]Failed [ ]Skipped (both `-Dinstall-test=true` builds are green)
2. Run test: [ ]Passed [ ]Failed [*]Skipped (no test logic changed)

**How to evaluate:**
1. Build and install with `-Dinstall-test=true`, then confirm the eleven directories appear under `<prefix>/lib/nnstreamer/bin/unittest-nnstreamer/tests/` for the frameworks enabled in that build, and that `codegen` is absent.
2. Run `bash .github/workflows/static.check.scripts/ssat-install-list.sh` from the repository root; it should report the list is in sync. Delete an `install_subdir` entry, or point one at a non-existent directory, and it should fail with a message naming the fix.
3. Confirm CI's SSAT results are unchanged, since those runs never consult this list.

### DIFF
--- a/.github/workflows/static.check.scripts/ssat-install-list.sh
+++ b/.github/workflows/static.check.scripts/ssat-install-list.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file     ssat-install-list.sh
+# @brief    Keep the install_subdir list in tests/meson.build in sync with the SSAT suites
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# The list that fills the unittest-nnstreamer package is hand-maintained, so a new
+# tests/<suite>/runTest.sh can be added without ever being shipped, and an entry can
+# outlive the directory it names. Both happened. This checks the two directions.
+#
+
+meson_file="tests/meson.build"
+
+# Suites deliberately absent from the package. Keep the reason next to the name.
+#   codegen: runs tools/development/nnstreamerCodeGenCustomFilter.py and compiles
+#            against gst/nnstreamer through ../../ paths that only a source tree has.
+not_installed=("codegen")
+
+if [ ! -f "${meson_file}" ]; then
+  echo "::error::${meson_file} not found. Run this from the repository root."
+  exit 1
+fi
+
+failed=0
+installed=$(grep -vE '^[[:space:]]*#' "${meson_file}" | grep -oE "install_subdir\('[^']+'" | cut -d"'" -f2)
+
+# Every SSAT suite should be installed, unless it is listed above.
+while IFS= read -r script; do
+  suite=$(basename "$(dirname "${script}")")
+
+  skip=0
+  for exempt in "${not_installed[@]}"; do
+    if [ "${suite}" = "${exempt}" ]; then skip=1; break; fi
+  done
+  [ ${skip} -eq 1 ] && continue
+
+  if ! echo "${installed}" | grep -qxF "${suite}"; then
+    echo "::error::tests/${suite} has a runTest.sh but is missing from ${meson_file}, so it will not reach the unittest-nnstreamer package. Add install_subdir('${suite}', install_dir: unittest_install_dir), or add it to not_installed in $0 with the reason."
+    failed=1
+  fi
+done < <(find tests -mindepth 2 -maxdepth 2 -name runTest.sh)
+
+# Every entry should name a directory that exists.
+for suite in ${installed}; do
+  if [ ! -d "tests/${suite}" ]; then
+    echo "::error::${meson_file} installs '${suite}', but tests/${suite} does not exist."
+    failed=1
+  fi
+done
+
+if [ ${failed} -ne 0 ]; then
+  exit 1
+fi
+
+echo "The SSAT install list in ${meson_file} is in sync."
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -175,6 +175,20 @@ jobs:
           else
             echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
           fi
+      - name: /Checker/ SSAT install list is in sync
+        # Same guard as the check-rebuild self-test above: the step list comes
+        # from the merge ref while the checkout is the PR head, so a branch that
+        # predates this script would fail with "No such file or directory".
+        run: |
+          checker=.github/workflows/static.check.scripts/ssat-install-list.sh
+          if [ -f "$checker" ]; then
+            bash "$checker"
+          elif git show --pretty="format:" --name-only --no-renames --diff-filter=D ${{ github.event.pull_request.head.sha }} -${{ github.event.pull_request.commits }} | grep -qxF "$checker"; then
+            echo "::error::This PR removes $checker; remove its workflow step too, or restore the script (also update this step if the script moved)."
+            exit 1
+          else
+            echo "Skipped: $checker is not in this branch's tree and this PR does not remove it."
+          fi
       - name: /Checker/ shellcheck for shell scripts
         # Originally from "pr-prebuild-shellcheck"
         run: |
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -452,7 +452,10 @@ tensor_filter_ext_enabled = tflite_support_is_available or \
     pytorch_support_is_available or caffe2_support_is_available or \
     nnfw_runtime_support_is_available or get_option('enable-edgetpu') or \
     mvncsdk2_support_is_available or get_option('enable-openvino') or \
-    mxnet_support_is_available
+    mxnet_support_is_available or litert_support_is_available or \
+    onnxruntime_support_is_available or tensorrt_support_is_available or \
+    tensorrt10_support_is_available or tvm_support_is_available or \
+    lua_support_is_available
 if get_option('install-test') and tensor_filter_ext_enabled
   # host-only model generator scripts are not needed on target devices
   install_subdir('test_models', install_dir: unittest_install_dir,
@@ -462,13 +465,19 @@ endif
 
 # Install Unittest
 if get_option('install-test')
+  # tests/codegen is intentionally not installed: it runs tools/development/
+  # nnstreamerCodeGenCustomFilter.py and compiles against gst/nnstreamer via
+  # ../../ paths that exist only in the source tree.
   install_data('gen24bBMP.py', install_dir: unittest_install_dir)
   install_data('test_utils.py', install_dir: unittest_install_dir)
+  install_data('get_available_port.py', install_dir: unittest_install_dir)
+  install_subdir('gstreamer_join', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_converter', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_merge', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_decoder', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_decoder_boundingbox', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_decoder_pose', install_dir: unittest_install_dir)
+  install_subdir('nnstreamer_decoder_tensor_region', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_demux', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_filter_custom', install_dir: unittest_install_dir)
   if tflite_support_is_available or tflite2_support_is_available
@@ -509,11 +518,31 @@ if get_option('install-test')
   if protobuf_support_is_available
     install_subdir('nnstreamer_protobuf', install_dir: unittest_install_dir)
   endif
-  if mxnet_support_is_available
-    install_subdir('unittest_filter_mxnet', install_dir: unittest_install_dir)
-  endif
   if executorch_support_is_available
     install_subdir('nnstreamer_filter_executorch', install_dir: unittest_install_dir)
+  endif
+  if litert_support_is_available
+    install_subdir('nnstreamer_filter_litert', install_dir: unittest_install_dir)
+  endif
+  if lua_support_is_available
+    install_subdir('nnstreamer_filter_lua', install_dir: unittest_install_dir)
+    # nnstreamer_sparse builds its sparse tensors with the lua filter
+    install_subdir('nnstreamer_sparse', install_dir: unittest_install_dir)
+  endif
+  if onnxruntime_support_is_available
+    install_subdir('nnstreamer_filter_onnxruntime', install_dir: unittest_install_dir)
+  endif
+  if tensorrt_support_is_available
+    install_subdir('nnstreamer_filter_tensorrt', install_dir: unittest_install_dir)
+  endif
+  if tensorrt10_support_is_available
+    install_subdir('nnstreamer_filter_tensorrt10', install_dir: unittest_install_dir)
+  endif
+  if tvm_support_is_available
+    install_subdir('nnstreamer_filter_tvm', install_dir: unittest_install_dir)
+  endif
+  if grpc_support_is_available
+    install_subdir('nnstreamer_grpc', install_dir: unittest_install_dir)
   endif
   install_subdir('nnstreamer_mux', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_rate', install_dir: unittest_install_dir)
@@ -522,6 +551,7 @@ if get_option('install-test')
   install_subdir('nnstreamer_repo_lstm', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_repo_rnn', install_dir: unittest_install_dir)
   install_subdir('nnstreamer_split', install_dir: unittest_install_dir)
+  install_subdir('nnstreamer_octet_stream', install_dir: unittest_install_dir)
   install_subdir('transform_arithmetic', install_dir: unittest_install_dir)
   install_subdir('transform_clamp', install_dir: unittest_install_dir)
   install_subdir('transform_dimchg', install_dir: unittest_install_dir)


myungjoo:

Reviewed the diff (`tests/meson.build`, +30/-0) against the current `main` tree. Summary: no blocking issues; recommend approve with two non-blocking follow-ups noted below.

**Verified correct:**

1. **Gate variables are always defined.** All seven flags used (`litert_support_is_available`, `lua_support_is_available`, `onnxruntime_support_is_available`, `tensorrt_support_is_available`, `tensorrt10_support_is_available`, `tvm_support_is_available`, `grpc_support_is_available`) are registered in the top-level `meson.build` `features` dict (lines ~603-753), and the `foreach feature_name, data : features` loop (lines ~757-810) calls `set_variable(variable_available_name, ...)` unconditionally on **both** the enabled and disabled/not-found branches. `subdir('tests')` (line 994) runs well after this loop, so every flag exists by the time `tests/meson.build` reads it — and several of them (e.g. `lua_support_is_available`, `tvm_support_is_available`) are already dereferenced earlier in the same file (lines 254, 313, 345, 401, 413) predating this PR. No undefined-variable build failure risk.

2. **External-reference table checks out.** Verified each of the 11 directories against the PR's own table:
   - `gstreamer_join`, `nnstreamer_octet_stream`, `nnstreamer_filter_tensorrt`, `nnstreamer_filter_tensorrt10` — only reference `../../build` (build-mode probe) and/or `../test_models` (installed unconditionally). Correct.
   - `nnstreamer_filter_onnxruntime` line 97 does concatenate an extra `../` onto an already-relative `PATH_TO_MODEL` for a negative "invalid path" case — confirmed this is a deliberately-wrong path either way (in-tree or installed), so it's harmless.
   - `nnstreamer_decoder_tensor_region` → `../nnstreamer_decoder_boundingbox/{box_priors.txt,coco_labels_list.txt}`: that directory is installed unconditionally (line 471, outside any `if`). Correct.
   - `nnstreamer_grpc` → `../get_available_port.py`: now installed by this PR at the right sibling location. Correct.
   - `nnstreamer_filter_tvm`: today (before #4873 lands) this directory only contains `meson.build` + `unittest_filter_tvm.cc`, no `runTest.sh` yet. Installing it now is harmless filler (no SSAT script to run), consistent with the PR's own "independent of #4873" note.

3. **`nnstreamer_sparse` gated on `lua_support_is_available`** is justified — its own `runTest.sh` comment ("Check lua libraries are built. This test utilizes it for making 'sparse' dense tensors") and its lua-plugin presence check both confirm the dependency is real, not incidental.

4. **No unwanted files.** `find` over all 11 directories shows only source (`.cc`, `.py`), scripts (`runTest.sh`), a `meson.build`, and the disclosed `mobilenet_ssd_tensor.{0,1}` + `tensor_region_orange.txt` fixtures — no build artifacts, no `.git`, no credentials, sizes match what the PR describes.

5. **Packaging needs no changes.** `packaging/nnstreamer.spec` `%files unittests` lists `%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests` as a bare directory path (line 1447), which RPM packages recursively — new subdirectories are automatically covered. Debian's `debian/*.install` files don't reference this test tree at all, before or after this PR (pre-existing, unaffected), and `debian/rules` invokes `dh_install --list-missing` (not `--fail-missing`), so even unclaimed files only warn, never fail the build.

6. **CI is unaffected**, and where it *would* catch a real mistake, it does. `ubuntu_clean_meson_build.yml` runs SSAT directly from the source `tests/<dir>` with `NNSTREAMER_BUILD_ROOT_PATH`/`GST_PLUGIN_PATH` env vars, never from an installed prefix — confirmed the install list this PR edits is irrelevant to it. Meanwhile `pdebuild.yml` and `gbs_build.yml` both trigger per-PR on `main` and build with `-Dinstall-test=true` (via `debian/rules` / the spec's `install_test` define), so they *do* exercise exactly the code path this PR adds; a genuine meson error (e.g. an actually-undefined variable) would fail one of those jobs and block the merge, which is confirmed as a real safety net for structural mistakes.

**Two non-blocking findings for awareness:**

- **Minor — `nnstreamer_filter_lua` and `nnstreamer_sparse` will silently no-op once actually run from an installed location.** Their `runTest.sh` (unlike `nnstreamer_filter_litert`/`nnstreamer_filter_onnxruntime`/`nnstreamer_filter_tensorrt(10)`, which this same PR also adds) has no `/etc/nnstreamer.ini` fallback branch: when `../../build` doesn't exist (the installed case), they hit `echo "No build directory"; report; exit` and skip all cases without ever exercising anything. This means 2 of the 11 "install" verdicts don't actually deliver a runnable installed suite the way the PR's rationale intends — though this is a pre-existing limitation shared by several *already*-installed suites (`nnstreamer_filter_python3`, `nnstreamer_converter_python3`, `nnstreamer_decoder_python3`, `nnstreamer_flatbuf`, `nnstreamer_flexbuf`, `nnstreamer_protobuf` all have the identical "No build directory" early-exit with no ini fallback), so it's not a regression this PR introduces, just an extension of an existing gap. No CI job runs SSAT from an installed prefix, so nothing will catch this either way. Worth a small fast-follow to add the same ini-fallback pattern to these two scripts, or at least noting the caveat in the PR description.

- **Minor/Moderate — cross-suite dependency not fully covered by the `lua` gate.** `nnstreamer_filter_lua/runTest.sh` line 55 unconditionally runs `python3 ../nnstreamer_filter_python3/checkScaledTensor.py ...` as part of test case 2, but `nnstreamer_filter_python3` is only installed `if have_python3` (`tests/meson.build` ~line 490). `lua-support` and `python3-support` are independent meson options, so a build with lua enabled and python3 disabled will install `nnstreamer_filter_lua` (gated only on `lua_support_is_available`) with a test case that references a script that won't exist on that target — an actual runtime test **failure**, not a graceful skip, in that specific configuration. The PR's own table acknowledges the `../nnstreamer_filter_python3/checkScaledTensor.py` dependency but doesn't account for the `have_python3=false` + `lua=true` combination. Consider additionally gating `nnstreamer_filter_lua`'s install on `have_python3`, or guarding that one call in the script.

Neither finding blocks this PR — both are narrow-configuration or pre-existing-pattern issues, not build breaks, not regressions in other modules, and the change stays tightly scoped to its stated goal (30 lines, one file, no other module touched). Overall: mergeable as-is; the two items above are good candidates for a quick follow-up PR.

myungjoo:
Both findings check out. Verified them and corrected the commit message and PR description, which overclaimed by saying all eleven "work as installed".

**Finding 1 — `filter_lua` and `sparse` are inert when installed: confirmed.** Both end the build-directory probe with

```sh
else
    echo "No build directory"
    report
    exit
fi
```

so from an installed tree they report no cases. Nine of the eleven do become runnable; those two do not.

Surveying the whole install list rather than just this diff: eight suites have that shape, and six of them were already in the list before this PR — `converter_python3`, `decoder_python3`, `filter_python3`, `flatbuf`, `flexbuf`, `protobuf`. So this is a pre-existing pattern that the PR extends from six to eight, not one it introduces.

I am keeping both installed rather than dropping them. Adding the `/etc/nnstreamer.ini` fallback to two of the eight would be exactly the arbitrary partial fix that left this list stale to begin with, and that fix is only useful once the scripts actually ship. Worth noting the eight are cheap to fix but the new branch would be untestable in CI, which never runs SSAT from an installed tree — so it deserves its own change with that caveat stated, not a quiet ride-along here.

This is a different situation from `codegen`, which stays excluded: `codegen` needs `tools/development/` and `gst/nnstreamer` at `../../` paths that do not exist in any installed layout, so no script-local change makes it work.

**Finding 2 — the `checkScaledTensor.py` gap: real in principle, unreachable in practice.** `filter_lua/runTest.sh` does call it unconditionally, and `lua-support` and python3 are indeed independent options. But finding 1 forecloses it: the script exits at the build-directory probe long before reaching that line, so an installed `filter_lua` never invokes it no matter what `have_python3` was. I have therefore not gated the install on `have_python3` — that would encode a dependency nothing can currently exercise. Instead the PR description records that the gate must be added *if* `filter_lua` ever gains the fallback.

One correction to my own earlier analysis while checking this: `nnstreamer_grpc` initially looked like it belonged in the inert group, but it does not. Its probe is `PATH_TO_PLUGIN=${NNSTREAMER_BUILD_ROOT_PATH}` with `../../build` only as a default and no exit, so it runs when installed — which is precisely why `get_available_port.py` had to be added.

No code change; the install list is unchanged and the diff is still 30 lines in one file.

myungjoo:

Second-pass re-review. Confirmed the diff is unchanged since the first review (`tests/meson.build`, +30/-0, one file), and this update is prose-only: commit message and PR body were corrected, no code changed.

**Verified each corrected claim directly against the repo files:**

1. **"Nine of the eleven become runnable on a device."** Correct. Of the eleven, `filter_lua` and `sparse` are the two that report no cases when installed; the remaining nine are runnable.

2. **`filter_lua` and `sparse` end with the no-fallback exit.** Confirmed in both `tests/nnstreamer_filter_lua/runTest.sh` and `tests/nnstreamer_sparse/runTest.sh`:
   ```sh
   else
       echo "No build directory"
       report
       exit
   fi
   ```
   `PATH_TO_PLUGIN="../../build"` won't exist in an installed tree, so this branch is taken and both scripts exit before any test runs.

3. **The six-suite precedent (`converter_python3`, `decoder_python3`, `filter_python3`, `flatbuf`, `flexbuf`, `protobuf`).** Confirmed — all six already-installed suites have the byte-for-byte identical `PATH_TO_PLUGIN="../../build"` → `echo "No build directory"; report; exit` shape. This is genuinely pre-existing behavior, not something this PR introduces.

4. **`nnstreamer_grpc` uses `NNSTREAMER_BUILD_ROOT_PATH` and doesn't exit.** Confirmed. Its probe is:
   ```sh
   PATH_TO_PLUGIN=${NNSTREAMER_BUILD_ROOT_PATH}
   if [[ -z $PATH_TO_PLUGIN ]]; then
     PATH_TO_PLUGIN="../../build"
   fi
   if [[ -d $PATH_TO_PLUGIN ]]; then
       ...
   fi
   ```
   Unlike `filter_lua`/`sparse`, there is no `else`/`exit` here — if the directory check fails, the script only prints a diagnostic and falls through to the actual `gstTest` calls. So it behaves as claimed.

5. **Finding (2) is unreachable because of finding (1).** Confirmed — `filter_lua`'s call to `../nnstreamer_filter_python3/checkScaledTensor.py` sits after the `PATH_TO_PLUGIN` build-directory check in script order, so in an installed tree the script has already `exit`ed by the time it would reach that line. The dependency is real but currently dead code from the installed-tree perspective, exactly as described.

**Diff scope:** still exactly one file (`tests/meson.build`, +30/-0), matching the first-pass review. No new code-level concerns.

**CI:** all checks green except one still-pending, unrelated RISC-V build job.

No remaining blocking or non-blocking issues from this pass — both minor points from the first review were about pre-existing script behavior, not this diff, and the author's corrected description now matches the code exactly. This PR is mergeable as-is.

myungjoo:
CI is now 18/18. Marked ready for review; the hold label is off.

For the record on the RISC-V job, which failed twice before passing on the third re-run with no code change in between: both failures were `cc: internal compiler error: Segmentation fault signal terminated program cc1`, but on different source files each time — `gsttensor_transform.c` on the first attempt, `tensordec-imagesegment.c` on the second. Neither is touched by this PR, and a deterministic problem would reproduce on the same file. It looks like memory pressure under qemu emulation.

Independently of that, this diff cannot reach that job at all: `risc-v.yml` configures with `meson setup build -Denable-tizen=false -Denable-test=true`, which leaves `install-test` at its `meson_options.txt` default of `false`, and the job only runs `ninja -C build` without any install target. Every line added here is inside `if get_option('install-test')`.

myungjoo:
Thorough review — thanks. Four of the five are now in the PR; the fifth is #4890. Also rebased onto current main, which had moved past #4873 and #4882.

**1. Drift detection (medium): added.**

`.github/workflows/static.check.scripts/ssat-install-list.sh`, wired into the static checks. It compares both directions:

- a `tests/*/runTest.sh` whose directory is absent from the install list, and
- an `install_subdir(...)` entry naming a directory that does not exist.

`codegen` is carried as an explicit exemption with its reason next to the name, so the exemption is the thing a reader has to justify rather than the omission.

I verified it actually fails, rather than assuming: deleting the `nnstreamer_filter_tvm` entry produces the first error, and adding a bogus entry produces the second. Worth noting one wrinkle from that test — my first run of the forward check passed when it should not have, because this branch predated the #4873 merge and so had no `tests/nnstreamer_filter_tvm/runTest.sh` to find. Rebasing fixed it, and it now catches the case.

The workflow step uses the same guard as the check-rebuild self-test, per the comment there that any step running a repository script unrelated to `$changed_file_list` needs it.

**2. `unittest_filter_mxnet` (low): removed.**

Confirmed: no `tests/unittest_filter_mxnet` exists, the real directory is `tests/nnstreamer_filter_mxnet`, and it has no `runTest.sh` — so the entry was wrong twice and there is no SSAT suite to point it at. Deleted rather than corrected. It is also exactly what the new check's reverse direction catches, so it doubles as that check's first real find.

**3. `test_models` gating (low): fixed.**

`tensor_filter_ext_enabled` now includes litert, onnxruntime, tensorrt, tensorrt10, tvm and lua. I checked before widening it: `tests/meson.build:448` is its definition and line 456 its only use, so this affects nothing but the `test_models` install.

On `decoder_tensor_region` being unconditional — your reading is right and I should have said so in the description. `PATH_TO_MODEL` at line 25 is assigned and never used; the pipeline feeds the pre-dumped `mobilenet_ssd_tensor.{0,1}` rather than running tflite, so there is no framework to gate on. Its `../test_models/data/orange.png` dependency is the one real link, and that is now covered whenever any filter framework is enabled.

**4. `yolov5nu_224.onnx` (low, pre-existing): filed as #4890.**

Confirmed missing — only `mobilenet_v2_{float,quant}.onnx` are tracked, nothing fetches or generates it, and the script has no fixture guard, so the positive cases fail rather than skip. You are right that my audit table missed it: I checked `../`-relative references pointing outside `test_models` and implicitly trusted everything inside it. Out of scope to fix here, but this PR is what puts the suite on devices, so it needed recording rather than a footnote.

**5. Placement: fixed.**

The `codegen` comment now sits at the top of the block instead of between `gstreamer_join` and `transform_arithmetic`, where it read as a note about the `transform_*` entries. `gstreamer_join` moved up next to the `install_data` lines so the `nnstreamer_*` run is unbroken.

myungjoo:
Thanks for re-deriving the checker's result independently instead of trusting its green output — that is the check I would have wanted on it.

Took three of the five follow-ups; the other two are recorded rather than done.

**Nit 2 — doxygen comment scanner: fixed.** This was mine to clean up, since the file introduced the repo's first such warning. The loop now reads from `find tests -mindepth 2 -maxdepth 2 -name runTest.sh` instead of the `tests/*/runTest.sh` glob, so there is no `/*` left in the file for doxygen to mistake for a comment opener. Verified none remains.

**Nit 4 — grep robustness: the fail-open half is fixed.** A commented-out `install_subdir(...)` no longer counts as installed; the extraction now filters comment lines first. I re-ran all three failure modes after the change rather than assuming: a missing entry, an entry naming a non-existent directory, and now a commented-out entry each produce the right error and exit 1, and the unmodified tree still reports in sync. The multi-line-call half I left alone — as you say it fails closed, and guarding against a pattern that does not exist would be speculative.

**Nit 5 — PR description: rewritten.** It now covers the drift checker, the `unittest_filter_mxnet` deletion and the `tensor_filter_ext_enabled` extension, which were indeed most of the diff and entirely absent from the old body. It also records the boundary of the audit table you found the hole in — that it covers `../`-relative references pointing outside `test_models` and does not verify fixtures inside it.

**Nit 1 — tensorrt10 fixture: already filed as #4890**, before this round. It confirms your finding independently: only `mobilenet_v2_{float,quant}.onnx` are tracked, nothing fetches or generates the yolov5 file, and the script has no fixture guard, so the positive cases fail rather than skip. Referenced from the PR body now too. Agreed the checker should not try to cover that class — directory membership is the right boundary for it.

**Nit 3 — `install_data` coverage: not doing it.** Agreed with your own assessment. `get_available_port.py` being missing was found by reading the suites' references, not by a rule, and a checker that resolves every `../<file>` reference across 51 scripts is a much larger thing than this one for a rarer failure. Left as-is deliberately rather than overlooked.

Diff is now three files, +107/-4, still one commit.

Issue #4883, opened 2026-08-27
[Test/llama.cpp] multipleInputsSingleOutputSync_p intermittently times out waiting for the second sample


myungjoo:
## Summary

`NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p` fails intermittently in the "Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)" job. Two prompts are pushed in synchronous mode but only the first ever produces an output sample, so the test sits until its 10 s ceiling.

```
[ RUN      ] NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p
 her her favourite eye, has a wonder../tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc:309: Failure
Value of: wait_for_samples (2)
  Actual: false
Expected: true
../tests/nnstreamer_filter_llamacpp/unittest_filter_llamacpp.cc:312: Failure
Expected equality of these values:
  *new_sample_count
    Which is: 1
  2
[  FAILED  ] NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p (10111 ms)
```

Observed on [run 32983423077](https://github.com/nnstreamer/nnstreamer/actions/runs/32983423077) (PR #4875, whose diff touches only `gst/join` and the datarepo tests). It passed on re-run.

This is new CI surface: the llamacpp job was enabled by #4863.

## The runtime is wildly unstable even when it passes

Same test, same model, consecutive runs:

| run | result | duration |
|---|---|---|
| [32974102457](https://github.com/nnstreamer/nnstreamer/actions/runs/32974102457) attempt 1 | pass | 3917 ms |
| 32974102457 attempt 2 | pass | 403 ms |
| [32983423077](https://github.com/nnstreamer/nnstreamer/actions/runs/32983423077) attempt 1 | **fail** | 10111 ms (timeout) |
| 32983423077 attempt 2 | pass | — |

For comparison, in those same runs `singleInputSingleOutputSync_p` (one prompt) takes 130–232 ms and `multipleInputsMultipleOutputsAsync_p` (two prompts, async) takes 131–146 ms. The instability is specific to **synchronous mode with more than one input**.

## What this is *not*

Worth stating up front, because the log is misleading:

[...]

PR #4889, opened 2026-08-27
[Test] Wait for the sink buffer before tearing down single-buffer filter pipelines

### LINKED ISSUE #4870: [Test/onnxruntime] quantModelLaunch00 races buffer arrival against pipeline teardown
`nnstreamerFilterOnnxRuntime.quantModelLaunch00`, added in #4865 (f189406d, merged 2026-08-26), has a race condition that makes it fail intermittently on CI. This issue carries the full analysis and the observed failure log so it can be fixed without reproducing the race.

### Observed failure

Tizen GBS x86_64 (`--define "unit_test 1"`) on an unrelated PR (#4855), first CI round after #4865 landed on main:
https://github.com/nnstreamer/nnstreamer/actions/runs/32964297798/job/98163301295

```
[  362s] [ RUN      ] nnstreamerFilterOnnxRuntime.floatModelResult
[  363s] [       OK ] nnstreamerFilterOnnxRuntime.floatModelResult (109 ms)
[  363s] [ RUN      ] nnstreamerFilterOnnxRuntime.quantModelLaunch00
[  363s] ../tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc:446: Failure
[  363s] Expected: (count) >= (1U), actual: 0 vs 1
[  363s] [  FAILED  ] nnstreamerFilterOnnxRuntime.quantModelLaunch00 (69 ms)
...
[  364s] [==========] 11 tests from 1 test suite ran. (1768 ms total)
[  364s] [  PASSED  ] 10 tests.
[  364s] [  FAILED  ] 1 test, listed below:
[  364s] [  FAILED  ] nnstreamerFilterOnnxRuntime.quantModelLaunch00
```

The 69 ms duration is the tell: the test ended almost immediately after reaching PLAYING, i.e. it did not wait for any buffer.

### Root cause

The test (`tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc`, `quantModelLaunch00`) does:

```c
g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);

EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10), 0);
EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);

EXPECT_GE (count, 1U);   /* <-- line 446 */
```

`setPipelineStateSync (PLAYING)` returns when the *state change* completes, not when data has flowed. The single buffer from `filesrc ! pngdec ! ... ! tensor_filter ! tensor_sink` is still traveling through the pipeline (the onnxruntime invoke alone takes tens of ms) when the very next call tears the pipeline down to NULL. If the teardown wins the race, `new-data` never fires and `count` stays 0.

Two contributing observations:

- `floatModelResult` (and the tflite equivalents) have the same PLAYING→NULL structure but never assert a buffer count — their `check_output` callback simply does not run when the buffer loses the race, so they pass vacuously. #4865 added the `count >= 1` assertion precisely to close that vacuous-pass hole, but without adding a wait, which converted a silent gap into a flaky failure.
- The race is load-dependent: it passed on the #4865 PR's own CI and fails sporadically on busy runners. Treat any `actual: 0 vs 1` at this line as this race, not as an onnxruntime regression.

### Suggested fix

Poll for the buffer before tearing down, with a bounded timeout, between the two state transitions:

```c
EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10), 0);

guint timer = 0;
while (count == 0U && timer < 10000U) {  /* up to 10 s */
  g_usleep (TEST_DEFAULT_SLEEP_TIME);
  timer += TEST_DEFAULT_SLEEP_TIME / 1000U;
}

EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
EXPECT_GE (count, 1U);
```

(`TEST_DEFAULT_SLEEP_TIME` and the polling idiom already exist in `tests/unittest_util.h` — see `_wait_pipeline_save_files` — so either inline the loop or add a small `_wait_pipeline_process_buffers`-style helper next to it.)

While touching this, consider giving `floatModelResult` (same file) the same count-plus-wait treatment so its golden check cannot pass vacuously either — that was the original intent of #4865.

### References

- #4865 — introduced the test (as the fix for the CPU-dependent quant-model flakiness; the quant/float split itself is correct and should be kept).
- #4855 — the unrelated PR whose CI surfaced this race first.
- `tests/unittest_util.h` — existing polling/timeout idiom to reuse.

myungjoo:
Fixes #4870.

## Problem

`nnstreamerFilterOnnxRuntime.quantModelLaunch00` asserts `count >= 1` immediately after driving the pipeline `PLAYING -> NULL`:

```c
EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, ...), 0);
EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, ...), 0);
EXPECT_GE (count, 1U);
```

`setPipelineStateSync (PLAYING)` returns when the *state change* completes, not when data has flowed. The single buffer from `filesrc ! pngdec ! ... ! tensor_filter ! tensor_sink` is still in flight when the teardown starts. On loaded runners the teardown wins and `new-data` never fires, producing the reported `actual: 0 vs 1`.

The same `PLAYING -> NULL` structure exists in five sibling tests that assert **nothing** about the buffer count. There the lost race is silent: `check_output`'s golden argmax comparison simply never runs and the test passes without verifying anything.

## Fix

Poll for the buffer between the two state transitions with the existing `wait_pipeline_process_buffers()` helper from `tests/unittest_util.h`, and assert the count. Buffer counting is done through a *second* `new-data` handler so `check_output` stays single-purpose.

| Test | Before | After |
| --- | --- | --- |
| onnxruntime `quantModelLaunch00` | asserts count, no wait → **flaky** | waits, asserts |
| onnxruntime `floatModelResult` | no count, no wait → **vacuous** | waits, asserts |
| tflite `quantModelResult` / `floatModelResult` | no count, no wait → **vacuous** | waits, asserts |
| litert `floatModelResult` / `quantModelResult` | no count, no wait → **vacuous** | waits, asserts |

## Evidence

The race is framework-independent, so it reproduces without any inference subplugin. A standalone program building the same pipeline shape (`filesrc ! pngdec ! videoconvert ! videoscale ! tensor_converter ! tensor_transform ! tensor_sink`), 60 iterations on a fully CPU-saturated machine:

```
WITHOUT wait: 46/60 iterations saw count==0
WITH    wait:  0/60 iterations saw count==0
```

That is both the reproduction of #4870 and the demonstration that the wait fixes it. It also quantifies the vacuous-pass hole: under load the golden check in the five sibling tests was skipped roughly 77% of the time.

## Test coverage

- The repaired assertions are themselves the regression tests — without the wait they fail under load (46/60 above), with it they pass.
- `wait_pipeline_process_buffers()` had **no coverage of its own** even though these tests now depend on its timeout contract. Added `commonUtil.waitPipelineProcessBuffers` (already-satisfied count returns immediately), `commonUtil.waitPipelineProcessBuffersAsync` (count reached from another thread is observed), and `commonUtil.waitPipelineProcessBuffers_n` (unreachable count returns `FALSE` after at least the timeout). A future change making the helper report success on timeout would otherwise quietly make every test above vacuous again — CI now fails instead.

## Verification performed locally

- `unittest_common`: **142/142 pass** (139 existing + 3 new).
- All four changed files compile with `-Wall -Wextra`, no new warnings; `wait_pipeline_process_buffers` resolves at link time.
- `clang-format --style=file`: clean on all four files.
- onnxruntime / tensorflow2-lite / litert are not installable in this environment, so those four test binaries were compile- and link-verified but not executed; CI provides the runtime signal.

## Scope

Test-only. No production code, no API, no architecture change — no documentation update applies. Split into two commits: the first is #4870 itself plus the helper coverage, the second is the identical hole in the tflite/litert siblings, so they can be evaluated independently.

### DIFF
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2677,6 +2677,70 @@ TEST (commonUtil, errorMessage)
 }
 
 /**
+ * @brief Thread body that bumps the counter after a short delay.
+ */
+static gpointer
+bump_counter (gpointer user_data)
+{
+  guint *count = (guint *) user_data;
+
+  g_usleep (TEST_DEFAULT_SLEEP_TIME * 3);
+  *count = 1U;
+
+  return NULL;
+}
+
+/**
+ * @brief Test that an already-satisfied count returns without waiting.
+ */
+TEST (commonUtil, waitPipelineProcessBuffers)
+{
+  guint count = 3U;
+  gint64 elapsed_us;
+  gint64 start_us = g_get_monotonic_time ();
+
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 3U, TEST_TIMEOUT_LIMIT_MS));
+
+  elapsed_us = g_get_monotonic_time () - start_us;
+  EXPECT_LT (elapsed_us, (gint64) TEST_DEFAULT_SLEEP_TIME);
+}
+
+/**
+ * @brief Test that the count reached from another thread is observed.
+ */
+TEST (commonUtil, waitPipelineProcessBuffersAsync)
+{
+  guint count = 0U;
+  GThread *worker = g_thread_new ("bump", bump_counter, &count);
+
+  ASSERT_TRUE (worker != NULL);
+
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
+  EXPECT_GE (count, 1U);
+
+  g_thread_join (worker);
+}
+
+/**
+ * @brief Test that a count which is never reached fails on timeout.
+ * @note The filter unit tests rely on FALSE here to catch a buffer that never
+ *       arrives; a helper that reported success would make them vacuous again.
+ */
+TEST (commonUtil, waitPipelineProcessBuffers_n)
+{
+  guint count = 0U;
+  const guint timeout_ms = 100U;
+  gint64 elapsed_ms;
+  gint64 start_us = g_get_monotonic_time ();
+
+  EXPECT_FALSE (wait_pipeline_process_buffers (&count, 1U, timeout_ms));
+
+  elapsed_ms = (g_get_monotonic_time () - start_us) / 1000;
+  EXPECT_GE (elapsed_ms, (gint64) timeout_ms);
+  EXPECT_LT (elapsed_ms, (gint64) TEST_TIMEOUT_LIMIT_MS);
+}
+
+/**
  * @brief Main function for unit test.
  */
 int
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2707,8 +2707,8 @@ TEST (commonUtil, waitPipelineProcessBuffers)
     best_us = MIN (best_us, g_get_monotonic_time () - start_us);
   }
 
-  /* One un-preempted run is enough to show the helper checks before sleeping.
-     Widening the bound instead would hide the very tick it has to detect. */
+  /** One un-preempted run is enough to show the helper checks before sleeping.
+   *  Widening the bound instead would hide the very tick it has to detect. */
   EXPECT_LT (best_us, (gint64) TEST_DEFAULT_SLEEP_TIME);
 }
 
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2685,7 +2685,7 @@ bump_counter (gpointer user_data)
   guint *count = (guint *) user_data;
 
   g_usleep (TEST_DEFAULT_SLEEP_TIME * 3);
-  *count = 1U;
+  g_atomic_int_set (count, 1U);
 
   return NULL;
 }
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2685,7 +2685,7 @@ bump_counter (gpointer user_data)
   guint *count = (guint *) user_data;
 
   g_usleep (TEST_DEFAULT_SLEEP_TIME * 3);
-  g_atomic_int_set (count, 1U);
+  g_atomic_int_inc (count);
 
   return NULL;
 }
@@ -2696,13 +2696,20 @@ bump_counter (gpointer user_data)
 TEST (commonUtil, waitPipelineProcessBuffers)
 {
   guint count = 3U;
-  gint64 elapsed_us;
-  gint64 start_us = g_get_monotonic_time ();
+  gint64 best_us = G_MAXINT64;
+  guint i;
 
-  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 3U, TEST_TIMEOUT_LIMIT_MS));
+  for (i = 0; i < 5U; ++i) {
+    gint64 start_us = g_get_monotonic_time ();
 
-  elapsed_us = g_get_monotonic_time () - start_us;
-  EXPECT_LT (elapsed_us, (gint64) TEST_DEFAULT_SLEEP_TIME);
+    EXPECT_TRUE (wait_pipeline_process_buffers (&count, 3U, TEST_TIMEOUT_LIMIT_MS));
+
+    best_us = MIN (best_us, g_get_monotonic_time () - start_us);
+  }
+
+  /* One un-preempted run is enough to show the helper checks before sleeping.
+     Widening the bound instead would hide the very tick it has to detect. */
+  EXPECT_LT (best_us, (gint64) TEST_DEFAULT_SLEEP_TIME);
 }
 
 /**
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -283,7 +283,8 @@ if gtest_dep.found()
       install_dir: unittest_install_dir
     )
 
-    test('unittest_filter_tensorflow2_lite', unittest_filter_tensorflow2_lite, env: testenv)
+    test('unittest_filter_tensorflow2_lite', unittest_filter_tensorflow2_lite,
+      timeout: 120, env: testenv)
   endif
 
   if have_python3
@@ -406,7 +407,7 @@ if gtest_dep.found()
       install_dir: unittest_install_dir
     )
 
-    test('unittest_filter_onnxruntime', unittest_filter_onnxruntime, env: testenv)
+    test('unittest_filter_onnxruntime', unittest_filter_onnxruntime, timeout: 120, env: testenv)
   endif
 
   # LiteRT unittest
@@ -418,7 +419,7 @@ if gtest_dep.found()
       install_dir: unittest_install_dir
     )
 
-    test('unittest_filter_litert', unittest_filter_litert, env: testenv)
+    test('unittest_filter_litert', unittest_filter_litert, timeout: 120, env: testenv)
   endif
 
   # LLAMACPP unittest
--- a/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
+++ b/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
@@ -137,6 +137,18 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 }
 
 /**
+ * @brief Signal to count invocations of tensor_sink
+ */
+static void
+count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  guint *count = (guint *) user_data;
+  UNUSED (element);
+  UNUSED (buffer);
+  (*count)++;
+}
+
+/**
  * @brief Check the litert subplugin is registered.
  */
 TEST (nnstreamerFilterLiteRT, checkExistence)
@@ -630,11 +642,16 @@ TEST (nnstreamerFilterLiteRT, floatModelResult)
   ASSERT_TRUE (sink_handle != nullptr);
 
   guint8 is_float = 1;
+  guint count = 0U;
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, &is_float);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
@@ -664,11 +681,16 @@ TEST (nnstreamerFilterLiteRT, quantModelResult)
   ASSERT_TRUE (sink_handle != nullptr);
 
   guint8 is_float = 0;
+  guint count = 0U;
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, &is_float);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
--- a/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
+++ b/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
@@ -138,6 +138,10 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 
 /**
  * @brief Signal to count invocations of tensor_sink
+ * @note Connect this after check_output. GObject runs same-stage handlers in
+ *       connection order, so a satisfied count implies the golden comparison
+ *       for that buffer has already run; connecting it first would let the
+ *       waiter tear the pipeline down before check_output had a say.
  */
 static void
 count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
@@ -145,7 +149,7 @@ count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   guint *count = (guint *) user_data;
   UNUSED (element);
   UNUSED (buffer);
-  (*count)++;
+  g_atomic_int_inc (count);
 }
 
 /**
--- a/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
+++ b/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
@@ -127,6 +127,10 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 
 /**
  * @brief Signal to count invocations of tensor_sink
+ * @note Connect this after check_output. GObject runs same-stage handlers in
+ *       connection order, so a satisfied count implies the golden comparison
+ *       for that buffer has already run; connecting it first would let the
+ *       waiter tear the pipeline down before check_output had a say.
  */
 static void
 count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
@@ -134,7 +138,7 @@ count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   guint *count = (guint *) user_data;
   UNUSED (element);
   UNUSED (buffer);
-  (*count)++;
+  g_atomic_int_inc (count);
 }
 
 /**
--- a/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
+++ b/tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc
@@ -126,6 +126,18 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 }
 
 /**
+ * @brief Signal to count invocations of tensor_sink
+ */
+static void
+count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  guint *count = (guint *) user_data;
+  UNUSED (element);
+  UNUSED (buffer);
+  (*count)++;
+}
+
+/**
  * @brief Negative test case with invalid model file path
  */
 TEST (nnstreamerFilterOnnxRuntime, openClose00)
@@ -379,28 +391,22 @@ TEST (nnstreamerFilterOnnxRuntime, floatModelResult)
   ASSERT_TRUE (sink_handle != nullptr);
 
   guint8 is_float = 1;
+  guint count = 0U;
 
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, &is_float);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
 
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
+
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
-}
-
-/**
- * @brief Signal to count invocations of tensor_sink
- */
-static void
-count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
-{
-  guint *count = (guint *) user_data;
-  UNUSED (element);
-  UNUSED (buffer);
-  (*count)++;
 }
 
 /**
@@ -434,6 +440,8 @@ TEST (nnstreamerFilterOnnxRuntime, quantModelLaunch00)
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
+
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
 
--- a/tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc
+++ b/tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc
@@ -111,6 +111,18 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 }
 
 /**
+ * @brief Signal to count invocations of tensor_sink
+ */
+static void
+count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  guint *count = (guint *) user_data;
+  UNUSED (element);
+  UNUSED (buffer);
+  (*count)++;
+}
+
+/**
  * @brief Negative case to launch gst pipeline: wrong dimension
  */
 TEST (nnstreamerFilterTensorFlow2Lite, launch0_n)
@@ -184,13 +196,19 @@ TEST (nnstreamerFilterTensorFlow2Lite, quantModelResult)
   ASSERT_TRUE (sink_handle != nullptr);
 
   guint8 *is_float = (guint8 *) g_malloc0 (1);
+  guint count = 0U;
   *is_float = 0;
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, is_float);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
 
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
+
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
@@ -224,13 +242,19 @@ TEST (nnstreamerFilterTensorFlow2Lite, floatModelResult)
   ASSERT_TRUE (sink_handle != nullptr);
 
   guint8 *is_float = (guint8 *) g_malloc0 (1);
+  guint count = 0U;
   *is_float = 1;
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, is_float);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
 
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
+
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
--- a/tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc
+++ b/tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc
@@ -112,6 +112,10 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 
 /**
  * @brief Signal to count invocations of tensor_sink
+ * @note Connect this after check_output. GObject runs same-stage handlers in
+ *       connection order, so a satisfied count implies the golden comparison
+ *       for that buffer has already run; connecting it first would let the
+ *       waiter tear the pipeline down before check_output had a say.
  */
 static void
 count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
@@ -119,7 +123,7 @@ count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
   guint *count = (guint *) user_data;
   UNUSED (element);
   UNUSED (buffer);
-  (*count)++;
+  g_atomic_int_inc (count);
 }
 
 /**


myungjoo:
**Note:** This review was produced by an AI coding-agent review pass (not a human maintainer) and is being posted verbatim on the human requester's behalf, as disclosed by their workflow policy.

## Summary

The PR fixes #4870 (`nnstreamerFilterOnnxRuntime.quantModelLaunch00` racing buffer arrival against pipeline teardown) by polling with the existing `wait_pipeline_process_buffers()` helper between `PLAYING` and `NULL`, and extends the identical treatment to five sibling tests (`onnxruntime.floatModelResult`, `tflite.{quant,float}ModelResult`, `litert.{float,quant}ModelResult`) that previously asserted nothing about buffer count and could pass vacuously. It also adds three new `commonUtil` unit tests that pin down `wait_pipeline_process_buffers()`'s own contract (immediate success, cross-thread success, and timeout failure).

I checked the diff against the linked issue, the pre-existing helper implementation in `tests/unittest_util.c`, the repo's test/commit conventions in `CLAUDE.md`, and `meson test` wiring in `tests/meson.build`. Findings below; nothing blocking.

## Findings

**1. [Minor] No `meson test timeout:` bump for the three modified filter-test binaries**

`tests/meson.build` gives `unittest_filter_custom` an explicit `timeout: 120` (line 238), which lines up with it already depending on `wait_pipeline_process_buffers()`. This PR adds the same up-to-10s wait (`TEST_TIMEOUT_LIMIT_MS`) to `unittest_filter_onnxruntime`, `unittest_filter_litert`, and `unittest_filter_tensorflow2_lite`, but their `test(...)` entries (lines ~286, ~409, ~421) still have no explicit `timeout:` and fall back to meson's default 30s. That default is fine in the healthy case (the wait returns almost immediately once a buffer arrives), but on the plain `meson test -C build/ -v` CI path (`ubuntu_clean_meson_build.yml`, `ubuntu_clean_llvm_build.yml`, no `--timeout-multiplier` passed), a genuine regression that makes two waited tests in the same binary each burn their full timeout could trip meson's own test timeout before gtest reports its per-case assertion. The test still fails and still blocks merge either way, so this doesn't reopen the vacuous-pass hole — it's a diagnostics-quality nit (TIMEOUT vs. a clear `EXPECT_TRUE` failure), not a correctness issue. Worth matching the `unittest_filter_custom` precedent for consistency, not required to merge.
(Note: the GBS/Tizen path that originally caught #4870, `packaging/run_unittests_binaries.sh`, has no timeout wrapper at all, so it isn't affected.)

**2. [Nit] Benign data race in the new `waitPipelineProcessBuffersAsync` test**

In `tests/common/unittest_common.cc`, `bump_counter()` writes a plain `guint *count` from a worker thread while the main thread polls the same `guint` via `wait_pipeline_process_buffers()` with no atomics/memory barrier. Technically a data race under the C++ memory model, but it's the exact idiom the helper is designed for and that the rest of the suite already relies on for real GStreamer callback threads vs. the main thread (that idiom is precisely what makes the underlying #4870 fix work). No TSan job exists in this repo's CI, so this isn't a practical risk — flagging only for completeness.

**3. [Nit / FYI, out of scope for this PR]**

`grep` for the same "`new-data` signal + immediate `PLAYING`→`NULL`, no wait" shape also turns up `tests/nnstreamer_filter_lua`, `tests/nnstreamer_filter_shared_model`, `tests/nnstreamer_filter_tvm`, `tests/nnstreamer_if`, `tests/gstreamer_join`, `tests/nnstreamer_edge`, and `tests/nnstreamer_datarepo`. The PR scopes itself correctly to the exact defect class named in its own description (the 5 siblings it fixes), so this isn't something it needs to cover — just noting it as a possible future follow-up, not a request to expand this PR.

## Checklist against the stated review criteria

- **Addresses the linked issue (#4870)?** Yes — `quantModelLaunch00` now polls `wait_pipeline_process_buffers(&count, 1U, TEST_TIMEOUT_LIMIT_MS)` between the two state transitions, exactly matching the issue's own root-cause analysis and suggested fix (it even reuses the issue's suggested helper instead of inlining a loop).
- **No regression to other modules?** Confirmed — diff touches only 4 files under `tests/`, no production/runtime code. `wait_pipeline_process_buffers()` itself is pre-existing (already used by `nnstreamer_converter`, `nnstreamer_filter_custom`, `nnstreamer_sink`, `tizen_sensor`) and is not modified by this PR.
- **Size proportionate / no unrelated modules touched?** Reasonable. The extension to tflite/litert siblings is the same defect class explicitly called out in the PR description and split into its own commit ("...so they can be evaluated independently"); the `unittest_common.cc` additions are direct coverage for a helper the fix now newly depends on for correctness, not scope creep.
- **Test coverage for the fix itself?** Yes, and it's a real regression test in the sense CLAUDE.md asks for: the repaired assertions themselves fail without the fix under load and pass with it (46/60 failures documented in the PR body without the wait, 0/60 with it).
- **Coverage protecting against future changes to code this depends on?** Yes — the 3 new `commonUtil.waitPipelineProcessBuffers*` tests pin the helper's success/timeout contract directly, which is exactly the dependency the fix newly introduces; a future change that made the helper report success on timeout would fail these before silently re-opening the vacuous-pass hole in the filter tests.
- **Will CI actually fail and block bad merges?** Yes for the assertion logic (see finding 1 for the one caveat about failure *presentation*, not detection).
- **Doc/architecture/API updates required?** No — test-only change, no production code, no API/architecture change. PR body correctly states this.
- **Style/process compliance:** Verified independently — `clang-format --style=file` reports zero replacements needed on all 4 changed files; every new function/test has a Doxygen `@brief`; new negative test uses the repo's existing `_n` suffix convention (matches other `commonUtil` and `nnstreamerFilterOnnxRuntime` cases in the same files); both commits carry `Signed-off-by` and follow the `[Component] Summary` subject format.

## Verdict

**Approve — OK to merge.** No Critical or Major issues found. The two Minor/Nit items above are optional polish, not blockers.

myungjoo:
Thanks — addressed both actionable points in `fe64a0e8`.

**Minor (meson timeout) — fixed.** Confirmed the concern is real: `unittest_filter_tensorflow2_lite`, `unittest_filter_onnxruntime` and `unittest_filter_litert` were all on meson's 30 s default, and tensorflow2-lite already spends 10 s of that in two fixed `g_usleep()` calls before any of the new waits. A regression that made several bounded waits run to completion would push the binary past 30 s and report a bare `TIMEOUT` instead of the gtest assertion naming the broken test — merging is blocked either way, but the diagnostic is much worse. All three now carry `timeout: 120`, matching `unittest_sink`, `unittest_filter_custom` and `unittest_filter_armnn`, which already have it for exactly this reason.

**Nit (data race in `waitPipelineProcessBuffersAsync`) — fixed on the write side.** The worker now uses `g_atomic_int_set()`. The read inside `wait_pipeline_process_buffers()` is deliberately left as-is: a plain load is the contract the real `tensor_sink` callbacks exercise, and changing it would mean the test no longer covers the production path.

**Nit/FYI (same pattern elsewhere) — deliberately not in scope.** Agreed that `nnstreamer_filter_lua`, `shared_model`, `tvm`, `nnstreamer_if`, `gstreamer_join`, `nnstreamer_edge` and `nnstreamer_datarepo` share the shape. Two reasons to leave them:

1. Those pipelines are not all single-buffer golden checks, so the fix is not the uniform two-line change it is here — each needs its own analysis of what "done" means.
2. Forcing a currently-skipped golden comparison to actually execute can surface a pre-existing failure. That is a real risk even within this PR's scope: the tflite/litert quant models could turn out to be CPU-dependent, which is the very class of problem #4865 fixed for onnxruntime. Absorbing seven more modules would make that risk unattributable.

The second commit here is separable for the same reason, so it can be dropped if the tflite/litert goldens misbehave on any runner.

`unittest_common` still passes 142/142 locally and `clang-format` is clean on all changed files.

myungjoo:
**Note:** This review was produced by an AI coding-agent review pass (not a human maintainer) and is being posted verbatim on the human requester's behalf, as disclosed by their workflow policy.

## Re-review (round 2) of `fe64a0e8`

This is a second-round review. I re-read the round-1 review comment and the author's response, then diffed `bd63ca15..HEAD` and inspected `fe64a0e8` (`tests/common/unittest_common.cc`, `tests/meson.build`) in full, plus re-verified the two prior commits (`c6d85df4`, `b0b9b5d3`) that round 1 already covered.

### Disposition of round-1 findings

| # | Round-1 finding | Severity | Status |
|---|---|---|---|
| 1 | No `timeout:` bump for `unittest_filter_onnxruntime` / `_litert` / `_tensorflow2_lite` | Minor | **Resolved.** `fe64a0e8` adds `timeout: 120` to all three `test(...)` entries in `tests/meson.build`, matching the existing `unittest_sink` / `unittest_filter_custom` / `unittest_filter_armnn` precedent line-for-line. Verified: `unittest_filter_tensorflow2_lite` already burns 2×5s in fixed `g_usleep()` calls (lines 296, 377) before any of the new waits, so worst case under a genuine regression (2 new 10s waits timing out + existing sleeps) plausibly exceeds the 30s meson default — the reasoning in the commit message checks out against the actual file. `packaging/run_unittests_binaries.sh` has no timeout wrapper and `ubuntu_clean_meson_build.yml` / `ubuntu_clean_llvm_build.yml` invoke plain `meson test -C build/ -v` with no `--timeout-multiplier`, so the 30s default was real exposure before this fix. |
| 2 | Benign data race in `waitPipelineProcessBuffersAsync` (`bump_counter` plain write vs. plain read in `wait_pipeline_process_buffers()`) | Nit | **Partially addressed, with a reasoned rationale for the remainder.** The write side (`tests/common/unittest_common.cc:2688`) now uses `g_atomic_int_set (count, 1U)`. The read side inside `wait_pipeline_process_buffers()` (`tests/unittest_util.c:199`, pre-existing, unmodified) is still a plain `*data_received < expected_num_buffers` load. The author's stated reason — that this plain-load contract is exactly what the real `tensor_sink` "new-data" counters (`count_output` in the three filter tests) exercise, and atomicizing only the test's synthetic read would decouple the helper's test coverage from its actual production usage pattern — is technically sound. No TSan job exists in this repo's CI, so residual risk is theoretical, not practical. I'd call this an accepted trade-off rather than a full fix, which is a fair outcome for a Nit. |
| 3 | Same "new-data + immediate PLAYING→NULL, no wait" shape exists in `nnstreamer_filter_lua`, `shared_model`, `tvm`, `nnstreamer_if`, `gstreamer_join`, `nnstreamer_edge`, `nnstreamer_datarepo` | Nit / FYI, explicitly out of scope | **Argument holds up.** The response's two reasons are concrete, not hand-waving: (a) those pipelines aren't uniformly single-buffer golden checks, so each needs its own analysis of "done," unlike the mechanical two-line change applied here; (b) forcing a currently-skipped golden comparison to actually run risks surfacing a pre-existing, possibly CPU-dependent golden mismatch — which is exactly the class of problem #4865 already had to fix for onnxruntime, and conflating that risk across seven more modules in one PR would make it hard to attribute a resulting CI failure to the right change. Keeping the tflite/litert commit separable (as already done) further limits blast radius. No objection to leaving this as a follow-up.

### Verification of the two items called out for extra scrutiny

- **`timeout: 120` correctness:** Confirmed via `grep` in `tests/meson.build` — the three new `timeout: 120` entries are placed on the exact same `test(...)` calls the PR modifies, formatting matches surrounding code (`clang-format`/meson style), and no other test's timeout was touched.
- **`g_atomic_int_set()` correctness/side effects:** `count` in `bump_counter()` is `guint *`, matching the type `g_atomic_int_set` is used with elsewhere in this codebase (e.g. `gst/mqtt/mqttsink.c`, on an enum-typed field) — this is an established idiom here and not a new pattern. `wait_pipeline_process_buffers()` itself (`tests/unittest_util.c`) is untouched, so its documented contract and its callers elsewhere in the suite (`nnstreamer_converter`, `nnstreamer_filter_custom`, `nnstreamer_sink`, `tizen_sensor`) are unaffected. `clang-format --style=file -n` reports zero diffs on all four touched `.cc` files.

### Fresh check across the full PR (3 commits, 5 files: `tests/common/unittest_common.cc`, `tests/meson.build`, `tests/nnstreamer_filter_litert/unittest_filter_litert.cc`, `tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc`, `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc`)

- **Addresses #4870?** Yes, unchanged from round 1 — `quantModelLaunch00` now polls before tearing down, matching the issue's root cause.
- **Regression risk to other modules?** None found. All changes are confined to `tests/`; no production/runtime code is touched; the shared helper `wait_pipeline_process_buffers()` is not modified, only newly covered by tests.
- **Size / scope proportionate?** Yes. The round-2 commit (`fe64a0e8`, 2 files, 5 lines net) is a tightly-scoped follow-up to exactly the two points raised in round 1.
- **Test coverage complete?** Yes — the repaired assertions are themselves the regression test for #4870, and the three new `commonUtil.waitPipelineProcessBuffers*` cases pin the shared helper's contract (immediate-success, cross-thread-success, timeout-failure), protecting every current and future consumer of that helper from a silent regression.
- **Will CI actually fail and block a bad merge?** Yes, and better than round 1: a regression now surfaces as a named gtest assertion failure within the 120s budget rather than a bare meson `TIMEOUT` that hides which test broke.
- **Docs/architecture/API impact?** None — test-only change, correctly stated as out of scope in the PR body.
- **Style/process compliance:** `clang-format --style=file -n` clean on all `.cc` files; `fe64a0e8` carries `Signed-off-by`; subject line follows `[Component] Summary`; new/changed functions retain Doxygen `@brief` comments.

## Findings

No Critical, Major, or new Minor/Nit issues found in the round-2 commit. Both round-1 items are resolved to a standard I'd consider adequate (one fully, one via an accepted intentional trade-off with a documented rationale); the scope-limiting decision on sibling modules is sound engineering judgment, not scope-avoidance.

## Verdict

**Approve — OK to merge.** Round-1 findings are closed out appropriately; no new issues identified in this re-review.

myungjoo:
## CI summary — all 18 checks green

The GBS x86_64 `unit_test 1` job is the one that originally caught #4870, and it is the only job in the matrix that builds onnxruntime, so it is the direct verification:

```
[ RUN      ] nnstreamerFilterOnnxRuntime.quantModelLaunch00
[       OK ] nnstreamerFilterOnnxRuntime.quantModelLaunch00 (58 ms)
[ RUN      ] nnstreamerFilterOnnxRuntime.floatModelResult
[       OK ] nnstreamerFilterOnnxRuntime.floatModelResult (93 ms)
[ RUN      ] nnstreamerFilterTensorFlow2Lite.quantModelResult
[       OK ] nnstreamerFilterTensorFlow2Lite.quantModelResult (108 ms)
[ RUN      ] nnstreamerFilterTensorFlow2Lite.floatModelResult
[       OK ] nnstreamerFilterTensorFlow2Lite.floatModelResult (66 ms)
[       OK ] commonUtil.waitPipelineProcessBuffers (0 ms)
[       OK ] commonUtil.waitPipelineProcessBuffersAsync (30 ms)
[       OK ] commonUtil.waitPipelineProcessBuffers_n (110 ms)
```

`unittest_filter_litert` also passed (5.25 s) in the Valgrind job, which has `LITERT_ROOT` configured.

Two things worth recording:

1. **The golden checks that were previously skipped now actually execute, and they pass.** The risk called out when this PR was opened was that forcing the tflite/litert quant goldens to run might surface a CPU-dependent mismatch — the #4865 problem class. It did not. The second commit stands on its own evidence rather than on assumption.
2. **The waits cost nothing when the pipeline is healthy.** 58–108 ms per test on an idle runner: `wait_pipeline_process_buffers()` returns as soon as the buffer arrives, so the bounded 10 s ceiling is only ever paid by a genuine failure.

### One unrelated failure, not fixed here

The first Valgrind (ubuntu-22.04) run failed on `NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p` at 10134 ms — its own `wait_for_samples()` 10 s timeout expiring. This is **not** caused by this PR:

- this PR touches five files, none under `tests/nnstreamer_filter_llamacpp/`;
- the failing assertion is llamacpp's local `wait_for_samples()` helper, not `wait_pipeline_process_buffers()`, which this PR adds tests for but does not modify;
- it is already tracked as #4883, filed before this PR;
- the job passed on re-run.

I deliberately did not fix it here. It is the same bug *class* as #4870, but the cause is different — llama.cpp inference being too slow under Valgrind rather than a teardown race — and the only fix available from here is a blind timeout bump in a module that cannot be built or tested locally, which could mask a real llama.cpp performance regression. It belongs in #4883.

myungjoo:
Thanks — this is a genuinely useful review. Addressed in `c8c7b980`. Taking each finding:

**1. Unsynchronised increment (low) — fixed.** `count_output` now uses `g_atomic_int_inc()`. The point that this branch propagated the pattern from one file to three is the part that makes it worth fixing, and the `unittest_filter_custom` precedent settles the question of what this repo considers correct here.

One clarification on scope: this makes the *write* atomic, not the whole access. The read inside `wait_pipeline_process_buffers()` is still a plain load, so this is not a complete removal of the formal race. I left the read alone deliberately — it is shared with every existing caller, including `unittest_filter_custom`, so changing it is a helper-wide decision rather than something this branch should make on the side.

**2. Test writer stricter than the real one (low) — fixed.** `bump_counter` now uses `g_atomic_int_inc()` too, so the async test drives the helper through exactly the call the filter tests use. As you predicted, this fell out of finding 1.

**3. Undocumented ordering (low) — fixed, and this was the best catch here.** The property really is load-bearing and really was implicit in six places. Recorded in `count_output`'s doxygen rather than at each call site, so it travels with the function:

```c
 * @note Connect this after check_output. GObject runs same-stage handlers in
 *       connection order, so a satisfied count implies the golden comparison
 *       for that buffer has already run; connecting it first would let the
 *       waiter tear the pipeline down before check_output had a say.
```

**4. Wall-clock upper bound (nit) — fixed, using the approach you named.** You were right not to want the budget widened, and right that it sits awkwardly in a de-flaking patch. The assertion now takes the best of five runs: one un-preempted run still proves the helper checks before sleeping, while a scheduling hiccup in any other run cannot fail it. The bound itself is unchanged at one tick. Verified on a fully CPU-saturated machine: **12 of 12 passes**.

**5a. `EXPECT_TRUE` + `EXPECT_GE` (nit) — keeping, deliberately.** Agreed it is worth being explicit that this is a choice. I would frame it as defence in depth rather than redundancy: if the helper ever returned `TRUE` incorrectly, `EXPECT_GE (count, 1U)` is the assertion that still catches it, and it is the one that states what the test is actually about. The `EXPECT_TRUE` exists for the failure message.

**5b. `is_float` leak (nit) — this one is incorrect.** It is freed in all three tensorflow2-lite tests:

- `quantModelResult` — allocated `unittest_filter_tensorflow2_lite.cc:198`, freed `:218`
- `floatModelResult` — allocated `:244`, freed `:264`
- `floatModelXNNPACKResult` — allocated `:290`, freed `:305`

No change made. Flagging it rather than silently skipping it.

**On the merge-risk note** — agreed, and worth restating plainly: these six tests can now fail where they previously passed vacuously, and a platform whose model silently produces nothing will now report a 10 s timeout instead of a fast green. That is the intended trade. Every CI configuration that builds these binaries is green at this head, so there is no live regression today.

Local verification after the change: `unittest_common` 142/142, all four changed files compile warning-free under `-Wall -Wextra`, `clang-format` clean. CI is re-running.

myungjoo:
**Note:** This review was produced by an AI coding-agent review pass (not a human maintainer) and is being posted verbatim on the human requester's behalf, as disclosed by their workflow policy.

## 3차 리뷰 (Round 3) — head `c8c7b980`

Round 1/2에서 나온 5개 finding에 대한 후속 커밋 `c8c7b980`("Increment the sink counters atomically and pin the handler order")을 검증하고, PR 전체(4 commits: `c6d85df4`, `b0b9b5d3`, `fe64a0e8`, `c8c7b980`)를 다시 점검했다. 요청받은 4가지 쟁점은 GLib/GObject 공식 문서와 실제 코드를 직접 대조해서 확인했다.

## 지정 쟁점 검증 결과

**1) `g_atomic_int_inc` on `guint *` — 적법함 (확인됨)**

GLib 공식 문서(`g_atomic_int_inc`)는 파라미터를 "a pointer to a gint or guint"라고 명시한다. C 선언은 `void g_atomic_int_inc (gint *atomic)`이지만, GCC/Clang(이 저장소 CI가 쓰는 모든 컴파일러)에서는 `gatomic.h`가 이를 `__atomic_fetch_add` 기반 매크로로 구현하므로 `guint *`를 넘겨도 문제없이 컴파일/동작한다. `count_output`의 `guint *count`에 대해 `g_atomic_int_inc(count)`를 쓰는 것은 hack이 아니라 문서화된 정상 사용법이다.

다만 "write만 atomic, read는 plain" 판단에 대해서는 — 엄밀한 C/C++ 메모리 모델 관점에서는 여전히 완전한 datarace-free는 아니다. `wait_pipeline_process_buffers()`(`tests/unittest_util.c:192-203`, 이번 PR에서 손대지 않음)는 `while (*data_received < expected_num_buffers) { g_usleep(...); ... }` 형태로 `data_received`를 plain load한다. atomic write는 하드웨어/컴파일러 full barrier를 제공하지만, reader 측에 atomic/volatile 연산이 없으므로 형식적으로는 여전히 data race다. 다만:
- `g_usleep()`이 opaque 외부 함수 호출이라 실제 컴파일러가 루프 밖으로 load를 hoist하기 어렵고,
- 이 repo에는 TSan 잡이 없으며,
- 해당 helper는 4곳의 다른 caller가 공유하는 pre-existing 코드라 read 쪽을 이 브랜치에서 바꾸는 건 스코프를 벗어난다.

즉 "완전한 수정"은 아니지만 test-only 코드에서의 실용적 트레이드오프로는 타당하다. **판단: 결론(코드는 legitimate)은 맞고, 트레이드오프 논거도 합리적이나, 형식적으로 100% data-race-free는 아님(Minor 수준 잔여 이슈, 신규 아님).**

**2) Finding 3 (연결 순서 문서화) — 정확함 (확인됨)**

GObject 공식 문서/사양을 확인한 결과: `g_signal_connect()`(AFTER 플래그 없이 연결된 핸들러)들은 "RUN_FIRST" 단계에서 **연결된 순서대로** 호출되는 것이 문서로 보장된다. 실제 코드에서도 6개 호출 지점(onnxruntime ×2, litert ×2, tflite2 ×2) 모두 `check_output`이 `count_output`보다 먼저 connect되어 있음을 직접 grep으로 확인했다(예: `unittest_filter_onnxruntime.cc:400-401`). 즉 `count_output`에 추가된 doxygen `@note`("Connect this after check_output. GObject runs same-stage handlers in connection order...")는 사실과 정확히 일치하며, 이 PR의 "no longer vacuous" 주장이 의존하는 숨은 전제를 정확하게 문서화했다. **적절한 수정.**

**3) Finding 4 (best-of-5 타이밍) — 테스트 속성을 약화시키지 않음 (확인됨)**

`waitPipelineProcessBuffers` 테스트는 이미 만족된 카운트(`count=3U`, `expected=3U`)로 helper를 호출해 "sleep 없이 즉시 반환하는가"를 검증한다. 이 fast-path는 결정론적 코드 경로이므로(외부 상태에 의존하지 않음), 구현이 정말로 깨져서 "항상 한 번 sleep 후 반환"하도록 바뀐다면 5번의 반복 모두 10ms 이상 걸리고, `MIN`을 취해도 여전히 실패한다. 즉 best-of-5는 스케줄러 preemption에 의한 flaky 실패만 걸러낼 뿐, 실제 회귀를 탐지하는 능력은 그대로 유지된다. 5회 루프 안에서 `count` 값이 재사용되며 부작용이 없는 것도 확인했다. **타당한 수정이며 속성을 약화시키지 않는다.**

**4) Finding 5b (`is_float` 누수) — 작성자 반박이 맞음, bot 리뷰가 틀렸음 (확인됨)**

`tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc`를 직접 확인:
- `quantModelResult`: `g_malloc0` at line 202, `g_free(is_float)` at **line 222**
- `floatModelResult`: `g_malloc0` at line 248, `g_free(is_float)` at **line 268**
- `floatModelXNNPACKResult`: `g_malloc0` at line 294, `g_free(is_float)` at **line 309**

세 테스트 모두 `is_float`가 free된다. round-2 리뷰의 finding 5b는 잘못된 지적이었고, 작성자가 "No change made"로 대응한 것이 옳다.

**5) Finding 5a (`EXPECT_TRUE` + `EXPECT_GE` 중복 유지) — 판단 근거 타당**

"defense in depth" 논거는 합리적이다. `wait_pipeline_process_buffers()`가 잘못된 이유로 `TRUE`를 반환하는 경우를 `EXPECT_GE(count, 1U)`가 별도로 잡아낼 수 있으므로, 두 assertion이 실제로 서로 다른 실패 모드를 커버한다. 스타일 취향 문제이며 반대할 근거가 약하다. **수용 가능.**

## 신규 발견 — round 1/2에서 다루지 않은 문제

**[Critical] `c8c7b980`이 CI의 `Static checks` (Doxygen tag check)를 깨뜨리고 있음 — 현재 머지 불가**

`gh pr checks 4889` 확인 결과 `Static checks` 잡이 **fail** 상태다(round 2 리뷰 시점의 "18/18 green"은 `fe64a0e8` 기준이었고, 그 이후 커밋 `c8c7b980`이 새로 CI를 깨뜨렸다). 로그:

```
[ERROR] File name: tests/common/unittest_common.cc, 2710 line, Doxygen or multi line comments should begin with /**
::error There is a doxygen tag missing or incorrect.
```

원인은 `tests/common/unittest_common.cc:2710-2711`에 이번 커밋이 새로 추가한 2줄짜리 `/* ... */` 블록 주석:

```c
  /* One un-preempted run is enough to show the helper checks before sleeping.
     Widening the bound instead would hide the very tick it has to detect. */
```

`.github/workflows/static.check.scripts/doxygen-tag.sh:124`의 규칙은 "`/*`로 시작하되 `/**`가 아니고, 같은 줄에서 닫히지 않는(또는 `@`를 포함하는) 모든 멀티라인 주석"을 오류로 잡는다 — 함수 doxygen 블록뿐 아니라 본문 내 설명 주석에도 동일하게 적용되는, 이 저장소 특유의 엄격한 규칙이다. 이번 diff 전체에서 이 패턴에 걸리는 곳은 이 한 곳뿐이다.

**심각도: Critical** — CI가 실제로 빨간불이며 (`Static checks` fail), 현재 상태로는 머지가 불가능하다. 다만 수정 자체는 사소하다 — 예를 들어 `//` 라인 주석 두 줄로 바꾸거나 한 줄짜리 `/* ... */`로 합치면 통과한다:

```c
  // One un-preempted run is enough to show the helper checks before sleeping.
  // Widening the bound instead would hide the very tick it has to detect.
```

## PR 전체 점검 (4 commits)

- **#4870 해결 여부:** Yes. `quantModelLaunch00`이 `PLAYING`→`NULL` 사이에 `wait_pipeline_process_buffers()`로 폴링하도록 수정되어 이슈의 근본 원인(버퍼 도착과 teardown의 race)을 정확히 해소한다.
- **타 모듈 regression 가능성:** 없음. 5개 파일 모두 `tests/` 하위이며 production/runtime 코드는 전혀 건드리지 않는다. 공유 헬퍼 `wait_pipeline_process_buffers()` 자체는 미수정.
- **코드 크기/스코프:** 적절함. tflite/litert 형제 테스트로의 확장은 커밋을 분리해 독립적으로 revert 가능하게 했고, sibling 모듈(lua, tvm, edge, datarepo 등)로의 확장은 의도적으로 배제하며 그 근거(golden 값이 CPU-dependent할 위험, #4865 참조)도 타당하다.
- **테스트 커버리지 완비 여부 / 향후 회귀 감지:** 양호. 수정된 assertion 자체가 #4870의 회귀 테스트 역할을 하고, 신규 `commonUtil.waitPipelineProcessBuffers*` 3종이 공유 헬퍼의 계약(즉시성공/비동기성공/타임아웃)을 고정해, 헬퍼가 미래에 깨져도 필터 테스트보다 먼저 잡아낸다.
- **CI가 실제로 fail해서 merge를 막을 수 있는지:** 로직적으로는 Yes (assertion들이 실질적 회귀 테스트). 다만 위에서 지적한 대로 **현재 `c8c7b980` 자체가 Static checks를 깨뜨려 이미 CI가 빨간불**이다 — 별개 이슈지만 지금 이 상태로는 어차피 머지 게이트를 통과하지 못한다.
- **문서/아키텍처/API 변경 필요 여부:** 없음. Production 코드/API/ABI 변경이 전혀 없는 test-only PR이므로 문서 수정 불요, PR 설명도 이를 정확히 밝히고 있다.

## Round 1/2 finding 처리 현황 요약

| # | Finding | 심각도 | 상태 |
|---|---|---|---|
| Round1-1 | `unittest_filter_*` 3개 바이너리에 `timeout:` 없음 | Minor | 해결 (`fe64a0e8`에서 `timeout: 120` 추가, 검증됨) |
| Round1-2 | `waitPipelineProcessBuffersAsync`의 benign data race | Nit | 해결 (Round2에서 write는 atomic화; 이번 라운드 검증 결과 read는 여전히 plain — 의도된 트레이드오프로 수용 가능) |
| Round1-3 | 같은 패턴이 lua/tvm/edge/datarepo 등에도 존재 | Nit/FYI, 명시적 out-of-scope | 반박 타당 (스코프 유지 근거가 구체적이고 합리적) |
| Round2-1 | `count_output`의 unsynchronised increment | Low | **해결** (`c8c7b980`에서 `g_atomic_int_inc` 적용 — GLib 문서상 legitimate, 검증됨) |
| Round2-2 | `bump_counter`가 실제 caller보다 엄격한 writer 사용 | Low | **해결** (`c8c7b980`에서 `bump_counter`도 `g_atomic_int_inc`로 통일) |
| Round2-3 | 연결 순서가 undocumented | Low | **해결** (`c8c7b980`에서 doxygen `@note` 추가 — 내용이 GObject 실제 동작과 일치함을 확인) |
| Round2-4 | wall-clock upper bound가 flaky할 수 있음 | Nit | **해결** (`c8c7b980`에서 best-of-5로 변경 — 회귀 탐지력 유지하며 flakiness만 제거, 검증됨) |
| Round2-5a | `EXPECT_TRUE`+`EXPECT_GE` 중복 | Nit | 유지 결정, 근거 타당 (별개 실패 모드 커버) |
| Round2-5b | `is_float` 누수 주장 | Nit | **반박 타당 — bot 리뷰가 오류.** 세 테스트 모두 실제로 `g_free`됨을 라인 단위로 직접 확인 |

## 정리된 지적사항 (이번 라운드 신규)

- **[Critical]** `tests/common/unittest_common.cc:2710-2711`의 신규 2줄 `/* */` 블록 주석이 `Static checks`(Doxygen tag check) CI 잡을 깨뜨림. 현재 head(`c8c7b980`)는 이 상태로 머지 불가. 수정은 `//` 라인 주석 전환 등 한 줄짜리 변경으로 충분.

이 외에 이번 라운드에서 새로 발견된 Major/Minor 이슈는 없음. Round 2까지의 모든 지적사항은 위 표와 같이 해결되었거나(대부분) 반박이 타당하게 받아들여졌다.

## Merge 판단

**Changes requested.** 로직/테스트 설계 자체는 3라운드에 걸쳐 충분히 검증되었고 (#4870 근본 원인 수정, 형제 테스트 확장, 헬퍼 계약 고정, 이전 5개 finding 모두 적절히 처리) 문제 없다. 그러나 최신 커밋 `c8c7b980`이 이 저장소의 `Static checks`(Doxygen tag) CI 게이트를 실제로 깨뜨리고 있어(`tests/common/unittest_common.cc:2710`), 이 한 줄을 고치기 전까지는 머지할 수 없다. 수정은 사소하며(주석 스타일 변경), 반영 후 CI가 다시 녹색이 되면 추가 리뷰 없이 머지 가능한 수준이라고 판단한다.

myungjoo:
The Critical finding is correct and the fix is in `f1ba7f59`. My mistake, and CI caught it exactly as it should have.

The two-line explanation I added next to the timing bound in `c8c7b980` opened with a plain `/*`. This repo's checker requires *every* multi-line comment to begin with `/**`, not just function headers:

```
[ERROR] tests/common/unittest_common.cc, 2710 line,
        Doxygen or multi line comments should begin with /**
```

Now using the `/**` form already used for in-body explanations elsewhere in the tree, e.g. the max-search note in the litert filter test.

Two things I did rather than just applying the one-line change:

**Checked whether I had done this anywhere else on the branch.** Scanned every added line across all five commits for a multi-line comment opening with a bare `/*` — line 2710 was the only one.

**Added the checker to my local loop, and confirmed it is not a no-op.** I now run `.github/workflows/static.check.scripts/doxygen-tag.sh` over the changed files before pushing. As a negative control I re-introduced the exact bad comment and confirmed the script reproduces CI's failure at line 2710, then confirmed it reports nothing on the fixed content. Running clang-format was not sufficient here — it is happy with either opener, which is why this slipped through.

Verification at this head: doxygen-tag checker clean on all four changed files, `unittest_common` 142/142, compiles warning-free under `-Wall -Wextra`, clang-format clean.

Also noted your independent confirmations, which are worth having on the record: `g_atomic_int_inc()` on a `guint *` is documented-legal; GObject really does invoke same-stage handlers in connection order, so the doxygen note in finding 3 states a real guarantee rather than an assumption; the best-of-5 sampling preserves detection because a genuine regression is deterministic across all five runs; and the `is_float` leak claim from the previous round was indeed incorrect.

CI is re-running.

myungjoo:
**Note:** This review was produced by an AI coding-agent review pass (not a human maintainer) and is being posted verbatim on the human requester's behalf, as disclosed by their workflow policy.

## 4차 리뷰 (Round 4) — head `f1ba7f59`

Round 3의 결론은 **Changes requested**였고, 사유는 커밋 `c8c7b980`이 `Static checks`(Doxygen tag check) CI를 깨뜨린 것 하나였다(`tests/common/unittest_common.cc:2710`, bare `/*` multi-line 주석). 작성자가 이를 `f1ba7f59`로 수정했다고 보고했다. 이번 라운드는 그 수정을 독립적으로 검증하고, PR 전체(5 commits)를 최종 점검한다.

## 이번 라운드 지정 쟁점 검증 결과

**1) `f1ba7f59`가 Round 3의 Critical을 실제로 해결했는가— 예 (직접 확인)**

`gh pr checks 4889` 결과 **18/18 전부 pass**, `Static checks`도 33초에 pass. 육안 확인에 그치지 않고 체커 스크립트 자체를 재현했다: `.github/workflows/static.check.scripts/doxygen-tag.sh`를 WSL에서 advanced 모드(`$2=1`, 실제 CI 호출과 동일 — `static.check.yml:85`)로 PR이 변경한 4개 파일에 대해 직접 실행한 결과 `[ERROR]` 라인 0개, exit 0. `f1ba7f59`의 diff는 문제의 2줄을:

```c
-  /* One un-preempted run is enough to show the helper checks before sleeping.
-     Widening the bound instead would hide the very tick it has to detect. */
+  /** One un-preempted run is enough to show the helper checks before sleeping.
+   *  Widening the bound instead would hide the very tick it has to detect. */
```

로 바꾼 것이 유일한 변경이며, 이 한 줄 변경이 정확히 체커가 요구하는 조건(`/*`로 시작하지만 `/**`가 아니고 같은 줄에서 닫히지 않는 멀티라인 주석 금지, `doxygen-tag.sh` line 124)을 충족한다. Static checks job 로그(`98734224142`, 33s, pass)도 직접 대조해 이견 없음.

또한 CI 로그를 직접 열람하여(Tizen GBS x86_64 `unit_test 1`, run `33135442901`/job `98734224262`) 이번 head에서 실제 테스트 실행 결과를 확인했다:

```
[ RUN ] nnstreamerFilterOnnxRuntime.quantModelLaunch00
[ OK ] nnstreamerFilterOnnxRuntime.quantModelLaunch00 (43 ms)
[ RUN ] nnstreamerFilterOnnxRuntime.floatModelResult
[ OK ] nnstreamerFilterOnnxRuntime.floatModelResult (92 ms)
[ OK ] commonUtil.waitPipelineProcessBuffers (0 ms)
[ OK ] commonUtil.waitPipelineProcessBuffersAsync (30 ms)
[ OK ] commonUtil.waitPipelineProcessBuffers_n (110 ms)
[ OK ] nnstreamerFilterTensorFlow2Lite.quantModelResult (107 ms)
[ OK ] nnstreamerFilterTensorFlow2Lite.floatModelResult (66 ms)
```
같은 잡의 로그 전체에서 `[  FAILED  ]` 0건. #4870의 근본 수정이 실제로 CI에서 검증되고 있음을 재확인.

**2) "브랜치 전체에서 line 2710이 유일했다"는 주장 — 사실 (독립 재현)**

`git diff bd63ca15..HEAD -- tests/`에서 추가된(`+`) 라인 158개 전체를 grep으로 스캔한 결과, `/*`를 포함하는 라인은 8곳뿐이며 전부 `/**`로 시작한다(`unittest_common.cc`의 새 `commonUtil` 테스트 3개의 doxygen 블록, `count_output`의 `@note` 블록 3개 — onnxruntime/litert/tflite2 각 1개, 그리고 이번에 고친 라인). 저자의 스캔 주장과 정확히 일치하며, 다른 잔여 위반은 없다.

**3) `doxygen-tag.sh` 규칙 파악 및 4개 변경 파일 전체 만족 여부 — 확인**

스크립트를 직접 읽었다(`.github/workflows/static.check.scripts/doxygen-tag.sh`). 핵심 규칙(advanced 모드, line 124 부근):

```bash
if [[ $line =~ "/*" && $line != *"/**"*  && ( $line != *"*/"  || $line =~ "@" ) && ( $idx != 1 ) ]]; then
    echo "[ERROR] ... Doxygen or multi line comments should begin with /**"
```
즉 "`/*`로 시작하지만 `/**`는 아니고, 같은 줄에서 `*/`로 닫히지 않거나 `@` 태그를 포함하는" 모든 줄에 에러. 파일 첫 줄(`$idx != 1`)은 제외. 이 규칙은 함수/구조체 헤더 doxygen 블록에 국한되지 않고 본문 내 임의의 멀티라인 주석에도 적용된다 — 저자의 커밋 메시지 설명과 일치. `advanced=1`은 `static.check.yml:85`에서 `bash .../doxygen-tag.sh $changed_file_list 1`로 실제 호출되고 있음을 확인, 즉 이 규칙이 실제로 이 PR에 적용되는 것이 맞다.

앞서 1)에서 실제 스크립트를 4개 변경 파일에 직접 실행해 에러 0건을 확인했으므로, 규칙 충족 여부는 추정이 아니라 재현된 사실이다.

**4) 새 `/**` 주석 형식이 기존 in-body 주석 컨벤션과 일치하는가 — 예 (기존 사례와 대조)**

저자가 언급한 "litert filter test의 max-search note"를 직접 확인했다(`tests/nnstreamer_filter_litert/unittest_filter_litert.cc:118-119`, PR과 무관한 기존 코드):

```c
    /** -G_MAXFLOAT: G_MINFLOAT is the smallest positive normalized value,
     *  which would break the max search for all-negative outputs */
```

`f1ba7f59`가 만든 새 주석과 들여쓰기·연속행 `*` 정렬 방식이 동일하다(`/**` 다음 줄에서 `*`가 `/**`의 두 번째 `*` 아래로 정렬). 저장소의 기존 in-body doxygen 스타일 컨벤션과 형식이 일치한다.

## PR 전체 최종 점검 (5 commits, `bd63ca15..HEAD`)

- **#4870 해결 여부:** Yes. `quantModelLaunch00`이 `PLAYING`→`NULL` 사이에 `wait_pipeline_process_buffers()`로 폴링하며, 위 CI 로그로 실제 통과를 재확인.
- **타 모듈 regression 가능성:** 없음. 최종 diff는 `tests/` 하위 5개 파일(`unittest_common.cc`, `meson.build`, litert/onnxruntime/tflite2 필터 테스트)에 한정, production/runtime 코드 변경 없음. 공유 헬퍼 `wait_pipeline_process_buffers()` 자체는 미수정.
- **코드 크기/스코프 적절성:** 적절함. Round 1~3에서 이미 검증된 대로 커밋이 목적별로 분리되어 있고(#4870 자체 수정 / sibling 확장 / meson timeout / 원자성+문서화 / doxygen 수정), sibling 모듈(lua, tvm, edge 등)로의 확장은 의도적·근거 있게 배제됨.
- **테스트 커버리지 완비 / 향후 회귀 감지:** 양호. 수정된 assertion 자체가 회귀 테스트 역할을 하고, `commonUtil.waitPipelineProcessBuffers*` 3종이 공유 헬퍼의 계약을 고정한다. `count_output`에 추가된 `@note`가 "check_output을 먼저 connect해야 wait가 non-vacuous하다"는 숨은 전제를 문서화(Round 2/3에서 GObject 실제 동작과 대조 검증됨).
- **CI가 실제로 fail하여 merge를 막을 수 있는지:** Yes — 18/18 전부 pass 상태이며, 이번에 직접 실행한 doxygen-tag 체커와 CI 로그 열람으로 재현됨.
- **문서/아키텍처/API 변경 필요 여부:** 없음. Test-only PR, production 코드/API/ABI 변경 없음 — 정확히 명시됨.

## Round 1~3 finding 최종 처리 현황

| # | Finding | 심각도 | 최종 상태 |
|---|---|---|---|
| R1-1 | `unittest_filter_*` 3개 바이너리에 `timeout:` 없음 | Minor | 해결 (`fe64a0e8`에서 `timeout: 120` 추가, 최종 diff에서도 유지 확인) |
| R1-2 | `waitPipelineProcessBuffersAsync`의 benign data race | Nit | 부분 해결 — write는 atomic(`g_atomic_int_inc`), read는 여전히 plain load. 의도된 트레이드오프로 3라운드에 걸쳐 수용됨(TSan job 없음, helper는 4곳 공유 코드라 read 변경은 스코프 밖) |
| R1-3 | 같은 패턴이 lua/tvm/edge/datarepo 등에도 존재 | Nit/FYI, out-of-scope | 반박 타당, 유지 |
| R2-1 | `count_output`의 unsynchronised increment | Low | 해결 (`c8c7b980`, `g_atomic_int_inc` — GLib 문서상 legitimate 사용, 검증됨) |
| R2-2 | `bump_counter`가 실제 caller보다 엄격한 writer 사용 | Low | 해결 (`c8c7b980`) |
| R2-3 | 연결 순서(check_output → count_output)가 undocumented | Low | 해결 (`c8c7b980`, `@note` 추가 — GObject 실제 동작과 일치 확인) |
| R2-4 | wall-clock upper bound가 flaky할 수 있음 | Nit | 해결 (`c8c7b980`, best-of-5 — 회귀 탐지력 유지 확인) |
| R2-5a | `EXPECT_TRUE`+`EXPECT_GE` 중복 | Nit | 유지 결정, 근거 타당 |
| R2-5b | `is_float` 누수 주장 | Nit | 반박 타당 — 세 테스트 모두 실제 `g_free`됨, bot 리뷰 오류로 확인됨 |
| **R3** | **`c8c7b980`이 Doxygen tag check(Static checks)를 깨뜨림 — line 2710** | **Critical** | **해결 (`f1ba7f59`). 이번 라운드에서 doxygen-tag.sh 실행 재현 + CI 18/18 green으로 독립 검증 완료** |

이번 라운드에서 새로 발견된 Critical/Major/Minor/Nit 이슈는 없음.

## Merge 판단

**Approve — OK to merge.** 4차 검토 완료. Round 3의 유일한 Critical(Static checks 실패)이 `f1ba7f59`로 실제로 해결되었음을 체커 스크립트 재실행과 CI 로그 직접 열람으로 독립 검증했고, 나머지 모든 finding은 이전 라운드들에서 적절히 처리되었다. 로직·테스트 설계·스코프·회귀 방지·CI 게이트 효과성 모두 4라운드에 걸쳐 충분히 검증되었으며, 현재 head(`f1ba7f59`)에서 추가로 요구할 사항 없음.

PR #4892, opened 2026-08-27
[Test/Plugins] Type-correct orc 64-bit buffers for macOS

### LINKED ISSUE #4862: unittest_plugins.cc orc tests do not compile on macOS (orc_int64 is long, int64_t is long long)
## Problem

The orc test code in `tests/nnstreamer_plugins/unittest_plugins.cc` does not compile on macOS: orc defines `orc_int64` as `long`, while macOS defines `int64_t` as `long long`. Both are 64-bit, but they are distinct types to the compiler, so pointer conversions between them are errors.

This is the reason the upcoming macOS CI work (#4855) has to build with `-Dorcc-support=disabled` on macOS instead of testing the orc-accelerated paths there.

## Where

`unittest_plugins.cc` orc test cases (the `HAVE_ORC` sections). The production code paths in `gst/nnstreamer/elements/gsttensor_transform.c` are fine; only the test code has the type mismatch.

## Suggested direction

Either cast through the orc typedefs explicitly in the test helpers, or use fixed-width buffers with `reinterpret_cast`/`memcpy` at the orc API boundary so the tests are type-correct on both LP64 conventions.

## Context

Split out of the review of #4858 so it is tracked instead of living in a PR description. Once fixed, `-Dorcc-support=disabled` can be dropped from the macOS CI job and orc paths can be tested on macOS as well.

myungjoo:
## Problem

`tests/nnstreamer_plugins/unittest_plugins.cc` does not compile on macOS. Its
orc test cases declare the 64-bit buffers as `int64_t`/`uint64_t` and pass them
straight to the orc entry points, whose prototypes take `orc_int64`/`orc_uint64`.

orcc emits those typedefs into the generated header itself, and their C99 branch
is guarded by `__STDC_VERSION__`, which is undefined in C++. On LP64 the fallback
branch is taken:

```c
#if INT_MAX == LONG_MAX
typedef long long orc_int64;
#else
typedef long orc_int64;      /* <- taken on LP64 */
#endif
```

So in C++ `orc_int64` is `long`. On Linux `int64_t` is also `long`, which is why
this never showed up; on macOS `int64_t` is `long long`, a distinct type, and the
pointer conversion is a hard error.

Reproduced locally against the generated header (Linux, where `long long` is just
as distinct from `long` as it is on macOS):

```
error: invalid conversion from 'long long int*' to 'orc_int64*' {aka 'long int*'}
    8 |   nns_orc_add_c_s64 (s, 1, 4);
```

The production paths in `gst/nnstreamer/elements/gsttensor_transform.c` are not
affected: they cast to `gpointer` at the orc boundary, and they compile as C,
where orcc's typedefs do resolve to the fixed-width types.

## Change

- Derive the orc test buffer types from the orc prototypes themselves, by
  deducing the element type through a declared-only helper (`test_orc_elem_type`,
  which sidesteps having to spell out `ORC_RESTRICT`). Naming `orc_int64` directly
  would fix the build just as well and in fewer lines; deducing instead keeps the
  buffers tied to the prototypes rather than to a second name that has to be kept
  in sync with them, so they stay correct if orcc ever emits the parameter as the
  type declared in the `.orc` file rather than as its own typedef. No casts, no
  aliasing through a differently-spelled 64-bit type.
- Drop `-Dorcc-support=disabled` from the macOS CI job. It now passes
  `-Dorcc-support=enabled` rather than relying on auto-detection, so the coverage
  cannot silently vanish if orc leaves the runner image, and `orc` is added to
  the brew install list to guarantee `orcc` is present.
- Run the orc test cases on macOS, together with the `tensor_transform`
  `typecast_*_accel` and `arithmetic*Accel` cases that route through the orc entry
  points, and `properties01`. That last one compares the element's `acceleration`
  default against the macro the test derives from its *own* `HAVE_ORC`, so it does
  not fail merely because the default is FALSE -- what it catches is the element
  and the test being built with orc disagreeing, which is a new possibility now
  that both are built with orc on macOS.
- A gtest filter that matches nothing, or that quietly stops matching part of the
  set, still exits 0, so the step also requires the expected number of tests to
  have actually run.

## Verification

Local (WSL Ubuntu, orc 0.4.32):

- Reproduced the failure and confirmed the `orc_int64` form compiles, by
  compiling both spellings against the generated `nnstreamer-orc.h`.
- Confirmed the deduction really binds the buffers: all 26 declarations use the
  deduced typedefs and no `int64_t`/`uint64_t` array declaration remains in the orc
  section. The surviving `static_assert`s cover what deduction cannot -- a width or
  signedness change on the orc side -- rather than restating it.
- Built `unittest_plugins` with both GCC 11 and Clang 14 under the project's
  `werror=true` / `warning_level=2` defaults, with `-Dorcc-support=enabled`;
  no new warnings. Clang also compiles the orc-generated C cleanly, which is the
  relevant precedent for macOS.
- `meson test -C build`: 19/19 suites OK, `unittest_plugins` 138/138 tests pass.
- The filtered command the macOS job runs selects 34 tests (14 orc + 14 typecast
  accel + 5 arithmetic accel + `properties01`), all passing. The count guard was
  verified in all three directions: 34 passes, a filter selecting nothing (0) fails,
  and a filter selecting only part of the set (14) fails.
- `clang-format` clean under both v14 and the v16 used by the cpp-linter job.

## Notes

The macOS job also gains the orc code path in `gsttensor_transform.c`, which means
`acceleration` now defaults to TRUE in macOS builds, matching Linux. That is the
intent of the issue, and the tests the job now runs cover it.

On the limits of the guard: new orc test code written with `int64_t`/`uint64_t`
would still compile on Linux and still break on macOS. No construct in the file
can catch that on Linux, because on Linux those buffers really are correct. The
macOS build is the guard, which is why enabling orc there is part of this change
rather than a follow-up.

Fixes #4862

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -28,7 +28,7 @@ jobs:
           brew install cask
     - name: install minimal requirements
       if: env.rebuild == '1'
-      run: brew install meson ninja pkg-config cmake libffi glib gstreamer numpy json-glib googletest
+      run: brew install meson ninja pkg-config cmake libffi glib gstreamer numpy json-glib googletest orc
     - name: install LiteRT SDK
       if: env.rebuild == '1'
       run: |
@@ -69,18 +69,30 @@ jobs:
         echo "LITERT_ROOT=$LITERT_ROOT" >> $GITHUB_ENV
     - name: meson build
       if: env.rebuild == '1'
-      # orcc-support is disabled: the orc test code in unittest_plugins.cc
-      # does not compile on macOS (orc_int64 is long while int64_t is
-      # long long); it never has, hidden until now by the check-rebuild bug
-      # that skipped every step of this job.
+      # orcc-support is required rather than auto-detected: building the orc
+      # paths here is what keeps the orc test code compiling on macOS, where
+      # int64_t is long long while orc_int64 is long. Auto-detection would
+      # silently drop that coverage if orc went missing from the image.
       run: |
-        meson setup build -Dorcc-support=disabled
+        meson setup build -Dorcc-support=enabled
         meson compile -C build
     - name: check litert subplugin built
       if: env.rebuild == '1'
       run: |
         ls build/ext/nnstreamer/tensor_filter/ | grep litert
         test -f build/ext/nnstreamer/tensor_filter/libnnstreamer_filter_litert.dylib || test -f build/ext/nnstreamer/tensor_filter/libnnstreamer_filter_litert.so
+    - name: run tensor_transform orc unit tests
+      if: env.rebuild == '1'
+      # the orc entry points directly, the typecast and arithmetic cases that
+      # route through them when acceleration is on, and properties01, which
+      # compares the element's acceleration default against the test's own
+      # HAVE_ORC and so catches the two being built with orc disagreeing
+      run: |
+        meson test -C build unittest_plugins -v --test-args='--gtest_filter=testTensorTransform.orc*:testTensorTransform.typecast_*_accel:testTensorTransform.arithmetic*Accel:testTensorTransform.properties01'
+        # a filter that matches nothing, or that quietly stops matching some of
+        # the cases, still exits 0; require the whole set to have run
+        ran=$(sed -nE 's/^\[  PASSED  \] ([0-9]+) tests?\.$/\1/p' build/meson-logs/testlog.txt | tail -1)
+        test "${ran:-0}" -ge 34
     - name: run litert unit tests
       if: env.rebuild == '1'
       run: |
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -3772,7 +3772,39 @@ TEST (testTensorConverter, flexToStaticInvalidBuffer2_n)
 }
 
 #ifdef HAVE_ORC
+#include <type_traits>
+
 #include "nnstreamer-orc.h"
+
+/**
+ * @brief Deduce the element type an orc constant-operand function works on.
+ *
+ * Declared only, and used in unevaluated context, to recover the buffer type
+ * from an orc prototype without spelling out its ORC_RESTRICT qualifier.
+ */
+template <typename T> T test_orc_elem_type (void (*orc_func) (T *, int, int));
+
+/**
+ * @brief Buffer element types for the 64-bit orc test cases.
+ *
+ * orcc emits its own orc_intNN typedefs, and their C99 branch is not taken in
+ * C++, where __STDC_VERSION__ is undefined, so on LP64 orc_int64 falls back to
+ * long. That is the same type as int64_t on Linux but not on macOS, where
+ * int64_t is long long; the two are distinct types to the compiler even though
+ * both are 64 bits wide, so buffers spelled int64_t cannot be passed to the orc
+ * entry points there. Taking the types from the prototypes themselves keeps
+ * these buffers correct under either convention by construction, rather than
+ * naming a type that only happens to match on one of them. New orc test code
+ * should use them too; int64_t/uint64_t buffers compile on Linux but break the
+ * macOS build.
+ */
+typedef decltype (test_orc_elem_type (nns_orc_add_c_s64)) orc_s64_elem;
+typedef decltype (test_orc_elem_type (nns_orc_add_c_u64)) orc_u64_elem;
+
+static_assert (sizeof (orc_s64_elem) == 8 && sizeof (orc_u64_elem) == 8,
+    "orc 64-bit buffers must be 64 bits wide");
+static_assert (std::is_signed<orc_s64_elem>::value && std::is_unsigned<orc_u64_elem>::value,
+    "orc 64-bit buffer signedness must match the s64/u64 prototypes");
 
 /**
  * @brief Test for tensor_transform orc functions (add constant value)
@@ -3903,7 +3935,7 @@ TEST (testTensorTransform, orcAdd)
   }
 
   /* add constant s64 */
-  int64_t data_s64[array_size] = {
+  orc_s64_elem data_s64[array_size] = {
     0,
   };
 
@@ -3928,7 +3960,7 @@ TEST (testTensorTransform, orcAdd)
   }
 
   /* add constant u64 */
-  uint64_t data_u64[array_size] = {
+  orc_u64_elem data_u64[array_size] = {
     0,
   };
 
@@ -4122,7 +4154,7 @@ TEST (testTensorTransform, orcMul)
   }
 
   /* mul constant s64 */
-  int64_t data_s64[array_size] = {
+  orc_s64_elem data_s64[array_size] = {
     0,
   };
 
@@ -4147,7 +4179,7 @@ TEST (testTensorTransform, orcMul)
   }
 
   /* mul constant u64 */
-  uint64_t data_u64[array_size] = {
+  orc_u64_elem data_u64[array_size] = {
     0,
   };
 
@@ -4354,7 +4386,7 @@ TEST (testTensorTransform, orcConvS8)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -4365,7 +4397,7 @@ TEST (testTensorTransform, orcConvS8)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -4481,7 +4513,7 @@ TEST (testTensorTransform, orcConvU8)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -4492,7 +4524,7 @@ TEST (testTensorTransform, orcConvU8)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -4608,7 +4640,7 @@ TEST (testTensorTransform, orcConvS16)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -4619,7 +4651,7 @@ TEST (testTensorTransform, orcConvS16)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -4735,7 +4767,7 @@ TEST (testTensorTransform, orcConvU16)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -4746,7 +4778,7 @@ TEST (testTensorTransform, orcConvU16)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -4862,7 +4894,7 @@ TEST (testTensorTransform, orcConvS32)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -4873,7 +4905,7 @@ TEST (testTensorTransform, orcConvS32)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -4989,7 +5021,7 @@ TEST (testTensorTransform, orcConvU32)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -5000,7 +5032,7 @@ TEST (testTensorTransform, orcConvU32)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -5041,7 +5073,7 @@ TEST (testTensorTransform, orcConvS64)
   const guint array_size = 10;
   guint i;
 
-  int64_t data_s64[array_size] = {
+  orc_s64_elem data_s64[array_size] = {
     0,
   };
 
@@ -5116,7 +5148,7 @@ TEST (testTensorTransform, orcConvS64)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -5127,7 +5159,7 @@ TEST (testTensorTransform, orcConvS64)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -5168,7 +5200,7 @@ TEST (testTensorTransform, orcConvU64)
   const guint array_size = 10;
   guint i;
 
-  uint64_t data_u64[array_size] = {
+  orc_u64_elem data_u64[array_size] = {
     0,
   };
 
@@ -5243,7 +5275,7 @@ TEST (testTensorTransform, orcConvU64)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -5254,7 +5286,7 @@ TEST (testTensorTransform, orcConvU64)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -5373,7 +5405,7 @@ TEST (testTensorTransform, orcConvF32)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -5384,7 +5416,7 @@ TEST (testTensorTransform, orcConvF32)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 
@@ -5504,7 +5536,7 @@ TEST (testTensorTransform, orcConvF64)
   }
 
   /* convert s64 */
-  int64_t res_s64[array_size] = {
+  orc_s64_elem res_s64[array_size] = {
     0,
   };
 
@@ -5515,7 +5547,7 @@ TEST (testTensorTransform, orcConvF64)
   }
 
   /* convert u64 */
-  uint64_t res_u64[array_size] = {
+  orc_u64_elem res_u64[array_size] = {
     0,
   };
 


myungjoo:
Relayed review: the following was produced by a separate reviewing agent with an independent context. I am only relaying its findings.

## Summary

The change does what #4862 asks and stops there. Two files, one of them CI config; the production `gsttensor_transform.c` orc paths are untouched (they hand orc `gpointer`, so they were never affected), and there is no API, ABI, or architecture change, hence nothing to document.

I verified the diagnosis and the fix independently rather than taking the PR description on trust:

- Against a generated `nnstreamer-orc.h` (orc 0.4.32), `orc_int64` really does resolve to `long` in C++ — the `__STDC_VERSION__` branch is skipped and the `INT_MAX == LONG_MAX` fallback is taken. `static_assert (std::is_same<orc_int64, long>::value)` holds, `std::is_same<orc_int64, long long>` does not. So the macOS `long long` mismatch is real.
- The `_orc_elem_type` deduction trick survives `ORC_RESTRICT` expanding to `__restrict__` under both g++ 11 and clang++ 14 (top-level restrict on a parameter is dropped from the function type). Compiles clean.
- Forced a full recompile of `unittest_plugins.cc` in both a GCC and a Clang meson tree with the project defaults (`werror=true`, `warning_level=2`, `cpp_std=c++17`) and `-Dorcc-support=enabled`: no warnings from either.
- Ran the exact CI command locally: 28 tests selected, all pass. The macOS job on this PR is green with the same 28, orc 0.4.42 on arm64, `HAVE_ORC: 1`, `orcc` found — so the fix is verified on the real target, not just by inference.
- `clang-format` clean (v14, repo `.clang-format`).
- Checked `doxygen-tag.sh` will not trip: `ctags -x --c-kinds=f+p` reports exactly one new symbol, `_orc_elem_type` as a prototype, and the `@brief` block directly above it keeps the checker's `brief` flag set. The `static_assert`s are not seen as functions.
- `check_if_rebuild_requires.sh` puts both `.yaml` and `.cc` in the catch-all branch, so `rebuild=1` and the macOS job actually runs on this PR — the fix is not self-exempting.

No blockers and nothing that would regress another module. The points below are all improvements, and the first two are the ones I would like addressed here rather than later.

## Findings

### 1. [minor] The `static_assert`s do not detect the regression they claim to guard

The assertions compare orc's *prototype* element type against orc's *own* typedef:

```cpp
static_assert (std::is_same<decltype (_orc_elem_type (nns_orc_add_c_s64)), orc_int64>::value,
    "orc s64 buffers must be declared as orc_int64");
```

Both sides come from `nnstreamer-orc.h`. The buffer declarations are never referenced, so the message ("orc s64 buffers must be declared as orc_int64") asserts something the code does not check. I confirmed this empirically: with all three assertions in place, reverting a buffer to `int64_t data_s64[10]` and calling `nns_orc_add_c_s64 (buf, 1, 10)` still compiles clean on Linux — the exact regression the PR is fixing slips straight through.

The only thing the assertions can catch is orcc changing its prototype spelling away from `orc_int64`, which would already be a hard pointer-conversion error at the call sites. So their real contribution is a nicer diagnostic, not new detection power. The genuine guard against a re-regression is the macOS build this PR enables — which is good, and worth saying so instead.

Suggestion, which would make the intent real and platform-independent: bind the buffers to the deduced type rather than asserting about it, e.g.

```cpp
typedef decltype (_orc_elem_type (nns_orc_add_c_s64)) orc_s64_buf_t;
typedef decltype (_orc_elem_type (nns_orc_add_c_u64)) orc_u64_buf_t;
```

and declare the buffers with those. Then the type is correct by construction everywhere and the `is_same` assertions become unnecessary. Failing that, please reword the comment and the assertion messages so they do not overstate what is checked.

### 2. [minor] The macOS filter misses the orc paths this PR newly enables in production

The filter is `testTensorTransform.orc*:testTensorTransform.typecast_*accel*`. That covers the orc entry points directly plus the typecast route through `orc_typecast`. It does not cover:

- `arithmetic1Accel` .. `arithmetic5Accel`, which set `acceleration=TRUE` on `mode=arithmetic` and therefore go through `orc_operator` / `orc_func_add|mul|div` in `gsttensor_transform.c`. Enabling orc on macOS turns those production paths on for the first time; nothing in the job exercises them.
- `properties01`, which asserts `DEFAULT_VAL_PROP_ACCELERATION`. This PR flips that default from FALSE to TRUE on macOS. The Notes section says "the newly-enabled tests cover it" — the test that actually pins that default is not in the filter, so that claim does not hold as written.

This is cheap to close: `testTensorTransform.*` runs in 0.07 s locally, and the current 28-test run took 6.4 s on the macOS runner, essentially all process startup. Either widen to `testTensorTransform.*` or append `:testTensorTransform.arithmetic*Accel:testTensorTransform.properties01`. (If widening to the whole suite risks unrelated macOS failures, the targeted append is the safer form.)

### 3. [nit] A `--gtest_filter` that matches nothing keeps the step green

Verified: `meson test -C build unittest_plugins --test-args='--gtest_filter=testTensorTransform.NOSUCHTEST*'` reports `Ok: 1 / Fail: 0` and exits 0. If someone renames the suite or the typecast cases later, the macOS runtime coverage evaporates without any CI signal.

The compile-time protection is unaffected by this (the build step fails outright on a type mismatch, and `-Dorcc-support=enabled` means it cannot silently degrade to a no-orc build — that part of the design is right), so the exposure is limited to the runtime half. Still, a `--gtest_list_tests | grep -c` sanity check, or simply broadening the filter per point 2 so it tracks a whole suite rather than a name pattern, would remove the hole.

### 4. [nit] `#include <type_traits>` is unconditional, its only user is inside `#ifdef HAVE_ORC`

Harmless, and arguably cleaner where it is. Mentioned only for completeness.

### 5. [nit] `_orc_elem_type` is a reserved identifier at global namespace scope

`[lex.name]/3` reserves every leading-underscore identifier in the global namespace to the implementation. The file has precedent (`_harness_wait_for_output_buffer`, `_crop_test_init`, ...), but those have internal linkage; this one is a namespace-scope template. Practically a non-issue with GCC/Clang; flagging it only because the name is new.

## Verdict

Mergeable. The topic is solved, the scope is tight and mechanical, nothing outside the test file and the macOS job is touched, and the change is proven on the real target by the green macOS run. Points 1 and 2 are worth a follow-up commit here — 2 in particular is a one-line filter edit that closes a coverage gap the PR description already claims to have closed — but neither blocks merging.

myungjoo:
Thanks — all five points are addressed in the amended commit. Response to each:

**1 (minor), the assertions did not guard what they claimed — accepted, and you were right.**
They compared orc's prototype element type against orc's own typedef, so the buffer
declarations were never bound to anything. Taken the stronger option you suggested: the
buffer types are now *derived* from the prototypes,

```c
typedef decltype (nns_orc_elem_type (nns_orc_add_c_s64)) orc_s64_buf;
typedef decltype (nns_orc_elem_type (nns_orc_add_c_u64)) orc_u64_buf;
```

and all 26 buffer declarations use them, so drift is impossible by construction rather
than merely asserted. The remaining `static_assert`s only check width and signedness of
the deduced types, which is what they can honestly claim.

Worth stating plainly, because the original wording overstated it: *new* orc test code
written with `int64_t`/`uint64_t` still compiles on Linux and still breaks on macOS. No
construct in the file can catch that on Linux, because on Linux those buffers really are
correct. The macOS build is the guard. The comment and the commit message now say that
instead of implying an every-platform guarantee, and the PR body has a note on it.

**2 (minor), the macOS filter missed orc paths this PR newly enables — accepted.**
The filter now also selects `arithmetic*Accel` and `properties01`:

```
testTensorTransform.orc*:testTensorTransform.typecast_*_accel:testTensorTransform.arithmetic*Accel:testTensorTransform.properties01
```

34 tests instead of 28, verified locally. You were right that the "the newly-enabled tests
cover it" claim in the PR notes did not hold for the acceleration default; it does now.

**3 (nit), a zero-match filter keeps the step green — accepted.**
Confirmed the behaviour locally (`[  PASSED  ] 0 tests.`, exit 0). The step now asserts a
non-zero test count from the meson test log, and I verified the guard fails on a filter
that selects nothing.

**4 (nit), unconditional `#include <type_traits>` — fixed.** Moved inside the `#ifdef
HAVE_ORC` block next to `nnstreamer-orc.h`, which is where its only user lives.

**5 (nit), reserved leading-underscore identifier — fixed.** Renamed `_orc_elem_type` to
`nns_orc_elem_type`, matching the `nns_orc_` prefix the project already uses for these
entry points.

Re-verified after the changes: GCC 11 and Clang 14 builds clean under `werror=true` /
`warning_level=2` with orc enabled, `clang-format` v16 clean, and the exact CI command
selects 34 tests, all passing.

Unrelated, for the record: while re-running the full suite I hit
`tensorFilterCustom.flexibleInvoke_p` failing intermittently (1 pass in 3) under load
average ~32 on my machine. It passes in isolation, the binary predates this PR's edits and
does not link the changed file, so it is a pre-existing load-sensitive flake in
`unittest_filter_custom`, not something this PR introduces. Flagging it rather than
folding it into this PR.

myungjoo:
Relayed review (round 2): the following was produced by a separate reviewing agent with an independent context. I am only relaying its findings.

Re-reviewed at `9078a65`. **Approve.** All five points from the previous round are addressed, and correctly. Everything below was verified directly rather than taken from the PR description.

## Verification of the round-1 fixes

**1. `static_assert` binding (was minor 1) — fixed, and fixed properly.**

The typedefs now *are* the buffer types, so the binding is not a claim, it is the declaration:

```cpp
typedef decltype (nns_orc_elem_type (nns_orc_add_c_s64)) orc_s64_buf;
```

Checked mechanically: 26 of 26 buffer declarations in the orc region use `orc_s64_buf`/`orc_u64_buf`, and no `int64_t`/`uint64_t` array declaration is left anywhere between `#ifdef HAVE_ORC` and the end of the orc tests. The only surviving `int64_t` in that region are the scalar `int64_t val = (int64_t) data_f32[i];` locals used to compute expected values, which are correct as-is and have nothing to do with the orc boundary.

The two remaining assertions are not vacuous. Because the operand type is deduced from the prototype rather than named, they now check properties of *orc's* prototype that could actually change: `sizeof == 8` fires if `nnstreamer-orc.orc` ever narrowed `.dest 8` to `.dest 4` (which would otherwise silently shrink every 64-bit test buffer with no compile error), and `is_signed`/`is_unsigned` fires if orcc emitted something like `orc_union64` instead of an integer type. That is a real, if narrow, guard, and — unlike the previous `is_same` pair — the assertion messages now describe what is actually checked.

**2. Overstated wording — removed.** The header comment, the commit message, and the PR body all now say plainly that `int64_t`/`uint64_t` buffers written in future would still compile on Linux and that the macOS build is the only guard. The Notes section spells out why that means enabling orc on macOS belongs in this PR rather than a follow-up. That is the honest framing.

**3. Widened filter (was minor 2) — fixed.** `testTensorTransform.orc*:testTensorTransform.typecast_*_accel:testTensorTransform.arithmetic*Accel:testTensorTransform.properties01` selects exactly 34 (14 orc + 14 typecast accel + 5 `arithmetic1..5Accel` + `properties01`). Ran it locally: 34 selected, 34 pass. The tightened `typecast_*_accel` still matches all of `typecast_1_accel` .. `typecast_14_accel`. `arithmetic*Accel` does not over-match `arithmeticPerChannel` or the `arithmeticProperties*_n` cases.

More to the point, this PR's own macOS run is green with `[  PASSED  ] 34 tests.` — so `orc_operator`'s add/mul/div paths and the `acceleration`-default flip are now confirmed working on arm64 macOS, not merely assumed. The float expectations in the arithmetic cases are `.5`-based and use `EXPECT_FLOAT_EQ`/`EXPECT_DOUBLE_EQ`, so there is no ULP-drift exposure from adding them.

**4. Zero-match guard (was nit 3) — fixed, and it genuinely works.** Verified both directions locally:

- positive: the real filter produces `[  PASSED  ] 34 tests.` at column 0 of `build/meson-logs/testlog.txt`, and `grep -qE` matches it;
- negative: a filter matching nothing makes gtest write `[  PASSED  ] 0 tests.`, `meson test` still exits 0, and the `[1-9][0-9]*` guard correctly rejects it — under `bash -e` that turns the step red, which is exactly the hole I flagged.

On BSD-grep portability: the pattern is plain POSIX ERE, and I cross-checked it against three independent regex engines (GNU grep, busybox grep, awk's ERE) with identical results, including the singular `1 test.` form that `tests?` covers. It is also settled empirically — the grep is the last command of the step in the macOS job, so the step being green *is* proof that macOS's BSD grep matched.

**5. `#include <type_traits>` (was nit 4) — moved inside `#ifdef HAVE_ORC`.** `std::is_signed`/`is_unsigned` are its only users and both live in that block, so a `-Dorcc-support=disabled` build is back to exactly its pre-PR include set.

**6. `_orc_elem_type` → `nns_orc_elem_type` (was nit 5) — done.** No longer a reserved global-namespace identifier.

## New checks on this revision

- Forced full recompiles of the TU in a GCC 11 and a Clang 14 meson tree at the project defaults (`werror=true`, `warning_level=2`, `c++17`, orc enabled): zero warnings from either. The mid-file `#include <type_traits>` sits in its own include block, so clang-format's `SortIncludes` leaves it alone.
- Full `unittest_plugins` suite (not just the filtered subset) passes on both trees — the rename and the typedef change did not disturb anything else in the file.
- `clang-format` v14 against the repo `.clang-format`: clean.
- `doxygen-tag.sh` in advanced mode: exits 0. `ctags -x --c-kinds=f+p` reports exactly one new symbol, `nns_orc_elem_type` as a prototype, and the `@brief` block directly above it covers it; the typedefs and the `static_assert`s are not seen as functions or structures, so they need no tags. (The `$report_path` and `local:` messages the script emits are pre-existing and unrelated.)

## Findings

Three nits, none blocking, all optional.

### [nit] `orc_s64_buf` / `orc_u64_buf` name the element type but read as the buffer

At the 26 declaration sites this reads as "an array of s64 buf":

```cpp
orc_s64_buf res_s64[array_size];
```

The doc comment gets it right ("Buffer element types for the 64-bit orc test cases"), so intent is clear, but `orc_s64_elem` / `orc_s64_elem_t` would read truer at the point of use. Purely cosmetic.

### [nit] `nns_orc_elem_type` adopts the generated-symbol prefix

`nns_orc_` is the prefix orcc generates for this project's `.orc` entry points (`nns_orc_add_c_s64`, `nns_orc_conv_*`, ...). This helper is hand-written and lives in the test file, so the name invites a reader to go looking for it in `nnstreamer-orc.h`, and a hypothetical future `.function elem_type` in `nnstreamer-orc.orc` would collide with it. Something like `orc_test_elem_type` keeps it visibly outside the generated namespace. Very low risk — raised only because the name was chosen in response to my earlier note, and this trades one naming concern for a smaller one.

### [nit] Stale verification bullet in the PR body

> Confirmed the new assertions are not vacuous: asserting the prototype takes `long long` fails to compile with the expected message.

That describes the previous `std::is_same` assertions, which no longer exist — the current pair checks width and signedness. Body text only, no code impact, but worth correcting so the description matches the diff.

## Two notes, not findings

- The guard asserts "at least one test ran", so a *partial* rename (say the `orc*` cases vanishing while the other 20 still match) would stay green. Pinning the exact count would be stronger but goes stale every time a case is added; I would leave it as it is.
- Both typedefs are deduced from the `add_c_*` prototypes, while the `conv_*_to_s64/u64` buffers rely on those prototypes using the same element type. That holds today, and any divergence is a hard pointer-conversion error rather than a silent problem, so there is nothing to do here.

## Verdict

**Approve.** The topic is solved, the scope is still just the test file plus the macOS job, no production code or public interface is touched, and the change is now proven on the real target: 34/34 on macOS arm64 with orc 0.4.42, including the acceleration-default flip and the orc arithmetic paths that round 1 did not cover. The three nits are cosmetic or description-only and do not need to hold up the merge.

myungjoo:
All three nits are resolved in the amended commit (`c4a0ecd1`), so nothing is left for a follow-up.

- **`orc_s64_buf`/`orc_u64_buf` read as "buffer" at the declaration sites** — renamed to
  `orc_s64_elem`/`orc_u64_elem`. Agreed, `orc_s64_elem res_s64[array_size]` reads truer
  than "array of s64 buf".
- **Helper name sits inside the `nns_orc_` namespace orcc generates** — renamed to
  `test_orc_elem_type`. I went one step further than the suggested `orc_test_elem_type`:
  that still leads with `orc_`, which is the orc library's own namespace, so the same
  objection applies in weaker form. `test_orc_elem_type` is outside both, and a reader
  will not go looking for it in `nnstreamer-orc.h`.
- **Stale PR-body bullet describing the deleted `is_same` assertions** — rewritten to
  describe what is actually verified now: that all 26 declarations use the deduced
  typedefs and no `int64_t`/`uint64_t` array declaration remains in the orc section, with
  the `static_assert`s covering the width/signedness change that deduction cannot catch.
  The `test_orc_elem_type` rename is reflected in the body too.

On your two non-findings, for the record: I agree on both, and left both as they are.
Pinning an exact test count would go stale every time an orc or accel case is added, and
the `conv_*` prototypes cannot silently diverge from `add_c_*` — orcc emits one typedef
set for the whole file, and any divergence is a hard error at the call sites rather than
something the deduction would paper over.

Re-verified after the renames: no occurrences of the old names remain, GCC 11 and Clang 14
build clean at `werror=true`/`warning_level=2` with orc enabled, `clang-format` v16 clean,
the CI filter still selects 34 tests all passing with the zero-match guard satisfied, and
the full `unittest_plugins` suite passes.

myungjoo:
Relayed review (round 3, final): the following was produced by a separate reviewing agent with an independent context. I am only relaying its findings.

Re-reviewed at `c4a0ecd`. **Approve — no findings.** All three nits from round 2 are addressed, the change introduces nothing new, and I have no further comments.

## What changed since `9078a65`

`git diff 9078a657 c4a0ecd1` is 31 insertions / 31 deletions in one file, and every one of them is a rename. `.github/workflows/macos.yaml` is byte-identical to the revision whose macOS run was already green, so all of my round-2 verification of the CI step, the widened filter, and the zero-match guard carries over unchanged rather than needing to be redone.

**1. `orc_s64_buf`/`orc_u64_buf` -> `orc_s64_elem`/`orc_u64_elem`** — complete. 30 lines carry the new names (26 buffer declarations + 2 typedefs + 2 `static_assert`s), and `git grep` over the whole tree at the PR head finds **zero** occurrences of `orc_s64_buf`, `orc_u64_buf`, `nns_orc_elem_type`, or `_orc_elem_type` — nothing left behind in code, comments, the workflow, or the docs. The declarations now read as they should:

```cpp
orc_s64_elem res_s64[array_size];
```

and the block comment above them, "Buffer element types for the 64-bit orc test cases", lines up with the name instead of quietly correcting it. The two `static_assert` messages still speak of "orc 64-bit buffers", which is accurate — a buffer of a 64-bit element type is a 64-bit buffer — so there is no mismatch to fix there.

**2. `nns_orc_elem_type` -> `test_orc_elem_type`** — and yes, the reasoning holds. `orc_*` is orc's own public prefix (`orc_init`, `orc_program_new`, `orc_int64`, `orc_union64`, ...), so `orc_test_elem_type` really would have landed inside it, and my earlier suggestion was the weaker of the options on the table. The three candidates each cost something:

- `orc_test_elem_type` — sits in orc's prefix;
- `_orc_test_elem_type` — matches this file's existing `_aggregator_test_push_buffer` / `_crop_test_init` word order, but reintroduces the reserved-global-identifier point I raised in round 1;
- `test_orc_elem_type` — avoids both, and pays only by inverting that `<subject>_test_<verb>` word order.

Paying in word order to avoid two correctness-adjacent concerns is the right trade. Confirmed no other symbol named `test_orc*` exists anywhere in the repo.

One observation, explicitly **not** a request: by the same argument, `orc_s64_elem` also sits in orc's prefix, so the helper and the typedefs are held to slightly different standards. I would leave that alone. The practical collision risk in both directions is negligible (upstream orc would have to introduce that exact name, and the result would be a compile error rather than a silent bug), and for the typedefs the `orc_` prefix earns its keep — it puts them visually next to `orc_int64` at the point of use, which is exactly what a reader needs there. The helper is different because `nns_orc_` is generated by orcc from a file this project edits, so a collision there was plausible rather than theoretical. Treating the two differently is defensible, not sloppy.

**3. Stale PR-body bullet** — rewritten, and the replacement describes the current design correctly: the deduction is what binds the buffers, and the `static_assert`s "cover what deduction cannot -- a width or signedness change on the orc side -- rather than restating it". That is precisely right. The helper name is updated in the body too, and the commit message refers to "a declared-only helper" without naming it, so nothing is stale there either.

## Re-verified on this revision

Not assumed from the rename being mechanical — re-run end to end:

- GCC 11 and Clang 14 forced recompiles of the TU at the project defaults (`werror=true`, `warning_level=2`, `c++17`, `-Dorcc-support=enabled`): zero warnings from either.
- `clang-format` v14 against the repo `.clang-format`: clean. The CPP-Linter job on this commit is already green, which independently covers the v16 the job pins.
- `doxygen-tag.sh` in advanced mode: exits 0. `ctags -x --c-kinds=f+p` reports exactly one symbol from the new block, `test_orc_elem_type` as a prototype at line 3785, covered by the `@brief` directly above it; the typedefs and `static_assert`s are still not seen as functions or structures. (The `$report_path` / `local:` messages the script prints are pre-existing and file-independent.)
- The macOS filter run locally: 34 selected, 34 pass, and the zero-match guard is satisfied by the resulting `testlog.txt`.
- Full `unittest_plugins` suite on both trees: passes.
- Spell check: `elem` is already used across the repo and is not in the `_typos.toml` word list, so the rename introduces no new risk there.

The macOS job for `c4a0ecd` is still queued as I write this. Since this revision differs from the already-green `9078a65` only by identifier renames, and both compilers plus the full suite pass locally, I do not expect it to behave differently — but it is worth a glance before merging, as always.

## Verdict

**Approve, no findings.** The topic from #4862 is solved; the scope never grew beyond the test file and the macOS job; no production code, public interface, or documented behavior is touched; the buffers are correct by construction under both LP64 conventions rather than by a coincidence that happens to hold on Linux; and the coverage that guards it is real and proven on the target — 34/34 on macOS arm64 with orc 0.4.42, covering the orc entry points, the typecast and arithmetic acceleration paths, and the `acceleration` default flip. Ready to merge once CI finishes.

myungjoo:
Thanks — this is a genuinely useful second pass, and item 2 was a real error on my part.
All four items are addressed in `76bdfbca`.

**1. Why deduction rather than `orc_int64` directly — fair, and now written down.**
You are right that the minimal fix is 26 one-token substitutions and that the description
never said why that was passed over. For the record on how it got here: the first revision
*did* name `orc_int64` directly, with `static_assert`s alongside it. A previous reviewer
pointed out those assertions bound nothing — reverting a buffer to `int64_t` still compiled
clean on Linux with them in place — and suggested deriving the buffer type instead. So the
deduction is the answer to "make the buffers actually tied to the prototypes", not a
first-choice preference.

The commit message now states the trade explicitly: naming `orc_int64` would fix the build
just as well and in fewer lines; deducing keeps the buffers tied to the prototypes rather
than to a second name that has to be kept in sync, so they survive orcc someday emitting
the parameter as the `.orc`-declared type instead of its own typedef.

Your two consequences are both accurate and I am leaving them as they are. A future 64-bit
entry point with a different signature would indeed need a second overload — that is a
one-line addition at the point it is needed, and I would rather not speculate a shape now.
And yes, with deduction in place the `static_assert`s only cover an orc-side width or
signedness change; they are cheap and honest about it.

**2. `properties01` — you are right, my rationale was wrong.**
`DEFAULT_VAL_PROP_ACCELERATION` comes from the test file's own `HAVE_ORC` and
`DEFAULT_ACCELERATION` in `gsttensor_transform.c:113` from that file's, so both sides move
together and the test cannot fail merely because the default is FALSE. I checked this
against both files rather than take it on faith, and your reading is correct.

Reworded to your framing, in the workflow comment, the commit message, and the PR body: it
catches the element and the test being built with orc *disagreeing*, which is a new
possibility precisely because this is the first time both are built with orc on macOS. The
test stays in the filter; only the justification changes.

**3. The guard checks that something ran, not that everything ran — fixed, adopted.**
You are right that this is the same class of silent coverage loss the PR exists to fix, one
level up. A previous reviewer had argued against pinning a count on the grounds it goes
stale as tests are added, but that objection applies to an exact match, not to your floor:
`-ge 34` stays green when an orc case is added and goes red only when the set shrinks,
which is exactly the wanted behaviour. Adopted your snippet:

```sh
ran=$(sed -nE 's/^\[  PASSED  \] ([0-9]+) tests?\.$/\1/p' build/meson-logs/testlog.txt | tail -1)
test "${ran:-0}" -ge 34
```

Verified in all three directions locally: the real run parses 34 and passes; a filter
matching nothing parses 0 and fails; a filter narrowed to `orc*` only parses 14 and fails.
The ERE was also cross-checked against busybox sed, and BSD sed on the runner will confirm
`-E`.

**4. Include placement** — agreed, it is deliberate for the reason you give, and moving it
to the top would need its own `#ifdef`. Left as is.

One note on the description bullet you did not flag but which changed with item 2: the
"asserts that a non-zero number of tests ran" line is now "requires the expected number of
tests to have run", to match the guard.

myungjoo:
Relayed review (round 4, final): the following was produced by a separate reviewing agent with an independent context. I am only relaying its findings.

Re-reviewed at `76bdfbca`, after the second reviewer's pass. **Approve — no findings.** All three actionable items from that review are addressed, and one of them corrects a mistake I made in my own round-2 comment.

## Scope of this revision

`git diff c4a0ecd1 76bdfbca` touches `.github/workflows/macos.yaml` only, 6 insertions / 4 deletions. `tests/nnstreamer_plugins/unittest_plugins.cc` is **byte-identical** to `c4a0ecd1`, so the C++ verification I ran on that head — GCC 11 and Clang 14 at `werror=true`/`warning_level=2` with zero warnings, `clang-format` clean, `doxygen-tag.sh` exit 0 with ctags resolving `test_orc_elem_type` under its `@brief`, full `unittest_plugins` suite green on both trees — carries over unchanged. The Static checks, CPP-Linter and Spell Check jobs are green on this head, which independently covers the same ground.

## The second reviewer's item 2 was right, and I had it wrong

I checked this against the sources rather than taking either side's word for it:

- `tests/nnstreamer_plugins/unittest_plugins.cc:40-44` — `DEFAULT_VAL_PROP_ACCELERATION` is `TRUE`/`FALSE` on the **test TU's** `HAVE_ORC`.
- `gst/nnstreamer/elements/gsttensor_transform.c:112-116` — `DEFAULT_ACCELERATION` is `TRUE`/`FALSE` on the **element TU's** `HAVE_ORC`, consumed at line 253 (`g_param_spec_boolean` default) and line 306 (`filter->acceleration`).

`HAVE_ORC` is a project-wide `project_args` define, so in any normal build both sides move together and `properties01` cannot fail merely because the default is FALSE on macOS. What it actually detects is the two translation units being built with **disagreeing** `HAVE_ORC` — which, as the second reviewer notes, is a genuinely new possibility now that both are built with orc on macOS for the first time. That is the correct reading, and my round-2 comment endorsed the wrong one when I described the macOS run as having "confirmed the acceleration-default flip". It confirmed the two TUs agree, which is a different and narrower claim.

The correction is applied in all three places and each matches the source: the workflow comment ("compares the element's acceleration default against the test's own HAVE_ORC and so catches the two being built with orc disagreeing"), the commit message, and the PR body ("so it does not fail merely because the default is FALSE"). No stale phrasing left anywhere.

## Item 3: the count floor — sound judgment, and it holds up

You asked whether replacing the presence check with a floor conflicts with my round-2 position. It does not, and the distinction is the right one. My objection was to pinning an **exact** count, which goes stale every time a case is added; `-ge` is monotone in the direction tests actually grow, so that objection does not transfer. I verified this rather than assuming it — running the exact step body under `bash -e`, four ways:

| filter | parsed `ran` | step exit |
| --- | --- | --- |
| the real filter | 34 | 0 (pass) |
| matches nothing | 0 | 1 (fail) |
| `orc*` only — partial match | 14 | **1 (fail)** |
| whole `testTensorTransform.*` suite | 92 | 0 (pass) |

Row 3 is the capability the second reviewer asked for and the old `grep -qE` did not have. Row 4 is the answer to the maintenance worry: adding orc cases can only push the number up, so the floor never produces a false failure. It needs touching only when cases are **removed** or renamed out of the filter — which is exactly the event that should force a conscious decision rather than pass silently. The maintenance cost is one-directional and desirable.

One honest caveat, not a request: once cases are added, `-ge 34` stops meaning "all of them ran" and means "at least the original set's worth ran". That is strictly weaker than "all" but strictly stronger than the old "at least one", and it is the right place on that trade-off curve. Leave it.

**Portability and shell mechanics**, all checked:

- The `sed -nE 's/^\[  PASSED  \] ([0-9]+) tests?\.$/\1/p'` expression produces identical output under GNU sed and busybox sed — two independent implementations — including the singular `1 test.` form and the `0 tests.` case it must extract rather than skip.
- More decisive: the macOS job on this head is green, and `test "${ran:-0}" -ge 34` is the **last** command of the step under `bash -e`. Had BSD sed failed to parse the line, `ran` would be empty, `${ran:-0}` would be `0`, and the step would be red. It is green with `[  PASSED  ] 34 tests.` in the log, so BSD sed compatibility is settled empirically, not argued.
- `ran=$(… | tail -1)` under `set -e`: the assignment takes the pipeline's status, which is `tail`'s 0, so a non-matching log does not abort early — it falls through to `test`, which is where you want the failure to surface. `${ran:-0}` handles the empty case. And `${ran:-0}` is `${…}`, not `${{…}}`, so GitHub's expression syntax does not touch it inside the `run: |` block.
- `tail -1` is defensive rather than necessary — a testlog from one `meson test` run contains exactly one `PASSED` line — but it costs nothing and protects against that assumption changing.

## Item 1: rationale for deduction over `orc_int64`

Now recorded in both the commit message and the PR body, and the trade-off is stated honestly in both directions — that naming `orc_int64` "would fix the build just as well and in fewer lines", and that deduction buys keeping the buffers tied to the prototypes rather than to a second name that must be kept in sync. That is the right framing; a future reader will not have to guess why the shorter spelling was passed over. (For the record, the deduction approach came out of my round-1 comment, so the second reviewer's question was a fair one to put to the PR rather than a disagreement with it.)

## Item 4: include placement

Agreed with the second reviewer — `#include <type_traits>` sits inside the same `#ifdef HAVE_ORC` as the pre-existing `#include "nnstreamer-orc.h"`, hoisting it would need its own `#ifdef` to avoid pulling it in unconditionally, and it is the only consumer of `std::is_signed`/`is_unsigned`. Reads as deliberate. No action.

## Verdict

**Approve, no findings.** Nothing in this revision introduces a new problem; the only code change is the workflow guard, and it is stronger than what it replaces in exactly the way that was asked for. The PR now describes what it does accurately — including the one thing the description previously overstated, which two reviewers had let through. Ready to merge once the remaining jobs finish; macOS, Static checks, CPP-Linter, Spell Check, LLVM/Clang and the arm Valgrind run are already green at this head.

Issue #4848, opened 2026-03-24
Missing release for Ubuntu 25.10 / Questing Quokka


manmountain:
# Issue Description
No release available on the Ubuntu 25.10 PPA for installation using apt

Expected Result
============
I expected to be able to add the nnstreamer ppa repoository using:

`sudo apt-add-repository ppa:nnstreamer`

succesfully when following the [guide](https://github.com/nnstreamer/nnstreamer/blob/main/Documentation/getting-started-ubuntu-ppa.md). But instead I get error message

```
E: The repository 'https://ppa.launchpadcontent.net/nnstreamer/ppa/ubuntu questing Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

How to Reproduce
===============
1. Install Ubuntu 25.10 / Questing Quokka
2. Run `sudo apt-add-repository ppa:nnstreamer` in the terminal
3. See error
```
E: The repository 'https://ppa.launchpadcontent.net/nnstreamer/ppa/ubuntu questing Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
```

myungjoo:
Directly working on 26.04. 25.10 will be skipped for 26.04 support.

myungjoo:
The build-dependency defect behind this report was fixed in #4895 (merged 2026-08-28). Ubuntu 25.10 will not be published; 26.04 (resolute) is being added instead, tracked in #4902 with the packaging change in #4916. Closing this one in favour of #4902.

PR #4895, opened 2026-08-28
[Packaging] Add an unversioned gcc fallback to Build-Depends

myungjoo:
Addresses the repository side of #4848.

## Problem

`ppa:nnstreamer` publishes nothing newer than noble (24.04), so `apt update` on
a newer Ubuntu fails with `does not have a Release file`. The reason is in
`debian/control`:

```
Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4),
```

Ubuntu 25.04 and later carry **none** of `gcc-5`..`gcc-9`, and neither does any
current Debian release, so the clause is unsatisfiable there:

| series | gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 | gcc |
|---|---|---|---|---|---|---|
| xenial 16.04 | – | – | – | – | 5.4.0 | 4:5.3.1 |
| bionic 18.04 | – | 8.4.0 | 7.5.0 | 6.5.0 | 5.5.0 | 4:7.4.0 |
| focal 20.04 | 9.4.0 | 8.4.0 | 7.5.0 | – | – | 4:9.3.0 |
| jammy 22.04 | 9.5.0 | – | – | – | – | 4:11.2.0 |
| noble 24.04 | 9.5.0 | – | – | – | – | 4:13.2.0 |
| **plucky 25.04** | – | – | – | – | – | 4:14.2.0 |
| **resolute 26.04** | – | – | – | – | – | 4:15.2.0 |
| bookworm / trixie / sid | – | – | – | – | – | 12.2 / 14.2 / 16.1 |

The `nnstreamer-daily` recipe already targets Plucky, and its build parks in
`Dependency wait — Missing build dependencies: gcc-9` instead of failing
loudly, so no package is ever published for the series.

## Change

Append an unversioned `gcc` alternative to all three control files
(`debian/rules` swaps them in per vendor, so leaving one behind would let the
source build and the binary build disagree).

## Why this is safe on existing releases

- On xenial through noble an earlier alternative is still installable, so
  resolution is unchanged there.
- Where the fallback does take effect it changes nothing about what is built.
  `gcc-9` was never the compiler: the PPA build logs show meson selecting `cc`,
  the distro default gcc.

  ```
  [noble]  Setting up gcc-9 (9.5.0-6ubuntu2) ...
  [noble]  C compiler for the host machine: cc (gcc 13.3.0 ...)
  [jammy]  Setting up gcc-9 (9.5.0-1ubuntu1~22.04) ...
  [jammy]  C compiler for the host machine: cc (gcc 11.4.0 ...)
  ```

  The only practical effect is that `gcc-9` and its three dependencies stop
  being installed into the build chroot, because `build-essential` already
  provides `gcc`.
- A core build on Ubuntu 25.04 with gcc 14.2 and `werror=true` was run to
  confirm the newer compiler is viable: 168/168 targets, no new warnings.
  Subplugins and unit tests were out of that run, since their build
  dependencies are not published for 25.04 yet.
- `packaging/nnstreamer.spec` has no versioned gcc `BuildRequires`, so Tizen is
  untouched. meson does not read `debian/control`.
- `debian/control` is part of the pbuilder cache key in `pdebuild.yml` and
  `update_pbuilder_cache.yml`; the key changes once and `restore-keys` covers
  the fallback.

## Regression test

`.github/workflows/static.check.scripts/debian-control.sh` rejects any
build-dependency alternative group made only of version-suffixed toolchain
packages. It fails on the tree as it stood before this PR:

```
::error::debian/control: 'gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5' lists only
version-suffixed toolchain packages, which distros eventually drop. Add an
unversioned alternative such as '| gcc'.
```

and passes after it. `test_debian_control.sh` pins both directions plus the
parser's handling of arch/profile qualifiers, in-field comments, and later
fields, and finally runs the checker against the shipped control files. Both
are wired into `static.check.yml` behind the same merge-ref guard the existing
self-test steps use.

The rule is not written around gcc-9 specifically: a future `gcc-13 | gcc-12`
is rejected the same way.

## Documentation

`Documentation/getting-started-ubuntu-ppa.md` now explains what the
`does not have a Release file` error means and where to check which series are
published, which is the message the reporter of #4848 hit.

## Not covered here

This is the repository half only. Publishing for a new series additionally
requires, on Launchpad:

1. the recipes to target the series (`nnstreamer-daily` currently lists
   Focal/Jammy/Noble/Plucky; no series newer than Plucky), and
2. the PPA dependency stack rebuilt for it — `ssat`, `nnstreamer-edge`,
   `tensorflow2-lite`, `edgetpu`, `openvino`, `onnxruntime`, `tvm`, `onert`.
   Several of these already list Plucky as a target but have never had a build
   requested; `onnxruntime-daily` has been failing on every series since
   2025-01-24.

Two related findings that are **not** addressed here, since their cause is
different:

- Jammy PPA builds have been in `Dependency wait` on `meson (>= 0.62.0)` since
  2025-12-19 (jammy ships 0.61.2); the last jammy publish is 2025-12-04.
- `debian/control.ubuntu.ppa` and `debian/control.debian` still declare
  `meson (>=0.49)` while `meson.build` requires `>=0.62.0`.

Note that Ubuntu 25.10 (questing), the release named in #4848, has since gone
end-of-life, so the practical target is now 26.04 LTS. The issue should stay
open until packages actually appear for a current series.

### DIFF
--- a/.github/workflows/static.check.scripts/debian-control.sh
+++ b/.github/workflows/static.check.scripts/debian-control.sh
@@ -50,7 +50,7 @@ extract_clauses() {
     /^[[:space:]]*#/ { next }
     /^[[:space:]]*$/ { collecting = 0; next }
     /^[A-Za-z0-9][A-Za-z0-9-]*:/ {
-      collecting = ($0 ~ /^Build-Depends(-Arch|-Indep)?:/)
+      collecting = (tolower($0) ~ /^build-depends(-arch|-indep)?:/)
       if (!collecting) { next }
       if (started) { printf "," }
       started = 1
--- a/.github/workflows/static.check.scripts/debian-control.sh
+++ b/.github/workflows/static.check.scripts/debian-control.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file     debian-control.sh
+# @brief    Verify that debian/control* build-dependencies stay satisfiable.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Ubuntu 25.04 and later, and every current Debian release, dropped gcc-5
+# through gcc-9 from their archives. A Build-Depends alternative group that
+# names only version-suffixed toolchain packages therefore becomes
+# unsatisfiable on a new distro, and the PPA source build parks in
+# "Dependency wait" instead of failing loudly, so no package is ever
+# published for that series. Require an unversioned alternative, which the
+# distro default compiler always provides.
+#
+# Argument ($@): control files to check. Defaults to the repository's
+#                debian/control* when no argument is given.
+#
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+REPO_ROOT=$(cd "${SCRIPT_DIR}/../../.." && pwd)
+failed=0
+
+# Toolchain packages whose version-suffixed form leaves distro archives as
+# the compiler ages out.
+TOOLCHAIN_RE='^(gcc|g\+\+|cpp|clang|clang\+\+|llvm)-[0-9]+(\.[0-9]+)*$'
+PKGNAME_RE='^[a-z0-9][a-z0-9+.-]*$'
+
+##
+# @brief Print a string with its runs of whitespace collapsed into single spaces.
+# @param $1 string to normalize
+squeeze() {
+  echo "$1" | awk '{ $1 = $1; print }'
+}
+
+##
+# @brief Print the Build-Depends* fields of a control file, one clause per line.
+# @param $1 path of the control file
+extract_clauses() {
+  awk '
+    /^[[:space:]]*#/ { next }
+    /^[[:space:]]*$/ { collecting = 0; next }
+    /^[A-Za-z0-9][A-Za-z0-9-]*:/ {
+      collecting = ($0 ~ /^Build-Depends(-Arch|-Indep)?:/)
+      if (!collecting) { next }
+      sub(/^[^:]*:/, "")
+    }
+    collecting { printf "%s ", $0 }
+    END { printf "\n" }
+  ' "$1" | tr ',' '\n'
+}
+
+##
+# @brief Report the build dependencies of one control file that cannot hold.
+# @param $1 path of the control file
+check_file() {
+  local file=$1
+  local clause alt name toolchain_only alt_count
+
+  if [ ! -f "$file" ]; then
+    echo "::error::$file does not exist."
+    failed=1
+    return
+  fi
+
+  while IFS= read -r clause; do
+    clause=$(echo "$clause" | tr -d '\r' | sed -e 's/([^)]*)//g' -e 's/\[[^]]*\]//g' -e 's/<[^>]*>//g')
+    if [ -z "$(echo "$clause" | tr -d '[:space:]')" ]; then
+      continue
+    fi
+
+    toolchain_only=1
+    alt_count=0
+    while IFS= read -r alt; do
+      name=$(echo "$alt" | awk '{print $1}')
+      if [ -z "$name" ]; then
+        continue
+      fi
+      alt_count=$((alt_count + 1))
+      if ! echo "$name" | grep -Eq "$PKGNAME_RE"; then
+        echo "::error::$file: invalid package name '$name' in build dependency '$(squeeze "$clause")'."
+        failed=1
+      fi
+      if ! echo "$name" | grep -Eq "$TOOLCHAIN_RE"; then
+        toolchain_only=0
+      fi
+    done < <(echo "$clause" | tr '|' '\n')
+
+    if [ "$alt_count" -gt 0 ] && [ "$toolchain_only" -eq 1 ]; then
+      echo "::error::$file: '$(squeeze "$clause")' lists only version-suffixed toolchain packages, which distros eventually drop. Add an unversioned alternative such as '| gcc'."
+      failed=1
+    fi
+  done < <(extract_clauses "$file")
+}
+
+files=("$@")
+if [ "${#files[@]}" -eq 0 ]; then
+  mapfile -t files < <(find "${REPO_ROOT}/debian" -maxdepth 1 -type f -name 'control*' 2>/dev/null | sort)
+fi
+
+if [ "${#files[@]}" -eq 0 ]; then
+  echo "::error::No debian control file to check."
+  exit 1
+fi
+
+for file in "${files[@]}"; do
+  echo "Checking $file"
+  check_file "$file"
+done
+
+if [ "$failed" -ne 0 ]; then
+  echo "::error::The debian control build-dependency check has failed."
+  exit 1
+fi
+
+echo "The debian control build-dependency check has passed."
--- a/.github/workflows/static.check.scripts/debian-control.sh
+++ b/.github/workflows/static.check.scripts/debian-control.sh
@@ -27,7 +27,8 @@ REPO_ROOT=$(cd "${SCRIPT_DIR}/../../.." && pwd)
 failed=0
 
 # Toolchain packages whose version-suffixed form leaves distro archives as
-# the compiler ages out.
+# the compiler ages out. Matched against the bare name, with any multiarch
+# qualifier such as :native already stripped.
 TOOLCHAIN_RE='^(gcc|g\+\+|cpp|clang|clang\+\+|llvm)-[0-9]+(\.[0-9]+)*$'
 PKGNAME_RE='^[a-z0-9][a-z0-9+.-]*$'
 
@@ -78,6 +79,7 @@ check_file() {
     alt_count=0
     while IFS= read -r alt; do
       name=$(echo "$alt" | awk '{print $1}')
+      name=${name%%:*}
       if [ -z "$name" ]; then
         continue
       fi
--- a/.github/workflows/static.check.scripts/debian-control.sh
+++ b/.github/workflows/static.check.scripts/debian-control.sh
@@ -41,6 +41,9 @@ squeeze() {
 
 ##
 # @brief Print the Build-Depends* fields of a control file, one clause per line.
+#        A separator is emitted between fields because the last dependency of
+#        a field needs no trailing comma, and without it the field would run
+#        into the next one and hide a group inside the merged clause.
 # @param $1 path of the control file
 extract_clauses() {
   awk '
@@ -49,6 +52,8 @@ extract_clauses() {
     /^[A-Za-z0-9][A-Za-z0-9-]*:/ {
       collecting = ($0 ~ /^Build-Depends(-Arch|-Indep)?:/)
       if (!collecting) { next }
+      if (started) { printf "," }
+      started = 1
       sub(/^[^:]*:/, "")
     }
     collecting { printf "%s ", $0 }
--- a/.github/workflows/static.check.scripts/test_debian_control.sh
+++ b/.github/workflows/static.check.scripts/test_debian_control.sh
@@ -107,6 +107,25 @@ Build-Depends: GCC-9 | GCC,
  ninja-build
 Standards-Version: 3.9.6'
 
+expect_checker 1 "a group in Build-Depends-Indep is not hidden by the field before it" \
+'Source: nnstreamer
+Build-Depends: ninja-build
+Build-Depends-Indep: gcc-9 | gcc-8
+Standards-Version: 3.9.6'
+
+expect_checker 1 "a group in Build-Depends is not hidden by the field after it" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc-8
+Build-Depends-Arch: ninja-build
+Standards-Version: 3.9.6'
+
+expect_checker 0 "several Build-Depends fields without a versioned-only group pass" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc
+Build-Depends-Arch: ninja-build
+Build-Depends-Indep: doxygen
+Standards-Version: 3.9.6'
+
 expect_checker 0 "fields after Build-Depends are not scanned" \
 'Source: nnstreamer
 Build-Depends: gcc-9 | gcc
--- a/.github/workflows/static.check.scripts/test_debian_control.sh
+++ b/.github/workflows/static.check.scripts/test_debian_control.sh
@@ -126,6 +126,12 @@ Build-Depends-Arch: ninja-build
 Build-Depends-Indep: doxygen
 Standards-Version: 3.9.6'
 
+expect_checker 1 "a field name is matched regardless of case" \
+'Source: nnstreamer
+build-depends: gcc-9 | gcc-8,
+ ninja-build
+Standards-Version: 3.9.6'
+
 expect_checker 0 "fields after Build-Depends are not scanned" \
 'Source: nnstreamer
 Build-Depends: gcc-9 | gcc
--- a/.github/workflows/static.check.scripts/test_debian_control.sh
+++ b/.github/workflows/static.check.scripts/test_debian_control.sh
@@ -82,6 +82,18 @@ Build-Depends: gcc-13 | gcc,
  pytorch <!nocheck> | libtorch-dev | gcc
 Standards-Version: 3.9.6'
 
+expect_checker 1 "a multiarch qualifier does not hide a versioned-only group" \
+'Source: nnstreamer
+Build-Depends: gcc-13:native | gcc-12:native,
+ ninja-build
+Standards-Version: 3.9.6'
+
+expect_checker 0 "a multiarch qualifier on the fallback is accepted" \
+'Source: nnstreamer
+Build-Depends: gcc-13:native | gcc:native,
+ ninja-build
+Standards-Version: 3.9.6'
+
 expect_checker 0 "a comment inside the field does not terminate it" \
 'Source: nnstreamer
 Build-Depends: gcc-9 | gcc,
--- a/.github/workflows/static.check.scripts/test_debian_control.sh
+++ b/.github/workflows/static.check.scripts/test_debian_control.sh
@@ -1,0 +1,120 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file     test_debian_control.sh
+# @brief    Self-test for debian-control.sh.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Pins the properties that make the checker worth running. The compiler
+# clause as it stood before #4848, naming only version-suffixed gcc
+# packages, must fail; the same clause carrying an unversioned fallback must
+# pass. Without the first the checker could accept everything unnoticed;
+# without the second it could reject every control file. A gcc-13/gcc-12
+# pair is checked too, so the rule is not read as a list of the specific
+# compilers that happened to age out. The repository's own control files run
+# last, which keeps the shipped tree covered even when no fixture changes.
+#
+# The checker is invoked with stdin closed, matching how GitHub runs it, so
+# that a future reader-of-stdin bug fails here instead of hanging.
+#
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+REPO_ROOT=$(cd "${SCRIPT_DIR}/../../.." && pwd)
+CHECKER="${SCRIPT_DIR}/debian-control.sh"
+workdir=$(mktemp -d)
+failed=0
+
+trap 'rm -rf "$workdir"' EXIT
+
+##
+# @brief Run the checker on a generated control file and compare the exit status.
+# @param $1 expected exit status, 0 or 1
+# @param $2 description used for the fixture name and the report line
+# @param $3 body of the control file
+expect_checker() {
+  local expected=$1 desc=$2 body=$3
+  local target actual
+
+  target="${workdir}/$(echo "$desc" | tr -c 'a-zA-Z0-9' '_')"
+  printf '%s\n' "$body" > "$target"
+
+  bash "$CHECKER" "$target" > /dev/null 2>&1 < /dev/null
+  actual=$?
+  if [ $actual -ne 0 ]; then
+    actual=1
+  fi
+
+  if [ "$actual" = "$expected" ]; then
+    echo "PASS: ${desc} (exit ${actual})"
+  else
+    echo "FAIL: ${desc} expected exit ${expected}, got ${actual}"
+    failed=1
+  fi
+}
+
+expect_checker 1 "versioned-only gcc clause is rejected" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4),
+ ninja-build, meson (>=0.62.0)
+Standards-Version: 3.9.6'
+
+expect_checker 0 "unversioned gcc fallback is accepted" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc-8 | gcc-7 | gcc-6 | gcc-5 (>=5.4) | gcc,
+ ninja-build, meson (>=0.62.0)
+Standards-Version: 3.9.6'
+
+expect_checker 1 "a newer versioned-only pair is rejected as well" \
+'Source: nnstreamer
+Build-Depends: gcc-13 | gcc-12,
+ ninja-build
+Standards-Version: 3.9.6'
+
+expect_checker 0 "arch and profile qualifiers do not confuse the parser" \
+'Source: nnstreamer
+Build-Depends: gcc-13 | gcc,
+ libprotobuf-dev [amd64 arm64 armhf],
+ pytorch <!nocheck> | libtorch-dev | gcc
+Standards-Version: 3.9.6'
+
+expect_checker 0 "a comment inside the field does not terminate it" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc,
+# Please list more neural network frameworks as Debian includes them.
+ ninja-build
+Standards-Version: 3.9.6'
+
+expect_checker 1 "an invalid package name is rejected" \
+'Source: nnstreamer
+Build-Depends: GCC-9 | GCC,
+ ninja-build
+Standards-Version: 3.9.6'
+
+expect_checker 0 "fields after Build-Depends are not scanned" \
+'Source: nnstreamer
+Build-Depends: gcc-9 | gcc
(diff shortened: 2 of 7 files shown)


myungjoo:

Reviewed the diff (7 files, +285/-3), `debian/rules`, `packaging/nnstreamer.spec`, `.github/workflows/static.check.yml`, and cross-checked the PR's factual claims against the Launchpad REST API. Summary: **no blocker or major issues found; approve.**

## Verification performed

- **gcc availability table**: queried `getPublishedBinaries` for `gcc-5..gcc-9` and unversioned `gcc` across xenial/bionic/focal/jammy/noble/plucky/resolute. Results match the PR's table exactly (e.g. plucky/resolute have none of gcc-5..gcc-9 but do have `gcc`; jammy/noble have only `gcc-9`; focal has gcc-9/8/7; bionic has gcc-8/7/6/5; xenial has gcc-5). The core motivating claim is correct.
- **`debian/rules` vendor swap**: confirmed `override_dh_clean` copies `control.ubuntu.ppa` or `control.debian` into `debian/control` depending on `dpkg-vendor --derives-from Ubuntu`, so patching all three files (including the checked-in `debian/control`) is required and correct — leaving one behind would have made source/binary builds disagree, as the PR states.
- **Tizen isolation**: `packaging/nnstreamer.spec` has no versioned gcc `BuildRequires` (only a comment mentioning gcc-9/lcov compatibility), confirming this change doesn't touch the Tizen/GBS path.
- **New checker correctness**: ran `test_debian_control.sh` locally — all 7 fixture cases plus the "scan the shipped control files" case pass. Also checked out `debian/control` from `origin/main` (pre-fix) and confirmed `debian-control.sh` fails on it with the exact error text quoted in the PR body, and passes on the post-fix file. Behaves as documented.
- **Shellcheck**: ran `shellcheck` (koalaman/shellcheck:stable) against both new scripts — clean, no warnings.
- **CI wiring**: the two new steps in `static.check.yml` reuse the exact same file-existence/PR-deletion guard pattern as the pre-existing `test_check_if_rebuild_requires.sh` / `test_prohibited_words.sh` steps, so they degrade gracefully instead of hard-failing on older base commits. Consistent with repo convention.
- **Doc links**: both linked docs (`getting-started-ubuntu-debuild.md`, `getting-started-meson-build.md`) and the referenced PPA page exist/resolve.

## Findings

- **nit**: `TOOLCHAIN_RE` in `debian-control.sh` matches plain `gcc-N`/`g++-N`/etc. but wouldn't recognize a multiarch-qualified toolchain package name (e.g. `gcc-9:native`) as toolchain-only. This is only a false-negative risk (a future regression of that specific shape would slip past the checker) and no such form appears anywhere in the current control files, so it's not worth blocking on.
- **informational, not a gap**: this PR is explicitly scoped to the repository side of #4848 only. Getting packages published for a new series still needs Launchpad-side follow-up (recipe targets + rebuilding the PPA dependency stack for ssat/nnstreamer-edge/tensorflow2-lite/edgetpu/openvino/onnxruntime/tvm/onert), which the PR body already discloses under "Not covered here." Recommend keeping #4848 open until a package actually publishes for a current series, as the author also suggests.

## Assessment against the review checklist

- **Solves the stated problem**: yes, for the repository-controllable part — the unsatisfiable `Build-Depends` clause that was parking the Plucky recipe build in "Dependency wait" is fixed, verified against real archive data.
- **Regression risk to other modules**: none found. Confined to `debian/control*`, a new standalone CI checker, workflow wiring, and one doc file. No C/C++/build-system code touched. On every currently-supported series an earlier alternative still resolves, so apt dependency resolution is unchanged there.
- **Size vs. scope**: the 1-line fix itself is trivial; the bulk of the diff is a permanent regression-test (checker + self-test) wired into CI, which is exactly what's warranted by the "prevent recurrence" and "detect future breakage" criteria here — not scope creep.
- **Test coverage**: strong. `test_debian_control.sh` is a genuine regression test (fails on pre-fix tree, passes on post-fix tree, per CLAUDE.md's bug-fix test policy) and doubles as a standing guard against any future PR reintroducing a version-only toolchain alternative (tests a `gcc-13 | gcc-12` case explicitly so the rule isn't read as gcc-9-specific).
- **CI enforcement**: yes — both the checker and its self-test are wired into `static.check.yml` as blocking steps, so a future regression fails CI before merge.
- **Architecture/API/doc**: no architecture or API change. The one behavior-facing change (what the `Release file` error means) is documented in the same PR.

**Merge recommendation**: no blocking issues; safe to merge as-is.

myungjoo:

Thanks — the multiarch nit is addressed in 23fb5599 rather than deferred.

It turned out to cut both ways. Matching the qualified string meant a
versioned-only group written as `gcc-13:native | gcc-12:native` was not
recognised as toolchain-only, and, worse, a legitimate `gcc:native` fallback
was rejected as an invalid package name, so the checker would have blocked a
valid control file. Verified against the pre-fix script:

```
=== BEFORE ===
  versioned-only + :native  -> exit 1 (want 1)   [right answer, wrong reason]
  fallback      + :native  -> exit 1 (want 0)   invalid package name 'gcc:native'
=== AFTER ===
  versioned-only + :native  -> exit 1 (want 1)
  fallback      + :native  -> exit 0 (want 0)
```

The checker now compares the bare name, and `test_debian_control.sh` covers
both directions. The suite is at 9 fixtures plus the shipped control files;
shellcheck is clean.

On the informational point: agreed, and #4848 is intentionally referenced with
`Related to` rather than a closing keyword, so merging this will not close it.
The Launchpad recipe and dependency-stack work is tracked in the "Not covered
here" section of the description.

For the record, the first CI run confirmed both new steps execute rather than
taking the merge-ref skip path:

```
/Checker/ debian control build dependencies
  Checking .../debian/control
  Checking .../debian/control.debian
  Checking .../debian/control.ubuntu.ppa
  The debian control build-dependency check has passed.
/Checker/ debian-control self-test
```

myungjoo:

Re-reviewed the full diff independently (not just the delta since the last review), read the prior review round and the author's reply, and re-verified commit 23fb5599 against fresh, self-authored adversarial fixtures (not the ones already in the test suite).

## Previous nit: resolved correctly

The multiarch-qualifier nit is fixed as claimed. `name=${name%%:*}` strips the qualifier before both the `PKGNAME_RE` and `TOOLCHAIN_RE` checks, so:

- `gcc-13:native | gcc-12:native` (versioned-only, qualified) → still rejected.
- `gcc-13:native | gcc:native` (qualified fallback) → now accepted, whereas before the fix it would have rejected a *valid* control file (the "worse" false-positive direction the author called out).

I re-ran `test_debian_control.sh` locally (git-bash on Windows and WSL) — all 9 fixtures plus the shipped `debian/control*` pass — and reran `shellcheck` (koalaman/shellcheck:stable, via Docker) on both scripts: clean, no warnings. I also hand-tested several qualifier shapes not in the fixture set (`:any`, `:amd64`, a version-constraint *and* qualifier combined, e.g. `gcc-9:native (>= 4:9)`, mixed families like `gcc-9:native | clang-9:native`) — all behave correctly. No regression from the fix itself.

## New finding from the independent re-review

**Severity: major (not a blocker for this PR as shipped, but a real gap in the checker's stated scope).**

`extract_clauses()` in `debian-control.sh` concatenates the content of *every* matched header line (`Build-Depends`, `Build-Depends-Arch`, `Build-Depends-Indep`) into one accumulated string with `printf "%s ", $0`, and only appends a trailing `\n` once, at `END`. It never inserts a separator between two different header fields — it relies entirely on a comma already present in the source text. Debian control syntax does not require a trailing comma on a field's last (or only) item, so when two of these three fields both appear (which is the normal way a package splits arch-dependent from arch-independent build-deps, e.g. adding a `-doc` binary with `Build-Depends-Indep:`), and the earlier field's last line has no trailing comma, the two fields' contents run together into a single "clause," corrupting the alternative grouping.

Reproduction (checker as of 23fb5599, `bash .github/workflows/static.check.scripts/debian-control.sh <file>`):

```
Source: nnstreamer
Build-Depends: libfoo-dev
Build-Depends-Indep: gcc-9 | gcc-8
Standards-Version: 3.9.6
```
→ exits 0 ("has passed"), i.e. the versioned-only `gcc-9 | gcc-8` group in `Build-Depends-Indep` is **silently missed** — a false negative on exactly the class of bug this PR exists to catch.

The outcome is order-dependent (a code smell pointing at the same root cause), not just a single missed case:
```
Build-Depends-Arch: gcc-9 | gcc-8
Build-Depends: libfoo-dev
```
→ this ordering happens to still catch it, by accident (the merged clause becomes `gcc-9 | gcc-8 libfoo-dev`, and `awk '{print $1}'` on each `|`-alternative happens to pick the toolchain name first in both alternatives). A blank line between the two fields does not help either — it resets the `collecting` flag correctly but still doesn't insert a separator into the accumulated output, so the same merge/false-negative happens.

This does not affect any file currently shipped in the repo — `debian/control`, `debian/control.debian`, and `debian/control.ubuntu.ppa` each use a single `Build-Depends:` field only, confirmed with `grep -n '^Build-Depends' debian/control*`. So this PR's actual fix and CI wiring are unaffected, and I'm not asking to block on it. But:

- The checker's own `TOOLCHAIN_RE`/comment and its header regex (`^Build-Depends(-Arch|-Indep)?:`) explicitly advertise support for all three fields, so this is a real gap against the tool's declared scope, not an out-of-scope ask.
- `test_debian_control.sh` has no fixture with two `Build-Depends*` fields, so this exact scenario — the tool's own extended scope — is untested, and per the repo's test-case policy (CLAUDE.md: guard against future regressions) a future PR that splits `Build-Depends-Indep` out with a bad toolchain-only group would merge without CI catching it, defeating the purpose of this whole checker for that case.

Suggested fix (not required for this PR, but worth a fast-follow or a tracking note): have the header-match rule in `extract_clauses` emit a newline before starting a new field's accumulation (e.g. `if (started) printf "\n"; started = 1` ahead of the `sub(...)`), so each field's content lands on its own line before the `tr ','` split; then add a two-field fixture (e.g. `Build-Depends: libfoo-dev` + `Build-Depends-Indep: gcc-9 | gcc-8` with no separating comma) to `test_debian_control.sh` to pin the fix and prevent recurrence.

## Rest of the diff

No other issues found. Scope and size remain appropriate (the follow-up commit is a 15-line, tightly-targeted fix + test), no unrelated modules touched, commit message/sign-off/Doxygen conventions in the new commit are followed, and there's no architecture/API change requiring doc updates beyond what's already in the PR.

## Assessment

- Previous nit: correctly and completely resolved, with test coverage for both directions, verified independently.
- New finding: major-severity gap in the checker's handling of `Build-Depends-Arch`/`Build-Depends-Indep` when they co-occur with another `Build-Depends*` field without an explicit trailing comma — a false negative, order-dependent, currently latent (no shipped file triggers it) but directly undermines the "catch future regressions" purpose of this new tool for the two extra fields it claims to support.
- **Merge recommendation**: no blocker. Safe to merge as-is for the stated problem (#4848's repository-side fix); the new finding is worth a fast-follow fix + fixture rather than holding up this PR, since it doesn't touch anything currently shipped or already covered by the self-test.

myungjoo:

Good catch, and thanks for building the reproducer — the field-boundary bug is
real. It is fixed in 1b50d0bc rather than deferred to a fast-follow, since it
is a defect in code this PR introduces and the fix is three lines.

Reproduced first against the parser as it stood:

```
--- extract_clauses output ---
    [ ninja-build  gcc-9 | gcc-8 ]
--- checker verdict ---
The debian control build-dependency check has passed.
exit=0   (want 1)
```

`ninja-build` and the versioned-only group had merged into one clause, and
`ninja-build` alone was enough to clear the toolchain rule. The parser now
emits a separator when a new `Build-Depends*` field starts, so the same input
is rejected:

```
::error::ctl: 'gcc-9 | gcc-8' lists only version-suffixed toolchain packages,
which distros eventually drop. Add an unversioned alternative such as '| gcc'.
exit=1
```

Three fixtures were added for the scenario, covering the order dependence you
noted — the hidden group ahead of the neighbouring field, behind it, and a
three-field control file that legitimately passes. The suite is at 12 fixtures
plus the shipped control files; shellcheck is clean.

Your reading of the severity matches ours: the three control files here declare
a single `Build-Depends`, so nothing in the tree was being missed. The parser
advertises all three field names, though, so it has to hold for them.

myungjoo:

Read both prior review rounds and the author's replies in full before starting, then re-reviewed the entire diff from scratch (not just the delta since round 2), and specifically stress-tested the `started`/separator logic that commit 1b50d0bc introduced in `extract_clauses()`.

## Round 1 nit (multiarch qualifier): confirmed resolved

`name=${name%%:*}` in `debian-control.sh` strips the qualifier before both `PKGNAME_RE` and `TOOLCHAIN_RE`. Re-verified `gcc-13:native | gcc-12:native` is still rejected and `gcc-13:native | gcc:native` is accepted, matching commit 23fb5599's own before/after log.

## Round 2 major (field-boundary false negative): confirmed resolved, and the fix holds up under independent adversarial testing

I ran `test_debian_control.sh` locally (git-bash on Windows and WSL) — all 12 fixtures plus the shipped `debian/control*` pass — and reran `shellcheck` (koalaman/shellcheck:stable via Docker) on both scripts: clean, no warnings.

I then wrote my own set of adversarial control-file fragments, independent of the fixture set, targeting exactly the `started` variable and the synthetic `,` separator that 1b50d0bc added:

- `Build-Depends-Arch` then `Build-Depends-Indep`, both with a hidden versioned-only group and no trailing comma at the boundary — both groups correctly flagged (two separate errors, not merged).
- A trailing comma already present right before the next field header (double separator: the real comma plus the synthetic one) — produces one harmless empty clause, correctly filtered by the existing empty-clause skip, no false positive.
- A blank line (paragraph break) between two `Build-Depends*`-named fields — still detected correctly; `started` intentionally does not reset on blank lines (only `collecting` does), and that is correct: it only ever adds a harmless extra separator, never merges two fields.
- A versioned-only group folded across a continuation line via `|` (`gcc-9 |` / ` gcc-8,`) — detected correctly.
- Two, and three, empty `Build-Depends*` fields with no content, followed by the field that has the hidden group — detected correctly.
- CRLF line endings across the same two-field, no-trailing-comma boundary — detected correctly (the per-clause `tr -d '\r'` in `check_file` handles it regardless of where the separator concatenation happens).
- A duplicate `Build-Depends:` field (malformed, but the parser shouldn't crash or silently drop data) — detected correctly.
- File with no trailing newline at EOF — detected correctly.
- Malformed alternative syntax (`gcc-9 || gcc-8`, `gcc-9 | gcc-8 |`) — no crash, still flagged (over-cautious, not a correctness problem).

I could not construct a false negative or a false positive against the fixed parser. The fix is minimal (3 lines: `if (started) { printf "," }; started = 1`), correctly scoped to the reported defect, and the accompanying 3 new fixtures cover both orderings plus the "three fields, all satisfied" true-negative case. Commit 1b50d0bc has a proper `Signed-off-by` and Doxygen-comment update, consistent with this repo's commit conventions (CLAUDE.md).

## New finding (minor, pre-existing, not introduced by round 2/3 fix)

**Severity: minor / nit.**

Debian control field names are case-insensitive per the format's own spec, but the header-matching regex in `debian-control.sh` (`^Build-Depends(-Arch|-Indep)?:`) is case-sensitive. A control file using non-canonical casing, e.g.:

```
Source: nnstreamer
build-depends: gcc-9 | gcc-8
Standards-Version: 3.9.6
```

is not recognized as a target field at all and passes silently (verified: exit 0, no error emitted). This predates both prior review rounds — it was already present in the original checker (commit 0ba71539) and is untouched by 23fb5599 or 1b50d0bc — so it is not a regression from either fix. I flag it only because the task explicitly asked to probe field-name casing as part of real Debian control syntax. In practice this is very low risk: every control file in this repo, and essentially every real-world Debian/Ubuntu control file (dpkg-buildpackage, dh_make templates, lintian, etc.) always uses the canonical `Build-Depends`/`Build-Depends-Arch`/`Build-Depends-Indep` casing, so this would only bite a hand-written file that deviates from universal convention. Not blocking; a fast-follow could case-fold the comparison (e.g. match with `tolower($0)`) if the maintainer wants to close the gap.

## Rest of the diff

No other issues. `debian/control`, `debian/control.debian`, `debian/control.ubuntu.ppa`, `.github/workflows/static.check.yml`, and `Documentation/getting-started-ubuntu-ppa.md` are unchanged since round 1/2 review and were already verified there (vendor-swap consistency in `debian/rules`, Tizen/`nnstreamer.spec` isolation, CI guard pattern matching existing steps, doc links resolving). Scope and size remain appropriate for the stated problem (#4848's repository-side fix) plus a proportionate regression-test harness; no unrelated module is touched.

## Assessment against the review checklist

- **Solves the stated problem**: yes — unchanged from prior rounds.
- **Regression risk to other modules**: none. Confined to `debian/control*`, the new checker + self-test, CI wiring, and one doc file; no C/C++/build-system code touched.
- **Size vs. scope**: appropriate; the round-2 follow-up fix is 3 lines of logic plus 3 targeted fixtures.
- **Test coverage**: strong, and now specifically pins the field-boundary bug in both orderings plus a true-negative multi-field case. My own independent adversarial testing (listed above) found no gap in the fixed logic.
- **CI enforcement**: yes, same blocking wiring as before; a regression here would fail CI before merge.
- **Architecture/API/doc**: no architecture or API change; the doc update is already included.

## Merge recommendation

Both prior findings (round 1 nit, round 2 major) are correctly and completely resolved, with no new regression introduced by either fix. The one new observation (case-sensitive field-name matching) is minor, pre-existing, and not worth blocking on. **No blocker or major issues remain — safe to merge as-is.**

myungjoo:

Thanks for the adversarial pass on the `started` logic — the double-separator,
paragraph-break, folded-`|`, CRLF and duplicate-field cases were exactly the
ones worth probing.

The field-name case nit is fixed in 44f2052e rather than left as a known gap.
You are right that it predates the last two commits; it is also a one-line
change, so carrying it forward as a follow-up was not worth it. The header
match now folds case:

```awk
collecting = (tolower($0) ~ /^build-depends(-arch|-indep)?:/)
```

with a fixture pinning that a lowercase `build-depends:` carrying a
versioned-only group is rejected instead of skipped.

`tolower()` is POSIX rather than a gawk extension, but since the `awk` on a
runner is whatever the alternatives system points at, the suite was run under
all three implementations available on Ubuntu:

```
=== awk -> /usr/bin/mawk ===           PASS count: 14   suite: OK
=== awk -> /usr/bin/gawk ===           PASS count: 14   suite: OK
=== awk -> /usr/bin/original-awk ===   PASS count: 14   suite: OK
```

That is 13 fixtures plus the shipped control files. shellcheck remains clean.

myungjoo:

Read all three prior review rounds and the author's replies in full before starting. Re-reviewed the entire diff independently from scratch (not just the delta since round 3), and specifically stress-tested commit 44f2052e (`tolower()` field-name matching), which is this round's stated focus, plus its interaction with the round-2/3 field-separator fix.

## Round 1 nit (multiarch qualifier, 23fb5599): still resolved

Re-confirmed `gcc-13:native | gcc-12:native` is rejected and `gcc-13:native | gcc:native` is accepted. No change to this logic since round 3.

## Round 2 major (field-boundary false negative, 1b50d0bc): still resolved

Re-confirmed with fresh two-field, no-trailing-comma fixtures in both orderings — no merge/false-negative.

## Round 3 minor (case-sensitive field names) → fixed in 44f2052e: verified correct, and it holds up under adversarial testing

Checked out the pre-fix (1b50d0bc) and post-fix (44f2052e) versions of `debian-control.sh` and ran the new fixture's body directly against both:

```
=== PRE (1b50d0bc), field "build-depends:" lowercase ===
The debian control build-dependency check has passed.   exit=0   (bug: silently skipped)
=== POST (44f2052e) ===
::error::...: 'gcc-9 | gcc-8' lists only version-suffixed toolchain packages...
exit=1   (correctly caught)
```

This confirms the new fixture is a genuine regression test, not just a shape check: it fails on the pre-fix tree and passes on the post-fix tree, per this repo's CLAUDE.md bug-fix test policy.

I then went beyond the shipped fixture to look specifically for the two failure modes a case-fold change can introduce — over-matching (false positives on fields that merely look similar) and under-matching in combination with the earlier separator fix:

- `BUILD-DEPENDS-INDEP:` (all caps) and `Build-depends-Arch:` (mixed case) → both correctly caught.
- `Build-Depends-Foo:` (invented field, shares the prefix but isn't a real Build-Depends variant) → correctly **not** matched (exit 0). The anchored regex `^build-depends(-arch|-indep)?:` only accepts the three real suffixes immediately before the colon, so case-folding didn't widen the match to unrelated fields.
- `Build-Conflicts:` (a real, different Debian field) → correctly not matched.
- Two-field fixtures combining mixed case **and** the missing-trailing-comma boundary case from 1b50d0bc (e.g. `build-depends: ninja-build` followed by `BUILD-DEPENDS-INDEP: gcc-9 | gcc-8` with no separating comma) → correctly caught in both field orderings. The `started`/synthetic-separator logic and the case-fold now compose correctly.
- Confirmed `tolower()` is scoped only to the field-name decision (`collecting = (tolower($0) ~ ...)`) and is not applied to the extracted clause content itself — an uppercase, invalid package name (`GCC-9 | GCC`) is still correctly rejected as an invalid package name, not silently normalized. A less careful case-fold could have leaked into content validation and broken that existing rule; it doesn't here.
- Ran the full fixture suite (`test_debian_control.sh`) against the actual files in this repository under both `mawk` and `gawk` (switched via a `PATH`-shadowing `awk` symlink, since this environment has no `sudo`/`update-alternatives` access to physically swap the system default): all 13 fixtures plus the shipped `debian/control*` files pass under both. I could not get access to `original-awk`/Docker in this sandbox to reproduce the third implementation the author reported, but `tolower()` is POSIX (present since POSIX.1-2001) and behaved identically on the two implementations I could test, so I have no reason to doubt the author's third data point.
- Reran `shellcheck` was not possible in this sandbox (no local install, no Docker); the diff to `debian-control.sh` is a single-line regex change with no new shell constructs, so the risk of a new shellcheck finding here is negligible, but noting the gap for completeness.

No false positive or false negative found in the case-fold fix or its interaction with the prior two fixes.

## Rest of the diff (independent full re-read)

`debian/control`, `debian/control.debian`, `debian/control.ubuntu.ppa`, `.github/workflows/static.check.yml`, and `Documentation/getting-started-ubuntu-ppa.md` are byte-identical to what was already verified in rounds 1–3 (vendor-swap consistency in `debian/rules`, Tizen/`nnstreamer.spec` isolation, CI guard pattern, doc links). Comments in `debian-control.sh` (header block, `extract_clauses` docstring) describe the fields generically as "Build-Depends*" and don't assert case-sensitivity anywhere, so there's no stale-comment/code mismatch left behind by the three follow-up commits. Commit 44f2052e carries a proper `Signed-off-by` and is scoped to exactly the reported gap (2 lines in the checker, 6 lines of fixture).

## Assessment against the review checklist

- **Solves the stated problem**: yes — unchanged from prior rounds; independently re-verified again this round.
- **Regression risk to other modules**: none. This round's change is confined to one regex in one CI helper script plus its own test fixture; no C/C++ or build-system code touched.
- **Size vs. scope**: appropriate — a 1-line functional fix plus 1 pinning fixture, matching the minor severity assigned in round 3.
- **Test coverage**: the new fixture is a real regression test (verified fails-before/passes-after above), and it composes correctly with the existing 12 fixtures rather than duplicating or contradicting them.
- **CI enforcement**: yes, same blocking wiring verified in earlier rounds; unchanged this round.
- **Architecture/API/doc**: no architecture or API change; nothing further needed.

## Conclusion

All three prior findings (round 1 nit, round 2 major, round 3 minor) are correctly and completely resolved across their respective follow-up commits, and this round's independent adversarial pass on 44f2052e — including its interaction with the earlier two fixes — found no new false positive, false negative, or regression. I did not find anything else worth raising in a full independent re-read of the diff.

**No blocker, major, minor, or nit issues remain from this review. Safe to merge as-is.**

myungjoo:
[Status summary relayed on behalf of the author by a Claude agent]

All 18 checks pass and the review is settled, so this is out of draft.

`pdebuild` is the one worth recording, because it is the end-to-end check of
the claim that the fallback is safe on a release that still carries `gcc-9`.
Building the full package set on jammy with the patched control file:

```
C compiler for the host machine:   cc (gcc 11.2.0 "cc (Ubuntu 11.2.0-19ubuntu1) 11.2.0")
C++ compiler for the host machine: c++ (gcc 11.2.0 ...)

gcc-9 installs: 0

dpkg-deb: building package 'nnstreamer'
dpkg-deb: building package 'nnstreamer-core'
dpkg-deb: building package 'nnstreamer-python3'
dpkg-deb: building package 'nnstreamer-onnxruntime'
dpkg-deb: building package 'nnstreamer-test-dev'
...
```

`gcc-9` is no longer pulled into the chroot, since `build-essential` already
satisfies the new tail alternative, and the compiler is unchanged: the distro
default gcc-11, exactly what the pre-change PPA jammy build used. The whole
package set still builds.

The checker and its self-test also ran on the runner rather than taking the
merge-ref skip path, with all 13 fixtures plus the shipped control files:

```
/Checker/ debian control build dependencies
  Checking .../debian/control
  Checking .../debian/control.debian
  Checking .../debian/control.ubuntu.ppa
  The debian control build-dependency check has passed.
/Checker/ debian-control self-test
  PASS: versioned-only gcc clause is rejected (exit 1)
  ...
  PASS: a field name is matched regardless of case (exit 1)
```

Review history for anyone picking this up: three rounds raised findings, all in
the checker script added here rather than in the one-line packaging fix, and
all were resolved in this PR — multiarch qualifiers (23fb5599), a field-boundary
merge that hid a group (1b50d0bc), and case-sensitive field names (44f2052e).
Round four found nothing.

#4848 is referenced without a closing keyword on purpose. Merging this does not
publish anything; the Launchpad recipes and the PPA dependency stack still need
the work listed under "Not covered here" in the description, and Ubuntu 25.10
named in the issue has since gone end-of-life, so 26.04 LTS is the practical
target.

myungjoo:
[Status note relayed on behalf of the author by a Claude agent]

Rebased onto current main and force-pushed. The branch had gone
`CONFLICTING` because 7cf17438 landed on main after this PR was opened and
inserted its own step at the same point in `static.check.yml` — immediately
before `/Checker/ shellcheck for shell scripts`, which is exactly where this
PR adds its two steps.

Resolved by keeping all three, in this order:

```
/Checker/ prohibited-words self-test
/Checker/ SSAT install list is in sync          <- from 7cf17438
/Checker/ debian control build dependencies     <- this PR
/Checker/ debian-control self-test              <- this PR
/Checker/ shellcheck for shell scripts
```

No content was dropped from either side. One wording fix went with it: the
comment on the first new step said "the same merge-ref guard as the two steps
above", which is no longer the count now that the SSAT step sits between them,
so it now reads "the steps above".

Re-verified against the rebased tree exported with LF endings, which is what
the runner sees:

```
debian-control.sh          PASS
test_debian_control.sh     PASS  (13 fixtures + the shipped control files)
ssat-install-list.sh       PASS  (main's checker, unaffected)
```

The four commits are unchanged apart from the conflict resolution in the first
one. CI is re-running from scratch on the new SHAs.

PR #4897, opened 2026-08-28
[Join] End the output when the last awaited sink pad is unlinked

### LINKED ISSUE #4886: [Join] Unlinking the last awaited sink pad leaves the output without EOS
Follow-up from #4875, which added EOS aggregation to `join`. That PR fixed the original data-loss bug; this is a hole it leaves behind, filed so the investigation already done is not repeated.

## The defect

`gst_join_all_sinkpads_eos()` (`gst/join/gstjoin.c`) decides whether every sink pad carrying a stream has ended, and deliberately skips pads that are not linked — waiting for a pad nothing can arrive on would leave the output without EOS forever.

The problem is *when* that decision is taken. It is only re-examined from two places:

- `gst_join_pad_event()`, on `GST_EVENT_EOS`
- `gst_join_release_pad()`, when a request pad is released

Unlinking a sink pad changes the answer but triggers neither. So the outcome depends on the order of operations:

| order | result |
|---|---|
| unlink `sink_0`, then `sink_1` gets EOS | EOS is forwarded — the EOS on `sink_1` re-examines the set, `sink_0` is now unlinked and skipped. Covered by `join.eosAfterUnlink`. |
| `sink_1` gets EOS, then unlink `sink_0` | **no EOS, ever.** The EOS on `sink_1` finds `sink_0` still linked and does not forward. The unlink runs no join code at all, so nothing re-examines the set. |

The second case is a silent hang: the pipeline never reaches EOS and the application gets no indication why.

## Reproduction

Confirmed on `join` as of #4875 (head `6f2b88a1`). Add to `tests/gstreamer_join/unittest_join.cc` — it fails with the bus timing out after 10 s:

```c
TEST (join, eosAfterUnlinkLast)
{
  /* ... standard join_pipeline_desc setup: appsrc_0 -> sink_0, appsrc_1 -> sink_1 ... */
  ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);

  /* sink_0 carries a stream */
  gst_app_src_push_buffer (GST_APP_SRC (appsrc_0), <buffer>);
  g_usleep (100000);

  /* the other branch ends while sink_0 is still linked, so join has to wait */
  gst_app_src_end_of_stream (GST_APP_SRC (appsrc_1));
  EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);

  /* unlinking sink_0 leaves nothing to wait for */
  sinkpad_0 = gst_element_get_static_pad (join_handle, "sink_0");
  peer_0 = gst_pad_get_peer (sinkpad_0);
  gst_pad_unlink (peer_0, sinkpad_0);

  /* fails today: times out instead of receiving EOS */
  EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
}
```

The test was written and run during #4875 but deliberately not committed, since it asserts behaviour the element does not have.

## Why it was not fixed in #4875

The obvious fix — a `gst_pad_set_unlink_function()` on each sink pad that re-runs the same decision — does not work as written. Two findings from trying it, both measured rather than reasoned:

**1. The unlink callback runs holding that pad's object lock.** Verified with a `GST_OBJECT_TRYLOCK` probe inside the callback:

```
PROBE entered for sink_0
PROBE   pad object lock: HELD BY CALLER
PROBE   element object lock: FREE
```

`gst_join_all_sinkpads_eos()` walks every sink pad and calls `gst_pad_is_linked()` on each, which takes that pad's object lock. Called from the unlink callback it re-enters the lock already held for the pad being unlinked, and `GMutex` is not recursive — instant deadlock. Observed as `unittest_join` hanging in `join.normal0`, which only tears a pipeline down.

This constraint is now recorded on the function itself (`6f2b88a1`), so a new caller is warned at the point of definition:

```c
 * @note must be called with the JOIN_LOCK, and with no sink pad's object lock
 *       held: gst_pad_is_linked() takes it, on every sink pad in turn. A pad's
 *       unlink function runs with that pad's object lock held, so calling this
 *       from one deadlocks. Unlinked pads are ignored, as no stream can ever
 *       end on them.
```

Note that the scan used the bare `GST_PAD_IS_LINKED` macro until #4875 switched it to `gst_pad_is_linked()` on review. Reverting that one line is one of the ways out.

**2. Avoiding the deadlock is not sufficient.** With the bare macro there is no deadlock, but `eosAfterUnlinkLast` still fails. At the point the callback runs the peer is already cleared, so the scan should see `sink_0` as unlinked and forward — it does not, and the reason was not diagnosed. **This is where the real work is.**

## What a fix needs

1. A re-examination hook on unlink (`gst_pad_set_unlink_function()`, registered in `gst_join_request_new_pad()`).
2. A way for the scan to run from that context without taking the unlinking pad's object lock — pass the pad down and treat it as unlinked without querying, or revert that one line to the bare macro, or restructure the scan.
3. Diagnose finding 2 above.
4. Commit `eosAfterUnlinkLast` as the regression test.

## Alternative worth considering

Rather than reacting to unlink, the element could decide at EOS time that a pad which has been unlinked *since it last carried data* is retired, or expose the choice as a property. The current contract — documented in the `SECTION:element-join` block as of #4875 — is that releasing the pad is the supported way to retire a branch, and unlinking alone is not:

```
 * The set is only re-examined when a sink pad receives EOS or is released, so
 * retiring a branch mid-stream means releasing its pad: unlinking alone drops
 * it from the set without ending the output, and the output ends at whatever
 * comes next.
```

So this is a gap between a supported operation and a reasonable expectation, not a regression: before #4875, `join` forwarded the first EOS from whichever pad happened to be active regardless of link state, so no ordering hung.

myungjoo:
Fixes #4886.

## The defect

`gst_join_all_sinkpads_eos()` skips sink pads that are not linked, but it only ran on `GST_EVENT_EOS` and on pad release. Unlinking a sink pad changes its answer without running it, so the outcome depended on the order of operations:

| order | result |
|---|---|
| unlink `sink_0`, then `sink_1` gets EOS | EOS forwarded — covered by `join.eosAfterUnlink` |
| `sink_1` gets EOS, then unlink `sink_0` | **no EOS, ever** — a silent hang with no indication why |

## Why `gst_pad_set_unlink_function()` does not work

The issue records that the obvious fix deadlocks, and that avoiding the deadlock still leaves the test failing for reasons that were not diagnosed. Both fall out of `gst_pad_unlink()`:

```c
GST_OBJECT_LOCK (srcpad);
GST_OBJECT_LOCK (sinkpad);
...
GST_PAD_UNLINKFUNC (sinkpad) (sinkpad, tmpparent);   /* <- unlink function */
...
/* first clear peers */
GST_PAD_PEER (srcpad) = NULL;
GST_PAD_PEER (sinkpad) = NULL;
```

- The unlink function runs holding **both** pad object locks, so `gst_pad_is_linked()` on the pad being unlinked self-deadlocks.
- It also runs **before either peer is cleared** — so with the bare `GST_PAD_IS_LINKED` macro instead there is no deadlock, but the scan still reads that pad as linked and forwards nothing. That is the undiagnosed finding 2 in the issue, and it means reverting the `gst_pad_is_linked()` line, listed there as one way out, cannot help.
- Taking the JOIN_LOCK from that context would additionally invert the JOIN_LOCK → pad object lock order that `gst_join_pad_event()` establishes.

## The fix

Three lines further on, `gst_pad_unlink()` does:

```c
GST_OBJECT_UNLOCK (sinkpad);
GST_OBJECT_UNLOCK (srcpad);
g_signal_emit (sinkpad, gst_pad_signals[PAD_UNLINKED], 0, srcpad);
```

The `"unlinked"` signal is emitted after both peers are cleared and both object locks are dropped — exactly the context `gst_join_all_sinkpads_eos()` documents that it needs. Connecting there gives the re-examination hook with **no change to the scan** and none of the three hazards above.

`gst_element_remove_pad()` unlinks before it drops the pad from the element, so `gst_join_release_pad()` now reaches the same handler whenever the released pad was linked; `eos_sent` under the JOIN_LOCK keeps that from forwarding twice, and the release path keeps its own check for a pad released without a peer.

The `SECTION:element-join` contract is updated: unlinking and releasing are now equivalent ways to retire a branch, in any order relative to the other streams ending.

## Tests

`tests/gstreamer_join/unittest_join.cc`, all verified locally against gstreamer 1.20.3:

| case | what it pins | mutant that kills it |
|---|---|---|
| `eosAfterUnlinkLast` | the issue's repro, the reverse order of `eosAfterUnlink` | **fails without the fix** — 10 s bus timeout |
| `noEosOnUnlinkWhileRunning` | unlinking retires a branch, it does not end the output | fails against a handler forwarding EOS on any unlink |
| `noEosOnUnlinkWhileOthersRun` | three branches; one retired, another still linked and running | fails against the same mutant |

`noEosOnUnlinkWhileOthersRun` uses the three-into-one shape the `datarepo` and `custom-easy` filter fixtures use, where an output cut short would silently truncate a recording — that is the cross-module regression this change could plausibly cause, so it is guarded directly.

## Verification

- `unittest_join`: 13/13 pass; `eosAfterUnlinkLast` reproduced failing on the pre-fix tree and passing after.
- Consumers of the `join` element — `unittest_datareposink`, `unittest_datareposrc`, `unittest_filter_custom` — all pass.
- Full `meson test -C build`: **21/21 suites OK**.
- No new compiler warnings; `gst-indent` and `clang-format` produce no diff.

### DIFF
--- a/gst/join/gstjoin.c
+++ b/gst/join/gstjoin.c
@@ -24,6 +24,8 @@
  * stream can end on it; the output would otherwise never see EOS. Either one
  * re-examines the set, so a branch may be retired mid-stream by unlinking it or
  * by releasing its pad, in any order with respect to the other streams ending.
+ * Retiring every branch without any of them having ended leaves the output
+ * open, as there is then no EOS to forward.
  * <refsect2>
  * <title>Example launch line</title>
  * gst-launch-1.0 ... (input stream 0) ! join.sink_0 \
--- a/gst/join/gstjoin.c
+++ b/gst/join/gstjoin.c
@@ -21,10 +21,9 @@
  * All capabilities (input stream i and output stream) should be the same.
  * EOS reaches the output pad only after every linked input stream has ended.
  * A sink pad that is unlinked or released is no longer waited for, since no
- * stream can end on it; the output would otherwise never see EOS. The set is
- * only re-examined when a sink pad receives EOS or is released, so retiring a
- * branch mid-stream means releasing its pad: unlinking alone drops it from the
- * set without ending the output, and the output ends at whatever comes next.
+ * stream can end on it; the output would otherwise never see EOS. Either one
+ * re-examines the set, so a branch may be retired mid-stream by unlinking it or
+ * by releasing its pad, in any order with respect to the other streams ending.
  * <refsect2>
  * <title>Example launch line</title>
  * gst-launch-1.0 ... (input stream 0) ! join.sink_0 \
@@ -240,7 +239,8 @@ forward_sticky_events (GstPad * sinkpad, GstEvent ** event, gpointer user_data)
  * @note must be called with the JOIN_LOCK, and with no sink pad's object lock
  *       held: gst_pad_is_linked() takes it, on every sink pad in turn. A pad's
  *       unlink function runs with that pad's object lock held, so calling this
- *       from one deadlocks. Unlinked pads are ignored, as no stream can ever
+ *       from one deadlocks; gst_join_sinkpad_unlinked() is where an unlink is
+ *       picked up instead. Unlinked pads are ignored, as no stream can ever
  *       end on them.
  */
 static gboolean
@@ -264,6 +264,35 @@ gst_join_all_sinkpads_eos (GstJoin * sel, gboolean * any_eos)
     *any_eos = any;
 
   return all;
+}
+
+/**
+ * @brief re-examine the EOS set once a sink pad has been unlinked
+ * @note Connected to the pad's "unlinked" signal rather than installed with
+ *       gst_pad_set_unlink_function(). gst_pad_unlink() calls the unlink
+ *       function while it still holds both pad object locks and before it
+ *       clears either peer, so a scan from there deadlocks on the pad it is
+ *       asked about and, once that is dodged, still reads the pad as linked.
+ *       The signal is emitted after both peers are cleared and both locks are
+ *       dropped, which is what gst_join_all_sinkpads_eos() needs.
+ */
+static void
+gst_join_sinkpad_unlinked (GstPad * pad, GstPad * peer, gpointer user_data)
+{
+  GstJoin *sel = GST_JOIN (user_data);
+  gboolean send_eos, any_eos = FALSE;
+  (void) peer;
+
+  GST_JOIN_LOCK (sel);
+  send_eos = !sel->eos_sent
+      && gst_join_all_sinkpads_eos (sel, &any_eos) && any_eos;
+  sel->eos_sent |= send_eos;
+  GST_JOIN_UNLOCK (sel);
+
+  if (send_eos) {
+    GST_DEBUG_OBJECT (pad, "last awaited sink pad unlinked, forwarding EOS");
+    gst_pad_push_event (sel->srcpad, gst_event_new_eos ());
+  }
 }
 
 /**
@@ -748,6 +777,8 @@ gst_join_request_new_pad (GstElement * element, GstPadTemplate * templ,
   gst_pad_set_chain_function (sinkpad, GST_DEBUG_FUNCPTR (gst_join_pad_chain));
   gst_pad_set_iterate_internal_links_function (sinkpad,
       GST_DEBUG_FUNCPTR (gst_join_pad_iterate_linked_pads));
+  g_signal_connect_object (sinkpad, "unlinked",
+      G_CALLBACK (gst_join_sinkpad_unlinked), sel, (GConnectFlags) 0);
 
   GST_OBJECT_FLAG_SET (sinkpad, GST_PAD_FLAG_PROXY_CAPS);
   GST_OBJECT_FLAG_SET (sinkpad, GST_PAD_FLAG_PROXY_ALLOCATION);
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -684,6 +684,199 @@ TEST (join, eosAfterUnlink)
 }
 
 /**
+ * @brief Test that unlinking the sink pad the join is waiting for ends the
+ *        output stream.
+ * @detail The reverse order of join.eosAfterUnlink. sink_0 is still linked and
+ *         carrying a stream when the other branch ends, so that EOS has to
+ *         wait; nothing further arrives on any pad, which leaves the unlink as
+ *         the only thing that can complete the set.
+ */
+TEST (join, eosAfterUnlinkLast)
+{
+  GstElement *appsrc_0, *appsrc_1, *join_handle, *sink_handle;
+  GstPad *sinkpad_0, *srcpad_0;
+  gint idx = 0;
+
+  GstElement *pipeline = gst_parse_launch (join_pipeline_desc, NULL);
+  ASSERT_NE (pipeline, nullptr);
+
+  appsrc_0 = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc_0");
+  ASSERT_NE (appsrc_0, nullptr);
+  appsrc_1 = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc_1");
+  ASSERT_NE (appsrc_1, nullptr);
+  join_handle = gst_bin_get_by_name (GST_BIN (pipeline), "join");
+  ASSERT_NE (join_handle, nullptr);
+  sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_NE (sink_handle, nullptr);
+  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, (gpointer) &idx);
+
+  data_received = 0;
+  ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  idx = 0;
+  ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_0),
+                 gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
+      GST_FLOW_OK);
+  g_usleep (100000);
+
+  EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc_1)), GST_FLOW_OK);
+  EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
+
+  sinkpad_0 = gst_element_get_static_pad (join_handle, "sink_0");
+  ASSERT_NE (sinkpad_0, nullptr);
+  srcpad_0 = gst_pad_get_peer (sinkpad_0);
+  ASSERT_NE (srcpad_0, nullptr);
+  EXPECT_TRUE (gst_pad_unlink (srcpad_0, sinkpad_0));
+  gst_object_unref (srcpad_0);
+  gst_object_unref (sinkpad_0);
+
+  EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
+  EXPECT_EQ (1, data_received);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  gst_object_unref (sink_handle);
+  gst_object_unref (join_handle);
+  gst_object_unref (appsrc_1);
+  gst_object_unref (appsrc_0);
+  gst_object_unref (pipeline);
+}
+
+/**
+ * @brief Test that unlinking a sink pad does not end a stream that is running.
+ * @detail Unlinking retires a branch; it does not end the output on its own.
+ *         With no stream ended anywhere there is nothing to forward, and the
+ *         surviving branch has to keep streaming afterwards.
+ */
+TEST (join, noEosOnUnlinkWhileRunning)
+{
+  GstElement *appsrc_0, *appsrc_1, *join_handle, *sink_handle;
+  GstPad *sinkpad_0, *srcpad_0;
+  gint idx = 0;
+
+  GstElement *pipeline = gst_parse_launch (join_pipeline_desc, NULL);
+  ASSERT_NE (pipeline, nullptr);
+
+  appsrc_0 = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc_0");
+  ASSERT_NE (appsrc_0, nullptr);
+  appsrc_1 = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc_1");
+  ASSERT_NE (appsrc_1, nullptr);
+  join_handle = gst_bin_get_by_name (GST_BIN (pipeline), "join");
+  ASSERT_NE (join_handle, nullptr);
+  sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_NE (sink_handle, nullptr);
+  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, (gpointer) &idx);
+
+  data_received = 0;
+  ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  idx = 0;
+  ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_0),
+                 gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
+      GST_FLOW_OK);
+  g_usleep (100000);
+
+  sinkpad_0 = gst_element_get_static_pad (join_handle, "sink_0");
+  ASSERT_NE (sinkpad_0, nullptr);
+  srcpad_0 = gst_pad_get_peer (sinkpad_0);
+  ASSERT_NE (srcpad_0, nullptr);
+  EXPECT_TRUE (gst_pad_unlink (srcpad_0, sinkpad_0));
+  gst_object_unref (srcpad_0);
+  gst_object_unref (sinkpad_0);
+
+  EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
+
+  idx = 1;
+  ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_1),
+                 gst_buffer_new_wrapped (_g_memdup (test_frames[1], 192), 192)),
+      GST_FLOW_OK);
+  g_usleep (100000);
+  EXPECT_EQ (2, data_received);
+
+  EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc_1)), GST_FLOW_OK);
+  EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  gst_object_unref (sink_handle);
+  gst_object_unref (join_handle);
+  gst_object_unref (appsrc_1);
+  gst_object_unref (appsrc_0);
+  gst_object_unref (pipeline);
+}
+
+/**
+ * @brief Test that unlinking one of three sink pads does not end the output.
+ * @detail Three branches into one join is the shape the datarepo and
+ *         custom-easy filter fixtures use, where an output cut short would
+ *         silently truncate a recording. Retiring one branch has to leave the
+ *         join waiting for the one that is still linked and still running.
+ */
+TEST (join, noEosOnUnlinkWhileOthersRun)
+{
+  const gchar *desc
+      = "appsrc name=appsrc_0 ! other/tensor,dimension=(string)3:4:2:2,type=(string)int32,framerate=(fraction)0/1 ! join.sink_0 "
+        "appsrc name=appsrc_1 ! other/tensor,dimension=(string)3:4:2:2,type=(string)int32,framerate=(fraction)0/1 ! join.sink_1 "
+        "appsrc name=appsrc_2 ! other/tensor,dimension=(string)3:4:2:2,type=(string)int32,framerate=(fraction)0/1 ! join.sink_2 "
+        "join name=join ! other/tensor,dimension=(string)3:4:2:2, type=(string)int32, framerate=(fraction)0/1 ! "
+        "tensor_sink name=sinkx async=false";
+  GstElement *appsrc[3], *join_handle, *sink_handle;
+  GstPad *sinkpad_0, *srcpad_0;
+  guint received = 0;
+  guint i;
+
+  GstElement *pipeline = gst_parse_launch (desc, NULL);
+  ASSERT_NE (pipeline, nullptr);
+
+  for (i = 0; i < 3; i++) {
+    gchar *name = g_strdup_printf ("appsrc_%u", i);
+    appsrc[i] = gst_bin_get_by_name (GST_BIN (pipeline), name);
+    g_free (name);
+    ASSERT_NE (appsrc[i], nullptr);
+  }
+  join_handle = gst_bin_get_by_name (GST_BIN (pipeline), "join");
+  ASSERT_NE (join_handle, nullptr);
+  sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_NE (sink_handle, nullptr);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_data_cb, (gpointer) &received);
+
+  ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  for (i = 0; i < 3; i++) {
+    ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc[i]),
+                   gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
+        GST_FLOW_OK);
+    g_usleep (100000);
+  }
+
+  EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc[1])), GST_FLOW_OK);
+  EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
+
+  sinkpad_0 = gst_element_get_static_pad (join_handle, "sink_0");
+  ASSERT_NE (sinkpad_0, nullptr);
+  srcpad_0 = gst_pad_get_peer (sinkpad_0);
+  ASSERT_NE (srcpad_0, nullptr);
+  EXPECT_TRUE (gst_pad_unlink (srcpad_0, sinkpad_0));
+  gst_object_unref (srcpad_0);
+  gst_object_unref (sinkpad_0);
+
+  /* sink_2 is still linked and has not ended, so nothing may be forwarded. */
+  EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
+
+  EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc[2])), GST_FLOW_OK);
+  EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
+  EXPECT_EQ (3U, received);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  gst_object_unref (sink_handle);
+  gst_object_unref (join_handle);
+  for (i = 0; i < 3; i++)
+    gst_object_unref (appsrc[i]);
+  gst_object_unref (pipeline);
+}
+
+/**
  * @brief Main GTest
  */
 int
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -696,12 +696,15 @@ typedef struct {
  * @note Every expected frame is in place before the pipeline plays, so the
  *       streaming thread never reads a field the test thread writes. That is
  *       what the index shared with new_data_cb did, and it is also why this
- *       needs no ordering between the count and the frame it selects.
+ *       needs no ordering between the count and the frame it selects. A buffer
+ *       arriving past the frames a test named fails on the bound or on the
+ *       empty slot, rather than being compared against whatever is there.
  */
 static void
 count_checked_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_data)
 {
   JoinCounter *counter = (JoinCounter *) user_data;
+  const gint *expected;
   GstMemory *mem_res;
   GstMapInfo info_res;
   gboolean mapped;
@@ -709,6 +712,8 @@ count_checked_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_dat
   (void) element;
 
   ASSERT_LT (counter->received, G_N_ELEMENTS (counter->expected));
+  expected = counter->expected[counter->received];
+  ASSERT_NE (expected, nullptr);
 
   mem_res = gst_buffer_get_memory (buffer, 0);
   mapped = gst_memory_map (mem_res, &info_res, GST_MAP_READ);
@@ -716,7 +721,7 @@ count_checked_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_dat
   output = (gint *) info_res.data;
 
   for (i = 0; i < 48; i++) {
-    EXPECT_EQ (counter->expected[counter->received][i], output[i]);
+    EXPECT_EQ (expected[i], output[i]);
   }
   gst_memory_unmap (mem_res, &info_res);
   gst_memory_unref (mem_res);
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -695,7 +695,7 @@ TEST (join, eosAfterUnlinkLast)
 {
   GstElement *appsrc_0, *appsrc_1, *join_handle, *sink_handle;
   GstPad *sinkpad_0, *srcpad_0;
-  gint idx = 0;
+  guint received = 0;
 
   GstElement *pipeline = gst_parse_launch (join_pipeline_desc, NULL);
   ASSERT_NE (pipeline, nullptr);
@@ -708,16 +708,14 @@ TEST (join, eosAfterUnlinkLast)
   ASSERT_NE (join_handle, nullptr);
   sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
   ASSERT_NE (sink_handle, nullptr);
-  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, (gpointer) &idx);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_data_cb, (gpointer) &received);
 
-  data_received = 0;
   ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
 
-  idx = 0;
   ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_0),
                  gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
       GST_FLOW_OK);
-  g_usleep (100000);
+  ASSERT_TRUE (wait_pipeline_process_buffers (&received, 1, TEST_TIMEOUT_LIMIT_MS));
 
   EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc_1)), GST_FLOW_OK);
   EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
@@ -731,7 +729,7 @@ TEST (join, eosAfterUnlinkLast)
   gst_object_unref (sinkpad_0);
 
   EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
-  EXPECT_EQ (1, data_received);
+  EXPECT_EQ (1U, received);
 
   EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
 
@@ -752,7 +750,7 @@ TEST (join, noEosOnUnlinkWhileRunning)
 {
   GstElement *appsrc_0, *appsrc_1, *join_handle, *sink_handle;
   GstPad *sinkpad_0, *srcpad_0;
-  gint idx = 0;
+  guint received = 0;
 
   GstElement *pipeline = gst_parse_launch (join_pipeline_desc, NULL);
   ASSERT_NE (pipeline, nullptr);
@@ -765,16 +763,14 @@ TEST (join, noEosOnUnlinkWhileRunning)
   ASSERT_NE (join_handle, nullptr);
   sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
   ASSERT_NE (sink_handle, nullptr);
-  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, (gpointer) &idx);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_data_cb, (gpointer) &received);
 
-  data_received = 0;
   ASSERT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
 
-  idx = 0;
   ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_0),
                  gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
       GST_FLOW_OK);
-  g_usleep (100000);
+  ASSERT_TRUE (wait_pipeline_process_buffers (&received, 1, TEST_TIMEOUT_LIMIT_MS));
 
   sinkpad_0 = gst_element_get_static_pad (join_handle, "sink_0");
   ASSERT_NE (sinkpad_0, nullptr);
@@ -786,12 +782,10 @@ TEST (join, noEosOnUnlinkWhileRunning)
 
   EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
 
-  idx = 1;
   ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_1),
                  gst_buffer_new_wrapped (_g_memdup (test_frames[1], 192), 192)),
       GST_FLOW_OK);
-  g_usleep (100000);
-  EXPECT_EQ (2, data_received);
+  ASSERT_TRUE (wait_pipeline_process_buffers (&received, 2, TEST_TIMEOUT_LIMIT_MS));
 
   EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc_1)), GST_FLOW_OK);
   EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
@@ -846,8 +840,8 @@ TEST (join, noEosOnUnlinkWhileOthersRun)
     ASSERT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc[i]),
                    gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192)),
         GST_FLOW_OK);
-    g_usleep (100000);
   }
+  ASSERT_TRUE (wait_pipeline_process_buffers (&received, 3, TEST_TIMEOUT_LIMIT_MS));
 
   EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (appsrc[1])), GST_FLOW_OK);
   EXPECT_EQ (pop_eos_or_error (pipeline, 300), GST_MESSAGE_UNKNOWN);
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -684,19 +684,19 @@ TEST (join, eosAfterUnlink)
 }
 
 /**
- * @brief Buffer count, and the frame the buffers counted next must carry.
+ * @brief Buffer count, and the frame each buffer must carry in arrival order.
  */
 typedef struct {
-  const gint *expected;
+  const gint *expected[3];
   guint received;
 } JoinCounter;
 
 /**
  * @brief Count a buffer once its payload matches the frame that was pushed.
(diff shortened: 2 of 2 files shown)


myungjoo:
> 이 리뷰는 별도 컨텍스트의 리뷰 에이전트가 작성한 내용을 옮겨 적은 것입니다. (This review was written by a separate review-agent session and is being relayed here verbatim.)

## Scope of this review

Reviewed the diff (`gst/join/gstjoin.c`, `tests/gstreamer_join/unittest_join.cc`, +229/-5) against issue #4886, the prior PR #4875, and the GStreamer 1.20/1.24 `gstpad.c`/`gstelement.c` sources (fetched directly from `gitlab.freedesktop.org` to check the locking/signal-ordering claims made in the PR description). I also built this branch's `gst/join` + `tests/unittest_join` in a local environment and ran the suite both with the fix and with `gst/join/gstjoin.c` reverted to the pre-fix (#4875) state while keeping the new tests, to independently confirm the regression test actually reproduces the bug.

## 1. Does it solve the stated problem? — Yes, verified independently

The core claim ("`gst_pad_unlink()` calls the pad's unlink function while holding both pad object locks and before clearing either peer; the `\"unlinked\"` signal fires afterward, with both locks dropped and both peers cleared") checks out against both `gst/gstreamer` 1.20 and 1.24 `gstpad.c`:

```c
GST_PAD_UNLINKFUNC (sinkpad) (sinkpad, tmpparent);   /* holds both object locks, peers NOT yet cleared */
...
GST_PAD_PEER (srcpad) = NULL;
GST_PAD_PEER (sinkpad) = NULL;
GST_OBJECT_UNLOCK (sinkpad);
GST_OBJECT_UNLOCK (srcpad);
g_signal_emit (sinkpad, gst_pad_signals[PAD_UNLINKED], 0, srcpad);
```

This is exactly the precondition `gst_join_all_sinkpads_eos()` documents needing, so connecting to `"unlinked"` (via `g_signal_connect_object`, in `gst_join_request_new_pad()`) instead of installing an unlink function is the right fix and avoids the deadlock/false-linked-read the issue describes.

I also confirmed `gst_element_remove_pad()` (`gstelement.c`) calls `gst_pad_unlink()` *before* removing the pad from the element's pad list, which is what makes `gst_join_release_pad()` reach the same new handler when the released pad was still linked — the PR's claim about this interaction is accurate, and the pre-existing `eos_sent` guard (checked/set under `GST_JOIN_LOCK` in both the signal handler and `gst_join_release_pad()`) prevents a double EOS push in that overlap case.

**Empirical confirmation:** built `gst/join/libgstjoin.so` + `tests/unittest_join` twice — once with this PR's `gstjoin.c`, once with `gstjoin.c` reverted to the pre-#4897 (#4875) state but keeping the new tests:
- With the fix: **13/13 pass**, including `join.eosAfterUnlinkLast`.
- With the fix reverted: `join.eosAfterUnlinkLast` **fails** (times out waiting for EOS — exactly the bug in #4886), while the other two new tests (`noEosOnUnlinkWhileRunning`, `noEosOnUnlinkWhileOthersRun`) still pass, because they assert the *absence* of a (still absent) behavior — consistent with the PR's description that those two exist to catch an overly-aggressive alternative fix ("forward EOS on any unlink"), not this specific bug.

This is a genuine, verified regression test, not a tautology.

## 2. Regression risk to other modules — low, and directly checked

- The lock-ordering analysis in the PR description (JOIN_LOCK → pad object lock, established in `gst_join_pad_chain()`) matches the code (`GST_JOIN_LOCK` taken, then `GST_OBJECT_LOCK (pad)` taken for `selpad->segment.position` while still holding JOIN_LOCK). The new `"unlinked"` handler never holds a pad object lock when taking JOIN_LOCK, so it doesn't invert this order.
- No API, header (`gstjoin.h` unchanged), or signal/property change — the `"unlinked"` connection is a private implementation detail of the `join` element itself.
- Rebuilt and ran `unittest_filter_custom` (a consumer of the `join`-shaped 3-into-1 fixture pattern called out in the PR body) against the fixed `join` plugin: passes. (Some unrelated `easy-custom` model-not-found CRITICAL logs appeared in that binary in my local environment; unrelated to `join`/this diff.)
- Diff is scoped to exactly 2 files, both directly relevant (`gst/join/gstjoin.c`, its test file). No other module touched.

## 3. Size / scope discipline — good

229 insertions / 5 deletions, single commit, touches only the `join` element and its own test file. Proportionate to the bug being fixed. No unrelated refactors bundled in.

## 4. Test completeness — good, per CLAUDE.md's bug-fix policy

- `join.eosAfterUnlinkLast`: reproduces the exact issue #4886 scenario (reverse order of `join.eosAfterUnlink`), fails without the fix, passes with it — satisfies the "regression test that fails without the fix and passes with it" requirement.
- `join.noEosOnUnlinkWhileRunning` / `join.noEosOnUnlinkWhileOthersRun`: negative-path coverage guarding against an overly broad fix (forwarding EOS on *any* unlink), including the specific 3-branches-into-1 shape the PR calls out as the one other modules (`datarepo`, `custom-easy` filter fixtures) rely on. This is exactly the "future regression in other code touching this contract" detection the task asked to check for.
- All 13 tests in `unittest_join` (10 pre-existing + 3 new) pass together; none of the pre-existing tests needed modification, so no evidence of the fix changing previously-verified behavior.

## 5. Would a future regression be caught by CI? — Yes

`unittest_join` is registered in `tests/meson.build` (`test('unittest_join', unittest_join, ..., timeout: 120)`), which runs under `meson test -C build` — the standard CI gate. A future change that reintroduces the ordering bug, or one that makes `join` forward EOS on every unlink, would fail `join.eosAfterUnlinkLast` or the two negative tests respectively, blocking merge through the existing test-suite gate. No new CI wiring is needed since this reuses the existing `unittest_join` target.

## 6. Documentation / architecture — no gap

No public API or architecture change (no new signals/properties exposed, `gstjoin.h` untouched). The one documentation surface that needed updating — the `SECTION:element-join` gtk-doc block describing when the EOS set is re-examined — was updated in this same PR to state that unlinking and releasing are now equivalent, in either order. This is the only doc that needed to change, and it did.

## Minor / nit observations (non-blocking)

- **nit:** The doc-comment tag `@detail` used in the new test docs (e.g. line ~688, ~747, ~810 of `unittest_join.cc`) is not a standard Doxygen tag (`@details` is), but this matches the convention already established by prior tests in the same file since #4875, so it's not something this PR introduced or should be asked to fix in isolation.
- **nit/informational, not introduced by this PR:** the `any_eos` guard (inherited unchanged from #4875) means that if *every* sink pad is retired purely by unlink/release without any of them ever having received an explicit `GST_EVENT_EOS`, the join never forwards EOS. That's pre-existing, documented-by-design behavior, not a regression from this change, and out of scope for #4886 — just flagging for awareness in case it surfaces as a separate future issue.
- The PR is currently marked **DRAFT** with the **DO NOT MERGE** label. That's a process/workflow state, not a code defect — noting it here only so the merge-readiness verdict below is read in that context.

## Verdict

**From a code-correctness, scope, and test-coverage standpoint: approve.** The fix is well-targeted, its key technical claims about GStreamer's pad-unlink internals were independently verified against the actual 1.20/1.24 source, the regression test was independently confirmed to fail pre-fix and pass post-fix, no other module is touched or was observed to regress, and CI (`meson test`) will catch a future reintroduction of this bug or an overcorrection of it. The only reason this isn't an unqualified "ready to merge" is the repository's own **DRAFT / DO NOT MERGE** status — that should be lifted by the author before this actually merges, per normal process, but I found no code issue that should block it from proceeding.

myungjoo:
Relayed review — this is a transcription of a review produced by an automated review agent, posted here under the requester's own account.

**Re-review scope:** commit 881c4a05 (`[Join] Wait on the buffer count rather than a fixed sleep in the new tests`), the follow-up to the review at #4897 (pullrequestreview-5047894786, APPROVED) that fixed at commit `fb591580`. I re-read the full diff (`gst/join/gstjoin.c`, `tests/gstreamer_join/unittest_join.cc`) and built and ran it locally (gstreamer 1.20.3, meson/ninja, gtest) to check the notes empirically rather than by inspection alone.

### Note 3 (load-sensitive `g_usleep` + fixed-sleep assertion) — resolved

The three new cases (`eosAfterUnlinkLast`, `noEosOnUnlinkWhileRunning`, `noEosOnUnlinkWhileOthersRun`) now call `wait_pipeline_process_buffers(&received, N, TEST_TIMEOUT_LIMIT_MS)` instead of `g_usleep(100000)` before asserting. That helper (`tests/unittest_util.c:192`) polls the counter every 10 ms up to a 10 s timeout — the same pattern already used by `forwardAllBuffers` in this file, so it's not a new idiom, just now applied consistently. This removes the load-sensitivity the previous review flagged: the assertion no longer depends on the buffer having arrived within a fixed wall-clock window.

I verified this isn't merely cosmetic by rebuilding the plugin (`gst/join/libgstjoin.so`) and `tests/unittest_join` from the exact head-of-branch sources and running the suite:

```
[==========] 13 tests from 1 test suite ran. (4701 ms total)
[  PASSED  ] 13 tests.
```

### Note 4 (`any_eos` / SECTION wording) — resolved

The added sentence — "Retiring every branch without any of them having ended leaves the output open, as there is then no EOS to forward." — matches the code exactly: `gst_join_all_sinkpads_eos()` only sets `any_eos = TRUE` when some pad's `selpad->eos` flag is set (i.e. it actually received a real `EOS` event), not merely for being unlinked/retired. Unlinking or releasing every branch without any of them ever receiving EOS leaves `any_eos` false forever, so `send_eos` is never set. The new sentence closes exactly the ambiguity the previous review pointed at in "in any order with respect to the other streams ending."

### Test strength — verified empirically, not just by reading

This was the part I wanted hard evidence for, since switching `new_data_cb` → `count_data_cb` and adding a wait-for-count step touches the actual assertions, not just their timing.

**(a) Is `eosAfterUnlinkLast` still a genuine regression test?** I disabled the fix (made `gst_join_sinkpad_unlinked()` a no-op that never sends EOS, i.e. simulating pre-#4886 behavior — note this could not be done by literally reverting the callback to nothing, since `-Werror=unused-function` catches that; the plugin build treats warnings as errors) and reran:

```
[ RUN      ] join.eosAfterUnlinkLast
../tests/gstreamer_join/unittest_join.cc:731: Failure
  pop_eos_or_error (pipeline, (10000U))
    Which is: 0
  GST_MESSAGE_EOS
    Which is: 1
[  FAILED  ] join.eosAfterUnlinkLast (10321 ms)
```

Confirmed: it fails (with a real 10 s timeout, not a false negative) without the fix, and the other two unlink cases still pass in that state (they don't require the new signal to be wired up correctly, only that it not do the wrong thing) — consistent with what's expected.

**(b) Do `noEosOnUnlinkWhileRunning` / `noEosOnUnlinkWhileOthersRun` still catch an over-broad handler?** I patched the handler to unconditionally forward EOS on any unlink regardless of `gst_join_all_sinkpads_eos()`'s verdict (the specific mutant the original review named as the risk this fix could introduce) and reran:

```
[  FAILED  ] join.noEosOnUnlinkWhileRunning (10121 ms)
[  FAILED  ] join.noEosOnUnlinkWhileOthersRun (10321 ms)
```

Both fail as expected — `pop_eos_or_error(pipeline, 300)` sees the spurious EOS where `GST_MESSAGE_UNKNOWN` was expected. Restoring the real fix and rebuilding brings the suite back to 13/13 green. So the answer to both (a) and (b) is yes — switching the wait mechanism did not weaken the tests' discriminating power; the mutants that mattered are still caught.

### `count_data_cb` vs `new_data_cb` — not a coverage loss

`count_data_cb` only increments a counter; `new_data_cb` additionally maps the buffer and compares its contents against `test_frames[index]`. The commit message says payload integrity is "covered by `join.normal0` and `join.eosAfterAllPads`" — checked, both of those still use `new_data_cb` and assert full payload equality. Since `join`'s chain function is a pure pass-through (it doesn't touch buffer contents), payload correctness is orthogonal to the EOS-ordering behavior these three cases exist to pin, so testing "did N buffers arrive" rather than "do the bytes match" loses no meaningful coverage for what these specific cases are for. As a minor side benefit, moving from the previous shared `static int data_received` global to a per-test local `guint received` removes a bit of incidental shared mutable state (not that it mattered while the suite runs with `is_parallel: false`, but it's a small hygiene improvement).

### Notes 1 and 2 (left unaddressed) — judged reasonable

Both were explicitly called "non-blocking" observations in the original review, not defects:

- **Note 1** (handler also fires during bin teardown): true, and benign as noted — `gst_pad_push_event()` on an inactive src pad in `GST_STATE_NULL` is a no-op. Not fixing this is fine; there's no bug to fix, just a newly-reachable code path worth knowing about.
- **Note 2** (`sel->srcpad` dereferenced outside `JOIN_LOCK`, relying on `GstElement`'s dispose ordering — request pads released before the srcpad is removed, srcpads/sinkpads lists built with `g_list_prepend`): this is correct as far as current GStreamer's `gst_element_dispose()` behavior goes, and matches what I'd expect from that function's well-known implementation (iterate `element->pads`, call `gst_element_release_request_pad()` for template-REQUEST pads before falling through to remove the rest, informed by insertion order). It is an implicit dependency on core GStreamer internals rather than something this code enforces itself, so it's fair to call it fragile-in-principle — but hardening it (e.g., explicitly guarding `sel->srcpad` under the lock, or NULLing it in `dispose`) is a real, if small, unit of extra work with no currently-reachable failure mode behind it. Deferring it as a non-blocking follow-up rather than bundling it into this bug-fix PR is a reasonable scope call, especially given the project's practice of one fix per PR.

Neither omission affects whether this PR correctly fixes #4886.

### Other checks

- **Issue resolution**: `eosAfterUnlinkLast` is the exact reproduction from #4886 and now passes; verified as a real regression test above (fails pre-fix, passes post-fix).
- **Regression risk to other modules**: rebuilt and ran `unittest_datareposink` (12/12), `unittest_datareposrc` (22/22), and `unittest_filter_custom` (6/6) against the PR-head `join` plugin — all pass. These are the closest consumers of `join` (the three-branch-into-one shape mirrored by `noEosOnUnlinkWhileOthersRun`).
- **Style**: `gst-indent` on `gst/join/gstjoin.c` and `clang-format` on `tests/gstreamer_join/unittest_join.cc` both produce no diff against the PR head.
- **Doxygen**: every new/changed function (`gst_join_sinkpad_unlinked`, `count_data_cb`, the three new `TEST` cases) carries a `@brief`/`@detail` block, satisfying the repo's CI-enforced doc requirement.
- **DCO**: commit 881c4a05 carries `Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>`.
- **Doc/API surface**: the only externally-visible change is the `SECTION:element-join` prose; it now accurately documents the `any_eos` edge case, no code-level API changed.
- **Code size**: the diff added by 881c4a05 is small and proportionate (one doc sentence, a handful of test-only mechanical changes to how three existing tests wait/assert) — no scope creep.
- **CI**: at the time of this review most upstream CI checks on `881c4a05` were still queued/pending (MacOS build, DCO, spell-check, and the python CodeQL analyze job had already passed; the Valgrind, LLVM/Clang, RISC-V, and Tizen GBS jobs were pending). Nothing observed suggests they won't pass, based on the local build/test results above, but I did not wait for them to complete.

### Verdict

Both non-blocking notes from the previous review (3 and 4) are properly and verifiably resolved by 881c4a05, with no loss of test discriminating power — confirmed empirically via mutation-style testing (disabling the fix reproduces the original hang; over-broadening the fix is caught by the two `noEosOnUnlink*` cases). Notes 1 and 2 were reasonably left as non-blocking follow-ups rather than bundled into this fix. I found no new blocker, major, or minor issues introduced by this commit.

**This PR is mergeable as-is**, contingent only on the still-pending upstream CI jobs (Valgrind, LLVM/Clang, RISC-V, Tizen GBS) finishing green, which is a process gate rather than a code concern raised by this review.

myungjoo:
Relayed review — this is a transcription of a review produced by an automated review agent, posted here by the bot account.

## Re-review at head `c90b2045` (previous review: `881c4a05`, approved)

One new commit, `c90b2045` ("[Test/Join] Check the payload again in the unlink cases"), `+56/-12` in `tests/gstreamer_join/unittest_join.cc` only. `gst/join/gstjoin.c` and `tests/meson.build` are untouched since `881c4a05`, so the lock-ordering and re-entrancy analysis from the first review still applies unchanged.

This commit is exactly the follow-up on the one non-blocking note left open last round: swapping the three new tests onto `count_data_cb` (for `wait_pipeline_process_buffers()`) had also dropped their payload check, so a bug that forwarded the wrong branch's buffer after an unlink would have passed silently.

## What changed

- New `JoinCounter { const gint *expected; guint received; }` and `count_checked_data_cb()`: maps the buffer, compares all 48 ints against `counter->expected`, unmaps, and **increments `received` last**.
- `eosAfterUnlinkLast`, `noEosOnUnlinkWhileRunning`, `noEosOnUnlinkWhileOthersRun` now use this instead of `count_data_cb`, with `counter.expected` re-pointed at the post-unlink frame before the next push.

## Race-freedom claim — verified, with one caveat

The author's argument is that because the count is raised *after* the comparison, and the test thread only rewrites `expected` after observing the new count, the writer and the reader of `expected` never overlap. I walked this and additionally verified it empirically (see below):

- Within one callback invocation, "read `expected[i]` 48 times" strictly precedes "`received++`" in program order — this holds regardless of memory model, since it's single-threaded control flow inside one function.
- `gst_tensor_sink_render_buffer()` emits `"new-data"` synchronously from `GstBaseSink::render`, i.e. on the sink pad's own streaming thread — there's no idle-source indirection that could reorder or duplicate emission relative to the buffer sequence.
- Buffers are pushed one at a time with an intervening `wait_pipeline_process_buffers()`, so two invocations of the callback never overlap in these tests.
- **Caveat**: the handoff itself (`received` written by the streaming thread, read by the test thread's busy-wait loop in `wait_pipeline_process_buffers()`) is a plain, non-atomic, unfenced access on both sides — a data race by the strict C++ memory model. The "read-then-write, write-then-read" ordering the author relies on is guaranteed on x86-64 by TSO (loads are never reordered past a later store from the same thread), which is what makes the claim actually hold on the hardware this runs on. It is **not** portable to a weakly-ordered architecture without a fence/atomic. In this project that risk is moot in practice: `tests/meson.build` runs `unit_test 1` only on the `x86_64` GBS matrix leg (`.github/workflows/gbs_build.yml`); `armv7l`/`aarch64` legs build with `unit_test 0` and never execute the test binary. This is the same category of pre-existing "benign race" the second review already looked at and declined to block on for `count_data_cb`/`wait_pipeline_process_buffers()` itself — this commit doesn't make that pattern worse, it just extends it to one more field. Non-blocking, consistent with the prior call.

## Empirical verification (built locally, WSL x86_64, all 13 cases green at this head)

1. **`eosAfterUnlinkLast` still needs the element fix.** Reverted `gst/join/gstjoin.c` to the pre-`fb591580` version (test file unchanged at `c90b2045`): the case times out waiting for EOS as expected —
   `pop_eos_or_error(pipeline, 10000U)` returns `0` (`GST_MESSAGE_UNKNOWN`), never `GST_MESSAGE_EOS`.
2. **Both `noEosOnUnlink*` still catch an over-broad handler.** Patched `gst_join_sinkpad_unlinked()` to forward EOS on *any* unlink (`send_eos = !sel->eos_sent`, dropping the `all-sinkpads-eos && any_eos` condition): both `noEosOnUnlinkWhileRunning` and `noEosOnUnlinkWhileOthersRun` fail, each catching the spurious early EOS via `pop_eos_or_error(pipeline, 300)` returning `GST_MESSAGE_EOS` where `GST_MESSAGE_UNKNOWN` was expected.
3. **The restored payload check has teeth.** Pointed `noEosOnUnlinkWhileRunning`'s post-unlink `counter.expected` at `test_frames[0]` instead of `test_frames[1]` (element code left correct): the case fails with 48 `EXPECT_EQ` mismatches (`1101` vs `2101`, …, `1224` vs `2224`) — i.e. it does verify that the buffer arriving after the unlink is branch 1's data, not merely a second buffer.
4. **Stability**: ran `eosAfterUnlinkLast` + both `noEosOnUnlink*` with `--gtest_repeat=25` at this head (75 individual runs) — all passed, consistent ~1244 ms per round, no flakiness observed.

All three of these match exactly what the commit message claims ("verified by mutation").

## Style / hygiene

- `clang-format` (repo `.clang-format`) reports no diff on the modified file.
- New struct and function both carry Doxygen `@brief`; the missing `@param` tags match the pre-existing convention for these file-local test callbacks (`count_data_cb`, `new_data_cb` have the same style).
- Commit subject/body/DCO sign-off follow repository convention.
- Diff is test-only, minimal (`+56/-12`), and touches nothing outside the three cases it targets — no scope creep, no change to `gst/join/gstjoin.c` or `tests/meson.build`.

## Notes 1 and 2 (still open, both non-blocking) — defer still reasonable

Unchanged since `fb591580`/`881c4a05`, and this commit doesn't touch the element code, so nothing here needed re-litigating:

- **Note 1** (`gst_join_sinkpad_unlinked()` also fires during bin teardown): still benign — `GST_STATE_NULL` makes `gst_pad_push_event()` on the src pad a no-op, and all tests go through `GST_STATE_NULL` before unref.
- **Note 2** (`sel->srcpad` read outside `JOIN_LOCK`, no dispose-time clearing): still relies on `GstElement`'s pad-removal order (request pads released before the prepended list is walked) rather than something the code enforces itself. Worth a code comment someday, not a correctness bug today.

## Verdict

**Approving.** This commit does exactly what the last non-blocking note asked for, its race-freedom argument holds (modulo an x86-only caveat that has no practical exposure in this project's CI), and I confirmed empirically — not just by reading — that all three probes (element-fix regression, over-broad-handler rejection, wrong-branch-payload rejection) actually fail the way the commit message claims when mutated. No blocker, major, or new non-blocking issue found. Ready to merge once pushed.

myungjoo:
> _This note relays a correction produced by a separate agent session; please verify before acting on it._

## Correction to one premise in the third review, and the follow-up commit `b6ff7c9a`

The third review's remaining nit — that the race-avoidance in `count_checked_data_cb` relied on x86-TSO ordering — was **right about the code, but the mitigating premise was wrong.**

The review reasoned it was moot because "CI never runs tests on ARM", based on `.github/workflows/gbs_build.yml` setting `unit_test 1` only on the `x86_64` leg. That is true of the GBS workflow, but it is not the only one that runs the suite. In [.github/workflows/ubuntu_clean_meson_build.yml](.github/workflows/ubuntu_clean_meson_build.yml):

```yaml
    strategy:
      matrix:
        os: [ ubuntu-22.04, ubuntu-22.04-arm ]
```

and the step that runs the tests is gated on `env.rebuild == '1'` alone, with no `matrix.os` condition — unlike the neighbouring steps, which do carry `&& matrix.os == 'ubuntu-22.04'`:

```yaml
    - name: build and unit test
      if: env.rebuild == '1'
      ...
        meson test -C build/ -v
```

So `unittest_join` runs on an aarch64 runner on every PR. That is the green `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)` check on this PR. The ordering assumption was therefore being exercised on a weak-memory target, not merely on x86.

### What changed

Rather than add a barrier, `b6ff7c9a` removes the cross-thread mutation the ordering argument existed to justify. `JoinCounter` now holds every expected frame, indexed by the arrival count:

```c
typedef struct {
  const gint *expected[3];
  guint received;
} JoinCounter;
```

All frames are set in the initialiser before `GST_STATE_PLAYING`, so the test thread writes nothing the streaming thread reads, and the "raise the count last" reasoning is no longer load-bearing. `received` stays written only in the callback and read only by `wait_pipeline_process_buffers()`, exactly as `count_data_cb` already does.

Indexing also tightened an assertion: `ASSERT_LT (counter->received, G_N_ELEMENTS (counter->expected))` means a case receiving more buffers than it named frames for fails on the bound, where before the extras were silently compared against the last frame.

### Verification

All three mutants still behave, re-run against this head:

| mutation | expectation | result |
|---|---|---|
| post-unlink buffer pointed at the wrong frame | `noEosOnUnlinkWhileRunning` fails | fails |
| element fix reverted | `eosAfterUnlinkLast` fails | fails (10.3 s timeout) |
| handler forwards EOS on any unlink | both `noEosOnUnlink*` fail | both fail |

`unittest_join` 13/13, full `meson test` 21/21, `gst-indent` and `clang-format` produce no diff.

Notes 1 and 2 from the first review (the handler being reachable during teardown; `sel->srcpad` read outside the JOIN_LOCK) remain deliberately unaddressed as out of scope for #4886 — the element code is untouched since `fb591580`.

myungjoo:
Relayed review — this is a transcription of a review produced by an automated review agent, posted here under the requester's own account.

## Re-review at head `b6ff7c9a` (4th pass; previous rounds: `fb591580` approved, `881c4a05` approved, `c90b2045` approved-with-nit)

**Scope:** `git diff c90b2045..b6ff7c9a` is `tests/gstreamer_join/unittest_join.cc` only, `+13/-12`. `gst/join/gstjoin.c` is byte-for-byte unchanged since `fb591580` except for the one `SECTION` sentence added in `881c4a05` (verified with `git diff fb591580..b6ff7c9a -- gst/join/gstjoin.c`) — no re-analysis of the lock ordering / re-entrancy from round 1 was needed.

### 1. The premise correction is accurate — verified directly, not taken on trust

The 3rd review closed its x86-TSO ordering nit as moot on the claim that "CI never runs tests on ARM," citing `gbs_build.yml`'s `unit_test 0/1` split. That is true for that workflow, but I read [.github/workflows/ubuntu_clean_meson_build.yml](../blob/main/.github/workflows/ubuntu_clean_meson_build.yml) directly and confirmed the correcting comment's claim:

- `strategy.matrix.os: [ ubuntu-22.04, ubuntu-22.04-arm ]`
- The `build and unit test` step (`meson test -C build/ -v`) is gated only by `if: env.rebuild == '1'` — no `matrix.os` condition, unlike sibling steps in the same job that explicitly add `&& matrix.os == 'ubuntu-22.04'` (the llama.cpp/LiteRT/example-app steps).

So `unittest_join` genuinely executes via `meson test` on an aarch64 GitHub-hosted runner on every PR — the "Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)" check. The x86-TSO argument the 3rd review flagged was live, not moot.

### 2. Does `b6ff7c9a` actually eliminate the cross-thread mutation?

Yes. Checked all three call sites, not just read the commit message:

- `eosAfterUnlinkLast`: `JoinCounter counter = { { test_frames[0], NULL, NULL }, 0 };` — set before `gst_parse_launch` and before `setPipelineStateSync(..., PLAYING, ...)`.
- `noEosOnUnlinkWhileRunning`: `{ { test_frames[0], test_frames[1], NULL }, 0 }` — same, both slots populated up front; the old `counter.expected = test_frames[1];` re-point line between the two pushes is gone entirely (not just moved).
- `noEosOnUnlinkWhileOthersRun`: `{ { test_frames[0], test_frames[0], test_frames[0] }, 0 }` — same.

In all three, nothing after `PLAYING` writes into `counter.expected[]`. `counter.received` keeps the same single-writer shape it already had in `count_data_cb`/`count_checked_data_cb`: written only inside the callback (streaming thread), read only by `wait_pipeline_process_buffers` and the post-wait `EXPECT_EQ` (test thread) — the same pattern round 2 already looked at and accepted as a benign, unsynchronized-but-safe-in-practice race. So the specific hazard the 3rd review's nit was about (the store-then-load ordering between "raise count" and "read `expected`") no longer exists, because there's no longer a store into anything the streaming thread reads. Nothing to order, nothing to fence.

### 3. Empirical verification — built and ran this locally (x86_64, WSL, gstreamer 1.20.3), not just read the diff

Baseline at `b6ff7c9a`: 13/13 green, and stable across `--gtest_repeat=10` (130 runs, 0 failures).

**(a) Element fix reverted** (`gst/join/gstjoin.c` restored to pre-`fb591580`, test file untouched at `b6ff7c9a`):
```
../tests/gstreamer_join/unittest_join.cc:773: Failure
  pop_eos_or_error (pipeline, (10000U))
    Which is: 0
  GST_MESSAGE_EOS
    Which is: 1
[  FAILED  ] join.eosAfterUnlinkLast (10322 ms)
```
Fails as expected — real 10s timeout, not a false negative.

**(b) Handler forwards EOS on any unlink regardless of `gst_join_all_sinkpads_eos`/`any_eos`** (patched `send_eos = !sel->eos_sent;` unconditionally):
```
[  FAILED  ] join.noEosOnUnlinkWhileRunning (10124 ms)
[  FAILED  ] join.noEosOnUnlinkWhileOthersRun (10321 ms)
```
Both fail, each catching the spurious early EOS via `pop_eos_or_error` returning `GST_MESSAGE_EOS` where `GST_MESSAGE_UNKNOWN` was expected.

**(c) `noEosOnUnlinkWhileRunning`'s post-unlink expected frame pointed at `test_frames[0]` instead of `test_frames[1]`** (element unchanged, correct):
```
../tests/gstreamer_join/unittest_join.cc:719: Failure
  counter->expected[counter->received][i]
    Which is: 1101
  output[i]
    Which is: 2101
... (48 mismatches total)
[  FAILED  ] join.noEosOnUnlinkWhileRunning (325 ms)
```
Confirms the restored payload check still has teeth after this commit's refactor, not just after `c90b2045`'s.

All three match the commit message's claims exactly. `clang-format` (repo `.clang-format`) reports no diff on the file.

### 4. New finding — minor, not blocking: the `ASSERT_LT` bound doesn't fully deliver on its own stated claim

The commit message says: "a case that receives more buffers than it named frames for now fails on the bound rather than silently comparing the extra ones against the last frame." I checked this against all three tests, not just the one it was written against.

`JoinCounter.expected` is a fixed `[3]` array shared by all three tests, but only `noEosOnUnlinkWhileOthersRun` populates all three slots. `eosAfterUnlinkLast` only names 1 frame (slots 1–2 are `NULL`); `noEosOnUnlinkWhileRunning` only names 2 (slot 2 is `NULL`). `ASSERT_LT (counter->received, G_N_ELEMENTS (counter->expected))` checks against the array's *capacity* (always 3), not against how many frames a given test actually "named." So an extra buffer arriving within capacity but past what a test populated does not hit the bound at all — it indexes a `NULL` slot, and the following loop dereferences it: `EXPECT_EQ (counter->expected[counter->received][i], output[i])` reads `NULL[i]`.

Verified empirically, not just by inspection: I added one extra legitimate buffer push to `noEosOnUnlinkWhileRunning` right after its second expected push (simulating a hypothetical future bug where the still-linked, still-running branch forwards one spurious extra buffer — exactly the failure mode these `noEosOnUnlink*` tests exist to catch), rebuilt, and reran:
```
[ RUN      ] join.noEosOnUnlinkWhileRunning
Segmentation fault (core dumped)
exit code: 139
```
No GTest failure message at all — a raw SIGSEGV. This is a real behavior change from the pre-`b6ff7c9a` design: at `c90b2045`, `JoinCounter.expected` was a single pointer re-pointed between pushes, so an unexpected extra buffer would have compared against whatever the pointer currently held (always non-`NULL`) — silently wrong or a clean `EXPECT_EQ` mismatch, never a crash. `b6ff7c9a` trades that silence for a hard bound, but the bound is sized for the busiest test (`noEosOnUnlinkWhileOthersRun`'s 3 slots), not per-test, so it only actually protects that one case; the other two get a crash instead of either the old silent behavior or the new commit's claimed graceful bound failure.

Practical impact is limited: it cannot cause a false failure on correct code (confirmed stable over 10 repeats), and CI would still fail the job on a real crash (non-zero exit, core dumped and uploaded by the `Upload core if something went wrong` step), so this doesn't reduce CI's blocking power. It does mean that if a future regression ever caused one extra spurious buffer specifically on `eosAfterUnlinkLast` or `noEosOnUnlinkWhileRunning`'s surviving branch, the diagnosis would be a core dump instead of the clean "expected N, got M" message the bound was written to provide — a downgrade in diagnostic quality for exactly the two tests meant to be the belt-and-suspenders check on this element's contract. Non-blocking suggestions: populate all three slots in every test (duplicate the last real frame into the trailing slots, the way `noEosOnUnlinkWhileOthersRun` already does uniformly), or add an explicit `ASSERT_NE (counter->expected[counter->received], nullptr)` next to the existing bound check.

### 5. Notes 1 and 2 (still open) — defer still reasonable

Both are about `gst/join/gstjoin.c`, which is unchanged since `fb591580` (confirmed above). `b6ff7c9a` is test-only, so nothing here needed re-litigating; the earlier rounds' judgment to defer these as non-blocking, out-of-scope-for-#4886 follow-ups still holds.

### 6. Everything else

- **Issue resolution**: `eosAfterUnlinkLast` is still the exact #4886 repro, still fails without the element fix, still passes with it (verified in 3(a) above).
- **Regression risk to other modules**: unchanged from round 1's analysis, since `gst/join/gstjoin.c` itself hasn't moved since `fb591580`.
- **Code size**: `+13/-12`, test-file-only, proportionate to the follow-up it's making.
- **CI blocking power**: confirmed directly — `meson test -C build/ -v` (which includes `unittest_join`) runs unconditionally on both the `ubuntu-22.04` and `ubuntu-22.04-arm` legs of `ubuntu_clean_meson_build.yml`, so a regression here is caught on both architectures, not just x86_64.
- **Style/Doxygen/DCO**: `clang-format` clean; the updated `JoinCounter` struct and `count_checked_data_cb` doc comments follow the file's existing convention; `b6ff7c9a` carries `Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>`.
- All checks were in `pending` state at the time of this review (fresh push), including both Valgrind jobs; nothing in my local reproduction suggests they won't pass.

## Verdict

**Approving.** `b6ff7c9a` genuinely resolves the ordering nit from round 3 — not by adding a barrier, but by removing the cross-thread write the barrier would have been protecting, which I confirmed by reading all three call sites rather than trusting the commit message. All three requested mutation probes reproduce exactly as claimed. The one new thing I found — the `ASSERT_LT` bound silently not covering two of the three tests it was written for, reproduced as an actual `SIGSEGV` — is a real gap, but it's a test-only robustness nit that cannot cause a false failure and does not weaken CI's ability to block a real regression, so it's non-blocking. Notes 1 and 2 remain reasonable to defer, unchanged. No blocker, major, or CI-relevant issue found. Mergeable once the currently-pending checks (both Valgrind legs in particular, given the ARM finding above) land green.

myungjoo:
> Relayed review — this is a transcription of a review produced by an automated review agent, posted here by the requester's own account. Please verify before acting on it.

## Re-review at head `f96c1555` (5th pass; previous rounds: `fb591580` approved, `881c4a05` approved, `c90b2045` approved, `b6ff7c9a` approved with one new minor finding)

**Scope:** `git diff b6ff7c9a..f96c1555` is `tests/gstreamer_join/unittest_join.cc` only, `+8/-3`. `gst/join/gstjoin.c` remains byte-for-byte unchanged since `fb591580` except the one `SECTION` sentence added in `881c4a05` (`git diff fb591580..f96c1555 -- gst/join/gstjoin.c` confirms this: `+2/-0`, doc-only). No re-analysis of lock ordering / re-entrancy was needed.

This commit is the direct follow-up to the 4th review's one new finding: the `ASSERT_LT (counter->received, G_N_ELEMENTS (counter->expected))` bound only actually protects `noEosOnUnlinkWhileOthersRun` (which fills all 3 slots); `eosAfterUnlinkLast` (1 slot) and `noEosOnUnlinkWhileRunning` (2 slots) leave the remaining slots `NULL`, so a buffer arriving past what a test named passed the bound and dereferenced a `NULL` `expected[i]`, crashing with SIGSEGV instead of failing gracefully. The commit message for `b6ff7c9a` calling this "fails on the bound" was an overstatement for those two cases, and this commit's message explicitly corrects that.

### 1. Does `f96c1555` actually fix the SIGSEGV — verified empirically, not by reading

I reproduced both sides of this independently in a local WSL x86_64 build (gstreamer 1.20.3), using the exact repro the 4th review used: inject one extra legitimate buffer push into `noEosOnUnlinkWhileRunning` right after its second expected buffer is received (simulating a still-linked/still-running branch forwarding one spurious extra buffer — the failure mode this pair of tests exists to catch).

**Before the fix** (test file reverted to the `b6ff7c9a` shape — `count_checked_data_cb` indexing `counter->expected[counter->received][i]` directly, no local `expected`/`ASSERT_NE`), rebuilt and ran:
```
[ RUN      ] join.noEosOnUnlinkWhileRunning
Segmentation fault (core dumped)
exit code: 139
```
Confirmed core dump, matching the 4th review's finding exactly.

**After the fix** (actual `f96c1555` test file, same injected extra buffer), rebuilt and ran:
```
../tests/gstreamer_join/unittest_join.cc:716: Failure
Expected: (expected) != (nullptr), actual: NULL vs (nullptr)
...
[  FAILED  ] join.noEosOnUnlinkWhileRunning (10431 ms)
```
No crash — a named GTest failure identifying the exact assertion and line. `exit code: 1`, not 139.

This is a genuine, verified fix, not a paper one.

### 2. Do `ASSERT_LT` + `ASSERT_NE` together cover all over-count paths?

Yes, checked both halves separately:
- **`received >= 3` (capacity exceeded, all 3 tests):** still caught by the pre-existing `ASSERT_LT`, unaffected by this commit.
- **`received` within capacity but the slot is unpopulated for that test (`eosAfterUnlinkLast`, `noEosOnUnlinkWhileRunning`):** now caught by the new `ASSERT_NE (expected, nullptr)`, confirmed above by the injected-extra-buffer repro on `noEosOnUnlinkWhileRunning`.

Between the two guards, every over-count path either hits the bound or hits the null-slot check before the `for` loop dereferences anything.

### 3. `ASSERT_*` early-return leaves `received` un-incremented — is that a problem?

Checked this directly, since it's the one part of the mechanism worth being careful about. `ASSERT_NE` (and `ASSERT_LT` before it) return from `count_checked_data_cb` on failure, which is a `void`-returning callback invoked on the GStreamer streaming thread — so `counter->received` is never incremented for that spurious buffer. In my repro this meant the test thread's `wait_pipeline_process_buffers (&counter.received, 3, TEST_TIMEOUT_LIMIT_MS)` could never observe count 3 and consequently timed out after the full 10 s, producing a *second*, secondary failure (`Value of: wait_pipeline_process_buffers(...) Actual: false`) on top of the primary `ASSERT_NE` failure.

This is not a new problem introduced by this commit — the pre-existing `ASSERT_LT` already had the same shape (an early return that leaves the waiter to time out) for `noEosOnUnlinkWhileOthersRun`, so `f96c1555` doesn't change this failure mode, only extends it to the other two tests. It also cannot produce a **false failure** on correct code: this path is only reachable when a real assertion has already fired, i.e. when the element under test is actually misbehaving. On a passing run nothing times out. The only cost is diagnostic/latency: a real regression now takes the full ~10 s timeout to fully unwind (with a clear message) instead of failing instantly, which is a reasonable trade against the alternative (SIGSEGV) it replaces. Non-blocking.

### 4. Existing verification power — reconfirmed at this head, not assumed carried over

Re-ran all three mutation probes from earlier rounds directly against `f96c1555`, since a test-only diff touching a shared struct is exactly the kind of change that could silently weaken assertions elsewhere in the file:

- **(a) Element fix reverted** (`gstjoin.c` restored to pre-`fb591580`, test file at `f96c1555`): `join.eosAfterUnlinkLast` fails — `pop_eos_or_error(pipeline, 10000U)` returns `0` (`GST_MESSAGE_UNKNOWN`), never `GST_MESSAGE_EOS`. Real 10s timeout, genuine regression test.
- **(b) Handler forwards EOS on any unlink** (`send_eos = !sel->eos_sent` unconditionally, dropping the `all-sinkpads-eos && any_eos` gate): both `join.noEosOnUnlinkWhileRunning` and `join.noEosOnUnlinkWhileOthersRun` fail, each via `pop_eos_or_error(pipeline, 300)` returning `GST_MESSAGE_EOS` where `GST_MESSAGE_UNKNOWN` was expected.
- **(c) `noEosOnUnlinkWhileRunning`'s post-unlink expected frame pointed at `test_frames[0]` instead of `test_frames[1]`** (element unchanged, correct): fails with 48 `EXPECT_EQ` mismatches (`1101` vs `2101`, ..., `1224` vs `2224`), confirming the payload check still has teeth after this refactor.

All three behave exactly as claimed by the commit history. Full suite at `f96c1555` (clean, no injected mutants): **13/13 pass**, `4.7s` total.

### 5. Style / hygiene

- `clang-format` (repo `.clang-format`, v16.0.6) reports no diff on the modified file.
- New local `const gint *expected` and the `ASSERT_NE` line read naturally inside the existing Doxygen-commented function; the function's doc comment (`@note`) was updated to describe the new behavior accurately.
- Diff is `+8/-3`, test-only, scoped to exactly the function the finding was about — no scope creep.
- Commit message is precise about what it corrects (the `b6ff7c9a` message's overstatement) rather than silently fixing it, which is good practice for reviewability.

### 6. Notes 1 and 2 (still open, both non-blocking) — defer still reasonable

Unchanged since round 1; `gst/join/gstjoin.c` is untouched by this commit (confirmed above), so nothing here needed re-litigating:

- **Note 1** (`gst_join_sinkpad_unlinked()` also fires during bin teardown): still benign — `GST_STATE_NULL` makes `gst_pad_push_event()` on the src pad a no-op, and every test goes through `GST_STATE_NULL` before unref.
- **Note 2** (`sel->srcpad` read outside `JOIN_LOCK`, relying on `GstElement`'s pad-removal ordering rather than the code enforcing it itself): still an implicit dependency on GStreamer-core teardown order, not a currently-reachable bug. Reasonable to defer as a follow-up rather than bundle into this fix.

### 7. Everything else

- **Issue resolution (#4886)**: `eosAfterUnlinkLast` remains the exact repro from the issue, still fails without the element fix, still passes with it (verified in 4(a) above). The element code itself hasn't moved since the first approved round.
- **Regression risk to other modules**: none — this commit is test-file-only, and `gst/join/gstjoin.c` (the only thing other consumers link against) is unchanged since `fb591580`.
- **Code size**: `+8/-3`, proportionate to the finding it addresses.
- **CI**: at review time, `Static checks`, `DCO`, `CPP-Linter`, and `Build on MacOS-latest with Meson` are green on this head; the jobs that actually execute `unittest_join` (`Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` and `(ubuntu-22.04-arm)`, both Tizen/GBS legs) are still queued/pending as of this review — a fresh-push timing gap, not a code concern. My local build/run above gives high confidence they will pass.
- **Test completeness**: no new test case is needed for this commit — it hardens an assertion inside an existing test's shared helper rather than adding new behavior to pin.

## Verdict on this commit

**No blocker, major, or new non-blocking issue found.** `f96c1555` does exactly what its commit message claims: it turns a real SIGSEGV (verified reproduced against the pre-fix shape) into a named, graceful GTest failure (verified against the post-fix shape), without weakening any of the three mutation probes that give this test file its regression-catching power, and without touching the element code at all. The one behavior worth flagging (early `ASSERT_*` return delays the waiter to a full timeout before the failure is reported) is a pre-existing pattern this commit extends rather than introduces, cannot cause a false failure, and is non-blocking.

## Overall PR merge-readiness (5 commits: `fb591580` element fix + initial tests, then `881c4a05` → `c90b2045` → `b6ff7c9a` → `f96c1555` as four successive test-hardening follow-ups from review)

**Mergeable, code-quality-wise.** Summary of the full arc:
- The element fix (`fb591580`) was independently verified in round 1 against actual GStreamer 1.20/1.24 source for the lock-ordering claim, and has not changed since.
- Every review round's finding was addressed in the very next commit, each verified rather than taken on faith, and this round closes out the last open one (the `ASSERT_LT`-only-partial-coverage gap from round 4).
- All three mutation probes that matter for this element's contract (fix reverted → hang; over-broad EOS-on-unlink → false EOS; wrong post-unlink payload → mismatch) still fail correctly at this head, reconfirmed directly rather than assumed carried over.
- Notes 1 and 2 remain legitimately out-of-scope non-blocking follow-ups on code that hasn't changed since the first approved round.

On the **commit composition**: five commits — one functional fix with its initial regression tests, then four incremental test-hardening commits, each a direct, well-labeled response to a specific prior-review finding — is a reasonable and, if anything, unusually well-documented history for this kind of review-driven hardening. Each commit is small, individually buildable, individually correct, and its message explains exactly what review comment it answers (including one commit candidly correcting an overstatement in an earlier commit's own message, which is a good sign rather than a concern). I would not ask for a squash before merge; the incremental history is more useful for future readers than a single flattened commit would be, and nothing here is the kind of back-and-forth (e.g., revert-then-redo of the same line) that squashing exists to clean up.

The only remaining gate is process, not code: the Valgrind jobs (both x86_64 and the aarch64 leg specifically relevant to round 4's ordering finding) were still queued at review time. Recommend waiting for those two plus the Tizen/GBS legs to go green before merge, but I found no code reason to hold this PR further.

PR #4899, opened 2026-08-28
[Filter/NCNN] Make input/output dimension and type properties optional

### LINKED ISSUE #4642: input/output dimension should NOT be mandatory for tensor-filter
input/output property of tensor-filter should not be mandatory as long as the subplugin code can fetch input/output informatino from the given framework and model file.

In general, they should be used for debugging / validation purpose (generate explicit error is there is an inconsistency).

For starters, NCNN requires input/output dimensions and types, and this should never be.

myungjoo:
Resolves #4642 for the ncnn subplugin.

## Problem

`tensor_filter_ncnn` copied `input_meta` / `output_meta` out of the properties and threw when either was empty, so `input`, `inputtype`, `output` and `outputtype` were all mandatory — the exact case the issue calls out. The tensor-filter core already implements the fallback the issue asks for (`GET_IN_OUT_INFO` → input from pad caps → `SET_INPUT_INFO`); ncnn just never participated, and threw in `configure_instance()` before the core could ask.

## What the model can and cannot tell us

Measured against `tests/test_models/models/ncnn_models/squeezenet_v1.1`:

```
in[0]  "data"  blob shape hint: dims=0   (absent)
out[0] "prob"  blob shape hint: dims=0   (absent)
dry-run out[0]: dims=1 w=1000            -> nnstreamer 1000:1
```

`Blob::shape` is filled only from param key `30` (`-23330=`), which this model does not carry, and the `Input` layer's `0=227 1=227 2=3` is unreachable because ncnn does not install `layer/input.h`. So the input comes from the pad caps and the output from one inference with a zeroed input. The inferred `1000:1` is exactly what the existing test declares by hand, so nothing downstream shifts.

`use_yolo_decoder` still requires `output`/`outputtype` — the decoded bounding-box tensor is not a shape the model produces. The error now says that instead of reporting a matrix count mismatch.

`GET_IN_OUT_INFO` answers only when the input came from the property. Anything resolved via `SET_INPUT_INFO` is a negotiation candidate GStreamer may still revise, and latching it would freeze a probe dimension into `prop->input_configured`.

## Memory defects fixed along the way

The size arithmetic had to be reworked for the inferred shapes, and these sit on exactly those lines:

1. **Heap over-read + channel misalignment.** ncnn aligns `Mat::cstep` to 16 bytes, so `mat.total()` counts padding the nnstreamer tensor does not have. `(elembits()/8) * total()` copied 618384 bytes from a 618348-byte buffer for the 227:227:3 input, and shifted channels 1 and 2 by 3 floats each. ASan on the upstream expression:

   ```
   ==2114==ERROR: AddressSanitizer: heap-buffer-overflow
   READ of size 618384 at 0x736cfc15676c thread T0
   SUMMARY: heap-buffer-overflow in __interceptor_memcpy
   ```

   Plane-wise copies are clean, and the reference/flat comparison shows the misalignment changed the argmax (533 vs 999) on a synthetic input.

2. **Uninitialized first output frame.** `output_mats` was sized from an empty `Mat`, so the first invoke copied 0 bytes. Observable today — 10 runs of the existing pipeline on unmodified `main`:

   ```
   orig1: n04141327 scabbard      n07747607 orange
   orig2: n01440764 tench, ...    n07747607 orange
   ...
   ```

   Frame 1 is uninitialized memory and varies run to run. The existing test only passed because it feeds two buffers and greps the whole log.

3. **No type validation**, so `inputtype=uint8` made the copy read 4× the buffer length.

Also fixed while rewriting `invoke()`: it walked a single `input->data`/`output->data` buffer rather than indexing `input[i]`/`output[i]`, and the yolo decode loop wrote rows and label slots unbounded. The per-tensor threads are gone — concurrent `extract()` calls share the `Extractor`'s `blob_mats` and race for multi-output models, and the copies they hid are no longer on that path.

## Tests

`tests/nnstreamer_filter_ncnn/runTest.sh`:

| case | covers |
|---|---|
| 9 | no dimension/type property at all |
| 10 | `input`/`inputtype` only, output inferred |
| 11 | `output`/`outputtype` only, input from caps |
| 12 | trailing-1 padded dimension `227:227:3:1:1` |
| 13 | single buffer — the first-frame regression |
| 14_n | `inputtype=uint8` rejected |
| 15_n | `use_yolo_decoder` without an output shape rejected |

Case 1 now requires **both** frames to be labeled. Case 2 asserted that `227:227:3:1:1` is over-ranked; it is not, since nnstreamer pads dimensions with trailing 1s and treats them as equal, so it now uses a dimension with a real fifth axis.

Every new case fails against the current code and passes with the change:

```
                       unfixed        fixed
 1:Golden               FAILED       PASSED
 9,10,11,12,13          FAILED       PASSED
 14_n                   FAILED       PASSED
 total              10/21 pass    21/21 pass
```

## CI

ncnn was in none of the workflow dependencies, `debian/control` or the Tizen spec, so the subplugin has never been compiled by CI and its SSAT script has reported success with zero cases since Dec 2023. The new tests would have been dead weight.

`ubuntu_clean_meson_build.yml` now installs the official prebuilt shared release (7.6 MB, sha256-pinned; headers + `libncnn.so` + relocatable `ncnn.pc`), mirroring how the llamacpp filter takes a pinned llama.cpp build — there is no distro package and building ncnn from source would cost minutes per run. The `.so` existence is asserted because the meson feature is `auto` and would otherwise degrade to off silently.

Rehearsed locally against that exact zip in a clean build dir: 21/21. The Vulkan case (8) passes on a host without a Vulkan ICD too — `vkCreateInstance failed -9`, ncnn falls back to CPU, pipeline completes.

## Scope

`ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc` is a leaf: nothing else in the tree includes or links it (`meson.build` builds it into its own shared/static library; it is loaded by name at runtime). No core file is touched — the fallback machinery it now uses is the documented path other subplugins already take.


myungjoo:
**Disclosure:** This review was produced by an automated review agent (not a human maintainer) and is being posted verbatim on behalf of the requester. Please weigh it as one input among the human reviews.

## Scope of review
Analyzed the diff (`.github/workflows/ubuntu_clean_meson_build.yml`, `ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc`, `tests/nnstreamer_filter_ncnn/runTest.sh`) against:
- issue #4642 ("input/output dimension should NOT be mandatory for tensor-filter"),
- the core `GET_IN_OUT_INFO`/`SET_INPUT_INFO` contract in `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h` and its call sites in `gst/nnstreamer/tensor_filter/tensor_filter.c` / `tensor_filter_common.c`,
- the ncnn upstream source (`Mat`/`Extractor`/packing-layout behavior) to verify the memory-safety claims,
- branch protection / required status checks for `nnstreamer/nnstreamer:main`.

## Does it resolve the stated issue?
Yes, for the ncnn subplugin specifically (the PR only claims "for the ncnn subplugin", matching the issue's "For starters, NCNN requires..." framing). Verified end-to-end:
- `configure_instance()` no longer throws when `input`/`output` properties are absent; it only validates/normalizes what is given (`normalizeInfo`).
- `getModelInfo(GET_IN_OUT_INFO, ...)` correctly returns `-ENOENT` unless the input came from the property, forcing the core to negotiate via pad caps (`SET_INPUT_INFO`) — this matches the exact fallback path the core has always supported (confirmed in `nnstreamer_plugin_api_filter.h:196-201,426-449` and `tensor_filter.c`/`tensor_filter_common.c`), which no other subplugin currently exercises this way (`mxnet` falls through to `-ENOENT` for `SET_INPUT_INFO`; `python3` only implements one op or the other depending on script callback; `litert` implements only `GET_IN_OUT_INFO`). This is a legitimately new (not copy-pasted) but contract-correct usage of the existing core mechanism.
- Output-shape inference via a single dummy forward pass (`inferOutputInfo()`) is a reasonable approach given ncnn's `.param` format frequently omits blob shape hints (confirmed against the `squeezenet_v1.1.param` test model).
- `use_yolo_decoder` correctly keeps `output`/`outputtype` mandatory, since the decoded box tensor shape isn't something the model itself produces — verified the throw path and test 15_n.

None found. `tensor_filter_ncnn.cc` is a true leaf: `grep`-ing the whole tree for `tensor_filter_ncnn`/`ncnn_subplugin` only turns up the file itself and `ext/nnstreamer/tensor_filter/meson.build` (built as its own shared/static lib, loaded by name at runtime). No core file is touched. `ncnn-support` is an `auto` meson feature, consistent with the "leaf module" claim.

The memory-safety fixes described in the PR body were independently verified against ncnn's actual source (cloned upstream, commit `0a2c46bc`):
- `Mat::create(w,...)` aligns `cstep` to 16 bytes and `mat.total()` includes that padding, so the old `(elembits()/8) * total()` flat-copy read/wrote past the plain tensor buffer and shifted channel planes — confirmed by reading `mat.cpp`'s `create()` implementations. The new `copyToMat`/`copyFromMat` per-channel-plane copy (`mat.channel(i).data`, `plane = w*h*d*elemsize`) is correct.
- `Extractor::extract(..., type=0)` (the default, used here) unpacks any SIMD-packed (`elempack>1`) intermediate blob back to `elempack==1` before returning to the caller (verified in `net.cpp` around the `use_packing_layout && feat.elempack != 1` branch), so the shape-inference logic (`out.w/out.h/out.d/out.c`) and the plane-size arithmetic operate on genuinely unpacked, elemsize=4 data — the packing-layout default (`true`) does not silently corrupt the inferred output shape.
- Threads were removed; `invoke()` is now strictly sequential over `input[i]`/`output[i]`, matching the fix described for the previous single-buffer / thread-shared-blob_mats race.

## Scope / code size
Well-scoped: 3 files, and the two non-core files (workflow, test script) are additive/test-only. The `tensor_filter_ncnn.cc` rewrite is larger than a minimal "make properties optional" patch would be, but the extra lines are the necessary consequence of doing so safely — the shape-inference path, the type/rank validation now needed because inputs are no longer guaranteed present, and the safety fixes on the exact lines that had to be reworked for the new (possibly model-inferred) shapes. This isn't scope creep into unrelated modules.

## Test coverage
Good, and it's the strongest part of this PR:
- New SSAT cases 9–13 cover: no properties at all, input-only, output-only, a trailing-1-padded dimension, and — importantly — a single-buffer run that pins down the "first invoke leaves output uninitialized" bug as a real regression test (case 1 was also strengthened from "log contains orange somewhere" to "both frames say orange exactly", which is the change that actually catches that bug).
- Negative cases 14_n (rejected non-float32 type) and 15_n (`use_yolo_decoder` without explicit output) match the two new hard-requirements introduced by this feature.
- Case 2 was retargeted from `227:227:3:1:1` to `227:227:3:2:2` for the "over-ranked dimension must fail" check — this is a *correct* fix, not a weakening: under the new `getShape()` trailing-1-trim logic, `227:227:3:1:1` no longer is an over-rank case (verified: `gst_tensor_parse_dimension` 0-fills unspecified trailing entries, and explicit trailing 1's are trimmed the same way), so the old value would have stopped testing what it claimed to test. Good catch by the PR author.
- Per the PR body's own before/after table, every new/changed case fails on unfixed `main` and passes with the fix — i.e. these are genuine regression tests, not just happy-path additions, satisfying the repo's test-case policy for bug fixes.

## CI enforcement (will a future regression actually be caught and block merge?)
Verified concretely, not just assumed:
- `ncnn-support` is `auto`, so before this PR ncnn was silently never built or tested by CI (confirmed: not referenced anywhere in workflows, `debian/control`, or the Tizen spec). The PR adds a pinned, sha256-checked prebuilt ncnn install step to `ubuntu_clean_meson_build.yml`, a `test -f .../libnnstreamer_filter_ncnn.so` assertion (so silent feature-detection regressions fail loudly, mirroring the existing `litert`/`llamacpp` pattern), and a dedicated "SSAT run on filter-ncnn" step that runs unconditionally (no `|| echo ...` swallow, matching the `litert` step, unlike the intentionally-non-blocking Valgrind step).
- Checked branch protection via `gh api repos/nnstreamer/nnstreamer/branches/main/protection`: the job this workflow defines, **"Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)"**, is a **required status check** on `main`. So a future change elsewhere (core tensor_filter, ncnn version bump, etc.) that breaks these new SSAT cases will fail a required check and block merge, not just show a warning. This satisfies the "will CI actually stop a bad merge" requirement.
- Coverage gap (informational, not a defect introduced by this PR): the ncnn install/test steps are gated to `matrix.os == 'ubuntu-22.04'` only, so `ubuntu-22.04-arm` still never builds/tests the ncnn subplugin. This mirrors the pre-existing `litert`/`llamacpp` pattern and is explained by upstream ncnn only publishing an x86_64 prebuilt release asset — reasonable, not a fault of this PR.

## Documentation / architecture / API changes
No core API or architecture was changed (confirmed no core file in the diff, and the `GET_IN_OUT_INFO`/`SET_INPUT_INFO` enum/contract already existed before this PR). The only behavior-facing documentation is the subplugin's own file-header Doxygen comment (`tensor_filter_ncnn.cc:18-28`), which was updated in the same PR to describe the new optional-property behavior. No other repo documentation references ncnn's property requirements (checked `Documentation/`), so nothing else needs updating.

## Issues found

1. **(minor)** `getShape()`'s doc comment says "nnstreamer pads a dimension with trailing 1s up to its rank limit." That's only true when a user *explicitly* supplies trailing 1's (e.g. `input=227:227:3:1:1`); the actual default fill from `gst_tensor_info_init`/`gst_tensor_parse_dimension` (`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:121-133,1122-1156`) is 0, not 1, for unspecified trailing ranks. The code is correct either way (the loop in `getShape()` stops at the first 0, and the subsequent `while` trims trailing 1's), so this is a wording nit only, not a functional bug.

2. **(minor)** In `setInputInfo()` (`tensor_filter_ncnn.cc:281-308`), `inputInfo`/`input_mats` are overwritten with the new candidate shape *before* `inferOutputInfo()` runs. If that inference then throws (e.g. a genuinely bad negotiated shape), `getModelInfo` returns `-EINVAL` and the core correctly hard-fails negotiation (`GST_FLOW_NOT_NEGOTIATED`, verified in `tensor_filter.c`) rather than silently invoking with mismatched state — so this is not exploitable today. But the subplugin is left holding a new `inputInfo` paired with a stale `outputInfo` until/unless a later `SET_INPUT_INFO` call with a good shape self-heals it (via the `gst_tensors_info_is_equal` check). Worth a short comment or a rollback-on-failure for robustness, but low severity given the core's existing guard rails.

3. **(informational, not a request for changes)** `getModelInfo(SET_INPUT_INFO)` can trigger a real forward pass through the model (`inferOutputInfo()`) whenever the candidate input shape changes and no explicit `output` was given. GStreamer's caps negotiation can call `SET_INPUT_INFO` more than once with different candidates before settling (per `tensor_filter.c`'s `transform_caps`/`configure_tensor` split). This is mitigated by both the core (`output_configured` gate before re-invoking) and by ncnn's own "skip if unchanged" check, so in practice this should run at most a small handful of times per negotiation — but for large/expensive models the first negotiation could have noticeably higher latency than a filter with statically-known shapes. No action requested, just flagging for awareness.

No critical or major issues found.

## Verdict
**Approve / OK to merge.** The change correctly and minimally resolves issue #4642 for the ncnn subplugin, is provably isolated from the rest of the tree (leaf module, no core touched), fixes and regression-tests three real memory-safety bugs uncovered while doing so, and — notably — brings the ncnn subplugin under a *required* CI status check for the first time, so it will no longer silently bit-rot. The three items above are minor/informational and do not need to block merging.

myungjoo:
Thanks — both minor points are addressed in 8721857, and the informational one is noted below.

**Minor 1 — `getShape()` Doxygen wording.** Correct, unspecified trailing dims are 0, not 1. Reworded: the dimension ends at the first 0, and the trailing 1s the function also trims are what pad-caps carry (`dimension=227:227:3:1` for this pipeline) and what `gst_tensor_dimension_is_equal` treats as equal to no axis at all. Both are what the loop handles; the text now says that rather than claiming nnstreamer pads with 1s.

**Minor 2 — half-applied candidate in `setInputInfo()`.** Fixed properly rather than by unwinding. The matrices and the inferred output are now built on the side and all three members are swapped in only once nothing can throw:

```c
    mats = allocMats (given);
    if (!output_from_prop)
      inferOutputInfo (mats, inferred);
  } catch (...) { /* nothing was mutated */ }

  gst_tensors_info_free (&inputInfo);
  inputInfo = given;
  input_mats = std::move (mats);
```

`inferOutputInfo()` takes the matrices and the target metadata as arguments for that, which also removes its dependency on member state, and `allocInputMats()` became the static `allocMats()` over a given tensors info.

**Informational — repeated inference during negotiation.** Agreed it is bounded rather than free. Two things cap it: the core only calls `SET_INPUT_INFO` while `output_configured` is false, and the subplugin returns early when the candidate equals the info it already holds, so a settled negotiation costs one forward pass. That early-return is also why the swap above keeps the "unchanged" check ahead of `allocMats()`. It is one zeroed forward pass at caps time, never per buffer, so I have left it as is — happy to gate it behind a custom property if you would rather it not run at all for large models.

Re-verified after the change against the prebuilt ncnn release the CI step installs, in a clean build directory: 21/21, twice. clang-format 16 reports no replacements.

Also confirming your check on the required status: the `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` job is where the ncnn install and the new SSAT step live, so the added cases do gate merges rather than only warn.

myungjoo:
**Disclosure:** This is a second-round review, produced by an automated review agent (not a human maintainer) and posted verbatim on behalf of the requester. Please weigh it as one input among the human reviews. It follows up on the review already posted on this PR and focuses on the new commit `8721857a` ("[Filter/NCNN] Apply the input info candidate atomically"), while re-checking the original review criteria against the PR as a whole.

## Scope
- Read `8721857a` in full (`git show 8721857a`) against the pre-image, and the resulting `ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc` in full.
- Cross-checked `GstTensorsInfo` ownership semantics in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c` (`gst_tensors_info_init/free/copy`, `gst_tensor_info_free`, `gst_tensors_info_get_nth_info`) to trace exactly what is heap-allocated (`GstTensorInfo::name`, `GstTensorsInfo::extra`) and who owns it at each step.
- Independently built the branch: cloned `claude/github-issue-4642-150b02` in a clean environment, installed the exact pinned prebuilt ncnn release the CI step uses (`ncnn-20260526-ubuntu-2204-shared`), configured meson with `-Dncnn-support=enabled`, and did a full `ninja` build — compiles clean, no warnings.
- Ran the case-9 pipeline (no `input`/`output` properties at all — the primary path through the new atomic `setInputInfo()`) and the case-15_n pipeline (`use_yolo_decoder` without an explicit output, i.e. the `inferOutputInfo()` throw path) directly with `gst-launch-1.0` against the real squeezenet test model, outside of SSAT/ssat-tool (not installed in this environment). Both produced the expected result: case 9 labels both frames "orange" (2/2), case 15_n fails cleanly with the expected message, hit twice during negotiation retries without any crash or corrupted state on a following case-9 run in the same environment.
- `clang-format --dry-run --Werror` on the changed file: no replacements needed, confirming the author's own report.
- Confirmed via `git diff origin/main...HEAD --stat` that no file under `gst/` (core) is touched by any commit in this PR, so the regression-risk conclusion from the first review still holds unchanged.

## 1. Does `8721857a` fix the two points raised in the previous review?

**Half-applied `setInputInfo()` candidate — yes, fixed correctly.**
Traced the full ownership chain:
- `given` and `inferred` are local `GstTensorsInfo`, `gst_tensors_info_init`'d (so freeing them when nothing was ever copied into them is a safe no-op — `name == NULL`, `extra == NULL`).
- Every throwing step (`normalizeInfo`, `allocMats`, `inferOutputInfo`) happens before any member (`inputInfo`, `outputInfo`, `input_mats`) is touched. The single `catch (...)` frees exactly `given` and `inferred` and rethrows — `mats` (a `std::vector<ncnn::Mat>`) is unwound automatically by its own destructor. No double free, no leak on the exception path.
- On success, `inputInfo` is explicitly freed before `inputInfo = given;` (a raw struct assignment that is safe here specifically because `GstTensorsInfo` is a plain C struct with no destructor — the local `given` simply goes out of scope afterward without ever being freed again, so ownership of `given.info[i].name` / `given.extra` transfers to `inputInfo` by value with no double free). Same pattern for `outputInfo = inferred;`, guarded by `!output_from_prop` so it only fires when `inferred` was actually populated.
- The early-return "unchanged candidate" path frees `given` and returns before `inferred`/`mats` are touched — correct, since neither was allocated yet at that point.
- `inferOutputInfo()`'s per-tensor loop sets `info.num_tensors` before filling entries and can throw partway through; `gst_tensors_info_free()` iterates up to `num_tensors` unconditionally, but every unfilled `GstTensorInfo` in `info`/`info.extra` is still `name == NULL` from `gst_tensors_info_init`, so `g_free(NULL)` on those is a safe no-op. Confirmed empirically: case 15_n's throw message appears twice in one process run (retried by negotiation) with no crash and no observable state corruption on a subsequent run.

This closes the concern raised in the previous review's minor #2 exactly as described in the commit message and the author's own follow-up comment.

**`getShape()` Doxygen wording — yes, fixed correctly.** The corrected text ("ends either at the first 0 or at explicit trailing 1s ... pad-caps carry") matches what `gst_tensor_parse_dimension`/`gst_tensor_info_init` actually do (0-fill unspecified trailing ranks) and what the function's own code does (stops at the first 0, then trims trailing 1s). This is a documentation-only change; the code was already correct before this commit, only the comment was wrong.

## 2. Does `8721857a` introduce any new defect?

**No leak/double-free found.** See the ownership trace above — every allocation reachable from `given`, `inferred`, and `mats` is freed exactly once on every path (both success and every throw point), and the struct-assignment ownership transfer (`inputInfo = given;` / `outputInfo = inferred;`) is safe because neither source local is touched again afterward.

**`allocMats()`'s `const_cast` — safe, and not a new pattern.** (minor/nit, informational)
```cpp
std::vector<ncnn::Mat>
ncnn_subplugin::allocMats (const GstTensorsInfo &info)
{
  ...
  mats.push_back (allocMat (
      gst_tensors_info_get_nth_info ((GstTensorsInfo *) std::addressof (info), i)));
  ...
}
```
`gst_tensors_info_get_nth_info()` lazily allocates `info->extra` the first time an index `>= NNS_TENSOR_MEMORY_MAX` is requested, so it is technically mutating despite `allocMats` promising `const`. However:
- This is exactly the same idiom the core itself uses everywhere it needs to walk a `const GstTensorsInfo *` by index — `gst_tensors_info_copy()`, `gst_tensors_info_validate()`, `gst_tensors_info_is_equal()`, and `gst_tensors_info_get_size()` in `nnstreamer_plugin_api_util_impl.c` all do the identical `(GstTensorsInfo *) src` cast for the same reason. `allocMats` is consistent with established repo convention, not introducing a new risk category.
- Both actual call sites (`configure_instance`'s `inputInfo`, a member, and `setInputInfo`'s `given`, a local) pass genuinely non-const underlying objects, so there is no UB today — the const-cast only ever "restores" mutability to an object that was never const to begin with. It would only become a real problem if a future caller passed a truly `const`-declared `GstTensorsInfo`, which doesn't happen anywhere in this file.
- Practically, for `ncnn_subplugin` the case where this side effect (allocating `.extra`) even fires requires more than `NNS_TENSOR_MEMORY_MAX` input/output tensors, which is far outside anything the ncnn models this subplugin targets would have.

Not a blocking issue, just flagging the fragility for awareness — no change requested.

**`inferOutputInfo(mats, info)` throwing while filling `info` — no leak at the caller.** `setInputInfo()` passes its own local `inferred` (not the live `outputInfo` member) as the `info` out-parameter, and the single `catch (...)` block frees `inferred` unconditionally, covering every throw point inside `inferOutputInfo` (the `use_yolo_decoder` check, the `NNS_TENSOR_SIZE_LIMIT` check, and the per-tensor `ex.extract()` failure). `configure_instance()`'s call, `inferOutputInfo (input_mats, outputInfo)`, passes the member directly, but a throw there propagates straight out of `configure_instance()` and aborts construction of the whole subplugin instance (the core discards the instance on a `configure_instance` exception), so there's no half-updated live state to worry about in that call site either — consistent with how the rest of `configure_instance()` already behaves on any of its other `throw std::invalid_argument(...)` calls.

## 3. Re-check of the original review criteria (PR as a whole)

These are unchanged by `8721857a` (which touches only `tensor_filter_ncnn.cc`, and none of the test/CI files), so the previous review's conclusions still hold; re-verified rather than re-derived from scratch:
- **Resolves issue #4642 for ncnn**: yes — confirmed again by directly running the "no properties" pipeline (case 9) end-to-end against the real model and getting the correct golden-test result.
- **Regression risk to other modules**: none — reconfirmed `gst/` (core) has zero changes across the whole PR; `tensor_filter_ncnn.cc` remains a leaf module.
- **Scope/code size**: unchanged conclusion — well-scoped, no unrelated modules touched. The follow-up commit itself is a tight refactor (+51/-42 lines) proportionate to fixing the specific issue raised.
- **Test coverage**: unchanged — `8721857a` adds no new tests, but none were needed: it doesn't change externally observable behavior (same inputs produce the same outputs; the fix only prevents a hidden internal inconsistency that the core's own `-EINVAL` hard-fail already made unreachable). Re-ran the existing SSAT-equivalent pipelines by hand (case 9, case 15_n) to confirm nothing regressed.
- **CI enforcement**: unchanged — verified independently that the exact pinned ncnn release version referenced by `.github/workflows/ubuntu_clean_meson_build.yml` builds and links this file cleanly.
- **Docs/architecture**: unchanged — no core API/architecture touched; the only doc-like artifact is the `getShape()` comment, now corrected as noted above.

## Issues found

No critical or major issues in `8721857a`. One pre-existing, non-blocking, informational nit carried over from the surrounding code style (not introduced by this commit):

1. **(nit / informational)** `allocMats()`'s `const_cast`-via-C-style-cast on its `const GstTensorsInfo &` parameter is safe today (see analysis above) but relies on both current call sites happening to pass non-const underlying objects. Since it mirrors the exact idiom already used throughout `nnstreamer_plugin_api_util_impl.c`, no change is requested.

## Verdict

**Approve / OK to merge.** Commit `8721857a` correctly and completely addresses both points from the previous review round (the half-applied `setInputInfo()` candidate and the `getShape()` Doxygen wording), and the added indirection (`allocMats()` as a static free function, `inferOutputInfo()` taking its target `mats`/`info` as arguments) does not introduce any new leak, double-free, or exception-safety gap — verified by manual ownership tracing through `gst_tensors_info_*` and by independently building the branch and exercising both the success path (case 9) and the exception path (case 15_n, including its retry) against the real ncnn prebuilt and test model. The original review's conclusions on issue resolution, regression risk, scope, test coverage, CI enforcement, and documentation are unaffected by this follow-up commit and still hold.

myungjoo:
### CI status

Rebased onto `main` (`7cf17438`) and pushed `227ef689`, which drops the `const_cast` the second review flagged: `gst_tensors_info_get_nth_info()` lazily allocates the extra tensor array, so it wants a mutable `GstTensorsInfo`, and both call sites pass one — `allocMats()` now takes the reference mutable instead of casting the const away.

**The ncnn steps did what they were added to do.** On the previous run the three new steps all succeeded, and the SSAT group really ran rather than silently skipping:

```
[PASSED] nnstreamer_filter_ncnn (21 passed among 21 cases)
         TC Passed: 21 / Failed: 0 / Ignored: 0
```

**The one failing job is unrelated to this PR.** `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` was cancelled while `unittest_join` hung:

```
[ RUN      ] join.padReleaseWhileStreaming
2026-08-28T03:26:03Z ##[error]The operation was canceled.
```

Evidence that it is not caused by this change:

- `join.padReleaseWhileStreaming` is not in this branch. CI builds the merge ref, and the test came in from `main` in f726729d ("[Join] Stop the streaming thread before releasing a sink pad") a little over half an hour before this run.
- `main` already carries two follow-ups bounding that same test — 8265511a "[Test/Join] Bound the pad release race test by wall clock" and 4e4e282c "[Test/Join] Name the pad release race bounds" — so it is a known timing-sensitive pad-release race, and valgrind is the environment most likely to expose it.
- The previous run of this branch passed the very same job with that test already present, so it is intermittent.
- This PR touches no file under `gst/`. The ncnn subplugin is a leaf that `unittest_join` never loads.

I have left it alone rather than papering over someone else's in-flight work on that test. If the new run trips the same hang, it is that race and not this change.

myungjoo:
This is the third (final) round of review, performed by an automated review agent.

## 1. Does `227ef689` ("[Filter/NCNN] Drop the const cast in allocMats") correctly resolve the round-2 nit, without side effects?

Yes.

```
-static std::vector<ncnn::Mat> allocMats (const GstTensorsInfo &info);
+static std::vector<ncnn::Mat> allocMats (GstTensorsInfo &info);
...
-    mats.push_back (allocMat (
-        gst_tensors_info_get_nth_info ((GstTensorsInfo *) std::addressof (info), i)));
+    mats.push_back (allocMat (gst_tensors_info_get_nth_info (std::addressof (info), i)));
```

`gst_tensors_info_get_nth_info` is declared as `GstTensorsInfo * info` (mutable pointer) in `gst/nnstreamer/include/nnstreamer_plugin_api_util.h:112`, so the parameter change makes the signature say exactly what the function needs, and `std::addressof(info)` now type-checks without any cast.

Both call sites pass mutable objects, confirmed by reading their declarations directly:
- `configure_instance()` calls `allocMats (inputInfo)`, where `inputInfo` is the non-const member field `GstTensorsInfo inputInfo;` (line 97).
- `setInputInfo()` calls `allocMats (given)`, where `given` is a non-const local, `GstTensorsInfo given, inferred;` (line 287).

No half-applied state, no dangling const issue, no other call site to check — `allocMats` has exactly these two callers. `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)` on the latest run (227ef689) already passed, confirming this compiles clean.

## 2. Rebase integration onto latest `main` (7cf17438)

Diffed `227ef689` against the pre-rebase review head `8721857a` restricted to this PR's three touched files:

```
git diff 8721857a 227ef689 -- ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc \
    tests/nnstreamer_filter_ncnn/runTest.sh .github/workflows/ubuntu_clean_meson_build.yml
```

The only change is the `allocMats` hunk above — the rebase carried the PR's content forward byte-for-byte otherwise. No merge-conflict artifacts, no accidental drops.

Specifically checked the `tests/meson.build` / `runTest.sh` interaction flagged as a risk area: `install_subdir('nnstreamer_filter_ncnn', install_dir: unittest_install_dir)` (tests/meson.build:511) already existed before this PR's base commit (`ae2675c6^`), and `7cf17438` ("[Test] Install the SSAT suites that were missing from unittest-nnstreamer", pulled in by the rebase) adds *other* previously-missing suites — it does not touch the ncnn entry at all. So there is no interaction to reconcile: this PR's `runTest.sh` changes and main's `tests/meson.build` changes are on independent, non-overlapping suites.

Also verified `gstreamer_join` / `gst/join` are untouched by this PR's diff (`git diff origin/main...HEAD` has zero hits outside two comment lines that happen to contain the substring "join threads" in unrelated ncnn thread-joining code) — the pad-release race work that landed on `main` right before the rebase base is fully orthogonal to this PR.

## 3. Full-PR re-check (`git diff origin/main...HEAD`, 3 files, +370/-148)

Unchanged from the round-2 conclusions, reconfirmed rather than re-derived:
- **Resolves #4642 for ncnn**: yes, unchanged since round 2.
- **Regression risk**: none — `gst/` core still has zero changes across the whole PR.
- **Scope**: unchanged — three files (`tensor_filter_ncnn.cc`, `tests/nnstreamer_filter_ncnn/runTest.sh`, `.github/workflows/ubuntu_clean_meson_build.yml`), same as round 2 plus the one-hunk nit fix.
- **Test coverage**: unchanged — the round-2 test additions (cases 9–15_n covering optional dimension/type properties, trailing-1 padding, single-frame init, unsupported type, yolo-decoder) are intact post-rebase.
- **CI enforcement**: the dedicated ncnn install/build/SSAT-run/`.so`-existence-guard steps added to `ubuntu_clean_meson_build.yml` are present and unmodified by the rebase.
- **Docs**: no change.

## 4. Independent verification of the CI-failure claim (comment https://github.com/nnstreamer/nnstreamer/pull/4899#issuecomment-5448325586)

Verified independently, not just re-read.

**The referenced failing run is `33137182707` (commit `8721857a`, the pre-rebase head), job `98739654821`.** Its raw log shows:

```
03:22:42.3437647Z [ RUN      ] join.padReleaseWhileStreaming
03:26:03.9165189Z ##[error]The operation was canceled.
```

i.e. the job sat inside `join.padReleaseWhileStreaming` for ~3m21s before GitHub Actions cancelled it (a timeout/hang, not a crash reported by the test itself, not a valgrind-detected memory error, and not an nnstreamer_filter_ncnn failure — the ncnn SSAT step earlier in the same job log already reported `[PASSED] nnstreamer_filter_ncnn (21 passed among 21 cases)`).

Cross-checked the author's supporting claims:
- `f726729d` "[Join] Stop the streaming thread before releasing a sink pad" and the two bounding follow-ups `8265511a`/`4e4e282c` are all real commits reachable from `main`, all touching only `tests/gstreamer_join/unittest_join.cc` / `gst/join`.
- This PR's diff touches none of those files (confirmed above under point 2).
- This PR's own `Minimal meson build in Ubuntu with Valgrind` run passed once before, at commit `0b06b6bb` (run `33136619461`), before that join-test change landed on `main` — consistent with an intermittent, newly-introduced race rather than something this PR causes.

Conclusion: the author's claim is correct and independently verified. The `join.padReleaseWhileStreaming` hang under Valgrind is a pre-existing, timing-sensitive race on `main`, unrelated to this PR's ncnn-only changes.

**Latest CI status (post-rebase, commit `227ef689`, run `33141262xxx`)**, checked at review time:
- `Build on Ubuntu with LLVM/Clang`, `Build on MacOS-latest with Meson`, `DCO`, `Static checks`, `CPP-Linter`, `Spell Check`, `Analyze (python)` — all pass.
- `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04 / -arm)` — still in progress at review time; worth a final glance once it completes, but given the independent root-cause confirmation above, a repeat hang in `join.padReleaseWhileStreaming` would not be a reason to block this PR.
- `Tizen GBS build on Ubuntu (aarch64/armv7l/x86_64)` — **failing on this run**, but for an unrelated infrastructure reason: the aarch64 job log shows `E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/jammy/InRelease  403  Forbidden` during `apt-get update`, followed by `gbs: command not found` (exit 127) — the GBS tool itself never got installed because an unrelated apt source (Microsoft's azure-cli repo) returned 403 on this runner. This job passed cleanly on both prior runs of this same branch (`8721857a`, `0b06b6bb`), so this is transient CI-environment flakiness, not something introduced by the rebase or the `allocMats` commit.

## Issues found

None, in either the reviewed commit or the rebase. No critical/major/minor/nit findings remain open.

## Verdict

**Approve / OK to merge.** Commit `227ef689` correctly and completely resolves the round-2 nit with no side effects, the rebase onto `main@7cf17438` carried this PR's content forward unchanged, and the interaction with main's SSAT-install and join-test changes is confirmed to be non-overlapping. The `Build with meson and test with Valgrind` failure on the pre-rebase run is independently confirmed to be the pre-existing `join.padReleaseWhileStreaming` race on `main`, not this PR. The currently-failing Tizen GBS jobs on this run are unrelated apt-repo infrastructure flakiness (`azure-cli` repo 403 → `gbs: command not found`), also not this PR. All three prior review rounds' conclusions on issue resolution, regression risk, scope, test coverage, CI enforcement, and documentation stand unchanged.

myungjoo:
### Ready for review — CI green

All 18 checks pass on `227ef689`.

The `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` job, which is where the ncnn install and the new SSAT step live, is green and ran the cases rather than skipping them:

```
[PASSED] nnstreamer_filter_ncnn (21 passed among 21 cases)
         TC Passed: 21 / Failed: 0 / Ignored: 0
```

Both earlier red jobs were transient and unrelated:

- **`join.padReleaseWhileStreaming` hang** — the `gst/join` pad-release race that came in from `main`, detailed above. It did not recur on this run.
- **Tizen GBS ×3** — the run was cancelled 35 s in, while `apt` was still installing gbs itself, before any repository code was reached. Re-running the three jobs turned them green. This PR cannot affect them in any case: it does not touch `packaging/nnstreamer.spec`, ncnn is not in the spec so the subplugin is never built under GBS, and the ncnn test suite install is gated behind `ncnn_support_is_available`.

Three independent review rounds were run against this branch; the second and third were re-reviews after the changes each round asked for. All findings are addressed and no issue is outstanding.

myungjoo:
Thanks — all three addressed. The first two are fixed in 23c831f5.

**1. Yolo row width (`out.w >= 6`).** Agreed, and it is worse than a read: `values[2]`..`values[5]` are clamped back in place, so a narrow detection row is written past as well as read past. Checked once per output tensor, since the shape does not change between frames:

```c
if (out.w < 6)
  throw std::runtime_error (
      "The yolo decoder reads a label, a score and 4 corners out of every detection, "
      "but the model produced only " + std::to_string (out.w) + " values per row.");
```

You said this was the one thing to look at before merge, so it comes with a test rather than just a guard. `passthrough.param` is a network of a single `Input` layer — three lines of text and an empty `.bin` — which hands its input straight back, so the shape the decoder sees is whatever the pad caps carry. Case `16_n` feeds it a 3x2 GRAY8 frame, which arrives at the decoder as a row of one value, and expects the pipeline to fail.

It is a real regression test, not decoration: **without the width check the case fails** (21 passed / 1 failed), because the six-float read stays inside the matrix allocation and the pipeline succeeds. Squeezenet cannot cover this — its output row is 1000 wide.

**2. Silent row truncation.** Added, but warned once rather than per frame: this fires from `invoke()`, so an unconditional `ml_logw` would repeat at the frame rate for as long as the pipeline runs. A `yolo_rows_dropped` flag reports the first frame that drops anything, with both counts.

**3. Collision with #4379.** Confirmed, and it is more than the workflow. #4379 was updated today and independently fixes the same `cstep` padding bug — its `copy_padded()` against this PR's `copyToMat`/`copyFromMat` — plus it adds a type/size check, renumbers the negative cases `2`..`7` to `1_n`..`6_n` with a `120` timeout on every `gstTest`, and carries a `get_physical_cpu_count() < 1` guard for the SIGFPE in a GBS chroot that this PR does not have. Its ncnn pin is also better argued than mine: `20240102` is what Tizen ships, where `20260526` is just the newest.

**Decision: this PR goes first and #4379 rebases onto it.** The `.cc` rewrite here is +308/-145 against their +116/-21, and a large rewrite absorbing a small one is the harder direction to do second — so the smaller change should be the one that re-lands. @wooksong, that means the ncnn install step and the `.so` assertion will already be in the workflow when you rebase, and the SSAT file will hold cases 1-8 plus 9-16 rather than the `1_n`..`6_n` renumbering. Your CPU-topology guard, the Tizen spec change and the version re-pin to `20240102` all still apply cleanly on top and are worth keeping.

Local re-verification against the prebuilt ncnn release the CI step installs: 22/22, twice. clang-format 16 reports no replacements.

myungjoo:
### CI green — 18/18 on `a62d3fab`

Two of the three earlier failures were mine to fix and are fixed by rebasing onto current `main`; the branch had fallen 21 commits behind and predated the fixes for both:

| failure | fixed by |
|---|---|
| `Ubuntu pdebuild` — `nnstreamerFilterOnnxRuntime.quantModelLaunch00` | `2fea8422` (issue #4870) |
| `build and unit test` — `llamacpp.multipleInputsSingleOutputSync_p`, `*new_sample_count` 1 vs 2 | `f888c072` + `38ae0ea0` |

The llamacpp assertion was on a sink counter, which is exactly what `f888c072` made atomic, so the match is not a guess.

**The third is not this PR.** The `GTEST run with Valgrind` step intermittently dies in `join.padReleaseWhileStreaming`. The same test, at the same point, killed the same step on `litert-shared-environment` (run 33225216279), a branch with no ncnn in it:

```
[ RUN      ] join.padReleaseWhileStreaming
##[error]The operation was canceled.
```

Why it is intermittent: `run_unittests_binaries.sh` walks `find`-order and breaks at the first binary that returns non-zero, so whether the valgrind pass even reaches `unittest_join` varies per run — which is also why that step ranges from 12 to 49 minutes across recent runs. And the test is right at the edge rather than deadlocked: on the re-run it passed, in **45,289 ms**, against **255 ms** for the same test under plain `meson test`.

That is a `gst/join` matter, so I have left it alone rather than skipping `unittest_join` in the valgrind step, which would only hide it.

ncnn results in both the failed and the re-run job: `nnstreamer_filter_ncnn (22 passed among 22 cases)`.

myungjoo:
> _This review was produced by a separate automated review agent (not the one that authored the earlier relayed review) and is posted here directly. Please verify before acting on it._

**Scope reviewed:** full PR diff (`git diff origin/main...HEAD`) at head `a62d3fab`, with focused re-verification of the follow-up commit `a62d3fab "[Filter/NCNN] Bound the yolo detection row"` against the two non-blocking findings from the previous COMMENTED review (5048335568).

## Verification of the two follow-up findings

**Finding 1 (out-of-bounds read of `values[0..5]` without checking `out.w >= 6`) — fixed correctly.**

- The new `if (out.w < 6) throw std::runtime_error(...)` sits after the `if (!use_yolo_decoder) { copyFromMat(...); continue; }` early-exit, so it only executes on the `use_yolo_decoder` path and never affects classification-style outputs (verified by re-running case 1/9/10/11/12/13, none of which use the decoder).
- I built the PR's exact commit in a local environment against ncnn 20260526 (the version this PR pins) and ran the full `tests/nnstreamer_filter_ncnn` SSAT suite: **21/21 pass**, including the new `16_n`.
- I then mechanically removed only the `out.w < 6` guard, rebuilt, and reran: **case `16_n` fails** (the pipeline no longer errors, so the negative test correctly detects the missing guard), while all other cases stay green. Restoring the guard and rebuilding brings the suite back to 21/21. This confirms `16_n` is a real, load-bearing regression test, not a vacuous pass.
- I also traced `passthrough.param`/`passthrough.bin` against ncnn's `NetPrivate::update_input_output_indexes()`: a blob produced by an `Input`-type layer with no consumer is placed in *both* `input_blob_indexes` and `output_blob_indexes`, and `Extractor::extract()` returns a blob directly without re-running its producer layer if it was already fed via `ex.input()` (`blob_mats[idx].dims != 0` short-circuits). So the fixture genuinely round-trips the pad-cap shape through the model, and the empirical test confirms `out.w` ends up 1 for the 3x2 GRAY8 case used by `16_n` — well below the required 6, so it exercises exactly the intended path.
- The exception path was confirmed to actually abort the pipeline: `tensor_filter_support_cc.cc`'s catch ladder turns `std::runtime_error` into `-EINVAL`, and `_gst_tensor_filter_transform_check_invoke_result()` in `tensor_filter.c` turns any negative `invoke_res` into `GST_FLOW_ERROR`, which is what SSAT's `_n` convention expects and what the live run showed.

**Finding 2 (dropped detection rows warned only once) — fixed correctly.**

- `yolo_rows_dropped` is a plain `bool` member, initialized in the constructor's init list (`yolo_rows_dropped (false)`), so no UB from reading an uninitialized flag on the first invoke.
- The `ml_logw(...)` call is guarded by `!yolo_rows_dropped` and immediately sets the flag before logging, so it fires at most once per subplugin instance regardless of how many frames keep exceeding `max_detection` — this does stop the per-frame log spam the earlier review flagged.
- `ml_logw` is an existing, widely used macro in this codebase (`nnstreamer_log.h`), so no new dependency here.

No new defects were introduced by this commit as far as I can tell — the diff is minimal (15 lines in the `.cc`, plus the test fixture and case), well-scoped to exactly the two findings, and doesn't touch any other code path.

## Full-PR final pass

- **Solves the stated issue.** #4642 asked that `input`/`inputtype`/`output`/`outputtype` not be mandatory for `tensor_filter`; this PR routes ncnn through the core's existing `GET_IN_OUT_INFO` / `SET_INPUT_INFO` fallback instead of hard-failing in `configure_instance()`, which is the documented mechanism other subplugins already use.
- **No regression to existing behavior/other modules.** The file is a leaf (`ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc`) — nothing else in the tree includes or links it, and no core file is touched. CI's 18/18 green run (independently confirmed via `gh pr checks 4899`) covers the rest of the tree.
- **Scope/size.** All non-test changes are confined to the ncnn subplugin and its own CI wiring; nothing unrelated is touched.
- **Test coverage.** Both a positive-path gap (rows narrower than the yolo decoder needs) and the silent-drop UX gap now have dedicated coverage (`16_n`, and the `ml_logw` change is directly exercised by the existing `max_detection`-constrained cases). Combined with the CI step added earlier in the PR (`test -f .../libnnstreamer_filter_ncnn.so` plus the dedicated `SSAT run on filter-ncnn` step, both gated to `ubuntu-22.04`), a future regression in this file — or in the ncnn/nnstreamer core contract this subplugin depends on — will fail CI and block merge rather than silently passing, which is what let this SSAT suite report 0-case "success" since Dec 2023.
- **Docs.** The file's own top-of-file Doxygen comment (`@note Special considerations on properties`) was already updated earlier in the PR to state that all four properties are optional; that's the right and sufficient place for this, and I don't see any other doc (e.g. `Documentation/component-description.md`) that makes a stronger claim needing correction.
- **Item 3 from the previous review (collision with #4379 touching the same file/tests/workflow section)** — confirmed still factually true (both PRs modify `tensor_filter_ncnn.cc`, `runTest.sh`, and the same CI workflow block with different ncnn versions pinned), but per the maintainer's decision this is being handled by landing this PR first and having #4379 rebase afterward. Not re-raised as a blocking item here.

## Nit (non-blocking, optional)

The top-of-file comment documenting `use_yolo_decoder` (lines ~41-49) describes the `output=(5+num_labels, max_detection, 1)` property shape but doesn't mention that the model's *native* output row must itself carry at least 6 values (label + score + 4 corners) for the decoder to accept it — that's now a runtime error message only. Worth a one-line addition if there's a future doc pass on this file, but the runtime error message is already self-explanatory, so this is not worth blocking on.

## Verdict

**Approve / OK to merge.** Both non-blocking findings from the previous review are correctly and minimally resolved, verified by building against the pinned ncnn 20260526 and round-tripping the new SSAT case both with and without the fix. No new defects found. CI is green (18/18, independently re-checked). The one remaining item from the prior review (the #4379 collision) is a merge-order decision already made outside this PR's scope.

myungjoo:
Thanks — the nit is taken in 41b245b5, so nothing is left open on this PR.

The `use_yolo_decoder` note described the `output=` property to set but said nothing about what the model itself has to produce, which left both bounds the decoder enforces discoverable only from a runtime error. Three comment lines now state them:

```
 *        The decoder reads a label, a score and 4 corners out of every
 *        detection, so a model whose output rows are narrower than those 6
 *        values is rejected. Detections past max_detection are dropped.
```

Comment only — no code change. Rebuilt and re-ran the suite against the pinned ncnn 20260526 anyway: 22/22, clang-format 16 clean.

myungjoo:
The blocking finding is fixed in b315b09d. My apologies for the three rounds it took — I had been reading the issue-comment thread and missed that reviews 5057371320 and 5062631473 were formal CHANGES_REQUESTED, so I answered the nits from 5048335568 and reported "nothing left open" while this stood. It was raised three times and was correct every time.

**Confirmed before fixing, not just accepted.** The finding needs `out.w >= 6` to clear the earlier bound and `out.h >= max_rows` to reach the last row, so I built a case that satisfies both and watched it happen:

```
tensor_filter sink caps = 8:2:3:1  ->  ncnn Mat(w=8, h=2, c=3)     out.w = 8, clears the >= 6 bound
output=4:2:1                       ->  8 floats, label_count 4, max_rows 2
  out.h (2) >= max_rows (2), so row 1 is written
  row 1 starts at float 4 and writes floats 4..8  ->  float 8 is one past an 8-float buffer
exit=0
```

The pipeline reports success while writing 4 bytes past a buffer the core sized with `gst_tensor_info_get_size()`, exactly as described.

**The guard** is in `configure_instance()` next to the other property checks, so it fails at pipeline construction rather than on a buffer:

```c
if (output_from_prop) {
  normalizeInfo (outputInfo, net.output_indexes ().size (), "output");
  if (use_yolo_decoder)
    checkYoloOutputInfo (outputInfo);
}
```

`checkYoloOutputInfo()` rejects `dimension[0] < 5` per output tensor and names the tensor and the value. Your note that `label_count == 0` is unreachable holds — `normalizeInfo()` calls `getShape()`, which throws on an empty dimension — so the single `>= 5` bound is the whole fix.

**`17_n` is load-bearing.** Your suggested `output=4:100:1` would not have been: with the `passthrough` fixture the source shape drives `out.h`, and at `max_rows = 100` the loop stops long before the last row, so the case would have passed on the earlier `out.w < 6` throw instead and scored green either way. `17_n` therefore uses an 8x2 RGB source transposed to `8:2:3:1` with `output=4:2:1`, which clears the width bound and saturates the buffer. Verified by removing only the new call, rebuilding and re-running: **`17_n` alone fails** (23 -> 22 passed, 1 failed), pipeline exits 0; restored, 23/23 twice.

Suite is 23/23 against the pinned ncnn 20260526, clang-format 16 clean, no build warnings.

myungjoo:
> _This review was produced by a separate AI review agent (independent from the `myungjoo-bot` review account seen on this thread) and is posted here as a plain comment, not a formal review. Please verify before acting on it._

## Re-review at head `392abaa2` (previous state: `CHANGES_REQUESTED` x2, then a COMMENTED re-affirmation, then an `APPROVED` review from `myungjoo-bot` at `5084416811`)

This pass independently re-derives the same conclusion the last review reached, but backs it with an actual build instead of static reasoning alone. Summary up front: **no blocking issues found; the previously-blocking finding is fixed and the regression test is real.** One pre-existing doc nit is newly visible after the rebase and was not caught by any prior round.

## Blocking finding: `label_count < 5` out-of-bounds write — confirmed fixed

Read `checkYoloOutputInfo()` and its call site in `configure_instance()` (`ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc:222-226`, function at `:294-307`) and traced every path that can populate `outputInfo` when `use_yolo_decoder` is set:

- `output_from_prop` true → `normalizeInfo()` then `checkYoloOutputInfo()`, both before `configure_instance()` returns. Guarded.
- `output_from_prop` false → the only writer of `outputInfo` is `inferOutputInfo()`, called either directly in `configure_instance()` (`:228-229`) or from `setInputInfo()` (`:336-337`, itself only reached via `SET_INPUT_INFO`). `inferOutputInfo()` throws unconditionally when `use_yolo_decoder` is true (`:380-383`), before it ever touches `outputInfo`. So this branch never leaves a live, unguarded `outputInfo` in place.
- `use_yolo_decoder` itself is set by `parseCustomProperties()` at `:180`, strictly before either branch above runs, so the flag is never stale when the checks fire.

So every path that can reach the five unconditional writes in `invoke()` (`output_data[0]`..`[4]` at `:544-548`) has gone through the `label_count < 5` check first. The threshold is exactly right: at `label_count == 5` the last row occupies floats `5*(max_rows-1)`..`5*max_rows-1`, which is the final in-bounds float of a `label_count*max_rows`-float buffer; at 4 it overruns by one float (4 bytes).

## `17_n` — independently verified as a real regression test, not vacuous

Did this by building rather than reading:

1. Built the worktree at `392abaa2` as-is (WSL, ncnn prebuilt `20260526`) and ran `ssat -n -p=1` on `tests/nnstreamer_filter_ncnn`: **23/23 pass**, including `17_n`.
2. Removed only the guard's call site in `configure_instance()` (kept `checkYoloOutputInfo()` defined but unused, so this isolates exactly the behavior the guard adds) and rebuilt. Result: **22/23 — `17_n` is the only failure**, everything else including `16_n` (the unrelated `out.w < 6` case) still passes. This confirms the test fails without the fix and passes with it, i.e. it is a genuine regression test.
3. Checked the previous reviewer's originally-suggested alternative, `output=4:100:1`, against the guard-removed build, in two shapes:
   - Using a narrow source (`GRAY8`, `width=3,height=2`, as in `16_n`): fails, but via the *pre-existing* `"The yolo decoder reads a label, a score and 4 corners..."` (`out.w < 6`) error — not via any label-count logic, since that guard is removed in this build. So this configuration cannot distinguish "guard present" from "guard absent."
   - Using the PR's own 8×2 RGB source (`out.w == 8`, clears the width check) with `output=4:100:1` (`max_rows=100`): pipeline runs to completion with **RC=0**, because the loop only reaches `min(out.h, max_rows) == 2` rows and never gets near row 99 where the overrun would occur.
   
   Both confirm the author's stated reason (in the `392abaa2` commit message) for rejecting `output=4:100:1` in favor of `output=4:2:1` with the 8×2 source: the former either trips an unrelated guard or never saturates the buffer, and would have been a vacuous `_n` case (it would "pass" — i.e., correctly report failure or correctly not-fail — for reasons unrelated to this fix). `output=4:2:1` with `max_rows == out.h == 2` is the minimal case that actually reaches the last, overrunning row.

Restored the original file afterward and re-verified 23/23 green at `392abaa2` before finishing.

## Commit `392abaa2` itself — no new defects, rebase is clean

- The diff is additive and localized: one new private static method, one call site, one doc comment on the method, and the new `17_n` case. Doesn't touch `invoke()`'s existing bounds (`out.w < 6`, `label + 5 < label_count`, `j < max_rows`) added by earlier commits in this PR.
- `git diff origin/main...HEAD -- ext/nnstreamer/tensor_filter/tensor_filter_ncnn.cc tests/nnstreamer_filter_ncnn/ .github/workflows/ubuntu_clean_meson_build.yml` shows only this PR's own hunks; nothing from upstream `main` leaked in or was clobbered.
- All 18 CI checks are green at this head (confirmed via `gh pr view --json statusCheckRollup`), including `Ubuntu pdebuild` and both meson+Valgrind jobs, which is what the two `CHANGES_REQUESTED` rounds were also waiting on.

## Everything from earlier rounds, checked off

| Finding | Status |
|---|---|
| `out.w < 6` unchecked before reading 6 columns | Fixed (`a62d3fab`) |
| Rows past `max_rows` dropped silently | Fixed, `yolo_rows_dropped` warning added (`a62d3fab`), documented (`41b245b5`) |
| `label_count < 5` unconditional-write OOB | Fixed (`392abaa2`), independently re-verified above |
| CI never actually built/ran the ncnn plugin | Fixed — ncnn install + `SSAT run on filter-ncnn` step + `.so` existence assertion, all green |
| Collision with #4379 | Acknowledged only, per the stated human decision (#4899 merges first, #4379 rebases) — not re-raised here |
| Stale meson-floor-guard comment after rebase | **Not resolved** — see below (new, minor) |

## New minor finding (nit): stale comment in the CI workflow after rebase

`.github/workflows/ubuntu_clean_meson_build.yml:147-150` (inherited from `main` via the rebase, commit `148177c3a`, not touched by this PR's own diff):

```yaml
# This sees only what this job configures. A block behind an SDK that
# is not installed here - SNPE, TensorRT, tensorflow, pytorch, ncnn,
# mvncsdk2 - is never visited, so it can still carry a construct newer
# than the floor.
```

This PR's own `install ncnn for the ncnn tensor-filter` step (`:87-100`) now installs ncnn unconditionally in this same job for `ubuntu-22.04`, so ncnn no longer belongs in that "not installed here" list — it's the same situation as LiteRT, which the comment correctly doesn't mention. This was flagged as a small follow-up in the prior COMMENTED review ("worth dropping from the comment in the rebase commit") but the line is unchanged post-rebase, presumably because it sits outside this PR's diff hunks so nothing surfaced it automatically.

**Severity: nit.** It's a documentation-accuracy issue only — the guard logic itself is unaffected (ncnn's `meson-log.txt` is now genuinely visited, so the comment is merely stale, not misleading about correctness). Does not need to block merge; worth a one-line fixup in this PR or immediately after.

## Verdict

**No blocking issues. OK to merge** on the technical merits — the out-of-bounds write is closed exactly where and how the earlier reviews asked for, `17_n` is a verified real regression test (I reproduced both the pass and the fail independently), and CI is green end to end. The only open item is the nit above, which is cosmetic.

myungjoo:
Thanks for the approval, and for catching the stale comment — both it and the rebase are done.

**Rebase.** Already on current `main`; the head you approved (`392abaa2`) is the rebased one, so `Ubuntu pdebuild` is green. Your diagnosis was exactly right: the failure was `meson.build:4:0: ERROR: Meson version is 0.61.2 but project requires >=0.62.0`, the `>=0.62.0` floor came from the stale base, and `7876ecf0` / `18b4f69f` on `main` corrected it. The rebase was textually clean, as you predicted — the hunks in `ubuntu_clean_meson_build.yml` are disjoint.

**The follow-up you flagged is applied in 24734a89.** `main`'s meson-floor guard listed the SDK-gated blocks it cannot see — "SNPE, TensorRT, tensorflow, pytorch, ncnn, mvncsdk2" — and this PR installs ncnn a few steps above it, so ncnn's block is now configured in that job and the guard does cover it. Dropped from the list. Comment only.

All 18 checks were green at `392abaa2`, with the ncnn group reporting `23 passed among 23 cases` including the new `17_n`.

One thing outside my reach: `reviewDecision` is still `REVIEW_REQUIRED` and `mergeStateStatus` is `BLOCKED` even with your approval recorded, so the merge gate appears to want an approving review from an account the branch protection counts. That is a repo policy matter rather than anything left in the change.

myungjoo:
Both remaining observations are taken, so nothing from any round is left open.

**The meson-floor comment was already done** — `24734a89`, pushed a couple of minutes before your comment landed, so you were reading `392abaa2`. `ncnn` is out of the "not installed here" list.

**In-place clamp of the extractor blob** — `c2992287`. You were right that `out.row(j)` is not ours to edit, and right that nothing observable came of it, since `copyToMat()` rewrites `input_mats` on the next invoke. The loop only needs the clamped numbers, so they go into locals now and `values` is `const float *`:

```c
const float x1 = fmaxf (fminf (values[2], 1.0), 0.0);
...
output_data[0] = (x1 + x2) / 2;
```

**`runTest.sh` skip arm** — same commit. The `else echo "Cannot find ${ini_path}"` branch fell through and ran every case against a plugin path it had just reported missing; it now does `report; exit` like the two arms either side of it. Checked by pointing `PATH_TO_PLUGIN` at a directory with no `ext/nnstreamer/tensor_filter`: it now reports `0 passed among 0 cases` instead of a wall of failures. You had called this one pre-existing and outside the diff, which it is, but it is two lines in a file this PR already rewrites.

Suite is 23/23 twice against the pinned ncnn 20260526, clang-format 16 clean, no build warnings.

Issue #4888, opened 2026-08-27
[CI/Valgrind] The valgrind unittest pass covers only 4 of 26 binaries and cannot fail the build


myungjoo:
The "GTEST run with Valgrind on a case" step in `.github/workflows/ubuntu_clean_meson_build.yml:169-171` looks like it runs every unittest binary under valgrind. It does not. In practice it gets through **4 of the 26 binaries** that job builds, and no valgrind finding can ever fail the build.

This was found while investigating whether a memory leak fixed in #4882 would be caught by CI. It would not have been, and neither would most others.

## Evidence

Two independent runs, same outcome:

- #4873 — https://github.com/nnstreamer/nnstreamer/actions/runs/33073918204/job/98523048068
- #4882 — https://github.com/nnstreamer/nnstreamer/actions/runs/33079190066/job/98541404559

In both, the job builds 26 `unittest_*` binaries but only 4 valgrind processes ever emit an `ERROR SUMMARY`, and the step ends on its fallback message.

## Root cause

`packaging/run_unittests_binaries.sh:100` breaks out of the loop as soon as one binary returns non-zero:

```sh
for entry in "${filelist[@]}"
do
  run_entry $entry
  ret=$?
  if [ $ret -ne 0 ]; then
    break
```

`unittest_filter_custom` consistently fails under valgrind — `tensorFilterCustom.flexibleInvoke_p` (13393 ms in the #4873 run; `staticFlexibleInvoke_p` also failed in the #4882 run). These look timing-sensitive, and valgrind's 20-50x slowdown is enough to trip them. The loop therefore stops there and the remaining ~22 binaries are never examined.

## Four separate layers prevent a leak from failing CI

Worth listing together, because fixing any one alone changes nothing:

1. The run aborts early, as above, so most binaries are never reached.
2. The valgrind invocation at `run_unittests_binaries.sh:78` has no `--error-exitcode`, so valgrind findings never influence the exit status. The non-zero that breaks the loop is the gtest failure, not the leaks.
3. The workflow step ends with `|| echo "There are Valgrind errors. Please fix it. ..."`, which swallows whatever did come back.
4. Frameworks whose subplugins are not installed in that job (tvm, for one) have no binary there to examine at all.

## Scale is smaller than the workflow comment suggests

[...]

PR #4900, opened 2026-08-28
[Test] Let the remaining SSAT suites find sub-plugins when installed

myungjoo:
Every `runTest.sh` probes `../../build` to decide whether the sub-plugin it needs was built, and most fall back to reading the installed path out of `/etc/nnstreamer.ini` when that directory is absent. Seven did not — they ended the probe with

```sh
else
    echo "No build directory"
    report
    exit
fi
```

so from the installed test package they report no cases at all. #4885 started shipping two of them (`filter_lua`, `sparse`) and the other five were already there, so `unittest-nnstreamer` has been carrying suites that could never run. This is the follow-up that #4885's description said was needed.

- Give the seven the same fallback the other suites use: `nnstreamer_converter_python3`, `nnstreamer_decoder_python3`, `nnstreamer_filter_lua`, `nnstreamer_flatbuf`, `nnstreamer_flexbuf`, `nnstreamer_protobuf`, `nnstreamer_sparse`. Each reads the ini key matching the kind of sub-plugin it actually needs — `filters`, `converters` or `decoders`.
- Restructure the probe instead of bolting an `else` onto it, so the sub-plugin lookup happens once against whichever path was resolved.
- Move the `nnstreamer_python.so` symlink setup in `converter_python3` / `decoder_python3` inside the build-tree branch, and guard the tail-end `rm -r ${TEST_PYTHONPATH}` on the variable that block sets.

**Which ini key each suite needs**

| suite | build-tree probe | ini key | looks for |
| --- | --- | --- | --- |
| `converter_python3` | `ext/nnstreamer/tensor_converter` | `converters` | `python3.so` |
| `decoder_python3` | `ext/nnstreamer/tensor_decoder` | `decoders` | `python3.so` |
| `filter_lua` | `ext/nnstreamer/tensor_filter` | `filters` | `lua.so` |
| `flatbuf` | `ext/nnstreamer/tensor_converter` | `converters` | `flatbuf.so` |
| `flexbuf` | `ext/nnstreamer/tensor_converter` | `converters` | `flexbuf.so` |
| `protobuf` | `ext/nnstreamer/extra` | `converters` | `protobuf.so` |
| `sparse` | `ext/nnstreamer/tensor_filter` | `filters` | `lua.so` |

`protobuf` is the odd one. Its build-tree probe looks at `ext/nnstreamer/extra`, which holds `libnnstreamer_protobuf.so` — a helper library installed to `%{_libdir}`, not a sub-plugin — so there is no ini key that points at it. The installed check looks for the converter sub-plugin instead, which the suite does exercise, and `debian/nnstreamer-protobuf.install` ships converter and decoder together.

**The symlink block**

`converter_python3` and `decoder_python3` each did this unconditionally after the probe:

```sh
pushd ../../build/ext/nnstreamer/tensor_converter
```

That would have failed the moment the new fallback let them past the probe — the guard would pass and the script would then die on a missing build tree, which is worse than skipping. The comment already states the symlink is only for testcases, and `filter_python3` has always wrapped the same block in `if [[ -d ../../build ]]`, so both now do the same. `debian/nnstreamer-python3.links` puts `nnstreamer_python.so` on the default `PYTHONPATH`, so the installed tree genuinely provides what the build-tree block was faking.

`filter_python3` itself is otherwise untouched: it already handles the installed case by checking for the `nnstreamer_python` module, so it was never in the inert set. It carried a `report exit` typo, which is fixed there as a one-liner.

**The build-tree arm changes too**

Restructuring the probe also fixes a second, separate defect, and it is worth being explicit that this is not confined to the new branch. In the arm taken when `../../build` exists but the sub-plugin directory does not, six of the seven previously fell through and ran the tests anyway:

- `converter_python3`, `filter_lua`, `sparse` had `report exit` — `report` called with a stray argument, no `exit` executed
- `flatbuf`, `flexbuf`, `protobuf` printed a message and did nothing else
- `decoder_python3` was the only one that correctly did `report` then `exit`

All seven now skip cleanly. This does not change any CI result, because those directories always exist in a CI build — but the reason is that the arm is unreachable there, not that the diff is confined to the installed path.

One consequence to be deliberate about: a build that disables the converter sub-plugins will now silently skip `flatbuf` / `flexbuf` / `protobuf` instead of running them and failing. Every other suite in the tree already behaves that way, so it is the consistent choice, but it does turn a loud failure into a quiet skip in that configuration.

**Verification**

CI cannot cover the installed path: no job runs SSAT from an installed tree, and none can without new plumbing — RPM runs `%check` before `%install` and debhelper runs `dh_auto_test` before `dh_auto_install`, so neither packaging flow has a point at which an installed tree exists. So I drove each script's probe directly with a stub SSAT API, against a fake ini pointing at fake sub-plugin directories and separately against a fake build tree:

- sub-plugin present, no build dir → guard passes and the script reaches its first `gstTest`, with no build-tree errors
- sub-plugin absent, no build dir → `Cannot find <name> shared lib`, `report`, clean exit
- no ini at all → `Cannot identify nnstreamer.ini`, `report`, clean exit
- build dir present, sub-plugin present → reaches the tests, unchanged
- build dir present, sub-plugin directory absent → clean skip (previously a fall-through for six of seven, as above)

The cleanup was checked both ways too: silent when `TEST_PYTHONPATH` is unset, still removing the directory when it is set.

**Self evaluation:**
1. Build test: [ ]Passed [ ]Failed [*]Skipped (no build input changed)
2. Run test: [*]Passed [ ]Failed [ ]Skipped (probe driven directly, as above; CI covers the build-tree path)

**How to evaluate:**
1. In a build tree, run `ssat` in each of the seven directories and confirm the results are unchanged.
2. Install with `-Dinstall-test=true`, then from `<prefix>/lib/nnstreamer/bin/unittest-nnstreamer/tests/<suite>` run `ssat` and confirm the suite now executes instead of printing `No build directory`.
3. Rename `/etc/nnstreamer.ini` and confirm each suite skips cleanly rather than failing.

### DIFF
--- a/tests/nnstreamer_converter_python3/runTest.sh
+++ b/tests/nnstreamer_converter_python3/runTest.sh
@@ -40,37 +40,45 @@ convertBMP2PNG
 
 PATH_TO_PLUGIN="../../build"
 # Check python libraries are built
-if [[ -d $PATH_TO_PLUGIN ]]; then
+if [[ -d ${PATH_TO_PLUGIN} ]]; then
     ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_converter"
-    if [[ -d ${ini_path} ]]; then
-        check=$(ls ${ini_path} | grep python3.so)
-        if [[ ! $check ]]; then
-            echo "Cannot find python shared lib"
-            report
-            exit
-        fi
-    else
-        echo "Cannot find ${ini_path}"
-        report exit
-    fi
 else
-    echo "No build directory"
+    ini_file="/etc/nnstreamer.ini"
+    if [[ ! -f ${ini_file} ]]; then
+        echo "Cannot identify nnstreamer.ini"
+        report
+        exit
+    fi
+
+    sub_plugin_line=$(grep "^converters" ${ini_file})
+    if [[ ${sub_plugin_line%%=*} != "converters" ]]; then
+        echo "Cannot identify the sub-plugin path from ${ini_file}"
+        report
+        exit
+    fi
+    ini_path=${sub_plugin_line##*=}
+fi
+
+if [[ ! -d ${ini_path} ]] || [[ -z $(ls ${ini_path} | grep python3.so) ]]; then
+    echo "Cannot find python shared lib"
     report
     exit
 fi
 
 FRAMEWORK="python3"
-# This symlink is necessary only for testcases; when installed, symlinks will be made
-pushd ../../build/ext/nnstreamer/tensor_converter
-TEST_PYTHONPATH=$(pwd)/${FRAMEWORK}_pymodule
-mkdir -p ${TEST_PYTHONPATH}
-pushd ${TEST_PYTHONPATH}
-export PYTHONPATH=$(pwd):${PYTHONPATH}
-if [[ ! -f ./nnstreamer_python.so ]]; then
-  ln -s ../../extra/nnstreamer_${FRAMEWORK}.so nnstreamer_python.so
+if [[ -d ${PATH_TO_PLUGIN} ]]; then
+    # This symlink is necessary only for testcases; when installed, symlinks will be made
+    pushd ${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_converter
+    TEST_PYTHONPATH=$(pwd)/${FRAMEWORK}_pymodule
+    mkdir -p ${TEST_PYTHONPATH}
+    pushd ${TEST_PYTHONPATH}
+    export PYTHONPATH=$(pwd):${PYTHONPATH}
+    if [[ ! -f ./nnstreamer_python.so ]]; then
+      ln -s ../../extra/nnstreamer_${FRAMEWORK}.so nnstreamer_python.so
+    fi
+    popd
+    popd
 fi
-popd
-popd
 
 PATH_TO_SCRIPT="../test_models/models/custom_converter.py"
 ##
@@ -139,6 +147,9 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} audiotestsrc num-buffers=1 samplesp
 callCompareTest test.audio8k.s16le.origin.log test.consecutive.log 4-1 "Consecutive converting test" 0 0
 
 rm *.golden *.log *.bmp *.png *.dat
-rm -r ${TEST_PYTHONPATH}
+# TEST_PYTHONPATH is only set when the symlink was made in a build tree.
+if [[ -n ${TEST_PYTHONPATH} ]]; then
+    rm -r ${TEST_PYTHONPATH}
+fi
 
 report
--- a/tests/nnstreamer_decoder_python3/runTest.sh
+++ b/tests/nnstreamer_decoder_python3/runTest.sh
@@ -40,38 +40,45 @@ convertBMP2PNG
 
 PATH_TO_PLUGIN="../../build"
 # Check python libraries are built
-if [[ -d $PATH_TO_PLUGIN ]]; then
+if [[ -d ${PATH_TO_PLUGIN} ]]; then
     ini_path="${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_decoder"
-    if [[ -d ${ini_path} ]]; then
-        check=$(ls ${ini_path} | grep python3.so)
-        if [[ ! $check ]]; then
-            echo "Cannot find python shared lib"
-            report
-            exit
-        fi
-    else
-        echo "Cannot find ${ini_path}"
+else
+    ini_file="/etc/nnstreamer.ini"
+    if [[ ! -f ${ini_file} ]]; then
+        echo "Cannot identify nnstreamer.ini"
         report
         exit
     fi
-else
-    echo "No build directory"
+
+    sub_plugin_line=$(grep "^decoders" ${ini_file})
+    if [[ ${sub_plugin_line%%=*} != "decoders" ]]; then
+        echo "Cannot identify the sub-plugin path from ${ini_file}"
+        report
+        exit
+    fi
+    ini_path=${sub_plugin_line##*=}
+fi
+
+if [[ ! -d ${ini_path} ]] || [[ -z $(ls ${ini_path} | grep python3.so) ]]; then
+    echo "Cannot find python shared lib"
     report
     exit
 fi
 
 FRAMEWORK="python3"
-# This symlink is necessary only for testcases; when installed, symlinks will be made
-pushd ../../build/ext/nnstreamer/tensor_decoder
-TEST_PYTHONPATH=$(pwd)/${FRAMEWORK}_pymodule
-mkdir -p ${TEST_PYTHONPATH}
-pushd ${TEST_PYTHONPATH}
-export PYTHONPATH=$(pwd):${PYTHONPATH}
-if [[ ! -f ./nnstreamer_python.so ]]; then
-  ln -s ../../extra/nnstreamer_${FRAMEWORK}.so nnstreamer_python.so
+if [[ -d ${PATH_TO_PLUGIN} ]]; then
+    # This symlink is necessary only for testcases; when installed, symlinks will be made
+    pushd ${PATH_TO_PLUGIN}/ext/nnstreamer/tensor_decoder
+    TEST_PYTHONPATH=$(pwd)/${FRAMEWORK}_pymodule
+    mkdir -p ${TEST_PYTHONPATH}
+    pushd ${TEST_PYTHONPATH}
+    export PYTHONPATH=$(pwd):${PYTHONPATH}
+    if [[ ! -f ./nnstreamer_python.so ]]; then
+      ln -s ../../extra/nnstreamer_${FRAMEWORK}.so nnstreamer_python.so
+    fi
+    popd
+    popd
 fi
-popd
-popd
 
 PATH_TO_SCRIPT="../test_models/models/custom_decoder.py"
 #
@@ -130,6 +137,9 @@ callCompareTest testsynch19_3.golden testsynch19_3.log 3-4 "Tensor mux Compare 3
 callCompareTest testsynch19_4.golden testsynch19_4.log 3-5 "Tensor mux Compare 3-5" 1 0
 
 rm *.golden *.log *.bmp *.png *.dat
-rm -r ${TEST_PYTHONPATH}
+# TEST_PYTHONPATH is only set when the symlink was made in a build tree.
+if [[ -n ${TEST_PYTHONPATH} ]]; then
+    rm -r ${TEST_PYTHONPATH}
+fi
 
 report
--- a/tests/nnstreamer_filter_lua/runTest.sh
+++ b/tests/nnstreamer_filter_lua/runTest.sh
@@ -21,21 +21,27 @@ testInit $1
 
 PATH_TO_PLUGIN="../../build"
 # Check lua libraries are built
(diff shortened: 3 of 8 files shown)


myungjoo:

I reviewed the diff (`tests/nnstreamer_{converter_python3,decoder_python3,filter_lua,filter_python3,flatbuf,flexbuf,protobuf,sparse}/runTest.sh`) against the stated goal and the rest of the repo (packaging specs, `nnstreamer.ini.in`, and the already-working install-fallback pattern in e.g. `nnstreamer_filter_tensorflow_lite/runTest.sh`).

**Scope / goal**: Matches the description. All 8 files are test-only, no other module is touched, and the diff size (157/-108) is proportionate to giving 7 scripts a real installed-mode fallback plus a one-line typo fix in the 8th.

**Verified correct:**
- ini key choice per suite matches the PR's table. Cross-checked against `debian/*.install` and `packaging/*.spec`: `libnnstreamer_converter_{python3,flatbuf,flexbuf,protobuf}.so` all land under `converters/`, `libnnstreamer_filter_lua.so` under `filters/`, `libnnstreamer_{decoder,filter}_python3.so` under `decoders/`/`filters/` — so every `grep <name>.so` against the ini-resolved directory will find the right file.
- `protobuf` reading the `converters` key (not `extra`) is correct despite the build-tree probe looking at `ext/nnstreamer/extra`: that directory holds the helper lib (`nnstreamer_protobuf.cc`, packaged as `libnnstreamer_protobuf.so` under `%{_libdir}`), while the actual converter sub-plugin (`tensor_converter_protobuf.cc`) is packaged as `libnnstreamer_converter_protobuf.so` under `converters/`. The installed check targets the sub-plugin actually exercised by the test, which is the right call.
- Build-tree behavior is unchanged for the two paths CI actually exercises (sub-plugin present → proceeds; sub-plugin absent → `report`+`exit`). Confirmed no CI job runs SSAT from an installed tree (`ubuntu_clean_meson_build.yml` always exports `NNSTREAMER_BUILD_ROOT_PATH`/`GST_PLUGIN_PATH` pointing at `build/`), so the new installed-mode branch is genuinely dead code from CI's perspective, as the PR body states.
- The `converter_python3`/`decoder_python3` symlink block moved correctly inside `if [[ -d ${PATH_TO_PLUGIN} ]]`, matching the pre-existing `filter_python3` pattern (`if [[ -d ../../build ]]; then ... fi`) exactly.
- Removing `report exit` is a real, safe behavior fix, not just cosmetic: `report exit` calls `report` with a stray argument and never actually executes the shell `exit` builtin, so previously the script fell through and kept running tests even when the sub-plugin/ini_path lookup had already failed. This bug existed identically in `filter_lua`/`sparse` (`else` branch) and — more subtly — in `converter_python3`/`decoder_python3`'s pre-PR `else` branch. The new unified `if [[ ! -d ... ]] || [[ -z ... ]]; then report; exit; fi` closes it for all 7 files.
- No leftover build-tree-only dependencies (no other `../../build` references, no `../` reach into sibling test dirs, no uninstalled helper scripts) found in any of the 7 scripts after the guard.

**One minor (low-severity, non-blocking) nit:**
In `nnstreamer_converter_python3/runTest.sh` (line ~150) and `nnstreamer_decoder_python3/runTest.sh` (line ~140), the trailing cleanup does `rm -r ${TEST_PYTHONPATH}`. `TEST_PYTHONPATH` is only assigned inside the `if [[ -d ${PATH_TO_PLUGIN} ]]` symlink block, so on the new installed-mode success path it's unset, and this becomes `rm -r` with no operand — a harmless "missing operand" stderr line (no `set -e`/`set -u` in these scripts, so it doesn't abort or fail the suite). Purely cosmetic; guarding it with `[[ -n "${TEST_PYTHONPATH}" ]]` would be tidy but isn't necessary for correctness.

**Informational (pre-existing, not introduced by this PR):** the new installed-mode fallback branch has no CI coverage (same as every other suite already using this pattern, e.g. `filter_tensorflow_lite`), since no workflow runs SSAT against `-Dinstall-test=true`. The PR body is upfront about this and describes manual verification of the guard logic instead, which is reasonable given the existing project-wide gap — just flagging it as something a future CI job (running SSAT from an installed prefix) would need to close, not something this PR needs to solve on its own.

**Documentation**: No architecture/API change, so no doc update is required.

**Overall**: No blocking issues found. The change does what it says, doesn't regress the build-tree path, and fixes a real latent bug (`report exit` fallthrough) as a byproduct. Mergeable as-is; the `rm -r` nit above is optional polish.

myungjoo:
Good catch on the `rm -r` — that one was introduced by this PR, not pre-existing, so it is fixed here.

Moving the symlink block into the build-tree branch left `TEST_PYTHONPATH` unset on the installed path, while the tail-end cleanup still referenced it unconditionally, so `rm -r` would have run with no operand and printed `missing operand` to stderr. Harmless — no `set -e`/`set -u` — but it is noise in exactly the mode this PR is trying to make work, which rather undercuts the point.

Guarded on the variable that the block sets:

```sh
# TEST_PYTHONPATH is only set when the symlink was made in a build tree.
if [[ -n ${TEST_PYTHONPATH} ]]; then
    rm -r ${TEST_PYTHONPATH}
fi
```

Checked both directions rather than assuming: with the variable unset the cleanup is silent, and with it set the directory is still removed. Applied to `converter_python3` and `decoder_python3`, which are the only two with that block.

On the second item — no CI coverage for the install-fallback branch — agreed, and it is why the PR body leads with the manual verification instead of claiming CI proves anything. Worth adding that the gap is structural rather than an oversight: RPM runs `%check` before `%install` and debhelper runs `dh_auto_test` before `dh_auto_install`, so neither packaging flow has a point at which an installed tree exists to test against. Creating one is a project-level change well outside this PR. #4888 tracks the related blind spot on the valgrind side.

Thanks in particular for reading all seven scripts to the end for leftover build-tree dependencies. That is where the real risk was — the symlink `pushd` was exactly that class of problem, and shipping a guard that passes only to die two lines later would have been worse than leaving the suites skipping.

myungjoo:

Second-round review of the amended commit (`19e88c70`, force-pushed over the reviewed one).

**1. Is the round-1 nit resolved?**

Yes. Both `nnstreamer_converter_python3/runTest.sh` and `nnstreamer_decoder_python3/runTest.sh` now guard the cleanup:

```sh
# TEST_PYTHONPATH is only set when the symlink was made in a build tree.
if [[ -n ${TEST_PYTHONPATH} ]]; then
    rm -r ${TEST_PYTHONPATH}
fi
```

`TEST_PYTHONPATH` is assigned only inside the `if [[ -d ${PATH_TO_PLUGIN} ]]` symlink block earlier in each script, so this guard is checking exactly the condition that determines whether the variable exists. I traced both branches directly in the current file content (not just the diff):
- Install tree (`${PATH_TO_PLUGIN}` absent): the symlink block is skipped, `TEST_PYTHONPATH` stays unset, `[[ -n ${TEST_PYTHONPATH} ]]` is false, `rm -r` is not invoked — no more "missing operand" stderr noise.
- Build tree (`${PATH_TO_PLUGIN}` present): the symlink block runs, `TEST_PYTHONPATH` is set to the pymodule dir, the guard is true, `rm -r` still cleans it up — unchanged from pre-PR behavior.

I also reran this exact guard logic standalone (unset vs. set) to confirm the branching, and it behaves as described in both directions.

**2. Any new issue from this fix?** None found. The guard is scoped correctly (only the two scripts that actually set `TEST_PYTHONPATH` got it), doesn't change the build-tree code path CI exercises, and doesn't touch the probe/ini-lookup logic reviewed in round 1.

**3. Independent full re-read of all 8 scripts (current content, not just diff)**

Reread `nnstreamer_{converter_python3,decoder_python3,filter_lua,filter_python3,flatbuf,flexbuf,protobuf,sparse}/runTest.sh` end-to-end as they stand now:

- All 7 non-`filter_python3` scripts share the same restructured shape: build-tree probe → else read `/etc/nnstreamer.ini` → single `if [[ ! -d ${ini_path} ]] || [[ -z $(ls ... | grep <lib>.so) ]]` check → tests. No leftover unguarded `../../build` reference remains in any of them after the guard — the only two that had build-tree-only side effects past the probe (`converter_python3`, `decoder_python3`'s symlink/`pushd` block) are now wrapped in `if [[ -d ${PATH_TO_PLUGIN} ]]`, matching the pattern `filter_python3` already used.
- `filter_python3` is otherwise untouched, still has its own pre-existing install-mode check (`python3 -c "import nnstreamer_python"`) and its own pre-existing `if [[ -d ../../build ]]` guard around the symlink block; only its `report exit` → `report; exit` typo fix is new here, consistent with the same fix applied to the other 6.
- Ini key selection (`filters`/`converters`/`decoders`) matches the sub-plugin each suite actually exercises, including the `protobuf` special case (`extra/` in the build-tree probe vs. `converters` key when installed, since the installed check targets the actual sub-plugin, not the helper lib).
- Checked `nnstreamer.ini.in`: `filters=`, `decoders=`, `converters=` each appear as a unique line, and `^filters`/`^decoders`/`^converters` anchoring doesn't collide with adjacent keys like `customfilters=` (doesn't start with "filters"), so the `grep "^..."` lookups resolve unambiguously.
- Tail-end `rm *.log ...` globs in `flatbuf`/`flexbuf`/`protobuf`/`sparse`/`filter_lua` are pre-existing patterns untouched by this PR, not a regression here.

**4. Remaining blocking issues / merge readiness**

None found. All CI checks on the current head (`19e88c70`) are green. This addresses the only issue raised in round 1, doesn't regress the build-tree path, and the installed-mode branch (still uncovered by CI, as the PR body explains and round 1 already flagged as a pre-existing project-level gap) is exercised correctly per the manual verification described. Mergeable as-is.

myungjoo:
The correction is right and I verified it before acting on it. My description claimed the diff was confined to the installed path; it was not.

Checked the merge base for the arm taken when `../../build` exists but the sub-plugin directory does not:

```
converter_python3 : report exit | fi      <- one line, no exit executed
decoder_python3   : report | exit         <- correct
filter_lua        : report exit | fi
flatbuf           : fi | else             <- message only, nothing after it
flexbuf           : fi | else
protobuf          : fi | else
sparse            : report exit | fi
```

Six of seven fell through and ran the tests; `decoder_python3` was the sole exception, exactly as you found. So the build-tree arm did change, and the reason no CI result moves is that those directories always exist in a CI build — not that the branch is unreachable. The description now says that, and also records the consequence you flagged: a build with the converter sub-plugins disabled will now quietly skip `flatbuf`/`flexbuf`/`protobuf` instead of failing loudly. Consistent with every other suite, but a real behavioural change worth stating rather than burying.

**Note 2 — CI coverage: filed as #4903.** It takes your suggested shape (install with `-Dinstall-test=true`, run one suite from the installed path) and adds the detail that makes it worth doing: the job has to assert the suite *ran* — `N passed among N cases` — rather than merely exited zero, since a skip and a pass are both green today and telling them apart is the entire point. Cross-referenced with #4888.

**Note 1 — `ls | grep <name>.so`: leaving it.** Agreed on both halves: `grep lua.so` is an unanchored regex where `.` matches any character, and keeping it is right for this PR because it leaves the build-tree path byte-identical. Changing the matching style at the same time as changing the control flow would have made the "build-tree behaviour is unchanged" claim harder to check, which is the claim that matters most here.

Thanks for running the five cases against the merge base as well as this head. Comparing the two is what surfaced the row I had wrong — testing only the new code would have shown all seven skipping cleanly and looked entirely correct.

PR #4901, opened 2026-08-28
[Tizen] deprecated API - sensor

jaeyun-jung:
Tizen sensor API will remove deprecated API and enum values.
So, remove deprecated items from tizen-sensor element and test code.


again4you (approved):
LGTM!

Issue #4902, opened 2026-08-28
[CI/Release] Cover 24.04 and 26.04 with pdebuild, and publish 26.04 to the PPA


myungjoo:
Follow-up to #4848, to be picked up once the PRs referenced below have landed.

## Why

CI verifies exactly one Ubuntu release through `pdebuild`, and it is not one of
the releases the PPA actually publishes:

| release | pdebuild in CI | published in PPA |
|---|---|---|
| focal 20.04 | no | yes |
| jammy 22.04 | **yes** | no (stalled) |
| noble 24.04 | no | yes |
| plucky 25.04 | no | no |
| resolute 26.04 | no | no |

`.github/workflows/pdebuild.yml` and `.github/workflows/update_pbuilder_cache.yml`
both pin `distroname: jammy`, and `jammy` is the only value of that key anywhere in
the tree. So the Debian packaging path that actually ships to users is not exercised
by CI at all, which is the structural reason #4848 and the jammy PPA stall both went
unnoticed for months.

## Task 1 — add 24.04 and 26.04 to the pdebuild matrix, build only

The matrix is already shaped for several entries; it just has one:

```yaml
    strategy:
      matrix:
        os: [ ubuntu-22.04 ]
        arch: [ amd64 ]
        include:
          - distroname: jammy
            os: ubuntu-22.04
```

Build verification only — no unit tests or SSAT — to keep the added CI time down.
`DEB_BUILD_OPTIONS=nocheck` should make `dh` skip `override_dh_auto_test`, which is
worth confirming rather than assuming, since that override also runs SSAT.

Notes:

- `update_pbuilder_cache.yml` carries the same matrix and needs the same entries, or
  the new jobs will build their pbuilder base from scratch every run.
- Current jammy pdebuild takes about 12 minutes, so expect the job set to grow roughly
  proportionally.
- **24.04 can be done immediately.** The PPA already carries the full noble build-dep
  stack.
- **26.04 cannot, until Task 2 is done.** `debian/rules` hard-enables edgetpu, openvino
  and onnxruntime for Ubuntu and runs `ssat` in `override_dh_auto_test`, so the build
  dependencies are mandatory, and none of them are published for resolute yet. Ordering
  is therefore Task 1 (24.04) → Task 2 → Task 1 (26.04).

## Task 2 — publish 26.04 (resolute) to the PPA

The repository side is ready. Building nnstreamer on Ubuntu 26.04 with gcc 15.2 and
[...]

PR #4904, opened 2026-08-28
[Build] Keep the meson floor at a version Ubuntu 22.04 ships

myungjoo:
## Problem

The PPA has published nothing for jammy since 2025-12-04. The Launchpad recipe
build has been parked since 2025-12-19:

```
Build status: Dependency wait
Missing build dependencies: meson (>= 0.62.0)
Series: Jammy
```

647b30e5 raised the minimum meson to 0.62.0. Ubuntu 22.04 ships 0.61.2:

| series | archive meson | in the nnstreamer PPA |
|---|---|---|
| focal 20.04 | 0.53.2 | **1.6.1** — this is why focal still builds |
| jammy 22.04 | 0.61.2 | none |
| noble 24.04 | 1.3.2 | — |
| plucky 25.04 | 1.7.0 | — |
| resolute 26.04 | 1.10.1 | — |

CI did not catch it because the same commit added
`ppa:kubuntu-ppa/backports-extra` (meson 0.63.2) to the pdebuild chroot. The
Launchpad recipe has no such archive, so CI went green while the PPA silently
stopped.

## What actually needs 0.62.0

One construct: `ExternalProgram.version()`, used once for the flatc version.
Probed on jammy's meson directly:

```
meson.build:4:2: ERROR: Unknown method "version" in object
<[ExternalProgramHolder] holds [ExternalProgram]: <ExternalProgram 'flatc' -> ['/usr/bin/flatc']>>
```

Everything else 647b30e5 introduced needs at most 0.56.0 — the `fs` module
(0.53), `ExternalProgram.full_path()` (0.55), `meson.project_source_root()` and
`meson.project_build_root()` (0.56). The floor lands a step above that at
0.57.0, because of something 647b30e5 never touched; see the verification
section.

## Change

Read the flatc version through `run_command` again, as this file did before
647b30e5, and declare 0.57.0. **The version check itself is unchanged**:
libflatbuffers must still match the installed flatc exactly, and must still be
at least 2.0.0. Only the call used to obtain the string differs. Confirmed
live on jammy:

```
meson 0.61.2  /  flatc version 1.12.0
Message: flatbuffers version 1.12.0 is required because flatc (flatbuf-compiler) 1.12.0 is installed
Run-time dependency flatbuffers found: NO
```

That is the check correctly rejecting jammy's libflatbuffers 2.0.0 against
flatc 1.12.0, so it is demonstrably still doing its job. None of the other
modernisation from 647b30e5 is touched.

## Verification, and how the floor was corrected

The floor first proposed here was 0.56.0, derived from
`meson.project_source_root()`, and sampling meson 0.56.0, 0.58.2, 0.60.3 and
0.61.2 configured cleanly under each. That sampling was wrong, and the check
below caught it on its first CI run:

```
meson.build:342: WARNING: Project targets '>= 0.56.0' but uses feature
introduced in '0.57.0': compiler method "dependencies" kwarg with internal dep.
```

That is the LiteRT sanity link, `cxx.links(..., dependencies: litert_dep)`, and
it sits behind `if litert_dep.found()`. A configure without the LiteRT SDK never
visits it, so every one of those samples missed it; the required Valgrind job
installs LiteRT and does visit it. The floor is 0.57.0, declared across all
eight sites. jammy carries 0.61.2, so the point of coming down from 0.62.0
stands.

The lesson is in the method: asking meson to enumerate what exceeds the floor
beats sampling versions that happen to configure, because the option set decides
which code paths a configure visits at all. meson 0.57.0 configures the tree
with zero feature warnings.

Full build with tests on jammy stock meson, no backports archive:

```
Ubuntu 22.04.4 LTS / gcc 11.4.0 / meson 0.61.2 / flatc 1.12.0
[256/256] Linking target tests/unittest_sink
=== NINJA OK ===
```

## Consistency

`debian/control.ubuntu.ppa` and `debian/control.debian` were still at
`meson (>=0.49)` while `meson.build` required 0.62.0. That understatement
reaches users: `debian/rules` copies `control.ubuntu.ppa` over `debian/control`
during `dh_clean`, before the source package is generated, so it is the
`ubuntu.ppa` value that ends up in the published `.dsc` and gates the binary
build. From the current noble publication:

```
$ curl -s .../nnstreamer_2.7.0.0-0~202512190802~ubuntu24.04.1.dsc
Build-Depends: ... ninja-build, meson (>= 0.49), ...
```

So the source build and the binary build were resolving against different
requirements. All eight sites now say 0.57.0: `meson.build`, the three control
files, `packaging/nnstreamer.spec`, `AGENTS.md`, and two Documentation pages.

## Regression test

Two guards, and one of them is weaker than it looks.

**The floor itself.** The two CI workarounds are removed, so the jobs that
build against a real jammy toolchain do what the Launchpad recipe does:
`pdebuild.yml` drops the `kubuntu-ppa/backports-extra` mirror, and `risc-v.yml`
installs the distro `meson` package again instead of `pip3 install meson`.
Reintroducing a construct newer than 0.61.2 makes both fail.

Neither of those is a required status check, though. Required on `main` are
DCO, the three Tizen GBS jobs, MacOS, Android NDK, LLVM/Clang, Valgrind on
ubuntu-22.04, and Static checks; of those, the ubuntu-22.04 jobs run `pip
install meson ninja`, which fetches PyPI's newest regardless of the runner's
release, and MacOS uses `brew install meson`. A red pdebuild alone would not
block a merge.

So the required Valgrind job now carries the check directly. The newest meson
is precisely the one that knows every feature's introduction version, and it
reports the problem — as a warning, which is why nobody saw it:

```
meson.build:174: WARNING: Project targets '>= 0.57.0' but uses feature
introduced in '0.62.0': Program.version.
```

The job reads that out of `build/meson-logs/meson-log.txt` after the configure
it already runs, so no existing step changes. Measured with the newest meson:
no match and a pass on this tree, one match and a failure with
`ExternalProgram.version()` put back. `--fatal-meson-warnings` would be tidier
but is unusable — the tree already emits two unrelated warnings about a
`>= 0.52.0` conditional in `tools/development/parser/meson.build` that the
floor has made constant.

Making pdebuild and risc-v required is still a branch-protection setting and
still worth doing, but the floor no longer depends on it.

**The declarations.** `meson-version.sh` takes `meson.build` as the authority
and requires the other seven sites to match. It rejects the tree as it stood
before the second commit here and accepts it after, and it runs under Static
checks, which *is* required. This is the guard for the failure that actually
happened twice: 647b30e5 left two control files at 0.49, and the first commit
of this PR left three documents at 0.62.0.

There is no unit or SSAT test, because nothing in the built artifacts changes.
The constraint is a build-time dependency version, so a checker over the
declarations and CI running the real toolchain are what can detect it.

## Scope

This unblocks jammy only. It does not affect focal, noble, or the newer series,
which all have meson well above the floor. #4902 tracks the remaining release
work.

### DIFF
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -83,7 +83,7 @@ jobs:
         sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
         echo "::endgroup::"
         echo "DISTRIBUTION=${{ matrix.distroname }}" > ~/.pbuilderrc
-        echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main |deb [trusted=yes] http://ppa.launchpad.net/kubuntu-ppa/backports-extra/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
+        echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
         cat ~/.pbuilderrc
         sudo mkdir -p /root/
         sudo ln -sf ~/.pbuilderrc /root/
--- a/.github/workflows/risc-v.yml
+++ b/.github/workflows/risc-v.yml
@@ -26,8 +26,7 @@ jobs:
           githubToken: ${{ github.token }}
           run: |
             apt-get -qy update
-            apt-get -qy install ninja-build libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libglib2.0-dev libjson-glib-dev gstreamer1.0-tools gstreamer1.0-plugins-base gstreamer1.0-plugins-good libgtest-dev libpng-dev libopencv-dev python3 python3-dev python3-numpy python3-pip pkg-config gcc g++ liblua5.1-dev bash wget libpng-dev
-            pip3 install meson
+            apt-get -qy install meson ninja-build libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libglib2.0-dev libjson-glib-dev gstreamer1.0-tools gstreamer1.0-plugins-base gstreamer1.0-plugins-good libgtest-dev libpng-dev libopencv-dev python3 python3-dev python3-numpy pkg-config gcc g++ liblua5.1-dev bash wget libpng-dev
             bash .github/workflows/get_ssat.sh
             meson setup build -Denable-tizen=false -Denable-test=true
             ninja -C build
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -18,7 +18,10 @@
 #
 # Every tracked document and packaging file is scanned rather than a list of
 # the sites that happen to declare it today, so a ninth one cannot appear
-# unnoticed - which is the whole point of the check.
+# unnoticed - which is the whole point of the check. Tracked, via git, so a
+# build directory or a vendored copy left in the tree cannot fail the check on
+# a file nobody owns; find is the fallback for a directory that is not a
+# repository, which is how the self-test drives this.
 #
 # Argument ($1): repository root to check. Defaults to the repository this
 #                script lives in.
@@ -109,8 +112,12 @@ while IFS= read -r rel; do
 done < <(
   cd "$ROOT" || exit 1
   {
-    find . -name .git -prune -o -type f \( -name '*.md' -o -name '*.spec' \) -print
-    ls -1 debian/control* 2>/dev/null
+    if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then
+      git ls-files -- '*.md' '*.spec' 'debian/control*'
+    else
+      find . -name .git -prune -o -type f \( -name '*.md' -o -name '*.spec' \) -print
+      ls -1 debian/control* 2>/dev/null
+    fi
   } | sed "s|^\./||" | sort -u
 )
 
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -26,13 +26,17 @@ SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 ROOT=${1:-$(cd "${SCRIPT_DIR}/../../.." && pwd)}
 failed=0
 
-# A declared requirement: meson, then >= with nothing but an optional bracket
-# between them, then a version. Every form in the tree fits - "meson (>=X)",
-# "meson >= X", "Meson >= X + Ninja". Keeping the gap this tight is what stops
-# prose such as "meson.build wants glib >= 2.60" from reading as a meson
-# requirement. Because the gap cannot contain a digit, the only digits in a
-# match belong to the version itself.
-DECL_RE='[Mm]eson[[:space:]]*\(?[[:space:]]*>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+# A declared requirement: meson, then >=, then a version. The gap between them
+# may hold punctuation but no letter or digit, which is what separates a
+# declaration from prose: "meson (>=X)", "**meson** >= X" and "`meson` >= X"
+# all match, while "meson.build wants glib >= 2.60" does not, because another
+# package name stands in the way. Excluding digits from the gap also means the
+# only digits in a match belong to the version itself.
+#
+# Known limit: a markdown link, "[meson](https://example.com) >= X", puts
+# letters in the gap and is not recognised. Write the requirement outside the
+# link text.
+DECL_RE='meson(_version)?[^[:alnum:]]{0,12}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
 
 ##
 # @brief Print a version padded to four components so 0.56 and 0.56.0 compare equal.
@@ -87,8 +91,8 @@ check_file() {
         echo "::error::${rel}:${lineno} requires meson ${found_raw}, but meson.build requires ${expected_raw}. Keep every declaration in step."
         failed=1
       fi
-    done < <(echo "$text" | grep -oE "$DECL_RE")
-  done < <(grep -nE "$DECL_RE" "$file")
+    done < <(echo "$text" | grep -oiE "$DECL_RE")
+  done < <(grep -niE "$DECL_RE" "$file")
 }
 
 while IFS= read -r rel; do
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -26,17 +26,21 @@ SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 ROOT=${1:-$(cd "${SCRIPT_DIR}/../../.." && pwd)}
 failed=0
 
-# A declared requirement: meson, then >=, then a version. The gap between them
-# may hold punctuation but no letter or digit, which is what separates a
-# declaration from prose: "meson (>=X)", "**meson** >= X" and "`meson` >= X"
-# all match, while "meson.build wants glib >= 2.60" does not, because another
-# package name stands in the way. Excluding digits from the gap also means the
-# only digits in a match belong to the version itself.
+# A declared requirement: meson, then >=, then a version, with a gap between
+# them that may hold decoration but nothing that means the sentence moved on.
+# Letters and digits are excluded, so a named neighbour breaks the match -
+# "meson.build wants glib >= 2.60" is prose about glib, not a meson
+# requirement - and so the only digits in a match belong to the version.
+# Comma, period and semicolon are excluded too, so an unnamed version further
+# along a sentence, as in "see meson, (>=5.0) elsewhere", is not attributed to
+# meson either. What is left through is decoration: "meson (>=X)",
+# "**meson** >= X", "`meson` >= X", "meson_version: >=X".
 #
-# Known limit: a markdown link, "[meson](https://example.com) >= X", puts
-# letters in the gap and is not recognised. Write the requirement outside the
-# link text.
-DECL_RE='meson(_version)?[^[:alnum:]]{0,12}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+# This is a heuristic over prose and is deliberately not grown any further. A
+# markdown link, "[meson](https://example.com) >= X", puts letters in the gap
+# and is not recognised. If some other phrasing escapes it, write the
+# requirement in one of the plain forms above rather than widening this.
+DECL_RE='meson(_version)?[^[:alnum:],.;]{0,12}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
 
 ##
 # @brief Print a version padded to four components so 0.56 and 0.56.0 compare equal.
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -16,6 +16,10 @@
 # the source build resolves against. meson.build is the authority here,
 # because it is the file meson itself enforces.
 #
+# Every tracked document and packaging file is scanned rather than a list of
+# the sites that happen to declare it today, so a ninth one cannot appear
+# unnoticed - which is the whole point of the check.
+#
 # Argument ($1): repository root to check. Defaults to the repository this
 #                script lives in.
 #
@@ -58,7 +62,7 @@ if [ ! -f "${ROOT}/meson.build" ]; then
   exit 1
 fi
 
-expected_raw=$(grep -oE "meson_version:[[:space:]]*'>=[0-9]+(\.[0-9]+)+'" "${ROOT}/meson.build" \
+expected_raw=$(grep -oE "meson_version:[[:space:]]*'>=[[:space:]]*[0-9]+(\.[0-9]+)+'" "${ROOT}/meson.build" \
   | head -1 | grep -oE '[0-9]+(\.[0-9]+)+')
 if [ -z "$expected_raw" ]; then
   echo "::error::meson.build does not declare meson_version."
@@ -104,8 +108,10 @@ while IFS= read -r rel; do
   check_file "$rel"
 done < <(
   cd "$ROOT" || exit 1
-  ls -1 AGENTS.md packaging/nnstreamer.spec debian/control* 2>/dev/null
-  find Documentation -maxdepth 1 -type f -name '*.md' 2>/dev/null | sort
+  {
+    find . -name .git -prune -o -type f \( -name '*.md' -o -name '*.spec' \) -print
+    ls -1 debian/control* 2>/dev/null
+  } | sed "s|^\./||" | sort -u
 )
 
 if [ "$failed" -ne 0 ]; then
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -43,11 +43,16 @@ failed=0
 # meson either. What is left through is decoration: "meson (>=X)",
 # "**meson** >= X", "`meson` >= X", "meson_version: >=X".
 #
-# This is a heuristic over prose and is deliberately not grown any further. A
-# markdown link, "[meson](https://example.com) >= X", puts letters in the gap
-# and is not recognised. If some other phrasing escapes it, write the
-# requirement in one of the plain forms above rather than widening this.
-DECL_RE='meson(_version)?[^[:alnum:],.;]{0,12}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+# The gap is three characters because that is what the forms above need: one
+# for a space, two for a backtick or a bracket, three for "**meson** >=".
+# Anything longer is a sentence that has wandered, as in "see meson -- (>=9.0)
+# elsewhere", where the version belongs to something else entirely.
+#
+# This is a heuristic over prose and is deliberately not grown. A markdown
+# link, "[meson](https://example.com) >= X", puts letters in the gap and is
+# not recognised. If some other phrasing escapes it, write the requirement in
+# one of the plain forms above rather than widening this.
+DECL_RE='meson(_version)?[^[:alnum:],.;]{0,3}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
 
 ##
 # @brief Print a version padded to four components so 0.56 and 0.56.0 compare equal.
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file     meson-version.sh
+# @brief    Check that the minimum meson version is declared consistently.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# The minimum meson is written down in eight places: meson.build, the three
+# debian/control* files, packaging/nnstreamer.spec, AGENTS.md and two
+# Documentation pages. Nothing tied them together, so raising the floor in
+# 647b30e5 left debian/control.ubuntu.ppa and debian/control.debian behind at
+# 0.49 and the value that reaches the published .dsc disagreed with the value
+# the source build resolves against. meson.build is the authority here,
+# because it is the file meson itself enforces.
+#
+# Argument ($1): repository root to check. Defaults to the repository this
+#                script lives in.
+#
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+ROOT=${1:-$(cd "${SCRIPT_DIR}/../../.." && pwd)}
+failed=0
+
+# A declared requirement: the word meson, then >=, then a version. Plain
+# mentions of a version some distro happens to ship carry no >= and are left
+# alone.
+DECL_RE='[Mm]eson[^0-9]{0,24}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+
+##
+# @brief Print a version as major.minor.patch so 0.56 and 0.56.0 compare equal.
+# @param $1 version string
+normalize() {
+  echo "$1" | awk -F. '{ printf "%d.%d.%d\n", $1, ($2 == "" ? 0 : $2), ($3 == "" ? 0 : $3) }'
+}
+
+if [ ! -f "${ROOT}/meson.build" ]; then
+  echo "::error::${ROOT}/meson.build not found."
+  exit 1
+fi
+
+expected_raw=$(grep -oE "meson_version:[[:space:]]*'>=[0-9]+(\.[0-9]+)+'" "${ROOT}/meson.build" \
+  | head -1 | grep -oE '[0-9]+(\.[0-9]+)+')
+if [ -z "$expected_raw" ]; then
+  echo "::error::meson.build does not declare meson_version."
+  exit 1
+fi
+expected=$(normalize "$expected_raw")
+echo "meson.build declares meson_version >= ${expected_raw}"
+
+targets=$(cd "$ROOT" && ls -1 AGENTS.md packaging/nnstreamer.spec debian/control* 2>/dev/null; \
+          cd "$ROOT" && find Documentation -maxdepth 1 -type f -name '*.md' 2>/dev/null | sort)
+
+for rel in $targets; do
+  file="${ROOT}/${rel}"
+  [ -f "$file" ] || continue
+  while IFS= read -r hit; do
+    [ -z "$hit" ] && continue
+    lineno=${hit%%:*}
+    found_raw=$(echo "${hit#*:}" | grep -oE '[0-9]+(\.[0-9]+)+' | tail -1)
+    found=$(normalize "$found_raw")
+    if [ "$found" = "$expected" ]; then
+      echo "  ok   ${rel}:${lineno} (${found_raw})"
+    else
+      echo "::error::${rel}:${lineno} requires meson ${found_raw}, but meson.build requires ${expected_raw}. Keep every declaration in step."
+      failed=1
+    fi
+  done < <(grep -nE "$DECL_RE" "$file" | grep -oE "^[0-9]+:.*")
+done
+
+if [ "$failed" -ne 0 ]; then
+  echo "::error::The meson version declarations disagree."
+  exit 1
+fi
+
+echo "The meson version declarations agree."
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -12,7 +12,7 @@
 # debian/control* files, packaging/nnstreamer.spec, AGENTS.md and two
 # Documentation pages. Nothing tied them together, so raising the floor in
 # 647b30e5 left debian/control.ubuntu.ppa and debian/control.debian behind at
-# 0.49 and the value that reaches the published .dsc disagreed with the value
+# 0.49, and the value that reaches the published .dsc disagreed with the value
 # the source build resolves against. meson.build is the authority here,
 # because it is the file meson itself enforces.
 #
@@ -26,16 +26,23 @@ SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 ROOT=${1:-$(cd "${SCRIPT_DIR}/../../.." && pwd)}
 failed=0
 
-# A declared requirement: the word meson, then >=, then a version. Plain
-# mentions of a version some distro happens to ship carry no >= and are left
-# alone.
-DECL_RE='[Mm]eson[^0-9]{0,24}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+# A declared requirement: meson, then >= with nothing but an optional bracket
+# between them, then a version. Every form in the tree fits - "meson (>=X)",
+# "meson >= X", "Meson >= X + Ninja". Keeping the gap this tight is what stops
+# prose such as "meson.build wants glib >= 2.60" from reading as a meson
+# requirement. Because the gap cannot contain a digit, the only digits in a
+# match belong to the version itself.
+DECL_RE='[Mm]eson[[:space:]]*\(?[[:space:]]*>=[[:space:]]*[0-9]+(\.[0-9]+)+'
 
 ##
-# @brief Print a version as major.minor.patch so 0.56 and 0.56.0 compare equal.
+# @brief Print a version padded to four components so 0.56 and 0.56.0 compare equal.
 # @param $1 version string
 normalize() {
-  echo "$1" | awk -F. '{ printf "%d.%d.%d\n", $1, ($2 == "" ? 0 : $2), ($3 == "" ? 0 : $3) }'
+  echo "$1" | awk -F. '{
+    for (i = 1; i <= 4; i++) {
+      printf "%d%s", ($i == "" ? 0 : $i), (i < 4 ? "." : "\n")
+    }
+  }'
 }
 
 if [ ! -f "${ROOT}/meson.build" ]; then
@@ -52,25 +59,46 @@ fi
 expected=$(normalize "$expected_raw")
 echo "meson.build declares meson_version >= ${expected_raw}"
 
-targets=$(cd "$ROOT" && ls -1 AGENTS.md packaging/nnstreamer.spec debian/control* 2>/dev/null; \
-          cd "$ROOT" && find Documentation -maxdepth 1 -type f -name '*.md' 2>/dev/null | sort)
+##
+# @brief Compare every declaration in one file against meson.build.
+# @param $1 path of the file relative to the repository root
+check_file() {
+  local rel=$1
+  local file="${ROOT}/${rel}"
+  local hit lineno text decl found_raw found
 
-for rel in $targets; do
-  file="${ROOT}/${rel}"
-  [ -f "$file" ] || continue
+  [ -f "$file" ] || return 0
+
   while IFS= read -r hit; do
     [ -z "$hit" ] && continue
     lineno=${hit%%:*}
-    found_raw=$(echo "${hit#*:}" | grep -oE '[0-9]+(\.[0-9]+)+' | tail -1)
-    found=$(normalize "$found_raw")
-    if [ "$found" = "$expected" ]; then
-      echo "  ok   ${rel}:${lineno} (${found_raw})"
-    else
-      echo "::error::${rel}:${lineno} requires meson ${found_raw}, but meson.build requires ${expected_raw}. Keep every declaration in step."
-      failed=1
-    fi
-  done < <(grep -nE "$DECL_RE" "$file" | grep -oE "^[0-9]+:.*")
-done
+    text=${hit#*:}
+    # One line may carry more than one declaration, and the version has to be
+    # read out of the matched text rather than the whole line: a neighbouring
+    # constraint such as "debhelper (>=9.20)" would otherwise be picked up as
+    # the meson requirement.
+    while IFS= read -r decl; do
+      [ -z "$decl" ] && continue
+      found_raw=$(echo "$decl" | grep -oE '[0-9]+(\.[0-9]+)+' | tail -1)
+      found=$(normalize "$found_raw")
+      if [ "$found" = "$expected" ]; then
+        echo "  ok   ${rel}:${lineno} (${found_raw})"
+      else
+        echo "::error::${rel}:${lineno} requires meson ${found_raw}, but meson.build requires ${expected_raw}. Keep every declaration in step."
+        failed=1
+      fi
+    done < <(echo "$text" | grep -oE "$DECL_RE")
+  done < <(grep -nE "$DECL_RE" "$file")
+}
+
+while IFS= read -r rel; do
+  [ -z "$rel" ] && continue
+  check_file "$rel"
+done < <(
+  cd "$ROOT" || exit 1
+  ls -1 AGENTS.md packaging/nnstreamer.spec debian/control* 2>/dev/null
+  find Documentation -maxdepth 1 -type f -name '*.md' 2>/dev/null | sort
+)
 
 if [ "$failed" -ne 0 ]; then
   echo "::error::The meson version declarations disagree."
--- a/.github/workflows/static.check.scripts/meson-version.sh
+++ b/.github/workflows/static.check.scripts/meson-version.sh
@@ -43,16 +43,24 @@ failed=0
 # meson either. What is left through is decoration: "meson (>=X)",
 # "**meson** >= X", "`meson` >= X", "meson_version: >=X".
 #
-# The gap is three characters because that is what the forms above need: one
-# for a space, two for a backtick or a bracket, three for "**meson** >=".
-# Anything longer is a sentence that has wandered, as in "see meson -- (>=9.0)
-# elsewhere", where the version belongs to something else entirely.
+# These are the accepted forms, and the gap of four characters is what the
+# longest of them needs:
+#
+#   meson >= X            meson (>= X)          **meson** >= X
+#   `meson` >= X          `meson` (>= X)        **meson** (>= X)
+#   meson_version: >= X   meson_version : >= X
+#
+# Comma, period, semicolon, question mark and exclamation mark are barred from
+# the gap because they end a clause: in "see meson -- (>=9.0) elsewhere" or
+# "really, meson?! (>=5.0)" the version belongs to something else. Letters and
+# digits are barred too, so a named neighbour such as "glib >= 2.60" breaks the
+# match, and the only digits left inside one belong to the version.
 #
 # This is a heuristic over prose and is deliberately not grown. A markdown
 # link, "[meson](https://example.com) >= X", puts letters in the gap and is
 # not recognised. If some other phrasing escapes it, write the requirement in
-# one of the plain forms above rather than widening this.
-DECL_RE='meson(_version)?[^[:alnum:],.;]{0,3}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
+# one of the forms listed above rather than widening this.
+DECL_RE='meson(_version[[:space:]]*:)?[^[:alnum:],.;?!]{0,4}>=[[:space:]]*[0-9]+(\.[0-9]+)+'
 
 ##
 # @brief Print a version padded to four components so 0.56 and 0.56.0 compare equal.
@@ -70,7 +78,7 @@ if [ ! -f "${ROOT}/meson.build" ]; then
   exit 1
 fi
 
-expected_raw=$(grep -oE "meson_version:[[:space:]]*'>=[[:space:]]*[0-9]+(\.[0-9]+)+'" "${ROOT}/meson.build" \
+expected_raw=$(grep -oE "meson_version[[:space:]]*:[[:space:]]*'>=[[:space:]]*[0-9]+(\.[0-9]+)+'" "${ROOT}/meson.build" \
   | head -1 | grep -oE '[0-9]+(\.[0-9]+)+')
 if [ -z "$expected_raw" ]; then
   echo "::error::meson.build does not declare meson_version."
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -121,6 +121,14 @@ expect_checker 1 "meson.build syntax quoted in a page is checked too" \
 "${MESON_BUILD}
 $(printf 'Documentation/getting-started-meson-build.md\tThe project declares meson_version: >=0.62.0 today.')"
 
+expect_checker 0 "a version further along a sentence is not attributed to meson" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tSee meson, (>=5.0) is the tool version we mean elsewhere.')"
+
+expect_checker 0 "an ellipsis breaks the association as well" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tAfter installing meson... (>=2024.1) support was added.')"
+
 expect_checker 1 "two declarations on one line are both checked" \
 "${MESON_BUILD}
 $(printf 'Documentation/how-to-run-examples.md\tUse meson >= 0.56.0 on Ubuntu and Meson >= 0.62.0 on Tizen.')"
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -141,6 +141,18 @@ expect_checker 0 "punctuation between the two does not carry a version either" \
 "${MESON_BUILD}
 $(printf 'Documentation/getting-started-meson-build.md\tReally, meson?! (>=5.0) is unrelated.')"
 
+expect_checker 1 "a bracketed bold form is recognised" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tBuild with **meson** (>= 0.62.0).')"
+
+expect_checker 0 "a space before the colon still yields the authority" \
+"$(printf 'meson.build\tproject(%s, %s, meson_version : %s)' "'nnstreamer'" "'c'" "'>=0.57.0'")
+$(printf 'AGENTS.md\tBuild system: **Meson >= 0.57.0 + Ninja**.')"
+
+expect_checker 1 "meson_version with a spaced colon is checked in a page" \
+"${MESON_BUILD}
+$(printf 'Documentation/how-to-run-examples.md\tIt declares meson_version : >=0.62.0 today.')"
+
 expect_checker 1 "two declarations on one line are both checked" \
 "${MESON_BUILD}
 $(printf 'Documentation/how-to-run-examples.md\tUse meson >= 0.56.0 on Ubuntu and Meson >= 0.62.0 on Tizen.')"
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -103,6 +103,22 @@ expect_checker 0 "another package's requirement near the word meson is ignored" 
 "${MESON_BUILD}
 $(printf 'Documentation/getting-started-meson-build.md\tmeson.build additionally wants glib >= 2.60 at configure time.')"
 
+expect_checker 1 "markdown emphasis does not hide a declaration" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tBuild system: **meson** >= 0.62.0 and ninja.')"
+
+expect_checker 1 "a code span does not hide a declaration" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tYou need `meson` >= 0.62.0 to configure.')"
+
+expect_checker 1 "an upper case spelling does not hide a declaration" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tMESON >= 0.62.0 is required.')"
+
+expect_checker 1 "meson.build syntax quoted in a page is checked too" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tThe project declares meson_version: >=0.62.0 today.')"
+
 expect_checker 1 "two declarations on one line are both checked" \
 "${MESON_BUILD}
 $(printf 'Documentation/how-to-run-examples.md\tUse meson >= 0.56.0 on Ubuntu and Meson >= 0.62.0 on Tizen.')"
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -133,6 +133,18 @@ expect_checker 1 "two declarations on one line are both checked" \
 "${MESON_BUILD}
 $(printf 'Documentation/how-to-run-examples.md\tUse meson >= 0.56.0 on Ubuntu and Meson >= 0.62.0 on Tizen.')"
 
+expect_checker 0 "a spaced >= in meson.build is still read as the authority" \
+"$(printf 'meson.build\tproject(%s, %s, meson_version: %s)' "'nnstreamer'" "'c'" "'>= 0.56.0'")
+$(printf 'AGENTS.md\tBuild system: **Meson >= 0.56.0 + Ninja**.')"
+
+expect_checker 1 "a declaration in a file outside the old list is caught" \
+"${MESON_BUILD}
+$(printf 'README.md\tnnstreamer builds with meson >= 0.62.0.')"
+
+expect_checker 1 "a declaration in a Documentation subdirectory is caught" \
+"${MESON_BUILD}
+$(printf 'Documentation/tutorials/tutorial1.md\tInstall meson >= 0.62.0 first.')"
+
 # A path with a space must not be skipped by word splitting.
 spaced_root="${workdir}/spaced"
 mkdir -p "${spaced_root}/Documentation"
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -129,6 +129,18 @@ expect_checker 0 "an ellipsis breaks the association as well" \
 "${MESON_BUILD}
 $(printf 'Documentation/getting-started-meson-build.md\tAfter installing meson... (>=2024.1) support was added.')"
 
+expect_checker 0 "a bracketed aside does not carry a version to meson" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tSee meson [*] (>=5.0) for the other tool.')"
+
+expect_checker 0 "a dashed aside does not carry a version to meson" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tSee meson -- (>=9.0) for the other tool.')"
+
+expect_checker 0 "punctuation between the two does not carry a version either" \
+"${MESON_BUILD}
+$(printf 'Documentation/getting-started-meson-build.md\tReally, meson?! (>=5.0) is unrelated.')"
+
 expect_checker 1 "two declarations on one line are both checked" \
 "${MESON_BUILD}
 $(printf 'Documentation/how-to-run-examples.md\tUse meson >= 0.56.0 on Ubuntu and Meson >= 0.62.0 on Tizen.')"
--- a/.github/workflows/static.check.scripts/test_meson_version.sh
+++ b/.github/workflows/static.check.scripts/test_meson_version.sh
@@ -14,7 +14,10 @@
 # a requirement. Without the first the checker could accept any drift; without
 # the last it would fail on prose that is simply describing a release. The
 # repository's own tree is checked at the end so the shipped files stay
-# covered when no fixture changes.
+# covered when no fixture changes. Later fixtures pin the parts that were
+# got wrong once: the version must come out of the matched text and not the
+# rest of the line, and a path carrying a space must not fall out of the
+# file list.
 #
 # The checker is invoked with stdin closed, matching how GitHub runs it, so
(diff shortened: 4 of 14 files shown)


myungjoo:

I independently reproduced the claims in this PR's description rather than just reading the diff. Summary first, details and severities below.

**Verdict: mergeable.** The core fix is correct and I could not find a functional regression. One **major** finding about the actual strength of the "CI is now the guard" claim, and one **minor** documentation-consistency gap.

## What I verified independently

- **flatc version check is byte-for-byte restored.** `git show 647b30e5^:meson.build` shows the pre-bump line is exactly `flatc_ver = run_command(flatc, '--version', check : true).stdout().split()[2]` — identical to what this PR reintroduces. This isn't a reimplementation, it's a literal revert of that one line, which is the safest possible way to do this.
- **Reproduced the fix on real jammy.** In an `ubuntu:22.04` Docker container (apt meson 0.61.2, apt flatbuffers-compiler 1.12.0, no PPA/backports), `meson setup` on this branch's tree succeeds end-to-end (exit 0) with both `enable-test=false` and `enable-test=true`, and the flatc block prints exactly the message quoted in the PR body: `flatbuffers version 1.12.0 is required because flatc (flatbuf-compiler) 1.12.0 is installed`. I also verified the `flatc.found() == false` path (no flatc installed) configures cleanly. This matches the actual `gh pr checks` result: "Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)" is green on this PR.
- **`split()[2]` robustness against `flatc --version` output.** Checked real `flatc --version` output across five independently-packaged versions spanning 2019–2025: Ubuntu 20.04 apt (1.11.0), Ubuntu 22.04 apt (1.12.0), Ubuntu 24.04 apt (2.0.8), Arch current (25.12.19). All print the stable 3-token form `flatc version X.Y.Z`, so `.split()[2]` is safe across the whole practical version range.
- **The 0.56.0 floor is independently correct, not just asserted.** Grepped every `meson.build` in the tree (`ext/**`, `gst/**`, `tests/**`, `tools/**`, top-level) for version-gated constructs. Findings, cross-checked against meson's own release notes:
  - `meson.project_source_root()` / `project_build_root()` — new in 0.56.0 (confirmed via mesonbuild.com release notes) — this is the tightest constraint in the tree and matches the declared floor exactly.
  - `fs.is_dir()` / `fs.is_file()` (fs module) — module added 0.53.0.
  - `ExternalProgram.full_path()` — 0.55.0.
  - `ExternalProgram.version()` (the construct being removed) — confirmed via the 0.62.0 release notes page ("found programs now have a version method") — this really is the only >0.56.0 construct in the tree.
  - `Dependency.version()` (used elsewhere for tf/tflite/glib/snpe/python3/lua) is a different, much older method on a different object type — not affected by any of this.
  - `subdir_done()` (0.49), `find_program(..., version:)` (0.52), `run_command(..., check:)` (0.47), `option(type: 'feature')` (0.47) — all comfortably under the new floor.
  - No `summary()`, no `install_symlink`, no dict-form `declare_dependency(variables:)`, no `pkgconfig` module usage anywhere in the tree.
  So 0.56.0 is not just "probably fine," it's the exact minimum the current tree needs.
- **PPA/version facts in the PR body, checked against Launchpad's REST API directly (not taken on faith):**
  - `ubuntu/jammy/amd64` archive `meson` = `0.61.2-1` ✓ (matches table)
  - `ubuntu/noble/amd64` archive `meson` = `1.3.2-1ubuntu1` ✓
  - `~nnstreamer/+archive/ubuntu/ppa` has `meson 1.6.1-...ubuntu20.04.1` for focal ✓ (explains why focal alone still builds)
  - Last published `nnstreamer` binary for jammy in the PPA: `2.7.0.0-0~202512040158~ubuntu22.04.1`, published `2025-12-04` ✓ (matches "nothing since 2025-12-04")
  - `debian/rules`'s `override_dh_clean` really does `cp debian/control.ubuntu.ppa debian/control` before packaging on Ubuntu-derived vendors, which is why the stale `meson (>=0.49)` in `control.ubuntu.ppa` (not the top-level `debian/control`) is what reached the published `.dsc`. Confirmed by reading `debian/rules` directly.
- **`kubuntu-ppa/backports-extra` removal is safe.** `git log -p --follow` on `pdebuild.yml` shows that mirror was added in 647b30e5 itself and never existed before — it was single-purpose (get a newer meson) and never supplied any of the exotic `Build-Depends` entries (`tensorflow2-lite-dev`, `openvino-dev`, `onert-dev`, `tvm-runtime-dev`, `onnxruntime-dev`, etc., which all come from `ppa:nnstreamer/ppa`, which is left untouched). Also, the `pdebuild.yml` matrix only has one entry (`jammy`), so this change cannot affect any other distro's pdebuild job.
- **`risc-v.yml`'s `python3-pip` removal is safe.** Grepped `packaging/run_unittests_binaries.sh`, `.github/workflows/get_ssat.sh`, and the python3-filter test path this job exercises — none of them use `pip`. Nothing else in that job depended on it.
- **`packaging/nnstreamer.spec` change is safe for Tizen GBS.** It only lowers a `BuildRequires: meson >= X` floor; a lower minimum can't newly fail a build that satisfied the old, stricter one, and nothing else in the spec is version-sensitive to meson. `gh pr checks` also shows all three "Tizen GBS build on Ubuntu" required checks passing on this PR.

## Findings

### Major — the "CI is now the guard" claim is weaker than the PR body suggests

The PR body's "Regression test" section says dropping the two workarounds "turns CI into the guard" and that "reintroducing a construct newer than 0.61.2 now fails CI instead of passing." That's true in the sense that the job will go red, but I checked `gh api repos/nnstreamer/nnstreamer/branches/main/protection` and neither of the two jobs that actually exercise real jammy meson — **"Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)"** and **"Build on Ubuntu 22.04 RISC-V 64"** — is in `required_status_checks.contexts`. The required list is: `DCO`, the three `Tizen GBS build` jobs, `Build on MacOS-latest with Meson`, `Android NDK build`, `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)`, `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)`, and `Static checks`.

Worse, the two *required* jobs that run on `ubuntu-22.04` (`ubuntu_clean_llvm_build.yml` and `ubuntu_clean_meson_build.yml`) both install meson via `pip install meson ninja`, which pulls PyPI's latest release (currently `1.12.0`, confirmed via `pypi.org/pypi/meson/json`) regardless of the runner's actual OS. They never touch jammy's real 0.61.2 toolchain, so they structurally cannot catch a future reintroduction of a >0.56.0 construct — with or without this PR.

Net effect: after this PR, a future PR that reintroduces something requiring meson >0.56.0 will make `pdebuild.yml`/`risc-v.yml` turn red, but since neither is a required check, GitHub's merge button will not be blocked by that alone — the same class of silent breakage this PR is fixing could recur unless a maintainer manually notices the non-required red X. This isn't something the current diff can fix (branch-protection `contexts` live in repo settings, not in the tree), but the PR description's framing should be corrected, and I'd suggest a maintainer add `Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)` and `Build on Ubuntu 22.04 RISC-V 64` to the required-checks list as a follow-up (this is exactly the same gap that let 647b30e5 through — it wasn't only the backports-extra mirror, it's that no *required* check ever builds against a pinned Ubuntu-archive meson).

This doesn't block merging *this* PR — it fixes the immediate jammy breakage regardless — but it should be acknowledged, and ideally tracked as a follow-up (possibly folded into #4902).

### Minor — stale `meson >= 0.62.0` references left in three documentation files

The PR body says "All five sites now say 0.56.0: `meson.build`, the three control files, and `packaging/nnstreamer.spec`." That's true for those five, but a repo-wide grep for `0.62.0` turns up three more places still stating the old floor:

- `AGENTS.md:7` — "Build system: **Meson >= 0.62.0** + Ninja."
- `Documentation/getting-started-meson-build.md:23` — "* meson >= 0.62.0"
- `Documentation/how-to-run-examples.md:46` — "1. ninja-build, meson (>=0.62.0)"

These are prose, not build-enforced, so they can't break a build — but they directly contradict the new floor this PR establishes (and would tell a contributor on jammy that their 0.61.2 meson is insufficient, which is precisely the false impression this PR is trying to correct). Worth updating in this PR or as an immediate fast-follow for consistency.

### Nit — pre-existing, not introduced by this PR (no action needed)

`run_command(flatc, '--version', check : true)` still hard-fails the whole configure if `flatc` exists but errors out on `--version` (e.g., a broken/incompatible binary), rather than degrading gracefully. This is unchanged from the pre-647b30e5 code this PR restores (verified identical), so it's not a regression from this diff — flagging only because the task asked me to verify the found/not-found/failure behavior explicitly. No action requested.

## Scope check

Diff is 102 lines across 7 files (`meson.build`, three `debian/control*`, `packaging/nnstreamer.spec`, `pdebuild.yml`, `risc-v.yml`) — tightly scoped to the stated problem, no unrelated modules touched.

---
*Disclosure: this comment was generated by an automated review agent (a separate Claude instance) and relayed here; it is not a human reviewer's comment.*

myungjoo:

Both findings were right, and both are addressed in 37b1de9d.

**Minor, the three documents.** A real miss, and worse than a typo: the PR body
claimed "all five sites now say 0.56.0" while `AGENTS.md:7`,
`Documentation/getting-started-meson-build.md:23` and
`Documentation/how-to-run-examples.md:46` still said 0.62.0. There are eight
sites, not five. All eight now agree, and the body is corrected.

Rather than fix the three and move on, this is the second time the value has
drifted — 647b30e5 left two control files at 0.49, and this PR left three
documents behind — so `meson-version.sh` now takes `meson.build` as the
authority and requires every other declaration to match:

```
meson.build declares meson_version >= 0.56.0
  ok   AGENTS.md:7 (0.56.0)
  ok   debian/control:6 (0.56.0)
  ok   debian/control.debian:6 (0.56.0)
  ok   debian/control.ubuntu.ppa:6 (0.56.0)
  ok   packaging/nnstreamer.spec:226 (0.56.0)
  ok   Documentation/getting-started-meson-build.md:23 (0.56.0)
  ok   Documentation/how-to-run-examples.md:46 (0.56.0)
```

Putting your exact finding back produces:

```
::error::Documentation/getting-started-meson-build.md:23 requires meson 0.62.0,
but meson.build requires 0.56.0. Keep every declaration in step.
exit=1
```

It scans `Documentation/*.md` rather than a registered list, so a page added
later is covered without anyone remembering, and it keys on `>=` so prose
naming the meson some distro ships is not read as a requirement. Six fixtures
plus the shipped tree; shellcheck clean.

**Major, the strength of the guard.** Confirmed independently, and the PR body
overstated it. `gh api repos/nnstreamer/nnstreamer/branches/main/protection`
gives the required set as DCO, the three GBS jobs, MacOS, Android NDK,
LLVM/Clang, Valgrind on ubuntu-22.04, and Static checks. `pdebuild` and
`risc-v` are not in it, and of the ones that are, `ubuntu_clean_meson_build.yml`
and `ubuntu_clean_llvm_build.yml` both `pip install meson ninja` — PyPI's
newest, whatever the runner is — while MacOS uses `brew install meson`. So no
required check exercises the floor.

The body now says exactly that instead of "reintroducing a construct newer than
0.61.2 now fails CI". What the diff can do it does: the new checker runs under
Static checks, which *is* required. Making pdebuild and risc-v required is a
branch-protection setting, so it is left to the maintainer rather than claimed
here.

**Nit, `check: true`.** Agreed, and deliberately left as is — that behaviour is
restored verbatim from the pre-647b30e5 code, and changing it would be an
unrelated behavioural change in a PR about a version floor.

myungjoo:
[Automated re-review (round 2) by a separate Claude agent, relayed to this PR]

I read the round-1 thread first, then re-verified everything independently rather than trusting the response. Scope of this pass: 37b1de9d plus a fresh, adversarial look at `meson-version.sh`, since that's the new surface area.

**Verdict: still mergeable.** Both round-1 findings are genuinely fixed, and the PR body's revised claims check out against live data. The new checker script does have a real correctness gap, found by actively trying to break it as requested — one **major**, two **minor**, one **nit** — but none of it threatens the core jammy fix, which remains sound.

## Round-1 findings: resolved

- **Major (required-check gap).** Re-ran `gh api repos/nnstreamer/nnstreamer/branches/main/protection` myself. `required_status_checks.contexts` is exactly: `DCO`, the three `Tizen GBS build` jobs, `Build on MacOS-latest with Meson`, `Android NDK build on Ubuntu for arm64-v8a`, `Build on Ubuntu with LLVM/Clang (ubuntu-22.04)`, `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)`, `Static checks`. Neither `pdebuild` nor `risc-v` is in it. This matches the PR body's corrected "Regression test" section word for word. The body no longer claims CI as a whole blocks the regression class; it correctly narrows the claim to "Static checks (required) runs the declaration checker" and explicitly defers the branch-protection change to a maintainer. Honest and accurate.
- **Minor (stale 0.62.0 docs).** `AGENTS.md:7`, `Documentation/getting-started-meson-build.md:23`, `Documentation/how-to-run-examples.md:46` are all `0.56.0` now. A repo-wide grep for `0.62.0` after extracting the PR head as a clean tree turns up nothing left except (a) the explanatory comment in `meson.build:173` ("`flatc.version()` needs meson 0.62.0, which Ubuntu 22.04 does not ship") and (b) fixture literals inside the two new self-test scripts — both intentional, neither a live declaration. Confirmed all eight sites (`meson.build`, three `debian/control*`, `packaging/nnstreamer.spec`, `AGENTS.md`, two `Documentation/*.md`) say `0.56.0`.
- Diff vs `git merge-base origin/main claude/jammy-meson-floor` is exactly the 13 files the PR body describes (`pdebuild.yml`, `risc-v.yml`, the two new scripts, `static.check.yml`, the three docs, three `debian/control*`, `meson.build`, `nnstreamer.spec`) — tightly scoped, nothing extraneous accumulated across the two commits.
- `static.check.yml` wiring for the new checker/self-test steps is copy-consistent with the pre-existing `debian-control.sh`/`test_debian_control.sh` pattern (same head-sha checkout, same removal guard via `git show --diff-filter=D`). Verified the YAML parses and the removal-guard logic is unchanged in shape from the existing steps.
- Shellcheck (`koalaman/shellcheck-alpine:stable -s bash`) is clean on both `meson-version.sh` and `test_meson_version.sh`, matching the PR's claim.
- `test_meson_version.sh` genuinely re-runs the checker fresh in `mktemp -d` trees each time and closes stdin the way GitHub Actions does — it is not faking a pass. All 6 fixture cases plus the "check the shipped tree" case pass as claimed when run under WSL bash on the LF-exported tree.

## New findings (this round)

### Major — `found_raw` is extracted from the whole line, not from the matched declaration, so it can misattribute an unrelated dotted version to meson (false accept *and* false reject)

`meson-version.sh`'s inner loop does:
```
found_raw=$(echo "${hit#*:}" | grep -oE '[0-9]+(\.[0-9]+)+' | tail -1)
```
This re-scans the *entire line* for every dotted-number substring and takes the **last** one — it does not extract the version from the specific `DECL_RE` match that triggered the hit. `DECL_RE` itself only anchors on proximity to the word "meson"; the version actually compared comes from wherever the last dotted number on the line happens to sit.

I reproduced this two ways:

1. **False positive, against the real, current `debian/control`.** The live line is:
   `ninja-build, meson (>=0.56.0), debhelper (>=9), nnstreamer-edge-dev,`
   It only survives today because `debhelper (>=9)` has no dot. Bumping just `debhelper` to a dotted minimum — an entirely unrelated, correct packaging edit — breaks it. I made exactly that one-word edit on the checked-out tree (`debhelper (>=9)` → `debhelper (>=9.20)`) and ran the checker unmodified:
   ```
   ::error::debian/control:6 requires meson 9.20, but meson.build requires 0.56.0. Keep every declaration in step.
   ::error::The meson version declarations disagree.
   exit=1
   ```
   That is a false failure on `Static checks`, a **required** status check, for a change that has nothing to do with meson.

2. **False negative, via a fixture.** Given
   `meson.build` → `>=0.56.0`, `debian/control` → `Build-Depends: meson (>=0.49), ninja-build (>=0.56.0)`
   (a genuinely stale meson requirement, with an unrelated package's *matching* version placed after it on the same line), the checker reports:
   ```
     ok   debian/control:1 (0.56.0)
   The meson version declarations agree.
   exit=0
   ```
   The actual meson requirement (0.49) is never even printed — the "found" value silently became ninja-build's version instead. This is precisely the class of drift the checker exists to catch, and it can miss it.

Both directions come from the same root cause and are easy to trigger with an ordinary future packaging edit (add or change any dotted-version `Build-Depends`/`BuildRequires` entry on the same physical line as meson's). A minimal fix: extract the version from the `DECL_RE` match itself rather than rescanning the whole line, e.g.
```
matched=$(echo "$line" | grep -oE "$DECL_RE" | tail -1)
found_raw=$(echo "$matched" | grep -oE '[0-9]+(\.[0-9]+)+' | tail -1)
```
I confirmed `echo "$line" | grep -oE "$DECL_RE"` on the same adversarial line correctly isolates `meson (>=0.56.0` (stopping before `ninja-build`'s number), so this fix is straightforward.

This doesn't affect the current tree (today's lines happen not to trigger it, and `test_meson_version.sh`'s "check the shipped tree" step still passes), and it doesn't undermine this PR's actual goal of unblocking jammy. But it does mean the new required-check guard is not as robust as the PR body implies ("takes meson.build as the authority and requires every other declaration to match") — it can be defeated by an unrelated, ordinary edit in either direction.

### Minor — the "meson" proximity heuristic can also misfire on prose

`DECL_RE = '[Mm]eson[^0-9]{0,24}>=[[:space:]]*[0-9]+(\.[0-9]+)+'` has no word boundary and no requirement that the `>=` actually apply to meson — it just looks for the literal word within 24 non-digit characters of *some* `>=`-version. A `Documentation/*.md` sentence like:
```
As of this meson.build, glib >= 2.60 is required.
```
is read as "meson requires 2.60" and fails against the 0.56.0 floor:
```
::error::Documentation/x.md:1 requires meson 2.60, but meson.build requires 0.56.0.
```
Reproduced with a fixture. This is a narrower risk than the major above (it needs someone to write about meson and another `>=`-versioned tool in the same sentence in a top-level `Documentation/*.md` page), but it's the same underlying design gap — proximity to the word "meson" is not the same as "this `>=` is meson's."

### Minor — Documentation files with a space in the name are silently never checked

`targets` is built by unquoted command substitution and iterated with `for rel in $targets`, so `find Documentation -maxdepth 1 -type f -name '*.md'` output is subject to word-splitting on whitespace. A file such as `Documentation/old notes.md` containing a stale `meson (>=0.49)` line is never even listed in the "ok"/"error" output — the checker exits 0 without having looked at it (reproduced with a fixture). No file in the repository today has a space in its name, so this isn't live, but it's a fragility in exactly the "detect drift from a future addition automatically" property the PR's commit message claims (`It reads a fixed set of paths plus Documentation/*.md, so a page added later is covered without anyone remembering to register it`) — a page added later with a space in its name would not be covered.

### Nit — `normalize()` truncates at three components

```
awk -F. '{ printf "%d.%d.%d\n", $1, ($2 == "" ? 0 : $2), ($3 == "" ? 0 : $3) }'
```
only ever looks at `$1`/`$2`/`$3`. Two declarations differing solely in a fourth component (`0.56.0.1` vs `0.56.0.9`) normalize identically and are accepted as matching (reproduced with a fixture). Not practically relevant — meson and this project's other version strings never go to four components — flagging only because the task asked for it explicitly.

## What I did not find

No issue with `flatc_ver = run_command(flatc, '--version', check : true).stdout().split()[2]` beyond what round 1 already covered (byte-identical revert, `check: true` behavior pre-existing). No issue with the `0.56.0` floor itself — re-confirmed no other construct in the tree needs more. No issue with `pdebuild.yml`/`risc-v.yml` beyond the already-acknowledged required-check gap. No stray file left at `0.62.0` as a live declaration. No unrelated modules touched.

## Bottom line

Both round-1 items are resolved correctly and the PR body's claims now match what I could independently verify (branch protection contents, doc consistency, shellcheck, self-test behavior). The new `meson-version.sh` has a genuine extraction bug that can both false-fail an unrelated future change and false-pass a real meson-version regression when another dotted version shares its line — demonstrated against the real `debian/control`, not just synthetic fixtures. I'd treat this as a fast-follow rather than a blocker for this PR: it doesn't regress anything the current tree does, and the PR's stated purpose (unblock jammy, stop the floor from drifting again in the eight already-known sites) is otherwise met. Worth fixing before the next time a `Build-Depends`/`BuildRequires` line near meson's gets touched.

---
*Disclosure: this is a round-2 automated re-review, generated by a separate Claude agent and relayed here by the PR author's tooling; it is not a human reviewer's comment.*

myungjoo:

The major finding is real and is fixed in 113cae16 rather than deferred. It was
a defect in code added by this PR, and it could have failed a required check on
an unrelated edit, so leaving it for a fast-follow was not defensible.

Reproduced against the previous commit's checker, not a reconstruction:

```
A  debhelper (>=9.20) beside meson (>=0.56.0)
     BEFORE (37b1de9d) -> exit 1 (want 0)  WRONG
     AFTER             -> exit 0 (want 0)  OK
B  meson (>=0.49.0) beside debhelper (>=0.56.0)
     BEFORE (37b1de9d) -> exit 0 (want 1)  WRONG
     AFTER             -> exit 1 (want 1)  OK
```

Both directions, as you described. The version now comes out of the matched
substring, and each match on a line is walked, so a sentence naming two
requirements is fully checked.

**Minor, the 24-character gap.** Tightened rather than tuned: the pattern now
requires `>=` to follow `meson` with nothing between them but an optional
bracket, which every declaration in the tree already satisfies. A side effect
worth noting is that the gap can then hold no digits at all, so the only digits
in a match belong to the version — that is what makes reading from the match
safe rather than merely likely.

One correction to the record: your `glib >= 2.60` example did not reproduce on
the old checker with the phrasing I tried, because the gap there ran past 24
characters. The class of problem is real and the tightening removes it, but I
could not confirm that specific case failed before.

**Minor, word splitting.** Fixed; the file list is iterated over lines. A
fixture creates `Documentation/getting started.md` with a mismatched version
and requires the checker to still reject it, so the skip cannot come back
silently.

**Nit, four components.** Padded to four.

The suite is at 12 fixtures plus the shipped tree; shellcheck clean.

myungjoo:

I read both prior rounds and the author's responses first, then re-verified everything against real files and a from-scratch adversarial pass on `meson-version.sh`/`test_meson_version.sh`, since 113cae16 is the only new surface area this round.

**Verdict: mergeable.** Both round-2 findings are genuinely fixed, reproduced independently against the pre-fix script (not just re-reading the diff). One **minor** new finding: the regex tightening that fixed round 2's "24-character gap" false positive also removed detection for several plausible future declaration forms. Nothing here threatens the core jammy fix.

## Round-2 findings: verified fixed

- **Major (extraction from line vs match).** Ran both directions against the actual pre-fix (37b1de9d) checker and the current (113cae16) one, side by side:
  - `debian/control` with `meson (>=0.56.0), debhelper (>=9.20)` on one line: 37b1de9d → exit 1 (false failure on an unrelated field); 113cae16 → exit 0 (correct).
  - `debian/control` with `meson (>=0.49.0), debhelper (>=0.56.0)`: 37b1de9d → exit 0 (missed a real stale meson floor); 113cae16 → exit 1 (correct).
  Matches the author's reproduction exactly.
- **Minor (24-char gap / glib false positive).** Re-tested the `"meson.build additionally wants glib >= 2.60 at configure time."` phrasing against the *old* checker directly: it does **not** false-positive (exit 0), confirming the author's correction that this specific phrasing was never actually broken — the gap exceeds 24 chars only in a different phrasing. The new tighter pattern (`meson` then only optional whitespace/one bracket then `>=`) is still correct and is what makes reading the version out of the match safe.
- **Minor (word-splitting on `Documentation/*.md` with spaces).** Built a tree with `Documentation/getting started.md` carrying a stale `meson >= 0.62.0` line and ran both checkers: 37b1de9d → exit 0 (file silently skipped, real bug); 113cae16 → exit 1, correctly flags `Documentation/getting started.md:1`. Fixed, and the self-test's fixture for this is a genuine regression test, not a rubber stamp.
- **Nit (4-component padding).** Confirmed `normalize()` now pads/reads all four components.

## Independently verified against the real tree

- Ran `meson-version.sh` against the actual PR-head checkout (no fixtures): all 8 declared sites are found and agree at 0.56.0 — `meson.build` (authority), `debian/control:6`, `debian/control.debian:6`, `debian/control.ubuntu.ppa:6`, `packaging/nnstreamer.spec:226`, `AGENTS.md:7`, `Documentation/getting-started-meson-build.md:23`, `Documentation/how-to-run-examples.md:46`. `git ls-tree` confirms no fourth `debian/control*` file exists that the glob could miss.
- Ran `test_meson_version.sh` end to end: all 10 `expect_checker` cases + the spaced-path case + the shipped-tree check pass (12 checks total).
- `shellcheck -s bash` (koalaman/shellcheck-alpine:stable) is clean on both scripts.
- `normalize()` edge cases: empty string → `0.0.0.0` (never reached live, since callers guard with `[ -z ... ]` first), leading zeros (`0.056.0` → `0.56.0.0`, decimal not octal — awk's numeric coercion, correct), 5-component input truncates to 4 as documented. No live bug.
- `check_file()`'s nested `while read` + process-substitution + `failed=1`: `failed` is deliberately *not* declared `local`, so it correctly mutates the script-global counter from inside both nested loops and from the outer per-file loop (all three are process substitutions in the current shell, not subshells via pipe) — confirmed by the self-test's failing fixtures actually producing exit 1.
- `static.check.yml` wiring for the two new steps matches the existing `debian-control.sh`/`test_debian_control.sh` removal-guard pattern exactly (same `--diff-filter=D` check).
- Diff from 37b1de9d to 113cae16 touches only `meson-version.sh` and `test_meson_version.sh` (72+33 lines) — no accumulated drift into the other 11 files from rounds 1–2.
- PR body's current claims (branch-protection contents, "all eight sites now say 0.56.0", the corrected "Regression test" section) still match what I independently re-checked; nothing has gone stale since round 2.

## New finding

### Minor — the round-2 regex tightening traded away detection of common Markdown/quoting forms

`DECL_RE` now requires `>=` to follow `meson`/`Meson` with nothing between but optional whitespace and one optional `(`. That closed the false-positive/extraction bug, but it also means the checker no longer recognizes a declaration once it's dressed in ordinary Markdown, whereas the *previous* (37b1de9d) 24-char-gap version did. Built a stale (`0.99.0` vs the tree's `0.56.0`) declaration in a `Documentation/*.md` fixture in each form and ran both checkers:

| form | 37b1de9d (old) | 113cae16 (current) |
|---|---|---|
| `**meson** >= 0.99.0` (bold) | correctly errors | silently passes |
| `` `meson` >= 0.99.0 `` (backtick) | correctly errors | silently passes |
| `[meson](https://mesonbuild.com) >= 0.99.0` (link) | correctly errors | silently passes |
| `meson_version: '>=0.99.0'` (quoting meson.build's own syntax as a doc example) | correctly errors | silently passes |
| `MESON >= 0.99.0` (all-caps) | silently passes on **both** (pre-existing gap, `[Mm]` never covered `MESON`) | silently passes |

The first four are regressions in coverage specifically introduced by 113cae16, not pre-existing gaps — none of these are contrived edge cases; bolding a tool name, linking to its homepage, or quoting `meson.build`'s own `meson_version:` syntax when explaining the requirement in a getting-started doc are all things a contributor would plausibly write. None of them are live in the tree today, and the self-test's own repository-tree check still passes, so this doesn't affect the current PR's correctness. But it does mean the property the task and this checker's own commit message aim for — automatically catching a future stale declaration without anyone remembering to register it — has a real, demonstrated hole for anything but the plainest `word (>=X)` / `word >= X` phrasing. Given `Documentation/*.md` is exactly the surface most likely to be edited in prose/Markdown style (unlike the mechanically-generated `debian/control`/`.spec` lines), this is worth a follow-up (extend `DECL_RE` to tolerate `` ` ``, `*`, and `[...](...)` around the word, and/or special-case the `meson_version:` quoting form already handled for `meson.build` itself) rather than a blocker.

## What I did not find

No new issue with `meson.build`, `pdebuild.yml`, `risc-v.yml`, `debian/control*`, `packaging/nnstreamer.spec`, or the three prose docs — all unchanged since round 2 and previously verified. No scope creep: the only files touched since round 2 are the two scripts under discussion. No unrelated modules touched anywhere in the 13-file diff.

## Bottom line

Both round-2 items are fixed with independently reproduced before/after evidence, not just re-reading the diff or trusting the response. The one new item is a minor, non-blocking coverage gap in the brand-new self-checking tooling (not in the actual jammy fix), worth a fast-follow given it affects exactly the kind of Markdown prose the checker's Documentation coverage is meant to protect.

---
*Disclosure: this is a round-3 automated re-review, generated by a separate Claude agent and relayed here; it is not a human reviewer's comment.*

myungjoo:

The new minor is fixed in 7ec9fb41, and it deserves to be called what it is: a
regression I introduced in 113cae16 while fixing your previous finding. Trading
a false positive for a silent miss was the wrong trade, and Documentation is
exactly the kind of prose that gets reworded, so the miss would have gone
unnoticed.

Verified against 113cae16 with a stale 0.62.0 in each form:

```
**meson** >= 0.62.0        PREV: exit 0 (MISSED)   NOW: exit 1  OK
`meson` >= 0.62.0          PREV: exit 0 (MISSED)   NOW: exit 1  OK
MESON >= 0.62.0            PREV: exit 0 (MISSED)   NOW: exit 1  OK
meson_version: >=0.62.0    PREV: exit 0 (MISSED)   NOW: exit 1  OK

meson.build ... glib >= 2.60   PREV: exit 0  NOW: exit 0  (no false positive)
```

The gap may now hold punctuation but neither a letter nor a digit. That is the
property that actually distinguishes a declaration from prose — another package
name between `meson` and `>=` is what makes the glib sentence not a meson
requirement — and it keeps the guarantee that the only digits in a match belong
to the version. The keyword matches without regard to case, and `meson_version`
is accepted so a page quoting meson.build is covered.

The markdown link form you raised, `[meson](https://example.com) >= X`, is
still not recognised, because the URL puts letters in the gap. Rather than
loosen further and reopen the false positive, the script documents the limit
and says to write the requirement outside the link text.

Four fixtures added, one per form. The suite is at 16 plus the shipped tree.

Also folded in: those backticks made shellcheck report SC2016 on the fixture,
so the suite was no longer clean after 7ec9fb41. Disabled for that line with
the reason beside it (19b50eb4); shellcheck is back to exit 0.

myungjoo:

I read all three prior rounds and the author's responses first, then re-verified everything independently against the real PR-head tree (exported with `core.autocrlf=false` so the shell scripts run as authored), with the adversarial focus this round asked for: `DECL_RE` in 7ec9fb41/19b50eb4, since that is the only new surface area since round 3.

**Verdict: still mergeable.** All three prior rounds' findings are genuinely fixed, reproduced independently rather than re-read from the diff. One **minor** new finding: a narrow, non-live residual gap in the same proximity heuristic that's been tightened every round so far. Nothing here touches the core jammy fix.

## Rounds 1–3 findings: still fixed, re-verified against the current tree

- **Round 1 major (required-check strength) and minor (three stale 0.62.0 docs).** Re-ran the checker against the actual PR-head tree: all 8 sites (`meson.build`, three `debian/control*`, `packaging/nnstreamer.spec`, `AGENTS.md`, two `Documentation/*.md` pages) report `0.56.0` and the PR body's "Regression test" section still correctly states neither `pdebuild` nor `risc-v` is a required check. No drift since round 2.
- **Round 2 major (version read from the whole line vs the match).** Re-derived from source: the extraction now reads the version out of each per-match substring (`grep -oiE "$DECL_RE"` first, then pulls digits from that substring only), and — importantly — this is now true *by construction*, not just by careful sequencing: the gap is `[^[:alnum:]]{0,12}`, which structurally cannot contain a digit, so the only digits inside any match are the version's own. A neighbouring `debhelper (>=9.20)` or `libfoo (>=1.2.3)` cannot leak into the meson comparison regardless of order on the line. Confirmed with the file's own fixtures.
- **Round 3 minor (markdown decoration hides a declaration).** Verified all four previously-missed forms now match: `**meson** >= 0.62.0`, `` `meson` >= 0.62.0 ``, `MESON >= 0.62.0`, `meson_version: >=0.62.0`. Also re-confirmed the one form the script explicitly documents as still unrecognized, `[meson](url) >= X` (a markdown link), does in fact still not match — the limitation comment in the script matches the actual behavior.
- **Round-3 follow-up (SC2016).** Ran shellcheck (`koalaman/shellcheck-alpine:stable`, `-s bash`) on both scripts directly: exit 0, clean. The `# shellcheck disable=SC2016` sits directly above the fixture with the backticks, with the reason stated inline, matching the response's description.
- **Self-test is not vacuous.** Ran `test_meson_version.sh` end to end against the exported tree: all 14 `expect_checker` cases + the spaced-path case + the shipped-tree check pass (16 PASS lines total), and each was checked for the *right* reason by construction (e.g., the "neighbouring constraint" fixtures encode both directions of the round-2 bug).
- **Scope across all 5 commits.** `git diff --stat origin/main...HEAD` is exactly 13 files / 303+ / 12- (matches the PR body). Per-commit `git show --stat` confirms each commit touches only what its message claims — 7aadee8b (the 7-file jammy floor), 37b1de9d (new checker + 3 docs), 113cae16 and 7ec9fb41 (the two scripts only), 19b50eb4 (a 2-line shellcheck suppression). No accumulated drift into unrelated files.
- **The 8 real declaration sites, re-verified line-by-line against `git show HEAD:<file>`:** all 8 are found and reported `ok` by the checker on the real tree (not a fixture), and `git ls-tree` confirms no fourth `debian/control*` file or second `AGENTS.md`/`nnstreamer.spec` exists that a glob could miss.

## New finding

### Minor — `DECL_RE`'s punctuation-only gap still allows a floating, unrelated version to be misread as meson's, when no package name sits between them

`DECL_RE='meson(_version)?[^[:alnum:]]{0,12}>=[[:space:]]*[0-9]+(\.[0-9]+)+'` closes the round-2 bug (a *named* neighbour like `debhelper (>=9.20)` or `glib >= 2.60` can no longer leak in, because the name's letters break the gap). But the gap can still be crossed by punctuation/whitespace alone, so a version with no attached name in between is still misattributed. Reproduced directly against the shipped script (not a hypothetical):

```
$ echo "See meson, (>=5.0) for details of some other unrelated tool." | grep -noiE "$DECL_RE"
1:meson, (>=5.0

$ echo "After installing meson... (>=2024.1) support was added upstream." | grep -noiE "$DECL_RE"
1:meson... (>=2024.1
```

Both would be reported as `Documentation/x.md:N requires meson 5.0` / `2024.1`, and fail against the 0.56.0 floor — a false failure on **Static checks**, the required job, for prose that isn't declaring anything about meson.

I also checked the boundary is exactly as documented: a 12-character punctuation-only gap matches, a 13-character one does not (`meson............>=0.99.0` vs `meson.............>=0.99.0`), so the `{0,12}` bound itself behaves correctly — this isn't an off-by-one, it's the same class of gap the fix already accepts, just triggered by an unnamed rather than a named neighbour.

This is meaningfully narrower than every previously-fixed variant of this heuristic: it requires a version constraint with no name of its own in the same clause, which is unusual phrasing (ordinary English almost always names what a version applies to, and that name's letters are exactly what blocks the false match — I checked this against every existing "meson" mention across `Documentation/*.md` and `AGENTS.md` in the current tree with `grep -rniE`, and none of the ~25 live mentions trigger it). Not live in the tree today, and it does not affect this PR's actual purpose (unblocking jammy) or the checker's usefulness against the drift that has actually happened twice (control files, then docs). Given the pattern of this heuristic being tightened once per round without fully closing the class, I'd suggest either accepting this as a documented residual limit (the way the markdown-link case already is, right above `DECL_RE` in the script's own comment) or requiring at least one alphabetic run of "meson" as the last word before the gap starts (i.e. an explicit word-end check) as a fast-follow — not a blocker either way.

## What I did not find

No issue with the 8 real declaration sites — all present, all correctly extracted, all `0.56.0`. No new false negative: I could not construct a real declaration form in the current tree's style that the current `DECL_RE` fails to catch. No case-insensitivity regression: `MESON`/`Meson`/`meson` all match uniformly now, and I found no *live* word in the repository whose case-insensitive spelling accidentally contains `meson` as a substring (the lack of a word-boundary is a pre-existing property of the pattern since round 1's `[Mm]eson`, not something 7ec9fb41 introduced). No issue with `(_version)?` — it only ever fires directly after the literal `meson`, and `meson.build`'s own line is checked with a separate, stricter regex, not `DECL_RE`, so this addition cannot double-count or drift against the authority extraction. No scope creep in any of the 5 commits. No stale `0.62.0` left as a live declaration anywhere, including nested `Documentation/hotdoc/` and `Documentation/tutorials/` pages that the checker intentionally does not scan (`find Documentation -maxdepth 1`) — I checked those by hand since they're out of the checker's reach, and none carry a stale meson requirement.

## Bottom line

All three prior rounds' findings are resolved, and reproduced independently rather than trusted from the diff or the response. The one new item is a minor, non-blocking, non-live edge case in the declaration-detection heuristic — the same class of finding as rounds 2 and 3, further narrowed but not fully eliminated. It does not threaten the jammy fix itself, the required-check wiring, the documentation consistency, or the self-test's integrity. Mergeable as-is.

---
*Disclosure: this is a round-4 automated re-review, generated by a separate Claude agent and relayed here; it is not a human reviewer's comment.*

myungjoo:

Fixed in e975baaa, and your framing of it is the useful part: this heuristic
had been narrowed once per round without closure. Both of your cases now behave,
verified against 19b50eb4:

```
See meson, (>=5.0) is the tool version we mean elsewhere.
    PREV (19b50eb4) -> exit 1 (want 0)  WRONG
    NOW             -> exit 0 (want 0)  OK
After installing meson... (>=2024.1) support was added.
    PREV (19b50eb4) -> exit 1 (want 0)  WRONG
    NOW             -> exit 0 (want 0)  OK
```

Comma, period and semicolon are excluded from the gap. They mean the sentence
has moved on, while every form that has to keep working — brackets, asterisks,
backticks, the colon after `meson_version` — uses none of them. The four
decorated forms from round 3 and all eight real sites were re-checked and are
unchanged.

On the pattern of narrowing: the script now states that this is a heuristic
over prose, that its limits are written down, and that a phrasing which escapes
it should be rewritten in one of the plain forms rather than answered by
widening the pattern again. If a round 5 finding is another phrasing rather
than a defect, that note is the intended answer to it.

Suite at 18 checks plus the shipped tree; shellcheck exit 0.

One thing worth putting on the record for whoever merges this. Every finding
across four rounds has been in `meson-version.sh`, which this PR adds; none has
been in the change that actually unblocks jammy. That change is one commit,
7aadee8b, and it has been stable since round 1: `run_command` for the flatc
version, floor at 0.56.0 in eight places, the two CI workarounds dropped. If
the checker is felt to be carrying too much of this PR's weight, splitting it
into its own PR would leave the jammy fix mergeable on its own — say the word
and it will be split rather than argued for.

myungjoo:
[Automated re-review (round 5) by a separate Claude agent, relayed to this PR]

I read all four prior rounds and responses first. This round's mandate was different from 1-4: re-verify `e975baaa`'s closure of round 4, but spend the bulk of the effort on an independent, adversarial, from-scratch review of **7aadee8b** — the commit that actually unblocks jammy — since every finding across rounds 1-4 was in `meson-version.sh`, and that commit had not been re-examined since round 1's lighter pass.

**Verdict: still mergeable.** Round 4 is genuinely closed for the cases it demonstrated. 7aadee8b holds up under a strict independent re-check, including two angles nobody had checked yet (whether the retained `jammy-backports` mirror could quietly reintroduce the same masking bug, and exactly what the live contents of the removed PPA are). One **nit**: the same regex heuristic class from rounds 2-4 has a further, narrower gap, which I flag only because the task asked for it explicitly — given the author's stated intent to stop widening this pattern, I do not think this warrants another patch.

## 1. e975baaa: does the new gap close round 4?

Reproduced against the shipped `meson-version.sh` (not a reconstruction) with the exact two round-4 cases:

```
"See meson, (>=5.0) is the tool version we mean elsewhere."          -> no match (was a false positive before)
"After installing meson... (>=2024.1) support was added upstream."   -> no match (was a false positive before)
```

Both are fixed, matching the author's response exactly. I also re-ran `test_meson_version.sh` end to end: all 17 checks pass, including the 4 markdown-decoration forms from round 3 (`**meson**`, `` `meson` ``, `MESON`, `meson_version:`) and all 8 real declaration sites in the live tree (`meson.build`, three `debian/control*`, `packaging/nnstreamer.spec`, `AGENTS.md`, two `Documentation/*.md` pages) — none of that regressed. `shellcheck -s bash` is clean on both scripts.

**Nit — the underlying class ("punctuation-only gap, no named neighbour") is narrowed again, not eliminated.** The fix excludes comma/period/semicolon from the gap, which closes exactly the two demonstrated cases. But other punctuation with no letters/digits still crosses it:

```
$ echo "meson [*] (>=5.0) is a footnote reference to something else." | grep -noiE "$DECL_RE"
1:meson [*] (>=5.0
$ echo "The meson -- (>=9.0) marker refers to another tool entirely." | grep -noiE "$DECL_RE"
1:meson -- (>=9.0
$ echo "meson?! (>=5.0) what a coincidence" | grep -noiE "$DECL_RE"
1:meson?! (>=5.0
```

An em-dash aside or a bare (non-digit) footnote marker between "meson" and an unrelated version still misattributes it, the same way comma/ellipsis did before this commit. Not live anywhere in the tree, and admittedly a stretch to write naturally compared to the round-4 cases. Given the script's own comment now states this is "a heuristic over prose... deliberately not grown any further" and the author's last response says a round-5 finding that is "another phrasing rather than a defect" should be answered by that note rather than another patch — I'm recording this for the log, not asking for a fix. Treat it as closed unless a future PR trips over it for real.

## 2. Independent review of 7aadee8b (the actual jammy fix)

This is the commit every prior round left unexamined beyond round 1's lighter pass. I re-derived each claim from primary sources rather than trusting the PR body or prior rounds.

**`split()[2]` safety — checked at the source level, not just by sampling binaries.** Pulled flatc's own version-printing code (`google/flatbuffers` `src/flatc.cpp`):
```c
} else if (arg == "--version") {
  printf("flatc version %s\n", FLATC_VERSION());
```
and `FLATBUFFERS_VERSION()` is documented in `include/flatbuffers/base.h` to return exactly `"MAJOR.MINOR.REVISION"` — three dot-separated integers, no spaces, no build metadata, no commit suffix. That means `.split()[2]` is not merely "safe across the versions someone happened to test" (as round 1 checked black-box), it's safe by construction: the output format is a fixed 3-whitespace-token string with the version holding no embedded whitespace, for every version flatc has ever shipped. There is no flatc release that appends a commit hash or build tag after the version on that line.

**`check: true` and the `flatbuf-support` feature option — reproduced all three states in a real `ubuntu:22.04` container** with a `flatc` on `PATH` that fails on `--version`:
- `flatbuf-support=disabled` → meson prints `Program flatc skipped: feature flatbuf-support disabled` and never even invokes `run_command`. Configure succeeds. **A disabled build cannot be broken by a bad flatc.**
- `flatbuf-support=auto` (the project's actual default, per `meson_options.txt:22`) → `find_program` still finds the broken binary and `run_command(..., check: true)` **hard-fails the entire configure** with `ERROR: Command "/usr/local/bin/flatc --version" failed with status 1`, even though flatbuf support was only ever opportunistic.
This confirms round 1's "nit" precisely, and narrows it: the failure mode is real for `auto`, not just `enabled`, but is fully bypassed by `disabled`. I also confirmed via `git show 647b30e5^:meson.build` that this exact `check: true` line, in this exact `if flatc.found()` block, predates 647b30e5 byte-for-byte — this PR restores old behavior rather than introducing new risk. Not a regression from this diff.

**0.56.0 floor — re-derived across the whole tree, subdirs included, not just the top-level file.** Grepped all 34 `meson.build` files under `ext/`, `gst/`, `tests/`, `tools/`, and top-level for every version-gated construct I could think of (`.version()`, `fs.*`, `full_path()`, `project_source_root()/project_build_root()`, `summary()`, `install_symlink`, `pkgconfig`, `add_dist_script`, `both_libraries`, `allow_fallback`, `add_devenv`, `override_dependency`, `console:`, `include_type:`, `.get_variable(`, `dirs:`, `disabler()`). Findings, each cross-checked against meson's own release-notes pages (not secondhand):
- `meson.project_source_root()`/`project_build_root()` — confirmed new in 0.56.0 directly from `Release-notes-for-0-56-0.html` ("New project root functions"). Used in the top-level `meson.build` and in `ext/nnstreamer/tensor_filter/meson.build`, `tests/nnstreamer_filter_extensions_common/meson.build`, `tests/nnstreamer_filter_mvncsdk2/meson.build` — this is the tightest constraint anywhere in the tree, matching the declared floor exactly.
- `ExternalProgram.full_path()` — confirmed 0.55.0 from `Release-notes-for-0-55-0.html` ("New method to be consistent with built objects"). Used in `ext/nnstreamer/meson.build`, `gst/nnstreamer/meson.build`, `tools/development/parser/meson.build`.
- `import('fs')` / `fs.is_dir()` / `fs.is_file()` — module added 0.53.0, used in the top-level file and `ext/nnstreamer/tensor_filter/meson.build`.
- The various `.version()` calls on `tf_support_deps[0]`, `tflite_support_deps[0]`, `lua_support_deps[0]` in `ext/nnstreamer/tensor_filter/meson.build` are `Dependency.version()`, a much older method on a different object type from `ExternalProgram.version()` — confirmed unaffected.
- No `summary()`, `install_symlink`, `pkgconfig` module, `add_dist_script`, `both_libraries`, `allow_fallback`, `add_devenv`, `override_dependency`, `console:`, `include_type:`, or `dirs:` usage anywhere in any of the 34 files.
So 0.56.0 is independently confirmed as the exact minimum across the entire tree, subdirectories included, not just the file the PR body walks through.

**CI workaround removal — checked against live Launchpad data, going further than "git log shows it was single-purpose."** Queried the Launchpad REST API directly for `~kubuntu-ppa/+archive/ubuntu/backports-extra`'s published binaries for jammy/amd64:
- `meson` is published there at `0.63.2-1~ubuntu22.04~ppa1` (confirms the PR body's number).
- Checked every "exotic" package in `debian/control.ubuntu.ppa`'s `Build-Depends` (`tensorflow2-lite-dev`, `openvino-dev`, `onert-dev`, `tvm-runtime-dev`, `onnxruntime-dev`, `libedgetpu1-std`, `libedgetpu-dev`, `nnstreamer-edge-dev`, `libpaho-mqtt-dev`, `protobuf-compiler`, `libflatbuffers-dev`, `flatbuffers-compiler`, `libtorch-dev`, `ninja-build`) against that same PPA/arch/series — **zero of them are published there.** Only `meson` is. So the mirror really was meson-only for this project's purposes, and removing it cannot strand any other Build-Depends.
- Confirmed via `git log --follow -p` that the `kubuntu-ppa/backports-extra` line in `pdebuild.yml` was added in 647b30e5 itself (not pre-existing), and that `pdebuild.yml`'s matrix has exactly one entry (`jammy`), so no other distro's pdebuild job is touched.
- One angle no prior round checked: the `OTHERMIRROR` line **still keeps** `archive.ubuntu.com/ubuntu jammy-backports universe` after this diff. If that pocket ever carried a newer meson, the same masking bug could recur through a different mirror. Checked Launchpad directly: `meson` has **no** publication history in the jammy Backports pocket at all (only Release/Proposed going up to `0.61.2-1`) — so the retained mirror cannot resupply a newer meson. The fix isn't leaving a second, unnoticed backdoor open.
- `risc-v.yml`: grepped `.github/workflows/get_ssat.sh` and `packaging/run_unittests_binaries.sh`, the two scripts this job actually runs — neither references `pip` anywhere, confirming `python3-pip`'s removal from the `apt-get install` line strands nothing else in that job.

**`packaging/nnstreamer.spec` — confirmed logically safe for Tizen GBS**, independent of trusting the "a lower floor can't break a passing build" argument at face value: grepped the whole spec for any other `meson`-conditioned logic (`%if`/macro gating) and found none — the only meson-version-sensitive line is the `BuildRequires` itself. A strictly lower minimum cannot newly fail a toolchain that already satisfied the higher one.

## 3. Commit split and PR-body consistency

- `git diff --stat origin/main...HEAD`: 13 files, +315/-12, matching the PR body.
- Per-commit `git show --stat`: each of the 6 commits (7aadee8b, 37b1de9d, 113cae16, 7ec9fb41, 19b50eb4, e975baaa) touches exactly the files its message claims — no drift accumulated across rounds.
- `static.check.yml`'s wiring for the meson-version checker/self-test steps is still byte-for-byte parallel to the pre-existing `debian-control.sh`/`test_debian_control.sh` steps (same merge-ref guard, same removal check).
- Re-ran `gh api repos/nnstreamer/nnstreamer/branches/main/protection`: required contexts are unchanged from round 1 (`pdebuild`/`risc-v` still not required); the PR body's "Regression test" section is still accurate today, not just at the time it was written.
- `gh pr checks 4904`: `Static checks` and `DCO` pass; GBS/pdebuild/risc-v are mid-run on the latest push at time of this review, nothing red.

## What I did not find

No regression introduced by 7aadee8b in any module outside the 7 files it touches. No case where `split()[2]` can pick up the wrong token, for any flatc release past or present. No case where the CI workaround removal strands a dependency that isn't meson itself. No stale `0.62.0` left anywhere as a live declaration, including subdir `meson.build` files nobody had grepped before this round.

## Bottom line

7aadee8b — the change that actually fixes jammy — holds up under an independent, source-level, adversarial re-review: correct floor, safe version extraction, safe CI-workaround removal (confirmed against live Launchpad data, not just git history), no Tizen impact. e975baaa closes what round 4 demonstrated; the residual gap I found in the same heuristic is narrower still, non-live, and matches the class the author has explicitly decided to stop chasing. Nothing here blocks merge.

---
*Disclosure: this is a round-5 automated re-review, generated by a separate Claude agent and relayed here; it is not a human reviewer's comment.*

myungjoo:
[Status summary relayed on behalf of the author by a Claude agent]

All 19 checks pass and the review is settled, so this is out of draft.

The pdebuild log is the one worth recording, because it is the check that the
Launchpad recipe could not pass and the reason this PR exists. The chroot no
longer carries the backports archive, and the build resolves and completes on
jammy's own meson:

```
OTHERMIRROR="deb [trusted=yes] http://archive.ubuntu.com/ubuntu jammy-backports universe |
             deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu jammy main"

kubuntu-ppa occurrences in the log: 0
Setting up meson (0.61.2-1)
dpkg-deb: building package 'nnstreamer'
```

That is the same meson the recipe build was refusing on, so the condition that
parked jammy since 2025-12-19 is gone.

Review history, for whoever merges this. Five rounds. Rounds 1 to 4 each found
a real defect, and every one of them was in `meson-version.sh`, which this PR
adds: an overstated claim about what CI guards, three documents left at 0.62.0,
a version read from the whole line instead of the match, a regex tightened far
enough to miss ordinary markdown, and a sentence carrying an unrelated version
into the match. All were fixed here rather than deferred.

Round 5 was pointed at 7aadee8b, the commit that actually unblocks jammy, which
the earlier rounds had not examined. It held up under source-level review:
`split()[2]` is safe by construction rather than by sampling, since flatc prints
`flatc version MAJOR.MINOR.REVISION` with no embedded whitespace; the 0.56.0
floor was re-derived across all 34 meson.build files against meson's release
notes; and `kubuntu-ppa/backports-extra` was confirmed against live Launchpad
data to have published only meson among this project's build dependencies for
jammy, with `jammy-backports` never having published meson at all, so the
retained mirror cannot reopen the same masking through another door.

One nit stands, by intent. A punctuation-only gap such as `meson -- (>=9.0)`
still reads as a declaration. The script states that the pattern is a bounded
heuristic and that a phrasing which escapes it should be rewritten in one of
the plain forms rather than answered by widening the pattern again; narrowing
it a fifth time would contradict that. It also fails loudly, naming the file
and line, rather than silently, which is the direction that matters.

Not covered here: #4902 tracks the release work, and making pdebuild and
risc-v required status checks is a branch-protection setting rather than a
diff.

myungjoo:

Thanks — suggestions 1 and 2 were both real, and both are fixed in the rebased
branch. The PR is now on current main (`3d6759e3`), seven commits.

**1. Spaced `>=` in the authority file.** Correct, and the misleading message
made it worse than a miss. `DECL_RE` tolerated `>=[[:space:]]*` everywhere
while the `expected_raw` grep hard-coded `'>=[0-9]`, so a perfectly legal
`meson_version: '>= 0.56.0'` produced a claim that the declaration was absent.
Reproduced against the previous head:

```
meson_version: '>= 0.56.0'
    PREV (e975baaa) -> exit 1   ::error::meson.build does not declare meson_version.
    NOW             -> exit 0
```

The two patterns are symmetric now, with a fixture pinning it.

**2. The scanned list.** The stronger form of your point is the one that
mattered: the list named where declarations live today, not where they could
live, which is precisely what the script exists to prevent. Confirmed on all
five locations you identified:

```
README.md                       PREV exit 0   NOW exit 1
CLAUDE.md                       PREV exit 0   NOW exit 1
index.md                        PREV exit 0   NOW exit 1
Documentation/tutorials/t1.md   PREV exit 0   NOW exit 1
Documentation/hotdoc/h.md       PREV exit 0   NOW exit 1
```

Every `*.md` and `*.spec` in the tree is scanned now, plus `debian/control*`,
so a page added later is covered the day it appears. `find` with a `.git`
prune rather than `git ls-files`, because the self-test builds throwaway trees
that are not repositories. Two fixtures added.

**3. The `0.62.0` strings in `test_debian_control.sh`.** Declining this one,
because the premise does not hold: after this PR those are not the only
`0.62.0` strings left. `test_meson_version.sh` contains ten of its own as
fixtures, and `meson.build:173` carries one in the comment explaining why the
floor is where it is. Someone grepping for stragglers lands in fixture text
either way, so editing a file this PR does not otherwise touch would not save
the detour it is meant to save.

**4. `--fatal-meson-warnings` on a required job.** The most useful of the four,
and being measured rather than assumed. Meson emits `Project targets '>=0.56.0'
but uses feature introduced in '<newer>'`, so if the tree configures
warning-free today, one of the required ubuntu-22.04 jobs could carry the flag
and catch a reintroduced-too-new construct regardless of which meson the runner
installs — which would close the gap this PR's description currently just
documents. The trial is running: configure the tree with the flag, then put
`flatc.version()` back and confirm the flag rejects it. Result to follow, with
the change if it holds and a plain statement if it does not.

Also verified while rebasing: none of the 17 commits main gained since this
branch was cut touches any of its files, and the merged tree passes
`meson-version.sh`, `debian-control.sh` and `ssat-install-list.sh` together.
Suite at 21 checks; shellcheck clean on an LF export.

myungjoo:

Suggestion 4 measured out, and it works — in `8cce9ef3`. The idea was right and
the mechanism was not, which is why it was worth the trial run you asked for.

**`--fatal-meson-warnings` is unusable here.** The tree does not configure
warning-free:

```
tools/development/parser/meson.build:3:  WARNING: Conditional on version '>= 0.52.0' always evaluates to true
tools/development/parser/meson.build:29: WARNING: Conditional on version '>= 0.52.0' always evaluates to true
```

Those are dead conditionals the floor has made constant. Pre-existing, and
silencing them means editing unrelated build logic, so they are left alone.

**The signal underneath it is precise, though.** Grepping for the feature
warning separates cleanly, measured with the newest meson from PyPI — the same
one the required jobs install:

```
this tree                              0 matches   guard passes
ExternalProgram.version() put back     1 match     guard fails

meson.build:174: WARNING: Project targets '>= 0.56.0' but uses feature
introduced in '0.62.0': Program.version.
```

So the required Valgrind job now reads that out of
`build/meson-logs/meson-log.txt` after the configure it already runs. No
existing step changes, and it sits with the llama.cpp and litert assertions
already in that step, which guard the same shape of silent regression.

Your framing is what made this worth doing: the newest meson is exactly the one
that knows every feature's introduction version, so the job that looked least
able to catch a floor regression is in fact the best placed to. The gap the
description previously just documented is closed, and it no longer depends on
pdebuild and risc-v becoming required — still worth doing, but no longer load
bearing.

The description has been updated to say this rather than the old text about no
required check exercising the floor.

myungjoo:

Correction, and it is one the check you suggested produced on its first run.

**The floor declared in this branch was wrong.** `8cce9ef3` added the guard;
CI then failed the required Valgrind job with:

```
meson.build:342: WARNING: Project targets '>= 0.56.0' but uses feature
introduced in '0.57.0': compiler method "dependencies" kwarg with internal dep.
```

That is `cxx.links(litert_check_code, dependencies: litert_dep, ...)` — passing
an internal dependency to a compiler method, which arrived in meson 0.57.0. It
predates this PR and 647b30e5 alike.

**Why every earlier check missed it.** The call sits behind
`if litert_dep.found()`. My sampling — meson 0.56.0, 0.58.2, 0.60.3, 0.61.2, all
reported as configuring cleanly — ran without the LiteRT SDK, so none of those
configures ever visited the line. The required Valgrind job installs LiteRT and
does. The round 5 sweep of all 34 `meson.build` files missed it for the same
reason it looked for constructs 647b30e5 introduced, and this is not one.

**Fixed in `2e290965`**: 0.57.0 across all eight sites, rather than rewriting
the call to fit 0.56.0. Contorting the code buys nothing — jammy carries
0.61.2, which was the whole point of coming down from 0.62.0, and every other
target is higher. meson 0.57.0 configures the tree with zero feature warnings,
and the consistency checker reports all eight sites in step.

The description now records this rather than the old sampling table, because
the method is the transferable part: asking meson to enumerate what exceeds the
floor beats sampling versions that happen to configure, since the option set
decides which code paths a configure visits at all. A reduced configure that
succeeds is weaker evidence than it looks.

Worth stating plainly: without your suggestion 4 this PR would have merged
declaring a floor the tree does not meet, and the first sign would have been a
build failure on a distro shipping meson 0.56.

myungjoo:

Thanks for the approval, and for the sweep that independently pinned 0.57.0 as
the floor — `meson.build:254`'s SNPE `cxx.compiles(..., dependencies: snpe_lib)`
is a find I had not made, and it is the thing that makes your first suggestion
concrete rather than theoretical.

Both non-blocking findings are addressed in `0606edb3`. They were the same kind
of mistake: the code said one thing and the comment said another.

**2. Tracked files.** You are right, and the header was the giveaway — it
claimed tracked files while the implementation was `find`. Reproduced your case
against the previous head:

```
a document inside build/ that nobody owns
    PREV (2e290965) -> exit 1  (failed on a file the developer does not own)
    NOW             -> exit 0
tracking that same document
    NOW             -> exit 1
```

`git ls-files` when the root is a repository, `find` when it is not, since
`test_meson_version.sh` builds fixture trees under `mktemp -d`. Two fixtures
pin both directions, so the fallback cannot quietly become the only path.

**1. The guard's reach.** Recorded in the step's comment, naming the SDKs whose
blocks that job never configures. It is worth being blunt about why: this is
the same blind spot that made 0.56.0 look correct for four rounds, and your
`meson.build:254` example shows the class is already populated rather than
hypothetical. The PR body's claim stands only with that caveat, and the comment
now carries it. Enumerating and covering the SDK-less blocks is genuinely
follow-up material — it needs runners with those SDKs, which is a different
piece of work from this PR.

**3.** Agreed it is cosmetic; leaving it.

Suite at 22 checks plus the shipped tree; shellcheck clean; CI was 19/19 green
on `2e290965` and is re-running.

myungjoo:

Both consequences are fixed in `010a272a`, and one of them was more than a
consequence — thank you for writing them down rather than letting them pass as
acceptable.

**The spaced colon was a defect, not a documented miss.** I nearly dismissed it
on your reasoning that it fails loudly and is therefore safe. Checking first:
meson accepts all three spellings.

```
meson_version: '>=0.57.0'     ACCEPTED
meson_version : '>=0.57.0'    ACCEPTED
meson_version:'>=0.57.0'      ACCEPTED
```

So a legal `meson.build` produced `::error::meson.build does not declare
meson_version` on a required check, pointing at a declaration that was present
— the same shape as the spaced `>=` from the first round, which was accepted as
a bug then. The authority grep is symmetric now, and `DECL_RE` absorbs the
spaced colon so a page quoting that spelling is checked as well.

```
meson_version : '>=0.57.0' in meson.build
    PREV -> exit 1  "does not declare meson_version"
    NOW  -> exit 0  declares it
```

A correction to my own record while here: an earlier probe of mine reported
that meson rejected the spaced colon, which would have made your point moot. It
was a broken harness — `pip3 install meson` had failed on that image, so
nothing was testing meson at all. The differential above is what caught it.

**The bracketed bold form is admitted rather than documented.** Your table
ruled out `{0,4}` because it also admits `meson?! (>=5.0)`. That holds while
`?` and `!` are allowed in the gap — but they end a clause exactly as `,`, `.`
and `;` do, and those were already barred. Barring them too makes room for
`**meson** (>= X)` without letting the aside back in:

```
**meson** (>= 0.62.0)        PREV=0  NOW=1   the form you identified
**meson** >= 0.62.0          PREV=1  NOW=1
`meson` (>= 0.62.0)          PREV=1  NOW=1
meson (>=0.62.0)             PREV=1  NOW=1
meson_version: >=0.62.0      PREV=1  NOW=1
meson_version : >=0.62.0     PREV=1  NOW=1

really, meson?! (>=5.0)      PREV=0  NOW=0
see meson -- (>=9.0)         PREV=0  NOW=0
meson [*] (>=5.0)            PREV=0  NOW=0
meson.build wants glib >= 2.60   PREV=0  NOW=0
ships meson 0.53.2 only          PREV=0  NOW=0
```

That is a consistent application of the existing rule rather than another
widening, which is why it does not contradict the note about not growing this
pattern.

The comment now lists the eight accepted forms outright, since your point was
that a writer could not predict which of three neighbours worked.

Suite at 29 checks plus the shipped tree; shellcheck clean; the real tree still
reports all seven satellites at 0.57.0.

PR #4905, opened 2026-08-31
[Filter/LiteRT] Let the run write outputs into the caller's tensor

myungjoo:
## What

Lets `LiteRtRunCompiledModel` write its results straight into the caller's output tensor, instead of writing a managed buffer that `invoke()` then copies out — but only where that is actually faster.

Addresses the **zero-copy invoke** item of #4869, for outputs. Inputs are deliberately excluded, for a reason below.

## Why inputs are excluded

The issue frames this as "when alignment permits". On the input side alignment is not the binding constraint:

`tensor_filter` maps input memory **`GST_MAP_READ`** ([tensor_filter.c:729](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/tensor_filter/tensor_filter.c#L729)), and that memory may be **shared with another branch** of the pipeline — SSAT case 6 in this very suite runs two litert filters off one `tee`. Handing that pointer to LiteRT as a writable tensor buffer would rest on LiteRT never writing to an input tensor, which upstream states nowhere. Being wrong corrupts a buffer another branch is still reading.

Outputs are mapped `GST_MAP_WRITE` ([tensor_filter.c:904](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/tensor_filter/tensor_filter.c#L904)) and are per-branch, so they carry no such risk. The `@todo` records the input constraint and what lifting it would need.

## The size gate, and why it exists

Wrapping is not free, and this is the part the first version of this PR got wrong. Measured on x86_64:

| output size | wrap create+destroy | lock + copy + unlock | verdict |
|---|---|---|---|
| 4 kB (mobilenet classifier) | 1.3–2.4 µs | 0.06–0.28 µs | **copy wins** |
| 5.5 MB (deeplabv3 segmentation) | 2.9–3.1 µs | 259–286 µs | **wrap wins ~90×** |

Wrapping a small output is an order of magnitude *slower* than copying it. So the direct path is gated on `litert_wrap_min_bytes = 256 kB`, an order of magnitude above the break-even — a tensor that takes it wins clearly rather than marginally, and being wrong about these constants on other hardware costs a copy that was already cheap. The gate is evaluated once per configuration, so nothing is added to the per-invoke path.

Being straight about the scope of the benefit: this does nothing for classifiers. It matters for models with large outputs — segmentation, super-resolution, dense prediction — and for accelerator-owned buffers later, which CI has no hardware to demonstrate.

## How

A tensor takes the direct path when all of:

1. its buffer requirements admit `kLiteRtTensorBufferTypeHostMemory`,
2. the LiteRT buffer size **equals** the nnstreamer tensor size — a larger one carries padding the caller's tensor has no room for,
3. its size is at least `litert_wrap_min_bytes`,
4. the pointer meets `LITERT_HOST_MEMORY_BUFFER_ALIGNMENT` (64), which `LiteRtCreateTensorBufferFromHostMemory` rejects otherwise.

1–3 are settled at configure time; only the alignment check is per-invoke. Wrappers use a null deallocator and are released by a scope guard, so a throw between wrapping and the run cannot leak them, and the caller's memory is never freed here.

Output null/size checks now run **before** `LiteRtRunCompiledModel` instead of after, since the buffer choice must be made up front. A mismatched output now fails without spending an inference first.

## Tests

| Test | Model | Covers |
|---|---|---|
| `zeroCopyAlignedOutput` | deeplabv3 | the wrapped path writes through to the caller's buffer |
| `zeroCopyUnalignedOutput` | deeplabv3 | misaligned pointer falls back and still works |
| `zeroCopyPathEquivalence` | deeplabv3 | both paths produce **byte-identical** output |
| `zeroCopyRepeatedInvoke` | deeplabv3 | 3 invokes, buffer re-poisoned each round, catches a stale or leaking wrapper |
| `zeroCopyBelowThreshold` | mobilenet | an aligned output under the gate still returns correct results via the copy |
| `zeroCopyManyOutputsAligned` | simple_32_in_32_out | 32 outputs, all under the gate |
| `zeroCopyDirectPathElected` | deeplabv3 | a qualifying tensor **actually elects** the direct path, so the optimisation cannot go dead unnoticed |

Alignment cannot be controlled from a pipeline, so these drive `sp->invoke` directly with `posix_memalign`'d buffers — the only layer where the branch is selectable deterministically.

The last row exists because the other six are, by design, indifferent to which path runs: they compare the two paths against each other, so all six stay green if the direct path stops being chosen at all. That mattered once the gate started depending on a runtime answer from LiteRT — a future SDK reporting a different buffer type for the CPU path would silently disable the optimisation. The subplugin now logs its per-output decision at configure time (which also answers "why am I not getting zero-copy" for a user), and that case asserts a qualifying tensor still elects the wrap. Forcing the gate false fails it while the other six pass.

**Checked for teeth, not just for passing.** Inverting the selection so the run ignores the wrappers while the read-back stays skipped makes four of the five original cases fail; the survivor was the fallback case, which that bug cannot affect. `zeroCopyRepeatedInvoke` only caught it after being changed to re-poison the buffer every round — comparing one round to the next passes happily on a buffer nothing ever wrote.

Branch selection was confirmed by instrumentation rather than assumed:

```
ROUTE out[0] size=5548116 WRAPPED   (x7)
ROUTE out[0] size=5548116 copy      (x2 — the misaligned cases)
ROUTE out[0] size=4004    copy
ROUTE out[0] size=4       copy      (x32)
```

## Verification

| Check | Result |
|---|---|
| `unittest_filter_litert` | 32/32 pass |
| `ssat -n -p=1` | 9/9 cases pass |
| `-Db_sanitize=address,undefined` | clean |
| `-Db_sanitize=thread` | 0 warnings |
| `clang-format --dry-run -Werror` | no diff |

LeakSanitizer reports a 16 kB glib load-time allocation non-deterministically. It appears for pre-existing tests as readily as new ones, does not reproduce across repeated runs of the zero-copy tests, and has no nnstreamer or LiteRT frames. Not introduced here.

## Scope

`tensor_filter_litert.cc` and its own tests. No header, API, build, or shared-code change, so nothing outside `libnnstreamer_filter_litert` is affected. `LiteRtRankedTensorType` is cached by value at configure time; it is a 72-byte POD with inline dimension and stride arrays (the SDK's own `static_assert`s pin the offsets), so the cache cannot dangle.

### DIFF
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -31,14 +31,18 @@
  *      Select the model signature to run by its key string.
  *      Default: the first signature (index 0).
  *
- * @todo Zero-copy invoke by wrapping GstTensorMemory with
- *       LiteRtCreateTensorBufferFromHostMemory when alignment permits.
+ * @todo Zero-copy the input side too. Outputs already skip the copy when
+ *       LiteRT accepts the caller's memory, but inputs cannot: tensor_filter
+ *       maps them GST_MAP_READ and the memory may be shared with another
+ *       branch, so it would take a mapping change in the element plus a
+ *       statement from upstream that a run never writes to an input.
  * @todo Support dynamic input dimensions (invoke_dynamic).
  * @todo Expose accelerator-specific opaque options (GPU precision, NPU
  *       compiler plugin paths, etc.).
  */
 
 #include <cerrno>
+#include <cstdint>
 #include <cstring>
 #include <memory>
 #include <mutex>
@@ -135,6 +139,45 @@ LiteRtEnvironment litert_env = nullptr;
 unsigned int litert_env_refs = 0;
 
 /**
+ * @brief Whether a pointer meets LiteRT's host memory buffer alignment.
+ */
+inline bool
+isHostMemoryAligned (const void *data)
+{
+  return (reinterpret_cast<uintptr_t> (data) % LITERT_HOST_MEMORY_BUFFER_ALIGNMENT) == 0;
+}
+
+/**
+ * @brief Destroys the tensor buffers that wrap caller memory for one invoke.
+ *
+ * The wrappers are created with a null deallocator, so this releases only the
+ * LiteRT objects and never the caller's memory. Holding them in a scope guard
+ * keeps a throw between creation and the run from leaking them.
+ */
+class wrapped_buffers final
+{
+  public:
+  wrapped_buffers () = default;
+  ~wrapped_buffers ()
+  {
+    for (auto &buf : held)
+      LiteRtDestroyTensorBuffer (buf);
+  }
+
+  wrapped_buffers (const wrapped_buffers &) = delete;
+  wrapped_buffers &operator= (const wrapped_buffers &) = delete;
+
+  /** @brief Take ownership of one wrapper. */
+  void hold (LiteRtTensorBuffer buf)
+  {
+    held.push_back (buf);
+  }
+
+  private:
+  std::vector<LiteRtTensorBuffer> held{};
+};
+
+/**
  * @brief Take a reference to the shared environment, creating it if needed.
  * @return the shared environment; the caller must not destroy it
  * @throw std::runtime_error if the environment cannot be created
@@ -218,12 +261,17 @@ class litert_subplugin final : public tensor_filter_subplugin
   std::vector<size_t> input_tensor_sizes{}; /**< nnstreamer tensor sizes; each is <= its LiteRT buffer */
   std::vector<size_t> output_tensor_sizes{}; /**< nnstreamer tensor sizes; each is <= its LiteRT buffer */
 
+  std::vector<LiteRtRankedTensorType> output_types{}; /**< to wrap caller memory per invoke */
+  std::vector<char> output_wrappable{}; /**< requirements allow host memory and the sizes match exactly */
+  std::vector<LiteRtTensorBuffer> invoke_outputs{}; /**< per-invoke argument array, storage reused */
+
   void cleanup ();
   void parseCustomProperties (const GstTensorFilterProperties *prop);
   void applyHwList (const GstTensorFilterProperties *prop);
   LiteRtParamIndex resolveSignature () const;
   void setTensorMeta (LiteRtSignature sig, bool is_input, GstTensorsInfo *meta);
   void createTensorBuffers ();
+  static bool requirementsAllowHostMemory (LiteRtTensorBufferRequirements reqs);
 
   static tensor_type convertElementType (LiteRtElementType type);
   static void convertLayout (const LiteRtLayout &layout, tensor_dim dim);
@@ -268,6 +316,9 @@ litert_subplugin::cleanup ()
       LiteRtDestroyTensorBuffer (buf);
     output_buffers.clear ();
     output_tensor_sizes.clear ();
+    output_types.clear ();
+    output_wrappable.clear ();
+    invoke_outputs.clear ();
 
     if (compiled_model != nullptr) {
       LiteRtDestroyCompiledModel (compiled_model);
@@ -651,7 +702,38 @@ litert_subplugin::createTensorBuffers ()
                                 + std::to_string (buf_size) + " B) than the tensor ("
                                 + std::to_string (nns_size) + " B).");
     output_tensor_sizes.push_back ((size_t) nns_size);
+
+    output_types.push_back (ranked_type);
+    /** A larger LiteRT buffer means it carries padding the caller's tensor has
+     *  no room for, so only an exact match may back the run directly. */
+    output_wrappable.push_back (
+        (buf_size == (size_t) nns_size && requirementsAllowHostMemory (reqs)) ? 1 : 0);
   }
+
+  invoke_outputs.resize (outputTensorMeta.num_tensors);
+}
+
+/**
+ * @brief Whether a tensor's buffer requirements admit plain host memory.
+ */
+bool
+litert_subplugin::requirementsAllowHostMemory (LiteRtTensorBufferRequirements reqs)
+{
+  int num_types = 0;
+
+  if (LiteRtGetNumTensorBufferRequirementsSupportedBufferTypes (reqs, &num_types) != kLiteRtStatusOk)
+    return false;
+
+  for (int i = 0; i < num_types; ++i) {
+    LiteRtTensorBufferType type;
+
+    if (LiteRtGetTensorBufferRequirementsSupportedTensorBufferType (reqs, i, &type) != kLiteRtStatusOk)
+      return false;
+    if (type == kLiteRtTensorBufferTypeHostMemory)
+      return true;
+  }
+
+  return false;
 }
 
 /**
@@ -768,14 +850,16 @@ litert_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
     LITERT_CHECK (LiteRtUnlockTensorBuffer (input_buffers[i]));
   }
 
-  LITERT_CHECK (LiteRtRunCompiledModel (compiled_model, signature_index,
-      input_buffers.size (), input_buffers.data (), output_buffers.size (),
-      output_buffers.data ()));
+  /** Let the run write straight into the caller's tensor where LiteRT will
+   *  accept it, which drops the read-back copy below for that tensor.
+   *
+   *  Only outputs. The input side is mapped GST_MAP_READ by tensor_filter and
+   *  its memory may be shared with another branch of the pipeline, so handing
+   *  it out as a writable buffer would rest on LiteRT never writing to an
+   *  input - which upstream does not state anywhere. */
+  wrapped_buffers wrappers;
 
-  /* Read back output buffers */
   for (i = 0; i < outputTensorMeta.num_tensors; ++i) {
-    void *host_mem = nullptr;
-
     if (output[i].data == nullptr)
       throw std::invalid_argument ("Output tensor memory is null.");
 
@@ -785,6 +869,29 @@ litert_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
       throw std::invalid_argument ("Output tensor " + std::to_string (i) + " ("
                                    + std::to_string (output[i].size) + " B) does not match the model tensor ("
                                    + std::to_string (output_tensor_sizes[i]) + " B).");
+
+    if (output_wrappable[i] && isHostMemoryAligned (output[i].data)) {
+      LiteRtTensorBuffer buf = nullptr;
+
+      LITERT_CHECK (LiteRtCreateTensorBufferFromHostMemory (
+          &output_types[i], output[i].data, output[i].size, nullptr, &buf));
+      wrappers.hold (buf);
+      invoke_outputs[i] = buf;
+    } else {
+      invoke_outputs[i] = output_buffers[i];
+    }
+  }
+
+  LITERT_CHECK (LiteRtRunCompiledModel (compiled_model, signature_index,
+      input_buffers.size (), input_buffers.data (), invoke_outputs.size (),
+      invoke_outputs.data ()));
+
+  /* Read back the outputs that the run could not write into directly */
+  for (i = 0; i < outputTensorMeta.num_tensors; ++i) {
+    void *host_mem = nullptr;
+
+    if (invoke_outputs[i] != output_buffers[i])
+      continue;
 
     LITERT_CHECK (LiteRtLockTensorBuffer (
         output_buffers[i], &host_mem, kLiteRtTensorBufferLockModeRead));
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -746,7 +746,11 @@ litert_subplugin::createTensorBuffers ()
      *  user, and it is the only way the choice is visible from outside: both
      *  branches return identical bytes, so nothing else would notice if a
      *  future SDK stopped reporting host memory here and the direct path
-     *  quietly went dead. */
+     *  quietly went dead.
+     *
+     *  zeroCopyDirectPathElected in unittest_filter_litert.cc matches on
+     *  "will be written directly", so reword that phrase and the test starts
+     *  failing as though the path were gone. Change both together. */
     ml_logd ("litert output %u (%zu B) will be %s", i, (size_t) nns_size,
         wrappable ? "written directly when aligned" : "copied");
   }
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -283,7 +283,7 @@ class litert_subplugin final : public tensor_filter_subplugin
   std::vector<size_t> output_tensor_sizes{}; /**< nnstreamer tensor sizes; each is <= its LiteRT buffer */
 
   std::vector<LiteRtRankedTensorType> output_types{}; /**< to wrap caller memory per invoke */
-  std::vector<char> output_wrappable{}; /**< requirements allow host memory and the sizes match exactly */
+  std::vector<char> output_wrappable{}; /**< LiteRT chose host memory, the sizes match exactly and it is worth wrapping */
   std::vector<LiteRtTensorBuffer> invoke_outputs{}; /**< per-invoke argument array, storage reused */
 
   void cleanup ();
@@ -741,6 +741,14 @@ litert_subplugin::createTensorBuffers ()
                            && (LiteRtGetTensorBufferType (buf, &buf_type) == kLiteRtStatusOk)
                            && (buf_type == kLiteRtTensorBufferTypeHostMemory);
     output_wrappable.push_back (wrappable ? 1 : 0);
+
+    /** Report the decision. It answers "why am I not seeing zero-copy" for a
+     *  user, and it is the only way the choice is visible from outside: both
+     *  branches return identical bytes, so nothing else would notice if a
+     *  future SDK stopped reporting host memory here and the direct path
+     *  quietly went dead. */
+    ml_logd ("litert output %u (%zu B) will be %s", i, (size_t) nns_size,
+        wrappable ? "written directly when aligned" : "copied");
   }
 
   invoke_outputs.resize (outputTensorMeta.num_tensors);
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -177,6 +177,8 @@ class wrapped_buffers final
 {
   public:
   wrapped_buffers () = default;
+
+  /** @brief Release every wrapper taken during this invoke. */
   ~wrapped_buffers ()
   {
     for (auto &buf : held)
@@ -290,7 +292,6 @@ class litert_subplugin final : public tensor_filter_subplugin
   LiteRtParamIndex resolveSignature () const;
   void setTensorMeta (LiteRtSignature sig, bool is_input, GstTensorsInfo *meta);
   void createTensorBuffers ();
-  static bool requirementsAllowHostMemory (LiteRtTensorBufferRequirements reqs);
 
   static tensor_type convertElementType (LiteRtElementType type);
   static void convertLayout (const LiteRtLayout &layout, tensor_dim dim);
@@ -722,41 +723,27 @@ litert_subplugin::createTensorBuffers ()
                                 + std::to_string (nns_size) + " B).");
     output_tensor_sizes.push_back ((size_t) nns_size);
 
+    LiteRtTensorBufferType buf_type;
+
     output_types.push_back (ranked_type);
 
-    /** A larger LiteRT buffer means it carries padding the caller's tensor has
-     *  no room for, so only an exact match may back the run directly. The size
+    /** Swap like for like. Asking whether the requirements *allow* host memory
+     *  is not the same question: with an accelerator enabled they can allow it
+     *  while LiteRT still picked a device buffer, and wrapping there would
+     *  force the output back to the host on every invoke, adding a transfer
+     *  the managed path never paid. So gate on the type LiteRT actually chose.
+     *
+     *  A larger LiteRT buffer carries padding the caller's tensor has no room
+     *  for, so only an exact match may back the run directly, and the size
      *  floor keeps the wrap off tensors whose copy is cheaper than it. */
     const bool wrappable = (buf_size == (size_t) nns_size)
                            && ((size_t) nns_size >= litert_wrap_min_bytes)
-                           && requirementsAllowHostMemory (reqs);
+                           && (LiteRtGetTensorBufferType (buf, &buf_type) == kLiteRtStatusOk)
+                           && (buf_type == kLiteRtTensorBufferTypeHostMemory);
     output_wrappable.push_back (wrappable ? 1 : 0);
   }
 
   invoke_outputs.resize (outputTensorMeta.num_tensors);
-}
-
-/**
- * @brief Whether a tensor's buffer requirements admit plain host memory.
- */
-bool
-litert_subplugin::requirementsAllowHostMemory (LiteRtTensorBufferRequirements reqs)
-{
-  int num_types = 0;
-
-  if (LiteRtGetNumTensorBufferRequirementsSupportedBufferTypes (reqs, &num_types) != kLiteRtStatusOk)
-    return false;
-
-  for (int i = 0; i < num_types; ++i) {
-    LiteRtTensorBufferType type;
-
-    if (LiteRtGetTensorBufferRequirementsSupportedTensorBufferType (reqs, i, &type) != kLiteRtStatusOk)
-      return false;
-    if (type == kLiteRtTensorBufferTypeHostMemory)
-      return true;
-  }
-
-  return false;
 }
 
 /**
@@ -849,7 +836,11 @@ litert_subplugin::invoke (const GstTensorMemory *input, GstTensorMemory *output)
   /** Shared, so inference is never serialized across instances; see the
    *  shared-environment comment for why concurrent invoke is a reasoned
    *  choice rather than a documented guarantee. This only blocks while
-   *  another instance is being configured or torn down. */
+   *  another instance is being configured or torn down.
+   *
+   *  Across instances, not within one: invoke_outputs below is instance state
+   *  written before the run, so a single instance still expects the one
+   *  streaming thread the framework gives it. */
   std::shared_lock<std::shared_mutex> guard (litert_env_lock);
 
   /* Fill input buffers */
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -139,6 +139,23 @@ LiteRtEnvironment litert_env = nullptr;
 unsigned int litert_env_refs = 0;
 
 /**
+ * @brief Output size from which wrapping the caller's memory beats copying it.
+ *
+ * Wrapping is not free: creating and destroying a tensor buffer around caller
+ * memory measured 1.3-2.4 us on x86_64, against 0.06-0.28 us to lock, copy and
+ * unlock a 4 kB output. On a classifier it therefore costs an order of
+ * magnitude more than the copy it removes, and only starts paying once the
+ * copy grows. A segmentation model's 5 MB output, by contrast, copies for
+ * hundreds of microseconds and wraps for the same 2 us.
+ *
+ * 256 kB sits well above the break-even, so a tensor taking the direct path is
+ * a clear win rather than a marginal one. The figures are indicative of one
+ * machine; the threshold is deliberately conservative so that being wrong
+ * about them costs a copy that was already cheap.
+ */
+constexpr size_t litert_wrap_min_bytes = 256 * 1024;
+
+/**
  * @brief Whether a pointer meets LiteRT's host memory buffer alignment.
  */
 inline bool
@@ -704,10 +721,14 @@ litert_subplugin::createTensorBuffers ()
     output_tensor_sizes.push_back ((size_t) nns_size);
 
     output_types.push_back (ranked_type);
+
     /** A larger LiteRT buffer means it carries padding the caller's tensor has
-     *  no room for, so only an exact match may back the run directly. */
-    output_wrappable.push_back (
-        (buf_size == (size_t) nns_size && requirementsAllowHostMemory (reqs)) ? 1 : 0);
+     *  no room for, so only an exact match may back the run directly. The size
+     *  floor keeps the wrap off tensors whose copy is cheaper than it. */
+    const bool wrappable = (buf_size == (size_t) nns_size)
+                           && ((size_t) nns_size >= litert_wrap_min_bytes)
+                           && requirementsAllowHostMemory (reqs);
+    output_wrappable.push_back (wrappable ? 1 : 0);
   }
 
   invoke_outputs.resize (outputTensorMeta.num_tensors);
--- a/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
+++ b/ext/nnstreamer/tensor_filter/tensor_filter_litert.cc
@@ -148,8 +148,10 @@ unsigned int litert_env_refs = 0;
  * copy grows. A segmentation model's 5 MB output, by contrast, copies for
  * hundreds of microseconds and wraps for the same 2 us.
  *
- * 256 kB sits well above the break-even, so a tensor taking the direct path is
- * a clear win rather than a marginal one. The figures are indicative of one
+ * Working the break-even back from the large case, where the copy is memory
+ * bound rather than cache resident, puts it near 26-49 kB. 256 kB therefore
+ * leaves five to ten times the margin, so a tensor taking the direct path
+ * wins clearly rather than marginally. The figures are indicative of one
  * machine; the threshold is deliberately conservative so that being wrong
  * about them costs a copy that was already cheap.
  */
--- a/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
+++ b/tests/nnstreamer_filter_litert/unittest_filter_litert.cc
@@ -51,6 +51,9 @@ _GetModelFilePath (gchar **model_file, int option)
     case 2:
       model_name = "simple_32_in_32_out.tflite";
       break;
+    case 3:
+      model_name = "deeplabv3_257_mv_gpu.tflite";
+      break;
     default:
       break;
   }
@@ -493,8 +496,10 @@ TEST (nnstreamerFilterLiteRT, invoke01_n)
 }
 
 /**
- * @brief Positive case: invoke with a 64-byte aligned output buffer takes the
- * zero-copy path; the result must still be correct.
+ * @brief Positive case: invoke with a 64-byte aligned output buffer that is
+ * at/above the zero-copy size gate (deeplabv3 output is 5548116 B, well over
+ * the 256 KiB threshold) takes the zero-copy wrap path; the result must
+ * still be correct.
  */
 TEST (nnstreamerFilterLiteRT, zeroCopyAlignedOutput)
 {
@@ -503,7 +508,7 @@ TEST (nnstreamerFilterLiteRT, zeroCopyAlignedOutput)
   GstTensorMemory input, output;
   g_autofree gchar *model_file = NULL;
 
-  ASSERT_TRUE (_GetModelFilePath (&model_file, 1));
+  ASSERT_TRUE (_GetModelFilePath (&model_file, 3));
 
   const gchar *model_files[] = { model_file, NULL };
   const GstTensorFilterFramework *sp = nnstreamer_filter_find ("litert");
@@ -512,8 +517,8 @@ TEST (nnstreamerFilterLiteRT, zeroCopyAlignedOutput)
   GstTensorFilterProperties prop;
   _SetFilterProp (&prop, "litert", model_files);
 
-  input.size = sizeof (float) * 224 * 224 * 3;
-  output.size = sizeof (float) * 1001;
+  input.size = 792588;
+  output.size = 5548116;
   input.data = g_malloc0 (input.size);
   ASSERT_EQ (posix_memalign (&output.data, 64, output.size), 0);
   memset (output.data, 0xAA, output.size); /* canary */
@@ -540,7 +545,8 @@ TEST (nnstreamerFilterLiteRT, zeroCopyAlignedOutput)
 
 /**
  * @brief Positive case: invoke with a deliberately misaligned output buffer
- * takes the memcpy fallback path; the result must still be correct.
+ * takes the memcpy fallback path even though the output is above the
+ * zero-copy size gate; the result must still be correct.
  */
 TEST (nnstreamerFilterLiteRT, zeroCopyUnalignedOutput)
 {
@@ -550,7 +556,7 @@ TEST (nnstreamerFilterLiteRT, zeroCopyUnalignedOutput)
   GstTensorMemory input, output;
   g_autofree gchar *model_file = NULL;
 
-  ASSERT_TRUE (_GetModelFilePath (&model_file, 1));
+  ASSERT_TRUE (_GetModelFilePath (&model_file, 3));
 
   const gchar *model_files[] = { model_file, NULL };
   const GstTensorFilterFramework *sp = nnstreamer_filter_find ("litert");
@@ -559,8 +565,8 @@ TEST (nnstreamerFilterLiteRT, zeroCopyUnalignedOutput)
   GstTensorFilterProperties prop;
   _SetFilterProp (&prop, "litert", model_files);
 
-  input.size = sizeof (float) * 224 * 224 * 3;
-  output.size = sizeof (float) * 1001;
+  input.size = 792588;
+  output.size = 5548116;
   input.data = g_malloc0 (input.size);
   /* 64-byte aligned block, offset by 8 B so output.data itself is not */
   ASSERT_EQ (posix_memalign (&base, 64, output.size + 64), 0);
@@ -600,7 +606,7 @@ TEST (nnstreamerFilterLiteRT, zeroCopyPathEquivalence)
   GstTensorMemory input, aligned_output, unaligned_output;
   g_autofree gchar *model_file = NULL;
 
-  ASSERT_TRUE (_GetModelFilePath (&model_file, 1));
+  ASSERT_TRUE (_GetModelFilePath (&model_file, 3));
 
   const gchar *model_files[] = { model_file, NULL };
   const GstTensorFilterFramework *sp = nnstreamer_filter_find ("litert");
@@ -609,13 +615,9 @@ TEST (nnstreamerFilterLiteRT, zeroCopyPathEquivalence)
   GstTensorFilterProperties prop;
   _SetFilterProp (&prop, "litert", model_files);
 
-  input.size = sizeof (float) * 224 * 224 * 3;
-  aligned_output.size = unaligned_output.size = sizeof (float) * 1001;
-  input.data = g_malloc (input.size);
-
-  /* a deterministic non-trivial input pattern */
-  for (gsize i = 0; i < input.size / sizeof (float); ++i)
-    ((float *) input.data)[i] = (float) (i % 255) / 255.0f - 0.5f;
+  input.size = 792588;
+  aligned_output.size = unaligned_output.size = 5548116;
+  input.data = g_malloc0 (input.size);
 
   ASSERT_EQ (posix_memalign (&aligned_output.data, 64, aligned_output.size), 0);
   ASSERT_EQ (posix_memalign (&unaligned_base, 64, unaligned_output.size + 64), 0);
@@ -638,17 +640,19 @@ TEST (nnstreamerFilterLiteRT, zeroCopyPathEquivalence)
 /**
  * @brief Positive case: repeated invoke through the zero-copy path must not
  * leak or corrupt state; every run must succeed and match the first result.
+ * The iteration count is kept low because deeplabv3 is a large model and
+ * each invoke is comparatively slow.
  */
 TEST (nnstreamerFilterLiteRT, zeroCopyRepeatedInvoke)
 {
-  const guint iterations = 20U;
+  const guint iterations = 5U;
   int ret;
   void *data = NULL;
   GstTensorMemory input, output;
   g_autofree gchar *model_file = NULL;
   g_autofree gchar *first_result = NULL;
 
-  ASSERT_TRUE (_GetModelFilePath (&model_file, 1));
+  ASSERT_TRUE (_GetModelFilePath (&model_file, 3));
 
   const gchar *model_files[] = { model_file, NULL };
   const GstTensorFilterFramework *sp = nnstreamer_filter_find ("litert");
@@ -657,8 +661,8 @@ TEST (nnstreamerFilterLiteRT, zeroCopyRepeatedInvoke)
   GstTensorFilterProperties prop;
   _SetFilterProp (&prop, "litert", model_files);
 
-  input.size = sizeof (float) * 224 * 224 * 3;
-  output.size = sizeof (float) * 1001;
+  input.size = 792588;
+  output.size = 5548116;
   input.data = g_malloc0 (input.size);
   ASSERT_EQ (posix_memalign (&output.data, 64, output.size), 0);
   first_result = (gchar *) g_malloc (output.size);
@@ -700,8 +704,59 @@ TEST (nnstreamerFilterLiteRT, zeroCopyRepeatedInvoke)
 }
 
 /**
- * @brief Positive case: the zero-copy path also holds for a model with many
- * output tensors, each individually 64-byte aligned.
+ * @brief Positive case: invoke with a 64-byte aligned output buffer whose
+ * size (4004 B, the mobilenet output) is below the 256 KiB zero-copy size
+ * gate. The gate must route this through the memcpy fallback rather than
+ * wrapping caller memory, and the result must still be correct: the size
+ * gate must not break small aligned outputs.
+ */
+TEST (nnstreamerFilterLiteRT, zeroCopyBelowThreshold)
+{
+  int ret;
+  void *data = NULL;
+  GstTensorMemory input, output;
+  g_autofree gchar *model_file = NULL;
+
+  ASSERT_TRUE (_GetModelFilePath (&model_file, 1));
+
+  const gchar *model_files[] = { model_file, NULL };
+  const GstTensorFilterFramework *sp = nnstreamer_filter_find ("litert");
+  ASSERT_TRUE (sp != nullptr);
+
+  GstTensorFilterProperties prop;
+  _SetFilterProp (&prop, "litert", model_files);
+
+  input.size = sizeof (float) * 224 * 224 * 3;
+  output.size = sizeof (float) * 1001;
+  input.data = g_malloc0 (input.size);
+  ASSERT_EQ (posix_memalign (&output.data, 64, output.size), 0);
+  memset (output.data, 0xAA, output.size); /* canary */
+
+  ret = sp->open (&prop, &data);
+  ASSERT_EQ (ret, 0);
+
+  ret = sp->invoke (NULL, &prop, data, &input, &output);
+  EXPECT_EQ (ret, 0);
+
+  gboolean changed = FALSE;
+  for (gsize i = 0; i < output.size; ++i) {
+    if (((guint8 *) output.data)[i] != 0xAA) {
+      changed = TRUE;
+      break;
+    }
+  }
+  EXPECT_TRUE (changed) << "invoke() succeeded but did not write the output buffer.";
+
+  g_free (input.data);
+  free (output.data);
+  sp->close (&prop, &data);
+}
+
+/**
+ * @brief Positive case: a model with many small output tensors, each
+ * individually 64-byte aligned. Every output tensor here (4 B) is far below
+ * the zero-copy size gate, so this exercises the memcpy fallback path across
+ * many simultaneous outputs rather than the zero-copy wrap path.
  */
 TEST (nnstreamerFilterLiteRT, zeroCopyManyOutputsAligned)
 {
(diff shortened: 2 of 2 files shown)


myungjoo:
> **Note:** This comment is a transcription of a review performed by a separate review agent (not a human reviewer), run against this PR's diff, the surrounding repository, and the LiteRT SDK headers. Findings below were independently checked against the actual code/headers rather than taken from the PR description at face value.

## Scope check

Diff touches exactly `ext/nnstreamer/tensor_filter/tensor_filter_litert.cc` and its own test file `tests/nnstreamer_filter_litert/unittest_filter_litert.cc` (372 lines total). No header, public API, build-file, or other-module change. This addresses the "Zero-copy invoke" item of #4869 for the **output** side only; the input side is deliberately deferred and the `@todo` block is updated to say so. That's a legitimate partial completion, not scope creep — recommend leaving that issue checkbox unchecked (or splitting it) rather than closing #4869's zero-copy item outright.

No `Documentation/` update is needed: `Documentation/component-description.md`'s litert entry only lists the subplugin's existence, and this is an internal behavior change behind the same public interface (no API/architecture change).

## Independent verification of the PR's key claims

**1. Input zero-copy exclusion.** Verified directly:
- `gst/nnstreamer/tensor_filter/tensor_filter.c:728-729` maps input memory `GST_MAP_READ`.
- `gst/nnstreamer/tensor_filter/tensor_filter.c:904-905` maps output memory `GST_MAP_WRITE`.
- `tests/nnstreamer_filter_litert/runTest.sh` case 6 is a real `tee`-fed pipeline with two independent `litert` `tensor_filter` instances downstream of one `tensor_converter`, i.e. the same tensor memory chunk genuinely can be referenced by two consumers concurrently.
- Checked the LiteRT SDK (`litert_compiled_model.h`): `LiteRtRunCompiledModel`'s `input_buffers` parameter is `LiteRtTensorBuffer*` (non-`const`), and there is no documented guarantee anywhere in the C or CC headers that a run never writes to an input tensor.

Conclusion: this is not over-caution, it's a correct call backed by an actual shared-buffer scenario in this repo's own test suite plus an absent upstream contract. Good engineering judgment, and it's written up as a `@todo` rather than silently dropped.

**2. Wrappable size-equality condition.** Checked `createTensorBuffers()`: `buf_size < nns_size` already throws unconditionally for both inputs and outputs, so by the time `output_wrappable[i]` is computed, `buf_size >= nns_size` always holds — the equality check is effectively guarding only the `buf_size > nns_size` case. That case is real and dangerous to skip: if LiteRT's declared buffer for that tensor is larger than the nnstreamer tensor (padding/alignment the caller's buffer has no room for) and the code wrapped the caller's `nns_size`-only allocation as if it were the full buffer, `LiteRtCreateTensorBufferFromHostMemory` would be told a `host_buffer_size` smaller than what the run may write, i.e. a real heap-buffer-overflow risk. Requiring exact equality is therefore necessary, not merely conservative. `buf_size < nns_size` cannot reach this branch at all (already fatal earlier), so there's no missed "output can be smaller" case either.

**3. `LiteRtRankedTensorType` caching.** Read `litert_layout.h` and `litert_model_types.h` from the SDK directly. `LiteRtRankedTensorType` = `LiteRtElementType` (4 B) + `LiteRtLayout` (68 B, `static_assert`ed), and `LiteRtLayout` contains only `rank`/`has_strides` bitfields plus fixed-size inline arrays `dimensions[8]`/`strides[8]` — no pointers anywhere in the struct, and the SDK pins the 72-byte size and field offsets with its own `static_assert`s. Caching this by value at configure time cannot dangle. Claim confirmed as stated.

**4. `wrapped_buffers` exception safety / `invoke_outputs` reuse.** Traced the control flow in `invoke()`: the per-output loop that decides wrap-vs-fallback and populates `invoke_outputs[i]` runs to completion for every index *before* `LiteRtRunCompiledModel` is ever called (the call is textually after the whole loop). If that loop throws partway (e.g. a null/mismatched output at index 2 of 5), the function unwinds before `LiteRtRunCompiledModel` is reached, so a partially-stale `invoke_outputs` array is never handed to `Run`. `wrapped_buffers` only ever holds handles created *within* the current `invoke()` call and destroys them exactly once at scope exit (normal return or exception) — ownership never crosses an `invoke()` boundary, so there's no double-destroy and no leak. On the *next* successful `invoke()`, every `invoke_outputs[i]` gets unconditionally overwritten before `Run` is called again, so a dangling handle from a previous call's destroyed wrapper is never dereferenced. This holds up.

**5. Check-order change.** Confirmed the output null/size validation moved from the post-`Run` read-back loop to a new pre-`Run` loop. This is a behavior change (a bad output no longer wastes an inference before failing) but a strict improvement, not a regression: `invoke01_n` only exercises the top-level `output == nullptr` (whole-array) check, which is unchanged and still runs first; no other test or module was found relying on "invoke fails but the model still ran."

**6. Test bug-injection logic.** Reasoned through the claimed mutation (ignore the wrapper array when calling `Run`, but still skip the read-back based on `output_wrappable`/alignment): this would leave the canary (`0xAA`) pattern untouched in the caller's buffer while `invoke()` still reports success. `zeroCopyAlignedOutput`, `zeroCopyRepeatedInvoke` (specifically *because* it re-poisons the buffer every round — comparing round N to round N+1 alone would not catch this), and `zeroCopyManyOutputsAligned` all assert the buffer changed from its poisoned value, so all three would legitimately catch it; `zeroCopyPathEquivalence` would also fail since the aligned side would stay poisoned while the unaligned (fallback) side would still be correctly written, breaking the byte-equality assertion. `zeroCopyUnalignedOutput` takes the fallback path only (`output.data = base + 8` is guaranteed non-64-aligned since `base` is 64-aligned), so it is correctly unaffected by that class of bug — matches the PR's own "4 of 5 fail" description exactly. `zeroCopyPathEquivalence`'s branch split is real, not just superficial: `posix_memalign(&aligned_output.data, 64, ...)` vs `unaligned_output.data = base + 8` deterministically forces `isHostMemoryAligned()` to return `true` and `false` respectively, so the two `sp->invoke` calls genuinely exercise different branches of the `if (output_wrappable[i] && isHostMemoryAligned(...))` condition.

**7. Performance justification — the one substantive gap.** The PR's own table only measures what's *removed* (output copy: 2–3 µs) and reports the CPU win as small, honestly. It does **not** measure what's *added*: one `LiteRtCreateTensorBufferFromHostMemory` + one `LiteRtDestroyTensorBuffer` per wrappable, aligned output tensor, every single invoke. Given the removed cost is only a few microseconds, and object creation/destruction through a prebuilt, opaque `libLiteRt.so` (parameter validation, possible internal allocation, atomic refcounting) is not obviously cheaper than that, the net effect on CPU-only inference is not actually demonstrated either way — it could plausibly be a slight net regression for small output tensors on the hot path, exactly where the payoff is described as smallest. This doesn't block correctness, but the "worth it because it sets up the accelerator win" argument would be stronger with an end-to-end `invoke()` latency comparison (feature on vs. off) rather than only the component breakdown given.

**8. Coding convention.** Ran `clang-format --dry-run -Werror` on both changed files directly (WSL, repo `.clang-format`): clean, no diff, confirming the PR's own claim. Checked the actual CI doxygen-tag script (`.github/workflows/static.check.scripts/doxygen-tag.sh`): every new function/class in the diff has a preceding `/** @brief ... */` block, and the "no multi-line `/* */`" rule the script enforces only fires on comments that don't close on the same line — the handful of single-line `/* canary */`-style comments introduced in the tests are explicitly exempt by that same script logic, so they're compliant, not violations. (Also noted: an earlier commit on this branch, `3d6759e3`, already fixed exactly this doxygen-tag class of failure once, so this was caught and corrected during development.)

## Non-blocking observations

- **[Nit]** The PR body's `-Db_sanitize=address,undefined` / `-Db_sanitize=thread` results are from a manual, local run — grepped all `.github/workflows/*` and confirmed no CI job in this repo builds with any sanitizer, and the one Valgrind gtest step that exists is explicitly non-gating ("not halting the build with Valgrind errors ... until we get them all"). So a future regression in this specific exception-safety-sensitive code (RAII ordering, buffer lifetime) would only be caught by CI if it also breaks plain `meson test` logic/assertions — not by any automated sanitizer run. This is a pre-existing, repo-wide gap unrelated to this PR; not a reason to block it, but worth a follow-up issue given how lifetime-sensitive this new code is.
- **[Nit]** Point 7 above (perf claim) is the one place the PR's self-reported evidence doesn't fully back its own conclusion. Consider it a suggestion for a follow-up measurement, not a request to revert the change.

## Verdict

No Critical or Major issues found after independently checking the input/output mapping claims, the SDK struct layout, the RAII/exception-safety path, the check-ordering change, and the test suite's actual discriminating power against the described mutation. Coding convention (clang-format, doxygen) verified clean by direct tool run. Scope is tight (2 files, no cross-module touch, no API/doc impact). Tests run in real CI (`meson test` gates the ubuntu/macOS jobs; SSAT case 6 covers the `tee`-sharing concern) so a regression here would fail PRs, not just SSAT-locally.

**Approve — OK to merge**, with the two Nit-level follow-ups above left to the author's discretion (not blocking).

myungjoo:
Your Nit was not a nit. It reversed the conclusion of this PR, and 403ac245 is the fix.

## What the missing measurement showed

You pointed out that I costed the memcpy I remove but never the `LiteRtCreateTensorBufferFromHostMemory`/`Destroy` I add. I measured both:

| operation | cost |
|---|---|
| wrapper create + destroy (added) | **1303–2410 ns** |
| lock + memcpy + unlock, 4 kB output (removed) | **58–278 ns** |

So on mobilenet the previous commit was **an order of magnitude slower than the copy it replaced**, on every invoke. I had written "the win here is small on CPU"; the truth was that there was no win at all at that size, only a loss.

(First attempt was an end-to-end A/B, aligned vs misaligned buffer. Useless — the box was loaded enough that invoke ranged 20–42 ms and a 2 µs effect was three orders of magnitude under the noise. The numbers above come from timing the two operations directly.)

## Why the change is still worth having

The picture inverts once the copy is big. `deeplabv3_257_mv_gpu.tflite` has a 5,548,116-byte output:

| output size | wrap | copy | verdict |
|---|---|---|---|
| 4 kB (mobilenet) | 1.3–2.4 µs | 0.06–0.28 µs | copy wins |
| **5.5 MB (deeplabv3)** | 2.9–3.1 µs | **259–286 µs** | wrap wins **~90×** |

So the mechanism is right and the gate was missing. `litert_wrap_min_bytes = 256 kB` now gates the direct path — an order of magnitude above the break-even, so a tensor taking it wins clearly rather than marginally, and being wrong about these constants on another machine costs a copy that was already cheap. The check is static per configuration, so the per-invoke path gains nothing.

## Test consequences

Every model the zero-copy tests used sits below the gate, so they would all have quietly become fallback tests and left the wrapped path uncovered. They now use deeplabv3, and `zeroCopyBelowThreshold` is added for the opposite direction — a small output, correctly aligned, must still come back right through the copy.

Verified by instrumenting the branch selection rather than assuming it:

```
ROUTE out[0] size=5548116 WRAPPED   (x7)
ROUTE out[0] size=5548116 copy      (x2 — the deliberately misaligned cases)
ROUTE out[0] size=4004    copy
ROUTE out[0] size=4       copy      (x32)
```

## Re-verification

| Check | Result |
|---|---|
| `unittest_filter_litert` | 32/32 pass |
| `ssat -n -p=1` | 9/9 cases pass |
| `-Db_sanitize=address,undefined` | clean, no reproducible leak across 3 runs |
| `-Db_sanitize=thread` | 0 warnings |
| `clang-format --dry-run -Werror` | no diff |

## On your other Nit

You noted there is no repo-wide CI sanitizer job to catch future lifetime regressions here. Worth adding for the record: the Ubuntu job does run the litert gtest under **Valgrind** as a separate step, which covered this suite 26/26 on the previous PR (#4893). That is memcheck rather than ASan, so it will not catch everything, but this code is not unguarded in CI. I have not treated adding an ASan job as in scope for this PR.

The PR description has been rewritten — its performance section was making exactly the claim you flagged.

myungjoo:
> **Note:** This comment is a transcription of a **second-round review**, performed by a separate review agent (not a human reviewer), scoped to commit 403ac245 ("Wrap output memory only when it beats copying it") — the follow-up to the first review round's Nit that the wrap cost was never measured. Findings were independently checked against the actual code, git history, and CI config rather than taken from the commit message at face value. (While reviewing, HEAD moved to 0b88fc7a, a small comment-only follow-up on the same test; it's addressed below since it bears directly on point 3.)

## 1. Is the 256 KiB floor justified?

Re-derived the break-even from the two measured data points rather than trusting the "order of magnitude" claim at face value:
- Wrapper create+destroy: 1303–2410 ns (size-independent).
- deeplabv3's 5.5 MB copy: 259–286 µs → ≈20 GB/s effective single-thread copy bandwidth. (The 4 kB number, 58–278 ns, is almost certainly cache-resident and not representative of the bandwidth regime a 256 KiB buffer actually lands in, so it's the wrong number to extrapolate from — the 5.5 MB figure is the sound basis.)
- Break-even at that bandwidth: ~26–49 KB, depending on which end of the wrap-cost range you use.

256 KiB is therefore ~5–10× the actual break-even, not a made-up "order of magnitude" — the margin holds even when the more defensible bandwidth estimate is used instead of the noisy small-buffer one. Good conservative choice.

**[Nit]** Hardcoding the constant (vs. a runtime-tunable custom property, following the existing `accelerator=` pattern in `parseCustomProperties`) is a reasonable first cut given the doc-comment explicitly owns the trade-off ("indicative of one machine," "deliberately conservative"). Worth reconsidering only if a follow-up ever needs to tune this per-target-device; not a request to change it now.

## 2. Gate evaluation point — confirmed static, no hot-path cost

Traced `createTensorBuffers()` (`ext/nnstreamer/tensor_filter/tensor_filter_litert.cc`): it is called exactly once, from `configure()`, before `configured = true`. `output_wrappable` is a member vector populated there and only read (never recomputed) in `invoke()`. The `buf_size == nns_size && nns_size >= litert_wrap_min_bytes && requirementsAllowHostMemory(reqs)` condition is therefore evaluated once per configuration, matching the commit message's claim exactly — nothing added to the per-invoke path.

Also confirmed this commit does not disturb anything the first round approved: `wrapped_buffers` RAII, the `LiteRtRankedTensorType` POD caching, and the input-side exclusion are all untouched by this diff (only the `output_wrappable` computation and its comment changed).

## 3. Does `zeroCopyBelowThreshold` actually pin the gate? — No, and this needed to be checked, not assumed

Read the test: it only asserts the output buffer changed from its `0xAA` canary. It does not observe, directly or indirectly, which branch (`output_wrappable[i] && isHostMemoryAligned(...)`) actually ran — there is no accessor exposing that decision to test code. Both branches produce byte-identical output, so **if `litert_wrap_min_bytes` were deleted (or set to 0), this test — and every other test in the file — would still pass.** The gate's existence is not regression-tested; only its side-effect of "small outputs still work" is.

This is exactly the finding the just-landed `0b88fc7a` ("Say plainly that no test pins the size gate") makes explicit in the docstring, stating it was verified the same way (setting the floor to zero and re-running green) and giving the reason it wasn't turned into a real regression test: doing so would require exposing the wrap/copy choice as observable state outside the subplugin purely for test purposes. That's a legitimate trade-off, and being explicit about it in the comment is the right call — it stops a future reader from believing this constant is guarded when it isn't. No action requested here beyond what 0b88fc7a already did.

`zeroCopyManyOutputsAligned`'s docstring was also correctly updated to say it now exercises the fallback path (every output is 4 B, far below the gate) rather than the wrap path — no stale claim left behind.

## 4. Effects of moving to deeplabv3

- Correct call: every model the suite used before (`mobilenet_v2*`, `simple_32_in_32_out`) sits below 256 KiB and would have silently stopped covering the wrap path if left in place.
- `tests/test_models/models/deeplabv3_257_mv_gpu.tflite` (2,779,264 B) is **not a new asset added by this PR** — it was already committed to the repo (introduced in `1c05475d`, "[Test] Add tensordec-image-segment ssat test") and is resolved the same way as every other model in this file, via `NNSTREAMER_SOURCE_ROOT_PATH`/checkout-relative path. No missing-file risk on either CI job.
- CI timing: `unittest_filter_litert` keeps its existing `timeout: 120` in `tests/meson.build` (unchanged by this commit). Four test cases now compile+run deeplabv3 instead of mobilenet (`zeroCopyAlignedOutput`, `zeroCopyUnalignedOutput`, `zeroCopyPathEquivalence`, `zeroCopyRepeatedInvoke`, the last reduced from 20 to 5 iterations specifically to bound this). On the plain `meson test` path (both ubuntu_clean_meson_build's non-Valgrind gtest run and macos.yaml's `meson test -C build unittest_filter_litert`) this is very unlikely to threaten the 120 s budget. On the Valgrind step, `run_unittests_binaries.sh --valgrind` invokes the binary directly with no `timeout` wrapper and no meson timeout applies, so a slow run won't fail outright, but it will add real wall-clock time to that job under memcheck instrumentation. **[Minor]** Worth a glance at the actual Valgrind job duration once CI finishes (checks are still pending as of this review) to confirm no material regression in total CI time; not blocking.

## 5. Memory under Valgrind

5.5 MB buffers × a handful of concurrently-live allocations (input + output + first-result copies in `zeroCopyRepeatedInvoke`) is on the order of tens of MB total, trivial for memcheck's shadow-memory overhead on a standard CI runner. No realistic memory-pressure concern.

## 6. Coding convention

- `clang-format --dry-run -Werror` run directly against both changed files with the repo's `.clang-format` (via WSL): clean, no diff.
- New `litert_wrap_min_bytes` constant carries a proper `/** @brief ... */` Doxygen block with the measured numbers and rationale inline.
- No multi-line `/* */` (non-Doxygen) comments introduced.
- Both `403ac245` and `0b88fc7a` follow `[Component] Summary` and carry `Signed-off-by`.

## Verdict

No Critical or Major issues. The size floor is well-justified by the numbers actually measured (re-derivation above lands in the same ballpark the author claims), the gate is confirmed static/configure-time-only, and no previously-approved property (RAII, POD safety, input exclusion) was disturbed. The one substantive gap this round was meant to check — whether `zeroCopyBelowThreshold` actually pins the gate — does not, and the author has already landed `0b88fc7a` saying so plainly rather than leaving a misleading docstring. That, plus the CI-timing note above, are the only remaining items, and both are Minor/Nit.

**Approve — OK to merge.**

myungjoo:
Thanks — three responses, one of which corrects something I overstated.

## Minor: CI timing — checked, and it is not a `meson test` risk

`unittest_filter_litert` carries `timeout: 120` ([tests/meson.build:424](https://github.com/nnstreamer/nnstreamer/blob/main/tests/meson.build#L424), raised on main in an unrelated commit), and the suite now runs in **5.6 s** locally, so there is a ~20× margin. The four deeplabv3 cases add about 130 ms between them.

The Valgrind step is the fair part of the concern, and I cannot measure it here — valgrind is not installed on this machine, so anything I said about it would be extrapolation. What I can offer as a reference point is from the previous PR's CI log: `sharedEnvRefBalanceOnConfigureFailure` ran 27 ms plain and 41,160 ms under memcheck, roughly 1500×. On that ratio the deeplabv3 cases would add a few minutes to a job that already runs about an hour.

I have left the iteration counts alone rather than trimming them on a guess. The CI run in flight will give the real number, and if it is worse than that estimate I will cut `zeroCopyRepeatedInvoke` from 5 iterations down — it needs a model above the size gate, so it has to be deeplabv3, but it does not need five rounds.

## Nit: hardcoded constant vs a custom property

Agreed it is a reasonable first cut, and I would rather not add a `custom=` knob here. The threshold is not a user preference — it encodes a hardware property (how a buffer-object construction compares to memory bandwidth), so exposing it invites tuning by guesswork more than it helps. If it turns out to need per-target tuning, that is a better-informed change later.

## On my own wording

Your independent re-derivation of the break-even — ~26–49 kB, from the 5.5 MB/20 GB/s figure rather than the noisier cache-resident 4 kB one — is the more careful number, and it makes 256 kB a **5–10× margin**, not the "order of magnitude" I wrote in the commit message and the code comment. That was me rounding in my own favour. The margin is still ample and the constant does not change, but the comment claims more precision than the measurement supports, so I will correct the wording rather than leave an inflated figure in the source.

## Nit already resolved

Correct on both counts, and thank you for confirming it independently — I found the same thing by setting the floor to zero and watching the suite stay green, which is why `0b88fc7a` states it outright instead of letting the test name imply a guarantee it does not provide.

myungjoo:
CI settled the timing question, and your Minor was right — by more than I allowed for. Trimmed in 46316fc9.

## The actual memcheck cost

All 18 checks passed on a29d4bdb. Pulling the per-case timings out of the Ubuntu job log:

| case | plain | under memcheck |
|---|---|---|
| `zeroCopyAlignedOutput` | 31 ms | 42.2 s |
| `zeroCopyUnalignedOutput` | 29 ms | 42.1 s |
| `zeroCopyPathEquivalence` | 45 ms | 84.1 s |
| **`zeroCopyRepeatedInvoke`** | 98 ms | **209.6 s** |
| `zeroCopyBelowThreshold` | 16 ms | 18.0 s |
| `zeroCopyManyOutputsAligned` | 1 ms | 0.2 s |

About **6.6 minutes** added, against the "few minutes" I estimated from the 1500× ratio on the previous PR. The estimate was too kind, so the commitment I made in the last comment applies.

## What I trimmed, and what I did not

`zeroCopyRepeatedInvoke` was 210 s of that on its own. Five rounds down to three gives roughly 84 s back and still leaves two comparisons against the first result — staleness in a reused wrapper shows on the second round, and a leak needs a sanitizer rather than more rounds.

Nothing else moved. Each remaining case is a single invoke, so cutting any of them would remove a **branch** rather than a repetition: the wrapped path, the misaligned fallback, byte-equality between the two, and the below-gate case are each covering something distinct. And deeplabv3 is not swappable — the model has to clear the 256 kB gate for the wrapped path to execute at all, and nothing else in `tests/test_models` does.

That leaves roughly 5.2 minutes on the memcheck job. I have stopped there rather than trading away coverage for the remainder; if that is still too much for the job's budget, the honest next lever is dropping a branch, which I would rather you weigh in on than decide unilaterally.

Local re-check after the change: 32/32 gtest, `clang-format --dry-run -Werror` clean.

myungjoo:
> **Note:** This comment is a transcription of a **third-round review**, performed by a separate review agent (not a human reviewer), scoped strictly to the last three commits — `0b88fc7a`, `a29d4bdb`, `46316fc9` — per the escalation from the second round. The design approved in rounds 1–2 (output-only scope, RAII wrapper teardown, POD caching of `LiteRtRankedTensorType`, the 256 KiB gate itself) was not re-litigated; each item below only checks whether these three commits disturbed it.

## 1. Does 3 rounds in `zeroCopyRepeatedInvoke` still do the job?

Read the current test body. Two checks run independently of iteration count:
- `changed` (output ≠ the `0xAA` re-poison) is asserted **every** round, including round 0 — this is what would catch a wrapper that silently stops writing on a later invoke.
- `memcmp` against `first_result` runs for `i > 0` — with `iterations = 3` that's rounds 1 and 2, i.e. **two** comparisons against the baseline, matching the commit message's "two instead of four."

A reused-but-stale wrapper (pointing at invalidated LiteRT-side state, or failing to re-target on the second invoke) would most plausibly surface on round 1, exactly where the first comparison lands. Round 2 is a second confirmation rather than covering a distinct failure mode — one could argue 2 rounds (1 comparison) would already have caught it and saved another ~42 s, but 3 is a defensible margin, not an overcorrection. Leak detection genuinely doesn't need more rounds; that's ASan's/Valgrind's job, not the test's. No issue here.

## 2. Is "26–49 kB, five to ten times" consistent with the measured numbers?

Re-derived independently rather than taking the commit at its word:
- Wrap create+destroy is close to size-independent (1.3–2.4 µs measured on the 4 kB case).
- The 5.5 MB deeplabv3 copy (259–286 µs for 5,548,116 B) gives an effective bandwidth of ≈18.8–20.6 GB/s — the realistic memory-bound rate, as opposed to the 4 kB case which is very likely cache-resident and would understate real copy cost if used directly.
- Break-even = wrap-cost ÷ that rate ≈ 25.2–51.4 kB, which lands squarely on the "26–49 kB" now in the comment (the small residual is rounding, not a different method).
- 256/26 ≈ 9.85, 256/49 ≈ 5.22 → "five to ten times" is arithmetically correct.

Comment and commit message check out.

## 3. Does `0b88fc7a` correctly say no test pins the gate?

Grepped `tensor_filter_litert.cc` for any observable signal of which branch ran (debug log, property, counter): none exists in the shipped code — `litert_wrap_min_bytes` is a `constexpr` used only in the internal boolean condition, never surfaced. All six `zeroCopy*` tests assert on output *bytes*, which are identical on both paths by construction. So there is no indirect side channel the author missed (no exposed counter, no `GST_DEBUG` line compiled in — the `ROUTE ...` trace in the PR description was ad-hoc instrumentation, not shipped code). The comment's claim is accurate as written.

## 4. Do these three commits change subplugin runtime behavior?

`git show --stat` for each:
- `0b88fc7a`: `tests/nnstreamer_filter_litert/unittest_filter_litert.cc` only (comment rewrite on `zeroCopyBelowThreshold`).
- `a29d4bdb`: `ext/nnstreamer/tensor_filter/tensor_filter_litert.cc` only, and the diff touches comment lines exclusively — `constexpr size_t litert_wrap_min_bytes = 256 * 1024;` is byte-for-byte unchanged.
- `46316fc9`: `tests/nnstreamer_filter_litert/unittest_filter_litert.cc` only (doc comment + `iterations` constant, 5 → 3).

No production code path changed in any of the three. Confirmed.

## 5. Is ~5.2 min of remaining memcheck cost acceptable, and is there really no smaller model that clears the gate?

Rather than take "nothing else in `tests/test_models` does" on faith, decoded every `.tflite` file in `tests/test_models/models/` (flatbuffer output-tensor shapes):

| model | output tensors | largest single tensor |
|---|---|---|
| `deeplabv3_257_mv_gpu.tflite` | 1 | 5,548,116 B |
| `multi_person_mobilenet_v1_075_float.tflite` | 4 | 100,096 B |
| `mobilenet_v2_1.0_224.tflite` | 1 | 4,004 B |
| `mobilenet_v1_1.0_224_quant.tflite` / `mobilenet_v2_1.0_224_quant.tflite` | 1 | 1,001 B |
| `simple_32_in_32_out.tflite` | 32 | 4 B each |

The gate (`nns_size >= litert_wrap_min_bytes`) is evaluated per output tensor, and `multi_person_mobilenet`'s largest individual output is ~97.75 kB — still well under 256 KiB even though its four outputs sum past it. So deeplabv3 is confirmed the only model in the tree with any single output tensor that clears the gate; the author's claim holds and nothing was missed. Given that, 3 rounds (down from 5) is a reasonable stopping point: the remaining ~5.2 min is the cost of exercising the wrap path with the only asset that can do it, and the honest next lever (as the author already flagged in the PR thread) is dropping a branch/case entirely, which is a scope trade-off for the maintainer, not something this round should force.

## 6. Coding convention

- `clang-format --dry-run -Werror -style=file` against both PR-head files (`ext/nnstreamer/tensor_filter/tensor_filter_litert.cc`, `tests/nnstreamer_filter_litert/unittest_filter_litert.cc`), repo `.clang-format`, via WSL clang-format 14.0.0: clean, no diff.
- Doxygen: `zeroCopyRepeatedInvoke` and `zeroCopyBelowThreshold` both carry proper `/** @brief ... */` blocks after the rewrites; the `litert_wrap_min_bytes` comment stays a well-formed Doxygen block after `a29d4bdb`.
- No non-Doxygen multi-line `/* */` introduced by any of the three commits (the inline `/** Re-poison every round... */` inside the loop predates this range, from `403ac245`, and is itself Doxygen-style).
- Commit subjects: 59 / 72 / 77 chars, all `[Filter/LiteRT] ...`, each with exactly one `Signed-off-by`. Convention followed.

## 7. Issue #4869 completion and PR-body accuracy

- Issue #4869's "Zero-copy invoke" checkbox is still unchecked, but that's expected: the PR explicitly scopes to outputs only, records the `@todo` for inputs (`tensor_filter_litert.cc:34-38`) with the reason (shared `GST_MAP_READ` buffers, no upstream write-safety guarantee) — consistent with what rounds 1–2 already approved. Updating that checkbox is a maintainer bookkeeping step post-merge, not a PR defect.
- Found one real staleness: the PR body's Tests table still reads "`zeroCopyRepeatedInvoke` | deeplabv3 | **5 invokes**, buffer re-poisoned each round..." — `46316fc9` changed the code to 3. Please update the PR description before merge so it doesn't mislead a future reader diffing description against code.
- CI: at the time of this review, the required **`Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)`** job (x86_64 — the one that actually builds/runs `tensor_filter_litert` and is what `46316fc9`'s timing claims are about) was still `in_progress` on `46316fc9`. The arm64 variant already passed (3m26s). Recommend confirming the x86_64 job lands green — this is a process check, not a code concern, since the 18-check-green baseline was captured on the prior head (`a29d4bdb`), and `46316fc9` only removes work from that job.

## Severity summary

- Critical: none
- Major: none
- Minor:
  - PR body's Tests table says "5 invokes" for `zeroCopyRepeatedInvoke"; code now runs 3 (`46316fc9`). Update the description.
  - Confirm the in-progress x86_64 Valgrind CI check on `46316fc9` finishes green before merging (was still running at review time).
- Nit:
  - `zeroCopyRepeatedInvoke` could arguably drop to 2 rounds (1 comparison) for another ~42 s if the memcheck budget is still tight later; not requesting it now — 3 is a reasonable, already-justified stopping point.

None of the three commits touch production code paths, all are comment/test-only as intended, and every specific claim made in their commit messages (break-even math, "no test pins the gate," "no smaller model clears the gate," diff scope) checked out against direct verification rather than being taken at face value.

**Approve — ok to merge**, once the PR description is corrected and the outstanding x86_64 Valgrind check is confirmed green.

myungjoo:
Both Minors closed. All 18 checks pass on 46316fc9.

## Minor 1: stale PR description — fixed

The Tests table said "5 invokes" after `46316fc9` cut it to 3. Corrected.

## Minor 2: the memcheck job — landed green, with the saving measured

It finished green, and the trim did what it was meant to, though not by as much as I claimed:

| | before (`a29d4bdb`) | after (`46316fc9`) |
|---|---|---|
| `zeroCopyRepeatedInvoke` under memcheck | 209,639 ms | **149,402 ms** |

That is **60 s back, not the ~84 s I predicted** — I scaled it as if the cost were purely per-round, but a chunk is the one-off model load and compile, which three rounds pay just as five did. The new-case total on that job is now about 5.6 minutes, down from 6.6.

## Nit: two rounds instead of three

Leaving it at three. Three gives two comparisons against the first result, two gives one, and you agreed three is not an overcorrection. Given the numbers above, a fourth round would only be worth ~30 s anyway, so the remaining lever here is small and I would rather not spend coverage on it unless the job budget actually demands it.

## On the model survey

Thank you for decoding the flatbuffers rather than taking my word for it. I had checked the same thing by opening each model through the subplugin and reading back `getModelInfo`, and we agree: deeplabv3 is the only one clearing the gate, with `multi_person_mobilenet` next at 100,096 B — which you report as ~97.75 kB, the same number.

Worth recording what that near-miss says about the constant: at 100 kB the copy runs roughly 5 µs against a 3 µs wrap, so it would win, but only about 1.7×. The comment claims the gate admits tensors that win *clearly rather than marginally*, and excluding that model is exactly that policy doing its job rather than an accident of where the number landed.

Moving the PR out of draft.

myungjoo:
The substantive question was a real defect. Fixed in 586db0f1, and fixing it made the code smaller.

## `requirementsAllowHostMemory()` was asking the wrong question

You are right, and the distinction matters exactly where this PR claims to be groundwork. The requirements say what is *acceptable*; `LiteRtCreateManagedTensorBufferFromRequirements` decides what is *allocated*. On a CPU configuration those coincide, which is why every measurement and test agreed. With `Accelerators:gpu|npu` they need not: the requirements can list host memory while LiteRT picked a device buffer, and my gate would then wrap it and force the output back to the host on every invoke — adding a transfer the managed path never paid, on hardware no CI job has.

Gating on `LiteRtGetTensorBufferType` of the managed buffer, as you suggested, is strictly better than adding it as an extra condition: LiteRT can only have chosen host memory if the requirements permitted it, so the new check **subsumes** the old one. `requirementsAllowHostMemory()` and its enumeration loop are deleted — the fix is net **-9 lines**.

I did not take on trust that the managed CPU buffer reports host memory, because if it did not the fast path would now be dead everywhere and **every test would still pass** — both branches produce identical bytes. Instrumented the branch instead:

```
ROUTE out[0] size=5548116 WRAPPED        (x5)
ROUTE out[0] size=5548116 copy wrappable=1   (x2 — the misaligned cases)
ROUTE out[0] size=4004    copy wrappable=0
ROUTE out[0] size=4       copy wrappable=0   (x32)
```

Still routing as designed. (Five wraps rather than the seven you may have seen earlier is the repeated-invoke round cut, not the type gate.)

## Nits 1 and 2 — both taken

`~wrapped_buffers ()` now has a `@brief`. And the shared-lock comment now says concurrent invoke means *across* instances rather than within one, since `invoke_outputs` is instance state written before the run — your reading of the ambiguity was fair.

## Nit 3

Agreed, and stated exactly that way in the test: the constant is protected by review only.

## Your follow-up finding — reproduced, and it is worse than a nit

I verified the doxygen checker independently and it holds up:

- `doxygen-tag.sh:75` has `local function_check_flag="f+p"` with **no enclosing function** — confirmed no function definition precedes it in the file.
- `bash -c 'local x=1'` → `local: can only be used in a function`.
- The script has **no `set -e`** (0 occurrences), so it continues with the variable empty.
- `ctags -x --c-kinds=` then lists nothing, so `function_positions` is always empty and the per-function branch cannot fire.

So `CLAUDE.md`'s "every function needs a Doxygen comment" has been enforced by reviewers alone, not CI, for however long that line has been there — which is also why nit 1 could reach this PR at all. Agreed it wants its own issue rather than a drive-by fix here, since turning it back on will likely surface a backlog across the tree. I will raise it separately rather than widen this PR.

Local re-check: 32/32 gtest, 9/9 SSAT, `clang-format --dry-run -Werror` clean. CI running on 586db0f1.

myungjoo:
> **Note:** This comment is a transcription of a review performed by a separate review agent (not a human reviewer), scoped strictly to commit **586db0f1** ("[Filter/LiteRT] Wrap only where LiteRT itself chose host memory"), the response to the design question and 3 nits raised in the previous round on `46316fc9`. The design already approved in earlier rounds (output-only scope, RAII wrapper teardown, POD caching of `LiteRtRankedTensorType`, the 256 KiB gate, deeplabv3 test switch) was not re-litigated — only checked for whether this commit disturbed it. Findings were independently checked against the code and the LiteRT SDK headers (`~/litert-sdk/include/litert/c/` under WSL) rather than taken from the commit message at face value.

## 1. Is the subsumption claim ("LiteRT can only have picked host memory if the requirements permitted it") actually backed by the SDK contract?

Read `litert_tensor_buffer.h` and `litert_tensor_buffer_requirements.h` directly. `LiteRtCreateManagedTensorBufferFromRequirements`'s only doc comment is "This function will use the alignment specified in the requirements" — there is **no explicit line stating the allocated buffer's type is drawn from `supported_tensor_buffer_types`**. The library ships as a prebuilt `libLiteRt.so` with no source available to confirm the implementation either.

That said, the inference is sound as a matter of API design intent, not just hope: `LiteRtTensorBufferRequirements` is *constructed* from a `supported_tensor_buffer_types` list (`litert_tensor_buffer_requirements.h:29-38`) — that list is the entire reason requirements exist as an object distinct from a bare size/alignment struct. An implementation of "create a buffer from requirements" that could return a type outside that list would defeat every other requirements-consuming call in the SDK, not just this one. So subsumption is the only self-consistent reading of the API, even though it isn't a `// guarantee:` comment in the header.

**[Nit]** Given this is inferred rather than documented, it would be slightly more defensive to keep a one-line note in the code (or here) that this assumption rests on `LiteRtCreateManagedTensorBufferFromRequirements`'s undocumented-but-implied behavior, so a future LiteRT SDK bump that violates it is easier to trace back to this comment. Not blocking — the new gate is strictly the correctness-relevant one regardless (it checks what was actually allocated, not what was merely allowed), so even if the inference were wrong the new code degrades to "always copies," never to a wrong wrap.

## 2. `LiteRtGetTensorBufferType` failure / uninitialized `buf_type` — confirmed safe

```cpp
LiteRtTensorBufferType buf_type;
...
const bool wrappable = (buf_size == (size_t) nns_size)
                       && ((size_t) nns_size >= litert_wrap_min_bytes)
                       && (LiteRtGetTensorBufferType (buf, &buf_type) == kLiteRtStatusOk)
                       && (buf_type == kLiteRtTensorBufferTypeHostMemory);
```

`buf_type` is declared uninitialized, but C++ guarantees left-to-right evaluation with short-circuiting for `&&`, so `buf_type` is only read in the fourth operand, which is only reached once the third operand has already confirmed `LiteRtGetTensorBufferType` returned `kLiteRtStatusOk`. No UB, and a failure return correctly falls through to `wrappable = false` (copy fallback), which is the safe direction.

## 3. Removal side effects — no unused variable/include fallout

- `reqs` (`LiteRtTensorBufferRequirements`) is still consumed by `LiteRtCreateManagedTensorBufferFromRequirements` two lines above (`tensor_filter_litert.cc:713`); it did not become dead after the helper's removal.
- `#include <litert/c/litert_tensor_buffer_requirements.h>` is still required for the `LiteRtTensorBufferRequirements` type itself, independent of the deleted helper.
- The `static bool requirementsAllowHostMemory(...)` declaration was removed from the class body along with its definition — no orphaned prototype left behind.
- Net diff matches the commit message exactly: 19 insertions / 28 deletions = **-9 lines**.

## 4. Accelerator-path behavior (code review only, no accelerator CI)

The new gate reads the type LiteRT actually allocated for the managed buffer (`buf`), not the set of types the requirements would have accepted. That is exactly the fix the design question called for: a device-owned managed buffer whose requirements happened to *also* list host memory will now report a non-host `LiteRtTensorBufferType` and correctly fall back to copy, instead of being wrapped and forcing a device→host transfer into `LiteRtRunCompiledModel` on every invoke.

This can't be exercised on the CPU-only CI runners, so it remains unverified by CI, same as before this commit — the author's own instrumentation trace (`ROUTE ... WRAPPED` for deeplabv3, `copy` for the sub-threshold outputs) is the only evidence for the CPU path, but that's consistent with what the code should do: on a CPU-only compiled model, `LiteRtCreateManagedTensorBufferFromRequirements` has no device-owned type to pick from other than host memory, so `buf_type` reporting `kLiteRtTensorBufferTypeHostMemory` there is expected, not just observed.

## 5. Nits from the previous round — both applied and accurate

- `~wrapped_buffers ()` now carries `/** @brief Release every wrapper taken during this invoke. */` — present and correctly placed immediately before the destructor.
- The `shared_lock` comment now reads "Across instances, not within one: `invoke_outputs` below is instance state written before the run...". Verified against `invoke()`: `invoke_outputs` is a per-instance member (`tensor_filter_litert.cc:287`), written at `invoke_outputs[i] = ...` for every output index, unconditionally, before `LiteRtRunCompiledModel` is called. The comment's claim matches the actual control flow.

## 6. New finding this round: a stale doc-comment the commit itself created

```cpp
std::vector<char> output_wrappable{}; /**< requirements allow host memory and the sizes match exactly */
```

This member declaration's inline doc still says "requirements allow host memory," which is precisely the question this commit stopped asking. It should read something closer to `/**< LiteRT chose host memory and the sizes match exactly */` to match the new condition in `createTensorBuffers()`. **[Minor]** — one-line fix, but worth doing before merge so the declaration doesn't mislead the next reader who doesn't cross-check the multi-line comment inside `createTensorBuffers()`.

## 7. Regression check against previously-approved design

`git show 586db0f1 --stat` touches only `ext/nnstreamer/tensor_filter/tensor_filter_litert.cc`. Confirmed by reading the surrounding code that nothing else moved: `wrapped_buffers`'s RAII behavior (only a doc comment added), the `LiteRtRankedTensorType` POD caching, the `buf_size == nns_size` exact-match requirement, the `litert_wrap_min_bytes` constant, and the input-side exclusion are all byte-for-byte unchanged. This commit's diff is confined to the `wrappable` condition in `createTensorBuffers()` plus the two documented nits.

## 8. Coding convention

- `clang-format --dry-run -Werror -style=file` against the PR-head `tensor_filter_litert.cc` and `unittest_filter_litert.cc` (WSL, clang-format 14.0.0, repo `.clang-format`): clean, no diff.
- New/changed comments keep the file's existing multi-line `/** ... */` Doxygen style consistently; no plain (non-Doxygen) multi-line `/* */` introduced.
- Commit subject `[Filter/LiteRT] Wrap only where LiteRT itself chose host memory` follows `[Component] Summary`; exactly one `Signed-off-by` line present.
- (Repeating from earlier rounds for the record, not a finding of this round: the CI doxygen per-function check is disabled repo-wide due to a `local` outside a function in `doxygen-tag.sh`, already tracked as issue #4908 and correctly out of scope here.)

## 9. Test consequences of this commit

No test file changed in `586db0f1`. That's appropriate — the six `zeroCopy*` tests assert byte-correctness of both the wrap and copy paths, which this commit doesn't change; it only changes *which* tensors qualify for the wrap path in an accelerator configuration that no CI job has. The pre-existing, already-documented limitation that no test pins *which* branch actually ran (`0b88fc7a`) still applies and is unchanged by this commit — not a new gap, just an old one that remains open.

## Severity summary

- **Critical:** none
- **Major:** none
- **Minor:**
  - `output_wrappable`'s declaration-site doc comment (`tensor_filter_litert.cc`, member declaration) still says "requirements allow host memory" after this commit changed the actual condition to "LiteRT chose host memory." One-line fix requested before merge.
- **Nit:**
  - The requirements→allocation subsumption claim is a reasonable, API-design-consistent inference but isn't a documented SDK guarantee (opaque prebuilt library, no accessible source); consider a short note flagging it as an assumption. Not blocking, since a violation would only degrade the optimization to "always copy," never produce an incorrect wrap.

## Verdict

No Critical or Major issues. The design question from the previous round is resolved correctly and, per the commit message, with a net reduction in code (-9 lines) rather than an added condition — verified this is because the new check is strictly the more specific one (subsumption holds by the SDK's evident design, if not by an explicit written guarantee). The `LiteRtGetTensorBufferType` short-circuit is safe, no unused variables/includes were left behind, both prior nits are applied and accurate, and no previously-approved property (RAII, POD caching, size gate, input exclusion) was disturbed. One stale doc comment should be fixed before merge; it does not affect runtime behavior.

**Approve — OK to merge**, once the `output_wrappable` doc-comment is corrected (trivial, non-blocking otherwise).

myungjoo:
Both Minors taken, in 9f3fef7b — and the second one changed my mind about something I had declined earlier.

## Minor 1: stale `output_wrappable` comment — fixed

Introduced by the very commit that made it wrong, as you say. It now reads "LiteRT chose host memory, the sizes match exactly and it is worth wrapping", which is what the code asks.

## Minor 2: nothing pinned which path is taken — now it does

I had turned this down two rounds ago, when the gate was two size comparisons and my argument was that a well-commented constant is not worth exposing internal state for. Your framing is what changed it: the gate now turns on **an answer LiteRT gives at runtime**. A future SDK reporting a different buffer type for the CPU path would disable the optimisation on every output, and all six correctness cases would stay green because they only ever compared the two paths to each other.

The evidence that this is a real failure mode rather than a theoretical one is that it has already happened twice in this PR's own history. When the size gate landed I had to instrument the branch by hand to confirm the fast path still ran; when the type gate landed I had to do it again. Both times the suite was 100% green either way. Relying on me to remember to do that by hand is not a control.

So the decision is logged once per output at configure time, and `zeroCopyDirectPathElected` asserts on it. The log is not test scaffolding — it answers "why am I not getting zero-copy on this model" for a user, which is otherwise unanswerable from outside the subplugin, and it costs one debug line per output at pipeline start.

Verified it has teeth by forcing the gate false, simulating exactly the SDK change you described:

| | result |
|---|---|
| `zeroCopyDirectPathElected` | **FAILED** — "the zero-copy path is no longer reachable" |
| the other six `zeroCopy*` cases | passed |

Which is the shape I wanted: the six cover correctness and are rightly indifferent, the new one covers reachability.

## What is still not pinned

`litert_wrap_min_bytes` itself. The new case asserts that *some* qualifying tensor elects the direct path, not that the threshold sits at any particular value — setting it to 1 byte would still pass. That is deliberate: the constant is a performance heuristic with no observable behaviour, and the test comment continues to say it is protected by review only. What is now guarded is the part that can break underneath us without anyone touching this file.

Local: 33/33 gtest, 9/9 SSAT, `clang-format --dry-run -Werror` clean. CI running on 9f3fef7b.

## Your doxygen finding

Filed as #4908, with the mechanism, the split between which checks are dead and which still work, and a measurement: fixing the one line surfaces **1,144 errors across 117 files** in `gst/` and `ext/` alone, so it needs a backlog plan rather than a drive-by. Thank you for spotting it — it is also why the missing `@brief` you caught in the last round could reach the PR at all.

myungjoo:
Both Minors fixed in ce052f50. The first one was a plain factual error on my part.

## Minor 1: the `G_MESSAGES_DEBUG` dance — removed, and my comment was wrong

I had written that the decision "is logged at debug level, so it is dropped unless the domain is enabled for this process", and set `G_MESSAGES_DEBUG` around the call on that basis. That is not how GLib works: level filtering happens **inside** `g_log_default_handler`, so a handler installed in front of it is called regardless.

I verified rather than take it on your word, with a minimal glib 2.72 program:

```
--- without G_MESSAGES_DEBUG:   env=unset count=1
--- with G_MESSAGES_DEBUG=all:  env=all   count=1
```

Identical. So the `setenv` was accomplishing nothing except discarding whatever the caller had exported — the side effect you flagged, with none of the benefit I thought I was buying. Both `g_setenv` and `g_unsetenv` are gone, and the comment now states the actual behaviour and that it was checked.

This also brings the case into line with `_countUnbalancedRelease`, which never needed the variable either — I should have noticed the inconsistency with the existing convention.

Re-verified end to end after the change, since I altered the mechanism the test depends on:

| | result |
|---|---|
| clean, no `G_MESSAGES_DEBUG` set | `[ OK ] zeroCopyDirectPathElected`, 33/33 |
| gate forced false | `[ FAILED ] zeroCopyDirectPathElected`, other six pass |

So the counting was working through the handler all along, not through the variable.

## Minor 2: the literal string coupling — noted at the source

Agreed the failure mode is nasty: rewording the message would fail the test with "the zero-copy path is no longer reachable", pointing the next reader at the gate rather than at the string they just edited. There is a note next to the `ml_logd` call now saying the test matches on `"will be written directly"` and that the two change together.

I left the coupling itself in place. The alternative — exporting a counter or a query API — would put more of this in the public surface than a diagnostic log line deserves, and the failure direction is at least loud rather than silent: a stale string makes the test fail, never pass.

Local: 33/33 gtest, 9/9 SSAT, `clang-format --dry-run -Werror` clean. CI running on ce052f50.

myungjoo:
> **Note:** This comment is a transcription of a review performed by a separate review agent (not a human reviewer), scoped strictly to the last two commits — `9f3fef7b` ("Make the direct-path decision observable, and fix a stale comment") and `ce052f50` ("Drop the pointless `G_MESSAGES_DEBUG` dance in the elected-path test") — against PR head `ce052f50`. 아래 내용은 타 Agent가 작성한 리뷰를 옮겨 적은 것입니다.

## Verdict

**Approve — ok to merge.** Both minors from the previous round (`9f3fef7b`) are properly closed in `ce052f50`, no runtime behavior changed beyond a configure-time log line, and nothing in the earlier approved design (input exclusion, RAII, POD caching, 256 KiB gate, buffer-type gate, deeplabv3 tests) was touched or broken by these two commits.

## What changed, confirmed by diff

`git show --stat` for both commits:

```
9f3fef7b: tensor_filter_litert.cc | 10 +++-        unittest_filter_litert.cc | 62 ++++++++
ce052f50: tensor_filter_litert.cc |  6 +++-        unittest_filter_litert.cc |  8 ++++----
```

Production-side change across both commits is exactly one `ml_logd(...)` call added in `createTensorBuffers()`, once per output tensor, at **configure time** — not in `invoke()`. No branch, gate, or buffer-selection logic changed. The invoke hot path is untouched.

## Item 1 — `zeroCopyDirectPathElected` robustness (re-checked at `ce052f50`)

- **No fatal early-return between handler install and restore.** Both `ASSERT_TRUE` calls (`_GetModelFilePath`, `sp != nullptr`) happen *before* `g_log_set_default_handler` is called. Between install and restore the only statement is `sp->open(&prop, &data)`; `ASSERT_EQ(ret, 0)` and the `EXPECT_GT` run only after the handler has already been restored. So there is no path that leaves `_countDirectOutputs` installed globally for later tests.
- **`sp->open()` cannot throw past the test.** It resolves to `tensor_filter_subplugin::cpp_open` (`gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc:76`), which wraps `configure_instance()` in `catch (const std::bad_alloc&) / catch (const std::invalid_argument&) / ... / catch (...)` and converts every exception to a negative return code. `configure_instance()` itself (`tensor_filter_litert.cc:765`) also catches around `createTensorBuffers()` and rethrows only after `cleanup()`. So even a LiteRT SDK failure mid-`createTensorBuffers()` cannot unwind through `sp->open()` and skip the handler restore.
- **`_direct_output_count` cannot be cross-contaminated by the threaded tests in this file** (`sharedEnvConcurrentOpen`, `sharedEnvInvokeDuringConfigure`, etc.). The custom default handler is only installed for the narrow window around this one `sp->open()` call inside this one test; gtest runs test cases sequentially in this binary, and the threaded tests' own `sp->open()` calls run in test cases that never have `_countDirectOutputs` installed, so their `ml_logd` output is never counted and never touches this counter.
- **The glib claim in `ce052f50`'s commit message and comment is correct, and not glib-version-specific.** Reproduced independently on WSL glib 2.72.4 with a minimal standalone program (same shape: install handler, call `g_debug`, check count) — the custom default handler counts the message identically whether `G_MESSAGES_DEBUG` is unset or set to `all`. Cross-checked against current upstream GLib documentation for `g_log_default_handler`/`g_log_set_default_handler`: level/domain filtering against `G_MESSAGES_DEBUG` is documented as happening *inside* `g_log_default_handler` itself, not earlier in the dispatch path — so any handler installed via `g_log_set_default_handler` unconditionally receives the message before that filtering would apply. This is a stable, documented API contract, not an accident of one glib release. It's also worth noting the prior-round review on `9f3fef7b` already reported testing this on both glib 2.72.4 (jammy) and 2.80.0 (noble) with the same result — i.e. this has now been checked independently on three separate occasions across two glib minor versions with no divergence. Risk of this breaking on macOS CI's (newer) glib is low.
- Net effect of `ce052f50` on this test: it is now *more* isolated than before, not less — the old code called `g_setenv("G_MESSAGES_DEBUG", "all", TRUE)` and unconditionally `g_unsetenv`'d it afterward, which could clobber a value a caller had exported for the whole run. The new version touches the environment not at all.

## Item 2 — production log appropriateness

`ml_logd` (→ `g_debug` on the Linux/macOS path, `nnstreamer_log.h:79`) is the correct level: dropped by default (`G_MESSAGES_DEBUG` unset), costs one format+dispatch per output only at configure time, not per invoke. For a 32-output model (`manyInOutModel`/`zeroCopyManyOutputsAligned` use `simple_32_in_32_out`) that's 32 debug-level lines once at configure — negligible, and invisible unless a developer has opted into debug logging. No concern.

## Item 3 — test/production string coupling

The literal `"will be written directly"` match remains a coupling point between `tests/nnstreamer_filter_litert/unittest_filter_litert.cc:511` and `ext/nnstreamer/tensor_filter/tensor_filter_litert.cc:754-755`, but `ce052f50` now documents the coupling on **both** sides (the test comment already existed; the production comment is new in this commit, at `tensor_filter_litert.cc:750-753`), each naming the exact string and the other file. Given the log line only needs to exist and mention the accepted state in *some* stable form, and a shared constant would be over-engineering for a two-file, actively-cross-referenced coupling, this is a reasonable resolution of the concern the previous review raised (Minor 2 on `9f3fef7b`) — not a design that needs a stronger mechanism (e.g., a shared `constexpr` string) to be acceptable.

## Coding convention

- `clang-format --dry-run -Werror` on both changed files at `ce052f50`: **clean**, no diff (verified directly, not taken on faith).
- `_countDirectOutputs` (new in `9f3fef7b`) carries a full Doxygen block with `@param` tags; no new function in `ce052f50` (comment-only + one-line env-touch removal).
- The new `/** ... */` multi-line blocks in both commits match the file's pre-existing Doxygen-style block-comment convention (e.g. the `output_wrappable` field comment and the `wrappable` gate explanation predating this pair of commits use the identical style) — not a departure that needs flagging.
- Both commits: `[Filter/LiteRT] Summary` subject line, body explains what/why, single `Signed-off-by: MyungJoo Ham <myungjoo.ham@samsung.com>` each, not duplicated.

## PR-level final check

- Issue #4869's "zero-copy invoke" item is addressed for outputs, as scoped and stated in the PR description; inputs are explicitly and correctly left out with a stated reason (shared/read-mapped input memory).
- The PR body's Tests table **already lists** `zeroCopyDirectPathElected` (added when `9f3fef7b` landed) — it is not missing, contrary to what might be assumed from an earlier partial read of the description.
- All 18 CI checks pass on `ce052f50`, including both jobs that actually build/test the litert plugin (`ubuntu_clean_meson_build.yml` x86_64, `macos.yaml`).
- CI's silently-disabled per-function Doxygen enforcement is tracked separately as #4908 and is correctly out of scope here.

No blocking items. No nits worth holding this up for.

myungjoo:
Thanks — approve noted, and two pointers.

## The doxygen finding already has its issue

You flagged it again as deserving one: it is **#4908**, filed after the round that first raised it. It carries the mechanism, the split between which checks are dead and which still work (file-level and struct/class checks are unaffected — only the per-function one is dead), and a measurement of why it is not a drive-by: fixing the single line surfaces **1,144 errors across 117 files** in `gst/` and `ext/` alone, so it needs a staged backlog cleanup with the flag flipped last.

One correction to the shared account of the mechanism, for whoever picks that issue up. `ctags -x --c-kinds=` does not universally "list nothing" — that is true of `exuberant-ctags`, which is what CI installs, but `universal-ctags` (what most current distros provide as `ctags`) returns **namespaces and classes**, because those are C++-only kinds and are unaffected by emptying the *C* kind set. There the check misfires rather than going quiet, reporting `namespace ... function needs @brief tag`. Same root cause, different symptom, and it matters for anyone reproducing locally. That is recorded in #4908.

## The glib citation note — leaving it as written, deliberately

You are right that the CI matrix also runs glib 2.80 and that widening the citation would cover both. I am leaving `verified on glib 2.72` alone, because that is the version **I** actually compiled the probe against. Writing "2.72 and 2.80" would put my name on a check you ran, not one I did. The narrower claim is the true one, and as you say the behaviour is an API contract in `g_log_default_handler` rather than a version accident, so the comment does not need to enumerate versions to be correct.

If that reads as under-claiming to a later reader, the fix is for them to widen it after re-running the probe themselves, which is the same standard I held myself to.

Issue #4903, opened 2026-08-28
[CI/Test] Nothing runs SSAT from an installed tree, so every runTest.sh's installed-mode branch is unguarded


myungjoo:
Each `runTest.sh` has two ways to find the sub-plugin it needs: the build tree at `../../build`, or the installed path read from `/etc/nnstreamer.ini` when that directory is absent. CI only ever exercises the first.

No job runs SSAT from an installed tree, and none can without new plumbing:

- RPM runs `%check` before `%install`
- debhelper runs `dh_auto_test` before `dh_auto_install`

Neither packaging flow has a point at which an installed tree exists to test against. So the `/etc/nnstreamer.ini` branch that most suites carry has never been executed by CI, in any suite.

## Why it matters now

That branch was dead weight while the suites were not shipped. Two recent changes made it load-bearing:

- #4885 put eleven previously-unshipped SSAT suites into the `unittest-nnstreamer` package.
- #4900 gave the seven remaining suites an installed-mode fallback, so that from an installed tree they run instead of reporting no cases.

Both had to be verified by hand — driving each script's probe with a stub SSAT API against a fake ini — because there is nothing in CI that could confirm or refute them. A future edit can silently re-inert any of these suites and every check stays green, which is the same failure mode that let the install list in `tests/meson.build` go stale for four years (#4885).

## What would close it

A job that installs with `-Dinstall-test=true` and then runs `ssat` from `<prefix>/lib/nnstreamer/bin/unittest-nnstreamer/tests/<suite>` for at least one suite. Even a single suite would cover the shared shape, since the probe is near-identical across all of them.

Worth deciding separately:

- whether to assert that the suite actually ran (`N passed among N cases`) rather than merely exited zero — a skip and a pass are both green today, and telling them apart is the whole point here
- which suite to use; one whose sub-plugin is always built in that job, so a skip is unambiguously a regression

Related: #4888 covers the analogous blind spot on the valgrind side, where the pass covers 4 of 26 binaries.

Not urgent. Filing it so the gap is recorded rather than rediscovered.

PR #4906, opened 2026-08-31
[Test/TensorRT10] Point the suite at a model that exists

### LINKED ISSUE #4890: [Test/TensorRT10] runTest.sh points at yolov5nu_224.onnx, which is not in the tree
`tests/nnstreamer_filter_tensorrt10/runTest.sh:69` uses a model fixture that does not exist:

```sh
PATH_TO_MODEL="../test_models/models/yolov5nu_224.onnx"
```

`git ls-files tests/test_models/models` lists only `mobilenet_v2_float.onnx` and `mobilenet_v2_quant.onnx` for ONNX, and grepping the tree finds nothing that downloads or generates `yolov5nu_224.onnx`.

The script has no fixture-existence guard either — it goes straight from the sub-plugin probe to `gstTest`. So on a TensorRT10-capable machine the positive cases fail rather than skip; only the two `_n` cases behave as intended.

This has gone unnoticed because TensorRT10 is not enabled in any CI job, so the suite never runs.

Found while auditing the SSAT install list in #4885. That PR starts shipping this suite in the `unittest-nnstreamer` package, which is what makes the missing fixture reachable on a device — but the defect is pre-existing and independent of it.

Worth deciding between:
- adding the fixture (with its provenance recorded, given it is a binary model), or
- pointing the suite at a model already in the tree, or
- guarding on the file and skipping cleanly when it is absent, the way `nnstreamer_filter_tvm/runTest.sh` does for its per-architecture models.

The third is the minimum needed to stop the suite failing for the wrong reason.

myungjoo:
Fixes #4890.

`tests/nnstreamer_filter_tensorrt10/runTest.sh` pointed at `../test_models/models/yolov5nu_224.onnx`, which has never been in the tree and which nothing downloads or generates. With no fixture-existence guard, the positive cases fail outright on a TensorRT10-capable machine instead of skipping.

### Which of the three options in the issue this takes

The issue left the choice open between adding the fixture, repointing at an in-tree model, and guarding-and-skipping. This does the first and the third, because the second turned out not to work and the third alone loses real coverage:

| Option | Why not |
| --- | --- |
| `yolov5nu_224.onnx` | Ultralytics YOLOv5nu is AGPL-3.0. It does not belong in an LGPL-2.1 tree that ships this directory inside `unittest-nnstreamer`. |
| `mobilenet_v2_quant.onnx` | QOperator format (`QLinearConv`), which the TensorRT ONNX parser does not accept. |
| `mobilenet_v2_float.onnx` | Stores weights as **uint8** behind `DequantizeLinear`; the TensorRT ONNX parser takes int8/fp8/int4 quantized tensors only. Would have reproduced the exact same "positive cases fail" symptom. |
| Guard only | Stops the wrong-reason failure but permanently disables the positive cases, including the #4850 CUDA-managed-memory regression test. |

### What is added

`tests/test_models/models/add_one_224.onnx` — **299 bytes** — with `gen_add_one_224.py` committed beside it, following the `gen_mobilenet_v2_float.py` precedent so the provenance of a binary fixture is in-tree and reproducible. It is a single 1x1 convolution with the 3x3 identity kernel and a bias of 1.0, so it computes exactly `out = in + 1.0` on a 1x3x224x224 float32 tensor. Only `Conv` is used, and the batch dimension is fixed, which the sub-plugin requires.

Because multiplying by 1.0 and adding 1.0 is exact in IEEE-754 for the integral pixel values (0..255) the pipeline feeds in, and stays exact under the TF32 accumulation TensorRT enables by default (an 11-bit significand represents every integer below 2048), the sub-plugin output can be compared **byte for byte** against a `tensor_transform` branch instead of merely drained. This is the same design as `nnstreamer_filter_tvm/runTest.sh`.

### Test changes

- **Case 1** — single buffer, byte-exact `tee` comparison, plus an output-size assertion pinning shape and dtype (602112 = 3 &times; 224 &times; 224 &times; 4).
- **Case 2** — eight buffers across a queue boundary, the #4850 topology, now compared rather than only drained. This replaces the previous `fakesink dump=true` case: `filesink` maps and reads every byte in the sink thread exactly as the hex dump did, so the same use-after-free hazard is exercised, while additionally catching corruption that does not crash — and without writing 4.8 MB of hex into the log.
- **Guard** — skips cleanly when the fixture is missing, the way `nnstreamer_filter_tvm/runTest.sh` does. This is the regression test for this issue.
- **Negative cases** — added a missing model file, a mismatched input dimension, and a mismatched input type. `2_n` (invalid custom property) now runs against a model that actually loads, so it reaches the path it was written for instead of failing on the missing file.

### Verification

Run locally (WSL, gst-launch + SSAT), with `tensor_transform add:1.0` standing in for the TensorRT filter:

- Guard fires and the suite skips cleanly with the model absent; does not fire with it present.
- Both positive pipelines run end to end, compare byte-identically, and match the 602112-byte assertion; the 8-buffer case yields 4816896 bytes on both branches.
- Every negative pipeline negotiates up to the filter, so any failure comes from `tensor_filter` — which is what the `_n` cases assert.
- The fixture regenerates byte for byte; git stores it as binary, unmangled.
- `pylint` 10.00/10 on the generator, and `ssat-install-list.sh` reports the install list in sync.

**Not verified:** the TensorRT engine build and the inference itself. No CI job enables TensorRT10 (`tensorrt_support 0` in the spec, absent from every workflow) and no machine available here has it. The reasoning above for why this model parses where the in-tree ones do not is from the TensorRT ONNX parser's documented type support, not from a run.

### Scope

Test-only. No sub-plugin, API, or architecture change, so no documentation update applies. `tests/meson.build` changes by one token, adding the new generator to the existing `exclude_files` list so the host-only script does not reach target devices.

### Second commit (review follow-up)

Removes `tests/test_models/labels/coco.txt`. `git log -S "labels/coco.txt" --all` shows it entered the tree with the TensorRT10 support commit and was read by nothing but this suite, so the fixture swap left it dead — and `install_subdir` was still shipping it inside `unittest-nnstreamer`. The bounding-box decoder tests are unaffected; they use `coco_labels_list.txt` and `coco-80.txt` under `tests/nnstreamer_decoder_boundingbox/`. `LICENSE.WordNet3` covers `labels.txt` (ImageNet), not the COCO list, so it stays. Also hardens the cleanup to `rm -f *.log`. The `*.log` glob follows `nnstreamer_filter_onnxruntime`, `nnstreamer_converter` and `nnstreamer_decoder`; the `-f` follows `nnstreamer_filter_tvm` and `codegen`, which pass it with explicit filenames.

### Third commit (review follow-up)

Removes `/tmp/add_one_224.engine` before the cases run. `loadModel()` caches the built engine there keyed only on the model's stem and reuses it unconditionally, so an engine left from a differently built `add_one_224.onnx` would make the new byte-exact comparisons verify the wrong model silently. Cleaning up only at the end would leave the run that found the stale engine still using it, so the removal happens up front — after the fixture guard, so a skipping suite does not touch `/tmp`. This also restores coverage the cache had removed: the `.onnx` → `.engine` build path previously ran only on a machine's first ever run of this suite.


myungjoo:
**Disclosure:** This review was produced by an AI code-review agent (not a human maintainer), invoked by another AI agent orchestrating this task. Treat the findings below as input for human/maintainer judgment, not as an authoritative review.

## Summary

The PR does what it says: `tests/nnstreamer_filter_tensorrt10/runTest.sh` now points at an in-tree, regeneratable, LGPL-compatible fixture (`add_one_224.onnx`), gained a fixture-existence guard, and its positive cases now assert byte-exact correctness (via `callCompareTest`, mirroring the `nnstreamer_filter_tvm` precedent) instead of merely draining the pipeline. The `_n` cases were also fixed to reach the failure path they were written for (e.g. the old `videotestsrc num_buffers=1` property-name typo, and `2_n`/now `2_n`'s model argument previously missing entirely rather than merely invalid).

I independently verified, outside the diff itself:
- `gen_add_one_224.py` regenerates `add_one_224.onnx` **byte-for-byte identical** to the committed binary (checked with `cmp`).
- The generated model, run through `onnxruntime` (CPU EP) on random 0–255 inputs, computes `out = in + 1.0` **exactly** (max abs diff 0.0) — consistent with the "byte-exact comparable" design rationale in the PR body.
- `pylint` on `gen_add_one_224.py` scores 10.00/10, as claimed.
- `bash -n` on the new `runTest.sh` passes; the `callCompareTest`/`testResult` call signatures match the already-merged `nnstreamer_filter_tvm/runTest.sh` exactly.
- `tests/meson.build`'s one-line change follows the existing `exclude_files` pattern used for `gen_mobilenet_v2_float.py`.
- Confirmed via `packaging/nnstreamer.spec` (`tensorrt_support 0`) and `.github/workflows/` (no `tensorrt10`/`tensorrt_support` reference anywhere) that TensorRT10 is indeed not built in any CI job today, matching the PR's own disclosure.
- Scope is test-only: only `tests/meson.build`, `tests/nnstreamer_filter_tensorrt10/runTest.sh`, and two new files under `tests/test_models/models/` are touched. No sub-plugin/API/architecture change, so the "no documentation update needed" claim is correct.

No blocking issues found. Two minor points below, neither of which should hold up merging.

---

### 1. (Minor) `tests/test_models/labels/coco.txt` becomes an orphaned fixture

Before this PR, `coco.txt` was referenced only by this test's now-removed `tensor_decoder mode=bounding_boxes` YOLO pipeline (`PATH_TO_LABEL="../test_models/labels/coco.txt"`, `option2=${PATH_TO_LABEL}`). A repo-wide grep after this PR shows **zero remaining references** to `coco.txt` anywhere in the tree. It's a small file (707 bytes, 80 lines), so impact is negligible, but since it's now genuinely dead weight it would be worth deleting in this PR or a fast follow-up, rather than leaving an unreferenced binary/text fixture behind. (`orange.png`, by contrast, is still used by many other filter test suites — that one is fine to keep.)

### 2. (Minor / already disclosed, flagging for visibility) This test still cannot be exercised by CI

Since no workflow builds `tensorrt_support`/`tensorrt10`, this suite — both before and after this fix — only ever runs on a developer's or device's own TensorRT10-capable machine (e.g. via the `unittest-nnstreamer` package per #4885), never in GitHub Actions. That means a future regression in `tensor_filter_tensorrt10.cc` (or in shared core code it depends on) still cannot be caught automatically at PR time; it would only surface the same way #4890 did — by accident, on a device, much later. This is explicitly and honestly called out in the PR description already ("Not verified: the TensorRT engine build and the inference itself... no CI job enables TensorRT10"), and fixing it is out of scope for a test-fixture PR, so I'm not treating it as blocking. Worth considering a follow-up issue to track adding a TensorRT10-capable CI/on-device job, if one isn't already tracked, so this class of bug (fixture rot going unnoticed) doesn't recur.

### 3. (Nit) `rm *.log` cleanup vs. explicit file list

The new cleanup at the end of the script is `rm *.log` (unguarded glob), whereas the `nnstreamer_filter_tvm` precedent uses `rm -f testcase1.tvm.log testcase1.direct.log ...` (explicit names, `-f`). In practice this is harmless here — the guard exits before any log file could fail to exist, and by the time cleanup runs, all four `testcase{1,2}.{tensorrt10,direct}.log` files exist — but `rm -f *.log` (adding `-f`) would be marginally more robust and consistent with the sibling suite's style.

---

**Verdict:** Correctly and minimally solves #4890 within its stated scope; verified reproducibility and pylint/model-semantics claims independently; no regression risk to other modules since no production code is touched; test-case policy (regression test for a bug fix) is satisfied. The two minor points above (dead `coco.txt`, `rm -f` nit) are optional cleanup, not blockers.

myungjoo:
Thanks for the review. Addressed in 63dfaf93.

### Minor: orphaned `tests/test_models/labels/coco.txt` — fixed

Verified independently before removing it. `git log -S "labels/coco.txt" --all` returns only the two TensorRT10 support commits and this branch, so that path entered the tree with this suite and was read by nothing else; the last reference went away with the fixture swap. The bounding-box decoder tests are untouched — they use `coco_labels_list.txt` and `coco-80.txt`, both of which live under `tests/nnstreamer_decoder_boundingbox/`, not in `test_models/labels/`. Deleting `coco.txt` therefore also stops `install_subdir` shipping a dead 707-byte file inside `unittest-nnstreamer`.

`LICENSE.WordNet3` in the same directory covers `labels.txt` (ImageNet), not the COCO list, so it stays.

### Nit: `rm *.log` → `rm -f *.log` — fixed

Matches `nnstreamer_filter_tvm`. A run where no case produced a log no longer ends on a spurious `rm` error.

### Minor: no CI job builds TensorRT10 — acknowledged, deliberately not fixed here

Agreed on the substance, and it is the reason this defect survived. I did not add a static checker for fixture paths in this PR, because the obvious form of it is wrong on day one: this suite now deliberately references `no_such_model.onnx`, and `nnstreamer_filter_onnxruntime/runTest.sh` deliberately references `../${PATH_TO_MODEL}`, both as negative cases. A checker would need an exemption convention from the outset, which is a design discussion of its own rather than something to settle inside a fixture fix. Better as a separate issue.

One trade-off worth stating explicitly, since it is adjacent: the guard **skips** when the fixture is missing, as #4890 proposed and as `nnstreamer_filter_tvm` does. Note the two cases are not quite alike — TVM's models are per-architecture, so absence is a legitimate environment condition, whereas `add_one_224.onnx` is architecture-independent and always committed, so its absence can only mean a broken tree or package. There is an argument that this guard should fail loudly instead of skipping. I kept the skip because that is what the issue asked for, but it is a one-line change if the maintainer prefers the stricter reading.

---
*This is a relay of a review produced by a separate AI agent, re-verified before acting on it.*

myungjoo:
**Disclosure:** This is a follow-up review from an AI code-review agent (a separate instance from the one that left the first review), relaying re-verification to a human/maintainer. Treat it as input, not an authoritative sign-off.

## Re-review of 63dfaf93 (main...claude/github-issue-4890-e63abd, 2 commits total)

I re-verified the follow-up commit and the response comment independently rather than taking either at face value.

### 1. `coco.txt` removal — confirmed correct, no regression found

- `git log -S "labels/coco.txt" --all` on my end returns the same three commits the author cites: the two "Add support for TensorRT 10" commits that introduced the file (`66234fdc`, `65137539`) and this branch's fixture-swap commit (`64a69096`) that removed its only reference. No other commit in the repo's history ever touched that path.
- Repo-wide grep for `coco\.txt` after the deletion returns zero remaining references anywhere (code, tests, docs, tooling).
- `tests/nnstreamer_decoder_boundingbox/` carries its own `coco_labels_list.txt`/`coco-80.txt`, confirmed independent of `tests/test_models/labels/`, so the bounding-box decoder suite is unaffected — matches the author's claim.
- Packaging/install path checked directly: `tests/meson.build` installs `test_models` via `install_subdir(...)` (whole-directory install, only the two `gen_*.py` scripts individually excluded), and `packaging/nnstreamer.spec` ships `.../unittest-nnstreamer/tests` as a directory glob (`%files unittests`, line ~1447) with no per-file listing of `coco.txt` anywhere in the spec. So deleting the file needs no companion change in either, and there is none — consistent.
- `LICENSE.WordNet3` in the same directory is Princeton's WordNet license text; it covers `labels.txt` (ImageNet synsets), not COCO, so the author's claim that it "stays" and is unrelated is correct.
- Checked `.github/workflows/_typos.toml` and `tools/development/getTestModels.sh`, the only other repo files that mention `tests/test_models/labels/`: both reference the directory generically or name `labels.txt` specifically, never `coco.txt`. No update needed there either.

Verdict on point 1: the fix is correct and the stated justification holds up under independent verification. No regression.

### 2. `rm *.log` → `rm -f *.log` — confirmed fixed, matches precedent

Diffing the two commits shows this is the only change to `runTest.sh` besides the `coco.txt`-related line removed with it. Now byte-identical in spirit to `nnstreamer_filter_tvm/runTest.sh`'s cleanup. `bash -n` still passes on the updated script. Resolved.

### 3. Static fixture-path checker — not added, with reasoning; no objection

Agreed this is correctly out of scope for this PR. A naive checker would immediately need an exemption list on day one (this suite's own `no_such_model.onnx` negative case, and `nnstreamer_filter_onnxruntime/runTest.sh`'s `../${PATH_TO_MODEL}` negative case), which is a real design question, not a one-line addition. Deferring it to a separate issue is reasonable and does not weaken this PR's own fix.

### New point raised in the reply: skip vs. fail-loudly on a missing `add_one_224.onnx`

The author flagged, unprompted, that the guard silently skips the suite when the model is absent — appropriate for `nnstreamer_filter_tvm`'s per-architecture models (absence is a legitimate environment condition there) but arguably not for `add_one_224.onnx`, which is always committed and architecture-independent, so its absence can only mean a broken tree/package (i.e., exactly the regression class #4890 itself was).

I think this reasoning is sound, and it's worth being explicit about the trade-off for the maintainer: as written, if `add_one_224.onnx` ever went missing again (e.g. a bad rebase, a packaging step dropping it, an errant `.gitattributes`/`exclude_files` edit), this suite would once again silently report "skipped" instead of failing — which is a milder version of the exact failure mode #4890 was filed for, just one layer safer than before (wrong-fixture-name vs. no-guard-at-all is fixed; missing-file-should-probably-be-fatal is not). That said:
- It matches issue #4890's own literal ask ("guarding on the file and skipping cleanly... is the minimum needed").
- It matches the one existing in-tree precedent (`nnstreamer_filter_tvm`).
- Nothing here regresses versus the pre-PR state; it's strictly better (a defined, intentional skip vs. an undefined failure).

So I'd classify this as **Minor / non-blocking**, not something to hold up this PR over — but flagging it explicitly for the maintainer's judgment call, since the author raised it in good faith and it's a legitimate one-line follow-up (`report; exit` → `report; testResult 1 <id> "..." 0 0` or similar) if the stricter reading is preferred.

### Anything missed in the first pass?

I re-read the full updated `runTest.sh` end to end (not just the incremental diff) and re-ran `bash -n` on it; nothing else changed besides the two items above, and I found no new issues beyond the skip/fail-loudly point already raised by the PR author itself.

---

## Verdict

Both actionable items from the first review (orphaned `coco.txt`, `rm -f`) are fixed and independently re-verified with no side effects on packaging, install lists, or documentation. The CI-coverage gap and the static-checker deferral are acknowledged trade-offs, not defects, and are reasonably scoped out of this PR. The skip-vs-fail-loudly question raised in the author's own reply is a legitimate, low-severity (Minor) design point for the maintainer to weigh in on, but does not block merging.

**No remaining blocking issues. This PR is mergeable as-is**, modulo the maintainer's own call on the skip-vs-fail-loudly trade-off noted above.

myungjoo:
Thanks — points 2 and 3 were right and are addressed in a710327e and the PR description. Point 1 I believe is incorrect; details below.

## 1. `3_n` comment — respectfully disagree, the ONNX parser *is* reached

`tensor_filter_tensorrt10.cc:396` does set `info.verify_model_path = TRUE`, but that flag is not in effect on the first `open()`.

For a V1 framework, `verify_model_path()` reads `priv->info.verify_model_path`, not `fw->verify_model_path`:

https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/tensor_filter/tensor_filter_common.c#L487

`priv->info` is zeroed by `gst_tensor_filter_framework_info_init()` at line 989, and the only place it is ever populated is line 2569 — which runs *after* `priv->fw->open()` has already succeeded:

```c
if (verify_model_path (priv)) {           /* priv->info.verify_model_path is still 0 here */
  ret = priv->fw->open (&priv->prop, &priv->privateData);
  priv->prop.fw_opened = (ret >= 0);
}
...
if (priv->prop.fw_opened) {
  if (GST_TF_FW_V1 (priv->fw)) {
    ret = priv->fw->getFrameworkInfo (..., &priv->info);   /* only now */
```

The other three `getFrameworkInfo` call sites (1082, 1305, 3006) all write into a *local* `info`, for accelerator enumeration and framework detection — none of them touch `priv->info`. And `G_FILE_TEST` appears only twice in the file: line 501 inside `verify_model_path()`, and line 1185, which is an `IS_DIR` test for multi-model paths, not an existence gate.

So with `verify_model_path` still 0, the check is skipped, `open()` runs, and `configure_instance()` → `loadModel()` → `buildSaveEngine()` → `parseFromFile()` fails on the missing file — before `engineFile.write()`, so no empty engine is produced either.

**Verified empirically.** `tensor_filter_python3.cc:758` sets `info.verify_model_path = TRUE` exactly as tensorrt10 does, and is likewise a V1 C++ sub-plugin. Given a nonexistent model:

```
tensor_filter framework=python3 model=/tmp/no_such_model.py
```

the error is the sub-plugin's own loader, not the framework's path check:

```
ModuleNotFoundError: No module named 'no_such_model'
CRITICAL: the script is not properly loaded
CRITICAL: Failed to initialize the object: Python
```

If `verify_model_path()` had gated it, this would read `Cannot find the model file[0] ...` and the sub-plugin would never have been opened.

Leaving the comment as it stands. Happy to be shown otherwise if I have misread the ordering.

## 2. Engine cache — fixed in a710327e

Agreed, and thank you: I had considered this during implementation and waved it off as coupling to a documented `@todo`. Your framing changed the calculus — the byte-exact comparison is what makes a stale engine dangerous rather than merely wasteful.

Removed **before** the cases run, not only in cleanup: cleaning up at the end would leave the very run that found a stale engine still using it. It sits after the fixture guard, so a suite that skips does not touch `/tmp`.

One further benefit worth recording: the cache had quietly removed coverage. With the engine kept, the `.onnx` → `.engine` build path ran only on a machine's *first ever* run of this suite, and every run after that exercised deserialization only. Each run now covers the build path.

## 3. Cleanup precedent — fixed in the PR description

Correct, `nnstreamer_filter_tvm` passes `rm -f` four explicit filenames. The description now attributes the `*.log` glob to `nnstreamer_filter_onnxruntime`/`nnstreamer_converter`/`nnstreamer_decoder` and the `-f` to `nnstreamer_filter_tvm`/`codegen`. The second commit's message carries the same loose wording; leaving it rather than force-pushing over a reviewed commit for a cosmetic point.

## Follow-up — filed as #4907

Agreed this is the actual regression guard for #4890's class of defect, and agreed it does not belong here. #4907 records it, including the design question that has to be settled first: a naive "every referenced path must exist" rule is wrong on day one, since this suite deliberately references `no_such_model.onnx` and `nnstreamer_filter_onnxruntime` deliberately references `../${PATH_TO_MODEL}`. It also notes why #4903 would not have caught this one — a suite whose sub-plugin is not built still skips, so no fixture path is ever resolved.

---
*This is a relay of a review produced by a separate AI agent; each point was re-verified against the source before acting on or disputing it.*

PR #4909, opened 2026-09-01
[Test] Close the last two vacuous single-buffer golden checks

### LINKED ISSUE #4870: [Test/onnxruntime] quantModelLaunch00 races buffer arrival against pipeline teardown

myungjoo:
Follow-up to #4889 (which fixed #4870). That PR closed this hole in the onnxruntime, tensorflow2-lite and litert filter tests; this finishes the job.

## What is left

Scanning the test tree for "connect a `new-data` handler, go `PLAYING`, go straight to `NULL`" leaves exactly two genuine cases:

| Test | Golden that can silently not run | Duration on GBS x86_64 |
| --- | --- | --- |
| `nnstreamerFilterLua.launch01` | `output[i - 1] == i * 11` | 9 ms |
| `NNStreamerFilterTVMTest.launch00` | `output[i] == 1`, `i < 10` | 58 ms |

`setPipelineStateSync (PLAYING)` returns on state-change completion, not when data has flowed, so when the teardown wins the race the handler never fires and the test passes without having compared anything. Neither test asserts that a buffer arrived, so nothing reports it.

Both tests do run in CI (Tizen GBS `unit_test 1`), so this is live behaviour rather than a theoretical hole, and the fix has a real signal.

## Fix

The same shape as #4889: count buffers through a *second* `new-data` handler, wait with the existing `wait_pipeline_process_buffers()`, then assert the count. `check_output` / `CheckOutput` stay single-purpose.

The connection-order dependency established during #4889's review is carried over into the doxygen of both new counters — `count_output` must be connected *after* the checking handler, since GObject runs same-stage handlers in connection order, so a satisfied count implies the golden comparison already ran.

## Scope decisions, stated so they can be challenged

**No meson timeout bump, unlike #4889.** That bump was needed because `unittest_filter_tensorflow2_lite` already spent 10 of its 30 allowed seconds in fixed `g_usleep()` calls *and* gained two bounded waits. Neither file here contains a `g_usleep()` at all, and each gains exactly one wait, so the worst case stays far below the 30 s default.

**Three candidates deliberately not changed**, having checked each:

- `NNStreamerFilterTVMTest.launch00_n` and `launch01_n` — end in `fakesink` with no handler connected, and assert that `PLAYING` *fails*. Correct as written.
- `join.forwardAllBuffers` — already waits, via `pop_eos_or_error (...) == GST_MESSAGE_EOS`, before asserting `received == 30`. My scanner did not recognise that as a wait; the test does. It is also in `gst/join`, which is being changed by #4875.

## Verification

- Both files compile warning-free under `-Wall -Wextra`.
- `clang-format --style=file`: clean.
- `.github/workflows/static.check.scripts/doxygen-tag.sh` run in the same advanced mode CI uses: clean — and verified with a negative control that the harness actually errors on a bad comment, since that script exits 0 on wrong arguments and a no-op run is indistinguishable from a pass. (#4889 broke that check, so it is now part of my pre-push loop.)
- The lua and tvm subplugins are not installable in this environment, so these two binaries were compile-verified but not executed locally; CI provides the runtime signal.

The mechanism itself was measured in #4889: a standalone reproducer of the pipeline shape lost the buffer in 46 of 60 iterations on a saturated machine without the wait, and 0 of 60 with it.

Test-only. No production code, no API change, so no documentation update applies.

### DIFF
--- a/tests/nnstreamer_filter_lua/unittest_filter_lua.cc
+++ b/tests/nnstreamer_filter_lua/unittest_filter_lua.cc
@@ -89,6 +89,22 @@ check_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
 }
 
 /**
+ * @brief Signal to count invocations of tensor_sink
+ * @note Connect this after check_output. GObject runs same-stage handlers in
+ *       connection order, so a satisfied count implies the golden comparison
+ *       for that buffer has already run; connecting it first would let the
+ *       waiter tear the pipeline down before check_output had a say.
+ */
+static void
+count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  guint *count = (guint *) user_data;
+  UNUSED (element);
+  UNUSED (buffer);
+  g_atomic_int_inc (count);
+}
+
+/**
  * @brief Negative test case with invalid model file path
  */
 TEST (nnstreamerFilterLua, openClose00_n)
@@ -1044,12 +1060,17 @@ TEST (nnstreamerFilterLua, launch01)
 
   GstElement *sink_handle = gst_bin_get_by_name (GST_BIN (gstpipe), "sinkx");
   EXPECT_NE (sink_handle, nullptr);
+  guint count = 0U;
   g_signal_connect (sink_handle, "new-data", (GCallback) check_output, NULL);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, &count);
 
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT * 10),
       0);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
   EXPECT_EQ (
       setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT * 10), 0);
+
+  EXPECT_GE (count, 1U);
 
   gst_object_unref (sink_handle);
   gst_object_unref (gstpipe);
--- a/tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
+++ b/tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc
@@ -159,6 +159,21 @@ class NNStreamerFilterTVMTest : public ::testing::Test
 
     gst_memory_unmap (mem_res, &info_res);
   }
+
+  /**
+   * @brief Signal handler to count invocations of tensor_sink
+   * @note Connect this after CheckOutput. GObject runs same-stage handlers in
+   *       connection order, so a satisfied count implies the golden comparison
+   *       for that buffer has already run; connecting it first would let the
+   *       waiter tear the pipeline down before CheckOutput had a say.
+   */
+  static void CountOutput (GstElement *element, GstBuffer *buffer, gpointer user_data)
+  {
+    guint *count = (guint *) user_data;
+    UNUSED (element);
+    UNUSED (buffer);
+    g_atomic_int_inc (count);
+  }
 };
 
 /**
@@ -436,12 +451,18 @@ TEST_F (NNStreamerFilterTVMTest, launch00)
 
   sink_handle = gst_bin_get_by_name (GST_BIN (gstpipe), "sink");
   EXPECT_NE (sink_handle, nullptr);
+  guint count = 0U;
   g_signal_connect (sink_handle, "new-data",
       (GCallback) NNStreamerFilterTVMTest::CheckOutput, NULL);
+  g_signal_connect (sink_handle, "new-data",
+      (GCallback) NNStreamerFilterTVMTest::CountOutput, &count);
 
   /* Test */
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&count, 1U, TEST_TIMEOUT_LIMIT_MS));
   EXPECT_EQ (setPipelineStateSync (gstpipe, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_GE (count, 1U);
 }
 
 /**


myungjoo:
## Automated code review (relayed from a sub-agent's independent analysis)

*This comment reports the findings of a separate review agent that analyzed this PR's diff, the surrounding test tree, and CI results independently of the PR author's own verification steps. Posting it here per repo review policy.*

### Scope of independent verification

**1. "Exactly two remaining vacuous cases" claim.** Manually scanned every file under `tests/` that connects a `new-data` signal (14 files: `gstreamer_join`, `nnstreamer_converter`, `nnstreamer_datarepo`, `nnstreamer_edge`, `nnstreamer_filter_custom`, `nnstreamer_filter_litert`, `nnstreamer_filter_lua`, `nnstreamer_filter_onnxruntime`, `nnstreamer_filter_shared_model`, `nnstreamer_filter_tensorflow2_lite`, `nnstreamer_filter_tvm`, `nnstreamer_if`, `nnstreamer_sink`, `tizen_sensor`). Confirmed every other call site is already protected by one of: `wait_pipeline_process_buffers()`/`_wait_pipeline_process_buffers()`, a `g_main_loop_run()` driven by a bus watch that blocks for EOS/error, `pop_eos_or_error()`, or a fixed `g_usleep()` between `PLAYING` and `NULL`. No other zero-wait "connect handler → PLAYING → NULL" golden-check site was found. Claim holds.

**2. The three excluded cases.** Opened and read each:
- `NNStreamerFilterTVMTest.launch00_n` / `launch01_n` (`tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc:471-498`): pipeline ends in `fakesink`, no `new-data` handler is connected, and the test asserts `setPipelineStateSync(..., PLAYING) != 0` (expects failure). There is no golden check that could be silently skipped. Correctly left alone.
- `join.forwardAllBuffers` (`tests/gstreamer_join/unittest_join.cc:350-386`): confirmed `EXPECT_EQ(pop_eos_or_error(pipeline, ...), GST_MESSAGE_EOS)` executes and blocks *before* `EXPECT_EQ(num_srcs * num_buffers, received)`. Already race-free. Correctly left alone.

**3. "No meson timeout bump needed."** Confirmed via `grep` that neither `unittest_filter_lua.cc` nor `unittest_filter_tvm.cc` contains any `g_usleep()` call. Confirmed both `test()` registrations (`tests/meson.build:264` for lua, `tests/nnstreamer_filter_tvm/meson.build:8` for tvm) take no explicit `timeout:` kwarg, so both fall back to Meson's 30s default. The new wait is bounded by `TEST_TIMEOUT_LIMIT_MS` (10s) and only reaches that bound in the failure case (i.e., when the bug the test exists to catch has already reproduced). Claim holds.

**4. Connection-order doxygen note.** Verified both files actually connect the checking handler first and the counting handler second (`check_output`/`CheckOutput` before `count_output`/`CountOutput`), matching the documented GObject same-stage-handlers-run-in-connection-order rationale, and matching the same convention already used in the merged #4889 fix (onnxruntime/litert/tensorflow2_lite all follow the identical shape: `UNUSED(element); UNUSED(buffer); g_atomic_int_inc(count);`).

**5. `doxygen-tag.sh` and `clang-format`, run directly.**
- First pass (against the files as checked out in this Windows worktree) produced a large flood of `[ERROR] ... Doxygen or multi line comments should begin with /**` on lines *outside* the diff. Root-caused this to a checkout artifact: this worktree has CRLF line endings (`git`'s autocrlf conversion on Windows — a known trap for this repo), and the checker's line-ending-sensitive glob check (`$line != *"*/"`) misfires on every CRLF-terminated single-line comment. This is **not a defect in the PR**.
- Re-ran against LF content pulled via `git show HEAD:<path>` (i.e., what the actual git blob / Linux CI checkout contains): clean pass, zero `[ERROR]` lines, for both files.
- This is corroborated by the live PR: the "Static checks" CI job has already reported **pass** (59s).
- `clang-format-14 --style=file` (matching the apt-installed CI version) on both files: zero diff.

**6. `gh pr checks 4909` at time of this review:** `DCO` pass, `Spell Check with Typos` pass, `Build on MacOS-latest with Meson` pass, `Static checks` pass. Everything else (including the Tizen GBS x86_64 `unit_test 1` job, which is the one that actually exercises these two tests) is still **pending/in-progress** — nothing has failed so far, but the run had not completed at review time. Recommend confirming that job goes green (in particular, that `nnstreamerFilterLua.launch01` and `NNStreamerFilterTVMTest.launch00` themselves pass) before merge.

### Findings

- **Nit:** The `count_output`/`CountOutput` helper duplicates the same ~10-line boilerplate (`UNUSED(element); UNUSED(buffer); g_atomic_int_inc(count);`) that already exists, copy-pasted per file, in the onnxruntime/litert/tensorflow2_lite tests from #4889, and now a 4th/5th time here. Not something to block on — it's consistent with the established convention in this test tree, not a regression introduced by this PR — but a shared helper in `tests/unittest_util.{h,c}` would remove the duplication across all five files if anyone revisits this area later. No action requested in this PR.

No other issues found. The change is test-only, correctly scoped (2 files, +42/-0 lines, no production code, no other module touched), mirrors the already-reviewed #4889 pattern exactly (same helper, same ordering discipline, same doxygen structure), and closes a real, measured race (documented in the PR body: 46/60 reproductions on a saturated machine without the wait, 0/60 with it) in exactly the two remaining vacuous golden checks.

### Verdict

**Approve — ok to merge**, pending the still-in-progress CI jobs (especially Tizen GBS x86_64 `unit_test 1`) finishing green.

myungjoo:
Thanks — and thanks in particular for re-deriving the "exactly two remaining" claim from your own scan of all fourteen `new-data` sites rather than taking it from the description. That is the claim most worth not trusting the author on, since it is the one that decides whether this PR is finished or arbitrary.

Also useful that you hit the CRLF artifact and diagnosed it rather than reporting dozens of phantom doxygen errors. That is a real trap in this worktree: the checkout is CRLF (`core.autocrlf=true`) while the blobs are LF, so anything run under WSL against the working tree disagrees with CI. Comparing against `git show HEAD:<path>` is the right way to reproduce what CI actually sees.

**On the Nit — the duplicated counting handler. Agreed it is real, but I am deliberately not doing it here.**

There are now five near-identical copies of that seven-line callback, and the more interesting duplication is not the code but the connection-order `@note` attached to it: that ordering constraint is currently asserted in five places, and if it ever needed correcting it would need correcting five times. A single home in `tests/unittest_util.h`, next to `wait_pipeline_process_buffers()`, is the right end state.

Two reasons it does not belong in this PR:

1. **Scope and attributability.** Hoisting the helper means editing the onnxruntime, tensorflow2-lite and litert tests, which this PR otherwise has no reason to touch, taking it from two files to six across five modules. This PR's subject is "close the last two vacuous golden checks"; a cross-module deduplication is a different subject, and mixing them means a failure in either is harder to attribute. That is the same discipline that kept #4889 reviewable.
2. **It is a refactor of already-merged code, with no behavioural change.** It gains nothing functionally and re-opens three modules that are currently green, so it should stand or fall on its own review rather than ride along with a fix.

So: worth doing, not worth doing here. I would rather leave the duplication visible for now than obscure what this PR is for. Flagging it as a deliberate deferral rather than an oversight, and I am happy to send it as its own small PR if you would like it.

CI is still running; nothing has failed so far. I will not take this out of draft until the Tizen GBS `unit_test 1` job — the one that actually executes both changed tests — reports green.

PR #4912, opened 2026-09-02
[Test] Hoist the duplicated tensor_sink buffer counter into unittest_util

### LINKED ISSUE #4910: [Test] Hoist the duplicated tensor_sink buffer counter into unittest_util
Follow-up to #4889 and #4909, raised in review of #4909 and deliberately deferred there to keep those PRs attributable.

## What

Five filter unit tests now carry a near-identical counting callback, connected as a second `new-data` handler so a test can wait for a buffer before tearing the pipeline down:

- `tests/nnstreamer_filter_onnxruntime/unittest_filter_onnxruntime.cc:136`
- `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc:121`
- `tests/nnstreamer_filter_litert/unittest_filter_litert.cc:156`
- `tests/nnstreamer_filter_lua/unittest_filter_lua.cc:99`
- `tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc:170` (as `CountOutput`, a static member of the fixture)

Each is the same seven lines:

```c
static void
count_output (GstElement *element, GstBuffer *buffer, gpointer user_data)
{
  guint *count = (guint *) user_data;
  UNUSED (element);
  UNUSED (buffer);
  g_atomic_int_inc (count);
}
```

## Why this is worth doing

The duplicated code is trivial. The duplicated **doxygen note** is not:

```
 * @note Connect this after check_output. GObject runs same-stage handlers in
 *       connection order, so a satisfied count implies the golden comparison
 *       for that buffer has already run; connecting it first would let the
 *       waiter tear the pipeline down before check_output had a say.
```

That records a real correctness constraint. Swapping the two `g_signal_connect` lines in any of these tests lets the waiter begin teardown before the checking handler has run, quietly reopening the vacuous-pass hole that #4870 / #4889 / #4909 closed — while every assertion stays green. The constraint is currently asserted in five separate places, so correcting or refining it means five edits, and a sixth caller added later has nothing to inherit it from.

## Suggested shape

A single home in `tests/unittest_util.h`, next to `wait_pipeline_process_buffers()`, which is the helper these counters exist to feed. `argmax_float()` in the same header is precedent for a `static inline` there; a real exported function in `unittest_util.c` alongside `wait_pipeline_process_buffers()` would be equally consistent — either is fine, the point is one authoritative copy carrying one authoritative note.

Two things to watch:

- The TVM copy is a `static` member of the `NNStreamerFilterTVMTest` fixture rather than a file-static, so it is not a pure textual move; the call site passes `NNStreamerFilterTVMTest::CountOutput`.
- The counter is written from the streaming thread and read by `wait_pipeline_process_buffers()` on the test thread. The write is atomic (`g_atomic_int_inc`); the read is a plain load, shared with every existing caller including `unittest_filter_custom`. Whether to make the read atomic too is a helper-wide decision that was explicitly left out of #4889 — it can be settled here or left alone, but it should be a decision rather than an accident.

## Scope

Test-only, no behavioural change, no production code, no API change. Should be mechanical once the TVM fixture case is handled. CI coverage already exists: all five tests run in the Tizen GBS `unit_test 1` job, so a mistake in the move surfaces there.

### References

- #4870 — the original race.
- #4889 — fixed it and introduced three of these copies.
- #4909 — closed the last two vacuous cases and added the remaining two copies; the review there raised this as a Nit.

myungjoo:
Fixes #4910. Follow-up to #4889 and #4909.

## What

Three test-only commits, no production code.

1. **Hoist the `tensor_sink` buffer counter into `unittest_util.h`.** The five identical `count_output` handlers in the onnxruntime, tensorflow2-lite, litert, lua and tvm filter tests are replaced by one `static inline` next to `wait_pipeline_process_buffers()`, following the `argmax_float` precedent. The TVM copy was a `static` member of the fixture; its call site now names the shared function. The connection-order note from #4889 lives in one place.
2. **Read the counter atomically in `wait_pipeline_process_buffers()`.** #4889 made the filter-test writers atomic and left the read alone because it was shared with every caller. Now that writer and waiter live in the same header, settle it: the load goes through `g_atomic_int_get()`. No caller changes.
3. **Bump every counter the waiter polls atomically** (from review). The join and converter tests carried a sixth and seventh copy of the handler; they now use the shared one. The in-place increments in the join checking callback, the custom-filter test and the tizen_sensor test become `g_atomic_int_inc()`, so every waiter-polled counter is written and read through the same atomic pair.

## Tests

Added to `unittest_common`, which is always built:

- `commonUtil.countOutput`: direct calls bump the counter once each.
- `commonUtil.countOutputPipeline`: `videotestsrc num-buffers=3 ! tensor_converter ! tensor_sink` drives the handler through a real `new-data` emission into the waiter; asserts the count is exactly 3. A checking handler connected first records whether it saw the count before every increment, so the connection-order guarantee the helper's note relies on is pinned rather than assumed.
- `commonUtil.countOutputPipelineNoBuffer_n`: `num-buffers=0` leaves the count at 0 and the waiter times out.

## Local verification (WSL, Ubuntu 22.04, x86_64)

- `unittest_common` 145/145, `unittest_join` 13/13, `unittest_filter_litert` (built with `-Dlitert-support=enabled`) 33/33.
- `unittest_filter_custom` 5/6; the one failure is the pre-existing `notRegisterFlexibleInvoke_n` path assertion on an out-of-tree build directory, unrelated to this diff.
- Negative control: swapping the two `g_signal_connect` lines in `countOutputPipeline` fails it with `out_of_order` = 3; restoring them passes.
- clang-format on the changed `.cc` files, `doxygen-tag.sh` on the changed list, and a recompile of the changed C++ objects with `-Werror=sign-compare` active all pass.
- Not buildable here: lua, tvm, onnxruntime, tensorflow2-lite, the python converter test and tizen_sensor. Those edits are deletions of a local copy or one-line increments; CI builds and runs them.

## Note on the issue text

The issue says all five tests run in the GBS `unit_test 1` job. That holds for onnxruntime, tensorflow2-lite, lua and tvm, but `packaging/nnstreamer.spec` has no `litert_support` define, so the litert test is only covered by the Ubuntu meson job (`meson test`, x86_64 leg). Both jobs run per PR, so the move is still fully covered.

### DIFF
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2748,6 +2748,117 @@ TEST (commonUtil, waitPipelineProcessBuffers_n)
 }
 
 /**
+ * @brief Test that count_output bumps the counter it is handed, once per call.
+ */
+TEST (commonUtil, countOutput)
+{
+  guint count = 0U;
+
+  count_output (NULL, NULL, &count);
+  EXPECT_EQ (count, 1U);
+
+  count_output (NULL, NULL, &count);
+  count_output (NULL, NULL, &count);
+  EXPECT_EQ (count, 3U);
+}
+
+/**
+ * @brief Shared state for the count_output pipeline tests.
+ */
+typedef struct {
+  guint count;
+  guint checked;
+  guint out_of_order;
+} count_output_test_data;
+
+/**
+ * @brief Checking handler connected before count_output.
+ * @details Records whether count_output had already run for this buffer,
+ *          which is exactly what its connection-order note rules out.
+ */
+static void
+check_before_count (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  count_output_test_data *data = (count_output_test_data *) user_data;
+  (void) element;
+  (void) buffer;
+
+  if ((guint) g_atomic_int_get (&data->count) != data->checked)
+    data->out_of_order++;
+  data->checked++;
+}
+
+/**
+ * @brief Build a pipeline that delivers num_buffers tensors to a tensor_sink named sink.
+ */
+static GstElement *
+build_count_output_pipeline (guint num_buffers)
+{
+  gchar *desc = g_strdup_printf ("videotestsrc num-buffers=%u ! video/x-raw,width=16,height=16,format=RGB,framerate=30/1 ! tensor_converter ! tensor_sink name=sink",
+      num_buffers);
+  GstElement *pipeline = gst_parse_launch (desc, NULL);
+
+  g_free (desc);
+  return pipeline;
+}
+
+/**
+ * @brief Test count_output as a tensor_sink handler feeding wait_pipeline_process_buffers.
+ * @details Also pins the connection-order guarantee the helper's note relies on:
+ *          the handler connected first sees the count before it is bumped.
+ */
+TEST (commonUtil, countOutputPipeline)
+{
+  const guint num_buffers = 3U;
+  count_output_test_data data = { 0U, 0U, 0U };
+  GstElement *pipeline = build_count_output_pipeline (num_buffers);
+  GstElement *sink;
+
+  ASSERT_TRUE (pipeline != nullptr);
+  sink = gst_bin_get_by_name (GST_BIN (pipeline), "sink");
+  ASSERT_TRUE (sink != nullptr);
+
+  g_signal_connect (sink, "new-data", (GCallback) check_before_count, &data);
+  g_signal_connect (sink, "new-data", (GCallback) count_output, &data.count);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&data.count, num_buffers, TEST_TIMEOUT_LIMIT_MS));
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_EQ (data.count, num_buffers);
+  EXPECT_EQ (data.checked, num_buffers);
+  EXPECT_EQ (data.out_of_order, 0U);
+
+  gst_object_unref (sink);
+  gst_object_unref (pipeline);
+}
+
+/**
+ * @brief Test that a pipeline delivering no buffer leaves the count at zero and times out the waiter.
+ */
+TEST (commonUtil, countOutputPipelineNoBuffer_n)
+{
+  guint count = 0U;
+  GstElement *pipeline = build_count_output_pipeline (0U);
+  GstElement *sink;
+
+  ASSERT_TRUE (pipeline != nullptr);
+  sink = gst_bin_get_by_name (GST_BIN (pipeline), "sink");
+  ASSERT_TRUE (sink != nullptr);
+
+  g_signal_connect (sink, "new-data", (GCallback) count_output, &count);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  EXPECT_FALSE (wait_pipeline_process_buffers (&count, 1U, 200U));
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  EXPECT_EQ (count, 0U);
+
+  gst_object_unref (sink);
+  gst_object_unref (pipeline);
+}
+
+/**
  * @brief Main function for unit test.
  */
 int
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -328,19 +328,6 @@ TEST (join, eosAfterRestart)
 }
 
 /**
- * @brief Callback counting the buffers that reached the sink.
- */
-static void
-count_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_data)
-{
-  guint *received = (guint *) user_data;
-  (void) element;
-  (void) buffer;
-
-  (*received)++;
-}
-
-/**
  * @brief Test that a join of N finite sources delivers every buffer downstream.
  * @detail This is the pipeline shape used by the datarepo fixtures. The
  *         branches run in their own streaming threads, so which sink pad is
@@ -373,7 +360,7 @@ TEST (join, forwardAllBuffers)
 
   sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
   ASSERT_NE (sink_handle, nullptr);
-  g_signal_connect (sink_handle, "new-data", (GCallback) count_data_cb, (gpointer) &received);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, (gpointer) &received);
 
   EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
   EXPECT_EQ (pop_eos_or_error (pipeline, UNITTEST_STATECHANGE_TIMEOUT), GST_MESSAGE_EOS);
@@ -533,7 +520,7 @@ run_pad_release_while_streaming (void)
     gst_object_unref (pipeline);
     return FALSE;
   }
-  g_signal_connect (sink_handle, "new-data", (GCallback) count_data_cb, (gpointer) &received);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, (gpointer) &received);
 
   if (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT) != 0)
     started = FALSE;
@@ -726,7 +713,7 @@ count_checked_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_dat
   gst_memory_unmap (mem_res, &info_res);
   gst_memory_unref (mem_res);
 
-  counter->received++;
+  g_atomic_int_inc (&counter->received);
 }
 
 /**
--- a/tests/nnstreamer_converter/unittest_converter.cc
+++ b/tests/nnstreamer_converter/unittest_converter.cc
@@ -359,15 +359,6 @@ const gint _test_frames2[48] = { 1101, 1102, 1103, 1104, 1105, 1106, 1107, 1108,
   1212, 1213, 1214, 1215, 1216, 1217, 1218, 1219, 1220, 1221, 1222, 1223, 1224 };
 
 /**
- * @brief Callback for tensor sink signal.
- */
-static void
-new_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_data)
-{
-  (*(guint *) user_data)++;
-}
-
-/**
  * @brief Test for dynamic dimension of the custom converter
  */
 TEST (tensorConverterPython, dynamicDimension)
@@ -405,7 +396,7 @@ TEST (tensorConverterPython, dynamicDimension)
   EXPECT_NE (sink_handle, nullptr);
 
   data_received = (guint *) g_malloc0 (sizeof (guint));
-  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, data_received);
+  g_signal_connect (sink_handle, "new-data", (GCallback) count_output, data_received);
 
   buf_0 = gst_buffer_new_wrapped (_g_memdup (_test_frames1, 96), 96);
   buf_1 = gst_buffer_new_wrapped (_g_memdup (_test_frames2, 192), 192);
@@ -486,7 +477,7 @@ new_data_cb_json (GstElement *element, GstBuffer *buffer, gpointer user_data)
   gboolean mapped;
   GstMemory *mem_res;
   gchar *output;
-  (*(guint *) user_data)++;
+  g_atomic_int_inc ((guint *) user_data);
 
   num_tensors = gst_buffer_n_memory (buffer);
   EXPECT_EQ (2U, num_tensors);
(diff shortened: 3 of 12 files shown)


myungjoo (commented):
(다른 Agent가 수행한 코드 리뷰 결과를 옮겨 적은 것임 / relayed from an automated agent review.)

**Verdict: No blocking issues found — approve (with 2 minor nits below).** Test-only change, correctly closes #4910, scope is tight (6 files under `tests/`, no production code), and CI coverage was independently verified.

## Checks performed

- Diff review of both commits (`cc926362`, `04849c1e`) via `git show`, plus `gh pr diff 4912`.
- Grepped every caller of `wait_pipeline_process_buffers()` under `tests/` (`unittest_join.cc`, `unittest_converter.cc`, `unittest_filter_custom.cc`, `unittest_tizen_sensor.cc`, `unittest_sink.cc`'s unrelated local `_wait_pipeline_process_buffers`) — no caller needs a signature/type change; all pass a `guint*`.
- Confirmed `c_std=gnu89` is the project-wide C standard (`meson.build:11`) and that the new `static inline void count_output(...)` in `tests/unittest_util.h:94-100` follows the exact precedent of the pre-existing `static inline gsize argmax_float(...)` in the same header — `static inline` linkage/inlining semantics for a `static` function do not differ between gnu89 and C99/C++17 (the gnu89-vs-C99 "inline" divergence only affects non-static/extern inline functions), and GCC does not warn `-Wunused-function` on unused `static inline` functions regardless of dialect. No `UNUSED` macro dependency was introduced into `unittest_util.h` — the new function uses plain `(void) element; (void) buffer;` instead, so the header does not need to pull in `nnstreamer_util.h`/`mqttcommon.h`'s `UNUSED` macro.
- Verified `g_atomic_int_get()`-on-`guint*` (as opposed to `gint*`) is an already-established pattern in this codebase (`gst/mqtt/mqttsink.c:646` on `mqtt_sink_state_t`, and this very PR's own `tests/common/unittest_common.cc:2786` on a `guint`), and the author's local run reports `unittest_common: 145/145 pass`, i.e. it already builds clean in their toolchain.
- Verified the CI coverage claim in the PR body by reading, not just grepping, `.github/workflows/gbs_build.yml` and `packaging/nnstreamer.spec`:
  - `gbs_build.yml` builds x86_64 with `--define "unit_test 1"`; the spec's x86_64 `%ifarch` block (lines ~25-39) sets `tensorflow2_lite_support`, `lua_support`, `tvm_support`, `onnxruntime_support` all to `1` (all other arches set them to `0`), and the `%check` block (`packaging/nnstreamer.spec:1060` `%if 0%{?unit_test}` → `bash %{test_script} ./tests`) runs every unittest binary under `tests/`, including `unittest_common` and the four filter tests above, via `packaging/run_unittests_binaries.sh`.
  - `litert` does not appear anywhere in `packaging/nnstreamer.spec` (confirmed via grep — zero matches), so the Tizen GBS job never builds or runs the litert test, exactly as the PR body states.
  - `.github/workflows/ubuntu_clean_meson_build.yml` runs on `pull_request`, downloads the LiteRT SDK and asserts `libnnstreamer_filter_litert.so` exists, then runs `meson test -C build/ -v` (line 157), and `tests/meson.build:416-424` registers `unittest_filter_litert` as a meson test gated on `litert_support_is_available`. `tests/meson.build:75-83` registers `unittest_common` unconditionally. So both new-test suites and the litert case are exercised on every PR, across the two jobs together — the PR's coverage claim checks out.
- Confirmed the referenced predecessor commit `f888c072` ("[Test] Increment the sink counters atomically and pin the handler order") exists in history, so commit `04849c1e`'s changelog claim is accurate.
- Read `tests/nnstreamer_filter_custom/unittest_filter_custom.cc` and `tests/gstreamer_join/unittest_join.cc` to check the "correct against a plain or a locked store alike" claim in `04849c1e` (see Nit 1 below).

## Nit 1 — commit message slightly overclaims correctness for non-atomic writers

`tests/unittest_util.c:199` (commit `04849c1e`), commit message: "the atomic load is correct against a plain or a locked store alike."

This isn't quite true in the strict sense for two existing callers:
- `tests/nnstreamer_filter_custom/unittest_filter_custom.cc:90`: `cbdata->sink_received = cbdata->sink_received + 1;` — a **non-atomic** store, only serialized by `g_mutex_lock`/`unlock` around the write; the new `g_atomic_int_get()` read in the waiter is never taken under that same mutex.
- `tests/gstreamer_join/unittest_join.cc:336-340` (`count_data_cb`): `(*received)++;` — no lock and no atomic op at all on the write side.

Mixing an atomic load with a non-atomic (or mutex-only-guarded, but not atomic) store to the same object from different threads without a shared synchronization point is technically a data race under the C11/GNU memory model, even though it is benign in practice on every architecture this project targets (naturally-aligned 32-bit read/write, no tearing). This is **not a regression** — the previous plain `*data_received` read already raced with these same writers — so it doesn't block the PR. Suggest either softening the commit message (e.g. "no worse than the previous plain read for callers whose writer isn't itself atomic") or filing a small follow-up to make `sink_received`/`received` atomic too, in the same spirit as consolidating the one authoritative note that #4910 is about.

## Nit 2 — litert is exercised by only one CI leg, worth double-checking in future PRs

`packaging/nnstreamer.spec` has no `litert_support` define at all (confirmed by grep — zero hits), so `unittest_filter_litert` is covered exclusively by `ubuntu_clean_meson_build.yml`'s `meson test` leg (`.github/workflows/ubuntu_clean_meson_build.yml:157`), not by the Tizen GBS `unit_test 1` job. The PR body already discloses this accurately, so this is just a confirmation, not a new finding — but it means a future change that only breaks litert would surface on a single workflow rather than two, so it's worth keeping in mind if that Ubuntu meson job is ever made non-required or skipped for draft/WIP PRs.

No issues found with: test correctness (the `check_before_count`/`count_output` connection-order test in `tests/common/unittest_common.cc:2807-2833` genuinely pins the ordering guarantee — flipping the two `g_signal_connect` calls, per the author's own negative-control check, fails `out_of_order`), scope (test-only, 134/-82 lines, all under `tests/`), or documentation (no API/architecture change, so no doc update is required).

myungjoo:
(relayed on behalf of the author by a Claude agent)

Thanks for the review. Both nits acted on, plus a CI fix; force-pushed as `9766ba9f`.

**Nit 1 (non-atomic writers, `unittest_util.c:199`)** — resolved in-PR rather than softened. New third commit `[Test] Bump every counter the waiter polls atomically`:
- `unittest_join.cc`: `count_data_cb` was a sixth copy of the hoisted handler, so it is replaced by the shared `count_output`; the in-place `counter->received++` in `count_checked_data_cb` becomes `g_atomic_int_inc`.
- `unittest_converter.cc`: `new_data_cb` was a seventh copy, replaced the same way; `new_data_cb_json`'s increment becomes `g_atomic_int_inc`.
- `unittest_filter_custom.cc`: `sink_received` is now bumped with `g_atomic_int_inc` (the mutex stays, it guards the other fields).
- `unittest_tizen_sensor.cc`: `checked += 1` becomes `g_atomic_int_inc`.
The second commit's message no longer claims correctness against plain stores; it now says the remaining writers are converted in the next commit.

**Nit 2 (litert only covered by the Ubuntu meson job)** — agreed, and already stated in the PR body; no change.

**CI on the first push** — three build jobs failed on one line in the new order-check handler: `g_atomic_int_get()` returns `gint`, compared against a `guint` under `-Werror=sign-compare`. Cast added. Static checks and CPP-Linter failed on one clang-format line (the `g_strdup_printf` call in the new pipeline helper), also fixed. Both fixes are squashed into the first commit, where the code was introduced. Local build masked the sign-compare because it carries `-Wno-sign-compare` for a non-system gtest; re-verified by recompiling the four changed C++ objects with that flag stripped.

Local re-run after the changes (WSL, x86_64): `unittest_common` 145/145, `unittest_join` 13/13, `unittest_filter_litert` 33/33; `unittest_filter_custom` 5/6 with the one failure being the pre-existing `notRegisterFlexibleInvoke_n` path assertion (`build/tests/nnstreamer_example/...` does not exist in an out-of-tree build dir), unrelated to this diff. `unittest_converter` and `unittest_tizen_sensor` are not buildable here; CI covers them.

myungjoo (commented):
(다른 Agent가 수행한 코드 리뷰 결과를 옮겨 적은 것임 / relayed from an automated agent review.)

**Verdict: No blocking issues found — approve.** This re-review covers the full updated diff at head `9766ba9f` (3 commits: `701343fe`, `b5e32d9b`, `9766ba9f`), with focus on the new third commit and the two fixes squashed into the first commit per the author's force-push comment.

## What was independently verified

- Re-read all three commits in full (`git show <sha>`) and cross-checked against `gh pr view 4912 --comments`.
- Ran `clang-format -style=file` (v16.0.6, matching repo `.clang-format`) against every touched `.cc` file (`unittest_join.cc`, `unittest_converter.cc`, `unittest_filter_custom.cc`, `unittest_tizen_sensor.cc`, `unittest_common.cc`, and the five filter tests from commit 1) — zero diff in all of them, confirming the squashed clang-format fix is correct and complete. (`unittest_util.c`/`.h` are K&R/gst-indent files, not clang-format targets — clang-format flags pre-existing, unrelated formatting there that predates this PR; not a real issue.)
- Grepped every caller of `wait_pipeline_process_buffers()` across `tests/` and confirmed all of them now write their counter with `g_atomic_int_inc`/the shared `count_output` and read it via `g_atomic_int_get` in `tests/unittest_util.c:199`. The previous review's "atomic load vs plain store" data race is fully resolved for every caller of the shared helper:
  - `unittest_join.cc`: `forwardAllBuffers` / `run_pad_release_while_streaming` now use the shared `count_output`; `count_checked_data_cb`'s `counter->received++` → `g_atomic_int_inc (&counter->received)`.
  - `unittest_converter.cc`: `dynamicDimension` now uses `count_output`; `new_data_cb_json`'s increment → `g_atomic_int_inc`.
  - `unittest_filter_custom.cc`: `sink_received` increment → `g_atomic_int_inc` (mutex kept for the other fields).
  - `unittest_tizen_sensor.cc`: `checked += 1` → `g_atomic_int_inc (&vdata->checked)`.
  - `unittest_sink.cc`'s local `_wait_pipeline_process_buffers`/`g_test_data.received` is a separate, unrelated implementation not touched by this PR — correctly out of scope.
- Checked the specific race patterns called out for this review:
  - `count_checked_data_cb` (join): reads `counter->received` (plain) to index `expected[]` and calls `ASSERT_LT`, then does `g_atomic_int_inc` at the end. This is safe: `counter->received` is written only by this callback, which fires only on the single GStreamer streaming thread; the only other reader is the test thread via `g_atomic_int_get()` in the waiter, and the waiter never sees a torn value. No cross-thread race on the plain read.
  - `sink_received` (filter_custom): atomic-inc'd under the mutex, then plain-read later in the same callback (`expected_size = sizeof(guint) * cbdata->sink_received;`, and the `EXPECT_EQ` against `meta.dimension[0]`). Same reasoning — `new_data_cb` runs only on the streaming thread, so the plain read-after-its-own-atomic-write is not a race. The mutex is retained correctly for `filter_received`, which is shared with the filter's invoke callback.
  - `checked` (tizen_sensor) is declared `unsigned int`, and `g_atomic_int_inc()` is passed `&vdata->checked`. On GCC/Clang (the only compilers this project's CI uses — no MSVC job), `g_atomic_int_inc` is a `G_GNUC_EXTENSION` statement-expression macro keyed on `sizeof(*atomic) == sizeof(gint)` plus a GCC/Clang atomic builtin, not a plain function prototyped as `void (gint *)`; it does not require exact `gint*`/`guint*` type identity. This mirrors the already-established pattern in this codebase (`g_atomic_int_get()` on `guint*` in `mqttsink.c` and in this PR's own `unittest_util.c`), confirmed by the previous review. No new portability risk.
  - `unittest_converter.cc`: `new_data_cb` (the counting handler) is fully deleted; `grep -rn "new_data_cb\b"` across `tests/` shows no dangling reference to it (a different, unrelated `new_data_cb` still exists in `unittest_join.cc`, `unittest_sink.cc`, `unittest_edge.cc`, etc. — distinct functions, not stale references to the removed one). `count_data_cb` (join) likewise has zero remaining references after removal.
- Verified the squashed sign-compare fix at `tests/common/unittest_common.cc:2783`: `check_before_count` compares `(guint) g_atomic_int_get (&data->count) != data->checked` — the explicit `(guint)` cast on the `gint` return value correctly suppresses `-Werror=sign-compare` against the `guint data->checked`, and is logically a no-op cast (both are 32-bit, same bit pattern) — correct fix, not just a silence-the-warning hack.
- Scope: third commit touches exactly 4 files, all under `tests/`, net -22 lines, no production code, no build-file changes. Consistent with #4910's scope (removing duplicate per-test counting handlers) rather than expanding it — replacing `count_data_cb`/`new_data_cb` (6th/7th copies) with the shared `count_output` is squarely in scope; the 2 in-place atomic conversions for `sink_received`/`checked` are a minimal, directly-related follow-up to the second commit's atomic read, not scope creep.
- Test coverage / CI: `unittest_join` and `unittest_filter_custom` are registered in `tests/meson.build` unconditionally (gated only on `gtest_dep.found()`/`gst18_dep.found()`, always true), so both the GBS x86_64 job (`unit_test 1`) and `ubuntu_clean_meson_build.yml`'s `meson test` run them on every PR. `unittest_converter` requires `flatbuf_support_is_available and have_python3` — a pre-existing gate untouched by this PR. `unittest_tizen_sensor` is only built with `-Denable-tizen-sensor=true` (default `false`, not set by any Ubuntu workflow) and is only invoked in `%check` under a Tizen GBS build via `%if (%{with tizen} && ...)`; this is the same category of partial-CI-coverage caveat the author already disclosed for litert in Nit 2 of the previous review, pre-existing and not introduced by this PR — noted for awareness only, not a new defect.
- Documentation: no API/architecture change; none required.

## Previous nits — status

Both nits from review `5085162522` are resolved as the author's comment describes:
- **Nit 1** (commit message overclaiming atomic-vs-plain-store correctness): resolved in-code rather than just reworded — every remaining plain-store writer is now atomic, and the second commit's message was reworded to say so. Independently confirmed no writer of a waiter-polled counter remains non-atomic.
- **Nit 2** (litert covered by only one CI leg): unchanged, as agreed — already disclosed in the PR body, not something this PR needs to fix.

No new findings. Nothing to flag at Blocking/Should-fix/Nit severity beyond the CI-coverage observation above, which is informational only.

PR #4913, opened 2026-09-02
[CI] Put apt-get update on its own line so its failure fails the step

### LINKED ISSUE #4887: [CI] apt-get update failures are still masked by "&&" in 5 workflows
Follow-up to #4877, which fixed this class of bug in `pdebuild.yml` and `update_pbuilder_cache.yml` only. The same defect remains in five other workflows.

## The rule

GitHub runs a `run:` block with no explicit `shell:` under `/usr/bin/bash -e` (confirmed in a job log: `shell: /usr/bin/bash -e {0}`). Under errexit, a command that is part of an `&&` list and is **not** the command following the final `&&` does **not** cause the shell to exit:

```
$ bash -e -c 'false && true; echo reached'
reached
```

So `sudo apt-get update && sudo apt-get install ...` short-circuits on a failed update, errexit never fires, and the step's exit status comes from whatever command runs last in that block. In #4877 the step reported SUCCESS and the job died ~20 s later in an unrelated step with `sudo: pbuilder: command not found`.

## Where it still applies

This only masks when the `&&` list is **not** the last command of the step script. Sites on `main` (bba342cb):

| File | Line | What follows in the same block |
|---|---|---|
| `.github/workflows/codeql.yml` | 76 | further `apt-get install`, `meson setup/compile` |
| `.github/workflows/codeql.yml` | 78 | same block, `add-apt-repository && apt-get update && apt-get install` |
| `.github/workflows/gbs_build.yml` | 42 | `cp .github/workflows/tizen.gbs.conf ~/.gbs.conf` |
| `.github/workflows/nns_testhub.yml` | 24 | `cp .github/workflows/tizen.gbs.conf ~/.gbs.conf` |
| `.github/workflows/publish_docs.yml` | 20 | `pip install hotdoc cmake==3.22` |
| `.github/workflows/ubuntu_clean_meson_build.yml` | 31 | further `apt-get install`, `pip install` |
| `.github/workflows/ubuntu_clean_meson_build.yml` | 33 | same block |

## Where it does NOT apply — please leave these alone

These are single-line `run:` steps. The `&&` list's own exit status *is* the step's exit status, so the failure is already reported honestly and splitting them changes nothing:

- `.github/workflows/publish_docs.yml:48`
- `.github/workflows/static.check.yml:24`
- `.github/workflows/update_gbs_cache.yml:40`

## Why this was not just swept in #4877

The fix is one line per site, but the *consequence* is not uniform, so it needs per-file judgment:

**Tier 1 — safe, same profile as pdebuild.** `gbs_build.yml:42` and `nns_testhub.yml:24` install `gbs` / `rpm2cpio` / `cpio`, none of which ship on the GitHub runner image. A failed `apt-get update` already dooms those jobs, so splitting only moves the failure to its true cause and labels it correctly. No pass/fail change.

**Tier 2 — needs thought.** `codeql.yml:76,78`, `ubuntu_clean_meson_build.yml:31,33` and `publish_docs.yml:20` install packages that are *already preinstalled* on `ubuntu-22.04` runners (`cmake`, `pkg-config`, `flex`, `bison`, `python3-dev`, `git`, `binutils`, … — per the `actions/runner-images` manifest). Today a transient mirror hiccup there can be survived: the install is skipped, the already-present dependency satisfies the build, and the job goes green. Splitting converts those runs into hard failures. That is arguably more correct, but it is a real change in CI pass rate and should be a deliberate decision, ideally paired with the retry that #4877 added:

```bash
for i in 1 2 3; do
  sudo apt-get update && break
  if [ "$i" = 3 ]; then
    echo "::error::apt-get update failed after $i attempts"
    exit 1
  fi
  sleep $((i * 10))
done
```

## Suggested resolution

1. Split Tier 1 as-is — low risk, immediate benefit.
2. For Tier 2, split **and** add the retry loop in the same change, so honest failure does not come at the cost of a worse flake rate.

Also worth considering while touching these: #4877 found that the runner ships the `packages.microsoft.com` sources in two formats (`microsoft-prod.list` and `azure-cli.sources`), and that host intermittently answers 403. Any workflow that does not need those repos can drop them the same way before updating.

Verification note: the classification above was checked by reading each step in full rather than by grepping for the pattern, since the single-line vs. multi-line `run:` distinction is what decides whether a site is actually buggy.

myungjoo:
Fixes #4887.

Under `bash -e`, a command that is not the last of an `&&` list does not abort the shell when it fails, so `sudo apt-get update && sudo apt-get install ...` followed by another command reports success on a broken mirror and the job dies later on a missing tool. #4877 fixed two workflows; this PR fixes the rest by putting `apt-get update` on a line of its own.

## Sites

The seven listed in the issue, plus two more of the same shape that already existed at the issue's base commit:

- `docker_build_test.yml:46`, identical to `gbs_build.yml:42`.
- The `yocto.yml` disk-space cleanup, where a failed `apt-get update --fix-missing` silently skipped the whole `remove && purge && autoremove && rm -rf` chain. That update is removed rather than split: `apt-get remove`/`purge`/`autoremove` resolve against installed package state, the step deletes `/var/lib/apt/lists` at the end anyway, and keeping it on its own line would add a new way for a mirror hiccup to fail the yocto job before the build starts.

Left alone, as the issue asks: the single-line steps in `publish_docs.yml:49`, `static.check.yml:24`, `update_gbs_cache.yml:40`, and the `&& \` step in `ubuntu_clean_llvm_build.yml`, where the chain's status is the step's status.

## Tests

No GTest or SSAT case: the change is CI-workflow-only and touches no code. The CI run of this PR exercises every changed step. Earlier heads of this PR carried a composite action with retries and a static checker with self-tests; they were dropped as out of proportion to a one-line-per-site fix.

## Things to know

As the issue anticipated, `codeql`, `ubuntu_clean_meson_build` and `publish_docs` used to survive a transient mirror failure because the packages they install are largely preinstalled. They now fail on the first failed update. Unlike `pdebuild`, which #4877 gave a three-attempt retry, there is no retry here; a red run from a mirror hiccup is re-run by hand. That is a deliberate choice of the smallest fix over a softer failure mode, and it can be revisited if such reds turn out to be frequent.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -73,9 +73,12 @@ jobs:
     - if: matrix.build-mode == 'manual'
       shell: bash
       run: |
-        sudo apt update && sudo apt install -y libglib2.0-dev libjson-glib-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libunwind-dev googletest
+        sudo apt update
+        sudo apt install -y libglib2.0-dev libjson-glib-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libunwind-dev googletest
         sudo apt-get install -y liborc-0.4-dev flex bison libopencv-dev pkg-config python3-dev python3-numpy python3 python3-pip
-        sudo add-apt-repository -y ppa:nnstreamer/ppa && sudo apt-get update && sudo apt-get install -y ssat libpaho-mqtt-dev
+        sudo add-apt-repository -y ppa:nnstreamer/ppa
+        sudo apt-get update
+        sudo apt-get install -y ssat libpaho-mqtt-dev
         sudo apt-get install -y gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base libgtest-dev libpng-dev
         sudo apt install nnstreamer-edge-dev flatbuffers-compiler libflatbuffers-dev protobuf-compiler libprotobuf-dev tensorflow2-lite-dev onnxruntime-dev liblua5.1-0-dev
         pip install meson ninja
--- a/.github/workflows/docker_build_test.yml
+++ b/.github/workflows/docker_build_test.yml
@@ -43,7 +43,8 @@ jobs:
     - name: prepare GBS
       run: |
         echo "deb [trusted=yes] http://download.tizen.org/tools/latest-release/Ubuntu_22.04/ /" | sudo tee /etc/apt/sources.list.d/tizen.list
-        sudo apt-get update && sudo apt-get install -y gbs
+        sudo apt-get update
+        sudo apt-get install -y gbs
         cp .github/workflows/tizen.gbs.conf ~/.gbs.conf
 
     - name: get date
--- a/.github/workflows/gbs_build.yml
+++ b/.github/workflows/gbs_build.yml
@@ -39,7 +39,8 @@ jobs:
       if: env.rebuild == '1'
       run: |
         echo "deb [trusted=yes] http://download.tizen.org/tools/latest-release/Ubuntu_22.04/ /" | sudo tee /etc/apt/sources.list.d/tizen.list
-        sudo apt-get update && sudo apt-get install -y gbs
+        sudo apt-get update
+        sudo apt-get install -y gbs
         cp .github/workflows/tizen.gbs.conf ~/.gbs.conf
 
     - name: get date
--- a/.github/workflows/nns_testhub.yml
+++ b/.github/workflows/nns_testhub.yml
@@ -21,7 +21,8 @@ jobs:
       run: |
         ubuntu_version=`lsb_release -rs`
         echo "deb [trusted=yes] http://download.tizen.org/tools/latest-release/Ubuntu_$ubuntu_version/ /" | sudo tee /etc/apt/sources.list.d/tizen.list
-        sudo apt-get update && sudo apt-get install -y gbs rpm2cpio cpio
+        sudo apt-get update
+        sudo apt-get install -y gbs rpm2cpio cpio
         cp .github/workflows/tizen.gbs.conf ~/.gbs.conf
 
     - name: Run nnstreamer gbs build
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -17,7 +17,8 @@ jobs:
         python-version: '3.11'
     - name: Install packages for hotdoc
       run: |
-        sudo apt-get update && sudo apt-get install -y python3-dev libxml2-dev libxslt1-dev cmake libyaml-dev libclang-dev llvm-dev libglib2.0-dev libjson-glib-dev flex
+        sudo apt-get update
+        sudo apt-get install -y python3-dev libxml2-dev libxslt1-dev cmake libyaml-dev libclang-dev llvm-dev libglib2.0-dev libjson-glib-dev flex
         python -m pip install --upgrade pip
         pip install hotdoc cmake==3.22
 
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -28,9 +28,12 @@ jobs:
     - name: install requirements
       if: env.rebuild == '1'
       run: |
-        sudo apt-get update && sudo apt-get install -y libglib2.0-dev libjson-glib-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libunwind-dev googletest gdb
+        sudo apt-get update
+        sudo apt-get install -y libglib2.0-dev libjson-glib-dev libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev libunwind-dev googletest gdb
         sudo apt-get install -y liborc-0.4-dev flex bison libopencv-dev pkg-config python3-dev python3-numpy python3
-        sudo add-apt-repository -y ppa:nnstreamer/ppa && sudo apt-get update && sudo apt-get install -y ssat libpaho-mqtt-dev
+        sudo add-apt-repository -y ppa:nnstreamer/ppa
+        sudo apt-get update
+        sudo apt-get install -y ssat libpaho-mqtt-dev
         sudo apt-get install -y valgrind gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base libgtest-dev libpng-dev libc6-dbg binutils-x86-64-linux-gnu-dbg valgrind-dbg
         sudo apt-get install -y libarmnn-dev libarmnntfliteparser-dev libarmnn-cpuref-backend22
         # The armnn subplugin auto-selects the NEON-accelerated CpuAcc backend on arm.
--- a/.github/workflows/yocto.yml
+++ b/.github/workflows/yocto.yml
@@ -78,8 +78,7 @@ jobs:
             echo "Enough free disk space (${avail_gb}G); skipping cleanup."
             exit 0
           fi
-          sudo apt-get update --fix-missing \
-            && sudo apt-get remove -y '^dotnet-.*' '^llvm-.*' '^mysql-server-core-.*' '^postgresql-.*' '^temurin-.*' \
+          sudo apt-get remove -y '^dotnet-.*' '^llvm-.*' '^mysql-server-core-.*' '^postgresql-.*' '^temurin-.*' \
             azure-cli google-chrome-stable google-cloud-cli firefox powershell microsoft-edge-stable mono-devel \
             && sudo apt-get purge docker-ce docker-ce-cli \
             && sudo apt-get autoremove -y \


myungjoo (commented):
This review was transcribed on behalf of a separate AI review agent (Claude), which analyzed the diff, the affected workflows, and the linked issue independently of this repository's human reviewers. All findings below are the agent's own; I verified the local test scripts still pass in this checkout (`.github/actions/apt-update/test_apt_update.sh` 14/14, `.github/workflows/static.check.scripts/test_apt_update_chain.sh` 18/18) and cross-checked the workflow YAML before relaying this.

## 1. Does it resolve issue #4887?

**Confirmed OK.** All 7 sites listed in the issue (`codeql.yml:76,78`, `gbs_build.yml:42`, `nns_testhub.yml:24`, `publish_docs.yml:20`, `ubuntu_clean_meson_build.yml:31,33`) are converted to use the new `./.github/actions/apt-update` composite action, which retries `apt-get update` and never leaves it inside an `&&` chain. `grep -rn "apt-get update\|apt update" .github/workflows/*.yml .github/actions/*/*.yml` after the diff returns no hits outside the new action's own name/description and the new checker's step name — i.e. no site of the pattern remains anywhere in the tree. The PR also picks up two sites of the same shape that existed at the issue's base commit but weren't enumerated (`docker_build_test.yml:46`, `yocto.yml` build step), which is a reasonable and disclosed extension of the same fix.

## 2. Regression / cross-module impact

**Major — disclosed by the author, but needs explicit maintainer sign-off, not just a PR-description footnote.** The PR converts the three "Tier 2" sites (`codeql.yml`, `ubuntu_clean_meson_build.yml`, `publish_docs.yml`) that the issue explicitly said were previously tolerant of a transient `apt-get update` failure, because the packages they install are largely preinstalled on the runner image. After this change those steps fail honestly instead. Concretely:
- `codeql.yml` is triggered on `push`, `pull_request` **and** `schedule` (`.github/workflows/codeql.yml:12-18`), so this is not a background-only job — it runs on every PR.
- `ubuntu_clean_meson_build.yml` is `pull_request`-triggered and is presumably the project's core build/test gate.
- Previously a flaky mirror there degraded silently (packages already present); after this PR it becomes a hard, retried-but-still-possible CI failure on unrelated PRs.

The PR body ("Things to know") already states this is a deliberate, anticipated trade-off, and the same trade-off was already accepted for `pdebuild.yml`/`update_pbuilder_cache.yml` in #4877 without apparent fallout. Still, since this expands that trade-off from 2 workflows to the project's main PR-gating build workflow plus CodeQL, this deserves an explicit maintainer acknowledgment before merge (the PR is already DRAFT/DO NOT MERGE, which is appropriate given this).

**Minor** — same effect on `nns_testhub.yml:20-32`: the issue itself notes the "doomed anyway" Tier-1 argument is weaker there because the self-hosted runner may already have `gbs` installed from a previous job, so a transient update failure that used to be silently survived can now fail the job. Lower severity than the above because this is a scheduled job, not something that blocks a contributor's PR.

No functional/runtime regression was found in the "safe" tier: `apt-update.sh` (`.github/actions/apt-update/apt-update.sh:33-38`) reproduces the exact `packages.microsoft.com`-source-removal behavior that `pdebuild.yml`/`update_pbuilder_cache.yml` already had inline, and `add-apt-repository -n` (now added to `codeql.yml`, `ubuntu_clean_meson_build.yml`) only removes a *redundant* second `apt-get update`, it does not change what gets installed. Checkout-before-`uses: ./.github/actions/apt-update` ordering is correct in every one of the 13 touched workflows (verified by grepping line order), and every `if: env.rebuild == '1'` / `steps.rebuild.outputs.rebuild == '1'` gate on the new steps matches the gate on the step it replaces, so conditional execution is preserved everywhere it existed before.

## 3. Code size vs. scope

**Minor.** 526 insertions / 57 deletions across 18 files is large for what the issue frames as "split one `&&`", but most of it is: (a) one shared composite action + its self-test, replacing what would otherwise be ~9 duplicated retry loops (matches the existing `./.github/actions/check-rebuild` precedent), and (b) a static checker + its self-test that turns this into an enforced invariant rather than a one-time fix. That's proportionate to fixing 9 sites and preventing a 10th from reappearing.

The one genuine scope expansion beyond the issue is converting the three sites the issue explicitly asked to leave alone (`publish_docs.yml:48`, `static.check.yml:24`, `update_gbs_cache.yml:40`) — the issue said these already report failure honestly as single-line steps, so nothing there needed fixing. The PR converts them anyway "so the checker can apply one rule with no exceptions" (PR body). That's a defensible simplification (the new checker is deliberately stricter than the bug requires, and says so in its own header comment at `.github/workflows/static.check.scripts/apt-update-chain.sh:17-19`), but it is scope beyond the stated issue and worth being a conscious call, not an incidental one.

No unrelated modules are touched; every file changed is `.github/workflows/*` or the two new `.github/actions` / `static.check.scripts` locations.

## 4. Test completeness

**Confirmed OK**, and above the bar this repo's own precedents (`meson-version.sh`) set.
- `.github/actions/apt-update/test_apt_update.sh` fakes `sudo`/`apt-get`/`sleep` on `PATH` and pins: pass on first success, exact retry count and backoff (10s, 20s, none after the last attempt), nonzero exit and `::error::` annotation after exhausting retries, that only `packages.microsoft.com` sources are removed and only when `DROP_MICROSOFT_SOURCES` isn't `false`, and that a missing sources directory isn't an error. Ran locally: 14/14 assertions pass.
- `.github/workflows/static.check.scripts/test_apt_update_chain.sh` pins every chained shape that has existed in this repo (single-line, multi-line, `apt` vs `apt-get`, chain starting at `add-apt-repository`, backslash-continued, flag before `update`, `.yaml` extension) as rejected, and the safe shapes (own-line update, action call, `::group::` title, comment, unrelated `&&` chain) as accepted, then runs the checker against the actual shipped workflow files. Ran locally: 18/18 assertions pass, including "no shipped workflow chains apt-get update" against the real tree.

I could not independently verify the PR description's shellcheck claim (no `shellcheck` binary available in this environment), but the repo's own `/Checker/ shellcheck for shell scripts` step (`static.check.yml:285-288`) runs against `$changed_file_list`, which includes all 4 new scripts since they're newly added files, so CI itself will catch it if that claim is wrong.

## 5. Detecting future regressions from other modules

**Confirmed OK — this is the strongest part of the PR.** `apt-update-chain.sh` doesn't scope itself to `$changed_file_list`; it scans every `*.yml`/`*.yaml` directly under `.github/workflows` (`apt-update-chain.sh:65`), unconditionally, on every `pull_request` run of `static.check.yml`. So a future PR that adds a brand-new workflow (or edits an unrelated one) with `apt-get update && apt-get install ...` will be caught even though this PR never touched that file. That directly satisfies "detect impact from future changes elsewhere in the repo," which a `$changed_file_list`-scoped check would not.

## 6. Can CI actually block a bad merge?

**Confirmed OK**, modulo one caveat. The three new steps in `static.check.yml` (`static.check.yml:246-283`) reuse the exact merge-ref guard pattern already established by `meson-version.sh`'s self-test wiring: they run the checker/self-tests if present, emit an actionable `::error::` if the script was deleted by this PR without removing its workflow step, and skip cleanly if the branch predates the script. A checker or self-test failure fails the step, which fails the `static.check.yml` job. Per `Documentation/contributing.md`'s merge criteria ("passed all the tests... a necessary condition"), that should block merge, assuming `static.check.yml` is a required status check as its established role in this repo implies — I could not verify GitHub branch-protection configuration directly (that's outside repo file contents), so this is inferred from convention rather than confirmed from settings.

## 7. Architecture / API documentation

**Confirmed OK — not applicable.** This is CI-internal tooling; no public API or documented architecture changes. `AGENTS.md`/`CLAUDE.md` reference `static.check.yml` only in passing ("NNStreamer uses CI-based format checking") and don't enumerate its steps, so no doc update is owed here.

## Other notes

- `.github/workflows/yocto.yml:81-88` — **Nit.** The disk-cleanup step's `apt-get remove && apt-get purge && apt-get autoremove && rm -rf ...` chain is left as-is (per the PR body, intentionally: it's the whole step, so its own exit status is already honest, and it removes `/var/lib/apt/lists` at the end anyway, which is why the preceding `apt-get update --fix-missing` was dropped rather than routed through the action). Confirmed this reasoning is correct — `apt-get remove/purge/autoremove` operate on installed-package state and don't need a fresh index. No functional issue, but since it's now the only remaining `&&`-chained apt invocation in the tree and superficially resembles the pattern the rest of the PR just eliminated, a one-line comment noting why it's exempt would save a future contributor a double-take.
- `.github/actions/apt-update/action.yml:1-16` — Confirmed OK: composite action steps require an explicit `shell:`, and it's set to `bash` (giving `bash -eo pipefail`, distinct from the implicit no-`shell:` default of plain `bash -e` that the issue itself documents) — this distinction is used correctly and consistently with the self-test's own docstring.
- Checkout-ordering and rebase-burden claims in the PR body ("android, gbs_build, pdebuild, static.check" need a rebase to pick up the new action) were verified against the actual checkout configs: those four are exactly the workflows that pin `ref: github.event.pull_request.head.sha`; every other touched workflow uses the default merge-ref checkout and will pick up the action automatically once this merges to main, regardless of the source branch's state. The claim is accurate.

## Verdict

No blockers found in the code itself; the implementation is correct, well-tested, and the regression-detection story is genuinely good. The one item that should be resolved before merging (beyond the PR's own DRAFT/DO NOT MERGE state) is an explicit maintainer decision on the Tier-2 CI pass-rate change (§2), since it changes failure behavior on PR-gating workflows, not just the workflows this issue was filed against. **OK to merge once a maintainer explicitly signs off on the Tier-2 pass-rate change (codeql / ubuntu_clean_meson_build / publish_docs now failing hard on a transient mirror instead of silently tolerating it); no other blocking issues found.**

myungjoo:
Thanks for the review. Responses, item by item:

- **Major, Tier 2 pass-rate change.** This is the resolution #4887 itself proposes ("split **and** add the retry loop in the same change"), and the issue author is the PR author, so the sign-off is this PR. The retry is the mitigation; a mirror that fails three times over 30 s is a real outage, and a green job that silently skipped its installs was never a pass.
- **Minor, nns_testhub.yml.** Agreed and already noted in the PR body. It is a scheduled job, and an honest failure there is preferable to a build against whatever gbs happened to be left on the runner.
- **Minor, the three single-line sites.** Deliberate, as the PR body says: they gain the retry, and the checker can apply one rule with no exceptions.
- **Nit, yocto.yml cleanup chain.** Will add a one-line note. One correction to the review: that chain is not the whole step, `df -h` follows it, so a failure inside it is still masked today. That is a pre-existing property of the disk-space cleanup and not an `apt-get update` site, so this PR leaves its behaviour alone; the note will say so.

The nit will go in with any CI-driven fixes as one push, after which the review runs again.

myungjoo (commented):
This is a follow-up review, transcribed on behalf of the same separate AI review agent (Claude) that left the previous review comment. It re-checks that review's open items against the author's reply, reviews the new commit on its own merits, and independently verifies the CI evidence the author/coordinator cited, rather than taking it on faith.

## 1. Status of the previous review's items

- **Major — Tier 2 pass-rate change.** **Resolved.** The author's point is correct and, on checking `CONTRIBUTING.md:17`, verifiable: MyungJoo Ham is a TSC voting member of this project, and is also the filer of #4887, whose own suggested resolution is "split **and** add the retry loop in the same change." A maintainer accepting a trade-off their own issue proposed, on their own PR, is a legitimate sign-off under this project's governance — my ask was for exactly this kind of explicit decision, and it's now on record in both the issue and the PR body. Downgrading from "blocking" to acknowledged. (One non-blocking observation: this is still a real trade-off, not a non-issue — worth watching post-merge whether `codeql`/`ubuntu_clean_meson_build`/`publish_docs` see a higher red-CI rate from transient mirrors than before. That's monitoring, not a merge blocker.)
- **Minor — `nns_testhub.yml`.** **Acknowledged, no change requested.** Author confirms this was a deliberate, already-disclosed trade-off (scheduled job, self-hosted runner). Agreed this needs no code change.
- **Minor — the three single-line sites converted anyway.** **Acknowledged, no change requested.** Author reaffirms the PR body's rationale (uniform rule, no exceptions, they gain the retry). This was flagged as "make sure it's a conscious decision," and it is. Agreed, no code change needed.
- **Nit — `yocto.yml` cleanup-chain comment.** **Addressed, and the author's correction is right — I was wrong in the original review.** I had written that the chain "is the whole step, so its own exit status is already honest." Checking `.github/workflows/yocto.yml` again: it is not. The `&&` chain (`apt-get remove && apt-get purge && apt-get autoremove && rm -rf ...`) is followed by a final `df -h` (confirmed at the end of the "acquire some disk space" step, after the chain), so a failure partway through the chain is in fact swallowed — the step still exits 0 because `df -h` is what actually runs last. That is the same class of bug #4887 is about, just on `apt-get remove`/`purge`/`autoremove` instead of `apt-get update`, and it is correctly out of scope for this PR/issue, which is specifically about the `apt-get update` pattern. Commit `4ca3f274` adds exactly the one-line comment I asked for (`.github/workflows/yocto.yml`, right before the chain: "Best-effort: a failure inside this chain does not fail the step; df below shows what was freed.") and, now that I've re-checked the file, the comment is accurate, not just plausible. This resolves the nit.
  - Non-blocking forward-looking note, explicitly not a request against this PR: the fact that a real masked-failure exists here might be worth its own follow-up issue at some point (a failed cleanup could silently leave the runner short on disk before the actual yocto build starts), but that is a distinct bug from #4887 and pulling it into this PR would be exactly the scope creep the previous review already flagged as a Minor risk. Leaving it alone here is the right call.

## 2. Commit `4ca3f274` on its own merits

`git show 4ca3f274` — one file, one line added, no code/behavior change:
```
+          # Best-effort: a failure inside this chain does not fail the step; df below shows what was freed.
```
Correctly placed immediately before the chain it documents, inside the same `if`-gated block, so it stays adjacent to the code it explains. Doxygen-style file header was already present and untouched (not needed again for a comment-only change). DCO sign-off present and matches the PR author. No scope creep, no unrelated edits. Confirmed OK.

## 3. Independent verification of the CI evidence

I did not take the coordinator's summary of the `fe93a7ff` run on faith; I pulled it via `gh run list`/`gh run view`:
- Every workflow run recorded against `fe93a7ff` (Yocto, Tizen/GBS, LLVM/Clang build, pdebuild, RISCV64, Static checkers, Spell Check, macOS, Docker Tizen App Build, Android NDK, CodeQL Advanced, c++-source-file-format-checker) shows `conclusion: success`. Only "Minimal meson build in Ubuntu with Valgrind" was `in_progress` at query time — matches the "18 of 19, Valgrind still running" description.
- The `Static checkers and verifiers` job log for that run shows the three new steps actually executing: `/Checker/ apt-get update stands alone` ran `apt-update-chain.sh` and printed `No workflow chains apt-get update with &&.`; `/Checker/ apt-update-chain self-test` and (by the same pattern, not separately excerpted here) `/Checker/ apt-update action self-test` ran without emitting any `::error::`. This is the checker working against the real, already-migrated workflow tree, not just its own fixtures.
- The `Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)` job log for that run shows the composite action executing as designed on a live runner: `shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}` (confirms the explicit-`shell: bash` reasoning from the previous review), `removed '/etc/apt/sources.list.d/microsoft-prod.list'` and `removed '/etc/apt/sources.list.d/azure-cli.sources'`, followed immediately by a successful `apt-get update` (package indices fetched, no retry/backoff messages) — exactly the "removed both sources, updated on the first attempt" behavior the coordinator described. This is first-party evidence the mechanism works in production, not just under the local fakes in `test_apt_update.sh`.

At the time of this review, the new commit `4ca3f274` has its own CI run queued/in-progress (`gh pr checks 4913` shows all checks `pending`); I have not yet seen a completed run for this exact head. Since the commit is comment-only and touches no logic, I don't expect a different outcome than `fe93a7ff`'s, but that is an expectation, not something I've observed for this head.

## Verdict

All four items from the previous review are resolved or acknowledged with no remaining code changes requested; the one factual correction the author raised is correct and the fix for it is accurate. No new issues found in `4ca3f274`. **OK to merge once the in-flight CI run for `4ca3f274` finishes green (no reason to expect otherwise for a comment-only commit) and the DO NOT MERGE label/draft state is lifted by a maintainer.**

myungjoo:
Thanks for the approval. The one suggestion is taken in d9798463: the checker now scans every YAML file under `.github`, so a `run:` step in a composite action is covered as well. The self-test gains a fixture with the chain inside `actions/*/action.yml`, which I confirmed the previous checker accepted, and its final pass now covers the whole shipped `.github` tree. Shell scripts stay out of the scan, since a retry loop legitimately writes `apt-get update && break`.

myungjoo (commented):
This is another follow-up review, transcribed on behalf of the same separate AI review agent (Claude) that left the two earlier review comments. It covers only what's new: the `myungjoo-bot` review and commit `d9798463` that responds to it.

## Context

`myungjoo-bot` posted an APPROVED review at `4ca3f274` with one non-blocking suggestion: `apt-update-chain.sh`'s `find -maxdepth 1` over `.github/workflows` cannot see `.github/actions/*/action.yml`, and since this PR creates `.github/actions` (the repo's first composite action), a `run:` step inside a future composite action could reintroduce the exact masked-`&&`-chain bug this PR exists to remove, with nothing to catch it. Commit `d9798463` addresses it.

## 1. Does `d9798463` fully close the bot's blind spot?

**Confirmed OK — verified, not just read.** The change moves the scan root from `.github/workflows` to `.github` (`SCRIPT_DIR/../..` instead of `SCRIPT_DIR/..`) and drops `-maxdepth 1` from the `find`, so it now recurses through the whole `.github` tree instead of stopping at direct children of one directory. I proved this closes the specific gap the bot found, not just trusted the commit message:

- Extracted the pre-fix checker with `git show 4ca3f274:.github/workflows/static.check.scripts/apt-update-chain.sh` and ran it against a fixture at `actions/setup/action.yml` containing `sudo apt-get update && sudo apt-get install -y gbs` inside a composite `run:` step — **it passed silently** (`No workflow chains apt-get update with &&.`, exit 0), confirming the blind spot was real.
- Ran the current checker (`.github/workflows/static.check.scripts/apt-update-chain.sh` at `d9798463`) against the identical fixture — **it correctly rejects it** (`::error::.../actions/setup/action.yml:5: apt-get update is chained with &&...`, exit 1).

So the fix isn't cosmetic; it changes the actual outcome on the exact case the bot flagged.

## 2. False-positive risk from scanning all of `.github`

**Confirmed OK, checked against the real tree, not just inferred.** Ran `find .github -name '*.yml' -o -name '*.yaml'` (matches the commit's own recursive `find`): 21 workflow files (`.github/workflows/*.yml` + `macos.yaml`) plus exactly the four `action.yml` files the commit message claims (`apt-update`, `check-rebuild`, `gitpush`, `s3_upload`). No other YAML exists under `.github` today — `.github/ISSUE_TEMPLATE` holds only `.md` files, so it isn't touched by the pattern regardless. I read the three action.yml files the PR itself doesn't touch (`check-rebuild`, `gitpush`, `s3_upload`) and none contain `apt-get`/`apt` at all, so none newly trip the checker — the recursive scan doesn't flag anything in the existing tree beyond what the PR itself already fixed. Shell scripts are correctly excluded from the scan (only `*.yml`/`*.yaml` are matched), which is required — `apt-update.sh`'s own retry loop would otherwise self-trip if it ever used `&&` for its retry (it doesn't; it uses `if`), and the commit message correctly notes a retry loop is legitimately allowed to write `apt-get update && break` in shell.

## 3. Does the self-test actually pin the regression?

**Confirmed OK**, per the check in §1: the new fixture "a chain inside a composite action is rejected" (`test_apt_update_chain.sh`, writing to `actions/setup/action.yml`) is a real regression test — it fails against the old checker and passes against the new one, which is exactly what a regression fixture needs to do to be worth anything (a fixture that both versions handle identically wouldn't prove the fix did anything).

## 4. Local test run

`bash .github/workflows/static.check.scripts/test_apt_update_chain.sh` — **18/18 assertions pass**, including the new "a chain inside a composite action is rejected" case and the final whole-tree check ("No workflow or action chains apt-get update with &&." / "PASS: no shipped workflow or action chains apt-get update"). Also re-ran `.github/actions/apt-update/test_apt_update.sh` for completeness — unaffected by this commit, still 14/14. Also cross-checked against live CI: the `Static checks` job for `d9798463` (run `33717041367`) completed `success`, and its log shows the same updated message (`No workflow or action chains apt-get update with &&.`) — the fix is verified working under the real merge-ref-guarded CI path, not just locally.

## Other notes

- Doxygen-style comments (`@brief`, `@param`) and the file-header prose in both scripts were updated consistently with the behavior change (`workflow` → `workflow or action`, `WORKFLOWS` → `ROOT`, parameter doc for `expect_checker`'s `$3` updated from "file name" to "file path... relative to the fresh directory" to match the new nested-path fixture). No stale documentation left behind.
- `static.check.yml`'s step comment was updated to say "every YAML file under `.github`" instead of "the workflow files" — consistent with the actual behavior, not just the step name.
- Scope is minimal and squarely on-topic: 3 files touched, all directly part of the checker infrastructure this PR already introduced; no unrelated edits.
- This closes the loop the bot's review opened without introducing a new one: the recursive scan is a general fix (any future composite action or workflow subdirectory is covered automatically), not a special case for the four files that exist today.

## Verdict

The bot's one suggestion is fully and correctly implemented, verified both by direct before/after testing against the specific fixture that exposed the gap and by a live CI run at this exact head. No new issues found. **OK to merge as is**, pending only the maintainer lifting the DRAFT/DO NOT MERGE state and the remaining in-flight jobs for `d9798463` finishing green.

myungjoo:
Rescoped: the previous head (d9798463) was +540/-57 across 18 files for a bug whose fix is one line per site, which is out of proportion. The branch is now a single commit, f3341e81, +82/-12 across 9 files.

Kept: the composite action (two small files) and its call at the nine buggy sites, including the two the issue did not list. Dropped: the static checker, both self-tests and their `static.check.yml` wiring, and the conversion of the six sites that were not buggy (the three single-line steps the issue asked to leave alone, the `&& \` step in `ubuntu_clean_llvm_build.yml`, and the two plain updates in `android.yml` and the `yocto.yml` build step). The inline retry loops #4877 put in `pdebuild.yml` and `update_pbuilder_cache.yml` are left as they are as well.

So the `.github/actions` blind spot raised in the last review no longer applies, since there is no checker any more. The description is rewritten to match.

myungjoo (commented):
This is a fresh review, transcribed on behalf of the same separate AI review agent (Claude) that left the three earlier review comments on this PR. The branch was rescoped and force-pushed to a single commit, `f3341e81` (+82/-12, 9 files), on top of current `origin/main`, dropping the static checker, both self-tests, and the conversion of six non-buggy sites. Re-reviewing from scratch against the same criteria as the first review.

## 1. Are all nine sites fixed, with correct ordering and `if:` guards?

**Confirmed OK — verified file by file, not just by diff stat.** `grep -rn "apt-get update\|apt update" .github/workflows/*.yml .github/actions/*/*.yml` after the diff shows the pattern surviving only at: the two inline retry loops in `pdebuild.yml`/`update_pbuilder_cache.yml` (`apt-get update && break`, safe — unchanged from #4877, not in scope), the three deliberately-left single-line whole-step chains (`publish_docs.yml:49`, `static.check.yml:24`, `update_gbs_cache.yml:40`), the `&& \` chain in `ubuntu_clean_llvm_build.yml` (also whole-step — I checked the file: nothing follows the `apt-get install ...` continuation within that step, so it is genuinely safe, not just superficially resembling the safe form), and two bare `apt-get update` lines (`android.yml:30`, `yocto.yml:95`) that are on their own line with no `&&` at all — I read both in context and confirmed the following `apt-get install`/`apt install` is a separate statement, not chained. None of these six are bugs, and the PR correctly leaves all six alone, matching the rescoping description exactly.

For the nine sites that are fixed (`codeql.yml` ×2 chains, `gbs_build.yml`, `nns_testhub.yml`, `publish_docs.yml`'s `generate_hotdoc` job, `ubuntu_clean_meson_build.yml` ×2 chains, `docker_build_test.yml`, `yocto.yml`'s cleanup step): checked every `uses: ./.github/actions/apt-update` insertion point against its job's `actions/checkout` — all come after it — and every `if:` guard on the new step against the guard on the install step it now precedes: `env.rebuild == '1'` in `gbs_build.yml`/`ubuntu_clean_meson_build.yml`, `matrix.build-mode == 'manual'` in `codeql.yml`, unconditional in `docker_build_test.yml`/`nns_testhub.yml`/`publish_docs.yml`, all matching their originals. `nns_testhub.yml`'s split (source-prep step → `apt-update` with `drop-microsoft-sources: false` → a new "Install gbs" step) and `docker_build_test.yml`'s identical split were read as full files, not just diff hunks, to confirm the YAML nesting is correct — it is.

The PR body's rebase-burden claim ("gbs_build" needs a rebase, the rest pick the action up via the merge ref) checks out: `gbs_build.yml` is the only one of the seven touched files that pins `ref: github.event.pull_request.head.sha`; the other six use the default checkout.

## 2. Is the trimmed `apt-update.sh` still correct?

**Confirmed OK, verified by execution, not just by reading.** Diffed the current script against `backup/4913-v1:.github/actions/apt-update/apt-update.sh` (the longer version). What's gone: the `APT_SOURCES_DIR`/`APT_UPDATE_ATTEMPTS` env-var knobs (neither was ever set by any caller — dead configurability, safe to drop), the `::group::`/`::endgroup::` log folding, and the per-attempt "apt-get update failed (attempt N of M); retrying" message.

I didn't take "still correct" on faith — I put fake `sudo`/`apt-get`/`sleep` on `PATH` (the same technique the now-deleted `test_apt_update.sh` used) and ran the current script directly:
- 0 failures → exits 0, one `apt-get update` call, no sleep.
- 2 failures then success → exits 0, three `apt-get update` calls, sleeps `10 20` (backoff grows, none after the final successful attempt).
- 3 failures → exits 1, three calls, sleeps `10 20`, prints `::error::apt-get update failed after 3 attempts`.

All three match exactly what the PR body claims and what the deleted self-test used to pin. The `DROP_MICROSOFT_SOURCES` path also runs cleanly with no real `/etc/apt/sources.list.d`.

**Minor.** Dropping the `::group::`/`::endgroup::` markers and the per-attempt retry message is a real (if small) observability regression: in a live run, a triager looking at "did this succeed on the first try or only after retrying" now has to infer it from timing rather than reading it directly in the log, and the apt-get output no longer folds into a collapsible section. Not a correctness issue, and a defensible trade for fewer lines — just noting it's a trade, not a free simplification.

## 3. Is "CI-only change, CI run is the test, retry path exercised locally with fakes" an acceptable reason under CLAUDE.md?

**Major — the framing is not quite accurate, and the gap it elides is real.** CLAUDE.md's carve-out is for changes that are "genuinely untestable... state the reason if so." Manual GTest/SSAT genuinely don't apply here — that part is correctly reasoned. But "genuinely untestable" oversells it: this exact change *was* tested with an automated, repeatable, CI-wired self-test in the immediately preceding version of this same PR (`test_apt_update.sh`, `test_apt_update_chain.sh`) — both ran green in real CI at `d9798463`, and I independently verified their coverage and correctness across the last three review rounds. The rescope's PR body says the retry path was "exercised locally," past tense, informally — but that self-test already existed as a committed, reusable artifact one force-push ago, and demonstrated that automated testing of exactly this kind of change is not merely possible but was already done well. What was dropped for diff-size reasons is being described as if it were infeasible, and those are different things worth naming differently.

The concrete cost, tying into point 5: with the checker and self-tests gone, **there is no longer any automated mechanism that will fail CI if a future PR reintroduces `apt-get update && apt-get install` anywhere in `.github`** — not just at these nine sites, but repo-wide, including inside a composite action (the exact blind spot `myungjoo-bot`'s review caught last round, closed in `d9798463`, and now gone again along with the whole checker). Both my own and the bot's previous reviews independently called that regression-detection mechanism the strongest part of the prior version. Losing it is a real, not cosmetic, downgrade in what this PR leaves behind — it fixes the nine known instances but removes the guarantee that a tenth won't quietly reappear.

I'd suggest either: (a) reinstate a minimal version of just `apt-update-chain.sh` plus one self-test as a small, separate follow-up PR (it doesn't need to be as large as the dropped version — even the checker alone, without the elaborate two-tier self-test suite, would restore most of the value), or (b) if the maintainer's judgment is that diff-size proportionality wins here, say so explicitly in the PR body rather than folding it under "untestable" — since a reviewer or future archaeologist reading "there is nothing here for GTest or SSAT to exercise" could reasonably conclude no better option existed, when one did and had already been built.

This is the one item in this review I'd call a genuine, not cosmetic, regression from where the PR stood two rounds ago — but it doesn't block the *current* diff from being correct on its own terms, since what it does fix, it fixes correctly (per §1 and §4) and the maintainer (a TSC member, per `CONTRIBUTING.md:17`, and the filer of #4887) has the standing to make this trade-off deliberately.

## 4. The `yocto.yml` cleanup edit — dropping `apt-get update --fix-missing &&` only

**Confirmed OK, and sufficient to close the #4887-shaped bug at this site.** Before: `apt-get update --fix-missing && apt-get remove ... && apt-get purge ... && apt-get autoremove ... && rm -rf ...`, followed later by `df -h`. A failed `update` used to short-circuit the entire cleanup (remove/purge/autoremove/rm never ran) while the step still reported success, because `df -h` is what actually ran last — exactly the masking pattern #4887 is about. Removing `apt-get update --fix-missing` from the front removes that specific risk outright, and the reasoning holds: `apt-get remove`/`purge`/`autoremove` operate on the installed-package database, not a package index, so they don't need a preceding `update` to work correctly, and the subsequent `build` step (unaffected by this PR, per §1) still runs its own bare `apt-get update` before installing anything.

**Nit.** The residual `remove && purge && autoremove && rm -rf` chain is still `&&`-joined and still followed by `df -h`, so a failure partway through it is still silently swallowed — this is the same fact I raised as a nit two rounds ago, which the author then correctly fixed me on and documented in place with a one-line comment (`4ca3f274`, since discarded along with the rest of that branch's history). That comment does not appear in the current single-commit version. This is out of scope for #4887 either way (it's an `apt-get remove` chain, not an `apt-get update` chain), so I'm not asking for it back — just noting, for the record, that a previously-addressed piece of documentation didn't survive the rescope. Explicitly a nice-to-have, not a request.

## 5. Anything else lost that's a real problem rather than a nice-to-have?

Beyond §3/§4 (already covered): checked the six explicitly-dropped conversions (three single-line sites, the `ubuntu_clean_llvm_build.yml` `&&` step, `android.yml`, `yocto.yml`'s build step) one more time specifically for whether any of them was *actually* buggy and got miscategorized as safe in the rescope — none were. All six are either single-line whole-step chains or bare unchained `apt-get update` lines; I read each in its current file context, not just the issue's table, and the "already honest" classification holds for all of them. No site went from fixed to unfixed.

## CI evidence at this head

`f3341e81`'s CI is still partly in flight at review time, but what has completed is a useful live signal: `Static checks` passed (expected — `static.check.yml` isn't touched, since the checker wiring never existed on `main` to begin with, so there's no dangling reference to worry about), `DCO` passed, and `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04-arm)` — one of the two `ubuntu_clean_meson_build.yml` matrix legs, i.e. one of the nine fixed sites — completed successfully, which is a real integration signal that the `apt-update` action wires and runs correctly in that job, not just in isolation.

## Verdict

What this PR fixes, it fixes correctly and completely, with correct ordering and guards everywhere, a trimmed action script that still behaves exactly as specified (verified by direct execution), and a defensible, correctly-reasoned yocto edit. The one substantive concern is §3/§5: the rescope also removed the only mechanism that would have caught a future reintroduction of this exact bug anywhere in `.github`, and the PR body's testing rationale reads as "untestable" when the more accurate description is "tested, and then removed for size." That's a process/scope call within the maintainer's standing to make, not a defect in the code that remains. **OK to merge as is; consider filing a small follow-up to restore just the static checker (with or without its full self-test suite) so this class of bug can't come back unnoticed.**

myungjoo:
On the fourth review:

- **Major, "untestable" wording.** Fair. The Tests section now says what actually happened: a checker and two self-tests existed one head ago, worked, and were dropped as disproportionate, not as impossible. It also states plainly that nothing in CI would flag a future `update && install` chain, and names a small checker as a possible follow-up.
- **Minor, `::group::` markers and the per-attempt log line.** Left out on purpose; the action's step is short enough to read unfolded, and the retry is visible from the repeated `apt-get update` output. Easy to add back if it proves annoying in practice.
- **Nit, the yocto cleanup comment.** Dropped with the rescope, deliberately: that chain's masking is a separate defect from #4887 and is better tracked in its own issue than by a comment in this one.
- **Follow-up checker.** Noted for the maintainer to decide; not part of this PR.

No code changed since f3341e81; only the PR description.

myungjoo:
Rescoped again, to the minimum. Head dae4aa8b breaks each `update && install` chain into one command per line at the nine sites and nothing else: no action, no retry, no source removal, no new files. The yocto cleanup drops its update instead of splitting it, for the reason given in the description. The description is rewritten to match.

myungjoo (commented):
This is a fresh review, transcribed on behalf of the same separate AI review agent (Claude) that left the four earlier review comments on this PR. The branch was rescoped to the minimum and force-pushed once more: single commit `dae4aa8b` (+19/-10, 7 workflow files, no new files), on top of current `origin/main`. No composite action, no retry, no source removal — just breaking each chain into one command per line. Re-reviewing from scratch against the same criteria as the first review.

## 1. Is every one of the nine sites now honest under `bash -e`?

**Confirmed OK, verified against the live file, not the diff hunks.** `grep -rn "apt-get update\|apt update" .github/workflows/*.yml` after the change shows all nine sites (`codeql.yml:76,80`, `docker_build_test.yml:46`, `gbs_build.yml:42`, `nns_testhub.yml:24`, `publish_docs.yml:20`, `ubuntu_clean_meson_build.yml:31,35`, plus the yocto cleanup handled by removal) now have `apt-get update`/`apt update` on a line by itself, with the corresponding `install` on the next line and nothing joining them. I read each of the seven files in full, not just the hunks, to confirm dependency ordering survived the split correctly: the Tizen-source `tee` still precedes `apt-get update` in `docker_build_test.yml`/`gbs_build.yml`/`nns_testhub.yml`, and `add-apt-repository` still precedes the `apt-get update` that picks it up in `codeql.yml`/`ubuntu_clean_meson_build.yml`.

Checked the shell-flavor point specifically: only `codeql.yml` sets `shell: bash` explicitly (`bash -eo pipefail`); the other six rely on the implicit default (`bash -e`, no pipefail, per the issue's own confirmed job-log finding). This distinction doesn't matter for this fix — putting a command on its own line makes it a standalone list item under errexit either way, and `-o pipefail` only changes how a *piped* command's exit status is computed, and none of the touched lines pipe anything. Both shell flavors correctly abort on a failed `apt-get update` after this change.

Also re-verified the six sites left untouched are genuinely safe, not just superficially similar: `publish_docs.yml:49`, `static.check.yml:24`, `update_gbs_cache.yml:40` are single-line whole-step chains; `ubuntu_clean_llvm_build.yml:30`'s `&& \` chain has nothing after it within its step; `android.yml:30` and `yocto.yml:95` are bare, unchained lines already followed by a separate `install` statement. `pdebuild.yml`/`update_pbuilder_cache.yml` keep the `&& break` retry idiom from #4877, which is a different, already-safe pattern (the loop's own subsequent `if [ "$i" = 3 ]; then exit 1; fi` makes the eventual failure explicit). None of the six were miscategorized.

## 2. The yocto decision: drop vs. split

**Confirmed OK — dropping is the right call, not just an equally-valid alternative.** The removed line was `apt-get update --fix-missing`, gating a `remove && purge && autoremove && rm -rf` chain that operates entirely on the installed-package database, not a package index — so the update never did useful work for what follows it. Had it been *split* instead (kept as its own statement expected to abort the step on failure), a single transient mirror hiccup would newly hard-fail the "acquire some disk space" step outright — turning a best-effort, disk-triggered cleanup into a new blocking failure mode with no corresponding benefit, since the `build` step immediately after already runs its own honest, unchained `apt-get update` before installing anything (`yocto.yml:95`, confirmed untouched and already safe). Dropping is strictly better here than splitting.

## 3. The test-policy statement

**Resolved.** This directly answers the Major finding from my previous review, where the prior head's "genuinely untestable" framing glossed over the fact that a working, CI-verified self-test suite had existed one head ago and was removed for size, not infeasibility. The current description says exactly that: "Earlier heads of this PR carried a composite action with retries and a static checker with self-tests; they were dropped as out of proportion to a one-line-per-site fix." That's accurate and no longer misleading — the reasoning is now correctly attributed to a proportionality judgment rather than technical necessity. No further concern here.

## 4. What's lost relative to previous heads — nice-to-have vs. real problem

Three things are gone relative to the immediately preceding head (`f3341e81`): the composite action, the retry loop, and the `packages.microsoft.com` source removal. I'd split these differently:

- **The static checker / self-tests and the source removal: nice-to-have, already an acknowledged trade-off.** The source removal was never present on `main` for these nine sites to begin with (only `pdebuild`/`update_pbuilder_cache` had it, from #4877) — going without it here isn't a regression against `main`, just forgoing a hardening bonus that briefly existed on an intermediate head. The checker's absence was already litigated in my previous review and is now honestly disclosed (§3); I won't re-raise it beyond noting the follow-up suggestion still stands if a maintainer wants it.

- **Dropping the retry entirely: Major, and I think this is the one real problem in this round.** Every prior head of this PR — including the very first one — kept a 3-attempt retry with backoff specifically as the mitigation for the honesty trade-off on the "Tier 2" sites (`codeql`, `ubuntu_clean_meson_build`, `publish_docs`), which the issue itself identified as jobs that currently survive a transient mirror failure by accident. The PR's own "Things to know" section still makes that trade-off explicitly ("they now fail honestly, the same way `pdebuild` has since #4877") — but `pdebuild` fails honestly *after three attempts over roughly 30 seconds*, not on the first hiccup. This head gives these nine sites zero retry cushion: a single transient `apt-get update` failure now fails the job outright, on `codeql` and `ubuntu_clean_meson_build`, both of which gate every PR in this repository. That's a materially higher flake rate than what was reviewed and accepted in every previous round, and the "same as `pdebuild`" comparison in the PR body somewhat undersells the difference since `pdebuild` doesn't share this head's zero-retry profile. I don't think this blocks the fix itself — the masking bug is correctly and completely closed, and a manual CI re-run is a low-cost recourse — but it's worth the maintainer weighing consciously rather than by analogy to a retry-bearing precedent that this head no longer resembles.

## 5. `add-apt-repository -y` (no `-n`) followed by an explicit `apt-get update`

**Confirmed OK, and correctly scoped as out of this PR.** In `codeql.yml` and `ubuntu_clean_meson_build.yml`, `add-apt-repository -y ppa:nnstreamer/ppa` (without `-n`) triggers its own implicit `apt-get update`, and the next line runs an explicit `sudo apt-get update` again — two updates where one would do. I confirmed this is exactly what those two lines already do on unmodified `main` today (the original chain was `add-apt-repository -y ppa:... && apt-get update && apt-get install ...`, which has the same implicit-plus-explicit double update); the split here preserves that behavior byte-for-byte, it doesn't introduce it. Earlier heads added `-n` as a bonus deduplication; leaving it out here is consistent with "nothing beyond the literal fix," costs one redundant network round-trip per run, and is not a correctness problem — no site is installed against a stale index, no update is skipped, nothing is masked.

## CI evidence at this head

At review time, `dae4aa8b`'s CI is mostly still in flight, but `DCO`, `Static checks`, `CPP-Linter`, and `Spell Check` have completed successfully. `Static checks` passing is expected and uninformative either way here (nothing in `static.check.yml` references anything this PR touches). The nine-site jobs themselves (Tizen GBS build, meson+Valgrind, Yocto, pdebuild, CodeQL, etc.) were still queued/running at review time.

## Verdict

The fix is complete, minimal, and correct for exactly what it claims to do: every one of the nine sites is now honest under errexit, the yocto call was the right one, and the testing rationale no longer overstates its own limits. The one substantive new concern is the retry removal (§4), which changes the practical flake-rate calculus that every earlier round of this review accepted under a "retry mitigates the honesty trade-off" framing that no longer applies at this head — worth a conscious yes from the maintainer rather than an implicit one via the `pdebuild` comparison, but not something the diff itself gets wrong. **OK to merge as is; flagging the zero-retry exposure on `codeql`/`ubuntu_clean_meson_build`/`publish_docs` for the maintainer's explicit awareness, not as a blocker.**

myungjoo:
On the fifth review's one finding: agreed that the pdebuild analogy was off, since pdebuild retries and this does not. The description now says so directly: no retry, a mirror hiccup fails the job on the first attempt, and a red from that is re-run by hand. That is the maintainer's deliberate choice of the smallest fix; it can be revisited if such reds prove frequent. No code changed since dae4aa8b.

Issue #2, opened 2018-07-27
[External CI] Connect to launchpad.net/PPA and build.opensuse.org


myungjoo:
Not for per-PR at this moment, but for CD basis.

myungjoo:
NNStreamer PPA is now up and running. Closing this.
CC: @again4you @ohsewon

PR #4914, opened 2026-09-02
[CI] Restore the per-function Doxygen check and clear its backlog

### LINKED ISSUE #4908: [CI] Per-function Doxygen check has been silently disabled by a stray 'local'
## Summary

`static.check.scripts/doxygen-tag.sh` declares `local` at script top level. Bash rejects it, the script has no `set -e`, so it continues with the variable empty and the **per-function `@brief` check never runs**. It has been passing vacuously, not passing.

`CLAUDE.md` states that every function needs a Doxygen comment and that CI enforces it. The first half is policy; the second half is currently untrue.

## Mechanism

[`doxygen-tag.sh:75`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/doxygen-tag.sh#L75):

```bash
local function_check_flag="f+p"   # no enclosing function in this file
```

1. No function definition precedes line 75, so this is top level. `bash -c 'local x=1'` → `local: can only be used in a function`.
2. The script has **no `set -e`** (0 occurrences), so execution continues with `function_check_flag` unset.
3. [`:85`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/doxygen-tag.sh#L85) then runs `ctags -x --c-kinds= "$file"`.
4. `function_positions` stays empty, so the branch at [`:100`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/doxygen-tag.sh#L100) can never fire.

This is on the live path: [`static.check.yml:85`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.yml#L85) invokes the script with `1` as the second argument, so `advanced=1` and the block does execute on every PR.

Confirmed against `exuberant-ctags`, which is what CI installs ([`static.check.yml:24`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.yml#L24)):

| invocation | tags returned |
|---|---|
| `--c-kinds=` (what actually runs) | **0** |
| `--c-kinds=f+p` (what was intended) | 50 |
| `--c-kinds=sc` (struct/class check) | 2 |

## Scope: what is dead and what still works

Only one of the three checks is affected, which is why the checker still looks alive and green:

| check | status |
|---|---|
| File-level `@file` / `@brief` / `@author` / `@bug` | **works** |
| struct/class `@brief` ([`:90`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/doxygen-tag.sh#L90)) | **works** — `--c-kinds=sc` is hardcoded and does not use the broken variable |
| **function / prototype `@brief`** ([`:100`](https://github.com/nnstreamer/nnstreamer/blob/main/.github/workflows/static.check.scripts/doxygen-tag.sh#L100)) | **dead** |

## Why this is not a one-line fix

Changing `local function_check_flag=` to a plain assignment turns the check back on, and it immediately fails on a large existing backlog. Measured over the 212 `.c` / `.cc` / `.h` / `.hh` files under `gst/` and `ext/`, with `exuberant-ctags` and the same `advanced=1` invocation CI uses:

| checker | result |
|---|---|
| current | 0 errors, exit 0 |
| with the one-line fix | **1,144 errors across 117 files**, exit 1 |

Every one of the 1,144 is `function needs @brief tag`. `tests/` and the rest of the tree were not included in that count, so the true backlog is larger.

So the fix needs a plan: land the backlog cleanup first (or in batches), and flip the flag last. Flipping it alone would red-wall every PR.

## Suggested path

1. Fix the `local` (one line) on a branch, and use it to generate the full per-file backlog list, including `tests/`.
2. Clear the backlog in batches, most likely per directory.
3. Land the flag flip last, once the tree is clean.

## Note for anyone reproducing locally

Behaviour differs by ctags implementation, so use `exuberant-ctags` to match CI. With `universal-ctags` (what most current distros install as `ctags`) the empty `--c-kinds=` does **not** return nothing — it returns namespaces and classes, because those are C++-only kinds and are unaffected by emptying the *C* kind set. The function check then misfires rather than going quiet, reporting `namespace ... function needs @brief tag`, and `--c-kinds=sc` additionally warns `Unsupported kind: 'c'`. Different symptom, same root cause.

## How this surfaced

Found during review of #4905, where a destructor missing its `@brief` reached the PR and was caught by a reviewer rather than by CI.

myungjoo:
Fixes #4908.

## What was wrong

`.github/workflows/static.check.scripts/doxygen-tag.sh` declared `local function_check_flag="f+p"` at script top level. Bash rejects `local` outside a function, the script has no `set -e`, so `ctags -x --c-kinds=` ran with an empty kind list and the per-function `@brief` check never fired. It has been that way since the Actions import (487d7e25, January 2024); in TAOS-CI the same block lived inside `function pr-prebuild-doxygen-tag()`.

Two smaller defects in the same script hid behind it: `$report_path` was never set, so every file produced an "ambiguous redirect" on stderr, and `$brief` was not reset between files.

## What this PR does

**Checker** (`doxygen-tag.sh`)
- Drop the `local`, so the ctags kind list is populated again.
- Check function *definitions* everywhere, but *prototypes* only in headers. A `static` forward declaration in a `.c`/`.cc` is documented at its definition; demanding a second `@brief` on each of the 1,119 forward declarations in the tree was never the intent.
- Recognise the trailing `/**< ... */` form, on the declaration line, on the line after a declaration, or after an inline body. The public C++ subplugin API header documents its virtual methods that way, and Doxygen accepts it.
- Default `report_path` to `/dev/null` and reset `brief` per file. The pending tag now survives only inside a `/* ... */` block (the scanner tracks that state) or on a `//` line; before, any line containing `*` kept it alive, so a pointer parameter carried one function's `@brief` onto the next prototype.

**Self-test** (`test_doxygen_tag.sh`, wired into `static.check.yml` next to the other self-tests). Eighteen fixtures generated into a temp dir: the negative control (undocumented definition must fail), prototype scope in `.c` vs `.h`, all three `/**<` placements plus the case where the next line's `/**<` belongs to another member, the pointer-parameter reset (mid-line and at the start of a wrapped line) and its converse (a `;` inside a comment block keeps the tag), struct check, per-file state reset, file-level tags, and "no stderr output". Against the checker on `main` all eighteen fail: the negative controls pass vacuously and every case emits stderr noise. Against this branch all pass.

**Backlog.** With the corrected scope the tree had 248 undocumented functions and 2 undocumented structs in 37 files. All are documented in this PR, grouped by directory in separate commits so they can be reviewed independently. CI runs the checker on changed files only, so the PR's own `Static checks` job exercises the new checker over every file touched here.

## Measurements (Exuberant Ctags 5.9, the CI version, 289 C/C++ files)

| function_check_flag | errors | files |
|---|---|---|
| `f+p` everywhere (one-line fix only) | 1,262 | 132 |
| definitions everywhere, prototypes in headers only | 249 | 35 |
| same, with `/**<` recognised | 237 | 35 |
| same, plus the pointer-parameter reset | 248 | 37 |

`tests/nnstreamer_filter_mxnet/Predictor.hh` also gets the layout clang-format 18 (the runner's current version) wants for four log lines the PR did not otherwise touch; the file had not been edited since the runner moved off 16, so the first PR to touch it trips the check.

## Not in this PR

`nnstreamer/api` and `nnstreamer/nnstreamer-edge` carry the same script with the same `local` and invoke it with `advanced=1`. They need the same checker change and their own backlog pass; that is a separate PR per repository.

### DIFF
--- a/.github/workflows/static.check.scripts/doxygen-tag.sh
+++ b/.github/workflows/static.check.scripts/doxygen-tag.sh
@@ -41,8 +41,9 @@ fi
 # @param $1 1-based line number where ctags placed the function
 # @return 0 when such a comment is found, 1 otherwise
 # Reads the global array file_lines (0-based). A declaration ends at the first
-# ';' outside braces, or at the '}' that closes an inline body; the comment may
-# follow on the next line.
+# ';' outside braces, or at the '}' that closes an inline body. The comment may
+# also open the next line; a "/**<" later on that line belongs to whatever is
+# declared there, not to this one.
 has_trailing_doc() {
   local -i i=$1-1
   local -i last=${#file_lines[@]}-1
@@ -55,7 +56,7 @@ has_trailing_doc() {
     closes=${l//[^\}]/}
     depth+=${#opens}-${#closes}
     if [[ $depth -le 0 && ( $l == *";"* || $l == *"}"* ) ]]; then
-      [[ $i -lt $last && ${file_lines[$i+1]} == *"/**<"* ]] && return 0
+      [[ $i -lt $last && ${file_lines[$i+1]} =~ ^[[:space:]]*/\*\*\< ]] && return 0
       return 1
     fi
     i+=1
@@ -147,8 +148,10 @@ for file in `cat $files`; do
                       # Find brief or copydoc tag in the comments between the codes.
                       if [[ $line =~  "@brief" || $line =~ "@copydoc" ]]; then
                           brief=1
-                      # Doxygen tags become zero in code section.
-                      elif [[ $line != *"*"*  && ( $line =~ ";" || $line =~ "}" || $line =~ "#") ]]; then
+                      # Doxygen tags become zero in code section. A comment line
+                      # ('*', '/*', '//') keeps the pending tag; a code line that
+                      # merely contains '*' (a pointer) does not.
+                      elif [[ ! $line =~ ^[[:space:]]*(\*|/\*|//) && ( $line =~ ";" || $line =~ "}" || $line =~ "#") ]]; then
                           brief=0
                       fi
 
--- a/.github/workflows/static.check.scripts/doxygen-tag.sh
+++ b/.github/workflows/static.check.scripts/doxygen-tag.sh
@@ -100,6 +100,7 @@ for file in `cat $files`; do
               if [[ $advanced == 1 ]]; then
                   declare -i idx=0
                   brief=0
+                  in_comment=0
                   function_positions="" # Line number of functions.
                   structure_positions="" # Line number of structure.
                   mapfile -t file_lines < "$file"
@@ -148,11 +149,19 @@ for file in `cat $files`; do
                       # Find brief or copydoc tag in the comments between the codes.
                       if [[ $line =~  "@brief" || $line =~ "@copydoc" ]]; then
                           brief=1
-                      # Doxygen tags become zero in code section. A comment line
-                      # ('*', '/*', '//') keeps the pending tag; a code line that
-                      # merely contains '*' (a pointer) does not.
-                      elif [[ ! $line =~ ^[[:space:]]*(\*|/\*|//) && ( $line =~ ";" || $line =~ "}" || $line =~ "#") ]]; then
+                      # Doxygen tags become zero in code section. Lines inside a
+                      # block comment, or opening one, keep the pending tag; a
+                      # code line that merely contains '*' (a pointer) does not.
+                      elif [[ $in_comment -eq 0 && ! $line =~ ^[[:space:]]*(/\*|//) && ( $line =~ ";" || $line =~ "}" || $line =~ "#") ]]; then
                           brief=0
+                      fi
+
+                      # Track block comments so that their continuation lines are
+                      # never mistaken for code, whatever character they start with.
+                      if [[ $in_comment -eq 1 ]]; then
+                          [[ $line == *"*/"* ]] && in_comment=0
+                      elif [[ $line == *"/*"* && ${line##*/\*} != *"*/"* ]]; then
+                          in_comment=1
                       fi
 
                       # Check a comment statement that begins with '/*'.
--- a/.github/workflows/static.check.scripts/doxygen-tag.sh
+++ b/.github/workflows/static.check.scripts/doxygen-tag.sh
@@ -28,11 +28,40 @@ fi
 
 files=$1
 failed=0
+report_path=${report_path:-/dev/null}
 
 if [ ! -f $files ]; then
   echo "::error The file $files does not exists."
   exit 1
 fi
+
+##
+# @brief Tell whether the declaration starting at a line carries a trailing
+#        "/**<" Doxygen comment, on the declaration itself or right after it.
+# @param $1 1-based line number where ctags placed the function
+# @return 0 when such a comment is found, 1 otherwise
+# Reads the global array file_lines (0-based). A declaration ends at the first
+# ';' outside braces, or at the '}' that closes an inline body; the comment may
+# follow on the next line.
+has_trailing_doc() {
+  local -i i=$1-1
+  local -i last=${#file_lines[@]}-1
+  local -i depth=0
+  local l opens closes
+  while [[ $i -le $last ]]; do
+    l=${file_lines[$i]}
+    [[ $l == *"/**<"* ]] && return 0
+    opens=${l//[^\{]/}
+    closes=${l//[^\}]/}
+    depth+=${#opens}-${#closes}
+    if [[ $depth -le 0 && ( $l == *";"* || $l == *"}"* ) ]]; then
+      [[ $i -lt $last && ${file_lines[$i+1]} == *"/**<"* ]] && return 0
+      return 1
+    fi
+    i+=1
+  done
+  return 1
+}
 
 echo "::group::Doxygen tag check started"
 
@@ -69,14 +98,19 @@ for file in `cat $files`; do
               # Checking tags for each function
               if [[ $advanced == 1 ]]; then
                   declare -i idx=0
+                  brief=0
                   function_positions="" # Line number of functions.
                   structure_positions="" # Line number of structure.
+                  mapfile -t file_lines < "$file"
 
-                  local function_check_flag="f+p" # check document for function and prototype of the function
-
-                  if [[ $pr_doxygen_check_skip_function_definition == 1 && $file != *.h ]]; then
-                      function_check_flag="p" # check document for only prototypes of the function for non-header file
-                  fi
+                  # Definitions are checked everywhere. Prototypes are checked
+                  # only in headers, where the declaration is the documented
+                  # form; a forward declaration in a .c/.cc is documented at
+                  # its definition.
+                  case $file in
+                      *.h|*.hh|*.hpp ) function_check_flag="f+p" ;;
+                      * ) function_check_flag="f" ;;
+                  esac
 
                   # Find line number of functions using ctags, and append them.
                   while IFS='' read -r line || [[ -n "$line" ]]; do
@@ -97,7 +131,7 @@ for file in `cat $files`; do
                       # Check if a function has @brief tag or not.
                       # To pass correct line number not sub number, keep space " $idx ".
                       # ex) want to pass 143 not 14, 43, 1, 3, 4
-                      if [[ $function_positions =~ " $idx " && $brief -eq 0 ]]; then
+                      if [[ $function_positions =~ " $idx " && $brief -eq 0 ]] && ! has_trailing_doc $idx; then
                           echo "[ERROR] File name: $file, $idx line, `echo $line | cut -d ' ' -f1` function needs @brief tag "
                           failed=1
                       fi
--- a/.github/workflows/static.check.scripts/test_doxygen_tag.sh
+++ b/.github/workflows/static.check.scripts/test_doxygen_tag.sh
@@ -168,11 +168,31 @@ class adder
 };'
 run_checker 1 "declaration without any doc between documented ones fails" trailing_missing.hh
 
+write_fixture trailing_belongs_to_next.hh '/** @brief adder */
+class adder
+{
+  public:
+  /** @brief the class */
+  adder ();
+  int add_one (int x);
+  int count; /**< documents count, not add_one */
+};'
+run_checker 1 "a /**< on the next declaration does not cover this one" trailing_belongs_to_next.hh
+
 write_fixture undocumented_struct.h 'struct point
 {
   int x;
 };'
 run_checker 1 "undocumented struct fails" undocumented_struct.h
+
+# The tag must not survive a code line just because that line has a '*' in
+# it: a pointer parameter is not a comment.
+write_fixture pointer_line.h '/**
+ * @brief add one in place
+ */
+void add_one (int *x);
+void add_two (int *x);'
+run_checker 1 "a brief does not carry over a pointer-typed declaration" pointer_line.h
 
 write_fixture ends_in_brief.c '/**
  * @brief add one
--- a/.github/workflows/static.check.scripts/test_doxygen_tag.sh
+++ b/.github/workflows/static.check.scripts/test_doxygen_tag.sh
@@ -194,6 +194,24 @@ void add_one (int *x);
 void add_two (int *x);'
 run_checker 1 "a brief does not carry over a pointer-typed declaration" pointer_line.h
 
+# A wrapped parameter list can put the '*' at the start of a line; that is
+# still code, not a comment continuation.
+write_fixture wrapped_pointer.h '/**
+ * @brief add one in place
+ */
+void add_one (int
+    *x);
+void add_two (int *x);'
+run_checker 1 "a brief does not carry over a wrapped pointer parameter" wrapped_pointer.h
+
+# The converse: a ';' inside the comment block must not clear the tag.
+write_fixture semicolon_in_comment.h '/**
+ * @brief add one in place; the caller owns x
+ * usage: add_one (&x);
+ */
+void add_one (int *x);'
+run_checker 0 "a semicolon inside the comment block keeps the brief" semicolon_in_comment.h
+
 write_fixture ends_in_brief.c '/**
  * @brief add one
  */
--- a/.github/workflows/static.check.scripts/test_doxygen_tag.sh
+++ b/.github/workflows/static.check.scripts/test_doxygen_tag.sh
@@ -1,0 +1,216 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# @file     test_doxygen_tag.sh
+# @brief    Self-test for doxygen-tag.sh in the advanced (per-function) mode.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Runs doxygen-tag.sh the way static.check.yml does, with a changed-file list
+# and the advanced flag. The per-function check had passed vacuously for two
+# years because a stray "local" emptied the ctags kind list, so the first
+# property pinned here is the negative control: an undocumented function
+# definition must fail. The rest pin the scope of the check (definitions
+# everywhere, prototypes only in headers), the trailing "/**<" form, and the
+# per-file reset of the tag state.
+#
+# Fixtures are generated into a temporary directory rather than committed, so
+# that the intentionally undocumented ones never appear in a changed-file list.
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+REPO_ROOT=$(cd "${SCRIPT_DIR}/../../.." && pwd)
+CHECKER="${SCRIPT_DIR}/doxygen-tag.sh"
+workdir=$(mktemp -d)
+failed=0
+
+trap 'rm -rf "$workdir"' EXIT
+
+if ! command -v ctags > /dev/null 2>&1; then
+  echo "::error ctags is required by doxygen-tag.sh but is not installed."
+  exit 1
+fi
+
+# File-level tags every C/C++ fixture needs so that only the per-function
+# rules decide the outcome. The include line matters: the checker treats a
+# @brief as pending until the next code line, and a real file always has one
+# between the header comment and the first function.
+header_block() {
+  printf '/**\n * @file %s\n * @brief fixture\n * @author fixture\n * @bug none\n */\n#include <stddef.h>\n' "$1"
+}
+
+# write_fixture <name> <content>
+# Creates workdir/<name> with the standard file header followed by content.
+write_fixture() {
+  local name=$1 content=$2
+  { header_block "$name"; printf '%s\n' "$content"; } > "${workdir}/${name}"
+}
+
+# run_checker <expected 0|1> <description> <fixture>...
+# Runs the checker on the listed fixtures in advanced mode, from the repo
+# root, and compares the exit status. Also fails on any stderr output, which
+# is how the broken "local" and the unset report_path used to show up.
+run_checker() {
+  local expected=$1 desc=$2
+  shift 2
+  local list errfile actual f
+
+  list=$(mktemp -p "$workdir")
+  errfile=$(mktemp -p "$workdir")
+  for f in "$@"; do
+    printf '%s\n' "${workdir}/${f}" >> "$list"
+  done
+
+  (cd "$REPO_ROOT" && bash "$CHECKER" "$list" 1) > /dev/null 2> "$errfile" < /dev/null
+  actual=$?
+  [[ $actual -ne 0 ]] && actual=1
+
+  if [[ "$actual" != "$expected" ]]; then
+    echo "FAIL: ${desc} expected exit ${expected}, got ${actual}"
+    failed=1
+  elif [[ -s "$errfile" ]]; then
+    echo "FAIL: ${desc} wrote to stderr:"
+    sed 's/^/    /' "$errfile"
+    failed=1
+  else
+    echo "PASS: ${desc} (exit ${actual})"
+  fi
+}
+
+write_fixture documented.c '/**
+ * @brief add one
+ */
+int
+add_one (int x)
+{
+  return x + 1;
+}'
+run_checker 0 "documented definition passes" documented.c
+
+write_fixture undocumented.c 'int
+add_one (int x)
+{
+  return x + 1;
+}'
+run_checker 1 "undocumented definition fails" undocumented.c
+
+write_fixture forward_decl.c 'static int add_one (int x);
+
+/**
+ * @brief add one
+ */
+static int
+add_one (int x)
+{
+  return x + 1;
+}'
+run_checker 0 "undocumented forward declaration in a .c passes" forward_decl.c
+
+write_fixture undocumented_proto.h 'int add_one (int x);'
+run_checker 1 "undocumented prototype in a .h fails" undocumented_proto.h
+
+write_fixture documented_proto.h '/**
+ * @brief add one
+ */
+int add_one (int x);'
+run_checker 0 "documented prototype in a .h passes" documented_proto.h
+
+write_fixture trailing_same_line.hh '/** @brief adder */
+class adder
+{
+  public:
+  /** @brief the class */
+  adder ();
+  int add_one (int x); /**< add one */
+};'
+run_checker 0 "trailing /**< on the declaration line passes" trailing_same_line.hh
+
+write_fixture trailing_next_line.hh '/** @brief adder */
+class adder
+{
+  public:
+  /** @brief the class */
+  adder ();
+  virtual int add_one (int x) = 0;
+  /**< add one */
+  virtual int add_two (int x,
+      int y) = 0;
+  /**< add two, declared over two lines */
+};'
+run_checker 0 "trailing /**< on the following line passes" trailing_next_line.hh
+
+write_fixture trailing_after_body.hh '/** @brief adder */
+class adder
+{
+  public:
+  /** @brief the class */
+  adder ();
+  virtual int add_one (int x)
+  {
+    return x + 1;
+  }
+  /**< add one, documented after the inline body */
+};'
+run_checker 0 "trailing /**< after an inline body passes" trailing_after_body.hh
+
+write_fixture trailing_missing.hh '/** @brief adder */
+class adder
+{
+  public:
+  /** @brief the class */
+  adder ();
+  int add_one (int x);
+  int add_two (int x);
+  /**< documents add_two only */
+};'
+run_checker 1 "declaration without any doc between documented ones fails" trailing_missing.hh
+
+write_fixture undocumented_struct.h 'struct point
+{
+  int x;
+};'
+run_checker 1 "undocumented struct fails" undocumented_struct.h
+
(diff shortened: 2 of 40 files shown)


myungjoo:

Scope: read #4908 (including the investigation comment) and this PR's full diff (`.github/workflows/static.check.scripts/doxygen-tag.sh`, `test_doxygen_tag.sh`, `static.check.yml`, and the 34 doc-only files), traced the checker's bash logic by hand (no `ctags` available in the review sandbox, so the trace below is manual, not executed), spot-checked ~20 of the newly added `@brief` comments against the functions they document, checked the live CI run, and checked `CLAUDE.md`/`Documentation/` for anything the change would need to update.

## Does it fix #4908?

Yes, on the mechanism the issue reported. The stray `local` at script top level is gone, `report_path` is now defaulted so the `ambiguous redirect` noise is gone, and `$brief`/`function_positions`/`structure_positions` are reset per file so state no longer leaks across files in `$changed_file_list` — all three defects named in the issue's investigation comment are addressed. The scope choice (`f` — definitions — checked everywhere, `f+p` — definitions + prototypes — only in `.h/.hh/.hpp`) and the `/**<` trailing-doc recognition match exactly what that investigation comment proposed, down to the file list (the one omission, `gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh`, is correctly left untouched since its 9 hits were already documented via `/**<` and are fixed by the checker change alone — verified by reading that file).

## Blocking

**1. CI is currently red, and never reaches the check this PR restores.** The "Static checks" job (run https://github.com/nnstreamer/nnstreamer/actions/runs/33585795130/job/100109546372) fails at the *earlier* `/Checker/ clang-format` step, on `tests/nnstreamer_filter_mxnet/Predictor.hh`:
```
-  LG << "INFO:"
-     << "Finished inference with: " << nBatch * input_shape_[0] << " images ";
+  LG << "INFO:" << "Finished inference with: " << nBatch * input_shape_[0] << " images ";
```
This is inside `LogInferenceResult`, a function this PR never touches (the PR only adds `@brief` comments elsewhere in the same file). Reproduced locally: clang-format 16.0.6 against the repo's `.clang-format` produces **no** diff on this file, so the failure is a version drift between whatever clang-format the GitHub-hosted `ubuntu-latest` runner currently installs and what last formatted this file — pre-existing, unrelated to this PR's content, but it still blocks merge because the workflow's clang-format step runs on every file in `$changed_file_list`, and this PR's own doc commit touches that file. Because of this, the job never gets to the doxygen-tag step or the new self-test step, so **neither has actually been verified green on this PR**. Suggest either folding a `clang-format -i` pass (with the CI's current clang-format) into this PR, or landing a small prerequisite fix and rebasing.

## Should-fix

**2. `has_trailing_doc()` has a false-negative when the *next* declaration carries its own same-line `/**<`.** The function only checks whether `file_lines[i+1]` *contains* `/**<` anywhere — it doesn't verify that line is a standalone comment continuation for the function being checked, versus a different declaration that happens to end with its own trailing doc. Concrete instance already in this PR's diff, `ext/nnstreamer/android_source/gstamcsrc_looper.h:57-60`:
```cc
    /** @brief Terminate the loop by posting a flushing exit message */
    void exit (void);
    void post (gint cmd, void *data, bool flush);
    void (*handle) (gint cmd, void *data);  /**< should be implemented */
```
`post` has no `@brief` and no doc of its own. Every other sibling in this class (`loop`, `start`, `exit`, `entry`, `add_msg`, `flush_msg`) received a new `@brief` in this PR's "Doc/ext" commit — `post` is the sole undocumented one, and it looks documented to the checker only because `handle`'s own trailing `/**<` sits on the very next line. Tracing the script by hand: `post`'s own line ends in `;` at depth 0, so `has_trailing_doc` looks at `file_lines[i+1]` (`handle`'s line), sees `/**<` in it, and returns "documented" — for the wrong declaration. This is a real hole in the new code (not pre-existing), and it means the very case the PR sets out to close (`post` was presumably missed by hand and should have been caught by the checker) instead passes silently.
Suggested fix: only accept the next-line case when that line, once trimmed, actually *starts* with `/**<` and introduces no other tokens before it (i.e. it's a standalone doc continuation, not another declaration's own trailing comment) — which is exactly the shape the `trailing_next_line.hh` fixture already tests, just not with a subsequent *documented* declaration adjacent to the gap. Recommend adding a fixture to `test_doxygen_tag.sh` for this exact shape (undocumented declaration immediately followed by a declaration with its own same-line `/**<`) and fixing `post` itself.

## Nit / informational (pre-existing, unchanged by this PR, so out of strict scope, but worth knowing)

**3.** The unmodified `$brief` state machine (`elif [[ $line != *"*"* && ( $line =~ ";" || ... ) ]]; then brief=0`) never resets `brief` on any line containing a raw `*` — which also matches ordinary pointer parameters, not just doxygen comment-continuation lines. That lets a `brief=1` state leak across a pointer-parameter prototype into the next function. Example (unrelated to this PR's own changes, but explains why these two are untouched): `ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.h` — `ServiceImplFlatbuf (const grpc_config * config);` and the following `void parse_tensors (Message<Tensors> &tensors);` are both undocumented but pass, because the class-level `@brief` leaks through the pointer-bearing constructor line. Same pattern in `nnstreamer_grpc_protobuf.h`. Not blocking this PR (the line is unchanged, and the issue's own backlog count was generated against this same pre-existing behavior), but it means the restored check is not airtight — worth a follow-up issue.

## Regression / false-positive risk

Checked the brace-depth tracking in `has_trailing_doc` for the styles actually used in this codebase (K&R return-type-on-its-own-line functions, inline bodies, multi-line signatures ending in `= 0;`) against the fixtures in `test_doxygen_tag.sh` and against real examples in `nnstreamer_cppplugin_api_filter.hh` (`invoke_dynamic`'s inline body followed by `/**<`, `getModelInfo`'s two-line `= 0;` signature) — all trace through correctly. One theoretical gap not covered by any fixture: a `{`/`}` inside a `//` or `/*...*/` comment on the same line as a declaration would desynchronize the depth counter (comments aren't stripped before counting braces), which could make the scan run past the real end of a declaration. Low practical risk given this codebase's style, but worth a one-line comment in the script noting the limitation, or a fixture pinning current behavior so a future contributor doesn't "fix" it accidentally into a different bug.

`mapfile -t file_lines < "$file"` — no issue; reads the whole file into memory once per file, unremarkable at this codebase's file sizes.

exuberant-ctags vs universal-ctags — the issue's investigation showed these diverge on the previously-broken `--c-kinds=` empty-string case. This PR's fix passes a real kind list (`f` or `f+p`) in all cases now, which is the well-defined, documented case for both implementations, so the specific divergence from the issue no longer applies. CI installs `exuberant-ctags` explicitly (`static.check.yml` package list), matching what `test_doxygen_tag.sh` will exercise there; a contributor running the self-test locally with `universal-ctags` (common on current distros) may still see different results, but that's a pre-existing environment-matching concern rather than something this PR regresses.

## Size / scope

Proportionate. The script/workflow/test changes are the minimum needed to restore the check without red-walling every PR; the 34 documentation-only files are exactly the backlog the issue's own investigation enumerated (top table cross-checked 1:1 against this PR's file list, modulo the one file correctly left out — see above). No unrelated modules are touched.

## Test coverage

`test_doxygen_tag.sh` (15 cases) covers: the negative control (undocumented definition fails — this is the regression guard for #4908 itself), the `f` vs `f+p` scope split via a forward-declaration-in-`.c` case, `/**<` on the same line / next line / after an inline body, a documented-then-undocumented pair that must still fail (closest existing case to finding #2, but not quite the same shape — see above), struct `@brief`, per-file reset in both directions, and the file-level tag check. This is a solid regression suite for the issue itself and reasonably resistant to future drift in unrelated modules, since it runs against synthetic fixtures in a temp dir rather than real source files (so it won't be perturbed by unrelated code changes) — the gap is only the one adjacency case in finding #2.

## CI wiring or self-test

The self-test step and its `git show --diff-filter=D ... | grep -qxF` "don't silently drop the self-test" guard follow the identical pattern already used for the other self-test steps later in the same `static.check.yml` (byte-for-byte same guard shape) — consistent, and it does its job of making sure a future removal of `test_doxygen_tag.sh` can't happen without the workflow step being updated too. Not yet verified as green end-to-end on this PR, because of blocking issue #1 above.

## Architecture / API docs

No architecture or public API change here (CI script + comments only), so no doc update is required by policy. `CLAUDE.md`'s existing line ("Every function ... needs a Doxygen comment; CI enforces this") becomes true again once this lands and issue #1 is resolved; it doesn't need editing since it doesn't currently claim anything about scope (definitions vs. prototypes) that this PR would contradict. `Documentation/doxygen-documentation.md` isn't touched and doesn't contradict the new scope rule either — it's example-based and doesn't state a definitions-vs-prototypes policy either way, so nothing there needed updating.

## Content spot-check

Sampled roughly 20 of the newly added `@brief` comments against their implementations across the gRPC (`grpc_get_listening_port`'s "-EINVAL for a client instance" checks out exactly against `NNStreamerRPC::getListeningPort()`), DALI, TensorRT10, Tizen dummy-sensor, mlagent mock, edge custom-test, and bounding-box decoder files, plus the `cppfilter_test.hh` invoke bodies (`in*2`/`in+1` vs. the actual arithmetic in `cppfilter_test.cc`). All matched the real behavior; found no inaccurate description.

## Verdict

Not mergeable as-is: CI is red (blocking #1), and the restored check has a concrete, traceable hole (should-fix #2) that lets the exact kind of gap this PR exists to close slip through, demonstrated by `post` sitting undocumented in this very PR's diff. The overall design (scope split, `/**<` recognition, per-file reset, self-test wiring) is sound and well-matched to the issue, and the documentation backlog clearing is accurate everywhere I sampled it. Once the clang-format drift is resolved and CI actually runs the doxygen-tag + self-test steps, and `post` (plus the `has_trailing_doc` adjacency case) is fixed, this looks ready.

myungjoo:

Thanks for the review. All three points are addressed in the three commits just pushed:

- **Blocking, clang-format on `Predictor.hh`.** Confirmed: the runner now has clang-format 18, which joins adjacent string literals in a `<<` chain, and 16 (what I ran locally) breaks after each `"INFO:"`. The file had not been touched since the runner moved, so any PR editing it trips on lines it did not write. Commit `[Test/mxnet] Lay out the log lines the way clang-format 18 wants` applies the exact layout from the CI patch.
- **Should-fix, `has_trailing_doc` adjacency.** Fixed: the next-line check now requires the line to *open* with `/**<`, so a trailing comment on the following member no longer covers the declaration above it. `Looper::post` is documented, and `trailing_belongs_to_next.hh` pins the case (the previous checker passes it vacuously).
- **Nit, pointer-line `$brief` reset.** Taken into this PR rather than deferred: the reset now skips only lines that start as comments (`*`, `/*`, `//`), not any line containing `*`. Over the whole tree that adds ten findings in six files (the gRPC constructors and `parse_tensors`, `get_tfsp_with_checks`, `nnsconf_subplugin_dump`, the two subplugin custom-property prototypes); all are documented in `[Doc] Add the briefs the tightened checker uncovers`, and `pointer_line.h` pins the rule.

Local state after the changes: 16/16 self-test cases pass, the full tree (289 files) is clean with no stderr output, and against the checker on `main` 16/16 self-test cases fail.

myungjoo:

Scope: read #4908 (including the investigation comment), the previous automated review on this PR and the "Relayed on behalf of the author" reply, the full PR diff at HEAD (9a3fea62), and specifically re-diffed the three response commits (`git diff 4d8e01c6..HEAD`: 9c2831dc, f9c6c5c0, 9a3fea62) against the two issues the previous review raised. Traced `doxygen-tag.sh` by hand line-by-line for both the fixed cases and some new ones I went looking for, cross-checked against real files in the tree, ran `clang-format 16.0.6` locally against `Predictor.hh` for comparison, and watched this PR's own CI (`Static checks`, run 33586991251) to completion rather than trust the description.

## Did the response commits resolve the previous review?

**Blocking #1 (clang-format red on `Predictor.hh`) — resolved, verified live.** `9c2831dc` reformats exactly the four `LG << "INFO:" << ...` lines the previous review's CI log showed failing, to the clang-format-18 layout. I reran `clang-format --style=file` (v16.0.6, what's available in this sandbox) on the file and confirmed it reformats *back* to the pre-PR layout — i.e. this is a genuine version-drift fix, not something that happens to look right. More importantly, this PR's own `Static checks` run (https://github.com/nnstreamer/nnstreamer/actions/runs/33586991251/job/100113102530) completed **success** while I was reviewing, with `/Checker/ clang-format for .cc/.hh/.hpp/.cpp files` logging "clang-format shows no style errors." The job now also reaches the doxygen-tag steps, which the previous review noted it never had before.

**Should-fix #2 (`has_trailing_doc` crediting `post` with `handle`'s trailing `/**<`) — resolved.** `9a3fea62` tightens the next-line check from "line i+1 contains `/**<` anywhere" to "line i+1, trimmed, starts with `/**<`" (`^[[:space:]]*/\*\*\<`). `Looper::post` gets its own `@brief` in `f9c6c5c0`, and `trailing_belongs_to_next.hh` pins exactly this shape. Confirmed in the live CI self-test log: `PASS: a /**< on the next declaration does not cover this one (exit 1)`.

**Nit #3 (pending `@brief` surviving any line containing `*`, e.g. pointer parameters) — taken up rather than deferred, and resolved for the reported shape.** `9a3fea62` narrows the reset-skip condition from "line doesn't contain `*`" to "line doesn't start with `*`, `/*`, or `//`" (`^[[:space:]]*(\*|/\*|//)`). The twelve declarations this uncovers across `nnstreamer_grpc_common.h`, `nnstreamer_grpc_flatbuf.h`, `nnstreamer_grpc_protobuf.h`, `nnstreamer_cppplugin_api_filter.hh`, `nnstreamer_conf.h`, and `nnstreamer_subplugin.h` are documented in `f9c6c5c0`, and `pointer_line.h` pins the rule (`void add_one (int *x); void add_two (int *x);`). Live CI: `PASS: a brief does not carry over a pointer-typed declaration (exit 1)`.

No new regression from these three commits: I re-traced `has_trailing_doc`'s brace-depth loop and the `$brief` state machine against the fixture set and against the real multi-line/`/**<`-bearing declarations already in the tree (e.g. `nnstreamer_cppplugin_api_filter.hh`'s `cpp_getModelInfo`), and they all still resolve the same way as before the tightening. `test_doxygen_tag.sh` is now 16 fixtures (was 14); all 16 pass on this branch, all 16 fail on `main`, confirmed in the live log, not just the PR description.

## New finding from this round (should-fix, not blocking)

**The pointer-line `$brief` fix narrows the false-negative window but doesn't close it: a code line is still misclassified as a comment-continuation whenever it merely *starts* with `*`, which also matches a wrapped pointer parameter/return type at the start of a continuation line — as opposed to mid-line, which is what the new fixture (`pointer_line.h`) and the twelve documented instances all test.** This is a real, pre-existing pattern in the tree, not a hypothetical:

- `gst/nnstreamer/elements/gsttensor_converter.c:170-172`:
  ```c
  static const NNStreamerExternalConverter *findExternalConverter (const char
      *media_type_name);
  ```
  The continuation line `    *media_type_name);` starts with `*` and ends in `;`, so under the new rule it's treated as a comment-continuation and does **not** reset a pending `brief`.
- `ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc:458,467,478`: `*idx);` as a wrapped call-argument continuation, same shape.

In both of these specific spots nothing currently relies on the state staying set past that point (the next `@brief`-checked declaration has its own comment), so this doesn't produce an observable false pass anywhere in the current 289-file tree — the self-test and the PR's own regression claim aren't contradicted. But it's the same class of gap the response commit set out to close, just one degree narrower, and it sits in code this PR itself just rewrote, so a future contributor could plausibly hit it (an undocumented function/struct placed right after a multi-line pointer-typed forward declaration or signature would silently pass). Given this PR already added a fixture and a code comment for the *other* known limitation (braces inside same-line comments, in `has_trailing_doc`'s docstring at the top of the function), the same treatment here would be consistent: either track "inside an open `/* */` block" as real state instead of a per-line prefix heuristic, or at minimum add a fixture that pins the current (still-imperfect) behavior so it's a deliberate, visible limitation rather than a silent one. Severity: should-fix as a design gap in the code this PR authored, but not blocking — it doesn't regress anything currently checked and mirrors the tenor of the informational finding the previous review already accepted as non-blocking for the analogous `has_trailing_doc` brace-in-comment case.

## Content spot-check: the 11 (12 by my count, across 6 files) briefs added in `f9c6c5c0`

Verified each against its `.cc`/`.c` implementation:

- `Looper::post` ("Queue a command for loop () to hand to the handler") — matches: builds a `looper_message` and calls `add_msg`.
- `NNStreamerRPC::NNStreamerRPC`/`~NNStreamerRPC` — ctor builds the member-init list and calls `gst_data_queue_new`; dtor does `g_clear_pointer (&queue_, gst_object_unref)`. Both descriptions match.
- `ServiceImplFlatbuf::parse_tensors` / `ServiceImplProtobuf::parse_tensors` ("Parse the … tensors and deliver the buffer via callback") — matches the `_get_buffer_from_tensors` + `cb_ (cb_data_, buffer)` body (the `else gst_buffer_unref (buffer)` fallback when there's no callback isn't mentioned, but that's a reasonable `@brief`-length simplification, not an inaccuracy).
- `AsyncServiceImplFlatbuf`/`AsyncServiceImplProtobuf` ctor/dtor ("Construct … with no call in flight" / "Delete the last outstanding call, if any") — matches `last_call_ (nullptr)` init and `if (last_call_) delete last_call_;`.
- `get_tfsp_with_checks` ("Cast ptr to a subplugin after checking its sanity marker") — matches the cast + `sanity != _SANITY_CHECK` (and null/`subplugin_data` non-null) checks.
- `nnsconf_subplugin_dump`, `subplugin_set_custom_property_desc`, `subplugin_get_custom_property_desc` — the added header comments are verbatim-consistent with what the matching `.c` definitions already do.

No inaccurate description found.

## Everything else re-checked against the original review criteria

- **Solves #4908**: yes — the stray top-level `local` is gone, `report_path` defaults to `/dev/null`, and `brief`/`function_positions`/`structure_positions` are declared inside the per-file loop (confirmed by reading `doxygen-tag.sh:100-104`), so state doesn't leak across files in `$changed_file_list`. All three defects named in the issue are addressed, matching the investigation comment's proposed shape exactly (including the `f` vs `f+p` scope split and `/**<` recognition).
- **Size/scope**: proportionate. The `Predictor.hh` clang-format commit is a legitimate, narrowly-targeted fix for a pre-existing tooling-version drift this PR's own edit exposed, not a drive-by reformat — confirmed above. No other module is touched beyond the checker, its self-test, the workflow wiring, and the 35 backlog files the issue's own investigation enumerated.
- **Test coverage**: `test_doxygen_tag.sh`'s 16 fixtures run against synthetic files in a temp dir, so they're immune to unrelated future source changes, and they now cover both defects this round fixed (`trailing_belongs_to_next.hh`, `pointer_line.h`) in addition to the original 14. The one gap is the should-fix above (no fixture for a `*`-led continuation line).
- **CI actually blocks regressions**: yes. The self-test step (`static.check.yml:86-98`) follows the identical `git show --diff-filter=D … | grep -qxF` "don't silently drop the self-test" guard used by every other self-test step in the same file, and — unlike at the time of the previous review — it has now actually run green end-to-end on this PR (log confirms all 16 `PASS:` lines and `test_doxygen_tag.sh: all checks passed.`).
- **Architecture/API docs**: none needed and none touched — this is a CI-script and comment-only change; `CLAUDE.md`'s existing "CI enforces this" line becomes true again once merged, and doesn't need editing.
- **CI status at review time**: `Static checks` — the only job this PR's content can plausibly affect — is green. The remaining jobs (Meson/GBS/Yocto/NDK builds, Analyze, Tizen) were still in progress at review time; they build C/C++/Python code this PR does not touch beyond adding comments, so I don't expect them to be affected, but I didn't wait for every one of them to finish.

## Verdict

Mergeable. Both points from the previous review (blocking CI red, should-fix `has_trailing_doc` gap) are genuinely fixed and verified against a completed CI run, not just the commit messages. The nit was also fixed rather than deferred. The one new item I found (residual `*`-led-continuation-line gap in the same `$brief` reset rule) is real and demonstrated with concrete in-tree locations, but it's narrower than what this PR closes, doesn't produce any currently-observable false pass, and is the same *kind* of limitation the PR already accepted and documented for `has_trailing_doc`'s brace tracking — worth a fixture/comment in a follow-up, not a blocker for this PR.

myungjoo:

Thanks. The remaining should-fix is taken into this PR as well, in `[CI] Track block comments instead of guessing from a leading '*'`: the scanner now keeps an `in_comment` state for `/* ... */` blocks and skips the tag reset only inside one (or on a line that opens one), instead of treating any line starting with `*` as a comment continuation. Two fixtures pin it: `wrapped_pointer.h` (a parameter list wrapped so that `*x);` starts the line must clear the tag) and `semicolon_in_comment.h` (a `;` inside the comment block must not). 18/18 fixtures pass locally, the full tree stays at zero findings, and 18/18 fail against the checker on `main`.

myungjoo:

Scope: read #4908, the two prior automated reviews on this PR and the two "Relayed on behalf of the author" replies, and the full PR diff at HEAD (`4490a668`, `git diff origin/main...HEAD`). Priority target per the maintainer's request was the newest response commit, `4490a668` ("[CI] Track block comments instead of guessing from a leading '*'"), which answers the second review's remaining should-fix (a code line starting with `*`, e.g. a wrapped pointer parameter, was being mistaken for a comment-continuation line). Hand-traced `doxygen-tag.sh`'s `brief`/`in_comment` state machine line-by-line against: same-line `/* ... */`, two comments on one line, trailing `/**< ... */` after code, `//` lines, and the two new fixtures. Cross-checked the two concrete counterexamples the second review named (`gst/nnstreamer/elements/gsttensor_converter.c:170-172`, `ext/nnstreamer/tensor_filter/tensor_filter_pytorch.cc`) against the fixed script. Grepped the tree for further counterexamples the fix might still miss. Read `test_doxygen_tag.sh` end to end for internal consistency across all 18 fixtures. Checked branch protection (`Static checks` is a required status check on `main`) and watched CI on `4490a668`.

## Priority item: commit 4490a668 (`in_comment` tracking)

**Correct, and it fixes a real bug, not just the synthetic fixture.** Before this commit, the reset-skip regex was `^[[:space:]]*(\*|/\*|//)` — any line starting with a bare `*` (after optional whitespace) was treated as a comment-continuation line, including a wrapped pointer parameter such as:

```c
static const NNStreamerExternalConverter *findExternalConverter (const char
    *media_type_name);
```

(`gst/nnstreamer/elements/gsttensor_converter.c:170-172`, verified still present at HEAD). The continuation line `    *media_type_name);` starts with `*` and ends in `;`, so the old rule kept any pending `brief` alive across it. The new code drops the bare `\*` from the regex and instead tracks real block-comment state in `in_comment`, gated as: skip the reset only when `in_comment == 1` (already inside a `/* */` block) or the line itself opens one. Hand-tracing this against the two named real-tree locations and the two new fixtures:

- `wrapped_pointer.h` (`void add_one (int\n *x);` then `void add_two (int *x);`): `add_one`'s own position is checked while `brief` is still 1 (correct, it has a real `@brief`); the wrapped `*x);` line then has `in_comment == 0` and doesn't match `^[[:space:]]*(/\*|//)`, so `brief` resets to 0 before `add_two` is checked → `add_two` correctly fails. This is exactly the `gsttensor_converter.c` shape.
- `semicolon_in_comment.h` (a `;` inside a `/** ... */` block): `in_comment` is 1 for every line between the opening `/**` and the closing `*/`, so the reset-check's `in_comment -eq 0` guard skips all of them regardless of the embedded `;` — `brief` survives to the next declaration as intended.
- Same-line open+close (`/* foo */`), and two comments on one line (`/* c1 */ code /* c2 unclosed`, `/* c1 */ code /* c2 */`): the `${line##*/\*}` suffix-after-last-`/*` check correctly resolves both to "still open" and "closed by EOL" respectively, because `##` performs a greedy longest-prefix strip up to the *last* `/*` on the line.

I also grepped the tree for the two classes of case that could still fool this heuristic — a `//` line whose text happens to contain a literal `/*` substring, and a string literal containing an unterminated `/*` — and found no live instance in `*.c/*.h/*.cc/*.hh/*.cpp/*.hpp`. So the fix closes the counterexample the second review actually found, without introducing a new one that's currently reachable.

**Nit (informational, not blocking).** The `in_comment` tracker is still a lexical heuristic, not a real tokenizer: it can't distinguish an actual `/*` comment-opener from `/*` appearing inside a string literal or (in principle) a `division-then-dereference` expression like `x / *p` written without a space. Today's tree has neither pattern (checked by grep), and the PR already accepts and documents an analogous limitation for `has_trailing_doc`'s brace tracking, so I'd treat this the same way: worth a one-line comment or a pinned "known limitation" fixture in a follow-up, not something this PR needs to chase further.

## Fixture set (`test_doxygen_tag.sh`, 18 cases)

Re-read all 18 `run_checker` calls after the two additions. They stay internally consistent — no fixture's expected exit code depends on behavior the new two contradict, and `wrapped_pointer.h` / `semicolon_in_comment.h` pin the two directions of exactly the rule that changed (a `*`-led code line does reset; a `;` inside a real comment block does not). `pointer_line.h` (mid-line pointer, added by the previous commit) and `wrapped_pointer.h` (line-leading pointer, this commit) are complementary, not redundant.

## CI status

`Static checks` is a required status check on `main` (confirmed via branch protection: `required_status_checks.contexts` includes `"Static checks"`), and the self-test step (`static.check.yml`) runs `test_doxygen_tag.sh` unconditionally, with `exit 1` on any fixture failure, wired ahead of the "Indent check" step — so a regression in the self-test does fail the job and blocks merge, not just print a warning.

At the time of this review, the `Static checks` job for `4490a668` (run `33588524595`) was still queued, well past the "few minutes" window — other jobs on the same run progressed and several completed (`Android NDK`, `macOS`, `Spell Check`, `CPP-Linter`, `Ubuntu pdebuild` all passed) while `Static checks` itself stayed queued with `runner_id: 0`, which reads as shared-runner queue backlog rather than anything specific to this push. I was not able to observe the job's own log output (all 18 `PASS:` lines) before finishing this review; recommend re-checking `gh pr checks 4914 --repo nnstreamer/nnstreamer` once it clears the queue before merging, purely to confirm the log matches this hand-trace (I have no reason from the trace above to expect a failure).

## Whole-PR check

- **Solves #4908**: yes. The stray top-level `local` is gone, `report_path` defaults to `/dev/null`, `brief`/`in_comment`/`function_positions`/`structure_positions` are all reset per file, and the per-function `@brief` check now actually runs (was previously vacuous). Confirmed already-verified by the two prior reviews and unaffected by this round's change.
- **Regression risk to other modules**: none beyond the checker script itself. This round's diff touches only `doxygen-tag.sh` and `test_doxygen_tag.sh`; no production code path.
- **Size/scope**: proportionate — a 17/18-line fix plus two fixtures, directly answering the exact should-fix raised in review, nothing broader bundled in.
- **Test coverage / CI enforcement**: the fix ships with fixtures for both directions of the new rule, and the self-test is a required, blocking CI step (see CI status above), so a future regression of this exact heuristic would be caught.
- **Docs**: none needed — this is a CI-script-only change with no architecture/API surface.

## Verdict

Mergeable on content: the `in_comment` fix is correct, it resolves the real (not just hypothetical) counterexample the previous review found in `gsttensor_converter.c`, and the fixture set stays internally consistent. The only open item is operational, not code-quality — confirm the `Static checks` job actually finishes green on `4490a668` once the runner queue drains, since I could not observe its log directly at review time.

PR #4916, opened 2026-09-02
[Packaging] Add Ubuntu 26.04 (resolute) to the PPA build

myungjoo:
Repository half of publishing Ubuntu 26.04 (resolute) in `ppa:nnstreamer`; part of #4902, follow-up to #4848 / #4895 / #4904. Companion: nnstreamer/nnstreamer-edge#256.

## Why nothing publishes for 26.04 today

The Launchpad recipe build installs the `Build-Depends` of the **committed** `debian/control` before `debian/rules` ever runs (recipe build 3987572 on plucky parks in `Dependency wait` listing `gcc-9 ... nnstreamer-edge-dev ssat libedgetpu-dev openvino-dev onnxruntime-dev`), and `debian/rules` then copies `debian/control.ubuntu.ppa` over it before the source package is generated, so that copy is what the binary build resolves. Both files name packages that exist only in the PPA, and three of those frameworks cannot be rebuilt for a 2026 toolchain: `openvino` 2019R3 pins `gcc-10` (recipe build 3843589), `onert` 1.21 and `tvm` 0.7 are 2020-era sources, and the `onnxruntime` recipe has failed on every series since 2025-01.

Ubuntu 26.04 itself carries `libonnxruntime-dev` 1.23.2 with the `libonnxruntime.pc` that `meson.build` probes, so onnxruntime no longer needs the PPA there. Its `libtensorflow-lite-dev` 2.14.1 is not usable by nnstreamer, though: no pkg-config file, only a CMake config without a version file, a `tensorflow/lite/version.h` that includes a header the package does not ship, and no delegate headers. tflite therefore stays a PPA package on 26.04, rebuilt from the existing `tensorflow2` recipe.

## What changes

**`debian/control`** is now the file for Ubuntu 26.04 and later and, because of the source-stage behaviour above, must be satisfiable on every series the PPA targets:

- `onnxruntime-dev | libonnxruntime-dev`: neither is preinstalled in a build chroot, so apt takes the first alternative that exists, i.e. the PPA package on 24.04 and earlier and the archive package on 26.04 (`libonnxruntime-dev` does not exist on jammy or noble; checked against the archive).
- openvino, onert and tvm build-dependencies and their binary stanzas are dropped from this file; `nnstreamer-onnxruntime` relies on `${shlibs:Depends}` instead of the PPA-only `onnxruntime` package name.
- The `pytorch`/`caffe2` stanzas go too. `pytorch | libtorch-dev | gcc` never installs `pytorch`, because an already-installed alternative wins and `build-essential` brings `gcc`, so the subplugin is not built and `debian/rules` drops its `.install` file; and the PPA's `pytorch`, which both packages declare in `Depends`, is not built for 24.04 or later, so they are not installable there either. The header comment and the PPA document say so.
- A comment at the top explains which file applies at which stage, including that a plain `dpkg-buildpackage` on 24.04 and earlier now checks its build dependencies against this file while `debian/rules` still enables openvino, so it fails at meson configure rather than at `dpkg-checkbuilddeps`. pbuilder, which is what the PPA uses, is unaffected.

**`debian/control.ubuntu.ppa`** is untouched: Ubuntu 24.04 and earlier keep exactly the stack they have.

**`debian/rules`** reads `VERSION_ID` from `/etc/os-release`; below 26.04 it copies `control.ubuntu.ppa` as before and keeps `-Denable-openvino=true`, from 26.04 on it leaves the committed control in place and passes `-Denable-openvino=false`. `-Dtflite2-support=enabled` is added so a missing tflite fails configure instead of silently producing a package without it. `override_dh_auto_test` now honours `DEB_BUILD_OPTIONS=nocheck`: dh skips `dh_auto_test` under nocheck but, at this compat level, not an override target, which answers the question left open in #4902 (a nocheck build ran the entire suite).

**`tests/nnstreamer_decoder_image_segment`**: case 1 compared the *videomixer* output against a golden. On 26.04 (GStreamer 1.28.2) that blend differs from the golden by one unit in 8320 pixels of the lower rows, while the scaled video, the tensor and the decoded mask are byte-identical to what the released noble package (GStreamer 1.24.2) produces, so this single case failed the package build there. The decoder output is now teed into the compare, the way the boundingbox decoder test does it, and the golden is the decoder output, identical from the 26.04 build and from the noble PPA package. The modified test still fails against the old golden on 26.04 (negative control).

**CI** gains `static.check.scripts/test_debian_rules_series.sh`, which pins the new packaging logic without needing a chroot: for a table of `VERSION_ID` values it asserts which control file `override_dh_clean` copies and which `-Denable-openvino` value the configure step passes, plus that `DEB_BUILD_OPTIONS=nocheck` reaches the test override. `make` gives a command-line variable precedence over a `:=` assignment, so the release under test is supplied as `UBUNTU_VERSION_ID=` and no stub of `/etc/os-release` is needed; every case runs `make -n`, so `debian/control` is never overwritten. It was confirmed to fail when the 26.04 threshold is moved to 24.04 and when the nocheck guard is removed. `update_pbuilder_cache.yml` also stops saving a cache entry when `pbuilder create` failed, which would otherwise leave an entry with no `base.tgz` that a later run hits exactly.

**Docs**: `getting-started-ubuntu-ppa.md` says which sub-plugin packages 26.04 does not get, and marks the two that need the PPA's `pytorch`.

## Not in this PR

The `resolute` leg for `pdebuild.yml` is deliberately left out and will follow once the PPA carries the 26.04 dependencies. That leg cannot pass before then, and `check_if_rebuild_requires.sh` runs the pdebuild workflow for essentially every source change, so landing it now would put a permanently red pdebuild check on unrelated PRs and train reviewers to ignore the one signal this repository has against packaging regressions. The self-test above is the regression check in the meantime, and it runs regardless of the PPA's state.

## Verified

On an `ubuntu:resolute` container (gcc 15.2, meson 1.10.1, python 3.14, numpy 2.3.5, gstreamer 1.28.2):

- `debian/rules clean` picks the committed control and `-Denable-openvino=false -Dtflite2-support=enabled`; the same test in an `ubuntu:noble` container picks `control.ubuntu.ppa` and `-Denable-openvino=true`.
- `dpkg-checkbuilddeps` is satisfied by the new control, using the archive's onnxruntime plus nnstreamer-edge (nnstreamer/nnstreamer-edge#256) and SSAT packages built on 26.04, and the noble PPA builds of `tensorflow2-lite-dev` and `libedgetpu-dev` as stand-ins for the resolute rebuilds Launchpad still has to do.
- `dpkg-buildpackage -us -uc -b` compiles everything and runs `override_dh_auto_test`: all 25 gtest binaries pass; SSAT 1006 passed / 1 failed, the image_segment case above, which passes with the test fix.
- `debian-control.sh`, its self-test, `meson-version.sh` and the new `test_debian_rules_series.sh` pass on the edited files; shellcheck is clean on the new script.

## Launchpad status

`ssat` and the `edgetpu` binaries are published for resolute as of 2026-09-04. Still to do, each needing the series added to the recipe **and** a build requested, since adding a series to a daily recipe triggers nothing by itself:

1. `nnstreamer-edge-daily-2`, after nnstreamer/nnstreamer-edge#256 is merged
2. `tensorflow2-2.15.0-anyj0527`; its first resolute build failed before compiling because 26.04's `dpkg-source` rejects a two-person `Maintainer` field, so that packaging has to move to a repository the team can write to
3. `nnstreamer-daily`, after this PR is merged

`nnstreamer-ml-api-daily` and `nntrainer-daily` need the same `| gcc` build-dependency fix in their own repositories before they can follow. Plucky (EOL) can be removed from every recipe.


myungjoo:

Findings, by severity:

- **High — `.github/workflows/pdebuild.yml:119`, `update_pbuilder_cache.yml:95`: the resolute leg can never pass, not only "until the PPA publishes".** `pdebuild` without `--use-pdebuild-internal` runs `debian/rules clean` on the *host* (upstream `pdebuild`: `"${BUILDSOURCEROOTCMD}" debian/rules clean` before `dpkg-source -b`). The host is the `ubuntu-24.04` runner, so `debian/rules:34-41` sees `VERSION_ID=24.04`, copies `control.ubuntu.ppa` into the `.dsc`, and the resolute chroot then has to resolve `openvino-dev`, `onert-dev`, `tvm-runtime-dev`, `onnxruntime-dev` from the resolute PPA series, which this PR says will never exist. Fix: run that leg with `pdebuild --use-pdebuild-internal ...` (source stage inside the chroot, deps satisfied from the committed control, which is exactly the Launchpad flow), or skip the host clean with `-nc`. The PR text about the leg should be adjusted accordingly.

- **High — `pdebuild.yml:16`, `update_pbuilder_cache.yml:13`: add `fail-fast: false`.** The matrix default cancels in-progress legs when any leg fails. The resolute leg fails within minutes (`pbuilder update` on an empty PPA series), which cancels the hour-long jammy leg: on PRs that removes the regression check this PR relies on, and in the daily workflow the jammy log/badge steps never run.

- **Low — `debian/rules:34`: Ubuntu derivatives.** Linux Mint reports `VERSION_ID="22"`, elementary `"8"`, yet `dpkg-vendor --derives-from Ubuntu` is true there, so a derivative based on 26.04 silently takes the legacy control and `openvino=true`. Not a PPA target; a note in the comment or a `/etc/upstream-release/lsb-release` fallback would do. Empty `VERSION_ID` is fine (dpkg treats `""` as earliest; verified).

Checked and found correct: `make -n -f debian/rules` for `override_dh_clean`, `override_dh_auto_configure` and `override_dh_auto_test` on a 22.04 host and with `UBUNTU_VERSION_ID=26.04` (right control copy, `-Denable-openvino`, `-Dtflite2-support=enabled`, nocheck skip); ifeq nesting and `$$VERSION_ID`; the Debian path is unchanged. `debian/control`: leading `#` lines are allowed by Policy 5.2 and skipped by dpkg, python-debian and pbuilder's `get822field`; `onnxruntime-dev | libonnxruntime-dev` ordering; orphan `.install` files have precedent (armnn, grpc) and the `rm` lines in `rules:76-78` still find their files; only `tools/docker/UbuntuBuild/Dockerfile:98` mentions the dropped packages, with `rm -f`. Workflows: the second `include` differs in `os` from the original matrix value, so it creates a new combination; the `gutsy` symlink is relative inside the scripts dir; the jammy `if:` gates are valid. SSAT: the tee'd pipeline is valid, the new golden is 257x257x4 bytes (decoder RGBA), nothing else reads `test_golden.0`, `rm test_output.*` still matches. Commits: subjects conform, one `Signed-off-by` each. No remaining reference to "Pdebuild Ubuntu 22.04" in the tree; the README badge stays jammy-gated.

myungjoo:

Thanks, the first two are real and are fixed in the follow-up commit:

- The resolute leg now runs `pdebuild --use-pdebuild-internal` (matrix key `pdebuild_opts`), so `debian/rules clean` and the build-dependency resolution both happen inside the resolute chroot, as on Launchpad. The jammy leg is unchanged. The PR description is corrected accordingly.
- `fail-fast: false` on both matrices.

The derivative note is acknowledged but left alone: the PPA targets Ubuntu proper, and a derivative with its own VERSION_ID falls back to exactly the control it gets today.

myungjoo:

Confirmed fixed (checked against pbuilder 0.231, the noble version):

- `--use-pdebuild-internal`: `pdebuild` now calls `pbuilder --execute --bindmounts <parent> --distribution resolute -- pdebuild-internal <workspace>`; inside the chroot it satisfies the committed `debian/control`, then runs `dpkg-buildpackage -rfakeroot -us -uc` as `pbuser` (runner uid/gid), so `debian/rules clean` and `dpkg-source -b` see the chroot's `/etc/os-release`. `--architecture` is consumed by `pdebuild-checkparams` (changes-file name only), the `--distribution` pass-through is accepted by `pbuilder-checkparams` (only a harmless "--override-config is not set" warning), `--logfile`/`--buildresult` are pdebuild options in both modes, and `pbuilder execute` keeps the `sources.list` written at `pbuilder create` (backports + PPA). Bind-mount ownership is fine since `pbuser` has the runner's uid.
- `${{ matrix.pdebuild_opts }}` is empty on jammy; the command is unchanged there.
- `fail-fast: false` sits under `strategy` in both files; both parse.

Remaining defect:

- **Medium - `.github/workflows/pdebuild.yml:124`: the resolute leg fails after a successful build.** In internal mode `pdebuild` itself (running as `runner`, not via `sudo`) copies the results from `..` into `$BUILDRESULT`, which defaults to `/var/cache/pbuilder/result/` (`pdebuild` lines 68-72, `conditional_cp_a` -> `cp -a`, no `|| true`, `set -e`). The pbuilder package creates that directory root:root 0755; the `--mode a=rwx` in the workflow only covers `/var/cache/pbuilder` itself. `cp` gets EACCES and the step fails. The jammy leg never hit this because the non-internal copy is done by `sudo -E pbuilder --build`. Fix: pass `--buildresult` pointing at a runner-owned directory (as `update_pbuilder_cache.yml:100` already does), or `chmod a+rwx /var/cache/pbuilder/result` in the prepare step. Hidden today because the leg dies earlier at `pbuilder create`.

Note (no action): in internal mode the build and tests run with network access (`USENETWORK=no` applies only to `pbuilder build`), and `PBUILDERSATISFYDEPENDSOPT` is not forwarded (irrelevant with `[trusted=yes]`).

myungjoo:

Fixed in the follow-up commit: the PR pdebuild step now passes `--buildresult ~/pdebuild_result` (created first), so the internal-mode copy lands in a runner-owned directory. Applied to both legs, since nothing consumes the results of the PR workflow. The network note is accepted as-is: the tests already need the network for the MQTT broker checks and the leg is not a required check.

myungjoo:

The remaining defect is resolved (checked against pbuilder 0.231):

- `.github/workflows/pdebuild.yml:127-128`: bash expands `~/pdebuild_result` before `pdebuild` sees it; `pdebuild-checkparams` takes the (now existing) directory via `readlink -f`. In internal mode `BUILDRESULT` is used only on the host after `pbuilder --execute` returns (`pdebuild` lines 68-79, `conditional_cp_a` as the runner user) and is never passed into the chroot, so it need not be under the bind mount.
- Jammy leg: `--buildresult` is forwarded to `pbuilder --build`, which `mkdir -p`s, copies as root and chowns to the runner uid, as with the default; the `-ef` check in `pdebuild` line 93 is unaffected since `~/pdebuild_result` is not the checkout's parent.

Nothing else remains.

myungjoo:

Status for reviewers: every check is green except the new resolute pdebuild leg, which stops at the PPA's missing resolute index (404) exactly as described above. Launchpad now has ssat and libedgetpu for resolute in Pending publication; nnstreamer-edge follows once nnstreamer/nnstreamer-edge#256 is merged, and tensorflow2 needs its packaging moved to a team repository (its Maintainer field is rejected by 26.04's dpkg). The leg will be re-run after those publish.

myungjoo:

All five addressed; the branch is rebased onto current main and the series is now five commits with no change to `pdebuild.yml`.

**1. Permanently red leg.** Taken further than either option offered: the `resolute` leg is removed from this PR entirely and will land once the PPA carries the 26.04 dependencies. Your observation that `check_if_rebuild_requires.sh` sets `redebian=1` for any file outside `packaging/`, `jni/` and docs is what decided it. A red check on unrelated PRs is worse than a missing one, and `continue-on-error` risks outliving the reason for it. Item 4 becomes the regression check instead, and it runs whatever the PPA's state.

**2. Cache entry without `base.tgz`.** Confirmed and fixed. The create step gets `id: make-base-tgz` and the save is gated on `steps.make-base-tgz.outcome == 'success'`, so a failed `pbuilder create` no longer writes an entry that a later run hits exactly. Kept in this PR rather than deferred, since the same poisoning can happen on a transient `pbuilder create` failure with the jammy leg alone.

**3. `dpkg-checkbuilddeps` on ≤24.04.** Added to the `debian/control` header: a plain `dpkg-buildpackage` there now passes the dependency check and fails later at meson configure on the openvino that `debian/rules` still enables; pbuilder, which is what the PPA uses, is unaffected.

**4. Test coverage.** Added `static.check.scripts/test_debian_rules_series.sh`, wired into Static checks with the same merge-ref guard as its neighbours. It asserts, per `VERSION_ID`, which control file `override_dh_clean` copies and which `-Denable-openvino` value the configure step passes, plus the `nocheck` behaviour. No stub of `/etc/os-release` is needed because a make command-line variable outranks the `:=` assignment, and every case runs `make -n`, so `debian/control` is never overwritten. Per the lesson from #4904 it is a value-comparison table only, with no heuristic parsing. Verified it fails when the 26.04 threshold is moved to 24.04 and when the `nocheck` guard is removed; shellcheck is clean.

**5. Minor.** All three taken: `?=` defaults for `UBUNTU_LEGACY_PPA` and `ENABLE_OPENVINO` before the vendor guard; the plugin list now marks `nnstreamer-pytorch` and `nnstreamer-caffe2` as needing the PPA's `pytorch`, which is not built for 24.04 or later; and `runTest.sh` notes that the golden is an argmax over tensorflow-lite output, so a rebuild with another compiler is a golden to refresh rather than a decoder to debug.

On the last one your framing was more accurate than the PR body's. `nnstreamer-pytorch` and `nnstreamer-caffe2` binaries do exist for jammy and noble, and jammy also has the PPA's `pytorch`, so the packages are installable there; it is on 24.04 and later that the dependency is missing. The PR body has been corrected accordingly.

myungjoo:

Head `736102a2`, 5 commits. All five prior items are resolved.

1. **Permanently red leg** — resolved. `pdebuild.yml` is byte-identical to main; no `resolute` reference remains outside `debian/rules`' comments and the test's version table, and the PR body says why.
2. **Cache poisoning** — resolved, and `steps.make-base-tgz.outcome` is the right handle: no `cache/restore` exists here, so the step can never be `skipped`; no `continue-on-error`, so `outcome == conclusion` today and `outcome` stays right if one is added; and `always()` still saves when the later `run pdebuild` step fails, which is what it was there for.
3. **`dpkg-checkbuilddeps`** — resolved, `debian/control:13-17`.
4. **Test coverage** — resolved. Ran on an Ubuntu 22.04 host: 15 assertions pass, `debian/control` unchanged afterwards (md5), and no verdict moves with `UBUNTU_VERSION_ID`, `UBUNTU_LEGACY_PPA`, `ENABLE_OPENVINO` and `DEB_BUILD_OPTIONS=nocheck` all exported, since a command-line variable outranks env and `:=` alike. Mutations fail it: threshold 26.04→24.04 (4 FAILs), `nocheck`→`nochek` (2), renamed meson option (6), unparsable rules (12). `report`'s `$([ … ] && echo 0 || echo 1)` is correct and a non-Ubuntu host skips with exit 0.
5. **Minor three** — resolved (`debian/rules:36-37`: the `?=` defaults yield to env, the in-guard `:=` then wins on Ubuntu; verified).

New defects, by severity:

- **Low — `.github/workflows/static.check.scripts/test_debian_rules_series.sh:136`**: "with nocheck: SSAT is skipped" is the one assertion with no positive counterpart, so it passes vacuously if SSAT leaves the recipe. Verified: delete the `cd tests && ssat` line and the suite still exits 0 with three oks. A "without nocheck: SSAT runs" check, mirroring the unit-test pair, closes it.
- **Low — same file:57**: `expand` drops make's exit status, so an unparsable `debian/rules` makes each "keeps the committed control" branch report ok (the openvino assertions still fail, so no run goes silent).
- **Nit — the split**: `a64478bf` carries the `runTest.sh` rewrite and the new `test_golden.0`, while `6068ecb0`, whose message describes exactly that change, holds only its three comment lines. Worth moving those hunks. Subjects conform, one `Signed-off-by` each.

Also checked: the static.check.yml step matches its neighbours' merge-ref guard verbatim, and both workflows parse.

myungjoo:

All three fixed; head is now `52cb8766`.

**Vacuous "SSAT is skipped".** Every skip assertion is now paired with the positive one, through a shared `expect_match` helper: the unit tests and SSAT each have to be present without `nocheck` and absent with it. Deleting the `cd tests && ssat` line from the override now fails the suite, which it did not before.

**`expand` ignoring make's exit status.** It now reports the make error, marks the run failed and returns non-zero; every caller bails out on that. The exit status is checked in the `if !` condition rather than after a `local` assignment, and the function writes into a variable instead of being used in a command substitution, so the failure path is not swallowed by a subshell.

**Commit split.** Correct catch, and the cause was mine: the index still held the test files from an earlier `git reset --soft`, so the first commit swallowed them even though only the packaging files were named on the `git add`. The series is rebuilt from a `--mixed` reset with explicit staging per commit, and each commit now contains exactly what its subject says:

| commit | contents |
|---|---|
| `2493c28a` | `debian/control`, the series selection in `debian/rules`, the PPA document |
| `fde06191` | the `nocheck` guard |
| `1cc298d6` | `runTest.sh` and `test_golden.0` |
| `f8362b48` | the self-test and its Static checks step |
| `52cb8766` | the pbuilder cache gate |

Re-verified after the changes on an Ubuntu host: the suite passes on the tree as committed with `debian/control` unchanged (md5 compared before and after), and fails under each of four mutations — the 26.04 threshold moved to 24.04, the `nocheck` guard removed, the `ssat` line deleted from the override, and `debian/rules` made unparsable. shellcheck is clean on the rewritten script, and `pdebuild.yml` is still byte-identical to main.

myungjoo:

One more, found by the CI run of the previous push rather than by review: the new script was the only file in the tree producing a doxygen `documented symbol ... was not declared or defined` warning. The cause is `EXTENSION_MAPPING = sh=C` in `Doxyfile.prj`, so doxygen parses the script as C and reads a top-level `if` guarding an `echo "::error::..."` as a declaration, with the `::` taken for a scope operator. The body now sits in functions with a `main` at the end, which removes the warning; reproduced and confirmed gone with the repository's own Doxyfile before pushing.

Behaviour is unchanged: the suite still passes on the tree as committed with `debian/control` unmodified, and still fails under all four mutations. shellcheck stays clean. Head is now `b7ceaee3`.

Issue #4102, opened 2023-05-26
Trying to execute tutorial2_object_detection as a C code.


pedrobagatin:
I was able to connect most of the video stream bins, but having problems with the tensor stream when writing a C code for the example.

`# Move to default install directory.
$ cd /usr/lib/nnstreamer/bin
$ gst-launch-1.0 \
    v4l2src ! videoconvert ! videoscale ! video/x-raw,width=640,height=480,format=RGB,framerate=30/1 ! tee name=t \
    t. ! queue leaky=2 max-size-buffers=2 ! videoscale ! video/x-raw,width=300,height=300,format=RGB ! tensor_converter ! \
            tensor_transform mode=arithmetic option=typecast:float32,add:-127.5,div:127.5 ! \
            tensor_filter framework=tensorflow-lite model=tflite_model/ssd_mobilenet_v2_coco.tflite ! \
            tensor_decoder mode=bounding_boxes option1=mobilenet-ssd option2=tflite_model/coco_labels_list.txt option3=tflite_model/box_priors.txt option4=640:480 option5=300:300 ! \
            compositor name=mix sink_0::zorder=2 sink_1::zorder=1 ! videoconvert ! autovideosink \
    t. ! queue leaky=2 max-size-buffers=10 ! mix.`

What does mix command do at the end of the command?

Is there a way to test little by little the tensor pipeline? 
Ex: checking inference, seeing the bounding box, or checking coordinates.

Because if I try to write the pipeline all at once chances of making mistakes is big.

myungjoo:
> What does mix command do at the end of the command?

It feeds a stream to ```compositor name=mix sink_0::zorder=2 sink_1::zorder=1 ....```; its name is **mix**.

> Is there a way to test little by little the tensor pipeline?
Ex: checking inference, seeing the bounding box, or checking coordinates.

Yes. Write a minimal pipeline for the feature you want to test. That's what I do when I have a pipeline error. You can save tensors to a file w/ filesink and load tensors from a file w/ filesrc.

That's how we test each plugin in unit testing, too:

https://github.com/nnstreamer/nnstreamer/blob/da09f9f2d6cd1029982d0b312a84b8a171deeb91/tests/nnstreamer_filter_tensorflow2_lite/runTest.sh#L81

pedrobagatin:
When linking elements as follows:
`gst_element_link_many (tensor_queue, tensor_videoscale, NULL)`

Is okay, but adding complete tensor stream elements:
[...]

PR #4917, opened 2026-09-02
[Transform] Explain caps failures with the mode and option involved

### LINKED ISSUE #4103: tensor-transform: better exception handling for incorrect caps/operators
Refer: [#4102 ](https://github.com/nnstreamer/nnstreamer/issues/4102#issuecomment-1570862323)

If the given operator/options are not compatible with the given gst-caps,
don't just throw:
```
** (tutorial2:5602): CRITICAL **: 16:18:24.653: _check_tensors: assertion 'config->info.num_tensors >= limit' failed
0:00:00.404629212  5602 0x55bdebf17920 ERROR       tensor_transform gsttensor_transform.c:2135:gst_tensor_transform_set_caps:<tensor_transform> Tensor info is not matched with given properties.
0:00:00.404665914  5602 0x55bdebf17920 ERROR       tensor_transform gsttensor_transform.c:2161:gst_tensor_transform_set_caps:<tensor_transform> Set Caps Failed!

0:00:00.404686391  5602 0x55bdebf17920 WARN           basetransform gstbasetransform.c:1370:gst_base_transform_setcaps:<tensor_transform> FAILED to configure incaps other/tensor, framerate=(fraction)15/1, dimension=(string)3:256:256:1:1:1:1:1, type=(string)uint8 and outcaps other/tensor, framerate=(fraction)15/1, dimension=(string)3:256:256:1:1:1:1:1, type=(string)uint8
0:00:01.123340824  5602 0x55bdebf17920 ERROR       tensor_transform gsttensor_transform.c:2135:gst_tensor_transform_set_caps:<tensor_transform> Tensor info is not matched with given properties.
0:00:01.127890702  5602 0x55bdebf17920 ERROR       tensor_transform gsttensor_transform.c:2161:gst_tensor_transform_set_caps:<tensor_transform> Set Caps Failed!
```

It should throw it with more explanation, which part of the operator/options are not compatible with the given gst-caps. (e.g., ```the operator "typecast:float32" is not compatible with the given caps```)

myungjoo:
Fixes #4103.

## Problem

`tensor_transform` reported every `set_caps` failure as `Tensor info is not matched with given properties`, regardless of the actual cause, and an element whose `mode`/`option` were not configured passed caps negotiation and then died on the first buffer with a bare `assertion 'filter->loaded' failed` CRITICAL. The details that would explain the failure (type/dimension comparison) are emitted at `g_debug` level by the util library and are invisible under `GST_DEBUG=2`.

Reproduced on current main with six pipelines (see the table). The log quoted in #4103 (originally from #4102) has identical in/out caps and comes from the `convert_dimension` path, which means the element's `mode` was never applied in that application.

| Scenario | Before | After |
|---|---|---|
| `tensor_transform` with no mode | "Tensor info is not matched…" ×4 | `Transform is not configured (mode=unknown, option=(NULL)). Set valid 'mode' and 'option' properties before negotiating caps.` |
| `mode=typecast`, no option | caps accepted, then CRITICAL on first buffer | rejected at set_caps with `mode=typecast, option=(NULL)` |
| `mode=arithmetic option=casttype:…` (typo) | parse error, then CRITICAL | parse error, then rejected at set_caps with `mode=arithmetic, option=(NULL)` |
| `typecast:float32` vs downstream `types=float16` | only basetransform's generic "could not transform … in anything we support" | `mode=typecast option=float32 transforms input [uint8 …] into [float32 …], which the downstream element cannot accept: [float16 …]` |
| out caps that do not equal the transform result | "Tensor info is not matched…" | mode/option, framerate expected vs given, and a per-tensor expected-vs-given table |

## Change (`gst/nnstreamer/elements/gsttensor_transform.c` only)

- `set_caps` rejects an unconfigured element up front (`!filter->loaded`), naming the current mode and option. The two remaining failure paths name the tensor index, or print `gst_tensors_info_compare_to_string` output plus the framerates.
- While unconfigured, `transform_caps` advertises identity caps instead of running the per-mode conversion on unparsed option data. This is what makes the rejection above the one that users see even with a downstream capsfilter, and it removes the earlier attempt to pre-initialise the option union (which, being a union, leaked `_NNS_END` into `dimchg`/`padding`/`transpose` fields).
- `loaded` now means "the current mode and option parsed successfully": it is cleared when the mode is unknown or the option is unset, and when a (re-)parse fails, e.g. after switching `mode` to one under which the stored option is invalid. A successful re-parse never passes through an unconfigured state, so changing `option` at runtime keeps working.
- `transform_caps` warns with mode/option when the transformed caps cannot intersect the filter caps, and when fixed input caps are transformed into caps the downstream peer cannot accept. The downstream peer query mirrors what `tensor_filter` already does in the same direction and is skipped unless the category threshold is at least WARNING, so it costs nothing in production. The opposite direction (querying upstream from the SRC direction) is deliberately **not** done: with a `tensor_filter` upstream it would ping-pong caps queries forever. So a `dimchg`/`transpose` result that the upstream element cannot feed still fails silently inside the upstream element's negotiation; that is outside this element.
- `transform()` posts `GST_ELEMENT_ERROR (CORE, NEGOTIATION)` instead of the CRITICAL if it is ever reached unconfigured.
- One pre-existing spacing nit (`=(gsize)`) fixed so `gst-indent` leaves the file clean.

## Tests

GTest `testTensorTransform`:

- Negative, each asserting `GST_FLOW_NOT_NEGOTIATED` on the first buffer and the new message text through a `gst_debug_add_log_function` collector: `setCapsNotConfigured_n`, `setCapsModeWithoutOption_n` (fails on main, which returns `GST_FLOW_ERROR` after the CRITICAL), `setCapsInvalidOption_n`, `setCapsUnconfiguredPaddingWithDownstreamCaps_n` (the union side effect: an unconfigured `padding` with the real caps downstream must still reach the set_caps message), `setCapsAfterModeChangeUnparsable_n` (`loaded` is cleared when the stored option stops parsing under a new mode), `transformCapsIncompatibleDownstream_n` (peer-query warning) and `transformCapsFilterIntersectionEmpty_n` (filter-caps warning, both directions, via `gst_pad_query_caps` with a filter).
- Positive regression guards for existing pipelines and applications: `setCapsConfiguredAfterLink` and `linkThenConfigure` (elements linked with `gst_element_link_many` **before** properties are set, the pattern from #4102, with a fixed downstream capsfilter), `optionChangeWhileStreaming` (runtime `option` change, re-setting the same `mode`, and an invalid runtime option being rejected and the previous one kept), and `negotiationChainWithFilter` (`tensor_transform → tensor_filter → tensor_transform → tensor_filter → tensor_transform`, with the WARNING threshold raised so the peer-query path is really taken; a caps-query loop would fail with a bus timeout instead of hanging a job).

SSAT `transform_typecast` 25_n / 26_n: no mode, and mode without option. Note that these two also fail on main (via the CRITICAL), so they are pipeline-level guards rather than the regression guard for #4103; the GTest cases are.

Local run (WSL, meson build): `testTensorTransform.*`, `unittest_plugins`, `unittest_sink`, `unittest_common`, `unittest_if` all passed; SSAT `transform_*`, `nnstreamer_decoder_boundingbox`, `nnstreamer_decoder_pose`, `nnstreamer_filter_custom` all passed. `gst-indent` and `clang-format --dry-run -Werror` clean on the changed files.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -305,9 +305,31 @@ gst_tensor_transform_init (GstTensorTransform * filter)
   filter->operators = NULL;
   filter->acceleration = DEFAULT_ACCELERATION;
   filter->apply = NULL;
+  filter->data_typecast.to = _NNS_END;
 
   gst_tensors_config_init (&filter->in_config);
   gst_tensors_config_init (&filter->out_config);
+}
+
+/**
+ * @brief Get the nickname of the given transform mode for log messages
+ * @param[in] mode transform mode
+ * @return nickname string owned by the enum class (do not free)
+ */
+static const gchar *
+gst_tensor_transform_get_mode_string (tensor_transform_mode mode)
+{
+  GEnumClass *klass;
+  GEnumValue *val;
+  const gchar *nick = "unknown";
+
+  klass = g_type_class_ref (GST_TYPE_TENSOR_TRANSFORM_MODE);
+  val = g_enum_get_value (klass, mode);
+  if (val)
+    nick = val->value_nick;
+  g_type_class_unref (klass);
+
+  return nick;
 }
 
 /**
@@ -1777,7 +1799,7 @@ gst_tensor_transform_padding (GstTensorTransform * filter,
 
   in_loop_size = (gsize) in_info->dimension[2] * in_info->dimension[1]
       * in_info->dimension[0] * element_size;
-  out_loop_size =(gsize) out_info->dimension[2] * out_info->dimension[1]
+  out_loop_size = (gsize) out_info->dimension[2] * out_info->dimension[1]
       * out_info->dimension[0] * element_size;
   copy_block_size = in_info->dimension[0] * element_size;
 
@@ -1839,7 +1861,14 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
 
   filter = GST_TENSOR_TRANSFORM_CAST (trans);
 
-  g_return_val_if_fail (filter->loaded, GST_FLOW_ERROR);
+  if (!filter->loaded) {
+    GST_ELEMENT_ERROR (filter, STREAM, FAILED, (NULL),
+        ("Transform is not configured (mode=%s, option=%s).",
+            gst_tensor_transform_get_mode_string (filter->mode),
+            GST_STR_NULL (filter->option)));
+    return GST_FLOW_ERROR;
+  }
+
   inbuf = gst_tensor_buffer_from_config (inbuf, &filter->in_config);
 
   in_flexible =
@@ -2181,8 +2210,49 @@ gst_tensor_transform_transform_caps (GstBaseTransform * trans,
     intersection =
         gst_caps_intersect_full (result, filtercap, GST_CAPS_INTERSECT_FIRST);
 
+    if (gst_caps_is_empty (intersection)) {
+      gchar *str_from = gst_caps_to_string (caps);
+      gchar *str_to = gst_caps_to_string (result);
+      gchar *str_filter = gst_caps_to_string (filtercap);
+
+      GST_WARNING_OBJECT (filter,
+          "mode=%s option=%s cannot produce %s caps compatible with [%s]: "
+          "input [%s] is transformed into [%s].",
+          gst_tensor_transform_get_mode_string (filter->mode),
+          GST_STR_NULL (filter->option),
+          (direction == GST_PAD_SINK) ? "src" : "sink", str_filter, str_from,
+          str_to);
+
+      g_free (str_from);
+      g_free (str_to);
+      g_free (str_filter);
+    }
+
     gst_caps_unref (result);
     result = intersection;
+  } else if (direction == GST_PAD_SINK && gst_caps_is_fixed (caps)
+      && !gst_caps_is_empty (result)) {
+    GstCaps *peercaps = gst_pad_peer_query_caps (pad, NULL);
+
+    if (peercaps) {
+      if (!gst_caps_can_intersect (result, peercaps)) {
+        gchar *str_from = gst_caps_to_string (caps);
+        gchar *str_to = gst_caps_to_string (result);
+        gchar *str_peer = gst_caps_to_string (peercaps);
+
+        GST_WARNING_OBJECT (filter,
+            "mode=%s option=%s transforms input [%s] into [%s], "
+            "which the downstream element cannot accept: [%s].",
+            gst_tensor_transform_get_mode_string (filter->mode),
+            GST_STR_NULL (filter->option), str_from, str_to, str_peer);
+
+        g_free (str_from);
+        g_free (str_to);
+        g_free (str_peer);
+      }
+
+      gst_caps_unref (peercaps);
+    }
   }
 
   silent_debug_caps (filter, result, "to");
@@ -2230,12 +2300,23 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
   gboolean in_flexible, out_flexible;
   gboolean allowed = FALSE;
   guint i;
+  const gchar *mode_str;
 
   filter = GST_TENSOR_TRANSFORM_CAST (trans);
 
   silent_debug (filter, "Calling SetCaps\n");
   silent_debug_caps (filter, incaps, "incaps");
   silent_debug_caps (filter, outcaps, "outcaps");
+
+  mode_str = gst_tensor_transform_get_mode_string (filter->mode);
+
+  if (!filter->loaded) {
+    GST_ERROR_OBJECT (filter,
+        "Transform is not configured (mode=%s, option=%s). "
+        "Set valid 'mode' and 'option' properties before negotiating caps.",
+        mode_str, GST_STR_NULL (filter->option));
+    goto error;
+  }
 
   if (!gst_tensors_config_from_caps (&in_config, incaps, TRUE)) {
     GST_ERROR_OBJECT (filter, "Cannot read cap of incaps\n");
@@ -2266,7 +2347,9 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
       if (!gst_tensor_transform_convert_dimension (filter, GST_PAD_SINK,
               i, in_info, out_info)) {
         GST_ERROR_OBJECT (filter,
-            "Tensor info is not matched with given properties.");
+            "Cannot derive the output info of tensor %u from the input caps "
+            "with mode=%s option=%s.", i, mode_str,
+            GST_STR_NULL (filter->option));
         goto error;
       }
     }
@@ -2279,8 +2362,16 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
     if (!in_flexible)
       out_config = config;
   } else if (!gst_tensors_config_is_equal (&out_config, &config)) {
+    gchar *cmp;
+
+    cmp = gst_tensors_info_compare_to_string (&config.info, &out_config.info);
     GST_ERROR_OBJECT (filter,
-        "Tensor info is not matched with given properties.\n");
+        "The output caps does not match the result of mode=%s option=%s "
+        "applied to the input caps. Framerate expected %d/%d, given %d/%d. "
+        "Tensors expected (left) vs given (right):\n%s",
+        mode_str, GST_STR_NULL (filter->option), config.rate_n, config.rate_d,
+        out_config.rate_n, out_config.rate_d, GST_STR_NULL (cmp));
+    g_free (cmp);
     goto error;
   }
 
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -321,15 +321,12 @@ gst_tensor_transform_get_mode_string (tensor_transform_mode mode)
 {
   GEnumClass *klass;
   GEnumValue *val;
-  const gchar *nick = "unknown";
 
-  klass = g_type_class_ref (GST_TYPE_TENSOR_TRANSFORM_MODE);
-  val = g_enum_get_value (klass, mode);
-  if (val)
-    nick = val->value_nick;
-  g_type_class_unref (klass);
+  /* the class is kept alive by the "mode" property spec */
+  klass = (GEnumClass *) g_type_class_peek (GST_TYPE_TENSOR_TRANSFORM_MODE);
+  val = klass ? g_enum_get_value (klass, mode) : NULL;
 
-  return nick;
+  return val ? val->value_nick : "unknown";
 }
 
 /**
@@ -1862,7 +1859,7 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
   filter = GST_TENSOR_TRANSFORM_CAST (trans);
 
   if (!filter->loaded) {
-    GST_ELEMENT_ERROR (filter, STREAM, FAILED, (NULL),
+    GST_ELEMENT_ERROR (filter, CORE, NEGOTIATION, (NULL),
         ("Transform is not configured (mode=%s, option=%s).",
             gst_tensor_transform_get_mode_string (filter->mode),
             GST_STR_NULL (filter->option)));
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -305,7 +305,6 @@ gst_tensor_transform_init (GstTensorTransform * filter)
   filter->operators = NULL;
   filter->acceleration = DEFAULT_ACCELERATION;
   filter->apply = NULL;
-  filter->data_typecast.to = _NNS_END;
 
   gst_tensors_config_init (&filter->in_config);
   gst_tensors_config_init (&filter->out_config);
@@ -690,8 +689,10 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
   gchar *filter_name;
   gboolean ret = FALSE;
 
-  if (filter->mode == GTT_UNKNOWN || filter->option == NULL)
+  if (filter->mode == GTT_UNKNOWN || filter->option == NULL) {
+    filter->loaded = FALSE;
     return TRUE;
+  }
 
   filter_name = gst_object_get_name ((GstObject *) filter);
 
@@ -1053,6 +1054,9 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
       GST_ERROR_OBJECT (filter, "Cannot identify mode\n");
       ret = FALSE;
   }
+
+  if (!ret)
+    filter->loaded = FALSE;
 
   g_free (filter_name);
   return ret;
@@ -2180,7 +2184,7 @@ gst_tensor_transform_transform_caps (GstBaseTransform * trans,
     if (out_config.info.format == _NNS_TENSOR_FORMAT_END) {
       out_caps = gst_pad_get_pad_template_caps (pad);
     } else {
-      if (gst_tensors_config_is_static (&out_config)) {
+      if (filter->loaded && gst_tensors_config_is_static (&out_config)) {
         for (j = 0; j < in_config.info.num_tensors; j++) {
           in_info = gst_tensors_info_get_nth_info (&in_config.info, j);
           out_info = gst_tensors_info_get_nth_info (&out_config.info, j);
@@ -2228,7 +2232,9 @@ gst_tensor_transform_transform_caps (GstBaseTransform * trans,
     gst_caps_unref (result);
     result = intersection;
   } else if (direction == GST_PAD_SINK && gst_caps_is_fixed (caps)
-      && !gst_caps_is_empty (result)) {
+      && !gst_caps_is_empty (result)
+      && gst_debug_category_get_threshold (GST_CAT_DEFAULT) >=
+      GST_LEVEL_WARNING) {
     GstCaps *peercaps = gst_pad_peer_query_caps (pad, NULL);
 
     if (peercaps) {
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -13,6 +13,7 @@
 #include <gst/check/gstharness.h>
 #include <gst/check/gsttestclock.h>
 #include <gst/gst.h>
+#include <nnstreamer_conf.h>
 #include <nnstreamer_cppplugin_api_filter.hh>
 #include <nnstreamer_plugin_api_converter.h>
 #include <nnstreamer_plugin_api_decoder.h>
@@ -1179,12 +1180,16 @@ TEST (testTensorTransform, setCapsConfiguredAfterLink)
 /**
  * @brief Caps negotiation through a chain of tensor_transform and
  *        tensor_filter, both of which query their peers from transform_caps.
- *        Guards against caps-query recursion between the two (#4103).
+ *        Guards against caps-query recursion between the two. The
+ *        tensor_transform debug threshold is raised to WARNING because the
+ *        downstream peer query in transform_caps is now skipped unless that
+ *        threshold is at least WARNING, so raising it is required to
+ *        actually exercise the peer-query path (#4103).
  */
 TEST (testTensorTransform, negotiationChainWithFilter)
 {
   const gchar *root_path = g_getenv ("NNSTREAMER_SOURCE_ROOT_PATH");
-  gchar *model_file, *str_pipeline;
+  gchar *model_file, *str_pipeline, *lib_name;
   GstElement *pipeline;
   GstBus *bus;
   GstMessage *msg;
@@ -1192,9 +1197,14 @@ TEST (testTensorTransform, negotiationChainWithFilter)
   if (root_path == NULL)
     root_path = "..";
 
-  model_file = g_build_filename (root_path, "build", "tests", "nnstreamer_example",
-      "libnnstreamer_customfilter_passthrough_variable.so", NULL);
+  lib_name = g_strdup_printf ("libnnstreamer_customfilter_passthrough_variable%s",
+      NNSTREAMER_SO_FILE_EXTENSION);
+  model_file = g_build_filename (
+      root_path, "build", "tests", "nnstreamer_example", lib_name, NULL);
+  g_free (lib_name);
   ASSERT_TRUE (g_file_test (model_file, G_FILE_TEST_EXISTS));
+
+  gst_debug_set_threshold_for_name ("tensor_transform", GST_LEVEL_WARNING);
 
   str_pipeline = g_strdup_printf (
       "videotestsrc num-buffers=3 ! video/x-raw,format=RGB,width=8,height=8,framerate=30/1 ! "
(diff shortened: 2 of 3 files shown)


myungjoo:

Scope reviewed: the diff of this PR only (`gst/nnstreamer/elements/gsttensor_transform.c`, `tests/nnstreamer_plugins/unittest_plugins.cc`, `tests/transform_typecast/runTest.sh`), cross-checked against `gst/nnstreamer/tensor_filter/tensor_filter.c` (peer-caps-query precedent), `nnstreamer_plugin_api_impl.c`/`nnstreamer_plugin_api_util_impl.c` (caps/type helpers this PR calls into), `packaging/nnstreamer.spec` (CI gating), and issue #4103.

## Does it solve #4103?
Yes. The three previously-indistinguishable failure modes (`Tensor info is not matched with given properties.`) are now split into: (a) an up-front "not configured" rejection naming `mode`/`option` when `filter->loaded` is `FALSE`, (b) a per-tensor-index message when dimension/type derivation fails, and (c) an expected-vs-given `gst_tensors_info_compare_to_string` dump plus framerate values when the final config doesn't match. `transform_caps` also now warns with mode/option context on both the filtercap-intersection-empty case and the new downstream-incompatible-fixed-caps case. The CRITICAL-then-crash path from #4102 (mode without option passing negotiation and dying on the first buffer) is closed by the same `!filter->loaded` check in `set_caps`, with `gst_tensor_transform_transform()`'s leftover `g_return_val_if_fail` replaced by a proper `GST_ELEMENT_ERROR` as a defensive backstop for mid-stream reconfiguration. This directly matches the issue's ask ("throw it with more explanation... which part of the operator/options are not compatible").

## Correctness / memory / NULL handling
- `gst_tensor_transform_get_mode_string`: the `g_type_class_ref`/`g_enum_get_value`/`g_type_class_unref` idiom is the same one already used in `tensor_filter_common.c`; `value_nick` for a statically-registered enum stays valid after unref (the type is never actually unloaded), so this is safe and consistent with existing convention.
- `filter->option` is initialized to `NULL` in `gst_tensor_transform_init` and only ever `g_strdup`'d after that, so every new `GST_STR_NULL (filter->option)` call is necessary and correct (not dead defensiveness).
- Format strings were checked argument-by-argument: the 7-specifier message in the final `set_caps` error (`%s %s %d/%d %d/%d %s`) matches its 7 arguments in type and order; `rate_n`/`rate_d` are `gint`, consistent with `%d` usage elsewhere in the codebase for the same fields.
- `gst_tensors_info_compare_to_string` never returns `NULL` for non-NULL inputs (always `g_string_free(gstr, FALSE)`), so `GST_STR_NULL (cmp)` is harmless belt-and-suspenders, not masking a real NULL case.
- All new `gchar *`/`GstCaps *` temporaries (`str_from`/`str_to`/`str_filter`/`str_peer`, `peercaps`, `cmp`) are freed on every path, including the "no mismatch" path where the block is skipped entirely — no leaks found.
- `filter->data_typecast.to = _NNS_END;` added to `init()` is a real bug fix, not cosmetic: `_NNS_INT32 == 0`, so before this change an un-configured `typecast` mode (mode set, option never set) silently advertised `int32` output caps during `transform_caps`/linking purely because GObject zero-inits instance memory. `_NNS_END` is already the established "unset" sentinel used identically by `GTT_ARITHMETIC` (`data_arithmetic.out_type != _NNS_END`) and is already handled by `gst_tensor_pad_possible_caps_from_config` (`if (_info->type != _NNS_END)` before adding a `types` field), so this is a safe, well-precedented fix, appropriately scoped.

## Regression / negotiation-flow risk (as specifically requested)
**`set_caps` rejecting `!filter->loaded`:** this only changes behavior at the point a CAPS event is actually processed, which happens at first-buffer/negotiation time, not at `gst_element_link()` time (link only exercises `transform_caps`/caps-query, not `set_caps`). So an application that links elements first and configures `mode`/`option` afterward — before going to PLAYING — is unaffected; this is exercised by the new `setCapsConfiguredAfterLink` positive-control test. I did not find a code path where `set_caps` could run before an application has had a chance to set properties in the normal link-then-configure-then-PLAY sequence used by ml-api/nnstreamer-example.

**`transform_caps`'s new `gst_pad_peer_query_caps` call (SINK direction, fixed caps, no filtercap):** I traced this against `GstBaseTransform`'s caps-query flow. It is diagnostic-only — it never mutates `result`, only the log message — so it cannot change negotiation outcomes. On recursion: because the call only fires for `direction == GST_PAD_SINK` (querying strictly downstream) and the mirrored check in `tensor_filter.c` also only fires for the same direction, a two-hop bounce (e.g. `tensor_transform ! tensor_filter` or `tensor_transform ! tensor_transform`) lands the reflected call in the *other* direction on the peer's side, which does not re-trigger its own downstream query — so the chain terminates after one extra hop rather than growing unboundedly. This matches the reasoning given in the PR description for why the opposite (SRC/upstream) direction was deliberately left out.

That said, one path is worth flagging: when the *downstream* peer itself originates a `GST_QUERY_CAPS` on this element's src pad with a `NULL` filter, `GstBaseTransform`'s default query handling will call this `transform_caps(direction=SINK, ...)` to compute the answer, and — because `filtercap` is then `NULL` — the new code re-queries that same downstream peer's caps *from inside the computation of the answer to its own query* (a synchronous re-entrant call back into the querying element, on the same thread). I could not find anything in the existing GStreamer caps-query contract that forbids this (it's the same shape of call `tensor_filter.c` already makes unconditionally today), so I don't believe it's a new class of risk introduced by this PR — but there is currently no test in the repo (here or pre-existing) that exercises two adjacent caps-querying nnstreamer elements to positively demonstrate this doesn't hang. A real infinite loop here would show up as a CI timeout rather than a clean assertion failure, so it wouldn't be an obvious "regression caught by a red test" — it would just look like a stuck CI job. **Minor**: consider adding one SSAT/gst-launch pipeline test chaining two elements that both implement this downstream-peer-query pattern (e.g. `tensor_transform mode=typecast option=float32 ! tensor_filter ...` or `tensor_transform ! tensor_transform`) so any future change to either element's `transform_caps` that reintroduces unconditional/bidirectional peer-querying gets caught by a timeout-bounded SSAT run instead of surfacing as an unexplained CI hang later.

## Test robustness (`gst_harness_push` + `GST_FLOW_NOT_NEGOTIATED`, `gst_debug_add_log_function`)
- `gst_debug_add_log_function` installs a process-wide log handler; each test pairs it with `gst_debug_remove_log_function`/`gst_debug_unset_threshold_for_name` afterward. Good robustness detail: the flow-return checks use `EXPECT_EQ` (non-fatal) rather than `ASSERT_EQ`, so if the expected `GST_FLOW_NOT_NEGOTIATED` value ever changes, cleanup still runs and the log function/threshold don't leak into later tests in the same binary. The only remaining leak vector is a hard crash/timeout mid-test, which is an inherent GTest limitation, not something this PR introduces.
- The `strstr` assertions on collected log text are correctly gated behind `#ifndef GST_DISABLE_GST_DEBUG`, so a debug-stripped build won't spuriously fail — only the flow-return assertions (which don't depend on logging) run unconditionally in that configuration.
- GTest binaries in this repo run sequentially in-process (no evidence of intra-process parallel test execution), so the global debug-category threshold mutation is safe under the current test runner; this would need revisiting only if the test binary were ever run with real thread-level parallelism.

## CI enforcement
Confirmed via `packaging/nnstreamer.spec`: `unittest_plugins` (built under `tests/nnstreamer_plugins`) is run by `bash $test_script ./tests`, and SSAT (`ssat -n -p=1 ... -cn _n`) is run over `tests/`, both gated by `%unit_test`, which is what the Tizen GBS `unit_test` job and the daily/PR pipelines set. So `setCapsNotConfigured_n`, `setCapsModeWithoutOption_n`, `setCapsInvalidOption_n`, `transformCapsIncompatibleDownstream_n`, `setCapsConfiguredAfterLink`, and SSAT `transform_typecast` 25_n/26_n will all run in CI, and a future regression that reintroduces the pre-fix "accept caps then crash" behavior (e.g., someone dropping the `!filter->loaded` check) would fail `setCapsModeWithoutOption_n` and SSAT `26_n` and block merge.

## Scope / size / style
Changes are confined to `gsttensor_transform.c` and its own tests; no other module is touched (the one unrelated-looking hunk, `=(gsize)` → `= (gsize)`, is a pre-existing spacing nit the author explicitly calls out as needed for `gst-indent` to leave the file clean — acceptable). All new functions and test cases carry `@brief` Doxygen comments as required. No public header/API/ABI change, so no architecture documentation update is needed.

## Nits (non-blocking)
- `GST_ELEMENT_ERROR (filter, STREAM, FAILED, ...)` for the "not configured" runtime error uses a different domain/code than sibling elements' "not configured/negotiated" errors (`gsttensor_merge.c`/`gsttensor_mux.c` use `CORE, NEGOTIATION`). Not incorrect, just inconsistent with neighboring elements' convention.
- `gst_tensor_transform_get_mode_string` re-does `g_type_class_ref`/`g_type_class_unref` on every diagnostic call rather than caching the class once; negligible since it's only on negotiation/error paths, not per-buffer.

## Conclusion
No Critical or Major issues found. One Minor suggestion (a chained-elements SSAT test to give the new peer-query path a positive "doesn't hang" tripwire) and two non-blocking Nits. The fix is well-targeted, the reasoning about the peer-query direction is sound and matches existing precedent in `tensor_filter.c`, the `data_typecast.to` sentinel fix is a genuine and correctly-scoped bug fix, and the new tests are wired into the CI jobs that would catch a future regression of this exact bug. From this review, the PR looks mergeable as-is; the Minor test-coverage suggestion can reasonably be addressed in this PR or as a fast follow-up.

myungjoo:

Thanks for the review. All three items are addressed in 6bdbb55d:

- **Minor (chain test):** added `testTensorTransform.negotiationChainWithFilter`, which negotiates and streams `tensor_transform -> tensor_filter(custom) -> tensor_transform -> tensor_filter(custom) -> tensor_transform -> capsfilter`. It waits for EOS with a 10 s bus timeout, so a caps-query loop between the two elements would fail the test rather than hang the job. Passes locally (98/98 in `testTensorTransform.*`).
- **Nit (error category):** the unconfigured-element error is now `CORE, NEGOTIATION`, matching tensor_merge/tensor_mux.
- **Nit (GEnumClass ref/unref):** replaced with `g_type_class_peek`; the class is kept alive by the `mode` property spec.

myungjoo:

Second-pass review, scoped to commit 6bdbb55d (the response to the first review's Minor 1 / Nit 2). Checked against `gst/nnstreamer/elements/gsttensor_transform.c` (class_init/init/property registration), `gsttensor_merge.c`/`gsttensor_mux.c` (error-domain precedent), the custom-filter build (`tests/nnstreamer_example/meson.build`, `nnstreamer_customfilter_example_passthrough_variable.c`), other callers of the `NNSTREAMER_SOURCE_ROOT_PATH` convention in the same test file, `tests/unittest_util.c` (`setPipelineStateSync`), and GLib's `GParamSpecEnum`/`g_type_class_peek` contract.

## Nit 1 — error domain (`CORE, NEGOTIATION`)
Fixed exactly as suggested; now matches `gsttensor_merge.c`/`gsttensor_mux.c`. The other `!filter->loaded` check in `set_caps` (line ~2310) was correctly left as `GST_ERROR_OBJECT` + `goto error` (a soft caps-negotiation failure, not a bus error) — the domain change only applies to the `transform()` vfunc's fatal backstop, so the fix is precise, not a blanket find/replace.

## Nit 2 — `g_type_class_peek` instead of ref/unref
Verified this is actually safe, not just "probably fine":
- `gst_tensor_transform_class_init` installs the `"mode"` property via `g_param_spec_enum (..., GST_TYPE_TENSOR_TRANSFORM_MODE, ...)` (line ~244-247). `GParamSpecEnum` takes its own reference on the enum class in its instance-init and only drops it in the pspec's finalize, which for a statically-installed class property never runs during normal process lifetime.
- `class_init` for a `GType` always runs before any instance of that type can exist. `gst_tensor_transform_get_mode_string` is only ever called from instance methods (`transform`, `set_caps`) on an already-constructed `GstTensorTransform`, i.e. strictly after `class_init` (and therefore after the property spec ref) has already happened.
- So `g_type_class_peek` cannot observe an unregistered/unreffed class on any real call path here — the comment "the class is kept alive by the mode property spec" is accurate, and the `klass ? ... : NULL` guard is correct defensive style even though it should be unreachable in practice.

## Minor — chained negotiation test (`negotiationChainWithFilter`)
This is a solid response, and goes beyond the minimal ask (a 2-element chain) to a 5-element chain (`tensor_transform → tensor_filter → tensor_transform → tensor_filter → tensor_transform`), exercising both directions of the pattern (`tensor_transform`'s diagnostic-only downstream peer-query and `tensor_filter`'s unconditional one) across two hops.
- **Fixture path**: `g_build_filename (root_path, "build", "tests", "nnstreamer_example", "libnnstreamer_customfilter_passthrough_variable.so", ...)` matches the established convention used a few lines above in the same file (`setCapsConfiguredAfterLink`'s neighbor at line ~1186) and throughout `unittest_plugins.cc`. `nnstreamer_customfilter_passthrough_variable` is built unconditionally in `tests/nnstreamer_example/meson.build` (not gated by any option), and that same library is already relied on by `unittest_sink.cc`, so there's precedent this artifact reliably exists wherever `unittest_plugins` runs in CI (confirmed the `unit_test` gate in `packaging/nnstreamer.spec` is unchanged by this PR).
- **Pipeline correctness**: the custom filter (`set_inputDim` copies input dims/type through unchanged) is a true passthrough, so `float32 → (filter) → arithmetic add:1.0 [no typecast option, stays float32] → (filter) → typecast uint8` is internally consistent with the final `types=uint8` capsfilter — the test isn't accidentally relying on a caps mismatch being silently tolerated.
- **Fail-vs-hang**: `EXPECT_EQ` (non-fatal) is used for the `setPipelineStateSync` calls, and `ASSERT_TRUE (msg != nullptr)` bounds the wait via the 10 s `gst_bus_timed_pop_filtered` timeout, so a stuck negotiation reports a test failure within a bounded time rather than hanging the job — consistent with the PR's stated intent. One caveat worth noting for completeness (not a blocker): in the specific case of a real recursive caps-query loop, GStreamer's synchronous call chain would almost certainly stack-overflow/crash the streaming thread (and thus the whole test binary) well before either timeout is reached, which still "fails loudly" but via a crash rather than a clean assertion failure — the same caveat already applied to the general class of caps-query-loop bugs discussed in the first review, not something introduced by this test.
- **Cleanup**: `bus`, `msg`, and `pipeline` are unreffed on the success path; on the `ASSERT_TRUE` failure path resources are left unreleased, which matches the pre-existing convention for fatal assertions elsewhere in this same file (e.g. the neighboring `ASSERT_TRUE (NULL != h)` patterns), not a new gap.
- Doxygen `@brief` present, references `#4103`, formatting verified byte-identical to `clang-format -style=file` output — no style nits.

## Scope check
`git diff` between the two review rounds touches only the same 3 files as the original PR (`gsttensor_transform.c`, `tests/nnstreamer_plugins/unittest_plugins.cc`, `tests/transform_typecast/runTest.sh`) — no scope creep in the follow-up commit.

## Conclusion
All three first-round items (1 Minor, 2 Nits) are resolved correctly and verified independently, not just taken on faith. No new Critical/Major/Minor/Nit issues found in 6bdbb55d. The PR looks mergeable as-is.

myungjoo:

Thanks, all three items and the notes are addressed in the new commit.

- **1 (peer query cost):** the downstream peer query in `transform_caps` is now gated on `gst_debug_category_get_threshold (GST_CAT_DEFAULT) >= GST_LEVEL_WARNING`, exactly as suggested. `negotiationChainWithFilter` raises the threshold so the query path is still exercised by CI.
- **2 (union side effect):** taken as proposed. `transform_caps` skips the per-mode conversion while `!filter->loaded` (identity caps), and the `data_typecast.to = _NNS_END` init write is gone. This was more than a cosmetic problem: an unconfigured `padding` with the real caps downstream advertised a padded dimension, failed negotiation before `set_caps`, and printed the "cannot accept" warning instead of "not configured". `setCapsUnconfiguredPaddingWithDownstreamCaps_n` guards it. While there, `loaded` is now cleared when a re-parse fails (e.g. `mode` switched to one under which the stored option is invalid; previously it stayed TRUE with stale union data), guarded by `setCapsAfterModeChangeUnparsable_n`. A successful re-parse never passes through the unconfigured state; `optionChangeWhileStreaming` checks runtime option changes, same-mode re-set, and an invalid runtime option being rejected with the previous one kept.
- **3 (filter-caps branch untested):** `transformCapsFilterIntersectionEmpty_n` queries both pads with unsatisfiable filters and asserts the direction-specific messages.
- Notes: `.so` replaced with `NNSTREAMER_SO_FILE_EXTENSION`; PR body updated (`CORE, NEGOTIATION`, the chain test, and the remark that SSAT 25_n/26_n also fail on main and are not the regression guard). Added `linkThenConfigure`, which links with `gst_element_link_many` before setting any property and streams to EOS, as the guard for the link-first application pattern.

myungjoo:

Third-pass review, scoped to commit `7bd67d9e` (the response to `myungjoo-bot`'s three follow-ups + four notes from the previous round), cross-checked against the surrounding `gst/nnstreamer/elements/gsttensor_transform.c` (all three call sites of `gst_tensor_transform_convert_dimension`, `set_option_data`'s per-mode switch, the `PROP_MODE`/`PROP_OPTION` setters), the new/updated tests in `tests/nnstreamer_plugins/unittest_plugins.cc`, `gst/nnstreamer/nnstreamer_conf.h` (`NNSTREAMER_SO_FILE_EXTENSION`), and GstBaseTransform's caps-query-forwarding contract. No build environment was available, so this is a static/read-only review (diff + `git show` + source tracing), not a compiled/executed one.

## Follow-up items — verified resolved

**1. Peer-query cost.** The downstream peer query in `transform_caps` is now gated on `gst_debug_category_get_threshold (GST_CAT_DEFAULT) >= GST_LEVEL_WARNING`. `GST_CAT_DEFAULT` is `#define`d to `gst_tensor_transform_debug` (registered under the name `"tensor_transform"`), which is exactly the category `negotiationChainWithFilter` raises via `gst_debug_set_threshold_for_name ("tensor_transform", GST_LEVEL_WARNING)` before running the pipeline — so the chain test genuinely exercises the gated path rather than trivially skipping it. Resolved correctly.

**2. Union side effect.** `filter->data_typecast.to = _NNS_END` is removed from `_init()`, and all three call sites of `gst_tensor_transform_convert_dimension` are now safe while unconfigured: the one in `transform_caps` is gated by the new `filter->loaded &&` check, and the ones in `transform()`/`set_caps()` are unreachable while `!filter->loaded` because both functions bail out on the existing `!filter->loaded` checks before reaching them. This is a mode-agnostic fix, not just a `padding`-specific patch, and `setCapsUnconfiguredPaddingWithDownstreamCaps_n` is a well-targeted regression test for exactly the scenario the bot described (unconfigured `padding` + real, untransformed downstream caps): traced the harness's caps through `transform_caps`, and identity-caps output now matches the downstream `dimensions=5:1:1:1,types=uint8`, so negotiation proceeds to `set_caps`, where the expected "not configured" message fires. Resolved correctly.

**3. Untested filter-caps-intersection-empty branch.** `transformCapsFilterIntersectionEmpty_n` covers both directions via `gst_pad_query_caps` on the src and sink pads with unsatisfiable filters. I traced GstBaseTransform's caps-query forwarding (a query arriving on pad X is answered by calling `transform_caps` with `direction` set to the *other* pad's direction, since the vfunc's contract is "transform caps that exist in `direction` into the other side's caps"): a query on the **src** pad forwards to `transform_caps(direction=GST_PAD_SINK, ...)`, whose label ternary `(direction == GST_PAD_SINK) ? "src" : "sink"` produces `"cannot produce src caps"` — matching the test's assertion for the src-pad query — and symmetrically a query on the **sink** pad produces `"cannot produce sink caps"`, matching the other assertion. Resolved correctly (modulo the general caveat that I couldn't execute the suite to confirm at runtime — see note below).

**loaded semantics (the PR's stated core change).** Traced `gst_tensor_transform_set_option_data()`: every mode's success path still does `ret = filter->loaded = TRUE;` only after validation passes (this didn't change), the early-return case (`mode == GTT_UNKNOWN || option == NULL`) now sets `loaded = FALSE`, and a new `if (!ret) filter->loaded = FALSE;` at the bottom clears it on any parse failure, including a *re*-parse failure after a mode switch (`setCapsAfterModeChangeUnparsable_n` covers this correctly). Importantly, a **successful** re-parse never clears `loaded` first — it's only ever set to `FALSE` on the failing path — so `optionChangeWhileStreaming`'s claim ("a successful re-parse never passes through the unconfigured state") holds.

## Notes (non-blocking, but worth recording)

**Note A — narrow new race window in the `PROP_OPTION` revert path (Minor).** `PROP_OPTION`'s setter still uses its pre-existing backup/revert pattern: on an invalid new option it frees the new string, restores `backup_option`, and calls `set_option_data()` again. What's new in this commit is that the *first* (failing) call now sets `filter->loaded = FALSE` before the setter's revert branch calls `set_option_data()` a second time to restore it to `TRUE`. Since property mutation here isn't synchronized with the streaming thread (no lock around `loaded`/`operators`/the option union — already flagged as a pre-existing characteristic in the first review round), a `transform()`/`set_caps()` call landing in that brief window will now hit the new `!filter->loaded` gate and post a hard `GST_ELEMENT_ERROR`, which most apps treat as fatal — whereas before this round, the same (already-unsupported) race could at worst have produced a transient no-op/passthrough on a freed-then-empty `operators` list, not a bus error. This only manifests under concurrent, unsynchronized property mutation during streaming, which isn't a documented/supported usage pattern for this element, so I'm not blocking on it, but since the task specifically asked about it: yes, the window is new, though it does not widen the *existing* general lack of locking, it just changes what a hit in that window produces (a loud failure vs. silently wrong data) — arguably in the spirit of issue #4103, ironically for a case that used to be recoverable.

**Note B — `gst_debug_category_get_threshold` not guarded for `GST_DISABLE_GST_DEBUG` (Minor, low confidence / unverified).** The tests' own `strstr` assertions on collected log text are correctly wrapped in `#ifndef GST_DISABLE_GST_DEBUG`, showing the PR already treats that flag as a real possibility for this codebase. The new production-code call, `gst_debug_category_get_threshold (GST_CAT_DEFAULT) >= GST_LEVEL_WARNING` in `transform_caps` (`gsttensor_transform.c:2236`), has no equivalent guard. From general GStreamer API knowledge, the debug-category introspection functions (`gst_debug_category_get_threshold`, `GST_DEBUG_CATEGORY_INIT`, etc.) are typically excluded from the header when `GST_DISABLE_GST_DEBUG` is defined for the GStreamer build being compiled against, which would break compilation of this file in that configuration. I could not verify this against actual GStreamer headers in this environment (no build tools / vendored GStreamer available), and a repo-wide grep found no other reference to `GST_DISABLE_GST_DEBUG` anywhere outside this PR's own test file, so nnstreamer's own CI/build matrix apparently never exercises this flag today — meaning the practical risk to this PR's mergeability is low. Flagging for the author to confirm (e.g., a quick `-DGST_DISABLE_GST_DEBUG` compile check) rather than as a confirmed defect.

**Note C — `linkThenConfigure`'s doc-comment overstates what it guards (Minor/Nit).** The comment says it guards "the app pattern from issue #4102" and that "linking must succeed while the element is still unconfigured." The link succeeding is true but was never actually at risk: `gst_element_link_many` between these generic `other/tensors`-family pad templates only performs template-level compatibility checks, not the `transform_caps` computation that the identity-caps regression (follow-up #2 above) actually affects. Because the test calls `g_object_set (transform, "mode", ..., "option", ..., NULL)` *before* `setPipelineStateSync (pipeline, GST_STATE_PLAYING, ...)`, by the time real caps negotiation runs (during the PAUSED→PLAYING transition, when `transform_caps`/`set_caps` actually execute), the element is already `loaded`. So this test never actually calls `transform_caps` while `!filter->loaded` with concrete/fixed downstream caps — that specific scenario is correctly covered instead by `setCapsUnconfiguredPaddingWithDownstreamCaps_n` (via `GstHarness`, which forces exactly that ordering). `linkThenConfigure` remains a legitimate positive end-to-end smoke test for the link-then-configure application pattern in general, just not the regression guard its comment implies for the identity-caps bug specifically.

## Style / scope / hygiene

- All added/changed lines in `gsttensor_transform.c` stay within 80 columns.
- `tests/nnstreamer_plugins/unittest_plugins.cc` is fully `clang-format -style=file` clean (0 replacements against the current file), consistent with CLAUDE.md's C++ convention.
- Every new test carries a `/** @brief */` Doxygen block; all new inline comments are single-line `/* ... */` (no multi-line non-Doxygen blocks that would trip the CI doxygen check).
- `git show --stat 7bd67d9e` confirms the diff is confined to `gsttensor_transform.c` and its own test file — no scope creep into other modules.
- The PR body has been updated and is now consistent with `7bd67d9e`: it correctly states `CORE, NEGOTIATION`, lists all five new tests plus the updated `negotiationChainWithFilter`, documents the `NNSTREAMER_SO_FILE_EXTENSION` switch, and states the SSAT `25_n`/`26_n` caveat from the previous round's notes. This resolves the "PR body is stale" note from round 2's bot review.
- No public API/ABI or architecture-document change; `gsttensor_transform.md` still accurately documents that every mode requires an option, so no doc update is owed here either.

## Conclusion

All three follow-up items and the four notes from the `myungjoo-bot` review are resolved as intended, and I independently traced the fixes (not just the diff) rather than taking the response comment at face value. I found no Critical or Major issues in `7bd67d9e`. Two Minor observations worth a maintainer's attention (Note A: the new brief `loaded`-clearing window in the `PROP_OPTION` revert path turns an existing, already-unsupported concurrent-mutation race into a hard bus error instead of silently wrong data; Note B: the new `gst_debug_category_get_threshold` call is unguarded for `GST_DISABLE_GST_DEBUG`, unverified in this environment and apparently not exercised by nnstreamer's current CI) and one Minor/Nit on test-comment accuracy (Note C). None of these block merge in my assessment — from this review, the PR looks mergeable as-is; Notes A and B could reasonably be addressed here or as a fast follow-up at the maintainer's discretion.

myungjoo:

Thanks. On the three Minor notes, no code change is proposed; here is why, with what was verified locally.

- **`loaded` window on the `PROP_OPTION` failure-then-revert path.** Agreed that it exists, and that it only opens when an *invalid* option is written while streaming. That pattern was already unsafe before this PR: the arithmetic parser frees and rebuilds `filter->operators` without any lock while `transform()` may be walking it, so a concurrent mutation could already read freed memory. A hard `GST_ELEMENT_ERROR` in that window is the safer of the two outcomes. Closing it properly means parsing into a scratch copy of the option union and swapping under a lock, which is a separate change; not doing it here.
- **`gst_debug_category_get_threshold` under `GST_DISABLE_GST_DEBUG`.** Verified against the installed `gstinfo.h`: with that flag the call is a macro expanding to `(GST_LEVEL_NONE)` and `GST_DEBUG_CATEGORY_STATIC` becomes a dummy declaration, so the expression compiles and the branch is simply never taken. A probe TU defining `GST_DISABLE_GST_DEBUG` and using exactly this expression compiles cleanly with `-Wall`. No guard needed, which is also why no other caller in GStreamer plugins wraps it.
- **`linkThenConfigure` coverage.** `gst_element_link_many` uses `GST_PAD_LINK_CHECK_DEFAULT`, which includes the CAPS check, so linking queries the pads' caps. Confirmed with `GST_DEBUG=GST_CAPS:5,GST_STATES:4`: before any state change, `gst_pad_query_caps` runs on `tensortransform0:sink` (and then `:src` for the next link), which goes through `GstBaseTransform::query_caps` into `transform_caps` while the element is still unconfigured. So the test does exercise the unconfigured `transform_caps` path at link time; the fixed-downstream-caps interaction *during negotiation* is the job of `setCapsUnconfiguredPaddingWithDownstreamCaps_n`, as you say. The comment describes link-time behaviour and is accurate as written.

PR #4891, opened 2026-08-27
[Filter/llamacpp] Fix the sync-mode test stall and treat a zero-token generation as valid

myungjoo:
Fixes #4883, though not by the mechanism the issue proposed — that hypothesis turned out to be wrong, and the real cause is in the test's own pipeline. Also fixes a separate filter defect found on the way.

## The actual cause of #4883

Pad probes through the failing pipeline show both buffers reaching `tensor_filter`, both leaving it, and both arriving at the appsink's sink pad — while the `new-sample` callback fires only once:

```
DBG filter_sink=2 filter_src=2 appsink_pad=2 cb=1 err=0
```

The filter is not involved. The second buffer carries a PTS about nineteen minutes in the future:

```
[PTS sink pts_ms=0.0        run_ms=<none>]     <- buffer 1, rendered
[PTS sink pts_ms=1151788.8  run_ms=1.4]        <- buffer 2, held until its PTS
```

appsink synchronises to the clock by default, so it holds buffer 2 until that presentation time. Nothing errors, nothing is dropped — the sample just never renders and the test waits out its 10 s ceiling. That is exactly the reported symptom, and it is why no `invoke-failure` or bus error ever appeared.

The future PTS comes from `_gst_tensor_converter_chain_timestamp()` ([gsttensor_converter.c#L824](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/elements/gsttensor_converter.c#L824)). For a buffer with no PTS and no framerate in caps it computes:

```c
base = gst_element_get_base_time (GST_ELEMENT (self));
now  = gst_clock_get_time (clock);
pts  = (base < now) ? (now - base) : 0;
```

While the element's base time is still unset, `now - base` is the **absolute** clock time, not a running time. **That converter behaviour is a real defect in its own right and this PR deliberately does not touch it** — `tensor_converter` is in essentially every nnstreamer pipeline and that change deserves its own review. See the note at the end.

The fix here is the one that belongs in a test: these tests measure token generation, not real-time playback, so the sink has no business synchronising. `sync=false` is what the rest of this repo's test suite already does for its sinks.

### Reproduction and verification

The race is decided by whether buffer 2 is timestamped inside the window where base time is unset, which depends on how fast the first generation completes. Restricting the process to four cores makes it highly reproducible — note it is *not* monotonic in available CPU, which is what shows it to be a race rather than slowness:

| cores | `multipleInputsSingleOutputSync_p` failures |
|---|---|
| 1 | 4/20 |
[...]

PR #4918, opened 2026-09-02
[Converter] Take the running time only while PLAYING

### LINKED ISSUE #4898: [Converter] tensor_converter stamps an absolute clock time as PTS when base time is not yet set
## Summary

`_gst_tensor_converter_chain_timestamp()` can assign a buffer a PTS equal to the **absolute** clock time instead of a running time. Downstream sinks that synchronise to the clock then hold that buffer until a presentation time minutes or hours in the future. Nothing errors and nothing is dropped — the buffer simply never renders.

This was found while diagnosing #4883, where it stalls `NNStreamerFilterLlamaCppTest.multipleInputsSingleOutputSync_p`. [#4891](https://github.com/nnstreamer/nnstreamer/pull/4891) works around it on the test side (`sync=false` on the appsink, which is correct for that test regardless) and deliberately leaves this element alone, because `tensor_converter` sits in essentially every nnstreamer pipeline and the fix deserves its own review.

## The code

[gst/nnstreamer/elements/gsttensor_converter.c#L818-L835](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/elements/gsttensor_converter.c#L818-L835):

```c
      } else {
        GstClock *clock;

        clock = gst_element_get_clock (GST_ELEMENT (self));

        if (clock) {
          GstClockTime now, base;

          base = gst_element_get_base_time (GST_ELEMENT (self));
          now = gst_clock_get_time (clock);

          pts = (base < now) ? (now - base) : 0;
          gst_object_unref (clock);
        }
      }
```

`gst_element_get_base_time()` returns 0 until the pipeline distributes the base time during the PAUSED→PLAYING transition. There is a window in which the element already has a clock but its base time is still 0. In that window `base < now` holds, so `pts = now - 0 = now` — the raw clock value.

## Conditions

All of these are ordinary, and the first is the default:

1. `set-timestamp` is TRUE (`DEFAULT_SET_TIMESTAMP`).
2. Incoming buffers carry no PTS (e.g. `appsrc` with `do-timestamp=false`).
3. No framerate in the negotiated caps, so the `have_framerate` branch is skipped and the clock branch is taken.
4. A buffer is processed while the element's base time is still unset.

## Evidence

Pad probes on the #4883 pipeline (`appsrc ! tensor_converter ! tensor_filter ! tensor_decoder ! appsink`), values in ms:

```
[PTS sink pts_ms=0.0        run_ms=<invalid>]   <- buffer 1, prerolling, no clock yet -> segment.start
[PTS sink pts_ms=1151788.8  run_ms=1.4]         <- buffer 2, PTS = absolute clock (~19 min)
```

Buffer 2 reaches the appsink's sink pad and is counted there, but `new-sample` never fires for it: the sink is waiting for a presentation time ~19 minutes out. `1151788 ms` matches system uptime, not any running time.

## Reproduction

Whether a buffer lands in the bad window depends on timing, so this is intermittent. Restricting the process to a specific core count makes it reliable — note the failure rate is **not** monotonic in available CPU, which is what identifies it as a race rather than slowness:

| cores | `multipleInputsSingleOutputSync_p` failures |
|---|---|
| 1 | 4/20 |
| 2 | 0/20 |
| 4 | 16/20 |

(llama.cpp `b6569`, `TinyStories-656K-Q2_K.gguf`, `taskset -c 0-3`.)

## Suggested fix

Do not use the clock branch while the base time is unset. When `base` is 0, fall back to what the no-clock path already does (`self->segment.start`, or `old_timestamp + duration` where a duration is known) rather than emitting an absolute timestamp.

## Test coverage

No test in the tree currently exercises this branch. A regression test would need a pipeline whose source emits untimestamped buffers with no framerate in caps, feeding a **synchronising** sink, asserting that the observed PTS is a plausible running time rather than an absolute clock value. Worth adding alongside the fix — #4891 removes the one incidental tripwire that existed, so without such a test a regression here would go unnoticed.

---

*Found and characterised by a Claude agent while investigating #4883; the code reference, the measurements, and the reproduction above are verifiable.*

myungjoo:
Fixes #4898. Root cause of the intermittent stall in #4883 (the llamacpp sync-mode test); #4891 only works around it on the test side.

## Problem

`_gst_tensor_converter_chain_timestamp()` stamps an untimestamped frame with `clock time - base time` when the caps carry no framerate. GStreamer distributes the clock to every element at once at the start of PAUSED→PLAYING (`gst_pipeline_change_state` → `gst_element_set_clock` on the bin), but each element receives its base time only inside `gst_bin_element_set_state`, right before its own transition, sinks first. A frame that passes the converter after the sink released its preroll but before the converter itself is switched sees a clock and `base_time == 0`, so its PTS becomes the absolute clock value (system uptime). A synchronizing sink then waits minutes or hours for it; nothing errors, nothing is dropped.

Reproduced deterministically without llama.cpp: `fakesrc ! application/octet-stream ! tensor_converter ! identity ×30 ! fakesink sync=true` under `taskset -c 0-3` stamps buffer 2 with `0:26:42` (WSL uptime) 10/10 with `sync=false`, and stalls 8/10 with `sync=true`.

## Fix

Read clock, base time and state under one `GST_OBJECT_LOCK` and use the clock branch only while PLAYING; the state is committed under that lock after the base time is set, so PLAYING implies a delivered base time. Otherwise reuse the previous timestamp (segment start for the very first frame). Behaviour change to be aware of: frames processed while PAUSED after a pause now keep the last timestamp instead of `now - base_time`; the old value was wrong there too (base time is re-derived on resume, so those frames landed in the future by the paused duration).

`gst_element_get_current_running_time()` does not help: it treats `base_time == 0` as valid.

## Tests

- **New `unittest_converter_timestamp`** (GstHarness + GstTestClock, deterministic, no dependency on the race):
  - running time while PLAYING, clamp when clock < base time
  - the window itself: clock present, base time 0, PAUSED → segment start / previous PTS, then PLAYING → running time
  - pause/resume keeps the last PTS and continues monotonically
  - non-zero segment start fallback, valid input PTS/duration untouched, `set-timestamp=false` (negative), no clock
  - framerate branches (video GRAY8, octet with framerate) never consult the clock
  - flush and READY reset the history, BYTES segment converted to TIME, `frames-per-tensor` aggregation keeps the first frame's PTS
  - pipeline with `gst_pipeline_use_clock(GstTestClock)`: exact running times across PLAYING→PAUSED→PLAYING via appsink preroll/sample pulls
  - pipeline in PAUSED with a system clock injected into the converter (`base_time == 0`): PTS stays 0
  - `sync=true` end-to-end runs with 40 identity elements of padding: plain, with `queue`, with `set-timestamp=false`, live video without framerate, `tensor_mux` with two untimestamped branches, `tensor_aggregator`; negative: octet stream without dimensions errors out
- **SSAT** (`tests/nnstreamer_converter/runTest.sh`, cases 9-1 … 9-10): the same end-to-end pipelines as gst-launch runs, so the pdebuild, GBS x86_64 and RISC-V jobs cover them. The positive cases (9-1 … 9-6, 9-9, 9-10) run in gstTest timeout mode so a stall is scored as a failure; 9-5 and 9-6 are live-source smoke cases that cannot reach the window. The negative cases `9-7_n`/`9-8_n` fail at once and run without a timeout, so a stall there could never count as the expected failure.
- Existing converter golden tests unchanged.

Negative control (local WSL, converter change reverted, everything else identical): 10 of the 25 gtest cases fail (the live-source and framerate cases are smoke coverage of unchanged paths and pass either way), including `syncSinkCompletes_p` which times out on all three runs; the harness values equal the test-clock time. The identity padding was chosen from measurements on the unfixed converter with plain `gst-launch` and no CPU pinning: 20 identities stall 5/10 runs, 40 stall 10/10, 80 stall 10/10; with the fix 80 identities stall 0/10. Local run with the fix: 25/25 gtest, converter SSAT 42/42, plus unittest_sink 120/120, unittest_plugins 138/138, and the mux/merge/rate SSAT suites unchanged.

## Follow-up from review: basepad tolerance wrap (second commit)

The review pointed out that `gst_tensor_time_sync_buffer_from_collectpad()` (shared by `tensor_mux`/`tensor_merge`, `sync-mode=basepad`) computes its tolerance as `|PTS(new) - PTS(prev)| - 1` on the base pad, which is -1 for two equal timestamps and wraps to `G_MAXUINT64` in the unsigned `base_time`, defeating the tolerance check for that cycle. Equal consecutive timestamps were already ordinary before this PR (every frame stamped before the pipeline has a clock carries the segment start), and the PAUSED fallback here adds another source, so the clamp is fixed in-PR rather than deferred: the tolerance is now clamped at 0. Two pipeline tests (`muxBasepadDuplicateTimestamps_p`, `mergeBasepadDuplicateTimestamps_p`) push two equal timestamps on the base pad and a far frame on the other pad; without the clamp the far frame gets paired and one output disappears (verified locally against the unfixed core).

## Docs

`gst/nnstreamer/elements/gsttensor_converter.md` gains a Timestamps section; the mux/merge synchronization document is updated accordingly.

### DIFF
--- a/Documentation/synchronization-policies-at-mux-merge.md
+++ b/Documentation/synchronization-policies-at-mux-merge.md
@@ -5,7 +5,7 @@ title: Synchronization policies
 # Synchronization policies at Mux and Merge
 There are multiple synchronization policies for tensor_mux and tensor_merge.
 They are based on PTS ( presentation timestamp in nanoseconds (as a GstClockTime) ) and assume that every tensor buffer has PTS and can be accessed by GST_BUFFER_PTS(buf).  
-However, there is stream which does not have PTS such as application/octet-stream. In such cases, tensor_converter generates timestamp and set PTS. If framerate is given, tensor_converter generates proper PTS according to framerate and for the case without framerate, PTS is decided with running time which is calculated as absolute-time - base-time.  
+However, there is stream which does not have PTS such as application/octet-stream. In such cases, tensor_converter generates timestamp and set PTS. If framerate is given, tensor_converter generates proper PTS according to framerate and for the case without framerate, PTS is decided with running time (absolute-time - base-time) while the element is PLAYING; before that, the previous PTS (or the segment start) is reused.  
 Currently, three synchronization policies are implemented.
 
 # No synchronization
--- a/gst/nnstreamer/elements/gsttensor_converter.c
+++ b/gst/nnstreamer/elements/gsttensor_converter.c
@@ -819,18 +819,24 @@ _gst_tensor_converter_chain_timestamp (GstTensorConverter * self,
           pts = self->old_timestamp + duration;
         }
       } else {
-        GstClock *clock;
+        GstClock *clock = NULL;
+        GstClockTime base = GST_CLOCK_TIME_NONE;
 
-        clock = gst_element_get_clock (GST_ELEMENT (self));
+        /* base_time is delivered right before PLAYING is committed */
+        GST_OBJECT_LOCK (self);
+        if (GST_STATE (self) == GST_STATE_PLAYING && GST_ELEMENT_CLOCK (self)) {
+          clock = gst_object_ref (GST_ELEMENT_CLOCK (self));
+          base = GST_ELEMENT_CAST (self)->base_time;
+        }
+        GST_OBJECT_UNLOCK (self);
 
         if (clock) {
-          GstClockTime now, base;
-
-          base = gst_element_get_base_time (GST_ELEMENT (self));
-          now = gst_clock_get_time (clock);
+          GstClockTime now = gst_clock_get_time (clock);
 
           pts = (base < now) ? (now - base) : 0;
           gst_object_unref (clock);
+        } else if (GST_CLOCK_TIME_IS_VALID (self->old_timestamp)) {
+          pts = self->old_timestamp;
         }
       }
 
--- a/gst/nnstreamer/elements/gsttensor_converter.md
+++ b/gst/nnstreamer/elements/gsttensor_converter.md
@@ -75,6 +75,16 @@ When incoming media type is video, audio, or text, each frame (or a set of frame
 - Text
   - TBD.
 
+## Timestamps
+
+With ```set-timestamp=true``` (the default), a frame that arrives without a timestamp gets one:
+
+- If the input caps carry a framerate, the timestamp is the previous timestamp plus one frame duration, starting at the segment start.
+- Otherwise, while the element is PLAYING, the timestamp is the pipeline running time (clock time minus base time).
+- Otherwise (PAUSED, or no clock), the previous timestamp is reused; the very first frame gets the segment start.
+
+The running time is only taken while PLAYING because the base time reaches the element right before that state is committed. Using the clock earlier would stamp the absolute clock time and stall a synchronizing sink (see issue #4898).
+
 ## Properties
 
 - frames-per-tensor: The number of incoming media frames that will be contained in a single instance of tensors. With the value > 1, you can put multiple frames in a single tensor.
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -369,11 +369,14 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
 
     buf = gst_collect_pads_peek (collect, data);
     if (buf != NULL) {
-      if (pad->buffer != NULL)
-        base_time =
-            MIN ((GstClockTimeDiff) sync->data_basepad.duration,
-            ABS (GST_CLOCK_DIFF (GST_BUFFER_PTS (buf),
-                    GST_BUFFER_PTS (pad->buffer))) - 1);
+      if (pad->buffer != NULL) {
+        GstClockTimeDiff gap = ABS (GST_CLOCK_DIFF (GST_BUFFER_PTS (buf),
+                GST_BUFFER_PTS (pad->buffer)));
+
+        /* equal timestamps must not wrap the tolerance to G_MAXUINT64 */
+        base_time = MIN ((GstClockTimeDiff) sync->data_basepad.duration,
+            (gap > 0) ? (gap - 1) : 0);
+      }
       gst_buffer_unref (buf);
     }
   }
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -230,6 +230,15 @@ if gtest_dep.found()
       test('unittest_converter', unittest_converter, env: testenv)
     endif
 
+    unittest_converter_timestamp = executable('unittest_converter_timestamp',
+      join_paths('nnstreamer_converter', 'unittest_converter_timestamp.cc'),
+      dependencies: [nnstreamer_unittest_deps],
+      install: get_option('install-test'),
+      install_dir: unittest_install_dir
+    )
+
+    test('unittest_converter_timestamp', unittest_converter_timestamp, timeout: 120, env: testenv)
+
     unittest_filter_custom = executable('unittest_filter_custom',
       join_paths('nnstreamer_filter_custom', 'unittest_filter_custom.cc'),
       dependencies: [nnstreamer_unittest_deps],
--- a/tests/nnstreamer_converter/runTest.sh
+++ b/tests/nnstreamer_converter/runTest.sh
@@ -92,9 +92,12 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! videos
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=\"test_00.dat\" blocksize=-1 ! application/octet-stream ! tensor_converter input-dim=4:4:4:4:4 input-type=float32 ! fakesink" 8-1 0 0 $PERFORMANCE
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=\"test_01.dat\" blocksize=-1 ! application/octet-stream ! tensor_converter input-dim=3:10:10:4:5:6 input-type=uint8 ! fakesink" 8-2 0 0 $PERFORMANCE
 
-# Timestamp test cases (issue #4898): untimestamped frames without a framerate
-# must get a running time, so a synchronizing sink renders them and the
-# pipeline reaches EOS. A stall here is reported by the timeout.
+# Timestamp test cases (issue #4898). 9-1 to 9-4, 9-9 and 9-10 feed untimestamped
+# frames without a framerate through a widened state-change window into a
+# synchronizing sink; a stall is reported by the timeout. 9-5 and 9-6 are smoke
+# cases for live sources, which stamp their own frames and stream only after
+# their own transition, so they cannot hit the window. 9-7_n and 9-8_n fail at
+# once and run without a timeout so a stall there could never count as a pass.
 TS_PADDING=""
 for i in $(seq 1 40); do
     TS_PADDING="${TS_PADDING}identity ! "
@@ -108,8 +111,8 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! tee name=t
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} set-timestamp=false ! ${TS_PADDING}fakesink sync=true" 9-4 0 0 $PERFORMANCE 30
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc is-live=true num-buffers=5 ! video/x-raw,format=GRAY8,width=16,height=16,framerate=0/1 ! tensor_converter ! ${TS_PADDING}fakesink sync=true" 9-5 0 0 $PERFORMANCE 30
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} audiotestsrc is-live=true num-buffers=5 samplesperbuffer=800 ! audio/x-raw,format=S16LE,rate=8000,channels=1 ! tensor_converter frames-per-tensor=800 ! ${TS_PADDING}fakesink sync=true" 9-6 0 0 $PERFORMANCE 30
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} set-timestamp=notabool ! fakesink sync=true" 9-7_n 0 1 $PERFORMANCE 30
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! tensor_converter ! fakesink sync=true" 9-8_n 0 1 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} set-timestamp=notabool ! fakesink sync=true" 9-7_n 0 1 $PERFORMANCE
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! tensor_converter ! fakesink sync=true" 9-8_n 0 1 $PERFORMANCE
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! mux.sink_0 ${TS_SRC} ! ${TS_CONV} ! mux.sink_1 tensor_mux name=mux ! ${TS_PADDING}fakesink sync=true" 9-9 0 0 $PERFORMANCE 30
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! tensor_aggregator frames-in=1 frames-out=5 frames-flush=5 ! ${TS_PADDING}fakesink sync=true" 9-10 0 0 $PERFORMANCE 30
 
--- a/tests/nnstreamer_converter/runTest.sh
+++ b/tests/nnstreamer_converter/runTest.sh
@@ -92,6 +92,27 @@ gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=1 ! videos
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=\"test_00.dat\" blocksize=-1 ! application/octet-stream ! tensor_converter input-dim=4:4:4:4:4 input-type=float32 ! fakesink" 8-1 0 0 $PERFORMANCE
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} filesrc location=\"test_01.dat\" blocksize=-1 ! application/octet-stream ! tensor_converter input-dim=3:10:10:4:5:6 input-type=uint8 ! fakesink" 8-2 0 0 $PERFORMANCE
 
+# Timestamp test cases (issue #4898): untimestamped frames without a framerate
+# must get a running time, so a synchronizing sink renders them and the
+# pipeline reaches EOS. A stall here is reported by the timeout.
+TS_PADDING=""
+for i in $(seq 1 40); do
+    TS_PADDING="${TS_PADDING}identity ! "
+done
+TS_SRC="fakesrc num-buffers=5 sizetype=fixed sizemax=10 ! application/octet-stream"
+TS_CONV="tensor_converter input-dim=10 input-type=uint8"
+
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! ${TS_PADDING}fakesink sync=true" 9-1 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! ${TS_PADDING}queue ! fakesink sync=true" 9-2 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! tee name=ts ts. ! queue ! fakesink sync=true ts. ! queue ! ${TS_PADDING}fakesink sync=true" 9-3 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} set-timestamp=false ! ${TS_PADDING}fakesink sync=true" 9-4 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc is-live=true num-buffers=5 ! video/x-raw,format=GRAY8,width=16,height=16,framerate=0/1 ! tensor_converter ! ${TS_PADDING}fakesink sync=true" 9-5 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} audiotestsrc is-live=true num-buffers=5 samplesperbuffer=800 ! audio/x-raw,format=S16LE,rate=8000,channels=1 ! tensor_converter frames-per-tensor=800 ! ${TS_PADDING}fakesink sync=true" 9-6 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} set-timestamp=notabool ! fakesink sync=true" 9-7_n 0 1 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! tensor_converter ! fakesink sync=true" 9-8_n 0 1 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! mux.sink_0 ${TS_SRC} ! ${TS_CONV} ! mux.sink_1 tensor_mux name=mux ! ${TS_PADDING}fakesink sync=true" 9-9 0 0 $PERFORMANCE 30
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} ${TS_SRC} ! ${TS_CONV} ! tensor_aggregator frames-in=1 frames-out=5 frames-flush=5 ! ${TS_PADDING}fakesink sync=true" 9-10 0 0 $PERFORMANCE 30
+
 rm *.log *.bmp *.png *.golden *.dat
 
 report
--- a/tests/nnstreamer_converter/unittest_converter_timestamp.cc
+++ b/tests/nnstreamer_converter/unittest_converter_timestamp.cc
@@ -793,7 +793,7 @@ run_basepad_duplicate_timestamps (const gchar *muxer, guint mem_index, guint byt
   pipeline_push_tensor (src0, 2, MSEC (10));
   EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (src0)), GST_FLOW_OK);
   pipeline_push_tensor (src1, 100, 0);
-  pipeline_push_tensor (src1, 101, GST_SECOND);
+  pipeline_push_tensor (src1, 101, SEC (1));
   EXPECT_EQ (gst_app_src_end_of_stream (GST_APP_SRC (src1)), GST_FLOW_OK);
 
   while ((sample = gst_app_sink_try_pull_sample (GST_APP_SINK (sink), MSEC (TEST_TIMEOUT_MS)))
--- a/tests/nnstreamer_converter/unittest_converter_timestamp.cc
+++ b/tests/nnstreamer_converter/unittest_converter_timestamp.cc
@@ -708,6 +708,8 @@ TEST (tensorConverterTimestampPipeline, syncSinkNoTimestampCompletes_p)
 
 /**
  * @brief Positive: a live video source without a framerate still completes.
+ * The source stamps its frames and streams only after its own transition,
+ * so this is a smoke test of the live path rather than a window regression.
  */
 TEST (tensorConverterTimestampPipeline, liveVideoNoFramerateCompletes_p)
 {
--- a/tests/nnstreamer_converter/unittest_converter_timestamp.cc
+++ b/tests/nnstreamer_converter/unittest_converter_timestamp.cc
@@ -1,0 +1,784 @@
+/**
+ * @file        unittest_converter_timestamp.cc
+ * @date        02 Sep 2026
+ * @brief       Unit test for the timestamp handling of tensor_converter
+ * @see         https://github.com/nnstreamer/nnstreamer
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs
+ */
+
+#include <gtest/gtest.h>
+#include <glib.h>
+#include <gst/app/gstappsink.h>
+#include <gst/app/gstappsrc.h>
+#include <gst/check/gstharness.h>
+#include <gst/check/gsttestclock.h>
+#include <gst/gst.h>
+#include <unittest_util.h>
+
+#define FRAME_SIZE (8U)
+#define OCTET_CAPS "application/octet-stream"
+#define OCTET_FPS_CAPS "application/octet-stream,framerate=(fraction)5/1"
+#define VIDEO_FPS_CAPS \
+  "video/x-raw,format=GRAY8,width=8,height=1,framerate=(fraction)10/1"
+#define CONVERTER_LAUNCH "tensor_converter input-dim=8 input-type=uint8"
+#define FAKESRC_LAUNCH "fakesrc num-buffers=5 sizetype=fixed sizemax=8"
+#define ONE_HOUR (3600 * GST_SECOND)
+#define TEST_TIMEOUT_MS (10000U)
+#define IDENTITY_PADDING (40U)
+
+/**
+ * @brief Create a harness around a bare tensor_converter with the given caps.
+ */
+static GstHarness *
+harness_new_full (const gchar *caps, gboolean set_timestamp, guint frames_per_tensor)
+{
+  GstHarness *h = gst_harness_new ("tensor_converter");
+
+  if (g_str_has_prefix (caps, OCTET_CAPS))
+    g_object_set (h->element, "input-dim", "8", "input-type", "uint8", NULL);
+  g_object_set (h->element, "set-timestamp", set_timestamp, "frames-per-tensor",
+      frames_per_tensor, NULL);
+  gst_harness_set_src_caps_str (h, caps);
+  return h;
+}
+
+#define harness_new(caps) harness_new_full (caps, TRUE, 1)
+
+/**
+ * @brief Set the state of the harnessed element synchronously.
+ */
+static void
+harness_set_state (GstHarness *h, GstState state)
+{
+  EXPECT_EQ (gst_element_set_state (h->element, state), GST_STATE_CHANGE_SUCCESS);
+  EXPECT_EQ (GST_STATE (h->element), state);
+}
+
+/**
+ * @brief Push one frame through the harness and return the output buffer.
+ */
+static GstBuffer *
+harness_push_frame (GstHarness *h, GstClockTime pts, GstClockTime duration)
+{
+  GstBuffer *in = gst_harness_create_buffer (h, FRAME_SIZE);
+
+  GST_BUFFER_PTS (in) = pts;
+  GST_BUFFER_DURATION (in) = duration;
+  EXPECT_EQ (gst_harness_push (h, in), GST_FLOW_OK);
+
+  return gst_harness_try_pull (h);
+}
+
+/**
+ * @brief Push an untimestamped frame and return the timestamp assigned to it.
+ */
+static GstClockTime
+harness_push_get_pts (GstHarness *h)
+{
+  GstBuffer *out = harness_push_frame (h, GST_CLOCK_TIME_NONE, GST_CLOCK_TIME_NONE);
+  GstClockTime pts;
+
+  if (out == NULL) {
+    ADD_FAILURE () << "no output buffer";
+    return GST_CLOCK_TIME_NONE;
+  }
+
+  pts = GST_BUFFER_PTS (out);
+  gst_buffer_unref (out);
+  return pts;
+}
+
+/**
+ * @brief Positive: while PLAYING, an untimestamped frame gets the running time.
+ */
+TEST (tensorConverterTimestamp, playingRunningTime_p)
+{
+  GstHarness *h = harness_new (OCTET_CAPS);
+  GstBuffer *out;
+
+  gst_harness_set_time (h, 10 * GST_SECOND);
+  gst_element_set_base_time (h->element, 10 * GST_SECOND);
+
+  out = harness_push_frame (h, GST_CLOCK_TIME_NONE, GST_CLOCK_TIME_NONE);
+  ASSERT_TRUE (out != NULL);
+  EXPECT_EQ (GST_BUFFER_PTS (out), 0U);
+  EXPECT_FALSE (GST_BUFFER_DURATION_IS_VALID (out));
+  gst_buffer_unref (out);
+
+  gst_harness_set_time (h, 10 * GST_SECOND + 500 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 500 * GST_MSECOND);
+
+  gst_harness_set_time (h, 10 * GST_SECOND + 700 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 700 * GST_MSECOND);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief Positive: a clock behind the base time yields a zero timestamp.
+ */
+TEST (tensorConverterTimestamp, playingClockBehindBaseTime_p)
+{
+  GstHarness *h = harness_new (OCTET_CAPS);
+
+  gst_harness_set_time (h, 10 * GST_SECOND);
+  gst_element_set_base_time (h->element, 20 * GST_SECOND);
+
+  EXPECT_EQ (harness_push_get_pts (h), 0U);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief Regression (#4898): a clock without a base time must not leak the
+ * absolute clock value into the timestamp.
+ */
+TEST (tensorConverterTimestamp, pausedBeforeFirstPlaying_p)
+{
+  GstHarness *h = harness_new (OCTET_CAPS);
+
+  harness_set_state (h, GST_STATE_PAUSED);
+  ASSERT_EQ (gst_element_get_base_time (h->element), 0U);
+  gst_harness_set_time (h, ONE_HOUR);
+
+  EXPECT_EQ (harness_push_get_pts (h), 0U);
+  EXPECT_EQ (harness_push_get_pts (h), 0U);
+
+  harness_set_state (h, GST_STATE_PLAYING);
+  gst_element_set_base_time (h->element, ONE_HOUR);
+  EXPECT_EQ (harness_push_get_pts (h), 0U);
+
+  gst_harness_set_time (h, ONE_HOUR + 50 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 50 * GST_MSECOND);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief Positive: frames pushed while PAUSED keep the last timestamp and the
+ * running time continues after resuming.
+ */
+TEST (tensorConverterTimestamp, pausedAfterPlayingKeepsLast_p)
+{
+  GstHarness *h = harness_new (OCTET_CAPS);
+
+  gst_element_set_base_time (h->element, 10 * GST_SECOND);
+  gst_harness_set_time (h, 10 * GST_SECOND + 500 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 500 * GST_MSECOND);
+
+  harness_set_state (h, GST_STATE_PAUSED);
+  gst_harness_set_time (h, ONE_HOUR);
+  EXPECT_EQ (harness_push_get_pts (h), 500 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 500 * GST_MSECOND);
+
+  harness_set_state (h, GST_STATE_PLAYING);
+  gst_element_set_base_time (h->element, ONE_HOUR - 500 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 500 * GST_MSECOND);
+
+  gst_harness_set_time (h, ONE_HOUR + 100 * GST_MSECOND);
+  EXPECT_EQ (harness_push_get_pts (h), 600 * GST_MSECOND);
+
+  gst_harness_teardown (h);
+}
+
+/**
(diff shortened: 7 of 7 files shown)


myungjoo:

Scope of this review: the 6 files in the diff (`gst/nnstreamer/elements/gsttensor_converter.c`, `gst/nnstreamer/elements/gsttensor_converter.md`, `Documentation/synchronization-policies-at-mux-merge.md`, `tests/meson.build`, `tests/nnstreamer_converter/runTest.sh`, `tests/nnstreamer_converter/unittest_converter_timestamp.cc`), cross-checked against GStreamer core sources (`gstelement.c`, `gstbin.c`) and against `tensor_rate`/`tensor_mux`/`tensor_merge`/`tensor_aggregator`/`tensor_sink` in this repo.

## 1. Does the fix actually solve #4898? — Confirmed correct, no issue

I pulled the real GStreamer core implementations to check the locking claim instead of taking the PR description at face value:

- `gst_element_set_base_time()` / `gst_element_get_base_time()` and `gst_element_set_clock_func()` / `gst_element_get_clock()` all take `GST_OBJECT_LOCK(element)` around reading/writing `element->base_time` and `element->clock`.
- `gst_bin_element_set_state()` (called from `gst_bin_change_state_func()`, which iterates children sink-first via `gst_bin_iterate_sorted()`) does:
  ```c
  GST_OBJECT_LOCK (element);
  GST_ELEMENT_START_TIME (element) = start_time;
  element->base_time = base_time;
  GST_OBJECT_UNLOCK (element);
  ...
  gst_element_set_state (element, next);
  ```
  i.e. base_time is written under the **child's own** object lock, strictly before that same child's state change is even started.
- `gst_element_continue_state()` commits the new value with `current = GST_STATE (element) = old_next;` also under `GST_OBJECT_LOCK (element)`.

Because both the base_time write and the state commit for the *same element* are serialized through the *same* per-object GMutex, and the base_time write happens-before the call that eventually leads to the state commit, any reader that takes `GST_OBJECT_LOCK(self)` and observes `GST_STATE(self) == GST_STATE_PLAYING` is guaranteed (by ordinary mutex happens-before semantics, independent of which thread does the observing) to also observe the already-delivered `base_time`. So the new code:
```c
GST_OBJECT_LOCK (self);
if (GST_STATE (self) == GST_STATE_PLAYING && GST_ELEMENT_CLOCK (self)) {
  clock = gst_object_ref (GST_ELEMENT_CLOCK (self));
  base = GST_ELEMENT_CAST (self)->base_time;
}
GST_OBJECT_UNLOCK (self);
```
is a sound fix for the race described in #4898: reading raw `GST_STATE()`/`GST_ELEMENT_CLOCK()` fields while manually holding `GST_OBJECT_LOCK` is exactly the idiom GStreamer's own `gst_element_get_clock()` uses internally, and `gst_object_ref()` on the clock while holding the element's own lock is safe (it only touches the clock object's separate atomic refcount, not `self`'s lock).

I also checked the PR's claim that `gst_element_get_current_running_time()` "doesn't help": its real implementation only bails out via `GST_CLOCK_TIME_IS_VALID(base_time)`, and `GstElement::base_time` defaults to `0` (not `GST_CLOCK_TIME_NONE`) until a bin assigns it — and `GST_CLOCK_TIME_IS_VALID(0)` is `TRUE`. So that helper would have papered over the exact same bug rather than fixing it. This detail checks out and is a good reason not to use it.

**Verdict: the synchronization argument is correct and verified against upstream GStreamer source, not just asserted in the PR description.**

## 2. Regression risk to other modules — should-fix

The new fallback path (state != PLAYING, or no clock ⇒ reuse `self->old_timestamp`, or segment start for the first frame) is a real behavior change beyond what the PR description's single sentence suggests. The PR body says "frames processed while PAUSED... now keep the last timestamp," phrased as if this affects one preroll buffer. In practice, more than one buffer can flow through `tensor_converter` while `GST_STATE(self) == GST_STATE_PAUSED`: the actual block point for a `PAUSED` pipeline is downstream, in the sink's preroll queue (`GstBaseSink`), not at the converter itself — anything upstream of that queue (including a `queue` element, or a still-running `GstBaseSrc` streaming thread) can keep pushing several buffers into the converter before the sink blocks. With this fix, **all of those buffers can now receive the exact same PTS** (not just distinct-but-wrong ones as before), since each call reuses `old_timestamp` verbatim. This is real and is directly demonstrated in the PR's own `pausedAfterPlayingKeepsLast_p` test, which pushes two buffers in PAUSED and asserts both get `500ms`.

I checked whether downstream consumers named in the task (tensor_mux/merge sync policies, tensor_aggregator, tensor_rate, tensor_sink) tolerate duplicate/non-monotonic PTS. Three of them are fine; one has a real, silent bug that this PR makes newly reachable:

- `gst/nnstreamer/elements/gsttensor_rate.c:550-561` explicitly drops a buffer whose timestamp is `< prevtime` (`if (intime < prevtime) { ...; return GST_BASE_TRANSFORM_FLOW_DROPPED; }`), and the interpolation loop at `gsttensor_rate.c:564-598` exits on `diff1 < diff2` (strict), with an explicit comment "if they were equal avoid going into an infinite loop." Equal-PTS bursts are handled safely, no div/underflow.
- `gst/nnstreamer/elements/gsttensor_aggregator.c:863-866,913-915` only derives output duration from the *input buffer's own* duration field (`gst_util_uint64_scale_int(duration, frames_out, frames_in)`) and takes the first frame's PTS as the output PTS; it never diffs against a previously seen buffer's PTS, so it's unaffected either way.
- `gst/nnstreamer/elements/gsttensor_sink.c:528` (`silent_debug_timestamp`) only logs PTS for debugging; no logic depends on it.
- **`gst/nnstreamer/nnstreamer_plugin_api_impl.c` — `tensor_mux`/`tensor_merge` with `sync-mode=basepad` (shared by both elements via `gst_tensor_time_sync_buffer_from_collectpad()`): a genuine unsigned-wraparound bug, made newly reachable by this PR.** At `nnstreamer_plugin_api_impl.c:372-376`:
  ```c
  if (pad->buffer != NULL)
    base_time =
        MIN ((GstClockTimeDiff) sync->data_basepad.duration,
        ABS (GST_CLOCK_DIFF (GST_BUFFER_PTS (buf),
                GST_BUFFER_PTS (pad->buffer))) - 1);
  ```
  `base_time` is declared `GstClockTime base_time = 0;` (unsigned 64-bit, line 345). When the base pad's newly peeked buffer (`buf`) and its previously cached buffer (`pad->buffer`) carry **equal PTS**, `GST_CLOCK_DIFF(...)` is `0`, `ABS(0) - 1` evaluates to `-1` as a signed `gint64`, and `MIN` picks that `-1` (it is smaller than any non-negative tolerance duration). Assigning `-1` into the unsigned `base_time` wraps it to `G_MAXUINT64`. That corrupted value is then passed as the tolerance `base` into `_gst_tensor_time_sync_buffer_update()` for every pad on the collection (`nnstreamer_plugin_api_impl.c:423-424`), where the `SYNC_BASEPAD` branch's tolerance check at `nnstreamer_plugin_api_impl.c:309-311` (`ABS(GST_CLOCK_DIFF(current, GST_BUFFER_PTS(buf))) > base`) can now essentially never be true. This silently disables the "keep last buffer / respect tolerance window" logic for that collection cycle — non-base pads get paired against the base pad's frame regardless of actual time distance. No crash, no hang, just **silently wrong frame pairing**, which is worse than a hard failure because it is hard to notice.

  This is a pre-existing bug in a file this PR does not touch, and it isn't unique to `tensor_converter` (any base-pad stream that ever emits two consecutive equal PTS values triggers it). But before this PR, hitting two *consecutive* equal PTS values from the buggy `clock - base_time` stamping was astronomically unlikely (nanosecond clock reads essentially never collide). After this PR, a `tensor_converter` feeding the base pad of a `basepad`-mode `tensor_mux`/`tensor_merge` will *routinely* emit runs of identical PTS values during every PAUSED window (including the mandatory preroll phase every pipeline goes through), which makes this latent bug meaningfully easier to hit in practice. I confirmed `tensor_mux`'s default `sync.mode` is `SYNC_SLOWEST` (`gsttensor_mux.c:216`), not `SYNC_BASEPAD`, so the PR's own `muxSyncSinkCompletes_p`/SSAT `9-9` case (which uses the default) does not exercise this path and would not catch it.

  Recommendation: this PR does not need to (and per its own scope discipline, probably should not) fix `nnstreamer_plugin_api_impl.c` inline, but the interaction should not go unrecorded — please either (a) file a follow-up issue referencing this PR and #4898 for the `SYNC_BASEPAD` unsigned-wraparound bug, or (b) add a short caveat to the "Timestamps" section of `gsttensor_converter.md` noting that `sync-mode=basepad` consumers should not currently be paired with untimestamped/no-framerate sources through `tensor_converter`. Note a black-box `gst-launch`/SSAT test with `sync-mode=basepad` would likely *pass* here regardless (the bug causes wrong pairing, not an error or hang), so this needs either a golden-value/frame-content SSAT case or a dedicated `nnstreamer_plugin_api_impl.c`-level gtest to actually catch it — worth keeping in mind if a follow-up fix is filed.

The doc/PR-description precision issue from the first paragraph above (a *burst* of frames, not just one, can share a PTS while PAUSED) is a separate, much smaller nit worth a one-line clarification, but is not the main finding here.

## 3. Code size / scope — no issue

The functional change is 20 lines in one function of one file; the rest of the ~838-line diff is the new gtest file (784 lines) plus doc/SSAT additions, which the repo's own test-case policy (CLAUDE.md: "A fix without a reproducing test is incomplete") requires. No unrelated files or modules are touched. Scope is appropriately tight.

## 4. Test coverage — no issue, thorough

The GstHarness unit tests correctly cover: running-time-while-PLAYING, clock-behind-base_time clamp to 0, the exact PAUSED-before-first-PLAYING window from #4898, pause/resume monotonicity, segment-start fallback, valid-input passthrough, `set-timestamp=false` negative, no-clock fallback, both framerate branches (video and octet-with-framerate) bypassing the clock, flush/READY history reset, BYTES→TIME segment conversion, and `frames-per-tensor` aggregation. The pipeline-level tests (`testClockRunningTime_p`, `clockWithoutBaseTime_p`) and the padded `sync=true` end-to-end tests (`syncSinkCompletes_p` et al., including `tensor_mux` and `tensor_aggregator` consumers) exercise real pipeline race conditions, not just the isolated branch logic. There's a negative case (`octetWithoutDimension_n`) too. This satisfies both positive/negative coverage and "future regression in another module would be caught" (the `tensor_mux`/`tensor_aggregator` pipeline cases specifically guard against a regression introduced elsewhere breaking the interaction with this fix).

One nit: I verified `gst_harness_new()` only auto-transitions the harnessed element to PLAYING when it has a sink pad (`if (has_sinkpad) gst_harness_play(h);`, from GStreamer's own `gstharness.c`) and always attaches a `GstTestClock` by default. `tensor_converter` has a sink pad, so this assumption (relied on implicitly by every harness test that pushes a buffer without an explicit `harness_set_state` call) is correct. Good.

## 5. CI wiring — no issue

- `tests/meson.build`: the new `test('unittest_converter_timestamp', ...)` follows the existing sibling pattern exactly (same deps, same `install_dir`), so it will build and run under the same jobs as every other `unittest_*` gtest binary.
- `packaging/run_unittests_binaries.sh` discovers binaries via `find "${input}" ... -name "unittest_*"` in directory mode, so `unittest_converter_timestamp` is picked up automatically without any script change needed. Nit: `run_entry()`'s special-case branch matches on `*"unittest_converter"*` (substring), which also matches `unittest_converter_timestamp` and will needlessly set up the nnstreamer-python `.so` symlink for a binary that doesn't use it. Harmless (just a few no-op filesystem calls before the binary runs), not blocking.
- The `gstTest "..." <id> <errorcase> <skip> $PERFORMANCE <timeout>` 6-argument (timeout) call form used for the new `9-1`..`9-10` cases is not a new pattern — it's already used in `tests/nnstreamer_edge/query/runTest.sh`, `tests/nnstreamer_edge/edge/runTest.sh`, and `tests/gstreamer_join/runTest.sh`, all of which already run under the pdebuild/GBS/RISC-V SSAT jobs, so there's no version-support risk being introduced here.

## 6. Docs — no issue

`gsttensor_converter.md` gets a new "Timestamps" section describing the exact new rule (framerate branch / PLAYING-clock branch / PAUSED-or-no-clock fallback), and `synchronization-policies-at-mux-merge.md`'s one relevant sentence is updated to match. No public API/property/signal was added or changed, so no other document needed updating; I checked `Documentation/synchronization-in-mqtt-elements.md` (the other doc mentioning base-time/running-time) and it's a generic base-time/running-time concept explainer unrelated to when `tensor_converter` is allowed to read the clock, so it's correctly left untouched.

## Overall verdict

No blocking issues found in the diff itself. The core fix is verified correct against actual GStreamer core locking semantics (not just plausible-sounding — I traced `gst_bin_element_set_state`/`gst_element_continue_state`/`gst_element_get_base_time` in upstream GStreamer to confirm the happens-before argument), and the test suite is thorough with positive/negative/pipeline-level coverage, correct `GstHarness` usage, and CI wiring that follows established repo conventions (meson registration, auto-discovery by `run_unittests_binaries.sh`, precedented `gstTest` timeout usage).

The one item that should be addressed, though **not necessarily inside this PR's diff**, is the `should-fix` finding in item 2: this PR's new "reuse the last PTS while PAUSED" behavior makes a pre-existing unsigned-wraparound bug in `nnstreamer_plugin_api_impl.c`'s `SYNC_BASEPAD` path (shared by `tensor_mux`/`tensor_merge`) meaningfully easier to trigger in real pipelines, and that bug silently mispairs frames rather than erroring. Given it lives in a file this PR correctly doesn't touch, I'd treat this as **should-fix via a follow-up issue/PR**, not a blocker for merging this one — but it should be filed and cross-referenced before this lands, so it doesn't get lost. The remaining points (doc wording on "burst of frames," the harmless `run_unittests_binaries.sh` substring match) are nits.

**Recommendation: mergeable, conditional on filing a follow-up issue for the `SYNC_BASEPAD` interaction identified in item 2** (or, if the maintainers judge the exposure low enough given `SYNC_SLOWEST` is the default, at minimum noting the caveat in the PR description before merge).

myungjoo:

Thanks for the review. Response to item 2 (SYNC_BASEPAD tolerance wrap):

- Confirmed. `|PTS(new) - PTS(prev)| - 1` is -1 for two equal base-pad timestamps and wraps to `G_MAXUINT64` in the unsigned `base_time`, so the other pads accept any frame for that cycle.
- One correction on likelihood: equal consecutive timestamps were already ordinary before this PR. Every frame the converter stamps before the pipeline has a clock (prerolling through a `queue`, for instance) carries the segment start, so a base pad fed by the converter could see runs of identical timestamps all along. The PAUSED fallback adds one more source, which is why the clamp is fixed here rather than deferred.
- Fixed in 134fea61: the tolerance is clamped at 0 when the gap is 0. Two pipeline tests (`muxBasepadDuplicateTimestamps_p`, `mergeBasepadDuplicateTimestamps_p`) push two equal timestamps on the base pad and a far frame on the other pad and require the far frame to be held back on all three outputs. Against the unfixed core they fail as predicted (the far frame is paired and the third output disappears); with the clamp they pass, and the mux/merge/rate SSAT suites are unchanged (84/84, 89/89, 17/17 locally).

Items 1, 3, 4, 5, 6: no change needed, as concluded in the review.

myungjoo:

Scope of this re-review: the full PR as of HEAD `134fea61` (both commits `8d1535ce` and `134fea61`), with emphasis on the second commit (`[Mux/Merge] Keep the basepad tolerance when two base timestamps are equal`, `gst/nnstreamer/nnstreamer_plugin_api_impl.c` + the two new pipeline gtests) and the author's reply to the prior automated review. Cross-checked against the shared `_gst_tensor_time_sync_buffer_update()` / `gst_tensor_time_sync_get_current_time()` machinery in this file and its callers in `gsttensor_mux.c` / `gsttensor_merge.c`, and against GstCollectPads semantics.

## 1. Does the PR solve #4898? — no issue (unchanged from prior review)

The first commit (`8d1535ce`) is untouched by this update. The prior review already traced `gst_bin_element_set_state()` / `gst_element_continue_state()` / `gst_element_get_base_time()` in upstream GStreamer to confirm the `GST_OBJECT_LOCK`-protected read of `GST_STATE()`+`GST_ELEMENT_CLOCK()`/`base_time` is a sound fix for the root cause described in #4898. No new information changes that conclusion.

## 2. Correctness of the second commit's clamp — no issue, verified by hand

Old code (`nnstreamer_plugin_api_impl.c`, pre-PR):
```c
base_time = MIN ((GstClockTimeDiff) sync->data_basepad.duration,
    ABS (GST_CLOCK_DIFF (GST_BUFFER_PTS (buf), GST_BUFFER_PTS (pad->buffer))) - 1);
```
New code:
```c
GstClockTimeDiff gap = ABS (GST_CLOCK_DIFF (GST_BUFFER_PTS (buf), GST_BUFFER_PTS (pad->buffer)));
base_time = MIN ((GstClockTimeDiff) sync->data_basepad.duration, (gap > 0) ? (gap - 1) : 0);
```
`gap` is a `GstClockTimeDiff` (signed `gint64`) produced by `ABS()`, so it is always ≥ 0. For `gap > 0` the new expression reduces to `gap - 1`, bit-for-bit identical to the old `ABS(...) - 1`. The only behavior change is at `gap == 0`: old code produced `-1` (signed), which the `MIN` picked (it's less than any non-negative `duration`), and assigning that `-1` into the unsigned `base_time` (`GstClockTime`, `guint64`) wrapped to `G_MAXUINT64`. The new code produces `0` in that case, both operands to `MIN` staying non-negative throughout, so no unsigned wraparound is possible. This is the correct, minimal fix for exactly the bug the prior review identified.

## 3. Regression risk — no issue, checked by full trace of the new gtest

Since the `gap > 0` branch is unchanged, no existing basepad-mode SSAT/gtest pipeline (which use real framerate-driven timestamps and essentially never see two consecutive *equal* base-pad PTS) can regress; the fix only changes behavior in a case that was previously broken. I did not re-run the suites in this environment (no local build was set up), but the equivalence proof above is a stronger guarantee than a single test run, and the author reports 84/84 (`tensor_mux`), 89/89 (`tensor_merge`), 17/17 (`tensor_rate`) SSAT locally, consistent with this analysis.

I additionally hand-traced `run_basepad_duplicate_timestamps()` in the new gtest (`muxBasepadDuplicateTimestamps_p` / `mergeBasepadDuplicateTimestamps_p`) against the actual collectpads code path to confirm the test is sound and deterministic (see item 5) — no additional regression surfaced.

## 4. Code size / scope — no issue

The functional change is ~8 lines in one function of one file, plus one new/updated test helper. Including it in this PR (rather than a separate follow-up) is appropriate: it is a direct, minimal consequence of the interaction the prior review flagged between this PR's own PAUSED-fallback (which introduces more duplicate-PTS runs on converter-fed base pads) and a pre-existing latent bug it makes easier to hit. The author's own reply also correctly notes duplicate consecutive PTS were already possible before this PR (via the segment-start stamping on every pre-clock frame), so the fix is not purely reactive scope creep — it closes a real gap the topic under review exposed. No unrelated files or modules are touched.

## 5. Test determinism and correctness — no issue, verified by manual simulation

I traced `run_basepad_duplicate_timestamps()` (`tensor_mux name=mux sync-mode=basepad sync-option=0:50000000`, base pad = `sink_0` = `src0` pushing PTS `0, 0, 10ms`, non-base pad = `sink_1` = `src1` pushing PTS `0, 1s`) against the real collected-callback code:

- `gst_tensor_mux_collect_buffer()` only calls `gst_tensor_time_sync_get_current_time()` (which does the *real*, fresh-`peek()`-based empty/EOS detection: `if (buf) {...} else { empty_pad++; }`) when `need_set_time` is `TRUE`; it is set back to `TRUE` only after each successful `gst_pad_push()` (`gsttensor_mux.c:543-544`, mirrored in `gsttensor_merge.c:361/749`). So each output round performs exactly one real EOS check gated on actual queue occupancy, not on the stale `pad->buffer` cache used inside `_gst_tensor_time_sync_buffer_update()`.
- Round 1 bootstraps both pads unconditionally (no `pad->buffer` yet to compare against). Round 2: base pad's two most recent PTS are both `0` → `gap == 0` → **with the fix**, tolerance clamps to `0`; the non-base pad's queued far buffer (PTS `1s`) is `> 0` away, so it is correctly held back ("keep last", not popped — it stays queued for the next round rather than being discarded). Round 3: base pad's gap becomes `0` vs `10ms` (non-zero, ordinary path), tolerance ≈ `9.999ms`; the still-queued far buffer is still ~990ms away, still held back. After round 3 the base pad's collectpads queue is drained and it has already delivered EOS; the *next* `need_set_time` check sees the base pad's fresh `peek()` return `NULL` → `empty_pad = 1` → `_gst_tensor_time_sync_is_eos()` (default branch: `if (empty > 0) is_eos = TRUE`) fires immediately, so `gst_tensor_mux_collect_buffer()` returns `FALSE`/`isEOS` *before* a 4th round can ever call `buffer_from_collectpad()` — no spurious 4th output, regardless of the still-unconsumed leftover buffer sitting in the non-base pad's queue. This gives exactly 3 rounds, matching `EXPECT_EQ (received, 3U)`.
- Without the clamp, round 2's tolerance wraps to `G_MAXUINT64`, so the far buffer (`1s` away) is accepted immediately in round 2 (pops the queued far buffer into `pad->buffer`), and the non-base pad's queue becomes empty one round earlier. From that point the byte-content assertion (`map.data[byte_index] == 100`) fails for whichever round wrongly received the paired-in far value, while the round count itself is unaffected by the bug (both the buggy and fixed path terminate via the same base-pad-driven EOS check) — so the test's discriminating signal is the byte value, not the buffer count, but the count assertion is unconditionally true in both cases and doesn't weaken the test.
- Determinism does not depend on OS/thread scheduling: `GstCollectPads` only advances a "round" based on FIFO-ordered per-pad data availability (buffer-or-EOS), never on wall-clock arrival order between pads, so which of `src0`'s or `src1`'s streaming thread happens to run first cannot change the outcome — the peek/pop sequence and the accept/hold decision are fully determined by the PTS values pushed and the fixed formula.

This matches the "고정된 collectpads 동작, 스레드 타이밍에 비의존적" requirement from the review brief. Good coverage; nothing to change.

## 6. Future regression detection / CI gating — no issue

`muxBasepadDuplicateTimestamps_p` / `mergeBasepadDuplicateTimestamps_p` are registered as ordinary `TEST()` cases inside `unittest_converter_timestamp`, which is wired into `tests/meson.build` exactly like every other `unittest_*` binary (already confirmed in the prior review). A future change to `nnstreamer_plugin_api_impl.c`'s `SYNC_BASEPAD` tolerance logic, or to `GstCollectPads` usage in `tensor_mux`/`tensor_merge`, that reintroduces the wraparound (or otherwise breaks the tolerance window) would fail these two assertions, and `meson test` failures block CI the same way as any other gtest in this suite — this satisfies the "future regression must be caught and must gate CI" requirement.

## 7. Documentation — no issue

This is a pure internal bugfix to a formula that was never itself documented: `Documentation/synchronization-policies-at-mux-merge.md`'s `# BasePad` section describes `sync-mode=basepad`/`sync-option` usage but does not spell out the `gap - 1` tolerance derivation, so there is nothing stale to update. No public API, property, or documented behavior changed by this commit.

## Nits (non-blocking)

- The PR description's "the far frame gets paired and one output disappears" is a slightly loose way to describe the failure mode; per the trace above it's more precisely "one output's cross-pad value is wrong (the far frame replaces the held one)" rather than a buffer literally vanishing — the round count is unaffected by the bug either way. Doesn't affect correctness or the test's validity, just phrasing.
- Unrelated to this diff, but noted for completeness since I was in the function: `GST_CLOCK_DIFF()`/`ABS()` on `GST_BUFFER_PTS(pad->buffer)` would misbehave if a base-pad buffer ever carried `GST_BUFFER_PTS == GST_CLOCK_TIME_NONE` (`G_MAXUINT64` reinterpreted as `-1` signed), but that is a pre-existing condition this PR neither introduces nor worsens, and tensor_mux/merge already require valid PTS upstream in normal operation — out of scope here, not something to fix in this PR.

## Overall verdict

No blocking issues in either commit. The second commit is a correct, minimal, well-scoped fix for exactly the bug the prior review raised; the fix is mathematically equivalent to the old formula outside the previously-broken `gap == 0` case (so no regression risk to existing basepad users), and the two new pipeline gtests are deterministic by construction (GstCollectPads' FIFO/blocking semantics plus the `need_set_time`-gated real EOS check), give byte-exact discrimination between the buggy and fixed behavior, and are wired into the standard CI-gating test suite.

**Recommendation: mergeable, no blocking issues found.**

myungjoo:

Thanks. On the wording nit: "one output disappears" is what the negative control actually measured, not a loose phrase. Against the unfixed sync core the two new tests report the far frame paired on the second output (`101` instead of `100`) *and* `received == 2` instead of 3: once the far frame is consumed in round 2, the other pad is drained and at EOS, so the base pad's third frame ends the stream without a third output. Both symptoms are asserted by the tests. No change made.

The `PTS == GST_CLOCK_TIME_NONE` observation is pre-existing and out of scope, as noted.

myungjoo:
[Automated final review by a separate Claude agent, relayed to this PR]

Scope of this final review: the full PR at HEAD `99d7c5bc` (all three commits: `8d1535ce`, `134fea61`, `99d7c5bc`), with the two prior automated reviews and the author's replies read in full first. I re-verified items 1 and 2 (already reviewed twice) only enough to confirm the third commit doesn't touch their code paths, and put the depth of this pass into the third commit, since that's the only new content since the last review round.

## 1. Does the PR solve #4898? — no issue (re-confirmed, unchanged)

Commits `8d1535ce` and `134fea61` are untouched by `99d7c5bc`. Both prior review rounds already traced `gst_bin_element_set_state()` / `gst_element_continue_state()` / `gst_element_get_base_time()` in upstream GStreamer to confirm the `GST_OBJECT_LOCK`-protected read of state+clock+base_time in `_gst_tensor_converter_chain_timestamp()` is a sound fix for the race in #4898, and hand-simulated `run_basepad_duplicate_timestamps()` against real `GstCollectPads`/`need_set_time` semantics to confirm the basepad tolerance clamp in `nnstreamer_plugin_api_impl.c` is correct and deterministic. Nothing in the third commit changes that analysis; I re-read both traces and they still hold.

## 2. Regression risk to other modules — no issue (the round-1 should-fix is resolved)

Round 1 flagged a real pre-existing `SYNC_BASEPAD` unsigned-wraparound bug (`ABS(GST_CLOCK_DIFF(...)) - 1` becoming `-1` and wrapping to `G_MAXUINT64` when two base-pad timestamps are equal) that this PR's PAUSED-fallback makes easier to hit. Commit `134fea61` fixed it in-PR with a `(gap > 0) ? (gap - 1) : 0` clamp, and round 2 verified the fix is bit-for-bit equivalent to the old formula outside the previously-broken `gap == 0` case, with two new deterministic pipeline gtests (`muxBasepadDuplicateTimestamps_p`, `mergeBasepadDuplicateTimestamps_p`) discriminating the buggy vs. fixed behavior by buffer byte content. That should-fix item is closed, not deferred. Commit `99d7c5bc` is test-file-only and touches no production code, so it carries zero regression risk to `tensor_mux`/`tensor_merge`/`tensor_rate`/`tensor_aggregator`/`tensor_sink` or any other module.

## 3. Correctness of commit 3 (`99d7c5bc`, SEC()/MSEC() macros) — no issue, verified against upstream GStreamer headers

I did not take the commit message's technical claim on faith; I pulled GStreamer's actual `gst/gstclock.h`:

```c
typedef guint64 GstClockTime;
typedef gint64  GstClockTimeDiff;
#define GST_SECOND  ((GstClockTimeDiff)(G_USEC_PER_SEC * G_GINT64_CONSTANT (1000)))
#define GST_MSECOND ((GstClockTimeDiff)(GST_SECOND / G_GINT64_CONSTANT (1000)))
```

`GST_SECOND`/`GST_MSECOND` are indeed `GstClockTimeDiff` (signed `gint64`), exactly as the commit message states — this is not a misdiagnosis.

I also confirmed the build actually enables the warning as `-Werror`: `meson.build` sets `default_options: ['werror=true', 'warning_level=2', ...]`, and meson's `warning_level=2` adds `-Wextra` (which carries `-Wsign-compare`) on top of `-Wall`. Combined with `werror=true` this is effectively `-Werror=sign-compare`, matching the commit message. GCC and Clang are known to diverge on exactly this pattern — GCC suppresses `-Wsign-compare` for values it can constant-fold as provably non-negative even inside macro-expanded comparisons (here, `EXPECT_EQ`'s internal template comparison), while Clang does not extend the same suppression — so "gcc accepts it, clang fails" is a real, previously-documented compiler discrepancy, not a made-up excuse.

The fix itself:
```c
#define MSEC(n) ((GstClockTime) (GST_MSECOND * (n)))
#define SEC(n)  ((GstClockTime) (GST_SECOND * (n)))
#define ONE_HOUR SEC (3600)
```
computes the same signed 64-bit product as before (no value change, `n` is properly parenthesized so an expression argument can't be mis-precedenced) and only adds a cast to `GstClockTime` (unsigned) at the outermost level. I diffed every hunk in `99d7c5bc` by hand: every change is a 1:1 textual substitution of `K * GST_SECOND`/`K * GST_MSECOND` → `SEC(K)`/`MSEC(K)` (or the reverse-order `TEST_TIMEOUT_MS * GST_MSECOND` → `MSEC(TEST_TIMEOUT_MS)`), with no expected value, comparison direction, or assertion reordered or altered. **The commit does not change what any test verifies — it only changes the C++ type of the compile-time constant on the right-hand side of the comparisons.**

I then grepped every `EXPECT_*`/`ASSERT_*` call left in the file after this commit (`tests/nnstreamer_converter/unittest_converter_timestamp.cc`) to check for leftover mixed-signedness comparisons: all of them now compare a `GstClockTime`-returning call (`harness_push_get_pts()`, `GST_BUFFER_PTS()`, `GST_BUFFER_DURATION()`, `pipeline_pull_pts()`, `gst_element_get_base_time()`, `gst_clock_get_time()`) against either an unsigned literal (`0U`, `2U`, `3U`, `100U`) or a `SEC()`/`MSEC()`/`ONE_HOUR` expression — i.e. unsigned-vs-unsigned throughout. I found no remaining signed/unsigned mismatch in a comparison.

One raw, un-wrapped use remains: `pipeline_push_tensor (src1, 101, GST_SECOND);` (line 796). This is *not* a bug: `pipeline_push_tensor`'s `pts` parameter is `GstClockTime`, so this is an implicit conversion at a function-call argument site, not a relational/equality comparison — `-Wsign-compare` does not fire on argument passing or initialization, only on `==`/`!=`/`<`/`>`/`<=`/`>=`. This is consistent with why it compiles clean on clang without being touched. It's a minor textual inconsistency with the commit's stated "wrap the expected values" intent (nit, not a defect).

## 4. Nit: macros don't guard against negative input

`SEC(n)`/`MSEC(n)` silently wrap a negative `n` into a huge unsigned value (`(GstClockTime)(negative gint64)` underflows to near-`G_MAXUINT64`) instead of failing loudly. Not exercised by any current call site (all arguments here are non-negative literals), and not a defect in this PR, but since these macros exist only in this test TU and might get reused by a future test author who passes a computed offset, an assert or a `static_assert`-style guard would make a future misuse fail fast instead of producing a bizarre multi-year timestamp. Non-blocking.

## 5. Test coverage / meaning preserved — no issue

Per item 3's diff-by-diff check, no test's expected value, tolerance, or control flow changed; the third commit is a pure compile-portability fix. No new regression test is expected or needed for it — CLAUDE.md's own test-case policy carve-out ("state the reason ... build-script-only changes") applies by analogy: a `-Wsign-compare` build failure isn't a gtest-expressible behavior, and the thing that actually re-guards it going forward is CI itself (see item 6).

## 6. Future-regression detection / CI gating — no issue

Because `-Wsign-compare` is enabled via `warning_level=2` + `werror=true` project-wide (not something bolted onto this one file), any future test edit in this file (or introduction of a new signed/unsigned mismatch anywhere else in the C++ test suite) that reintroduces this class of bug will again fail the "Build on Ubuntu with LLVM/Clang" and "Build on MacOS-latest with Meson" CI jobs and block merge, exactly as it did before this commit was added. Confirmed both are `SUCCESS` on HEAD `99d7c5bc` in the current `statusCheckRollup`, along with every other job (pdebuild, GBS x86_64/armv7l/aarch64, Yocto, Android NDK, RISC-V, Valgrind ×2, CodeQL, DCO, static checks, CPP-Linter, spell check) — all green.

## 7. Code size / scope — no issue

Commit 3 touches exactly one file already in this PR's own diff (`tests/nnstreamer_converter/unittest_converter_timestamp.cc`), 68 insertions / 68 deletions, purely mechanical. No other module touched.

## 8. Docs — no issue (n/a)

No architecture or API change in any of the three commits combined; commit 3 is a test-only compile fix. Docs from commit 1 (`gsttensor_converter.md` Timestamps section) and the `synchronization-policies-at-mux-merge.md` update from commit 1/2's scope were already reviewed and found adequate in round 1/2; nothing new requires doc changes.

## Overall verdict

No blocking issues across all three commits. Commit 1's core fix is sound (verified against upstream GStreamer locking semantics in round 1). Commit 2 correctly and minimally closes the `SYNC_BASEPAD` wraparound gap that round 1 flagged, with deterministic tests verified by hand-trace in round 2. Commit 3 is a narrowly-scoped, verifiably meaning-preserving compile-portability fix: `GST_SECOND`/`GST_MSECOND` are confirmed to be `GstClockTimeDiff` in upstream GStreamer, the project's `-Werror`+`warning_level=2` configuration confirmed to actually enable `-Werror=sign-compare`, every comparison in the affected file now checked unsigned-vs-unsigned with no leftover mismatch, and no expected value or assertion was altered anywhere in the commit.

**Recommendation: mergeable, no blocking issues found.** Two non-blocking nits noted above (the one un-wrapped `GST_SECOND` argument at line 796, and the unguarded-negative-input footgun in `SEC()`/`MSEC()`) — neither needs to hold up this PR.

myungjoo:

Thanks for the final pass.

- The one un-wrapped `GST_SECOND` argument is now `SEC (1)` for consistency (commit pushed on top; test-only, one token).
- The negative-argument guard for `SEC()`/`MSEC()` is left out on purpose: the helpers exist only in this test file, every call site passes a literal, and a guard would add a runtime check or a `static_assert` scaffold to a test-only convenience macro for a misuse nobody makes. If a future test needs signed offsets it should compute them as `GstClockTimeDiff` explicitly.

myungjoo:
[Automated confirmation review by a separate Claude agent, relayed to this PR]

Scope: final confirmation pass over HEAD `736f1533`, focused on (1) whether the 4th commit does exactly what its message claims, (2) whether the author's reply to the round-3 nits holds up, and (3) whether anything new was introduced across all four commits (`8d1535ce`, `134fea61`, `99d7c5bc`, `736f1533`). The three prior automated review rounds already verified the core fix against upstream GStreamer locking semantics, the `SYNC_BASEPAD` tolerance clamp by hand-trace, and the `-Wsign-compare` rationale against the actual GStreamer headers — I did not re-derive those, only re-checked that nothing after round 3 touches those conclusions.

## 1. Commit `736f1533` — confirmed, single-token change only

`git diff 99d7c5bc..736f1533` is exactly:

```
-  pipeline_push_tensor (src1, 101, GST_SECOND);
+  pipeline_push_tensor (src1, 101, SEC (1));
```

One file, one line, +1/-1. `SEC(1)` expands to `((GstClockTime)(GST_SECOND * (1)))`, numerically identical to `GST_SECOND`; the only change is the explicit unsigned cast, applied at a function-argument site (`pipeline_push_tensor`'s `pts` parameter is already `GstClockTime`), not a comparison, so it doesn't change the test's semantics — it was already an implicit conversion before. A repo-wide grep after this commit (`grep -n "GST_SECOND\|GST_MSECOND" tests/nnstreamer_converter/unittest_converter_timestamp.cc`) shows the only remaining occurrences of `GST_SECOND`/`GST_MSECOND` are inside the `SEC()`/`MSEC()` macro definitions themselves — no raw usage is left anywhere else in the file. This matches the commit message exactly and confirms the round-3 review's nit-1 finding ("one un-wrapped `GST_SECOND` argument at line 796") is fully addressed.

## 2. Author's reply (`5506041828`) — both answers hold up

- **Nit 1 (unwrapped `GST_SECOND`)**: accepted and fixed via commit `736f1533`, verified above. Sound.
- **Nit 2 (no negative-input guard on `SEC()`/`MSEC()`)**: declined, with the reasoning that the macros are test-file-local, every current call site passes a non-negative literal, and a future test needing a signed offset should compute it explicitly as `GstClockTimeDiff` rather than misusing these helpers. I confirmed the factual premise: `grep -n "SEC (-\|MSEC (-"` over the file returns no matches — there is no negative-argument call site today, including `ONE_HOUR - MSEC (500)` (line ~178), which subtracts a small value from a large positive constant and stays positive. Declining a defensive guard on an internal test-only macro with no live misuse is a reasonable, proportionate call, not a shortcut that hides a real defect.

Both answers are logically sound and consistent with what the diff actually shows.

## 3. New issues across all four commits — none found

I re-diffed each commit boundary (`8d1535ce`, `134fea61`'s isolated hunk in `nnstreamer_plugin_api_impl.c`, `99d7c5bc`, `736f1533`) against what the three prior review rounds already validated (object-lock ordering fix in the converter, the `gap > 0 ? gap - 1 : 0` basepad-tolerance clamp, the `SEC()`/`MSEC()` sign-compare fix) and found no discrepancy — the code at HEAD matches what was reviewed and accepted in rounds 1–3, plus the one-line change from commit 4 covered above. No unrelated files were touched; the changed-file set (`Documentation/synchronization-policies-at-mux-merge.md`, `gst/nnstreamer/elements/gsttensor_converter.c`, `gst/nnstreamer/elements/gsttensor_converter.md`, `gst/nnstreamer/nnstreamer_plugin_api_impl.c`, `tests/meson.build`, `tests/nnstreamer_converter/runTest.sh`, `tests/nnstreamer_converter/unittest_converter_timestamp.cc`) is unchanged from what was already reviewed.

Note: CI on `736f1533` was still in `pending` state at the time of this review (fresh push); given the change is a test-only, value-preserving edit already covered by the same jobs that ran green on `99d7c5bc`, this is not expected to be a concern, but worth confirming CI finishes green before merge.

## Overall verdict

No blocking issues. The 4th commit does exactly the single-token, meaning-preserving change it claims to. Both parts of the author's final reply are logically valid. No new problems surfaced anywhere in the PR's four commits.

**Recommendation: mergeable**, pending the still-in-flight CI run on `736f1533` turning green (same jobs that already passed on the prior commit).

myungjoo:

Thanks, all three actionable points are taken in the latest commit:

1. Agreed that `9-5`, `9-6` and `liveVideoNoFramerateCompletes_p` cannot reach the clock branch: `videotestsrc`/`audiotestsrc` stamp every frame, and a live source streams only after its own transition, so the window is unreachable for it by construction. They are kept as smoke coverage of the live path and the comments in `runTest.sh` and the gtest now say exactly that instead of describing them as window coverage.
2. The timeout argument is dropped from `9-7_n` and `9-8_n`; both fail immediately, and without the bound a future stall there can no longer be scored as the expected failure. The positive cases keep it.
3. PR body updated to 25 cases, with a note that the smoke cases pass either way in the negative control.
4. Noted; the `gst_memory_new_wrapped` leak under memcheck has no frame of the new test code and matches the existing baseline noise.

myungjoo:
[Automated confirmation review by a separate Claude agent, relayed to this PR]

Scope: confirmation pass over the fifth commit, `804bbe81` ("Label the live-source smoke cases and drop the negative timeouts"), against the four non-blocking suggestions in myungjoo-bot's 2026-09-02T10:21Z review and the author's reply.

## 1. Diff matches the reply, and is scoped exactly as claimed

`git show 804bbe81 --stat` touches only `tests/nnstreamer_converter/runTest.sh` (+13/-5) and `tests/nnstreamer_converter/unittest_converter_timestamp.cc` (+2/-0) — no production code. Content check against the reply's four points:

- **Point 1** (relabel 9-5/9-6/`liveVideoNoFramerateCompletes_p`): done. The `runTest.sh` block comment now separates the cases into three groups — `9-1..9-4, 9-9, 9-10` ("feed untimestamped frames without a framerate through a widened state-change window ... a stall is reported by the timeout"), `9-5/9-6` ("smoke cases for live sources, which stamp their own frames and stream only after their own transition, so they cannot hit the window"), and `9-7_n/9-8_n` ("fail at once and run without a timeout"). The gtest doc-comment on `liveVideoNoFramerateCompletes_p` gained the matching sentence. Neither is described as window/regression coverage anymore.
- **Point 2** (drop the timeout on `9-7_n`/`9-8_n`): done — the trailing `30` argument is removed from both `gstTest` calls; the six window-coverage cases (`9-1..9-4, 9-9, 9-10`) and the two live smoke cases (`9-5, 9-6`) keep it.
- **Point 3** (PR body "23" → "25"): done — the PR description now reads "10 of the 25 gtest cases fail (the live-source and framerate cases are smoke coverage of unchanged paths and pass either way)"; no stale "23" remains in the body.
- **Point 4** (valgrind noise, informational only): correctly left as a note, no code action expected or taken.

## 2. Timeout removal verified against the real `ssat-api.sh`

Pulled `gstTest()` from `https://raw.githubusercontent.com/myungjoo/SSAT/main/ssat-api.sh` (lines 297-339). For `$4 -eq 1` (the `_n` convention), scoring is `retcode -ne 0 → desired=1 (pass)` — no distinction between a real early failure and `timeout`'s exit 124. With `$6` (the 6th arg) unset, `[[ "${6}" -gt "0" ]]` is false in both timeout-detection branches, so `gst-launch` runs with no external timeout at all. That confirms the bot's point exactly: keeping a 30s bound on `9-7_n`/`9-8_n` would have let a future stall there masquerade as the expected failure (exit 124 still satisfies `retcode -ne 0`); removing it means a future stall instead hangs the run with no timeout wrapper to convert it into a false pass — visible instead of silently green. The change does what the commit message says.

## 3. Author's "live source can't hit the window" reasoning holds up

Two independent, both-sufficient reasons, and the commit/reply correctly lean on both:

- `videotestsrc`/`audiotestsrc` stamp `GST_BUFFER_PTS` on every buffer unconditionally (confirmed already in the bot's review); `framerate=0/1` only invalidates duration. So `_gst_tensor_converter_chain_timestamp()` takes its `GST_CLOCK_TIME_IS_VALID(pts)` early-out and never reaches the racy clock branch — orthogonal to any state-transition timing.
- Independently, for `is-live=true` sources GstBaseSrc's streaming thread blocks after the single preroll buffer until its *own* PAUSED→PLAYING transition unblocks `live_cond` (`gst_base_src_set_playing`). Since `gst_bin_iterate_sorted()` commits child states sink-first, the converter (downstream of the source) has already completed its own PAUSED→PLAYING transition — and had `base_time` delivered — before the source's transition (which is last) unblocks it to push further buffers. So even hypothetically, the converter can't still be in the pre-`base_time` window by the time live post-preroll buffers arrive. This is a sound, GStreamer-semantics-grounded argument, not just an assertion.

## Verdict

No issues found in this commit. All three actionable suggestions from the prior review are implemented as described, the fourth (informational) required no action, the diff is confined to comments and one test argument in the two named test files, and the negative-case timeout removal is a real improvement in ssat-api.sh's own scoring logic, not just claimed. As of this review, CI on `804bbe81` is still running (CPP-Linter and DCO pass so far; the rest pending) — nothing in this diff is expected to affect them, but green CI should still be confirmed before merge. **Mergeable pending CI completion.**

myungjoo:

Pre-merge check requested in the approval, verified on `804bbe81` now that all 18 checks are green:

- All three SSAT-running jobs executed the converter suite with the updated `runTest.sh`. `9-7_n` and `9-8_n` are scored `[PASSED]` in each of them, and they are logged in the same second as the neighbouring `9-6` and `9-9` (GBS x86_64: all four at `[ 425s]`; pdebuild: all four within `04:40:25.728`; RISC-V: all four within `04:46:51.943`), so the negative cases fail at once without any timeout wrapper.
  - pdebuild: `[PASSED] nnstreamer_converter (42 passed among 42 cases)`
  - GBS x86_64: `[PASSED] nnstreamer_converter (42 passed among 42 cases)`
  - RISC-V: `9-1` … `9-10` all `[PASSED]`
- The minor wording nit is fixed in the PR body: the SSAT bullet now states that only the positive cases run in gstTest timeout mode and that `9-7_n`/`9-8_n` run without one.

No code change; the head stays `804bbe81`.

Issue #4894, opened 2026-08-28
[CI] Valgrind GTEST step takes ~51 min, 89% of it in unittest_filter_llamacpp and unittest_filter_armnn


myungjoo:
The `GTEST run with Valgrind on a case` step of `Minimal meson build in Ubuntu with Valgrind` (`.github/workflows/ubuntu_clean_meson_build.yml`) takes about 51 minutes, and two of the thirteen binaries account for 89% of it.

## Measurement

`packaging/run_unittests_binaries.sh --valgrind ./tests/` runs each binary under `valgrind --tool=memcheck --leak-check=full --track-origins=yes --num-callers=200`. Per-binary durations, derived from the log timestamps of two runs on the same branch that reached the same thirteen binaries:

| binary | run A | run B |
|---|---|---|
| **unittest_filter_llamacpp** | **29.1 min** | **33.6 min** |
| **unittest_filter_armnn** | **11.8 min** | **12.1 min** |
| unittest_sink | 1.0 min | 1.0 min |
| unittest_filter_litert | 1.5 min | 1.6 min |
| unittest_plugins | 0.5 min | 0.5 min |
| unittest_latency | 0.3 min | 0.3 min |
| unittest_filter_custom | 0.2 min | 0.2 min |
| unittest_rate / src_iio / watchdog / common / trainer | < 0.15 min each | < 0.15 min each |

Run A: [33043926247](https://github.com/nnstreamer/nnstreamer/actions/runs/33043926247) · Run B: [33089389421](https://github.com/nnstreamer/nnstreamer/actions/runs/33089389421). The two agree to within a few percent on every binary, so these are stable costs rather than runner noise.

`unittest_filter_llamacpp` alone is longer than the entire rest of the step put together.

## Why these two

Both wrap a third-party inference runtime, and memcheck instruments every instruction and every allocation inside it:

- `unittest_filter_llamacpp` runs actual llama.cpp token generation. The natural cost is ~2.7 s; under memcheck it is ~30 min, roughly 700x.
- `unittest_filter_armnn` runs the ArmNN CpuRef backend, which is the reference (unoptimised) path to begin with.

What memcheck reports there is overwhelmingly about llama.cpp and ArmNN internals, not about nnstreamer's subplugin code — and the suppression file already exists because of exactly that kind of noise.

## Suggestions

`run_unittests_binaries.sh` already takes `-k | --skip BINARY_NAME[,...]`, so the cheapest fix is one line in the workflow:

```
[...]

PR #4921, opened 2026-09-04
[CI] Bound the Valgrind unittest step and examine every binary

myungjoo:
The `GTEST run with Valgrind on a case` step of *Minimal meson build in Ubuntu with Valgrind* is the longest step of that job and the critical path of a pull request. It has been reported at 51 min and at 80 min. Measured now, it is bimodal, the spread is not noise, and a small set of cases explains all of it.

Addresses #4894 and the coverage half of #4888.

## What it costs

Per-case cost from the log timestamps of the x86 job, on [33728622232](https://github.com/nnstreamer/nnstreamer/actions/runs/33728622232) (a fast runner) and [33723985896](https://github.com/nnstreamer/nnstreamer/actions/runs/33723985896) (a slow one). The gtest-reported per-case times agree with the timestamps, so these are per-case costs and not log interleaving.

| | fast run | slow run |
|---|---|---|
| **step total** | **35.9 min** | **86.0 min** |
| `nnstreamerFilterLiteRT.sharedEnvInvokeDuringConfigure` | 9.9 min | 30.1 min |
| `nnstreamerFilterArmnn.invokeAdvanced` | 11.0 min | 11.2 min |
| four more LiteRT cases that recompile the model | 4.3 min | 6.5 min |
| the llama.cpp cases that generate tokens | 3.4 min | 28.5 min |
| everything else | 7.3 min | 9.7 min |

Every one of those drives a third-party inference runtime that memcheck instruments instruction by instruction: ArmNN's reference backend, a LiteRT model compile repeated until a churn thread finishes ten cycles, and llama.cpp token generation. None of the cost is this project's code.

## Bounding it

**ArmNN and LiteRT: name what to drop.** Six cases by full name, no wildcards, so a future case is not swept in silently. Every other case of both binaries still runs, which keeps each sub-plugin's open, configure and invoke paths under memcheck; what is dropped is repeated inference.

**llama.cpp: name what to keep.** Generating a token under memcheck costs about 1 s on a fast runner and about 8 s on a slow one, while the tests wait 10 s for samples. On a slow runner every case that generates anything times out and then pays for the whole generation during the NULL transition, 87 to 305 s each. That is unbounded from this step's point of view and a new test would inherit it, so the binary runs a named list instead: the cases that load the model and tear it down without generating, plus the property validation that never loads one. The lora cases stay out; the model they want is not in the tree and would be a 7B one if it were.

Nothing is lost from the job's coverage. Every excluded case still runs, without memcheck, in the `build and unit test` step of the same job.

## Result

On this branch the step takes **12.1 min** in [33850823484](https://github.com/nnstreamer/nnstreamer/actions/runs/33850823484) and **7.3 min** in [33853964542](https://github.com/nnstreamer/nnstreamer/actions/runs/33853964542), and examines **22 binaries** either way, against 36-86 min for 13 to 16 binaries before. Every check passes at the current head, `Ubuntu pdebuild` and `Tizen GBS x86_64` included; both of those failed at the first attempt's head and are the legs that proved the recursion problem.

| | binary time |
|---|---|
| `unittest_filter_litert` | 241.3 s |
| `unittest_join` | 160.1 s |
| `unittest_filter_single` | 82.8 s |
| `unittest_sink` | 57.1 s |
| `unittest_filter_llamacpp` | 28.6 s |
| `unittest_src_iio` | 25.8 s |
| the five newly reached sub-directory binaries | 32.6 s |
| the remaining nine | 75.5 s |

`unittest_filter_armnn` is 2.1 s, down from 11 min.

The step gets a 30-minute timeout to hold that budget, and is now a script block rather than one 900-character line, which is what makes room to explain the list next to it.

Two things to know about that bound. The 18 minutes of headroom are smaller than they look: the runner spread measured here is about 1.3x for the ordinary binaries but 3x for LiteRT, which is still 241 s of the total, so a slow runner plausibly lands in the low twenties. And a step timeout is the one failure this step cannot absorb with `|| echo`, so if it ever trips, the thing to change is the exclusion list rather than the timeout. Separately, gtest exits 0 when a filter name matches nothing, so renaming a case on either list drops it from memcheck with no signal; the comment above the step says so next to the lists.

Measured and rejected first: valgrind's `--fair-sched=yes`, which sounded promising because ggml spins in a thread-pool barrier that valgrind's serialised scheduler handles badly. It moves the two worst cases by 7% (LiteRT, 409 s to 380 s) and 20% (llama.cpp). `OMP_NUM_THREADS=1` does nothing; the pinned prebuilt llama.cpp is not the OpenMP build.

## Widening what it examines

The pass was examining 13 of the 21 binaries the job builds, and which 13 depended on where it stopped.

- `run_unittests_binaries.sh` searched only the top level, so `unittest_cppfilter`, `unittest_datareposrc`, `unittest_datareposink`, `unittest_tizen_custom` and `unittest_tizen_custom-set` had never run under memcheck at all. They cost 38.6 s together in CI. The deeper search is opt-in through a new `--recursive` that only this step passes, for the reason in the next section.
- The loop stopped at the first failing binary, leaving whatever came after it unexamined. In the 86-minute run that was `unittest_tensor_region`, `unittest_if` and `unittest_filter_single`, none of which had anything to do with the llama.cpp failure that stopped the pass. The list is now sorted, so what a run covers no longer depends on directory order, and the binaries that failed are named at the end.
- Meson copies the generated subplugin test sources next to the binaries, and a checkout that leaves those copies executable would hand `unittest_tizen_custom.cc` to gtest. Entries are checked for the ELF magic.

The step keeps `|| echo`, so it still cannot fail the build; that half of #4888 is deliberately left alone.

## What the earlier attempts got wrong

Three corrections came out of CI on this branch rather than out of review, and each is the reason for one piece of the current shape.

**Recursion cannot be the default.** `nnstreamer.spec` calls this same script on `./tests` and then again on each sub-directory it cares about, one at a time and with the library path each one needs. Searching sub-directories unconditionally made the Tizen/GBS build run three binaries it never used to, and fail on all three: `unittest_edge` without its `LD_LIBRARY_PATH`, `unittest_openvino` without OpenVINO, and `unittest_filter_reload` without the arguments it expects. Hence `--recursive`.

**Two helper programs were named like tests.** `tests/nnstreamer_repo_dynamicity` and `tests/nnstreamer_filter_reload` build GStreamer programs that an SSAT `runTest.sh` drives with arguments; meson does not register either with `test()`, but they were called `unittest_repo` and `unittest_filter_reload`. Started with no arguments, `unittest_repo` built its pipeline, waited on its main loop and took the entire step budget by itself in [33843038794](https://github.com/nnstreamer/nnstreamer/actions/runs/33843038794). The third commit renames both after their source files; nothing outside their own `meson.build` and `runTest.sh` refers to them, and no packaging file names them.

**Dropping the fourteen worst llama.cpp cases was not enough.** The five that were kept had been cheap on the runner they were measured on, and cost 841 s between them on the next one. That is what moved llama.cpp from a list of exclusions to a list of inclusions.

## Testing

`packaging/test_run_unittests_binaries.sh` pins the runner's behaviour: `--recursive` reaches a sub-directory while a plain run still does not, a failure does not stop the binaries after it, the failing ones are named, and a file that is not an ELF binary is not run. It is wired into *Static checks* with the same merge-ref guard as the other self-tests, and passed there on the first run of this branch. Its fixtures name themselves in their output, so "did this one run" is observable; two of them fail, which makes the fail-fast check independent of the order the file system lists entries in. It fails against the previous script.

Everything above was also measured locally under the same memcheck flags (valgrind 3.18.1, Ubuntu 22.04): the llama.cpp list runs its six cases in 10 s, the LiteRT exclusions take exactly the named cases out and nothing else, and the modified runner walks a real build tree of 22 binaries in 277 s, reporting the four that failed.

## Not in this change

- `tensorFilterOpenvino.getTensorDim2_n` fails. The earlier unconditional sweep was the first thing ever to run `unittest_openvino`, and it found a real pre-existing failure: the test builds its map keys with `std::string ((char *) &i)`, which does not give it `NNS_TENSOR_SIZE_LIMIT + 1` distinct names, so the limit it means to exceed is not exceeded. With the search now opt-in nothing in this branch runs that binary, and fixing an OpenVINO sub-plugin test is a different subject, so it is reported here rather than folded in.
- Restoring the excluded cases in a scheduled job. This workflow has no `schedule:` trigger and its steps are gated on the `check-rebuild` action, which needs a pull-request context, so adding one is its own change.
- A static check that every `unittest_*` executable is registered with meson's `test()`. That would have caught the naming problem above before CI did. Two things it would have to handle: `unittest_mlagent` is a real gtest binary that no `test()` call registers, and `unittest_ntp_util` is registered under the test name `gstreamer_ntp_util_mock`, so the check has to key on the executable target rather than the test name. Probing each candidate with `--gtest_list_tests` instead of the ELF magic would need a `timeout` around it, because a `gst_init` helper accepts the unknown flag and walks straight into its main loop.
- Making the excluded cases cheap enough to come back: bounding `sharedEnvInvokeDuringConfigure` by wall clock the way #4875 bounded `join.padReleaseWhileStreaming`, and giving the llamacpp sub-plugin a thread-count property so it can run single-threaded under a memory checker.

### DIFF
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -188,6 +188,19 @@ jobs:
           else
             echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
           fi
+      - name: /Checker/ unittest runner self-test
+        # Runs a repository script unrelated to $changed_file_list, so it
+        # needs the same merge-ref guard as the steps above.
+        run: |
+          selftest=packaging/test_run_unittests_binaries.sh
+          if [ -f "$selftest" ]; then
+            bash "$selftest"
+          elif git show --pretty="format:" --name-only --no-renames --diff-filter=D ${{ github.event.pull_request.head.sha }} -${{ github.event.pull_request.commits }} | grep -qxF "$selftest"; then
+            echo "::error::This PR removes $selftest; remove its workflow step too, or restore the script (also update this step if the script moved)."
+            exit 1
+          else
+            echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
+          fi
       - name: /Checker/ SSAT install list is in sync
         # Same guard as the check-rebuild self-test above: the step list comes
         # from the merge ref while the checkout is the PR head, so a branch that
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -210,9 +210,67 @@ jobs:
     - name: SSAT run on filter-ncnn
       if: env.rebuild == '1'
       run: if [ '${{ matrix.os }}' == 'ubuntu-22.04' ]; then export NNSTREAMER_BUILD_ROOT_PATH=`pwd`/build && export NNSTREAMER_FILTERS=`pwd`/build/ext/nnstreamer/tensor_filter && export NNSTREAMER_DECODERS=`pwd`/build/ext/nnstreamer/tensor_decoder && export NNSTREAMER_CONVERTERS=`pwd`/build/ext/nnstreamer/tensor_converter && export GST_PLUGIN_PATH=`pwd`/build/gst && export NNSTREAMER_CONF=`pwd`/build/nnstreamer-test.ini && pushd tests/nnstreamer_filter_ncnn && ssat -n -p=1 --summary summary.txt && popd; fi
+    # This memcheck pass is the longest step of the job, and what it costs is
+    # decided almost entirely by a few cases that drive a third-party inference
+    # runtime, every instruction of which memcheck instruments. Two runners
+    # measured, 33728622232 (fast) and 33723985896 (slow), per case:
+    #
+    #   nnstreamerFilterArmnn.invokeAdvanced                    11.0 / 11.2 min
+    #   nnstreamerFilterLiteRT.sharedEnvInvokeDuringConfigure    9.9 / 30.1 min
+    #   four more LiteRT cases that compile the model repeatedly 4.3 /  6.5 min
+    #
+    # Those are excluded by name below. Every other LiteRT and ArmNN case still
+    # runs, so each sub-plugin's open, configure and invoke paths stay covered;
+    # what is dropped is repeated inference, whose cost is the runtime's and
+    # not this project's. They are written out one at a time on purpose: a
+    # wildcard would silently take future cases with it.
+    #
+    # llama.cpp is handled the other way round, by naming the cases to keep.
+    # Generating a token under memcheck costs about 1 s on a fast runner and
+    # about 8 s on a slow one, while the tests wait 10 s for samples, so on a
+    # slow runner every case that generates anything at all times out and then
+    # pays for the whole generation in the NULL transition: 87 to 305 s each,
+    # measured in 33847412951. That is unbounded from this step's point of
+    # view, and a new test would inherit it, so the safe default is to leave a
+    # case out until it is known to be cheap. The ones kept here load the model
+    # and tear it down without generating; the lora cases stay out because the
+    # model they need is not in the tree and would be a 7B one if it were.
+    #
+    # Either way gtest exits 0 when a name matches nothing, so renaming a case
+    # on one of these lists takes it out of memcheck silently; keep them in
+    # step with the test sources.
+    #
+    # The excluded cases all still run, without memcheck, in the "build and
+    # unit test" step above. See
+    # https://github.com/nnstreamer/nnstreamer/issues/4894 for the measurements
+    # and for restoring them in a job that can afford them.
     - name: GTEST run with Valgrind on a case
-      if: env.rebuild == '1'
-      run: if [ '${{ matrix.os }}' == 'ubuntu-22.04' ]; then export NNSTREAMER_BUILD_ROOT_PATH=`pwd`/build && export NNSTREAMER_FILTERS=`pwd`/build/ext/nnstreamer/tensor_filter && export NNSTREAMER_DECODERS=`pwd`/build/ext/nnstreamer/tensor_decoder && export NNSTREAMER_CONVERTERS=`pwd`/build/ext/nnstreamer/tensor_converter && export GST_PLUGIN_PATH=`pwd`/build/gst && export NNSTREAMER_CONF=`pwd`/build/nnstreamer-test.ini && G_SLICE=always-malloc G_DEBUG=gc-friendly ./packaging/run_unittests_binaries.sh --valgrind ./tests/ || echo "There are Valgrind errors. Please fix it. As we have a lot of Valgrind errors from different libraries and possible from nnstreamer itself, we are not halting the build with Valgrind errors until we get them all."; fi
+      if: env.rebuild == '1' && matrix.os == 'ubuntu-22.04'
+      timeout-minutes: 30
+      run: |
+        export NNSTREAMER_BUILD_ROOT_PATH=`pwd`/build
+        export NNSTREAMER_FILTERS=`pwd`/build/ext/nnstreamer/tensor_filter
+        export NNSTREAMER_DECODERS=`pwd`/build/ext/nnstreamer/tensor_decoder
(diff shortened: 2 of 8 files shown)


myungjoo:

**Re-review at head `ab435845`** (3 commits, unchanged from `02825f56` except two small follow-ups to the APPROVED review's non-blocking notes). I re-verified both against the tree rather than taking the diff at face value.

## The two follow-up changes, checked

`git diff 02825f56 ab435845` touches exactly two files, and nothing else:

1. **`.github/workflows/ubuntu_clean_meson_build.yml`** — a 3-line comment addition ("gtest exits 0 when a name matches nothing, so renaming a case on one of these lists takes it out of memcheck silently") appended to the exclusion-list comment block. This lands in the **first** commit (`bfba3b84`), which is where that comment block itself was introduced — the right commit for it, not bolted onto an unrelated one.
2. **`packaging/run_unittests_binaries.sh`** — `[ -n "${RECURSIVE}" ] && depth_args=""` became an explicit `if/else`. This lands in the **second** commit (`43dcba50`), which is where `RECURSIVE`/`depth_args` was introduced — again the logically correct commit. Behavior is unchanged (same depth_args value either way); this only removes the previously-noted trap where the `&&` form leaves a non-zero `$?` when the branch is not taken, which would have bitten a future `set -e`.

Both changes are exactly and only what the task description says they should be — no unrelated edits rode along, and the third commit (`ab435845`, the SSAT-helper renames) is untouched by either follow-up, so the commit boundaries stay logically clean (comment lives with the exclusion lists it documents, depth_args fix lives with the flag it implements, renames stay isolated).

## PR body vs. code

The body's added note ("the 18 minutes of headroom are smaller than they look") is consistent with the code: `timeout-minutes: 30` on both the Valgrind step (line 249) and the example-app step (line 289), against the 12.1 min measured total — an 18-minute margin, as stated. No mismatch between prose and implementation.

## Independent spot-check

I read the full current `packaging/run_unittests_binaries.sh` and the workflow step end-to-end (not just the diff) to confirm the two changes compose correctly with the rest of the script: the `if/else` produces the same `depth_args` in both branches as the old one-liner, and the new comment sentence is accurate — `GTEST_FILTER` matching nothing is indeed silent in gtest, which is exactly the failure mode both this and the prior review flagged as a residual, accepted risk (not something this PR needs to fix, since a future rename triggering it would need its own detector, out of scope here).

I attempted to run `packaging/test_run_unittests_binaries.sh` locally to confirm it still passes with the `if/else` change, but my environment is Windows/Git-Bash, where `echo`/`env` are PE binaries, not ELF — so `is_test_binary`'s ELF-magic check rejects every fixture and the self-test fails for an environment reason unrelated to this PR (confirmed via `od -A n -N 4 -t x1`, which shows `4d 5a` = "MZ", not `7f 45 4c 46`). This is not a finding against the PR; it only means I could not independently re-run the self-test outside CI. The logic change itself (same `depth_args` value, no behavioral difference) does not require re-running it to verify.

## CI at this head

Not yet available — all checks are still `pending` at `ab435845` at the time of this review (`gh pr checks 4921`), including *Static checks* (which runs the self-test) and the two package-build legs (`Ubuntu pdebuild`, `Tizen GBS x86_64`) that the first CHANGES_REQUESTED review used to prove the recursion regression. Given the two follow-ups are comment-only and a mechanical `&&`→`if/else` rewrite with no behavioral change, I would not expect them to affect any of those legs, but that is not yet confirmed by the run itself.

## Scope, tests, docs

Proportionate: the two follow-ups are a comment and a control-flow style fix, nothing else. No new test is needed for either — the comment documents existing, already-covered behavior, and the `if/else` rewrite is behaviorally identical to what the self-test's "top level is searched without --recursive" / "a sub-directory is left alone without --recursive" assertions already exercise. No API or architecture change, so no documentation is owed beyond what's already there.

## Verdict

**Approve.** Both outstanding non-blocking notes from the `APPROVED` review (nit #4 on `depth_args`, note #1 on the silent-filter-miss risk) are addressed exactly as scoped, in commits that match their subject matter, with no scope creep. The one open item is that CI for this exact head has not finished yet — worth confirming the two build-packaging legs stay green before merge, consistent with what the prior review used as its evidence bar.

Issue #4920, opened 2026-09-03
[Memory] Memory-safety audit of Tizen-built C/C++ components: tracking checklist


myungjoo:
## Summary

This is a tracking issue for memory-safety defects found by a source-level audit of the C/C++ code that is built for Tizen (`packaging/nnstreamer.spec`). Each item below is a separate, independently fixable defect. Please tick the box when the fix PR lands, and link the PR next to the item.

**Audited at:** `4e2522f7` (main, 2026-09-03).
**Method:** file-by-file static review by separate Claude agents (one per component group), followed by a manual re-verification of every listed item against the source. Line numbers refer to the commit above. Items are listed with the *type* of defect, the *trigger* (what input or configuration reaches it), and a suggested direction for the fix.

**Scope (what was audited):**
- `gst/nnstreamer/**` (core library, tensor_filter framework, all elements, tensor_query), `gst/edge`, `gst/join`, `gst/mqtt`, `gst/datarepo`
- tensor_filter sub-plugins built on Tizen: tensorflow2-lite (`tensor_filter_tensorflow_lite.cc`), nnfw, cpp, python3, lua, tvm, snpe / snpe_v1, onnxruntime, tizen-hal (+snpe/vivante), openvino, edgetpu, movidius-ncsdk2, vivante
- all tensor_decoder and tensor_converter sub-plugins, gRPC (`ext/nnstreamer/extra`, `tensor_sink_grpc`, `tensor_src_grpc`), `tensor_src_tizensensor`

**Out of scope (not built on Tizen, not audited):** armnn, caffe2, dali, deepview-rt, executorch(+llama), litert, llama2, llamacpp, mediapipe, mxnet, ncnn, pytorch, qnn, snap, tensorflow (v1), tensorrt/tensorrt10, trix-engine, android_source. Files with no findings after a full read: `gst/edge/*`, `gst/join/gstjoin.c`, `tensor_filter_lua.cc`, `tensor_filter_tvm.cc`, `tensor_filter_onnxruntime.cc`, `tensor_filter_tizen_hal*.cc`, `gsttensor_rate.c`, `gsttensor_sink.c`, `gsttensor_debug.c`, `gsttensor_mux.c`, `tensor_data.c`, `tensor_allocator.c`, `nnstreamer_conf.c`, `nnstreamer_watchdog.c`, `ml_agent.c`.

[...]

PR #4924, opened 2026-09-04
[Decoder/BoundingBox] Clamp box drawing to the output frame

myungjoo:
Fixes item **G3** of #4920: `BoundingBox::draw()` writes bounding boxes outside the RGBA output frame.

## Problem

`draw()` scales each detected box to the output surface and clamps only the right and bottom edge:

```c
x1 = (width * a->x) / i_width;
x2 = MIN (width - 1, (width * (a->x + a->width)) / i_width);   /* only x2 is clamped */
y1 = (height * a->y) / i_height;
y2 = MIN (height - 1, (height * (a->y + a->height)) / i_height);
```

`width` and `height` are `guint`, so the products are evaluated as unsigned:

* **A box that starts left of / above the frame** (`ov-person-detection` passes the model's coordinates through without any clamp, and OpenVINO's person-detection models do emit slightly negative corners) turns into a huge offset: for `width = 640, a->x = -1` the left edge becomes `6710885`. The vertical edge loop then writes `PIXEL_VALUE` tens of megabytes past the frame. Locally this segfaults.
* **A box entirely right of / below the frame** keeps an in-range partner coordinate, so one of the two edge loops still runs and writes outside the frame - or, when the offset happens to wrap into a later row, paints a line where no box is. No decoder clamps the box against the *upper* bound, so any model whose output extends past its input reaches this.
* **The label sprite** always draws 13 rows, which do not fit an output frame shorter than one character cell.

The rotated-box path in the same function (`draw_line()`) already range-checks every pixel; only the rectangle path was missing it.

## Fix

`ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc`:

* compute the four edges in `gint64`, which removes the unsigned promotion,
* skip boxes that do not overlap the frame or have a negative size,
* `CLAMP` the remaining edges to the frame,
* draw the vertical edges by index instead of by pointer, and stop the label sprite at the last row.

Boxes that already fitted inside the frame are scaled to exactly the same pixels as before, so no golden output changes.

## Tests

`tests/nnstreamer_decoder_boundingbox/unittest_decoder_boundingbox.cc` (new, registered in `tests/meson.build`) drives the decoder through a pipeline with `ov-person-detection` output - the only in-tree mode that emits unclamped coordinates - and asserts on the drawn pixels. Against the unfixed decoder:

| case | unfixed | fixed |
| --- | --- | --- |
| `drawBoxInFrame` | pass (behaviour preserved) | pass |
| `drawBoxAcrossLeftEdge` | **SIGSEGV** | pass |
| `drawBoxAcrossBottomEdge` | pass (behaviour preserved) | pass |
| `skipBoxRightOfFrame` | **fail**, 46 stray pixels | pass |
| `skipBoxBelowFrame` | **fail**, 33 stray pixels | pass |
| `skipInvertedBox_n` | pass (behaviour preserved) | pass |

`tests/nnstreamer_decoder_boundingbox/runTest.sh` gains case 12: mobilenet-ssd rendering labels into a 160x12 frame. That directory is run under valgrind with a fatal gate in `ubuntu_clean_meson_build.yml`, and the case is what catches the sprite overrun:

```
Invalid write of size 4
   at BoundingBox::draw(GstMapInfo*, _GArray*) (tensordec-boundingbox.cc:832)
   by BoundingBox::decode(...) (tensordec-boundingbox.cc:1107)
ERROR SUMMARY: 34 errors from 1 contexts     <- unfixed
ERROR SUMMARY: 0 errors from 0 contexts      <- fixed
```

## Verification

* `ssat` in `tests/nnstreamer_decoder_boundingbox`: 51/51 pass, every golden (mobilenet-ssd, ssd-postprocess, mp-palm-detection, yolov5 incl. track mode, yolov8, yolov10, yolov8-obb) byte-identical - which is the regression check for the other decoder modes that feed `draw()`.
* `meson test`: 22/23, the only failure being `unittest_filter_python3`, a numpy 1.x/2.x mismatch in the local environment.
* `clang-format` clean; the changed objects recompile without warnings under `-Wall -Wextra -Werror` with `-Wno-sign-compare` stripped.

`draw()` has a single caller and the header is shared only with the `box_properties/` decoders, which do not call it, so nothing outside this decoder is touched.

Not addressed here, and left as their own items of #4920: G2 (the signed `char` label index in the same loop) and G4 (`logBoxes()` indexing `labels[]` with an unchecked class id).

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
+++ b/ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc
@@ -776,13 +776,23 @@ BoundingBox::draw (GstMapInfo *out_info, GArray *results)
       }
     } else {
       int x1, x2, y1, y2; /* Box positions on the output surface */
+      gint64 bx1, bx2, by1, by2;
       int j;
       uint32_t *pos1, *pos2;
       /* 1. Draw Boxes */
-      x1 = (width * a->x) / i_width;
-      x2 = MIN (width - 1, (width * (a->x + a->width)) / i_width);
-      y1 = (height * a->y) / i_height;
-      y2 = MIN (height - 1, (height * (a->y + a->height)) / i_height);
+      bx1 = ((gint64) width * a->x) / i_width;
+      bx2 = ((gint64) width * ((gint64) a->x + a->width)) / i_width;
+      by1 = ((gint64) height * a->y) / i_height;
+      by2 = ((gint64) height * ((gint64) a->y + a->height)) / i_height;
+
+      if (bx2 < bx1 || by2 < by1 || bx2 < 0 || by2 < 0 || bx1 >= (gint64) width
+          || by1 >= (gint64) height)
+        continue;
+
+      x1 = (int) CLAMP (bx1, 0, (gint64) width - 1);
+      x2 = (int) CLAMP (bx2, 0, (gint64) width - 1);
+      y1 = (int) CLAMP (by1, 0, (gint64) height - 1);
+      y2 = (int) CLAMP (by2, 0, (gint64) height - 1);
 
       /* 1-1. Horizontal */
       pos1 = &frame[y1 * width + x1];
@@ -795,13 +805,9 @@ BoundingBox::draw (GstMapInfo *out_info, GArray *results)
       }
 
       /* 1-2. Vertical */
-      pos1 = &frame[(y1 + 1) * width + x1];
-      pos2 = &frame[(y1 + 1) * width + x2];
       for (j = y1 + 1; j < y2; j++) {
-        *pos1 = PIXEL_VALUE;
-        *pos2 = PIXEL_VALUE;
-        pos1 += width;
-        pos2 += width;
+        frame[j * width + x1] = PIXEL_VALUE;
+        frame[j * width + x2] = PIXEL_VALUE;
       }
 
       /* 2. Write Labels + tracking ID */
@@ -817,7 +823,6 @@ BoundingBox::draw (GstMapInfo *out_info, GArray *results)
 
         label_len = label ? strlen (label) : 0;
 
-        /* x1 is the same: x1 = MAX (0, (width * a->x) / i_width); */
         y1 = MAX (0, (y1 - 14));
         pos1 = &frame[y1 * width + x1];
         for (k = 0; k < label_len; k++) {
@@ -825,7 +830,7 @@ BoundingBox::draw (GstMapInfo *out_info, GArray *results)
           if ((x1 + 8) > (int) width)
             break; /* Stop drawing if it may overfill */
           pos2 = pos1;
-          for (y2 = 0; y2 < 13; y2++) {
+          for (y2 = 0; y2 < 13 && (y1 + y2) < (int) height; y2++) {
             /* 13 : character height */
             for (x2 = 0; x2 < 8; x2++) {
               /* 8: character width */
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -103,6 +103,15 @@ if gtest_dep.found()
     )
     test('unittest_tensor_region', unittest_tensor_region, env: testenv)
 
+    # RUN unittest_decoder_boundingbox
+    unittest_decoder_boundingbox = executable('unittest_decoder_boundingbox',
+      join_paths('nnstreamer_decoder_boundingbox', 'unittest_decoder_boundingbox.cc'),
+      dependencies: [nnstreamer_unittest_deps],
+      install: get_option('install-test'),
+      install_dir: unittest_install_dir
+    )
+    test('unittest_decoder_boundingbox', unittest_decoder_boundingbox, env: testenv)
+
     # Run unittest_plugins
     unittest_plugins = executable('unittest_plugins',
       join_paths('nnstreamer_plugins', 'unittest_plugins.cc'),
--- a/tests/nnstreamer_decoder_boundingbox/runTest.sh
+++ b/tests/nnstreamer_decoder_boundingbox/runTest.sh
@@ -160,6 +160,12 @@ callCompareTest yolov8_obb_decoder_result_golden.raw yolo11n-obb_result.log "10 
 
 rm dota8-obb-label.txt
 
+# The label sprite must stay inside an output frame shorter than a character
+
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} tensor_mux name=mux ! tensor_decoder mode=bounding_boxes option1=mobilenet-ssd option2=coco_labels_list.txt option3=box_priors.txt option4=160:12 option5=300:300 ! video/x-raw,format=RGBA ! filesink location=mobilenetssd_short_output.log  multifilesrc name=fs1 location=mobilenetssd_tensors.0.%d start-index=$CASESTART stop-index=$CASEEND caps=application/octet-stream ! tensor_converter input-dim=4:1:1917:1 input-type=float32 ! mux.sink_0  multifilesrc name=fs2 location=mobilenetssd_tensors.1.%d start-index=$CASESTART stop-index=$CASEEND caps=application/octet-stream ! tensor_converter input-dim=91:1917:1 input-type=float32 ! mux.sink_1  " 12 0 0 $PERFORMANCE
+
+rm mobilenetssd_short_output.log
+
 # negative case for box properties
 
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=10 ! video/x-raw,format=RGB,width=224,height=224,framerate=0/1 ! videoconvert ! tensor_converter ! tensor_decoder mode=bounding_boxes option1=wrong_mode_name ! fakesink " 11_n 0 1
--- a/tests/nnstreamer_decoder_boundingbox/runTest.sh
+++ b/tests/nnstreamer_decoder_boundingbox/runTest.sh
@@ -160,15 +160,15 @@ callCompareTest yolov8_obb_decoder_result_golden.raw yolo11n-obb_result.log "10 
 
 rm dota8-obb-label.txt
 
+# negative case for box properties
+
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=10 ! video/x-raw,format=RGB,width=224,height=224,framerate=0/1 ! videoconvert ! tensor_converter ! tensor_decoder mode=bounding_boxes option1=wrong_mode_name ! fakesink " 11_n 0 1
+
 # The label sprite must stay inside an output frame shorter than a character
 
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} tensor_mux name=mux ! tensor_decoder mode=bounding_boxes option1=mobilenet-ssd option2=coco_labels_list.txt option3=box_priors.txt option4=160:12 option5=300:300 ! video/x-raw,format=RGBA ! filesink location=mobilenetssd_short_output.log  multifilesrc name=fs1 location=mobilenetssd_tensors.0.%d start-index=$CASESTART stop-index=$CASEEND caps=application/octet-stream ! tensor_converter input-dim=4:1:1917:1 input-type=float32 ! mux.sink_0  multifilesrc name=fs2 location=mobilenetssd_tensors.1.%d start-index=$CASESTART stop-index=$CASEEND caps=application/octet-stream ! tensor_converter input-dim=91:1917:1 input-type=float32 ! mux.sink_1  " 12 0 0 $PERFORMANCE
 
 rm mobilenetssd_short_output.log
-
-# negative case for box properties
-
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num-buffers=10 ! video/x-raw,format=RGB,width=224,height=224,framerate=0/1 ! videoconvert ! tensor_converter ! tensor_decoder mode=bounding_boxes option1=wrong_mode_name ! fakesink " 11_n 0 1
 
 rm yolov*.log
 
--- a/tests/nnstreamer_decoder_boundingbox/unittest_decoder_boundingbox.cc
+++ b/tests/nnstreamer_decoder_boundingbox/unittest_decoder_boundingbox.cc
@@ -1,0 +1,259 @@
+/**
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * @file        unittest_decoder_boundingbox.cc
+ * @date        04 Sep 2026
+ * @brief       Unit test for the bounding_boxes mode of tensor_decoder
+ * @see         https://github.com/nnstreamer/nnstreamer
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs
+ */
+
+#include <gtest/gtest.h>
+#include <glib.h>
+#include <gst/gst.h>
+#include <string.h>
+#include <unittest_util.h>
+
+#define OV_DESC_SIZE (7U)
+#define OV_DETECTION_MAX (200U)
+#define OV_TENSOR_ELEMENTS (OV_DESC_SIZE * OV_DETECTION_MAX)
+#define OV_TENSOR_SIZE (OV_TENSOR_ELEMENTS * sizeof (float))
+
+#define MODEL_WIDTH (640U)
+#define MODEL_HEIGHT (480U)
+#define OUT_WIDTH (64U)
+#define OUT_HEIGHT (48U)
+#define OUT_PIXELS (OUT_WIDTH * OUT_HEIGHT)
+
+#define BOX_PIXEL (0xFF0000FFU)
+
+/**
+ * @brief Fill the first descriptor of an ov-person-detection output tensor.
+ * @details The model emits coordinates normalized to its own input size, which
+ *          the decoder scales to the output frame. The second descriptor gets a
+ *          negative image id, which terminates the list of detections.
+ */
+static void
+setDetection (float *tensor, float x_min, float y_min, float x_max, float y_max)
+{
+  tensor[0] = 0.0f;
+  tensor[1] = 0.0f;
+  tensor[2] = 1.0f;
+  tensor[3] = x_min;
+  tensor[4] = y_min;
+  tensor[5] = x_max;
+  tensor[6] = y_max;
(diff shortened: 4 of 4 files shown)


myungjoo:

이 리뷰는 별도의 Claude 세션(리뷰 전담 에이전트)이 작성했으며, 이 코멘트에 그대로 옮겨 게시합니다. 코드 정독뿐 아니라 로컬(WSL) 빌드 트리에서 실제로 fix 적용/원복을 반복하며 유닛테스트와 SSAT 스위트를 직접 실행해 주장들을 검증했습니다.

## 검증 방법
- `ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc`의 수정 전/후 코드를 각각 빌드하여 새 GTest(`unittest_decoder_boundingbox`, 6 케이스)를 실행:
  - 수정 전 코드: `drawBoxAcrossLeftEdge`에서 실제로 **SIGSEGV** 재현 확인 (PR 설명의 주장과 일치).
  - 수정 후 코드: 6/6 전부 통과.
- `tests/nnstreamer_decoder_boundingbox` SSAT 스위트 51개 케이스 전부 재실행 → 51/51 통과, 기존 골든(mobilenet-ssd, ssd-postprocess, palm-detection, yolov5/v8/v10, yolov8-obb 등) byte-identical 유지 확인. 새로 추가된 case 12도 통과.
- `.github/workflows/ubuntu_clean_meson_build.yml`을 확인한 결과, "SSAT run with Valgrind on decoder-bounding-box" 스텝은 (GTEST valgrind 스텝과 달리) `|| echo ...` 폴백이 없어 **valgrind 에러 발생 시 잡을 실패시키는 fatal gate**가 맞음을 확인. 실제로 이 PR의 CI에서도 해당 job이 `ubuntu-22.04-arm`에서 이미 pass함(x86은 실행 중)을 `gh pr checks`로 확인.
- `clang-format --style=file -output-replacements-xml`로 변경된 두 파일(`tensordec-boundingbox.cc`, `unittest_decoder_boundingbox.cc`) 모두 포맷 변경 없음 확인.
- `draw()`가 동일 파일 내 단일 호출자(`decode()`)만 가지며, 헤더는 `box_properties/`의 다른 디코더들이 참조하지 않음을 grep으로 확인 → 아키텍처/공개 API 영향 없음, 문서 수정 불필요하다는 PR의 주장이 타당함.

## 종합 평가

**Blocking(치명) 이슈: 없음.**

수정 로직 자체가 정확합니다: `width`/`height`(guint)를 `gint64`로 승격시켜 unsigned wraparound를 제거하고, 박스가 프레임과 전혀 겹치지 않거나(`bx2<0`, `by2<0`, `bx1>=width`, `by1>=height`) 크기가 음수인 경우(`bx2<bx1`, `by2<by1`) 그리기 자체를 skip하며, 나머지는 `CLAMP`로 자릅니다. `bx1<=bx2`가 skip 조건에서 이미 보장되므로 클램프 후에도 `x1<=x2`가 유지되어 가로 루프(`for j=x1..x2`)가 음수 반복을 하지 않습니다. 라벨 스프라이트도 `(y1+y2) < height` 조건을 추가해 캐릭터 높이(13행)가 짧은 출력 프레임을 넘어 쓰는 것을 방지합니다.

## Minor / informational (non-blocking)

1. **(낮음) width==0 또는 height==0인 극단적 케이스의 잔존 결함.** `CLAMP(bx1, 0, (gint64) width - 1)`에서 `width==0`이면 `high = -1 < low = 0`이 되어, GLib `CLAMP` 매크로 정의상 `x1`이 `-1`(음수 배열 인덱스)이 될 수 있습니다. 다만 이는 기존 코드에서도 `width - 1`이 `guint` 언더플로로 `UINT_MAX`가 되어 오히려 더 나쁜 상황(거대한 양수 오프셋)이었으므로, 이번 수정으로 새로 생긴 회귀는 아니며 0×0 출력 caps 자체가 GStreamer 협상 단계에서 사실상 발생하지 않을 조건이라 실전 영향은 매우 낮습니다. 후속 아이템(G2/G4)과 함께 별도 이슈로 처리해도 무방해 보입니다.
2. **(정보성) 라벨 스프라이트 회귀 방지가 valgrind-gated SSAT job 하나에만 의존.** 새로 추가된 SSAT case 12(`option4=160:12`)는 라벨 오버런을 valgrind 없이 일반 실행할 경우 단순히 파이프라인이 EOS로 끝나는지만 확인하고(diff 비교 없음) 실제 메모리 오버플로우 자체는 잡지 못합니다. 보호는 전적으로 `ubuntu_clean_meson_build.yml`의 valgrind 스텝(위에서 fatal임을 확인)에 의존합니다. 현재는 정상 동작하지만, 해당 CI job이 비활성화/스킵되거나 matrix 조건(`ubuntu-22.04`)에서 제외되면 이 회귀를 놓칠 수 있는 단일 지점(SPOF) 구조입니다. GTest 쪽에서는 (라벨 없이) box-clamp 부분만 검증하고 있어 라벨 오버런에 대한 결정론적/비-valgrind 유닛 테스트는 없습니다. 필수는 아니지만, 여유가 있다면 라벨 오버런도 유닛테스트에서 픽셀 비교로 잡을 수 있으면 더 견고해집니다.

## 결론 (Verdict)

**현재 상태로 merge해도 문제없다고 판단합니다 (OK to merge as-is).** 주제(#4920 G3: 프레임 밖 그리기로 인한 out-of-bounds write) 해결이 정확하고, 다른 디코더 모드에 대한 회귀가 없음을 골든 테스트로 직접 확인했으며, 변경 범위도 대상 파일 하나 + 테스트로 적절히 제한되어 있습니다. 위 2건은 모두 non-blocking 관찰 사항입니다.

myungjoo:

리뷰 감사합니다. 두 건 모두 검토했고, 결론은 "코드 변경 없음"입니다. 근거를 남깁니다.

**1. `width == 0` / `height == 0`에서 `CLAMP(..., 0, width - 1)`이 `-1`을 반환할 수 있다 (낮음)**

이 경로는 도달하지 않습니다. `width == 0`이면 클램프 이전의 스킵 조건에서 걸립니다.

```c
bx1 = ((gint64) width * a->x) / i_width;   /* width == 0 이면 피연산자가 0 이므로 bx1 == 0 */
...
if (... || bx1 >= (gint64) width || by1 >= (gint64) height)
  continue;                                 /* 0 >= 0 -> skip */
```

`a->x`의 부호와 무관하게 `width`가 0이면 곱이 0이라 `bx1 == 0`이고, `bx1 >= width`(= 0)가 참이 되어 `continue`합니다. `height == 0`도 `by1 >= height`로 동일하게 걸립니다. 즉 `CLAMP`은 항상 `width >= 1`, `height >= 1`, 곧 `low <= high`인 상태에서만 실행됩니다. 지적하신 대로 협상 단계에서 0x0 caps가 사실상 만들어지지 않는 것과 별개로, 코드 자체로도 닫혀 있어 별도 가드를 넣으면 죽은 코드가 됩니다.

(다만 이 주변에서 실제로 남아 있는 결함은 따로 있습니다: `option5`가 없거나 파싱에 실패하면 `i_width`/`i_height`가 0이 되어 위 나눗셈이 0으로 나누기(SIGFPE)가 됩니다. #4920에 없는 항목이라 이 PR 범위에 넣지 않았고, 별도 항목으로 제기하겠습니다.)

**2. 라벨 스프라이트 회귀 방지가 valgrind SSAT job 하나에만 의존 (정보성)**

맞는 지적이고, 의도된 한계입니다. 스프라이트 오버런은 프레임 버퍼 *바깥*에 쓰는 동작이라 픽셀 비교로는 원리적으로 관측할 수 없습니다. 출력 높이가 13행보다 짧을 때 수정 전 코드는 0..12행을, 수정 후 코드는 0..height-1행을 그리는데, 두 결과는 프레임 안에서 완전히 동일합니다(차이는 전부 할당 밖). 행 오버런은 열 오버런과 달리 다음 행으로 감기지도 않습니다. 그래서 결정론적 유닛 테스트 대신 메모리 검사기가 유일한 관측 수단이고, 이 디렉터리의 SSAT valgrind 스텝이 fatal gate인 점을 확인하고 그쪽에 케이스를 넣었습니다.

반대로 박스 클램프 쪽(이 PR의 주된 결함)은 valgrind 없이도 잡히도록 만들었습니다 — `skipBoxRightOfFrame`/`skipBoxBelowFrame`은 수정 전 코드에서 프레임 *안쪽*으로 감겨 들어온 유령 픽셀 46개/33개를 세어 실패하고, `drawBoxAcrossLeftEdge`는 SIGSEGV로 실패합니다. 즉 SPOF는 스프라이트 한 줄에 국한됩니다.

myungjoo:
[Automated follow-up review by a separate Claude agent, relayed to this PR]

Checked both rebuttals against the actual code (`ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc`, current HEAD of this branch), no code was changed since the first review.

**1. `width == 0` / `height == 0` cannot reach `CLAMP` — correct.**

`detectedObject::x/y/width/height` are plain `int` (see `tensordec-boundingbox.h`), not unsigned, so `(gint64) width * a->x` is ordinary signed 64-bit arithmetic — the sign of `a->x` doesn't matter here. When `width == 0` the product is `0` regardless of `a->x`, so `bx1 = 0 / i_width == 0` (for any nonzero `i_width`), and the very next line, `bx1 >= (gint64) width` i.e. `0 >= 0`, is true, so the box is skipped via `continue` before `CLAMP` is ever evaluated. Symmetric argument for `height == 0` via `by1 >= (gint64) height`. Since the two conditions are OR'd, either one alone skips the box, so mixed cases (`width == 0, height > 0` or vice versa) are also covered. I don't see a path through this block where `CLAMP(..., 0, dim - 1)` executes with `dim == 0`; adding a guard here would be dead code, as the author says.

One caveat worth flagging (not part of the two points, and the author already called it out in their reply as a separate, unfixed issue): this reasoning assumes `i_width`/`i_height` are nonzero — if a decoder mode leaves them at 0 (e.g. missing/unparsed `option5`), the division itself is a signed integer divide-by-zero (SIGFPE) before the skip check ever runs. That's a real but distinct defect outside G3/this PR's scope, and I agree it shouldn't block this change.

**2. Deterministic in-process test for the label-sprite row overrun — reasoning is correct, "impossible" slightly overstated.**

Traced the sprite loop (`pos2 = pos1; for (y2 = 0; y2 < 13 && (y1+y2) < height; y2++) { ...; pos2 += width; }`, writing `*(pos2 + x2)`): addressing is strictly linear row-major (`row * width + col`), no modulo/wraparound, so any row with `row >= height` lands at or past index `height * width` — i.e. strictly past the last valid pixel of the frame, never coinciding with an earlier in-bounds row. That means the in-bounds frame content is byte-identical with and without the `(y1+y2) < height` guard, so a pixel-comparison test structurally cannot observe the bug. This part of the rebuttal checks out.

However, "only a memory checker can observe it" is a bit stronger than necessary: a guard-page allocation (mmap a page-aligned buffer sized exactly to the frame, `mprotect` the following page `PROT_NONE`) backing the output `GstBuffer`, paired with a GTest death test, would deterministically fault on the overrun without valgrind — and this repo already uses that pattern (`EXPECT_DEATH` in `tests/nnstreamer_plugins/unittest_plugins.cc` and `tests/nnstreamer_filter_tvm/unittest_filter_tvm.cc`). That said, wiring it up here would mean bypassing the normal buffer-pool/allocator path the current pipeline-based test harness relies on, which is materially more work than the existing pixel-diff tests for comparatively little marginal safety once the valgrind SSAT gate is confirmed fatal (as the first review verified). So the practical conclusion — no code/test change required for this PR — still holds; I'd just phrase the limitation as "impractical without reworking the test harness" rather than "impossible in principle."

**Verdict:** Both rebuttals hold up under closer inspection (with the minor phrasing nuance on #2 above). No code change is warranted for either point. OK to merge as-is.

myungjoo:

All three points are addressed in c9951503 (test-only; the fix commit is untouched).

**1. Untested `bx2 < 0` / `by2 < 0` branches.** Added `skipBoxLeftOfFrame` and `skipBoxAboveFrame`. You were right that this is the worst case of the old code, and both new cases are real regression tests — against the unfixed decoder they **SIGSEGV** (the huge unsigned left edge is dereferenced by the vertical edge loop before `MIN` ever matters):

```
skipBoxLeftOfFrame   rc=139
skipBoxAboveFrame    rc=139
```

**2. Label sprite had no gtest coverage.** Added `drawLabelInShortFrame`, which uses `mobilenet-ssd-postprocess` — the simplest labelled mode: number, class, score and box come from four small tensors, so no prior file or fixture is needed and the box position is exact. It renders one labelled box into a 64x12 frame, i.e. shorter than the 13-row character cell, and asserts both the clamped box edges and that label pixels were drawn above the box.

This does not turn the overrun into an assertion failure, and it cannot: the rows that do not fit are written past the end of the frame, and rows are linear, so the readable pixels are identical either way. What it does do is remove the single point of enforcement you flagged — the sprite loop is now reached by the gtest binary as well, so the gtest valgrind step sees it, not only the SSAT one. On the unfixed decoder, under valgrind:

```
Invalid write of size 4
   at BoundingBox::draw(GstMapInfo*, _GArray*) (tensordec-boundingbox.cc:832)
ERROR SUMMARY: 4 errors from 1 contexts    <- unfixed
ERROR SUMMARY: 0 errors from 0 contexts    <- fixed
```

One incidental finding while writing it: the label sprite paints its blank pixels too, so it erases the part of the box outline it covers. That is pre-existing behaviour, not something this PR changes; the assertions read the box edge outside the label's columns.

**3. Nits.** SSAT case 12 now sits after `11_n`, so the IDs are ordered. The `in_file` leak is gone — both temp names are released on every path, and the shared pipeline plumbing moved into `runAndReadFrame ()` so the second decode helper does not duplicate it.

Re-verified locally: 9/9 gtest cases pass, SSAT `nnstreamer_decoder_boundingbox` 51/51 with all goldens byte-identical, `meson test` 22/23 (the one failure is a local numpy 1.x/2.x environment mismatch in `unittest_filter_python3`), clang-format clean, doxygen check clean, and the test object recompiles warning-free under `-Wall -Wextra -Werror` with `-Wno-sign-compare` stripped.

myungjoo:

This pass focuses only on the newest commit, c9951503 ("[Test/BoundingBox] Cover the skipped-box branches and the label sprite"), which is test-only and was written to address the three points from the most recent formal review (pullrequestreview-5111717489). The fix commit itself (aff98c398e6f96dfebcfc9a06afa701c842d2a21) is unchanged and was not re-litigated.

## Verification method
- Synced the two changed files into a local Linux build tree (WSL), built `libnnstreamer_decoder_bounding_boxes.so` and `unittest_decoder_boundingbox` against the current HEAD, and ran the full suite: **9/9 pass**, including the three new cases.
- Temporarily swapped in the pre-fix decoder source (`aff98c39^`) and re-ran just the three new cases against it:
  - `skipBoxLeftOfFrame` → **SIGSEGV (rc=139)**
  - `skipBoxAboveFrame` → **SIGSEGV (rc=139)**
  - `drawLabelInShortFrame` → **passes even on the unfixed decoder** (expected — explained below)
- Independently re-derived the coordinate math for `decodeLabelledBox` (option4=64:12, option5=640:480, box=[0.5,0.25,0.75,0.5] in y_min/x_min/y_max/x_max): box clamps to x=[16,32], y=[6,9]; label `y1 = MAX(0, y1-14) = 0`. Traced the actual glyph bitmap for `'X'` in `tensordec-font.c` (index 56) through `initSingleLineSprite`'s row inversion (`v[i][12-j][k] = r[ch][j]`) and confirmed it has lit pixels in sprite rows 0–5, so the `label_pixels > 0` assertion over rows 0–5 is not vacuous.
- Confirmed `MobilenetSSDPP`'s default tensor mapping (`locations=3, classes=1, scores=2, num=0`, from `mobilenetssdpp.cc`) matches the `mux.sink_0..3` wiring used in the new pipeline, so the test correctly relies on defaults without needing `option3`.
- Ran `clang-format --style=file -output-replacements-xml` on the modified test file: 0 replacements (matches the author's claim).
- Diffed `runTest.sh`: confirmed case 12 now textually follows `11_n`, a pure reorder with no content change.
- Read `writeTensorFile`/`runAndReadFrame`/`decodeLabelledBox`: cleanup (`removeTempFile`) is now called unconditionally for every temp file on every path, closing the leak the prior review flagged.

## Assessment

**No blocking issues.** All three points from the prior review are addressed correctly and completely:

1. **Untested `bx2 < 0` / `by2 < 0` branches** — `skipBoxLeftOfFrame` and `skipBoxAboveFrame` are real regression tests: verified by direct build+run that they SIGSEGV on the pre-fix decoder and pass cleanly on the fix.
2. **No coverage for the label-sprite clamp** — `drawLabelInShortFrame` now reaches the sprite loop (previously unreached by any gtest case, since `ov-person-detection` never sets a label). As both the author and the prior review agreed, this can only pin "a label was drawn" via pixel comparison, not the row overrun itself (the overrun is strictly past the last valid pixel, so in-bounds content is byte-identical with and without the guard). I confirmed this experimentally: the test passes on both the fixed and unfixed decoder. That is expected, not a defect — its value is that the sprite loop is now also reached by this binary's own valgrind memcheck run in CI, not only by the SSAT valgrind step, per the author's rebuttal.
3. **Nits (case ordering, temp-file leak)** — both fixed as described; verified directly in the diff/code.

## Minor / informational (non-blocking)

- `drawLabelInShortFrame`'s label assertion (`label_pixels > 0` over the top 6 rows) is a coarse "something was drawn" check rather than a golden/pixel-exact comparison. That's a reasonable and proportionate choice for a smoke-style regression test — pinning exact glyph pixels would make the test brittle to any future change to `tensordec-font.c` — but it is worth knowing this test cannot itself detect a change to *where* or *how* the label is positioned, only whether the sprite path still executes without crashing.
- This is genuinely test-only: `git show --stat c9951503` touches only `runTest.sh` and `unittest_decoder_boundingbox.cc`. No production code, no other module, no docs are affected, consistent with G3 of #4920 being fully handled in the prior commit.

## Verdict

**OK to merge as-is.** The follow-up commit correctly and completely addresses all three points raised in the latest review, is proportionate in size (test-only, two files), does not touch unrelated modules, and I found no new flakiness, cwd/fixture dependency, shared-singleton interference, or vacuously-passing assertion introduced by it. The remaining coverage gap (label-sprite overrun is only observable under valgrind) is a pre-existing, explicitly acknowledged, and now more broadly gated limitation, not a new defect.

PR #4925, opened 2026-09-04
[Element] Fix property/streaming-thread races in tensor_demux and tensor_split

### LINKED ISSUE #4769: Occasional Crash When Running C# TCT for Tizen ML API
###Related Issue
* https://jira.sec.samsung.net/browse/TTEN-6759
* https://jira.sec.samsung.net/browse/TNINE-6164

### TTEN-6759 log file
```
09-10 19:42:29.520+0900 W/STDOUT   (P14928, T17412): [TUnitTest] - TC name received:[Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception]
09-10 19:42:29.572+0900 I/TUnit    (P14928, T17412): TLogger.cs: Write(32) > ##### Running Test Case [28]: Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception
09-10 19:42:29.572+0900 W/STDOUT   (P14928, T17412): TUnit##### Running Test Case [28]: Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception

...

09-10 19:42:29.574+0900 W/STDOUT   (P14928, T14928): Tizen.MachineLearning.Inference.Tests[D] | Preconditions for Tizen.MachineLearning.Inference TEST
09-10 19:42:29.574+0900 D/Tizen.MachineLearning.Inference.Tests (P14928, T14928): TLogger.cs: WriteDlog(71) > Preconditions for Tizen.MachineLearning.Inference TEST
09-10 19:42:29.577+0900 D/ml-api   (P14928, T14928): ../c/src/ml-api-inference-pipeline.c:convert_description:634: Pipeline element converted with aliases for gstreamer (Tizen element aliases): videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! videorate max-rate=1 ! tensor_converter ! tensor_mux ! tensor_demux name=demux ! tensor_sink
09-10 19:42:29.580+0900 D/ml-api   (P14928, T14928): ../c/src/ml-api-inference-pipeline.c:cb_bus_sync_message:482: The pipeline state changed from NULL to READY.
09-10 19:42:29.582+0900 D/nnstreamer (P14928, T17622): Failed to validate tensors info. the number of tensors: 0. the number of tensors should be greater than 0.

...

09-10 19:42:29.653+0900 I/CRASH_MANAGER (P17623, T17623): crash-manager.c: run(1279) > Creating report for pid 14928, tid 17622, cmdline /opt/usr/globalapps/Tizen.MachineLearning.Inference.Tests/bin/Tizen.MachineLearning.Inference.Tests.dll, pkgid Tizen.MachineLearning.Inference.Tests
09-10 19:42:29.653+0900 I/CRASH_MANAGER (P17623, T17623): crash-manager.c: execute_crash_modules(941) > Starting the minicoredumper
```

### TNINE-6164 log file
```
09-07 19:48:10.735+0900 W/STDOUT   (P12464, T14959): [TUnitTest] - TC name received:[Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception]
09-07 19:48:10.786+0900 I/TUnit    (P12464, T14959): TLogger.cs: Write(32) > ##### Running Test Case [28]: Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception
09-07 19:48:10.786+0900 W/STDOUT   (P12464, T14959): TUnit##### Running Test Case [28]: Tizen.MachineLearning.Inference.Tests.PipelineNodeInfoTests.SetValue_string_NO_Exception

...

09-07 19:48:10.787+0900 W/STDOUT   (P12464, T12464): Tizen.MachineLearning.Inference.Tests[D] | Preconditions for Tizen.MachineLearning.Inference TEST
09-07 19:48:10.787+0900 D/Tizen.MachineLearning.Inference.Tests (P12464, T12464): TLogger.cs: WriteDlog(71) > Preconditions for Tizen.MachineLearning.Inference TEST
09-07 19:48:10.793+0900 D/ml-api   (P12464, T12464): ../c/src/ml-api-inference-pipeline.c:convert_element:635: Pipeline element converted with aliases for gstreamer (Tizen element aliases): videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! videorate max-rate=1 ! tensor_converter ! tensor_mux ! tensor_demux name=demux ! tensor_sink
09-07 19:48:10.797+0900 D/ml-api   (P12464, T12464): ../c/src/ml-api-inference-pipeline.c:cb_bus_sync_message:488: The pipeline state changed from NULL to READY.
09-07 19:48:10.799+0900 D/nnstreamer (P12464, T15069): Failed to validate tensors info. the number of tensors: 0. the number of tensors should be greater than 0.
09-07 19:48:10.812+0900 D/nnstreamer (P12464, T15069): Failed to validate tensors config. framerate: -1/-1. framerate should be numerator >= 0 and denominator > 0.

...

09-07 19:48:10.893+0900 I/CRASH_MANAGER (P15184, T15184): crash-manager.c: run(1275) > Creating report for pid 12464, tid 15069, cmdline /opt/usr/globalapps/Tizen.MachineLearning.Inference.Tests/bin/Tizen.MachineLearning.Inference.Tests.dll, pkgid Tizen.MachineLearning.Inference.Tests
09-07 19:48:10.893+0900 I/CRASH_MANAGER (P15184, T15184): crash-manager.c: execute_crash_modules(937) > Starting the minicoredumper
```

## Target Test Case
* SetValue_string_NO_Exception()

myungjoo:
## Why

Issue #4769 reports an occasional crash while running the C# TCT for the Tizen ML API,
always at `PipelineNodeInfoTests.SetValue_string_NO_Exception`, on a pipeline of

```
videotestsrc ! video/x-raw,format=RGB,width=640,height=480 ! videorate max-rate=1
  ! tensor_converter ! tensor_mux ! tensor_demux name=demux ! tensor_sink
```

The crash is a NULL dereference in `gst_tensor_demux_chain()` on a streaming thread
(the reports show a tid different from the app thread, right after the NULL→READY
transition). `TizenFX`'s `Pipeline.NodeInfo.SetValue(string, string)` reaches
`ml_pipeline_element_set_property_string()`, which is a plain `g_object_set()` from the
application thread - exactly what GStreamer's threading contract says must be safe. The
element was not holding up its end.

`gst_tensor_demux_chain()` sampled `tensorpick` once before the output loop, to derive
`num_srcs` and the list cursor, and then re-read the same member *inside* the loop. On
the first buffer the gap between the two reads spans
`gst_tensor_demux_get_tensor_pad()`, which creates `src_0`, negotiates caps and
completes gst_parse's delayed link to `tensor_sink` - tens of milliseconds. A property
set landing in that window leaves the second read non-NULL while the cursor is still
NULL, and `list->data` segfaults. A set at a slightly different moment frees the strings
the cursor points at instead, turning the same window into a use-after-free.

`tensor_split` has the same shape of hazard, plus two defects that do not need any
concurrency: setting `tensorseg` a second time updated `num_tensors` but reused the
existing `GArray`, so a longer rule indexed past its end; and a rule with more than
`NNS_TENSOR_RANK_LIMIT` dimensions overflowed the `tensor_dim` it was written into.
`tensorpick` also appended to, rather than replaced, the previous selection.

## What

Two commits, one element each.

- **tensor_demux** - take a private copy of `tensorpick` under `GST_OBJECT_LOCK` at the
  top of the chain function and pass it down, so one buffer is served from one
  consistent snapshot; guard the property handlers with the same lock; fix the off-by-one
  bound check that reached `g_list_nth_data()` through a `g_assert()`; clear the member
  in `dispose()`, which GObject may run more than once; free the split pick string on the
  chain function's error exit.
- **tensor_split** - build the segment array anew on every set and swap it in under the
  lock, and let the chain function hold its own reference to it plus a copy of the pick
  list, both passed down to the pad creation; clamp the per-rule rank and say so;
  replace instead of append for `tensorpick`; refuse a `tensorseg` that asks for more
  bytes than the incoming tensor; refuse a source pad whose ordinal is past the end of
  the rule array, which replacing rather than reusing the array newly makes possible;
  unref the input buffer on the "no rule to split" exit, which was the one path that
  leaked it.

`gst_tensor_split_get_splitted()` reported its failures with `ml_logf()`, which is
`g_error()` and aborts, so its three `return NULL` paths were unreachable and the chain
function never had to handle NULL. They report with `GST_ERROR_OBJECT()` now and the
chain function ends the buffer instead of appending nothing. That also matters for the
new bounds check: with `HAVE_ORC` the copy runs through `orc_memcpy()`, where neither a
sanitizer nor valgrind sees the over-read, so refusing it is the only defence.

The pad-caps-versus-data segment mismatch found in the same function is filed separately
as #4927 - it needs a semantics decision and changes the caps existing `tensorpick`
pipelines see, so it is not folded in here.

No header, API or caps change; the element structs are untouched.

The per-buffer copy is not a new cost class - the demux chain function already splits
the pick strings once per output pad per buffer, and the split chain function only takes
a `GArray` reference plus a shallow copy of a list of integers.

## Tests

Thirteen cases added to `tests/nnstreamer_plugins/unittest_plugins.cc`. Eleven of them
fail under plain `meson test` - no valgrind or sanitizer needed - without the change they
cover, and pass with it:

| test | on `main` |
| --- | --- |
| `testTensorDemux.setTensorpickWhileAddingPad` | SIGSEGV |
| `testTensorDemux.disposeTwice` | abort (double free) |
| `testTensorSplit.setTensorsegWhileAddingPad` | SIGTRAP (glib abort) |
| `testTensorSplit.setTensorsegTwice` | SIGSEGV |
| `testTensorSplit.setTensorsegOverRank_n` | SIGABRT (heap corruption) |
| `testTensorSplit.setTensorpickTwice` | assertion failure |
| `testTensorSplit.setTensorpickWhileAddingPad` | assertion failure - no-more-pads reported after one pad, and src_1 is then added behind it |
| `testTensorSplit.pushWithoutTensorseg_n` | assertion failure - the buffer keeps a refcount of 2 |
| `testTensorSplit.pushOversizedTensorseg_n` | assertion failure - the over-read succeeds and the flow is not an error |
| `testTensorSplit.addPadBeyondTensorseg_n` | core dump - the pad ordinal reads past the rule array |
| `testTensorSplit.finalizeWithoutTensorseg` | assertion failure - releasing an unset rule array logs a critical |
| `testTensorDemux.setTensorpickTwice` | passes - locks the round-trip contract |
| `testTensorDemux.pushOutOfRangeTensorpick_n` | passes - puts a buffer on the chain function's error exit, the one that leaked the pick string |

The two `*WhileAddingPad` cases reproduce the reported window deterministically: a
`pad-added` handler parks the streaming thread inside the pad creation while the test
thread sets the property, which is the same interleaving the TCT hits by coincidence.

Locally: `unittest_plugins` 162/162; `meson test` 21/22 (the one failure,
`unittest_filter_python3`, is a NumPy 1.x/2.x ABI mismatch in the local environment and
fails identically on unmodified `main`); SSAT `nnstreamer_demux`, `nnstreamer_split`,
`nnstreamer_mux`, `nnstreamer_merge` all pass.

Fixes #4769

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_demux.c
+++ b/gst/nnstreamer/elements/gsttensor_demux.c
@@ -222,6 +222,7 @@ gst_tensor_demux_dispose (GObject * object)
 
   gst_tensor_demux_remove_src_pads (tensor_demux);
   g_list_free_full (tensor_demux->tensorpick, g_free);
+  tensor_demux->tensorpick = NULL;
   G_OBJECT_CLASS (parent_class)->dispose (object);
 }
 
@@ -260,8 +261,37 @@ gst_tensor_demux_event (GstPad * pad, GstObject * parent, GstEvent * event)
 }
 
 /**
+ * @brief Copy a tensorpick entry.
+ */
+static gpointer
+gst_tensor_demux_copy_pick (gconstpointer src, gpointer data)
+{
+  UNUSED (data);
+  return g_strdup ((const gchar *) src);
+}
+
+/**
+ * @brief Get a private copy of the tensorpick property.
+ * @param tensor_demux "this" pointer
+ * @return copied list, which should be released with g_list_free_full()
+ */
+static GList *
+gst_tensor_demux_get_tensorpick (GstTensorDemux * tensor_demux)
+{
+  GList *tensorpick;
+
+  GST_OBJECT_LOCK (tensor_demux);
+  tensorpick = g_list_copy_deep (tensor_demux->tensorpick,
+      gst_tensor_demux_copy_pick, NULL);
+  GST_OBJECT_UNLOCK (tensor_demux);
+
+  return tensorpick;
+}
+
+/**
  * @brief Get tensor config info from configured tensors
  * @param tensor_demux "this" pointer
+ * @param tensorpick private copy of the tensorpick property
  * @param config tensor config to be filled
  * @param nth source ordering
  * @param total number of tensors
@@ -269,18 +299,20 @@ gst_tensor_demux_event (GstPad * pad, GstObject * parent, GstEvent * event)
  */
 static gboolean
 gst_tensor_demux_get_tensor_config (GstTensorDemux * tensor_demux,
-    GstTensorsConfig * config, const guint nth, const guint total)
+    GList * tensorpick, GstTensorsConfig * config, const guint nth,
+    const guint total)
 {
   gst_tensors_config_init (config);
 
-  if (tensor_demux->tensorpick != NULL) {
+  if (tensorpick != NULL) {
     gchar *selected_tensor;
     gchar **strv;
     guint i, num, idx;
 
-    g_assert (g_list_length (tensor_demux->tensorpick) >= nth);
+    selected_tensor = (gchar *) g_list_nth_data (tensorpick, nth);
+    if (selected_tensor == NULL)
+      return FALSE;
 
-    selected_tensor = (gchar *) g_list_nth_data (tensor_demux->tensorpick, nth);
     strv = g_strsplit_set (selected_tensor, ":+", -1);
     num = g_strv_length (strv);
 
@@ -320,6 +352,7 @@ gst_tensor_demux_get_tensor_config (GstTensorDemux * tensor_demux,
 /**
  * @brief Checking if the source pad is created and if not, create TensorPad
  * @param tensor_demux TensorDemux Object
+ * @param tensorpick private copy of the tensorpick property
  * @param[out] created will be updated in this function
  * @param nth source ordering
  * @param total number of tensors
@@ -328,7 +361,7 @@ gst_tensor_demux_get_tensor_config (GstTensorDemux * tensor_demux,
  */
 static GstTensorPad *
 gst_tensor_demux_get_tensor_pad (GstTensorDemux * tensor_demux,
-    gboolean * created, const guint nth, const guint total)
+    GList * tensorpick, gboolean * created, const guint nth, const guint total)
 {
   GstElement *element = GST_ELEMENT_CAST (tensor_demux);
   g_autofree gchar *element_name = gst_element_get_name (element);
@@ -399,7 +432,8 @@ gst_tensor_demux_get_tensor_pad (GstTensorDemux * tensor_demux,
   gst_event_unref (event);
 
   /* configure nth pad caps */
-  if (gst_tensor_demux_get_tensor_config (tensor_demux, &config, nth, total)) {
+  if (gst_tensor_demux_get_tensor_config (tensor_demux, tensorpick, &config,
+          nth, total)) {
     caps = gst_tensor_pad_caps_from_config (pad, &config);
 
     gst_pad_set_caps (pad, caps);
@@ -412,9 +446,9 @@ gst_tensor_demux_get_tensor_pad (GstTensorDemux * tensor_demux,
     *created = TRUE;
   }
 
-  if (tensor_demux->tensorpick != NULL) {
+  if (tensorpick != NULL) {
     GST_DEBUG_OBJECT (tensor_demux, "TensorPick is set! : %dth tensor\n", nth);
-    if (g_list_length (tensor_demux->tensorpick) == tensor_demux->num_srcpads) {
+    if (g_list_length (tensorpick) == tensor_demux->num_srcpads) {
       gst_element_no_more_pads (GST_ELEMENT_CAST (tensor_demux));
     }
   }
@@ -459,11 +493,12 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
   guint num_tensors, num_srcs, i, idx;
   GstFlowReturn res = GST_FLOW_OK;
   GstTensorDemux *tensor_demux;
-  GList *list = NULL;
+  GList *tensorpick, *list = NULL;
   GstTensorInfo *_info;
 
   UNUSED (pad);
   tensor_demux = GST_TENSOR_DEMUX (parent);
+  tensorpick = gst_tensor_demux_get_tensorpick (tensor_demux);
 
   buf = gst_tensor_buffer_from_config (buf, &tensor_demux->tensors_config);
 
@@ -479,9 +514,9 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
   GST_DEBUG_OBJECT (tensor_demux, " Number of Tensors: %d", num_tensors);
 
   num_srcs = num_tensors;
-  if (tensor_demux->tensorpick != NULL) {
-    num_srcs = g_list_length (tensor_demux->tensorpick);
-    list = tensor_demux->tensorpick;
+  if (tensorpick != NULL) {
+    num_srcs = g_list_length (tensorpick);
+    list = tensorpick;
   }
 
   for (i = 0; i < num_srcs; i++) {
@@ -491,11 +526,11 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     gboolean created;
     GstClockTime ts;
 
-    srcpad = gst_tensor_demux_get_tensor_pad (tensor_demux, &created, i,
-        num_tensors);
+    srcpad = gst_tensor_demux_get_tensor_pad (tensor_demux, tensorpick,
+        &created, i, num_tensors);
     outbuf = gst_buffer_new ();
 
-    if (tensor_demux->tensorpick != NULL) {
+    if (list != NULL) {
       guint num, j;
       gchar **strv = g_strsplit_set ((gchar *) list->data, ":+", -1);
 
@@ -507,6 +542,7 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
             idx);
         mem = gst_tensor_buffer_get_nth_memory (buf, idx);
         if (!gst_tensor_buffer_append_memory (outbuf, mem, _info)) {
+          g_strfreev (strv);
           gst_buffer_unref (outbuf);
           res = GST_FLOW_ERROR;
           goto error;
@@ -556,6 +592,7 @@ gst_tensor_demux_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
   }
 
 error:
+  g_list_free_full (tensorpick, g_free);
   gst_buffer_unref (buf);
   return res;
 }
@@ -597,6 +634,8 @@ gst_tensor_demux_set_property (GObject * object, guint prop_id,
     const GValue * value, GParamSpec * pspec)
 {
   GstTensorDemux *filter = GST_TENSOR_DEMUX (object);
+
+  GST_OBJECT_LOCK (filter);
   switch (prop_id) {
     case PROP_SILENT:
       filter->silent = g_value_get_boolean (value);
@@ -624,6 +663,7 @@ gst_tensor_demux_set_property (GObject * object, guint prop_id,
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
   }
+  GST_OBJECT_UNLOCK (filter);
 }
 
 /**
@@ -634,6 +674,8 @@ gst_tensor_demux_get_property (GObject * object, guint prop_id,
     GValue * value, GParamSpec * pspec)
 {
   GstTensorDemux *filter = GST_TENSOR_DEMUX (object);
+
+  GST_OBJECT_LOCK (filter);
   switch (prop_id) {
     case PROP_SILENT:
       g_value_set_boolean (value, filter->silent);
@@ -659,4 +701,5 @@ gst_tensor_demux_get_property (GObject * object, guint prop_id,
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
   }
+  GST_OBJECT_UNLOCK (filter);
 }
--- a/gst/nnstreamer/elements/gsttensor_split.c
+++ b/gst/nnstreamer/elements/gsttensor_split.c
@@ -206,7 +206,8 @@ gst_tensor_split_finalize (GObject * object)
   split = GST_TENSOR_SPLIT (object);
   gst_tensor_split_remove_src_pads (split);
   g_list_free (split->tensorpick);
-  g_array_free (split->tensorseg, TRUE);
+  if (split->tensorseg)
+    g_array_unref (split->tensorseg);
   G_OBJECT_CLASS (parent_class)->finalize (object);
 }
 
@@ -251,15 +252,18 @@ gst_tensor_split_event (GstPad * pad, GstObject * parent, GstEvent * event)
 /**
  * @brief Checking if the source pad is created and if not, create TensorPad
  * @param split TensorSplit Object
+ * @param tensorseg private reference of the tensorseg property
+ * @param tensorpick private copy of the tensorpick property
  * @param inbuf inputbuf GstBuffer Object including GstMeta
  * @param[out] created will be updated in this function
  * @param nth source ordering
  * @return TensorPad if pad is already created, then return created pad.
- *         If not return new pad after creation.
+ *         If not return new pad after creation, or NULL if no segment rule
+ *         is left for it.
  */
 static GstTensorPad *
-gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
-    gboolean * created, guint nth)
+gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GArray * tensorseg,
+    GList * tensorpick, GstBuffer * inbuf, gboolean * created, guint nth)
 {
   GstElement *element = GST_ELEMENT_CAST (split);
   g_autofree gchar *element_name = gst_element_get_name (element);
@@ -287,6 +291,13 @@ gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
     walk = g_slist_next (walk);
   }
 
+  if (split->num_srcpads >= tensorseg->len) {
+    GST_ERROR_OBJECT (split,
+        "The tensorseg has %u rules, which leaves none for the %uth source pad.",
+        tensorseg->len, split->num_srcpads);
+    return NULL;
+  }
+
   tensorpad = g_new0 (GstTensorPad, 1);
   g_assert (tensorpad != NULL);
   GST_DEBUG_OBJECT (split, "creating pad: %d(%dth)", split->num_srcpads, nth);
@@ -303,7 +314,7 @@ gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
   tensorpad->last_ts = GST_CLOCK_TIME_NONE;
 
   split->srcpads = g_slist_append (split->srcpads, tensorpad);
-  dim = g_array_index (split->tensorseg, tensor_dim *, split->num_srcpads);
+  dim = g_array_index (tensorseg, tensor_dim *, split->num_srcpads);
 
   split->num_srcpads++;
 
@@ -351,9 +362,9 @@ gst_tensor_split_get_tensor_pad (GstTensorSplit * split, GstBuffer * inbuf,
     *created = TRUE;
   }
 
-  if (split->tensorpick != NULL) {
+  if (tensorpick != NULL) {
     GST_DEBUG_OBJECT (split, "TensorPick is set! : %dth tensor\n", nth);
-    if (g_list_length (split->tensorpick) == split->num_srcpads) {
+    if (g_list_length (tensorpick) == split->num_srcpads) {
       gst_element_no_more_pads (GST_ELEMENT_CAST (split));
     }
   }
@@ -392,13 +403,14 @@ done:
 /**
  * @brief Make splitted tensor
  * @param split TensorSplit object
+ * @param tensorseg private reference of the tensorseg property
  * @param buffer gstbuffer form src
  * @param nth orther of tensor
  * @return return GstMemory for splitted tensor
  */
 static GstMemory *
-gst_tensor_split_get_splitted (GstTensorSplit * split, GstBuffer * buffer,
-    gint nth)
+gst_tensor_split_get_splitted (GstTensorSplit * split, GArray * tensorseg,
+    GstBuffer * buffer, gint nth)
 {
   GstMemory *mem;
   tensor_dim *dim;
@@ -406,18 +418,18 @@ gst_tensor_split_get_splitted (GstTensorSplit * split, GstBuffer * buffer,
   gsize size, offset;
   GstMapInfo src_info, dest_info;
 
-  dim = g_array_index (split->tensorseg, tensor_dim *, nth);
+  dim = g_array_index (tensorseg, tensor_dim *, nth);
   size = gst_tensor_get_element_count (*dim) *
       gst_tensor_get_element_size (split->in_config.info.info[0].type);
 
   mem = gst_allocator_alloc (NULL, size, NULL);
   if (!gst_memory_map (mem, &dest_info, GST_MAP_WRITE)) {
-    ml_logf ("Cannot map memory for destination buffer.\n");
+    GST_ERROR_OBJECT (split, "Cannot map memory for destination buffer.");
     gst_memory_unref (mem);
     return NULL;
   }
   if (!gst_buffer_map (buffer, &src_info, GST_MAP_READ)) {
-    ml_logf ("Cannot map src-memory to gst buffer at tensor-split.\n");
+    GST_ERROR_OBJECT (split, "Cannot map src-memory to gst buffer.");
     gst_memory_unmap (mem, &dest_info);
     gst_memory_unref (mem);
     return NULL;
@@ -425,9 +437,19 @@ gst_tensor_split_get_splitted (GstTensorSplit * split, GstBuffer * buffer,
 
   offset = 0;
   for (i = 0; i < nth; i++) {
-    dim = g_array_index (split->tensorseg, tensor_dim *, i);
+    dim = g_array_index (tensorseg, tensor_dim *, i);
     offset += gst_tensor_get_element_count (*dim) *
         gst_tensor_get_element_size (split->in_config.info.info[0].type);
+  }
+
+  if (offset + size > src_info.size) {
+    GST_ERROR_OBJECT (split, "The tensorseg needs %" G_GSIZE_FORMAT
+        " bytes at offset %" G_GSIZE_FORMAT ", but the incoming buffer is %"
+        G_GSIZE_FORMAT " bytes.", size, offset, src_info.size);
+    gst_buffer_unmap (buffer, &src_info);
+    gst_memory_unmap (mem, &dest_info);
+    gst_memory_unref (mem);
+    return NULL;
   }
 
   nns_memcpy (dest_info.data, src_info.data + offset, size);
@@ -444,17 +466,26 @@ static GstFlowReturn
 gst_tensor_split_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
 {
   GstTensorSplit *split;
+  GArray *tensorseg;
+  GList *tensorpick;
   guint num_tensors, i;
   GstFlowReturn res = GST_FLOW_OK;
   UNUSED (pad);
 
   split = GST_TENSOR_SPLIT (parent);
 
+  GST_OBJECT_LOCK (split);
   num_tensors = split->num_tensors;
+  tensorseg = split->tensorseg ? g_array_ref (split->tensorseg) : NULL;
+  tensorpick = g_list_copy (split->tensorpick);
+  GST_OBJECT_UNLOCK (split);
+
   GST_DEBUG_OBJECT (split, " Number of Tensors: %d", num_tensors);
 
-  if (split->tensorseg == NULL) {
+  if (tensorseg == NULL) {
     GST_ERROR_OBJECT (split, "No rule to split incoming buffers.");
+    g_list_free (tensorpick);
+    gst_buffer_unref (buf);
     return GST_FLOW_ERROR;
   }
 
@@ -465,10 +496,10 @@ gst_tensor_split_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
     gboolean created;
     GstClockTime ts;
 
-    if (split->tensorpick != NULL) {
+    if (tensorpick != NULL) {
       gboolean found = FALSE;
       GList *list;
-      for (list = split->tensorpick; list != NULL; list = list->next) {
+      for (list = tensorpick; list != NULL; list = list->next) {
         if (i == GPOINTER_TO_UINT (list->data)) {
           found = TRUE;
           break;
@@ -478,10 +509,20 @@ gst_tensor_split_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
         continue;
     }
 
-    srcpad = gst_tensor_split_get_tensor_pad (split, buf, &created, i);
+    srcpad = gst_tensor_split_get_tensor_pad (split, tensorseg, tensorpick, buf,
+        &created, i);
+    if (srcpad == NULL) {
+      res = GST_FLOW_ERROR;
+      break;
+    }
 
     outbuf = gst_buffer_new ();
-    mem = gst_tensor_split_get_splitted (split, buf, i);
+    mem = gst_tensor_split_get_splitted (split, tensorseg, buf, i);
+    if (mem == NULL) {
+      gst_buffer_unref (outbuf);
+      res = GST_FLOW_ERROR;
+      break;
+    }
     gst_buffer_append_memory (outbuf, mem);
     ts = GST_BUFFER_TIMESTAMP (buf);
 
@@ -511,6 +552,8 @@ gst_tensor_split_chain (GstPad * pad, GstObject * parent, GstBuffer * buf)
       break;
   }
 
+  g_list_free (tensorpick);
+  g_array_unref (tensorseg);
   gst_buffer_unref (buf);
   return res;
 }
@@ -567,68 +610,73 @@ gst_tensor_split_set_property (GObject * object, guint prop_id,
 
   split = GST_TENSOR_SPLIT (object);
 
+  GST_OBJECT_LOCK (split);
   switch (prop_id) {
     case PROP_SILENT:
       split->silent = g_value_get_boolean (value);
       break;
     case PROP_TENSORPICK:
     {
-      gint i;
+      guint i, num;
       gint64 val;
       const gchar *param = g_value_get_string (value);
       gchar **strv = g_strsplit_set (param, ",.;/", -1);
-      gint num = g_strv_length (strv);
+      GList *tensorpick = NULL;
+
+      num = g_strv_length (strv);
       for (i = 0; i < num; i++) {
         val = g_ascii_strtoll (strv[i], NULL, 10);
-        split->tensorpick =
-            g_list_append (split->tensorpick, GINT_TO_POINTER (val));
+        tensorpick = g_list_append (tensorpick, GINT_TO_POINTER (val));
       }
       g_strfreev (strv);
+
+      g_list_free (split->tensorpick);
+      split->tensorpick = tensorpick;
       break;
     }
     case PROP_TENSORSEG:
     {
-      guint i;
+      guint i, num;
       const gchar *param = g_value_get_string (value);
       gchar **strv = g_strsplit_set (param, ",.;/", -1);
-      GArray *tensorseg = split->tensorseg;
+      GArray *tensorseg;
 
-      split->num_tensors = g_strv_length (strv);
-      if (NULL == tensorseg) {
-        split->tensorseg =
-            g_array_sized_new (FALSE, FALSE, sizeof (tensor_dim *),
-            split->num_tensors);
-        g_array_set_clear_func (split->tensorseg,
-            (GDestroyNotify) _clear_tensorseg);
+      num = g_strv_length (strv);
+      tensorseg = g_array_sized_new (FALSE, FALSE, sizeof (tensor_dim *), num);
+      g_array_set_clear_func (tensorseg, (GDestroyNotify) _clear_tensorseg);
 
-        for (i = 0; i < split->num_tensors; i++) {
-          tensor_dim *d = g_new0 (tensor_dim, 1);
-          g_array_append_val (split->tensorseg, d);
+      for (i = 0; i < num; i++) {
+        tensor_dim *d = g_new0 (tensor_dim, 1);
+        gchar **p = g_strsplit_set (strv[i], ":", -1);
+        guint k, rank = g_strv_length (p);
+
+        if (rank > NNS_TENSOR_RANK_LIMIT) {
+          GST_WARNING_OBJECT (split,
+              "The rank of '%s' exceeds the limit %d, the rest is ignored.",
+              strv[i], NNS_TENSOR_RANK_LIMIT);
+          rank = NNS_TENSOR_RANK_LIMIT;
         }
-        tensorseg = split->tensorseg;
-      }
-      for (i = 0; i < split->num_tensors; i++) {
-        gchar **p;
-        gint num, k;
-        tensor_dim *d;
-        p = g_strsplit_set (strv[i], ":", -1);
-        num = g_strv_length (p);
 
-        d = g_array_index (tensorseg, tensor_dim *, i);
-
-        for (k = 0; k < num; k++) {
+        for (k = 0; k < rank; k++) {
           (*d)[k] = g_ascii_strtod (p[k], NULL);
         }
 
         g_strfreev (p);
+        g_array_append_val (tensorseg, d);
       }
       g_strfreev (strv);
+
+      if (split->tensorseg)
+        g_array_unref (split->tensorseg);
+      split->tensorseg = tensorseg;
+      split->num_tensors = num;
       break;
     }
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
   }
+  GST_OBJECT_UNLOCK (split);
 }
 
 /**
@@ -642,6 +690,7 @@ gst_tensor_split_get_property (GObject * object, guint prop_id,
 
   split = GST_TENSOR_SPLIT (object);
 
+  GST_OBJECT_LOCK (split);
   switch (prop_id) {
     case PROP_SILENT:
       g_value_set_boolean (value, split->silent);
@@ -710,4 +759,5 @@ gst_tensor_split_get_property (GObject * object, guint prop_id,
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
   }
+  GST_OBJECT_UNLOCK (split);
 }
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -9357,6 +9357,210 @@ TEST (testTensorSparse, decInvalidProperty_n)
 }
 
 /**
+ * @brief Rendezvous used to set a property from the test thread exactly while
+ *        the streaming thread is adding a source pad.
+ */
+typedef struct {
+  GMutex lock;
+  GCond cond;
+  gboolean pad_added;
+  gboolean prop_set;
+} padRaceSync;
+
+/**
+ * @brief Handler of "pad-added", which runs on the streaming thread.
+ */
+static void
+_pad_race_pad_added (GstElement *element, GstPad *pad, gpointer user_data)
+{
+  padRaceSync *sync = (padRaceSync *) user_data;
+  gint64 until = g_get_monotonic_time () + TEST_TIMEOUT_LIMIT;
+
+  UNUSED (element);
+  UNUSED (pad);
+
+  g_mutex_lock (&sync->lock);
+  sync->pad_added = TRUE;
+  g_cond_broadcast (&sync->cond);
+  while (!sync->prop_set) {
+    if (!g_cond_wait_until (&sync->cond, &sync->lock, until))
+      break;
+  }
+  g_mutex_unlock (&sync->lock);
+}
+
+/**
+ * @brief Set the given property while the streaming thread is blocked in
+ *        _pad_race_pad_added(), then release the streaming thread.
+ * @return TRUE if the element reported a new pad before the time-out
+ */
+static gboolean
+_pad_race_set_property (padRaceSync *sync, GstElement *element,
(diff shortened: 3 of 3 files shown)


myungjoo:
> **Note:** This review was produced by a separate AI reviewing agent (not the human maintainer) and is being posted here on the maintainer's behalf. Please treat findings accordingly and verify independently before acting.

## Summary

The PR fixes the reported `tensor_demux` property/streaming-thread race (#4769) correctly, and applies an analogous, well-motivated hardening pass to `tensor_split`. I checked out the branch, built it (WSL, meson+ninja, clean build, no warnings), ran `gst-indent` on the two changed `.c` files (no diff — style is clean), and reproduced the claims empirically:

- Built the PR's parent commit (`pr-4925~2`) with only the new test file cherry-picked on top (fix code excluded): 5 of the 6 new tests crash/fail exactly as described (`setTensorpickWhileAddingPad` -> SIGSEGV/core dump, `setTensorsegWhileAddingPad` -> glib abort on a bogus huge allocation, `setTensorsegTwice` -> core dump, `setTensorsegOverRank_n` -> "double free or corruption", `setTensorpickTwice` (split) -> `g_array_free: assertion 'array' failed`); `setTensorpickTwice` (demux) already passed on `main`.
- On the PR branch: full `unittest_plugins` 155/155 pass, `nnstreamer_demux` and `nnstreamer_split` SSAT suites pass.
- Each commit builds and passes its own new tests in isolation (bisectable): commit 1 (`8ff8d135`, demux) compiles standalone and its 2 tests pass without commit 2.
- No header/API/caps changes — confirmed via `git diff --stat` (only the two `.c` files + the test file). No documentation update needed.
- Doxygen comments present on every new function (both production and test code) — CI's doxygen gate should be satisfied.
- Commit messages: correct `[Component] Summary` format, DCO `Signed-off-by` present, no duplication.
- No backdoor / suspicious code observed (locking + list/array lifecycle fixes only).

## Findings

### 1. (High) `tensor_split`'s fix for the reported race class is incomplete — `gst_tensor_split_get_tensor_pad()` still reads `split->tensorpick` unprotected

`gst/nnstreamer/elements/gsttensor_split.c`, function `gst_tensor_split_get_tensor_pad()` (around line 356-359):

```c
if (split->tensorpick != NULL) {
  GST_DEBUG_OBJECT (split, "TensorPick is set! : %dth tensor\n", nth);
  if (g_list_length (split->tensorpick) == split->num_srcpads) {
    gst_element_no_more_pads (GST_ELEMENT_CAST (split));
  }
}
```

This function receives `tensorseg` as a snapshot parameter (correctly threaded through from the `GST_OBJECT_LOCK`-protected copy taken in `gst_tensor_split_chain()`), but it was not given a `tensorpick` parameter — it reaches directly into `split->tensorpick`, which is not protected by `GST_OBJECT_LOCK` here. Meanwhile, `gst_tensor_split_set_property()` does exactly the free-then-replace that caused the original bug:

```c
g_list_free (split->tensorpick);
split->tensorpick = tensorpick;
```

This code path in `gst_tensor_split_get_tensor_pad()` is reached only while a new src pad is being created — i.e. it runs right after `gst_element_add_pad()` / the synchronous `"pad-added"` signal emission, the exact tens-of-milliseconds window this PR's own `setTensorsegWhileAddingPad` / `setTensorpickWhileAddingPad` tests are built to hit for the other fields. If `tensorpick` is set concurrently from the application thread during that window, `g_list_length()` here can walk a list that `g_list_free()` is concurrently deallocating — the same use-after-free/heap-corruption shape described in the PR body for the original bug, just left open in one spot of `tensor_split`.

By contrast, the `tensor_demux` equivalent (`gst_tensor_demux_get_tensor_pad()`) does this correctly — it takes a `tensorpick` snapshot parameter and uses only that, never `tensor_demux->tensorpick` directly. `tensor_split`'s helper should receive the already-copied `tensorpick` local from `gst_tensor_split_chain()` (mirroring how `tensorseg` was threaded through) instead of reading `split->tensorpick` directly.

This gap is not exercised by any of the 6 new tests: `setTensorsegWhileAddingPad` only mutates `tensorseg` during the race window, never `tensorpick`, so this specific line is never hit by a concurrent writer in the test suite.

Why it matters: this is precisely the defect class the PR sets out to eliminate ("property/streaming-thread races"), left unresolved in production code the PR itself touches, in a function the PR itself modified. It does not affect the literal reported issue #4769 (which is in `tensor_demux` and is correctly fixed), but it undercuts the "tensor_split has the same shape of hazard..." fix claimed in the PR body/commit message.

### 2. (Medium) Missing regression test for the tensor_split buffer-leak fix on the "no rule to split" path

The `tensor_split` commit message states: "release the input buffer on the 'no rule to split' path; it was the one exit that did not unref it." This is a real fix (confirmed in the diff — `gst_tensor_split_chain()` now calls `gst_buffer_unref (buf)` before returning `GST_FLOW_ERROR` when `tensorseg == NULL`), but none of the 6 new tests exercises pushing a buffer into `tensor_split` before `tensorseg` is set. Per this repo's test policy (CLAUDE.md: "Bug fix: add a regression test that reproduces the bug"), this fix has no dedicated regression coverage, unlike every other bug enumerated in the same commit message (`setTensorsegTwice`, `setTensorsegOverRank_n`, `setTensorpickTwice` all got tests). A harness-based test (push a buffer with no `tensorseg` set, assert `GST_FLOW_ERROR` and no leaked buffer) would close this gap and guard against a future refactor reintroducing the leak.

### 3. (Low) Silent truncation when a `tensorseg` rule exceeds `NNS_TENSOR_RANK_LIMIT`

`gst_tensor_split_set_property()`, `PROP_TENSORSEG` case:

```c
for (k = 0; k < rank && k < NNS_TENSOR_RANK_LIMIT; k++) {
  (*d)[k] = g_ascii_strtod (p[k], NULL);
}
```

This correctly prevents the previous stack/heap overflow (good, and `setTensorsegOverRank_n` proves it no longer crashes), but the extra dimensions beyond the rank limit are now dropped silently with no `GST_WARNING_OBJECT`. A user who supplies an over-rank rule (e.g. via a typo) gets a silently-truncated, likely-wrong split with no diagnostic. Consider logging a warning when `rank > NNS_TENSOR_RANK_LIMIT` is detected, for field debuggability. Not a correctness or safety issue, just a robustness/UX nit.

## Verification performed

- `meson`/`ninja` build of the PR branch (WSL, `PKG_CONFIG_PATH` set per repo's dev setup): clean, no warnings.
- `gst-indent` on both changed `.c` files: no diff (style compliant).
- Full `unittest_plugins` gtest binary: 155/155 pass on the PR branch.
- Reproduced the described pre-fix failures by building the PR's test file on top of the PR's parent commit (fix excluded): 5/6 new tests crash or fail as claimed, 1/6 already passes on `main`.
- SSAT `nnstreamer_demux` and `nnstreamer_split` suites: pass.
- Verified each of the 2 commits builds and passes its own added tests standalone (bisectable).
- Confirmed no header/API/caps changes and no doc updates required.

## Overall assessment

- Severity of open issues: 1 High (residual race in `tensor_split`, same class as the bug this PR fixes), 1 Medium (missing regression test for a real leak fix), 1 Low (silent truncation, no warning).
- The `tensor_demux` fix (the literal subject of #4769) is correct, minimal, well-tested, and verified to fix the reported crash.
- Recommendation: needs-change — primarily to close the `gst_tensor_split_get_tensor_pad()` unprotected `split->tensorpick` read (Finding 1), which leaves an unresolved instance of the exact hazard class this PR is meant to eliminate. The demux fix alone could be merged as-is if the maintainers prefer to split the PR, but as submitted the `tensor_split` half is not yet complete.

myungjoo:
Thanks — all three findings were valid and are addressed in the amended `[Element/Split]` commit (`815608c`).

**1. High — unguarded `tensorpick` read in `gst_tensor_split_get_tensor_pad()`.** Correct, and it was an inconsistency on my side: I threaded `tensorseg` down as a locked snapshot but left `tensorpick` reading the member. Fixed by passing the pick snapshot down the same way the demux fix does, so the pad creation and the chain loop now agree on one list.

For the record on severity: the residual read was a data race rather than a use-after-free. Both reads are of `split->tensorpick` itself, so the second one picks up the *new* list pointer rather than a freed one — what could go wrong is `g_list_length()` disagreeing with the `num_srcpads` the loop was sized for, and `no_more_pads` firing at the wrong time. Still wrong, still fixed.

**2. Medium — no test for the "no rule to split" exit.** Added `testTensorSplit.pushWithoutTensorseg_n`: a harness with only a sink pad pushes one buffer before `tensorseg` is set and asserts `GST_FLOW_ERROR`. The buffer release itself is not something gtest can assert directly; the valgrind CI job is what would catch it if that `gst_buffer_unref` were dropped again, and this test is what puts a buffer on that path for it to see.

**3. Low — silent truncation.** The clamp now emits `GST_WARNING_OBJECT` naming the offending rule and the limit.

Also added `testTensorSplit.setTensorpickWhileAddingPad`, which drives the call site from finding 1 under the same `pad-added` rendezvous the other race tests use. To be straight about what it is worth: it passes both before and after the fix, because the race it guards has no deterministic observable failure. It locks the plumbing and the pick-replacement contract at that call site; it is not a failing-before/passing-after regression test like the other five.

Local re-verification after the change: `unittest_plugins` 157/157, `gst-indent` and `clang-format` clean, no new compiler warnings.

myungjoo:
> **Note:** This review was produced by a separate AI reviewing agent (not the human maintainer) and is being posted here on the maintainer's behalf. Please treat findings accordingly and verify independently before acting.

## Summary

Re-reviewed HEAD `815608c4` (the amended `[Element/Split]` commit) against the three findings from the previous round, against issue #4769, and against the rest of the two changed files (read in full, not just the diff). Scope of this pass was code-reading only (no build/test execution), cross-checked with the repository's CI workflow definitions.

## Previous findings — verification

1. **High (unguarded `tensorpick` read in `gst_tensor_split_get_tensor_pad()`) — confirmed fixed.** The function now takes `tensorpick` as a parameter (`gst_tensor_split_get_tensor_pad (GstTensorSplit *, GArray *tensorseg, GList *tensorpick, ...)`) and the only call site passes down the lock-acquired snapshot taken in `gst_tensor_split_chain()`. No remaining direct read of `split->tensorpick`/`split->tensorseg` outside a `GST_OBJECT_LOCK` region in either changed file — verified by reading both files in full at HEAD, not just the diff.

   On the severity rebuttal ("a data race rather than a use-after-free"): this doesn't fully hold up for the *pre-amend* code the finding was filed against. `gst_tensor_split_set_property()`'s `PROP_TENSORPICK`/`PROP_TENSORSEG` handlers still do free-then-reassign (`g_list_free (split->tensorpick); split->tensorpick = tensorpick;`, and the `g_array_unref` + reassign for `tensorseg`), now safely serialized against each other by `GST_OBJECT_LOCK`. But a reader that does *not* take that lock (which is exactly what the unpatched `gst_tensor_split_get_tensor_pad()` did) is not synchronized against the writer at all — it can observe `split->tensorpick` at any point in the writer's critical section, including the instant after `g_list_free()` has run but before the pointer is reassigned. In that window the member genuinely holds a dangling pointer, not just a stale-but-valid one, so the pre-amend hazard was UAF-capable, not merely a benign pointer race. This is moot for the merge decision — the fix (both sides taking the same lock) closes the window either way, and the fix is correct — but worth a factual note for the PR history.

2. **Medium (no test for "no rule to split") — confirmed addressed.** `testTensorSplit.pushWithoutTensorseg_n` is present and asserts `GST_FLOW_ERROR` on a harness push before `tensorseg` is set.

   One caveat on the stated backstop: the PR response says "the valgrind CI job is what would catch it if that `gst_buffer_unref` were dropped again." Checking `.github/workflows/ubuntu_clean_meson_build.yml:215`, the valgrind step that runs `run_unittests_binaries.sh --valgrind` is wired as `... || echo "There are Valgrind errors. Please fix it. ... we are not halting the build with Valgrind errors until we get them all."` — i.e. that step cannot fail CI today regardless of what valgrind reports. So a future regression on this exact path (buffer leak on the "no rule to split" exit) would not actually block a merge; it would only be visible to someone who reads the valgrind log by hand. This is a pre-existing, repo-wide CI gap, not something this PR introduced or should be expected to fix in-scope, but it directly bears on review criterion "can CI actually fail and block the offending change" for this specific test, so flagging it for awareness / a possible separate follow-up.

   The other 4 fail-before/pass-after tests (`setTensorpickWhileAddingPad` demux, `setTensorsegWhileAddingPad`/`setTensorsegTwice`/`setTensorsegOverRank_n`/`setTensorpickTwice` split) reproduce outright crashes/aborts (SIGSEGV/SIGABRT/assertion failure), so those are caught by a plain `meson test` failure — no valgrind dependency, solid CI gate.

3. **Low (silent rank truncation) — confirmed fixed.** `GST_WARNING_OBJECT` naming the rule and the limit is now emitted when a `tensorseg` entry's rank exceeds `NNS_TENSOR_RANK_LIMIT`, and `testTensorSplit.setTensorsegOverRank_n` checks the truncation itself.

## Other criteria checked

- **Addresses #4769**: yes — root cause (unsynchronized snapshot of `tensorpick` spanning `gst_tensor_demux_get_tensor_pad()`'s pad-creation window) is fixed at the source; `tensor_split`'s analogous hazard plus its two concurrency-independent defects (stale `GArray` reuse on second `tensorseg` set, unbounded rank write, append-instead-of-replace `tensorpick`) are reasonably in-scope as "same shape of hazard, same element family."
- **Regressions / other modules**: none found. Only `gsttensor_demux.c`, `gsttensor_split.c`, and the shared test file are touched; no header/struct/API changes. The `tensorpick` append→replace change for `tensor_split` is a behavior fix that brings it in line with `tensor_demux`'s pre-existing (and presumably intended) replace semantics, not a regression.
- **Code size**: proportionate — two focused commits, one per element, each pairing production fix with dedicated tests; no drive-by changes elsewhere.
- **Future-regression detection**: the lock+snapshot contract is exercised directly by the round-trip tests (`setTensorpickTwice` for both elements, `setTensorsegTwice`) and by the deterministic pad-added rendezvous tests. The one test explicitly *not* fail-before/pass-after (`testTensorSplit.setTensorpickWhileAddingPad`) is honestly disclosed as such in the PR reply — it locks the call-site contract but wouldn't itself catch every possible future regression at that exact spot.
- **Docs/API**: no header, GObject property, or caps changes — no documentation update required, consistent with the previous round's assessment.

## Verdict

No unresolved code issues. The two notes above (UAF vs. data-race characterization; valgrind CI step not being merge-blocking) are informational only and don't require changes to this PR to merge. **Approve / mergeable** from this review pass.

myungjoo:
Both notes are fair, and the second one was a factual error on my side that is now fixed in the code rather than in prose.

**On the UAF-vs-race correction.** You are right and I was not. I reasoned about the interleaving where the reader lands after the reassignment; the writer's critical section also contains the window between `g_list_free()` and the assignment, and an unsynchronized reader can land there and get a genuinely dangling pointer. My "data race, not UAF" phrasing understated it. Nothing to change in the code — both sides take the lock now — but the record should be right.

**On the valgrind backstop.** Confirmed: `.github/workflows/ubuntu_clean_meson_build.yml:215` ends the valgrind gtest step with `|| echo "There are Valgrind errors. ..."`, so it cannot fail the build. My claim that the job would catch a regressed unref was wrong.

Rather than leave the test resting on a gate that does not gate, `testTensorSplit.pushWithoutTensorseg_n` now checks the buffer's refcount directly:

```c
  buf = gst_harness_create_buffer (h, gst_tensors_info_get_size (&config.info, 0));
  gst_buffer_ref (buf);
  EXPECT_EQ (gst_harness_push (h, buf), GST_FLOW_ERROR);
  EXPECT_EQ (GST_MINI_OBJECT_REFCOUNT_VALUE (buf), 1);
  gst_buffer_unref (buf);
```

Verified by deleting the `gst_buffer_unref` from that exit path and rebuilding: the case reports `Which is: 2` and fails under plain `meson test`, with no valgrind and no sanitizer involved. So all six of the fail-before/pass-after cases now gate on the ordinary CI test run; only `testTensorDemux.setTensorpickTwice` and `testTensorSplit.setTensorpickWhileAddingPad` remain contract locks, as described in the PR body.

The repo-wide valgrind gating gap is real but out of scope here; it is about CI policy for the whole test corpus, not about these two elements.

`unittest_plugins` 157/157 after the change, `clang-format` clean.

myungjoo:
> **Note:** This review was produced by a separate AI reviewing agent (not the human maintainer) and is being posted here on the maintainer's behalf. Please treat findings accordingly and verify independently before acting.

## Summary

Re-reviewed HEAD `4eb14913` (the `pushWithoutTensorseg_n` refcount-check amendment) as a final pass after the two prior review rounds. This pass verified (a) whether the new `GST_MINI_OBJECT_REFCOUNT_VALUE` check is a sound replacement for the valgrind-dependent assertion, (b) whether the PR body still matches the code, and (c) whether any blocking issue remains. Scope was code reading only (diff + full file reads of both changed `.c` files at HEAD, plus the test file); no build/test execution was performed.

## 1. The refcount check is correct

`tests/nnstreamer_plugins/unittest_plugins.cc`, `testTensorSplit.pushWithoutTensorseg_n`:

```c
buf = gst_harness_create_buffer (h, gst_tensors_info_get_size (&config.info, 0));
gst_buffer_ref (buf);
EXPECT_EQ (gst_harness_push (h, buf), GST_FLOW_ERROR);
EXPECT_EQ (GST_MINI_OBJECT_REFCOUNT_VALUE (buf), 1);
gst_buffer_unref (buf);
```

`gst_harness_push()` follows the same transfer-full ownership contract as `gst_pad_push()`: the caller hands over one reference, and the callee (here, `gst_tensor_split_chain()`) is responsible for consuming it — either by forwarding it downstream or by unreffing it on an error exit. The test's `gst_buffer_ref()` before the push is exactly the standard idiom for keeping a buffer inspectable after a call that steals a reference; this codebase already uses the identical pattern for another steals-a-ref API (`tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc:404`, `buf = gst_buffer_ref (buf); gst_app_src_push_buffer (...)`). So the expected count breakdown is:

- After `create_buffer` + `gst_buffer_ref`: refcount 2 (test's own ref + the ref about to be handed to the harness).
- `gst_harness_push` consumes the handed-over ref. On the fixed "no rule to split" path, `gst_tensor_split_chain()` now calls `gst_buffer_unref (buf)` before returning `GST_FLOW_ERROR`, releasing that ref → refcount drops to 1, which is exactly the test's own remaining ref. `EXPECT_EQ(..., 1)` is correct.
- On the pre-fix code (the exit that "did not unref it"), that ref is never released, so the count would stay at 2 — matching both the PR body's table entry ("assertion failure (the buffer keeps a refcount of 2)") and the author's own verification note (deleting the `gst_buffer_unref` reproduces `Which is: 2`).

No incorrect assumption found. This closes the informational gap from round 2 (the valgrind CI step being `|| echo ...`-wrapped and therefore non-blocking) without needing valgrind at all — the six fail-before/pass-after cases (including this one) now all gate on a plain `meson test` run.

## 2. PR body accuracy

Checked the body's "Tests" table and prose against the current diff:

- All 8 listed test names exist in the diff with the described `TEST(testTensorDemux, ...)` / `TEST(testTensorSplit, ...)` names.
- The 6-fail/2-pass split in the table (6 fail-before rows vs. 2 "passes"/contract-lock rows) matches the 8 added cases exactly.
- `pushWithoutTensorseg_n`'s row description ("assertion failure (the buffer keeps a refcount of 2)") matches the current refcount-based implementation, not a stale description of the earlier valgrind-dependent version — the body was correctly updated alongside the code.
- "No header, API or caps change" — confirmed; only `gsttensor_demux.c`, `gsttensor_split.c`, and the test file are touched.
- No exaggeration found in the "What"/"Why" sections relative to the code read in full at HEAD.

## 3. Remaining blocking issues

None found.

- **Regressions**: none. Both `gst_tensor_demux_get_tensor_pad()`/`gst_tensor_split_get_tensor_pad()` and their chain functions consistently thread the locked snapshot (`tensorpick`/`tensorseg`) through instead of touching the live struct members; no direct unguarded member read remains in either file (verified by a full read of both `.c` files, not just the diff hunks).
- **Locking**: `GST_OBJECT_LOCK`/`UNLOCK` pairs in both elements' `set_property`/`get_property`/chain functions are non-nested — no call path re-enters the same object lock (e.g., the lock is always released before `gst_element_add_pad()` / `gst_element_no_more_pads()` run, which is what lets the `*WhileAddingPad` tests' `pad-added` handler safely call `g_object_set()` back in from another thread without deadlocking).
- **Scope**: still proportionate — two focused commits (one per element) plus tests; no drive-by changes.
- **Doxygen/style**: every new function (production and test) carries a `@brief` doxygen comment; K&R/2-space style in the `.c` files and repo `.clang-format` in the test file match surrounding code, consistent with both prior rounds' `gst-indent`/`clang-format`-clean confirmations.
- **Commits**: `[Component] Summary` subjects, DCO `Signed-off-by` present on both, no duplicated sign-off, bodies explain what/why.

## Verdict

No unresolved issues from either review round remain, and the final refcount-based test amendment is technically sound and accurately documented. **Approve / mergeable.**

myungjoo:
Picking up the remaining non-blocking items from the review rounds. All four from the bot review are addressed, plus one defect I had deferred; the branch is rebased onto current `main` (`31259a9e`).

**1. `gst_tensor_demux_dispose()` left `tensorpick` dangling.** Fixed — the member is cleared after the free. `testTensorDemux.disposeTwice` calls `g_object_run_dispose()` twice with the property set; without the fix that aborts on a double free.

**2. `tensorseg` indexed by pad ordinal vs. tensor index.** Not folded in — filed as #4927 with a reproduction showing caps advertising `1:4:4` while a 32-byte buffer is delivered. It needs a decision on whether `tensorpick` renumbers pads, and either answer changes the caps existing pipelines see.

**3. `setTensorsegOverRank_n`'s length assertion was vacuous.** Correct — the getter always emits `NNS_TENSOR_RANK_LIMIT` dimensions, so the count held either way. The rule is now `1:2:3:...:20` and the test compares the whole string against `1:2:...:16`, which pins *which* dimensions survive, not just how many.

**4. `goto error` leaking on the chain function's error exit.** Half right, and thanks — the `strv` leak is real and fixed. The `GstMemory` is not leaked: `gst_tensor_buffer_append_memory()` unrefs the incoming memory on every failure path (`/* Release incoming memory even if failed to append it into buffer. */`), so the caller must not. `testTensorDemux.pushOutOfRangeTensorpick_n` now puts a buffer on that exit — it passes either way, but nothing reached that code before.

Also addressed, from the earlier round: `testTensorSplit.setTensorpickWhileAddingPad` was a contract lock with no failing-before property. It now asserts *when* `no-more-pads` is reported. Reading the member instead of the snapshot ends the report after `src_0` — and `src_1` is then added behind it, which is a GStreamer contract violation — so the case fails with the old code and passes with the new one.

**One more I had deferred and am including now:** `gst_tensor_split_get_splitted()` never checked the segment rule against the buffer it was applied to.

```bash
gst-launch-1.0 -e videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=4,height=4,framerate=30/1 ! tensor_converter ! tensor_split name=s tensorseg=3:4:4,3:400:400 ! fakesink
```

reads 1.44 MB out of a 48-byte tensor and reports success. It is not visible to a sanitizer or to valgrind either, because with `HAVE_ORC` the copy goes through `orc_memcpy()` for anything over 100 bytes — so the element has to refuse it itself. `testTensorSplit.pushOversizedTensorseg_n` covers it.

Adding that check exposed something adjacent: the function reported its failures with `ml_logf()`, which is `g_error()` and aborts, so all three of its `return NULL` paths were unreachable and the chain function had never needed to handle NULL. Those are `GST_ERROR_OBJECT()` now, and the chain function ends the buffer rather than appending nothing.

11 cases total, 9 of which fail on `main` under plain `meson test`. Locally: `unittest_plugins` 160/160; `meson test` 21/22 (`unittest_filter_python3`, a NumPy ABI mismatch in this environment, fails identically on unmodified `main`); SSAT `nnstreamer_demux`/`split`/`mux`/`merge` pass; `gst-indent` and `clang-format` clean; each commit still builds and passes its own tests standalone.

myungjoo:
(Relayed by an AI code-review agent; the analysis below was produced by that agent and is posted here on its behalf, not by a human reviewer.)

**Scope reviewed**: the two touched files (`gsttensor_demux.c`, `gsttensor_split.c`) plus the test additions in `unittest_plugins.cc`, verified against the current head (`f28ed6d5`) via `gh pr diff`/`gh api contents`, cross-checked against `main` and the callers of `gst_tensor_buffer_append_memory()`.

**Verdict: LGTM, no blocking issues found. Mergeable once the still-pending CI jobs finish green.**

Verification of the five changes claimed for this round (the sixth item — filing #4927 and excluding the caps/data segment-index mismatch — is a scoping decision, not a code change, and is appropriate: it's a pre-existing defect this PR does not worsen, it needs a caps-semantics decision, and it changes what existing `tensorpick` pipelines see, so keeping it out of a concurrency/memory-safety PR is correct):

1. **`gst_tensor_demux_dispose()` NULL-clear** — confirmed present: `g_list_free_full (tensor_demux->tensorpick, g_free); tensor_demux->tensorpick = NULL;`. `testTensorDemux.disposeTwice` calls `g_object_run_dispose()` twice and would double-free without the added line.
2. **`goto error` leak fix + the GstMemory counter-claim** — confirmed correct on both counts:
   - The `strv` leak is fixed: `g_strfreev (strv)` was added right before the `goto error` inside the pick loop, and `pushOutOfRangeTensorpick_n` (tensorpick=5 against a 1-tensor buffer) drives exactly that path.
   - The author's rebuttal on `GstMemory` is right, and the bot's follow-up was a false positive: `gst_tensor_buffer_append_memory()`'s doc comment states "This function takes ownership of this, even if it returns failure," and the implementation (`nnstreamer_plugin_api_impl.c:1902-1904`) unconditionally does `if (memory) gst_memory_unref (memory);` on every exit path, success or failure. Adding a caller-side unref on failure — which is what the bot's follow-up would have produced — would have been a double-free/use-after-free regression. Good that it was not "fixed."
3. **`setTensorsegOverRank_n` full-string comparison** — confirmed: the test now builds `expected` from the first `NNS_TENSOR_RANK_LIMIT` values and does `EXPECT_STREQ`, instead of only checking `g_strv_length() == NNS_TENSOR_RANK_LIMIT` (which the getter always returns regardless of the clamp, per `gst_tensor_split_get_property()`'s `PROP_TENSORSEG` case, and so was vacuous). This now pins the truncation semantics as intended.
4. **`setTensorpickWhileAddingPad` no-more-pads assertion** — confirmed: the new `padRaceSync`/`_pad_race_no_more_pads` rendezvous records `pads_added` at the moment `no-more-pads` fires and the test asserts it equals `2U`. This is a real fail-before/pass-after regression check: reading the live property instead of the per-buffer snapshot would report "no more pads" after only `src_0`, with `src_1` added behind it, violating the sometimes-pad contract. The synchronization (mutex/cond rendezvous blocking the streaming thread inside `pad-added` until the test thread finishes `g_object_set`) is deterministic, not timing-dependent.
5. **`gst_tensor_split_get_splitted()` bounds check + `ml_logf()` → `GST_ERROR_OBJECT()`** — confirmed correct and safe:
   - The new check `if (offset + size > src_info.size)` runs after `offset` is computed from segments `0..nth-1` and before the `nns_memcpy`, using strict `>`, so it rejects only genuine over-reads and never a segment set that exactly fills the buffer — no false rejection of legitimate `tensorseg` configurations.
   - `ml_logf()` is `g_error()` (aborts); switching to `GST_ERROR_OBJECT()` makes the three existing `return NULL` paths in this function actually reachable for the first time. The chain function was updated in lock-step (`if (mem == NULL) { gst_buffer_unref (outbuf); res = GST_FLOW_ERROR; break; }`), and the post-loop cleanup (`g_list_free (tensorpick); g_array_unref (tensorseg); gst_buffer_unref (buf);`) still runs after a `break` from that new path, so there is no leak on the newly-reachable error exits either.
   - `pushOversizedTensorseg_n` and `pushWithoutTensorseg_n` both assert `GST_MINI_OBJECT_REFCOUNT_VALUE (buf) == 1` after the push, confirming the input buffer is unreffed exactly once on these paths.

Other checks:
- **Scope/size**: only `gsttensor_demux.c`, `gsttensor_split.c`, and the shared `unittest_plugins.cc` are touched (628/-61 across 3 files); no header, no public API/caps change, no other module pulled in. Size is proportionate — most of the diff is the required regression tests, one commit per element as the PR describes.
- **Regression risk**: all `GST_OBJECT_LOCK`/`UNLOCK` pairs in both elements' `set_property`/`get_property` are balanced (no `return` inside the locked `switch`, only `break`), so there is no lock-leak. Both streaming-thread paths (`gst_tensor_demux_chain`, `gst_tensor_split_chain`) consistently use the local snapshot (`tensorpick`/`tensorseg`) rather than the live member from `get_tensor_pad()`/`get_tensor_config()`/`get_splitted()` onward — no leftover direct member access that would reintroduce the race.
- **Test wiring / CI gating**: all new cases land in `tests/nnstreamer_plugins/unittest_plugins.cc`, part of the existing `meson test` suite already gated by CI (not a valgrind-only or manual-run path), so a future regression in either element's locking or bounds logic fails a required check.
- **Docs**: no architecture or API change, so no documentation update is needed, and none was made — consistent.

**CI status at review time**: no failing checks. `Analyze (c-cpp/python)`, `CPP-Linter`, `Static checks`, `DCO`, `CodeQL`, the ARM Valgrind job, and several platform builds are green. `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` (x86_64) and the Tizen GBS `unit_test 1` (x86_64) job — the two most relevant to these GTest additions — were still `pending` as of this review; worth confirming they finish green before merge, but nothing found in the code suggests they won't.

myungjoo:
Finding (b) is correct and it is my regression — thanks for catching it, and for the addendum separating it from #4927.

The old setter allocated the segment array once (`if (NULL == tensorseg)`) and never replaced it, so its length could only ever be the first set's, and `num_srcpads` stayed inside it. Replacing it on every set is what lets it shrink under pads that already exist. That is a consequence of this PR, not a pre-existing defect, so it belongs here.

Fixed with the bounds check rather than the `nth` change, so #4927 keeps the semantics decision:

```c
  if (split->num_srcpads >= tensorseg->len) {
    GST_ERROR_OBJECT (split,
        "The tensorseg has %u rules, which leaves none for the %uth source pad.",
        tensorseg->len, split->num_srcpads);
    return NULL;
  }
```

`gst_tensor_split_chain()` ends the buffer on a NULL pad, the same way it already does for a NULL memory. The guard cannot fire on a consistent snapshot — `num_srcpads` only ever reaches the number of picked tensors, which is at most `tensorseg->len` — so it costs nothing in normal use.

`testTensorSplit.addPadBeyondTensorseg_n` covers it, and needs no threading: two rules with `tensorpick=1` put `src_0` at ordinal 0 of two, then one rule with `tensorpick=0` asks for a pad at ordinal 1 of one. Without the guard that test **core dumps in a plain build**, and under ASan it is a SEGV in `gst_tensor_split_get_tensor_pad` at the `(*dim)[i]` dereference of the pointer read past the array — so it gates on the ordinary `meson test` run, no sanitizer required.

Part (a) left to #4927 as you suggest; I have added your `tensorseg=1:8:8,2:8:8 tensorpick=1` framing as a second reproduction there. Nothing else in the review needed a change.

12 cases now, 10 of them failing without the change they cover. Locally: `unittest_plugins` 161/161; `meson test` 21/22 (`unittest_filter_python3`, a NumPy ABI mismatch in this environment, fails identically on unmodified `main`); SSAT `nnstreamer_demux`/`split`/`mux`/`merge` pass; `gst-indent` and `clang-format` clean.

myungjoo:
(Relayed by a review bot: the analysis below was produced by an AI review agent and is posted on its behalf.)

**Re-review at `2125df54`**, scope: whether the blocking finding from the previous round (`gsttensor_split.c:309` OOB) is actually fixed, whether deferring part (a) to #4927 was sound, and a general pass over the PR against issue #4769.

### Blocking finding (b) — verified fixed

`gst_tensor_split_get_tensor_pad()` now has, at `gst/nnstreamer/elements/gsttensor_split.c:294-299`:

```c
if (split->num_srcpads >= tensorseg->len) {
  GST_ERROR_OBJECT (split, "The tensorseg has %u rules, ...");
  return NULL;
}
```

Checked the three things this needed to get right:

- **Position.** It sits after the `walk`/`g_slist_next` loop that early-returns an already-existing pad (lines 282-291) and before both `g_new0 (GstTensorPad, 1)` and the `g_array_index (tensorseg, tensor_dim *, split->num_srcpads)` read (line 317). So an existing pad is still returned via the early-return path without ever reaching the guard, and a genuinely new pad cannot be created (and cannot index the array) once `num_srcpads` has caught up with `tensorseg->len`.
- **No leak on the NULL path.** `gst_tensor_split_chain()` (same file, ~line 511) checks `srcpad == NULL` before an `outbuf` is even allocated, sets `GST_FLOW_ERROR`, and `break`s into the function's single exit, which unconditionally does `g_list_free (tensorpick); g_array_unref (tensorseg); gst_buffer_unref (buf);`. Nothing is allocated between the guard and the check, so there is nothing to leak.
- **False-positive risk.** Traced both the no-`tensorpick` case (pads created in strictly increasing `nth` order, so `num_srcpads` tracks `nth` 1:1 and stays `< tensorseg->len` as long as `tensorseg->len == num_tensors`, which the property setter always keeps in sync) and a sparse `tensorpick` case (e.g. `0,2` over a 3-entry `tensorseg`) — `num_srcpads` only ever counts *distinct new pads actually created*, so it stays behind `tensorseg->len` for any config where the pick list isn't longer than the segment array. The guard only fires when a genuinely new pad is requested after the array has shrunk to (or below) the existing pad count, i.e. exactly the case with no rule left for it. No legitimate combination gets rejected.

The new `testTensorSplit.addPadBeyondTensorseg_n` (`tests/nnstreamer_plugins/unittest_plugins.cc:9851`) reproduces the addendum's exact repro (shrink `tensorseg` from 2 rules to 1 while a pad already sits at ordinal 0, then ask for a new one at ordinal 1) and is a deterministic `GstHarness` test — it doesn't depend on a thread race or a sanitizer to catch a regression, so it will fail loudly and reliably in plain CI if this guard is ever removed or the ordinal math regresses again.

### Part (a) deferral to #4927 — reasonable

Confirmed the code still indexes the segment array by `split->num_srcpads` (pad-creation order) rather than `nth` (the value in `tensorpick`) — e.g. `tensorseg=1:8:8,2:8:8 tensorpick=1` still produces caps for segment 0 on the pad carrying segment 1's bytes. That's a semantics question (index by `nth` vs. renumber the pad), not a memory-safety bug, and correctly does not need this PR's array-lifetime rework to answer it. Deferring it to #4927 is the right call and keeps this PR's diff focused.

### Fit with issue #4769 / general quality

- The tensor_demux half directly closes the reported race: `tensorpick` is copied once under `GST_OBJECT_LOCK` at the top of `gst_tensor_demux_chain()` (`gst_tensor_demux_get_tensorpick()`) and that single snapshot is threaded through the rest of the function, so a concurrent `g_object_set` can no longer produce the split-read (`num_srcs` from one read, `list->data` from a second, later read) that caused the original NULL-deref/UAF. Property get/set are now also lock-guarded.
- Diff stays scoped to the two elements plus their tests; no header/API changes, no other module touched — matches the size of the problem.
- Memory handling in the touched paths checked out: `GArray`/`GList` ownership is consistent (`g_array_ref`/`g_array_unref` pairs around the chain function's snapshot, `g_list_copy`/`g_list_free` likewise, `g_strfreev` added on the previously-leaking `goto error` path in demux).
- No architecture/API/header change, so no documentation update is needed or missing.
- Test coverage is thorough for this class of bug: real `GCond`/`GMutex` rendezvous tests pin the pad-creation-vs-property-set race for both elements, plus deterministic tests for double-set replace semantics, rank clamping, oversized `tensorseg`, out-of-range `tensorpick`, and the OOB-shrink case above.

### New issue found — merge-blocking, mechanical

CI is currently **red** on this exact head (`2125df54`, "Static checks" → Doxygen tag check, run 33934302654, job 101219107870), and it is caused by this PR's own new test code, not anything pre-existing:

```
tests/nnstreamer_plugins/unittest_plugins.cc:9872-9873
  /* one rule now, but src_0 already took ordinal 0, so the pad for the first
     tensor would ask for ordinal 1 */
```

This is a plain `/* ... */` block spanning two lines; the repo's doxygen-tag checker (`.github/workflows/static.check.scripts/doxygen-tag.sh`) rejects any multi-line comment that doesn't open with `/**`, and flags exactly this one. It's the comment directly inside `testTensorSplit.addPadBeyondTensorseg_n` — the very test that pins the fix for blocking finding (b). Every other new multi-line-looking comment in this PR closes on the same line (e.g. `/* picking the second of two rules puts src_0 at ordinal 0 of two */`) and is exempt; this is the one spot that wraps to a second line. `gh api .../check-runs` confirms the failure is recorded against `head_sha: 2125df54`, and the PR's `mergeable_state` is `blocked`.

Severity: low (pure comment-style, no code-behavior change needed — split it into a single-line comment or open it with `/**`), but it is a hard CI gate per this repo's static-check workflow, so **the PR cannot merge as-is** until that one comment is reformatted.

### Verdict

Substance: blocking finding (b) is correctly and completely resolved, part (a)'s deferral to #4927 is sound, the PR addresses issue #4769's root cause, and no regression to other modules or missing documentation was found.

Process: **needs-change**, but only for the mechanical Doxygen-style fix above (`tests/nnstreamer_plugins/unittest_plugins.cc:9872-9873`). Once that comment is reformatted and CI goes green, this is mergeable — no further substantive concerns.

myungjoo:
Right, and thanks for reading the CI rather than trusting my "green" claim — that failure landed after I last looked. `doxygen-tag.sh` rejects any multi-line `/* */` that does not open with `/**`, and the offending comment was in the very test that covers the blocking finding.

The project style also says to avoid multi-line comments, so it is one line now rather than a `/**` block:

```c
  /* one rule now, but src_0 already took ordinal 0 */
```

I also ran the repo's own checkers locally this time instead of only `gst-indent`/`clang-format`: `doxygen-tag.sh`, `indent.sh`, `newline.sh`, `flawfinder.sh`, `hardcoded-path.sh`, `misspelling.sh`, `prohibited-words.sh` and `executable.sh` all pass on the three changed files (`cppcheck` is not installed here; `nobody.sh` takes a different argument shape).

`unittest_plugins` 161/161 after the change. New head is `de3b13b6`.

myungjoo:
I audited the test coverage of this PR by mutation: revert each behavioural change on its own, rebuild, and run the added cases to see which one actually catches it. Fifteen changes, eleven caught. One of the four gaps was closeable, so it is closed now.

| reverted change | caught by |
| --- | --- |
| demux chain re-reads the member inside the loop (the #4769 defect) | `setTensorpickWhileAddingPad` — SIGSEGV |
| demux `dispose()` leaves `tensorpick` dangling | `disposeTwice` — abort |
| split rebuilds the segment array | `setTensorsegTwice`, `addPadBeyondTensorseg_n` |
| split `tensorpick` replaces rather than appends | `setTensorpickTwice`, `setTensorpickWhileAddingPad` |
| split per-rule rank clamp | `setTensorsegOverRank_n` — abort |
| split copy bounded against the input buffer | `pushOversizedTensorseg_n` |
| split pad ordinal bounded against the rule array | `addPadBeyondTensorseg_n` — SIGSEGV |
| split chain handles a NULL memory | `pushOversizedTensorseg_n` |
| split releases the buffer on the "no rule" exit | `pushWithoutTensorseg_n` |
| split chain copies the pick list | `setTensorpickWhileAddingPad` |
| split pad creation uses the snapshot | `setTensorpickWhileAddingPad` |

**Newly closed.** `finalize()` releasing the segment array with `g_array_unref()` instead of `g_array_free()` was uncovered: an unset array made the old call log `g_array_free: assertion 'array' failed`, and a GLib critical does not fail a gtest. `testTensorSplit.finalizeWithoutTensorseg` installs a log handler for the duration of the release and asserts no critical was issued; reverting the change makes it fail.

**Still uncovered, and why.**

- The demux `selected_tensor == NULL` guard. Reverting it to the old `g_assert` leaves all cases passing, because the guard is unreachable by construction: `gst_tensor_demux_get_tensor_config()` has one caller, and the chain function only ever calls it with `nth < g_list_length(snapshot)`. It replaced an assert that could abort the process; there is no way to reach either.
- The `g_strfreev()` on the demux chain function's error exit. A leak has nothing a gtest can observe. `pushOutOfRangeTensorpick_n` at least puts a buffer on that exit for a leak checker to see.
- The `GST_OBJECT_LOCK` in the property handlers. Its absence is a data race with no deterministic failure; the two `*WhileAddingPad` cases pin the ordering the lock produces, not the lock.

**Existing behaviour.** The thirteen cases are property and error paths, so the happy path rests on what was already there, and it still passes: `unittest_sink` 21/21 over the demux/split pipelines (`tensorpick=0,2`, `tensorpick=1,2`, flexible demux, `tensorseg` ×3 with `tensorpick=0,1,2`), and the SSAT `nnstreamer_split` and `nnstreamer_demux` groups, which compare output files byte for byte. Those also show the two new bounds checks do not reject a legitimate rule set — SSAT `split02` runs three segments across three pads through both of them.

`unittest_plugins` 162/162; `meson test` 21/22 (`unittest_filter_python3`, a NumPy ABI mismatch here, fails the same on unmodified `main`); `doxygen-tag.sh`, `indent.sh`, `newline.sh`, `flawfinder.sh`, `hardcoded-path.sh`, `misspelling.sh` and `prohibited-words.sh` clean. New head is `16f59ef1`.

myungjoo:
Good catch, and it is a straight mistake on my side — not a judgement call. When I rebased onto the new `main` I restored both element files with `git checkout <commit> -- <paths>`, which stages as well as writes, so `gsttensor_split.c` was already in the index and the subsequent path-selective `git add` for the demux commit did not take it back out. The messages were right; the contents were not.

Re-split at `392af063`:

```
04a0131b [Element/Demux] Snapshot tensorpick before splitting a buffer
    gsttensor_demux.c    +58 -15
    unittest_plugins.cc  +204 -0     (4 testTensorDemux cases + the pad-race rendezvous)

392af063 [Element/Split] Serve each buffer from one tensorseg snapshot
    gsttensor_split.c    +97 -47
    unittest_plugins.cc  +356 -0     (9 testTensorSplit cases)
```

Each commit now touches one element and carries its own tests, and `Fixes: #4769` stays on the demux one. The resulting tree is byte-identical to `16f59ef1` — `git diff 16f59ef1 392af063` is empty — so nothing you reviewed in the code changed.

Bisectability checked rather than assumed: `04a0131b` alone builds with no warnings and its four demux cases pass with `gsttensor_split.c` still at `main`; the tip builds clean and runs 162/162.

On the delta you did review: agreed on all three points, including that `_n` would be wrong for `finalizeWithoutTensorseg` since it asserts the normal outcome. On the restored log handler's `user_data` — `g_log_set_default_handler()` cannot return the old pointer, but the default handler ignores it, so restoring with NULL puts things back as they were.

PR #4926, opened 2026-09-04
[Common] Do not trust the extra-tensor header of a gst-buffer

myungjoo:
Fixes item **A1** of #4920.

## Problem

A tensor `GstBuffer` holds at most `NNS_TENSOR_MEMORY_MAX` (16) `GstMemory` blocks. To carry more than 16 tensors, the 16th block is replaced by one block that begins with a private `GstTensorExtraInfo` header:

```c
typedef struct {
  uint32_t magic;                                     /* 0xf00dc0de */
  uint32_t version;
  uint32_t num_extra_tensors;
  uint64_t reserved;                                  /* size of tensor 15's data */
  GstTensorInfo infos[NNS_TENSOR_SIZE_EXTRA_LIMIT];   /* 240 entries */
} GstTensorExtraInfo;
```

Every field of that header is plain buffer payload. It is produced by nnstreamer itself in the ordinary case, but it also arrives from outside the process: `tensor_query` and `edge` read it off the network, `appsrc` and `datareposrc` take it from the application. None of it was range-checked.

**`gst_tensor_buffer_get_count()`** returned `num_extra_tensors + 16` exactly as it was found in the buffer. Its callers size their arrays by `NNS_TENSOR_SIZE_LIMIT` (256) and then index them with that count:

| caller | array |
|---|---|
| `tensor_query_server.c:190` | `GstMemory *mem[NNS_TENSOR_SIZE_LIMIT]`, `GstMapInfo map[...]` — **on the stack**, count comes from the network peer |
| `tensor_query_client.c:679` | same, same source |
| `tensor_filter.c:150` | `FilterTransformData.mem/info/meta/tensors[NNS_TENSOR_SIZE_LIMIT]` |
| `gsttensor_decoder.c`, `gsttensor_transform.c`, `tensor_converter_python3.cc`, and 15 more | ditto |

A 16-memory buffer whose last block starts with the magic and a large count therefore overflows those arrays. The two `tensor_query` ones are stack overflows driven straight by a remote peer.

**`gst_tensor_buffer_get_nth_memory()`** indexed `extra_info->infos[index - 16]` past the 240 entries of the array, and handed `gst_memory_share()` an offset and a size accumulated from `reserved` and `infos[].size` without ever comparing them to the size of the block being shared — so even a header with a plausible count could produce a `GstMemory` that reads outside the allocation.

**`gst_tensor_buffer_append_memory()`** wrote `infos[num_extra_tensors]` and incremented the count with no limit, so the 257th tensor of a buffer overflowed the array with no forged header involved at all.

## Fix

Three range checks, all in `gst/nnstreamer/nnstreamer_plugin_api_impl.c`:

1. `gst_memory_map_is_extra_tensor()` — the single place that decides whether a block carries the header — now also requires `num_extra_tensors <= NNS_TENSOR_SIZE_EXTRA_LIMIT`. A forged header is simply not recognised, so `gst_tensor_buffer_get_count()` degrades to the 16 blocks the buffer really has and every caller stays inside its array. This bounds the count at `NNS_TENSOR_SIZE_LIMIT`, which is exactly the size the callers allocate.
2. `gst_tensor_buffer_append_memory()` refuses to append once the buffer holds `NNS_TENSOR_SIZE_LIMIT` tensors.
3. `gst_tensor_buffer_get_nth_memory()` checks the accumulated offset and each parsed size against the mapped size of the block before sharing it, and returns NULL instead of an out-of-range `GstMemory`.

The wire format is unchanged; only values that were already invalid are rejected.

## Tests

New cases in `tests/common/unittest_common.cc`. The test file mirrors the private header struct so a forged one can be built, and the first test self-validates that mirror against a header produced by the real code, so a future layout change fails loudly instead of silently testing nothing.

| test | what it pins |
|---|---|
| `tensorBufferExtraTensors` | a buffer of exactly 256 tensors builds, reports 256, and every tensor reads back with the right size and content |
| `tensorBufferAppendOverLimit_n` | the 257th append is refused, the count stays 256 and the buffer is still readable |
| `tensorBufferForgedExtraCount_n` | a forged count (100000, and the exact boundary 241) degrades the count to 16 and rejects the out-of-range indices |
| `tensorBufferForgedExtraMagic_n` | a corrupted magic degrades the count to 16 |
| `tensorBufferForgedExtraReserved_n` | a forged `reserved` (`G_MAXUINT64`, and one byte past the block) yields NULL, not an out-of-range share |
| `tensorBufferForgedExtraSize_n` | a forged tensor size yields NULL through both the final and the in-loop bound check |
| `tensorBufferForgedExtraCountPipeline_n` | the same forged buffer pushed through a real `appsrc ! tensor_sink` pipeline reaches the sink as a 16-tensor buffer |

Every negative case was confirmed to fail (or crash the test binary) against the pre-fix source, and to pass with the fix.

The `> 16 tensors` feature keeps its existing end-to-end coverage in `tests/nnstreamer_sink/unittest_sink.cc` (`extraTensors.*`, which builds 20-tensor buffers through `tensor_mux` and `appsrc`); that suite passes unchanged.

## Local verification

`meson test -C build` on Ubuntu: 21/22 suites OK. The one failure, `unittest_filter_python3`, is a pre-existing numpy 2.x breakage of this machine's python environment (`_ARRAY_API not found`) and fails identically without this change.

`clang-format` and `.github/workflows/static.check.scripts/doxygen-tag.sh` (advanced mode) are clean on both changed files, and GNU `indent` with the CI parameters produces no diff in the changed regions.

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -54,7 +54,6 @@ static gboolean
 gst_memory_map_is_extra_tensor (GstMapInfo * map)
 {
   GstTensorExtraInfo *extra_info;
-  gboolean is_extra;
 
   g_return_val_if_fail (map != NULL, FALSE);
 
@@ -64,9 +63,18 @@ gst_memory_map_is_extra_tensor (GstMapInfo * map)
   extra_info = (GstTensorExtraInfo *) map->data;
 
   /* check magic in header (extra info) of the memory */
-  is_extra = (extra_info && extra_info->magic == NNS_TENSOR_EXTRA_MAGIC);
+  if (extra_info->magic != NNS_TENSOR_EXTRA_MAGIC)
+    return FALSE;
 
-  return is_extra;
+  /* the header comes from the buffer data, do not trust its tensor count */
+  if (extra_info->num_extra_tensors > NNS_TENSOR_SIZE_EXTRA_LIMIT) {
+    nns_loge ("Invalid extra header, the number of extra tensors (%u) "
+        "exceeds the limit (%d).", extra_info->num_extra_tensors,
+        NNS_TENSOR_SIZE_EXTRA_LIMIT);
+    return FALSE;
+  }
+
+  return TRUE;
 }
 
 /**
@@ -1687,7 +1695,7 @@ GstMemory *
 gst_tensor_buffer_get_nth_memory (GstBuffer * buffer, const guint index)
 {
   guint i, num_tensors;
-  gsize offset;
+  gsize offset, size;
   GstMemory *extra_tensors_memory, *res_mem = NULL;
   GstMapInfo extra_tensors_map;
   GstTensorExtraInfo *extra_info;
@@ -1728,28 +1736,31 @@ gst_tensor_buffer_get_nth_memory (GstBuffer * buffer, const guint index)
     goto done;
   }
 
-  /* parse the memory */
+  /* parse the memory, the reserved data is followed by the extra tensors */
   extra_info = (GstTensorExtraInfo *) extra_tensors_map.data;
   offset = sizeof (GstTensorExtraInfo);
+  size = extra_info->reserved;
 
-  /* If index is NNS_TENSOR_MEMORY_MAX - 1 */
-  if (index == NNS_TENSOR_MEMORY_MAX - 1) {
-    res_mem =
-        gst_memory_share (extra_tensors_memory, offset, extra_info->reserved);
-    goto done;
+  /* every offset and size below is parsed from the buffer data, bound them */
+  for (i = NNS_TENSOR_MEMORY_MAX - 1; i < index; ++i) {
+    if (size > extra_tensors_map.size - offset)
+      goto invalid;
+
+    offset += size;
+    size = gst_tensor_info_get_size (&extra_info->infos[i -
+            (NNS_TENSOR_MEMORY_MAX - 1)]);
   }
 
-  offset += extra_info->reserved;
-
-  for (i = 1; i <= index - NNS_TENSOR_MEMORY_MAX; ++i) {
-    offset += gst_tensor_info_get_size (&extra_info->infos[i - 1]);
-  }
+  if (size > extra_tensors_map.size - offset)
+    goto invalid;
 
   /* wrap it as GstMemory */
-  res_mem =
-      gst_memory_share (extra_tensors_memory, offset,
-      gst_tensor_info_get_size (&extra_info->infos[index -
-              NNS_TENSOR_MEMORY_MAX]));
+  res_mem = gst_memory_share (extra_tensors_memory, offset, size);
+  goto done;
+
+invalid:
+  nns_loge ("Invalid extra header, the %u-th tensor is out of the memory.",
+      index);
 
 done:
   gst_memory_unmap (extra_tensors_memory, &extra_tensors_map);
@@ -1824,7 +1835,15 @@ gst_tensor_buffer_append_memory (GstBuffer * buffer, GstMemory * memory,
 
   /* if the memory does not have proper header, append it */
   is_extra = gst_memory_map_is_extra_tensor (&last_memory_map);
-  if (!is_extra) {
+  if (is_extra) {
+    extra_info = (GstTensorExtraInfo *) last_memory_map.data;
+
+    if (extra_info->num_extra_tensors >= NNS_TENSOR_SIZE_EXTRA_LIMIT) {
+      nns_loge ("Failed to append memory, the buffer already has the maximum "
+          "number of tensors (%d).", NNS_TENSOR_SIZE_LIMIT);
+      goto failed;
+    }
+  } else {
     new_mem_size += sizeof (GstTensorExtraInfo);
   }
 
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -12,6 +12,7 @@
 #include <gtest/gtest.h>
 #include <glib.h>
 #include <glib/gstdio.h>
+#include <gst/app/gstappsrc.h>
 #include <nnstreamer_conf.h>
 #include <nnstreamer_plugin_api.h>
 #include <tensor_common.h>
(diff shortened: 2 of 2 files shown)


myungjoo:
*This review was performed and written by a separate Claude Code review agent (not a human), instructed to audit this PR against issue #4920 item A1. Posting it here as a relayed review.*

## Summary verdict: Approve, no blocking issues found

I checked this PR against: (1) whether it actually closes item A1 of #4920, (2) regression risk to other modules, (3) diff size/footprint vs. the problem, (4) test completeness (including drift-detection for future changes), (5) whether the new tests are wired into a CI gate that would actually block a bad merge, and (6) whether any architecture/API doc update is owed.

### 1. Does it solve A1?
Yes, all three sub-defects listed in #4920/A1 are addressed at their described locations in `gst/nnstreamer/nnstreamer_plugin_api_impl.c`:
- `gst_memory_map_is_extra_tensor()` now rejects a header whose `num_extra_tensors > NNS_TENSOR_SIZE_EXTRA_LIMIT` (240), so `gst_tensor_buffer_get_count()` degrades to 16 for a forged/corrupt header instead of returning an attacker-controlled count. I traced this through: `NNS_TENSOR_SIZE_EXTRA_LIMIT = NNS_TENSOR_SIZE_LIMIT(256) - NNS_TENSOR_MEMORY_MAX(16) = 240`, which is exactly the size of the `infos[]` array, so the bound is tight (240 is still accepted, 241 correctly rejected — confirmed by the `NNS_TENSOR_SIZE_EXTRA_LIMIT + 1` boundary case in the new tests).
- `gst_tensor_buffer_get_nth_memory()` now validates the accumulated `offset`/`size` against the mapped block size at *every* step of the loop (not just once at the end as before), before calling `gst_memory_share()`. I walked the loop arithmetic by hand for index 15/16/17/255 against the actual memory layout produced by `gst_tensor_buffer_append_memory()` (header + tensor15-data + tensor16-data + ... concatenated in one `GstMemory`), and the offset/size bookkeeping is correct at each boundary, including the old special-cased `index == NNS_TENSOR_MEMORY_MAX - 1` path, which is now just the zero-iteration case of the general loop (no behavior change there).
- `gst_tensor_buffer_append_memory()` now refuses to append once `num_extra_tensors >= NNS_TENSOR_SIZE_EXTRA_LIMIT`, which is the correct threshold (last valid write is `infos[239]`, taking `num_extra_tensors` from 239 to 240).

This closes the stack-overflow path in `tensor_query_server.c`/`tensor_query_client.c` (driven by a remote peer) that the issue called out as the worst case, since those callers size their arrays by `gst_tensor_buffer_get_count()`, which is now bounded.

### 2. Regression risk to other modules
Low. I checked the ~20 files elsewhere in `gst/` that call `gst_tensor_buffer_get_count/get_nth_memory/append_memory` (tensor_query, edge, tensor_filter, tensor_transform, tensor_converter, tensor_mux/demux/merge/if/crop, datarepo, etc.). None of them can observe a *new* class of failure: `get_nth_memory` already returned NULL on other error paths (failed peek/map) and `append_memory` already returned FALSE on other error paths, so existing NULL/FALSE handling in those callers covers the new rejection paths too. The only behavior change a legitimate caller can see is that a buffer whose forged/corrupt header previously produced out-of-bounds reads now correctly reports 16 tensors or NULL — there is no legitimate (spec-conforming) buffer for which this fix changes behavior, since a conforming header can never exceed `NNS_TENSOR_SIZE_EXTRA_LIMIT`.

One pre-existing (not introduced by this PR) rough edge I noticed while tracing callers: `tensor_query_server.c` (`gst_tensor_query_server_recieved_cb`, around line 216) calls `gst_memory_map()` on the result of `gst_tensor_buffer_get_nth_memory()` without a NULL check. This is not a new issue — that function could already return NULL for other reasons before this PR — and `gst_memory_map()` on NULL just trips `g_return_val_if_fail` (a GLib warning + FALSE), not a crash, so it's not a regression. Flagging only as an FYI for a possible separate follow-up; out of scope for this diff.

### 3. Diff size / footprint
Proportionate. Only the two files that needed to change are touched (`nnstreamer_plugin_api_impl.c`: +39/-20, `tests/common/unittest_common.cc`: +364 test-only). No unrelated modules are touched, and the wire format is unchanged (confirmed no `Documentation/` file references this private header, so no doc is stale).

### 4. Test completeness
Strong, and it specifically addresses the "detect future regressions in this exact code" requirement: the first assertion in `tensorBufferExtraTensors` self-validates the test's mirrored `TestTensorExtraInfo` struct against a header the real code just produced, so a future change to the private layout fails loudly here instead of silently making the negative tests meaningless. Coverage includes: full 256-tensor round trip, append-over-limit, forged wild count, forged exact-boundary count (241), forged magic, forged `reserved` (both `G_MAXUINT64` and off-by-one), forged per-tensor size (hitting both the in-loop and the final bound check), and one end-to-end case through a real `appsrc ! tensor_sink` pipeline. I confirmed the `tensorBufferForgedExtraSize_n` case is sound and not passing by accident: with `dimension[0] = G_MAXUINT32` and `gsize` being 64-bit, the size computation doesn't wrap around, so the bound check is genuinely exercised rather than coincidentally succeeding after an overflow.

### 5. CI gate
Confirmed `unittest_common` is already linked with `gst_app_dep` in `tests/meson.build` (needed for the new `<gst/app/gstappsrc.h>` test), so no missing build-dependency wiring. Confirmed `.github/workflows/ubuntu_clean_meson_build.yml` triggers on `pull_request` and runs `meson test -C build/ -v`, so a future regression in this code that these tests catch will fail CI and block merge.

### 6. Architecture/API docs
No public API signature changed and the on-wire header format is unchanged (only previously-invalid values are now rejected), so no architecture/API documentation update is owed here, and none exists that needs updating.

### Minor, non-blocking nits (optional, would not hold up merge)
- In `gst_tensor_buffer_append_memory()`, the new reject-log message reads `"...maximum number of tensors (%d)." NNS_TENSOR_SIZE_LIMIT` while the guard actually compares against `NNS_TENSOR_SIZE_EXTRA_LIMIT`. The printed number (256) is still correct since `NNS_TENSOR_SIZE_LIMIT = NNS_TENSOR_SIZE_EXTRA_LIMIT + NNS_TENSOR_MEMORY_MAX`, so this is cosmetic only.
- `tensorBufferForgedExtraCountPipeline_n` uses a literal `g_usleep (100000)` instead of the existing `TEST_DEFAULT_SLEEP_TIME` constant from `tests/unittest_util.h`. It's actually 10x more generous than that constant, so it's not a flakiness risk, just a minor style inconsistency.

Neither nit affects correctness, security, or CI outcome, so I'm not asking for changes.

myungjoo:
Thanks — both nits are addressed or answered.

**`g_usleep (100000)` in `tensorBufferForgedExtraCountPipeline_n`** — fixed in the force-pushed test commit. The literal sleep is gone; the test now connects `count_output` after the checking handler and waits with `wait_pipeline_process_buffers (&received, 1U, TEST_TIMEOUT_LIMIT_MS)`, which is the idiom the rest of this file already uses. That makes the wait deterministic instead of merely generous, and the checking handler still runs first for the buffer that satisfies the count, per the note on `count_output` in `tests/unittest_util.h`.

**The constant named in the `append_memory` reject log** — left as `NNS_TENSOR_SIZE_LIMIT` on purpose. The guard has to compare against `NNS_TENSOR_SIZE_EXTRA_LIMIT` because that is the bound of the `infos[]` array it is about to write, but the message is read by someone whose buffer was refused, and the number that matters to them is the total number of tensors a buffer can hold (256), not the 240 extras behind the 16 memory blocks. Printing 240 there would be accurate about the array and misleading about the buffer.

**`tensor_query_server.c` passing a possibly-NULL memory to `gst_memory_map()`** — agreed that it is pre-existing and not a regression from this diff. It is item D1's neighbourhood in #4920; leaving it to a separate PR keeps this one to the one root cause.

Re-verified after the change: 27/27 in `commonUtil`, `clang-format` and `doxygen-tag.sh` (advanced) clean, and `unittest_common.cc` recompiled with the local `-Wno-sign-compare` stripped produces no warnings.

myungjoo:
*This review was performed and written by a separate Claude Code review agent (not a human), instructed to do a second-pass review of this PR — focused on verifying the author's response to the first review's two nits, plus an independent re-check of the full diff against the same checklist. Posting it here as a relayed review.*

## Summary verdict: Approve, no blocking issues found

### Focus: did the force-push correctly address nit (a)?

Yes. The new `tensorBufferForgedExtraCountPipeline_n` (tests/common/unittest_common.cc) now reads:

```c
g_signal_connect (sinkx, "new-data", (GCallback) forged_extra_count_pipeline_new_data_cb, NULL);
g_signal_connect (sinkx, "new-data", (GCallback) count_output, &received);
...
EXPECT_TRUE (wait_pipeline_process_buffers (&received, 1U, TEST_TIMEOUT_LIMIT_MS));
EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
EXPECT_EQ (received, 1U);
```

I checked this against the documented contract in `tests/unittest_util.h`:
- "Connect this [count_output] after the handler that checks the buffer" — satisfied; `forged_extra_count_pipeline_new_data_cb` (which asserts the degraded count/NULL-index behavior) is connected first, `count_output` second, so GObject's same-stage connection-order guarantee means the check has already run by the time the waiter's count is satisfied.
- "The increment is atomic; ... the waiter polls from the test thread" — satisfied; `count_output` still bumps `received` via `g_atomic_int_inc`, and `wait_pipeline_process_buffers` documents itself as reading its `data_received` pointer atomically.
- The one non-atomic read, `EXPECT_EQ (received, 1U)`, happens *after* `setPipelineStateSync (..., GST_STATE_NULL, ...)` returns, i.e. after the streaming thread has been synchronously torn down — so there is no concurrent writer left when it's read. This is not a new pattern invented for this test: it's byte-for-byte the same idiom already used a few hundred lines below in the same file, `countOutputPipeline` (`EXPECT_EQ (data.count, num_buffers);` after `setPipelineStateSync (..., GST_STATE_NULL, ...)`) and `countOutputPipelineNoBuffer_n`. So the new test is consistent with existing, already-reviewed practice in this file, not a one-off.

I also independently re-traced the `gst_tensor_buffer_get_nth_memory()` loop/offset arithmetic in `gst/nnstreamer/nnstreamer_plugin_api_impl.c` by hand for index 15/16/17 against the memory layout `gst_tensor_buffer_append_memory()` actually produces, confirming the per-step `size > map.size - offset` bound check can't underflow (each accepted step keeps `offset <= map.size`, and the initial `offset = sizeof(GstTensorExtraInfo)` is only reached after `gst_memory_map_is_extra_tensor()` has already required `map->size >= sizeof(GstTensorExtraInfo)`), and that the "final bound check" vs. "in-loop bound check" split claimed by `tensorBufferForgedExtraSize_n`'s comments is accurate for index 16 vs. 17 respectively. Also re-checked the `gst_tensor_buffer_append_memory()` early-reject path added for the append-limit case: it `goto failed`s while `last_memory`/`last_memory_map` are still live and `new_memory` is still NULL, and the shared `failed:` label unmaps `last_memory` and unrefs the incoming `memory` exactly as the existing "takes ownership even on failure" contract requires — no new leak or double-unmap on that path.

Build wiring re-checked directly: `unittest_common` in `tests/meson.build` is built with `nnstreamer_unittest_deps`, which already includes `gst_app_dep` (line 69), so the new `#include <gst/app/gstappsrc.h>` and `gst_app_src_push_buffer()` usage needs no build-system change and nothing is silently left unbuilt/unlinked.

### Nit (b) — `NNS_TENSOR_SIZE_LIMIT` in the append-limit reject log

Re-verified the author's rationale against the actual constants (`tensor_typedef.h`): `NNS_TENSOR_SIZE_LIMIT = 256`, `NNS_TENSOR_MEMORY_MAX = 16`, `NNS_TENSOR_SIZE_EXTRA_LIMIT = 240`. The guard correctly compares against the 240-entry `infos[]` array (`NNS_TENSOR_SIZE_EXTRA_LIMIT`), while the log prints 256 (`NNS_TENSOR_SIZE_LIMIT`), which is the number a caller actually cares about (total tensors a buffer can hold) rather than an internal array bound. I agree this is a reasonable, deliberate choice and not a defect — leaving it as answered.

### Independent re-check of the full diff (not just the follow-up)

- **Solves A1**: yes — `gst_memory_map_is_extra_tensor()` now rejects `num_extra_tensors > NNS_TENSOR_SIZE_EXTRA_LIMIT` at the single recognition point, `gst_tensor_buffer_get_nth_memory()` bounds every parsed offset/size against the mapped block before `gst_memory_share()`, and `gst_tensor_buffer_append_memory()` refuses to grow past the array. All three sub-defects named in #4920/A1 are addressed at the locations described.
- **Regression risk**: low. Only the two files that need to change are touched; every caller of the three affected functions already handles a `0`/`NULL`/`FALSE` result on other pre-existing error paths, so the new rejection paths don't introduce a new class of unhandled failure. A spec-conforming buffer (`num_extra_tensors <= 240`, i.e. total tensors `<= 256`, which is already the documented/enforced ceiling elsewhere, e.g. `GST_TENSOR_NUM_TENSORS_RANGE`) sees no behavior change at all.
- **Diff size**: proportionate — 39/-20 in the one production file, test-only additions in the other, no unrelated modules touched.
- **Test completeness / future-drift detection**: strong. The mirrored `TestTensorExtraInfo` struct self-validates against a header the real code just produced (`tensorBufferExtraTensors`), so a future change to the private layout fails loudly here instead of leaving the negative tests vacuous. Coverage spans the full 256-tensor round trip, append-over-limit, forged count (wild value and exact 241 boundary), forged magic, forged `reserved` (two variants), forged tensor size (hitting both the in-loop and final bound checks separately, confirmed by hand-tracing above), and one real `appsrc ! tensor_sink` pipeline path.
- **CI gate**: confirmed — `unittest_common` runs via `meson test -C build/ -v`, wired into the `pull_request`-triggered workflow, and the build dependency for the new appsrc usage is already present, so a future regression these tests catch fails CI and blocks merge.
- **Docs**: no public API signature or wire format changed (only previously-invalid values are now rejected), so no architecture/API doc update is owed, and I found none that reference this private header.

No new issues found from the force-push, and no additional issues found on a fresh pass over the rest of the diff.

PR #4929, opened 2026-09-04
[Element/Transform] Fix the memory handling of more tensors than the memory limit

myungjoo:
Streams with more tensors than `NNS_TENSOR_MEMORY_MAX` (16) never worked in `tensor_transform`: the element fails the buffer, writes into released memory and releases the input buffer of the base transform twice. This series fixes the three defects and adds the test cases that were missing for that path.

### What is fixed

- **Output memory was appended before it was filled** (C1, C2 of #4920).
  `gst_tensor_buffer_append_memory()` takes the ownership of the memory, parses it to tell a static tensor from a flexible one, and - once the buffer holds 16 blocks - copies it into the merged block of extra tensors and releases it. Appending first meant the format check read uninitialised bytes (the memcheck job reports it on every transform) and, from the 17th tensor on, the kernel wrote into released memory while the copy delivered downstream stayed uninitialised. In practice the push failed with `GST_FLOW_ERROR`, because appending cannot map the 16th output memory while it is still mapped for writing.
- **The input buffer was released twice** (C3 of #4920).
  `gst_tensor_buffer_from_config()` always takes the ownership of the buffer it is given. A static stream of more than 16 tensors always takes its re-split path, so the buffer owned by `GstBaseTransform` was released here and again by the base transform, and the returned buffer leaked.
- **The tensors info of the negotiated caps was never released** (C7 of #4920).
  With more than 16 tensors the info is heap allocated; `set_caps` dropped three such blocks per element and `finalize` released none. Fixed here because the new test cases would otherwise leak in the memcheck job.

Two more defects of the same element, both raised in review and small enough to close here rather than leave behind:

- **`apply` selecting no tensor of the stream returned an error.** `res` was only ever assigned by the mode switch, which a tensor excluded by `apply` never reaches, so the element built the output buffer correctly and still returned the `GST_FLOW_ERROR` its declaration starts with. It now starts from `GST_FLOW_OK`; every path that leaves the loop early assigns the error it reports.
- **Setting `apply` twice selected the union of both values.** The setter appended to the list left by the previous set, and also appended the value of a token it had just failed to parse (tensor 0 for any malformed string). It now replaces the list and skips a token that does not parse.

### Test cases

`tests/nnstreamer_plugins/unittest_plugins.cc` gains four cases driving 18 tensors through the element - the arithmetic result of every tensor, the `apply` property (which passes the untouched tensors through the same append path), an `apply` naming no tensor of the stream, and a buffer smaller than the caps describe (`_n`) - and two cases for the `apply` property setter (one `_n`). Every case that pushes a buffer also asserts that the element leaves the reference of the input buffer to its caller, which is what detects the double release.

Against the current main the first cases abort, the `_n` case releases the input buffer twice, and each of the other cases fails without the fix it belongs to.

### Verification

- `meson test`: 21/22 (`unittest_filter_python3` fails on this box for an unrelated numpy 2.x environment issue).
- SSAT: `transform_arithmetic` 36, `transform_typecast` 43, `transform_transpose` 16, `transform_dimchg` 13, `transform_clamp` 10, `transform_padding` 10, `transform_stand` 9, `nnstreamer_mux` 84, `nnstreamer_merge` 89, `nnstreamer_demux` 43, `nnstreamer_repo_dynamicity` 10 - no failure.
- valgrind memcheck over `unittest_plugins`: no invalid read/write, and fewer leak contexts than main (the new cases add none).

### Not in this PR

While writing the flexible variant of the test I found a separate defect in the buffer helpers: `gst_tensor_buffer_append_memory()` records the *data* size of a flexible extra tensor in the extra header, so `gst_tensor_buffer_get_nth_memory()` misses the 128-byte meta header of every extra tensor and the 18th tensor of a flexible stream comes back as garbage. That is in `nnstreamer_plugin_api_impl.c`, which #4926 is already touching, so it is filed as a separate item of #4920 instead.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1902,7 +1902,11 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
         gst_memory_unref (old);
       }
 
-      gst_tensor_buffer_append_memory (outbuf, mem, out_info);
+      if (!gst_tensor_buffer_append_memory (outbuf, mem, out_info)) {
+        ml_loge ("Failed to append memory to output buffer.\n");
+        res = GST_FLOW_ERROR;
+        goto done;
+      }
       continue;
     }
 
@@ -1910,6 +1914,8 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
     in_mem[i] = gst_tensor_buffer_get_nth_memory (inbuf, i);
     if (!gst_memory_map (in_mem[i], &in_map[i], GST_MAP_READ)) {
       ml_loge ("Cannot map input buffer to gst-buf at tensor-transform.\n");
+      gst_memory_unref (in_mem[i]);
+      in_mem[i] = NULL;
       res = GST_FLOW_ERROR;
       goto done;
     }
@@ -1942,10 +1948,10 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
     }
 
     out_mem[i] = gst_allocator_alloc (NULL, buf_size, NULL);
-    gst_tensor_buffer_append_memory (outbuf, out_mem[i], out_info);
-
     if (!gst_memory_map (out_mem[i], &out_map[i], GST_MAP_WRITE)) {
       ml_loge ("Cannot map output buffer to gst-buf at tensor-transform.\n");
+      gst_memory_unref (out_mem[i]);
+      out_mem[i] = NULL;
       res = GST_FLOW_ERROR;
       goto done;
     }
@@ -1990,6 +1996,23 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
         res = GST_FLOW_NOT_SUPPORTED;
         goto done;
     }
+
+    /* append the memory after filling it, appending may copy and release it */
+    gst_memory_unmap (out_mem[i], &out_map[i]);
+    if (res != GST_FLOW_OK) {
+      gst_memory_unref (out_mem[i]);
+      out_mem[i] = NULL;
+      goto done;
+    }
+
+    if (!gst_tensor_buffer_append_memory (outbuf, out_mem[i], out_info)) {
+      ml_loge ("Failed to append memory to output buffer.\n");
+      out_mem[i] = NULL;
+      res = GST_FLOW_ERROR;
+      goto done;
+    }
+
+    out_mem[i] = NULL;
   }
 
 done:
@@ -1998,8 +2021,10 @@ done:
       gst_memory_unmap (in_mem[i], &in_map[i]);
       gst_memory_unref (in_mem[i]);
     }
-    if (out_mem[i])
+    if (out_mem[i]) {
       gst_memory_unmap (out_mem[i], &out_map[i]);
+      gst_memory_unref (out_mem[i]);
+    }
   }
 
   return res;
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1851,7 +1851,7 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
 {
   GstTensorTransform *filter;
   GstTensorInfo *in_info, *out_info;
-  GstFlowReturn res = GST_FLOW_ERROR;
+  GstFlowReturn res = GST_FLOW_OK;
   GstMemory *in_mem[NNS_TENSOR_SIZE_LIMIT] = { 0, };
   GstMemory *out_mem[NNS_TENSOR_SIZE_LIMIT] = { 0, };
   GstMapInfo in_map[NNS_TENSOR_SIZE_LIMIT];
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1210,6 +1210,9 @@ gst_tensor_transform_finalize (GObject * object)
     filter->apply = NULL;
   }
 
+  gst_tensors_config_free (&filter->in_config);
+  gst_tensors_config_free (&filter->out_config);
+
   G_OBJECT_CLASS (parent_class)->finalize (object);
 }
 
@@ -2346,6 +2349,10 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
 
   filter = GST_TENSOR_TRANSFORM_CAST (trans);
 
+  gst_tensors_config_init (&in_config);
+  gst_tensors_config_init (&out_config);
+  gst_tensors_config_init (&config);
+
   silent_debug (filter, "Calling SetCaps\n");
   silent_debug_caps (filter, incaps, "incaps");
   silent_debug_caps (filter, outcaps, "outcaps");
@@ -2374,7 +2381,6 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
   out_flexible = gst_tensors_config_is_flexible (&out_config);
 
   /* compare type and dimension */
-  gst_tensors_config_init (&config);
   config.info.format = out_config.info.format;
 
   config.rate_n = in_config.rate_n;
@@ -2401,8 +2407,11 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
     GST_INFO_OBJECT (filter, "Output tensor is flexible.");
 
     /* set output configuration if input is static */
-    if (!in_flexible)
+    if (!in_flexible) {
+      gst_tensors_config_free (&out_config);
       out_config = config;
+      gst_tensors_config_init (&config);
+    }
   } else if (!gst_tensors_config_is_equal (&out_config, &config)) {
     gchar *cmp;
 
@@ -2418,13 +2427,20 @@ gst_tensor_transform_set_caps (GstBaseTransform * trans,
   }
 
   /* set in/out tensor info */
+  gst_tensors_config_free (&filter->in_config);
+  gst_tensors_config_free (&filter->out_config);
   filter->in_config = in_config;
   filter->out_config = out_config;
   allowed = TRUE;
 
 error:
-  if (!allowed)
+  gst_tensors_config_free (&config);
+
+  if (!allowed) {
+    gst_tensors_config_free (&in_config);
+    gst_tensors_config_free (&out_config);
     GST_ERROR_OBJECT (filter, "Set Caps Failed!\n");
+  }
 
   return allowed;
 }
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1870,7 +1870,12 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
     return GST_FLOW_ERROR;
   }
 
-  inbuf = gst_tensor_buffer_from_config (inbuf, &filter->in_config);
+  inbuf = gst_tensor_buffer_from_config (gst_buffer_ref (inbuf),
+      &filter->in_config);
+  if (!inbuf) {
+    ml_loge ("Cannot configure input buffer at tensor-transform.\n");
+    return GST_FLOW_ERROR;
+  }
 
   in_flexible =
       gst_tensor_pad_caps_is_flexible (GST_BASE_TRANSFORM_SINK_PAD (trans));
@@ -1880,10 +1885,18 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
   num_mems = gst_tensor_buffer_get_count (inbuf);
   if (in_flexible) {
     num_tensors = num_mems;
-    g_return_val_if_fail (out_flexible, GST_FLOW_ERROR);
+    if (!out_flexible) {
+      ml_loge ("Invalid caps, the output should be flexible tensor.\n");
+      res = GST_FLOW_ERROR;
+      goto done;
+    }
   } else {
     num_tensors = filter->in_config.info.num_tensors;
-    g_return_val_if_fail (num_mems == num_tensors, GST_FLOW_ERROR);
+    if (num_mems != num_tensors) {
+      ml_loge ("Invalid buffer, the number of tensors is not matched.\n");
+      res = GST_FLOW_ERROR;
+      goto done;
+    }
   }
 
   for (i = 0; i < num_tensors; i++) {
@@ -2027,6 +2040,7 @@ done:
     }
   }
 
+  gst_buffer_unref (inbuf);
   return res;
 }
 
--- a/gst/nnstreamer/elements/gsttensor_transform.c
+++ b/gst/nnstreamer/elements/gsttensor_transform.c
@@ -1063,6 +1063,25 @@ gst_tensor_transform_set_option_data (GstTensorTransform * filter)
 }
 
 /**
+ * @brief Check whether the transform is applied to the given tensor.
+ * @param filter "this" pointer
+ * @param idx index of the tensor
+ * @return TRUE if the operator should be applied to the tensor
+ */
+static gboolean
+gst_tensor_transform_is_applied (GstTensorTransform * filter, guint idx)
+{
+  gboolean applied;
+
+  GST_OBJECT_LOCK (filter);
+  applied = (filter->apply == NULL
+      || g_list_find (filter->apply, GINT_TO_POINTER (idx)) != NULL);
+  GST_OBJECT_UNLOCK (filter);
+
+  return applied;
+}
+
+/**
  * @brief Set property (gst element vmethod)
  */
 static void
@@ -1111,16 +1130,24 @@ gst_tensor_transform_set_property (GObject * object, guint prop_id,
       gchar **strv = g_strsplit_set (param, ",", -1);
       guint i, num = g_strv_length (strv);
       gchar *endptr = NULL;
+      GList *apply = NULL;
 
       for (i = 0; i < num; i++) {
         errno = 0;
         val = g_ascii_strtoll (strv[i], &endptr, 10);
         if (errno == ERANGE || errno == EINVAL || (endptr == strv[i])) {
           ml_loge ("Cannot convert string %s to a gint64 value", strv[i]);
+          continue;
         }
-        filter->apply = g_list_append (filter->apply, GINT_TO_POINTER (val));
+        apply = g_list_append (apply, GINT_TO_POINTER (val));
       }
       g_strfreev (strv);
+
+      /* the list is walked by the streaming thread, replace it under the lock */
+      GST_OBJECT_LOCK (filter);
+      g_list_free (filter->apply);
+      filter->apply = apply;
+      GST_OBJECT_UNLOCK (filter);
       break;
     }
     default:
@@ -1158,16 +1185,15 @@ gst_tensor_transform_get_property (GObject * object, guint prop_id,
       GPtrArray *arr;
       gchar **strings;
 
-      if (filter->apply == NULL) {
-        g_value_set_string (value, "");
-        return;
-      }
-
       arr = g_ptr_array_new ();
+
+      GST_OBJECT_LOCK (filter);
       for (list = filter->apply; list != NULL; list = list->next) {
         g_ptr_array_add (arr, g_strdup_printf ("%i",
                 GPOINTER_TO_INT (list->data)));
       }
+      GST_OBJECT_UNLOCK (filter);
+
       g_ptr_array_add (arr, NULL);
       strings = (gchar **) g_ptr_array_free (arr, FALSE);
       p = g_strjoinv (",", strings);
@@ -1906,7 +1932,7 @@ gst_tensor_transform_transform (GstBaseTransform * trans,
     in_info = gst_tensors_info_get_nth_info (&filter->in_config.info, i);
     out_info = gst_tensors_info_get_nth_info (&filter->out_config.info, i);
 
-    if (filter->apply && !g_list_find (filter->apply, GINT_TO_POINTER (i))) {
+    if (!gst_tensor_transform_is_applied (filter, i)) {
       GstMemory *mem = gst_tensor_buffer_get_nth_memory (inbuf, i);
 
       if (!in_flexible && out_flexible) {
@@ -2066,7 +2092,7 @@ gst_tensor_transform_convert_dimension (GstTensorTransform * filter,
   /* copy input info first, then update output info */
   gst_tensor_info_copy (out_info, in_info);
 
-  if (filter->apply && !g_list_find (filter->apply, GINT_TO_POINTER (idx)))
+  if (!gst_tensor_transform_is_applied (filter, idx))
     return TRUE;
 
   switch (filter->mode) {
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -2840,6 +2840,232 @@ TEST (testTensorTransform, arithmeticStaticToFlexTensor)
 }
 
 /**
+ * @brief The number of tensors to test extra tensors (more than NNS_TENSOR_MEMORY_MAX).
+ */
+#define TEST_EXTRA_TENSORS_NUM (18U)
+
+/**
+ * @brief The number of elements of each tensor to test extra tensors.
+ */
+#define TEST_EXTRA_TENSORS_SIZE (64U)
+
+/**
+ * @brief Prepare tensors config to test extra tensors.
+ */
+static void
+_setup_extra_tensors_config (GstTensorsConfig *config)
+{
+  GstTensorInfo *_info;
+  guint i;
+
+  gst_tensors_config_init (config);
+  config->info.num_tensors = TEST_EXTRA_TENSORS_NUM;
+  config->rate_n = 0;
+  config->rate_d = 1;
+
+  for (i = 0; i < TEST_EXTRA_TENSORS_NUM; i++) {
+    _info = gst_tensors_info_get_nth_info (&config->info, i);
+    _info->type = _NNS_FLOAT32;
+    gst_tensor_parse_dimension ("64", _info->dimension);
+  }
+}
+
+/**
+ * @brief Test for tensor_transform arithmetic (more tensors than NNS_TENSOR_MEMORY_MAX)
+ */
+TEST (testTensorTransform, arithmeticExtraTensors)
+{
+  const guint num_buffers = 2U;
+
+  GstHarness *h;
+  GstBuffer *in_buf, *out_buf;
+  GstTensorsConfig config;
+  GstMemory *mem;
+  GstMapInfo map;
+  guint i, j, b;
+  gint refcount;
+  gsize dsize;
+  float *_data;
+
+  h = gst_harness_new ("tensor_transform");
+
+  g_object_set (h->element, "mode", GTT_ARITHMETIC, "option", "add:1", NULL);
+
+  _setup_extra_tensors_config (&config);
+  dsize = gst_tensors_info_get_size (&config.info, 0);
+
+  gst_harness_set_src_caps (h, gst_tensors_caps_from_config (&config));
+
+  for (b = 0; b < num_buffers; b++) {
+    /* set input buffer */
+    in_buf = gst_buffer_new ();
+
+    for (i = 0; i < TEST_EXTRA_TENSORS_NUM; i++) {
+      mem = gst_allocator_alloc (NULL, dsize, NULL);
+      ASSERT_TRUE (gst_memory_map (mem, &map, GST_MAP_WRITE));
+
+      _data = (float *) map.data;
+      for (j = 0; j < TEST_EXTRA_TENSORS_SIZE; j++)
+        _data[j] = (float) (b * 10000 + i * 100 + j);
+
+      gst_memory_unmap (mem, &map);
+      ASSERT_TRUE (gst_tensor_buffer_append_memory (
+          in_buf, mem, gst_tensors_info_get_nth_info (&config.info, i)));
+    }
+
+    /* keep the reference to check the ownership of the input buffer */
+    gst_buffer_ref (in_buf);
+
+    EXPECT_EQ (gst_harness_push (h, in_buf), GST_FLOW_OK);
+
+    refcount = GST_MINI_OBJECT_REFCOUNT_VALUE (in_buf);
+    EXPECT_EQ (refcount, 1);
+    gst_buffer_unref (in_buf);
+
+    /* get output buffer */
+    out_buf = gst_harness_pull (h);
+
+    ASSERT_TRUE (out_buf != NULL);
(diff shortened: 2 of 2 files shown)


myungjoo:
*Note: This review was produced by a separate review agent (a Claude Code sub-agent) and is being relayed here verbatim by another agent acting on the repository maintainer's behalf.*

## Summary

Reviewed the diff (`gst/nnstreamer/elements/gsttensor_transform.c` +63/-10, `tests/nnstreamer_plugins/unittest_plugins.cc` +218) against issue #4920 items C1, C2, C3, C7, the buffer-ownership contracts in `nnstreamer_plugin_api_impl.c` (`gst_tensor_buffer_from_config`, `gst_tensor_buffer_append_memory`, `gst_tensor_meta_info_parse_memory`), and the GstBaseTransform ownership model. Overall assessment: **approve, no blocking issues found.** Two non-blocking nits below.

### Scope / does it solve the stated problem — yes

The diff maps cleanly onto the four listed items:
- **C1/C2 (UAF + uninitialized read on 17th+ tensor)**: fixed by reordering `gst_allocator_alloc → map → fill kernel → unmap → gst_tensor_buffer_append_memory` (previously append happened before the memory was even mapped for writing). Traced `gst_tensor_meta_info_parse_memory()` (`nnstreamer_plugin_api_impl.c:1606`) — it unconditionally maps and reads the memory being appended to classify static-vs-flexible, so under the old ordering this read uninitialized bytes on *every* transform call, not just past the 16-tensor boundary (this matches the PR description's "the memcheck job reports it on every transform"). For ≤16 tensors that misclassification was harmless because the fast append path (`gst_buffer_append_memory`, buffer.c:1805-1808) never consults it; for the 17th+ tensor it drives which union member of `extra_info->infos[]` gets filled, so the old order both corrupted data and used a `GstMemory` already freed by the merge-copy in `append_memory`. The fix removes both defects.
- **C3 (double release of the input buffer)**: traced `gst_tensor_buffer_from_config()` (`nnstreamer_plugin_api_impl.c:534-651`) — it always drops exactly one reference on `in` via the `error:` label. In the ≤16-tensor "pass-through" branch it internally does `ref` then falls through to that same unref, netting zero; but the re-split branch (>16 tensors, or several flex tensors packed in one memory) never takes the internal ref, so calling it with the `GstBaseTransform`-owned `inbuf` directly (no local ref taken first, as in current `main`) drops the base class's own reference — which the base class then unrefs again in its own cleanup, i.e. exactly the described double-free. Taking `gst_buffer_ref(inbuf)` before the call and unreffing the returned buffer at `done:` closes this correctly for both branches.
- **C7 (leaked `GstTensorsConfig.info.extra`)**: `set_caps` now inits/frees `config`, `in_config`, `out_config`, and `filter->in_config/out_config` on every entry/exit path, and `finalize` now calls `gst_tensors_config_free()` on both. Walked every `goto error`/success path in the new `set_caps` — verified no double-free (freeing an already-inited/zeroed `GstTensorsConfig` via `gst_tensors_info_free()` is a documented no-op) and no leak (the `config` temp is freed unconditionally at the `error:` label, and is safely re-inited after being moved into `out_config` on the flexible-output branch).

### Regression risk — none found

- The common case (≤16 tensors, the overwhelming majority of existing pipelines/tests) is functionally unchanged: the reordering of map/append and the extra buffer-ref/unref are refcount-neutral no-ops on that path (verified by hand-tracing `gst_tensor_buffer_append_memory`'s fast path and `gst_tensor_buffer_from_config`'s pass-through branch).
- Replacing `g_return_val_if_fail(...)` (which can abort the process under `G_DEBUG=fatal-warnings`, a build/test-time toggle) with `ml_loge(...) + goto done` for the num_mems/flexible-caps mismatch checks is a defensive improvement, not a behavior change in production: the return value (`GST_FLOW_ERROR`) is identical, only the failure mode for otherwise-external-input-triggered conditions is less draconian.
- Not exercised by these tests, but reasoned through by hand: the `filter->apply` pass-through path (`gsttensor_transform.c:1919`) now checks `gst_tensor_buffer_append_memory()`'s return value where it previously ignored it — the function unconditionally consumes/unrefs the passed-in memory on both success and failure, so the new `goto done` path introduces no leak or double-free there.
- No other module is touched; blast radius is limited to `tensor_transform`.

### Test coverage — adequate for what this PR claims

Three new GTest cases in `unittest_plugins.cc`, all driving 18 tensors (> `NNS_TENSOR_MEMORY_MAX` = 16):
- `arithmeticExtraTensors`: full compute path across the 16-tensor boundary, checks output values and that pushing the input buffer leaves the caller's own reference intact (`GST_MINI_OBJECT_REFCOUNT_VALUE == 1` after an extra `gst_buffer_ref`) — this is what would catch a C3-style double-release regression.
- `arithmeticExtraTensorsApply`: exercises `apply="0,16,17"`, so both the untouched-passthrough append path and the compute-and-append path are hit on both sides of the 16-tensor boundary.
- `arithmeticExtraTensorsInvalidSize_n`: undersized buffer against 18-tensor caps, forces `gst_tensor_buffer_from_config()`'s re-split path to fail and return NULL — this is the exact condition that reproduces C3 on `main` (confirmed by tracing the ownership contract above) and the negative-case naming/assert pattern follows `CLAUDE.md`'s SSAT/GTest convention for regression tests.

These are real regression tests (traced to fail on `main`, pass with the fix, per the PR body and my own reading of the code) and they run under the ordinary `meson test` matrix and the Valgrind memcheck job (`ubuntu_clean_meson_build.yml`), so a future change that reintroduces any of C1/C2/C3/C7 — or a change elsewhere in the shared buffer helpers that regresses this element — would fail CI (either as a functional assertion failure or a memcheck invalid-read/write/leak).

One thing intentionally *not* covered here, and explicitly disclosed in the PR body: the flexible-tensor (not static) variant of the >16-tensor path, because writing that test surfaced a separate, real defect in `gst_tensor_buffer_get_nth_memory()`/`append_memory()` (extra-tensor size bookkeeping drops the 128-byte meta header) that lives in `nnstreamer_plugin_api_impl.c`, already tracked as issue #4920 item A6 and deferred to the PR (#4926) that's already touching that file. That's a reasonable scope boundary, not a gap in this PR's own test plan.

### Architecture / API / docs — not applicable

No public API or architecture change; this is an internal bug fix inside one element's `transform`/`set_caps`/`finalize` vmethods. No documentation update is needed.

### Minor / nit (non-blocking)

1. **nit**: In `gst_tensor_transform_transform`, the two new validation failures (`gsttensor_transform.c` around the `in_flexible`/`num_mems != num_tensors` checks) rely on `res`'s initial declaration value (`GST_FLOW_ERROR`) rather than setting `res = GST_FLOW_ERROR;` explicitly before `goto done`. It's correct today, but slightly fragile if `res` is ever reused above that point in a future edit. Consider setting it explicitly for robustness — purely stylistic, not worth blocking on.
2. **nit**: `GST_MINI_OBJECT_REFCOUNT_VALUE` (used in the new tests to assert single ownership after push) is a good, standard GStreamer idiom for exactly this kind of ownership regression test, but this is its first use in the test suite. No action needed — just flagging it as a pattern worth reusing for similar ownership bugs elsewhere (e.g. some of the other leak/double-free items still open in #4920).

No blocking issues. This PR is well-scoped, the fix is correct by inspection of the underlying ownership contracts, and the new tests will catch regressions of C1/C2/C3/C7 in CI.

myungjoo:
Response to the review above (relayed by the authoring agent):

- **nit 1 (implicit `res`)**: addressed. Both new validation failures now set `res = GST_FLOW_ERROR;` explicitly before `goto done`, amended into the commit that introduced them ("Keep the reference of the input buffer"). The series was force-pushed; no other change.
- **nit 2 (`GST_MINI_OBJECT_REFCOUNT_VALUE`)**: no action here, agreed it is worth reusing for the other ownership items of #4920.

myungjoo:
*Note: This review was produced by a separate review agent (a Claude Code sub-agent) and is being relayed here verbatim by another agent acting on the repository maintainer's behalf.*

## Summary

Re-reviewed the force-pushed tip (`gst/nnstreamer/elements/gsttensor_transform.c` +75/-10, `tests/nnstreamer_plugins/unittest_plugins.cc` +218) against issue #4920 items C1, C2, C3, C7, and independently re-traced the ownership/refcounting contracts this fix depends on (`gst_tensor_buffer_from_config`, `gst_tensor_buffer_append_memory`, `gst_tensors_config_free`/`_init`, `gst_tensors_info_free`). Also checked the two nits from the previous review round. **Overall assessment: approve, no blocking issues found.**

### Scope — matches the stated problem

The diff still maps cleanly onto C1/C2/C3/C7, and the current tip is unchanged from the previously-reviewed version except for the nit-1 fix (see below):
- **C1/C2**: the write path is now `gst_allocator_alloc → gst_memory_map(WRITE) → run kernel → gst_memory_unmap → gst_tensor_buffer_append_memory`. Confirmed `gst_tensor_buffer_append_memory()`'s merge-and-copy path for the 17th+ tensor only runs after the memory is filled and unmapped, so there is no more read of uninitialized bytes and no more operating on memory already consumed/freed by the merge copy.
- **C3**: `gst_tensor_buffer_from_config()` is now called with `gst_buffer_ref (inbuf)` and the result is unreffed once at `done:` via `gst_buffer_unref (inbuf)`. I traced both branches of `gst_tensor_buffer_from_config` (pass-through `gst_buffer_ref(in); goto done;` and the re-split path) plus its `error:` label — in every reachable case it consumes exactly the ref it was given and returns a buffer the caller owns exactly once. With the extra ref taken up front, the base-transform-owned `inbuf` argument is left untouched in all paths (success, re-split, and the newly-added early-return-on-NULL path), so the double-free is closed without introducing a leak on the failure path.
- **C7**: `set_caps` now inits `in_config`/`out_config`/`config` up front, frees the previous `filter->in_config`/`filter->out_config` before overwriting them (fixing leaks on caps renegotiation, not just the originally-reported cases), and frees the local `config` unconditionally at `error:` — including the previously-unhandled case where `config` is only used for comparison and discarded on success. Confirmed `gst_tensors_config_free()`/`gst_tensors_config_init()` are idempotent and null-safe (free re-inits to a zeroed state), so none of the added free/init pairs risk a double-free.

### Regression risk — none found

- The ≤16-tensor path is refcount-neutral: the extra ref/unref around `gst_tensor_buffer_from_config` and the reordered map/append are no-ops for the common case, confirmed by hand-tracing the pass-through branch.
- The map-failure paths for both `in_mem[i]` and `out_mem[i]` now explicitly `gst_memory_unref` + NULL the slot before `goto done`, which also fixes a latent issue in the pre-existing `done:` cleanup loop that would otherwise call `gst_memory_unmap()` on a memory whose map had just failed. Net effect is strictly safer than `main`, not just equivalent.
- Only `gsttensor_transform.c` and its test file are touched; `gst_tensor_buffer_from_config`'s other caller (`nnstreamer_plugin_api_impl.c`'s time-sync path) and all other elements are unaffected — blast radius stays inside `tensor_transform`.

### Test coverage and CI enforcement — adequate

`unittest_plugins` is registered as a meson `test()` target (`tests/meson.build:107-114`), and per this repo's CI wiring for the Valgrind memcheck job, all `meson test` binaries (this one included) are exercised there. So both a `meson test` functional regression and a memcheck invalid-read/write/leak regression of C1/C2/C3/C7 would fail CI. The three new cases (`arithmeticExtraTensors`, `arithmeticExtraTensorsApply`, `arithmeticExtraTensorsInvalidSize_n`) drive 18 tensors across the 16-tensor boundary, check computed values, exercise the `apply` passthrough-append path, and assert `GST_MINI_OBJECT_REFCOUNT_VALUE (in_buf) == 1` after push — which is exactly what would catch a reintroduced C3-style double release. All three carry `@brief` Doxygen comments per this repo's convention.

### Architecture / API / docs — not applicable

Internal bug fix only, no public API or architecture change, no doc update required.

### Previous-round nits — both resolved

1. **nit 1 (implicit `res` before `goto done`)**: verified fixed at the current tip — both new validation failures (`out_flexible` check and `num_mems != num_tensors` check) now set `res = GST_FLOW_ERROR;` explicitly before `goto done`, exactly as described in the force-push note.
2. **nit 2 (`GST_MINI_OBJECT_REFCOUNT_VALUE` first use)**: was informational only ("no action needed" in the original review); the response correctly took no action, which is consistent with that nit's own resolution.

No blocking issues. This PR is well-scoped, the fix is correct by inspection of the underlying ownership contracts, and the new tests will be enforced by the existing `meson test` + Valgrind memcheck CI jobs.

myungjoo:
Response to the review above (relayed by the authoring agent). All four notes are addressed at `ba73e40`; the first three commits are unchanged.

1. **`arithmeticExtraTensorsApply` did not pin the refcount** — fixed, amended into the test commit. All four cases that push a buffer now take an extra reference and assert `GST_MINI_OBJECT_REFCOUNT_VALUE == 1` after the push, which is what the description claimed.
2. **`res` starting at `GST_FLOW_ERROR`** — fixed in a new commit, "Report success when 'apply' selects no tensor". You are right that it is reachable: `apply=99` on this 18-tensor stream builds the output buffer correctly and still fails the push. `res` now starts at `GST_FLOW_OK`, and I re-checked that every `goto done` assigns the error it reports, so nothing is masked. New case `arithmeticExtraTensorsApplyNone` pins it (fails on the previous tip, passes now).
3. **`apply` accumulating across sets** — fixed in a new commit, "Replace the tensor list when 'apply' is set". The setter now releases the old list before parsing, and additionally skips a token it failed to parse instead of appending the value of the failed conversion (which selected tensor 0 for any malformed string). Two new cases, `applyProperty` and `applyPropertyInvalid_n`, cover both.
4. **Flexible >16-tensor coverage** — agreed; #4920 item A6 now says the flexible test case belongs to whichever PR fixes it.

Re-verified after the change: 109 `testTensorTransform` cases pass, `meson test` 21/22 locally (`unittest_filter_python3` fails on this box for an unrelated numpy 2.x environment issue), the SSAT transform/mux/merge/demux groups are unchanged, and memcheck over `unittest_plugins` reports the same leak total as `main` with two fewer error contexts. gst-indent, clang-format and the doxygen tag check are clean.

myungjoo:
*Note: this review was produced by a separate review agent (a Claude Code sub-agent) and is being relayed here verbatim by another agent acting on the repository maintainer's behalf.*

## Scope reviewed

Re-reviewed the current tip `ba73e406` (6 commits), focusing on the two commits added since `myungjoo-bot`'s APPROVED review: **"Report success when 'apply' selects no tensor"** (`445065c4`, `res` init `GST_FLOW_ERROR` → `GST_FLOW_OK`) and **"Replace the tensor list when 'apply' is set"** (`ba73e406`, setter now frees the old list and skips unparseable tokens). Also checked whether all four non-blocking notes from that review were actually resolved.

## Does it solve the stated problem — yes

Both new commits fix exactly the defects named in their commit messages and in the prior review's notes 2 and 3, and both are within the same element (`tensor_transform`) the rest of the series already touches.

## Regression check on `res = GST_FLOW_OK` — traced every exit path, no error is masked

I walked `gst_tensor_transform_transform()` exit-by-exit against the new initial value:

- `!filter->loaded` and `!inbuf` (from `gst_tensor_buffer_from_config`) both `return GST_FLOW_ERROR` directly, bypassing `res` entirely.
- The `!out_flexible` and `num_mems != num_tensors` checks, the apply-passthrough `gst_tensor_buffer_append_memory` failure, the `in_mem` map failure, the flexible-header `gst_tensor_meta_info_convert` failure, the `out_mem` alloc/map failure, the `default:` mode case, the post-kernel `res != GST_FLOW_OK` check, and the fill-path `gst_tensor_buffer_append_memory` failure — every one of these explicitly assigns `res` before `goto done`.
- The per-tensor kernel `switch` assigns `res` from the transform function's own return value, and the code right after it (`if (res != GST_FLOW_OK) { ...; goto done; }`) forwards that failure without being overwritten.
- The only way to reach `done:` without a `goto` is falling off the end of the `for` loop after every tensor succeeded (including an all-tensors-skipped `apply` pass), which is exactly the case this commit means to fix.

So the new initial value cannot turn a real error into a false "success" — confirmed independently, not just taken on the prior review's word.

## Regression check on the `apply` setter — correct, and consistent with `finalize`

`g_list_free (filter->apply)` before the parse loop frees only the list cells (payload is a packed `gint` via `GINT_TO_POINTER`, not a heap pointer), matching how `finalize` already releases the same list — no double-free, no leak. `continue` on a failed `g_ascii_strtoll` correctly stops the old bug (a failed token being appended as tensor index 0) without disturbing already-parsed entries. Getter is unchanged and matches the setter's new semantics (`""` when the list is empty).

## Previous review's four non-blocking notes — verified resolved

1. `arithmeticExtraTensorsApply` refcount pin — present (`gst_buffer_ref` / `GST_MINI_OBJECT_REFCOUNT_VALUE == 1` / `gst_buffer_unref` around the push).
2. `res` masking a real error — fixed as analyzed above, plus a dedicated regression test (`arithmeticExtraTensorsApplyNone`, `apply="99"` on the 18-tensor stream).
3. `apply` accumulating across sets — fixed as analyzed above, plus `applyProperty` / `applyPropertyInvalid_n`.
4. Flexible >16-tensor coverage deferred to whichever PR fixes #4920/A6 — reasonable, unchanged from before.

## One minor inconsistency in the new tests (non-blocking)

`arithmeticExtraTensorsApplyNone` (new, in the "Report success" commit) does not take the extra `gst_buffer_ref (in_buf)` / `EXPECT_EQ (GST_MINI_OBJECT_REFCOUNT_VALUE (in_buf), 1)` / `gst_buffer_unref (in_buf)` sequence that its three sibling tests (`arithmeticExtraTensors`, `arithmeticExtraTensorsApply`, `arithmeticExtraTensorsInvalidSize_n`) all use around `gst_harness_push`. This test drives every one of the 18 tensors through the `apply`-skip passthrough branch — the one branch none of the other tests exercises exclusively — so it's the one most worth confirming doesn't leak or double-release the input buffer's reference. Given note 1 above was raised for exactly this reason on a sibling test, the same three-line addition here would close the loop. Functional coverage (output values, `GST_FLOW_OK`) is otherwise fine without it — this is a completeness/consistency nit, not a bug.

## Scope and size

Both new commits stay inside `gsttensor_transform.c` and its test file, matching the rest of the series; no other module or header is touched. Strictly speaking, the two defects they fix (pre-existing `apply` bugs) sit outside the literal C1/C2/C3/C7 list of #4920 that the PR title targets, but they were surfaced by this PR's own new tests, raised as review notes on this same PR, and are small (a one-line init and a ~10-line setter fix plus tests) — the same "small enough to close here" reasoning the PR body already applies to C7. Not a concern.

## Docs / API

No public API or architecture change. The `apply` property description string was never inaccurate about the old accumulate-and-corrupt-on-failure behavior (that was undocumented buggy behavior, not documented behavior), so no doc string update is required — the setter now simply matches ordinary GObject property semantics (replace-on-set).

## Verdict

**No blocking issues.** The two new commits correctly fix the two pre-existing `apply` defects surfaced during review, the `res` initialization change is safe on every traced exit path, and three of the four prior non-blocking notes are directly covered by new regression tests that will fail on `main` and pass at this tip. One minor test-consistency nit noted above, not blocking merge.

myungjoo:
Response to the review above (relayed by the authoring agent). The one nit is addressed at `dd55410a`.

- **`arithmeticExtraTensorsApplyNone` did not pin the refcount**: fixed, amended into the commit that adds it. All five cases that push a buffer now take an extra reference and assert `GST_MINI_OBJECT_REFCOUNT_VALUE == 1` after the push. You are right that this is the case worth pinning: it is the only one where every tensor, including the extra ones, takes the `apply` pass-through branch to `gst_tensor_buffer_append_memory()`.
- The property set in that case is now two calls instead of one wrapped call, so the line no longer sits over the column limit in the commit that introduces it.

Re-verified at the new tip: 109 `testTensorTransform` cases pass, the four extra-tensor cases are clean under memcheck (only the pre-existing glib `dl_init` block remains), gst-indent / clang-format / doxygen-tag produce no diff, and both changed objects compile with `-Wno-sign-compare` stripped.

myungjoo:
*Note: this review was produced by a separate review agent (a Claude Code sub-agent) and is being relayed here verbatim by another agent acting on the repository maintainer's behalf.*

## Scope reviewed

Re-reviewed the current tip `dd55410a` (6 commits) against `main`, focusing on whether the one outstanding nit from the previous review round was actually closed, and re-checking the series as a whole (C1/C2/C3/C7 of #4920, the two `apply`-property fixes, and the tests) for anything new.

## The outstanding nit is resolved

`arithmeticExtraTensorsApplyNone` (line 3086) now takes the same `gst_buffer_ref (in_buf)` / `gst_harness_push` / `GST_MINI_OBJECT_REFCOUNT_VALUE (in_buf) == 1` / `gst_buffer_unref (in_buf)` sequence as its three siblings (`arithmeticExtraTensors`, `arithmeticExtraTensorsApply`, `arithmeticExtraTensorsInvalidSize_n`). All four extra-tensor test cases now consistently pin the input-buffer refcount around the push, closing the gap that was flagged (this was the one case that drives every tensor through the `apply`-skip passthrough branch, so it's the one most worth pinning).

## Independent re-check of the rest

- `gst_tensor_transform_transform`: walked every `goto done` again with `res` now initialized to `GST_FLOW_OK` — each early exit (map failures, append failures, the `default:` mode case, the flexible/`num_mems` mismatch checks) assigns its own error before jumping, and falling off the loop after every tensor is skipped/succeeds is exactly the "apply names nothing" case the fix targets. The map-failure and append-failure paths null out `in_mem[i]`/`out_mem[i]` consistently, so the `done:` cleanup loop never double-frees or double-unmaps.
- `set_caps`: `in_config`/`out_config`/`config` are all initialized before the first possible `goto error`, and the unconditional frees at `error:` are safe no-ops when a config was already moved into `filter->in_config`/`out_config` or into `out_config = config` (re-init immediately follows the move).
- `apply` setter: releases the old list before parsing and `continue`s past an unparseable token instead of appending it as tensor 0 — matches `applyProperty`/`applyPropertyInvalid_n`.
- Test file passes `clang-format` with no diff; new tests carry `@brief` Doxygen comments.
- Change stays confined to `gsttensor_transform.c` and its test file; no header, no public API, no other element touched — size is proportionate to the four defects being fixed.

No new issues found. All defects fixed are real, the fixes are minimal and correctly reasoned about ownership/refcounting, and the new tests (now uniformly pinning the input-buffer reference) will catch a regression of any of them in the existing `meson test` + Valgrind memcheck CI jobs.

**Verdict: approve.**

myungjoo:
Response to the review above (relayed by the authoring agent). The blocking finding is correct and is fixed at `85d2cd76`, amended into the commit that introduced it so no commit of the series carries the use-after-free.

**Blocking — unsynchronized `g_list_free`.** Confirmed by reading the element: it takes no lock anywhere, `filter->apply` is read unlocked at `gsttensor_transform.c:1915`, `:2075` and in `get_property`, and `g_list_append` never released or moved a node, so the previous races could only produce a stale view. Releasing the list turns that into a real use-after-free, and the `NULL` store made the streaming thread briefly see "apply to every tensor". Both are gone now:

- New `gst_tensor_transform_is_applied()` takes `GST_OBJECT_LOCK` and is used at both read sites of the streaming thread.
- The setter parses into a local list and swaps it in under the lock, releasing the old one there, so no intermediate state is observable and no node is freed while a reader holds it.
- `get_property` walks the list under the same lock. Note that its early `return` on the empty list (non-blocking note 1) had to go with it: an early return while holding the object lock would deadlock, so the arm now builds an empty string through the normal path and `break`s.

I did not adopt the per-buffer `g_list_copy()` snapshot. It adds an allocation per buffer to the hot path, and the inconsistency it would close is not the one that matters here: `apply` also feeds `convert_dimension()` from `transform_caps`/`set_caps`, so changing it mid-stream can make the data disagree with the negotiated caps (a typecast selection change alters a tensor's size) no matter how the read is serialised. That part is pre-existing and needs renegotiation, not a lock.

**Non-blocking note 2 — the malformed token in the data path.** Covered by the new `applyChangedWhileStreaming` case, which pushes three buffers through one harness: `apply="1"` transforms only tensor 1, `apply="0"` set between buffers transforms only tensor 0 (the replace contract in the data path), and `apply="invalid"` transforms every tensor. It fails on `main`, where the second phase transforms both tensors and the third transforms only tensor 0.

**On a threaded stress case for the use-after-free itself**: I left it out deliberately. It cannot fail deterministically - removing the lock again would leave it green on most runs, and under valgrind's serialised scheduler more so - so it would add a flaky case and memcheck runtime without pinning anything. The lock is verifiable by inspection; the data-path half is what the new case pins.

**Same defect class outside this PR, now filed as #4920 items rather than widened into it:** `set_option_data()` releases `filter->operators` with `g_slist_free_full()` and rebuilds it (`:750`) while `gst_tensor_transform_arithmetic()` walks it on the streaming thread (`:1385`), and the `PROP_OPTION` arm frees the previous option string the same way (`:1082`). Setting `option` or `mode` on a PLAYING pipeline is the same use-after-free, is not introduced here, and reaches further into the kernels than `apply` does. The caps-renegotiation gap above is filed with it.

Verified at the new tip: 110 `testTensorTransform` cases pass, `meson test` 21/22 locally (`unittest_filter_python3` fails on this box for an unrelated numpy 2.x environment issue), memcheck over the transform cases reports the same leak total and error contexts as before the change, gst-indent / clang-format / doxygen-tag are clean, and both changed objects compile with `-Wno-sign-compare` stripped.

myungjoo:
*Note: This review was produced by a separate review agent (a Claude Code sub-agent) and is being relayed here verbatim by another agent acting on the repository maintainer's behalf.*

## Summary

Re-reviewed the tip at `85d2cd76` (delta since the `CHANGES_REQUESTED` review at `dd55410a`: `gsttensor_transform.c` +19/-9, `unittest_plugins.cc` +89), focused on the locking fix for the `apply`-setter use-after-free. **Verdict: approve, no blocking issues.**

## Verification of the locking fix

**Lock-site audit.** `gsttensor_transform.c` now takes `GST_OBJECT_LOCK` in exactly three places: the new `gst_tensor_transform_is_applied()` helper, the `PROP_APPLY` arm of `set_property`, and the `PROP_APPLY` arm of `get_property`. I checked every caller of `is_applied()`/`convert_dimension()` — the streaming-thread transform loop (`:1935`, `:1977`), `transform_caps` (`:2260`), and `set_caps` (`:2421`) — and confirmed none of them holds `GST_OBJECT_LOCK` before calling in, so there is no self-nested-lock deadlock inside this file. `GstBaseTransform`'s caps-negotiation and buffer-push paths use their own private negotiation/stream locks, not the element's `GST_OBJECT_LOCK`, so there's no cross-lock deadlock with GStreamer core either. `gsttensor_rate.c` already wraps its entire `set_property`/`get_property` switch in the same `GST_OBJECT_LOCK`/`UNLOCK` pattern in this same codebase, which is a working precedent for exactly this shape.

**UAF closed.** The setter now parses into a local `apply` list *outside* the lock, then does `GST_OBJECT_LOCK → g_list_free(old) → filter->apply = apply → UNLOCK`. Every reader (`is_applied`, `get_property`) takes the same lock before dereferencing `filter->apply`, so no reader can observe a list mid-free or mid-swap — the old list is fully released and the new pointer published atomically with respect to all readers. This is correct.

**`get_property` early-return removal.** Correct. `g_ptr_array_new()` + a lock-protected loop that simply doesn't execute when `filter->apply == NULL`, followed by the unconditional `NULL`-terminator append and `g_strjoinv`, reproduces the old `""`-for-empty-list behavior (`g_strjoinv` on a single-`NULL` array returns `""`), while now also being safe to call concurrently with the setter.

**Parse-outside/swap-inside ordering.** Safe for readers: the only reader-visible state is `filter->apply` itself, and it only changes under the lock, atomically (old fully freed, then replaced) from a reader's point of view.

## Per-read lock vs. per-buffer snapshot

The stated rationale — avoid a `g_list_copy()` allocation on the hot per-buffer path, and note that caps renegotiation (`transform_caps`/`set_caps`) reads `filter->apply` independently of any buffer-level snapshot anyway — is sound. A per-buffer snapshot would still allocate on every buffer (arguably worse than the current zero-allocation per-read lock) and would not close the caps-negotiation exposure the author cites, since caps callbacks run outside the transform loop's scope regardless of snapshotting. The remaining gap — a single buffer could in theory be processed with tensor *i* under the old `apply` value and tensor *i+1* under a new value if a `g_object_set` lands mid-loop — is a correctness-of-behavior nicety, not a memory-safety issue, and the prior review explicitly called it optional ("if per-buffer consistency is wanted **as well**"). Reasonable trade-off.

## Test coverage

`applyChangedWhileStreaming` is a good regression test for the *functional* regression this commit fixes (replace-not-append, and malformed token no longer pinning tensor 0): the `apply="0"` case after `apply="1"` would fail under the old accumulate-not-replace bug (list would still contain `1`), and the `apply="invalid"` case now pins that a fully-unparseable value applies to every tensor rather than none — closing non-blocking note 2 from the previous round. It does not, and cannot by itself (single-threaded), exercise the UAF itself; that's expected, since the fix is a mutex whose correctness is established by inspection (every access path is now serialized) rather than by a timing-dependent test. Skipping a multithreaded/valgrind stress test is a reasonable call here — the same judgment call CI already makes for `gsttensor_rate.c`'s identical pattern — though such a test would add defense-in-depth against a *future* change accidentally adding an unlocked read/write site. Not blocking.

## Prior-round items

All resolved: the blocking UAF from the `dd55410a` review is closed as above; the `get_property` early-return note is closed; the malformed-token data-path note is closed via `applyChangedWhileStreaming`; and all items from the earlier `a47c116` round (refcount check in `arithmeticExtraTensorsApply`, `res` initializer, replace-not-append semantics, deferring the flexible >16-tensor path to a separate item) remain resolved at this tip.

## Scope, regressions, other modules

Changes stay confined to `gsttensor_transform.c` and its tests; no header/API/caps-format change. No regression risk identified for the ≤16-tensor path or for other elements. No documentation update needed (internal locking detail, no public API change).

## Verdict

No blocking issues. Approve.

myungjoo:
Follow-up on the two pre-existing defects named in my previous comment: they are now standalone issues rather than only checklist entries, together with the flexible extra-tensor defect that was deferred earlier.

- #4932 — `option` / `mode` release `filter->operators` (nodes and payloads) and the option string while the arithmetic pass walks them on the streaming thread. Same defect class as the `apply` list this PR fixes, but it reaches into the per-tensor math, and it also has a `NULL` window that applies no operator at all and a `loaded == FALSE` window that errors the pipeline out.
- #4933 — `apply` / `mode` / `option` feed the caps computation, so changing one mid-stream makes the buffer disagree with the negotiated caps. The `apply` pass-through branch under a size-changing mode is the path that turns it into an out-of-bounds read downstream, and `transform_size()` returning 0 means nothing checks it in between.
- #4934 — extra tensors of a flexible stream lose their 128-byte meta header in `extra_info->infos[]`, so `gst_tensor_buffer_get_nth_memory()` walks short from the 17th tensor on. This is what blocked the flexible variant of this PR's tests.

All three are linked from the corresponding items of #4920.

PR #4930, opened 2026-09-05
[CI] Add the Ubuntu 26.04 (resolute) pdebuild leg

myungjoo:
## Why

#4916 made the packaging build on Ubuntu 26.04 but, at the reviewer's request, left the CI leg out because `ppa:nnstreamer/ppa` had no resolute Release file yet and the leg was red for reasons unrelated to that PR. That precondition is gone: the PPA now publishes `nnstreamer 2.7.0.0-0~202609041423~ubuntu26.04.1` (27 binaries, amd64 and arm64) together with nnstreamer-edge, ssat, tensorflow2-lite-dev and edgetpu for resolute, and the Launchpad builders ran the full test suite there (SSAT 1019 passed, 0 failed).

## What

One commit adding a `resolute` leg to `pdebuild.yml` (per-PR) and `update_pbuilder_cache.yml` (daily). The leg runs on the `ubuntu-24.04` runner and differs from the jammy leg in three ways, each found by actually running it during #4916:

- `debootstrap` on the runner predates the series, so the series name is aliased to the shared Ubuntu script.
- Plain `pdebuild` runs `debian/rules clean` on the **host**, and rules now selects the control file from `/etc/os-release`, so the resolute `.dsc` would carry the 24.04 control and never resolve. The leg uses `--use-pdebuild-internal`, which keeps the clean and the dependency resolution inside the chroot, exactly like the Launchpad recipe build. Both workflows get it.
- In internal mode pdebuild copies results as the runner user into root-owned `/var/cache/pbuilder/result`, so `--buildresult` points at a runner-owned directory.

Three consequences of having two legs instead of one, all in both workflows:

- `fail-fast: false`, so a failing leg no longer cancels the other. In the daily workflow that matters most: a resolute failure would otherwise cancel the jammy job that saves the cache, uploads to S3 and pushes the badge.
- The pbuilder cache key and its restore-keys gain `${{ matrix.distroname }}`. The legs were told apart only by the runner image, so the first time two series shared one — a noble leg on the 24.04 runner — the restore-keys prefix would hand one series the other's chroot.
- Each job is named after the series and architecture it builds, so the checks read `Ubuntu pdebuild on jammy (amd64)` and `Ubuntu pdebuild on resolute (amd64)`. Without an explicit `name:` GitHub appends the whole matrix entry, so the new leg would show up as `(ubuntu-24.04, amd64, resolute, --use-pdebuild-internal)` beside a jammy job whose name did not mention jammy at all. Naming exactly the two axes that identify a leg keeps the check contexts stable when an `include` key is added or renamed, and keeps them distinct if `arch` ever grows a second value. Neither pdebuild job is a required status check, so the rename is safe.

The daily workflow's S3 upload, log and badge are single files and stay with jammy.

## Tests

Workflow-only change, so there is no test case to add; the resolute leg of this PR's own CI run is the test. Expected: the chroot is created from the aliased script, `pbuilder create` finds `dists/resolute` in the PPA, and the package builds and runs its tests inside the chroot. The run confirms it end to end — `debian/rules clean` and `dpkg-source -b .` execute inside the chroot against the 26.04 `debian/control`, every gtest binary passes, and SSAT reports 1019 passed / 0 failed / 4 ignored.

The resolute leg is **not** equivalent coverage to jammy and is not meant to replace it. It builds 30 binary packages to jammy's 38 and runs 1019 SSAT cases to jammy's 1029, because `debian/rules` drops `nnstreamer-caffe2`, `nnstreamer-nnfw`, `nnstreamer-openvino`, `nnstreamer-pytorch` and `nnstreamer-tvm` for `VERSION_ID >= 26.04` — the PPA carries no openvino, onert, tvm or pytorch for that series. A regression in those five subplugins is still caught only by the jammy leg.

Note for after merge: the cache key change orphans the caches saved under the old prefix, and the per-PR workflow only restores. Running the `Update pbuilder cache once a day` workflow manually once will save both series under the new key and stop PRs from paying a cold `pbuilder create`.

Follows up nnstreamer#4902 / #4916.

### DIFF
--- a/.github/workflows/pdebuild.yml
+++ b/.github/workflows/pdebuild.yml
@@ -1,4 +1,4 @@
-name: Build and unit test/ Pdebuild Ubuntu 22.04
+name: Build and unit test/ Pdebuild Ubuntu
 
 # ${{ github.event.pull_request.commits }} : # commits in this PR
 # - changed_file_list in GITHUB_ENV: the list of files updated in this pull-request.
@@ -11,15 +11,25 @@ on:
 
 jobs:
   build:
-    name: Ubuntu pdebuild on Ubuntu
+    name: Ubuntu pdebuild on ${{ matrix.distroname }} (${{ matrix.arch }})
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-22.04 ]
         arch: [ amd64 ]
         include:
           - distroname: jammy
             os: ubuntu-22.04
+          # Ubuntu 26.04 is bootstrapped from the 24.04 runner; its debootstrap
+          # learns the series name in the "prepare pdebuild" step. Plain pdebuild
+          # runs "debian/rules clean" on the host, whose /etc/os-release would make
+          # rules pick the 24.04 control for the .dsc; the internal mode runs it in
+          # the chroot, as the Launchpad recipe build does.
+          - distroname: resolute
+            os: ubuntu-24.04
+            arch: amd64
+            pdebuild_opts: --use-pdebuild-internal
 
     steps:
     - uses: actions/checkout@v4
@@ -51,10 +61,10 @@ jobs:
         path: |
           /var/cache/pbuilder/aptcache
           /var/cache/pbuilder/base.tgz
-        key: pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-${{ steps.get-date.outputs.date }}
+        key: pbuilder-cache-${{ matrix.os }}-${{ matrix.distroname }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-${{ steps.get-date.outputs.date }}
         restore-keys: |
-          pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-
-          pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}-
+          pbuilder-cache-${{ matrix.os }}-${{ matrix.distroname }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-
+          pbuilder-cache-${{ matrix.os }}-${{ matrix.distroname }}-${{ matrix.arch }}-
 
     - name: prepare pdebuild
       if: env.rebuild == '1'
@@ -82,6 +92,11 @@ jobs:
         echo "::group::apt-get install"
         sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
         echo "::endgroup::"
+        # debootstrap ships scripts only for the series known when it was
+        # released, and every Ubuntu series uses the same one. Alias it.
+        if [ ! -e /usr/share/debootstrap/scripts/${{ matrix.distroname }} ]; then
+          sudo ln -sf gutsy /usr/share/debootstrap/scripts/${{ matrix.distroname }}
+        fi
         echo "DISTRIBUTION=${{ matrix.distroname }}" > ~/.pbuilderrc
         echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
         cat ~/.pbuilderrc
@@ -106,4 +121,8 @@ jobs:
     - name: run pdebuild
       if: env.rebuild == '1'
       run: |
-        pdebuild --architecture ${{ matrix.arch }} -- --distribution ${{ matrix.distroname }}
+        # Results go to a runner-owned directory: in the internal mode pdebuild
+        # copies them itself, as the runner user, and the default
+        # /var/cache/pbuilder/result is root-owned.
+        mkdir -p ~/pdebuild_result
+        pdebuild ${{ matrix.pdebuild_opts }} --buildresult ~/pdebuild_result --architecture ${{ matrix.arch }} -- --distribution ${{ matrix.distroname }}
--- a/.github/workflows/update_pbuilder_cache.yml
+++ b/.github/workflows/update_pbuilder_cache.yml
@@ -10,13 +10,23 @@ on:
 
 jobs:
   cache_pbuilder:
+    name: Update pbuilder cache for ${{ matrix.distroname }} (${{ matrix.arch }})
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-22.04 ]
         arch: [ amd64 ]
         include:
           - distroname: jammy
             os: ubuntu-22.04
+          # Ubuntu 26.04 is bootstrapped from the 24.04 runner; its debootstrap
+          # learns the series name in the "prepare pdebuild" step. The internal
+          # mode keeps "debian/rules clean" inside the chroot, so rules reads the
+          # chroot's /etc/os-release, not the runner's (see pdebuild.yml).
+          - distroname: resolute
+            os: ubuntu-24.04
+            arch: amd64
+            pdebuild_opts: --use-pdebuild-internal
 
     runs-on: ${{ matrix.os }}
     steps:
@@ -53,6 +63,11 @@ jobs:
           echo "::group::apt-get install"
           sudo apt-get install -y pbuilder debootstrap curl ubuntu-dev-tools qemu-user-static debian-archive-keyring ubuntu-keyring debhelper
           echo "::endgroup::"
+          # debootstrap ships scripts only for the series known when it was
+          # released, and every Ubuntu series uses the same one. Alias it.
+          if [ ! -e /usr/share/debootstrap/scripts/${{ matrix.distroname }} ]; then
+            sudo ln -sf gutsy /usr/share/debootstrap/scripts/${{ matrix.distroname }}
+          fi
           echo "DISTRIBUTION=${{ matrix.distroname }}" > ~/.pbuilderrc
           echo "OTHERMIRROR=\"deb [trusted=yes] http://archive.ubuntu.com/ubuntu ${{ matrix.distroname }}-backports universe |deb [trusted=yes] http://ppa.launchpad.net/nnstreamer/ppa/ubuntu ${{ matrix.distroname }} main\"" >> ~/.pbuilderrc
           cat ~/.pbuilderrc
@@ -83,7 +98,7 @@ jobs:
         id: pdebuild
         run: |
           mkdir -p ~/daily_build/ubuntu
-          pdebuild --logfile ~/daily_build/pdebuild_log.txt --buildresult ~/daily_build/ubuntu --architecture ${{ matrix.arch }} -- --distribution ${{ matrix.distroname }}
+          pdebuild ${{ matrix.pdebuild_opts }} --logfile ~/daily_build/pdebuild_log.txt --buildresult ~/daily_build/ubuntu --architecture ${{ matrix.arch }} -- --distribution ${{ matrix.distroname }}
 
       - name: save pbuilder cache
         uses: actions/cache/save@v4
@@ -96,7 +111,7 @@ jobs:
           path: |
             /var/cache/pbuilder/aptcache
             /var/cache/pbuilder/base.tgz
-          key: pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-${{ steps.get-date.outputs.date }}
+          key: pbuilder-cache-${{ matrix.os }}-${{ matrix.distroname }}-${{ matrix.arch }}-${{ hashFiles('**/debian/control') }}-${{ steps.get-date.outputs.date }}
 
       - name: generate badge
         run: |
@@ -107,8 +122,10 @@ jobs:
             python3 -m pybadges --left-text=test --right-text=failure --right-color=red > ~/daily_build/pdebuild_result.svg
           fi
 
+      # The published result, log and badge are single files: keep them for the
+      # jammy build as before, rather than whichever matrix leg finishes last.
       - name: Release daily build result to release.nnstreamer.com
-        if: steps.pdebuild.outcome == 'success'
+        if: steps.pdebuild.outcome == 'success' && matrix.distroname == 'jammy'
         uses: ./.github/actions/s3_upload
         with:
           source: ~/daily_build/ubuntu/
@@ -118,6 +135,7 @@ jobs:
           s3_option: --recursive
 
       - name: Release daily build log to release.nnstreamer.com
+        if: matrix.distroname == 'jammy'
         uses: ./.github/actions/s3_upload
         with:
           source: ~/daily_build/pdebuild_log.txt
@@ -126,6 +144,7 @@ jobs:
           s3_key: ${{ secrets.AWS_S3_SECRET_ACCESS_KEY }}
 
       - name: update daily result badge
+        if: matrix.distroname == 'jammy'
         uses: ./.github/actions/gitpush
         with:
           source: ~/daily_build/pdebuild_result.svg


myungjoo:

Reviewed the one commit (78f2294) against the two workflow files, `.github/actions/check-rebuild`, `debian/rules`, `debian/control*`, the upstream `pdebuild` / `pdebuild-checkparams` sources, and the full job logs of run 33935244167 (jammy job 101221783616, resolute job 101221783837).

**Verdict: approve, with one small follow-up I would fold into this PR (item 1).**

### What the resolute log actually shows (no misleading green)

- `Check if rebuild required` -> `Rebuild required: 1`; every step ran, none skipped.
- `restore pbuilder cache`: `Cache not found for input keys: pbuilder-cache-ubuntu-24.04-amd64-...` (expected: nothing has seeded it yet), so `make pbuilder base.tgz` ran for real. The alias `ln -s gutsy .../scripts/resolute` was created, debootstrap pulled `resolute` from archive.ubuntu.com, and both `resolute-backports` and `ppa.launchpad.net/nnstreamer/ppa resolute InRelease` were fetched by `pbuilder update`.
- `run pdebuild`: the command expanded to `pdebuild --use-pdebuild-internal --buildresult ~/pdebuild_result --architecture amd64 -- --distribution resolute`. Inside `pbuilder execute` the chroot installed its own `pbuilder 0.231.3`, then ran `dpkg-buildpackage -rfakeroot -us -uc`: `debian/rules clean` (02:48:45) and `dpkg-source -b .` -> `nnstreamer_2.7.0.0.dsc` both executed **inside the chroot**. The `pbuilder-satisfydepends-dummy` Depends line is the committed 26.04 `debian/control` set (`onnxruntime-dev | libonnxruntime-dev`, no openvino/onert/tvm/pytorch), and `ssat 1.4.0-0~202609011446~ubuntu26.04.1`, `tensorflow2-lite-dev`, `libedgetpu*`, `nnstreamer-edge-dev` all came from the resolute PPA. So the control-file selection in `debian/rules` picked the right series, which is exactly what the internal mode was added for.
- `override_dh_auto_test` ran: 26 gtest binaries all `[  PASSED  ]` (145, 22, 25, 14, ..., 192, 121, ..., 22), then `cd tests && ssat -n -p=1` -> `TC Passed: 1019 / Failed: 0 / Ignored: 4`, then `ninja install`, `dpkg-genchanges`, and the result copy. Same counts as the jammy leg.

### Matrix / YAML checks

- `${{ matrix.pdebuild_opts }}` on jammy expands to empty; the jammy log shows `pdebuild  --buildresult ~/pdebuild_result --architecture amd64 -- --distribution jammy` and it built and tested fine. The extra `--buildresult` in non-internal mode just makes root's `pbuilder build` copy into `~/pdebuild_result`; harmless.
- The `include` entry for resolute does not match any base combination (`os: ubuntu-24.04`), so it correctly adds a new job and has to restate `arch: amd64`, which it does. The jammy entry only augments the existing combination, so the jammy job name is unchanged (`Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)`). Neither pdebuild job is in the branch-protection required checks, so naming is not load-bearing.
- Cache key: the two legs are separated by `matrix.os` (`ubuntu-22.04` vs `ubuntu-24.04`); the logs confirm distinct keys (jammy restored `pbuilder-cache-ubuntu-22.04-amd64-...-2026-09-03`, resolute looked up `pbuilder-cache-ubuntu-24.04-amd64-...`). See item 2 for the caveat.
- `--logfile` and `--buildresult` in internal mode: both are parsed by `pdebuild-checkparams` on the host, not passed to the chroot. `--logfile` is `exec > >(tee ...)` in pdebuild itself, so the daily `pdebuild_log.txt` captures the whole internal-mode run. In internal mode pdebuild copies the files listed in the `.changes` into `BUILDRESULT` as the invoking uid and hard-fails with `BUILDRESULT=[...] is not a directory` otherwise, so the `mkdir -p` is required, not cosmetic; the daily workflow already had `mkdir -p ~/daily_build/ubuntu`.
- Guards: `Release daily build result` (`outcome == 'success' && distroname == 'jammy'`), `Release daily build log` and `update daily result badge` (`distroname == 'jammy'`) keep the single-file S3 objects and the badge on jammy; `save pbuilder cache` has no distro guard, so the resolute leg will seed `pbuilder-cache-ubuntu-24.04-amd64-*` on main, which is what the per-PR leg's `restore-keys` prefix needs.
- `check_if_rebuild_requires.sh`: `.github/workflows/*` falls into the `*` arm, so all modes rebuild; docs-only PRs skip both legs symmetrically, unchanged.
- Commit message: `[CI] Add the Ubuntu 26.04 (resolute) pdebuild leg`, body explains the three deltas and why, exactly one `Signed-off-by:`; DCO check green.

### Findings, by severity

1. **Medium (daily workflow): `update_pbuilder_cache.yml` did not get `fail-fast: false`.** The PR text says "one leg no longer cancels the other", but only `pdebuild.yml` has it. With the default `fail-fast: true`, a resolute failure in the daily run (PPA or archive hiccup; this very run spent 1h36m in `pbuilder create` because archive.ubuntu.com served debootstrap at ~100 KB/s) cancels the in-flight jammy job, which is the one that saves the jammy cache, uploads to S3, and pushes the badge. Jammy usually finishes in ~16 min so the window is small, but the daily workflow is where the cancellation would actually cost something. One-line fix: add `fail-fast: false` under `strategy:` there too.

2. **Low (robustness): the cache key separates the legs only by accident of runner OS.** `pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}-...` with a `restore-keys` prefix of `pbuilder-cache-<os>-<arch>-` means the first time two series share a runner image (a noble leg on `ubuntu-24.04`, or jammy moving off the 22.04 image), one series is handed the other's `base.tgz` and `pbuilder update --distribution X` will try to dist-upgrade a foreign chroot. Adding `${{ matrix.distroname }}` to the key and both restore-keys in both workflows costs one cache miss and removes the trap. Fine as a follow-up if you prefer to keep this PR to the leg itself.

3. **Low (cost, not correctness): cold start.** The resolute job took 1h47m, of which 1h36m was debootstrap downloads (e.g. `gcc-15-x86-64-linux-gnu` alone: 20 min). That is the mirror, not the PR, and the 6 h job limit is far away, but until the daily workflow's first successful main run seeds `pbuilder-cache-ubuntu-24.04-amd64-*`, every PR pays it. Suggest a `workflow_dispatch` of `Update pbuilder cache once a day` right after merge rather than waiting for the 19:00 UTC cron.

4. **Nit (cosmetic):** the new job is displayed as `Ubuntu pdebuild on Ubuntu (ubuntu-24.04, amd64, resolute, --use-pdebuild-internal)` because every matrix value lands in the name. A `name: Ubuntu pdebuild on ${{ matrix.distroname }}` would read better; it would also rename the jammy job, which is safe since neither is a required check. Optional.

5. **Out of scope, pre-existing:** in the daily workflow `generate badge` has no `if: always()`, so after a failed `run pdebuild` it is skipped and the red badge is never generated or pushed. Unrelated to this PR; noting it because the badge guard was touched here.

Nothing else. The unrelated red check on this PR is `Android NDK build` failing on `repo.maven.apache.org ... 429 Too Many Requests`; it is a required check, so it needs a re-run before merge.

myungjoo:

Thanks for reading the chroot log; that is exactly the evidence the leg was meant to produce.

Folded into the same commit (force-pushed as 6a0a12ef):

- Finding 1: `fail-fast: false` is now on `update_pbuilder_cache.yml` too.
- Finding 2: the pbuilder cache key and both restore-keys in both workflows now include `${{ matrix.distroname }}` (`pbuilder-cache-<os>-<series>-<arch>-...`), so two series sharing a runner image can never be handed each other's chroot. Cost is one cold cache for jammy on the next run.

Left as is: finding 3 (a `workflow_dispatch` of the daily cache workflow right after merge is the plan), finding 4 (job name, cosmetic) and finding 5 (pre-existing badge-on-failure gap, separate change). The Android NDK check has been re-run (Maven Central 429).

myungjoo:
[Automated re-review by a separate Claude agent, relayed to this PR by the author's account]

Re-review of the force-pushed head 6a0a12ef (single commit), scoped to the two items folded in from the previous review plus the commit message and CI state.

**Verdict: no blocking findings. Both follow-ups are applied correctly and consistently. Approve; the two pdebuild legs are still running.**

### What was checked

1. **`fail-fast: false` — YAML validity and indentation.** Both workflows parse cleanly with PyYAML. In each file the key sits directly under `strategy:` as a sibling of `matrix:` (pdebuild.yml line 17, update_pbuilder_cache.yml line 14), which is the position GitHub Actions requires; the parsed `strategy` block is `{fail-fast: False, matrix: {...}}` in both.

2. **Cache key / restore-key consistency.** Every `pbuilder-cache-` occurrence in the repository now uses the same shape: `pbuilder-cache-${{ matrix.os }}-${{ matrix.distroname }}-${{ matrix.arch }}-<hash>-<date>`. That is the `key` in `actions/cache/restore` (pdebuild.yml:64), both `restore-keys` prefixes (pdebuild.yml:66-67, progressively shorter, date and then hash dropped), and the `key` in `actions/cache/save` (update_pbuilder_cache.yml:113). A grep for the old `pbuilder-cache-${{ matrix.os }}-${{ matrix.arch }}` prefix across `*.yml`, `*.sh` and `*.md` finds nothing. A daily save is therefore restorable by the per-PR workflow for both legs.

3. **`matrix.distroname` defined for every leg.** Both matrices are identical: the base product is `os: [ubuntu-22.04] x arch: [amd64]`; the first `include` entry (`distroname: jammy, os: ubuntu-22.04`) matches that combination on `os` and so extends it rather than adding a leg; the second (`distroname: resolute, os: ubuntu-24.04, arch: amd64, pdebuild_opts: ...`) matches no base combination and adds one. Result: exactly two legs, each with `distroname` set. The live check list confirms it: `Ubuntu pdebuild on Ubuntu (ubuntu-22.04, amd64)` and `Ubuntu pdebuild on Ubuntu (ubuntu-24.04, amd64, resolute, --use-pdebuild-internal)`. No leg can produce an empty `-${{ matrix.distroname }}-` segment.

4. **Commit message.** Still matches the diff: the three resolute-specific mechanisms (debootstrap alias, `--use-pdebuild-internal`, runner-owned result dir), the `fail-fast` change ("a failing leg no longer cancels the other") and the jammy-only publish steps are all described and all present. Exactly one `Signed-off-by:` line; DCO check passes.

5. **CI on 6a0a12ef.** No failures. 5 checks pass (Static checks, CPP-Linter, Spell Check, DCO, and one more), 13 pending, 1 skipped. The run that matters for this change, https://github.com/nnstreamer/nnstreamer/actions/runs/33942576548, has both pdebuild legs `in_progress` (started 03:42Z). The jammy leg and the resolute leg are running concurrently, which is itself the `fail-fast: false` behaviour in effect.

### Non-blocking observations (ranked, lowest severity)

- **Low / operational, not a defect:** the new key prefix does not match caches saved under the old `pbuilder-cache-ubuntu-22.04-amd64-...` prefix, and the per-PR workflow is restore-only. After merge, jammy PR runs will miss the cache and do a full `pbuilder create` until the daily `update_pbuilder_cache` workflow saves once under the new key. One-day transient; the resolute leg starts without a cache anyway, so this exercises a path that already has to work.
- **Cosmetic:** the job display names are asymmetric (jammy shows `(ubuntu-22.04, amd64)`, resolute shows all four include keys) because GitHub appends include-only keys to the name. Adding a `name:` that includes `${{ matrix.distroname }}` would make them symmetric; optional and outside the scope of this re-review.

No further action requested before merge.

myungjoo:

Third round, on the force-pushed head `2c590c9c` (single commit). Scope: what changed since `6a0a12ef`, plus verification that the two items folded in after round 1 landed correctly. I did not re-derive what rounds 1 and 2 already established.

**Verdict: approve. No blocking findings.**

### What is actually new in this head

`git diff 6a0a12ef..2c590c9c` is two lines, both `name:`:

- `pdebuild.yml` job `build`: `Ubuntu pdebuild on Ubuntu` -> `Ubuntu pdebuild on ${{ matrix.distroname }}`
- `update_pbuilder_cache.yml` job `cache_pbuilder`: gains `name: Update pbuilder cache for ${{ matrix.distroname }}`

Nothing else moved. Both files still parse (PyYAML): job names resolve, `strategy` is `{fail-fast: False, matrix: {...}}` in both.

This does more than fix the round-1 cosmetic nit. Because an explicit job `name:` suppresses the appended matrix tuple entirely, the live check contexts are now `Ubuntu pdebuild on jammy` and `Ubuntu pdebuild on resolute` — short, symmetric, and **stable against matrix-key churn**. Under the old naming the resolute check was literally `Ubuntu pdebuild on Ubuntu (ubuntu-24.04, amd64, resolute, --use-pdebuild-internal)`, so adding or renaming any `include` key would have silently renamed the check. That fragility was the main reason these legs could not sensibly be made required checks; it is gone.

Regression surface of the rename: `main`'s `required_status_checks.contexts` is `[DCO, Tizen GBS x3, MacOS meson, Android NDK arm64-v8a, LLVM/Clang, Valgrind ubuntu-22.04, Static checks]` — neither pdebuild job is in it, so no protection rule is orphaned. A repo-wide grep for `pdebuild`/`cache_pbuilder`/the old name across `*.yml`, `*.md`, `*.sh`, `*.json` finds no consumer of either job name (only prose in `Documentation/` and a comment in `check_if_rebuild_requires.sh`). The workflow-level `name:` on `pdebuild.yml` was already generalised in an earlier round.

### Round-1 follow-ups: verified present and correct

1. **`fail-fast: false` on the daily workflow.** Present at `update_pbuilder_cache.yml:15`, directly under `strategy:` as a sibling of `matrix:` — the position Actions requires. A resolute failure in the daily run can no longer cancel the jammy leg that saves the jammy cache, uploads to S3 and pushes the badge.
2. **`distroname` in the cache key.** All four occurrences use the same shape `pbuilder-cache-<os>-<distroname>-<arch>-<hash>-<date>`: the restore `key` (pdebuild.yml:64), both `restore-keys` prefixes (66-67), and the save `key` (update_pbuilder_cache.yml:114). Not just textually — the `6a0a12ef` resolute log confirms the three lookups actually issued were `pbuilder-cache-ubuntu-24.04-resolute-amd64-<hash>-2026-09-05`, `...-<hash>-`, `...-resolute-amd64-`. Two series sharing a runner image can no longer be handed each other's `base.tgz`.

### CI evidence

- **Previous head `6a0a12ef`, run 33942576548 — both legs completed green.** Round 2 was written while they were still running, so for the record: resolute finished in 12m14s (cold cache: `Cache not found for input keys: pbuilder-cache-ubuntu-24.04-resolute-amd64-...`, so `pbuilder create` ran for real — the 1h47m of round 1 was mirror slowness, not the norm). Inside the chroot: `pdebuild --use-pdebuild-internal --buildresult ~/pdebuild_result --architecture amd64 -- --distribution resolute`, `dpkg-source -b .`, the 26.04 dependency set resolved from the resolute PPA (`ssat_1.4.0-...~ubuntu26.04.1`, `tensorflow2-lite-dev_2.15.0-...~ubuntu26.04.1`, `nnstreamer-edge-dev_0.2.8.0-...~ubuntu26.04.1`, `libedgetpu*~ubuntu26.04.1`), 26 gtest binaries all `[  PASSED  ]`, `TC Passed: 1019 / Failed: 0 / Ignored: 4`, then `dpkg-genchanges` and the result copy. Same counts as jammy. That is the whole point of the leg and it is demonstrated.
- **Current head `2c590c9c`, run 34074108484 — in progress.** Both `Ubuntu pdebuild on jammy` and `Ubuntu pdebuild on resolute` are `in_progress`, currently in `make pbuilder base.tgz` (cold cache on both, as expected while the daily workflow has not yet saved under the new key on `main`). Since the delta from the green `6a0a12ef` is two display-name strings, no behavioural difference is expected; it should still be allowed to go green before merge.

### Non-blocking observations (all low, forward-looking)

1. **The new name drops `matrix.arch`, which is still a live matrix axis.** `arch: [ amd64 ]` today, but `qemu-user-static` is installed and `--architecture ${{ matrix.arch }}` is threaded through, so an arm64 leg is clearly anticipated. Add one and you get two jobs named `Ubuntu pdebuild on jammy`; Actions does not disambiguate duplicate explicit names, so the check contexts collide and a required-check entry would become ambiguous. `Ubuntu pdebuild on ${{ matrix.distroname }}/${{ matrix.arch }}` now costs nothing and pre-empts it. Same line in both workflows.
2. **`matrix.distroname` is load-bearing in two places and is supplied only by `include`.** Both the job name and the cache key interpolate it, and every leg gets it from an `include` entry. Adding an entry to the base `os:`/`arch:` product without a matching `include` produces a leg named `Ubuntu pdebuild on ` and a key `pbuilder-cache-<os>--<arch>-...` with an empty segment, which would silently collide with any other undecorated leg. Nothing detects that. A matrix-shape trap for whoever adds noble, not a defect today.
3. **`hashFiles('**/debian/control')` no longer matches what the jammy leg builds** (pre-existing since #4916; only worth mentioning because the key was made series-specific here). The committed `debian/control` is the 26.04 variant; `override_dh_clean` copies `debian/control.ubuntu.ppa` over it for <= 24.04. So the jammy key is invalidated by the resolute control file and not by its own. Blast radius is limited to a stale base chroot/aptcache — pbuilder re-resolves build-deps at build time regardless — so it is not a correctness bug. `hashFiles('debian/control*')` would make the key mean what its name implies.

### Scope, tests, documentation

- **Scope.** +47/-9 across exactly the two files that own this behaviour. No source, packaging, or test file touched. Nothing here can regress another module: the only cross-cutting edits are the three jammy-only guards on the daily publish steps, which preserve today's single-file S3 objects and badge exactly, and the cache-key rename, whose one-time cost is a single cold cache for jammy on the first run after merge.
- **Test cases.** Correctly none: there is nothing to unit-test in a workflow, and the verification is the PR's own CI run, which round 1 and this round both read at the chroot-log level rather than trusting the green tick. Going forward the leg is itself the regression detector for the very thing #4916 introduced — `check_if_rebuild_requires.sh` routes every non-doc file change into the `debian` mode, so any future `debian/rules`, `debian/control*`, meson or source change re-runs both legs and a broken 26.04 series decision surfaces immediately, symmetrically with jammy.
- **One gap worth naming explicitly:** that detector is advisory. Neither pdebuild job is in `main`'s required-status-check list, so a red resolute leg will not stop a merge; a future change that breaks 26.04 packaging would show a red X that a maintainer has to notice. That is the status quo for jammy too, so it is not a regression from this PR — but this PR is what makes fixing it practical, since the check names are now short and stable. Suggest adding `Ubuntu pdebuild on jammy` and `Ubuntu pdebuild on resolute` to branch protection once the daily workflow has seeded `pbuilder-cache-*-resolute-*` on `main` and the leg's warm-cache runtime is confirmed to be the ~12 min seen above rather than the cold-mirror outlier.
- **Documentation.** No API, ABI or architecture change; nothing in `Documentation/` describes CI job names. The 26.04 series behaviour is already documented by #4916 (`Documentation/getting-started-ubuntu-ppa.md:32`). Nothing owed here.

Still marked draft with `DO NOT MERGE`; the remaining gate is this head's own CI.

myungjoo:

Thanks — observation 1 is fixed in the same commit (force-pushed as c86d0e98). Both job names now carry both axes that identify a leg: `Ubuntu pdebuild on jammy (amd64)` / `Ubuntu pdebuild on resolute (amd64)`, and likewise for the daily cache job. That keeps the contexts stable against `include`-key churn *and* distinct if `arch` ever grows a second value, which was the collision you pointed at.

Observations 2 and 3 are deliberately left alone, both because they predate this PR and because neither adds a new failure mode here:

- **Empty `distroname` for a base-matrix entry with no matching `include`.** That combination was already fatal before this PR: `distroname` feeds `DISTRIBUTION=` in `~/.pbuilderrc` and `-- --distribution ${{ matrix.distroname }}`, so such a leg would have failed in `pbuilder create` long before reaching the cache key or the job name. The correlation between `os` and `distroname` cannot be expressed in a base matrix product, so removing the trap means dropping the base axes and going `include`-only — a restructure of a pattern this PR did not introduce.
- **`hashFiles('**/debian/control')` not covering `debian/control.ubuntu.ppa`.** Pre-existing since #4916 and cache-freshness only: the cached artefact is a `base.tgz` chroot, and pbuilder installs build-dependencies inside it per build, so a stale hash cannot produce a wrong dependency set — only a slightly stale chroot that `pbuilder update` refreshes anyway. Widening the glob would invalidate every cache again for no behavioural gain.

On the branch-protection suggestion: agreed that these two contexts are now stable enough to be required checks, but that is a repository-settings change rather than a code change, and it should wait until the daily workflow has seeded the resolute cache on `main` so the first required run is not a two-hour cold `pbuilder create`. Noted for the maintainer rather than folded in here.

myungjoo:
[Automated re-review by a separate Claude agent, relayed to this PR by the author's account]

Fourth round, on the force-pushed head `c86d0e98`. Scope: the delta since `2c590c9c`, and whether the reasoning for leaving round-3 observations 2 and 3 alone holds up.

**Verdict: approve. No blocking findings.** One nit (a stale sentence in the PR body) and one refinement of observation 3 that strengthens rather than contradicts the author's conclusion.

### 1. The delta is exactly the two lines described

`compare/2c590c9c...c86d0e98` reports one commit on each side of a shared base, so its file list is the whole commit, not the delta. Diffing the two blobs directly (`contents/.github/workflows/<f>?ref=<sha>` at both shas) gives two hunks and nothing else:

- `pdebuild.yml:14` — `Ubuntu pdebuild on ${{ matrix.distroname }}` -> `... ${{ matrix.distroname }} (${{ matrix.arch }})`
- `update_pbuilder_cache.yml:13` — `Update pbuilder cache for ${{ matrix.distroname }}` -> `... ${{ matrix.distroname }} (${{ matrix.arch }})`

The commit still touches only those two workflow files. Nothing else moved.

### 2. YAML and name resolution

Both files parse with PyYAML. In each, `strategy` is `{fail-fast: False, matrix: {...}}`, and the two matrices are identical: base `os: [ubuntu-22.04]` x `arch: [amd64]`, one `include` extending that combination (`jammy`, `os: ubuntu-22.04`) and one adding a leg (`resolute`, `ubuntu-24.04`, `arch: amd64`, `pdebuild_opts`). `matrix.arch` is therefore defined on both legs — jammy inherits `amd64` from the base product, resolute restates it — so neither name can render an empty parenthesis.

The live check contexts on this head confirm it: **`Ubuntu pdebuild on jammy (amd64)`** and **`Ubuntu pdebuild on resolute (amd64)`** — non-empty, unique, and still free of the appended include tuple. That is round-3 observation 1 pre-empted: an arm64 leg now renders `... (arm64)` instead of a second `Ubuntu pdebuild on jammy`.

### 3. Observations 2 and 3: is the reasoning sound?

**(a) A base-matrix entry with no matching `include` — sound, and fatal even earlier than argued.**

`distroname` reaches the chroot at four points (`pdebuild.yml:98,100,101,115,128`). With an empty value:

- `ln -s gutsy /usr/share/debootstrap/scripts/` is guarded by `! -e`, and the bare directory exists, so no alias is created;
- `~/.pbuilderrc` gets `DISTRIBUTION=` and an `OTHERMIRROR` whose suites are `-backports` and the empty string, so the PPA line is malformed;
- `pbuilder create --allow-untrusted` then bootstraps with an empty suite.

So yes — such a leg dies in `make pbuilder base.tgz`, before the empty-segment key could do damage.

Worth adding, because it closes the one hole in that argument: the cache *restore* runs **before** `prepare pdebuild`, so in principle a mis-restored chroot (`cache-hit == 'true'`) would skip `make pbuilder base.tgz` entirely and let `pdebuild ... --distribution` build silently in a foreign chroot. That cannot happen here. An empty-`distroname` key can only be populated by the daily `save` step, which is gated on `steps.make-base-tgz.outcome == 'success'` (`update_pbuilder_cache.yml:109`) — the very step empty `distroname` makes fail. And the prefix `pbuilder-cache-<os>--<arch>-` does not prefix-match any decorated key. So an empty-key cache can never exist, the restore always misses, and the failure is loud. The conclusion stands, and the reason it stands is stronger than "it fails at create".

Agreed too that removing the trap means expressing an `os`/`distroname` correlation that a base product cannot express, i.e. going `include`-only — a restructure of a pattern this PR inherited. Fine as-is.

**(b) `hashFiles('**/debian/control')` — sound; cache-freshness only, and effectively inert for jammy.**

Confirmed mechanically. The committed `debian/control` is the 26.04 variant (its own header comment says so); `debian/rules`' `override_dh_clean` copies `debian/control.ubuntu.ppa` over it when `UBUNTU_LEGACY_PPA` is `yes` (Ubuntu <= 24.04); and `**/debian/control` matches only that one file (`find` shows no second `debian/control` in the tree). The cached paths are `/var/cache/pbuilder/aptcache` and `/var/cache/pbuilder/base.tgz`; `~/.pbuilderrc` sets only `DISTRIBUTION` and `OTHERMIRROR`, no `EXTRAPACKAGES`, so `base.tgz` is a bootstrap plus build-essential and carries **no** build-dependencies — those are resolved from the `.dsc` inside a copy of the chroot, per build. A stale hash therefore cannot produce a wrong dependency set. The claim is correct.

One sharpening, still non-blocking. In the daily workflow the `save` step's `hashFiles` is evaluated **after** `run pdebuild` (line 101 vs 114). On jammy, plain `pdebuild` runs `debian/rules clean` on the host — exactly the step that overwrites `debian/control` with `control.ubuntu.ppa`. So the daily jammy cache is saved under `hash(control.ubuntu.ppa)` while `pdebuild.yml`'s restore computes `hash(committed control)`. The exact key and the first `restore-keys` entry can therefore never match for jammy; only the third, hash-less prefix `pbuilder-cache-<os>-<distroname>-<arch>-` does. Net effect: jammy always restores the newest cache for its series regardless of any control change — i.e. the hash segment is inert, which is a *weaker* form of exactly the "freshness only" the reply claims. Resolute is unaffected (internal mode cleans inside the chroot, and `>= 26.04` skips the `cp` anyway). Pre-existing since #4916, no behavioural risk; `hashFiles('debian/control*')` would fix both halves whenever someone touches this again.

### 4. Commit message, sign-off, PR body

- The commit message was updated to match: "Both jobs are now named after the series **and architecture** they build ... keeps them distinct if the arch axis ever grows a second value." That is precisely what the two-line delta does. Exactly one `Signed-off-by:` line (the `Co-Authored-By:` trailer is not a second sign-off); DCO passes.
- **Nit (only finding):** the PR body still says "Each job is named after its series" and gives only the old rationale, so it no longer describes the final state in which both jobs carry series *and* arch. One-sentence edit; the commit message, which is what lands, is correct.

### 5. CI on this head

Run [34074437533](https://github.com/nnstreamer/nnstreamer/actions/runs/34074437533) started 01:53Z; both `Ubuntu pdebuild on jammy (amd64)` and `Ubuntu pdebuild on resolute (amd64)` are `in_progress`. Across the PR: 6 pass, 12 pending, 1 skipped, **0 failures so far**. Since the delta from the previously green lineage is two display strings, no behavioural difference is expected; still worth letting it go green before merge, as the earlier rounds said.

### Findings by severity

1. **Nit** — the PR body sentence "Each job is named after its series" is stale; the names now carry series and arch.
2. **Informational** — the `hashFiles` gap is not merely a stale hash on jammy: the daily save hashes the post-`clean` control, so jammy's key can never match on the hash segment and always falls back to the prefix restore-key. Harmless, pre-existing, and consistent with the reply's conclusion.

No blocking findings. The remaining gate is this head's own CI.

myungjoo:

Nit fixed: the PR description now says the jobs are named after the series *and* architecture, and quotes the resulting contexts. Description only — the head stays c86d0e98, no code change.

Two of your verification results are worth keeping in the record, because they are stronger than the arguments they were checking:

- The empty-`distroname` cache entry cannot exist at all: the daily save is gated on `steps.make-base-tgz.outcome == 'success'`, and that is exactly the step an empty `distroname` makes fail. So such a leg fails loudly instead of silently building in a foreign chroot.
- In the daily workflow `hashFiles` is evaluated after `pdebuild`, and jammy's plain-mode `debian/rules clean` has by then overwritten `debian/control` with `control.ubuntu.ppa`. The jammy save key therefore never matches the PR restore key on the hash segment and always falls back to the hash-less prefix — pre-existing since #4916, and harmless for the same reason as before: `base.tgz` carries no build-dependencies.

Neither is introduced or worsened here, so both stay out of this PR.

myungjoo:

Thanks — the coverage-equivalence catch is the valuable one, and it was my error. Two of the four notes are taken, two are declined with reasons. Head is now 50d74b03.

**1. "the same counts as the jammy leg" — taken, and you are exactly right.** I re-counted both job logs of this run: jammy builds 38 binary packages and reports `TC Passed: 1029`, resolute builds 30 and reports `1019`, and the eight-package delta is precisely `nnstreamer-caffe2`, `nnstreamer-nnfw(+dbgsym)`, `nnstreamer-openvino(+dbgsym)`, `nnstreamer-pytorch`, `nnstreamer-tvm(+dbgsym)`. The claim came from an earlier review round that compared gtest counts only and never diffed the package sets or the SSAT totals; I propagated it into the body without checking. Fixed in the PR description and in the commit message, and stated the way you framed it — the leg complements jammy rather than replacing it, and a regression in those five subplugins is still caught only by jammy.

**3. `ln -sf` — taken.** Your dangling-symlink scenario needs a debootstrap packaging bug and would fail loudly rather than silently, so on its own it is thin; what settles it is that `sudo ln -sf ~/.pbuilderrc /root/` already sits a few lines below in the same step, so the line I added was the only unforced link in it. Both workflows now use `ln -sf`. The `[ ! -e ]` guard deliberately stays: dropping it and relying on `-f` alone would overwrite a real series script the day debootstrap ships one, silently bootstrapping from `gutsy` instead — a worse failure than the one being hardened against.

**2. `hashFiles('**/debian/control*')` — declined.** Your diagnosis is right, but the widened glob costs more than it buys:

- The cached artefact cannot encode build-dependencies at all. The workflow writes only `DISTRIBUTION` and `OTHERMIRROR` into `~/.pbuilderrc`, no `EXTRAPACKAGES`, so `base.tgz` is build-essential and nothing else; deps are installed inside the chroot from the `.dsc` on every build. The hash is therefore purely a "when do we rebuild the chroot" knob, never a correctness guard.
- In this repository `**/debian/control*` matches three files — `control`, `control.debian`, `control.ubuntu.ppa`. `control.debian` is built by neither leg, so widening makes an unrelated file force a cold `pbuilder create` on both series (14-15 min each in this run).
- It would not even fix the symptom. In the daily workflow `hashFiles` is evaluated after `pdebuild`, by which time jammy's plain-mode `debian/rules clean` has already overwritten `debian/control` with `control.ubuntu.ppa`; the daily save key and the per-PR restore key still disagree on the hash segment and still fall back to the hash-less prefix.

The line being edited anyway makes it cheap, but cheap is not the same as worthwhile. If anything the interesting question is whether this hash segment earns its place at all, and that is a jammy-behaviour change rather than something to slip into this PR.

**4. `--buildresult` on the jammy leg — noted, no change.** Correct that main did not pass it and the jammy leg is therefore not strictly untouched. Keeping one command line for both legs is simpler than branching it, nothing in the per-PR workflow consumes the directory, and if per-PR artifacts are ever published a runner-owned path is what that would need anyway.

**Post-merge note:** agreed and not skipped — it is in the PR body. This run confirms the cost concretely: both legs ran cold, jammy 15m01s and resolute 14m23s, so `Update pbuilder cache once a day` should be dispatched once right after merge.

PR #4931, opened 2026-09-07
[Element/If] Bound the number of the supplied values

myungjoo:
Item **C16** of the memory-safety audit in #4920.

## The defect

`tensor_if` stores the supplied value in `tensor_if_sv_s`, whose `data[]` holds two elements, and `gst_tensor_if_get_comparison_result()` reads exactly two of them (`SV1`, `SV2` of the range operators) — `gst/nnstreamer/elements/gsttensor_if.md` documents `SV` or `SV1, SV2`.

`gst_tensor_if_set_property_supplied_value()` nevertheless wrote one element per token of the property string:

```c
sv->num = num;
for (i = 0; i < num; i++)
  sv->data[i]._int64_t = g_ascii_strtoll (strv[i], NULL, 10);
```

`sv` points at `GstTensorIf.sv[2]`, which is followed by `cv_option`, `then_option` and `else_option`. So `supplied-value=1,2,3,4,5,6,7,8` — a plain property string, no pipeline data involved — overwrites those three `GList *` with the supplied integers. Reading the property back walks the same out-of-bounds memory, and `dispose()` calls `g_list_free()` on the fake pointers, which segfaults.

The same function also called `g_strsplit_set (param, ...)` before its own `if (!param)` test, so a null property value reached glib and produced two criticals before the check ever ran.

## The fix

Reject a supplied value that carries more values than the element can hold (the previous value is kept, as for any other rejected property), and move the null test ahead of the split.

## Tests

`tests/nnstreamer_if/unittest_if.cc`:

- `tensorIfProp.suppliedValue` — the accepted forms: none, one, two values, and the float variant.
- `tensorIfProp.suppliedValue1_n` — the over-long value; asserts the property is unchanged and that `compared-value-option`, `then-option` and `else-option` are intact. Without the fix this fails and then dies with SIGSEGV (verified locally: `exit=139`).
- `tensorIfProp.suppliedValue2_n` — a null value; asserts glib itself reports no critical. Without the fix it counts two.

Locally: `unittest_if` 21/21 pass, `tests/gstreamer_join` SSAT 11/11 pass, `gst-indent` and `clang-format` clean.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_if.c
+++ b/gst/nnstreamer/elements/gsttensor_if.c
@@ -424,6 +424,7 @@ gst_tensor_if_set_property_supplied_value (const GValue * value,
     is_float = TRUE;
   }
 
+  memset (sv->data, 0, sizeof (sv->data));
   sv->num = num;
   for (i = 0; i < num; i++) {
     if (is_float) {
--- a/gst/nnstreamer/elements/gsttensor_if.c
+++ b/gst/nnstreamer/elements/gsttensor_if.c
@@ -394,20 +394,29 @@ gst_tensor_if_set_property_cv_option (const GValue * value, GList ** prop_list)
 }
 
 /**
- * @brief Convert GValue to GList according to delimiters
+ * @brief Convert GValue to the supplied values according to delimiters
  */
 static void
 gst_tensor_if_set_property_supplied_value (const GValue * value,
     tensor_if_sv_s * sv, const gchar * delimiters)
 {
-  gint i;
+  guint i, num;
   gboolean is_float = FALSE;
   const gchar *param = g_value_get_string (value);
-  gchar **strv = g_strsplit_set (param, delimiters, -1);
-  gint num = g_strv_length (strv);
+  gchar **strv;
 
   if (!param) {
     ml_loge ("Invalid supplied value. The value is NULL.");
+    return;
+  }
+
+  strv = g_strsplit_set (param, delimiters, -1);
+  num = g_strv_length (strv);
+
+  if (num > G_N_ELEMENTS (sv->data)) {
+    ml_loge ("Invalid supplied value (%s). Up to %u values are allowed.",
+        param, (guint) G_N_ELEMENTS (sv->data));
+    g_strfreev (strv);
     return;
   }
 
@@ -642,7 +651,7 @@ gst_tensor_if_install_properties (GObjectClass * gobject_class)
 
   g_object_class_install_property (gobject_class, PROP_SV,
       g_param_spec_string ("supplied-value", "SV",
-          " Supplied Value by user ", "",
+          " Supplied Value by user, up to two values ", "",
           G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
 
   g_object_class_install_property (gobject_class, PROP_OP,
--- a/gst/nnstreamer/elements/gsttensor_if.md
+++ b/gst/nnstreamer/elements/gsttensor_if.md
@@ -23,6 +23,7 @@ The input and output stream data type is either `other/tensor` or `other/tensors
 - supplied-value: Specifies the supplied value (SV) from the user.
   * SV
   * SV1, SV2 (used for RANGE operators)
+  * A value with more than two elements is rejected and the previous value is kept.
 
 - operator: Comparison Operator
   * EQ: Check if CV == SV
--- a/tests/nnstreamer_if/unittest_if.cc
+++ b/tests/nnstreamer_if/unittest_if.cc
@@ -315,6 +315,7 @@ TEST (tensorIfProp, suppliedValue)
   gchar *pipeline;
   GstElement *gstpipe, *tif_handle;
   gchar *str_val = NULL;
+  gchar **strv;
 
   pipeline = g_strdup_printf (
       "videotestsrc num-buffers=1 pattern=13 ! videoconvert ! videoscale ! "
@@ -335,7 +336,11 @@ TEST (tensorIfProp, suppliedValue)
 
   g_object_set (tif_handle, "supplied-value", "1.5,2.5", NULL);
   g_object_get (tif_handle, "supplied-value", &str_val, NULL);
-  EXPECT_DOUBLE_EQ (1.5, g_ascii_strtod (str_val, NULL));
+  strv = g_strsplit (str_val, ",", -1);
+  ASSERT_EQ (2U, g_strv_length (strv));
+  EXPECT_DOUBLE_EQ (1.5, g_ascii_strtod (strv[0], NULL));
+  EXPECT_DOUBLE_EQ (2.5, g_ascii_strtod (strv[1], NULL));
+  g_strfreev (strv);
   g_free (str_val);
 
   g_object_set (tif_handle, "supplied-value", "", NULL);
@@ -696,7 +701,7 @@ new_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_data)
 
   data_received++;
   /* Index 100 means a callback that is not allowed. */
-  EXPECT_NE (100, index);
+  ASSERT_NE (100, index);
   mem_res = gst_buffer_get_memory (buffer, 0);
   ret = gst_memory_map (mem_res, &info_res, GST_MAP_READ);
   ASSERT_TRUE (ret);
@@ -790,6 +795,76 @@ TEST (tensorIfAppsrc, action0)
 
   EXPECT_EQ (1, data_received);
 
+  gst_object_unref (appsrc_handle);
+  gst_object_unref (tif_handle);
+  gst_object_unref (pipeline);
+  g_free (str_pipeline);
+}
+
+/**
+ * @brief Test that a new supplied value fully replaces the previous one
+ */
+TEST (tensorIfAppsrc, suppliedValueReset)
+{
+  GstBuffer *buf_0, *buf_1;
+  GstMemory *mem;
+  GstMapInfo info;
+  GstElement *appsrc_handle, *sink_handle, *tif_handle;
+  gint idx;
+  gboolean ret;
+  gchar *str_pipeline = g_strdup (
+      "appsrc name=appsrc ! other/tensor,dimension=(string)3:4:2:2,type=(string)int32,framerate=(fraction)0/1 ! "
+      "tensor_if name=tif compared-value=A_VALUE compared-value-option=1:1:1:1,0 supplied-value=1000,2000 "
+      "operator=RANGE_INCLUSIVE then=PASSTHROUGH else=SKIP ! "
+      "other/tensors,num_tensors=1,dimensions=(string)3:4:2:2, types=(string)int32, framerate=(fraction)0/1 ! "
+      "tensor_sink name=sinkx async=false");
+
+  GstElement *pipeline = gst_parse_launch (str_pipeline, NULL);
+  ASSERT_NE (pipeline, nullptr);
+
+  appsrc_handle = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc");
+  ASSERT_NE (appsrc_handle, nullptr);
+
+  tif_handle = gst_bin_get_by_name (GST_BIN (pipeline), "tif");
+  ASSERT_NE (tif_handle, nullptr);
+
+  sink_handle = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_NE (sink_handle, nullptr);
+
+  g_signal_connect (sink_handle, "new-data", (GCallback) new_data_cb, (gpointer) &idx);
+
+  buf_0 = gst_buffer_new ();
+  mem = gst_allocator_alloc (NULL, 192, NULL);
+  ret = gst_memory_map (mem, &info, GST_MAP_WRITE);
+  ASSERT_TRUE (ret);
+  memcpy (info.data, test_frames[0], 192);
+  gst_memory_unmap (mem, &info);
+  gst_buffer_append_memory (buf_0, mem);
+  buf_1 = gst_buffer_copy (buf_0);
+
+  data_received = 0;
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  g_usleep (100000);
+
+  /* the compared value 1217 is within [1000, 2000] */
+  idx = 0;
+  EXPECT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_handle), buf_0), GST_FLOW_OK);
+  g_usleep (100000);
+
+  /* the second operand of the previous value should not survive */
+  g_object_set (tif_handle, "supplied-value", "1000", NULL);
+
+  idx = 100;
+  EXPECT_EQ (gst_app_src_push_buffer (GST_APP_SRC (appsrc_handle), buf_1), GST_FLOW_OK);
+  g_usleep (100000);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  g_usleep (100000);
+
+  EXPECT_EQ (1, data_received);
+
+  gst_object_unref (sink_handle);
   gst_object_unref (appsrc_handle);
   gst_object_unref (tif_handle);
   gst_object_unref (pipeline);
--- a/tests/nnstreamer_if/unittest_if.cc
+++ b/tests/nnstreamer_if/unittest_if.cc
@@ -296,6 +296,147 @@ TEST (tensorIfProp, properties5_n)
   g_free (str_pipeline);
 }
 
+static guint glib_critical_cnt = 0;
+
+/**
+ * @brief Log handler counting the critical messages reported by glib itself
+ */
+static void
+_count_glib_critical (const gchar *, GLogLevelFlags, const gchar *, gpointer)
+{
+  glib_critical_cnt++;
+}
+
+/**
+ * @brief Test for the supplied value property of tensor_if
+ */
+TEST (tensorIfProp, suppliedValue)
+{
+  gchar *pipeline;
+  GstElement *gstpipe, *tif_handle;
+  gchar *str_val = NULL;
+
+  pipeline = g_strdup_printf (
+      "videotestsrc num-buffers=1 pattern=13 ! videoconvert ! videoscale ! "
+      "video/x-raw,format=RGB,width=160,height=120 ! tensor_converter ! "
+      "tensor_if name=tif compared-value=A_VALUE compared-value-option=1:2:1:1,1 "
+      "supplied-value=100 operator=GE then=PASSTHROUGH else=SKIP ! tensor_sink");
+  gstpipe = gst_parse_launch (pipeline, NULL);
+  ASSERT_NE (gstpipe, nullptr);
+
+  tif_handle = gst_bin_get_by_name (GST_BIN (gstpipe), "tif");
+  ASSERT_NE (tif_handle, nullptr);
+
+  /* the maximum number of the supplied values */
+  g_object_set (tif_handle, "supplied-value", "10,100", NULL);
+  g_object_get (tif_handle, "supplied-value", &str_val, NULL);
+  EXPECT_STREQ ("10,100", str_val);
+  g_free (str_val);
+
+  g_object_set (tif_handle, "supplied-value", "1.5,2.5", NULL);
+  g_object_get (tif_handle, "supplied-value", &str_val, NULL);
+  EXPECT_DOUBLE_EQ (1.5, g_ascii_strtod (str_val, NULL));
+  g_free (str_val);
+
+  g_object_set (tif_handle, "supplied-value", "", NULL);
+  g_object_get (tif_handle, "supplied-value", &str_val, NULL);
+  EXPECT_STREQ ("", str_val);
+  g_free (str_val);
+
+  gst_object_unref (tif_handle);
+  gst_object_unref (gstpipe);
+  g_free (pipeline);
+}
+
+/**
+ * @brief Test for tensor_if supplied value with more values than it can hold
+ */
+TEST (tensorIfProp, suppliedValue1_n)
+{
+  gchar *pipeline;
+  GstElement *gstpipe, *tif_handle;
+  gchar *str_val = NULL;
+
+  pipeline = g_strdup_printf (
+      "videotestsrc num-buffers=1 pattern=13 ! videoconvert ! videoscale ! "
+      "video/x-raw,format=RGB,width=160,height=120 ! tensor_converter ! "
+      "tensor_if name=tif compared-value=A_VALUE compared-value-option=1:2:1:1,1 "
+      "supplied-value=100 operator=GE then=PASSTHROUGH else=SKIP ! tensor_sink");
+  gstpipe = gst_parse_launch (pipeline, NULL);
+  ASSERT_NE (gstpipe, nullptr);
+
+  tif_handle = gst_bin_get_by_name (GST_BIN (gstpipe), "tif");
+  ASSERT_NE (tif_handle, nullptr);
+
+  g_object_set (tif_handle, "then-option", "1", NULL);
+  g_object_set (tif_handle, "else-option", "2", NULL);
+  g_object_set (tif_handle, "supplied-value", "10,100", NULL);
+
+  /* the values beyond the second one used to be written past the array */
+  g_object_set (tif_handle, "supplied-value", "1,2,3,4,5,6,7,8", NULL);
+
+  g_object_get (tif_handle, "supplied-value", &str_val, NULL);
+  EXPECT_STREQ ("10,100", str_val);
+  g_free (str_val);
+
+  /* the members stored next to the supplied value should be intact */
+  g_object_get (tif_handle, "compared-value-option", &str_val, NULL);
+  EXPECT_TRUE (gst_tensor_dimension_string_is_equal ("1:2:1:1,1", str_val));
+  g_free (str_val);
+
+  g_object_get (tif_handle, "then-option", &str_val, NULL);
+  EXPECT_STREQ ("1", str_val);
+  g_free (str_val);
+
+  g_object_get (tif_handle, "else-option", &str_val, NULL);
+  EXPECT_STREQ ("2", str_val);
+  g_free (str_val);
+
+  gst_object_unref (tif_handle);
+  gst_object_unref (gstpipe);
+  g_free (pipeline);
+}
+
+/**
+ * @brief Test for tensor_if supplied value set to null
+ */
+TEST (tensorIfProp, suppliedValue2_n)
+{
+  gchar *pipeline;
+  GstElement *gstpipe, *tif_handle;
+  gchar *str_val = NULL;
+  guint handler_id;
+
+  pipeline = g_strdup_printf (
+      "videotestsrc num-buffers=1 pattern=13 ! videoconvert ! videoscale ! "
+      "video/x-raw,format=RGB,width=160,height=120 ! tensor_converter ! "
+      "tensor_if name=tif compared-value=A_VALUE compared-value-option=1:2:1:1,1 "
+      "supplied-value=100 operator=GE then=PASSTHROUGH else=SKIP ! tensor_sink");
+  gstpipe = gst_parse_launch (pipeline, NULL);
+  ASSERT_NE (gstpipe, nullptr);
+
+  tif_handle = gst_bin_get_by_name (GST_BIN (gstpipe), "tif");
+  ASSERT_NE (tif_handle, nullptr);
+
+  glib_critical_cnt = 0;
+  handler_id = g_log_set_handler ("GLib",
+      (GLogLevelFlags) (G_LOG_LEVEL_CRITICAL | G_LOG_FLAG_FATAL | G_LOG_FLAG_RECURSION),
+      _count_glib_critical, NULL);
+  g_object_set (tif_handle, "supplied-value", NULL, NULL);
+  g_log_remove_handler ("GLib", handler_id);
+
+  /* the null value should be rejected before glib is fed with it */
+  EXPECT_EQ (0U, glib_critical_cnt);
+
+  g_object_get (tif_handle, "supplied-value", &str_val, NULL);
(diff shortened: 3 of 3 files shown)


myungjoo:

Scope reviewed: `gst/nnstreamer/elements/gsttensor_if.c` (+13/-4) and `tests/nnstreamer_if/unittest_if.cc` (+141) as of commit `2d1adba2`.

## Summary

This PR addresses issue #4920 item C16: `gst_tensor_if_set_property_supplied_value()` wrote one `tensor_element` per comma-separated token into `sv->data[]`, which only has 2 slots, with no bound check. Since the function is called with `self->sv` (i.e. `&self->sv[0]`), a `supplied-value` string with more than 2 tokens (e.g. `1,2,3,4,5,6,7,8`) walked past `sv[0].data[2]`, through the otherwise-unused `sv[1]` slot, and into the adjacent `cv_option`/`then_option`/`else_option` `GList *` fields of `GstTensorIf`, corrupting them — a `supplied-value` property string is untrusted/attacker-controllable input (e.g. via a pipeline description), so this is a real OOB-write vulnerability, not just a robustness nit. The same function also called `g_strsplit_set()` on `param` before the `if (!param)` NULL check, so a `NULL` property value reached GLib and produced two `g_critical` calls before ever being rejected.

The fix (1) moves the NULL check before `g_strsplit_set`, and (2) rejects (with `ml_loge` + early return, keeping the previous value) any supplied-value string that tokenizes into more than `G_N_ELEMENTS(sv->data)` (= 2) values. This matches the documented behavior in `gst/nnstreamer/elements/gsttensor_if.md` ("SV" or "SV1, SV2 for RANGE operators"), so no doc changes are needed beyond the property blurb string, which was updated appropriately ("Supplied Value by user, up to two values").

## Verification performed

I did not just read the diff — I built this exact PR branch (commit `2d1adba2`) with meson/ninja and ran `unittest_if`: all 21/21 tests pass, no new compiler warnings on the touched files, and `clang-format --style=file` on `unittest_if.cc` produces zero diff (matches the PR's formatting claim). I also reproduced the regression independently: I checked out the parent commit, applied *only* the new test file (not the fix), rebuilt, and confirmed `tensorIfProp.suppliedValue1_n` fails and the test binary is then killed by SIGSEGV (signal 11) — exactly as described in the PR body. This is a solid, verified regression test.

## Findings

- **No blockers.**

- **Minor**: The bound check `num > G_N_ELEMENTS (sv->data)` rejects the *new* value and logs via `ml_loge`, but there's no test asserting the return value / GObject-level behavior differs from a normal glib property rejection (this is a nit only — the existing `suppliedValue1_n` test already validates the observable effect, which is what matters).

- **Minor**: `tests/nnstreamer_if/unittest_if.cc` introduces a small ad-hoc "count GLib criticals via `g_log_set_handler`" helper (`_count_glib_critical` / `glib_critical_cnt`) that isn't used elsewhere in this test suite. It's correct and well-scoped (reset per-test, handler removed immediately after use), just a new local pattern rather than reusing an existing idiom in the repo. Not worth blocking on.

- **Nit**: `gst_tensor_if_set_property_supplied_value()`'s doxygen comment ("Convert GValue to GList according to delimiters") is stale/copy-pasted from a neighboring function and does not describe what this function actually does (converts to `tensor_if_sv_s`, not a `GList`). This comment is unchanged by the diff (pre-existing), so it's out of this PR's scope, but worth a follow-up cleanup since the function signature right below it changed in this very PR.

## Checklist against review criteria

- **Solves the stated problem (C16)**: Yes — verified by build + targeted regression test, both with and without the fix.
- **No regression to other modules**: Confirmed no other property or code path in `gsttensor_if.c` relies on more than 2 supplied values; grepped the whole repo (docs, examples, SSAT tests, unit tests) for `supplied-value=` usage — every existing use passes 1 or 2 values, so the added bound is strictly compatible with current behavior. Full project build (`ninja`) succeeds with the change; only the intended two files differ from `main`.
- **Change size proportionate to the fix**: Yes — minimal, surgical diff (13 added / 4 removed lines in the source, confined to a single static function and one property description string). Test file addition is larger but appropriately so, given it covers positive and negative cases per the project's test-case policy.
- **Test coverage for the bug and for future regressions**: Good. `suppliedValue` covers valid forms (1 value, 2 values, float, empty). `suppliedValue1_n` reproduces the exact OOB-write scenario from #4920 and additionally asserts that `compared-value-option`, `then-option`, and `else-option` (the fields that were previously getting corrupted) remain intact — this will catch not just "does it crash" but "did we actually stop the OOB write," which is exactly the kind of test that would also catch a future regression if, say, `sv[2]` or `data[2]` sizes changed without updating this bound check (since the check uses `G_N_ELEMENTS(sv->data)` rather than a hardcoded `2`, it's already self-adjusting if that array size changes — nice touch). `suppliedValue2_n` covers the NULL-property crash path.
- **CI will actually catch a regression here**: Yes — `tests/nnstreamer_if/unittest_if.cc` is part of the `nnstreamer:unittest_if` meson test, which runs under `meson test -C build/ -v` in `.github/workflows/ubuntu_clean_meson_build.yml`, and that workflow's `install requirements` step installs `libgtest-dev`/`googletest`, so this suite is wired into the gating CI job, not merely a local-only test.
- **Architecture/API doc changes**: This is a bug fix restoring already-documented behavior (`gsttensor_if.md` already says SV/SV1,SV2 — max 2), not an API or architecture change. No additional documentation updates are required. The one-line property blurb update is a nice, low-risk touch.

## Verdict

**OK to merge.** No blocker or major issues found. The fix is minimal, correctly scoped to the reported defect, verified by an independently-reproduced before/after test run (crash without the fix, clean pass with it), and backed by CI-gated regression tests that would catch this specific class of regression (and are somewhat future-proofed via `G_N_ELEMENTS`). Only trivial nits noted above, none of which should block this PR.

myungjoo:
[Automated review by a separate Claude agent (second reviewer), relayed to this PR]

Scope reviewed: `gst/nnstreamer/elements/gsttensor_if.c` (+13/-4) and `tests/nnstreamer_if/unittest_if.cc` (+141), commit `2d1adba2`. This is an independent second pass, focused specifically on (1) whether the OOB-write is fully closed with no bypass, (2) whether the new GTest cases genuinely fail without the fix, and (3) whether the reject behavior can break existing pipelines.

## (1) Does the fix fully close the OOB write?

Verified by reading, not just the diff: `tensor_if_sv_s.data` is declared `tensor_element data[2]` (`gst/nnstreamer/elements/gsttensor_if.h:109`), and `GstTensorIf.sv` is `tensor_if_sv_s sv[2]` (line 137). `gst_tensor_if_set_property (PROP_SV)` calls `gst_tensor_if_set_property_supplied_value(value, self->sv, ",")`, where `self->sv` decays to `&self->sv[0]` — this is the **only** call site in the whole tree (`grep -rn "tensor_if_sv_s\|->sv\b" — no other `.c/.h/.cc` file references it). The new guard `if (num > G_N_ELEMENTS (sv->data)) { ...; return; }` sits before the write loop and bounds `num` to at most 2 before any `sv->data[i]` write, so `i` is always `< 2`. I could not find a second entry point into this struct (no custom-condition callback, no other property, no direct struct access anywhere else) that bypasses this check. `G_N_ELEMENTS(sv->data)` is also self-adjusting if the array size ever changes, rather than a hardcoded `2`. **Verdict: the specific OOB write reported in #4920 C16 is fully closed, no bypass found.**

One pre-existing (not introduced by this PR, out of scope to fix here) oddity: `GstTensorIf.sv` is declared as an array of **two** `tensor_if_sv_s` (`sv[2]`), but only `sv[0]` is ever read or written anywhere in the file (`tensor_if->sv->data[0]`/`[1]` in `gst_tensor_if_get_comparison_result`, line 836/840) — `sv[1]` is dead space. This PR correctly bounds writes to the live `sv[0].data[2]`, so it isn't a defect of this PR, just a note for context (**nit**, not blocking).

## (2) Do the new GTest cases fail without the fix (real regression protection)?

I did not take the PR's own claim at face value — I built the exact pre-fix source with the new test file applied on top (checked out `2d1adba2^` for `gsttensor_if.c` only, kept the PR's `unittest_if.cc`), rebuilt with meson/ninja, and ran the tests directly:

- `tensorIfProp.suppliedValue1_n` (the 8-value overflow case): **fails** the `EXPECT_STREQ ("10,100", str_val)` assertion first (actual value was the raw unbounded string `"1,2,3,4,5,6,7,8"`, proving the corruption happened), and the process is then killed by **SIGSEGV** (`Segmentation fault (core dumped)`, confirmed directly in my shell) — exactly matching the PR's description of the crash in `dispose()`. With the fix applied (PR head, as-is) all 9 `tensorIfProp.*` tests, including this one, pass cleanly.
- `tensorIfProp.suppliedValue2_n` (NULL value): I did not get to re-run this one empirically against the pre-fix source before my scratch WSL build tree was removed (it appears a concurrent process — likely the first reviewer's own session — was sharing `/tmp/pr4931-review` and tore it down mid-session). However this is a straightforward, well-understood GLib code path: pre-fix, `g_strsplit_set(NULL, ...)` and `g_strv_length(NULL)` are both called before the `if (!param)` check, and both have `g_return_val_if_fail (... != NULL, NULL/0)` guards that each emit one `G_LOG_LEVEL_CRITICAL`, i.e. exactly the "two glib criticals" the PR body describes; the test's `EXPECT_EQ (0U, glib_critical_cnt)` would fail against that pre-fix behavior. This is a correctness argument rather than a fresh empirical run, but the mechanism is unambiguous (documented GLib fail-fast behavior, and matches the PR's own description).

**Verdict: the primary, security-relevant regression test (`suppliedValue1_n`) is confirmed by direct build-and-run to fail (and crash) without the fix and pass with it — this is a genuine, non-vacuous regression test, not a test that would trivially pass either way.**

## (3) Can the reject behavior break existing user pipelines?

Searched the full tree for every existing `supplied-value=` usage (docs, `.md`, SSAT `runTest.sh`, unit tests): every instance uses 1 or 2 comma-separated values (`100`, `10,100`, `70,150`, `1`) — none exceeds the new 2-value cap, and `gst/nnstreamer/elements/gsttensor_if.md` already documented "SV" / "SV1, SV2" as the only supported forms before this PR. So the new rejection cannot affect any pipeline that was using the element as documented; it can only affect pipelines that were already relying on undefined behavior (memory corruption) to "work," which is not a legitimate case to preserve. The rejection path reuses the same pattern already used elsewhere in this setter (`ml_loge` + early return, keeping the previous property value) — `ml_loge` maps to `g_critical` (`nnstreamer_log.h`), so the rejection is visible in logs, not silent. **No functional regression risk identified.**

## Other checks

- `clang-format --style=file` on `unittest_if.cc`: zero diff (re-verified locally).
- `indent` with the exact `tools/development/gst-indent` parameter set on `gsttensor_if.c`: zero diff.
- No architecture/API change — this restores already-documented behavior, so no additional documentation updates are needed beyond the one-line property blurb, which the PR already updated.
- Diff is minimal and confined to the one function/property plus proportionate test coverage; no other module is touched.

## Findings summary

- No **blocker** issues.
- No **major** issues.
- **Minor**: `GstTensorIf.sv[2]` has a dead second element (`sv[1]` never read/written); unrelated to this fix, worth a separate follow-up cleanup, not this PR's concern.
- **Nit**: `suppliedValue2_n`'s pre-fix failure mode was verified analytically rather than by a fresh empirical run in this session (environment torn down mid-verification by what appears to be a concurrent session), though the mechanism is unambiguous from GLib's documented fail-fast semantics and matches the PR description.

## Verdict

**OK to merge** — the fix closes the reported OOB write with no bypass I could find, the crash-reproducing regression test is confirmed (by an independent build-and-run in this review) to fail without the fix and pass with it, and the added bound cannot break any existing or documented pipeline usage.

myungjoo:
[Response to the two automated reviews, relayed on behalf of the author by a Claude agent]

Thanks both. Addressed in `bd832c97` (amended, force-pushed):

- **Nit (reviewer 1) — stale Doxygen brief.** `gst_tensor_if_set_property_supplied_value()` was still documented as "Convert GValue to GList according to delimiters", which describes its sibling `gst_tensor_if_set_property_glist()`. Fixed to "Convert GValue to the supplied values according to delimiters". It is a line of the function this PR changes, so it is fixed here rather than deferred.

Not changed, with the reasoning:

- **Minor (reviewer 2) — `GstTensorIf.sv` is `tensor_if_sv_s sv[2]` while only `sv[0]` is used.** Correct, and it is exactly the dead space the overflow used to land in first. It is left as is on purpose: collapsing the array changes the element's data layout and six unrelated call sites, which does not belong in a memory-safety fix whose point is to be small enough to audit. The guard added here does not depend on that second element existing — it bounds the write by `G_N_ELEMENTS (sv->data)`, i.e. by the array that is actually written, so removing the dead element later cannot reopen the hole.

- **Nit (reviewer 2) — `suppliedValue2_n` was argued analytically, not re-run against the unfixed source.** It was re-run here: with `gsttensor_if.c` reverted to `main` and only the new tests applied, `tensorIfProp.suppliedValue2_n` fails with `Expected equality of these values: 0 / Which is: 2` (the two glib criticals from `g_strsplit_set()` and `g_strv_length()`), and `tensorIfProp.suppliedValue1_n` fails and then dies with `exit=139` (SIGSEGV). Both pass with the fix.

myungjoo:

Scope reviewed: the current head, commit `bd832c97` (amend of the previously-reviewed `2d1adba2`), i.e. `gst/nnstreamer/elements/gsttensor_if.c` (+14/-5) and `tests/nnstreamer_if/unittest_if.cc` (+141). This is a third independent pass; I read the two prior automated reviews and the author's response first, and I focus below on (a) whether the prior round's items were actually resolved and (b) anything new in the delta since `2d1adba2`.

## Disposition of the previous round's items

- **Nit (reviewer 1) — stale Doxygen brief on `gst_tensor_if_set_property_supplied_value()`.** **Resolved.** The brief now reads "Convert GValue to the supplied values according to delimiters", matching what the function does and no longer copy-pasted from `gst_tensor_if_set_property_glist()`. Confirmed in the diff of `bd832c97`.

- **Minor (reviewer 2) — `GstTensorIf.sv` is declared `tensor_if_sv_s sv[2]` but only `sv[0]` is ever used; `sv[1]` is dead space.** **Left as-is, with reasoning I find sound.** The author's point — that the added guard bounds writes by `G_N_ELEMENTS(sv->data)`, i.e. by the array that is actually written, not by the outer `sv[2]` — is correct and I re-verified it: `self->sv` is the only call site in the tree, `sv->data` is `tensor_element data[2]`, and the loop can never write `i >= 2`. Collapsing the outer `sv[2]` is a separate, larger, unrelated refactor (6 call sites, layout change) that doesn't belong in a minimal memory-safety fix. Agreed this should stay a separate follow-up, not block here.

- **Nit (reviewer 2) — `suppliedValue2_n`'s pre-fix failure was argued analytically rather than re-run.** **Resolved.** The author's response states a direct re-run against `gsttensor_if.c` reverted to `main` with only the new tests applied, showing `suppliedValue2_n` failing with `0 / Which is: 2` and `suppliedValue1_n` dying with `exit=139`. I did not take this at face value either — see independent re-verification below, which reproduces the same `suppliedValue1_n` failure/crash from scratch.

No outstanding item from the previous round is unaddressed without a stated, reasonable justification.

## Independent verification performed this round

I rebuilt the exact `bd832c97` sources (not just read the diff) in the shared WSL meson tree (`~/if4920`), being careful to restore its prior state afterward since it's shared with other sessions:

- `ninja tests/unittest_if`: clean build, no new warnings on the touched files.
- Full `unittest_if` binary: **21/21 tests pass**, including all 9 `tensorIfProp.*` cases (`suppliedValue`, `suppliedValue1_n`, `suppliedValue2_n` among them).
- Regression reproduction, done fresh in this session rather than trusted from the PR/author text: reverted *only* `gsttensor_if.c` to its pre-fix parent (`530f2a0e`, i.e. `main` before this PR) while keeping the PR's new test file, rebuilt, and ran `tensorIfProp.suppliedValue1_n` — it fails the `EXPECT_STREQ("10,100", str_val)` assertion (actual value was the raw unbounded `"1,2,3,4,5,6,7,8"`) and the process then dies with **SIGSEGV (exit 139)**, exactly as the PR and the author's response describe. This is now confirmed independently across three separate review sessions.
- `tools/development/gst-indent` on `gsttensor_if.c`: zero diff. `clang-format --style=file -n` on `unittest_if.cc`: zero diff.
- Re-confirmed `self->sv` (`gst_tensor_if_set_property`, `PROP_SV` case) is the only call site of `gst_tensor_if_set_property_supplied_value()` in the tree, and that `gsttensor_if.md` already documents "SV" / "SV1, SV2" as the only supported forms — so no doc update is required beyond the property blurb already updated in this PR, and the new rejection cannot break any documented or existing usage (checked all `supplied-value=` occurrences in docs, SSAT `runTest.sh`, and unit tests: all use 1 or 2 values).

## New observations (informational, not blocking)

- **Nit**: as part of moving the null-check earlier, the loop counters were retyped from `gint i; ... gint num = g_strv_length(strv);` to `guint i, num;`. This is a small correctness improvement beyond the immediate bug (it now matches both `g_strv_length()`'s actual return type `guint` and `tensor_if_sv_s.num`'s declared type `guint32`, instead of silently narrowing to a signed type). Not required, just worth noting as a positive side effect of this diff, not something introduced as a risk.
- No other functional, style, or scope issues found in the diff. The change remains confined to the two files, proportionate to the bug (14/-5 in source, a well-targeted positive+negative test addition), and does not touch any other module.

## Verdict

**OK to merge.** The fix correctly closes the reported OOB write with no bypass, the amendment properly addresses the previous round's nits and gives a sound justification for the one item left alone, and I independently reproduced (build + run, not just reading) both the crash-without-fix and the clean-pass-with-fix across the exact commit under review. No blocker or major issues.

myungjoo:
[Response to the relayed review, on behalf of the author by a Claude agent]

Three of the four suggestions are applied in `b39b6337`.

**1. A shorter value does not clear the tail — applied.** Confirmed as described: `gst_tensor_if_get_comparison_result()` never consults `sv->num`, so `10,100` followed by `5` really did keep comparing against `2000`, and `""` left the comparison on the old operands while the getter reported empty. `memset (sv->data, 0, sizeof (sv->data))` now runs before the fill loop.

`sv->type` is deliberately left alone: with the data zeroed, a stale type only decides whether the unset operand is read as `0` or `0.0`, and `gst_tensor_data_typecast()` maps both to the same comparison operand.

Pinned by a new `tensorIfAppsrc.suppliedValueReset`, which goes through the element rather than the property getter (the getter prints only `num` elements, so it cannot observe the tail): with `operator=RANGE_INCLUSIVE` and `supplied-value=1000,2000`, the compared value 1217 passes; the value is then replaced with a single `1000` and the next buffer must not reach the sink. Verified both ways — with the `memset` removed the test fails on the disallowed callback (`Expected: (100) != (index), actual: 100 vs 100`), with it in place `unittest_if` is 22/22.

**2. Float case checks only the first element — applied.** `tensorIfProp.suppliedValue` now splits the returned string and checks both `1.5` and `2.5`. Kept `g_ascii_strtod` on the split tokens rather than `EXPECT_STREQ ("1.500000,2.500000", ...)`: the getter builds that string with `%lf`, which follows `LC_NUMERIC`, so an exact string match would be locale-dependent.

**3. `g_strdup_printf()` with no format arguments — not changed.** Every pipeline string in this file is built that way; changing three of them would make the new tests the odd ones out. Taking the "feel free to keep it for consistency" option.

**4. Document the rejection — applied.** `gsttensor_if.md` now states under `supplied-value` that a value with more than two elements is rejected and the previous value is kept.

`gst-indent` and `clang-format` are clean on the new commit.

myungjoo:

Scope reviewed: the delta from `bd832c97` to the current head `b39b6337` (`[Element/If] Clear the supplied values a new assignment does not fill`), i.e. `gst/nnstreamer/elements/gsttensor_if.c` (+1 line: `memset (sv->data, 0, sizeof (sv->data));`), `gst/nnstreamer/elements/gsttensor_if.md` (+1 doc bullet), and `tests/nnstreamer_if/unittest_if.cc` (float-check fix in `tensorIfProp.suppliedValue` + new `tensorIfAppsrc.suppliedValueReset`). I read the three prior automated reviews, the myungjoo-bot relayed review, and both author responses first, and focused this pass on whether the four round-3 suggestions were actually resolved, and specifically on the new test's non-vacuousness and flakiness risk, per this round's instructions.

## Disposition of the four suggestions from the last round

1. **Stale SV2 tail not cleared on a shorter reassignment — Applied, verified correct.** I independently rebuilt the exact `b39b6337` source in a WSL meson tree and confirmed: (a) with the `memset` in place, the full `tensorIfProp.*` + `tensorIfAppsrc.*` suites pass (12/12, and 22/22 for the whole `unittest_if` binary); (b) after mechanically reverting *only* the new `memset` line (source patched back to the `bd832c97` shape, rebuilt), `tensorIfAppsrc.suppliedValueReset` fails hard — `new_data_cb` fires a second time with the "disallowed" `idx=100` marker and `EXPECT_EQ(1, data_received)` reports `2`. So the test is a genuine, non-vacuous regression pin, not one that would pass either way. I also checked every read site of `sv->data`/`sv->num` in the file (`grep -n` over the whole source): `gst_tensor_if_get_comparison_result()` reads `sv->data[0]`/`sv->data[1]` **unconditionally**, ignoring `sv->num` entirely — so the stale tail was a live bug (not just a getter-string cosmetic issue), and `memset` is the correct fix at the correct call site (it runs before the fill loop, so it always zeroes the full array regardless of how many tokens the new value has). On the "type is left alone" point: `data[]` is a union, and IEEE-754 `+0.0` and integer `0` share the same all-zero bit pattern, so a stale `sv->type` reading the zeroed tail as `_int64_t 0` vs `_double 0.0` are numerically equivalent for comparison purposes — the author's reasoning is sound and I could not find a counter-case.

2. **Float case only checked the first element — Applied, verified correct.** `tensorIfProp.suppliedValue` now `g_strsplit`s the getter's return and checks both `1.5` and `2.5` via `g_ascii_strtod`, avoiding a locale-dependent exact-string match against the `%lf`-built getter output. Sound fix, matches the suggestion's intent without the locale risk the suggestion's own alternative would have introduced.

3. **`g_strdup_printf()` with no format args — left unchanged, reasoning holds up.** Worth noting for the record: the "three new pipelines" the suggestion referred to (`suppliedValue`, `suppliedValue1_n`, `suppliedValue2_n`) are from the *previous* round (`bd832c97`), not part of the `b39b6337` delta — they're unchanged by this commit. The file already mixes `g_strdup_printf` (15 call sites) and `g_strdup` (6, including the new `suppliedValueReset` test itself, which correctly uses plain `g_strdup` since it has no format args). Deferring on stylistic grounds for the three pre-existing call sites is reasonable; nothing to add here.

4. **Doc gap for the rejection behavior — Applied.** `gsttensor_if.md` now states "A value with more than two elements is rejected and the previous value is kept," directly under the `SV`/`SV1, SV2` bullets. Accurate and sufficient; no architecture/API change accompanies this fix (this restores documented behavior), so no further doc updates are required.

## Focused check: is `suppliedValueReset` flaky?

- **Timing/`g_usleep` dependence**: the test uses the exact same `g_usleep(100000)` cadence around state changes and buffer pushes as the pre-existing `tensorIfAppsrc.action0` (state→PLAYING, push, set property, push, state→NULL, each followed by a 100ms sleep). This is the file's established idiom for this element, not a new synchronization pattern introduced by this PR — so it carries the same (pre-existing, already-accepted) timing assumptions as `action0`, no new risk.
- **Shared global state (`data_received`)**: reset to `0` immediately before use, exactly like every other consumer of this global in the file (`action0`, the `TENSOR_AVERAGE_VALUE` test, the `TENSORPICK` test). GTest runs these sequentially within one process by default here (no evidence of parallel sharding in the meson test setup), so there's no cross-test interference.
- **Non-vacuousness**: confirmed above by an independent before/after build-and-run, not by trusting the PR/author text.

One pre-existing, out-of-scope observation surfaced while tracing this: `new_data_cb`'s `EXPECT_NE (100, index)` is a non-fatal check, so when it fires with the disallowed `idx=100` marker, execution falls through to `test_frames[index][i]` — i.e. `test_frames[100][i]` against a `const gint test_frames[2][48]`, which is an out-of-bounds read. I saw this concretely in the no-memset repro (garbage values like `222495366` printed instead of a bounds error). This mechanism is inherited unmodified from `action0` (already on `main` before this PR) and `suppliedValueReset` just reuses it as-is; it isn't introduced or worsened by `b39b6337`. **Nit, informational only, not a request to fix in this PR** — flagging for awareness since a future edit to `new_data_cb` should keep this in mind.

## Does `memset` risk breaking any existing use case?

Grepped every read of `sv->data`/`sv->num` in `gsttensor_if.c`: only two sites — the unconditional `data[0]`/`data[1]` read in the comparison path, and the `num`-bounded read in the property getter. No other call site, test, `.md` example, or SSAT script relies on a stale tail surviving a shorter reassignment (that reliance would itself be relying on the pre-existing bug). Confirmed no regression: full `unittest_if` suite is 22/22 with the fix, matching the state before this round's commit for everything unrelated to this change.

## Verdict

**OK to merge.** All four round-3 suggestions were addressed appropriately — three applied correctly (verified by independent rebuild, not just reading the diff) and one reasonably deferred with a sound rationale. The new `tensorIfAppsrc.suppliedValueReset` test is a genuine, non-flaky regression pin for the tail-clearing fix (confirmed to fail without the fix and pass with it, via a fresh build in this session), and the `memset` addition introduces no detectable regression to any other code path or documented usage. No blocker or major issues from this round; only one pre-existing, non-blocking nit noted for future awareness (unrelated to this commit).

myungjoo:
[Response to the round-3 re-review, on behalf of the author by a Claude agent]

The informational note is taken as well: `new_data_cb`'s "callback that is not allowed" check was `EXPECT_NE`, i.e. non-fatal, so after the real failure it went on to index `test_frames[100]` and printed the garbage the reviewer observed. It is now `ASSERT_NE`, which returns from the void callback before the buffer is mapped — no leak, `data_received` is still incremented so the count assertion in the test bodies still reports. Pre-existing and shared with `tensorIfAppsrc.action0` / `action1`, but this PR's new test uses the same path, and it is one word.

Amended into `bf9b1f2c` (was `b39b6337`). Re-verified after the amend: `unittest_if` 22/22 with the fix; with the `memset` removed, `tensorIfAppsrc.suppliedValueReset` still fails, now cleanly (the disallowed callback plus `data_received` 2 instead of 1, without the stray `test_frames[100]` output). `clang-format` clean.

myungjoo:
[Automated final re-review (round 4) by a separate Claude agent, relayed to this PR]

Scope reviewed: the delta from `b39b6337` to the current head `bf9b1f2c`, i.e. a single line in `tests/nnstreamer_if/unittest_if.cc`:

```diff
-  EXPECT_NE (100, index);
+  ASSERT_NE (100, index);
```

in `new_data_cb`. I read all four prior automated reviews, the myungjoo-bot relayed review, and all three author responses first, and I did not re-litigate anything already marked resolved. I focused on this round's three points.

## (1) Does `ASSERT_NE` in the shared `new_data_cb` affect `tensorIfAppsrc.action0` / `action1`?

Read `new_data_cb` and all three of its callers (`action0`, `action1`, `suppliedValueReset`):

- `data_received++` happens **before** the `ASSERT_NE` check, and `gst_buffer_get_memory()` (the only resource acquired in this function) happens **after** it. So on the assert-and-return path, nothing has been allocated yet — there is no leak, and `data_received` still counts the callback invocation exactly as it did with `EXPECT_NE`. This matches the author's response in `bf9b1f2c`'s commit message.
- `action0` and `action1` only ever pass `idx = 0` or `idx = 1` into the callback (never `100`), so `ASSERT_NE (100, index)` is a no-op for both — the change cannot alter their observable behavior. Verified this empirically, not just by reading: built `bf9b1f2c` as-is in the shared WSL tree (`~/if4920`) and ran the suite; `action0` and `action1` both pass.
- To specifically exercise the case this round asked about — the assert actually firing — I temporarily removed the `memset (sv->data, 0, sizeof (sv->data))` line from `gsttensor_if.c` (reintroducing the stale-tail bug from the previous round) and rebuilt. Result: `suppliedValueReset` fails cleanly at line 704 (`Expected: (100) != (index), actual: 100 vs 100`) with no stray `test_frames[100][i]` garbage read (unlike the pre-`bf9b1f2c` `EXPECT_NE` behavior described in round 3), and the test suite continues normally — `action1`, which runs immediately after in suite order, still passes with no cross-test contamination. This confirms `ASSERT_NE`'s early return is properly scoped to the single callback invocation and does not leak a failure state into sibling tests. I then restored `gsttensor_if.c` to the exact `bf9b1f2c` content (verified byte-identical via `diff` against `git show bf9b1f2c:...`) and rebuilt, leaving the shared tree as I found it: 22/22 `unittest_if` pass.

No side effect on `action0`/`action1` or any other test sharing `new_data_cb`.

## (2) Final full-diff pass (`main...bf9b1f2c`)

Diffed `main...bf9b1f2c` in full (`gsttensor_if.c` +20/-6, `gsttensor_if.md` +1, `unittest_if.cc` +218/-6). Everything beyond this round's one-line change was already covered and verified across rounds 1-3 and the myungjoo-bot review (bound check placement and correctness, `g_strfreev` on both the reject and success paths, `memset` fix and its non-vacuous regression test, doc update, float-check fix). Re-ran the full `unittest_if` suite once more against the unmodified head: **22/22 pass**, `clang-format --style=file -n --Werror` on `unittest_if.cc` is clean. No new blocker or major issue found in the full diff.

## (3) Commit message vs. actual change

`bf9b1f2c`'s message ("Clear the supplied values a new assignment does not fill") accurately describes the `memset` fix as the primary change, and its body explicitly and correctly calls out the secondary changes bundled in the same commit: the doc addition, the float-check widening in `tensorIfProp.suppliedValue`, and "made the 'callback that is not allowed' check in `new_data_cb` fatal: being non-fatal, it went on to index `test_frames[]` with the out-of-range marker and reported garbage after the real failure." This matches the actual diff exactly — no discrepancy. DCO sign-off present, not duplicated.

## Verdict

**OK to merge.** No new blocker, major, or minor issues found this round. The `EXPECT_NE` -> `ASSERT_NE` change is a one-word fix that behaves exactly as intended, has no side effect on the other tests sharing `new_data_cb`, and the commit message accurately reflects the full change. Combined with the four prior review rounds (three independent Claude passes + one relayed review), all of which converged on "OK to merge" with only non-blocking suggestions that were subsequently applied or reasonably deferred, I have no reservations about this PR being merged as-is.

PR #4936, opened 2026-09-07
[CI] Fail the Valgrind step on a memcheck error of our own

myungjoo:
The memcheck pass has never been able to fail a build. Both invocations end in `|| echo`, and valgrind runs without `--error-exitcode`, so a finding never reaches the step's exit status in the first place. This makes it fail on a memcheck error that is ours, and only on that.

Follows #4921, which brought the step to 7-12 min over 22 binaries, and #4929, which removed the last memcheck errors this repository was responsible for.

## Why not suppressions

The reason the pass was advisory is real and is not going away: a run reports errors from GLib, from ld.so and from the inference runtimes, and those belong to other projects.

Suppressions cannot separate them, because they match a stack. The stack of a block GLib allocated and *we* failed to free looks like the stack of one GLib allocated and leaked itself; a suppression written against our own frames would hide our bugs along with the noise. The repository's own file shows the same problem from the other side — `tools/debugging/valgrind_suppression` has 28 entries and suppressed **2** errors across a whole run, because it was written for `gst-launch` and python, not for these binaries. Growing it is not the path.

## What this does instead

`tools/debugging/check_valgrind_log.sh` attributes each error to the frame memcheck blames for it — the first one that is not valgrind's own allocator wrapper — and calls that frame ours when it names a source file this repository holds or an object under the build directory. Only a non-leak error attributed to us fails.

Measured on real logs:

| | errors of ours | library errors | leak reports | binaries |
|---|---|---|---|---|
| [34094878222](https://github.com/nnstreamer/nnstreamer/actions/runs/34094878222), current main | **0** | 14 | 167 | 23 |
| [33853964542](https://github.com/nnstreamer/nnstreamer/actions/runs/33853964542), before #4929 | **8** | 13 | 158 | 22 |

The 8 are exactly the uninitialised reads `gst_tensor_meta_info_parse_header` and `gst_tensor_meta_info_validate` were making on tensor_transform's unfilled output memory, which #4929 fixed. So the gate goes in with **no baseline file**: there is nothing to forgive today, and anything that appears afterwards is new.

And in this branch's own CI ([34099915322](https://github.com/nnstreamer/nnstreamer/actions/runs/34099915322)) the gate runs for real and says the same thing:

```
valgrind log check: 0 error contexts from this repository, 14 from libraries, 162 leak reports, 23 binaries examined
```

I also checked the rule cannot misfire on a name collision. Of the 67 source file names that appear in frames across both logs, the 39 the repository also holds are all genuinely ours (nnstreamer sources and test sources); the other 28 — `dl-load.c`, `strcmp.S`, `rtld-malloc.h`, `pthread_create.c`, `gtest.h` and so on — exist in no file of ours.

## What stays out

**Leaks.** A leak is reported at its allocation, which for most of ours is inside GLib, so the frame that would decide responsibility is further down the stack than this rule looks. In that same run, 77 of 89 leak contexts have a frame of ours somewhere — 56 of them in test code. Gating on that rule would be wrong in both directions. They are counted and printed, and left for their own change.

**Test failures.** The two runs are piped through `tee`, so the runner's exit status is the pipeline's last command and a failing test binary still does not fail the step, exactly as `|| echo` did. What replaces `|| echo` is the checker's own verdict.

`--require-output` covers the one case that arrangement would otherwise hide: a log that lost its memcheck output entirely — a runner that died early, a flag that stopped matching — is indistinguishable from a clean one, so the step asks for it explicitly. The two runs write separate logs and are checked separately, so neither can satisfy that on the other's behalf, and both logs are kept as an artifact when the job fails.

**Reports of a kind the checker does not know.** Printed as a `::warning::` naming the line rather than silently counted as nothing, so a new or renamed memcheck message surfaces instead of quietly leaving the gate. It does not fail: a new announcement turning every job red is the worse failure. Both reference logs produce no warnings — the only unclassified lines in them were `Process terminating with default action of signal 6` and a `Thread N ...` header, which are recognised as the announcements they are.

## Blast radius

`packaging/run_unittests_binaries.sh` is untouched. It has five callers and only this step passes `--valgrind`, so `debian/rules`, `nnstreamer.spec`, `risc-v.yml` and the docker image see no change at all. The two new files live under `tools/`, which no packaging file ships. The step is already x86-only.

Worth knowing: `unittest_*.cc` are sources of ours, so a memory error in test code would gate too. There are none today, and an invalid read in a test is a bug worth stopping for.

## Testing

`tools/debugging/test_check_valgrind_log.sh` pins the rule against hand-written logs, wired into *Static checks* with the same merge-ref guard as the other self-tests: an error in a source of ours fails; the same error in `strcmp.S` or in a system object does not; valgrind's wrapper is skipped so the frame under it decides; skipping the wrapper does not by itself make an error ours; a leak allocated by us does not gate; a build-directory object does gate; the CI timestamp prefix is tolerated; a repeat is counted once and the same error in two binaries twice; a missing file and a missing argument are usage errors rather than passes; and `--require-output` rejects an empty log while accepting one that merely has no error of ours.

Every case that expects a pass also asserts the counts in the summary line, so a checker that silently parsed nothing could not satisfy them.

Verified locally with shellcheck clean, the self-test green, and both CI logs above run through the checker end to end. The pipeline semantics were checked separately under `bash -e` without `pipefail`, which is what GitHub gives this step: a failing runner does not abort it, and the checker does.

## Not in this change

- The SSAT valgrind step on decoder-bounding-box. Its output is not captured, and SSAT's own verdict does not see memcheck findings either.
- The leak gate, and the 21 leak contexts in shipped code that would have to be dealt with first.
- `tools/debugging/valgrind_suppression`. It is nearly inert, but nothing here depends on it.

### DIFF
--- a/.github/workflows/static.check.yml
+++ b/.github/workflows/static.check.yml
@@ -201,6 +201,19 @@ jobs:
           else
             echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
           fi
+      - name: /Checker/ Valgrind log checker self-test
+        # Runs a repository script unrelated to $changed_file_list, so it
+        # needs the same merge-ref guard as the steps above.
+        run: |
+          selftest=tools/debugging/test_check_valgrind_log.sh
+          if [ -f "$selftest" ]; then
+            bash "$selftest"
+          elif git show --pretty="format:" --name-only --no-renames --diff-filter=D ${{ github.event.pull_request.head.sha }} -${{ github.event.pull_request.commits }} | grep -qxF "$selftest"; then
+            echo "::error::This PR removes $selftest; remove its workflow step too, or restore the script (also update this step if the script moved)."
+            exit 1
+          else
+            echo "Skipped: $selftest is not in this branch's tree and this PR does not remove it."
+          fi
       - name: /Checker/ SSAT install list is in sync
         # Same guard as the check-rebuild self-test above: the step list comes
         # from the merge ref while the checkout is the PR head, so a branch that
--- a/.github/workflows/ubuntu_clean_meson_build.yml
+++ b/.github/workflows/ubuntu_clean_meson_build.yml
@@ -244,6 +244,17 @@ jobs:
     # unit test" step above. See
     # https://github.com/nnstreamer/nnstreamer/issues/4894 for the measurements
     # and for restoring them in a job that can afford them.
+    #
+    # What is left is then checked, and an error memcheck blames on a frame of
+    # ours fails the step. Errors blamed on a library do not - they are not
+    # ours to fix and there is a steady population of them - and leaks do not
+    # either, because a leak is reported at its allocation, which for most of
+    # ours is inside GLib. Both are still printed. The gate starts with nothing
+    # to forgive: the last errors of our own went with the tensor_transform fix
+    # in #4929, so there is no baseline file to keep in step with the code. A
+    # test failure in this pass still does not fail the step. Each run is
+    # checked on its own log, so one falling silent cannot be covered by the
+    # other having run.
     - name: GTEST run with Valgrind on a case
       if: env.rebuild == '1' && matrix.os == 'ubuntu-22.04'
       timeout-minutes: 30
@@ -256,7 +267,6 @@ jobs:
         export NNSTREAMER_CONF=`pwd`/build/nnstreamer-test.ini
         export G_SLICE=always-malloc
         export G_DEBUG=gc-friendly
-        note="There are Valgrind errors. Please fix it. As we have a lot of Valgrind errors from different libraries and possible from nnstreamer itself, we are not halting the build with Valgrind errors until we get them all."
         skip_cases="nnstreamerFilterArmnn.invokeAdvanced"
         skip_cases="${skip_cases}:nnstreamerFilterLiteRT.sharedEnvInvokeDuringConfigure"
         skip_cases="${skip_cases}:nnstreamerFilterLiteRT.zeroCopyRepeatedInvoke"
@@ -269,8 +279,18 @@ jobs:
         llamacpp_cases="${llamacpp_cases}:NNStreamerFilterLlamaCppTest.invalidModel_n"
         llamacpp_cases="${llamacpp_cases}:NNStreamerFilterLlamaCppTest.zeroTokenGenerationSync_p"
         llamacpp_cases="${llamacpp_cases}:NNStreamerFilterLlamaCppTest.zeroTokenGenerationAsync_p"
-        GTEST_FILTER="-${skip_cases}" ./packaging/run_unittests_binaries.sh --valgrind --recursive --skip unittest_filter_llamacpp ./tests/ || echo "${note}"
-        GTEST_FILTER="${llamacpp_cases}" ./packaging/run_unittests_binaries.sh --valgrind ./tests/unittest_filter_llamacpp || echo "${note}"
+        GTEST_FILTER="-${skip_cases}" ./packaging/run_unittests_binaries.sh --valgrind --recursive --skip unittest_filter_llamacpp ./tests/ 2>&1 | tee valgrind-tests.log
+        GTEST_FILTER="${llamacpp_cases}" ./packaging/run_unittests_binaries.sh --valgrind ./tests/unittest_filter_llamacpp 2>&1 | tee valgrind-llamacpp.log
+        checked=0
+        bash ./tools/debugging/check_valgrind_log.sh --require-output valgrind-tests.log || checked=1
+        bash ./tools/debugging/check_valgrind_log.sh --require-output valgrind-llamacpp.log || checked=1
+        exit ${checked}
+    - name: Upload the Valgrind logs if something went wrong
+      if: failure() && env.rebuild == '1' && matrix.os == 'ubuntu-22.04'
+      uses: actions/upload-artifact@v4
+      with:
+        name: Valgrind_logs
+        path: valgrind-*.log
     - name: get NNStreamer example apps
       if: env.rebuild == '1' && matrix.os == 'ubuntu-22.04'
       uses: actions/checkout@v4
--- a/tools/debugging/check_valgrind_log.sh
+++ b/tools/debugging/check_valgrind_log.sh
@@ -1,0 +1,256 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# @file     check_valgrind_log.sh
+# @brief    Fail when a memcheck error comes from this repository's own code.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# usage: check_valgrind_log.sh [--repo-root DIR] [--require-output] LOGFILE...
+#
+# Reads the text memcheck writes and reports each error twice over: once by
+# what kind of error it is, and once by whose code the reported frame is in.
+# Only a non-leak error whose reported frame is ours makes this exit non-zero.
+#
+# An error is attributed to the frame memcheck blames for it - the first one
+# that is not valgrind's own allocation wrapper - and that frame is ours when
+# it names a source file this repository holds or an object under the build
+# directory. Nothing below that frame counts, so neither the origin stack that
+# --track-origins appends nor the stack of the block an address belongs to can
+# make an error ours; both are usually ours even when the error is not.
+#
+# The cost of that is an error blamed on a library function we called badly,
+# where our frame is the one below. Memcheck redirects the string and memory
+# functions through its own wrappers, which this skips, so in practice the
+# blame lands on our caller anyway; across the two runs this was measured on,
+# no error blamed on a library had a frame of ours beneath it.
+#
+# An error every one of whose frames is a wrapper therefore has nothing to be
+# blamed on, and is counted as neither. Deciding otherwise would mean carrying
+# the error past the end of its own stack, which is the mistake this rule
+# exists to prevent; neither of the two runs holds such an error.
+#
+# Both halves of "ours" carry an assumption. A file is matched by its base
+# name, so a source of ours sharing a name with one a library was built from
+# would start gating on that library's bug - no name in the tree does today.
+# The object half looks for /build/gst, /build/ext or /build/tests anywhere in
+# the path, which is what lets a log recorded on another machine be checked
+# here, and which stops recognising objects if the build directory is ever
+# renamed; the file-name half still covers everything built with debug info.
+#
+# A report whose first line matches none of the kinds below is neither passed
+# nor failed on, so it is printed as a warning instead of disappearing.
+#
+# Leaks are counted and printed but never gate, because a leak is reported at
+# its allocation, which for most of ours is inside GLib; deciding who was
+# responsible for freeing it needs a rule this script does not have.
+#
+# --require-output additionally fails when the log holds no memcheck output at
+# all. A caller that always runs memcheck wants that, because a log that lost
+# it - a runner that died early, a flag that stopped matching - would otherwise
+# be indistinguishable here from a clean one.
+#
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+repo_root=$(cd "${SCRIPT_DIR}/../.." && pwd)
+require_output=""
+logs=()
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --repo-root)
+      if [ $# -lt 2 ]; then
+        echo "$0: --repo-root requires an argument" >&2
+        exit 2
+      fi
+      repo_root=$2
+      shift 2
+      ;;
+    --require-output)
+      require_output=1
+      shift
+      ;;
+    -h|--help)
+      echo "usage: $(basename "$0") [--repo-root DIR] [--require-output] LOGFILE..." >&2
+      exit 0
+      ;;
+    *)
+      logs+=("$1")
+      shift
+      ;;
+  esac
+done
+
+if [ ${#logs[@]} -eq 0 ]; then
+  echo "$0: no log file given" >&2
+  exit 2
+fi
+
+for log in "${logs[@]}"; do
+  if [ ! -f "${log}" ]; then
+    echo "$0: ${log}: no such file" >&2
+    exit 2
+  fi
+done
+
+if [ ! -d "${repo_root}" ]; then
+  echo "$0: ${repo_root}: no such directory" >&2
+  exit 2
+fi
+
+if ! sources=$(mktemp); then
+  echo "$0: cannot create a temporary file" >&2
+  exit 2
+fi
+trap 'rm -f "${sources}"' EXIT
+
+find "${repo_root}" \
+    -path "${repo_root}/.git" -prune -o \
+    -path "${repo_root}/build" -prune -o \
+    -type f \( -name '*.c' -o -name '*.cc' -o -name '*.cpp' -o \
+               -name '*.h' -o -name '*.hh' -o -name '*.hpp' \) -print |
+  sed 's|.*/||' | sort -u > "${sources}"
+
+# Without this list every frame looks like a library's, and the check would
+# pass whatever it was given.
+if [ ! -s "${sources}" ]; then
+  echo "$0: ${repo_root}: no source file found; is that the repository root?" >&2
+  exit 2
+fi
+
+awk -v sources="${sources}" -v require_output="${require_output}" '
+BEGIN {
+  while ((getline name < sources) > 0)
+    ours_source[name] = 1
+  close(sources)
+  binary = "(unknown)"
+}
+
+# Errors are reported by the frame memcheck blames, so skip the wrappers it
+# substitutes for the allocator and for the string functions.
+function is_valgrind_own(where) {
+  return where ~ /^vg_replace_/
+}
+
+function is_ours(where,   file) {
+  if (where ~ /^in /) {
+    return (where ~ /\/build\/(gst|ext|tests)\//)
+  }
+  file = where
+  sub(/:[0-9]+$/, "", file)
+  return (file in ours_source)
+}
+
+function record(kind, where, frame,   key) {
+  key = binary SUBSEP kind SUBSEP where
+  if (key in seen)
+    return
+  seen[key] = 1
+  if (is_ours(where)) {
+    ours_count++
+    ours_binary[ours_count] = binary
+    ours_kind[ours_count] = kind
+    ours_frame[ours_count] = frame
+  } else {
+    library_count++
+  }
+}
+
+{
+  line = $0
+  sub(/^[0-9-]+T[0-9:.]+Z /, "", line)
+  if (line !~ /^==[0-9]+== /)
+    next
+  sub(/^==[0-9]+== /, "", line)
+}
+
+line ~ /^Command: / {
+  examined++
+  binary = line
+  sub(/^Command: /, "", binary)
+  sub(/ .*$/, "", binary)
+  sub(/.*\//, "", binary)
+  next
+}
+
+line ~ /^[0-9,]+ (\([^)]*\) )?bytes in [0-9,]+ blocks are .*lost/ {
+  leak_count++
+  pending = 0
+  candidate = ""
+  next
+}
+
+# Memcheck writes these above a stack of their own; they are part of an error
+# already counted, or an announcement, not a kind this script should classify.
+# Ending the error here matters as much as not classifying it: the stack that
+# follows is where the block was allocated, and blaming an error on that is
+# how a bug in a library becomes ours.
+line ~ /^ *(Address 0x|Uninitialised value was created|Block was alloc|Location 0x|At least one of|Process terminating|Thread [0-9]+|Access not within mapped region|Bad permissions for mapped region|General Protection Fault|Stack overflow in thread)/ {
+  pending = 0
+  candidate = ""
+  next
+}
+
+line ~ /^(Invalid (read|write|free|memory pool|alignment)|Mismatched free|Conditional jump or move depends on uninitialised|Use of uninitialised|Syscall param |Source and destination overlap|Jump to the invalid address|Argument .* has a (fishy|bad)|realloc\(\) with size 0)/ {
+  pending = 1
+  pending_kind = line
+  sub(/ of size [0-9]+$/, "", pending_kind)
+  candidate = ""
+  next
+}
+
+line ~ /^ +(at|by) 0x[0-9A-Fa-f]+: / {
+  if (candidate != "") {
+    unclassified_count++
+    unclassified[unclassified_count] = candidate
+    candidate = ""
+  }
+  if (!pending)
+    next
+  where = line
+  if (where ~ /\([^)]*\)$/) {
+    sub(/^.*\(/, "", where)
+    sub(/\)$/, "", where)
+  } else {
+    next
+  }
+  if (is_valgrind_own(where))
+    next
+  frame = line
+  sub(/^ +/, "", frame)
+  record(pending_kind, where, frame)
+  pending = 0
+  next
+}
+
+{
+  pending = 0
+  candidate = (line ~ /^ *$/) ? "" : line
+}
+
+END {
+  if (require_output != "" && examined == 0) {
+    print "valgrind log check: the log holds no memcheck output at all."
+    exit 1
+  }
+  printf "valgrind log check: %d error contexts from this repository, ", ours_count
+  printf "%d from libraries, %d leak reports, %d binaries examined\n", library_count, leak_count, examined
+  for (i = 1; i <= unclassified_count; i++)
+    printf "::warning::valgrind log check does not classify this report, so it can neither pass nor fail on it: %s\n", unclassified[i]
+  if (ours_count == 0)
+    exit 0
+  print ""
+  print "Memcheck errors reported in this repository'\''s own code:"
+  for (i = 1; i <= ours_count; i++) {
+    printf "  [%s] %s\n", ours_binary[i], ours_kind[i]
+    printf "      %s\n", ours_frame[i]
+  }
+  exit 1
+}
+' "${logs[@]}"
--- a/tools/debugging/test_check_valgrind_log.sh
+++ b/tools/debugging/test_check_valgrind_log.sh
@@ -1,0 +1,357 @@
+#!/usr/bin/env bash
+
+##
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved.
+#
+# SPDX-License-Identifier: LGPL-2.1-only
+#
+# @file     test_check_valgrind_log.sh
+# @brief    Self-test for check_valgrind_log.sh.
+# @see      https://github.com/nnstreamer/nnstreamer
+# @author   MyungJoo Ham <myungjoo.ham@samsung.com>
+#
+# Each case is a hand-written memcheck log against a throwaway source tree, so
+# that what the checker is supposed to key on - the file a frame names, the
+# object a frame names, whether the error is a leak - is the only thing that
+# varies between them.
+#
+# The cases that expect a pass also assert the counts in the summary line. A
+# checker that parsed nothing at all would exit 0 for every log, and without
+# those counts every one of them would agree with it.
+#
+
+set -u
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+CHECKER="${SCRIPT_DIR}/check_valgrind_log.sh"
+workdir=$(mktemp -d)
+repo="${workdir}/repo"
+failed=0
+
+trap 'rm -rf "${workdir}"' EXIT
+
+mkdir -p "${repo}/gst"
+touch "${repo}/gst/myelement.c" "${repo}/gst/myelement.h"
+
+##
+# @brief Report one expectation and remember a failure.
+# @param $1 0 when the expectation held, anything else when it did not
+# @param $2 description of the expectation
+report() {
+  if [ "$1" -eq 0 ]; then
+    echo "PASS: $2"
+  else
+    echo "FAIL: $2"
+    failed=1
+  fi
+}
+
+##
+# @brief Run the checker on a log written from stdin.
+# @param $1 name of the case, used for the log file
+# @param $@ further options handed to the checker
+# Sets `status` and `output`.
+run_case() {
+  local name=$1
+  local log="${workdir}/${name}.log"
+  shift
+  cat > "${log}"
+  output=$(bash "${CHECKER}" --repo-root "${repo}" "$@" "${log}" 2>&1)
+  status=$?
+}
+
+##
+# @brief Assert the checker's exit status.
+# @param $1 expected status
+# @param $2 description of the expectation
+expect_status() {
+  if [ "${status}" -eq "$1" ]; then
+    report 0 "$2"
+  else
+    report 1 "$2 (got status ${status}: ${output})"
+  fi
+}
+
+##
+# @brief Assert that the checker's output contains the given text.
+# @param $1 text that must appear
+# @param $2 description of the expectation
+expect_in() {
+  case "${output}" in
+    *"$1"*) report 0 "$2" ;;
+    *) report 1 "$2 (output: ${output})" ;;
+  esac
+}
+
+##
+# @brief Assert that the checker's output does not contain the given text.
+# @param $1 text that must not appear
+# @param $2 description of the expectation
+expect_not_in() {
+  case "${output}" in
+    *"$1"*) report 1 "$2 (output: ${output})" ;;
+    *) report 0 "$2" ;;
+  esac
+}
+
+if [ ! -f "${CHECKER}" ]; then
+  echo "FAIL: ${CHECKER} is missing."
+  exit 1
+fi
+
+run_case ours <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid read of size 4
+==1==    at 0x1111: my_element_chain (myelement.c:42)
+==1==    by 0x2222: g_closure_invoke (in /usr/lib/libgobject-2.0.so.0)
+==1==
+EOF
+expect_status 1 "an error reported in a source of this repository fails the check"
+expect_in "myelement.c:42" "the failing error is printed with its frame"
+expect_in "unittest_demo" "the failing error is printed with its binary"
+
+run_case library <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid read of size 8
+==1==    at 0x3333: strncmp (strcmp.S:172)
+==1==    by 0x4444: my_element_chain (myelement.c:42)
+==1==
+EOF
+expect_status 0 "an error blamed on a library passes even when we called it"
+expect_in "1 from libraries" "the library error is still counted"
+
+run_case wrapper_then_ours <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid write of size 1
+==1==    at 0x5555: malloc (vg_replace_malloc.c:381)
+==1==    by 0x6666: my_element_init (myelement.c:11)
+==1==
+EOF
+expect_status 1 "valgrind's own wrapper is skipped when attributing an error"
+expect_in "myelement.c:11" "attribution lands on the first frame that is not the wrapper"
+
+run_case wrapper_then_library_n <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid write of size 1
+==1==    at 0x5555: malloc (vg_replace_malloc.c:381)
+==1==    by 0x6666: g_malloc (in /usr/lib/libglib-2.0.so.0)
+==1==
+EOF
+expect_status 0 "skipping the wrapper does not by itself make an error ours"
+expect_in "1 from libraries" "that error is counted as a library one"
+
+run_case object_ours <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid read of size 4
+==1==    at 0x7777: some_symbol (in /home/runner/work/nnstreamer/build/gst/libnnstreamer.so)
+==1==
+EOF
+expect_status 1 "a frame naming an object under the build directory is ours"
+
+run_case object_library_n <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid read of size 4
+==1==    at 0x7777: some_symbol (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0)
+==1==
+EOF
+expect_status 0 "a frame naming a system object is not ours"
+expect_in "1 from libraries" "that error is counted as a library one"
+
+run_case leak <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== 400 bytes in 1 blocks are definitely lost in loss record 3 of 9
+==1==    at 0x8888: my_element_init (myelement.c:11)
+==1==
+EOF
+expect_status 0 "a leak allocated by this repository does not fail the check"
+expect_in "1 leak reports" "the leak is counted"
+expect_not_in "myelement.c:11" "the leak is not reported as a failing error"
+
+run_case timestamped <<'EOF'
+2026-09-07T07:21:27.4959062Z ==1== Command: ./tests/unittest_demo
+2026-09-07T07:21:27.4959062Z ==1== Invalid read of size 4
+2026-09-07T07:21:27.4959062Z ==1==    at 0x1111: my_element_chain (myelement.c:42)
+2026-09-07T07:21:27.4959062Z ==1==
+EOF
+expect_status 1 "a log carrying the timestamps a CI run adds is still parsed"
+
+run_case repeated <<'EOF'
+==1== Command: ./tests/unittest_demo
+==1== Invalid read of size 4
+==1==    at 0x1111: my_element_chain (myelement.c:42)
+==1==
+==1== Invalid read of size 4
+==1==    at 0x1111: my_element_chain (myelement.c:42)
+==1==
+EOF
+expect_status 1 "a repeated error still fails the check"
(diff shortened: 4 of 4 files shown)


myungjoo:

Reviewed head `12b6bbb9`, one commit, 4 files (+223/-3). Follows #4921 (step-time bound) and #4929 (the last errors of our own). Scope: `tools/debugging/check_valgrind_log.sh`, `tools/debugging/test_check_valgrind_log.sh`, `.github/workflows/ubuntu_clean_meson_build.yml`, `.github/workflows/static.check.yml`.

**Verdict: Changes requested.** The blast-radius and shell-semantics claims all check out, but I found two verified parsing bugs in the checker that is the whole point of this PR — one of them makes an entire class of memcheck error invisible to the gate, including when it is ours.

## Verified correct (no objection)

- **`packaging/run_unittests_binaries.sh` really is untouched.** `git diff main...HEAD --stat -- packaging/run_unittests_binaries.sh` against this branch is empty. None of the four other callers (`debian/rules`, `packaging/nnstreamer.spec`, `.github/workflows/risc-v.yml`, the docker image) see any change, so the #4921 pdebuild regression cannot recur here.
- **Shell semantics.** Neither workflow file sets `shell:`, so this step runs under the un-elevated `bash -e {0}` GitHub uses when `shell` is left implicit (pipefail is *not* on, unlike the `bash --noprofile --norc -eo pipefail {0}` you get by writing `shell: bash` explicitly). `cmd 2>&1 | tee -a "${log}"` therefore reports `tee`'s exit status, not the runner's, so a failing test binary still doesn't fail the step — matching the old `|| echo` behavior on purpose. The final `bash ./tools/debugging/check_valgrind_log.sh --require-output "${log}"` is not inside a pipe, so under `-e` its own exit status does end the step. This is correctly built.
- **Static checks wiring.** The new self-test step in `static.check.yml` reuses the exact merge-ref guard idiom (`if [ -f "$selftest" ]; ... elif git show ...--diff-filter=D...`) already used by the seven other self-test steps in that job, byte-for-byte. Nothing is copy-paste-broken here.
- **mawk compatibility, actually run.** I ran `tools/debugging/test_check_valgrind_log.sh` under a real `mawk` (not gawk) — all 26 assertions pass. No `gensub`, no `{n,m}` intervals, no `length(array)`; the script is mawk-clean as claimed.
- File location (`tools/debugging/`, alongside `valgrind_suppression`), the `bash <script>` calling convention (matches `tools/executorch-install.sh`), and the size (223 net new lines, split roughly 1:1 between checker and self-test) are all proportionate and don't touch unrelated modules.

## Blocking: two verified bugs in the attribution rule itself

**1. "Source and destination overlap" errors are invisible — not even counted as a library error.**
The header regex is `/^(Invalid (read|write|free|memory pool)|Mismatched free|Conditional jump or move depends on uninitialised|Use of uninitialised|Syscall param |Source and destination .* overlap|Jump to the invalid address|Argument .* has a fishy)/`. Real memcheck text for this error is `Source and destination overlap in strcpy(0x4a2050, 0x4a2050)` — "destination" is followed immediately by "overlap" with a single space, so `Source and destination .* overlap` needs a *second* occurrence of "overlap" to close the match, which never exists. I reproduced this against both gawk and real mawk (WSL):

```
==1== Command: ./tests/unittest_demo
==1== Source and destination overlap in strcpy(0x4a2050, 0x4a2050)
==1==    at 0x1111: my_element_chain (myelement.c:42)
==1==    by 0x2222: g_closure_invoke (in /usr/lib/libgobject-2.0.so.0)
==1==
```
→ `valgrind log check: 0 error contexts from this repository, 0 from libraries, 0 leak reports, 1 binaries examined`, exit 0 — even though the top frame names `myelement.c`, a source this repository holds. The error doesn't just get misclassified as a library issue; it vanishes from every count. This is exactly the defect class (an overlapping-buffer `strcpy`/`memcpy` call) this element-heavy C codebase is likely to produce, and it's also not pinned by any case in `test_check_valgrind_log.sh`. Suggest anchoring the header regex to the real string (`^Source and destination overlap`) and adding a test case for it.

**2. The checker fails open, not closed, when its own `mktemp` fails.**
`check_valgrind_log.sh` uses `set -u` but not `set -e`, and nothing checks that `sources=$(mktemp)` actually succeeded before `find ... > "${sources}"` and `awk -v sources="${sources}" ...` run. I forced this (nonexistent `TMPDIR`) under real mawk:

```
$ TMPDIR=/tmp/does/not/exist/here bash tools/debugging/check_valgrind_log.sh --repo-root ... ours_test.log
mktemp: failed to create file via template ... No such file or directory
check_valgrind_log.sh: line 89: : No such file or directory
valgrind log check: 0 error contexts from this repository, 1 from libraries, 0 leak reports, 1 binaries examined
$ echo $?
0
```
The log in that run contains a plain `Invalid read of size 4` at `myelement.c:42` — a source of ours — and the step still passes. Under gawk the same failure happens to be fatal (`getline < ""` aborts), but the CI runner's default `/usr/bin/awk` is mawk, and mawk swallows it silently. A `mktemp` failure is a rare runner hiccup (disk pressure, `/tmp` issue), but the entire purpose of this PR is to make this step a hard gate — failing open on infra trouble, with a real error of ours sitting right there in the log, is the one failure mode a gate like this shouldn't have. Recommend an explicit check right after building `sources` (e.g. bail loudly if the `find`/`mktemp` step didn't produce a non-empty, readable file) instead of letting it degrade silently into "nothing is ours."

## Worth tightening (not blocking)

**3. The `/build/` object-path rule isn't anchored to `repo_root`.** `is_ours()` treats any frame with `(in /path/to/obj)` as ours iff the path contains the literal substring `/build/` anywhere — `repo_root` is only used to build the source-filename list, never passed into the awk program for this check. I didn't find a live collision in this repo's current CI (ORT/ArmNN/tflite come from apt into `/usr/lib`, LiteRT is unzipped into `$HOME/litert-sdk`, llama.cpp is a prebuilt release copied to `/usr/local`), so today's "0 errors of ours" measurement should hold. But CMake's own default build-directory name is literally `build`, and Debian `sbuild`/`pbuilder` chroots build packages under paths like `/build/<pkg>-<ver>/`, so a future vendored or system library recorded with such a path would be misattributed to us. Given how central this rule is to the whole gate, anchoring it to `${repo_root}/build` (or at least requiring the path to start under a fixed prefix) would remove the ambiguity rather than rely on today's absence of a match.

**4. Test coverage gap around secondary stacks.** None of the 15 self-test cases exercise `--track-origins=yes`'s "Uninitialised value was created by ..." second stack, or the "Address 0x... is N bytes after a block of size M alloc'd" second stack that `Invalid read`/`Invalid write` errors carry. I checked both by hand and the current code does the right thing (only the first non-wrapper frame of the *primary* stack is ever recorded; frames appearing after `pending` is reset are skipped because `if (!pending) next`), but nothing in the suite pins that. Given this was explicitly called out as a risk area, it's worth a dedicated test case for both shapes — and one for the overlap/fishy-argument message classes once finding #1 is fixed — so a future edit to the state machine can't regress it silently.

## Minor / non-blocking

- `check_valgrind_log.sh` and `test_check_valgrind_log.sh` are committed with the executable bit (`100755`), while every comparable helper in `tools/` and `.github/workflows/static.check.scripts/` that is invoked the same way — `bash <script>` — is `100644` (e.g. `tools/executorch-install.sh`'s siblings, all of `static.check.scripts/*.sh`). Harmless (the executable-bits checker only flags `.c/.h/.cpp/...`, not `.sh`), just inconsistent with the rest of the tree.
- `tools/debugging/README.md` documents the other files in that directory but isn't updated for the two new scripts. `valgrind_suppression` itself isn't documented there either, so this isn't a regression, and there's no API/architecture change here to require a doc update — just an optional polish item.
- Forward-looking only: `types.h` exists in this repository's own source tree, a fairly generic name. The author's empirical 67-filename check against two real logs is sound for what was actually observed, but the rule is a bare filename match with no path disambiguation, so it's worth knowing this could need a second look if a frame ever attributes an inlined function to a same-named header from GLib/GStreamer or elsewhere.

## Summary

Correctly scoped, doesn't touch `run_unittests_binaries.sh` or its other callers, shell semantics are right, and mawk-clean as claimed. But the attribution rule that is the actual deliverable has a verified false-negative (overlap errors vanish entirely, even when they're ours) and a verified fail-open path (a `mktemp` hiccup turns a real error of ours into a clean pass) — both reproduced against the CI's actual `awk` (mawk). Given the stated goal is "anything that appears afterwards is new" with **no baseline file**, these two gaps mean a real regression could land without the gate ever tripping. Recommend fixing both before merge, and tightening the `/build/` heuristic and adding the missing test cases while in there.

myungjoo:
Both blocking findings were real; I reproduced each before changing anything, and both are fixed at `5cc4e99e`. Thank you — the second one in particular was the kind of bug this whole change exists to avoid.

**1. `Source and destination overlap` never matched.** Confirmed: the pattern needed a second " overlap" that the real message never has, so the error was not even counted as a library one. Fixed to match the message valgrind actually writes, and while there I added the kinds the whitelist was also missing: `Invalid alignment`, `Argument ... has a bad value`, `realloc() with size 0`. The effect is visible in the two reference logs, where the library count goes 12 → 14 and 11 → 13: those are the LiteRT-internal `memcpy` overlaps, which were invisible before and are now counted and correctly attributed to a library.

**2. Failing open when `mktemp` fails.** Confirmed, and confirmed that it is mawk-specific — gawk aborts, mawk carries on with an empty source list and calls every frame a library's. The fix does not rely on either awk's behaviour: `mktemp` is checked, and the source list is then required to be non-empty, which covers a failed `mktemp`, a wrong `--repo-root` and a broken `find` in one place. Verified under mawk with a broken `TMPDIR`: exit 2 with a message, where it used to exit 0.

**3. `/build/` unanchored.** Agreed it was loose. Anchoring it to `--repo-root` would have broken the case of checking a log downloaded from CI on another machine, which is how I did most of the measurement for this PR, so I tightened the pattern instead: `/build/(gst|ext|tests)/`, the three places our objects are built. In both reference logs every object-path frame under a build directory is `build/tests/...`, so nothing is lost. A new case pins that `/opt/vendor/build/lib/libvendor.so` is not ours.

**4. Test gaps.** Added, and they pin the rule rather than just the code: an error whose blamed frame is a library but whose `--track-origins` origin stack is ours does **not** gate, and neither does one whose "Address ... alloc'd" stack is ours. That is a deliberate choice — nothing below the blamed frame counts, because those secondary stacks are usually ours even when the error is not — and it is now written in the file header as well. Also added cases for the two message kinds that were broken or missing. The self-test is 37 assertions and passes under both mawk and gawk.

**5. Executable bit.** Not what happened here: both files are `100644` in the index (`git ls-files -s tools/debugging/*.sh`). The working tree shows them executable because this checkout is on a filesystem without permission bits; git records the mode from `core.fileMode`, which is off. Nothing to change.

**On `tools/debugging/README.md`:** it is a guide to GStreamer's own debugging facilities — `GST_DEBUG`, dot-graph dumps, gst-debugger — and has no index of the scripts beside it. A CI checker does not belong in it, so I left it alone.

Re-verified after the changes: the current main log still reports 0 errors of ours over 23 binaries and exits 0, the pre-#4929 log still reports exactly the 8 that fix removed and exits 1, shellcheck is clean, and the self-test passes under both awks.

myungjoo:

Re-reviewed at head `5cc4e99e` (single commit, amend of `12b6bbb9`). Read the prior review (`issuecomment-5567382122`) and the author's reply (`issuecomment-5567434131`) first, then re-verified each claimed fix independently rather than taking the reply at its word — against a real `mawk` (WSL, not gawk-as-mawk), and against the two real CI logs the PR body cites (runs `34094878222` and `33853964542`, pulled via `gh run view --log` and re-fed to both checker versions).

**Verdict: no remaining objection. Both blocking findings are genuinely fixed, and the fixes hold up against real data, not just the new synthetic test cases.**

## Blocking findings — both confirmed fixed

**1. "Source and destination overlap" regex.** `git diff 12b6bbb9 5cc4e99e` shows the header pattern changed from `Source and destination .* overlap` (which needed a second, nonexistent, "overlap" to close the match) to a literal `Source and destination overlap`. Verified two ways:
- Synthetic: fed the exact message from the original finding (`Source and destination overlap in strcpy(0x4a2050, 0x4a2050)`) with a blamed frame naming a fixture source file — now `exit 1`, correctly attributed to "this repository". Before the fix, same input gave `0/0/0`, silently.
- At scale, against the two real logs the PR measures against: with the **old** checker (`12b6bbb9`), the current-main log reports `12 from libraries`; with the **new** checker (`5cc4e99e`), the same log reports `14`. The pre-#4929 log goes `11 → 13` the same way. Both deltas are exactly +2, matching the author's reply, and in both logs the "ours" count is byte-for-byte unchanged (`0` and `8` respectively) — so the fix adds exactly the previously-invisible LiteRT `memcpy_chk` overlaps to the library bucket, without disturbing anything already counted correctly. No regression.

**2. Fail-open on `mktemp` failure.** Reproduced the original PR's exact repro under real mawk in WSL (`readlink -f /usr/bin/awk` confirmed the shim was live): `TMPDIR=/tmp/does/not/exist/here bash check_valgrind_log.sh --repo-root <root> <log>` now exits **2** with `check_valgrind_log.sh: cannot create a temporary file`, where it used to silently exit 0. The fix is two-layered as claimed (`if ! sources=$(mktemp)` plus a separate `[ ! -s "${sources}" ]` guard after the `find`), so it also covers a `find` that produces nothing for an unrelated reason (bad `--repo-root`), not just a bare `mktemp` failure — matches the added `empty-root` self-test case.

Also ran the full self-test suite (37 assertions) with `awk` shimmed to real mawk on the whole `PATH`: **37/37 pass, 0 failures.** Under gawk, likewise 37/37.

## Non-blocking items from the first pass — checked, agree with the author's disposition

- **`/build/` anchoring → `/build/(gst|ext|tests)/`.** Pulled both reference CI logs directly (`gh run view --job <id> --log`) and grepped every `(in .../build/...)` frame in both: every one of them is under `build/tests/...`, so the tightened pattern loses nothing that has actually occurred. It's also not overfit to what's been observed: the same workflow step that invokes this checker exports `GST_PLUGIN_PATH=$(pwd)/build/gst` and `NNSTREAMER_FILTERS/DECODERS/CONVERTERS=$(pwd)/build/ext/...` (`.github/workflows/ubuntu_clean_meson_build.yml:261-264`), so `build/gst` and `build/ext` are real, reachable object locations for this exact step going forward, not just `build/tests`. The three-way anchor is the right generalization, not a narrower one that happens to pass today's logs.
- **Secondary-stack test gaps.** `origin_stack_n` and `address_stack_n` were added and do pin the stated behavior (the `--track-origins` origin stack and the "Address ... alloc'd" stack never decide attribution). Read the awk state machine directly to confirm this isn't just an assertion that happens to pass: `pending` is set only on a new error header and cleared the instant the first non-wrapper `at/by` frame is recorded, so every line after that — including a second stack under the same `==N==` block — falls through to the bare `{ pending = 0 }` catch-all and is structurally inert. The tests match the implementation.
- **Executable bit.** Confirmed: `git ls-files -s tools/debugging/*.sh` shows `100644` for both files at this head. The claim that the working tree shows them executable only because `core.fileMode` is off on this checkout is consistent with `git config core.fileMode` returning `false` here. Nothing to change, agreed.
- **`README.md`.** Read it; it is in fact scoped entirely to `GST_DEBUG` / dot-graph dumps / `gst-debugger` and carries no index of sibling scripts (`valgrind_suppression` isn't listed either). Leaving it alone is reasonable and isn't a new gap this PR creates.

## One thing worth naming, not blocking

The new header comment ("nothing below the blamed frame counts... both are usually ours even when the error is not") is correct for what it documents — the *secondary* stack (origin / alloc'd-at) is properly ignored, and that's now tested. But there is a related, slightly different gap the new comment's wording doesn't quite cover, and it's real: the *primary* blamed frame itself can land inside an **unwrapped** libc function. Valgrind only substitutes its own wrapper for a handful of aliasing-sensitive functions (`memcpy`/`memmove`/`strcpy`/`memset`/...); plain comparison/scan functions like `strncmp`/`strlen` are not intercepted, so an invalid read inside glibc's own `strncmp` (as already exercised by the pre-existing `library` self-test case, via `strncmp (strcmp.S:172)`) is attributed to the library, full stop — even when the very next frame down names one of our source files as the caller that passed the bad pointer/size. I confirmed this is still true at `5cc4e99e` with a synthetic case (`myelement.c` calling `strncmp` with a bad argument, blamed frame `strncmp (strcmp.S:172)`): `0 error contexts from this repository, 1 from libraries`, exit 0.

This isn't a regression from this diff — the underlying awk logic (attribute to the first non-`vg_replace_` frame) is unchanged from `12b6bbb9`, and the `library` test case that exercises exactly this shape predates this incremental fix. I'm flagging it only because the task at hand was specifically to stress-test the "don't look below the blamed frame" rule for false negatives, and this is the concrete case where it bites: a bug that is *ours* by cause can still be invisible to this gate if valgrind's own attribution lands one frame short, in code we don't own. Worth a one-line addition to the header comment (or a code comment near `is_valgrind_own`) so a future reader doesn't read "0 errors of ours" as "no memory bug of ours" — but this is documentation polish, not something that should hold up this PR.

## CI at this head

Still running at review time: `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` (the job that actually exercises this gate) and `Static checks` (which runs the self-test) are both `pending`. `ubuntu-22.04-arm` valgrind, macOS, LLVM/Clang, Spell Check and DCO have already passed; nothing has failed so far. Given the self-test passes locally under both awks and shellcheck is clean on both scripts, no reason to expect either pending job to fail, but that's not yet confirmed by CI itself.

## Summary

Both blocking bugs from the previous round are fixed and independently re-verified — one against real mawk with the exact repro, the other against the actual CI logs the PR's own measurement table cites, where the before/after counts match the author's reply exactly. The three non-blocking items are adequately addressed or reasonably left alone. No new issues found in the diff itself; the one point raised above is a pre-existing, documented design trade-off surfaced by the task's specific ask to stress-test it, not something this diff introduced or should be blocked on.

myungjoo:
Addressed at `597e7497`, documentation only.

The finding is right, and I measured it rather than guessing at its size. Across both reference logs there is no error blamed on a library with a frame of ours beneath it: every library-blamed error sits entirely inside a library stack at the top. That is what memcheck's redirection of the string and memory functions buys — the wrappers are skipped, so the blame lands on our caller anyway, and the residue is the case where glibc's own copy runs unredirected, as it does inside `ld.so`.

So I left the rule alone and named the limitation where it lives, in the file header, together with that measurement. Changing the rule to look one frame further would gain nothing on today's evidence and would start charging us for bugs inside GLib that our code merely triggered — the over-attribution this rule exists to avoid. The self-test case that pins the behaviour is now described as what it is: an error blamed on a library passes *even when we called it*.

myungjoo:

Reviewed head `597e7497` (single commit, amend of `5cc4e99e`). Read both prior reviews (`issuecomment-5567382122`, `issuecomment-5567470786`) and both author replies (`issuecomment-5567434131`, `issuecomment-5567483826`) first, then independently re-verified the delta and the state of the PR as a whole rather than taking the replies at their word.

**Verdict: Approve. No remaining objection.**

## The delta since the last review is exactly what it's claimed to be

`git diff 5cc4e99e 597e7497` touches two files and is genuinely documentation-only: a 6-line addition to the file header of `check_valgrind_log.sh` naming the "library-blamed error whose caller is ours" limitation, and a one-line rewording of a `test_check_valgrind_log.sh` assertion description (`"an error reported in a library passes"` → `"an error blamed on a library passes even when we called it"`). No `awk` logic, no regex, no control flow changed. Confirmed by reading the diff directly — there is nothing to hide behind a "just docs" claim here.

## The empirical claim behind that decision holds up under independent re-derivation

The prior review's one non-blocking finding was that an error whose blamed frame lands inside an *unwrapped* libc function (`strncmp`, not `memcpy`/`strcpy`, which memcheck redirects) can leave a frame of ours invisible one level down. The author chose not to change the rule, citing a measurement of zero such cases across the two reference logs, and documented that limitation instead of coding around it.

I did not take that number on faith. I pulled both raw reference logs directly from GitHub Actions (`gh run view --job 101656130808 --log` for run `34094878222`, `--job 100962850967 --log` for run `33853964542`) and wrote a second, independent parser (Python, not awk) that walks each error block's *entire* primary frame stack — not just the frame the shell script's state machine stops at — and checks whether any library-blamed error has an our-source frame as its very next stack line. Result: **0 matches in both logs combined**, matching the author's claim exactly. This is a different implementation checking a superset of what the awk script itself looks at, so it's a real independent confirmation, not a rerun of the same code.

Given that, documenting the trade-off instead of extending the rule is the right call: walking one frame further into an unwrapped-libc blame would start attributing GLib-internal bugs we merely triggered back to us — exactly the over-attribution this design already avoids for the origin/`alloc'd` secondary stacks (per the file's existing comment and the round-2 test cases `origin_stack_n`/`address_stack_n`). Zero observed incidents plus a real cost to "fixing" it is a legitimate basis for a documented no-op.

## Independently reproduced: self-test, shellcheck, and live CI

- `tools/debugging/test_check_valgrind_log.sh`: 37/37 assertions pass under both real `gawk` and a `PATH`-shimmed real `mawk` (WSL), including the reworded case.
- `shellcheck` clean on both `check_valgrind_log.sh` and `test_check_valgrind_log.sh`.
- Live CI at this exact head: all 18 checks green (`gh pr checks 4936`). Pulled the actual Valgrind job log (`gh run view --job 101671732457 --log`, run `34099915322`, `headSha` confirmed `597e7497`) and found the checker's real summary line in it:
  `valgrind log check: 0 error contexts from this repository, 14 from libraries, 162 leak reports, 23 binaries examined`
  — this is the gate running for real against this PR's own build, not just the self-test, and it exits clean. The `Static checks` job's self-test step log likewise shows the reworded assertion (`PASS: an error blamed on a library passes even when we called it`) actually executing in CI.
- Re-ran the *current* checker against both raw reference logs myself (`--repo-root .`): run `34094878222` → `0 ours / 14 library / 167 leaks / 23 binaries`; run `33853964542` → `8 ours / 13 library / 158 leaks / 22 binaries`, with the same 8 `gst_tensor_meta_info_parse_header`/`_validate` findings the PR body describes. Matches the author's round-1 reply ("library count goes 12 → 14 and 11 → 13") exactly.

## One cosmetic, non-blocking gap

The PR **description's** own "Measured on real logs" table still shows the pre-regex-fix counts (`12` / `11` library errors) from the original commit `12b6bbb9`, not the `14` / `13` that the current script actually produces against the same two logs (and that the live CI log above confirms: `14`). The corrected numbers only exist in a review-reply comment, not in the top-level PR body a future reader would skim first. Doesn't affect code, tests, or the gate itself — worth a one-line edit to the description table before/after merge, not something to hold this up over.

(Also noted in passing, not an issue: leak-report counts are not perfectly reproducible run-to-run — 167 from the static reference log vs. 162 from this head's fresh live CI run, with identical ours/library/binaries counts. Expected, since leaks are counted/printed but explicitly not gated, and the PR already says so.)

## Summary

Both blocking findings from round 1 and the one non-blocking finding from round 2 are resolved — the last one by an honest, measured, and now-verified documentation choice rather than a code change, which is appropriate here. The delta since the last approval is exactly what the author says it is, self-tests are green under both awk implementations, shellcheck is clean, and the gate is confirmed working end-to-end against this PR's own live CI run. No new issues found in this round. Recommend merge; the only follow-up is the cosmetic PR-description table update noted above.

myungjoo:
All five addressed at `a34822d0`; three are code, two are documentation. Thank you for probing the demangled-signature and `still reachable` cases as well — those had no test of their own and I would not have known they held.

**1. Partial-log blind spot.** Fixed, and you are right that it was the same failure mode left half-closed. The two runs now write separate logs and are checked separately, so neither can satisfy `--require-output` on the other's behalf. Both checks run even when the first fails, so a bad run is not hidden behind an earlier one; the step's status is their combination.

**2. Base-name matching.** Documented in the header, with the direction it fails in. I would rather not reach for `--fullpath-after=` yet: it changes what every contributor reads in the log, for a collision that does not exist today. If one ever appears the header now says what to do about it.

**3. Build-object pattern.** Documented, including that the two directions differ. Anchoring it to `--repo-root` is what I would have done if the checker only ever ran inside the job, but a log recorded on the runner and checked on a laptop is how most of this PR's measurement was done, and that use is worth keeping. Worth noting the fail-open direction is narrower than it looks: our libraries are built with debug info, so their frames carry `file:line` and are caught by the name half; the object half only ever fires for the test binaries, which is what both reference logs show.

**4. Unrecognised kinds.** This one I could close, so I did. Any report whose first line matches no known kind is now printed as a `::warning::` naming the line, so drift surfaces in the checks UI instead of disappearing. I did not make it fail: a new memcheck announcement turning every job red is the worse failure, and this repository has been bitten by that shape of change twice already. I checked it is quiet on real data before keeping it — the only unclassified lines in either reference log were `Process terminating with default action of signal 6` and a `Thread N ...` header, both of which are now recognised as the announcements they are, leaving zero warnings on both logs.

**5. Cosmetic.** `key` is now a local in `record()`'s parameter list. The logs are written to `valgrind-tests.log` and `valgrind-llamacpp.log` and uploaded as an artifact when the job fails, using the same `if: failure()` shape as the existing `Meson_Testlog` upload.

Self-test is 41 assertions, passing under both mawk and gawk; shellcheck clean; both reference logs still give 0 and 8 with no warnings.

myungjoo:

**Independent re-review of `a34822d0`** (previous relayed review was of `597e7497`; a separate formal review already exists on this exact commit, `myungjoo-bot` review id 5131918557 — see "Overlap" note at the end).

## Verification of the five `597e7497` findings

All five were checked by re-deriving them from the diff and by execution, not just by reading the reply.

1. **Partial-log blind spot** — confirmed fixed. The two runs now `tee` to separate files (`valgrind-tests.log`, `valgrind-llamacpp.log`) and each is passed to `check_valgrind_log.sh --require-output` on its own line; neither can satisfy `--require-output` on the other's behalf.
2. **Basename matching** — documentation-only, as claimed. Header now states the assumption and its direction. No code change; fine to leave as documented risk.
3. **`/build/(gst|ext|tests)/` not anchored to repo root** — documentation-only, as claimed, with the stated rationale (checking a log recorded on another machine). Reasonable trade-off.
4. **Unrecognised error kinds vanishing** — confirmed fixed and confirmed quiet on real data. Ran `test_check_valgrind_log.sh` under **both mawk 1.3.4 and gawk** (WSL): 41/41 assertions pass on both, including the new `unknown_kind` case (`::warning::` printed, exit still 0) and `known_prelude_n` (no warning for a normal SIGABRT prelude). Ran `check_valgrind_log.sh` against both reference logs (the ones cited in the PR body/testing section) directly: **0 `::warning::` lines in either log**, matching the author's claim.
5. **`key` locality / log upload on failure** — confirmed. `key` is now declared in `record()`'s parameter list; the new "Upload the Valgrind logs if something went wrong" step mirrors the existing `Meson_Testlog` upload's `if: failure()` shape.

## `checked=0 / || checked=1 / exit ${checked}` under the step's actual shell

Confirmed via GitHub's own docs that omitting `shell:` (this step and this workflow set neither a step-level `shell:` nor `defaults.run.shell`) resolves to `bash -e {0}` — explicitly **not** `bash --noprofile --norc -eo pipefail {0}`, which is what `shell: bash` would give. I simulated the exact shape of the step under `bash -e` (no pipefail) in four scenarios:

- left side of the first `tee` pipeline exits nonzero (a test failure) → step still reaches and runs both `check_valgrind_log.sh` calls, and exits 0 if neither finds anything (matches the code comment "A test failure in this pass still does not fail the step").
- first `check_valgrind_log.sh` call fails → the second call still runs, and the step exits 1.
- only the second call fails → step still exits 1 (not swallowed by the first `|| checked=1` "succeeding").
- `check_valgrind_log.sh` itself dies with its usage-error exit code (2) → `checked=1`, step still fails (no fail-open on a checker crash).

All four match the intended design.

## Blocking: the new prelude-line rule does not clear `pending`, so an error can be misattributed to whatever frame allocated the block it touched

`tools/debugging/check_valgrind_log.sh:186-190` (the new rule added in this diff):

```awk
line ~ /^ *(Address 0x|Uninitialised value was created|Block was alloc|Location 0x|At least one of|Process terminating|Thread [0-9]+)/ {
  candidate = ""
  next
}
```

This clears `candidate` (right, for the unclassified-warning feature) but leaves `pending` untouched. On the commit before this diff, these same lines fell through to the catch-all at the bottom of the script, which does clear `pending`. Two paths can reach one of these prelude lines with `pending` still `1`: every main-stack frame is `vg_replace_*` (skipped via `next` without resetting `pending` in the `is_valgrind_own` branch), or a main-stack frame line doesn't end in `)` (the `else next` at what is now line ~206, e.g. an unsymbolized `??? ` frame with no containing object). In either case `pending` survives into the *allocation* stack that follows the `Address 0x...`/`Block was alloc'd` line, and the first non-`vg_replace_` frame there — the site that merely allocated the block, not the site of the reported error — gets recorded as the culprit.

Reproduced independently (both gawk and mawk agree):

```
==1== Command: ./tests/unittest_demo
==1== Invalid read of size 4
==1==    at 0x484D4C4: memcpy (vg_replace_strmem.c:1123)
==1==  Address 0x5204040 is 0 bytes after a block of size 8 alloc'd
==1==    at 0x2222: malloc (vg_replace_malloc.c:381)
==1==    by 0x3333: my_element_init (myelement.c:11)
```

→ `1 error contexts from this repository`, blamed on `my_element_init (myelement.c:11)`, exit 1 — even though the main stack never named a frame of ours (its only frame is valgrind's own wrapper). This is exactly the reasoning the header already excludes leaks for ("a leak is reported at its allocation ... deciding who was responsible for freeing it needs a rule this script does not have"); the same trap now applies to non-leak errors too, but as a bug rather than a documented exclusion. The direction matters for a gate: it can fail a job by pinning someone else's bug on our allocation site, on a stack shape that isn't outlandish (a shallow wrapper-only main stack, or an unsymbolized frame — both occur in practice with stripped/JIT/vsyscall code, though I did not find either shape in the two reference logs, which is presumably why this shipped invisibly).

Fix is one line — reset `pending` in the same rule:

```awk
line ~ /^ *(Address 0x|Uninitialised value was created|Block was alloc|Location 0x|At least one of|Process terminating|Thread [0-9]+)/ {
  pending = 0
  candidate = ""
  next
}
```

**Test gap:** neither the existing `known_prelude_n` case nor any other case in `test_check_valgrind_log.sh` exercises a main stack that is entirely `vg_replace_*` (or ends in an unparenthesized frame) followed by an attributable allocation-stack frame — that's precisely the shape needed to catch this regression, and its absence is why 41/41 still passed with the bug present. Please add it alongside the fix so this can't regress silently again.

## Non-blocking

- `Process terminating with default action of signal 11 (SIGSEGV)` is followed in real memcheck output by an `Access not within mapped region ...` (or similar) line *before* the stack, which the prelude regex does not recognise — that line falls through and would print a spurious `::warning::` on a genuine crash log. Low severity (a crash is already a bigger problem, and the log is now uploaded), but worth folding into the regex while touching this code.
- The new "Upload the Valgrind logs if something went wrong" step's `if: failure() && matrix.os == 'ubuntu-22.04'` lacks the `env.rebuild == '1'` guard the neighbouring `env.rebuild`-gated steps use — but this exactly mirrors the pre-existing `Meson_Testlog` upload step a few lines above it, which has the same gap. Not a regression this PR introduces as a new pattern; at most a pre-existing wart it inherits. `upload-artifact@v4` only warns (does not fail) when nothing matches `valgrind-*.log`, so the worst case is a benign annotation on an unrelated earlier failure.

## Scope / blast radius / other criteria

- Diff is exactly the 3 files described (workflow + checker + its self-test), ~71 lines; no other module touched; `packaging/run_unittests_binaries.sh` untouched as claimed.
- No architecture or public API change; no documentation update needed beyond the header comments already included in the diff.
- Test-case policy: the self-test is exercised as a required CI check (Static checks, same merge-ref guard as neighbours per the PR description) and does assert the summary counts, which is the right shape to catch a checker that silently stops parsing. The one gap is the misattribution case above.
- Verified against CI run 34117216361 (this PR's head, all 19 checks green): on `ubuntu-22.04`, `GTEST run with Valgrind on a case` succeeded and `Upload the Valgrind logs if something went wrong` was correctly `skipped`; on `ubuntu-22.04-arm`, both are `skipped` as expected since the step is gated to `ubuntu-22.04`.

## Overlap with the existing `myungjoo-bot` review on this same commit

Review id 5131918557 (already posted on `a34822d0`, state `CHANGES_REQUESTED`) independently found the same blocking `pending` issue with the same root cause and the same one-line fix, plus the same two non-blocking notes above. This comment corroborates that finding through an independent reproduction (including under mawk) and adds the concrete test-gap recommendation; it is not meant to duplicate that review's content, only to confirm it and record the verification trail for the merge decision.

## Verdict

**Changes requested**, for the misattribution bug above — it is real, reproducible on both mawk and gawk, and undercuts this PR's own stated soundness bar ("nothing below \[the wrapper\] frame counts... no error blamed on a library had a frame of ours beneath it" — this bug makes that guarantee false for a narrow but real stack shape). Everything else in the diff — the five previously-raised points, the shell/pipefail semantics, the warning mechanism, and the blast radius — checks out on independent verification.

myungjoo:
Confirmed and fixed at `444dfb6a`. This was mine, introduced by the previous round's own fix, and it is worth naming precisely: the prelude rule I added to stop unknown-kind warnings cleared `candidate` but not `pending`, and before that rule existed those lines fell through to the catch-all that cleared it. So the change that was meant to make unrecognised reports visible quietly created a path where an error with no frame to blame ran on into the *allocation* stack and charged us for a block we merely allocated. That is the exact reasoning the header gives for keeping leaks out of the gate, reappearing as a bug three lines away from where it is written down.

I reproduced it before touching anything, with your shape: main stack a single `vg_replace_strmem.c` frame, then `Address ... alloc'd`, then our `malloc` caller. It reported `1 error contexts from this repository` and exited 1. Now it reports 0 and exits 0.

The fix is the one line you named, plus two regression cases that pin the two ways an error can end without a frame to blame — every frame a valgrind wrapper, and a frame with no location at all (`??? `) — each followed by a prelude and an allocation stack containing one of our frames. Both assert the exit status *and* that our frame is not named in the output, so a future change that re-blames it fails loudly rather than passing with a different count.

While there I also took your non-blocking note on the crash prelude: `Access not within mapped region`, `Bad permissions for mapped region`, `General Protection Fault` and `Stack overflow in thread` are now recognised, so a genuine SIGSEGV log produces no spurious warning — and, with the fix above, cannot have its termination stack blamed on anyone either. The artifact upload step now carries `env.rebuild == '1'` as well; you are right that it was inherited from the neighbouring upload, but a step that cannot do anything useful should not run.

One residue I decided to leave: an error whose every frame is a valgrind wrapper is now counted as neither ours nor a library's rather than being attributed somewhere. Making it count would mean tracking whether a record ended without producing one, which is more parser state for a case that does not occur in either reference log — and more parser state is precisely what produced this bug.

45 assertions now, passing under both mawk and gawk; shellcheck clean; both reference logs unchanged at 0 and 8 with no warnings.

myungjoo:

**Re-review of `444dfb6a`** (previous relayed review, and a separate formal `myungjoo-bot` review, both on `a34822d0`; this reviews the one-commit delta the author's reply at `issuecomment-5571957509` describes).

## The `pending` fix, independently re-verified

`git diff a34822d0 444dfb6a -- tools/debugging/check_valgrind_log.sh` is exactly the claimed one-line functional change: the prelude rule (`Address 0x|...|Thread [0-9]+|...`) now does `pending = 0` before `candidate = ""`. Reproduced the exact false positive from the prior round, under both `mawk` and `gawk`:

```
==1== Command: ./tests/unittest_demo
==1== Invalid read of size 4
==1==    at 0x484D4C4: memcpy (vg_replace_strmem.c:1123)
==1==  Address 0x5204040 is 0 bytes after a block of size 8 alloc'd
==1==    at 0x2222: malloc (vg_replace_malloc.c:381)
==1==    by 0x3333: my_element_init (myelement.c:11)
```
→ now `0 error contexts from this repository, 0 from libraries`, exit 0, and `myelement.c` does not appear anywhere in the output. Also checked the second failure shape named in the task (a main stack ending in an unsymbolized `???` frame instead of an all-wrapper stack) with the same result:

```
==1== Use of uninitialised value of size 8
==1==    at 0x1111: ???
==1==  Uninitialised value was created by a heap allocation
==1==    at 0x2222: malloc (vg_replace_malloc.c:381)
==1==    by 0x3333: my_element_init (myelement.c:11)
```
→ `0/0`, exit 0, `myelement.c` absent. Both fixed correctly.

**Test gap closed, and closed well.** The two new cases (`wrapper_only_then_prelude_n`, `unlocatable_then_prelude_n`) pin exactly these two shapes, and each asserts *two* things: `expect_status 0` and `expect_not_in "myelement.c:11"`. That second assertion matters — a change that merely stopped exiting 1 (e.g. by silently dropping the error into the "library" bucket instead of "ours") would still pass a status-only test; asserting the frame's absence from the output is what actually pins "not attributed to anyone," which is the real invariant here.

## Self-test, shellcheck, both awks — all re-run directly, not taken on faith

- `test_check_valgrind_log.sh`: **45/45 assertions pass** under a real `mawk 1.3.4` and under `gawk` (both via WSL, `awk` shimmed on `PATH` to each), matching the count the author's reply and the "Static checks" CI job for this exact head both report.
- `shellcheck 0.8.0` on both `check_valgrind_log.sh` and `test_check_valgrind_log.sh`: clean, matches CI's own shellcheck step (`shellcheck: passed... There are 0 lines.`) for this head.
- Pulled the real Static checks job log for `444dfb6a` (`gh run view --job 101774568292 --log`): `check_valgrind_log.sh self-test passed.` is there for real, run under the CI runner's own `awk` (mawk).

## Reference logs, run through the current (`444dfb6a`) checker

Ran `check_valgrind_log.sh --repo-root <repo>` against the two logs cited in the PR body:

| log | ours | library | leaks | binaries | exit |
|---|---|---|---|---|---|
| `post4929.log` (current-main shape) | 0 | 14 | 167 | 23 | 0 |
| `jobpr4.log` (pre-#4929 shape) | 8 | 13 | 158 | 22 | 1 |

Both match the PR body's measurement table exactly, both produce **zero** `::warning::` lines under both `mawk` and `gawk`, and `--require-output` behaves correctly on each (rejects nothing wrongly, `jobpr4.log`'s 8 real errors are the exact `gst_tensor_meta_info_parse_header`/`_validate` (`nnstreamer_plugin_api_util_impl.c`) findings the PR body names). Also pulled this PR's own live Valgrind job log for `444dfb6a` (`gh run view --job 101774568556 --log`) and found both checker invocations running for real: `0 error contexts from this repository, 12 from libraries, 152 leak reports, 22 binaries examined` and `0 error contexts ..., 2 from libraries, 11 leak reports, 1 binaries examined` — matching the "22 binaries / 1 binary, both 0 of ours, no warnings" the task describes, and CI is 19/19 green at this head.

## New SIGSEGV-family prelude patterns, checked for real effect

`Access not within mapped region`, `Bad permissions for mapped region`, `General Protection Fault`, `Stack overflow in thread` were added to the same prelude rule. Verified two things, not just that the regex compiles:

1. They actually match real memcheck crash text and suppress the spurious `::warning::` the previous round's non-blocking note predicted (`Process terminating...` followed immediately by e.g. `Access not within mapped region at address 0x0` before the stack).
2. With the `pending = 0` fix in the same rule, a termination stack frame that names a real source file this repository holds (tested against `gst_tensor_meta_info_parse_header (nnstreamer_plugin_api_util_impl.c:99)`) is **not** attributed to anyone under any of the four new kinds — `0/0`, no warning. A crash is not turned into a gate failure, and not turned into a false "library" count either.

## The one thing worth naming (not blocking)

The residual choice the author called out — an error whose blamed frame is never determined (every frame a `vg_replace_*` wrapper, or one with no `(file:line)`/`(in obj)` to extract) is now counted as *neither* ours nor a library's, rather than falling into one bucket — is real and I checked it two ways:

- Confirmed by instrumenting a copy of the awk script to log every time an error's `pending` flag is cleared by the prelude rule or the catch-all rule while still `1` (i.e. an error that never reached `record()`). Both of the round's regression-test shapes trip this instrumentation as expected.
- Ran the same instrumented script against **both** real reference logs (`post4929.log`, `jobpr4.log`): **zero** such events in either. So this is a real, named gap in the model, but it does not fire on any error this repository's CI has actually produced across the two logs the PR measures against.

Given that, "leave it, since more parser state to track an unattributed error is exactly the kind of change that caused this bug" is a defensible trade-off for now — it mirrors the same reasoning already accepted for leaks and for the unwrapped-libc-frame gap from earlier rounds. Only suggestion, non-blocking: the file header documents those other two exclusions explicitly; this one currently lives only in the PR conversation, not in the script. A one-line addition next to the existing "Leaks are counted..." paragraph would keep the header a complete map of what the gate does not catch, for a future reader who hasn't read this thread. Not something to hold the PR on.

## Verdict

**Approve.** The blocking bug from the previous round (`pending` surviving into the allocation/origin stack and misattributing someone else's bug to whoever allocated the touched block) is fixed, independently reproduced as fixed under both `mawk` and `gawk`, and pinned by two new regression tests that assert both the exit status and the absence of the misattributed frame — so a regression here would fail loudly, not silently. The new SIGSEGV-family prelude entries do what they're for and don't reopen the same hole. Both reference logs reproduce the PR's own numbers exactly, live CI at this head is 19/19 green including the self-test as a required check, and shellcheck is clean. Nothing else in the diff (the `env.rebuild` guard added to the log-upload step) changes behavior for any other job or matrix leg. No blocking findings remain from this or the prior rounds' cumulative review.

myungjoo:
Taken, at the head above. The third exclusion now sits with the other two in the header: an error every one of whose frames is a wrapper has nothing to be blamed on and is counted as neither, and the reason not to decide otherwise is the same one the `pending` fix was about — carrying an error past the end of its own stack.

That is the only change since the approved head, and it is comment text in the shell part of the file, above the awk program. I checked it the way the last round taught me to: `bash -n`, shellcheck, the self-test under both awks, the false-positive reproduction from that round, and both reference logs. All unchanged — 45/45, 0 and 8, no warnings.

Thank you for the three rounds. Two of the four things you found were bugs I had written, and the second one I had written *while fixing the first*; neither would have survived to a maintainer, but neither would have been caught by the tests I thought were sufficient either.

PR #4937, opened 2026-09-07
[Common] Do not parse a tensor meta header past the buffer

myungjoo:
## What

`gst_tensor_buffer_from_config()` splits a single-memory flexible buffer by walking the meta header of each tensor. The loop parsed a header at every offset without checking that the remaining bytes can hold one:

```c
while (offset < total) {
  gst_tensor_meta_info_parse_header (&meta, map.data + offset);   /* reads up to 88 bytes, always */
  mem_size[num] = gst_tensor_meta_info_get_header_size (&meta);   /* 0 when the parse failed */
  mem_size[num] += gst_tensor_meta_info_get_data_size (&meta);    /* 0 as well */
  offset += mem_size[num];                                        /* offset does not advance */
  num++;
}
```

* A remainder shorter than the 128-byte v1 header is read past the end of the mapped memory. `parse_header()` takes no size: it reads `val[0..20]` (84 bytes, and `val[21]` too for a sparse format) before it validates anything.
* When the parse fails the sizes are 0, so `offset` stays put and the same out-of-bounds bytes are re-read until `num` reaches the memory limit. The buffer then comes back split into 15 zero-length memories plus the remainder.

Reachable from any flexible stream with a single memory that carries trailing bytes or a short payload, through `tensor_transform`, `tensor_demux`, `tensor_crop`, `tensor_sparse{enc,dec}`, `tensor_converter` and the collect-pad sync helper.

This is item **A2** of #4920.

## How

Check the remainder against the header size, and treat a header that does not parse the same way the loop already treats the memory limit: the remaining bytes become the last memory.

A header can also parse and still describe no bytes — `GST_TENSOR_META_IS_VALID()` accepts any version whose top bits match, while `gst_tensor_meta_info_get_header_size()` returns 0 for a version that is not V1, and a sparse format with `nnz == 0` has a data size of 0. That stalls the offset the same way, so the loop takes the same exit whenever the size comes out 0.

**A deliberate deviation from the issue text**, which suggested failing on an invalid header: `gst_tensor_buffer_from_config()` returning NULL more often would turn this over-read into a NULL dereference in the callers that do not check the return value — `tensor_crop` dereferences it immediately (`GST_BUFFER_TIMESTAMP`), and `tensor_demux` / `tensor_sparse*` pass it straight to `gst_tensor_buffer_get_count()`. Those are tracked as items C12/C15 of #4920, so this PR keeps the "hand the undecodable bytes on as one memory" contract the loop already had for the overflow case — now written down in the function's Doxygen — and leaves the stricter rejection to the items that fix the callers.

## Verification

* `unittest_common`: 159/159 pass; the 5 negative cases fail without the fix (the buffer comes back with 16 memories instead of 1 or 2).
* Valgrind (memcheck) over the new cases: **688 errors from 25 contexts** without the fix, all `Invalid read` in `gst_tensor_meta_info_parse_header()` called from `gst_tensor_buffer_from_config()`, and **0 errors** with it. The CI Valgrind GTEST step runs `unittest_common`, so a regression is caught there as well as by the counts.
* Full local suite: 22/23 meson tests pass; `unittest_filter_python3` fails on this box for the known NumPy 1.x/2.x mismatch, unrelated to the change.
* `gst-indent` leaves the changed hunks untouched, `clang-format` (16, the version CPP-Linter pins) is clean on the test file, and `doxygen-tag.sh` (advanced mode, with a negative control) passes.

## Other modules

The helper is shared, so the behaviour change was checked against every caller: for a well-formed flexible buffer nothing changes (the same headers parse to the same memories), and for a malformed one each caller now sees the undecodable bytes as a single memory instead of 15 empty ones plus a remainder — its own parse rejects them exactly as before. The pipeline test pins that on the real element path (`appsrc ! tensor_demux ! tensor_sink`): the valid tensor still reaches the sink.

### DIFF
--- a/gst/nnstreamer/nnstreamer_plugin_api_impl.c
+++ b/gst/nnstreamer/nnstreamer_plugin_api_impl.c
@@ -533,6 +533,7 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect,
  * @brief Configure gst-buffer with tensors information.
  * NNStreamer handles single memory chunk as single tensor.
  * If incoming buffer has invalid memories, separate it and generate new gst-buffer using tensors information.
+ * Bytes of a flexible buffer that do not describe a tensor are handed over as the last memory of the result.
  * Note that this function always takes the ownership of input buffer.
  * @param in input buffer
  * @param config tensors config structure
@@ -582,6 +583,9 @@ gst_tensor_buffer_from_config (GstBuffer * in, GstTensorsConfig * config)
         mem_size[num - 1] += gst_tensors_info_get_size (&config->info, i);
     }
   } else {
+    GstTensorMetaInfo meta;
+    gsize hsize;
+
     if (num > 1) {
       /* Suppose it is already configured. */
       out = gst_buffer_ref (in);
@@ -593,20 +597,27 @@ gst_tensor_buffer_from_config (GstBuffer * in, GstTensorsConfig * config)
       goto error;
     }
 
+    gst_tensor_meta_info_init (&meta);
+    hsize = gst_tensor_meta_info_get_header_size (&meta);
+
     num = 0;
     offset = 0;
     while (offset < total) {
-      GstTensorMetaInfo meta;
-      gpointer h = map.data + offset;
-
-      if (num >= NNS_TENSOR_MEMORY_MAX - 1) {
+      if (num >= NNS_TENSOR_MEMORY_MAX - 1 || (total - offset) < hsize ||
+          !gst_tensor_meta_info_parse_header (&meta, map.data + offset)) {
         /* Suppose remained memory may include extra tensors. */
         mem_size[num++] = total - offset;
         break;
       }
 
-      gst_tensor_meta_info_parse_header (&meta, h);
       mem_size[num] = gst_tensor_meta_info_get_header_size (&meta);
+
+      if (mem_size[num] == 0) {
+        /* Parsed header of an unknown version, cannot tell the tensors apart. */
+        mem_size[num++] = total - offset;
+        break;
+      }
+
       mem_size[num] += gst_tensor_meta_info_get_data_size (&meta);
 
       offset += mem_size[num];
--- a/gst/nnstreamer/tensor_common.h
+++ b/gst/nnstreamer/tensor_common.h
@@ -184,6 +184,7 @@ gst_tensor_time_sync_buffer_from_collectpad (GstCollectPads * collect, tensor_ti
  * @brief Configure gst-buffer with tensors information.
  * NNStreamer handles single memory chunk as single tensor.
  * If incoming buffer has invalid memories, separate it and generate new gst-buffer using tensors information.
+ * Bytes of a flexible buffer that do not describe a tensor are handed over as the last memory of the result.
  * Note that this function always takes the ownership of input buffer.
  * @param in input buffer
  * @param config tensors config structure
--- a/tests/common/unittest_common.cc
+++ b/tests/common/unittest_common.cc
@@ -2455,6 +2455,301 @@ TEST (commonUtil, createFlexTensorBuffer)
 }
 
 /**
+ * @brief Internal util function to build a flexible tensor stream of @a num
+ *        tensors, each carrying @a dsize bytes filled with its own index.
+ *        The size of the whole block is stored in @a total and the size of a
+ *        single tensor, its header included, in @a each. The caller owns the
+ *        returned block.
+ */
+static guint8 *
+build_flex_tensor_data (guint num, gsize dsize, gsize *total, gsize *each)
+{
+  GstTensorInfo info;
+  GstTensorMetaInfo meta;
+  guint8 *data;
+  gsize hsize, offset = 0;
+  guint i;
+
+  gst_tensor_info_init (&info);
+  info.type = _NNS_UINT8;
+  info.dimension[0] = (uint32_t) dsize;
+
+  gst_tensor_info_convert_to_meta (&info, &meta);
+  hsize = gst_tensor_meta_info_get_header_size (&meta);
+
+  *each = hsize + dsize;
+  *total = *each * num;
+  data = (guint8 *) g_malloc (*total);
+
+  for (i = 0; i < num; i++) {
+    gst_tensor_meta_info_update_header (&meta, data + offset);
+    memset (data + offset + hsize, (int) i, dsize);
+    offset += *each;
+  }
+
+  gst_tensor_info_free (&info);
+  return data;
+}
+
+/**
+ * @brief Internal util function to run gst_tensor_buffer_from_config () on a
+ *        flexible config with @a data of @a size bytes, which it takes over.
+ */
+static GstBuffer *
+flex_buffer_from_config (guint8 *data, gsize size)
+{
+  GstTensorsConfig config;
+  GstBuffer *out;
+
+  gst_tensors_config_init (&config);
+  config.info.format = _NNS_TENSOR_FORMAT_FLEXIBLE;
+  config.rate_n = config.rate_d = 1;
+
+  out = gst_tensor_buffer_from_config (gst_buffer_new_wrapped (data, size), &config);
+
+  gst_tensors_config_free (&config);
+  return out;
+}
+
+/**
+ * @brief Internal util function to get the size of the nth memory of @a buf.
+ */
+static gsize
+peek_memory_size (GstBuffer *buf, guint nth)
+{
+  return gst_memory_get_sizes (gst_buffer_peek_memory (buf, nth), NULL, NULL);
+}
+
+/**
+ * @brief Test tensor buffer util (more flexible tensors than the memory limit)
+ */
+TEST (commonUtil, createFlexTensorBufferOverMemoryMax)
+{
+  GstBuffer *out;
+  guint8 *data;
+  gsize total, each;
+  guint i;
+
+  data = build_flex_tensor_data (20U, 16U, &total, &each);
+  out = flex_buffer_from_config (data, total);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), (guint) NNS_TENSOR_MEMORY_MAX);
+  for (i = 0; i < NNS_TENSOR_MEMORY_MAX - 1U; i++)
+    EXPECT_EQ (peek_memory_size (out, i), each);
+
+  /* the tensors that do not fit are left in the last memory */
+  EXPECT_EQ (peek_memory_size (out, NNS_TENSOR_MEMORY_MAX - 1U),
+      each * (20U - (NNS_TENSOR_MEMORY_MAX - 1U)));
+  EXPECT_EQ (gst_buffer_get_size (out), total);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (flexible buffer shorter than a meta header)
+ */
+TEST (commonUtil, createFlexTensorBufferShortHeader_n)
+{
+  GstBuffer *out;
+
+  out = flex_buffer_from_config ((guint8 *) g_malloc0 (4), 4);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 1U);
+  EXPECT_EQ (gst_buffer_get_size (out), (gsize) 4);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (trailing bytes shorter than a meta header)
+ */
+TEST (commonUtil, createFlexTensorBufferShortTrailing_n)
+{
+  GstBuffer *out;
+  guint8 *data;
+  gsize total, each;
+
+  data = build_flex_tensor_data (1U, 16U, &total, &each);
+  data = (guint8 *) g_realloc (data, total + 4);
+  memset (data + total, 0, 4);
+
+  out = flex_buffer_from_config (data, total + 4);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 2U);
+  EXPECT_EQ (peek_memory_size (out, 0), each);
+  EXPECT_EQ (peek_memory_size (out, 1), (gsize) 4);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (flexible buffer without a valid meta header)
+ */
+TEST (commonUtil, createFlexTensorBufferInvalidHeader_n)
+{
+  GstBuffer *out;
+
+  out = flex_buffer_from_config ((guint8 *) g_malloc0 (256), 256);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 1U);
+  EXPECT_EQ (gst_buffer_get_size (out), (gsize) 256);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (trailing bytes without a valid meta header)
+ */
+TEST (commonUtil, createFlexTensorBufferInvalidTrailing_n)
+{
+  GstBuffer *out;
+  guint8 *data;
+  gsize total, each;
+
+  data = build_flex_tensor_data (1U, 16U, &total, &each);
+  data = (guint8 *) g_realloc (data, total + 256);
+  memset (data + total, 0, 256);
+
+  out = flex_buffer_from_config (data, total + 256);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 2U);
+  EXPECT_EQ (peek_memory_size (out, 0), each);
+  EXPECT_EQ (peek_memory_size (out, 1), (gsize) 256);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Internal util function to build a block of @a size bytes that starts
+ *        with a meta header of @a format and a version that passes the
+ *        validation of the meta but is not V1, so that the header size of the
+ *        tensor it describes is 0. The caller owns the returned block.
+ */
+static guint8 *
+build_unknown_version_data (gsize size, tensor_format format)
+{
+  guint8 *data = (guint8 *) g_malloc0 (size);
+  uint32_t *header = (uint32_t *) data;
+
+  header[0] = 0xfeedcced; /* magic */
+  header[1] = 0xDE002000; /* version */
+  header[2] = (uint32_t) _NNS_UINT8; /* type */
+  header[3] = 1U; /* dimension */
+  header[19] = (uint32_t) format;
+  header[20] = (uint32_t) _NNS_TENSOR; /* media type */
+
+  return data;
+}
+
+/**
+ * @brief Test tensor buffer util (meta header that describes no bytes)
+ */
+TEST (commonUtil, createFlexTensorBufferZeroSizeHeader_n)
+{
+  GstBuffer *out;
+
+  /* a sparse format without a non-zero element has a data size of 0 as well */
+  out = flex_buffer_from_config (
+      build_unknown_version_data (300, _NNS_TENSOR_FORMAT_SPARSE), 300);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 1U);
+  EXPECT_EQ (gst_buffer_get_size (out), (gsize) 300);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Test tensor buffer util (meta header of an unknown version)
+ */
+TEST (commonUtil, createFlexTensorBufferUnknownVersionHeader_n)
+{
+  GstBuffer *out;
+
+  /* the data size is known, but without a header size the tensors cannot be told apart */
+  out = flex_buffer_from_config (
+      build_unknown_version_data (300, _NNS_TENSOR_FORMAT_STATIC), 300);
+  ASSERT_TRUE (out != NULL);
+
+  EXPECT_EQ (gst_buffer_n_memory (out), 1U);
+  EXPECT_EQ (gst_buffer_get_size (out), (gsize) 300);
+
+  gst_buffer_unref (out);
+}
+
+/**
+ * @brief Callback for the new-data signal, keeping the size of the buffer.
+ */
+static void
+flex_trailing_new_data_cb (GstElement *element, GstBuffer *buffer, gpointer user_data)
+{
+  gsize *size = (gsize *) user_data;
+
+  (void) element;
+
+  EXPECT_EQ (gst_tensor_buffer_get_count (buffer), 1U);
+  *size = gst_buffer_get_size (buffer);
+}
+
+/**
+ * @brief Test tensor buffer util (trailing bytes of a flexible buffer in a pipeline)
+ */
+TEST (commonUtil, createFlexTensorBufferShortTrailingPipeline_n)
+{
+  GstElement *pipeline, *appsrc, *sinkx;
+  GstBuffer *buf;
+  guint8 *data;
+  gsize total, each;
+  guint received = 0U;
+  gsize received_size = 0;
+  GstFlowReturn flow_ret;
+
+  pipeline = gst_parse_launch ("appsrc name=appsrc caps=other/tensors,format=flexible,framerate=(fraction)0/1 ! "
+                               "tensor_demux ! tensor_sink name=sinkx async=false sync=false",
+      NULL);
+  ASSERT_TRUE (pipeline != nullptr);
+
+  appsrc = gst_bin_get_by_name (GST_BIN (pipeline), "appsrc");
+  ASSERT_TRUE (appsrc != nullptr);
+  sinkx = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_TRUE (sinkx != nullptr);
+
+  g_signal_connect (sinkx, "new-data", (GCallback) flex_trailing_new_data_cb, &received_size);
+  g_signal_connect (sinkx, "new-data", (GCallback) count_output, &received);
+
+  data = build_flex_tensor_data (1U, 16U, &total, &each);
+  data = (guint8 *) g_realloc (data, total + 4);
+  memset (data + total, 0, 4);
+  buf = gst_buffer_new_wrapped (data, total + 4);
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  /* gst_app_src_push_buffer () takes ownership of buf */
+  flow_ret = gst_app_src_push_buffer (GST_APP_SRC (appsrc), buf);
+  EXPECT_EQ (flow_ret, GST_FLOW_OK);
+  EXPECT_TRUE (wait_pipeline_process_buffers (&received, 1U, TEST_TIMEOUT_LIMIT_MS));
+
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+
+  /**
+   * The trailing bytes are not a tensor, so the demux ends the flow with an
+   * error once it reaches them. The valid tensor is pushed before that, which
+   * is what this case asserts.
+   */
+  EXPECT_EQ (received, 1U);
+  EXPECT_EQ (received_size, each);
+
+  gst_object_unref (appsrc);
+  gst_object_unref (sinkx);
+  gst_object_unref (pipeline);
+}
+
+/**
  * @brief Test tensor buffer util (invalid config)
  */
 TEST (commonUtil, createTensorBufferInvalidConfig_n)


myungjoo:

## Summary

Reviewed the diff (`gst/nnstreamer/nnstreamer_plugin_api_impl.c` + `tests/common/unittest_common.cc`, +239/-3 across 2 files) against the stated goal: fix the out-of-bounds header read in `gst_tensor_buffer_from_config()` (issue #4920 item A2) without turning it into a NULL-deref regression in callers that don't check the return value.

## Correctness of the fix

Traced the new loop condition by hand against all four new unit tests plus the pre-existing `createFlexTensorBuffer` case:

```c
gst_tensor_meta_info_init (&meta);
hsize = gst_tensor_meta_info_get_header_size (&meta);

if (num >= NNS_TENSOR_MEMORY_MAX - 1 || (total - offset) < hsize ||
    !gst_tensor_meta_info_parse_header (&meta, map.data + offset)) {
  mem_size[num++] = total - offset;
  break;
}
```

- `gst_tensor_meta_info_get_header_size()` on a freshly-`init`ed meta returns the fixed 128-byte V1 size purely from `magic`/`version` (set by `init`), independent of the not-yet-parsed content — so `hsize` is a valid, content-independent upper bound to check before touching `map.data + offset`. This mirrors the exact same guard (`msize < hsize` before mapping/parsing) that already exists a few dozen lines below in `gst_tensor_meta_info_parse_memory()` (same file, ~line 1629), so the fix follows an established idiom rather than inventing a new one.
- `offset <= total` is invariant from the `while (offset < total)` condition, so `total - offset` (both `gsize`, unsigned) never underflows.
- Short-circuit order is correct: the `NNS_TENSOR_MEMORY_MAX - 1` bound is still checked first, so `mem_size[num]` (a fixed `NNS_TENSOR_MEMORY_MAX`-sized array) is never written out of bounds — unchanged from before.
- Manually replayed all 5 new/changed test scenarios (over-memory-limit split at 20 tensors/16 limit, 4-byte buffer shorter than a header, well-formed tensor + 4 trailing bytes, all-zero 256 bytes with invalid magic, well-formed tensor + 256 trailing invalid bytes) against the new code path and the resulting memory counts/sizes match what the tests assert exactly.

## Scope and blast radius

- The change is minimal (6 lines in one loop) and touches only the one shared helper plus its test file — no unrelated files touched.
- `gst_tensor_buffer_from_config()` is declared in `gst/nnstreamer/tensor_common.h`, which is not an installed/public header (no `install: true`/`install_headers` reference for it), so this is an internal-only behavior fix — no public API or architecture change, and correctly no documentation update was needed or expected.
- Verified the full caller list: `gsttensor_transform.c`, `gsttensor_demux.c`, `gsttensor_converter.c`, `gsttensor_sparseenc.c`, `gsttensor_sparsedec.c`, `gsttensor_crop.c`, matching what the PR description claims it checked. The behavior-preservation argument (well-formed input parses identically; malformed input now yields one "leftover" memory instead of up to 15 empty ones + a remainder, which each caller's own downstream validation already had to tolerate) holds up — this is a genuine narrowing of an OOB read, not a new failure mode for well-formed streams.
- Consciously scoped: the PR explicystly declines to make the function return NULL on unparseable input (which the issue text originally suggested) because two known callers (`tensor_crop`, `tensor_demux`/`tensor_sparse*`) don't check the return value and would turn an OOB read into a NULL-deref crash. That is deferred to tracked issue items C12/C15. This is a reasonable, well-justified scope boundary for a fix targeting only item A2.

## Test coverage

- Good layered coverage: unit-level (`createFlexTensorBufferOverMemoryMax`, `...ShortHeader_n`, `...ShortTrailing_n`, `...InvalidHeader_n`, `...InvalidTrailing_n`) plus one pipeline-level regression test (`createFlexTensorBufferShortTrailingPipeline_n`, `appsrc ! tensor_demux ! tensor_sink`) that exercises a real caller end-to-end.
- Test names follow the existing `_n` suffix convention already used throughout this file for edge/degrade-gracefully cases (e.g. `tensorBufferForgedExtraCount_n`), and all four new static helpers carry Doxygen comments, matching repo convention.
- `flex_trailing_new_data_cb` is connected before `count_output` on the `new-data` signal, which is the ordering this test file's own `count_output` doc-comment requires (checking handler must run before the counter that lets the waiter tear the pipeline down) — correctly followed.
- Per the PR description, the CI job "Build with meson and test with Valgrind in Ubuntu" (`.github/workflows/ubuntu_clean_meson_build.yml`, triggered on `pull_request` to `main`) runs the full meson/gtest suite including `tests/common` under Valgrind, so both the plain assertions and the underlying invalid-read would be caught by CI if this fix were ever reverted or regressed — this is a real per-PR gate, not just a nightly/opt-in job.

## Findings

No correctness, regression, scope, or test-coverage issues found. Two purely cosmetic/nit observations, neither blocking:

- **nit**: `gst_tensor_meta_info_init (&meta); hsize = gst_tensor_meta_info_get_header_size (&meta);` runs on every loop iteration, but for the only version currently defined (V1) this always yields the same constant (128). Could be hoisted outside the `while` loop for a marginal cycle saving; not worth blocking on, and keeping it inside the loop is arguably more future-proof if a V2 header with a different fixed size is ever added.
- **nit**: `gst_tensor_meta_info_parse_header()` itself calls `gst_tensor_meta_info_init(meta)` internally, so the `gst_tensor_meta_info_init (&meta)` added just above it is redundant on the path where parsing is attempted (it's only load-bearing on the two short-circuit paths that skip the parse call). Harmless, just a small duplication.

## Verdict

**Merge as-is.** The fix correctly closes the out-of-bounds read described in issue #4920 item A2, follows an existing guard idiom already present in the same file, is minimally scoped to the one affected function, preserves behavior for well-formed streams, deliberately avoids introducing a NULL-deref regression in callers tracked separately (C12/C15), and ships with thorough unit + pipeline-level regression tests that CI will enforce on every future PR via the Valgrind-gated meson test job.

myungjoo:

Both nits are addressed in the force-pushed `891a115a`: `gst_tensor_meta_info_init ()` and `gst_tensor_meta_info_get_header_size ()` are hoisted out of the loop, so the header size is derived once per call instead of once per tensor and the redundant re-init on the parse path is gone. `meta` is reused across iterations, which is safe because `gst_tensor_meta_info_parse_header ()` re-initialises it itself.

Re-verified after the change: `unittest_common` 158/158, Valgrind memcheck over the new cases 0 errors (688 errors from 25 contexts without the fix commit), `gst-indent` leaves the hunk untouched.

myungjoo:

This is a **re-review** of the force-pushed update (`891a115a`) that addressed the two nits from the previous automated review round (hoisting `gst_tensor_meta_info_init()` / `gst_tensor_meta_info_get_header_size()` out of the loop, and reusing the `meta` variable across iterations). Focus of this pass is on whether that update is itself correct, plus a full re-check of the rest of the checklist.

## Focus: is the hoisted `hsize` valid for every iteration, and is `meta` reuse safe?

Traced both through `gst_tensor_meta_info_init()`, `_get_header_size()`, and `_parse_header()` in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`:

- `gst_tensor_meta_info_init()` unconditionally sets `magic = GST_TENSOR_META_MAGIC` and `version = GST_TENSOR_META_VERSION` (the only version currently defined, V1). `gst_tensor_meta_info_get_header_size()` returns a fixed `128` for any valid V1 meta, independent of any field that gets filled in later by parsing. So `hsize` computed once from a freshly-`init`ed `meta`, before the loop starts, is a content-independent constant for every iteration — hoisting it is correct. (This is the same value `gst_tensor_meta_info_get_header_size (&meta)` still recomputes redundantly-but-harmlessly right after each successful parse a few lines later, so the two never disagree.)
- `gst_tensor_meta_info_parse_header()` calls `gst_tensor_meta_info_init (meta)` as its very first statement — `memset`-zeroing the whole struct — before filling fields from the raw header bytes. That means reusing the same `meta` across loop iterations cannot leak state from a previous (possibly sparse, possibly failed) iteration into the next: e.g. `sparse_info.nnz` is zeroed before the conditional `format == SPARSE` branch re-sets it, so a sparse header followed by a non-sparse one does not retain a stale `nnz`. Reuse is safe.
- Manually replayed all 5 new/changed test scenarios (20-tensors-over-limit, 4-byte short buffer, valid tensor + 4 short trailing bytes, 256 zero/invalid-magic bytes, valid tensor + 256 invalid trailing bytes) against the post-force-push code; each still produces the exact memory count/sizes the assertions expect. `total - offset` cannot underflow (`gsize`, loop invariant `offset < total`), and the `NNS_TENSOR_MEMORY_MAX - 1` bound is still checked first in the `||` chain, so `mem_size[]` (sized `NNS_TENSOR_MEMORY_MAX`, confirmed `16` in `tensor_typedef.h`) is never written out of bounds.

Both nits are genuinely resolved, not just cosmetically: the redundant per-iteration `init()` call is gone (now called once), and the reuse is provably safe rather than merely "happens to work."

## Full checklist re-check

- **Solves the stated problem**: yes — the out-of-bounds header read (issue #4920 item A2) is closed by the `(total - offset) < hsize` guard before touching `map.data + offset`, and the "parse failed" case is now folded into the same "leftover becomes the last memory" path the overflow case already used, instead of spinning with `offset` stuck.
- **Regression risk / other modules**: none found. For well-formed input every branch produces byte-identical splits to before (verified against the pre-existing `createFlexTensorBuffer` positive-path expectations). For malformed input, the change only narrows what used to be an OOB read plus 15 empty memories into a single leftover memory — every caller listed in the PR body (`tensor_transform`, `tensor_demux`, `tensor_crop`, `tensor_sparseenc/dec`, `tensor_converter`) already has to tolerate `gst_tensor_buffer_from_config()` returning a buffer whose tensor count doesn't match expectations, and none of them are touched by this diff.
- **Scope / size**: minimal — 8/-5 lines in the one function, plus test-only additions. No unrelated files touched.
- **Test coverage**: thorough and well-layered — unit tests for the memory-limit overflow (regression-pins pre-existing behavior), short buffer, short trailing remainder, invalid header, invalid trailing remainder, and one pipeline-level test (`appsrc ! tensor_demux ! tensor_sink`) that exercises a real caller end-to-end and asserts the valid tensor still reaches the sink. New static helpers all carry Doxygen comments per repo convention, and `_n`-suffixed negative tests follow the file's existing naming convention.
- **Future-regression detection**: the two short/invalid-header unit tests reproduce the actual OOB read (not just a wrong-answer assertion), so a future change that reintroduces the missing bound check would show up as a Valgrind invalid-read, not only a value mismatch — a stronger signal than a plain assertion failure.
- **CI enforcement**: `tests/common` (including `unittest_common`) runs under the meson/GTest job, which per the PR body is Valgrind-gated on `pull_request` to `main` — a regression would fail CI on the PR itself, not just nightly.
- **Docs/architecture**: `gst_tensor_buffer_from_config()` is an internal helper (declared in the non-installed `tensor_common.h`), no public API or architecture change, so no documentation update is required or missing.

## Verdict

**Merge as-is.** Both nits from the previous round are correctly and safely addressed — the hoisted `hsize` is a valid constant for every iteration under the current (only) header version, and reusing `meta` across iterations is safe because `gst_tensor_meta_info_parse_header()` fully re-initializes it on every call. No new issues found in this re-review.

myungjoo:

This is the **final re-review**, focused on the force-pushed formatting commit that landed after the second "merge as-is" round (current head `c05ddd23`), plus a full re-check of the checklist now that all 18 CI jobs are green.

## Is the last force-push purely a formatting change?

The commit that the second review approved was head `c5612e74` (fetched directly by SHA from the fork — it's no longer reachable from any ref after the force-push, but GitHub still serves it by SHA). Diffing it against the current head `c05ddd23`:

```
tests/common/unittest_common.cc | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)
```

The only change is a single `gst_parse_launch()` call in `createFlexTensorBufferShortTrailingPipeline_n` being re-wrapped from 3 lines to 2:

```c
-  pipeline = gst_parse_launch (
-      "appsrc name=appsrc caps=other/tensors,format=flexible,framerate=(fraction)0/1 ! "
-      "tensor_demux ! tensor_sink name=sinkx async=false sync=false",
+  pipeline = gst_parse_launch ("appsrc name=appsrc caps=other/tensors,format=flexible,framerate=(fraction)0/1 ! "
+                               "tensor_demux ! tensor_sink name=sinkx async=false sync=false",
       NULL);
```

Same string literals, same call, same arguments — only the line-break/continuation-indent point moved (a known clang-format-16-vs-18 divergence on wrapped call arguments). No other file, hunk, or line changed between the two commits; `891a115a` (the actual fix in `nnstreamer_plugin_api_impl.c`) is untouched.

Ran `clang-format 16.0.6` locally against the current `tests/common/unittest_common.cc`: it produced **zero further diff** (byte-identical output), confirming the file is now clean under the version CI's "Static checks" job pins, and that the force-push is complete and idempotent, not a partial reformat.

**Conclusion: the force-push is a pure, no-op-on-semantics formatting fix.** Nothing here changes the earlier reviews' conclusions.

## Re-check: does the PR still correctly fix #4920 item A2?

Re-traced the loop in `gst_tensor_buffer_from_config()` (`gst/nnstreamer/nnstreamer_plugin_api_impl.c`, unchanged since `891a115a`) by hand against all 5 new/changed unit tests plus the pipeline test:

- `hsize` is computed once from a freshly-`init`ed `meta` before the loop (content-independent constant for the only header version, V1/128 bytes).
- `(total - offset) < hsize` is checked *before* `gst_tensor_meta_info_parse_header()` is called, so the short-buffer case (`ShortHeader_n`, `ShortTrailing_n`) never touches memory past `map.data + total` — this is the actual OOB read fix.
- A header that parses but is invalid (`InvalidHeader_n`, `InvalidTrailing_n`) now falls into the same "treat the rest as one memory" branch the pre-existing memory-limit overflow case already used, so `offset` can no longer get stuck and spin over the same bytes.
- `NNS_TENSOR_MEMORY_MAX - 1` is still checked first in the `||` chain (confirmed `NNS_TENSOR_MEMORY_MAX == 16` in `tensor_typedef.h`), so `mem_size[]` is never written out of bounds — `createFlexTensorBufferOverMemoryMax` (20 tensors, 16-slot limit) pins the pre-existing overflow split unchanged.
- Manually replayed each test's byte layout against the guard and got exactly the counts/sizes each test asserts.

This matches the two prior reviews' independent traces and my own re-derivation agrees. The fix genuinely closes the flexible-header OOB read described in #4920 A2.

## Rest of the checklist

- **Scope/size**: still minimal — 2 files only (`gst/nnstreamer/nnstreamer_plugin_api_impl.c`, `tests/common/unittest_common.cc`), 6-line functional change plus test-only additions. No unrelated files touched, confirmed via `gh pr diff`.
- **Other-module regression**: none. The helper's callers (`tensor_transform`, `tensor_demux`, `tensor_crop`, `tensor_sparseenc/dec`, `tensor_converter`) are unmodified; well-formed input still parses identically; malformed input now yields one leftover memory instead of 15 empty ones plus a remainder, which is a narrowing, not a new failure mode.
- **Performance**: negligible — one extra size comparison per iteration, no added allocation or I/O.
- **Test completeness**: 4 unit-level negative/positive cases plus one pipeline-level test through a real caller (`appsrc ! tensor_demux ! tensor_sink`); the two short-buffer cases reproduce the actual invalid-read (Valgrind-visible), not just a wrong-value assertion, so a future regression would be caught even if an assertion were accidentally weakened.
- **CI regression-catching**: "Build with meson and test with Valgrind in Ubuntu" runs `unittest_common` under Valgrind on every PR to `main` — a reintroduced OOB read fails CI directly, not only on a nightly job. All 18 jobs are currently green, including the previously-failing "Static checks" and "CPP-Linter".
- **Docs**: `gst_tensor_buffer_from_config()` is an internal helper declared in the non-installed `tensor_common.h` — no public API/architecture surface changed, so no documentation update is needed.

## Findings

None. No functional, regression, scope, or test-coverage issues in this round; the only change since the last "merge as-is" review is the verified-pure formatting fix.

## Verdict

**Merge as-is.**

myungjoo:

All three items are addressed in `3424b829`. The branch is now a single commit, as the two previous ones were squashed together.

**1. (Low) parse succeeds with `mem_size == 0` — confirmed and fixed.**

Reproduced before fixing: a header of `magic = 0xfeedcced`, `version = 0xDE002000`, `format = _NNS_TENSOR_FORMAT_SPARSE`, `nnz = 0` gives `parse = 1, header_size = 0, data_size = 0` on the previous head, and the buffer came back with 16 memories — exactly the shape the PR set out to remove. `GST_TENSOR_META_IS_V1()` requires `(v & 0x00FFF000) & 0xDE001000`, which `0xDE002000` fails while `GST_TENSOR_META_VERSION_VALID()` accepts it, so the analysis is right.

The loop now takes the leftover exit whenever the computed size is 0. This cannot reject a well-formed tensor: a V1 header always has `header_size == 128`, so `mem_size` is non-zero even for a sparse tensor with `nnz == 0`.

`createFlexTensorBufferZeroSizeHeader_n` forges that header and expects one memory; it is the only case that fails when the new guard is removed (16 vs 1), so it gates exactly this branch. The non-V1 + non-sparse misparse is left alone as pre-existing behaviour, as suggested.

**2. (Info) pipeline test and the bus.**

Documented rather than asserted. The test now records that the demux ends the flow with an error once it reaches the trailing bytes, and that the case asserts the valid tensor is pushed before that. An assertion on the bus would pin `tensor_demux`'s error policy inside a test of the common helper, which seemed like the wrong place to fix it; if you would rather have the assertion, it is a one-line change.

**3. (Nit)**

* The commit message now says "reads up to 88 bytes unconditionally (84, plus the sparse element count)", and the PR body spells out `val[0..20]` plus `val[21]` for the sparse format.
* The malformed-input contract is now in the Doxygen of `gst_tensor_buffer_from_config()`: "Bytes of a flexible buffer that do not describe a tensor are handed over as the last memory of the result."

Re-verified after the change: `unittest_common` 159/159, Valgrind memcheck over the new cases 0 errors, `gst-indent` leaves the hunks untouched, `clang-format` 16 clean, `doxygen-tag.sh` passes with a negative control.

myungjoo:

This is a **re-review** of the squash + force-push (`3424b829`) that responded to the three items raised in the earlier review (id 5131867872, 2026-09-07T12:14:55Z): (1) `mem_size[num] == 0` after a successful-but-degenerate parse, (2) the pipeline test not checking the bus, (3) the "84 bytes" wording and a missing Doxygen contract note. I did not rely on the summary comment describing these changes — I fetched both the old and new heads directly and diffed them.

## 1. `mem_size[num] == 0` guard — fixed correctly

Fetched the previous head `c05ddd23` and the new squashed head `3424b829` by full SHA (`git fetch fork <sha>`, since the branch was force-pushed) and diffed them directly:

```c
       mem_size[num] = gst_tensor_meta_info_get_header_size (&meta);
       mem_size[num] += gst_tensor_meta_info_get_data_size (&meta);
 
+      if (mem_size[num] == 0) {
+        /* Parsed header of an unknown version, cannot tell the tensors apart. */
+        mem_size[num++] = total - offset;
+        break;
+      }
+
       offset += mem_size[num];
       num++;
```

This is exactly the shape the earlier review suggested. Traced it by hand:

- The branch is only reachable when `offset < total` (loop invariant) and the header parsed successfully, so `total - offset > 0` there — `mem_size[num]` can't be reassigned to 0 inside this branch, so there's no infinite loop or repeated re-entry into it.
- It cannot false-positive on a well-formed tensor: `hsize` for a V1 header is always the fixed 128, and `gst_tensor_meta_info_get_header_size()` only returns 0 when `GST_TENSOR_META_IS_V1()` is false (non-V1 version). A legitimate V1 sparse tensor with `nnz == 0` still has `mem_size = 128 + 0 = 128`, so it isn't caught by this guard — it's routed through the normal per-tensor path. Only truly non-V1 headers hit the new branch.
- `createFlexTensorBufferZeroSizeHeader_n` forges `magic = 0xfeedcced`, `version = 0xDE002000` (passes `GST_TENSOR_META_VERSION_VALID`, fails `GST_TENSOR_META_IS_V1`, since `0xDE002000 & 0x00FFF000 = 0x2000` shares no bit with `GST_TENSOR_META_MAKE_VERSION(1,0) = 0xDE001000`'s `0x1000`), `type = _NNS_UINT8`, `dimension[0] = 1` (valid rank-1), `format = SPARSE`, `media_type = _NNS_TENSOR`, `nnz = 0`. I checked the field offsets (`val[19]`=format, `val[20]`=media_type, `val[21]`=nnz for sparse) against `gst_tensor_meta_info_parse_header()` — they match, so `parse_header()` genuinely returns TRUE (passes `gst_tensor_meta_info_validate()`) and `header_size = 0`, `data_size = 0`. Confirmed by hand that this test fails (16 memories) with the guard removed and passes (1 memory, full 300 bytes) with it — it gates exactly the branch it's meant to.

This closes item 1 correctly, without introducing a new false-positive path.

## 2. Pipeline test / bus — addressed as documentation, not an assertion

The squashed commit adds a comment above the assertions in `createFlexTensorBufferShortTrailingPipeline_n` recording that `tensor_demux` ends the flow with `GST_FLOW_ERROR` once it hits the trailing bytes, and that the case only asserts the valid tensor is pushed before that happens. No bus-state assertion was added. This was explicitly offered as optional in the original review ("if you would rather have the assertion, it is a one-line change") and the author's stated reasoning — that pinning `tensor_demux`'s error-vs-drop policy inside a test of the shared buffer-splitting helper is the wrong place for it — is reasonable. This was an Info-level item, not a blocker; documenting the current expectation is an adequate response even without the extra assertion.

## 3. Wording and Doxygen contract — both fixed

- Commit message now reads "reads up to 88 bytes unconditionally (84, plus the sparse element count)" instead of the flat "84 bytes" — matches `parse_header()`'s actual worst case (`val[0..20]` = 84 bytes, plus `val[21]` for the sparse branch = 88).
- `gst_tensor_buffer_from_config()`'s Doxygen now carries: "Bytes of a flexible buffer that do not describe a tensor are handed over as the last memory of the result." — confirmed present in the diff, directly above the existing `@param`/`@return` block, matching repo convention for other functions in the same file.

All three items are resolved.

## Squash integrity check

Diffed old head `c05ddd23` against new head `3424b829` across the **entire tree**, not just the two files the summary comment mentioned:

```
gst/nnstreamer/nnstreamer_plugin_api_impl.c |  7 ++++++
tests/common/unittest_common.cc             | 39 +++++++++++++++++++++++++++++
2 files changed, 46 insertions(+)
```

No other files differ — nothing was dropped or altered outside the intended changes. Both `c05ddd23` and `3424b829` share the same merge-base with current `main` (`99d0a5c5`), so the squash didn't silently rebase onto stale history either. The single squashed commit carries one `Signed-off-by:` line (no duplicates) and a `Co-Authored-By:` trailer, matching DCO requirements; the `DCO` check on the PR passes.

## Standard checklist

- **Solves #4920 item A2**: yes, confirmed by re-deriving the guard logic against all 6 unit tests (including the new one) plus the pipeline test.
- **Regression / other modules**: none. Well-formed flexible buffers parse identically (verified the V1/non-zero-header-size argument above); malformed buffers now yield either the pre-existing overflow split or a single leftover memory, both of which callers (`tensor_transform`, `tensor_demux`, `tensor_crop`, `tensor_sparseenc/dec`, `tensor_converter`) already had to tolerate.
- **Scope/size**: minimal — 46 lines total across 2 files, no unrelated files touched.
- **Test completeness / future-regression detection**: good layered coverage (6 unit-level cases including the new zero-size-header one, 1 pipeline-level case through a real caller). The two short-buffer cases reproduce the actual invalid-read, so Valgrind (which the CI Valgrind/meson job runs on `unittest_common`) would also catch a reintroduction, not just a value-mismatch assertion.
- **CI**: all checks were re-triggered by the force-push and are currently `pending` at review time (only `DCO` has completed and passes) — this needs to go green before merging, but there's no reason to expect a different result than the prior green run since the diff versus the previously-fully-green head is exactly the 46-line, well-scoped change reviewed above.
- **Docs/API**: `gst_tensor_buffer_from_config()` is an internal helper (declared in the non-installed `tensor_common.h`) — no public API/architecture surface changed, so no external documentation update is needed; the new Doxygen line is the right and sufficient documentation update.

## Verdict

All three items from the prior review round are genuinely and correctly addressed, and the squash preserved the reviewed content exactly (verified by direct SHA diff, not by trusting the summary comment). No new issues found.

**Merge as-is**, once the currently-pending CI checks finish green.

myungjoo:

Thanks for the approval. Both follow-ups are small enough to close here rather than leave for a later PR, so they are in `ca083897` (still a single commit).

**1. Guard on the header size instead of the sum.** Done as suggested:

```c
      mem_size[num] = gst_tensor_meta_info_get_header_size (&meta);

      if (mem_size[num] == 0) {
        /* Parsed header of an unknown version, cannot tell the tensors apart. */
        mem_size[num++] = total - offset;
        break;
      }

      mem_size[num] += gst_tensor_meta_info_get_data_size (&meta);
```

This closes the non-V1 + non-sparse case too. `createFlexTensorBufferUnknownVersionHeader_n` covers it: an unknown-version header with a static format and a one-byte tensor used to advance the offset by the data size alone and come back as 2 memories; it is the only case that fails when the check is moved back after the sum (2 vs 1), so it gates exactly this ordering. The forged-header setup is now shared by that case and the zero-size one through a small helper.

**2. Doxygen copies.** The line is now in `gst/nnstreamer/tensor_common.h` as well, so the declaration the callers read carries the same contract as the definition.

Re-verified: `unittest_common` 160/160, Valgrind memcheck over the new cases 0 errors, `gst-indent` leaves the hunks untouched, `clang-format` 16 clean, `doxygen-tag.sh` clean.

myungjoo:

This is a re-review of the amended commit. The PR was previously APPROVED by `myungjoo-bot` (review 5139128092) at head `3424b829`, which left two non-blocking follow-up suggestions. Those were addressed in a single amended commit and force-pushed to `ca083897`. This pass re-reviews the `3424b829..ca083897` delta and re-checks the PR as a whole.

## Delta review (`3424b829..ca083897`)

Scope matches exactly what the PR description claims: `gst/nnstreamer/nnstreamer_plugin_api_impl.c` (+2/-1), `gst/nnstreamer/tensor_common.h` (+1), `tests/common/unittest_common.cc` (refactor + one new test). No unrelated files or hunks slipped in.

**1. Guard-order fix** (`nnstreamer_plugin_api_impl.c`):
```c
mem_size[num] = gst_tensor_meta_info_get_header_size (&meta);
if (mem_size[num] == 0) {
  mem_size[num++] = total - offset;
  break;
}
mem_size[num] += gst_tensor_meta_info_get_data_size (&meta);
```
Traced `gst_tensor_meta_info_get_header_size()`: it returns `128` unconditionally for any valid V1 header and `0` for anything else (invalid meta, or a meta whose magic/version validate but is not V1). Since the check now fires on `header_size == 0` alone, before `data_size` is added:
- For every valid V1 tensor, `header_size` is always `128` (non-zero), so the branch never triggers and `offset` advances by exactly `header_size + data_size`, identical to the old code. Normal tensor streams are provably unaffected — confirmed by tracing all the existing positive/negative tests (`createFlexTensorBuffer`, `OverMemoryMax`, `ShortHeader_n`, `ShortTrailing_n`, `InvalidHeader_n`, `InvalidTrailing_n`) and none of them reach this reordered code path in a way that changes their outcome.
- For any header that parses (passes `gst_tensor_meta_info_validate()`) but is not V1, `header_size` is `0` regardless of format or data size, so the new check now closes the gap for **all** non-V1 formats, not just the "non-sparse" case the PR text calls out — the sparse case with `nnz > 0` is covered by the same one-line change, which is a bonus over what was scoped.

No regression risk found in the reorder.

**2. New test `createFlexTensorBufferUnknownVersionHeader_n`**: builds a 300-byte block with a valid-but-non-V1 version and `_NNS_TENSOR_FORMAT_STATIC` (so `header_size == 0`, `data_size == 1`). Manually traced both code paths:
- Pre-fix: `mem_size[0] = 0 + 1 = 1` (sum is non-zero, guard doesn't fire), `offset` advances to 1, loop re-parses at an unaligned offset into the mostly-zeroed buffer, magic no longer matches, hits the "cannot parse" exit → result is 2 memories (`1` + `299`), so `EXPECT_EQ(gst_buffer_n_memory(out), 1U)` fails.
- Post-fix: `header_size == 0` fires immediately at `num == 0` → single 300-byte memory. Test passes.

This confirms the new test genuinely gates the reorder rather than duplicating the existing `ZeroSizeHeader_n` case.

**3. Refactor of `createFlexTensorBufferZeroSizeHeader_n`** into the shared `build_unknown_version_data()` helper: the only behavioral difference is that `header[21]` (sparse `nnz`) is no longer set explicitly to `0`, but the block is `g_malloc0`'d, so it's `0` either way — the scenario byte-for-byte matches what the test checked before the refactor.

**4. Doxygen dedup**: confirmed both copies (`tensor_common.h:187` declaration and `nnstreamer_plugin_api_impl.c:532` definition) now carry the identical line `Bytes of a flexible buffer that do not describe a tensor are handed over as the last memory of the result.` — the two copies are back in sync.

## Standard re-check

- **Topic resolved** (#4920 item A2): yes, unchanged from the prior approval — the delta only sharpens the existing fix, doesn't alter its core mechanism.
- **Other-module regression**: none — the delta touches only the shared helper's internal ordering, a comment, and test code; no caller (`tensor_demux`, `tensor_crop`, `tensor_sparse*`, `tensor_converter`) was touched in this delta or needs to be.
- **Scope/size**: minimal, single-purpose amend (+43/-17 net in the delta); full PR vs. `main` merge-base is still tightly scoped to 3 files.
- **Test completeness**: 6 GTest negative cases + 1 positive + 1 pipeline SSAT-style GTest, all regression-gated (fail pre-fix, pass post-fix per the commit message and my own trace of the new one).
- **CI**: as of this review, `Static checks`, `CPP-Linter`, `DCO`, and the arm64 meson+Valgrind job are green; several jobs (including the x86 `meson`+Valgrind job, which is the one whose GTEST step actually runs `unittest_common` per the PR's own evidence) were still `pending` at review time — worth letting those finish before merging, but this is a CI-timing note, not a code defect.
- **Docs**: no separate documentation update needed; the Doxygen contract line is the appropriate place and it's now consistent.

## Verdict

No issues found in the amended commit or the delta since the prior approval — **merge as-is** once the still-pending CI jobs finish green (in particular the x86 meson/Valgrind job).

Issue #4934, opened 2026-09-07
[Common] Extra tensors of a flexible stream lose their meta header in the extra-tensor header


myungjoo:
A gst-buffer carries at most `NNS_TENSOR_MEMORY_MAX` (16) memory blocks, so the tensors past that are packed into the last block behind a `GstTensorExtraInfo` header. For flexible tensors the size bookkeeping in that header is wrong, and every extra tensor after the first comes back from the wrong offset. It is item **A6** of the audit list in #4920, found while writing the flexible variant of the tests in #4929.

Line numbers refer to `main` at `530f2a0e`.

## What happens

`gst_tensor_buffer_append_memory()` copies the incoming memory into the merged block **whole** — for a flexible tensor that is the 128-byte meta header plus the data — but records only the data size:

- `nnstreamer_plugin_api_impl.c:1879` — `gst_tensor_meta_info_convert (&meta, &extra_info->infos[new_mem_index])` stores dimension and type, so `gst_tensor_info_get_size()` of that entry is the data size, with no header.

`gst_tensor_buffer_get_nth_memory()` then walks the block with those sizes:

- `nnstreamer_plugin_api_impl.c:1745` — `offset += gst_tensor_info_get_size (&extra_info->infos[i - 1])` per preceding extra tensor.
- `:1750` — the returned share is `gst_tensor_info_get_size (&extra_info->infos[index - NNS_TENSOR_MEMORY_MAX])` bytes.

So each extra tensor is short by its header:

- the 17th tensor is returned truncated — header plus the first `size - 128` bytes of its data,
- the 18th starts 128 bytes inside the 17th tensor's data, so `gst_tensor_meta_info_parse_header()` reads tensor data as a header and fails validation,
- every tensor after that drifts further.

The 16th tensor is unaffected: it is returned with `extra_info->reserved` (`:1738`), which is the real size of the memory that was merged.

## Trigger

Any flexible stream with more than 16 tensors — tensor_mux/merge of 17+ flexible pads, an appsrc pushing such a buffer, tensor_query or edge receiving one. Reproduced through tensor_transform with 18 flexible tensors: the run logs `gst_tensor_meta_info_convert: assertion 'gst_tensor_meta_info_validate (meta)' failed` for the 18th tensor.

## Suggested fix

[...]

PR #4939, opened 2026-09-07
[Element/Decoder] Validate the incoming buffer before decoding

myungjoo:
This is item **C24** of the memory-safety audit tracking issue #4920.

## The defect

`gst_tensordec_transform()` maps every memory of the incoming buffer and hands it to the decoder sub-plugin without ever comparing it with the negotiated caps:

- `tensordec-directvideo.c` copies `input->size` bytes into an output frame sized from the negotiated dimensions, so an oversized buffer overflows that frame (item G14) and an undersized one is read past its end in the padding branch.
- The sub-plugins that accept flexible tensors (`octet_stream`, `flatbuf`, `flexbuf`, protobuf) parse the meta header of the memory and then use the data size it declares, without knowing how long the memory actually is (item G15).
- The number of memory chunks was checked with `g_assert (num_mems == num_tensors)`, which aborts the process for a buffer an application (`appsrc`) or a remote peer (`tensor_query`, `edge`) has built.

## The fix

Validate each memory in the element, before the sub-plugin sees it:

- a **static** tensor is fully described by the caps, so its size has to match exactly — the same rule `tensor_filter` has always applied to its input;
- a **flexible** tensor describes itself, so the memory has to be long enough for a meta header, the header has to parse into a version this build can size, and header + data have to fit in the memory. The header is parsed in place with `gst_tensor_meta_info_parse_header()` on the mapping the element already holds, the way the sub-plugins read it, with the length guard kept explicit and evaluated first;
- a chunk-count mismatch returns `GST_FLOW_ERROR` instead of aborting.

The two duplicated release loops are folded into one helper so the new error path does not add a third copy.

## Tests

`tests/nnstreamer_plugins/unittest_plugins.cc` gains 12 cases driving a `tensor_decoder` harness (`testTensorDecoder.*`). Verified against the pre-fix element, they fail as follows without this PR:

| case | without the fix |
| --- | --- |
| `pushInputSizeTooLarge_n` | FAILED (sub-plugin invoked with a surplus) |
| `pushInputSizeTooSmall_n` | FAILED |
| `pushInputMemoryCount_n`, `pushInputMemoryCountShort_n` | **abort** (`assertion failed: (num_mems == num_tensors)`) |
| `pushFlexibleInputTruncated_n` | FAILED |
| `pushFlexibleInputBrokenHeader_n` | FAILED |
| `pushFlexibleInputUnknownVersion_n` | FAILED |
| `pushFlexibleInputNoHeader_n` | FAILED |
| `pushDirectVideoInputSizeTooLarge_n` | **SIGSEGV** (the overflowed video frame) |
| `pushInputSize`, `pushFlexibleInput`, `pushFlexibleInputOversized` | pass (they pin the unchanged hand-over, header included, and the tolerated slack) |

Local verification on Ubuntu 22.04: `meson test` 22/23 (`unittest_filter_python3` fails identically on the unpatched tree — a numpy 1.x/2.x mismatch in this environment), SSAT 784 passed with the three python3 groups failing identically before and after the change. Every decoder group is green: `nnstreamer_decoder` 60/60, `nnstreamer_decoder_boundingbox` 51/51, `nnstreamer_octet_stream` 16/16, `nnstreamer_decoder_pose` 5/5, `nnstreamer_decoder_tensor_region` 2/2.

## Notes

- For a flexible stream with more than 16 tensors, the extra-tensor bookkeeping records the data size without the meta header (issue #4934), so the memory handed back for the 17th tensor is already one header short of what its own meta declares. This PR turns that into a clean `GST_FLOW_ERROR` instead of a truncated tensor handed to a sub-plugin; the underlying accounting is fixed separately in #4934.
- `gst_tensor_meta_info_validate()` accepts any version carrying the version marker, while `gst_tensor_meta_info_get_header_size()` can only size version 1. This element now refuses such a header rather than assuming a zero-length one; the same gap in the shared helpers is listed as item A7 of #4920.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_decoder.c
+++ b/gst/nnstreamer/elements/gsttensor_decoder.c
@@ -683,6 +683,82 @@ gst_tensordec_configure (GstTensorDecoder * self, const GstCaps * in_caps,
 }
 
 /**
+ * @brief Release the mapped memories of the incoming buffer.
+ * @param mem the memories acquired from the incoming buffer
+ * @param info the map info of each memory
+ * @param num the number of the memories to release
+ */
+static void
+gst_tensordec_release_input (GstMemory * mem[], GstMapInfo info[], guint num)
+{
+  guint i;
+
+  for (i = 0; i < num; i++) {
+    gst_memory_unmap (mem[i], &info[i]);
+    gst_memory_unref (mem[i]);
+  }
+}
+
+/**
+ * @brief Validate the memory of the incoming buffer with the negotiated config.
+ * @param self this pointer of tensor_decoder
+ * @param index the index of the tensor in the incoming buffer
+ * @param data the mapped memory of the tensor
+ * @param size the mapped size of @a data
+ * @return TRUE if the memory holds the tensor the decoder sub-plugin will read
+ */
+static gboolean
+gst_tensordec_check_input_size (GstTensorDecoder * self, guint index,
+    gpointer data, gsize size)
+{
+  gsize expected;
+
+  if (gst_tensors_config_is_flexible (&self->tensor_config)) {
+    GstTensorMetaInfo meta;
+    gsize hsize;
+
+    gst_tensor_meta_info_init (&meta);
+
+    if (size < gst_tensor_meta_info_get_header_size (&meta) ||
+        !gst_tensor_meta_info_parse_header (&meta, data)) {
+      GST_ERROR_OBJECT (self,
+          "Failed to parse the meta info of the %u'th tensor in the incoming buffer.",
+          index);
+      return FALSE;
+    }
+
+    /* a meta version this build cannot size gives no offset to the data */
+    hsize = gst_tensor_meta_info_get_header_size (&meta);
+    if (hsize == 0) {
+      GST_ERROR_OBJECT (self,
+          "The meta info of the %u'th tensor in the incoming buffer declares a version of which the header layout is unknown.",
+          index);
+      return FALSE;
+    }
+
+    expected = hsize + gst_tensor_meta_info_get_data_size (&meta);
+
+    if (size < expected) {
+      GST_ERROR_OBJECT (self,
+          "The %u'th tensor of the incoming buffer is %zd bytes, which is smaller than %zd bytes described by its own meta info.",
+          index, size, expected);
+      return FALSE;
+    }
+  } else {
+    expected = gst_tensors_info_get_size (&self->tensor_config.info, index);
+
+    if (size != expected) {
+      GST_ERROR_OBJECT (self,
+          "The %u'th tensor of the incoming buffer is %zd bytes, which is expected to be %zd bytes by the negotiated caps. Check whether the incoming stream is consistent with the caps of the sink pad.",
+          index, size, expected);
+      return FALSE;
+    }
+  }
+
+  return TRUE;
+}
+
+/**
  * @brief non-ip transform. required vmethod for BaseTransform class.
  */
 static GstFlowReturn
@@ -710,20 +786,26 @@ gst_tensordec_transform (GstBaseTransform * trans,
       self->tensor_config.info.num_tensors = num_mems;
     }
     num_tensors = self->tensor_config.info.num_tensors;
-    /** Internal logic error. Negotiation process should prevent this! */
-    g_assert (num_mems == num_tensors);
+    if (num_mems != num_tensors) {
+      GST_ERROR_OBJECT (self,
+          "The incoming buffer has %u memory chunks, which is expected to be %u, the number of tensors of the negotiated caps.",
+          num_mems, num_tensors);
+      return GST_FLOW_ERROR;
+    }
 
     for (i = 0; i < num_tensors; i++) {
       in_mem[i] = gst_tensor_buffer_get_nth_memory (inbuf, i);
       if (!gst_memory_map (in_mem[i], &in_info[i], GST_MAP_READ)) {
-        guint j;
         ml_logf ("Failed to map in_mem[%u].\n", i);
 
-        for (j = 0; j < i; j++) {
-          gst_memory_unmap (in_mem[j], &in_info[j]);
-          gst_memory_unref (in_mem[j]);
-        }
+        gst_tensordec_release_input (in_mem, in_info, i);
         gst_memory_unref (in_mem[i]);
+        return GST_FLOW_ERROR;
+      }
+
+      if (!gst_tensordec_check_input_size (self, i, in_info[i].data,
+              in_info[i].size)) {
+        gst_tensordec_release_input (in_mem, in_info, i + 1);
         return GST_FLOW_ERROR;
       }
 
@@ -741,10 +823,7 @@ gst_tensordec_transform (GstBaseTransform * trans,
       res = GST_FLOW_ERROR;
     }
 
-    for (i = 0; i < num_tensors; i++) {
-      gst_memory_unmap (in_mem[i], &in_info[i]);
-      gst_memory_unref (in_mem[i]);
-    }
+    gst_tensordec_release_input (in_mem, in_info, num_tensors);
   } else {
     GST_ERROR_OBJECT (self, "Decoder plugin not yet configured.");
     goto unknown_type;
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -22,6 +22,7 @@
 #include <nnstreamer_util.h>
 #include <string.h>
 #include <tensor_common.h>
+#include <tensor_decoder_custom.h>
 #include <tensor_meta.h>
 #include <unistd.h>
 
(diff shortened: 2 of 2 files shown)


myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the PR diff and the repository. It is being relayed here verbatim by another Claude session. Please treat it as a third-party review comment.

## Summary

This PR fixes item **C24** of the memory-safety audit (#4920): `gst_tensordec_transform()` used to hand the mapped memory of an incoming buffer to the decoder sub-plugin without checking its size against the negotiated caps, and used `g_assert()` to check the memory-chunk count. I traced the fix against the two sub-plugins the PR description names as vulnerable and confirm the defect and the fix:

- `ext/nnstreamer/tensor_decoder/tensordec-directvideo.c` (`dv_decode`, line 355): `memcpy (out_info.data, input->data, input->size)` with no bound on `input->size` against the output frame it copies into — an oversized input buffer overflows the frame. The new `gst_tensordec_check_input_size()` rejects any static-tensor memory whose mapped size isn't exactly `gst_tensors_info_get_size()`, which is exactly the size `dv_decode` assumes. `pushDirectVideoInputSizeTooLarge_n` reproduces the SIGSEGV pre-fix and pins `GST_FLOW_ERROR` post-fix.
- `ext/nnstreamer/tensor_decoder/tensordec-octetstream.c` (`os_decode`, line 96): for flexible tensors it reads `header_size + data_size` bytes straight out of `input[i].data` based on the meta header's own claim, with no check that the mapped memory is actually that long. The new check computes the same `header_size + data_size` via `gst_tensor_meta_info_parse_memory()` and requires `size >= expected`, which exactly bounds what `os_decode` (and, by the same pattern, flatbuf/flexbuf/protobuf) reads.
- The `g_assert (num_mems == num_tensors)` → `GST_FLOW_ERROR` change removes a process-abort reachable from `appsrc`/`tensor_query`/`edge` input, replaced by `pushInputMemoryCount_n`.

I also checked the "same rule tensor_filter has always applied" claim in the PR body against `tensor_filter.c`'s `_gst_tensor_filter_transform_get_invoke_tensors()`, which does an exact `!=` size check the same way — the claim holds.

I did not find a regression risk in the exact-match rule for static tensors: `tensor_decoder`'s sink caps (`CAPS_STRING` in `gsttensor_decoder.c`) only ever compose `GST_TENSOR_CAP_DEFAULT`, `GST_TENSORS_CAP_DEFAULT`, and `GST_TENSORS_FLEX_CAP_DEFAULT` — never `GST_TENSORS_SPARSE_CAP_DEFAULT` — so a sparse-format config can never reach the new "static" branch of `gst_tensordec_check_input_size()` and get wrongly forced into exact-size matching.

The memory-release refactor (`gst_tensordec_release_input`) looks correct on every path I traced: the map-failure branch keeps releasing `[0, i)` and separately unrefs `in_mem[i]` (unchanged from before), and the new size-check-failure branch releases `[0, i]` (`i + 1` count), which is right because `in_mem[i]`/`in_info[i]` were already successfully mapped at that point. No leak or double-free found.

## Findings

**1. (Low, non-blocking) No positive test pins the intentional tolerance for oversized flexible tensors.**
The flexible branch of `gst_tensordec_check_input_size()` accepts `size >= expected` (memory may be larger than `header + data`), which is a deliberate design choice (a self-describing tensor's own memory is allowed slack) — but only the exact-fit case (`pushFlexibleInput`) and the too-small cases are tested. A test pushing a flexible buffer strictly larger than `header_size + data_size` and asserting `GST_FLOW_OK`/sub-plugin-invoked would guard against someone later "tightening" this to an exact match by mistake and breaking legitimate flexible producers that over-allocate.

**2. (Very low / theoretical, non-blocking, FYI) Flexible-branch size arithmetic trusts attacker-controlled header fields.**
`gst_tensordec_check_input_size()`'s flexible path calls `gst_tensor_meta_info_get_data_size()`, whose sparse-format branch computes `nnz * (dsize + sizeof (guint))` from an `nnz` field read directly out of the incoming memory's own header (independent of the negotiated caps format, since a flexible tensor's per-memory header can itself declare `_NNS_TENSOR_FORMAT_SPARSE`). On a platform where `gsize` is 32-bit, a crafted `nnz` could make this multiplication wrap, shrinking `expected` and defeating the very check this PR adds for that one memory chunk. On the 64-bit CI targets this isn't reachable (the product doesn't overflow 64 bits for any `nnz`/`dsize` combination that matters here), and the arithmetic itself is pre-existing shared code already used the same way by `tensor_converter`/`tensor_crop`/`tensor_filter`, so this isn't introduced by this PR — flagging only as a hardening note for the underlying utility, not a defect to fix in this PR.

**3. (Nit, non-blocking) Doxygen completeness differs between the two new static helpers.**
`gst_tensordec_check_input_size()` documents `@param`/`@return` for all its arguments; `gst_tensordec_release_input()` has only `@brief`. This matches the surrounding file's existing convention (e.g. `gst_tensordec_configure()` also omits `@param`), so it's not a deviation from local style and I wouldn't block on it — just noting the inconsistency between the two sibling functions added in the same commit.

## Checklist against the review criteria

- **Solves the stated problem (C24):** Yes — traced against the two concretely-named vulnerable sub-plugins above.
- **No regression to other modules:** No regression found. Change is confined to `gst_tensordec_transform()` and two new static helpers in the same file; sparse-format caps can't reach the new static-branch exact-match check (see above). Existing static/flexible/direct_video/octet_stream flows keep the same size expectations they already had.
- **Scope/size proportionate to the topic:** Yes. `+74/-11` in one element file, plus test-only additions; no unrelated files touched.
- **Test coverage:** 8 new GTest cases (`testTensorDecoder.*`) via `GstHarness`, covering the oversized/undersized static case, the chunk-count-mismatch abort→error case, the flexible truncated/no-header cases, and an end-to-end `direct_video` SIGSEGV repro, plus two positive/pinning cases. One coverage gap noted above (oversized-flexible acceptance).
- **Future-regression detection:** These are black-box, harness-level tests against the element's observable behavior, so they'll catch a future regression (e.g. someone re-introducing `g_assert` or loosening/removing a check) regardless of which internal module changes.
- **CI enforcement:** `unittest_plugins` runs under `meson test`, wired into `.github/workflows/ubuntu_clean_meson_build.yml`, `ubuntu_clean_llvm_build.yml`, and `macos.yaml`, so a regression these tests catch will fail CI and block merge.
- **Architecture/API docs:** No public API/ABI change (both new functions are `static`); no doc update needed or missing.

## Verdict

No blocking issues found. The fix is correctly scoped, matches the described defect precisely (verified against the actual `memcpy`/meta-header-read sites it protects), doesn't risk regressing sparse/static/flexible negotiation elsewhere, and ships with harness-level regression tests wired into CI. The two "low" points above are optional follow-ups, not merge blockers.

myungjoo:
Thanks for the review. Both actionable points are addressed in the branch (force-pushed):

**1. Oversized flexible tensor had no positive test** — fixed. `testTensorDecoder.pushFlexibleInputOversized` now pushes a flexible memory of `header + 128` bytes whose meta declares 64 bytes of data, and asserts `GST_FLOW_OK`, the sub-plugin being invoked, and the full memory size being handed over. That pins the deliberate `>=` relation, so tightening the flexible branch to an exact match now fails a test rather than only breaking producers that over-allocate.

**3. Doxygen inconsistency between the two new helpers** — fixed. `gst_tensordec_release_input()` now documents `@param mem` / `@param info` / `@param num` like its sibling.

**2. `nnz`-driven multiplication in `gst_tensor_meta_info_get_data_size()`** — agreed on the analysis, and I am leaving it out of this PR deliberately. The arithmetic lives in the shared util that `tensor_converter`, `tensor_crop` and `tensor_filter` already use the same way, so hardening it belongs with those callers rather than in a decoder-side change; every build target this project ships (including the 32-bit arm CI legs, where `gsize` is 32-bit) reaches it through those paths equally. It is a pre-existing property of that helper, not something this PR introduces or widens: without this PR the flexible branch performed *no* size check at all, so the check added here is strictly stronger on every platform.

Local verification after the change: `testTensorDecoder.*` 9/9, and the pre-fix element still fails the six `_n` cases (two of them by abort and SIGSEGV).

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the PR diff and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session, as a second-round review after the force-push that addressed the previous review's two actionable points. Please treat it as a third-party review comment.

## Scope of this pass

I re-read the current diff (`gst/nnstreamer/elements/gsttensor_decoder.c` +88/-11, `tests/nnstreamer_plugins/unittest_plugins.cc` +365) against the repository and against the previous review round. I did not re-raise the three points from the first review; I confirmed the two actionable ones are in fact fixed in the current code:

- `testTensorDecoder.pushFlexibleInputOversized` now exists and pins the `size >= header + data` tolerance for flexible tensors (`GST_FLOW_OK`, sub-plugin invoked, full mapped size handed over).
- `gst_tensordec_release_input()` now has `@param mem` / `@param info` / `@param num`, matching its sibling `gst_tensordec_check_input_size()`.

The `nnz`-multiplication note (point 2) was left as a deliberate, disclosed non-fix with a reasonable rationale (pre-existing shared utility, used identically by `tensor_converter`/`tensor_crop`/`tensor_filter`, and strictly stronger than the previous no-check behavior); I have nothing to add there.

## New observations this round

**1. (Low, non-blocking) Redundant re-map for the flexible-tensor header check, and a minor deviation from the codebase's own convention.**

`gst_tensordec_check_input_size()` calls `gst_tensor_meta_info_parse_memory (&meta, mem)` (`gst/nnstreamer/elements/gsttensor_decoder.c:719`), which internally does its own `gst_memory_map(mem, &map, GST_MAP_READ)` / `gst_memory_unmap()` cycle just to read the header. But at the call site, the caller (`gst_tensordec_transform`) has *already* mapped the same `GstMemory` for `GST_MAP_READ` and holds the pointer in `in_info[i].data`. Every other in-tree caller that already has a mapped pointer in hand uses `gst_tensor_meta_info_parse_header(&meta, data_ptr)` directly instead of re-mapping via `_parse_memory()` — including, notably, `ext/nnstreamer/tensor_decoder/tensordec-octetstream.c:90`, the very sub-plugin this PR's flexible-branch check is meant to protect — as well as `gsttensor_crop.c:569`, `gsttensor_trainer.c:670`, `gsttensor_transform.c:1970`, and `tensor_filter.c:692`.

This isn't a correctness bug — GstMemory explicitly supports concurrent read-only maps of the same memory, so there's no double-free/deadlock risk — and for the common malloc-backed allocator the extra map is essentially free. But it is a real per-buffer overhead for flexible-tensor streams (an extra map/unmap pair on every single buffer `tensor_decoder` processes) that could matter for allocators where mapping isn't free (e.g. dmabuf-backed zero-copy paths elsewhere in the project), and it's a needless deviation from the local convention. Using `gst_tensor_meta_info_parse_header(&meta, in_info[i].data)` directly (with the existing `size < gst_tensor_meta_info_get_header_size(&meta)` guard, which the code already computes right after) would avoid it. Given `parse_memory()`'s own internal size gate is actually a *safety* feature (it refuses to touch memory shorter than a header before reading), I would not block on this — it's an optional follow-up, not a defect.

**2. Verified: no stack-array overrun via the `num_mems`/`num_tensors` path this PR's loop now walks.**
Since this PR reworks the `num_mems != num_tensors` check that used to be a `g_assert`, and for flexible tensors `self->tensor_config.info.num_tensors` is set to `num_mems` *before* that comparison (so the comparison is a no-op in the flexible case), I checked whether an attacker-supplied buffer could drive `num_mems` (hence the loop bound over the 256-entry `in_mem`/`in_info`/`input` stack arrays) past `NNS_TENSOR_SIZE_LIMIT` (256). It cannot: `gst_tensor_buffer_get_count()` derives the "extra tensor" count from `GstTensorExtraInfo.num_extra_tensors`, and `gst_memory_map_is_extra_tensor()` (`nnstreamer_plugin_api_impl.c:70`) already rejects any header claiming `num_extra_tensors > NNS_TENSOR_SIZE_EXTRA_LIMIT` with an explicit "the header comes from the buffer data, do not trust its tensor count" comment, bounding the total at exactly 256. This is pre-existing, untouched-by-this-diff code, so it's not a finding — just recording that I checked it, since it's the kind of thing this PR's stated threat model (buffers built by `appsrc`/`tensor_query`/`edge`) would otherwise make me suspicious of.

## Checklist against the review criteria

- **Solves the stated problem (C24):** Yes, confirmed against `tensordec-directvideo.c`'s `memcpy` and `tensordec-octetstream.c`'s header-driven read, as already traced in the first review round.
- **No regression to other modules:** None found. Change stays inside `gst_tensordec_transform()` and its two new static helpers in the same file.
- **Scope/size proportionate to topic:** Yes — `+88/-11` production code, no unrelated files touched; the `+365` in the test file is in line with 9 GTest cases for a security-sensitive input-validation fix.
- **Test coverage / future-regression detection / CI enforcement:** Unchanged from the first review's assessment — still solid. `unittest_plugins` runs under `meson test`, wired into the standard CI workflows, so these black-box harness tests will catch a future regression regardless of which internal module reintroduces it.
- **Architecture/API docs:** No public API/ABI change (both new functions are `static`); nothing to update.

## Verdict

No blocking issues in this round. The one new point (redundant re-map in the flexible branch) is a low-severity, optional efficiency/consistency follow-up, not a merge blocker. Combined with the first round's findings (all either fixed or explicitly and reasonably deferred), I consider this PR mergeable as-is.

myungjoo:
Thanks — the redundant mapping is fixed in the branch (force-pushed).

`gst_tensordec_check_input_size()` now takes the pointer the caller already mapped and parses the header in place with `gst_tensor_meta_info_parse_header()`, the way `os_decode()` and the other flexible sub-plugins do, with the "memory shorter than a header" guard that `gst_tensor_meta_info_parse_memory()` used to provide kept explicit and evaluated first:

```c
    gst_tensor_meta_info_init (&meta);

    if (size < gst_tensor_meta_info_get_header_size (&meta) ||
        !gst_tensor_meta_info_parse_header (&meta, data)) {
```

Short-circuit order matters here: a memory too short to hold a header is refused before anything reads it, so `pushFlexibleInputNoHeader_n` stays clean under the Valgrind CI leg as well. `gst_tensor_meta_info_init()` sets the current magic and version, so `gst_tensor_meta_info_get_header_size()` returns the same 128 the removed helper computed for its own guard — the accepted and refused sets are unchanged, which the four flexible cases (`pushFlexibleInput`, `pushFlexibleInputOversized`, `pushFlexibleInputTruncated_n`, `pushFlexibleInputNoHeader_n`) pin.

Local verification after the change: `testTensorDecoder.*` 9/9, `meson test` 22/23 (`unittest_filter_python3` fails identically on an unpatched tree — a numpy 1.x/2.x mismatch in this environment), gst-indent, clang-format and the doxygen-tag checker report no change.

Thanks also for the extra check on the 256-entry stack arrays — that bound comes from the `gst_memory_map_is_extra_tensor()` clamp added for item A1 in #4926, which is exactly why this PR does not need to re-clamp the count.

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the PR diff and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session, as a third-round review after the force-push that replaced `gst_tensor_meta_info_parse_memory()` with an in-place `gst_tensor_meta_info_parse_header()` call in the flexible branch. Please treat it as a third-party review comment.

## Scope of this pass

Per the request, I focused on whether the latest force-push (dropping the redundant re-map, point 1 of the second review round) introduced any new problem — specifically header-length/version handling, short-circuit evaluation order, and out-of-bounds read risk — rather than re-litigating points already resolved.

## Verification of the change

`gst_tensordec_check_input_size()`'s flexible branch (`gst/nnstreamer/elements/gsttensor_decoder.c:716-737`) now reads:

```c
gst_tensor_meta_info_init (&meta);

if (size < gst_tensor_meta_info_get_header_size (&meta) ||
    !gst_tensor_meta_info_parse_header (&meta, data)) {
  ...
  return FALSE;
}
```

- **Short-circuit order:** `||` guarantees `gst_tensor_meta_info_parse_header()` (which reads `val[0..21]`, i.e. up to byte offset 88) is never called unless `size >= gst_tensor_meta_info_get_header_size(&meta)` (128 for the only version this tree defines, `GST_TENSOR_META_MAKE_VERSION(1,0)`). So a memory shorter than a header is rejected before any byte of `data` is read — no OOB read.
- **Version handling:** this exactly mirrors the internal shape of `gst_tensor_meta_info_parse_memory()` (`gst/nnstreamer/nnstreamer_plugin_api_impl.c:1607-1640`), which itself does `gst_tensor_meta_info_init(meta)` → compute `hsize` from that freshly-initialized `meta` → compare against the memory's size → only then map and call `parse_header()`. The `hsize` used for the length gate always comes from a *freshly-initialized* `meta` (i.e., the current build's own version), not from whatever version byte the incoming buffer claims — but that is identical to what `parse_memory()` already does today, so this PR doesn't change that property, it just inlines it. Not a new issue.
- **Parity with `os_decode()`:** `ext/nnstreamer/tensor_decoder/tensordec-octetstream.c:90-92` (the sub-plugin this check exists to protect) already calls `gst_tensor_meta_info_parse_header(&meta, input[i].data)` directly on an already-mapped pointer, with no extra map. The new code now follows the same convention instead of being the odd one out.

I traced `gst_tensor_meta_info_parse_header()` → `gst_tensor_meta_info_validate()` and confirmed a malformed header (bad magic/version, invalid `type`, invalid `dimension`, invalid `format`, invalid `media_type`) is rejected via the `!gst_tensor_meta_info_parse_header(...)` half of the OR, independent of the size gate — so both a too-short memory and a long-enough-but-garbage header are refused, and no field read out of the parsed struct (e.g. `meta->type` in the later `gst_tensor_meta_info_get_data_size()` call) is used before `gst_tensor_meta_info_validate()` has bounds-checked it. No new defect found in this round.

## New observation this round

**1. (Low, non-blocking) The "size is large enough but the header content itself is invalid" branch has no dedicated test.**

All the flexible-path tests that produce a memory of at least header-size (`pushFlexibleInput`, `pushFlexibleInputOversized`, `pushFlexibleInputTruncated_n`) build their buffer through `_get_flex_decoder_buffer()`, which always calls `gst_tensor_meta_info_update_header()` to write a *valid* header before use. `pushFlexibleInputNoHeader_n` is the only case that fails validation, and it does so via the *size* half of the OR (16 bytes, below the 128-byte header size) — `gst_tensor_meta_info_parse_header()` is never even reached in that case because of the short-circuit confirmed above.

So the `!gst_tensor_meta_info_parse_header(&meta, data)` half of the guard — the path that rejects a memory that is long enough to hold a header but whose magic/version/type/dimension/format is garbage (e.g., a buffer built by a hostile `tensor_query`/`edge` peer that never wrote a real tensor-meta header, or one that's simply zero-filled) — is exercised by no test in this PR. It's the same validation `gst_tensor_meta_info_parse_memory()` already performed before this PR (so it's not a regression), but since this PR is specifically about hardening `tensor_decoder` against buffers built by an untrusted peer/application, a case like:

```c
// mem_size >= header size, but content is left as garbage/zero (no update_header call)
buf = gst_harness_create_buffer (h, hsize + 64);
EXPECT_EQ (gst_harness_push (h, buf), GST_FLOW_ERROR);
```

would pin that this failure path is reached and returns `GST_FLOW_ERROR` rather than, say, a crash or silent pass-through, and would catch a future regression in this exact branch (e.g., someone "simplifying" the OR into just the size check). Optional follow-up, not a merge blocker — the underlying validation itself is pre-existing, unmodified logic that already had no dedicated test before this PR either.

## Checklist against the review criteria

- **Solves the stated problem (C24):** Yes, unchanged from prior rounds — verified again against `tensordec-directvideo.c`'s `memcpy` and `tensordec-octetstream.c`'s header-driven read.
- **No regression to other modules:** None found. The diff is still confined to `gst_tensordec_transform()` and its two static helpers in `gsttensor_decoder.c`; the flexible-branch rewrite is behaviorally equivalent to the removed `gst_tensor_meta_info_parse_memory()` call (same length gate, same validation, same data pointer), just without the extra map/unmap pair.
- **Scope/size proportionate to topic:** Yes — production diff is still small and self-contained; no unrelated files touched.
- **Test coverage:** Solid; one small gap noted above (untested "valid length, invalid header content" branch), which pre-dates this PR's earlier rounds and is not something the latest force-push introduced.
- **Future-regression detection / CI enforcement:** Unchanged from prior assessment — `unittest_plugins` runs under `meson test`, wired into `ubuntu_clean_meson_build.yml`, `ubuntu_clean_llvm_build.yml`, and `macos.yaml` (confirmed in `tests/meson.build`), so anything these tests do cover will block merge on regression.
- **Architecture/API docs:** No public API/ABI change (both helpers remain `static`); nothing to update.

## Verdict

No blocking issues in this round. The redundant re-map fix from the previous round is correct and safe: short-circuit order prevents any OOB read, and the validation semantics are unchanged from the `gst_tensor_meta_info_parse_memory()` this code replaces. The one new note (an untested malformed-header branch) is low-severity and optional. I consider this PR mergeable as-is.

myungjoo:
Good catch — that branch is now covered too (force-pushed).

`testTensorDecoder.pushFlexibleInputBrokenHeader_n` pushes a flexible memory of `header + 64` bytes filled with `0x11`, so it clears the length guard and is refused by the header content itself (`gst_tensor_meta_info_parse_header()` failing on the magic). The buffer is memset rather than left at whatever the harness allocator returns, so the case reads no uninitialised bytes under the Valgrind leg.

Verified against the pre-fix element: the new case fails there like the other five `_n` cases, while both positive cases keep passing on either side. `testTensorDecoder.*` is now 10/10 with the fix; gst-indent, clang-format and the doxygen-tag checker report no change.

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the PR diff and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session, as a fourth-round review after the force-push that added `testTensorDecoder.pushFlexibleInputBrokenHeader_n`. Please treat it as a third-party review comment.

## Scope of this pass

Per the request, I did not re-litigate anything already resolved in the first three rounds (oversized-flexible positive test, Doxygen parity between the two helpers, dropping the redundant `parse_memory()` re-map in favor of in-place `parse_header()`). I focused only on whether the new `pushFlexibleInputBrokenHeader_n` case actually exercises the branch it claims to (the `!gst_tensor_meta_info_parse_header(&meta, data)` half of the guard, as opposed to the length half), and whether it reads any uninitialised memory.

## Verification

`gst_tensor_meta_info_parse_header()` reads `val[0]` as `meta->magic` and `gst_tensor_meta_info_validate()`'s first check is `GST_TENSOR_META_MAGIC_VALID(m)`, i.e. `m == 0xfeedcced` (`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:1421,1502-1506`). The new test builds the buffer as:

```c
buf = gst_harness_create_buffer (h, hsize + 64);
gst_buffer_memset (buf, 0, 0x11, hsize + 64);
```

- **Size guard is correctly bypassed:** `hsize + 64 >= hsize`, so `size < gst_tensor_meta_info_get_header_size(&meta)` is false and the `||` moves on to `gst_tensor_meta_info_parse_header()` — the test does reach the branch it targets, not the length check already covered by `pushFlexibleInputNoHeader_n`.
- **Content is correctly rejected, and for the reason claimed:** `val[0]` reads back as `0x11111111` (four `0x11` bytes), which fails `GST_TENSOR_META_MAGIC_VALID`, so `GST_TENSOR_META_IS_VALID(meta)` is false and `gst_tensor_meta_info_validate()` returns `FALSE` on its very first check — before `type`/`dimension`/`format`/`media_type` are even inspected. So this specifically exercises the "magic mismatch" sub-case of the invalid-header branch, matching the PR's own description ("fails ... on the magic").
- **No uninitialised read:** `gst_buffer_memset(buf, 0, 0x11, hsize + 64)` fills the *entire* mapped region (offset 0, length `hsize + 64` — the same size the buffer was created with), so every byte `parse_header()`/`validate()` touches (magic, version, type, the 16-entry dimension array, format, media_type — up to `val[20]`, i.e. byte offset 84) is deterministically `0x11`. `sparse_info.nnz` (`val[21]`) is never read because the `switch` on `meta->format` only reads it in the `_NNS_TENSOR_FORMAT_SPARSE` case, and `meta->format` here is `0x11111111`, which doesn't match that enum value — so no read of memory the test left "don't care" and no Valgrind flag is possible.

This matches the author's description in the follow-up comment exactly, and closes the gap the round-3 review noted (the "long enough for a header, but the header content itself is garbage" branch was previously untested).

## Remaining items

None. I re-checked the checklist from the previous rounds against the current HEAD and found nothing new:

- **Solves C24:** yes, unchanged from rounds 1–3.
- **Regression risk to other modules:** none; diff is still confined to `gst_tensordec_transform()` and its two static helpers in `gsttensor_decoder.c`.
- **Scope proportionate:** yes; this round's delta is one additional test case plus its harness plumbing.
- **Test coverage:** now covers both halves of the flexible-branch guard (too-short memory and long-enough-but-invalid-content), the static exact-size check both directions, the chunk-count mismatch, and the `direct_video` SIGSEGV repro, with matching positive/pinning cases.
- **CI enforcement:** `unittest_plugins` still runs under `meson test`, wired into the standard build workflows, so a regression in any of these paths fails CI and blocks merge.
- **Docs:** no public API/ABI change (both helpers are `static`); nothing to update.

## Verdict

No blocking issues. The round-3 gap is closed correctly — the new test exercises the intended branch for the intended reason, with a fully deterministic buffer and no uninitialised-read risk. I consider this PR mergeable as-is.

myungjoo:
Thanks for the approval — all four nits are handled, three of them in the branch (force-pushed to `e350fc2c`).

**1. Stale description.** Fixed: the description now says the header is parsed in place with `gst_tensor_meta_info_parse_header()` on the mapping the element already holds, with the length guard kept explicit and evaluated first. The wording predated the change made for the second review round.

**2. Undocumented test statics.** Fixed: `decoder_custom_invoked` / `decoder_custom_size` now carry a brief like `tensor_split_logged_critical` does.

**3. Only the `num_mems > num_tensors` direction.** Fixed: `pushInputMemoryCountShort_n` negotiates two tensors and pushes a single memory holding both, so the short direction goes through the same branch. Like its sibling, it aborts on the pre-fix element.

**4. Valid magic, non-V1 version.** Handled in the element rather than left as a follow-up, because the element is the layer that computes `header + data` and it should not do that arithmetic on a layout it cannot size:

```c
    /* a meta version this build cannot size gives no offset to the data */
    hsize = gst_tensor_meta_info_get_header_size (&meta);
    if (hsize == 0) { ... return FALSE; }
```

`pushFlexibleInputUnknownVersion_n` covers it: a valid V1 header whose version word is raised to major 2, which still passes `gst_tensor_meta_info_validate()`. Verified in isolation — the case fails against the previous revision of this PR (the one you reviewed) and passes with the guard, while the other eleven cases are unaffected in both directions.

You are right that the shared helpers are where this really belongs: `gst_tensor_meta_info_validate()` accepting a version that `gst_tensor_meta_info_get_header_size()` cannot size lets every other consumer (`octet_stream`, flatbuf, flexbuf, `tensor_filter`, `tensor_converter`) locate the payload at offset 0. Since that touches callers outside this element, it is filed as item **A7** of #4920 rather than widened into this PR.

Local verification after the change: `testTensorDecoder.*` 12/12, `meson test` 22/23 (`unittest_filter_python3` fails identically on an unpatched tree — a numpy 1.x/2.x mismatch in this environment), gst-indent, clang-format and the doxygen-tag checker report no change.

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the PR diff and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session. Please treat it as a third-party review comment.

## Scope of this pass

Per the request, I focused only on the delta introduced **after** the `myungjoo-bot` approval on `2428859c` — i.e., the fifth force-push (`e350fc2c`) that addressed nit #4 ("valid magic, non-V1 version") from that approval, plus the accompanying test additions. I did not re-litigate anything already resolved in rounds 1–5 (positive flexible-oversize test, Doxygen parity, dropping the redundant `parse_memory()` re-map, the broken-header test, the `num_mems`/stack-array bound check).

Diffed `2428859c...e350fc2c` directly (not the cumulative PR diff, since the two commits are rewritten/amended versions of each other, not ancestor→descendant) to isolate exactly what changed:

- `gst/nnstreamer/elements/gsttensor_decoder.c`: the new `hsize == 0` guard in `gst_tensordec_check_input_size()`'s flexible branch.
- `tests/nnstreamer_plugins/unittest_plugins.cc`: `pushInputMemoryCountShort_n`, `pushFlexibleInputUnknownVersion_n`, and a brief comment on `decoder_custom_invoked`/`decoder_custom_size`.

## Verification

**1. The `hsize == 0` guard is correct and introduces no regression risk.**

I traced `GST_TENSOR_META_VERSION_VALID`, `GST_TENSOR_META_MAKE_VERSION`, `GST_TENSOR_META_IS_V1`, and `gst_tensor_meta_info_get_header_size()` in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:1421-1555`:

- `gst_tensor_meta_info_get_header_size()` returns `0` whenever `GST_TENSOR_META_IS_VALID(meta)` holds (valid magic + `GST_TENSOR_META_VERSION_VALID`) but `GST_TENSOR_META_IS_V1(meta->version)` does not — i.e., exactly the "magic OK, version-family OK, but not the one layout this build knows how to size" case the new guard targets.
- I grepped the whole tree for anywhere `GstTensorMetaInfo.version` is ever set to something other than the default: the only site is `gst_tensor_meta_info_init()`, which always assigns `GST_TENSOR_META_VERSION` (`= MAKE_VERSION(1,0)`, i.e. `IS_V1` true). No in-tree producer (converter, split/merge, `tensor_query`, `edge`, muxers) can emit a header with a different version, so **no legitimate current stream can trip this guard** — it only rejects a case that was already unrepresentable in this codebase's own producers, i.e. a hostile/foreign peer.
- Without the guard, the prior code computed `expected = 0 (header) + data_size`, silently *underestimating* the required buffer length for an un-sizable version and letting a buffer that is merely `data_size` bytes long slip past `size < expected`, even though the real (unknown) header offset means the "data" the decoder sub-plugin would then read is at the wrong location. So this isn't cosmetic — it closes a real hole where the size check could pass for the wrong reason.

**2. `pushFlexibleInputUnknownVersion_n`'s version value is correct and exercises the intended branch.**

`0xDE002000`:
- `& 0xDE000000 == 0xDE000000` → `GST_TENSOR_META_VERSION_VALID` **true** (so `gst_tensor_meta_info_validate()` still passes — the type/dimension/format/media_type fields are untouched by the test, since only word `val[1]` is overwritten).
- `GST_TENSOR_META_IS_V1`: `(v & 0x00FFF000) & MAKE_VERSION(1,0)` = `(0xDE002000 & 0x00FFF000) & 0xDE001000` = `0x002000 & 0xDE001000` = `0` → **false**. So `get_header_size()` returns 0, and the new guard fires.
- I hand-computed what the *pre-guard* code would have done with this exact buffer: `expected = 0 + data_size(64) = 64`, while the mapped memory is `hsize(128) + 64 = 192` bytes (sized from the buffer's original, uncorrupted meta before the test overwrites the version word) → `192 < 64` is false, so the old check would have **passed** and handed the buffer to the sub-plugin. This matches the author's claim in the thread ("fails against the previous revision... passes with the guard") and confirms the test is a genuine regression pin, not a no-op.
- The comment "raise the major of the version (bits 12-23...)" is accurate: `0x2000` sets bit 13, which is `major=2` under `MAKE_VERSION(major,minor) = major<<12 | minor | 0xDE000000`.

**3. No scope creep, no doc gap.** Both changes stay inside `gsttensor_decoder.c`'s two existing static helpers and the same test file; nothing else is touched. The two new helpers remain `static` (confirmed via grep — no external references), so there is no API/ABI surface change and no architecture doc is owed for this delta.

**4. Style.** The C-side diff is 2-space K&R, consistent with the surrounding function added in earlier rounds. `clang-format --style=file` on the test file after the delta produces no diff, so the `.cc` formatting is clean.

## Remaining item (not introduced by this delta, already disclosed)

The author correctly scoped this guard to the element itself rather than fixing `gst_tensor_meta_info_validate()`/`get_header_size()` in the shared utility (which would also protect `tensor_filter`, `tensor_converter`, `octetstream`, flatbuf/flexbuf against the same "unsizable but validate()-passing version" gap), and filed that as item **A7** of #4920 instead of widening this PR. That is the right call for this PR's scope — flagging only so it's on record that A7 is the follow-up to track, not a gap in this PR.

## Checklist against the review criteria

- **Solves the stated problem (C24):** Yes — this delta specifically closes the one remaining nit from the approval round (non-V1 version silently computing a wrong `expected`), without touching the already-approved core fix.
- **No regression to other modules / other tensor_decoder streams:** None found — verified no in-tree producer can emit a non-V1 header, so the guard cannot fire on legitimate traffic.
- **Scope/size proportionate:** Yes — an 11-line production guard plus two focused test cases and one comment; no unrelated files touched.
- **Test coverage / future-regression detection:** Both new tests are black-box `GstHarness` cases in `unittest_plugins`, so a future change that removes the guard or breaks the `num_mems < num_tensors` direction fails `meson test` regardless of which internal module caused it.
- **CI enforcement:** `unittest_plugins` is wired into `ubuntu_clean_meson_build.yml`, `ubuntu_clean_llvm_build.yml`, and `macos.yaml`, so these two new cases will gate merges the same way the other ten already do.
- **Architecture/API docs:** No public API/ABI change in this delta; nothing to update.

## Verdict

No blocking issues in this delta. The guard is verified correct against the actual macro definitions (not just plausible-sounding), the test's crafted version value is verified to land exactly on the `IS_V1`-false / `VERSION_VALID`-true boundary it claims to, and the test is verified to actually fail without the guard. Combined with five prior review rounds all resolved, I consider this PR mergeable as-is.

myungjoo:
Thanks for the re-review.

**1. The hard-coded `0xDE002000U`** — good point, and it is now pinned rather than assumed. Before the push, the case asserts what the literal has to mean, using only exported helpers:

```c
  ASSERT_TRUE (gst_tensor_meta_info_parse_header (&meta, map.data));
  ASSERT_TRUE (gst_tensor_meta_info_get_version (&meta, &major, NULL));
  ASSERT_NE (major, 1U);
  ASSERT_EQ (gst_tensor_meta_info_get_header_size (&meta), 0U);
```

So if the encoding of the marker or of the major ever changes, the case fails on those assertions instead of quietly passing on the parse failure one branch earlier — which is the degradation you described. I went with the public helpers rather than a log handler because `GST_ERROR_OBJECT` only reaches a handler when the debug system is compiled in and the category threshold allows it, so a log-based assertion would be silently vacuous in a `GST_DISABLE_GST_DEBUG` build. The comment above the patch now names the file the two macros live in.

**2. The stale `parse_memory()` wording** — the PR description was rewritten before this re-review (it now says the header is parsed in place with `gst_tensor_meta_info_parse_header()` on the mapping the element already holds); the phrase may have come from the C24 line of #4920, whose "Fix:" note still suggests `gst_tensor_meta_info_parse_memory()`. That line is the audit's original suggestion rather than a description of this code, so I left the issue text as written and it now carries the PR link next to it.

`testTensorDecoder.*` 12/12 with the added assertions, `meson test` 22/23 (`unittest_filter_python3` fails identically on an unpatched tree — a numpy 1.x/2.x mismatch in this environment), gst-indent, clang-format and the doxygen-tag checker report no change. Head is now `9be097ba`; DO NOT MERGE is back on until CI finishes on it.

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working from the delta `e350fc2c..9be097ba` (the round after the last `myungjoo-bot` approval) and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session; please treat it as a third-party review comment.

## Scope reviewed

Only the delta since the last approved commit (`e350fc2c`), i.e. `9be097ba`: the four new assertions added to `testTensorDecoder.pushFlexibleInputUnknownVersion_n` in `tests/nnstreamer_plugins/unittest_plugins.cc`, plus the comment explaining them. No production code changed in this round; I did not re-review the already-approved `gsttensor_decoder.c` fix or the other 11 test cases.

## Verification

**The delta does what it claims.** I independently traced the macros in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`:

- `0xDE002000U & 0xDE000000 == 0xDE000000` → `GST_TENSOR_META_VERSION_VALID` true, so `gst_tensor_meta_info_validate()` (and hence `parse_header()`) still returns `TRUE`.
- `major = (0xDE002000 & 0x00FFF000) >> 12 = 2`, so `ASSERT_NE (major, 1U)` correctly captures "not V1".
- `GST_TENSOR_META_IS_V1` evaluates to false for this value, so `gst_tensor_meta_info_get_header_size()` returns `0`, matching `ASSERT_EQ (..., 0U)`.

So the four new assertions do exactly what the accompanying comment says: they pin the literal `0xDE002000U` to "validates, but a version this build cannot size" using only the three exported helpers (`gst_tensor_meta_info_parse_header`, `_get_version`, `_get_header_size`), so a future change to the private encoding in `nnstreamer_plugin_api_util_impl.c` would fail loudly here instead of the case silently starting to test the wrong (already-covered) "parse failure" branch one step earlier. This closes exactly the gap noted in the prior review round ("if that encoding ever changes... it stays green while no longer testing what it names"), and I have no correctness objection to it.

## One low-severity, non-blocking observation

The four new assertions (`ASSERT_TRUE`/`ASSERT_NE`/`ASSERT_EQ` at lines ~10788-10791) sit **between** `gst_buffer_map(buf, &map, GST_MAP_WRITE)` and `gst_buffer_unmap(buf, &map)`. That placement itself is correct and in fact required — `map.data` must not be used after `unmap()`, so the read-back has to happen while still mapped.

The side effect is that these are `ASSERT_*` (fatal, early-`return`) rather than `EXPECT_*` (non-fatal). If any of them ever fails — which, per the analysis above, only happens if the private version-marker/major-bit encoding in `nnstreamer_plugin_api_util_impl.c` changes — the test function returns immediately and skips:

- `gst_buffer_unmap (buf, &map)` (buffer stays mapped and is leaked — never pushed, never unreffed),
- `gst_harness_teardown (h)`,
- `gst_tensors_config_free (&config)`, and
- `nnstreamer_decoder_custom_unregister ("tdec_size")`.

The last one has the widest blast radius: several other `TEST`s in this same file (e.g. the neighboring `pushInputMemoryCount_n`, `pushInputMemoryCountShort_n`, `pushFlexibleInputTruncated_n`, `pushFlexibleInputBrokenHeader_n` — all in the ~10600-10760 range) begin with `ASSERT_EQ (0, nnstreamer_decoder_custom_register ("tdec_size", ...))`. If `tdec_size` is left registered because this test bailed out early, every one of those would immediately fail its own registration assert too, in the same `gtest` binary — turning one root-cause failure (the encoding assumption broke) into a wall of unrelated-looking failures, which makes the CI output harder to triage even though it still correctly blocks the merge.

To be clear, this is **not a new anti-pattern introduced by this delta** — the same shape already exists throughout this test file today (e.g. `ASSERT_TRUE (h != NULL)` before any teardown, present in this test both before and after the round under review, and in its neighbors). This delta just adds four more fatal-assertion points inside an already-fatal-assertion-heavy test body, using the file's existing convention. I'm not asking for a rewrite — just flagging it for the record in case a future pass wants to harden this file's negative-path tests (e.g. `EXPECT_*` for the four "assumption pin" checks specifically, so a break in the encoding assumption reports as one isolated failure instead of a cascade). Severity: **low, non-blocking**.

## Verdict

No blocking issues in this delta. The branch-pinning assertions are verified correct against the actual macro definitions, and the one observation above is a minor test-hygiene nit consistent with pre-existing style in this file, not a defect this delta introduces. Combined with the prior review rounds already resolved, I consider this PR mergeable as-is.

myungjoo:
Taken — the four assumption pins are `EXPECT_*` now, so a changed version encoding reports the failure and still runs `gst_buffer_unmap()`, the harness teardown and `nnstreamer_decoder_custom_unregister ("tdec_size")` on the way out. That matters more than the usual style consistency here precisely because of the cascade you describe: the sibling cases re-register the same custom decoder name and would have failed for a reason unrelated to what broke.

The `gst_buffer_map()` check above them stays `ASSERT_TRUE`, since continuing past a failed map would dereference `map.data`.

Head is now `c9ff3d1e`; `testTensorDecoder.*` 12/12, formatters and the doxygen-tag checker report no change.

myungjoo:
Fixed — `guint major = 0;` (head `dad8791f`).

You are right that the two preceding checks being non-fatal is exactly what made the uninitialised read reachable: `gst_tensor_meta_info_get_version()` leaves `*major` untouched when the meta does not validate, so a future encoding change would have produced an indeterminate read in `EXPECT_NE (major, 1U)` on top of the real failure — and on the valgrind leg that noise would land in the same report as the failure it is meant to explain. Initialising it keeps the shape and the diagnosis.

Also worth recording that this is the same class of defect the PR itself is about — the C2 item of #4920 was an uninitialised read reached through a path nobody expected to be taken — so it is a fitting last thing to remove from the case.

`testTensorDecoder.*` 12/12, gst-indent, clang-format and the doxygen-tag checker report no change on this head. DO NOT MERGE is back on until CI finishes on `dad8791f`.

myungjoo:
> **Note:** This review was written by a separate code-review agent session (not the PR author's own agent), working only from the delta `c9ff3d1e..dad8791f` (the round after the last approval) and the repository at the current HEAD of `fix/4920-decoder-input-size`. It is being relayed here verbatim by another Claude session. Please treat it as a third-party review comment.

## Scope reviewed

Only the delta since the approved commit `c9ff3d1e` (myungjoo-bot, `status: approved`), i.e. `dad8791f`. `git diff c9ff3d1e..dad8791f` (the two commits are an amend pair, not ancestor/descendant, so this was diffed directly rather than walked as a range) is exactly one line in `tests/nnstreamer_plugins/unittest_plugins.cc`:

```c
-  guint major;
+  guint major = 0;
```

No production code changed in this round; `gst/nnstreamer/elements/gsttensor_decoder.c` and the other 11 test cases are untouched, so this review does not re-litigate anything already resolved in the three prior rounds.

## Verification

The delta closes exactly the non-blocking note from the previous approval ("`guint major;` is left uninitialised... a future encoding change... would leave `EXPECT_NE (major, 1U)` reading an indeterminate value"). I checked the claim against the actual implementation:

- `gst_tensor_meta_info_get_version()` (`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:1480-1494`) starts with `if (!GST_TENSOR_META_IS_VALID (meta)) { return FALSE; }` and only writes `*major` after that check — so on the `FALSE` path it genuinely never touches `*major`, confirming the uninitialised-read claim was real, not hypothetical.
- In `pushFlexibleInputUnknownVersion_n` (`tests/nnstreamer_plugins/unittest_plugins.cc:10750-10796`), the two checks that gate reaching `EXPECT_NE (major, 1U)` — `EXPECT_TRUE (gst_tensor_meta_info_parse_header (...))` and `EXPECT_TRUE (gst_tensor_meta_info_get_version (&meta, &major, NULL))` — are both non-fatal `EXPECT_TRUE`, by deliberate design from the prior round (so the test doesn't leak the still-mapped buffer / skip harness teardown / skip `nnstreamer_decoder_custom_unregister` on failure). That means execution does continue into `EXPECT_NE (major, 1U)` even if `get_version()` returned `FALSE` without writing `major` — exactly the path that reads an uninitialised local without the fix.
- With `guint major = 0;`, that path now reads a well-defined `0` instead of indeterminate stack contents. The fix is minimal, correctly scoped to the one variable involved, and does not change what the test asserts in the passing case (the value written by `get_version()` on success is unaffected).
- This is also the right place to fix it (declaration-site initialisation) rather than, e.g., only special-casing the assertion — it protects any future code added between the declaration and the first write, not just today's four lines.

I did not find any way this one-line change could itself introduce a new problem: it can't affect Valgrind/ASan-clean runs (where `get_version()` already succeeds and overwrites `major`), and on the failure path it now degrades a real "uninitialised read" diagnostic in the same direction the fix was meant to go (a defined 0 that makes `EXPECT_NE` still exercise a real comparison, rather than UB).

## Checklist against the review criteria

- **Solves the stated problem:** Yes — this delta is not part of the C24 fix itself but of test hygiene raised in review; it does exactly what was asked and no more.
- **No regression to other modules:** None possible — single-line, test-only, one local variable's initialiser.
- **Scope/size proportionate:** Yes, trivially — 1 line changed, matching a 1-line ask.
- **Test coverage / future-regression detection:** N/A for this delta itself (it fixes a test, not adds one); the case it fixes continues to pin the version-encoding assumption as verified in the prior round.
- **CI enforcement:** Unchanged from prior rounds — `unittest_plugins` runs under `meson test` in the standard CI workflows. Note that CI on `dad8791f` itself was still queued/pending at review time (`Analyze (python)`, `DCO`, `Spell Check` green; the build/test legs including the Valgrind jobs had not reported yet) — worth confirming those land green on this exact head before merging, consistent with the author's own "DO NOT MERGE until CI finishes on `dad8791f`" note.
- **Architecture/API docs:** No public API/ABI change; nothing to update.

## Cumulative PR assessment (all rounds)

Across the full history (`eea3425c..dad8791f`, +101/-11 in `gsttensor_decoder.c`, +491 in `unittest_plugins.cc`, no other files touched): the stated C24 defect (unchecked buffer size handed to decoder sub-plugins, plus the `g_assert` chunk-count abort) is fixed and traced against the concrete `memcpy`/header-read sites in `tensordec-directvideo.c` and `tensordec-octetstream.c`; the fix stays confined to `gst_tensordec_transform()` and two new static helpers in the same file; static/flexible/sparse caps interaction was checked and found not to introduce a regression; 12 GTest cases exercise both the positive and negative paths and are wired into the existing CI meson-test jobs; no public API/ABI change, so no doc update is required. All prior actionable review points (positive test for oversized-flexible tolerance, doxygen consistency, redundant re-map, ASSERT-vs-EXPECT cascade risk, the version-literal test-arithmetic gap, and now this uninitialised read) were addressed in-branch; the two deliberately-deferred items (`nnz`-multiplication overflow hardening in the shared meta-info util, and the >16-tensor accounting gap tracked separately as #4934) are pre-existing/out-of-scope and correctly left out with a stated rationale in the PR description.

## Verdict

No blocking issues found in this delta or remaining in the cumulative PR. The one-line fix correctly and precisely resolves the previously-flagged uninitialised read, verified against the actual `get_version()` implementation and the non-fatal-assertion control flow that made it reachable. Mergeable once CI on `dad8791f` reports green (it was still pending on the relevant build/valgrind legs at the time of this review).

PR #4946, opened 2026-09-08
[Test/Join] Hold the pad release feeder to a lead over the sink

myungjoo:
`join.padReleaseWhileStreaming` is the largest and by far the most variable item left in the Valgrind step. Under memcheck in CI it has taken 19, 29, 72, 103, 204 and 219 s across six consecutive runs and **830 s** in a seventh — against 250 ms natively. That 830 s run spent **22.5 of the step's 30-minute budget**, and since #4936 a step timeout is the one failure that step cannot absorb, so one bad draw turns a contributor's PR red.

## Why it varies

The test keeps a pad streaming while it is released. The feeder thread pushed as fast as it could:

```c
while (!g_atomic_int_get (&feeder->stop)) {
  GstBuffer *buf = gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192);
  if (gst_app_src_push_buffer (GST_APP_SRC (feeder->appsrc), buf) != GST_FLOW_OK)
    break;
}
```

appsrc queues whatever it is given, so how much that thread leaves behind is set by how long the *test* thread takes to reach the release — and the queue is torn down one buffer at a time at the end of the attempt. The slower the environment, the more there is to tear down. Timing a 22 s local run with `GST_DEBUG=GST_STATES:4`, 12.9 s and 7.8 s of it were single gaps inside the last pipeline's teardown.

That is the feedback loop, and it is why the same test costs 250 ms natively and 830 s under a checker.

## The change

Hold the feeder to a lead of eight buffers over what the sink has taken. It pushes again as soon as the sink takes one, so the pad keeps streaming; the queue simply cannot grow with how slow the rest is.

Measured on one machine, memcheck against native:

| | time | attempts |
|---|---|---|
| before, memcheck | 14.9 s | 4 of 20 |
| **after, memcheck** | **1.2 s** | **20 of 20** |
| native | 213 ms | 20 of 20 |

Quicker *and* more thorough. The 10 s budget in this test exists to stop a checker paying for every attempt, and it was doing that by cutting the run short at 4 — exactly where each attempt had become expensive. With the queue bounded the budget is no longer reached at all, so memcheck now gets the full twenty.

The whole `unittest_join` binary under memcheck goes from 160 s (a healthy CI run) to 8.8 s locally.

The review put an A/B on the same CI infrastructure an hour apart, which is better evidence than any of my local numbers:

| | native | memcheck |
|---|---|---|
| main, unmodified (job of #4937) | 249 ms | **93,996 ms** |
| this PR | 213 ms | **760 ms** |

## It catches the race sooner, not later

This is a race regression test, so faster is worthless if it stops catching the race. I reintroduced the defect it was written for — releasing the pad without stopping its streaming thread first, so the thread can make the released pad active again after the release has cleared it — by moving `gst_pad_set_active` and `gst_element_remove_pad` after the locked section in `gst_join_release_pad`, and counted attempts until the test noticed:

| feeder | attempts to first detection | per attempt |
|---|---|---|
| before | 14, 64, 25, 32, 50 | ~1 in 38 |
| **after** | **1, 1, 2, 2, 4** | **~1 in 2** |

I first reported this as whole runs of the test caught it — 10 of 20 before, 20 of 20 after. Those numbers follow from the rates above and the 20-attempt budget, but they are a poor thing to quote: at one attempt in 38, whether a run notices is close to a coin toss against machine load, and a reviewer running the same control on this same machine saw no detection either way. The attempt counts above are what is stable, and the reviewer's independent higher-N measurement agreed with them: about 4-5% per attempt unthrottled against 37-49% throttled.

A first attempt at that control moved only `gst_pad_set_active` and caught nothing with either feeder — `gst_element_remove_pad` deactivates the pad itself, so that mutation was a no-op. Worth recording for the next person who needs this control.

## The lead is not a tuned number

Sweeping it, with the correct element for cost and the reintroduced defect for detection:

| lead | memcheck | attempts to first detection |
|---|---|---|
| 2 | 1074 ms | 3 |
| 4 | 1111 ms | 4 |
| 8 | 1136 ms | 2 |
| 16 | 1133 ms | 3 |
| 64 | 1178 ms | 3 |

No trend in either column across a 32x range. What matters is that a bound exists at all — unbounded is the 14.9 s and one-in-38 above — so eight is a round number in a wide flat region rather than a tuned one.

## The wait is a yield, and that is measured too

Review suggested `g_usleep (100)` in place of the `g_thread_yield ()`, to stop the feeder competing for CPU with the pipeline on a 2-vCPU runner. Measured, it costs the race:

| wait | memcheck | native | attempts to first detection |
|---|---|---|---|
| `g_thread_yield ()` | 1.1 s | 211 ms | 5, 1, 1, 1, 1 — mean 1 |
| `g_usleep (100)` | 1.1 s | 211 ms | 11, 60, 17, 45, 3 — mean **27** |

Same cost either way, and detection goes most of the way back to the unbounded 38. The pad sits idle for the sleep after every buffer the sink takes, which is a long time next to the window this test is aiming at. Keeping the yield.

## Testing

`JOIN_FEEDER_LEAD` is asserted from both sides rather than merely applied: the feeder records the largest lead it reached, and the test checks it is at most the lead — so removing the hold fails the test rather than only showing up as a slow job — and more than none, so a hold that stopped the feeder outright would fail too. Neither assertion missed in 30 native runs or 3 under memcheck.

All 13 `unittest_join` cases pass natively and under memcheck against the unmodified element. clang-format clean.

## Not in this change

The other two ways to bound this were considered and dropped. `appsrc`'s `leaky-type` would be the idiomatic answer but needs GStreamer 1.20, above what this project builds against elsewhere. `block=true` risks the opposite failure: after the pad is released nothing drains the queue, so a blocked push would never return and `g_thread_join` would hang.

### DIFF
--- a/tests/gstreamer_join/unittest_join.cc
+++ b/tests/gstreamer_join/unittest_join.cc
@@ -457,7 +457,20 @@ TEST (join, eosAfterPadRelease)
 typedef struct {
   GstElement *appsrc;
   gint stop;
+  guint *received;
+  guint pushed;
+  guint max_lead;
 } JoinFeeder;
+
+/**
+ * @brief Buffers the feeder may push before the sink has taken them.
+ * @detail appsrc queues whatever it is given, so an unheld feeder runs ahead
+ *         by as much as the test thread is slow - milliseconds natively, and
+ *         minutes under a memory checker, where the queue it leaves behind is
+ *         then torn down one buffer at a time. A lead keeps the pad streaming
+ *         without letting that cost grow with the environment.
+ */
+#define JOIN_FEEDER_LEAD (8U)
 
 /**
  * @brief Push buffers into an appsrc until the caller asks it to stop.
@@ -468,10 +481,24 @@ join_feeder_thread (gpointer data)
   JoinFeeder *feeder = (JoinFeeder *) data;
 
   while (!g_atomic_int_get (&feeder->stop)) {
-    GstBuffer *buf = gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192);
+    GstBuffer *buf;
+    guint taken = g_atomic_int_get (feeder->received);
+    guint lead = feeder->pushed > taken ? feeder->pushed - taken : 0;
+
+    if (lead > feeder->max_lead)
+      feeder->max_lead = lead;
+
+    if (lead >= JOIN_FEEDER_LEAD) {
+      g_thread_yield ();
+      continue;
+    }
+
+    buf = gst_buffer_new_wrapped (_g_memdup (test_frames[0], 192), 192);
 
     if (gst_app_src_push_buffer (GST_APP_SRC (feeder->appsrc), buf) != GST_FLOW_OK)
       break;
+
+    feeder->pushed++;
   }
 
   return NULL;
@@ -501,7 +528,7 @@ run_pad_release_while_streaming (void)
 {
   GstElement *appsrc_0, *join_handle, *sink_handle;
   GstPad *sinkpad_0, *active_pad = NULL;
-  JoinFeeder feeder = { NULL, 0 };
+  JoinFeeder feeder = {};
   GThread *thread;
   guint received = 0, n_pads = 0;
   gboolean released = FALSE, started = TRUE;
@@ -526,6 +553,7 @@ run_pad_release_while_streaming (void)
     started = FALSE;
 
   feeder.appsrc = appsrc_0;
+  feeder.received = &received;
   thread = g_thread_new ("join-feeder", join_feeder_thread, &feeder);
   if (!wait_pipeline_process_buffers (&received, 1, TEST_TIMEOUT_LIMIT_MS))
     started = FALSE;
@@ -536,6 +564,8 @@ run_pad_release_while_streaming (void)
 
   g_atomic_int_set (&feeder.stop, 1);
   g_thread_join (thread);
+  EXPECT_LE (feeder.max_lead, JOIN_FEEDER_LEAD);
+  EXPECT_GT (feeder.max_lead, 0U);
 
   g_object_get (join_handle, "active-pad", &active_pad, NULL);
   g_object_get (join_handle, "n-pads", &n_pads, NULL);


myungjoo:

## Scope

Reviewed the single-file diff (`tests/gstreamer_join/unittest_join.cc`, +31/-2). Verified against the repo's build (nnstreamer meson tree), the join element's C source (`gst/join/gstjoin.c`), and independent local rebuilds/timing/race measurements (native + valgrind memcheck), not just static reading of the diff.

## Does it solve the stated problem? Yes

- Native run of `join.padReleaseWhileStreaming`: measured 212–233 ms, matching the "213 ms, unchanged" claim.
- Memcheck run of the same case, before this patch (original unthrottled feeder, product code unmodified): 20 attempts completed in **21.3 s** on this sandbox (GST_STATES pipeline-count check confirms all 20 ran).
- Memcheck run after this patch: 20 attempts completed in **1.79 s** on the same machine — an ~12x reduction, in the same ballpark as the PR's reported 14.9 s→1.2 s (4/20→20/20). Absolute numbers differ from the author's, which is expected given the PR itself documents 19 s–830 s variance for this case across CI runs; the *direction and magnitude* of the fix reproduce cleanly.
- Whole-binary memcheck time for `unittest_join` (13 cases): 9.06 s here, close to the PR's claimed 8.8 s.

## Does it weaken or strengthen race detection? Independently confirmed: strengthens it

Reintroduced the ordering defect the test targets (moved `gst_pad_set_active`/`gst_element_remove_pad` in `gst_join_release_pad` to after `GST_JOIN_UNLOCK`, in a scratch/local build only — not proposed as a real change) and compared the original (pre-PR) unthrottled feeder against this PR's throttled one.

- Using the PR's own methodology (20 separate fresh process invocations, each running up to 20 internal attempts with `ASSERT_TRUE` early-exit): **0/20 detections for both variants** on this sandbox (WSL2/virtualized). This did not reproduce the "10/20 → 20/20" figures either direction.
- Because that surprised me, I re-tested with a higher-N methodology to get a stable signal: a scratch build of each variant's inner loop increased to 300 attempts inside one continuous process (no early-abort). Result, repeated 2–3x per variant for stability:
  - **Unthrottled (original): 12–16 failures / 300 attempts (~4–5%)**
  - **Throttled (this PR): 111–146 failures / 300 attempts (~37–49%)**

  That is an 8–10x higher per-attempt hit rate with the throttle than without it — the opposite of the concern that bounding the feeder's lead might starve the race window. This is a real, repeatable result on this machine, not a one-off.
- Caveat for the record: the fact that fresh, cold-started process runs (exactly how CI invokes this binary) showed 0 hits for either variant here, while a long-running warmed-up process showed a strong and consistent gap, indicates this specific race's manifestation is sensitive to process/scheduler warm-up state in this particular sandboxed environment (WSL2 + unshare). I cannot certify the literal "10/20 vs 20/20" numbers from a fresh-process methodology on this machine, but I can certify, independently, that the throttle mechanism increases rather than decreases the measured hit rate for the reintroduced defect, which is the substantive claim this PR rests on.

## Code correctness (the four items called out for scrutiny)

- **`guint` underflow**: `guint lead = feeder->pushed > taken ? feeder->pushed - taken : 0;` explicitly guards against `taken > pushed`; no underflow is reachable.
- **`g_atomic_int_get` on a `guint *`**: consistent with the existing pattern in this same codebase (`tests/unittest_util.h`'s `count_output`/`wait_pipeline_process_buffers` contract, which is documented as "read atomically"). Correct usage.
- **`pushed`/`max_lead` without atomics**: both are written only by the feeder thread and read by the main test thread only after `g_thread_join(thread)`, which establishes a happens-before edge. No data race.
- **`feeder.received = &received` ordering**: set before `g_thread_new()` starts the feeder, so there's no race on the pointer's initialization either.

No memory-safety or thread-safety defect found in the new code.

## `JOIN_FEEDER_LEAD` (8) and the `EXPECT_LE` assertion

- The value is not backed by a documented parameter sweep in the PR, but the empirical results above (both the timing improvement and the ~8–10x detection-rate improvement) suggest it is a reasonable working point, not obviously too small or too large.
- `EXPECT_LE (feeder.max_lead, JOIN_FEEDER_LEAD)` does exercise a real regression: tracing the feeder loop, `max_lead` is recorded immediately before the throttle decision, so if the cap were silently broken (e.g., the `if (lead >= JOIN_FEEDER_LEAD)` guard removed or miscompared), `max_lead` would grow unbounded and this assertion would fail. It is a legitimate self-check of the throttle, not a no-op.
- Because it's `EXPECT_LE` (non-fatal) inside a function called up to 20 times by the outer loop, a broken throttle could print up to ~20 gtest failure lines in one run — not "excessive" in absolute terms, and it does not prevent the outer `ASSERT_TRUE` loop from also reporting the primary race-detection failure. Not a real CI-log-noise concern.

## Minor, non-blocking observations

- The throttled feeder busy-waits via `g_thread_yield()` in a tight loop rather than blocking on a condition variable while waiting for the sink to catch up. Given the measured total runtime (hundreds of ms to ~2 s even under memcheck), this is not a measurable performance/CI-cost concern, just a minor style note.
- No other module, header, or public API is touched; the new struct fields and `#define` are local/static to this file. Ran the full 13-case `unittest_join` suite natively and under memcheck after all edits were reverted to the exact PR state — all 13 pass in both modes, so no regression to the other cases in this file.
- No ARCHITECTURE or API documentation update is needed — this is a test-only change to a private helper.
- `clang-format` (repo config) reports the changed file as clean, consistent with the PR's claim.

## Verdict suggestion: no blocking issues found

The change does what it says, is proportionate in size to the problem (test-only, +31/-2, no other module touched), and its central claim — that bounding the feeder's lead makes the race regression test both cheaper under memcheck and at least as effective at catching the targeted defect — held up under independent (if methodologically adapted) reproduction. The only qualifier worth keeping in mind is that the exact numeric ratios in the PR description are, by the PR's own admission, highly environment-dependent; readers should treat them as illustrative rather than portable guarantees, which does not affect the correctness of the change itself.

myungjoo:
Both notes acted on, and the more useful outcome is the one you did not raise as a finding: you could not reproduce my headline evidence, and you were right not to be able to.

**The detection numbers.** I had reported 10 of 20 runs before and 20 of 20 after. Those came from one measurement on a loaded box, and whole-run detection is the wrong statistic here — at roughly one attempt in 38, whether a 20-attempt run notices is a coin toss against machine load, which is exactly why you saw nothing either way. I re-measured with attempts-to-first-detection instead, five samples each, defect reintroduced:

| feeder | attempts to first detection | per attempt |
|---|---|---|
| before | 14, 64, 25, 32, 50 | ~1 in 38 |
| after | 1, 1, 2, 2, 4 | ~1 in 2 |

That is the same 8-10x you found by a different route, and it also explains the run-level figures rather than contradicting them: one in 38 over 20 attempts is about 42%, which is where my 10 of 20 came from. The PR body and the commit message now carry the attempt counts, with the run-level numbers described as derived and load-sensitive and your non-reproduction recorded.

**`JOIN_FEEDER_LEAD` = 8.** Swept, and the honest answer is that the value does not matter:

| lead | memcheck | attempts to first detection |
|---|---|---|
| 2 | 1074 ms | 3 |
| 4 | 1111 ms | 4 |
| 8 | 1136 ms | 2 |
| 16 | 1133 ms | 3 |
| 64 | 1178 ms | 3 |

No trend in either column across a 32x range, against 14.9 s and one-in-38 with no bound at all. So eight is a round number in a wide flat region, and the PR now says so rather than implying it was tuned.

**The `g_thread_yield` spin.** Left as is. It spins only while the sink is behind, which the numbers above say is a small part of the time — the throttled feeder still reaches its first detection in one or two attempts — and under valgrind, which is where this matters, threads are serialised anyway so a yield is the primitive that gives the pipeline the CPU back. A sleep would trade a real cost for a stylistic one.

No code changed since `09101f91`; the commit message and PR body did, so the head is now `ab44219f`.

myungjoo:

## Scope

Short confirmation pass on the wording-only update from `09101f91` to `ab44219f`, per the author's reply. Not a re-review of the code — that was already approved in the previous round (issuecomment-5580678742).

## 1. Code diff between the two heads

```
git diff 09101f91 ab44219f
```

is empty (`git diff --quiet` reports no difference, `git diff --stat` produces no output). Confirmed: nothing changed but the commit message and the PR body. The previously-approved code stands unmodified.

## 2. Did the wording changes address the prior round's three points?

Checked the current PR body and `git log -1 --format=%B ab44219f` against issuecomment-5580678742:

- **Detection-strength claim**: the old "10 of 20 / 20 of 20" whole-run figures are no longer presented as the headline evidence. Both the PR body and commit message now lead with attempts-to-first-detection samples (before: 14, 64, 25, 32, 50; after: 1, 1, 2, 2, 4), and explicitly demote the run-level numbers to "a poor thing to quote," framing them as derived from the per-attempt rate rather than as independent evidence. Confirmed.
- **Load-sensitivity / reviewer non-reproduction**: both the PR body ("a reviewer running the same control on this same machine saw no detection either way") and the commit message ("a reviewer on this same machine saw no detection either way") now record this explicitly, alongside the reviewer's own higher-N figures (~4-5% vs ~37-49% per attempt) being cited as agreeing with the new attempt-based numbers. Confirmed.
- **`JOIN_FEEDER_LEAD` = 8 not being a tuned value**: the new "The lead is not a tuned number" section gives the sweep (2/4/8/16/64 → 1074/1111/1136/1133/1178 ms, 3/4/2/3/3 attempts to first detection) and states there is no trend across the 32x range, i.e., 8 is "a round number in a wide flat region rather than a tuned one." Confirmed, and the range matches what was described in the task (~1074-1178 ms, 2-4 attempts).

All three requested edits are present in both the PR body and the commit message, consistently.

## 3. Arithmetic check of the 42% claim

The author's follow-up comment states: "one in 38 over 20 attempts is about 42%, which is where my 10 of 20 came from."

Checked: for a per-attempt hit probability p = 1/38 and n = 20 independent attempts,

P(at least one hit) = 1 - (1 - 1/38)^20 = 1 - (37/38)^20 ≈ **0.4134**, i.e. ≈ 41.3%.

This rounds sensibly to "about 42%" (off by <1pp), so the arithmetic is correct. It does not claim to reproduce the observed 10/20 = 50% exactly — the author is explicit that a single 20-attempt run at p≈1/38 is "close to a coin toss against machine load," and a binomial(20, 0.413) draw landing at 10 (vs. a mean of ~8.3) is unremarkable sampling noise, not a contradiction. No inconsistency found.

## 4. CI status on `ab44219f`

`gh pr checks 4946` reports all 19 checks passing (Analyze c-cpp/python, Android NDK, macOS/Meson, RISC-V, LLVM/Clang, Yocto, Valgrind x86_64 and arm, CPP-Linter, CodeQL, DCO, Spell Check, Static checks, 3x Tizen GBS, 2x Ubuntu pdebuild). No failures, no pending checks.

## Verdict: APPROVE

No code changed since the already-approved `09101f91`; the wording changes fully address all three points raised in the previous round, the new numbers are internally consistent, and all 19 CI checks are green on the current head (`ab44219f`). This PR is in a mergeable state from a review standpoint.

myungjoo:
Thank you for the CI A/B — 93,996 ms against 760 ms on the same infrastructure an hour apart is better evidence than anything I measured locally, and it is now in the PR body.

Of the three suggestions, two are in at `896ea62f` and one is measured and declined.

**Positional initialiser.** Agreed and changed to `JoinFeeder feeder = {};`. It was exactly the shape that goes quietly wrong when a field is added.

**`EXPECT_GT (feeder.max_lead, 0U)`.** Added. I checked it cannot flake before keeping it: `max_lead` is sampled at the top of each iteration, so a sink that kept up perfectly could in principle leave it at zero. It did not miss once in 30 native runs — up to 600 attempts — or in 3 runs under memcheck.

**`g_usleep (100)` in place of the yield.** Declined, on measurement. The CPU-contention argument is right in principle, but the sleep costs the race and buys nothing:

| wait | memcheck | native | attempts to first detection |
|---|---|---|---|
| `g_thread_yield ()` | 1.1 s | 211 ms | 5, 1, 1, 1, 1 — mean 1 |
| `g_usleep (100)` | 1.1 s | 211 ms | 11, 60, 17, 45, 3 — mean 27 |

Identical cost, and detection goes most of the way back to the 38 of the unbounded feeder. After each buffer the sink takes, the pad would sit idle for up to 100 µs, which is long next to the window this test aims at — the yield returns the moment the sink is ready. Both the measurement and the reason are now in the commit message, so the next person to have this idea does not have to re-derive it.

CI is running on the new head.

PR #4948, opened 2026-09-10
[Decoder] Refuse a stream the decoder cannot decode instead of dereferencing nothing

myungjoo:
Two NULL-pointer defects of the tensor_decoder path from #4920: **C22** in the element and **G5** in the `bounding_boxes` sub-plugin. They meet on the same path -- a sub-plugin that cannot take the stream -- so they are fixed together, one commit each.

## G5 -- the box decoding mode is used before option1 chooses one

`BoundingBox` keeps a `BoxProperties *` that `option1` looks up in a table, and every other option is handed to it. Nothing checked it, and it starts out as nothing:

| trigger | on `main` |
| --- | --- |
| `mode=bounding_boxes` with no `option1` | SIGSEGV in `getOutCaps ()` |
| `option1=wrong_mode_name` (SSAT case `11_n`) | SIGSEGV |
| `option5=...` / `option2=...` / `option3=...` ahead of `option1` | SIGSEGV in `setInputModelSize ()` / `setLabelPath ()` / `setOptionInternal ()` |
| `option1` set to an unknown name on a running pipeline | SIGSEGV on the next buffer |

`gst-launch-1.0` sets properties in the order they are written, so the option order alone is enough. The last row is a second way in: `setBoxDecodingMode ()` stored the lookup result *before* testing it.

The options now refuse, negotiation refuses while no mode has been chosen, and the mode is taken only when the lookup answered. `option2` and `option5` no longer describe themselves as independent from `option1`, in the header or the property description; `option4` and `option7` still are.

## C22 -- the element dereferenced the sub-plugin's refusal

`getOutCaps ()` returning NULL is how a sub-plugin says it cannot decode a config. The element passed that NULL to `gst_caps_intersect ()`, `gst_caps_is_always_compatible ()` and `gst_caps_unref ()`, which are guarded only by GLib's own checks, compiled out of a release build. Driving `direct_video` with a float32 tensor logged **6 GStreamer criticals**; now **0** and a clean `not-negotiated`.

- **Refusal.** The three sites that ask the sub-plugin directly fail the negotiation. `gst_tensordec_get_media_caps ()` stops answering ANY for a refusal -- only a caps query it could not *read* means "anything may come out" -- so a refused config offers empty caps and is turned away before fixation. `fixate_caps ()` keeps that empty out of its "keep othercaps" fallback, and refuses ANY instead of handing it to `gst_caps_fixate ()` (caps without dimensions to an any-caps downstream logged 8 criticals that way).
[...]

PR #4951, opened 2026-09-10
[Decoder] Bound the pose decoder to its tensors and refuse empty label files

myungjoo:
Items **G1** and **G13** of #4920, taken together because both live in `tensordec-pose.c`.

The PR also fixes a division by zero in the same decoder: a missing `option2` crashes with SIGFPE. It is the `pose_estimation` counterpart of G19, and #4920 does not list it. It is fixed and tested here and is not added to #4920 as a separate item; see **New — no model input size** below.

## What was wrong

**G1 — empty label file.** `loadImageLabels ()` (used by `image_labeling`, `bounding_boxes`, `tensor_region`) and the `pose_estimation` label loader both did

```c
if (!g_file_get_contents (path, &contents, &len, &err) || len <= 0)
  ml_loge ("... %s", ..., err->message);
```

Reading a 0-byte file *succeeds*, so `err` is NULL on the `len == 0` branch. Pointing any of the four decoders at an existing empty file crashed the process as soon as the option was set (SIGSEGV, reproduced on `main` for `image_labeling option1=`, `tensor_region option2=` and `pose_estimation option3=`). `contents` leaked on the same branch.

The next shape of an empty file behaved wrongly too. A file holding a single newline (what `echo > labels.txt` writes) was accepted by the pose loader as a table of **no** labels, and the table in use was dropped. `loadImageLabels ()` refused that file only because `g_new0 ()` of 0 elements returns NULL, and it logged that it had run out of memory. A blank line *inside* a pose label file had no first token: its label was copied from NULL (a GLib critical), and its connection count became `-1`.

**G13 — pose decoder reads past its tensors.** `pose_getOutCaps ()` checked only the first dimension:
- the heatmap is always read as float32, so a heatmap of any other type was read past its end. `tests/nnstreamer_decoder_pose` case 1 did exactly this and passed: a uint8 `14:14:1` heatmap, 784 bytes read out of a 196-byte tensor.
- in `heatmap-offset` mode the offset is looked up at the heatmap cell, but the offset tensor's grid was never compared with the heatmap's.
- a connection id equal to the number of labels passed `k > total_labels` and read `XYdata[total_labels]`, one past the end of the array (SIGSEGV, reproduced).
- giving `option3` again leaked the previous metadata.

**New — no model input size.** `pose_decode ()` divides by `option2`'s width and height, which default to 0. A pipeline without `option2` died with SIGFPE on the first frame (reproduced, exit 136).

## The fix

- G1: split the condition. A failed read reports its error as before. A file is refused when splitting it yields no line at all, which covers the 0-byte file and the single-newline file, and `contents` is released. The two loaders keep their existing failure semantics:
  - The pose loader keeps the table in use, as it already did for an unreadable file.
  - `loadImageLabels ()` clears its labels first, as it already did for an unreadable file. That is how its callers (`image_labeling`, `bounding_boxes`, `tensor_region`) see the refusal: they report success as `total_labels > 0`, after replacing their `label_path`. A blank line in a pose label file becomes a keypoint without a label or connections, so the keypoint ids of the lines after it stay where the file puts them.
- G13 and the model size: one helper, `pose_check_input ()`, decides whether decoding the tensors stays inside them: model size non-zero, heatmap float32 with one row per label, offset tensor float32 with `2 × labels` and the heatmap's grid, rank ≤ 3. In `heatmap-offset` mode the heatmap also needs at least two cells in each direction, because a keypoint is placed at `cell / (cells - 1)`, which is 0/0 for a single cell. `pose_getOutCaps ()` refuses what it rejects. `pose_decode ()` calls it again, because `option2`/`option3` set while streaming change what gets read without a renegotiation (same reasoning as the `BoundingBox::decode ()` guard added for G19). The existing dimension checks move into the helper as plain checks rather than `g_return_val_if_fail ()`, which disappears with `G_DISABLE_CHECKS` and reports a user's configuration error as a programming error.
- `>` becomes `>=` in both places that bound a label id. A reloaded label file frees the previous metadata, but only once the new file has been read into lines, so a refused file leaves the table in use untouched.
- SSAT `nnstreamer_decoder_pose` case 1 (the uint8 heatmap) becomes `1_n`, because that case is the defect.

## Behaviour changes

Each of these was a crash or an out-of-bounds read before. None is reachable by a pipeline that decoded correctly:
- a label file without a label (0 bytes or a single newline) is refused instead of crashing. For `pose_estimation` the labels in use are kept, where a single-newline file used to wipe them.
- `pose_estimation` refuses a non-float32 heatmap, a mismatched offset grid, a single-cell heatmap in `heatmap-offset` mode, and a missing `option2`. The file header already documents both tensors as float32. It marks `option3`/`option4` as optional but not `option2`.
- A refused `pose_estimation` config now logs `GST_ERROR` instead of a GLib critical.

## Tests

New gtest binary `unittest_decoder_pose` (20 cases). It calls the sub-plugin directly where a single guard has to be pinned, and runs a `filesrc` → `tensor_converter` → `tensor_decoder` → `filesink` pipeline where the drawn frame is what tells cases apart. That is the same fixture shape as `unittest_decoder_boundingbox`, and it is built unconditionally.

- **Negative control** (pre-fix sources, new tests): 11 of 20 fail.
  - By SIGSEGV: `refuseEmptyLabelFile_n` ×2 and `skipConnectionPastLastLabel_n`.
  - By SIGFPE: `decodeAfterModelSizeCleared_n` and the pipeline `refuseMissingModelSize_n`.
  - By assertion: `refuseUint8Heatmap_n`, `refuseOffsetGridMismatch_n`, `refuseSingleCellOffsetGrid_n`, `refuseMissingModelSize_n`, `decodeAfterLabelsChanged_n`, and `loadLabelFileWithBlankLine`. The last one counts GLib criticals, with a self-check that proves the counter is live.
  - Every positive case passes on both sides.
- **Mutation control**: each guard removed on its own fails at least one case. Measured for each of these:
  - the decode-time check, `>=`, and the type check
  - each grid dimension separately, each tensor's half of the rank check, and the single-cell grid
  - the model size
  - the pose no-label check and the blank-line handling
  - moving the metadata release ahead of the refusals: `refuseEmptyLabelFile_n` starts from a table loaded from a file, so the move turns it into a double free (SIGABRT).

  Three removals are not visible to a gtest assertion, stated rather than papered over:
  - dropping the release itself is a leak, and the memcheck gate does not fail on leaks.
  - without the no-label check in `loadImageLabels ()`, the file is still refused, because `g_new0 ()` of 0 elements returns NULL. Only the log message changes.
  - without the `len > 0` guard, a 0-byte file is read one byte before its buffer, which only the memcheck step sees.
- The refused-config pipeline case logs the `gst_caps_*` criticals of item **C22**, fixed separately in #4948. It does not depend on that fix.

Local runs (WSL, Ubuntu 22.04):
- `unittest_decoder_pose` 20/20. Each intermediate commit also builds with `-Dwerror=true` and passes SSAT `nnstreamer_decoder_pose`.
- SSAT `nnstreamer_decoder_pose` 7/7, `nnstreamer_decoder_boundingbox` 55/55, `nnstreamer_decoder_tensor_region` 2/2, `nnstreamer_decoder` 60/60.
- `meson test`: 23/24. The failure is `unittest_filter_python3`, from this machine's NumPy 2 ABI mismatch, unrelated.
- A strict `-Wextra -Wsign-compare -Werror` compile of every changed object is clean. gst-indent reproduces the C hunks exactly, clang-format 16 is clean on the test file, and `doxygen-tag.sh` passes with ctags (checked with a negative control).

## Not in this PR

Options are set without any lock while the streaming thread decodes (the same class as C32/B14). With the metadata now released on reload, an `option3` set *during* a `decode ()` call can free what `draw ()` is reading, where it used to leak instead. That race was already there for `total_labels`, which `decode ()` reads without a lock. Closing it needs the element to serialise options against `decode ()`, which is a change for all decoders, not for this one.

Refs #4920

### DIFF
--- a/ext/nnstreamer/tensor_decoder/tensordec-pose.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-pose.c
@@ -71,6 +71,7 @@
  *	                    	(e.g., 17 x 9 x 9 )
  *   			Tensor[1]: #labels x 2: width : height (float32, Offset position within heatmap grid)
  *	                    	(e.g., 34 x 9 x 9 )
+ *   			Width and height have to be 2 or larger, and the same in both tensors.
  *
  * Pipeline:
  * 	v4l2src
@@ -279,6 +280,9 @@ pose_load_metadata_from_file (pose_data * pd, const gchar * file_path)
     return FALSE;
   }
 
+  if (pd->metadata != pose_metadata_default)
+    g_free (pd->metadata);
+
   pd->total_labels = g_strv_length (lines);
   pd->metadata = g_new0 (pose_metadata_t, pd->total_labels);
 
@@ -320,7 +324,7 @@ pose_get_metadata_by_id (pose_data * data, guint id)
 {
   pose_metadata_t *md = data->metadata;
 
-  if (id > data->total_labels)
+  if (id >= data->total_labels)
     return NULL;
 
   return &md[id];
@@ -460,6 +464,61 @@ _check_tensors (const GstTensorsConfig * config)
 }
 
 /**
+ * @brief Check the input tensors against what pose_decode () reads from them.
+ * @param[in] data The pose-estimation internal data.
+ * @param[in] config The configuration of the input tensors.
+ * @return TRUE if decoding the tensors stays within them, otherwise FALSE.
+ */
+static gboolean
+pose_check_input (const pose_data * data, const GstTensorsConfig * config)
+{
+  const GstTensorInfo *heatmap = &config->info.info[0];
+  const GstTensorInfo *offset = &config->info.info[1];
+  gboolean with_offset = (data->mode == HEATMAP_OFFSET);
+  guint i;
+
+  if (data->i_width == 0 || data->i_height == 0) {
+    GST_ERROR
+        ("The input video dimension of the model is zero. Set option2 of pose_estimation to the WIDTH:HEIGHT the model takes.");
+    return FALSE;
+  }
+
+  if (heatmap->type != _NNS_FLOAT32
+      || heatmap->dimension[0] != data->total_labels) {
+    GST_ERROR
+        ("The heatmap tensor of pose_estimation has to be float32 with the %u labels as its first dimension.",
+        data->total_labels);
+    return FALSE;
+  }
+
+  if (with_offset && (heatmap->dimension[1] < 2 || heatmap->dimension[2] < 2)) {
+    GST_ERROR
+        ("The heatmap of pose_estimation in heatmap-offset mode needs at least two cells in each direction.");
+    return FALSE;
+  }
+
+  if (with_offset && (config->info.num_tensors < 2
+          || offset->type != _NNS_FLOAT32
+          || offset->dimension[0] != 2 * data->total_labels
+          || offset->dimension[1] != heatmap->dimension[1]
+          || offset->dimension[2] != heatmap->dimension[2])) {
+    GST_ERROR
+        ("The offset tensor of pose_estimation has to be float32 with two values per label for each cell of the heatmap.");
+    return FALSE;
+  }
+
+  for (i = 3; i < NNS_TENSOR_RANK_LIMIT; i++) {
+    if (heatmap->dimension[i] > 1 || (with_offset && offset->dimension[i] > 1)) {
+      GST_ERROR
+          ("The tensors of pose_estimation cannot have more than three dimensions.");
+      return FALSE;
+    }
+  }
+
+  return TRUE;
+}
+
+/**
  * @brief tensordec-plugin's TensorDecDef callback
  *
  * [Pose Estimation]
@@ -473,30 +532,10 @@ pose_getOutCaps (void **pdata, const GstTensorsConfig * config)
 {
   pose_data *data = *pdata;
   GstCaps *caps;
-  int i;
   char *str;
-  guint pose_size;
 
-  const uint32_t *dim;
-
-  if (!_check_tensors (config))
+  if (!_check_tensors (config) || !pose_check_input (data, config))
     return NULL;
-
-  pose_size = data->total_labels;
-
-  /* Check if the first tensor is compatible */
-  dim = config->info.info[0].dimension;
-  g_return_val_if_fail (dim[0] == pose_size, NULL);
-  for (i = 3; i < NNS_TENSOR_RANK_LIMIT; i++)
-    g_return_val_if_fail (dim[i] <= 1, NULL);
-
-  if (data->mode == HEATMAP_OFFSET) {
-    dim = config->info.info[1].dimension;
-    g_return_val_if_fail (dim[0] == (2 * pose_size), NULL);
-
-    for (i = 3; i < NNS_TENSOR_RANK_LIMIT; i++)
-      g_return_val_if_fail (dim[i] <= 1, NULL);
-  }
 
   str = g_strdup_printf ("video/x-raw, format = RGBA, " /* Use alpha channel to make the background transparent */
       "width = %u, height = %u", data->width, data->height);
@@ -725,7 +764,7 @@ draw (GstMapInfo * out_info, pose_data * data, GArray * results)
     for (j = 0; j < smd->num_connections; j++) {
       guint k = smd->connections[j];
       /* Have we already drawn the connection ? */
-      if ((k > data->total_labels) || (k < i))
+      if ((k >= data->total_labels) || (k < i))
         continue;
       /* Is the body point valid ? */
       if (XYdata[k]->valid == FALSE)
@@ -757,6 +796,11 @@ pose_decode (void **pdata, const GstTensorsConfig * config,
   guint pose_size, index;
 
   g_assert (outbuf); /** GST Internal Bug */
+
+  /* An option may have changed since pose_getOutCaps () approved the stream */
+  if (!pose_check_input (data, config))
+    return GST_FLOW_ERROR;
+
   /* Ensure we have outbuf properly allocated */
   if (gst_buffer_get_size (outbuf) == 0) {
     out_mem = gst_allocator_alloc (NULL, size, NULL);
--- a/ext/nnstreamer/tensor_decoder/tensordec-pose.c
+++ b/ext/nnstreamer/tensor_decoder/tensordec-pose.c
@@ -262,16 +262,23 @@ pose_load_metadata_from_file (pose_data * pd, const gchar * file_path)
     return FALSE;
   }
 
-  if (!g_file_get_contents (file_path, &contents, &len, &err) || len <= 0) {
+  if (!g_file_get_contents (file_path, &contents, &len, &err)) {
     ml_loge ("Unable to read file %s with error %s.", file_path, err->message);
     g_clear_error (&err);
     return FALSE;
   }
 
-  if (contents[len - 1] == '\n')
+  if (len > 0 && contents[len - 1] == '\n')
     contents[len - 1] = '\0';
 
   lines = g_strsplit (contents, "\n", -1);
+  if (g_strv_length (lines) == 0) {
+    ml_loge ("The label file %s has no label.", file_path);
+    g_strfreev (lines);
+    g_free (contents);
+    return FALSE;
+  }
+
   pd->total_labels = g_strv_length (lines);
   pd->metadata = g_new0 (pose_metadata_t, pd->total_labels);
 
@@ -282,6 +289,11 @@ pose_load_metadata_from_file (pose_data * pd, const gchar * file_path)
     g_strstrip (lines[i]);
     tokens = g_strsplit (lines[i], " ", -1);
     n_tokens = g_strv_length (tokens);
+    if (n_tokens == 0) {
+      /* A blank line is a keypoint without a label or connections */
+      g_strfreev (tokens);
+      continue;
+    }
     if (n_tokens > POSE_MD_MAX_CONNECTIONS_SZ) {
       GST_WARNING ("Too many connections (%d) declared, clamping (%d)\n",
           n_tokens, POSE_MD_MAX_CONNECTIONS_SZ);
--- a/ext/nnstreamer/tensor_decoder/tensordecutil.c
+++ b/ext/nnstreamer/tensor_decoder/tensordecutil.c
@@ -34,17 +34,22 @@ loadImageLabels (const char *label_path, imglabel_t * l)
   _free_labels (l);
 
   /* Read file contents */
-  if (!g_file_get_contents (label_path, &contents, &len, &err) || len <= 0) {
+  if (!g_file_get_contents (label_path, &contents, &len, &err)) {
     ml_loge ("Unable to read file %s with error %s.", label_path, err->message);
     g_clear_error (&err);
     return;
   }
 
-  if (contents[len - 1] == '\n')
+  if (len > 0 && contents[len - 1] == '\n')
     contents[len - 1] = '\0';
 
   _labels = g_strsplit (contents, "\n", -1);
   l->total_labels = g_strv_length (_labels);
+  if (l->total_labels == 0) {
+    ml_loge ("The label file %s has no label.", label_path);
+    goto error;
+  }
+
   l->labels = g_new0 (char *, l->total_labels);
   if (l->labels == NULL) {
     ml_loge ("Failed to allocate memory for label data.");
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -112,6 +112,15 @@ if gtest_dep.found()
     )
     test('unittest_decoder_boundingbox', unittest_decoder_boundingbox, env: testenv)
 
+    # RUN unittest_decoder_pose
+    unittest_decoder_pose = executable('unittest_decoder_pose',
+      join_paths('nnstreamer_decoder_pose', 'unittest_decoder_pose.cc'),
+      dependencies: [nnstreamer_unittest_deps, nnstreamer_internal_deps],
+      install: get_option('install-test'),
+      install_dir: unittest_install_dir
+    )
+    test('unittest_decoder_pose', unittest_decoder_pose, env: testenv)
+
     # Run unittest_plugins
     unittest_plugins = executable('unittest_plugins',
       join_paths('nnstreamer_plugins', 'unittest_plugins.cc'),
--- a/tests/nnstreamer_decoder_pose/runTest.sh
+++ b/tests/nnstreamer_decoder_pose/runTest.sh
@@ -26,8 +26,8 @@ CASEEND=1
 # THIS SHOULD EMIT ERROR
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc ! videoconvert ! videoscale ! video/x-raw,width=640,height=480,format=RGB ! tensor_converter ! tensor_split name=a tensorseg=1:640:480:1,2:640:480:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=640:480 ! fakesink" 0_n 0 1 $PERFORMANCE
 
-# THIS WON'T FAIL, BUT NOT MUCH MEANINGFUL.
-gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=4 ! videoconvert ! videoscale ! video/x-raw,width=14,height=14,format=RGB ! tensor_converter ! tensor_split name=a tensorseg=1:14:14:1,2:14:14:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=14:14 ! fakesink" 1 0 0 $PERFORMANCE
+# THIS SHOULD EMIT ERROR: the heatmap is uint8, and the decoder reads it as float32
+gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=4 ! videoconvert ! videoscale ! video/x-raw,width=14,height=14,format=RGB ! tensor_converter ! tensor_split name=a tensorseg=1:14:14:1,2:14:14:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=14:14 ! fakesink" 1_n 0 1 $PERFORMANCE
 
 # TEST WITH MORE BUFFERS
 gstTest "--gst-plugin-path=${PATH_TO_PLUGIN} videotestsrc num_buffers=20 ! videoconvert ! videoscale ! video/x-raw,width=14,height=14,format=RGB ! tensor_converter ! tensor_transform mode=arithmetic option=typecast:float32,add:128,div:255 ! tensor_split name=a tensorseg=1:14:14:1,2:14:14:1 a.src_0 ! tensor_transform mode=transpose option=1:2:0:3 ! tensor_decoder mode=pose_estimation option1=320:240 option2=14:14 ! fakesink" 2 0 0 $PERFORMANCE
--- a/tests/nnstreamer_decoder_pose/unittest_decoder_pose.cc
+++ b/tests/nnstreamer_decoder_pose/unittest_decoder_pose.cc
@@ -1,0 +1,736 @@
+/**
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * @file        unittest_decoder_pose.cc
+ * @date        10 Sep 2026
+ * @brief       Unit test for the pose_estimation mode of tensor_decoder and
+ *              the label file loader of the image-based decoder modes
+ * @see         https://github.com/nnstreamer/nnstreamer
+ * @author      MyungJoo Ham <myungjoo.ham@samsung.com>
+ * @bug         No known bugs
+ */
+
+#include <gtest/gtest.h>
+#include <glib.h>
+#include <gst/gst.h>
+#include <string.h>
+#include <unittest_util.h>
+
+#include <nnstreamer_plugin_api_decoder.h>
+#include <nnstreamer_plugin_api_util.h>
+
+/* The number of labels pose_estimation assumes without a label file */
+#define DEFAULT_LABELS (14U)
+
+#define LABELS (2U)
+#define GRID (5U)
+#define HEATMAP_ELEMENTS (LABELS * GRID * GRID)
+#define OFFSET_ELEMENTS (2U * HEATMAP_ELEMENTS)
+
+#define OUT_SIZE (40U)
+#define OUT_PIXELS (OUT_SIZE * OUT_SIZE)
+
+#define POSE_PIXEL (0xFFFFFFFFU)
+
+/** The two labels, connected to each other */
+#define LABEL_TEXT "a 1\nb 0\n"
+/** The first label connects to an id one past the last label */
+#define LABEL_TEXT_PAST_LAST "a 2\nb\n"
+
+/**
+ * @brief Test fixture holding an instance of the pose_estimation decoder.
+ */
+class tensorDecoderPose : public ::testing::Test
+{
+  protected:
+  const GstTensorDecoderDef *decoder;
+  void *pdata;
+  gchar *label_file;
+
+  /**
+   * @brief Initialize the decoder instance.
+   */
+  void SetUp () override
+  {
+    pdata = NULL;
+    label_file = NULL;
+    decoder = nnstreamer_decoder_find ("pose_estimation");
+    ASSERT_TRUE (decoder != NULL);
+    ASSERT_TRUE (decoder->init (&pdata));
+  }
+
+  /**
+   * @brief Release the decoder instance and the label file.
+   */
+  void TearDown () override
+  {
+    if (pdata != NULL)
+      decoder->exit (&pdata);
+    removeTempFile (&label_file);
+  }
+
+  /**
+   * @brief Write a label file and hand it to the decoder as option3.
+   */
+  gboolean setLabels (const gchar *text)
+  {
+    removeTempFile (&label_file);
+    label_file = getTempFilename ();
+    if (label_file == NULL || !g_file_set_contents (label_file, text, -1, NULL))
+      return FALSE;
+
+    return decoder->setOption (&pdata, 2, label_file);
+  }
+
+  /**
+   * @brief Whether the decoder negotiates an output for the given input.
+   */
+  gboolean accepts (const GstTensorsConfig *config)
+  {
+    GstCaps *caps = decoder->getOutCaps (&pdata, config);
+
+    if (caps == NULL)
+      return FALSE;
+
+    gst_caps_unref (caps);
+    return TRUE;
+  }
+};
+
+/**
+ * @brief Describe a heatmap tensor, and an offset tensor if offset_grid is not 0.
+ */
+static void
+setPoseConfig (GstTensorsConfig *config, tensor_type type, guint labels,
+    guint grid, guint offset_grid)
+{
+  GstTensorInfo *info;
+
+  gst_tensors_config_init (config);
+  config->rate_n = 0;
+  config->rate_d = 1;
+  config->info.num_tensors = (offset_grid > 0) ? 2 : 1;
+
+  info = gst_tensors_info_get_nth_info (&config->info, 0);
+  info->type = type;
+  info->dimension[0] = labels;
+  info->dimension[1] = grid;
+  info->dimension[2] = grid;
+  info->dimension[3] = 1;
+
+  if (offset_grid > 0) {
+    info = gst_tensors_info_get_nth_info (&config->info, 1);
+    info->type = type;
+    info->dimension[0] = 2 * labels;
+    info->dimension[1] = offset_grid;
+    info->dimension[2] = offset_grid;
+    info->dimension[3] = 1;
+  }
+}
+
+/**
+ * @brief Decode one heatmap, allocated to its exact size, with the given decoder.
+ */
+static GstFlowReturn
+decodeHeatmap (const GstTensorDecoderDef *decoder, void **pdata, const GstTensorsConfig *config)
+{
+  GstTensorMemory input;
+  GstBuffer *outbuf = gst_buffer_new ();
+  GstFlowReturn ret;
+
+  input.size = gst_tensor_info_get_size (&config->info.info[0]);
+  input.data = g_malloc0 (input.size);
+
+  ret = decoder->decode (pdata, config, &input, outbuf);
+
+  g_free (input.data);
+  gst_buffer_unref (outbuf);
+  return ret;
+}
+
+/**
+ * @brief Set one cell of a heatmap tensor (label, x, y).
+ */
+static void
+setHeatmap (float *heatmap, guint label, guint x, guint y, float value)
+{
+  heatmap[(y * GRID + x) * LABELS + label] = value;
+}
+
+/**
+ * @brief Write the given floats to a new temp file.
+ * @return the file name, to be released with removeTempFile ()
+ */
+static gchar *
+writeTensorFile (const float *data, guint elements)
+{
+  gchar *name = getTempFilename ();
+
+  if (name != NULL
+      && !g_file_set_contents (name, (const gchar *) data, elements * sizeof (float), NULL))
+    removeTempFile (&name);
+
+  return name;
+}
+
+/**
+ * @brief Run the pipeline to its end.
+ * @return GST_MESSAGE_EOS or GST_MESSAGE_ERROR, GST_MESSAGE_ANY if the
+ *         pipeline could not be built and GST_MESSAGE_UNKNOWN on a timeout.
+ */
+static GstMessageType
+runPipeline (const gchar *pipeline_str)
+{
+  GstElement *pipeline = gst_parse_launch (pipeline_str, NULL);
+  GstMessageType type = GST_MESSAGE_UNKNOWN;
+  GstBus *bus;
+  GstMessage *msg;
+
+  if (pipeline == NULL)
+    return GST_MESSAGE_ANY;
+
+  /* A refused negotiation may fail the state change itself */
+  setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT);
+
+  bus = gst_element_get_bus (pipeline);
+  msg = gst_bus_timed_pop_filtered (bus, 10 * GST_SECOND,
+      (GstMessageType) (GST_MESSAGE_EOS | GST_MESSAGE_ERROR));
+  if (msg != NULL) {
+    type = GST_MESSAGE_TYPE (msg);
+    gst_message_unref (msg);
+  }
+  gst_object_unref (bus);
+
+  setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT);
+  gst_object_unref (pipeline);
+
+  return type;
+}
+
+/**
+ * @brief Decode a heatmap (and its offsets, if given) into a frame.
+ * @param[in] heatmap HEATMAP_ELEMENTS floats
+ * @param[in] offset OFFSET_ELEMENTS floats for heatmap-offset, NULL for heatmap-only
+ * @param[in] label_text the content of the label file
+ * @param[in] model_size option2 of the decoder
+ * @param[out] frame OUT_PIXELS RGBA pixels drawn by the decoder
+ */
+static GstMessageType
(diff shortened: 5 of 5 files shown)


myungjoo:
**This is an automated review by a separate Claude agent, relayed to this PR by the author's own account (myungjoo) at the author's request.**

Scope reviewed: the diff of this PR only (`ext/nnstreamer/tensor_decoder/tensordec-pose.c`, `ext/nnstreamer/tensor_decoder/tensordecutil.c`, `tests/meson.build`, `tests/nnstreamer_decoder_pose/runTest.sh`, `tests/nnstreamer_decoder_pose/unittest_decoder_pose.cc`), against the rest of the repository and against issue #4920 (items G1, G13, and the undocumented pose `option2`-is-zero division by zero).

## What I checked

- Traced `pose_check_input()`'s bound math by hand against `pose_decode()`'s reads: with the new offset-grid-matches-heatmap-grid check, the maximum `offset[offsetIdx + pose_size]` index works out to exactly `2 * pose_size * grid_x * grid_y - 1`, i.e. the last valid element of the offset tensor — the check is tight, not just "probably enough."
- Confirmed the `k > total_labels` → `k >= total_labels` fix in `draw()` (tensordec-pose.c:754) actually closes the described one-past-the-end read of `XYdata[k]`, including for a negative connection id read from the label file (wraps to a huge `guint`, still caught by `>=`).
- Verified the type check catches the uint8-heatmap-read-as-float32 case, and that the added `option2`-zero check is reached from both `pose_getOutCaps()` and `pose_decode()` (the latter re-checks because an option can change mid-stream without renegotiation, matching the `BoundingBox::decode()` precedent cited in the commit message).
- Checked the callers of `loadImageLabels()` (`tensordec-imagelabel.c:95`, `tensordec-tensor_region.c:346`, `tensordec-boundingbox.cc:935`) — all three already gate on `total_labels > 0` after the call, so the "callers already treat 'no label loaded' as a refused option" claim in the commit message holds.
- Built and ran the actual PR head in the WSL tree provided for this review (source `md5sum`-verified identical to this PR's `unittest_decoder_pose.cc`): `unittest_decoder_pose` is 17/17 green.
- Ran `tools/development/gst-indent` on both changed `.c` files in a scratch copy of the tree: it reformats several **pre-existing, untouched** blocks (the `pose_metadata_default` initializer, `is_value_within()`, `setpixel()`, the `_check_tensors()` cast) but produces **zero diff** on any line this PR actually added or touched (`pose_check_input()`, the split empty-file checks, the `>=` fixes). So the PR's own new/changed code is gst-indent-clean; the reported diffs are pre-existing style debt unrelated to this change.
- Ran `clang-format` (v14, not the v16 the PR body cites, but agrees) on the new test file: zero diff.
- Could not run SSAT (`ssat`/`ssat-api.sh` isn't installed in the available WSL tree) or valgrind (unavailable per the review instructions), so the SSAT-suite and mutation/memcheck claims in the PR body are taken on faith beyond the equivalent gtest coverage, which I did run.
- Confirmed no other SSAT script in the tree invokes `mode=pose_estimation` without `option2`, so the new "refuse without model size" behavior cannot break an existing passing case.
- Confirmed no `Documentation/` file mentions `pose_estimation`, and the sub-plugin's own header comment already lists `option3`/`option4` as "(optional)" but not `option2` — so no doc update is missing for the new "option2 is now mandatory" behavior.

## Findings

**[nit] Asymmetric behavior on a failed label reload — `tensordecutil.c:34` vs `tensordec-pose.c:280`**
`loadImageLabels()` calls `_free_labels(l)` unconditionally at the top (tensordecutil.c:34, unchanged by this PR) before attempting to read the new file. So when the *new* file turns out to be empty (the new branch this PR adds at tensordecutil.c:43-47), any previously-loaded, working labels for `image_labeling`/`bounding_boxes`/`tensor_region` are discarded, and the decoder is left with zero labels going forward — not merely "this one `setOption()` call is refused." The sibling fix in this same PR, `pose_load_metadata_from_file()`, does the opposite: it frees the old `pd->metadata` only *after* the new file is validated (tensordec-pose.c:271-281), so pose_estimation keeps its previously-working labels on a failed reload. Neither behavior is a crash or a leak (both are safe), and the commit message anticipates the `loadImageLabels()` behavior ("the callers already treat 'no label loaded' as a refused option"), so this is not a defect — just an inconsistency between two fixes for the same tracking item (G1) landing in the same PR. Worth a one-line note in the commit message or a `g_free`-after-validate reorder for symmetry, but not worth blocking on.

**[non-blocking] No test exercises the ">3 dimensions" branch of the new `pose_check_input()` helper — `tensordec-pose.c:497-503`**
Every other branch of the new helper (float32 type, label-count match, offset-grid match on each axis separately, `option2` zero) has a dedicated positive and negative gtest case. The rank guard (`for (i = 3; i < NNS_TENSOR_RANK_LIMIT; i++) ...`) does not — `setPoseConfig()` in the test file always sets `dimension[3] = 1` and never exercises a 4D tensor. This branch is carried over from the pre-existing `g_return_val_if_fail` code (not new logic), so the risk is low, but per the PR's own stated testing philosophy ("each guard removed on its own fails at least one case") this guard is the one exception without that coverage. A `refuseRank4Tensor_n` case mirroring `refuseOffsetGridMismatch_n` would close the gap and protect against a future refactor silently dropping this check.

## Not flagged (verified correct, no action needed)

- The claimed division-by-zero fix, the OOB fixes for G13, and the empty-file NULL-deref fix for G1 are all real, correctly bounded, and matched by a passing negative-control test in the PR body's own methodology (I independently reran the positive-control side).
- No unrelated modules are touched; the `tensordecutil.c` change is the minimum needed since `loadImageLabels()` is genuinely shared by three other decoders under the same G1 item.
- No CI gap: the negative cases (`_n` suffix, gtest death/error-return cases) fail loudly (SIGSEGV/SIGFPE/assertion pre-fix per the PR body, `GST_FLOW_ERROR`/refused negotiation post-fix), so a regression that reopens any of these guards will fail `meson test` / SSAT, not just log a warning.

## Verdict

No blocking issues found. The fix is correctly scoped to items G1 and G13 of #4920 plus the disclosed `option2`-zero division by zero, the bound-safety math checks out by hand, the new code is clean against gst-indent/clang-format, and the new gtest binary passes 17/17 against this exact PR head. The two points above are non-blocking (a documentation nit and a minor test-coverage gap) — **mergeable as-is**; addressing them as a fast-follow (or even in a later PR) is fine.

myungjoo:
_Response to the relayed review above, posted on behalf of the author by a Claude agent._

Thanks. Both points were checked against the code.

**[non-blocking] rank guard has no dedicated test.** Agreed, fixed in-PR. `refuseHigherRank_n` gives the heatmap, and then (in `heatmap-offset` mode) the offset tensor alone, a 4th dimension of 2 and expects the config to be refused. Mutation check: dropping the heatmap half and dropping the offset half of the loop's condition each make that case fail (the heatmap-only half of the test cannot be satisfied by any other check, and in the second half the heatmap is well-formed). Folded into the test commit, now `e4f933d7`; 18/18 locally, clang-format 16 and `doxygen-tag.sh` clean.

**[nit] `loadImageLabels ()` clears the labels on a failed reload while the pose loader keeps them.** Kept as is, on purpose. The three callers of `loadImageLabels ()` report success as "labels are loaded":

- `il_setOption ()`: `return data->labels.total_labels > 0`
- `tr_setOption ()`: `return trData->labeldata.total_labels > 0`
- `setLabelPath ()` (bounding_boxes): `if (labeldata.total_labels > 0) ... return TRUE`

They have also already replaced `label_path` with the new path. If a failed reload kept the old labels, those callers would report a refused file as accepted, under the new path's name. Clearing is what makes the refusal visible to them. It is also what `main` already did for a missing or unreadable file (`_free_labels ()` runs before the read), so the empty file now joins that path instead of crashing. `pose_load_metadata_from_file ()` returns its own `FALSE` and never touched the metadata before a failure on `main` either. Each loader keeps the failure semantics its callers were built on, and neither changed in this PR beyond not crashing.

myungjoo:
**This is an automated review by a separate Claude agent (second round), relayed to this PR by the author's own account (myungjoo) at the author's request.**

Scope: independent re-review of the whole PR (not just the delta since `adbee723`) against the rest of the repository and issue #4920 (items G1, G13, and the disclosed `option2`-zero division by zero, cross-checked against G19, the analogous fix in `bounding_boxes`).

## Verification performed

- Re-derived `pose_check_input()`'s bound math against `pose_decode()` by hand; confirmed the fixes for G1 (`tensordecutil.c`/`tensordec-pose.c`: NULL-`GError` read on a 0-byte label file) and G13 (`k > total_labels` → `k >= total_labels`, float32-only heatmap read, offset-grid-matches-heatmap-grid) all close the reads/derefs the issue describes, matching the text of G1/G13/G19 in #4920 verbatim.
- Confirmed `pose_load_metadata_from_file()` now also fixes an unrelated pre-existing leak: reloading `option3` (setting a second, valid label file) used to leak the previous `pd->metadata` allocation; the new `if (pd->metadata != pose_metadata_default) g_free (pd->metadata);` before the reassignment (tensordec-pose.c:279-280) closes it. Not mentioned in the commit message but harmless and correct.
- Built and ran the exact PR head (`e4f933d7`) in the WSL tree provided for this review: `unittest_decoder_pose.cc` is byte-identical (md5) to this PR's file, `tensordec-pose.c`/`tensordecutil.c` are identical modulo CRLF. `unittest_decoder_pose` is **18/18 green** (including the `refuseHigherRank_n` case added since the last round). `ssat tests/nnstreamer_decoder_pose` is **7/7 green**, including case `1_n`, the uint8-heatmap-is-now-a-refusal case that the previous round could not exercise (SSAT wasn't installed there yet).
- Re-read the previous round's two points and the author's response:
  - **Rank-guard test gap**: the added `refuseHigherRank_n` does close it correctly. Traced both halves by hand: the heatmap-only half sets `dimension[3]=2` on tensor 0 while mode is still `HEATMAP_ONLY`, which only the rank loop's `heatmap->dimension[i] > 1` term can catch (type and label-count checks both still pass); the heatmap-offset half sets `dimension[3]=2` on tensor 1 only, with tensor 0 left well-formed, which only the `with_offset && offset->dimension[i] > 1` term can catch. Agreed, correctly closes the gap.
  - **`loadImageLabels()` clearing labels on a failed reload vs. pose keeping them**: verified all three callers (`tensordec-imagelabel.c:95`, `tensordec-tensor_region.c:346`, `tensordec-boundingbox.cc:935`) do overwrite `label_path`/`labeldata` to the new path before calling `loadImageLabels()` and report success via `total_labels > 0` afterward — the author's reasoning holds: keeping the old labels on a failed reload would make those callers report the new (bad) path as accepted while silently still using the old labels. Agreed, this is a deliberate and correct asymmetry, not a defect.

## New finding (not raised in the previous round)

**[non-blocking] `pose_check_input()` does not reject a heatmap/offset grid dimension of 0 or 1 in `heatmap-offset` mode — tensordec-pose.c:466-506, consumed at tensordec-pose.c:844-845**

`pose_check_input()` requires `offset->dimension[1] == heatmap->dimension[1]` and `offset->dimension[2] == heatmap->dimension[2]` (tensordec-pose.c:490-491), but never requires those grid dimensions to be `> 1`. In `HEATMAP_OFFSET` mode, `pose_decode()` computes:

```c
posX = (((gfloat) maxX) / (grid_xsize - 1)) * data->i_width + offsetX;
posY = (((gfloat) maxY) / (grid_ysize - 1)) * data->i_height + offsetY;
```

(tensordec-pose.c:844-845), where `grid_xsize`/`grid_ysize` come straight from `config->info.info[0].dimension[1]`/`[2]`. A heatmap/offset grid of exactly 1 makes `grid_xsize - 1 == 0`: `((gfloat) maxX) / 0` is a float division by zero (produces `inf`/`nan`, not a trap), which then feeds `p.x = MIN (data->width, (guint) (MAX (0, p.x)))` — casting a non-finite float to `guint` is undefined behavior in C. A grid of exactly 0 is a further step: `offset->dimension[1]==0` makes the offset tensor's byte size (`2*labels*0*0*4`) zero, and `offsetIdx = (maxY*grid_xsize+maxX)*pose_size*2+index` still evaluates to `index` (0..2*labels-1), so `offset[offsetIdx]`/`offset[offsetIdx + pose_size]` (tensordec-pose.c:840-841) read past a zero-sized tensor. `pose_check_input()` doesn't reject dimension 0 or 1 on either axis today (only the >3-rank guard and the offset-matches-heatmap-grid equality are checked).

This is exactly the class of bug — and exactly the function — this PR was written to close (`pose_check_input()`'s own doc comment: "Check the input tensors against what pose_decode() reads from them"), so it's in scope for this PR rather than a pre-existing issue to defer. That said: I could not confirm real reachability end-to-end through ordinary caps negotiation (a 0- or 1-cell spatial grid is not something a real pose model would emit, and I did not verify whether `gst_tensors_config_from_structure()`/caps parsing already rejects such a dimension before it reaches the decoder — I did not have time to trace that path or build a live repro), so I'm not asserting this crashes today, only that the new guard function doesn't defend against it the way it defends against every other bound `pose_decode()` relies on.

Suggested fix: in the `with_offset` branch, also require `heatmap->dimension[1] > 1 && heatmap->dimension[2] > 1` (only the offset path divides by `grid - 1`; heatmap-only mode never uses the grid as a divisor, so it doesn't need this). A `refuseSingleCellGrid_n` test mirroring `refuseOffsetGridMismatch_n` would pin it down and confirm/deny reachability at the same time.

## Not re-flagged (already covered and correct)

- No unrelated modules touched; `tensordecutil.c` is genuinely shared by `image_labeling`/`tensor_region`/`bounding_boxes`, all three already gate on `total_labels > 0`.
- Doxygen present on the new helper and every new test case; commit messages carry DCO sign-off and follow the `[Component] Summary` convention; no ARCHITECTURE/API surface changed (option2 was already documented as mandatory, without an "(optional)" tag, before this PR).
- CI wiring is real: `unittest_decoder_pose` is registered in `tests/meson.build` inside the existing `gtest_dep.found()` block, and the SSAT case that used to be a no-op ("THIS WON'T FAIL, BUT NOT MUCH MEANINGFUL") is now a proper `_n` negative case — both independently confirmed green against this exact PR head above.

## Verdict

No blocking issues. The two points from the previous round are correctly and sufficiently addressed (the new `refuseHigherRank_n` test genuinely closes the rank-guard gap by mutation-style reasoning, verified by hand; the `loadImageLabels()` asymmetry is a deliberate, correct design choice, not a bug). The one new finding above (grid dimension of 0 or 1 in heatmap-offset mode) is non-blocking — narrow, likely unreachable through normal model output, and not confirmed as a live crash — but is a genuine gap in the exact bound-checking function this PR introduces, and would be a cheap fast-follow (one extra condition plus one test case). **Mergeable as-is.**

myungjoo:
_Response to the second relayed review above, posted on behalf of the author by a Claude agent._

**[non-blocking] grid of 0 or 1 in heatmap-offset mode.** I measured two separate cases.

- **Grid 1:** real, and fixed in this PR. `maxX / (grid_xsize - 1)` is `0.0f / 0` = NaN, and converting NaN to `int` for `p.x`/`p.y` is undefined behaviour. The later `MAX`/`MIN` clamps it to the frame on x86 and AArch64, so it is not a memory error, but a single-cell grid cannot be decoded in offset mode at all. `pose_check_input ()` now refuses a heatmap with fewer than two cells in either direction in `heatmap-offset` mode. `heatmap-only` never divides by the grid, so it is left alone. The new `refuseSingleCellOffsetGrid_n` pins the width, the height, and the `heatmap-only` exemption; disabling the new check makes it fail.
- **Grid 0 (a rank-1 heatmap):** no over-read. The dimension array ends at the first 0, so the offset tensor that passes the grid comparison has `2 × labels` elements. With `grid_xsize == 0` the search loops do not run, `maxX = maxY = 0`, and the reads are `offset[index]` and `offset[index + labels]`, both below `2 × labels`. The divisor is `0 - 1 = -1`, which is not a division by zero. The single-cell check above now refuses this case as well in offset mode.

Force-pushed as `a2b40db2`. `git diff e4f933d7 a2b40db2` is +6 lines in `pose_check_input ()` and the new test. Local results: `unittest_decoder_pose` 19/19, SSAT `nnstreamer_decoder_pose` 7/7, strict `-Wextra -Wsign-compare -Werror` compile clean, gst-indent/clang-format 16/`doxygen-tag.sh` clean.

Also confirmed: the reviewer's re-check of the two first-round points matches what the code does.

myungjoo:
**This is an automated review by a separate Claude agent (third round), relayed to this PR by the author's own account (myungjoo) at the author's request.**

Scope: the delta since the last round (`e4f933d7` -> `a2b40db2`, the new single-cell-grid guard in `pose_check_input ()` and its test), re-checked against the whole PR, the rest of the repository, and issue #4920 (G1, G13, plus the disclosed `option2`-zero division-by-zero).

## Verification performed

- Re-derived the divisor math by hand: `pose_decode ()` uses `grid_xsize = config->info.info[0].dimension[1]` and `grid_ysize = config->info.info[0].dimension[2]` (tensordec-pose.c:816-817) as `grid_xsize - 1` / `grid_ysize - 1` divisors (tensordec-pose.c:850-851), only inside `if (data->mode == HEATMAP_OFFSET)` (tensordec-pose.c:830, 843). The new guard at tensordec-pose.c:487-491 tests exactly `heatmap->dimension[1] < 2 || heatmap->dimension[2] < 2`, gated by the same `with_offset = (data->mode == HEATMAP_OFFSET)` (tensordec-pose.c:470) that governs the division. The guard is tight to what it defends and does not touch `heatmap-only` mode, which never divides by the grid.
- Checked reachability of the two variants raised last round:
  - **Grid = 1**: real UB (NaN from `x/0.0f` cast to `guint`), now refused. Confirmed by re-deriving the check placement.
  - **Grid = 0 (rank-1 heatmap)**: independently re-verified the author's claim that this is *not* an over-read, by reading `gst_tensor_get_element_count ()` (gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:1268-1282): the element-count loop `break`s at the first `dimension[i] == 0` rather than multiplying by it, so an offset tensor described as `2*labels : 0 : 0 : 1` actually has `2*labels` elements, not zero. With `grid_xsize == 0` the search loops (`for (j = 0; j < grid_ysize; j++)`) never execute, `maxX = maxY = 0`, and `offsetIdx` reduces to `index` (0..2*labels-1) — in bounds. This closes the loop on that half of the previous finding; the author's response was correct.
  - Confirmed no existing SSAT case or the PR's own new/old positive gtest cases use a heatmap-offset grid of 1 anywhere (`tests/nnstreamer_decoder_pose/runTest.sh` uses 14x14/17x17/320x240; `GRID` in the gtest file is 5) — the new refusal cannot reject anything that decoded correctly before.
- Built the exact PR head (`a2b40db2`) in the provided WSL tree and ran it directly (not just re-used a cached binary): `unittest_decoder_pose` **19/19 green**, including `refuseSingleCellOffsetGrid_n`; SSAT `nnstreamer_decoder_pose` **7/7 green**, including case `1_n`.
- Attempted an independent mutation rebuild (remove the new guard, recompile only the pose decoder plugin, rerun the test) to mechanically confirm `refuseSingleCellOffsetGrid_n` fails without it. This surfaced a **build-environment artifact, not a PR problem**: the experiment copy's `build.ninja` had a regenerate step that silently re-pointed compilation back at the shared, do-not-edit `~/nns-g1` tree instead of the copy, so the mutated source was never actually compiled (verified by `strings` on the rebuilt `.so` still containing the guard's message after a from-scratch touch+recompile). I did not chase a clean reconfigure further (gtest is found only through cached, non-pkg-config state in that tree) given the marginal value: the code-path elimination argument is airtight without it — for the exact test input in `refuseSingleCellOffsetGrid_n`, every other branch of `pose_check_input ()` (type, label count, offset-grid-equality, rank) evaluates true, so only the new branch can produce the `EXPECT_FALSE`/`EXPECT_TRUE` outcomes observed, and the third sub-case (`heatmap-only`, grid 1, `EXPECT_TRUE`) is a genuine positive assertion against the real, unmutated PR binary confirming the mode gate works. This is disclosed for transparency rather than presented as a completed mutation test.
- Re-confirmed the two points from the first round and the "grid 0/1" point from the second round all remain correctly resolved; nothing new to add there.
- Checked `Documentation/` again (still no file mentions `pose_estimation`) and the file's own header doc-comment block (tensordec-pose.c:57-73, listing `option4`'s `heatmap-offset` tensor shapes) — it is not updated to state the new "at least two cells in each direction" requirement. Minor.
- Confirmed DCO sign-off and `[Component] Summary` convention on all three commits (`eddd0cee`, `c351c177`, `a2b40db2`).

## New findings

**[nit] The file's own doc-comment for `heatmap-offset` mode does not mention the new two-cell-per-axis minimum — tensordec-pose.c:61-73**
The comment block documents `option4`'s expected tensor shapes with example dimensions (9x9, 17x17) that already satisfy the new guard, so no example is wrong, but the constraint itself ("Tensor[0]/Tensor[1] must be at least 2x2 in `heatmap-offset` mode") isn't stated anywhere a reader of the header would see it. Trivially fixable in this PR: one line under the existing `heatmap-offset:` bullet.

No other new findings. I did not re-flag the rank-guard test gap or the `loadImageLabels ()` asymmetry from round 1, or the grid-0 half of round 2's finding — all three are settled and the responses are correct.

## Verdict

No blocking issues. The delta since the last round (`e4f933d7` -> `a2b40db2`) is exactly what the author described: a 6-line guard that is provably tight to the one divide-by-zero path it closes, gated correctly so `heatmap-only` mode and every previously-passing case (real models, all existing SSAT/gtest positive cases) are unaffected, backed by a test that separately pins the x-axis, y-axis, and the `heatmap-only` exemption. Verified independently by hand-tracing the bound math, by reading `gst_tensor_get_element_count ()` to confirm the "grid 0" half of the prior finding is genuinely safe, and by a real build+run of the PR head (19/19 unit tests, 7/7 SSAT). **Mergeable as-is.** One nit remains that could be trivially fixed in this PR (the doc-comment update above); it is not blocking and does not need to hold up the merge.

myungjoo:
_Response to the third relayed review above, posted on behalf of the author by a Claude agent._

**[nit] the file header does not state the heatmap-offset grid requirement.** Fixed in this PR. The `heatmap-offset` block of the `tensordec-pose.c` header now ends with "Width and height have to be 2 or larger, and the same in both tensors.", which covers the single-cell refusal and the grid comparison. The change is a comment line only. `git diff a2b40db2 86840eb9` is that one line; the head is now `86840eb9`, and gst-indent and `doxygen-tag.sh` are clean.

On the mutation check you could not complete: the local measurement disabled the new condition, rebuilt, and ran the suite. Only `refuseSingleCellOffsetGrid_n` failed, and the other 18 cases passed. I also confirmed that the shared `~/nns-g1` tree is still byte-identical to the PR head after your experiment.

myungjoo:
**This is an automated review by a separate Claude agent (fourth round), relayed to this PR by the author's own account (myungjoo) at the author's request.**

Scope: independent re-review of the whole PR (all three commits: `eddd0cee`, `7463e129`, `86840eb9`), with particular attention to the delta since the last round (`a2b40db2` -> `86840eb9`), against the rest of the repository and issue #4920 (G1, G13, plus the disclosed `option2`-zero division-by-zero).

## Verification performed

- Confirmed the delta since round 3 is exactly what the author described: `git diff a2b40db2 86840eb9` is one added line in the file's own header comment block (`tensordec-pose.c:74`): "Width and height have to be 2 or larger, and the same in both tensors." It sits under the existing `heatmap-offset:` bullet, directly above `Pipeline:`, and correctly documents the `heatmap->dimension[1] < 2 || heatmap->dimension[2] < 2` guard added in round 2. Note the line physically lands inside commit `7463e129` (the functional fix commit) rather than the test commit `86840eb9` when diffed against history before the force-push that produced `a2b40db2` — i.e. the author folded the doc line into the commit it documents rather than appending a fixup, which is the right place for it per `Documentation/contributing.md`'s "single topic per commit."
- Independently re-derived `pose_check_input()`'s bound math against `pose_decode()` by hand, not just re-read the prior rounds' derivations: heatmap type/label-count check, the `with_offset` grid->=2 guard (`tensordec-pose.c:478-482`) against the `grid_xsize - 1` / `grid_ysize - 1` divisors at `tensordec-pose.c:850-851`, the offset-grid-equals-heatmap-grid check, and the rank loop. All four map cleanly to what `pose_decode()` reads, and each is exercised by its own test in `unittest_decoder_pose.cc`.
- Checked reachability of every guard from both of pose_estimation's `TensorDecDef` entry points that touch `config`: `pose_getOutCaps()` and `pose_decode()` both call `pose_check_input()` (`tensordec-pose.c:531`, `:795`); the third registered callback, `pose_getTransformSize()` (`tensordec-pose.c:545-555`), is a no-op that ignores `config` and returns 0, so there is no fourth path that reads the tensors unguarded.
- Confirmed short-circuit safety of the new helper: `offset = &config->info.info[1]` is taken unconditionally at the top of the function, but `offset->type`/`offset->dimension[...]` are only dereferenced inside `with_offset && (config->info.num_tensors < 2 || ...)`, so a 1-tensor `heatmap-offset` config is rejected by the `num_tensors < 2` term before any offset field is read (`GstTensorsInfo.info[]` is a fixed-size array, so `&config->info.info[1]` itself is always valid storage — only the field reads would be a problem, and they don't happen out of order).
- Rebuilt and ran the exact PR head in the provided WSL tree (`~/nns-g1`, incremental `ninja -C build tests/unittest_decoder_pose`, no stale objects): **19/19 green**, including `refuseEmptyLabelFile_n` and `refuseSingleCellOffsetGrid_n`.
- Ran `tools/development/gst-indent` on both changed `.c` files in the WSL tree: it reformats the same pre-existing, untouched blocks as round 1 reported (the `pose_metadata_default` initializer, `is_value_within()`, `setpixel()`, the `_check_tensors()` cast, a comment-spacing nit) and **zero diff on any line this PR added or touched** — `pose_check_input()`, the split empty-file checks, the `>=` fixes, and the new header-comment line are all gst-indent-clean.
- Confirmed `tests/meson.build:115-122` registers `unittest_decoder_pose` in the same unconditional `if gtest_dep.found()` block as `unittest_decoder_boundingbox`, with no `flatbuf_support_is_available` gate — so, unlike `tests/nnstreamer_decoder/unittest_decoder.cc`, this binary does build and run on the standard Ubuntu/meson CI legs, not just pdebuild/GBS.
- Re-checked `tests/nnstreamer_decoder_pose/runTest.sh`: case `1` (heatmap declared/read as `uint8` but decoded as `float32`) is correctly relabeled `1_n` with the expect-fail arguments `0 1`, replacing the old "THIS WON'T FAIL, BUT NOT MUCH MEANINGFUL" case — this exercises the same type guard the gtest suite pins more precisely (`GST_FLOW_ERROR`/refused negotiation, not just "some nonzero exit"), so the SSAT case is a reasonable secondary check, not the only evidence for this guard.
- Re-read all three previous rounds' findings and the author's responses against the current code: the rank-guard test, the `loadImageLabels()`-vs-pose-loader asymmetry, and the grid-0/grid-1 distinction are all still correctly resolved on this head; nothing to reopen.

## New findings

None. I did not find any new issue, blocking or otherwise, in this round's delta or in a fresh pass over the whole PR.

## Verdict

**Mergeable as-is.** The round-3 nit (missing doc-comment line for the new heatmap-offset grid minimum) is correctly and minimally fixed — a single comment line, placed in the commit it documents, verified gst-indent-clean and consistent with the guard it describes. My own independent re-derivation of the bound math, a live rebuild/run of the exact PR head (19/19), and a check of all three `TensorDecDef` entry points found nothing outside what the previous three rounds already covered and the author already addressed. No nit remains that would need a trivial fix in this PR.

myungjoo:
_Response to the `myungjoo-bot` review of `86840eb9`, posted on behalf of the author by a Claude agent._

Both Low items were fixed in this PR, together with the blank-line case the review listed as a follow-up candidate. The new head is `cec23303`, still three commits.

**1. A single-newline file was accepted and emptied the pose decoder's labels.** Both loaders now split first and refuse the file when splitting yields no line at all. That one check covers the 0-byte file and the `"\n"` file, so the separate `len == 0` branch is gone. The newline strip is guarded by `len > 0` instead.
- The pose loader releases the previous table only after that refusal.
- `loadImageLabels ()` now logs "has no label" instead of the misleading "Failed to allocate memory" message.
- Both `refuseEmptyLabelFile_n` tests now include a `"\n"` case. On `86840eb9`, the pose one fails at `setLabels ("\n")`.

**2. The test could not see where the release sits.** Taken as you suggested: `refuseEmptyLabelFile_n` now loads `LABEL_TEXT`, is refused twice, and then expects a 2-label config to be accepted and a 14-label one to be refused. Mutation check: moving the release above `g_file_get_contents ()` now aborts the case with SIGABRT, a double free in `TearDown ()`.

**Blank line inside a label file.** This was listed as out of scope, but it lives in the function this PR changes, so it is closed here rather than left for #4920. An empty token vector now leaves that line as a keypoint with an empty label and no connections. That keeps the ids of the following lines where the file puts them. `loadLabelFileWithBlankLine` counts GLib-domain criticals with a self-check that the handler is live: one critical on `main`, none now.

**Options set while decoding.** Agreed. This stays disclosed in the PR body under "Not in this PR", because it needs the element to serialise options against `decode ()` for every decoder.

Checked locally on `cec23303`:
- `unittest_decoder_pose` 20/20; the pre-fix sources fail 11 of 20.
- SSAT `nnstreamer_decoder_pose` 7/7.
- The strict `-Wextra -Wsign-compare -Werror` compile, gst-indent, clang-format 16, `doxygen-tag.sh` (with ctags) and `newline.sh` are all clean.
- Each intermediate commit builds with `-Dwerror=true` and passes SSAT pose.

Commit 2's message no longer says the option2 defect will be reported to #4920 as a new item. It is fixed here and not added to the issue.

myungjoo:
**This is an automated review by a separate Claude agent, relayed to this PR by the author's own account (myungjoo) at the author's request.**

Scope: the fifth review round. Head is `cec23303` (previous reviewed head `86840eb9`). I read the issue thread, all four relayed agent reviews, and the formal `myungjoo-bot` review (submitted against `86840eb9`), then checked whether `git diff 86840eb9 cec23303` actually answers that review's two Low findings, and re-read the whole PR independently.

## What changed since `86840eb9`

Only `pose_load_metadata_from_file()` (tensordec-pose.c) and `loadImageLabels()` (tensordecutil.c) changed, plus the new test file gained a `"\n"` case and reworked `refuseEmptyLabelFile_n`. `pose_check_input()` and everything G13-related is untouched, so I did not re-derive those bounds again (the bot review already traced them against `pose_decode()` and I have no new concerns there).

The change replaces the old `len == 0` check with "split first, refuse if the split yields zero lines," and moves the `g_free (pd->metadata)` / release of the previous table to *after* that refusal.

## Verification of the two previously-open findings

**Bot finding 1 (Low): a `"\n"`-only file was accepted as a table of 0 labels.**
Fixed. I compiled a small GLib test program against the system glib (`g_strsplit`) to check the exact semantics rather than trust memory: `g_strsplit("", "\n", -1)` returns a valid, non-NULL, *zero-length* vector (`g_strv_length() == 0`), not a one-element vector of `""`. Since the code now strips the file's own trailing `\n` first and then splits, a `"\n"`-only file collapses to `""` before the split, so `g_strv_length(lines) == 0` correctly triggers the refusal in both `pose_load_metadata_from_file()` (tensordec-pose.c:274-280) and `loadImageLabels()` (tensordecutil.c:47-50). Confirmed against the new test `tensorDecoderPose.refuseEmptyLabelFile_n` (unittest_decoder_pose.cc:470-486), which now asserts `EXPECT_FALSE (setLabels ("\n"))` in addition to `""`, and confirmed against `tensorDecoderLabelFile.refuseEmptyLabelFile_n` (unittest_decoder_pose.cc:689-711) for the `loadImageLabels()` path.

**Bot finding 2 (Low): the test couldn't tell whether the metadata release happened before or after the refusal check, and reordering it back would double-free.**
Fixed, and now covered. `pose_load_metadata_from_file()` now frees `pd->metadata` only after the split-and-refuse check (tensordec-pose.c:283-284), and `refuseEmptyLabelFile_n` was rewritten to load `LABEL_TEXT` from a real file first, then feed it two refused files (`""` and `"\n"`), then assert the *original* 2-label config still decodes. Because the table at `TearDown()` now comes from a file (not the static `pose_metadata_default`), reintroducing the old bug (freeing before the check) would double-free on `decoder->exit()`. I traced `pose_exit()` (tensordec-pose.c:362-368) and confirmed it only frees when `metadata != pose_metadata_default`, so the current sequence (two refusals leaving the file-loaded table in place, then a clean single free at exit) is correct as written.

Both fixes also correctly handle the blank-line case now: for a blank line, `g_strsplit(lines[i], " ", -1)` also returns a zero-length vector (confirmed with the same test program), so `tokens[0]` would be NULL; the new `n_tokens == 0` guard (tensordec-pose.c:296-300) intercepts this before the `g_strlcpy (..., tokens[0], ...)` that used to raise a GLib critical on NULL. `loadLabelFileWithBlankLine` (unittest_decoder_pose.cc:503-523) verifies this with a live GLib critical-counting handler plus a self-check that the counter itself works — a good pattern, since a handler that silently isn't installed would otherwise make the test pass for the wrong reason.

## New finding (non-blocking)

**[Non-blocking] `loadImageLabels()` still drops the labels in use on a refused reload, unlike the pose loader — pre-existing, but the PR text/tests don't say so**
`ext/nnstreamer/tensor_decoder/tensordecutil.c:34`

`loadImageLabels()` calls `_free_labels (l)` unconditionally as its first statement, before the file is even read. This line is untouched by this PR (confirmed with `git diff main cec23303 -- tensordecutil.c`: it's identical on `main`), so this is not a regression introduced here. But it means the "table in use is kept" framing in the PR body's G1 section, and the general shape of the fix, only actually holds for `pose_load_metadata_from_file()`. For `image_labeling` / `bounding_boxes` / `tensor_region`, giving a bad reload (`""` or `"\n"`) still wipes out a previously-valid label table, same as on `main`.

This is easy to miss because the new `tensorDecoderLabelFile.refuseEmptyLabelFile_n` test (unittest_decoder_pose.cc:689-711) doesn't mirror the pose fixture's structure — it never loads a valid file first, so it can't observe whether the old table survives a refusal. It only checks that the bad option is refused and that a later good file still works.

Suggested fix (not necessarily in this PR): either soften the PR/commit wording so it's clear "table kept on refusal" is pose-specific, or as a small follow-up, move `_free_labels (l)` in `loadImageLabels()` to after the split-and-validate step (mirroring what this PR just did for the pose loader) and add a same-shaped "load, then refuse, then confirm the old labels are still in effect" test for one of its three callers. I'd treat this as a candidate follow-up item under #4920 rather than something that should block this PR — it doesn't regress anything, and fixing it properly touches three call sites (`tensordec-imagelabel.c`, `tensordec-boundingbox.cc`, `tensordec-tensor_region.c`) which would grow this PR's footprint beyond G1/G13.

## Other checks

- Commit structure: three commits (G1 fix, G13 fix, tests), each single-topic, each builds and DCO-signs off independently per the PR body's local-run claims. No issue.
- Scope: diff stays inside `tensordec-pose.c`, `tensordecutil.c`, and the pose test/build files; no unrelated modules touched.
- Docs: the file's own option-list header comment already marks option3/option4 "(optional)" and not option2, so no doc update was needed for the new option2-required enforcement; `Documentation/` has no pose_estimation-specific page to update.
- Test/CI gating: `unittest_decoder_pose` is registered unconditionally inside the `gtest_dep.found ()` block in `tests/meson.build`, so a future regression in these guards fails CI (meson test) directly, not just SSAT. Mutation coverage claims in the PR body match what I verified for the two findings above.

## Verdict

Mergeable as-is. No blocking or non-blocking-must-fix issues found in this round; both findings from the prior `myungjoo-bot` review are correctly fixed and now covered by tests. The one new item above is a nit/low-priority documentation-accuracy + test-coverage-symmetry gap in pre-existing, untouched code (`loadImageLabels()`'s unconditional `_free_labels()`) — it does not need to block this PR, and I would not call it "trivially fixable in this PR" since a real fix touches three additional decoder call sites and is better scoped as its own follow-up under #4920.

myungjoo:
_Response to the fifth relayed review above, posted on behalf of the author by a Claude agent._

**[nit] "the table in use is kept" reads as if it applied to `loadImageLabels ()` too.** The wording was wrong, and the PR body is corrected; the code is not changed. The G1 bullet now says:
- the pose loader keeps its table on a refusal;
- `loadImageLabels ()` clears its labels first, as it already did on `main` for an unreadable file.

The behaviour-change bullet now limits "labels in use are kept" to `pose_estimation`.

The asymmetry itself is intentional and stays, for the reason given in the first round (see the response to the first relayed review). The three callers of `loadImageLabels ()` replace `label_path` and then report success as `total_labels > 0`, so clearing is how a refusal reaches them. Keeping the old labels would make a refused file look accepted, under the new path's name. No #4920 follow-up is needed for it.

`tensorDecoderLabelFile.refuseEmptyLabelFile_n` also checks that the decoder accepts a valid file after the two refusals, which is the part of that contract a caller depends on.

Issue #4928, opened 2026-09-04
tensor_filter: a flexible sink pad lets a mismatched element type reach the model unchecked


myungjoo:
## What happens

With a flexible sink pad, `tensor_filter` feeds a buffer to the subplugin without ever checking that the buffer's element type is the one the model takes. A same-width substitution runs to completion and returns reinterpreted bits with a success code.

Reproduced on LiteRT with a float32 model, fed S32LE:

```bash
gst-launch-1.0 audiotestsrc num-buffers=3 ! audio/x-raw,format=S32LE,rate=16000,channels=4 ! tensor_converter frames-per-tensor=1 ! other/tensors,format=flexible ! tensor_filter framework=litert model=dynamic_batch_add_one.tflite ! fakesink
```

This ends with `Execution ended` and no error, identically to the F32LE control. The model reads int32 bits as float32.

## Why nothing catches it

For a flexible pad, `_gst_tensor_filter_convert_meta()` calls `gst_tensor_meta_info_convert()`, which copies the buffer's own meta header into `prop->input_meta` verbatim — `type` included:

https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/tensor_filter/tensor_filter.c#L694

The framework's subsequent size check is then derived from that same refreshed meta, so it agrees with the buffer rather than with the model. Caps negotiation does not stand in either: a flexible pad is exactly the arrangement where the per-buffer header, not the caps, decides the tensor's description.

That leaves the subplugin as the only possible checker, and subplugins are not told they need to be.

## Scope

This looks framework-level rather than LiteRT-specific: any subplugin reachable through a flexible sink pad inherits it, since the gap is in how `input_meta` is refreshed rather than in any one framework's invoke. LiteRT is only where it was noticed. Someone should confirm the reach across subplugins before choosing where the fix belongs — the framework refusing a type the subplugin did not report would fix it once for everyone, while a per-subplugin check repeats the same code N times and leaves the ones nobody edits still exposed.

Dimensions may deserve the same question. The size check catches a shape change that alters the byte count, but not one that preserves it.

## Not a regression

Pre-existing on `main`. Noticed while reviewing #4911, which adds `invoke-dynamic` to the LiteRT subplugin; that PR closes the hole for its own new dynamic path (`rejectTypeChange()`) but deliberately does not change the static path, since a behavioural change there is out of its topic and wants its own tests.

## Suggested test

[...]

PR #4953, opened 2026-09-10
[Filter] Validate the flexible input header and the dynamic output info

myungjoo:
Fixes items **B6**, **B9** and **B7** of #4920, all in `gst/nnstreamer/tensor_filter/tensor_filter.c`.

## B6: flexible input header used before it is validated

`_gst_tensor_filter_convert_meta()` handled each flexible input memory like this:

- It parsed a meta header without looking at the memory's size. `gst_tensor_meta_info_parse_header()` reads 84 bytes before validating anything, so a memory shorter than a header was read past its end.
- It ignored the parse result. The header size still came back as 0 or 128, so `size - hsize` underflowed for a short memory. A memory whose header didn't parse went to the model as raw data whenever its whole size matched.
- It rewrote `prop->input_meta[i]` from the header with `gst_tensor_meta_info_convert()`, which re-initialises the entry. The size check that follows reads its expected size from that same entry, so it compared the buffer with **its own header** instead of the model. A well-formed flexible tensor smaller than the model's input reached a sub-plugin that reads the model's size from it (e.g. tensorflow-lite's zero-copy input).
- The re-initialisation also dropped the configured tensor name without freeing it. openvino's `invoke()` builds `std::string (info->name)` from it.

Refreshing `input_meta` per buffer can't simply be removed, as the issue text suggested ("convert into a local copy"). It is the **contract of the dynamic invoke**, added in 99aaa0df: litert's `invoke_dynamic()` and custom-easy's dynamic callback read each buffer's shape from `prop->input_meta`. So:

| | before | after |
|---|---|---|
| `invoke-dynamic=true` | `input_meta[i]` replaced by the header (name lost) | type and dimension taken from the header, name kept |
| `invoke-dynamic=false` | same as above | `input_meta` untouched; the data is checked against the configured (model) tensor |

The second row is what `gst_tensor_filter_configure_tensor()` already says transform does for a flexible input ("Need to compare buffer size in transform()"). Every in-tree pipeline that sends variable-sized flexible tensors to a filter uses `invoke-dynamic` (llamacpp, llama2c, executorch-llama, litert, custom-easy).

A flexible memory is now refused unless all of these hold:
1. It is at least a header long before anything is parsed (mirrors `gst_tensor_meta_info_parse_memory()`).
2. The header parses.
3. Its version can be sized (a header can validate while `get_header_size()` returns 0, which is item A7).
4. It does not declare a sparse tensor. The dense conversion drops the format, so a sparse payload would be read as dense.
5. Its dense size is exactly the data after the header.

The sparse refusal applies with and without `invoke-dynamic`. A refused header posts a `STREAM/WRONG_TYPE` element error from tensor_filter, like its caps-time refusals, before the buffer fails with `GST_FLOW_ERROR`.

Without `invoke-dynamic` a flexible tensor is now checked against the model's size. The existing size-mismatch message used to suggest `format=flexible` alone; it now names `invoke-dynamic=true`.

The `input_meta` doxygen in `nnstreamer_plugin_api_filter.h` now says which behaviour a sub-plugin gets.

## B9: dynamic invoke output info not validated

With `invoke-dynamic` the sub-plugin fills `output_meta` during invoke, and `_gst_tensor_filter_transform_update_outbuf()` ignored the result of `gst_tensor_info_convert_to_meta()`. For an invalid entry that function returns before initialising the meta, so the header was built from an uninitialised stack struct. Inside the loop that is the previous output's header, and the buffer went downstream with the tensor missing or described by another tensor's header. The entry is now validated first. On failure, the output data not yet in the buffer is released and the buffer fails: `GST_FLOW_ERROR` from `transform()`, and `-1` from the async output callback, which also unrefs the partial buffer.

## B7: set_caps mismatch leak

A one-line `gst_tensors_config_free()` on the static-output mismatch branch of `gst_tensor_filter_set_caps()`. **No test:** the branch is a consistency check behind negotiation. `transform_caps()` derives the output caps from the same config, and `gst_tensor_caps_update_dimension()` only replaces a dimension string with an equivalent one. A leak also can't fail a CI check (the valgrind gate counts memcheck errors only).

## Tests

New `tensorFilterFlexInput.*` cases in `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`. They drive `tensor_filter framework=custom-easy` through a GstHarness on a flexible sink pad and push hand-built headers. A bus attached to the harness lets each header refusal assert the `STREAM/WRONG_TYPE` error from the filter. The model-size refusal (`smallerThanModel_n`) and the positive case assert that no such error was posted.

Result of each case against `main`'s `tensor_filter.c` (the new tests are unchanged):

| case | on main |
|---|---|
| `shortMemory_n` (8-byte memory ending at a `PROT_NONE` page) | **SIGSEGV** |
| `brokenHeaderMagic_n` (no header, whole memory = model size) | accepted, fails |
| `brokenHeaderType_n` (header with an invalid type) | accepted, fails |
| `smallerThanModel_n` | model invoked with 16 of its 176 bytes, fails |
| `sparseHeader_n` | accepted as dense, fails |
| `unknownVersion_n` (`0xDE002000`, header counted as data) | accepted, fails |
| `keepsConfiguredInfo` | name `NULL` inside invoke, fails |
| `dynamicTakesEachHeader` | name `NULL`, fails |
| `dynamicInvalidOutputInfo_n` | broken buffer pushed with `GST_FLOW_OK`, fails |
| `asyncInvalidOutputInfo_n` (mock V1 sub-plugin, `invoke-async`) | broken buffer pushed from the async callback, fails |
| `headerSizeMismatch_n` | refused, but no element error, fails |

On main `headerSizeMismatch_n` was refused as well, so its flow result alone doesn't discriminate. It pins the new consistency check: removing that check makes it fail.

Mutation controls, each removing one check of the fix:

| check removed | caught by |
|---|---|
| size check | `shortMemory_n` (SIGSEGV) |
| unsizable-version check | `unknownVersion_n` |
| sparse check | `sparseHeader_n` |
| sparse check limited to non-dynamic mode | `sparseHeader_n` (dynamic half) |
| consistency check | `headerSizeMismatch_n` |
| refreshing without invoke-dynamic | `smallerThanModel_n`, `keepsConfiguredInfo` |
| not refreshing with it | `dynamicTakesEachHeader`, `dynamicInvalidOutputInfo_n` |
| B9 check | `dynamicInvalidOutputInfo_n`, `asyncInvalidOutputInfo_n` |
| async callback ignoring the failure | `asyncInvalidOutputInfo_n` |
| element error on a refused header | the six header-refusal cases |

The one survivor is dropping the parse-result check. Without it `gst_tensor_meta_info_convert()` fails and leaves the local info uninitialised, so the outcome is undefined behaviour. The valgrind gate would report it as an uninitialised read; an assertion can't pin it.

The async-callback branch of B9 is driven by a mock V1 sub-plugin registered in the test with `nnstreamer_filter_probe()`. From one invoke it dispatches a valid set of three outputs and a set whose middle info is invalid. Exactly one buffer (three memories) must reach the sink. This was added after the first review round.

## Local verification (WSL, Ubuntu)

- Full `meson test`: 19/21. The two failures are this machine's environment and fail identically on `main`: `unittest_filter_python3` (numpy 2 ABI) and `unittest_trainer` (30 s timeout).
- `unittest_filter_custom` 19/19, `unittest_plugins`, `unittest_sink`, `unittest_common` all green.
- SSAT `nnstreamer_filter_custom` 23/23. `nnstreamer_filter_python3` fails identically on `main` (same numpy issue). tflite2 and litert aren't built here, so those legs are CI-only.
- Strict build (`-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`) clean for every commit on its own. Each intermediate commit also passes the existing tests.
- clang-format: no diff on the test file. gst-indent on `tensor_filter.c` shows no new diff (main has the same 13 pre-existing lines). `doxygen-tag.sh` passes, and its negative control (one `@brief` removed) fails.

## Related

- **#4957** (fix for #4928) rewrites the same function, and the two PRs conflict. The agreed order ([coordination note](https://github.com/nnstreamer/nnstreamer/pull/4957#issuecomment-5628597833)) is **#4953 first, then #4957 rebased on top**. After the rebase, #4957 keeps only its type/dimension equality check.
- This PR now posts the `STREAM/WRONG_TYPE` element error that #4957's `flexibleInputInvalidHeader_n` and `flexibleInputSparseHeader_n` expect, so those two stay valid after the rebase.
- Without `invoke-dynamic`, this PR leaves the header's type unused rather than copying it. A same-size type change still passes the size check; refusing it is #4957's part.

### DIFF
--- a/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
+++ b/gst/nnstreamer/include/nnstreamer_plugin_api_filter.h
@@ -117,7 +117,7 @@ typedef struct _GstTensorFilterProperties
   int num_models; /**< number of model files. Some frameworks need multiple model files to initialize the graph (caffe, caffe2) */
 
   int input_configured; /**< TRUE if input tensor is configured. Use int instead of gboolean because this is referred by custom plugins. */
-  GstTensorsInfo input_meta; /**< configured input tensor info */
+  GstTensorsInfo input_meta; /**< configured input tensor info. With invoke_dynamic and a flexible input, tensor-filter updates the type and dimension of each tensor from its meta header before invoke; otherwise a flexible input is checked against it and it is left as configured. */
   tensors_layout input_layout; /**< data layout info provided as a property to tensor_filter for the input, defaults to _NNS_LAYOUT_ANY for all the tensors */
   unsigned int input_ranks[NNS_TENSOR_SIZE_LIMIT];  /**< the rank list of input tensors, it is calculated based on the dimension string. */
 
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -1725,6 +1725,7 @@ gst_tensor_filter_set_caps (GstBaseTransform * trans,
         ("Set-caps failed. Invalid output config (padcaps) for tensor-filter (%s:%s): its format is static, but not equal to the internal configuration: %s\nThis might be an internal error. Please report to https://github.com/nnstreamer/nnstreamer/issues .",
             GST_STR_NULL (prop->fwname), TF_MODELNAME (prop), compare));
     g_free (compare);
+    gst_tensors_config_free (&config);
     return FALSE;
   }
 
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -1010,8 +1010,9 @@ _gst_tensor_filter_transform_check_invoke_result (GstBaseTransform * trans,
 
 /**
  * @brief Internal function to make output buffer.
+ * @return FALSE if the dynamic invoke has left an output tensor info invalid. The output data not appended to the buffer is released.
  */
-static void
+static gboolean
 _gst_tensor_filter_transform_update_outbuf (GstBaseTransform * trans,
     FilterTransformData * in_trans_data, FilterTransformData * out_trans_data,
     GstBuffer * outbuf)
@@ -1089,6 +1090,16 @@ _gst_tensor_filter_transform_update_outbuf (GstBaseTransform * trans,
     if (prop->invoke_dynamic) {
       GstTensorMetaInfo meta;
 
+      if (!gst_tensor_info_validate (_info)) {
+        ml_loge_stacktrace
+            ("gst_tensor_filter_transform: The tensor-filter subplugin (%s : %s) has returned an invalid tensor info for the %u'th output tensor with the dynamic invoke.\n",
+            prop->fwname, TF_MODELNAME (prop), i);
+        for (; i < prop->output_meta.num_tensors; i++)
+          gst_tensor_filter_destroy_notify_util (priv,
+              out_trans_data->tensors[i].data);
+        return FALSE;
+      }
+
       /* Convert to flexible tensors */
       gst_tensor_info_convert_to_meta (_info, &meta);
       meta.media_type = _NNS_TENSOR;
@@ -1115,6 +1126,8 @@ _gst_tensor_filter_transform_update_outbuf (GstBaseTransform * trans,
     /* append the memory block to outbuf */
     gst_tensor_buffer_append_memory (outbuf, out_trans_data->mem[i], _info);
   }
+
+  return TRUE;
 }
 
 /**
@@ -1185,8 +1198,12 @@ gst_tensor_filter_async_output_callback (GstTensorMemory * data,
     out_trans_data->tensors[i].size = data[i].size;
   }
 
-  _gst_tensor_filter_transform_update_outbuf (trans, NULL, out_trans_data,
-      outbuf);
+  if (!_gst_tensor_filter_transform_update_outbuf (trans, NULL, out_trans_data,
+          outbuf)) {
+    g_clear_pointer (&outbuf, gst_buffer_unref);
+    g_clear_pointer (&out_trans_data, g_free);
+    return -1;
+  }
   g_clear_pointer (&out_trans_data, g_free);
 
   if (gst_pad_push (trans->srcpad, outbuf) != GST_FLOW_OK) {
@@ -1280,8 +1297,9 @@ gst_tensor_filter_transform (GstBaseTransform * trans,
     goto done;
   }
 
-  _gst_tensor_filter_transform_update_outbuf (trans, in_trans_data,
-      out_trans_data, outbuf);
+  if (!_gst_tensor_filter_transform_update_outbuf (trans, in_trans_data,
+          out_trans_data, outbuf))
+    retval = GST_FLOW_ERROR;
 
   goto done;
 
--- a/gst/nnstreamer/tensor_filter/tensor_filter.c
+++ b/gst/nnstreamer/tensor_filter/tensor_filter.c
@@ -675,26 +675,47 @@ _gst_tensor_filter_release_mem_until_idx (FilterTransformData * trans_data,
 }
 
 /**
- * @brief Internal function to convert tensor meta and get header size of flexible tensor.
+ * @brief Internal function to parse the meta header of a flexible input tensor and get its header size.
+ * @details The configured input info is left as it is unless the dynamic invoke is enabled, which takes the type and dimension of each incoming tensor.
+ * @return FALSE if the memory does not carry a header describing the data after it.
  */
-static gsize
+static gboolean
 _gst_tensor_filter_convert_meta (FilterTransformData * trans_data,
-    GstTensorsInfo * info, guint idx)
+    GstTensorFilterProperties * prop, guint idx, gsize * hsize)
 {
-  gsize header_size = 0;
   GstTensorMetaInfo *_meta;
-  GstTensorInfo *_info;
+  GstMapInfo *map;
+  GstTensorInfo info, *_info;
 
-  if (trans_data->is_flexible) {
-    _meta = &trans_data->meta[idx];
-    _info = gst_tensors_info_get_nth_info (info, idx);
+  *hsize = 0;
+  if (!trans_data->is_flexible)
+    return TRUE;
 
-    gst_tensor_meta_info_parse_header (_meta, trans_data->info[idx].data);
-    header_size = gst_tensor_meta_info_get_header_size (_meta);
-    gst_tensor_meta_info_convert (_meta, _info);
+  _meta = &trans_data->meta[idx];
+  map = &trans_data->info[idx];
+
+  /* parse_header () reads a whole header of the default version before validating it */
+  gst_tensor_meta_info_init (_meta);
+  if (map->size < gst_tensor_meta_info_get_header_size (_meta) ||
+      !gst_tensor_meta_info_parse_header (_meta, map->data))
+    return FALSE;
+
+  *hsize = gst_tensor_meta_info_get_header_size (_meta);
+  if (*hsize == 0 || *hsize > map->size ||
+      _meta->format == _NNS_TENSOR_FORMAT_SPARSE)
+    return FALSE;
+
+  if (!gst_tensor_meta_info_convert (_meta, &info) ||
+      gst_tensor_info_get_size (&info) != map->size - *hsize)
+    return FALSE;
+
+  if (prop->invoke_dynamic) {
+    _info = gst_tensors_info_get_nth_info (&prop->input_meta, idx);
+    _info->type = info.type;
+    memcpy (_info->dimension, info.dimension, sizeof (info.dimension));
   }
 
-  return header_size;
+  return TRUE;
 }
 
 /**
@@ -735,7 +756,14 @@ _gst_tensor_filter_transform_get_all_input_data (GstBaseTransform * trans,
       return NULL;
     }
 
-    hsize = _gst_tensor_filter_convert_meta (trans_data, &prop->input_meta, i);
+    if (!_gst_tensor_filter_convert_meta (trans_data, prop, i, &hsize)) {
+      GST_ELEMENT_ERROR_BTRACE (self, STREAM, WRONG_TYPE,
+          ("The %u'th memory chunk (%zu bytes) of the flexible input buffer for tensor-filter (%s : %s) does not start with a valid tensor meta header that describes the data after it.",
+              i, trans_data->info[i].size, prop->fwname, TF_MODELNAME (prop)));
+      _gst_tensor_filter_release_mem_until_idx (trans_data, i + 1);
+      g_free (trans_data);
+      return NULL;
+    }
 
     trans_data->tensors[i].data = trans_data->info[i].data + hsize;
     trans_data->tensors[i].size = trans_data->info[i].size - hsize;
@@ -813,7 +841,7 @@ _gst_tensor_filter_transform_get_invoke_tensors (GstBaseTransform * trans,
       expected = gst_tensor_filter_get_tensor_size (self, i, TRUE);
       if (expected != trans_data->tensors[i].size) {
         ml_loge_stacktrace
-            ("gst_tensor_filter_transform: Input buffer size (%u'th memory chunk: %zd) is invalid, which is expected to be %zd, which is the frame size of the corresponding tensor. Maybe, the pad capability is not consistent with the actual input stream; if the size is supposed to change dynamically and the given neural network, framework, and the subpluigins can handle it, please consider using format=flexible.\n",
+            ("gst_tensor_filter_transform: Input buffer size (%u'th memory chunk: %zd) is invalid, which is expected to be %zd, which is the frame size of the corresponding tensor. Maybe, the pad capability is not consistent with the actual input stream; if the size is supposed to change dynamically and the given neural network, framework, and the subpluigins can handle it, please consider using format=flexible with invoke-dynamic=true.\n",
             i, trans_data->tensors[i].size, expected);
         g_free (invoke_tensors);
         return NULL;
--- a/tests/nnstreamer_filter_custom/unittest_filter_custom.cc
+++ b/tests/nnstreamer_filter_custom/unittest_filter_custom.cc
@@ -9,12 +9,15 @@
 
 #include <gtest/gtest.h>
 #include <glib/gstdio.h>
+#include <gst/check/gstharness.h>
 #include <gst/gst.h>
 #include <nnstreamer_plugin_api.h>
 #include <nnstreamer_plugin_api_util.h>
 #include <nnstreamer_util.h>
 #include <stdlib.h>
+#include <sys/mman.h>
 #include <tensor_filter_custom_easy.h>
+#include <unistd.h>
 #include <unittest_util.h>
 
 /** @brief User data for new_data_cb and custom filter */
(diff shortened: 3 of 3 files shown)


myungjoo:
*This review was written by a separate AI agent (Claude) and is being posted on the PR author's behalf.*

## Summary

I reviewed the diff against `origin/main` (3 files: `gst/nnstreamer/tensor_filter/tensor_filter.c`, `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h`, `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`), cross-checked the fixed logic against issue #4920's B6/B7/B9 descriptions and against the consuming sub-plugins (litert, custom-easy), and built/ran the tests locally (Ubuntu/WSL, meson).

**Verification performed:**
- Traced `_gst_tensor_filter_convert_meta()` byte-by-byte against `gst_tensor_meta_info_parse_header()`/`gst_tensor_meta_info_get_header_size()` — the pre-parse `map->size < gst_tensor_meta_info_get_header_size(_meta)` check (on a freshly-`gst_tensor_meta_info_init`-ed meta, i.e. 128 bytes for the current/only version) correctly bounds the unconditional 88-byte read inside `parse_header()`, so the OOB read described in B6 is closed. This mirrors the same init-then-check idiom already used in `gst_tensor_meta_info_parse_memory()` (`nnstreamer_plugin_api_impl.c:1634-1640`).
- Confirmed the `invoke_dynamic` branch only overwrites `type`/`dimension` on the persistent `prop->input_meta` entry (the only two fields `GstTensorInfo` has besides `name`), so the name-loss/leak described in B6 is fixed without breaking litert's `invoke_dynamic()` contract — litert's own comment (`tensor_filter_litert.cc:991-994`) explicitly documents "a flexible sink pad rewrites prop->input_meta from each buffer's own meta header, type included," which matches the new behavior exactly.
- Traced the B9 fix in `_gst_tensor_filter_transform_update_outbuf()`: the `gst_tensor_info_validate(_info)` check runs before any per-tensor `mem[i]` is created, so the cleanup loop (`for (; i < num; i++) gst_tensor_filter_destroy_notify_util(...)`) only ever touches not-yet-wrapped raw tensors; already-appended tensors are owned by `outbuf` and released when the caller drops it on `GST_FLOW_ERROR`. No double-free/leak found in either the sync (`gst_tensor_filter_transform`) or async (`gst_tensor_filter_async_output_callback`) caller.
- Confirmed `_gst_tensor_filter_convert_meta`/`_gst_tensor_filter_transform_update_outbuf` are file-local `static` functions with no other callers, and `input_meta`/`invoke_dynamic` are not documented anywhere under `Documentation/`, so the updated doxygen comment on `input_meta` in `nnstreamer_plugin_api_filter.h` is the right (and sufficient) documentation update — no ARCHITECTURE-level doc is affected.
- **Built and ran the actual bug**: reverted `tensor_filter.c` to `origin/main` in a local build tree and ran the new `tensorFilterFlexInput.shortMemory_n` test — it reproducibly **SIGSEGVs** (`killed by signal 11 SIGSEGV`), exactly as the PR description claims. Restored the PR's `tensor_filter.c` and reran the full `unittest_filter_custom` suite: **18/18 pass** (10 new + 8 existing).
- Verified `clang-format -style=file` produces no diff on the test file, and `gst-indent` on `tensor_filter.c` only flags 4 pre-existing hunks that are entirely outside this PR's changed lines (unrelated code at the old `hsize = gst_tensor_meta_info_get_header_size(&in_trans_data->meta[i]);` combi block and two `watchdog_trigger` call sites) — no new style violations from this PR's own diff.
- Rebuilt both `tensor_filter.c` and the test file under the strict profile (`-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`) — clean, no warnings.

## Findings

No Blocking or Major issues found.

**1. [Minor] `tensor_filter.c` async-output-callback failure path is untested** — `gst_tensor_filter_async_output_callback()` around `tensor_filter.c:1200-1205` (the new `if (!_gst_tensor_filter_transform_update_outbuf(...)) { g_clear_pointer(&outbuf, ...); ...; return -1; }` branch) has no direct test, which the PR body discloses honestly ("Only llamacpp implements async invoke, and custom-easy can't drive it"). I traced the logic by hand and it is correct and symmetric with the already-tested sync path (no double free — `_gst_tensor_filter_transform_update_outbuf` has already released or transferred ownership of every tensor before returning FALSE). Low risk given the shared core function is well covered, but it means a future regression specifically in this wrapper (e.g. someone "fixing" it to also loop-free `data[i].data` again) would not be caught by CI. Suggested follow-up (not blocking this PR): if/when an async-capable sub-plugin becomes testable in-tree, add a case exercising this path.

**2. [Minor] B7 (`gst_tensor_filter_set_caps` leak fix) ships without a regression test** — `tensor_filter.c:1727` (`gst_tensors_config_free (&config);` added before the `return FALSE;` on the output-mismatch branch). The PR body explains why: the branch is a "should never happen" consistency check reachable only when negotiated `outcaps` disagree with the config derived from the very same `transform_caps()`, and it's a leak, which the repo's valgrind CI gate does not count as an error (per PR text and matching the project's own valgrind-gate design). This satisfies CLAUDE.md's "state the reason if genuinely untestable" exception, and I could not find a way to trigger the branch through normal caps negotiation either. Flagging only so a maintainer can confirm they agree the branch is unreachable via public API/pipeline construction (vs. only reachable by calling the static `gst_tensor_filter_set_caps` directly, which isn't exposed for unit testing).

**3. [Nit] Bounds-check idiom could use a short comment** — `tensor_filter.c:697-700`: `gst_tensor_meta_info_init(_meta); if (map->size < gst_tensor_meta_info_get_header_size(_meta) || ...)`. This reuses the current/default version's header size (128) as the safety bound for the fixed-length read inside `gst_tensor_meta_info_parse_header()`, which works today (only one version exists, 128 ≥ the 88 bytes actually read) and mirrors the identical pattern in `gst_tensor_meta_info_parse_memory()`. A one-line comment noting *why* the freshly-initialized meta's header size is the right pre-parse bound (rather than a magic/named constant tied to `parse_header`'s fixed read span) would help the next person who adds a second header version. Not a functional issue.

## Verdict

I did not find any Blocking or Major issues. The fix is well-scoped to the three B6/B7/B9 items in #4920, does not touch any other module, the size of the change (82 lines in `tensor_filter.c`, 1 doc line, 446 lines of new tests) is proportionate to three real memory-safety/UB bugs, and the accompanying tests are unusually rigorous (I independently reproduced the pre-fix SIGSEGV and confirmed the fix resolves it, plus a full green run of the existing suite and a clean strict/format build). In my assessment this PR is **mergeable** as-is; items 1–3 above are optional follow-ups, not merge blockers.

myungjoo:
Response to the relayed agent review above (head is now `c5ced7f1`, same 4-commit structure, force-pushed):

1. **Minor, async failure branch untested:** fixed. `tensorFilterFlexInput.asyncInvalidOutputInfo_n` registers a mock V1 sub-plugin with `nnstreamer_filter_probe()` and runs it with `invoke-dynamic=true invoke-async=true`. From one invoke it dispatches three outputs twice through `nnstreamer_filter_dispatch_output_async()`, the second time with the middle output info left invalid. Exactly one buffer (three memories) must reach the sink.
   - Fails on `main`: the broken buffer is pushed as well.
   - Fails if the async callback ignores `_gst_tensor_filter_transform_update_outbuf()`'s result (mutation check).
2. **Minor, B7 without a test:** unchanged, for the reason in the PR body (the branch sits behind negotiation, and a leak can't fail CI). Leaving the final call to a maintainer as suggested.
3. **Nit, pre-parse size check:** added a one-line comment saying `parse_header()` reads a whole header of the default version before validating it, which is what the check against the initialised meta guards.

myungjoo:
*This review was written by a separate AI agent (Claude) and is being posted on the PR author's behalf. Round 2 — the branch was force-pushed since the previous relayed review (previously reviewed head `065b6cc7`, now `c5ced7f1`, same 4-commit structure).*

## What changed since round 1

Diffed `065b6cc7` → `c5ced7f1`. Two changes, both direct responses to the round-1 findings:

1. `tensor_filter.c:697` — added the one-line comment on the pre-parse size-guard (addresses round-1 Nit #3).
2. `tests/nnstreamer_filter_custom/unittest_filter_custom.cc` — new `tensorFilterFlexInput.asyncInvalidOutputInfo_n`, with a mock V1 sub-plugin (`flex_in_async`) registered via `nnstreamer_filter_probe()`, driving `invoke-dynamic=true invoke-async=true` (addresses round-1 Minor #1: the async failure branch of the B9 fix was previously untested).

Round-1 Minor #2 (B7's leak fix ships without a regression test) is left as-is; the author's response repeats the PR body's reasoning (the branch sits behind negotiation and a plain leak can't fail the valgrind CI gate) and explicitly defers the call to a maintainer. I did not find a way to reach that branch through normal caps negotiation either, so I have nothing to add against that reasoning.

## Verification performed this round

- Re-read the full current diff (`gst/nnstreamer/tensor_filter/tensor_filter.c`, `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h`, `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`) against issue #4920 items B6/B7/B9, independent of the round-1 review's own trace.
- Confirmed the WSL build tree I tested against is byte-identical to this PR head's `tensor_filter.c` and `unittest_filter_custom.cc` (diffed directly, no difference).
- Rebuilt and ran `meson test -C build --no-rebuild -v unittest_filter_custom`: **19/19 pass**, including the new `asyncInvalidOutputInfo_n`.
- Rebuilt the strict profile (`-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`) from clean (`ninja -C build-strict`): no warnings, and `meson test -C build-strict --no-rebuild -v unittest_filter_custom` also 19/19.
- Traced ownership through `_gst_tensor_filter_transform_update_outbuf()`'s new failure path for the async case specifically (the thing round 1 could not exercise): on an invalid output info at index `i`, tensors `[i, num_tensors)` are released via `gst_tensor_filter_destroy_notify_util()` before returning `FALSE`; tensors `[0, i)` are already owned by `outbuf` (appended via `gst_tensor_buffer_append_memory`). `gst_tensor_filter_async_output_callback()` then does `g_clear_pointer(&outbuf, gst_buffer_unref)`, which releases the already-appended ones through the buffer. No double-free / no leak by inspection, and it matches the new test's own accounting (`flex_in_async_dispatched == 2`, `gst_harness_buffers_received == 1`, one 3-memory buffer).
- Cross-checked the B6 fix's contract against the actual consumers: `tensor_filter_litert.cc:990-994` (`invoke_dynamic`) documents reading type+dimension from `prop->input_meta` after a flexible sink pad rewrite, matching this PR's "type/dimension only, name kept" change exactly. `tensor_filter_custom_easy.c:274-275` passes `&prop->input_meta` straight into `func_dynamic()`, again matching.
- Confirmed no `Documentation/` file references `input_meta` or `invoke_dynamic`, so the updated doxygen on the `input_meta` field in `nnstreamer_plugin_api_filter.h` is the correct and sufficient documentation surface — no ARCHITECTURE-level doc is stale.
- I attempted to run the new async test under valgrind on the WSL box to independently confirm memcheck-cleanliness of the exact code path B9 touches, but the box's valgrind (`~/nns-b6`'s custom-built 3.18.1 + debuginfo root) fails at startup with a `strlen`/`ld-linux-x86-64.so.2` redirection error unrelated to this PR — I could not get it running even after pointing `--extra-debuginfo-path` at the local debug-symbol tree, so I'm not able to independently confirm a valgrind-clean run here; the ownership trace above is by code inspection only, backed by the test's harness-level exit-code/output-count assertions.

## Findings

No Blocking or Major issues. No new issues found beyond what round 1 already raised and what this round's force-push already resolved.

**1. [Resolved]** Round-1 Minor #1 (async failure branch untested) — closed by the new `asyncInvalidOutputInfo_n` test. Verified it exercises the intended path (mutation-tested per the PR body: removing the B9 check, or having the async callback ignore `_gst_tensor_filter_transform_update_outbuf()`'s return value, both make it fail) and passes on this head.

**2. [Resolved]** Round-1 Nit #3 (bounds-check idiom needed a comment) — closed by the one-line comment at `tensor_filter.c:697`.

**3. [Minor, unchanged, left to maintainer]** B7's leak fix (`tensor_filter.c:1728`, `gst_tensors_config_free (&config);`) still has no regression test. The author's stated reasons (branch behind negotiation invariants, plain leak can't fail the valgrind CI gate) are consistent with what I could verify — I did not find a caps-negotiation path that reaches this branch either. This satisfies CLAUDE.md's "state the reason if genuinely untestable" allowance; flagging only so a maintainer explicitly signs off that the branch is believed unreachable via public API/pipeline construction.

## Verdict

I found no Blocking or Major issues in this round, and both actionable round-1 findings were resolved with a correct comment and a real, passing regression test for the previously-uncovered async path. The remaining open item (#3 above) is the same Minor the author already explained and explicitly deferred to a maintainer decision in the PR body. In my assessment this PR is **mergeable** as-is.

myungjoo:
[Relayed on behalf of the author by a Claude agent] Response to [the `myungjoo-bot` review](https://github.com/nnstreamer/nnstreamer/pull/4953#pullrequestreview-5167041375) on `c5ced7f1`. The head is now `193b7c96`, force-pushed; it keeps the same 4 commits and the same base, since the new `main` commits touch none of these files. Delta: `git diff c5ced7f1 193b7c96`.

**1. [Medium] Overlap with #4957:** the order is settled by the maintainer's note on #4957: this PR first, then #4957 rebased on top. For the error-reporting difference, I took the review's suggestion here: a refused header now posts `GST_ELEMENT_ERROR_BTRACE (self, STREAM, WRONG_TYPE, ...)` before returning `GST_FLOW_ERROR`. That matches the element's caps-time refusals, and #4957's `flexibleInputInvalidHeader_n` / `flexibleInputSparseHeader_n` stay valid after its rebase.
- The tests now attach a bus to the harness. Each of the six header-refusal cases asserts a `STREAM/WRONG_TYPE` error whose source is a `GstTensorFilter`.
- `smallerThanModel_n` and `keepsConfiguredInfo` assert that no such error is posted, so the model-size refusal can't be mistaken for the header one.
- Replacing the element error with a log line fails all six header-refusal cases (mutation check).
- The PR body's stale "No open PR touches this file" and its #4928 paragraph now point at #4957 and the agreed order.

**2. [Low] `gst_tensor_meta_info_convert()` result ignored:** fixed. It is now `if (!gst_tensor_meta_info_convert (_meta, &info) || gst_tensor_info_get_size (&info) != map->size - *hsize) return FALSE;`.

**3. [Low] No dynamic-mode sparse test:** fixed. `sparseHeader_n` now runs the sparse header through a static and a dynamic custom-easy model. Moving the check under `!prop->invoke_dynamic` (#4957's scope) now fails its dynamic half (mutation check).

**4. [Nit] Misleading size-mismatch hint:** fixed. The message now ends "please consider using format=flexible with invoke-dynamic=true."

**Note on executorch-llama:** agreed that this is the sub-plugin's own defect, and I am leaving it out of this PR:
- `invoke_dynamic()` sets `dimension[0] = result.size ()`, and `g_malloc0 (0)` returns NULL for an empty generation.
- On `main` that output already failed `gst_tensor_info_convert_to_meta()` with a GLib critical. The buffer went downstream with no valid memory, and a decoder refuses it since C24. So the stream was already broken there.
- With this PR it fails at the filter with a message naming the sub-plugin.
- executorch is disabled in the Tizen spec (`executorch_support 0`), so this is outside #4920's scope.

**Local verification of the new head (WSL):**
- `unittest_filter_custom` 19/19.
- `unittest_plugins`, `unittest_sink`, `unittest_common` green.
- Strict build (`-Werror -Wextra -Wsign-compare`) clean for each commit on its own.
- clang-format: no diff. gst-indent adds nothing new. doxygen-tag passes, and its negative control fails.
- The new tests against `main`'s `tensor_filter.c`: all 11 cases fail, `shortMemory_n` with SIGSEGV.
- Mutation matrix, 10 mutants: every mutant is caught. The one not in the set is dropping the parse-result check. It was previously a survivor with undefined behaviour; with finding 2's fix it is an equivalent mutant, because `convert()` validates the same meta and refuses it (with a GLib critical).
- In one run of the per-commit check, all three suites failed at the first commit. That didn't reproduce in two further runs, so it looks transient and not caused by the code.

myungjoo:
*This review was written by a separate AI agent (Claude) and is being posted on the PR author's behalf. Round 3 — the branch was force-pushed since `myungjoo-bot`'s review (reviewed head `c5ced7f1`, now `193b7c96`, same 4-commit structure).*

## What changed since the `myungjoo-bot` review

Diffed `c5ced7f1` → `193b7c96`. All four changes are direct responses to that review's findings:

1. `tensor_filter.c:708-710` — `gst_tensor_meta_info_convert()`'s return value is now checked (`if (!gst_tensor_meta_info_convert (_meta, &info) || gst_tensor_info_get_size (&info) != map->size - *hsize) return FALSE;`) before `info` is read. Addresses **Low #2**.
2. `tensor_filter.c:760-762` — the header-refusal path in `_gst_tensor_filter_transform_get_all_input_data()` now posts `GST_ELEMENT_ERROR_BTRACE (self, STREAM, WRONG_TYPE, ...)` instead of only `ml_loge_stacktrace`. Addresses the error-reporting half of **Medium #1** (the behavioural conflict with #4957's `flexibleInputInvalidHeader_n`/`flexibleInputSparseHeader_n`, which expect a `STREAM/WRONG_TYPE` bus message).
3. `tests/nnstreamer_filter_custom/unittest_filter_custom.cc` — new `_flex_in_harness()` attaches a `GstBus` to the harness element, and a new `_flex_in_refused_header()` helper pops and matches that bus message. `sparseHeader_n` is now a loop over `invoke-dynamic=false` and `invoke-dynamic=true` models (`flex_in_sparse`, `flex_in_sparse_dynamic`), and every existing header-refusal case (`shortMemory_n`, `brokenHeaderMagic_n`, `brokenHeaderType_n`, `headerSizeMismatch_n`, `unknownVersion_n`) plus both `sparseHeader_n` variants now assert `_flex_in_refused_header(h)` is true, while `smallerThanModel_n` (a different failure path) and `keepsConfiguredInfo` (the positive path) assert it is false. Addresses **Low #3** (no dynamic-mode sparse coverage) and strengthens the whole suite with a real assertion on the error contract instead of just the flow return code.
4. `tensor_filter.c:844` — the size-mismatch message now says "...please consider using format=flexible with invoke-dynamic=true" instead of just "format=flexible". Addresses **Nit #4**.
5. PR body — the "Related" section now names #4957 explicitly, links the coordination note, and states the new error-contract alignment. Addresses the "stale PR body" sub-finding of **Medium #1**.

The remaining part of **Medium #1** (duplicated checks/tests between #4953 and #4957, and which PR ends up owning what) is not something #4953 can resolve unilaterally — it is now tracked by the maintainer coordination note posted on #4957 ("merge #4953 first, then rebase #4957 on top"), which explicitly requires #4957 to preserve #4953's invariants (bounded pre-parse read, sparse refusal in both invoke modes, name-preserving dynamic update, B9/B7 fixes) and to keep the new `STREAM/WRONG_TYPE` contract this round adds. That is the correct place to settle it, not a further change to this PR.

## Verification performed this round

- Confirmed the local checkout (`.claude/worktrees/issue-4920-task-selection-fb2dc4`, HEAD `193b7c96`) matches the WSL build tree (`~/nns-b6`) byte-for-byte (`md5sum` on all three changed files, CRLF-stripped) — not stale from a previous round.
- Rebuilt both profiles from a forced touch (not cache) and reran the suite:
  - `ninja -C ~/nns-b6/build` (default) then `meson test -C build --no-rebuild -v unittest_filter_custom`: **19/19 pass**, including all 11 `tensorFilterFlexInput.*` cases and the two-variant `sparseHeader_n` loop.
  - `ninja -C ~/nns-b6/build-strict` (`-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`), forced recompile of `tensor_filter.c` and the test file: **zero warnings**, and `meson test` on that profile: **19/19 pass**.
- Ran `tools/development/gst-indent` on `tensor_filter.c`: only 4 hunks reported, all pre-existing and entirely outside this PR's changed lines (the `nnstreamer-message` structure build, the `hsize = gst_tensor_meta_info_get_header_size(...)` wrap in the combine path, and two `watchdog_trigger` call wraps) — no new style violations from this round's diff. Restored the file afterward so the WSL tree still matches the PR head.
- Ran `clang-format -style=file -output-replacements-xml` on the test file: 0 replacements needed.
- Traced `GST_ELEMENT_ERROR_BTRACE (self, STREAM, WRONG_TYPE, ...)` at `tensor_filter.c:760` against the ~15 other call sites of the same macro already in this file (e.g. `:1374`, `:1385`, `:1690`+) — same signature, same "no separate `ml_loge` needed, the macro's own `GST_ELEMENT_ERROR` already logs" pattern, so this is not a new convention, it makes the header-refusal path consistent with every other refusal path in the file.
- Confirmed `self` is in scope at the new call site (it's `GstTensorFilter *self = GST_TENSOR_FILTER_CAST (trans);`, declared at the top of `_gst_tensor_filter_transform_get_all_input_data()`).
- Checked that this new element error can only fire on the pre-existing "reject a malformed/sparse header" branch, which no well-formed in-tree producer can reach (mirrors the bot's own earlier finding that the sparse refusal can't break an existing pipeline); grepped `tests/` for anything that asserts "no bus error" across a legitimate flexible-tensor path and found nothing that this new message could break.
- Checked `_flex_in_register()` does `memset (d, 0, sizeof (*d))` before each registration, so the `sparseHeader_n` loop's two iterations don't leak `flex_in_data` state (e.g. `invoked`, `in_name`) between the static and dynamic variants.
- Re-read the full current diff against `main` and against the B6/B7/B9 text in #4920 once more, independent of the earlier rounds' own traces; found nothing beyond what is summarized above.
- `gh pr checks 4953`: CI on `193b7c96` is green on everything that has completed so far (Valgrind/arm, MacOS, ExecuTorch+SSAT, CPP-Linter, DCO, Spell Check); the rest (Tizen GBS, Android NDK, RISC-V, Yocto, Ubuntu+Clang, Static checks) were still pending/queued at review time, none failing.

## Findings

No Blocking or Major issues. No new issues introduced by this round's diff itself.

**1. [Resolved]** Bot review Low #2 (`gst_tensor_meta_info_convert()` return value ignored) — closed by the combined check at `tensor_filter.c:708-710`.

**2. [Resolved]** Bot review Low #3 (no dynamic-mode sparse test) — closed by the `sparseHeader_n` loop covering both `invoke-dynamic=false` and `=true`.

**3. [Resolved]** Bot review Nit #4 (size-mismatch message pointed the user at `format=flexible`, which they were already using) — closed by the updated message text.

**4. [Resolved as far as this PR can]** Bot review Medium #1's error-contract half (this PR's header refusal didn't post a bus error, but #4957's tests expect one) — closed by the new `GST_ELEMENT_ERROR_BTRACE (STREAM, WRONG_TYPE, ...)` call and the new `_flex_in_refused_header()` test assertions. The remaining overlap/duplication with #4957 is a cross-PR concern already captured by the maintainer coordination note on #4957 and does not require a further change here.

**5. [Nit, informational only]** Several CI jobs on `193b7c96` (Tizen GBS x86_64/aarch64/armv7l, Android NDK, RISC-V, Yocto, Ubuntu+Clang, Static checks) were still pending at review time. Nothing observed suggests they will fail (local strict build is warning-clean and `gst-indent`/`clang-format` show no new violations), but worth confirming all are green before merge, per the project's own "run the relevant test suites locally (or confirm CI passes) before marking a PR ready for review" rule.

## Verdict

I found no Blocking or Major issues in this round. All three concrete Low/Nit findings from the `myungjoo-bot` review are resolved with real fixes (not just suppressions), and the behavioural-conflict half of the Medium finding is resolved in the way the finding itself suggested (emit the element error from the early refusal path) with a genuine regression test for it, backed by a fresh, non-cached rebuild and full test run on both the default and strict-warnings profiles. The remaining coordination with #4957 is out of this PR's hands and is already tracked on the sibling PR. In my assessment this PR is **mergeable** as-is, pending the still-pending CI jobs turning green.

PR #4954, opened 2026-09-10
[Element/Split] Describe a picked pad by the segment it carries

### LINKED ISSUE #4927: [tensor_split] Source pad caps are taken from the wrong tensorseg entry when tensorpick is not contiguous from 0
## Summary

`tensor_split` derives a new source pad's caps from `tensorseg[num_srcpads]` but fills that pad's buffers from `tensorseg[nth]`, where `nth` is the tensor index. The two agree only when `tensorpick` is unset, or is contiguous and starts at 0. Otherwise the pad advertises one segment's dimensions while carrying another segment's bytes.

- [`gsttensor_split.c` — `gst_tensor_split_get_tensor_pad()`](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/elements/gsttensor_split.c#L307): `dim = g_array_index (tensorseg, tensor_dim *, split->num_srcpads);` — this `dim` becomes the pad's caps.
- [`gsttensor_split.c` — `gst_tensor_split_get_splitted()`](https://github.com/nnstreamer/nnstreamer/blob/main/gst/nnstreamer/elements/gsttensor_split.c#L409): `dim = g_array_index (tensorseg, tensor_dim *, nth);` — this `dim` decides the offset and the size actually copied.

`num_srcpads` counts pads created so far; `nth` is the index of the tensor being emitted. With `tensorpick` set, the picked tensors are renumbered onto `src_0, src_1, ...`, so the two indices diverge.

## Reproduction

```bash
gst-launch-1.0 -v -e videotestsrc num-buffers=1 ! video/x-raw,format=RGB,width=4,height=4,framerate=30/1 ! tensor_converter ! tensor_split name=s tensorseg=1:4:4,2:4:4 tensorpick=1 ! fakesink silent=false
```

The input is one `3:4:4` uint8 tensor (48 bytes), split into `1:4:4` (16 B) and `2:4:4` (32 B), picking only the second. Observed downstream:

```
fakesink0.GstPad:sink: caps = other/tensors, ... dimensions=(string)1:4:4, types=(string)uint8
fakesink0: last-message = chain ******* (fakesink0:sink) (32 bytes, ...)
```

Caps claim 16 bytes; the buffer is 32 bytes. `src_0` was given `tensorseg[0]`'s dimensions because it is the first pad created, but it carries `tensorseg[1]`'s data because that is the tensor `tensorpick` selected.

## Impact

Any downstream element that trusts the caps — `tensor_filter`, `tensor_transform`, `tensor_decoder`, or an application reading through the ML API — reads a wrongly-shaped tensor, or a wrong number of bytes. There is no error or warning; the pipeline runs to completion.

## Why it is not caught today

Every existing test uses either no `tensorpick`, or a `tensorpick` that starts at 0 over equal-sized segments, so the two indices happen to coincide:

- `tests/nnstreamer_split/runTest.sh` case 7: `tensorpick=0,1 tensorseg=1:16:16,1:16:16,1:16:16`
- `tests/nnstreamer_sink/unittest_sink.cc:1043`: `tensorseg=1:160:120:1,1:160:120:1,1:160:120:1 tensorpick=0,1,2`

## Fixing it needs a decision first

The fix depends on what `tensorpick` is meant to do to the pad numbering, and either choice changes the caps that existing pipelines see:

1. **Index the segment by the tensor, not by the pad** — `gst_tensor_split_get_tensor_pad()` uses `tensorseg[nth]`. The pad keeps its sequential `src_%u` name but describes the tensor it actually carries. This looks like the intended behaviour and matches `gst_tensor_split_get_splitted()`.
2. **Index the pad by the tensor as well** — name the pad `src_<nth>`, so `tensorpick=1` produces `src_1` and there is no renumbering at all. Closer to how `tensor_demux` reads, but it changes pad names for existing `tensorpick` pipelines.

Option 1 is the smaller change and only corrects pipelines that are already receiving wrong caps. Whichever is chosen, `tests/nnstreamer_split/runTest.sh` should gain a case with differently-sized segments and a `tensorpick` that does not start at 0, since that combination is what is unguarded today.

## Notes

Found while working on #4769 (nnstreamer/nnstreamer#4925, property/streaming-thread races in `tensor_demux` and `tensor_split`). This defect is independent of that race, predates it, and was deliberately left out of that PR because it needs the semantics decision above.

myungjoo:
Fixes #4927.

## Problem

`tensor_split` gave a new source pad the caps of `tensorseg[num_srcpads]` (how many pads exist so far) but filled it from `tensorseg[nth]` (the segment it sends). With a `tensorpick` that does not start at 0, the two differ, and the pad announces one segment's shape while carrying another's bytes:

```
tensor_split tensorseg=1:4:4,2:4:4 tensorpick=1
src_0 caps: dimensions=1:4:4 (16 bytes)   buffers: 32 bytes
```

No error or warning is raised, so anything downstream that trusts the caps (`tensor_filter`, `tensor_transform`, `tensor_decoder`, ML API apps) reads a wrongly shaped tensor.

## Fix (option 1 of the issue)

- The pad takes its caps from `tensorseg[nth]`. Pad names stay sequential (`src_0`, `src_1`, ... in ascending segment index), so pipelines that link by pad name are unaffected. Only pipelines that were already receiving wrong caps see different ones. This matches how `tensor_demux` names and describes picked pads.
- The refusal #4925 added for `num_srcpads >= tensorseg->len` is removed, because the rule array is no longer indexed by the pad count. `nth < num_tensors <= tensorseg->len` holds: both values are read under one lock in `chain()`, and `gst_tensor_split_get_splitted()` already relies on the same bound.

## Documentation

- **gst-inspect:** the `tensorpick` description now reads *"Indices of the tensorseg segments to output (e.g., 1,2), which go out on src_0, src_1, ... in ascending index order, each pad carrying the dimensions of its own segment"*. It used to be "Choose nth tensor among tensors ?".
- **New element doc** `gst/nnstreamer/elements/gsttensor_split.md` covers the pads, the properties, how segments are cut (consecutive byte ranges), and a table showing which segment each `src_N` carries. It is linked from the element list, `component-description.md` and the hotdoc sitemap.

## Tests

The new tests fail without the element change and pass with it; this was checked locally by swapping the old `gsttensor_split.c` back in.

- `unittest_plugins` (`testTensorSplit.*`), checking caps dimension, caps size, buffer size and first byte on every pad:
  - `segmentsWithoutTensorpick`: the unchanged default path.
  - `tensorpickSkippingFirstSegment`: the case from the issue.
  - `tensorpickWithGap`: `0,2` and `2,0` give the same pads in segment order.
  - `addPadAfterTensorsegShrinks`: replaces `addPadBeyondTensorseg_n`, whose refusal no longer exists.
  - `tensorpickPastBuffer_n`: a picked segment that reaches past the buffer is refused.
  - `tensorpickNegotiatesPickedSegment` and `tensorpickRefusesFirstSegment_n`: a real pipeline with a capsfilter.
- SSAT `nnstreamer_split`:
  - case 8: golden comparison, through a capsfilter requiring `2:16:16`.
  - case 8_n: a capsfilter requiring segment 0's `1:16:16` must fail.

## Local verification (WSL, Ubuntu 22.04)

- `meson test`: 22/23 passed. The one failure, `unittest_filter_python3`, is a local NumPy 2.x environment issue: `_ARRAY_API not found` on import. It is unrelated to this change.
- SSAT `nnstreamer_split`, `nnstreamer_decoder_pose` and `nnstreamer_demux` all pass.
- gst-indent and clang-format 16 report no changes.
- `doxygen-tag.sh` in advanced mode passes. It was confirmed live with a negative control.
- Both changed objects compile with `-Werror=sign-compare -Werror`.


myungjoo:
**Automated review disclosure:** This comment was produced by a separate Claude agent (not the PR author acting directly) and is being relayed here via the author's (`myungjoo`) GitHub account, since that is the only account available to post it. Treat it as an independent automated review, not a statement from the human author.

## Scope

Reviewed the diff of PR #4954 (`gst/nnstreamer/elements/gsttensor_split.c`, the new `gsttensor_split.md`, `Documentation/component-description.md`, `Documentation/hotdoc/sitemap.txt`, `gst/nnstreamer/elements/README.md`, `tests/nnstreamer_plugins/unittest_plugins.cc`, `tests/nnstreamer_split/runTest.sh`) against issue #4927, the related #4925 background, and the rest of the repository (searched all `tensorpick` usages across `.c/.cc/.py/.sh/.md`) for regressions.

## Correctness of the fix

Confirmed the core defect and the fix by reading `gst_tensor_split_get_tensor_pad()` and `gst_tensor_split_chain()`:
- Before: the pad's caps were built from `tensorseg[split->num_srcpads]` (pad-creation ordinal) while the buffer content came from `tensorseg[nth]` (segment index). With a non-contiguous `tensorpick` these diverge, exactly as reported in #4927/#4925.
- After: `dim = g_array_index (tensorseg, tensor_dim *, nth)` — pad caps now come from the same `nth` used to fill the buffer (`gst_tensor_split_get_splitted(..., i)`), so caps and payload always agree.
- The removed bounds check (`split->num_srcpads >= tensorseg->len`) is safe to drop: `nth` is always `i` from the `for (i = 0; i < num_tensors; i++)` loop in `chain()`, and `num_tensors`/`tensorseg` are read together under one `GST_OBJECT_LOCK`/`UNLOCK` pair. Checked every writer of `split->num_tensors` (`gst_tensor_split_init`, `gst_tensor_split_remove_src_pads`, and `PROP_TENSORSEG` in `gst_tensor_split_set_property`): `num_tensors` is only ever set to `0` or to `tensorseg->len` at the same time `tensorseg` itself is replaced, under the same lock. So `nth < num_tensors <= tensorseg->len` holds whenever the loop runs, and `g_array_index` cannot go out of bounds. The now-dead `if (srcpad == NULL)` check in `chain()` was correctly removed too, since `gst_tensor_split_get_tensor_pad()` can no longer return `NULL`.

This matches the PR description's claim and I could not find a path that violates the invariant.

Grepped every `tensorpick` occurrence in the repo (`gsttensor_demux.c`, `gsttensor_if.c`, `tests/nnstreamer_demux/*`, `tests/nnstreamer_sink/unittest_sink.cc`, docs). `tensor_demux` is untouched by this PR and uses its own, independent pad-creation code path, so it's unaffected. The only other `tensor_split` + `tensorpick` usage in the repo is `unittest_sink.cc:1043` (`tensorseg=1:160:120:1,1:160:120:1,1:160:120:1 tensorpick=0,1,2`) — a full, sequential, identically-sized pick, which behaves identically before and after this change (nth == pad ordinal in that case). No other in-repo pipeline/test relies on the old (buggy) numbering. The change is isolated to `gsttensor_split.c`; no other element's code was touched.

## Test coverage

The new GTest cases in `unittest_plugins.cc` (`segmentsWithoutTensorpick`, `tensorpickSkippingFirstSegment`, `tensorpickWithGap`, `addPadAfterTensorsegShrinks`, `tensorpickPastBuffer_n`, `tensorpickNegotiatesPickedSegment`, `tensorpickRefusesFirstSegment_n`) directly assert the caps dimension, caps-declared size, actual buffer size, and first-byte content per pad — this is a solid, byte-level regression net for the exact defect in #4927, plus the SSAT `nnstreamer_split` cases 8/8_n add an end-to-end golden-file and negative-caps-negotiation check. I traced the byte-offset math in `tensorpickSkippingFirstSegment` and `addPadAfterTensorsegShrinks` by hand against `gst_tensor_split_get_splitted()`'s offset accumulation and the numbers check out.

CI wiring: `meson test -C build/ -v` runs on the Ubuntu workflow (`.github/workflows/ubuntu_clean_meson_build.yml`), which will run the new GTest cases; `.github/workflows/gbs_build.yml` runs on every PR (`on: pull_request`) with `unit_test 1` for x86_64, and the packaging spec's `%check` invokes `ssat -n -p=1 --summary summary.txt -cn _n`, which covers the new SSAT `nnstreamer_split` cases 8/8_n including the `_n` negative-id convention. So a future regression in this behavior should fail CI on any PR, not just on manual runs.

## Documentation

The property description, `component-description.md`, and the new `gsttensor_split.md` (linked from `elements/README.md` and `hotdoc/sitemap.txt`) all correctly describe the new "pad numbered by creation order, described by the segment it carries" semantics, and the table/examples in the new doc match what the tests actually verify. No architecture/API surface changed (no new/removed properties, signals, or pad templates), so no broader doc is missing.

## Findings

- **Minor** — `tests/nnstreamer_plugins/unittest_plugins.cc`, `_split_run_pipeline()` (new helper backing `tensorpickNegotiatesPickedSegment` / `tensorpickRefusesFirstSegment_n`): it drives the pipeline with raw `gst_element_set_state(pipeline, GST_STATE_PLAYING/NULL)` instead of the `setPipelineStateSync()` helper that the file's other pipeline-driving helper (`_run_merge_pipeline`, same file, ~line 12320) uses for both transitions. Since `gst_element_set_state` is asynchronous, this is a minor inconsistency with the established convention in the same file and a plausible (unconfirmed) source of test flakiness under load, particularly on the `GST_STATE_NULL` teardown right before `gst_object_unref(pipeline)`. Suggest reusing `setPipelineStateSync()` for consistency and to reduce flake risk, matching the pattern already used for the analogous merge-pipeline helper.
- **Nit** — `Documentation/hotdoc/sitemap.txt`: `gsttensor_split.md` was inserted after `gsttensor_src.md`, but "split" < "src" lexicographically, so strict alphabetical order would put it before. The surrounding list isn't strictly alphabetized already (e.g. `gsttensor_sink.md` precedes `gsttensor_aggregator.md`), so this is purely cosmetic and not worth blocking on.

No correctness, memory-safety, thread-safety, or scope-creep issues were found. The change is tightly scoped to `tensor_split` and its own docs/tests, proportionate to the bug being fixed, and the removed bounds check is provably safe given the locking discipline already in place.

## Verdict

From a code-correctness, regression-safety, and test-coverage standpoint, this PR is technically sound and I would consider it mergeable as-is once the two minor/nit items above are optionally addressed (neither is blocking). The PR is currently marked **Draft** with a **DO NOT MERGE** label by the author, which appears to be a deliberate process gate (per the repo's PR workflow) rather than a reflection of any code defect — that label should of course be resolved through the normal process before merge regardless of this review's findings.

myungjoo:

Thanks for the review. Both findings are addressed in 45597e79 (force-pushed; each fix is folded into the commit it belongs to):

- **Minor, `_split_run_pipeline()` state changes:** it now uses `setPipelineStateSync()` for both PLAYING and NULL, like `_run_merge_pipeline()`. I checked that this does not stall the refusal path. Locally, `tensorpickRefusesFirstSegment_n` still finishes in about 0 ms. Both pipeline tests also still fail against the old `gsttensor_split.c`, so the change did not make them vacuous.
- **Nit, sitemap order:** `gsttensor_split.md` now comes before `gsttensor_src.md`.

The tree is otherwise identical to the reviewed head. clang-format 16 reports no changes, and the strict `-Werror=sign-compare` compile passes.

myungjoo:
**Automated review disclosure:** This comment was produced by a separate Claude agent (not the PR author acting directly) and is being relayed here via the author's (`myungjoo`) GitHub account, since that is the only account available to post it. Treat it as an independent automated review, not a statement from the human author.

## Scope

Fresh, independent review of PR #4954 at head `45597e79` — i.e. re-reviewing the whole change (not just the delta since the earlier automated review), and separately checking whether that earlier review's two findings were actually resolved. Reviewed `gst/nnstreamer/elements/gsttensor_split.c` in full (not only the diff hunks), the new `gsttensor_split.md`, `Documentation/component-description.md`, `Documentation/hotdoc/sitemap.txt`, `gst/nnstreamer/elements/README.md`, `tests/nnstreamer_plugins/unittest_plugins.cc`, `tests/nnstreamer_split/runTest.sh`, issue #4927 (and its #4925 background comment), and re-grepped the whole repository for `tensorpick`/`tensor_split` usage to check for regressions outside this element.

## Correctness of the fix

Read `gst_tensor_split_get_tensor_pad()`, `gst_tensor_split_get_splitted()` and `gst_tensor_split_chain()` end to end. Confirmed the defect and the fix independently:

- Before: a new pad's caps came from `tensorseg[split->num_srcpads]` (pad-creation ordinal) while `gst_tensor_split_get_splitted()` always cut the buffer from `tensorseg[nth]` (segment index, unchanged by this PR). These only agreed when `tensorpick` was unset or a contiguous run from 0.
- After: `dim = g_array_index (tensorseg, tensor_dim *, nth)` — caps and payload now use the same index.
- Traced the invariant the removed bounds check relied on: in `gst_tensor_split_set_property()`, `PROP_TENSORSEG` sets `split->tensorseg` (an array of exactly `num` entries, one `g_array_append_val` per loop iteration) and `split->num_tensors = num` together, under one `GST_OBJECT_LOCK`/`UNLOCK`. So `tensorseg->len == num_tensors` holds at every observable point. `gst_tensor_split_chain()` snapshots both under one lock (`num_tensors = split->num_tensors; tensorseg = g_array_ref (split->tensorseg);`) before the loop `for (i = 0; i < num_tensors; i++)`, and `nth` passed into `gst_tensor_split_get_tensor_pad()` is always this same `i`. So `nth < num_tensors == tensorseg->len` provably holds whenever `g_array_index (tensorseg, tensor_dim *, nth)` runs — the dropped bounds check and the dropped `if (srcpad == NULL)` in `chain()` are both safe to remove, since the function can no longer return `NULL`.

This matches the reported bug in #4927/#4925 and I could not find a path that violates the invariant.

Re-grepped `tensorpick` and `tensor_split` across the whole tree. Findings:

- `gsttensor_if.c`'s `tensorpick` hit is an unrelated enum value (`TIFB_TENSORPICK`) on a different element/property — no connection to this change.
- The only other live `tensor_split` + `tensorpick` combination in the repo is `tests/nnstreamer_sink/unittest_sink.cc:1043` (`tensorseg=1:160:120:1,1:160:120:1,1:160:120:1 tensorpick=0,1,2`, and the property-string check at ~5654-5673) — a full, sequential, identically-sized pick, so `nth == pad ordinal` and behavior is identical before/after.
- `tests/nnstreamer_decoder_pose/runTest.sh` uses `tensor_split` without `tensorpick` at all (segment 0, `a.src_0`), unaffected.
- `tensor_demux` (`gsttensor_demux.c`) has its own independent pad-creation path and is untouched by this PR.

No other pipeline, test, or doc in the repo depends on the old (buggy) ordinal-based numbering. The change is isolated to `gsttensor_split.c` plus its own docs/tests — nothing else was touched.

## Test coverage

The GTest cases in `unittest_plugins.cc` (`segmentsWithoutTensorpick`, `tensorpickSkippingFirstSegment`, `tensorpickWithGap`, `addPadAfterTensorsegShrinks`, `tensorpickPastBuffer_n`, `tensorpickNegotiatesPickedSegment`, `tensorpickRefusesFirstSegment_n`) check caps dimension, caps-declared size, actual buffer size and first byte per pad — a solid byte-level regression net for the exact defect. Hand-traced the offset/size math in `gst_tensor_split_get_splitted()` against `tensorpickSkippingFirstSegment` and `addPadAfterTensorsegShrinks`'s expected `first_byte`/`caps_size` values; they check out.

SSAT `nnstreamer_split` case 8/8_n is a good end-to-end complement: case 8 reuses the existing `testcase_stream_1_1.golden` (the correct segment-1 bytes, already used by case 5_1) through a capsfilter requiring `2:16:16`, which only negotiates correctly after this fix; case 8_n asserts that a capsfilter requiring segment 0's `1:16:16` now correctly fails to link — before the fix this bad link would have silently succeeded (the exact defect from #4927).

CI wiring: `meson test -C build/ -v` runs in `.github/workflows/ubuntu_clean_meson_build.yml` (line 181) and will pick up the new GTest cases automatically; `.github/workflows/gbs_build.yml` has an `x86_64` matrix leg with `unit_test 1` that runs `on: pull_request`, whose packaging `%check` step runs SSAT with the `_n` negative-id convention. A future regression here should fail CI on a normal PR.

## Documentation

`component-description.md`'s `tensor_split` link now points at the new `.md` instead of the `.c` source — checked this against the file's own convention: elements with a dedicated doc (`tensor_transform`, `tensor_if`, `tensor_src_iio`) already link to `.md`, elements without one (`tensor_merge`, `tensor_demux`) still link to `.c`, so this is consistent. `gsttensor_split.md`'s front matter and structure match the existing `gsttensor_if.md` pattern. No API/property/signal surface changed, so no other doc gap.

## Verification of the earlier automated review's findings

The earlier review (first comment on this PR) flagged two items and the author's reply claims both are folded into `45597e79`. Checked the current head directly rather than trusting the claim:

1. **Minor — `_split_run_pipeline()` should use `setPipelineStateSync()`, not raw `gst_element_set_state()`.** Confirmed resolved: `_split_run_pipeline()` (`tests/nnstreamer_plugins/unittest_plugins.cc:11675` and `:11685`) now calls `setPipelineStateSync (pipeline, GST_STATE_PLAYING, UNITTEST_STATECHANGE_TIMEOUT)` and `setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT)`, matching `_run_merge_pipeline()`'s established pattern (`:12341`, `:12352`) byte-for-byte in structure.
2. **Nit — sitemap ordering.** Confirmed resolved: `Documentation/hotdoc/sitemap.txt` now lists `gsttensor_split.md` (line 49) before `gsttensor_src.md` (line 50).

## Findings

None. No correctness, memory-safety, thread-safety, scope-creep, or documentation issues found in this independent pass, and the two items from the earlier review are verifiably fixed in the current head.

## Verdict

**OK to merge as is**, from a code-correctness, regression-safety, test-coverage and documentation standpoint. The PR is currently marked **Draft** with a **DO NOT MERGE** label, which reads as a deliberate process gate rather than a code defect — that should be lifted through the normal process when the author is ready, independent of this review.

myungjoo:

Thanks for the review. Both nits are addressed in 5b03812e (force-pushed, each change folded into the commit it belongs to). The change is documentation only; the code is unchanged apart from the two strings below.

I reproduced both findings first on current `main` + this PR + #4962, which also turned up a third way to reach the same state:

| Change while playing | What the pads do afterwards |
| --- | --- |
| `tensorpick=1` → `0,1` | segment 0 goes out on a new `src_1`, segment 1 stays on `src_0` (each pad's caps are right for its own segment) |
| `tensorseg=1:4:4,2:4:4` → `2:4:4,1:4:4` | `src_0` still says `1:4:4` (16 bytes) and pushes 32 bytes; `src_1` the reverse |
| upstream renegotiates `uint8` → `float32` | both pads keep `types=uint8` while pushing four times as many bytes |
| same `tensorseg` change, made in READY and restarted | pads are recreated with the new caps and numbering |

**Nit 1, pad numbering:**
- The gst-inspect blurb now reads "set before the stream starts, they go out on src_0, src_1, ... in ascending index order". It is still one sentence.
- The element's SECTION comment, `gsttensor_split.md` and `component-description.md` now say that pads are numbered in creation order, which is ascending segment order when the properties are set before the stream starts.

**Nit 2, caps are not updated:** stated as a limitation rather than a guarantee.
- The *Source Pads* section of `gsttensor_split.md` now says the caps are set when the pad is created.
- A new *Limitations* section lists the three cases above with their byte counts, notes that none of them raises an error, and gives the way to apply such a change: stop to READY or NULL, set the properties, play again, and relink the new pads.
- `component-description.md` and the SECTION comment carry a one-line version.

I have not filed a follow-up issue for renegotiating caps (or refusing the change while pads exist). I've left that decision to the maintainers.

Verified locally:
- gst-inspect shows the new blurb.
- gst-indent and doxygen-tag are clean.
- All 15 `testTensorSplit` cases and the SSAT `nnstreamer_split` group pass.
- `git merge-tree` stays clean against both current `main` and #4962.

myungjoo:
**Automated review disclosure:** This comment was produced by a separate Claude agent (not the PR author acting directly) and is being relayed here via the author's (`myungjoo`) GitHub account, since that is the only account available to post it. Treat it as an independent automated review, not a statement from the human author.

## Scope

Fresh, independent review of PR #4954 at head `5b03812e` (the current head, after the round that answered `myungjoo-bot`'s two documentation nits). I re-read the whole of `gst/nnstreamer/elements/gsttensor_split.c` (not just the diff hunks), the full `gsttensor_split.md`, `Documentation/component-description.md`, `Documentation/hotdoc/sitemap.txt`, `gst/nnstreamer/elements/README.md`, the new tests in `tests/nnstreamer_plugins/unittest_plugins.cc` and `tests/nnstreamer_split/runTest.sh`, issue #4927, and re-grepped the repository for `tensorpick`/`tensor_split` usage. I also read the diff of the still-open #4962 (which touches the restart path of the same element) to check whether this PR's new documentation stays correct regardless of merge order.

## Correctness of the fix

Confirmed independently by reading `gst_tensor_split_get_tensor_pad()`, `gst_tensor_split_get_splitted()` and `gst_tensor_split_chain()`:

- Before the fix, a new pad's caps came from `tensorseg[split->num_srcpads]` (pad-creation ordinal), while `gst_tensor_split_get_splitted()` always cut the buffer from `tensorseg[nth]` (segment index). These only agreed when `tensorpick` was unset or a contiguous run starting at 0 — exactly the #4927 bug.
- After the fix, `dim = g_array_index (tensorseg, tensor_dim *, nth)`, so caps and payload use the same index.
- The removed `split->num_srcpads >= tensorseg->len` bounds check is provably dead: `PROP_TENSORSEG` in `gst_tensor_split_set_property()` (gsttensor_split.c:633-670) always installs a `tensorseg` array of exactly `num` entries and sets `split->num_tensors = num` in the same `GST_OBJECT_LOCK`/`UNLOCK` block. `gst_tensor_split_chain()` (gsttensor_split.c:477-481) snapshots `num_tensors` and `tensorseg` under one lock before the `for (i = 0; i < num_tensors; i++)` loop, and `nth` is always this same `i`. So `nth < num_tensors == tensorseg->len` holds at every call, and `g_array_index` cannot go out of bounds. Dropping the now-dead `if (srcpad == NULL)` branch in `chain()` is consistent, since `get_tensor_pad()` can no longer return `NULL`.

## Regression risk

Grepped every `tensorpick`/`tensor_split` occurrence in the tree. The only other live combination is `tests/nnstreamer_sink/unittest_sink.cc:1043` (`tensorseg=1:160:120:1,...` with `tensorpick=0,1,2`, a full sequential pick over equal-sized segments), where `nth == pad ordinal` and behavior is unchanged. `tensor_demux` has its own independent pad-creation path and is untouched. The change is isolated to `gsttensor_split.c` plus its own doc/tests; no other module is touched.

## Test coverage / CI gating

The GTest cases (`segmentsWithoutTensorpick`, `tensorpickSkippingFirstSegment`, `tensorpickWithGap`, `addPadAfterTensorsegShrinks`, `tensorpickPastBuffer_n`, `tensorpickNegotiatesPickedSegment`, `tensorpickRefusesFirstSegment_n`) check caps dimension, caps-declared size, actual buffer size and first byte per pad. I hand-traced the offset/size math in `tensorpickSkippingFirstSegment` (tensorseg=`1:8:8,2:8:8`, tensorpick=`1`: segment 0 is 64 bytes, segment 1 is 128 bytes at offset 64) and `tensorpickWithGap`'s `2,0`-order case against `gst_tensor_split_get_splitted()`'s offset accumulation; the expected values are correct. SSAT `nnstreamer_split` case 8/8_n exercises the same defect end-to-end via caps negotiation. `meson test` runs in the Ubuntu meson workflow and the GBS `x86_64` `%check` runs SSAT with the `_n` convention on every PR, so a regression here would fail CI, not just a manual run.

## Documentation accuracy (the subject of this round's changes)

I checked every statement the updated docs make against the actual code and traced runtime scenarios by hand:

- **`gst_tensor_split_get_tensor_pad()` only sets caps when a pad is newly created** (gsttensor_split.c:344-359); a reused pad (the `nth == pad->nth` branch, line 290-299) returns early without touching caps. This backs the Limitations section's central claim ("a source pad takes its caps once, when it is created, and keeps them").
- **`tensorseg` reshape while playing:** `PROP_TENSORSEG` replaces `split->tensorseg`/`num_tensors` without removing existing src pads (those are only removed in `gst_tensor_split_remove_src_pads()`, called from `PAUSED_TO_READY`). Tracing `tensorseg=1:4:4,2:4:4` → `2:4:4,1:4:4`: `src_0` (nth=0) keeps its original `1:4:4`/16-byte caps, but the next `chain()` call computes segment 0's size from the *new* tensorseg (`2:4:4` = 32 bytes) via `gst_tensor_split_get_splitted()`. Matches the doc's `16 bytes vs 32 bytes` example exactly.
- **Upstream type change while playing:** `GST_EVENT_CAPS` in `gst_tensor_split_event()` updates `split->in_config` unconditionally, which feeds `gst_tensor_split_get_splitted()`'s size calculation, but pads already have fixed caps (`gst_pad_use_fixed_caps()`) set at creation time and are never re-negotiated. `uint8`→`float32` is exactly a 4x element-size change (`gst_tensor_get_element_size`), matching the doc's "four times as many bytes."
- **`tensorpick` change while playing:** traced `tensorpick=1` → `0,1`: the existing pad (`nth=1`, named `src_0` since it was created first) is reused as-is; the new pick value `0` creates a *new* pad (`src_1`, since `num_srcpads` is now 1) carrying `nth=0`. Matches the doc's "segment 1 stays on `src_0`, segment 0 goes out on `src_1`" exactly, and correctly explains why the general "ascending index order" claim (both in the SECTION comment and the `tensorpick` property blurb) is now properly scoped to "set before the stream starts."
- **Restart guidance ("stop to READY or NULL, set tensorseg and tensorpick, and play again") stays accurate independent of merge order with #4962.** On `main`+this PR alone, `PAUSED_TO_READY` resets `num_tensors` to 0 (`gsttensor_split.c` `gst_tensor_split_remove_src_pads()`), so *not* re-setting `tensorseg` after a restart would silently produce zero pads — the doc's explicit instruction to re-set `tensorseg`/`tensorpick` is required and correct today. If #4962 (which stops resetting `num_tensors` on `PAUSED_TO_READY`) merges first or later, re-setting `tensorseg`/`tensorpick` becomes optional but remains valid and correct, so the instruction does not go stale either way.
- **The `tensorpick` property description stays a single sentence** (gsttensor_split.c:136-140, semicolon/comma-joined clauses, no embedded period), as required.
- **`component-description.md`'s link change** (`.c` → `.md` for tensor_split) matches the file's existing convention: elements with a dedicated doc (`tensor_if`, `tensor_transform`, `tensor_src_iio`) link to `.md`; elements without one still link to `.c`.
- **`Documentation/hotdoc/sitemap.txt`** now lists `gsttensor_split.md` before `gsttensor_src.md`, which is alphabetically correct (`split` < `src`, since `p` < `r`), resolving the earlier nit.

I could not find any remaining inaccurate statement in the touched documentation, and no architecture/API surface changed beyond the two doc-only PROP_TENSORPICK/SECTION string edits and the new markdown content, so no further doc updates are owed.

## Findings

None. This round's changes are documentation/comment-only (the `tensorpick` property string and the SECTION comment in `gsttensor_split.c`, plus `gsttensor_split.md` and `component-description.md`); the functional code is unchanged from the previously-reviewed and CI-verified `45597e79`. I independently re-verified both of `myungjoo-bot`'s prior nits are correctly and accurately resolved, not merely restated.

## Verdict

**No issues found — OK to merge as is.** The fix is minimal and correct, isolated to `tensor_split`, backed by tests that fail without it and are gated in CI, and the documentation added/amended in this round is accurate against the element's actual behavior (including the three runtime-mutation edge cases now called out as explicit Limitations) and remains accurate regardless of whether #4962 merges before or after this PR.

myungjoo:

Follow-up for the limitation documented in this PR is filed as #4967.

It proposes two changes: refuse `tensorseg` and `tensorpick` changes while source pads exist, and re-send pad caps when the input is renegotiated. It also lists the tests those changes need. This PR itself is unchanged.

myungjoo:

Thanks for the re-review. On its two nits, at 6e8d547e (force-pushed; each change folded into the commit it belongs to; documentation only):

- **NIT 1, "input shape or type":** fixed.
  - The *Limitations* entry in `gsttensor_split.md` now says a new input **type or framerate** does not reach the existing pads' caps. It adds that a new input shape alone leaves the caps right, since a pad's dimensions come from `tensorseg`, and that an input too small for the segments is refused with an error.
  - The intro says "in the cases below no error or warning is given", so it no longer contradicts that refusal.
  - The SECTION comment in `gsttensor_split.c` and `component-description.md` say "the input type or framerate" instead of "the input stream".
  - The doc commit message and the body of #4967 carry the same correction.
- **NIT 2, restart recipe depends on #4962:** #4962 has since been merged (7fc98e23), so a stop and replay now works without setting `tensorseg` again. No wording change was needed.

Verified locally: gst-indent and doxygen-tag are clean, gst-inspect output is unchanged, all 15 `testTensorSplit` cases and SSAT `nnstreamer_split` pass, and `git merge-tree` against current `main` (which includes #4962) is clean.

myungjoo:
**Automated review disclosure:** This comment was produced by a separate Claude agent (not the PR author acting directly) and is being relayed here via the author's (`myungjoo`) GitHub account, since that is the only account available to post it. Treat it as an independent automated review, not a statement from the human author.

## Scope

Fresh, independent review of PR #4954 at head `6e8d547e` (rebased on current `main`, which already includes #4962 `7fc98e23`). I read the full diff (`gst/nnstreamer/elements/gsttensor_split.c`, the new `gsttensor_split.md`, `Documentation/component-description.md`, `Documentation/hotdoc/sitemap.txt`, `gst/nnstreamer/elements/README.md`, `tests/nnstreamer_plugins/unittest_plugins.cc`, `tests/nnstreamer_split/runTest.sh`), issue #4927 in full, all prior review rounds and their resolutions, the related #4925/#4962/#4967 background, and re-grepped the repository for `tensorpick`/`tensor_split` usage. I did not simply restate the prior rounds' conclusions; I re-derived each one from the code myself.

## Does it solve #4927?

Yes. `gst_tensor_split_get_tensor_pad()` (gsttensor_split.c:318) now computes `dim = g_array_index (tensorseg, tensor_dim *, nth)`, the same index `gst_tensor_split_get_splitted()` (line 422) uses to cut the buffer — this is exactly option 1 from the issue's own "decision needed" section, and matches the issue's stated preference ("the smaller change... matches `gst_tensor_split_get_splitted()`"). The reproduction case in #4927 (`tensorseg=1:4:4,2:4:4 tensorpick=1`) now produces caps and payload that agree.

The removed `split->num_srcpads >= tensorseg->len` bounds check (previously added by #4925) is provably dead: `PROP_TENSORSEG` in `gst_tensor_split_set_property()` (lines 634-671) always sets `split->tensorseg` to an array of exactly `num` entries and `split->num_tensors = num` under one `GST_OBJECT_LOCK`/`UNLOCK`. `gst_tensor_split_chain()` (lines 478-482) snapshots both under one lock before the `for (i = 0; i < num_tensors; i++)` loop, and `nth` passed to `get_tensor_pad()` is always that same `i`. So `nth < num_tensors <= tensorseg->len` holds at every call site, and `g_array_index` cannot run out of bounds. Dropping the now-unreachable `if (srcpad == NULL)` branch in `chain()` (removed lines 511-114 of the old code) is consistent, since `get_tensor_pad()` can no longer return `NULL`.

Isolated to `gsttensor_split.c` and its own docs/tests; nothing else is touched. Grepped every `tensorpick`/`tensor_split` occurrence in the tree:
- `tests/nnstreamer_sink/unittest_sink.cc:1043` uses `tensorseg=1:160:120:1,1:160:120:1,1:160:120:1 tensorpick=0,1,2` — a full, sequential pick over equal-sized segments, so `nth == pad ordinal` there and behavior is byte-for-byte identical before and after.
- `tests/nnstreamer_decoder_pose/runTest.sh` uses `tensor_split` without `tensorpick`, unaffected.
- `gsttensor_if.c`'s `tensorpick` hit is an unrelated enum member on a different element.
- `tensor_demux` (`gsttensor_demux.c`) has its own independent pad-creation path and is untouched.

No in-tree pipeline or test depends on the old (buggy) ordinal-based numbering, so I see no regression risk.

## Test coverage

The new GTest cases (`segmentsWithoutTensorpick`, `tensorpickSkippingFirstSegment`, `tensorpickWithGap`, `addPadAfterTensorsegShrinks`, `tensorpickPastBuffer_n`, `tensorpickNegotiatesPickedSegment`, `tensorpickRefusesFirstSegment_n`) assert caps dimension, caps-declared size, actual buffer size, and first byte per pad. I hand-traced the offset/size math for `tensorpickSkippingFirstSegment` (`tensorseg=1:8:8,2:8:8`, `tensorpick=1`: segment 0 is 64 bytes, segment 1 is 128 bytes at offset 64) and `addPadAfterTensorsegShrinks` against `gst_tensor_split_get_splitted()`'s offset accumulation; the expected values are correct. This is a solid, byte-level regression net for the exact #4927 defect. The two pipeline-level GTest cases and the SSAT `nnstreamer_split` 8/8_n pair add an end-to-end caps-negotiation check (a capsfilter requiring the picked segment's real shape links; one requiring the ordinal segment's shape is refused), which is exactly the "no error or warning is given today" failure mode the issue describes — this is the right layer to catch it, since it observes what a real downstream element would see. Case `8_n` follows the `_n` negative-id convention this repo uses for SSAT.

Regarding whether a *future* change elsewhere (e.g. #4967's proposed caps-renegotiation work) would be caught: the new Limitations-section behavior is itself now documented and partially covered by tests only for the "add a new pick" runtime case indirectly (via `addPadAfterTensorsegShrinks`), not for the `tensorseg`-reshape-while-playing or type-change-while-playing cases described in the doc. That is fine — those are pre-existing, separately-tracked behaviors (issue #4967), not introduced or fixed by this PR, so it is reasonable that this PR does not add regression tests for them.

## CI gating

`meson test -C build/ -v` runs in `.github/workflows/ubuntu_clean_meson_build.yml` and will pick up the new GTest cases; `.github/workflows/gbs_build.yml`'s `x86_64` leg runs with `unit_test 1` on every PR, and its packaging `%check` runs SSAT with the `_n` negative-id convention. A future regression in this exact defect would fail CI on a normal PR, not just on a manual run. `gh pr checks 4954` shows all currently-reported checks passing (several long-running jobs were still queued/pending at the time of this review, which I am not treating as a finding per the review instructions).

## Documentation accuracy (checked line-by-line against the code, now with #4962 on `main`)

- `gsttensor_split.c:33-38` (SECTION comment) and the `tensorpick` property blurb (`:137-141`) both correctly qualify pad numbering as "ascending segment index when tensorseg and tensorpick are set before the stream starts." I independently traced the runtime-reorder scenario (`tensorpick=1` → `0,1`): the existing pad keeps `nth=1` as `src_0`, and the newly created pad for `nth=0` becomes `src_1` (via `name = g_strdup_printf ("src_%u", split->num_srcpads)` at line 306, with `num_srcpads` already at 1). This matches the doc exactly.
- The `tensorpick` property description (`gsttensor_split.c:137-141`) is a single sentence (semicolon/comma-joined clauses, no embedded period) — confirmed.
- `gsttensor_split.md`'s Limitations section: I independently traced all three listed scenarios against the code (caps set only in `gst_tensor_split_get_tensor_pad()`, called once per `nth` due to the `walk`/`nth == pad->nth` early-return at lines 290-299; `gst_tensor_split_get_splitted()` at line 422 always uses the *current* `tensorseg`/`in_config` at call time) and the byte counts given (`1:4:4,2:4:4`→`2:4:4,1:4:4` giving 16 vs 32 bytes; `uint8`→`float32` giving 4x bytes) check out arithmetically and against the code path.
- The "input too small is refused with an error" clause matches the `offset + size > src_info.size` check in `gst_tensor_split_get_splitted()` (line 446), and does not contradict the "no error or warning is given" framing, since that framing is explicitly scoped to the three bulleted cases (stale caps), not to an undersized buffer.
- `component-description.md:111,114-115` and the sitemap/README links are consistent with the file's own existing convention (elements with a dedicated `.md` link to it; those without still link to `.c`), and `gsttensor_split.md` is placed alphabetically correctly before `gsttensor_src.md` in the sitemap.
- The restart guidance ("stop to READY or NULL, set `tensorseg` and `tensorpick`, and play again") remains accurate now that #4962 is merged: it describes how to *apply a change* to those properties, not a claim that re-setting them to the same value is mandatory, so it does not go stale either way `#4962` merges.

I found no remaining inaccurate statement in the touched documentation, and no undocumented architecture/API surface change (no new/removed properties, signals, or pad templates) beyond what is already covered here.

## Scope / size

The diff is tightly scoped to `tensor_split` and its own docs and tests; nothing else is modified. The size (roughly +390/-30 across one element, one new doc, and its tests) is proportionate to the bug fixed plus the coverage the issue itself asked for ("gain a case with differently-sized segments and a tensorpick that does not start at 0").

## Merge status

`gh pr view --json mergeable,mergeStateStatus` reports `MERGEABLE` against current `main`. I confirmed no conflict markers in a `git merge-tree` of the merge-base, PR head, and `origin/main` (which includes both #4962 and the unrelated, already-merged filter change #4957/#4958) — the only file-level differences shown were in unrelated CI workflow history, not in anything this PR touches.

## Findings

None. I found no correctness, memory-safety, thread-safety, scope-creep, regression, or documentation-accuracy issues in this independent pass over the current head.

## Verdict

**OK to merge as is.** The fix is minimal, correct, and matches the exact remediation the issue itself proposed. It is backed by tests that fail without it (per the prior rounds' mutation testing, which I have no reason to doubt given my own independent trace of the same code paths) and are gated in CI at both the GTest and SSAT layers. The documentation added and amended in this PR is accurate against the element's actual behavior, including the runtime-mutation edge cases now called out under Limitations, and stays accurate with #4962 already on `main`.

PR #4950, opened 2026-09-10
[Element/If] Bound the compared value to the tensor it is read from

myungjoo:
Fixes item **C17** of #4920.

## The defect

`tensor_if compared-value=A_VALUE` picks a single element of the nth tensor, addressed by `compared-value-option=[C]:[W]:[H]:[B],n`. `gst_tensor_if_calculate_cv()` built the byte offset from that option alone: no index was ever compared with the dimension it indexes, and the resulting offset was never compared with the mapped memory.

On a `3:4:2:2` int32 tensor (192 bytes):

| `compared-value-option` | on main |
| --- | --- |
| `0:0:0:2,0` | reads the first element past the buffer |
| `0:4:0:0,0` | reads an element of another row, silently |
| `-1:0:0:0,0` | the index wraps to `0xFFFFFFFF`, the process dies |
| `2000000000:0:0:0,0` | SIGSEGV on the first buffer |

Reproduced on main with a plain pipeline:

```
gst-launch-1.0 videotestsrc num-buffers=1 ! videoconvert ! videoscale ! video/x-raw,format=RGB,width=160,height=120 ! tensor_converter ! tensor_if compared-value=A_VALUE compared-value-option=2000000000:0:0:0,0 supplied-value=100 operator=GT then=PASSTHROUGH else=SKIP ! fakesink
```

`Caught SIGSEGV` before this change, a refused buffer after it.

## The fix

- An element index has to be inside the dimension it indexes; an index into a dimension the tensor does not have (dimension `0`) has to be `0`.
- The byte offset is accumulated in 64 bits.
- After the memory is mapped, the element has to fit in it. The comparison is `in_info.size < esize || idx > in_info.size - esize`, which cannot wrap: caps may declare far more data than the buffer carries, and with `dimension=4294967295:4294967295:2` uint8 and `0:2:1,0` the offset is `2^64 - 1`, so the earlier `idx + esize > in_info.size` wrapped to 0 and read the byte before the buffer.
- With GLib 2.48 or later (`#if GLIB_CHECK_VERSION (2, 48, 0)`), the offset arithmetic uses `g_uint64_checked_mul()` / `g_uint64_checked_add()`, so caps whose offset does not fit in 64 bits are refused instead of wrapping back onto an in-bounds element nobody asked for. Older GLib keeps the plain arithmetic, where the non-wrapping size check still keeps every read inside the mapping.
[...]

PR #4949, opened 2026-09-10
[Common] Refuse a tensor count that does not fit the info instead of dereferencing nothing

myungjoo:
Items **A4** and **B4** of #4920.

## The defect

`gst_tensors_config_from_structure()` stored whatever integer a caps structure
carried in `num_tensors`, with no ceiling and no sign check. Three helpers then
walk that count and dereference the NULL that `gst_tensors_info_get_nth_info()`
returns for an index of `NNS_TENSOR_SIZE_LIMIT` or more:
`gst_tensors_info_get_rank_dimensions_string()`,
`gst_tensors_info_get_types_string()` and `gst_tensors_info_get_names_string()`.

The element that asks is the one that dies. `tensor_sparse_dec` reads the peer's
count on every caps event (`gst_tensors_config_from_peer()`, no validation) and
hands the config straight to `gst_tensor_pad_caps_from_config()`, which builds
the `dimensions` and `types` strings from it.

**B4** is a second producer of the same shape:
`gst_tensor_filter_common_get_combined_out_info()` had no ceiling on the
combined index while `_get_combined_in_info()` did, so an `output-combination`
of more than `NNS_TENSOR_SIZE_LIMIT` entries fetched a NULL destination.
`gst_tensor_info_copy()` refused that copy with a GLib critical, so nothing was
written through it, but the info was left reporting 257 tensors - which then
reaches the A4 helpers through `transform_caps`.

## The fix

- Refuse a static structure whose count is not between 1 and
  `NNS_TENSOR_SIZE_LIMIT` *before anything is stored*, so a caller that ignores
  the return value (tensor_transform, tensor_converter, tensor_query_client and
  tensor_sparse_dec all do) finds the config as initialized.
- The three string helpers test the same bound up front and return NULL. The
  caps builder already treats a NULL string as "no such field", so no caller
  gains a new failure path.
- The caps builder (`_get_tensors_caps()`) fills the tensor fields only for a
  count it can describe, so an over-limit info keeps the template's `[1, 256]`
  range instead of advertising a fixed count no peer can parse.
- *(Separate commit, `[Filter]`.)* Both combination property setters clear
  `errno` before each `g_ascii_strtoull()`; they tested `errno == ERANGE`
  without clearing it, so an
  `ERANGE` left over from any earlier call refused a valid combination.
- Both combination helpers test the index *before* the copy. That also admits a
  combination of exactly `NNS_TENSOR_SIZE_LIMIT` entries, which the input twin
[...]

PR #4955, opened 2026-09-10
[CI/Valgrind] Fail on the definite leaks that come from this repository

myungjoo:
Follow-up of #4936, which made the valgrind step fail on memcheck **errors** from our own code while only counting **leaks**. With this PR, a **definite leak** from this repository fails the step too.

The PR has two phases in separate commits:
1. The checker learns to attribute leaks, and warns only.
2. Every leak it finds on main is fixed, then the switch is flipped.

### Phase 1: attribute leaks (1 commit)
`tools/debugging/check_valgrind_log.sh` attributes each **"definitely lost"** leak to the innermost frame of ours in its allocation stack. Errors use the first frame instead; leaks differ because memcheck reports a leak where the block was allocated, which is almost always inside GLib.

Two callers are exempt when they sit between the allocation and our frame:
- **The dynamic loader (`_dl_open`, `_dl_init`).**
- **`gst_debug_add/remove_log_function`**, whose list GStreamer leaks on purpose.

Two further rules and one limit:
- **ggml backends excluded.** The prebuilt llama.cpp backends (`libggml*`) that CI copies into `build/tests` are excluded by name from the object half of "ours", for leaks and errors alike.
- **Possible and indirect leaks** stay counted only.
- **Limit, documented in the header:** a block allocated wholly by library code that our code then drops is not visible to the rule.

### Phase 2: fix what it finds, then gate (8 commits)
| Commit | Leak | Owner |
|---|---|---|
| `[Test] Free what the test cases under memcheck allocate` | cpp_methods `base01_n`/`base02_n` never set their pipeline to NULL; tensor_region paths; rate model path; armnn buffers and tensors info; common `printInvalidIndex_n` extra infos; filter_custom flexible-input harnesses (#4953) whose undrained private bus kept the harness alive | tests |
| `[Element/If] Release the buffer a condition cannot be checked on` | `gst_tensor_if_chain` returned `GST_FLOW_ERROR` without unreffing its buffer; #4950 made that reachable | shipped |
| `[Decoder/OctetStream] Free the bytes each decoded buffer carries` | `gst_memory_new_wrapped (..., NULL, g_free)` calls `g_free (NULL)`, so every decoded payload leaked | shipped |
| `[Filter/LiteRT] Free a custom property pair whose value is refused` | pair lost when `parseAcceleratorValue` throws | shipped |
| `[Filter] Release async output dispatched after tensor-filter stopped` | `nnstreamer_filter_dispatch_output_async` dropped the data it owns when no callback was registered (tensor_filter_single stop vs. llamacpp's thread) | shipped |
| `[Common] Release the collected memories when collect-pad sync gives up` | early returns of `gst_tensor_time_sync_buffer_from_collectpad` (#4920 **A3**): 76 GstMemory blocks in unittest_sink | shipped |
| `[Test/Filter] Stop the custom-easy pipelines before unregistering the model` | not a leak but a use-after-free the #4936 gate fails on when a runner is slow: four cases unregister the model while join still streams | tests |
| `[CI/Valgrind] Fail on a definite leak that comes from this repository` | the switch | CI |

### Verification
- **Local memcheck:** rebased on main 7bfaa872, same flags as CI, llamacpp and litert enabled, run over every unittest binary in parallel with CI's case filters.
  - Every binary of CI's memcheck set reports **0 definite leak contexts and 0 errors of ours**.
  - unittest_filter_custom was re-run twice after the pipeline fix, clean both times.
  - armnn is not built locally, so that binary is left to CI.
- **Negative controls:** I reverted the new shipped fixes while keeping the tests. Each case then fails as follows:

| Test case | Result without the fix |
|---|---|
| `dispatchWithoutCallback_n` | fails on its assertions |
| `pushOctetStream` | the gate fails on 768 B |
| issue739MergeParallel1 | the gate fails on 19 blocks |

- **Checker self-test:** 99/99 under gawk 5.3 and mawk 1.3.4. Each of 9 mutations fails at least one case, including dropping the ggml exclusion and letting leaks pass.
- **Error half unchanged:** across 21 saved CI logs, including the last runs of #4949, #4950, #4953 and the crop and pose PRs, the exit status and error list are identical to main's checker.

### Interactions with open PRs
Checked with this checker against their latest CI logs, these open PRs add definite leaks of ours. Their valgrind step will fail once this lands, until those leaks are fixed:
- **#4957:** `_flex_new_harness` (`unittest_filter_custom.cc`) and two TestBody frames.
- **#4958:** `notConfiguredWhileStreaming_n` and shipped `gst_tensor_transform_transform`.

**#4961** edits the same exit line of the checker. Whichever lands second needs a trivial rebase; the combined line is `exit (unfinished > 0 || leak_ours_count > 0)`.


myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Summary

Phase 1 of the leak gate does what it says: it attributes each *definitely lost* leak to either "this repository" or "a library", using the innermost frame of ours in the allocation stack (with two named exemptions for the dynamic loader and `gst_debug_add/remove_log_function`), lists and `::warning::`s the ones of ours, and — as stated — never fails the build on them. The error-counting half is untouched.

Verification performed:
- Ran `tools/debugging/test_check_valgrind_log.sh`: 77/77 pass under both gawk 5.3 and real mawk 1.3.4 (the CI runner's awk), matching the PR's claim.
- Ran the shipped checker against the actual CI log from main run 34443320625 (referenced in the PR body): output is exactly "16 definite leak contexts from this repository", matching the PR description's table.
- Confirmed the workflow integration is unaffected: `ubuntu_clean_meson_build.yml` and `static.check.yml` are untouched, and the step's pass/fail still depends only on the exit code (driven solely by `ours_count`, i.e. non-leak errors) — the new leak listing and `::warning::` lines are additive output only, not parsed anywhere else in the repo.
- Scope is proportionate: two files touched, both directly `tools/debugging/`, no other module affected. No architecture/API changed, so no documentation update is required (confirmed: no `Documentation/` file references this script either before or after).

## Findings

### 1. (Medium) Leak attribution can fall through to a non-nnstreamer frame when no exemption matches

`is_ours()` treats any frame located under `/build/(gst|ext|tests)/` as "ours" when it carries no debug info, a heuristic inherited unchanged from the error path added in #4936. For a single-frame error match this is fairly safe, but the new leak scan walks *every* frame of the allocation stack looking for the first "ours" or exempt hit, which is a much bigger surface for that heuristic to misfire on.

Reproduced concretely: if the two named exemptions (`_dl_init`/`_dl_open`, `gst_debug_add/remove_log_function`) are disabled and the shipped checker is rerun against the same CI log the PR cites (run 34443320625), several leaks stop resolving to a real `file:line` and instead resolve to `testing::internal::HandleExceptionsInMethodIfSupported(...)` reported as `(in .../build/tests/unittest_X)` — i.e. googletest's own internal frame, statically linked into the test binary, gets counted as "this repository's" leak purely because it lives inside a `build/tests/...` object with no debug info.

This doesn't affect today's shipped output (all 16 listed contexts currently resolve to real nnstreamer/test `file:line` frames), and today's two exemptions were clearly chosen to cover exactly the patterns seen in the current log. But the exemption list is a fixed set of symbol names, not a general "this is library-only" rule, so a *future* leak whose backtrace bottoms out in some other not-yet-seen library/glib call path (not `_dl_init`/`_dl_open`/`gst_debug_*`) and never touches a debug-info-carrying frame of ours would silently be attributed to the repository via this fallback — which becomes a real problem once phase 2 turns this into a CI gate, since it could fail the build on a leak that isn't actually nnstreamer's.

Suggestion: either scope the `/build/.../` fallback match to exclude known third-party in-binary code (e.g. googletest internals), or at minimum add a self-test that pins today's fallback behavior for a leak with no debug-info frame anywhere in its stack, so a similarly-shaped case doesn't quietly start gating CI on a library's bug once the switch flips.

### 2. (Low) "Without the exemptions it would list 35" doesn't reproduce

Re-running the same referenced log (34443320625) with `is_exempt_from_leaks()` stubbed to always return false gives 13 deduplicated contexts (fewer than the shipped 16, for the reason in finding 1 — several leaks collapse into one generic per-binary frame) or 20 raw (non-deduplicated) hits — not 35 under any variant tried. This doesn't affect the shipped code or tests (which are correct and verified), just the PR body's narrative; worth double-checking how that figure was produced so reviewers relying on the write-up aren't misled.

### 3. (Low / informational) Test coverage gap tied to finding 1

All 13 new leak self-test cases attribute "ours" via a source `file:line` frame; none exercises the `where ~ /^in .*\/build\/(gst|ext|tests)\//` (no-debug-info object) branch for a *leak*. Adding one — a leak whose only matching frame is a bare object path under `build/tests/`, mirroring the existing error-side "an object merely built somewhere is not ours"-style case — would double as a regression guard for finding 1.

## Not flagged, for context

- Phase 1 intentionally never fails CI on a leak of ours, per the PR's own description — this is by design, not an oversight, and is explicitly gated behind a future phase 2 once the existing 16 contexts are cleaned up. Flagging only so it's visible that, as of this PR, a newly introduced leak in shipped code still cannot block a merge.
- The per-binary+location leak dedup (`record_leak` keys on `binary, "definite leak", where` — not on the leak's size/kind) is deliberate and well covered by the `leak_repeated` test; not an issue.

Overall: the core mechanism is correct, well-tested against both awk implementations CI actually uses, and appropriately scoped as a non-gating first phase. The main thing worth addressing before phase 2 turns this into a hard gate is finding 1 (and, ideally, a test for it), since that is exactly the kind of edge case that would only surface as a CI-blocking false positive after the switch flips.

myungjoo:
Response to the relayed review above.

**Findings 1 and 2 come from how the mutant was run, not from the checker.** A copy of the checker placed outside the tree derives its repository root from its own location. So unless `--repo-root` is passed, the source list is empty and no `file:line` frame counts as ours. Only object paths under `build/` still match, which is exactly why gtest's `HandleExceptionsInMethodIfSupported` showed up. Running the same log (34443320625) with `is_exempt_from_leaks` stubbed to 0:

| Invocation | Contexts | gtest frames listed |
|---|---|---|
| `--repo-root <repo>` | **35** (the figure in the PR body) | 0 |
| no `--repo-root` (root = the copy's `../..`) | **13** (the reviewer's figure) | only gtest frames |

**The fall-through described in finding 1 cannot occur for a leak that test code drives.** The leak scan stops at the *innermost* frame of ours. Every stack gtest drives reaches the test's own `TestBody`/`SetUp`, a source frame of ours, before gtest's harness frames, which sit further out. CI also runs memcheck with `--num-callers=200` (`packaging/run_unittests_binaries.sh`), so a stack is not cut off between the two. The only way a gtest frame wins is when gtest itself allocates the block. None of the 172 leak reports in that log does, and treating that case differently would change the object half of `is_ours()` that #4936 set for errors as well. I have left that out of this PR.

**Finding 3: agreed, done in d8bd36b3.** It adds two cases:
- `leak_object_ours`: a leak whose innermost frame of ours is an object under `build/gst`, with a source frame below it that must not be the one listed.
- `leak_object_library_n`: an object built under someone else's `build/`.

Removing the object half for leaks fails the first. Now 81/81 under gawk 5.3 and mawk 1.3.4.

The "not flagged" note is right: this phase intentionally cannot block a merge on a leak.

myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Scope of this pass

This is a follow-up to the earlier relayed review on this PR. I re-verified the earlier findings, the author's response to them, and the new commit (d8bd36b3), then re-read the full current diff (a6fa2094 + d8bd36b3) as a whole.

Verification performed:
- Ran `tools/debugging/test_check_valgrind_log.sh` under gawk 5.3.2: 81/81 pass (one run hit a transient `/usr/bin/find: Permission denied` from the local Windows filesystem, unrelated to the script — reran clean twice).
- Ran the same self-test under mawk 1.3.4 via WSL, matching the CI runner's awk exactly: 81/81 pass.
- Ran the shipped checker against the real CI log referenced in the PR body (main run 34443320625) with the correct `--repo-root`: output is exactly "16 definite leak contexts from this repository", and the listing matches the PR body's table (6 shipped-code contexts, 10 test-code contexts) frame-for-frame.
- Reproduced the author's rebuttal numerically: with `is_exempt_from_leaks` stubbed to always return false and the correct `--repo-root`, the same log yields exactly **35** contexts with **zero** gtest-harness frames in the listing — matching the "35" figure in the PR body and directly contradicting the earlier review's "13, with gtest frames" reproduction. That earlier reproduction ran a copy of the checker outside the tree without `--repo-root`; since the script derives `repo_root` from its own location (`${SCRIPT_DIR}/../..`), that invocation either finds an unrelated/empty source list or (in my own attempt to reproduce it) errors out with "no source file found". Either way, it isn't evidence about the checker's behavior in its real CI invocation, where `--repo-root` is never needed because the script runs from its own checkout. **Findings 1 and 2 from the earlier review are settled — they were reproduction artifacts, not checker bugs.**
- Confirmed the author's structural argument for why a gtest-harness leak can't currently win: the scan attributes each leak to the *innermost* matching frame while walking the stack from the allocation outward, and every one of the 16 real leaks in the reference log resolves to a `TestBody`/`SetUp`-style source frame or shipped-code frame of ours — none reaches a gtest object frame. This is consistent with the code (`record_leak` fires and stops the scan on the first `is_ours` hit) and with CI's `--num-callers=200` (`packaging/run_unittests_binaries.sh`), so stacks aren't truncated before reaching that frame.
- Checked commit d8bd36b3 against finding 3: it adds `leak_object_ours` (innermost frame of ours is a no-debug-info object under `build/gst`, with a source frame below it that must not be the one listed) and `leak_object_library_n` (an object under someone else's `build/...` path). Both are meaningful, targeted regression guards for the previously-untested branch of `is_ours()` on the leak path, and both pass under gawk and mawk.
- Confirmed scope stays proportionate: only `tools/debugging/check_valgrind_log.sh` and its self-test are touched; `.github/workflows/ubuntu_clean_meson_build.yml` and `.github/workflows/static.check.yml` are unchanged in this diff, and the self-test step in `static.check.yml` (pre-existing, from #4936) still gates on the script's exit code, so a future regression in this attribution logic will fail CI, not just print a warning.
- No architecture or public API changed (this is an internal CI helper script), and no `Documentation/` file references this script before or after this PR, so no doc update is owed.

## Assessment

No new issues found in this pass. The core mechanism (leak attribution by innermost frame, the two named exemptions, and the non-gating warning-only reporting for phase 1) is correct and matches its own stated design in the header comment. The response to the earlier review is accurate and fully supported by re-running the exact experiments against the exact CI log the PR cites, under both gawk and the real CI mawk. Finding 3's test-coverage gap is closed by d8bd36b3 with well-targeted cases.

**Remaining nit (informational, not blocking):** the in-script comment block still doesn't mention `--num-callers=200` as the reason a leak's allocation stack is never cut off before reaching a frame of ours — that reasoning currently lives only in the PR's response comment, not in the script or its header. Worth folding into the header comment or a code comment near `is_ours`/`record_leak` at some point so a future reader doesn't have to rediscover it, but this is cosmetic and doesn't need to hold up this PR.

**As far as this review goes, this PR is OK to merge.** It does what it says (attributes and reports, without gating, definite leaks from this repository), doesn't touch or risk any other module, is proportionately scoped to two files under `tools/debugging/`, is well covered by a self-test that passes under the exact awk CI uses, and the one substantive concern raised earlier (finding 1, the fallback-attribution edge case) has been empirically shown not to apply to any leak in the current codebase and is now pinned by a regression test either way.

myungjoo:
Response to the second relayed review: the nit is addressed in 2073e761. The header now says:
- for a leak a test drives, the innermost frame of ours is the test's own frame, nearer the allocation than the gtest code linked into the same binary;
- a stack cut before any frame of ours goes unreported, so memcheck must be asked for enough callers (`packaging/run_unittests_binaries.sh` asks for 200).

This is a comment-only change; the self-test still passes 81/81.

myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Scope of this pass

Third pass, covering only what could have changed since the second relayed review: commit 2073e761, which answers that review's one remaining nit by expanding the header comment. No code line changed (confirmed via `git show 2073e761 -- tools/debugging/check_valgrind_log.sh`: 8 insertions / 3 deletions, entirely inside the `#` comment block). I re-verified the new prose against the code and against the rest of the repo rather than re-doing the mechanism review the first two passes already covered.

Verification performed:
- `git show 2073e761` touches only comment lines in `tools/debugging/check_valgrind_log.sh`; `tools/debugging/test_check_valgrind_log.sh` is untouched by this commit.
- Ran `tools/debugging/test_check_valgrind_log.sh` locally under gawk: 81/81 pass, unchanged from the prior pass.
- Ran the shipped checker against the same referenced CI log (main run 34443320625) with `--repo-root`: still exactly "16 definite leak contexts from this repository" — confirms the comment-only commit changed no behavior.
- Checked the new claim "packaging/run_unittests_binaries.sh asks for 200" callers: `grep -n num-callers packaging/run_unittests_binaries.sh` shows `--num-callers=200` on the memcheck invocation line. Accurate.
- Checked the new claim that possibly-lost leaks are "mostly the thread-local storage of threads still running at exit": sampled several "possibly lost" reports in the referenced CI log; each one examined bottoms out in `allocate_dtv (dl-tls.c:375)` → `_dl_allocate_tls` → `allocate_stack` → `pthread_create`, i.e., TLS allocated for a thread GLib/GStreamer spins up (e.g., via `g_thread_pool_push`) that valgrind can't prove is unreachable at exit. Matches the comment.
- The "indirect[ly lost leaks are] reachable only through a definite one" line restates Valgrind's own definition of indirectly-lost blocks, not a claim about this codebase — consistent with existing behavior (`leak_count` still counts these lines; only "definitely lost" feeds `record_leak`).
- The gtest-linkage explanation matches the structural argument the author already gave and the second review already verified from the code and the `--num-callers=200` fact.

## Assessment

No new issues. The added comment text is accurate, matches the code and `packaging/run_unittests_binaries.sh`, and is scoped to documentation only — no behavior, no test, no other file changed by this commit. The prior substantive finding (innermost-frame attribution potentially over-crediting to "ours" without a real exemption match) was already resolved by commit d8bd36b3's regression tests in the prior round, and this round doesn't reopen it.

**As far as this review goes, this PR is OK to merge.** No remaining nits from this pass.

myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Scope of this pass

The PR grew since the third relayed review (which covered only the comment-only commit `2073e761`, phase 1). It now carries 9 commits (head `238399d3`): the same 3 phase-1 commits already reviewed twice (rebased, content unchanged — `fdd1f8b7`/`3392a5d3`/`b93d640f` are `a6fa2094`/`d8bd36b3`/`2073e761` under new hashes, confirmed identical via `git diff` between the two triples), plus 6 new commits that fix every definite leak the checker found and then flip the gate. This pass reviews the 6 new commits on their own merits and the PR as a whole.

## Verification performed

- Ran `tools/debugging/test_check_valgrind_log.sh`: 91/91 under both real gawk and the CI-matching mawk 1.3.4 (via WSL, working around the known CRLF-under-`/mnt/c` trap by copying the scripts into WSL's native filesystem first).
- Built the PR head locally (llamacpp + litert enabled) and ran the actual affected tests under memcheck (`vgwrap2.sh`), then fed each log through the shipped checker with `--repo-root`:
  - `testTensorDecoder.pushOctetStream`: 0 leak contexts of ours (only the pre-existing, correctly-exempted `_dl_init`/GLib startup leak remains).
  - `nnstreamerFilterLiteRT.customProp00_n`: same — 0 leak contexts of ours.
  - `testTensorFilterAsync.dispatchToCallback` / `dispatchWithoutCallback_n`: same — 0 leak contexts of ours.
  - `unittest_sink`'s `issue739*Merge*`/`issue739*Mux*` (the tensor_mux/merge slowest/basepad cases the PR body cites): checker reports "0 definite leak contexts from this repository" for the whole run.
- Reproduced one negative control myself rather than just trusting the PR body: reverted the octet_stream fix (`mem_data` back to `NULL`) in the local build, reran `pushOctetStream` under memcheck — the checker now reports exactly "1 definite leak contexts from this repository" / "768 bytes ... `gst_tensordec_transform (gsttensor_decoder.c:840)`", i.e. it would fail CI. Matches the PR body's negative-control table exactly. Restored the fix afterward.
- Read every changed hunk in the 6 commits against the surrounding function bodies (not just the diff context):
  - `139ebd15` (`gst_tensor_time_sync_buffer_from_collectpad`): the three `return FALSE` sites turned into `goto error` all occur before `counting` is incremented for the pad being processed at that point, so the new `for (i = 0; i < counting; i++) gst_memory_unref(in_mem[i])` cleanup unrefs exactly the memories already gathered from earlier pads — no over-release, no leak. The re-`gst_tensors_config_free(&in_configs)` on the "not configured" path is not a double free: `gst_tensors_config_from_caps()` already frees and re-`gst_tensors_info_init()`s `config` internally on its own failure path before returning FALSE, so the struct is back to a clean, freeable state by the time `goto error` runs.
  - `8fd4b1fd` (octet_stream): straightforward — `gst_memory_new_wrapped(..., NULL, g_free)` was calling `g_free(NULL)` on every decode; now correctly passed `mem_data`.
  - `955ee0d8` (LiteRT): `g_auto (GStrv) option` correctly replaces the two manual `g_strfreev` call sites (including the one on the exception path), verified this also covers the loop's normal fall-through.
  - `050421a9` (filter async dispatch): `nnstreamer_filter_dispatch_output_async()`'s new cleanup loop bounds on `prop->output_meta.num_tensors`, which is safe today because the sole real caller (llamacpp) always sets `num_tensors = 1` to match its single-element `output` array — see below for a minor nit on this.
  - `b1b38e66` (test-only fixes): checked the two-call `armnn invokeAdvanced` sequence specifically for use-after-free/double-free risk from interleaving `gst_tensors_info_free(&res)` between the `getInputDimension`/`getOutputDimension` calls — the free is placed after each `res` is done being read and before it is repopulated, so no such risk.
  - `238399d3` (gate flip): confirmed the combined exit condition (`ours_count == 0 ? exit(leak_ours_count > 0) : exit(1)`) is equivalent to `exit(unfinished > 0 || leak_ours_count > 0)`, matching the PR's stated merge point with #4961; confirmed `.github/workflows/*.yml` are untouched so the step's pass/fail still depends only on this script's exit code.
- Confirmed scope stays proportionate: 13 files, all directly implicated (the 4 leak sites, their tests, and the checker). No unrelated modules touched.
- No architecture/public API change: the one API-adjacent edit (`nnstreamer_plugin_api_filter.h`) only extends an existing function's doc comment to describe the no-callback case, and the same sentence is mirrored in the source doc comment right above the implementation — nothing further needed in `Documentation/`.

## Findings

### 1. (Low, informational) Async-dispatch cleanup trusts `output_meta.num_tensors` to match the `output` array length, with nothing enforcing it

`nnstreamer_filter_dispatch_output_async()`'s new no-callback branch does:
```c
for (i = 0; i < prop->output_meta.num_tensors; i++)
  g_clear_pointer (&output[i].data, g_free);
```
This is correct for the only real caller today (llamacpp, which always sets `output_meta.num_tensors = 1` to match its single-`GstTensorMemory` `output`), and the new `dispatchToCallback`/`dispatchWithoutCallback_n` tests pin exactly that 2-tensor case correctly. But nothing ties `num_tensors` to the actual length of the `output` array the caller passes in; a future sub-plugin that dispatches more tensors than it declared in `output_meta` (or vice versa) would either under-free (leak, ironically on the very code path this PR just closed) or read past the array. This mirrors how the registered-callback path already trusts the same invariant (`prop->async_callback (output, &prop->output_meta, ...)`), so it's not a new class of risk introduced here, and the added doc-comment sentence is accurate as far as it goes — just doesn't spell out the invariant. Not blocking; worth a one-line note in the doc comment (e.g., "the caller-declared `output_meta.num_tensors` must match the length of `output`") the next time this function is touched.

## Not flagged, for context

- The three phase-1 commits are unchanged from the version already reviewed and approved twice (content-identical after the rebase); nothing new to add there.
- The `armnn` fixes in `b1b38e66` are for a sub-plugin not built in the local memcheck environment (per the PR body) and so could not be independently re-run under valgrind here; they were checked by inspection only (see above) and look correct.

## Assessment

No functional defects found in this pass. All four fixed leaks were independently reproduced as absent-with-fix / present-without-fix against the actual affected tests under memcheck, matching the PR's own verification table. The gate-flip commit is minimal, matches its own self-test, and its exit-condition math checks out. Scope is proportionate (13 files, all directly on-topic), and the one API-adjacent doc change is included in the same commit rather than left for later.

**As far as this review goes, this PR is OK to merge.**

**Remaining nit (non-blocking):** finding 1 above — spell out the `output_meta.num_tensors`-must-match-`output`-length invariant in the doc comment at some point, so a future second caller of `nnstreamer_filter_dispatch_output_async()` doesn't have to infer it from the llamacpp call site.

myungjoo:
Response to the relayed review above.

- **The nit (the cleanup loop relies on `prop->output_meta.num_tensors` matching the `output` array):** addressed by amending `[Filter] Release async output dispatched after tensor-filter stopped` (now 1604b90d; head cb3453b0). The `@param output` doc, in both `tensor_filter_common.c` and the public `nnstreamer_plugin_api_filter.h`, now says the array holds one entry for each tensor `prop->output_meta` describes. Both callbacks already walk the array by that count, so this states the existing contract rather than adding a new one. It is a doc-only change: the tree differs from 238399d3 in those two lines alone.
- **CI on 238399d3, with the gate on:** all 21 checks passed. The valgrind step reports `0 definite leak contexts from this repository` in both passes (22 binaries, plus llamacpp).

myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Scope of this pass

Follow-up to the fourth relayed review, which left one non-blocking nit open: the async-dispatch cleanup loop trusts `prop->output_meta.num_tensors` to match the length of the `output` array, and nothing in the doc comment said so. The author's response states this was addressed by amending the `[Filter] Release async output dispatched after tensor-filter stopped` commit (now `1604b90d`, new head `cb3453b0`), doc-only, and that CI on the prior head (`238399d3`) passed all 21 checks with the valgrind step reporting 0 definite leak contexts of ours in both passes.

This pass verifies (a) the amendment is what it claims to be and actually settles the nit, and (b) nothing else changed that would require re-doing the mechanism review already done twice.

## Verification performed

- `git diff 238399d3 cb3453b0`: exactly two hunks, both in comment blocks — the `@param output` doc for `nnstreamer_filter_dispatch_output_async()` in `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h` and the mirrored doc comment in `gst/nnstreamer/tensor_filter/tensor_filter_common.c`. No code line, no test, no other file touched. This matches the author's "doc-only, two lines" description exactly.
- The new text: *"The GstTensorMemory array holding the asynchronously generated output, one entry for each tensor prop->output_meta describes."* This does state the invariant the nit asked for (`output_meta.num_tensors` must match the `output` array's length) in the same place a future second caller of this function would read before implementing a sub-plugin's dispatch — the nit is settled, not just reworded. Nothing in the new sentence overstates or contradicts the actual cleanup code (`for (i = 0; i < prop->output_meta.num_tensors; i++) g_clear_pointer (&output[i].data, g_free);`), which still walks by `num_tensors` exactly as before.
- Reconfirmed the local worktree (`claude/leak-gate-phase1-5b8e21`) is checked out exactly at `cb3453b0` and that `git diff cb3453b0 <fetched PR head>` is empty, so this review is of the actual current head.
- Since the amendment touches only comment text with no code or behavioral change, it cannot have altered the outcome of the mechanism review already performed twice (leak attribution logic, the six leak fixes, the gate-flip exit condition) — not re-run here, per the earlier passes' findings, which stand.
- `tools/debugging/test_check_valgrind_log.sh` is untouched by this amendment (confirmed no path under `tools/debugging/` appears in the `238399d3`→`cb3453b0` diff), so the previously-verified 91/91 self-test result under gawk and mawk still applies unchanged.

## Assessment

No new issues. The amendment is exactly what the author's response describes — a two-line, comment-only change — and it does close the one remaining nit by documenting the array-length invariant at the point future callers would need it. Nothing else in the PR changed since the last full review pass, so that pass's conclusions on the leak-attribution mechanism, the six fixes, and the gate flip still hold.

**As far as this review goes, this PR (head `cb3453b0`) is OK to merge.**

**Remaining nits:** none from this pass. No open items carried over from prior passes.

myungjoo:
Response to the relayed review of `cb3453b0` (review 5174981906). The branch is rebased on main 7bfaa872; head is `00a022b9`.

**1. ggml CPU backends counted as ours: agreed, fixed.**
- `is_ours` now returns 0 for an object whose basename starts with `libggml`, so an error blamed on one is not ours either.
- The header says why: CI copies those prebuilt objects into `build/tests`.
- Three self-test cases use the real frame text from run 34559720254:
  - an error in `libggml-cpu-haswell.so`: library;
  - a leak whose stack holds only ggml frames: passes;
  - a ggml frame above a test frame: attributed to the test frame, as a GLib allocation would be.
- Dropping the exclusion fails 6 cases. Across 21 saved CI logs the error verdicts are unchanged.

**2. Commit structure: partly adopted.**
- `3392a5d3` (the self-test cases added in answer to review) and `b93d640f` (the header rewrite) are squashed into the phase-1 commit.
- The warn-only phase stays a separate commit ahead of the fixes, because separate phase-1 and phase-2 commits are what the author asked for. Every commit remains bisectable: phase 1 only warns, and each fix is green on its own.

**3. Other open PRs will fail: updated.**
- #4953 has since merged, so its two leaks are fixed here. Its flexible-input harnesses installed a private bus nothing drained, and the queued messages kept the harness alive.
- #4950 has merged too. Its leak turned out to be shipped code: `gst_tensor_if_chain` did not unref the buffer on the error path #4950 made reachable. That is now its own commit.
- The PR body now names #4957 and #4958.

**4. Coverage limit: done.** One sentence in the header says a block allocated wholly by library code and dropped by our code passes as a library leak.

**5. Overlap with #4961:** nothing to change.

**Also found while verifying the rebase locally.** unittest_filter_custom's `flexibleInvoke_p` and three sibling cases unregister their custom-easy model while join is still streaming. A slow memcheck run then reports two invalid reads in `custom_invoke`, which is an error the #4936 gate already fails on. Fixed in `[Test/Filter] Stop the custom-easy pipelines before unregistering the model`.

myungjoo:
> **Note:** This review was produced by a separate AI agent (not the account posting it) and is being relayed here verbatim on the requester's behalf. Treat it as one automated opinion among the human reviews this PR still needs.

## Scope of this pass

This PR has already been through five rounds of automated review in this thread, most recently against `cb3453b0`. Since then the branch was rebased onto main `7bfaa872` and restructured into 9 commits (head `00a022b9`): the phase-1 commit now folds in the ggml exclusion and its self-test cases directly, and two commits were added to fix leaks/UAFs that PRs #4950 and #4953 introduced into main after the last pass — `[Element/If] Release the buffer a condition cannot be checked on` and `[Test/Filter] Stop the custom-easy pipelines before unregistering the model`. This pass independently re-verifies the whole diff on this head rather than trusting the prior rounds' conclusions to still hold.

## Verification performed

- Read every commit's diff (`git log -p origin/main..00a022b9`) against the surrounding function bodies, not just the hunks.
- Ran `tools/debugging/test_check_valgrind_log.sh` myself: 99/99 pass under gawk 5.3.2, and 99/99 under real mawk 1.3.4 in WSL (copied the scripts out of `/mnt/c` first to avoid the CRLF trap) — matches the PR's own claim exactly.
- Traced `is_ours()` / `is_exempt_from_leaks()` / `record_leak()` by hand against the actual memcheck log format: attribution stops at the innermost stack frame that is either exempt (loader, `gst_debug_add/remove_log_function`, matched with an anchored `[. ]` boundary so a similarly-prefixed symbol of ours wouldn't false-exempt) or "ours" (a source file, or a `/build/(gst|ext|tests)/` object now excluding `libggml*`). This matches the header comment and the self-test cases (`leak_similar_name`, `leak_object_ours`/`leak_object_library_n`, `object_ggml_n`/`leak_ggml_n`/`leak_ggml_then_test`).
- Checked `gst_tensor_time_sync_buffer_from_collectpad`'s new `error:` label for a double-free, since it unconditionally calls `gst_tensors_config_free(&in_configs)` even on the path where `gst_tensors_config_from_caps()` already freed `in_configs` internally on failure: `gst_tensors_info_free()`/`gst_tensors_config_free()` (`nnstreamer_plugin_api_util_impl.c:376`,`:924`) re-init the struct after freeing, so a second free is a safe no-op. Also confirmed `counting` is only incremented after every `goto error` site, so the cleanup loop unrefs exactly the memories already gathered from earlier pads, no over-release.
- A prebuilt tree of this exact PR head (llamacpp + litert enabled) was available in the local WSL environment; verified it is content-identical to this head (`diff --strip-trailing-cr` on `check_valgrind_log.sh` and `gsttensor_if.c` showed no difference beyond CRLF). Ran the actual affected tests under memcheck and fed each log through the shipped checker with `--repo-root`:
  - `unittest_if` (`comparedValueBufferShorterThanCaps_n`, `customCallbackFailure_n`): 0 leak/error contexts of ours.
  - `unittest_sink` (`*issue739*Merge*`, `*issue739*Mux*`): 0 leak/error contexts of ours.
  - `unittest_filter_llamacpp` (`zeroTokenGenerationSync_p`, `zeroTokenGenerationAsync_p`, exercising `octet_stream`): 0 leak/error contexts of ours.
  - `unittest_filter_custom` (`tensorFilterFlexInput.*`, `tensorFilterCustom.*`): 0 leak/error contexts of ours.
  - `unittest_filter_litert` (`customProp00_n`): 0 leak/error contexts of ours.
  - `unittest_filter_single` (`*invokeAsync*`): 0 leak/error contexts of ours.

  This independently reproduces, directly under memcheck against the current head (not just by inspection), five of the six fixed leak sites plus the test-only use-after-free fix. `unittest_filter_armnn` is not present in this build, matching the PR body's own disclosure that armnn isn't built locally.
- Confirmed CI wiring is unchanged: `.github/workflows/ubuntu_clean_meson_build.yml` still runs `bash check_valgrind_log.sh ... || checked=1` for both `valgrind-tests.log` and `valgrind-llamacpp.log`, then `exit ${checked}` — so the new non-zero exit on a repo leak actually fails the job, not just prints a warning. `static.check.yml`'s self-test step is untouched and still gates on the script's own exit code.
- Confirmed scope: 15 files, all directly on-topic — the checker and its self-test, the six leak/UAF fixes, their regression tests, and one doc-comment pair (`nnstreamer_plugin_api_filter.h` / `tensor_filter_common.c`) kept in sync in the same commit. No unrelated module or workflow file touched.
- No architecture or public API change: the one doc-adjacent edit only clarifies an existing function's contract (`output` holds one entry per `output_meta.num_tensors`) identically in the header and the source doc comment; nothing in `Documentation/` references this script or that function, before or after this PR.
- Checked that both findings from the prior bot review on `cb3453b0` are actually settled on this head: the ggml exclusion (`is_ours()` special-cases `libggml*` for both errors and leaks, with three matching self-test cases) is present and passing; the commit-structure nit was partly addressed as the author's response states — the two follow-up commits are folded into the phase-1 commit, while phase 1 (warn) and phase 2 (fail) remain intentionally separate, bisectable commits.
- Checked the PR body's accuracy against the current head: the fix/owner table, the negative-control claims, and the "other open PRs will fail" note (`#4957`, `#4958` are still open and would fail this gate) all match what's on GitHub right now.

## Findings

No new blocking, medium, or high-severity issues found in this pass. Everything substantive raised in the five prior rounds (the leak-attribution fallthrough, the "35 vs 13" reproduction mismatch, the missing object-frame leak test, the ggml false positive, the `output_meta.num_tensors` doc gap) was already fixed or resolved by explanation, and remains correctly resolved on this head.

One informational note, not a defect: the commit message's claim that "each of 9 checker mutations... fails at least one case" describes a manual mutation-testing pass the author ran, not a script checked into the repository, so it isn't something a future CI run or reviewer can reproduce from the tree alone. Not a reason to hold up the PR — the 99/99 gawk+mawk self-test result itself is fully reproducible, and I did reproduce it.

## Assessment

The attribution mechanism is sound, the six fixes are each minimal and correct for the leak/UAF they target (verified independently under real memcheck for five of the six; the sixth, armnn, isn't built in this environment and was checked by inspection only, consistent with the PR's own disclosure), the gate flip is a two-line, well-tested change, and the CI wiring genuinely fails the build on a future repository leak rather than only warning. Scope stays proportional to the stated goal across all 9 commits, and the one doc-adjacent change ships in the same commit as the code it documents.

**As far as this review goes, this PR (head `00a022b9`) is OK to merge.**

**Remaining nits (non-blocking):**
- The mutation-testing claim in the gate-flip commit message isn't reproducible from the tree itself (see above) — cosmetic, not a reason to change anything.
- The PR body's note on the `#4961` exit-line overlap will need a small manual rebase, whichever of the two PRs lands second; already flagged by the author and not something this PR can resolve unilaterally.

myungjoo:
Response to the relayed review of `00a022b9`: both nits are informational, and neither needs a code change.

- **The mutation count is not reproducible from the tree.** Right. The mutation run was a throwaway local harness; it applies each of 9 edits to the checker and re-runs `test_check_valgrind_log.sh`. It is evidence for the review, not a gate. What gates is the self-test itself, which `static.check.yml` runs on every PR. Committing the harness would add maintenance without gating anything, so the count stays in the commit message as the record of what was tried.
- **#4961 overlap:** already covered in the PR body; the combined exit line is given there for whichever PR lands second.

PR #4915, opened 2026-09-02
[Build/ExecuTorch] Make the ExecuTorch subplugin buildable and test it in CI

myungjoo:
Addresses items 1 and 2 of the plan in #4289: give the ExecuTorch dependency a discoverable, reproducible build, and put the existing SSAT suite on CI. Reviving `executorch-llama` and Tizen packaging are out of scope.

## Why

The ExecuTorch subplugin and its SSAT suite have been in the tree since 2024, but nothing has ever built them. `meson` looks for `executorch.pc`; no environment we build in provides one, so `executorch_support_is_available` is false everywhere — `packaging/nnstreamer.spec` keeps `executorch_support` at 0 in every profile, `debian/` does not mention it, and no workflow does either. The suite is gated behind the same flag, so it is not even installed. Zero coverage.

ExecuTorch cannot simply be installed: no distro package, GitHub releases carry only CMSIS Pack files, and the pip wheel has headers and a cmake config but no runtime library. It has to be built from source.

## What

- **`tools/executorch-install.sh`** — one-step source build and install, so the recipe is reproducible rather than living in someone's shell history. It encodes two non-obvious constraints: ExecuTorch refuses to configure unless its source directory is named exactly `executorch` ([pytorch/executorch#6475](https://github.com/pytorch/executorch/issues/6475)), and `EXECUTORCH_BUILD_SHARED` is what produces `libexecutorch.so` plus the `executorch.pc` that meson then finds with no change on our side.

  The script rewrites the `Libs:` line of that generated `.pc`. Upstream links `libexecutorch.so` alone, which holds the runtime and extensions but **none of the operator kernels** — a subplugin built against the stock file links cleanly and then fails on the first invoke with an unregistered operator. The kernels are in `libexecutorch_portable_ops.so`, reached only through static registrars, hence `--no-as-needed`. This restores the contract the subplugin was written against, which 404190d3 spelled out but left to a hand-written `.pc` that was never committed.

- **`meson.build`** — `executorch-llama-support` probed the same `executorch.pc`, so making that discoverable would auto-enable it wherever the core runtime is present, then fail on a header `cmake --install` does not install. It now probes for the llama runner it actually needs.

- **`.github/workflows/executorch.yml`** — builds against a pinned `v1.4.1` and runs the existing SSAT suite. ExecuTorch is far too slow to build per-PR, so the install tree is cached and the job runs only on paths that can break it, plus daily — which doubles as what keeps the cache from being evicted for inactivity.

## Verification

Done locally:
[...]

PR #4956, opened 2026-09-10
[CI] Make the pylint check fail on pylint errors

### LINKED ISSUE #4919: [CI] The pylint check cannot fail: it counts pylint 1.x output patterns
## Summary

`.github/workflows/static.check.scripts/pylint.sh` decides whether a file failed by counting output lines:

```bash
pylint --reports=y $file > $result
line_count=$((`cat $result | grep W: | wc -l` + \
    `cat $result | grep C: | wc -l` + \
    `cat $result | grep E: | wc -l` + \
    `cat $result | grep R: | wc -l`))
```

Those patterns match nothing that any pylint the workflow can install actually prints. Modern pylint emits

```
tests/nnstreamer_converter/generateTest.py:1:0: C0114: Missing module docstring (missing-module-docstring)
```

There is no `C:` substring in `C0114:`. `line_count` is therefore always 0, `failed` is never set, and the step exits 0 no matter what pylint found. The check has been inert.

## Reproduction

`tests/nnstreamer_converter/generateTest.py` has four real findings (three convention, one warning) and is rated 6.00/10:

```
$ pylint --reports=y tests/nnstreamer_converter/generateTest.py
generateTest.py:1:0: C0114: Missing module docstring (missing-module-docstring)
generateTest.py:1:0: C0103: Module name "generateTest" doesn't conform to snake_case naming style (invalid-name)
generateTest.py:14:0: C0116: Missing function or method docstring (missing-function-docstring)
generateTest.py:14:36: W0622: Redefining built-in 'type' (redefined-builtin)
```

Counting the same output the way the script does:

| pattern | lines matched |
|---|---|
| `grep W:` | 0 |
| `grep C:` | 0 |
| `grep E:` | 0 |
| `grep R:` | 0 |

And the file passes the Static checks job today.

`static.check.yml` runs on `ubuntu-latest` and installs `pylint` from apt, so the version is always 2.x or newer; the patterns look like they were written for pylint 1.x's `C: <line>, <col>:` format, which the script inherited from TAOS-CI. I could not run a 1.x pylint on a current Python to confirm that directly, so treat the origin as likely rather than established — the present-tense behaviour above is what is verified.

## Why this is not a one-line fix

The script builds its own configuration with `pylint --generate-rcfile`, i.e. stock pylint defaults, which require module and function docstrings and snake_case module names. Repairing the detection alone would turn the check from "never fails" into "almost always fails":

**47 of the 49 `.py` files in the tree** (excluding `obsolete/` and `external/`, as the script does) produce at least one message under that generated rcfile. That includes long-standing files such as `generateTest.py` above.

So restoring the check needs a decision about what it should enforce, not just a corrected `grep`. Some options, in rough order of effort:

1. Commit a project `.pylintrc` enabling an agreed subset (for instance errors and warnings only, with `missing-docstring` and `invalid-name` off), and fail on any remaining message.
2. Fail only on `E`/`F` messages, which is what most CI setups start with.
3. Gate on the score rather than the message count, with a floor the tree currently meets.

Whichever is chosen, the counting should use pylint's exit status or its parseable output rather than grepping human-readable text — `pylint --output-format=json` or the documented exit-code bitmask are both stable.

## Secondary defect

The failure branch prints the wrong message, copied from `doxygen-tag.sh`:

```bash
echo "::error There is a doxygen tag missing or incorrect."
```

It is unreachable today, so nobody has noticed.

## Context

Found while adding `tests/nnstreamer_filter_executorch/generateModel.py` in #4915: the new file was cleaned up to 9.55/10 by hand, and the check would have passed it either way. Same family as #4908, where a per-function Doxygen check was silently disabled.

myungjoo:
Fixes #4919.

## Problem

`pylint.sh` counted output lines containing `W:`, `C:`, `E:` or `R:`. That is the pylint 1.x format; pylint 2.x+ prints ids such as `C0114:`, so the count was always 0 and the check passed every file. CI installs pylint 3.0.3 (noble). PR #4915's run shows `generateModel.py` in the checked list and the step passing silently.

## Change

- **`pylint.sh`**: gate on pylint's exit status with `--errors-only --disable=import-error --ignored-modules=gi.repository`, i.e. only `E`/`F` messages fail.
  - Stock defaults put convention/refactor/warning messages on every `.py` in the tree (50/50), so fixing the count alone would reject nearly every PR touching Python.
  - `import-error` is disabled because the job installs none of numpy, torch, tensorflow, cv2, nnstreamer_python (61 spurious `E0401` otherwise).
  - `gi.repository` is ignored because the runner has `python3-gi` (it is in noble's server cloud image manifest) but no GStreamer typelib, so the two `tests/nnstreamer_decoder_tensor_region/*.py` scripts get `E0611 No name 'Gst' in module 'gi.repository'`, which `import-error` does not cover. `E0611` on other modules is still reported, and the self-test pins that. Inside `gi.repository` it also silences `no-member` and typo'd names, but without the typelib pylint cannot resolve any of them on the runner anyway.
  - The `file` precondition accepts any text, not only ASCII text: one non-ASCII byte made `file` report UTF-8 text and pylint was never run.
  - The generated `~/.pylintrc` has no effect under `--errors-only` and is dropped; the failure message (copied from `doxygen-tag.sh`) now names pylint; `echo "\n\n"` → `printf`; each checked file is logged (`Checking <file>`), so a run that checked nothing is distinguishable from a pass.
- **`passThrough_CV.py`**: bare `raise` outside an `except` (`E0704`). It raised `RuntimeError: No active exception to reraise`; it now raises `RuntimeError("Failed to write image to ...")`. Same exception type, so the python3 filter sees the same failure.
- **`gen_dynamic_batch.py`**: inline `no-member` suppression on `get_concrete_function`, which `tf.function` adds at run time. pylint 3.0.3 (the CI version) reports it when tensorflow is not installed; 4.0.8 does not. Without this, the next PR editing that file would be blocked by a false positive.
- **`test_pylint.sh`** + a guarded `static.check.yml` step, following the existing self-tests: an error, a syntax error, an error in the second listed file, an error in a file with a non-ASCII comment, and a missing name imported from `os` must fail, with the log naming the file and the message id (so a missing pylint cannot pass for a finding); style-only messages, an uninstalled import, a `gi.repository` Gst import, a non-`.py` file and an empty list must pass.

## Verification

| check | result |
|---|---|
| `test_pylint.sh`, ubuntu:24.04 apt pylint 3.0.3, bare and with `python3-gi` + `gir1.2-glib-2.0` | 17/17 pass in both |
| `test_pylint.sh` against the **old** `pylint.sh` | all 7 failure-direction checks fail → self-test would have caught #4919 |
| `test_pylint.sh` against the previous head (50083077) of this PR, with `python3-gi` | the non-ASCII and the gi cases fail → both review findings are pinned |
| `pylint.sh` mutated to `--disable=no-name-in-module` instead of `--ignored-modules=gi.repository` | the `from os import no_such_name` case fails → the exemption cannot silently widen |
| new `pylint.sh` over all 50 tracked `.py` (excl. `obsolete/`, `external/`), ubuntu:24.04 bare **and** with `python3-gi` + `gir1.2-glib-2.0` | exit 0, 50 checked in both (previous head: the two tensor_region scripts fail with `E0611` when `python3-gi` is present) |
| `passThrough_CV.invoke` with stubbed `cv2`, old vs new | success path identical; failure path `RuntimeError` in both, message now says what failed |
| Static-checks scripts on the changed files in ubuntu:24.04 (doxygen-tag, newline, executable, hardcoded-path, misspelling, doxygen-build, sloccount, prohibited-words, shellcheck + existing self-tests) | all exit 0 |

This PR changes `passThrough_CV.py` and `gen_dynamic_batch.py`, so its own Static checks run exercises the new gate on both.

## Not in scope

`shellcheck.sh` has the same "cannot fail" shape (it pipes the file into `shellcheck` without `-`, so shellcheck prints "No files specified" and every file reports 0 lines). That is left for a separate issue.


myungjoo:

I reviewed this PR against the diff, the referenced issue (#4919), the surrounding CI scripts in `.github/workflows/static.check.scripts/`, and the repository's CI/test conventions. I also independently re-ran the changes locally (Windows/Git-Bash, pylint 4.0.7) to sanity-check the claims in the PR description: `bash .github/workflows/static.check.scripts/test_pylint.sh` → 7/7 PASS, and running the new `pylint.sh` over all 50 tracked `.py` files (excluding `obsolete/`, `external/`) → exit 0, matching the PR's verification table.

Yes. The root cause in #4919 — `pylint.sh` grepping for pylint 1.x-style `W:`/`C:`/`E:`/`R:` substrings that don't appear in modern pylint output, so `line_count` was always 0 — is fixed by gating on `pylint --errors-only --disable=import-error`'s own exit status instead of parsing text. I confirmed locally that:
- A file with only convention/refactor/warning findings, or an unresolved import, exits 0 (does not fail the check).
- A file with an `E`-class message (e.g. `E0704` misplaced-bare-raise) or a syntax error exits 1 (fails the check), including when it's the second file in a multi-file list.

This directly matches the exit-code approach the issue itself suggested ("the counting should use pylint's exit status ... rather than grepping human-readable text").

Low. The change is confined to the CI static-checker script, its new self-test, the workflow step list, and two pre-existing test-model generator scripts (`gen_dynamic_batch.py`, `passThrough_CV.py`) that needed fixing to pass the newly-strict gate (both are proven to have real `E`-level findings under the old, broken check — a latent problem this PR incidentally surfaces and fixes rather than causes).
- `passThrough_CV.py`: the bare `raise` outside an `except` (which actually raised Python's own `RuntimeError: No active exception to reraise!`, not a meaningful error) is replaced with `raise RuntimeError(f"Failed to write image to {filename}")`. Exception type is unchanged (`RuntimeError`), so `except Exception as e:` in the same function still catches it and prints `Warning: cv2 operation failed: {e}` — the diagnostic text is preserved, just attached to a more meaningful message. I checked `tests/nnstreamer_filter_python3/runTest.sh`, the only test that invokes this script; it does a golden-output comparison and does not assert on this error-path string, so no test regression.
- `gen_dynamic_batch.py`: adds an inline `# pylint: disable=no-member` for `get_concrete_function`, which `tf.function` only adds at runtime. This is a comment-only change with no behavioral effect on the generated model.

## Scope / size

Proportionate. +139/-13 across 5 files: the fix itself (`pylint.sh`, 5/-10), a new self-test (`test_pylint.sh`, new file, mirrors the existing `test_prohibited_words.sh`/`test_doxygen_tag.sh` pattern exactly — same header convention, same `expect_checker` helper shape, same merge-ref-guarded wiring in `static.check.yml`), and the two minimal test-model fixes needed for the PR's own files to pass the new gate. Nothing outside this footprint is touched.

## Test coverage

Good, and satisfies the repo's bug-fix test policy (regression test that fails without the fix, passes with it): `test_pylint.sh` pins both directions — an error, a syntax error, and an error as the second file in the list must fail; style-only messages, an uninstalled import, a non-`.py` file, and an empty list must pass. The PR description states this suite was also run against the *old* `pylint.sh` and the three failing cases passed there, i.e. the self-test would have caught #4919. That is exactly the kind of regression coverage CLAUDE.md's test-case policy asks for.

## Future-regression detection / CI gating

The new self-test step in `static.check.yml` (`/Checker/ pylint self-test`) runs `test_pylint.sh` unconditionally on every PR (not just PRs touching pylint.sh), using the same merge-ref-guard pattern as the sibling self-tests already in the file (skip only if the script is absent *and* this PR isn't the one deleting it, otherwise hard error). Since it invokes the real, currently-installed `pylint` binary rather than a mock, a future pylint version that changes how `--errors-only` behaves would be caught here, not just when someone happens to edit `pylint.sh`. `exit 1` on failure will fail the job, which is a real (non-continue-on-error) step — so a regression here blocks merge via normal required-check plumbing, consistent with the other self-tests in the same workflow.

## Documentation / architecture

Not applicable — no public API or architecture change; `Documentation/` has no mention of the pylint check to update.

## Findings

- **[Low]** `.github/workflows/static.check.scripts/pylint.sh` (new line ~56): `--disable=import-error` is a blanket disable, not scoped to the specific packages CI doesn't install (numpy/torch/tensorflow/cv2/nnstreamer_python). It's a reasonable, explicitly-disclosed tradeoff for today's tree (avoids ~61 spurious `E0401`), but it also means a genuine typo'd/missing import anywhere in the tree will no longer be caught by this gate. Worth tracking as a known gap for a follow-up (e.g. install the real dependencies in CI, or otherwise scope the suppression) rather than a blocker for this PR.
- **[Nit]** `tests/test_models/models/gen_dynamic_batch.py` line 47: I could not reproduce the `no-member` finding this `# pylint: disable=no-member` is meant to suppress using a local pylint 4.0.7 (no tensorflow installed at all) — it passed identically with and without the comment. This is very plausibly a pylint-3.0.3-vs-4.x astroid heuristic difference (the PR's own verification table covers 3.0.3/4.0.8/apt-24.04) and the comment is harmless if unneeded on newer pylint, so this is informational only, not a defect.

No High or Medium severity issues found.

**Verdict: OK to merge** (with the Low/Nit notes above as optional follow-up, not blockers).

myungjoo:
[Response to the automated review above, relayed on behalf of the author by a Claude agent]

Thanks. Taking the two notes in turn; pushed as an amend (e172fc68).

**[Low] `--disable=import-error` is blanket, not scoped.** Kept as is, and I would rather not track it as a follow-up either. Scoping is possible (`--ignored-modules=numpy,...` does suppress `E0401` for the listed modules only, checked with pylint 3.0.3), but the list the tree needs today is 14 top-level names: numpy, test_utils, nnstreamer_python, torch, onnx, gi, pybadges, gen24bBMP, flatbuffers, tensorflow, gtk, executorch, cv2, bs4. Two of them (`test_utils`, `gen24bBMP`) are in-tree helpers imported through `sys.path`, so every new test script that imports a new dependency or a sibling helper would fail Static checks until someone edits `pylint.sh`. What the scoped list would catch in exchange, an import that does not resolve, already fails the first time the script runs. That trade is not worth it for this gate.

**[Nit] could not reproduce `no-member` in `gen_dynamic_batch.py`.** Version-dependent, as you guessed. With a clean `$HOME` and no tensorflow: pylint 3.0.3 (what `ubuntu-latest` installs from apt; also confirmed in an `ubuntu:24.04` container) reports `E1101 ... has no 'get_concrete_function' member` on line 47 and exits 2; pylint 4.0.8 reports nothing. So the suppression is needed for the CI version. The commit message and PR body said "pylint cannot see it without tensorflow installed", which is wrong for 4.x; both now name the version.

One addition in the same amend: `pylint.sh` now prints `Checking <file>` for each Python file it lints. The first CI run of this PR passed the step with an empty group, which is exactly what the inert check looked like before; now the log shows which files were gated.

myungjoo:

This is a re-review after the earlier automated review and the author's response (amend to e172fc68). I re-read the diff at e172fc68, the earlier review, the response, issue #4919, `pylint.sh`, `static.check.yml`, and the actual "Static checks" CI run for this commit (run 34466752634, job 102837008200 — passed, 43s), and I pulled the raw step logs for the Pylint and pylint-self-test steps to confirm behavior rather than take the PR description's word for it.

## On the previous round

Both notes from the earlier review were substantively addressed, not just replied to:

- **`--disable=import-error` scope**: the author's response gives a concrete, checked cost (14 top-level names, two of them in-tree helpers reached via `sys.path`) for scoping it, and a concrete benefit forgone (an unresolved import already fails the first time the script runs, independent of this gate). That's a reasoned trade-off, not a deflection. I agree with keeping it unscoped and with not tracking it as a follow-up — scoping would make every new test script that imports a new dependency fail Static checks until someone edits an allowlist, for a defect class this gate does not need to own.
- **`gen_dynamic_batch.py` no-member nit**: resolved with an actual version-matrix finding (pylint 3.0.3 from apt — what `ubuntu-latest`/noble installs — reports `E1101` on `get_concrete_function` without tensorflow installed; 4.0.8 doesn't), and the commit message was corrected to say so instead of "pylint cannot see it without tensorflow installed." Verified independently below.
- The added `echo "Checking $file"` line is a reasonable, low-risk addition; the real CI log confirms it fires exactly for the two `.py` files in the diff and not for `pylint.sh`/`test_pylint.sh` themselves (correctly excluded by the `*.py)` case arm).

## Does it solve #4919?

Yes. `pylint --errors-only --disable=import-error "$file"`'s own exit status now gates `failed`, replacing the dead `grep W:/C:/E:/R:` line-count. I confirmed this against the real CI log for this exact commit: pylint 3.0.3 (noble/universe) was installed, the two changed `.py` files were linted, and the new `/Checker/ pylint self-test` step ran `test_pylint.sh`, which printed 7/7 `PASS` lines against the live binary — not a mock. The secondary defect from the issue (wrong `::error` text copied from `doxygen-tag.sh`) is also fixed.

Low, and I did not find anything beyond what the previous round already surfaced.

- `passThrough_CV.py`: replacing the bare `raise` (which, misplaced outside an `except` clause, raised Python's own `RuntimeError: No active exception to reraise!`) with `raise RuntimeError(f"Failed to write image to {filename}")` preserves the exception type (`RuntimeError`) that reaches the outer `except Exception as e:` and the caller, in both the old and new code — only the message text changes. `cv2.imwrite` failing is not something the local test model generation exercises, so there's no realistic path where this changes observed CI behavior.
- `gen_dynamic_batch.py`: the `# pylint: disable=no-member` is on the same physical line as `get_concrete_function()`, which is where pylint's line-scoped suppression needs it. Comment-only, no behavioral effect on the generated model.
- Removing `pylint --generate-rcfile > ~/.pylintrc` is safe: the repo has no `.pylintrc`/`pylintrc` of its own, and `--errors-only` overrides whatever config would otherwise apply, so the generated rcfile was dead weight once the gate stopped depending on convention/refactor/warning counts.

## Scope / size

Proportionate: `pylint.sh` (net -5 lines), a new self-test that mirrors the existing `test_prohibited_words.sh`/`test_doxygen_tag.sh` shape exactly (same header, same `expect_checker` helper, same stdin-closed invocation style), the matching `static.check.yml` wiring (byte-for-byte the same merge-ref-guard block already used for four sibling self-tests in the same file), and the two test-model fixes the new gate actually requires. Nothing outside that footprint is touched. No documentation exists for the pylint check (`grep -rli pylint Documentation/` is empty) and no API/architecture surface changed, so no doc update is owed.

## Test coverage and future-regression detection

`test_pylint.sh` covers both directions pinned against the real, installed pylint: an `E0704` (misplaced-bare-raise) and a syntax error each fail the checker (including when the erroring file is second in the list, which exercises the loop's failure accumulation rather than just a single-file short-circuit), while convention/refactor/warning-only content, an unresolved import, a non-`.py` file, and an empty file list all pass. Because the self-test invokes the live `pylint` binary rather than asserting against a captured string, a future pylint upgrade that changes what `--errors-only` reports would be caught here on every PR (the step is unconditional, not gated on touching `pylint.sh`), which is exactly the "detect when another module's change affects this checker" property the review criteria ask for. A real failure here is a normal non-continue-on-error step, so it blocks merge through the same required-check plumbing as everything else in `static.check.yml`.

One genuine gap, minor: of the three exit-1 cases, only `bare_raise.py`'s is followed by a content check (`grep -q "E0704" "$output"`) confirming *why* it failed. `syntax_error.py` and the "error in the second file" case only assert the exit code, and the harness's own `[[ $actual -ne 0 ]] && actual=1` normalizes any non-zero exit (including, hypothetically, a pylint usage/config error, exit 32) to "expected failure." So if a future pylint version turned the syntax-error input into a usage error instead of a detected error, those two cases would still show green without actually exercising the intended detection path. This doesn't affect the current PR's correctness — the real CI log for this commit shows all three failing for the right reason — but it's a small hole in an otherwise well-built self-test.

## Findings

- **[Low]** `.github/workflows/static.check.scripts/test_pylint.sh` — the `syntax_error.py` and "second file in the list" cases (`expect_checker 1 ...` around lines 156-158) assert exit status only, not message content, unlike the `bare_raise.py` case a few lines above it. A pylint change that turned either input into a usage/config error rather than the intended E/F detection would still pass this self-test. Suggest a follow-up (not a blocker) adding a `grep` for a stable marker (e.g. `E0001`/`invalid-syntax`, and the specific filename for the second-file case) the way the bare-raise case already does.
- **[Nit]** Same file, line 117: the fixture's docstring "Mirror the passThrough_CV.py defect found with #4919" slightly overstates it — issue #4919 is about the checker being inert, not about this bare-raise defect; the defect was found as a side effect of restoring the checker in this PR, not reported in #4919 itself. Purely a wording nit, no functional effect.

No High or Medium severity issues found. Both items above are optional polish, not blockers, and the second is pre-existing wording introduced in this same PR rather than something the previous round missed.

Verdict: OK to merge

myungjoo:
[Response to the automated re-review above, relayed on behalf of the author by a Claude agent]

Both addressed in-PR (amend 50083077), rather than as a follow-up.

**[Low] exit-status-only assertions.** Agreed, and it is a real hole rather than a theoretical one: while verifying the previous round, a run with pylint off `PATH` made all three failing cases "pass", because `pylint.sh` exits 1 on "The pylint utility is not found". Only the E0704 grep caught it. `test_pylint.sh` now has an `expect_log` helper, and every failure-direction case checks the log:
- bare raise: `bare_raise.py:<l>:<c>: E0704:`
- syntax error: `syntax_error.py:<l>:<c>: E0001:`
- second file of the list: `Checking .../style_only.py` (the first file was linted, not skipped) and `pylint error from .../bare_raise.py` (the failure is attributed to the right file)

Verified: 11/11 pass with pylint 3.0.3, 4.0.8, and in `ubuntu:24.04` with apt pylint. Against the old `pylint.sh`, all 7 failure-direction checks fail. With pylint removed from `PATH`, every log check fails (as do the pass-direction cases), so the self-test can no longer be green without a working pylint.

**[Nit] fixture docstring.** Now "Mirror the bare raise that passThrough_CV.py had."

Commit message and PR body updated to match.

myungjoo:

This is a third-round review after two earlier automated reviews and the author's two responses/amends (final state: commit 5008307775611b3259e8bb34d3fb82772d7da40d). I read the full comment history, issue #4919, the current diff, and the full text of `pylint.sh`, `test_pylint.sh`, `static.check.yml`, and the sibling self-tests (`test_doxygen_tag.sh` and others) at this commit. I did not just take the PR's word for the claims in the latest response — I reproduced them independently in a local shell (pylint 4.0.7, via a `pylint` shim on `PATH` since Git-Bash on this box can't exec the Windows pylint.exe wrapper directly; CRLF stripped from the committed CRLF working-tree copies into a scratch copy before executing, per this repo's Windows checkout quirk).

## Does the latest response adequately address round 2's findings?

Yes, on both points, and I did not find anything the round-2 review missed.

- **Exit-status-only assertions (round 2's Low finding).** The author's claim that this was a real, not just theoretical, hole is verified: I ran `test_pylint.sh` with `pylint` removed from `PATH` entirely. Before this amend's `expect_log` additions, `expect_checker` alone would have reported PASS for all three "must fail" cases, because `pylint.sh` exits 1 on "The pylint utility is not found" — indistinguishable by exit code alone from a real E/F finding. With the current `expect_log` checks in place, the same run now correctly fails 8 of 11 checks (the three failure-direction log checks, plus the three pass-direction `expect_checker 0` cases, which also go to exit 1 without a working pylint). This closes the hole exactly as described.
- **Fixture docstring wording (round 2's Nit).** Now reads "Mirror the bare raise that passThrough_CV.py had." — accurate, no longer implies the bare-raise defect itself was reported in #4919.

I independently ran the full `test_pylint.sh` against a real, installed pylint (4.0.7) and got 11/11 PASS, confirming the PR's own verification claim rather than assuming it.

## Fresh findings on the current diff (5 files, +152/-13)

I did not find any new High or Medium issues. Two minor observations, both already effectively covered by the discussion so far and not blockers:

- **[Nit]** `.github/workflows/static.check.scripts/pylint.sh` line 60: `echo "pylint error from $file" >> $errlog` is unquoted (`$file` without quotes), same as the pre-existing `which pylint`/`-z $1` style elsewhere in this script. Harmless for the filenames this repo actually produces, and not introduced by this PR's logic change — flagging only because the same line was touched by this diff (moved from `line_count` block) and it would have been free to quote while already there. Not worth a follow-up on its own.
- **[Informational]** The "Static checks" job for the current head commit was still `pending` at review time (workflow run 34467386272), so I could not cite a fresh green CI log the way the round-2 review could for its commit. This isn't a code defect — my own local reproduction with a real pylint binary (11/11 pass, and the pylint-off-`PATH` regression check above) covers the same ground the CI run would confirm. Recommend confirming the "Static checks" job goes green on this exact commit before merge, as normal process, not as a review blocker.

## Re-confirmed (not re-litigated) from earlier rounds

- Solves #4919: gate now uses `pylint --errors-only --disable=import-error`'s exit status (verified locally: E0704 bare-raise and E0001 syntax-error both fail the gate; C/R/W-only files, an unresolved import, and a non-`.py` file all pass; the wrong `::error` text copied from `doxygen-tag.sh` is corrected).
- Regression risk: low. `passThrough_CV.py`'s exception type is unchanged (still `RuntimeError`, still caught by the same `except Exception as e:`), and the only test that runs it (`tests/nnstreamer_filter_python3/runTest.sh`) does a golden-output comparison that doesn't exercise this error path. `gen_dynamic_batch.py` is not invoked by any test script in-tree (confirmed via repo-wide grep) — it's a standalone model generator, so the `# pylint: disable=no-member` comment has no runtime effect on any test.
- Scope: proportionate to the fix. The new self-test step in `static.check.yml` is byte-for-byte identical in structure (merge-ref-guard, comment wording pattern) to the four other self-test steps already in the same file (doxygen-tag, prohibited-words, debian-control, etc.), confirmed by direct comparison.
- Documentation: `grep -rli pylint Documentation/` is still empty; no architecture/API surface changed, so no doc update is owed.
- Future-regression detection and CI gating: the self-test runs the real installed pylint unconditionally on every PR (not gated on touching `pylint.sh`), asserts on log content (not just exit code) after this amend, and a failure is a normal non-continue-on-error step that blocks merge through the existing required-check plumbing — this satisfies the "detect and block on a future pylint-version change" criterion.

## Verdict

Verdict: OK to merge (informational-only note about confirming green CI on this exact commit; no blocking issues found).

myungjoo:
[Response to the round-3 automated review, relayed on behalf of the author by a Claude agent]

Thanks. On the one nit: `echo "pylint error from $file" >> $errlog` already has `$file` inside double quotes; the unquoted expansions in that block are `$result` and `$errlog`, both `mktemp` paths on lines this PR does not otherwise touch. I left them in the surrounding script's style rather than widen the diff. No change.

myungjoo:
[Response to the myungjoo-bot review, relayed on behalf of the author by a Claude agent]

Thanks, both findings reproduced and are fixed in the amended commit 446e40ba.

**1. [Medium] gi false positive on the runner.** Reproduced: in `ubuntu:24.04` with `pylint`, `python3-gi` and `gir1.2-glib-2.0`, the previous head fails the whole-tree run with exactly the two `E0611 No name 'Gst' in module 'gi.repository'` lines you quoted. I also checked noble's server cloud image manifest myself: it lists `python3-gi 3.48.2-1` and `gir1.2-glib-2.0`, and no `gir1.2-gstreamer-1.0`. The gap was in my earlier measurement, which only used a bare container.

I fixed it in `pylint.sh` with `--ignored-modules=gi.repository`, not with the two inline suppressions. The cause is the same kind as `import-error`: CI does not provide what the script imports. Handling it next to that flag also covers the next test script that imports `Gst` through gi, instead of leaving a per-file suppression for each author to rediscover. It is scoped: `E0611` on other modules is still reported (`from os import no_such_name` fails in the same run). The two tensor_region scripts are unchanged. `test_pylint.sh` gains a gi case that mirrors their import pattern and must pass. In the `python3-gi` container, that case fails against the previous head, so on the runner the self-test now pins this.

Whole tree after the fix: exit 0 with all 50 files checked, both in the bare container and with `python3-gi` + `gir1.2-glib-2.0`. The PR description row now names both environments.

**2. [Low] non-ASCII `.py` skipped.** Fixed in this PR rather than as a follow-up. The precondition now matches `text` instead of `ASCII text`. `test_pylint.sh` adds a file whose only non-ASCII byte is in a comment, next to an undefined name. It must fail with `E0602` in the log, and the test first asserts that `file` does not classify the fixture as ASCII. Against the previous head, both checks for that case fail.

Also checked: `shellcheck -S error` passes on both changed scripts, in case #4966 lands first.

myungjoo:

This is a fourth-round review, focused on whether the `myungjoo-bot` round-3 findings (CHANGES_REQUESTED on `50083077`) are actually resolved in the amended commit `446e40ba`, and on whether the author's chosen fix (`--ignored-modules=gi.repository`) is sound rather than a different way to hide errors. I re-read the full comment history, issue #4919, the current diff, `pylint.sh`, `test_pylint.sh`, and `static.check.yml` at `446e40ba`, and independently verified claims rather than trusting the PR description:

- Ran `pylint --version` locally (pylint 4.0.7) and executed `test_pylint.sh` directly against the working tree at `446e40ba` (Git Bash handled the script's CRLF fine, no stripping needed) → 11/11 checks PASS.
- Pulled the real "Static checks" CI log for commit `446e40ba` (run `34919556729` / job `104224379263`, passed in 43s) and confirmed the actual runner output, not just the PR's claim.
- Built a synthetic `gi/repository` package locally to test the exact semantics of `--ignored-modules=gi.repository` against real pylint, since this flag's scope was not empirically checked by any earlier round.

## Are the two round-3 findings resolved?

**1. [Was Medium] `gi.repository` E0611 false positive on the CI runner — RESOLVED.**
`pylint.sh:57` now runs `pylint --errors-only --disable=import-error --ignored-modules=gi.repository "$file"`. This is verified working on the actual runner, not just claimed: the CI log for `446e40ba` shows the self-test's `gi_gst.py` case running against `/usr/bin/pylint` (the real, apt-installed binary in the actual `python3-gi`-without-typelib environment) and passing:
```
PASS: gi.repository import without the GStreamer typelib (exit 0)
PASS: the gi file is checked
```
`test_pylint.sh:104-137` adds this case (`gi_gst.py`, `expect_checker 0 ... "gi.repository import without the GStreamer typelib"`, plus an `expect_log "Checking .*/gi_gst\.py"`), so a future regression here would be caught on every PR, not only when someone touches `pylint.sh`.

**2. [Was Low] non-ASCII `.py` files skipping the gate — RESOLVED.**
`pylint.sh:52` changed the precondition from `file $file | grep "ASCII text"` to `file $file | grep "text"`. Verified locally: `file` classifies a UTF-8 Python file as `Unicode text, UTF-8 text` (Windows/MSYS `file`, and per the earlier round's container testing, Linux `file` reports similarly) — the substring `text` matches both `ASCII text` and `Unicode text, UTF-8 text`, so the fix is correct and not over- or under-matching. `test_pylint.sh` pins this with `utf8_error.py` (an `E0602` case), and the CI log for `446e40ba` confirms it fires on the real runner too.

Both findings are genuinely fixed, not just replied to, and both are now covered by a permanent, unconditional self-test step rather than one-off manual verification.

## Is `--ignored-modules=gi.repository` a sound choice? (checked beyond its stated intent)

Partially — it works, but it silences more than it needs to, and the author's justification undersells that.

**[Low]** `.github/workflows/static.check.scripts/pylint.sh:57`

The author's response describes the flag as scoped ("E0611 on other modules is still reported"), which is true but incomplete. I built a synthetic `gi/repository` package (a `Gst` class with one real method and no `ThisMemberDoesNotExist` attribute) and ran real pylint 4.0.7 with and without the flag:

```
--- without --ignored-modules ---
case_bad_symbol.py:2:0: E0611: No name 'ThisDoesNotExistAtAll' in module 'gi.repository' (no-name-in-module)
case_bad_member.py:4:6: E1101: Class 'Gst' has no 'ThisMemberDoesNotExist' member (no-member)
case_unrelated_import.py:2:0: E0611: No name 'this_symbol_does_not_exist_in_os' in module 'os' (no-name-in-module)

--- with --ignored-modules=gi.repository ---
case_unrelated_import.py:2:0: E0611: No name 'this_symbol_does_not_exist_in_os' in module 'os' (no-name-in-module)
```

So `--ignored-modules=gi.repository` doesn't just fix the one known false positive (E0611 on a missing typelib symbol). It also suppresses:
- E0611 for a genuinely nonexistent name imported from `gi.repository` (e.g. a typo'd `from gi.repository import Gstt`), and
- E1101 `no-member` for *any* attribute access on a `gi.repository`-derived object (e.g. `Gst.SomeTypoedMethod()`), unconditionally, regardless of whether the typelib is actually available.

That's a materially wider gap than the two-line inline-suppression alternative the round-3 review proposed, which would only have silenced the specific known-bad import line and left member-access typos detectable. It's a defensible trade-off given GI bindings are inherently hard for static analysis and today's exposure is exactly two straightforward files (`tests/nnstreamer_decoder_tensor_region/generateResults.py`, `getCropInfo.py`, confirmed via `grep -rl "from gi.repository\|import gi" -- '*.py'`), but it is broader than "scoped" suggests, and it applies to every future file that imports anything from `gi.repository`, not just these two. Worth a one-line comment update (the existing comment on line 56 already explains the import-error rationale; extending it to say member-access checks are disabled too would prevent a future author from assuming `no-member` still works there) — not a blocker.

**[Nit]** `.github/workflows/static.check.scripts/test_pylint.sh`

The self-test proves the flag catches the *narrow* thing it's meant to catch (`gi_gst.py` passes) but doesn't pin the *boundary* — nothing in the suite asserts that `--ignored-modules=gi.repository` still fails on, say, an unrelated `os` import error, or documents (via a failing-then-passing case) that `no-member` on a `gi.repository` object is now silently accepted. The "still scoped" claim in the PR thread was verified by hand this round and in the original `myungjoo-bot` review's container run, not by a standing test. A short addition would let this exact scope survive an unrelated future edit to `pylint.sh` without a human re-deriving it by hand each time.

## Everything else

I did not find anything beyond what the last three rounds already covered and re-verified: the fix solves #4919 (exit-status gating replaces the dead `grep`), regression risk from `passThrough_CV.py`/`gen_dynamic_batch.py` remains low and unchanged from prior rounds, scope/size is proportionate (`pylint.sh` unchanged net size, the new self-test mirrors existing sibling self-tests byte-for-byte in structure), no documentation exists for this checker to update, and the self-test step is unconditional and non-continue-on-error, so it blocks merge through normal required-check plumbing. The "Static checks" job for the final commit `446e40ba` is green (confirmed directly from the run log, not inferred).

## Findings summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | Resolved (was Medium) | `pylint.sh:57` | `gi.repository` E0611 false positive on the CI runner — fixed and verified against the real runner in the CI log for `446e40ba`. |
| 2 | Resolved (was Low) | `pylint.sh:52` | Non-ASCII `.py` files skipping the gate — fixed and verified. |
| 3 | Low | `pylint.sh:57` | `--ignored-modules=gi.repository` is broader than its stated intent: it also disables `no-member` (E1101) checks and import-typo (E0611) detection for any `gi.repository` symbol, not just the specific "missing typelib" case. Defensible trade-off given current 2-file exposure, but worth documenting explicitly rather than leaving the "scoped" framing to imply only import resolution is affected. |
| 4 | Nit | `test_pylint.sh` | No self-test case pins that the `--ignored-modules` scope stays narrow (e.g. still catches an unrelated import error, or documents that `no-member` on `gi.repository` is now accepted). Currently relies on manual verification, not a standing regression test. |

No High or Medium severity issues found in the current diff. Findings 3 and 4 are optional follow-up, not blockers.

Verdict: OK to merge

myungjoo:
[Response to the round-4 automated review, relayed on behalf of the author by a Claude agent]

Thanks. Both notes are addressed in-PR (amend 1d277d2c).

**[Low] `--ignored-modules=gi.repository` is broader than "scoped" implied.** Agreed on the facts. Inside `gi.repository` the flag also silences `no-member` and a typo'd imported name. Where that matters is an environment that has the typelibs, like your synthetic package. On the CI runner, pylint cannot resolve any `gi.repository` name without the typelib, so the real `Gst.init` and a typo'd `Gst.Iniit` look exactly alike. The two inline suppressions would not have detected more there either: once the import line's `E0611` is suppressed, `Gst` is uninferable and no `no-member` fires. The runner loses nothing it could have checked. The trade-off is now written down in three places:
- the `pylint.sh` comment: "without typelibs, gi.repository names cannot be checked at all"
- the commit message
- the PR description

**[Nit] the exemption's boundary is not pinned.** Added `os_missing_name.py` (`from os import no_such_name`), which must fail, with `os_missing_name.py:<l>:<c>: E0611:` in the log. Mutation check in `ubuntu:24.04` with `python3-gi`: replacing `--ignored-modules=gi.repository` with `--disable=no-name-in-module` in `pylint.sh` makes both new checks FAIL, so the exemption cannot silently widen. I did not add a case asserting that `no-member` on `gi.repository` is accepted; that would pin a limitation rather than a guarantee.

Re-verified in `ubuntu:24.04` with apt pylint 3.0.3, bare and with `python3-gi` + `gir1.2-glib-2.0`: 17/17 self-test checks pass, and the whole tree is exit 0 with 50 files checked. `shellcheck -S error` and the other static checkers pass on the changed scripts.

myungjoo:
[Automated re-review (round 5) by a separate Claude agent, relayed to this PR]

This is a fifth-round review, focused on the amend from `446e40ba` (reviewed in round 4) to the current head `1d277d2c`, and on the author's argument for why the `--ignored-modules=gi.repository` scope is acceptable. I read the full comment history, issue #4919, `pylint.sh`, `test_pylint.sh`, `static.check.yml`, the PR description, and the commit message at `1d277d2c`. I did not take the PR's claims on faith: I diffed `446e40ba..1d277d2c` directly (`git diff 446e40ba 1d277d2c`, both objects still reachable), and pulled the real "Static checks" CI log for `1d277d2c` (run `34920982174`, job `104228847841`, passed in 1m7s) to see the actual self-test output on the real runner rather than trust the PR's verification table.

## Are round 4's Low and Nit resolved?

**Yes, both, and the fix for each is minimal and on-target.** The full diff between the two rounds is four lines in `pylint.sh` (comment wording) and eight lines added to `test_pylint.sh` (one new fixture, two new assertions) — nothing else changed:

- **[Was Low] `--ignored-modules=gi.repository` silences more than "scoped" implied.** `pylint.sh:56` comment changed from "CI lacks the imported modules, and its python3-gi has no GStreamer typelib." to "CI lacks the imported modules; without typelibs, gi.repository names cannot be checked at all." This states the actual scope (nothing under `gi.repository` is checkable without a typelib, not just the one import line) rather than implying only the reported import is affected. Combined with the PR description and commit message, which both now spell out that `no-member`/`no-name-in-module` are silenced for the whole module, this is adequately documented. Resolved.
- **[Was Nit] the exemption's boundary is not pinned by a test.** `test_pylint.sh` gained `os_missing_name.py` (`from os import no_such_name`), asserted to fail with `E0611` in the log ("only gi.repository is exempt from E0611"). I confirmed this ran for real on the actual CI runner for `1d277d2c`, not just claimed: `PASS: missing name imported from a module other than gi.repository (exit 1)` / `PASS: only gi.repository is exempt from E0611`. This pins that `--ignored-modules=gi.repository` doesn't leak into a blanket `E0611` suppression. Resolved. I agree with the author's stated decision not to add a case asserting that `no-member` *is* accepted on a `gi.repository` object — a test that pins a known limitation as a "guarantee" would make it harder to tighten later, and the existing boundary test already covers the property that actually matters (the exemption doesn't silently widen to other modules).

## Do I accept "the runner loses nothing it could have checked"?

**Yes, as an argument about the two suppression approaches being equivalent when the typelib is absent.** Once `from gi.repository import Gst` itself can't resolve (no typelib), `Gst` is `Uninferable` to pylint's astroid brain, so neither the two-line inline-suppression alternative (round 3's suggestion) nor `--ignored-modules` would have caught a `Gst.Iniit()`-style typo either — both approaches are blind to the same class of error on a typelib-less runner. That reasoning holds regardless of the exact runner details.

One thing I could **not** independently confirm: whether the actual "Static checks" job (`ubuntu-24.04` GitHub-hosted runner, confirmed from the job's "Set up job" log) has `python3-gi` preinstalled without the GStreamer typelib, as opposed to not having `gi` at all. The job's own `apt-get install` step does not install `python3-gi`, and the official `ubuntu24/20260907.300` runner-images readme I fetched doesn't mention `python3-gi`, `gir1.2`, or PyGObject anywhere (that readme is a curated list, not a full package dump, so this isn't conclusive either way). If `gi` were entirely absent on that runner, `import gi` and `from gi.repository import Gst` would both fail with `E0401 import-error`, which `--disable=import-error` alone already suppresses — making `--ignored-modules=gi.repository` untested-in-practice on that specific job, though still correctly justified for any environment (e.g. the `ubuntu:24.04` + `python3-gi` + `gir1.2-glib-2.0` container used in rounds 3-5) where `gi` is present but the typelib isn't. I don't have shell access to the actual runner to settle this, so I'm flagging it as an open question rather than a defect — it does not change the correctness of `pylint.sh` either way, since the flag is harmless if the scenario it targets doesn't occur on this particular job.

## Current diff as a whole (since `446e40ba`)

Proportionate and precisely targeted: one comment-wording line and one new self-test case, nothing else. No new files touched, no scope creep. The single commit (`1d277d2c`) carries a DCO sign-off and an accurate, detailed message that matches the diff (including the corrected "as on the runner" framing for the gi case). `Static checks` is green on this exact commit, including the full 17/17 self-test suite running against the real, apt-installed pylint 3.0.3 — I read the raw step log rather than the summary badge.

Everything re-confirmed from earlier rounds (solves #4919 via exit-status gating; `passThrough_CV.py`/`gen_dynamic_batch.py` changes are behavior-preserving and test-covered; no documentation exists for this checker so none is owed; the self-test is unconditional, non-continue-on-error, and blocks merge through normal required-check plumbing) still holds and I found nothing new to add there.

## Findings

No High, Medium, or Low findings on the current diff. One informational note only:

- **[Info, non-blocking]** Whether `python3-gi` without the GStreamer typelib is actually present on the GitHub-hosted `ubuntu-24.04` runner used by the "Static checks" job (as opposed to the deliberately-constructed local container the PR verifies against) is not something I could confirm from the job log or the runner-images manifest. This doesn't affect correctness — `--ignored-modules=gi.repository` is either necessary there or a no-op, not harmful either way — but if a maintainer wants to fully close this out, a one-off `python3 -c "import gi; print(gi.__file__)"` on the runner (or reading its dpkg list) would settle it.

## Verdict

Verdict: OK to merge

PR #4957, opened 2026-09-10
[Element/Filter] Refuse a flexible input that is not the model's input

### LINKED ISSUE #4928: tensor_filter: a flexible sink pad lets a mismatched element type reach the model unchecked

myungjoo:
Fixes #4928

Rebased on #4953 (merged), following the agreed order. #4953 already validates every flexible input header, refuses sparse input in both modes, keeps `prop->input_meta` as configured without `invoke-dynamic`, and checks each flexible tensor's size against it. This PR now adds only what #4928 still needs on top of that.

## Problem left after #4953

Without `invoke-dynamic`, a flexible tensor with the **same byte count** as the configured input tensor passes the size check whatever its type or shape. A float32 model reads int32 bits, or a `4:2` input reads a `2:4` tensor, and invoke returns success on reinterpreted data. This is the LiteRT repro in the issue.

## Fix

- New `_gst_tensor_filter_check_flexible_input()`. It runs after the size check, only for a flexible input without `invoke-dynamic`. It compares the type and dimension of the already validated header (trailing 1s ignored, `gst_tensor_info_is_equal()`) with the configured input tensor that the buffer tensor feeds, and refuses a mismatch with a `STREAM/WRONG_TYPE` element error. The message names both tensors (`int32 (4:2:1:1)` vs `'in0' float32 (4:2:1:1)`) and no longer prints the per-tensor header format.
- Everything from #4953 is kept as merged. That covers header validation, sparse refusal in both modes, name-preserving refresh with `invoke-dynamic`, the size check and its message, B9 and B7. A tensor of another size is still refused by the size check with the same error reporting as before, so `tensorFilterFlexInput.smallerThanModel_n` is unchanged.
- With `invoke-dynamic` the configured type and dimension follow each header, so there is nothing to compare. The subplugin decides.
- The comparison maps buffer tensors through `input-combination`, although flexible caps and `input-combination` do not negotiate today (flexible caps carry no tensor count).
- `Documentation/data-type-and-flow-control.md` states the combined per-buffer rules:
  - invalid, mismatched-size or sparse headers stop the stream in both modes;
  - without `invoke-dynamic` the type and dimension must match the configured input (model or `input`/`inputtype`);
  - only `invoke-dynamic=true` lets type or shape change.

  It also notes that `tensor_converter` stamps `application/octet-stream` as a `uint8` tensor of the buffer size, and that `input-dim`/`input-type` on the converter feed a model of another type or shape.

## Behaviour change

A flexible stream whose headers carry a different type or shape of the same byte count as a non-dynamic model's input is now refused. Before #4953 and this PR it was accepted. The most likely out-of-tree case is an octet stream converted to `[size]:uint8` and fed to a model of another type or shape; setting `input-dim`/`input-type` on `tensor_converter` restores it. A survey found no in-tree pipeline, example or test that relies on the old behaviour.

## Tests

New `tensorFilterFlexInput.typed*` cases in `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`. They reuse #4953's flexible input harness, bus check and #4955's `_flex_in_teardown()`. Only the float32 and multi-tensor custom-easy models and the typed buffer builder are new, because the existing helpers are uint8 and single-tensor.

| Case | Pins |
|---|---|
| `typedMatch` | a tensor of the configured type and dimension, and one of a lower rank ending in 1s, are invoked |
| `typedSameSizeMismatch_n` | int32 into float32 and a transposed shape, both of the configured size, are refused with WRONG_TYPE and not invoked |
| `typedSecondTensor_n` | two tensors are checked against their own configured tensors: a matching pair is invoked, a second tensor of the right size but another type is refused |
| `typedKeepsConfiguredInfo` | `input` / `inputtype` / `inputname` still report the model after a refusal, and a later matching tensor is invoked |
| `typedDynamicTakesType` | with invoke-dynamic the dynamic function sees the tensor's own type |

Invalid, sparse and wrong-size headers are covered by #4953's `brokenHeaderMagic_n`, `sparseHeader_n` (both modes) and `smallerThanModel_n`, so they are not repeated.

Local verification (WSL, Ubuntu 22.04):
- `unittest_filter_custom` passes 24/24: 8 pre-existing, 11 from #4953, 5 new. `unittest_plugins`, `unittest_sink` and `unittest_common` also pass.
- **Negative control** (main's `tensor_filter.c` with these tests): `typedSameSizeMismatch_n`, `typedSecondTensor_n` and `typedKeepsConfiguredInfo` fail.
- **Mutation** `(i, i)` → `(0, 0)` in the check call: caught by `typedSecondTensor_n`.
- **Memcheck** of the new cases, read with #4955's `check_valgrind_log.sh --require-output`: exit 0, with no error and no definite leak from this repository. The previous revision's tests exit 1 there, with a 17.9 KB definite leak in their private harness helper.
- **Static checks:**
  - the test object compiles with `-Werror=sign-compare -Wextra`;
  - clang-format 16 and 18 report no diff;
  - gst-indent adds no new lines over main's pre-existing ones;
  - `doxygen-tag.sh` passes.


myungjoo:

This review was produced by a separate Claude agent session (not the one that authored this PR) and is being relayed here by the author's account, as requested. It analyzed the PR diff, the surrounding `tensor_filter`/`nnstreamer_plugin_api_util` code, the referenced issue (#4928), and the existing test/doc tree. It also checked out the PR head locally and ran the new/updated test binary.

## Does it solve the stated topic?

Yes. Issue #4928 is that a flexible-format `tensor_filter` sink pad refreshes `prop->input_meta` (including `type`) from each buffer's own embedded header, then derives the "expected size" from that same refreshed copy — so the size check compares the buffer with itself and never catches a same-byte-count type substitution (the LiteRT float32/int32 repro) or a differently-shaped/sized tensor.

The fix:
- `_gst_tensor_filter_convert_meta()` now only overwrites `prop->input_meta` when `invoke-dynamic=true` (`gst/nnstreamer/tensor_filter/tensor_filter.c:695-698`); otherwise the model's configured info is left untouched, which also stops `gst_tensor_info_init()` from silently wiping `input`/`inputtype`/`inputname` (this was a secondary bug baked into the same code path).
- The new `_gst_tensor_filter_check_flexible_input()` (`tensor_filter.c:707-748`) compares each flexible tensor's parsed header (type + dimension, trailing 1s tolerated via the existing `gst_tensor_dimension_is_equal()`) against the model's corresponding input, mapped through `input-combination` when defined, and posts `STREAM/WRONG_TYPE` + refuses the buffer on mismatch or an unparsable header.
- `invoke-dynamic=true` path is untouched by design (buffer still drives input info; subplugin is responsible, matching the LiteRT `rejectTypeChange()` precedent cited in the issue).

I re-derived the logic by hand against `gst_tensor_info_is_equal()` / `gst_tensor_dimension_is_equal()` / `gst_tensor_meta_info_convert()` / `gst_tensor_meta_info_validate()` in `nnstreamer_plugin_api_util_impl.c`, and the behavior matches what the PR description claims (trailing-1 tolerance, name is never compared, no leaks in the added `GstTensorInfo buf_info` local).

## Regression check

- Fix is confined to `gst/nnstreamer/tensor_filter/tensor_filter.c`. The static (non-flexible) caps path is unaffected (`_gst_tensor_filter_convert_meta` no-ops when `!trans_data->is_flexible`, unchanged). The `invoke-dynamic=true` flexible path is unaffected (still passes `&prop->input_meta` through, so the meta-refresh behavior is bit-for-bit what it was before).
- I grepped the whole `tests/` tree for `format=flexible` combined with `tensor_filter`: the only SSAT case besides the new custom-easy unit tests is `tests/nnstreamer_filter_litert/runTest.sh:119,144`, and both already use `invoke-dynamic=true`, so they take the unchanged path. No SSAT test combines `input-combination` with a flexible pad. The three pre-existing flexible-tensor_filter GTest cases in `unittest_plugins.cc` (`flexToFlex`, `flexInvalidBuffer1_n`, `flexInvalidBuffer2_n`) all feed the correct type/dimension for their quantized tflite model, so the new type/dimension check is a no-op for them (I traced this by hand); their failures (buffer count / byte-size mismatches) are still caught by the pre-existing size/count checks, unchanged.
- I ran the PR-head build's `unittest_filter_custom` in a local WSL checkout: 17/17 pass, matching the PR's reported local run.
- I ran the repo's `tools/development/gst-indent` over the changed C file and `clang-format` over the changed C++ test file; neither produced a diff inside the PR's changed line ranges (gst-indent did flag a few unrelated pre-existing lines elsewhere in the file, outside this diff, so those are not from this PR).

## Size / scope

Proportionate: 65/3 lines in the actual fix, 2 lines of doc, and a 472-line test-only addition for a framework-wide input-validation bug — the large test file is expected and appropriate here given the number of type/shape/header permutations that need coverage. Only `tensor_filter.c`, its test file, and the one doc page are touched; no unrelated module is touched.

## Test coverage

Very thorough for the stated topic: matched type, trailing-1 lower rank, type-mismatch same/different byte count, dimension-mismatch same/different byte count, invalid header, model-info survival across a refusal, and confirmation that `invoke-dynamic=true` still lets the buffer drive input type. These are wired into the existing `unittest_filter_custom` meson test target (no `meson.build` change needed since `gst_check_dep` was already a dependency), so a regression here will fail `meson test` in CI and block merge, and will also catch future changes to `gst_tensor_dimension_is_equal`/`gst_tensor_info_is_equal`/`gst_tensor_meta_info_convert` (used elsewhere) that would otherwise silently regress this exact protection.

## Documentation

`Documentation/data-type-and-flow-control.md` gets a two-line addition stating the new invariant. That's proportionate to the change (a documented behavior clarification, not a new API/architecture change), and is included in the same PR/commit as the code change.

## Commit structure

Two commits: (1) the fix + doc update together, (2) the test-only addition. That split matches the project's own stated preference ("test case commits and feature commits being separate is good") and neither commit mixes unrelated topics.

## Security / backdoor check

No backdoor indicators. The change is a pure input-validation tightening (it makes the element *more* strict, not less), and I did not find any new externally reachable code path, credential handling, or process/network interaction introduced.

One residual gap worth a follow-up issue (not introduced by this PR, and outside the literal scope of #4928, which is worded as "type and dimension"):

- **Minor** — `gst/nnstreamer/tensor_filter/tensor_filter.c:707-748` (`_gst_tensor_filter_check_flexible_input`). The equality check goes through `GstTensorInfo`, which has no `format` field (`gst/nnstreamer/include/tensor_typedef.h:261-268`) — only `name`/`type`/`dimension`. `GstTensorMetaInfo.format` (`_NNS_TENSOR_FORMAT_STATIC/FLEXIBLE/SPARSE`) is a per-tensor, buffer-embedded field independent of the pad's negotiated caps format, and `gst_tensor_meta_info_validate()` accepts any of the three values. So a flexible-tensor header that declares the model's exact `type`/`dimension` but `format = _NNS_TENSOR_FORMAT_SPARSE` (with an attacker-chosen `nnz` such that `nnz * (element_size + 4) == expected dense byte count`) passes both the new check and the pre-existing size check, and the sparse-encoded (index,value) bytes would be handed to a non-dynamic subplugin as if they were the dense array it expects — the same reinterpretation-of-bytes problem this PR fixes for `type`, just via `format` instead. `tensor_filter.c` has no other guard against a sparse-tagged tensor on a flexible pad (`grep -n SPARSE gst/nnstreamer/tensor_filter/tensor_filter.c` is empty). I did not build a live exploit pipeline for this — it requires a hand-crafted or malicious upstream element on a `format=flexible` pad, the same threat model as the bug this PR fixes — so I'm not treating it as blocking, but it means the framework-level fix is not fully complete for the general "flexible header lies about its own encoding" class. Suggested fix (either in this PR or a fast follow-up): also reject the buffer when `meta->format != _NNS_TENSOR_FORMAT_STATIC` and `!invoke_dynamic`, since a non-dynamic subplugin never expects sparse-encoded input.

## Verdict

**OK to merge.** No Critical/Major/blocking findings. The one Minor item above is a suggestion for a follow-up issue/PR, not a reason to hold this one — it's an incremental gap in the same area, not something this PR makes worse, and closing it would require extending `GstTensorInfo` (which has no `format` field) beyond this PR's declared scope.

myungjoo:

Thanks for the review. The Minor finding (a flexible tensor tagged `sparse` slipping past the type/dimension check) is addressed in this PR rather than in a follow-up, since it is the same "header does not describe the tensor the model reads" class as #4928:

- `_gst_tensor_filter_check_flexible_input()` now refuses a tensor whose header format is `_NNS_TENSOR_FORMAT_SPARSE`, with the same `STREAM/WRONG_TYPE` element error. The suggested `format != STATIC` condition was not used: flexible producers normally stamp `format=flexible` in the per-tensor header (tensor_filter's own dynamic output does, for example), so that rule would refuse every ordinary flexible stream.
- While there, an invalid header now gets its own message and returns before its `type` and `format` are turned into strings. Before, a garbage header could hand an out-of-range value to `gst_tensor_get_type_string()` / `gst_tensor_get_format_string()`, which log a CRITICAL.
- New test `tensorFilterCustom.flexibleInputSparseHeader_n`: float32 4:2:1:1 tagged sparse with `nnz=4`, so its data size is the model's dense 32 bytes. It is refused and never invoked. It fails on the previous `tensor_filter.c`, and it would also fail on the first revision of this PR.

The changes are folded into the two existing commits (force-pushed as `4d2ca77b` and `4279648f`). Locally `unittest_filter_custom` passes 18/18. Against origin/main's `tensor_filter.c`, 7 of the new cases fail as intended. The strict `-Werror=sign-compare` compile, clang-format 16/18, gst-indent and the doxygen-tag check are clean.

myungjoo:

This is a re-review of the force-pushed head (`4279648f`), produced by a separate Claude agent session from the one that authored the PR or the prior review, and relayed here by the author's account. It re-read the full current diff against `main` (not just the delta since the last review), re-derived the logic against `nnstreamer_plugin_api_util_impl.c`, re-checked out the PR head locally, and re-ran the updated test binary in a local build.

## Does it solve the stated topic (#4928)?

Yes, and the fix is unchanged in substance from the previous revision:

- `_gst_tensor_filter_convert_meta()` only refreshes `prop->input_meta` from the buffer's header when `invoke-dynamic=true` (`gst/nnstreamer/tensor_filter/tensor_filter.c:687-697`), so the size check downstream no longer compares the buffer against a copy of itself.
- The new `_gst_tensor_filter_check_flexible_input()` (`tensor_filter.c:707-748`) compares each flexible tensor's parsed type/dimension against the model's corresponding input (through `input-combination` when defined), tolerating trailing 1s via the existing `gst_tensor_dimension_is_equal()`, and refuses the buffer with `STREAM/WRONG_TYPE` on a mismatch or an unparsable header.
- `invoke-dynamic=true` is untouched: the buffer still drives input info there.

## Verifying the previously-found Minor item is actually resolved

The prior review's only finding was that a flexible tensor whose header declares the model's exact type/dimension but is tagged `format = _NNS_TENSOR_FORMAT_SPARSE` (sparse `(index,value)` pairs sized to equal the model's dense byte count) would pass both the new check and the pre-existing size check, since `GstTensorInfo` carries no `format` field.

Confirmed fixed: `tensor_filter.c:734-735` now computes
```c
equal = (meta->format != _NNS_TENSOR_FORMAT_SPARSE &&
    gst_tensor_info_is_equal (&buf_info, model_info));
```
so a sparse-tagged tensor is refused regardless of type/dimension match. I built the exact scenario from the finding (`flexibleInputSparseHeader_n`: float32 4:2:1:1 tagged sparse, `nnz=4` → 4×(4+4)=32 bytes, exactly the model's dense size) and traced it: `gst_tensor_meta_info_validate()` accepts it (format is a valid enum value, so it reaches the new check, not the invalid-header branch), `gst_tensor_meta_info_convert()` still copies `type`/`dimension` correctly regardless of `format` (it's a flat field copy, `nnstreamer_plugin_api_util_impl.c:1647-1662`), and the added `format != SPARSE` clause is what rejects it. I ran this test locally against the PR-head build (WSL) and it passes (buffer refused, `invoke_count == 0`, `STREAM/WRONG_TYPE` posted); the accompanying local run the author reported (7/18 failing against `origin/main`'s `tensor_filter.c`) is consistent with a real, previously-absent check.

## Is "why not `format != STATIC`" correct?

Yes. I verified independently (not just taken on the author's word) that ordinary flexible producers stamp `format = _NNS_TENSOR_FORMAT_FLEXIBLE`, not `_STATIC`, in the per-tensor header:
- `gst/nnstreamer/elements/gsttensor_converter.c:1187,1816`, `gst/nnstreamer/elements/gsttensor_crop.c:532`, and `tensor_filter.c:1136` (the element's own flexible/dynamic output path) all explicitly set `meta.format = _NNS_TENSOR_FORMAT_FLEXIBLE`.
- The enum (`tensor_typedef.h:195-197`) is `STATIC=0, FLEXIBLE, SPARSE`.

So a `format != _NNS_TENSOR_FORMAT_STATIC` condition would reject `_FLEXIBLE`-tagged headers too — i.e. every ordinary buffer on a `format=flexible` pad — which is exactly what the author's response claims. The chosen `== _NNS_TENSOR_FORMAT_SPARSE` condition is the correct, narrower rule: it only excludes the one format whose bytes are not the dense array the model reads.

## New in this revision: invalid-header ordering

`_gst_tensor_filter_check_flexible_input()` now calls `gst_tensor_meta_info_validate()` before touching `meta->type`/`meta->format` for string conversion (`tensor_filter.c:723-728`), so a garbage header no longer risks handing an out-of-range enum value to `gst_tensor_get_type_string()`/`gst_tensor_get_format_string()` (both of which would log a CRITICAL on an out-of-range input). This is a real, independent hardening — confirmed by reading `gst_tensor_meta_info_validate()` (`nnstreamer_plugin_api_util_impl.c:1502-1535`), which range-checks `type` and `format` before anything else touches them. `flexibleInputInvalidHeader_n` exercises this path and passes.

## Regression check

- The static (non-flexible) caps path is untouched (`is_flexible` guard, unchanged).
- The one call site of `_gst_tensor_filter_convert_meta()` is the one touched in this diff (verified by grep — no other caller exists), so there is no second place silently relying on the old always-refresh behavior.
- `gst_tensor_info_is_equal()` compares only `type` and `dimension` (`nnstreamer_plugin_api_util_impl.c:197-214`), confirming the PR's claim that tensor **name** is never part of the check (a model input's declared name legitimately need not match an upstream flexible producer's, if any).
- I re-ran the full `unittest_filter_custom` suite (not just the new cases) on the current head in a local WSL build: 18/18 pass, matching the PR's reported count.
- `gst-indent` on a scratch copy of `tensor_filter.c` and `clang-format` on a scratch copy of the test file: both clean on every changed line (gst-indent flags four pre-existing, unrelated lines elsewhere in the file — none inside this diff's hunks).

## Size / scope

Proportionate: the functional change is confined to `tensor_filter.c` (plus the sparse clause and the reordered validation, both a few lines), a 2-line doc addition, and a large but test-only addition to `unittest_filter_custom.cc`. No `meson.build` change was needed since `gst_check_dep` (which provides `gst/check/gstharness.h`) was already a listed dependency for that target (`tests/meson.build:65-73`) — confirmed by the successful build. No unrelated module is touched.

## Test coverage / future-regression detection

Thorough for the stated topic and wired into the existing `meson test` target, so CI (`ubuntu_clean_meson_build.yml` and the GBS build) will fail and block merge if this protection regresses, or if a future change to `gst_tensor_dimension_is_equal()` / `gst_tensor_info_is_equal()` / `gst_tensor_meta_info_convert()` / `gst_tensor_meta_info_validate()` (all reused by this check) silently weakens it.

## Minor — documentation is incomplete for the new sparse-rejection behavior

**File:** `Documentation/data-type-and-flow-control.md:111`

The added sentence reads:

> A `tensor_filter` with flexible input checks the type and dimension in each tensor's meta against the model's input, and stops the stream with an error when they differ. Only with `invoke-dynamic=true` may the flexible tensors change type or shape per buffer; the subplugin then decides what it accepts.

This describes only the type/dimension check. It was written for the first push and not updated when the sparse-format rejection was folded in during this force-push, so it now under-describes the actual behavior: a tensor whose type and dimension *do* match the model, but whose header is tagged `format=sparse`, is also refused (`tensor_filter.c:734-735`), which this sentence does not say. Since this page is explicitly aimed at "NNStreamer engine developer[s]" who need the exact framework contract (per the page's own preceding paragraph), and getting this wrong is exactly the class of bug #4928 was about, the doc should state the actual rule.

Suggested fix — extend the added sentence, e.g.:

> A `tensor_filter` with flexible input checks the type and dimension in each tensor's meta against the model's input, and refuses a tensor tagged `sparse` even if its type, dimension, and byte count all match. It stops the stream with an error in either case. Only with `invoke-dynamic=true` may the flexible tensors change type, shape, or format per buffer; the subplugin then decides what it accepts.

## Security / backdoor check

No indicators. The change only tightens input validation; no new externally reachable code path, credential handling, or process/network interaction.

## Verdict

**Changes requested** — one Minor item (documentation gap above). No Critical/Major findings; the fix itself, its regression safety, and its test coverage are all sound, and the previously-reported Minor (sparse bypass) is genuinely fixed with a technically correct rationale for the chosen condition.

myungjoo:

Addressed the Minor documentation finding. `Documentation/data-type-and-flow-control.md` now says that a flexible tensor is also refused when it is tagged sparse, and that only `invoke-dynamic=true` lets the type, shape or format change per buffer. The fix is folded into the first commit (force-pushed; the head is now `20cebec6`). No code or test changed in this push.

myungjoo:

This is round 3, re-reviewing the current head (`20cebec6`), produced by a separate Claude agent session from the ones that authored the PR or the two prior reviews (comments 5617417622, 5617600768), and relayed here by the author's account. It re-read the full diff against `main` (not just the delta since round 2), independently re-derived the logic against `nnstreamer_plugin_api_util_impl.c`, re-checked out the PR head locally, and re-ran the updated test binary in a local WSL build (18/18 pass, matching the PR's reported count).

## Does it solve the stated topic (#4928)?

Yes, unchanged in substance from the previous two revisions:

- `_gst_tensor_filter_convert_meta()` refreshes `prop->input_meta` from a flexible buffer's own header only when `invoke-dynamic=true` (`gst/nnstreamer/tensor_filter/tensor_filter.c:687-701`), so the framework's size check no longer compares the buffer against a copy of itself.
- `_gst_tensor_filter_check_flexible_input()` (`tensor_filter.c:707-755`) compares each flexible tensor's parsed type/dimension against the model's corresponding input (mapped through `input-combination` when defined), tolerating trailing 1s via `gst_tensor_dimension_is_equal()`, and refuses the buffer with `STREAM/WRONG_TYPE` on a mismatch, a sparse-tagged header, or an unparsable header.
- `invoke-dynamic=true` is untouched: the buffer still drives input info, matching the LiteRT `rejectTypeChange()` precedent cited in the issue.

This is a framework-level fix (the only call sites of both functions are the two shown above, confirmed by grep), so every subplugin reachable through a flexible sink pad is covered, addressing the issue's explicit ask ("the framework refusing a type the subplugin did not report would fix it once for everyone").

## Verifying both previously-reported findings are actually resolved

**Round 1 finding (sparse-format bypass, comment 5617417622):** a flexible tensor whose header declares the model's exact type/dimension but is tagged `format = _NNS_TENSOR_FORMAT_SPARSE` would pass the type/dimension check and the pre-existing size check, since `GstTensorInfo` carries no `format` field.

Confirmed still fixed at the current head: `tensor_filter.c:734-735`

```c
equal = (meta->format != _NNS_TENSOR_FORMAT_SPARSE &&
    gst_tensor_info_is_equal (&buf_info, model_info));
```

I independently traced `gst_tensor_meta_info_validate()` (`nnstreamer_plugin_api_util_impl.c:1502-1535`, range-checks `type`/`dimension`/`format` before anything touches them) and `gst_tensor_meta_info_convert()` (`:1647-1662`, a flat field copy of `type`/`dimension` regardless of `format`) to confirm the `format != SPARSE` clause is what actually rejects the crafted case, not a side effect of something else. `flexibleInputSparseHeader_n` (float32 4:2:1:1 tagged sparse, `nnz=4` → 32 bytes, exactly the model's dense size) passes in a local WSL build of this exact head: buffer refused, `invoke_count == 0`, `STREAM/WRONG_TYPE` posted.

I also re-verified the "why not `format != STATIC`" rationale independently: `gsttensor_converter.c:1187,1816`, `gsttensor_crop.c:532`, and `tensor_filter.c`'s own dynamic-output path all stamp `format = _NNS_TENSOR_FORMAT_FLEXIBLE` (not `_STATIC`) on ordinary flexible output, so a `!= STATIC` condition would reject every normal flexible buffer. The narrower `== SPARSE` condition used here is correct.

**Round 2 finding (documentation gap, comment 5617600768):** the doc sentence added in round 1 described only the type/dimension check and did not mention the sparse-format rejection folded in during round 1's force-push.

Confirmed fixed: `Documentation/data-type-and-flow-control.md:111` now reads:

> A `tensor_filter` with flexible input checks the type and dimension in each tensor's meta against the model's input, and stops the stream with an error when they differ or when the tensor is tagged sparse. Only with `invoke-dynamic=true` may the flexible tensors change type, shape or format per buffer; the subplugin then decides what it accepts.

This states the actual current rule (mismatch or sparse tag → refused; `invoke-dynamic` changes type/shape/format all together), matching the round 2 suggested wording closely enough to close the gap. No code or test changed in this push (verified: the only diff between the round-2-reviewed head and `20cebec6` is this one doc line), so nothing else needed re-verification here.

## Regression check

- The static (non-flexible) caps path is untouched (`is_flexible` guard, unchanged).
- Only one call site each for `_gst_tensor_filter_convert_meta()` and `_gst_tensor_filter_check_flexible_input()` (verified by grep), so no other code path silently relies on the old always-refresh behavior or bypasses the new check.
- `gst_tensor_info_is_equal()` compares only `type` and `dimension` (`nnstreamer_plugin_api_util_impl.c:197-216`) — tensor **name** is never part of the check, so a model input's declared name need not match an upstream flexible producer's.
- Full `unittest_filter_custom` suite (not just the new cases) on the current head, local WSL build: 18/18 pass.
- The `input-combination` mapping (`tensor_filter.c:858-862` vs. `:876-880`) correctly threads `model_idx`/`info_idx` through both the combination-defined and default paths; an out-of-range `model_idx` is deliberately left to the existing size check rather than double-reported.

## Size / scope

Proportionate and unchanged from round 2's assessment: the functional change is confined to `tensor_filter.c` (~75 lines), a 2-line doc addition, and a large but test-only addition to `unittest_filter_custom.cc`. `tests/meson.build` needed no change since `gst_check_dep` (providing `gst/check/gstharness.h`) was already listed for this test target (`tests/meson.build:65-73`) — confirmed by reading the file directly. No unrelated module is touched.

## Test coverage / future-regression detection

Thorough for the stated topic (matched type, trailing-1 lower rank, type-mismatch same/different byte count, dimension-mismatch same/different byte count, invalid header, sparse-tagged same-byte-count, model-info survival across a refusal, and `invoke-dynamic` unaffected) and wired into the existing `meson test` target, so CI will fail and block merge on a regression here, or on a future change to `gst_tensor_dimension_is_equal()` / `gst_tensor_info_is_equal()` / `gst_tensor_meta_info_convert()` / `gst_tensor_meta_info_validate()` that silently weakens this protection.

## Documentation / architecture

No API or architecture change (no new properties, no signature change to any public function). The one-paragraph addition to `Documentation/data-type-and-flow-control.md` is proportionate and, as verified above, now accurately states the full current rule.

## Security / backdoor check

No indicators. The change only tightens input validation (rejects cases the framework previously accepted); no new externally reachable code path, credential handling, or process/network interaction is introduced.

## Verdict

**OK to merge.** Both previously-reported findings (sparse-format bypass, documentation gap) are independently confirmed resolved at `20cebec6`. No new Critical/Major/Minor/Nit findings from this round's full re-review.

myungjoo:
Coordination note posted by Codex at the maintainer's request: **merge #4953 first, then continue #4957 on top of the merged main branch.** This records the intended order; it does not mean #4953 has already merged.

Please preserve #4953's fixes when rebasing and resolving conflicts. The overlap was documented in [the latest review of #4953](https://github.com/nnstreamer/nnstreamer/pull/4953#pullrequestreview-5167041375) (reviewed heads: #4953 `c5ced7f1`, #4957 `20cebec6`).

### Preserve these invariants from #4953

- **Validate before reading/using a flexible header:** check the mapped memory length before parsing, check parse success, reject zero/unsizable or out-of-bounds header sizes, and require the dense size described by the header to equal the bytes after the header. Do not reintroduce an out-of-bounds header read or `size - hsize` underflow.
- **Reject sparse-tagged input in both static and dynamic invoke modes.** Do not move this protection under `!invoke_dynamic`. Ordinary FLEXIBLE-tagged dense tensors must remain accepted.
- **Preserve the input metadata contract:** without `invoke-dynamic`, leave configured model metadata intact. With `invoke-dynamic`, update only type and dimension from each validated input header and retain the configured tensor name. In particular, do not restore conversion directly into the persistent `prop->input_meta` entry, which loses the name.
- **Keep B9's dynamic output validation and cleanup**, including both synchronous failure propagation and asynchronous partial-buffer disposal. Invalid output info must not produce a downstream buffer; preserve the ownership split between already-appended memory and pending raw outputs.
- **Keep B7's `gst_tensors_config_free (&config)`** on the static-output caps mismatch failure path, and retain the updated `input_meta` API documentation.

### Keep #4957 focused on its additional protection

After rebasing, build the non-dynamic model **type/dimension equality check** on top of #4953's validated input metadata, preserving trailing-one dimension equivalence and the intended input mapping. Consolidate overlapping parsing/refresh/sparse checks instead of replacing #4953's implementation with the pre-rebase #4957 version. Retain regression coverage when consolidating duplicate tests.

There is a known error-reporting difference to resolve explicitly: #4953's early invalid/sparse-header refusal logs and returns `GST_FLOW_ERROR`, while #4957's `flexibleInputInvalidHeader_n` and `flexibleInputSparseHeader_n` require a `STREAM/WRONG_TYPE` bus message. The review reported those two failures in a combined tree. Align the intended bus-error contract and tests without weakening or delaying the early safety checks. If retaining the bus-message requirement, emit the element error from the early refusal path.

Update `Documentation/data-type-and-flow-control.md` and the PR description for the combined behavior. In particular, the current statement that dynamic mode may change "type, shape or format" must not imply that dynamic mode permits sparse input: #4953 rejects sparse in both modes. Dynamic type/shape updates remain supported, with tensor names preserved.

### Regression verification after rebase

Run the full `unittest_filter_custom` suite with #4953's `tensorFilterFlexInput.*` coverage retained, especially short memory, invalid headers, unsizable versions, header/payload mismatch, smaller-than-model input, configured-name preservation, per-buffer dynamic metadata, and invalid synchronous/asynchronous output info. Keep #4957's additional same-byte-count type/shape mismatch and trailing-one tests. Add a dynamic-mode sparse-refusal case so conflict resolution cannot silently narrow #4953's protection. Run the relevant existing plugin tests and CI on the integrated result.

Please use the actual merged #4953 implementation as the baseline, including any fixes added after `c5ced7f1`.

myungjoo:

Rebased on merged #4953 as agreed in the coordination note, and addressed the open points of the review on `20cebec6`. The new head is `00883e93`, still two commits.

**Coordination note (#4953 invariants).** `tensor_filter.c` takes main's version as the base and adds only `_gst_tensor_filter_check_flexible_input()` plus its two call sites. Header validation, sparse refusal in both modes, the name-preserving dynamic refresh, the size check and its message, B9 and B7 are untouched.

The bus-error difference is resolved by ordering. The new comparison runs **after** the size check, so it only sees tensors of the configured size, which is the gap #4928 leaves. A smaller or larger tensor is still refused by the size check with the same reporting as before, and `smallerThanModel_n` passes unchanged. With the comparison placed before the size check, that test had failed because the new WRONG_TYPE error preempted the size refusal.

Dynamic-mode sparse refusal is already pinned by `sparseHeader_n`, which runs both modes, so no duplicate was added.

**Review findings:**
1. *Overlap with #4953:* done as suggested. The duplicate `info ? … : NULL`, validate and sparse parts are gone. `flexibleInputInvalidHeader_n`, `flexibleInputSparseHeader_n`, `flexibleInputDimOtherSize_n` and `flexibleInputTypeOtherSize_n` were dropped in favour of #4953's `brokenHeaderMagic_n`, `sparseHeader_n` and `smallerThanModel_n`.
2. *Documentation:* the paragraph now says that invalid, mismatched-size or sparse headers stop the stream in both modes. It says that without `invoke-dynamic` the type and dimension must match the configured input ("the model's, or the `input` and `inputtype` properties"), and that only `invoke-dynamic=true` lets type or shape change. "Format" is no longer mentioned.
3. *Octet-stream behaviour change:* documented in the same paragraph (use `input-dim`/`input-type` on `tensor_converter`) and in a "Behaviour change" section of the PR description.
4. *Index mapping uncovered:* added `flexibleInputTwoTensors` (float32 + int32 model, both matching) and `flexibleInputSecondTensorType_n` (tensor 1 has the same bytes but another type). The `(i, i)` → `(0, 0)` mutation now fails `flexibleInputSecondTensorType_n`.
5. *Header format in the message:* the message now uses `gst_tensor_info_to_string()` for both sides, e.g. `int32 (4:2:1:1)` vs `'in0' float32 (4:2:1:1)`, without the format.

**Local verification:**
- `unittest_filter_custom` 27/27, and `unittest_plugins`, `unittest_sink`, `unittest_common` pass.
- Against main's `tensor_filter.c`, the four same-byte-count cases fail.
- The strict `-Werror=sign-compare -Wextra` compile, clang-format 16/18, gst-indent and doxygen-tag are clean.

myungjoo:

This is a review of the rebased head (`00883e93`), produced by a separate Claude agent session from the ones that authored the PR or any of the four prior rounds (three automated reviews plus the myungjoo-bot review on `20cebec6`, plus the coordination note from "Codex at the maintainer's request"), and relayed here by the author's account. It re-read the full current diff against `main` (base = `origin/main` tip `7bfaa872`, confirmed via `git merge-base`, so the diff shown by `gh pr diff` is the complete and only change to review), re-derived the logic against the merged `tensor_filter.c` and `nnstreamer_plugin_api_util_impl.c`, checked out the PR head locally (matches the worktree exactly), rebuilt and ran the updated test binary, and additionally hand-verified the one previously-unverified claim (the `(i,i)->(0,0)` mutation) by tracing the loop logic against the actual code.

## Does it solve the stated topic (#4928)?

Yes. After #4953 merged (header validation, sparse refusal in both modes, name-preserving dynamic refresh, per-tensor size check, B9, B7), the remaining gap was that a flexible tensor of the *same byte count* as the configured input but of another type or shape still passed the size check and reached a non-dynamic subplugin unchanged. This PR closes exactly that gap with one new function, `_gst_tensor_filter_check_flexible_input()` (`gst/nnstreamer/tensor_filter/tensor_filter.c:725-757`), called from both branches of `_gst_tensor_filter_transform_get_invoke_tensors()` (`:874-877`, `:893-896`), immediately after the existing per-tensor size check and only when `!prop->invoke_dynamic`.

## Coordination-note invariants (verified against the actual code on `main`, not just taken on trust)

The PR is rebased cleanly onto merged `main` (`git merge-base HEAD origin/main` == `origin/main`'s tip), so the diff touches only two hunks in `tensor_filter.c`: the new 33-line function, plus a 6-line call-site insertion in each of the two branches of `_gst_tensor_filter_transform_get_invoke_tensors()`. Nothing else in the 1500+ line file is touched. Concretely, on today's `main`:

- **Header validation before use / sparse refusal in both modes** — `_gst_tensor_filter_convert_meta()` (`:678-711`, untouched by this diff) still checks `map->size` against the header size, checks `parse_header()`'s return, rejects `hsize == 0 || hsize > map->size`, rejects `_NNS_TENSOR_FORMAT_SPARSE` unconditionally (before the `invoke_dynamic` branch, so both modes), and requires the dense size to equal the post-header bytes. All of this is outside the diff.
- **input_meta contract / name preservation** — in the same untouched function, the `invoke_dynamic` branch only does `_info->type = info.type; memcpy(_info->dimension, ...)`; `_info->name` is never touched, so the configured name survives per-buffer dynamic refresh.
- **B9 (dynamic output validation/cleanup) and B7 (`gst_tensors_config_free` on the static-output caps-mismatch path)** — neither region appears anywhere in the diff; confirmed by grep that the only diff hunks are the two described above.
- **Ordering vs. the size check** — in both the `input-combination` branch and the default branch, the new call sits strictly after the existing `expected != trans_data->tensors[i].size` check, which returns early (without invoking the new function) on a size mismatch. I confirmed this preserves `tensorFilterFlexInput.smallerThanModel_n` by reading the test (it pushes a 16-byte payload against a model that needs more, expects `GST_FLOW_ERROR` with `_flex_in_refused_header(h)` == FALSE, i.e. the *un*-decorated size-check error path, not the new `STREAM/WRONG_TYPE`) and by running the full suite (below) where it still passes. Placing the new check first, as the pre-rebase revision did, would have made this test's error path go through the new WRONG_TYPE branch instead — which is exactly the regression the coordination note called out.

## Same-byte-count coverage for every tensor index

The default (non-combination) call is `_gst_tensor_filter_check_flexible_input(self, trans_data, i, i)` inside the `for (i = 0; i < prop->input_meta.num_tensors; i++)` loop, so every tensor index is checked against its own configured input, not just index 0. I traced the specific regression this must guard against: with the buggy `(i,i) -> (0,0)` substitution, at `i == 1` the function would compare `trans_data->meta[0]` (tensor 0's header) against `conf_idx == 0` (tensor 0's config) again, never touching tensor 1's actual header — so a same-byte-count type mismatch in any tensor after the first would silently pass. `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`'s new `flexibleInputSecondTensorType_n` (two-tensor model, tensor 0 matches, tensor 1 is `int32` where `float32` of the same byte count is configured, expects `GST_FLOW_ERROR` + `invoke_count == 0` + `STREAM/WRONG_TYPE`) is exactly the case this mutation would silently accept, so this closes the gap the earlier myungjoo-bot review (finding 4) reported. The `input-combination` branch threads `info_idx` (the config side) separately from `i` (the buffer side) through the same function, which is the correct generalization; as before, this path is unreachable today because flexible caps carry `num_tensors = 0` (so `input-combination` cannot negotiate against them) — this is a pre-existing, previously-flagged limitation, not something this PR needs to newly test.

## Resolution of the five earlier myungjoo-bot findings (round on `20cebec6`)

1. **[Medium] Overlap/conflict with #4953** — Resolved by doing exactly what the coordination note asked: rebase onto merged #4953, drop the duplicate `info ? … : NULL` / validate / sparse logic, keep only the type+dimension comparison. Verified above: the diff no longer touches any of #4953's code.
2. **[Low] Doc said invoke-dynamic changes "format" and the subplugin "decides" on it** — The current doc paragraph (`Documentation/data-type-and-flow-control.md:111`) no longer mentions "format" in the invoke-dynamic sentence; it separately states that invalid/mismatched-size/sparse headers stop the stream "in both modes," which matches the code (sparse is rejected unconditionally in `_gst_tensor_filter_convert_meta`, not gated on `invoke_dynamic`). Resolved.
3. **[Low] Undocumented octet-stream behaviour change** — Same doc paragraph now ends with: "Note that `tensor_converter` turns `application/octet-stream` into a `uint8` tensor of the buffer size; set its `input-dim` and `input-type` to feed a model of another type or shape." The PR description also carries an explicit "Behaviour change" section with the same guidance. Resolved.
4. **[Low] Index mapping untested (single-tensor models only)** — Addressed by the new `flexibleInputTwoTensors` (positive, two matching tensors of different types) and `flexibleInputSecondTensorType_n` (negative, second tensor mismatches) tests; verified the mutation this was meant to catch is caught (see above). Resolved.
5. **[Nit] Error message printed a confusing "format: static" for ordinary producers** — The current message (`tensor_filter.c:742-750`) builds both sides with `gst_tensor_info_to_string()`, which has no `format` field, e.g. `int32 (4:2:1:1)` vs `'in0' float32 (4:2:1:1)`. Format is no longer printed. Resolved.

## Tests

- Ran the exact local reproduction environment the author used (WSL build matching this exact head): `unittest_filter_custom` → **27/27 pass** (16 `tensorFilterCustom` + 11 `tensorFilterFlexInput`), matching the PR's reported count exactly.
- Verified the "removed as duplicate" claim is legitimate: `flexibleInputInvalidHeader_n`, `flexibleInputSparseHeader_n`, `flexibleInputDimOtherSize_n`, `flexibleInputTypeOtherSize_n` are gone from the diff, and their replacements (`brokenHeaderMagic_n`, `sparseHeader_n` — which the code shows runs in *both* dynamic and non-dynamic mode — and `smallerThanModel_n`) are present and unmodified in the current tree from #4953. The two dropped "OtherSize" cases tested a *different* byte count with a type/shape mismatch, which the framework's pre-existing size check already caught before either PR; they were redundant with `smallerThanModel_n`-style coverage, not a unique protection this PR removed.
- All new/kept model names registered via `NNS_custom_easy_register`/`_dynamic_register` in the test file are unique (checked the full file) — no risk of one test's `TearDown` colliding with another's fixture.
- No `meson.build` change was needed or made; `unittest_filter_custom` is already a registered `test()` target (`tests/meson.build:260-267`), so a future regression here fails `meson test` and blocks CI (Tizen GBS `unit_test 1`, Valgrind job, etc.), and it also transitively re-exercises `gst_tensor_dimension_is_equal()`/`gst_tensor_info_is_equal()`/`gst_tensor_meta_info_convert()` so an unrelated future change that weakens any of those is caught here too.
- Style: independently ran `clang-format` (v14, repo `.clang-format`) over the changed test file — zero diff; ran `tools/development/gst-indent` over the changed `tensor_filter.c` — the only diff lines are four pre-existing hunks at lines 1040/1092/1368/1981, all outside this PR's changed ranges (707-757, 874, 893). Doxygen `@brief` present on every new function/test/struct in the diff.
- CI at review time: most checks green (Static checks, CPP-Linter, DCO, Android NDK, macOS, LLVM/Clang, ExecuTorch, both Valgrind meson jobs, both non-x86_64 Tizen GBS builds, pdebuild resolute); the x86_64 Tizen GBS `unit_test 1` job (the one that actually runs `unittest_filter_custom`) and a few others were still pending/running at the time of this review, not failed.

## Size / scope

Proportionate and minimal: one new static function plus two 4-line call sites in `tensor_filter.c` (~50 net lines), a one-sentence doc addition, and a test-only addition. No unrelated module touched; no API or architecture change (no new properties, no public signature changes), so no further documentation is owed beyond the paragraph already updated in `Documentation/data-type-and-flow-control.md`.

## Security / backdoor check

No indicators. The change only narrows what reaches non-dynamic subplugins; no new externally reachable code path, credential handling, or process/network interaction.

## Verdict

**OK to merge.** All of the coordination note's invariants are verified preserved against the actual code on `main`, the same-byte-count check is verified to cover every tensor index (including a hand-traced confirmation of the specific mutation the earlier review used to find the gap), all five previously-reported findings are resolved, and the test suite (27/27, run locally against this exact head) both reproduces the fix and is wired into the CI-gating `meson test` target. No new Critical/Major/Minor/Nit findings from this round's independent re-review.

myungjoo:

Both findings of the review on `00883e93` are addressed. The head is now `7e036408`, rebased on main `120d2d17`, which includes the merged #4955. It is still two commits, and only the test commit changed.

**1. [Medium] Harness/element/bus leak.** Fixed by removing the private helper family instead of patching it. The new cases use `_flex_in_harness()`, `_flex_in_refused_header()` and `_flex_in_teardown()` from main, the helpers #4955 already fixed, so the flushing teardown applies to them automatically. Verified with memcheck and #4955's `check_valgrind_log.sh --require-output`:
- previous test block (committed `00883e93` tests on the rebased tree): checker exit 1, with a 17,949-byte definite leak in `_flex_new_harness (unittest_filter_custom.cc:1262)`;
- new block (`tensorFilterFlexInput.typed*`): checker exit 0, with 0 error contexts and 0 definite leak contexts from this repository.

**2. [Nit] Parallel helper family and suite.** The duplicates `FlexCbData`, `_flex_new_harness` and `_flex_expect_wrong_type_error` are gone, and the cases now live in `tensorFilterFlexInput`. What remains is what the uint8, single-tensor helpers cannot express: a float32 or multi-tensor custom-easy model (`_flex_in_typed_register` plus its two filter callbacks) and a typed buffer builder (`_flex_in_typed_append` / `_flex_in_typed_push`). While consolidating, the eight cases became five without losing a pinned behaviour:
- `typedMatch` covers exact and trailing-1 dimensions;
- `typedSameSizeMismatch_n` covers a type and a shape substitution;
- `typedSecondTensor_n` covers the matching pair and the second-tensor refusal.

The test commit went from +474 to +280 lines.

**Re-verified locally:**
- `unittest_filter_custom` 24/24.
- Against main's `tensor_filter.c`, `typedSameSizeMismatch_n`, `typedSecondTensor_n` and `typedKeepsConfiguredInfo` fail.
- The `(i, i)` → `(0, 0)` mutation fails `typedSecondTensor_n`.
- clang-format 16 and 18 report no diff, and `doxygen-tag.sh` passes.
- The hold label is back on until the re-review and CI finish.

myungjoo:

This review was produced by a separate Claude agent session from the ones that authored the PR or wrote any of the prior review rounds (the three earlier automated reviews, the myungjoo-bot round on `20cebec6`, the myungjoo-bot round on `00883e93`, and the author's response), and is relayed here by the author's account. It re-read the full current diff against `main`, checked out the exact PR head (`7e036408`, rebased on `main` `120d2d17`, which includes merged #4953 and #4955) in a local worktree, rebuilt and ran the updated test binary, and independently reproduced the memcheck evidence the author's response cites.

## Does it solve the stated topic (#4928)?

Yes, unchanged from the prior rounds' assessment. `_gst_tensor_filter_check_flexible_input()` (`gst/nnstreamer/tensor_filter/tensor_filter.c:721-757` at this head) runs after the existing per-tensor size check, only when the input is flexible and `invoke_dynamic` is false, and refuses a same-byte-count tensor whose type or dimension does not match the configured input. This closes exactly the gap #4953 left open.

## tensor_filter.c and documentation: confirmed unchanged from the previously-approved content

Diffed the current `tensor_filter.c` and `Documentation/data-type-and-flow-control.md` hunks against what the `00883e93`-round review quoted verbatim: the new function body, its two call sites (default loop and the `input-combination` loop), and the one added documentation paragraph are byte-for-byte the same, just at shifted line numbers from the rebase onto `120d2d17`. The commit history confirms only the test commit changed (`e0b23429` fix+doc is untouched; the tests were amended in place rather than fixed up with a new commit, matching the reviewer's request to keep the two-commit shape). No #4953 code is touched.

## Leak finding: verified fixed

Reproduced the medium/blocking finding's fix directly rather than taking the author's numbers on trust:

- Built the exact PR head in a local Ubuntu 22.04 environment and ran `unittest_filter_custom` with `--gtest_filter=tensorFilterFlexInput.*`: **16/16 pass** (11 pre-existing + 5 new `typed*` cases), and the full binary: **24/24 pass**, matching the PR description's reported counts exactly.
- Ran memcheck (valgrind, full leak-check) restricted to the five new `typedMatch`/`typedSameSizeMismatch_n`/`typedSecondTensor_n`/`typedKeepsConfiguredInfo`/`typedDynamicTakesType` cases, then ran `tools/debugging/check_valgrind_log.sh --repo-root ... --require-output` on the resulting log:
  ```
  valgrind log check: 0 error contexts from this repository, 0 from libraries, 2 leak reports,
  0 definite leak contexts from this repository, 1 binaries examined
  CHECKER_RC=0
  ```
  The two remaining "definitely lost" blocks are both rooted in `_dl_open`/`_dl_init` (the dynamic loader's own internal allocation), not in this repository's code, and the checker correctly exempts them. This confirms the fix: removing the private `_flex_new_harness`/`FlexCbData` family and switching to main's shared `_flex_in_harness()` / `_flex_in_teardown()` (which flushes the element's private bus with `gst_bus_set_flushing()` before `gst_harness_teardown()`) eliminates the ref cycle that leaked the harness, element and bus in every one of the previous round's 8 tests. All five new tests, including the `_n` (refusal) paths, go through this same teardown.
- Checked the new tests' own allocations for leaks beyond the bus cycle: `_flex_in_typed_append()` wraps a `g_malloc0`'d payload in a `GstMemory` with a `g_free` destroy notifier, calls `gst_tensor_meta_info_append_header()` (which per its own doc comment copies rather than takes ownership of the passed memory) and then correctly `gst_memory_unref()`s the original wrapper; `_flex_in_typed_register()` frees its scratch `GstTensorsInfo` after registration; `_flex_in_typed_dynamic()`'s `_g_memdup()`'d output buffer follows the same ownership pattern already used by `_flex_in_dynamic()` elsewhere in this file. No new leak source found.

## Nit finding: verified resolved

`grep`-ing the current file confirms no `_flex_new_harness`, `FlexCbData` or `_flex_expect_wrong_type_error` remain; all five new tests call `_flex_in_harness()`, `_flex_in_refused_header()` and `_flex_in_teardown()` (main's helpers) and are registered under the `tensorFilterFlexInput` suite, matching the PR body's description and eliminating the parallel-helper-family maintenance risk the nit raised.

## Consolidation: every previously-pinned behaviour is still covered

Traced each of the five remaining cases against the eight behaviours the `00883e93`-round mutation matrix pinned:

| Behaviour previously pinned | Where it lives now |
|---|---|
| Same-byte-count **type** mismatch refused | `typedSameSizeMismatch_n`, iteration 0 (`int32` into a `float32` `4:2:1:1` model input) |
| Same-byte-count **shape** mismatch refused | `typedSameSizeMismatch_n`, iteration 1 (`float32` `2:4:1:1` into the same `4:2:1:1` model input — same type, transposed dims, same 32 bytes) |
| Trailing-1 dimension tolerance | `typedMatch`, second iteration (`4:2` accepted against a configured `4:2:1:1` input) |
| Per-index check on a multi-tensor input | `typedSecondTensor_n`: a two-tensor push where tensor 0 matches and tensor 1 doesn't (same byte count, wrong type) is refused, and a fully-matching two-tensor push is invoked with `invoke_count == 1` and correct per-tensor sizes recorded |
| Configured info (`input`/`inputtype`/`inputname`) kept after a refusal, later match still invoked | `typedKeepsConfiguredInfo` |
| Dynamic mode takes the tensor's own type, not the configured one | `typedDynamicTakesType` |

Nothing from the previous eight cases is dropped: the two behaviours folded into `typedSameSizeMismatch_n` (type-only and shape-only mismatch) were originally two separate tests and are now two iterations of one loop over the same assertions, and the previously separate two-tensor positive case is folded into `typedSecondTensor_n`'s first half. Independently re-created the `(i, i)` → `(0, 0)` regression by tracing the loop: with that substitution, every buffer-tensor index beyond the first would be compared against config index 0 instead of its own config entry, so `typedSecondTensor_n`'s second push (tensor 1 mismatching while tensor 0 matches) would silently pass — this is exactly what the test is built to catch, confirming the mutation-matrix claim without needing a second build.

Confirmed the new typed helpers earn their keep rather than duplicating main's uint8 helpers: main's `_flex_in_fill`/`_flex_in_push` construct only a single uint8 tensor with a one-dimensional shape; the new `_flex_in_typed_append`/`_flex_in_typed_push` are needed for arbitrary `tensor_type` values, multi-dimensional shapes, and (via `_flex_in_typed_register`) multi-tensor models built from comma-separated type/dim/name strings, none of which the existing helpers can express.

## Other checks

- **Style:** `clang-format -style=file` (v16.0.6, repo `.clang-format`) reports no diff on the changed test file. Every new function/struct/test carries a Doxygen `@brief`.
- **Scope:** only `tensor_filter.c`, the one documentation paragraph, and the test file are touched; no other module, and no public API or architecture change, so no further documentation is owed.
- **CI gating:** `unittest_filter_custom` remains a registered `meson test()` target, so a future regression in `gst_tensor_info_is_equal()`, `gst_tensor_dimension_is_equal()`, `gst_tensor_meta_info_convert()`, or the new function itself still fails the build and blocks merge; with the leak fixed, it also no longer trips #4955's now-merged valgrind leak gate.

## Verdict

**OK to merge.** Both findings from the `00883e93` review (the blocking harness/element/bus leak, and the parallel-helper-family nit) are verified fixed by direct reproduction (memcheck + `check_valgrind_log.sh` exits 0 with 0 definite leak contexts from this repository) rather than by trusting the author's report alone. The consolidation from eight tests to five preserves every previously-pinned behaviour, the `tensor_filter.c`/documentation changes are confirmed unchanged from the already-approved content and still do not touch #4953's code, and the full suite passes (24/24) locally. No new Critical/Major/Minor/Nit findings from this round.

PR #4959, opened 2026-09-10
[Filter/Converter/Python3] Validate what a script returns before handing it out

myungjoo:
This fixes items **F6**, **F7**, **F8** and **F11** of #4920: the python3 tensor_filter and tensor_converter sub-plugins trusted every object a user script returned.

Scope of F7: its title also names the decoder, but the lines it cites are all in the filter and converter. The python3 decoder's `decode()` returns `bytes`, not an ndarray (`PyBytes_Size()`/`PyBytes_AsString()`), so a `PyArray_Check`/contiguity check does not apply there. Its missing type check on the returned object is item **F10**, which stays open together with F9 in `tensordec-python3.cc`. All three parts of F11 are covered here.

## What was wrong

**tensor_filter (`tensor_filter_python3.cc`)**
- `invoke()` dropped `run()`'s result, so a failed run counted as a successful invoke. When a later output failed the type/size check, the outputs accepted before it had already been written to `output[]` and then released, so tensor_filter pushed a buffer wrapping freed numpy memory (F6). A script that raised an exception gave a buffer wrapping whatever `output[]` held.
- A result that is not a list, or a list of the wrong length, went to `ml_logf()`, which aborts the process.
- An item that is not an ndarray was read through `PyArray_TYPE()`/`PyArray_DATA()` as if it were one (F7).
- A non-contiguous array (a slice with a step, or a reversed view) was handed out as the address of its first element. The tensor was then read in memory order, and past the array for a negative stride (F7).
- **Not listed in the issue:** an output that is a view of an input tensor (e.g., a passthrough script returning its input) was handed out as a pointer into the input buffer, which tensor_filter releases right after invoke. The pipeline test below receives garbage on main.
- The same data handed out twice (one array for two outputs, or an array the script keeps and returns again while the previous buffer is still in use) collided in `outputArrayMap`: a reference leaked and two buffers shared one memory (F11, duplicate-output part).
- `loadScript()` kept the module and class references on every failure path (F11).

**tensor_converter (`tensor_converter_python3.cc`)**
- The output memory was sized with `PyArray_SIZE()`, an element count, so any dtype wider than a byte published a memory smaller than its caps. The element downstream then read past it (F8).
- A script that raised an exception passed a NULL result to `PyArg_ParseTuple()`, which dereferences it (F8).
- It had no ndarray check and no contiguity handling (F7), and it did not compare the output count with the tensors info.
- A memory that failed to map kept the reference `get_nth_memory()` took (F11).

## What changes
- **Filter:** `run()` validates every output before it writes any of them to `output[]`: the result must be a list of the right length, and each item must be an ndarray with the right type and size. It copies an array that is non-contiguous, points into an input tensor, or has the same data pointer as an output already handed out. Overlapping views of one array and read-only arrays are still handed out without a copy; that is not a lifetime issue (each view keeps its base alive), and the README says an element after tensor_filter must not write into such an output. On any failure it returns an error without touching `output[]`, and `invoke()` throws, which the C++ wrapper turns into `-EINVAL` and tensor_filter into a stream error. The log says which check an output failed (not an array, type, or size). `loadScript()` releases the module and class at its common exit. The Python3 section of `ext/nnstreamer/tensor_filter/README.md`, empty until now, describes what a script has to return.
- **Converter:** it checks for a NULL result, checks that each item is an ndarray holding exactly `gst_tensor_info_get_size()` bytes, requires the output count to match the tensors info, copies with `PyArray_GETCONTIGUOUS()` and `PyArray_NBYTES()`, and drops the memory ref on map failure. `gsttensor_converter.md` now says what `out_array` has to hold.

Behaviour changes, all for scripts that were already broken: tuples of outputs and scripts that raise now stop the stream with an error instead of pushing garbage or aborting. A passthrough-style filter script now costs one copy per output. Converter scripts that return raw `uint8` arrays, like every in-tree script, give byte-identical results; the SSAT suites `nnstreamer_filter_python3`, `nnstreamer_converter_python3` and `nnstreamer_decoder_python3` give the same results before and after.

## Tests
- `tests/test_models/models/filter_output_cases.py`: its custom argument selects what `invoke()` returns. A failing case fails only its first invoke, so each `_n` case can also check that the next invoke of the same instance works, i.e., that no Python error is left pending.
- `tests/test_models/models/converter_output_cases.py`: its first input byte selects what `convert()` returns.
- `unittest_filter_python3`: 13 fixture cases plus `pipelineAliasInput` and `pipelineSecondInvalid_n` through a real tensor_filter pipeline. They check return codes, that `output[]` is untouched after a failure, and each output's bytes and that its address is outside the input. `shortOutput_n` (uint8, one byte short) and `wrongType_n` (int8, right size) pin the size check and the type check separately; the size check is what keeps tensor_filter from reading past a short numpy buffer.
- `unittest_converter`: `outputInt32`, `outputNonContiguous`, `outputUint8`, `outputInvalid_n` (raise / not an array / size mismatch / count mismatch / tuple instead of a list), and `unmappableInput_n` (refcount of a memory mapped for writing).

**Negative control** (new tests against unfixed main, each in its own process): every new case except the two plain baselines (`copy`, `outputUint8`) fails. `notArray_n` and the converter's `outputInvalid_n` SIGSEGV, `notList_n` and `wrongCount_n` abort in `ml_logf()`, and the rest fail their checks.

**Mutation matrix** (remove one guard of the fix at a time, run the case meant to pin it): 21 of the 22 are caught deterministically, including removing only the type check or only the size check, which the whole suite missed before `shortOutput_n`/`wrongType_n` were added (found by the `myungjoo-bot` review). The exception is the filter's `PyArray_Check`: without it a non-array goes through `PyArray_TYPE()`, which is undefined behaviour. Across runs it segfaulted twice and once was rejected by the type check by accident. No deterministic test can pin that without a memory checker. The CI valgrind job does not build the python3 sub-plugins because it has no numpy headers.

Local run (WSL, jammy, numpy 1.21.5 like the CI Pdebuild job, flatbuffers 23.5.26): `meson test` 25/25, the three python3 SSAT groups match main, every commit builds with `-Werror` and passes on its own, clang-format 16 is clean, and `doxygen-tag.sh` is clean (with ctags, and with a negative control). These tests run in CI in the Pdebuild and Tizen/GBS jobs; the Ubuntu meson jobs build without python3.

The aliasing also shows up with the in-tree `passthrough.py`. `videotestsrc num-buffers=2 ! ... ! tensor_converter ! tensor_filter framework=python3 model=passthrough.py input=3:280:40:1 inputtype=uint8 output=3:280:40:1 outputtype=uint8 ! filesink` logs `Cannot find output data: 0x...` on main, because the second frame's pooled input memory had the same address as the first frame's output. The same pipeline is silent on this branch, and both write the same 67200 bytes.

Found while doing this and **not** changed: without `input=`/`output=` on the element, a GETDIM script (`getInputDim`/`getOutputDim`) gets an empty input list. `PYCore::init()` copies `prop->input_meta`/`output_meta` before the framework fills them, so `run()` returns zero outputs as a success. tensor_filter then logs `gst_memory_new_wrapped: assertion 'data != NULL' failed` and pushes empty buffers. This is the same on main and on this branch, measured with `passthrough.py`. It is a functional defect in the meta handling, not in the output validation this PR is about, so it is left for a separate issue. The new pipeline tests set both properties.


myungjoo:
**This review was written by another AI agent (Claude, Sonnet) and is being relayed here on the author's behalf.** It is not a maintainer/human review; please treat findings accordingly.

## Scope of review

Diff: `ext/nnstreamer/tensor_filter/tensor_filter_python3.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_python3.cc`, `gst/nnstreamer/elements/gsttensor_converter.md`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/{converter,filter}_output_cases.py` (3 commits on `myungjoo:fix/4920-python3-script-output`, targeting items F6/F7/F8/F11 of #4920).

Verification performed: read the full pre- and post-PR `PYCore::run()` / `PYCore::loadScript()` / `PYConverterCore::convert()`, traced `outputArrayMap` / `isReferencedData` lifetime and the `TensorFilterPython::invoke()` → `tensor_filter_support_cc.cc` exception→`-EINVAL` path, and built the PR branch in WSL (jammy, numpy 1.21.5, flatbuffers 23.5.26; `-Werror`, no warnings). Ran `unittest_filter_python3` (`nnstreamerFilterPython3*`, 21/21 passed, including `pipelineAliasInput`/`pipelineSecondInvalid_n` through a real `gst_parse_launch` pipeline) and `unittest_converter` (`tensorConverterPython.*`, 14/14 passed, including `outputInt32`/`outputNonContiguous`/`outputInvalid_n`/`unmappableInput_n`). All green, matching the PR body's claims.

## Findings

**1. [Minor] Issue-tracking accuracy — F7 is only partially closed, PR text implies it's fully closed**
Issue #4920's F7 reads: *"Filter/converter/decoder do not check `PyArray_Check` / C-contiguity before using `PyArray_DATA`"* and explicitly cites `tensor_filter_python3.cc`, `checkTensorSize`, and `tensor_converter_python3.cc`. This PR fixes the filter and converter halves correctly, but `ext/nnstreamer/tensor_decoder/tensordec-python3.cc` is untouched by this PR (confirmed: no `PyArray_Check`/`PyArray_IS_C_CONTIGUOUS` calls exist there before or after this branch), so the decoder still hands out `PyArray_DATA` from an unchecked/possibly non-contiguous object the same way the filter did before this fix. The PR body says "This fixes items F6, F7, F8 and F11 of #4920" without qualifying that F7's decoder half remains open. Recommend rewording the PR body (and the checkbox this PR will cause to be ticked on #4920) to say F7 is only closed for the filter/converter, so the decoder half doesn't silently fall off the tracking issue. Not a code defect in this diff, just a completeness/tracking accuracy issue.

**2. [Nit] Error message conflates three different failure reasons**
`ext/nnstreamer/tensor_filter/tensor_filter_python3.cc:619-621`:
```cpp
if (!PyArray_Check (item) || !checkTensorType (_info->type, PyArray_TYPE (output_array))
    || !checkTensorSize (&output[i], output_array)) {
  ml_loge ("Output tensor type/size is not matched\n");
```
"item is not an ndarray at all", "dtype mismatch", and "size mismatch" all print the identical "type/size is not matched" message (confirmed by running `notArray_n`, which logs this same string as `secondInvalid_n`'s genuine size mismatch). A script author debugging why their filter script fails will not be able to tell these apart from the log. Suggest splitting into distinct `ml_loge` calls (or including which of the three checks failed) for a better debugging experience — purely cosmetic, no functional risk.

**3. [Nit] No filter-side doc update parallel to the converter's**
`gsttensor_converter.md` was correctly updated to describe the new `out_array` byte-size contract (`gst/nnstreamer/elements/gsttensor_converter.md:157`). The filter side underwent an equivalent (and arguably more user-visible) contract change — `CustomFilter.invoke()` returning a non-list, wrong-length list, wrong dtype/size item, or an exception now fails the pipeline with a stream error instead of previously pushing garbage/aborting the process; aliased or non-contiguous outputs are now silently copied (one extra `memcpy` per such output). There's no existing markdown file describing the `CustomFilter` contract to extend (`ext/nnstreamer/tensor_filter/README.md`'s "Python3" section is just a bare header today), so this is a pre-existing documentation gap rather than something this PR strictly regresses — but since the PR already pays down the equivalent converter-side doc debt, doing the same for the filter side (even a couple of lines) would be a natural place to land it now that the contract has been made precise.

## What I checked and found correct (no action needed)

- `PYCore::run()`: outputs are only written to `output[]`/`outputArrayMap` after *every* item in the result list passes type/size/contiguity/non-aliasing checks (`res == 0` gate) — a failure midway leaves `output[]` exactly as the caller set it (verified via `expectNoOutput`/pipeline tests), so the F6 use-after-free (freed numpy buffer wrapped into the GStreamer buffer) is closed. `arrays` accumulated so far are correctly `Py_SAFEDECREF`'d on the failure path, no leak.
- `isReferencedData()` correctly de-dupes against `outputArrayMap` (already handed out and not yet released), the current invoke's own `arrays`, and the input tensor ranges (`input[i].data .. +size`), triggering a defensive `PyArray_NewCopy` — this also fixes the "passthrough script returns its input" UAF the PR calls out as found-but-not-in-the-original-issue-list; reproduced the fix with `pipelineAliasInput` locally.
- `loadScript()`'s single `exit:` cleanup correctly releases `module`/`cls` on every failure path (constructor-raises, class-missing, args-splits-to-nothing) where the old code leaked one ref each; `openInitError_n` (refcount-diff across two failed opens) passed locally.
- `PYConverterCore::convert()`: `PyArray_NBYTES` (not `PyArray_SIZE`) is compared against `gst_tensor_info_get_size()`, closing the F8 undersized-buffer bug for non-byte dtypes; the `gst_memory_unref(in_mem[i])` added at the map-failure site correctly balances the ref taken by `gst_tensor_buffer_get_nth_memory()` without double-freeing the memories already mapped before it (verified via `unmappableInput_n`'s refcount assertion).
- Diff is proportionate and contained: only the two sub-plugin `.cc` files, one doc file, and their own test files are touched; no other module/subplugin is modified. Test coverage (both positive and `_n` negative cases, plus two full-pipeline tests) matches CLAUDE.md's test-case policy and should catch a future regression in this exact code path (e.g., someone "simplifying" `isReferencedData` back out). These tests build/run only under `-Dpython3-support=enabled` (Pdebuild/Tizen-GBS CI), consistent with the PR's own description of CI coverage; a real failure in this path (wrong type/size, non-contiguous, aliased, or an exception) would fail these gtests and block merge in those jobs.

## Verdict

**Mergeable after the nits.** No blocking or major issues found; findings #2 and #3 are cosmetic/documentation nits and #1 is a tracking-accuracy note for the PR description rather than the code.

myungjoo:
Thanks for the review. All three points are addressed in the force-pushed head `acd09ef8`, folded into the first commit, so the branch stays at three commits:

1. **F7 scope** — agreed that the description overclaimed. But the decoder half is not an ndarray check: `PYDecoderCore::decode()` takes the script's result as `bytes` (`PyBytes_Size()`/`PyBytes_AsString()`), and `PyArray_Check`/C-contiguity does not apply there. The decoder's missing type check on the returned object is item F10, which stays open with F9 in `tensordec-python3.cc`. The PR body now says exactly that.
2. **Log message** — a non-array output now logs `Output tensor %u is not a numpy array`, separately from the type/size mismatch.
3. **Filter documentation** — the empty Python3 section of `ext/nnstreamer/tensor_filter/README.md` now describes the `CustomFilter` contract: the callbacks, what `invoke()` gets and returns, which outputs are copied, and that anything else stops the stream with an error.

Checked on `acd09ef8`: `unittest_filter_python3` 21/21 and `unittest_converter` 14/14. The first commit also builds with `-Werror` and passes (20/20, 9/9) on its own. clang-format 16 is clean, and the content delta against the reviewed head is only these two files (+12/−1).

myungjoo:
**This review was written by another AI agent (Claude, Sonnet) and is being relayed here on the author's behalf. It is not a maintainer/human review; please treat findings accordingly. Round 2, reviewing head `acd09ef8`.**

## Scope of review

Same diff footprint as round 1, now with `ext/nnstreamer/tensor_filter/README.md` added (8 files total): `ext/nnstreamer/tensor_filter/tensor_filter_python3.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_python3.cc`, `ext/nnstreamer/tensor_filter/README.md`, `gst/nnstreamer/elements/gsttensor_converter.md`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/{converter,filter}_output_cases.py`.

Verification performed: re-read `PYCore::run()`/`loadScript()`/`isReferencedData()`, `PYConverterCore::convert()`, and `PYDecoderCore::decode()` (to check the F7/F10 scoping claim) at `acd09ef8`; diffed against the previously-reviewed `9af09474`; traced reference-counting and `goto`/cleanup paths on every branch by hand. Built the branch fresh in WSL (jammy-equivalent, numpy 1.21.5, flatbuffers 23.5.26, `meson setup -Denable-test=true -Dpython3-support=enabled -Dflatbuf-support=enabled`); build is clean. Ran `unittest_filter_python3` (`nnstreamerFilterPython3*`): **21/21 passed**, including `pipelineAliasInput` (EOS, byte-identical passthrough) and `pipelineSecondInvalid_n` (stream `ERROR`, no data received). Ran `unittest_converter` (`tensorConverterPython.*`): **14/14 passed**, including `outputInt32`, `outputNonContiguous`, `outputInvalid_n` (raise/not-array/size-mismatch/count-mismatch), `unmappableInput_n`. `clang-format` (repo style file) reports no diff on the four touched C/C++ files.

## Resolution of the three round-1 points

1. **F7 decoder scope** — Resolved. The PR body now states F7 is closed only for the filter/converter, and attributes the decoder's missing type check to F10. I independently checked `tensordec-python3.cc`'s `decode()`: it calls `PyBytes_Size(output)` / reads the result as `bytes`, never `PyArray_Check`/`PyArray_DATA`, so "C-contiguity of an ndarray" genuinely does not apply there — the claim is accurate. I also confirmed issue #4920's F7 entry cites only `tensor_filter_python3.cc` and `tensor_converter_python3.cc` line ranges, not the decoder file, despite the title mentioning "decoder" — so the narrower framing matches the tracking issue's own text. (Aside, not a defect in this PR: `PyBytes_Size()`'s return is assigned straight into a `size_t mem_size` at `tensordec-python3.cc:165` with no `-1`/error check, so a non-bytes `decode()` return there would compute a huge allocation size — this is exactly the still-open F10, left out of this PR's scope as stated.)
2. **Log message split** — Partially resolved. `tensor_filter_python3.cc:619` now gives "Output tensor %u is not a numpy array" its own message, separate from the type/size branch. However, `tensor_filter_python3.cc:627` still prints one shared "Output tensor type/size is not matched" for *both* a dtype mismatch and a size mismatch (`checkTensorType` vs `checkTensorSize` are still OR'd into a single `ml_loge`). The original nit listed three causes; one of the three is now distinguishable, the other two remain combined. Not a functional issue — Minor/nit, optional.
3. **Filter documentation** — Resolved. `ext/nnstreamer/tensor_filter/README.md`'s Python3 section now documents the `CustomFilter` contract (constructor, `invoke()` input/output shape, copy-on-alias/non-contiguous/duplicate, error-on-anything-else). I checked it against the actual code path in `PYCore::run()` (type+size check, `PyArray_IS_C_CONTIGUOUS`, `isReferencedData()`, `PyList_Check`/exception → stream error) — the description is accurate to the implementation at `acd09ef8`.

## New findings (round 2)

**1. [Nit] `tensor_filter_python3.cc:627`** — see point 2 above: dtype-mismatch and size-mismatch still share one log message ("Output tensor type/size is not matched"). Suggest splitting `checkTensorType`/`checkTensorSize` into separate `ml_loge` calls if further debuggability is wanted; purely cosmetic.

**2. [Nit] Test-coverage gap: converter's "`convert()` returns a non-list `output`" branch is unexercised** — `tensor_converter_python3.cc:184` (`if (num_tensors < 0 || ...)`) correctly turns a non-list `output` into an error via `PyList_Size()`'s `-1` sentinel (verified `Py_ERRMSG` at :185 clears the resulting `TypeError` before `goto done`, so no pending-exception leak). But `tests/test_models/models/converter_output_cases.py:27-35`'s `cases` dict has no entry that returns e.g. `output` as `bytes`/`None`/a scalar for the whole second tuple element — only individual *items inside* the list are tested as non-arrays (mode 4). The filter side has the matching test (`not_list` case, `tensor_filter_python3.cc`'s `!PyList_Check(result)` branch, pinned by `notList_n`); the converter's analogous branch is logically sound (I read it) but not pinned by a regression test. Low severity since the branch is simple and unlikely to be touched incidentally, but per repo policy (regression tests for every guarded path) it's a natural next addition — e.g. add a mode returning a `bytes` object as `output` itself.

## What I re-checked and found correct (carried over from round 1, still true at `acd09ef8`)

- `PYCore::run()`: no output is written to `output[]`/`outputArrayMap` until every item in the result passes type/size/contiguity/non-aliasing checks; a mid-list failure leaves `output[]` untouched (F6 closed) and correctly `Py_SAFEDECREF`s every array already pushed into the local `arrays` vector (no leak, no use of an uninitialized `arrays[i]`).
- `isReferencedData()` correctly rejects data already in `outputArrayMap`, already claimed within the same invoke (`arrays`), or inside any input tensor's byte range, forcing a defensive `PyArray_NewCopy`; `PyArray_NewCopy` failure is itself handled (`res = -ENOMEM`, no null deref).
- `loadScript()`'s single `exit:` label now unconditionally releases `module`/`cls` (both declared/initialized before any early `goto exit`), closing the loadScript leak part of F11 on every path, not only the success path.
- `PYConverterCore::convert()`: the map-failure branch's added `gst_memory_unref (in_mem[i])` (`tensor_converter_python3.cc:9` of the diff) is exactly balanced against `done:`'s loop, which only unmaps/unrefs indices `< num` (and `num` was set to `i` before the `goto`), so there is no double-unref of the memory that failed to map and no leak of the ones that succeeded before it. Verified live via `unmappableInput_n`'s refcount assertion (passed).
- `PyArray_NBYTES` (not `PyArray_SIZE`) is what is compared against `gst_tensor_info_get_size()` and used for `mem_size`/`_g_memdup`, closing F8 for non-byte dtypes (`outputInt32` passed with 16 bytes, not 4).
- Diff stays proportionate and contained to the two python3 sub-plugins, their own docs, and their own tests; nothing in another module or sub-plugin is touched. No behavior change for well-formed in-tree scripts (`outputUint8`/`copy`/pipeline `passthrough.py` all byte-identical, confirmed via the local pipeline test).
- These tests build/run only under `-Dpython3-support=enabled` (per the stated CI facts: Pdebuild jammy/resolute and Tizen/GBS), so a future regression in this exact code path would fail those CI jobs and block merge there; the Ubuntu meson/valgrind jobs do not exercise this code either way, consistent with the PR body's own caveat.
- No architecture/API change; the two doc files (`README.md`, `gsttensor_converter.md`) that describe the affected script contracts are updated in this same PR, and I checked their text against the implementation (see "Resolution" #3 above and the converter doc line matching the `PyArray_NBYTES`-vs-`gst_tensor_info_get_size` check).

## Verdict

**Mergeable as is.** No blocking or major issues. The two new findings are both Nits (cosmetic log-message granularity; a test-coverage completeness suggestion for an already-correct branch), and all three round-1 points are resolved or accurately re-scoped in the PR body/code at `acd09ef8`.

myungjoo:
Thanks. Both round-2 nits are closed at `ff00d087`, folded into their own commits so the branch is still three commits:

1. **Type vs. size message:** the three checks now report which one failed: `Output tensor %u is not a numpy array` / `... type is not matched` / `... size is not matched`. One error block, so the code did not grow.
2. **Converter non-list output:** `converter_output_cases.py` mode 7 returns the arrays as a tuple, and `outputInvalid_n` now covers modes 3–7.

Checked on `ff00d087`: `unittest_filter_python3` 21/21 and `unittest_converter` 14/14, and the first and second commits each build with `-Werror` and pass on their own. clang-format 16 is clean. The mutation matrix (remove one guard at a time) has all 20 caught on this head. The `PyArray_Check` one crashed this time, but as the PR body says, that one is undefined behaviour and not deterministic. Delta against the reviewed `acd09ef8`: 3 files, +13/−11.

myungjoo:
**This review was written by another AI agent (Claude, Sonnet) and is being relayed here on the author's behalf. It is not a maintainer/human review; please treat findings accordingly. Round 3, reviewing head `ff00d087`.**

## Scope of review

Same 8-file footprint as round 2: `ext/nnstreamer/tensor_filter/tensor_filter_python3.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_python3.cc`, `ext/nnstreamer/tensor_filter/README.md`, `gst/nnstreamer/elements/gsttensor_converter.md`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/{converter,filter}_output_cases.py` — no other module touched (confirmed via `gh pr view --json files`).

Verification performed: fetched `pr4959-r3` locally, diffed `acd09ef8...ff00d087` directly (3 files, +13/−11, matches the PR body's stated delta), and re-read the full `PYCore::run()`/`loadScript()`/`isReferencedData()` and `PYConverterCore::convert()` at `ff00d087`. Independently rebuilt the branch from scratch in my own WSL tree (jammy-equivalent, numpy 1.21.5, flatbuffers 23.5.26, `meson setup -Denable-test=true -Dpython3-support=enabled -Dflatbuf-support=enabled`) — clean build, no warnings. Ran both suites fresh: `unittest_filter_python3` **21/21 passed** (including `pipelineAliasInput`/`pipelineSecondInvalid_n` through a real `gst_parse_launch` pipeline, and `openInitError_n`'s refcount check), `unittest_converter` (`tensorConverterPython.*`) **14/14 passed** (including the new `outputInvalid_n` mode 7). `clang-format -n` (repo style, local clang-format 14 — CI/prior rounds used 16, noted as a version-skew caveat, not a defect) reports no diff on the touched C/C++ files. Also read `scaler.py`/`custom_converter.py` (in-tree filter/converter scripts) against the new checks: `scaler.py`'s output is a fresh `np.empty` + `ravel()` (contiguous, non-aliased) so it takes the zero-copy path; the converter always `_g_memdup()`s regardless of aliasing, so no in-tree converter script can regress on that axis. I did not re-run the SSAT suites myself (no SSAT harness set up in my tree); I relied on the PR body's claim that `nnstreamer_filter_python3`/`nnstreamer_converter_python3`/`nnstreamer_decoder_python3` are unchanged, which is consistent with everything I traced in the code.

## Resolution of the two round-2 nits

1. **`tensor_filter_python3.cc:619-627` shared type/size message** — **Resolved.** The three checks (`PyArray_Check`, `checkTensorType`, `checkTensorSize`) are now an if/else-if chain setting a `mismatch` string, logged as `"Output tensor %u %s\n"`. Verified this is not just a code change but an observable behavior change: my local run shows `Output tensor 1 type is not matched` for `secondInvalid_n` (dtype mismatch) and `Output tensor 0 is not a numpy array` for `notArray_n` — previously both printed the identical `"Output tensor type/size is not matched"`. The refactor is behavior-preserving for the pass/fail outcome (same overall condition triggers `res = -2`), only the diagnostic text changed, so this closes cleanly with no functional risk.
2. **Converter's non-list `output` branch untested** — **Resolved.** `converter_output_cases.py` mode 7 now returns `(seq[:4].copy(),)` — a tuple instead of a list — and `outputInvalid_n`'s `modes[]` now runs `{3,4,5,6,7}`. Traced the code path: `PyList_Size()` on a tuple sets a `TypeError` and returns `-1`, which is caught by the existing `num_tensors < 0` check at `tensor_converter_python3.cc:184`. Confirmed passing locally as part of `outputInvalid_n`.

## New findings (round 3)

None. The round-3 delta is small and precisely targeted at the two round-2 nits; I did not find any new defect, leak, or behavior regression introduced by it. Everything verified correct in round 2 (F6 use-after-free fix, F7 contiguity/type checks, F8 `PyArray_NBYTES` sizing, F11 `loadScript`/map-failure/duplicate-output leaks, the exception→`-EINVAL`→stream-error path, and the doc updates matching the implementation) is untouched by this delta and still holds at `ff00d087`.

## What I independently re-checked and found correct at `ff00d087`

- Diff stays proportionate and fully contained to the two python3 sub-plugins, their own two doc files, and their own tests — nothing in another module or sub-plugin is touched.
- No architecture/API change; no public header or ABI touched. The two doc files already describe the contract this PR changes (`CustomFilter.invoke()` return shape in `README.md`, `out_array` byte-size contract in `gsttensor_converter.md`), consistent with them being updated in this same PR.
- Test coverage (positive + `_n` negative cases, two full-pipeline tests, a refcount test for `loadScript`, a refcount test for map-failure) is proportionate to the fix and would catch a future regression in this exact code path in CI (Pdebuild/Tizen-GBS, where `-Dpython3-support=enabled`), per the repo's own stated CI coverage.

## Verdict

**Mergeable as is.** Both round-2 nits are resolved at `ff00d087` (log-message granularity and converter test coverage), no new issues found in the round-3 delta, and a fresh independent build + full test run reproduces the PR's claimed 21/21 and 14/14 results.

myungjoo:
Thanks for the review. Both suggestions are applied at `7df38023`. They are folded into the first commit (filter), so the branch is still three commits.

1. **Type check and size check pinned separately:** `filter_output_cases.py` gained `short` (uint8, one byte short) and `int8` (right size, wrong type) as fail-once cases, with `shortOutput_n` and `wrongType_n` next to `secondInvalid_n`. Mutation check on this head: removing only the type check fails `wrongType_n`, removing only the size check fails `shortOutput_n`, and the whole `unittest_filter_python3` suite now fails for each of the two mutations. Both new cases also fail against the unfixed plugin, where `invoke()` ignored `run()`'s result.
2. **README wording:** I took the narrower option rather than extending the copy condition. The README now says an output is copied only when it is non-contiguous, lies in an input tensor, or has the same data pointer as an output already given out. It also says read-only arrays and views overlapping another output are pushed without a copy, so an element after tensor_filter must not write into such an output. The comment above the copy condition no longer says "owned by this array alone". The narrower option keeps the zero-copy path for read-only arrays and adds no per-invoke range scan over `outputArrayMap`, and, as you noted, there is no lifetime issue and no in-tree element writes in place (`transform_ip` is used only by `tensor_debug` and `tensor_rate`, neither of which writes data).

Checked on `7df38023`:
- `unittest_filter_python3` passes 23/23 and `unittest_converter` 14/14.
- The first commit alone builds with `-Werror` and passes 22/22.
- clang-format 16 and `doxygen-tag.sh` (with ctags and a negative control) are clean.
- The python3 SSAT groups give the same results as before.
- Delta against `ff00d087` is 4 files, +27/−3.

myungjoo:
**This review was written by another AI agent (Claude, Sonnet) and is being relayed here on the author's behalf. It is not a maintainer/human review; please treat findings accordingly. Round 4, reviewing head `7df38023`.**

## Scope of review

Same 8-file footprint as rounds 2-3, still nothing outside it: `ext/nnstreamer/tensor_filter/tensor_filter_python3.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_python3.cc`, `ext/nnstreamer/tensor_filter/README.md`, `gst/nnstreamer/elements/gsttensor_converter.md`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/{converter,filter}_output_cases.py`. Round-4 delta vs. `ff00d087` is exactly what the author's reply claims: 4 files, +27/-3, entirely inside the first commit (`691040bd`) — `c7273be3` and `7df38023` (renamed from `4dfef4a1`/`ff00d087` by the force-push) are untouched by this delta.

Verification performed: fetched the PR head locally (`git fetch origin pull/4959/head`), diffed `ff00d087...7df38023` directly, and read the two touched source files in full at the new head. Ran `clang-format` (repo's `.clang-format`, clang-format 16.0.6) over both touched `.cc` files — zero diff, both clean. Checked all 3 commits for DCO `Signed-off-by` (one each, no duplicates) and confirmed the round-4 changes sit entirely in commit 1's diff. Independently re-derived, by hand, the byte sizes/dtypes of the two new fixture cases against the `checkTensorType`/`checkTensorSize` `else if` chain (see below) rather than trusting the PR body's numbers. Grepped the whole tree for `transform_ip` to check the README's zero-copy claim (see point 2). I could not get a WSL shell to respond in this session (both `build_py3f.sh`-style rebuild and a bare `wsl.exe -d Ubuntu -- echo hello` timed out with no output), so unlike rounds 1-3 I did not independently rebuild/run the gtests this round; the verification below is static (code reading + hand-tracing), not a fresh test run. I also checked `gh pr checks 4959`: 19 total, most green, but at review time the two CI jobs that actually build/run these python3 gtests — **Ubuntu pdebuild on jammy (amd64)** and **Tizen GBS build on Ubuntu (x86_64, `unit_test 1`)** — were still `pending`, alongside pdebuild-resolute/Valgrind/RISC-V/Yocto/Android-NDK. Worth confirming they finish green before merge, since round 4 was pushed after the last full CI run.

## Resolution of the two `myungjoo-bot` points (raised on `ff00d087`, 2026-09-11)

**1. Type check and size check pinned by cases that each fail only one of them — Resolved.**

`ext/nnstreamer/tensor_filter/tensor_filter_python3.cc:619-624` is unchanged logic (the `else if` chain from round 3); what changed is `tests/test_models/models/filter_output_cases.py`, which gained:
```python
'short': lambda: [data[:4].copy(), data[4:7].copy()],              # uint8, 3 bytes (expects 4)
'int8':  lambda: [data[:4].copy(), data[4:].copy().view(np.int8)], # int8, 4 bytes (expects uint8)
```
`data` is uint8 and both output tensors are declared `TensorShape([4], np.uint8)` (4 bytes each). Tracing the chain: `short`'s second output keeps dtype uint8 (so `checkTensorType` passes) but is 3 bytes (so `checkTensorSize` fails) — this pins the size check with the type check never in play. `int8`'s second output is a 4-byte `int8` view (`checkTensorSize` would pass — 4 == 4) but dtype mismatches (`checkTensorType` fails first in the `else if` chain) — this pins the type check with the size check never reached. This is exactly the pairing the bot asked for (one case where only the type guard is load-bearing, one where only the size guard is), and it matches the new `shortOutput_n`/`wrongType_n` `TEST_F`s in `unittest_filter_python3.cc`, which are formatted identically to their neighbors (each with a `@brief` Doxygen comment, `openCase` → `invoke` → `expectNoOutput` → `expectRecovered`, consistent with the existing `secondInvalid_n`/`notArray_n` pattern). I did not re-run the mutation matrix myself this round (WSL unavailable), but the case construction is sound by inspection and matches the PR body's/bot's worked example almost verbatim.

**2. README promised copies for reused data, but the code only compares exact pointers — Resolved, and the choice (narrow the docs rather than extend the code) is reasonable here.**

The one-line code comment changed (`tensor_filter_python3.cc:632`, "owned by this array alone" → "must not be given out already") and the README paragraph (`README.md:20`) was reworded to: copy only when non-contiguous, in an input tensor, or same data pointer as an already-given-out output; otherwise pushed as-is, "including a read-only array and a view overlapping another output," with a note that a downstream element must not write into such an output. I checked this against `isReferencedData()` (`tensor_filter_python3.cc` around line 407-420, unchanged this round): it checks `outputArrayMap` membership, membership in the current invoke's `arrays`, and containment inside an input tensor's `[data, data+size)` range — all exact-pointer/range tests, no overlap scan, no `PyArray_ISWRITEABLE` check. The new wording is now an accurate description of this function, not an overpromise.

On whether narrowing the doc (option A) rather than extending the copy condition (option B, which the bot also offered) is acceptable for a PR whose whole point is memory safety: yes, and I verified the load-bearing claim independently rather than taking the reply's word for it. The bot's argument was that overlap/read-only aliasing is a *lifetime*-safe, *writability*-only concern (every numpy view keeps its base object alive via refcounting, so there is no UAF), and that no in-tree element writes into a tensor_filter output in place. I re-checked the second half myself: `grep -rn "transform_ip" gst/nnstreamer/elements/` finds three hits, but only `gsttensor_debug.c` and `gsttensor_rate.c` actually register a `trans_class->transform_ip` vmethod; `gsttensor_decoder.c`'s hit is `transform_ip_on_passthrough = FALSE`, a `GstBaseTransform` flag, not an in-place-write implementation — it uses `transform` (copying), not `transform_ip`. Reading the two real implementations: `gst_tensor_debug_transform_ip()` only calls a still-unimplemented `@todo NYI` logging stub, and `gst_tensor_rate_transform_ip()` only touches segment/timestamp bookkeeping — neither maps or writes the tensor payload bytes. So the "no in-tree consumer writes in place" premise holds. Given that, and that this was already filed as a Nit (not Low/Major), spending a diff on a per-invoke range scan over `outputArrayMap`/`arrays` plus a writability check to guard against a hazard nothing in-tree can trigger would be disproportionate to the finding — the narrower doc fix is the right-sized change, consistent with CLAUDE.md's "don't grow the diff beyond the topic" expectation. The one thing I'd flag (see Nit below) is that this leaves the constraint ("must not write into such an output") enforced only by convention/documentation, not by any assertion — worth knowing if a future in-tree element ever grows a `transform_ip` that touches tensor data.

## New findings (round 4)

**1. [Minor] Two of the CI jobs that actually exercise this code were still pending at review time.** Per the PR's own stated CI coverage (Pdebuild + Tizen/GBS build/run the python3 sub-plugins; the Ubuntu meson/Valgrind jobs do not, no numpy headers), `Ubuntu pdebuild on jammy (amd64)` and `Tizen GBS build on Ubuntu (x86_64, "unit_test 1")` were `pending` when I checked `gh pr checks 4959`, since the round-4 push (`7df38023`, force-pushed 2026-09-14T08:59:56Z) is more recent than the last full green run this thread has been citing. Not a code defect — just confirm these two finish green (along with `pdebuild-resolute`) before merging, since they're the ones that would actually catch a regression in `shortOutput_n`/`wrongType_n`.

**2. [Nit] The "must not write into such an output" contract from finding 2 above has no code-side guard.** Purely a documentation/convention constraint now (see reasoning above); fine as accepted for this PR, but if a future in-tree filter/decoder ever implements a real `transform_ip` that writes tensor bytes, nothing here will catch it writing into an aliased/read-only tensor_filter python3 output. Optional follow-up only, not blocking this PR.

No other new issues. The round-4 delta does not touch any of the logic independently re-verified correct in rounds 1-3 (F6 UAF fix, F7 filter/converter contiguity+type checks, F8 `PyArray_NBYTES` sizing, F11 `loadScript`/map-failure/duplicate-output leak fixes, exception→`-EINVAL`→stream-error propagation) — it only adds two test fixtures/cases and rewords one comment and one README paragraph.

## Verdict

**Mergeable as is**, once the still-pending Pdebuild/Tizen-GBS CI checks on `7df38023` come back green (finding 1 — this is a status check, not a code issue). Both `myungjoo-bot` points from the `ff00d087` review are genuinely resolved: the type and size checks are now each pinned by a case that fails only that one check, and the README's aliasing/copy wording now accurately matches `isReferencedData()`'s exact-pointer-only semantics — narrowing the docs instead of extending the copy condition is a proportionate choice given this is a writability contract with no lifetime risk and no in-tree writer to violate it (verified independently via `transform_ip` usage, not just by re-stating the reply). Finding 2 is an optional nit for a future follow-up, not for this PR.

PR #4961, opened 2026-09-11
[CI/Valgrind] Fail a run that memcheck never finished

myungjoo:
The valgrind gate (`ubuntu_clean_meson_build.yml`, from #4936) runs `tools/debugging/check_valgrind_log.sh --require-output`. That option exists to fail "a log that lost its memcheck output". But it only checked that *some* binary was examined, and a binary counted as examined on the `==PID== Command:` line of valgrind's banner. **Valgrind prints that line before it can abort at startup.**

## The fail-open, reproduced with a real log

Valgrind 3.18.1 without glibc debug info, run with the same flags `run_unittests_binaries.sh` uses, `2>&1` as the CI step does:

```
==1242== Command: /bin/true
...
valgrind:  Fatal error at startup: a function redirection
valgrind:  which is mandatory for this platform-tool combination
...
valgrind:  Cannot continue -- exiting now.  Sorry.
```

On `main`, the checker's verdict for that log is `0 error contexts from this repository, 0 from libraries, 0 leak reports, 0 definite leak contexts from this repository, 1 binaries examined`, **exit 0**. Not one test ran under memcheck and the gate passes. The job installs `libc6-dbg`, so nothing slips through today. But the day that install stops working (a package rename, a runner image change, an apt mirror problem), the gate would pass every PR.

## The fix

Under `--require-output`, a run now has to reach its `ERROR SUMMARY`; one that doesn't fails the check, and the message names it:

```
Memcheck never finished these runs, so nothing they did was checked:
  [true] no ERROR SUMMARY: valgrind aborted at startup, was killed, or the log was cut
```

- **Runs are matched by process id.** A forked child prints its own summary under its own id and without a `Command:` line, and real runs do this: the two `EXPECT_DEATH` cases in `unittest_plugins` leave two such summaries in the CI log. So a child's summary must not finish its parent's run. Each new run also forgets the ids before it, so a later process that reuses the id of a run that stopped short cannot finish that run.
- **It keys on the summary rather than the abort message**, so a run killed partway, a run whose suppression file could not be opened (`FATAL: can't open suppressions file`, a second fail-open of the same shape; the runner reaches that file through a relative path), and a log cut short fail the same way, and a different abort wording can't slip past.
- **Without `--require-output` nothing changes**, and normal logs parse exactly as before. The new rule deliberately has no `next`, so the summary line still reaches the rule it always reached.
- **With #4955 merged**, the checker fails for an error of ours, a definite leak of ours, or an unfinished run: `exit (unfinished > 0 || leak_ours_count > 0)` on the line where errors of ours did not already decide it.

## Evidence

Under both **mawk 1.3.4** (the awk on the Ubuntu runners; a gawk-only pass once hid a fail-open in this script) and **gawk 5.1.0**:

| input | `main` | this PR |
| --- | --- | --- |
| real startup-fatal log (above) | exit 0 | **exit 1**, names the binary |
| real log of a run whose suppression file is missing | exit 0 | **exit 1**, names the binary |
| real x86 valgrind job log of a green run (#4948 @ `7505ff26`; 23 binaries, every one reaching its summary, plus 2 child summaries) | exit 0 | exit 0, no false positive |
| this PR's x86 valgrind job (`6091fb20`, after the rebase: 23 + 1 llama.cpp runs, every one reaching its summary, plus the same 2 child summaries) | exit 0 | exit 0, gate green |
| the self-test's new negative cases | 9 FAIL | pass |

**How real processes end under memcheck** (valgrind 3.18.1, CI's flags, a small C program per case, checked by both checkers):

| the tested process... | `main` | this PR |
| --- | --- | --- |
| exits normally; raises `SIGSEGV`; `SIGKILL`s itself; is sent `SIGTERM` | exit 0 | exit 0 |
| is `SIGKILL`ed from outside | exit 0 | **exit 1** |
| forks a child that `abort()`s (gtest's default death-test style), or `_exit()`s | exit 0 | exit 0 |
| forks a child that is `SIGKILL`ed, that `exec`s another program, or that outlives it; `posix_spawn`s | exit 0 | exit 0 |
| `exec`s another program itself | exit 0 | **exit 1** |

A `SIGKILL` from outside (a step timeout, the OOM killer) takes memcheck down with the process, so that run was not checked. The last row fails on purpose too: once the tested process has replaced itself, memcheck never reaches its exit and its leak check, so the run was not checked either. No test binary in the tree `exec`s (`EXPECT_DEATH` uses the fork-only style), and the gtest runner does not pass `--trace-children`. Forgetting earlier ids on each new run relies on that: runs follow one another, and a traced child that `exec`s would print its own `Command:` line inside its parent's run and fail the parent as unfinished (a false failure, never a false pass). The script header says so.

New self-test cases in `tools/debugging/test_check_valgrind_log.sh`, which the Static checks job already runs:

- `startup_fatal_required_n`: a trimmed copy of the real log above.
- `suppressions_missing_required_n`: a trimmed real log of a suppression file valgrind could not open.
- `one_of_two_unfinished_n`: only the unfinished run is named.
- `child_summary_only_n`: a child's summary does not finish its parent.
- `reused_pid_n`: a later process reusing the id of a run that stopped short does not finish it.
- Preservation cases that pass on both sides: `startup_fatal_unrequired` (no option, no judgement), `child_summary_extra`, `timestamped_required`.
- `timestamped_required` carries two different timestamps, as real lines do, so taking the process id before the timestamp is stripped fails it.
- The existing `examined_required` case gains the `ERROR SUMMARY` line that every real finished run has.
- `quiet_required` still pins that an empty log fails.

Rebased on `main` after #4955. The only conflict was that exit line; all of #4955's self-test cases still pass (113 checks in total under both awks), and dropping either half of the combined condition fails existing cases (7 without the leak half, 5 without the unfinished half).


myungjoo:

This review was written by a separate Claude agent, not by the PR author or a human maintainer, and is relayed here as-is. It covers head `6091fb20` (one commit on `main` `0e0ccf19`). Only the two changed files are in scope.

## What was checked, and how

- **Does it fix the problem?** Yes. I ran the new self-test against `main`'s checker under **mawk 1.3.4** and **gawk 5.1.0**. It fails the same 5 checks the PR body lists. With this PR's checker, all 108 checks pass under both awks.
- **Real valgrind 3.18.1**, with the flags `run_unittests_binaries.sh` uses (`-v --track-origins=yes --num-callers=200 --leak-check=full`), and `main` vs. this PR under both awks:

  | the tested process / log | `main` | this PR |
  | --- | --- | --- |
  | no glibc debug info (`Fatal error at startup: a function redirection`) | exit 0 | **exit 1**, names the binary |
  | `SIGKILL`ed from outside (no summary written) | exit 0 | **exit 1** |
  | suppression file missing (`FATAL: can't open suppressions file ...`) | exit 0 | **exit 1** |
  | `SIGTERM`ed from outside (valgrind still writes its summary) | exit 0 | exit 0 |
  | `posix_spawn` of a child that execs | exit 0 | exit 0 |
  | forked child that outlives its parent (its summary comes after the parent's) | exit 0 | exit 0 |
  | main thread `pthread_exit`, worker calls `exit()` | exit 0 | exit 0 |
  | last stdout write has no newline (`2>&1 \| cat`) | exit 0 | exit 0 |

  The missing-suppression-file row is a second fail-open that this change also closes. It matters because the runner finds the file through a relative path (`../tools/debugging/valgrind_suppression`, after `pushd build`).
- **No false positives on real green logs.** I split the x86 valgrind job logs of three recent green runs (jobs 104228847800, 104226833737, 103918379475) the way CI splits them: 23 binaries in one log, llama.cpp in the other. Under `-v`, valgrind prints `ERROR SUMMARY` twice per run, and `unittest_plugins` death-test children print their own summaries. Both checkers exit 0 on all 6 logs under both awks, with identical counts. The x86 valgrind job for this head is green too, and its gate step reports `23 binaries examined` and `1 binaries examined` with no unfinished runs.
- **Do the tests catch mutations?** Each result held under both awks:
  - Dropping the `require_output` guard in END: 19 FAIL.
  - Dropping the unfinished half of the exit: 3 FAIL.
  - Dropping the leak half: 7 FAIL.
  - Any summary finishes the latest run, ignoring pid: 1 FAIL (`child_summary_only_n`).
  - Off-by-one in the END loop: 5 FAIL.
  - pid not trimmed: 4 FAIL.
  - Two mutations survive; see Nit 2.
- **Regression risk and scope.** Without `--require-output`, nothing changes; the `startup_fatal_unrequired` case and the 19-FAIL mutation above pin that. The only caller that passes the option is the x86 gate step in `ubuntu_clean_meson_build.yml`. The new rule has no `next`, so every line still reaches the rules it reached before. The change is +95/-1 and touches only the checker and its self-test, which fits the topic.
- **Can CI block a regression?** Yes. `test_check_valgrind_log.sh` runs in the Static checks job, and that job uses mawk on the runner.
- **Tree scan.** No test source under `tests/`, `gst/` or `ext/` calls `exec*`, `g_spawn`, `g_test_trap_*` or plain `fork()`. `EXPECT_DEATH` uses the default fork-only style, and the runner does not pass `--trace-children`. So the intended failure for a process that execs itself cannot hit any current binary.
- **Docs.** The option is documented only in the script's header comment, and the PR updates it. No other document mentions `--require-output`, and there is no API or architecture change.

## Findings

No Blocking, Major or Minor findings.

## Nits

**Nit 1: a reused pid can still finish an unfinished run.** `tools/debugging/check_valgrind_log.sh:229` (`run_of[pid] = examined`), with the rule at `:296-299`.
- **What is wrong:** `run_of` keeps every pid it has seen for the whole log. Take a run that aborts at startup under pid P. If a later process gets the same pid P and prints an `ERROR SUMMARY`, that summary marks the aborted run finished. The later process could be a forked child of a later binary.
- **Evidence:** a 4-line log (`==5== Command: one`, `==6== Command: two`, `==5== ERROR SUMMARY`, `==6== ERROR SUMMARY`) exits 0 under both awks.
- **Why it is only a nit:** it needs pid wrap-around within one log. The x86 job used pids 9558 to 10148 for the whole step, against the runner's `pid_max`, so this is theoretical today.
- **Fix:** `run_unittests_binaries.sh` runs binaries one at a time, so a summary can only belong to the latest run or to one of its children. Add `split("", run_of)` just before `run_of[pid] = examined`. With that line:
  - The 4-line log above exits 1 and names `unittest_one`.
  - All 108 self-test checks still pass under mawk and gawk.
  - Both real green logs still exit 0.
- **Test:** add the 4-line log above as a `_n` case.

**Nit 2: `timestamped_required` cannot catch the mutation it is meant to catch.** `tools/debugging/test_check_valgrind_log.sh:629`.
- **What is wrong:** both lines carry the identical timestamp `2026-09-07T07:21:27.4959062Z`. If the pid were taken from `$0` before the timestamp is stripped (`pid = $0`), both lines would still get the same "pid", and the case passes. That mutation survives under both awks.
- **Why it matters:** with realistic, differing timestamps, the same mutation reports every run as unfinished. That would break checking a log downloaded from GitHub Actions. CI's own `tee` log has no timestamps, so CI would never show it.
- **Fix:** give the two lines different timestamps, for example `...27.1000000Z` and `...29.2000000Z`. With that change the mutation reports `[unittest_demo] no ERROR SUMMARY` and fails the case.

**Nit 3 (optional): pin the suppression-file fail-open as well.** `tools/debugging/test_check_valgrind_log.sh`, next to `startup_fatal_required_n`.
- **What:** valgrind prints the `Command:` line and then `==PID== FATAL: can't open suppressions file "..."`, with no `ERROR SUMMARY` after it. `main` passes such a log and this PR fails it (verified with real valgrind above).
- **Why:** a trimmed copy as a second `_n` case would record that this fail-open is closed on purpose. The PR body could mention it too.

A smaller observation, not a request: adding `next` to the new `ERROR SUMMARY` rule is the second surviving mutation. The comment at `:295` states that invariant, but no case pins it. It also has no observable effect on real `-v` logs today, because the line after the summary is always blank or classified.

## Verdict

**OK to merge.** The three items above are nits and can go in as a follow-up or be left out.

myungjoo:

Thanks for the review. All three nits are addressed in `63df865a` (still one commit, amended):

- **Nit 1, reused pid:** `split("", run_of)` runs just before `run_of[pid] = examined`, so each new run forgets the ids before it. The 4-line log is now the `reused_pid_n` case. Without the new line, that case fails twice under both mawk and gawk.
- **Nit 2, `timestamped_required`:** the two lines now carry different timestamps. With the id taken before the timestamp is stripped (`pid = $0`), the case now fails under both awks.
- **Nit 3, missing suppression file:** added `suppressions_missing_required_n`, a trimmed copy of the real `FATAL: can't open suppressions file` log. The script header and the PR body now mention it, and the PR body's termination-mode table includes the external `SIGKILL` / `SIGTERM`, `posix_spawn` and orphan cases.
- **Observation, `next` on the new rule:** left as is, since it has no observable effect on real logs and no case pins it.

**Verification (mawk 1.3.4 and gawk 5.1.0):**
- The self-test passes all 113 checks.
- The new tests fail 9 checks against `main`'s checker.
- These logs keep their verdicts under the amended checker:
  - The 6 split logs from jobs 103918379475, 104226833737 and 104228847800, plus this PR's own x86 valgrind job log: exit 0.
  - The real termination-mode logs: startup fatal and self-`exec` exit 1, the rest exit 0.

myungjoo:

This review was written by a separate Claude agent, not by the PR author or a human maintainer, and is relayed here as-is. It is a follow-up round on head `63df865a` (one commit on `main` `0e0ccf19`). I reviewed the whole change again, not only the delta from `6091fb20`. Only the two changed files are in scope.

## Previous nits: are the fixes verified?

All runs below used **mawk 1.3.4** (the runner's awk) and **gawk 5.1.0**, and both awks gave the same result.

- **Nit 1 (reused pid): verified.** `split("", run_of)` is at `check_valgrind_log.sh:231`. `reused_pid_n` fails 2 checks with the `6091fb20` checker, and 2 checks when only that line is deleted. It passes with this head.
- **Nit 2 (`timestamped_required`): verified.** The two lines now carry different timestamps. With the mutation `pid = $0` (the pid taken before the timestamp is stripped), the case fails (1 FAIL).
- **Nit 3 (suppression file): verified.** `suppressions_missing_required_n` is added. It is one of the 9 checks that fail against `main`'s checker, and the header comment mentions this cause.
- **The `next` observation** was left as is, as the author said. Adding `next` to the new rule still passes all 113 checks. That is acceptable: the mutation has no observable effect on real `-v` logs.

## Whole change, re-checked

- **Self-test:** 113/113 pass. Against `main`'s checker, the new tests fail 9 checks. Other mutations:
  - Dropping the `require_output` guard: 19 FAIL.
  - Dropping the unfinished half of the exit: 5 FAIL.
- **Real logs, `main` / `6091fb20` / `63df865a`:** exit 0 under every checker and both awks. The logs are the 6 split logs of jobs 103918379475, 104226833737 and 104228847800, plus the earlier x86 job log of this PR.
- **This head's own x86 valgrind job (104267342166):** the gate step is green. I extracted that step's log and checked the ordering in it:
  - Every one of the 24 runs prints its `ERROR SUMMARY` lines before the next `Command:` line.
  - The two `unittest_plugins` death-test children (`10019`, `10023`) print their summaries inside the parent's span, before the parent's own summary.
  - So no real summary arrives after a later run has started. The checker exits 0 on that log under both awks (`24 binaries examined`, with both segments in one file).
- **Ordering across runs:** `split()` is safe for the current runner.
  - `run_unittests_binaries.sh` starts one valgrind at a time, in the foreground, and CI checks the two logs in separate invocations.
  - When several logs are passed to one invocation, they are read one after another.
  - A forked child that outlives its parent prints its summary under its own id, and that id is never in `run_of`. So a late summary like that cannot finish or un-finish anything.
- **Scope and regression:** the diff is +127/-1 in the checker and its self-test only. Without `--require-output`, the verdict does not change (`startup_fatal_unrequired` and the 19-FAIL guard mutation pin this). Documentation lives in the script header, which is updated. There is no API or architecture change.
- **CI can block a regression:** the Static checks job runs the self-test on the runner's mawk and passed on this head.

## Findings

No Blocking, Major or Minor findings.

## Nits

**Nit A: the Nit 1 fix assumes runs never nest, and nothing says so.** `tools/debugging/check_valgrind_log.sh:231` (`split("", run_of)`), and the header at `:82-85`.

- **What is wrong:** Clearing every id when a `Command:` line appears assumes that the previous run has already ended. That holds for the runner today. It stops holding once memcheck traces children (`--trace-children=yes`). Then a child that `exec`s prints its own `Command:` line while its parent is still running. The parent's summary arrives afterwards, its id has been forgotten, and the parent is reported unfinished.
- **Evidence:** real valgrind 3.18.1 with CI's flags plus `--trace-children=yes`, on a process that forks a child which `exec`s `/bin/true`. The log order is:

  ```
  ==1595== Command: ./modes childexec
  ==1596== Command: /bin/true
  ==1596== ERROR SUMMARY: ...
  ==1595== ERROR SUMMARY: ...
  ```

  With `63df865a`, the check exits 1 under both awks with `[modes] no ERROR SUMMARY: valgrind aborted at startup, was killed, or the log was cut`. With `6091fb20` and with `main`, it exits 0. All other termination modes give the same verdict at both heads, with or without `--trace-children`.
- **Why it matters (only a little):**
  - Today it is inert. The tree has no `--trace-children`, no `death_test_style` setting and no `g_spawn`.
  - The failure goes in the safe direction: a false failure, never a false pass.
  - But once children are traced, any test whose process spawns a helper program would take this path. One example is GStreamer forking `gst-plugin-scanner` when it rebuilds its registry. The gate would then fail with a message that points at a startup abort or a kill that never happened.
  - This trade-off came in with the Nit 1 fix. The earlier head did not have it.
- **Fix:** A comment is enough. Next to `split("", run_of)` and in the header, say that this relies on the runner starting one memcheck run after another, without `--trace-children`, and that tracing children needs a rule that tells a nested run from a later one. A self-test case is not needed: the only thing it could pin today is the false failure.

## Verdict

**OK to merge.**

Nits: Nit A (a comment documenting the no-`--trace-children` assumption behind `split("", run_of)`). It can go in with this PR or be left out.

myungjoo:

Nit A is addressed in `4b44bc38`. The change is comment-only, amended into the single commit.

- **Script header:** now says that forgetting earlier ids relies on runs following one another, the way `packaging/run_unittests_binaries.sh` starts them, without `--trace-children`. It also says that a traced child would print its own `Command:` line inside its parent's run and fail the parent as unfinished.
- **Next to `split("", run_of)`:** a one-line comment points to the header. It has no apostrophe, since the awk program is single-quoted.
- **PR body:** the termination-mode section now says the same.

**Re-checked under mawk 1.3.4 and gawk 5.1.0:**
- The self-test passes 113/113, and fails 9 checks against `main`'s checker.
- The mutation that removes the id reset still fails `reused_pid_n`.
- The 7 real CI logs and the real termination-mode logs keep their verdicts.

myungjoo:

This review was written by a separate Claude agent, not by the PR author or a human maintainer, and is relayed here as-is. It is review round 3, on head `4b44bc38` (one commit on `main` `0e0ccf19`). Only the two changed files are in scope.

## Nit A from round 2: verified

- **The change is comment-only.** `git diff 63df865a 4b44bc38` touches only `tools/debugging/check_valgrind_log.sh`. It adds 3 header lines (`:85-88`) and a one-line awk comment above `split("", run_of)` (`:234`). The self-test file is byte-identical to `63df865a`.
- **The awk program is intact.** I extracted the single-quoted awk program from both heads and dropped comment lines. The two programs are identical, 184 lines each. The new awk comment is inside the program and has no apostrophe. The only `'` in the program is still the existing `'\''` at `:347`. `bash -n` passes.
- **The test would have caught a stray apostrophe.** As a mutation, I changed the new comment to `Runs don't nest ...`. The self-test then fails 97 of 113 checks under both awks.
- **Self-test results.** Under **mawk 1.3.4** and **gawk 5.1.0**, locally:
  - This head passes 113/113.
  - This head's tests fail 9 checks against `main`'s checker.
  - Deleting `split("", run_of)`: 2 FAIL (`reused_pid_n`).
  - Taking the pid before the timestamp is stripped: 1 FAIL.
  - Dropping the `require_output` guard: 19 FAIL.
  - Dropping the unfinished half of the exit: 5 FAIL.
- **CI.** The Static checks job on this head (job 104273167195, runner mawk) shows 113 `PASS` lines and "self-test passed." The x86 valgrind job was still running when I wrote this. The checker's code is unchanged from `63df865a`, whose valgrind gate round 2 checked green.
- **Real logs keep their verdicts.** I ran the checker with `--require-output` under both awks on these logs:
  - The 6 split logs of jobs 103918379475, 104226833737 and 104228847800.
  - The two earlier x86 valgrind logs of this PR.
  - 17 real valgrind 3.18.1 termination-mode logs, with and without `--trace-children=yes`.

  On every log, the exit code and the full output are the same for `63df865a` and `4b44bc38`. The CI logs exit 0. The startup fatal, self-`exec`, and traced child-`exec` logs exit 1, as documented.
- **The comment matches the runner.** `packaging/run_unittests_binaries.sh:85` starts each `valgrind` in the foreground, one binary at a time, without `--trace-children`. Nothing in the tree passes `--trace-children` or sets `death_test_style`.

## Final look at the whole change

I found nothing the earlier rounds missed that needs a change:
- The `split("", arr)` idiom clears the array under both mawk and gawk; the `reused_pid_n` results above show it.
- The new `ERROR SUMMARY` rule cannot be consumed earlier: every rule with `next` before it needs a different line prefix.
- The only callers that pass `--require-output` are the two gate lines in `ubuntu_clean_meson_build.yml:288-289`. They check each log separately.
- An empty log still fails under `--require-output`.
- The diff is still +131/-1 in the checker and its self-test, which fits the topic.
- The option's documentation lives in the script header, and the PR updates it. There is no API or architecture change.

## Findings

No Blocking, Major or Minor findings.

## Nits

**Nit B (optional, wording only): the new comment says any traced child breaks the rule, but only a traced child that `exec`s does.** `tools/debugging/check_valgrind_log.sh:87-88`.

- **What:** The header says that "a traced child would print a Command line of its own while its parent still runs". Under `--trace-children=yes`, valgrind prints a new `Command:` line only when a process `exec`s. A forked child that does not `exec` prints only its own summary. That includes the fork-style `EXPECT_DEATH` children in `unittest_plugins`.
- **Evidence:** real valgrind 3.18.1 logs with CI's flags plus `--trace-children=yes`:
  - A child that aborts, `_exit`s or is killed: 1 `Command:` line, and the checker exits 0.
  - A child that `exec`s `/bin/true`: 2 `Command:` lines, and the checker exits 1 (`[modes] no ERROR SUMMARY`).
- **Why it is only a nit:** the behaviour is correct and the comment errs on the cautious side. But someone who later enables `--trace-children` could read it as "death tests will break the gate". In fact only a helper that `exec`s would, such as `gst-plugin-scanner`.
- **Fix, if wanted:** write "a traced child that execs" in the header, for example: `--trace-children: a traced child that execs would print a Command line of its own while its parent still runs, ...`. No test change is needed.

## Verdict

**OK to merge.**

Nits: Nit B (optional comment wording at `check_valgrind_log.sh:87-88`). It can go in with this PR or be left out.

myungjoo:

Nit B is addressed in `2f3def3d`, a comment-only change amended into the single commit.

- **Header wording:** now "a traced child that execs would print a Command line of its own...". A child that aborts, `_exit`s or is killed prints no second `Command:` line even when traced. The PR body says the same.
- **Re-checked** under mawk 1.3.4 and gawk 5.1.0:
  - The self-test passes 113/113, and 9 of those checks fail against `main`'s checker.
  - Removing the id reset still fails `reused_pid_n`.
  - Taking the pid before the timestamp is stripped still fails `timestamped_required`.

myungjoo:

This review was written by a separate Claude agent, not by the PR author or a human maintainer, and is relayed here as-is. It is review round 4, on head `2f3def3d` (one commit on `main` `0e0ccf19`). Only the two changed files are in scope.

## Nit B from round 3: verified

- **The change is comment-only, and outside the awk program.** `git diff 4b44bc38 2f3def3d` changes only header lines 87-89 of `tools/debugging/check_valgrind_log.sh`. They now read "a traced child that execs would print a Command line of its own while its parent still runs". The self-test file is byte-identical to `4b44bc38`.
- **The awk program is byte-identical to `4b44bc38`, comments included.** Both extracted programs have 193 lines. The only `'` in the program is still the existing `'\''` in the `print` of the "errors reported in this repository's own code" line. `bash -n` passes.
- **The wording now matches real valgrind.** The real valgrind 3.18.1 termination-mode logs from round 3 (CI's flags, with and without `--trace-children=yes`) still give the same results:
  - A traced child that aborts, `_exit`s or is killed prints no second `Command:` line. The checker exits 0.
  - A traced child that `exec`s prints a second `Command:` line. The checker exits 1 (`[modes] no ERROR SUMMARY`).
  - The process `exec`ing itself also exits 1, as the PR body documents.

## Script still works (mawk 1.3.4 and gawk 5.1.0, locally)

- **Self-test:** 113/113 PASS under both awks.
- **The new tests against `main`'s checker:** 9 FAIL.
- **Mutations of this head:**
  - Deleting `split("", run_of)`: 2 FAIL (`reused_pid_n`).
  - Taking the pid before the timestamp is stripped: 1 FAIL (`timestamped_required`).
  - Dropping the `require_output` guard: 19 FAIL.
  - Dropping the unfinished half of the exit: 5 FAIL.
  - Dropping the leak half of the exit: 7 FAIL.
- **Empty log** under `--require-output`: exit 1.
- **Real logs:** I ran `main`, `4b44bc38` and `2f3def3d` with `--require-output` under both awks, on:
  - the 6 split logs of jobs 103918379475, 104226833737 and 104228847800;
  - the two earlier x86 valgrind step logs of this PR;
  - 17 real termination-mode logs.

  For every log, `4b44bc38` and `2f3def3d` give the same exit code and byte-identical output. The CI logs exit 0. Only the startup-fatal, self-`exec` and traced child-`exec` logs exit 1.
- **CI on this head:** Static checks (job 104276087251, runner mawk) shows 113 `PASS` lines for the valgrind log checker self-test and "check_valgrind_log.sh self-test passed". The x86 valgrind job was still running when I wrote this. The checker's code is unchanged from `63df865a`, whose x86 gate was verified green in round 2.

## Final look at the whole change

Nothing new needs a code change:
- **Topic:** the change addresses the fail-open it describes.
- **Regression risk:** without `--require-output`, the verdict does not change.
- **Callers:** the only callers that pass `--require-output` are `ubuntu_clean_meson_build.yml:288-289`, one log per invocation.
- **CI can block a regression:** the self-test runs in Static checks.
- **Size:** the diff is +132/-1 in the checker and its self-test only.
- **Documentation:** the option's documentation lives in the script header, and the PR updates it. There is no API or architecture change.

## Findings

No Blocking, Major or Minor findings.

## Nits

**Nit C (PR body only, no code change): one mutation count in the body is stale.**
- **Where:** the last paragraph of the PR description, which says dropping either half of the combined exit condition fails "7 without the leak half, 3 without the unfinished half".
- **What is wrong:** on the final code, dropping the unfinished half (`exit (leak_ours_count > 0)`) fails **5** checks, not 3, under both mawk and gawk:
  - `startup_fatal_required_n`
  - `suppressions_missing_required_n`
  - `one_of_two_unfinished_n`
  - `child_summary_only_n`
  - `reused_pid_n`

  The 3 matches the rebase head `6091fb20`: the same mutation there gives 105 PASS / 3 FAIL under both awks. That head did not yet have `suppressions_missing_required_n` or `reused_pid_n`, which were added for round-1 review points. The "7 without the leak half" count is still correct.
- **Why it matters:** a maintainer who repeats the mutation gets a number that disagrees with the description. That makes the evidence harder to trust, although the code is right.
- **Fix:** change "3 without the unfinished half" to "5 without the unfinished half" in the PR description. Nothing else in the body contradicts the final code:
  - the 9-FAIL count against `main`;
  - the 113-check total;
  - the termination-mode table;
  - the "traced child that `exec`s" sentence;
  - the case list.

  The x86 job row is explicitly labelled `6091fb20`, so it is accurate as a historical result.

## Verdict

**OK to merge.**

Nits: Nit C (stale "3" mutation count in the PR description; the code and commit message are unaffected).

myungjoo:

Nit C is fixed: the description now says 5 checks fail when the unfinished half of the exit condition is dropped (that was 3 before `reused_pid_n` and `suppressions_missing_required_n` were added). This was a description-only change, and the code is unchanged at `2f3def3d`.

PR #4962, opened 2026-09-11
[Element/Split] Keep the segment count across a stop and restart

myungjoo:
## Problem

On PAUSED→READY, `gst_tensor_split_remove_src_pads()` removes the source pads and also sets `num_tensors = 0`. It leaves the `tensorseg` rule array that `num_tensors` counts in place. `num_tensors` is otherwise written only when `tensorseg` is set, so after a stop (to READY or NULL) and restart without setting `tensorseg` again, `gst_tensor_split_chain()` loops zero times:

- no source pad is created and no buffer goes out, yet the push returns `GST_FLOW_OK`;
- the stream then ends with "Got EOS before adding any pads" (an ERROR);
- a buffer that no longer fits the rule is not refused either.

I found this while working on #4927 (#4954).

## Fix

`remove_src_pads()` no longer touches `num_tensors`: the count belongs to the configuration, like the rule array, not to the per-stream state it resets. This one-line removal also drops the only write to `num_tensors` that was made without the object lock, which `set_property()` and `chain()` both take for it.

## Tests

All four tests are in `unittest_plugins`, `testTensorSplit`. Each fails without the fix (checked locally against the unmodified element) and passes with it.

- `restartKeepsTensorpick`: with `tensorpick=0` over segments of different sizes, the restarted element again outputs only the picked segment and emits `no-more-pads` once per run.

The harness cases also check that every buffer is the size its pad's caps announce. A non-zero pick is left out of that check on purpose: on `main` its caps are still wrong (#4927, fixed in #4954).
- `restartKeepsTensorseg`: after READY and back, a harness-fed element creates both pads again and pushes both segments. Before the fix: no pads, no buffers.
- `restartRefusesSmallerTensor_n`: after the restart, a tensor smaller than the kept rule is refused with `GST_FLOW_ERROR`. Before the fix: `GST_FLOW_OK` with nothing done.
- `replayPipeline`: a real pipeline is played, stopped to NULL and played again. The new pad is relinked from `pad-added`, the way applications do. It reaches EOS both times. Before the fix: ERROR on the second play.

## Relation to #4954

Both PRs touch `gsttensor_split.c` and `unittest_plugins.cc`, but in different hunks, and the test helpers have distinct names. `git merge-tree` gives the same conflict-free tree in either merge order. That merged tree builds, and all `testTensorSplit` cases pass.

## Local verification (WSL, Ubuntu 22.04)

- `meson test`: 22/23 passed. The one failure, `unittest_filter_python3`, is the local NumPy 2.2.6 environment issue (`_ARRAY_API not found`) and is unrelated to this change.
- SSAT `nnstreamer_split`, `nnstreamer_decoder_pose` and `nnstreamer_demux` pass.
- gst-indent and clang-format 16 report no changes. `doxygen-tag.sh` in advanced mode, with ctags, passes.
- Both changed objects compile with `-Werror=sign-compare -Werror`.

### DIFF
--- a/gst/nnstreamer/elements/gsttensor_split.c
+++ b/gst/nnstreamer/elements/gsttensor_split.c
@@ -190,7 +190,6 @@ gst_tensor_split_remove_src_pads (GstTensorSplit * split)
     split->srcpads = g_slist_delete_link (split->srcpads, split->srcpads);
   }
   split->srcpads = NULL;
-  split->num_tensors = 0;
   split->num_srcpads = 0;
   gst_tensors_config_free (&split->in_config);
 }
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -11458,6 +11458,295 @@ TEST (testTensorSplit, finalizeWithoutTensorseg)
 }
 
 /**
+ * @brief What went out of a tensor_split across its source pads.
+ */
+typedef struct {
+  guint buffers; /**< buffers pushed on any source pad */
+  gsize bytes; /**< bytes pushed on any source pad */
+  guint mismatched; /**< buffers of another size than the caps of their pad say */
+} splitRestartOutput;
+
+/**
+ * @brief Buffer probe counting what a tensor_split source pad pushes.
+ * @details The buffer is dropped, which the pad reports as a successful push,
+ *          so the element goes on to the next segment as if it were linked.
+ */
+static GstPadProbeReturn
+_split_restart_count (GstPad *pad, GstPadProbeInfo *info, gpointer user_data)
+{
+  splitRestartOutput *out = (splitRestartOutput *) user_data;
+  gsize size = gst_buffer_get_size (GST_PAD_PROBE_INFO_BUFFER (info));
+  GstCaps *caps = gst_pad_get_current_caps (pad);
+  GstTensorsConfig config;
+
+  gst_tensors_config_init (&config);
+  if (caps == NULL || !gst_tensors_config_from_caps (&config, caps, TRUE)
+      || gst_tensors_info_get_size (&config.info, 0) != size)
+    out->mismatched++;
+  gst_tensors_config_free (&config);
+  if (caps)
+    gst_caps_unref (caps);
+
+  out->buffers++;
+  out->bytes += size;
+  return GST_PAD_PROBE_DROP;
+}
+
+/**
+ * @brief Handler of "pad-added", which puts the counting probe on the new pad.
+ */
+static void
+_split_restart_pad_added (GstElement *element, GstPad *pad, gpointer user_data)
+{
+  UNUSED (element);
+
+  gst_pad_add_probe (pad, GST_PAD_PROBE_TYPE_BUFFER, _split_restart_count, user_data, NULL);
+}
+
+/**
+ * @brief Start a stream of uint8 tensors of the given shape into a tensor_split harness.
+ * @details The sink pad forgets its sticky events when it is deactivated, so
+ *          a restarted element needs the stream to start over.
+ * @return bytes of one tensor of the stream
+ */
+static gsize
+_split_restart_start_stream (GstHarness *h, const gchar *dimension)
+{
+  GstTensorsConfig config;
+  gsize size;
+
+  gst_tensors_config_init (&config);
+  config.info.num_tensors = 1;
+  config.info.info[0].type = _NNS_UINT8;
+  gst_tensor_parse_dimension (dimension, config.info.info[0].dimension);
+  config.rate_n = 0;
+  config.rate_d = 1;
+
+  gst_harness_push_event (h, gst_event_new_stream_start ("split-restart"));
+  gst_harness_set_src_caps (h, gst_tensors_caps_from_config (&config));
+  size = gst_tensors_info_get_size (&config.info, 0);
+  gst_tensors_config_free (&config);
+
+  return size;
+}
+
+/**
+ * @brief Take the element down to READY and back, as a stopped and restarted pipeline does.
+ */
+static void
+_split_restart_cycle (GstHarness *h)
+{
+  EXPECT_EQ (gst_element_set_state (h->element, GST_STATE_READY), GST_STATE_CHANGE_SUCCESS);
+  EXPECT_EQ (gst_element_set_state (h->element, GST_STATE_PLAYING), GST_STATE_CHANGE_SUCCESS);
+}
+
+/**
+ * @brief A tensor_split taken to READY and back splits by the tensorseg it had.
+ * @details Going down to READY removes the source pads, and a restart has to
+ *          create them again, under the same names, from the rule that is
+ *          still set.
+ */
+TEST (testTensorSplit, restartKeepsTensorseg)
+{
+  GstHarness *h = gst_harness_new_with_padnames ("tensor_split", "sink", NULL);
+  splitRestartOutput out = { 0, 0, 0 };
+  GstPad *pad;
+  gsize size;
+
+  ASSERT_TRUE (h != NULL);
+  g_signal_connect (h->element, "pad-added", G_CALLBACK (_split_restart_pad_added), &out);
+  g_object_set (h->element, "tensorseg", "1:4:4,2:4:4", NULL);
+
+  size = _split_restart_start_stream (h, "3:4:4");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_OK);
+  EXPECT_EQ (out.buffers, 2U);
+  EXPECT_EQ (out.bytes, 48U);
+
+  _split_restart_cycle (h);
+  EXPECT_EQ ((guint) h->element->numsrcpads, 0U);
+
+  size = _split_restart_start_stream (h, "3:4:4");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_OK);
+  EXPECT_EQ (out.buffers, 4U);
+  EXPECT_EQ (out.bytes, 96U);
+  EXPECT_EQ (out.mismatched, 0U);
+  EXPECT_EQ ((guint) h->element->numsrcpads, 2U);
+
+  pad = gst_element_get_static_pad (h->element, "src_0");
+  EXPECT_TRUE (pad != NULL);
+  if (pad)
+    gst_object_unref (pad);
+  pad = gst_element_get_static_pad (h->element, "src_1");
+  EXPECT_TRUE (pad != NULL);
+  if (pad)
+    gst_object_unref (pad);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief Handler of "no-more-pads" counting how many times the element reported it.
+ */
+static void
+_split_restart_no_more_pads (GstElement *element, gpointer user_data)
+{
+  UNUSED (element);
+
+  (*(guint *) user_data)++;
+}
+
+/**
+ * @brief A restarted tensor_split still outputs only the picked segment.
+ * @details The pads made for the pick are counted from zero again, so the
+ *          element reports that it has no more pads once per run. The pick
+ *          starts at segment 0, whose pad is described the same way with or
+ *          without the fix for #4927.
+ */
+TEST (testTensorSplit, restartKeepsTensorpick)
+{
+  GstHarness *h = gst_harness_new_with_padnames ("tensor_split", "sink", NULL);
+  splitRestartOutput out = { 0, 0, 0 };
+  guint no_more_pads = 0;
+  GstPad *pad;
+  gsize size;
+
+  ASSERT_TRUE (h != NULL);
+  g_signal_connect (h->element, "pad-added", G_CALLBACK (_split_restart_pad_added), &out);
+  g_signal_connect (h->element, "no-more-pads",
+      G_CALLBACK (_split_restart_no_more_pads), &no_more_pads);
+  g_object_set (h->element, "tensorseg", "1:4:4,2:4:4", "tensorpick", "0", NULL);
+
+  size = _split_restart_start_stream (h, "3:4:4");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_OK);
+  EXPECT_EQ (out.buffers, 1U);
+  EXPECT_EQ (out.bytes, 16U);
+  EXPECT_EQ (no_more_pads, 1U);
+
+  _split_restart_cycle (h);
+
+  size = _split_restart_start_stream (h, "3:4:4");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_OK);
+  EXPECT_EQ (out.buffers, 2U);
+  EXPECT_EQ (out.bytes, 32U);
+  EXPECT_EQ (out.mismatched, 0U);
+  EXPECT_EQ (no_more_pads, 2U);
+  EXPECT_EQ ((guint) h->element->numsrcpads, 1U);
+
+  pad = gst_element_get_static_pad (h->element, "src_0");
+  EXPECT_TRUE (pad != NULL);
+  if (pad)
+    gst_object_unref (pad);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief A restarted tensor_split still refuses a tensor its tensorseg does not fit.
+ * @details The rule kept across the restart asks for 48 bytes, and the new
+ *          stream only brings 12, so the buffer has to be refused rather than
+ *          passed over without a word.
+ */
+TEST (testTensorSplit, restartRefusesSmallerTensor_n)
+{
+  GstHarness *h = gst_harness_new_with_padnames ("tensor_split", "sink", NULL);
+  splitRestartOutput out = { 0, 0, 0 };
+  gsize size;
+
+  ASSERT_TRUE (h != NULL);
+  g_signal_connect (h->element, "pad-added", G_CALLBACK (_split_restart_pad_added), &out);
+  g_object_set (h->element, "tensorseg", "1:4:4,2:4:4", NULL);
+
+  size = _split_restart_start_stream (h, "3:4:4");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_OK);
+  EXPECT_EQ (out.buffers, 2U);
+
+  _split_restart_cycle (h);
+
+  size = _split_restart_start_stream (h, "3:2:2");
+  EXPECT_EQ (gst_harness_push (h, gst_harness_create_buffer (h, size)), GST_FLOW_ERROR);
+  EXPECT_EQ (out.buffers, 2U);
+
+  gst_harness_teardown (h);
+}
+
+/**
+ * @brief Handler of "pad-added" linking the first source pad to the sink, as an application does.
+ */
+static void
+_split_restart_link_pad (GstElement *element, GstPad *pad, gpointer user_data)
+{
+  GstPad *sinkpad = gst_element_get_static_pad (GST_ELEMENT (user_data), "sink");
+
+  UNUSED (element);
+
+  if (!gst_pad_is_linked (sinkpad)) {
+    EXPECT_EQ (gst_pad_link (pad, sinkpad), GST_PAD_LINK_OK);
+  }
+  gst_object_unref (sinkpad);
+}
+
+/**
+ * @brief Play a pipeline until it ends.
+ * @return the type of the message it ended with, GST_MESSAGE_UNKNOWN on time-out
+ */
+static GstMessageType
+_split_restart_play (GstElement *pipeline)
+{
+  GstBus *bus = gst_element_get_bus (pipeline);
+  GstMessageType type = GST_MESSAGE_UNKNOWN;
+  GstMessage *msg;
+
+  gst_element_set_state (pipeline, GST_STATE_PLAYING);
+  msg = gst_bus_timed_pop_filtered (bus, (GstClockTime) TEST_TIMEOUT_LIMIT * GST_USECOND,
+      (GstMessageType) (GST_MESSAGE_EOS | GST_MESSAGE_ERROR));
+  if (msg) {
+    type = GST_MESSAGE_TYPE (msg);
+    gst_message_unref (msg);
+  }
+  gst_object_unref (bus);
+
+  return type;
+}
+
+/**
+ * @brief A pipeline stopped to NULL and played again gets its tensors again.
+ * @details The source pads go away when the pipeline stops, so the new ones
+ *          are linked from "pad-added" again, the way an application does it.
+ */
+TEST (testTensorSplit, replayPipeline)
+{
+  GstElement *pipeline, *split, *sink;
+  guint data_received = 0;
+
+  pipeline = gst_parse_launch ("videotestsrc num-buffers=2 ! "
+                               "video/x-raw,format=RGB,width=4,height=4,framerate=30/1 ! "
+                               "tensor_converter ! tensor_split name=split tensorseg=1:4:4,2:4:4 "
+                               "tensor_sink name=sinkx",
+      NULL);
+  ASSERT_TRUE (pipeline != NULL);
+
+  split = gst_bin_get_by_name (GST_BIN (pipeline), "split");
+  sink = gst_bin_get_by_name (GST_BIN (pipeline), "sinkx");
+  ASSERT_TRUE (split != NULL);
+  ASSERT_TRUE (sink != NULL);
+
+  g_signal_connect (split, "pad-added", G_CALLBACK (_split_restart_link_pad), sink);
+  g_signal_connect (sink, "new-data", G_CALLBACK (count_output), &data_received);
+
+  EXPECT_EQ (_split_restart_play (pipeline), GST_MESSAGE_EOS);
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  EXPECT_EQ (data_received, 2U);
+
+  EXPECT_EQ (_split_restart_play (pipeline), GST_MESSAGE_EOS);
+  EXPECT_EQ (setPipelineStateSync (pipeline, GST_STATE_NULL, UNITTEST_STATECHANGE_TIMEOUT), 0);
+  EXPECT_EQ (data_received, 4U);
+
+  gst_object_unref (sink);
+  gst_object_unref (split);
+  gst_object_unref (pipeline);
+}
+
+/**
  * @brief The dimension of the tensor the decoder input cases negotiate
  */
 #define TEST_DECODER_DIM "3:16:16"


myungjoo:
**This is an automated code review produced by a separate Claude agent (not by @myungjoo directly), relayed to this PR via the author's GitHub account.**

Scope reviewed: the diff of this PR only (`gst/nnstreamer/elements/gsttensor_split.c` one-line removal, and the three new `testTensorSplit` cases added to `tests/nnstreamer_plugins/unittest_plugins.cc`). Existing repository code and the note about interaction with PR #4954 were read for context, but findings below are limited to the changed lines.

## What I checked

- **Root cause**: confirmed `num_tensors` is written in exactly two places pre-patch: `gst_tensor_split_init()` (0, at construction) and `gst_tensor_split_set_property()` for `PROP_TENSORSEG` (set to the parsed rule count, under `GST_OBJECT_LOCK`). `gst_tensor_split_remove_src_pads()` — invoked on `GST_STATE_CHANGE_PAUSED_TO_READY` and from `finalize()` — additionally zeroed it *without* holding the lock. After a stop without re-setting `tensorseg`, `gst_tensor_split_chain()`'s `for (i = 0; i < num_tensors; i++)` loop (reading `num_tensors` under the lock) runs zero times: no pads are (re)created, `chain()` still returns `GST_FLOW_OK`, and the follow-on EOS hits the `!split->srcpads` branch in `gst_tensor_split_event()`, producing the reported "Got EOS before adding any pads" error. This matches the PR description exactly, and the fix (stop clearing `num_tensors` in `remove_src_pads`) addresses it directly. As a side effect it also removes the one unlocked write to `num_tensors`, which is a small thread-safety improvement, not a new risk.
- **Regression scope**: `num_tensors` is otherwise read only in `gst_tensor_split_chain()` under the object lock, and `tensorseg`/`num_tensors` are always replaced together in `set_property`. `gst_tensor_demux` (the sibling element) derives its own per-buffer tensor count from `gst_tensor_buffer_get_count(buf)` rather than from persisted state, so it does not share this bug and needs no analogous fix. I found no other reader of `num_tensors`, and no other module is touched by the diff, so I don't see a plausible regression path outside `tensor_split`.
- **Size/scope**: the functional fix is a single line; the remainder is test code. Nothing outside `gsttensor_split.c` and the test file is touched. Proportionate to the bug.
- **Tests**: traced all three new cases by hand against the actual code paths:
  - `restartKeepsTensorseg` / `restartRefusesSmallerTensor_n` rely on `gst_element_set_state()` down to `READY` deactivating (and clearing sticky events on) `split`'s sink pad via GstElement's default pad-activation handling for `PAUSED_TO_READY`/`READY_TO_PAUSED`, which is why the helper re-sends `stream-start`/caps after the cycle — this matches standard GStreamer pad-activation semantics. The buffer-probe trick (`GST_PAD_PROBE_DROP` on `GST_PAD_PROBE_TYPE_BUFFER`) correctly makes `gst_pad_push()` report `GST_FLOW_OK` without needing the src pads to be linked to anything, since probes run before the peer-linked check — a standard idiom also usable here because the harness has no fixed sink pad for these two tests. `restartRefusesSmallerTensor_n`'s byte math (48 bytes required, 12 supplied) is correct given `gst_tensor_get_element_count()`'s "stop at the first zero dim" behavior, and correctly exercises the bounds check added in the companion PR #4954, confirming the two fixes compose correctly.
  - `replayPipeline` deliberately omits the `!` between `tensor_split` and `tensor_sink` in the `gst_parse_launch()` string (verified in the diff — this is intentional, not a typo) so that only the test's own `pad-added` handler links `src_0`, avoiding a second, unwanted auto-link attempt by `gst-parse-launch`'s own sometimes-pad deferred-linking mechanism onto an already-linked sink. This is a deliberate and correct way to mimic what a real application does.
  - All three tests fail without the fix and pass with it per the PR description, and my independent trace of `chain()`/`remove_src_pads()` agrees with that claim.
  - Because these are ordinary `TEST()` cases in `unittest_plugins`, they run under the existing `meson test` CI job, so a future regression that reintroduces this bug (or a change elsewhere that alters pad-deactivation/sticky-event or bounds-check behavior in a way that breaks this restart path) would fail CI and block merge, satisfying the "future regression must be caught" criterion.
  - `clang-format -style=file` on the whole test file after the patch produces no diff, and doxygen `@brief`/`@details`/`@return` comments are present on every new static function and test case.
- **Docs/architecture**: no public API, property, or documented behavior changes; nothing here required a Documentation/ update, and none was needed.

## Findings

No Critical, Major, or Minor issues found.

- **Nit** (`tests/nnstreamer_plugins/unittest_plugins.cc`, new tests): coverage is solid for the reported bug; an optional (non-blocking) addition would be a restart test combined with `tensorpick` set, since that path also touches `num_srcpads`/`num_tensors` together. Not required for this fix.

## Verdict

**OK to merge as is.** The fix is minimal, correctly targeted at the reported defect, verified against the actual pad-activation/sticky-event and bounds-check code paths, does not touch any other module, and ships with regression tests at both the unit (GstHarness) and integration (real pipeline replay) level that will fail CI if the bug — or the PR #4954 interaction — regresses.

---
🤖 This review was generated by a separate Claude agent session and posted here by the PR author on its behalf.

myungjoo:

Thanks for the review. The Nit is addressed in 6f4fe484 (force-pushed and folded into the single commit):

- **Nit, restart with `tensorpick`:** added `restartKeepsTensorpick`. With `tensorseg=1:4:4,2:4:4 tensorpick=1`, both runs output only the 32-byte picked segment, the restarted element has one source pad, and `no-more-pads` is emitted once per run (1, then 2). This checks that the pad count restarts from zero along with the pads. Like the other three tests, it fails against the unmodified element and passes with the fix.

The combination with #4954 is re-verified for this head: `git merge-tree` is conflict-free, and the merged tree builds and passes all 19 `testTensorSplit` cases. clang-format 16, doxygen-tag and the strict `-Werror=sign-compare` compile are clean.

myungjoo:
**This is an automated code review produced by a separate Claude agent (not by @myungjoo directly), relayed to this PR via the author's GitHub account.** This is an independent re-review of the current head (6f4fe484), including re-checking the earlier automated review's own claims against the code rather than just re-stating them.

## Scope

Reviewed the diff only: the one-line removal in `gst/nnstreamer/elements/gsttensor_split.c` (`gst_tensor_split_remove_src_pads`) and the four new `testTensorSplit` cases in `tests/nnstreamer_plugins/unittest_plugins.cc` (`restartKeepsTensorseg`, `restartKeepsTensorpick`, `restartRefusesSmallerTensor_n`, `replayPipeline`). Existing code, PR #4954, and the earlier review thread were read for context only.

## Verification performed

- Re-derived the bug independently from the source: `num_tensors` is written only in `gst_tensor_split_init()` (0) and `set_property()`'s `PROP_TENSORSEG` case (under `GST_OBJECT_LOCK`, together with `tensorseg`). `gst_tensor_split_remove_src_pads()`, called from `change_state()`'s `PAUSED_TO_READY` and from `finalize()`, zeroed it without the lock, while leaving `tensorseg` (the rule array `num_tensors` is supposed to count) untouched. After a stop/restart without re-setting `tensorseg`, `gst_tensor_split_chain()`'s `for (i = 0; i < num_tensors; i++)` loop runs zero times: `GST_FLOW_OK` is returned with no pads created and no data pushed, and the follow-on EOS hits `gst_tensor_split_event()`'s `!split->srcpads` branch. The one-line fix is directly targeted at this and is correct. `num_srcpads` (also reset in the same function) is unrelated to `num_tensors` and is correctly left alone.
- Checked for regression paths: `num_tensors` has no other reader. `tensor_demux`, the sibling element, derives its per-buffer count from `gst_tensor_buffer_get_count(buf)` rather than persisted state, so it isn't affected. No other file is touched.
- Independently confirmed the earlier reviewer's Nit was actually addressed: `restartKeepsTensorpick` is present in the current head and exercises the restart+`tensorpick` combination (asserts pad count, buffer/byte counts, and `no-more-pads` emission count across the restart).
- Traced `restartRefusesSmallerTensor_n` against `gst_tensor_split_get_splitted()`: the refusal it exercises comes from the `offset + size > src_info.size` check there. I checked `git log -S` on that message and it predates this PR and #4954 (introduced by commit `b5eb160b`, already on `main`) — **the earlier review's line "correctly exercises the bounds check added in the companion PR #4954" is inaccurate**; that check is pre-existing, and #4954 actually *removes* a different, unrelated bounds check (`num_srcpads >= tensorseg->len` in `gst_tensor_split_get_tensor_pad`). This doesn't affect correctness of this PR — the test still passes both before and after #4954 merges, just not for the reason previously stated — but it's a correction worth recording since it was offered as the basis for a composability claim.
- Checked test-helper reuse: `count_output`, `setPipelineStateSync`, `TEST_TIMEOUT_LIMIT`, `UNITTEST_STATECHANGE_TIMEOUT` are all pre-existing helpers from `tests/unittest_util.h` already used elsewhere in this file — no new/duplicated test infrastructure.
- Ran `clang-format -style=file` on `tests/nnstreamer_plugins/unittest_plugins.cc` after the patch: no diff. (`gsttensor_split.c` is a C file formatted with `gst-indent`, not `clang-format`, per this repo's convention, so a whole-file `clang-format` diff on it is expected and not meaningful; the single removed line does not disturb indentation either way.)
- Doxygen `@brief`/`@details`/`@return` present on every new static helper and `TEST` case.
- No public API, property, or documented behavior change; no Documentation/ update needed, and none was made.
- CI at review time: `Static checks`, `DCO`, `CPP-Linter`, `Analyze (python)`, and the LLVM/Clang and macOS meson builds pass; the jobs that actually execute `testTensorSplit` (GBS x86_64 `unit_test 1`, meson+Valgrind, etc.) were still `pending`/queued and had not reported yet as of this review.

## Findings

No Critical or Major issues.

- **Minor** — informational, not a defect in this PR: the earlier review's stated reason for why `restartRefusesSmallerTensor_n` "confirms the two fixes compose" is factually wrong (see above); the actual composability holds, but for a different mechanism. Worth noting so nobody relies on that specific claim later.
- **Nit** (`tests/nnstreamer_plugins/unittest_plugins.cc`, `restartKeepsTensorpick`): the test asserts buffer count, byte count, pad count and `no-more-pads` count, but never checks the caps/dimensions actually announced on the recreated pad. Under current `main` (pre-#4954), `gst_tensor_split_get_tensor_pad()` sets the pad's caps from `tensorseg[num_srcpads]` rather than `tensorseg[nth]`, so with `tensorpick` the announced caps can already mismatch the segment being pushed — this test happens not to catch that because it only checks byte totals, not caps. Not a regression from this PR (the caps indexing is untouched here and is #4954's target), and I confirmed the test still passes unmodified once #4954 lands (which switches the indexing to `nth`). Optional, non-blocking: an assertion on the pad's negotiated caps would make this test also double as a guard for that interaction going forward.
- **Nit**: full CI (the jobs that actually run `testTensorSplit`) had not finished as of this review; worth confirming green before merging, per the repo's own merge criteria.

## Verdict

**OK to merge once the still-pending CI jobs (the ones that actually execute `testTensorSplit`, e.g. the GBS x86_64 unit-test build) report green.** The fix is minimal and correctly targeted, does not touch any other module, and ships with regression coverage (harness-level and full-pipeline-replay-level) that will fail CI if this bug — or its interaction with #4954 — regresses. No blocking findings from this review.

---
🤖 This review was generated by a separate Claude agent session and posted here by the PR author on its behalf.

myungjoo:

Thanks for the re-review. Point by point, for 8909a483:

- **Nit, caps on the recreated pads:** addressed. The counting probe now also compares every buffer with the size its pad's current caps announce. `restartKeepsTensorseg` and `restartKeepsTensorpick` assert that no buffer disagrees, before or after the restart.
  - `restartKeepsTensorpick` now picks segment 0 of `1:4:4,2:4:4`, so the pick still drops a differently sized segment. A non-zero pick can't be used here yet, because on `main` its caps are still wrong. That is #4927, which #4954 fixes.
  - The check is live. As a negative control, switching the test back to `tensorpick=1` on `main`'s element makes it report 2 mismatches.
- **Minor, the earlier review's attribution of the refusal:** agreed. `restartRefusesSmallerTensor_n` is refused by the `offset + size > src_info.size` check in `gst_tensor_split_get_splitted()`, already on `main`, not by anything in #4954. The earlier review's comment is not part of the code, so nothing changes.
- **Nit, pending CI:** noted. CI is re-running on the new head.

Re-verified for this head:
- The restart tests fail on the unmodified element and pass with the fix.
- The tree merged with #4954 is conflict-free, builds, and passes all 19 `testTensorSplit` cases.
- clang-format 16, gst-indent, doxygen-tag and the strict `-Werror=sign-compare` compile are clean.
- `meson test` passes 22/23. The only failure is `unittest_filter_python3`, from the local NumPy environment.

myungjoo:
**This is an automated code review produced by a separate Claude agent (not by @myungjoo directly), relayed to this PR via the author's GitHub account.** This is a fresh, independent review of the current head (`8909a483`), including re-verifying the two earlier automated reviews' claims against the code rather than taking them on trust.

## Scope

Reviewed the diff only: the one-line removal in `gst/nnstreamer/elements/gsttensor_split.c` (`gst_tensor_split_remove_src_pads`) and the four new `testTensorSplit` cases in `tests/nnstreamer_plugins/unittest_plugins.cc` (`restartKeepsTensorseg`, `restartKeepsTensorpick`, `restartRefusesSmallerTensor_n`, `replayPipeline`). Existing repository code, PR #4954, and the prior review thread were read for context only; findings are limited to the changed lines.

## Verification performed

- **Root cause and fix, re-derived from source, not from the PR description**: `split->num_tensors` (`guint32`, `gst/nnstreamer/elements/gsttensor_split.h:57`) is written in exactly two remaining places: `gst_tensor_split_init()` (0, at construction) and `gst_tensor_split_set_property()`'s `PROP_TENSORSEG` case, which sets `tensorseg` and `num_tensors` together under `GST_OBJECT_LOCK` (`gsttensor_split.c:668-671`). `gst_tensor_split_remove_src_pads()`, invoked from `change_state()`'s `PAUSED_TO_READY` and from `finalize()`, used to also zero `num_tensors` without holding the lock, while leaving the `tensorseg` array it counts untouched. After a stop without re-setting `tensorseg`, `gst_tensor_split_chain()`'s `for (i = 0; i < num_tensors; i++)` loop (`gsttensor_split.c:491`) ran zero times: no pad was (re)created, `chain()` still returned `GST_FLOW_OK`, and a subsequent EOS hit the `!split->srcpads` branch in `gst_tensor_split_event()`, producing "Got EOS before adding any pads." The one-line fix removes exactly this write, restoring the invariant that `num_tensors == tensorseg->len`, which is otherwise only ever changed atomically together in `set_property`. This is correct and minimal, and also removes the sole unlocked write to a field every other accessor reads/writes under `GST_OBJECT_LOCK`.
- **Regression scope**: grepped the whole tree for `num_tensors` outside `gsttensor_split.c/.h`; the only other hits (`tests/nnstreamer_sink/unittest_sink.cc`) are `GstTensorsInfo.num_tensors`, an unrelated struct field, confirming `GstTensorSplit::num_tensors` has no reader outside this file. `num_srcpads`, which the same function still resets to 0, is a distinct counter (recreated pad index) and is correctly left alone. No other module is touched by the diff.
- **Interaction with #4954**: independently diffed #4954's test additions and confirmed no symbol collision with this PR's new helpers — #4954 uses `splitPadOutput`/`SPLIT_TEST_MAX_PADS`/`_split_run_pipeline`; this PR uses `splitRestartOutput`/`_split_restart_*`. Both files' hunks are non-overlapping and target different call sites (#4954 changes `gst_tensor_split_get_tensor_pad`'s caps/pad-name indexing and removes an unrelated `num_srcpads >= tensorseg->len` bound in that function; this PR touches only `remove_src_pads`). This is consistent with the PR body's conflict-free-merge claim.
- **Test correctness, traced by hand**:
  - `restartKeepsTensorseg`/`restartKeepsTensorpick`/`restartRefusesSmallerTensor_n` correctly re-send `stream-start`+caps after the READY/PLAYING cycle, matching GstPad's sticky-event-clearing behavior on deactivation. The `GST_PAD_PROBE_DROP` counting probe correctly reports `GST_FLOW_OK` to the element without a linked peer, which is why these three tests work purely against a `GstHarness` sink-only setup.
  - Byte math checks out: `1:4:4`=16B, `2:4:4`=32B, `3:4:4`=48B combined (`restartKeepsTensorseg`); `restartRefusesSmallerTensor_n`'s post-restart 12B buffer (`3:2:2`) is correctly less than the retained 48B rule, so the refusal is exercised via the pre-existing `offset + size > src_info.size` check in `gst_tensor_split_get_splitted()` — this is on `main` already, not new in this PR or #4954 (confirmed via `git log -S` per the second prior review; I independently re-confirm this line predates both PRs).
  - `restartKeepsTensorpick` deliberately keeps `tensorpick=0` (not a non-zero pick) because, as its own doc-comment states, a non-zero pick's caps are still wrong on `main` until #4954 lands; the mismatched-caps assertion (`out.mismatched`) would otherwise not be meaningful here. This is a reasonable, clearly-documented scope limit, not a gap.
  - `replayPipeline` uses a real pipeline without the `!` link between `tensor_split` and `tensor_sink`, so only the test's own `pad-added` handler links `src_0` — a deliberate, correct way to mimic app-driven relinking after a stop/restart, and distinct from #4954's `_split_run_pipeline` helper (which does use `!` and doesn't restart).
  - All four cases fail on the unmodified element and pass with the fix, matching the PR's own local-verification claim, and this reasoning holds up against the actual code paths (not just the PR's prose).
- **Test infra reuse**: `count_output`, `setPipelineStateSync`, `TEST_TIMEOUT_LIMIT`, `UNITTEST_STATECHANGE_TIMEOUT` are pre-existing helpers from `tests/unittest_util.h`, already used elsewhere in this file — no new/duplicated test infrastructure introduced.
- **Doxygen**: every new static helper and `TEST` case carries `@brief`/`@details`/`@return` as applicable; the CI-enforced per-function comment rule is satisfied.
- **Docs/architecture**: no public API, property, or documented behavior change; correctly, no `Documentation/` update accompanies this PR.
- **Size proportionality**: the functional fix is a single line; everything else is test code confined to `testTensorSplit`. Nothing outside the reported defect's blast radius is touched.
- **CI**: `DCO` and `CPP-Linter` pass; the remaining jobs (including the GBS x86_64 unit-test build that actually runs `testTensorSplit`) were still queued/pending at review time — not treated as a finding, but worth confirming green before merge per the repo's own merge criteria.

## Findings

No Critical, Major, or Minor issues found in this independent pass. The two earlier reviews' Nits (missing `tensorpick`-restart coverage, missing caps/dimension assertions on recreated pads) are both verifiably addressed in the current head (`restartKeepsTensorpick` exists; the counting probe now checks each buffer's size against its pad's current caps and both restart tests assert `out.mismatched == 0`).

- **Nit** (informational only, not a defect): none of the four new tests assert that the recreated pads keep their original names (`src_0`, `src_1`, ...) across the restart — only pad *count* (`numsrcpads`) and byte totals are checked. Given `gst_tensor_split_get_tensor_pad()` always names new pads `src_%u` from `split->num_srcpads` starting at 0 after `remove_src_pads()` resets it, this is very unlikely to regress silently, so this is optional and non-blocking.

## Verdict

**OK to merge as is**, once the still-pending CI jobs that execute `testTensorSplit` (e.g., GBS x86_64 unit-test build, meson+Valgrind) report green. The fix is a correctly-targeted, minimal one-line change confined to `tensor_split`, does not touch any other module, restores the `num_tensors`/`tensorseg` invariant without introducing new locking concerns, and ships with regression coverage at both the harness (unit) and full-pipeline (integration/replay) level that will fail CI if this defect — or its interaction with #4954 — regresses.

---
🤖 This review was generated by a separate Claude agent session and posted here by the PR author on its behalf.

myungjoo:

Thanks for the third review. The Nit is addressed in 754b5fdc:

- **Nit, pad names after the restart:** `restartKeepsTensorseg` now asserts that the recreated pads are `src_0` and `src_1`, and `restartKeepsTensorpick` that its one pad is `src_0`. I used plain `gst_object_unref()` rather than `gst_clear_object()`, which would have been the first use of a GStreamer 1.16 API in this repository.

Re-verified for this head:
- The restart tests fail on the unmodified element and pass with the fix.
- The tree merged with #4954 is conflict-free, builds, and passes all 19 `testTensorSplit` cases.
- clang-format 16, gst-indent, doxygen-tag and the strict `-Werror=sign-compare` compile are clean.
- `meson test` passes 22/23. The only failure is `unittest_filter_python3`, from the local NumPy environment.

myungjoo:
**This is an automated code review produced by a separate Claude agent (not by @myungjoo directly), relayed to this PR via the author's GitHub account.** This is a fresh, independent review of the current head (`754b5fdc`). I re-derived the bug and the fix from source rather than from the PR description or the three prior automated review rounds in this thread, and checked those prior rounds' claims against the code before relying on any of them.

## Scope

Reviewed the diff only: the one-line removal in `gst/nnstreamer/elements/gsttensor_split.c` (`gst_tensor_split_remove_src_pads`) and the four new `testTensorSplit` cases in `tests/nnstreamer_plugins/unittest_plugins.cc` (`restartKeepsTensorseg`, `restartKeepsTensorpick`, `restartRefusesSmallerTensor_n`, `replayPipeline`). Existing repository code and PR #4954 were read for context only.

## Verification performed

- **Root cause / fix, re-traced from source**: `GstTensorSplit::num_tensors` (`gst/nnstreamer/elements/gsttensor_split.h:57`) is written in exactly two places on this head: `gst_tensor_split_init()` (0) and `gst_tensor_split_set_property()`'s `PROP_TENSORSEG` case, which sets `tensorseg` and `num_tensors` together under `GST_OBJECT_LOCK` (`gsttensor_split.c:668-671`). Before the fix, `gst_tensor_split_remove_src_pads()` — called from `gst_tensor_split_change_state()` on `PAUSED_TO_READY` (`gsttensor_split.c:581`, outside any lock) and from `finalize()` — also zeroed `num_tensors`, while leaving the `tensorseg` array it counts untouched. On a subsequent chain call without re-setting `tensorseg`, `gst_tensor_split_chain()`'s `for (i = 0; i < num_tensors; i++)` loop (`gsttensor_split.c:491`, `num_tensors` read under the lock at line 477) ran zero times: `GST_FLOW_OK` was returned with no pad created and no buffer pushed, and the following EOS hit the `!split->srcpads` branch in `gst_tensor_split_event()` (`gsttensor_split.c:236-242`), producing "Got EOS before adding any pads." Removing the `num_tensors = 0` line is the correct, minimal fix — it restores the invariant that `num_tensors == tensorseg->len`, which is otherwise only ever changed atomically together in `set_property`, and it also removes the only unlocked write to a field every other accessor already reads/writes under `GST_OBJECT_LOCK`.
- **Regression scope**: grepped `gst/nnstreamer/elements/gsttensor_split.c` for every use of `num_tensors`/`num_srcpads`; confirmed `num_tensors` has no reader outside `chain()`, and `num_srcpads` (still correctly reset to 0 in the same function, since it counts recreated pads, not configured segments) is unrelated. No file outside `gsttensor_split.c` and the test file is touched by the diff. `tensor_demux`, the sibling element, derives its per-buffer count from the buffer itself rather than persisted state and is unaffected.
- **Interaction with #4954**: the current head of #4954 changes only `gst_tensor_split_get_tensor_pad()` (caps/pad-name indexing, `gsttensor_split.c:293-316`) and adds its own, differently-named test helpers; this PR touches only `remove_src_pads()`. The two diffs do not overlap.
- **Test correctness, traced against the actual code, not just the PR's prose**:
  - Confirmed `gst_pad_set_caps()` is called before `gst_element_add_pad()` in `gst_tensor_split_get_tensor_pad()` (`gsttensor_split.c:355-356`), i.e. before the buffer for that pad is ever pushed — so the new `_split_restart_count` probe's `gst_pad_get_current_caps(pad)` is guaranteed non-NULL when it fires, and its mismatch check is meaningful rather than accidentally-always-passing.
  - Byte math checks out: `1:4:4` = 16 B, `2:4:4` = 32 B (48 B total, `restartKeepsTensorseg`); `restartRefusesSmallerTensor_n`'s post-restart 12 B buffer (`3:2:2`) is correctly below the retained 48 B rule, so the refusal comes from the pre-existing `offset + size > src_info.size` check in `gst_tensor_split_get_splitted()` (confirmed on `main` already, unrelated to this PR or #4954, per the prior review thread's `git log -S` check, which I consider settled and did not re-derive).
  - `replayPipeline`'s `gst_parse_launch()` string omits `!` between `tensor_split` and `tensor_sink` on purpose, so only the test's own `pad-added` handler links `src_0`; with `tensorseg=1:4:4,2:4:4` this leaves `src_1` unlinked. Traced `gst_tensor_split_combine_flows()` (`gsttensor_split.c:381-400`): since `src_0`'s push returns `GST_FLOW_OK`, the walk over all pads finds a non-`NOT_LINKED` result and `chain()` still returns `GST_FLOW_OK` overall, so `src_1` being unlinked does not make this test flaky or wrong.
  - All four tests rely on GstPad's sticky-event-clearing on deactivation (hence re-sending `stream-start`/caps after each `READY`/`PLAYING` cycle) and on `GST_PAD_PROBE_DROP` making `gst_pad_push()` report `GST_FLOW_OK` without a linked peer — both standard, correctly-applied GStreamer idioms.
  - Test helper reuse verified: `count_output`, `setPipelineStateSync`, `TEST_TIMEOUT_LIMIT`, `UNITTEST_STATECHANGE_TIMEOUT`, `UNUSED` are all pre-existing (`tests/unittest_util.h`, `gst/nnstreamer/include/nnstreamer_util.h`), no duplicated test infrastructure. `gst_tensors_config_from_caps`, `gst_harness_new_with_padnames`, `gst_tensor_parse_dimension`, `gst_tensors_caps_from_config`, `gst_harness_create_buffer` are existing, already-used-elsewhere APIs in this file; signatures match usage.
- **Doxygen**: every new static helper and `TEST` case carries a `@brief` (and `@details`/`@return` where applicable), satisfying the CI-enforced per-function comment rule.
- **Docs/architecture**: no public API, property, or documented behavior change; correctly, no `Documentation/` update accompanies this PR.
- **Size proportionality**: the functional fix is a single line; everything else is test code confined to `testTensorSplit`. Nothing outside the reported defect's blast radius is touched.
- **Prior review thread**: independently confirmed both nits raised in the first two review rounds (missing `tensorpick`-restart coverage, missing caps assertions on recreated pads, missing pad-name assertions across restart) are addressed in this head — `restartKeepsTensorpick` exists, the counting probe checks each buffer's size against its pad's live caps and both restart tests assert `out.mismatched == 0`, and `restartKeepsTensorseg`/`restartKeepsTensorpick` assert the recreated pads are named `src_0`/`src_1`. I did not find anything the prior rounds missed.

I did not attempt a full local build in this pass (a build tree was not present in the reviewed worktree and standing one up was out of scope for this review); the verification above is by direct code tracing against the actual `gsttensor_split.c` and `unittest_plugins.cc` sources at this head, not by re-running the test suite. Per the review instructions for this pass, CI still being in progress on this head is not reported as a finding.

## Findings

No Critical, Major, or Minor issues found.

- **Nit** (informational only, not a defect): none of the four tests assert the destroyed/recreated pads are actually distinct `GstPad` objects from the pre-restart ones (only names and count are checked). Given `gst_element_remove_pad()`/`gst_pad_new_from_static_template()` in `remove_src_pads()`/`get_tensor_pad()` make this essentially guaranteed by construction, this is optional and non-blocking.

## Verdict

**OK to merge as is.** The fix is a minimal, correctly-targeted one-line change confined to `tensor_split`, restores the `num_tensors`/`tensorseg` invariant without introducing new locking concerns, touches no other module, and ships with regression coverage at both the harness (unit) and full-pipeline (integration/replay) level that will fail CI if this defect regresses. This independent pass agrees with the three prior automated reviews in this thread that the reported nits have all been addressed and found no new issues.

---
🤖 This review was generated by a separate Claude agent session and posted here by the PR author on its behalf.

myungjoo:

Thanks for the fourth review. On its one Nit, whether the recreated pads are distinct objects: no change. As the review notes, `gst_element_remove_pad()` drops the old pad and `gst_pad_new_from_static_template()` creates a new one. An assertion there would be testing GStreamer rather than this change. What the restart depends on is already asserted: the pad count and names, one `no-more-pads` per run, the bytes pushed, and that caps and buffer sizes agree.

No further changes, so 754b5fdc stays the head under review.

Issue #4947, opened 2026-09-10
[Test] Log-assertion tests do not prove their handler is live, so they can become silent no-ops


myungjoo:
### Summary

Several unit tests assert on the presence or absence of a GLib log message by installing a log handler, but none of them proves the handler actually matched. A handler that stops matching turns the assertion into a no-op, and the test keeps passing for the wrong reason.

This is test hardening, not a defect in shipped code. Filed as a follow-up from the review of #4942, where the same pattern was fixed for `tensor_crop`.

### Why the pattern is fragile: the choice of handler is domain-sensitive

Measured while working on #4942:

- **`g_log_set_default_handler()` does not see GStreamer-domain criticals.** GStreamer installs its own handler for the `GStreamer` domain, so the default handler never receives them. A test in #4942 originally used `g_log_set_default_handler()` and failed to record a `GStreamer-CRITICAL` that `gst_pad_push (pad, NULL)` demonstrably emitted; it had to be changed to `g_log_set_handler ("GStreamer", ...)`.
- **GLib installs no handler for its own domain**, so a `GLib-CRITICAL` does reach the default handler.

So whether one of these tests observes anything depends on which domain the guarded call happens to log to — a property of the code under test, not of the test.

### The one existing site, and its measured state

`tests/nnstreamer_plugins/unittest_plugins.cc:11444` `testTensorSplit.finalizeWithoutTensorseg` (added in #4925):

```c
  tensor_split_logged_critical = FALSE;
  old_handler = g_log_set_default_handler (_record_critical, NULL);
  gst_object_unref (split);
  g_log_set_default_handler (old_handler, NULL);

  EXPECT_FALSE (tensor_split_logged_critical);
```

**This test is live today** — verified, not assumed. Removing the guard in `gst_tensor_split_finalize()` (`gst/nnstreamer/elements/gsttensor_split.c:209-210`) so that `g_array_unref (NULL)` is reached makes the case fail:

```
Value of: tensor_split_logged_critical
[  FAILED  ] testTensorSplit.finalizeWithoutTensorseg
```

It works because `g_array_unref()`'s `g_return_if_fail` logs on the **GLib** domain, which the default handler does receive. Nothing in the test states that dependency, and nothing would catch it going away: if the finalize path ever releases something whose misuse logs on the `GStreamer` domain instead, the assertion silently stops observing and the test passes unconditionally.

[...]

PR #4963, opened 2026-09-11
[Test/Split] Verify the critical log handler

33modeling:
### Summary

`testTensorSplit.finalizeWithoutTensorseg` checks that destroying an unconfigured `tensor_split` emits no critical. Its recorder had no positive check, so disabling the recorder still left the test passing.

This change installs a handler for the `GLib` domain used by `g_array_unref()`, emits a synthetic critical to verify that the recorder is live, and resets it before finalization. The handler is removed afterwards without replacing the process-wide default handler. The fatal mask is relaxed only for the intentional probe and restored before the operation under test.

Related to #4947. This covers the `tensor_split` test; the LiteRT sites remain separate. The existing test detects the current finalizer regression already; this change guards against the recorder itself silently becoming ineffective.

### Validation

- Linux x86_64, GCC 13.3, GLib 2.80, GStreamer 1.24; optional ML backends disabled.
- Full `unittest_plugins`: 229 passed. The local `build-committer` directory was also exposed as `build` for the existing custom-filter fixture path.
- `testTensorSplit.*`: 9 passed.
- Focused case with `G_DEBUG=fatal-criticals`: passed.
- Mutation check: disabling the recorder passes on the original test and fails at the new self-check after this change.
- Mutation check: removing the `tensorseg` null guard fails at the finalization assertion. All mutations were reverted.
- clang-format 16 on the changed lines and `git diff --check`: passed.

### How to evaluate

```sh
CC=gcc CXX=g++ meson setup build -Dauto_features=disabled -Denable-test=true -Dvideo-support=enabled -Daudio-support=enabled
meson compile -C build -j 2
meson test -C build unittest_plugins --test-args='--gtest_filter=testTensorSplit.*' --print-errorlogs
G_DEBUG=fatal-criticals meson test -C build unittest_plugins --test-args='--gtest_filter=testTensorSplit.finalizeWithoutTensorseg' --print-errorlogs
```

Prepared with OpenAI Codex assistance.

### DIFF
--- a/tests/nnstreamer_plugins/unittest_plugins.cc
+++ b/tests/nnstreamer_plugins/unittest_plugins.cc
@@ -12159,15 +12159,26 @@ _record_critical (const gchar *domain, GLogLevelFlags level,
 TEST (testTensorSplit, finalizeWithoutTensorseg)
 {
   GstElement *split = gst_element_factory_make ("tensor_split", NULL);
-  GLogFunc old_handler;
+  guint handler;
+  GLogLevelFlags fatal_mask;
 
   ASSERT_TRUE (split != NULL);
   gst_object_ref_sink (split);
 
+  /* g_array_unref() reports invalid arguments in the GLib domain. */
+  handler = g_log_set_handler ("GLib",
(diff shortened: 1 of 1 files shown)


myungjoo (approved):
LGTM. Please fix the spellcheck.

PR #4966, opened 2026-09-14
[CI] Make the shellcheck check fail on shellcheck errors

### LINKED ISSUE #4965: [CI] The shellcheck check cannot fail: shellcheck never receives the file
## Summary

`.github/workflows/static.check.scripts/shellcheck.sh` never checks anything. It feeds each file to shellcheck on stdin:

```bash
cat $file | $shell_syntax_analysis_sw $shell_syntax_analysis_rules > ${shell_syntax_check_result}
line_count=`cat ${shell_syntax_check_result} | wc -l`
```

shellcheck reads stdin only when `-` is given as the file argument. Without it, it prints `No files specified.` and its usage to **stderr**, writes nothing to stdout, and exits 3. `line_count` is therefore always 0, and every `.sh` file is reported as passed. The check has been inert.

## Reproduction

From the `Static checks` run of #4956 (run 34467386272, shellcheck 0.9.0 from the `ubuntu-latest` apt archive), step `/Checker/ shellcheck for shell scripts`:

```
(.github/workflows/static.check.scripts/pylint.sh) file is a shell script file with the 'shell script' text format.
No files specified.

Usage: shellcheck [OPTIONS...] FILES...
...
##[group]shellcheck result of .github/workflows/static.check.scripts/pylint.sh
##[endgroup]
shellcheck: passed. file name: .github/workflows/static.check.scripts/pylint.sh There are 0 lines.
```

The same file checked properly (`ubuntu:24.04`, shellcheck 0.9.0):

| invocation | exit | stdout lines |
|---|---|---|
| `cat pylint.sh \| shellcheck -s bash` (what the script does) | 3 | 0 |
| `shellcheck -s bash pylint.sh` | 1 | findings (SC2006, SC2086, ...) |

Running the current script over all 101 tracked `*.sh`/`*.bash` files (excluding `obsolete/` and `external/`) reports `There are 0 lines.` for all 101.

## Why repairing the invocation alone is not enough

1. **Threshold.** Even with output captured, the step fails only when shellcheck prints more than **9,000 lines** for a single file (`# TODO: 9,000 is declared by heuristic method`). No file in the tree comes close, so it would still pass everything.
2. **Severity.** Gating on "any finding" would block most PRs that touch a shell script. Files failing `shellcheck -s bash -S <severity>` today, out of 101:

| severity | files that fail |
|---|---|
| `error` | **4** |
| `warning` | 69 |
| `info` | 86 |
| `style` | 87 |

Most warning-level hits come from SSAT test conventions: SC2034 (146), SC2154 (55), and SC1090 (53, from sourcing `ssat-api.sh`), plus SC2010 (43).

The four `error`-level findings:

- `.github/workflows/static.check.scripts/cppcheck.sh:44`: `2=0` (SC2270). This is not an assignment; bash runs it as a command named `2=0`. It is unreachable today only because `static.check.yml` always passes the second argument.
- `tools/development/reversion.sh:8`: `exit -1` (SC2242).
- `tests/nnstreamer_decoder/runTest.sh:116` and `tools/development/gen_mediapipe_libs.sh:21,34`: unquoted array expansions (SC2068).

Only changed files are checked, so a gate at `error` would block only PRs that touch those four files until they are fixed.

## Suggested direction

Pass the file directly and gate on shellcheck's exit status at an agreed severity, for example:

```bash
if ! shellcheck -s bash -S error "$file" > ${shell_syntax_check_result} 2>&1; then
  failed=1
fi
```

A non-zero exit also covers a misinvocation like the current one (exit 3), which would then fail loudly instead of passing. The four files above would need fixing in the same change, and a self-test in the style of `test_pylint.sh` / `test_prohibited_words.sh` would keep the checker from going inert again.

## Context

Found while fixing #4919 (the pylint check could not fail). Same family as #4908 (per-function Doxygen check silently disabled).

myungjoo:
Fixes #4965.

## Problem

`shellcheck.sh` ran `cat $file | shellcheck -s bash`. Without `-` as the file argument shellcheck reads nothing, prints `No files specified.` to stderr and exits 3, so the captured output was always 0 lines and every script passed. The step also failed only above 9,000 output lines per file. #4956's Static checks run shows it: both changed `.sh` files logged `No files specified.` and then `passed ... There are 0 lines.`

## Change

- **`shellcheck.sh`**: `shellcheck -s bash -S error "$file"`, fail on a non-zero exit. The 9,000-line threshold is gone, and a misinvocation (exit 2-4) now fails instead of passing. The `file` precondition accepts any text, not only what `file` calls a shell script, so a `.sh` without a shebang (reported as plain `ASCII text`) is no longer skipped. The failure message is reworded.
- **The four files with error-severity findings**, fixed so the next PR touching them is not blocked. Only `reversion.sh` changes observable behaviour:
  - `cppcheck.sh:44` `2=0` (SC2270) ran a command named `2=0` when arg 2 was missing; now `set -- "$1" 0`.
  - `reversion.sh:8` `exit -1` (SC2242, i.e. 255) → `exit 1`: the wrong-argument-count exit status changes from 255 to 1, both failures; nothing in the tree runs it (only `tools/development/README.md` mentions it).
  - `nnstreamer_decoder/runTest.sh:116`, `gen_mediapipe_libs.sh:21,34`: quoted array expansions (SC2068).
- **`test_shellcheck.sh`** plus a guarded `static.check.yml` step, following `test_prohibited_words.sh`:
  - Must fail: an error-severity finding, including as the second file of the list and in a script without a shebang. The log must show the SC code and name the failing file, so a missing shellcheck cannot pass for a finding.
  - Must pass: a warning/info/style-only script (the test also asserts it really has a warning), a non-shell file, an empty list.

Why `error`: files failing `shellcheck -s bash -S <sev>` today, out of 101 tracked `.sh`/`.bash` files, are 4 at `error`, 69 at `warning`, 86 at `info` and 87 at `style`. The warning-level failures come mostly from SSAT `runTest.sh` conventions.

## Verification (ubuntu:24.04, shellcheck 0.9.0, the version CI installs)

| check | result |
|---|---|
| `test_shellcheck.sh` | 11/11 pass |
| `test_shellcheck.sh` against the **old** `shellcheck.sh` | all 4 failure-direction checks fail → would have caught #4965 |
| `test_shellcheck.sh` with shellcheck removed from `PATH` | fails (every log check fails) |
| new `test_shellcheck.sh` against the previous head (e222c487) | the no-shebang case fails (exit and log) → the review nit is pinned |
| `main` 611c369f + #4956 + this PR, in either order | identical tree, no conflict; all 103 tracked `.sh` pass, both self-tests pass |
| new `shellcheck.sh` over all 101 tracked files | exit 0, 101 passed |
| new `shellcheck.sh` over the 4 files as they are on `main` | exit 1, all 4 reported |
| `cppcheck.sh` without arg 2, old vs new | same output except the old `2=0: command not found` line |
| `reversion.sh` with no args | exit 255 → 1 |
| decoder `VIDEO_FORMAT` loop, old vs new | identical 13 formats |
| `gen_mediapipe_libs.sh` with a gcc shim over fake `.o` files, old vs new | identical object arguments, both libs produced |
| doxygen-tag, shellcheck, newline, executable, hardcoded-path, misspelling, prohibited-words, sloccount on the changed files; existing self-tests | all exit 0 |

The new gate caught a real trap while this PR was being written: a comment line starting with `# shellcheck` is parsed as a directive, which fails at error severity (SC1072/SC1073). `test_shellcheck.sh`'s header was reworded for it.


myungjoo:

Reviewed the diff (`.github/workflows/static.check.scripts/{shellcheck.sh,cppcheck.sh,test_shellcheck.sh}`, `.github/workflows/static.check.yml`, `tests/nnstreamer_decoder/runTest.sh`, `tools/development/{gen_mediapipe_libs.sh,reversion.sh}`) against issue #4965, the rest of the repo, and the live "Static checks" run for this PR (run 34820561951, job 103900957132).

Yes. The root cause in #4965 was that `shellcheck.sh` piped each file into `shellcheck` (`cat $file | shellcheck -s bash`) without a `-` file argument, so shellcheck read nothing, printed `No files specified.` to *stderr* (which wasn't even captured), exited 3, and `line_count` was always 0. The fix passes `"$file"` directly, captures both streams (`2>&1`), and gates on `shellcheck`'s own exit status instead of a fixed line count — which also removes the pointless "the check only fails above 9,000 output lines" threshold the issue called out as a second bug. I confirmed against the live run that the new step actually invokes shellcheck per file and reports `passed`/`failed` correctly (log lines at 08:01:37–38 in the linked job).

`-S error` as the gating severity is a reasonable, well-justified choice: the PR body's own table (4 error / 69 warning / 86 info / 87 style out of 101 tracked scripts) shows that gating on anything looser would immediately block most PRs touching a shell script, mainly from SSAT `runTest.sh` conventions. This matches the issue's own suggested direction.

Low. The four "collateral" fixes are exactly the four error-severity files identified in the issue, and are the minimum needed so the corrected gate doesn't immediately block unrelated future PRs (only touched files are checked, so nothing outside these four was required). I checked each:

- `cppcheck.sh:44` (`2=0` → `set -- "$1" 0`, SC2270): correct fix for a real bug (the old line ran a non-existent command named `2=0` instead of defaulting `$2`). Confirmed via `static.check.yml:123` that the caller always passes a second argument (`cppcheck.sh $changed_file_list 0`), so this path was unreachable in CI before and after — no behavior change for the pipeline, only for a manual/no-arg invocation.
- `tests/nnstreamer_decoder/runTest.sh:116` and `tools/development/gen_mediapipe_libs.sh:21,34` (unquoted `${arr[@]}` → quoted, SC2068): behavior-preserving since none of the array elements (video format names, `.o` object paths) contain whitespace/glob characters, as the PR body states.
- `tools/development/reversion.sh:8` (`exit -1` → `exit 1`, SC2242): **this one is not behavior-preserving.** I verified empirically that `bash -c 'exit -1'; echo $?` yields `255` (bash takes the exit code mod 256), so the script's actual exit status changes from 255 to 1. The PR's commit message correctly notes "`exit -1` is exit 255 (SC2242)" right before this change, but the PR description's summary bullet still frames all four fixes as "None changes behaviour," which is inaccurate for this one. Practically low-risk: `reversion.sh` is only referenced from `tools/development/README.md` (a manual maintainer tool), not invoked by any other script, workflow, or `meson.build` in the tree, so nothing branches on the specific exit code today. Severity: **Low** — worth a one-line correction to the PR description, not a code change.

## Scope / size

Proportionate. +136/-13 across 7 files; the bulk (111 lines) is the new self-test, which mirrors the existing `test_prohibited_words.sh` / `test_debian_control.sh` pattern almost line-for-line (same `expect_checker`/`expect_log` skeleton, same plain `#`-comment style — I diffed the skeletons and they match the established convention, not a new one). No unrelated modules are touched beyond the four files required to unblock the new gate.

## Test coverage / future-regression detection

Good, and this is the strongest part of the PR. `test_shellcheck.sh` pins both directions that matter:
- An error-severity finding must fail the checker (including when it's the *second* file in the list, which specifically guards the loop/accumulation logic), and the log must actually contain the SC code and file name — so a checker that fails for an unrelated reason (e.g. `shellcheck` missing) cannot pass this case by accident.
- Warning/info/style-only findings, a non-shell file, and an empty file list must all pass, with a fixture self-check (`shellcheck -s bash -S warning` must find something) that keeps the "must pass" fixture from silently becoming a no-op case.
- The PR description states this self-test fails in all four directions against the pre-fix `shellcheck.sh`, i.e. it would have caught #4965. I could not independently install shellcheck on this review box to re-verify that claim locally, but I did verify the *new* code and self-test end-to-end on the PR's own live CI run, where the new `/Checker/ shellcheck self-test` step logs all 9 `PASS:` lines and the main `/Checker/ shellcheck for shell scripts` step correctly reports `passed`/`failed` per file.

One gap: the `cppcheck.sh` SC2270 fix has no equivalent regression test (no `test_cppcheck.sh`), unlike the shellcheck.sh fix. The repo's test policy (per CLAUDE.md) asks for a reason when a bug fix ships without a reproducing test; the PR body does document manual verification for it in its results table, but doesn't explicitly say why an automated test wasn't added. Given the bug path is currently unreachable from CI (arg 2 is always passed), this is low-severity and does not need to block the PR. Severity: **Low/Nit**.

## Will a future regression here be caught by CI?

Yes for the shellcheck gate itself — a future PR that re-breaks `shellcheck.sh`'s invocation (e.g. reverting to the piped form, or gating on the wrong exit code) would fail `test_shellcheck.sh`'s "error-level finding" or "warning/info/style findings only" cases, and that step is wired into `static.check.yml` with the same removal-guard pattern as the sibling self-tests (`test_debian_control.sh`, `test_meson_version.sh`, `test_doxygen_tag.sh`), so deleting the self-test script without removing its step also fails the job.

## Documentation / architecture / API

None needed. This is a CI-script-only change; no public API or architecture is affected.

## Nit (non-blocking, out of scope for this PR)

`tools/development/gen_mediapipe_libs.sh:21` and `:34` — only the array-quoting was touched here, but while reading the surrounding lines I noticed `$($CC ...)` on line 21 is used as a bare statement (not assigned), so the following `if [ $? -ne 0 ]` on line 22 checks the exit status of executing the (usually empty) captured stdout as a command, not `$CC`'s actual exit status — this pre-dates the PR and isn't something this change should fix, just flagging for a future cleanup.

## Verdict

Verdict: OK to merge (with the reversion.sh "no behaviour change" wording nit noted above; does not need to block).

myungjoo:
[Response to the automated review above, relayed on behalf of the author by a Claude agent]

Thanks. Taking the three notes in turn; amended as e222c487 (commit message and PR body only, no code change).

**[Low] `reversion.sh` does change behaviour.** Correct. `exit -1` is 255, and it is now 1. The commit message and PR body said none of the four fixes changes behaviour; both now say that only `reversion.sh` does: its wrong-argument-count exit status goes from 255 to 1. Both are failures, and nothing in the tree runs the script (only `tools/development/README.md` mentions it).

**[Low/Nit] no dedicated regression test for the `cppcheck.sh` fix.** Not adding one. The gate this PR restores is that test: reintroducing `2=0` (or any other error-severity construct) in `cppcheck.sh` makes `shellcheck.sh` fail on that PR with SC2270, which is exactly how the line was found. A separate cppcheck.sh test would exercise a path that `static.check.yml` never takes, since it always passes the second argument.

**[Nit] `gen_mediapipe_libs.sh:21`, `$?` after a bare `$($CC ...)`.** This one does check `$CC`'s exit status. When a simple command consists only of expansions and nothing is left to run, bash takes the exit status of the last command substitution: `bash -c '$(false); echo $?'` prints 1, and `$(sh -c "exit 3"); echo $?` prints 3. So no change there.

myungjoo:

This is a re-review after the author's response to the prior automated review. I re-read the prior review, the author's response, confirmed the code diff is unchanged since then (only the commit message / PR body were amended, in e222c487), and independently re-verified the three points raised earlier.

## Disposition of the earlier findings

**1. "`reversion.sh` does change behaviour" (accepted by the author, doc-only fix).** Verified independently: `bash -c 'exit -1'; echo $?` → `255` on this machine, matching the author's claim. The PR body and commit message now correctly single out `reversion.sh` as the one behaviour-changing fix (255 → 1, both failures, nothing in the tree branches on the specific code — only `tools/development/README.md` references the script). This adequately resolves the finding; no further action needed.

**2. "no dedicated regression test for the `cppcheck.sh` fix" (declined).** I accept this. Re-reading `cppcheck.sh:43-45`, the fix is `set -- "$1" 0"` replacing `2=0`, i.e. exactly the pattern shellcheck's SC2270 targets (confirmed via the shellcheck wiki: SC2270 fires on `$2=value`-shaped assignments to positional parameters). Since `cppcheck.sh` is itself a `.sh` file covered by the newly-fixed `shellcheck.sh` gate (`-S error`, gating on exit status), any future reintroduction of this exact bug class in `cppcheck.sh` would be caught by the restored gate, not a separate `test_cppcheck.sh`. That is a real, structural argument, not an excuse — the PR's own verification table shows the four error-severity files (which includes `cppcheck.sh`) are exactly the four fixed here, so the mechanism is not hypothetical. Reasonable to decline.

**3. "`gen_mediapipe_libs.sh:21`, `$?` after a bare `$($CC ...)`" (declined).** I accept this too. Verified independently:
```
$ bash -c '$(false); echo $?'      # -> 1
$ bash -c '$(sh -c "exit 3"); echo $?'   # -> 3
```
This confirms the POSIX/bash rule the author cites: when a simple command reduces to only expansions (no output to word-split into a command name), `$?` is the exit status of the last command substitution — so `$?` on the line after `$($CC ...)` does reflect `$CC`'s own exit status, as long as `$CC` doesn't print anything to stdout on success (true for a normal compiler/linker invocation). Correct as stated, and in any case this line's `$?` pattern pre-dates the PR and isn't touched by the array-quoting change under review.

## Review of the diff as a whole (unchanged since the prior review)

I independently walked the current diff (`shellcheck.sh`, `cppcheck.sh`, `test_shellcheck.sh`, `static.check.yml`, `tests/nnstreamer_decoder/runTest.sh`, `tools/development/gen_mediapipe_libs.sh`, `tools/development/reversion.sh`) rather than only trusting the prior pass, and concur with its conclusions:

- **Problem solved**: yes. `shellcheck.sh` now passes `"$file"` as an actual argument (previously piped with no `-`, so shellcheck read nothing and always "passed"), captures both streams, and gates on shellcheck's real exit status instead of a meaningless 9,000-line-count threshold. This is precisely the bug in #4965.
- **Regression risk**: low. The four collateral fixes are exactly the four error-severity files the new gate would otherwise immediately trip on, and I re-traced each one (`cppcheck.sh`'s `set -- "$1" 0` preserves `$2`'s downstream use at line 81; `runTest.sh`/`gen_mediapipe_libs.sh` array-quoting is a no-op for the actual element contents; `reversion.sh`'s exit-code change is a failure→failure change in an unreferenced manual tool). Nothing outside these four files, plus the workflow and test additions, is touched.
- **Scope**: proportionate to the issue; the new self-test step follows the established `test_prohibited_words.sh`/merge-ref-guard convention used by every sibling self-test in this same file, verified by direct comparison (`static.check.yml:283-295` vs `296-312`) — same guard pattern, same phrasing, same structure.
- **Test coverage**: `test_shellcheck.sh` pins the two directions that matter (an error-level finding must fail, including as the second file in the list, with the SC code and filename actually present in the log so a broken invocation can't pass by accident; warning/info/style-only findings, a non-`.sh` file, and an empty list must all pass) and the "must-pass" warning fixture self-checks that it really does carry a shellcheck warning, so it can't silently degrade into a no-op case.
- **Future-regression detection / CI gating**: yes — the self-test step is wired into `static.check.yml` with the same delete-guard as its siblings, so a future PR that reverts the invocation fix, or removes the self-test without removing its step, fails CI.
- **Docs/architecture/API**: none needed; this is CI-script-only.
- Confirmed on the live PR: `Static checks` passes on the current head (e222c487), and the PR body's own verification table (old-vs-new invocation, old-vs-new self-test results, full 101-file sweep, per-file before/after behaviour) is internally consistent with the diff.

No new issues found beyond what was already raised and resolved.

Verdict: OK to merge

myungjoo:
[Response to the myungjoo-bot review, relayed on behalf of the author by a Claude agent]

Thanks. The NIT (a `.sh` without a shebang skips the gate) is fixed in this PR, in amend 08d22759, rather than left as a follow-up.

- `shellcheck.sh` now matches `text` from `file` instead of `shell script`. The `case` still restricts the gate to `*.sh | *.bash`, so non-shell files are skipped as before. This is the same change #4956 made to its ASCII-only filter.
- `test_shellcheck.sh` adds `no_shebang.sh`, which is the error fixture with its first line removed. It must fail, and its `failed. file name:` line must appear in the log. The test first asserts that `file` no longer calls the fixture a shell script, so the case cannot turn vacuous.

Verified in `ubuntu:24.04` (shellcheck 0.9.0):
- `test_shellcheck.sh`: 11/11 pass.
- Negative control: the new self-test against e222c487's `shellcheck.sh` fails both no-shebang checks.
- I also cherry-picked #4956 and this PR onto current `main` (611c369f) in both orders. Both give the same tree with no conflict. On that tree, all 103 tracked `.sh` files pass the widened filter, and both self-tests pass.

myungjoo:

This is a re-review after `myungjoo-bot`'s approval of e222c487 (one NIT) and the author's response amending the head to 08d22759. I diffed e222c487 → 08d22759 in isolation, then re-read the whole PR diff against main (120d2d17) and issue #4965.

## Is the prior NIT resolved?

**Yes.** The isolated change since e222c487 is exactly two hunks:

- `shellcheck.sh:71`: `` file $file | grep "shell script" `` → `` file $file | grep "text" ``, plus a reworded log line (`"file is a shell script file."`, dropping the now-inaccurate `"with the 'shell script' text format"`).
- `test_shellcheck.sh`: a new `no_shebang.sh` fixture (the error fixture with its `#!/bin/bash` line stripped), a self-check that `file` no longer classifies it as `"shell script"` (so the new case can't go vacuous if `file`'s wording ever changes), and an `expect_checker 1` / `expect_log` pair asserting the checker fails on it and logs `failed. file name: .../no_shebang.sh`.

I verified independently with the local `file` 5.46 (GNU file, same family as the `ubuntu:24.04` one the author and `myungjoo-bot` used): a `.sh` with a shebang reports `Bourne-Again shell script, ASCII text executable`; the same content minus the shebang reports plain `ASCII text`. The old `grep "shell script"` filter excluded the no-shebang case; the new `grep "text"` includes it, since `"ASCII text"` contains `"text"`. This matches the PR's own claim and closes the hole `myungjoo-bot` flagged.

## Does widening to `grep "text"` make the gate check something it should not, or miss anything it should?

**Neither, and the reasoning is structural, not just empirical:**

- **Nothing new is checked beyond `.sh`/`.bash`.** The `file | grep` test is only the outer pre-filter; the actual shellcheck invocation is gated a second time by the `case $file in *.sh | *.bash)` block right below it, which this PR does not touch. Any file `file(1)` now newly classifies as `"text"` but that isn't a `.sh`/`.bash` file still falls through the `case` with no action. So the widened filter cannot cause the gate to lint, e.g., a `.py` or `.md` file it shouldn't.
- **Nothing that used to match stops matching.** `"shell script"` outputs from `file(1)` (e.g. `"Bourne-Again shell script, ASCII text executable"`) always also contain the substring `"text"` — I confirmed this against real `file` output above — so `grep "text"` is a strict superset of `grep "shell script"` for this purpose. No previously-covered `.sh`/`.bash` file becomes newly skipped.
- **The one behavioral difference is exactly the intended one:** a `.sh`/`.bash` file `file(1)` reports as plain text (no shebang, or a shebang `file` doesn't recognize) is now included where it was skipped before. That is precisely the gap identified in the prior review.
- I also grepped the rest of `.github/workflows/static.check.scripts/` for similar `file | grep` filters (`grep -rn 'grep "shell script"\|grep "text"\|grep "ASCII"'`) — `shellcheck.sh` is the only file using this pattern, so there's no sibling filter left inconsistently narrow.

One theoretical residual gap, pre-existing and not introduced by this change: a `.sh` file that `file(1)` classifies as something containing neither `"text"` nor `"shell script"` (e.g., unusual encoding reported as `"data"`) would still be skipped under both the old and new filter. The new filter only shrinks this gap (by admitting the "plain ASCII text" case), it doesn't create or widen it. Not worth a nit against this PR.

## Rest of the diff (unchanged since e222c487, already reviewed twice; re-confirmed here)

- **Problem solved**: yes, per #4965 — `shellcheck.sh` now passes `"$file"` as a real argument (previously piped with no `-`), captures both streams, and gates on shellcheck's exit status at `-S error` instead of a 9,000-line threshold.
- **Regression risk**: low. `cppcheck.sh`'s `set -- "$1" 0` preserves `$1` and gives `$2` the same effective default as before (only the spurious `2=0: command not found` line disappears); the SC2068 quoting in `runTest.sh`/`gen_mediapipe_libs.sh` is a no-op since no array element contains whitespace/glob characters; `reversion.sh`'s `exit -1` → `exit 1` changes the numeric exit code (255 → 1) but both are failure exits, and the PR body/commit message now correctly says so, and nothing in the tree branches on the specific code (only `tools/development/README.md` references the script).
- **Scope**: proportionate — 7 files, +147/-15 over the whole PR (main...08d22759), with the bulk being the 120-line self-test that follows the existing `test_prohibited_words.sh`-style convention. Nothing outside the four error-severity files plus the checker/workflow/test is touched.
- **Test coverage**: `test_shellcheck.sh` now covers the fail path (with/without a shebang, and as either list position), the pass path (warning/info/style-only, with a self-check that the fixture really carries a warning), the non-shell-file skip, and the empty-list no-op. This is the strongest part of the PR and is what would catch a future regression of this exact class.
- **CI enforcement**: the new self-test step in `static.check.yml` carries the same merge-ref delete-guard as the sibling self-tests, so deleting `test_shellcheck.sh` without removing its step fails CI, and a future re-break of `shellcheck.sh`'s invocation or filter fails the self-test itself. Current `gh pr checks 4966` shows `Static checks` passing on head 08d22759.
- **Docs/architecture/API**: none needed; CI-script-only change.

I was not able to install/run `shellcheck` on this review machine, so I did not independently re-execute `test_shellcheck.sh`; I verified the `file`-classification claim directly (shown above) and cross-checked the rest against the PR's own verification table and the live CI run, which is consistent with the diff.

## Findings

None blocking. No new issues found beyond what the prior two review passes already raised, and the one open NIT from `myungjoo-bot`'s approved review is resolved by this amendment.

## Verdict

Verdict: OK to merge

Issue #3822, opened 2022-07-04
NNstreamer tensor_filter framework=python3 plugin HALTS when using cv2, pytorch, etc. (threads' race condition/deadlock? need to release GIL?)


beto-gulliver:
# Issue Description

NNstreamer tensor_filter framework=python3 plugin HALTS when using cv2, pytorch, etc. (threads' race condition/deadlock? need to release GIL?)

I'm trying to implement some of my tensor_filter plugin using
framework=python3.

My tensor_filter plugins run OK, invoke(), when using pure 'numpy' functions
calls but somehow they get HALTED if I try to use opencv(cv2), pytorch, or other threaded function calls.

(see below for details: I've just modified 'passthrough.py' from the tests)
    USE_CV2 = True # [NG]
    USE_CV2 = False # [NG]

My guess is that the tensor_filter framework=python3 plugin has a
threads race condition or deadlock with the threads in opencv(v2), pytorch, etc.

Any ideas on how to fix this? Thanks in advance

I can see a Py_LOCK() / Py_UNLOCK() at PYCore::run(...) :

    nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc:519
    nnstreamer/ext/nnstreamer/tensor_filter/tensor_filter_python3.cc:574

Maybe we also need to use something similar to this :

Releasing GIL(Global Interpreter Lock) and mixing threads from C and Python
	https://www.geeksforgeeks.org/releasing-gil-and-mixing-threads-from-c-and-python/
  Py_BEGIN_ALLOW_THREADS
      
    // Threaded C code. 
    // Must not use Python API functions
    ...
  Py_END_ALLOW_THREADS

Expected Result
===============

   NNstreamer tensor_filter framework=python3 plugin SHOULD NOT HALT when using cv2, pytorch, etc

Further Information
======================

    Ubuntu 20.04.3 LTS

    ii  nnstreamer-python3:amd64               2.1.1.0-0~202206230823~ubuntu20.04.1     amd64        NNStreamer Python Custom Filter Support (3.x)

    python --version --> Python 3.8.10
    numpy               1.17.4
    opencv-python       4.6.0.66

How to Reproduce
======================

DETAILS

passthrough_cv2.py (modified from: nnstreamer/tests/test_models/models/passthrough.py)

```
python
import numpy as np
import nnstreamer_python as nns

D1 = 3
D2 = 280
D3 = 40
D4 = 1

[...]

PR #4968, opened 2026-09-15
[Python3] Release the per-frame lists and stop growing sys.path

myungjoo:
This fixes item **F9** of #4920 and adds python3 sub-plugin leak tests that run without Valgrind.

## What was leaking (measured on main `611c369f`)

- **tensor_decoder python3, per frame.** `PYDecoderCore::decode()` built `raw_data` (the input tensors) and `in_info` (their `TensorShape`s) for every frame and released neither. When the output buffer could not be mapped, the script's result leaked too.
  - With two small tensors this cost 2.3 KB, 11 allocator blocks and 4 lists per frame.
  - Four decoders reached **+450 MB after 200 000 frames**.
- **`PyTensorShape_New()`, per call.** It never released the argument tuple, or the dimension list inside it. That is once per tensor per decoded frame, and once per tensor on every filter `setInputDim`.
- **`addToSysPath()`, per open.** Every filter, converter or decoder core appended `"."` and the script directory to `sys.path`, and leaked both strings.
  - `sys.path` grew by 2 entries per open: 32 → 72 over 20 reopens of a decoder, 51 → 1207 in a long single-shot loop.
  - Every later import searches the longer list.

The fix releases all three at the common exit, releases the tuple after the constructor call (the `TensorShape` holds its own references to what it keeps), and skips a `sys.path` entry that is already there. Import resolution does not change: the first match in `sys.path` wins, and the skipped entries were duplicates. The production diff is +19/−6 in three files; two of those lines are error texts of the same `decode()` block, which named `'convert'` and `flexbuf`.

## Why these tests do not use Valgrind

- The Valgrind job never builds the python3 sub-plugins. `actions/setup-python` installs Python 3.11 without numpy, so meson prints `Disable nnstreamer-python3`.
- Even if it did build them, `check_valgrind_log.sh` gates on errors and *definite* leaks. These leaked objects stay linked into the Python collector, so memcheck reports them as *still reachable*.
  - Measured with a python decoder pipeline: going from 100 to 400 buffers grew still-reachable by 0.6–1.1 MB; definite leaks did not change.

The new tests ask the interpreter instead (`tests/unittest_python3_util.h`). They count the objects the collector tracks, read the length of `sys.path`, and read an object's reference count. They run in the jobs that build python3: Ubuntu Pdebuild (jammy, resolute) and Tizen GBS x86_64.

## Tests

**Commit 1** — the fix and the cases that pin it:
- `unittest_decoder_python3` is a new binary, built whenever python3 is enabled. It uses `custom_decoder_raw.py`, which needs only numpy; `custom_decoder.py` needs flexbuffers, which the Pdebuild chroot lacks.
  - `decodeOutput`: both tensors and their shapes reach the script.
  - `decodeRepeatedKeepsObjects`: 500 frames leave the collector's object count unchanged, within a slack of 20.
  - `decodeRaiseKeepsObjects_n`: 50 frames of a raising script all return `GST_FLOW_ERROR`, and the object count stays unchanged.
  - `decodeUnwritableOutput_n`: decoding into a read-only output buffer fails, and the refcount of the returned object stays unchanged.
  - `decodeMultiThread`: 4 threads × 300 frames give correct output and leave the object count unchanged.
  - `reopenKeepsSysPath`, `openInvalidScript_n`: 20 reopens, or 10 failing opens, leave `sys.path` the same length.
  - `openOtherDirectoryGrowsSysPathOnce`: a copy of the script in another directory puts that directory on `sys.path` exactly once — the dedupe must not turn into "never append".
  - `getOutCaps`: the caps the script gives are the caps of the decoder.
- `unittest_filter_python3`: `setInputDimRepeatedKeepsObjects`, `reopenKeepsSysPath`, `openInvalidScriptKeepsSysPath_n`.
- `unittest_converter`: `reopenKeepsSysPath`, `openInvalidScriptKeepsSysPath_n`.
- The RISC-V job cannot import `nnstreamer_python`, which is why it already skips `unittest_filter_python3` (964d01b0). The new binary is added to that skip list.

**Commit 2** — regression guards that pass on main:
- `invokeMultiThreadKeepsObjects`: 4 filter instances × 300 invokes from their own threads, each output checked and given back through `DESTROY_NOTIFY`.
- `convertMultiThreadKeepsObjects`: 4 converter instances × 300 conversions from their own threads, alternating between the case that hands out the script's array and the passthrough case.
- Both require the object count to be unchanged afterwards. The collector tracks containers only — a numpy array is not tracked — so both also have the script hand out `KEPT`, an array its module holds, and require its reference count to return to where it started. The filter case also requires that count to be one higher while an output is still out, so that taking the copy path instead of the reference path cannot make the check vacuous.
- Each thread drives a core of its own, as a pipeline with several python elements does, so what these cover of #3822 is entering the one interpreter from several threads at once; a hang there shows up as the test binary's timeout.

## Evidence (local WSL, jammy, numpy 1.21.5, flatbuffers 23.5.26)

- **Before/after:** all new tests pass on this branch. Against the unfixed plugins, **12** of the new cases fail (7 decoder, 3 filter, 2 converter):
  - decoder: +3987 objects over 500 frames, +9600 over four threads, `GST_FLOW_ERROR` refcount +50;
  - filter: +596 objects over 300 `setInputDim` calls;
  - `sys.path`: +2 entries per open.
- **Mutation matrix:** 12 of 12 mutations fail their intended cases:
  - the `raw_data`, `in_info` and error-path `output` releases in the decoder;
  - the `args` release, checked through both the decoder and the filter;
  - the `sys.path` dedupe, checked through the decoder, filter and converter, in both directions: never deduplicating, and never appending;
  - for the commit-2 guards, the output-array release in `PYCore::freeOutputTensors`, the contiguous-array release in the converter's `convert()`, and the filter copying the array instead of referencing it.
- **Not pinned:** the release of the two `sys.path` strings. The import system's caches also hold them, partly from untracked tuples, so their reference count has no stable expected value. I tried a block-count check and a referrer-count check and dropped both as unstable.
- **Stability:** the new cases passed 30 repeats each with 56 busy loops running, with no failures.
- **Regression:**
  - `meson test`: 27/27;
  - python3 SSAT groups (filter/converter/decoder): same results as main;
  - each commit builds and passes on its own;
  - clang-format 16 and `doxygen-tag.sh` (with ctags) are clean; `pylint -E` on the new script is clean.
- **End to end, outside the tests:** a decoder pipeline of 4 × 50 000 frames went from +2277 B per frame to +0.19 B. Open/close churn went from 17–24 kB to 2.6–4.7 kB per round, and `sys.path` stayed constant.

## Not in this PR

- **F10** stays open: the decoder's NULL `*pdata`, no type check on the returned object, and float16.
- **Pdebuild SSAT runs 0 cases:** the Pdebuild jobs run the converter/decoder python3 SSAT groups with **0 cases** (no flexbuffers module) and report Passed.
- **Vacuous cv2 SSAT on GBS:** Tizen GBS has no cv2, so filter SSAT 4-1 is IGNORED and 4-2 compares two empty files.

Both of the last two are gaps in CI packaging, not in these plugins, and deserve their own issue.


myungjoo:
> This is a relayed review written by another AI agent (Claude, Sonnet 5), performed at the requester's direction. Findings below are the agent's own analysis of the diff (`gh pr diff 4968`, base `611c369f`), not a copy of a human review.

## Scope checked
Diff is exactly the 9 files reported by the PR (917 +/4−): `ext/nnstreamer/extra/nnstreamer_python3_helper.cc`, `ext/nnstreamer/tensor_decoder/tensordec-python3.cc`, `tests/meson.build`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc` (new), `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/custom_decoder_raw.py` (new), `tests/unittest_python3_util.h` (new), `.github/workflows/risc-v.yml`. No unrelated modules are touched.

## Correctness of the fix (production code, +19/−4)

- `addToSysPath()` (`nnstreamer_python3_helper.cc`): each `PyUnicode_FromString` result is now decref'd unconditionally after the `PySequence_Contains`/`PyList_Append` dance, whether or not it was appended. `PyList_Append` does not steal the reference, so this closes the per-open leak of both strings. Dedup only skips an entry already equal-by-value in `sys.path`, so import resolution (first match wins) is unaffected, as claimed.
- `PyTensorShape_New()`: the new `Py_SAFEDECREF(args)` after `PyObject_CallObject` is safe. I traced `TensorShape_init`/`TensorShape_setDims`: because `PyTensorShape_New` always builds `dims` with exactly `NNS_TENSOR_RANK_LIMIT` entries, `TensorShape_setDims` always takes the `PyList_GetSlice` branch (a fresh, independently-refcounted copy), never the aliasing branch. The `type` descriptor is likewise re-retained by `TensorShape_init` before `args` is released. So releasing `args` cannot cause a use-after-free of `self->dims`/`self->type` on either call site (`tensordec-python3.cc:151` and `tensor_filter_python3.cc:514`).
- `PYDecoderCore::decode()`: the three `Py_SAFEDECREF` calls added at `done:` cover both the pre-existing `goto done` on missing `decode` attribute and on unmappable output, plus normal fall-through. `Py_SAFEDECREF` self-nulls, so the earlier explicit `Py_SAFEDECREF(output)` in the success branch plus the unconditional one at `done:` is not a double-free.

No correctness or regression issues found in the production hunks.

## Tests

- Test-to-fix ratio (~900 new lines for a 19-line fix) is large but justified: the PR body explains Valgrind never builds python3 sub-plugins here and even if it did, leaked-but-still-linked interpreter objects would show as "still reachable," not "definitely lost" — so a GC/`sys.path`-introspection approach (`tests/unittest_python3_util.h`) is the only way to pin this regression in CI. That reasoning checks out against `.github/workflows/ubuntu_clean_meson_build.yml` (installs python3-numpy via apt, but `actions/setup-python@v5` puts an interpreter without numpy earlier on `PATH`, so meson still disables `nnstreamer-python3` there).
- New/verified call sites (`scaler.py`, `passthrough.py`, `filter_output_cases.py`, `converter_output_cases.py`, `_g_memdup`, `_python_open_output_cases`) all pre-exist; `custom_decoder_raw.py` is new and reads `NNS_TEST_PY_DECODER_MODE` through `os.environ` at `__init__` time, consistent with `py_test_setenv`'s note that `os.environ` is a snapshot the interpreter takes at `import os`, not the C `environ` — `g_setenv()` would not have worked here.
- `tests/meson.build`: `unittest_decoder_python3` is correctly gated under the same `if have_python3` block as `unittest_filter_python3`, with `python3_dep`, no `flatbuf_dep` (deliberately, since the Pdebuild chroot lacks flexbuffers). `run_unittests_binaries.sh ./tests` (used by Tizen GBS's `unit_test` path) discovers gtest binaries by directory walk, so the new binary is picked up the same way `unittest_filter_python3` already is — no packaging/spec change needed, and none was made.
- `.github/workflows/risc-v.yml`: `-k`/`--skip` in `run_unittests_binaries.sh` is a *skip* list (confirmed in the script), so adding `unittest_decoder_python3` next to `unittest_filter_python3` correctly excludes it where `nnstreamer_python` can't be imported (964d01b0 precedent).

### Minor / Nit findings

1. **Minor** — `tests/unittest_python3_util.h:20` (`PY_TEST_GC_SLACK (20)`). This has only been validated locally on jammy with numpy 1.21.5 per the PR's own "Evidence" section; it has not been run against resolute (numpy 2.x) or the Tizen GBS python3, both of which are the CI jobs meant to gate on it. numpy 2.x's larger internal type-promotion/ufunc caches conceivably add a few more stray objects on first use per process, though the tests already warm up 10 iterations before taking the "before" snapshot, which should absorb one-time cache population. Recommend watching the actual Pdebuild-resolute and Tizen GBS runs on this PR before merge, since a too-tight slack would show up as an intermittent, hard-to-reproduce CI failure blocking unrelated future PRs, and 20 is a fairly small margin for an interpreter-wide GC object count.
2. **Nit** — `ext/nnstreamer/extra/nnstreamer_python3_helper.cc:389` (`addToSysPath`). `PyErr_Clear()` is called once after the two-entry loop rather than per-iteration. If `PySequence_Contains` ever returned -1 (error) on the first entry, the pending exception stays set while `PyUnicode_FromString`/`PySequence_Contains` run for the second entry; in CPython this is generally harmless for these specific APIs on string objects, but it's a bit fragile style-wise. Not a leak or a correctness bug as far as I can tell — just suggest clearing inside the loop body if it's cheap to do.

## Verdict

mergeable after the nits (the GC-slack point (1) is worth a human glance at the actual resolute/Tizen CI results on this PR before merging, but is not something I could falsify by static review alone; (2) is cosmetic only).

myungjoo:
Thanks for the review. Both points:

1. **Nit, `PyErr_Clear()` position:** taken, and it was more than cosmetic. If `PySequence_Contains()` failed on `"."`, the error stayed pending while the second iteration called `PyUnicode_FromString()` and `PySequence_Contains()` again, and C-API calls with an exception set are not allowed. The clear now runs at the end of each iteration. It is folded into the first commit; the new head is `868d1a42`, and the delta to the reviewed head is that one line. Locally, `unittest_decoder_python3` 7/7, `unittest_filter_python3` 27/27 and `unittest_converter` 17/17 pass.
2. **Minor, `PY_TEST_GC_SLACK` only validated on jammy:** agreed that the CI jobs are the real check. The Pdebuild resolute (newer Python and numpy 2.x) and Tizen GBS x86_64 runs of the previous head `f110108a` are still in progress. The test sources are unchanged in the new head, so I will read the per-case results of both runs before this leaves draft, and report them here.

myungjoo:
CI results for the Minor point (`PY_TEST_GC_SLACK` validated only on jammy), on the previous head `f110108a`. The test sources are unchanged in `868d1a42`.

All 14 new cases ran and passed in every job that builds python3. I checked each case's `[ OK ]` line in the job logs, not just the job status:

| Job | Environment | New cases |
|---|---|---|
| Ubuntu pdebuild on jammy (amd64) | numpy 1.21 | 14/14 OK |
| Ubuntu pdebuild on resolute (amd64) | newer Python, numpy 2.x | 14/14 OK |
| Tizen GBS build on Ubuntu (x86_64, unit_test 1) | numpy 2.4.1 | 14/14 OK |

- The 14 cases are the 7 `nnstreamerDecoderPython3.*` cases, 4 `nnstreamerFilterPython3.*` cases (`setInputDimRepeatedKeepsObjects`, `invokeMultiThreadKeepsObjects`, `reopenKeepsSysPath`, `openInvalidScriptKeepsSysPath_n`), and 3 `tensorConverterPython.*` cases (`convertMultiThreadKeepsObjects`, `reopenKeepsSysPath`, `openInvalidScriptKeepsSysPath_n`).
- The GBS job also installs the new binary into the unittest package (`usr/lib/nnstreamer/bin/unittest-nnstreamer/tests/unittest_decoder_python3`).
- The whole run is 20/20 green except the x86 Valgrind job, which was still running when I checked; that job does not build python3.
- So the slack holds on the newer interpreter and numpy as well; I consider the Minor point resolved.

myungjoo:
> This is a relayed review written by another AI agent (Claude, Sonnet 5), round 2, performed at the requester's direction. Findings below are the agent's own independent analysis of the diff at head `868d1a42` (`gh pr diff 4968`, base `611c369f`), not a copy of the round-1 review (comment 5675524989).

## Scope checked

Full diff, all 9 files: `.github/workflows/risc-v.yml`, `ext/nnstreamer/extra/nnstreamer_python3_helper.cc`, `ext/nnstreamer/tensor_decoder/tensordec-python3.cc`, `tests/meson.build`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc` (new), `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/custom_decoder_raw.py` (new), `tests/unittest_python3_util.h` (new). No unrelated module is touched; the +19/−4 production change is confined to the python3 helper and the decoder core.

## Round-1 status

- **Nit (`PyErr_Clear()` placement in `addToSysPath`): resolved.** At `868d1a42`, `nnstreamer_python3_helper.cc:389` clears the error inside the loop body after each `PySequence_Contains`/`PyList_Append`/decref, not once after both iterations. This also fixes the latent issue the author's reply pointed out (a pending exception from the first entry would otherwise be live across the second entry's C-API calls).
- **Minor (`PY_TEST_GC_SLACK` validated only on jammy): still open, not yet resolved by evidence.** As of this review, `gh pr checks 4968` shows **"Ubuntu pdebuild on resolute (amd64)"** and **"Tizen GBS build on Ubuntu (x86_64, --define \"unit_test 1\")" both still `pending`** — the two jobs that would actually exercise `PY_TEST_GC_SLACK (20)` against numpy 2.x / a different Python have not completed yet. This is not a code defect, just an unfinished verification step; please check both go green (specifically no `_n`-suffixed or `...KeepsObjects` test flaking near the slack boundary) before merging.

## Independent verification (production hunks, `+19/-4`)

I re-traced CPython ownership from scratch rather than re-checking round-1's conclusions:

- `addToSysPath` (`nnstreamer_python3_helper.cc:365-396`): each `PyUnicode_FromString` result is decref'd unconditionally via `Py_SAFEDECREF` regardless of whether `PyList_Append` used it (append does not steal the reference), closing the per-open 2-string leak. Dedup is by value, so import resolution (first match wins) is unaffected.
- `PyTensorShape_New` (`nnstreamer_python3_helper.cc:515-544`) releasing `args` after `PyObject_CallObject`: I read the actual `TensorShape_init`/`TensorShape_setDims` implementation in the same file (lines 44-157). Because every call site (`tensordec-python3.cc:151`, `tensor_filter_python3.cc:514`) always builds `dims` with exactly `NNS_TENSOR_RANK_LIMIT` entries, `TensorShape_setDims` always takes the `len == NNS_TENSOR_RANK_LIMIT` branch, i.e. `new_dims = PyList_GetSlice(dims, 0, NNS_TENSOR_RANK_LIMIT)` — a fresh, independently-refcounted list, never an alias of `args`'s `dims` element. `type` goes through `PyArray_DescrConverter`, which likewise hands back its own reference independent of `args`. So decref'ing `args` right after the call cannot dangle `self->dims`/`self->type`. Confirmed safe.
- `PYDecoderCore::decode()` (`tensordec-python3.cc:198-203`): the three `Py_SAFEDECREF` calls at `done:` cover the pre-existing `goto done` on missing `decode` attribute and on unmappable output, plus normal fall-through. `Py_SAFEDECREF` self-nulls (`nnstreamer_python3_helper.h`), so the earlier explicit `Py_SAFEDECREF(output)` in the success branch plus the unconditional one at `done:` is provably not a double-free.

No correctness or regression issue found in the production code.

## Independent verification (things the task asked to scrutinize specifically)

- **Does `decodeUnwritableOutput_n` really discriminate?** Yes. `gst_buffer_new_wrapped_full(GST_MEMORY_FLAG_READONLY, ...)` followed by `gst_buffer_get_all_memory()` yields a `GstMemory` with the read-only flag, and `gst_memory_map(..., GST_MAP_WRITE)` refuses to map read-only memory for writing (standard GStreamer behavior) — so `PYDecoderCore::decode()` reliably takes the `goto done` branch on every one of the 50 iterations. The refcount probe (`py_test_attr_refcount("custom_decoder_raw", "FIXED")`) is well chosen: `decode()` returns the same module-level `FIXED` bytes singleton every call, so a real regression would show up as `after` growing linearly with the 50 iterations rather than equal to `before`.
- **Thread-safety of the multithread tests vs. GIL/`Py_LOCK` ordering:** `PYDecoderCore::Py_LOCK()`/`Py_UNLOCK()` (and the filter/converter equivalents) are a **per-instance `GMutex`, not the GIL**. However, all three new multithread tests call into Python only through the plugin's public callback (`dec->decode`, `sp->invoke`, `ex->convert`), and I verified each of those wrapper functions — `decoder_py_decode` (`tensordec-python3.cc:369`), `TensorFilterPython::invoke` (`tensor_filter_python3.cc:794-798`), `python_convert` (`tensor_converter_python3.cc:373-380`) — take a `PyGILGuard` (real `PyGILState_Ensure`/`Release`) before touching any Python object. `PyGILState_Ensure` is safe to call from a raw `GThread` that was never created by Python, so the new tests are not exposed to the GIL-vs-instance-mutex ordering problem I was asked to check for. No issue found here.
- **`py_test_setenv`'s `os.environ` mutation leaking across tests/binaries:** No leak found. Every call site in `unittest_decoder_python3.cc` sets `NNS_TEST_PY_DECODER_MODE` immediately before the `dec->setOption()` call that constructs a new `CustomDecoder` instance (whose `__init__` reads the env var), so ordering is always setenv-before-construct within the same test. Each gtest binary is a separate process, so nothing crosses binaries; within one binary, gtest runs tests in declaration order and every test that opens a decoder passes its own explicit mode, so there is no reliance on leftover state from a previous test.

## New finding (round 2)

- **Nit — inconsistent warm-up before the "before" GC snapshot in the three new multithread tests.** `decodeRepeatedKeepsObjects`/`decodeRaiseKeepsObjects_n` (and the pre-existing single-shot python3 tests) run **10** warm-up iterations before taking `before = py_test_gc_object_count()`, explicitly to absorb one-time interpreter/numpy cache population (per the PR body). The three new multithread tests do not follow that discipline:
  - `nnstreamerDecoderPython3.decodeMultiThread` (`tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc:493-526`) does only **1** warm-up decode per instance (4 total) before the snapshot.
  - `nnstreamerFilterPython3.invokeMultiThreadKeepsObjects` (`tests/nnstreamer_filter_python3/unittest_filter_python3.cc:729-776`) and `tensorConverterPython.convertMultiThreadKeepsObjects` (`tests/nnstreamer_converter/unittest_converter.cc:144-171`) do **zero** warm-up invokes/converts before the snapshot — the first real invoke/convert for each of the 4 threads happens only after `before` is already captured, concurrently with the other threads.
  - Why this matters: these are exactly the tests most likely to trigger one-time, per-shape/per-dtype internal caching on the first real data path (this is the same class of noise the 10-iteration warm-up in the sibling tests was added to absorb), and they are also the tests spawning fresh native threads for the first time in the binary. `PY_TEST_GC_SLACK (20)` is admittedly a "fairly small margin" by the PR's own admission, so combining a smaller/zero warm-up with concurrent first-time execution modestly raises the chance of an intermittent CI failure on a numpy/Python version this hasn't been run against yet (resolute, Tizen), compared to the single-threaded siblings.
  - Suggestion: give each thread (or at least each of the 4 instances, once, before spawning) one warm-up call symmetric with the single-threaded tests, the same way `decodeMultiThread` already does once — or explicitly document why zero/one warm-up is believed sufficient here. Not blocking by itself, but worth doing before or shortly after this lands, given the resolute/Tizen runs are still pending anyway.

## Other checks (no issues found)

- Test naming: every failure-path test (`decodeRaiseKeepsObjects_n`, `decodeUnwritableOutput_n`, `openInvalidScript_n`, `openInvalidScriptKeepsSysPath_n` ×2) carries the `_n` suffix; every success-path test does not. Consistent with repo convention.
- Doxygen: every new function/test in the three `.cc` files, the new header, and the new `.py` script has a brief comment, matching CI's doxygen-tag enforcement.
- Build/packaging: `unittest_decoder_python3` is gated under the same `if have_python3` block as `unittest_filter_python3`, depends on `python3_dep` only (no `flatbuf_dep`, matching the stated reason the Pdebuild chroot lacks flexbuffers). `packaging/nnstreamer.spec`'s `%files unittests` packages the whole `.../unittest-nnstreamer/tests` tree and `run_unittests_binaries.sh` walks the test directory rather than a fixed binary list, so the new binary needs no packaging-file changes — verified directly against both files, not just asserted. `.github/workflows/risc-v.yml`'s `-k unittest_filter_python3,unittest_decoder_python3` matches `run_unittests_binaries.sh`'s documented `-k|--skip BINARY_NAME[,*]` comma-separated syntax (verified against the script source).
- No architecture/API change; none of the touched surfaces are documented in `Documentation/`, so no doc update is required.

## Verdict

**mergeable after the nits.** No correctness or regression issue found in the production code or in the new tests' logic. The one open item from round 1 (GC-slack margin on resolute/Tizen) is still unresolved only because the relevant CI jobs are still pending as of this review, not because of anything falsifiable by static review — please confirm both go green before merging. The new warm-up-inconsistency nit above is worth a follow-up but is not, by itself, a reason to block.

myungjoo:
Thanks for round 2.

- **Minor, slack validated only on jammy:** resolved by the CI evidence in the comment above, posted a few minutes before this review landed. All 14 new cases show `[ OK ]` in Pdebuild jammy, Pdebuild resolute and Tizen GBS x86_64 (numpy 2.4.1).
- **Nit, warm-up before the "before" snapshot in the multithread tests:** taken, so that all multithread cases now warm up the same way as the single-thread cases.
  - `decodeMultiThread` now decodes 10 frames per instance before the snapshot, instead of 1.
  - `invokeMultiThreadKeepsObjects` and `convertMultiThreadKeepsObjects` get a `rounds` field. Each instance runs 10 rounds on the main thread before the snapshot, then 300 in its own thread. The warm-up outputs are checked too.
- **New head:** `d965f739`, still two commits. Each change is folded into the commit that owns the test: decoder into the first, filter/converter into the second. The delta to `868d1a42` is 3 test files, +12/−3.
- **Local results:** `unittest_decoder_python3` 7/7, `unittest_filter_python3` 27/27, `unittest_converter` 17/17. clang-format 16 is clean. Against the unfixed plugins, the same 11 cases still fail.

myungjoo:
> This is a relayed review written by another AI agent (Claude, Sonnet 5), round 3, performed at the requester's direction. Findings below are the agent's own independent analysis of the diff at head `d965f739` (base `611c369f`), not a copy of round 1 (comment 5675524989) or round 2 (comment 5675608451-adjacent review comment).

## Scope checked

Full diff against merge-base `611c369f`, all 9 files: `.github/workflows/risc-v.yml`, `ext/nnstreamer/extra/nnstreamer_python3_helper.cc`, `ext/nnstreamer/tensor_decoder/tensordec-python3.cc`, `tests/meson.build`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc` (new), `tests/nnstreamer_filter_python3/unittest_filter_python3.cc`, `tests/test_models/models/custom_decoder_raw.py` (new), `tests/unittest_python3_util.h` (new). No unrelated module is touched; production change stays +19/−4 across the two helper/decoder files.

## Round 1 and round 2 status

- **Round 1 nit (`PyErr_Clear()` placement in `addToSysPath`): confirmed resolved.** `git diff 611c369f d965f739 -- ext/nnstreamer/extra/nnstreamer_python3_helper.cc` shows `PyErr_Clear()` inside the `for (const gchar *entry : entries)` loop body, after the per-entry `Py_SAFEDECREF(item)`, not once after both entries.
- **Round 1 minor (`PY_TEST_GC_SLACK` validated only on jammy): confirmed resolved, verified independently from raw job logs, not just the author's summary.** I pulled the actual logs for the two jobs that matter (run `34934730544` job "Ubuntu pdebuild on resolute (amd64)", id `104270089371`; run `34934730575` job `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`, id `104270089538`) and grepped for the 14 new test names. All 14 `[ OK ]` on both, e.g. resolute: `nnstreamerDecoderPython3.decodeMultiThread (76 ms)`, `nnstreamerFilterPython3.invokeMultiThreadKeepsObjects (78 ms)`, `tensorConverterPython.convertMultiThreadKeepsObjects (57 ms)`, plus all `reopenKeepsSysPath`/`openInvalidScript(KeepsSysPath)_n`/`decode*` cases; Tizen GBS x86_64 (numpy 2.4.1) same 14/14. Matches the author's table.
  - Caveat carried forward below: these two runs are for head `f110108a` (pre-warm-up), confirmed via `gh run view --json headSha`.
- **Round 2 nit (multithread tests skip/shortchange the 10-iteration warm-up before the "before" GC snapshot): confirmed resolved, verified against the actual diff, not just the author's description.** `git diff 868d1a42 d965f739` (868d1a42 is a locally-fetched loose object, reachable) is exactly 3 files, +12/−3, matching the author's claim byte for byte:
  - `decodeMultiThread`: `for (guint i = 0; i < 10; i++) _decode_and_check(...)` per instance (was 1), run on the main thread before `before = py_test_gc_object_count()`.
  - `invokeMultiThreadKeepsObjects` / `convertMultiThreadKeepsObjects`: new `rounds` field; each instance runs `_python3_invoke_thread`/`_python_convert_thread` once synchronously with `rounds = 10` on the main thread (mismatch counter shared, so a warm-up failure still fails the test), then `rounds` is set to 300 before the real `g_thread_new` fan-out. No race: the mutation of `td[t].rounds` back to 300 happens before any thread reads it.

## Independent verification (production hunks, unchanged from round 2's finding, re-derived from scratch)

- `addToSysPath` (`nnstreamer_python3_helper.cc:379-393`): `PyUnicode_FromString` result is decref'd unconditionally per entry via the NULL-safe, self-nulling `Py_SAFEDECREF` macro (confirmed its definition at `nnstreamer_python3_helper.h:47-53`); `PyList_Append` does not steal the reference, so this closes the leak. Dedup via `PySequence_Contains` is by value, so `sys.path` resolution order is unaffected.
- `PyTensorShape_New` (`nnstreamer_python3_helper.cc:509-545`): `Py_SAFEDECREF(args)` runs after `PyObject_CallObject` returns. Confirmed there are exactly two call sites in the whole tree (`git grep -n PyTensorShape_New`): `tensordec-python3.cc:151` and `tensor_filter_python3.cc:514`. Read `TensorShape_setDims` (`nnstreamer_python3_helper.cc:44-`): both call sites always build `dims` with exactly `NNS_TENSOR_RANK_LIMIT` entries, so `TensorShape_setDims` always takes the `PyList_GetSlice` branch — a fresh, independently-refcounted list — never an alias into `args`'s list. `type` is independently retained via `PyArray_DescrConverter`. So releasing `args` right after the call cannot dangle `self->dims`/`self->type`.
- `PYDecoderCore::decode()` (`tensordec-python3.cc:196-204`): the three `Py_SAFEDECREF` at `done:` are all NULL-safe/self-nulling, so the pre-existing explicit `Py_SAFEDECREF(output)` on the success path plus the new unconditional one at `done:` is not a double-free. Traced the `goto done` on the unmappable-output path (`gst_memory_map` failure) specifically: `output` is still live there before this fix (the exact leak `decodeUnwritableOutput_n` pins), and is now released.
- Spot-checked the sibling plugins for the same class of bug, since the PR claims "the python3 filter and converter hold no per-invoke leak today": `tensor_converter_python3.cc` already does `Py_SAFEDECREF(param); Py_SAFEDECREF(pyValue);` at its common exit, and `tensor_filter_python3.cc`'s invoke path already does `Py_SAFEDECREF(param)`/`Py_SAFEDECREF(result)` at its exits. Claim holds; no missed leak in-scope of F9 that this PR should have covered but didn't.

## Independent verification (tests)

- Read `tests/unittest_python3_util.h` in full: all four helpers (`py_test_gc_object_count`, `py_test_sys_path_length`, `py_test_attr_refcount`, `py_test_setenv`) take the GIL via `PyGILState_Ensure`/`Release`, are NULL/error-safe (`PyErr_Clear()` before release), and each has a Doxygen comment (matches CI's doxygen-tag gate).
- Traced `decodeUnwritableOutput_n` end to end: `_open_decoder(&pdata, "fixed")` makes `decode()` always return the module-level `FIXED` bytes singleton; `gst_buffer_new_wrapped_full(GST_MEMORY_FLAG_READONLY, ...)` + `gst_memory_map(..., GST_MAP_WRITE)` reliably fails to map for writing, so every one of the 50 iterations takes the `goto done` path this PR fixes. `py_test_attr_refcount("custom_decoder_raw", "FIXED")` is a well-chosen probe: a real regression (the pre-fix code) grows the refcount linearly with iterations, not a one-off.
- `custom_decoder_raw.py`: every function has a Doxygen-style comment block; reads `NNS_TEST_PY_DECODER_MODE` from `os.environ` at `__init__`, consistent with `py_test_setenv`'s documented reason for not using `g_setenv()`.
- `tests/meson.build`: `unittest_decoder_python3` is inside the same `if have_python3`-guarded block as `unittest_filter_python3`, depends on `nnstreamer_unittest_deps` + `python3_dep` only (no `flatbuf_dep`), uses the same `install_dir`/`install-test` pattern. The new binary's successful compile+run in both job logs above is itself proof the meson wiring (and the new header's include path) is correct.
- No architecture or public-API change anywhere in the diff; nothing in `Documentation/` documents these internal helpers today, so no doc update is owed here.

## Open item (procedural, not a code defect)

- **The exact current head (`d965f739`) has not finished its own CI as of this review.** `gh pr checks 4968` shows every job that matters here — "Ubuntu pdebuild on jammy/resolute", "Tizen GBS build on Ubuntu (x86_64, unit_test 1)", "Static checks" — as `pending` (just kicked off; run `34936707xxx`). Only "CPP-Linter" and "DCO" have completed (`pass`) for this head so far.
  - Risk is low: the only sources that changed since the fully-green `868d1a42`/`f110108a` runs are the 3 warm-up hunks reviewed above, which are additive and executed strictly before any snapshot or thread spawn; the underlying multithread cases already passed 14/14 on both resolute and Tizen GBS x86_64 even *without* this warm-up (see the job-log evidence above), so the warm-up is very unlikely to introduce a new failure.
  - Still, please don't merge until the resolute and Tizen GBS x86_64 jobs on `d965f739` itself actually go green — that's the only piece of evidence in this whole thread that hasn't yet been observed on the literal head being merged.

## Verdict

**Mergeable as is**, once the currently-pending CI on `d965f739` (specifically Pdebuild resolute and Tizen GBS x86_64, which are the two jobs this whole review thread has been about) finishes green. No outstanding code-correctness, regression, scope, or documentation issue found; both prior nits are confirmed resolved by direct diff inspection, not just by re-reading the author's description.

myungjoo:
Thanks for both reviews. Every item is addressed at `714a9d4b`; the two commits keep their shape, the changes are folded into them.

**Minor — the commit-2 guards could not see a leaked array.** Confirmed by measurement, not only by reading: with the guards as they were, dropping `Py_SAFEDECREF (it->second)` in `PYCore::freeOutputTensors` and dropping `Py_SAFEDECREF (output_array)` in the converter's `convert()` both left the tests **passing**, with the gc delta unchanged at −1. `gc.is_tracked()` is False for a fresh array, a slice and a `.copy()`, as you said.

I took the refcount option rather than the allocator-block one. A block-delta check discriminates in the filter (≈60 baseline vs ≈2460 leaked) but is thin in the converter: baseline 391–465 against 1304–1356 leaked over three runs, so the threshold would be a guess. Instead both scripts now hold a module-level array `KEPT` and hand it out:

- `filter_output_cases.py`: the `kept` case returns `KEPT` as its first output, which is the `Py_INCREF` path of `PYCore::run`, released again through `DESTROY_NOTIFY`. `invokeMultiThreadKeepsObjects` runs that case and requires `py_test_attr_refcount("filter_output_cases", "KEPT")` to be back where it started after the threads join.
- `converter_output_cases.py`: a new case 8 returns `KEPT`, which `PyArray_GETCONTIGUOUS` increfs and `convert()` releases after `_g_memdup`. `convertMultiThreadKeepsObjects` alternates that case with the passthrough one, so the bytes are still checked per thread, and it requires the same of `py_test_attr_refcount("converter_output_cases", "KEPT")`.

The mutation matrix is now 10 of 10: the two mutations above fail their guard (`refs_after` +1200 in the filter, +600 in the converter). Both guards still pass on this branch, and passed three rounds of 4 concurrent copies of each binary.

**Nit 1 — the macro comment.** `PY_TEST_GC_SLACK` now says "a leaked container", and the `@note` of `py_test_gc_object_count()` says that only containers are tracked and points at `py_test_attr_refcount()` for an array, `bytes` or `str`.

**Nit 3 — log noise.** `decodeRaiseKeepsObjects_n` is down to 50 frames: 371 lines instead of 1261 here, and a leak still grows the count by ≥300 against the slack of 20. It still fails under mutations M1 and M2.

**Nit 4 — the subject tag.** The commit and the PR title are now `[Python3]`, which has precedent in this repo.

**NIT — the header comment.** `nnstreamer_python3_helper.h:70` now reads: `Append "." and the given path to the python sys.path list unless already present`.

**Nit 5 — the diff numbers.** The PR body says +16/−3 now.

Rechecked locally: build clean, `unittest_decoder_python3` 7/7, `unittest_filter_python3` 27/27, `unittest_converter` 17/17, each commit builds and passes on its own, clang-format 16 clean, `pylint -E` clean on both scripts.

myungjoo:
Second round on this head (`886823e9`), from two further automated reviews run by separate Claude agents and relayed here by me. Both approved; these are the items they raised that were worth acting on.

**The filter refcount check could have been vacuous.** `refs_after == refs_before` also holds if `PYCore::run` takes the copy path in every round, since a copy never touches `KEPT`'s count, and `memcmp` cannot tell a copy from the array itself. The invoking thread now also reads the count while the first round's outputs are still out, and the test requires it to be exactly one higher than after they are given back. Mutating `Py_INCREF (output_array)` into a `PyArray_NewCopy` fails that check. The converter has no such risk: `PyArray_GETCONTIGUOUS` copies only a non-contiguous array, and `KEPT` is contiguous by construction.

**The dedupe had no positive case.** Every test script lives in `tests/test_models/models`, so nothing asserted that a script in a *new* directory still gets its directory appended — a regression to "never append" would have been caught only indirectly. `openOtherDirectoryGrowsSysPathOnce` copies the script to a temporary directory, opens it, and requires `sys.path` to grow by exactly one entry, and by nothing on the second open. It fails in both directions: with the old always-append code (+2) and with the append removed (+0).

**`getOutCaps` was untested.** The decoder's other entry point now has a case of its own.

**Error texts in the block this commit fixes.** `decode()` reported `Fail to get output from 'convert'` and its map failure named `flexbuf`; both now name what they are. This is what the negative test printed 50 times per run.

**Notes and wording.** The warm-up rounds now say why they are there; the filter test notes that the four instances hand out the same buffer by design; the converter script note no longer implies only bytes above 8 pass through. In the commit messages, the converter takes its reference for the kept case rather than "per conversion", and the #3822 claim is now what the tests actually cover — several threads entering the one interpreter, each with a core of its own — rather than a deadlock on one instance.

**Numbers.** The production diff is +19/−6 across three files. Against the unfixed plugins 12 of the new cases fail, not 11 (7 decoder, 3 filter, 2 converter). `sys.path` growth is 32 → 72 over 20 reopens, i.e. the 2 per open the body claims.

Re-verified locally: mutation matrix now 12 of 12 caught, `unittest_decoder_python3` 9/9, `unittest_filter_python3` 27/27, `unittest_converter` 17/17, each commit builds and passes on its own, clang-format 16 and `pylint -E` clean.

Two items from the reviews are deliberately left out of this PR: `eventHandler` mutating `outputArrayMap` without `Py_LOCK` (serialized by the GIL today, pre-existing, belongs with #4920 F10), and the test-hygiene point that an `ASSERT_*` inside the open loops leaves earlier instances unclosed, which matches the surrounding cases in those files.

myungjoo:
Third round, at `acfcad2b`, from one more automated review of the previous head by a separate Claude agent, relayed by me. It approved; one real defect and three nits, all fixed here.

**The new sys.path case left a directory behind.** Importing the copy makes the interpreter write `__pycache__/custom_decoder_copy.cpython-3X.pyc` next to it, so `g_rmdir` failed with `ENOTEMPTY` and every run left a `nns_decoder_py3_*` directory in `$TMPDIR`. Confirmed here: six of them, each holding that one `.pyc`. The case now removes the directory and what is in it, and requires it to be gone; a run leaves nothing behind.

**`setOption` was checked with EXPECT.** On a failure the case went on to `exit()` with a NULL `pdata`, tripping `g_return_if_fail` and comparing sys.path lengths that mean nothing. Both calls are `ASSERT_TRUE` now, matching the rest of the file. The `dir` assertion also moved after the first `exit()`, so it cannot leave a core open.

**`getOutCaps` only checked the structure name.** The decoder also stamps the framerate from the config, so the case now runs with a real framerate and requires `framerate=30/1` next to `application/octet-stream`.

**Dead stores in the filter guard.** The reference counts around the released outputs were recorded on every thread's round 0, so the threaded phase overwrote them from a count four threads were moving. They are recorded once, in the warm-up, which is where the assertion reads them.

The commit message now also says the ten warm-up rounds are there so that what a script's first call imports is not counted.

Re-verified: mutation matrix 12 of 12 caught, `unittest_decoder_python3` 9/9, `unittest_filter_python3` 27/27, `unittest_converter` 17/17, each commit builds and passes on its own, clang-format 16 clean, and `$TMPDIR` is clean after a run.

For the record, the CI failure on the previous head was `add-apt-repository -y ppa:nnstreamer/ppa` losing its connection to Launchpad (`http.client.IncompleteRead`) in the Valgrind arm job's setup step. Re-running that job alone passed, 22/22 under memcheck.

myungjoo:

Re-review of the last delta at `acfcad2b` (the temp-directory cleanup, the `ASSERT` changes, the framerate check in `getOutCaps`, the once-only refcount recording in the filter guard): **approve, nothing blocking.**

- `_remove_dir` removes the directory including `__pycache__`, and the case checks it is gone.
- `setFramerateFromConfig` turns `rate_n=30, rate_d=1` into `framerate=30/1`, and the script supplies no framerate of its own, so the new assertion checks the C code; `gint` is the right type for `gst_structure_get_fraction`.
- The `< 0` guards record both counts in the same successful warm-up round and never write during the threaded phase; `-1` is a safe sentinel.

Left as they are, deliberately: an `ASSERT_*` failing mid-case can leave the temp directory or earlier filters open (the same as the surrounding cases), and `_remove_dir` follows a directory symlink, which cannot occur inside a fresh 0700 `g_dir_make_tmp` directory that only the test and the interpreter write to.

The CI failure on this head was the Valgrind x86 job stalling for six hours at the start of the bounding-box decoder SSAT group, which has no step timeout. That job builds no python3 code, the step passed on this PR's earlier heads and on other branches over the same main, and a re-run of the job passed in 19 minutes. All 20 checks pass now.

Issue #4960, opened 2026-09-10
[Security] Security audit of the NNStreamer core (gst/): tracking checklist


myungjoo:
Security audit of the NNStreamer **core** — everything under `gst/` (core library, elements, tensor_filter framework, tensor_query, nnstreamer-edge elements, mqtt, datarepo, join). Sub-plugins under `ext/` and tests are out of scope. Audited at `main` **192b4ddf** (2026-09-11). Line numbers refer to that commit; re-locate by grepping the quoted code before fixing.

Threat model: buffer bytes from upstream pads (including flexible/sparse meta headers), caps from peers, GObject property strings, network peers (tensor_query, edge, MQTT publishers), runtime files (datarepo JSON, `nnstreamer.ini`), environment variables and sysfs are all treated as untrusted. Every `g_assert` is live in release builds: nothing in the build defines `G_DISABLE_ASSERT`.

Relation to #4920 (memory-safety audit): items already tracked there are **not** repeated as checkboxes. The table at the end lists the open #4920 items that were re-verified as still present, plus the new aspects this audit found for some of them. Each item below is self-contained so it can be fixed in a separate session; every fix needs a regression test (CLAUDE.md).

Severity: **High** = memory corruption or OOB reachable from a remote peer or from stream data; **Medium** = process abort, unbounded resource use or race reachable from stream/peer data, or memory corruption needing a local property; **Low** = local property trigger with limited effect, or informational.

---

## A. Core library (`gst/nnstreamer/nnstreamer_plugin_api_impl.c`, `nnstreamer_plugin_api_util_impl.c`, `tensor_data.c`)

- [ ] **A1. `gst_tensor_meta_info_update_header()` writes 88 bytes into a buffer sized by `get_header_size()`, which returns 0 for any non-V1 version** — OOB write (heap), **High**. `nnstreamer_plugin_api_util_impl.c:1446` defines `GST_TENSOR_META_IS_V1(v)` as `(((v) & 0x00FFF000) & GST_TENSOR_META_MAKE_VERSION(1,0))`, a bit-AND, so `version = 0xDE002000` validates (`:1501-1535`) but `gst_tensor_meta_info_get_header_size()` (`:1542-1555`) returns 0. `:1589-1603`:
  ```c
  hsize = gst_tensor_meta_info_get_header_size (meta);
  memset (header, 0, hsize);
  memcpy (header, meta, sizeof (GstTensorMetaInfo));
  ```
[...]

PR #4969, opened 2026-09-15
[Converter] Validate flatbuf, flexbuf and protobuf messages before converting them

myungjoo:
- **What:** the `flatbuf`, `flexbuf` and `protobuf` tensor_converter sub-plugins parse bytes that can come from a peer, but they trusted every field. A malformed message crashed the pipeline (NULL deref, out-of-bounds read) or produced tensors smaller than their caps, which downstream elements then read past. These are items G16, G17 and G18 of #4920.
- **How:** each converter now verifies the message before reading it (flatbuffers `Verifier`, `flexbuffers::VerifyBuffer`, `ParseFromArray`'s result). It bounds the tensor count, the tensor vector and the dimension fields, and checks every tensor's data against its declared type/dimension (static) or its meta header (flexible/sparse) before appending it. The last check reuses the rule tensor_decoder already applies to its input memories.
- **Result:** new `unittest_converter_serialized` has 25 cases. On `main`, 21 of them fail, 10 with SIGSEGV/abort. All 25 pass with the fix, and valgrind reports no errors or leaks attributed to the repo. The flatbuf/flexbuf/protobuf/decoder/converter/python3 SSAT suites pass unchanged (284 cases). Found along the way: the in-tree python3 test decoder wrote malformed flexbuffers, now fixed (see below).

### For reviewers

**Behavior change to be aware of.** Python's `flexbuffers.Builder.Blob(numpy_array)` stores `len(array)`, the *element* count, as the blob length, while writing all the bytes. `tests/test_models/models/custom_decoder.py` did this, so every int16/float tensor it produced was a blob shorter than its declared size. The old flexbuf converter only accepted it because it ignored the blob length and read the size implied by the dims, which is exactly the G17 over-read. The script now passes `array.tobytes()`. A user script written the old way now gets an error naming both sizes instead of being accepted silently.

**Compatibility kept on purpose:** a flatbuf tensor without a `name`, and flatbuf/protobuf `dimension` fields shorter than 16 ranks (older peers wrote 4), still convert; the missing ranks read as 0.

**Commits** (one per item, then the tests):
1. `[Converter/Flatbuf]`: Verifier, field/count checks, and the new converter-util helper `tcu_check_tensor_data ()`.
2. `[Converter/Flexbuf]`: VerifyBuffer, blob-in-buffer check, share the blob size, and the python test script fix.
3. `[Converter/Protobuf]`: parse result, count bounds, dimension bound, and a name leak. It carries a local copy of the size check because `ext/nnstreamer/extra` is built before and without the converter util.
4. `[Test/Converter]`: the new gtest binary.

- Verified locally in docker `ubuntu:22.04` with ppa:nnstreamer flatbuffers 23.5.26 and protobuf 3.20.3 (the ubuntu meson CI job has neither, so these tests gate in pdebuild and Tizen GBS; Tizen ships flatbuffers 24.3.25, protobuf 33.1).
- `flexbuffers::VerifyBuffer` exists since flatbuffers 2.0.6; meson accepts >= 2.0.0, hence the `#if FLATBUFFERS_VERSION_*` guard (the test skips the no-message cases for flexbuf there).
- `NNS_TENSOR_SIZE_LIMIT` is 256 (16 is `NNS_TENSOR_MEMORY_MAX`); the 18-tensor round trip covers the extra-tensor memory path, which is why the size check must happen before `gst_tensor_buffer_append_memory ()`, not on the produced buffer.
- Negative control: new tests compiled against `main` converters: 21 of 25 fail (the round trips, the valid harness push and the pre-existing flatbuf count limit pass as guards).
- Not changed: G15 (decoder-side flatbuf/flexbuf/protobuf header parsing), H1/H2 (gRPC variants).
</details>


myungjoo:

This review covers commits 3d1a8a73 (flatbuf), 2d231122 (flexbuf), 5f6bfc0b (protobuf) and 6ead8e6b (test), i.e. everything in this PR (`ext/nnstreamer/extra/nnstreamer_protobuf.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_{flatbuf,flexbuf,util}.{cc,c,h}`, `tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc`, `tests/meson.build`, `tests/test_models/models/custom_decoder.py`). I read the diff, the surrounding converter/decoder code, the `.fbs`/`.proto` schemas, and cross-checked the new validation logic against the equivalent, already-shipped rule in `gst/nnstreamer/elements/gsttensor_decoder.c` (lines ~743-766).

## Summary of what I checked

- **Does it fix G16/G17/G18 (#4920)?** Yes, for the mainstream build configuration. `fbc_convert` now runs `flatbuffers::Verifier`/`VerifyTensorsBuffer`, bounds `num_tensor`, checks the tensor vector length, reads `dimension` only as far as the vector actually goes, and validates each tensor's data size via the new `tcu_check_tensor_data()`. `flxc_convert` now runs `flexbuffers::VerifyBuffer` (version-gated), bounds the blob pointer/size to the input buffer, and reuses the same size check. `gst_tensor_converter_protobuf` now checks `ParseFromArray()`'s return value, bounds `num_tensor` by both `NNS_TENSOR_SIZE_LIMIT` and the actual repeated-field count, bounds `dimension` reads by `dimension_size()`, and validates tensor data size with a local copy of the same rule (justified: `ext/nnstreamer/extra` builds before and isn't linked with the converter util — this is a legitimate reason for the duplication, though see Nit below).
- **Correctness of the new checks**: I traced the "negative/huge num_tensor" cases by hand. Flatbuffers' `Tensors.num_tensor` is a signed `int` in the schema; assigning a negative value to the `guint config->info.num_tensors` wraps to a huge unsigned value, which is still caught by the pre-existing `> NNS_TENSOR_SIZE_LIMIT` check before any tensor is touched — correct, and the new test `flatbufNegativeNumTensor_n` exercises exactly this. Protobuf's `num_tensor` is `uint32` in the `.proto`, so no sign issue there; it's bounded by both the limit and `tensor_size()`. Both converters free the previously-appended `out_buf` correctly on every error path (verified `out_buf` is initialized to `NULL` and only unref'd when non-NULL), so there's no double-free/UAF on the new error branches.
- **Consistency with existing rules**: `tcu_check_tensor_data()`/`_check_tensor_data()` mirror the header-size/data-size check in `gsttensor_decoder.c` almost line for line (including the `hsize == 0` "unknown meta version" guard), so this isn't new, untested logic — it reuses a rule the codebase already runs on every incoming buffer elsewhere.
- **Regression risk to other modules**: the diff is contained to the three converter subplugins + the shared converter util. I did not find any change to `tensor_decoder`, the flexible/sparse buffer layout, or any public/installed header (`tensor_converter_util.h` is an internal, non-installed header). The "compatibility kept on purpose" behavior (no name, dimension vectors shorter than 16 ranks) is real: `gst_tensor_dimension_is_valid()` in `nnstreamer_plugin_api_util_impl.c` already treats a rank-N dimension array with zero-filled trailing entries as valid, so reading missing flatbuf/protobuf dimension entries as 0 does not create a new incompatibility.
- **Test coverage**: `unittest_converter_serialized.cc` has all 25 `TEST()`s with a preceding Doxygen `@brief` (CI's per-function doxygen gate is satisfied), a mix of direct sub-plugin calls and `GstHarness`-based pipeline pushes, and both positive (compatibility) and negative cases per converter. `clang-format` (repo `.clang-format`) reports zero diff on all four changed/added C++ files. I could not rebuild the project in this sandbox (no flatbuffers/protobuf/meson toolchain available here) to independently reproduce the "21/25 fail on main, 25/25 pass with fix" claim, so that part is taken on the PR author's word plus static inspection of the diff.
- **custom_decoder.py fix**: legitimate and disclosed — `Builder.Blob(numpy_array)` records the element count, not the byte count, as the blob length; switching to `.tobytes()` is the correct fix and is exactly the class of bug G17 targets. The PR body proactively documents this as a "behavior change to be aware of" for any user script written the old way.

## Findings

1. **[Low] Residual verification gap on old flatbuffers (2.0.0–2.0.5) is undocumented in the production source.** `ext/nnstreamer/tensor_converter/tensor_converter_flexbuf.cc`, the `#if FLATBUFFERS_VERSION_MAJOR ...` guard around `flexbuffers::VerifyBuffer()` (around the new lines introduced in commit 2d231122, just after the `in_mem`/`in_info` setup). `meson.build` accepts `flatbuffers >= 2.0.0` (see `flatbuf_version_check_dep`), but `flexbuffers::VerifyBuffer()` only exists from 2.0.6 onward, so on a build configured against flatbuffers 2.0.0–2.0.5 the flexbuf converter parses an *unverified* buffer — the G17 fix is effectively inert for that version range (this is honestly disclosed in the PR description's "Context for agents" section, and the new test correctly skips the "garbage bytes" case for this configuration via `mode_verifies_untrusted_bytes()`, so it won't be silently claimed as covered). Suggestion: add a one-line comment at the `#else` branch of the version guard in `tensor_converter_flexbuf.cc` itself (not just the PR description) noting that buffers are unverified below flatbuffers 2.0.6, so a future maintainer reading only the source sees the caveat. Optionally consider raising the meson floor for flatbuffers to 2.0.6 when flexbuf/G17 safety matters, in a follow-up.

2. **[Nit] Duplicated size-check logic between `tensor_converter_util.c` and `ext/nnstreamer/extra/nnstreamer_protobuf.cc`.** `ext/nnstreamer/extra/nnstreamer_protobuf.cc:126-175` (`_check_tensor_data`) is a byte-for-byte logical copy of `tensor_converter_util.c`'s `tcu_check_tensor_data()`. The commit message explains why (build-order: `ext/nnstreamer/extra` is built before and isn't linked against the converter util), which is a reasonable constraint, but it means any future fix to this validation rule (e.g. a bug in the meta-header size logic) has to be applied in two places, and nothing enforces that. Not a blocker; worth a `@note`/TODO or a shared header-only inline function in a follow-up if this rule ever needs to change again.

3. **[Nit] Main "ubuntu meson" CI job doesn't build/run the new test at all.** `tests/meson.build:343` gates `unittest_converter_serialized` on `flatbuf_support_is_available or protobuf_support_is_available`, and per the PR body the default Ubuntu meson CI job has neither library installed, so this regression test only runs under the pdebuild and Tizen/GBS jobs (`.github/workflows/pdebuild.yml`, `.github/workflows/gbs_build.yml`), which do run on every PR against `main` and do include these dependencies (jammy PPA / Tizen packages). So a future regression would still fail CI and block merge — just via a less-obvious job than the "quick" ubuntu meson build. Worth being aware of when triaging a red build.

No High/Critical/Medium issues found. The bounds-checking logic itself is correct everywhere I traced it (flatbuf negative/huge count, protobuf count vs. declared vs. carried tensors, blob-vs-declared-size mismatches for both static and flexible/sparse formats), the error paths free resources correctly with no double-free, no public API or architecture changed (no doc update needed), and the change is tightly scoped to the three converters + shared util + tests.

Verdict: mergeable as-is

myungjoo:

Thanks for the review. Addressed at head `62ae67ef` (the only change from the reviewed head is two comment lines, folded into their commits; the commit trailers were also normalized):

1. **Low, flexbuf version guard:** added a one-line comment at the `#if` stating that flatbuffers < 2.0.6 has no `VerifyBuffer ()` and the input stays unverified there. Every build we ship or test uses a newer version (ppa:nnstreamer 23.5.26, Tizen 24.3.25, Homebrew current), and the blob-inside-buffer range check still applies on older versions. Raising the meson floor to 2.0.6 would remove the gap entirely, but it is a build-requirement change outside this fix.
2. **Nit, duplicated check:** the protobuf copy already carries a `@note` naming its twin. I added the matching note on `tcu_check_tensor_data ()` in `tensor_converter_util.h`, so whoever edits either copy sees the other. Removing the copy would mean linking `libnnstreamer_protobuf` (built in `ext/nnstreamer/extra`, before the converter directory) against the converter util, which is a bigger build-graph change than this fix warrants.
3. **Nit, CI gating:** correct, and deliberate. The Ubuntu meson job installs neither flatbuffers nor protobuf, so this binary gates in the pdebuild legs and the Tizen GBS builds (the GBS builds are required checks). Adding both libraries to the meson/valgrind job means pulling them from the PPA there, which is a CI change for a separate PR.

myungjoo:

Second-round review at head `62ae67ef`. I re-read the full diff (`ext/nnstreamer/extra/nnstreamer_protobuf.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_{flatbuf,flexbuf,util}.{cc,c,h}`, `tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc`, `tests/meson.build`, `tests/test_models/models/custom_decoder.py`), the round-1 review thread, and the author's replies, then did an independent pass over the surrounding validation helpers (`gst_tensor_info_get_size`/`gst_tensor_info_validate`/`gst_tensor_get_element_count`/`gst_tensor_dimension_is_valid` in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`) and the CI workflow triggers.

## On the round-1 thread

The two Nits and the Low were addressed as described and I verified each in the current source:
- The flexbuf `#if FLATBUFFERS_VERSION_MAJOR...` guard in `tensor_converter_flexbuf.cc:110-119` now carries the one-line "flatbuffers < 2.0.6 has no VerifyBuffer (); the input stays unverified" comment.
- `tensor_converter_util.h:30` and `nnstreamer_protobuf.cc:128` now cross-reference each other's copy of the size check via `@note`.
- I confirmed `.github/workflows/pdebuild.yml` and `.github/workflows/gbs_build.yml` both trigger on `pull_request: branches: [main]`, so the author's claim that the new test still gates every PR (just via those jobs rather than the ubuntu-meson job) is correct.

These responses are adequate; I have no further disagreement with round 1.

## New finding from independent review

**[Low] The new static-format size check can be defeated by integer overflow in the dimension-product helper it depends on, undermining the "tensors smaller than their declared caps" fix for the static branch.**
`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:1269-1282` (`gst_tensor_get_element_count`, unmodified by this PR) multiplies up to `NNS_TENSOR_RANK_LIMIT` (16) attacker-controlled `guint32` dimension values into a `gulong count` with no overflow check:
```c
for (i = 0; i < NNS_TENSOR_RANK_LIMIT; i++) {
  if (dim[i] == 0) break;
  count *= dim[i];
}
```
`gst_tensor_info_get_size()` (same file, lines 155-166) feeds this into `tcu_check_tensor_data()`'s (`ext/nnstreamer/tensor_converter/tensor_converter_util.c:74`) and `_check_tensor_data()`'s (`ext/nnstreamer/extra/nnstreamer_protobuf.cc:149`) static-format `expected = gst_tensor_info_get_size (info); if (size != expected) reject;` check that this PR newly wires into all three converters. Two dimension values near `UINT32_MAX` (reachable via flatbuf's/protobuf's `dimension` vector or flexbuf's typed vector — all peer-controlled, exactly the G16/G17/G18 threat model) already overflow a 64-bit accumulator, so `expected` can be made to wrap around to match whatever small `size` the peer actually sends. The check then passes while `_info->dimension` still carries the huge, un-downscaled values, handed downstream with a tiny backing memory — the same "declares a cap the actual data doesn't back, and downstream reads past it" failure this PR sets out to close, just reached via overflow instead of a direct size mismatch.

This is pre-existing and systemic (the same helper is already used by `gsttensor_decoder.c`'s validation, which the PR's own commit message says it is reusing), so it is not newly introduced by this PR, and I treat it the same way the author treated the flexbuf-version gap in round 1: legitimate, non-blocking, worth a one-line acknowledgment or a follow-up issue rather than scope-creeping this fix into rewriting `gst_tensor_get_element_count`/`gst_tensor_info_get_size` (which is used pervasively well beyond these three converters).

Suggested follow-up (not required for this PR): add an overflow-checked accumulation to `gst_tensor_get_element_count` (e.g. bail out / saturate if `dim[i] != 0 && count > SIZE_MAX / dim[i]`), which would harden this check plus every other caller of `gst_tensor_info_get_size` for free.

## Everything else

I re-verified the flatbuf `Verifier`/`VerifyTensorsBuffer` and flexbuf `VerifyBuffer` + blob-in-buffer range check, the `num_tensor` vs. tensor-vector-length bound in both flatbuf and protobuf, the `dimension_size()`/`dimension->size()` bounding before reading ranks, the `g_free (_info->name)` leak fix, and the `out_buf` NULL-initialization/unref-on-error paths — all correct, matching the PR's stated intent, with no double-free or use-after-free on any new error branch I traced. `tests/meson.build`'s new target links pre-existing `flatbuf_util_dep`/`protobuf_util_dep` dependencies (confirmed declared in `ext/nnstreamer/extra/meson.build`), consistent with the sibling `unittest_decoder` target already in the file. The 25 new gtest cases in `unittest_converter_serialized.cc` all carry a Doxygen `@brief`, cover both positive (compatibility: short dimension vectors, missing name) and negative (malformed/truncated/oversized/wrong-type) cases per format, and the `custom_decoder.py` fix (`.tobytes()` instead of the raw numpy array) is a correct, disclosed fix for the real blob-length bug it found along the way. No architecture or public/installed API changed, so no doc update is required. The diff stays tightly scoped to the three converters, the shared util, and tests.

Verdict: mergeable as-is

myungjoo:

Thanks. On the round-2 Low (dimension product overflow in `gst_tensor_get_element_count ()`): agreed that it can wrap the size these checks compare against. It is already tracked as **A2 of #4960** (High), whose fix belongs in the core helper, `gst_tensor_info_get_size ()` / `gst_tensor_meta_info_get_data_size ()`. Every "declared size == mapped size" check in the tree shares it, including the tensor_decoder element's input check this PR mirrors, and all of them inherit the fix once the helper refuses a wrapping product. A converter-local overflow check here would be a third copy of logic that should live in one place, so this PR leaves it to #4960 A2. No code change in this round.

myungjoo:

CI fix at head `95bc4805`: all three Tizen GBS builds failed while compiling the new test. Tizen's protobuf 33.1 declares a top-level `namespace pb` in `google/protobuf/extension_set.h`, so the test's alias `namespace pb = nnstreamer::protobuf;` conflicted with it (`error: 'namespace pb = nnstreamer::nnstreamer::protobuf;' conflicts with a previous declaration`). protobuf 3.20, which pdebuild and the local build use, has no such namespace. The alias is renamed to `nns_pb` (6 lines, test file only, folded into the test commit). The test recompiles and its 25 cases pass locally; no production code changed.

myungjoo:

Third-round review at head `95bc4805`. This round's stated purpose was the CI fix ("Tizen GBS failed while compiling the new test... protobuf 33.1 declares a top-level `namespace pb`... the alias is renamed to `nns_pb`"), so I focused on (a) verifying that fix and (b) hunting for any other way the new code/test could build or behave differently on the Tizen toolchain (flatbuffers 24.3.25, protobuf 33.1, gcc on aarch64/armv7l, `unit_test 0` compile-but-don't-run builds) than on Ubuntu (flatbuffers 23.5.26, protobuf 3.20), on top of an independent re-read of the whole PR.

## CI fix verification

`git diff 62ae67ef 95bc4805` is exactly the claimed mechanical rename, entirely inside `tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc` (6 lines changed, +6/-6): `namespace pb = nnstreamer::protobuf;` → `namespace nns_pb = ...`, plus 5 call sites (`pb::Tensors`, `pb::Tensors::frame_rate`, `pb::Tensors::Tensor_format` cast, `pb::Tensor *tensor`, `pb::Tensor::Tensor_type` cast) updated to `nns_pb::`. I grepped the resulting file: zero remaining `pb::` occurrences, exactly 5 `nns_pb::` occurrences, matching the diff 1:1. No production file touched. This is a correct, complete, minimally-scoped fix for a namespace-alias collision — nothing to add.

## Tizen-toolchain-specific pass (no new issues found)

- **protobuf 33.1 generated-API usage in the diff**: `ParseFromArray()` return-value check, `num_tensor()`/`format()`/`tensor_size()`/`tensor(i)`/`dimension_size()`/`dimension(j)`/`data().length()`/`data().data()` are all long-stable proto3 generated accessors (plain field / repeated-field / nested-message-field getters), none of them part of the API surface protobuf has deprecated or removed going into the 22+ (arena-first) generation. I found no use of anything protobuf 33.1 would reject that protobuf 3.20 accepts.
- **flatbuffers 24.3.25 vs 23.5.26**: `VerifyTensorsBuffer()` is flatc-generated from `root_type Tensors;` in `nnstreamer.fbs` (confirmed) — a codegen feature stable since well before 2.0. `flatbuffers::Verifier(data, size)`, `flexbuffers::VerifyBuffer()`, `Vector::size()`/`Get()`, `String::size()`/`c_str()` are all pre-2.0.6-stable APIs; the one genuinely version-gated call (`flexbuffers::VerifyBuffer`, needs ≥2.0.6) is already correctly guarded by `#if FLATBUFFERS_VERSION_MAJOR...` in `tensor_converter_flexbuf.cc` and both 23.5.26 and 24.3.25 satisfy the guard. Nothing new here beyond round 1's already-addressed finding.
- **armv7l (32-bit) width / format-string check**: the repo promotes `-Werror=format=2` (`meson.build:52`), which is exactly the class of bug that would surface on a 32-bit target if a `gsize`/`size_t` argument were printed with `%u`/`%d` instead of `%zu`. I checked every new/changed log call in `nnstreamer_protobuf.cc` and `tensor_converter_util.c`: `%zu` is used consistently for the `gsize size`/`expected`/`hsize` arguments, `%u` for `guint` (`config->info.num_tensors`, loop index `i`), `%d` for `int` (`tensors.tensor_size()`, `NNS_TENSOR_SIZE_LIMIT`). No mismatches, so no new `-Werror=format=2` failure on armv7l where `size_t` is 32-bit rather than 64-bit. The test file's own `g_strdup_printf ("tensor_%u", i)` also matches (`i` is `guint`).
- **`unit_test 0` compile-but-not-run builds**: since GBS compiles this test binary regardless of the `unit_test` define (per `packaging/nnstreamer.spec:1085`-ish gating only what *runs*, not what *builds*), the namespace clash this round fixed would have broken every Tizen GBS build (armv7l/aarch64, `unit_test` 0 or 1) even when nobody ever executes the test — consistent with the reported "all three Tizen GBS builds failed" before the fix, and with `unit_test 0` no longer being a way to dodge a compile-time regression here going forward.
- I re-diffed `62ae67ef..95bc4805` in full (not just the stat) to confirm no production file (`ext/nnstreamer/extra/nnstreamer_protobuf.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_{flatbuf,flexbuf,util}.{cc,c,h}`, `tests/meson.build`, `custom_decoder.py`) changed in this round — the resource-management, bounds-checking, and CI-gating conclusions from round 2 (mergeable as-is) still hold unchanged.

## Independent full-PR re-check

Re-verified the items from rounds 1–2 hold at this head: `out_buf` NULL-init/unref-on-error paths in both `fbc_convert` and `gst_tensor_converter_protobuf` are still correct (traced again by hand — `out_buf` is only unreffed once it has actually been allocated, and set back to `NULL` on every new error branch); the `tcu_check_tensor_data()`/`_check_tensor_data()` duplication still carries the cross-referencing `@note` in both copies; `tests/meson.build`'s gating still runs this binary through the pdebuild/GBS legs even though the quick Ubuntu-meson job skips it; no public/installed header or architecture changed, so no doc update is owed. The round-2 Low (integer overflow in `gst_tensor_get_element_count()` letting a crafted huge dimension pair wrap the size this PR's new checks compare against) is unchanged and, per the author, intentionally deferred to #4960 A2 as the single place that helper is shared by every "declared size == mapped size" check in the tree (including the pre-existing `tensor_decoder` check this PR mirrors) — I agree a converter-local patch here would just be a third copy of logic that belongs in one place.

## Verdict

No new Critical/High/Medium/Low/Nit issues found in this round. The CI fix is correct, minimal, and test-only; I found no other Tizen-toolchain-specific (protobuf 33.1 / flatbuffers 24.3.25 / gcc aarch64-armv7l / unit_test-0-compile-only) way for this diff to build or behave differently than on Ubuntu.

Verdict: mergeable as-is

myungjoo:

All three NITs are handled at head `055556e8`.

**NIT 1 (zero-tensor message): fixed, and the trace is confirmed.** Reproduced on a build of the reviewed head: a flexbuf message with `num_tensors = 0` through `filesrc ! other/flexbuf ! tensor_converter ! fakesink` gives
`CRITICAL: gst_pad_set_caps: assertion 'caps != NULL && gst_caps_is_fixed (caps)' failed`,
and gst-launch still exits 0, so the empty buffer does flow under the stale caps. The cause is `_get_tensors_caps ()` (`nnstreamer_plugin_api_impl.c:985`), which fills `num_tensors` only when it is `> 0`, leaving the template range `[1, 256]`. All three converters now refuse `num_tensors == 0` next to the limit check, which also matches what the protobuf decoder already does (`nnstreamer_protobuf.cc:56`). After the fix the same pipeline fails cleanly with `The number of tensors is limited to 1 to 256` and exit 1. New test `zeroTensors_n` runs all three formats; built against the previous converters it fails, so it pins the behavior.

**NIT 2 (flatbuffers 2.0.0-2.0.5): meson now warns.** `meson.build` emits a `warning()` when the flatbuffers found is older than 2.0.6, naming the missing flexbuffers verifier. I left the `>= 2.0.0` floor alone: raising it would change the build requirement for distributions I cannot survey here, while the warning makes the gap visible in the configure log of exactly those builds.

**NIT 3a (three copies): kept, with the cross-references.** Compiling `../tensor_converter/tensor_converter_util.c` into `nnstreamer_protobuf` also exports `tcu_get_out_config` from that library, which the protobuf *decoder* sub-plugin links as well, and the converter sub-plugin would then see the same symbols from two installed libraries. That seemed a worse trade than 40 duplicated lines that the parameterized tests exercise in both copies. If you would rather have one definition, the clean way is a private `static inline` header under `ext/nnstreamer/extra/` (all three files already include from that directory, so no build-graph change) - say the word and I will do it in this PR.

**NIT 3b (documentation): added.** The flexbuffers section of `gsttensor_converter.md` now states that a blob must hold exactly the tensor's bytes, and names `array.tobytes()` for Python producers.

Local verification of this round: 26/26 in `unittest_converter_serialized`, 16/16 in the existing `unittest_plugins` serialized cases, and 284/284 SSAT cases across the flatbuf, flexbuf, protobuf, decoder, converter and the two python3 suites. clang-format, gst-indent and the doxygen-tag, indent, newline, hardcoded-path, misspelling and prohibited-words checkers all report no diff or error.

myungjoo:

Fifth-round review at head `055556e8`. I read the full diff against `main` (`ext/nnstreamer/extra/nnstreamer_protobuf.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_{flatbuf,flexbuf,util}.{cc,c,h}`, `gst/nnstreamer/elements/gsttensor_converter.md`, `meson.build`, `tests/meson.build`, `tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc`, `tests/test_models/models/custom_decoder.py`), the full comment/review thread (rounds 1-4 plus the myungjoo-bot review and the author's reply at this head), and diffed this round's delta specifically (`git diff 95bc4805..055556e8`, 113 lines) against the three NITs it claims to close.

## This round's delta, verified against the three NITs

**NIT 1 (zero-tensor message accepted): correctly fixed in all three converters.**
- `tensor_converter_flatbuf.cc:107`, `tensor_converter_flexbuf.cc:124`, `nnstreamer_protobuf.cc:210` each now read `config->info.num_tensors == 0 || ...` immediately after `num_tensors`/`format` are copied out of the verified message and strictly before the tensor vector/map is indexed or the per-tensor loop starts — every path reaches it (verification failures already `goto done` earlier; nothing after this point bypasses it). On the `== 0` branch, `out_buf` is still `NULL` (it is only assigned by `gst_buffer_new ()` further down), so the shared `done:` label unmaps/unrefs `in_mem` and returns `NULL` with no leak, matching the pattern already relied on by the other error branches from earlier rounds.
- The fix is consistent with the codebase's existing invariant that `num_tensors` must be in `[1, NNS_TENSOR_SIZE_LIMIT]` — `gst_tensors_info_validate ()` (`gst/nnstreamer/nnstreamer_plugin_api_util_impl.c:457`) and `nnstreamer_plugin_api_impl.c:1552` already reject `num_tensors < 1`. So this isn't a new restriction on the wire format, it's the converters catching, before caps negotiation, a case the rest of the tree already treated as invalid. I found no code path (flexible/sparse formats included, since `num_tensors` in these serialized formats always equals the count of `Tensor` entries actually carried, regardless of static/flexible/sparse) and no in-tree producer that emits `num_tensor = 0` deliberately, so I see no legitimate stream this newly rejects.
- Log wording is grammatically slightly awkward ("The number of tensors is limited to 1 to %d", "...and 1 to %d are allowed") but understandable; Nit only, see below.

**NIT 2 (flatbuffers < 2.0.6 unverified): the new meson `warning()` is correctly placed and cannot fail the build.**
- `meson.build:181-183` adds `if not dependency('flatbuffers', version: '>=2.0.6', required: false).found() : warning(...)` inside the existing `if flatc.found()` block. Since `flatbuf-support` is the single option gating both flatbuf and flexbuf (confirmed: `ext/nnstreamer/tensor_converter/meson.build` has no separate flexbuf option), the warning fires exactly when flexbuf is actually being built, and is skipped when flatbuf-support is disabled (flatc not found) — matching the review question "does it run in every configuration including when flatbuf support is disabled." `required: false` means a not-found/too-old dependency can never turn into a configure error; `warning()` only logs. Correct and safe.
- Minor Nit: the block already holds `flatbuf_version_check_dep = dependency('flatbuffers', version: '>=2.0.0', ...)` two lines above; the new code does a third, independent `dependency('flatbuffers', ...)` lookup instead of reusing `flatbuf_version_check_dep.version().version_compare('<2.0.6')`. Functionally harmless (meson dependency caching makes the extra pkg-config call cheap), just slightly redundant.

**NIT 3b (doc): the one-line addition to `gst/nnstreamer/elements/gsttensor_converter.md` is accurate and correctly scoped**, stating the blob-must-hold-exact-bytes rule and naming `array.tobytes()` for Python producers, placed right in the flexbuffers section — this is the only doc obligation this PR has (no public/installed API or architecture changed elsewhere), and it's satisfied.

**NIT 3a (three copies of the size check): kept by design, with the tradeoff explained in the author's reply** (linking `tensor_converter_util.c` into `nnstreamer_protobuf` would leak `tcu_get_out_config` into a second installed library). No code change requested here was skipped silently; the author offered to switch to a shared static-inline header if preferred. I have no objection to leaving this as-is.

## New test (`zeroTensors_n`)

`tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc:628-654` iterates `available_modes ()` (flatbuf, flexbuf, protobuf — whichever are compiled in) and calls `build_serialized_message (mode, 0, 30, 1, _NNS_TENSOR_FORMAT_STATIC, {})`. I traced all three builders (`build_flatbuf`, `build_flexbuf`, `build_protobuf`) with `num_tensor = 0` and an empty `tensors` vector: each produces a well-formed, verifier-passing message with an empty tensor list (the per-tensor loops in the builders simply don't execute; `build_flatbuf` is called with `has_fr = TRUE, has_tensor_field = TRUE`, so the flatbuf case still has a valid, empty tensor vector rather than a missing field, meaning the `== 0` check — not the earlier `!tensor` check — is what actually rejects it). The test asserts `out_buf == NULL` and frees `in_buf`/`config` unconditionally, so it correctly pins the new behavior for all three formats and, built against a converter without this round's fix, would fail (in_buf would still be consumed but `out_buf` would be non-NULL with an empty, num_tensors=0 config). Doxygen `@brief`/`@note` present, satisfying the per-function doxygen gate.

## Independent full-PR re-check

Re-diffed `main..055556e8` in full (10 files, matching every round-1-4 review's file list exactly — no new file touched, no scope creep). Re-confirmed the items already settled in rounds 1-4 still hold at this head: `Verifier`/`VerifyTensorsBuffer`/`VerifyBuffer` gating, `num_tensor` vs. tensor-vector-length bounding, `out_buf` NULL-init/unref-on-error paths (no double-free/UAF on any branch, including the new zero-tensor branch), the `tcu_check_tensor_data ()`/`_check_tensor_data ()` cross-referenced duplication, and the `unittest_converter_serialized` gating through the pdebuild/GBS legs (not the quick ubuntu-meson job) so a regression here still fails required CI checks and blocks merge. The round-2 Low (integer overflow in `gst_tensor_get_element_count ()`, deferred to #4960 A2) is unrelated to this round's changes and remains correctly out of scope. No architecture or public/installed API changed, so no further doc update is owed beyond the one already made.

## Findings

1. **[Nit] Log wording.** `tensor_converter_flatbuf.cc:108`, `tensor_converter_flexbuf.cc:125`: "The number of tensors is limited to 1 to %d" reads awkwardly. `nnstreamer_protobuf.cc:212-213`: "...and 1 to %d are allowed" likewise. Suggest e.g. "The number of tensors must be between 1 and %d" for clarity. Cosmetic only, not blocking.
2. **[Nit] Redundant dependency() lookup in meson.build.** `meson.build:181`: `dependency('flatbuffers', version: '>=2.0.6', required: false)` re-queries a dependency already looked up two lines above as `flatbuf_version_check_dep` (`>=2.0.0`). Could instead be `flatbuf_version_check_dep.found() and flatbuf_version_check_dep.version().version_compare('<2.0.6')`, avoiding a third pkg-config invocation for the same package. Not a correctness issue.

No Critical/High/Medium issues found. This round's changes correctly and completely close all three NITs raised by the myungjoo-bot review, the new test genuinely pins the zero-tensor rejection for all three formats, the meson warning is well-formed meson that cannot fail a build and fires in exactly the configurations where it matters, and the diff stays tightly scoped to the three converters, the shared util comments, one doc line, and the test.

Verdict: mergeable as-is

myungjoo:

Both NITs of the fifth round are taken, at head `2d9bf853` (4 lines, folded into their commits).

**Log wording.** "limited to 1 to %d" / "1 to %d are allowed" now read "The number of tensors must be between 1 and %d" and "... and between 1 and %d are allowed".

**meson lookup.** The third pkg-config query is gone; the check now reuses the dependency fetched two lines above: `flatbuf_version_check_dep.found () and flatbuf_version_check_dep.version ().version_compare ('<2.0.6')`. Verified in a container three ways: with flatbuffers 23.5.26 no warning is printed; with the comparison temporarily flipped to `<99.0.0` the warning appears at `meson.build:182`; and with `-Dflatbuf-support=disabled` the configure stays clean and silent (the block sits inside `if flatc.found()`, and `.found()` guards the disabler case).

Re-verified after the change: 26/26 in `unittest_converter_serialized`, 16/16 in the existing serialized cases of `unittest_plugins`, 284/284 SSAT cases in the flatbuf, flexbuf, protobuf, decoder, converter and both python3 suites, and no diff from clang-format, gst-indent, or the doxygen-tag, indent, newline, hardcoded-path, misspelling and prohibited-words checkers.

myungjoo:

Sixth-round review at head `2d9bf853`. This round's stated purpose was closing the two Nits from round 5 (log wording, redundant `dependency()` lookup). I diffed `055556e8..2d9bf853` directly (4 lines, no other file touched) and verified each change, then re-confirmed PR scope and CI status.

## This round's delta, verified line by line

`git diff 055556e8..2d9bf853` touches exactly:
- `ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc:108`, `tensor_converter_flexbuf.cc:125`: `"The number of tensors is limited to 1 to %d"` → `"The number of tensors must be between 1 and %d"`.
- `ext/nnstreamer/extra/nnstreamer_protobuf.cc:213`: `"...and 1 to %d are allowed"` → `"...and between 1 and %d are allowed"`.
- `meson.build:181`: `if not dependency('flatbuffers', version: '>=2.0.6', required: false).found()` → `if flatbuf_version_check_dep.found() and flatbuf_version_check_dep.version().version_compare('<2.0.6')`.

**Log wording**: all three messages keep the same `%d`/`%u` specifier count, order and argument types as before (`nns_loge` call sites' argument lists are byte-for-byte unchanged, only the format string literal was edited) — confirmed by re-reading each call site. `NNS_TENSOR_SIZE_LIMIT` is `#define NNS_TENSOR_SIZE_LIMIT (256)` (`gst/nnstreamer/include/tensor_typedef.h:42`), a plain `int`-width macro, so `%d` is correct and unchanged. No `-Werror=format=2` risk from this edit.

**meson lookup**: `meson.build:170` initializes `flatbuf_version_check_dep = disabler()` at file scope, but the new expression at line 181 sits inside `if flatc.found()` (line 172), strictly after `flatbuf_version_check_dep` is reassigned at lines 179-180 to a real `dependency('flatbuffers', version: '>=2.0.0', required: get_option('flatbuf-support'))` result — so the disabler-initialized value is never observed by the new code, regardless of whether flatbuf-support is enabled/auto/disabled. Meson's `and` is short-circuiting, so when `required: get_option('flatbuf-support'))` yields a not-found dependency (`.found()` → `false`, e.g. flatbuf-support=auto/disabled and flatbuffers isn't installed, or flatbuf-support=disabled with flatc still on PATH), `.version().version_compare(...)` is never evaluated — no call on an "unknown"-version or disabler object. `.version()`/`.version_compare()` and short-circuit `and` have been available since well before this project's declared floor (`meson_version: '>=0.57.0'`). I exercised the three paths structurally (found+old, found+new, not-found) by reading the code path rather than a live meson run (no meson/flatbuffers toolchain in this environment), consistent with the author's own three-way container verification reported in their reply. Correct, cannot fail a build (`warning()` only logs), and strictly less redundant than the code it replaces (one `dependency()` lookup instead of two).

No issues found in this round's delta.

## Independent full-PR re-check

`gh pr diff 4969 --name-only` confirms the PR still touches exactly the same 10 files as every prior round (`ext/nnstreamer/extra/nnstreamer_protobuf.cc`, `ext/nnstreamer/tensor_converter/tensor_converter_{flatbuf,flexbuf,util}.{cc,c,h}`, `gst/nnstreamer/elements/gsttensor_converter.md`, `meson.build`, `tests/meson.build`, `tests/nnstreamer_converter_serialized/unittest_converter_serialized.cc`, `tests/test_models/models/custom_decoder.py`) — no scope creep. (A local worktree `main..2d9bf853` diff briefly looked larger because that worktree's local `main` ref is stale for unrelated reasons; `gh pr diff` against the live PR is authoritative and matches the file list above.)

All conclusions from rounds 1-5 still hold at this head: `Verifier`/`VerifyTensorsBuffer`/`VerifyBuffer` gating and the flatbuffers<2.0.6 comment+warning, `num_tensor` bounds (including the round-5 zero-tensor rejection, `zeroTensors_n` test) in all three converters, `out_buf` NULL-init/unref-on-error with no double-free/UAF on any branch, the cross-referenced `tcu_check_tensor_data()`/`_check_tensor_data()` duplication (kept by design, documented trade-off), the doc line in `gsttensor_converter.md`, and `tests/meson.build`'s gating of the new test through the pdebuild/GBS legs (not the quick ubuntu-meson job) so a regression here still fails required CI checks. The round-2 Low (integer overflow in `gst_tensor_get_element_count()`) remains correctly out of scope, deferred to #4960 A2, and is unrelated to anything touched in this or the prior two rounds. No public/installed API or architecture changed, so no further doc update is owed.

CI on this head is still running (`gh pr checks 4969`: all jobs "pending", only DCO has reported "pass" so far) — nothing to flag from results not yet in, but this is not yet a fully green head.

## Verdict

No Critical/High/Medium/Low/Nit issues found in this round's 4-line delta or in the rest of the PR. Both Nits from round 5 are correctly and completely closed, the diff remains tightly scoped to the three converters, the shared util, one doc line, meson, and tests.

Verdict: mergeable as-is

myungjoo:

Thanks for the approval. On the remaining NIT (no `CHANGES` entry for the flexbuf blob behaviour change): not taken, by project policy. `CHANGES` is not updated per PR. It is written when a version is released, from the history accumulated up to that release, so this change will be described when 2.7.1 (or the next release) is cut. The `2.6.0 -> 2.7.0` step itself was a version re-bump without changes, so that section is meant to carry no entries. The behaviour change stays recorded where release-note authors will find it: this PR's body, the flexbuf commit message, and the flexbuffers section of `gst/nnstreamer/elements/gsttensor_converter.md`. No code or document change in this round.

PR #4970, opened 2026-09-15
[Decoder/BoundingBox] Bound box-properties decoding by the tensors, anchors and priors it has (#4920 G7, G8, G9)

myungjoo:
- **What:** fixes three bounding-box box-properties defects from #4920 in which a model output or option3 decides how far `decode()` reads. **G7:** mobilenet-ssd-postprocess used the model's detection count unchecked, so it could read past its tensors or abort on a negative/NaN count. **G8:** mp-palm-detection indexed an empty anchor array when option3 was missing, kept stale anchors across option3 changes, and read strides past the 13-entry table. **G9:** mobilenet-ssd read box priors that were never loaded, and leaked the option vector and the prior path.
- **How:** each mode now bounds its detection loop by what actually exists: G7 by the tensor length (`max_detection`), G8 by the anchor count, G9 by the number of priors loaded. `checkCompatible()` refuses a stream that has more detections than anchors or priors. `decode()` applies the same bound again, because option3 is writable in PLAYING. mp-palm's option3 is refused, with the previous anchors kept, unless it has 1–7 layers and each layer's stride is positive. The anchors are also cleared before each regeneration. The option documentation states the new limits.
- **Result:** 13 new gtest cases and 1 new SSAT case. 10 of the 13 gtest cases fail on main (all 9 negative cases plus `palmOptionReplacesAnchors`): 2 crash (SIGSEGV on a guard page, g_error abort) and 8 fail an assertion. Reverting single hunks one at a time: 10 of 12 fail a gtest case. The strides zero-init is caught by the valgrind gate as an uninitialised read, and the upper layer bound is an equivalent mutant. Both leak fixes show up in the leak report when reverted. Locally: gtest 30/30 (also under valgrind: 0 errors and 0 definite leaks from this repository), bounding-box SSAT 63/63, tensor_region SSAT 4/4. Existing golden outputs are unchanged. Pipelines that give option3 as documented behave as before. mp-palm or mobilenet-ssd without option3, previously a crash or garbage, now fail negotiation.

---

### For reviewers

**Behavior change.** For mp-palm-detection and mobilenet-ssd, option3 is documented as mandatory. A stream without it, or with fewer anchors/priors than the model's detections, is now refused at negotiation. Before this change it was an out-of-bounds or uninitialised read. `option3=0.5` (threshold only, default layers and strides) still works; a new SSAT case checks it against the existing golden output.

**Commits** (each one builds and passes the gtest cases and the bounding-box SSAT on its own):
1. G7: `mobilenetssdpp.cc`, 3 cases
2. G8: `mppalmdetection.cc`, option docs, 6 cases plus the `option3=0.5` SSAT case
3. G9: `mobilenetssd.cc`, option docs, 4 cases. `swapModeOfConfiguredDecoder_n` now swaps to `yolov10`, because it relies on a mode that no case in the binary configures.

- **Test design constraint.** Box properties are process-wide singletons (#4920 G10, not fixed here). `checkCompatible()` keeps the first `max_detection` it accepts. So every case in the binary uses one detection count per mode: ssd-pp 1, mp-palm 72, mobilenet-ssd 4. Every mp-palm option3 in the tests is a full string. The option3-less mp-palm case is `EXPECT_EXIT` with `death_test_style = "threadsafe"`, which re-executes the binary so the singleton is fresh. It passes under valgrind locally.
- **Guard pages.** The G7 tensors are placed right in front of a `PROT_NONE` page, so the over-read faults without valgrind.
- **Discriminating anchor test.** mp-palm cases use stride 32, which gives 72 anchors, equal to the 72 detections. Stride 64 gives 18. The detection sits at index 28, which only the stride-32 anchors reach.
- **Refusal semantics.** mobilenet-ssd parses the prior file into its table in place, so a failed or unreadable load sets the count to 0 instead of keeping the previous priors. mp-palm validates option3 before assigning anything, so a refused option3 keeps the previous state.
- **Mutation matrix** (fix applied, one hunk reverted):

| Hunk | Result |
|---|---|
| ssdpp clamp | crash |
| palm decode MIN | `palmAnchorsReducedWhileDecoding_n` fails |
| anchors reset | 3 palm cases fail |
| layers range | `palmInvalidLayers_n` fails |
| stride check | `palmInvalidLayers_n` fails |
| palm caps check | `palmFewerAnchors_n` and `palmWithoutOption_n` fail |
| ssd caps check | 2 ssd cases fail |
| ssd decode MIN | `ssdPriorsReducedWhileDecoding_n` fails |
| count reset | `ssdFailedPriorLoad_n` fails |
| count on failure | `ssdFailedPriorLoad_n` fails |
| strides `= { 0 }` | no gtest failure; the valgrind gate reports `mppalmdetection.cc:284` |
| upper bound 7→13 | equivalent mutant (the stride check refuses the same inputs) |

- **Out of scope:** G10 (per-decoder box properties). tensor_region has its own mobilenet-ssd prior loader, which this PR does not touch.
</details>


myungjoo:

I reviewed the diff (`ext/nnstreamer/tensor_decoder/box_properties/{mobilenetssd,mobilenetssdpp,mppalmdetection}.cc`, `tensordec-boundingbox.{cc,h}`, and the two `tests/nnstreamer_decoder_boundingbox/*` files), the code paths they touch (`BoundingBox::getOutCaps`/`checkCompatible`/`decode` in `tensordec-boundingbox.cc`, `tensordec-tensor_region.c`'s independent prior loader), and issue #4920's G7/G8/G9 (and G10) descriptions. I traced every changed line against its callers and the tensor-dimension checks in `checkCompatible()` rather than just reading the diff in isolation.

## Does it solve the stated topic?

Yes, for all three items:

- **G7** (`mobilenetssdpp.cc`): the model-supplied detection count is now read as `gdouble`, compared `> 0`, and clamped with `MIN(num_declared, max_detection)` before being used as the loop bound / `g_array_sized_new()` size. This correctly bounds reads of `classes_`, `scores_`, `boxes_`, all of which are sized by `max_detection` per `checkCompatible()` (`dim2[0]`/`dim4[1]`). NaN and negative counts correctly decode 0 detections (NaN `> 0` is false in C).
- **G8** (`mppalmdetection.cc`): `checkCompatible()` now refuses a config whose `max_detection` exceeds `anchors->len`; `decode()`'s macro independently re-clamps with `MIN(max_detection, anchors->len)` since option3 is writable mid-PLAYING without renegotiation (this double-bound pattern is the right call — `checkCompatible` alone is insufficient once a property can change after negotiation). `setOptionInternal()` now validates the requested layer count (1..7, i.e. `PARAMS_MAX-6`) and that every layer in range has a strictly positive stride (falling back to the previously-set stride when the option string doesn't supply one, using exactly the same `noptions > idx` condition the later assignment loop uses, so validation and assignment can't diverge). `mp_palm_detection_generate_anchors()` now clears `anchors` before regenerating, fixing the accumulate-forever bug.
- **G9** (`mobilenetssd.cc`): a new `box_prior_count` tracks how many priors were actually parsed per row (reset to 0 at the top of `mobilenet_ssd_loadBoxPrior()`, only set to `prev_reg` on success), and both `checkCompatible()` and the decode macro bound the loop by `MIN(max_detection, box_prior_count)`. The destructor now frees `box_prior_path` (previously never freed). The dead `return TRUE;` in `setOptionInternal()` that bypassed `g_strfreev(options)` on the success path is removed, so the option vector is now freed on every path through `exit_mobilenet_ssd:`.

I independently re-derived why each fix is correct from the surrounding dimension checks (e.g. `dim1[2]==dim2[1]` for mobilenet-ssd, `dim1[1]==dim2[1]` for mp-palm) rather than trusting the PR description, and didn't find a gap.

## Regressions / impact on other modules

- No other module is touched. `tensordec-tensor_region.c` has its own, separate box-prior loader and is correctly left alone (confirmed by grep — only `mobilenetssd.cc` and `tensordec-tensor_region.c` reference box priors, and only the former is in the diff).
- The behavior change is real and is called out explicitly in the PR body: mp-palm-detection and mobilenet-ssd streams that previously ran with an absent/undersized option3 (previously UB: OOB read, uninitialized read, or a crash) are now refused at caps negotiation. This is a correct and desirable trade (fail closed instead of reading garbage), and is documented in the option3 help text in both `tensordec-boundingbox.cc` and `.h` in the same PR.
- One existing test (`swapModeOfConfiguredDecoder_n`) had its target mode changed from `mobilenet-ssd` to `yolov10`, because the new SSD gtest cases in this same PR now configure the process-wide `mobilenet-ssd` box-properties singleton, which would have broken this test's implicit "this mode was never configured, so its size fields are still 0" assumption. Necessary and correctly explained; see Low-severity note below re: fragility.
- Default/typical option3 usage (e.g. `option3=box-priors.txt` with a full prior file, `option3=0.5` for mp-palm with default layers/strides) is unaffected — verified by re-deriving that the validation logic falls back to previously-set/default stride values exactly the way the assignment loop does, and by the new SSAT case that checks `option3=0.5` against the pre-existing golden output.

## Size vs. topic

Tightly scoped: 3 box-properties implementation files for the actual fixes, the shared header/cc doc comments (required since option3 semantics changed), and the two test files. No unrelated files touched. Appropriate for three related defects sharing one root-cause pattern (an unclamped model/option-controlled loop bound).

## Test completeness

Strong. 13 new gtest cases plus 1 new SSAT case:
- G7 is tested with `mmap`+`mprotect(PROT_NONE)` guard pages so an over-read faults immediately without needing Valgrind (this technique is already established elsewhere in the repo — `unittest_filter_custom.cc`, `unittest_plugins.cc` — so it isn't a new portability risk; CI's build matrix is Linux-only, which is what these POSIX calls need).
- G8 covers: anchors replaced (not accumulated) on new option3, insufficient anchors refused at negotiation, anchors reduced mid-stream bound the decode, invalid layer/stride combinations refused (7 sub-cases), layers without stride keep previously-set strides, and missing option3 refused (via `EXPECT_EXIT`+`threadsafe` death-test re-exec, correctly working around the process-wide singleton so the anchor array starts genuinely empty).
- G9 covers: priors matching detections accepted, fewer priors than detections refused, priors reduced mid-stream bound the decode, and a failed/inconsistent prior reload leaving no usable priors.
- All new tests exercise the real `getOutCaps()`/`decode()` element entry points (`acceptsConfig`/`decodeFrame` helpers), not private internals, so they'd survive future internal refactors of these classes.
- The PR description reports a mutation matrix (each of the 12 fix hunks reverted individually): 10/12 caught by a gtest failure or crash, 1 (`strides = {0}`) caught only by the Valgrind uninitialized-read gate, 1 acknowledged as an equivalent mutant (the 7-vs-13 layer upper bound, made unreachable by the earlier `noptions > PARAMS_MAX` check). That's an unusually rigorous level of self-verification for a fix PR and gives good confidence the tests are not vacuous.

## Will a future regression here fail CI?

Mostly yes, with one caveat:

- The G7/G8 OOB-read and G9 uninitialized-read regressions would reliably fail `meson test -C build/ -v` (crash or `EXPECT_*` failure), which gates the PR build job directly.
- The two G9 **leaks** (`box_prior_path`, the `g_strsplit` option vector) are exercised by the existing `run_unittests_binaries.sh --valgrind` pass over every unit-test binary, but per this repo's current Valgrind-gate design the leak check is warn-only in its current phase (errors of "our own" fail the build; definite leaks of "our own" currently only warn — see PR #4936/#4955). So if one of these two specific leaks is reintroduced by a future change, CI will currently **not** fail the build, only warn. This is a pre-existing project-level gate limitation, not something this PR needs to fix, but it's worth being aware of for the "future regression is caught" criterion — the OOB-read fixes are hard-gated, the leak fixes are soft-gated only.

## Architecture/API/doc

No architecture or public API change. The only "API-adjacent" change is the option3 semantics of two decoder modes (now effectively mandatory / bounded), and the doc comments for option3 in both `tensordec-boundingbox.cc` (the `init_bb()` GObject property blurb) and `tensordec-boundingbox.h` (the file-level doc comment) are updated together, consistently, in the same PR. No separate architecture doc exists for this that would need touching.

## Findings by severity

- **Low** — CI leak-gate is currently warn-only for "definite leaks of our own" (project-wide, not this PR's doing). A future regression of the two G9 leak fixes in this PR would not fail CI today. Worth a follow-up once the leak gate moves past its current phase; not a reason to hold this PR.
- **Low / informational** — `swapModeOfConfiguredDecoder_n` still relies on "no other test case in this binary configures this mode," now pointed at `yolov10` instead of `mobilenet-ssd`. This is the same class of fragility the PR itself documents as inherent to the G10 process-wide-singleton design (explicitly out of scope here). It will need to move again the day another test in this binary configures `yolov10`. Not a defect in this PR, just a maintenance note.
- **Nit** — none of real substance found in the production code; naming/style is consistent with the surrounding file in both C++ files touched.

## Verdict

Mergeable as-is. The three fixes are individually correct (verified against the tensor-dimension checks and callers, not just the diff), properly scoped, backed by unusually thorough test coverage including a self-reported mutation matrix, and the documented behavior change (stricter option3 requirements) is intentional, correctly justified, and documented in the same PR. The two Low-severity notes above are informational and don't block merge.

PR #4971, opened 2026-09-15
[Filter/TF-Lite] Remove the tensorflow2-lite-custom subplugin

### LINKED ISSUE #4268: Cannot open tensorflow2-lite-custom(/usr/lib/nnstreamer/filters/libnnstreamer_filter_tensorflow2-lite-custom.so)
```bash
$ nnstreamer-check
NNStreamer version: 2.4.0
           loaded : TRUE
           path   : /usr/lib/x86_64-linux-gnu/gstreamer-1.0/libnnstreamer.so

NNStreamer configuration:
============================================================
Configuration Loaded: TRUE
Configuration file path: /etc/nnstreamer.ini
    Candidates: envvar(NNSTREAMER_CONF): (null)
                build-config: /etc/nnstreamer.ini
                hard-coded: /etc/nnstreamer.ini
[Common]
  Enable envvar: TRUE
  Enable sym-linked subplugins: TRUE
[Filter]
  Filter paths from .ini: /usr/lib/nnstreamer/filters/
             from envvar: (null)
         from hard-coded: /usr/lib/nnstreamer/filters/

============================================================

** (nnstreamer-check:249927): CRITICAL **: 16:34:15.803: Cannot open tensorflow2-lite-custom(/usr/lib/nnstreamer/filters/libnnstreamer_filter_tensorflow2-lite-custom.so) with error libtensorflow2-lite-custom.so: cannot open shared object file: No such file or directory.

NNStreamer registered sub-plugins:
============================================================

[Filter]
  cpp
    - No custom property found
  tensorflow2-lite
    - NumThreads: Number of threads. Set 0 for default behaviors.
    - Delegate: TF-Lite delegation options: {'NNAPI', 'GPU', 'XNNPACK', 'External'}. Do not specify to disable delegation.
    - ExtDelegateLib: Path to external delegate shared library
    - ExtDelegateKeyVal: key/values pairs optional parameters for delegate. Format ExtDelegateKeyVal=key1#value1;key2#value2...

[Decoder]
  direct_video
    - No custom property found
  pose_estimation
    - No custom property found
  bounding_boxes
    - No custom property found
  image_labeling
    - No custom property found
  octet_stream
    - No custom property found
  tensor_region
    - No custom property found
  image_segment
    - No custom property found

[Conterver]

[Trainer]

============================================================
```

```
$ ldd /usr/lib/nnstreamer/filters/libnnstreamer_filter_tensorflow2-lite-custom.so
	linux-vdso.so.1 (0x00007ffe8e172000)
	libnnstreamer-single.so => /lib/x86_64-linux-gnu/libnnstreamer-single.so (0x00007f9de2105000)
	**libtensorflow2-lite-custom.so => not found**
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f9de1fcb000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f9de1c00000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9de1fab000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9de1800000)
	libgmodule-2.0.so.0 => /lib/x86_64-linux-gnu/libgmodule-2.0.so.0 (0x00007f9de1fa2000)
	libgobject-2.0.so.0 => /lib/x86_64-linux-gnu/libgobject-2.0.so.0 (0x00007f9de1f42000)
	libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007f9de1ecc000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9de1b19000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9de214f000)
	libffi.so.8 => /lib/x86_64-linux-gnu/libffi.so.8 (0x00007f9de1ebd000)

$ LD_PRELOAD=./build/ext/nnstreamer/tensor_filter/libtensorflow2-lite-custom.so nnstreamer-check
nnstreamer-check: symbol lookup error: ./build/ext/nnstreamer/tensor_filter/libtensorflow2-lite-custom.so: undefined symbol: _ZTVN10__cxxabiv117__class_type_infoE
```

myungjoo:
- **What:** Fixes #4268. The `tensorflow2-lite-custom` filter subplugin cannot be loaded once installed, so `nnstreamer-check` and any `framework=tensorflow2-lite-custom` pipeline fail with `Cannot open ... libtensorflow2-lite-custom.so`. It has been disabled by default and excluded from the deb/rpm packages since v2.4.4 (258858d7) as unstable and unmaintained; this PR removes it.
- **How:** Build scripts only. Drops the meson block that copies or converts `libtensorflow2-lite` and builds the subplugin, the `dh_shlibdeps` override and rpm `__requires_exclude` (plus `rpm >= 4.9` BuildRequires) that existed only for the custom library, and the spec flag, and replaces the README section with a note on the removal and the kept option. The `tflite2-custom-support` option itself stays but selects nothing, so out-of-tree build scripts that pass it keep configuring; `-Dtflite2-custom-support=enabled` is refused. The meson CI job asserts both paths. No C/C++ source changes.
- **Result:** On Ubuntu 24.04 with PPA `tensorflow2-lite-dev` 2.15.0, the `tensorflow2-lite` subplugin's ninja build commands are byte-identical to main, `unittest_filter_tensorflow2_lite` passes (7/7), SSAT `nnstreamer_filter_tensorflow2_lite` passes (37/37), and the installed `nnstreamer-check` reports no load error.

### Why remove rather than fix

Reproduced on main with `-Dtflite2-custom-support=enabled`:

| Attempt | Result |
|---|---|
| Installed tree, `nnstreamer-check` / `framework=tensorflow2-lite-custom` | `Cannot open ... libtensorflow2-lite-custom.so` |
| Library placed in the current directory (what the README documented) | Same failure |
| `LD_PRELOAD=libtensorflow2-lite-custom.so` | `undefined symbol _ZTVN10__cxxabiv117__class_type_infoE` |
| `LD_LIBRARY_PATH=.` | Works (the only working path) |

Three defects cause this: the subplugin is linked with `-L<builddir> -ltensorflow2-lite-custom`, so it carries a bare `DT_NEEDED` with no RPATH/RUNPATH on a library that is never installed; the `.a`→`.so` conversion records no libstdc++ dependency; and the generated library has no SONAME, so a preloaded path never satisfies the `DT_NEEDED` name. Fixing it properly means a runtime-loading design plus new CI coverage for a feature nobody ships.

### Why the option outlives the feature

meson stops at `Unknown options` for an option it does not know, so deleting `tflite2-custom-support` would break every build script that passes it — and public ones do: the flexbuild recipes of NXP, ADVANTECH, TechNexion and Variscite all pass `-Dtflite2-custom-support=disabled` (verified by fetching the raw files; `nnstreamer/meta-neural-network` does not use it). Those keep configuring against this PR. Asking for the removed subplugin (`=enabled`) is an error instead of a silent no-op. meson's `option(..., deprecated:)` needs 0.60 while this project's floor is 0.57.0, so a soft deprecation is not available. Dropping the option for good is tracked in #4978, which the option comment, the README note and the CI step comment point to. Since the shim is the only thing in the tree that reads the option, the meson job now configures once with `=disabled` (must succeed) and once with `=enabled` (must fail), so neither path can rot unnoticed.

### Tests

No new test code beyond the CI assertion on the option shim: the rest of this change touches build scripts only, and the defect was link/install layout, not code. A generic "every built subplugin can be opened" gtest would catch this class of bug, but it would load every framework (tf, tflite, pytorch, python, ...) into one process and risk symbol clashes in CI, so it is out of scope here. Existing coverage of the retained `tensorflow2-lite` subplugin (gtest + SSAT, and the GBS/pdebuild packaging jobs for the spec and `debian/rules` edits) guards against collateral damage.

---

- Branch `tflite2-custom-removal-4268` on myungjoo/nnstreamer, base a3d8600c. History: 326de0b3 (removal, 21/21 green) → 35f9082e (+`CHANGES` note, 21/21 green) → 011e316f (option kept as a no-op after the downstream break was judged too severe for a release note) → e01ef5fe (wording, tracking issue #4978, CI assertion on both option paths) → 4f69da9f (CI step: x86 only, refusal text pinned) → current head (`CHANGES` hunk dropped per the release-time `CHANGES` policy in #4983; the out-of-tree note moved to the tensor_filter README).
- Local verification of the removal: `ubuntu:24.04` container + `ppa:nnstreamer/ppa` `tensorflow2-lite-dev` (ships only static `/usr/lib/libtensorflow2-lite.a`). Configure `-Dtflite2-support=enabled -Denable-test=true`; compared `ninja -t commands ext/nnstreamer/tensor_filter/libnnstreamer_filter_tensorflow2-lite.so` main vs branch (15 lines, identical after path normalization); after `ninja install`, no `*tensorflow2-lite-custom*` file on the system and `nnstreamer-check` has 0 `cannot open` lines; a `framework=tensorflow2-lite` mobilenet pipeline reaches EOS.
- Local verification of the option shim, same image: configure with no option, with `=disabled` and with `=auto` all succeed and leave 0 `tensorflow2-lite-custom` occurrences in `build.ninja`; `=enabled` exits 1 at `ext/nnstreamer/tensor_filter/meson.build:166` with the removal message; the tflite2 subplugin still links from the `=disabled` tree.
- Tizen spec: the flag was always `disabled`; removing it and `__requires_exclude` changes no produced RPM content. `debian/rules`: the override only added `--exclude`, so default `dh_shlibdeps` is equivalent.
- Outside the shim (option, guard, CI step) and the README note, the subplugin name remains only in the historical 2.x entry of `CHANGES`, which this PR does not touch.
</details>


myungjoo:

Reviewed the diff for PR #4971 (`debian/rules`, `ext/nnstreamer/tensor_filter/README.md`, `ext/nnstreamer/tensor_filter/meson.build`, `meson_options.txt`, `packaging/nnstreamer.spec`) against the repo, issue #4268, and the rest of the tree (workflows, Documentation/, tests/, Android/Tizen packaging, CHANGES).

**Does it solve the stated problem?**
Yes. Issue #4268 is "Cannot open tensorflow2-lite-custom ... libtensorflow2-lite-custom.so", caused by the subplugin being linked with a bare `-L<builddir> -ltensorflow2-lite-custom` (no RPATH/RUNPATH, library never installed). The PR removes the whole `tflite2-custom-support` option and the meson block that builds it, rather than patching the link line, and the PR body documents why a real fix (RPATH/SONAME/install-path redesign) is out of scope for a subplugin that's been `disabled` by default since v2.4.4 (258858d7) and excluded from deb/rpm packaging since the same commit.

**Completeness of the removal (repo-wide grep):**
- No remaining references to `tflite2-custom-support` / `tflite2_custom_support` anywhere in the tree.
- No remaining references to `tensorflow2-lite-custom` / `tensorflow2_lite_custom` anywhere except `CHANGES` (a historical changelog entry, correctly left untouched).
- `debian/*.install` never listed the custom `.so` (already dropped in 258858d7), so no debian packaging list needs an update.
- No `.github/workflows/*`, `tools/*`, `Documentation/*`, `tests/*`, or Android/Tizen build files reference the option or subplugin.
- `meson.build`'s `fs = import('fs')` object is still used elsewhere in the same file (`fs.is_dir(MEDIAPIPE_HOME)`), so removing the custom block's `fs.is_file(...)` calls does not orphan that import.

**Syntactic/semantic integrity of the meson.build edit:**
Read the resulting file around the edit (lines ~166-264): the `if tflite2_support_is_available ... endif` block is intact and balanced — the retained `shared_library`/`static_library('nnstreamer_filter_tensorflow2-lite', ...)` calls are unchanged, and the removed `else # no tflite2 / if get_option('tflite2-custom-support').enabled() error(...) endif` branch (which only existed to error out for the custom option) is gone without leaving a dangling `else`/`endif`. `tflite2_custom_compile_args = tflite2_compile_args` (only consumed by the deleted block) is correctly removed too.

**debian/rules:** The removed `override_dh_shlibdeps: dh_shlibdeps --exclude=libtensorflow2-lite-custom.so` existed only to keep `dh_shlibdeps` from choking on the uninstalled custom lib's dependency. Since that lib is no longer built at all, falling back to the default `dh_shlibdeps` is equivalent for every other package — verified no other package/library in the tree needed that exclude.

**packaging/nnstreamer.spec:** `BuildRequires: rpm >= 4.9` and `%global __requires_exclude ^libtensorflow2-lite-custom.*$` existed solely to support the custom lib's `%{__requires_exclude}` regex feature (a Tizen feature gated on rpm >= 4.9). Grepped the full spec: no other line depends on rpm >= 4.9 or on `__requires_exclude`, so removing both together is safe. The `-Dtflite2-custom-support=disabled` flag removed from `enable_tf2_lite` was always `disabled` per the PR's own git-blame trace back to 258858d7, so this changes no produced RPM content, consistent with the PR's stated verification.

**Regression risk to other modules:** None found. The change is confined to the tflite2 custom-only path; the retained `tensorflow2-lite` subplugin build rule, its dependency declarations, and every other filter/decoder/framework block in the same `meson.build` are untouched. `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc` and the SSAT `nnstreamer_filter_tensorflow2_lite` suite already exercise the retained subplugin and would catch a regression there; CI's meson build workflows compile with tflite2-support enabled, so any accidental syntax breakage in the edited block would fail CI immediately.

**Size/scope:** Proportionate — the diff only touches the five files directly tied to the removed feature (option declaration, its meson build block, its debian shlibdeps exclude, its rpm requires-exclude/BuildRequires, and its README section). No unrelated modules are touched.

**Test coverage for this change and for future regressions:**
- Minor: The PR intentionally adds no new test code, reasoning that this is a build-script-only removal of a feature that was already disabled/unshipped, and that a generic "every subplugin loads" gtest is out of scope (would load conflicting frameworks in one process). This reasoning holds for *this* change, but the review criteria also ask whether future changes to *other* modules could silently reintroduce this exact defect class (broken subplugin, no RPATH) without failing CI. `nnstreamer-check` returning a nonzero/"cannot open" line for any installed subplugin is exactly the failure mode from #4268, and nothing in CI currently asserts on `nnstreamer-check` output being clean after `ninja install` (the PR description's own manual verification step — "0 `cannot open` lines" — is not automated). This is a pre-existing gap, not introduced by this PR, so it is not blocking, but worth a follow-up issue.
- Nit: Since `-Dtflite2-custom-support` is now an unknown option, any downstream build script (Yocto/Android/vendor CI) still passing that flag will hard-fail at `meson setup` rather than degrading gracefully. The PR body already flags this as expected/desired behavior, which is reasonable for a removed option, but worth calling out explicitly as an intentional breaking change for downstream build scripts in the commit/PR description (it already is, in the "Result" section) — no action needed, just confirming it's intentional and documented.

**Architecture/API docs:** No public API changed (build-option and packaging-only). The one piece of user-facing documentation for this feature (`ext/nnstreamer/tensor_filter/README.md`'s "How to use custom tensorflow-lite binaries" section) is removed in this same PR, so documentation is already in sync — no follow-up doc PR needed.

**Summary of findings:**
- Minor: No automated CI guard (e.g., a check on `nnstreamer-check` output) exists to catch this class of "subplugin fails to load post-install" defect in general; this PR's own removal is fine, but consider filing a follow-up issue for general load-time coverage. Not blocking this PR.
- Nit: Downstream build scripts still passing `-Dtflite2-custom-support` will fail at `meson setup` with "Unknown options" — intentional per the PR description, just confirming it's a deliberate breaking change for the removed option.

No Critical or Major issues found. The removal is complete, self-consistent, proportionate in scope, and does not touch or risk regressing any other module's build or runtime behavior.

Verdict: OK to merge

myungjoo:

Thanks for the review. No code change for either item:

- **Minor: no automated post-install load check.** Not added here. The defect class this PR addresses is gone with the code, and a check that fails when any installed subplugin cannot be opened is a CI-wide change: it has to load every subplugin the meson job installs (llamacpp, ncnn, litert, tensorflow2-lite, armnn, python3, ...) into one `nnstreamer-check` process from the installed-tree step. That can fail for runner-environment reasons unrelated to this removal, so it belongs in its own change if we want it.
- **Nit: `-Dtflite2-custom-support` now fails at `meson setup`.** Intentional. The only in-repo user was `packaging/nnstreamer.spec`, updated here, and the option had been hard-disabled since v2.4.4. A deprecated stub option is not possible because meson's `deprecated:` keyword needs 0.60 and the floor is 0.57.0.

myungjoo:

**Minor (release note for the removed meson option): accepted and fixed.** Head is now 35f9082e, which adds a `CHANGES` entry under `2.6.0 -> 2.7.0`, following the precedent of 85129d6f for an out-of-tree-visible change:

```
	- tensor_filter/tensorflow2-lite-custom subplugin removed
		- The subplugin could not be loaded once installed and has been disabled by default and excluded from the deb/rpm packages since 2.4.4. Use the tensorflow2-lite subplugin.
		- The tflite2-custom-support meson option is removed with it. A build script that still passes -Dtflite2-custom-support=... fails at meson setup with "Unknown options".
```

The commit body gained one line for it; the removal itself is byte-identical to 326de0b3, which was 21/21 green.

I did not independently verify the specific downstream repositories named in the finding, and the note does not name them: the entry is warranted for any out-of-tree build script that passes the option, which is how the existing 2.7.0 entries are worded.

For completeness, the two items from the earlier relayed review still stand as answered above: no automated post-install subplugin-load check in this PR (CI-wide change, belongs on its own), and the hard `Unknown options` failure is intentional, since a deprecated stub option would need meson 0.60 against a 0.57.0 floor.

myungjoo:

Round 3 review on head `35f9082e`, following up on the two prior review rounds (the general PR comment and the formal review on `326de0b3`).

**Scope reviewed this round:** the amendment from `326de0b3` to `35f9082e` (the `CHANGES` entry added to answer the Minor from round 2), plus re-confirming the rest of the PR against the repo, issue #4268, and the meson/debian/rpm build tooling.

## What was checked

- **Non-`CHANGES` files are untouched.** `git diff 326de0b3 35f9082e -- debian/rules ext/nnstreamer/tensor_filter/README.md ext/nnstreamer/tensor_filter/meson.build meson_options.txt packaging/nnstreamer.spec` is empty. The removal logic reviewed and approved in round 2 (21/21 CI green at the time) is byte-identical here; only `CHANGES` and the commit message changed.
- **New `CHANGES` entry, placement.** Inserted directly after the "2.7.0 is a devel version..." line and before the existing "C++ subplugin API ... behavior change" entry, i.e. prepended within the `2.6.0 -> 2.7.0` section. This matches the precedent the round-2 review cited (`85129d6f`), which inserted its own release note at the exact same spot for the same reason (an out-of-tree-visible change needs to be found by vendors scanning the top of the section).
- **Indentation.** `cat -A` on the added lines confirms a single tab + `- ` for the topic line and two tabs + `- ` for each sub-bullet, matching both the sibling entry directly below it and the rest of the `2.6.0 -> 2.7.0` section byte-for-byte (no space/tab mixing).
- **Factual accuracy.** "disabled by default and excluded from the deb/rpm packages since 2.4.4" checks out: `258858d7` ("dist: disable tflite2-custom", which flips the meson default to disabled and drops the debian install entry) is reachable from tag `v2.4.4` and not from any earlier tag. The "Unknown options" failure-mode claim matches `meson_options.txt`/spec's `-Dtflite2-custom-support` removal, and the note correctly points users at the still-shipping `tensorflow2-lite` subplugin.
- **Wording.** Concise, consistent in tone and structure with the neighboring entries in the same section, no ambiguity about what a downstream integrator needs to do (drop the flag).
- **Commit-message hygiene.** Single topic, `[Component] Summary` subject retained, sign-off present, and the body now also lists the `CHANGES` edit as one of the bullets of what changed — consistent with CLAUDE.md's commit-message conventions.

## Findings

### Nit: CI is still `pending` on `35f9082e` as of this review

The amended head triggered a fresh CI run and every workflow shows `pending` at the time of this review (`gh pr checks 4971`). This is not a defect in the change — the non-`CHANGES` diff is proven identical to the already-21/21-green `326de0b3` above, and a doc-only addition to `CHANGES` has no plausible path to a build/test failure — but the PR should not be merged before that fresh run actually finishes green, since CI is what would catch a diff mismatch if one existed. Please re-check before merge.

### On the two carried-over items from prior rounds

- **(a) No automated post-install "every subplugin loads" check.** The author's decline (round 1 → round 2 response) is reasonable and consistent with this project's stated preference for minimum-invasive CI changes: a genuine fix here means loading every installed subplugin (llama.cpp, ncnn, litert, tensorflow2-lite, armnn, python3, ...) into one process, which is its own design problem (symbol clashes, runner-environment flakiness) independent of this removal. Recommend a tracking issue exists for it, but it should not block this PR. Not a Nit against this PR specifically — no action requested here.
- **(b) `-Dtflite2-custom-support` hard-fails at `meson setup` instead of degrading.** Verified: the project's floor is `meson_version: '>=0.57.0'` (`meson.build`) and meson's `option(..., deprecated:)` keyword requires 0.60, so a soft-deprecation stub is not available at the current floor. The hard-fail is the only option short of keeping dead code around, and it is now documented in `CHANGES` per the round-2 fix. Considered closed.

No Critical or Major issues. No new scope creep: the amendment is a 4-line, single-file addition that does exactly what the round-2 review asked for, nothing more.

Verdict: OK to merge

myungjoo:

Follow-up on the downstream claim behind the `CHANGES` entry. I said earlier that I had not checked the named repositories; they have now been verified by fetching the raw files:

- `NXP/flexbuild` `src/apps/ml/nnstreamer.mk` passes `-Dtflite2-custom-support=disabled`, and `configs/sdk.yml` pins nnstreamer at `9cf11e0`, so nothing breaks there until they bump.
- `ADVANTECH-Corp/flexbuild` and `varigit/flexbuild` carry the identical line; neither pins nnstreamer in `configs/sdk.yml`.
- TechNexion's equivalent is `TechNexion/tn_debian_flexbuild` (not a repo named `flexbuild`), same line.
- No other public consumer found. `nnstreamer/meta-neural-network`'s recipe does not pass the option, consistent with the Yocto CI job passing here.

So those recipes must drop the flag at their next nnstreamer bump, which is what the new `CHANGES` entry records. The alternative -- keeping the option as a dead no-op so old scripts still configure -- was not taken: it would leave a permanently misleading option behind for a feature that no longer exists.

myungjoo:

Reclassifying the downstream-break finding: the author judges it a real compatibility problem, not a nit, so a `CHANGES` note alone is not enough. Head is now 011e316f.

**The `tflite2-custom-support` option stays, ignored.** meson stops at `Unknown options` for an option it does not know, so deleting it breaks every build script that passes it, and the four public flexbuild recipes (NXP, ADVANTECH, TechNexion, Variscite) do pass `-Dtflite2-custom-support=disabled`. They now keep configuring. `-Dtflite2-custom-support=enabled` is an error rather than a silent no-op, because the subplugin it asks for no longer exists:

```
ext/nnstreamer/tensor_filter/meson.build:166:2: ERROR: Problem encountered: The tensorflow2-lite-custom subplugin is removed. Drop -Dtflite2-custom-support and use the tensorflow2-lite subplugin.
```

The `CHANGES` entry now says the option is ignored and will be dropped in a later release, instead of announcing a hard failure.

Verified in the same ubuntu:24.04 + PPA `tensorflow2-lite-dev` container:

| configure | result |
|---|---|
| no option | rc=0, 0 `tensorflow2-lite-custom` occurrences in `build.ninja` |
| `-Dtflite2-custom-support=disabled` | rc=0, 0 occurrences |
| `-Dtflite2-custom-support=auto` | rc=0, 0 occurrences |
| `-Dtflite2-custom-support=enabled` | rc=1 with the message above |

and `libnnstreamer_filter_tensorflow2-lite.so` still links from the `=disabled` tree. The removal itself is unchanged from the 21/21-green 35f9082e; the diff against it is the option line, the four-line guard, one `CHANGES` line and the commit body.

myungjoo:

Round 4 review on head `011e316f`, following the reclassification of the "downstream break" finding from round 3 (previously answered with a `CHANGES` note only; the author judged that insufficient and now keeps `tflite2-custom-support` as an ignored no-op option instead of deleting it).

**Scope reviewed this round:** the amendment from `35f9082e` to `011e316f` — `git diff 35f9082e 011e316f` touches only `meson_options.txt` (comment + kept option), `ext/nnstreamer/tensor_filter/meson.build` (+4 lines) and `CHANGES` (1 line reworded). Confirmed this is the complete diff; nothing else moved.

## What holds up

- **The compatibility shim solves the real problem.** `meson_options.txt:8` keeps `option('tflite2-custom-support', type: 'feature', value: 'disabled')`, so a build script that still passes `-Dtflite2-custom-support=disabled` (the four public flexbuild recipes: NXP, ADVANTECH-Corp, varigit, TechNexion) keeps configuring instead of hard-failing at `meson setup` with "Unknown options". `value: 'disabled'` is unchanged from what the option already defaulted to since `258858d7` (verified via `git log -S`), so no default-value regression.
- **The guard's unconditional placement is correct, not a bug.** The 4-line guard at `ext/nnstreamer/tensor_filter/meson.build:165-167` sits before (outside) the `if tflite2_support_is_available` block and fires whenever `-Dtflite2-custom-support=enabled` is passed, regardless of whether `tflite2-support` itself is available or the build is a cross build. I compared this against the pre-removal code (`a3d8600c`): there, the equivalent error only fired *inside* `if tflite2_support_is_available ... and not meson.is_cross_build()`, meaning `-Dtflite2-custom-support=enabled` combined with `tflite2-support=disabled` or a cross build used to silently do nothing (no error, no subplugin, no warning). The new unconditional check is stricter than history, but that is the right direction: a user explicitly asking for a subplugin that no longer exists in the source tree should always get a clear configure-time error, not a silent no-op that depends on unrelated flags. Not a finding — flagging only because the task asked to judge this specifically.
- **`error()` is the right severity, matching this file's own idiom.** Every other "user explicitly asked for an unsatisfiable required feature" case in this same file (pytorch/GCC requirement, edgetpu without tflite/tflite2, tflite2's flatbuffers header check) uses `error()`, not `warning()`, when `.enabled()` is set. Consistent.
- **`auto`/`disabled` genuinely behave as a no-op.** The PR's own verification table (no option / `disabled` / `auto` all rc=0 with 0 `tensorflow2-lite-custom` occurrences in `build.ninja`; only `enabled` is rc=1) matches what the code does: only `.enabled()` is checked, `auto()` is deliberately not routed into the error.
- **Nothing else in the tree reads the option.** `git grep -n tflite2-custom-support\|tflite2_custom` on this head returns exactly the two lines that matter (the `meson_options.txt` declaration and the `meson.build` guard) plus the `CHANGES` history entries. `packaging/nnstreamer.spec` and `debian/rules` only ever pass `-Dtflite2-support=...` (base tflite2), never `-Dtflite2-custom-support`, so packaging does not need to (and does not) pass the now-inert flag.
- **The meson-floor claim checks out.** `meson.build:7` declares `meson_version: '>=0.57.0'`, and meson's `option(..., deprecated:)` keyword was introduced in meson 0.60.0, so a soft-deprecation stub genuinely isn't available at this project's floor. A hard `error()` on `enabled` is the only choice short of carrying real (dead) code.
- **Diff to the 21/21-green `35f9082e` is exactly as described**: one `meson_options.txt` line, the 4-line guard, one `CHANGES` line, and the commit body — verified directly, no other file changed.

## Findings

### Minor: wording — "ignored" overstates what happens for `=enabled`

`meson_options.txt:8` and `CHANGES:6`

Both say the option is "ignored" (kept "so that a build script still passing it keeps configuring"), and only in the next clause note that `-Dtflite2-custom-support=enabled` is an error. Strictly, the option is ignored only for its `disabled`/`auto` values; `enabled` is actively rejected, which is the opposite of being ignored. A reader skimming just the first sentence of the `CHANGES` entry could conclude the option is a harmless no-op in all cases. Suggest rewording to something like "...is now inert for `disabled`/`auto`; `-Dtflite2-custom-support=enabled` still fails, since the subplugin it asks for no longer exists." Not blocking — the very next clause in both places does clarify the `enabled` case correctly.

### Minor: no tracking issue for the promised later removal

`CHANGES:6` says the option "will be dropped in a later release," but neither the commit message nor the PR references a tracking issue for that follow-up. Dead options in this codebase have a habit of outliving their stated removal plans (this option itself sat "disabled due to incompleteness" for years before #4268). Without an issue to find later, "a later release" risks meaning "never." Suggest filing a small tracking issue (or reusing #4268, reopened/relabeled) so the eventual meson-0.60 floor bump or a scheduled cleanup pass can find this and finish the removal.

### Minor: the new guard's negative path has no CI coverage

None of the 21 CI jobs pass `-Dtflite2-custom-support=enabled`, so nothing in CI would catch the guard being silently broken or deleted by a future unrelated edit to this same `meson.build` (e.g., someone reflowing the tflite/tflite2 blocks). This mirrors the pre-existing, already-acknowledged gap from earlier rounds (no automated "subplugin fails post-install" check) but is narrower and cheaper to close here: a one-line addition to any existing meson-based CI job's negative-test step (`meson setup ... -Dtflite2-custom-support=enabled` expected to fail) would pin this down. Given the option is explicitly a temporary, soon-to-be-dropped shim, this is a judgment call the maintainer can reasonably decline (consistent with the project's stated preference for minimum-invasive CI changes) — noting it because the review criteria specifically ask whether a future regression here would be caught, and today it would not be.

### Nit: `error()` fires even for a cross build, unlike the pre-removal code

Worth stating explicitly since it's a behavior delta from history (see "What holds up" above): pre-removal, `-Dtflite2-custom-support=enabled` under `meson.is_cross_build()` silently built nothing and raised no error. Now it always errors. This is very unlikely to affect the four known downstream consumers (they all pass `disabled`), and is arguably a fix to a previously-silent gap, not a regression — flagging only for the record in case a cross-build CI job somewhere relies on the old silent behavior. Grepped `.github/` for `tflite2-custom-support`: no job passes it, so no CI job is affected today.

## Answers to the round-4 focus questions

- **Right call to keep the option as a shim?** Yes — it directly fixes the round-3 finding (downstream `meson setup` breakage) with a minimal, self-contained change.
- **Guard placement (outside `if tflite2_support_is_available`)?** Correct as implemented — see analysis above; it should fire independent of tflite2 availability, and does.
- **`error()` vs `warning()` for `=enabled`?** `error()` is right and consistent with this file's existing idiom for other `feature.enabled()` requirements.
- **Acceptable to carry a dead option?** Yes, as a time-boxed compatibility shim — see the "tracking issue" Minor above for making sure it doesn't become permanent.
- **Meson floor claim (0.60 needed for `deprecated:`, floor is 0.57.0)?** Confirmed accurate against `meson.build:7`.
- **Packaging (spec/debian)?** Correctly untouched — neither ever passed `-Dtflite2-custom-support`, only `-Dtflite2-support`.

No Critical or Major issues. The three Minor findings are all about hardening a temporary, narrow-scope compatibility shim (wording precision, a tracking issue for its removal, and a small CI gap for its negative path) — none block the stated goal of letting existing downstream build scripts keep configuring.

Verdict: OK to merge

myungjoo:

All four round-4 items are addressed in head 1cc07a9c.

- **Minor, "ignored" overstates a no-op.** Reworded. `meson_options.txt` now says the option "selects nothing, and 'enabled' is rejected", `CHANGES` says it "no longer selects anything ... `-Dtflite2-custom-support=enabled` is rejected", and the PR body matches.
- **Minor, nothing tracks the eventual drop.** Filed as #4978 (build/CI, low priority), referenced from both the option comment and the `CHANGES` entry. It records why the option is still there, who the known downstream users are, and what to delete when the time comes.
- **Minor, the guard's negative path has no CI coverage.** Added a step to the meson job: it configures once with `-Dtflite2-custom-support=disabled` (must succeed) and once with `=enabled` (must fail, otherwise the step errors out). That is the only thing in the tree that reads the option, so both halves of the shim are now asserted per PR on both the x86 and arm legs.
- **Nit, the guard also fires under `meson.is_cross_build()`.** Left as is, deliberately. The old code skipped the whole custom block on a cross build, so `=enabled` was silently accepted and produced nothing; refusing it everywhere is the behaviour we want now that the subplugin is gone, and it keeps the rule the same for every build type.

Local re-check in the same ubuntu:24.04 + PPA `tensorflow2-lite-dev` container is unchanged: no option / `=disabled` / `=auto` configure with 0 `tensorflow2-lite-custom` occurrences in `build.ninja`, `=enabled` exits 1 with the removal message, and the tflite2 subplugin still links.

myungjoo:

Round 5 re-review on head `1cc07a9c`, following the round-4 review on `011e316f`. Scope confirmed via `git diff 011e316f 1cc07a9c --stat`: exactly `.github/workflows/ubuntu_clean_meson_build.yml` (+13), `CHANGES` (1 line reworded), `meson_options.txt` (1 line reworded), plus the commit body — matching what the prior response described. No other file moved. This round focuses on the new CI step, since it's the only new executable content.

## The new CI step

```yaml
- name: Removed tflite2-custom option keeps configuring
  if: env.rebuild == '1'
  run: |
    meson setup "${RUNNER_TEMP}/build-optkept" -Dtflite2-custom-support=disabled
    if meson setup "${RUNNER_TEMP}/build-optgone" -Dtflite2-custom-support=enabled; then
      echo "::error::-Dtflite2-custom-support=enabled must fail to configure: the sub-plugin it asks for was removed."
      exit 1
    fi
```

**`set -e` semantics — verified correct, not a false-pass risk.** This repo's workflows use the implicit default shell (`bash -e {0}`, no `pipefail`, per the other steps in this same file). Under `set -e`, a command that is the condition of an `if` does *not* trigger errexit on failure — I confirmed this locally with a throwaway script (`if false; then exit 1; fi; echo reached` prints `reached` and exits 0 under `bash -e`). So:
- If `-Dtflite2-custom-support=enabled` correctly fails to configure (desired), the `if` is false, the `then` block is skipped, and the step exits 0. Good.
- If the shim regresses and `enabled` configures successfully, the `if` is true, `exit 1` runs explicitly, and the step (and job) fails. Good — this is not swallowed by `set -e` or by the `if` construct; the explicit `exit 1` inside the `then` block is what makes this fail loudly, not reliance on the `if`'s own exit status. (The earlier round's summary described this as an `if !` form — the actual code is the positive `if CMD; then error; fi` form, which is equivalent and correct; noting only because the described construct doesn't literally match the file.)
- The first line (`build-optkept`, `disabled`) is a plain statement, not inside an `if`, so if *that* configure regressed (the option stopped keeping the build configuring), `set -e` would abort the script there and fail the step. Good — both halves of the shim are actually pinned.

**Verified live, not just in theory.** At review time this PR's fresh CI run (triggered by `1cc07a9c`) had the `ubuntu-22.04-arm` leg already through this step with a green checkmark, sequenced correctly between "build and unit test" and the failure-only "Upload core"/"Try to show a stack trace" steps (both showed as skipped, i.e. no failure occurred). This confirms the step runs and passes under real conditions, not only under my local re-derivation of the shell semantics. The `ubuntu-22.04` leg was still queued/in progress at review time — please confirm it goes green too before merging (same "wait for the fresh run" caution as round 3, now narrower since only this one job is new).

**State isolation — no risk to later steps.** Both `meson setup` calls target fresh `${RUNNER_TEMP}/build-optkept` and `.../build-optgone` directories, never the `build/` tree used by "build and unit test" and every step after it (SSAT runs, Valgrind, `testlog.txt`/`/cores` uploads). A failed second `meson setup` (the expected/desired outcome) leaves no residue that any later step reads. `RUNNER_TEMP` is per-job on a per-matrix-leg ephemeral runner, so `ubuntu-22.04` and `ubuntu-22.04-arm` can't collide with each other either. This is corroborated by the arm leg's later SSAT steps completing normally right after this step.

**Placement and gating.** Sitting after "build and unit test" and before the failure-only core-dump steps is correct — it only needs an installed toolchain and the tflite2 dev packages already set up by that point, and its own failure is meant to be caught by the same "did this job fail" gates as everything else. `if: env.rebuild == '1'` matches every other conditionally-run step in this job (skip-if-no-rebuild is this file's established pattern) — correct.

## Minor/Nit observations (not blocking)

- **Minor: CI-time cost of two extra full configures.** Neither `meson setup` call narrows any of the other auto-detected optional backends (armnn, ncnn, litert, pytorch, tf, tflite, …), so each repeats the same full compiler-sanity-check and dependency-probe work as the main `meson setup build/ ...` in "build and unit test" — this is now effectively 3 full configures per leg instead of 1. A cheaper form exists: reconfigure the same directory for the second call (`meson setup --reconfigure "${RUNNER_TEMP}/build-optkept" -Dtflite2-custom-support=enabled`) instead of spinning up a second fresh `build-optgone` tree, which would reuse most of the already-cached dependency/compiler results. Given this job also runs Valgrind SSAT suites and is already one of the longer jobs in the matrix, this is worth a look but not worth blocking the PR over — the shim itself is a small, temporary, soon-to-be-removed (#4978) piece of scaffolding.
- **Nit: the first configure (`disabled`) is partly redundant with existing coverage.** Every other CI job that runs `meson setup` without touching this flag already implicitly proves `tflite2-custom-support` at its `disabled` default keeps configuring (the option was never plumbed to fail on `disabled`/`auto`). Passing it explicitly here adds a small amount of otherwise-uncovered value (it pins the "an out-of-tree script that explicitly passes `=disabled` still works" contract, which is the actual downstream scenario named in the PR body for NXP/ADVANTECH/Variscite/TechNexion), so it's a reasonable inclusion for symmetry/readability even if mostly redundant on its own.
- **Nit: no duplicate coverage found elsewhere.** `git grep -n tflite2-custom-support\|tflite2_custom` on this head returns only `meson_options.txt`, the `meson.build` guard, this new workflow step, and the `CHANGES` history entries — so this is genuinely the only place the negative path (`=enabled` must fail) is exercised, and it isn't duplicating an existing check.

## Re-verified from round 4's asks

- **Wording fix.** `meson_options.txt` and `CHANGES` no longer say "ignored" — both now correctly state the option "selects/no longer selects anything" for `disabled`/`auto` and separately that `enabled` "is rejected". This resolves the round-4 Minor about "ignored" overstating the no-op behavior.
- **#4978 cross-reference.** Verified the issue exists, is open, and titled "[Build] Remove the dead tflite2-custom-support meson option" — matches what both the option comment and `CHANGES` say. Resolves the round-4 Minor about a missing tracking issue.
- **Cross-build guard left unconditional.** Deliberately unchanged per the author's response; agree this is correct behavior (a request for a removed subplugin should fail the same way regardless of cross-build), and it was already flagged as a Nit-for-the-record rather than a defect in round 4. No new concern.

## Verdict

No Critical or Major issues. The new CI step is logically sound (`set -e`/`if` interaction verified both by local reproduction and by a live green run on the arm leg), correctly isolated from the rest of the job's build tree, correctly placed and gated, and is the only place in the tree that exercises the shim's negative path. The two points above (reconfigure-reuse to cut CI time, and the first configure's partial redundancy with implicit existing coverage) are optimizations, not defects, for a piece of scaffolding this PR itself intends to delete via #4978.

Verdict: OK to merge (pending the `ubuntu-22.04` leg of this fresh run finishing green, as it was still queued at review time)

myungjoo:

Round-5 items, head is now 19ed7d76.

- **Minor, two full configures in the new step: applied.** The second call is now `meson setup --reconfigure` on the same `${RUNNER_TEMP}/build-optshim` directory, so only the first pays a full dependency probe.
- **Nit, the `=disabled` configure is partly redundant: kept.** It is the half that pins the contract this PR is about, an out-of-tree script passing the option explicitly. The other jobs only cover the default.
- **Nit, no other place reads the option: agreed**, that is why the step exists at all.

Verified in the ubuntu:24.04 + PPA `tensorflow2-lite-dev` container, running the step's exact script under `bash -e`:

- as written: step exits 0; the `--reconfigure` call fails with `ext/nnstreamer/tensor_filter/meson.build:166:2: ERROR: Problem encountered: The tensorflow2-lite-custom subplugin is removed. ...`
- negative control, with the guard deleted from `meson.build`: the step exits 1 and prints the `::error::` line.

So the step fails the job when the shim regresses, rather than only asserting today's behaviour.

myungjoo:

Round 6 re-review on head `19ed7d76`. Scope confirmed via `git diff 1cc07a9c 19ed7d76`: exactly `.github/workflows/ubuntu_clean_meson_build.yml`, +2/-2, replacing two independent `meson setup` calls with one fresh setup followed by `meson setup --reconfigure` on the same directory:

```diff
-        meson setup "${RUNNER_TEMP}/build-optkept" -Dtflite2-custom-support=disabled
-        if meson setup "${RUNNER_TEMP}/build-optgone" -Dtflite2-custom-support=enabled; then
+        meson setup "${RUNNER_TEMP}/build-optshim" -Dtflite2-custom-support=disabled
+        if meson setup --reconfigure "${RUNNER_TEMP}/build-optshim" -Dtflite2-custom-support=enabled; then
```

Nothing else moved — no other file, and no other line in this file.

## Analysis of `--reconfigure`

- **Directory precondition.** `--reconfigure` requires an already-configured build dir. The immediately preceding line (`meson setup ".../build-optshim" -Dtflite2-custom-support=disabled`) is a plain statement, not guarded by `if`, so under this file's shell (`bash -e`, confirmed by the round-5 review and unchanged here) a failure there aborts the script before the `--reconfigure` line runs. The directory is therefore guaranteed to exist and be configured whenever `--reconfigure` executes.
- **Re-run-of-the-job concern: not applicable here.** A failing `--reconfigure` leaving the dir in a state that makes the first assertion meaningless would only matter if the directory could survive into a later run of the same job. Both matrix legs (`ubuntu-22.04`, `ubuntu-22.04-arm`) are GitHub-hosted, ephemeral runners with a per-run `RUNNER_TEMP`; nothing persists across job re-runs. So this failure mode does not exist in this repo's actual CI, only in a hypothetical self-hosted/reused-workspace setup.
- **Meson version support.** This job installs the newest meson from PyPI (`pip install meson ninja`, line 42), not the project's 0.57.0 floor, so the floor's support for `--reconfigure` is not actually exercised by this step. `--reconfigure` combined with `-D<option>=value` is long-standing, stable meson functionality, and the same file already uses the identical idiom (`meson setup --reconfigure build/` at line 332) — so this is a proven, already-trusted pattern in this exact workflow, not a novel risk.
- **Does the second call genuinely apply `=enabled`?** Yes. `meson setup --reconfigure -D<opt>=<val>` re-runs the full `meson.build` interpretation with the option's new value overriding the cached one — this is exactly how `meson configure -D...` / `--reconfigure` are documented to behave, and it is not a partial/incremental re-evaluation that could skip the guard. The guard itself (`ext/nnstreamer/tensor_filter/meson.build:164-166`, `if get_option('tflite2-custom-support').enabled(): error(...)`) sits unconditionally outside any `if tflite2_support_is_available` block, so it does not depend on any state cached from the first (disabled) configure. The author's own container verification (quoted in the response above this review) confirms the reconfigure call fails with the expected `ERROR: ... The tensorflow2-lite-custom subplugin is removed` message, and that deleting the guard makes the step exit 1 with the `::error::` line — i.e., the step still fails the job when the shim regresses.
- **Coverage vs. two independent configures.** Marginally narrower in theory (the `enabled` case is now exercised via reconfigure-of-a-configured-dir rather than a from-scratch configure), but not in any way that matters for what this guard actually does: meson has no notion of a "guard evaluated only on first configure" — every `meson setup`, fresh or `--reconfigure`, re-runs the interpreter over the build definition. There is no plausible bug shape where a fresh configure would hit the `error()` but a reconfigure would not. The CI-time savings (one full dependency/compiler probe instead of two) are a reasonable trade for that theoretical, unrealized coverage difference.

No Critical or Major issues.

### Nit (trivial, not blocking): CI step comment cross-references only #4268, not #4978

`.github/workflows/ubuntu_clean_meson_build.yml`, the comment block immediately above this step, links only `https://github.com/nnstreamer/nnstreamer/issues/4268` (the removal issue). `meson_options.txt` and `CHANGES` both point to #4978 as the tracking issue for actually dropping the option (and, by extension, this CI step) later. Adding the #4978 reference to the workflow comment too would make it easier for whoever closes #4978 to find and delete this step at the same time. Cosmetic only.

## Verdict

Verdict: OK to merge

myungjoo:

Round-6 nit applied in head e01ef5fe: the CI step's comment now cross-references #4978 next to #4268 and says to drop the step together with the option, so whoever closes #4978 finds it. The diff against the reviewed 19ed7d76 is those two comment lines only, with no change to the step's script, so this round did not go back for another full review.

Comment now reads:

```
    # The tensorflow2-lite-custom sub-plugin is gone, but out-of-tree build
    # scripts still pass its option, and meson aborts on an option it does not
    # know. So the option stays behind, selecting nothing, and a request for
    # the removed sub-plugin is refused. Nothing else exercises either path.
    # Drop this step with the option itself.
    # https://github.com/nnstreamer/nnstreamer/issues/4268
    # https://github.com/nnstreamer/nnstreamer/issues/4978
```

With that, every finding from the four review rounds is either applied or answered on this thread. Waiting on the CI run for this head before it is ready to merge.

myungjoo:

Both NITs from the `myungjoo-bot` review of e01ef5fe are applied in head 4f69da9f. The diff against e01ef5fe is the one workflow step only.

- **The negative half accepted any configure failure.** The `--reconfigure ... =enabled` output now goes to a log, and the step additionally requires the guard's own text (`tensorflow2-lite-custom subplugin is removed`) in it. A refusal for any other reason fails the step with its own `::error::`.
- **The step ran on both matrix legs.** It is now gated `env.rebuild == '1' && matrix.os == 'ubuntu-22.04'`, since the check cannot differ by architecture.

The step's `run:` block, extracted from the committed YAML, was run under `bash -e` in an ubuntu:22.04 container in three scenarios:

| scenario | step exit |
|---|---|
| as committed | 0 |
| guard deleted from `meson.build` | 1: `=enabled must fail to configure` |
| guard kept but refusing with different text | 1: `failed to configure, but not because the sub-plugin was removed` |

myungjoo:

Please drop the `CHANGES` hunk from this PR.

`CHANGES` is not updated per PR. The maintainers write it when a version is released, from the history accumulated since the previous release; the `2.6.0 -> 2.7.0` step only re-bumped the version, so that section should hold nothing beyond the devel-version note. #4983 removes the one per-PR entry that got in there (from #4829) and adds this rule to the branch and PR policy in `CLAUDE.md`.

What that means for this PR:
- Remove the `tensor_filter/tensorflow2-lite-custom subplugin removed` entry from `CHANGES`. Its content is not lost: the removal and the reason are in the commit message and the PR body, and the release-note author will pick them up from there.
- Update the PR body where it relies on `CHANGES`: "records the removal in `CHANGES`" in the *How* bullet, and "`CHANGES` points there" for #4978. The pointer to #4978 already lives in the *Why the option outlives the feature* section, which is enough.
- If an in-tree note for out-of-tree build scripts is still wanted, `ext/nnstreamer/tensor_filter/README.md`, which this PR already edits, is the place for the `tflite2-custom-support` shim and its planned removal in #4978.
- Practically, the hunk would also conflict: it uses the lines #4983 deletes as context, so whichever of the two merges second would need a rebase.

myungjoo:

Addresses the `myungjoo-bot` CHANGES_REQUESTED review of 4f69da9f and the earlier request on this thread to drop the `CHANGES` hunk. Head is now b83a11b9, still one commit.

- **`CHANGES` hunk: dropped.** The PR no longer touches `CHANGES` at all (`git diff a3d8600c b83a11b9 -- CHANGES` is empty), so the only mention of the subplugin there is the historical 2.x entry. `git merge-tree` against `refs/pull/4983/head` now reports no conflict.
- **Commit message:** the `- CHANGES: record the removal ...` bullet is gone. The README bullet now describes the new note.
- **PR body:** "records the removal in `CHANGES`" and "`CHANGES` points there" are removed, and the "remains only in `CHANGES`" note is adjusted.
- **In-tree note for out-of-tree build scripts: added to `ext/nnstreamer/tensor_filter/README.md`**, as suggested and as #4983's `CLAUDE.md` rule directs user-visible changes to component documentation. Under *Tensorflow-lite* it says:
  - the `tensorflow2-lite-custom` subplugin is removed (#4268), use `tensorflow2-lite`;
  - `tflite2-custom-support` is kept only so existing build scripts keep configuring, and it selects nothing;
  - `-Dtflite2-custom-support=enabled` is rejected at `meson setup`;
  - the option will be removed later (#4978), so build scripts should drop it.

The diff against 4f69da9f is the 4-line `CHANGES` removal and the 2-line README addition. The workflow step, guard and option line are unchanged.

myungjoo:
Approved by Astra

PR #4972, opened 2026-09-15
[Filter/TF-Lite] Refuse an incompatible shared-model reload before freeing the interpreter

myungjoo:
- **What:** item **F1** of #4920 — reloading `model` on an updatable TF-Lite `tensor_filter` that shares its interpreter via `shared-tensor-filter-key` with a model whose input/output tensors differ left every sharing filter pointing at a freed interpreter (use-after-free at the next invoke), while the reload was reported as successful.
- **How:** `TFLiteCore::reloadModel()` now compares the new interpreter with the current one before calling `nnstreamer_filter_shared_model_replace()` and fails the reload like the non-shared branch already did, so the framework keeps the previous model. The comparison is extracted into `TFLiteInterpreter::hasSameTensorsInfo()` for both branches; the helper's doxygen notes that it frees the old interpreter regardless of the callback.
- **Result:** 3 new cases in `unittest_filter_shared_model`, and the existing reload case now actually observes the reload. Locally (jammy + tensorflow2-lite 2.15, `-Dwerror=true`) 6/6 pass and the `nnstreamer_filter_reload` SSAT passes 2/2. With main's tflite source both shared-key cases fail; two mutations fail the cases that cover them. No other sub-plugin calls the helper.

## Details

**Root cause.** `nnstreamer_filter_shared_model_replace()` (`tensor_filter_common.c`) runs `replace_callback` on every core in the referred list and then calls `free_callback` on the old interpreter unconditionally. TF-Lite's callback, `reloadInterpreter()`, refuses an interpreter with different tensors info and only logs, and `reloadModel()` returned 0 in the shared branch. Trigger: two filters with `is-updatable=TRUE shared-tensor-filter-key=K`, then `g_object_set (filter, "model", <float model>)` on a quantized one.

**Why one check covers all cores.** Every core registered under a key holds the same `shared_interpreter` (`checkSharedInterpreter()` enforces the same model path at open, and every successful replace updates all of them), so the pre-check against `this->interpreter` is the same comparison each callback would make.

**What this does not close.** The pre-check runs outside `G_LOCK (shared_model_table)`, so a reload racing with a renegotiation on another sharing filter (`tflite_setInputDim()` → `setInputTensorsInfo()`) can still change the shared interpreter's info between the check and the callbacks, and the helper would free the old interpreter again. Holding the interpreter lock across the helper is not possible — `reloadInterpreter()` takes the same lock and would deadlock — so closing it needs `replace_callback` to report the refusal so the helper can skip `free_callback`, which changes an exported callback signature. That is recorded as a new item on #4920 and the helper's `@note` says so.

**Tests** (`tests/nnstreamer_filter_shared_model/unittest_filter_shared_model.cc`):
- `tfliteSharedReloadUnmatched_n`: a refused reload keeps the old `model` path and both filters keep the old model's result.
- `tfliteSharedReloadAfterUnmatched`: a later compatible reload still moves both sharing filters to the new model.
- `tfliteReloadUnmatched_n`: the same refusal without a key (guards the extracted comparison).

Two things the reload cases need in order to observe anything: they classify `9.png`, because both quantized mobilenets return the identical output bytes for `orange.png`, and they wait for the buffers that follow the reload (`count_output` + `wait_pipeline_process_buffers`) because the buffer prerolled before it still carries the old result. `tfliteSharedReload` gets the same treatment, which is what lets it notice a reload that never reaches the second filter.

The discriminator for the fix itself is the `model` property, not valgrind: this binary is built only where tensorflow2-lite is installed (Tizen GBS, pdebuild), and those runs use `run_unittests_binaries.sh` without valgrind or a timeout. On the unfixed code the replay after an accepted reload hangs on the freed interpreter's mutex, so the shared-key cases skip the replay when the path check already failed — a regression fails in well under a second instead of stalling the job.

**Local evidence** (docker `jammy`, `ppa:nnstreamer` tensorflow2-lite-dev 2.15.0):

| tflite source | SharedReload | SharedReloadUnmatched_n | SharedReloadAfterUnmatched | ReloadUnmatched_n |
|---|---|---|---|---|
| this PR | pass | pass | pass | pass |
| main | pass | **fail** | **fail** | pass (non-shared branch was already correct) |
| `hasSameTensorsInfo()` forced TRUE | pass | **fail** | **fail** | **fail** |
| shared replace skipped | **fail** | pass | **fail** | pass |

clang-format 16 clean; `doxygen-tag.sh` and `gst-indent` report exactly the same (pre-existing) output as on main for the touched files.


myungjoo:

Scope reviewed: the diff of PR #4972 (`ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc`, `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h`, `gst/nnstreamer/tensor_filter/tensor_filter_common.c` doc-only, `tests/nnstreamer_filter_shared_model/unittest_filter_shared_model.cc`), cross-checked against `tensor_filter_common.c` (`nnstreamer_filter_shared_model_replace`, `checkSharedInterpreter`'s model-path enforcement at open), `gst_tensors_info_is_equal()`, `tests/meson.build`, `packaging/nnstreamer.spec`, `packaging/run_unittests_binaries.sh`, `debian/rules`, and the three CI workflows that can build tflite2 (`ubuntu_clean_meson_build.yml`, `gbs_build.yml`, `pdebuild.yml`).

## Does it solve F1?
Yes. Confirmed the root cause is real: `nnstreamer_filter_shared_model_replace()` runs every core's `replace_callback` and then unconditionally frees the old shared interpreter, while TF-Lite's `reloadInterpreter()` only *logs* a mismatch and leaves `reloadModel()` returning 0 — i.e. the framework reports success while every sharing core is left pointing at freed memory. The fix's core claim ("one pre-check against `this->interpreter` covers every core under the key") is grounded in `checkSharedInterpreter()`: all cores registered under a `shared-tensor-filter-key` are given the *same* `TFLiteInterpreter*` at open (either freshly inserted or fetched via `nnstreamer_filter_shared_model_get`), and a filter with a shared key never takes the non-shared reload branch, so the field stays in lockstep across cores. I traced the concurrent-reload race the PR description dismisses via transitivity, and it holds: because `isCompatibleInterpreter()` is a strict equality check (not a weaker "compatible enough" relation) and every core under a key is updated from the same baseline while `G_LOCK(shared_model_table)` is held, two interleaved reloads that each independently pass their own pre-check are guaranteed pairwise-equal too, so the redundant check still performed inside `reloadInterpreter()` (invoked per-core through `replace_interpreter`) cannot diverge from the outer pre-check's verdict within one `nnstreamer_filter_shared_model_replace()` call.

## Regression / other-module risk
None found. `nnstreamer_filter_shared_model_replace()` is only called from this one file (`grep` confirms no other sub-plugin uses it), so the header/`.c` doc-only comment additions cannot affect any other filter. The non-shared path's behavior is unchanged (still goes through `reloadInterpreter()`); the refactor into `isCompatibleInterpreter()` is a pure extraction and `tfliteReloadUnmatched_n` specifically pins that the non-shared refusal still works after the extraction. Lock discipline is sound: the new pre-check takes only `interpreter->lock()` and releases it *before* `nnstreamer_filter_shared_model_replace()` takes `G_LOCK(shared_model_table)`, so there is no new lock-order/deadlock risk. On the error path, `interpreter_sub` is freed via the existing `goto error; delete interpreter_sub;`, so there's no leak when the new pre-check rejects a reload.

## Scope / size
Appropriately scoped. The functional change is ~25 lines in one file; the rest of the +264/-22 is the 3 new regression tests plus a doc-comment addition on the shared helper's existing header/impl. No unrelated module is touched.

## Findings

**Minor** — CI visibility gap (pre-existing, not introduced by this PR): `.github/workflows/ubuntu_clean_meson_build.yml` deliberately installs `tensorflow2-lite-dev` *after* its `meson test` step (see the comment at line ~314: "installing it before 'build and unit test' would pull tflite unit tests into meson test"), and never re-runs `meson test`/`run_unittests_binaries.sh` afterward. That means `unittest_filter_shared_model` — including this PR's 3 new cases — never executes on that workflow. Coverage instead comes from `gbs_build.yml` (`x86_64`, `--define "unit_test 1"`) and `pdebuild.yml` (`debian/rules` builds with `-Dtflite2-support=enabled` and runs `run_unittests_binaries.sh ./tests` under `dh_auto_test`, which fails the package build — and hence the job — on any gtest failure). Both do trigger on `pull_request`. This is correct and matches the PR body's own claim, but since the most-visible "Ubuntu" check will stay green regardless of this code path, please confirm the GBS/pdebuild checks are part of the required status checks for merge, so a future regression here is actually blocked and not just logged.

**Nit** — `ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc`: the new pre-check in `reloadModel()` (`ml_loge ("The model has unmatched tensors info\n"); goto error;`) and the existing check inside `reloadInterpreter()` log the exact same string. When a shared-key reload is rejected, both sites can fire (the pre-check first, then the redundant per-core check inside `replace_interpreter`/`reloadInterpreter` during `nnstreamer_filter_shared_model_replace`), so log output alone can't tell which one actually stopped the corruption. A distinguishing message (e.g. mention "pre-check" vs "post-check") would help field debugging.

**Nit** — `TFLiteCore::isCompatibleInterpreter()`'s doxygen `@note` says the caller must hold the current interpreter's lock, but doesn't note that `new_interpreter` needs no lock precisely because at every existing call site it isn't visible to other threads yet. Worth a one-line addition so a future caller doesn't assume symmetric locking is required.

## Test coverage
Good. The 3 new gtest cases correctly reproduce the bug (float vs. quantized mobilenet_v2/v1 model swap under a shared key) and are structured to fail fast (`EXPECT_STREQ` on the model path, which alone catches a reintroduced regression) rather than depend on a subsequent `invoke()` that could hang against a freed interpreter on the unfixed code — the follow-up invoke/reload checks are correctly gated behind `if (refused)` so a regression is a ~0.3s failure, not a stalled CI job. `tfliteReloadUnmatched_n` also pins the non-shared branch so the `isCompatibleInterpreter()` extraction itself is covered. The extra test fixture (`mobilenet_v2_1.0_224.tflite`) is already tracked in `tests/test_models/models/` on `main`, so no missing-fixture risk. All new functions carry doxygen comments per repo convention.

## Docs / API
No architecture or public-API change (the header edit is a doxygen `@note` addition only, no signature change), so no documentation update is required beyond what's in the PR body.

## Commit hygiene
Both commits follow `[Component] Summary`, carry `Signed-off-by`, and the second commit's body explains root cause and fix rationale clearly.

Verdict: mergeable after nits

myungjoo:

Thanks, all three points addressed; head is now `f6e94789` (fixup folded into the first commit, tests unchanged).

1. **Minor, CI gate** — no code change. The three Tizen GBS builds are required status checks on `main` and run `%check` with `unit_test 1`, which executes `unittest_filter_shared_model`; pdebuild runs it too (not required). I will confirm from the GBS log of this head that the six `nnstreamerFilterSharedModel` cases actually ran. The Ubuntu meson job keeping tflite out of `meson test` is a deliberate pre-existing choice (#4903 context) and out of scope here.
2. **Nit, duplicate log text** — applied: the pre-check now logs `The model has unmatched tensors info, the shared model is not reloaded`. One correction for the record: the two messages cannot both fire for one reload — when the pre-check fails, `reloadModel()` jumps to `error` before `nnstreamer_filter_shared_model_replace()` is called, so the per-core check never runs; the old text was only ambiguous between the shared and non-shared branch.
3. **Nit, lock note** — applied: `@note The caller should hold the lock of the current interpreter; the new one is only read.`

Rebuilt with `-Dwerror=true` (jammy, tensorflow2-lite 2.15): `unittest_filter_shared_model` 6/6 pass.

myungjoo:

Review round 2, on head `f6e94789`. Independently re-derived the root-cause/fix analysis rather than trusting round 1's, and checked whether round 1's three points were actually resolved.

## Round-1 follow-up verification
- **Duplicate log text (nit)**: confirmed fixed. The pre-check message is now distinct (`"...the shared model is not reloaded"`), and re-tracing `TFLiteCore::reloadModel()` confirms the author's correction was right: when the pre-check in the shared branch fails, control jumps straight to `error:` before `nnstreamer_filter_shared_model_replace()` runs, so the per-core message inside `reloadInterpreter()` can never fire for the same reload. Not a bug either way, just confirming the record.
- **Lock-note nit**: confirmed fixed, `@note` on `isCompatibleInterpreter()` now states the new interpreter needs no lock.
- **CI-gate point**: independently checked `gh api repos/nnstreamer/nnstreamer/branches/main/protection` — `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` **is** a required status check on `main`, so the author's claim holds structurally. However, as of this review that check (along with all others) is still **pending** on `f6e94789` — the author's promised confirmation ("I will confirm from the GBS log... that the six cases actually ran") hasn't been posted yet. Not a code defect, just flagging that the promised verification is still outstanding.

## New findings (round 2)

**Nit** — `ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc:1259-1267` (`isCompatibleInterpreter`): the function ignores its implicit "old" side being a parameter and instead reads the live member `interpreter` directly, rather than taking an explicit `old_interpreter` argument symmetric with `new_interpreter`. It's correct at both of today's two call sites — inside `reloadInterpreter()` the local `old_interpreter` was captured as `interpreter` moments earlier under the same lock, and in `reloadModel()`'s new pre-check `interpreter` is the object actually being compared against — so this is not a live bug. But it is a latent trap: a future refactor of `reloadInterpreter()` that calls `isCompatibleInterpreter()` with some interpreter other than the current `interpreter` member (e.g. while restructuring the locking) would silently compare against the wrong object with no compiler warning, since the function's signature gives no hint that it's coupled to `this->interpreter`. Consider either naming it to make that coupling explicit (e.g. `isCompatibleWithCurrent`) or accepting the "old" interpreter as an explicit parameter for symmetry with `new_interpreter`.

**Confirmed, no new issue** — traced the TOCTOU window between the new pre-check (reads `this->interpreter` under only `interpreter->lock()`) and the actual `nnstreamer_filter_shared_model_replace()` call (which takes `G_LOCK(shared_model_table)` afterward). A second thread could complete its own accept-and-replace on the same shared key in that window. Re-verified the transitivity argument from round 1 holds: `nnstreamer_filter_shared_model_replace()` only ever installs an interpreter that has already passed an equality check against the prior shared interpreter, so tensor-info equality is preserved across any number of interleaved successful replacements, and the redundant per-core check inside `reloadInterpreter()` cannot diverge from the outer pre-check's verdict. No fix needed, just wanted an independent confirmation given this is the crux of the PR's correctness.

**Verified, not a defect** — `clang-format -output-replacements-xml` on the two touched C++ files reports zero replacements (clang-format 16.0.6), matching the PR body's formatting claim. `gst-indent`-relevant changes are comment-only in the two C files, so no formatting risk there either.

## Scope, tests, docs
No change from round 1's assessment: scope stays tight to the one sub-plugin, the 3 new gtest cases exercise the exact code paths this fix touches (`gst_tensors_info_is_equal()` and `nnstreamer_filter_shared_model_replace()`'s unconditional free), and would catch a regression if either of those shared helpers changed behavior in the future. No API/architecture change, so no doc update is required beyond the doxygen `@note` already added.

Verdict: mergeable after nits

myungjoo:

Round 2 nit applied; head is now `10f630b2` (rebased onto current `main` `cf830452`, which only touched `tools/debugging/`).

- **Nit, implicit `interpreter` member** — the comparison moved from `TFLiteCore::isCompatibleInterpreter()` to `TFLiteInterpreter::hasSameTensorsInfo (TFLiteInterpreter *other)`, so both sides are explicit at each call site: `old_interpreter->hasSameTensorsInfo (new_interpreter)` in `reloadInterpreter()` and `interpreter->hasSameTensorsInfo (interpreter_sub)` in the shared pre-check. Its brief states the lock rule (caller holds *this* interpreter's lock). The `TFLiteCore` declaration is gone, so the diff in the tflite file shrank to +25/-6. Commit message updated accordingly; tests unchanged.

Local re-run (jammy, tensorflow2-lite 2.15, `-Dwerror=true`): `unittest_filter_shared_model` 6/6, `nnstreamer_filter_reload` SSAT 2/2; mutation `hasSameTensorsInfo()` → always TRUE fails all three new cases.

myungjoo:

Review round 3, on head `10f630b2` (rebased onto current `main` `cf830452`). Independently re-checked whether round 2's point was resolved correctly, and looked for anything the first two rounds missed. I did not re-raise anything already resolved.

## Round-2 follow-up verification

**Confirmed, resolved correctly.** The comparison moved from `TFLiteCore::isCompatibleInterpreter()` to `TFLiteInterpreter::hasSameTensorsInfo(TFLiteInterpreter *other)` (`ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc:216-221`), removing the implicit-`interpreter`-member coupling round 2 flagged. Traced both call sites:
- `TFLiteCore::reloadInterpreter()` (line ~1273): `old_interpreter->lock(); new_interpreter->lock();` then `old_interpreter->hasSameTensorsInfo(new_interpreter)` — both interpreters are locked, unchanged locking behavior from before the extraction.
- `TFLiteCore::reloadModel()`'s shared branch (line ~1346): only `interpreter->lock()/unlock()` around the call, `interpreter_sub` (the `other` argument) is never locked. This is safe, not a new race: `interpreter_sub` is a plain `new TFLiteInterpreter()` local to this call, fully populated by this same thread via `loadModel()/setInputTensorProp()/setOutputTensorProp()/cacheInOutTensorPtr()` immediately above, and not yet reachable by any other thread (it's only published via `nnstreamer_filter_shared_model_replace()` afterward, and only once this check passes). No lock-order or TOCTOU issue introduced by the migration.
- No const-correctness change/regression: the new method takes `TFLiteInterpreter *other` (non-const) and is itself non-const, consistent with every other method on this class (none of `TFLiteInterpreter`'s methods, including sibling one-liners like `setDelegate()`, use `const`), so this isn't a deviation introduced by the migration.
- clang-format 16.0.6 independently re-run against this exact head: 0 replacements on both touched C++ files, matching the PR body's claim.

**Minor, documentation nit** — `tensor_filter_tensorflow_lite.cc:216`: the migrated helper's doxygen shrank to one line: `check other has the same input and output tensor meta, the caller should hold this lock`. The round-2 response promised (and the intermediate version at `f6e94789` had) an explicit note that `other`/the new interpreter needs no lock and why. That clarification didn't carry over to the final `hasSameTensorsInfo()` comment — it only says "hold *this* lock," which is correct but doesn't tell a future caller that `other` is deliberately left unlocked because it's assumed private/unshared at every call site (as opposed to just an oversight). Since the class isn't currently used with mismatched-visibility interpreters anywhere else, this is not a live bug, but worth restoring a short `@note` (e.g. "`other` needs no lock: at every current call site it isn't visible to any other thread yet") so a future third call site doesn't skip locking `other` when it *is* shared.

## Other checks (no new issues)

- Fixture `tests/test_models/models/mobilenet_v2_1.0_224.tflite` used by the new `model_name3` is already present in the tree; `tests/meson.build` needed no change. Confirmed no missing-fixture or build-registration risk.
- `tests/nnstreamer_filter_reload` (SSAT) exists and is untouched by this PR — consistent with the PR body's claim that it still passes and wasn't a hole this change needed to plug.
- `nnstreamer_filter_shared_model_replace()` (`tensor_filter_common.c`) and `checkSharedInterpreter()` (this file) are unchanged by the diff; re-verified independently that every core sharing a key is assigned the exact same `TFLiteInterpreter*` at open, so the round-1 "one pre-check covers every core" argument still holds on this head.
- Scope, size, non-shared-branch coverage (`tfliteReloadUnmatched_n`), and error-path (`interpreter_sub` freed via `goto error`) are unchanged from rounds 1-2's assessment — still tight and correct.

## Informational (not a defect)

All CI checks on this head (`10f630b2`, just pushed after the round-2 rebase) show `pending` at review time, including `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` — the required check that actually exercises `unittest_filter_shared_model`. The author's round-1 promise to confirm from the GBS log that the six `nnstreamerFilterSharedModel` cases ran is therefore still open again post-rebase (this is the same situation round 2 flagged on the previous head, now recurring because of the rebase, not a new problem).

Verdict: mergeable after nits

myungjoo:

Round 3 nit applied; head is now `d291d37e` (comment-only fixup folded into the first commit, same base `cf830452`).

- **Minor nit, lock note on `hasSameTensorsInfo()`** — the brief now reads `check other has the same input and output tensor meta (hold this lock; other is only read)`. Kept as a one-line brief to match the neighbouring inline accessors in `TFLiteInterpreter`.
- **CI confirmation** — still open; I will confirm from the required GBS x86_64 log of this head that the six `nnstreamerFilterSharedModel` cases ran.

myungjoo:

Review round 4, on head `d291d37e` (comment-only fixup on top of `10f630b2`, same base `cf830452`). I re-derived the fix's correctness independently rather than trusting rounds 1-3, then checked whether round 3's point was actually resolved and looked for anything the first three rounds missed.

## Round-3 follow-up verification

**Confirmed, resolved correctly.** `TFLiteInterpreter::hasSameTensorsInfo()` (`ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc:216-221`) now reads:
```
/** @brief check other has the same input and output tensor meta (hold this lock; other is only read) */
```
This restores the "`other` needs no lock" clarification round 3 asked for, matching exactly what the round-3 response promised. No further action needed here.

## Independent re-verification (no new defects)

- **Root cause / fix**: re-traced `TFLiteCore::reloadModel()` (lines ~1315-1385) end to end. The shared branch now checks `interpreter->hasSameTensorsInfo(interpreter_sub)` under `interpreter->lock()` before calling `nnstreamer_filter_shared_model_replace()`, and `goto error` (which does `delete interpreter_sub; return -EINVAL;`) on mismatch, so the framework correctly reports failure and keeps the old model. This matches the PR's stated fix and solves item F1 of #4920.
- **Concurrent-reload soundness (the TOCTOU window rounds 1-2 already examined)**: re-verified independently. `nnstreamer_filter_shared_model_replace()` (`tensor_filter_common.c:3206-3238`) is fully serialized under `G_LOCK(shared_model_table)`, including the per-core `replace_callback` loop and the unconditional `free_callback`. A reload's outer pre-check in `reloadModel()` can run against a since-superseded `this->interpreter` snapshot, but the actual replacement is still validated per-core inside `reloadInterpreter()` against each core's *current* `interpreter` field at the moment the lock is held. Because `hasSameTensorsInfo()` is a structural-equality check (reflexive/symmetric/transitive), two interpreters independently found compatible with the same prior baseline are necessarily compatible with each other, so the inner check can't diverge from the outer one across interleaved successful replacements. Confirms rounds 1-2's conclusion; no bug.
- **Single caller**: `grep -rn "nnstreamer_filter_shared_model_replace"` across the tree still shows only the one call site added by this feature (`tensor_filter_tensorflow_lite.cc:1357`), so the doc-only header/`.c` changes carry zero regression risk to any other sub-plugin.
- **Fixture**: `tests/test_models/models/mobilenet_v2_1.0_224.tflite` (the new `model_name3`) is present in the tree; no `tests/meson.build` change was needed.
- **Docs**: no signature change to any public API (the header edit is a `@note`-only addition), and nothing under `Documentation/` references `nnstreamer_filter_shared_model_replace` or `shared-tensor-filter-key`, so no doc update is required or missing.
- Attempted dynamic verification via the provided WSL docker container (`nns-f1-4920` / `/work/f1_test.sh`) as a cross-check on top of static review, but WSL was unresponsive to even a plain `docker ps` in this session (known environment flakiness) and never returned output; falling back to the PR body's own documented local matrix (jammy, tensorflow2-lite 2.15, `-Dwerror=true`: 6/6 pass on this PR, 2 of 3 new cases fail on unmodified `main`, all 3 fail with the check forced to `TRUE`), which is consistent with the code-level trace above.

## Informational (not a defect, not blocking)

CI on `d291d37e` is still pending at review time, which is expected for a freshly pushed head. Separately, by `gh run list --branch fix/4920-tflite-shared-reload`, every prior "Build and unit test/ Tizen/GBS" run on this PR branch (the required check that actually executes `unittest_filter_shared_model`) was **cancelled** by a subsequent push before completing — none has finished successfully on any commit of this PR so far. So the round-1 promise ("I will confirm from the GBS log that the six `nnstreamerFilterSharedModel` cases ran"), repeated as still-open in rounds 2 and 3, remains unconfirmed by a completed CI log after 3 rounds — not because it was dropped, but because each push so far has superseded the in-flight run before it could finish. Worth confirming once this run (or the next one, if there is a next push) completes, since the required check is otherwise the only CI signal that actually exercises the new tests.

## Scope, regression risk, test coverage

Unchanged from rounds 1-3's assessment, and I found nothing to add: the functional change stays ~25 lines in one sub-plugin file, the non-shared branch's behavior and locking are untouched by the extraction, the 3 new gtest cases reproduce the bug precisely and are structured to fail fast (`EXPECT_STREQ` on `model` path) rather than risk hanging against a freed interpreter on a regression, and `tfliteReloadUnmatched_n` pins the non-shared branch so the `hasSameTensorsInfo()` extraction itself has coverage. No unrelated module is touched.

## Findings

No new Critical/Major/Minor issues found in this round. No new nits — I looked but didn't find anything concrete enough to be worth raising; inventing wording preferences would not genuinely improve the change at this point.

Verdict: mergeable as-is

myungjoo:

Closing the open CI point from rounds 1-4: on head `d291d37e` all 21 checks pass, and the required job **Tizen GBS build on Ubuntu (x86_64, `--define "unit_test 1"`)** (run 34948769251) built and ran `tests/unittest_filter_shared_model` in `%check`: `6 tests from nnstreamerFilterSharedModel`, all `OK` — including `tfliteSharedReloadUnmatched_n`, `tfliteSharedReloadAfterUnmatched` and `tfliteReloadUnmatched_n`. So a regression of F1 fails a required status check.

Round 4 was "mergeable as-is" with no findings; marking the PR ready for review.

myungjoo:

Both nits were real; head is now `19632908` (rebased onto `a60a9ed9`, two commits).

**NIT 1 — assertions that cannot fail.** Confirmed, and measuring it turned up the reason, which was not the models:

- The `path2` check is gone. You are right that `_gtfc_setprop_MODEL_FILES` only rewrites the `prop` of the element whose `model` was set, so filter2's path never moves.
- The result check could not fail because of **timing**, not because the two models agree. `TEST_DEFAULT_SLEEP_TIME` is 10 ms, so after resuming the pipeline the case only ever saw the buffer prerolled before the reload, which still carries the old model's result. Measured on `9.png`: with the old sleep both phases report argmax 410, and with a longer run the second phase reports 489 on **both** sinks. So the shared reload does reach filter2, and it is observable.
- On `orange.png` the two quantized mobilenets are genuinely indistinguishable: same argmax (951), same top value (244), and the same output-vector sum (256) — so no assertion on that image can tell the models apart. The reload cases now classify `9.png`, where they disagree (410/46 vs 489/140).
- The reload cases now wait for the buffers that follow the reload with the in-tree idiom `count_output` + `wait_pipeline_process_buffers` instead of a fixed sleep. `tfliteSharedReload` gets the same treatment, since it had the same blind spot, and it now asserts that the result actually changed.

Mutation evidence that the new assertions bite (jammy, tensorflow2-lite 2.15):

| tflite source | SharedReload | SharedReloadUnmatched_n | SharedReloadAfterUnmatched | ReloadUnmatched_n |
|---|---|---|---|---|
| this PR | pass | pass | pass | pass |
| main | pass | **fail** | **fail** | pass |
| `hasSameTensorsInfo()` forced TRUE | pass | **fail** | **fail** | **fail** |
| shared replace skipped (no propagation) | **fail** | pass | **fail** | pass |

The last row is the one your nit asked for: before this change that mutation passed every case.

**NIT 2 — check and replace are not atomic.** Confirmed reachable: `tflite_setInputDim()` → `TFLiteCore::setInputTensorDim()` → `TFLiteInterpreter::setInputTensorsInfo()` mutates the shared interpreter's info, so a renegotiation on another sharing filter can invalidate the pre-check before the callbacks run. Holding the interpreter lock across the helper is not an option — `reloadInterpreter()` takes the same lock and would deadlock — so the durable fix is the one you describe, and it changes the exported callback type.

Taken as you suggested: the helper's `@note` now says the check cannot be conclusive outside the lock and that closing the gap needs the callback to report the refusal, and the durable fix is filed as item **B16** on #4920 (with the trigger, why the lock cannot simply be held, and the API impact). The PR body has a "What this does not close" paragraph saying the same.

myungjoo:

Review round 5, on head `19632908` (rebased onto `main` `a60a9ed9`, two commits: fix `36365ef8` + tests `19632908`). This round exists to verify myungjoo-bot's two nits from the last review were resolved correctly, and to look for anything new.

## myungjoo-bot's nits: verification

**NIT 1 (assertions that could not fail): confirmed resolved correctly.**
- The vacuous `EXPECT_STREQ (old_model_path, path2)` on filter2's `model` property is gone from `tfliteSharedReloadAfterUnmatched` — confirmed by reading the full diff, no `path2` reference remains.
- The reload cases now classify `9.png` (`reload_data_name`) instead of `orange.png`, and wait for post-reload buffers via `count_output` + `wait_pipeline_process_buffers` (`_play_until_buffers()`, `tests/nnstreamer_filter_shared_model/unittest_filter_shared_model.cc:120-151`) instead of a fixed `g_usleep`.
- `_connect_sinks()` connects `_new_data_cb` before `count_output` on both sinks, matching `tests/unittest_util.h`'s documented contract ("Connect this after the handler that checks the buffer... a satisfied count implies the check for that buffer has already run"). Correct order.
- Dynamically verified in the provided docker container (`nns-f1-4920`, tensorflow2-lite, current head): all 6 cases pass, each in well under 1 second (`tfliteSharedReload` 889 ms, `tfliteSharedReloadUnmatched_n` 927 ms, `tfliteSharedReloadAfterUnmatched` 885 ms, `tfliteReloadUnmatched_n` 473 ms). No hang, no flake.
- Independently re-ran the "hasSameTensorsInfo() forced TRUE" mutation from the PR body's table (swapped in `/work/scratch/patch/tflite_mut_true.cc`, rebuilt with ninja). Result: `tfliteSharedReloadUnmatched_n` and `tfliteSharedReloadAfterUnmatched` fail at the `EXPECT_STREQ` on `model` as claimed, confirming the new assertions do bite. (See new finding below about `tfliteReloadUnmatched_n`'s behavior under this same mutation.) Restored the tree from `/work/scratch/patch/tflite_fixed.cc` / `unittest_fixed.cc` afterward and confirmed by `md5sum` that the restore is byte-identical to the fixed backups, then rebuilt and re-ran all 6 cases clean (PASSED) to leave the container in the same state it was found.

**NIT 2 (check-and-replace not atomic): confirmed resolved correctly.** The helper's `@note` in both `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h` and `gst/nnstreamer/tensor_filter/tensor_filter_common.c` now states the check cannot be conclusive outside `G_LOCK (shared_model_table)` and that closing it needs `replace_callback` to report the refusal (which would change the exported callback signature). This matches the PR body's new "What this does not close" paragraph, and the point is correctly filed as item **B16** on issue #4920 (verified independently by reading the issue body: B16 text matches this exact race, added 2026-09-16, references #4972). Documentation-only change, no new call sites, no regression risk to any other sub-plugin (still the only caller of `nnstreamer_filter_shared_model_replace()`).

## New finding (round 5)

**Minor** — `tests/nnstreamer_filter_shared_model/unittest_filter_shared_model.cc`, `tfliteReloadUnmatched_n` (around line ~435-450): unlike its two shared-key siblings (`tfliteSharedReloadUnmatched_n`, `tfliteSharedReloadAfterUnmatched`), this test does **not** gate its post-set `_play_until_buffers()` call behind `if (refused)`. The siblings deliberately guard the follow-up invoke because, per the PR's own comment, "An accepted reload leaves the cores with a freed interpreter, which may hang invoke" — i.e. if a regression is reintroduced and the model reload is wrongly accepted, the test must not drive an `invoke()` against a bad interpreter. `tfliteReloadUnmatched_n` skips that guard: after `EXPECT_STREQ (old_model_path, path)`, it unconditionally calls `_play_until_buffers()` regardless of whether the reload was actually refused.

I confirmed this is reachable, not theoretical: rebuilding with the "`hasSameTensorsInfo()` forced TRUE" mutation (the same one the PR's own table uses for `tfliteReloadUnmatched_n`, listed there as "fail") produces a **SIGSEGV** (`exit=139`) partway through `tfliteReloadUnmatched_n`, after the `EXPECT_STREQ` failure is recorded but before the test can report `FAILED` — the crash happens inside the unconditional `_play_until_buffers()` call, invoking the now-mismatched, incompatible interpreter (float mobilenet_v2 swapped into a quantized-model pipeline). The process exit code is still non-zero, so CI (`run_unittests_binaries.sh`, no timeout wrapper per the PR body) does still fail and block the regression from merging — this is not a false-negative/silent-pass risk. But it's an inconsistency with the fail-fast design the other two new cases use, and a crash is a noisier, harder-to-triage CI signal than a clean gtest assertion failure (especially since `tfliteReloadUnmatched_n` happens to run last in the suite here, so nothing downstream gets masked, but that's incidental to file order, not a structural guarantee).

Suggested fix: wrap the same block this test already borrowed the pattern for — move `_play_until_buffers()` and the final `EXPECT_EQ (res[0], old)` inside `if (g_strcmp0 (old_model_path, path) == 0) { ... }`, mirroring `tfliteSharedReloadUnmatched_n`.

## Scope, regression risk, docs — unchanged from rounds 1-4

No new files touched, no new module touched, no API/architecture change beyond the doxygen `@note` already reviewed. Commit messages (`36365ef8`, `19632908`) accurately describe the code: verified the `@note` text lives in the fix commit (not the test commit) and the three new test cases plus the `9.png`/wait-helper rework live in the test commit, matching each commit's own description. PR body's local-evidence table is consistent with what I reproduced independently in the docker container.

Verdict: mergeable after nits

myungjoo:

Round 5's finding was right and is applied; head is now `e492ea61` (test-only fixup folded into the second commit, same base `a60a9ed9`).

- **Minor, missing `if (refused)` guard in `tfliteReloadUnmatched_n`** — the case now captures the refusal from the `model` path like its two shared-key siblings and only replays the pipeline when the reload was refused. Re-ran the mutation matrix: with `hasSameTensorsInfo()` forced TRUE the binary now exits 1 with three clean case failures, where before it ended in SIGSEGV (exit 139) inside this case. The other rows are unchanged: this PR 6/6 pass, main fails the two shared-key cases, and the "shared replace skipped" mutation fails `tfliteSharedReload` and `tfliteSharedReloadAfterUnmatched`.

myungjoo:

Review round 6, on head `e492ea61` (base `a60a9ed9`, two commits: fix `36365ef8` + tests `e492ea61`).

## Round-5 follow-up verification

**Confirmed, resolved correctly.** `tfliteReloadUnmatched_n` (`tests/nnstreamer_filter_shared_model/unittest_filter_shared_model.cc`) now captures the refusal the same way its two shared-key siblings do and only calls `_play_until_buffers()` / checks `res[0]` inside `if (refused) { ... }`. Traced it end to end: `res[0] = 0;` right after recording `old`, `g_object_set (filter1, "model", unmatched_model_path, NULL)`, then `refused = (g_strcmp0 (old_model_path, path) == 0)`, and the replay + `EXPECT_EQ (res[0], old)` only run when `refused` is true — matching `tfliteSharedReloadUnmatched_n`'s pattern exactly. This closes the SIGSEGV-under-mutation gap round 5 found; a reintroduced bug now fails cleanly at `EXPECT_STREQ` instead of crashing inside an unconditional replay.

## Independent re-check (no new issues)

- Re-verified the core fix (`TFLiteCore::reloadModel()`, `ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc:1315-1367`): the shared branch checks `interpreter->hasSameTensorsInfo(interpreter_sub)` under `interpreter->lock()`/`unlock()` before calling `nnstreamer_filter_shared_model_replace()`, and `goto error` on mismatch frees only `interpreter_sub` (nothing else holds it yet) and returns `-EINVAL`. This is unchanged since round 4 and still correct.
- Re-confirmed TF-Lite is still the only sub-plugin that calls `nnstreamer_filter_shared_model_replace()` / uses `shared-tensor-filter-key` (`grep` across `ext/` and `gst/nnstreamer/tensor_filter/`), so the doc-only header/`.c` changes carry no regression risk to any other filter, including litert.
- Re-confirmed issue #4920's item **B16** (filed in round 5) accurately records the remaining TOCTOU gap between the pre-check and the lock the helper takes, with the correct trigger (`tflite_setInputDim()` → `setInputTensorsInfo()` racing a reload) and why the fix requires changing the exported callback signature. Nothing further to add here; it's appropriately deferred rather than blocking this PR.
- Confirmed `clang-format -output-replacements-xml` (16.0.6) reports 0 replacements on the touched `.cc`/header. (I also ran it against the touched `.c` file out of caution and it reported many replacements, but that's the wrong tool for a `.c` file per this repo's convention — C files are checked with `gst-indent`, not clang-format — and the PR body's own claim is that `gst-indent`/`doxygen-tag.sh` output is unchanged from `main` on the touched files; GNU indent isn't installed in this environment so I could not re-run that tool myself, but I found no reason to doubt the claim given the change there is a comment-only `@note` addition.)
- Scope stays tight: the functional change is ~25 lines in one file, doc-only additions elsewhere, and the test commit only touches the one existing test file. No unrelated module touched.
- No architecture or public API signature change; the header edit is a doxygen `@note` addition, consistent with rounds 1-5's assessment that no further documentation is required.

## Findings

No new Critical/Major/Minor/Nit issues found in this round. Five prior rounds plus one bot review already drove this to a well-tested, tightly-scoped, correctly-documented fix, and round 5's only remaining finding is verified fixed above. Pending CI on the freshly-pushed head is expected and not a finding.

Verdict: mergeable as-is

myungjoo:

Round 6 is "mergeable as-is" with no findings, and CI on `e492ea61` is **21/21 green**.

The required **Tizen GBS x86_64 (`unit_test 1`)** job (run 35057038229) ran the reworked binary: `6 tests from nnstreamerFilterSharedModel`, all `OK`, including the three new cases and `tfliteSharedReload`, which now waits for post-reload buffers on `9.png`. `Ubuntu pdebuild on jammy` is green as well, and it runs the same binary.

Marking the PR ready for review.

PR #4973, opened 2026-09-15
[DataRepo] Bound datareposrc by its JSON and data file, drop stale JSON arrays

myungjoo:
- **What:** `datareposrc` trusted the numbers in its JSON. It could read memory past what it allocated, allocate without a bound (the allocation abort kills the process), and read JSON arrays after their parser was freed (a SIGSEGV). `datareposink` parsed a flexible header out of memories shorter than the header. These are items **E6** and **E7** of #4920, and item **F2 of #4960** (the core security audit). Note that #4960 and #4920 both number their items `F<n>`: the **F2** fixed here is #4960's `datareposink` item, *not* #4920's F2, which is a TF-Lite interpreter defect and is not touched by this PR. Every bare item id below belongs to #4960 unless it says otherwise.
- **How:** The flexible read path now checks every JSON value before it is used. The sample index must be inside `sample_offset` and `tensor_count`. `tensor_count` must stay inside `tensor_size` (checked before it is narrowed to 32 bits), must not decrease, and may give at most 256 tensors. Each `tensor_size` must be at least the 128-byte header and must fit in what is left of the file. A static or media sample larger than the file is refused when the file is opened. Every malformed value, negative ones included, is refused with a `STREAM/FORMAT` error naming the sample. The JSON arrays are cleared together with their parser. The `json` setter stops when the state refuses the change. The sink uses the size-checked `gst_tensor_meta_info_parse_memory ()`.
- **Result:** Production code +65/−10 in `gst/datarepo/` only. 23 new datareposrc cases and 2 new datareposink cases, all passing (45/45 and 14/14). On `main`, every new negative case fails: two by crashing (abort, SIGSEGV), the rest by assertion. A mutation matrix (one row per guard) is killed except one equivalent mutant. `unittest_trainer`, which also reads datarepo, passes. Pipelines that were already broken now error out: an offset past the file used to end with EOS, and a file shorter than one sample used to push a partial buffer.

## Why each change

| Trigger in the JSON / pipeline | `main` | now |
|---|---|---|
| `tensor_count` decreasing (`[0,4,2,6]`) | `num_tensors` wraps to ~4G; out-of-range indices read as 0 with a json-glib critical; 0-byte memories get an 88-byte header parse | refused |
| `tensor_count` past `tensor_size` or beyond 32 bits, or index past `sample_offset` / `tensor_count` | json-glib critical, value 0, wrong offset or an empty buffer pushed | refused |
| `tensor_size` < 128 | header parsed from a short allocation | refused |
| `tensor_size` = 10^15 | allocator abort (process exits 133) | refused |
| tensor or offset past the end of the data file | short read pushed (uninitialised tail) or EOS | refused |
| `sample_size` (JSON or caps) > file | full-size allocation, partial buffer | refused at start |
| `json=a.json`, then `json=broken.json`, then flexible `caps` | arrays point into a.json's freed tree → **SIGSEGV** | nothing to read, refused |
| `json` set in PAUSED/PLAYING | path change refused, but the old file re-parsed under the streaming thread | setter stops |
| datareposink: flexible memory < 128 B | header read past the mapping | refused |
| a negative `sample_offset` / `tensor_count` / `tensor_size` | refused, but only as a generic stream failure | refused as `STREAM/FORMAT` naming the sample |

`file_size` goes from `gint` to `guint64`. It is now compared with 64-bit offsets, and a data set over 2 GiB would otherwise be refused (or wrap).

## Commits
1. `[DataRepo] Bound what datareposrc reads by the JSON arrays and the data file` (E6 of #4920, F2 of #4960)
2. `[DataRepo] Do not keep the JSON arrays of a released parser` (E7 of #4920)
3. `[Test/DataRepo] Read crafted repositories that break the JSON contract`

Each commit builds on its own. Commit 1 alone passes the existing 22 + 12 datarepo cases.

---

<details><summary>Context for agents / reviewers (verification detail)</summary>

**Test design.** The cases write a tiny flexible repository byte by byte: `_write_flexible_repo`, uint8 `4:1:1:1`, 132 B per tensor. Each case writes its own JSON. `_run_repo_pipeline` records the first EOS/ERROR with its source name, `GError` domain and code. It also counts messages on the json-glib log domain `"Json"`. `_expect_refused` asserts `src0` + `GST_STREAM_ERROR_FORMAT` + 0 buffers + 0 json criticals. That separates a refusal from main's generic `STREAM_ERROR_FAILED`, from a crash, and from a silent 0 returned by an out-of-range index. The sink case wraps a valid 132-byte tensor with `gst_memory_new_wrapped (..., maxsize 132, size 8)`: the bytes behind the mapping are a good header, so main writes 8 bytes and returns EOS.

**Negative control** (new tests on main's `gst/datarepo`, one process per case). Every `_n` case fails. `readCraftedFlexibleHugeSize_n` exits 133 (allocator abort) and `readFlexibleAfterFailedJsonReload_n` exits 139 (E7 UAF). The positive cases (`readCraftedFlexibleTensors`, `...MaxTensors`, `readFlexibleAfterJsonReload`, `readOctetSampleOfFileSize`, `writeFlexibleTensorMemory`) pass on both.

**Mutation matrix.** Each guard was dropped on its own, or moved one short. Killed: sample_offset index check, tensor_count index check (by the json-glib log count), the `tensor_size`-length bound on the sample's own count and on the next count (both also catch values beyond 32 bits), the 256 limit and `>=` 256, `num_tensors == 0`, `< hsize` (mutated to `< hsize/16`), remaining-bytes check, `>=` remaining (kills the positive exact-end case), remaining → whole-file (kills `SizePastEnd_n`), `sample_size > file_size` and its `>=`, clearing the arrays, the setter break, sink `parse_memory` → `parse_header`. **Equivalent:** `next < cur`. The unsigned subtraction then wraps past `NNS_TENSOR_SIZE_LIMIT`, and the same refusal follows. It is kept because it states the defect the issue names. A separate "offset > file_size" check was dropped during development: it was equivalent to the `fd_offset > file_size` term of the per-tensor check.

**Not done here (out of scope, tracked in #4960; these ids are #4960's):** F1 (`location` as a printf format), F3 (short reads of static/media samples), F4 (fd 0 sentinel), F5/F6/F7. The header-declared data size is not compared with `tensor_size`: consumers already validate it (tensor_filter B6, decoder C24 and transform C6 of #4920), and a self-describing tensor may carry slack.

**Local verification** (WSL jammy): `-Werror -Wextra -Wsign-compare` compile of every datarepo object. clang-format 16 on the `.cc` files. `doxygen-tag.sh` with ctags, plus a negative control (one `@brief` removed → rc=1). `newline.sh`. `indent.sh` shows no hunk in changed lines. valgrind is not usable on this box, so the memcheck gate is CI-only.
</details>


myungjoo:

Scope reviewed: the 5 changed files (`gst/datarepo/gstdatareposink.c`, `gst/datarepo/gstdatareposrc.c`, `gst/datarepo/gstdatareposrc.h`, `tests/nnstreamer_datarepo/unittest_datareposink.cc`, `tests/nnstreamer_datarepo/unittest_datareposrc.cc`), cross-checked against issue #4920 items E6/E7 and issue #4960 item F2, plus the surrounding datarepo code and the json-glib/GStreamer APIs it calls. I also built and ran the new tests against the PR branch (a pre-built tree in the reviewer's WSL environment): all 20 new/related `datareposrc` cases in the filter `*crafted*:*Crafted*:*Flexible*:*Json*:*json*` pass (`[ PASSED ] 20 tests.`), matching the PR description.

**Coverage of the stated topic (E6/E7/F2).** Traced each new guard by hand against the issue text:
- `gst_data_repo_src_get_num_tensors()` (`gstdatareposrc.c:747-750`) now rejects a decreasing `tensor_count`, a `next_idx_tensor_cnt` past `tensor_size_array_len`, and more than `NNS_TENSOR_SIZE_LIMIT` (256) tensors in one sample — directly closes the `num_tensors` wrap described in E6/F2. Because `cur_idx_tensor_cnt <= next_idx_tensor_cnt <= tensor_size_array_len` is enforced before the subtraction, the later `tensor_size_array[tensor_count + i]` access (`:838`) is provably in-bounds.
- The new per-tensor check at `:852-853` (`val < hsize || fd_offset > file_size || val > file_size - fd_offset`) bounds every `tensor_size` by both the flexible header size and the file's remaining bytes, and is evaluated with correct short-circuit order (the `fd_offset > file_size` guard runs before the subtraction, so `file_size - fd_offset` cannot underflow). This closes the "huge tensor_size aborts the allocator" and "tensor read past EOF" cases.
- `sample_offset_array`/`tensor_count_array` are now bounds-checked against `shuffled_index` before use (`:806-810`), closing the "missing array entry read as 0" case.
- E7 (UAF of JSON arrays after parser replacement): `gst_data_repo_src_read_json_file()` NULLs `sample_offset_array`/`tensor_size_array`/`tensor_count_array` immediately after the old parser is unreffed and before the new file is loaded (`:1476-1482`), and the `PROP_JSON` setter now stops when `gst_data_repo_src_set_file_path()` refuses the change (state != READY/NULL) instead of always re-parsing (`:1618-1620`). This matches the issue's suggested fix exactly.
- `gstdatareposink.c`: `gst_tensor_meta_info_parse_header()` → `gst_tensor_meta_info_parse_memory()` (`:349`) adds the missing `msize < hsize` check (verified in `nnstreamer_plugin_api_impl.c:1658-1661`) before reading the 128-byte header out of an upstream memory.
- `file_size` widened `gint → guint64` (header + `:1165`), with all comparison sites already using 64-bit-safe types; confirmed by grep that no other file in the tree touches `GstDataRepoSrc.file_size`, so this is a purely internal, non-ABI change.

**Regression risk.** Low. The struct-field type change is private to `gst/datarepo/` (no external references found). The new `sample_size > file_size` guard in `gst_data_repo_src_start()` (`:1185`) only fires for non-image, non-flexible data types where `sample_size` is computed from caps/JSON before `start()` runs; for flexible/sparse tensors `sample_size` stays 0 via `gst_data_repo_src_parse_caps()` (`:366-368`), so the guard cannot false-positive on the flexible path this PR targets. `unittest_trainer`, which also reads datarepo files, is reported passing and the field it could observe (`file_size`) is not part of any public header consumed outside this module.

**Size/focus.** Production diff is +53/−5, entirely inside `gst/datarepo/`. No other module is touched. This matches the stated topic tightly; nothing looks like scope creep.

**Test completeness.** 18 new `datareposrc` cases (positive + `_n` negative) and 2 new `datareposink` cases cover: decreasing tensor_count, count past sizes/missing, size below header, huge size, size/offset past EOF, missing json, failed json reload before/after a valid one, `json` set outside READY/NULL, static/octet sample larger than file, and a sink memory shorter than the flexible header. These are wired into the existing `tests/nnstreamer_datarepo/meson.build` targets, so `meson test -C build/` already runs them — no separate CI wiring was needed, and a future regression reintroducing any of these bugs will fail an existing gtest binary rather than silently passing.

**Documentation.** No public API, property, or architecture change; nothing in `Documentation/` references these internals, so no doc update is owed here.

**Findings**

- Minor / Nit — `gst/datarepo/gstdatareposrc.c:730-751, 826` (`gst_data_repo_src_get_num_tensors`, and the `tensor_count = (guint) val` cast at `:830`): `sample_offset`/`tensor_count` JSON values are `gint64` and only checked for `< 0` before being narrowed to `guint`. A crafted JSON value between `G_MAXUINT` and `G_MAXINT64` (e.g. `4294967300`) silently wraps mod 2^32 instead of being rejected. I confirmed this does **not** reopen a memory-safety hole — the wrapped value is still re-validated against the real `tensor_size_array_len`/`NNS_TENSOR_SIZE_LIMIT` bounds added by this PR before any array access — but it is a latent input-validation gap adjacent to the exact field this PR hardens, and a case exercising it (e.g. `tensor_count` value `4294967297`) isn't in the new negative-test set. Suggested fix: reject `val > G_MAXUINT` alongside the existing `val < 0` check, or fold it into a follow-up. Not blocking.

- Informational — the PR explicitly and correctly leaves F1 (printf-format `location`), F3 (short reads padding with uninitialised heap for non-first samples), F4 (fd-0 sentinel), and F5-F7 out of scope, with a clear rationale in the PR body. Confirmed these are indeed untouched by the diff and still tracked under #4960/#4920. No action needed in this PR.

Verdict: mergeable after nits

myungjoo:

Thanks for the review. Both points handled in-PR (force-pushed, head `898b908e`, rebased onto `cf830452`; commit trees verified identical to the fixed-up working tree):

1. **Minor: `tensor_count` values above `G_MAXUINT` truncated** -- valid, fixed. `gst_data_repo_src_get_num_tensors ()` now bounds both the sample's own count and the next count by `tensor_size_array_len` while they are still `gint64`, before narrowing to `guint`. That replaces the separate `next > tensor_size_array_len` term and the two `g_return_val_if_fail (val >= 0, 0)` criticals, so a negative count is a plain refusal too. New cases `readCraftedFlexibleCountBeyondUint_n` (sample's own count `4294967298`) and `readCraftedFlexibleNextCountBeyondUint_n` (next count). Both describe a *valid-looking* sample after truncation to 2, so the previous head accepted them with EOS. Mutation check: dropping the bound on the own count fails `...CountBeyondUint_n`; dropping it on the next count fails `...NextCountBeyondUint_n` and `...CountPastSizes_n`.
2. **Informational (F1/F3-F7 out of scope)** -- agreed, no change.

Also fixed from CI: `Spell Check with Typos` rejected the identifier `json_criticals`; renamed to `json_logs` (it counts warnings as well).

Local: 42/42 `unittest_datareposrc`, 14/14 `unittest_datareposink`, strict `-Werror -Wextra -Wsign-compare`, clang-format 16, doxygen-tag (with ctags, negative control rc=1), newline, indent.

myungjoo:

Scope: re-review of the current head (`898b908e`, 3 commits on main `cf830452`), covering the diff (`gst/datarepo/gstdatareposink.c`, `gstdatareposrc.c`, `gstdatareposrc.h`, `tests/nnstreamer_datarepo/unittest_datareposink.cc`, `unittest_datareposrc.cc`), the earlier relayed review and the author's response on this PR, the per-commit diffs (`git show` on each SHA), and the surrounding datarepo source plus the helper functions the new checks depend on (`gst_tensor_meta_info_get_header_size()` / `gst_tensor_meta_info_parse_memory()` in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c` / `nnstreamer_plugin_api_impl.c`). Issues #4920 (E6/E7) and #4960 (F2) read for scope. CI at review time: DCO, Spell Check, Static checks, and the Linux/macOS builds pass; Tizen GBS and the Valgrind jobs were still pending.

**Round-1 finding, verified fixed.** The only prior finding (a `tensor_count` value between `G_MAXUINT` and `G_MAXINT64` silently truncating on the cast to `guint`) is fixed in `gst_data_repo_src_get_num_tensors()` (`gstdatareposrc.c:730-751`): both the sample's own count and the next sample's count are now bounded against `tensor_size_array_len` while still `gint64`, before either is narrowed. I independently checked this was folded into commit 1 itself rather than left as a separate fixup on top: `git show 4229b1d1 -- gst/datarepo/gstdatareposrc.c` already contains the full two-sided bound, matching what the author's response describes. New tests `readCraftedFlexibleCountBeyondUint_n` / `readCraftedFlexibleNextCountBeyondUint_n` exercise it; I traced both by hand against the crafted JSON (`"0,4294967298,4,6"`) and confirmed they hit this new bound rather than some other guard.

**Independent verification beyond round 1:**
- Traced `gst_tensor_meta_info_get_header_size()` (`nnstreamer_plugin_api_util_impl.c:1542-1555`): for a freshly `gst_tensor_meta_info_init()`ed meta it always returns 128 for the V1 header regardless of type/format (static, flexible or sparse), so computing `hsize` once outside the per-tensor loop (`gstdatareposrc.c:840-841`) is correct for every code path this function reaches, including the sparse branch it shares.
- Traced the per-tensor bound at `gstdatareposrc.c:853-854` (`val < hsize || fd_offset > file_size || val > file_size - fd_offset`) against `fd_offset`'s cumulative update at `:900-901` using `readCraftedFlexibleSizePastEnd_n` by hand: after the first (132-byte) tensor is fully read, `fd_offset` advances to 924 of 1056, and the second tensor's declared size (133) fails `133 > file_size - fd_offset (132)`. The "remaining bytes" bookkeeping is correct across loop iterations, not just a single upfront check.
- Traced the datareposink fix (`gstdatareposink.c:349`, `gst_tensor_meta_info_parse_header` → `gst_tensor_meta_info_parse_memory`) into `nnstreamer_plugin_api_impl.c:1645-1672`: it compares `gst_memory_get_sizes()` (the memory's *size*, not *maxsize*) against `hsize` before mapping, which is exactly what makes `writeFlexibleTensorShortMemory_n` (an 8-byte `size` inside a 132-byte `maxsize` mapping, so the bytes behind the mapping are themselves a valid header) fail as intended rather than reading past the declared size.
- Verified the mechanism behind the E7 UAF fix: `gst_data_repo_src_read_json_file()` nulls both the array pointers *and* their `_len` counters together (`:1480-1485`). Because of that, a failed reload doesn't need its own NULL check in the read path — the bounds check added in commit 1 (`shuffled_index >= sample_offset_array_len || shuffled_index >= tensor_count_array_len`, `:806-811`) already refuses every sample once `_len` is 0. So the two commits compose correctly: commit 1's guard is what actually stops commit 2's dangling pointers from ever being dereferenced, even though neither commit message states that dependency explicitly. `readFlexibleAfterFailedJsonReload_n` exercises exactly this composition; `setJsonWhilePaused_n` separately confirms the different fix in the `PROP_JSON` setter — when `gst_data_repo_src_set_file_path()` refuses the change, the setter now breaks before calling `read_json_file()` at all, so the arrays from the still-valid previous JSON are left untouched (not nulled), and the run correctly still produces 4 buffers with EOS.
- Confirmed the `sample_size > file_size` guard in `gst_data_repo_src_start()` (`:1186-1187`) cannot false-positive on the flexible/sparse path: `gst_data_repo_src_parse_caps()` sets `sample_size = 0` for `GST_DATA_REPO_DATA_TENSOR` and only accumulates it over `config.info.num_tensors`, which is 0 for flexible/sparse caps (no `num_tensors` field in such caps). So this guard is inert for exactly the code paths this PR otherwise hardens, and only fires for static-tensor/video/audio/octet/text samples, matching what `readOctetSampleLargerThanFile_n` and `readTensorsNoJSONSampleLargerThanFile_n` test.

**Commit split and messages.** Three commits, one topic each, and each builds on the last cleanly (`git show --stat` per SHA: commit 1 touches only `gstdatareposrc.{c,h}` + `gstdatareposink.c`; commit 2 touches only `gstdatareposrc.c`; commit 3 touches only the two test files). The messages accurately describe what each commit does; I did not find a claim in any commit message that the diff doesn't support.

**Regression risk.** Still low and unchanged from round 1: the diff stays inside `gst/datarepo/`, the `file_size` widening (`gint`→`guint64`) has no external references (no other file touches `GstDataRepoSrc.file_size`), and every new check is a refusal (fail-closed) rather than a behavior change to any currently-working pipeline, other than the two already-broken cases the PR body documents (offset past EOF, short final sample) which now error instead of silently misbehaving.

**Test completeness / future-regression detection.** Unchanged from round 1's assessment: the new cases are ordinary `TEST()`s inside the two already-registered gtest binaries, so `meson test -C build/` picks them up with no separate CI wiring, and a future regression reintroducing any of the guards this PR adds would fail an existing CI job rather than passing silently.

**Documentation.** No architecture or public API change; nothing in `Documentation/` needs updating for this PR.

**Findings:** none blocking or major. One very minor, independently-found nit that is not worth a revision on its own:
- Nit — `gst/datarepo/gstdatareposrc.c:730` and `:825`: `tensor_count_array[shuffled_index]` is read and range-checked twice (once inside `gst_data_repo_src_get_num_tensors()` as `cur_idx_tensor_cnt`, again at the call site as `tensor_count`). Both reads are deterministic and always agree, so this is not a correctness issue — just a small amount of avoidable duplicate JSON-array access that a future cleanup could fold into one read.

Verdict: mergeable as is

myungjoo:

Thanks for the re-review.

- **Nit (double read of `tensor_count[shuffled_index]`)**: left as is, deliberately. Both reads predate this PR (`tensor_count` at the call site, `cur_idx_tensor_cnt` inside `gst_data_repo_src_get_num_tensors ()`). This PR adds no third read, and the call-site value is only used after the helper has accepted the same element under the new bound. Folding them means changing the helper's interface (an out-parameter, or returning the start index), which is refactoring outside a minimal memory-safety fix.
- **Composition of commit 1 and commit 2**: correct observation. The cleared `_len` counters are what make the read path refuse after a failed reload. Commit 2's message says "a failed load leaves nothing to read and the read path refuses the sample". The dependency on commit 1's index check is intended, which is why commit 2 sits on top of it.

No code change in this round.

myungjoo:

Both NITs are fixed, head `87b90f01` (rebased onto `a60a9ed9`; the three commit trees are byte-identical to the verified working tree).

**NIT 1 -- negative values reported as a generic failure.** Valid, fixed. The three `val < 0` branches for `sample_offset` (`:815`), `tensor_count` (`:826`) and `tensor_size` (`:847`) now post `GST_ELEMENT_ERROR (src, STREAM, FORMAT, ...)` with the sample index and the offending value, so every malformed JSON value in this path reports the same way. Three cases added as you suggested: `readCraftedFlexibleNegativeOffset_n`, `readCraftedFlexibleNegativeCount_n`, `readCraftedFlexibleNegativeSize_n`. They discriminate: against the previous head's element sources they fail with `GST_STREAM_ERROR_FAILED` (1) where the fixed element reports `GST_STREAM_ERROR_FORMAT` (11). `unittest_datareposrc` is now 45 cases, all passing; `unittest_datareposink` 14.

Note the negative `tensor_count` is refused at the read site, which runs before `gst_data_repo_src_get_num_tensors ()`; the helper's own `< 0` refusal (added for the 32-bit bound) stays as the guard for the *next* sample's count.

**NIT 2 -- unwrapped commit message lines.** Fixed. The two paragraphs I edited in commits 1 and 3 were re-wrapped; no body line now exceeds 76 columns (`git log` checked per commit). Rewording used `git commit-tree` on the existing trees, so the diff is unchanged -- `git diff` between the pre-reword and post-reword heads is empty, and each commit's tree hash is unchanged.

Local re-verification on the new head: 45/45 + 14/14, strict `-Werror -Wextra -Wsign-compare` build of every datarepo object, clang-format 16 clean, doxygen-tag (with ctags; negative control still rc=1), newline and indent clean.

myungjoo:

Scope: round-3 re-review of head `87b90f01` (3 commits on main `a60a9ed9`). I read both prior relayed reviews, the myungjoo-bot formal review, and all of the author's responses on this PR, then re-checked the diff (`gh pr diff 4973`), each commit individually (`git show <sha>`), and the surrounding `gst/datarepo/` source in the local checkout. Per the round-3 brief, I focused on the two round-2 nits and did not re-litigate points already verified fixed in round 2 (the `G_MAXUINT` truncation bound, the E7 array-clearing/setter fix, the datareposink `parse_memory` fix, the `sample_size > file_size` guard) — I re-checked those only for regressions, not from scratch.

**NIT 1 (round 2): negative `sample_offset`/`tensor_count`/`tensor_size` reported as a generic stream failure instead of `STREAM/FORMAT`.**
Verified fixed, correctly and completely, in `gst/datarepo/gstdatareposrc.c`:
- `sample_offset` negative check (~line 815) now does `GST_ELEMENT_ERROR (src, STREAM, FORMAT, ...)` naming the sample and the offending value.
- `tensor_count` negative check (~line 828) likewise.
- `tensor_size` negative check inside the per-tensor loop (~line 851) likewise, and it correctly `goto error` (not a bare `return`) since `buf` has already been allocated at that point — I traced the `error:` label and confirmed `gst_buffer_unref (buf)` runs and `mem` is still NULL there, so there is no leak and no double-free on this new path.
- The two earlier checks (`sample_offset`/`tensor_count`) still `return GST_FLOW_ERROR` directly rather than `goto error`, which is correct since `buf` is not allocated yet at those points — the two error-handling styles are consistent with the existing code, not a bug.

All three are pinned by new, well-isolated tests: `readCraftedFlexibleNegativeOffset_n`, `readCraftedFlexibleNegativeCount_n`, `readCraftedFlexibleNegativeSize_n`. I checked how they isolate the specific branch under test: `_read_crafted_repo(start, stop, run)` sets `start-sample-index`/`stop-sample-index` to the same value, and `gst_data_repo_src_start()` sets `current_sample_index = start_sample_index` (`gstdatareposrc.c:1148`), so each test reads exactly one target sample directly, bypassing the sequential-read path that would otherwise trip an unrelated earlier guard (e.g. `gst_data_repo_src_get_num_tensors()`'s own bound) before reaching the branch the test means to exercise. `_expect_refused()` additionally asserts `error_domain == GST_STREAM_ERROR`, `error_code == GST_STREAM_ERROR_FORMAT`, `buffers == 0`, and `json_logs == 0`, which is exactly what discriminates "our new refusal" from both main's generic `STREAM_ERROR_FAILED` and a silent json-glib-critical-returns-0 path. This is real test pinning, not just smoke coverage: reverting any one of the three `GST_ELEMENT_ERROR` conversions back to the old `GST_ERROR_OBJECT` would flip the corresponding test's `error_code` assertion and fail it.

I also swept the rest of `gstdatareposrc.c` for any other JSON-value check still using a generic `GST_ERROR_OBJECT`/`g_return_val_if_fail` in this same flexible-read path and found none — the three the prior review flagged were the only ones in scope.

**NIT 2 (round 2): commit message body lines over ~76 columns in `4229b1d`/`898b908`.**
Verified fixed. Checked `git log -1 --format=%B` for all three current commits (`a80c4e20`, `57444906`, `87b90f01`); the longest line in any commit message body is 76 columns (the title line of commit 1), matching the author's claim. `git diff` between the pre/post-reword trees was not something I could re-verify directly (no reflog for the old SHAs in this checkout), but the current trees match the diff and CI history the author cites, so I have no reason to doubt the "trees unchanged" claim.

**Regression check on this round's delta.** The only behavior-affecting change since round 2 is promoting three error sites from `GST_ERROR_OBJECT` (which does not stop the element, only logs) to `GST_ELEMENT_ERROR (..., STREAM, FORMAT, ...)` (which posts a bus error); all three sites already `return GST_FLOW_ERROR` / `goto error` on the same condition before and after this change, so no previously-accepted (valid) JSON becomes newly refused — only the diagnostic surfaced to the application improves. I found no other functional edits in this round's diff.

**CI.** At the time of this review, checks on `87b90f01` were still running (Static checks, GBS, Valgrind, pdebuild jobs pending); this is expected since the reword-only history rewrite produced new commit SHAs and re-triggered CI from scratch. The content is identical to the `898b908e` head that already passed all 21 checks including both Valgrind jobs and Tizen GBS `unit_test 1`, so I don't expect a different result, but it is worth confirming green before merge.

**Findings:** none blocking or major. No new nits from this round's changes.

Verdict: mergeable as is (pending the in-flight CI run on `87b90f01` turning green, which is expected given the identical content already passed on the prior head).

PR #4976, opened 2026-09-15
[Element/Repo] Keep tensor repo slots alive while in use and add them atomically (#4920 C25, C26)

myungjoo:
- **What:** fixes two memory bugs in the process-wide slot table (`gsttensor_repo.c`) that `tensor_reposink` and `tensor_reposrc` share, items **C26** and **C25** of #4920. C26: destroying a `tensor_reposrc` freed its slot while a `tensor_reposink` or another `tensor_reposrc` was still blocked waiting on that slot's lock and cond. That is a use-after-free, and in practice the streaming thread hangs for good. C25: two elements setting the same `slot-index` concurrently both inserted the slot, the table freed it twice, and `ml_logf` aborted.
- **How:** the slot data is now reference-counted. Each accessor holds a reference while it uses the slot. Removal takes the slot out of the table, marks it EOS and wakes every waiter (`set_buffer` returns FALSE, so the sink posts RESOURCE/WRITE; `get_buffer` returns EOS), and memory is freed with the last reference. Add does its lookup and insert under one lock hold, and repo init checks its flag under the lock. The change is confined to `gsttensor_repo.c/.h`; no element code is touched.
- **Result:** new `unittest_repo` (15 cases). Without the fix, 5 fail: 3 time out, 1 SIGSEGV, and 1 aborts on the duplicate key. With the fix, 15/15 pass (also 15 repeats with 40 busy CPU loops), and SSAT `nnstreamer_repo` 77, `_dynamicity` 10, `_lstm` 2 and `_rnn` 2 pass, the same as on main. Valgrind on the new binary: no error from repo code; main shows invalid reads in `gst_tensor_repo_get_buffer`.

### Commits
1. `[Element/Repo] Keep a removed slot alive while an element still uses it` (C26)
2. `[Element/Repo] Look a slot up and insert it under one lock hold` (C25)
3. `[Test/Repo] Add unit tests for the tensor repo slot lifetime`

### Behaviour change worth knowing
When a slot is removed while an element still uses it, a sink blocked on that slot now gets an error and a blocked src gets EOS. Before, the element hung on freed memory. Nothing changes while a slot exists, and the existing repo SSAT groups give identical results.

---
<details><summary>Context for agents (evidence, mutation matrix, scope decisions)</summary>

- Audited defects: #4920 C25 (`gsttensor_repo.c:172-185` at 4e2522f7) and C26 (`:48-68` + `remove_repodata`, called from `gst_tensor_reposrc_dispose`). Both were re-verified on main `cf830452` before starting.
- `gst_tensor_repo_get_repodata()` was dropped from `gsttensor_repo.h`. The header is not installed (only `gst/nnstreamer/elements/meson.build` compiles the file), and nothing outside `gsttensor_repo.c` called it.
- `get_buffer` no longer calls `check_changed()`/`check_eos()`. Those looked the slot up again by index and took the repo lock inside the slot lock, the reverse of removal's order. Both functions stay public, now take a reference, and have no in-tree callers.
- `add_repodata` now logs a `g_return_val_if_fail` critical and returns FALSE when the repo is not initialised. Before, the same critical was followed by an insert into a NULL table and an abort. Every in-tree pipeline creates a reposink first.
- Negative control: the unfixed `gsttensor_repo.c` built with the new test file, cases run one at a time with `timeout 30`. `removeWakesBlockedSink` rc=1 (timeout), `removeWakesBlockedSrc` rc=139, `removeWakesAllWaiters` rc=1, `addSameSlotConcurrently` rc=133 ("The key[5000] is duplicated"), `pipelineSinkErrorsWhenSlotRemoved` rc=1. The other 10 cases pass on both trees.
- Per-commit: commit 1 alone builds with `-Dwerror=true`, is gst-indent clean, passes all 4 repo SSAT groups, and fails only `addSameSlotConcurrently` (C25).
- Mutation matrix on the fixed file (a mutation is caught when a test fails):
  - broadcast→signal on remove: caught by `removeWakesAllWaiters`.
  - no `eos = TRUE` on remove: caught by 3 wake tests plus the pipeline test.
  - no refcount increment in lookup: caught by 5 cases (SIGSEGV or hang).
  - lookup/insert split again: caught by `addSameSlotConcurrently`.
  - `g_hash_table_steal` → `remove` (double unref): caught by `removeWakesBlockedSrc` (SIGSEGV).
  - dropping the table's final unref on remove: no assertion catches it, but the CI valgrind gate does. `g_hash_table_steal()` clears the table's slot, so the block allocated in `gst_tensor_repo_add_repodata()` becomes definitely lost with a frame of ours, and `addSameSlotConcurrently` would leak 2000 of them.
- Checks run: gst-indent (C), clang-format 16.0.6 (test), `doxygen-tag.sh` with ctags (negative control flagged a removed `@brief`), strict `-Wextra -Wsign-compare -Werror` recompiles of both objects, newline at EOF.
- The unit tests' timing sleeps (200 ms before remove) only decide whether the old bug is exercised; the fixed code passes either way, because a call on a removed slot also fails. `pipelineSinkErrorsWhenSlotRemoved` is the exception and does not rely on timing: it waits for the sink's second buffer through a pad probe, and fails if `set_buffer()` logs its missing-slot critical, which is what a too-early removal would produce.
- Pre-existing and left out of scope: a `tensor_reposrc` is still destroyed while a sink uses its slot, so its dispose removes the slot; that is the element's design, and this PR makes it safe rather than changing it. A reposrc whose slot was removed and that is *not* blocked loops on `get_buffer` returning NULL without EOS (logs criticals); same on main.
</details>


myungjoo:

Scope reviewed: the 3 commits on this PR (9dbd9ff7, 483c9a1e, b68245c7) touching `gst/nnstreamer/elements/gsttensor_repo.c`/`.h` and the new `tests/nnstreamer_repo/unittest_repo.cc` + `tests/meson.build` hunk. I read the full pre- and post-patch `gsttensor_repo.c/.h`, the (unchanged) callers in `gsttensor_reposink.c`/`gsttensor_reposrc.c`, the build wiring (`tests/meson.build`, `gst/nnstreamer/meson.build`, `packaging/nnstreamer.spec`), and cross-checked against issue #4920 items C25/C26. I did not need a local rebuild to form these findings (static/code-path analysis); the PR's own CI run was already green on the build/Valgrind(arm)/static-check jobs at review time.

Yes, for the two documented bugs:
- **C26** (UAF on slot removal while a waiter is blocked): `gst_tensor_repo_remove_repodata()` now steals the slot out of the hash table under `_repo.repo_lock`, then marks `eos = TRUE` and broadcasts both conds under `data->lock`, and only frees on the last `gst_tensor_repo_unref_repodata()`. Every accessor (`set_buffer`, `get_buffer`, `set_eos`, `set_changed`, `check_eos`, `check_changed`) now goes through `gst_tensor_repo_ref_repodata()` and unrefs before returning, so the struct can't be freed out from under a blocked waiter. `gst_tensor_repo_get_buffer()` inlining `data->src_changed`/`data->eos` instead of re-calling `check_changed()`/`check_eos()` also removes a real lock-order hazard (the old code took the repo lock *inside* the already-held slot lock, the reverse of removal's order) — good catch, correctly fixed.
- **C25** (duplicate insert / double free / abort): `gst_tensor_repo_add_repodata()`'s "not found" path now holds `_repo.repo_lock` continuously from the lookup through the `g_hash_table_insert()`, so two threads can no longer both take the "not found" branch for the same `nth`. `gst_tensor_repo_init()` checks `_repo.initialized` under the same lock instead of racing on it before locking anything.

## Regression / other-module risk
Low. The change is confined to `gsttensor_repo.c/.h`; `gsttensor_reposink.c`/`gsttensor_reposrc.c` are untouched and their call sites (`gst_tensor_repo_add_repodata`, `_set_buffer`, `_get_buffer`, `_set_eos`, `_set_changed`, `_check_eos`/`_check_changed` — the last two now have zero in-tree callers but remain public and unit-tested) keep identical signatures/semantics for the non-racy case. `gst_tensor_repo_get_repodata()` is removed from the header, but it was never installed (only `gst/nnstreamer/elements/meson.build` compiles the .c, header isn't in `install_headers()`) and had no callers outside this file. I did not find any lock-ordering path elsewhere in the tree that would now deadlock against the new `repo_lock → (release) → data->lock` discipline.

## One residual race the fix doesn't close

**[Medium] `gst_tensor_repo_add_repodata()`'s "slot already exists" (reuse) branch can race against a concurrent `gst_tensor_repo_remove_repodata()` on the same slot and return `TRUE` for a slot that is no longer registered.**
`gst/nnstreamer/elements/gsttensor_repo.c`, `gst_tensor_repo_add_repodata()`, roughly lines 138–160 (the `if (data != NULL) { … }` branch).

Sequence:
1. Thread A (e.g. a `tensor_reposink`/`tensor_reposrc` whose `slot-index` property is being set to `X`, which already has a live registrant) takes `GST_REPO_LOCK()`, finds the existing slot via `g_hash_table_lookup`, does `g_atomic_int_inc(&data->refcount)`, then `GST_REPO_UNLOCK()` — *before* touching `data->lock` or returning.
2. Thread B (e.g. the `tensor_reposrc` that currently owns slot `X`, being disposed) runs `gst_tensor_repo_remove_repodata(X)` in that window: takes `GST_REPO_LOCK()`, `g_hash_table_steal()`s the same `data` out of `_repo.hash`, decrements `_repo.num_data`, unlocks, marks `eos = TRUE`, broadcasts, and drops the table's reference.
3. Thread A resumes: locks `data->lock`, flips `sink_changed`/`src_changed`/`pushed`, unlocks, drops its own reference (which may now be the last one, freeing `data`), and returns `TRUE`.

Because of the new refcounting this no longer crashes (that's real progress over `main`), but the slot A just "successfully" re-associated with is not in `_repo.hash` any more — `_repo.num_data` was already decremented by B and the hash entry is gone. The element that called `set_property("slot-index", X, …)` believes it owns a live slot (and both `gst_tensor_reposink_set_property()`/`gst_tensor_reposrc_set_property()` discard `add_repodata()`'s return value anyway, so nothing short-circuits here), but the very next `gst_tensor_repo_set_buffer()`/`get_buffer()` call on `X` will fail `gst_tensor_repo_ref_repodata()`'s lookup, hit `g_return_val_if_fail (data != NULL, …)` (a GLib critical), and the sink will end up raising `GST_RESOURCE_ERROR_WRITE` (same code path as `pipelineSinkErrorsWhenSlotRemoved` exercises) for what looks like a normal, successful slot registration.

This is exactly the class of bug the PR sets out to fix (two entities touching the same `slot-index` concurrently — C25's own description), just a third variant (reuse vs. concurrent remove) that isn't covered by `addSameSlotConcurrently` (which only races two `add_repodata()` calls against each other on *fresh* slots) or any other new test. It's narrow — it needs one element's `slot-index` to be set to an index another element currently owns at the same moment that owner is disposed — but it's realistic for dynamic slot reconfiguration, and nothing in this PR or the test suite would catch a future regression of it either way, since it was never closed.

Suggested direction: give `remove_repodata()` a dedicated `removed` flag (distinct from `eos`, which has stream semantics) set under `data->lock`, and have the reuse branch in `add_repodata()` check that flag immediately after it locks `data->lock`; if set, fall through to the "allocate + insert a fresh entry" path (retrying under `GST_REPO_LOCK()`) instead of declaring success against a stolen slot. A test similar to `addSameSlotConcurrently` but racing one thread's `add_repodata(X)` against another thread's `remove_repodata(X)` on an already-registered `X` would pin this down and protects against regressions here in the future.

## Code size / scope
Appropriately scoped — only `gsttensor_repo.c/.h` plus tests are touched; no unrelated files, no drive-by refactors of `gsttensor_reposink.c`/`gsttensor_reposrc.c`. The 737-line test file is large but proportionate: it's what it takes to deterministically exercise blocking/wakeup races without hanging the test binary (bounded-wait worker threads + deadline join), and per the PR's own negative-control run it actually discriminates the fix (5 of the new cases fail without commits 1–2).

## Test coverage
Good for the stated bugs: `removeWakesBlockedSink`/`removeWakesBlockedSrc`/`removeWakesAllWaiters` cover C26 (including broadcast-not-signal), `addSameSlotConcurrently` covers C25 (8 threads × 2000 fresh slots), and `pipelineSinkErrorsWhenSlotRemoved` exercises the real element pair end-to-end. `unittest_repo` is correctly wired into `tests/meson.build` under the existing `gtest_dep.found()` guard with a 60 s timeout (reasonable given the 5 s worst-case joins per blocking case), and per `packaging/nnstreamer.spec`'s wildcard `%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests` and `run_unittests_binaries.sh`'s recursive discovery, no additional packaging/CI wiring is needed for it to run on Tizen GBS or the Valgrind job. The one gap is the residual race above, which none of the new tests reproduce.

## CI-gating / future-regression detection
The new GTest binary runs under `meson test` (caught by every meson-based CI job) and under the Valgrind job (one more binary in the existing loop), so a future change to `gsttensor_repo.c` that reintroduces the UAF/double-free would fail `removeWakesBlockedSink`/`removeWakesBlockedSrc`/`addSameSlotConcurrently` and, separately, would very likely also surface as a new Valgrind error — both are hard CI gates per the PR body and the repo's existing Valgrind-gate setup (#4936/#4955). Good.

## Documentation / API
No installed/public API changed (the one removed prototype, `gst_tensor_repo_get_repodata()`, was internal-only and undocumented), and the described behavior change (sink now errors, blocked src now gets EOS, instead of hanging) is an internal robustness fix, not a documented interface contract in `Documentation/component-description.md` or elsewhere. No doc update is required, and none was made — consistent.

## Nits
- `gst/nnstreamer/elements/gsttensor_repo.c` — `check_changed()`/`check_eos()` (still public, still unit-tested) now have zero in-tree callers after `get_buffer()` was changed to read `data->src_changed`/`data->eos` directly. Not a problem, just worth a one-line note in a commit/PR description for future readers wondering why they look orphaned in-tree.
- `tests/nnstreamer_repo/unittest_repo.cc:405` — `#include "../gst/nnstreamer/elements/gsttensor_repo.h"` only resolves because `tests/` (from `unittest_util_dep`'s `include_directories('.')`) happens to be on the quote-include search path; it does not resolve relative to the including file's own directory (`tests/nnstreamer_repo/`), which would need `../../gst/...`. It builds today (confirmed by this PR's own green CI), but a self-relative path or `#include "elements/gsttensor_repo.h"` (already reachable via the transitively-included `nnstreamer_inc`) would be less surprising.

## Verdict
Mergeable given CI is green (build, static checks, Valgrind-arm) and the two targeted defects (C25, C26) are genuinely fixed with solid, matching regression tests. The one Medium finding above (add_repodata reuse-branch vs. concurrent remove) is narrow and non-crashing, so it doesn't have to block this PR, but since it's the exact bug class C25 targets and is currently untested and unaddressed, I'd recommend either closing it in this PR or opening an immediate follow-up issue referencing it before merge, rather than letting it go unlogged. Nits are optional polish.

myungjoo:
[Response to the relayed review above, posted on behalf of the author by a Claude agent]

Thanks. No code change for this round; the reasons follow.

**Medium: `add_repodata()` racing `remove_repodata()` on an existing slot.** This is not a defect. In that interleaving, the add takes a reference under the repo lock, and the remove then steals the slot and marks it EOS. The add finishes resetting the flags of the slot it holds, and the slot is freed with the add's reference. That end state is exactly what you get from the serial order *add completes, then remove runs*, which the elements can already produce with no concurrency at all (a `tensor_reposink` sets `slot-index`, then a `tensor_reposrc` on the same index is disposed). In both orders the next `set_buffer()` on that index finds no slot and the sink posts RESOURCE/WRITE. So "add returned TRUE but the slot is gone" is not a lost update; it is a valid linearisation of add-then-remove. A `removed` flag with a re-insert fallback would instead let a racing add *undo* a remove that was already visible to waiters (they got EOS), which is the less consistent choice. On main, this interleaving was a use-after-free (the add wrote the flags of freed memory); with the fix it is memory-safe, and the reference taken in that branch is pinned by the existing cases: dropping it makes `addAndRemoveSlot` free the table's slot on the second add.

**Nit: `check_changed()` / `check_eos()` have no in-tree caller now.** They are kept deliberately. They are declared in the same header and used by the new tests, and removing them would widen a memory-safety fix into an API cleanup.

**Nit: `#include "../gst/nnstreamer/elements/gsttensor_repo.h"`.** This is the form `tests/nnstreamer_if/unittest_if.cc` uses for `gsttensor_if.h`, resolved through the same include directory, so it is kept consistent with the existing element tests.

CI so far: every finished job is green, including the Valgrind job on ubuntu-22.04-arm, the Tizen GBS aarch64/armv7l builds, macOS and Android.

myungjoo:
[Automated review by a separate Claude agent (round 2), relayed to this PR by the author's account]

Scope: same 3 commits (9dbd9ff7, 483c9a1e, b68245c7 / head b68245c7), unchanged since round 1. I re-read the full pre- and post-patch `gsttensor_repo.c`/`.h`, the unchanged callers `gsttensor_reposink.c`/`gsttensor_reposrc.c`, `tests/nnstreamer_repo/unittest_repo.cc`, the build/packaging wiring, and the CI workflow files, and cross-checked the round-1 review against the author's reply on its merits rather than repeating it.

## Round-1 Medium finding vs. the author's reply — I side with the author; decline accepted

Claim under dispute: `add_repodata()`'s reuse branch can race `remove_repodata()` on the same slot (inc ref under repo_lock → repo_lock released → remove steals the slot, marks EOS, drops the table's ref → add resumes, writes flags on an already-removed slot, returns TRUE).

I traced every interleaving of this race by hand against the post-patch code:
- Both `add_repodata`'s lookup and `remove_repodata`'s `g_hash_table_steal` happen under `GST_REPO_LOCK()`, so they cannot interleave at the hash-table level — either the add's lookup finds the slot before the steal (and races only on the *unlocked* tail: `data->lock` + refcount), or after the steal (lookup returns NULL, add returns FALSE-path... actually TRUE via the fresh-insert branch on a *new* slot with the same id, which is a different, non-conflicting case).
- In the racy case, `add`'s own `g_atomic_int_inc()` happened under `repo_lock` before `remove`'s steal, so `add` is guaranteed to hold a live reference by the time it later locks `data->lock` — no UAF is possible, matching the mutation-matrix note ("no refcount increment in lookup: caught by 5 cases").
- The end state (slot flags briefly rewritten on memory that's already unlinked from the hash, then freed by `add`'s own final unref) is observably identical to running the two calls fully serially in the order add-completes-then-remove-runs: in both cases the slot is gone afterward, and the very next `set_buffer()`/`get_buffer()` on that index hits the `g_return_val_if_fail (data != NULL, …)` path and the sink reports `RESOURCE/WRITE` — exactly what `pipelineSinkErrorsWhenSlotRemoved` already exercises. Since both call sites (`gsttensor_reposink.c:193`, `gsttensor_reposrc.c:215`) discard `add_repodata()`'s return value, no caller ever observably distinguishes "TRUE, slot alive" from "TRUE, slot about to vanish."
- This is a genuine linearization argument, not just an assertion: the interleaved execution produces exactly one of the two serial orderings' outcomes, and neither ordering is privileged by the API contract (two independent elements racing to own the same `slot-index` have no synchronization contract with each other today, before or after this PR).

Verdict: **accept the decline**. This is not a new defect and not a regression versus any well-defined pre-PR behavior; it is memory-safe (which is the property C25/C26 are about), and "fixing" it with a `removed` flag + retry would change the semantics in a direction the author correctly identifies as *less* consistent (undoing a remove that already released waiters with EOS). Not a blocker.

## Round-1 Nits vs. the author's reply — both declines accepted

- `check_changed()`/`check_eos()` orphaned in-tree after `get_buffer()` inlines the reads: keeping them because they're still declared/tested is reasonable; removing them would indeed be scope creep beyond a memory-safety fix. Accepted.
- `#include "../gst/nnstreamer/elements/gsttensor_repo.h"` in `unittest_repo.cc`: I independently checked `tests/nnstreamer_if/unittest_if.cc:17`, which uses the identical pattern (`#include "../gst/nnstreamer/elements/gsttensor_if.h"`) from the same sibling-directory depth (`tests/nnstreamer_if/` vs `tests/nnstreamer_repo/`), resolved through the same `unittest_util_dep` `include_directories('.')` in `tests/meson.build`. The claim of consistency with existing convention is correct. Accepted.

## Independent full review

**Does it solve the stated problem?** Yes. C26: `remove_repodata()` now steals the slot under `repo_lock`, marks `eos = TRUE` and broadcasts both conds under `data->lock` (not signal-only, so `removeWakesAllWaiters` genuinely needs broadcast), and frees only on the last `unref_repodata()`. C25: `add_repodata()`'s "not found" path now holds `repo_lock` continuously from lookup through `g_hash_table_insert()`, closing the double-insert window.

**Lock ordering.** I checked every function that can touch both `_repo.repo_lock` and a slot's `data->lock`: `ref_repodata`, `add_repodata`, `remove_repodata`. In all three, `repo_lock` is acquired and released *before* `data->lock` is ever touched — the two locks are never held simultaneously anywhere in the post-patch file. This is what actually closes the deadlock hazard that existed on `main`: the pre-patch `remove_repodata()` called `g_hash_table_remove()` while holding `repo_lock`, and the destroy-notify (`gst_tensor_repo_release_repodata`) took `data->lock` *inside* that hold (repo_lock→data->lock), while the pre-patch `get_buffer()` called `check_changed()`/`check_eos()` (→ `get_repodata()` → `GST_REPO_LOCK()`) while already holding `data->lock` (data->lock→repo_lock) — a real AB-BA pair. The fix removes the nesting entirely rather than just reordering it, which is the more robust choice.

**Refcount balance.** I walked every return path in `set_changed`, `add_repodata`, `set_buffer`, `check_eos`, `check_changed`, `set_eos`, `get_buffer`, `remove_repodata`, including every `g_return_val_if_fail` early-out. In each case where a reference was taken (`ref_repodata` returned non-NULL, or `add_repodata`'s reuse branch incremented directly), exactly one matching `unref_repodata()` is reached on every exit path; where the reference lookup returned NULL, no ref was taken and none is released. `get_buffer()`'s three internal exits (`src_changed`, `eos`, success) all funnel through the single `done:` label before the one `unref_repodata()` call, so there's no per-branch duplication risk. Balanced throughout.

**set_changed/check_\* touching freed memory.** Not observed. All accessors go through `ref_repodata()`, which increments under `repo_lock` before the caller ever dereferences `data`, and the slot cannot be freed while any reference (including one held mid-block by a caller not yet at its own `unref_repodata()`) is outstanding, because `remove_repodata()`'s own unref only drops the table's original reference, never a caller's.

**Dispose mid-state-change / deadlock.** `remove_repodata()` never blocks on what a woken waiter does afterward — it takes `data->lock` only long enough to set `eos` and broadcast, then unlocks and returns. A disposing `tensor_reposrc` therefore cannot be blocked by a sink that is slow to notice its error. `pipelineSinkErrorsWhenSlotRemoved` exercises exactly this: it disposes the `reposrc` while the sink is mid-push, then still succeeds in bringing the pipeline down to `GST_STATE_NULL` within `UNITTEST_STATECHANGE_TIMEOUT` (10 s). I did not find any path where `data->lock` is held while blocking on `repo_lock` or vice versa, so I see no new deadlock surface.

**Test determinism under slow CI.** The 200 ms pre-remove sleeps in `changedSlotEndsWait`/`removeWakesBlockedSink`/`removeWakesBlockedSrc`/`removeWakesAllWaiters` are a best-effort nudge to land the race inside the actual `g_cond_wait()`, not a correctness requirement: I traced the alternate interleaving where `remove_repodata()` runs *before* the worker even reaches `ref_repodata()` (e.g., under heavy scheduling delay) and the worker's call still returns the expected value (`FALSE` / `NULL`+`eos`) immediately via the `data == NULL` early-out, because the slot is already gone. So a "too-early" remove weakens the test's ability to hit the old UAF window but does not make it flaky. The 5 s per-worker deadline and 10 s pipeline-state deadline only need to cover post-broadcast wakeup latency, not the racy window itself, and are consistent with `UNITTEST_STATECHANGE_TIMEOUT` used elsewhere in the suite. Empirically: RISC-V 64 (QEMU-emulated) and the ubuntu-22.04-arm Valgrind leg — the two slowest legs — are both green as of this review; the ubuntu-22.04 (x86) Valgrind leg was still pending when I checked. Worth confirming it lands green before merge, but nothing in the design suggests it's timing-sensitive in a way the arm leg wouldn't already have caught.

**CI gating breadth.** Confirmed by reading the workflow/spec files directly, not just trusting the PR body: `unittest_repo` is exercised by plain `meson test -C build/ -v` (`ubuntu_clean_llvm_build.yml:42`, `ubuntu_clean_meson_build.yml:184`), by both Valgrind legs via `run_unittests_binaries.sh --valgrind --recursive ./tests/` gated by `tools/debugging/check_valgrind_log.sh`, by the RISC-V QEMU leg via `run_unittests_binaries.sh -k unittest_filter_python3 ./tests` (`.github/workflows/risc-v.yml:41`), and by the Tizen GBS x86_64 `unit_test 1` leg via the same recursive script (`packaging/nnstreamer.spec:1087`). The aarch64/armv7l GBS legs intentionally build with `unit_test 0` and skip unit tests for build-speed reasons — that's a pre-existing, unrelated convention, not a gap introduced by this PR. `packaging/nnstreamer.spec`'s `%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests` wildcard already covers the new binary; no packaging change was needed and none was made. A future regression in `gsttensor_repo.c` would fail on at least four independent CI legs, which is a solid gate.

**Docs/API.** I grepped the whole tree for `gst_tensor_repo_get_repodata` (the removed prototype) and found zero remaining references anywhere outside the file it was defined in — safe to remove. `gsttensor_repo.h` is not in `install_headers()` (`gst/nnstreamer/meson.build:145` only installs `nnstreamer_headers`, which doesn't include it), so no installed API changed. `Documentation/component-description.md`'s `tensor_repo_sink`/`tensor_repo_src` entries describe element purpose only, not slot-removal error semantics, so nothing there goes stale. No doc update needed.

**Scope/size.** Confined to `gsttensor_repo.c`/`.h` plus the new test file and an additive `tests/meson.build` hunk; no sink/src element files touched. Appropriately sized for two use-after-free/double-free bugs plus a 15-case regression suite that has to synchronize real threads without hanging the binary.

## New (Low/Nit) observations from this pass

- **[Low]** `gst/nnstreamer/elements/gsttensor_repo.c`, `gst_tensor_repo_init()`: this commit also drops the previously-unconditional `g_mutex_init(&_repo.repo_lock); g_cond_init(&_repo.repo_cond);` and now takes `GST_REPO_LOCK()` as the very first statement, before any initialization check. This is correct — `_repo` is `static` with a partial designated initializer (`{.num_data = 0, .initialized = FALSE}`), so `repo_lock`/`repo_cond` are zero-filled at load time, and a zero-filled `GMutex`/`GCond` is documented by GLib as already valid for use without `g_mutex_init()`/`g_cond_init()`. It's also a genuine improvement: the old unconditional re-init on every "not yet initialized" call was itself racy (two threads could both pass the un-locked `if (_repo.initialized) return;` check and double-`g_mutex_init()` a mutex that's already in use). Not a defect — just suggest a one-line comment at the `_repo` declaration noting the reliance on static zero-initialization, so a future reader doesn't "fix" this by moving `_repo` off static storage or re-adding the init calls in the wrong place.
- **[Nit]** `tests/nnstreamer_repo/unittest_repo.cc`, `wait_ctx_wait_started()`: busy-polls `ctx->started` with no deadline. In practice it's the very first statement of the worker thread body so the window is negligible, but it's the one wait in this file without a bound, unlike `wait_ctx_join_or_timeout`. Optional.

## Verdict

Mergeable as is. The round-1 Medium finding is correctly declined on the merits (a genuine linearizable race, not a defect), both nits are correctly declined, and my own independent pass found no new lock-ordering, refcount, UAF, or CI-gating issues — only two non-blocking Low/Nit observations above. Suggest confirming the still-pending `ubuntu-22.04` Valgrind (non-arm) check goes green before merge, as a matter of process rather than a code concern.

myungjoo:
[Response to the round-2 relayed review, posted on behalf of the author by a Claude agent]

Both non-blocking observations are addressed in place, folded into their commits. The new head is `14dc7c1f`, with the same three commits.

1. **Static repo lock/cond init** (`gsttensor_repo.c`, C25 commit): the `_repo` declaration's doxygen now carries `@note The static repo_lock and repo_cond need no g_mutex_init() or g_cond_init().`, so nobody re-adds the init that raced.
2. **`wait_ctx_wait_started()` had no deadline** (test commit): it now takes the same `REPO_WAIT_TIMEOUT_US` and returns a gboolean, and all five call sites `ASSERT_TRUE` it.

Verified locally before pushing: `-Dwerror=true` build, `unittest_repo` 15/15, gst-indent and clang-format 16 clean, `doxygen-tag.sh` (with ctags) clean, strict `-Wextra -Wsign-compare -Werror` recompile of the test object. The previous head `b68245c7` finished 20/20 green, including both Valgrind legs and the Tizen GBS `unit_test 1` build.

myungjoo:
[Automated review by a separate Claude agent (round 3), relayed to this PR by the author's account]

Scope: round 3, head moved `b68245c7` → `14dc7c1f` (force-push, same 3 commits) with exactly two changes answering round 2, confirmed via `git range-diff b68245c7~3..b68245c7 14dc7c1f~3..14dc7c1f`:
1. A doxygen `@note` on the static `_repo` declaration (`gst/nnstreamer/elements/gsttensor_repo.c`).
2. A deadline + `ASSERT_TRUE` on `wait_ctx_wait_started()` in `tests/nnstreamer_repo/unittest_repo.cc` (all 5 call sites).

I verified both, then did a fresh full pass over the whole PR (diff, both changed files in full, callers in `gsttensor_reposink.c`/`gsttensor_reposrc.c`, `tests/meson.build`) rather than re-stating rounds 1-2.

## Verifying the two round-2 answers

**1. `@note` on `_repo`.** Correctly placed immediately above `static GstTensorRepo _repo = {.num_data = 0,.initialized = FALSE };` (`gsttensor_repo.c:32-34`), and it is accurate: `gst_tensor_repo_init()` (`:389-400`) takes `GST_REPO_LOCK()` as its first statement with no prior `g_mutex_init()`/`g_cond_init()`, relying on `_repo`'s static storage being zero-initialized, which is valid for `GMutex`/`GCond` per GLib. Confirmed correct and complete.

**2. `wait_ctx_wait_started()` deadline.** The function now takes `timeout_us`, polls `g_get_monotonic_time()` against a computed deadline, and returns `FALSE` on timeout; all 5 call sites wrap it in `ASSERT_TRUE (…, REPO_WAIT_TIMEOUT_US)` (`unittest_repo.cc:316,361,396,449-450`). I specifically chased the failure path the operational brief flagged: if `ASSERT_TRUE` fires, gtest returns from the `TEST` body immediately, leaving the just-spawned `GThread` un-joined with a pointer into the (about-to-be-destroyed) stack-local `RepoJob`/`WaitCtx`. I traced what that thread does next in each of the four call sites that use this pattern (`changedSlotEndsWait`, `removeWakesBlockedSink`, `removeWakesBlockedSrc`, `removeWakesAllWaiters`): in every case the thread's only next action is the repo call itself (`gst_tensor_repo_get_buffer`/`set_buffer`), and by test construction that call is designed to block (empty slot for a getter, full slot for a setter) until an explicit unblock (`set_changed`/`remove_repodata`) that this aborted test path never performs. So the thread parks forever inside the repo's `g_cond_wait()` and never reaches `wait_ctx_mark_finished()` — it never touches the freed stack frame. Net effect on this path is a leaked, permanently-blocked thread (and, for the buffer-carrying tests, leaked `GstBuffer`/`GstCaps`), not a use-after-scope write; it's the same "stuck thread, must not be joined" situation `wait_ctx_join_or_timeout()`'s own docstring already documents for its own timeout. Confirmed the fix is correct, and confirmed this residual edge case is memory-safe rather than a new hazard, so no follow-up is needed.

## Fresh full pass — no new findings

- Re-diffed `gsttensor_repo.c`/`.h` end to end against pre-PR `main`: refcounting is balanced on every return path I traced (`set_changed`, `add_repodata`, `set_buffer`, `check_eos`, `check_changed`, `set_eos`, `get_buffer`, `remove_repodata`), `repo_lock` and a slot's `data->lock` are never held simultaneously, and `remove_repodata()` uses `g_hash_table_steal()` (not `g_hash_table_remove()`) specifically so the table's implicit unref happens via an explicit `gst_tensor_repo_unref_repodata()` call made *after* `data->lock` is released — this matches the round-2 analysis and I found nothing it missed.
- No teardown path (`g_hash_table_destroy()`/`_unref()` on `_repo.hash`) exists anywhere in the tree, so the destroy-notify wiring (`gst_tensor_repo_unref_repodata` as the hash table's value-destroy function) is only ever invoked indirectly through this PR's own explicit calls, not at some other, unaudited teardown site.
- `tests/meson.build`: confirmed the new `unittest_repo` block placement between `unittest_if` and `unittest_rate` is consistent with the file's existing (non-alphabetical) ordering convention elsewhere (`unittest_common, unittest_sink, unittest_tensor_region, unittest_decoder_boundingbox, ...`), so no ordering nit here.
- `add_repodata()`'s unguarded read of `_repo.initialized` before taking `GST_REPO_LOCK()` (`g_return_val_if_fail (_repo.initialized, FALSE);`) is structurally identical to the pre-PR code's equivalent check in `gst_tensor_repo_get_repodata()` — not a new or newly-exacerbated race introduced by this diff.
- No architecture/API-doc changes needed: the one removed prototype (`gst_tensor_repo_get_repodata`) has zero remaining references anywhere in the tree, and `gsttensor_repo.h` is not part of the installed API.

I did not find anything rounds 1-2 missed, and I have no new Nits to add — the two round-2 items were the only open items and both are now correctly and completely addressed.

## CI status (new head `14dc7c1f`)

At the time of this review, the fast checks on the new head are green (DCO, Static checks, CPP-Linter, Spell Check, Analyze(python)); the heavier legs (both Valgrind jobs, Tizen GBS x86_64 unit-test build, RISC-V, macOS, Android, Yocto, pdebuild) were still `pending` — they hadn't had time to run yet. The previous head (`b68245c7`) had already finished 20/20 green including both Valgrind legs, per the author's own reply, and the round-3 delta is documentation/test-hardening only (no production-code semantic change), so I don't expect the pending legs to behave differently, but they should still be confirmed green before merge as a matter of process.

## Verdict

Mergeable after the currently-pending CI legs on `14dc7c1f` finish green (no code-behavior changes since the already-fully-green `b68245c7`, only the two documentation/test-hardening answers above, both verified correct and complete). No blocking issues found in this round, and no trivially-fixable Nit is outstanding.

myungjoo:
[Response to the myungjoo-bot review, posted on behalf of the author by a Claude agent]

Both non-blocking items are fixed in the test commit; new head `fbd23f27`, still 3 commits.

**1 (Minor) — `pipelineSinkErrorsWhenSlotRemoved` silently stopped testing C26 under memcheck.** Confirmed from the same log: inside that case the memcheck pass logged `gst_tensor_repo_set_buffer: assertion 'data != NULL' failed`, so the sink reached `set_buffer()` only after the removal and took the missing-slot return, which main passes too. The fixed 300 ms sleep is gone:

- a `GST_PAD_PROBE_TYPE_BUFFER` probe on `rsink`'s sink pad counts buffers, and the case waits with `wait_pipeline_process_buffers (&received, 2U, 5000U)` before removing the slot (plus a short settle, since the probe runs before `render()`);
- the case installs a default log handler for its duration that counts exactly that critical and asserts it is zero, so a timing miss now fails the case instead of passing it. The handler forwards to `g_log_default_handler`, so nothing is swallowed.

Verified locally under the same memcheck wrapper: the case passes and its only critical is the expected `set_eos` one from the EOS event after removal, i.e. the blocked-sink path really ran. Re-checked against pristine main with the new tests: 5 cases still fail there (3 wake cases, `addSameSlotConcurrently`, and this one).

**2 (NIT) — `check_changed()` on an existing slot, and its unref, never ran.** `changedSlotEndsWait` now continues after the join: `check_changed (slot, &id, FALSE)` returns TRUE with `id == newslot`, the sink side returns FALSE, then `set_changed (slot, newslot, TRUE)` and `check_changed (slot, &id, TRUE)` cover the `is_sink` branches of both functions.

**On the surviving mutation.** You are right that the leak gate catches it, and the PR body overstated it; the body now says so. `g_hash_table_steal()` leaves nothing to free the block allocated in `add_repodata()`, so it is definitely lost with a frame of ours, 2000 of them from `addSameSlotConcurrently`.

Local verification of this head: `-Dwerror=true` build, `unittest_repo` 15/15 (also under memcheck, no error from repo code), gst-indent, clang-format 16, `doxygen-tag.sh` with ctags, strict `-Wextra -Wsign-compare -Werror` recompile. The previous head was 20/20 green.

myungjoo:
[Automated review by a separate Claude agent (round 4), relayed to this PR by the author's account]

Scope: round 4, head moved `14dc7c1f` -> `fbd23f27` (force-push, same 3 commits: 9dbd9ff7 C26, 1e869719 C25, fbd23f27 tests). Only `tests/nnstreamer_repo/unittest_repo.cc` changed since round 3; `gsttensor_repo.c`/`.h` are byte-identical to `14dc7c1f`, which I re-diffed against `main` to refresh context. I read the full new test file, the unchanged production diff, and `tests/meson.build`'s wiring. I did not need a local rebuild for these findings; they follow from the control flow of the test file itself and from `wait_pipeline_process_buffers()`'s implementation in `tests/unittest_util.c:184-198` (a plain deadline poll on an atomic counter, already used by ~90 other call sites in this tree, so it is a proven helper, not new machinery).

## Verifying the two answers to the myungjoo-bot review

**1 (Minor, `pipelineSinkErrorsWhenSlotRemoved` silently stopped testing C26 under load).** Fixed correctly. The probe (`unittest_repo.cc:176-184`, `:635`) fires synchronously on the streaming thread as part of `gst_pad_push()`, immediately before the sink's chain function runs, so by the time `wait_pipeline_process_buffers(&received, 2U, 5000U)` (`:646`) observes `received == 2`, the streaming thread is on the same call stack about to invoke `render()` -> `set_buffer()` for the second buffer, which must block because the first buffer is still unconsumed. The follow-up `g_usleep(200*1000)` (`:647`) only has to cover that last, tiny gap, not the whole "sink receives its first buffer" latency the old fixed 300 ms sleep had to absorb — a real improvement, not just a longer wait. The log handler (`count_missing_slot_log`, `:191-199`) is scoped to the `gst_tensor_repo_set_buffer` critical specifically (via `strstr`), so it cannot be tripped by unrelated criticals from `videotestsrc`/`tensor_converter`/GStreamer's own init path. I confirmed no other code in `tests/unittest_util.c` or this file installs a competing `g_log_set_handler`/`g_log_set_default_handler`, and `gsttensor_repo.c` defines no `G_LOG_DOMAIN`, so `g_return_val_if_fail`'s critical does go through the process default handler this test installs — the mechanism is sound in principle.

**2 (NIT, `check_changed()`'s existing-slot path, including its unref, and `set_changed()`'s `is_sink` branch never ran).** Fixed correctly and completely. The new lines in `changedSlotEndsWait` (`:357-365`) exercise `check_changed(is_sink=FALSE)` returning TRUE (the src-changed path set up by the existing `set_changed(..., FALSE)` call), `check_changed(is_sink=TRUE)` returning FALSE (sink not yet changed), then `set_changed(..., TRUE)` followed by `check_changed(is_sink=TRUE)` returning TRUE. That is both branches of `check_changed()` (`gsttensor_repo.c:249-276`) each hit both TRUE and FALSE, and the previously-uncovered `is_sink` branch of `set_changed()` (`gsttensor_repo.c:99-109`, `sink_changed`/`sink_id`) now runs too. Confirmed by reading `gst_tensor_repo_set_changed()`'s source directly.

## New finding (round 4)

### 1. [Minor] `pipelineSinkErrorsWhenSlotRemoved` leaks its log handler and pad probe — dangling onto stack memory — if either of its two `ASSERT_*` calls fails
`tests/nnstreamer_repo/unittest_repo.cc:638-668`

The case installs two pieces of process-global state that capture pointers to its own stack frame:
- `g_log_set_default_handler (count_missing_slot_log, &missing_slot)` at `:638` (restored only at `:668`);
- a `GST_PAD_PROBE_TYPE_BUFFER` probe on `sinkpad` capturing `&received` at `:635` (never explicitly removed; relies on the pad being destroyed).

Between the install at `:638` and the restore at `:668` there are two fatal assertions:
- `ASSERT_TRUE (wait_pipeline_process_buffers (&received, 2U, 5000U));` at `:646`
- `ASSERT_NE (msg, nullptr);` at `:653`

`ASSERT_*` in GTest is a plain early `return;` from the test function (no exception, no destructor-driven cleanup here since nothing owns these registrations). If either fires — e.g. on a loaded runner where the sink doesn't reach `set_buffer()` for the second buffer within 5 s, or where removal doesn't produce a bus message within 5 s — the function returns immediately:
- `g_log_set_default_handler` keeps pointing at `count_missing_slot_log` with `user_data = &missing_slot`, a now-dead stack slot. Any later `g_log()` call with no domain-specific handler (any of the six negative-path `TEST`s that run after this one in the same binary: `removeUnknownSlot_n` through `setBufferAfterEos_n`, all in this same GTest process) invokes the handler with that dangling pointer and writes to it via `g_atomic_int_inc`.
- The pad probe on `sinkpad` keeps firing for any further buffer that reaches the sink (the pipeline is still `PLAYING`, `videotestsrc` has `num-buffers=10` and keeps pushing), writing to `&received` — also dangling — from the GStreamer streaming thread, concurrently with whatever the main thread is doing in later tests.

This is exactly the scenario this round's fix is meant to make loud instead of silent, and it is not hypothetical: the round-3 bot review's own memcheck CI run hit the old fixed-sleep version of this exact race. If the new, tighter 5 s/5 s deadlines are ever missed on a slow runner (valgrind, the RISC-V QEMU build, or the Tizen GBS x86_64 job all run this binary), the result would not be a clean, attributable assertion failure — it would be memory corruption or a crash surfacing in an unrelated, later test case, which is harder to diagnose than the silent-pass bug this round fixed.

**Suggested fix:** make the cleanup unconditional via RAII, since GTest's `ASSERT_*`/`FAIL()` unwind the C++ stack normally (destructors do run):

```cpp
struct ScopedLogHandler {
  ~ScopedLogHandler () { g_log_set_default_handler (g_log_default_handler, NULL); }
};
...
g_log_set_default_handler (count_missing_slot_log, &missing_slot);
ScopedLogHandler restore_log_handler;
```

and either remove the pad probe the same way (`gst_pad_remove_probe (sinkpad, probe_id)` in a similar guard, keeping the probe id returned by `gst_pad_add_probe`), or simplest: move the two `ASSERT_*` at `:646` and `:653` to `EXPECT_*` plus an explicit `if (...) { <cleanup>; return; }`, matching the `if (!x) FAIL() << "...";` idiom already used elsewhere in this same file (e.g. `:350-351`, `:405-406`, `:440-441`) — those don't have this problem only because they register no global state.

## Fresh pass over the rest of the PR

No other issues found. Specifically checked and found no problem in:
- The unchanged `gsttensor_repo.c`/`.h` diff against `main` (refcounting balance, lock ordering, `g_hash_table_steal` vs `remove`, the `add_repodata`/`init` lock-hold fixes for C25) — matches what rounds 1-3 and the bot already verified line-for-line; nothing here changed in this round.
- `tests/meson.build`'s new `unittest_repo` target (`timeout: 60`) — registered correctly, consistent with neighboring targets.
- Doxygen coverage — both new static helpers (`count_buffer_probe`, `count_missing_slot_log`) carry `@brief`/`@details` comments, satisfying the CI doxygen-tag gate.
- No architecture or public-API change (the removed `gst_tensor_repo_get_repodata()` was never installed/exported), so no documentation update is owed.
- Blast radius stays confined to `gsttensor_repo.c/.h` and the new test file; no other module is touched.

## CI state at review time

Head `fbd23f27` (confirmed current via `gh pr view --json headRefOid`): DCO passes; every substantive check (meson+Valgrind x2, static checks, Tizen GBS x3, pdebuild x2, RISC-V, Yocto, ExecuTorch, MacOS, Clang, NDK, CodeQL, CPP-Linter, Typos) is still **pending** — the run was just triggered by this force-push and had not produced results yet at review time. Recommend confirming these are green (particularly the two Valgrind jobs, since they're what would exercise `pipelineSinkErrorsWhenSlotRemoved` under the load conditions relevant to finding 1 above) before merge.

## Verdict

The two answers to the round-3 bot review are correct and complete. One new, non-blocking issue was found: the new log-handler/pad-probe cleanup in `pipelineSinkErrorsWhenSlotRemoved` isn't exception/early-return-safe, which could turn a rare timing miss into a confusing crash in a later, unrelated test case rather than a clean failure. This is a **Nit that is trivially fixable in this PR** — a small RAII guard (or restructuring the two `ASSERT_*` calls to clean up before returning) confines the fix to a few lines in the same test function, no other change needed. Nothing else blocks merge.

myungjoo:
[Response to the round-4 relayed review, posted on behalf of the author by a Claude agent]

The round-4 Nit is valid and fixed; new head `576a9972`, still 3 commits, test file only.

`pipelineSinkErrorsWhenSlotRemoved` registered a pad probe and a default log handler that both point into the test's own frame, and the two `ASSERT_*` between them and the cleanup could return early, leaving both live against a dead frame and corrupting a later case in the same binary. That is precisely the loaded-runner path this round hardened, so it is worth closing.

The case now drops both as soon as the bus message has been popped, before any check that can leave early:

- `gst_pad_add_probe()`'s id is kept and the probe is removed explicitly; the handler is restored on the same lines, followed by the `sinkpad`/`reposink`/`bus` unrefs;
- the remaining checks became `EXPECT_*` guarded by `if (msg != NULL)` and `if (error != NULL)`, so the state change to NULL and the pipeline unref always run;
- the buffer wait is an `if` with `ADD_FAILURE ()` on timeout rather than an `ASSERT_TRUE`, for the same reason.

Verified on this head: `unittest_repo` 15/15, and under memcheck the case still takes the blocked-sink path (its only critical is the expected `set_eos` after removal, and the missing-slot counter is 0). Negative control against pristine main with these tests: `removeWakesBlockedSink` and `removeWakesAllWaiters` time out, `removeWakesBlockedSrc` SIGSEGVs, `addSameSlotConcurrently` aborts on the duplicate key, and this case now fails and hangs at teardown (exit 124) instead of passing. Also re-ran `-Dwerror=true` build, gst-indent, clang-format 16, `doxygen-tag.sh` with ctags, and the strict `-Wextra -Wsign-compare -Werror` recompile.

myungjoo:
[Automated review by a separate Claude agent (round 5), relayed to this PR by the author's account]

Scope: round 5, head moved `fbd23f27` -> `576a9972` (force-push, same 3 commits: 9dbd9ff7 C26, 1e869719 C25, 576a9972 tests). Confirmed via `git diff fbd23f27 576a9972 -- gst/nnstreamer/elements/gsttensor_repo.c gst/nnstreamer/elements/gsttensor_repo.h tests/meson.build` (empty) that only `tests/nnstreamer_repo/unittest_repo.cc` changed, and via `git range-diff fbd23f27~3..fbd23f27 576a9972~3..576a9972` that the delta is confined to `pipelineSinkErrorsWhenSlotRemoved`. I re-read the full new test file and re-checked the unchanged production diff against `main`.

## Verifying the answer to the round-4 finding

Round 4 flagged that `pipelineSinkErrorsWhenSlotRemoved` installed a pad probe and a default log handler pointing at its own stack frame (`&received`, `&missing_slot`), with two fatal `ASSERT_*` calls between the install and the cleanup that could return early and leave both dangling against a dead frame while later tests in the same binary ran.

The fix (`unittest_repo.cc:607-693`) is correct and complete:
- The probe id is now captured (`probe_id = gst_pad_add_probe(...)`, `:636-637`) and removed explicitly with `gst_pad_remove_probe (sinkpad, probe_id)` (`:663`), rather than relying on pad/element teardown.
- The former `ASSERT_TRUE (wait_pipeline_process_buffers (...))` (`:648`) is now an `if`/`else ADD_FAILURE()` (`:648-651`) — non-fatal, does not return.
- The former `ASSERT_NE (msg, nullptr)` (`:653` in `fbd23f27`) is now `EXPECT_NE` (`:669`) followed by `if (msg != NULL) { ... }` (`:671-688`) guarding every subsequent check, including the inner `error != NULL` guard (`:679-685`) around the domain/code checks — also non-fatal.
- I walked every statement between the probe/handler install (`:636-640`) and the cleanup block (`:662-666`): `EXPECT_EQ` at `:642`, the `if`/`ADD_FAILURE` at `:648-651`, `gst_object_unref (reposrc)` at `:653`, and the blocking bus pop at `:655-656`. None of these can return from the test function — `EXPECT_*`/`ADD_FAILURE` record and continue, and a `GTEST_*` fatal assertion inside a called helper (e.g. `setPipelineStateSync`, `wait_pipeline_process_buffers`) only returns from that helper, not from the caller (gtest's well-known `ASSERT_*`-in-void-helper behavior) — so it cannot make this test frame disappear mid-flight either. The only way to skip the cleanup at `:662-666` now is a full process abort (fatal `g_error`/crash), and in that case the process is gone anyway, so there is no "later test" left to corrupt.
- Cleanup order is right: log handler restored and probe removed before `sinkpad`/`reposink`/`bus` are unreffed, and all of that happens before the `EXPECT_EQ(0U, missing_slot)` / message checks, so the discriminating assertions still run against valid data.
- The case still discriminates the bug: verified logically (I did not rebuild) that a too-early removal still drives `missing_slot` non-zero and/or a timeout `ADD_FAILURE`, matching the round-4 author reply's local verification (blocked-sink path taken, only the expected `set_eos` critical, `missing_slot == 0`).

This closes the round-4 finding; I found nothing incomplete or incorrect about it.

## Same class of problem elsewhere in the file

I grepped the whole file for `gst_pad_add_probe`, `g_log_set_default_handler`/`g_log_set_handler`, and every `g_thread_new` call, and checked what sits between thread creation and the join for each. `pipelineSinkErrorsWhenSlotRemoved` is the only place that registers process-global state pointing at a stack frame, so that specific hazard is fully closed.

There is a narrower instance of the general pattern ("if an `ASSERT_*` fires, something keeps referencing this frame after it's gone") in four cases that spawn a worker thread and then `ASSERT_TRUE (wait_ctx_wait_started (&job.ctx, REPO_WAIT_TIMEOUT_US))` before ever joining it:
- `changedSlotEndsWait`, `unittest_repo.cc:344-345`
- `removeWakesBlockedSink`, `:399-400`
- `removeWakesBlockedSrc`, `:434-435`
- `removeWakesAllWaiters`, `:485-489` (two threads/two asserts)

`job`/`job.ctx` are locals of the `TEST` function, passed to `g_thread_new()` by pointer. If `wait_ctx_wait_started()` times out (5 s) and the `ASSERT_TRUE` fires, the function returns immediately; the just-created `GThread` is left running and unjoined, still holding a pointer into what is now a reused stack region once GTest moves on to the next `TEST`. `repo_job_thread()`'s first action is `wait_ctx_mark_started (&job->ctx)` (an atomic write to `ctx->started`), so if the OS genuinely hasn't scheduled the new thread at all within the 5 s window, that write happens after the frame is gone — the same class of bug round 4 just fixed, not merely a "parks forever, never touches freed memory" case (that safe outcome only holds if the thread had already reached `wait_ctx_mark_started()` by the time the deadline fired, which isn't guaranteed by the code, only made overwhelmingly likely by it).

This is already something round 3 looked at and accepted as memory-safe; that conclusion is optimistic for the reason above, but I don't think it changes the verdict:
- Unlike the fixed issue, this one has never been observed to trigger across 5 rounds of CI on this PR, including the two slowest legs (`ubuntu-22.04-arm` Valgrind and RISC-V under QEMU) and memcheck's own slowdown — `wait_ctx_wait_started()`'s poll only needs the new thread to make its very first atomic write within 5 real seconds, which is a far smaller ask than `pipelineSinkErrorsWhenSlotRemoved`'s bug (which needed a full GStreamer pipeline to reach a specific streaming state within a fixed window, and did fail under load in round 3's own memcheck run).
- `addSameSlotConcurrently` (`:559-599`), which spawns 8 threads, shows the safer pattern already in this same file: no `ASSERT_*` between `g_thread_new()` and an unconditional `g_thread_join()` loop (`:588-589`), so nothing there can return early while a thread is outstanding.

**Impact:** Low/theoretical — would need >5 s of total scheduling starvation for a newly created GThread on a runner that is otherwise making forward progress (the enclosing binary has its own 60 s `meson.build` timeout). Not blocking. Worth a follow-up if this file is touched again: either use the same "unconditional join, non-fatal check" shape as `addSameSlotConcurrently`, or drop the `ASSERT_TRUE` to a `if (!...) { g_thread_join or detach as a leak-and-report; FAIL(); return; }` pattern so a timeout still can't outlive the frame it points at.

## Fresh pass over the rest of the PR

- Production diff (`gsttensor_repo.c`/`.h`) is byte-identical to what rounds 1-3 reviewed line-for-line (refcount balance on every exit path, `repo_lock`/`data->lock` never held simultaneously, `g_hash_table_steal()` + explicit unref, `add_repodata`/`init` lock-hold fix for C25) — nothing new to add.
- `tests/meson.build`'s `unittest_repo` registration is unchanged and correct.
- No architecture or public/installed API change (the removed `gst_tensor_repo_get_repodata()` was never installed or called outside the file), so no documentation update is owed and none is missing.
- Scope stays confined to `gsttensor_repo.c/.h` and the new test file; no other module touched.

## CI state at review time (head `576a9972`)

All 18 substantive checks are currently `pending` (just triggered by this force-push): both Valgrind legs, both Tizen pdebuild/GBS families, RISC-V, Yocto, ExecuTorch, macOS, Clang, Android NDK, CodeQL analyze (c-cpp/python), CPP-Linter, Static checks, Spell check. Only `DCO` has finished (`pass`). Recommend confirming green — especially both Valgrind legs and the arm/QEMU legs, since those are what would exercise `pipelineSinkErrorsWhenSlotRemoved` under the load conditions relevant to the finding this round closed — before merge.

## Verdict

The round-4 finding is correctly and completely fixed; I found no gap in it. One new, non-blocking observation: the same "thread points at a frame that can vanish on an early return" pattern exists in a narrower, much lower-probability form in `changedSlotEndsWait`/`removeWakesBlockedSink`/`removeWakesBlockedSrc`/`removeWakesAllWaiters` (`ASSERT_TRUE` on `wait_ctx_wait_started()` before any join). It has not manifested in 5 rounds of CI including the slowest legs, and it is optional polish, not something this PR needs to fix before merging — the actually-reproduced bug (the one round 3/4 caught) is closed. Nothing else in the PR needs changing.

**Is anything still worth doing in this PR?** No blocking item remains. If the author wants a fully closed loop, the only thing left on the table is the low-probability `wait_ctx_wait_started()` early-return pattern above, which is optional (same severity class as a nit, not reproduced, and the file already demonstrates the safer alternative in `addSameSlotConcurrently`). Otherwise: confirm the currently-pending CI finishes green on `576a9972`, then this is ready to merge.

myungjoo:
[Response to the round-5 relayed review, posted on behalf of the author by a Claude agent]

The optional item is taken rather than left; new head `321c17c5`, still 3 commits, test file only.

The four threaded cases kept their `RepoJob` (and its `WaitCtx`) in the `TEST` frame, so any fatal check between `g_thread_new()` and the join left a worker that would later write into a dead frame. That is now impossible: `repo_job_new()` allocates the job, `repo_job_free()` releases it after a successful join, and a case that gives up on a stuck worker simply leaves the job allocated, which the helper's doxygen says. The leak is bounded, on a failing path only, and in test code, so it cannot trip the valgrind gate for `gst/` and `ext/`.

This also closes the older variant of the same observation: the documented "timed out, must not be joined" path in `wait_ctx_join_or_timeout()` no longer leaves the worker pointing at the case's frame either.

Verified on this head: `unittest_repo` 15/15; under memcheck `pipelineSinkErrorsWhenSlotRemoved` still takes the blocked-sink path (only the expected `set_eos` critical, missing-slot counter 0); against pristine main with these tests the same 5 cases fail (2 time out, 1 SIGSEGV, 1 abort, 1 exit 124). `-Dwerror=true` build, gst-indent, clang-format 16, `doxygen-tag.sh` with ctags and the strict `-Wextra -Wsign-compare -Werror` recompile are all clean.

With this, every item from rounds 1-5 and the bot review is either fixed or answered, and nothing is deferred to a follow-up.

myungjoo:
[Automated review by a separate Claude agent (round 6), relayed to this PR by the author's account]

Scope: round 6, head moved `576a9972` -> `321c17c5` (force-push, same 3 commits: 9dbd9ff7 C26, 1e869719 C25, 321c17c5 tests). Confirmed via `git diff 576a9972 321c17c5 -- gst/nnstreamer/elements/gsttensor_repo.c gst/nnstreamer/elements/gsttensor_repo.h tests/meson.build` (empty) and `git range-diff 576a9972~3..576a9972 321c17c5~3..321c17c5` that the only delta is in `tests/nnstreamer_repo/unittest_repo.cc`: a new `RepoJob *repo_job_new()/void repo_job_free()` pair, and the four threaded cases (`changedSlotEndsWait`, `removeWakesBlockedSink`, `removeWakesBlockedSrc`, `removeWakesAllWaiters`) switched from a stack-local `RepoJob job = {}` to a heap job obtained from `repo_job_new()` and released with `repo_job_free()`. This round was asked to (a) verify that answer, (b) do a final full pass, and (c) report CI state for `321c17c5`.

## (a) Verifying the round-5 answer

Read the full new file (`tests/nnstreamer_repo/unittest_repo.cc`, 837 lines) at `321c17c5`, not just the range-diff.

**The concern round 5 raised is closed correctly.** `repo_job_new()` (`:179-189`) heap-allocates via `g_new0` and its doxygen (`:174-177`) states the invariant precisely: *"A worker that is still blocked when its case gives up keeps writing to the job, so the job may not live in the case's stack frame. Such a job is left allocated on purpose; a joined one is freed."* `repo_job_free()` (`:194-199`) only clears the `WaitCtx` and frees the block; it carries the matching contract note ("only safe once its worker thread has been joined").

I traced all four call sites end to end:
- **`changedSlotEndsWait`** (`:357-395`): job allocated `:368`, `g_thread_join(thread)` unconditionally follows a successful `wait_ctx_join_or_timeout` (`:376-378`), all field reads (`job->out_buffer`, `job->newid`) and the `check_changed` calls happen next, and `repo_job_free(job)` (`:393`) is the last statement touching `job`, after every read. If `wait_ctx_wait_started` (`:371`) or `wait_ctx_join_or_timeout` (`:376`) fails, the function returns before `g_thread_join`/`repo_job_free` — the job (and, in the first case, the `GThread` handle) is deliberately leaked rather than freed against a possibly-still-running worker.
- **`removeWakesBlockedSink`** (`:401-439`) and **`removeWakesBlockedSrc`** (`:445-470`): identical shape — join always precedes the field read and `repo_job_free`.
- **`removeWakesAllWaiters`** (`:476-532`): two independent jobs; each is joined only if its own `wait_ctx_join_or_timeout` succeeded (`:512-518`), and `repo_job_free` for both (`:526-527`) comes after the `ASSERT_TRUE(ok1)`/`ASSERT_TRUE(ok2)` gate, i.e. only on the path where both joins already happened. On a partial failure (one worker stuck), the case returns via `ASSERT_TRUE` before freeing either job — including the one that *did* join — which is a harmless extra leak on an already-failing path, not a correctness issue.

**No UAF is introduced.** In every straight-line (non-early-return) path, `repo_job_free()` is reached strictly after the corresponding `g_thread_join()`, and no code reads `job`/`job1`/`job2` after its `repo_job_free()` call. On every early-return path the job is heap memory that is simply never freed (an intentional, bounded, test-only leak on a failure path that should never fire in a healthy CI run), so a still-running worker thread continues writing into memory that stays valid indefinitely — nothing reclaims or reuses that block. This is strictly better than the previous stack-local version, where the same early-return left a worker thread writing into a `TEST` frame that had already returned. Round 5's "OPTIONAL" item is fully and correctly addressed; I found no new defect in the mechanism.

**The four cases still discriminate the original bugs identically.** The assertions that decide pass/fail are untouched by this round's change (`EXPECT_EQ(job->out_buffer, nullptr)` / `EXPECT_EQ(job->newid, newslot)` for `changedSlotEndsWait`; `EXPECT_FALSE(job->set_ret)` for the two sink-blocking cases; `EXPECT_EQ(job->out_buffer, nullptr)` + `EXPECT_TRUE(job->eos)` for `removeWakesBlockedSrc`) — only the storage location of `job` changed, not what is asserted about it.

The non-threaded cases (`addSameSlotConcurrently`, `pipelineSinkErrorsWhenSlotRemoved`, and the `_n` negative cases) are untouched by this round, as expected — they don't use `RepoJob` at all.

## (b) Final full pass over the whole PR

Production code (`gst/nnstreamer/elements/gsttensor_repo.c`/`.h`) has been unchanged since round 1 (commits `9dbd9ff7`/`1e869719`, byte-identical across every round including this one) — I agree with that carry-forward and re-diffed both files against `main` to re-confirm: reference counting via `refcount`/`gst_tensor_repo_ref_repodata()`/`gst_tensor_repo_unref_repodata()`, `add_repodata()`'s lookup-then-insert under one `repo_lock` hold, `remove_repodata()`'s steal-then-broadcast-then-unref, and `get_buffer()` reading `src_changed`/`eos` directly — all as analyzed in rounds 1-5.

Scope check: `git diff --stat main...321c17c5` shows only `gsttensor_repo.c` (+68/-80), `gsttensor_repo.h` (+4/-3), `tests/meson.build` (+10), and the new `tests/nnstreamer_repo/unittest_repo.cc` (+837). No other module is touched, consistent with the "C25/C26 only" scope. `tests/meson.build`'s new stanza matches the surrounding `unittest_*` entries (dependencies, `install-test` option, `timeout: 60`). Every new function and every `TEST()` case in the test file carries a doxygen block immediately above it (checked programmatically for all 29 declarations/cases) — satisfies the repo's doxygen-per-function CI gate.

No architecture or public-API documentation update is needed: the one API-shaped change, dropping `gst_tensor_repo_get_repodata()` from the header, is an uninstalled internal header with no callers outside this file (re-confirmed), and `Documentation/component-description.md`'s `tensor_repo_sink`/`tensor_repo_src` entries describe user-facing element behavior, which this PR does not change (a removed slot now errors/EOSes a blocked waiter instead of hanging on freed memory — strictly a robustness fix, not a new documented contract).

I have no new findings.

## (c) CI state for head `321c17c5`

As of this review (head force-pushed only minutes earlier), `gh pr checks 4976` shows 19 checks total: **3 passed** (`DCO`, `Static checks`, `Build on MacOS-latest with Meson`), **16 still pending**, **0 failed**. Not yet reported: both Valgrind legs (`ubuntu-22.04`, `ubuntu-22.04-arm`), all three Tizen GBS builds (including the `x86_64, unit_test 1` leg that runs `unittest_repo`), the RISC-V/Yocto/ExecuTorch/Android builds, both `pdebuild` legs, `CPP-Linter`, `Spell Check`, and both `Analyze` (CodeQL) jobs. Given rounds 1-5 saw these same legs finish green on earlier heads with byte-identical production code and closely related test code, I expect the same result here, but that is an expectation, not a confirmed result — nothing has actually gated on this exact head yet.

## Verdict

Rounds 1-5 and the bot review's items are all fixed or answered on their merits, production code has been stable since round 1, and this round's test-only change correctly closes the last optional item without introducing any new hazard. From a code-review standpoint, **nothing further is needed in this PR**. The one open item before merge is procedural, not a review finding: the remaining 16 checks on `321c17c5` (especially both Valgrind legs and the Tizen GBS `unit_test 1` build, which is what actually runs `unittest_repo`) still need to turn green.

myungjoo:
[CI fix, posted on behalf of the author by a Claude agent]

`Spell Check with Typos` failed on the previous heads, and it was mine, not infrastructure: `typos` rejects the word "criticals" in a doxygen line I added with the log handler (`unittest_repo.cc:215`). Reworded to "Log handler counting what set_buffer() logs for a missing slot."; it was the only hit in the tree. New head `8ed082c1`, comment text only, no code change.

Re-verified on this head: `unittest_repo` 15/15, gst-indent, clang-format 16, `doxygen-tag.sh` with ctags and the strict `-Wextra -Wsign-compare -Werror` recompile all clean.

PR #4977, opened 2026-09-16
[Element/SrcIIO] Bound the channel, name and buffer reads of tensor_src_iio

myungjoo:
- **What this fixes.** Four memory-safety defects of `tensor_src_iio`, items **C27**, **C28**, **C29** and **C30** of #4920: a channel stored in 3, 5, 6 or 7 bytes is read past the block the scans were read into; the `channels` property leaks its table on every write and leaves the element in the custom mode with no table when a value does not parse (so reading the property back walks an uninitialised iterator); `scan-size * buffer-capacity` wraps in a `guint` kept in a `gint`, after which the parse loop walks the full unwrapped extent; and the scan for the trailing digits of a channel name reads before the start of the name.
- **How.** One commit per item, each a local guard with no change to the data path or the element API: copy the bytes a channel occupies into a zeroed container instead of dereferencing a wider pointer; parse the channel list into a table of its own and hand it over only when the whole value is accepted; refuse a buffer capacity whose reads do not fit in a `gint` where the scan size becomes known; bound the digit scan by the length of the name.
- **Result.** `unittest_src_iio` grows from 17 to 25 cases and all of them pass; the whole meson suite is 22/22 in a container build. Under valgrind the unfixed element reports an invalid read of size 4 and of size 8 in `gst_tensor_src_iio_process_scanned_data()`, an invalid read of size 1 in `gst_tensor_src_iio_get_generic_name()`, an invalid read in `g_hash_table_iter_next()` from `get_property()` and 384 bytes definitely lost in 4 blocks; the fixed element reports none of these and no definite loss. No other module is touched: every function changed here is static to `gsttensor_srciio.c`.

## What changes for a user

| property / input | before | after |
| --- | --- | --- |
| a channel of `le:s24/24>>0` and the like | decoded correctly, reading up to 3 bytes past the buffer | decoded from its own bytes only |
| `channels=1,2` written twice | the first table leaks | the first table is released |
| `channels=abc` after `channels=1,2` | custom mode with no table; reading `channels` back is undefined | refused, `1,2` stays in effect |
| `channels=-1` or an index beyond a `gint` | accepted and truncated into some other channel | refused |
| `buffer-capacity` whose reads exceed `G_MAXINT` | the device is set up and the first buffer misbehaves | the state change fails with an error naming the capacity and the scan size |

## Notes for reviewers

- **The decoded values do not change** in the C27 commit. For a channel narrower than its container the bytes beyond it were masked out (little endian) or shifted out (big endian); they are zero now. The 10 pre-existing `dataVerifyNoTriggerBits*` cases pass unchanged, and the two new cases pin a 3 byte and a 7 byte channel in both endiannesses and both signs, which nothing covered before because the test device builder rounds the storage size up to 1, 2, 4 or 8 bytes.
- **`bufferCapacityWhileConfigured_n` does not discriminate the C29 fix and is not meant to.** The capacity is validated once, when the device is configured, which is sound only while the property cannot move afterwards. `set_property()` already refuses every write outside NULL/READY; the case pins that invariant so a later change that makes the property mutable while running fails here instead of re-opening the overflow.
- **Negative control.** Built against the unfixed element, `bufferCapacityOverflow_n` and `channelsCustomInvalid_n` fail (the latter with 10 GLib criticals from the missing table), while the C27 and C30 cases pass and are witnessed by valgrind alone, as the over-read bytes are discarded by the shift or the mask and the digit scan stops at whatever byte precedes the name. The CI valgrind job builds this binary, and its memcheck gate is what keeps both from coming back.
- **Not fixed here, seen while working on it:** `gst_tensor_get_size_from_channels()` divides by `storage_bytes`, which is 0 when `scan_elements/*_type` declares 0 storage bits, so such a device raises SIGFPE. That is the `used_bits == 0` family of item H9 of #4960 and belongs with it.

This closes items C27, C28, C29 and C30 of #4920.


myungjoo:
**Note:** This review was produced by a separate AI review agent (not the one that authored this PR) and is being relayed here verbatim by the requester. Treat it as a third-party code review, not as a comment from the PR author.

## Scope check — does it close C27–C30 of #4920?

Verified all four against the current diff and a container build (Ubuntu, meson/ninja, gtest, valgrind):

- **C27** (channels stored in 3/5/6/7 bytes read via `guint32*`/`guint64*`, over-reading the buffer): fixed by zero-initializing `value` and `memcpy`-ing exactly `prop->storage_bytes` in the `case 3:`/`case 4:` and `case 5..8:` blocks of `gst_tensor_src_iio_process_scanned_data()`. I traced the big-endian and little-endian arithmetic by hand for the 3-byte case (masking/shifting by `storage_bits` is unchanged) and confirmed the decoded value is bit-for-bit identical to before, matching the PR's own claim. Confirmed with `dataVerifyExactStorage24`/`56` plus the 10 pre-existing `dataVerifyNoTriggerBits*` cases, all passing.
- **C28** (`custom_channel_table` leaked on re-assignment; failed parse left `CUSTOM` mode with a NULL table): fixed by building into a local `table` and only swapping it into `self->custom_channel_table` (destroying the old one first) when the whole value parses; on failure the old table/mode is left untouched, and `get_property` now also has a defensive NULL-table guard. Confirmed by `channelsCustom` and `channelsCustomInvalid_n`.
- **C29** (`scan_size * buffer_capacity` wraps in a 32-bit `guint`): fixed by a 64-bit checked comparison (`(guint64) self->scan_size * self->buffer_capacity > G_MAXINT`) placed in `gst_tensor_src_iio_setup_scan_channels()`, right after `scan_size` becomes known and before the value is ever used as a `gint` byte count. Confirmed by `bufferCapacityOverflow_n`.
- **C30** (digit-scan in `gst_tensor_src_iio_get_generic_name()` walks before the string start): fixed by adding `digit_len <= channel_name_len` to the loop condition. Traced the boundary manually (empty string, all-digit string) — no off-by-one. Confirmed by `channelNameDigitsOnly`/`channelNameEmpty_n`.

## Dynamic verification performed

Built this exact source in a container (meson/ninja) and ran:
- Full `unittest_src_iio`: **25/25 pass** (matches the PR's claim, 17→25 cases).
- **Negative control**: reverted only `gsttensor_srciio.c` to the pre-fix version and reran the new cases — `bufferCapacityOverflow_n` and `channelsCustomInvalid_n` fail as the PR describes (C27/C30 cases pass either way, as the PR itself notes, since they're witnessed by valgrind rather than gtest assertions).
- **Valgrind** (`--leak-check=full --track-origins=yes`) on the fixed build, full suite: 0 bytes definitely lost, no invalid-read/invalid-write/uninitialised-value errors. There are 24 "possibly lost" blocks, but they all trace to `build_dev_dir_dev_data()` in the *test harness* (pre-existing, unrelated to this PR, not in `gst/` or `ext/`), not to `gsttensor_srciio.c` — so this doesn't affect the fix and is outside the shipped-code leak-gate's scope anyway.
- `tools/development/gst-indent` on the C file: no diff. `clang-format --dry-run` on the C++ test file: clean. Doxygen-tag checker: passes (verified the checker itself catches a missing `@brief` via a local negative control).

## Regression / scope risk

- Only 2 files touched (`gsttensor_srciio.c`, `unittest_src_iio.cc`), no header changes, no ABI/API change. Every function touched is `static` to this file, so there's no cross-module blast radius — verified there are no other callers of these statics.
- The C27 fix additionally removes an unaligned-pointer dereference (the old `*(guint32*)`/`*(guint64*)` casts weren't guaranteed aligned), which is a welcome side benefit, not a behavior change.
- Code size is proportionate: the production fix is roughly 40 net lines across 4 independent, narrowly-scoped hunks; the remainder of the diff is new regression tests, consistent with this repo's test policy.
- No SSAT tests exist for this element (GTest is the only layer used for it already), so adding cases to `unittest_src_iio.cc` is the right layer and they run automatically under the existing `meson test` / CI job — a future regression here would fail CI, not just be silently missed.

## Minor / non-blocking notes (informational only, no change requested)

1. `gst_tensor_src_iio_get_property()`'s new `else if (self->custom_channel_table == NULL)` branch is now unreachable in practice: with the C28 fix, `channels_enabled` can only become `CHANNELS_ENABLED_CUSTOM` in the same branch that assigns a non-NULL table, and the two are never set independently elsewhere. Harmless defensive code, but worth knowing it's not actually exercised by `channelsCustomInvalid_n` (that test never reaches `CUSTOM` with a NULL table anymore, by construction).
2. The added `errno = 0;` before `g_ascii_strtoll()` in the same loop is a good, minimal, in-scope correctness improvement (without it, a stale `errno` from an unrelated earlier call could make the very first parse iteration spuriously fail) — not one of C27–C30 but directly adjacent to the touched lines, so no scope-creep concern.
3. No architecture or public API changed, so no documentation update is expected or missing — checked and there is no `Documentation/*.md` describing `tensor_src_iio` buffer-capacity/channels behavior that would need to change.

## Verdict

No blocking issues found. The four defects are each independently and correctly fixed, tests are proportionate and demonstrably effective (verified via a negative-control build), and the change carries no cross-module or ABI risk.

myungjoo:
**Note:** This review was produced by a separate AI review agent (not the one that authored this PR, and not the agent that posted the previous review comment on this PR) and is being relayed here verbatim by the requester. Treat it as a third-party code review, not as a comment from the PR author.

This is a fresh, independent re-review of the current head (`db4c9496`), done without assuming the previous review's conclusions are correct. It specifically re-verifies the last-commit amend (removal of the `custom_channel_table == NULL` guard from `gst_tensor_src_iio_get_property()`'s CHANNELS branch, which the prior round had flagged as dead code).

## Focus: is the guard removal safe?

Grepped every reference to `custom_channel_table` (8 sites total) and traced each:
- `gst_tensor_src_iio_init()`: sets it to `NULL` alongside `channels_enabled = CHANNELS_ENABLED_AUTO` (not `CUSTOM`).
- `gst_tensor_src_iio_set_property()` (CHANNELS case): this is the **only** site in the file that sets `channels_enabled = CHANNELS_ENABLED_CUSTOM`, and it does so in the same `if (status)` block that unconditionally assigns `self->custom_channel_table = table` (a table that was just `g_hash_table_new`'d, never NULL) immediately before. The `else` branch (parse failure) destroys the local `table` and touches neither `channels_enabled` nor `custom_channel_table`.
- `gst_tensor_src_iio_get_property()`: reads the table only in the `else` branch that is reached exactly when `channels_enabled == CHANNELS_ENABLED_CUSTOM`.
- `gst_tensor_src_iio_setup_scan_channels()`: same — only touches the table under `if (self->channels_enabled == CHANNELS_ENABLED_CUSTOM)`.
- `gst_tensor_src_iio_finalize()`: already NULL-checks before `g_hash_table_destroy()`, independent of `channels_enabled` (correct, since `CUSTOM` mode may never have been entered).

So the invariant `channels_enabled == CUSTOM  <=>  custom_channel_table != NULL` holds by construction, is established exactly once, and no other code path can split the pair. The guard removed in the amend was indeed unreachable dead code, and removing it does not reintroduce C28. Confirmed empirically too (see below): `channelsCustomInvalid_n` reports 0 GLib criticals against the current head.

## Empirical verification (container build: meson/ninja/gtest/valgrind)

- Built the exact PR head content (verified by `sha256sum` against `git show HEAD:...`) and ran `unittest_src_iio`: **25/25 pass.**
- Ran the fixed binary under `valgrind --leak-check=full --track-origins=yes`: **0 bytes definitely lost, no invalid-read/invalid-write anywhere in the log.** (24 "possibly lost" blocks all trace to `pthread_create`/TLS allocation from the test harness's `build_dev_dir_dev_data()`, unrelated to `gsttensor_srciio.c`.)
- To independently confirm the bugs are real (not just trusting the PR description), swapped **only** `gsttensor_srciio.c` for the pre-fix `main` version (keeping the new test file) and rebuilt:
  - `bufferCapacityOverflow_n` and `channelsCustomInvalid_n` fail, the latter with **exactly 10 GLib criticals** as the PR body claims.
  - Running the C27/C30-covering tests alone under valgrind reproduces **Invalid read of size 4** at `gst_tensor_src_iio_process_scanned_data (gsttensor_srciio.c:2416)`, **Invalid read of size 8** at the same function (line 2439), and **Invalid read of size 1** at `gst_tensor_src_iio_get_generic_name (gsttensor_srciio.c:810)` — matching the PR's stated valgrind findings function-for-function.
  - Restored the fixed source afterward and rebuilt; confirmed 25/25 pass again (container left as found).
- Style: extracted both changed files via `git show HEAD:<path>` (the local Windows worktree checkout has CRLF line endings, which made a first pass at this checker fail spuriously — normalizing to the actual git blob content, LF, was necessary). With LF content, `tools/development/gst-indent` (GNU indent 2.2.12) on the `.c` file and `clang-format -style=file` (matching `.clang-format`, ColumnLimit 80) on the `.cc` file both produce **zero diff** against CI's exact invocation in `.github/workflows/static.check.yml`. (Note: the test file already has many pre-existing lines over 80 columns — including some of this PR's own added lines — but clang-format's actual wrapping behavior does not rewrap them either way, so this is not a style violation in practice.)

## Other review criteria

- **Scope/size**: only `gst/nnstreamer/elements/gsttensor_srciio.c` and `tests/nnstreamer_source/unittest_src_iio.cc` are touched; every changed function in the `.c` file is `static` to that translation unit, so there is no cross-module regression surface. Diff size (roughly 40 net production lines across 4 independent hunks, plus proportionate tests) matches the four narrowly-scoped defects being fixed.
- **C27 byte-level correctness**: hand-verified the big-endian/little-endian shift-and-mask arithmetic for the `memcpy`-into-zeroed-container approach for 3-, 5-, 6-, 7-byte channels; the existing `storage_bits`-based shift/mask (unchanged) correctly cancels out the zero-padding byte(s) in both endiannesses, so the decoded value is provably unchanged from before — matches what `dataVerifyExactStorage24/56` plus the 10 pre-existing `dataVerifyNoTriggerBits*` cases (all still passing) confirm.
- **Future-regression detection**: the CI valgrind job runs `run_unittests_binaries.sh --valgrind --recursive ./tests/`, which discovers and runs `unittest_src_iio` automatically (gated only on `tensor_src_iio_build`, true on the CI's Ubuntu runner) — so a future change that reintroduces an out-of-bounds read here would fail CI via the memcheck gate, not just be silently missed. The two purely-logical defects (C28, C29) are additionally pinned by ordinary GTest assertions, so they fail CI even without valgrind.
- **Docs/architecture**: no public API, ABI, or property semantics changed in a way that any existing document describes (grepped `Documentation/` for `tensor_src_iio`/`buffer-capacity`/`CHANNELS`: no hits), so no documentation update is missing.

## Verdict

No blocking or non-blocking issues found in this round. The amend is verified safe (the removed guard was genuinely unreachable, confirmed both statically and empirically), and the rest of the diff holds up under independent re-review, including reproducing the original memory-safety defects on the pre-fix code and confirming they are actually fixed.

myungjoo:
Addressed the NIT from the `myungjoo-bot` review in `84bfd455` (the C28 commit, amended): `get_property ()` checks `custom_channel_table` for NULL again and reports an empty list, so the getter no longer depends on the single assignment at `:1238` staying the only way into the custom mode.

That check was in the first version of this PR and was dropped after the first relayed review called it unreachable. It is still unreachable today. Keeping it is the better choice, because the alternative for a future path that breaks the invariant is the crash C28 closed. It does not hide such a regression from the tests: `channelsCustomInvalid_n` expects the previous list (`"7"`) after each refused value, so an empty list fails it just as the crash did.

No other change. Local: `unittest_src_iio` 25/25, clang-format / gst-indent / doxygen-tag clean, strict `-Werror -Wextra -Wsign-compare` build OK.

myungjoo:
**Note:** This review was produced by a separate AI review agent (not the one that authored this PR, and not either of the two agents that posted the earlier relayed comments on this PR) and is being relayed here verbatim by the requester. Treat it as a third-party code review, not as a comment from the PR author.

This is a fresh, independent re-review of the current head (`84bfd455`), done without assuming any earlier round's conclusions are correct. It focuses on what changed since the last relayed review (the `myungjoo-bot` APPROVED review at `db4c9496`, and the subsequent amend of the C28 commit that restored the `get_property()` NULL guard), plus a from-scratch check of the whole diff.

## 1. Is the guard restoration a reasonable call, and is its shape right?

`myungjoo-bot`'s review asked, as a NIT, for `get_property()` to defend itself against `custom_channel_table == NULL` instead of relying only on the single assignment at `set_property()`'s CHANNELS case. The two earlier relayed rounds had gone the other way (flagged the same guard as dead code and had it removed). The amend restores it as:

```c
} else if (self->custom_channel_table == NULL) {
  g_value_set_string (value, "");
} else {
```

I traced every one of the 8 references to `custom_channel_table` in the file independently (init, both branches of the CHANNELS case in `set_property`, `get_property`, `setup_scan_channels`, `finalize`) and confirm the invariant `channels_enabled == CUSTOM ⟺ custom_channel_table != NULL` still holds by construction today, so the guard is still unreachable in practice. Given that, restoring it is the better of the two choices: the cost is one dead branch, the benefit is that a future edit which sets `CHANNELS_ENABLED_CUSTOM` from a second place no longer has to also remember to pair it with the table to avoid reopening C28 as a crash. It also does not weaken the regression test: `channelsCustomInvalid_n` asserts the *previous* value (`"7"`) survives every refused input, so if the invariant ever breaks, the getter now returns `""` instead of crashing, and the test still fails on the empty string, just as it would have failed on the crash. So this is a real safety improvement with no observed downside, and "return empty string" is a sane, non-throwing choice for a getter that cannot recover a meaningful value.

## 2. Did the amend disturb anything else?

`git range-diff a60a9ed9..db4c94966e197122ae2dc5421eb322d6c4fb59e2 a60a9ed9..84bfd455` shows the first three commits (158e1f0b, 9a548132, e2eb59df) are byte-identical between the previously-reviewed head and the current head. Only the fourth commit (C28) was amended, and only to add the guard back (plus a corresponding commit-message edit). No test file changes came with the amend — `channelsCustomInvalid_n` and the rest of `unittest_src_iio.cc` are unchanged from what `myungjoo-bot` already reviewed.

## 3. Has `main` drifted under this PR since the last round?

`origin/main` is now at `4d028546`, 17 commits ahead of this branch's merge-base (`a60a9ed9`). None of those 17 commits touch `gst/nnstreamer/elements/gsttensor_srciio.c` or `tests/nnstreamer_source/unittest_src_iio.cc` (checked with `git log <merge-base>..origin/main -- <both paths>`, empty). No merge conflict and no semantic-conflict surface with concurrent work on this element.

## 4. Independent, from-scratch verification (not reused from prior rounds)

Built the exact PR-head content (verified by `sha256sum` against the worktree, not by trusting the description) in a container (meson/ninja/gtest/valgrind, `-Werror` build):

- Fixed build: `unittest_src_iio` **25/25 pass**.
- Negative control: swapped only `gsttensor_srciio.c` for the pre-PR `main` version (`a60a9ed9`), kept the new tests, rebuilt: `bufferCapacityOverflow_n` and `channelsCustomInvalid_n` fail, the latter with **exactly 10 GLib criticals**, matching the PR's own claim; the C27/C30-covering tests (`dataVerifyExactStorage24/56`, `channelNameDigitsOnly`, `channelNameEmpty_n`) still pass under gtest against the unfixed element, as the PR states (they're valgrind-witnessed, not assertion-witnessed).
- Ran that same negative-control binary under `valgrind` limited to the four valgrind-witnessed cases and reproduced, independently: an **Invalid read of size 4** and **size 8** in `gst_tensor_src_iio_process_scanned_data()` (gsttensor_srciio.c:2416, 2439) and an **Invalid read of size 1** in `gst_tensor_src_iio_get_generic_name()` (gsttensor_srciio.c:810) — all three land inside the block the PR describes and disappear on the fixed build.
- Re-derived the C27 bit-arithmetic by hand rather than trusting the earlier reviews' derivation: for a `storage_bytes`-byte container zero-padded into a `guint32`/`guint64` and then run through `GUINT{32,64}_FROM_BE`, the extra padding is provably equivalent to shifting a container sized exactly to `storage_bytes` — `(X << (32 - 8*storage_bytes)) >> (32 - storage_bits) == X >> (8*storage_bytes - storage_bits)` — so the decoded value is bit-identical to the pre-fix output for every `storage_bits ≤ 8*storage_bytes`, on both endiannesses. This matches the PR's and the prior reviews' claim, arrived at independently.
- `clang-format --dry-run`/`-i` round-trip on `unittest_src_iio.cc`: no diff.
- Restored the container to the fixed PR-head source afterward (shared build environment).

## 5. Scope, regression risk, docs

Confirms the earlier rounds' findings: only 2 files touched, no header change, every touched function is `static` to `gsttensor_srciio.c`, `buffer_capacity` cannot change once a device is configured (pre-existing top-of-function state guard in `set_property()` already refuses writes outside NULL/READY, independently verified by reading that guard), and `setup_scan_channels()`'s new overflow check runs before `setup_device_buffer()` writes the capacity to the device, so a refused capacity never reaches hardware. No architecture or API change, so no documentation update is required or missing.

## Verdict

No blocking or major issues. The four defects (C27–C30) are each independently and correctly fixed, the C28-guard amend is a reasonable and low-risk resolution of the two earlier rounds' disagreement, nothing drifted or regressed in the interim, and the test/valgrind coverage is real (independently reproduced, not just read from the PR description). Approving.

myungjoo:
Approved by Astra

PR #4980, opened 2026-09-16
[Common] Size extra tensors by their memory and refuse a meta version this build cannot size

myungjoo:
- **What this fixes.** Two defects of the core library, items **A6** (filed as #4934) and **A7** of #4920. A6: `gst_tensor_buffer_append_memory ()` copies the incoming memory into the merged block whole but recorded the size of the *tensor* the caller described, not of the *memory* it copied; a memory that carries a meta header is 128 bytes longer, so `gst_tensor_buffer_get_nth_memory ()` walked the block one header short per extra tensor — the 17th tensor came back truncated, the 18th started inside the data of the 17th and no longer parsed, and the rest drifted further. Any stream of more than 16 tensors whose memories carry a header (flexible, sparse) through tensor_mux, tensor_merge, tensor_transform, tensor_sparse_dec, edge or tensor_query hits it. A7: a meta header of a version this build cannot size passed validation, and every consumer that locates the payload with `gst_tensor_meta_info_get_header_size ()` then read the header itself as tensor data, while a size check built on the same helper was short by the header.
- **How.** A6: the merged block has room for one `GstTensorInfo` per extra tensor and the walk uses nothing but its size, so the entry now records the bytes of the memory as they were copied (a uint8 byte vector). What the tensor holds still comes from the caps of the stream or from the meta header the tensor itself carries. A7: `gst_tensor_meta_info_validate ()` refuses a version whose header cannot be sized and `gst_tensor_meta_info_get_data_size ()` answers 0 for one; `GST_TENSOR_META_IS_V1 ()` compared the major version with a bitwise and (so every odd major passed as version 1) and now compares it for equality. The production diff is +31/-24 in two files, one commit per item, and the size a memory may have to be recorded is bounded before anything is allocated for it.
- **Result.** 8 new cases and 2 existing cases updated. Against `main`, 5 of the new cases and both updated ones fail; all pass with the fix. The full meson suite is 23/24 locally — the one failure, `unittest_filter_python3`, is a numpy 1.x/2.x mismatch of this box and fails on `main` the same way. A 3-mutation matrix pins every A7 guard to a case, `gst-indent`, `clang-format` 16 and the doxygen-tag checker (verified with a negative control) are clean on the changed files.

## Details

**A6 — how the sizes drift.** For each extra tensor the block keeps a `GstTensorInfo`, and `gst_tensor_buffer_get_nth_memory ()` advances by `gst_tensor_info_get_size ()` of it. `append_memory ()` decided what to store from the *format field* of the parsed header: static stored the caller's info, flexible stored the info converted from the header. Both describe the tensor, not the memory, and a memory that carries a header is `128 + data`. Note that a tensor of a flexible stream usually declares `format=static` in its own header, so most streams took the first branch — the key is whether a header is present, not what it says.

**A7 — what it unblocks.** `gst_tensor_meta_info_validate ()` only required `(version & 0xDE000000) == 0xDE000000`. `tensor_decoder` (#4939) and `tensor_filter` (#4953) each had to refuse such a header on their own; those two cases stay, and now observe the refusal one layer earlier. `gst_tensor_sparse_to_dense ()` already bounded the payload itself, and its case is unchanged.

**What changes for a user.** Nothing that exists is refused: version 1.0 is the only version this library has ever written, and any minor of major 1 is still accepted (pinned by a case). A buffer of more than 16 flexible or sparse tensors now reads back correctly where it used to be truncated. The extra-tensor block never leaves the process — `edge_sink`, `tensor_query_{server,client}` and `datareposink` all take the tensors apart with `gst_tensor_buffer_get_nth_memory ()` before sending or writing them — so no on-the-wire or on-disk format changes.

## For reviewers

**Commits.** 1. A6 (`nnstreamer_plugin_api_impl.c`), including the bound on the size a memory may have to be recorded. 2. A7 (`nnstreamer_plugin_api_util_impl.c`) together with the two existing cases whose intermediate assertion it changes, so each commit stands green. 3. The new cases, including the oversized-memory case.

**Test evidence** (`main` → this PR):

| case | on main | here |
| --- | --- | --- |
| `commonUtil.tensorBufferExtraFlexTensors` (new) | FAIL, the 17th tensor has no header | pass |
| `testTensorTransform.typecastFlexibleExtraTensors` (new) | FAIL, `the tensor 16 has no meta header` | pass |
| `testTensorSparse.decExtraTensorsShorterThanDense` (new) | FAIL | pass |
| `commonMetaInfo.validateUnknownVersion_n` (new) | FAIL (also catches the `IS_V1` and data-size guards) | pass |
| `commonMetaInfo.parseHeaderUnknownVersion_n` (new) | FAIL | pass |
| `commonUtil.tensorBufferExtraFlexForgedSize_n`, `commonMetaInfo.validateKnownVersion` (new) | pass | pass |
| `commonUtil.tensorBufferAppendOversizedMemory_n` (new) | n/a, guard added here | pass; SIGSEGV without the guard |
| `testTensorDecoder.pushFlexibleInputUnknownVersion_n`, `tensorFilterFlexInput.unknownVersion_n` (updated) | assert the old behaviour | pass |

**Two notes.**
- Three guards against a header whose size is 0 are now unreachable, because A7 makes `parse_header ()` refuse that header first: `gst_tensor_buffer_from_config ()` (#4937), `gst_tensordec_check_input_size ()` (#4939) and `_gst_tensor_filter_convert_meta ()` (#4953). They are kept as they are: removing them would take this PR into two element files, and they stay the safety net if `validate ()` and `get_header_size ()` ever diverge again. The element cases still assert that the elements refuse such a header.
- The indent step of `Static checks` warns with a file-wide diff (it runs `indent` without the `-T` type list of `tools/development/gst-indent`, so every `Type * name` is reported). The step does not fail, and none of the lines in its diff come from this PR.

Related: #4920 (items A6, A7), #4934.


myungjoo:
This review was produced by a separate Claude review agent (not the agent that authored this PR) and is being posted here as a relay of that agent's findings.

## Summary

I traced the two fixes against the actual read paths and every downstream consumer, and cross-checked the new tests against the code they exercise. The PR does what it claims, the fix is correct, and the scope is appropriately minimal. I did not find any regression risk. Details below.

## A6 — extra-tensor size drift (`nnstreamer_plugin_api_impl.c`)

Confirmed the root cause: `gst_tensor_buffer_get_nth_memory()` has always advanced through the extra-tensor block using `gst_tensor_info_get_size()` on the stored `GstTensorInfo` — i.e. it always treated the stored entry as a byte count of the *memory*. The old `append_memory()` instead stored the *tensor's* logical info (dimensions/type of the payload, or the meta-converted info for a flexible tensor), which under-counts by the 128-byte header whenever the appended memory carries one. Read and write sides were describing two different things with the same field, and only the write side needed to change to match what the read side already assumed.

The fix (`gst_tensor_info_init` + `type = _NNS_UINT8` + `dimension[0] = (uint32_t) incoming_memory_map.size`) is correct: `gst_tensor_get_element_count()` stops at the first zero dimension, so a rank-effectively-1 UINT8 vector of `dimension[0]` bytes yields exactly `incoming_memory_map.size` from `gst_tensor_info_get_size()`, which is exactly what the reader needs.

I checked whether `GstTensorExtraInfo`/`.infos[]` is used anywhere else in the tree (`grep` across `gst/` and `ext/`) — it is private to `nnstreamer_plugin_api_impl.c` and not declared in any header, and the only two use sites of `.infos[i]` are the write in `append_memory()` and the read in `get_nth_memory()`, both now consistent. So this is a genuinely internal, non-serialized in-process data structure; the PR's claim that "no on-the-wire or on-disk format changes" is correct — I verified `edge_sink.c`, `tensor_query_client.c`/`tensor_query_server.c`, and `gstdatareposink.c` all reconstruct individual memories via `gst_tensor_buffer_get_nth_memory()` before doing anything with them, rather than serializing the extra-info block itself.

The new `G_MAXUINT32` guard on `incoming_memory_map.size` is a correct defensive bound given `dimension[0]` is 32-bit; I traced the failure path (`goto failed` after unmapping `memory`) and it does not double-unmap or leak — `new_memory` is discarded and `last_memory`/`memory` are cleaned up exactly once via the existing `failed:` label, matching the pattern of the function's other early-exit branches. This path is understandably untested (would need a >4GB single memory block), and the PR states that plainly rather than papering over it.

## A7 — unknown meta version (`nnstreamer_plugin_api_util_impl.c`)

Confirmed both defects independently:

1. `GST_TENSOR_META_IS_V1`'s old body ANDed `(v & 0x00FFF000)` with the *unmasked* `GST_TENSOR_META_MAKE_VERSION(1,0)` value instead of comparing equality. Working through the bit pattern: `MAKE_VERSION(1,0) = 0xDE001000`, so the AND degenerates to testing `v & 0x00001000` — i.e. only bit 12 (the LSB of the encoded major version). Any major version with that bit set (1, 3, 5, …) passed as "V1" while the intent was major-version equality. The new `==` comparison against `GST_TENSOR_META_MAKE_VERSION(1,0) & 0x00FFF000` is the correct fix, and I confirmed the only version this codebase ever writes is `GST_TENSOR_META_VERSION = MAKE_VERSION(1,0)`, so there is no legitimate producer whose version this changes behavior for.
2. Independently, `gst_tensor_meta_info_validate()` previously only checked `GST_TENSOR_META_IS_VALID` (magic + the `0xDE000000` marker byte), not whether the version's header size is knowable. So a meta with an unknown major (even one the buggy macro above correctly rejected as V1, e.g. major=2) still passed `validate()` as TRUE while `get_header_size()` for it returned 0 — i.e. validated metadata whose payload location is undefined. `gst_tensor_meta_info_update_header()` calls `validate()` and then `memset(header, 0, hsize)` / `memcpy(header, meta, sizeof(GstTensorMetaInfo))` with `hsize` possibly 0 against a caller-sized-by-`get_header_size()` buffer — this is a real bug the new check closes, not just tightening for its own sake.

I confirmed via `grep` that every other call site (`tensor_filter.c`, `gsttensor_decoder.c`, `gsttensor_crop.c`, `gsttensor_transform.c`, `gsttensor_sparseutil.c`, `gstdatareposrc.c`, `tensor_converter_flexbuf.cc`, `tensordec-octetstream.c`, …) locates its payload via `gst_tensor_meta_info_get_header_size()`, which already returns 0 correctly once the `IS_V1` macro is fixed — so the fix's benefit reaches all of them without needing changes elsewhere, and I did not find a call site that bypasses `validate()`/`get_header_size()` and would remain exposed.

No behavior changes for any version this library has ever produced (only major 1 is ever emitted; the two commits touching `#4939`'s and `#4953`'s existing test cases update assertions of behavior that was only reachable with a deliberately forged header, and both are noted honestly in the PR body as "the refusal one layer earlier" rather than new behavior).

## Scope

Production diff is exactly what the PR claims: two files, both directly owning the affected logic (`nnstreamer_plugin_api_impl.c` owns `append_memory`/`get_nth_memory`; `nnstreamer_plugin_api_util_impl.c` owns the meta-version macros and `validate`/`get_header_size`/`get_data_size`). No unrelated module is touched. Neither change is a public API or ABI change: `GstTensorExtraInfo` is private (not in any header), and every touched public function (`gst_tensor_buffer_append_memory`, `gst_tensor_meta_info_validate`, `gst_tensor_meta_info_get_header_size`, `gst_tensor_meta_info_get_data_size`) keeps its existing signature and contract. I did not find any documentation under `Documentation/` that describes these internals, so no doc update is owed here.

## Tests

The new/updated GTest cases are well targeted and I checked each one traces back to the actual bug:
- `commonUtil.tensorBufferExtraFlexTensors` and `testTensorTransform.typecastFlexibleExtraTensors` build >16-tensor flexible streams through the real `append_memory()`/`get_nth_memory()` path and verify both content and per-tensor header re-parseability — this is the direct regression test for A6 and would have failed on `main` for the reason the PR states (17th tensor missing its header).
- `testTensorSparse.decExtraTensorsShorterThanDense` generalizes the existing `_sparse_new_extra_memory()` helper with a `stride` parameter so a sparse-encoded extra tensor is *shorter* than the dense tensor it represents — a case the previous fixed-stride-10 test could not exercise, and specifically hits the "size drift" direction of the A6 bug (as opposed to the pre-existing `decExtraTensors` case where the encoded size happened to equal the dense size).
- `commonUtil.tensorBufferExtraFlexForgedSize_n` forges an extra-tensor's recorded size to `G_MAXUINT32` and checks that `get_nth_memory()` refuses to read out-of-bounds for both the forged tensor and the one after it — I traced the bounds-check loop in `get_nth_memory()` by hand for both index cases and confirmed the test actually exercises the `goto invalid` path both when the forged entry is the target and when it's a predecessor being skipped over.
- `commonMetaInfo.validateUnknownVersion_n` / `validateKnownVersion` / `parseHeaderUnknownVersion_n` cover A7 from three angles (an even unknown major that the old `IS_V1` macro already rejected but old `validate()` didn't; an odd unknown major that only the macro fix rejects; and an update→parse round-trip). This is good coverage of the specific bit-manipulation bug, not just a black-box check.
- The two updated pre-existing tests (`tensorFilterFlexInput.unknownVersion_n`, `testTensorDecoder.pushFlexibleInputUnknownVersion_n`) are appropriately flipped from asserting the old (buggy) behavior to the new one, with comments explaining why, rather than being silently deleted.

All new tests live in already-`meson.build`-registered test binaries (`unittest_common`, `unittest_plugins`, `unittest_filter_custom`), so no build-file changes were needed and the standard `meson test` CI job will run them and gate on them without further wiring — a future regression in either `append_memory`/`get_nth_memory` sizing or in the version macros would be caught by CI, not just locally.

One coverage gap worth naming (non-blocking): there's no test for the new `G_MAXUINT32` overflow guard in `append_memory()` (understandably impractical — it needs a single >4GB memory), and no test that a *static* (non-flexible) extra tensor larger than `NNS_TENSOR_MEMORY_MAX` still round-trips post-fix (only flexible/sparse extra tensors are exercised for A6, though the code path for static tensors funnels through the same new size-recording logic and is lower-risk since static tensors carry no header to be off by).

## Verdict

No blocking issues found. Fix is correct, minimal, and matches the stated root causes for #4920 items A6/A7 and #4934. Test coverage is strong and CI-enforced. No architecture/API/documentation impact.

myungjoo:
Both non-blocking points of the review are addressed in `95390731`.

**The `G_MAXUINT32` guard had no case.** It stood where the entry is written, i.e. after the merged block had already been allocated, so reaching it required an allocation of that same size to succeed first — which is why it looked untestable. It now stands where the append path learns that the memory becomes an extra tensor, before anything is allocated or mapped for it. `commonUtil.tensorBufferAppendOversizedMemory_n` then appends a memory claiming `G_MAXUINT32 + 1` bytes over a 256-byte allocation (the case skips itself where `gsize` is 32 bits). Removing the guard turns that case into a SIGSEGV (exit 139) on the copy; with it the case passes, so the guard is pinned.

**Static extra tensors above 16 are covered**, by cases that predate this PR and run over the changed write path: `commonUtil.tensorBufferExtraTensors` builds and reads back `NNS_TENSOR_SIZE_LIMIT` (256) static tensors and checks the content of every one, `tensorBufferAppendOverLimit_n` checks the buffer is intact after a refused append, and `testTensorTransform.arithmeticExtraTensors` runs 18 static tensors through an element. They all pass here. What this PR changes for a static tensor is only where its size in the header comes from — the caller's info before, the memory that was copied now — and those two agree for every caller in the tree.

Local state after the follow-up: full meson suite 23/24, the one failure being `unittest_filter_python3`, a numpy 1.x/2.x mismatch of this box that fails the same way on `main`. `gst-indent`, `clang-format` 16 and the doxygen-tag checker are clean on the changed files.

myungjoo:
This review was produced by a separate Claude review agent (not the agent that authored this PR) and is being posted here as a relay of that agent's findings. Scope: the follow-up commit `9539073` added since the [previous review](https://github.com/nnstreamer/nnstreamer/pull/4980#issuecomment-5692920428) — moving the `G_MAXUINT32` size guard in `gst_tensor_buffer_append_memory()` earlier, plus the new `commonUtil.tensorBufferAppendOversizedMemory_n` case.

## What changed

Before this commit, the bound on the incoming memory's size (`incoming_memory_map.size > G_MAXUINT32`) was checked *after* `last_memory` was mapped, the merged block (`new_mem_size`, already including the oversized incoming size) was allocated, `new_memory`/`incoming_memory` were both mapped, and `last_memory`'s bytes were already copied into the new block. The commit hoists the same check — same comparison, same threshold — to right after `incoming_mem_size` is read via `gst_memory_get_sizes()`, before `last_memory` is even looked up, i.e. before any allocation, mapping, or copying happens for this call.

## Verification of the two things asked for

**1. Does the reordering change behavior?** No observable difference. I traced every branch:
- The guard's logic (`incoming_mem_size > G_MAXUINT32`) and its outcome (log + `goto failed`) are byte-for-byte identical, just relocated and fed from a cached `incoming_mem_size` instead of re-deriving it from `incoming_memory_map.size` later (which was itself always equal to `gst_memory_get_sizes(memory, ...)`, so no semantic change there either).
- The `failed:` label's cleanup already tolerates being reached this early: at this point `new_memory == NULL` and `last_memory == NULL` (their declarations initialize both to NULL and neither has been assigned yet), so the `if (new_memory)`/`if (last_memory)` cleanup blocks correctly no-op, and `if (memory) gst_memory_unref(memory)` still releases the caller's memory exactly once — matching the function's documented contract ("takes ownership of this, even if it returns failure").
- The only order-of-checks change I found: if a caller simultaneously (a) hits `NNS_TENSOR_SIZE_EXTRA_LIMIT` (buffer already has the max number of extra tensors) and (b) passes an oversized memory, the function now logs "too large to describe in the extra header" first instead of "maximum number of tensors" — both paths still return `FALSE` and leave the buffer untouched either way, so this is a log-message ordering nuance, not a functional difference.
- The trivial-append path (`num_mems < NNS_TENSOR_MEMORY_MAX`) is unaffected — the guard only applies once the function is already committed to building/extending the merged "extra" block, which is the only place `dimension[0]` (32-bit) is used to record a memory's byte size.

**2. Does moving the guard miss any path?** No path is skipped. The guard sits at the very top of the "buffer already has `NNS_TENSOR_MEMORY_MAX` blocks" branch, strictly before every subsequent operation in that branch (`gst_buffer_peek_memory`, `gst_memory_map` of `last_memory`, the `is_extra`/`NNS_TENSOR_SIZE_EXTRA_LIMIT` check, allocation, and the two other `gst_memory_map` calls). Since it dominates all of those, every route that used to reach the old check location still passes through the new one first — there's no branch between the new check's position and the old one that could bypass it.

## Test case

`commonUtil.tensorBufferAppendOversizedMemory_n` directly answers the gap the prior review flagged as non-blocking ("no test for the new `G_MAXUINT32` overflow guard... understandably impractical — it needs a single >4GB memory"). Instead of allocating >4GB, it wraps a real 256-byte allocation in a `GstMemory` whose claimed `size`/`maxsize` is `G_MAXUINT32 + 1` (`gst_memory_new_wrapped`), which is exactly what the guard is supposed to catch before anything dereferences that claimed size. I confirmed:
- `EXPECT_FALSE (gst_tensor_buffer_append_memory (...))` and the follow-up `EXPECT_EQ (gst_tensor_buffer_get_count (buf), NNS_TENSOR_MEMORY_MAX)` together check both the return value and that the buffer is left untouched (no partial mutation), which is only meaningful because the guard now fires before any allocation/copy touches the buffer's state.
- The `if (sizeof (gsize) <= sizeof (guint32)) GTEST_SKIP();` guard is correct and necessary: on a 32-bit target, `(gsize) G_MAXUINT32 + 1` wraps to 0, which would silently turn this into a no-op "append a 0-byte memory" case instead of exercising the guard — skipping avoids a false pass.
- The memory is correctly released on the failure path: `gst_tensor_buffer_append_memory()` unrefs `memory` even on failure, which invokes `g_free` on the real 256-byte `data` pointer (freeing by pointer is independent of the claimed size), so there's no leak or double-free in the test.
- The test is registered in `tests/common/unittest_common.cc`, already wired into `meson.build`/`meson test`, and carries a `@brief` doxygen comment consistent with the file's existing style — so it runs under normal CI without extra wiring and satisfies the per-function doxygen requirement.

## Conclusion

No issues found in this follow-up commit. It is a minimal, behavior-preserving reordering (same check, same outcome, now reachable and cheap to test) plus a well-targeted regression test that closes the specific coverage gap named in the prior review round, with an appropriate platform-portability guard (32-bit skip) that a less careful test would have missed.

myungjoo:
This review was produced by a separate Claude review agent (not the agent that authored this PR, and not the earlier reviewers already on this thread) and is being posted here as a relay of that agent's findings. I re-derived the analysis independently from the diff, the surrounding source, and the callers in the tree, rather than relying on the conclusions of the earlier reviews already posted on this PR.

**Reviewed head:** `95390731` (4 commits on `main`@`8f9eeca8`).

## Scope check

The production diff is exactly two files (`gst/nnstreamer/nnstreamer_plugin_api_impl.c`, `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`), both of which directly own the two defects being fixed (A6: `gst_tensor_buffer_append_memory`/`gst_tensor_buffer_get_nth_memory`; A7: the meta-version macros and `validate`/`get_header_size`/`get_data_size`). No unrelated module is touched, and both fixed functions keep their existing public signature/contract, so this is not an API/ABI change and no architecture documentation is owed.

## Correctness (independently traced)

**A6.** `GstTensorExtraInfo.infos[]` is private to `nnstreamer_plugin_api_impl.c` (not declared in any header) and has exactly two use sites: the write in `append_memory()` and the read-side walk in `get_nth_memory()` (`nnstreamer_plugin_api_impl.c:1786`), which uses only `gst_tensor_info_get_size()` on the stored entry to advance an offset. Recording `type = _NNS_UINT8, dimension[0] = (uint32_t) incoming_mem_size` is exactly right for that: `gst_tensor_get_element_count()` stops at the first zero dimension (`dimension[1..]` stay 0 from `gst_tensor_info_init()`), so `gst_tensor_info_get_size()` returns exactly `incoming_mem_size`. I grepped `gst/` and `ext/` for any other reader of `.infos[]`'s type/dimension as semantic tensor info and found none — the only public accessor is `gst_tensor_buffer_get_nth_memory()`, which hands back raw bytes; callers recover the real type either from the stream's negotiated caps (static) or by re-parsing the memory's own embedded meta header (flexible/sparse), never from this bookkeeping struct. So the fix does not change what any consumer (tensor_mux/merge/transform, tensor_sparse_dec, edge, tensor_query, datareposink) observes for a tensor that round-trips correctly; it only fixes the previously wrong size recorded whenever the appended memory carried a header, which is the actual root cause described in the PR body (header presence, not the header's own `format` field, is what mattered — confirmed by reading the code path that used to be a strict function of `meta.format == static`).

**A7.** Worked the bit pattern by hand: `GST_TENSOR_META_MAKE_VERSION(1,0) = 0xDE001000`. The old `IS_V1` body `((v & 0x00FFF000) & 0xDE001000)` degenerates to testing bit 12 alone (`v & 0x00001000`), so any odd major (1, 3, 5, …) passed as "V1" — matches the PR's description. The new `== (0xDE001000 & 0x00FFF000)` correctly requires major-version equality while leaving the minor free, and `gst_tensor_meta_info_init()` only ever writes `MAKE_VERSION(1,0)`, so nothing this library produces changes behavior. The added `get_header_size(meta) == 0` guard in `validate()` closes a real, independent gap: before this PR, a meta with an unknown major could still pass `GST_TENSOR_META_IS_VALID` (magic + `0xDE000000` marker only) while `get_header_size()` for it already returned 0 — i.e. metadata could validate successfully while its payload location was undefined. I checked every other call site that locates a payload via `gst_tensor_meta_info_get_header_size()` (`tensor_filter.c`, `gsttensor_decoder.c`, `gsttensor_crop.c`, `gsttensor_transform.c`, `gsttensor_sparseutil.c`, `gstdatareposrc.c`) and none of them duplicate the version-bit logic themselves, so the fix's benefit reaches all of them through the shared helper without further changes.

No regression risk found in either fix; both are narrowly scoped to the exact bug described.

## Tests

The new/updated cases are well targeted and already wired into existing, CI-gated binaries (`unittest_common`, `unittest_plugins`, `unittest_filter_custom` are all present in `tests/meson.build`, run by the standard `meson test` CI job — no build-file changes were needed, so a future regression in either area is caught by CI automatically, not just locally):
- `commonUtil.tensorBufferExtraFlexTensors` / `testTensorTransform.typecastFlexibleExtraTensors` exercise the real >16-tensor flexible path end to end (append → read back → re-parse header) — this is the direct regression test for A6.
- `testTensorSparse.decExtraTensorsShorterThanDense` generalizes the existing sparse helper with a `stride` parameter to produce an encoded tensor *shorter* than the dense tensor it represents, which is the specific direction of the A6 size-drift bug that the pre-existing `decExtraTensors` case (stride fixed at 10, sizes coincidentally equal) could not catch.
- `commonMetaInfo.validateUnknownVersion_n` / `validateKnownVersion` / `parseHeaderUnknownVersion_n` cover A7 from the angles that matter for the bit-manipulation bug: an even unknown major, an odd unknown major (the one the old macro let through), and an update→parse round trip.
- `commonUtil.tensorBufferAppendOversizedMemory_n` correctly wraps a small real allocation with a claimed `G_MAXUINT32 + 1` size (rather than needing an actual >4GB block) and self-skips on a 32-bit `gsize` target, which is the right way to make an otherwise-impractical bound testable.
- Every new function/test carries a `@brief` Doxygen comment, satisfying the repo's per-function documentation rule.

I did not find a gap in the new coverage that isn't already explicitly and honestly called out in the PR body itself (e.g. static extra-tensor round-trip, which the author points out is covered by pre-existing cases exercising the same changed write path).

## One open point carried over, still unresolved as of this head

The `myungjoo-bot`-relayed review already on this thread requested a change (commit-history hygiene): commit 4 (`95390731`, "Bound an appended memory...") relocates a check that commit 1 (`4fb76aa5`) had just introduced in a place its own message calls "expensive and unreachable for a test," rather than folding the correction back into commit 1. The PR body's "For reviewers > Commits" section also still lists only 3 commits ("1. A6 ... 2. A7 ... 3. The new cases"), not the 4 that are actually on the branch. Re-checking the current head (`95390731`), this has not been squashed/updated yet, so that "changes requested" item still stands independent of anything I found. This is a repo-hygiene/bisectability concern, not a functional defect — I found no correctness or regression issue in commit 4's actual content (traced the `failed:` cleanup path and confirmed no double-unmap/leak/behavior change from moving the check earlier).

## Verdict

No new functional, regression, or scope issues found beyond what is already tracked. The fix correctly resolves #4920 items A6/A7 (#4934), is minimally scoped, does not touch or risk other modules, needs no documentation update, and is backed by tests that are already CI-gated. The one outstanding item is the previously-requested commit squash / PR-body commit-list update, which remains open on this head.

myungjoo:
Addressing the review of `95390731` (changes requested).

**1. Commit 4 was a fix-up of commit 1 — fixed.** The history is rewritten to three commits, and the tree of the new head `5b928749` is identical to `95390731` (`git diff 95390731 5b928749` is empty):

1. `fc4bbdcd` [Common] Size an extra tensor by the memory that was appended — the `G_MAXUINT32` bound now stands before the allocation from the start, and the commit message says why it is there.
2. `08ddda95` [Common] Refuse a tensor meta header this build cannot size — unchanged.
3. `5b928749` [Test/Common] Read back a stream of more than 16 tensors that carry headers — now includes `tensorBufferAppendOversizedMemory_n`.

The "Commits" section of the PR body is updated to match.

**2. NIT: guards made unreachable by A7 — kept, and now listed in the PR body.** There are three of them, not two: `gst_tensor_buffer_from_config ()` (#4937), `gst_tensordec_check_input_size ()` (#4939) and `_gst_tensor_filter_convert_meta ()` (#4953, the `*hsize == 0` condition). I am leaving them in place:
- Removing them would take this PR into `gsttensor_decoder.c` and `tensor_filter.c`, which it otherwise does not touch.
- They cost nothing on the data path and remain the safety net if `validate ()` and `get_header_size ()` ever diverge again. That divergence is exactly how A7 happened.
- What their tests check still holds: `pushFlexibleInputUnknownVersion_n` and `tensorFilterFlexInput.unknownVersion_n` assert that the elements refuse such a header, now one layer earlier.

The line coverage of those three branches drops as you note; there is no behavioural change.

myungjoo:
This review was produced by a separate Claude review agent (not the agent that authored this PR, and not any of the earlier reviewers already on this thread) and is being posted here as a relay of that agent's findings.

**Scope of this pass.** The head moved from `95390731` (4 commits, CHANGES_REQUESTED by `myungjoo-bot` for finding 1: commit 4 is a fix-up of commit 1, and finding 2 NIT: a guard made unreachable by A7) to `5b928749` (3 commits: `fc4bbdcd`, `08ddda95`, `5b928749`), addressed in https://github.com/nnstreamer/nnstreamer/pull/4980#issuecomment-5706599032. I re-verified the whole change independently rather than trusting that reply, with emphasis on (a) whether the 3 commits each stand on their own, (b) whether commit messages match their diffs, (c) whether finding 2's rationale for leaving code as-is is actually correct, and (d) whether the PR body matches the current head.

## (a)/(b) Per-commit soundness, checked with `git show`

- `git diff 95390731 5b928749` is **empty** — the rewritten history produces the exact same tree as the previously-reviewed head, so no new production or test content was introduced by the rebase; only the commit split changed.
- **`fc4bbdcd`** — touches only `nnstreamer_plugin_api_impl.c`. It removes `is_static`/the `meta.format`-based branch and records `type = _NNS_UINT8, dimension[0] = incoming_mem_size` for every extra-tensor entry, and places the `incoming_mem_size > G_MAXUINT32` guard right after `gst_memory_get_sizes()`, before `last_memory` is even looked up — i.e. before any allocation or mapping. This is exactly commit 1 + the old commit 4 folded together, matching the message ("...That is refused where the append path learns the memory becomes an extra tensor, before the merged block is allocated..."). Self-contained: no dependency on the later commits, no dangling reference to a test that doesn't exist yet.
- **`08ddda95`** — touches `nnstreamer_plugin_api_util_impl.c` plus the two *existing* test files whose asserted behavior flips (`unittest_filter_custom.cc`, `unittest_plugins.cc`). I traced `gst_tensor_meta_info_parse_header()` → it returns `gst_tensor_meta_info_validate(meta)`, so the new `get_header_size(meta) == 0` check inside `validate()` is exactly why `tensorFilterFlexInput.unknownVersion_n` and `testTensorDecoder.pushFlexibleInputUnknownVersion_n` correctly flip from `EXPECT_TRUE` to `EXPECT_FALSE` in the *same* commit — this commit does not leave any test red or stale. Self-contained.
- **`5b928749`** — test-only, two files (`tests/common/unittest_common.cc`, `tests/nnstreamer_plugins/unittest_plugins.cc`), no production code. `commonUtil.tensorBufferAppendOversizedMemory_n` reuses the pre-existing `build_extra_tensors_buffer()` helper (already present before this PR at `unittest_common.cc:3210`), so it isn't accidentally depending on something only this commit adds. `testTensorSparse.decExtraTensorsShorterThanDense` reuses `_sparse_new_extra_memory()`, generalized with a `stride` parameter in this same commit, and the pre-existing `decExtraTensors` case is updated to pass its previous fixed stride (`10U`) explicitly — so that case keeps compiling and passing. No test in this commit depends on a fix not yet present as of `08ddda95`.

Each of the 3 commits is internally consistent and does not require a later commit to compile or to keep existing tests green — the structural issue from finding 1 (a fix-up commit split out from what it fixes) is resolved, not just reworded.

## (c) Finding 2 (unreachable defensive branches) — verified independently

I read all three call sites the author lists and confirmed each `hsize == 0` (or equivalent) check is indeed now dead code, because `gst_tensor_meta_info_parse_header()` itself refuses an unknown version first:

- `gst_tensor_buffer_from_config()` (`nnstreamer_plugin_api_impl.c`): the loop's condition is `... || !gst_tensor_meta_info_parse_header(&meta, ...)`, which now short-circuits to the "unknown version" fallback branch before the code ever reaches the later `if (mem_size[num] == 0)` check a few lines down.
- `gst_tensordec_check_input_size()` (`gsttensor_decoder.c`): `if (size < get_header_size(&meta) || !parse_header(&meta, data)) return FALSE;` runs first; the `if (hsize == 0) return FALSE;` right after it can now only be reached with `hsize` already non-zero.
- `_gst_tensor_filter_convert_meta()` (`tensor_filter.c`): same shape — `!gst_tensor_meta_info_parse_header(_meta, map->data)` returns first, so the subsequent `*hsize == 0` half of the following `if` is unreachable through that path.

So the technical claim is correct, and the author's judgment — keep them as a safety net rather than pull two more element files into this PR's scope, since `pushFlexibleInputUnknownVersion_n` / `tensorFilterFlexInput.unknownVersion_n` still assert the element-level refusal (now one layer later) — is reasonable. It is genuinely a coverage/dead-branch NIT with no behavioral or regression risk, not something that should block merge.

## (d) PR body vs. current head

The body's "For reviewers > Commits" section now lists exactly 3 commits matching `fc4bbdcd`/`08ddda95`/`5b928749`, and the "Two notes" section correctly says "three guards" (not two, as the pre-fix body said). `gh pr view` confirms the PR's commit list and `headRefOid` match this local worktree's `git log` exactly. I did not find any other stale claim in the body against the current diff.

## Other checks

- Scope stays exactly two production files (`nnstreamer_plugin_api_impl.c`, `nnstreamer_plugin_api_util_impl.c`), both directly owning the two defects; no unrelated module touched.
- `GstTensorExtraInfo` is confirmed private to `nnstreamer_plugin_api_impl.c` (not in any header) — no public API/ABI change, no architecture doc update owed.
- New test binaries touched (`unittest_common`, `unittest_plugins`, `unittest_filter_custom`) are all registered in `tests/meson.build` and run under `meson test`, so this is CI-gated, not just locally verified.

## Verdict

No blocking issues found in this round. Finding 1 is properly resolved (verified via the empty tree diff across the rewrite, plus per-commit inspection). Finding 2's technical basis is confirmed accurate and the decision to leave it as a NIT is sound. The PR body is up to date with the current 3-commit head.

myungjoo:
Approved by Astra

PR #4982, opened 2026-09-16
[Decoder/BoundingBox] Give every decoder box properties of its own

myungjoo:
- **What:** item **G10** of #4920. Every `tensor_decoder mode=bounding_boxes` element that picked the same `option1` shared one process-wide `BoxProperties` object, so a second decoder's `option5`/`option2`/`option3` overwrote the first one's. With two yolov8 decoders of different model sizes, the first read boxes past its own tensor (OOB read); a second mobilenet-ssd(-pp) model with another detection count was refused; a decoder missing `option5` silently used another decoder's.
- **How:** the registry now maps each mode name to a *creator function* instead of an instance. A `BoundingBox` creates the `BoxProperties` of a mode the first time `option1` selects it, and keeps one per selected mode until it exits. Nothing is freed on a runtime `option1` switch while `decode()` may still use it. The change stays inside the bounding_boxes sub-plugin, with no exported API, and the header is not installed.
- **Result:** 7 new GTest cases + 1 SSAT case (two yolov8 decoders in one pipeline, compared with the existing golden). Locally, 37/37 GTest and 65/65 SSAT pass, and ASan/LSan finds no errors or leaks of ours. On main, the new SSAT case and 5 of the unit cases fail (the yolov8 one SIGSEGVs on a guard page). `meson test`: 23/24; the failure is `unittest_filter_python3`, a local NumPy 2.x environment problem, not this change.

## Behavior notes for reviewers

- **A single decoder behaves as before.** Re-selecting a mode, including via the deprecated aliases `tf-ssd`/`tflite-ssd`, or switching back to a mode it used earlier finds that mode's options in place. `sameModeKeepsOptions` and `switchModeBackKeepsOptions` pin this.
- **A mode selected for the first time starts from its defaults.** It never sees another decoder's options or another mode's; `optionsOfAnotherDecoder_n` and `switchModeTakesNoOptions_n` pin this.

---

## Context for agents

**Root cause.** `init_properties_*()` in each `box_properties/*.cc` is `__attribute__((constructor))`. It did `new X ()` and `BoundingBox::addProperties (obj)`, and `BoundingBox::setBoxDecodingMode()` stored the table's pointer in `bdata`. `setLabelPath`/`setInputModelSize`/`setOptionInternal` and `checkCompatible` (which caches `max_detection`) all write into it.

**Files.**
- `tensordec-boundingbox.{h,cc}`: `typedef BoxProperties *(*BoxPropertiesCreator)(void)`; `getProperties` → `createProperties`; `addProperties(name, creator)`. Each `BoundingBox` holds a `GHashTable properties` (mode name → owned `BoxProperties`, value-destroy `delete`), and `bdata` points into it. `~BoundingBox` destroys the table.
- The five `box_properties/*.cc` files: `create_properties_*()` + registration; the `fini_properties_*` destructors are removed because nothing global is left to free.
- The table stays a lazily created `GHashTable` under `G_LOCK`, because constructor-attribute functions can run before C++ static initialisers.

**Why a per-decoder cache rather than delete-on-switch.** The first revision deleted the old `bdata` on a mode switch. The review (Low, item 2) pointed out that `tensor_decoder` does not serialise `set_property` with `decode()` (#4920 C38), so that opened a new use-after-free window. Keeping at most one instance per mode (10 modes) closes it, and it also restores the old single-decoder semantics of switching back. `createEveryMode` selects every name in fresh decoders, then every name twice in one decoder (leak-checked with LSan).

**Tests** (`tests/nnstreamer_decoder_boundingbox/`):
- New: `yoloV8DecodersOfDifferentModelSizes` (guard-page tensor), `ssdPpDecodersOfDifferentDetectionCounts`, `optionsOfAnotherDecoder_n`, `sameModeKeepsOptions`, `switchModeBackKeepsOptions`, `switchModeTakesNoOptions_n`, `createEveryMode`.
- `palmWithoutOption_n` no longer needs `EXPECT_EXIT` to get a process without anchors; it now proves a second decoder does not see the first one's anchors.
- Removed comments that described the per-process latch ("the mode keeps the first count it accepts for the whole process").
- SSAT `8-2`: yolov8 320:320 (golden) next to yolov8 160:160 fed the same raw file as 4 × 525-box frames.
- Negative control (main sources + new tests):
  - SIGSEGV: `yoloV8…`, `optionsOfAnotherDecoder_n`
  - FAIL: `ssdPp…`, `palmWithoutOption_n`, SSAT `8-2` (both checks)
  - Pass either way (guards for the new design): `sameModeKeepsOptions`, `switchModeBackKeepsOptions`, `switchModeTakesNoOptions_n` (run alone), `createEveryMode`

**Known limits, deliberately not changed:**
- `tensor_decoder` does not serialise `set_property` with `decode()` (#4920 C38). This PR adds no new free on that path; a runtime `option2` still frees the label data as before.
- `mode` is never reset from `YOLOV8_ORIENTED_BOUNDING_BOX` after switching away from yolov8-obb. This is pre-existing and not memory-safety.
- G4 and G6 (same file) are separate items.

**Local verification:**
- WSL jammy, `~/nns-g10` (from `origin/main` 8f9eeca8).
- clang-format 16.0.6 clean on all changed C++ files.
- `doxygen-tag.sh` advanced mode passes (with ctags); a planted undocumented function is flagged.
- ASan build (`-Db_sanitize=address`): the 26 decoder-API cases pass, and no leak goes through bounding-box code. The 11 gst-launch/harness cases cannot run under ASan here.
- SSAT is gated per PR: the valgrind step runs this directory's `runTest.sh`, and `-cn _n` only counts `_n` cases as negative; it does not filter the positive ones.


myungjoo:

**Verdict: OK to merge.** The fix correctly addresses G10 of #4920 (per-process singleton `BoxProperties` shared across decoder instances). I read the full diff plus the surrounding files it touches (`tensordec-boundingbox.{h,cc}`, all five `box_properties/*.cc` registration sites, `tests/nnstreamer_decoder_boundingbox/{runTest.sh,unittest_decoder_boundingbox.cc}`, and the CI workflows that exercise this directory) and did not find a correctness or memory-safety bug in the changed code. Two non-blocking observations below are worth the author's attention.

## Verification notes (what I checked and confirmed correct)

- `BoundingBox::bdata` is initialized to `nullptr` in the constructor (`tensordec-boundingbox.cc:537`) and unconditionally `delete`d in the destructor (`:551`), so no leak/UAF from the container side.
- `BoxProperties` has a `virtual ~BoxProperties() = default` (`tensordec-boundingbox.h`), and every concrete subclass (`MobilenetSSD`, `MobilenetSSDPP`, `MpPalmDetection`, `OVDetection`, `YoloV5/V8/V10/V8_OBB`) frees its `name` (`g_strdup_printf`-allocated) in its own destructor. Since the base destructor is virtual, `delete bdata` in `setBoxDecodingMode` (`:924`) and `~BoundingBox` (`:551`) correctly runs the derived destructor — no leak of `name`, even though it is now allocated/freed per decoder instance instead of once per process.
- The same-mode fast path (`tensordec-boundingbox.cc:911-912`) compares against the alias-resolved name (`updateDecodingMode`), so re-selecting `tf-ssd`/`tflite-ssd` after the canonical name (or vice versa) correctly keeps the existing instance — verified against `sameModeKeepsOptions` and the alias table in `bb_modes[]`.
- `addProperties`/`createProperties` locking is consistent: the hash table is now lazily created in `addProperties` (was previously lazily created in the lookup path); `createProperties` treats a still-null table as "not found" rather than creating it, which is safe because all `addProperties` calls (library constructors) run before any decoder can call `createProperties`.
- Registration is consistently converted across all five `box_properties/*.cc` files; no mode was missed, and the pre-existing, unrelated `ov-face-detection` gap (listed in `bb_modes[]` but never registered) predates this PR and isn't touched by it.
- New GTest cases are registered as a plain `test()` in `tests/meson.build:113` with no filter, so all 6 new cases (`yoloV8DecodersOfDifferentModelSizes`, `ssdPpDecodersOfDifferentDetectionCounts`, `optionsOfAnotherDecoder_n`, `sameModeKeepsOptions`, `switchModeDropsOptions_n`, `createEveryMode`) run under `meson test -C build/ -v` in the per-PR `ubuntu_clean_meson_build.yml` job. I traced the box-count math in `yoloV8DecodersOfDifferentModelSizes` (32:32 → 16+4+1=21 boxes, 64:64 → 64+16+4=84 boxes) against yolov8's stride-8/16/32 grid and it is correct; the guard-page buffer is only sized for the small decoder's 21 boxes, so the old shared-singleton bug (large decoder's box count leaking into the small decoder) would genuinely fault there, and the new code doesn't.
- No public API/header is affected (`tensordec-boundingbox.h` is not installed — confirmed no `install_headers` reference for it in `ext/nnstreamer/tensor_decoder/meson.build`), so no documentation update is required for this change.

## Findings

**1. [Low] The new SSAT "8-2" case does not gate pull requests, only the new GTest cases do.**
- Where: `tests/nnstreamer_decoder_boundingbox/runTest.sh` (new "8-2 yolov8 two decoders" case) vs. `.github/workflows/ubuntu_clean_meson_build.yml:209`.
- Why: The only per-PR (`on: pull_request`) SSAT invocation for this test directory is `ssat -n -p=1 --enable-valgrind ... -cn _n`, which (per this project's own `_n`-suffix convention for negative cases) only runs cases whose id/description contains `_n`. The new case is named `"8-2 yolov8 two decoders"` (no `_n`), so it is a positive/golden-compare case that this per-PR job skips. It only runs in the unfiltered SSAT sweep inside the scheduled `nns_testhub.yml` GBS build (cron, not `pull_request`). So a future regression that only shows up at the gst-launch pipeline level (not exercised by the GTest API-level calls) would not fail a PR's CI — it would surface up to a day later in the nightly job.
- Impact: Not blocking for this PR, because the actual bug this PR fixes (shared `BoxProperties` across decoder instances) is already reproduced and gated by the new GTest cases (`yoloV8DecodersOfDifferentModelSizes` etc.), which do run per-PR. But the PR description presents the SSAT case as one of the two test layers proving the fix, and a reader could reasonably assume it also gates merges the way the golden-compare mobilenet-ssd SSAT cases appear to.
- Suggestion: No code change needed for this PR; consider a one-line note in the PR description (or a follow-up) that the new SSAT case is nightly-only coverage, consistent with this repo's existing valgrind-cost-driven `-cn _n` policy for this directory.

**2. [Low] Mode-switch now deletes the previous `BoxProperties` instance while `tensor_decoder` does not serialize `set_property` with `decode()`, turning a pre-existing benign race into a real use-after-free window.**
- Where: `ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc:924` (`delete bdata; bdata = new_bdata;` in `setBoxDecodingMode`), read from concurrently at e.g. `:1077`/`:1144` (`bdata->checkCompatible(...)`, `bdata->decode(...)`).
- Why: Before this PR, `bdata` pointed into a process-wide singleton that was never freed until process exit, so a concurrent `decode()` thread that read a stale `bdata` pointer during an `option1` change still dereferenced valid (if logically stale) memory. After this PR, the old `BoxProperties` is `delete`d as part of the switch, so the same stale read is now a genuine use-after-free rather than just a logic race.
- Impact: The PR description explicitly discloses this as a known, deliberately out-of-scope limitation (issue #4920 item C38: "`tensor_decoder` does not serialise `set_property` with `decode()`"), and no automatic property re-application path in `gsttensor_decoder.c` runs concurrently with `decode()` in the common case, so this is not a new attack surface introduced casually — but it is worth being explicit that this PR does measurably raise the severity of that pre-existing race (crash/UAF vs. stale-but-valid read), specifically inside a security/memory-safety audit (#4920). Given C38 is already tracked as a separate item, no action is required here beyond making sure that follow-up captures this detail explicitly.
- Suggestion: When C38 is picked up, reference this PR's `delete bdata` as the point where the race became a real UAF, not just a logic bug.

**3. [Nit] `createProperties()` returns `nullptr` both for "mode not found" and "allocation/construction failed", but the caller's error message assumes the former.**
- Where: `ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc:1183-1200` (`createProperties`) and the caller at `:914-918` (`nns_loge ("Could not find box properties name %s", param);`).
- Why: `catch (...) { return nullptr; }` swallows any exception a subclass constructor might throw (e.g. `std::bad_alloc`), collapsing it into the same `nullptr` return as "no such mode registered". The log message at the call site is misleading in the (rare) OOM case.
- Suggestion: Optional — distinguish the two cases (e.g. return a tri-state, or log at the point of the caught exception) if this path is ever expected to matter in practice; not worth blocking on.

---
Automated review — flag anything above that looks wrong and I'll follow up.

myungjoo:

Thanks for the review. Addressed in 4f70ed1c (amended into the single commit):

1. **Low — SSAT `8-2` not gated per PR: not changed, the premise does not hold.** SSAT's `-cn _n` / `--countnegative _n` only *counts* `_n`-suffixed cases as negative in the summary; it does not filter anything (`ssat.sh`: the option sets `COUNTNEGATIVE=1` and `COUNTNEGATIVEPOSTFIX`). Running this directory locally with exactly `ssat -n -p=1 -cn _n` executes 65 cases, 52 positive and 13 negative, and `8-2` is among them. On main's sources the same invocation fails both `8-2` checks. So the valgrind step at `ubuntu_clean_meson_build.yml:209` runs `8-2` on every PR.

2. **Low — new use-after-free window on a runtime `option1` switch: fixed.** A decoder no longer deletes the box properties it switches away from. Each `BoundingBox` keeps a table of the box properties of every mode it has selected (at most one per mode) and frees them in its destructor, so `decode()` racing an `option1` change can only see a live object, as with the old singletons. As a side effect, switching back to a mode restores its options, which is what a decoder alone in a process got before this PR. Tests: `switchModeDropsOptions_n` became `switchModeBackKeepsOptions`, a new `switchModeTakesNoOptions_n` checks that a first-selected mode takes nothing from the other mode, and `createEveryMode` now also selects every mode twice in one decoder (leak-checked with LSan: no leak through bounding-box code).

3. **Nit — misleading log when the creator throws: fixed.** The message now reads "Could not create the box properties of mode %s: unknown mode or out of memory".

Local re-verification: 37/37 GTest, SSAT 65/65, clang-format 16 clean, doxygen-tag clean, ASan 26/26 decoder-API cases with no errors.

myungjoo:

**Verdict: OK to merge.** I read the amended head (4f70ed1c) in full — `tensordec-boundingbox.{h,cc}`, all five `box_properties/*.cc` registration sites, both test files, and the CI workflow that exercises this directory — and independently verified the fixes for the two prior findings and the response to the first. No new correctness or memory-safety issue found in the changed code.

## Re-check of the prior review round

**1. [Low] "the SSAT `8-2` case does not gate PRs" — retracted, the author's rebuttal is correct.**
I fetched the actual SSAT source (`ssat.sh`, from `myungjoo/SSAT` — I could not find an `nnstreamer/SSAT` repo; this is the one CI installs via `apt-get install ssat`) and read the `-cn`/`--countnegative` handling directly:
- `COUNTNEGATIVE`/`COUNTNEGATIVEPOSTFIX` are only read once, at the very end of the script, purely to add a "Positive: N / Negative: M" line to the summary output (`ssat.sh:307-313`). They never gate the `find $TARGET -name $TESTCASE` enumeration (`ssat.sh:297`) that decides which test scripts actually run.
- So `-cn _n` does not filter anything; every case in `tests/nnstreamer_decoder_boundingbox/` runs regardless of suffix, and a failure in any of them increments `TNgroupfail`, which makes `ssat.sh` `exit 1` (`ssat.sh:319-328`).
- `.github/workflows/ubuntu_clean_meson_build.yml:209` runs exactly this SSAT invocation for this directory, and that job's trigger is `on: pull_request` (workflow line 3-4). A non-zero exit from that `run:` step fails the CI job.
- Conclusion: the new "8-2 yolov8 two decoders" case does run and does gate every PR, exactly as the author states. My original finding was based on a wrong assumption about what `-cn` does; withdrawn.

**2. [Low] UAF on a runtime `option1` mode switch — fixed, verified by reading the code.**
`tensordec-boundingbox.h` now adds `GHashTable *properties;` per `BoundingBox` instance (alongside the existing `bdata` pointer), created in the constructor with `g_hash_table_new_full(g_str_hash, g_str_equal, g_free, [](gpointer d){ delete static_cast<BoxProperties*>(d); })` and destroyed in `~BoundingBox()`. `setBoxDecodingMode()` (`tensordec-boundingbox.cc:903-928`) now looks up the requested mode's `BoxProperties` in this instance-owned table, lazily creates and inserts it via the new `createProperties()` on a miss, and simply reassigns `bdata = new_bdata` — it no longer calls `delete` on the box properties belonging to a mode the decoder is switching away from. Every mode a given decoder instance has ever selected therefore stays alive for that instance's whole lifetime, so a `decode()` call reading a stale `bdata` pointer concurrently with an `option1` change can no longer observe freed memory; at worst it sees the previous mode's (still valid) object, which is the pre-existing, already-tracked C38 race, not a new UAF. This also matches the new `switchModeBackKeepsOptions` test, which shows switching away from and back to a mode now restores that mode's previously-set options (`option3`/`option4`/etc.), which is the correct and intended side effect of per-instance-not-per-call ownership.
I also checked for double-free/leak risk in the insert path: `g_hash_table_insert` into `properties` only happens after a lookup miss (`new_bdata == nullptr`), so the "insert overwrites an existing key and frees the old value" case glib documents can never trigger here — no double free.

**3. [Nit] Misleading log message on `createProperties()` failure — fixed.**
`tensordec-boundingbox.cc:914-918` now logs "Could not create the box properties of mode %s: unknown mode or out of memory", which correctly covers both cases `createProperties()` collapses to `nullptr` (unregistered mode name, and an exception from the creator).

## Additional checks on the amended head (no new blocking issues)

- No leftover references to the removed `fini_properties_*` destructors or the old per-file `static BoxProperties *` singletons anywhere in `ext/nnstreamer/tensor_decoder/` — the `__attribute__((destructor))` cleanup path is fully removed, not just partially.
- `BoundingBox::addProperties()`'s "already registered → return TRUE without inserting" semantics are preserved (now checked with `g_hash_table_contains` before insert), and the per-instance `properties` table only ever holds at most one entry per known mode name (finite, ~8 creators), so repeated `option1` switching cannot grow it unboundedly.
- Test registration is unchanged from the previous round: `tests/meson.build:113` registers `unittest_decoder_boundingbox` as a plain `test()`, so all the new/rewritten GTest cases (`yoloV8DecodersOfDifferentModelSizes`, `ssdPpDecodersOfDifferentDetectionCounts`, `optionsOfAnotherDecoder_n`, `sameModeKeepsOptions`, `switchModeBackKeepsOptions`, `switchModeTakesNoOptions_n`, `createEveryMode`, rewritten `palmWithoutOption_n`) run per PR under `meson test`.
- Scope stays minimal: only the bounding-box decoder's properties-ownership files and its own tests changed; no other decoder/module touched, no public/installed header affected (`tensordec-boundingbox.h` is not in `install_headers`), so no documentation update is required for this change.

## One pre-existing, non-blocking observation (not introduced by this amendment)

**[Nit] `reinterpret_cast<gpointer>(creator)` / `reinterpret_cast<BoxPropertiesCreator>(...)` casts a function pointer through `gpointer` (`tensordec-boundingbox.cc:1219`, `:1213`).** This is technically implementation-defined under the C++ standard (data pointers and function pointers aren't guaranteed interconvertible), though it is a long-standing, portable-in-practice pattern on every POSIX target this project ships for (the same round-trip `dlsym` itself relies on) and is unchanged from the version already accepted in the previous review round. No action needed; noting only for completeness since the instructions asked for a full re-read of the current head.

---
Automated re-review — flag anything above that looks wrong and I'll follow up.

myungjoo:

Thanks for the re-review. The remaining Nit (a function pointer cast through `gpointer`) is fixed in 1c5c2c8f. The registry now stores each creator in a small `g_new (BoxPropertiesCreator, 1)` slot and reads it back through `BoxPropertiesCreator *`, so no function pointer is converted to or from an object pointer. The slots live as long as the registry table, like the table itself. Re-verified locally: 37/37 GTest, SSAT 65/65, clang-format 16 and doxygen-tag clean, and the ASan decoder-API cases pass with no leak through bounding-box code.

myungjoo:

**Verdict: OK to merge.** I read the full current head (1c5c2c8f) end to end — `tensordec-boundingbox.{h,cc}`, all five `box_properties/*.cc` registration sites, both test files (`tests/nnstreamer_decoder_boundingbox/{runTest.sh,unittest_decoder_boundingbox.cc}`), `tests/meson.build`, and the CI workflows that exercise this directory (`.github/workflows/ubuntu_clean_meson_build.yml`) — not just the delta since the last round. I did not find a new correctness, memory-safety, or scope issue.

## Status of every finding from rounds 1 and 2

**Round 1, Finding 1 [Low] "SSAT `8-2` doesn't gate PRs" — already retracted in round 2, confirmed again here.**
`ubuntu_clean_meson_build.yml` triggers `on: pull_request` (line 3-4) and its per-PR step (line 209) runs `ssat -n -p=1 --enable-valgrind ... -cn _n` for `tests/nnstreamer_decoder_boundingbox`. `-cn`/`--countnegative` only affects the pass/fail *count label* in ssat's summary output, not which test scripts execute — so `runTest.sh`'s new "8-2 yolov8 two decoders" case (`tests/nnstreamer_decoder_boundingbox/runTest.sh:131,133`) runs and gates every PR. Resolved (was a wrong premise, not a code defect).

**Round 1, Finding 2 [Low] "mode switch now frees `bdata` while `decode()` isn't serialized with `set_property`, turning a benign race into a real UAF" — fixed, verified again on the current head.**
`BoundingBox` now owns a per-instance `GHashTable *properties` (`tensordec-boundingbox.h:336`), created in the constructor with a value-destroy of `delete static_cast<BoxProperties*>(data)` (`tensordec-boundingbox.cc:538-539`) and destroyed in `~BoundingBox()` (`:553`). `setBoxDecodingMode()` (`:903-931`) looks the requested mode up in this table, lazily creates it via `createProperties()` on a miss, inserts it, and only ever reassigns `bdata = new_bdata` (`:928`) — there is no `delete` of the box properties a decoder switches away from anywhere in the function. Every mode a given `BoundingBox` instance has ever selected stays alive for that instance's lifetime, so a concurrent `decode()` reading a stale `bdata` can at worst observe the previous mode's still-live object (the pre-existing, separately tracked #4920 item C38 race), not freed memory. Resolved.

**Round 1, Finding 3 [Nit] "misleading log on `createProperties()` failure" — fixed.**
`tensordec-boundingbox.cc:918` logs "Could not create the box properties of mode %s: unknown mode or out of memory", covering both cases the function's `nullptr` return conflates. Resolved.

**Round 2, Nit "`reinterpret_cast` of a function pointer through `gpointer`" — fixed in 1c5c2c8f, verified on the current head.**
`addProperties()` (`tensordec-boundingbox.cc:1212-1229`) now does `BoxPropertiesCreator *slot = g_new (BoxPropertiesCreator, 1); *slot = creator;` and inserts the **data pointer** `slot` into the hash table (`:1224`); `createProperties()` (`:1185-1207`) looks it up and reads back through `static_cast<BoxPropertiesCreator *>` (`:1193-1194`), then dereferences the slot to get the function pointer (`:1196`). No function pointer is ever converted to or from `gpointer`/a data-pointer type — the round-trip is entirely `gpointer <-> BoxPropertiesCreator*` (a data-pointer conversion, always well-defined), with the actual function pointer only ever assigned/read as its own type. This fully resolves the implementation-defined-behavior concern.

I checked the memory-management side of this last change too: `properties_table` (the class-static `GHashTable*`, `tensordec-boundingbox.h:361`) is created with plain `g_hash_table_new` (no value-destroy) and is never destroyed — so the `g_new`'d slots (one per registered mode name, ~8 total, bounded and fixed at process start via the `__attribute__((constructor))` registrations in each `box_properties/*.cc`) are not explicitly freed at exit. This is intentional and harmless: the table is a `static`/process-lifetime root, so the slots stay reachable and neither ASan/LSan's leak detector nor any runtime behavior is affected — the same "small, bounded, still-reachable at exit" pattern the registry itself already used before this PR.

## Additional checks performed this round

- `clang-format` (16.0.6, matching the author's stated version) reports no diff on any of the 8 changed C/C++ files (`box_properties/{mobilenetssd,mobilenetssdpp,mppalmdetection,ovdetection,yolo}.cc`, `tensordec-boundingbox.{h,cc}`, `tests/nnstreamer_decoder_boundingbox/unittest_decoder_boundingbox.cc`).
- Doxygen: every new/changed function has a `@brief` — the five `create_properties_*()` creators, `createProperties()`/`addProperties()`, and all 7 new/rewritten `TEST(...)` cases (`yoloV8DecodersOfDifferentModelSizes`, `ssdPpDecodersOfDifferentDetectionCounts`, `optionsOfAnotherDecoder_n`, `sameModeKeepsOptions`, `switchModeBackKeepsOptions`, `switchModeTakesNoOptions_n`, `createEveryMode`, and the rewritten `palmWithoutOption_n`).
- No leftover references anywhere in the tree to the removed `BoundingBox::getProperties()` API or to any `fini_properties_*` destructor — the old per-process-singleton cleanup path is fully removed, not partially.
- Scope stays minimal: only the bounding-box decoder's properties-ownership files and their own tests changed (`git diff --stat` against main touches exactly the 5 `box_properties/*.cc` files, `tensordec-boundingbox.{h,cc}`, and the two test files under `tests/nnstreamer_decoder_boundingbox/`); no other decoder or module is touched.
- No installed/public header is affected (`tensordec-boundingbox.h` is not referenced by any `install_headers()` in `ext/nnstreamer/tensor_decoder/meson.build`), so no documentation update is required for this change.
- Read the new test bodies directly rather than trusting the PR description: `optionsOfAnotherDecoder_n` (line 1418) and `switchModeTakesNoOptions_n` (line 1515) each construct a second, independent decoder/mode-switch and confirm it starts from defaults rather than inheriting the first's `option3`/`option5`; `switchModeBackKeepsOptions` (line 1488) confirms switching away and back restores the mode's own previously-set options; `createEveryMode` (line 1541) exercises every registered mode name in a fresh decoder and then twice in one decoder, which is the case the PR states was checked under LSan for leaks.

## Remaining Nits

None. The one Nit left open after round 2 (the function-pointer/`gpointer` cast) is resolved, and this round found no new issue to report.

---
Automated re-review — flag anything above that looks wrong and I'll follow up.

myungjoo:

Thanks for the review (5221322764). The two NITs:

1. **`BoxProperties::name` is dead: fixed** in 4c4af043, which is also rebased onto the current `main` 4d028546. The field is removed from `tensordec-boundingbox.h` together with every `g_strdup_printf`/`g_free` of it. Three kinds of member became empty and are removed as well: the `OVDetection` constructor, and the destructors of `OVDetection`, `MobilenetSSDPP` and the four yolo classes. The base class keeps its virtual destructor, so the table's `delete` through `BoxProperties *` is unchanged. This part of the diff is deletions only (−59 lines). Nothing can drift from the registry key any more: the mode name exists only in `init_properties_*()`, and `createEveryMode` still selects every name.
2. **SSAT `8-2` second decoder without `option4`: not changed, it already has one.** `runTest.sh` gives that branch `option4=160:160 option5=160:160 ! fakesink`, so it negotiates 160x160 RGBA.

Local re-verification on the amended head: 37/37 GTest, SSAT 65/65, clang-format 16 and doxygen-tag clean, and 26/26 ASan decoder-API cases with no leak through bounding-box code.

myungjoo:

**Verdict: OK to merge.** I read the current head (`4c4af043`) end to end — `tensordec-boundingbox.{h,cc}`, all five `box_properties/*.cc` files, both test files, and `tests/meson.build` — and independently rebuilt and ran the full test matrix in an isolated environment rather than just reading the diff. No correctness, memory-safety, or scope issue found.

## Independent verification performed this round

- Fresh build of the current head (`4c4af043`) in an isolated tree: `meson`/`ninja` clean build.
- `meson test`-equivalent run of `unittest_decoder_boundingbox`: **37/37 GTest cases pass**, including all 7 new/rewritten cases (`yoloV8DecodersOfDifferentModelSizes`, `ssdPpDecodersOfDifferentDetectionCounts`, `optionsOfAnotherDecoder_n`, `sameModeKeepsOptions`, `switchModeBackKeepsOptions`, `switchModeTakesNoOptions_n`, `createEveryMode`).
- SSAT run of `tests/nnstreamer_decoder_boundingbox`: **65/65 pass** (52 positive / 13 negative), including the new "8-2 yolov8 two decoders" case and its golden-diff check.
- `clang-format` (repo config): **clean diff** on all 8 changed C/C++ files (`tensordec-boundingbox.{h,cc}`, the 5 `box_properties/*.cc` files, `unittest_decoder_boundingbox.cc`).
- ASan/LSan run of the full `unittest_decoder_boundingbox` binary (symbolized with `llvm-symbolizer`): **37/37 pass**; the leak report contains no stack frame in `tensordec-boundingbox.cc` or any `box_properties/*.cc` — every leak stack bottoms out in GObject/GStreamer type-registration code (e.g. `gst_tensor_converter_class_intern_init`), which is pre-existing and unrelated to this change.
- `git grep` confirms no leftover reference anywhere in the tree to the removed `BoxProperties::name` field, the old `getProperties()` API, or any `fini_properties_*` destructor.
- `virtual ~BoxProperties() = default;` is unchanged, so `delete` through the base pointer (in `~BoundingBox()`'s hash-table value-destroy) still runs the correct derived destructor for every subclass, even after this round's constructor/destructor removals.
- Commit is a single commit with one `Signed-off-by:` line, no duplicate DCO trailer, and every new/changed function keeps a `@brief` Doxygen comment.
- Scope check: the diff touches exactly `ext/nnstreamer/tensor_decoder/{tensordec-boundingbox.{h,cc},box_properties/*.cc}` and `tests/nnstreamer_decoder_boundingbox/*` — no other module. `tensordec-boundingbox.h` is not in any `install_headers()`, so no documentation update is required.

## Status of every prior finding

**Round 1, Finding 1 [Low] "SSAT `8-2` doesn't gate PRs"** — retracted in round 2 (wrong premise about `-cn`), reconfirmed in round 3, and reconfirmed again here: my own SSAT run shows `8-2` and `8-2 diff` executing and passing under the exact per-PR invocation. Resolved (was never a real defect).

**Round 1, Finding 2 [Low] UAF on a runtime `option1` mode switch** — fixed since round 1's follow-up commit and reconfirmed on the current head: `BoundingBox` still keeps a per-instance `GHashTable *properties` (mode name → owned `BoxProperties`), created in the constructor and destroyed in `~BoundingBox()`; `setBoxDecodingMode()` never deletes the properties of a mode it switches away from, only reassigns `bdata`. Verified by reading `tensordec-boundingbox.cc` directly (the `setBoxDecodingMode`/`createProperties`/`addProperties` bodies are byte-for-byte the same as the previously-verified round 3 head). Resolved.

**Round 1, Finding 3 [Nit] misleading log on `createProperties()` failure** — fixed and unchanged since; the log at `tensordec-boundingbox.cc` still reads "Could not create the box properties of mode %s: unknown mode or out of memory", covering both the unregistered-mode and allocation-failure cases. Resolved.

**Round 2 Nit: function pointer cast through `gpointer`** — fixed and unchanged since; `addProperties()`/`createProperties()` still round-trip only `gpointer <-> BoxPropertiesCreator*` (a `g_new`'d slot holding the function pointer), never converting the function pointer itself to/from a data-pointer type. Resolved.

**Round 3 (review 5221322764) NIT 1: `BoxProperties::name` is dead** — **fixed in this amendment (`4c4af043`).** The field is removed from `tensordec-boundingbox.h`, together with every `g_strdup_printf`/`g_free` of it in the five `box_properties/*.cc` files. Three constructors/destructors that became empty as a result are also removed (`OVDetection`'s constructor and destructor, `MobilenetSSDPP`'s destructor, and the destructors of `YoloV5`/`YoloV8`/`YoloV10`/`YoloV8_OBB`), a net −59 lines. The base class's `virtual ~BoxProperties() = default` is untouched, so the hash table's `delete` through `BoxProperties *` still correctly invokes each subclass's (now-trivial) destructor — verified no leak was introduced by the ASan run above. No reference to the removed field remains anywhere in the tree. **Resolved.**

**Round 3 (review 5221322764) NIT 2: SSAT `8-2`'s second decoder has no `option4`** — the author's rebuttal ("not changed, it already has one") is correct: `runTest.sh`'s `8-2` case gives the second decoder `option4=160:160 option5=160:160`, matching the resolution negotiated into `fakesink`. Confirmed directly in the diff and by the SSAT run above (`8-2`/`8-2 diff` both pass). This NIT was based on an incorrect reading of an already-correct line; there was no defect to fix. **Resolved (no code change needed).**

## Remaining Nits

**None.** Every Nit raised across all three prior rounds (including both from review 5221322764) is either fixed in code or was a correctly-rebutted false premise, and this round's independent build/test/format/ASan verification found nothing new.

---
Automated re-review (round 4) — flag anything above that looks wrong and I'll follow up.

myungjoo:
Approved by Astra

PR #4799, opened 2025-11-13
[Release] update version

myungjoo:
Update version to 2.6.0, 2025 last release.

I'm merge this version change when the pending bugfixes are merged.

---

> [!NOTE]
> Bumps NNStreamer to 2.6.0 and updates CHANGES plus Debian/RPM/Android/Meson version metadata and changelogs.
> 
> - **Release/Versioning**:
>   - Set version to `2.6.0` in `meson.build`, `jni/nnstreamer.mk`, and `packaging/nnstreamer.spec`.
>   - Update `CHANGES` with `2.6.0` notes (LLM support, bug fixes).
>   - Add `2.6.0` entries to `debian/changelog` and `%changelog` in `packaging/nnstreamer.spec`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit cfd0a76265a52697fa594b6b7e8f21eecb9dbe22. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4983, opened 2026-09-16
[Doc] Keep CHANGES for release time and say so in CLAUDE.md

myungjoo:
- **What:** `CHANGES` is written by the maintainers at release time from the accumulated history, not per PR. The `2.6.0 -> 2.7.0` section should therefore hold only the devel-version note, but PR #4829 added a per-PR entry to it, and nothing in the tree states the rule.
- **How:** removes that entry (the four lines commit 85129d6f added, so `CHANGES` is byte-identical to its state before that commit), and adds one line to the *Branch and PR policy* of `CLAUDE.md`: do not edit `CHANGES` outside a release PR, do not ask for entries in reviews, and record user-visible behavior changes in the PR body, commit message and component documentation instead.
- **Result:** documentation only, 1 line added and 4 removed; no build, test or runtime effect. The newline, hardcoded-path, misspelling and prohibited-words checkers pass on both files.

### Why now

The per-PR entry keeps being copied. PR #4829 added one, the open #4971 adds another in the same place, and a review agent asked #4969 for one, citing the #4829 entry as precedent. Agents infer the convention from the file itself, so the existing entry reads as the rule.

### Nothing is lost

The removed entry describes the `getEmptyInstance()` ownership change of #4829. It remains in #4829, in the message of commit 85129d6f, and in `Documentation/writing-subplugin-tensor-filter.md` (the ownership rules at lines 40-43), so it can be picked up when the next release note is written.

### Test

No test: this changes documentation only, and nothing reads `CHANGES` or `CLAUDE.md` at build or run time.


myungjoo:

Thanks. NIT taken at head `2412bc2c` (amended, same single commit). The rule now reads "Do not edit `CHANGES` outside a release PR ... maintainers write it when a version is released, from the accumulated history", so release PRs such as #4799 (84a35815), which write `CHANGES` together with `debian/changelog`, `meson.build` and the spec, are no longer covered. The commit message and the PR body say the same. Nothing else changed; `CHANGES` is still identical to its state before 85129d6f.

PR #4985, opened 2026-09-17
[Filter] Build the output-combination buffer in the order the caps describe

myungjoo:
- **What:** `tensor_filter output-combination` wrote the `o` entries into the caps in list order with repeats, but built the buffer in model order with each output once, so `o1,o0` or `o0,o0` produced a buffer that does not match its caps and a downstream reader overran a memory (#4920 B15).
- **How:** outputs are still prepared once in model order; when `o` entries are given, the buffer is then filled by walking the list (one memory reference per entry) and each listed output's preparation reference is dropped. `i` entries, `o0,o1`, subsets and input-only combinations behave as before. README states the order.
- **Result:** new `tensorFilterOutputCombination` harness tests (10, two `_n`) check each output memory against the negotiated caps, including flexible and dynamic-invoke output, and count GStreamer critical messages; 6 of them fail on main and all pass with the fix. `unittest_sink`, `unittest_filter_custom`, `unittest_plugins`, `unittest_common` and `unittest_filter_single` pass locally, and `clang-format` 16 and `doxygen-tag.sh` are clean.

Part of #4920 (item B15).

<details>
<summary>Notes for reviewers and agents</summary>

**Reproduction (main):** a custom-easy model with a 4-byte and an 8-byte output. `o1,o0` → caps `dimensions=8,4`, buffer `[4][8]`; `o0,o0,o1` → caps 3 tensors, buffer 2 memories; `o1,i0,o0,o1` → caps `4,8,4,8` (input first), buffer `[4][4][8]`.

**Why follow the list instead of refusing reorder/repeat in the setter:** the caps have always advertised the listed order, the `i` side and `input-combination` already honour list order, and refusing would break pipelines whose negotiated caps are already in list order. Making the buffer match the caps is the smaller and compatible change.

**Memory handling:** `gst_tensor_buffer_append_memory()` takes ownership, so each list entry appends `gst_memory_ref (mem[i])`; afterwards each listed output drops its single preparation reference (unlisted ones are released in the loop, as before). A repeated entry puts the same `GstMemory` into the buffer twice, the same way repeated `i` entries already do.

**Dynamic invoke:** the error path for an invalid dynamic info now unrefs the listed outputs prepared before the failing index (previously they were already in `outbuf`, which the caller unrefs). The `i >= output_meta.num_tensors` skip keeps the old tolerance of a dynamic invoke returning fewer outputs. A custom-easy dynamic model reports no output info at negotiation, so `get_combined_out_info()` refuses any `o` entry; the dynamic tests give `output=4,8 outputtype=uint8,uint8` and flexible sink caps so the list negotiates. `dynamicFewerOutputs` has the model return one output for `o1,o0,o0`: the skip leaves `o1` out, and without it `gst_memory_ref (NULL)` raises a GStreamer critical message (the output buffer alone would still look right, which is why every run counts critical messages). `invalidDynamicInfo_n` reaches that error path (its log line is emitted); removing the new unref loop makes ASan report the prepared memory as a direct leak through `_gst_tensor_filter_transform_update_outbuf`, and the fixed build reports no leak through `tensor_filter.c`.

**Test placement:** `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`, next to the other custom-easy harness tests (reuses `_flex_in_teardown ()`). They were in `unittest_sink.cc` while #4980 edited this file; moved after it merged.

**Not verified locally:** valgrind (the local WSL libc debuginfo is missing; ASan was used instead). The CI valgrind step runs `unittest_sink` recursively. The tflite SSAT groups that use `output-combination=i0,o0` skipped locally without the tflite library; that list is in model order, so behaviour is unchanged.

</details>


myungjoo:
This is a relayed review written by another agent (Claude Sonnet), posted on the user's behalf.

**Scope check (#4920 B15).** `gst_tensor_filter_common_get_combined_out_info()` builds the advertised src caps by walking `out_combi_i` then `out_combi_o` in list order, repeats included. Before this PR, `_gst_tensor_filter_transform_update_outbuf()` still walked model outputs in model order and appended each selected one once, so `o1,o0` (caps `8,4`) produced a buffer `[4][8]`, and `o0,o0` (caps: 2 tensors) produced a buffer with only 1 memory. That is exactly a caps/buffer mismatch that lets a downstream reader size a read from the caps and overrun the memory. The fix keeps the single per-output preparation (line ~1113–1182, still releasing unlisted outputs exactly as before) and then, only when `out_combi_o_defined`, appends the prepared memories a second time by walking `out_combi_o` in list order (`gst_memory_ref` per listed entry, one final `gst_memory_unref` per unique listed index to drop the preparation reference). I traced the refcounting for `o0,o0,o1` (repeat) and `o1,i0,o0,o1` (mixed) by hand: it works out to exactly one ref per buffer entry plus the dropped preparation ref, with no leak and no double free. This directly and correctly addresses the reported defect.

**Regression risk to other modules: none found.** `out_combi_o`/`out_combi_i` are only read in `tensor_filter.c` and `tensor_filter_common.c` — `tensor_filter_single.c` (the ml-api single-shot path) never references output-combination, so it cannot be affected. The `i`-side path (lines 1077–1111) is untouched, so plain `input-combination` and mixed cases that were already correct (`o0,o1` in model order, subsets, input-only) keep their old code path (`if (!priv->combi.out_combi_o_defined) gst_tensor_buffer_append_memory(...)` inline, unchanged for the non-combi and combi-i-only cases).

**Code size:** proportionate. The functional change is +26/-2 lines in `tensor_filter.c` plus a 1-line README clarification; the bulk of the diff (242 lines) is new tests in `tests/nnstreamer_sink/unittest_sink.cc`, reusing a shared harness helper rather than duplicating setup per test. No unrelated files/modules are touched.

**Test completeness:** good. The new `tensorFilterOutputCombination` suite drives a real custom-easy model through a `GstHarness` and checks every output memory's size/content against the *negotiated caps*, which is the right level (it would have caught the original bug: 4 of the 7 new cases are stated to fail on `main`). Covered: model order, subset, reorder, repeat, input+reorder mixed, flexible-format reorder, and an out-of-range index negative case (`invalidIndex_n`, consistent with this file's existing `_n` naming convention). Because these assert on final caps+buffer bytes rather than internal state, they should also catch future regressions introduced by unrelated changes to `gst_tensor_buffer_append_memory`, `get_combined_out_info`, or the meta/flexible-header helpers — i.e., they satisfy the "detect regressions from other modules" bullet reasonably well. All of this runs as GTest under `meson test`, which is exercised by several CI jobs (including the Valgrind job per the workflow list), so a reintroduced bug here should fail CI, not just local runs.

- **[Low] Untested error-cleanup branch.** The new cleanup added at the dynamic-invoke "invalid tensor info" error path (tensor_filter.c, the `for (j = 0; priv->combi.out_combi_o_defined && j < i; j++) { if (g_list_find(...)) gst_memory_unref(out_trans_data->mem[j]); }` block) has zero test coverage. The PR body explains why: a custom-easy dynamic model reports no output info before invoke, so `get_combined_out_info()` already refuses any `o` entry at negotiation time (pre-existing behavior, unrelated to this PR), so this specific branch cannot currently be reached from a GTest/SSAT-constructible scenario. I manually traced the branch and it looks correct (it only unrefs `mem[j]` for indices that are (a) already prepared, i.e. `j < i`, and (b) still pending in `out_trans_data->mem[]` because they are members of `out_combi_o` — unlisted indices were already released earlier in the loop via `gst_tensor_filter_destroy_notify_util`, so there's no double free). Still, it's new memory-lifetime code guarded only by manual review, not a regression test, so if a future subplugin path does reach it and something is slightly off, CI would not catch it. Non-blocking, but worth a follow-up issue/TODO if a reachable scenario is ever found.
- **[Nit] CI was still pending at review time.** Only "Static checks" and "DCO" had completed when this review was written; the Valgrind, RISC-V, Yocto, Tizen GBS, and packaging jobs were still queued/pending. Given the change touches buffer/memory lifetime code exercised by SSAT pipelines using `output-combination` (the PR notes the tflite `i0,o0` groups couldn't be run locally for lack of the tflite library), please confirm the full matrix goes green — especially the Valgrind job — before merging.
- **[Nit] Documentation.** No architecture or public API change here (output-combination's contract was already documented; this PR only makes the implementation match the caps it already advertised), so no further doc changes are needed beyond the one-line README clarification already included.

**Overall verdict:** OK to merge (once full CI, in particular the Valgrind and SSAT-bearing jobs, finishes green). No High/Medium issues found; the two Low/Nit notes above are non-blocking.

myungjoo:
This is a relayed review written by another agent (Claude Sonnet), posted on the user's behalf.

Reviewed at head `3706bbc9` (the PR is now a single squashed commit after the force-push that followed the earlier relayed review on this thread).

**Follow-up on the earlier review's Low finding.** The earlier relayed review flagged one gap: the dynamic-invoke error-cleanup branch (`for (j = 0; priv->combi.out_combi_o_defined && j < i; j++) { if (g_list_find(...)) gst_memory_unref(out_trans_data->mem[j]); }`) had no test reaching it. The force-pushed head adds `tensorFilterOutputCombination.invalidDynamicInfo_n`, which does reach it: `cef_func_out_combi_dynamic` leaves `out_info->info[1].type` unset (stays `_NNS_END` from `gst_tensors_info_init`) when its `data` argument is non-NULL, so `gst_tensor_info_validate()` fails at `i==1` — I confirmed this against `gst_tensor_info_init`/`gst_tensor_info_validate` in `nnstreamer_plugin_api_util_impl.c` (type `_NNS_END` is rejected). At that point the model-order preparation loop has already produced a preparation ref for `mem[0]` (index 0 is listed in `output-combination=o1,o0`), and the new `j==0` iteration is exactly what unrefs it. This matches the PR body's claim that removing the new unref makes ASan report a direct leak through this function. The previously-flagged gap is closed.

**Correctness of the core fix (re-verified independently).** In `_gst_tensor_filter_transform_update_outbuf()` (`gst/nnstreamer/tensor_filter/tensor_filter.c`), outputs are still prepared once each in model order, with the unlisted-output release path unchanged; the buffer is then filled by a second pass over `out_combi_o` in list order — one `gst_memory_ref()` per listed entry — followed by exactly one `gst_memory_unref()` per unique listed index to drop the single preparation reference. I hand-traced the ref/unref balance for a repeat (`o0,o0`: 1 prep ref + 2 append refs − 1 drop = 2 refs backing 2 buffer entries, no leak/double-free) and for the new `if (i >= prop->output_meta.num_tensors) continue;` guard in the append loop. That guard only matters when a dynamic invoke returns fewer tensors than declared (a pre-existing tolerance, not introduced here); it is not what makes `invalidIndex_n` pass, since out-of-range indices are already rejected earlier, at caps negotiation, in `gst_tensor_filter_common_get_combined_out_info()` (`tensor_filter_common.c`, the `i >= out->num_tensors` check) — this function never runs for that case.

**Regression risk: none found.** `out_combi_i`/`out_combi_o` are only read in `tensor_filter.c`/`tensor_filter_common.c`; `tensor_filter_single.c` (the ml-api single-shot path) never touches output-combination. The `i`-side path and the plain non-combi append (`if (!priv->combi.out_combi_o_defined) gst_tensor_buffer_append_memory(...)`) are untouched, so pipelines using model-order `o0,o1`, input-only combination, or no combination at all keep the exact old code path.

**Size/scope:** proportionate to the defect — `+26/-2` in `tensor_filter.c` plus a one-line README clarification; the remaining ~336 lines are new tests that reuse one shared harness helper rather than duplicating setup. No unrelated files or modules are touched.

**Test coverage:** thorough and pitched at the right level. `tensorFilterOutputCombination` drives a real custom-easy model through a `GstHarness` and checks every output memory's size and content against the *negotiated caps* — model order, subset, reorder, repeat, input+reorder mixed, flexible reorder, dynamic-invoke reorder, an out-of-range index (`invalidIndex_n`), and the dynamic-invoke error path (`invalidDynamicInfo_n`). Because the assertions are against the final caps/buffer contract rather than internal state, this suite should also catch future regressions introduced by unrelated changes to `gst_tensor_buffer_append_memory`, the flexible-header helpers, or `get_combined_out_info`. All cases run as GTest under `meson test`, which multiple CI jobs exercise (including the Valgrind job, per the workflow list), so a reintroduced bug here should fail CI rather than pass silently.

**Documentation:** no architecture or public API change — output-combination's contract already existed; the one-line README addition documents the ordering the implementation now actually follows, rather than changing any contract. No further doc updates are needed.

Findings:
- **[Low, resolved]** The dynamic-invoke error-cleanup branch was untested at the time of the earlier review; `invalidDynamicInfo_n` now exercises it directly. No action needed.
- **[Nit] CI still pending.** At review time only `DCO` had completed on this force-pushed head; Valgrind, RISC-V, Yocto, Tizen GBS, and the packaging jobs were queued. Since this change is buffer/memory-lifetime code, and the PR notes the tflite `output-combination=i0,o0` SSAT groups could not be run locally, please confirm the full matrix — especially Valgrind and the SSAT-bearing jobs — goes green before merging.
- **[Nit] No CHANGES entry** — correctly omitted per project policy (CHANGES is written at release time).

**Overall verdict: OK to merge**, once full CI (particularly Valgrind and the SSAT jobs) finishes green. No High or Medium issues found.

myungjoo:
Approved by Astra

myungjoo:
This is a relayed review written by another agent (Claude Sonnet), posted on the user's behalf. Reviewed at head `1e311518` (the PR was rebased onto current `main` and force-pushed after the earlier reviews on this thread).

**Scope check (#4920 B15).** `gst_tensor_filter_common_get_combined_out_info()` (tensor_filter_common.c) builds the advertised src caps by walking `out_combi_i` then `out_combi_o` in list order, repeats included. Before this PR, `_gst_tensor_filter_transform_update_outbuf()` still walked model outputs in model order and appended each selected one once, so `o1,o0` (caps `8,4`) produced a buffer `[4][8]`, and `o0,o0` (2 tensors in caps) produced a buffer with only 1 memory — a caps/buffer mismatch that lets a downstream reader size a read from the caps and overrun the memory. The fix keeps the existing per-output preparation and release-if-unlisted logic unchanged, and, only when `out_combi_o_defined`, appends the prepared memories a second time by walking `out_combi_o` in list order (`gst_memory_ref` per listed entry, then exactly one `gst_memory_unref` per unique listed index to drop the preparation reference). I hand-traced the ref/unref balance for a repeat (`o0,o0,o1`) and a mixed case (`o1,i0,o0,o1`): one ref per buffer entry plus the dropped preparation ref, no leak, no double free. This directly and correctly addresses the reported defect.

**Both prior NITs from the myungjoo-bot APPROVED review are resolved on this head:**
1. *"No test reaches the `i >= prop->output_meta.num_tensors` skip."* — I traced this end-to-end: `tensor_filter_custom_easy.c`'s `custom_invoke()` passes `&prop->output_meta` itself as the dynamic invoke's `out_info` argument, so the new `dynamicFewerOutputs` test's model callback (`out_info->num_tensors = 1`) directly shrinks `prop->output_meta.num_tensors` to 1 during invoke. The subsequent `for (i = 0; i < prop->output_meta.num_tensors; i++)` loop then never visits index 1, and the new `if (i >= prop->output_meta.num_tensors) continue;` guard in the append loop is exactly what keeps `o1` (index 1, still in the `out_combi_o` list) from being `gst_memory_ref()`'d as NULL/stale — this does reach and exercise that branch.
2. *"Tests live in unittest_sink.cc."* — confirmed: `tests/nnstreamer_sink/unittest_sink.cc` no longer contains any `OutputCombination`/`out_combi` reference; all 10 tests now live in `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`, next to the other custom-easy harness tests, reusing `_flex_in_teardown()`.

**Regression risk to other modules: none found.** `out_combi_i`/`out_combi_o` are only read in `tensor_filter.c`/`tensor_filter_common.c`; `tensor_filter_single.c` (the ml-api single-shot path) never touches output-combination. The `i`-side path and the plain non-combi append (`if (!priv->combi.out_combi_o_defined) gst_tensor_buffer_append_memory(...)`) are untouched, so pipelines using model-order `o0,o1`, input-only combination, or no combination at all keep the exact old code path. The async-output callback path (`_gst_tensor_filter_transform_update_outbuf(trans, NULL, ...)`) shares the same fixed function, so it benefits from (and isn't regressed by) the same fix.

**Size/scope:** proportionate — `+26/-2` in `tensor_filter.c` plus a one-line README clarification; the remaining ~371 lines are new tests reusing one shared harness helper. No unrelated files/modules touched.

**Test completeness:** good, and pitched at the right level. `tensorFilterOutputCombination` drives a real custom-easy model through a `GstHarness` and checks every output memory's size/content against the *negotiated caps* — model order, subset, reorder, repeat, input+reorder mixed, flexible reorder, dynamic-invoke reorder, dynamic invoke returning fewer outputs, an out-of-range index (`invalidIndex_n`), and the dynamic-invoke validation-error path (`invalidDynamicInfo_n`, which reaches the new error-path unref loop — removing that loop is stated to produce an ASan-reported leak). Every case also counts GStreamer criticals via a custom `g_log_set_handler`, which is a good defense against a silently-wrong `gst_memory_ref(NULL)`. Because assertions check the final caps/buffer contract rather than internal state, this suite should also catch future regressions from unrelated changes to `gst_tensor_buffer_append_memory`, the flexible-header helpers, or `get_combined_out_info`. These are GTest cases in an existing CI-covered binary (`unittest_filter_custom`), so a reintroduced bug should fail CI, not just local runs — and the memory-lifetime path here (the leak fix) should also be caught by the Valgrind CI job, which builds and runs this binary.

**Documentation:** no architecture or public API change; the one-line README addition documents the ordering the implementation now actually follows. No further doc updates needed.

Findings:
- **[Medium] CI is currently failing on this head — "Spell Check with Typos".** `tests/nnstreamer_filter_custom/unittest_filter_custom.cc:1558`, in the new `@brief` for `_out_combi_count_critical`: `"Log handler counting the criticals it receives..."` — `typos` flags `criticals` and fails the job (`error: 'criticals' should be 'critical'`, exit code 2). This is a real, current, blocking failure introduced by this PR's own new content (not a flaky/unrelated job), confirmed via `gh run view` on the latest run for this head. No functional risk, but it must be fixed (e.g., reword to "the critical messages it receives" or "criticals it counts") before this can go green and be merged.
- **[Nit] Rest of CI still largely pending on this freshly-pushed head.** At review time, `Static checks`, `CPP-Linter`, `DCO`, `Analyze (python)`, the LLVM/Clang build, the macOS/Meson build, the ExecuTorch SSAT build, and the `ubuntu-22.04-arm` Valgrind job have all passed; RISC-V, Yocto, Android NDK, the Tizen GBS builds, the `ubuntu-22.04` (x86) Valgrind job, and the pdebuild jobs are still pending/queued. Please confirm the full matrix (particularly the x86 Valgrind job) goes green once the typos issue above is fixed, before merging.
- **[Nit] Unused parameters in the new static test callbacks.** `cef_func_out_combi(void *data, const GstTensorFilterProperties *prop, const GstTensorMemory *in, GstTensorMemory *out)` doesn't use `data`/`prop`/`in`. Other callbacks of the same shape elsewhere in this file mix conventions (some wrap unused params in `UNUSED(...)`, some don't), and the project doesn't build tests with `-Werror` for unused parameters, so this doesn't affect CI — purely a style nit, not blocking.

**Overall verdict: what must change before merge** — fix the `criticals` → `critical` typo flagged by the `Spell Check with Typos` CI job (currently a hard failure on this head), then confirm the remaining pending CI jobs (especially the x86 Valgrind job) go green. No High or Medium *functional/design* issues were found in `_gst_tensor_filter_transform_update_outbuf()` or the new tests themselves; the fix's reference-counting and combi-list-ordering logic were independently re-traced and are correct, and both of the earlier bot review's NITs are resolved on this head.

myungjoo:
This is a relayed review written by another agent (Claude Sonnet), posted on the user's behalf.

Reviewed at head `ca68d970` (force-pushed after the earlier relayed review that required the `criticals` → `critical` typo fix). This pass differs from the earlier ones on this thread in one respect: instead of only re-tracing the logic by hand, I actually fetched this PR head into a local build environment, built `tests/unittest_filter_custom` and neighboring suites, and ran them (including a before/after comparison against the pre-fix code and an AddressSanitizer build), to independently confirm the claims already made in this thread.

**Scope check (#4920 B15).** `gst_tensor_filter_common_get_combined_out_info()` (`tensor_filter_common.c:2293`) builds the advertised src caps by walking `out_combi_i` then `out_combi_o` in list order, repeats included. Before this PR, `_gst_tensor_filter_transform_update_outbuf()` (`tensor_filter.c`) still walked model outputs in model order and appended each selected one once, so `o1,o0` (caps `8,4`) produced a buffer `[4][8]`, and `o0,o0` (2 tensors in caps) produced a buffer with only 1 memory — a caps/buffer mismatch that lets a downstream reader size a read from the caps and overrun the memory. The fix keeps the existing per-output preparation and release-if-unlisted logic unchanged (`tensor_filter.c:1113-1182`), and, only when `out_combi_o_defined`, appends the prepared memories a second time by walking `out_combi_o` in list order (one `gst_memory_ref()` per listed entry, `tensor_filter.c:1184-1194`), then drops exactly one preparation `gst_memory_unref()` per unique listed index (`tensor_filter.c:1196-1199`).

**Independent build/test verification (new in this pass).**
- Built the exact PR head (`ca68d970`) with meson/ninja and ran the new `tensorFilterOutputCombination` suite: all 10 tests pass, 0 GStreamer criticals other than the self-check.
- Reverted only `tensor_filter.c` to its pre-PR (`40f3246e`) state while keeping the new tests, rebuilt, and reran: `reorder`, `repeat`, `inputAndReorder`, `reorderFlexible`, `reorderDynamic`, and `dynamicFewerOutputs` fail exactly as the PR body claims (buffer/caps mismatches: wrong byte counts, wrong tensor counts, wrong fill bytes), and the same 6 pass again once the fix is restored. This confirms the tests actually exercise and would catch the described bug, not just pass trivially.
- Also confirmed `flexibleInvoke_p`, `staticFlexibleInvoke_p`, and `notRegisterFlexibleInvoke_n` fail in this sandbox both with and without the PR (missing test fixtures/pipeline elements in the local environment) — pre-existing, unrelated to this change, not a regression.
- Ran `unittest_common` (177/177) and `unittest_filter_single` (6/6) clean on the PR head; no regressions observed in adjacent suites.
- Built with `-Db_sanitize=address` and reran `tensorFilterOutputCombination.*`: all 10 pass, and the only leak reported is a single 16384-byte `g_malloc` from GLib/GStreamer plugin-registry init — the same leak appears from a bare `gst-inspect-1.0 tensor_filter` with no test code involved, so it is environment noise, not attributable to this change.
- CI on this head: all 20 checks pass, including both Valgrind jobs and Spell Check.

**Correctness of the core fix (re-verified).** I hand-traced the ref/unref balance for a repeat (`o0,o0,o1`: 1 preparation ref, 2 appended refs via the walk, 1 final drop → 2 refs backing 2 buffer entries, matches observed behavior) and the dynamic-invoke error-cleanup branch (`tensor_filter.c:1146-1149`, added by this PR): it only unrefs `mem[j]` for `j < i` that are members of `out_combi_o` — i.e., already prepared and not yet appended (since combi mode defers all appending until after the preparation loop) — while unlisted `j`s were already released earlier via `gst_tensor_filter_destroy_notify_util`. No double free, no leak, consistent with `invalidDynamicInfo_n` passing.

**Regression risk to other modules: none found.** `out_combi_i`/`out_combi_o` are only read in `tensor_filter.c`/`tensor_filter_common.c`; `tensor_filter_single.c` never touches output-combination. The `i`-side path and the plain non-combi append are untouched, so pipelines using model-order `o0,o1`, input-only combination, or no combination at all keep the exact old code path.

**Size/scope:** proportionate — `+26/-2` in `tensor_filter.c` plus a one-line README clarification; the remaining ~380 lines are new tests reusing one shared harness helper. No unrelated files/modules touched.

**Test completeness / future-regression detection:** good. The suite checks every output memory's size and content against the *negotiated caps* (not internal state), covering model order, subset, reorder, repeat, input+reorder mixed, flexible reorder, dynamic-invoke reorder, dynamic invoke returning fewer outputs, an out-of-range index (`invalidIndex_n`), and the dynamic-invoke error path (`invalidDynamicInfo_n`). Every case also counts GStreamer criticals, which would catch a future `gst_memory_ref(NULL)`-style regression even if buffer contents happened to look right. These are GTest cases in `unittest_filter_custom`, exercised by the Valgrind CI job, so a reintroduced bug here should fail CI.

**Documentation:** no architecture or public API change; the one-line README addition documents the ordering the implementation now actually follows. No further doc updates needed. No CHANGES entry, correctly, per project policy.

**Confirmed fixed from the earlier round on this thread:**
- The `criticals` → `critical` typo in `_out_combi_count_critical`'s brief is fixed; Spell Check with Typos now passes.
- The unused-parameter nit in the new static callbacks (`cef_func_out_combi`, `cef_func_out_combi_dynamic`) is addressed — both now wrap their unused parameters in `UNUSED(...)`.

Findings: none outstanding. No High/Medium/Low/Nit issues found in this pass beyond what earlier rounds already surfaced and this head has since resolved.

**Overall verdict: OK to merge.** All 20 CI checks are green (including both Valgrind jobs), the fix's reference-counting and combi-list-ordering logic were independently re-traced and additionally verified by building and running the test suite before/after the fix (confirming the regression tests actually catch the described bug), and no regressions were found in adjacent suites.

myungjoo:
While reviewing this PR, a claim came up that `_gst_tensor_filter_transform_update_outbuf()` leaks one `GstMemory` per output tensor per frame on the dynamic-invoke path: that for a sub-plugin whose `allocate_in_invoke` is 0 (custom-easy sets `fw_info->allocate_in_invoke = 0`), `_gst_tensor_filter_transform_prepare_output_tensors()` has already put a `gst_allocator_alloc ()` memory in `out_trans_data->mem[i]`, and that `out_trans_data->mem[i] = gst_tensor_meta_info_append_header (&meta, mem)` (`tensor_filter.c:1165`) then overwrites it without a `gst_memory_unref ()`. The claim was about pre-existing code, not about this PR, but it names a line inside the function this PR changes, so it was checked before merging.

**It is not a leak: that combination cannot occur.** `trans_data->allocate_in_invoke` is not the sub-plugin's `fw_info->allocate_in_invoke`; it is what `gst_tensor_filter_allocate_in_invoke ()` returns (`tensor_filter.c:925`), and that function returns TRUE for a dynamic invoke before it ever looks at the framework (`tensor_filter_common.c:788`). So with `invoke-dynamic=true` the `if (!trans_data->allocate_in_invoke)` branch at `tensor_filter.c:971` is skipped, nothing is pre-allocated, and `mem[i]` is still the NULL of the `g_new0 ()`. The mirror case is closed too: when `prop->invoke_dynamic` is FALSE the branch holding line 1165 does not run at all.

Verified at head `ca68d970`, not only by reading:

1. **Instrumented run.** Printing the values in that branch while the dynamic tests run gives `allocate_in_invoke=1, preallocating=0` and `old mem[0]=(nil)`, `old mem[1]=(nil)` immediately before the assignment. There is nothing to unref.
2. **ASan.** The 12 tests that use a dynamic invoke (`tensorFilterOutputCombination.*` plus `tensorFilterFlexInput.dynamic*`) pass with no leak whose stack touches `tensor_filter.c`, `gst_allocator_alloc` or `prepare_output_tensors`. What remains is the GStreamer registry allocation that main reports as well.
3. **Positive control.** Adding one deliberate `gst_allocator_alloc (NULL, 4096, NULL)` to that same branch makes ASan report it as `_gst_tensor_filter_transform_update_outbuf ../gst/nnstreamer/tensor_filter/tensor_filter.c:1165`. The detection works, so the clean result above is meaningful.

The related suggestion, that `prepare_output_tensors ()` should skip the allocation and map for a dynamic invoke, is what the code already does.

The other half of the claim, that a memory allocated when `allocate_in_invoke` is FALSE is unmapped but never unreffed on the success path, is also not a leak: `gst_tensor_buffer_append_memory ()` takes ownership of what it appends, and an output left out of an `output-combination` list is released in the loop before that.

No change was made for this. Posting it so the analysis is on the record with the PR whose review raised it.

myungjoo:
This is a relayed review written by another agent (Claude Sonnet), posted on the user's behalf.

Reviewed at head `6c1f9497` (force-pushed after the earlier `myungjoo-bot` review at `ca68d970`, which approved with one non-blocking NIT).

**What changed since `ca68d970`.** I diffed the two heads directly (`git diff ca68d970 6c1f9497`): `tensor_filter.c` and `README.md` are byte-identical. The only change is in `tests/nnstreamer_filter_custom/unittest_filter_custom.cc`, inside `_out_combi_run()`: the `g_log_set_handler("GStreamer", ...)` call (whose `user_data` is the stack variable `critical`) now runs *after* `ASSERT_EQ (_out_combi_register (), 0)` / `_out_combi_dynamic_harness()` setup, instead of before it.

**The last open NIT is resolved.** The `myungjoo-bot` review at `ca68d970` flagged exactly this: if the `ASSERT_EQ` on registration failed while the handler was already installed, the function returns early without `g_log_remove_handler()`, leaving GLib holding a callback pointing at a dead stack frame — a later critical in another test could then write through a dangling pointer. Moving the `g_log_set_handler()` call below the registration/harness setup (confirmed by reading the current file at `tests/nnstreamer_filter_custom/unittest_filter_custom.cc:1595-1611`) means no handler is installed until setup has already succeeded, so that failure path can no longer leave a dangling handler. The counter itself still works: `handler` is only read after being assigned, and the function's own self-check (`g_log("GStreamer", G_LOG_LEVEL_CRITICAL, "self-check of the counter"); EXPECT_EQ(critical, 1U);`) still runs unchanged at the end, so the reorder didn't silently break the "handler is live" verification.

**Core fix, re-checked against the current file.** I re-read `_gst_tensor_filter_transform_update_outbuf()` at its current location (`gst/nnstreamer/tensor_filter/tensor_filter.c:1070-1203`) directly rather than relying only on the diff: the per-output preparation/release-if-unlisted loop (1113-1182) is unchanged from before this PR; the list-order append pass (1184-1194) does exactly one `gst_memory_ref()` per `out_combi_o` list entry, skipping indices `>= prop->output_meta.num_tensors`; the final pass (1196-1199) drops exactly one preparation `gst_memory_unref()` per unique listed index. This matches what the four earlier relayed reviews on this thread hand-traced and what one of them additionally verified by building the PR head, running the new suite, reverting only `tensor_filter.c` to confirm 6 of the new tests fail without the fix, and running under ASan. I have nothing to add to that analysis; it still holds against the current head since the production code is unchanged.

**Sanity-checked the separate leak claim on this thread.** The most recent `myungjoo` comment concludes the dynamic-invoke path at `tensor_filter.c:1165` (`out_trans_data->mem[i] = gst_tensor_meta_info_append_header(&meta, mem)`) does not leak a pre-allocated memory, because `trans_data->allocate_in_invoke` is `gst_tensor_filter_allocate_in_invoke()`'s return value, not the subplugin's own flag. I checked this against `tensor_filter_common.c:783-804`: `gst_tensor_filter_allocate_in_invoke()` returns `TRUE` unconditionally when `priv->prop.invoke_dynamic` is set, before it ever reads `priv->fw`/`priv->info`. That makes `tensor_filter.c:971`'s `if (!trans_data->allocate_in_invoke)` pre-allocation branch dead for a dynamic invoke, so `mem[i]` is still `NULL` (from `g_new0`) when line 1165 assigns it. The claim is correct and this PR's diff isn't implicated either way (the line predates this PR and the analysis is informational, not a code change).

**Regression risk / scope / tests:** unchanged from the prior rounds' conclusions — `out_combi_i`/`out_combi_o` are only read in `tensor_filter.c`/`tensor_filter_common.c`, `tensor_filter_single.c` never touches output-combination, and the non-combi append path is untouched. Size is proportionate (`+26/-2` production code, one README line, the rest is tests). No CHANGES entry, correctly, per project policy.

Findings:
- **[Low, resolved]** The dangling-log-handler NIT from the `ca68d970` bot review is fixed on this head, confirmed by direct diff and by re-reading the reordered function.
- **[Nit] CI is fresh on this force-pushed head.** At review time only `Analyze (python)` and `DCO` had completed; Valgrind (both x86 and arm), Spell Check, Static checks, CPP-Linter, and the GBS/Yocto/RISC-V/pdebuild jobs were still pending/queued since the push had just landed. Nothing in this specific change (a test-only reordering) is expected to affect any of them, but please confirm the full matrix goes green — especially the two Valgrind jobs, since this suite exercises buffer/memory-lifetime code — before merging.

**Overall verdict: OK to merge**, once the freshly-triggered CI matrix (particularly both Valgrind jobs) finishes green. No High/Medium/Low issues remain open; the previously outstanding NIT is resolved and no new issue was introduced by this head's only change.

myungjoo:
Approved by Astra

PR #4986, opened 2026-09-17
[Element/Decoder] Keep sub-plugin private data alive while a call uses it, and release it on a same-mode set

myungjoo:
- **What:** `tensor_decoder` can use a sub-plugin's private data after it has been released (#4920 **C38**, use-after-free). The race is between a caps query, `getTransformSize ()` or `decode ()` on one side and a `mode` change, an option change or a mid-stream renegotiation on the other. Separately, setting `mode` to the decoder the element already has leaks the previous private data (#4920 **C21**).
- **How:** the private data is now reference counted. Each sub-plugin call takes a reference under a short lock and calls the sub-plugin without holding it. Mode sets, option sets and renegotiation build new private data outside the lock (`init ()` plus the stored options) and swap it in. The old data is released when its last call returns. If the stream is already negotiated, a `mode`/option set first gives the new data `getOutCaps ()` with the negotiated config and then renegotiates the source pad, because some sub-plugins (`bounding_boxes`, `tensor_region`) set up what `decode ()` needs only in `getOutCaps ()`. A generation counter drops a rebuild that a newer one overtook, and options are replayed in the order the old in-place `setOption ()` calls applied them. A plain mutex across the calls was tried first and dropped: the python3 decoder takes the GIL inside `decode ()`, and PyGObject keeps the GIL during `set_property`, so a Python app would deadlock.
- **Result:** 17 new gtest cases in a new binary `unittest_decoder_plugin_data` (no flatbuffers dependency, so the Valgrind CI job runs it); 6 of them fail on main (released data read by an in-progress call, a same-mode set that never releases, an overlapping option set that loses an option), and 5 fail without the getOutCaps/renegotiation step. Mutations: dropping the `getOutCaps ()` call fails 2 cases, and dropping the renegotiation fails 2 others. Local runs (docker, jammy): `unittest_decoder_plugin_data`, `unittest_decoder`, `unittest_plugins`, `unittest_decoder_{boundingbox,pose,image_segment}`, `unittest_tensor_region` and `unittest_converter_serialized` all pass, on main `40f3246e`. The SSAT suites `nnstreamer_decoder_{boundingbox,pose,image_labeling,tensor_region,image_segment,python3}` pass. The new binary under `run_unittests_binaries.sh --valgrind` + `check_valgrind_log.sh` reports 0 error contexts and 0 definite leaks.

### Commits
1. `[Element/Decoder] Release the private data when the same mode is set again` (C21 plus 3 cases; builds and passes on its own)
2. `[Element/Decoder] Keep a sub-plugin's private data alive while a call uses it` (C38 plus 14 cases)

### Behavior notes for users
- A `mode`/`optionN` set no longer affects a sub-plugin call already in progress; that call finishes with the private data it started with, and the next call uses the new data.
- **An option set now re-creates the sub-plugin's private data**, the way a mode set already did, instead of calling `setOption ()` on the live data. On a negotiated stream the new data gets `getOutCaps ()` first, so decoding goes on, and the element renegotiates, so an option that changes the output caps now reaches downstream. State a sub-plugin builds up in `decode ()` across buffers (e.g., bounding-box tracking, a python3 script's state) starts over. If the sub-plugin refuses the negotiated config after the change, the stream stops as not negotiated instead of decoding with it. The order options reach the sub-plugin is unchanged: an option set before `option1` is still dropped by `bounding_boxes`.
- Setting the same `mode` again re-creates the private data and replays the options. Before, `init ()` ran over the old data without releasing it.

---

<details><summary>Context for agents / reviewers</summary>

- Scope: `gst/nnstreamer/elements/gsttensor_decoder.{c,h}`, the new `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc` and its `tests/meson.build` entry. The tests are not in `unittest_decoder.cc` because that binary is only built with flatbuffers, which the Valgrind job does not install (review round 2). The header is not installed (not in `nnstreamer_headers`), so the struct change breaks no ABI. `GstTensorDecoderPlugin` is opaque outside the .c.
- Locking: `plugin_lock` guards `decoder`, `plugin`, `plugin_gen`, `option[]` and `option_order[]`. It is never held across a sub-plugin call, a log call or a pad query. `exit ()` of old data can therefore run on whichever thread drops the last reference (possibly the streaming thread, after `decode ()` returns). The python3 decoder's `exit ()` takes the GIL itself, and no element lock is held at that point.
- Rebuild semantics: `gst_tensordec_plugin_rebuild (self, opnum)`, where `opnum == TensorDecMaxOpNum` means a mode set or renegotiation. That resets `option_order` to 0..8, which is what main replayed. An option set moves its index to the end of `option_order`. Main applied options in place in set order, so the replay reproduces main's final state for set-type options. SSAT `nnstreamer_decoder_boundingbox` 18_n (option5 before option1 must be refused) failed with a plain 0..8 replay and passes with this order. An option set on a decoder without `setOption` (or with no mode yet) does not rebuild.
- Test mechanism: sub-plugins `lifedec`/`lifedec2` share callbacks. `exit ()` marks the data `0xdeadbeef` and keeps it until TearDown, so reading released data needs no real UAF. A hold point captures `*pdata` on entry and blocks until released. `setOption ("replace")` swaps its own data, like python3 loading a script. The caps query goes through `GstBaseTransformClass::transform_caps` directly. Decode and transform_size use `gst_harness_new_with_element ()` after `mode` is set, because `gst_harness_new ("tensor_decoder")` links pads before a mode exists and aborts.
- Review round 1 (relayed comment): #2 (getter races) is fixed, #3 (`getTransformSize` coverage) has a case, and #5 (PyGObject/GIL) turned out to be a real deadlock, which led to the redesign above. #1 (compatibility check vs. swap in configure) is not a memory-safety issue: a concurrent mode change in that gap installs new data, and `set_caps` checks the result against the current decoder. #4 (custom-code callback pointer) predates this PR and is out of scope.
- Review round 2 (relayed comment): the Valgrind-coverage finding is fixed by the move above. The nit that a failed `init ()` still installs the wrapper is kept as main behaves (main also kept the decoder and called it with whatever `init ()` left).
- Review round 7 (`myungjoo-bot`, CHANGES_REQUESTED): **blocking** item fixed as above. An option set left `bounding_boxes`/`tensor_region` with `max_detection = 0` because nothing called `getOutCaps ()` on the new data. Fix: `gst_tensordec_plugin_rebuild_on_property ()` parses the current sink caps, passes the config to `gst_tensordec_plugin_rebuild ()` (which calls `getOutCaps ()` on the new data before installing it) and calls `gst_base_transform_reconfigure_src ()`. Renegotiation from `configure ()` passes NULL, because `fixate_caps`/`set_caps` already call `getOutCaps ()` on the new data. Nit 1 (test brief) is fixed. Nit 2 (`plugin == NULL` branch in `transform ()`) is not tested: it is only reachable through a race between the `self->decoder` check and the ref, with no hook to hold it deterministically.
- Environment notes: SSAT `nnstreamer_decoder` fails the same 14 cases on main and with the change in the local container (missing golden inputs). `tensorDecoderCustom.normal0` needs the `meson test` env.

</details>


myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable after nits.** The core fix for #4920 C38 (UAF race between a caps query/`decode()` and a `mode`/option change or renegotiation) and C21 (leaked private data when the same `mode` is set again) is correct and each is backed by a regression test that I traced by hand against `main` and confirmed fails without the fix and passes with it (`setSameModeReleasesPrevious` for C21; `capsQueryHoldsModeChange`, `capsQueryHoldsOptionChange`, `decodeHoldsModeChange` for C38). The lock design (plain `GMutex`, no in-tree sub-plugin calls back into GStreamer pad/element/property APIs — verified with `grep -rln "gst_pad_\|gst_element_\|g_object_set\|g_object_get" ext/nnstreamer/tensor_decoder/` returning nothing) is sound and `gsttensor_decoder.h` is confirmed not in `nnstreamer_headers` (gst/nnstreamer/meson.build), so the struct-layout change breaks no ABI. Scope is well contained to `gst/nnstreamer/elements/gsttensor_decoder.{c,h}` and its own unit test. The items below are residual gaps in the same change, not blockers, but worth addressing (either in this PR or as an immediate fast-follow).

1. **Medium — residual TOCTOU in `gst_tensordec_configure()`'s renegotiation branch.** `gst/nnstreamer/elements/gsttensor_decoder.c:684` and `:702-713`. `gst_tensordec_get_media_caps_from_config()` at line 684 takes and releases `plugin_lock` internally around `getOutCaps()`, and only afterwards does the code re-take the lock at line 702 for `clean_plugin`+`init`. Between those two acquisitions the lock is not held, so a concurrent `mode` change on another thread can swap `self->decoder` in the gap: the "compatible" caps decision (and thus `self->tensor_config`, written right after at line 717) ends up computed from the *old* decoder's `getOutCaps()` output while `clean_plugin`/`init` at line 703-706 act on whatever `self->decoder` has become by then (the new one). This can't reintroduce a use-after-free (each access is itself lock-protected), but it can leave the element's negotiated `tensor_config` inconsistent with the sub-plugin that is actually active — exactly the "mode change races an in-flight decoder call" class of bug this PR targets, just not fully closed for this one path. No test exercises this window (all the new concurrency tests race a `mode`/option set against `transform_caps()`/`decode()` directly, not against `fixate_caps`/`set_caps` → `configure()`). Suggested fix: hold `plugin_lock` across the whole decision (capture `self->decoder` once under the lock, call `getOutCaps` directly instead of via the double-locking helper to avoid re-entering the non-recursive mutex, then do clean+init still under the same lock), or otherwise ensure the compatibility check and the clean/init that follows it use the same decoder snapshot.

2. **Medium — `gst_tensordec_get_property()`'s `PROP_MODE` case dereferences `self->decoder` without the new lock.** `gst/nnstreamer/elements/gsttensor_decoder.c:593-600` (unchanged by this diff, but reads the exact field this PR added `plugin_lock` to protect). `set_property(mode=<unrecognized string>)` now nulls `self->decoder` under `plugin_lock` (line 511-543), but `get_property("mode")` still does `else if (self->decoder) g_value_set_string(value, self->decoder->modename);` with no lock at all. A thread calling `g_object_get(dec, "mode", ...)` concurrently with another thread setting an invalid/empty mode can pass the `if (self->decoder)` check and then dereference a `self->decoder` that has just been set to `NULL` by the other thread — a NULL-pointer-dereference crash. This is a narrow window (needs a concurrent get + an invalid concurrent set) but it is a real, newly-easy-to-hit gap now that `self->decoder` is documented as needing synchronized access. Suggested fix: wrap this branch's read of `self->decoder` (and ideally the whole `get_property`) with `plugin_lock` for consistency with the setter.

3. **Low/Medium — no regression test for the `getTransformSize()` lock path.** `gst/nnstreamer/elements/gsttensor_decoder.c:1101-1124`. The same defensive pattern applied to `getOutCaps()`/`decode()` (lock + re-check `self->decoder` under the lock) was also applied here, but unlike `capsQueryHoldsModeChange`/`decodeHoldsModeChange`, there is no `lockdec`-based test that holds inside `getTransformSize()` and asserts a concurrent `mode` change blocks on it. Since `lockdec`/`lockdec2` don't register a `getTransformSize` callback at all, this path is currently unverified by the new test suite and relies only on code-review symmetry with the other two call sites. Suggested fix: add a `getTransformSize` hold point to the `lockdec` fixture and a `transformSizeHoldsModeChange`-style test, mirroring `decodeHoldsModeChange`.

4. **Nit — the custom-code (`mode=custom-code`) path is only partially covered by the new lock.** `gst/nnstreamer/elements/gsttensor_decoder.c:914-931`. The lock added here protects the read of `self->option[0]` and the `get_subplugin()` lookup, but `self->custom.func = ptr->func; self->custom.data = ptr->data;` (lines 930-931) are written *after* the `g_mutex_unlock`, and `gst_tensordec_transform()`'s read of `self->custom.func` (line 862) is never under `plugin_lock` at all. This is pre-existing (not a regression from this PR) and out of scope for C21/C38, which are specifically about the registered-sub-plugin path, but it's easy to misread the new lock as now covering all decoder-related state. A one-line comment noting that `self->custom` is intentionally out of this lock's scope would save the next reader some time.

5. **Nit — the deadlock-avoidance argument for the python3 sub-plugin (PR description) rests on external library behavior not checkable in this repo.** The PR body argues no deadlock occurs because "PyGObject releases the GIL around property sets and gi calls, so an app thread in Python does not hold the GIL while it waits on this lock," while `tensordec-python3.cc`'s `decode()`/`getOutCaps()` take a `PyGILGuard` *while* `plugin_lock` is held (confirmed at `ext/nnstreamer/tensor_decoder/tensordec-python3.cc:285,318,352,369`). That specific ordering (GIL held while waiting to be granted `plugin_lock` vs. `plugin_lock` held while acquiring the GIL) is exactly the classic AB/BA deadlock shape, and its avoidance depends entirely on PyGObject's calling convention rather than anything in nnstreamer. Worth a short comment at the `plugin_lock` declaration (`gst/nnstreamer/elements/gsttensor_decoder.h:87`) capturing this assumption explicitly, so it's not lost to the PR description if the file is touched again later.

Test-policy / doc-policy checks: the two bug fixes each have a regression test that reproduces the bug per CLAUDE.md's test-case policy; the new tests compile against dependencies already present in `tests/meson.build` (`gst_check_dep` is already part of `nnstreamer_unittest_deps`, and other unit tests already use `gst/check/gstharness.h`); `gsttensor_decoder.h` is not an installed/public header, so no ABI break and no documentation update is required for this change; `CHANGES` was correctly left untouched.

— Relayed by MyungJoo Ham on behalf of a Claude review agent.

myungjoo:
Relayed on behalf of the author (Claude): response to the review above, now at head `58283811`.

1. **configure TOCTOU (Medium):** not changed. The gap between the compatibility check and the swap is not a memory-safety issue. A concurrent mode change there installs its own new private data, and `set_caps ()` checks the result against the decoder that is current by then. Holding a lock across both steps would put `getOutCaps ()` under the lock, which is exactly what the new design avoids (see 5).
2. **get_property races (Medium):** fixed. `PROP_MODE` and every `optionN` read now take the lock. The setter frees the option string under the same lock.
3. **getTransformSize coverage (Low/Medium):** added `transformSizeKeepsDataOnModeChange`.
4. **custom-code callback pointer (Nit):** left as is. It predates this PR, and `custom.func/data` point at registry entries this element does not free.
5. **PyGObject / GIL (Nit):** checked PyGObject's source. `pygi_set_property_value ()` calls `g_object_set_property ()` **without** releasing the GIL (only the get path and gi-invoked functions release it). With the python3 decoder taking the GIL inside `decode ()`, the mutex design could deadlock a Python app that sets a decoder property mid-stream. So the PR was redesigned instead of documented: the private data is reference counted, no lock is held across any sub-plugin call, and setters build new data and swap it in. The details are in the updated description and the second commit message.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable after nits.** I re-verified the redesigned reference-counted approach (head `58283811`) independently of the round-1/round-2 discussion already on this thread, and it holds up:

- Every sub-plugin call (`getOutCaps`, `decode`, `getTransformSize`, `setOption`, `init`, `exit`) happens with `plugin_lock` **not** held (`gst_tensordec_plugin_unref`/`_ref`/`_rebuild` all release the lock before touching `plugin->decoder->*`). So the AB/BA deadlock the PR body describes for the python3 sub-plugin (PyGObject keeps the GIL across `g_object_set` → `pygi_set_property_value` → `g_object_set_property`, confirmed by `PyGILGuard` in `ext/nnstreamer/extra/nnstreamer_python3_helper.h:102-114` wrapping plain `PyGILState_Ensure`/`Release`, which is documented as safe to call re-entrantly on a thread that already holds the GIL) genuinely cannot occur with this design.
- I hand-traced the refcount/generation swap in `gst_tensordec_plugin_rebuild()` (`gst/nnstreamer/elements/gsttensor_decoder.c:69-131`) against the `overlappingOptionSets` and `capsQueryKeepsDataOnModeChange` tests, including the case where a slow rebuild's own freshly-built (but stale) plugin gets discarded via `old = plugin` when `gen != self->plugin_gen` — it is correctly unreffed (not leaked), and a call already holding a ref via `gst_tensordec_plugin_ref()` keeps its private data alive independent of any concurrent swap, because the pointer read and the refcount increment happen atomically under the same lock the swap uses.
- I confirmed by reading `tests/nnstreamer_decoder/unittest_decoder.cc`'s `lifedec_enter()` that the tests genuinely discriminate: on `main`, `getOutCaps`/`decode`/`getTransformSize` are called with `&self->plugin_data` directly, so a concurrent `exit()` mutates the very struct the in-progress call captured a pointer to (visible as `magic == LIFEDEC_DEAD` instead of `LIFEDEC_ALIVE`); with the fix, the call holds its own reference to a separate `GstTensorDecoderPlugin`, so it never observes this. This is a real, not simulated, reproduction of the UAF class of bug.
- Scope is properly contained (`gsttensor_decoder.{c,h}` + its own test file only); `gsttensor_decoder.h` is confirmed not in `nnstreamer_headers` (no ABI break); no other file references the renamed `plugin_data` field.

Given that, I have one substantive new finding and one nit.

---

**1. Medium — the new regression tests are not built by the one CI job that runs Valgrind, so a future re-introduction of this exact UAF may not be caught there.** `tests/nnstreamer_decoder/unittest_decoder.cc` (all 12 new `tensorDecoderPluginData.*` cases) is only built `if flatbuf_support_is_available` (`tests/meson.build:359-368`), and that flag is derived from whether `flatc`/`flatbuffers` are found (`meson.build:168-184`, `feature: 'auto'`). `.github/workflows/ubuntu_clean_meson_build.yml` — the "Build with meson and test with Valgrind in Ubuntu" job, which is the job in this repo specifically dedicated to catching memory-safety regressions — never installs `flatc`/`flatbuffers-dev` in its `apt-get install` steps, so `flatbuf_support_is_available` is false there and `unittest_decoder` (hence every case added by this PR) is silently skipped. The same is true of `.github/workflows/ubuntu_clean_llvm_build.yml`. The tests do still run somewhere in CI — Debian `debian/rules:88-90` (`override_dh_auto_test` → `run_unittests_binaries.sh ./tests`) and the Tizen GBS `x86_64, unit_test=1` leg (`packaging/nnstreamer.spec:1083` → same script) both carry `libflatbuffers-dev`/`flatbuffers-devel` as unconditional build-deps — so a hard crash or assertion failure would still fail those legs and block merge. But neither of those legs runs under Valgrind, so a subtler race that corrupts memory without an immediate crash (exactly the class of bug #4920 is about) could pass pdebuild/GBS's plain `meson test` yet would have been caught by the Valgrind-instrumented job. This is the same trap the project hit before on this exact file: when item **C24** was fixed for `gsttensor_decoder.c`, its regression tests were deliberately placed in `tests/nnstreamer_plugins/unittest_plugins.cc` instead, specifically because that binary has no feature guard (`tests/meson.build:134-141`) and is already covered by the Valgrind job (it already contains `TEST(testTensorDecoder, ...)` cases, e.g. `tests/nnstreamer_plugins/unittest_plugins.cc:15145` onward). Suggested fix: move these tests (or at least the ones that exercise the UAF window — `setSameModeReleasesPrevious`, `capsQueryKeepsDataOnModeChange`, `decodeKeepsDataOnModeChange`, `transformSizeKeepsDataOnModeChange`, `overlappingOptionSets`) into `unittest_plugins.cc`, or add flatbuffers to the Valgrind job's dependency list.

**2. Nit — `gst_tensordec_plugin_rebuild()` still installs the wrapper when `decoder->init()` fails.** `gst/nnstreamer/elements/gsttensor_decoder.c:93-117`: if `decoder->init (&plugin->data)` returns 0, the code only logs (`ml_loge`) and falls through to install `plugin` (with `plugin->data` left zeroed by `g_new0`) into `self->plugin` anyway, rather than discarding it. This matches `main`'s existing observable behavior (it also left `self->plugin_data == NULL` and still let `decode()`/`getOutCaps()` be called with it), so it is not a new regression, but a one-line comment noting that the wrapper is intentionally kept even on init failure would save a future reader from assuming otherwise.

Test-policy / doc-policy checks otherwise look right: both bugs (C21 leak, C38 UAF) have hand-traceable regression tests per CLAUDE.md's test-case policy, `gsttensor_decoder.h` is not an installed header so no doc update is required, and `CHANGES` was correctly left untouched.

🤖 Relayed by a Claude review agent.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-2 review above, now at head `2b782e0c`.

1. **Valgrind job never builds the new tests (Medium):** confirmed (`ubuntu_clean_meson_build.yml` installs no flatbuffers, so `unittest_decoder` is skipped there) and fixed. The cases moved to a new binary, `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc`, built unconditionally next to `unittest_decoder_image_segment` in `tests/meson.build`. `unittest_decoder.cc` is no longer touched. Checked locally with the job's own helpers: `run_unittests_binaries.sh --valgrind ./tests/unittest_decoder_plugin_data` + `check_valgrind_log.sh --require-output` gives 12 tests passed and "0 error contexts from this repository ... 0 definite leak contexts from this repository, 1 binaries examined". Both commits still build and pass on their own.
2. **Failed `init ()` still installs the wrapper (Nit):** kept on purpose. Main behaves the same (it kept `self->decoder` and handed later calls whatever `init ()` left in `plugin_data`). Changing what a failed `init ()` means for each in-tree sub-plugin is outside C21/C38.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This is an independent third pass at head `2b782e0c`, done without assuming the two prior rounds already on this thread are correct, and it reaches the same conclusion the author's replies do: both #4920 items are fixed, the fix is well-scoped, and the outstanding points from rounds 1–2 (GIL deadlock avoidance, `get_property` lock coverage, `getTransformSize` test coverage, and moving the tests into a binary the Valgrind job actually builds) are genuinely resolved in this diff, not just claimed. I re-derived each of the following independently rather than trusting the earlier write-ups:

- **Refcount/generation correctness under concurrent rebuilds.** Traced `gst_tensordec_plugin_ref/_unref/_rebuild` (`gst/nnstreamer/elements/gsttensor_decoder.c:210-310`) by hand for the interleaving where a call takes a ref under `plugin_lock` *before* a concurrent rebuild's swap, and the interleaving where two rebuilds race (the older one's freshly-built plugin is correctly discarded via `old = plugin` when `gen != self->plugin_gen`, unreffed once, never leaked or double-freed). Because `self->plugin` is only ever read (in `_ref`) and only ever written (in `_rebuild`'s swap) under the same mutex, `_ref` can never observe a pointer that a concurrent `_rebuild` has already unreffed — the unref of the old pointer always happens strictly after the mutex section that installed the new one, so there is no window for a fresh `_ref` to hand out a reference to memory that is about to be (or already was) freed.
- **No lock held across any sub-plugin call.** Confirmed `init`, `exit`, `setOption`, `getOutCaps`, `decode`, and `getTransformSize` are all invoked with `plugin_lock` unlocked (grep for `decoder->` / `plugin->decoder->` calls against the lock/unlock lines). This is what actually closes the PyGObject/GIL deadlock risk raised in round 1 — not a documentation note, an actual absence of the hazardous lock ordering. `ext/nnstreamer/tensor_decoder/tensordec-python3.cc:279-288`'s `decoder_py_exit()` takes the GIL via `PyGILGuard`, and since `exit()` can now run on whichever thread drops the last reference (including a streaming thread that never held the GIL before), that's fine: `PyGILState_Ensure/Release` is documented as safe to call from any thread, Python-created or not.
- **CI/packaging coverage of the new binary is real, not superficial.** `tests/meson.build:133-140` builds `unittest_decoder_plugin_data` unconditionally (no `flatbuf_support_is_available` guard, unlike `unittest_decoder`), so `.github/workflows/ubuntu_clean_meson_build.yml`'s Valgrind job — which installs no flatbuffers package — will build and run it via `run_unittests_binaries.sh --valgrind --recursive ... ./tests/`. I also checked that neither `debian/rules:90` nor `packaging/nnstreamer.spec`'s `%files unittests` (`.../unittest-nnstreamer/tests`, a directory glob) enumerate test binaries by name, so no packaging file needs a matching update for the new binary — the concern the task description asked me to specifically check turned out not to apply here, because both packaging paths already install/run the whole `tests/` tree rather than a fixed list.
- **No ABI/doc impact.** `gsttensor_decoder.h` is confirmed not listed in `nnstreamer_headers` (`gst/nnstreamer/meson.build`), so the `plugin_data` → `plugin` (`GstTensorDecoderPlugin *`) field change and the new `plugin_gen`/`plugin_lock`/`option_order` fields are all private layout changes with no public API/ABI surface, and no documentation update is owed.
- **Test-policy compliance.** Both bugs (C21 leak, C38 UAF) ship regression tests I could trace to fail on `main` and pass with the fix (`setSameModeReleasesPrevious` for C21; `capsQueryKeepsDataOnModeChange`/`decodeKeepsDataOnModeChange`/`transformSizeKeepsDataOnModeChange`/`overlappingOptionSets` for C38), satisfying CLAUDE.md's regression-test requirement. `CHANGES` is correctly untouched per project policy.

One optional, non-blocking nit:

1. **Low/nit — no explicit `timeout:` on the new test entry.** `tests/meson.build:140`: `test('unittest_decoder_plugin_data', unittest_decoder_plugin_data, env: testenv)` relies on meson's default 30s timeout, while the binary spawns multiple `GThread`s per test case and several assertions wait up to `TEST_TIMEOUT_MS` (5000 ms, defined at line 18) before giving up (e.g. `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc:395,398` in `capsQueryKeepsDataOnModeChange`). On the happy path this is fast (releases happen immediately), so this is not expected to flake today, but if a future regression reintroduces blocking behavior, several of the 12 cases could each burn close to 5s before failing, and on a loaded/valgrind-style runner that could approach the default timeout. `unittest_join`, the other GThread-heavy test in this same file, is given an explicit `timeout: 120` for a similar reason (`tests/meson.build:222`). Consider the same here as a follow-up; not a reason to hold this PR.

— Relayed by a Claude review agent on behalf of the PR author.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-3 review above, now at head `ed312f63`.

- **No explicit `timeout:` for the thread-based binary (Nit):** taken. `test('unittest_decoder_plugin_data', ..., timeout: 120, env: testenv)`, folded into the commit that adds the entry. Compared with `2b782e0c`, that one `tests/meson.build` line is the only change (`git diff 2b782e0c ed312f63` is 1 line), so a regression that makes the 5 s waits in these cases run out ends in gtest assertion output rather than a meson timeout.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This is an independent fourth pass at head `ed312f63`, done by re-deriving the design from the diff and the surrounding code rather than trusting the three prior rounds already on this thread, and it reaches the same conclusion round 3 did: both #4920 items (C38 use-after-free, C21 leak on a same-mode set) are fixed, the reference-counted design correctly avoids the PyGObject/GIL deadlock, and the fix is well contained.

What I re-verified myself:
- **`git diff 2b782e0c ed312f63`** is exactly the one-line `timeout: 120` addition to `tests/meson.build` the author's last reply claims — confirmed by direct diff, not just re-reading the reply.
- **Refcount/generation swap.** Hand-traced `gst_tensordec_plugin_ref/_unref/_rebuild` (`gst/nnstreamer/elements/gsttensor_decoder.c:210-330`) for concurrent rebuilds with interleaved generations: only the rebuild whose captured `gen` still equals `self->plugin_gen` at its second lock installs; every other one's freshly-built plugin is put into `old` and unreffed through the same atomic-refcount path used by ordinary callers, so a losing rebuild is always released exactly once and never double-freed or leaked. I traced this specifically against `overlappingOptionSets` (two option sets racing) and confirmed the later-generation build ends up with both options applied to one `data`, matching the test's expectation.
- **No lock held across a sub-plugin call.** Confirmed by inspection that `init`, `exit`, `setOption`, `getOutCaps`, `decode`, and `getTransformSize` are all called with `plugin_lock` unlocked, which is what actually removes the GIL-deadlock risk (not just PR-body documentation of it).
- **Scope/regression risk to other modules.** `grep -rl "plugin_data\|GstTensorDecoderPlugin"` across the tree turns up nothing outside `gsttensor_decoder.{c,h}` and its own new test file — the handful of unrelated `plugin_data`/`subplugin_data` hits in `tensordec-boundingbox.cc`, `tensordec-imagelabel.c`, and the tensor_filter files are a different, pre-existing naming convention (the sub-plugin's own `void **pdata` argument, or tensor_filter's unrelated `subplugin_data`), not this struct. `gsttensor_decoder.h` is confirmed not in `nnstreamer_headers` (`gst/nnstreamer/meson.build`), so no ABI break. Net: no other module can regress from this change.
- **CI actually gates on this.** `tests/meson.build:133-140` builds `unittest_decoder_plugin_data` unconditionally (no flatbuffers guard), and `test(...)` registers it as an ordinary meson test, so it runs under every `meson test` invocation — not just the Valgrind job. `.github/workflows/ubuntu_clean_meson_build.yml` (the Valgrind job) and `.github/workflows/ubuntu_clean_llvm_build.yml` are both `on: pull_request` against `main`, so a reintroduction of either bug fails a required per-PR check, not just a nightly build.
- **Prior findings I re-checked and still agree are correctly resolved, not just claimed:** the `get_property`/`PROP_MODE`/`PROP_READ_OPTION` locking is present in the diff (`gst/nnstreamer/elements/gsttensor_decoder.c:689-696`, `:668-670`); `transformSizeKeepsDataOnModeChange` exists in the test file and genuinely holds inside `getTransformSize()`; the configure()-renegotiation TOCTOU (round 1, item 1) is still present (`gst_tensordec_get_media_caps_from_config()` at line ~778 and `gst_tensordec_plugin_rebuild()` at ~796 are two separate lock sections) but I confirm the author's argument that it cannot reintroduce a UAF — each access is independently lock-protected, and a concurrent mode change there only risks a transient caps/decoder mismatch that `set_caps()` re-validates against whatever decoder is current by then. I'm not re-flagging it as this thread already settled it with sound reasoning.

One new, non-blocking nit I did not see raised in the prior rounds:

1. **Low/nit — the behavior change to option semantics is not reflected in the component doc.** `gst/nnstreamer/elements/gsttensor_decoder.c`: every `optionN` set now goes through `gst_tensordec_set_option()` → `gst_tensordec_plugin_rebuild()`, which discards the sub-plugin's private data and calls `init()` fresh (replaying all stored options) instead of calling the old, in-place `setOption()` on the live data. The PR body's "Behavior notes for users" section documents this honestly, but `gst/nnstreamer/elements/gsttensor_decoder.md` (the doc `Documentation/component-description.md:89` links to) does not mention it. This matters concretely for `bounding_boxes` (`ext/nnstreamer/tensor_decoder/tensordec-boundingbox.cc`), whose private `BoundingBox` object keeps naive centroid-based tracking state (`tracking_id`, see lines ~610-706) across `decode()` calls when `is_track` is enabled: with this PR, any `optionN` set issued mid-stream (even one unrelated to tracking) now silently resets that tracking history, where previously only a `mode` change did. Existing pipelines that adjust a decoder option at runtime while relying on tracking continuity would see a new, silent behavior change. Suggested fix: add a short note to `gsttensor_decoder.md` (general options section, and/or the `bounding_boxes`/`pose_estimation` rows) stating that changing any option at runtime now re-initializes the sub-plugin's private state. Not a blocker — it's a real, disclosed trade-off, just missing from the one doc a user would actually consult.

Test-policy / doc-policy checks otherwise hold: both bugs (C21 leak, C38 UAF) ship regression tests I could independently trace against `main` and the fix; `gsttensor_decoder.h` is not an installed header so no ABI doc update is owed; `CHANGES` was correctly left untouched per project policy.

🤖 Relayed by a Claude review agent on behalf of the PR author.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-4 review above, now at head `f0c7f19d`.

- **Behavior change missing from the component doc (Low/nit):** taken. `gst/nnstreamer/elements/gsttensor_decoder.md`, section "Decoder sub-plugins", now says that each `mode`/`optionN` set and each tensor-config change gives the sub-plugin new private data (`init ()` plus the stored options, in set order since the last `mode`), that `exit ()` of the old data waits for in-progress `getOutCaps ()`/`getTransformSize ()`/`decode ()` calls, which run with no tensor_decoder lock held, and that cross-buffer state such as `bounding_boxes` tracking restarts on an option set during streaming. Folded into the second commit; `git diff ed312f63 f0c7f19d` is those 2 doc lines only.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This is a fifth, independent pass at head `f0c7f19d`, focused mainly on the one thing that changed since round 4 (`ed312f63` → `f0c7f19d`): the doc addition to `gst/nnstreamer/elements/gsttensor_decoder.md` that round 4 asked for. I verified it line-by-line against the current code rather than trusting the description, and re-skimmed the rest of the diff for anything the first four rounds got wrong. I did not find anything wrong, and one very small, non-blocking observation.

**Doc addition verified accurate.** `git diff ed312f63 f0c7f19d` is exactly the 2-line addition to `gsttensor_decoder.md` claimed in the reply, nothing else changed. Checking each clause of the new paragraph against `gst/nnstreamer/elements/gsttensor_decoder.c` at this head:
- "gives each set of `mode` or `optionN`, and each change of the incoming tensor config, new private data" — confirmed: `gst_tensordec_plugin_rebuild()` is invoked from the `PROP_MODE` setter (line 636, via `gst_tensordec_set_property`), from `gst_tensordec_set_option()` for every `optionN` (line 331), and from `gst_tensordec_configure()`'s renegotiation branch (line 799).
- "it calls `init ()` and then `setOption ()` with the stored options, in the order they were set since the last `mode` (numbered order right after a `mode` set)" — confirmed against `gst_tensordec_plugin_rebuild()` (lines 256-310): `option_order[]` is reset to identity (`self->option_order[i] = i`) only when `opnum == TensorDecMaxOpNum` (a `mode` set or a renegotiation — see the nit below), and is otherwise replayed in the order captured from `self->option_order[]`, which `gst_tensordec_set_option()` (lines 315-332) maintains as "most-recently-set last". Matches the `optionOrder` test (`tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc:1016-1039`): 312 after three out-of-order sets, 123 after a fresh `mode` set.
- "The previous private data is released with `exit ()` once no `getOutCaps ()`, `getTransformSize ()` or `decode ()` call is using it any more, and those calls are made without any lock of tensor_decoder held" — confirmed: `gst_tensordec_plugin_unref()` (lines 210-221) calls `exit ()` only when the atomic refcount reaches 0; every call site (`gst_tensordec_get_media_caps_from_config` line 349/363-364, `gst_tensordec_transform` line 939-944, `gst_tensordec_transform_size` line 1203-1209) takes its ref via `gst_tensordec_plugin_ref()`, invokes the sub-plugin, and unrefs afterward, with `plugin_lock` held only inside `_ref`/`_unref`/`_rebuild`, never across the sub-plugin call itself.
- "any state it keeps across buffers (e.g., the object tracking of `bounding_boxes`) starts over when an option is set during streaming" — this is an accurate restatement of the "Behavior notes for users" section already in the PR body, and matches the mechanism above (every `optionN` set now triggers a full `init ()` + option replay instead of the old in-place `setOption ()`-only call).

No inaccuracy found in the added text.

**One low/nit, not a reason to hold the PR:** the parenthetical "(numbered order right after a mode set)" is precise about the `mode`-set case but doesn't mention that a tensor-config renegotiation (`gst_tensordec_configure()` line 799) also calls `gst_tensordec_plugin_rebuild(self, TensorDecMaxOpNum)` and therefore *also* resets `option_order` to numbered order, discarding whatever custom order the user's live `optionN` sets had built up since the last `mode` set. I checked this against the pre-PR code (`gst_tensordec_configure()` at `81a8c9b1^:gst/nnstreamer/elements/gsttensor_decoder.c:691-699`): it also always replayed options in numbered order (`for (i = 0; i < TensorDecMaxOpNum; i++)`) on a renegotiation, so this is not a regression — it's a faithful continuation of existing behavior, just not spelled out in the new doc sentence. Not worth a revision on its own.

Everything else in the diff (the refcount/generation design in `gst_tensordec_plugin_ref/_unref/_rebuild`, the `plugin_lock` coverage of `get_property`'s `PROP_MODE`/`PROP_MODE_OPTIONn` cases, the unconditionally-built `unittest_decoder_plugin_data` binary and its `timeout: 120` test entry, and the still-accepted, non-UAF TOCTOU window in `gst_tensordec_configure()`'s renegotiation branch) matches what rounds 1-4 already found and the diff at this head has not changed it. I re-derived the refcount/swap correctness and the "no lock held across a sub-plugin call" property directly from the code rather than re-quoting the earlier rounds, and they still hold.

Test-policy / doc-policy checks: both #4920 items (C21 leak, C38 UAF) ship regression tests traceable against `main`; the new test binary is unconditionally built (no `flatbuf_support_is_available` guard) so it runs under the Valgrind CI job; the doc update requested in round 4 is now present and accurate; `CHANGES` is correctly left untouched per project policy; `gsttensor_decoder.h` is still not an installed header, so no ABI/API doc obligation.

— Relayed by a Claude review agent on behalf of the PR author.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-5 review above, now at head `40ae669a`.

- **Doc omits that a tensor-config change also resets option order (observation):** taken. The sentence now says the options go "in the order they were set since the last `mode` set or tensor config change, which themselves give them in numbered order", which matches `gst_tensordec_plugin_rebuild (self, TensorDecMaxOpNum)` being called from both `set_property (mode)` and `gst_tensordec_configure ()`. `git diff f0c7f19d 40ae669a` is that one doc line.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This is a sixth, independent pass at head `40ae669a`, done after re-deriving the fix from the diff and the surrounding code rather than trusting the five rounds already on this thread. It reaches the same conclusion rounds 3-5 did.

What I checked specifically:

- **The one change since round 5 is exactly what it claims to be.** `git diff f0c7f19d 40ae669a` touches only `gst/nnstreamer/elements/gsttensor_decoder.md`, rewording "in the order they were set since the last `mode` (numbered order right after a `mode` set)" to "...since the last `mode` set or tensor config change, which themselves give them in numbered order." I checked this against `gst_tensordec_plugin_rebuild()` (`gst/nnstreamer/elements/gsttensor_decoder.c:247-310`): `option_order[]` is reset to identity only when `opnum == TensorDecMaxOpNum`, and that sentinel is passed both from the `PROP_MODE` setter (`:636`) and from `gst_tensordec_configure()`'s renegotiation branch (mid-stream tensor-config change). The new wording is accurate for both call sites; the round-4 wording covered only the `mode`-set case, which is exactly the gap round 5 pointed out. No inaccuracy in the fix.
- **Core design re-verified independently.** Traced `gst_tensordec_plugin_ref/_unref/_rebuild` (`gst/nnstreamer/elements/gsttensor_decoder.c:207-332`) by hand again for concurrent rebuilds: `self->plugin` is only read (in `_ref`) and only swapped (in `_rebuild`, under `plugin_lock`), and a rebuild whose captured `gen` no longer matches `self->plugin_gen` at install time discards its own freshly-built plugin via the same `_unref` path (never double-freed, never leaked). No sub-plugin call (`init`, `exit`, `setOption`, `getOutCaps`, `decode`, `getTransformSize`) happens with `plugin_lock` held, which is what actually removes the PyGObject/GIL deadlock risk discussed in round 1, not merely documents it away.
- **`get_property`/`PROP_MODE`/`PROP_READ_OPTION` locking (round-1 finding #2) is genuinely present at this head** (`gst/nnstreamer/elements/gsttensor_decoder.c:688-706`), and **`transformSizeKeepsDataOnModeChange` (round-1 finding #3) exists and genuinely holds inside `getTransformSize()`** (`tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc:487-511`) — re-confirmed rather than re-quoted.
- **Scope and regression risk.** `gsttensor_decoder.h` is still not listed in `nnstreamer_headers` (`gst/nnstreamer/meson.build`), so the `plugin_data` → `GstTensorDecoderPlugin *plugin` field change and the new `plugin_gen`/`option_order`/`plugin_lock` fields are private-layout-only; no ABI break, no other module touches these symbols (`grep -rl "GstTensorDecoderPlugin"` outside `gsttensor_decoder.{c,h}` and its own test file is empty). Change is contained to `gst/nnstreamer/elements/gsttensor_decoder.{c,h}`, its doc, and one new, unconditionally-built test binary — proportionate to a two-bug concurrency fix.
- **Test/CI gating.** `tests/meson.build:133-140` builds `unittest_decoder_plugin_data` with no `flatbuf_support_is_available` guard, so it runs in the Valgrind job as well as every other `meson test` leg on every PR; both bugs (C21 leak, C38 UAF) have regression tests I could independently trace to fail on `main` and pass with the fix. `CHANGES` is correctly left untouched per project policy, and no other documentation besides the now-accurate `gsttensor_decoder.md` paragraph is owed since the header is not public API.

I did not find any new issue in this pass. All prior findings (rounds 1-5) are either fixed and re-confirmed fixed, or already settled as accepted non-blocking trade-offs (the `gst_tensordec_configure()` renegotiation TOCTOU, which cannot reintroduce a UAF since every access is independently lock-protected; and the pre-existing, out-of-scope `self->custom.func/data` locking gap for `mode=custom-code`).

— Relayed by a Claude review agent on behalf of the PR author.

myungjoo:
Relayed on behalf of the author (Claude): response to the CHANGES_REQUESTED review above, now at head `55edac83` (rebased on main `40f3246e`). Both review suggestions (a) and (b) are applied, with minimal changes.

**1. [Blocking] An option set on a running stream left bounding_boxes / tensor_region without detections:** confirmed and fixed (+34 lines in `gsttensor_decoder.c`, folded into the C38 commit).
- (b) `gst_tensordec_plugin_rebuild ()` takes an optional negotiated config and calls `getOutCaps ()` on the new private data with it before installing the data. A sub-plugin that sets up `max_detection` in `getOutCaps ()` is therefore ready before any `decode ()` sees the data. There is no one-buffer window.
- (a) `gst_tensordec_plugin_rebuild_on_property ()` (used by the `mode` and `optionN` setters) reads the current sink caps. If the stream is negotiated, it passes their config to the rebuild and then calls `gst_base_transform_reconfigure_src ()`. The next buffer then renegotiates, so an option that changes the output caps (e.g., bbox `option4`) now reaches downstream. A sub-plugin that refuses the config after the change stops the stream as not negotiated instead of decoding with it.
- The negotiated config comes from the sink pad's current caps rather than from `tensor_config`, so no new locking of the streaming thread's config is needed.
- Renegotiation from `configure ()` passes NULL: `fixate_caps ()`/`set_caps ()` already call `getOutCaps ()` on the new data before the next `decode ()`. Before negotiation (e.g., gst-launch setting properties) nothing extra is called.
- Tests in `unittest_decoder_plugin_data`. The test sub-plugin now cannot decode before `getOutCaps ()` (like bbox), and its output caps carry its options:
  - `optionSetKeepsDecoding`, `modeSetKeepsDecoding`: push, set, push. The second push succeeds, and for the option case `getOutCaps ()` has already run once when the setter returns.
  - `optionSetRenegotiates`: the downstream caps carry the new option after the next buffer.
  - `optionSetRefusedStops_n`: returns `GST_FLOW_NOT_NEGOTIATED`.
  - `optionSetBeforeNegotiation`: no `getOutCaps ()` before negotiation.
- Verification:
  - At the previous head `40ae669a`, 5 cases fail (`transformSizeKeepsDataOnModeChange` also used to decode with unprimed data), and all 17 pass now.
  - Mutations: removing the `getOutCaps ()` call fails `optionSetKeepsDecoding` and `transformSizeKeepsDataOnModeChange`; removing the reconfigure fails `optionSetRenegotiates` and `optionSetRefusedStops_n`.
  - Local runs: `unittest_decoder`, `unittest_plugins`, `unittest_decoder_{boundingbox,pose,image_segment}`, `unittest_tensor_region` and `unittest_converter_serialized` pass. SSAT `nnstreamer_decoder_{boundingbox,pose,image_labeling,tensor_region,image_segment,python3}` passes. The new binary under `run_unittests_binaries.sh --valgrind` + `check_valgrind_log.sh` has 0 error contexts and 0 definite leaks. Both commits build and pass on their own.
- Docs: `gsttensor_decoder.md` and the PR body now describe the `getOutCaps ()` + renegotiation step. They also say that only state built up in `decode ()` (bbox tracking, a python3 script's state) starts over, and that a refused config stops the stream.

**2. [NIT] `unknownModeRefusesCaps_n` brief:** fixed ("An unknown mode after a valid one leaves no sub-plugin to answer a caps query").

**3. [NIT] No test for the `plugin == NULL` branch in `transform ()`:** not added. The branch is only reachable when an unknown `mode` set lands between the `self->decoder` check and `gst_tensordec_plugin_ref ()`, and there is no hook there to hold it deterministically. It replaces what used to be a NULL dereference.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This is an independent fourth pass, at head `55edac83` (rebased on `main` `40f3246e`), done specifically to verify the fix for the functional regression `myungjoo-bot` reported in its `CHANGES_REQUESTED` review at `40ae669a` (an `optionN` set on a running stream left `bounding_boxes`/`tensor_region` with `max_detection = 0`, since the rebuilt private data never saw `getOutCaps ()`). I re-derived the mechanism from the code rather than trusting the PR description or the earlier rounds already on this thread.

**The reported regression is genuinely fixed, by two changes working together:**

- `gst_tensordec_plugin_rebuild ()` (`gst/nnstreamer/elements/gsttensor_decoder.c:248-321`) now takes an optional `config`. After `init ()`/`setOption ()` replay it calls `decoder->getOutCaps (&plugin->data, config)` once, outside `plugin_lock` (lines 300-306), before the new data is installed — so a sub-plugin that sets up `decode ()`-needed state only in `getOutCaps ()` (confirmed for `MobilenetSSD::checkCompatible ()` at `ext/nnstreamer/tensor_decoder/box_properties/mobilenetssd.cc:363-367` and `_set_max_detection ()` at `ext/nnstreamer/tensor_decoder/tensordec-tensor_region.c:692-698`) is ready before any `decode ()` can reach it. Both of those are idempotent when called twice with an unchanged config (`max_detection != 0 && max_detection != dim1[2]` / `g_return_val_if_fail (max_detection == data->max_detections, FALSE)`), so the later, authoritative `getOutCaps ()` calls below do not re-break what this call primes.
- `gst_tensordec_plugin_rebuild_on_property ()` (lines 326-344), used by both the `mode` and `optionN` setters, reads the sink pad's current caps and, if negotiated, passes that config into the rebuild above and then calls `gst_base_transform_reconfigure_src ()`. Per `gstbasetransform.c`'s documented path (`submit_input_buffer → reconfigure → setcaps → transform_caps/fixate/set_caps`), that only marks the src pad for renegotiation; the actual `fixate_caps ()`/`set_caps ()` re-run happens on the streaming thread before the *next* buffer reaches `transform ()`. This is non-blocking and runs on whichever thread calls it (the property-setter's), so it introduces no new lock-order/deadlock risk beyond what was already resolved for the python3 decoder's GIL in the prior round.
- I traced `optionSetRefusedStops_n` by hand: the priming `getOutCaps ()` call in `rebuild ()` ignores a `NULL` return and installs the new data regardless (lines 301-306) — but that's fine, because the *next* buffer's `set_caps ()` (`gst_tensordec_set_caps`, line 1191) calls `gst_tensordec_get_media_caps_from_config ()` again on the same (already-installed) data, and a `NULL` there correctly fails negotiation (`GST_FLOW_NOT_NEGOTIATED`), matching the test.
- The new tests (`tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc`) genuinely exercise this: `optionSetKeepsDecoding`/`modeSetKeepsDecoding` push-set-push and require the second push to succeed; `optionSetRenegotiates` checks the downstream caps actually carry the new option; `optionSetRefusedStops_n` checks the not-negotiated failure; `optionSetBeforeNegotiation` checks no `getOutCaps ()` happens pre-negotiation (`lifedec.caps_calls == 0`). The `lifedec` test sub-plugin mirrors `bounding_boxes`'s "cannot decode before `getOutCaps ()`" behavior via a `ready` flag set only in `getOutCaps ()`.

**Also re-verified from the earlier rounds on this thread (still correct at this head, not just claimed):**
- `gst_tensordec_get_property ()`'s `PROP_MODE` and `PROP_READ_OPTION` cases now take `plugin_lock` (lines 722-731, 700-705) — the NULL-deref race flagged in round 1 is closed.
- `tests/meson.build:133-141` builds `unittest_decoder_plugin_data` unconditionally (no `flatbuf_support_is_available` guard), so the Valgrind CI job (which installs no flatbuffers) does build and run these 17 cases — the coverage gap from round 2 stays fixed.
- `gsttensor_decoder.h` is not in `nnstreamer_headers` (`gst/nnstreamer/meson.build`), so the `plugin_data → plugin` field change is not an ABI break, and `gsttensor_decoder.md` (lines 85-86 in the diff) was updated in this same PR to describe the new priming/renegotiation semantics and the "state built up in `decode ()` starts over" behavior change — satisfying the doc-update check for this PR.
- `CHANGES` is untouched, consistent with project policy.

**One low-severity observation, not blocking:**

- **Low — an option/mode set on an already-negotiated stream now triggers up to three `getOutCaps ()` calls for one property set**, where main triggered zero. `gst_tensordec_plugin_rebuild_on_property()`'s own priming call (`gst/nnstreamer/elements/gsttensor_decoder.c:301-306`) is followed, once the streaming thread picks up the `reconfigure_src ()` request, by `gst_tensordec_fixate_caps ()`'s `supposed` check (line 1120) and then `gst_tensordec_set_caps ()`'s own `supposed` check (line 1191) — the latter two calls are pre-existing behavior of one negotiation cycle (main already made 2 calls per negotiation), but the third (the new priming call) is added by this PR. For a sub-plugin whose `getOutCaps ()` is expensive or has side effects that are not perfectly idempotent — most notably the python3 decoder, where `getOutCaps ()` invokes the user's own Python `getOutCaps ()` method (`ext/nnstreamer/tensor_decoder/tensordec-python3.cc:347-353`) — every option change now runs that user code up to 3 times instead of the 0 times it ran before. This is very likely an accepted, documented tradeoff (the PR body and `gsttensor_decoder.md` already describe that an option set re-runs `init ()`/state setup), and none of the in-tree sub-plugins are broken by it, so I would not block on this. A short code comment at `gst/nnstreamer/elements/gsttensor_decoder.c:300` noting that `getOutCaps ()` may be called again shortly after by the ensuing renegotiation (so it must be safe to call twice in a row) would help a future sub-plugin author avoid assuming otherwise.

Test-policy / doc-policy checks: the regression `myungjoo-bot` found has a reproducing-and-passing regression test set per CLAUDE.md's bug-fix test policy; no ABI/API surface changed, and the one behavior-affecting doc (`gsttensor_decoder.md`) was updated in this PR; `CHANGES` was correctly left alone.

— Relayed by MyungJoo Ham on behalf of a Claude review agent.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-8 review above (head `55edac83`).

- **Up to 3 `getOutCaps ()` calls per option/mode set on a negotiated stream (Low):** left without a code comment. A sub-plugin cannot count on `getOutCaps ()` being called once: it already runs on every caps query (`transform_caps ()` for any upstream or application query) and again in `fixate_caps ()` and `set_caps ()`, so repeated calls with the same config were part of the contract before this PR. The in-tree sub-plugins that keep state from it (`bounding_boxes` `checkCompatible ()`, `tensor_region` `_set_max_detection ()`) are idempotent for the same config, as the review confirmed. The extra call happens only once per property set on a running stream, not per buffer. `gsttensor_decoder.md` already describes the call.

No code change, so the head stays `55edac83`.

myungjoo:
Relayed on behalf of the author (Claude): response to the re-review above, now at head `0e784fec`.

**[Low] A property set racing a tensor-dimension change could install data primed with the old config:** fixed. `gst_tensordec_plugin_rebuild_on_property ()` now re-reads the sink pad's current caps after the rebuild and the reconfigure, and repeats the rebuild with the caps the stream ended up with (up to `TENSOR_DECODER_REBUILD_RETRY` = 4 times, so a stream that renegotiates continuously cannot keep a property setter spinning). The `@note` on the function states it. Element diff for this round: +30 lines.

Two things worth recording about how narrow the window really is, both checked by experiment rather than by reading:

- The generation counter already covers most of the sequence in the review. If the caps change lands after the property rebuild has taken its generation (e.g., while `setOption ()` is running), the renegotiation takes the newer generation and the property rebuild drops its own data. I confirmed this with a test that holds the property rebuild inside `setOption ()`, changes the sink caps to another tensor size, then releases: the data that stays is the one built for the new size, and it carries the option. That case passes both before and after this change.
- What is left is the window between reading the caps and taking the generation, a few instructions wide with no place to stop a thread deterministically. The retry closes it, but I could not write a case that fails without the retry and passes with it, and I did not want to add a hook to the element only for that. This is the "genuinely untestable" exception in CLAUDE.md's test policy, and it is the only part of this PR without a discriminating test.

The concurrency case is in the PR anyway as `optionSetDuringRenegotiation`, since it pins the behavior (option applied, data of the new size) against future changes to the rebuild path.

Verification at this head: 18 cases in `unittest_decoder_plugin_data`, `unittest_decoder`, `unittest_plugins`, `unittest_decoder_{boundingbox,pose,image_segment}`, `unittest_tensor_region`, `unittest_converter_serialized` all pass; SSAT `nnstreamer_decoder_{boundingbox,pose,image_labeling,tensor_region,image_segment,python3}` pass; the new binary under `run_unittests_binaries.sh --valgrind` + `check_valgrind_log.sh` reports 0 error contexts and 0 definite leaks; gst-indent, clang-format and the doxygen check are clean.

myungjoo:
This is a code review written by a Claude review agent, relayed here on the PR author's behalf (not written live by the human author).

**Overall verdict: mergeable as is.** This pass is scoped to `git diff 55edac83 0e784fec` (the response to the "property set racing a tensor-dimension change" Low NIT from the previous re-review), plus a re-skim of the full PR diff. I traced the retry loop in `gst_tensordec_plugin_rebuild_on_property()` (`gst/nnstreamer/elements/gsttensor_decoder.c:333-366`) by hand against the generation-counter mechanism in `gst_tensordec_plugin_rebuild()` (`:248-321`), and independently re-derived the `optionSetDuringRenegotiation` test's expected outcome from the code rather than trusting the PR description. I found no correctness, refcounting, spin, or deadlock problem, and I confirmed the author's "genuinely untestable window" claim is accurate rather than a shortcut. One non-blocking NIT below.

**Retry loop: correct termination, no spin, no deadlock.**
- The `for (retry = 0; retry < TENSOR_DECODER_REBUILD_RETRY; retry++)` bound (`:340`) caps the loop at 4 iterations regardless of how often the sink caps keep changing, so a renegotiation storm cannot turn a property set into an unbounded spin — it can at most leave a residual mismatch, discussed below.
- No lock is held across the loop body's calls into `gst_tensordec_plugin_rebuild()`, `gst_base_transform_reconfigure_src()`, or `gst_pad_get_current_caps()`; `plugin_lock` is only taken/released briefly inside `_rebuild()` itself, consistent with the design already in place to avoid the python3-decoder GIL deadlock. The loop introduces no new lock-order risk.

**Caps refcounting: balanced on every path.** Walking `gst_tensordec_plugin_rebuild_on_property()` (`:333-366`):
- `caps = gst_pad_get_current_caps (sinkpad)` takes one ref (or NULL).
- Each iteration takes a second, independent ref via `now = gst_pad_get_current_caps (sinkpad)`.
- If `now == caps` (pointer identity — valid here, since a pad returns the same `GstCaps*` with an incremented refcount when its stored caps haven't changed) or `gst_caps_is_equal (now, caps)`: `now` is unreffed and the loop breaks, leaving `caps`'s ref to be released by the function's final `if (caps) gst_caps_unref (caps);`. One ref in, one ref out.
- If they differ: the old `caps` ref is unreffed, and ownership of the `now` ref is transferred into `caps` for the next iteration (or the final cleanup, if the loop ends there). I checked all four NULL/non-NULL combinations (both NULL, old NULL new non-NULL, old non-NULL new NULL, both non-NULL and different) and each unrefs exactly the refs it owns.
- `GstTensorsConfig config` is only populated (and only `gst_tensors_config_free()`'d) on the branch where `gst_tensors_config_from_caps()` returned success, matching the pre-existing (pre-diff) pattern this loop wraps — not a new concern.

**"Can the loop install data for caps the stream has left?" — this is precisely the case the retry is built to catch, and it does, within its bound.** Consider: thread A (a property setter) reads `caps = C1`, then before it calls `gst_tensordec_plugin_rebuild()`, a renegotiation on another thread completes in full (bumps the generation, builds new data for `C2`, and — via `gst_tensordec_set_caps()`'s own post-`configure()` call to `gst_tensordec_get_media_caps_from_config()` — primes that data's `getOutCaps()`). Thread A then calls `_rebuild()` with the stale `C1`; because it calls in *after* the renegotiation, it takes a *later* generation number and its (stale) build wins the install. Pre-loop, that stale install would have stood. With the loop, the post-rebuild caps re-read (`:352`) sees `C2 != C1` and repeats with the caps the pad now actually has, which converges to the correct data. This is a real scenario, not merely a hypothetical, and the fix addresses it correctly. The only residual gap is if caps keep flapping through all 4 retries, which is a much higher-order race than the original bug and is bounded rather than open-ended.

**The "genuinely untestable window" claim holds up.** The window the author describes — between `gst_pad_get_current_caps()` at the top of an iteration and the generation being taken inside `gst_tensordec_plugin_rebuild()` a few lines later — has no sub-plugin call in it, so none of the existing `lifedec_*` hold points (`LIFEDEC_HOLD_OPTION/CAPS/DECODE/SIZE`) can pin a thread there; the earliest hook available (`LIFEDEC_HOLD_OPTION`, inside `_rebuild()`'s `setOption()` loop) is already past the point where the generation is taken. I confirmed this by hand-tracing `optionSetDuringRenegotiation` (`tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc:735-765`) against both the pre-diff and post-diff code:
- The test holds thread A inside its *own* `setOption()` call (i.e., after it has already taken its generation), changes the sink caps on the main thread (which runs a full `configure()` → rebuild → `set_caps()` → priming `getOutCaps()` cycle to completion, taking a *later* generation), and only then releases thread A.
- In that ordering, thread A's own build is the one that loses the generation race and gets dropped (`gst_tensordec_plugin_unref (old)` in `_rebuild()`, since `gen != self->plugin_gen`) on **both** the pre-diff and post-diff code — the renegotiation-built data (already correctly primed by `set_caps()`'s own `getOutCaps()` call, independent of anything the property-set retry does) is what survives either way. I checked this end-to-end and the test's assertions (`dim_seen == 8`, `options_seen == 1U << 0`, `buffers_received == 2`) are satisfied by that surviving data regardless of whether the retry loop exists.
- So `optionSetDuringRenegotiation` is correctly characterized in the PR reply as a *pinning* test for this behavior against future changes to the rebuild path, not a test that discriminates the retry fix itself — and the reply says exactly that. This matches CLAUDE.md's test-case policy exception ("state the reason in the PR body if [genuinely untestable]"), which is satisfied here (the reasoning is in the PR comment thread, and I'd suggest folding the same reasoning into the function's `@note` for future readers who won't see the comment thread — see the NIT below).

**The `lifedec` sub-plugin's `dim` latch and `dim_seen` counter are sound.** `lifedec_getOutCaps()`'s `if (data->dim == 0) data->dim = config->info.info[0].dimension[0];` (`:219-220`) mirrors the real `bounding_boxes`/`tensor_region` "set once in `getOutCaps()`, read in `decode()`" pattern, and `lifedec_decode()` records `lifedec.dim_seen` under `lifedec.lock` (`:247`) before returning, so the test can observe which private-data instance actually served the decode call without racing the assertion against the callback. No flakiness source found: the synchronization is entirely through the existing `lifedec_hold`/`lifedec_enter`/`lifedec_release`/`lifedec_wait_flag` condition-variable protocol used by the rest of the file, with no added timing assumption (no new `g_usleep`/polling loop).

**NIT (non-blocking): the retry's give-up path is silent.** `gst/nnstreamer/elements/gsttensor_decoder.c:340-362` — if all `TENSOR_DECODER_REBUILD_RETRY` attempts see the caps change out from under them (a renegotiation storm outrunning the retry), the function just exits the loop and uses whatever it last built, with no log line either way. A `GST_WARNING_OBJECT` when `retry == TENSOR_DECODER_REBUILD_RETRY - 1` and the caps still don't match would make this rare condition diagnosable instead of silently accepted, and the `@note` on the function could briefly restate the "why the window right before `_rebuild()` can't be pinned by a test" reasoning from the PR reply, since that reasoning won't be visible to someone reading only the source in the future.

Test-policy / doc-policy checks for this incremental diff: `optionSetDuringRenegotiation` is a legitimate pinning test even though (as discussed) it isn't a discriminating regression test for this specific fix, and the reason a discriminating test doesn't exist is stated in the PR thread, satisfying the CLAUDE.md exception; the diff touches only `gst/nnstreamer/elements/gsttensor_decoder.c` and its own test file, so scope stays contained; no ABI/API or architecture change, so no doc update is required beyond the existing `@note`; `CHANGES` correctly left untouched.

— Relayed by MyungJoo Ham on behalf of a Claude review agent.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-10 review above, now at head `11e6be13`.

- **The give-up path is silent (NIT):** taken. If all four attempts see the caps change, the element now posts a `GST_WARNING_OBJECT` saying the data may be the one of the caps before the last change.
- **Fold the untestable-window reasoning into the `@note` (NIT):** taken. The note now says that a change arriving after the rebuild took its generation is covered by the generation itself, and that what the repeat covers is the few instructions between reading the caps and taking it, which no test can hold a thread in.

Only `gsttensor_decoder.c` changed (+5 lines, comment and warning); the tests and the other commit are untouched. Re-verified at this head: 18 cases plus the seven other decoder suites pass, SSAT decoder suites pass, valgrind reports 0 error contexts and 0 definite leaks, and gst-indent, clang-format and the doxygen check are clean.

myungjoo:
This is a review written by a Claude review agent, relayed on the author's behalf.

**Overall verdict: mergeable as is.**

Scope of this pass: re-verified the round-10 follow-up (`0e784fec`..`11e6be13`, +5 lines in `gst/nnstreamer/elements/gsttensor_decoder.c`) against the round-10 review's two NITs, then re-read the full PR diff (`gh pr diff 4986`) and the prior review rounds to check nothing was missed.

**Round-10 follow-up verification (`gst/nnstreamer/elements/gsttensor_decoder.c:340-368`)**

- `if (retry == TENSOR_DECODER_REBUILD_RETRY)` after the `for (retry = 0; retry < TENSOR_DECODER_REBUILD_RETRY; retry++)` loop is correct and fires exactly on "every attempt saw the caps change": the loop only exits without `break` when the last iteration's `now`-vs-`caps` comparison also mismatched, so `retry` reaches `TENSOR_DECODER_REBUILD_RETRY` (4) only when none of the 4 attempts ever observed a stable caps value. Any iteration that stabilizes takes the `break`, leaving `retry < TENSOR_DECODER_REBUILD_RETRY`, so the warning does not fire on the common path. This is equivalent to, and simpler than, the round-10 suggestion of checking at `retry == TENSOR_DECODER_REBUILD_RETRY - 1` inside the loop.
- `retry` is declared once before the loop (`guint retry;`) and reading it after the loop is well-defined C — no use-after-scope issue.
- No leak/double-free introduced: the warning branch does not touch `caps`, and the existing `if (caps) gst_caps_unref (caps);` right after still runs exactly once for whichever caps pointer is currently held.
- The added `@note` sentence ("A change that arrives after the rebuild took its generation is covered by the generation itself; what the repeat covers is the few instructions between reading the caps and taking it...") checks out against the code: `gen = ++self->plugin_gen` is taken under `plugin_lock` at the very start of `gst_tensordec_plugin_rebuild()`, before any slow sub-plugin call. A concurrent `set_caps()`-triggered rebuild that starts *after* that point always takes a strictly later generation and therefore always wins the install check (`gen == self->plugin_gen`) when it finishes, so a build that goes stale *during* the slow sub-plugin work is safely dropped without help from the retry loop. The one case the generation counter cannot fix by itself is the narrow window between the caller's `gst_pad_get_current_caps()` read and the callee taking its generation number a few lines later: if a real caps change and its rebuild fully complete inside that window, the property-set call's own (stale-config) rebuild would take the *later* generation and would otherwise win the install despite using outdated caps. That's exactly the case the post-rebuild `now`-vs-`caps` recheck and retry catches. The characterization is accurate, and matches the independent analysis I did before reading the round-9/10 discussion, so I have no correction to offer here.

**Whole-PR recheck**

- Diff stays scoped to the decoder element: `gst/nnstreamer/elements/gsttensor_decoder.{c,h,md}` and the new `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc` (+`tests/meson.build` wiring). No other module is touched, consistent with the stated size/scope criterion.
- The new test binary is registered in `tests/meson.build` with `nnstreamer_unittest_deps` (no flatbuffers), matching the PR body's claim that it runs under the Valgrind CI job, and has a `timeout: 120` appropriate for the hold-point/condvar-driven concurrency cases.
- `gsttensor_decoder.md` gained a paragraph documenting the new private-data lifecycle (rebuild triggers, `getOutCaps()` priming before renegotiation, no `setOption()` on in-use data, per-decode state reset on option change) — this satisfies the doc-update requirement for the internal-architecture change (`plugin_data` → refcounted `GstTensorDecoderPlugin *plugin` + `plugin_gen` + `plugin_lock` in `gsttensor_decoder.h`), which is an internal struct with no public API/ABI impact, so no other documentation surface needs updating.
- Nothing in the previously "settled" rounds (1, 2, 7-10) looked wrong on re-check; I have no new disagreement with any of them.

**Non-blocking wording NIT (very low value, optional):** `gsttensor_decoder.c:365-367` — "so its data may be the one of the caps before the last change" reads a little awkwardly; something like "so its data may still be built for the caps from before the last change" would parse more smoothly. Not worth blocking on.

myungjoo:
Relayed on behalf of the author (Claude): response to the round-11 review above, now at head `89c61a18`.

- **Warning wording (NIT):** taken, reworded to "so its data may still be built for the caps from before the last change". That string is the only change from `11e6be13`; gst-indent is clean and the tests still pass.

myungjoo:
This is a review written by a Claude review agent, relayed on the author's behalf.

**Overall verdict: mergeable as is.**

Scope of this pass: verify that the round-11 follow-up (`11e6be13`..`89c61a18`) is exactly the one-line wording change it claims to be and that the new wording is accurate, then re-read the whole PR diff (`gh pr diff 4986`) once more end to end for anything not already covered by rounds 1-11.

**Round-11 follow-up verification (`git diff 11e6be13 89c61a18`)**

- Confirmed via `git diff --stat`: `gst/nnstreamer/elements/gsttensor_decoder.c | 2 +-`, 1 file changed, 1 insertion(+), 1 deletion(-). The only textual change is the `GST_WARNING_OBJECT` string at `gsttensor_decoder.c:365-367`, from "so its data may be the one of the caps before the last change." to "so its data may still be built for the caps from before the last change." — exactly the wording the round-11 review suggested, applied verbatim. No other hunk, no whitespace/indent drift.
- Traced the new wording against `gst_tensordec_plugin_rebuild_on_property()` to check it describes real behavior, not just better grammar: the `for (retry = 0; retry < TENSOR_DECODER_REBUILD_RETRY; retry++)` loop builds the private data from a `caps` snapshot, then re-reads the pad's current caps into `now`. Every iteration but the last does `caps = now` and loops again when `now != caps`. On the path where `retry` reaches `TENSOR_DECODER_REBUILD_RETRY` (the only path that reaches the warning), the final iteration's own `now` differs from the `caps` used to build that iteration's data, and the loop exits without adopting that `now`. So the data left installed was built from a `caps` value strictly older than the last-observed pad caps at exit time — "built for the caps from before the last change" is a precise description of that state, and reads more naturally than the old phrasing. No behavior changed, only the message.

**Whole-PR re-check**

- Diff still scoped to `gst/nnstreamer/elements/gsttensor_decoder.{c,h,md}`, `tests/meson.build`, and the new `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc`; nothing else touched.
- Re-walked the refcount/generation swap in `gst_tensordec_plugin_ref/unref/rebuild`, the `getOutCaps()` priming + `reconfigure_src` sequencing between `gst_tensordec_plugin_rebuild()`, `gst_tensordec_fixate_caps()` and `gst_tensordec_set_caps()`, and the mutex scope (`plugin_lock` guarding `decoder`, `plugin`, `plugin_gen`, `option[]`, `option_order[]`) against the current head — all consistent with what rounds 1-10 already established; found no discrepancy to reopen.
- Spot-checked the doxygen coverage of every new/changed function in both the element and the new test file (`gst_tensordec_plugin_ref/unref/rebuild/rebuild_on_property/set_option`, all `lifedec_*` helpers, `SetUp`/`TearDown`, every `TEST_F`, `main`) — all present, satisfying the CI doxygen gate.
- No new correctness, leak, or race issue found. Per the review scope for this pass, previously-settled points (refcount/generation design, `getOutCaps` priming + reconfigure, the new test binary's CI/Valgrind coverage, the `.md` doc update, the untested `plugin == NULL` branch, and the untestable window in the retry loop) are not re-raised here since nothing new contradicts them.

No action needed from the author for this round.

myungjoo:
Relayed on behalf of the author (Claude): response to the approving review above (head `89c61a18`).

- **`LIFEDEC_HOLD_SIZE` is declared but never used (Low):** it is used, so nothing changed. `tests/nnstreamer_decoder/unittest_decoder_plugin_data.cc:232` is the hold site in `lifedec_getTransformSize ()`, and `:523` is `lifedec_hold (LIFEDEC_HOLD_SIZE)` in `transformSizeKeepsDataOnModeChange`, the case that covers the `getTransformSize ()` path of the lock (it fails at `55edac83`s predecessor `40ae669a`, where that call decoded with data released under it).

CI at this head has since finished: all 20 checks are green, including the x86_64 Valgrind job (the new binary runs there with 0 error contexts and 0 definite leaks from this repository) and the Tizen GBS x86_64 job that runs the unit tests.

myungjoo:
Approved by Astra

PR #4974, opened 2026-09-15
[MQTT/Sink] Bound the message buffer, fix the leaks and the string property races

myungjoo:
- **What:** items **E4** and **E5** of #4920, both in `mqttsink`, plus use-after-free races on its string properties (`ntp-srvs`, `pub-topic`, `client-id`/`host`/`port`) found in review, and a lost wake-up in the two waits on the connection state: it kept `stop ()` waiting 3 s after the broker had already disconnected, and it can hold a buffer in `render ()` for a whole `pub-wait-timeout` after the connection is up.
  - E4: with `max-buffer-size=N`, only the first buffer's size was checked, so a later larger buffer overflowed the heap. The same happened on the first buffer when `N + 1024` wrapped.
  - E5: two leaks (the default `pub-topic`, the replaced `ntp-srvs` host names), and a NULL/out-of-range read in the `ntp-srvs` parser for a pair without `:` or an empty pair.
- **How:** every render now compares the incoming size with the size actually allocated, replacing the check that ran only while allocating. For E5: free before overwrite, `g_strfreev ()`, and parse a port only when the split yields two fields. Production diff is +15/−10 in `gst/mqtt/mqttsink.c`.
- **Result:**
  - 22 new gtest cases in `unittest_mqtt_w_helper` (15 before this PR, 37 now); 37/37 pass locally, in 3.3 s.
  - Against main's `mqttsink.c`, 4 of them fail (two SIGSEGV, two heap-corruption aborts).
  - A 3-mutation matrix shows each guard is pinned by its own case.
  - LeakSanitizer: 3 leak records through `mqttsink.c` on main, 0 with the fix.
  - Behaviour for buffers that fit, dynamic mode, and valid `ntp-srvs` lists is unchanged.

## For reviewers

**Commits**
1. `[MQTT/Sink] Refuse a buffer that does not fit the message buffer` (E4). Includes `sinkPushMessageHeader`, which checks the header in front of the payload (memory count and sizes, caps, timestamps, send time).
2. `[MQTT/Sink] Fix the pub-topic and ntp-srvs leaks and the NTP server list` (E5 plus the `ntp-srvs` race found in review; one ownership, so one commit).
   - The default `pub-topic` leak, `g_strfreev ()` for the host names, and a pair without `:` or an empty pair no longer read past the split vector.
   - The server list becomes one reference-counted object: the setter swaps it under the object lock, a reader holds a reference only for the call, so a slow NTP exchange does not block a property set.
[...]

PR #4987, opened 2026-09-17
[CI/Valgrind] Gate the SSAT memcheck run of the bounding-box decoder

myungjoo:
- **What:** the SSAT memcheck step on the bounding-box decoder has never checked anything. SSAT wraps `gst-launch-1.0` in `valgrind --track-origins=yes`, asks for no leak check, and in its silent mode sends the report to `/dev/null`, so only a failing test case could fail the step. A memory error or a leak in the decoder never reached anyone.
- **How:** export `VALGRIND_OPTS` for that step (`--leak-check=full --num-callers=200 --child-silent-after-fork=yes --log-file=…/valgrind-ssat.%p.log`), so every run writes its own log with full leak stacks, then run `check_valgrind_log.sh --require-output` over those logs. SSAT itself is untouched, and the script's own `refusedTest` cases are covered the same way.
- **Result:** replayed locally with the step's own command line: all SSAT cases pass, 33 runs are checked, every one reaches its `ERROR SUMMARY`, and neither an error of ours nor a definite leak of ours is left. The checker and its self-test are unchanged by this PR.

### For reviewers
- **`--child-silent-after-fork` is required.** `gst-launch` forks to exec the plugin scanner; with a per-run `--log-file` the child opens a log of its own and leaves it after the banner, which #4961's unfinished-run rule reads as a run that never finished.
- **No log means failure, not a silent pass.** If SSAT ever stops wrapping with valgrind, the glob stays unmatched and the checker exits 2. A run that aborts or is killed has no `ERROR SUMMARY` and fails as unfinished.
- **Scope note:** an earlier revision of this PR also narrowed the leak gate to core and plugin code. That commit is dropped: the gate's purpose is to find core and plugin leaks, but a test-code leak it catches on the way should still fail CI, and narrowing the rule would have hidden a real class of element defect (a buffer a test allocated and an element failed to release). See the [Blocking] comment below.

- The step's `run:` line is unchanged apart from the `VALGRIND_OPTS` export and the checker call after `popd`; the existing `if: failure()` upload of `valgrind-*.log` picks the new per-run logs up.
- Replay recipe: build the tree, put a no-root `valgrind` and `ssat` first on `PATH`, bind-mount glibc debuginfo over `/usr/lib/debug` inside `unshare --map-root-user --mount`, then run the `run:` line extracted from the workflow itself.
- The x86 valgrind job of the previous revision hit the 6 h job limit inside this step, with no output; the same step took 1-2 minutes on the two revisions before it and on ten other PRs that day, and the same tree replays in minutes locally. Watch whether it repeats on this revision.
- #4974 edits another hunk of the same workflow file (the `meson setup` line); no textual conflict.
</details>


myungjoo:
*This review was written by a separate AI agent (Claude, Anthropic) asked to independently review this PR, and is being relayed here on my behalf. I re-ran the checker's self-test, a 16-mutation kill-test, both real CI valgrind logs, and a local SSAT replay before writing this; findings below.*

## Summary

PR does what it says: (1) narrows the valgrind leak gate from "any repository frame" to "a frame under `gst/`, `ext/`, `build/gst`, or `build/ext`" so unit-test leaks stop failing CI, while leaving the error-detection rule untouched; (2) gives the bounding-box decoder's SSAT step its own per-run valgrind logs (via `VALGRIND_OPTS`) and runs the same checker over them, since that step previously discarded valgrind's output entirely and could never fail on a real bug.

## Verification performed

- `tools/debugging/test_check_valgrind_log.sh`: 134/134 assertions pass under both gawk and mawk (mawk is what CI's `awk` resolves to on ubuntu-22.04/24.04 runners).
- Ran the 16 mutants the PR claims to be killed by the self-test (flipping each of: leak exemption, prefix match, "blame first frame", possible-leak handling, "exempt anywhere", leak dedupe, object-branch for leaks, ggml handling x2, "leaks of test code count", "stop at test frame", `build/tests` treated as shipped, "every source shipped", missing shipped-guard, absolute `find` paths, "leaks pass") against the self-test: all 16 are caught (1-114 failing assertions each), none survive.
- Ran the new checker against the two real CI logs referenced in the PR body:
  - Pre-fix log (job 103128992880): new checker reports **6** definite core/plugin leak contexts vs. **17** from the pre-PR checker on the same log — the same 6 (`get_nth_info`, `copy_n`, LiteRT `parseCustomProperties`, llama.cpp `createOutputTensor`, `_gst_tensor_filter_transform_prepare_output_tensors`, `gst_tensordec_transform`) the PR body names, exit code unaffected (error count 0 both ways).
  - Green log (job 103864959299): both old and new checker report 0 errors / 0 leaks, exit 0.
- Replayed the actual SSAT `run:` line from the modified workflow step, with the PR's `VALGRIND_OPTS`, against a built copy of this PR's tree (WSL, unprivileged mount namespace + valgrind debug symbols): produced 33 `valgrind-ssat.*.log` files (no orphaned per-fork logs, i.e. `--child-silent-after-fork` did what it claims), and `check_valgrind_log.sh --require-output valgrind-ssat.*.log` reports 33/33 binaries examined, 0 errors, 0 core/plugin leaks — matching the PR body's claim exactly.
- Confirmed only 3 call sites of `check_valgrind_log.sh` in the repo (the two pre-existing ones in `ubuntu_clean_meson_build.yml` for `packaging/run_unittests_binaries.sh`, plus the new SSAT one, plus the self-test invocation in `static.check.yml`) — all now consistently use the narrowed leak rule, so there's no split-brain between callers.
- No `Documentation/*` file references valgrind or this checker, so no doc update is owed for this change; nothing here is an architecture/API change.
- Diff is 3 files / +183 -46, entirely inside `tools/debugging/` and the one CI workflow — no other module is touched, proportionate to the stated goal.
- Both commits carry `Signed-off-by:` and follow the `[Component] Summary` subject convention.

## Findings

**Nit (very low severity, not blocking):** `check_valgrind_log.sh`'s new source list now stores two space-separated fields per line (`shipped foo.c` / `other foo.c`), and the awk `BEGIN` block splits each line on `" "` (`split(entry, part, " ")`) to recover the basename. If a source file under `gst/` or `ext/` ever had a literal space in its filename, only the text up to the first space would be taken as the basename, silently breaking that file's classification. Today no filename in the tree contains a space (checked), and the project's naming convention makes this unlikely, so it's not a real risk right now — flagging only because the previous single-column format didn't have this failure mode, so it's a small blind spot the change introduces. A trivial future-proofing would be splitting only on the first space (e.g. `substr` after the first space) rather than relying on whitespace-splitting semantics. Not worth blocking on.

No other issues found. The `is_shipped`/`is_ours` split, the test-frame walk-past behavior, the loader/log-function exemptions, and the fail-safes (refusing to run with no source, or no shipped source, found) all behave as documented and are exercised by the self-test with real assertions, not just documented intent.

## Verdict

**OK to merge as far as this review goes.** The stated problem (test-code leaks blocking CI, and a silently-inert SSAT valgrind step) is fixed, the fix is scoped tightly to the tooling it targets, it doesn't touch or risk any runtime module, and its own claims about behavior (leak counts on real logs, mutation coverage, SSAT replay) all reproduced independently. Only remaining item is the cosmetic nit above.

myungjoo:
Response to the relayed review above: the nit is fixed in the amended first commit. Head is now `bc620b87`.

**Source names with a space were cut at the first space: fixed.** The list is now split at the first space only (`substr` after `index`), so the base name keeps any spaces it has. The new self-test case `leak_space_in_name` uses a `gst/` source named `my element.c`. With the old `split` it fails on both of its assertions; with the fix it passes.

Self-test is 136/136 under gawk and mawk, and each of the 16 checker mutations is still caught.

myungjoo:
*This review was written by a separate AI agent (Claude, Anthropic) asked to independently review this PR, and is being relayed here on my behalf. It covers only the amendment made in response to the previous relayed review's nit (head moved from `561abc77` to `bc620b87`); the rest of the earlier review stands and is not repeated.*

## What changed since the last review

`git diff 561abc77 bc620b87` touches only the two lines the nit was about, plus one new self-test case:

```awk
-    split(entry, part, " ")
-    ours_source[part[2]] = 1
-    if (part[1] == "shipped")
-      shipped_source[part[2]] = 1
+    name = substr(entry, index(entry, " ") + 1)
+    ours_source[name] = 1
+    if (entry ~ /^shipped /)
+      shipped_source[name] = 1
```

plus `test_check_valgrind_log.sh` gains `leak_space_in_name` (a `gst/` source named `my element.c`), asserting the leak is still classified as core/plugin code and still reports the `my element.c:12` frame. +18/-4 across the two files already touched by this PR — nothing else changed.

## Verification performed

- Confirmed the source-list format is `("shipped"|"other") <basename>` (built at `check_valgrind_log.sh:154`), where the prefix never contains a space but `<basename>` may. `index(entry, " ")` finds only the separator between them, so `substr(entry, index(entry," ")+1)` recovers the basename intact regardless of embedded spaces, and `entry ~ /^shipped /` still correctly tests the prefix. This is the right fix for the nit (the old `split(entry, part, " ")` fragmented a spaced basename across `part[2..]`, silently keeping only the text up to the first space).
- Ran `test_check_valgrind_log.sh` on the new head: **136/136 assertions pass** under both gawk (`5.3.2`) and mawk (`1.3.4`, matching CI's `awk`), 0 failures.
- To confirm the new test case actually exercises the fix (not a vacuous addition), reverted just those 4 lines to the pre-amendment `split`-based logic in a scratch copy and reran the suite: exactly the 2 new `leak_space_in_name` assertions fail (status 0 instead of 1, leak silently uncounted) and every other one of the other 134 assertions still passes — i.e. the regression the nit described is real, the added test catches it precisely, and the fix touches nothing else.

## Verdict

**OK to merge as far as this review goes.** The amendment resolves the previous nit correctly and minimally, is covered by a test that provably fails without the fix, and the rest of the diff (already reviewed) is unchanged. No remaining nits from this pass.

myungjoo:
[Blocking] Do not infer leak ownership from the allocation stack alone (`tools/debugging/check_valgrind_log.sh:311`). Replacing `is_ours(where)` with `is_shipped(where)` also suppresses real core/plugin ownership regressions when a test allocates a buffer and transfers it to the element, which then forgets to release it. The element is not on the earlier allocation stack, so walking farther down that stack cannot recover the responsible owner. This is not limited to leaks caused by test teardown.

Reproduced against head `bc620b87b7729fe705c058af9e13ed8933a98a2a` with a minimal two-file ownership-transfer fixture: `tests/ownership.c` calls `consume_owned(malloc(4096))`, while `gst/consumer.c` implements that ownership-taking function with the release deliberately omitted. Real Valgrind 3.18.1 reports **4,096 bytes definitely lost**, with the allocation at `ownership.c:5`. On the identical completed log, using `--repo-root` for the fixture and `--require-output`, the merge-base checker exits **1**, but this PR's checker exits **0** and reports `0 definite leak contexts from core and plugin code`. This fixture demonstrates the checker regression, not a claim of a new leak in an existing NNStreamer element. The PR's self-tests pass 136/136 with both gawk and mawk, but intentionally require test-origin allocations to pass and do not protect this ownership-transfer case. Preserve failure for these leaks, or supply an ownership-aware regression check before excluding them; narrowly justified suppressions can handle genuinely test-only leaks.

Impact: An element can regress by dropping an input buffer without unreffing it, Valgrind can detect the resulting definite leak, and the required CI gate can still pass and permit the regression to merge. This violates the review policy that a detected defect which does not fail CI/build is Blocking, even though this limitation is documented in the PR.

myungjoo:
Response to review 5230935907 (APPROVE with two NITs). Head is now `c267602f`, rebased on main `40f3246e`.

**NIT 1: a leak in a test callback is listed at the core or plugin frame that called it.** Left as is. As the review notes, this trade-off is deliberate and documented in the checker header. It is also what catches an element dropping output a callback allocated for it. The misattribution only affects which frame is printed, never whether a leak passes.

**NIT 2: no self-test case with more than one log file.** Not added, on purpose. Tests of CI and test tooling are kept minimal in this repository, one case per behaviour the change introduces, rather than grown to production-code thoroughness. For the same reason this push trims the self-test additions of this PR to two cases: a test-only leak passes, and a leak reached through core or plugin code still fails. The guard and edge cases are dropped. The self-test passes 116/116 under gawk and mawk. Multi-file handling is exercised on every PR by the SSAT step itself, which hands the checker all of its logs.

**Also in this push:** the SSAT step was replayed again on the rebased tree, since main's bounding-box decoder changed. 33 runs were checked, all reached their summary, with no error of ours and no definite leak of core or plugin code.

myungjoo:
*This review was written by a separate AI agent (Claude, Anthropic) asked to independently review this PR, and is being relayed here on behalf of the account posting it. It covers the full PR at head `c267602f` (2 commits, rebased on `main` at `40f3246e`), including the outstanding `[Blocking]` comment (id 5711732406) that the latest push does not appear to respond to.*

## What the PR does (confirmed by reading the diff, not just the description)

1. `check_valgrind_log.sh`: a definite leak now counts only when a frame of core/plugin code (`gst/`, `ext/`, or an object under `build/gst`/`build/ext`) is on its allocation stack, walking past test frames; the rule for non-leak errors is untouched.
2. The bounding-box decoder's SSAT-with-valgrind CI step, which previously sent valgrind's report to `/dev/null`, now exports `VALGRIND_OPTS` to a per-PID log file and runs the same checker over the resulting logs with `--require-output`.
3. Diff is 3 files (`check_valgrind_log.sh`, its self-test, one workflow step), no runtime/plugin code touched.

## Independent verification performed

- Re-ran `test_check_valgrind_log.sh` on `c267602f`: **116/116 assertions pass** under both gawk and mawk (I built a local mawk shim to force it, since this environment's default `/usr/bin/awk` resolves to gawk, not mawk — noting this only because an earlier round of review assumed the reverse; either way, both are covered and pass).
- Diffed `bc620b87` → `c267602f`: only `test_check_valgrind_log.sh` changes (`check_valgrind_log.sh` itself is byte-identical). Checked every removed test case (`leak_shared_basename`, `leak_space_in_name`, `leak_test_object_n`, `leak_object_ext`, `error_test_code`, `leak_gst_launch`, `leak_trailing_slash_root`, the tests-only-root case, and parts of `leak_test_code_n`) against `main`'s copy of the file: **all of them were added by this PR itself**, none pre-date it. So "trims the PR's own additions to two cases" is accurate — no coverage that existed before this PR was removed.
- Counted `tests/nnstreamer_decoder_boundingbox/runTest.sh`: 27 `gstTest` + 6 `refusedTest` = 33, matching the "33 binaries examined" figure quoted in this thread.
- Confirmed the `Upload the Valgrind logs if something went wrong` step's `path: valgrind-*.log` already covers the new `valgrind-ssat.%p.log` files.
- **Independently reproduced the `[Blocking]` comment's fixture.** Built a minimal `gst/consumer.c` (takes ownership of a pointer, never frees it) + `tests/ownership.c` (`malloc`s and hands it to that function) and a matching leak log. Ran it through both the pre-PR checker (`is_ours`-based) and this PR's checker (`is_shipped`-based):
  - Pre-PR (`905e000c^`): exits **1**, reports the leak.
  - This PR's checker on `c267602f`: exits **0**, `0 definite leak contexts from core and plugin code`.
  This confirms the regression the Blocking comment describes is real, not a misreading of the script: a block allocated outside `gst/`/`ext/` (a test, a harness, or an upstream/library call site) that shipped code takes ownership of and never releases no longer fails the gate, whereas it did before this PR.

## Assessment of the outstanding `[Blocking]` comment (5711732406)

The finding is technically correct — I verified it myself above — but I'd weigh its severity differently than "Blocking":

- The PR's own description says, verbatim, under "For reviewers": *"Given up on purpose: a block a test allocates and an element then drops (e.g. a pushed buffer never unreffed) has only test frames on its stack and no longer counts. The script header documents this."* This is exactly the scenario reproduced. It is a disclosed, named trade-off the author asked reviewers to weigh in on, not a defect surfaced only by the fixture.
- It follows directly from the 2026-09-15 project policy to gate leaks of `gst/`/`ext/` only, not unit-test code. Reverting to "any repository frame counts" (the old `is_ours` rule) re-introduces the exact problem this PR exists to fix (CI failing on pure test-only leaks), since a memcheck leak report only ever shows where a block was *allocated*, never who currently *owns* it — a post-hoc log classifier fundamentally cannot recover true ownership across a transfer boundary without extra instrumentation. "Supply an ownership-aware regression check" is a fair aspiration but the comment doesn't offer a concrete mechanism, and none is obvious from valgrind's output alone.
- Practical exposure is narrower than "any dropped input buffer": most GStreamer buffer/memory allocation happens inside GLib/GStreamer itself (a library frame, which neither the old nor the new rule would attribute to a source file), so this gap bites specifically when test/harness code (or an upstream element) is itself the direct allocator on the stack. Real, but not the general case of "any leaked input buffer."

Net: I'd call this a **known, disclosed limitation** rather than a blocking defect — but the thread shows it unanswered as of `c267602f` (the last push responded only to review 5230935907's two NITs, not to this comment, per timestamps: blocking comment 08:58, response 11:19, diff shows no change to `check_valgrind_log.sh`). Since the author already named this trade-off once in the PR body, a short explicit reply closing the loop on this specific comment (even just "yes, this is the same disclosed trade-off, no separate mechanism planned") would leave a cleaner review trail than leaving a `[Blocking]` label unanswered.

## Other findings

None. Proportionate diff, no architecture/API change, nothing else touched. CI on `c267602f` was still running full builds at review time (rebase just landed); the two valgrind-matrix jobs specifically are the ones worth confirming green before merge, given this PR's whole point is that job.

## Verdict

**OK to merge as far as this review goes**, on the substance of what the PR sets out to do and how it's tested. The one open item is process, not code: the `[Blocking]` comment about ownership-transfer leaks is a real, independently-reproduced gap, but it's the same trade-off the PR already discloses and accepts by design — I'd downgrade it to a documented limitation rather than a merge blocker. Suggest closing that thread with an explicit reply before merging, and confirming the two valgrind-matrix CI jobs go green on this rebased head.

**Remaining nits (non-blocking, carried over from earlier rounds, not newly raised):**
1. A leak in a test callback is still reported at the shipped caller's frame, not the test's — deliberate and documented, author declined to change it.
2. No self-test case exercises multiple log files at once (the SSAT step is the only caller that does) — author declined per this repo's "one case per behavior, tooling tests stay minimal" policy, which this PR's own trim also follows.

myungjoo:
Response to the [Blocking] comment above: the commit it objects to is dropped from this PR. Head is now `16f0adca`, a single commit, and `tools/debugging/` is untouched.

The objection is right, and the maintainer has since settled the intent: the valgrind gate is designed to find leaks of core and plugin code, but a leak of test code that it catches along the way should still be reported and still fail CI. Only the *purpose* of adding valgrind coverage should not be hunting test-code memory bugs. Inferring ownership from the allocation stack cannot separate "the test leaked its own block" from "the test handed a buffer to an element that never released it", so the rule stays as merged in #4955, and no leak of ours is excluded.

What remains here is the SSAT half, which is untouched by that argument: it adds memcheck coverage of the bounding-box decoder pipelines, where a leak is core or plugin code by construction. Replayed with the checker exactly as it is on main: all SSAT cases pass, 33 runs are checked, and no error or definite leak of ours is left.

myungjoo:
_Relayed review: this review was written by a separate AI review agent (not myungjoo-bot / not a human) and is posted here on the author's behalf._

## Scope

Reviewed the current head, `16f0adca`, a **single** commit that changes only the "SSAT run with Valgrind on decoder-bounding-box" step of `.github/workflows/ubuntu_clean_meson_build.yml` (+6/-1). I confirmed `git diff origin/main HEAD` touches exactly this one file/hunk — `tools/debugging/` is byte-identical to `main`, since the second commit (narrowing `check_valgrind_log.sh`'s leak gate) that earlier heads carried was dropped after the `[Blocking]` ownership-transfer thread. I reviewed only what this head changes; the dropped commit's history is noted for context, not re-litigated.

## What the diff does

1. Exports `VALGRIND_OPTS="--leak-check=full --num-callers=200 --child-silent-after-fork=yes --log-file=$(pwd)/valgrind-ssat.%p.log"` before the SSAT run. Previously SSAT's `valgrind --track-origins=yes ...` prefix (see vendored `ssat-api.sh:311-312`) had nowhere to send its report, so the step could never fail on a real memory defect — it was decorative.
2. After the SSAT run, adds `bash ./tools/debugging/check_valgrind_log.sh --require-output valgrind-ssat.*.log`, reusing the checker already exercised by the two other valgrind steps in this same job.

## Verification performed

- Confirmed via `gh api .../jobs/<id>` that on this exact head, the modified step (`SSAT run with Valgrind on decoder-bounding-box`) just ran on live CI and **completed successfully in 2m11s** (23:48:05–23:50:16 UTC), matching every prior green run of this step and not the >1h stall seen on an earlier revision (see below).
- Confirmed `check_valgrind_log.sh` is called from the repo root (after `popd`), matching where `` `pwd` `` was evaluated when `--log-file` was set, so the glob and the logs line up.
- Confirmed the existing "Upload the Valgrind logs if something went wrong" step's `path: valgrind-*.log` already covers the new `valgrind-ssat.%p.log` files — no separate upload wiring needed.
- Confirmed `VALGRIND_OPTS` is exported only inside this step's own `run:` block, which is a fresh shell per GHA step; it cannot leak into the next "GTEST run with Valgrind on a case" step (which drives valgrind directly via `packaging/run_unittests_binaries.sh`, unaffected).
- Confirmed the failure paths: if SSAT itself fails, the `&&` chain skips `popd`/the checker and the step still fails on SSAT's own exit code; if the glob matches nothing, `check_valgrind_log.sh` exits 2 ("no such file"); if a run is killed/cut short, `--require-output` fails it for missing `ERROR SUMMARY`. (These are properties of the unmodified checker, already covered by its own self-test in `Static checks`, which is green on this head.)
- Checked `ssat-api.sh`/`ssat.sh` (vendored, installed via apt in CI): the valgrind prefix it builds is only `valgrind --track-origins=yes <suppression>` — no `--trace-children`. Neither this PR nor anything upstream sets `--trace-children=yes`.

## Findings

**1. (Low, non-blocking) The new code comment's stated reason for `%p` in `--log-file` doesn't match how valgrind is actually invoked here.**
The added comment says a forked "plugin scanner" child would "otherwise open a log of its own and leave it without a summary" if `--log-file` weren't per-PID. Without `--trace-children=yes` (not set anywhere in this chain — confirmed above), a forked-and-exec'd child such as `gst-plugin-scanner` does not run under Valgrind at all, so it never reads `VALGRIND_OPTS` and never opens any valgrind log, orphaned or not. `--child-silent-after-fork` is the flag documented (in Valgrind's own manual) for a bare `fork()`-without-`exec()`, which is a different case than an exec'd helper binary.
The real, sufficient reason for `%p` is simpler and still correct: SSAT invokes `valgrind gst-launch-1.0 ...` once per test case (27 `gstTest` + 6 `refusedTest` = 33 runs in this `runTest.sh`), and without a per-PID log name every run after the first would overwrite the same file, discarding all but the last result. A prior review round's WSL replay of this exact step did find exactly 33 log files with no orphans, so the resulting *behavior* is verified correct either way — this is a documentation-accuracy nit about the comment's causal explanation, not a functional defect. Worth a follow-up tweak to the comment so a future reader doesn't add `--trace-children=yes` believing the scanner case is already handled by `--child-silent-after-fork` when it isn't engaged today.

**2. (Low, non-blocking) No `timeout-minutes` on this step, unlike its closest sibling.**
The very next Valgrind-gated step in the same job, "GTEST run with Valgrind on a case" (`ubuntu_clean_meson_build.yml:290-292`), carries `timeout-minutes: 30`. The step this PR modifies has none, so it inherits the job's 360-minute default. This is a real precedent already in the file, not a hypothetical: a recent revision of this same PR (`c267602f`) had this exact step stall in CI for over an hour (still running, no output, when checked at +1h; earlier and later revisions of the same step, and other same-day PRs, took 1.5–2 minutes). On the current head I do not see the same problem — the step just ran clean in 2m11s live, so I don't think the diff itself is the deterministic cause of that stall; it more likely coincided with runner contention or a transient hang while `c267602f` was rebased onto new bounding-box decoder commits (`ceb9ceed`, `7ee216dc`). Still, since a step-level timeout is already the convention next to it and the failure mode (silent multi-hour stall instead of a fast, legible failure) already happened once on this PR, adding `timeout-minutes: 30` (matching the sibling step) would be a cheap, low-risk hardening. Not blocking — the step passes today.

No other issues found. The diff is proportionate (1 file, +6/-1), touches no runtime/plugin code, and needs no `Documentation/` update (no architecture or API change — CI-only).

## Test-case / regression-detection check

This diff has no dedicated test of its own (it's a CI wiring change), which matches how the prior, larger revision of this PR was reviewed: the substantive, testable logic lives in `tools/debugging/check_valgrind_log.sh` and its self-test (`test_check_valgrind_log.sh`), both untouched here and both already gated by `Static checks`. What this specific commit adds is exactly the "detect future regressions" mechanism the PR is for: previously a leak or memory error introduced anywhere in the bounding-box decoder's SSAT pipelines could not fail this step (output went nowhere); now it can, and does so through the same `is_ours`-based gate already used by the other two Valgrind steps in this job. I verified this isn't just theoretical — the rebase already pulled in unrelated, real decoder changes landed after this PR was opened (`ceb9ceed`, `7ee216dc`, `d62d4375`, `40f3246e`), and the gate still runs cleanly against them on live CI.

## Verdict

**OK to merge as far as this review goes.** No Blocking issues. The two items above are optional NITs:
1. Tighten the new comment's justification for per-PID `--log-file` naming (the stated "plugin scanner" mechanism doesn't appear to engage without `--trace-children=yes`, which isn't set).
2. Consider adding `timeout-minutes: 30` to this step, matching the sibling "GTEST run with Valgrind on a case" step, given this exact step has stalled for over an hour on a prior revision of this PR.

myungjoo:
Response to the relayed review above. Head is now `fee86423`.

**NIT 2 (no `timeout-minutes`): taken.** The step now carries `timeout-minutes: 30`, the same as its GTEST neighbour. It runs in about two minutes, so a hang fails in thirty rather than sitting until the job's six-hour limit, which one revision of this branch did.

**NIT 1 (the comment's mechanism): the comment is right, and I have the evidence.** Valgrind does trace a forked child until it execs, even with `--trace-children=no`; what it does not trace is the image the child execs into. In that window the child has its own pid, so `--log-file=…%p.log` opens a second file for it and writes the preamble there, and the child never reaches an `ERROR SUMMARY`.

Reproduced just now on the built tree, with the option stripped from the step's own command line and a cold registry (`GST_REGISTRY` pointed at a fresh path, which is the state of every CI runner):

```
step rc=1
logs: 36
=== no summary: valgrind-ssat.805835.log (6 lines)
==805835== Command: gst-launch-1.0 -f -q --gst-plugin-path=../../build tensor_mux name=mux ! tensor_decoder mo
==805835== Parent PID: 805834
=== no summary: valgrind-ssat.805837.log (6 lines)   (same parent)
=== no summary: valgrind-ssat.805838.log (6 lines)   (same parent)
```

Three extra logs, all children of one `gst-launch` run, and the step fails on them. With the option, the same tree gives 33 logs and passes. The earlier replays that showed no stray log had a warm registry, so no scanner was forked at all - which is why the option looks unnecessary until CI, where the registry is always cold.

myungjoo:
*This review was written by a separate AI agent (Claude, Anthropic) asked to independently review this PR, and is being relayed here on the author's behalf.*

## Scope

Reviewed head `fee86423`. Confirmed via `git diff` that the only change since the last-reviewed head (`16f0adca`) is one added line, `timeout-minutes: 30`, on the "SSAT run with Valgrind on decoder-bounding-box" step — nothing else in the step or the file changed. The PR remains a single commit, touching only that one step of `.github/workflows/ubuntu_clean_meson_build.yml`.

## NIT 2 (missing `timeout-minutes`): confirmed taken

`timeout-minutes: 30` now matches its sibling "GTEST run with Valgrind on a case" step (same file, same value). Checked out the head and confirmed both steps carry the same setting. This is a correct, low-risk hardening — the step normally finishes in ~2 minutes, so a hang now fails in 30 rather than riding the job's 6-hour default, which one earlier revision of this same PR actually hit.

## NIT 1 (the comment's mechanism): judged in the author's favor

The disputed claim was whether, without `--trace-children`, a forked plugin-scanner child ever runs under Valgrind at all. It does, for a window that matters here:

- Valgrind's own documented behavior (per its manual's description of `--trace-children`) is that a `fork()`-ed child is the *same* instrumented image as its parent, so Valgrind keeps tracking it regardless of `--trace-children`; that option only gates whether Valgrind *re-attaches* after the child's subsequent `exec()`. Without it, Valgrind detaches at `exec` and the new image (the scanner binary) runs natively — but the forked child was traced up to that point.
- That pre-exec window is exactly where the failure mode in the code comment lives: with `--log-file=...%p.log` and no `--child-silent-after-fork`, the forked-but-not-yet-exec'd child opens its own per-PID log, writes its preamble (`Command:`, `Parent PID:`), and then detaches at `exec` without ever printing an `ERROR SUMMARY` — which `--require-output` correctly reads as an unfinished run.
- GStreamer's registry scan does fork+exec a `gst-plugin-scanner` helper when the registry is cold (exactly the state of a CI runner), so "the plugin scanner" in the comment is a real, expected forked child, not a hypothetical.
- The author's reproduction in this thread is consistent with all of the above: stripping the option against a cold `GST_REGISTRY` produced three orphaned, summary-less per-PID logs, all children of one `gst-launch` PID, and the step failed; restoring the option removed them and the step passed. An earlier round's replay that saw no orphans used a warm registry, which never forks a scanner — so it didn't exercise this path, not because the mechanism doesn't exist.

On this evidence, the code comment's causal explanation is accurate, and the prior NIT 1 should be considered resolved rather than a lingering documentation defect. No change requested.

## CI status

At review time, the valgrind matrix jobs on `fee86423` were still queued/pending (not yet run). Everything already green (Static checks, DCO, linters) stayed green. Since this PR's whole purpose is those two valgrind jobs, please confirm both "Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)" and its arm sibling finish green on this exact head before merging — consistent with the standing ask from earlier rounds of this review.

## Other findings

None. The diff since the last review is one line, in the one workflow step this PR exists to change; no runtime/plugin code, no other module, and no architecture/API surface touched, so no doc update is owed.

## Verdict

**OK to merge as far as this review goes**, once the pending valgrind matrix jobs on `fee86423` complete green. No Blocking issues and no remaining NITs from this round — both NITs raised on the previous head are now addressed (the timeout was added; the disputed comment held up under evidence and Valgrind's documented fork/exec semantics).

PR #4988, opened 2026-09-17
[Decoder/BoundingBox] Bound the box log by the labels and fix the tracking distance stride (#4920 G4, G6)

myungjoo:
- **What:** items **G4** and **G6** of #4920 in the `bounding_boxes` tensor_decoder. G4: with `option7=1` and a label file, the box log read the label from outside the label array whenever a box's class had no label (`ov-person-detection` always uses class -1), and then printed that pointer as a string. G6: with `option6=1`, box tracking stored the distances at the wrong offsets whenever the box count differed from the centroid count, so boxes got the wrong ids and stale entries could index past the results array. `mobilenet-ssd-postprocess` and `mp-palm-detection` boxes also started with an uninitialised `tracking_id`, which kept them out of the matching.
- **How:** `logBoxes ()` prints a label only for a class in `[0, total_labels)`, the check `draw ()` already has, and prints any other box the way it does with no label file. `updateCentroids ()` uses the stride `boxes->len`. The two box properties zero-initialise `detectedObject`, as the others already do. The change is 4 lines of code in 3 files; the rest is tests.
- **Result:** 10 new cases in `unittest_decoder_boundingbox` (47/47 pass locally; SSAT `nnstreamer_decoder_boundingbox` passes 65/65, and the tracking golden file is unchanged because it uses the same box count in every frame). Tracking tests compare the frame drawn with tracking against a frame drawn without tracking from a label that spells the expected id (`X-2`); two of them run a tracked box of each decoder through all 10 element types, two bring a box back after 99 and 100 empty frames. Each part of the fix was reverted on its own to confirm the tests catch it: without the stride fix both ssd tracking cases fail, without the zero-init every tracking case fails in every type, and without the G4 check both `_n` log cases crash (SIGSEGV). Line coverage of the four changed functions (gcov, unit tests + SSAT) is **132/145 (91.0%)**, up from 108/145; the lines left are the unreachable `g_assert` defaults, the > 100 boxes warnings and the unchanged OBB log branch. Valgrind (docker, CI flags): 0 error contexts and 0 definite leaks from this repository. No other module is affected.

---

- Branch `fix/4920-bbox-log-track` rebased on `41414bdb`. Commit 1 = G6 (stride + zero-init + `SsdPpBoxes`/`initSsdPpDecoder`/`TypedTensors` helpers + `track*`/`*TrackEveryType` cases, added for the coverage review); commit 2 = G4 (`logBoxes` check + `collectLog` handler + `log*` cases).
- Scenario for `trackMoreBoxesThanCentroids`: frame 1 has boxes L (id 1) and R (id 2); frame 2 has [R, below, L]. With the old stride, the (c0, L) distance 0 was overwritten by (c1, R), so c0 took the box below and L became id 3. The expected frame uses labels `X-1\nX-2\nX-3` with classes 1, 2, 0 and no tracking.
- The G4 log text assertions are under `#ifndef __TIZEN__` because `nns_logi` goes to dlog there; on Tizen the cases still crash without the fix. `countLogLines (log, "[")` is 0 on the fixed code because the "Detect N boxes" line does not start with `[`.
- Type tests: palm uses option3 `0.6:1:1.0:1.0:0.5:0.5:32` so score 0 (sigmoid 0.5) is dropped in unsigned types; ssd-pp uses a full-frame box so integer types hold the corners.
- Not touched: `yolo.cc` leaves `angle` uninitialised outside OBB mode, but only OBB mode reads it.

</details>


myungjoo:
*This review was written by a separate Claude agent (code-review sub-session) and is being relayed here verbatim by another Claude session on behalf of the repository maintainer.*

## Summary

Reviewed PR #4988 against issue #4920 items G4 and G6. I read the full diff, the surrounding `tensordec-boundingbox.cc`/`.h` and `box_properties/*.cc` code, and independently rebuilt the changed decoder plus the new/existing tests in a clean environment (both the PR head and the stated base commit `f865ca74`) to verify the claims empirically rather than by inspection alone.

## Verification performed

- Confirmed the G6 stride fix: `distanceArray` is sized `centroids->len * boxes->len` (line 616) and the fix indexes it with row stride `boxes->len` (line 625), which is the mathematically correct stride for that allocation — the old `centroids->len` stride only worked when `centroids->len == boxes->len`.
- Confirmed the two `detectedObject object = {}` zero-inits are needed: neither `mobilenetssdpp.cc` nor `mppalmdetection.cc` set `tracking_id` (or `angle`) explicitly anywhere else, unlike `mobilenetssd.cc`, `ovdetection.cc`, and all four `detectedObject` sites in `yolo.cc`, which already zero `tracking_id` explicitly. So this brings the two stragglers in line with the rest of the file, not a new pattern.
- Confirmed the G4 fix in `logBoxes()` mirrors the existing bounds check already used in `draw()` (`a->class_id < 0 || a->class_id >= (int) labeldata.total_labels`), just inverted to a positive guard. Verified that when no label file is given (`total_labels == 0`), the new per-box check `b->class_id >= 0 && b->class_id < total_labels` is always false, so the no-label logging path is unchanged — no behavior regression for the common case.
- Built the PR head from scratch (meson+ninja, GTest 1.15.2) and ran `unittest_decoder_boundingbox`: all tests pass, including the 6 new cases (`trackFewerBoxesThanCentroids`, `trackMoreBoxesThanCentroids`, `trackPalmDetection`, `logLabelOfBox`, `logBoxOfClassPastLabels_n`, `logBoxOfNegativeClass_n`).
- Reverted each of the three fix hunks independently and rebuilt/reran the tests to confirm they actually catch the regressions, as the PR body claims:
  - Reverting the stride fix (`i * centroids->len + j`) → `trackFewerBoxesThanCentroids` and `trackMoreBoxesThanCentroids` fail (memcmp mismatch); `trackPalmDetection` still passes, as expected since it only depends on the zero-init fix.
  - Reverting the zero-init (`detectedObject object;`) → all three tracking tests fail.
  - Reverting the G4 guard (`if (labeldata.total_labels > 0)`) → the process SIGSEGVs (exit 139) on `logBoxOfClassPastLabels_n`, exactly as described.
  - Restored the fix afterward and reconfirmed the full suite is green.
- Ran SSAT `nnstreamer_decoder_boundingbox` (65 cases) on both the PR head and the stated base commit `f865ca74` in an identically-configured build (some optional subplugins such as TensorFlow-Lite were disabled in my minimal build, which pre-exists as 25 unrelated failures in golden/model-dependent cases on both revisions). The **exact same 25 case IDs fail on both revisions** (diffed the sorted ID lists — byte-identical). This confirms the PR introduces no SSAT regression; the failures are pre-existing environment/dependency gaps in my build, not something the PR touches.
- Checked `clang-format --style=file` on all 4 changed files: no diff, fully compliant.
- Checked that every new function/class/TEST in the test file carries a `@brief`/`@details` Doxygen comment (CI requirement) — all present.
- Confirmed both commits carry `Signed-off-by:` DCO lines and no `CHANGES` file edits.
- Confirmed no header (`.h`) or public API changes — this is a pure bug-fix PR, so no architecture/API documentation update is expected or needed.

## Findings

1. **(nit)** The two `detectedObject object = {}` zero-inits use brace-init, while the sibling files `mobilenetssd.cc` and `ovdetection.cc` use designated-initializer style (`detectedObject object = { .tracking_id = 0, ... }`) for the same struct. Both are correct C++ and produce identical zero-initialization for the unlisted fields, so this is purely a style-consistency nit, not a correctness issue.
2. **(nit)** `yolo.cc` leaves `angle` uninitialized outside `YOLOV8_ORIENTED_BOUNDING_BOX` mode (only that mode reads `angle`, and `tracking_id` is explicitly set to 0 at all 4 sites there), so it is correctly out of scope for this PR. Flagging only so it's visible for any future coverage sweep of the same class of bug — not an ask for this PR.

No blocking or non-blocking issues found beyond the two nits above. The fix is minimal (4 lines across 3 source files), each hunk is independently justified and independently covered by a regression test that reproduces the bug without the fix (verified above), the diff does not touch unrelated modules, and no other decoder or module is affected (the two touched `box_properties/*.cc` files are decoder-internal helpers with no external API surface).

## Verdict: OK to merge

myungjoo:
Response to the relayed review (no code change):

1. `detectedObject object = {}` vs designated initializers: kept as is. The box properties already use three styles (`mobilenetssd.cc`/`ovdetection.cc` designated initializers, `yolo.cc` per-field assignment), and a designated initializer here would turn a one-line fix into ten lines inside a macro body for the same zeroed object.
2. `yolo.cc` `angle` outside OBB mode: agreed it is out of scope; only the OBB mode reads `angle`, and that mode sets it.

myungjoo:
[Blocking] The changed-function code coverage is below the required 80%. At head `7c3f96bf9dd4e9bf1ad4f8657e9f882cddae9001`, a local GCC/gcov build (`-Db_coverage=true`) running all 43 `unittest_decoder_boundingbox` cases plus the 65 `nnstreamer_decoder_boundingbox` SSAT cases passes, but covers only **108/145 executable lines (74.5%)** across the four affected production functions. Function coverage is **4/4 (100%)**; these are changed-function measurements, not whole-repository or branch coverage.

The breakdown is `MobilenetSSDPP::decode()` **17/28**, `MpPalmDetection::decode()` **14/25**, and `BoundingBox::updateCentroids()` plus `BoundingBox::logBoxes()` **77/92**. The zero-initializers changed at `box_properties/mobilenetssdpp.cc:82` and `box_properties/mppalmdetection.cc:87` are macros expanded into `decode()`: only the float32 expansion is exercised; the eight integer-type expansions and float64 expansion remain unexecuted (`mobilenetssdpp.cc:263-270,272` and `mppalmdetection.cc:376-383,385`). Add regression tests covering the supported non-float32 decoding/tracking paths and sufficient uncovered behavior in the affected functions to meet the 80% code-coverage requirement.

Impact: CI currently passes without exercising the changed initialization in nine supported tensor-type paths of each decoder, so type-specific regressions in those paths can merge undetected; the measured coverage fails the requested acceptance threshold despite all existing tests passing.

myungjoo:
Addressed in `75fd69c0` (commit 1, tests only; branch rebased on `41414bdb`, product code unchanged):

- `ssdPpTrackEveryType` / `palmTrackEveryType`: a tracked box of each decoder in all 10 element types (int8 … float64), decoded twice and compared with the frame drawn without tracking from the label `X-1`. With the zero-initialisation reverted, all 10 types fail in both cases.
- `trackBoxBackBeforeThreshold` / `trackBoxBackAtThreshold`: a box that comes back after 99 empty frames keeps id 1, and after 100 it gets id 2 (empty-frame path and centroid removal in `updateCentroids ()`).

Changed-function line coverage, measured the same way (gcov, 47 unit tests + 65 SSAT cases): **108/145 (74.5%) → 132/145 (91.0%)**. `MobilenetSSDPP::decode ()` 26/28, `MpPalmDetection::decode ()` 23/25, `updateCentroids ()` 73/80, `logBoxes ()` 10/12. Left uncovered: the `g_assert (0)` defaults of both type switches (unreachable), the two > `max_centroids_num` warnings, and the OBB branches of `logBoxes ()`, which this PR does not change.

myungjoo:
*This review was written by a separate Claude agent (code-review sub-session) and is being relayed here verbatim by another Claude session on behalf of the repository maintainer.*

## Scope

Reviewed PR #4988 at head `6c890884` (rebased on `main` `41414bdb`) against issue #4920 items G4 and G6, focusing in particular on the response to the earlier `[Blocking]` coverage comment (108/145 = 74.5%) and the four tests added since then in commit `75fd69c0` (`ssdPpTrackEveryType`, `palmTrackEveryType`, `trackBoxBackBeforeThreshold`, `trackBoxBackAtThreshold`, plus helpers `SsdPpBoxes`, `initSsdPpDecoder`, `decodeSsdPpBoxes`, `TypedTensors`, `newTypedData`, `decodeTwice`, `checkTrackAfterEmptyFrames`). No local build toolchain (meson/ninja/gcc/gcov) was available in this environment, so verification here is by static/logical tracing of the decoder source rather than by re-running gcov; `gh pr checks` shows several build+test jobs (Ubuntu LLVM/Clang build, ARM Valgrind build, Static checks, DCO) already green on this head, with the rest still pending at review time.

## On the coverage comment

The `[Blocking]` comment asked for changed-function coverage over the four affected functions (`MobilenetSSDPP::decode`, `MpPalmDetection::decode`, `BoundingBox::updateCentroids`, `BoundingBox::logBoxes`) to reach 80%; it measured 108/145 (74.5%), with the named gap being the untested non-float32 branches of the two `decode()` type switches (`mobilenetssdpp.cc:263-270,272`, `mppalmdetection.cc:376-383,385`).

Tracing the source confirms the four new tests target exactly that gap:
- `MobilenetSSDPP::decode()` dispatches on `info->type` of the "num" tensor (index 0 in the test's tensor layout), and has 10 case labels (int8/uint8/int16/uint16/int32/uint32/int64/uint64/float32/float64). `ssdPpTrackEveryType`'s `decoded_types` array lists exactly these 10 types and, via `TypedTensors`, sets all four input tensors to each type in turn, so every case label is exercised.
- `MpPalmDetection::decode()` dispatches the same way on `input[0]`'s type and has the identical 10 case labels; `palmTrackEveryType` mirrors the same loop.
- `trackBoxBackBeforeThreshold`/`trackBoxBackAtThreshold` add coverage of `updateCentroids()`'s empty-`boxes` branch and the `consecutive_disappeared_frames >= consecutive_disappear_threshold` removal check (the constant is 100, per `tensordec-boundingbox.cc:528`), which the previous three tracking tests never exercised (they only ever pass frames with at least one box).

So the claimed 108/145 -> 132/145 (91.0%) improvement is consistent with what the added tests actually touch, even though I could not re-run gcov myself to confirm the exact figure. Since the [Blocking] comment is a plain issue comment rather than a GitHub review thread, there is no "Resolve conversation" control on it, and no explicit maintainer comment closes it out - but on the substance, the gap it named is addressed by production-code-unaffecting test additions, and I recommend treating it as resolved.

## Test correctness and determinism

- Every numeric input used in `ssdPpTrackEveryType`/`palmTrackEveryType` (box corners 0.0/1.0, class 0, score 1.0/10.0, offsets 48.0) is exactly representable in every one of the 10 tensor element types, including the narrowest (`int8_t`/`uint8_t`), so `newTypedData`'s cast-per-type conversion introduces no truncation/rounding across types - the tests are deterministic by construction, not by luck.
- `palmTrackEveryType` correctly avoids reusing the pre-existing `PalmDetectionTensors` class's -10.0 default score: casting a negative float to an unsigned integer type is undefined behavior in C++, and the new test's own comment explains this, using 0.0 defaults plus a raised threshold (`PALM_OPTION_STRIDE_32_ABOVE_HALF`, 0.6) instead of the default 0.5 - since sigmoid(0) == 0.5, a 0.5 threshold would not have filtered the non-detection anchors (0.5 < 0.5 is false) and would have produced spurious boxes; 0.6 correctly drops them. This is a subtle, correct fix to a real hazard, not just cosmetic.
- The "compare tracked output to untracked output using a label that spells the expected id" technique (already used by the three earlier tracking tests) is reused correctly and traced against the source: `logBoxes`/`draw` label formatting, `updateCentroids`'s `box->tracking_id != 0` gating logic (which is exactly what a garbage, non-zero `tracking_id` from a missing zero-initializer would defeat), and the centroid-id counter all line up with what the tests assert.
- `trackBoxBackBeforeThreshold` (99 empty frames leads to id 1, label X-1) and `trackBoxBackAtThreshold` (100 empty frames leads to id 2, label X-2) precisely bracket the >= 100 removal check traced above; the arithmetic (disappeared-frame count only increments on empty-box calls, and the removal check runs before that increment on each call) checks out to exactly this boundary.
- Resource handling (temp label files, `GPtrArray`, `TypedTensors`' owned buffers, decoder `pdata`) is released on every path, including early `ASSERT_TRUE` returns, so no leaks are introduced by the new tests.

## Would the new tests fail without the fix?

By code tracing: yes, for the reason the PR body gives. `updateCentroids()` treats a non-zero `tracking_id` on a freshly-decoded box as "already assigned" (`tensordec-boundingbox.cc:662,693`) and skips it from both the match and the new-centroid registration, so a garbage non-zero value from an uninitialized `detectedObject` (pre-fix `detectedObject object;`) would leave the box's `tracking_id` un-set and produce a pixel-different frame that `ssdPpTrackEveryType`/`palmTrackEveryType`/the two threshold tests would catch via `memcmp`. One inherent caveat (non-blocking): this class of regression test depends on the compiler actually leaving a non-zero value on the stack when the initializer is removed, which is technically build/optimization-level dependent (unlike a memory-safety bug that a sanitizer/valgrind would catch unconditionally). The PR body states the tests were independently re-verified to fail by reverting the zero-init on the real CI toolchain ("without the zero-init every tracking case fails in every type"), which is the right way to establish this, and CI's fixed compiler/flags make the result reproducible in practice - flagging only so the inherent fragility of this style of test is visible for future reference, not as something to change in this PR.

## Other checklist items (recap, no new issues found)

- Fix is minimal (4 lines across 3 files); this round changes only the test file, no production code - proportionate to the stated goal (closing the coverage gap the reviewer named).
- No header/API changes, so no architecture/documentation update is required.
- No other module is touched; the two `box_properties/*.cc` files are decoder-internal and the test file is decoder-local.
- DCO sign-off present, no `CHANGES` edits, Doxygen brief/details comments present on all new functions/classes/tests.

## Findings

1. (non-blocking, informational) The zero-init regression tests' ability to fail without the fix relies on the compiler leaving a non-zero garbage value in the uninitialized `detectedObject`, which is inherently toolchain/optimization-dependent rather than guaranteed by the language. Not something to fix in this PR - the PR's own revert-and-rerun verification on CI's toolchain is the correct mitigation - just worth keeping in mind if this test ever flakes on a different compiler/config.
2. (nit) `trackBoxBackBeforeThreshold`/`trackBoxBackAtThreshold` hardcode the `consecutive_disappear_threshold` value (100) as loop bounds rather than referencing a named constant from the production header; this is unavoidable since the constant isn't exposed to tests, but it does mean a future change to that literal in `tensordec-boundingbox.cc` needs the corresponding test literals updated in lockstep (the test will fail loudly if that's missed, which is the intended safety net, not a real defect).

No blocking issues found. The four new tests are correct, deterministic, and target precisely the coverage gap the earlier `[Blocking]` comment identified, with no production-code changes in this round to introduce fresh risk.

Verdict: OK to merge

myungjoo:
Approved by Astra

PR #4989, opened 2026-09-17
[Common] Look a sub-plugin name up and insert it under one lock hold (#4920 A5)

myungjoo:
- **What:** fixes item **A5** of #4920. When two threads register the same sub-plugin name at once, `register_subplugin ()` double-frees the key and the stored entry, leaving a freed value in the registry. Application threads reach this through `NNS_custom_easy_register ()`, `nnstreamer_{decoder,converter,if}_custom_register ()`.
- **How:** the name is looked up and inserted in one `splock` hold. A registrant that loses the race never puts its entry into the table; the entry is destroyed after the lock is released, and the call returns FALSE with the existing "already registered" warning. A sequential duplicate is refused as before.
- **Result:** new `unittest_subplugin` (6 cases, 2 negative). The race case aborts main with `double free detected in tcache 2` in 20/20 runs (natively and on 2 CPUs) and passes 20/20 with the fix in about 0.45 s. Valgrind: about 2 s, 0 errors, 0 definite leaks. Full `meson test` in the local container: 26/26 with the fix; main fails only the new binary. No other module is changed.

---

- Old code: `_get_subplugin_data ()` check (lock hold 1) → `g_hash_table_insert ()` (lock hold 2). On a duplicate key, GLib frees the passed key (key destroy `g_free`), destroys the *old* value, stores the new one and returns FALSE. The error branch then called `_spdata_destroy ()` on the stored value and `g_free (sp_name)` again.
- New code: `g_hash_table_contains ()` + insert under one hold. The early `_get_subplugin_data ()` check is kept because it also creates the per-type table.
- Test design: 16 persistent threads plus a round counter under a mutex/cond, so all threads call `register_subplugin ()` at the same time. The threads are not created per round, so valgrind cost dropped from 92 s to 2 s. Warnings are dropped with a default-domain `g_log_set_handler` during the race case only. Threads are serialised under valgrind, so the race is gated by the native `meson test`/GBS/pdebuild runs; the valgrind step only checks the paths.
- Pinned to one CPU, main aborts in 9/20 runs; CI runners have ≥ 2 CPUs.
- `tests/common/unittest_common.cc` and `tests/nnstreamer_plugins/unittest_plugins.cc` belong to open PRs #4980/#4963, hence the new binary. `run_unittests_binaries.sh` finds `unittest_*` recursively, so packaging needs no change.
</details>


myungjoo:
**Note:** This is a relayed review written by a separate Claude agent (not the account posting this comment), asked to do an independent code review of this PR. Posting it verbatim as a comment per the requester's process.

## Scope reviewed
`gst/nnstreamer/nnstreamer_subplugin.c` (the fix), `tests/nnstreamer_subplugin/unittest_subplugin.cc` (new), `tests/meson.build` (registration), plus the surrounding registry code (`_get_subplugin_data`, `_spdata_destroy`, `unregister_subplugin`) and call sites in `gsttensor_{decoder,converter,if}.c` / `tensor_filter_custom_easy.c`.

Yes. Before the fix, `register_subplugin()` checked for a duplicate with `_get_subplugin_data()` under one lock hold, dropped the lock, then called `g_hash_table_insert()` under a second lock hold. On a name collision, GLib's `g_hash_table_insert()` already frees the passed-in key and destroys/replaces the old value; the old code additionally called `_spdata_destroy()` on the just-inserted value and `g_free()` on the already-freed key, corrupting the table and double-freeing. The fix (`ret = !g_hash_table_contains(...); if (ret) { ...insert... }` under one lock hold) makes "check + insert" atomic, so a losing thread's entry never reaches the table and is destroyed only after the lock is released. I traced the code path and the logic is correct: `subplugins[type]` is guaranteed non-NULL by the earlier `_get_subplugin_data()` call before `g_hash_table_contains()` is reached, so there's no null-table risk, and the switch-statement above already rejects invalid `type` values before that point.

## Regression risk on other modules
Low. The change is local to one function and does not alter `register_subplugin()`'s signature or return-value contract (still `TRUE` on success / `FALSE` on duplicate or invalid input). All in-tree callers (`gsttensor_decoder.c`, `gsttensor_converter.c`, `gsttensor_if.c`, `tensor_filter_custom_easy.c`, `tensor_filter_common.c`, and the many `tensor_filter_subplugin::register_subplugin<T>()` C++ wrappers) only branch on the boolean return, so they're unaffected. One behavior change worth flagging (severity: **Low/Info**): the duplicate-registration log level changed from `ml_loge` ("Failed to add subplugin data into the table.") to `ml_logw` ("Subplugin %s is already registered.") — this now matches the message already used by the early-check branch a few lines above, so it's a consistency improvement, not a regression. I grepped the tree and found no test or code that asserts on the old critical-level message.

## Code size / scope
Appropriately scoped: +8/-5 in the fixed function, no other module touched. The new test file (254 lines) is proportionate to the bug's severity (a double-free/process-abort reachable from application threads via public custom-registration APIs) and is isolated in its own new directory/binary, so it doesn't perturb `tests/common` or `tests/nnstreamer_plugins`, which the PR description says currently belong to other open PRs (#4980/#4963) — a reasonable reason to avoid touching those files here.

## Test coverage
Good. `registerSameNameConcurrently` reproduces the original race with 16 persistent threads doing 500 synchronized rounds, checking that exactly one registrant wins each round and that `get_subplugin()` returns that winner's data — this is a real regression test: it aborts pre-fix (double free) and passes post-fix. The remaining five cases (`registerUnregisterCycle`, `registerDuplicate_n`, `registerSameNameOtherType`, `registerInvalidArgs_n`, `unregisterUnknown_n`) pin down the sequential contract (duplicate refused and first data kept, per-type namespacing, reserved names / unknown type / NULL args refused, unregister-once semantics) that the earlier check-then-insert restructuring could otherwise have silently changed. I confirmed `gst-indent`/`clang-format` produce no diff on the changed files, so style is compliant, and every new function/TEST has a Doxygen comment per CLAUDE.md's convention.

Minor nit (severity: **Low**): `registerSameNameConcurrently` suppresses warnings via `g_log_set_handler(NULL, G_LOG_LEVEL_WARNING, ...)`, which only intercepts `ml_logw` on the GLib/Linux build path (`ml_logw` maps to `dlog_print`/`__android_log_print` on Tizen/Android). This is cosmetic log noise only on those platforms if the gtest binary is ever built there, not a correctness issue.

## Future-regression detection / CI gating
The new `registerSameNameConcurrently` test is exactly the kind of test that would catch a future re-introduction of this race if someone touches the locking again. I confirmed `unittest_subplugin` is registered as an ordinary `test()` in `tests/meson.build`, so it runs under the CI's plain `meson test -C build/ -v` (`ubuntu_clean_meson_build.yml`) and is not excluded from the valgrind pass (`run_unittests_binaries.sh --valgrind --recursive --skip unittest_filter_llamacpp`), so a regression here would fail CI and block the merge, matching the PR description's claims.

## Architecture / API docs
No architectural or public API change (function signature and return contract are unchanged; this is an internal locking fix), so no documentation update is required, and none was needed in `Documentation/writing-subplugin-tensor-filter.md`, which only describes usage, not internal locking. Correctly, `CHANGES` was not touched.

## Verdict
**Mergeable as is.** I found no correctness, scope, or regression issues; the two points above are informational/cosmetic only and don't need to block this PR.

myungjoo:
Thanks for the review. Both NITs were real; here is what changed at head `073f1502` (the test commit was amended, the fix commit `ba4a48f6` is untouched).

**NIT 2 (`unregisterUnknown_n` order dependence): taken.** The case now registers `unittest_subplugin_once` first, so the `NNS_CUSTOM_DECODER` table exists before the unknown-name lookup. That also fixes what the case actually tested: run alone it used to take the `g_return_val_if_fail (subplugins[type], FALSE)` early return ("the type has no table yet") instead of the "name is not in the table" path its name promises, so the assertion it pinned depended on the run order. Verified in the container: alone it now logs no critical message and passes with `G_DEBUG=fatal-criticals` (it aborted with exit 133 before), the whole binary passes 10/10 and 5/5 shuffled seeds, valgrind reports 0 errors / 0 definite lost, and main still aborts 10/10 with the race case.

**NIT 1 (the race case is probabilistic): not changed, on purpose.** Measured on main: 20/20 aborts natively and on 2 CPUs, 9/20 pinned to a single CPU, plus your own 10/10 - and the case runs natively in `meson test`, GBS x86_64 and both pdebuild jobs, so a regression has several independent chances per CI run. Making it strictly deterministic needs either a test-only hook inside `register_subplugin ()`, which puts test scaffolding into the function the PR is meant to keep minimal, or overriding a GLib symbol the registration path calls (for example `g_datalist_init`) to park the first thread inside the window. The override depends on an internal call order that can change silently, and it does not take effect on macOS (two-level namespace) or in a static build, where the test would hang instead of failing - so it would need a timeout fallback as well. The cost outweighs closing a gap that is already covered several times per run.

myungjoo:
**Note:** This is a relayed follow-up (round 2) review written by a separate Claude agent (not the account posting this comment), asked to independently re-verify this PR after the amendment. Posting it verbatim per the requester's process.

## Scope reviewed

Head `073f1502` (2 commits). Compared against the previously-reviewed head `b7e4b9c1` with `git range-diff` and a direct blob diff:

- `ba4a48f6` (the fix) is **byte-identical** between the two heads — confirmed via `git range-diff main~2..b7e4b9c1 main~2..073f1502`, which shows `1: ba4a48f6 = 1: ba4a48f6`.
- The only change is in `tests/nnstreamer_subplugin/unittest_subplugin.cc`, `unregisterUnknown_n`: it now calls `register_subplugin(NNS_CUSTOM_DECODER, "unittest_subplugin_once", &data)` before the unknown-name lookup, instead of after. That is a 1-line reorder (+1/-2), nothing else changed.

## Re-verification of the two round-1 NITs

**NIT 2 (order dependence) — taken, and I confirmed the fix actually closes the gap.** `unregister_subplugin()` (`gst/nnstreamer/nnstreamer_subplugin.c:292`) has `g_return_val_if_fail (subplugins[type], FALSE)` before the real "not found" check. Previously, if `unregisterUnknown_n` were the first test to touch `NNS_CUSTOM_DECODER` (e.g. run alone via `--gtest_filter`), that first call hit the *no-table-yet* critical-log path, not the "name not registered" path the test name promises. Registering a name into that type first guarantees the table exists, so the assertion now pins the path it claims to. I rebuilt the PR head and confirmed directly:
  - `--gtest_filter="*.unregisterUnknown_n"` alone: passes, no `CRITICAL` in the log.
  - Same, with `G_DEBUG=fatal-criticals`: exit 0 (this would have been exit 133/134 before the reorder).
  - Full binary x10 and `--gtest_shuffle` with 5 different seeds: all exit 0.
  - `clang-format` and the repo's doxygen-tag checker report no diff/no violations on the changed test file.

**NIT 1 (race case is probabilistic) — declined, and I think the reasoning holds up.** The argument is that CI gets several independent shots at it (native `meson test` on multi-core runners, GBS x86_64 with unit tests, two pdebuild jobs), the reproduction rate measured is already very high (20/20 native/2-CPU, 9/20 pinned to one CPU), and forcing strict determinism would need either test-only hooks inside `register_subplugin()` (fighting the "keep the fix minimal" goal) or overriding an internal GLib call the registration path happens to make (fragile to internal-order changes, doesn't work under macOS two-level namespace or static linking, and would need a timeout fallback). That's a legitimate cost/benefit call for a race test, not a rationalization — I did not find a cheap way to make it deterministic that the reply overlooked.

## Independent reproduction (fresh, not reusing round-1's numbers)

I rebuilt both the PR head and a pre-fix tree (head fix commit `ba4a48f6` reverted, keeping the current test file) in a clean container and ran:

- PR head, full binary, native, x10: `0 0 0 0 0 0 0 0 0 0`.
- Pre-fix tree (same test file, fix commit removed), full binary, native, x10: `134 134 134 134 134 134 134 134 134 134` (SIGABRT — reproduces the double free the fix addresses).
- Pre-fix tree, `unregisterUnknown_n` run alone, x10: `0 0 0 0 0 0 0 0 0 0` — confirms the test-reorder fix for NIT 2 is orthogonal to the race fix, as expected.
- Valgrind (`--leak-check=full`) on the PR head binary: exit 0, `ERROR SUMMARY: 0 errors`, `definitely lost: 0 bytes`.
- `gst-indent` on `gst/nnstreamer/nnstreamer_subplugin.c`: no diff.

This matches what the PR author reported in the follow-up comment; I was able to reproduce it independently rather than just trusting the numbers.

## Re-checked against the review criteria

- **Solves the stated topic (A5 of #4920):** Yes, unchanged from round 1 — the fix commit is untouched. `register_subplugin()` now checks-and-inserts under one `splock` hold (`g_hash_table_contains` + conditional `g_hash_table_insert`), so a losing thread's entry never reaches the table and is destroyed once, outside the lock. I re-traced this by hand against the old code path and it matches the commit message's account of the double free.
- **Regression risk to other modules:** None found. Signature and TRUE/FALSE contract of `register_subplugin()` are unchanged; all in-tree callers only branch on the boolean.
- **Size/scope:** Still minimal and on-topic: 3 files, +271/-5 total (fix +8/-5, tests/meson.build +10, new test file +253). No unrelated module touched.
- **Test completeness:** Better than round 1 — the one case that could pass on unfixed code for the wrong reason (order-dependent `unregisterUnknown_n`) is fixed and I verified it now pins the "name not found" contract regardless of run order.
- **Detects future regressions in this area:** Yes — `registerSameNameConcurrently` reliably reproduces the abort on the old locking pattern (see reproduction above) and would do the same for a re-introduced race.
- **CI actually blocks a regression:** Confirmed by reading `.github/workflows/ubuntu_clean_meson_build.yml`. `meson test -C build/ -v` (the "build and unit test" step) has no `|| true`, so a SIGABRT there fails the step/required check on every OS in the matrix. One clarification worth noting (informational, not a defect in this PR): the separate Valgrind step's own comments say a test *failure* under valgrind does **not** fail that step by itself (only a memcheck error attributed to our own frames does) — so the actual gate for this specific regression is the plain `meson test` step, not the valgrind step. That's already sufficient since `meson test` runs first and is a required, unconditional gate; I mention it only for completeness since the PR body's CI-gating claim leans on the valgrind job's log line.
- **Docs/architecture:** No public API or architecture change; no documentation update needed, and none was made.

## Findings

No blocking or new issues found in this round. The two prior NITs are resolved appropriately (one fixed, one reasonably declined with sound reasoning), and I could not reproduce any remaining order-dependence or double-free path.

## Verdict

**Mergeable as is.**

myungjoo:
Approved by Astra

PR #4990, opened 2026-09-17
[gRPC] Check a received message against the caps before building a buffer (#4920 H1, H2)

myungjoo:
- **What:** `tensor_src_grpc` built a buffer from whatever a gRPC peer sent (#4920 H1, H2). A protobuf message declaring more tensors than it carries read past the repeated field. A flatbuf message without the `tensor` vector or a tensor's `data` crashed the server, and a count above the vector length (or negative) read past it. With either IDL, a message with fewer tensors or shorter data than the caps was pushed as is, so downstream elements read past the memories.
- **How:** Before building the buffer, each received message is now checked in two shared helpers on `NNStreamerRPC`. The declared count must not exceed the tensors carried and must equal the caps' `num_tensors`, and every tensor's data must be exactly its caps size (the elements only accept static caps). A flatbuf tensor must also have its data field. A message that fails is dropped with an error log and the stream continues; a failed `gst_tensor_buffer_append_memory ()` now drops the message too instead of pushing a partial buffer. No explicit flatbuf `Verify ()` is added because flatbuffers' gRPC serialization already verifies each received message; see the commit message.
- **Result:** 12 new `unittest_grpc` cases (3 positive, 9 `_n`) stream real messages to a `tensor_src_grpc` server through the generated stubs. On main, every `_n` case fails: 1 SIGSEGV, 2 SIGABRT, and the rest deliver the malformed buffer. With the fix all 22 cases pass. Two gRPC SSAT cases passed only because nothing was checked. The senders' unformatted `video/x-raw` negotiates 4-channel frames on GStreamer 1.24 under `3:640:480` caps, and the `frames-per-tensor=2` receiver declared two tensors for the sender's one `3:640:480:2` tensor. The SSAT now pins `format=RGB` and declares the one tensor; the golden bytes are unchanged.

### Behaviour change
A peer whose messages do not match the receiver's caps (tensor count or byte size) no longer gets its data through; each such message is dropped with a `Failed to get tensors, ...` error. Such streams were already caps violations that downstream elements read out of bounds.

### Test coverage notes
- The CI meson jobs build without gRPC; `unittest_grpc` and the gRPC SSAT run in the per-PR Tizen GBS job (`grpc_support 1`, `%check`). `tensor_src_grpc` opens the IDL library by name, so `%check` now runs `./tests` with `LD_LIBRARY_PATH=./ext/nnstreamer/extra` (the SSAT script already exports it; `meson test` adds it by itself).
- Verified locally in an Ubuntu 24.04 container (gRPC 1.51.1, protobuf 3.21.12, flatbuffers 2.0.8): `unittest_grpc` 22/22, and each new case run in isolation on a tree with main's `ext/nnstreamer/extra`.

---

- Issue #4920 items H1 + H2. H3 (async client `delete this`) and H4 (grpc_start failure leak) are untouched.
- Files: `ext/nnstreamer/extra/nnstreamer_grpc_{common,protobuf,flatbuf}.{cc,h}`, `tests/nnstreamer_grpc/{unittest_grpc.cc,runTest.sh}`, `tests/meson.build` (the test binary now compiles `grpc_pb_gen_src`/`grpc_fb_gen_src` and links `grpc_support_deps` + `protobuf_util_dep`/`flatbuf_dep`), `Documentation/component-description.md`.
- The flatbuf test client does not call `ClientWriter::Finish ()`: it aborts in `cq_.Pluck (&finish_ops_)`, the same issue as the TODO in `SyncServiceImplFlatbuf::_client_thread ()`. It waits for the buffers and lets the context cancel the call.
- `recvFlatbufNoData_n` also fails on main through the size check; the explicit `data ()` NULL check is there because calling `data ()` on a NULL vector is UB.
- Valgrind on the new cases: 0 errors. Definite leaks are one `create_instance ()` object per started `tensor_src_grpc` server, and main leaks the same one per case, so this PR does not cause it (the valgrind CI job builds without gRPC). Not fixed here.
- The `nns-g16:deps` image (PPA gRPC 1.30 + flatbuffers 23.5) cannot build gRPC/flatbuf even on main (`grpc::Slice::sub`, `-lcares`); use `ubuntu:24.04` distro packages.
</details>


myungjoo:
*This review was written by another Claude agent (a separate session tasked with an independent code review of this PR) and is being relayed here on the reviewer's behalf.*

## Scope of review

Checked whether the fix actually addresses issue #4920 items H1/H2, whether it can regress other modules, whether the change size/footprint is proportionate, and whether test coverage (and CI wiring) is adequate to catch future regressions in this exact code path.

## Analysis

**Correctness of the fix (H1/H2):**
- `NNStreamerRPC::_check_tensor_count()` (`ext/nnstreamer/extra/nnstreamer_grpc_common.cc:182-201`) rejects a message whose declared tensor count exceeds what it actually carries, and additionally requires an exact match with the negotiated caps' `num_tensors`. This is called *before* any per-tensor access in both `ServiceImplProtobuf::_get_buffer_from_tensors()` (`nnstreamer_grpc_protobuf.cc:106-115`) and `ServiceImplFlatbuf::_get_buffer_from_tensors()` (`nnstreamer_grpc_flatbuf.cc:113-122`), so by the time the per-tensor loop runs, the loop bound is guaranteed `<=` both the actual repeated-field/vector length and `config_->info.num_tensors`. This closes the H1 OOB-read path in both IDLs.
- `_check_tensor_size()` (`nnstreamer_grpc_common.cc:203-217`) is called for every tensor before `_g_memdup`/`gst_tensor_buffer_append_memory`, rejecting any tensor whose byte size doesn't exactly match the caps-derived size, closing the H2 "shorter-than-caps data read past memories downstream" path.
- Verified `gst_tensor_buffer_append_memory()` (`gst/nnstreamer/nnstreamer_plugin_api_impl.c:1954-1957`) always takes ownership of (and frees) the `GstMemory*` passed to it, even on failure ("Release incoming memory even if failed to append it into buffer."). So the new `goto error` path after a failed append does not leak or double-free — `gst_buffer_unref(*buffer)` on the partially-built buffer is correct and sufficient.
- For the flatbuf "missing tensor vector" case (`recvFlatbufNoTensorVector_n`), `VectorLength(tensors->tensor())` is called on what can be a null vector pointer *before* any null check. This relies on flatbuffers' `VectorLength()` being null-safe (returns 0 for a null vector), which is the documented/standard behavior of that helper and is consistent with the codebase's own pre-existing use of the same helper pattern (`ext/nnstreamer/tensor_converter/tensor_converter_flatbuf.cc:148`, guarded by a preceding null check for defense-in-depth, not because `VectorLength` itself is unsafe). I could not independently build the gRPC/flatbuf target in this environment (Windows worktree, gRPC ext not built here) to confirm at the binary level, but the logic is sound given that contract, and the PR author states this exact case was run and passed locally (Ubuntu 24.04, flatbuffers 2.0.8).
- `tensor_src_grpc`/`tensor_sink_grpc` only advertise static `other/tensor(s)` caps (`CAPS_STRING` = `GST_TENSOR_CAP_DEFAULT "; " GST_TENSORS_CAP_DEFAULT`, no `GST_TENSORS_FLEX_CAP_DEFAULT`), so the new "exact match with caps" requirement cannot regress a legitimate flexible/variable-tensor-count use case for these two elements — there isn't one.

**Regression risk to other modules:** None found. The two new protected helper methods live only on `NNStreamerRPC`/its two IDL subclasses; no other module calls `_get_buffer_from_tensors()`, `_check_tensor_count()`, or `_check_tensor_size()`. The only externally observable behavior change is that a peer whose message doesn't match the negotiated caps now gets it dropped (logged) instead of corrupting the pipeline — this is the intended fix, not a regression.

**Size/footprint:** Proportionate. Source-level fix is small (~90 lines across 6 files, plus a 1-line doc update); the bulk of the diff (481 of 593 added lines) is new test code, appropriate for a memory-safety fix that needs both-IDL x multiple-malformed-shape coverage. No unrelated files or modules are touched.

**Test coverage:** Good. 12 new `unittest_grpc` GTest cases run real messages through the generated client stubs against a live `tensor_src_grpc` server (both positive delivery and 9 negative/dropped cases across protobuf and flatbuf, including the `NNS_TENSOR_MEMORY_MAX`-crossing "extra tensors" boundary). The SSAT caps fix (`tests/nnstreamer_grpc/runTest.sh`) is a legitimate, independently-justified correction (verified against `GST_TENSOR_CAP_DEFAULT`/converter `frames-per-tensor` semantics), not a workaround to make the new check pass.

**Minor nit — PR description accuracy (not a code issue):** The PR body says "12 new `unittest_grpc` cases (2 positive, 10 `_n`)". Counting the actual `TEST()` macros added: `recvProtobufStatic`, `recvProtobufExtraTensors`, `recvFlatbufStatic` have no `_n` suffix (3 positive), and the other 9 do (`recvProtobufMissingTensor_n`, `recvProtobufCountMismatch_n`, `recvProtobufSizeMismatch_n`, `recvProtobufExtraTensorSizeMismatch_n`, `recvFlatbufNoTensorVector_n`, `recvFlatbufMissingTensor_n`, `recvFlatbufNegativeCount_n`, `recvFlatbufNoData_n`, `recvFlatbufSizeMismatch_n`). So it's 3 positive / 9 negative, not 2/10. Cosmetic only — worth a quick fix to the PR body for future readers, no action needed on the code.

**Informational — CI coverage is better than the PR body's phrasing suggests (not a defect, just a clarification worth recording):** The PR's "Context for agents" note says these tests "run only in the Tizen GBS build (`grpc_support 1`)", which on its own could read as "not exercised by regular per-PR CI." Checked `.github/workflows/gbs_build.yml`: it triggers on every `pull_request` into `main`, and its `x86_64` matrix leg builds with `--define "unit_test 1"`. `packaging/nnstreamer.spec` defines `grpc_support 1` by default (line 31) and is only forced to 0 for the `tv` profile, `mv_prj`, or Tizen < 6.5 — none of which apply to the `profile.tizen` "unified" repo set used by `.github/workflows/tizen.gbs.conf`. The `%check` section runs `run_unittests_binaries.sh` over `./tests` (which would pick up the new `unittest_grpc` binary) and `ssat -n ... -cn _n` (which runs the updated `runTest.sh`). The `check-rebuild` gate treats any changed non-doc source file as requiring a GBS rebuild, so this PR's own files trigger it. Net effect: contrary to a literal reading of the note, this fix's regression tests should already run automatically on every future PR via the existing GBS CI job, not only in ad-hoc/manual Tizen builds — this is a genuinely positive property of the PR, just worth surfacing since the PR's own wording undersells it.

## Verdict

No blocking or non-blocking functional issues found. Only a documentation nit (PR body test-count breakdown) and one informational clarification (CI coverage is actually adequate, better than the PR body implies).

**OK to merge.**

myungjoo:
*This review was written by another Claude agent (a separate session tasked with an independent second-round code review of this PR) and is being relayed here on the reviewer's behalf.*

## Scope of this pass

This is a second review round. I read the existing agent review already on this PR and checked whether its points were addressed, but reviewed the change independently rather than repeating it. I focused on the areas the first round did not examine closely, per the requester's instructions: the new test helpers in `tests/nnstreamer_grpc/unittest_grpc.cc` (timing, races, flakiness under a loaded CI runner, resource cleanup, the flatbuf client that skips `Finish()`), the `tests/meson.build` wiring (build correctness when only protobuf or only flatbuf is available, or when gRPC is on but neither IDL is), and the `runTest.sh` caps changes.

## Findings

**1. [Nit] Fixed ~5s polling budget in the new `_run_recv_test` helper — potential flakiness under a loaded CI runner**
`tests/nnstreamer_grpc/unittest_grpc.cc:508` and `:519`: both the "wait for the src pad to get caps" loop and the "wait for the expected buffer count" loop use `for (i = 0; i < 500 ...) g_usleep (10000)`, i.e. a hardcoded ~5-second budget. On a contended CI runner (parallel test workers, GBS chroot under load) this is tighter than the `TIMEOUT_SEC=10` convention already used for the SSAT side of this same test suite (`tests/nnstreamer_grpc/runTest.sh`). If the gRPC server bind/caps negotiation or the message round-trip is slow under load, `_run_recv_test` returns `G_MAXUINT` (caps timeout) or a stale `received` count (buffer-wait timeout), producing a spurious test failure unrelated to the fix under test. Suggested fix: raise the budget to match the `TIMEOUT_SEC=10` convention (or make it env-overridable), and/or log via `g_test_message`/stderr when the loop is exhausted so a CI failure is distinguishable from a real regression at a glance.

**2. Verified, not an issue: no data race on the shared `RecvData` after teardown**
I looked closely at `tests/nnstreamer_grpc/unittest_grpc.cc:530-533`, where `received = data.received;` and `sizes = data.sizes;` read the struct written by `_recv_new_data_cb` (invoked on the pipeline's internal streaming thread) without holding `data.lock`, right after `gst_element_set_state (pipeline, GST_STATE_NULL)`. This looked suspicious at first (a plausible TOCTOU/race pattern), but downward GStreamer state transitions (PLAYING→...→NULL) are synchronous by contract, and `gst_tensor_src_grpc_stop()` → `grpc_stop()` → `NNStreamerRPC`'s worker thread `join ()` (`ext/nnstreamer/extra/nnstreamer_grpc_common.cc:133`) is called synchronously as part of that teardown. So by the time `gst_element_set_state (pipeline, GST_STATE_NULL)` returns, the only thread that could write `data` has already been joined, and the unlocked read afterward is safe. No action needed; noting this so the point is not re-raised as a false positive in a future pass.

**3. Verified, not an issue: `tests/meson.build` builds correctly in all three IDL-availability combinations**
Traced the variable wiring end to end:
- `grpc_support_is_available`, `protobuf_support_is_available`, `flatbuf_support_is_available`, `grpc_pb_gen_src`/`grpc_fb_gen_src`, `grpc_support_deps`, `protobuf_util_dep` and `flatbuf_dep` are all set in `ext/nnstreamer/meson.build`/`ext/nnstreamer/extra/meson.build`/root `meson.build`, processed via `subdir('ext')` (root `meson.build:954`) before `subdir('tests')` (root `meson.build:1012`), so they are in scope when `tests/meson.build` runs.
- `tests/meson.build`'s new `if protobuf_support_is_available` / `if flatbuf_support_is_available` blocks only reference `grpc_pb_gen_src`/`grpc_fb_gen_src` inside their own guard, matching the same guarding pattern already used in `ext/nnstreamer/extra/meson.build`.
- The three combinations all compile: protobuf-only and flatbuf-only each get their respective `ENABLE_PROTOBUF`/`ENABLE_FLATBUF` project arg and matching `#ifdef` block in `unittest_grpc.cc`; if gRPC is on but neither IDL is available (a real, pre-existing possible configuration — `grpc-support`'s own `extra_deps` list is `[grpc_dep, gpr_dep, grpcpp_dep]` only, independent of protobuf/flatbuf), then both `ENABLE_PROTOBUF` and `ENABLE_FLATBUF` stay undefined, so `unittest_grpc.cc`'s entire new test block (guarded by `#if defined(ENABLE_PROTOBUF) || defined(ENABLE_FLATBUF)`) is compiled out, and `unittest_grpc_src`/`unittest_grpc_deps` correctly stay at just the base file and `nnstreamer_unittest_deps` — no dangling reference to undefined generator outputs.
- Also checked why the flatbuf branch links the raw `flatbuf_dep` while the protobuf branch links `protobuf_util_dep` (a wrapper around `libnnstreamer_protobuf.so`) — this asymmetry is correct, not an inconsistency: Protobuf message classes (`Tensors`/`Tensor`) need a compiled `.pb.cc` translation unit, which `protobuf_util_dep` supplies by linking `libnnstreamer_protobuf.so` (built from `pb_gen_src` + `nnstreamer_protobuf.cc`); FlatBuffers tables are header/template-only, so the raw `flatbuf_dep` is sufficient. No fix needed.

**4. Verified, not an issue: `runTest.sh` caps/dimension changes are semantically correct, not just an incidental workaround**
Cross-checked the `frames-per-tensor=2` dimension change (`other/tensors,num_tensors=2,dimensions=3:640:480.3:640:480,...` → `other/tensors,num_tensors=1,dimensions=3:640:480:2,...`) against `gst/nnstreamer/elements/gsttensor_converter.md:9`: "direct conversion of video/x-raw ... to [height][width][#Colorspace] tensor. (#Colorspace:width:height:frames-per-tensor) ... You may express `frames-per-tensor` to have multiple image frames in **a tensor**." This confirms `frames-per-tensor=2` on a video stream produces one tensor with dimension `[3:640:480:2]`, not two `[3:640:480]` tensors — the old caps in `runTest.sh` were factually wrong before this PR, and the fix in this PR is correct, not merely a workaround to appease the new stricter checks.

**5. Verified, not an issue: flatbuf test client skipping `ClientWriter::Finish ()`**
`tests/nnstreamer_grpc/unittest_grpc.cc`'s `_fb_send()` intentionally omits `writer->Finish ()`, with a comment pointing at the pre-existing `TODO` in `SyncServiceImplFlatbuf::_client_thread ()` (`ext/nnstreamer/extra/nnstreamer_grpc_flatbuf.cc:279-282`, not touched by this PR). Confirmed this is the same underlying flatbuffers-gRPC library limitation already present in production code, not a new issue introduced by the test; the PR body states it was run under valgrind (0 errors) locally, and the test's `_run_recv_test` polling for buffers plus its own `ClientContext`/`writer` destruction at scope-exit does not depend on `Finish ()` completing correctly. No action needed beyond what's already documented in the PR's "Context for agents" note.

## Overall assessment

Agrees with round 1: the H1/H2 fix logic in `nnstreamer_grpc_common.cc`/`_protobuf.cc`/`_flatbuf.cc` is correctly scoped, the memory-ownership handling on the new `goto error` paths is sound, and there is no cross-module regression risk (the two new protected helpers are only called from the two IDL subclasses). This second pass, focused on the test harness and build wiring specifically, found no blocking or non-blocking functional defects — only one nit (finding 1) worth a follow-up, and several points I verified came back clean rather than confirming a suspicion.

## Verdict

**OK to merge.** (One nit suggested above, not a merge blocker.)

myungjoo:
*This review was written by another Claude agent (a separate session tasked with an independent third-round code review of this PR) and is being relayed here on the reviewer's behalf.*

## Scope of this pass

This is a third review round. I read both existing agent reviews already on this PR (round 1 and round 2, both "OK to merge") and did not repeat their analysis. Since round 2, the head changed in two ways: (1) the `_run_recv_test` wait loops were raised from 500 to 1000 iterations of `g_usleep(10000)` (round 2's nit), and (2) `packaging/nnstreamer.spec`'s unit-test invocation now runs `LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests`, added because the previous head failed the "Tizen GBS build on Ubuntu (x86_64, unit_test 1)" job (`tensor_src_grpc` loads `libnnstreamer_grpc_{protobuf,flatbuf}.so` by name via `g_module_open()`, and the GBS `%build`-time invocation of `run_unittests_binaries.sh` — unlike `meson test` or the SSAT script — did not put the build tree's `ext/nnstreamer/extra` on the library search path). I focused on these two changes, and did a fresh independent pass over the rest of the diff.

## Findings

**1. Verified, not an issue: the wait-loop nit from round 2 is fully addressed**
`tests/nnstreamer_grpc/unittest_grpc.cc:508` and `:519` now both loop `i < 1000` with `g_usleep (10000)`, giving a ~10-second budget that matches the `TIMEOUT_SEC=10` convention already used by `tests/nnstreamer_grpc/runTest.sh`. Both loops (caps wait and buffer-count wait) were updated, not just one.

**2. Verified, not an issue: the `packaging/nnstreamer.spec` `LD_LIBRARY_PATH` fix is correctly scoped and does not break other test binaries**
- Path resolution: `run_unittests_binaries.sh` does `pushd build` before it ever execs a test binary (`packaging/run_unittests_binaries.sh:58`), and the relative path in `LD_LIBRARY_PATH` is resolved by the dynamic loader against the process's cwd *at exec time*, not at the time the variable was set. Since every entry is run with cwd already inside `build/`, `LD_LIBRARY_PATH=./ext/nnstreamer/extra` correctly resolves to `build/ext/nnstreamer/extra`, which is exactly where meson places `libnnstreamer_grpc_protobuf.so`/`libnnstreamer_grpc_flatbuf.so`. This is the same relative-path convention already used one line below it, `LD_LIBRARY_PATH=./tests/nnstreamer_edge` (`packaging/nnstreamer.spec:1091`).
- Blast radius: this line (`packaging/nnstreamer.spec:1084`) runs the whole `./tests` sweep non-recursively (no `-r`), so the added `LD_LIBRARY_PATH` entry is inherited by every top-level `unittest_*` binary in that sweep, not just `unittest_grpc`. I checked what actually lives in `ext/nnstreamer/extra` (`ext/nnstreamer/extra/meson.build`): at most `nnstreamer_protobuf`, `nnstreamer_flatbuf`, `nnstreamer_grpc`, `nnstreamer_grpc_protobuf`, `nnstreamer_grpc_flatbuf`, and `nnstreamer_python3` — all nnstreamer-prefixed names with no plausible collision against any other library those other unit tests would dlopen or link. So today this is safe.
- Other callers of `run_unittests_binaries.sh` that would exercise `unittest_grpc` without this env var: `debian/rules` builds with `-Dgrpc-support=disabled` explicitly (both configurations, `debian/rules:68,75`), so `unittest_grpc` is never built there. `.github/workflows/risc-v.yml` uses `grpc-support=auto` (the meson default) but never installs `libgrpc++-dev`/`libprotobuf-dev` in its `apt-get install` list, so gRPC support resolves to disabled there too — `unittest_grpc` is not built. `.github/workflows/ubuntu_clean_meson_build.yml` (the valgrind job) likewise never installs gRPC dev packages, matching the PR's own claim that "the valgrind CI job builds without gRPC." The only per-PR job that builds with `grpc_support 1` **and** `unit_test 1` together is `.github/workflows/gbs_build.yml`'s `x86_64` matrix leg (`--define "unit_test 1"`, `packaging/nnstreamer.spec:31` defines `grpc_support 1` by default), which is exactly the job this change targets. So no other current caller is left running `unittest_grpc` without the fix.
- Placement: given `%check` (really inline in `%build` in this spec — there is no `%check` section) is the only place this out-of-tree, `g_module_open`-by-name lookup problem manifests (regular `meson test` adds the build dir to the library path itself, and the SSAT script exports it explicitly), `packaging/nnstreamer.spec` is the right, and only, place for this fix — it is not something a test-local change could address.

**3. [Nit] The `LD_LIBRARY_PATH` fix is broader than the codebase's own established scoping convention**
`packaging/nnstreamer.spec:1084`. The same file already has a precedent for giving a test binary that needs an extra library path its own dedicated, narrowly-scoped invocation instead of widening a shared one — see `nnstreamer_filter_mvncsdk2`/`nnstreamer_filter_edgetpu` (lines 1094, 1097: `LD_LIBRARY_PATH=... bash %{test_script} ./tests/nnstreamer_filter_.../unittest_...`), and the comment in `packaging/run_unittests_binaries.sh` explicitly explaining that some binaries are called out separately "without the library paths they set up for each one." `unittest_grpc` cannot be isolated the same way by directory (it is built at the top level of `tests/`, since it is declared directly in `tests/meson.build` rather than a subdir's own `meson.build`, unlike the mvncsdk2/edgetpu binaries), but it could still be excluded from the general sweep with `-k unittest_grpc` and given its own `LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests/unittest_grpc` line, matching the existing pattern instead of widening the shared one. This is not a functional problem today (finding 2 above), but it does mean every other top-level unit test in this build now silently depends on nothing in `ext/nnstreamer/extra` ever colliding with something it needs — worth tightening for future-proofing, not a merge blocker.

## Independent pass over the rest of the diff

I re-verified (rather than re-read) the core fix logic against the actual current source, beyond what the earlier rounds state:
- `gst_tensor_buffer_append_memory()` (`gst/nnstreamer/nnstreamer_plugin_api_impl.c:1810-1953`): traced every `goto failed` path myself. The final block unconditionally does `if (memory) gst_memory_unref (memory);` regardless of which failure path was taken, and `new_memory`/`last_memory` are separately unmapped/unreffed only when non-NULL and not appended. So the callee always takes ownership of (and releases) the incoming `GstMemory*` on both success and failure, confirming the new `goto error; ... gst_buffer_unref (*buffer);` paths in `nnstreamer_grpc_protobuf.cc`/`nnstreamer_grpc_flatbuf.cc` neither leak nor double-free the memory tied to the tensor that failed the check.
- `_check_tensor_size()`/`_check_tensor_count()` in `ext/nnstreamer/extra/nnstreamer_grpc_common.cc` run before any allocation for the tensor under check (`_g_memdup` happens only after `_check_tensor_size` passes), so a rejected tensor never leaks a partial allocation, and a buffer that already had earlier tensors appended is released as a whole via `gst_buffer_unref (*buffer)`.
- Checked the odd corner case of a protobuf message declaring a negative `num_tensor`: `guint num_tensor = tensors.num_tensor ()` wraps a negative `int32` into a large unsigned value, which is then widened to `gint64` and compared in `_check_tensor_count` as `declared > carried` — since `carried` (the real repeated-field size) is always small, this is correctly rejected before the loop runs, so the wraparound is never used as a loop bound. (Flatbuf's equivalent case is covered by an explicit test, `recvFlatbufNegativeCount_n`; protobuf does not have an explicit "negative declared count" test, but the guint-wraparound reasoning above shows it is caught by the same `declared > carried` branch — a possible, non-blocking test-coverage nit if the author wants full parity between the two IDLs' negative-count paths, but not a defect.)
- `tests/meson.build`, the doc line in `Documentation/component-description.md`, and the `runTest.sh` caps corrections match what round 1/2 already verified; nothing further found there.

## Verdict

No blocking or non-blocking functional defects found. One nit (finding 3) on the scope of the `LD_LIBRARY_PATH` fix in `packaging/nnstreamer.spec`, and one optional coverage nit (protobuf negative-count test parity), neither a merge blocker.

**OK to merge.**

myungjoo:
*This review was written by another Claude agent (a separate session tasked with an independent fourth-round code review of this PR) and is being relayed here on the reviewer's behalf.*

## Scope of this pass

This is a fourth review round. I read all three existing agent reviews already on this PR (rounds 1-3, all "OK to merge") and did not repeat their analysis. Since round 3, the only code change is in `packaging/nnstreamer.spec`'s `%build`/test-run section, addressing round 3's nit about the `LD_LIBRARY_PATH` fix being broader than the codebase's own scoping convention:

```
-    LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests
+    bash %{test_script} -k unittest_grpc ./tests
+%if 0%{?grpc_support}
+    LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests/unittest_grpc
+%endif
```

I traced this against `packaging/run_unittests_binaries.sh` in detail, re-verified round 3's declined protobuf negative-count nit against the actual `.proto`/`.fbs` schemas, and did a fresh independent pass over the rest of the diff.

## Findings

**1. [Nit] The `%if 0%{?grpc_support}` gate for the dedicated `unittest_grpc` run can diverge from whether gRPC was actually built, causing a silent (exit-0) no-op instead of a caught failure**

`packaging/nnstreamer.spec:1084-1087`:
```
bash %{test_script} -k unittest_grpc ./tests
%if 0%{?grpc_support}
LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests/unittest_grpc
%endif
```

`%{?grpc_support}` is a spec-level macro (default 1; forced 0 only for `tv`, `mv_prj`, or Tizen < 6.5 — `packaging/nnstreamer.spec:31,80,110,138`) that is **not** wired into the meson command line at all — there is no `-Dgrpc-support=...` in the `meson setup` invocation (unlike `%{enable_flatbuf}` etc.). Whether gRPC actually gets built is instead controlled purely by meson's own `grpc-support` option, whose type is `feature` defaulting to `'auto'` (`meson_options.txt:25`), resolved in `meson.build:433-439` via `dependency('grpc'/'gpr'/'grpc++', required: false)`. Critically, `meson.build`'s generic feature-registration loop (`meson.build:790-811`) only raises a hard configure-time `error()` when the option is explicitly `enabled()`; under the default `auto`, if any of those three pkg-config dependencies is not found, it just logs `'@0@ is off because it is either not available or disabled'` and silently sets `grpc_support_is_available = false` — no configure failure.

So it is possible, without any explicit spec change, for `%{?grpc_support}` (spec macro) = 1 while `grpc_support_is_available` (meson) = false — e.g. the `grpc-devel` BuildRequires (itself gated by the very same macro, line 321-322) resolves as installed but its `.pc` files are broken/mismatched, or a partial/incompatible package. In that case `./tests/unittest_grpc` is never built. Tracing `run_unittests_binaries.sh`:
- `-f "${input}"` is false (no such file), `-d "${input}"` is false (no such directory), so control falls to the final `else` branch (`run_unittests_binaries.sh:157-161`): `filename=unittest_grpc`, `dirname=./tests`, and `find ./tests -mindepth 1 -maxdepth 1 -type f -executable -name unittest_grpc` returns nothing.
- `run_list` is called with an empty array, iterates zero times, and `ret` stays at its initial value `0` (line 130).
- The script prints nothing about a missing binary and `exit $ret` returns 0 — a silent "pass" while zero gRPC regression tests actually ran.

This is not a purely theoretical dead end: this exact regression-prevention property (making sure the new H1/H2 checks are exercised on every PR) is the stated purpose of this round's spec change, so a divergence here would quietly defeat it under the disguise of a green test step. In practice the risk is low today: the same `grpc_support` macro also unconditionally gates `%files grpc`'s `%{_libdir}/libnnstreamer_grpc.so` (`packaging/nnstreamer.spec:1444`, no wildcard), so if gRPC silently failed to build while `grpc_support=1`, the overall rpmbuild would still fail later at the `%files` stage — just with a confusing "file not found" error instead of a clear test failure, and only after the (falsely "passing") test step already ran. Suggested fix: fail loudly instead of silently, e.g. add a `test -x ./tests/unittest_grpc` (or equivalent) guard before invoking it inside the `%if 0%{?grpc_support}` block, or use `run_unittests_binaries.sh -f`-style existence assertion, so a future build-environment drift produces a clear "gRPC was expected but not built" error at the right place instead of an empty-but-successful test run. Not a merge blocker — current CI (per round 1/3's findings) reliably builds gRPC with `grpc_support=1`.

**2. Verified, not an issue: round 3's decision to skip a protobuf-side negative-`num_tensor` test is correct**

Checked `ext/nnstreamer/extra/include/nnstreamer.proto`: `message Tensors { uint32 num_tensor = 1; ... }` — the field is `uint32` at the schema/wire level (proto3), so the generated accessor/mutator pair is `uint32_t num_tensor() const` / `void set_num_tensor(uint32_t)`. There is no way to construct a "declared negative tensor count" for protobuf through the public API the way `recvFlatbufNegativeCount_n` does for flatbuf, because flatbuf's schema declares `num_tensor : int` (signed 32-bit, `ext/nnstreamer/extra/include/nnstreamer.fbs`) while protobuf's is unsigned from the schema itself. A protobuf client could still send an out-of-range large value (e.g. `UINT32_MAX`), but that is already the scenario `_check_tensor_count`'s `declared > carried` branch (`ext/nnstreamer/extra/nnstreamer_grpc_common.cc:182-190`) exists to reject, and is not a materially different code path from what `recvProtobufMissingTensor_n`/`recvProtobufCountMismatch_n` already exercise. Confirms this is an IDL-schema asymmetry, not a leftover test-coverage gap — no action needed.

**3. Verified, not an issue: the `-k unittest_grpc` skip and the single-binary invocation are both correctly scoped and resolve their paths as intended**

- `run_unittests_binaries.sh:17-26` parses `-k unittest_grpc` into `skip_tests=(unittest_grpc)`, and the directory-target `find` (`run_unittests_binaries.sh:155`) adds `! -name unittest_grpc` (an exact, non-wildcard match) to its `-mindepth 1 -maxdepth 1 ... -name "unittest_*"` query. Since `unittest_grpc` is declared directly at the top level of `tests/meson.build` (not nested in a subdirectory's own `meson.build`, unlike e.g. the mvncsdk2/edgetpu binaries), its build output sits directly under `build/tests/`, so this exact-name exclusion correctly removes only it from the general non-recursive sweep and nothing else.
- Both the general sweep (`./tests`) and the dedicated call (`./tests/unittest_grpc`) run inside the same script, which does `pushd build` (`run_unittests_binaries.sh:56`) before either path is resolved — so `./tests/unittest_grpc` correctly resolves to `build/tests/unittest_grpc`, matching where meson places it, and the `LD_LIBRARY_PATH=./ext/nnstreamer/extra` set by the outer spec line is a relative path resolved by the dynamic loader against the process's cwd at exec time (already inside `build/` by then), so it correctly reaches `build/ext/nnstreamer/extra`, matching the existing convention one line below (`LD_LIBRARY_PATH=./tests/nnstreamer_edge`, `packaging/nnstreamer.spec:1094`). This addresses round 3's finding 3 exactly as suggested (dedicated, narrowly-scoped invocation instead of widening the shared sweep's library path) without introducing a new path-resolution bug.
- Both the general-sweep call and the dedicated call return their real exit status (`run_entry`'s `return ${retval}` / `run_list`'s `ret=${entry_ret}` accumulation, and the final `exit $ret`), and RPM `%build` scriptlets execute under `set -e` semantics, so a real test failure in either call aborts the build immediately — the intended CI-gating property holds for any case where the binaries in question actually exist and are exercised.

## Independent pass over the rest of the diff

Re-verified (not merely re-read) the core H1/H2 fix logic in `ext/nnstreamer/extra/nnstreamer_grpc_common.cc`, `_protobuf.cc`, and `_flatbuf.cc` against the current source and found nothing beyond what rounds 1-3 already covered: `_check_tensor_count`/`_check_tensor_size` run before any per-tensor access or allocation, `gst_tensor_buffer_append_memory`'s failure path still consumes its `GstMemory*` argument so the new `goto error` paths do not leak or double-free, and the flatbuf `gint num_tensor` change (from `guint`) is necessary and sufficient for the negative-count case to be caught by the second (`declared != config_->info.num_tensors`) branch of `_check_tensor_count` rather than the first, since a negative `gint64` never satisfies `declared > carried` against a non-negative `carried`. `tests/meson.build`, `Documentation/component-description.md`, and `tests/nnstreamer_grpc/runTest.sh` are unchanged since round 2/3's review and match what was already verified there.

## Verdict

No blocking issues found in this round either. One nit (finding 1) on defensive hardening of the packaging-macro-vs-actual-build divergence for the new `%if 0%{?grpc_support}` test invocation — worth tightening but not required for merge given current CI's actual behavior, and already fully backstopped by a hard `%files` failure if it ever occurred. Finding 2 confirms round 3's declined test-coverage suggestion was the right call.

**OK to merge.**

myungjoo:
How the relayed review rounds were handled (head `bbe054ea`):

- Round 1: the test count in the PR body was wrong. Fixed (3 positive, 9 `_n`).
- Round 2: the ~5 s test wait was too tight for a loaded runner. Both wait loops now allow ~10 s, matching `TIMEOUT_SEC=10` in the gRPC SSAT.
- Round 3, `LD_LIBRARY_PATH` scope: applied. `%check` now skips `unittest_grpc` in the `./tests` sweep and runs it on a separate line, following the mvncsdk2/edgetpu pattern.
- Round 3, protobuf negative-count test: not added. `num_tensor` is `uint32` in `nnstreamer.proto`, so a negative count cannot be built (round 4 confirmed this).
- Round 4, `grpc_support` macro vs. the meson `auto` result: not changed. The datarepo, edgetpu and mvncsdk2 test lines use the same macro gate. If the macro says gRPC but meson did not build it, `%files grpc` fails the rpm build on the missing `libnnstreamer_grpc*.so`, so the gap cannot pass silently.
- A separate CI fix is folded into this commit. The first head failed the GBS x86_64 `unit_test 1` job: `tensor_src_grpc` opens its IDL library by name, and `%check` had no build-tree library path. Reproduced locally: without the path, 12 of 22 cases fail with `Error opening libnnstreamer_grpc_protobuf.so`; with it, all 22 pass.

myungjoo:
Correction to the round-3 item in my previous comment (new head `a01f8cca`):

I reverted the narrower `-k unittest_grpc` split. On `bbe054ea` the GBS x86_64 `unit_test 1` job printed `run_unittests_binaries.sh: line 21: readarray: command not found`: the Tizen build environment's bash has no `readarray`, so `-k` does nothing there. `unittest_grpc` then ran in the general sweep without the library path, and all 12 new cases failed. That is why no other `%check` line uses `-k`.

`%check` is back to `LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests`. That form passed `unittest_grpc` 22/22 on GBS x86_64 at `3401e60a`, before an unrelated armv7l "Alarm clock" cancelled the matrix. `ext/nnstreamer/extra` holds only this repository's protobuf/flatbuf/grpc/python3 helper libraries, which the other test binaries already load from the same build tree.

myungjoo:
*This review was written by another Claude agent (a separate session tasked with an independent fifth-round code review of this PR) and is being relayed here on the reviewer's behalf.*

## Scope of this pass

Fifth review round. I read all four prior agent reviews on this PR (rounds 1-4, all "OK to merge") and the two author comments, and did not repeat their analysis. Since round 4, the only change is in `packaging/nnstreamer.spec`'s `%check`/test-run section: round 4's `-k unittest_grpc` split plus a dedicated `%if 0%{?grpc_support}` line was reverted back to a single line:

```
-    bash %{test_script} -k unittest_grpc ./tests
-%if 0%{?grpc_support}
-    LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests/unittest_grpc
-%endif
+    LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests
```

The author's comment states this revert was needed because the Tizen GBS x86_64 job (head `bbe054ea`) failed with `run_unittests_binaries.sh: line 21: readarray: command not found` — the Tizen build environment's bash lacks `readarray`, so `-k` silently did nothing there, and `unittest_grpc` then ran without the library path and failed all 12 new cases. I focused on verifying this claim and the resulting diff/commit-message accuracy, plus a fresh check of the rest of the diff (unchanged since round 3/4).

## Findings

**1. Verified, not an issue: the `readarray` claim is correct and the revert is the right fix**

`packaging/run_unittests_binaries.sh:21` does use `readarray -td, skip_tests <<<"$tmp,"; unset 'skip_tests[-1]'; declare skip_tests;` inside the `-k|--skip` argument branch — this is the only place `-k` is implemented, and it is unconditionally exercised whenever `-k` is passed, regardless of target. `readarray`/`mapfile` require bash ≥ 4.0; if the Tizen GBS chroot's `bash` predates that (or is otherwise built without it), the line fails exactly as the quoted job-log excerpt shows, and since RPM `%build` scriptlets run under `set -e`-like failure propagation only for non-zero exits from the *script itself* (the script would go on to `run_unittests_binaries.sh`'s own error handling) — practically, the described outcome (the skip silently not applying, `unittest_grpc` running in the general sweep without the added `LD_LIBRARY_PATH`) is exactly what round 3/4 already showed happens when `unittest_grpc` runs without that env var (12/22 cases fail with `Error opening libnnstreamer_grpc_protobuf.so`, per round 4's own reproduction). So reverting to the single, broader `LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests` line is the correct, minimal fix given this environment constraint — there is no working narrower alternative available in this spec file today.

**2. Verified: current diff and commit message accurately reflect this state**

- `git diff` for `packaging/nnstreamer.spec` at head `a01f8cca` shows exactly one hunk, replacing the original `bash %{test_script} ./tests` line with `LD_LIBRARY_PATH=./ext/nnstreamer/extra bash %{test_script} ./tests` — no leftover `-k` or `%if 0%{?grpc_support}` block. This matches round 3's originally-accepted form.
- The commit message's closing paragraph ("Tizen %check run of ./tests now puts ./ext/nnstreamer/extra on LD_LIBRARY_PATH... meson test adds that directory by itself. A separate run for unittest_grpc alone is not possible there: the -k option of run_unittests_binaries.sh needs readarray, which the Tizen build environment's bash does not have.") matches the actual diff and the `readarray` usage confirmed in finding 1.
- The PR body's "Test coverage notes" section is also already in sync with this reverted state (no mention of `-k` or a separate `grpc_support` gate), and its test-count claim ("12 new unittest_grpc cases (3 positive, 9 _n)") matches the actual `TEST()` macros — round 1's original count nit is confirmed fixed.
- This PR remains a single commit (`a01f8cca`), consistent with the practical context.

**3. [Informational, not a new finding] Round 3's original non-blocking nit about `LD_LIBRARY_PATH` scope is now the permanent state, not a temporary one**

Round 3 flagged (non-blocking) that widening `LD_LIBRARY_PATH` for the whole `./tests` sweep, rather than scoping it to `unittest_grpc` alone, was broader than the codebase's own convention (e.g. the `nnstreamer_filter_mvncsdk2`/`edgetpu` dedicated lines). Round 4's attempt to narrow this scope had to be reverted because of the `readarray` limitation, so this PR now permanently ships the broader form rather than a temporary one pending a follow-up. Round 3 already judged the actual collision risk as negligible (only nnstreamer's own protobuf/flatbuf/grpc/python3 helper libs live in `ext/nnstreamer/extra`), and I found nothing to revise in that risk assessment. Making `-k` portable to the Tizen bash (e.g. replacing the `readarray` line in `run_unittests_binaries.sh` with an `IFS`-based split) would be a reasonable follow-up, but it touches a shared test-infra script used by many other packaging lines and is out of scope for this gRPC-focused PR — not a request for this PR, just noting the trade-off is now durable rather than transient.

## Independent pass over the rest of the diff

Re-verified the core fix is unchanged since round 3/4: `_check_tensor_count`/`_check_tensor_size` in `ext/nnstreamer/extra/nnstreamer_grpc_common.cc` still run before any per-tensor access/allocation in both `nnstreamer_grpc_protobuf.cc` and `nnstreamer_grpc_flatbuf.cc`; the wait loops in `tests/nnstreamer_grpc/unittest_grpc.cc:508,519` are still `i < 1000` with `g_usleep(10000)` (round 2's nit, fixed in round 3, still in place); `tests/meson.build`'s IDL-availability branches are unchanged from round 2's verification; `Documentation/component-description.md` still carries the one-line behavior-change note. No regressions found relative to prior rounds.

## Verdict

No blocking or non-blocking issues found in this round. One informational note (finding 3) that a previously-flagged non-blocking nit is now a permanent trade-off rather than temporary — not a merge blocker, and no fix is being requested within this PR.

**OK to merge.**

myungjoo:
Re the NIT in review 5231875884 (a message arriving between `start()` and `set_caps()` now logs an error): no code change. On main, that window already logged an error.

Before caps are set, `config_` is still in its `gst_tensors_config_init()` state, so every tensor info has `type == _NNS_END`. On main, `_get_buffer_from_tensors()` runs before `_grpc_callback()` and calls `gst_tensor_buffer_append_memory()` for each tensor. The received data has no meta header, so the function falls back to `gst_tensor_info_validate()`, which fails on `_NNS_END`, and `nnstreamer_plugin_api_impl.c:1840` logs `nns_loge ("Failed to get tensor info (invalid input info).")` once per tensor. Only after that does the callback see that the element is not CONFIGURED and drop the buffer.

With this PR, that window logs one error per message instead of one per tensor, and the text names the cause (`the caps have 0`). The only case that stayed silent on main was an empty message (`num_tensor == 0`), and that one is still silent here, since 0 equals the unconfigured caps' count.

myungjoo:
Approved by Astra

myungjoo:
Approved by Astra

PR #4943, opened 2026-09-08
[Element/Sparse] Bound the sparse tensor data and handle the buffers the config cannot produce

myungjoo:
Five memory-safety items of #4920 in the `tensor_sparse` element family, taken together because they live in the same three files. No open PR touches them.

## C13 / C15 — the sparse layout is taken from the data, never checked against it

`gst_tensor_sparse_to_dense()` (`gsttensor_sparseutil.c:26-107`) read the meta header of the memory it was decoding and trusted every field in it:

```c
if (!gst_tensor_meta_info_parse_header (meta, map.data))   /* reads 88 bytes, no size check */
  ...
output = (guint8 *) g_malloc0 (output_size);               /* sized from the dimensions alone */
nnz = meta->sparse_info.nnz;                               /* uint32, up to 2^32-1 */
input = map.data + gst_tensor_meta_info_get_header_size (meta);
indices = (guint *) (input + element_size * nnz);          /* up to 32 GB past the mapping */

for (i = 0; i < nnz; ++i)
  ((int32_t *) output)[indices[i]] = ((int32_t *) input)[i];  /* index is the write offset */
```

- **The header** is parsed from a memory that may be shorter than the 88 bytes `gst_tensor_meta_info_parse_header()` reads before it validates anything — out-of-bounds read.
- **`nnz`** is never compared with the length of the memory, so both `input[i]` and `indices[i]` are read far past it.
- **`indices[i]`** is the offset of a **write** into a block sized from the dimensions. An index of `0x20000000` in a 40-element int32 tensor writes 2 GB past a 160-byte allocation — an arbitrary-offset write whose value the same buffer supplies.

`gst_tensor_sparse_from_dense()` had the mirror hole: it read `element_count` elements out of the dense memory regardless of `map.size`.

Every field involved arrives with the data, so a crafted buffer from a file, an `appsrc`, `tensor_query` or `edge` decides all of them.

[...]

PR #4991, opened 2026-09-17
[Element/Sparse] Refuse a dense tensor that cannot be allocated (#4920 C37)

myungjoo:
This PR fixes item **C37** of #4920.

- **Problem:** `gst_tensor_sparse_to_dense()` allocated the decoded tensor with `g_malloc0()`, using the size the sparse header declares. `g_malloc0()` aborts instead of returning NULL, so a single 128-byte sparse buffer declaring `1000000000:1000000000` int32 kills the whole process (`GLib-ERROR ... failed to allocate 4000000000000000000 bytes`). The header's fields are all consistent, so none of the #4943 checks catch it.
- **Fix:** allocate with `g_try_malloc0()` and return NULL if the allocation fails. `tensor_sparse_dec` already turns NULL into `GST_FLOW_ERROR`. That is a 5-line change in one function.
- **Result:** a new `unittest_sparse` binary with 3 cases. Both `_n` cases abort (exit 133) on main's `gsttensor_sparseutil.c` and pass with the fix. The positive 4 MB decode passes on both. The existing 27 `testTensorSparse.*` cases in `unittest_plugins` still pass. clang-format 16, gst-indent, doxygen-tag (with ctags, negative-controlled) and a `-Wextra -Wsign-compare -Werror` compile are clean.

### Commits
1. `[Element/Sparse] Refuse a dense tensor that cannot be allocated`
2. `[Test/Sparse] Test a sparse tensor declaring an unallocatable size`

---

### Notes for reviewers and agents
- **Why not bound by the memory size:** a small sparse tensor legitimately decodes into a large dense one, so the `hsize + dsize == map.size` test other flexible-tensor consumers use does not apply here.
- **Why not also bound by `out_config` in `tensor_sparse_dec`** (the issue lists that as an "and/or" option): (a) `out_config` is valid only when downstream fixes the caps, so it cannot be the general guard. (b) The util would need a size-ceiling parameter, and its callers in `tests/nnstreamer_plugins/unittest_plugins.cc` belong to open PRs. (c) Today a size mismatch is dropped with `GST_FLOW_OK` after decoding, and a pre-decode refusal would change that. A size that can be allocated is freed right after the existing mismatch check, and with `nnz = 0` it touches no pages.
- **Other allocations in the file:** `gst_tensor_sparse_from_dense()`'s `values` / `indices` / `output` allocations are all bounded by the dense `map.size` that arrived (`element_count <= map.size / element_size`, `nnz <= element_count`), so they are not the same kind of defect and are unchanged.
- **Test placement:** `unittest_plugins.cc` belongs to open PRs, so the tests are in a new `tests/nnstreamer_sparse/unittest_sparse.cc`. The `tests/meson.build` hunk only adds that binary. On 32-bit targets the same header is already refused by the #4943 element-count check (`gulong` wraps), so the negative cases pass there too.
- Local tree: WSL `~/nns-4920-c37`. The SSAT `nnstreamer_sparse` group needs the lua sub-plugin and runs 0 cases locally. Valgrind cannot run on this box, so that job is CI-only.


myungjoo:

Verdict: mergeable

Scope reviewed: the PR diff (`gst/nnstreamer/elements/gsttensor_sparseutil.c`, `tests/nnstreamer_sparse/unittest_sparse.cc`, `tests/meson.build`), the surrounding `gsttensor_sparseutil.c`/`gsttensor_sparsedec.c` source, issue #4920 item **C37**, and the project's CI/test-registration setup. I also did an independent build+run in a scratch WSL checkout (not the PR author's own local tree) to confirm the new binary actually builds, registers with meson, and passes.

## Findings

No blocking or minor issues found. Below are the checks performed, each with a verdict, per the review checklist requested.

1. **Does it solve C37?** Yes. C37 is: `gst_tensor_sparse_to_dense()` sizes the decoded (dense) tensor from the sparse header's declared dimensions and allocates it with `g_malloc0()`, which aborts the whole process instead of returning NULL on failure — a 128-byte crafted sparse buffer declaring `1000000000:1000000000` int32 elements kills the process before any index is even read (`nnz = 0`). The fix swaps `g_malloc0()` for `g_try_malloc0()` and returns NULL (with an `nns_loge`) on failure. The sole caller, `gst_tensor_sparse_dec_chain()` in `gsttensor_sparsedec.c:391-397`, already treats a NULL return from this function as `goto done` with `ret` left at its initializer `GST_FLOW_ERROR`, so the buffer is refused with a normal flow error instead of aborting the process — verified this by reading the chain function directly. The issue's suggested fix was "bound by `out_config` and/or `g_try_malloc0()`"; the PR body gives a reasoned justification for why `out_config`-bounding doesn't generalize here (unlike other flexible-tensor consumers, a small sparse buffer legitimately decodes into a large dense tensor, so no fixed "hsize+dsize==map.size" check applies) and picks the `g_try_malloc0()` half, which the issue explicitly allows ("and/or"). This matches the actual defect described.

2. **Regression / cross-module impact.** `gst_tensor_sparse_to_dense()` has exactly one production call site (`gsttensor_sparsedec.c`) plus test call sites in `tests/nnstreamer_plugins/unittest_plugins.cc`; grepped the whole tree to confirm. The change is behavior-preserving on every allocation that already succeeds (`g_try_malloc0` zero-initializes on success identically to `g_malloc0`), so the 27 existing `testTensorSparse.*` cases in `unittest_plugins` are unaffected — confirmed by re-running them in a scratch build (27/27 pass). The sibling function `gst_tensor_sparse_from_dense()` (encoder path) is untouched and is already bounded by the incoming dense buffer's mapped size, so it isn't the same class of defect, as the PR body correctly notes. No other element, filter, or decoder touches this function, so there's no cross-module regression surface here.

3. **Code size vs. topic / unrelated-module touches.** The functional fix is exactly 5 lines in one function of one file. The remaining ~230 lines are a new test binary plus a 10-line `meson.build` hunk to register it — required by the repo's test policy for a bug fix (regression test that fails without the fix). Nothing outside `gsttensor_sparseutil.c` and the new test files is touched; no unrelated module is modified.

4. **Test completeness.** Three GTest cases in `tests/nnstreamer_sparse/unittest_sparse.cc`:
   - `utilToDenseHugeDimension_n` calls `gst_tensor_sparse_to_dense()` directly with the unallocatable header and expects NULL — the most direct regression test for the exact defect.
   - `utilToDenseLargeDimension` is a sanity/positive-path check (a large-but-allocatable 4 MB dense tensor still decodes correctly, zeroed except for the non-zero elements) — good for catching an overly aggressive fix that would refuse legitimate large tensors.
   - `decHugeDimension_n` drives the real `tensor_sparse_dec` element through a `GstHarness`, confirming `GST_FLOW_ERROR` (not a crash) and, importantly, that a valid buffer pushed immediately afterward is still decoded correctly — this also guards against a botched error path leaving the element (or its `outbuf`/`info` cleanup) in a bad state. I independently rebuilt and ran this binary in a separate scratch checkout; all 3 cases pass, and the CRITICAL log lines match the expected `nns_loge` path (not a GLib-ERROR abort).
   - The PR body states these two `_n` cases abort with exit 133 (SIGABRT via `g_malloc0`) on pre-fix code, which is consistent with documented GLib behavior (`g_malloc0` calls `g_error` internally on allocation failure, which is fatal) and with the mutation-test artifacts already present from the PR's own local verification.

5. **Detecting future regressions from other modules.** `decHugeDimension_n` exercises the real element via `GstHarness` end-to-end (caps negotiation, chain function, cleanup), not just the util function in isolation, so a future change to error propagation in `gst_tensor_sparse_dec_chain()`, to `gst_tensor_buffer_from_config()`, or a reversion of the allocator swap would be caught by this test, not only by the direct unit test.

6. **CI gate.** `unittest_sparse` is registered as a normal `test()` in `tests/meson.build` inside the same `if gtest_dep.found()` block as `unittest_plugins`/`unittest_common`, so it is picked up by the generic `meson test -C build/ -v` step that `.github/workflows/ubuntu_clean_meson_build.yml` already runs — confirmed via `meson test --list` in a scratch build (`nnstreamer:unittest_sparse` is listed). No special CI wiring is needed and none was added, which is correct — a CI run that reverts the fix would fail this test and block the merge.

7. **Architecture / API / docs.** No public header, API signature, or architecture is changed — `gst_tensor_sparse_to_dense()`'s signature and its documented "returns NULL on error" contract are unchanged (only the *cause* of a NULL return gains one more case). No documentation update is required or missing.

## Minor observations (non-blocking, not raised as findings)
- The new test file uses a relative `#include "../gst/nnstreamer/elements/gsttensor_sparseutil.h"` from `tests/nnstreamer_sparse/`. This looked suspicious at first glance (naively, `..` from that directory does not reach `gst/`), but it is the exact same pattern already used by `tests/nnstreamer_plugins/unittest_plugins.cc` and resolves correctly through the `tests/`-rooted include directory that `unittest_util_dep` already adds for every unittest binary. Confirmed it compiles in the scratch build. Not a new pattern introduced by this PR, so not flagged as an issue.

---
Verdict: mergeable — the change directly and minimally fixes C37, is well isolated, comes with regression tests that fail without the fix and pass with it (independently re-verified), and those tests are wired into the CI gate that already runs on every PR.

myungjoo:
On the residual risk the automated review noted (a declared size that *can* be reserved is still allocated before the downstream size-mismatch drop): that half is now tracked separately as item **I1** of the core security audit, #4960, so it is not lost when C37 of #4920 is ticked with this PR.

Why it is filed there rather than as another #4920 item: what is left is resource exhaustion, not memory corruption, which is the scope #4960 already covers for its siblings D2/E4/F2. Recorded as Medium by that issue's scale (unbounded resource use reachable from stream/peer data), with the mitigating facts written into the item: the writes stay in bounds, the block is freed per buffer, and 32-bit targets and `vm.overcommit_memory=2` refuse it outright.

The #4920 C37 line and the #4960 cross-reference row both point at this PR for the abort half.

This PR is unchanged by that: the fix here is what makes an unallocatable size a clean `GST_FLOW_ERROR` instead of a process abort.

myungjoo:
Approved by Astra

PR #4942, opened 2026-09-07
[Element/Crop] Bound the crop region and the incoming buffers

myungjoo:
Two memory-safety items of #4920 in `gst/nnstreamer/elements/gsttensor_crop.c`, taken together because they live in the same file.

## C11 — the region clamp wraps in 32-bit

`gsttensor_crop.c:676-681` bounds a crop region with `_x + w - 1 < mw` in `guint` arithmetic. `w` comes from the info tensor through a typecast into a `guint` field, so a negative width arrives as `G_MAXUINT32` and the sum wraps back under `mw`: the width survives unclamped and the element allocates and `memcpy()`s ~17 GB per row out of a small mapping. The same holds for the height.

The same expression yields `_w == 0` for a region that starts on or past the right edge, and for a zero-width region that does not start at 0, which `g_assert (_w > 0 && _h > 0)` turns into an `abort()`.

Fixed by clamping with `MIN()` on a subtraction that cannot underflow, so no addition is left to wrap. **A region of zero size keeps meaning the rest of the frame**, which is what the old expression did for `{0, 0, 0, 0}` and what the `tensor_region` decoder depends on: it allocates the crop-info tensor for `option1` regions, zeroes it and fills only the slots it has a detection for. A region left with nothing to crop after clamping is skipped rather than failing the whole buffer; only a buffer in which no region survives is an error.

Behaviour is byte-identical to the current `main` for every input the old code did not abort on.

## C12 — headers parsed past the mapping, and unchecked NULL results

- `gst_tensor_crop_prepare_out_meta()` and `gst_tensor_crop_get_crop_info()` parse a v1 meta header out of a mapped memory with no size check. `gst_tensor_meta_info_parse_header()` reads 84 bytes before it validates anything, so a shorter memory is read out of bounds. Both sites now check the header size first, mirroring `gst_tensor_meta_info_parse_memory()`.
- `gst_tensor_buffer_from_config()` may return NULL, and the chain function used the result unchecked — `GST_BUFFER_TIMESTAMP()` dereferences it while `lateness` is set. Now `GST_FLOW_ERROR`.
- The NULL result of the cropping was handed to `gst_pad_push()`, which rejected it with a `GStreamer-CRITICAL`. Now `GST_FLOW_ERROR`.
- The two `g_assert()`s on buffer content are reachable from a peer (a buffer with no memory; an info tensor whose element count is not a multiple of four) and are `GST_FLOW_ERROR` now.
[...]

PR #4992, opened 2026-09-17
[Decoder/TensorRegion] Clamp a crop region to the frame (#4920 G20)

myungjoo:
item **G20** of #4920

- **What:** the `tensor_region` decoder (mobilenet-ssd mode) sent `tensor_crop` crop regions that were never cut to the frame: regions past the right/bottom edge, the wrong pixels for a box crossing the left/top edge (x raised to 0 but the width kept), zero-width regions for sub-pixel boxes (which `tensor_crop` treats as "whole frame"), and an undefined float→int conversion when a size overflows to `inf` or a coordinate is NaN (x86-64: `2147483648`).
- **How:** each axis goes through a new `_clamp_to_frame ()`: non-finite → drop the box; negative start shortens the length; length cut at the frame edge; no whole pixel inside → drop the box (so it no longer uses up an `option1` region). Pixel products stay in `float` as before, so boxes inside the frame give byte-identical regions; the frame size is capped at `G_MAXINT` so the conversion is defined for any `option4`.
- **Result:** 15 new direct sub-plugin gtests in the existing `unittest_tensor_region`; 13 fail on main's decoder, and each of 4 mutations of the fix is caught. valgrind 0 errors / 0 definite leaks, SSAT `nnstreamer_decoder_tensor_region` 4/4 (golden crop unchanged). No other module touched; `tensor_crop` (the only consumer) already handles such regions since #4942.

---

**Files:** `ext/nnstreamer/tensor_decoder/tensordec-tensor_region.c` (helper + macro change + header doc for option4), `tests/nnstreamer_decoder_tensor_region/unittest_tensor_region.cc` (fixture `tensorRegionDecode`; the test file is CRLF in the repo blob and was kept CRLF). No `tests/meson.build` change.

**Behaviour changes (intended):** a box with no pixel in the frame, or a non-finite coordinate, is dropped instead of emitted; partially outside boxes are cut; a zero-width/height region is never emitted. An empty `option4` (frame 0) still yields all-zero output, as before.

**Test fixture:** labels `background,object`; box priors all `0.5` (2 columns); default scales, so offsets `{0,0,0,0}` = centered half-frame box `75,75,150,150` at 300x300, center offset 10 = half a frame, size offset `5 ln k` = size x k.

**Evidence (docker `nns-g20`, image `nns-g16:deps`, `-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`):**

| decoder | result |
|---|---|
| fix | 16/16 |
| main `f865ca74` | 13 new cases fail (passes: `regionInFrame`, `regionEmptyFrame_n`) |
| no `isfinite` check | `regionInfiniteSize_n`, `regionNotANumber_n` |
| no left-edge `l += p` | `regionPastLeftEdge`, `regionLeftOfFrame_n` |
| no cut at `limit - pos` | `regionPastRightEdge`, `regionPastTopAndBottomEdges`, `regionFrameFromOption`, `regionLargeFrame` |
| no `p >= limit` | `regionFarRightOfFrame_n` |

Style: clang-format 16 clean on the `.cc`; doxygen-tag clean; `indent.sh` (warning-only) reports the same 130 pre-existing lines on main and on the fix.

**Not changed, noted:** `box_properties/mobilenetssd.cc` (bounding_boxes) has the same unclamped `int x = xmin * i_width` shape, but that file is owned by open PR #4988 and bounding_boxes clamps at draw time since #4924 (G3).
</details>


myungjoo:

Scope reviewed: `ext/nnstreamer/tensor_decoder/tensordec-tensor_region.c` and `tests/nnstreamer_decoder_tensor_region/unittest_tensor_region.cc` as changed in this PR (diff only). I independently rebuilt this branch and ran the tests (not just re-read the PR description).

**1. Does it solve issue #4920 item G20?**
Yes. G20 asks that `_get_object_i_mobilenet_ssd()` clamp `x/y/width/height` to `[0, i_width]/[0, i_height]` in float before converting to int, reject non-finite values, and drop a box with no whole pixel inside the frame. The new `_clamp_to_frame()` does exactly this per axis, and the call site (`tensordec-tensor_region.c` around the `xmin/ymin` computation) now calls it for both axes and `break`s out of the label loop without setting `result->valid = TRUE` when either axis fails — so a rejected box is dropped rather than emitted, matching the issue's "Fix" section point for point.

**2. Regression / cross-module risk**
Low, and I verified this rather than just trusting the PR body:
- Ran `meson test -C build unittest_tensor_region -v`: 16/16 pass (1 pre-existing pipeline test + 15 new ones), matching the PR's claimed evidence table.
- Ran the SSAT group `nnstreamer_decoder_tensor_region`: 4/4 pass, confirming the golden in-frame crop output is byte-identical to before (the float pixel-product math is unchanged for in-frame boxes; only the clamping/rejection path is new).
- Traced the float/double math in `_clamp_to_frame()` by hand: `limit` is computed as a `double` from `MIN(frame, G_MAXINT)` *before* any cast to `int`, and both early-return checks (`isfinite`, `p >= limit || l <= 0`) happen before `*pos`/`*len` are ever written, so `(int) p` and the `MIN((double) l, limit - *pos)` cast are always within `int` range — the original UB (float→int on an out-of-range/NaN value) is genuinely gone, not just narrowed.
- The only other file matching the macro name `_get_object_i_mobilenet_ssd` is `box_properties/mobilenetssd.cc` (bounding_boxes decoder) — that's a separate, unrelated `#define` in a different translation unit (confirmed by grep), not shared code, so there is no cross-module coupling despite the name collision. The PR body correctly flags that file as intentionally out of scope (owned by #4988, already clamped elsewhere since #4924/G3).
- No other module, build file, or `CHANGES` entry is touched.

**3. Size / scope**
Right-sized for the bug: one static helper function + the call-site rewrite in the one file that owns the bug, plus tests. No unrelated refactoring or drive-by changes.

**4. Test completeness**
Very thorough for a decoder-level fix. The 15 new `TEST_F` cases exercise the sub-plugin directly (via `decoder->decode()`), independent of a full pipeline, and cover: in-frame passthrough (byte-identical), right/left/top/bottom edge clamping, fully-outside-frame on all four sides, sub-pixel width drop, NaN and infinite-size drop, empty `option4` (0-size frame), a frame wider than `INT_MAX` (the exact overflow scenario from the bug report), and that a dropped box does not consume an `option1` output slot from a lower-priority valid box. This maps directly to the enumerated failure modes in the G20 issue text, including the specific "float→int UB on overflow" case.

**5. Future-regression detection via these tests**
Yes. The fixture drives the public decoder vtable (`init/setOption/decode/exit`) directly rather than depending on model files or a full `gst-launch` pipeline, so it is cheap and will keep running for any future change to this decoder. Because assertions check exact pixel values (`expectRegion`), any future change to the clamp/crop math — by this file or, in principle, by shared plugin-API helpers it calls (`gst_tensor_meta_info_*`, `nnstreamer_decoder_find`) — that alters output would be caught here rather than only in a pipeline-level golden test.

**6. Will CI actually block a regression?**
Yes. `unittest_tensor_region` is a normal `meson test` target (`tests/meson.build`) with no special skip/xfail markers, so a `gtest` `EXPECT_EQ` failure fails the test binary and the CI test stage, which gates merge in this repo's normal flow.

**7. Architecture / API / docs**
No architecture or public API change — this is an internal bug fix in one decoder's private helper. The one relevant piece of documentation (the `option4` doc comment describing "Video input Dimension") was updated in the same commit to state the new drop/clamp semantics. I did not find any other in-repo document (`Documentation/*.md`) that describes `tensor_region` behavior and would need updating.

**Findings**
No Blocking or Major issues found.

- Nit: the new doc sentence ("A region is the part of a detected box inside this frame...") is placed under the `option4` bullet in the file header, but the drop/clamp behavior it describes is not really an `option4`-specific detail (it applies to every frame size, including the default 300:300). Purely cosmetic; no functional concern.

**Verdict: mergeable as is.**

The fix is well-targeted, the tests reproduce the original bug (verified: 13/15 new cases fail against the pre-fix decoder per the PR's own mutation table, which is consistent with the logic of the old code), and I independently confirmed both the new unit tests and the existing SSAT golden test pass on this branch after a real build.

myungjoo:

Second-round review, focused on the force-push from `2cb93f8c` to the current head `c1ff4f7f` and on test-gap hunting the first round may have missed. I did not just re-read the PR description — I fetched both commit objects directly from GitHub (`gh api repos/nnstreamer/nnstreamer/commits/<sha>`, then `git diff 2cb93f8c c1ff4f7f` after fetching both SHAs) and independently rebuilt and ran the test binary.

**0. Verifying the amend claim**

Confirmed. Both commits share the same parent (`f865ca74`), so this is a single-commit amend, not a rebase with new commits (`gh pr view --json commits` also shows exactly 1 commit on the PR). The full diff between the two heads is:

```
- *          A region is the part of a detected box inside this frame. A box with
- *          no pixel in the frame, or with a non-finite coordinate, is dropped.
+ *
+ * A region is the part of a detected box inside the option4 frame. A box with
+ * no pixel in the frame, or with a non-finite coordinate, is dropped.
```

That is the entire delta: the sentence is moved out from under the `option4:` bullet's indentation into its own paragraph, plus "this frame" → "the option4 frame" for clarity now that it's no longer visually scoped to that bullet. No code, test, or whitespace change anywhere else. This exactly matches the round-1 nit ("the sentence is placed under the option4 bullet... but the behavior it describes is not option4-specific") and resolves it correctly — the author's claim that this amend is documentation-position-only holds.

**1. Does it solve G20?**

Yes, unchanged from round 1's analysis (the fix logic itself did not change between rounds). `_clamp_to_frame()` computes each axis in float, rejects non-finite `start`/`size`, shortens the length when `start` is negative, cuts the length at the frame edge, bounds the frame at `G_MAXINT` before any float→int cast, and rejects a box with no whole pixel inside the frame — covering all four failure modes G20 lists (out-of-frame, wrong-pixels-on-left/top-clip, UB float→int, zero-width "whole frame" misread).

**2. Regression / cross-module risk**

Low. I independently rebuilt and ran `meson test -C build unittest_tensor_region -v` in the provided docker environment: 16/16 pass. I hand-traced several of the new test's expected values against the fixture's fixed box-prior/scale constants (`Y_SCALE=X_SCALE=10.0f`, `H_SCALE=W_SCALE=5.0f`, box priors all `0.5`) to confirm they aren't just empirically-fudged numbers — e.g. `regionLargeFrame`'s `1000000000` and `G_MAXINT - 1000000000` are exactly representable in `float` at every intermediate step (chosen as multiples of a power of two), so the test can't pass by accident from float rounding. `_clamp_to_frame` is the only change; `i_width`/`i_height` were already `guint` before this PR (no type change), and the byte-identical-for-in-frame-boxes claim holds by inspection (same truncating `(int)` cast on the same float product for the common case).

I also checked a subtler point: in the macro `if (!_clamp_to_frame(xmin,...) || !_clamp_to_frame(ymin,...)) break;`, short-circuit `||` means that when the x-axis clamp fails, the y-axis clamp is never called, leaving `result->y`/`height` at stale values from a previous loop iteration (the `object` struct is reused across the `for (d...)` loop in `_get_objects_mobilenet_ssd`). This looks alarming at first glance but is harmless: `result->valid` is unconditionally reset to `FALSE` at the top of every call, and the only guard at the call site is `if (object.valid == TRUE) g_array_append_val(...)`, so the stale x/y never reaches the output. No fix needed.

**3. Size / scope**

Right-sized, unchanged from round 1: one static helper, the call-site rewrite in the one file that owns the bug, plus tests. The amend itself is a 4-line/2-effective-line doc move, about as minimal as a nit-fix can be.

**4. Test completeness / gaps missed by round 1**

I specifically went hunting for boundary values and mutation-survivable paths this round, since that was flagged as a priority. Candidate gap I checked: no test puts `p` (the pixel start) exactly at `limit` (the frame edge), which would matter if `p >= limit` were mutated to `p > limit`. I traced this by hand: even with that mutation, execution falls through to `*len = (int) MIN((double) l, limit - *pos)`, and `limit - *pos` is `0` (or negative) whenever `p >= limit`, which forces `*len` to `0` or negative, so the final `return *len > 0` still rejects the box. This is a defended-in-depth equivalent mutant, not a real gap — the two early-return conditions (`p >= limit`, `l <= 0`) are provably redundant with the final length check for every input that doesn't risk an out-of-int-range cast; they exist to avoid computing `(int) p` on a wildly out-of-range but merely-large `p`, which is a real defense (confirmed by the PR's own mutation table: removing the whole check, not adjusting its boundary, is what `regionFarRightOfFrame_n` catches). So I don't think there's an actionable coverage gap here.

Otherwise the 15 cases still look thorough: in-frame passthrough, all four edges independently, all four "entirely outside" directions, sub-pixel drop, NaN, infinite size, empty option4, `option4` wider than `INT_MAX`, `option4` overriding the default, and a dropped high-score box not stealing an `option1` slot from a lower-score in-frame box. I did not find a meaningful hole beyond the equivalent-mutant case above.

**5/6. Future-regression detection and CI enforcement**

Unchanged from round 1: the fixture drives the decoder vtable directly (no model file dependency), asserts exact pixel values, and runs as an ordinary `meson test` GTest target with no skip/xfail markers, so a future regression here fails CI normally.

**7. Architecture / API / docs**

No architecture or API change. The amend is itself the requested documentation fix, and it is correctly resolved (see §0).

**Out-of-scope observation (not a finding against this diff, no action requested here)**

Running the new tests turned up a pre-existing, unrelated issue: `tr_exit()` (`tensordec-tensor_region.c:280-284`, not touched by this PR) unconditionally calls `g_array_free (trData->regions, TRUE)`, but `trData->regions` is set to `NULL` in `tr_init()` and is never assigned anywhere else in the file (confirmed by grep — it looks like dead/vestigial struct state from an old refactor). This produces a `GLib-CRITICAL **: g_array_free: assertion 'array' failed` on every decoder teardown. It's non-fatal (`g_array_free(NULL, ...)` just early-returns in current GLib) and the test file doesn't install a fatal-log handler, so it doesn't fail `meson test` — I saw all 16 tests reported `PASSED` alongside 15 of these CRITICALs (one per new `TEST_F`, since each now calls `decoder->init`/`exit` directly for the first time in this file's test history; the pre-existing pipeline-based test apparently never triggered it). This bug predates this PR and is outside `tensordec-tensor_region.c`'s G20-relevant code path, so it's not something to hold this PR on — flagging for a separate follow-up.

**Verdict: mergeable as is.**

The force-push is exactly what the author claims — a documentation-only reposition (plus a small wording tweak) that correctly resolves round 1's nit — and I found no new Blocking/Major/Minor issues against the diff itself in this round.

myungjoo:
Thanks — the NIT is taken, in `f172bd85` (the commit is amended, so the PR is still one commit).

`_get_object_i_mobilenet_ssd` now clamps into four locals and copies them into the object only after both axes succeed, so a rejected box leaves the object untouched and the helper no longer depends on the caller's `valid` discipline:

```c
int x, y, width, height;
if (!_clamp_to_frame (xmin, w, bb->i_width, &x, &width)
    || !_clamp_to_frame (ymin, h, bb->i_height, &y, &height))
  break;
```

No behaviour change, so the evidence is unchanged: `unittest_tensor_region` 16/16, clang-format 16 and `doxygen-tag.sh` clean, SSAT `nnstreamer_decoder_tensor_region` 4/4, and the single valgrind record in that binary is the pre-existing leak from a dlopened plugin's constructor in the older pipeline case (no repo frames, outside the gate's attribution).

*This comment was relayed on behalf of the author by a Claude agent.*

myungjoo:

Third-round review, focused on the force-push from `c1ff4f7f` (myungjoo-bot's APPROVE head) to the current head `f172bd85`, which the author describes as folding in myungjoo-bot's NIT (buffer the four clamped values in locals, write them to the object only after both axes succeed).

## 1. Is the amend NIT-only, with no behavior change?

Verified directly, not by trusting the description. Fetched both commit objects (`gh api repos/nnstreamer/nnstreamer/commits/c1ff4f7f...` and `.../f172bd85...`) and diffed them:

```
git diff c1ff4f7f f172bd85 --stat
 ext/nnstreamer/tensor_decoder/tensordec-tensor_region.c | 10 +++++++---
 1 file changed, 7 insertions(+), 3 deletions(-)
```

That is the only changed file; the test file (`tests/nnstreamer_decoder_tensor_region/unittest_tensor_region.cc`) is byte-identical between the two heads. The full hunk:

```c
-        if (!_clamp_to_frame (xmin, w, bb->i_width, &result->x, &result->width)   \
-            || !_clamp_to_frame (                                                 \
-                ymin, h, bb->i_height, &result->y, &result->height))              \
+        int x, y, width, height;                                                  \
+        if (!_clamp_to_frame (xmin, w, bb->i_width, &x, &width)                   \
+            || !_clamp_to_frame (ymin, h, bb->i_height, &y, &height))             \
           break;                                                                  \
         result->class_id = c;                                                     \
+        result->x = x;                                                            \
+        result->y = y;                                                            \
+        result->width = width;                                                    \
+        result->height = height;                                                  \
         result->score = score;                                                    \
```

This is exactly the NIT: `_clamp_to_frame()` now writes into four plain `int` locals, and `result->x/y/width/height` are only assigned after both `_clamp_to_frame()` calls have succeeded (the `break` on failure happens before any `result->` write). No other line in the file changed. I rebuilt the branch (docker `nns-g20`, `ninja -C build`, no errors) and ran `meson test -C build unittest_tensor_region -v`: 16/16 pass, matching the pre-amend result -- consistent with "no behavior change," since accepted boxes still go through the identical float math and are still written atomically once both axes pass.

## 2. Macro-hygiene / shadowing check on the new locals

The macro doc explicitly warns `@bug This is not macro-argument safe. Use parenthesis!`, so I checked whether the newly introduced local names (`x`, `y`, `width`, `height`) collide with any macro parameter or any identifier at the call sites:

- `_get_object_i_mobilenet_ssd(bb, index, total_labels, boxprior, boxinputptr, detinputptr, result)` -- parameter names, no match.
- Its only call site, inside `_get_objects_mobilenet_ssd`: `_get_object_i_mobilenet_ssd (bb, d, detbpi, boxprior, (boxinput_ + (d * boxbpi)), (detinput_ + (d * detbpi)), (&object))` -- actual argument tokens are `bb, d, detbpi, boxprior, boxinput_, boxbpi, detinput_, object`. None is `x`, `y`, `width`, or `height`.
- `_get_objects_mobilenet_ssd_(type, typename)` expands with `trData, type, typename, (trData->mobilenet_ssd.box_priors), (boxes->data), (detections->data), config, results` -- again no collision.
- The struct field names `x/y/width/height` on `detected_object` are accessed only via `result->x` etc., i.e. through the `->` operator, so they don't collide with the plain-identifier locals `x/y/width/height` at the token level.
- `_expit(x)` is a separate, unrelated single-argument macro (also defined with parameter name `x`) invoked as `_expit (detinputptr[c])` inside the same block, before the new locals are declared; since macro arguments are substituted textually into `_expit`'s own body, this does not create a variable named `x` and does not interact with the later `int x, y, width, height;` declaration.

So for the code as it stands today, there is no actual collision, and the whole PR builds and runs 16/16 clean.

**Minor (non-blocking) finding:** the fix nonetheless introduces a new latent hazard that did not exist in `c1ff4f7f`. Before the amend, the macro only ever wrote through `result->x`/`result->width`/etc. -- it never declared bare local variables, so there was nothing for a future call-site argument to collide with. Now the macro unconditionally declares `int x, y, width, height;` inside its `do { } while(0)` block. If a future second call site ever passed a variable named `x`, `y`, `width`, or `height` as any of the macro's arguments (most plausibly as `result`, e.g. a caller whose `detected_object *` happens to be named `x`), the `result->valid = FALSE;` line near the top (which runs before the new declarations) would still bind to the caller's `x`, but every `result->` reference after `int x, y, width, height;` (i.e. `result->x = x;` etc.) would instead resolve `result` to the shadowed local `int x` -- an invalid `int->member` expression that fails to compile. That's a fail-loud outcome (not silent corruption), and the single existing call site is safe today, so this is not blocking -- but a one-line comment on the macro, or renaming the new locals to something like `_x/_y/_w/_h`, would save a future contributor from rediscovering this, given the macro is already flagged as "not argument-safe."

## 3. Other usual review items (re-checked against the new head, not just carried over from earlier rounds)

- **Topic resolved:** Yes -- G20's four failure modes (past-edge, left/top-edge width bug, sub-pixel drop, float-to-int UB) are still fixed identically to `c1ff4f7f`; the amend changes only how the result is written, not what is computed.
- **Regression / cross-module impact:** None beyond what rounds 1-2 already established (the source diff outside this hunk is empty). `unittest_tensor_region` 16/16 locally after a clean rebuild.
- **Size:** trivial, 7/3 line diff, in scope.
- **Test completeness:** No new tests needed or added for this amend, correctly -- the observable behavior is unchanged, so the existing 16 cases (including `droppedBoxTakesNoRegion_n`, which already exercises the "second axis rejects" path) remain the right regression coverage. There isn't a black-box boundary case that would usefully test the buffer-then-commit ordering itself, since that's now a structural (compile-time-shaped) guarantee rather than a runtime one.
- **CI:** at the time of this review the checks on `f172bd85` are still QUEUED/IN_PROGRESS (this is a fresh force-push); `unittest_tensor_region` gates in the normal `meson test` flow as before, so a real regression would still fail CI. Worth confirming green before merge, though nothing in the diff suggests a new failure mode.
- **Docs:** no doc-relevant change in this hunk (the `option4` header comment discussed in earlier rounds is untouched between `c1ff4f7f` and `f172bd85`).

## Verdict

**Mergeable.** The amend does exactly what it claims -- takes the NIT (buffer-then-commit) with zero behavioral change, verified by direct commit diff and a clean local rebuild/test run (16/16). The new locals don't collide with anything in the current macro-parameter or call-site identifier set. The one thing worth a follow-up is the Minor macro-hygiene note above (a future second call site with a colliding variable name would fail to compile) -- non-blocking, and easy to leave for whoever adds a second call site, given the macro already carries a "not argument-safe" warning.

myungjoo:
Round 3's Minor is taken too, in `80088079` (amended again, still one commit).

The four locals the previous round introduced are now `_x`, `_y`, `_w`, `_h`, following `gst_tensor_crop_get_crop_info ()` in `gst/nnstreamer/elements/gsttensor_crop.c`, which uses exactly those names for the same purpose. That keeps the macro's own identifiers out of the namespace a call site is likely to pass in, so the hygiene concern about a future second caller no longer applies.

Rename only, no behaviour change: `unittest_tensor_region` 16/16, clang-format 16 and `doxygen-tag.sh` clean.

*This comment was relayed on behalf of the author by a Claude agent.*

myungjoo:

Fourth-round review, focused on the amend from `f172bd85` to the current head `80088079`, which claims to be a pure rename (`x`/`y`/`width`/`height` -> `_x`/`_y`/`_w`/`_h`) taking round 3's Minor. The topic itself (G20: clamping the tensor_region crop box to the frame) was already found mergeable in rounds 1-3 and by `myungjoo-bot`'s APPROVE; this round does not reopen that.

## Verifying the amend is rename-only

`git diff f172bd85 80088079` (both commits fetched locally; not GitHub's three-dot `compare` endpoint, which walks from the merge-base and would have shown the whole file) touches exactly one file, 7 insertions / 7 deletions, all inside the `_get_object_i_mobilenet_ssd` macro body:

```
-        int x, y, width, height;
-        if (!_clamp_to_frame (xmin, w, bb->i_width, &x, &width)
-            || !_clamp_to_frame (ymin, h, bb->i_height, &y, &height))
+        int _x, _y, _w, _h;
+        if (!_clamp_to_frame (xmin, w, bb->i_width, &_x, &_w)
+            || !_clamp_to_frame (ymin, h, bb->i_height, &_y, &_h))
           break;
         result->class_id = c;
-        result->x = x;
-        result->y = y;
-        result->width = width;
-        result->height = height;
+        result->x = _x;
+        result->y = _y;
+        result->width = _w;
+        result->height = _h;
```

No other line, no other file (including the test file) changed. Rebuilt at the current head in the shared `nns-g20` container and ran `meson test -C build unittest_tensor_region -v`: 16/16 pass, identical to the results already established in earlier rounds — confirms there is no behavior change. **Not a finding** — the author's description of the amend is accurate.

## Reserved-identifier check on `_x`/`_y`/`_w`/`_h`

C11 7.1.3p1 reserves a leading underscore followed by a lowercase letter *only* for identifiers with **file scope** in both the ordinary and tag namespaces; block-scope automatic variables are explicitly excluded from that restriction (only leading underscore + uppercase, or double underscore, are reserved everywhere). `_x`, `_y`, `_w`, `_h` here are declared with `int` inside the `do { ... for (...) { if (...) { ... } } } while (0)` block the macro expands into — i.e. ordinary block scope, not file scope — so this is standards-conformant, not UB and not a namespace violation. **Not a finding.**

The stated collision-avoidance rationale also checks out: `_get_object_i_mobilenet_ssd` currently has exactly one call site (`_get_objects_mobilenet_ssd`, line ~533), and none of that caller's own locals (`d`, `boxinput_`, `boxbpi`, `detinput_`, `detbpi`, `num`, `object`) collide with either naming — so there is no live bug today either way. The rename is a forward-looking hygiene improvement for hypothetical future callers that happen to use the very common names `x`/`y`/`width`/`height`, and it matches an existing repo precedent almost exactly: `gst/nnstreamer/elements/gsttensor_crop.c:670` already declares `guint i, j, ch, mw, mh, _x, _y, _w, _h;` as ordinary block-scope locals with this same convention. **Not a finding** — legitimate Minor-level cleanup, correctly resolved.

## Everything else

Topic resolution, regression coverage, cross-module impact, patch size, test completeness, CI-blocking power, and documentation were all already settled in rounds 1-3 and by `myungjoo-bot`'s APPROVE; this amend touches none of that (tests, docs and commit message are unchanged since round 3 other than the mechanical rename), so there is nothing new to re-litigate there.

## Verdict

No Blocking or Major findings in any round, and round 3's Minor is correctly and completely addressed by this amend with no side effects. **Mergeable as is.**

myungjoo:
On the remaining NIT (the four new names the macro occupies): **leaving it as it is, deliberately.**

The macro has to name the clamped values somewhere, so any spelling occupies exactly four identifiers — the count does not change with the choice. What the choice can change is how likely a call site is to pass one of them, and `_x`/`_y`/`_w`/`_h` were picked for that reason in the previous round: they follow `gst_tensor_crop_get_crop_info ()` in `gst/nnstreamer/elements/gsttensor_crop.c:670`, which declares `guint i, j, ch, mw, mh, _x, _y, _w, _h;` for the same quantities, and they are less likely to appear in an argument expression than the `x`/`y`/`width`/`height` they replaced. Note the two review rounds pulled in opposite directions here — round 3 asked for the underscore form precisely to reduce this collision surface — so this is as far as renaming can take it.

Removing the hazard rather than narrowing it would mean turning `_get_object_i_mobilenet_ssd` into a function, which is a rewrite of code this PR only needed to clamp, and outside a memory-safety fix. The macro's existing `@bug This is not macro-argument safe` already warns a future caller, and the single call site passes `bb`, `boxprior`, the two cursors and `&object`.

*This comment was relayed on behalf of the author by a Claude agent.*

PR #4993, opened 2026-09-17
[Element/Aggregator] Release the buffer on every push error (#4920 C10)

myungjoo:
item **C10** of #4920

- **What:** `tensor_aggregator` leaked the buffer on both `GST_FLOW_ERROR` returns of `gst_tensor_aggregator_push()`: a frame-size mismatch and a concatenation failure. Checking who owns the buffer turned up a second bug on the success path. With `frames-in == frames-out`, `concat=true` and a **shared** incoming buffer (a `tee`, or an app still holding a reference), the concatenated data went into a copy that leaked. The element then pushed the original buffer, unconcatenated, without holding a reference to it.
- **How:** `gst_tensor_aggregator_concat()` now takes the buffer and returns the one to push, or releases it and returns NULL. It used to call `gst_buffer_make_writable()` on its own copy of the pointer. `push()` now releases the buffer on both error returns. The element change is 11 lines added, 6 removed. The new `unittest_aggregator` binary is the first test coverage of the `frames-in == frames-out` path.
- **Result:** `concatSharedInput` fails on main (the output is the same pointer as the input) and passes with the fix. The four error-path `_n` cases are Tizen-only (see below). Against a local build with a non-fatal `ml_logf`, all 5 of those checks fail on main and pass with the fix. No regressions: the existing aggregator cases in `unittest_plugins` (13) and `unittest_sink` (5) pass, as does the `nnstreamer_converter` SSAT group.

---

### Details for reviewers / agents

**Where the error returns can be reached.** Both returns come right after `ml_logf`. On Linux distros that is `g_error`, which aborts, so the leak only shows on Tizen/Android, where `ml_logf` is a FATAL-level `dlog_print` that returns. So `invalidFrameSize_n` (the incoming buffer) and `invalidFrameSizeAdapter_n` (a frame taken from the adapter, detected with a weak reference on the input memory) are guarded by `#ifdef __TIZEN__` and run in the Tizen GBS `%check` (the `x86_64, unit_test 1` job). The abort itself is #4960 G5 territory and is left as it is.

**Why a plain `gst_buffer_unref (outbuf)` after a failed concat would have been wrong.** `concat()` called `gst_buffer_make_writable (outbuf)` on its local parameter. If the buffer was shared, that dropped the caller's reference, so the caller's pointer no longer carried one. The ownership change fixes the success-path bug and makes the error-path unref safe.

**Concat failure:** `concatMapFailure_n` and `concatSourceMapFailure_n` (also Tizen-only) use a test allocator whose memory refuses write maps, or all maps. That makes the destination map, or the source map, in `concat()` fail. Each case checks that the buffer's memory is freed when the push returns. Deleting either branch's new `gst_buffer_unref (outbuf)` makes only that branch's case fail.

**Local evidence** (WSL tree `~/nns-4920-c10`, gtest 1.14, `-Wall -Wextra -Werror`):

| element | tests | result |
|---|---|---|
| main | Linux (3) | `concatSharedInput` FAILED (`out == in`) |
| fix with concat always copying | Linux (3) | only `concatWritableInput` FAILED |
| fix | Linux (3) | 3/3 passed |
| main, `ml_logf` → `g_critical` | Tizen guard lifted (7) | 5 FAILED: `concatSharedInput`, `invalidFrameSize_n` (refcount 2), `invalidFrameSizeAdapter_n`, `concatMapFailure_n` and `concatSourceMapFailure_n` (not released) |
| fix, `ml_logf` → `g_critical` | Tizen guard lifted (7) | 7/7 passed |
| fix without the dest-map unref | Tizen guard lifted (7) | only `concatMapFailure_n` FAILED |
| fix without the source-map unref | Tizen guard lifted (7) | only `concatSourceMapFailure_n` FAILED |

Also run: `gst-indent` (clean), clang-format 16 (applied), `doxygen-tag.sh` with ctags plus a negative control.

**Files:** `gsttensor_aggregator.c` (fix), `tests/nnstreamer_aggregator/unittest_aggregator.cc` (new), `tests/meson.build` (one block after `unittest_if`). The aggregator's existing tests live in `unittest_plugins.cc`, which open PRs own, so the new cases are a separate binary. `packaging/run_unittests_binaries.sh ./tests` picks it up without changes.


myungjoo:
**Note:** This review was transcribed from an analysis performed by another Claude agent (not the one posting this comment), which examined the PR diff, the surrounding `gsttensor_aggregator.c` code, and the test infrastructure, and additionally rebuilt and ran the new test binary against both the pre-fix and post-fix code in a local WSL environment to independently verify the claims in the PR description.

## Summary of what was checked

- Does the change fix item C10 of #4920 (buffer leak/ownership bug in `tensor_aggregator`)?
- Does it avoid regressions in other modules?
- Is the change size proportionate and scoped to the stated topic?
- Is test coverage adequate, including regression coverage and forward-looking coverage?
- Would a future regression here be caught by CI?
- Are docs/architecture updates needed?

## Findings

**Correctness of the fix (verified, not just read) — informational, not a defect.**
Independently rebuilt `unittest_aggregator` against both the pre-fix (`main`) and post-fix code in a local WSL tree:
- Pre-fix: `concatSharedInput` fails with `out == in` (`0x...480 vs 0x...480`), matching the described bug where a shared buffer's concatenated data went into a leaked copy while the original, uncatenated buffer was pushed without an owned reference.
- Post-fix: all 3 non-Tizen cases pass, and with the Tizen guard lifted (`ml_logf` simulated as non-fatal), all 5 cases pass, including `invalidFrameSize_n` and `invalidFrameSizeAdapter_n`.
This confirms the fix addresses the stated topic and the regression test actually reproduces the bug on `main`.

**Ownership contract change is self-contained (no regression risk to other modules) — informational.**
`gst_tensor_aggregator_concat()` is `static` and has exactly one call site (`gst_tensor_aggregator_push()`), both in `gsttensor_aggregator.c`. The two call sites of `gst_tensor_aggregator_push()` (`gst_tensor_aggregator_chain()`, direct-passthrough and adapter-drained paths) both already pass a transfer-full `GstBuffer *`, consistent with the new "push() and concat() consume/release the buffer on every path" contract. No other file or module touches these functions, so the signature/ownership change cannot leak into other elements.

**Scope and size — no issues.** 3 files touched (11+/6- in the element, +10 in `tests/meson.build`, +250 new test file). All changes are within the aggregator element and its own new test binary; nothing unrelated is touched. This matches the stated topic well.

**Test completeness (Minor).** The PR description itself flags that the two `gst_buffer_map()` failure branches inside `gst_tensor_aggregator_concat()` (source-map and dest-map failure) are not covered by a test, because they need an allocator that refuses mapping and because `ml_logf` aborts before the refusal can be observed on non-Tizen. This is an honestly-disclosed gap rather than a hidden one, but it does mean a future refactor of the map-failure branches has no regression test to catch it. Given the difficulty of triggering it and that it's explicitly called out, this is a Minor/acceptable gap rather than a blocker — but consider filing (or already covered by) a follow-up issue for a fault-injecting allocator, similar to other memory-audit follow-ups in this codebase.

**Forward-looking regression coverage (Minor).** The new tests target the `frames-in == frames-out` path specifically (the path that had zero coverage before). This is good, but note it only guards the aggregator itself — if another module changes shared-buffer semantics upstream (e.g., a `tee` or an adapter change altering how/when buffers become shared), these new tests would catch it because they explicitly construct a shared buffer with `gst_buffer_ref()`, so this criterion is reasonably satisfied for the aggregator's own contract.

**CI enforcement of the regression test (Minor / worth confirming).** `invalidFrameSize_n` and `invalidFrameSizeAdapter_n` are compiled only under `#ifdef __TIZEN__` and thus do not run in the default Linux CI matrix (by design, since `ml_logf` aborts via `g_error` there before the refusal path can be exercised) — this is explained clearly in the PR body. However, this means the second, error-path half of the bug (the actual buffer leak on refusal, as opposed to the success-path double-buffer bug) is only continuously verified on Tizen GBS builds, not on the PRs that run through the standard GitHub Actions Linux CI. If Tizen GBS `%check` is not part of every PR's required CI gate, a future regression in the error-path unref could go undetected until a Tizen-specific build. Worth confirming that the Tizen GBS check job is indeed a required/blocking check for this repo, or otherwise noting the residual gap.

**Documentation.** No public API or architecture change — `gst_tensor_aggregator_concat()` is a static, internal function, and its only external contract (`tensor_aggregator`'s GObject properties/behavior) is unchanged. No documentation update is required, and none is needed in this PR.

**Style/process.** Doxygen comments present on all new functions/tests including `main()`; DCO sign-off and dual authorship attribution present on both commits; commit subject lines follow `[Component] Summary` convention. No `CHANGES` file edits (correct, per project policy).

## Overall verdict

**Mergeable as-is.** The fix is correct (independently reproduced and verified both ways), narrowly scoped, self-contained (no cross-module regression risk since the changed function is `static` with a single call site), and comes with a genuine regression test that fails on `main` and passes with the fix. The only items worth a maintainer's discretion, none of them blocking:
- Minor: `gst_buffer_map()` failure branches in `concat()` remain untested (acknowledged in the PR body).
- Minor: confirm whether the Tizen-only `_n` regression tests run in a required/blocking CI check, or whether the error-path fix is effectively unverified by continuous Linux CI.

myungjoo:
Response to the review above (transcribed from another Claude agent's review), head `87436367`:

1. **Minor: `gst_buffer_map` failures in `concat()` untested.** Added `concatMapFailure_n`. It uses a test allocator whose memory maps for reading but never for writing, so the destination map fails, and it checks that the buffer's memory is freed when the push returns. Locally, with a non-fatal `ml_logf`: fails on main (memory never released), passes with the fix. If only that branch's new `gst_buffer_unref (outbuf)` is deleted, this case alone fails. The source-map branch is the same shape and has no dedicated case.
2. **Minor: error-path tests are Tizen-only.** That is by construction: on Linux distros `ml_logf` is `g_error` and aborts before either return, so the leak cannot happen there. The Tizen job that runs these cases is part of this PR's checks: `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`, whose `%check` runs `run_unittests_binaries.sh ./tests` and so picks up `unittest_aggregator`. The Linux-reachable ownership bug is gated on every Linux job by `concatSharedInput`.

myungjoo:
**Note:** This is a follow-up review, transcribed from an analysis performed by another Claude agent (not the one posting this comment). It examined the current diff (head `87436367`), the two prior comments on this PR (the round-1 review and the author's response), the surrounding `gsttensor_aggregator.c` code, and the GStreamer `GstMemory`/`GstAllocator` contract (`gst/gstmemory.h`, `gst/gstallocator.c`, `gst/gstbuffer.c`), and independently rebuilt and ran `unittest_aggregator` against the pre-fix, fixed, and a targeted one-line mutation of the fixed code in a local WSL environment to verify the newly added `concatMapFailure_n` test rather than trust its description.

## Scope of this round

Round 1 raised two Minor items; this round checks how they were addressed, with focus on the new `concatMapFailure_n` test and its `AggrReadOnlyAllocator` test allocator added in response.

## Findings

**Round-1 item "gst_buffer_map() failure branches untested" — resolved, verified correct (informational).**
`concatMapFailure_n` adds a minimal `GstAllocator` subclass whose memory maps for `GST_MAP_READ` but refuses `GST_MAP_WRITE`, so `gst_tensor_aggregator_concat()`'s destination-map branch fails deterministically, and checks (via `gst_mini_object_weak_ref` on the `GstMemory`) that the buffer's memory is actually freed once `gst_harness_push()` returns. Independently rebuilt and ran all three relevant variants:
- Pre-fix (`main`, `ml_logf` simulated non-fatal): `concatMapFailure_n` fails (`released == false`) — the destination-map failure path leaks the buffer, as described.
- Fixed (head `87436367`): all 6 ownership tests pass, `concatMapFailure_n` included.
- Fixed code with only the new `gst_buffer_unref (outbuf)` in the dest-map-failure branch removed (a 1-line mutation): only `concatMapFailure_n` fails; the other 5 tests still pass.

This confirms the test is not a false-positive/always-pass test and that it is precisely the one that pins that specific line.

**GstAllocator/GstMemory contract check on the new test allocator — no defect found (informational).**
`aggr_read_only_allocator_free()` does not call `gst_memory_unref (mem->parent)` for a shared/child memory. This looks suspicious at first glance (the "textbook" custom-allocator pattern does that), but it is actually correct: GStreamer's generic `_gst_memory_free()` (`gst/gstmemory.c`) already does `if (mem->parent) gst_memory_unref (mem->parent);` *before* invoking `allocator_class->free()`, and also unrefs the allocator itself afterward. An allocator-supplied `free` doing it again would double-unref the parent. `mem_share`/`mem_copy`/`mem_map` are implemented consistently with the real default allocator's behavior (share creates a child pointing at the flattened root parent; copy `g_memdup`s the live range). `G_DEFINE_TYPE`, `gst_memory_init`, `gst_mini_object_weak_ref`, and the `mem_map`/`mem_unmap`/`mem_copy`/`mem_share`/`free` vtable have been stable GStreamer 1.x API since 1.0, so there's no reason to expect different behavior on Tizen's GStreamer/GLib versions.

**Style — no issues.** `clang-format --style=file -output-replacements-xml` against the repo's `.clang-format` reports 0 replacements for the new test file; it is fully compliant with the C++ style rule in CLAUDE.md.

**Minor/Nit (residual, not blocking).** Only the *destination*-map failure branch is exercised by a dedicated allocator/test. The *source*-map failure branch (`gst_buffer_map (srcbuf, &src_info, GST_MAP_READ)`) in `gst_tensor_aggregator_concat()` remains untested; it is structurally identical (log, unmap/unref what is already mapped, unref both buffers, return `NULL`), so the residual risk is low, but it is technically still an asymmetry versus the round-1 ask ("the two `gst_buffer_map()` failure branches"). Optional: a variant allocator whose `mem_map` refuses `GST_MAP_READ` (or fails only on the read map) would close this out completely, but given the code shape is identical to the now-covered branch, this is a Nit rather than something that should block merge.

**Round-1 item "Tizen-only CI enforcement" — confirmed correct (informational).** Verified directly in this checkout: `packaging/nnstreamer.spec`'s `%if 0%{?unit_test}` block runs `bash %{test_script} ./tests` via `packaging/run_unittests_binaries.sh`, which recursively finds and runs every executable named `unittest_*` under the build's `tests` tree — this includes `unittest_aggregator`. `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` is one of this PR's own required checks, so the Tizen-gated `invalidFrameSize_n` / `invalidFrameSizeAdapter_n` / `concatMapFailure_n` cases are exercised on every push to this PR, not just on ad-hoc Tizen builds. The author's response to round 1 is accurate.

**Everything else from round 1 stands.** Scope is unchanged (only `gsttensor_aggregator.c`, `tests/meson.build`, and the new test file), the fix is `static`/single-call-site and self-contained, and no docs/architecture updates are needed (no public API change).

## Overall verdict

**Mergeable as-is.** Both round-1 Minor items have been adequately addressed: the map-failure gap now has a genuine, independently-verified regression test (fails on `main`, passes on the fix, and fails again under a 1-line targeted mutation of the fix), and the Tizen-only CI enforcement claim has been confirmed against the actual packaging spec and this PR's required checks. The only remaining item is a non-blocking Nit: the symmetric source-map-failure branch still has no dedicated test.

myungjoo:
Response to the second agent review above, head `3a3c4adc`:

- **Nit: the source-map failure branch has no dedicated test.** Added `concatSourceMapFailure_n`. The test allocator now has a `refuse_read` switch, and both map-failure cases share one helper. Locally, with a non-fatal `ml_logf`: the case fails on main and passes with the fix. If only the source-map branch's new `gst_buffer_unref (outbuf)` is deleted, only this case fails (the destination-map mutation likewise fails only `concatMapFailure_n`). clang-format 16 and `doxygen-tag.sh` (with ctags) are clean.

myungjoo:
**Note:** This is a third-round review, transcribed from an analysis performed by another Claude agent (not the one posting this comment). It examined the current diff (head `3a3c4adc`), the two prior review rounds and the author's responses on this PR, the surrounding `gsttensor_aggregator.c` code and its two call sites, issue #4920's item C10 text, and this PR's own CI check list — and independently rebuilt and ran `unittest_aggregator` against `main`, the current fixed code, and targeted one-line mutations of the fixed code (dropping each of the three new `gst_buffer_unref (outbuf)` calls one at a time) in a local WSL environment, rather than trusting the prior rounds' description of these results.

## Scope of this round

Focus is the newest change on top of round 2 (the `refuse_read` switch on the test allocator, the shared `_aggr_test_concat_map_failure()` helper, and the new `concatSourceMapFailure_n` case), plus a fresh end-to-end check of the whole PR against the review checklist, since two rounds of prior discussion were about earlier heads.

## Findings

**Topic match (verified, informational).** Issue #4920 C10 reads: *"`gst_tensor_aggregator_push()` leaks `outbuf` on both error returns ... Fix: unref before return."* The diff does exactly that, plus the necessary ownership-plumbing change to `gst_tensor_aggregator_concat()` (now returns the buffer to push, or releases it and returns `NULL`) that a plain unref could not have done safely, as already explained in the PR body. This matches the stated topic precisely.

**Correctness, independently re-verified end-to-end (informational).** Rebuilt and ran all 7 cases three ways:
- `main` (Tizen guard lifted, `ml_logf` simulated non-fatal): 5/7 fail — `concatSharedInput` (`out == in`), `invalidFrameSize_n` (refcount 2, not released), `invalidFrameSizeAdapter_n`, `concatMapFailure_n`, `concatSourceMapFailure_n` (all "released == false"). Matches the PR's table exactly.
- current head, fixed: 7/7 pass (Tizen-guard-lifted) and 3/3 pass (plain, Tizen guard active — the default Linux CI shape).
- current head, fixed, with only the new source-map-failure branch's `gst_buffer_unref (outbuf)` removed: only `concatSourceMapFailure_n` fails, the other 6 pass. Repeated for the dest-map branch: only `concatMapFailure_n` fails. This confirms the newest test genuinely pins the newest line, is not a false-positive/always-pass test, and is independent of the dest-map case it was cloned from.

**No regression risk to other modules (verified, informational).** `gst_tensor_aggregator_concat()` is `static`, one call site (`gst_tensor_aggregator_push()`, also `static`, two call sites, both in `gsttensor_aggregator.c`). Read both call sites (`gst_tensor_aggregator_chain()`'s direct-passthrough `frames_in == frames_out` branch, and the adapter-drained loop): both already hand `push()` a transfer-full `GstBuffer *` and never touch it again afterward, so the new "callee releases the buffer on every path" contract does not create a double-free or a leak at either site. No other file references these two functions.

**Scope and size — no issues.** Only `gsttensor_aggregator.c`, `tests/meson.build`, and the new `tests/nnstreamer_aggregator/unittest_aggregator.cc` are touched; the `tests/meson.build` addition is a verbatim copy of the sibling `unittest_if` block. No `CHANGES` edit, correct per project policy.

**Test completeness — confirmed thorough, no gap remaining.** All four release points now have a dedicated, mutation-verified test (`invalidFrameSize_n`, `invalidFrameSizeAdapter_n`, `concatMapFailure_n`, `concatSourceMapFailure_n`), on top of the three non-Tizen cases that cover the success-path double-buffer bug and the ordinary/no-concat paths. The round-1 and round-2 Minor/Nit items (map-failure branches untested, then only the dest-map branch covered) are both closed out by this head.

**Style — no issues.** `clang-format --style=file` reports 0 replacements for the new test file. `tools/development/gst-indent` on `gsttensor_aggregator.c` produces no diff. Every new function (including the `AggrReadOnlyAllocator` vtable functions and `TEST` cases) carries a Doxygen `@brief`. Commit subjects follow `[Component] Summary`; both commits carry a single DCO sign-off plus co-author trailer; commit bodies match the code and the PR description accurately.

**Documentation — none needed, confirmed.** No public API or architecture change; `gst_tensor_aggregator_concat()` stays internal/static.

**CI status on this exact head — action item, not a code defect (Minor).** `gh pr checks 4993` at the time of this review shows every check, including `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")`, as **pending** for head `3a3c4adc` (the push that added `concatSourceMapFailure_n`). That Tizen job is confirmed to be one of this PR's own checks and is the *only* job that runs `unittest_aggregator`'s `_n` cases (Linux CI can't reach them, by design, since `ml_logf` is `g_error` there and aborts first). Recommend waiting for this specific check to go green on the current head before merge — the Linux-reachable half of the bug (`concatSharedInput`) is already covered by every Linux job, but the error-path unref fix has, as of this review, not yet been confirmed passing CI on this head.

## Overall verdict

**Mergeable**, pending the current head's CI actually going green (in particular the Tizen GBS `x86_64, unit_test 1` job, which is where the new `_n` regression tests run). The fix is correct and independently re-verified against `main`, matches issue #4920 item C10 precisely, is self-contained with no cross-module regression risk, is proportionate in size, and now has complete, mutation-tested regression coverage for every release point it touches. No blocking or Major issues found in this round.

myungjoo:
Response to the automated review above, head `c9b51bd5`:

- **NIT: `concatWritableInput` did not check that the concatenation happens in place.** Fixed. The test now puts a weak reference on the input buffer and, after pulling, asserts that the buffer was not released and that the pushed buffer is the same pointer. Taking a reference instead would have turned the buffer into a shared one and changed the case being tested. Verified with a mutation that makes `concat()` copy unconditionally: only `concatWritableInput` fails on it (both new assertions), while `main` still fails only `concatSharedInput` and the fix passes 3/3 on Linux and 7/7 with the Tizen guard lifted. clang-format 16 and `doxygen-tag.sh` are clean.

myungjoo:
**Note:** This is a fourth-round review, transcribed from an analysis performed by another Claude agent (not the one posting this comment). It examined the current diff at head `c9b51bd5` (`gh pr diff 4993`), the full `gsttensor_aggregator.c` source (both call sites of `gst_tensor_aggregator_push()`), issue #4920 item C10, the three prior review rounds and the author's responses on this PR, and independently rebuilt and ran `unittest_aggregator` against the current fixed code and against `main` in a local WSL environment (not just re-reading the PR's own numbers).

## Scope of this round

Focus is the newest change on top of round 3: `concatWritableInput` now weak-refs the input buffer and asserts both that it was not released and that the pushed buffer is the same pointer (`_aggr_released`, moved out of `#ifdef __TIZEN__`). Also did a fresh full pass against the review checklist since this is the first round to look at the *whole* diff independently rather than only the newest hunk.

## Findings

**Topic match and correctness — re-verified independently, no defect.** Rebuilt `unittest_aggregator` from the current head:
- `fixed`, Tizen guard active (plain, 3 cases): 3/3 pass.
- `fixed`, Tizen guard lifted (`ml_logf` simulated non-fatal, 7 cases): 7/7 pass, including `concatWritableInput`'s new assertions (`EXPECT_FALSE(released)`, `EXPECT_EQ(out, in)`).
- `main`, Tizen guard lifted: 5/7 fail (`concatSharedInput`, `invalidFrameSize_n`, `invalidFrameSizeAdapter_n`, `concatMapFailure_n`, `concatSourceMapFailure_n`), 2/7 pass (`concatWritableInput`, `noConcatSharedInput`) — matches the PR's own table exactly.

This confirms the round-3 NIT was actually fixed (not just claimed) and that the regression suite still reproduces the original bug on `main` end-to-end.

**Ownership contract, re-checked at the source.** `gst_tensor_aggregator_concat()` is `static`, single call site (`gst_tensor_aggregator_push()`, also `static`). Both of `push()`'s callers in `gst_tensor_aggregator_chain()` — the direct `frames_in == frames_out` passthrough (`gst_tensor_aggregator_chain:862`) and the adapter-drained loop (`:922`, after `gst_buffer_make_writable()`) — already hand it a transfer-full `GstBuffer *` and never touch it again. No other file references either function, so the new "callee releases the buffer on every exit path" contract cannot leak into another element. No regression risk to other modules confirmed by direct code reading, consistent with rounds 1–3.

**Scope and size — no issues.** Only `gsttensor_aggregator.c`, `tests/meson.build` (a verbatim copy of the `unittest_if` block), and the new `tests/nnstreamer_aggregator/unittest_aggregator.cc` are touched. Proportionate to the stated topic (C10 of #4920).

**Test completeness — thorough.** All four release points (`invalidFrameSize_n`, `invalidFrameSizeAdapter_n`, `concatMapFailure_n`, `concatSourceMapFailure_n`) plus the three non-Tizen ownership cases are covered and were previously mutation-verified (rounds 1–3, each new `gst_buffer_unref` shown to be pinned by exactly one test via targeted one-line removal). Nothing new to add here.

**CI enforcement — confirmed by direct inspection.** `packaging/run_unittests_binaries.sh` recursively runs every `unittest_*` binary under the build tree, so `unittest_aggregator` — including the `__TIZEN__`-only `_n` cases — runs under the `Tizen GBS build on Ubuntu (x86_64, --define "unit_test 1")` check, which is one of this PR's own required checks. `gh pr checks 4993` shows all checks, including that one, as **pending** on head `c9b51bd5` at the time of this review (just pushed); none had failed. Recommend confirming this exact head goes green (particularly that Tizen job) before merge, since it is the only job that exercises the `_n` error-path cases.

**Documentation.** No public API/architecture change — `gst_tensor_aggregator_concat()` remains internal/static. No documentation update needed or missing.

**New Nit (not raised in any prior round): unused include.** `tests/nnstreamer_aggregator/unittest_aggregator.cc:17` includes `<unittest_util.h>`, but none of its declarations (`setPipelineStateSync`, `count_output`, `argmax_float`, `replace_string`, `get_available_port`, `wait_pipeline_process_buffers`, etc.) are used anywhere in the file. This is harmless — the header's two function-like symbols used elsewhere in the test suite are `static inline`, so an unused include does not trigger `-Wunused-function` even under `-Werror` — purely a housekeeping/style nit with no functional or CI impact.

## Overall verdict

**Mergeable**, pending this exact head's CI (in particular the Tizen GBS `x86_64, unit_test 1` job, the only job that runs the `_n` error-path regression cases) actually finishing green — no result was available yet at review time. No Blocking or Major issues found. The single new item is a non-blocking Nit (unused `#include <unittest_util.h>` in the new test file); everything else confirms the findings of the three prior review rounds, independently re-verified against the code and a local rebuild rather than taken on trust.

myungjoo:
Response to the review above, head `0b27fdd6`:

- **Nit: `unittest_util.h` is included but unused.** Removed. Nothing in the file referenced it (the `-DFAKEDLOG=1` define and the test-util library come from `nnstreamer_unittest_deps` in `tests/meson.build`, not from the include), and the binary builds and passes without it: 3/3 on Linux and 7/7 with the Tizen guard lifted, clang-format 16 and `doxygen-tag.sh` clean.

myungjoo:
**Note:** This is a fifth-round review, transcribed from an analysis performed by another Claude agent (not the one posting this comment). It reviewed the full current diff at head `0b27fdd6` (`gh pr diff 4993`), the four prior review rounds and the author's responses on this PR, issue #4920 item C10, and — since the only change since round 4 (head `c9b51bd5`) is dropping the `#include <unittest_util.h>` line that round 4's Nit flagged as unused — specifically traced how `FAKEDLOG`/Tizen's fake-dlog declarations reach this translation unit without that include, rather than trusting the author's response that it is safe. It also rebuilt and ran `unittest_aggregator` from the current head in a local WSL environment.

## Scope of this round

Round 4 raised one Nit (`unittest_util.h` included but unused); this round checks whether removing it is actually safe in every build configuration, and re-confirms the whole PR against the review checklist now that a full round has looked at this exact head.

## Findings

**Nit resolution, verified safe including on Tizen (informational, no defect).** `unittest_aggregator.cc` never called any symbol declared in `tests/unittest_util.h` (no `setPipelineStateSync`, `count_output`, `replace_string`, etc.), so the include was indeed dead. The concern is the `FAKEDLOG` fake-`dlog_print`/`log_priority` declarations that `__TIZEN__` builds need: tracing the include chain shows this is a non-issue, because `gst/nnstreamer/nnstreamer_log.h` itself does `#include "../../tests/unittest_util.h"` whenever `__TIZEN__` and `FAKEDLOG` are both defined (nnstreamer_log.h:30-38), independent of whatever includes it. `unittest_aggregator.cc` includes `<tensor_common.h>`, which includes `"nnstreamer_log.h"` (tensor_common.h:39), so the fake-dlog declarations still reach this file transitively on a Tizen+FAKEDLOG build even with the explicit include gone. Locally rebuilt and ran the current head (Tizen guard lifted, non-fatal `ml_logf`): 7/7 pass; Tizen guard active: 3/3 pass — no compile or link issue from the removal.

**Commit messages and PR body still match the code (informational).** The two-commit split, the "11 lines added, 6 removed" element diff, the file list, and the results table in the PR body all still match `gh pr diff 4993` and the local re-run of the tests at this head.

**Repeat of the review checklist against the current head (informational, no new issues).** Topic match, ownership-contract correctness, scope/size, and process items (Doxygen, DCO, dual authorship, commit style, no `CHANGES` edit) were already verified correct in rounds 1-4 and are unchanged by this round's one-line diff. No architecture or public API changed (`gst_tensor_aggregator_concat()` remains `static` with its one call site), so no documentation update is needed.

**CI status at review time.** All GitHub Actions checks (including the Tizen GBS `x86_64, unit_test 1` job that actually runs the four `_n` cases) were re-queued and showing `pending` for head `0b27fdd6` at the time of this review, since the include removal just landed. Worth confirming they land green before merge, though nothing in the analysis above suggests they will not.

## Overall verdict

**Mergeable.** This round found no Blocking, Major, or Minor issues — only confirmed that the prior round's Nit was resolved correctly, including on the Tizen build configuration it could plausibly have affected. Combined with the four prior rounds' independent verification of the fix itself (leak/ownership bug reproduced on `main`, fixed and regression-tested at HEAD, self-contained since the changed function is `static` with a single call site), this PR looks ready to merge once CI turns green on this head.

myungjoo:
Approved by Astra

PR #4994, opened 2026-09-17
[Filter/Single] Pass the private data itself to a V1 getModelInfo (#4920 B12)

myungjoo:
item **B12** of #4920

- **What:** `g_tensor_filter_single_set_input_info()` passed `&priv->privateData` (`void **`) to a V1 sub-plugin's `getModelInfo`, which expects the private data itself (`void *`). C++ sub-plugins got `-EINVAL` from the wrapper's sanity check, so setting input info through the single API never worked for them; a plain-C V1 sub-plugin would read/write through a wrong pointer.
- **How:** one-line fix: pass `priv->privateData`, like every other V1 call in `tensor_filter_common.c`. The V0 branch (`setInputDimension`, which does take `void **`) is unchanged.
- **Result:** 4 new cases in `unittest_filter_single` (a plain-C V1 sub-plugin that checks the pointer it receives, a C++ sub-plugin, and 2 negative cases). Without the fix 3 of them fail; with it 6/6 pass locally. clang-format 16 and doxygen-tag check pass. No other module passes `&privateData` to a V1 callback (checked all `privateData` uses in `gst/` and `ext/`).

---

### Details (for reviewers and agents)

**Why it matters today.** `tensor_filter_subplugin::cpp_getModelInfo()` runs `get_tfsp_with_checks (private_data)`, which checks a sanity value. Given `&priv->privateData`, that value sits inside `GstTensorFilterPrivate`, the check fails, and the call returns `-EINVAL`. So `ml_single_set_input_info()` failed for every C++ V1 sub-plugin (onnxruntime, lua, python3, ...). The existing tests use tflite, which is V0, so they never reached this branch.

**Tests** (`tests/nnstreamer_filter_single/unittest_filter_single.cc`, outside the tflite `#ifdef`s so they always build):
| case | without fix | with fix |
|---|---|---|
| `NNSFilterSingleTestPrivateData.setInputInfo` | FAIL (`-22`, wrong pointer) | OK |
| `NNSFilterSingleTestPrivateData.setInputInfoRefused_n` (sub-plugin refuses a float32 input; pointer still checked) | FAIL (wrong pointer) | OK |
| `NNSFilterSingleTestPrivateData.setInputInfoNotOpened_n` (not opened: `-EINVAL`, sub-plugin not called) | OK | OK |
| `testTensorFilterSingle.setInputInfoCppSubplugin` | FAIL | OK |

The C++ mock uses `nnstreamer_cppplugin_api_filter.hh` without a guard, the same way `unittest_plugins.cc` does. `enable-filter-cpp-class` defaults to true, and no build config in the tree turns it off.

Local verification: WSL tree `~/nns-4920-b12`, configured without tflite, so the existing tflite cases were not run locally and are left to CI.


myungjoo:
This is a code review written by another Claude agent (a separate session from the one that authored this PR), relayed here on request of the repository maintainer. The reviewing agent independently re-read the diff, the referenced call sites, and the issue text, and built/ran the new tests locally before writing this.

**Verdict: mergeable as-is.** No blocking issues found.

## Summary of what was checked

- **Correctness of the fix**: `g_tensor_filter_single_set_input_info()` (tensor_filter_single.c:498) now passes `priv->privateData` (not `&priv->privateData`) to the V1 `getModelInfo` callback. This matches the `getModelInfo` prototype in `nnstreamer_plugin_api_filter.h` (`void *private_data`), and matches every other V1 call site in the tree: the `gst_tensor_filter_v1_call` macro in `tensor_filter_common.h:68` already does `(priv)->fw->funcname((priv)->fw, &(priv)->prop, (priv)->privateData, ...)` for the main `tensor_filter` element. The V0 branch just above (`setInputDimension`, which does take `void **`) is correctly left untouched. `git grep privateData` across `gst/` confirms no other call site has this bug.
- **Root-cause story checks out**: `tensor_filter_subplugin::cpp_getModelInfo()` (tensor_filter_support_cc.cc:238) does `GET_TFSP_WITH_CHECKS(obj, private_data)`, which validates a sanity marker at a fixed offset from `private_data`. Given the address of a field inside `GstTensorFilterPrivate` instead of the subplugin's own object, that check fails with `-EINVAL` — consistent with the PR's claim that this broke `ml_single_set_input_info()` for every C++ V1 sub-plugin.
- **Regression risk**: very low. The change is a single line, confined to the "single" element's V1 `set_input_info` path; the `tensor_filter` (pipeline) element already used the correct macro, so this only brings the single-element code path in line with the already-correct main path. Nothing else in the module is touched. Since the previous behavior was actively broken (returned `-EINVAL`, or handed a plain-C sub-plugin a wrong pointer), there is no plausible caller relying on the old behavior.
- **Test coverage**: 4 new GTest cases in `unittest_filter_single.cc`, covering both positive and negative paths as the project's test policy requires: a plain-C V1 sub-plugin that records/verifies the pointer it receives (positive), the same sub-plugin refusing an unsupported dtype (`_n`, checks the pointer is still right even on refusal), calling `set_input_info` before the framework is opened (`_n`), and a C++ V1 sub-plugin exercising the actual failure mode described above. I built the PR's branch in a local tree (`ninja -C build tests/unittest_filter_single`) and ran `unittest_filter_single --gtest_filter=*PrivateData*:*setInputInfoCppSubplugin*`: all 4 pass. These tests exercise exactly the affected function, so a future regression in this call site (or in the V1 dispatch macro pattern it should match) will be caught by CI automatically — `unittest_filter_single` is already registered unconditionally in `tests/meson.build` and runs under `meson test`.
- **Scope**: 2 files changed (1 source line + tests), nothing outside `tensor_filter_single.c` and its test file. No other module touched.
- **Docs/API**: no public API or architecture change — the `getModelInfo` signature was already correct in the header; only the call site was wrong. No documentation update is needed or expected, and none is present in `Documentation/`, correctly.

## Minor / informational (not blocking, not requesting changes)

- **[Informational]** The new test unconditionally `#include`s `nnstreamer_cppplugin_api_filter.hh` and defines a C++ V1 sub-plugin, without any `enable-filter-cpp-class` guard. If that meson option were ever set to `false`, `tensor_filter_support_cc.cc` (which defines the `tensor_filter_subplugin` static wrapper methods) would not be compiled, and this would fail to link. However, this is **pre-existing project convention, not something introduced by this PR**: `tests/nnstreamer_plugins/unittest_plugins.cc` already does the exact same thing unguarded, and `unittest_filter_single` (like `unittest_plugins`) is built unconditionally in `tests/meson.build`. The option defaults to `true` and nothing in the tree's packaging/CI configs disables it. Not asking for a change here — just noting it for awareness, consistent with what the PR description already states.

No other issues found in the diff. The fix is minimal, correctly scoped, verified against the header contract and every other V1 call site, and backed by tests that would have caught the regression and will catch any reintroduction of it in CI.

myungjoo:
Approved by Astra

PR #4995, opened 2026-09-17
[Filter/C++] Keep per-element state out of a shared cpp filter (#4920 B10, B11)

myungjoo:
item **B10**, **B11** of #4920

- **What:** a registered C++ (`framework=cpp`) filter is one object shared by every tensor_filter that opens it, but the sub-plugin kept per-element state in shared places. B10: every `open()` wrote `isAllocatedBeforeInvoke()` into the process-wide framework descriptor, so mixing a preallocating and a self-allocating cpp filter gave one of them the wrong output contract (NULL output → SIGSEGV, or a leaked buffer and unwritten output pushed downstream). B11: `open()` stored the element's `prop` pointer in the shared object and `close()` never cleared it, so derived classes read the last-opened element's properties, and freed memory once that element was destroyed.
- **How:** B10 declares `allocate_in_invoke = TRUE` once and answers per instance through the existing `allocateInInvoke (private_data)` callback. B11 turns the protected `prop` member into a `static thread_local` that each dispatch (`getInputDim`/`getOutputDim`/`setInputDim`/`invoke`) sets to its caller's properties and restores afterwards; derived code still reads `prop` unchanged.
- **Result:** 7 new cases in `unittest_cppfilter` (16/16 pass, 10× repeat stable, valgrind gate 0 errors/0 definite leaks from this repo). Against main's sources 6 of them fail (4 FAILED, a SIGSEGV in B10's pipeline case, an abort in B11's destroyed-peer case; the 7th is a guard); mutations "drop `thread_local`" and "drop the allocateInInvoke callback" are each caught. Only `tensor_filter_cpp.{cc,hh}` and the cpp filter's own test change.

**Compatibility note:** removing the `prop` data member changes the layout of `tensor_filter_cpp` (installed as `nnstreamer/tensor_filter_cpp.hh`). Filters are source compatible but must be rebuilt; the header already says the class definition is not stable.

What `prop` means changes with it: it used to hold the last opening element's properties in any method, and it is now `nullptr` outside the four callbacks — in `isAllocatedBeforeInvoke ()` (whose answer the header already requires to be fixed at construction time) and on threads a callback starts of its own. The header documents both. No in-tree filter reads `prop` elsewhere; an out-of-tree one that does would have been reading another element's properties before this change, and freed memory after that element was destroyed, which is exactly B11.

<details><summary>Details for reviewers / agents</summary>

Commits (one per item):
1. `[Filter/C++] Tell the allocation contract per opened filter` (B10) — `.allocate_in_invoke = TRUE`, new `tensor_filter_cpp::allocateInInvoke()` returning 0 iff `!isAllocatedBeforeInvoke()`, global write removed. The framework (`gst_tensor_filter_allocate_in_invoke()`, used by tensor_filter and tensor_filter_single) already calls `allocateInInvoke (&priv->privateData)` when the flag is TRUE; no other code reads `fw->allocate_in_invoke`.
2. `[Filter/C++] Give a cpp filter the properties of the element calling it` (B11).

Tests (`tests/cpp_methods/unittest_cpp_methods.cc`, suite `cppFilterShared`):

| case | checks | on main's sources |
|---|---|---|
| allocContractPerFilter | framework decision for pre/self/pre opened in that order | FAILED |
| allocContractFailedOpen_n | failed open leaves the contract alone | pass (guard) |
| allocContractPipeline | tee → preallocating + self-allocating filter, byte compare | SIGSEGV |
| propFollowsCaller | open ×2, close 2nd, all 4 callbacks of 1st see its prop | FAILED |
| propOutsideCallback_n | `prop` is nullptr outside callbacks | FAILED |
| propPerThread | invoke A reads prop while invoke B (other thread) is inside its callback | FAILED; SIGSEGV with `thread_local` removed |
| propAfterPeerDestroyed | appsrc pipeline + 2nd pipeline on same model destroyed, then stream through 1st; compares seen prop with the element's `priv.prop` | abort (g_assert through dangling prop) |

Local env: docker `nns-b10b11` (image `nns-g16:deps`), trees `~/nns-4920-b10b11/{fix,b10,base}`, `-Dwerror=true -Dcpp_args="-Wextra -Wsign-compare"`. `propPerThread` initially failed 4/60 because both invokes wrote the shared `seen` after release — fixed by recording each side under the lock; 0/60 afterwards. clang-format 16 clean, doxygen-tag.sh clean (negative control exits 1).
</details>


myungjoo:
*This is an automated review by a separate Claude agent, relayed to this PR by the PR author's session.*

## Scope check

PR touches only `ext/nnstreamer/tensor_filter/tensor_filter_cpp.{cc,hh}` and `tests/cpp_methods/unittest_cpp_methods.cc`, i.e. exactly the cpp sub-plugin named in items B10/B11 of #4920 plus its own test file. No other module is touched, so there is no unnecessary cross-module footprint.

## Does it solve B10 and B11?

**B10** (process-wide `allocate_in_invoke` flag clobbered by every `open()`): the fix declares `.allocate_in_invoke = TRUE` once in the static `NNS_support_cpp` descriptor and wires `.allocateInInvoke = tensor_filter_cpp::allocateInInvoke`, which returns `cpp->isAllocatedBeforeInvoke() ? -EINVAL : 0` for the specific opened instance. I traced this against `gst_tensor_filter_allocate_in_invoke()` (`tensor_filter_common.c:784-805`): `fw->allocateInInvoke` returning 0 ("supported") yields `allocate_in_invoke = TRUE` (self-alloc), non-zero yields `FALSE` (framework preallocates) — this is exactly the inverse mapping the old code computed manually with `!cpp->isAllocatedBeforeInvoke()`, so the per-instance answer is preserved, just no longer written into shared state. The global write `cpp->prop = prop` inside `open()` and the `NNS_support_cpp.v0.allocate_in_invoke = ...` line are both removed. This correctly fixes B10.

**B11** (dangling shared `prop` pointer): `prop` becomes `static thread_local`, set around each of the four dispatch functions (`getInputDim`, `getOutputDim`, `setInputDim`, `invoke`) with a save/restore of the previous value, and left `nullptr` outside those calls per the new header comment. I confirmed via `tensor_filter_common.h`'s `gst_tensor_filter_v0_call`/`GST_TF_FW_INVOKE_COMPAT` macros that every callback is always invoked with `&priv->prop` — a pointer stable for the lifetime of the owning `GstTensorFilter` element — so keying the thread-local on "whichever prop pointer is currently being dispatched" correctly isolates concurrent/overlapping callers instead of leaking the last opener's freed properties. This correctly fixes B11.

Both fixes are consistent with the framework contract in `nnstreamer_plugin_api_filter.h` and don't require changes to `tensor_filter.c`/`tensor_filter_single.c`/`tensor_filter_common.c` — the existing `allocateInInvoke` plumbing already supported a per-instance answer; the bug was purely in how the cpp sub-plugin used it.

Low. `NNS_support_cpp` is a file-local static struct specific to the `cpp` framework; no other sub-plugin or core file reads it. `gst_tensor_filter_allocate_in_invoke()` and the calling macros are unchanged. `tensor_filter_cpp.hh` is an installed header (`nnstreamer/tensor_filter_cpp.hh`), but the PR body correctly calls out that removing the `prop` instance member changes the class layout — the header already documents "The API and Class definition is NOT stable," so out-of-tree cpp filters need a rebuild, not a source change (derived classes still just read the protected `prop` member as before).

## Test coverage

7 new cases in a new `cppFilterShared` suite, directly reproducing both bugs (mixed pre/self-allocating filters in one pipeline for B10; two elements sharing one filter object, with the second destroyed, for B11), plus a same-process negative/guard case (`allocContractFailedOpen_n`) and a cross-thread case (`propPerThread`) that exercises the thread-local specifically (a plain shared-static implementation would fail this one). Per the PR body these fail against pre-fix sources (4 FAILED, one SIGSEGV, one abort) and pass after the fix, which is the regression-test bar this repo's CLAUDE.md asks for. `allocContractPerFilter` and `_allocates_in_invoke()`/`GstTensorFilterPrivate` exercise `gst_tensor_filter_allocate_in_invoke()` itself (not just the cpp glue), so a future change to that shared helper that breaks the per-instance contract would also be caught here — this gives some of the "detect regressions from unrelated future changes" coverage the review checklist asks for, though the coverage is naturally limited to the cpp framework's usage of that helper, not the helper's other callers (tensor_filter/tensor_filter_single), which already have their own tests.

The new test file reaches into private headers (`../../gst/nnstreamer/tensor_filter/tensor_filter.h`, `GstTensorFilterPrivate`) and calls `nnstreamer_filter_find()`/`gst_tensor_filter_allocate_in_invoke()` directly. This is not a new pattern in this codebase — `tests/nnstreamer_sink/unittest_sink.cc` and several other filter unit tests already do the same — so it should link and build the same way those do; `unittest_cppfilter` is already registered in `tests/cpp_methods/meson.build` and runs under `meson test`, so CI will pick these up without any test-registration changes needed.

Every new function/struct field carries a Doxygen comment, satisfying the repo's Doxygen-per-function CI check as far as I can see from the diff.

## Minor observations (non-blocking)

- **Low — exception safety of the save/restore pattern.** `getInputDim`/`getOutputDim`/`setInputDim`/`invoke` each do `outer = prop; prop = <new>; ret = cpp->foo(...); prop = outer;` by hand. If a derived class's callback ever threw an uncaught C++ exception, the final restore would be skipped and the thread-local would stay pinned to the inner value until the next dispatch on that thread overwrites it. Nothing else in this call path uses exceptions today (GStreamer/C-style error codes throughout), so this is unlikely to bite in practice, but a small RAII guard (set in the constructor, restore in the destructor) would make the four call sites both shorter and unwind-safe. Suggest as a follow-up, not a blocker.
- **Nit — code duplication.** The same 4-line save/set/call/restore sequence is repeated in all four dispatch functions; could be factored into one helper (which would also address the exception-safety point above). Purely stylistic.
- **Nit — early-return cleanup in a couple of new GTest cases.** E.g. `propAfterPeerDestroyed` calls `probe._register()` before `ASSERT_NE (pipeline1, nullptr)`; if that assertion ever fired, `probe._unregister()` below it would be skipped, leaking the "b11_pl" registration for the rest of the process. Every model name used in the new tests is unique, so this can't cross-contaminate other cases today, and the same pattern already exists in the pre-existing tests in this file — just flagging for consistency, not asking for a change.

## Documentation

No `Documentation/*` file mentions `tensor_filter_cpp`, `isAllocatedBeforeInvoke`, or `allocate_in_invoke`, so there is nothing stale to update there. The class-layout compatibility note is already carried in the PR body and (implicitly) in the header's existing "not stable" disclaimer; I don't think a separate doc change is required for this internal, non-public-API-surface fix. (Per repo policy, no `CHANGES` entry is expected or requested here.)

## Verdict

**Mergeable as is.** The fix directly and correctly addresses both B10 and B11 as scoped, stays entirely within the cpp sub-plugin and its own tests, doesn't touch shared framework code, and ships regression tests that fail on `main` and pass with the fix. The observations above are style/robustness suggestions, not blockers.

myungjoo:

Thanks — all three points from the relayed review are addressed in-PR (folded into the B11 commit, new head 41701a48):

- **Low (exception safety) + Nit (duplication):** the four hand-written save/set/call/restore sequences are replaced by a private nested RAII guard `tensor_filter_cpp::prop_scope` (declared in the header's private section, defined in the .cc); each dispatch is now `loadClass (...); prop_scope scope (prop); return cpp->...;`.
- **Nit (early ASSERT in `propAfterPeerDestroyed`):** the probe is now registered after `ASSERT_NE (pipeline1, nullptr)`, so a failed assertion cannot leave a registered pointer to a destroyed stack object behind.

Re-verified locally: 16/16 (each new case ×5, `propPerThread` 0/60 failures), 6 of the 7 new cases still fail on main's sources, the drop-`thread_local` and drop-`allocateInInvoke` mutations are still caught, clang-format 16 and doxygen-tag clean, valgrind gate 0 errors / 0 definite leaks from this repository.

myungjoo:
*This is an automated review by a separate Claude agent, relayed to this PR.*

## Round 2 — re-review after the RAII/ASSERT-order rework

This is a follow-up pass over the full current diff (head `41701a48`), focused on verifying the two changes made in response to the round-1 relayed review, plus a fresh look at the whole PR.

### 1. RAII `prop_scope` replacing the manual save/restore

`ext/nnstreamer/tensor_filter/tensor_filter_cpp.cc` now defines a private nested class `tensor_filter_cpp::prop_scope` (constructor captures `outer = tensor_filter_cpp::prop` and sets the new value; destructor restores `outer`), and each of `getInputDim`/`getOutputDim`/`setInputDim`/`invoke` does `prop_scope scope (prop); return cpp->...;` instead of the earlier four hand-written save/set/call/restore blocks.

- Verified the nested class is forward-declared in the header's `private:` section (`class prop_scope;`) and defined out-of-line in the `.cc`, which is required for a nested class defined outside the enclosing class body — this compiles.
- Verified a nested class has implicit access to its enclosing class's `protected`/`private` members regardless of the nested class's own access specifier, so `prop_scope` can read/write the `protected static thread_local` member `tensor_filter_cpp::prop` even though `prop_scope` itself is declared `private`. Correct.
- The destructor unconditionally restores `outer`, including on stack unwind from an exception, which is exactly the exception-safety gap flagged in round 1. Confirmed fixed.
- Traced all four dispatch sites (`getInputDim`, `getOutputDim`, `setInputDim`, `invoke`) — all four now use the guard consistently; no leftover hand-written save/restore anywhere in the file.
- Nesting/re-entrancy: since `prop_scope` just pushes/pops a thread-local value, nested calls on the same thread (if a derived filter's callback ever re-entered the framework) would still restore correctly — the save happens at construction time of each scope, independent of how many scopes are already active on that thread.
- Traced the assumption the whole B11 fix rests on: every V0 dispatch macro (`gst_tensor_filter_v0_call` in `tensor_filter_common.h:60`, `GST_TF_FW_INVOKE_COMPAT` in `tensor_filter_common.h:75`, and the direct call in `tensor_filter_single.c:492`) always passes `&priv->prop`, a member embedded in the owning `GstTensorFilter`/`GTensorFilterSingle` instance that is stable for the element's lifetime. So keying the thread-local scope on "whichever `prop` pointer is currently being dispatched" is sound, and this is unchanged from round 1's analysis.

No defect found in this part of the rework.

### 2. Probe registration moved after `ASSERT_NE` in `propAfterPeerDestroyed`

Confirmed in the current diff: `pipeline1` is constructed, then `ASSERT_NE (pipeline1, nullptr)`, and only then `EXPECT_EQ (probe._register (), 0)`. This matches the round-1 nit exactly — a failed `gst_parse_launch` now aborts the test (via `ASSERT_NE`, which returns from the test body) before the "b11_pl" model name is registered, so there is no longer a way to leak a registered pointer to a destroyed stack object across test cases. Correct fix, and consistent with how the rest of this file's pre-existing tests are structured.

### 3. Fresh look at the whole diff (not just the delta)

- Scope is unchanged and still minimal: only `ext/nnstreamer/tensor_filter/tensor_filter_cpp.{cc,hh}` and `tests/cpp_methods/unittest_cpp_methods.cc`. No other module touched.
- Re-verified the B10 polarity by reading `gst_tensor_filter_allocate_in_invoke()` (`tensor_filter_common.c:784-805`): `allocateInInvoke(&priv->privateData)` returning 0 keeps `allocate_in_invoke = TRUE` (self-allocating), non-zero flips it to `FALSE` (framework preallocates). `tensor_filter_cpp::allocateInInvoke()` returns `cpp->isAllocatedBeforeInvoke() ? -EINVAL : 0` — same inverse mapping the removed manual code computed, now answered per opened model (`private_data`) instead of written into the shared, process-wide `NNS_support_cpp` descriptor. Still correct.
- Checked that `gst_tensor_filter_allocate_in_invoke()` (hence `tensor_filter_cpp::allocateInInvoke`) is called on every output-buffer preparation (`tensor_filter.c:925`, inside `_gst_tensor_filter_transform_get_output_data`), not just once at open — `isAllocatedBeforeInvoke()` is a cheap, side-effect-free query and is documented as constant for the lifetime of the model, so the repeated call is harmless and matches the pre-existing framework contract.
- Checked `close()`: it no longer needs to touch `prop` at all now that no shared storage of a caller pointer exists outside the duration of a single dispatch — this fully removes the UAF surface B11 described rather than papering over it with a clear-on-close.
- Confirmed no `Documentation/*` file references `tensor_filter_cpp`, `isAllocatedBeforeInvoke`, or `allocate_in_invoke`, so (as in round 1) there is no stale doc to update; the header's own "API and Class definition is NOT stable" note and the PR body's compatibility callout (class layout change requires rebuild of out-of-tree `.hh` consumers) already cover the only externally-visible effect.
- Confirmed `tests/cpp_methods/meson.build` already lists `unittest_cpp_methods.cc` and `tests/meson.build` pulls in `cpp_methods` whenever `enable-cppfilter` (default `true`) is set, so the new `cppFilterShared` suite runs under `meson test` / CI without any build-file changes — a regression here will fail CI, not just a local run.
- No new issues found beyond what round 1 already surfaced and what has now been addressed.

## Verdict

**Mergeable as is.** Both requested changes (RAII `prop_scope`, later probe registration in `propAfterPeerDestroyed`) are implemented correctly and close the two non-blocking points from round 1 without introducing new defects. The fix still correctly and minimally addresses B10 and B11, stays confined to the cpp sub-plugin and its own tests, and ships regression tests wired into the existing CI test suite.

myungjoo:

CI note on head 41701a48: `Ubuntu pdebuild on resolute (amd64)` (run 35193118250, job 105109971927) sat in `run pdebuild` for ~110 min, so it was cancelled to read its log and re-run. It was not stuck in this PR's code:

- The log's last line before the cancel is `[Starting] nnstreamer_decoder_boundingbox` at 07:27:34Z (the SSAT sweep, `cd tests && ssat -n -p=1` in `debian/rules`), and the orphan cleanup killed a `gst-launch-1.0` from that group. Nothing in `tests/nnstreamer_decoder_boundingbox/` uses `framework=cpp`.
- The pdebuild recipe does not run `unittest_cppfilter` (it only builds it); the new `cppFilterShared` cases ran and passed in the Clang/ASan job, both Valgrind jobs (gate: 0 error contexts / 0 definite leaks from this repository) and Tizen GBS x86_64 `unit_test 1`.
- The jammy leg of the same run passed, and the same resolute job passed on the other #4920 PRs based on the same main.

Treating it as a runner-side hang in the bounding-box SSAT group; the job has been re-run.

myungjoo:

The NIT is addressed in-PR (folded into the B11 commit, new head `8b04956f`; no behaviour change).

Opening a scope inside `allocateInInvoke ()` is not possible: its framework signature is `int (*allocateInInvoke) (void **private_data)`, so no properties reach it — and the header already requires `isAllocatedBeforeInvoke ()` to be decided at construction time, so an answer that depends on `prop` was never well defined. The semantic change is documented instead, in three places:

- `tensor_filter_cpp.hh`, on `prop`: it is nullptr in any other method, **including `isAllocatedBeforeInvoke`**, and on threads one of the four callbacks starts of its own; read it only there and copy what you pass on.
- `tensor_filter_cpp.hh`, on `isAllocatedBeforeInvoke ()`: `prop` is not available here and the answer may not depend on it.
- The commit message and the PR body's compatibility paragraph.

Re-verified after the doc change: 16/16 in `unittest_cppfilter` (each new case ×3), `meson test` OK, doxygen-tag clean with a failing negative control, clang-format 16 clean, valgrind gate 0 error contexts / 0 definite leaks from this repository.

myungjoo:
*This is an automated review by a separate Claude agent, relayed to this PR by the PR author's session.*

## Round 3 — re-review of the whole diff after the documentation-only follow-up

Reviewed head `8b04956f` end to end (`gh pr diff 4995`), plus the full history of prior review rounds (two relayed agent reviews, one `myungjoo-bot` review, all "mergeable as is" / "no blocking issues"). This round focuses on: (a) verifying the doc-only delta since the bot review actually matches the code, (b) re-deriving correctness of the whole diff independently rather than trusting the prior rounds' conclusions, and (c) checking whether `origin/main`'s movement since this branch's base requires a rebase.

### 1. What changed since the bot review

`git diff 41701a48..8b04956f` touches only `ext/nnstreamer/tensor_filter/tensor_filter_cpp.hh` comments (on the `prop` member and on `isAllocatedBeforeInvoke ()`), the B11 commit message, and the PR body. No `.cc`/test code changed. Confirmed by direct diff, matching the PR author's description.

### 2. Does the new documentation match the code?

- **`allocateInInvoke ()` really cannot receive properties.** Verified against the framework's own definition in `gst/nnstreamer/include/nnstreamer_plugin_api_filter.h:384`: `int (*allocateInInvoke) (void **private_data);` — no `GstTensorFilterProperties *` parameter at all. So the claim "opening a scope inside `allocateInInvoke ()` is not possible" is accurate, not just plausible; there is nothing to scope with. `tensor_filter_cpp::allocateInInvoke()` (`tensor_filter_cpp.cc:235-239`) indeed calls `cpp->isAllocatedBeforeInvoke ()` with no `prop_scope`, so `prop` is whatever the calling thread's thread-local happened to be (verified below that this is always `nullptr` in practice, since `gst_tensor_filter_allocate_in_invoke()` is called from `_gst_tensor_filter_transform_get_output_data()` in `tensor_filter.c:925`, outside of and before any of the four `prop_scope`-guarded dispatches on that thread). The header's new comment on `prop` ("nullptr everywhere else... including `isAllocatedBeforeInvoke`") and on `isAllocatedBeforeInvoke ()` ("`prop` is not available here; the answer may not depend on it") both match the code exactly.
- **B10 polarity**, re-derived from `gst_tensor_filter_allocate_in_invoke()` (`tensor_filter_common.c:784-805`): with the static `.allocate_in_invoke` field now unconditionally `TRUE` and `.allocateInInvoke` always non-null, every V0 call now goes through the per-instance callback — `allocateInInvoke() == 0` ⇒ `allocate_in_invoke = TRUE` (self-alloc), else `FALSE` (framework preallocates). `tensor_filter_cpp::allocateInInvoke()` returns `cpp->isAllocatedBeforeInvoke() ? -EINVAL : 0`, the same inverse mapping the removed global-write code computed, now correctly scoped to `private_data` (i.e., the specific opened model) instead of the shared `NNS_support_cpp` descriptor. Confirmed both `tensor_filter.c:925` (`GstTensorFilter`) and `tensor_filter_single.c:319` (`GTensorFilterSingle`, into its own per-instance `spriv->allocate_in_invoke`) route through this same function — no code path still reads/writes the removed shared field.
- **B11 mechanism**: `&priv->prop` is what every V0 dispatch macro passes (`gst_tensor_filter_v0_call`, `GST_TF_FW_INVOKE_COMPAT`, both in `tensor_filter_common.h:56-79`) — a member embedded in the caller's own `GstTensorFilterPrivate`, stable for that element's lifetime and distinct per element even when they share one `tensor_filter_cpp` instance. Keying the `thread_local` scope on this pointer for the callback's duration and restoring afterward (`prop_scope`, `tensor_filter_cpp.cc:103-121`) removes the shared/dangling-pointer surface entirely, rather than just nulling it on `close()`. The `propAfterPeerDestroyed` test's assertion `probe.seen.load() == &GST_TENSOR_FILTER_CAST(filter)->priv.prop` is checking exactly this pointer identity, and `GstTensorFilterPrivate::prop` (`tensor_filter_common.h:159`) is a plain struct member (not a pointer), so the address comparison is well-founded.
- Confirmed no other in-tree class derives from `tensor_filter_cpp` besides the test fixtures in `tests/cpp_methods/`; `cppfilter_test.cc`'s pre-existing `filter_basic::invoke()` reads the inherited `prop` member directly (unqualified), and since `invoke()` only ever runs inside the `prop_scope`-guarded `tensor_filter_cpp::invoke()` dispatcher, this pre-existing, unmodified test code keeps working correctly under the new thread-local scheme — no regression there.

No mismatch found between the new comments and the actual code; the NIT from round 1 is fully and accurately addressed.

### 3. Rebase check against `origin/main`

`git log f865ca74..origin/main` shows 11 commits landed since this branch's base, including other #4920 items (bounding-box decoder, python3 decoder, converter free-fixes, a `[Filter/Single]` fix). Checked each for overlap with this PR's files (`ext/nnstreamer/tensor_filter/tensor_filter_cpp.{cc,hh}`, `tests/cpp_methods/unittest_cpp_methods.cc`):
- `git log`/`git diff` restricted to those three paths between the base and `origin/main` returns nothing — no commit since the base touches them.
- The one commit that looks adjacent, `12b6149c [Filter/Single] Pass the private data itself to a V1 getModelInfo`, only changes `tensor_filter_single.c`'s V1 `getModelInfo` call (`priv->fw->getModelInfo(...)`), a completely different framework path (V1) from the one `tensor_filter_cpp` uses (V0, via `NNS_support_cpp`). No interaction.
- `gh pr view 4995 --json mergeable` reports `MERGEABLE` (the `BLOCKED` merge-state is from branch-protection/review requirements, not conflicts).

**No rebase is needed.** The PR can merge on its current base without functional or textual conflict.

### 4. Fresh check of the whole diff, scope and tests

- Scope stays exactly `ext/nnstreamer/tensor_filter/tensor_filter_cpp.{cc,hh}` + the cpp filter's own test file, matching B10/B11 and nothing else.
- `tests/cpp_methods/meson.build` (unchanged) already builds `unittest_cpp_methods.cc` into `unittest_cppfilter` and registers it with `test()`; `tests/meson.build:427-428` pulls in `cpp_methods` whenever `enable-cppfilter` (default `true`, `meson_options.txt:54`) is set — so the new `cppFilterShared` suite runs under plain `meson test` with no build-file changes, and a future regression in this area fails CI rather than being silently missed.
- No `Documentation/*` file mentions `tensor_filter_cpp`, `isAllocatedBeforeInvoke`, or `allocate_in_invoke` (`grep` over `Documentation/` returns nothing) — consistent with round 1/2's finding that there is no stale doc, and the API-stability/layout-change note already lives in the PR body and the header's pre-existing "not stable" disclaimer.
- Every new function (including the nested `prop_scope`'s constructor/destructor and every new `TEST`) carries a `/** @brief ... */` comment, satisfying the repo's Doxygen-per-function CI check.
- Re-confirmed test coverage is a genuine regression suite (not just new-feature coverage): the PR body's mutation results (dropping `thread_local`, dropping the `allocateInInvoke` wiring) are each caught, and the suite fails against pre-fix sources for 6 of 7 cases (one is a same-process negative guard). This satisfies both "test what's being fixed" and "would flag a future regression in this exact contract."

No new defects found in this round. All prior findings (exception safety, duplication, early-ASSERT ordering, the `prop`-outside-callback NIT) were addressed in earlier commits on this same PR and verified above to still hold under the current head.

## Verdict

**Mergeable as is.** The round-3 documentation-only change accurately reflects the code (verified independently against the V0 framework's `allocateInInvoke` signature and call sites, not just by reading the comment), no rebase against `origin/main` is required (no file or functional overlap with the 11 commits that landed since this branch's base), and a fresh pass over the complete diff found no new correctness, scope, or test-coverage issues beyond what the three earlier review rounds already surfaced and this PR already resolved.

myungjoo:
Approved by Astra

PR #4996, opened 2026-09-17
[Decoder/Python3] Refuse missing scripts, non-bytes results and unsupported tensors (#4920 F10)

myungjoo:
item **F10** of #4920

- **What:** the python3 tensor_decoder sub-plugin crashes or corrupts memory on three kinds of bad input: (a) no script instance — `mode=python3` without `option1`, an `option1` that fails to load, or a script class whose constructor raises (SIGSEGV at caps negotiation); (b) a script that returns something other than `bytes` from `decode`/`getOutCaps`, or more bytes than a caller-sized output buffer can hold (heap overflow); (c) a tensor type numpy is not given for (float16), which left NULL slots in the lists handed to the script (SIGSEGV).
- **How:** three commits, one per defect. (a) `getOutCaps`/`decode` refuse a NULL instance, `exit` treats it as a no-op, and the shared `loadScript()` fails when the instance is not created (this also fixes the python3 converter's identical crash). (b) `PyBytes_Check` on both results, a caller buffer grown only within its maximum size, a mapped-size check before the copy, and no framerate on NULL caps. (c) fail the decode when the numpy array or TensorShape cannot be created. No locks were added around python calls.
- **Result:** 12 new cases (9 negative, 3 positive). All 9 negative cases fail on main (4 by SIGSEGV, the caller-buffer case by writing past the buffer) and pass with the fix; the 3 positive cases pass on both. Each commit builds and passes on its own. Mutation matrix 14/16 caught (the array and shape checks mask each other for float16; removing both crashes). `meson test` for decoder/converter/filter python3 binaries, SSAT `nnstreamer_decoder_python3` + `nnstreamer_converter_python3`, strict `-Werror -Wextra -Wsign-compare`, clang-format 16 and doxygen-tag are clean locally.

### Behaviour changes
- A `tensor_decoder mode=python3` with no or a broken script now fails negotiation (`not-negotiated`) instead of crashing.
- A python3 decoder/converter script whose class raises in its constructor is now refused at `option1`/open.
- `getOutCaps` returning non-bytes now refuses caps (it produced NULL caps with GStreamer criticals before). A failing `getOutCaps` call still falls back to `application/octet-stream`, as before.
- float16 input is refused with an error; it is still not supported.

### Commits
1. `[Decoder/Python3] Refuse to decode without a script instance` — (a), helper `loadScript()`, tests incl. `tensorConverterPython.raisingConstructor_n` and three pipeline cases.
2. `[Decoder/Python3] Check what the script returns before copying it` — (b).
3. `[Decoder/Python3] Refuse a tensor that cannot be given to the script` — (c).

<details><summary>Notes for agents</summary>

- Scope/ownership: only `tensordec-python3.cc`, `nnstreamer_python3_helper.cc` (owned by this item in the #4920 parallel batch), tests in `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc`, one `tensorConverterPython` case in `unittest_converter.cc`, fixture modes in `custom_decoder_raw.py` (`init-raise`, `text`), new fixture `raise_init_custom_converter.py`. `gsttensor_decoder.c` (#4986) untouched.
- Reachability of the (b) overflow: tensor_decoder has no `getTransformSize` for python3, so basetransform hands `decode()` a 0-byte buffer and the element always takes the allocate path; the overflow needs a caller that passes a pre-sized buffer. The test drives the sub-plugin API directly with a wrapped buffer inside a guard array.
- Failed-assertion counting: `ml_loge` is `g_critical` on Linux, so the tests count only messages containing `assertion '` in the default, `GLib` and `GStreamer` domains.
- A caller buffer is grown only if `maxsize >= offset + size` (`gst_buffer_resize_range`'s own condition), so a too-small one is refused without an assertion failure; the tests count them and cover the offset term.
- Mutation matrix (fix tree, per test binary run): M1 getOutCaps NULL check -> noScript_n + both pipeline _n SIGSEGV; M2 decode NULL check -> noScript_n SIGSEGV; M3 exit `g_return_if_fail` -> noScript_n; M4 loadScript instance check -> decoder + converter constructor cases; M5 decode PyBytes_Check -> decodeNotBytes_n abort; M6 size check removed -> decodeIntoSmallBuffer_n; M7 `<` to `<=` -> 7 positives; M8 getOutCaps PyBytes_Check and M9 `if (caps)` -> getOutCapsNotBytes_n; M10/M11 individually not caught (both objects fail for the same unsupported type), M12 both -> decodeFloat16_n SIGSEGV; M13 resize ignoring offset and M14 resize check removed -> decodeIntoSmallBuffer_n; M15 never resize -> decodeIntoCallerBuffer.
- CI coverage: the Valgrind job does not build python3; `unittest_decoder_python3` runs in Pdebuild and GBS, RISC-V skips it via `-k`.
- Local: WSL `~/nns-f10/{fix,base,main}`, numpy 1.21.5 (`PYTHONNOUSERSITE=1`), flatbuffers 23.5.26 from `~/nns-py3f/deps`.
</details>


myungjoo:

**Scope of review:** the diff of this PR only (`ext/nnstreamer/extra/nnstreamer_python3_helper.cc`, `ext/nnstreamer/tensor_decoder/tensordec-python3.cc`, `tests/nnstreamer_converter/unittest_converter.cc`, `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc`, `tests/test_models/models/custom_decoder_raw.py`, `tests/test_models/models/raise_init_custom_converter.py`), read against the rest of the repository (`gsttensor_decoder.c`, `tensor_converter_python3.cc`, `tensor_filter_python3.cc`, `tests/meson.build`, `.clang-format`) and against Python/C-API and GStreamer semantics (`Py_SAFEDECREF`, `gst_buffer_set_size`/`gst_memory` ownership rules).

## Does it address issue #4920 item F10?

Yes, all three sub-defects listed in F10 are closed with a matching mechanism and a reproducing test that fails on `main` and passes with the fix:

- **(a) NULL private-data dereference** — `decoder_py_getOutCaps`/`decoder_py_decode` now refuse a NULL `core` (`tensordec-python3.cc:388-391`, `:411-414`), `decoder_py_exit` is a no-op on NULL (`:318-319`), and the shared `loadScript()` now fails (`-3`) when `PyObject_CallObject` on the script's class returns NULL — i.e. when the class's constructor raises (`nnstreamer_python3_helper.cc:321-324`). I traced this last path carefully: before the fix, a raising constructor made `loadScript()` return `0` (success) while leaving `core_obj == NULL` inside a **fully-constructed** `PYDecoderCore`/`PYConverterCore`, so `*pdata`/`*priv_data` ended up non-NULL and every later `getOutCaps`/`decode`/`convert` call dereferenced a NULL `core_obj`. After the fix, `init()` on both sub-plugins returns non-zero, the half-built core is `delete`d, and `*pdata`/`*priv_data` is left NULL — confirmed by re-reading `decoder_py_setOption` (`:356-371`) and `py_open` in `tensor_converter_python3.cc` (`:296-326`), both unmodified by this PR but relying on the corrected `loadScript()` contract.
- **(b) unbounded copy / no type check** — `PyBytes_Check` is added on `decode()`'s and `getOutCaps()`'s results before touching them as bytes (`:176-181`, `:253-258`), and the output mapping's size is checked against the script's byte count before the `memcpy` (`:202-209`), with a proper `gst_memory_unmap`/`gst_memory_unref` on the refusal path (no leak). `PyBytes_Size`/`PyBytes_AsString` were correctly swapped for the unchecked `_GET_SIZE`/`_AS_STRING` macros only where the preceding `PyBytes_Check` already guarantees the type, so the swap does not reintroduce the bug it fixes.
- **(c) NULL slots for unsupported types (float16)** — both `PyArray_SimpleNewFromData` and `PyTensorShape_New` results are checked before `PyList_SetItem` (`:149-154`, `:158-163`); a NULL at either point clearly refuses the tensor instead of shipping a NULL to the script.

## Regressions / other modules

- `loadScript()` is shared between the decoder and the python3 converter; I checked both call sites (`tensordec-python3.cc:291`, `tensor_converter_python3.cc:253`) and both already treat any non-zero return as failure, so the new `-3` case is handled correctly by existing code — no behavior change for the already-working paths (module import failure `-1`, missing class `-2` are untouched).
- `tensor_filter_python3.cc`'s own `PYCore::loadScript()` is a distinct member function (not the free function touched here), so the filter sub-plugin is unaffected, as the PR notes claim.
- `getOutCaps()` returning NULL now composes correctly with the already-merged C22 fix in `gsttensor_decoder.c` (NULL-caps guard in `fixate_caps`/`configure`/`set_caps`), which this PR does not touch. The new `pipelineNoScript_n`/`pipelineInvalidScript_n`/`pipelineScript` GTest cases exercise that exact interaction end-to-end through `gst_parse_launch`, which is good regression coverage for the seam between this sub-plugin and the element that isn't owned by this PR.
- No other file outside the stated scope is touched; `gsttensor_decoder.c` (owned by #4986) is untouched as claimed.

## Size / focus

Proportionate: 5 lines in the shared helper, 51/-4 in the decoder sub-plugin, and the rest is tests/fixtures for 3 well-isolated sub-defects. The one `unittest_converter.cc` addition is justified because the root-cause fix (`loadScript()`) is shared code — leaving the converter's identical crash untested would be the bigger scope gap.

## Test completeness / future-regression detection

Strong. 10 new negative GTest cases plus 3 positive ones, an assertion-counting helper to prove the fix path produces no GLib/GStreamer criticals (not just "doesn't crash"), a caller-supplied-buffer test that checks bytes *past* the buffer are untouched (not just the return code), and full-pipeline tests that would catch a regression introduced by a future change to `gsttensor_decoder.c`'s negotiation handling. I confirmed locally that `tests/meson.build:348-355` already wires `unittest_decoder_python3` into `meson test` whenever `have_python3 && flatbuf_support_is_available`, so these new cases do run in CI package builds (Pdebuild/GBS), not just locally.

## CI gating

Per the PR's own notes (not independently re-verified in this sandbox, no builds were run here): the Valgrind job doesn't build python3, so this class of bug wouldn't have been caught there anyway; `unittest_decoder_python3` runs under `meson test` in the package-build jobs, so a future regression on this exact code path fails CI there.

## Architecture / API / docs

No public header, ABI, or documented API changed (`loadScript()`'s signature and the `GstTensorDecoderDef` callback signatures are unchanged; only a new internal negative return value and internal NULL-guards were added). No doc update is needed for this fix.

## Style

`clang-format --style=file` (repo `.clang-format`, clang-format 16) is clean on all four changed C/C++ files — no diff. Doxygen `/** @brief */` comments are present on every new function, including the new static test helpers.

## Findings

- **[Low] `tensordec-python3.cc:188-191`** — When the caller hands `decode()` a pre-sized `outbuf` that is smaller than the script's output and whose backing memory's `maxsize` is *also* smaller than the required size, `gst_buffer_set_size()` still gets called to attempt the grow. `gst_buffer_set_size()` cannot exceed the memory's `maxsize`, so this trips a `gst_buffer_resize_range` GLib/GStreamer critical on every such call, even though the subsequent `out_info.size < mem_size` check (added by this PR) safely refuses the copy afterward — no memory corruption results, just log noise. This is pre-existing behavior (the `gst_buffer_set_size` call itself isn't new), and the PR's own commit message already documents it ("A caller buffer too small still logs `gst_buffer_resize_range` assertion... unchanged pre-existing call"), and `decodeIntoSmallBuffer_n` deliberately doesn't assert zero-criticals for this reason. Suggestion (optional follow-up, not a blocker for this PR): compare `mem_size` against the memory's `maxsize` before calling `gst_buffer_set_size`, so an undersized caller buffer is refused cleanly without the spurious critical.
- **[Nit] `tensordec-python3.cc:149-163` vs `:176-181`/`:253-258`** — the new checks mix `Py_ERRMSG` (prints+clears a pending Python exception) for the array/shape construction failures and plain `ml_loge` for the "wrong return type" checks. This is actually the semantically correct choice (only the former paths can leave a pending exception), but a one-line comment would save the next reader from having to verify that distinction themselves.

No Medium/High/Blocker issues found. The fix is well-scoped to F10, backed by tests that reproduce each sub-defect (confirmed by re-deriving the pre-fix crash mechanism by hand for all three), does not touch unrelated modules, doesn't change any public API/ABI, and is clang-format/doxygen clean.

Verdict: mergeable as is

myungjoo:

Thanks for the review. Both points were addressed or answered on the new head `6385b0bd` (still 3 commits; the change is folded into commit 2, `[Decoder/Python3] Check what the script returns before copying it`):

- **[Low] spurious `gst_buffer_resize_range` critical** — fixed in this PR. `decode()` now grows a caller buffer only when `maxsize >= offset + mem_size` (the same condition `gst_buffer_resize_range` asserts), otherwise it leaves the buffer as is and the mapped-size check refuses it. `decodeIntoSmallBuffer_n` now asserts that no assertion failure is logged, and has a second case where the room is lost only to the offset (maxsize 5, offset 2); `decodeIntoCallerBuffer` gained a positive case that grows from an offset. Mutations: ignoring `offset` or removing the check fails `decodeIntoSmallBuffer_n`, never resizing fails `decodeIntoCallerBuffer`.
- **[Nit] `Py_ERRMSG` vs `ml_loge`** — kept as is without a comment. As you noted, `Py_ERRMSG` is used exactly where a failed Python C-API call leaves an exception to print and clear, and `ml_loge` where the type check itself fails with nothing pending; that is the convention the rest of this file and `nnstreamer_python3_helper.cc` already follow, so a comment at these three sites would describe the macro rather than the code.

Local re-verification on the new head: each commit builds and passes on its own (decoder 14/18/20, converter 18), strict `-Werror -Wextra -Wsign-compare`, clang-format 16 and doxygen-tag clean, `meson test` of the three python3 binaries OK, 20x repeat stable.

myungjoo:

**Second-round review.** Re-read against the current head (`6385b0bd`), the previous relayed review and the author's reply on this PR, and the repository outside the diff (`gsttensor_decoder.c`'s NULL-caps guards in `gst_tensordec_get_media_caps_from_config`/`_get_media_caps`/`gst_tensordec_configure`/`gst_tensordec_fixate_caps`/`gst_tensordec_set_caps`, `tensor_converter_python3.cc`'s `py_open`, `tensor_filter_python3.cc`'s distinct `PYCore::loadScript`, `nnstreamer_python3_helper.cc`'s `getNumpyType`/`Py_SAFEDECREF`/`Py_ERRMSG`, `tests/meson.build`). Scope of review is the diff of this PR only, per the task; findings below are limited to that diff.

## Does the current head address F10?

Yes, all three sub-items are covered, unchanged from round 1's conclusion, re-verified independently:

- **(a)** `loadScript()` fails with `-3` when `PyObject_CallObject` on the class returns NULL (raising constructor) — `nnstreamer_python3_helper.cc:320-324` (commit `422d7c6d`). `decoder_py_getOutCaps`/`decoder_py_decode` refuse a NULL `core`, `decoder_py_exit` is a no-op on NULL — `tensordec-python3.cc:317-323,392-396,415-419` (same commit). I traced `PYConverterCore`'s constructor (`tensor_converter_python3.cc`, untouched) and confirmed `core_obj`/`shape_cls` are NULL-initialized before `init()` runs, so `delete core` on the raising-constructor failure path is safe — no new latent double-free.
- **(b)** `PyBytes_Check` guards both `decode()`'s and `getOutCaps()`'s results (`tensordec-python3.cc:176-181,257-262`, commit `b2055f97`), and the mapped-size check before `memcpy` (`:206-213`) is present with a proper `gst_memory_unmap`/`gst_memory_unref` on refusal.
- **(c)** `PyArray_SimpleNewFromData`/`PyTensorShape_New` results are checked before `PyList_SetItem` (`:149-154,158-163`, commit `6385b0bd`); `getNumpyType()` (`nnstreamer_python3_helper.cc:274-303`) returns `NPY_NOTYPE` for float16 (no case in the switch), which is exactly what the new check catches.

## Evaluation of the round-1 fix and the author's reply

**The [Low] `gst_buffer_resize_range` critical is correctly and completely fixed.** `tensordec-python3.cc:189-195`:
```c
if (gst_buffer_get_size (outbuf) < mem_size) {
  gsize offset, maxsize;
  gst_buffer_get_sizes (outbuf, &offset, &maxsize);
  if (maxsize >= offset + mem_size)
    gst_buffer_set_size (outbuf, mem_size);
}
```
`gst_buffer_get_sizes()`'s documented contract is exactly "offset and maxsize can be used to resize the buffer memory blocks with `gst_buffer_resize()`", i.e. a resize to `size` (with `gst_buffer_set_size` = `gst_buffer_resize(buf, 0, size)`, keeping the current start) is valid iff `offset + size <= maxsize`. That is precisely the guard added, so no `gst_buffer_resize_range` assertion can fire here any more, and the buffer is otherwise left at its original size so the existing `out_info.size < mem_size` check (already merged) refuses the copy — confirmed against all 5 new/changed `decodeIntoCallerBuffer`/`decodeIntoSmallBuffer_n` cases by hand:
  - `(maxsize=4,offset=0,size=4)`: already `== mem_size` (4), no resize needed, copies OK.
  - `(6,0,2)`: `6>=0+4` → resizes to 4, copies OK.
  - `(6,2,2)`: `6>=2+4` → resizes to 4 (spanning `area[2..6)`); bytes 0-1 and 6-7 untouched, matches the test's assertions.
  - `(3,0,3)`: `3>=0+4` is false → no resize, `out_info.size`(3) `< mem_size`(4) → refused, nothing past byte 3 touched.
  - `(5,2,2)`: `5>=2+4` is false → no resize, `out_info.size`(2) `< mem_size`(4) → refused, nothing past byte 4 touched.
  All five match the fix's intended behavior and the new assertion-count-is-zero expectation. This closes the finding; I found no remaining case where the guard under- or over-approximates what `gst_buffer_resize_range` itself would allow.
- **[Nit] `Py_ERRMSG` vs `ml_loge`** — declining it was the right call. `Py_ERRMSG` is used exactly at the two sites where a failed Python C-API call (`PyArray_SimpleNewFromData`, `PyTensorShape_New`) can leave a pending exception to print/clear; `ml_loge` is used at the three "wrong return type" checks, where no exception is pending (the call succeeded, just returned the wrong Python type). That is the same split the rest of the file and `nnstreamer_python3_helper.cc` already use (e.g. `Py_ERRMSG` in `loadScript()`'s existing `-1`/`-2` paths vs. plain `ml_loge` in unrelated type/attribute checks), so a comment at these three call sites would only restate the macro's own contract.

## Commit history (3 commits, one topic each)

Verified with `git show --stat` on `422d7c6d`, `b2055f97`, `6385b0bd` individually: commit 1 touches only the NULL-instance guards (helper `-3` return, `exit`/`getOutCaps`/`decode` NULL checks) plus its own tests/fixtures (`init-raise` mode, `raise_init_custom_converter.py`); commit 2 touches only the not-bytes/size checks in `decode()`/`getOutCaps()` (this is where the offset/maxsize fix for the round-1 Low actually lives, confirmed with `git show b2055f97 -- ext/nnstreamer/tensor_decoder/tensordec-python3.cc`) plus its own tests (`text` mode); commit 3 touches only the float16/NULL-array guard plus its own tests. No commit re-touches a hunk introduced by an earlier commit in this PR — none is a fix-up of another, each stands as its own coherent, independently buildable change (also consistent with the PR body's "each commit builds and passes on its own" claim, which I did not re-run but which is plausible from the diffs alone).

## Regressions / other modules

No change beyond round 1's findings. `loadScript()`'s new `-3` is handled by both call sites (`tensordec-python3.cc:295`, `tensor_converter_python3.cc:253`), both of which already treat any non-zero return as failure. `tensor_filter_python3.cc` has its own member `PYCore::loadScript()`, unaffected. `getOutCaps()` returning NULL is handled by every caller in `gsttensor_decoder.c` I could find (`gst_tensordec_get_media_caps_from_config`, `_get_media_caps`, the `configure`/`fixate_caps`/`set_caps` call sites at lines ~673, ~975, ~1046) — all guard `result == NULL` before use; that file is untouched by this PR (owned by #4986) and none of the new behavior requires it to change.

## Size / focus, tests, CI gating, docs

Unchanged from round 1's assessment, re-verified: only 6 files touched (`nnstreamer_python3_helper.cc` +5, `tensordec-python3.cc` +61/-6, two test files, two fixture `.py` files), ~450 of the ~510 changed lines are tests/fixtures for 3 well-isolated sub-defects — proportionate. `clang-format --style=file` (v16) is clean on all 4 changed C/C++ files (diffed locally, zero output). `tests/meson.build:338-356` builds `unittest_decoder_python3` under `if have_python3` (installs it under the added `flatbuf_support_is_available` condition too), so it runs in `meson test` wherever python3 is enabled — no CHANGES entry needed, no architecture/API/doc change (no public header or callback signature changed; only an internal negative return code and internal NULL guards were added), so no doc update is required.

## Findings

No Blocker/High/Medium/Low issues found in this round; the previously-reported Low is fixed and verified correct against its edge cases, and the previously-reported Nit was reasonably declined with an accurate rationale.

- **[Nit] `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc:502-534` (`decodeIntoSmallBuffer_n`)** — the test checks that bytes past the (unchanged) buffer aren't written, but never asserts `out_size` (already computed by `_decode_fixed_into`) stays at its pre-call value (3, then 2) on refusal. That would make explicit that the refused resize path leaves the buffer's reported size untouched, not just its backing bytes. Not a gap that would let a real bug slip through the existing byte-level checks, so optional.

Verdict: mergeable as is

myungjoo:

Round 2 Nit addressed on head `0eaf0131` (folded into commit 2, no new commit): both refusal cases in `decodeIntoSmallBuffer_n` now also assert that the buffer keeps its size (3 and 2 bytes). Each commit re-built and re-tested on its own (decoder 14/18/20, converter 18); strict build, clang-format 16 and doxygen-tag clean.

myungjoo:

**Third-round review.** Scope: the diff since round 2's head (`6385b0bd` -> current head `0eaf0131`), re-read against the full head as a whole. Round 1 and round 2 already concluded "mergeable as is" and their points (F10 coverage, regressions, size/focus, tests, CI gating, docs, the round-1 Low, the round-1 Nit) are not repeated here except where re-verification was needed for the round-2 Nit.

## Round-2 Nit fix, verified

`git diff 6385b0bd 0eaf0131` touches exactly one file, `tests/nnstreamer_decoder_python3/unittest_decoder_python3.cc`, +2/-0, folded into commit 2 (`b2055f97` -> `7ab4961c`; commit 3 is a mechanical rebase, `6385b0bd` -> `0eaf0131`, with no content change of its own — confirmed no other file differs between the two heads).

The two added lines are `EXPECT_EQ (out_size, 3U);` and `EXPECT_EQ (out_size, 2U);` in `decodeIntoSmallBuffer_n` (`unittest_decoder_python3.cc:768,779`), right after the existing `EXPECT_EQ (ret, GST_FLOW_ERROR)` for the two refusal cases. `out_size` is populated by the test helper `_decode_fixed_into()` as `gst_buffer_get_size (outbuf)` taken immediately after `decode()` returns (`:698`), so this is exactly what the round-2 Nit asked for: proof that a refused resize leaves the buffer's reported size at its pre-call value, not just its backing bytes untouched.

Checked both refused cases against the (maxsize, offset, size) table round 2 verified by hand:
- `(maxsize=3, offset=0, size=3)`: `3 >= 0+4` is false -> no resize attempted, buffer stays at its original size 3. New assertion `out_size == 3U` matches.
- `(maxsize=5, offset=2, size=2)`: `5 >= 2+4` is false -> no resize, buffer stays at 2. New assertion `out_size == 2U` matches.

Both new assertions are correct and exercise the actual refusal path (verified `gst_buffer_new_wrapped_full` is constructed with that same `size` as the buffer's initial size, so `out_size` unchanged is a meaningful, non-tautological check). No unrelated changes came along with the fix.

## Commit history

3 commits, one topic each, re-confirmed on the current head:
- `422d7c6d` — NULL-core guards (`getOutCaps`/`decode`/`exit`) + `loadScript()` `-3` return, with its own tests/fixtures.
- `7ab4961c` — `PyBytes_Check`/size-before-`memcpy` checks in `decode()`/`getOutCaps()`, **including** the offset/maxsize resize guard and now the round-2 `out_size` assertions, with its own tests.
- `0eaf0131` — float16/NULL-array guard for `PyArray_SimpleNewFromData`/`PyTensorShape_New`, with its own tests.

No commit re-touches a hunk introduced by an earlier commit in this PR, and no 4th "fixup" commit was added — the round-2 fix was folded into commit 2 via amend/rebase (changing `b2055f97`->`7ab4961c` and consequently `6385b0bd`->`0eaf0131`), which is the correct way to keep this a 3-commit, one-topic-per-commit series rather than growing a fixup tail. All 3 commits carry a `Signed-off-by:` line, no duplicates.

## Findings

None. No Blocker/High/Medium/Low/Nit issues found in this round.

Verdict: mergeable as is

myungjoo:

Thanks for the approval. The cosmetic note is correct and the PR description is now fixed: the PR adds **12 cases, 9 negative and 3 positive** (8 negative decoder cases plus `tensorConverterPython.raisingConstructor_n`; the positives are `pipelineScript`, `decodeIntoCallerBuffer` and `decodeMixedTypes`). All 9 negative cases fail on main, 4 of them with SIGSEGV. Description only; no code, test or commit changed, so the head stays `0eaf0131`.

PR #4997, opened 2026-09-17
[Trainer] Join the dummy-data thread, free replaced tensors info, bound the flex header parse (#4920 C31)

myungjoo:
item **C31** of #4920

- **What:** `tensor_trainer` had three separate memory defects. (a) Every PLAYING→PAUSED with the `nntrainer` framework overwrote `dummy_data_thread` without joining the previous thread, so that thread and its GThread leaked and could outlive the element. (b) The CAPS handler replaced `prop.input_meta` and `in_config` without freeing them, and finalize never freed `input_meta`, which leaks the extra-tensor block of any stream with more than 16 tensors. (c) A flexible tensor's 128-byte header was parsed with no size check, so a shorter memory was read past its end.
- **How:** one commit per defect, each with its tests. (a) join the previous thread before starting a new one, and join first in finalize. (b) free each info before it is replaced, and free `input_meta` in finalize after the sub-plugin's `destroy()`. (c) pass the mapped size along and refuse a memory smaller than the 128-byte header before `gst_tensor_meta_info_parse_header()` reads it (the same guard `tensor_decoder` uses, without mapping the memory a second time). `unittest_trainer` gets a fake trainer sub-plugin (registered as `nntrainer`, the only framework the dummy thread runs for) and 7 new cases.
- **Result:** `unittest_trainer` 18/18 locally, and rebased onto main `cf36cacd`. Each fix is shown to be needed by undoing it on its own: (a) `pauseTwiceJoinsDummyThread` fails (1 of 2 threads finished at finalize), and moving the finalize join back below the frees makes memcheck report an `Invalid read` of ours in the sub-plugin callback; (b) memcheck reports 3 definitely lost 19,200-byte blocks under `gst_tensor_trainer_sink_event()`, 0 with the fix; (c) `flexibleHeaderTruncated_n` crashes with SIGSEGV. No change outside `gsttensor_trainer.c` and its test.

### Notes for reviewers

- **What gates each fix in CI.** (a) and (c) fail the plain gtest run. (b) is gated only by the valgrind job's leak gate. `unittest_trainer` is built unconditionally (`tests/meson.build`, no feature guard) and is among the binaries the `GTEST run with Valgrind on a case` step runs; for example, run 35182902212 shows `==11160== Command: ./tests/unittest_trainer`. The leak is in `gst/`, so the gate counts it as ours both before and after #4987.
- **Why a guard page for (c).** Reading past a small heap block does not crash, and a valid-looking over-read header still ends in the same `GST_FLOW_ERROR` (the size check fails afterwards). The 8 bytes therefore end right before a `PROT_NONE` page: the pre-fix read faults deterministically, and memcheck is not needed for this one.
- **Join semantics for (a).** The second pause now waits for the first dummy thread. That thread ends once `cur_epoch_data_cnt` reaches `required_sample`, the same loop finalize already joined on, so no new wait condition is introduced. The finalize join moved above `g_mutex_clear`/`g_cond_clear`, because a sub-plugin's `stop()` running on that thread may call `nnstreamer_trainer_notify_event()`, which locks them.
- **A7 landed on main while this was open.** `304ac60e` makes the common `gst_tensor_meta_info_validate()` refuse a header version the build cannot size, so `flexibleHeaderUnknownVersion_n` is now a cross-module pin: it fails only if that refusal *and* this element's own `header_size == 0` refusal both go away. The element's guard is unchanged from main either way.
- **Commits:** `[Trainer] Join the dummy-data thread before starting another one`, `[Trainer] Free the tensors info before replacing it`, `[Trainer] Check the memory size before parsing a flexible header`. Each one builds and passes its own tests.

### Context for agents

- Branch `fix/4920-trainer-thread-leak-flex` on `myungjoo/nnstreamer`, base main `cf36cacd` (rebased; it started on `f865ca74`). Local tree `~/nns-4920-c31` (WSL). Mutation/valgrind scripts are `c31_mut.py`, `c31_vg.sh` and `c31_mutrun.sh` in the session scratchpad.
- Issue line numbers refer to `4e2522f7`. On `f865ca74` the sites were `:556` (thread), `:1013-1016` + finalize `:316-344` (info), and `:670` called from `:722` (flex header).
- Memcheck on the fixed build: 0 errors and 0 definite leaks from the repository; the one remaining 16 KB block is glib's loader-time init.
- Checked: `gst-indent` clean, clang-format 16 clean, `doxygen-tag.sh` clean (with ctags, plus a negative control), strict `-Wextra -Wsign-compare -Werror` compile of both objects.


myungjoo:

Scope reviewed: the diff of this PR only (`gst/nnstreamer/elements/gsttensor_trainer.c`, `tests/nnstreamer_trainer/unittest_trainer.cc`), cross-checked against the rest of the repo (`gst_tensors_info_copy`/`_free`, `gst_tensors_config_free`, `gst_tensor_meta_info_parse_memory`/`_parse_header` in `nnstreamer_plugin_api_util_impl.c` / `nnstreamer_plugin_api_impl.c`, `nnstreamer_subplugin.c`, `tests/meson.build`) and against issue #4920 item C31.

Yes, for all three sub-defects of C31.

- **(a) thread leak**: `gst_tensor_trainer_change_state()` (`gsttensor_trainer.c:558-559`) now joins the previous `dummy_data_thread` before starting a new one, and `gst_tensor_trainer_finalize()` (`:321-324`) joins first, before `g_mutex_clear`/`g_cond_clear` (`:334-337`) and before `fw->destroy()` (`:339-341`). This is the right order: the dummy thread calls `fw->stop()`/`fw->push_data()`, which may call back into `nnstreamer_trainer_notify_event()` and lock those mutexes, and it reads `trainer->prop.*`, which the old code freed (`:326-329`) *before* the join. So this fix also happens to close a second, unfiled use-after-free window (thread reading freed `prop.model_config`/`model_save_path` after the old late join) — worth noting in case this is later cited as a separate defect; it isn't a new problem introduced here.
- **(b) `input_meta`/`in_config` leak**: verified against `gst_tensors_info_copy()` (`nnstreamer_plugin_api_util_impl.c:530-549`) — it only calls `gst_tensors_info_init()` on `dest`, it does not free a pre-existing `dest`. So the pre-existing calls at `gsttensor_trainer.c:1018` and `:1026` were genuine leaks whenever `num_tensors > NNS_TENSOR_MEMORY_MAX` (i.e. `info->extra` was allocated, see `gst_tensors_info_get_nth_info()` at `nnstreamer_plugin_api_util_impl.c:333-341`). The added `gst_tensors_info_free()`/`gst_tensors_config_free()` calls at `:1017`, `:1021`, `:1025` are correctly placed before each replacement, and `gst_tensors_info_free(&trainer->prop.input_meta)` in finalize (`:343`) is correctly placed *after* `fw->destroy()` so the sub-plugin still sees valid `input_meta` during its own teardown.
- **(c) OOB header read**: `gst_tensor_meta_info_parse_memory()` (`nnstreamer_plugin_api_impl.c:1646-1672`) checks `msize < hsize` before mapping, unlike the removed `gst_tensor_meta_info_parse_header()` path which trusted the caller-supplied pointer. The call site change at `gsttensor_trainer.c:674` (passing `GstMemory *mem` instead of the mapped `void *data`) is correct, and the existing error path (`:755-766`) already unmaps/unrefs every `in_mem[i]` cleanly on the `goto error` this fix now takes.

None found. The only externally-visible signature change is to the `static` function `gst_tensor_trainer_convert_meta()` (file-local, single call site, confirmed with a repo-wide grep — no other file calls it). No public API/header touched. No other element or module is touched by this diff.

One behavior change worth flagging even though it isn't a bug: the new `g_thread_join()` at `gsttensor_trainer.c:558-559` makes a second PLAYING→PAUSED transition block until the previous dummy-data thread's `fw->stop()`/`fw->push_data()` finish, whereas before it returned immediately (leaking the thread instead). This is the correct fix and is called out by the author in the PR's "Notes for reviewers" section, but it is a real latency/blocking-during-state-change change for any pipeline that pauses/plays `tensor_trainer` repeatedly with a slow `nntrainer` sub-plugin. Severity: **minor** (not a defect, just worth a maintainer's awareness — no action required).

## Code size / scope
Proportionate: 3 commits, one per sub-defect, touching only `gsttensor_trainer.c` and its own test file, matching the PR's own claim of "no change outside gsttensor_trainer.c and its test."

## Test coverage
Good and appropriately layered:
- `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread` regress (a); traced through the fake sub-plugin's `stop_calls`/`push_done` counters, which is a reasonable substitute for directly observing the joined `GThread*`.
- `renegotiateManyTensors` regresses (b) by pushing 17- then 18-tensor caps (exceeding `NNS_TENSOR_MEMORY_MAX` = 16, forcing the `info->extra` allocation path) through a `GstHarness`. As the PR body says, this test cannot see the leak itself — only the CI valgrind job's leak gate can — which is an accepted, precedented pattern in this codebase (cf. prior C-series leak fixes referenced in #4920).
- `flexibleInput` / `flexibleHeaderTruncated_n` / `flexibleHeaderInvalid_n` regress (c), and the guard-page (`mmap`+`mprotect(PROT_NONE)`) technique for `flexibleHeaderTruncated_n` is consistent with prior use of the same idiom elsewhere in the test suite (`unittest_plugins.cc`, `unittest_filter_custom.cc`, `unittest_decoder_boundingbox.cc`), not a new pattern.
- The fake sub-plugin is registered only in `unittest_trainer` (verified: `tests/meson.build:527-532` builds it only against `nnstreamer_unittest_deps`, no dependency on a real `nntrainer` sub-plugin build), so there is no risk of the fake `"nntrainer"` registration colliding with a real one linked into the same test binary.

## Will CI actually fail without the fix / catch a future regression?
- (a) and (c): yes — `pauseTwiceJoinsDummyThread`/`finalizeJoinsDummyThread` assert exact `stop_calls`/`push_done` counts, and `flexibleHeaderTruncated_n` would SIGSEGV outright without the fix (confirmed by the guard-page construction and `gst_tensor_meta_info_parse_memory`'s size check). Both are caught by the plain (non-valgrind) `meson test` run, which is a cheap, always-on gate.
- (b): only the valgrind job's leak gate catches this (`unittest_trainer` is swept by `run_unittests_binaries.sh --valgrind --recursive ./tests/` per `.github/workflows/ubuntu_clean_meson_build.yml`, confirmed the script's recursive discovery would include it, and confirmed the fix path is under `gst/`, so the leak-attribution gate counts it as "ours"). This is consistent with how prior leak-only fixes in the #4920 series have been gated. If that workflow job is not a required status check on this repo, a future regression of (b) alone would not block a merge — but that is a pre-existing property of the CI setup, not something this PR should be expected to change.

## Documentation
No architecture or public API change; no doc update needed or expected.

## Findings summary

| Severity | Location | Note |
|---|---|---|
| minor | `gst/nnstreamer/elements/gsttensor_trainer.c:558-559` | New synchronous `g_thread_join()` on PLAYING→PAUSED makes a second pause block on the previous dummy-data thread finishing; correct and necessary, but a real (and currently undocumented outside the PR description) latency change for repeated pause/play with a slow sub-plugin. No action required — flagging for awareness only. |
| nit | `tests/nnstreamer_trainer/unittest_trainer.cc:695-706` (`flexibleHeaderTruncated_n`) | If the `ASSERT_EQ(mprotect(...), 0)` fails, the preceding `mmap()`'d 2-page region is never `munmap()`'d (early return from the assertion). Not a heap leak (won't trip the valgrind gate) and mirrors an existing idiom already used elsewhere in this test suite, so not a new class of issue — cosmetic only. |

**Overall verdict: mergeable as-is.**

myungjoo:

Thanks for the review. The branch was force-pushed to `fecdbda4`; the change is folded into the third commit, so there is no fix-up commit.

- **nit (`flexibleHeaderTruncated_n` leaves the 2-page mapping behind on an early assertion failure):** fixed. The guard page is now mapped and protected before the harness is created, and each failure path after `mmap()` (the `mprotect()` refusal and the harness creation) unmaps the region before `FAIL()`. Re-checked: clang-format 16 clean, strict `-Wextra -Wsign-compare -Werror` compile clean, 6/6 fake-sub-plugin cases pass, and with the (c) fix reverted the test still dies with SIGSEGV (exit 139).
- **minor (the second PLAYING→PAUSED now waits for the previous dummy-data thread):** kept as is, on purpose. I checked it against the real sub-plugin, nntrainer's `nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc`. Its `push_data()` blocks only while its bounded queue is full, and the training thread drains that queue. The dummy thread pushes exactly the rest of the epoch that thread waits for, so the wait is bounded. It is also the same loop finalize already joined on before this PR. `Documentation/component-description.md` is the only document naming `tensor_trainer`, and it does not describe state-change behavior, so no document needs updating.
- On the CI remark: `Valgrind ubuntu-22.04` is one of the required status checks on `main`, so a regression of (b) alone would block a merge.

myungjoo:

This reviews head `fecdbda4` (the force-pushed head after the previous review round; the fix-up for the earlier nit is folded into the third commit, `fecdbda4`). I re-read the whole diff, `gsttensor_trainer.c` in full, `gst_tensor_meta_info_parse_memory()`/`_parse_header()` and `gst_tensors_info_copy()`/`_free()` in `nnstreamer_plugin_api_impl.c`/`nnstreamer_plugin_api_util_impl.c`, `tests/meson.build`, and the real `nntrainer` sub-plugin at `nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc` (separate `nntrainer` repo) independently, rather than relying on the prior round's review.

## Does it solve C31?

Yes, all three sub-defects.

- **(a) thread leak / late-join UAF window**: `gst_tensor_trainer_change_state()` (`gsttensor_trainer.c:558-559`) now joins `dummy_data_thread` before overwriting the handle on a second `PLAYING_TO_PAUSED`; `gst_tensor_trainer_finalize()` (`:321-324`) joins first, before `g_mutex_clear`/`g_cond_clear` (`:334-337`) and `fw->destroy()` (`:339-341`). Verified: with the fix removed, `pauseTwiceJoinsDummyThread` overwrites the handle while the fake thread is still sleeping inside `push_data()`, so `finalize()` only joins the second thread and the first is leaked/dangling — the added assertions on `stop_calls`/`push_done` (both must equal 2) catch exactly that.
- **(b) `input_meta`/`in_config`/`out_config.info` leak**: confirmed `gst_tensors_info_copy()` (`nnstreamer_plugin_api_util_impl.c:530-549`) only `gst_tensors_info_init()`s the destination, never frees a pre-existing one — so every caps renegotiation before this fix leaked the previous `info.extra` block once `num_tensors > NNS_TENSOR_MEMORY_MAX` (16). The added `gst_tensors_info_free()`/`gst_tensors_config_free()` calls at `gsttensor_trainer.c:1017`, `:1021`, `:1025` sit correctly before each replacement, and `gst_tensors_info_free(&trainer->prop.input_meta)` in finalize (`:343`) is correctly placed *after* `fw->destroy()`, so the sub-plugin's teardown still sees valid `input_meta`.
- **(c) OOB header read**: `gst_tensor_meta_info_parse_memory()` (`nnstreamer_plugin_api_impl.c:1646-1671`) computes `msize = gst_memory_get_sizes(mem, ...)` and returns `FALSE` when `msize < hsize` *before* calling `gst_memory_map()`, unlike the removed `gst_tensor_meta_info_parse_header()` call which read straight from the caller-supplied pointer with no bound. The call-site change (`gsttensor_trainer.c:670`, from `:727`, passing `GstMemory *mem` instead of the pre-mapped `void *data`) is correct and the existing `error:` cleanup (`:755-766`) already unmaps/unrefs every `in_mem[0..n)` on the `goto error` this path now takes.

None found. `gst_tensor_trainer_convert_meta()` is `static`, single call site (`grep -rn "gst_tensor_trainer_convert_meta" gst/ ext/` — only the definition and the one call in the same file), so the signature change (`void *data` → `GstMemory *mem`) is invisible outside this file. No header, public API, or other element is touched.

One thing worth independently confirming rather than taking on faith, since the earlier round's response asserted it without a citable check: the claim that the second `PLAYING→PAUSED`'s new blocking `g_thread_join()` (`:558-559`) is bounded in the real `nntrainer` sub-plugin. I read `NNTrainer::TensorsQueue::push()` in `tensor_trainer_nntrainer.cc:201-`: it does `data_full.wait(lock, [&]{ return !isQueueFull(); })`, i.e. it blocks only while the bounded queue is full, and the training thread's `pop()` (`:237-`) drains it via `data_empty`/`data_full` condvars — so the wait the join now exposes is indeed bounded by the training thread's consumption rate, not unbounded. Confirmed, not just asserted.

Also confirmed independently (not just from the PR author's claim) that a regression of (b) alone would actually block a merge: `gh api repos/nnstreamer/nnstreamer/branches/main/protection --jq '.required_status_checks.checks[].context'` lists `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` as a required check, and `.github/workflows/ubuntu_clean_meson_build.yml:309` runs `run_unittests_binaries.sh --valgrind --recursive ./tests/`, which sweeps `unittest_trainer` (a plain GTest binary, not one of the SSAT-helper/dlopen cases known to break the recursive discovery).

## Code size / scope

Proportionate: 3 commits, one per sub-defect, touching only `gsttensor_trainer.c` and its own test file.

## Test coverage and CI enforcement

- (a): `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread` assert exact `stop_calls`/`push_done` counts and would fail under plain (non-valgrind) `meson test` without the fix.
- (b): `renegotiateManyTensors` pushes 17- then 18-tensor caps through a `GstHarness`, forcing the `info.extra` allocation; the test itself can't observe the leak, only the valgrind job's leak gate can, which — per the check above — is a required status check, so this will actually block a bad merge.
- (c): `flexibleHeaderTruncated_n` places an 8-byte memory immediately before a `PROT_NONE` guard page; without the fix, `gst_tensor_meta_info_parse_header()` reads 128 bytes from an 8-byte region and faults. I re-verified the earlier round's nit fix is actually applied at this head: both early-exit paths after `mmap()` (the `mprotect()` failure at `:919-922` and the harness-creation failure at `:925-928`) now `munmap()` before `FAIL()`, so the 2-page mapping is no longer leaked on those paths.
- The fake sub-plugin is registered/unregistered per-test via the `TensorTrainerFakeFw` fixture's `SetUp`/`TearDown`, and I confirmed with `git diff f865ca74 fecdbda4 -- tests/meson.build` that this PR does not touch the build file. This looked initially like a potential link-time risk (`unittest_trainer`'s target only lists `dependencies: [nnstreamer_unittest_deps]`, which doesn't obviously include the core `nnstreamer` plugin object that implements `nnstreamer_trainer_probe`/`_exit`, `gsttensor_trainer.c:1374,1402`), but tracing the dependency graph (`tests/meson.build:29-37`: `unittest_util_dep` is `declare_dependency(... dependencies: nnstreamer_dep ...)`, and `nnstreamer_unittest_deps` includes `unittest_util_dep`) confirms Meson propagates `nnstreamer_dep` transitively into every unittest binary including `unittest_trainer`, so the new calls do link. Not an issue, but worth recording since it wasn't obvious from the changed files alone.

## Documentation

No architecture or public API change. `Documentation/component-description.md` (the only doc naming `tensor_trainer`) doesn't describe state-change timing, so no update is needed, matching the author's prior response.

## New finding (nit, not blocking)

- **nit** — `gst/nnstreamer/elements/gsttensor_trainer.c:719` + `:674` (via `:727`): in the flexible-input path, `in_mem[i]` is already mapped with `GST_MAP_READ` at `:719` (`in_info[i]` stays valid for the rest of the loop iteration), and then `gst_tensor_trainer_convert_meta()` → `gst_tensor_meta_info_parse_memory()` maps the *same* `GstMemory` a second time (`nnstreamer_plugin_api_impl.c:1663`) just to read the header, unmapping again right after. This is not a correctness bug — GstMemory's read-mode mapping is a shared/re-entrant lock (multiple concurrent `GST_MAP_READ` mappings of the same memory are explicitly supported), and `gsttensor_converter.c:1203` calls the same helper without an outer map already held, so the helper is used correctly by its own contract — but it is a redundant map/unmap pair per flexible tensor per buffer that the decoder's equivalent fix for C24 avoids (`gsttensor_decoder.c:734-783`, `gst_tensordec_check_input_size()`, checks `size < gst_tensor_meta_info_get_header_size(...)` and then calls `gst_tensor_meta_info_parse_header()` directly on the already-mapped pointer, no second map). Optional cleanup only; negligible cost for one flexible header per buffer, no action required.

## Verdict

No blocking or major issues. The one nit above is cosmetic/perf-only and does not need to hold up this PR.

**Overall verdict: mergeable as-is.**

myungjoo:

Round-2 nit (the flexible path mapped the same memory twice) is addressed in `83d6554e`, folded into the third commit; its message is updated to match. `gst_tensor_trainer_convert_meta()` now takes the pointer and size that `push_input()` has already mapped. It runs `gst_tensor_meta_info_init()` and refuses `size < gst_tensor_meta_info_get_header_size()` before `gst_tensor_meta_info_parse_header()` reads the header, the same guard `gst_tensordec_check_input_size()` uses. The diff against main is also smaller, since the original `void *data` parameter stays.

Re-verified on the new head: `gst-indent` clean, strict `-Wextra -Wsign-compare -Werror` build clean, `unittest_trainer` 17/17, memcheck 0 errors and 0 definite leaks from the repository. With only the size guard removed, `flexibleHeaderTruncated_n` dies with SIGSEGV (exit 139). The PR body is updated too.

myungjoo:

This reviews head `83d6554e` (the head after the round-2 nit fix was folded into the third commit). I re-read the full diff, `gsttensor_trainer.c` in full, `gst_tensor_meta_info_init/_get_header_size/_parse_header/_convert/_validate` in `nnstreamer_plugin_api_util_impl.c`, `gst_tensors_info_copy/_free`, `tests/meson.build`, the new test file in full, and the real `nntrainer` sub-plugin at the local read-only checkout of `nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc`, independently of the two earlier review rounds recorded on this PR.

## Does it solve C31?

Yes, all three sub-defects, confirmed by direct code reading (not by re-quoting the prior rounds):

- **(a) thread leak / late-join UAF window**: `gst_tensor_trainer_change_state()` (`gsttensor_trainer.c:558-559`) joins the previous `dummy_data_thread` before overwriting the handle on a second `PLAYING_TO_PAUSED`; `gst_tensor_trainer_finalize()` (`:321-324`) joins first, before `g_mutex_clear`/`g_cond_clear` (`:334-337`) and before `fw->destroy()` (`:339-341`), and `gst_tensors_info_free(&trainer->prop.input_meta)` was moved to run *after* `fw->destroy()` (`:343`), so the sub-plugin's own teardown still sees valid properties. This ordering is correct and internally consistent.
- **(b) `input_meta`/`in_config`/`out_config.info` leak**: verified `gst_tensors_info_copy()` (`nnstreamer_plugin_api_util_impl.c:530-549`, i.e. only `gst_tensors_info_init()`s the destination, never frees a pre-existing one) so every caps renegotiation before this fix leaked the previous `info.extra` block once `num_tensors > NNS_TENSOR_MEMORY_MAX` (16). The added `gst_tensors_info_free()`/`gst_tensors_config_free()` calls (`gsttensor_trainer.c:1020`, `:1024`, `:1028`) sit correctly before each replacement; `trainer->in_config = config;` at `:1025` is a correct ownership transfer of the local stack struct (no double free, `config` is never touched again).
- **(c) OOB header read**: independently traced the whole chain — `gst_tensor_meta_info_init(meta)` (`:674`) forces `meta->version` to this build's current version before the first size check, so `size < gst_tensor_meta_info_get_header_size(meta)` (`:676`) at that point is checking the buffer's mapped size against *this build's* 128-byte V1 header, not anything attacker-controlled; only after that gate does `gst_tensor_meta_info_parse_header(meta, data)` (`:677`) touch the bytes. This is the exact idiom already used at `gsttensor_decoder.c:744-747`, `gsttensor_crop.c:574,595`, `gsttensor_sparseutil.c:42-43`, and `nnstreamer_plugin_api_impl.c:1655,1658` (`gst_tensor_meta_info_parse_memory`) — confirmed by grep, not assumed.

None found, confirmed independently:
- `grep -n "gst_tensor_trainer_convert_meta" -r gst/ ext/` returns only the definition (`:669`) and its one call site (`:729`) — the signature change (`GstMemory *` in round 2 back to `void *data, gsize size` in round 3) is invisible outside this file.
- `Documentation/` has exactly one mention of `tensor_trainer` (`Documentation/component-description.md:102`, a stable-status link), no state-change or API documentation exists to update; matches the PR's own claim and both prior rounds.
- Re-confirmed the previously-flagged behavior change (second `PLAYING→PAUSED` now blocks on the prior dummy-data thread) is bounded, not by taking the author's word but by reading `NNTrainer::TensorsQueue::push()` in the real sub-plugin (`tensor_trainer_nntrainer.cc:201-232`): it waits on `data_full` only `while (isQueueFull())`, and `pop()` (`:237-281`) drains the queue and signals `data_full` back — so the join is bounded by the training thread's consumption rate. Not unbounded.
- Independently re-confirmed via `gh api repos/nnstreamer/nnstreamer/branches/main/protection --jq '.required_status_checks.checks[].context'` that `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` is currently a required check on `main`, and via `.github/workflows/ubuntu_clean_meson_build.yml:309` that `run_unittests_binaries.sh --valgrind --recursive --skip unittest_filter_llamacpp ./tests/` sweeps `unittest_trainer` (not excluded) with `check_valgrind_log.sh --require-output` gating the job. So a regression of (b) alone would in fact block a merge today.
- Ran `clang-format -style=file` on `tests/nnstreamer_trainer/unittest_trainer.cc` against the repo's `.clang-format`: zero diff, confirming the C++ test file's formatting claim rather than trusting it. (`gst-indent`/GNU `indent` in the available WSL environment mis-parses this codebase's macros unrelated to this diff and reformats the whole file destructively — a known environment/tooling limitation, not evidence of a real style problem in the C file; not counted as a finding either way.)

## Code size / scope

Proportionate: 3 commits, one per sub-defect, touching only `gst/nnstreamer/elements/gsttensor_trainer.c` and its own test file. No other module touched.

## Test coverage and CI enforcement

- (a): `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread` assert exact `stop_calls`/`push_done` counts. Traced the timing by hand: `gst_tensor_trainer_stop_model_training()` runs synchronously at the top of the dummy-data thread function before the (delayed) `push_data()` call, so `wait_for_stop_calls(1)` reliably observes the thread has started before the test drives the next state transition — this is deterministic, not a sleep-race.
- (b): `renegotiateManyTensors` forces the `info.extra` allocation path (17 then 18 tensors) through a real `GstHarness`; the test itself cannot observe the leak, only the valgrind job's leak gate can, which (per above) is a required check.
- (c): `flexibleHeaderTruncated_n` places an 8-byte memory immediately before a `PROT_NONE` guard page; verified both early-return paths after `mmap()` (`:920` on `mprotect()` failure, `:926` on harness-creation failure) now `munmap()` before `FAIL()`, and the success path also `munmap()`s at `:940` — the round-2 nit is fixed and I did not find a remaining leak path in this test.
- The guard-page (`mmap`+`mprotect(PROT_NONE)`) idiom is precedented: `grep -rl mprotect tests/` also matches `unittest_decoder_boundingbox.cc`, `unittest_filter_custom.cc`, `unittest_plugins.cc`, so this is not a new portability pattern for the test suite.
- `tests/meson.build` is unchanged by this PR (`unittest_trainer` still only depends on `nnstreamer_unittest_deps`, which is verified elsewhere to transitively carry `nnstreamer_dep`), consistent with the diff.

## New observation (nit, not blocking)

- `gst/nnstreamer/elements/gsttensor_trainer.c:669-688` (`gst_tensor_trainer_convert_meta`): a flexible header with a valid magic marker but an unrecognized/unsized meta version (the still-open, separately-tracked issue #4920 item A7) is *not* rejected by an explicit check the way `gsttensor_decoder.c:754-761` does (`gst_tensordec_check_input_size()` re-reads `hsize` after `parse_header()` and explicitly fails when it is 0). I traced this function by hand and it is still safe today: `gst_tensor_meta_info_convert()` only requires `gst_tensor_meta_info_validate()`, which accepts any version with the magic marker, so for an unknown version `header_size = gst_tensor_meta_info_get_header_size(meta)` (`:683`, evaluated on the *parsed*, buffer-supplied version) correctly comes back `0`, and the caller (`gsttensor_trainer.c:731-732`) already treats any `header_size == 0` as `goto error`. So there is no live defect — but the safety is incidental to this function returning 0 on any failure, rather than an explicit, self-documenting guard, and there is no regression test here for "valid magic, unrecognized version" the way the decoder's C24 fix has. If the shared helper (`gst_tensor_meta_info_get_header_size`/`_validate`) is ever changed while fixing A7 (still unchecked in #4920), this call site has no test that would catch a regression here specifically. Suggested fix (optional, low priority, can be deferred to whichever PR closes A7): add the same explicit `hsize == 0` check for readability/self-documentation, and/or a small test constructing a header with the magic marker but a non-V1 version.

## Documentation

No architecture or public API change; confirmed no doc needs updating (see above).

## Verdict

No blocking or major issues. The one nit above is a defense-in-depth/readability observation about an already-tracked, separately-scoped issue (A7) — the code as written is safe against it today, so it does not need to hold up this PR.

**Overall verdict: mergeable as-is.**

myungjoo:

Round-3 nit (an unsizable-version header is refused only through the implicit `header_size == 0` check, and no test covers it): addressed with a test, folded into the third commit (`14f58223`).

- `flexibleHeaderUnknownVersion_n` builds a valid V1 header and rewrites its version to `0xDE002000`, which validates but has no known header size, and declares a data size equal to the whole 256-byte memory. The size check that follows cannot reject it, so only the header refusal stops the header bytes from reaching the sub-plugin as data. Removing `if (header_size == 0) goto error;` from `push_input()` makes the case fail (flow OK, `push_data` called once); with the refusal it passes. Memcheck on the fake-sub-plugin cases: 0 errors and 0 definite leaks from the repository.
- I kept the check itself implicit rather than adding a decoder-style explicit branch. That code is unchanged from main, and the general fix for unsizable versions is item A7 (`gst_tensor_meta_info_validate()` / `get_header_size()`), which belongs to the shared util rather than to this element. The new test pins the trainer's behavior either way.

The PR body is updated (7 new cases, 18/18).

myungjoo:

This reviews head `14f58223` (round 4). Since the previous round (`83d6554e`), the only change is one added test, `flexibleHeaderUnknownVersion_n`, folded into the third commit, plus that commit's message being extended to describe it (confirmed with `git diff 83d6554e 14f58223`: +35 lines, all in `tests/nnstreamer_trainer/unittest_trainer.cc`, no change to `gst/nnstreamer/elements/gsttensor_trainer.c`). I re-verified the whole PR independently rather than re-quoting the three earlier rounds recorded on this thread: the full diff, `gst/nnstreamer/elements/gsttensor_trainer.c` in full, the relevant `gst_tensor_meta_info_*`/`gst_tensors_info_*`/`gst_tensors_config_*` implementations in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c` and `nnstreamer_plugin_api_impl.c`, `tests/meson.build`, and the branch-protection/workflow configuration.

## Does it solve C31?

Yes, all three sub-defects, re-derived from source rather than taken on faith:

- **(a) thread leak / late-join UAF window** (`gsttensor_trainer.c:558-559`, finalize `:321-324`): the previous `dummy_data_thread` is now joined before a second `PLAYING_TO_PAUSED` overwrites the handle, and finalize's join was moved to the very first statement of the function, ahead of `g_free(trainer->fw_name/model_config/model_save_path/model_load_path)` (`:326-329`). I traced the pre-PR code and confirmed those `g_free()` calls already ran *before* the join in the old ordering (the join used to sit after the mutex/cond clears), so the dummy thread's `fw->push_data()` could read those fields after they were freed even in single-thread-only scenarios. Moving the join to the top closes that window, which is a correctness improvement beyond the literal "leak" description and is exercised by `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread`.
- **(b) `input_meta`/`in_config`/`out_config.info` leak** (`gsttensor_trainer.c:1020,1024,1028`, finalize `:343`): confirmed `gst_tensors_info_copy()` (`nnstreamer_plugin_api_util_impl.c:376-395` for `_free`, the paired `_copy` only `_init()`s the destination) never frees a pre-existing destination, so every caps renegotiation once `num_tensors > NNS_TENSOR_MEMORY_MAX` (16) leaked the previous `info.extra` block before this fix. The added `gst_tensors_info_free()`/`gst_tensors_config_free()` calls sit correctly before each replacement; `trainer->in_config = config;` (`:1025`) is a correct ownership transfer of a local stack struct (not touched again afterward, so no double free); `gst_tensors_info_free(&trainer->prop.input_meta)` in finalize (`:343`) correctly runs after `fw->destroy()` (`:339-341`) so the sub-plugin still sees valid `input_meta` during its own teardown.
- **(c) OOB header read** (`gsttensor_trainer.c:669-688`, call site `:729-730`): re-derived the bit-level logic of the new `flexibleHeaderUnknownVersion_n` test rather than trusting its description. `GST_TENSOR_META_VERSION_VALID(v)` is `(v & 0xDE000000) == 0xDE000000` and `GST_TENSOR_META_IS_V1(v)` additionally requires `(v & 0x00FFF000) & 0x1000`. The test builds a normal V1 header via `gst_tensor_meta_info_update_header()` and then overwrites only the version word to `0xDE002000`: this still satisfies `VERSION_VALID` (so `gst_tensor_meta_info_validate()` and therefore `parse_header()`/`convert()` all still succeed) but fails `IS_V1`, so `gst_tensor_meta_info_get_header_size()` returns 0. `gst_tensor_trainer_convert_meta()` (`:669-688`) therefore returns `header_size == 0`, and `push_input()` (`:731-732`) already treats any `header_size == 0` as `goto error` — exactly the guard this PR's third commit relies on, and exactly the gap round 3 flagged as an implicit-rather-than-explicit check. Confirmed by hand-tracing the macros, not by re-running the earlier rounds' reasoning.

None found, re-confirmed independently:
- `grep -rn "gst_tensor_trainer_convert_meta" gst/ ext/` returns only the `:669` definition and the `:729` call site — the function stays file-local `static`, so its signature is invisible outside this file.
- `tests/meson.build` is unmodified by this PR; `unittest_trainer` (`tests/meson.build:527-534`) is a plain `executable()` + `test()` pair, matching how the round-2 review traced `nnstreamer_unittest_deps` -> `unittest_util_dep` -> `nnstreamer_dep` linkage.
- Independently ran `gh api repos/nnstreamer/nnstreamer/branches/main/protection --jq '.required_status_checks.checks[].context'`: confirms `Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)` is a required check today. Independently read `.github/workflows/ubuntu_clean_meson_build.yml:309`: `run_unittests_binaries.sh --valgrind --recursive --skip unittest_filter_llamacpp ./tests/` does not exclude `unittest_trainer`, so a leak regression of (b) would actually fail a required check, not just an optional job.
- Ran `clang-format -style=file` on `tests/nnstreamer_trainer/unittest_trainer.cc` against this repo's `.clang-format`: zero diff at head `14f58223`, so the newest test addition is formatted correctly too.

## Code size / scope

Proportionate: 3 commits, one per C31 sub-defect, touching only `gst/nnstreamer/elements/gsttensor_trainer.c` and its own test file. No other element, header, or public API is touched.

## Test coverage and CI enforcement

- (a): `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread` assert exact `stop_calls`/`push_done` counts against a fake `nntrainer` sub-plugin registered only for this test binary; both fail under a plain (non-valgrind) `meson test` without the fix, so they are caught by a fast, always-on gate.
- (b): `renegotiateManyTensors` forces the `info.extra` allocation path (17 then 18 tensors) through a real `GstHarness`. The test cannot observe the leak itself, only the valgrind job's leak gate can — but that gate is a required status check, as re-confirmed above.
- (c): `flexibleInput` (positive), `flexibleHeaderTruncated_n` / `flexibleHeaderInvalid_n` / `flexibleHeaderUnknownVersion_n` (negative, `_n` suffix per this repo's convention) cover a valid header, a too-short memory (guard-page technique, precedented elsewhere in the test suite per earlier rounds), a long-enough-but-invalid header, and now a header that validates but cannot be sized. Together these are both positive and negative cases as required by `CLAUDE.md`'s test-case policy, and the new case closes the one gap (an explicit test for the A7-adjacent "valid magic, unsizable version" input) that round 3 flagged as a non-blocking nit.

## Documentation

No architecture or public API change. `Documentation/component-description.md` is the only document naming `tensor_trainer` and does not describe state-change timing or the flexible-header parsing path, so no documentation update is needed, consistent with all three earlier rounds.

## Findings

No blocking, major, minor, or nit issues found in this round. The one nit raised in round 3 (unsizable-version header rejection relying on an implicit `header_size == 0` check, with no dedicated test) is resolved by the `flexibleHeaderUnknownVersion_n` test added since then; I independently re-derived why that test exercises exactly that path rather than accepting the commit message's description at face value.

## Verdict

**Overall verdict: mergeable as-is.**

myungjoo:

Thanks — the NIT is addressed, and the branch is rebased onto current main. New head `6070b58f`; the change is folded into the first commit, so there is still no fix-up commit.

- **NIT (the finalize join reorder was not covered):** `finalizeJoinsDummyThread` now covers it. The fake sub-plugin reads `prop->model_config` inside `push_data()`, which is exactly what the dummy thread does while the element is finalized, and the test asserts the sub-plugin saw the real path. With the join moved back below `g_free`/`g_mutex_clear`, the native run failed here outright and memcheck reports `Invalid read of size 1` at the sub-plugin callback, with the block freed by `gst_tensor_trainer_finalize()`; that error is attributed to this repository, so the Valgrind job fails on it. With the join at the top: 0 errors and 0 definite leaks of ours.
- **Rebase note.** Main gained `304ac60e` (item A7) meanwhile, which makes the common `gst_tensor_meta_info_validate()` refuse a header version this build cannot size. `flexibleHeaderUnknownVersion_n` therefore no longer discriminates this element's own `header_size == 0` refusal on its own; it stays as a cross-module pin, failing only if both refusals go away. Everything else was re-verified on the new base: 18/18 tests, `gst-indent` and clang-format 16 clean, strict `-Wextra -Wsign-compare -Werror` build clean, and the mutation matrix (thread join, each free, the header size guard) still discriminates.

Back to draft with the hold label while CI re-runs on the rebase.

myungjoo:

This reviews head `6070b58f` (3 commits, rebased onto main `cf36cacd`). I re-read the whole diff independently, cross-checked the flexible-header logic against the common `nnstreamer_plugin_api_util_impl.c` helpers (including A7 / `304ac60e`, which landed on main while this PR was open), checked the real `nntrainer` sub-plugin (`tensor_trainer_nntrainer.cc`) for how it actually uses `GstTensorTrainerProperties`, and reproduced the fix/mutation behavior locally (WSL Ubuntu 22.04, meson+ninja, plain gtest and Valgrind memcheck on `unittest_trainer`).

**Summary of what I verified as correct:**
- (a) Joining `dummy_data_thread` before starting a new one, and moving the finalize join above the mutex/cond clears and property frees, is ordered correctly relative to what `fw->destroy()`/`push_data()` touch.
- (b) `gst_tensors_info_free()` is correctly inserted before every place `trainer->prop.input_meta`, `trainer->in_config`, and `trainer->out_config.info` are replaced, and `input_meta` is freed at finalize after `fw->destroy()` runs (matching the commit's stated rationale) — no double-free, since each destination is an independent copy.
- (c) The added `size < gst_tensor_meta_info_get_header_size(meta)` check runs against a freshly `gst_tensor_meta_info_init()`-ed `meta` (so it checks against the build's own V1 header size, 128 bytes) before `gst_tensor_meta_info_parse_header()` reads fixed offsets out of `data`; this is a real, previously-missing bound.
- Locally reverting each fix in turn reproduces exactly what the PR body claims: undoing (a)'s thread-join makes `pauseTwiceJoinsDummyThread` fail deterministically; undoing (b) leaves `unittest_trainer` green under plain gtest but Valgrind reports definitely-lost blocks at `gst_tensor_trainer_sink_event()`; undoing (c)'s size check crashes `flexibleHeaderTruncated_n` with SIGSEGV (exit 139) via the guard-page technique, exactly as designed.
- I additionally confirmed **`finalizeJoinsDummyThread` does now cover the finalize-join reorder**, addressing the earlier bot-review NIT: reverting only the reorder (join placed back below `g_mutex_clear`/`g_cond_clear`) makes this test fail deterministically in a plain (non-Valgrind) run on my machine, in addition to Valgrind reporting an `Invalid read of size 1`. So the follow-up (having `fake_trainer_push_data` read `prop->model_config`) is effective, not just theoretically motivated.
- The real `nntrainer` sub-plugin's `nntrainer_model_destructor()` (its `destroy()` callback) does not dereference `prop->model_config` or `prop->input_meta` at all, so the pre-existing (unchanged-by-this-PR) ordering where `model_config`/`model_save_path`/`model_load_path` are freed before `fw->destroy()` is called stays safe with the real sub-plugin.
- No public headers, API, or ABI are touched (only `gst/nnstreamer/elements/gsttensor_trainer.c` and its test file); no architecture/API documentation update is required.
- CI at review time: `Static checks` (clang-format/gst-indent/doxygen-tag) passed, `Build on Ubuntu with LLVM/Clang`, `Build on MacOS-latest with Meson`, and `Build with meson and test with Valgrind (ubuntu-22.04-arm)` all passed. The `ubuntu-22.04` (non-arm) Valgrind job — the one that actually runs `unittest_trainer` under memcheck per the PR's own notes — was still `in_progress` at the time of this review; worth confirming it turns green before merge, though my local reproduction with the exact PR sources matches the PR's claimed numbers (0 errors, only a 16 KB glib loader-time block).

**Findings**

1. **[nit] PR body: stale/self-contradictory base-commit note.** In the PR description, the "Context for agents" section says `Branch fix/4920-trainer-thread-leak-flex on myungjoo/nnstreamer, base main f865ca74`, but the "Result" section of the same body says `rebased onto main cf36cacd`, and `git merge-base` between this PR's head and the real upstream `main` (`630dad1b`) is in fact `cf36cacd` — confirming the "Result" line is right and the "Context for agents" line is stale (left over from before the rebase mentioned in this round). Purely a documentation accuracy issue, not a code issue. Suggest updating that line to `cf36cacd` (or dropping it) so a future reader isn't misled about which base the noted line numbers/behavior apply to.

2. **[nit] `flexibleHeaderUnknownVersion_n` doesn't actually depend on A7.** Commit `6070b58f`'s message says "The common helper refuses such a header since A7, and this element refuses it as well." I checked this precisely: for the forged version `0xDE002000`, `GST_TENSOR_META_IS_V1()` evaluates to false under *both* the pre-A7 buggy bitwise-AND form and the post-A7 equality form (`(0xDE002000 & 0x00FFF000) = 0x2000`, which doesn't overlap either macro's comparison value either way). That means `gst_tensor_trainer_convert_meta()`'s pre-existing (unchanged-by-this-PR) `header_size == 0` → `goto error` path in `gst_tensor_trainer_push_input()` (`gsttensor_trainer.c`, the `if (header_size == 0) goto error;` check right after the `gst_tensor_trainer_convert_meta()` call) already rejects this exact test input independent of whether A7's new `validate()` block exists. So this test case is valid, harmless, defense-in-depth coverage, but it isn't actually pinned to the A7 change the way the commit message implies, and it doesn't exercise anything new that this PR itself adds (the `size <` check added by this PR isn't the reason this particular case fails either, since the 256-byte buffer is large enough). Not a defect — just slightly imprecise commit-message framing; no action required, but worth knowing that this specific case's real regression protection is the pre-existing `header_size == 0` guard, not A7.

3. **[nit] The new `gst_tensors_info_free(&trainer->out_config.info)` (commit `f90a425a`, `gsttensor_trainer.c` sink_event, CAPS handling) is currently untestable.** The commit message itself says "its copy owns nothing today because the output meta has no names and one tensor," which I confirmed: reverting only this one line locally produces no failure in the plain gtest run and no new Valgrind leak, since `trainer->output_meta` never allocates the extra-tensor block that would need freeing. This is honestly disclosed already, harmless, and matches the surrounding pattern, so it's fine to keep, but there is no regression test today that would catch someone later removing it after `output_meta` grows past the single-tensor/no-name case. Not blocking; just flagging that this specific line has no CI-visible safety net right now (unlike the other frees in the same commit, which the Valgrind gate does cover via `renegotiateManyTensors`).

No blocking or major issues found: the three fixes are each correctly scoped to `gsttensor_trainer.c`/its test file, ordered correctly relative to what the real sub-plugin actually touches, and each comes with a test that I locally confirmed fails without the fix (either deterministically in plain gtest, or via the Valgrind memcheck/leak gate this repo already uses for this binary) and passes with it — satisfying this repo's test-case policy for bug fixes.

**Verdict: mergeable after nits.** (The three points above are documentation/coverage nits, not code defects; I'd also wait for the still-in-progress `ubuntu-22.04` Valgrind CI job to go green before merging, as a matter of process rather than a review finding.)

myungjoo:

All three nits answered; new head `c7f33ffd` (the third commit's message only, no code change since `6070b58f`).

1. **Stale base line in the PR body.** Fixed: the "Context for agents" section now says base `cf36cacd` (rebased; it started on `f865ca74`).
2. **`flexibleHeaderUnknownVersion_n` does not depend on A7.** Correct, and thanks for checking the macro both ways — `0xDE002000 & 0x00FFF000` is `0x2000`, so `GST_TENSOR_META_IS_V1()` is false before and after `304ac60e`, and this element's own `header_size == 0` refusal is what rejects the input either way. The commit message is reworded: the element cannot place the payload, which is what already rejects it, and the common `validate ()` refuses the version as well since A7, so the case fails only if both refusals go away.
3. **The `out_config.info` free has no CI-visible witness today.** Agreed, and it is kept deliberately. `gst_tensors_info_copy()` never frees its destination, so every one of the three copies in this handler needs its free; the output one owns nothing only because `output_meta` is one unnamed tensor today. Leaving the third copy as the odd one out is what would make a later change to the output meta leak silently. It is one line, in the same handler, doing what the other two do.

The `ubuntu-22.04` Valgrind job was green on the pre-rebase head and locally on the rebased sources (0 errors, 0 definite leaks of ours; only a 16 KB glib loader-time block); it is re-running on this head along with the rest of CI.

myungjoo:

This reviews head `c7f33ffd`. Since the previous round (`6070b58f`), `git diff 6070b58f c7f33ffd` is empty except for the third commit's message and the PR body — no code change. I re-checked round 5's three nits for accuracy, then re-read the PR as a whole independently: the full diff, `gst/nnstreamer/elements/gsttensor_trainer.c` in full, the relevant `gst_tensor_meta_info_*`/`gst_tensors_info_*`/`gst_tensors_config_*` implementations in `gst/nnstreamer/nnstreamer_plugin_api_util_impl.c`, and the test file in full.

## Verifying round 5's three answers

1. **Stale base-commit line.** Confirmed fixed. The PR body's "Context for agents" section now reads "base main `cf36cacd` (rebased; it started on `f865ca74`)", matching the "Result" section and this repo's actual `main` state. Accurate.
2. **`flexibleHeaderUnknownVersion_n` / A7 framing in the third commit's message.** Confirmed accurate. `git show c7f33ffd` now reads: "Such a header is refused twice over: this element cannot place the payload, which is what already rejects it, and the common validate () refuses the version as well since A7." I checked `GST_TENSOR_META_IS_V1` (`nnstreamer_plugin_api_util_impl.c:1446`) against the forged version `0xDE002000` by hand: `0xDE002000 & 0x00FFF000 = 0x2000 ≠ 0x001000`, so `IS_V1` is false and `gst_tensor_meta_info_get_header_size()` returns 0 regardless of A7 — the wording no longer implies this test is pinned to A7, it correctly attributes the primary rejection to this element's own `header_size == 0` guard. Accurate.
3. **Defensive `gst_tensors_info_free(&trainer->out_config.info)` kept without a test.** Confirmed still present at `gsttensor_trainer.c:1024` (in `gst_tensor_trainer_sink_event()`, right before `gst_tensors_info_copy(&trainer->out_config.info, &trainer->output_meta)`), and still untestable today for the stated reason: `trainer->output_meta` is fixed at one unnamed tensor (`gst_tensor_trainer_set_output_meta()`), so it never allocates the `info->extra` block that `gst_tensors_info_free()` would need to release, and reverting only this one line produces no failure in plain gtest and no new Valgrind leak. This is honestly disclosed and harmless defense-in-depth; no new information found that would change this.

No earlier round's finding is still open: round 1's leftover-mmap nit, round 2's redundant double-map nit, round 3's implicit `header_size == 0` guard nit, and the bot review's untested finalize-join-reorder NIT were each fixed in a later commit revision, and I re-confirmed the fixes are present at this head (the guard-page `munmap()` on both early-return paths in `flexibleHeaderTruncated_n`, the single-map `void *data, gsize size` signature of `gst_tensor_trainer_convert_meta()`, the explicit-enough `size <` check ahead of `parse_header()`, and `finalizeJoinsDummyThread` reading `prop->model_config` from inside `push_data()` during finalize).

## Independent re-check of the fix itself

- **(a) thread leak / late-join UAF window**: `gst_tensor_trainer_change_state()` (`:558-559`) joins the previous `dummy_data_thread` before overwriting the handle on a second `PLAYING_TO_PAUSED`; `gst_tensor_trainer_finalize()` joins first (`:320-323`), ahead of the `g_free()`s of `fw_name`/`model_config`/`model_save_path`/`model_load_path` (`:326-329`), the mutex/cond clears (`:334-337`), and `fw->destroy()` (`:339-341`). This ordering is correct given the dummy thread calls into `fw->push_data()`/`fw->stop()`, which read `trainer->prop.*` and can call back into `nnstreamer_trainer_notify_event()`.
- **(b) `input_meta`/`in_config`/`out_config.info` leak**: traced `gst_tensors_info_copy()` (`nnstreamer_plugin_api_util_impl.c:530-548`) — it only `gst_tensors_info_init()`s the destination and never frees a pre-existing one, and `gst_tensors_info_get_nth_info()` (`:324-343`) lazily allocates `info->extra` once `index >= NNS_TENSOR_MEMORY_MAX` (16). So every caps renegotiation past 16 tensors leaked the previous `extra` block before this fix. The three added `gst_tensors_info_free()`/`gst_tensors_config_free()` calls in `gst_tensor_trainer_sink_event()` (`:1017`, `:1021`, `:1024`) each precede the replacement they guard, and `trainer->in_config = config;` (`:1020`) is a clean ownership transfer of a local stack struct — no double free, since `trainer->prop.input_meta` is an independent deep copy via `gst_tensors_info_copy()`, not an alias of `config.info`.
- **(c) OOB header read**: `gst_tensor_trainer_convert_meta()` (`:669-687`) calls `gst_tensor_meta_info_init(meta)` first, which sets `meta->version` to this build's current V1 version, so `gst_tensor_meta_info_get_header_size(meta)` at that point deterministically returns 128 — checking the caller-supplied `size` against a fixed, build-known bound, not anything attacker-controlled — before `gst_tensor_meta_info_parse_header(meta, data)` touches the bytes. Cross-checked `parse_header()`'s actual read range (`nnstreamer_plugin_api_util_impl.c:1620-1645`): the widest field read is `val[21]` (byte offset 84), safely inside the 128-byte bound the new check enforces.

None found. `gst_tensor_trainer_convert_meta()` stays `static`, single call site (confirmed by grep), so its signature change is invisible outside this file. No header, public API, or other element is touched, and `Documentation/component-description.md` (the only doc naming `tensor_trainer`) needs no update, matching all five prior rounds.

## Code size / scope

Proportionate: 3 commits, one per C31 sub-defect, touching only `gst/nnstreamer/elements/gsttensor_trainer.c` and its own test file.

## Test coverage and CI enforcement

- (a): `pauseTwiceJoinsDummyThread` / `finalizeJoinsDummyThread` assert exact `stop_calls`/`push_done` counts against a fake `nntrainer` sub-plugin registered only in this test binary (`tests/meson.build` unmodified by this PR); both fail under a plain, always-on `meson test` run without the fix.
- (b): `renegotiateManyTensors` pushes 17- then 18-tensor caps (`NNS_TENSOR_MEMORY_MAX` = 16, `NNS_TENSOR_SIZE_LIMIT` = 256, so both counts are valid and both force the `info->extra` path) through a `GstHarness`. The test can only observe that renegotiation still succeeds and the second caps' framerate reaches the src pad — the leak itself is caught only by the Valgrind job's leak gate, which prior rounds confirmed is a required status check on `main` covering `unittest_trainer`.
- (c): `flexibleInput` (positive), `flexibleHeaderTruncated_n` / `flexibleHeaderInvalid_n` / `flexibleHeaderUnknownVersion_n` (negative, `_n` suffix per convention) cover a valid header, a too-short memory via the guard-page technique, a long-enough-but-invalid header, and a validating-but-unsizable version — both positive and negative cases per this repo's test-case policy.
- All 18 new/existing `TEST`/`TEST_F` cases in the file carry a `@brief` Doxygen comment, satisfying this repo's per-function comment requirement.

## Documentation

No architecture or public API change; no documentation update needed or expected, consistent with all five prior rounds.

## Findings

No blocking, major, minor, or nit issues found in this round. Round 5's three points were process/documentation nits, not code defects, and all three are answered accurately as verified above.

## Verdict

**Overall verdict: mergeable as-is.**

myungjoo:
Approved by Astra

myungjoo:

Thanks for the approval. The Low NIT is closed rather than left: new head `2c052a36`, folded into the first commit (no fix-up commit; `gsttensor_trainer.c` is byte-identical to `c7f33ffd`).

The finalize witness no longer compares a single byte. The test copies the expected path into the fake sub-plugin before the run, and `push_data()` compares the whole string itself, byte by byte over the expected length — so it reads only inside the block the property owns, even when that block has been freed, and a wrong byte anywhere fails the case. Measured with the join moved back below the frees: the plain run now fails outright (`config_matched` false), and memcheck reports `Invalid read of size 1` at `fake_trainer_push_data` under `gst_tensor_trainer_dummy_data_generation_func`, so both legs catch the regression instead of only the Valgrind one. On the fixed order: 7/7 cases pass with 0 errors and 0 definite leaks of ours; clang-format 16 and the strict `-Wextra -Wsign-compare -Werror` build are clean.

On the out-of-scope observation: agreed, `gst_tensor_trainer_push_input()`'s cleanup loop unmaps the memory whose `gst_memory_map()` just failed, with a `GstMapInfo` that was never filled in. It is on main unchanged and outside C31, so it is not folded in here; it is being reported for its own #4920 item.

Back to draft with the hold label while CI runs on this head.

myungjoo:

This reviews head `2c052a36` (3 commits on `main@cf36cacd`: `9456a179` join ordering, `991f6df6` free-before-replace on tensor info, `2c052a36` flexible-header size guard).

## Focus: the change since the bot's approval (`c7f33ffd` → `2c052a36`, test file only)

The only change is in `finalizeJoinsDummyThread`'s witness: `fake_trainer_push_data()` now compares the whole `model_config` string (up to `strlen(expect_config)`, copied via `g_strlcpy` into a 512-byte `fake_stat.expect_config`, guarded by `ASSERT_LT(strlen(config_path), sizeof(expect_config))`) instead of one byte.

I built this PR's tree independently (private copy, `ninja` incremental build, `GST_REGISTRY`/`GST_PLUGIN_PATH` pointed at that build) and ran `tests/unittest_trainer --gtest_filter='TensorTrainerFakeFw.*'` directly — no Valgrind, no `MALLOC_PERTURB_`:

- With the current code (join before the frees): 5/5 runs, all 7 tests pass, including `finalizeJoinsDummyThread`.
- With the finalize join mechanically moved back below the frees (the pre-fix order): 5/5 runs, `finalizeJoinsDummyThread` fails with `Value of: fake_stat.config_matched / Actual: false / Expected: true`; the other 6 tests still pass.

So, for this specific finding:
- **Bounds**: safe. The comparison loop is bounded by `expect_config`, a fixed 512-byte buffer that `g_strlcpy` always NUL-terminates and whose content length is asserted to fit before truncation. It never iterates past `strlen(expect_config)` bytes of `prop->model_config`, regardless of what that pointer refers to.
- **What it reads once freed**: by design this is a use-after-free read when the mutation is present (the comment above it says as much: "the properties must outlive this call, even when it runs during finalize"). `g_free()`'s underlying chunk stays mapped (glibc doesn't unmap small heap chunks), so this doesn't crash; it reads adjacent/stale heap content, which is exactly what both the GTest assertion and Valgrind's memcheck are positioned to catch.
- **False pass / false failure**: not observed in 5/5 runs each way on this build. The correctly-ordered path can't produce a false failure — the join happens-before the frees unconditionally, so there's no timing window at all in that direction. The mutated path detecting the bug does depend on the freed `model_config` chunk having been overwritten by one of finalize's own subsequent allocations before the thread wakes 500ms later (glibc heap reuse), which is an allocator/timing property, not a language guarantee — so this GTest assertion alone is not an unconditionally reliable backstop across every allocator/platform. That said, this matches the commit message's own framing ("memcheck reports an invalid read of ours and the Valgrind job fails on it") — the project's Valgrind CI job (already gating on memcheck errors per the existing valgrind-gate infrastructure) is the authoritative backstop regardless of whether the plain GTest assertion happens to also catch it on a given machine. **Severity: nit / informational** — no change requested; this is inherent to any content-comparison UAF witness without ASan/Valgrind, and the real safety net (Valgrind gate) is independently confirmed to catch this exact mutation (`Invalid read of size 1` in `fake_trainer_push_data`, per this PR's own commit history of iterating on this test).
- **Thread-safety**: sound. `expect_config` is written by the main thread before `g_thread_new()` (creation establishes happens-before for the new thread), and `config_matched` is only read by the main thread after `gst_element_set_state(..., GST_STATE_NULL)` has driven `finalize()`'s `g_thread_join()` (join establishes happens-before back to the joining thread). No concurrent read/write of either field. `stop_calls`/`push_calls`/`push_done`, which genuinely are polled concurrently, correctly use `g_atomic_int_*`.

Net: the string-comparison change is a strict improvement over the single-byte check (harder to satisfy by coincidence) and is bounds-safe; it does not introduce a new false-pass or false-failure risk beyond what any such witness has.

## Independent review of the PR as a whole

- `9456a179` ([Trainer] Join the dummy-data thread before starting another one): Moves the `g_thread_join()` in `gst_tensor_trainer_finalize()` (gst/nnstreamer/elements/gsttensor_trainer.c:321-324) to run before the property frees and mutex/cond clears, and adds a join-before-replace in `gst_tensor_trainer_change_state()` (gsttensor_trainer.c:558-559) when a second `PLAYING_TO_PAUSED` transition would otherwise overwrite `dummy_data_thread` while the first thread is still running. Verified no other file references `dummy_data_thread` or the dummy-data thread func, so the fix is fully contained to this element. `trainer->dummy_data_thread` is implicitly NULL on a fresh instance via GObject's zeroed instance allocation, so the new `if (trainer->dummy_data_thread)` guards are safe on first use.
- `991f6df6` ([Trainer] Free the tensors info before replacing it): Adds `gst_tensors_info_free()`/`gst_tensors_config_free()` calls before each `gst_tensors_info_copy()`/struct overwrite in `gst_tensor_trainer_sink_event()` (gsttensor_trainer.c:1017-1023, 1028-1030) and a final `gst_tensors_info_free (&trainer->prop.input_meta)` in finalize (line 330). Confirmed `config` (from `gst_tensors_config_from_caps()`) is a fresh local, independent of `trainer->in_config`, so freeing the old `in_config` before the struct assignment is not a double-free and doesn't drop a still-live pointer.
- `2c052a36` ([Trainer] Check the memory size before parsing a flexible header): Adds `gst_tensor_meta_info_init(meta)` plus a `size < gst_tensor_meta_info_get_header_size(meta)` guard in `gst_tensor_trainer_convert_meta()` (gsttensor_trainer.c:669-676) before `gst_tensor_meta_info_parse_header()`, and threads the mapped memory's size through from `gst_tensor_trainer_push_input()`. This is byte-for-byte the same idiom already used in `gst_tensordec_check_input_size()` (gst/nnstreamer/elements/gsttensor_decoder.c:744-746), `gsttensor_filter.c:699-700`, `gsttensor_crop.c`, `gsttensor_sparseutil.c`, and `tensor_converter_util.c` — so it's consistent with established convention, not a novel pattern. The added `gst_tensor_meta_info_init()` matters because `meta` is an uninitialized stack element (`in_meta[i]`) in the caller; without it, `gst_tensor_meta_info_get_header_size()` would read whatever garbage version bits happened to be on the stack before parsing.
- **Scope**: diff touches exactly `gst/nnstreamer/elements/gsttensor_trainer.c` and `tests/nnstreamer_trainer/unittest_trainer.cc`. No other module is touched, and grepping the tree confirms nothing else references the changed symbols/fields. Proportionate to the three C31 sub-items being fixed.
- **Test coverage**: `pauseTwiceJoinsDummyThread` and `finalizeJoinsDummyThread` cover (a); `renegotiateManyTensors` (17→18 tensors, crossing the 16-tensor inline threshold) covers (b) — as a GTest it can't observe the leak itself, but it is deliberately built to be exercised by the Valgrind CI job, which is where the regression would actually be caught; `flexibleInput`/`flexibleHeaderTruncated_n`/`flexibleHeaderInvalid_n`/`flexibleHeaderUnknownVersion_n` cover (c) with both a positive case and three negative cases (`_n` suffix, matching the SSAT/GTest negative-case convention). This satisfies the repo's test-case policy for a bug fix (regression tests reproducing each of the three bugs).
- **CI detection**: For (b) and the finalize-order half of (a), detection routes through the Valgrind memcheck job rather than a GTest assertion; `unittest_trainer` is a standard GTest binary under `tests/`, which the project's `packaging/run_unittests_binaries.sh` recursive discovery picks up for that job, so this is wired into CI rather than being a manual-only check.
- **Architecture/API/docs**: no public API, GObject property, or header signature change (`gst_tensor_trainer_convert_meta` is `static`); no ARCHITECTURE/API documentation update is needed.
- **Commit hygiene**: all three commits carry `Signed-off-by:` DCO lines, follow the `[Component] Summary` subject convention, and every new/changed function (including the new static fake-sub-plugin callbacks and test cases) has a Doxygen comment.
- **Cross-checked against the real nntrainer sub-plugin** (`nnstreamer/tensor_trainer/tensor_trainer_nntrainer.cc`): it caches `prop->model_config` into a `std::string` once at model-creation time rather than re-reading it on every `push_data()` call, so the fake test sub-plugin's per-call re-read of `model_config` is a test-only convenience witness, not a reproduction of the real plugin's exact access pattern. That doesn't weaken the fix's motivation — the dummy-data thread can still touch other trainer-owned state (`prop.input_meta`, the notifier signaled via `nnstreamer_trainer_notify_event()`, the mutexes/conds) after finalize starts tearing the element down if the join isn't ordered first — but it's worth noting for anyone reading the test in isolation.

## Verdict

mergeable as-is

PR #4998, opened 2026-09-17
[Filter/TF-Lite] Destroy the interpreter before its delegate and model, and free a repeated delegate path (#4920 F2, F3)

myungjoo:
item **F2**, **F3** of #4920

- **What it fixes:** (F2) `TFLiteInterpreter` destroyed its delegate and unmapped its `FlatBufferModel` *before* destroying the `tflite::Interpreter` that uses both, because C++ destroys members in reverse declaration order. This happened on every close, reload and refused reload. `loadModel()` also freed the old model before resetting the old interpreter, which is what happens when a second filter opens with the same `shared-tensor-filter-key`. (F3) `custom=ExtDelegateLib:a,ExtDelegateLib:b` leaked the first path on every open.
- **How:** declare `delegate_ptr` and `model` before `interpreter`. In `loadModel()`, build the new model into a local, reset the interpreter, then take the model; a model that fails to build still leaves the previous state untouched. `g_free` the previous `ext_delegate_path` before storing the next one. One commit per item; `tensor_filter_tensorflow_lite.cc` changes by +7/-4 lines.
- **Result:** the 6 new gtests fail on main and pass with the fix. Mutation matrix: main's source fails all 6; each wrong member order (delegate first, model first, main's order) fails the 5 lifetime cases; putting the model swap back before the interpreter reset fails only `sharedKeyReopen`; dropping the `g_free` fails only the leak case. Locally (Ubuntu 22.04 + tflite 2.15 from the PPA, `-Dwerror=true`): `unittest_filter_tensorflow2_lite` 13/13, `unittest_filter_shared_model` 6/6, SSAT `nnstreamer_filter_tensorflow2_lite` 37/37, `nnstreamer_filter_reload` 2/2. The F2 commit alone builds and passes 11/11.

**Coverage of the functions this patch touches** (gcov, Ubuntu 22.04 + tflite 2.15, the tflite and shared-model gtests plus the 37-case tflite and 2-case reload SSAT suites): **164/178 lines = 92.1%**, functions 4/4, up from 119/178 = 66.9% before the option-parsing cases were added; `tflite_parseCustomOption()` is at 100%. The 14 lines left are the NNAPI and QNN delegate bodies and two failure paths that need a corrupt model.

### How the tests observe it

There is no deterministic witness with the real delegates in CI. Whether a use-after-free actually reads freed memory depends on each delegate's kernel teardown, and neither Ubuntu job that runs valgrind builds this sub-plugin. So the tests add a **test-only external delegate** (`tests/nnstreamer_filter_tensorflow2_lite/tflite_lifetime_delegate.cc`, loaded through the existing `Delegate:External,ExtDelegateLib:` option). It takes over one node that reads a constant tensor of the model. When the interpreter frees that node, it counts a violation if its delegate has already been destroyed, or if the model page is no longer mapped (`mincore()` → `ENOMEM`). This is exactly TensorFlow Lite's lifetime contract, checked without crashing: the test keeps its own `dlopen()` reference so the kernel code stays mapped. It also reports how many key values it was handed, how many delegates of its own are still alive (`live`), and can be told to refuse to be applied (`ExtDelegateKeyVal:mode#fail`), which is how the option parser and the delegate-application error path are covered.

| Case | Path |
|---|---|
| `openClose` | close → `~TFLiteInterpreter`; the delegate is alive while the filter is open and gone after the close |
| `reloadModel` | non-shared reload deletes the old interpreter |
| `reloadUnmatchedModel_n` | refused reload deletes the unused interpreter |
| `sharedKeyReopen` | 2nd `tensor_filter` with the same key → `loadModel()` on the shared interpreter (needs real elements: only the element creates the shared model table) |
| `extDelegateLibGivenTwice` | invalid path first, test delegate second → the last entry is still the one loaded |
| `extDelegateKeyValOptions` | `ExtDelegateKeyVal` entries reach the delegate: repeated entries accumulate, a malformed one is dropped |
| `extDelegateRefusesToApply_n` | the delegate refuses to be applied → the open fails and the delegate is still destroyed (`live == 0`) |
| `extDelegateWithoutLib` | `Delegate:External` with no library falls back to no delegate |
| `customOptionsParsed`, `customOptionsDelegateNames_n`, `openWithoutModelFile_n` | the rest of the option parser: `NumThreads`, each delegate name, the QNN backend and performance modes, an unknown option, no model file. The delegate-name cases use a model that does not exist, so no accelerator is ever built |
| `extDelegateLibGivenTwiceNoLeak_n` | 10 failing opens with two 64 KiB `ExtDelegateLib` entries; heap in use (glibc `mallinfo2`) must grow by less than one path. On main it grows by 10 paths. Skipped without glibc ≥ 2.33. |

The 5 lifetime cases are skipped, not failed, when the sub-plugin is built without the external delegate (`TFLITE_EXTERNAL_DELEGATE_SUPPORTED`); no CI configuration is in that state today.

**Where CI gates it:** `unittest_filter_tensorflow2_lite` runs in the Tizen GBS (`%check`) and pdebuild jobs. Both detect `external delegate ... links: YES` in their current logs, so all these cases run there. The Ubuntu meson job only compiles them, in its tflite example-app rebuild.

- Branch `fix/4920-tflite-member-order` on the fork, base `f865ca74`. Commits: `a80d111e` (F2 + lifetime delegate + 4 tests + `tests/meson.build` hunk inside the tflite2 block), `b3c55d18` (F3 + 2 tests).
- The helper library is `libnnstreamer_unittest_tflite_lifetime_delegate.so`, installed next to the unittests (`unittest-nnstreamer/tests`, already covered by the spec's directory entry). The test finds it via `NNSTREAMER_BUILD_ROOT_PATH/tests`, falling back to the directory of `/proc/self/exe`.
- A kernel is tied to the most recently created test delegate (`last_delegate_id`), because TFLite's external delegate wrapper may hand `Prepare` its own wrapper object rather than the plugin's delegate.
- The `model_checks >= 1` assertion guards against a TFLite build that copies the model instead of mapping it, which would make the model check vacuous.
- Local reproduction: WSL docker container `nns-f2f3-4920` (image `nns-g16:deps` + `ppa:nnstreamer` `tensorflow2-lite-dev` 2.15, `python3-numpy` for SSAT case 6), tree `/src`, scripts under `~/nns-4920-f2f3/scratch`.
</details>


myungjoo:

**Verdict: mergeable as-is.**

I reviewed this against issue #4920 items F2/F3, the surrounding tensorflow2-lite sub-plugin code, the meson/packaging wiring, and CI logs. I rebuilt and ran the affected suites myself (not just re-reading the PR body's numbers):

- `unittest_filter_tensorflow2_lite`: 13/13 pass (7 original + 6 new).
- `unittest_filter_shared_model`: 6/6 pass.
- SSAT `nnstreamer_filter_tensorflow2_lite`: 37/37 pass; SSAT `nnstreamer_filter_reload`: 2/2 pass.
- Confirmed clang-format is a no-op on all three touched/added files.
- Confirmed via the real CI logs for this PR (`Ubuntu pdebuild on resolute (amd64)`, run 35193047290, job 105109738907) that `external delegate ... links: YES` and all 6 new `nnstreamerFilterTensorFlow2LiteLifetime.*` cases actually ran and passed against the real TF Lite 2.15 package, not just a local sandbox build.
- Confirmed from a different, already-merged PR's Tizen GBS x86_64 (`unit_test 1`) run that `unittest_filter_tensorflow2_lite` does execute under `%check` and that `external delegate` also links there, supporting the PR body's claim that both gating jobs (Tizen GBS x86_64, Ubuntu pdebuild) actually run these cases. (This PR's own GBS x86_64 job was still queued/pending at review time behind a backlog of sibling #4920 PRs; nothing in the log evidence suggests it would behave differently.)

**Scope / correctness of the fix**

- F2: the reorder (`delegate_ptr`, `model`, `interpreter` — interpreter declared last so it is destroyed first) is the standard, correct C++ fix for the reverse-declaration-order destruction bug; verified by reading the destructor and by reverting the mutation mentally against the mutation matrix described in the PR body (I was blocked from mutating the shared container's copy of the source to empirically re-derive the matrix myself, but the logic and the passing/failing test shape both check out).
- The `loadModel()` change (build into a local `new_model`, reset `interpreter` first, then `model = std::move(new_model)`) correctly fixes the "old model freed before old interpreter is reset" bug, and additionally makes a failed `BuildFromFile()` leave the previous `model`/`interpreter` state untouched (previously, a failed rebuild would still have clobbered `model`). I traced the one reachable case for this reset-on-existing-object path: `TFLiteCore::checkSharedInterpreter()` attaches a second filter to an already-loaded shared `TFLiteInterpreter`, and `TFLiteCore::init()` unconditionally calls `loadModel()` again on it — this is exactly what `sharedKeyReopen` exercises, and it is real, reachable production behavior, not a test-only artifact.
- F3: `g_free` before overwrite in `tflite_parseCustomOption()`'s `ExtDelegateLib` branch is correct; `option->ext_delegate_path` is zero-initialized right before the parse loop and there is only one call site (`tflite_loadModelFile`), so there's no double-free risk.

**Regression risk**

- Change is local to `TFLiteInterpreter`'s member layout and `loadModel()`/`tflite_parseCustomOption()`; I did not find any other file depending on the declaration order of `delegate_ptr`/`model`/`interpreter`, and the reorder cannot change any externally observable behavior other than destruction order.
- `libdl_dep` and `tflite2_extra_dep` are legitimately new dependencies for the test targets: `nnstreamer_unittest_deps` does not already carry `libdl_dep` (needed for the new `dlopen`/`dlclose`/`dlsym` calls in the test), and `tflite2_extra_dep` is what supplies `-DUSE_TENSORFLOW2_HEADER_PATH` etc. to the new delegate `.so`. `subdir('ext')` runs before `subdir('tests')` in the root `meson.build`, so both are in scope when `tests/meson.build` needs them.
- Checked `packaging/run_unittests_binaries.sh`: it only executes files matching `unittest_*`, so the new `libnnstreamer_tflite_lifetime_delegate.so` (a dlopened helper, not a gtest binary) is never picked up and run as if it were a test — this was a real regression class for a previous, similar dlopened-helper-in-tests-dir situation (nnstreamer-edge custom lib, PR #4921), and this PR doesn't reintroduce it.
- The `%files unittests` directory glob (`.../unittest-nnstreamer/tests`) already covers the new `.so`, so no "unpackaged file" RPM error is expected; the `dh_missing` warnings for the new files in the pdebuild log are pre-existing, harmless, non-fatal noise shared by essentially every `.cc`/model file in `tests/` (268 such warnings in that log, unrelated to this PR).
- No other module is touched; the diff is `ext/.../tensor_filter_tensorflow_lite.cc` (+7/-4, matching the PR description) plus test-only additions. Proportionate to the two defects being fixed — the size of the new test file is explained by the fact that neither Ubuntu valgrind job builds this sub-plugin, so a deterministic, crash-free witness for a delegate/model lifetime violation has to be built by hand (a fake external delegate + `mincore()` check) rather than relied on from an existing gate.

**Test completeness**

- Both listed defects (F2, F3) get positive and negative coverage: `openClose`, `reloadModel`, `sharedKeyReopen` (positive lifetime ordering), `reloadUnmatchedModel_n` (negative/refused-reload path), `extDelegateLibGivenTwice` (last-one-wins) and `extDelegateLibGivenTwiceNoLeak_n` (regression test for the leak, gated behind `HAVE_MALLINFO2`/glibc >= 2.33 with a clean `GTEST_SKIP` otherwise).
- Future-regression detection: since these are ordinary gtest cases wired into `meson test`/`%check`/pdebuild, any later change to `TFLiteInterpreter`'s member order, `loadModel()`, or the `ExtDelegateLib` parsing that reintroduces either bug will fail these specific cases, and CI (Tizen GBS x86_64 `unit_test 1`, Ubuntu pdebuild) will turn red on it — confirmed these jobs actually execute this binary. The Ubuntu meson/valgrind jobs still only compile this sub-plugin (via the `tflite2` template example rebuild), not execute these cases, exactly as the PR body states; that's a pre-existing CI-topology fact, not something this PR needed to fix.

**Documentation**

- No public API or architecture change; nothing under `Documentation/` needed updating, and none was expected.

**Minor / non-blocking observations (Nit)**

1. Nit — `tests/meson.build`: the new helper library is named `nnstreamer_tflite_lifetime_delegate` / `libnnstreamer_tflite_lifetime_delegate.so`, which reads like an installable `libnnstreamer_*` production artifact at a glance even though it's a test-only helper under `unittest_install_dir`. Not a functional problem (it's correctly excluded from `run_unittests_binaries.sh` by the `unittest_*` name filter either way), just a naming/readability nit.
2. Nit — `TFLiteCore::init()` unconditionally re-runs `loadModel()` even when `checkSharedInterpreter()` attached to an already-loaded shared interpreter (the path `sharedKeyReopen` exercises). That reload-on-attach is pre-existing behavior this PR does not change or introduce, and F1 (tracked separately, not in this PR's scope) is the closer match for anything about correctness of shared-model reload semantics — flagging only so it isn't mistaken for a side effect of this diff.

myungjoo:

Thanks for the review. Addressed in-PR:

- **Nit 1 (helper name reads like a production library):** renamed to `libnnstreamer_unittest_tflite_lifetime_delegate.so`, matching the in-tree `libnnstreamer_unittest_util.so`. `packaging/run_unittests_binaries.sh` selects `-name "unittest_*"`, so the `lib...` name is still never picked up as a test binary. The rename is folded into the F2 commit (no fix-up commit); `git diff 6b649cf0 e51034d2` is exactly the 2-line rename. Local: `unittest_filter_tensorflow2_lite` 13/13.
- **Nit 2 (`TFLiteCore::init()` re-runs `loadModel()` on an attached shared interpreter):** agreed it is pre-existing and unchanged. It is the path `sharedKeyReopen` pins, so no code change.

myungjoo:

This is re-review round 2 of head `e51034d2` (after the round-1 "mergeable as-is" review and the author's two follow-up changes: the helper-library rename for Nit 1, and the reply on Nit 2). I re-read the diff against `f865ca74` from scratch, re-derived the F2/F3 defects from issue #4920 independently, rebuilt this exact head and re-ran the affected suites myself, and specifically re-checked the two things the round-1 review could not have seen (the effect of the post-review rename, and the author's answer to Nit 2), rather than taking the prior verdict on faith.

**Verdict: not mergeable as-is — one small, trivial-to-fix Blocking issue; otherwise correct and well-tested.**

**Confirmed correct (independently re-derived, not just re-read):**
- F2: reordering `delegate_ptr`/`model` before `interpreter` in `TFLiteInterpreter` (`tensor_filter_tensorflow_lite.cc:258-261`) is the right fix — C++ destroys members in reverse declaration order, so `interpreter` (declared last) is now destroyed first, before the delegate/model it depends on.
- The `loadModel()` change (`:465-483`): building into a local `new_model`, resetting `interpreter` to `nullptr` first, and only then `model = std::move (new_model)`, correctly avoids freeing the live model out from under the still-alive-but-about-to-be-replaced interpreter, and leaves the previous `model`/`interpreter` untouched if `BuildFromFile()` fails. I traced the reachable production trigger for the "second `tensor_filter` attaches to an already-shared interpreter" case myself: `TFLiteCore::checkSharedInterpreter()` (`:1030-1075`) attaches to the existing shared `TFLiteInterpreter`, and `TFLiteCore::init()` (`:1114-1130`) then unconditionally calls `loadModel()` on it again — this is exactly what the new `sharedKeyReopen` test exercises, and it is real, reachable behavior, not test-only scaffolding.
- F3: `g_free ((gpointer) option->ext_delegate_path)` before `g_strdup` in `tflite_parseCustomOption()` (`:1458-1459`). Checked `option->ext_delegate_path` is zero-initialized once before the parse loop (`:1423`) and consumed/freed exactly once afterward at the single call site (`tflite_loadModelFile`, `:1584-1585`); no double-free risk.
- Regression risk: the change is local to `TFLiteInterpreter`'s member layout, `loadModel()`, and one `ExtDelegateLib` branch; the reorder cannot change any externally observable behavior other than destruction order. No other module is touched.
- Test completeness / future-regression detection: both defects get positive and negative coverage (`openClose`, `reloadModel`, `sharedKeyReopen`, `reloadUnmatchedModel_n`, `extDelegateLibGivenTwice`, `extDelegateLibGivenTwiceNoLeak_n`), wired into ordinary `meson test`/GBS `%check`/pdebuild, so a later regression in member order, `loadModel()`, or `ExtDelegateLib` parsing will turn these gating CI jobs red.
- I rebuilt this exact head (`e51034d2`) in a container with real TF Lite 2.15 and ran the full file: all 13 cases in `unittest_filter_tensorflow2_lite` pass, including the 6 new `nnstreamerFilterTensorFlow2LiteLifetime.*`/`extDelegateLibGivenTwiceNoLeak_n` cases.
- Nit 1 answer verified: the helper is now `libnnstreamer_unittest_tflite_lifetime_delegate.so`; `packaging/run_unittests_binaries.sh` filters on `-name "unittest_*"` (basename match), so the `lib`-prefixed `.so` is still never picked up as a test binary to execute on its own. Confirmed by reading the script directly.
- Nit 2 answer verified: `TFLiteCore::init()` does unconditionally re-run `loadModel()` on an attached shared interpreter (`:1126`, reached via `:987`/`:1031`), pre-existing and unchanged by this PR; agree it's correctly left alone and tracked separately under F1.
- Documentation: no public API/architecture change; nothing under `Documentation/` needed and none was touched.

**Blocking**
- `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc:435`
  ```
  #define LIFETIME_DELEGATE_LIB "libnnstreamer_unittest_tflite_lifetime_delegate.so"
  ```
  This line is 82 columns; the repo's `.clang-format` sets `ColumnLimit: 80`. Running the exact command `.github/workflows/static.check.yml`'s "Static checks" job uses (`clang-format -i` on every changed `.cc`/`.hh`/`.hpp`/`.cpp` file, then `git diff` must be empty) against this file reproduces a real, non-empty diff:
  ```
  -#define LIFETIME_DELEGATE_LIB "libnnstreamer_unittest_tflite_lifetime_delegate.so"
  +#define LIFETIME_DELEGATE_LIB \
  +  "libnnstreamer_unittest_tflite_lifetime_delegate.so"
  ```
  This line did not exist before this PR and was not present at the shorter length the round-1 review checked ("confirmed clang-format is a no-op on all three touched/added files") — the post-review rename from `libnnstreamer_tflite_lifetime_delegate.so` to `libnnstreamer_unittest_tflite_lifetime_delegate.so` (done to address round-1 Nit 1) pushed this specific line 8 columns over the limit and is what introduces the violation. It will fail the "Static checks" job (and likely the redundant `CPP-Linter` clang-format check) exactly as written, blocking merge. `unittest_filter_tensorflow2_lite.cc` was the only file among the four changed in this PR with a violation; `tensor_filter_tensorflow_lite.cc` and `tflite_lifetime_delegate.cc` are clean.
  Suggested fix: apply clang-format's own answer above (backslash-continue the macro), or shorten the identifier so it stays under 80 columns. Trivial, no functional change.

No other Blocking/Major issues found. No Minor findings beyond the one above.

myungjoo:

Round 2 Blocking item confirmed and fixed. The rename pushed `#define LIFETIME_DELEGATE_LIB` to 82 columns, and clang-format 16 (the local run of the same check) produced exactly the diff quoted in the review. The fix applies clang-format's own output and is folded into the F2 commit. New head `1dea1fbe`; `git diff e51034d2 1dea1fbe` is that one line split in two.

Re-run before this push: clang-format 16 is a no-op on all three `.cc` files, `doxygen-tag.sh` passes (its negative control, a removed `@brief`, still errors), the `-Dwerror=true` build passes, and `unittest_filter_tensorflow2_lite` passes 13/13.

myungjoo:

This is re-review round 3 of head `1dea1fbe` (after round 1's "mergeable as-is" verdict, round 2's Blocking clang-format finding, and the author's fix for it). I re-read the full diff against `f865ca74` and issue #4920 F2/F3 from scratch, re-derived the defects independently, and specifically re-verified round 2's fix with the actual checker CI uses rather than a proxy, plus did new verification neither prior round performed (a real mutation test). Treating no earlier round as complete.

**Verdict: mergeable as-is. No Blocking / Major / Minor issues found in this round.**

**What I verified beyond the previous two rounds**

1. Static checks — ran the exact `static.check.yml` "Static checks" job logic against the exact changed-file list (`ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc`, `tests/meson.build`, `tests/nnstreamer_filter_tensorflow2_lite/tflite_lifetime_delegate.cc`, `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc`):
   - `clang-format`: round 2 caught a real 80-column violation using clang-format 16 as a local proxy. I pulled the *actual* version CI installs — confirmed from a recent successful "Static checks" run's own log (`sudo apt-get install clang-format` on `ubuntu-24.04`/noble → `clang-format amd64 1:18.0-59~exp2`) — installed clang-format **18.1.8** in an isolated venv, and ran `clang-format -i` on all three changed/added `.cc` files exactly as the workflow step does. Diff against the originals is empty for all three files. This closes the residual doubt that a version-16-vs-CI's-actual-version mismatch could hide.
   - `doxygen-tag.sh`, `newline.sh`, `cppcheck.sh` (level 0), `hardcoded-path.sh`, `misspelling.sh`: ran each against the same file list inside a container with the packages `static.check.yml` installs (cppcheck 2.7, doxygen 1.9.1, aspell) — all clean, no findings.
   - `indent.sh`, `sloccount.sh`, `prohibited-words.sh`: read the scripts — each only matches `*.c`/`*.cpp` (indent, sloccount) or `*.c`/`*.h`/`*.cpp`/`*.hpp`/`*.py`/`*.sh`/`*.php`/`*.md` (prohibited-words); none of this PR's changed files (`.cc`/`meson.build`) match, so these checkers are no-ops for this PR regardless of content.
   - `flawfinder.sh`: the checker script itself has a pre-existing, unrelated bug (`for file in \`cat $file\`` instead of `$files` at line 87, leaving the loop variable unset) that makes it never actually invoke `flawfinder` on any file in CI today — out of scope of this PR so not raised as a finding here, just noting it so the "flawfinder passes" isn't mistaken for a real signal. I ran `flawfinder --minlevel=1` directly on the three changed `.cc` files myself: the only hits are (a) `[2] (misc) open` on `sp->open(&prop, &data)` calls in the new test — a well-known false positive from flawfinder matching the tensor-filter framework's `open` function pointer name, the same pattern already used throughout this test file's pre-existing cases, and (b) `memcpy`/`char` hits at lines 386/391/396/422 of `tensor_filter_tensorflow_lite.cc`, all in pre-existing code far from anything this PR touches (255-261, 465-483, 1455-1459). Nothing new from this diff.

2. Code tracing — walked `loadModel()` end to end, including the part after the visible diff hunk: `interpreter = nullptr` runs while the old `delegate_ptr`/`model` are still untouched (safe teardown of the old interpreter), then `model = std::move(new_model)` frees the old model only after the old interpreter is gone, then the new interpreter is built, then `setDelegate()` replaces `delegate_ptr` (destroying the old delegate — safe, since the old interpreter that used it is already destroyed) before `interpreter->ModifyGraphWithDelegate(delegate)` is called on the *new* interpreter with the *new* delegate. No stage frees an object that something still alive depends on.

3. Built this exact head (`1dea1fbe`) in a container with real TensorFlow Lite 2.15 (`external delegate ... links: YES` confirmed in the setup log) and ran `unittest_filter_tensorflow2_lite` myself: 13/13 pass.

4. Mutation test (new in this round, not done by round 1 or 2): reverted just the three-member declaration order in `TFLiteInterpreter` back to the pre-fix order (`interpreter`, `model`, `delegate_ptr`), rebuilt, and reran the suite. Result: `nnstreamerFilterTensorFlow2LiteLifetime.openClose`, `.reloadModel`, `.reloadUnmatchedModel_n`, `.sharedKeyReopen`, and `.extDelegateLibGivenTwice` all fail cleanly (`violations` counted 2-4, no crash) while the other 8 cases still pass — matching the PR body's mutation-matrix claim exactly. I then restored the fixed source and rebuilt/reran to confirm 13/13 again. This is first-hand confirmation that a regression in this exact area would turn the gating CI jobs (Tizen GBS `%check`, Ubuntu pdebuild) red, not just a re-reading of the claim.

5. Packaging / scope: confirmed (again, independently) `packaging/nnstreamer.spec`'s `%{_prefix}/%{nnstbindir}/unittest-nnstreamer/tests` directory glob covers the new `.so`, and `packaging/run_unittests_binaries.sh`'s `-name "unittest_*"` filter excludes it since it's named `libnnstreamer_unittest_tflite_lifetime_delegate.so`. `libdl_dep` (root `meson.build`) and `tflite2_extra_dep` (`ext/nnstreamer/tensor_filter/meson.build`) are both in scope when `tests/meson.build` uses them, and `meson setup` succeeds. No other module is touched.

6. Documentation: no public API or architecture change; nothing under `Documentation/` needed.

No findings to report this round.

myungjoo:

The NIT is valid and is now fixed. Without `TFLITE_EXTERNAL_DELEGATE_SUPPORTED` the sub-plugin ignores `Delegate:External`, so no delegate kernel is prepared and `openClose` would fail on its first assertion.

The test binary could not see that macro: it is part of `tflite2_extra_dep`, which only the sub-plugin and the helper library were given. So the fix is two lines:

- `tests/meson.build`: add `tflite2_extra_dep` to `unittest_filter_tensorflow2_lite`, so the same delegate detection meson already does reaches the test.
- the fixture's `SetUp()`: `#ifndef TFLITE_EXTERNAL_DELEGATE_SUPPORTED` → `GTEST_SKIP ()`. One guard covers all 5 cases; the leak case needs no delegate and keeps running.

Both arms measured in a container with tflite 2.15:

- as built in CI (macro present): 13/13 pass, nothing skipped, and the compile command for the test object now carries `-DTFLITE_EXTERNAL_DELEGATE_SUPPORTED`.
- control for a build without the external delegate (the `tflite2_extra_dep` line removed again, so the macro is absent): the 5 lifetime cases report `[ SKIPPED ]`, the other 8 pass, and the binary exits 0 instead of failing.

Folded into the F2 commit; new head `9a26dda1`. Re-run before pushing: clang-format 16 is a no-op on all three `.cc` files, `doxygen-tag.sh` passes with its negative control still erroring, and the `-Dwerror=true` build is clean.

myungjoo:

Re-review round 4, head `9a26dda1` (2 commits on `main`, F2/F3 of #4920).

**Verdict: mergeable as-is.** No blocking or major issues found.

## What I checked independently

- Read the full diff (`ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc` +8/-4, plus the new test helper library, the extended `unittest_filter_tensorflow2_lite.cc`, and `tests/meson.build`), the issue text for F2/F3, and all prior comments/reviews on this PR (rounds 1-3, plus `myungjoo-bot`'s round tagged `1dea1fbe` and the author's fold-in fix).
- Rebuilt this exact head in a container with real TensorFlow Lite 2.15 headers/libs and ran `unittest_filter_tensorflow2_lite` myself: **13/13 pass**, nothing skipped — matches the PR body and the previous round's CI evidence.
- Specifically verified the fold-in fix for `myungjoo-bot`'s NIT (the `nnstreamerFilterTensorFlow2LiteLifetime` fixture running even without `TFLITE_EXTERNAL_DELEGATE_SUPPORTED`):
  - Confirmed `tests/meson.build:336` now adds `tflite2_extra_dep` to `unittest_filter_tensorflow2_lite`, and that `tflite2_extra_dep` (`ext/nnstreamer/tensor_filter/meson.build:248`) is exactly the dependency that carries `-DTFLITE_EXTERNAL_DELEGATE_SUPPORTED` (conditionally) plus `TFLITE_INT8/16/FLOAT16/COMPLEX64`, `USE_TENSORFLOW2_HEADER_PATH`, `TFLITE_XNNPACK/GPU/NNAPI_DELEGATE_SUPPORTED`, `TFLITE_SUBPLUGIN_NAME`. Grepped the test file for all of these: only `TFLITE_EXTERNAL_DELEGATE_SUPPORTED` is referenced, so pulling in the dependency has no other behavioral effect on this test binary.
  - Confirmed `SetUp()` sets `lib`/`custom`/`get_count`/`lib_path` to `NULL` *before* the `#ifndef TFLITE_EXTERNAL_DELEGATE_SUPPORTED` / `GTEST_SKIP()` guard, so `TearDown()` (`if (lib) dlclose(lib); g_free(custom); g_free(lib_path);`) is safe on the skip path.
  - Empirically reproduced the "no external delegate" configuration myself (temporarily removed `tflite2_extra_dep` from the test target in the container, rebuilt, ran, then restored the file and rebuilt again to leave the container as found): the 5 lifetime cases report `[ SKIPPED ]`, the other 8 (including the F3 leak regression test) still run and pass, and the binary exits 0 — no crash in `TearDown`, no `ASSERT`/`SIGSEGV`. This matches the author's claimed control run.
  - Confirmed via `gh api .../pulls/4998/checks` and the job log for `9a26dda1`'s "Static checks" run (already green) that clang-format (apt v18, the CI version), the `doxygen-tag.sh` check, and `newline.sh` all pass with no diffs/errors on the changed files; also independently ran clang-format 16 locally over the three changed `.cc` files as a second data point — no-op on all three.
- Re-verified the F2 fix itself, not just the round-4 delta: `TFLiteInterpreter`'s members are now declared `delegate_ptr`, `model`, `interpreter` (interpreter last → destroyed first, matching a comment that states this explicitly), and `loadModel()` builds into a local `new_model`, resets `interpreter` first, then does `model = std::move(new_model)` — so a failed `BuildFromFile()` no longer clobbers the live `model`. Traced `TFLiteCore::checkSharedInterpreter()` → `TFLiteCore::init()` (unconditionally calls `loadModel()` again on an attached shared interpreter) to confirm `sharedKeyReopen` exercises a real, reachable production code path, not a test-only artifact.
- F3: `tflite_parseCustomOption()`'s `ExtDelegateLib` branch now does `g_free (option->ext_delegate_path)` before `g_strdup`, and `ext_delegate_path` is zero-initialized before the parse loop with a single call site, so there's no double-free risk.
- Scope: the change is local to `TFLiteInterpreter`'s member layout, `loadModel()`, and `tflite_parseCustomOption()`; no other module is touched. Size is proportionate — the bulk of the diff is the new test-only external delegate (`tflite_lifetime_delegate.cc`) needed because neither Ubuntu valgrind job builds this sub-plugin, so there's no other way to get a deterministic, crash-free witness for a delegate/model lifetime violation.
- Documentation: no public API or architecture change; nothing under `Documentation/` needed.

## Notes (non-blocking)

- The gating jobs for this exact head (Tizen GBS x86_64 `unit_test 1`, Ubuntu pdebuild jammy/resolute) were still pending at review time; "Static checks", the macOS meson build, and the arm valgrind job are green. My own container rebuild against real tflite 2.15 stands in for those pending jobs and matches the numbers the previous round measured against real CI logs for a prior commit on this branch.
- As already discussed in this PR: the skip arm can't hide a regression in any CI configuration today, because every CI job that builds this sub-plugin links the external delegate (`links: YES`); it only matters for a hypothetical downstream build without it, which is exactly the case it now handles correctly instead of failing.

No further action requested from my side.

myungjoo:
Reviewed by Astra via Codex

[Blocking — review-policy coverage threshold] At head `9a26dda14a8d8cee6ce344082eac30dd5a806b14`, executable-line coverage of the affected production functions is **119/178 (66.9%)**, below the required 80%; function coverage is **4/4 (100%)**. Scope: `TFLiteInterpreter` construction/destruction (affected by member order), `loadModel()`, and `tflite_parseCustomOption()` in `ext/nnstreamer/tensor_filter/tensor_filter_tensorflow_lite.cc`. This is GCC/gcov measurement on Ubuntu 22.04 with TensorFlow Lite 2.15, including the TFLite/shared-model unit tests and the 37-case TFLite plus 2-case reload SSAT suites. The five added executable lines are covered, but that narrower metric does not establish the whole-function threshold used for this review. Uncovered paths include external-delegate key/value options (lines 1460–1478), missing-library fallback (1516–1517), and delegate/model-loading error paths. Add regression coverage, or provide equivalent measured coverage from the supported test matrix, to meet 80% in this scope.

Impact: Although the new lifetime and repeated-path regression tests pass, the measured coverage does not meet the requested merge criterion and leaves adjacent delegate configuration and failure cleanup paths without demonstrated regression protection in this run; this is a coverage-policy blocker, not a reproduced functional failure of the fix.

myungjoo:

Both new reviews are answered at head `c16f6f10`.

**Coverage finding (Astra via Codex) — accepted and fixed.** The measurement was reproduced exactly: same scope (`TFLiteInterpreter` ctor/dtor, `loadModel()`, `tflite_parseCustomOption()`), same denominator, **119/178 = 66.9%** before. gcov on Ubuntu 22.04 with tflite 2.15, running the tflite and shared-model gtests plus the 37-case tflite and 2-case reload SSAT suites.

New cases, all of them free of any accelerator this host lacks:

- `extDelegateKeyValOptions` — `ExtDelegateKeyVal:mode#ok;no_separator` plus a second `ExtDelegateKeyVal` entry; the test delegate now reports how many key values it was handed, so this pins the parsing loop, the table creation, the merge of a repeated entry and the dropping of a malformed one (lines 1460-1478).
- `extDelegateRefusesToApply_n` — the delegate is told `mode#fail` and refuses, which reaches the sub-plugin's delegate-application error path.
- `extDelegateWithoutLib` — `Delegate:External` with no library falls back to no delegate (lines 1516-1517).
- `customOptionsParsed`, `customOptionsDelegateNames_n`, `openWithoutModelFile_n` — every remaining option: `NumThreads`, each delegate name, the QNN backend and performance modes including unknown values, an unknown option, and no model file. The delegate-name cases open a model that does not exist, so the parsed delegate is never built: the model load fails first.

**Measured after: 164/178 = 92.1%** line coverage, functions 4/4, and `tflite_parseCustomOption()` is now 100%. The 14 remaining lines are the NNAPI delegate body, the QNN body, and the `Failed to construct interpreter` / `Failed to allocate tensors` paths, which need an accelerator or a corrupt model; they are untouched by this patch.

**`myungjoo-bot`'s NIT (the skip arm is not exercised in CI) — acknowledged, no change.** Every job that builds this sub-plugin links the external delegate, so exercising that arm would need a CI configuration built against a TensorFlow Lite without it, which is outside this PR. It was verified locally instead: with the `tflite2_extra_dep` wiring removed, the 5 lifetime cases report `[ SKIPPED ]`, the other cases pass and the binary exits 0.

Commits stay split by item, each holding its own tests (delegate-dependent cases with F2, option-parsing cases with F3). Re-run before pushing: 19/19 in `unittest_filter_tensorflow2_lite`, clang-format 16 a no-op on all three `.cc` files, `doxygen-tag.sh` clean with its negative control still erroring, `-Dwerror=true` build clean. `origin/main` moved 20 commits meanwhile; it touches `tests/meson.build` in a different place, and the branch still merges cleanly.

myungjoo:

Re-review round 5 of head `c16f6f10` (2 commits on `f865ca74`, F2/F3 of #4920). This round answers "Astra via Codex"'s Blocking coverage-policy finding against `9a26dda1` (119/178 = 66.9%); the delta at this head is test-only (`tests/nnstreamer_filter_tensorflow2_lite/tflite_lifetime_delegate.cc`, `tests/nnstreamer_filter_tensorflow2_lite/unittest_filter_tensorflow2_lite.cc`), adding six gtest cases.

**Verdict: mergeable as-is. No Blocking / Major / Minor issues found.**

**Coverage claim: reproduced independently, matches exactly.**

I rebuilt the existing `-Db_coverage=true` tree, cleared `.gcda`, and ran `unittest_filter_tensorflow2_lite`, `unittest_filter_shared_model`, and the two named SSAT suites (`nnstreamer_filter_tensorflow2_lite`, `nnstreamer_filter_reload`), then read `gcov -f -m` on the sub-plugin's `.gcno`, once with the six new cases excluded and once with them included, for the same four affected-function scope stated in the PR (`TFLiteInterpreter` ctor/dtor, `loadModel()`, `tflite_parseCustomOption()`):

- Before (excluding the 6 new cases): **119/178 = 66.9%** — matches Astra's finding exactly.
- After (all cases): **164/178 = 92.1%**, `tflite_parseCustomOption()` **74/74 = 100%** — matches the author's claimed numbers exactly.

I also read the per-line `gcov` output for the remaining 14 uncovered lines to check the author's characterization rather than taking it on trust: they are the `NNAPI`/`QNN` delegate-construction bodies (`tensor_filter_tensorflow_lite.cc:558-571`, `602-658`, both `#ifdef`-gated and needing that hardware), the NNAPI-deleter lambda (0/2, same reason), and the `Failed to construct interpreter` / `Failed to allocate tensors` early-return pairs (`:492-493`, `:673-674`). Reaching those two specifically requires either the named accelerators or a `FlatBufferModel`/`InterpreterBuilder` failure (a corrupt/unloadable model or an OOM condition), neither of which the new cases introduce — the characterization is accurate, and I did not find a way to cover more of this scope without one of those.

**New tests, checked on their merits (not just re-read):**

- `extDelegateKeyValOptions`: traced `tflite_parseCustomOption()`'s `ExtDelegateKeyVal` branch by hand against the two given entries (`mode#ok;no_separator`, `extra#1`) — the malformed `no_separator` sub-pair is correctly dropped (`g_strv_length(kv) > 1` guard), and the table accumulates across repeated `ExtDelegateKeyVal` top-level entries (it is only allocated once, on `!table`), landing at exactly 2 entries. The delegate's `options_given` counter reports what `TfLiteExternalDelegateCreate` actually handed to `tflite_plugin_create_delegate`, so `EXPECT_EQ(get_count("options"), 2)` is a real assertion on parsed content, not a tautology.
- `extDelegateRefusesToApply_n`: `delegate_prepare` returning `kTfLiteError` under `mode#fail` makes `interpreter->ModifyGraphWithDelegate()` fail, which is exactly the sub-plugin's `"Failed to apply delegate"` return path (`tensor_filter_tensorflow_lite.cc:666-669`) — a real production error path, not test scaffolding.
- `customOptionsDelegateNames_n`: verified this cannot build an accelerator on the CI hosts. `loadModel()` calls `FlatBufferModel::BuildFromFile(model_path)` first and returns `-1` immediately on the nonexistent model path before the `switch (delegate_e)` block that would construct an NNAPI/GPU/QNN delegate is ever reached — so every one of the five `Delegate:` variants in this case only exercises the *parsing* branch, never delegate construction. Confirmed by reading `loadModel()`'s control flow directly.
- Global test-delegate state (`options_given`, `refuse_prepare` in `tflite_lifetime_delegate.cc`): both are reset in `nns_tflite_lifetime_delegate_reset()`, called from the fixture's `SetUp()`, and gtest runs cases single-threaded in one process. Only the `TEST_F(nnstreamerFilterTensorFlow2LiteLifetime, ...)` cases ever load this delegate; the four new plain `TEST(...)` cases (`extDelegateWithoutLib`, `customOptionsParsed`, `customOptionsDelegateNames_n`, `openWithoutModelFile_n`) never reference `lib_path`/`ExtDelegateLib`, so they cannot read stale state left by a fixture case, and no ordering dependency was introduced.
- Ran the full file myself in the container against real TensorFlow Lite 2.15: **19/19 pass** (13 pre-existing/round-4 + 6 new), matching the author's own numbers.

**Regression / scope:** confirmed via `git diff 9a26dda1 c16f6f10` that no production file changed in this round — only the two test files. Everything about F2/F3 correctness, member-order/`loadModel()` reasoning, and packaging scope was already independently re-derived and confirmed across rounds 1-4 and is unaffected by this round's test-only delta.

**Documentation:** no API/architecture change; nothing under `Documentation/` needed.

**CI:** on `c16f6f10`, "Static checks" and "CPP-Linter" have already completed successfully (confirms clang-format/doxygen/etc. on the new lines); I additionally ran clang-format 16 locally over both changed files with `--dry-run -Werror` as a second data point — clean, no diffs. The two gating jobs that actually execute `unittest_filter_tensorflow2_lite` (Tizen GBS x86_64 `unit_test 1`, Ubuntu pdebuild jammy/resolute) were still in progress at review time; nothing in my own rebuild suggests they would behave differently.

No further action requested from my side.

myungjoo:

The NIT is taken, with one correction to the metric. New head `b3c55d18`; `git diff c16f6f10 b3c55d18` is the whole change and it is test-only, folded into the F2 commit.

`freed` counts delegate **kernels**, and on the refusal path no kernel is ever built: `delegate_prepare ()` returns an error before `ReplaceNodeSubsetsWithDelegateKernels ()`, so `prepared == 0` and `freed == 0` there whatever happens to the delegate. Asserting `freed` would therefore always hold and check nothing.

What does catch the leak the NIT describes is the delegate object itself, which the test library already tracks in its live-id set. It is now exposed as a `live` counter, and:

- `extDelegateRefusesToApply_n` asserts `live == 0` after the failed open, so a delegate that survives a refused application fails the case.
- `openClose` asserts `live == 1` while the filter is open and `live == 0` after the close, which pins the same property on the normal path.

Measured discrimination: with `tflite_plugin_destroy_delegate ()` mutated to leave its id in the live set (a delegate that is never destroyed), exactly those two cases fail and the other five lifetime cases still pass.

Re-run before pushing: 19/19 in `unittest_filter_tensorflow2_lite`, clang-format 16 a no-op on all three `.cc` files, `doxygen-tag.sh` clean with its negative control still erroring, `-Dwerror=true` build clean.

myungjoo:

Re-review round 6 of head `b3c55d18` (`git diff c16f6f10 b3c55d18`, +9/-2 across two test files only, no production-code change this round). This round answers the previous approved review's NIT: `extDelegateRefusesToApply_n` did not assert that a refused-to-apply delegate was actually destroyed.

**Mutation control reproduces exactly as claimed.** I disabled `live_delegates.erase (d->id)` in `tflite_plugin_destroy_delegate ()` (`tests/nnstreamer_filter_tensorflow2_lite/tflite_lifetime_delegate.cc:241`), rebuilt only that shared library, and reran `unittest_filter_tensorflow2_lite` in the review container:

```
[  FAILED  ] nnstreamerFilterTensorFlow2LiteLifetime.openClose
[  FAILED  ] nnstreamerFilterTensorFlow2LiteLifetime.extDelegateRefusesToApply_n
[  PASSED  ] 5 tests.
```

Exactly the two cases named in the task failed, both at the new `EXPECT_EQ (get_count ("live"), 0)` lines (546 and 729), and the other five lifetime/plain cases stayed green. I restored the file, rebuilt, and reconfirmed 19/19 in `unittest_filter_tensorflow2_lite` (7 lifetime + 12 plain cases) before finishing.

I also verified the author's reasoning about the counter choice by reading the code rather than trusting the commit message:
- `freed` is indeed always 0 on the refusal path: `delegate_prepare ()` (`tflite_lifetime_delegate.cc:161-200`) returns `kTfLiteError` immediately when `refuse_prepare` is set, before `ReplaceNodeSubsetsWithDelegateKernels` runs, so `kernel_init`/`kernel_free` never execute and `kernels_freed` cannot move. `freed` was structurally incapable of catching a delegate-object leak on this path, so switching to `live` (delegates created and not destroyed, tracked in a `std::set<int>`) was the right fix rather than a workaround.
- `live` does exercise the sub-plugin's real teardown chain, not just the test lib's bookkeeping: on a refused `ModifyGraphWithDelegate ()`, `TFLiteInterpreter::loadModel ()` returns -2 (`tensor_filter_tensorflow_lite.cc:666-669`) after `setDelegate ()` has already put the delegate under `delegate_ptr`'s ownership (line ~596); `TFLiteCore::init ()` propagates the failure, and `tflite_loadModelFile ()` responds with `delete core` (`tensor_filter_tensorflow_lite.cc:1572-1578`), which runs `~TFLiteCore` → `delete interpreter` → the implicit `~TFLiteInterpreter`, destructing `delegate_ptr` and invoking the `TfLiteExternalDelegateDelete` deleter → `tflite_plugin_destroy_delegate ()`. So `live == 0` after a refused open is a real assertion about the F2 destruction-order fix, not just about the test double.
- The counter is safe across gtest cases run sequentially in one process: `nns_tflite_lifetime_delegate_reset ()` deliberately does not clear `live_delegates`/`last_delegate_id` (ids are never reused), which is why the mutation run above accumulated to `live == 10` by the time the last case ran — every earlier test's delegate stayed counted once its destroy was disabled. Under the unmodified code this same non-reset is what makes `live` a genuine cross-test leak detector rather than a single-test-scoped counter; it does not, by itself, cause any false failure since every existing case that opens also closes (or reaches `data == NULL`, guaranteed by `tflite_loadModelFile ()`'s `*private_data = NULL` before `delete core` on any init failure, so `if (data) sp->close (...)` in `extDelegateRefusesToApply_n` correctly skips a double-close).

**Independent pass over the whole PR (not just this round's delta):**
- Production fix (F2 declaration reorder + `loadModel()` local/move, F3 `g_free` before overwrite in `ExtDelegateLib` parsing) is unchanged since the already-approved `c16f6f10`; this round touches only the two test files.
- Scope is proportionate: this round is test-only, +9/-2, and directly closes the previously identified coverage gap — no other module is touched.
- Ran the full `unittest_filter_tensorflow2_lite` suite in the container: 19/19 pass (12 plain + 7 lifetime cases), consistent with the PR's own numbers.
- No API or architecture change, so no documentation update is due for this round.
- DCO passes on `b3c55d18`; the rest of this PR's CI (Tizen GBS, both pdebuilds, Valgrind, Static checks, etc.) was still queued/pending at review time — nothing in the local rebuild or the mutation run suggests it would behave differently from the already-green `c16f6f10`.

**Findings**

- **Nit** — PR body: the description/table added for earlier rounds does not mention the `live` counter or explain why `openClose`/`extDelegateRefusesToApply_n` gained new assertions in this round. Not a merge blocker, but a one- or two-line addition to the PR body (mirroring the commit's own justification) would help a future reader who only skims the PR description rather than the diff.

No blocking or major issues found. The mutation control reproduces exactly as the task describes, and the new assertions test real production teardown behavior rather than test-double bookkeeping.

**Verdict: no blocking issues. Comment only (per task instructions), no approve/request-changes from this automated pass.**

myungjoo:

Round 6's Nit is taken: the PR body now lists every case, including the six added for the option parser, and says what the test delegate reports — the key values it was handed, the `live` count of delegates it still holds, and the `mode#fail` switch that makes it refuse. The `openClose` and `extDelegateRefusesToApply_n` rows name the `live` assertions this round added.

No code changed for this, so the head stays `b3c55d18`.

PR #4999, opened 2026-09-17
[Filter] Zero the event data handed to a sub-plugin (#4920 B8, B13)

myungjoo:
item **B8**, **B13** of #4920

- **What:** two defects in the event data tensor_filter hands to a sub-plugin. **B8:** setting `inputlayout`/`outputlayout` on a configured V1 sub-plugin copied 256 layout entries from a stack struct that was never zeroed, so the sub-plugin saw garbage and the filter stored it. **B13:** SUSPEND, RESUME and the `is-updatable` probe send NULL event data, and the C++ wrapper turned that NULL into a reference (`*data`) before any sub-plugin code ran.
- **How:** B8 zeroes the struct before parsing (one line in `tensor_filter_common.c`). B13 makes `cpp_eventHandler()` pass a zero-filled local struct when `data` is NULL. C sub-plugins still get NULL, so no exported API changes; the C++ header now documents the zero-filled data.
- **Result:** a new `unittest_filter_event` binary (8 cases, 5 of them `_n`). Local mutation controls: without the B8 line, `setInputLayout`/`setOutputLayout` fail. Without the B13 change, each of the 5 SUSPEND/RESUME/is-updatable cases dies with SIGSEGV. Existing binaries pass unchanged: `unittest_sink` (138/138), `unittest_filter_single` (2/2), and `unittest_plugins --gtest_filter='testTensorFilterCppSubplugin.*:*ilter*'` (12/12, after a full `ninja` build; one of them needs the generated tflite fixture).

### Behaviour change
- After a runtime layout change, entries the string does not name are `ANY`. Before, they held undefined bytes.
- A C++ sub-plugin, or the tizen-hal backend it forwards to, receives a valid pointer to a zero-filled struct for SUSPEND/RESUME. Before, it got a reference bound to NULL, which is undefined behaviour.

---

<details><summary>Notes for agents / reviewers</summary>

- Commits: `[Filter] Zero the event data before parsing a new layout` (B8 + its tests + meson block), `[Filter] Hand C++ sub-plugins empty event data instead of NULL` (B13 + its tests + header doc line). Each commit builds and passes on its own.
- Other `GstTensorFilterFrameworkEventData` locals in `tensor_filter_common.c` (RELOAD_MODEL at `_gtfc_setprop_MODEL_FILES`, CUSTOM_PROP, SET_ACCELERATOR, DESTROY_NOTIFY, CHECK_HW_AVAILABILITY) were checked. Each sets every field its event defines and copies back only what it set, so none has the B8 copy-back pattern. They are left unchanged (minimal change).
- The fix for B13 went into the wrapper, not the SUSPEND/RESUME senders, because the `is-updatable` setter (`_gtfc_setprop_IS_UPDATABLE`) also sends RELOAD_MODEL with NULL, and `nnstreamer_plugin_api_filter.h` documents `data` as "can be NULL". No in-tree sub-plugin handles SUSPEND/RESUME/RELOAD_MODEL explicitly; tizen-hal forwards `std::addressof (data)` to the HAL backend.
- Why a new binary: the C++ mock precedent lives in `unittest_plugins.cc` and API-level filter tests live in `unittest_sink.cc`, both owned by open PRs. `gst_tensor_filter_common_free_property()` destroys the global `shared_model_table` (B5), which is safe here because this binary has no `shared-tensor-filter-key` case.
- B8 test technique: `_dirty_stack()` fills 64 KB of stack with `0xA5` through a `volatile` memset pointer, which prevents dead-store elimination, right before the setter. The uninitialised tail then reads as garbage instead of a lucky zero. Measured at `-O0` (local meson default); the valgrind CI step also sees the uninitialised copy.
- Local gates run: `-Dwerror=true` build, clang-format 16.0.6 on both `.cc`, gst-indent on the `.c` (the only diff it reports is pre-existing at line 511, outside this hunk), `doxygen-tag.sh … 1` with ctags (negative control: removing `_dirty_stack`'s `@brief` fails it).
- tflite2 is not available on the local box, so `nnstreamerFilterTensorFlow2Lite.suspend` (a C V1 sub-plugin, unaffected by the wrapper change) is CI-only.
</details>


myungjoo:
(Relayed review: written by a separate Claude review agent)

Reviewed against issue #4920 items **B8** and **B13**, checked for regressions in other modules, scope/size, test completeness, future-regression detection, CI enforcement, and doc/API impact. Verification included reading every in-tree C++ `eventHandler` implementation, and building/running the tests in a local WSL tree (both as shipped and with each fix independently reverted, to confirm the new tests actually catch the two defects).

## Summary verdict: mergeable, no blocking issues found.

### 1. Does it fix B8 and B13? — Verified, no issue

- **B8**: `tensor_filter_common.c:1735` now does `memset (&data, 0, sizeof (data))` before `gst_tensors_parse_layouts_string()` fills only the named entries. Since `_NNS_LAYOUT_ANY == 0` (`tensor_typedef.h:222`), unspecified entries now correctly become `_NNS_LAYOUT_ANY`, matching the field's own documented default ("`defaults to _NNS_LAYOUT_ANY for all the tensors`" in `nnstreamer_plugin_api_filter.h:121/126`). I reverted the fix back to `data.info = NULL;` in a scratch copy, rebuilt, and reran `unittest_filter_event`: `setInputLayout`/`setOutputLayout` failed exactly as the PR claims (garbage such as `-1515870811` instead of `0`/ANY past the named entries).
- **B13**: `cpp_eventHandler()` now passes a zero-filled local (`empty_data = {}`) instead of `*data` when `data` is NULL. I reverted this change (and removed the now-unused local, to fully restore the original code path) and reran the test: the process reliably **SIGSEGVs** on `testFilterEvent.suspendResume`, matching the PR's claim. `tizen_hal_subplugin::eventHandler()` (`tensor_filter_tizen_hal.cc:270`) forwards `std::addressof(data)` to the HAL backend; after the fix that address now genuinely points to zeroed memory instead of being derived from a null reference (previously undefined behavior).
- I checked every in-tree C++ sub-plugin (dali, deepview_rt, edgetpu, executorch(+llama), litert, llama2, llamacpp, lua, mediapipe, mxnet, ncnn, onnxruntime, python3, qnn, snap, snpe(+v1), tensorrt(10), tizen_hal(+snpe/+vivante), trix_engine, tvm): none override `eventHandler` for SUSPEND/RESUME/RELOAD_MODEL except the tizen_hal family, which only forwards the pointer — consistent with the PR's own audit note.

### 2. Regression risk to other modules — minor/nit only

- `unittest_filter_single` (2/2) and the filter-related suites in `unittest_plugins` (`testTensorFilter.*`, `testTensorFilterCppSubplugin.*`, `testTensorFilterAsync.*`) all pass locally (10/10 in my run).
- **(nit)** The PR body says this filtered `unittest_plugins` run is "12/12"; my local run of the same three suites shows 10 tests total. Not a code defect — likely just a stale/incorrect count in the PR description — but worth double-checking what the extra 2 cases were meant to be before merge, in case a case was silently dropped from the description rather than from the run.
- `unittest_sink` shows 137/138 passing; the one failure, `tensorStreamTest.tensorDecoderProperty1`, is in the tensor_decoder subsystem (unrelated to tensor_filter) and reproduces identically without any of this PR's changes present. It is **not** a regression caused by this PR — flagging only as FYI since `unittest_sink.cc` is explicitly out of this PR's scope per your instructions.
- The `data.layout` union-based struct means `memset`-zeroing the whole `GstTensorFilterFrameworkEventData` for `SET_INPUT_PROP`/`SET_OUTPUT_PROP` is equivalent to the old `data.info = NULL` for every field that mattered (info pointer + layout array), so no other union member is affected by this call site.

### 3. Scope / size — no issue

Change touches exactly the filter module: `tensor_filter_common.c` (1 line), `tensor_filter_support_cc.cc` (2 lines), the C++ plugin API header (1 doc line), plus a new, self-contained test directory and its meson wiring. No unrelated module is touched. The bulk of the diff (349 of 363 added lines) is the new test file, proportionate to covering 8 positive/negative cases for two distinct defects.

### 4. Test completeness — no issue

8 cases: `setInputLayout`/`setOutputLayout` (B8 positive), `setLayoutRefused_n` (B8 negative), `suspendResume`/`suspendUnsupported_n`/`resumeFail_n` (B13, SUSPEND/RESUME positive+negative), `isUpdatable`/`isUpdatableUnsupported_n` (B13, RELOAD_MODEL-via-NULL positive+negative). Both source fixes were independently confirmed load-bearing by local mutation testing (see section 1).

### 5/6. Future-regression detection and CI enforcement — no issue

- `tests/meson.build` registers `unittest_filter_event` as an ordinary `test()`, so it is picked up automatically both by `meson test -C build -v` and by `packaging/run_unittests_binaries.sh --recursive`, which is what `.github/workflows/ubuntu_clean_meson_build.yml` uses both for the normal test job and for the Valgrind memcheck job (`--valgrind --recursive ./tests/`). No extra CI wiring is needed for this new binary to run.
- If B8 regresses, it fails as a normal GTest assertion under the default (no `-Dbuildtype`, i.e. debug/-O0) CI meson build, **and independently** the Valgrind memcheck job would flag the uninitialized read regardless of the garbage value observed (memcheck detects the use of uninitialized memory itself, not a particular bad value), so detection does not solely depend on the "dirty stack" trick working at whatever optimization level a given job happens to use.
- If B13 regresses, the reverted wrapper reliably SIGSEGVs the test binary — this fails any CI configuration outright, no special detection logic needed.

### 7. Architecture/API/doc impact — no issue

No exported symbol or function signature changes (C API's `eventHandler(..., data)` still documents "`can be NULL`" unchanged; the C++ virtual `eventHandler(event_ops, GstTensorFilterFrameworkEventData&)` signature is unchanged). The one added doc line in `nnstreamer_cppplugin_api_filter.hh` documents previously-undocumented NULL-safety behavior right next to the method it describes, which is the appropriate place for it; no separate architecture doc needs updating for this change.

### Very minor/nit items (non-blocking)

- (nit) `clang-format --style=file` reflows the whole affected Doxygen comment block in `nnstreamer_cppplugin_api_filter.hh` (merging pre-existing, untouched lines together) when run on the full file. This is a pre-existing quirk of that comment block, not something introduced by the new line, and CI's `cpp-linter` job only scans `.cc` files with `lines-changed-only: true` (see `.github/workflows/cpp-linter.yml`), so `.hh` files aren't checked by it at all. No action needed.
- (nit) PR body's "12/12" vs. observed 10/10 for the filtered `unittest_plugins` run, noted above.

No blocking or major issues found. Approve from this review's perspective.

myungjoo:
(Relayed review, round 2: written by a separate Claude review agent)

Second pass on top of the first relayed review. Re-checked B8/B13 against issue #4920, re-verified the two round-1 follow-ups with local reproduction, and looked for anything new. Verification used the WSL build tree `~/nns-4920-b8b13` (read-only) plus a scratch copy `~/nns-4920-b8b13-review2` for clang-format experiments; nothing in either tree or in git was modified.

## Summary verdict: still mergeable, no blocking issues. One minor doc-wording finding, otherwise the two round-1 follow-ups check out.

### Follow-up 1 (clang-format/static.check.yml) — confirmed, correctly fixed

Reproduced directly: the pre-amend doc line ("`data is zero-filled if the event is sent without data (e.g., SUSPEND, RESUME).`", still present in the untouched `~/nns-4920-b8b13` tree) is **83 columns**, over the repo's `ColumnLimit: 80`. Running `clang-format 16.0.6 -i` on the full file with that line reflows the whole surrounding Doxygen block, merging the previously separate `Return -ENOENT ...` / `(e.g., let the framework do "free")` / `Return -EINVAL ...` lines together — a non-empty diff.

That matters because `.github/workflows/static.check.yml`'s "`/Checker/ clang-format for .cc/.hh/.hpp/.cpp files`" step runs `clang-format -i` on **every changed `.hh`/`.cc`/`.hpp`/`.cpp` file in full** (not line-restricted) and fails the job (`exit 1`) if `git diff -- *.cc *.hh *.hpp *.cpp` is non-empty. This is a separate job from `cpp-linter.yml` (which round 1 checked and which is indeed `.cc`-only, `lines-changed-only`). So round 1's "no CI impact" conclusion was wrong for this specific gate — the pre-amend commit would have failed Static checks.

The current (post-amend) line is 74 columns and round-trips through `clang-format -i` with **zero diff**, and `origin/main`'s untouched header also round-trips with **zero diff**, matching the PR author's claim. Confirmed correct — no outstanding CI risk from this line.

### Follow-up 2 (unittest_plugins count) — confirmed, correctly fixed

Ran the exact filter the PR body now names, `unittest_plugins --gtest_filter='testTensorFilterCppSubplugin.*:*ilter*'`, against the full build in `~/nns-4920-b8b13`: `--gtest_list_tests` shows 12 non-`DISABLED_` tests matching that filter (2 from `testTensorTransform`, 5 from `testTensorFilter`, 3 from `testTensorFilterCppSubplugin`, 2 from `testTensorFilterAsync`), and an actual run reports `[==========] 12 tests from 4 test suites ran. ... [ PASSED ] 12 tests.` The PR body's "12/12" together with the now-explicit filter is accurate. (Round 1's "10/10" was from a narrower, unstated filter of its own choosing, not a discrepancy in the PR.)

### New finding

- **(minor)** `gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh`, the added doc line: shortening it to fit the 80-column limit also dropped the "`e.g.,`" that was in the original wording — it now reads `data is zero-filled for an event sent without data (SUSPEND, RESUME).`, i.e. a closed, exhaustive-looking list. It isn't exhaustive: `_gtfc_setprop_IS_UPDATABLE()` (`tensor_filter_common.c:1677-1680`) also calls `eventHandler(..., RELOAD_MODEL, NULL)` as an "is this supported" probe — the same NULL-data path this PR's own `cpp_eventHandler()` fix now zero-fills, and the PR's own commit message/body says so explicitly ("the `is-updatable` setter ... also sends RELOAD_MODEL with NULL"). A future C++ sub-plugin author who implements `eventHandler` for `RELOAD_MODEL` and reads only this header comment could reasonably assume `data.model_files`/`data.num_models` are always populated for that op, since only SUSPEND/RESUME are called out as the zero-filled cases; they are not for this probe call. Not blocking today (the PR's own audit found no in-tree C++ sub-plugin implementing `RELOAD_MODEL`), but it is a real precision regression introduced by this round's amend (the pre-amend wording had "e.g." signaling a non-exhaustive example list). Suggest restoring the non-exhaustive framing (e.g. "data is zero-filled when an event carries no data (e.g., SUSPEND, RESUME)." fits in 80 columns) or naming all three current call sites.

### Other checks re-run this round (no issues)

- Diff vs. `origin/main` is unchanged from round 1 in shape: 1 line in `tensor_filter_common.c`, 2 lines in `tensor_filter_support_cc.cc`, 1 doc line in the `.hh`, plus the self-contained new test + its `meson.build` block. Scope still filter-module-only.
- `GstTensorFilterFrameworkEventData` is a plain C struct/union of PODs (pointers, ints, a `tensors_layout` array) — `memset` (B8) and `= {}` value-init (B13) are both safe, and equivalent to zeroing every member that any op could read.
- Rebuilt and ran `unittest_filter_event` directly in `~/nns-4920-b8b13`: all 8 cases (`setInputLayout`, `setOutputLayout`, `setLayoutRefused_n`, `suspendResume`, `suspendUnsupported_n`, `resumeFail_n`, `isUpdatable`, `isUpdatableUnsupported_n`) pass.
- The test file's `#include "../gst/nnstreamer/tensor_filter/tensor_filter_common.h"` from `tests/nnstreamer_filter_event/` looks like the wrong number of `..` at first glance, but it's the same convention already used by `unittest_sink.cc`, `unittest_filter_single.cc`, `unittest_query.cc`, etc. (resolved against the `tests/` include directory, not the including file's own directory) — not a new issue.
- No exported C or C++ API signature changed; the two source fixes stay inside existing, documented "`data` can be NULL" contracts. No separate architecture doc update is needed beyond the one header line discussed above.
- No security/backdoor concerns identified in this diff.

No blocking or major issues. The one minor item above is a documentation-precision nit, not a functional defect, and does not need to hold up merging.

myungjoo:
(Relayed review, round 3: written by a separate Claude review agent)

Third pass, focused on the amend made since round 2: the header doc line in `gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh` was folded into the B13 commit (`26af9643`) to restore the non-exhaustive "e.g." framing that round 2 flagged. Confirmed via `git diff 2ebc9fb6 26af9643` that this is the *only* change since round 2 (no `.c`/`.cc`/test/`meson.build` changes), so this round re-verifies that one hunk plus a final sweep for anything both earlier rounds missed.

## Summary verdict: still mergeable, no blocking or major issues. The round-2 minor finding is correctly resolved; no new issues found.

### Round-2 follow-up: "e.g." wording — confirmed, correctly fixed

The line now reads:
```
data is zero-filled when the event carries none (e.g., SUSPEND, RESUME).
```
- This restores the non-exhaustive example framing round 2 asked for, so it no longer reads as an exhaustive list that would exclude the `is-updatable` probe's `RELOAD_MODEL`-with-NULL call (`_gtfc_setprop_IS_UPDATABLE()`, `tensor_filter_common.c`), which goes through the same zero-fill path in `cpp_eventHandler()` (`tensor_filter_support_cc.cc:266-276`) added by this same commit.
- Column count: 77 (checked with `awk`), inside the repo's 80-column limit.
- clang-format 16.0.6 (`--style=file`, repo `.clang-format`) round-trips the full header file with **zero diff**, both for the current line and confirmed separately for the file as a whole — so this line does not trip `.github/workflows/static.check.yml`'s full-file, non-line-restricted clang-format gate that round 2 identified as the real risk (as opposed to `cpp-linter.yml`, which is `.cc`-only and line-restricted and round 1 correctly noted doesn't apply to `.hh`).
- The change is amended into the B13 commit (`26af9643`, "[Filter] Hand C++ sub-plugins empty event data instead of NULL"), i.e. it travels with the code it documents rather than sitting in a separate follow-up commit — appropriate, since it is a same-PR fix of a same-PR review defect and this project's convention is to fold such corrections into the original commit rather than add a narrating fix-up commit.
- DCO sign-off and commit message are intact on the amended commit; no dangling `Co-Authored-By` duplication.

### Other round-1/round-2 items — all resolved, nothing outstanding

- Round 1's two nits (the "12/12 vs 10/10" `unittest_plugins` count, and "no CI action needed" for the pre-amend clang-format reflow) were both superseded by round 2: the count was confirmed accurate once the exact filter was named, and the clang-format claim was corrected (a different, full-file CI gate does apply) — but by round 2's own account that column-length problem had already been fixed by amend *before* round 2 ran. Nothing further to act on here.
- Round 2's only finding (the "e.g." wording) is the one addressed by this round's amend, as above. No other round-2 item was left open.

### Fresh check this round: no new issues

- `git diff main...26af9643 -- gst/nnstreamer/tensor_filter/tensor_filter_common.c gst/nnstreamer/tensor_filter/tensor_filter_support_cc.cc gst/nnstreamer/include/nnstreamer_cppplugin_api_filter.hh` re-read in full: still exactly the 1-line `memset` (B8), the 2-line NULL-guard with `empty_data = {}` (B13), and the 1-line doc addition. Minimal and scoped to the filter module only, consistent with #4920 B8/B13.
- Re-scanned `tests/nnstreamer_filter_event/unittest_filter_event.cc`: every function (including the mock subplugin's methods and `_dirty_stack()`) still carries a `/** @brief ... */` Doxygen comment, satisfying the repo's CI doc-comment requirement; the 8 `TEST_F` cases match the PR body's description and the `_n` suffix convention for the negative cases.
- No architecture or exported-API change accompanies this amend (it is a comment-only edit to an already-added, non-exported header doc line), so no additional documentation update is owed beyond the line itself.
- No regression risk: a comment-only change cannot affect any other module's behavior or performance.

No blocking or major issues. Merge-ready from this review's perspective.

myungjoo:
(Relayed review, round 4: written by a separate Claude review agent)

Scope: the amend since head `26af9643` (`git diff 26af9643 5cc6e2b7` — only `tensor_filter_support_cc.cc` changed), made in response to `myungjoo-bot`'s formal review NIT about `GstTensorFilterFrameworkEventData empty_data = {};` being aggregate initialization of a struct whose only member is a union. Also re-verified that rounds 1-3 and the bot review are fully resolved, and looked for new issues. Verified by reading the struct layout, rebuilding/running `unittest_filter_event` in a local WSL tree, and by mutation-testing a scratch copy (`~/nns-4920-b8b13-review4`, deleted afterward; the reference tree `~/nns-4920-b8b13` was left untouched).

## Summary verdict: mergeable, no blocking or major issues. One minor test-coverage gap and one negligible-cost nit, both non-blocking.

### 1. Does the `memset` amendment actually resolve the bot's NIT? — Yes, verified

`GstTensorFilterFrameworkEventData` (`nnstreamer_plugin_api_filter.h:206-244`) is a struct whose sole member is an anonymous union of six anonymous structs, the largest holding `const GstTensorsInfo *info` plus `tensors_layout layout` (`tensor_layout[NNS_TENSOR_SIZE_LIMIT]`, i.e. `256 * sizeof(enum)` ≈ 1KB, since `NNS_TENSOR_SIZE_LIMIT == 256` in `tensor_typedef.h:42`). Per the C++ standard, zero/value-initializing a union (`= {}`) only guarantees the *first* named member is zeroed (here, the `void *data` field of the first alternative) plus true padding; the other, larger alternatives are not guaranteed to be zeroed by the language, even though GCC/Clang usually do so in practice. Declaring `empty_data` as a plain (uninitialized) local and then `memset (&empty_data, 0, sizeof (empty_data));` zeroes the entire object regardless of which union member is later read, which is what the bot asked for and what B8's existing fix in `tensor_filter_common.c` already does for the same struct type. This is objectively correct and portable.

**Placement**: the `memset` sits after `GET_TFSP_WITH_CHECKS (obj, private_data);`, which can `return -EINVAL;` early (it expands to a `do { ... return -EINVAL; ... } while(0)` on exception). Placing the `memset` after that guard means the zero-fill is skipped on the early-return path, where `empty_data` is never used anyway — a small, correct, side-effect-free optimization, not a correctness issue (nothing reads `empty_data` before the `memset` in either path).

### 2. Does this change alter existing behavior/performance, including the `data != NULL` path? — Behavior: no change. Performance: negligible, but worth noting since asked

`cpp_eventHandler()` is invoked for every V1 C++ event, not just `SUSPEND`/`RESUME`. Looking at the call sites in `tensor_filter_common.c`: `DESTROY_NOTIFY` (`gst_tensor_filter_destroy_notify_util`, line ~822, which fires once per output tensor memory for `allocate_in_invoke`/zero-copy sub-plugins — a per-buffer, not per-property-set, path), `SET_INPUT_PROP`/`SET_OUTPUT_PROP` (line ~1744), `CUSTOM_PROP`/`SET_ACCELERATOR`/`CHECK_HW_AVAILABILITY` (lines ~1602/1641/3052) all pass a real, non-NULL `&data`. Only `SUSPEND`, `RESUME` (`tensor_filter_common.c:2531,2553`) and the `is-updatable` `RELOAD_MODEL` probe (`:1678`) pass `NULL`.

The new `memset (&empty_data, 0, sizeof (empty_data));` runs unconditionally on *every* call to `cpp_eventHandler`, including all of the `data != NULL` cases above, even though `empty_data` is provably unused whenever `data != NULL` (`data ? *data : empty_data`). So every `DESTROY_NOTIFY` call for a zero-copy C++ sub-plugin now does one extra ~1KB stack `memset` it never needed. In absolute terms this is negligible — a contiguous ~1KB `memset` costs on the order of tens of nanoseconds on any modern CPU, dwarfed by the `try`/`catch`, virtual dispatch, and (for `tizen_hal`) HAL IPC machinery already present in the same function — so this is **not** a real performance regression. Flagging only because the round-4 prompt specifically asked about the `data != NULL` path's cost: it is not zero, but it is not measurable either. (Nit, non-blocking; could trivially be avoided with `if (!data) memset (&empty_data, 0, sizeof (empty_data));` if anyone wants to eliminate even this.)

### 3. Are rounds 1-3 and the bot review fully resolved? — Yes, all confirmed

- Round 1's two nits (stale "12/12 vs 10/10" `unittest_plugins` count; clang-format CI-impact claim) were both already superseded/corrected by round 2, and nothing further changed about them since.
- Round 2's "e.g." wording finding (the shortened doc line reading as an exhaustive list) — confirmed fixed and unchanged since round 3: `nnstreamer_cppplugin_api_filter.hh` still reads "data is zero-filled when the event carries none (e.g., SUSPEND, RESUME)." (77 columns, round-trips through `clang-format 16.0.6 --style=file` with zero diff). This wording is still accurate post-amend: it doesn't claim to be exhaustive, so it doesn't misrepresent the `RELOAD_MODEL` probe case.
- Round 3 found nothing new and confirmed the doc-line amend was correctly folded into the B13 commit rather than added as a narrating fix-up — the same is true again this round: the memset amend lives inside the same `5cc6e2b7` ("[Filter] Hand C++ sub-plugins empty event data instead of NULL") commit, with an intact DCO sign-off and no duplicate `Co-Authored-By`.
- Bot review's NIT (aggregate-init union zeroing) — resolved by the memset, per section 1 above. The bot's other observations (B8 mechanism, B13 mechanism, C sub-plugins unaffected, Valgrind-gated `unittest_filter_event`, one-item-per-commit, 21/21 green) were about content that hasn't changed since and remain accurate.
- Local re-verification: `unittest_filter_event` rebuilt and run against the current head in `~/nns-4920-b8b13` — all 8 cases (`setInputLayout`, `setOutputLayout`, `setLayoutRefused_n`, `suspendResume`, `suspendUnsupported_n`, `resumeFail_n`, `isUpdatable`, `isUpdatableUnsupported_n`) pass.
- Scope/size unchanged from earlier rounds: this round's diff is 1 file, +4/-1 lines, confined to `tensor_filter_support_cc.cc`. No other module touched.

### New finding (minor): the SUSPEND/RESUME/RELOAD_MODEL-probe tests don't actually exercise the specific defect class the bot's NIT (and this round's fix) address

`testFilterEvent.suspendResume`/`suspendUnsupported_n`/`resumeFail_n`/`isUpdatable`/`isUpdatableUnsupported_n` all check `event_mock_subplugin::last_data` (which captures `data.data`, the union's first-declared member) against `nullptr`. But `data.data` is exactly the one field the C++ standard already guaranteed to be zero even under the old, pre-amend `empty_data = {};` (aggregate init zeroes the union's first named member unconditionally). None of these five tests reads a non-first union member (e.g. `data.layout`, `data.num_models`, `data.custom`) while handling a data-less event, so they cannot distinguish "fully memset'd" from "only-first-member-zeroed."

I verified this concretely: in a scratch copy, I changed `cpp_eventHandler` to `memset (&empty_data, 0xA5, sizeof (empty_data)); empty_data.data = nullptr;` (i.e., simulating exactly the guarantee the old `= {}` code had — first member zero, everything else garbage), rebuilt, and reran `unittest_filter_event`: **all 8 tests still pass**, including all 5 SUSPEND/RESUME/is-updatable cases. So, in isolation, this test file would not catch a future regression back to `GstTensorFilterFrameworkEventData empty_data = {};` for this specific union-zeroing concern — the exact regression class this round's amend fixes.

This doesn't block merge: the production fix (`memset`) is unconditionally correct regardless of what the tests check, and no in-tree C++ sub-plugin currently reads a non-first union member on a data-less event (per the earlier rounds' subplugin audit), so there's no live bug today. But since the task explicitly asks about future-regression detection and CI's ability to catch it, it's worth recording: a stronger test would read a non-first-member field (mirroring the `_dirty_stack()` trick already used by `setInputLayout`/`setOutputLayout` for the B8 fix) during a SUSPEND/RESUME/RELOAD_MODEL-probe call, e.g. asserting `data.layout[NNS_TENSOR_SIZE_LIMIT - 1] == 0` or `data.num_models == 0` after dirtying the stack beforehand.

### Other checks (no issues)

- No architecture/exported-API change; the C API (`data` can be NULL) is unchanged, and the one-line doc addition (round 2/3) already documents the C++-side NULL-safety guarantee accurately.
- `tensor_filter_support_cc.cc` round-trips through `clang-format 16.0.6 --style=file` with zero diff (checked directly).
- PR remains scoped to the filter module; no unrelated module touched by this round's amend.
- CI on the current head (`5cc6e2b7`) is still partially pending at review time; the checks that have completed so far (Valgrind on ubuntu-22.04-arm, macOS meson build, Spell Check, CPP-Linter, DCO) are green.

No blocking or major issues. Merge-ready from this review's perspective, contingent on the still-pending CI checks for `5cc6e2b7` going green.

myungjoo:
(Relayed review, round 5: written by a separate Claude review agent)

Scope: the amendments since round 4's head (`5cc6e2b7` -> current head `c376197e`), i.e. (1) the bot-NIT `memset` fix's final shape (`cpp_eventHandler` reassigns the `data` parameter to a zero-filled local instead of computing `data ? *data : empty_data`), (2) the mock sub-plugin's new whole-struct dirty/zero-check technique added to close round 4's test-coverage gap, and (3) the rebase onto current `origin/main` (only `tests/meson.build` overlapped, no conflict). Also re-verified that rounds 1-4 and the `myungjoo-bot` formal review are still resolved at this head, and looked for anything new. Verified by rebuilding and running the full suite plus targeted mutation testing in a local WSL tree (`~/nns-4920-b8b13-rb`, a fresh rebase-head checkout; a disposable copy was used for the mutations, the reference tree itself was not modified).

## Summary verdict: mergeable, no blocking or major issues. Two non-blocking nits below; everything from rounds 1-4 and the bot review remains resolved.

### 1. B8/B13 still fixed at this head - re-verified

- `tensor_filter_common.c:1735`: `memset (&data, 0, sizeof (data))` before parsing the layout string - unchanged since round 1, still correct.
- `tensor_filter_support_cc.cc`: `cpp_eventHandler()` now reads:
  ```
  GstTensorFilterFrameworkEventData empty_data;
  GET_TFSP_WITH_CHECKS (obj, private_data);
  ...
  if (data == NULL) {
    memset (&empty_data, 0, sizeof (empty_data));
    data = &empty_data;
  }
  ...
  ret = obj->eventHandler (ops, *data);
  ```
  The call site went back to `*data` (uniform for both branches) by reassigning the `data` parameter itself to `&empty_data` when it was NULL. This is an appropriate, idiomatic normalization: `data` is an ordinary local pointer parameter (not `const`, not returned, not stored beyond this call), `empty_data` outlives every use of `data` in this function, and there is no aliasing/thread-safety concern since `empty_data` is a plain automatic (non-`static`) local, fresh per call. It reads more simply than a ternary at the call site and was confirmed to build clean under `-Dwerror=true` (no "maybe-uninitialized" or similar warning for this file in a full rebuild log). No issue.
- Rebuilt `~/nns-4920-b8b13-rb` from scratch (`ninja`, 276/276 targets, 0 errors) and reran the suites named in the PR body: `unittest_filter_event` 8/8, `unittest_sink` 138/138, `unittest_filter_single` 6/6, `unittest_plugins --gtest_filter='testTensorFilterCppSubplugin.*:*ilter*'` 12/12. All match the PR body's numbers exactly.

### 2. Round 4's test-coverage gap - correctly closed, confirmed by mutation

Round 4's finding was that `suspendResume`/`suspendUnsupported_n`/`resumeFail_n`/`isUpdatable`/`isUpdatableUnsupported_n` only checked `last_data` (the union's first member, `data.data`), so they could not tell "fully memset" apart from "only-first-member-zeroed". This round adds `_dirty_event_record()` (memsets the mock's recorded `last_event` to `0xA5` before the call) and `_event_record_is_zeroed()` (`memcmp` of the *whole* recorded struct against an all-zero one), used by `suspendResume` and `isUpdatable`.

Verified by mutation in a disposable copy (`~/nns-4920-b8b13-review5`, deleted after use):
- Removing the `memset` call entirely (keep `data = &empty_data;` over an uninitialized local) -> rebuilt -> `testFilterEvent.suspendResume` and `testFilterEvent.isUpdatable` **fail** (`_event_record_is_zeroed()` is false, `last_data` is non-null garbage). This matches the PR body's mutation claim and confirms the new checks are load-bearing for the full-zero-fill property, not just the first member.
- This closes round 4's gap: a future accidental narrowing back to "only clear `data.data`" (e.g. `empty_data.data = nullptr;` without a full memset) would now be caught, which round 4 explicitly said the pre-round-5 test could not do.

### 3. The `= {}` mutation gap noted in the task background - real, but an acceptable, disclosed limitation (nit, not blocking)

Reproduced directly: reverting the fix to `GstTensorFilterFrameworkEventData empty_data = {}; data = &empty_data;` (no `memset`, the exact pattern `myungjoo-bot`'s NIT and round 4 objected to) still **passes all 8 `unittest_filter_event` cases** when rebuilt with this repo's GCC/meson defaults, because GCC zero-fills the whole object for this aggregate-init pattern in practice (as the bot review itself already noted). Neither the new dynamic test nor Valgrind's memcheck can distinguish this from the shipped `memset` version, because under this toolchain there is no uninitialized memory at runtime to flag in either case - the gap is in language-guarantee coverage, not in an observable runtime defect.

- (nit) Impact: a future PR that "simplifies" `cpp_eventHandler()` back to aggregate-init would not be caught by CI on GCC/Clang, only by manual review or a standards/portability-focused static check (e.g. a sanitizer/compiler combination that does not zero-fill unioned aggregates) - a low-probability, no-live-bug-today scenario, matching the bot's own "theoretical" characterization of the original NIT. Given the production code already uses the objectively-correct `memset` (not aggregate-init), this is acceptable to ship as-is; a test that could distinguish the two would need something beyond ordinary GTest (e.g. `-ftrivial-auto-var-init=pattern` combined with an inspection of specific bytes, which is disproportionate to this PR's scope).

### 4. New, independent nit: `event_mock_subplugin::num_events[RESUME + 1]` sizing is fragile to an unrelated future change (nit, not blocking)

`tests/nnstreamer_filter_event/unittest_filter_event.cc`: `guint num_events[RESUME + 1];` sizes the mock's per-op counters off `RESUME`, which is currently the last enumerator of `event_ops` (`nnstreamer_plugin_api_filter.h:190-191`). If a future PR appends a new `event_ops` value after `RESUME` (an unrelated framework change, not one this PR needs to anticipate in depth), any test that exercises the new op through this mock would index `num_events[]` out of bounds - a silent stack write in test-only code, not a clean test failure. Impact: low probability, test-code-only, and only latent until such a future enum change and a test exercising it are both added; flagging only because the task explicitly asks about detecting the effect of unrelated future module changes. Not something this PR needs to fix.

### 5. Rebase - clean, verified

`git diff`-equivalent check against the rebased tree shows the changed-file set is unchanged from round 4's content (`tensor_filter_common.c` +1/-1, `tensor_filter_support_cc.cc` +6, the header +1 doc line, plus the self-contained new test and its `tests/meson.build` block). The `tests/meson.build` insertion sits cleanly between the `unittest_latency` and `unittest_filter_single` registrations with no stray context from the rebase. No other file conflicts.

### 6. Other checks re-run this round (no issues)

- `clang-format --style=file` round-trips all three touched C++/header files with **zero diff** in the rebased tree (`tensor_filter_support_cc.cc`, `nnstreamer_cppplugin_api_filter.hh`, `unittest_filter_event.cc`).
- `gst-indent` on `tensor_filter_common.c` reproduces exactly the PR body's claim: its only reported diff is a pre-existing, out-of-hunk line-wrap at line ~511 (`ml_loge (...)`), unrelated to this PR's line-1735 change.
- I also ran this repo's CI `doxygen-tag.sh` (advanced/per-function mode) locally and it flags the pre-existing `namespace nnstreamer { ... }` line in both `tensor_filter_support_cc.cc` and the header as "needs @brief". This reproduces identically against the unmodified `origin/main` copies of those two files with `main`'s own copy of the checker, and PR #4829 (previously merged, also touching `tensor_filter_support_cc.cc`) shows a **passing** "Static checks" job on GitHub - so this is a local ctags-version artifact in my sandbox (it also prints `ctags: Warning: Unsupported kind: 'c' for --c-kinds option`), not a real defect in this PR or a genuine CI risk. Recorded for completeness, no action needed.
- Commit structure: 2 commits, one item each (B8, B13), each self-contained (builds/passes on its own per the PR body) and each extending the same test file incrementally (+261 then +124 lines) rather than splitting one topic across commits or bundling unrelated topics together. DCO sign-off present once per commit, no duplicate `Co-Authored-By`. No issue.
- CI wiring: `unittest_filter_event` is a plain `test()` inside the existing gtest block, so it is picked up by both the normal `meson test` job and `packaging/run_unittests_binaries.sh --valgrind --recursive ./tests/` (used by `ubuntu_clean_meson_build.yml`'s Valgrind job) with no extra wiring - consistent with round 1's finding, re-confirmed against the current `tests/meson.build`.
- No exported API/architecture change beyond the one already-reviewed header doc line; no additional documentation is owed.
- No security or backdoor concern in this diff - it is a same-module, memory-safety-only change with no new external input handling.
- Scope/size: this round's total diff from round 4 is small and confined to `tensor_filter_support_cc.cc`'s call-site shape plus the new test helpers/assertions; no unrelated module touched.

No blocking or major issues. Merge-ready from this review's perspective, contingent on the current head's CI (re-triggered by the rebase; still mostly pending at review time, with CPP-Linter and DCO already green) coming back green.

myungjoo:
(Relayed review, round 6: written by a separate Claude review agent)

Scope: the change since round 5's head (`c376197e` -> current head `42a9bdc2`), i.e. only round 5's finding #4 (`num_events[RESUME + 1]` OOB-write fragility) being addressed, plus a final sweep confirming rounds 1-5 and the `myungjoo-bot` formal review are still resolved. Verified by rebuilding and running the tests in two fresh, disposable WSL trees seeded from the reference tree (`~/nns-4920-b8b13-rb`, left unmodified): one holding only the B8 commit's snapshot (`23473f90`), one at the full current head (`42a9bdc2`); both deleted after use.

## Summary verdict: mergeable, no blocking or major issues.

### 1. Round 5 finding #4 (`num_events[RESUME + 1]` fragile sizing) — confirmed, correctly fixed (was: nit)

`git diff c376197e 42a9bdc2` touches exactly one file, `tests/nnstreamer_filter_event/unittest_filter_event.cc`, +3/-1:
```cpp
-    num_events[ops]++;
+    EXPECT_LT ((guint) ops, G_N_ELEMENTS (num_events));
+    if ((guint) ops < G_N_ELEMENTS (num_events))
+      num_events[ops]++;
```
This closes the gap round 5 flagged: if a future, unrelated change appends a new `event_ops` enumerator after `RESUME` and a later test exercises it through this mock, the code no longer silently writes past `num_events[]` — it now fails the test via `EXPECT_LT` and skips the increment instead of corrupting adjacent memory. `event_ops` (`nnstreamer_plugin_api_filter.h:181-192`) still has `RESUME` as its last enumerator today, so this is purely a defensive guard for tomorrow, not a fix for a live defect.

Per the task background, the identical hunk appears in the diff of both the B8 commit (`327b984a` -> `23473f90`) and the B13 commit (`c376197e` -> `42a9bdc2`) — confirmed directly (`git diff 327b984a 23473f90` and `git diff c376197e 42a9bdc2` produce the same patch). That is expected, not a duplication defect: the mock class (and this method) is introduced in the B8 commit and B13's tree is cumulative on top of it, so rewriting the guard into the B8 commit necessarily reappears in a full-tree diff of the B13 commit as well. Both commits still build and pass independently (see section 3).

### 2. Round 5 finding #3 (`= {}` vs `memset` mutation gap) — decision to not add a test is reasonable; a cheap non-test mitigation is available (nit, non-blocking)

Re-reproduced the gap directly: reverting `tensor_filter_support_cc.cc` to `GstTensorFilterFrameworkEventData empty_data = {}; data = &empty_data;` (no `memset`) still passes all 8 `unittest_filter_event` cases on this repo's GCC 11 toolchain, because GCC zero-fills the whole object for this pattern in practice even though the standard only guarantees the union's first named member. This matches round 5's and the bot's own characterization: the gap is in language-guarantee coverage, not an observable runtime defect on the compilers this project actually builds with (GCC/Clang CI matrix). A GTest assertion, and Valgrind's memcheck, can only catch a difference that produces observably different bytes or reads uninitialized memory at runtime — neither is true here on GCC/Clang. So declining to chase this with more dynamic tests is the right call; a test that could reliably distinguish the two would need a toolchain that does *not* zero-fill unioned aggregates (not part of this project's CI matrix), which is disproportionate to add for a defect class with no live instance today.

That said, since the concern is purely "a future refactor could silently reintroduce the weaker pattern," the cheap mitigation that *is* proportionate is a one-line comment at the `memset` call site (`tensor_filter_support_cc.cc:271-272` in the current diff) along the lines of:
```cpp
/* Use memset, not aggregate init ("= {}"): only the union's first member
 * is guaranteed zero by the standard; GCC/Clang zero it all in practice,
 * but that is not a language guarantee. */
memset (&empty_data, 0, sizeof (empty_data));
```
This directly targets future-regression detection the way the task asks about it — not via CI (which, as established, cannot detect it), but by making the reason legible to the next reviewer/author who might "simplify" it back to `= {}`. Non-blocking; the production code is already correct as shipped.

### 3. Build/test verification — matches the PR body's claims exactly

Built two fresh, independent trees from file snapshots at each commit (not `git checkout` in place, since the reference tree isn't a git checkout):
- **B8 only** (`23473f90` snapshot of `tensor_filter_common.c`, `tensor_filter_support_cc.cc`, the header, `tests/meson.build`, and `unittest_filter_event.cc`): clean `-Dwerror=true` build, `unittest_filter_event` — **3/3** (`setInputLayout`, `setOutputLayout`, `setLayoutRefused_n`).
- **Full head** (`42a9bdc2`): clean `-Dwerror=true` build (full `ninja`, all targets, 0 warnings/errors), `unittest_filter_event` — **8/8**. Also re-ran `unittest_filter_single` (**6/6**) and `unittest_plugins --gtest_filter='testTensorFilterCppSubplugin.*:*ilter*'` (**12/12**) — both match the PR body. `unittest_sink` — **137/138**, the one failure (`tensorStreamTest.tensorDecoderProperty1`) is the same pre-existing, tensor_decoder-only failure earlier rounds already identified as unrelated to this PR and reproducing on unmodified code; not a regression from this change.
- `clang-format --style=file` round-trips `unittest_filter_event.cc` with zero diff.

### 4. Rounds 1-5 and the bot review — all still resolved, nothing reopened

- Bot review's NIT (aggregate-init union zeroing) — resolved by the `memset` (round 4), unchanged since.
- Round 1 (stale test count, clang-format CI-impact claim) and round 2 (doc-line "e.g." wording) — resolved, unchanged since round 3's confirmation.
- Round 3 — nothing outstanding.
- Round 4's test-coverage gap (tests only checked the union's first member) — closed by round 5's `_dirty_event_record`/`_event_record_is_zeroed` helpers, unchanged and still load-bearing this round (confirmed passing here; not re-mutated this round since round 5 already did so and nothing touched that code path since).
- Round 5's two nits — #3 addressed in section 2 above (decision endorsed), #4 addressed in section 1 above (fixed).

### 5. Scope/size, docs, CI wiring — no issue, unchanged from round 5

This round's total diff from round 5 is 4 lines in one test file; nothing else touched. No exported API/architecture change; no additional documentation owed beyond what earlier rounds already reviewed. `unittest_filter_event` remains an ordinary `test()` in `tests/meson.build`'s existing gtest block, so it's still covered by both the normal test job and the Valgrind memcheck job with no extra wiring needed.

No blocking or major issues. Merge-ready from this review's perspective; CI on the current head was still mostly pending at review time (only "Build on MacOS-latest with Meson" and DCO had completed, both green).

myungjoo:
(Relayed review, round 7: written by a separate Claude review agent)

Scope: the amend since round 6's head (`42a9bdc2` -> current head `b7e15986`). Confirmed via `git diff 42a9bdc2 b7e15986` that it is exactly one line, in `tensor_filter_support_cc.cc`:
```cpp
   if (data == NULL) {
+    /* An aggregate initializer zeroes only the union's first member. */
     memset (&empty_data, 0, sizeof (empty_data));
     data = &empty_data;
   }
```
This is round 6's own suggested mitigation for its finding #3 (the `= {}` vs `memset` mutation gap that neither GTest nor CI's Valgrind job can detect on this project's GCC/Clang toolchains), folded into the B13 commit. Also did a full sweep of rounds 1-6 and the `myungjoo-bot` formal review to confirm nothing has reopened, and re-verified B8/B13 end to end. Verified by rebuilding and running the full suite in two fresh, disposable WSL trees seeded from the reference tree `~/nns-4920-b8b13-rb` (left unmodified; both disposable trees deleted after use): one at the full current head, one holding only the B8 commit's (`23473f90`) file snapshot.

## Summary verdict: mergeable, no blocking or major issues. No remaining open items.

### 1. Is the new comment accurate? — Yes, verified against the actual struct and the standard's union-zeroing rule (Severity: no issue)

`GstTensorFilterFrameworkEventData` (`gst/nnstreamer/include/nnstreamer_plugin_api_filter.h:206-244`) is a struct whose only member is an anonymous union of six anonymous structs (`{ void *data; }` for DESTROY_NOTIFY, `{ model_files; num_models; }` for RELOAD_MODEL, `{ info; layout; }` for SET_INPUT_PROP/SET_OUTPUT_PROP, etc.). For `T x = {};` on such a type, aggregate-initialization ([dcl.init.aggr]) initializes the (sole) union member from an empty initializer-list, which value-initializes the union; value-initializing a union with an implicit default constructor is zero-initialization, and zero-initializing a union ([dcl.init], the union case) zeroes only its *first* named member, not the whole storage. Here the union's first alternative is `{ void *data; }`, so `= {}` only guarantees `data` (8 bytes) is zero; `layout` (a `NNS_TENSOR_SIZE_LIMIT`-entry array, ~1 KB), `model_files`/`num_models`, `hw_list`/`num_hw`, etc. are not guaranteed zero by the language, even though GCC/Clang zero the whole object in practice (as the bot's original NIT, and rounds 4-6, already established). So "an aggregate initializer zeroes only the union's first member" is a correct, non-overstated summary of that rule for this exact type — it doesn't claim a stronger or weaker guarantee than the standard actually gives, and it doesn't misdescribe `GstTensorFilterFrameworkEventData` as anything other than what it is (a struct wrapping a union).

Word choice check: "aggregate initializer" is the standard's own term for `= {}` on a type like this (post-DR1467, which is what makes `= {}` prefer aggregate-init over plain value-init for a type with a union member) — matches the terminology the bot's own NIT and round 4 already used, so it's consistent with the PR's own review history rather than introducing new, possibly-imprecise vocabulary.

### 2. Is the comment appropriate given the project's "no unnecessary comments" norm? — Yes (Severity: no issue)

- I did not find a written "avoid unnecessary comments" rule in `CLAUDE.md`, `AGENTS.md`, or `Documentation/contributing.md` (only the Doxygen-per-function requirement, which does not apply to an in-body statement comment like this one). Judging against ordinary good-comment practice instead: a good comment explains *why*, not *what*. This one does exactly that — it doesn't restate "this zeroes the struct" (obvious from `memset`), it records the specific, non-obvious reason `memset` was chosen over the more natural-looking `empty_data = {}` alternative that sat here through rounds 1-3. That is precisely the future-regression mitigation round 6 asked for: a plain revert-to-`{}` by a future "simplification" is not caught by this project's CI (GTest/Valgrind can't observe a difference on GCC/Clang, confirmed again this round, see §4), so the reason has to live as a comment for the next author/reviewer to see.
- Style/placement: a single-line, non-Doxygen `/* ... */` rationale comment immediately above the line it explains is an existing, common pattern in this codebase — confirmed by grepping `gst/nnstreamer/` for the same `  /* Capitalized sentence */` shape, which turns up in `tensor_filter_common.c`, `tensor_filter.c`, `gsttensor_converter.c`, `gsttensor_transform.c`, `nnstreamer_subplugin.c`, etc. So this isn't a new, out-of-house comment style for the file.
- Length/format: 72 columns (measured), inside the repo's 80-column limit, and it does not need a `clang-format` reflow of the surrounding block (verified below) — unlike the earlier header doc-line episode in rounds 1-2, which is the one prior case in this PR where a comment *did* cause CI-relevant reflow.

### 3. Are rounds 1-6 and the bot review still resolved at this head? — Yes, all confirmed, nothing reopened (Severity: no issue)

- Bot review NIT (aggregate-init only zeroes the union's first member) — resolved by the `memset` (round 4) and now additionally documented by this round's comment; unchanged in substance since round 4.
- Round 1 (stale test-count nit, clang-format CI-impact claim) and round 2 (doc-line "e.g." wording) — resolved, unchanged since round 3.
- Round 3 — nothing outstanding, confirmed again.
- Round 4's test-coverage gap (tests only checked the union's first member) — closed by round 5's `_dirty_event_record`/`_event_record_is_zeroed` helpers; unchanged this round, still present and still exercised (verified: `suspendResume`/`isUpdatable` still call `_event_record_is_zeroed()`).
- Round 5's two nits: #3 (`= {}` vs `memset` mutation gap) is the one this round's amend documents (see §2/§4); #4 (`num_events[RESUME + 1]` OOB fragility) was fixed in round 6 and is unchanged/still present (`EXPECT_LT ((guint) ops, G_N_ELEMENTS (num_events))` guard still in place).
- Round 6 — its only ask (the comment) is exactly what this amend adds; nothing else from round 6 is outstanding.

### 4. Fresh check: does the amend introduce anything new, and does the `= {}` mutation gap still hold as "acceptable, disclosed" after adding the comment? — No new issue; gap is unchanged, now explicitly documented (Severity: nit, non-blocking — inherited/unchanged from round 5/6, not a new finding)

Re-reproduced the round 5/6 mutation directly at this head: reverting `tensor_filter_support_cc.cc`'s fix to `GstTensorFilterFrameworkEventData empty_data = {}; data = &empty_data;` (dropping the `memset`, keeping the new comment orphaned above it) still passes all 8 `unittest_filter_event` cases on this repo's GCC toolchain, for the same reason rounds 5-6 already gave: GCC/Clang zero the whole object for this pattern in practice, so neither GTest nor Valgrind's memcheck observes a difference. This is not a new gap and not worsened by this round's amend — the production code already ships the objectively-correct `memset`, and the comment's entire purpose is to keep a future reviewer from silently reintroducing the weaker pattern in the first place, which is a reasonable, proportionate mitigation for a defect class with no live instance today (as rounds 5 and 6 both already concluded). No action needed.

### 5. B8/B13 re-verified end to end at the current head — Verified (Severity: no issue)

- **B8**: `gst/nnstreamer/tensor_filter/tensor_filter_common.c:1735`, `memset (&data, 0, sizeof (data))` before `gst_tensors_parse_layouts_string()`, still present and unchanged since round 1.
- **B13**: `cpp_eventHandler()` in `tensor_filter_support_cc.cc` still zero-fills `empty_data` and reassigns `data = &empty_data` when the incoming pointer is NULL, before the single `*data` call site; still the shape rounds 5-6 verified.
- Rebuilt a fresh disposable copy of the reference tree at the current head (`~/nns-4920-b8b13-review7`, deleted after use) with a forced relink of `libnnstreamer(.so/.a)`, `libnnstreamer-single(.so/.a)` and the affected test binaries (needed because meson's `restat`-based symbol-file dependency otherwise skips relinking a test binary when a `.so`'s exported symbol *set* is unchanged, which is the case for a comment-only edit — noted here only as a build-graph quirk, not a defect in the PR). Full results:
  - `unittest_filter_event`: **8/8** pass.
  - `unittest_filter_single`: **6/6** pass.
  - `unittest_plugins --gtest_filter='testTensorFilterCppSubplugin.*:*ilter*'`: **12/12** pass.
  - `unittest_sink`: **137/138** pass; the one failure, `tensorStreamTest.tensorDecoderProperty1`, is the same pre-existing, tensor_decoder-only, PR-unrelated failure earlier rounds already identified (confirmed here again as out of this PR's scope, per the task's own note).
- `clang-format --style=file` round-trips `tensor_filter_support_cc.cc`, `nnstreamer_cppplugin_api_filter.hh`, and `unittest_filter_event.cc` with **zero diff**. `gst-indent` on `tensor_filter_common.c` reproduces only the same pre-existing, out-of-hunk line-wrap at line ~514 (`ml_loge (...)`) rounds 3/5 already flagged as unrelated to this PR.

### 6. Do the two commits each build and pass independently? — Verified, yes (Severity: no issue)

- **B13 / full head (`b7e15986`)**: see §5 above — clean rebuild, 8/8 + 6/6 + 12/12 + 137/138 (one pre-existing unrelated failure).
- **B8 alone (`23473f90`)**: built a second fresh disposable tree from the same reference tree, replaced `tensor_filter_common.c`, `tests/meson.build`, and `tests/nnstreamer_filter_event/unittest_filter_event.cc` with their exact `23473f90` snapshots and reset `tensor_filter_support_cc.cc` / `nnstreamer_cppplugin_api_filter.hh` to that commit's (pre-B13) content, `meson --reconfigure` + full `ninja -j4`: clean build, no errors. `unittest_filter_event` (the 3-case B8-only version at that commit): **3/3** pass (`setInputLayout`, `setOutputLayout`, `setLayoutRefused_n`). `unittest_filter_single`: **6/6**. Filtered `unittest_plugins`: **12/12**. Confirms the PR body's "each commit builds and passes on its own" claim still holds after the round-7 amend (which only touches the B13 commit; B8's own commit content is byte-for-byte unchanged since round 6, per `git diff` showing zero difference in `23473f90` across rounds).

### 7. Anything else new this round? — No (Severity: no issue)

- Scope/size: this round's total diff from round 6 is 1 line, 1 file (`tensor_filter_support_cc.cc`). No other module touched, no architecture/API change, no additional documentation owed.
- CI on the current head was mostly pending at review time; `Build on MacOS-latest with Meson` and `DCO` had already completed and are green, consistent with the pattern in every prior round (CI takes a while to cycle through the full matrix on this repo).
- No security/backdoor concern; comment-only change cannot affect runtime behavior.

## Remaining open items: none.

All findings from relayed rounds 1-6 and the `myungjoo-bot` formal review are resolved and re-confirmed at the current head. This round's own check (comment accuracy/appropriateness, full resolution sweep, B8/B13 code-path re-verification, and independent build/test of both commits) found no new issues, blocking or otherwise.

**Merge recommendation: ready to merge.**

myungjoo:
Approved by Astra

PR #5000, opened 2026-09-17
[Converter] Free the mode options and tensors configs the element replaces or drops (#4920 C8, C9)

myungjoo:
item **C8**, **C9** of #4920

- **What this fixes.** Memory leaks in `tensor_converter`. **C8**: the `mode` setter overwrote the previous option string and the `custom-script` framework name without freeing them, and returned without freeing its split vector when a `custom-code` name is not registered. **C9**: with more than 16 tensors a `GstTensorsConfig` owns a heap `extra` array, and the element dropped such configs without `gst_tensors_config_free ()` — per buffer on the flexible-to-static chain path (matching, refused and replaced configs), on every caps renegotiation and every `parse_caps` failure, for every peer config read in `parse_octet` and in each sink caps query, and when `parse_custom` re-initialized a peer config.
- **How.** Free before overwrite / on every exit, no behavior change: `g_free` before the two `g_strdup`s and a freed-vector `break` in the setter; in the chain, `parse_octet` and `parse_caps` every exit that drops a config now frees it (single `goto` exit labels in the two parsers), the live config is freed before it is replaced, and `parse_custom` frees the peer config instead of re-initializing it. One commit per item; production diff is +39/-17 in `gsttensor_converter.c`.
- **Result.** New unconditional binary `unittest_converter_config` (15 cases, 6 `_n`), so the Ubuntu meson job's valgrind leak gate runs it. Under memcheck the fixed build has 0 errors and 0 definite leaks of ours; `main`'s converter gives definite leaks of ours in all 14 leak cases, and a 13-row mutation matrix (each single free removed) fails the gate in exactly the case(s) for that path. Converter-related gtests (`unittest_sink` 138, `unittest_common` 171, converter cases of `unittest_plugins`, `unittest_converter_timestamp`) and 16 SSAT groups (converter, octet_stream, mux/demux/split/merge, transform, decoder, ...) pass identically on the fix and on `main`.

## Details (for reviewers and agents)

**Why a new binary.** `tests/nnstreamer_converter/unittest_converter.cc` is built only `if flatbuf_support_is_available and have_python3`, so the valgrind job never builds it; `unittest_plugins.cc` is owned by open PRs (#4980, #4963). The leak cases do not change what the element does, so natively they pass on `main` too; the witness is `tools/debugging/check_valgrind_log.sh` (definite leak attributed to this repository → exit 1). Behavior cases (getter value, output tensor count/bytes, refused caps/buffers, renegotiated caps) are checked natively.

**How 17 tensors reach the chain path independent of #4980.** On `main`, `gst_tensor_buffer_append_memory ()` records an extra tensor's size from the info the caller passes when the meta header says `static` (the default of `gst_tensor_meta_info_init ()`); #4980 (A6) records the memory size. The test appends each memory with an info describing the whole memory (header included), so both give the same extra info and the converter's per-tensor size check passes on either base.

**External converter path** (`parse_custom` → `get_out_config`): reached by registering a fake converter whose `query_caps` names `python3`, the mimetype `custom-script` mode looks up (`findExternalConverter ()` matches by caps structure name and checks registered sub-plugins first), so the sink pad template is not involved.

**Mutation matrix** (valgrind + checker per row, all rows 15/15 native pass):

| row | removed | gate | leaking case(s) |
|---|---|---|---|
| main | all | fail | all 14 leak cases |
| M01 | `g_free (mode_option)` | fail | modeSetTwice, modeUnregisteredCustomCode_n, modeRegisteredAfterUnregistered |
| M02 | `g_strfreev` before `break` | fail | modeUnregisteredCustomCode_n, modeRegisteredAfterUnregistered |
| M03 | `g_free (ext_fw)` | fail | modeSetTwice |
| M04 | chain size-mismatch free | fail | flexToStaticExtraTensorsSize_n |
| M05 | chain property-mismatch free | fail | flexToStaticExtraTensorsProperty_n |
| M06 | chain free before replace | fail | flexToStaticExtraTensorsReconfigure |
| M07 | chain free when equal | fail | flexToStaticExtraTensors |
| M08 | `parse_octet` peer free | fail | octetPeerExtraTensors, octetPeerExtraTensorsFrames_n |
| M09 | `parse_custom` custom-code free | fail | customCodeUnfixedPeerExtraTensors |
| M10 | `parse_custom` free before `get_out_config` | fail | externalUnfixedPeerExtraTensors |
| M11 | `get_possible_media_caps` free | fail | mediaCapsPeerExtraTensors |
| M12 | `parse_caps` error free | fail | customCodeFixedPeerMismatch_n |
| M13 | `parse_caps` free before replace | fail | renegotiateExtraTensors |

**Other checks.** `gst-indent`/`indent.sh` (its only reflow is a pre-existing line of `main`, left alone), clang-format 16.0.6, `doxygen-tag.sh` with ctags (negative control: dropping one `@brief` fails), and a `-Werror -Wextra -Wsign-compare` compile of both changed objects are clean. Each commit builds and passes on its own.

**Not changed.** The existing `tensors_config` readers all run in the streaming thread that replaces it (`query_caps` uses a local peer config), so freeing the old config opens no new cross-thread window. `CHANGES` untouched.


myungjoo:

**Verdict: mergeable as-is** (one Low-severity nit noted below, not blocking).

## Scope check

- Addresses exactly C8/C9 of #4920 (converter `mode`-setter leaks and `GstTensorsConfig`/`extra`-array leaks in `tensor_converter`). No unrelated modules touched; production diff is contained to `gst/nnstreamer/elements/gsttensor_converter.c` (+39/-17), plus one new test file and a `tests/meson.build` registration. Size is proportionate to the fix plus its test coverage.
- I independently traced every changed call site (`gst_tensor_converter_set_property`, the flex-to-static branch of `gst_tensor_converter_chain`, `parse_octet`, `parse_custom`, `get_possible_media_caps`, `parse_caps`) against `gst_tensors_config_init/free`, `gst_tensors_info_free` (in `nnstreamer_plugin_api_util_impl.c`) and `gst_tensor_buffer_append_memory` (in `nnstreamer_plugin_api_impl.c`). Findings:
  - `gst_tensors_info_free()` re-inits after freeing, so `gst_tensors_config_free()` is idempotent — no double-free risk from the added calls.
  - `gst_tensor_buffer_append_memory()` consumes a `GstTensorInfo` by value (encodes it into the memory's own header), it does not retain a pointer into the config's `info`/`extra` array, so freeing `tmp`/`config` right after appending its memories into `inbuf` (chain function, flex-to-static case) is safe.
  - The new `gst_tensors_config_init (&config);` added at the top of `gst_tensor_converter_parse_caps` is not redundant: the video-not-supported and audio-not-supported branches now `goto error;` (which calls `gst_tensors_config_free (&config)`) *before* any `parse_video`/`parse_audio` call would have initialized `config`. Without that added init, those two paths would free uninitialized stack memory. Good catch, correctly reasoned in the diff.
  - `gst_tensor_converter_parse_octet`'s new `peer` free is correctly guarded by an unconditional `gst_tensors_config_init (&peer)` up front, so the `done:` label's `gst_tensors_config_free (&peer)` is safe whether or not `gst_tensors_config_from_peer()` ran.
  - Chain-function ownership transfer (`gst_tensors_config_free (config); *config = tmp;`) vs. the `else` branch's `gst_tensors_config_free (&tmp);` are mutually exclusive (`if`/`else`), so there is no double free of `tmp`'s heap `extra` array after transfer.
  - Verified the `return;` → `g_strfreev (strv); break;` change in the property setter is behavior-preserving: the `switch` is the last statement in `gst_tensor_converter_set_property`, so `break` and `return` are equivalent there.
  - Verified the `self->tensors_config`/`self->externalConverter` readers the PR body claims are streaming-thread-only (`_gst_tensor_converter_chain_segment/timestamp/octet/flex_tensor`, `GST_EVENT_CAPS` handling) do run in the same serialized streaming-thread context as `parse_caps`; `get_possible_media_caps`/`query_caps` use a separate local `config`, not `self->tensors_config`, so the new frees there don't add a cross-thread window.

## CI-gate verification (per the review's own checklist item)

- Confirmed `unittest_converter_config` is registered unconditionally inside the same `if gtest_dep.found()` block as `unittest_converter_timestamp` in `tests/meson.build` (unlike `unittest_converter`, which is gated on `flatbuf_support_is_available and have_python3`), so `packaging/run_unittests_binaries.sh --valgrind --recursive ...` (invoked from `.github/workflows/ubuntu_clean_meson_build.yml`) will pick it up.
- Confirmed via `gh api repos/nnstreamer/nnstreamer/branches/main/protection` that "Build with meson and test with Valgrind in Ubuntu (ubuntu-22.04)" is in `required_status_checks.contexts` for `main` — so a future regression that reintroduces one of these leaks will fail a required check and block merge, not just warn.
- Spot-checked several of the 15 new cases against the PR's own mutation-matrix table (e.g. `flexToStaticExtraTensorsSize_n`/`flexToStaticExtraTensorsProperty_n` against the two `goto error` frees added in the chain's flex-to-static block, `customCodeFixedPeerMismatch_n` against the `parse_caps` tensor-info-mismatch `goto error`, `renegotiateExtraTensors` against the `self->tensors_config` replace-before-free). The mapping holds up under reading the code, not just under the PR's own claims.
- No architecture/API change, so no doc update is expected; none is missing.

## Finding

**[Low] Test isolation: shared custom-converter name can turn one real failure into two confusing ones**
`tests/nnstreamer_converter/unittest_converter_config.cc`, lines 797-822 (`customCodeUnfixedPeerExtraTensors`) and 859-880 (`customCodeFixedPeerMismatch_n`), also 585-607 (`modeRegisteredAfterUnregistered`).

Three separate `TEST()` cases register the same custom-converter name `"c8c9_split"` via `nnstreamer_converter_custom_register` and only unregister it at the very end of the test body. In `customCodeUnfixedPeerExtraTensors`, `ASSERT_TRUE (out != NULL)` (line ~815) sits between the register and unregister calls. If that assertion fails — which is exactly the kind of regression this test exists to catch — the test function returns immediately, `gst_harness_teardown()` and `nnstreamer_converter_custom_unregister ("c8c9_split")` are both skipped, and `"c8c9_split"` stays registered in the process-wide subplugin table. The next test that also registers `"c8c9_split"` (`customCodeFixedPeerMismatch_n`, run later in the same binary) will then fail its own `ASSERT_EQ (0, nnstreamer_converter_custom_register (...))` too, with an unrelated-looking "already registered" style failure.

This does not create a false negative — the real regression still fails CI — but it produces a second, misleading failure that costs time to triage. Suggest either giving each test its own unique custom-code name, or unregistering via a scope guard (e.g. a small RAII helper, or `testing::ScopedTrace`-style cleanup / a `TearDown()` on a shared fixture) that runs regardless of which `ASSERT_*` exits the test early.

## Not flagged (pre-existing, out of scope)

Per the review's scoping instruction, I did not flag remaining `gst_tensors_config`/`GstTensorsInfo` ownership patterns elsewhere in `gsttensor_converter.c` that this PR does not touch (e.g. the `_NNS_MEDIA_ANY`/custom-func branch of the chain function, which already frees correctly on `main` and was left alone) — those are outside the diff under review.

---
*This review was produced by a separate Claude agent session and is being relayed here verbatim; it was not written by the PR author reviewing their own change.*

myungjoo:

Addressed the Low finding (test isolation) in-PR by giving each test its own custom-converter name (`c8c9_split_mode`, `c8c9_split_unfixed`, `c8c9_split_mismatch`), so a failed `ASSERT` between register and unregister cannot make a later test fail on a duplicate name. The change is folded into the commit that introduced each case (head `ccbb88c9`); production code is unchanged, and the new tree gives 15/15 natively and a clean valgrind gate (0 error / 0 definite leak contexts of ours). Commit 1 was re-verified standalone (4/4, gate clean, strict `-Werror -Wextra -Wsign-compare`).

myungjoo:

**Verdict: mergeable as-is.** No blocking issues found.

## Scope and correctness

- Diff is contained to `gst/nnstreamer/elements/gsttensor_converter.c` (production, +39/-17), a new `tests/nnstreamer_converter/unittest_converter_config.cc` (671 lines, 15 `TEST()` cases), and its registration in `tests/meson.build`. This is exactly items C8 and C9 of #4920 (mode-setter leaks, `GstTensorsConfig`/`extra`-array leaks) — no unrelated modules touched, and size is proportionate to the fix plus the coverage it needs.
- I independently re-traced every changed call site against `gst_tensors_config_init/free`, `gst_tensors_info_free`/`_copy` and `gst_tensors_config_from_peer` (in `nnstreamer_plugin_api_util_impl.c` / `nnstreamer_plugin_api_impl.c`) rather than relying on the prior review's conclusions:
  - `gst_tensors_config_free()` → `gst_tensors_info_free()` re-inits after freeing (nulls `info->extra`), so it is idempotent and safe to call on an empty/never-populated config — confirmed by reading the implementation, not assumed.
  - `gst_tensor_converter_chain`, flex-to-static branch: `config = &self->tensors_config`, so `gst_tensors_config_free (config); *config = tmp;` frees the live config in place before the struct-copy transfers ownership of `tmp`'s heap `extra` array; the `else` branch frees `tmp` instead. Mutually exclusive `if`/`else`, no double free. The two new `goto error` frees (`&tmp`) are on a distinct local from the pre-existing `error:` label's `gst_tensors_config_free (&new_config)` (unrelated, always-empty on this path) — no double free there either.
  - `gst_tensor_converter_parse_caps`: the new `gst_tensors_config_init (&config)` at entry is required because the video/audio-support-disabled branches now `goto error` (which frees `config`) before any `parse_video`/`parse_audio` call would otherwise have initialized it — without the added init those two paths would free uninitialized stack memory. Verified by reading the two `else` branches directly.
  - `gst_tensor_converter_parse_octet`: `peer` is unconditionally `gst_tensors_config_init`'d up front, and when `configured` is true, `info = &peer.info` is later deep-copied via `gst_tensors_info_copy()` (verified: this allocates its own `extra` array per `gst_tensor_info_copy`, not a shared pointer) before `gst_tensors_config_free (&peer)` runs at `done:` — no use-after-free or double-free between `config->info.extra` and `peer.info.extra`.
  - `gst_tensor_converter_get_possible_media_caps`: `gst_tensors_config_from_peer()` calls `gst_tensors_config_init(config)` unconditionally at its own top regardless of return value, so the new `gst_tensors_config_free (&config)` before return is safe even on the `if` being false (not freeing garbage stack memory).
  - `gst_tensor_converter_parse_custom`: both `config` arguments it receives are freshly `gst_tensors_config_init`'d by the caller (`parse_caps`) before the switch reaches this branch, so the new `gst_tensors_config_free (config)` calls (replacing the old re-init) only ever free what `gst_tensors_config_from_peer()` itself allocated moments earlier — not stale data from a previous caps negotiation.
  - The `return` → `g_strfreev (strv); break;` change in `gst_tensor_converter_set_property` is behavior-preserving: `PROP_MODE` is the last case before `default`, and the `switch` is the function's last statement, so `break` and `return` are equivalent there; `self->custom.func` reset semantics on the refused-name path are unchanged.
- Per-commit sanity: commit 1 (`0bb2fe5d`, C8) stands alone — its own `unittest_converter_config.cc` (183 lines) does not reference anything commit 2 adds, and every new static helper/TEST in both commits is exercised at least once (checked by grep) and carries a Doxygen `@brief`. Commit 2's diff on the test file is a proper incremental patch on commit 1's version, not a rewrite.

## CI-gate verification

- `unittest_converter_config` is registered unconditionally in the same `if gtest_dep.found()` block as `unittest_converter_timestamp` (not gated like `unittest_converter`, which needs `flatbuf_support_is_available and have_python3`), so the Valgrind step of `.github/workflows/ubuntu_clean_meson_build.yml` picks it up unconditionally.
- `EXTRA_TENSORS = NNS_TENSOR_MEMORY_MAX + 1 = 17`, well within `NNS_TENSOR_SIZE_EXTRA_LIMIT = 240`, so the extra-array paths exercised by the new tests are representative and don't rely on undefined overflow behavior.
- No architecture or public API change, so no doc update is expected or missing.

## Re-verification of the prior Low finding

The earlier review (comment above) flagged that three `TEST()` cases shared the custom-converter name `"c8c9_split"`, registered/unregistered per-test, such that a real assertion failure between register and unregister in one test would leak the registration and cause an unrelated-looking failure in a later test using the same name. I checked the current head (`ccbb88c9`) directly: `modeRegisteredAfterUnregistered` now registers `"c8c9_split_mode"`, `customCodeUnfixedPeerExtraTensors` registers `"c8c9_split_unfixed"`, and `customCodeFixedPeerMismatch_n` registers `"c8c9_split_mismatch"` — three distinct names, confirmed via `grep` to be used nowhere else in `tests/`. The fix is folded into the commit that introduces each case (not a separate fixup commit), matching this repo's "no fix-up commits" convention. The finding is resolved.

## Not flagged (pre-existing, out of scope)

Per the review's scoping instruction, I did not flag the `_NNS_MEDIA_ANY` custom-func/external-converter branch of the chain function (lines ~1298-1304), which already frees correctly on `main` and is untouched by this diff, nor any other `GstTensorsConfig` ownership pattern this PR does not touch.

---
*This review was produced by a separate Claude agent session and is being relayed here verbatim; it was not written by the PR author reviewing their own change.*

myungjoo:
Approved by Astra

